PR #2571, opened 2022-09-28
Cargo update

RalfJung:


RalfJung:
Lol, proc-macro2 had an update while I was preparing this PR... let's just include that as well. `cargo outdated` is clean now on both miri and cargo-miri.

@bors r+

PR #2572, opened 2022-09-29
bump rustc-build-sysroot

RalfJung:
This should fix sysroot build problems on nix.


RalfJung:
wtf... 'cache not found' but then it also didn't run the next step to install rustup-toolchain-install-master?

PR #2568, opened 2022-09-26
Merge rustc changes back into the miri repo

oli-obk:
This uses https://github.com/josh-project/josh (RIIR ftw)

The first time it looks a bit odd with the merges and so, all future syncs will work just like the instructions say. I'll add instructions for the sync from miri to rustc, too, starting to do that right now.

Initial cache building now takes 5 mins 🥳

PR #2573, opened 2022-09-29
Rustup

RalfJung:
@oli-obk trying your instructions for the first time. This seems to work fine (docker actually makes it quite convenient :D ), but it does create a copy of each merge commit, which over time will take a lot of room in our history I think.


RalfJung:
Hm, actually the history of this branch is strange. It seems entirely disconnected from master?

oli-obk:
I don't think we can avoid the merge commits with `josh`. I wonder if things break if we just rebase this PR and let josh figure it out during the other direction (though I have screwed up things enough for that to get even more confusing. We're hitting a josh bug now due to our multiple "unrelated" histories).

RalfJung:
I rebased and updated this PR accordingly. The history indeed looks much better now.

But that old history is something where josh could do better I think, so it might make sense to bring this up with the josh devs? https://github.com/josh-project/josh/issues/952

oli-obk (approved):

RalfJung:
Can we try what future syncs will look like before merging this, somehow?

oli-obk:
> Can we try what future syncs will look like before merging this, somehow?

wait for a rustc PR to get merged and sync again?

Or just sync again right now 😆

RalfJung:
Hm, I guess what that will do is that `josh/master` constructs parallel history and then git has increasing amounts of work to do in the rebase to deduplicate all that stuff...

RalfJung:
Yeah that does not look good
```
$ git merge josh/master
Auto-merging tests/fail/invalid_bool.rs
CONFLICT (content): Merge conflict in tests/fail/invalid_bool.rs
Automatic merge failed; fix conflicts and then commit the result.
```

oli-obk:
does it work cleanly with a rebase instead of a merge? Like we need to rebase on every sync, not just this one

RalfJung:
For this sync I did
```
git merge josh/master
git rebase origin/master
```
The first step already fails now the simulating the next sync.

`josh/master` has completely separate history, I don't even know how to do a rebase without first merging...

RalfJung:
So I moved this PR back to just `git merge josh/master`. But unfortunately, then doing `git merge josh-ralf/master` also leads to conflicts...

RalfJung:
Ah, this updates proc-macro2 which switched from unicode-xid to unicode-ident.

RalfJung:
We could consider adding the `:linear` filter to our josh remote, that makes for a much smaller history -- basically one commit per rustc PR that changed Miri. Slightly sub-optimal for rustc PRs that do heavy Miri work though, but that is rare, and it keeps the history here much less noisy.

Of course that will duplicate some stuff since we already merged the other josh branch. We should probably first figure out what we want and then merge more things. ;)

RalfJung:
Given that this matches https://github.com/rust-lang/miri/pull/2568, I'm going to land this now so that our CI cron job stays green.  We already have josh history in this repo so if we want to change that we need a force-push anyway.

@bors r+

PR #2576, opened 2022-09-29
use rustc_tools_util instead of vergen

RalfJung:
This avoids some 0BSD-licensed dependencies (which might or might not be a license we can use), and also just generally removes tons of dependencies and shares some code with clippy. Well, kind-of-shares -- clippy uses a path dependency (https://github.com/rust-lang/rust-clippy/issues/9553) so rustc will still build this twice in the rustc repo.


RalfJung:
This seems to work well enough for now and we can always tweak it later if needed to share more code with clippy. And ut unblocks the rustc sync.
@bors r+

PR #2554, opened 2022-09-20
Miri subtree sync

oli-obk:
replaces https://github.com/rust-lang/miri/pull/2550

PR #2583, opened 2022-10-04
initial josh subtree sync

RalfJung:
This demonstrates what a josh-based rustup would look like with my patched josh. To create it I did
```
git fetch http://localhost:8000/rust-lang/rust.git:start=75dd959a3a40eb5b4574f8d2e23aa6efbeb33573[:prefix=src/tools/miri]:/src/tools/miri.git master 
git merge FETCH_HEAD
./rustup-toolchain HEAD && ./miri fmt
git commit -am rustup
```
Unlike the [previous attempt](https://github.com/rust-lang/miri/pull/2554), this does not add a new root commit to the repo.

Once we merge this, we committed to using josh for subtree syncing, and in particular a version of josh that includes https://github.com/josh-project/josh/pull/961 (or something compatible).


RalfJung:
@oli-obk would be good if you could also check this out and take a look at the tree.

oli-obk:
Yea, this looks good now. Starting at the sync we see merge commits from rustc, but most of them are just merge commits of the B sort below, where we merge with a commit that is already in the history (X).

```
X -- Y
 \    \
   --- B
```

This is much better than the previous sync. Looking at the other direction now

RalfJung:
Also see https://github.com/josh-project/josh/issues/952#issuecomment-1264598227 for why these merge commits arise.

RalfJung:
I got the exact same sync commit with the other josh branch from https://github.com/josh-project/josh/pull/965 (by the josh devs) :)

RalfJung:
Given that two independent implementations of this josh feature produced the exact same commit -- should we risk it and land this before things land in josh upstream? I am not sure what their timeline is.

oli-obk:
worst case we'll force push again in the future

oli-obk:
so yea, let's go!

RalfJung:
We also have `MIRI_BLESS=1 ./run-test.py` now for test-cargo-miri. :)

@bors r=oli-obk

PR #2585, opened 2022-10-07
Add -Zmiri-always-two-phase

saethlin:
@RalfJung I think a while ago you said we could make all `Unique` retags two-phase, which would have the effect of delaying tag invalidation until the mutable reference is actually used for a write, as opposed to just reborrowed.

It occurs to me that we should at least be able to try out that idea in Miri?

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,6 +332,9 @@ The remaining flags are for advanced use only, and more likely to change or be r
 Some of these are **unsound**, which means they can lead
 to Miri failing to detect cases of undefined behavior in a program.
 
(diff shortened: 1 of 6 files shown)


RalfJung:
I am not sure what I said, but I think this will just make all mutable references effectively raw pointers. Which doesn't make a lot of sense?

saethlin:
:joy: Jakob warned me that would be the case too.

Search back in Zulip eventually reveals this aaaaaa

> Once we have a satisfying model of two-phase borrows, and we have a proof that this supports the optimizations &mut is supposed to enable, then yeah this could be a reasonable option.

I'll comment on the Zulip topic.

PR #2600, opened 2022-10-18
Fix ICE when trying to GC a Stack with an unknown bottom

saethlin:
Fixes https://github.com/rust-lang/rust/issues/103167

@RalfJung I prefer this approach because the whole GC system is sloppy already in order to be efficient (doesn't run often, ignores small stacks) so a bit more imprecision for a simple implementation seems worth it to me. But I'm of course willing to be convinced otherwise.

### DIFF
--- a/src/stacked_borrows/stack.rs
+++ b/src/stacked_borrows/stack.rs
@@ -43,10 +43,14 @@ impl Stack {
     pub fn retain(&mut self, tags: &FxHashSet<SbTag>) {
         let mut first_removed = None;
 
-        // For stacks with a known bottom, we never consider removing the bottom-most tag, because
-        // that is the base tag which exists whether or not there are any pointers to the
-        // allocation.
-        let mut read_idx = if self.unknown_bottom.is_some() { 0 } else { 1 };
+        // We never consider removing the bottom-most tag. For stacks without an unknown
+        // bottom this preserves the base tag.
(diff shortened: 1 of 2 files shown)


src/stacked_borrows/stack.rs
```diff
@@ -43,10 +43,11 @@ impl Stack {
...
         let mut first_removed = None;
 
-        // For stacks with a known bottom, we never consider removing the bottom-most tag, because
-        // that is the base tag which exists whether or not there are any pointers to the
-        // allocation.
-        let mut read_idx = if self.unknown_bottom.is_some() { 0 } else { 1 };
+        // We never consider removing the bottom-most tag.
+        // For stacks without an unknown bottom this preserves the base tag. Not considering the
+        // bottom-most tag for removal when we do have an unknown bottom does not have a
+        // significant impact on the degree to which the GC mititages memory growth.
```
RalfJung, line 49: No objection to doing this, but I don't understand why it can ICE -- the comment should explain why the obvious thing (the code before this PR) does not work.

saethlin:
```
            let left = self.borrows[read_idx - 1];
```
If `read_idx` is `0`, this is going to either panic due to out of bounds or integer overflow.

I've rewritten the comment to try to explain this.

RalfJung:
Ah I see, makes sense. Thanks!

@bors r+

PR #2606, opened 2022-10-19
Magic functions for writing to stdout/stderr.

cbiffle:
This enables I/O in no_std contexts (or, really, any Miri-specific OS-independent context). Combined with the `abort` intrinsic it should allow a reasonable test framework in no_std.

**Question for maintainers:** So, the `no_std` panic test needs work, for two reasons:

- First, its stdout includes Miri's whole message about the abort intrinsic having been used. I guess whatever panic handler you use in `std` contexts exits cleanly without triggering this message.
- Second, likely for the same reason, the test framework appears to expect the test to exit successfully, when in fact it exits with status 1 due to the abort.

What shall I do there?


RalfJung:
For the exist status, `tests/panic` is for tests with exit status 101.
If exit status 1 is expected here, this test should be in `tests/fail`.

cbiffle:
Fair enough! Every project has different opinions on golden tests like that.

I have moved the test and retained the output.

README.md
```diff
@@ -542,6 +542,16 @@ extern "Rust" {
     /// In particular, users should be aware that Miri will periodically attempt to garbage collect the
     /// contents of all stacks. Callers of this function may wish to pass `-Zmiri-tag-gc=0` to disable the GC.
     fn miri_print_stacks(alloc_id: u64);
+
+    /// Miri-provided extern function to print (from the interpreter, not the
+    /// program) the contents of a section of program memory, as bytes. Bytes
+    /// written using this function will emerge from the interpreter's stdout.
+    fn miri_write_to_stdout(ptr: *const u8, len: usize);
```
saethlin, line 549: Why is pointer + usize better than a slice?

cbiffle: Good question! I did this because I couldn't find existing examples of passing a slice in to a Miri-implemented extern function -- but maybe I didn't look hard enough! If you can point me at an example I'm happy to use a slice here.

cbiffle:
FYI, I have figured out how your `//~` patterns work (...I think) and fixed `tests/fail/no_std.rs` to use them.

RalfJung:
To support slices, you have to `read_immediate` the argument and then call `to_scalar_pair` on the result, and then `to_pointer` on the first component and `to_machine_usize` on the 2nd.

tests/fail/no_std.rs
```diff
@@ -0,0 +1,41 @@
+#![feature(lang_items, start, core_intrinsics)]
+#![no_std]
+// windows tls dtors go through libstd right now, thus this test
+// cannot pass. When windows tls dtors go through the special magic
+// windows linker section, we can run this test on windows again.
+//@ignore-target-windows
+
+// Plumbing to let us use `writeln!` to host stderr:
```
RalfJung, line 8: Please move the test to `tests/fail/panic`.

5225225:
If the API is meant to work on bytes, that might be worth testing. As in, have a test that emits something that's invalid UTF-8. That might make the test harder to read though, so I'll defer to other people as to if that's a useful test to have.

cbiffle:
> If the API is meant to work on bytes, that might be worth testing. As in, have a test that emits something that's invalid UTF-8. That might make the test harder to read though, so I'll defer to other people as to if that's a useful test to have.

Hmm. I do intend it to work on bytes, so that it can be used beneath implementations of (say) `std::io::Write` that may write partial buffers of bytes that _individually_ are invalid UTF-8, even if the overall task is a UTF-8 `println`. What sort of breakage would you imagine looking for in a test?

cbiffle:
- Moved test to `tests/fail/panic` (though note that it doesn't use the same panic path as the other tests in that dir)
- Converted API to slices.
- Deboilerplated implementation, which was starting to bother me

5225225:
> What sort of breakage would you imagine looking for in a test?

Only thing I can think of is "something along the chain does String::from_utf8_lossy and now your output is replacement characters".

I recall miri stdout having an issue with that in the past, but it seems to not do that now. But yeah, that's what I'd expect to be the most likely failure case, so just outputting some byte (`0xff`) that can never happen in UTF-8 would be fine.

README.md
saethlin: You updated the implementation, don't forget the docs :)

cbiffle:
> I recall miri stdout having an issue with that in the past, but it seems to not do that now. But yeah, that's what I'd expect to be the most likely failure case, so just outputting some byte (`0xff`) that can never happen in UTF-8 would be fine.

Attempting to do this causes something in compiletest to crash.

```
     Running tests/compiletest.rs (target/debug/deps/compiletest-5625b9d9b19cedc6)
## Running ui tests in tests/pass against miri for x86_64-unknown-linux-gnu
   Compiler flags: ["--error-format=json", "--edition", "2018", "-Dwarnings", "-Dunused", "-Zui-testing", "--target", "x86_64-unknown-linux-gnu"]
The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value: Utf8Error { valid_up_to: 31, error_len: Some(1) }
Location: /home/cbiffle/.cargo/registry/src/github.com-1ecc6299db9ec823/ui_test-0.3.1/src/lib.rs:467
```

This is because the test framework does this (`ui_test/src/lib.rs` line 466):

```rust
    let diagnostics = rustc_stderr::process(path, stderr);
    let stdout = std::str::from_utf8(stdout).unwrap();  // <-- not good
```

So, evidently, tests are forbidden from printing anything in non-UTF-8. Miri, however, handles it find, as attested by being able to `./miri run` that same test case.

cbiffle:
I've updated the README to reflect the new function signatures, fwiw.

RalfJung:
Looks good, thanks! @bors r+

If we want a non-UTF-8-test for this, we also want one for regular stdout/stderr for sure.

PR #2611, opened 2022-10-21
Update libffi-sys to 2.0.1

pfmooney:
Prior version of libffi [could not be cross-compiled to illumos](https://github.com/tov/libffi-rs/pull/59) due to host-triple complications.  This should fix rustup builds of miri for the illumos platform.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -320,9 +320,9 @@ dependencies = [
 
 [[package]]
 name = "libffi-sys"
-version = "2.0.0"
+version = "2.0.1"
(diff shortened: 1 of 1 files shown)


RalfJung:
This lockfile is irrelevant for the builds distributed by rustup. You'll need to update the lockfile in rustc as well.

But this will help for local Miri builds.
@bors r+

pfmooney:
> This lockfile is irrelevant for the builds distributed by rustup. You'll need to update the lockfile in rustc as well.

I'll go see to that.

Thanks!

PR #2612, opened 2022-10-21
add GetFileType stub

ChrisDenton:
because I'm planning to break miri again https://github.com/rust-lang/rust/pull/103360

See [GetFileType](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfiletype)


RalfJung:
Thanks. :)
Unfortunately we are struggling with getting Miri changes synced back to rustc, but still seems better to land this here.

@bors r+

PR #2322, opened 2022-07-04
Ideas on getting information about borrow stacks during execution

saethlin:
From time to time people ask what some borrow stack looks like in some code. I just know that I am terrible at doing Stacked Borrows by hand, so I always toss together something like this.

I know that Miri has logging, but I've never found it particularly useful because there's just too much output. Also I personally don't think about exactly what the state of a borrow stack is, but this seems to be something that newcomers to Stacked Borrows always want.

Update: This has been sitting as S-waiting-on-author for a long time. I bring it out from time to time to explain Stacked Borrows to people, and just now @JakobDegen said
>  Can we please merge that btw? It's such a valuable teaching tool
>  Interfaces can be fixed later

I'm inclined to trust Jake's judgement here.

PR #2614, opened 2022-10-23
Improve miri_print_borrow_stacks

saethlin:
Per post-merge review on https://github.com/rust-lang/miri/pull/2322

* `miri_print_stacks` renamed to `miri_print_borrow_stacks`
* A bit more details in docs, clarified how unstable these functions are meant to be
* Print an `unknown_bottom` if one exists

Open question: Currently `miri_get_alloc_id` gets the expected `AllocId` for `Wildcard` pointers, but for pointers with no provenance, the function reports UB and halts the interpreter. That's definitely wrong. But what _should_ we do? Is it reasonable to check if the pointer has `None` provenance and try to get an `AllocId` for its address? That still leaves us with a failure path, which in this case might be best-handled as an ICE? I'm just not sure that changing the return type of `miri_get_alloc_id` to `Option` is a win because it complicates all normal uses of this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -538,7 +538,9 @@ extern "Rust" {
     fn miri_start_panic(payload: *mut u8) -> !;
 
     /// Miri-provided extern function to get the internal unique identifier for the allocation that a pointer
-    /// points to. This is only useful as an input to `miri_print_borrow_stacks`, and it is a separate call because
+    /// points to. If this pointer is invalid (not pointing to an allocation), interpretation will abort.
+    ///
+    /// This is only useful as an input to `miri_print_borrow_stacks`, and it is a separate call because
     /// getting a pointer to an allocation at runtime can change the borrow stacks in the allocation.
     /// This function should be considered unstable. It exists only to support `miri_print_borrow_stacks` and so
     /// inherits all of its instability.
@@ -549,6 +551,7 @@ extern "Rust" {
     /// The format of what this emits is unstable and may change at any time. In particular, users should be
     /// aware that Miri will periodically attempt to garbage collect the contents of all stacks. Callers of
     /// this function may wish to pass `-Zmiri-tag-gc=0` to disable the GC.
+    ///
     /// This function is extremely unstable. At any time the format of its output may change, its signature may
     /// change, or it may be removed entirely.
     fn miri_print_borrow_stacks(alloc_id: u64);
--- a/README.md
+++ b/README.md
@@ -538,15 +538,20 @@ extern "Rust" {
     fn miri_start_panic(payload: *mut u8) -> !;
 
     /// Miri-provided extern function to get the internal unique identifier for the allocation that a pointer
-    /// points to. This is only useful as an input to `miri_print_stacks`, and it is a separate call because
+    /// points to. This is only useful as an input to `miri_print_borrow_stacks`, and it is a separate call because
     /// getting a pointer to an allocation at runtime can change the borrow stacks in the allocation.
+    /// This function should be considered unstable. It exists only to support `miri_print_borrow_stacks` and so
+    /// inherits all of its instability.
     fn miri_get_alloc_id(ptr: *const ()) -> u64;
 
     /// Miri-provided extern function to print (from the interpreter, not the program) the contents of all
-    /// borrow stacks in an allocation. The format of what this emits is unstable and may change at any time.
-    /// In particular, users should be aware that Miri will periodically attempt to garbage collect the
-    /// contents of all stacks. Callers of this function may wish to pass `-Zmiri-tag-gc=0` to disable the GC.
-    fn miri_print_stacks(alloc_id: u64);
+    /// borrow stacks in an allocation. The leftmost tag is the bottom of the stack.
+    /// The format of what this emits is unstable and may change at any time. In particular, users should be
+    /// aware that Miri will periodically attempt to garbage collect the contents of all stacks. Callers of
+    /// this function may wish to pass `-Zmiri-tag-gc=0` to disable the GC.
+    /// This function is extremely unstable. At any time the format of its output may change, its signature may
+    /// change, or it may be removed entirely.
+    fn miri_print_borrow_stacks(alloc_id: u64);
 
     /// Miri-provided extern function to print (from the interpreter, not the
     /// program) the contents of a section of program memory, as bytes. Bytes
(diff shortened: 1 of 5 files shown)


RalfJung:
>  Is it reasonable to check if the pointer has None provenance and try to get an AllocId for its address?

No, pointers with `None` provenance conceptually do nit have an AllocId.

src/shims/foreign_items.rs
```diff
@@ -420,10 +420,17 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
                 let ptr = this.read_pointer(ptr)?;
-                let (alloc_id, _, _) = this.ptr_get_alloc_id(ptr)?;
+                let (prov, offset) = ptr.into_parts();
+                let alloc_id = if prov.is_some() {
+                    let (alloc_id, _, _) = this.ptr_get_alloc_id(ptr)?;
+                    alloc_id
+                } else {
+                    intptrcast::GlobalStateInner::alloc_id_from_addr(this, offset.bytes())
+                        .expect("pointer passed to miri_get_alloc_id must not be dangling")
+                };
```
RalfJung, line 430: What is this supposed to achieve, compared to the previous implementation? The previous implementation looked good to me, it just lacked docs.

src/stacked_borrows/mod.rs
```diff
@@ -1153,7 +1153,11 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         let alloc_extra = this.get_alloc_extra(alloc_id)?;
         let stacks = alloc_extra.stacked_borrows.as_ref().unwrap().borrow();
         for (range, stack) in stacks.stacks.iter_all() {
-            print!("{:?}: [", range);
+            print!("{:?}: ", range);
+            if let Some(bottom) = stack.unknown_bottom() {
+                print!("{:?}..", bottom);
```
RalfJung, line 1158: The meaning of this tag is 'arbitrary many items whose tag is strictly less than `bottom`'. I don't think this way of printing reflects that. So maybe what about `(unknown, < {tag})..`? Or just `(< {tag})`? Annoyingly, `tag` is of the form `<number>`, so the double `<` will look funny either way...

saethlin: 🤦 yeah uh for starters I put the `..` on the wrong side of the unknown bottom. What if instead of `[ tag tag tag ]` we did `[ ?..unknown_bottom tag tag tag ]`

I'm not sure I like `<` because that to me says comparison and I feel like ranges are conceptually closer to what's going on here? But also we could be very verbose because this should be rare and print `[ unknown-bottom tag tag tag ]`.

src/shims/foreign_items.rs
saethlin: I think the previous implementation was wrong because it had a debugging tool reporting UB. I suppose we could `err_machine_stop` instead?

oli-obk: Should we test this failure path? Seems not strictly necessary for a debugging tool like this, but since we're already taking care thatthe message is helpful...

src/stacked_borrows/mod.rs
RalfJung: `?..unknown_bottom` to me looks like 'some sequence of elements, the last one being `unknown_bottom`'. So I think that is too confusing.

> [ unknown-bottom tag tag tag ].

Yeah verbose might be good, but it should still print the unk-bottom tag. So maybe `[ unknown-bottom(..$tag) $tag $tag ]`?

src/shims/foreign_items.rs
RalfJung: I think the previous implementation was a lot better since it used the shared codepath for determining the AllocId. This is messing with internals of the pointer representation that I don't think we want to unfold here. Calling `into_parts` is a red flag IMO.

saethlin: Ah, I thought we were more opposed to messing with InterpError

RalfJung: We are fairly strictly opposed to catching InterpError. But this is just replacing one error by another, and it seems strictly better than basically re-implementing `ptr_get_alloc_id`.

README.md
```diff
@@ -538,15 +538,20 @@ extern "Rust" {
     fn miri_start_panic(payload: *mut u8) -> !;
 
     /// Miri-provided extern function to get the internal unique identifier for the allocation that a pointer
-    /// points to. This is only useful as an input to `miri_print_stacks`, and it is a separate call because
+    /// points to. This is only useful as an input to `miri_print_borrow_stacks`, and it is a separate call because
     /// getting a pointer to an allocation at runtime can change the borrow stacks in the allocation.
```
RalfJung, line 542:
```suggestion
    /// points to. If this pointer is invalid (not pointing to an allocation), interpretation will abort.
    ///
    /// This is only useful as an input to `miri_print_borrow_stacks`, and it is a separate call because
    /// getting a pointer to an allocation at runtime can change the borrow stacks in the allocation.
```

README.md
```diff
@@ -538,15 +538,20 @@ extern "Rust" {
...
     /// Miri-provided extern function to print (from the interpreter, not the program) the contents of all
-    /// borrow stacks in an allocation. The format of what this emits is unstable and may change at any time.
-    /// In particular, users should be aware that Miri will periodically attempt to garbage collect the
-    /// contents of all stacks. Callers of this function may wish to pass `-Zmiri-tag-gc=0` to disable the GC.
-    fn miri_print_stacks(alloc_id: u64);
+    /// borrow stacks in an allocation. The leftmost tag is the bottom of the stack.
+    /// The format of what this emits is unstable and may change at any time. In particular, users should be
+    /// aware that Miri will periodically attempt to garbage collect the contents of all stacks. Callers of
+    /// this function may wish to pass `-Zmiri-tag-gc=0` to disable the GC.
+    /// This function is extremely unstable. At any time the format of its output may change, its signature may
```
RalfJung, line 552:
```suggestion
    /// this function may wish to pass `-Zmiri-tag-gc=0` to disable the GC.
    ///
    /// This function is extremely unstable. At any time the format of its output may change, its signature may
```

saethlin:
Thanks for improving the docs :)

PR #2619, opened 2022-10-26
ignore unused things in the stdlib

RalfJung:
The `cfg(miri)` make some things unused, and it's not worth allowing this all on the library side -- and https://github.com/rust-lang/rust/pull/102950 shows that these warnings currently can become a problem.


RalfJung:
Actually let's do https://github.com/rust-lang/rust/pull/103564 instead.

PR #2624, opened 2022-10-27
Implement `ptr_mask` intrinsic

WaffleLapkin:
I promised I'll implement it, but then forgot 😅

### DIFF
--- a/src/shims/intrinsics/mod.rs
+++ b/src/shims/intrinsics/mod.rs
@@ -11,7 +11,7 @@ use rustc_middle::{
     mir,
     ty::{self, FloatTy, Ty},
 };
-use rustc_target::abi::Integer;
+use rustc_target::abi::{Integer, Size};
 
 use crate::*;
 use atomic::EvalContextExt as _;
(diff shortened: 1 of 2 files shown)


src/shims/intrinsics/mod.rs
```diff
@@ -120,6 +120,18 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 this.write_bytes_ptr(ptr, iter::repeat(val_byte).take(byte_count.bytes_usize()))?;
             }
 
+            "ptr_mask" => {
+                let [ptr, mask] = check_arg_count(args)?;
+
+                let ptr = this.read_pointer(ptr)?;
+                #[allow(clippy::cast_sign_loss)]
+                let mask = this.read_scalar(mask)?.to_machine_isize(this)? as u64;
```
RalfJung, line 128: This is equivalent to `to_machine_usize` I think? And `usize` is the type used in the intrinsic so I don't understand why the `isize` method should be called here.

tests/pass/shims/ptr_mask.rs
```diff
@@ -0,0 +1,18 @@
...
+    let v: u32 = 0xABCDABCD;
+    let ptr: *const u32 = &v;
+
+    // u32 is 4 aligned,
+    // so the lower `log2(4) = 2` bits of the address are always 0
+    assert_eq!(ptr.addr() & 0b11, 0);
+
+    let tagged_ptr = ptr.map_addr(|a| a | 0b11);
+    let tag = tagged_ptr.addr() & 0b11;
+    let masked_ptr = ptr.mask(!0b11);
```
RalfJung, line 14: Shouldn't this use `tagged_ptr`?

WaffleLapkin: yep, oops

src/shims/intrinsics/mod.rs
WaffleLapkin: For some reason I couldn't find `to_machine_usize` when I was writing this, must be an overlook. Anyway, fixed now.

RalfJung:
I was wondering why we hadn't noticed this earlier... turns out there is no doctest and no libcore unit test for this function, so it is not hit by miri-test-libstd. Probably just some ui tests? AFAIK we're trying to avoid using ui tests for things where unit tests could be used.

WaffleLapkin:
fyi: I've filled a PR to add doc tests to the `pointer::mask` functions: https://github.com/rust-lang/rust/pull/103694

PR #2627, opened 2022-10-28
update ignore-windows comments

RalfJung:
Turns out 2 of these tests can actually be enabled. :)


oli-obk:
I could set up ui test to tell us when ignore-os tests start passing on that os.

For other things this may be too noisy, but for OSs this should be reliable

RalfJung:
@oli-obk sure, that sounds useful, but not very high priority.

Issue #2528, opened 2022-09-01
Tracking issue for enabling field retagging by default


RalfJung:
What has to happen for us to enable `-Zmiri-retag-fields` by default? LLVM [already does optimizations](https://github.com/rust-lang/rust/issues/16515#issuecomment-1234769355) that can only be explained on the Rust level if we do field retagging. This is the last case I am aware of of Rust code that has LLVM UB but is not considered UB by Miri. (There are other differences between Miri and the Reference, but none of those are LLVM UB -- all that extra UB is wiggle room for future optimizations, but not exploited yet.)

@saethlin has been doing some experiments here I think, but I am no aware of the current status. As a start, we should probably get [miri-test-libstd](https://github.com/rust-lang/miri-test-libstd) to pass with field retagging.

Related issues:
- [x] https://github.com/rust-lang/rust/issues/60076
- [x] https://github.com/rust-lang/rust/issues/101983
- [x] Release of hashbrown that includes https://github.com/rust-lang/hashbrown/pull/359
- [x] Release of scopeguard that includes https://github.com/bluss/scopeguard/pull/25

saethlin:
> has been doing some experiments here I think

I updated https://miri.saethlin.dev/ub to use field retagging a week or two ago. Generally, it looks like the ecosystem impact is minimal.

RalfJung:
So for now the primary blocker is https://github.com/rust-lang/rust/issues/60076.

RalfJung:
miri-test-libstd now runs the libcore tests with `-Zmiri-retag-fields`.

liballoc fails due to https://github.com/rust-lang/rust/issues/60076:
```
test collections::vec_deque::tests::test_drain ... error: Undefined Behavior: not granting access to tag <42280440> because that would remove [SharedReadOnly for <42659664>] which is protected because it is an argument of call 10005308
    --> alloc_miri_test/../library/alloc/src/collections/vec_deque/mod.rs:273:13
     |
273  |             ptr::copy(self.ptr().add(src), self.ptr().add(dst), len);
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not granting access to tag <42280440> because that would remove [SharedReadOnly for <42659664>] which is protected because it is an argument of call 10005308
     |
     = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
[...]

PR #2636, opened 2022-10-29
Stacked Borrows: make scalar field retagging the default

RalfJung:
I think it is time to finally close this soundness gap. Any objections? :)

Unfortunately the latest released versions of hashbrown and scopeguard can fail under full field retagging. The fixes have landed in the git repos but have not been released yet. I don't know if scalar field retagging as enabled by this PR is sufficient to cause problems with these crates, but it seems likely that this would be the case -- e.g. if both `value` and `dropfn` are scalars, the entire scopeguard struct will be a `ScalarPair` and thus get field retagging.

However, given that we actually generate LLVM `noalias` for these cases, it seems prudent to inform users of this risk. They can easily set `-Zmiri-field-retag=none` to opt-out of this change.

Cc https://github.com/rust-lang/miri/issues/2528

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,14 +374,15 @@ to Miri failing to detect cases of undefined behavior in a program.
   application instead of raising an error within the context of Miri (and halting
   execution). Note that code might not expect these operations to ever panic, so
   this flag can lead to strange (mis)behavior.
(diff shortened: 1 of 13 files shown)


saethlin (approved):
Yay!

PR #2231, opened 2022-06-13
Windows thread support: Part 1

DrMeepster:
This PR adds support for threads on Windows.

PR #2601, opened 2022-10-18
Add Windows InitOnceBeginInitialize and InitOnceComplete shims

DrMeepster:
Fixes #2595

PR #2638, opened 2022-10-30
Implement condvars for Windows

### LINKED ISSUE #2628: Support synchronization primitives on Windows
Thanks to @DrMeepster we now do support basic concurrency and atomics on Windows (https://github.com/rust-lang/miri/pull/2231). However, we do not support all of the standard library concurrency primitives. Based on my testing:
- [x] Mutex
- [x] RwLock
- [x] Condvar
- [x] thread parking

The fact that we don't support thread parking (which is needed by `std::sync::mpsc`) unfortunately blocks https://github.com/rust-lang/rust/pull/103681.

DrMeepster:
Adds 3 shims for Windows: `SleepConditionVariableSRW`, `WakeConditionVariable`, `WakeAllConditionVariable` to add support for condvars (which fixes #2628).

Salvaged from what was removed from #2231

### DIFF
--- a/src/concurrency/sync.rs
+++ b/src/concurrency/sync.rs
@@ -121,8 +121,10 @@ declare_id!(CondvarId);
 struct CondvarWaiter {
     /// The thread that is waiting on this variable.
     thread: ThreadId,
-    /// The mutex on which the thread is waiting.
-    mutex: MutexId,
+    /// The mutex or rwlock on which the thread is waiting.
+    lock: u32,
+    /// If the lock is shared or exclusive
+    shared: bool,
 }
 
 /// The conditional variable state.
@@ -569,16 +571,16 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
     }
 
     /// Mark that the thread is waiting on the conditional variable.
-    fn condvar_wait(&mut self, id: CondvarId, thread: ThreadId, mutex: MutexId) {
+    fn condvar_wait(&mut self, id: CondvarId, thread: ThreadId, lock: u32, shared: bool) {
         let this = self.eval_context_mut();
         let waiters = &mut this.machine.threads.sync.condvars[id].waiters;
         assert!(waiters.iter().all(|waiter| waiter.thread != thread), "thread is already waiting");
-        waiters.push_back(CondvarWaiter { thread, mutex });
+        waiters.push_back(CondvarWaiter { thread, lock, shared });
     }
 
     /// Wake up some thread (if there is any) sleeping on the conditional
     /// variable.
-    fn condvar_signal(&mut self, id: CondvarId) -> Option<(ThreadId, MutexId)> {
+    fn condvar_signal(&mut self, id: CondvarId) -> Option<(ThreadId, u32, bool)> {
         let this = self.eval_context_mut();
         let current_thread = this.get_active_thread();
         let condvar = &mut this.machine.threads.sync.condvars[id];
@@ -592,7 +594,7 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
             if let Some(data_race) = data_race {
                 data_race.validate_lock_acquire(&condvar.data_race, waiter.thread);
             }
-            (waiter.thread, waiter.mutex)
+            (waiter.thread, waiter.lock, waiter.shared)
         })
     }
 
--- a/src/concurrency/sync.rs
+++ b/src/concurrency/sync.rs
@@ -116,15 +116,25 @@ struct RwLock {
 
 declare_id!(CondvarId);
 
+#[derive(Debug, Copy, Clone)]
+pub enum RwLockMode {
+    Read,
+    Write,
+}
+
+#[derive(Debug)]
+pub enum CondvarLock {
+    Mutex(MutexId),
+    RwLock { id: RwLockId, mode: RwLockMode },
+}
+
 /// A thread waiting on a conditional variable.
 #[derive(Debug)]
 struct CondvarWaiter {
     /// The thread that is waiting on this variable.
     thread: ThreadId,
     /// The mutex or rwlock on which the thread is waiting.
-    lock: u32,
-    /// If the lock is shared or exclusive
-    shared: bool,
+    lock: CondvarLock,
 }
 
 /// The conditional variable state.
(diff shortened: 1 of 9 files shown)


DrMeepster:
I think this panic is caused by a mistake in #2601.

src/concurrency/sync.rs
```diff
@@ -121,8 +121,10 @@ declare_id!(CondvarId);
 struct CondvarWaiter {
     /// The thread that is waiting on this variable.
     thread: ThreadId,
-    /// The mutex on which the thread is waiting.
-    mutex: MutexId,
+    /// The mutex or rwlock on which the thread is waiting.
+    lock: u32,
```
oli-obk, line 125: Not sure if that's overkill, but it may be a bit cleaner to use a

```rust
enum Lock {
    Mutexes(MutexId),
    RwLock{ id: RwLockId, shared: bool }
}
```

src/shims/windows/sync.rs
```diff
@@ -8,6 +8,38 @@ use crate::*;
...
+        this.unblock_thread(thread);
+
+        if shared {
+            if this.rwlock_is_locked(lock) {
+                this.rwlock_enqueue_and_block_reader(lock, thread);
+            } else {
+                this.rwlock_reader_lock(lock, thread);
+            }
+        } else {
+            if this.rwlock_is_write_locked(lock) {
```
oli-obk, line 33: So if it's read locked when we write lock we don't enqueue but immediately do a write lock?

DrMeepster: Whoops, `rwlock_is_locked` and `rwlock_is_write_locked` are swapped here. It doesn't fail any tests because std only ever uses condvars in exclusive mode. I'll add tests for shared mode.

DrMeepster:
CI failure should be fixed by #2641.

src/concurrency/sync.rs
RalfJung: I'd actually like to go further and replace `shared: bool` by an `enum RwLockMode { Read, Write }` or something like that.

src/shims/windows/sync.rs
```diff
@@ -325,4 +358,129 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+
+        let condvar_id = this.condvar_get_or_create_id(condvar_op, CONDVAR_ID_OFFSET)?;
+        let lock_id = this.rwlock_get_or_create_id(lock_op, SRWLOCK_ID_OFFSET)?;
+        let timeout_ms = this.read_scalar(timeout_op)?.to_u32()?;
+        let flags = this.read_scalar(flags_op)?.to_u32()?;
+
+        let timeout_time = if timeout_ms == this.eval_windows("c", "INFINITE")?.to_u32()? {
+            None
+        } else {
+            this.check_no_isolation("`SleepConditionVariableSRW` with non-infinite timeout")?;
```
RalfJung, line 380: Since this is a monotone clock timeout, it can actually be allowed even with isolation. We have a fake monotone clock for that purpose now. :)

src/shims/windows/sync.rs
```diff
@@ -325,4 +358,129 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+
+            let duration = Duration::from_millis(timeout_ms.into());
+
+            Some(this.machine.clock.now().checked_add(duration).unwrap())
+        };
+
+        let shared_mode = 0x1; // CONDITION_VARIABLE_LOCKMODE_SHARED is not in std
+        let shared = flags == shared_mode;
```
RalfJung, line 388: What about other values of `flags` (neither 0 nor `shared_mode`)?

src/shims/windows/sync.rs
```diff
@@ -325,4 +358,129 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+
+        this.block_thread(active_thread);
+        this.condvar_wait(condvar_id, active_thread, CondvarLock::RwLock { id: lock_id, shared });
+
+        if let Some(timeout_time) = timeout_time {
+            struct Callback<'tcx> {
+                thread: ThreadId,
+                condvar_id: CondvarId,
+                lock_id: RwLockId,
+                shared: bool,
```
RalfJung, line 412: All these `shared: bool` should also use the new `RwLockMode` enum I suggested.

src/shims/windows/sync.rs
```diff
@@ -325,4 +358,129 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+
+            impl<'mir, 'tcx: 'mir> MachineCallback<'mir, 'tcx> for Callback<'tcx> {
+                fn call(&self, this: &mut MiriInterpCx<'mir, 'tcx>) -> InterpResult<'tcx> {
+                    this.reacquire_cond_lock(self.thread, self.lock_id, self.shared)?;
+
+                    this.condvar_remove_waiter(self.condvar_id, self.thread);
+
+                    let error_timeout = this.eval_windows("c", "ERROR_TIMEOUT")?;
+                    this.set_last_error(error_timeout)?;
+                    this.write_scalar(Scalar::from_i32(0), &self.dest)?;
```
RalfJung, line 431: I guess this could also use `eval_windows("c", "FALSE")`?

tests/pass/concurrency/windows_condvar_shared.rs
```diff
@@ -0,0 +1,223 @@
...
+                    condvar_ptr.0,
+                    lock_ptr.0,
+                    INFINITE,
+                    CONDITION_VARIABLE_LOCKMODE_SHARED,
+                )
+            };
+            assert_ne!(r, 0);
+
+            println!("reacquired lock");
+        }));
```
RalfJung, line 128: So these never release their lock?

tests/pass/concurrency/windows_condvar_shared.rs
```diff
@@ -0,0 +1,223 @@
...
+    let condvar_ptr = SendPtr(&mut condvar);
+
+    let mut handles = Vec::with_capacity(10);
+    for _ in 0..5 {
+        handles.push(thread::spawn(move || {
+            unsafe {
+                AcquireSRWLockExclusive(lock_ptr.0);
+            }
+
+            println!("waiter locked");
```
RalfJung, line 173: So "waiter" is sometimes shared and sometimes exclusive? That makes the output a bit confusing.

Also might be good to print the ID of the waiter here (the idx in the for loop)?

RalfJung (commented):
Generally looks good, thanks! Mostly stylistic comments.

src/concurrency/sync.rs
```diff
@@ -116,13 +116,25 @@ struct RwLock {
 
 declare_id!(CondvarId);
 
+#[derive(Debug, Copy, Clone)]
+pub enum RwLockMode {
+    Shared,
+    Exclusive,
```
RalfJung, line 122: Let's use Rut terminology here, not Windows terminology.
It's called "RwLock". The two modes should be "read" and "write". :)

tests/pass/concurrency/windows_condvar_shared.rs
DrMeepster: No. Is there a reason to release them? The lock is never used again

RalfJung: It's just surprising. :shrug: so please add a comment that it is intentional.

RalfJung:
> I think this panic is caused by a mistake in #2601.

That PR landed, so please rebase this one. :)

DrMeepster:
The rebase is done. CI's finally passing :)

RalfJung:
Great. :) Please squash the commits a little, then we can land this.

DrMeepster:
Squashed.

PR #2640, opened 2022-10-31
Fix dupe word typos

Rageking8:


saethlin:
@bors r+
Still amazes me how many of these you keep finding.

DrMeepster:
```rust
thread 'main' panicked at 'assertion failed: diff.as_millis() < 500', tests/pass/shims/time.rs:8:5
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /Users/runner/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/panicking.rs:575:5
   1: std::rt::panic_fmt
             at /Users/runner/.rustup/toolchains/miri/lib/rustlib/src/rust/library/core/src/panicking.rs:65:14
   2: core::panicking::panic
             at /Users/runner/.rustup/toolchains/miri/lib/rustlib/src/rust/library/core/src/panicking.rs:115:5
   3: duration_sanity
             at tests/pass/shims/time.rs:8:5
   4: main
             at tests/pass/shims/time.rs:35:5
   5: <fn() as std::ops::FnOnce<()>>::call_once - shim(fn())
             at /Users/runner/.rustup/toolchains/miri/lib/rustlib/src/rust/library/core/src/ops/function.rs:251:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```
Error is unrelated to this PR, it looks like the runner lagged. Do we even need this part of the time test?

saethlin:
> Do we even need this part of the time test?

We should probably have some kind of test like this, but I don't know how to write them so that they are strongly resilient to this situation. There are other tests in here for things like timeouts which flake if the interpreter is made too slow. Help welcomed.

cargo-miri/src/phases.rs
```diff
@@ -528,7 +528,7 @@ pub fn phase_runner(mut binary_args: impl Iterator<Item = String>, phase: Runner
     cmd.args(binary_args);
 
     // Make sure we use the build-time working directory for interpreting Miri/rustc arguments.
-    // But then we need to switch to the run-time one, which we instruct Miri do do by setting `MIRI_CWD`.
+    // But then we need to switch to the run-time one, which we instruct Miri do by setting `MIRI_CWD`.
```
RalfJung, line 531: This should have been "to do". the PR replaced one type by another. :)

Could you file a follow-up to fix that?

Rageking8: Haha, sure.

PR #2641, opened 2022-10-31
InitOnce: synchronize with completion when already complete

DrMeepster:
The completion of an InitOnce happens-before the threads waiting on it wake up. However, this is not the case for threads that call `InitOnceBeginInitialize` after the completion, leading to data races and outdated weak memory loads as observed in the CI for  #2638. This PR fixes this.

### DIFF
--- a/src/concurrency/init_once.rs
+++ b/src/concurrency/init_once.rs
@@ -141,18 +141,11 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         // Wake up everyone.
         // need to take the queue to avoid having `this` be borrowed multiple times
         for waiter in std::mem::take(&mut init_once.waiters) {
-            // End of the wait happens-before woken-up thread.
-            if let Some(data_race) = &this.machine.data_race {
-                data_race.validate_lock_acquire(
-                    &this.machine.threads.sync.init_onces[id].data_race,
-                    waiter.thread,
-                );
-            }
-
             this.unblock_thread(waiter.thread);
 
             // Call callback, with the woken-up thread as `current`.
             this.set_active_thread(waiter.thread);
+            this.init_once_acquire(id);
             waiter.callback.call(this)?;
             this.set_active_thread(current_thread);
         }
@@ -172,26 +165,17 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         );
 
         // Each complete happens-before the end of the wait
-        // FIXME: should this really induce synchronization? If we think of it as a lock, then yes,
-        // but the docs don't talk about such details.
         if let Some(data_race) = &this.machine.data_race {
             data_race.validate_lock_release(&mut init_once.data_race, current_thread);
         }
 
         // Wake up one waiting thread, so they can go ahead and try to init this.
         if let Some(waiter) = init_once.waiters.pop_front() {
-            // End of the wait happens-before woken-up thread.
-            if let Some(data_race) = &this.machine.data_race {
-                data_race.validate_lock_acquire(
-                    &this.machine.threads.sync.init_onces[id].data_race,
-                    waiter.thread,
-                );
-            }
-
             this.unblock_thread(waiter.thread);
 
             // Call callback, with the woken-up thread as `current`.
             this.set_active_thread(waiter.thread);
+            this.init_once_acquire(id);
             waiter.callback.call(this)?;
             this.set_active_thread(current_thread);
         } else {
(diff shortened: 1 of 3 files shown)


src/concurrency/init_once.rs
```diff
@@ -201,4 +185,19 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
 
         Ok(())
     }
+
+    /// Synchronize with the previous completion or failure of an InitOnce.
+    /// This is required to prevent data races.
+    #[inline]
+    fn init_once_acquire(&mut self, id: InitOnceId) {
```
RalfJung, line 192: Please keep the name consistent with the locks -- `validate_init_once_acquie` maybe? Also this should probably be on a private extension trait.

src/concurrency/init_once.rs
```diff
@@ -172,26 +165,17 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
-                );
-            }
-
             this.unblock_thread(waiter.thread);
 
             // Call callback, with the woken-up thread as `current`.
             this.set_active_thread(waiter.thread);
+            this.init_once_acquire(id);
             waiter.callback.call(this)?;
             this.set_active_thread(current_thread);
```
RalfJung, line 180: The entire inside of the `if let` could be shared with init_once_complete, couldn't it?

src/shims/windows/sync.rs
```diff
@@ -177,8 +177,10 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                     Box::new(Callback { init_once_id: id, pending_place }),
                 )
             }
-            InitOnceStatus::Complete =>
-                this.write_scalar(this.eval_windows("c", "FALSE")?, &pending_place)?,
+            InitOnceStatus::Complete => {
+                this.init_once_acquire(id);
```
RalfJung, line 181: This is the only actual change in behavior, right?

Something seems strange here. In a regular lock, we don't have a `validate_lock_acquire` before *and* after the wakeup, do we?

tests/pass/concurrency/windows_init_once.rs
```diff
@@ -131,8 +131,46 @@ fn retry_on_fail() {
...
+        place_ptr.0.read()
+    });
+
+    unsafe {
+        // this should not data race
+        place_ptr.0.write(1);
+    }
+
+    unsafe {
+        assert_eq!(InitOnceComplete(init_once_ptr.0, 0, null_mut()), TRUE);
```
RalfJung, line 163: This is where the reader thread gets unblocked, right? Please add a comment.

tests/pass/concurrency/windows_init_once.rs
```diff
@@ -131,8 +131,46 @@ fn retry_on_fail() {
...
+    }
+    //println!("complete");
+
+    // run reader
```
RalfJung, line 167:
```suggestion
    // wait for reader to finish
```

src/shims/windows/sync.rs
DrMeepster: This isn't after the wakeup, this is more like acquiring a lock that is unlocked

RalfJung: Oh, this is in `InitOnceBeginInitialize`.

Still something is odd here, having the data race stuff be partially handled in this file and partially in `init_once.rs` seems to me like there's improper layering going on.

RalfJung: Okay so the issue is that the code in `init_once.rs` doesn't even get told when `InitOnceBeginInitialize` encounters an alredy init'ed init_once. So we need a new public function, and that is what `init_once_acquire` is meant to be. Maybe call it `init_once_observe_completed` or so? 'acquire' to me sounded like this is relevant during the phase before the init_once is initialized, but actually this is only called after full init and could even assert the state, right?

src/concurrency/init_once.rs
```diff
@@ -172,26 +165,17 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
-                    &this.machine.threads.sync.init_onces[id].data_race,
-                    waiter.thread,
-                );
-            }
-
             this.unblock_thread(waiter.thread);
 
             // Call callback, with the woken-up thread as `current`.
             this.set_active_thread(waiter.thread);
+            this.init_once_acquire(id);
```
RalfJung, line 178: This is a strange one, since we didn't actually fully init the init_once yet. It's not really documented to provide any synchronization in that case, is it?

src/shims/windows/sync.rs
RalfJung: Ah no this is also called from init_once_fail... but that's an internal thing. Code outside the init_once module should only call this when the init_once is fully initialized. So the public method provided by init_once should assert that.

src/concurrency/init_once.rs
DrMeepster: Well, synchronization is not documented in any case. If there wasn't synchronization here, you could get a data race between a failed initialization attempt and the next initialization attempt, which doesn't make any sense.

RalfJung: If InitOnce used relaxed accesses plus fences only for the success case, this could totally happen. So it makes sense to me.

But I agree that is unlikely to be the implementation.

> Well, synchronization is not documented in any case.

Yeah and that's a problem, concurrent functions should *always* document their synchronization -- if they don't document any synchronization they should be assumed not to synchronize. This is an important part of the observable behavior of a function.

src/concurrency/init_once.rs
RalfJung: Okay never mind this part, I only later realized how this function is being used.

tests/pass/concurrency/windows_init_once.rs
DrMeepster: The reader thread is never blocked, so I added a comment clarifying that instead.

RalfJung:
Thanks, looks good!
@bors r+

Issue #2205, opened 2022-06-06
Print out both accesses involved in a data race


cbeuw:
Currently Miri throws a UB on the access that ultimately causes a data race, which is always the later one. It is not clear where the other access involved in the data race is in the source file.
```
error: Undefined Behavior: Data race detected between Write on Thread(id = 0, name = "main") and Write on Thread(id = 1) at alloc1 (current vector clock = VClock([5]), conflicting timestamp = VClock([0, 8]))
  --> src/lib.rs:21:18
   |
21 |         unsafe { V = 2 }
   |                  ^^^^^ Data race detected between Write on Thread(id = 0, name = "main") and Write on Thread(id = 1) at alloc1 (current vector clock = VClock([5]), conflicting timestamp = VClock([0, 8]))
   |
   = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
   = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
           
   = note: inside `tests::data_race` at src/lib.rs:21:18
```
Both [TSan](https://github.com/google/sanitizers/wiki/ThreadSanitizerCppManual#usage) and [Go's race detector](https://go.dev/doc/articles/race_detector#Report_Format) print out the line numbers of both accesses involved in the data race, we'd like to do the same.

It shouldn't be too hard to implement, we just need to augment `VTimestamp` with the current `Span` from `MiriEvalContext::cur_span()`
https://github.com/rust-lang/miri/blob/aca3b3a645e6d0295e9f1829c37af811a6de5251/src/vector_clock.rs#L45

RalfJung:
Stacked Borrows recently got some support for showing extra spans thanks to @saethlin; we probably want to reuse that -- in particular it doesn't use `cur_span` but tries to find a span "in the local crate".

Also, augmenting *every* timestamp might be a bit too expensive in terms of performance?

saethlin:
I don't know that augmenting every timestamp would be too slow. TSan manages to report backtraces in a debug build for these races, and those backtraces are not small and we're only trying to report one measly span. If someone is interested in doing this, it's probably worth giving the naïve approach a shot. If it's too slow I already have some ideas.

cbeuw:
Is there a way to get a backtrace in Miri, and how expensive would this be? Only showing one span is probably not too helpful

saethlin:
[...]

Issue #2200, opened 2022-06-06
Reduce redundancy in our error messages


RalfJung:
Our errors currently print the same message twice:
```
error: Undefined Behavior: pointer arithmetic failed: alloc1655 has size 4, so pointer to 1 byte starting at offset -1 is out-of-bounds
   --> /home/r/.rustup/toolchains/miri/lib/rustlib/src/rust/library/core/src/ptr/const_ptr.rs:455:18
    |
455 |         unsafe { intrinsics::offset(self, count) }
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ pointer arithmetic failed: alloc1655 has size 4, so pointer to 1 byte starting at offset -1 is out-of-bounds
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
```
That is unnecessarily redundant redundant. So maybe we could avoid duplication, e.g. by changing the above output to
```
error: Undefined Behavior: pointer arithmetic failed
   --> /home/r/.rustup/toolchains/miri/lib/rustlib/src/rust/library/core/src/ptr/const_ptr.rs:455:18
    |
455 |         unsafe { intrinsics::offset(self, count) }
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ alloc1655 has size 4, so pointer to 1 byte starting at offset -1 is out-of-bounds
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
```

This will require
- changing rustc: the `InterpError` type needs separate methods to compute the "title" and "text" of a message
- adjusting our test suite: many `ERROR` annotations will probably fail if the error details no longer appear in the "error:" message. On a case-by-case basis, we should either change this to just match on what the error message still says (we do have ui tests after all, so if the details change in unexpected ways we will notice), or add a second pattern to also match some key part of the details.

bjorn3:
What about the case where no source is available for the error origin?

RalfJung:
You are concerned about something like https://github.com/rust-lang/rust/issues/97699? IMO it is a bug in the rustc diagnostic printing that it entirely omits information attached to spans where the source is missing. We might have to work around that somehow on our side.

[...]

Issue #1689, opened 2021-01-25
Data race detector causes massive slowdown


shepmaster:
Steps to reproduce:

Check out https://github.com/shepmaster/sxd-string-slab/commit/8bf847eff2a8a1609c818cae68d227156f5f3dc1, then run Miri:

```zsh
time MIRIFLAGS='-Zmiri-disable-isolation -Zmiri-disable-stacked-borrows -Zmiri-disable-data-race-detector' \
  CARGO_TARGET_DIR=target/miri \
  cargo miri test -- --nocapture
# Reports ~23 seconds on my machine

time MIRIFLAGS='-Zmiri-disable-isolation -Zmiri-disable-data-race-detector' \
  CARGO_TARGET_DIR=target/miri \
  cargo miri test -- --nocapture
# Reports ~16.5 minutes on my machine

time MIRIFLAGS='-Zmiri-disable-isolation -Zmiri-disable-stacked-borrows' \
  CARGO_TARGET_DIR=target/miri \
  cargo miri test -- --nocapture
# Reports ~4.25 minutes on my machine
```

# Meta

- macOS

- `rustc --version --verbose`

    ```
    rustc 1.51.0-nightly (1d0d76f8d 2021-01-24)
    binary: rustc
    commit-hash: 1d0d76f8dd4f5f6ecbeab575b87edaf1c9f56bb8
    commit-date: 2021-01-24
    host: x86_64-apple-darwin
    release: 1.51.0-nightly
    LLVM version: 11.0.1
    ```

- `cargo miri --version --verbose`

    ```
    miri 0.1.0 (1cf1a2e4 2021-01-15)
    ```

RalfJung:
Cc @JCTyblaidd any idea what could be happening here? This is a 10x slowdown due to the data race detector.

OTOH, it's also a 40x slowdown due to Stacked Borrows, which is much more than what I observed elsewhere.

JCTyblaidd:
I will look into this properly later:

My guess would be that criterion is spawning a large set of threads for which the data-race detector is enabled for, potentially a large set with delayed synchronization - so from the perspective of the data-race detector there is a very large set of live concurrent threads to detect races for (probably crossing the smallvec threshold). I need to try track thread usage for a proper analysis.

shepmaster:
> that criterion is spawning a large set of threads

[...]

PR #2646, opened 2022-11-04
Data race spans

saethlin:
Fixes https://github.com/rust-lang/miri/issues/2205

This adds output to data race errors very similar to the spans we emit for Stacked Borrows errors. For example, from our test suite:
```
help: The Atomic Load on thread `<unnamed>` is here
  --> tests/fail/data_race/atomic_read_na_write_race1.rs:23:13
   |
23 | ...   (&*c.0).load(Ordering::SeqCst) //~ ERROR: Data race detected between Atomic Load on thread `<unnamed>` and Write o...
   |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: The Write on thread `<unnamed>` is here
  --> tests/fail/data_race/atomic_read_na_write_race1.rs:19:13
   |
19 |             *(c.0 as *mut usize) = 32;
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^```
```

Because of https://github.com/rust-lang/miri/pull/2647 this comes without a perf regression, according to our benchmarks.

### DIFF
--- a/src/concurrency/data_race.rs
+++ b/src/concurrency/data_race.rs
@@ -50,8 +50,11 @@ use rustc_ast::Mutability;
 use rustc_data_structures::fx::{FxHashMap, FxHashSet};
 use rustc_index::vec::{Idx, IndexVec};
 use rustc_middle::mir;
+use rustc_span::Span;
+use rustc_span::DUMMY_SP;
 use rustc_target::abi::{Align, Size};
 
+use crate::diagnostics::RacingOp;
 use crate::*;
 
 use super::{
@@ -144,8 +147,8 @@ impl ThreadClockSet {
     /// Increment the happens-before clock at a
     /// known index.
     #[inline]
-    fn increment_clock(&mut self, index: VectorIdx) {
-        self.clock.increment_index(index);
+    fn increment_clock(&mut self, index: VectorIdx, current_span: Span) {
+        self.clock.increment_index(index, current_span);
     }
 
     /// Join the happens-before clock with that of
@@ -361,6 +364,8 @@ impl MemoryCellClocks {
             atomic.read_vector.set_at_index(&clocks.clock, index);
             Ok(())
         } else {
+            let atomic = self.atomic_mut();
+            atomic.read_vector.set_at_index(&clocks.clock, index);
             Err(DataRace)
         }
     }
@@ -378,6 +383,8 @@ impl MemoryCellClocks {
             atomic.write_vector.set_at_index(&clocks.clock, index);
             Ok(())
         } else {
+            let atomic = self.atomic_mut();
+            atomic.write_vector.set_at_index(&clocks.clock, index);
             Err(DataRace)
         }
     }
@@ -386,46 +393,51 @@ impl MemoryCellClocks {
     /// returns true if a data-race is detected.
     fn read_race_detect(
         &mut self,
-        clocks: &ThreadClockSet,
+        clocks: &mut ThreadClockSet,
         index: VectorIdx,
+        current_span: Span,
     ) -> Result<(), DataRace> {
         log::trace!("Unsynchronized read with vectors: {:#?} :: {:#?}", self, clocks);
-        if self.write <= clocks.clock[self.write_index] {
+        let res = if self.write <= clocks.clock[self.write_index] {
             let race_free = if let Some(atomic) = self.atomic() {
                 atomic.write_vector <= clocks.clock
             } else {
                 true
             };
-            if race_free {
-                self.read.set_at_index(&clocks.clock, index);
-                Ok(())
-            } else {
-                Err(DataRace)
-            }
+            self.read.set_at_index(&clocks.clock, index);
+            if race_free { Ok(()) } else { Err(DataRace) }
         } else {
             Err(DataRace)
+        };
+        if res.is_ok() && current_span != DUMMY_SP {
+            clocks.clock[index].span = current_span;
         }
+        res
     }
 
     /// Detect races for non-atomic write operations at the current memory cell
     /// returns true if a data-race is detected.
     fn write_race_detect(
         &mut self,
-        clocks: &ThreadClockSet,
+        clocks: &mut ThreadClockSet,
         index: VectorIdx,
         write_type: WriteType,
+        current_span: Span,
     ) -> Result<(), DataRace> {
         log::trace!("Unsynchronized write with vectors: {:#?} :: {:#?}", self, clocks);
+        if current_span != DUMMY_SP {
+            clocks.clock[index].span = current_span;
+        }
         if self.write <= clocks.clock[self.write_index] && self.read <= clocks.clock {
             let race_free = if let Some(atomic) = self.atomic() {
                 atomic.write_vector <= clocks.clock && atomic.read_vector <= clocks.clock
             } else {
                 true
             };
+            self.write = clocks.clock[index];
+            self.write_index = index;
+            self.write_type = write_type;
             if race_free {
-                self.write = clocks.clock[index];
-                self.write_index = index;
-                self.write_type = write_type;
                 self.read.set_zero_vector();
                 Ok(())
             } else {
@@ -621,30 +633,35 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: MiriInterpCxExt<'mir, 'tcx> {
     /// Update the data-race detector for an atomic fence on the current thread.
     fn atomic_fence(&mut self, atomic: AtomicFenceOrd) -> InterpResult<'tcx> {
         let this = self.eval_context_mut();
+        let current_span = this.machine.current_span();
         if let Some(data_race) = &mut this.machine.data_race {
-            data_race.maybe_perform_sync_operation(&this.machine.threads, |index, mut clocks| {
-                log::trace!("Atomic fence on {:?} with ordering {:?}", index, atomic);
+            data_race.maybe_perform_sync_operation(
+                &this.machine.threads,
+                |index, mut clocks| {
+                    log::trace!("Atomic fence on {:?} with ordering {:?}", index, atomic);
 
-                // Apply data-race detection for the current fences
-                // this treats AcqRel and SeqCst as the same as an acquire
-                // and release fence applied in the same timestamp.
-                if atomic != AtomicFenceOrd::Release {
-                    // Either Acquire | AcqRel | SeqCst
-                    clocks.apply_acquire_fence();
-                }
-                if atomic != AtomicFenceOrd::Acquire {
-                    // Either Release | AcqRel | SeqCst
-                    clocks.apply_release_fence();
-                }
-                if atomic == AtomicFenceOrd::SeqCst {
-                    data_race.last_sc_fence.borrow_mut().set_at_index(&clocks.clock, index);
-                    clocks.fence_seqcst.join(&data_race.last_sc_fence.borrow());
-                    clocks.write_seqcst.join(&data_race.last_sc_write.borrow());
-                }
+                    // Apply data-race detection for the current fences
+                    // this treats AcqRel and SeqCst as the same as an acquire
+                    // and release fence applied in the same timestamp.
+                    if atomic != AtomicFenceOrd::Release {
+                        // Either Acquire | AcqRel | SeqCst
+                        clocks.apply_acquire_fence();
+                    }
+                    if atomic != AtomicFenceOrd::Acquire {
+                        // Either Release | AcqRel | SeqCst
+                        clocks.apply_release_fence();
+                    }
+                    if atomic == AtomicFenceOrd::SeqCst {
+                        data_race.last_sc_fence.borrow_mut().set_at_index(&clocks.clock, index);
+                        clocks.fence_seqcst.join(&data_race.last_sc_fence.borrow());
+                        clocks.write_seqcst.join(&data_race.last_sc_write.borrow());
+                    }
 
-                // Increment timestamp in case of release semantics.
-                Ok(atomic != AtomicFenceOrd::Acquire)
-            })
+                    // Increment timestamp in case of release semantics.
+                    Ok(atomic != AtomicFenceOrd::Acquire)
+                },
+                current_span,
+            )
         } else {
             Ok(())
         }
@@ -682,6 +699,7 @@ impl VClockAlloc {
         thread_mgr: &ThreadManager<'_, '_>,
         len: Size,
         kind: MemoryKind<MiriMemoryKind>,
+        current_span: Span,
     ) -> VClockAlloc {
         let (alloc_timestamp, alloc_index) = match kind {
             // User allocated and stack memory should track allocation.
@@ -693,7 +711,10 @@ impl VClockAlloc {
             )
             | MemoryKind::Stack => {
                 let (alloc_index, clocks) = global.current_thread_state(thread_mgr);
-                let alloc_timestamp = clocks.clock[alloc_index];
+                let mut alloc_timestamp = clocks.clock[alloc_index];
+                if current_span != DUMMY_SP {
+                    alloc_timestamp.span = current_span;
+                }
                 (alloc_timestamp, alloc_index)
             }
             // Other global memory should trace races but be allocated at the 0 timestamp.
@@ -704,7 +725,7 @@ impl VClockAlloc {
                 | MiriMemoryKind::ExternStatic
                 | MiriMemoryKind::Tls,
             )
-            | MemoryKind::CallerLocation => (0, VectorIdx::MAX_INDEX),
+            | MemoryKind::CallerLocation => (VTimestamp::NONE, VectorIdx::MAX_INDEX),
         };
         VClockAlloc {
             alloc_ranges: RefCell::new(RangeMap::new(
@@ -735,7 +756,7 @@ impl VClockAlloc {
                     let idx = l_remainder_slice
                         .iter()
                         .enumerate()
-                        .find_map(|(idx, &r)| if r == 0 { None } else { Some(idx) })
+                        .find_map(|(idx, &r)| if r == VTimestamp::NONE { None } else { Some(idx) })
                         .expect("Invalid VClock Invariant");
                     Some(idx + r_slice.len())
                 } else {
@@ -762,7 +783,7 @@ impl VClockAlloc {
     ) -> InterpResult<'tcx> {
         let (current_index, current_clocks) = global.current_thread_state(thread_mgr);
         let write_clock;
-        let (other_action, other_thread, _other_clock) = if range.write
+        let (other_action, other_thread, other_clock) = if range.write
             > current_clocks.clock[range.write_index]
         {
             // Convert the write action into the vector clock it
@@ -799,14 +820,19 @@ impl VClockAlloc {
         let other_thread_info = global.print_thread_metadata(thread_mgr, other_thread);
 
         // Throw the data-race detection.
-        throw_ub_format!(
-            "Data race detected between {} on {} and {} on {} at {:?}",
-            action,
-            current_thread_info,
-            other_action,
-            other_thread_info,
-            ptr_dbg,
-        )
+        Err(err_machine_stop!(TerminationInfo::DataRace {
+            ptr: ptr_dbg,
+            op1: RacingOp {
+                action: action.to_string(),
+                thread_info: current_thread_info,
+                span: current_clocks.clock.as_slice()[current_index.index()].span_data(),
+            },
+            op2: RacingOp {
+                action: other_action.to_string(),
+                thread_info: other_thread_info,
+                span: other_clock.as_slice()[other_thread.index()].span_data(),
+            },
+        }))?
     }
 
     /// Detect racing atomic read and writes (not data races)
@@ -840,12 +866,14 @@ impl VClockAlloc {
         range: AllocRange,
         machine: &MiriMachine<'_, '_>,
     ) -> InterpResult<'tcx> {
+        let current_span = machine.current_span();
         let global = machine.data_race.as_ref().unwrap();
         if global.race_detecting() {
-            let (index, clocks) = global.current_thread_state(&machine.threads);
+            let (index, mut clocks) = global.current_thread_state_mut(&machine.threads);
             let mut alloc_ranges = self.alloc_ranges.borrow_mut();
             for (offset, range) in alloc_ranges.iter_mut(range.start, range.size) {
-                if let Err(DataRace) = range.read_race_detect(&clocks, index) {
+                if let Err(DataRace) = range.read_race_detect(&mut clocks, index, current_span) {
+                    drop(clocks);
                     // Report data-race.
                     return Self::report_data_race(
                         global,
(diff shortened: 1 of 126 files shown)


RalfJung:
I assume this is intended to fix https://github.com/rust-lang/miri/issues/2205 ?

src/concurrency/vector_clock.rs
```diff
@@ -42,7 +47,37 @@ const SMALL_VECTOR: usize = 4;
 
 /// The type of the time-stamps recorded in the data-race detector
 /// set to a type of unsigned integer
```
oli-obk, line 49: that comment is odd. Is it just explaining that this is a `u32`?

src/concurrency/vector_clock.rs
```diff
@@ -88,11 +123,14 @@ impl VClock {
...
     /// this will panic if the vector index overflows
     #[inline]
-    pub fn increment_index(&mut self, idx: VectorIdx) {
+    pub fn increment_index(&mut self, idx: VectorIdx, current_span: Span) {
         let idx = idx.index();
         let mut_slice = self.get_mut_with_min_len(idx + 1);
         let idx_ref = &mut mut_slice[idx];
-        *idx_ref = idx_ref.checked_add(1).expect("Vector clock overflow")
+        idx_ref.time = idx_ref.time.checked_add(1).expect("Vector clock overflow");
+        if current_span != DUMMY_SP {
```
oli-obk, line 131:
```suggestion
        if !current_span.is_dummy() {
```

src/concurrency/vector_clock.rs
```diff
@@ -102,14 +140,31 @@ impl VClock {
...
+            let r_span = r.span;
             *l = r.max(*l);
+            if l.span == DUMMY_SP {
+                if r_span != DUMMY_SP {
+                    l.span = r_span;
+                }
+                if l_span != DUMMY_SP {
+                    l.span = l_span;
+                }
+            }
```
oli-obk, line 153:
```suggestion
            l.span = l.span.substitute_dummy(r_span.substitute_dummy(l_span));
```

src/concurrency/vector_clock.rs
```diff
@@ -102,14 +140,31 @@ impl VClock {
...
     pub fn set_at_index(&mut self, other: &Self, idx: VectorIdx) {
         let mut_slice = self.get_mut_with_min_len(idx.index() + 1);
+
+        let prev_span = mut_slice[idx.index()].span;
+
         mut_slice[idx.index()] = other[idx];
+
+        if other[idx].span == DUMMY_SP {
+            mut_slice[idx.index()].span = prev_span;
+        }
```
oli-obk, line 167:
```suggestion
        let span = &mut mut_slice[idx.index()].span;
        *span = span.substitute_dummy(prev_span);
```

src/concurrency/vector_clock.rs
```diff
@@ -313,7 +368,14 @@ impl Index<VectorIdx> for VClock {
 
     #[inline]
     fn index(&self, index: VectorIdx) -> &VTimestamp {
-        self.as_slice().get(index.to_u32() as usize).unwrap_or(&0)
+        self.as_slice().get(index.to_u32() as usize).unwrap_or(&VTimestamp::NONE)
+    }
+}
+
+impl IndexMut<VectorIdx> for VClock {
```
oli-obk, line 375: Side-note: we could probably use https://doc.rust-lang.org/nightly/nightly-rustc/rustc_index/vec/struct.IndexVec.html for `VClock` and https://doc.rust-lang.org/nightly/nightly-rustc/rustc_index/macro.newtype_index.html for `VectorIdx`

src/concurrency/vector_clock.rs
```diff
@@ -323,24 +385,25 @@ impl Index<VectorIdx> for VClock {
...
         assert_ne!(c1, c2);
-        c1.increment_index(VectorIdx(53));
+        c1.increment_index(VectorIdx(53), DUMMY_SP);
         assert_ne!(c1, c2);
-        c2.increment_index(VectorIdx(5));
+        c2.increment_index(VectorIdx(5), DUMMY_SP);
         assert_eq!(c1, c2);
     }
 
+    /*
```
oli-obk, line 406: why is this commented out?

tests/fail/data_race/atomic_write_na_read_race2.stderr
```diff
@@ -4,6 +4,16 @@ error: Undefined Behavior: Data race detected between Atomic Store on thread `<u
...
+help: The Atomic Store on thread `<unnamed>` is here
+  --> $DIR/atomic_write_na_read_race2.rs:LL:CC
+   |
+LL | ...   (&*c.0).store(32, Ordering::SeqCst);
+   |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+help: The Read on thread `<unnamed>` is here
+  --> RUSTLIB/core/src/ptr/mod.rs:LL:CC
+   |
+LL | }
+   | ^
```
oli-obk, line 16: not sure how useful these are without an extra annotation. I'm surprised the spans don't contain expansion information actually. That could be something we need to fix on the rustc side. I'm guessing it's destructor related?

oli-obk (changes requested):

tests/fail/data_race/atomic_write_na_read_race2.stderr
oli-obk: This may be because of the lack of overwriting when there are `DUMMY_SP`. We end up picking the span from the last successful use of that race site

saethlin: Pretty sure it was an old bug that I swear I fixed before... we need to update the span for a clock, even if the operation is a race. Failing to do so results in these wacky spans.

src/concurrency/vector_clock.rs
saethlin: This used to just be a type alias, I just failed to update the docs.

src/concurrency/vector_clock.rs
saethlin: Hm, `IndexVec` just wraps a `Vec` and currently this uses `SmallVec`, and this PR is already messing with the inline capacity. There are ambient concerns about the data race detector overhead: https://github.com/rust-lang/miri/issues/2716 https://github.com/rust-lang/miri/issues/1689 which in my experience are not backed up by profiling, but I still think the prudent thing to do is adjust one thing at a time here. Happy to look into this in a separate PR.

oli-obk: Makes sense

src/concurrency/vector_clock.rs
```diff
@@ -40,9 +45,40 @@ impl From<u32> for VectorIdx {
 /// clock vectors larger than this will be stored on the heap
 const SMALL_VECTOR: usize = 4;
 
-/// The type of the time-stamps recorded in the data-race detector
-/// set to a type of unsigned integer
-pub type VTimestamp = u32;
+/// The time-stamps recorded in the data-race detector consist of both
```
oli-obk, line 48: Everything starting at consists should be on the fields. Internal docs use the flag to generate docs for private items

oli-obk (approved):
r=me after a doc nit

src/concurrency/data_race.rs
```diff
@@ -361,6 +363,8 @@ impl MemoryCellClocks {
             atomic.read_vector.set_at_index(&clocks.clock, index);
             Ok(())
         } else {
+            let atomic = self.atomic_mut();
+            atomic.read_vector.set_at_index(&clocks.clock, index);
```
RalfJung, line 367: This is duplicated with the previous `if` arm. Can that be avoided? (Happens a few more times below.)

src/concurrency/data_race.rs
```diff
@@ -621,30 +631,35 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: MiriInterpCxExt<'mir, 'tcx> {
...
+                        clocks.write_seqcst.join(&data_race.last_sc_write.borrow());
+                    }
 
-                // Increment timestamp in case of release semantics.
-                Ok(atomic != AtomicFenceOrd::Acquire)
-            })
+                    // Increment timestamp in case of release semantics.
+                    Ok(atomic != AtomicFenceOrd::Acquire)
+                },
+                current_span,
```
RalfJung, line 661: Please keep the closure as the last argument

src/concurrency/data_race.rs
```diff
@@ -704,7 +721,7 @@ impl VClockAlloc {
                 | MiriMemoryKind::ExternStatic
                 | MiriMemoryKind::Tls,
             )
-            | MemoryKind::CallerLocation => (0, VectorIdx::MAX_INDEX),
+            | MemoryKind::CallerLocation => (VTimestamp::NONE, VectorIdx::MAX_INDEX),
```
RalfJung, line 724: "None" is a somewhat odd term for this. Maybe "zero"?

src/concurrency/init_once.rs
```diff
@@ -172,7 +173,7 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
 
         // Each complete happens-before the end of the wait
         if let Some(data_race) = &this.machine.data_race {
-            data_race.validate_lock_release(&mut init_once.data_race, current_thread);
+            data_race.validate_lock_release(&mut init_once.data_race, current_thread, current_span);
```
RalfJung, line 176: Would it make sense to move the `current_span` calls behind the fn call? That seems nicer than putting the burden on all callers.  Basically instead of passing current_thread and current_span it could pass `&Machine`?

EDIT: Ah, that doesn't work since it overlaps with `data_race`. Hm... I guess we can go ahead with separate arguments for now, but if you have a nice idea for how to avoid having to pass so many arguments, that would be great.

src/concurrency/vector_clock.rs
```diff
@@ -40,9 +45,40 @@ impl From<u32> for VectorIdx {
...
+/// a 32-bit unsigned integer which is the actual timestamp, and a `Span`
+/// so that diagnostics can report what code was responsible for an operation.
+#[derive(Clone, Copy, Debug, Eq)]
```
RalfJung, line 51: The `Eq` is suspicious. This will also compare spans. Is that really what we want? Why is it even needed?

EDIT: Oh, PartialEq contains all the logic. yeah that is confusing...

src/diagnostics.rs
```diff
@@ -205,6 +222,13 @@ pub fn report_error<'tcx, 'mir>(
                 vec![(Some(*span), format!("the `{link_name}` symbol is defined here"))],
             Int2PtrWithStrictProvenance =>
                 vec![(None, format!("use Strict Provenance APIs (https://doc.rust-lang.org/nightly/std/ptr/index.html#strict-provenance, https://crates.io/crates/sptr) instead"))],
+            DataRace { ptr: _, op1, op2 } =>
+                vec![
+                    (Some(op1.span), format!("The {} on {} is here", op1.action, op1.thread_info)),
```
RalfJung, line 227: Should probably start with `the`, not `The`, to be consistent?

tests/fail/data_race/alloc_read_race.stderr
```diff
@@ -4,6 +4,16 @@ error: Undefined Behavior: Data race detected between Read on thread `<unnamed>`
 LL |             *pointer.load(Ordering::Relaxed)
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Data race detected between Read on thread `<unnamed>` and Allocate on thread `<unnamed>` at ALLOC
    |
+help: The Read on thread `<unnamed>` is here
+  --> $DIR/alloc_read_race.rs:LL:CC
+   |
+LL | ...   *pointer.load(Ordering::Relaxed)
+   |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+help: The Allocate on thread `<unnamed>` is here
```
RalfJung, line 12: Nit: is "is" the right verb? One of them happened in the past. Do we know which one? Can we make the message "Earlier, the X on thread Y happened here" and "Now, the X' on thread Y' happened here"? In fact isn't the latter span redundant with the main error span?

RalfJung:
@saethlin could you do benchmarks with `-Zmiri-disable-stacked-borrows -Zmiri-disable-validation` so that we can an idea for how expensive this is?

tests/fail/data_race/alloc_read_race.stderr
saethlin: Hah I was also not sure what words to use here. This is a data race, so strictly does it make any sense to say that one of these operations happens before the other? They happen in a particular order in the interpreter, but should diagnostics be commenting on the interpreter or the program semantics (or lack thereof)?

I chose present tense because I imagine this diagnostic as a response to the current user experience, where someone would rightfully ask "But Miri, where is the read/write/allocate/deallocate?" so now Miri says "The read/write/allocate/deallocate is here"

(As I describe this I realize it is not consistent with how I worded the Stacked Borrows diagnostics. Ah well.)

RalfJung: With vector clocks, there could actually be considerate amounts of "real time" between the two operations, even in an observable way. But you also have a point.

What about the fact that one of these two spans is always redundant since it is the same as where the error points? I think it'd be better to just say "the other access was over there" or so?

src/concurrency/vector_clock.rs
saethlin: I could have sworn clippy wanted the Eq derived. I agree it's simpler explicitly.

saethlin:
Before:
```
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/backtraces/Cargo.toml
  Time (mean ± σ):      2.262 s ±  0.005 s    [User: 2.204 s, System: 0.054 s]
  Range (min … max):    2.256 s …  2.270 s    5 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/mse/Cargo.toml
  Time (mean ± σ):     389.0 ms ±   4.6 ms    [User: 351.8 ms, System: 36.8 ms]
  Range (min … max):   383.3 ms … 394.1 ms    7 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/serde1/Cargo.toml
  Time (mean ± σ):     930.7 ms ±   6.6 ms    [User: 882.4 ms, System: 47.3 ms]
  Range (min … max):   924.3 ms … 940.7 ms    5 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/serde2/Cargo.toml
  Time (mean ± σ):      1.844 s ±  0.033 s    [User: 1.795 s, System: 0.047 s]
  Range (min … max):    1.812 s …  1.885 s    5 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/slice-get-unchecked/Cargo.toml
  Time (mean ± σ):     322.3 ms ±   4.2 ms    [User: 282.1 ms, System: 40.0 ms]
  Range (min … max):   313.0 ms … 326.0 ms    9 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/unicode/Cargo.toml
  Time (mean ± σ):     717.5 ms ±   2.1 ms    [User: 681.5 ms, System: 35.4 ms]
  Range (min … max):   714.5 ms … 720.2 ms    5 runs
```

After:
```
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/backtraces/Cargo.toml
  Time (mean ± σ):      2.275 s ±  0.038 s    [User: 2.201 s, System: 0.057 s]
  Range (min … max):    2.245 s …  2.341 s    5 runs
  
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/mse/Cargo.toml
  Time (mean ± σ):     393.9 ms ±   1.2 ms    [User: 360.0 ms, System: 33.6 ms]
  Range (min … max):   392.4 ms … 396.1 ms    7 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/serde1/Cargo.toml
  Time (mean ± σ):     933.1 ms ±   3.2 ms    [User: 888.6 ms, System: 43.6 ms]
  Range (min … max):   927.8 ms … 935.5 ms    5 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/serde2/Cargo.toml
  Time (mean ± σ):      1.909 s ±  0.006 s    [User: 1.851 s, System: 0.056 s]
  Range (min … max):    1.899 s …  1.916 s    5 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/slice-get-unchecked/Cargo.toml
  Time (mean ± σ):     325.8 ms ±   2.0 ms    [User: 287.1 ms, System: 38.5 ms]
  Range (min … max):   323.5 ms … 329.6 ms    9 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/unicode/Cargo.toml
  Time (mean ± σ):     715.4 ms ±   3.4 ms    [User: 672.1 ms, System: 42.6 ms]
  Range (min … max):   711.0 ms … 720.2 ms    5 runs
```

This is about the size of the effect I expected.

src/concurrency/data_race.rs
saethlin: I think I got all of them?

tests/fail/data_race/alloc_read_race.stderr
saethlin: Yeah, I've tried to reduce the duplication here, but I feel backed into a corner again by the limited diagnostics API. Doesn't change the fact that this is still much better than before.

RalfJung: This seems potentially confusing, since after the `help` we have a backtrace that attaches to the "main span", not the "other access span". I think that's why I think it is good to say something about the "earlier conflicting access". Also if both accesses are of the same kind, saying "the Write" is not very informative.

So what about
```
error: Undefined Behavior: Data race detected between (1) Atomic Store on thread `<unnamed>` and (2) Read on thread `<unnamed>` at ALLOC. Access (2) just happened here
  --> $DIR/atomic_write_na_read_race1.rs:LL:CC
   |
LL |             *atomic_ref.get_mut()
   |             ^^^^^^^^^^^^^^^^^^^^^
   |
help: and access (1) occurred earlier here
  --> $DIR/atomic_write_na_read_race1.rs:LL:CC
   |
LL |             atomic_ref.store(32, Ordering::SeqCst)
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = note: BACKTRACE:
   = note: inside closure at $DIR/atomic_write_na_read_race1.rs:LL:CC
```
Yeah it is annoying that the same text is repeated for the error title and label -- that would be https://github.com/rust-lang/miri/issues/2200, which will be very annoying work since all the ui tests will need adjustment.

saethlin: I agree there is potential for confusion with the backtrace. I often find the Stacked Borrows errors a bit visually confusing, and have often wanted some other way to format all this information. I'll try to keep an eye on how this plays out.

I like this numbering strategy overall, I'm just not sure about calling these an "access". That word has a particular meaning in Stacked Borrows, and feel like calling an allocation or deallocation an access might be confusing to newcomers. I think we can just drop the word without loss of clarity?

---

Also it occurs to me that we're now comparing a "local" span with the topmost frame, so things could look a bit odd if the race occurs through a standard library function that isn't `#[track_caller]`. I can't  find any examples of this, so it may be that the way people write data races, this just isn't a concern. Just mentioning this so you're aware of it, I think we shouldn't cook up a solution until we have a real example.

src/concurrency/vector_clock.rs
RalfJung: This comment seems to be still open.

tests/fail/data_race/alloc_read_race.stderr
RalfJung: > Also it occurs to me that we're now comparing a "local" span with the topmost frame, so things could look a bit odd if the race occurs through a standard library function that isn't #[track_caller]. 

The main backtrace will also skip `track_caller` frames by default so I think one needs to pass `-Zmiri-backtrace=full` to get anything weird?

I pushed an attempt to clarify the backtrace situation a bit. What do you think?

tests/fail/data_race/atomic_read_na_write_race1.stderr
```diff
@@ -1,12 +1,17 @@
-error: Undefined Behavior: Data race detected between Atomic Load on thread `<unnamed>` and Write on thread `<unnamed>` at ALLOC
+error: Undefined Behavior: Data race detected between (1) Atomic Load on thread `<unnamed>` and (2) Write on thread `<unnamed>` at ALLOC. (1) just happened here
   --> $DIR/atomic_read_na_write_race1.rs:LL:CC
    |
 LL |             (&*c.0).load(Ordering::SeqCst)
-   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Data race detected between Atomic Load on thread `<unnamed>` and Write on thread `<unnamed>` at ALLOC
+   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Data race detected between (1) Atomic Load on thread `<unnamed>` and (2) Write on thread `<unnamed>` at ALLOC. (1) just happened here
    |
+help: and (2) occurred earlier here
```
RalfJung, line 7: So (2) happened before (1) in execution order. That seems confusing? I had deliberately swapped them in my proposal.

saethlin: 🤦 I was so sure I was following your proposal

tests/fail/data_race/alloc_read_race.stderr
saethlin: I think the clarification is good. It's a marginal improvement, but an improvement.

Here is an example of a program that produces the flavor of weird output I'm referring to:
```
use std::sync::atomic::AtomicPtr;
fn main() {
    let mut v: Vec<i32> = Vec::with_capacity(2);
    let ptr = AtomicPtr::new(v.as_mut_ptr());
    let handle = std::thread::spawn(move || {
        let mut v = unsafe { Vec::from_raw_parts(ptr.into_inner(), 0, 2) };
        v.push(0);
    });
    v.push(0);
    let _ = handle.join();
}
```

```
error: Undefined Behavior: Data race detected between (1) Write on thread `main` and (2) Write on thread `<unnamed>` at alloc1617. (2) just happened here
    --> /home/ben/.rustup/toolchains/miri/lib/rustlib/src/rust/library/alloc/src/vec/mod.rs:1839:13
     |
1839 |             ptr::write(end, value);
     |             ^^^^^^^^^^^^^^^^^^^^^^ Data race detected between (1) Write on thread `main` and (2) Write on thread `<unnamed>` at alloc1617. (2) just happened here
     |
    --> demo.rs:9:5
     |
9    |     v.push(2);
     |     ^^^^^^^^^
     = note: inside `std::vec::Vec::<i32>::push` at /home/ben/.rustup/toolchains/miri/lib/rustlib/src/rust/library/alloc/src/vec/mod.rs:1839:13: 1839:35
note: inside closure
    --> demo.rs:7:9
     |
7    |         v.push(1);
     |         ^^^^^^^^^
```
The wording of our diagnostic pretty clearly asks the user to compare the topmost span of the backtrace and the first `help` span, but the topmost span can be dislocated from the user code by a fair bit if it is buried inside a standard library data structure or algorithm. For example, if a data race is triggered by the `PartialOrd` impl used in the depths of `slice::sort`. (as I write this I'm realizing I could have cooked up a more extreme example, ah well...)

RalfJung: Oh I see. Yeah... I think this PR here is still an improvement, feel free to file an issue to track this problem and possible ideas for how to fix it.

PR #2647, opened 2022-11-04
Track local frames incrementally during execution

saethlin:
https://github.com/rust-lang/miri/pull/2646 currently introduces a performance regression. This change removes that regression, and provides a minor perf improvement.

The existing lazy strategy for tracking the span we want to display is as efficient as it is only because we often create a `CurrentSpan` then never call `.get()`. Most of the calls to the `before_memory_read` and `before_memory_write` hooks do not create any event that we store in `AllocHistory`. But data races are totally different, any memory read or write may race, so every call to those hooks needs to access to the current local span.

So this changes to a strategy where we update some state in a `Thread` and `FrameExtra` incrementally, upon entering and existing each function call.

Before:
```
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/backtraces/Cargo.toml
  Time (mean ± σ):      5.532 s ±  0.022 s    [User: 5.444 s, System: 0.073 s]
  Range (min … max):    5.516 s …  5.569 s    5 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/mse/Cargo.toml
  Time (mean ± σ):     831.4 ms ±   3.0 ms    [User: 783.8 ms, System: 46.7 ms]
  Range (min … max):   828.7 ms … 836.1 ms    5 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/serde1/Cargo.toml
  Time (mean ± σ):      1.975 s ±  0.021 s    [User: 1.914 s, System: 0.059 s]
  Range (min … max):    1.939 s …  1.990 s    5 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/serde2/Cargo.toml
  Time (mean ± σ):      4.060 s ±  0.051 s    [User: 3.983 s, System: 0.071 s]
  Range (min … max):    3.972 s …  4.100 s    5 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/slice-get-unchecked/Cargo.toml
  Time (mean ± σ):     784.9 ms ±   8.2 ms    [User: 746.5 ms, System: 37.7 ms]
  Range (min … max):   772.9 ms … 793.3 ms    5 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/unicode/Cargo.toml
  Time (mean ± σ):      1.679 s ±  0.006 s    [User: 1.623 s, System: 0.055 s]
  Range (min … max):    1.673 s …  1.687 s    5 runs
```
After:
```
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/backtraces/Cargo.toml
  Time (mean ± σ):      5.330 s ±  0.037 s    [User: 5.232 s, System: 0.084 s]
  Range (min … max):    5.280 s …  5.383 s    5 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/mse/Cargo.toml
  Time (mean ± σ):     818.9 ms ±   3.7 ms    [User: 776.8 ms, System: 41.3 ms]
  Range (min … max):   813.5 ms … 822.5 ms    5 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/serde1/Cargo.toml
  Time (mean ± σ):      1.927 s ±  0.011 s    [User: 1.864 s, System: 0.061 s]
  Range (min … max):    1.917 s …  1.945 s    5 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/serde2/Cargo.toml
  Time (mean ± σ):      3.974 s ±  0.020 s    [User: 3.893 s, System: 0.076 s]
  Range (min … max):    3.956 s …  4.004 s    5 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/slice-get-unchecked/Cargo.toml
  Time (mean ± σ):     780.0 ms ±   5.3 ms    [User: 740.3 ms, System: 39.0 ms]
  Range (min … max):   771.2 ms … 784.5 ms    5 runs
 
Benchmark 1: cargo +miri miri run --manifest-path /home/ben/miri/bench-cargo-miri/unicode/Cargo.toml
  Time (mean ± σ):      1.643 s ±  0.007 s    [User: 1.584 s, System: 0.058 s]
  Range (min … max):    1.635 s …  1.654 s    5 runs
```
(This change is marginal, but the point is that it avoids a much more significant regression)

### DIFF
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -118,6 +118,13 @@ pub struct Thread<'mir, 'tcx> {
     /// The virtual call stack.
     stack: Vec<Frame<'mir, 'tcx, Provenance, FrameData<'tcx>>>,
 
+    /// The index of the topmost user-relevant frame in `stack`. This field must contain
+    /// the value produced by `get_top_user_relevant_frame`.
+    /// The `None` state here represents
+    /// This field is a cache to reduce how often we call that method. The cache is manually
+    /// maintained inside `MiriMachine::after_stack_push` and `MiriMachine::after_stack_pop`.
+    top_user_relevant_frame: Option<usize>,
+
     /// The join status.
     join_status: ThreadJoinStatus,
 
@@ -147,6 +154,38 @@ impl<'mir, 'tcx> Thread<'mir, 'tcx> {
     fn thread_name(&self) -> &[u8] {
         if let Some(ref thread_name) = self.thread_name { thread_name } else { b"<unnamed>" }
     }
+
+    /// Return the top user-relevant frame, if there is one.
+    /// Note that the choice to return `None` here when there is no user-relevant frame is part of
+    /// justifying the optimization that only pushes of user-relevant frames require updating the
+    /// `top_user_relevant_frame` field.
+    fn compute_top_user_relevant_frame(&self) -> Option<usize> {
+        self.stack
+            .iter()
+            .enumerate()
+            .rev()
+            .find_map(|(idx, frame)| if frame.extra.is_user_relevant { Some(idx) } else { None })
+    }
+
+    /// Re-compute the top user-relevant frame from scratch.
+    pub fn recompute_top_user_relevant_frame(&mut self) {
+        self.top_user_relevant_frame = self.compute_top_user_relevant_frame();
+    }
+
+    /// Set the top user-relevant frame to the given value. Must be equal to what
+    /// `get_top_user_relevant_frame` would return!
+    pub fn set_top_user_relevant_frame(&mut self, frame_idx: usize) {
+        debug_assert_eq!(Some(frame_idx), self.compute_top_user_relevant_frame());
+        self.top_user_relevant_frame = Some(frame_idx);
+    }
+
+    pub fn top_user_relevant_frame(&self) -> usize {
+        debug_assert_eq!(self.top_user_relevant_frame, self.compute_top_user_relevant_frame());
+        // This can be called upon creation of an allocation. We create allocations while setting up
+        // parts of the Rust runtime when we do not have any stack frames yet, so we need to handle
+        // empty stacks.
+        self.top_user_relevant_frame.unwrap_or_else(|| self.stack.len().saturating_sub(1))
+    }
 }
 
 impl<'mir, 'tcx> std::fmt::Debug for Thread<'mir, 'tcx> {
@@ -167,6 +206,7 @@ impl<'mir, 'tcx> Default for Thread<'mir, 'tcx> {
             state: ThreadState::Enabled,
             thread_name: None,
             stack: Vec::new(),
+            top_user_relevant_frame: None,
             join_status: ThreadJoinStatus::Joinable,
             panic_payload: None,
             last_error: None,
@@ -184,8 +224,15 @@ impl<'mir, 'tcx> Thread<'mir, 'tcx> {
 
 impl VisitTags for Thread<'_, '_> {
     fn visit_tags(&self, visit: &mut dyn FnMut(SbTag)) {
-        let Thread { panic_payload, last_error, stack, state: _, thread_name: _, join_status: _ } =
-            self;
+        let Thread {
+            panic_payload,
+            last_error,
+            stack,
+            top_user_relevant_frame: _,
+            state: _,
+            thread_name: _,
+            join_status: _,
+        } = self;
 
         panic_payload.visit_tags(visit);
         last_error.visit_tags(visit);
@@ -414,7 +461,7 @@ impl<'mir, 'tcx: 'mir> ThreadManager<'mir, 'tcx> {
     }
 
     /// Get a shared borrow of the currently active thread.
-    fn active_thread_ref(&self) -> &Thread<'mir, 'tcx> {
+    pub fn active_thread_ref(&self) -> &Thread<'mir, 'tcx> {
         &self.threads[self.active_thread]
     }
 
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -179,12 +179,14 @@ impl<'mir, 'tcx> Thread<'mir, 'tcx> {
         self.top_user_relevant_frame = Some(frame_idx);
     }
 
-    pub fn top_user_relevant_frame(&self) -> usize {
+    /// Returns the topmost frame that is considered user-relevant, or the
+    /// top of the stack if there is no such frame, or `None` if the stack is empty.
+    pub fn top_user_relevant_frame(&self) -> Option<usize> {
         debug_assert_eq!(self.top_user_relevant_frame, self.compute_top_user_relevant_frame());
         // This can be called upon creation of an allocation. We create allocations while setting up
         // parts of the Rust runtime when we do not have any stack frames yet, so we need to handle
         // empty stacks.
-        self.top_user_relevant_frame.unwrap_or_else(|| self.stack.len().saturating_sub(1))
+        self.top_user_relevant_frame.or_else(|| self.stack.len().checked_sub(1))
     }
 }
 
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -936,31 +936,14 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
 }
 
 impl<'mir, 'tcx> MiriMachine<'mir, 'tcx> {
-    pub fn current_span(&self) -> CurrentSpan<'_, 'mir, 'tcx> {
-        CurrentSpan { current_frame_idx: None, machine: self }
-    }
-}
-
-/// A `CurrentSpan` should be created infrequently (ideally once) per interpreter step. It does
-/// nothing on creation, but when `CurrentSpan::get` is called, searches the current stack for the
-/// topmost frame which corresponds to a local crate, and returns the current span in that frame.
-/// The result of that search is cached so that later calls are approximately free.
-#[derive(Clone)]
-pub struct CurrentSpan<'a, 'mir, 'tcx> {
-    current_frame_idx: Option<usize>,
-    machine: &'a MiriMachine<'mir, 'tcx>,
-}
-
-impl<'a, 'mir: 'a, 'tcx: 'a + 'mir> CurrentSpan<'a, 'mir, 'tcx> {
-    pub fn machine(&self) -> &'a MiriMachine<'mir, 'tcx> {
-        self.machine
-    }
-
-    /// Get the current span, skipping non-local frames.
+    /// Get the current span in the topmost function which is workspace-local and not
+    /// `#[track_caller]`.
     /// This function is backed by a cache, and can be assumed to be very fast.
-    pub fn get(&mut self) -> Span {
-        let idx = self.current_frame_idx();
-        self.stack().get(idx).map(Frame::current_span).unwrap_or(rustc_span::DUMMY_SP)
+    pub fn current_span(&self) -> Span {
+        self.stack()
+            .get(self.top_user_relevant_frame())
+            .map(Frame::current_span)
+            .unwrap_or(rustc_span::DUMMY_SP)
     }
 
     /// Returns the span of the *caller* of the current operation, again
@@ -968,46 +951,27 @@ impl<'a, 'mir: 'a, 'tcx: 'a + 'mir> CurrentSpan<'a, 'mir, 'tcx> {
     /// current operation is not in a local crate.
     /// This is useful when we are processing something which occurs on function-entry and we want
     /// to point at the call to the function, not the function definition generally.
-    pub fn get_caller(&mut self) -> Span {
+    pub fn caller_span(&self) -> Span {
         // We need to go down at least to the caller (len - 2), or however
-        // far we have to go to find a frame in a local crate.
-        let local_frame_idx = self.current_frame_idx();
+        // far we have to go to find a frame in a local crate which is also not #[track_caller].
+        let frame_idx = self.top_user_relevant_frame();
         let stack = self.stack();
-        let idx = cmp::min(local_frame_idx, stack.len().saturating_sub(2));
-        stack.get(idx).map(Frame::current_span).unwrap_or(rustc_span::DUMMY_SP)
+        let frame_idx = cmp::min(frame_idx, stack.len().saturating_sub(2));
+        stack.get(frame_idx).map(Frame::current_span).unwrap_or(rustc_span::DUMMY_SP)
     }
 
     fn stack(&self) -> &[Frame<'mir, 'tcx, Provenance, machine::FrameData<'tcx>>] {
-        self.machine.threads.active_thread_stack()
+        self.threads.active_thread_stack()
     }
 
-    fn current_frame_idx(&mut self) -> usize {
-        *self
-            .current_frame_idx
-            .get_or_insert_with(|| Self::compute_current_frame_index(self.machine))
+    fn top_user_relevant_frame(&self) -> usize {
+        self.threads.active_thread_ref().top_user_relevant_frame()
     }
 
-    // Find the position of the inner-most frame which is part of the crate being
-    // compiled/executed, part of the Cargo workspace, and is also not #[track_caller].
-    #[inline(never)]
-    fn compute_current_frame_index(machine: &MiriMachine<'_, '_>) -> usize {
-        machine
-            .threads
-            .active_thread_stack()
-            .iter()
-            .enumerate()
-            .rev()
-            .find_map(|(idx, frame)| {
-                let def_id = frame.instance.def_id();
-                if (def_id.is_local() || machine.local_crates.contains(&def_id.krate))
-                    && !frame.instance.def.requires_caller_location(machine.tcx)
-                {
-                    Some(idx)
-                } else {
-                    None
-                }
-            })
-            .unwrap_or(0)
+    pub fn is_user_relevant(&self, frame: &Frame<'mir, 'tcx, Provenance>) -> bool {
+        let def_id = frame.instance.def_id();
+        (def_id.is_local() || self.local_crates.contains(&def_id.krate))
+            && !frame.instance.def.requires_caller_location(self.tcx)
     }
 }
 
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -940,9 +940,8 @@ impl<'mir, 'tcx> MiriMachine<'mir, 'tcx> {
     /// `#[track_caller]`.
     /// This function is backed by a cache, and can be assumed to be very fast.
     pub fn current_span(&self) -> Span {
-        self.stack()
-            .get(self.top_user_relevant_frame())
-            .map(Frame::current_span)
+        self.top_user_relevant_frame()
+            .map(|frame_idx| self.stack()[frame_idx].current_span())
             .unwrap_or(rustc_span::DUMMY_SP)
     }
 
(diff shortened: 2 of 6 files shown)


RalfJung:
So those benchmarks are without https://github.com/rust-lang/miri/pull/2646?

src/concurrency/thread.rs
```diff
@@ -118,6 +118,9 @@ pub struct Thread<'mir, 'tcx> {
     /// The virtual call stack.
     stack: Vec<Frame<'mir, 'tcx, Provenance, FrameData<'tcx>>>,
 
+    /// The index of the topmost local frame in `stack`
```
RalfJung, line 121:
```suggestion
    /// The index of the topmost local frame in `stack`.
```

RalfJung, line 121: Please state exactly what the invariant for this field is. Probably that it always holds the value computed by `reset_top_local`?

As usual, I am concerned about such redundant state and it getting out of sync. Can we have some sanity checks when debug assertions are enabled?

src/concurrency/thread.rs
```diff
@@ -414,10 +425,22 @@ impl<'mir, 'tcx: 'mir> ThreadManager<'mir, 'tcx> {
     }
 
     /// Get a shared borrow of the currently active thread.
-    fn active_thread_ref(&self) -> &Thread<'mir, 'tcx> {
+    pub fn active_thread_ref(&self) -> &Thread<'mir, 'tcx> {
         &self.threads[self.active_thread]
     }
 
+    /// Reset the current-span caching logic. Mostly called when a new frame is added or removed.
```
RalfJung, line 432: What about the active thread changing?

src/machine.rs
```diff
@@ -1036,21 +1039,23 @@ impl<'mir, 'tcx> Machine<'mir, 'tcx> for MiriMachine<'mir, 'tcx> {
         (alloc_id, prov_extra): (AllocId, Self::ProvenanceExtra),
         range: AllocRange,
     ) -> InterpResult<'tcx> {
-        if let Some(data_race) = &mut alloc_extra.data_race {
-            data_race.write(
+        let mut current_span = machine.current_span();
+        if let Some(stacked_borrows) = &mut alloc_extra.stacked_borrows {
```
RalfJung, line 1043: You swapped the order in which the subsystems are called. Why?

saethlin:
> So those benchmarks are without #2646?

Yes. That PR adds extra work to the data race machinery in order to store the spans which has its own very small regression, so I wanted to keep the effects separate.

src/helpers.rs
```diff
@@ -955,7 +955,6 @@ impl<'mir, 'tcx> MiriMachine<'mir, 'tcx> {
 /// The result of that search is cached so that later calls are approximately free.
 #[derive(Clone)]
 pub struct CurrentSpan<'a, 'mir, 'tcx> {
-    current_frame_idx: Option<usize>,
     machine: &'a MiriMachine<'mir, 'tcx>,
 }
```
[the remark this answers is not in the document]
saethlin: Well we can't just pass a `Span` around, because we do want the functionality of `CurrentSpan::get_caller`, on some retags. So yes I agree it is now a bit of a silly type (and I need to fix up its docs) but I am not sure the code is improved by removing it.

src/machine.rs
saethlin: This _might_ need to happen for the other PR, but honestly it's in here because I was sloppy when separating out this PR.

src/helpers.rs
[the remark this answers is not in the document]
RalfJung: IMO it'd make more sense to pass around a machine -- that might also make some other arguments redundant then.

src/concurrency/thread.rs
saethlin: I agree that state getting out of sync is a risk here and I don't like it either.

I will of course add some debug assertions, but what do you think about doing some encapsulation around the stack? This is really metadata for that one field in particular, but the problem is that it really has no encapsulation at all because there is a function in the public interface that just returns a mutable reference to the whole field. Is it too much extra complicatiion to refactor the use of `Machine::stack_mut` in rustc to be more clear operations like `push` and `pop` and `get_nth_frame`?

RalfJung: I'm always in favor of more encapsulation. :) However, would it help here? We also have the ThreadManager in Miri that has direct access to all the stacks.

src/concurrency/thread.rs
RalfJung: Oh I see, this is per-thread state... makes sense.

RalfJung, line 432:
```suggestion
    /// Reset the current-span caching logic of the active thread.
    /// Mostly called when a new frame is added or removed.
```

src/machine.rs
```diff
@@ -1122,13 +1115,19 @@ impl<'mir, 'tcx> Machine<'mir, 'tcx> for MiriMachine<'mir, 'tcx> {
             None
         };
 
+        let def_id = frame.instance.def_id();
+        let is_local = (def_id.is_local() || ecx.machine.local_crates.contains(&def_id.krate))
+            && !frame.instance.def.requires_caller_location(ecx.machine.tcx);
```
oli-obk, line 1120: with this it isn't purely about `is_local` anymore. Maybe at least sprinkle some (doc) comments around to make it clear what the `is_local` is being used for

saethlin: The choice here hasn't been purely about whether the Instance is local for a while.

In fact, even the logic for `Machine::is_local(&FrameInfo)` which does only check if a Frame is local is only ever used in the diagnostics code, where we also check if a Frame is `#[track_caller]`. 

In any case, I've made an attempt at picking a less-wrong name. But I don't think it is good, and would appreciate any suggestion for a better one.

src/concurrency/thread.rs
```diff
@@ -118,6 +119,17 @@ pub struct Thread<'mir, 'tcx> {
     /// The virtual call stack.
     stack: Vec<Frame<'mir, 'tcx, Provenance, FrameData<'tcx>>>,
 
+    /// The index of the topmost local frame in `stack`. If `Some`, this field must contain
```
RalfJung, line 122:
```suggestion
    /// The index of the topmost user-relevant frame in `stack`. If `Some`, this field must contain
```

src/machine.rs
```diff
@@ -1183,6 +1194,11 @@ impl<'mir, 'tcx> Machine<'mir, 'tcx> for MiriMachine<'mir, 'tcx> {
         mut frame: Frame<'mir, 'tcx, Provenance, FrameData<'tcx>>,
         unwinding: bool,
     ) -> InterpResult<'tcx, StackPopJump> {
+        if frame.extra.is_user_relevant {
+            // All that we store is whether or not the frame we just removed is local, so now we
+            // have no idea what if any the next topmost local frame is.
```
RalfJung, line 1199:
```suggestion
            // have no idea where the next topmost local frame is.
```

src/machine.rs
RalfJung: I like the new name. :shrug:

RalfJung:
LGTM, what are the benchmarks saying? How could the cache be invalidated less aggressively than on each pop?

src/machine.rs
```diff
@@ -1183,6 +1194,11 @@ impl<'mir, 'tcx> Machine<'mir, 'tcx> for MiriMachine<'mir, 'tcx> {
...
+            ecx.active_thread_mut().set_top_user_relevant_frame(None);
```
RalfJung, line 1200: I guess the alternative would be to do a search here, right? Then we'd always have a hot cache, which simplifies logic elsewhere so it might be worth it.

In fact we could speed that up by storing a 'pointer' to the next user-relevant frame -- on push, remember the current user-relevant frame and store it in the new frame as `prev_user_relevant_frame`, on pop restore that. Both of these are O(1) operations.

src/machine.rs
```diff
@@ -1120,13 +1117,20 @@ impl<'mir, 'tcx> Machine<'mir, 'tcx> for MiriMachine<'mir, 'tcx> {
             None
         };
 
+        let def_id = frame.instance.def_id();
+        let is_user_relevant = (def_id.is_local()
+            || ecx.machine.local_crates.contains(&def_id.krate))
+            && !frame.instance.def.requires_caller_location(ecx.machine.tcx);
```
RalfJung, line 1123: This is the main potential for slowdown, right? We are now computing this on each push, as opposed to only computing it when someone actually needs it.

But I guess this is cheap enough, pushes are rare enough, and it is needed often enough, that that is worth it?

src/machine.rs
saethlin: :facepalm: I guess I never tried this. Yes. This code path is so incredibly cold that it doesn't make sense to only invalidate here.

saethlin:
> How could the cache be invalidated less aggressively than on each pop?

I force-pushed after making this comment. Previously I was invalidating on every pop, not just pops of local frames. Most frames we create are in the standard library.

src/machine.rs
saethlin: Yes, this is the main potential for slowdown with respect to the previous strategy. And yes, I think there are ways to optimize this more, but we spend 5 times as many cycles in the call to `Frame::current_span` (all the rest of the code in this system totals 0.1% of cycles, `Frame::current_span` is 0.5%). So the slowest part of this is code which was already in here, and all data indicates that overall we are faster, so I favor simplicity.

saethlin:
> what are the benchmarks saying?

I updated the PR description. We have the same marginal improvement.

src/concurrency/thread.rs
```diff
@@ -147,6 +153,28 @@ impl<'mir, 'tcx> Thread<'mir, 'tcx> {
     fn thread_name(&self) -> &[u8] {
         if let Some(ref thread_name) = self.thread_name { thread_name } else { b"<unnamed>" }
     }
+
+    fn get_top_user_relevant_frame(&self) -> Option<usize> {
```
RalfJung, line 157: This sounds like a simple getter but is an expensive computation -- I think it should be named differently. Maybe `compute_`?

src/concurrency/thread.rs
```diff
@@ -147,6 +153,28 @@ impl<'mir, 'tcx> Thread<'mir, 'tcx> {
...
+        self.stack
+            .iter()
+            .enumerate()
+            .rev()
+            .find_map(|(idx, frame)| if frame.extra.is_user_relevant { Some(idx) } else { None })
+    }
+
+    pub fn recompute_top_user_relevant_frame(&mut self) {
```
RalfJung, line 165:
```suggestion
    /// Re-compute the top user-relevant frame from srcatch.
    pub fn recompute_top_user_relevant_frame(&mut self) {
```

src/concurrency/thread.rs
```diff
@@ -147,6 +153,28 @@ impl<'mir, 'tcx> Thread<'mir, 'tcx> {
...
+        self.top_user_relevant_frame = self.get_top_user_relevant_frame();
+    }
+
+    pub fn set_top_user_relevant_frame(&mut self, frame_idx: usize) {
```
RalfJung, line 169:
```suggestion
    /// Set the top user-relevant frame to the given value. Must be equal to what
    /// `get_top_user_relevant_frame` would return!
    pub fn set_top_user_relevant_frame(&mut self, frame_idx: usize) {
```

src/concurrency/thread.rs
```diff
@@ -147,6 +153,28 @@ impl<'mir, 'tcx> Thread<'mir, 'tcx> {
...
+        debug_assert_eq!(Some(frame_idx), self.get_top_user_relevant_frame());
+        self.top_user_relevant_frame = Some(frame_idx);
+    }
+
+    pub fn top_user_relevant_frame(&self) -> usize {
+        debug_assert_eq!(self.top_user_relevant_frame, self.get_top_user_relevant_frame());
+        self.top_user_relevant_frame.unwrap_or_else(|| self.stack.len().saturating_sub(1))
```
RalfJung, line 176: Any reason not to include the `.unwrap_or_else(|| self.stack.len().saturating_sub(1))` in `get_top_user_relevant_frame`?

Ah, then it wouldn't be enough to just update this when we push a user-relevant frame, right? Seems subtle enough to be worth documenting.

Can this even be called when the stack is empty? Looks like `current_span` would just panic then anyway.

RalfJung (commented):
All right so this is a simplification without perf loss. I like it, just some nits.

src/concurrency/thread.rs
saethlin: > Ah, then it wouldn't be enough to just update this when we push a user-relevant frame, right? Seems subtle enough to be worth documenting.

Correct. Will do.

> Can this even be called when the stack is empty? Looks like current_span would just panic then anyway.

Yes this is called when the stack is empty. We query the current span when creating an `AllocHistory`, and we create allocations before the program has started. I'll try to document that.

src/concurrency/thread.rs
```diff
@@ -147,6 +154,38 @@ impl<'mir, 'tcx> Thread<'mir, 'tcx> {
...
+    pub fn set_top_user_relevant_frame(&mut self, frame_idx: usize) {
+        debug_assert_eq!(Some(frame_idx), self.compute_top_user_relevant_frame());
+        self.top_user_relevant_frame = Some(frame_idx);
+    }
+
+    pub fn top_user_relevant_frame(&self) -> usize {
+        debug_assert_eq!(self.top_user_relevant_frame, self.compute_top_user_relevant_frame());
+        // This can be called upon creation of an allocation. We create allocations while setting up
+        // parts of the Rust runtime when we do not have any stack frames yet, so we need to handle
+        // empty stacks.
```
RalfJung, line 186: I searched for all top_user_relevant_frame and couldn't find this... I probably have tomatoes on my eyes (hm I dont think this German expression translates into English :P whatever) but could you point me to where this happens?

src/concurrency/thread.rs
```diff
@@ -147,6 +154,38 @@ impl<'mir, 'tcx> Thread<'mir, 'tcx> {
...
+    }
+
+    /// Set the top user-relevant frame to the given value. Must be equal to what
+    /// `get_top_user_relevant_frame` would return!
+    pub fn set_top_user_relevant_frame(&mut self, frame_idx: usize) {
+        debug_assert_eq!(Some(frame_idx), self.compute_top_user_relevant_frame());
+        self.top_user_relevant_frame = Some(frame_idx);
+    }
+
+    pub fn top_user_relevant_frame(&self) -> usize {
```
RalfJung, line 182:
```suggestion
    /// Returns the topmost frame that is considered user-relevant, or the
    /// top of the stack if there is no such frame, or 0 if the stack is empty.
    pub fn top_user_relevant_frame(&self) -> usize {
```

src/helpers.rs
```diff
@@ -936,78 +936,42 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
-                if (def_id.is_local() || machine.local_crates.contains(&def_id.krate))
-                    && !frame.instance.def.requires_caller_location(machine.tcx)
-                {
-                    Some(idx)
-                } else {
-                    None
-                }
-            })
-            .unwrap_or(0)
+    pub fn is_user_relevant(&self, frame: &Frame<'mir, 'tcx, Provenance>) -> bool {
```
RalfJung, line 971:
```suggestion
    /// This is the source of truth for the `is_user_relevant` flag in our `FrameExtra`.
    pub fn is_user_relevant(&self, frame: &Frame<'mir, 'tcx, Provenance>) -> bool {
```

src/helpers.rs
```diff
@@ -936,78 +936,42 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
-        let idx = cmp::min(local_frame_idx, stack.len().saturating_sub(2));
-        stack.get(idx).map(Frame::current_span).unwrap_or(rustc_span::DUMMY_SP)
+        let frame_idx = cmp::min(frame_idx, stack.len().saturating_sub(2));
+        stack.get(frame_idx).map(Frame::current_span).unwrap_or(rustc_span::DUMMY_SP)
     }
 
     fn stack(&self) -> &[Frame<'mir, 'tcx, Provenance, machine::FrameData<'tcx>>] {
-        self.machine.threads.active_thread_stack()
+        self.threads.active_thread_stack()
     }
```
RalfJung, line 965: `MiriMachine` already has a stack method in `machine.rs`, so I don't think we need this one?

src/concurrency/thread.rs
saethlin: In `helpers.rs`, `Machine::current_span` now calls `top_user_relevant_frame`: https://github.com/rust-lang/miri/pull/2647/files#diff-c2a0fff5575ebf83634ea4da334a24d52775bcef6700ed6f9d0bc4e3db79e42aR944

RalfJung: Ah, I missed the `unwrap_or`... thanks!

src/concurrency/thread.rs
RalfJung: Instead of returning an OOB index for the empty stack, would it make more sense to return `Option<usize>` here and promise that the index is always valid if it is `Some`?

src/helpers.rs
saethlin: `MiriMachine` has an associated function that takes an `InterpCx`, which is not what we have here.

Unless you are suggesting I pass `&InterpCx` all over stacked borrows now, as I suggested long ago ;)

RalfJung: > Unless you are suggesting I pass &InterpCx all over stacked borrows now, as I suggested long ago ;)

That's not even possible. ;)

src/machine.rs
```diff
@@ -1183,6 +1189,11 @@ impl<'mir, 'tcx> Machine<'mir, 'tcx> for MiriMachine<'mir, 'tcx> {
         mut frame: Frame<'mir, 'tcx, Provenance, FrameData<'tcx>>,
         unwinding: bool,
     ) -> InterpResult<'tcx, StackPopJump> {
+        if frame.extra.is_user_relevant {
+            // All that we store is whether or not the frame we just removed is local, so now we
+            // have no idea where the next topmost local frame is. So we recompute it.
+            ecx.active_thread_mut().recompute_top_user_relevant_frame();
```
RalfJung, line 1195:
```suggestion
            // (If this ever becomes a bottleneck, we could have `push` store the previous
            // user-relevant frame and restore that here.)
            ecx.active_thread_mut().recompute_top_user_relevant_frame();
```

src/helpers.rs
```diff
@@ -954,17 +953,17 @@ impl<'mir, 'tcx> MiriMachine<'mir, 'tcx> {
     pub fn caller_span(&self) -> Span {
         // We need to go down at least to the caller (len - 2), or however
         // far we have to go to find a frame in a local crate which is also not #[track_caller].
-        let frame_idx = self.top_user_relevant_frame();
-        let stack = self.stack();
-        let frame_idx = cmp::min(frame_idx, stack.len().saturating_sub(2));
-        stack.get(frame_idx).map(Frame::current_span).unwrap_or(rustc_span::DUMMY_SP)
+        self.top_user_relevant_frame()
+            .map(|frame_idx| cmp::min(frame_idx, self.stack().len() - 2))
```
RalfJung, line 957: This changes semantics I think?
```suggestion
            .map(|frame_idx| cmp::min(frame_idx, self.stack().len().saturating_sub(2)))
```

saethlin: Does it make sense to ask for the caller span (especially in an FnEntry retag) when we have one stack frame? I wrote this thinking "no, that would be a bug, so a crash would be good"

RalfJung: then it should be `checked_sub(2).unwrap()`

PR #2650, opened 2022-11-05
Rustup

RalfJung:


RalfJung:
I still need to test if a build against a locally built rustc still works.

RalfJung:
Looking good!
@bors r+

PR #2657, opened 2022-11-07
Add miri_alloc, miri_dealloc

mkroening:
`miri_alloc` and `miri_dealloc` are basically the same as `__rust_alloc` and `__rust_dealloc` respectively, but without the check for a global allocator.

This should allow bootstrapping an allocator in environments, where no fundamental way of allocating memory is available (`no_std` + `alloc` in Miri).

### DIFF
--- a/README.md
+++ b/README.md
@@ -568,6 +568,15 @@ extern "Rust" {
     /// program) the contents of a section of program memory, as bytes. Bytes
     /// written using this function will emerge from the interpreter's stderr.
     fn miri_write_to_stderr(bytes: &[u8]);
+
+    /// Miri-provided extern function to allocate memory from the interpreter.
+    /// 
+    /// This is useful when no fundamental way of allocating memory is
+    /// available, e.g. when using `no_std` + `alloc`.
+    fn miri_alloc(size: usize, align: usize) -> *mut u8;
+
+    /// Miri-provided extern function to deallocate memory.
+    fn miri_dealloc(ptr: *mut u8, size: usize, align: usize);
 }
 ```
 
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -123,7 +123,7 @@ impl fmt::Display for MiriMemoryKind {
         use self::MiriMemoryKind::*;
         match self {
             Rust => write!(f, "Rust heap"),
-            Miri => write!(f, "Miri heap"),
+            Miri => write!(f, "Miri bare-metal heap"),
             C => write!(f, "C heap"),
             WinHeap => write!(f, "Windows heap"),
             Machine => write!(f, "machine-managed memory"),
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -77,6 +77,8 @@ impl VisitTags for FrameData<'_> {
 pub enum MiriMemoryKind {
     /// `__rust_alloc` memory.
     Rust,
+    /// `miri_alloc` memory.
+    Miri,
     /// `malloc` memory.
     C,
     /// Windows `HeapAlloc` memory.
@@ -110,7 +112,7 @@ impl MayLeak for MiriMemoryKind {
     fn may_leak(self) -> bool {
         use self::MiriMemoryKind::*;
         match self {
-            Rust | C | WinHeap | Runtime => false,
+            Rust | Miri | C | WinHeap | Runtime => false,
             Machine | Global | ExternStatic | Tls => true,
         }
     }
@@ -121,6 +123,7 @@ impl fmt::Display for MiriMemoryKind {
         use self::MiriMemoryKind::*;
         match self {
             Rust => write!(f, "Rust heap"),
+            Miri => write!(f, "Miri heap"),
             C => write!(f, "C heap"),
             WinHeap => write!(f, "Windows heap"),
             Machine => write!(f, "machine-managed memory"),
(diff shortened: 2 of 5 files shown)


tests/pass/miri-alloc.rs
```diff
@@ -0,0 +1,12 @@
...
+    fn miri_dealloc(ptr: *mut u8, size: usize, align: usize);
+}
+
+fn main() {
+    unsafe {
+        let ptr = miri_alloc(123, 1);
+        std::ptr::write_bytes(ptr, 0, 123);
+        miri_dealloc(ptr, 123, 1);
+    }
+}
```
oli-obk, line 12: To make sure this actually works without a global allocator, add `#![no_std]` to the test and use `core::ptr::write_bytes`.

src/shims/foreign_items.rs
```diff
@@ -483,6 +483,40 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+
+                let ptr = this.allocate_ptr(
+                    Size::from_bytes(size),
+                    Align::from_bytes(align).unwrap(),
+                    MiriMemoryKind::Rust.into(),
+                )?;
+
+                this.write_pointer(ptr, dest)?;
+
+                return Ok(EmulateByNameResult::NeedsJumping);
```
oli-obk, line 502: this shares a lot of logic with the `__*` code paths. I think we could share the logic with some well placed closures and conditions.

README.md
```diff
@@ -568,6 +568,15 @@ extern "Rust" {
...
     fn miri_write_to_stderr(bytes: &[u8]);
+
+    /// Miri-provided extern function to allocate memory from the interpreter.
+    /// 
+    /// This is useful when no fundamental way of allocating memory is
+    /// available, e.g. when using `no_std` + `alloc`.
+    fn miri_alloc(size: usize, align: usize) -> *mut u8;
+
+    /// Miri-provided extern function to deallocate memory.
+    fn miri_dealloc(ptr: *mut u8, size: usize, align: usize);
```
saethlin, line 579: Why not use the same signatures as `std::alloc::alloc` and `std::alloc::dealloc`?

mkroening: @RalfJung told me it would be easier to implement if we take `size` and `align` separately instead of one `Layout`. We are able to share most of the code with the `__rust_*` paths now.

Apart from that, there is no real user-facing reason for this approach. If the maintainers preferred using `Layout`, I'll look into it.

src/shims/foreign_items.rs
```diff
@@ -528,7 +528,16 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                     )?;
```
RalfJung, line 528: I think this "bare Miri memory" should have a different memory kind, not `MiriMemoryKind::Rust`.

oli-obk: Yea... I was wondering about that. It's cleaner, but is there,much value to it?

RalfJung: It's easy to do so I'd say there's no reason not to do it.

README.md
saethlin: For sure it's easier to implement this way. I don't have any particular preference, I'm just wary of these being a regrettable API inconsistency. They're a Rust interface, it would make sense to speak the Rust types for allocation.

I know I have had my own share of struggles writing code reads non-primitive types from an ABI boundary like this. We should really have a go-to resource for how to do that. So if you and @RalfJung want to move your project forward by going with this interface (for now?) instead of getting hung up on writing the required code, that wouldn't bother me.

RalfJung: I view these as low-level machine interfaces; the nice Rust interfaces can be provided by a crate that wraps them.

src/shims/foreign_items.rs
mkroening: I added `MiriMemoryKind::Miri` for `miri_alloc` memory.

src/machine.rs
```diff
@@ -121,6 +123,7 @@ impl fmt::Display for MiriMemoryKind {
         use self::MiriMemoryKind::*;
         match self {
             Rust => write!(f, "Rust heap"),
+            Miri => write!(f, "Miri heap"),
```
RalfJung, line 126:
```suggestion
            Miri => write!(f, "Miri bare-metal heap"),
```

tests/pass/miri-alloc.rs
```diff
@@ -0,0 +1,29 @@
...
+extern "Rust" {
+    fn miri_alloc(size: usize, align: usize) -> *mut u8;
+    fn miri_dealloc(ptr: *mut u8, size: usize, align: usize);
+}
+
+#[start]
+fn start(_: isize, _: *const *const u8) -> isize {
+    unsafe {
+        let ptr = miri_alloc(123, 1);
+        core::ptr::write_bytes(ptr, 0, 123);
```
RalfJung, line 17:
```suggestion
        core::ptr::write_bytes(ptr, 0u8, 123);
```
I never know which argument is the data byte and which is the length; let's make the type system help us. :)

RalfJung:
Looks good, thanks!
@bors r+

Issue #2629, opened 2022-10-28
scoped threads vs thread leaks


RalfJung:
The standard library scoped threads do not use the OS facilities for joining threads, but instead they implement join logic with their own synchronization primitives. (This seems to help with cleaning up thread resources ASAP rather than keeping them around until the scope finishes and all threads are joined.)

Unfortunately this can show up in Miri as a thread leak, if a thread gets scheduled away just after the custom join logic is done, and then the main thread finishes.

For now, a possible work-around is to use `-Zmiri-ignore-leaks`. But of course it'd be nice to find a better solution. Maybe, after the main thread is done, instead of halting execution immediately we could yield that thread N times (for some suitable N), thus giving other threads a chance to finish?

oli-obk:
As an immediate step we should probably add a test for it ^^

The "leaked" threads don't have anything interesting to finish doing, so maybe we could have the scoped cleanup logic mark them as leakable or just join them under cfg(miri)

RalfJung:
Yeah I considered just adding this to the standard library somewhere after `main` returns:
```rust
if cfg!(miri) {
  for _ in 0..100 { thread::yield_now(); }
}
```
But honestly that seemed a bit too hacky.^^ Seems better to do this on the interpreter side.

oli-obk:
No I meant... actually join the threads in scoped thread code instead of just cleaning up their resources and letting them run to completion.

RalfJung:
That would be a much bigger and more annoying set of `cfg(miri)`.

RalfJung:
Yes it's a tight window. I definitely saw this happen, which is how I even noticed, but don't remember the exact code. But a test will always risk false negatives here.

RalfJung:
For example, this code
```rust
use std::thread;

fn main() {
let mut a = vec![1, 2, 3];
let mut x = 0;

thread::scope(|s| {
    s.spawn(|| {
        println!("hello from the first scoped thread");
        // We can borrow `a` here.
        dbg!(&a);
    });
    s.spawn(|| {
        println!("hello from the second scoped thread");
        // We can even mutably borrow `x` here,
        // because no other threads are using it.
        x += a[0] + a[2];
    });
});
}
```
reproduces the problem with seed 26.

saethlin:
Is there any particular reason to believe that "yield to all other threads on thread exit" isn't something we should/can do in the interpreter instead?

RalfJung:
That's exactly what I suggested in the OP. :)

PR #2660, opened 2022-11-09
Yield from the main thread a few times at shutdown

saethlin:
Closes https://github.com/rust-lang/miri/issues/2629

... how should we test this? The naive approach is to run `./miri many-seeds ./miri run example.rs`: https://github.com/rust-lang/miri/issues/2629#issuecomment-1295113360
```rust
use std::thread;

fn main() {
  thread::scope(|s| {
    s.spawn(|| {});
  });
}
```
But running _many_ seeds takes an annoyingly long time. How many seeds is enough? And how does such a test fit into our existing testing system?

(I'm also not enthused about the way I'm continuing to pile on complexity into the scheduler here, should this PR include a refactor?)

### DIFF
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -623,31 +623,34 @@ impl<'mir, 'tcx: 'mir> ThreadManager<'mir, 'tcx> {
     /// long as we can and switch only when we have to (the active thread was
     /// blocked, terminated, or has explicitly asked to be preempted).
     fn schedule(&mut self, clock: &Clock) -> InterpResult<'tcx, SchedulingAction> {
-        // If we are the main thread, and our call stack is empty but our state is Enabled, we are
-        // about to terminate.
-        // But, if there are any other threads which can execute, yield to them instead of falling
-        // through to the termination state. This is to give them a chance to clean up and quit.
-        let active_thread = &self.threads[self.active_thread];
-        if self.active_thread == MAIN_THREAD
-            && active_thread.should_terminate()
-            && self.threads.iter().any(|t| t.state == ThreadState::Enabled && !t.stack.is_empty())
-            && self.main_thread_yields_at_shutdown_remaining > 0
-        {
-            self.yield_active_thread = true;
-            self.main_thread_yields_at_shutdown_remaining -= 1;
-        } else {
-            // Check whether the thread ought to terminate, and if so start executing TLS
-            // destructors.
-            if self.threads[self.active_thread].should_terminate() {
-                self.threads[self.active_thread].state = ThreadState::Terminated;
-                return Ok(SchedulingAction::ExecuteDtors);
-            }
+        // Check whether the thread ought to terminate, and if so start executing TLS
+        // destructors.
+        if self.threads[self.active_thread].should_terminate() {
+            self.threads[self.active_thread].state = ThreadState::Terminated;
+            return Ok(SchedulingAction::ExecuteDtors);
         }
+
         // If we get here again and the thread is *still* terminated, there are no more dtors to run.
         if self.threads[MAIN_THREAD].state == ThreadState::Terminated {
-            // The main thread terminated; stop the program.
-            // We do *not* run TLS dtors of remaining threads, which seems to match rustc behavior.
-            return Ok(SchedulingAction::Stop);
+            // If we are the main thread and we are Terminated, we ought to stop.
+            // But, if there are any other threads which can execute, yield to them instead of falling
+            // through to the termination state. This is to give them a chance to clean up and quit.
+            let active_thread = &self.threads[self.active_thread];
+            if self.active_thread == MAIN_THREAD
+                && active_thread.state == ThreadState::Terminated
+                && self
+                    .threads
+                    .iter()
+                    .any(|t| t.state == ThreadState::Enabled && !t.stack.is_empty())
+                && self.main_thread_yields_at_shutdown_remaining > 0
+            {
+                self.yield_active_thread = true;
+                self.main_thread_yields_at_shutdown_remaining -= 1;
+            } else {
+                // The main thread terminated; stop the program.
+                // We do *not* run TLS dtors of remaining threads, which seems to match rustc behavior.
+                return Ok(SchedulingAction::Stop);
+            }
         }
         // This thread and the program can keep going.
         if self.threads[self.active_thread].state == ThreadState::Enabled
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -633,7 +633,7 @@ impl<'mir, 'tcx: 'mir> ThreadManager<'mir, 'tcx> {
         // If we are the main thread, and our call stack is empty but our state is Enabled, we are
         // about to terminate.
         // But, if there are any other threads which can execute, yield to them instead of falling
-        // through to the termination state.
+        // through to the termination state. This is to give them a chance to clean up and quit.
         let active_thread = &self.threads[self.active_thread];
         if self.active_thread == MAIN_THREAD
             && active_thread.stack.is_empty()
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -136,7 +136,7 @@ pub struct Thread<'mir, 'tcx> {
 }
 
 impl<'mir, 'tcx> Thread<'mir, 'tcx> {
-    /// Check if the thread is done executing (no more stack frames).
+    /// Check if the thread is just done executing (still enabled, but no more stack frames).
     fn should_terminate(&self) -> bool {
         self.state == ThreadState::Enabled && self.stack.is_empty()
     }
@@ -689,15 +689,16 @@ impl<'mir, 'tcx: 'mir> ThreadManager<'mir, 'tcx> {
         }
         self.yield_active_thread = false;
 
-        // The main thread is special, it is allowed to yield when it is shutting down and has
-        // nothing to execute. So if another thread yields back to the main thread when it is in
-        // this partly-shut-down state, we make another pass through the scheduler to either yield
-        // back off them main thread, or enter its shutdown sequence.
-        if self.active_thread == MAIN_THREAD && self.threads[self.active_thread].stack.is_empty() {
-            return self.schedule(clock);
-        }
-
         if self.threads[self.active_thread].state == ThreadState::Enabled {
+            // The main thread is special, it is allowed to yield when it is shutting down and has
+            // nothing to execute. So if another thread yields back to the main thread when it is in
+            // this partly-shut-down state, we make another pass through the scheduler to either yield
+            // back off them main thread, or enter its shutdown sequence.
+            if self.active_thread == MAIN_THREAD
+                && self.threads[self.active_thread].stack.is_empty()
+            {
+                return self.schedule(clock);
+            }
             return Ok(SchedulingAction::ExecuteStep);
         }
         // We have not found a thread to execute.
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -136,16 +136,9 @@ pub struct Thread<'mir, 'tcx> {
 }
 
 impl<'mir, 'tcx> Thread<'mir, 'tcx> {
-    /// Check if the thread is done executing (no more stack frames). If yes,
-    /// change the state to terminated and return `true`.
-    fn check_terminated(&mut self) -> bool {
-        if self.state == ThreadState::Enabled {
-            if self.stack.is_empty() {
-                self.state = ThreadState::Terminated;
-                return true;
-            }
-        }
-        false
+    /// Check if the thread is done executing (no more stack frames).
+    fn should_terminate(&self) -> bool {
+        self.state == ThreadState::Enabled && self.stack.is_empty()
     }
 
     /// Get the name of the current thread, or `<unnamed>` if it was not set.
@@ -636,18 +629,17 @@ impl<'mir, 'tcx: 'mir> ThreadManager<'mir, 'tcx> {
         // through to the termination state. This is to give them a chance to clean up and quit.
         let active_thread = &self.threads[self.active_thread];
         if self.active_thread == MAIN_THREAD
-            && active_thread.stack.is_empty()
-            && active_thread.state == ThreadState::Enabled
+            && active_thread.should_terminate()
             && self.threads.iter().any(|t| t.state == ThreadState::Enabled && !t.stack.is_empty())
             && self.main_thread_yields_at_shutdown_remaining > 0
         {
             self.yield_active_thread = true;
             self.main_thread_yields_at_shutdown_remaining -= 1;
         } else {
-            // Check whether the thread has **just** terminated (`check_terminated`
-            // checks whether the thread has popped all its stack and if yes, sets
-            // the thread state to terminated).
-            if self.threads[self.active_thread].check_terminated() {
+            // Check whether the thread ought to terminate, and if so start executing TLS
+            // destructors.
+            if self.threads[self.active_thread].should_terminate() {
+                self.threads[self.active_thread].state = ThreadState::Terminated;
                 return Ok(SchedulingAction::ExecuteDtors);
             }
         }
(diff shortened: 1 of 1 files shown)


oli-obk:
> how should we test this?

we could pick a seed that causes the problems and hardcode it into the test's arguments. When it fails due to rustc changes, we need to find a new seed.

saethlin:
> that causes the problems

The seed that causes the problem depends on the details of stdlib internals and MIR generation. Nearly all seeds are fine. So the problem we have is that any selection of a seed may not reproduce the problematic scheduling in the future.

> When it fails due to rustc changes, we need to find a new seed.

The problem would be the test passing with whatever seed we pick, even though the bug is back.

oli-obk:
> The problem would be the test passing with whatever seed we pick, even though the bug is back.

oh 🤦 yea, we'd need a way to disable this yielding scheme and then run the test once without it, expect it to fail and once with it, expecting it to pass... that's annoying and even more noise here.

RalfJung:
Yeah sadly I don't think we can test for this.

> (I'm also not enthused about the way I'm continuing to pile on complexity into the scheduler here, should this PR include a refactor?)

Not this PR, but if you have a good idea for a refactor that'd be great!
I keep wondering if we can use generators for this somehow, it's kind of a state machine thing... though this applies more to the TLS logic, which is quite terrible too, then the core scheduler.

src/concurrency/thread.rs
```diff
@@ -620,11 +630,26 @@ impl<'mir, 'tcx: 'mir> ThreadManager<'mir, 'tcx> {
...
     fn schedule(&mut self, clock: &Clock) -> InterpResult<'tcx, SchedulingAction> {
-        // Check whether the thread has **just** terminated (`check_terminated`
-        // checks whether the thread has popped all its stack and if yes, sets
-        // the thread state to terminated).
-        if self.threads[self.active_thread].check_terminated() {
-            return Ok(SchedulingAction::ExecuteDtors);
+        // If we are the main thread, and our call stack is empty but our state is Enabled, we are
+        // about to terminate.
+        // But, if there are any other threads which can execute, yield to them instead of falling
+        // through to the termination state.
```
RalfJung, line 636:
```suggestion
        // through to the termination state. This is to give them a chance to clean up and quit.
```

src/concurrency/thread.rs
```diff
@@ -620,11 +630,26 @@ impl<'mir, 'tcx: 'mir> ThreadManager<'mir, 'tcx> {
...
+        let active_thread = &self.threads[self.active_thread];
+        if self.active_thread == MAIN_THREAD
+            && active_thread.stack.is_empty()
+            && active_thread.state == ThreadState::Enabled
```
RalfJung, line 640: This is the same check as in `check_terminated`, right? I think we should make that function really just *check* and not also change the state -- this has confused be in the past several times. Then we can also reuse it here, right?

src/concurrency/thread.rs
```diff
@@ -671,6 +696,15 @@ impl<'mir, 'tcx: 'mir> ThreadManager<'mir, 'tcx> {
             }
         }
         self.yield_active_thread = false;
+
+        // The main thread is special, it is allowed to yield when it is shutting down and has
+        // nothing to execute. So if another thread yields back to the main thread when it is in
+        // this partly-shut-down state, we make another pass through the scheduler to either yield
+        // back off them main thread, or enter its shutdown sequence.
+        if self.active_thread == MAIN_THREAD && self.threads[self.active_thread].stack.is_empty() {
```
RalfJung, line 704: Should this be inside the `if self.threads[self.active_thread].state == ThreadState::Enabled`? I think it should.

src/concurrency/thread.rs
```diff
@@ -620,11 +630,26 @@ impl<'mir, 'tcx: 'mir> ThreadManager<'mir, 'tcx> {
...
+        // through to the termination state.
+        let active_thread = &self.threads[self.active_thread];
+        if self.active_thread == MAIN_THREAD
+            && active_thread.stack.is_empty()
+            && active_thread.state == ThreadState::Enabled
+            && self.threads.iter().any(|t| t.state == ThreadState::Enabled && !t.stack.is_empty())
+            && self.main_thread_yields_at_shutdown_remaining > 0
+        {
+            self.yield_active_thread = true;
+            self.main_thread_yields_at_shutdown_remaining -= 1;
```
RalfJung, line 645: Ideally we'd do this yielding *after* executing main thread TLS dtors (because it'd be UB to access them). Not sure how terrible that would be to implement though, so a FIXME is also fine.

src/concurrency/thread.rs
saethlin: Like this?
```diff
diff --git a/src/concurrency/thread.rs b/src/concurrency/thread.rs
index b59fcc8d..882c7eb7 100644
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -684,19 +684,20 @@ impl<'mir, 'tcx: 'mir> ThreadManager<'mir, 'tcx> {
             debug_assert_ne!(self.active_thread, id);
             if thread.state == ThreadState::Enabled {
                 self.active_thread = id;
+                // The main thread is special, it is allowed to yield when it is shutting down and has
+                // nothing to execute. So if another thread yields back to the main thread when it is in
+                // this partly-shut-down state, we make another pass through the scheduler to either yield
+                // back off them main thread, or enter its shutdown sequence.
+                if self.active_thread == MAIN_THREAD
+                    && self.threads[self.active_thread].stack.is_empty()
+                {
+                    return self.schedule(clock);
+                }
                 break;
             }
         }
         self.yield_active_thread = false;
 
-        // The main thread is special, it is allowed to yield when it is shutting down and has
-        // nothing to execute. So if another thread yields back to the main thread when it is in
-        // this partly-shut-down state, we make another pass through the scheduler to either yield
-        // back off them main thread, or enter its shutdown sequence.
-        if self.active_thread == MAIN_THREAD && self.threads[self.active_thread].stack.is_empty() {
-            return self.schedule(clock);
-        }
-
         if self.threads[self.active_thread].state == ThreadState::Enabled {
             return Ok(SchedulingAction::ExecuteStep);
         }
```
This doesn't compile because `IndexVec::iter_enumerated` uses `impl Trait` in return position:
```
error[E0502]: cannot borrow `*self` as mutable because it is also borrowed as immutable
   --> src/concurrency/thread.rs:694:28
    |
678 |           let threads = self
    |  _______________________-
679 | |             .threads
680 | |             .iter_enumerated()
    | |______________________________- immutable borrow occurs here
...
683 |           for (id, thread) in threads {
    |                               ------- a temporary with access to the immutable borrow is created here ...
...
694 |                       return self.schedule(clock);
    |                              ^^^^^^^^^^^^^^^^^^^^ mutable borrow occurs here
...
698 |           }
    |           - ... and the immutable borrow might be used here, when that temporary is dropped and runs the destructor for type `std::iter::Chain<std::iter::Skip<impl DoubleEndedIterator + ExactSizeIterator + Iterator<Item = (concurrency::thread::ThreadId, &concurrency::thread::Thread<'_, '_>)> + '_>, std::iter::Take<impl DoubleEndedIterator + ExactSizeIterator + Iterator<Item = (concurrency::thread::ThreadId, &concurrency::thread::Thread<'_, '_>)> + '_>>`
```

src/concurrency/thread.rs
saethlin: Yeah... I would like to understand how to do this correctly. The naive implementation is just shifting around the big check this PR adds so that the original `if should_terminate() { return Ok(ExecuteDtors) }` comes first, but that seems to produce data races whenever this yielding behavior kicks in: We run the main thread's dtors, then schedule another thread on the next step, and that thread reports a data race on one of a few allocations in the Rust runtime.

It wouldn't surprise me if these data races are legitimate. After all, we do have this comment:
```
                // The main thread terminated; stop the program.
                // We do *not* run TLS dtors of remaining threads, which seems to match rustc behavior.
 ```

File: src/concurrency/thread.rs [hunk 5]
```diff
@@ -136,16 +136,9 @@ pub struct Thread<'mir, 'tcx> {
...
-    fn check_terminated(&mut self) -> bool {
-        if self.state == ThreadState::Enabled {
-            if self.stack.is_empty() {
-                self.state = ThreadState::Terminated;
-                return true;
-            }
-        }
-        false
+    /// Check if the thread is done executing (no more stack frames).
+    fn should_terminate(&self) -> bool {
```
RalfJung, line 140:
```suggestion
    /// Check if the thread is just done executing (still enabled, but no more stack frames).
    fn should_terminate(&self) -> bool {
```

src/concurrency/thread.rs
RalfJung: Hard to say more without knowing the details of the race and how you implemented this. But we already currently will automatically yield to other threads while main thread Dtors are running, so we should already be seeing such races?

The best place to put this yielding behavior seems to be in the `if self.threads[MAIN_THREAD].state == ThreadState::Terminated` check, no? Instead of stopping, we keep going, but with some kind of bound.

src/concurrency/thread.rs
RalfJung: No i meant the one below. [Here](https://github.com/rust-lang/miri/blob/94ddca187760a1e139f8ac57d43f108268e74efd/src/concurrency/thread.rs#L674).

src/concurrency/thread.rs
RalfJung: Actually a bound on the steps is probably better than yielding... if automatic yielding is disabled, this might otherwise keep going forever even after the main thread stopped.

saethlin: I pushed a commit which does the change I was trying to describe above. I've seen two data races reported, one in the test suite and the other when I run this file:
```rust
use std::thread;

fn main() {
    thread::scope(|s| {
        s.spawn(|| {});
    });
}
```
With `./miri many-seeds ./miri demo.rs`

I agree that a bound on the extra steps sounds reasonable. Is the concern that this yielding behavior may take programs which currently exit under the non-yielding option and turn them into programs that never exit?

src/concurrency/thread.rs
```diff
@@ -623,31 +623,34 @@ impl<'mir, 'tcx: 'mir> ThreadManager<'mir, 'tcx> {
...
         if self.threads[MAIN_THREAD].state == ThreadState::Terminated {
-            // The main thread terminated; stop the program.
-            // We do *not* run TLS dtors of remaining threads, which seems to match rustc behavior.
-            return Ok(SchedulingAction::Stop);
+            // If we are the main thread and we are Terminated, we ought to stop.
+            // But, if there are any other threads which can execute, yield to them instead of falling
+            // through to the termination state. This is to give them a chance to clean up and quit.
+            let active_thread = &self.threads[self.active_thread];
+            if self.active_thread == MAIN_THREAD
+                && active_thread.state == ThreadState::Terminated
```
RalfJung, line 640: We already know `self.threads[MAIN_THREAD].state == ThreadState::Terminated` here, so this seems redundant?

src/concurrency/thread.rs
RalfJung: >  Is the concern that this yielding behavior may take programs which currently exit under the non-yielding option and turn them into programs that never exit?

Yes.

saethlin:
I believe alloc28 is `main`'s return place

RalfJung:
Is the active thread getting inconsistent or so? It is pointing at the `loop`; the last statements executed before that are
```
│ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ ├─11031ms  INFO rustc_const_eval::interpret::step return
│ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ ├─11031ms  INFO rustc_const_eval::interpret::eval_context popping stack frame (returning from function)
│ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │┌┘rustc_const_eval::interpret::eval_context::frame std::sys_common::thread_local_dtor::register_dtor_fallback::run_dtors
│ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ ├┘rustc_const_eval::interpret::eval_context::frame main::{closure#1}
│ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ │ ├─32962ms  INFO rustc_const_eval::interpret::step _12 = const ()
```
I think this race just happens when the main thread TLS dtors are done (i.e., immediately when the new logic kicks in).

RalfJung:
Oh I have a theory... when we are [reading the return place here](https://github.com/rust-lang/miri/blob/cb95524b403656a0a083fb3abeccfa9498b644c5/src/eval.rs#L369), is the active thread still the other thread? That would look like a race.

src/concurrency/thread.rs
saethlin: The TLS shims re-enable terminated threads. Is that a good thing? Do they have to be Enabled in order to execute user code?

RalfJung: Yes, threads must definitely be enabled when running TLS.

I wish they wouldn't be disabled and then re-enabled... this entire scheduler and TLS stuff needs a rework at some point, but I haven't yet found a nice way to factor this. I still think a generator would be good... basically `yield`ing a function to run, which is then executed until the stack is empty again, at which point the generator is consulted to figure out whether there is another function to run or whether this thread is done.

RalfJung: Ah sadly using generators for this [will not work](https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/Generators.20that.20borrow.20on.20each.20resume.3F).

RalfJung:
I implemented a scheduler refactoring with an alternative way to do this yielding in https://github.com/rust-lang/miri/pull/2699.

RalfJung:
Closing in favor of https://github.com/rust-lang/miri/pull/2699. Thanks for the PR!

Issue #2136, opened 2022-05-20
WinApi SYSTEM_INFO `dwPageSize` always returning 0


V0ldek:
This is a rather weird issue I reported first in the [`page_size`](https://github.com/Elzair/page_size_rs/issues/7) crate. When running Miri with any of the Tier 1 Windows targets the crate reports page size 0, which is a fatal error for my use, where I want to allocate aligned to the page boundary.

This is witnessed most easily by [the GitHub Action run](https://github.com/V0ldek/aligners/runs/6526456428?check_suite_focus=true). My code panics when the page size is not a power of two, so in this case it crashes with:

```bash
thread 'main' panicked at 'detected page size 0 that is not a power of two, this is unsupported', src/alignment.rs:111:21
```

The page size query works perfectly well for Unix targets.

## Replication

This can be confirmed by referencing `page_size` and asking for it like this:

```rust
fn main() {
    println!("Page size: {}", page_size::get());
}
```
and running Miri:

```bash
cargo +nightly miri run --target x86_64-pc-windows-gnu
```

resulting in:

```plain
> Page size: 0
```

Whereas running it without Miri gives the correct result (on my PC it's 4KB):

```bash
cargo run
```
```plain
> Page size: 4096
```

If it matters, the host OS seems to not affect this, the issue is the same running the command both from Unix and Windows.

RalfJung:
Miri's Windows support is a big hack, barely enough to get our own test suite to pass. Windows is effectively at best a Tier 2 target in Miri. (Not that we have a formal target policy.)

Basically, I don't have a lot of time to spend on Windows support in Miri myself. But help with the Windows support would certainly be appreciated. :D

This is probably caused by some code in [this file](https://github.com/rust-lang/miri/blob/master/src/shims/windows/foreign_items.rs), though I do not know which function is responsible.

V0ldek:
That would certainly be this code:

https://github.com/rust-lang/miri/blob/b96610b3de5bc9fd346d90700ba881aca8d503f0/src/shims/windows/foreign_items.rs#L112-L126

If I read it correctly, it fills the entire SYSTEM_INFO structure with zeroes and then sets a few selected fields to actual values.

V0ldek:
Hah, but the `page_size` crate just calls the WinApi SYSTEM_INFO :D

https://github.com/Elzair/page_size_rs/blob/290f3da944d0f1c86612172ec84958a9dbf910ac/src/lib.rs#L162-L170

[...]

Issue #1123, opened 2019-12-22
Figure out a way to do name resolution in value vs type namespace


RalfJung:
See the discussion at https://github.com/rust-lang/miri/pull/1101#discussion_r355505476: we'd like variants of `resolve_path` that specifically check the value *or* type namespace.

To my surprise, according to what @oli-obk said, this is actually not entirely trivial to do? Given that this is the basis of name resolution in rustc, I would have expected that to be much simpler than getting a weird "lookup in whatever namespace", which seems to be the operation we have right now.

PR #2661, opened 2022-11-10
Do field projections on known types

DrMeepster:
Currently, some shims will use `mplace_field` and related helper functions on pointers passed to them. This can cause problems such as #2136 and even ICEs if a field is not present.

This PR fixes this by having shims dereferencing pointers with a type from `std` or `libc` that we can rely on the layout and field names of instead of with whatever the user passed in.


oli-obk:
Do you have some concrete examples that failed before this PR? I have a hard time coming envisioning what this PR is actually doing.

RalfJung:
I tens to agree that if we can avoid trusting the user types, that is better. I was just too lazy to manually define the type each time... But I like the idea of taking the type from libc or std!

Did you just solve https://github.com/rust-lang/miri/issues/1123 as well in this? Nice!

oli-obk:
Instead of switching to using hardcoded types in miri, could we just verify the layout matches what we expect and keep the existing logic? Making wrong types an error instead of assuming types

RalfJung:
For raw ptr functions, wrong types are not really an error. Conceptually we should just tell Miri to read/write things at a given fixed offset, but that quickly gets awkward -- so this PR essentially uses the libc/std types  to determine that offset.

oli-obk:
Hmm I guess that makes sense, but still sad. Would it maybe be good to nudge people with a warning?

RalfJung:
At least on Windows people can't use the exact std type as those are private (and indeed https://github.com/rust-lang/miri/issues/2136 was caused by two different call sites using two different but equally valid types).

DrMeepster:
> Do you have some concrete examples that failed before this PR? I have a hard time coming envisioning what this PR is actually doing.

Here's one for windows. While I don't think anybody would actually do this, it still should work.
```rs
extern "system" {
    // this is completely legal, FILETIME has the same size and lower alignment than u64
    fn GetSystemTimeAsFileTime(out: *mut u64);
}

fn main() {
    unsafe {
        GetSystemTimeAsFileTime(&mut 0);
    }
}
```

src/helpers.rs
```diff
@@ -659,6 +664,25 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        let align = if this.machine.check_alignment != AlignmentCheck::None {
+            layout.align.abi
+        } else {
+            Align::ONE
+        };
+
+        this.check_ptr_access_align(ptr, layout.size, align, CheckInAllocMsg::DerefTest)?;
+
+        Ok(MPlaceTy::from_aligned_ptr(ptr, layout))
+    }
```
RalfJung, line 684: This could be just `deref_operand_and_offset` with offset 0, right?

DrMeepster: That has an assert to make sure its inbounds of the original type, so it behaves differently from this, which does all checks based on `layout`

RalfJung: Yeah... but I don't think there is a good reason for that.
what I don't like about this is that it essentially re-implements `deref_operand`. In particular we really shouldn't duplicate the `check_alignment` logic anywhere.

DrMeepster: Yeah, maybe this needs to live in rustc_const_eval so there can be shared code between this and deref_operand

RalfJung: I have a simpler idea, I'll just push to your branch.

RalfJung: Ah no that doesn't work. I'll make a rustc patch.

RalfJung:
@DrMeepster could you rebase? Now `check_mplace` is a public function which you can call to avoid most of the code duplication.

DrMeepster:
I noticed that almost every use of `deref_operand` is problematic, not just field accesses. While just reading from a pointer won't ice, it can cause false positives if the pointee type is wrong. I've changed these to use `deref_operand_as` too.

src/helpers.rs
```diff
@@ -659,31 +664,49 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         }
     }
 
+    /// Dereference a pointer operand to a place using `layout` instead of the pointer's declared type
+    fn deref_operand_as(
+        &self,
+        op: &OpTy<'tcx, Provenance>,
+        layout: TyAndLayout<'tcx>,
+    ) -> InterpResult<'tcx, MPlaceTy<'tcx, Provenance>> {
```
RalfJung, line 672: Now this is entirely equivalent to `deref_operand_and_offset(op, 0, layout, layout)` isn't it?

RalfJung: Oh you made `deref_operand_and_offset` call this one. Yeah that also works.

src/shims/time.rs
```diff
@@ -132,7 +134,7 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         })?;
         this.write_scalar(
             Scalar::from_i64(qpc),
-            &this.deref_operand(lpPerformanceCount_op)?.into(),
+            &this.deref_operand_as(lpPerformanceCount_op, this.machine.layouts.u64)?.into(),
```
RalfJung, line 137: I am not entirely sure this is worth it for primitive types... this adds quite a bit of churn, and makes the Miri code a lot harder to write. It can be hard to find out which Rust type some C API type corresponds to. Currently when we get it wrong, we get an error saying the size is wrong, which is pretty helpful. With this change we might not get an error at all, just clobbering bytes beyond what we are supposed to be writing to.

So if we do this I think it is important that `deref_operand_as` also checks that the layout given by the shim has the same size and the layout given by the user.

src/shims/unix/linux/sync.rs
```diff
@@ -85,16 +85,17 @@ pub fn futex<'tcx>(
...
+            let timeout_time = if this.ptr_is_null(this.read_pointer(timeout_op)?)? {
                 None
             } else {
                 if op & futex_realtime != 0 {
                     this.check_no_isolation(
                         "`futex` syscall with `op=FUTEX_WAIT` and non-null timeout with `FUTEX_CLOCK_REALTIME`",
                     )?;
                 }
+                let timeout =
+                    this.deref_operand_as(timeout_op, this.libc_ty_layout("timespec")?)?;
```
RalfJung, line 98: Here it seems like we'd want a `deref_pointer_as`, so that you don't need to use `timeout_op` twice.

src/shims/windows/sync.rs
```diff
@@ -7,10 +7,6 @@ use crate::concurrency::sync::{CondvarLock, RwLockMode};
 use crate::concurrency::thread::MachineCallback;
 use crate::*;
 
-const SRWLOCK_ID_OFFSET: u64 = 0;
-const INIT_ONCE_ID_OFFSET: u64 = 0;
-const CONDVAR_ID_OFFSET: u64 = 0;
```
RalfJung, line 12: Please either keep these constants, or add comments similar to [these](https://github.com/rust-lang/miri/blob/4ca3e276448ee5555f419d2a006ace4a60988388/src/shims/unix/sync.rs#L50-L57) to explain the data layout. Just a magic '0' without context is too confusing.

RalfJung:
We unfortunately had to force-push to Miri. Could you rebase this to new master? Let me know if you need help wrestling with git.

DrMeepster:
Rebase done, review comments not addressed yet.

src/shims/time.rs
RalfJung: > So if we do this I think it is important that deref_operand_as also checks that the layout given by the shim has the same size and the layout given by the user.

From what I understand this is explicitly something you did not want. So for now it's probably better to just stick to `deref_operand` for primitive types. We will already detect wrong sizes there anyway since we detect when the scalar written to a place does not match the size of the place. (And we deliberately changed that from an ICE to a UB error so that Miri could use this for its shims.)

RalfJung:
FYI I have cherry-picked the first commit from this PR into https://github.com/rust-lang/miri/pull/2696, to avoid conflicts. Please rebase once https://github.com/rust-lang/miri/pull/2696 lands.

DrMeepster:
Its been a while, sorry. Rebased and comments addressed.

RalfJung:
Just trying something here, please ignore.
@bors delegate+

PR #2671, opened 2022-11-16
remove unused `shell_escape` dep

klensy:


RalfJung:
Good catch, thanks!

@bors r+

PR #2673, opened 2022-11-17
Rustup

RalfJung:
This is unfortunately duplicating some Miri commits. When producing https://github.com/rust-lang/rust/pull/104456, josh did something odd in https://github.com/rust-lang/rust/commit/46d0df49d666c9ba1a1808a9adcca2cb6690d911 and added a rust-version file change. I have adjusted the way we update the rust-version file in `rustc-pull` and hope this will not happen again...


RalfJung:
Of course pushing this back to rustc fails again due to https://github.com/josh-project/josh/issues/998.

I really don't have the time to debug something every single pull/push... so I am beginning to wonder if josh might just not be ready yet for our needs. :/

RalfJung:
We might have to do a force-push to Miri again. :/ Basically, trying to get our current Miri master into rustc via josh, then re-fetching that and force-pushing Miri to that.

RalfJung:
I have identified the [source of the non-roundtrip](https://github.com/josh-project/josh/issues/998#issuecomment-1318391922). I don't think we can fix this without a force push, unfortunately.

https://github.com/rust-lang/miri/pull/2674 adjusts the script to
- not create the problematic kind of merge commit any more
- check that the created josh push round-trips, so we can avoid merging things into rustc that we cannot sync back to Miri

So at least next time this happens we should get a warning before things are in rustc history forever, which should let us avoid further force-pushes.

RalfJung:
This branch has identical content to https://github.com/rust-lang/miri/pull/2675, so we use it as CI proxy. (The other PR has conflicts so CI is not even started.)

PR #2675, opened 2022-11-17
Reset

RalfJung:
This is the Miri history as extracted from current rustc, with the PRs since the last push applied by hand. It is identical to https://github.com/rust-lang/miri/pull/2673. I think we should force-push Miri to this commit.


RalfJung:
@Mark-Simulacrum could I ask you again to temporarily disable branch protection in Miri? We ran into a josh bug and I don't know how to repair this without a Miri force-push. (The faulty commit landed in rustc with https://github.com/rust-lang/rust/pull/104456, but we can't force-push there obviously. This way we sync the fault back to Miri and therefore define it to be the correct history.)

I have updated our scripts so in the future we should be able to detect such faulty pushes before we merge them in rustc. I also confirmed that this commit does round-trip properly.

Cc @rust-lang/miri

RalfJung:
I think the oldest commit we need to get out of our history is https://github.com/rust-lang/miri/commit/218f60ec34004f9fd2d3de95d3d8ead28ad1ea68. I will verify by seeing where this history branches from master.

RalfJung:
Yes this branch does not contain https://github.com/rust-lang/miri/commit/218f60ec34004f9fd2d3de95d3d8ead28ad1ea68 but contains both of its parents. Good.

Mark-Simulacrum:
Allowed force push to users with push access.

RalfJung:
Hm, that does not seem to work...
```
remote: error: GH006: Protected branch update failed for refs/heads/master.
remote: error: You're not authorized to push to this branch. Visit https://docs.github.com/articles/about-protected-branches/ for more information.
To github.com:rust-lang/miri.git
 ! [remote rejected]   HEAD -> master (protected branch hook declined)
error: failed to push some refs to 'github.com:rust-lang/miri.git'
```

Mark-Simulacrum:
Try again now? I think there was another toggle I needed to hit

PR #2681, opened 2022-11-20
make miri-seed a regular integer, and also set layout-seed in many-seeds

RalfJung:
This makes the seed format consistent between `-Zlayout-seed` and `-Zmiri-seed`.


oli-obk (approved):

RalfJung:
I was about to ask. :)

PR #2684, opened 2022-11-20
Stack borrows: weak protectors

RalfJung:
This addresses the issue described in https://github.com/rust-lang/unsafe-code-guidelines/issues/376.

### DIFF
--- a/src/stacked_borrows/diagnostics.rs
+++ b/src/stacked_borrows/diagnostics.rs
@@ -288,7 +288,7 @@ impl<'span, 'history, 'ecx, 'mir, 'tcx> DiagnosticCx<'span, 'history, 'ecx, 'mir
             }
             Operation::Access(AccessOp { kind, range, .. }) =>
                 (*range, InvalidationCause::Access(*kind)),
-            _ => {
+            Operation::Dealloc(_) => {
                 // This can be reached, but never be relevant later since the entire allocation is
                 // gone now.
                 return;
--- a/src/stacked_borrows/diagnostics.rs
+++ b/src/stacked_borrows/diagnostics.rs
@@ -6,7 +6,7 @@ use rustc_span::{Span, SpanData};
 use rustc_target::abi::Size;
 
 use crate::helpers::CurrentSpan;
-use crate::stacked_borrows::{err_sb_ub, AccessKind, GlobalStateInner, Permission};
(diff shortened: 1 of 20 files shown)


RalfJung:
See https://github.com/rust-lang/unsafe-code-guidelines/pull/377 for a conceptual description of this change.

RalfJung:
At least the liballoc tests all this pass with this patch. I doubt there will be a lot of breakage out there, but I guess @saethlin will see. ;)

src/stacked_borrows/mod.rs
```diff
@@ -377,7 +401,7 @@ impl<'tcx> Stack {
                 0
             };
             self.pop_items_after(first_incompatible_idx, |item| {
-                Stack::item_popped(&item, global, dcx)?;
+                Stack::item_popped(&item, global, dcx, /* deallocation */ false)?;
```
saethlin, line 404: I think when you feel the need to put a comment next to a bool like this, a bool is no longer the best choice. Could this be two functions now? Or an enum, perhaps:
```rust
enum PopCause {
    Deallocation,
    Aliasing,
}
```
(I am not excited by these names)

src/stacked_borrows/diagnostics.rs
```diff
@@ -288,7 +288,11 @@ impl<'span, 'history, 'ecx, 'mir, 'tcx> DiagnosticCx<'span, 'history, 'ecx, 'mir
             }
             Operation::Access(AccessOp { kind, range, .. }) =>
                 (*range, InvalidationCause::Access(*kind)),
-            _ => unreachable!("Tags can only be invalidated during a retag or access"),
+            _ => {
+                // This can be reached, but never be relevant later since the entire allocation is
+                // gone now.
+                return;
```
saethlin, line 294: Without the runtime assertion, I would be happier with the variant we are ignoring listed explicitly so this is less fragile under refactoring.

src/stacked_borrows/mod.rs
RalfJung: Yeah, we have some of these and I'm just usually too lazy... I wish we had named function parameters.^^

saethlin:
I'm happy with the implementation. Thanks for cleaning up all the unnecessary mutability along the way.

PR #2685, opened 2022-11-21
Use `.wasm` extension when building for wasm in cargo-miri

Nilstrieb:
WASM uses the `.wasm` file extension for its binaries (just like how windows uses `.exe`), so we need to set that as well.

I'm not sure whether gating this behind the wasm target is a good idea, maybe it makes more sense to always do it just like on windows.

### DIFF
--- a/cargo-miri/src/phases.rs
+++ b/cargo-miri/src/phases.rs
@@ -281,9 +281,10 @@ pub fn phase_rustc(mut args: impl Iterator<Item = String>, phase: RustcPhase) {
             eprintln!("[cargo-miri rustc] writing run info to `{}`", filename.display());
         }
         info.store(&filename);
-        // For Windows, do the same thing again with `.exe` appended to the filename.
+        // For Windows and WASM, do the same thing again with `.exe`/`.wasm` appended to the filename.
         // (Need to do this here as cargo moves that "binary" to a different place before running it.)
         info.store(&out_filename("", ".exe"));
+        info.store(&out_filename("", ".wasm"));
     };
 
     let runnable_crate = !info_query && is_runnable_crate();
--- a/ci.sh
+++ b/ci.sh
@@ -103,6 +103,7 @@ case $HOST_TARGET in
     MIRI_TEST_TARGET=i686-pc-windows-msvc run_tests
     MIRI_TEST_TARGET=x86_64-unknown-freebsd run_tests_minimal hello integer vec panic/panic concurrency/simple atomic data_race env/var
     MIRI_TEST_TARGET=aarch64-linux-android run_tests_minimal hello integer vec panic/panic
+    MIRI_TEST_TARGET=wasm32-wasi MIRI_NO_STD=1 run_tests_minimal no_std # supports std but miri doesn't support it
     MIRI_TEST_TARGET=thumbv7em-none-eabihf MIRI_NO_STD=1 run_tests_minimal no_std # no_std embedded architecture
     ;;
   x86_64-apple-darwin)


cargo-miri/src/phases.rs
```diff
@@ -222,8 +222,8 @@ pub fn phase_rustc(mut args: impl Iterator<Item = String>, phase: RustcPhase) {
     /// never set for host crates. This matches what rustc bootstrap does,
     /// which hopefully makes it "reliable enough". This relies on us always
     /// invoking cargo itself with `--target`, which `in_cargo_miri` ensures.
-    fn is_target_crate() -> bool {
-        get_arg_flag_value("--target").is_some()
+    fn non_host_target_crate() -> Option<String> {
```
RalfJung, line 225: "target" and "host" are mutually exclusive, so I don't understand the point of the rename here.

Nilstrieb: Every build has a "target" in the sense that it targets some target. I wanted to make it clear that this is only available when `--target` is passed but I agree that the name isn't good.

cargo-miri/src/phases.rs
```diff
@@ -284,6 +284,11 @@ pub fn phase_rustc(mut args: impl Iterator<Item = String>, phase: RustcPhase) {
         // For Windows, do the same thing again with `.exe` appended to the filename.
         // (Need to do this here as cargo moves that "binary" to a different place before running it.)
         info.store(&out_filename("", ".exe"));
+
+        // cargo-miri never runs in wasm, so the --target should always be set when compiling for wasm.
+        if non_host_target_crate().map_or(false, |target| target.contains("wasm")) {
```
RalfJung, line 289: Let's keep this consistent... both here and below in the `emit_link_hack` section we currently just emit all the extensions; if we want to make that target-dependent we should do that for all extensions.

RalfJung:
Thanks!

We should also test this. Usually we just add new targets in `ci.sh` but I don't know if enough of wasm works for that to make sense?

cargo-miri/src/phases.rs
Nilstrieb: I removed the target check anyways so this doesn't need to be touched anymore, `is_target_crate` is a fine name when it returns a `bool`.

RalfJung: Note that `is_target_crate` is sometimes `true` even when target==host -- cargo has a distinction of host and target builds (these are my terms, not sure what they call it) that is different from whether target==host. It correlates with whether `--target` is passed to rustc.

ci.sh
```diff
@@ -85,7 +85,7 @@ case $HOST_TARGET in
     MIRI_TEST_TARGET=i686-pc-windows-msvc run_tests
     MIRI_TEST_TARGET=x86_64-unknown-freebsd run_tests_minimal hello integer vec panic/panic concurrency/simple atomic data_race env/var
     MIRI_TEST_TARGET=aarch64-linux-android run_tests_minimal hello integer vec panic/panic
-    MIRI_TEST_TARGET=wasm32-wasi run_tests
+    MIRI_TEST_TARGET=wasm32-wasi MIRI_NO_STD=1 run_tests_minimal # supports std but miri doesn't support it
```
RalfJung, line 88: `run_tests_minimal` expects as argument a filter to indicate which tests should be run. No filter means all tests.

However, `run_tests_minimal` also does not test `cargo miri`, so... it's not really helpful for this PR.

RalfJung:
Supporting wasi will take a while, that needs a bunch of wasi APIs to be implemented in Miri. We should probably find a way to land this without doing that.

I would suggest a 2-step process:
- in a first PR we make `run_tests_minimal` also test `cargo miri run` on some minimal no-std crate.
- then this PR can land with `MIRI_TEST_TARGET=wasm32-wasi run_tests_minimal no_std` (I don't think `MIRI_NO_STD` is needed, the sysroot can be built with std)

RalfJung:
https://github.com/rust-lang/miri/pull/2690 landed, so hopefully we can test cargo-miri on wasm in that way. :)

ci.sh
```diff
@@ -85,6 +85,7 @@ case $HOST_TARGET in
     MIRI_TEST_TARGET=i686-pc-windows-msvc run_tests
     MIRI_TEST_TARGET=x86_64-unknown-freebsd run_tests_minimal hello integer vec panic/panic concurrency/simple atomic data_race env/var
     MIRI_TEST_TARGET=aarch64-linux-android run_tests_minimal hello integer vec panic/panic
+    MIRI_TEST_TARGET=wasm32-wasi MIRI_NO_STD=1 run_tests_minimal # supports std but miri doesn't support it
```
RalfJung, line 88:
```suggestion
    MIRI_TEST_TARGET=wasm32-wasi MIRI_NO_STD=1 run_tests_minimal no_std # supports std but miri doesn't support it
```

Nilstrieb:
It works, CI passed!

RalfJung:
The bad news is that the smoke test isn't actually testing what it claims to test... see https://github.com/rust-lang/miri/compare/master...RalfJung:miri:smoke?expand=1.

RalfJung:
All right, let's see if they still pass.
@bors r+

PR #2690, opened 2022-11-24
Test a small cargo-miri smoke test even in `run_tests_minimal`

Nilstrieb:
This makes sure that cargo-miri works on all targets.

Implements the first step of https://github.com/rust-lang/miri/pull/2685#issuecomment-1325155967 to get that PR tested.


Nilstrieb:
libtest needs `std`, right. So uh, how exactly should I do this test crate? It should be `no_std`, but how should I do this? `no_std` binaries are tricky and platform specific and this should work on all platforms (and the ones that don't support `std` are the more funky ones usually :D).

RalfJung:
Yeah that's why I suggested a 'cargo run' smoketest, not 'cargo test'.
Or is that still too platform-specific?

Nilstrieb:
Getting a `no_std` binary running is still platform specific. I tried getting one to work (just on x86_64-unknown-linux-gnu) and after adding `panic = "abort"` and a `#[panic_handler]` it still needs an entry point and I didn't quite figure out what the best way to do that would be. Adding a `#[start]` function yielded linker errors about mising libc symbols (since libc isn't linked anymore) and I assume fixing this would require platform specific code to link the right things.

RalfJung:
This is Miri, no linking should be required...

We successfully do run https://github.com/rust-lang/miri/blob/master/tests/pass/no_std.rs on no_std targets in our test suite. Does that not also work with `cargo miri run`?

Nilstrieb:
Oh yes, this is Miri, I forgot about that :D

The example you have looks good, I think I can use that.

test-cargo-miri/Cargo.lock
```diff
@@ -81,6 +81,10 @@ version = "0.1.0"
 name = "issue_rust_86261"
 version = "0.1.0"
 
+[[package]]
+name = "no-std-smoke"
+version = "0.1.0"
```
RalfJung, line 86: If this is excluded, why is it listed in the lockfile...?

ci.sh
```diff
@@ -73,6 +73,10 @@ function run_tests_minimal {
   fi
 
   ./miri test -- "$@"
+
+  # Ensure that a small smoke test of cargo-miri works.
+  # Note: This doesn't work on windows because of TLS.
+  cargo miri run --manifest-path test-cargo-miri/no-std-smoke/Cargo.toml
```
RalfJung, line 79:
```suggestion
  echo "Runnig no-std cargo miri smoke test"
  cargo miri run --manifest-path test-cargo-miri/no-std-smoke/Cargo.toml
```

RalfJung: Though I guess with the `set -x` we don't really need this... never mind

Issue #2693, opened 2022-11-26
ICE when dealloc causes Stacked Borrows error


RalfJung:
I realized we don't have a test hitting [this](https://github.com/rust-lang/miri/blob/435faf0052ba0bd628bab5c531c74379a7462758/src/stacked_borrows/diagnostics.rs#L456), and indeed trying to do so currently ICEs:
```rust
use std::alloc::{alloc, dealloc, Layout};

fn main() { unsafe {
    let x = alloc(Layout::from_size_align_unchecked(1, 1));
    let ptr1 = (&mut *x) as *mut u8;
    let ptr2 = (&mut *ptr1) as *mut u8;
    // Invalidate ptr2 by writing to ptr1.
    ptr1.write(0);
    // Deallocate through ptr2.
    dealloc(ptr2, Layout::from_size_align_unchecked(1, 1));
} }
```
produces
```
thread '<unnamed>' panicked at 'internal error: entered unreachable code: access_error should only be called during an access', src/stacked_borrows/diagnostics.rs:398:13
```
The trouble is here:
https://github.com/rust-lang/miri/blob/2f92cf2d26c6b7e16715b900dc3d7d021f8afb40/src/stacked_borrows/mod.rs#L477-L480
We try to use `map_err` to make this a deallocation error, but when that happens `access` has already called `dcx.access_error`, causing the ICE. (The ICE is a regression introduced by https://github.com/rust-lang/miri/pull/2684.)

@saethlin any idea how to best fix this?

RalfJung:
I feel like `access_error` should just use the information present in `self.operation` to make a nice error.

saethlin:
`grant_error`, `access_error`, and `dealloc_error` only work with one kind of `Operation` each. `protector_error` however recognized 3 kinds of `Operation`. So if you can find a way to merge all 3 of `grant_error`, `access_error`, and `dealloc_error` together into a function with a decent name which branches on `Operation` to figure out what error text to emit, and when presented this program mentions a deallocation that sounds good to me. I'm not sure if that is workable though.

[...]

Issue #2648, opened 2022-11-05
Stacked Borrows retags should also be accesses for the data race model


RalfJung:
Currently, many Stacked Borrows retags are effectively read/write accesses for the aliasing model:

https://github.com/rust-lang/miri/blob/9568d7e440685584868cef516149123d43c45178/src/stacked_borrows/mod.rs#L508-L511

However, the data race model never hears about them, which as @JakobDegen [points out](https://rust-lang.zulipchat.com/#narrow/stream/136281-t-lang.2Fwg-unsafe-code-guidelines/topic/SB.20Retags.20interactions.20with.20C.2B.2B.20atomics.20model) leads us to accept some code that we probably should not accept.

PR #2694, opened 2022-11-26
fix handling of spurious accesses during retag

RalfJung:
The `dereferenceable` attribute we emit for LLVM is checked during retag in Stacked Borrows.
However, we currently don't properly do that for retagging of `&mut !Unpin`, which this PR fixes.
Also this adjusts retagging to inform the data race model of the accesses as well.

Fixes https://github.com/rust-lang/miri/issues/2648.
Also fixes https://github.com/rust-lang/miri/issues/2693 since the same issue arose for retagging as well.

r? @saethlin

### DIFF
--- a/src/stacked_borrows/diagnostics.rs
+++ b/src/stacked_borrows/diagnostics.rs
@@ -353,10 +353,12 @@ impl<'history, 'ecx, 'mir, 'tcx> DiagnosticCx<'history, 'ecx, 'mir, 'tcx> {
 
     /// Report a descriptive error when `new` could not be granted from `derived_from`.
     #[inline(never)] // This is only called on fatal code paths
-    pub(super) fn grant_error(&self, perm: Permission, stack: &Stack) -> InterpError<'tcx> {
+    pub(super) fn grant_error(&self, stack: &Stack) -> InterpError<'tcx> {
         let Operation::Retag(op) = &self.operation else {
             unreachable!("grant_error should only be called during a retag")
         };
+        let perm =
+            op.permission.expect("`start_grant` must be called before calling `grant_error`");
         let action = format!(
             "trying to retag from {:?} for {:?} permission at {:?}[{:#x}]",
             op.orig_tag,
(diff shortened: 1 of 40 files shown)


src/stacked_borrows/diagnostics.rs
```diff
@@ -447,14 +452,16 @@ impl<'span, 'history, 'ecx, 'mir, 'tcx> DiagnosticCx<'span, 'history, 'ecx, 'mir
...
-    pub fn dealloc_error(&self) -> InterpError<'tcx> {
+    pub fn dealloc_error(&self, stack: &Stack) -> InterpError<'tcx> {
         let Operation::Dealloc(op) = &self.operation else {
             unreachable!("dealloc_error should only be called during a deallocation")
         };
         err_sb_ub(
             format!(
-                "no item granting write access for deallocation to tag {:?} at {:?} found in borrow stack",
-                op.tag, self.history.id,
+                "attemtping deallocation using {tag:?} at {alloc_id:?}{cause}",
```
RalfJung, line 461: There is a small inconsistency here in the error message, when compared with accesses... accesses say `attempting a read/write access`, but here we leave away the `a`. IMO that sounds better. In fact `attempting read/write access` also works, doesn't it? Should we switch to that?

RalfJung:
While I was touching retagging anyway, I also added a fix for https://github.com/rust-lang/miri/issues/2648.

src/stacked_borrows/diagnostics.rs
saethlin: I think this is that this is an inconsistency in English, at least as it is spoken. I follow your logic, but the words sound wrong to me.

src/stacked_borrows/diagnostics.rs
```diff
@@ -428,14 +433,16 @@ impl<'history, 'ecx, 'mir, 'tcx> DiagnosticCx<'history, 'ecx, 'mir, 'tcx> {
...
```
saethlin, line 442:
```suggestion
                "attempting deallocation using {tag:?} at {alloc_id:?}{cause}",
```

src/stacked_borrows/mod.rs
```diff
@@ -914,8 +909,17 @@ trait EvalContextPrivExt<'mir: 'ecx, 'tcx: 'mir, 'ecx>: crate::MiriInterpCxExt<'
                         alloc_range(base_offset, size),
                     );
                     stacked_borrows.for_each(range, dcx, |stack, dcx, exposed_tags| {
-                        stack.grant(orig_tag, item, &global, dcx, exposed_tags)
-                    })
+                        stack.grant(orig_tag, item, access, &global, dcx, exposed_tags)
+                    })?;
+                    drop(global);
+                    if let Some(access) = access {
+                        assert!(access == AccessKind::Read);
```
saethlin, line 916: If I recall correctly, `assert_eq!` produces better panic messages, so I always use it when possible

saethlin:
I skimmed over this but I'm a bit fried right now, I'll look over it for real tomorrow.

src/stacked_borrows/diagnostics.rs
RalfJung: :shrug:

src/stacked_borrows/mod.rs
RalfJung: It prints what the values actually were. But here there are only two options.

But yeah probably better to get in the habit of using assert_eq.

saethlin:
I think I understand this decently well now. r=me whenever you want to merge (we have a number of outstanding PRs which may be disruptive so I'm not going to just send this)

RalfJung:
We'll have to land them in some order... :shrug: 
@bors r=saethlin

PR #2697, opened 2022-11-27
Reorganizing `stacked_borrows` in anticipation of a different model

Vanille-N:
These commits reorganize all the code inside the former `stacked_borrows` module and extract the part that really is specific to Stacked Borrows inside the `borrow_tracker/stacked_borrows` submodule.
Everything not specific to SB is put in `borrow_tracker/mod.rs`.
This is so that the future Tree Borrows model can be later added as a second submodule and reuse all the contents of `borrow_tracker/mod.rs`.

This reorganization is accompanied by renamings, mostly from "stacked borrows" to "borrow tracking".

### DIFF
--- a/src/borrow_tracker/mod.rs
+++ b/src/borrow_tracker/mod.rs
@@ -1,0 +1,365 @@
+use std::cell::RefCell;
+use std::fmt;
+use std::num::NonZeroU64;
+
+use log::trace;
+use smallvec::SmallVec;
+
+use rustc_data_structures::fx::{FxHashMap, FxHashSet};
+use rustc_middle::mir::RetagKind;
+use rustc_target::abi::Size;
+
+use crate::*;
+pub mod stacked_borrows;
+use stacked_borrows::diagnostics::RetagCause;
+
+pub type CallId = NonZeroU64;
+
+/// Tracking pointer provenance
+#[derive(Copy, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct BorTag(NonZeroU64);
+
+impl BorTag {
+    pub fn new(i: u64) -> Option<Self> {
+        NonZeroU64::new(i).map(BorTag)
+    }
+
+    pub fn get(&self) -> u64 {
+        self.0.get()
+    }
+
+    pub fn inner(&self) -> NonZeroU64 {
+        self.0
+    }
+
+    pub fn succ(self) -> Option<Self> {
+        self.0.checked_add(1).map(Self)
+    }
+
+    /// The minimum representable tag
+    pub fn one() -> Self {
+        Self::new(1).unwrap()
+    }
+}
+
+impl std::default::Default for BorTag {
+    /// The default to be used when borrow tracking is disabled
+    fn default() -> Self {
+        Self::one()
+    }
+}
+
+impl fmt::Debug for BorTag {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "<{}>", self.0)
+    }
+}
+
+/// Per-frame data for borrow tracking
+#[derive(Debug)]
+pub struct FrameExtra {
+    /// The ID of the call this frame corresponds to.
+    pub call_id: CallId,
+
+    /// If this frame is protecting any tags, they are listed here. We use this list to do
+    /// incremental updates of the global list of protected tags stored in the
+    /// `stacked_borrows::GlobalState` upon function return, and if we attempt to pop a protected
+    /// tag, to identify which call is responsible for protecting the tag.
+    /// See `Stack::item_popped` for more explanation.
+    ///
+    /// This will contain one tag per reference passed to the function, so
+    /// a size of 2 is enough for the vast majority of functions.
+    pub protected_tags: SmallVec<[BorTag; 2]>,
+}
+
+impl VisitTags for FrameExtra {
+    fn visit_tags(&self, _visit: &mut dyn FnMut(BorTag)) {
+        // `protected_tags` are fine to GC.
+    }
+}
+
+/// Extra global state, available to the memory access hooks.
+#[derive(Debug)]
+pub struct GlobalStateInner {
+    /// Borrow tracker method currently in use.
+    pub borrow_tracker_method: BorrowTrackerMethod,
+    /// Next unused pointer ID (tag).
+    pub next_ptr_tag: BorTag,
+    /// Table storing the "base" tag for each allocation.
+    /// The base tag is the one used for the initial pointer.
+    /// We need this in a separate table to handle cyclic statics.
+    pub base_ptr_tags: FxHashMap<AllocId, BorTag>,
+    /// Next unused call ID (for protectors).
+    pub next_call_id: CallId,
+    /// All currently protected tags.
+    /// An item is protected if its tag is in this set, *and* it has the "protected" bit set.
+    /// We add tags to this when they are created with a protector in `reborrow`, and
+    /// we remove tags from this when the call which is protecting them returns, in
+    /// `GlobalStateInner::end_call`. See `Stack::item_popped` for more details.
+    pub protected_tags: FxHashMap<BorTag, ProtectorKind>,
+    /// The pointer ids to trace
+    pub tracked_pointer_tags: FxHashSet<BorTag>,
+    /// The call ids to trace
+    pub tracked_call_ids: FxHashSet<CallId>,
+    /// Whether to recurse into datatypes when searching for pointers to retag.
+    pub retag_fields: RetagFields,
+}
+
+impl VisitTags for GlobalStateInner {
+    fn visit_tags(&self, _visit: &mut dyn FnMut(BorTag)) {
+        // The only candidate is base_ptr_tags, and that does not need visiting since we don't ever
+        // GC the bottommost tag.
+    }
+}
+
+/// We need interior mutable access to the global state.
+pub type GlobalState = RefCell<GlobalStateInner>;
+
+/// Indicates which kind of access is being performed.
+#[derive(Copy, Clone, Hash, PartialEq, Eq, Debug)]
+pub enum AccessKind {
+    Read,
+    Write,
+}
+
+impl fmt::Display for AccessKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            AccessKind::Read => write!(f, "read access"),
+            AccessKind::Write => write!(f, "write access"),
+        }
+    }
+}
+
+/// Policy on whether to recurse into fields to retag
+#[derive(Copy, Clone, Debug)]
+pub enum RetagFields {
+    /// Don't retag any fields.
+    No,
+    /// Retag all fields.
+    Yes,
+    /// Only retag fields of types with Scalar and ScalarPair layout,
+    /// to match the LLVM `noalias` we generate.
+    OnlyScalar,
+}
+
+/// The flavor of the protector.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum ProtectorKind {
+    /// Protected against aliasing violations from other pointers.
+    ///
+    /// Items protected like this cause UB when they are invalidated, *but* the pointer itself may
+    /// still be used to issue a deallocation.
+    ///
+    /// This is required for LLVM IR pointers that are `noalias` but *not* `dereferenceable`.
+    WeakProtector,
+
+    /// Protected against any kind of invalidation.
+    ///
+    /// Items protected like this cause UB when they are invalidated or the memory is deallocated.
+    /// This is strictly stronger protection than `WeakProtector`.
+    ///
+    /// This is required for LLVM IR pointers that are `dereferenceable` (and also allows `noalias`).
+    StrongProtector,
+}
+
+/// Utilities for initialization and ID generation
+impl GlobalStateInner {
+    pub fn new(
+        borrow_tracker_method: BorrowTrackerMethod,
+        tracked_pointer_tags: FxHashSet<BorTag>,
+        tracked_call_ids: FxHashSet<CallId>,
+        retag_fields: RetagFields,
+    ) -> Self {
+        GlobalStateInner {
+            borrow_tracker_method,
+            next_ptr_tag: BorTag::one(),
+            base_ptr_tags: FxHashMap::default(),
+            next_call_id: NonZeroU64::new(1).unwrap(),
+            protected_tags: FxHashMap::default(),
+            tracked_pointer_tags,
+            tracked_call_ids,
+            retag_fields,
+        }
+    }
+
+    /// Generates a new pointer tag. Remember to also check track_pointer_tags and log its creation!
+    pub fn new_ptr(&mut self) -> BorTag {
+        let id = self.next_ptr_tag;
+        self.next_ptr_tag = id.succ().unwrap();
+        id
+    }
+
+    pub fn new_frame(&mut self, machine: &MiriMachine<'_, '_>) -> FrameExtra {
+        let call_id = self.next_call_id;
+        trace!("new_frame: Assigning call ID {}", call_id);
+        if self.tracked_call_ids.contains(&call_id) {
+            machine.emit_diagnostic(NonHaltingDiagnostic::CreatedCallId(call_id));
+        }
+        self.next_call_id = NonZeroU64::new(call_id.get() + 1).unwrap();
+        FrameExtra { call_id, protected_tags: SmallVec::new() }
+    }
+
+    pub fn end_call(&mut self, frame: &machine::FrameData<'_>) {
+        for tag in &frame
+            .borrow_tracker
+            .as_ref()
+            .expect("we should have borrow tracking data")
+            .protected_tags
+        {
+            self.protected_tags.remove(tag);
+        }
+    }
+
+    pub fn base_ptr_tag(&mut self, id: AllocId, machine: &MiriMachine<'_, '_>) -> BorTag {
+        self.base_ptr_tags.get(&id).copied().unwrap_or_else(|| {
+            let tag = self.new_ptr();
+            if self.tracked_pointer_tags.contains(&tag) {
+                machine.emit_diagnostic(NonHaltingDiagnostic::CreatedPointerTag(
+                    tag.inner(),
+                    None,
+                    None,
+                ));
+            }
+            trace!("New allocation {:?} has base tag {:?}", id, tag);
+            self.base_ptr_tags.try_insert(id, tag).unwrap();
+            tag
+        })
+    }
+}
+
+/// Which borrow tracking method to use
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum BorrowTrackerMethod {
+    /// Stacked Borrows, as implemented in borrow_tracker/stacked
+    StackedBorrows,
+}
+
+impl BorrowTrackerMethod {
+    pub fn instanciate_global_state(self, config: &MiriConfig) -> GlobalState {
+        RefCell::new(GlobalStateInner::new(
+            self,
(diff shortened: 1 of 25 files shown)


RalfJung:
Please don't include any merge commits in PRs. If there is a conflict, do a rebase instead.

src/borrow_tracker/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+        match self {
+            State::Stack(sb) => sb.borrow_mut().remove_unreachable_tags(tags),
+        }
+    }
+
+    pub fn expose_tag(&self, tag: BorTag) {
+        match self {
+            State::Stack(sb) => sb.borrow_mut().expose_tag(tag),
+        }
+    }
```
saethlin, line 331: The fact that these take `&self` is surprising. I think it'll make more sense now or later to hoist the `RefCell` out to `State` instead of sinking it into both the borrow tracking implementations.

src/borrow_tracker/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+use std::fmt;
+use std::num::NonZeroU64;
+
+use log::trace;
+use smallvec::SmallVec;
+
+use rustc_data_structures::fx::{FxHashMap, FxHashSet};
+use stacked::diagnostics::RetagCause;
+
+use crate::*;
```
RalfJung, line 13:
```suggestion

use stacked::diagnostics::RetagCause;
use crate::*;
```
The organization is
- std imports
- other dependency imports
- rustc imports
- local crate imports

with an empty line between each block.

src/borrow_tracker/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+    pub fn inner(&self) -> NonZeroU64 {
+        self.0
+    }
+
+    pub fn succ(self) -> Option<Self> {
+        self.try_map(|n| n.checked_add(1))
+    }
+
+    // The default to be used when SB is disabled
+    pub fn one() -> Self {
```
RalfJung, line 39:
```suggestion
    pub fn default() -> Self {
```
No need to expose its value.

src/borrow_tracker/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+        Self::new(1).unwrap()
+    }
+
+    pub fn try_map<F>(self, f: F) -> Option<Self>
```
RalfJung, line 43: What is this for? I think all you need is an increment function; if that is true, please just implement that directly.

src/borrow_tracker/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+            }
+            trace!("New allocation {:?} has base tag {:?}", id, tag);
+            self.base_ptr_tags.try_insert(id, tag).unwrap();
+            tag
+        })
+    }
+}
+
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub enum Method {
```
RalfJung, line 225: This is missing a doc comment.

src/borrow_tracker/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+                )))),
+            Method::Off => None,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum State {
+    Stack(stacked::AllocExtra),
+}
```
RalfJung, line 270: This is missing a doc comment. it is also a confusing name... we have GlobalState and State?

What about calling this AllocState or so? Should then probably rename `stacked::AllocExtra` to `stacked::AllocState` for consistency.

src/borrow_tracker/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+
+impl State {
+    pub fn assume_sb(&self) -> &stacked::AllocExtra {
```
RalfJung, line 273: `assert_sb` I think would be more in line with how we usually call these -- I assume this will panic if the state is not sb.

src/borrow_tracker/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+                    machine,
+                )))),
+            Method::Off => None,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum State {
+    Stack(stacked::AllocExtra),
```
RalfJung, line 269: Why does this not have a `None` variant corresponding to `Method::Off`?

src/borrow_tracker/stacked/mod.rs
```diff
@@ -4,80 +4,34 @@
...
+    stacked::diagnostics::{AllocHistory, DiagnosticCx, DiagnosticCxBuilder, TagHistory},
+    AccessKind, GlobalState, GlobalStateInner, ProtectorKind, RetagCause, RetagFields,
+};
+use rustc_middle::{
+    mir::{Mutability, RetagKind},
+    ty::{
+        self,
+        layout::{HasParamEnv, LayoutOf},
+    },
+};
```
RalfJung, line 23: Please keep the import structure, and move this up to the other rustc imports.

src/borrow_tracker/mod.rs
RalfJung: We'll want the `RefCell` to be inside the `Option` though, so if we add a `None` variant to `State` then we wouldn't want a `RefCell` around that.

RalfJung: Actually this could just take `&mut self` I think, the only caller has a mutable reference available.

But I agree that since this file here is doing the `borrow_mut` calls, this file should also be adding the `RefCell`. We are already assuming that `stacked::AllocExtra` is a `RefCell`. So might as well define this more like
```rust
pub enum AllocState {
    Stack(RefCell<stacked::AllocState>),
    None,
}
```

src/borrow_tracker/stacked/mod.rs
```diff
@@ -929,14 +754,16 @@ trait EvalContextPrivExt<'mir: 'ecx, 'tcx: 'mir, 'ecx>: crate::MiriInterpCxExt<'
...
             .as_mut()
             .expect("we should have Stacked Borrows data")
-            .get_mut();
+            .assume_sb()
+            .borrow_mut();
         let item = Item::new(new_tag, perm, protect.is_some());
         let range = alloc_range(base_offset, size);
-        let global = machine.stacked_borrows.as_ref().unwrap().borrow();
+        let global = machine.borrow_tracker_data.as_ref().unwrap().borrow();
+        // FIXME: can't share this with the current_span inside log_creation
```
RalfJung, line 766: You are reverting some recent changes in master here. Please be careful when rebasing!

src/borrow_tracker/stacked/mod.rs
```diff
@@ -961,7 +788,8 @@ trait EvalContextPrivExt<'mir: 'ecx, 'tcx: 'mir, 'ecx>: crate::MiriInterpCxExt<'
 
     /// Retags an indidual pointer, returning the retagged version.
     /// `mutbl` can be `None` to make this a raw pointer.
-    fn retag_reference(
+    // FIXME: is this comment outdated ?
```
RalfJung, line 791:
```suggestion
```

src/borrow_tracker/stacked/mod.rs
```diff
@@ -961,7 +788,8 @@ trait EvalContextPrivExt<'mir: 'ecx, 'tcx: 'mir, 'ecx>: crate::MiriInterpCxExt<'
 
     /// Retags an indidual pointer, returning the retagged version.
     /// `mutbl` can be `None` to make this a raw pointer.
```
RalfJung, line 790:
```suggestion
    /// Retags an indidual pointer, returning the retagged version.
    /// `kind` indicates what kind of reference is being created.
```

src/borrow_tracker/stacked/mod.rs
```diff
@@ -1181,7 +1014,7 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 // uncovers a non-supported `extern static`.
                 let alloc_extra = this.get_alloc_extra(alloc_id)?;
                 trace!("Stacked Borrows tag {tag:?} exposed in {alloc_id:?}");
-                alloc_extra.stacked_borrows.as_ref().unwrap().borrow_mut().exposed_tags.insert(tag);
+                alloc_extra.borrow_tracker.as_ref().unwrap().expose_tag(tag);
```
RalfJung, line 1017: Why does Stacked Borrows call back into the generic borrow tracking machinery here? This should call `assume_sb`.

src/borrow_tracker/stacked/mod.rs
```diff
@@ -775,12 +598,13 @@ trait EvalContextPrivExt<'mir: 'ecx, 'tcx: 'mir, 'ecx>: crate::MiriInterpCxExt<'
                         orig_tag,
                         alloc_range(base_offset, size),
                     );
-                    let mut dcx = dcx.build(&mut stacked_borrows.history, base_offset);
+                    let mut stacked_borrows = sb.borrow_mut();
+                    let mut dcx = dcx.build(stacked_borrows.history_mut(), base_offset);
```
RalfJung, line 602: Why did you add a new method for this?

src/machine.rs
```diff
@@ -351,7 +351,8 @@ pub struct MiriMachine<'mir, 'tcx> {
     pub tcx: TyCtxt<'tcx>,
 
     /// Stacked Borrows global data.
-    pub stacked_borrows: Option<stacked_borrows::GlobalState>,
+    pub borrow_tracker_method: BorrowTrackerMethod,
+    pub borrow_tracker_data: Option<borrow_tracker::GlobalState>,
```
RalfJung, line 355: `borrow_tracker_method` should not be needed, since it is entirely determined by `borrow_tracker_data`. Then you can rename that field to `borrow_tracker`.

src/machine.rs
```diff
@@ -477,17 +478,13 @@ impl<'mir, 'tcx> MiriMachine<'mir, 'tcx> {
...
         let rng = StdRng::seed_from_u64(config.seed.unwrap_or(0));
-        let stacked_borrows = config.stacked_borrows.then(|| {
-            RefCell::new(stacked_borrows::GlobalStateInner::new(
-                config.tracked_pointer_tags.clone(),
-                config.tracked_call_ids.clone(),
-                config.retag_fields,
-            ))
-        });
+        let borrow_tracker_method = config.borrow_tracker;
+        let borrow_tracker_data = borrow_tracker_method.instanciate_global_state(config);
```
RalfJung, line 482:
```suggestion
        let borrow_tracker = config.borrow_tracker.instantiate_global_state(config);
```

src/machine.rs
```diff
@@ -900,9 +898,13 @@ impl<'mir, 'tcx> Machine<'mir, 'tcx> for MiriMachine<'mir, 'tcx> {
         }
 
         let alloc = alloc.into_owned();
-        let stacks = ecx.machine.stacked_borrows.as_ref().map(|stacked_borrows| {
-            Stacks::new_allocation(id, alloc.size(), stacked_borrows, kind, &ecx.machine)
-        });
+        let borrow_tracker = ecx.machine.borrow_tracker_method.new_allocation(
```
RalfJung, line 901: This can probably be `ecx.machine.borrow_tracker.method().new_allocation`.

src/machine.rs
```diff
@@ -1068,7 +1059,10 @@ impl<'mir, 'tcx> Machine<'mir, 'tcx> for MiriMachine<'mir, 'tcx> {
         kind: mir::RetagKind,
         place: &PlaceTy<'tcx, Provenance>,
     ) -> InterpResult<'tcx> {
-        if ecx.machine.stacked_borrows.is_some() { ecx.retag(kind, place) } else { Ok(()) }
+        match ecx.machine.borrow_tracker_method {
+            BorrowTrackerMethod::StackedBorrows => ecx.sb_retag(kind, place),
+            BorrowTrackerMethod::Off => Ok(()),
+        }
```
RalfJung, line 1065: Please move this `match` into `borrow_tracker/mod.rs`.
The machine file should never match on the `BorrowTrackerMethod`.

src/machine.rs
```diff
@@ -1091,10 +1085,10 @@ impl<'mir, 'tcx> Machine<'mir, 'tcx> for MiriMachine<'mir, 'tcx> {
             None
         };
 
-        let stacked_borrows = ecx.machine.stacked_borrows.as_ref();
+        let borrow_tracker = ecx.machine.borrow_tracker_data.as_ref();
 
         let extra = FrameData {
-            stacked_borrows: stacked_borrows.map(|sb| sb.borrow_mut().new_frame(&ecx.machine)),
+            borrow_tracker: borrow_tracker.map(|sb| sb.borrow_mut().new_frame(&ecx.machine)),
```
RalfJung, line 1091:
```suggestion
            borrow_tracker: borrow_tracker.map(|bt| bt.borrow_mut().new_frame(&ecx.machine)),
```

src/machine.rs
```diff
@@ -1154,7 +1148,11 @@ impl<'mir, 'tcx> Machine<'mir, 'tcx> for MiriMachine<'mir, 'tcx> {
             ecx.active_thread_mut().set_top_user_relevant_frame(stack_len - 1);
         }
 
-        if ecx.machine.stacked_borrows.is_some() { ecx.retag_return_place() } else { Ok(()) }
+        match ecx.machine.borrow_tracker_method {
```
RalfJung, line 1151: Another `match` to move into the borrow tracker.

src/machine.rs
```diff
@@ -1154,7 +1148,11 @@ impl<'mir, 'tcx> Machine<'mir, 'tcx> for MiriMachine<'mir, 'tcx> {
...
+            BorrowTrackerMethod::StackedBorrows => ecx.sb_retag_return_place(),
+            // FIXME: don't forget to also activate the args after entry
```
RalfJung, line 1153: I don't quite understand this FIXME?

src/borrow_tracker/mod.rs
Vanille-N: Or I could call it `AllocExtra` as it dispatches into each implementation's `AllocExtra` ? Or would that be too confusing ?

src/borrow_tracker/stacked/mod.rs
Vanille-N: `get_mut` doesn't work with `assert_sb` though. Or is that not the issue ?

src/borrow_tracker/mod.rs
Vanille-N: Ok I'll do that

src/borrow_tracker/mod.rs
RalfJung: We also call it `GlobalState`, not `GlobalExtra`. So I think we should rename the implementation's thing to `AllocState`, too.

src/borrow_tracker/stacked/mod.rs
RalfJung: Ah, then you need `assert_sb_mut`.

RalfJung:
@rustbot author

(call `rustbot ready` when this is ready for another round of review)

RalfJung (commented):
There are also still some unresolved comments from the previous round. (Github might hide them for you, be sure to click "load more" until that button disappears.)

src/borrow_tracker/stacked/mod.rs
```diff
@@ -1181,7 +989,7 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 // uncovers a non-supported `extern static`.
                 let alloc_extra = this.get_alloc_extra(alloc_id)?;
                 trace!("Stacked Borrows tag {tag:?} exposed in {alloc_id:?}");
-                alloc_extra.stacked_borrows.as_ref().unwrap().borrow_mut().exposed_tags.insert(tag);
+                alloc_extra.borrow_tracker.expose_tag(tag);
```
RalfJung, line 992: As I already said last round:
Why does Stacked Borrows call back into the generic borrow tracking machinery here? This should call assume_sb. Then you can probably remove the new expose_tag method on the per-alloc state.

src/intptrcast.rs
```diff
@@ -105,15 +105,16 @@ impl<'mir, 'tcx> GlobalStateInner {
...
         let global_state = ecx.machine.intptrcast.get_mut();
         // In strict mode, we don't need this, so we can save some cycles by not tracking it.
         if global_state.provenance_mode != ProvenanceMode::Strict {
             trace!("Exposing allocation id {alloc_id:?}");
             global_state.exposed.insert(alloc_id);
-            if ecx.machine.stacked_borrows.is_some() {
-                ecx.expose_tag(alloc_id, sb)?;
+            match ecx.machine.borrow_tracker_method {
+                BorrowTrackerMethod::StackedBorrows => ecx.sb_expose_tag(alloc_id, tag)?,
+                BorrowTrackerMethod::Off => (),
```
RalfJung, line 117: As I said last time, there shouldn't be any `match` on this outside the borrow_tracker module.

src/machine.rs
RalfJung: Oh I see, the global borrow_tracker_data is the same no matter the borrow tracking module.

But still, we should only have one field for the borrow tracker here. `new_allocation` should become a function on `borrow_tracker::GlobalState` which then has to internally remember whatever is needed.

src/intptrcast.rs
Vanille-N: Oops sorry I missed that one

src/borrow_tracker/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+impl VisitTags for FrameExtra {
+    fn visit_tags(&self, _visit: &mut dyn FnMut(BorTag)) {
+        // `protected_tags` are fine to GC.
+    }
+}
+
+/// Extra global state, available to the memory access hooks.
+#[derive(Debug)]
+pub struct GlobalStateInner {
+    /// Borrow tracker method currently in use (except BorrowTrackerMethod::Off)
```
RalfJung, line 84:
```suggestion
    /// Borrow tracker method currently in use.
```

src/borrow_tracker/mod.rs
```diff
@@ -274,20 +274,20 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         &mut self,
         kind: RetagKind,
         place: &PlaceTy<'tcx, Provenance>,
-        borrow_tracker_method: BorrowTrackerMethod,
     ) -> InterpResult<'tcx> {
         let this = self.eval_context_mut();
-        match borrow_tracker_method {
+        let method = this.machine.borrow_tracker.as_ref().unwrap().borrow().method();
```
RalfJung, line 275: Why do you call `method()` here? You should have access to the private field, it's in the same module. (Same below a few more times. Is `fn method` even needed?)

src/borrow_tracker/stacked_borrows/mod.rs
```diff
@@ -693,12 +494,16 @@ impl Stacks {
...
         let dcx = DiagnosticCxBuilder::dealloc(machine, tag);
-        let state = machine.stacked_borrows.as_ref().unwrap().borrow();
+        let state = machine.borrow_tracker.as_ref().unwrap().borrow();
         self.for_each(range, dcx, |stack, dcx, exposed_tags| {
             stack.dealloc(tag, &state, dcx, exposed_tags)
         })?;
         Ok(())
     }
+
+    pub fn expose_tag(&mut self, tag: BorTag) {
```
RalfJung, line 504:
```suggestion
    fn expose_tag(&mut self, tag: BorTag) {
```
This is only called in this file, right?

RalfJung:
Looks good. :) Just a few final nits. Then please squash this into one (or a few) commits and we are good to go!

RalfJung:
(force-pushes do not always generate GH notifications, so make sure to always post a note in the PR if you want another review)

RalfJung:
Thanks!

@bors r+

PR #2698, opened 2022-11-27
replace 'locally built rustc' instructions by 'Miri in rustc'

RalfJung:


CONTRIBUTING.md
```diff
@@ -203,65 +203,32 @@ for more information about configuring VS Code and `rust-analyzer`.
...
-rustup override set stage2
 ```
-
-Note: When you are working with a locally built rustc or any other toolchain that
-is not the same as the one in `rust-version`, you should not have `.auto-everything` or
-`.auto-toolchain` as that will keep resetting your toolchain.
-
-```sh
-rm -f .auto-everything .auto-toolchain
+./x.py test miri --stage 0
```
saethlin, line 222: FYI this isn't mentioned in `./x.py test -h -v`, it only lists `src/tools/miri`. Not sure if it is supposed to include just `miri` in that list.

RalfJung: I only fairly recently learned about these shortcuts and honestly have no idea why they work... but they are hand.y^^
@jyn514 is that something we can rely on even if it's not listed in `--help`?

jyn514: Yes. x.py supports all suffixes, they don't have to be an exact match: https://github.com/rust-lang/rust/blob/454784afba5bf35b5ff14ada0e31265ad1d75e73/src/bootstrap/builder.rs#L193

PR #2699, opened 2022-11-27
refactor scheduler

RalfJung:
Refactors the scheduler to use something akin to a generator -- a callback that will be invoked when the stack of a thread is empty, which has the chance to push a new stack frame or do other things and then indicates whether this thread is done, or should be scheduled again. (Unfortunately I think we [cannot use actual generators](https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/Generators.20that.20borrow.20on.20each.20resume.3F) here.) The interpreter loop is now a proper infinite loop, the only way to leave it is for some kind of interrupt to be triggered (represented as `InterpError`) -- unifying how we handle 'exit when calling `process::exit`' and 'exit when main thread quits'.

The last commit implements an alternative approach to https://github.com/rust-lang/miri/pull/2660 using this new structure. Fixes https://github.com/rust-lang/miri/issues/2629.

### DIFF
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -4,6 +4,7 @@ use std::ffi::{OsStr, OsString};
 use std::iter;
 use std::panic::{self, AssertUnwindSafe};
 use std::path::PathBuf;
+use std::task::Poll;
 use std::thread;
 
 use log::info;
@@ -20,6 +21,7 @@ use rustc_target::spec::abi::Abi;
 
 use rustc_session::config::EntryFnType;
 
+use crate::shims::tls;
 use crate::*;
 
 #[derive(Copy, Clone, Debug, PartialEq)]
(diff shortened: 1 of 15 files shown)


tests/pass/concurrency/scope.rs
```diff
@@ -0,0 +1,24 @@
...
+            // because no other threads are using it.
+            x += a[0] + a[2];
+        });
+        let _s = format!("hello from the main thread");
+    });
+
+    // After the scope, we can modify and access our variables again:
+    a.push(4);
+    assert_eq!(x, a.len());
+}
```
saethlin, line 24: Is this test safeguarding against anything in particular?

RalfJung: No, I just realized we didn't have any test that calls scoped threads in our test suite here, and that seemed like something we should fix. And since scoped threads are their own sufficiently complicated thing, it felt better to make it a separate test (potentially to be extended with more ways of invoking scoped threads in the future).

I *think* this is a test that triggered https://github.com/rust-lang/miri/issues/2629 for me in the past, but I am not sure and anyway we don't have a good way of testing for https://github.com/rust-lang/miri/issues/2629.

saethlin: Not sure if I proposed this before... but we could use `./miri many-seeds` in CI? That's not proof against some strange scheduler situation, but it definitely hits a lot of code paths.

RalfJung: We'd have to figure out which tests to run it on, though. 256 seeds takes quite a while so we can't do this for many tests.

saethlin: I would be happy with just this:
```rust
use std::thread;

fn main() {
  thread::scope(|s| {
    s.spawn(|| {});
  });
}
```

src/eval.rs
```diff
@@ -172,17 +179,75 @@ impl Default for MiriConfig {
...
+                match state.on_stack_empty(this)? {
+                    Poll::Pending => {} // just keep going
+                    Poll::Ready(()) => {
+                        // Give background threads a chance to finish by yielding the main thread a
+                        // couple of times -- but only if we would also preempt threads randomly.
+                        if this.machine.preemption_rate > 0.0 {
+                            *self = Yield { remaining: MAIN_THREAD_YIELDS_AT_SHUTDOWN };
+                        } else {
+                            *self = Done;
+                        }
```
saethlin, line 214: This is still the yield-based impl, not the step-based idea. I wouldn't mind saving the step-based impl for a new PR, but I just want to make sure: Did you use yielding here because the step-based idea is too hard now? Or just because the yield strategy is what was in my branch?

RalfJung: I think yielding is sufficient actually -- note how I only yield if preemption is enabled. So (a) there's anyway a chance that we might have scheduled to those other threads without this extra yield, and (b) there's also a chance we will schedule back to the main thread often enough to eventually terminate the program. Once we are in the `Yield` state, Miri is *guaranteed* to terminate, though with very small preemption rates it can take very long.

saethlin: And there's no way we could get these unhelpful leak reports with preemption disabled?

tests/pass/concurrency/scope.rs
RalfJung: I added that now. Even with just 64 seeds this takes a minute on my laptop... so it's a really expensive test. Not sure if we will keep it forever but I guess we can have it for now.

I counted 5 failing seeds below 64 before this PR so I hope this is enough coverage. (In the interval 64..128 there is only one failing test, so the 0..64 interval might just be lucky.)

src/eval.rs
RalfJung: I don't think so, no. The scoped thread logic would have to deliberately yield somewhere in the code that runs after the user thread is done.

PR #2702, opened 2022-11-27
Rustup

RalfJung:
No changes happened on the rustc side, but I want to do a push next and would rather make josh's life easier by integrating some recent history first.


RalfJung:
Ah actually josh can do the push just fine without this.

RalfJung:
Ah, but it would duplicate some things from https://github.com/rust-lang/rust/pull/104617.

Between two josh pushes there must always be a josh pull!

Issue #2709, opened 2022-12-02
ICE when setting -Zmiri-track-pointer-tag


RalfJung:
Trying to debug https://github.com/rust-lang/futures-rs/issues/2669 with `MIRIFLAGS="-Zmiri-track-pointer-tag=12798" cargo miri run`, Miri ICEs:
```
note: tracking was triggered
  --> /home/r/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-util-0.3.25/src/sink/feed.rs:20:16
   |
20 |         Feed { sink, item: Some(item) }
   |                ^^^^ created tag 12798 for unique reference (pointee type futures::futures_channel::mpsc::Sender<u32>) at alloc4773[0x10..0x28] derived from <12797>
   |
   = note: inside `futures::sink::Feed::<'_, futures::futures_channel::mpsc::Sender<u32>, u32>::new` at /home/r/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-util-0.3.25/src/sink/feed.rs:20:16
   = note: inside `futures::sink::Send::<'_, futures::futures_channel::mpsc::Sender<u32>, u32>::new` at /home/r/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-util-0.3.25/src/sink/send.rs:20:22
   = note: inside `<futures::futures_channel::mpsc::Sender<u32> as futures::SinkExt<u32>>::send` at /home/r/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-util-0.3.25/src/sink/mod.rs:224:53
note: inside closure at src/main.rs:6:9
  --> src/main.rs:6:9
   |
6  |         sender.send(n - x).await.unwrap();
   |         ^^^^^^^^^^^^^^^^^^
   = note: inside closure at /home/r/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-executor-0.3.25/src/local_pool.rs:317:23
   = note: inside closure at /home/r/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-executor-0.3.25/src/local_pool.rs:90:37
[...]

PR #2710, opened 2022-12-02
fix ICE in pointer tracking

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/2709


RalfJung:
macos being slow again...
@bors retry

Issue #2644, opened 2022-11-02
Miri's hardcoded PAGE_SIZE is sometimes inaccurate


thomcc:
Miri always [claims the page size is 4k](https://github.com/rust-lang/miri/blob/ccac25ee1cc7cc3a83f8db6a03f8c23deee45f03/src/machine.rs#L35), but on some targets this is not the case. Programs generally should query it at runtime, but a lot of code just hard-codes 4k (like miri), which is not right sometimes.

Plausibly this is an argument for it being configurable, but I think that's likely overkill (and deciding what values are allowable seems tricky). Thankfully, the targets I know of where this varies are seem to be statically fixed in practice, so Miri should just return the accurate-in-practice value. Specifically:

- On `aarch64-apple-darwin` the page size seems to be 16k (apple made a point about this changing in the move to M1). (Unsure if miri supports it, but this is true for `aarch64-apple-ios` too, and perhaps all `aarch64-apple-*`).
- The web assembly spec defines it as 64k (kinda wild, but so it goes): https://webassembly.github.io/spec/core/exec/runtime.html#memory-instances, and WASI seems to follow this.

I don't know of other targets which happen to have different values, and have looked around a bit (according to <https://linux.die.net/man/2/getpagesize>, on `sun4` is a real example where it depended on the specific machine rather than the target, but Rust probably doesn't support it).

Anyway, currently it doesn't matter much, but if #2520 lands (CC @saethlin) it would matter more. It might make sense to do as part of that PR, but this is certainly not necessary, so I filed it as a separate thing rather than asking on that PR.

RalfJung:
> I don't know of other targets which happen to have different values, and have looked around a bit (according to https://linux.die.net/man/2/getpagesize, on sun4 is a real example where it depended on the specific machine rather than the target, but Rust probably doesn't support it).

Seems okay for Miri to define it as 4K then as long as that is a legal option for the target. ;)

But indeed if other targets define it to definitely be something else, Miri should respect that. A separate PR that replaces the constant by a function which takes into account the target would be appreciated.

bjorn3:
> On aarch64-apple-darwin the page size seems to be 16k (apple made a point about this changing in the move to M1). (Unsure if miri supports it, but this is true for aarch64-apple-ios too, and perhaps all aarch64-apple-*).

[...]

PR #2717, opened 2022-12-08
Implement runtime checking for platform page size

a-b-c-1-2-3:
Miri currently assumes all platforms have 4k pages as a hard-coded constant (see #2644). This change replaces it with a `page_size()` function (which just wraps around `sysconf(_SC_PAGESIZE)` on *nix, all other platforms have known values on targets Rust supports) and minimal changes to make this work.

### DIFF
--- a/src/intptrcast.rs
+++ b/src/intptrcast.rs
@@ -56,7 +56,7 @@ impl GlobalStateInner {
             int_to_ptr_map: Vec::default(),
             base_addr: FxHashMap::default(),
             exposed: FxHashSet::default(),
-            next_base_addr: STACK_ADDR,
+            next_base_addr: stack_addr(),
             provenance_mode: config.provenance_mode,
         }
     }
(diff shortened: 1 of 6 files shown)


RalfJung:
I don't think this approach will work. This uses the *host* page size to say what the *target* should do. So if you run Windows target code on a Linux host it'll suddenly use the Linux host pagesize, which is wrong.

Also it is wrong for Miri behavior to depend on the host in any way whatsoever when isolation is enabled (which is the default). Miri must be fully deterministic and behave the same for a given target, no matter the host.

I don't know what the proper solution is here, but having information leak in from the host is not it. Probably we want to check the target in `tcx` and then e.g. return 16k on `aarch64-apple-darwin`.

RalfJung:
Also this is missing a test. Probably https://github.com/rust-lang/miri/blob/master/tests/pass-dep/page_size.rs should assert that on aarch-apple this is 16k, or so?

a-b-c-1-2-3:
Hm, okay. I'll figure out something better and close this for the time being I suppose

RalfJung:
If you have questions about how to implement what I sketched above, feel free to ask [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/269128-miri). :)

PR #2718, opened 2022-12-08
add graceful shim for the custom `O_TMPFILE` file opening flag plus test case

Pointerbender:
I'm trying if I can get the [`tempfile`](https://crates.io/crates/tempfile) crate to work nicely with miri. Right now miri errors out due to an unsupported flag `O_TMPFILE` (= `0x410000`) passed to [`OpenOptions::custom_flags`](https://github.com/Stebalien/tempfile/blob/92ae3e9d6e82bdcfcf114be459d234d3602be2d4/src/file/imp/unix.rs#L71-L86). Interestingly, `tempfile` has a fallback in case the underlying file system does not support the `O_TMPFILE` flag, in which case `open`/`open64` is expected to return the error code `EOPNOTSUPP` (= `95`). This PR adds support for this scenario and also includes a test case (relevant [zulip](https://rust-lang.zulipchat.com/#narrow/stream/269128-miri/topic/miri.20and.20the.20.60tempfile.60.20crate) discussion).

### DIFF
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -605,6 +605,15 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
             // (Technically we do not support *not* setting this flag, but we ignore that.)
             mirror |= o_cloexec;
         }
+        if this.tcx.sess.target.os == "linux" {
+            let o_tmpfile = this.eval_libc_i32("O_TMPFILE")?;
+            if flag & o_tmpfile != 0 {
+                // if the flag contains `O_TMPFILE` then we return a graceful error
+                let eopnotsupp = this.eval_libc("EOPNOTSUPP")?;
+                this.set_last_error(eopnotsupp)?;
+                return Ok(-1);
+            }
+        }
         // If `flag` is not equal to `mirror`, there is an unsupported option enabled in `flag`,
         // then we throw an error.
         if flag != mirror {
--- a/tests/pass-dep/shims/libc-fs.rs
+++ b/tests/pass-dep/shims/libc-fs.rs
@@ -6,7 +6,7 @@
 
 use std::convert::TryInto;
 use std::ffi::CString;
-use std::fs::{canonicalize, remove_file, File};
+use std::fs::{canonicalize, remove_dir_all, remove_file, File};
 use std::io::{Error, ErrorKind, Write};
 use std::os::unix::ffi::OsStrExt;
 use std::path::PathBuf;
@@ -18,6 +18,8 @@ fn main() {
     test_file_open_unix_allow_two_args();
     test_file_open_unix_needs_three_args();
     test_file_open_unix_extra_third_arg();
+    #[cfg(target_os = "linux")]
+    test_o_tmpfile_flag();
 }
 
 fn tmp() -> PathBuf {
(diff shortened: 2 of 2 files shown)


tests/pass-dep/shims/libc-fs.rs
```diff
@@ -135,3 +145,19 @@ fn test_readlink() {
     assert_eq!(res, -1);
     assert_eq!(Error::last_os_error().kind(), ErrorKind::NotFound);
 }
+
+fn test_o_tmpfile_flag() {
+    let dir_path = prepare_dir("miri_test_fs_dir");
+    create_dir(&dir_path).unwrap();
+    // test that the `O_TMPFILE` custom flag gracefully errors instead of panicking
```
RalfJung, line 152:
```suggestion
    // test that the `O_TMPFILE` custom flag gracefully errors instead of stopping execution
```
it wasn't panicking before, it was halting the interpreter

Pointerbender:
Thank you for the speedy review @RalfJung! I just pushed all the requested changes :)

src/shims/unix/fs.rs
```diff
@@ -605,12 +605,14 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
             // (Technically we do not support *not* setting this flag, but we ignore that.)
             mirror |= o_cloexec;
         }
-        let o_tmpfile = this.eval_libc_i32("O_TMPFILE")?;
-        if flag & o_tmpfile != 0 {
-            // if the flag contains `O_TMPFILE` then we return a graceful error
-            let eopnotsupp = this.eval_libc("EOPNOTSUPP")?;
-            this.set_last_error(eopnotsupp)?;
-            return Ok(-1);
+        if cfg!(target_os = "linux") {
```
oli-obk, line 608: This checks the miri binary's target, not the interpreted program's target.

RalfJung, line 608: In the `src` folder, this checks the compile-time target, which is wrong. You need to check the run-time target. You can find that in `this.tcx.sess.target.os`.

RalfJung: Ah, race condition. ;)

RalfJung:
LGTM if CI passes, but please squash the commits.

Pointerbender:
Thanks! The CI passed just now and the commits are squashed, too :+1:

src/shims/unix/fs.rs
```diff
@@ -605,6 +605,15 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
             // (Technically we do not support *not* setting this flag, but we ignore that.)
             mirror |= o_cloexec;
         }
+        if this.tcx.sess.target.os == "linux" {
+            let o_tmpfile = this.eval_libc_i32("O_TMPFILE")?;
+            if flag & o_tmpfile != 0 {
```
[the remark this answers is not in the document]
Pointerbender: Nice catch! Since this PR is already merged, I will address your feedback within my other related PR #2720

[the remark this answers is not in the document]
RalfJung: Since it looks like #2720 might be a bit more tricky, it'd be good to have a separate PR for this.

newpavlov:
Is there a reason why Miri can not properly support `O_TMPFILE`?

RalfJung:
To my knowledge, there's nothing in `std` that would let us properly implement its semantics, and we don't want to be directly calling `libc`/`windows` from Miri as that's just a nightmare to maintain.

Also, old closed PRs aren't really a place to have discussions. Feel free to open a thread on [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/269128-miri) or open an issue if you have a specific feature request.

Issue #1537, opened 2020-09-07
Implement missing shims for Windows


RalfJung:
Here we track shims that Miri currently only implements for Linux and macOS targets, but not for Windows.

If you run into this, try `cargo miri test --target x86_64-unknown-linux-gnu`. This works even on a Windows host (without installing anything extra) and interprets programs on a well-supported Linux target.

* [ ] File system access (https://github.com/rust-lang/miri/issues/3482)
* [x] Threading
* [x] `std::thread::sleep`
* [x] `std::sync`
* [ ] Console functions (https://github.com/rust-lang/miri/issues/2652)

jmerdich:
Also add RNG to the list (rand's default initial seed comes from the OS via `BCryptGenRandom`).

RalfJung:
We do test that `getrandom` works on Windows, why isn't `rand` using that? For this to work we implement `SystemFunction036` which is the actual name of `RtlGenRandom`. I thought supporting `getrandom` would be enough to cover RNG initialization.

RalfJung:
In fact we even test this through `rand`:

https://github.com/rust-lang/miri/blob/2cdd1744b896e8129322229f253f95fd7ad491f1/test-cargo-miri/tests/test.rs#L28

@jmerdich ` SmallRng::from_entropy()` and `rand::thread_rng()` both work on Windows, so I cannot reproduce the problem you are describing.

RalfJung:
Ah, looks like newer versions of `getrandom` changed which function they are using.

RalfJung:
@jmerdich thanks for the notification, https://github.com/rust-lang/miri/pull/1755 should fix this. :)

jmerdich:
@RalfJung thanks for the amazingly prompt fix! I got set up to contribute something for this and was surprised to find it already done! :D

RalfJung:
Ah, sorry for that.^^

Fixing `std::thread::sleep` should be rather easy, if you are looking for something small to get started with. :)

RalfJung:
That's great! I don't know a lot about the Windows File IO APIs, so I can't tell if it makes more sense to share code with Unix or to have a completely different implementation. But either way you can use the Unix version as a template. There's even a testcase already, `tests/run-pass/fs.rs`, that you just need to enable on Windows.

Yeah we don't have all the rustc tooling, but just leaving a comment is enough. :)

mbyt:
Miri is great project, thanks!

[...]

PR #2720, opened 2022-12-09
add support for the `tempfile` crate

Pointerbender:
* Support file creation modes where the owner must have read-write permissions and that none of the owner, group and other may have execute permissions.
* Add `tempfile` as a dependency in `test_dependencies/Cargo.toml`
* Add a test case for method `tempfile::tempfile()`.

### DIFF
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -585,8 +585,13 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 );
             };
 
-            if mode != 0o666 {
-                throw_unsup_format!("non-default mode 0o{:o} is not supported", mode);
+            // currently we only support that the owner must always have
+            // read-write permissions and that none of the owner, group
+            // and other may have execute permissions.
+            let owner_has_read_write_permissions = mode & 0o600 == 0o600;
+            let has_any_execute_permissions = mode & 0o111 != 0;
+            if !owner_has_read_write_permissions || has_any_execute_permissions {
+                throw_unsup_format!("mode 0o{:o} is not supported", mode);
             }
 
             mirror |= o_creat;
--- a/tests/pass-dep/tempfile.rs
+++ b/tests/pass-dep/tempfile.rs
@@ -1,6 +1,7 @@
 //@ignore-target-windows: no libc on Windows
 //@compile-flags: -Zmiri-disable-isolation
 
+use std::ffi::{CStr, CString};
 use std::path::PathBuf;
 
 /// Test that the [`tempfile`] crate is compatible with miri.
@@ -10,20 +11,21 @@ fn main() {
 }
 
 fn tmp() -> PathBuf {
-    std::env::var("MIRI_TEMP")
-        .map(|tmp| {
-            // MIRI_TEMP is set outside of our emulated
-            // program, so it may have path separators that don't
-            // correspond to our target platform. We normalize them here
-            // before constructing a `PathBuf`
+    let path = std::env::var("MIRI_TEMP")
+        .unwrap_or_else(|_| std::env::temp_dir().into_os_string().into_string().unwrap());
+    // These are host paths. We need to convert them to the target.
+    let path = CString::new(path).unwrap();
+    let mut out = Vec::with_capacity(1024);
 
-            #[cfg(windows)]
-            return PathBuf::from(tmp.replace("/", "\\"));
-
-            #[cfg(not(windows))]
-            return PathBuf::from(tmp.replace("\\", "/"));
-        })
-        .unwrap_or_else(|_| std::env::temp_dir())
+    unsafe {
+        extern "Rust" {
+            fn miri_host_to_target_path(path: *const i8, out: *mut i8, out_size: usize) -> usize;
+        }
+        let ret = miri_host_to_target_path(path.as_ptr(), out.as_mut_ptr(), out.capacity());
+        assert_eq!(ret, 0);
+        let out = CStr::from_ptr(out.as_ptr()).to_str().unwrap();
+        PathBuf::from(out)
+    }
 }
 
 fn test_tempfile() {
--- a/tests/pass-dep/tempfile.rs
+++ b/tests/pass-dep/tempfile.rs
@@ -1,0 +1,34 @@
+//@ignore-target-windows: no libc on Windows
+//@compile-flags: -Zmiri-disable-isolation
+
+//! Test that the [`tempfile`] crate is compatible with miri.
+fn main() {
+    test_tempfile();
+    test_tempfile_in();
+}
+
+fn tmp() -> PathBuf {
+    std::env::var("MIRI_TEMP")
+        .map(|tmp| {
+            // MIRI_TEMP is set outside of our emulated
+            // program, so it may have path separators that don't
+            // correspond to our target platform. We normalize them here
+            // before constructing a `PathBuf`
+
+            #[cfg(windows)]
+            return PathBuf::from(tmp.replace("/", "\\"));
+
+            #[cfg(not(windows))]
+            return PathBuf::from(tmp.replace("\\", "/"));
+        })
+        .unwrap_or_else(|_| std::env::temp_dir())
+}
+
+fn test_tempfile() {
+    tempfile::tempfile().unwrap();
+}
+
+fn test_tempfile_in() {
+    let dir_path = tmp();
+    tempfile::tempfile_in(dir_path).unwrap();
+}
--- a/tests/pass-dep/tempfile.rs
+++ b/tests/pass-dep/tempfile.rs
@@ -1,7 +1,9 @@
 //@ignore-target-windows: no libc on Windows
 //@compile-flags: -Zmiri-disable-isolation
 
-//! Test that the [`tempfile`] crate is compatible with miri.
+use std::path::PathBuf;
+
+/// Test that the [`tempfile`] crate is compatible with miri.
 fn main() {
     test_tempfile();
     test_tempfile_in();
--- a/tests/pass-dep/tempfile.rs
+++ b/tests/pass-dep/tempfile.rs
@@ -6,7 +6,7 @@ use std::path::PathBuf;
 
 /// Test that the [`tempfile`] crate is compatible with miri.
 fn main() {
-    test_tempfile();
+    // test_tempfile(); // does not work when host!=target
     test_tempfile_in();
 }
 
@@ -28,9 +28,10 @@ fn tmp() -> PathBuf {
     }
 }
 
-fn test_tempfile() {
-    tempfile::tempfile().unwrap();
-}
+// does not work when host!=target:
+// fn test_tempfile() {
+//     tempfile::tempfile().unwrap();
+// }
 
 fn test_tempfile_in() {
     let dir_path = tmp();
--- a/test_dependencies/Cargo.lock
+++ b/test_dependencies/Cargo.lock
@@ -27,6 +27,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
+name = "fastrand"
+version = "1.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a7a407cfaa3385c4ae6b23e84623d48c2798d06e3e6a1878f7f59f17b3f86499"
+dependencies = [
+ "instant",
+]
+
+[[package]]
 name = "getrandom"
 version = "0.1.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -55,6 +64,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "62b467343b94ba476dcb2500d242dadbb39557df889310ac77c5d99100aaac33"
 dependencies = [
  "libc",
+]
+
+[[package]]
+name = "instant"
+version = "0.1.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7a5bbe824c507c5da5956355e86a746d82e0e1464f65d862cc5e71da70e94b2c"
+dependencies = [
+ "cfg-if",
 ]
 
 [[package]]
@@ -110,6 +128,7 @@ dependencies = [
  "num_cpus",
  "page_size",
  "rand",
+ "tempfile",
  "tokio",
 ]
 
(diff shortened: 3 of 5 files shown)


src/shims/unix/fs.rs
```diff
@@ -585,8 +585,11 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 );
             };
 
-            if mode != 0o666 {
-                throw_unsup_format!("non-default mode 0o{:o} is not supported", mode);
+            // currently we only support that the owner must always have
+            // read-write permissions and that none of the owner, group
+            // and other may have execute permissions.
+            if mode & 0o666 != mode | 0o600 {
```
oli-obk, line 591: my brain broke trying to process this. Maybe we should start separating out the information and process it with exhaustive matches?

Pointerbender: To make this easier to digest, I can think of two flavors :)

```rust
            if !matches!(
                mode,
                0o600
                    | 0o602
                    | 0o604
                    | 0o606
                    | 0o620
                    | 0o622
                    | 0o624
                    | 0o626
                    | 0o640
                    | 0o642
                    | 0o644
                    | 0o646
                    | 0o660
                    | 0o662
                    | 0o664
                    | 0o666
            ) {
                throw_unsup_format!("mode 0o{:o} is not supported", mode);
            }
```

or

```rust
            let owner_read_write = 0o600;
            let group_read_write = 0o060;
            let other_read_write = 0o006;
            if mode & (owner_read_write | group_read_write | other_read_write)
                != mode | owner_read_write
            {
                throw_unsup_format!("mode 0o{:o} is not supported", mode);
            }
```

Which one do you prefer?

Pointerbender: p.s. another third option:

```rust
            let owner_read_write = 0o600;
            let execute = 0o111;
            if mode & owner_read_write != owner_read_write || mode & execute != 0
            {
                throw_unsup_format!("mode 0o{:o} is not supported", mode);
            }
```

oli-obk: I was thinking about something along the lines of

```rust
            let owner = (mode >> 6) & 0o600;
            let group = (mode >> 3) & 0o060;
            let other = mode & 0o006;
            if do_some_logic_here {
                throw_unsup_format!("mode 0o{:o} is not supported", mode);
            }
```

Pointerbender:
I see the windows CI is failing, but there is a `//@ignore-target-windows: no libc on Windows` at the top of `tests/pass-dep/shims/libc-fs.rs`. How can I best proceed in this case? (Can I run the Windows tests from Ubuntu 18.04? I don't have access to a Windows machine, unfortunately).

oli-obk:
> Can I run the Windows tests from Ubuntu 18.04? I don't have access to a Windows machine, unfortunately).

Yes, miri supports cross interpretation. For testing you can use the `MIRI_TEST_TARGET` env var to choose your target

Pointerbender:
I think I may still might be doing something the wrong way. If I look at the logs of the [previous](https://github.com/rust-lang/miri/actions/runs/3658832048/jobs/6184102502#step:8:1724) failed Windows run, I see two mentions of targets:

* i686-uwp-windows-msvc (at the very top  of the Github job)
* x86_64-unknown-linux-gnu (in the arguments passed to the command that failed)

I tried running both targets on my Ubuntu machine like this:

```shell
$ MIRI_TEST_TARGET="i686-uwp-windows-msvc" ./miri test tests/pass-dep/shims/libc-fs.rs
$ MIRI_TEST_TARGET="x86_64-unknown-linux-gnu" ./miri test tests/pass-dep/shims/libc-fs.rs
```

and for good measure also all the other targets that mention Windows:

```shell
$ rustc --print target-list | grep windows
aarch64-pc-windows-gnullvm
aarch64-pc-windows-msvc
aarch64-uwp-windows-msvc
i586-pc-windows-msvc
i686-pc-windows-gnu
i686-pc-windows-msvc
i686-uwp-windows-gnu
i686-uwp-windows-msvc
thumbv7a-pc-windows-msvc
thumbv7a-uwp-windows-msvc
x86_64-pc-windows-gnu
x86_64-pc-windows-gnullvm
x86_64-pc-windows-msvc
x86_64-uwp-windows-gnu
x86_64-uwp-windows-msvc

$ rustc --print target-list | grep windows | xargs -I {} env MIRI_TEST_TARGET="{}" ./miri test tests/pass-dep/shims/libc-fs.rs
```

But the test passes on all Windows targets on my machine, with the exception of targets `thumbv7a-uwp-windows-msvc` and `thumbv7a-pc-windows-msvc` for which it failed to compile the dependencies and could not even run the test. It might be something specific to running the miri test on a Windows machine for a `x86_64-unknown-linux-gnu` target, perhaps?

The [next](https://github.com/rust-lang/miri/actions/runs/3660294905/jobs/6187262816#step:8:1722) Windows CI also finished just now and has a slightly different error now:

The first time around, the error was:

```
actual output differed from expected
  --- tests/pass-dep\shims\libc-fs.stderr
  +++ <stderr output>
   hello dup fd
  +thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "entity not found" }', $DIR/libc-fs.rs:LL:CC
  +stack backtrace:
  +   0: std::panicking::begin_panic_handler
  + at RUSTLIB/std/src/panicking.rs:LL:CC
  +   1: std::rt::panic_fmt
  + at RUSTLIB/core/src/panicking.rs:LL:CC
  +   2: std::result::unwrap_failed
  + at RUSTLIB/core/src/result.rs:LL:CC
  +   3: std::result::Result::unwrap
  + at RUSTLIB/core/src/result.rs:LL:CC
  +   4: test_tempfile
  + at $DIR/libc-fs.rs:LL:CC
  +   5: main
  + at $DIR/libc-fs.rs:LL:CC
  +   6: <fn() as std::ops::FnOnce<()>>::call_once - shim(fn())
  + at RUSTLIB/core/src/ops/function.rs:LL:CC
  +note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

And the second time the error is:

```
actual output differed from expected
  --- tests/pass-dep\shims\libc-fs.stderr
  +++ <stderr output>
   hello dup fd
  +thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 36, kind: InvalidFilename, message: "invalid filename" }', $DIR/libc-fs.rs:LL:CC
  +stack backtrace:
  +   0: std::panicking::begin_panic_handler
  + at RUSTLIB/std/src/panicking.rs:LL:CC
  +   1: std::rt::panic_fmt
  + at RUSTLIB/core/src/panicking.rs:LL:CC
  +   2: std::result::unwrap_failed
  + at RUSTLIB/core/src/result.rs:LL:CC
  +   3: std::result::Result::unwrap
  + at RUSTLIB/core/src/result.rs:LL:CC
  +   4: test_tempfile
  + at $DIR/libc-fs.rs:LL:CC
  +   5: main
  + at $DIR/libc-fs.rs:LL:CC
  +   6: <fn() as std::ops::FnOnce<()>>::call_once - shim(fn())
  + at RUSTLIB/core/src/ops/function.rs:LL:CC
  +note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

The difference is that in the first time, the test said (now no longer visible due to my earlier force push):

```rust
/// Test that the [`tempfile`] crate is compatible with miri.
fn test_tempfile() {
    tempfile::tempfile().unwrap();
}
```

Which I changed to:

```rust
/// Test that the [`tempfile`] crate is compatible with miri.
fn test_tempfile() {
    let dir_path = tmp();
    tempfile::tempfile_in(dir_path).unwrap();
}
```

One other idea I could try is to add `#[cfg(not(target_family = "windows"))]` to `fn test_tempfile()`, but I suspect that won't matter in this case, because Windows fails on the test target for `x86_64-unknown-linux-gnu` :'-( I think I hit a dead end again on how to debug this one, might I solicit here again for another thing I could try? :)

Also the Mac OS build had a network blip just now and failed early. I tried force pushing it again, but that doesn't seem to trigger a restart of the CI without changes (`git` refuses to force push because it claims "*Everything up-to-date*"). Is there a recommended way to restart the Mac OS job in case I run into that again? :)

Thanks!

oli-obk:
ah, the problem is a windows host interpreting a linux target. Oof. Maybe this is due to backslashes in names or sth? Not sure how to debug this.

LegNeato:
I would imagine it is because the call to `tmp` is returning a unix path (because the host is unix and miri's fs implementation is sort of a passthrough with some path mangling), which is then passed into `tempfile_in()`, which is "running" on windows, so a windows API gets a unix path. With the old test, the path doesn't round trip through the host.

These sort of issues are going to keep coming up, as I mentioned in https://rust-lang.zulipchat.com/#narrow/stream/269128-miri/topic/Filesystem.20abstraction.

Pointerbender:
Thanks for that pointer @LegNeato . Going through the zulip topic that you linked, it sounds like it would be a proper effort add path translation from a unix target family to a windows Host. On the bright side, we now have an actual example of this being a problem on Windows running a linux target :P (cc #1537)

Considering that this can not be properly fixed without access to a Windows environment and a significant time investment, I believe that leaves me with two possible alternatives:

# Alternative 1
Adjust the test case for `tempfile` that it doesn't fail when the `tempfile` methods return a `InvalidFilename` or `NotFound` IO error kind and see if that is enough to make the Windows CI pass. Perhaps there is even a way to detect from a test whether it's running a linux target on a windows host, perhaps through an existing environment variable, to narrow it down even more? Then we track this support gap for `tempfile` on #1537 or in a separate Github issue, in the hopes that someone with a Windows machine and enough time finds it and fixes it properly some day.

# Alternative 2
We do not commit at this point in time to adding `tempfile` as a supported third party dependency by removing the `tempfile` test case(s) and the `tempfile` dependency; and replace those with test cases for the file creation modes only. This will make miri play nice with `tempfile` on host/target combinations that do not suffer from path translation issues, but without "official" support for the `tempfile` crate.

@oli-obk What do you think is the best course of action? Or maybe you know of a third viable alternative?

Thanks!

RalfJung:
The PR seems to effectively just ignore most of the mode. Is that really a good idea?

RalfJung:
Regarding the path problem: we already do path separator adjustment if host!=target, which so far was enough. It would be good to figure out what exactly happens here to understand why this is no longer enough. It is not too surprising that naive separator substitution is insufficient but it would still be good to learn why exactly it fails now.

I don't have a Windows machine either, but you can still add a bunch of debug printing and then see what happens on CI.

RalfJung:
My first guess would be that the `!path.is_absolute()` check is going wrong -- the path translation between Linux and Windows hosts maps `C:\dir` to `C:/dir` which is not considered absolute as a Unix path. I am trying to fix that in https://github.com/rust-lang/miri/pull/2725, so once that lands it might be worth rebasing and retrying.

RalfJung:
FWIW I think this would also not have been a problem if tempfile just called `canonicalize` -- but instead it hand-rolls that function:
```rust
    if !path.is_absolute() {
        let cur_dir = env::current_dir()?;
        tmp = cur_dir.join(path);
        path = &tmp;
    }
```

Pointerbender:
Thanks for all that input @RalfJung !

My next steps for this PR will be: 

* Wait for #2725 to be merged.
* Rebase master onto the branch of this PR.
* Move the `tempfile` tests to `tests/pass-dep/tempfile.rs`.
* Run the CI again to see what happens. Obtain some extra information by adding debug statements if it's still failing.
* In the mean time, explore if/how we could support the file modes cross-platform (due to that the host platform could differ from the target platform).

Pointerbender:
> the flags will need translation

From what I've been able to research so far, it appears to be the case that the Windows [file creation API](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea) has no concept of group permissions. Even if we find a work-around by calling the group permissions APIs ourselves after creating the file first, there is a discrepancy between the permission models of Windows and Unix: Unix distinguishes between the owner, a single additional group and everyone else; while Windows has a more complicated model that allows multiple users and multiple groups, and there is not a clear-cut way to map the only Linux owner/group to a Windows equivalent ("Everybody" is a group in Windows, though, so that part is at least somewhat straight-forward - assuming there is a Windows API that we can call to set the permissions for "Everyone", I was unable to find which API we need for that so far).

tests/pass-dep/tempfile.rs
```diff
@@ -0,0 +1,34 @@
...
+            // before constructing a `PathBuf`
+
+            #[cfg(windows)]
+            return PathBuf::from(tmp.replace("/", "\\"));
+
+            #[cfg(not(windows))]
+            return PathBuf::from(tmp.replace("\\", "/"));
+        })
+        .unwrap_or_else(|_| std::env::temp_dir())
+}
```
RalfJung, line 27: You copied outdated code from somewhere. Please see libc-fs.rs for the updated version of this function.

tests/pass-dep/tempfile.rs
```diff
@@ -0,0 +1,34 @@
...
+
+fn test_tempfile() {
+    tempfile::tempfile().unwrap();
```
RalfJung, line 30: This will call the target `temp_dir` function which might not work when host!=target. That's why we have this MIRI_TEMP stuff.

RalfJung:
Setting it later won't be entirely correct though, someone could access the file between it being created and the permissions being adjusted.

I guess the alternative would be to just not support this on a Windows host. So far we have kept our set of supported operations host-independent but as we move beyond what the Rust stdlib can provide, that might not always be feasible.

Pointerbender:
That is a good point. In that case, does it makes sense to support all file modes on Unix platforms (including execute permissions) or still just a subset of the file modes which lets miri stay forward compatible for when the std lib does add support for this in the future? So far I was only able to deduce that modes `0o600` and `0o666` are both compatible with the Unix and Windows permission models (there might be more compatible file modes, I just didn't have a chance to explore all of them yet).

To clarify what you mean by "*not support this on a Windows host*", should we either: 1) halt the interpreter by showing an unsupported error or 2) silently default to the default permissions when executing a Unix target on a Windows host?

Thanks!

RalfJung:
In this version we would
- support all modes on a Unix host, just forwarding the flag
- throw_unsup on a Windows host for non-default flags
- add `//@ignore-host-windows` to the test (which needs to be implemented in ui_test first)

RalfJung:
@Pointerbender I am closing this PR due to inactivity. Feel free to reopen or open a new PR if you want to get back to this!

Jefffrey:
Hey I was wondering on the status of this. Would it be possible for me to help push this along, as the PR has been inactive for a while now?

cc @Pointerbender

PR #2721, opened 2022-12-09
Allow configurable and platform-specific page sizes

### LINKED ISSUE #2644: Miri's hardcoded PAGE_SIZE is sometimes inaccurate

a-b-c-1-2-3:
This fixes #2644 by setting platform-default page sizes along with a command line flag to override size to a specific value (e.g. in the case of aarch64 Linux on M1 silicon). There's still some code cleanup to be done and tests need to be added but I'm opening this for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,6 +399,8 @@ to Miri failing to detect cases of undefined behavior in a program.
 * `-Zmiri-track-weak-memory-loads` shows a backtrace when weak memory emulation returns an outdated
   value from a load. This can help diagnose problems that disappear under
   `-Zmiri-disable-weak-memory-emulation`.
+* `-Zmiri-force-page-size=<num>` overrides the default page size for an architecture, in multiples of 1k.
+  `4` is default for most targets. This value should always be a power of 2 and nonzero.
 
 [function ABI]: https://doc.rust-lang.org/reference/items/functions.html#extern-function-qualifier
 
--- a/src/intptrcast.rs
+++ b/src/intptrcast.rs
@@ -51,12 +51,12 @@ impl VisitTags for GlobalStateInner {
 }
 
 impl GlobalStateInner {
-    pub fn new(config: &MiriConfig) -> Self {
+    pub fn new(config: &MiriConfig, stack_addr: u64) -> Self {
         GlobalStateInner {
             int_to_ptr_map: Vec::default(),
             base_addr: FxHashMap::default(),
             exposed: FxHashSet::default(),
-            next_base_addr: STACK_ADDR,
+            next_base_addr: stack_addr,
             provenance_mode: config.provenance_mode,
         }
     }
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -31,11 +31,6 @@ use crate::{
     *,
 };
 
-// Some global facts about the emulated machine.
-pub const PAGE_SIZE: u64 = 4 * 1024; // FIXME: adjust to target architecture
-pub const STACK_ADDR: u64 = 32 * PAGE_SIZE; // not really about the "stack", but where we start assigning integer addresses to allocations
-pub const STACK_SIZE: u64 = 16 * PAGE_SIZE; // whatever
-
 /// Extra data stored with each stack frame
 pub struct FrameExtra<'tcx> {
     /// Extra data for Stacked Borrows.
@@ -469,6 +464,10 @@ pub struct MiriMachine<'mir, 'tcx> {
     pub(crate) since_gc: u32,
     /// The number of CPUs to be reported by miri.
     pub(crate) num_cpus: u32,
+    /// Determines Miri's page size and associated values
+    pub(crate) page_size: u64,
+    pub(crate) stack_addr: u64,
+    pub(crate) stack_size: u64,
 }
 
 impl<'mir, 'tcx> MiriMachine<'mir, 'tcx> {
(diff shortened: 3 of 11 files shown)


src/intptrcast.rs
```diff
@@ -56,7 +56,7 @@ impl GlobalStateInner {
             int_to_ptr_map: Vec::default(),
             base_addr: FxHashMap::default(),
             exposed: FxHashSet::default(),
-            next_base_addr: STACK_ADDR,
+            next_base_addr: config.page_size.unwrap_or(128 * 1024),
```
RalfJung, line 59: This is ultimately arbitrary, but it might still make sense to keep this the way it was before... can you pass `stack_addr` into this constructor?

src/machine.rs
```diff
@@ -482,6 +481,21 @@ impl<'mir, 'tcx> MiriMachine<'mir, 'tcx> {
...
+        } else {
+            let target = &layout_cx.tcx.sess.target;
+            match target.arch.as_ref() {
+                "wasm32" | "wasm64" => 64 * 1024,
+                "aarch64" =>
+                    if target.options.vendor.as_ref().eq("apple") {
+                        16 * 1024
+                    } else {
+                        4 * 1024
+                    },
```
RalfJung, line 495: Are there references you can link here that document these non-4K page sizes?

src/machine.rs
```diff
@@ -482,6 +481,21 @@ impl<'mir, 'tcx> MiriMachine<'mir, 'tcx> {
...
         let borrow_tracker = config.borrow_tracker.map(|bt| bt.instanciate_global_state(config));
         let data_race = config.data_race_detector.then(|| data_race::GlobalState::new(config));
+        let page_size = if let Some(page_size) = config.page_size {
+            page_size
+        } else {
+            let target = &layout_cx.tcx.sess.target;
+            match target.arch.as_ref() {
+                "wasm32" | "wasm64" => 64 * 1024,
+                "aarch64" =>
+                    if target.options.vendor.as_ref().eq("apple") {
```
RalfJung, line 491:
```suggestion
                    if target.options.vendor.as_ref() == "apple" {
```

README.md
```diff
@@ -399,6 +399,8 @@ to Miri failing to detect cases of undefined behavior in a program.
 * `-Zmiri-track-weak-memory-loads` shows a backtrace when weak memory emulation returns an outdated
   value from a load. This can help diagnose problems that disappear under
   `-Zmiri-disable-weak-memory-emulation`.
+* `-Zmiri-force-page-size=<num>` overrides the default page size for an architecture, in multiples of 1k.
+  `4` is default for most targets. This value should always be a power of 2 and nonzero.
```
RalfJung, line 403: What is the motivation for this flag?

I don't have fundamental objections here, but if we have such a flag we definitely need a test for it, too.

a-b-c-1-2-3: Mostly because Linux allows variable page sizes (Asahi Linux is an important example I think) and some other platforms also have non-4k pages e.g. Windows on Alpha and IA64 (https://devblogs.microsoft.com/oldnewthing/20210510-00/?p=105200). I'll add a test, though

src/machine.rs
a-b-c-1-2-3: I'll add some in, thanks. I took the values based on #2644

src/intptrcast.rs
a-b-c-1-2-3: To clarify - should I just set a constant equal to 128k for this?

src/machine.rs
RalfJung: Yeah before we enshrine this in code I'd like to see sources for these numbers. That issue contains a source for wasm which you can just use; I assume it can't be too hard to find a source for apple.

src/intptrcast.rs
RalfJung: No it should use `machine.stack_addr`, as it did before.

src/machine.rs
a-b-c-1-2-3: Added some sources; are they fine? Seems like Apple never *definitively* said post-A8 iOS/macOS uses 16k pages in their docs

src/intptrcast.rs
a-b-c-1-2-3: Done

src/machine.rs
```diff
@@ -482,11 +481,30 @@ impl<'mir, 'tcx> MiriMachine<'mir, 'tcx> {
...
+            page_size
+        } else {
+            let target = &layout_cx.tcx.sess.target;
+            match target.arch.as_ref() {
+                "wasm32" | "wasm64" => 64 * 1024, // https://webassembly.github.io/spec/core/exec/runtime.html#memory-instances
+                "aarch64" =>
+                    if target.options.vendor.as_ref() == "apple" {
+                        16 * 1024 // No "definitive" source, but see:
+                    // https://github.com/ziglang/zig/issues/11308
+                    // https://www.wwdcnotes.com/notes/wwdc20/10214/ etc.
```
RalfJung, line 494:
```suggestion
                    if target.options.vendor.as_ref() == "apple" {
                        // No "definitive" source, but see:
                        // https://github.com/ziglang/zig/issues/11308
                        // https://www.wwdcnotes.com/notes/wwdc20/10214/ etc.
                        16 * 1024
```

RalfJung:
Looks good. :) Just a formatting nit, and please squash the commits.

src/machine.rs
```diff
@@ -482,11 +481,31 @@ impl<'mir, 'tcx> MiriMachine<'mir, 'tcx> {
...
+            page_size
+        } else {
+            let target = &layout_cx.tcx.sess.target;
+            match target.arch.as_ref() {
+                "wasm32" | "wasm64" => 64 * 1024, // https://webassembly.github.io/spec/core/exec/runtime.html#memory-instances
+                "aarch64" =>
+                    if target.options.vendor.as_ref() == "apple" {
+                        // No "definitive" source, but see:
+                        // https://www.wwdcnotes.com/notes/wwdc20/10214/
+                        // https://github.com/ziglang/zig/issues/11308 etc.
```
bjorn3, line 494: The zig issue is about linux on M1.

a-b-c-1-2-3:
Done! Thanks for the feedback and sorry for the hiccups

src/machine.rs
a-b-c-1-2-3: M1 is weird because some features require 16k pages (IOMMU as far as I recall?) but this also applies retroactively to Apple's ARM chips going back a long way (see https://developer.apple.com/library/archive/documentation/Performance/Conceptual/ManagingMemory/Articles/AboutMemory.html detailing that ever since A7 and A8, aarch64-apple-* has defaulted to 16k pages) and hackintoshes on aarch64 aren't a thing that exists right now AFAICT

bjorn3: > M1 is weird because some features require 16k pages (IOMMU as far as I recall?)

Yeah and Linux can't deal with mixed page sizes, unlike macOS which is just fine with 16k for IOMMU and 4k for Rosetta 2.

a-b-c-1-2-3: The issue I think still highlights that aarch64-apple prefers 16k pages by default

PR #2725, opened 2022-12-11
make unix path handling on Windows hosts (and vice versa) preserve absoluteness

RalfJung:
Also adds a magic Miri extern function so this conversion can be invoked by the program. That avoids having to duplicate that logic.


RalfJung:
This seems to actually work. :)

@rust-lang/miri are you fine with this patch or is it too terrible of a hack?^^

oli-obk:
The duplication in the tests is annoying, but that's a separate issue we have with all of those extern C miri helpers.

lgtm!

Issue #2730, opened 2022-12-16
Transition josh instructions to docker


RalfJung:
The 22.12.06 release of josh that came out last week has all the patches we need. So we can change our instructions to use released versions of josh, probably via docker, instead of telling people to clone the josh sources.

PR #2736, opened 2022-12-21
update josh instructions

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/2730

r\? @oli-obk


RalfJung:
Unfortunately it turns out that the docker container changed in ways that breaks Ctrl-C.

Maybe we should recommend `cargo install` instead... the question then is just, where to put the cache?

PR #2765, opened 2023-01-25
add scfix test

RalfJung:
I'm not sure if we currently *guarantee* to pass this test, but at least I was unable to get it to fail.

Cc @cbeuw

### DIFF
--- a/tests/pass/0weak_memory_consistency.rs
+++ b/tests/pass/0weak_memory_consistency.rs
@@ -286,6 +286,40 @@ fn test_iriw_sc_rlx() {
     assert!(c || d);
 }
 
+// Another test for C++20 SCfix.
+fn scfix() {
+    let x = static_atomic_bool(false);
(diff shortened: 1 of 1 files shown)


tests/pass/0weak_memory_consistency.rs
```diff
@@ -286,6 +286,40 @@ fn test_iriw_sc_rlx() {
...
+        fence(SeqCst);
+        let b = y.load(Relaxed);
+        (a, b)
+    });
+
+    let thread2 = spawn(move || {
+        x.store(true, Relaxed);
+    });
+    let thread3 = spawn(move || {
+        x.store(true, Relaxed);
```
cbeuw, line 305: This should've been `y.store(true, Relaxed)` right?

RalfJung: Oh... yes. Good catch!

PR #2767, opened 2023-01-26
Explain how to make josh authenticate with github via ssh

oli-obk:
Without this I was unable to push, even though pushing directly to https without josh works.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -242,8 +242,7 @@ josh-proxy --local=$HOME/.cache/josh --remote=https://github.com --no-background
 
 This uses a directory `$HOME/.cache/josh` as a cache, to speed up repeated pulling/pushing.
 
-Note that josh is unable to handle ssh auth natively, but you can force it to use ssh auth
-by adding the following to your `.gitconfig`:
+To make josh push via ssh instead of https, you can add the following to your `.gitconfig`:
 
 ```toml
 [url "git@github.com:"]
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -242,6 +242,14 @@ josh-proxy --local=$HOME/.cache/josh --remote=https://github.com --no-background
 
 This uses a directory `$HOME/.cache/josh` as a cache, to speed up repeated pulling/pushing.
 
+Note that josh is unable to handle ssh auth natively, but you can force it to use ssh auth
+by adding the following to your `.gitconfig`:
+
+```toml
+[url "git@github.com:"]
+    pushInsteadOf = https://github.com/
+```
+
 ### Importing changes from the rustc repo
 
 Josh needs to be running, as described above.


CONTRIBUTING.md
```diff
@@ -282,3 +285,18 @@ up-to-date master branch:
...
 main repository.
+
+#### Authentication issues during push
+
+Unfortunately `josh` does not support ssh authentication. Going via https may cause
+josh to fail to authenticate with github. In that case you can additionally pass
+`--require-auth` to your `josh-proxy` invocation.
+
+To authenticate with github, you'll need to enter your username and a Personal
+Access Token (PAT), *not* your passwword. You cannot authenticate with your password.
```
saethlin, line 296:
```suggestion
Access Token (PAT), *not* your password. You cannot authenticate with your password.
```

CONTRIBUTING.md
```diff
@@ -282,3 +285,18 @@ up-to-date master branch:
...
+
+To create a PAT, follow the instructions
+[here](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token#creating-a-fine-grained-personal-access-token).
+
+Make sure to enable the "Contents" reand *and* write permission for that PAT. If you forget,
```
saethlin, line 301:
```suggestion
Make sure to enable the "Contents" read and *and* write permission for that PAT. If you forget,
```

RalfJung:
The strange thing is that I am using SSH authentication and josh is working just fine. So the statement that josh does not support SSH authentication is definitely wrong. I don't know what's different about my setup vs yours, but maybe you could phrase it as "if you get authentication errors from josh, here is something you can do (but if you don't get any errors, just ignore this part)"?

CONTRIBUTING.md
```diff
@@ -282,3 +285,18 @@ up-to-date master branch:
...
+`--require-auth` to your `josh-proxy` invocation.
+
+To authenticate with github, you'll need to enter your username and a Personal
+Access Token (PAT), *not* your password. You cannot authenticate with your password.
+
+To create a PAT, follow the instructions
+[here](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token#creating-a-fine-grained-personal-access-token).
+
+Make sure to enable the "Contents" read and *and* write permission for that PAT. If you forget,
+you need to create a new PAT, you cannot edit the permissions after it has been created.
```
RalfJung, line 302: This should come with a warning that this access token gives r/w access to *all* your repos (without even requiring 2FA!) and hence should be guarded carefully; it's more security sensitive than the password (assuming 2FA is enabled).

CONTRIBUTING.md
```diff
@@ -282,3 +285,18 @@ up-to-date master branch:
 This will create a new branch called 'miri' in your fork, and the output should
 include a link to create a rustc PR that will integrate those changes into the
 main repository.
+
+#### Authentication issues during push
+
+Unfortunately `josh` does not support ssh authentication. Going via https may cause
+josh to fail to authenticate with github. In that case you can additionally pass
+`--require-auth` to your `josh-proxy` invocation.
```
RalfJung, line 293: Before we recommend this, could you maybe try adding this to your `~/.gitconfig`?
```
[url "git@github.com:"]
        pushInsteadOf = https://github.com/
```
I think that's what makes josh pushes work for me even though josh gets passed `--remote=https://github.com`.

oli-obk: that did the trick, thanks!

CONTRIBUTING.md
oli-obk: fwiw: the new tokens are per-repo, only the legacy ones are global.

CONTRIBUTING.md
```diff
@@ -242,6 +242,14 @@ josh-proxy --local=$HOME/.cache/josh --remote=https://github.com --no-background
 
 This uses a directory `$HOME/.cache/josh` as a cache, to speed up repeated pulling/pushing.
 
+Note that josh is unable to handle ssh auth natively, but you can force it to use ssh auth
+by adding the following to your `.gitconfig`:
```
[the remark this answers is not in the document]
oli-obk: I saw https://github.com/josh-project/josh/issues/22 and saw that using ssh paths didn't work for me either, so I just assumed it couldn't really support it

PR #2768, opened 2023-01-26
Add `miri_alloc_phys` to support basic MMIO

mkroening:
This adds `miri_alloc_sys` to support basic passive, non-emulated MMIO use cases.

Previous discussion: https://rust-lang.zulipchat.com/#narrow/stream/269128-miri/topic/MMIO.20Ideas

## Motivation

Applications that use MMIO look something like this:

```rust
#![feature(strict_provenance)]

pub const PHYS_ADDR: usize = 0xB8000;

fn main() {
    let ptr = paging::map(PHYS_ADDR, 2);
    unsafe {
        ptr.write_volatile(0u8);
        ptr.add(1).write_volatile(0u8);
    }
}

mod paging {
    /// Maps physical memory into virtual memory. 
    pub fn map(phys_addr: usize, size: usize) -> *mut u8 {
        // In real applications this would be more sophisticated.
        // We might want to ensure that physical memory is only mapped once.
        // For this example, we assume an existing identity mapping and don't perform any checks.
        core::ptr::from_exposed_addr_mut::<u8>(phys_addr)
    }
}
```

This example currently fails to run in Miri like this:

```console
warning: unused variable: `size`
  --> tests/pass/mmio_real.rs:15:34
   |
15 |     pub fn map(phys_addr: usize, size: usize) -> *mut u8 {
   |                                  ^^^^ help: if this is intentional, prefix it with an underscore: `_size`
   |
   = note: `#[warn(unused_variables)]` on by default

warning: integer-to-pointer cast
  --> tests/pass/mmio_real.rs:19:9
   |
19 |         core::ptr::from_exposed_addr_mut::<u8>(phys_addr)
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ integer-to-pointer cast
   |
   = help: This program is using integer-to-pointer casts or (equivalently) `ptr::from_exposed_addr`,
   = help: which means that Miri might miss pointer bugs in this program.
   = help: See https://doc.rust-lang.org/nightly/std/ptr/fn.from_exposed_addr.html for more details on that operation.
   = help: To ensure that Miri does not miss bugs in your program, use Strict Provenance APIs (https://doc.rust-lang.org/nightly/std/ptr/index.html#strict-provenance, https://crates.io/crates/sptr) instead.
   = help: You can then pass the `-Zmiri-strict-provenance` flag to Miri, to ensure you are not relying on `from_exposed_addr` semantics.
   = help: Alternatively, the `-Zmiri-permissive-provenance` flag disables this warning.
   = note: BACKTRACE:
   = note: inside `paging::map` at tests/pass/mmio_real.rs:19:9: 19:58
note: inside `main`
  --> tests/pass/mmio_real.rs:6:15
   |
6  |     let ptr = paging::map(PHYS_ADDR, 2);
   |               ^^^^^^^^^^^^^^^^^^^^^^^^^

error: Undefined Behavior: dereferencing pointer failed: 0xb8000[noalloc] is a dangling pointer (it has no provenance)
 --> tests/pass/mmio_real.rs:8:9
  |
8 |         ptr.write_volatile(0u8);
  |         ^^^^^^^^^^^^^^^^^^^^^^^ dereferencing pointer failed: 0xb8000[noalloc] is a dangling pointer (it has no provenance)
  |
  = note: BACKTRACE:
  = note: inside `main` at tests/pass/mmio_real.rs:8:9: 8:32

error: aborting due to previous error; 2 warnings emitted
```

There are cases where the underlying MMIO memory is not changed by an external device at all, for example VGA text buffers (which is where the address of this example comes from). In those cases of passive devices, a way of getting appropriate, passive memory from Miri would be enough to run those apps in Miri.

## Proposal

This PR adds a new Miri-provided extern function to allocate physical memory at a given physical address:

```rust
/// Miri-provided extern function to "allocate" "physical" memory.
/// 
/// This is useful when running MMIO applications with miri. Currently,
/// this returns normal memory without any emulated device modifying it.
fn miri_alloc_phys(size: usize, phys_addr: usize) -> *mut u8;
```

This function returns a new kind of memory — `MiriMemoryKind::Physical`, which is allowed to leak.

A corresponding adapted version of the above example that runs successfully with this PR looks like this (also included as test):

```rust
extern "Rust" {
    fn miri_alloc_phys(size: usize, phys_addr: usize) -> *mut u8;
}

pub const PHYS_ADDR: usize = 0xB8000;

fn main() {
    unsafe {
        let ptr = miri_alloc_phys(2, PHYS_ADDR);
        ptr.write_volatile(0u8);
        ptr.add(1).write_volatile(0u8);
    }
}
```

## Alternatives

Another name for this function to consider is `miri_map_phys` to reference the conceptual mapping of physical memory to virtual memory.

I think that allocate is the more fitting verb to use here. Though allocating memory instead of merely mapping it is more restrictive to the interpreted program, mapping would require us to do much more behind the scenes to keep track of mappings that might alias. The downside of this allocation semantic instead of the map semantic is that applications without central allocation of physical memory are more difficult to adapt. One might have to introduce a central manager who would allow easy drop-in aliasing. Still, I would think that this would be easier and more useful than building a fully fledged notion of physical memory with aliasing into Miri. 

## Future improvements

Currently, the provided `phys_addr` is unused, though it might be useful in the future:

If we stick to the _allocate once_ approach instead of the _map potentially many times_ approach, we should add a check on whether the physical memory has been allocated already by tracking physical allocations. This is currently omitted to keep this PR short, but can either be addressed in the future or in this PR if wanted.

In the future, we might want to introduce APIs to be able to plug software into Miri that emulates the behavior of active devices. The current design of the `miri_alloc_phys` API would be enough to support this without changes to the interpreted software, as far as I can tell.


oli-obk:
> Currently, the provided `phys_addr` is unused, though it might be useful in the future:

I think this is a problem beyond the "allocate same address multiple times" case you describe.

The physical address does not reserve that memory region, so there's no way to magically go from the address to the memory without preserving provenance from the `miri_alloc_phys` call. Basically `assert_eq!(miri_alloc_phys(42, 1) as usize, 42)` will fail almost always.

Can you elaborate why the current behaviour of the function you added is useful for testing applications that use MMIO?

mkroening:
> Basically `assert_eq!(miri_alloc_phys(42, 1) as usize, 42)` will fail almost always.

That's fine, though. That is the same behavior as the motivating example has if there is no identity mapping. On real hardware, it is fine to map `VirtAddr(4096 + 42)` to `PhysAddr(42)` and then cast `4096 + 42` to a pointer and use that. The offset of 4096 here is one page-size on x86_64, but the possible offsets are arch-dependent and conceptually arbitrary, I think.

There is no proper way to retrieve the same pointer again if you lost it. It is possible to map different virtual addresses to the same physical address, effectively aliasing it and making the reverse function going from physical to virtual ambiguous.

That's why we could either introduce this aliasing in Miri or warn/error on "allocating" the same physical address multiple times.

So either

```rust
/// A fully-fledged, aliasing paging emulation in miri
let ptr1 = miri_map_phys(42, 1);
let ptr2 = miri_map_phys(42, 1);
*ptr1 = 5;
assert_eq!(*ptr2, 5);
```

or

```rust
/// Warn/error on double allocation
let ptr1 = miri_alloc_phys(42, 1);
let ptr2 = miri_alloc_phys(42, 1); // This throws an error or warning
*ptr1 = 5;
assert_ne!(*ptr2, 5);
```

> Can you elaborate why the current behaviour of the function you added is useful for testing applications that use MMIO?

This makes applications run that only use passive devices that don't change memory, like the example. As of now, the only differences to using `miri_alloc` is that the memory is allowed to leak and that applications can supply the physical target address, which could right now be used to detect aliasing if we decide for that semantic. 

Future changes to support active devices would happen behind the scenes and would not require changes to this interpreted-program-facing API, I think.

mkroening:
Another possibility which would be an even more extreme and sophisticated paging implementation would be to allow choosing virtual addresses:

```rust
fn miri_map(virtual_address: usize, physical_address: usize);
```

This would be difficult regarding the virtual addresses that Miri's uses in other ways, as @saethlin noticed. Also, provenance would be more difficult, since casting to a pointer would still be necessary.

saethlin:
I was just entirely unaware of the use case you were aiming for. I think there are two hard aspects to doing the physical MMIO in Miri that Oli and I were thinking of:

* The provenance and address space issues we both raised
* Figuring out how to connect the memory system to extra logic

It sounds to me like what you're trying to support will deal with the second point without also getting tangled up in the first one. That sounds good to me.

oli-obk:
Yea, after reading this explanation and the zulip thread, this all makes a lot of sense now. I think we should still keep a map of the physical address space and raise a UB error if two `miri_alloc_phys` start overlapping.

btw, how is this kind of thing done on linux, mac or windows usually? Are there real APIs for this that we could emulate instead?

RalfJung:
If `miri_alloc_phys` is intended to return the *virtual* address mapping to that physical address, the docs should clearly say that.

But in that case I am also confused about its purpose. Miri / the Rust AM has no concept of physical vs virtual addresses. That's just the one address space that code runs in. In your mental model that seems to be the virtual address space. This then means that physical addresses are entirely unobservable. One can implement miri_alloc_phys "in userspace", without a Miri primitive, by just calling miri_alloc. What is the point of having a new primitive here?

mkroening:
I implemented the check and added a fail test for overlapping memory. I raised an "unsupported" error instead of an UB error, since this would not be UB, I think.

Furthermore, I reworded the docs of `miri_alloc_phys` and swapped the argument order. Hopefully, the docs are clearer now.

> What is the point of having a new primitive here?

The purpose of this is to have a path forward for emulated MMIO devices. Currently, the only difference to `miri_alloc` is that physical addresses are checked not to overlap and that the memory may leak.

Once we want to emulate devices in the future, we need physical addresses to recognize what device is accessed. This primitive is meant for apps to specify which device they access via the pointer via the device's physical addresses.

Another useful property of having a new memory type that I just thought of is making sure that the memory is only accessed via volatile operations.

Of these four points (device identification, volatile-ness, no overlapping, leakability), only device identification (and handling) is out of scope for Miri and must be provided externally. It might make sense to have this primitive declared centrally by Miri. It might also make sense to leave providing this primitive to external crates. I am not sure myself.

RalfJung:
What I'm worried about here is that we have no clear idea for what any of those features look like. To review if this makes any sense, I think we need a clear design description of what the final state will look like. I'd rather not spend a lot of time to review and maintain a bunch of experiments that don't go anywhere. This PR on its own doesn't achieve any of these goals, does it?

We currently don't have a Major Change Proposal or RFC process for Miri, but I think for changes like this we need one.

> Another useful property of having a new memory type that I just thought of is making sure that the memory is only accessed via volatile operations.

AFAIK MMIO doesn't always require volatile, in particular when DMA is involved? Sometimes you write a descriptor of sorts to some memory, then have a fence of sorts, and then store a pointer to that descriptor using a volatile write, thereby publishing the entire thing to the device. At least that's what I gathered in prior discussions around this subject, I am far from an expert here.^^

> Once we want to emulate devices in the future, we need physical addresses to recognize what device is accessed. This primitive is meant for apps to specify which device they access via the pointer via the device's physical addresses.

In real code they would write to the physical address though, wouldn't they? This part about having a separate virtual address is completely a Miri fiction?

I don't know what the best way to support MMIO in Miri is, but I feel like it should involve actually making the desired devices available at their actual addresses. As a very first idea, I could imagine (a) a Miri flag to control in which address range Miri puts stack and heap allocations (either by excluding certain regions or by saying "only use this region -- not sure which one makes more sense), and then (b) a way for programs to register physical devices at a given address. The (b) part is where it gets really interesting, and I don't have enough domain knowledge to say which approach is more suitable:
- We could imagine that there is a limited set of devices that Miri has hard-coded support for, similar to out "shims" that hard-code support for extern functions and statics. Then registering a device might be as simple as `miri_create_device(DEVICE_ID, DEVICE_ADDRESS)`.
- Or maybe that is a bad idea since devices are just too diverse in behavior.  Then we need to support "custom devices" somehow. In the extreme case, the API is like `miri_create_device(addr, size, read_fn, write_fn)` and each volatile access in that memory region leads to the read_fn/write_fn being called to determine what happens. Conveniently volatile accesses are already intrinsics to making them push new stack frames to the call stack should actually be fairly simple to implement...

Either way I think we need a clearer idea for where we want to go with this before we can start landing the first parts of it.

mkroening:
> This PR on its own doesn't achieve any of these goals, does it?

This PR is usable for passive devices and has benefits over using other methods, but verifying accesses and responding is not part of this PR. I agree that things should perhaps be more designed and tested around that before merging anything into Miri. I thought, this change as is would be the starting building block without many concerns around the design, but that has not turned out to be the case. I'll close this PR, but I am open for further discussion and will hopefully be back in summer with a more complete design proposal. This initial attempt and the surrounding discussion has been very helpful, though. Thank you all. I might also reach out to you in the future for discussing any approach, @RalfJung.

> AFAIK MMIO doesn't always require volatile, in particular when DMA is involved? Sometimes you write a descriptor of sorts to some memory, then have a fence of sorts, and then store a pointer to that descriptor using a volatile write, thereby publishing the entire thing to the device.

That does not collide with my view on things. The descriptor write is the volatile operation acting on MMIO memory. The referenced memory does not have to be special.

> In real code they would write to the physical address though, wouldn't they? This part about having a separate virtual address is completely a Miri fiction?

No, that's why I designed `miri_alloc_phys` like this. On a real x86_64 processor, for example, the VGA text buffer might be at the physical address `0xB8000`. I can now have any virtual address map to that address and would then write to that virtual address to access the MMIO memory. I might choose to do an identity mapping, or I might be on a device without paging, but in general the pointer address I use to access the device is independent of the physical address.

So no, generally this is not fiction at all. Supplying a physical address to Miri is still important, though, so Miri knows which device a pointer belongs to.

> I don't know what the best way to support MMIO in Miri is, but I feel like it should involve actually making the desired devices available at their actual addresses.

We'd have to decide if we want to support phys_addr -> ptr casts for MMIO, which imply identity mapping, or go for the more general approach that I presented. I went for the more general approach since it fits naturally with my experience on x86_64 with paging. Assuming identity mapping and allowing phys_addr -> ptr casts might require fewer changes for code to run on Miri, but does not feel quite *right* for me for my use case.

> (b) a way for programs to register physical devices at a given address

This might be difficult to do, since in principle, devices might be anywhere in the address space. Registering them after the program has started might cause some collisions, I think.

Regarding b.1: Emulating devices directly in Miri would slowly mutate Miri into some device emulator collection. I don't think that would be good.


I'll close this now since I don't see this landing soon. I'll be happy to continue the discussion though and hopefully open a new PR sometime. Thank you all for participating again! :)

Regarding b.2: That would move emulation into the guest program, right? Very interesting thought, but not really what we want, I think.

My approach would be to create a new frontend for Miri, which drives Miri as a library. That project would then emulate devices depending on invocation parameters. No registering inside the guest program except for `miri_alloc_phys` or similarly a phys_addr -> ptr cast.

RalfJung:
> No, that's why I designed miri_alloc_phys like this. On a real x86_64 processor, for example, the VGA text buffer might be at the physical address 0xB8000. I can now have any virtual address map to that address and would then write to that virtual address to access the MMIO memory. I might choose to do an identity mapping, or I might be on a device without paging, but in general the pointer address I use to access the device is independent of the physical address.

I see. I thought people would usually set up an identity mapping somewhere anyway and then it's easier to use that than set up a new mapping for each device, but then I don't actually do any work in this area so that might just be a misconception.

> This might be difficult to do, since in principle, devices might be anywhere in the address space. Registering them after the program has started might cause some collisions, I think.

That's why I suggested flags that control where Miri puts its allocations, so that these regions remain free and there won't be any collisions.

> My approach would be to create a new frontend for Miri, which drives Miri as a library. That project would then emulate devices depending on invocation parameters. No registering inside the guest program except for miri_alloc_phys or similarly a phys_addr -> ptr cast.

Ah I see. That is quite a big departure from how Miri is used so far, and I think would incur significant overhead on your size (Miri links to rustc so it's not the easiest Rust program to get building, and it can't just be a crates.io library).

Moving emulation into the guest program was meant to basically substitute that -- technically yes it would be the guest program, but conceptually the `main` function that registers these devices would be the "driver" you mention, which would run before handing control to the actual guest program. Then you could use the Miri that is distributed with rustup, making everything a lot easier.

mkroening:
> I see. I thought people would usually set up an identity mapping somewhere anyway and then it's easier to use that than set up a new mapping for each device, but then I don't actually do any work in this area so that might just be a misconception.

That's a proper argument. My experience is limited as well. But in principle, the virtual address can be anything.

> That's why I suggested flags that control where Miri puts its allocations, so that these regions remain free and there won't be any collisions.

Ah, that makes sense. I misread and thought you proposed either (a) *or* (b), not (a) *and* (b).

> the `main` function that registers these devices would be the "driver"

I see. That's a very good idea. I was cautious to depart too much from a real program, but thinking about it, I really like that approach!

PR #2770, opened 2023-01-27
Do the tokio tests work with the default preemption rate?

DebugSteven:
This is a test to see if CI will pass without the `-Zmiri-preemption-rate=0` flag. 

The `tokio_mvp.rs` test never failed for me locally, but this flag was needed for that test to pass CI. Since this branch has an implementation for sleep I want to see if the preemption flag is still needed or what CI will fail with.

There's a different [open PR](https://github.com/rust-lang/miri/pull/2764) already being reviewed for sleep.


DebugSteven:
Rad! The experiment is a success.

PR #2782, opened 2023-02-09
Rustup

RalfJung:


RalfJung:
Address space exhaustion on the AVR test. I guess if you only have 64 KiB we shouldn't skip half the address space for allocations...

@bors r+

PR #2784, opened 2023-02-11
Add tests for moving data across await point

bryangarza:
This patch adds a few tests to assert the current behavior when passing data across an await point. This will help to test out an upcoming fix for the issue of arguments in async functions growing in size because of the generator upvar that is generated when we desugar the async function.

See https://github.com/rust-lang/rust/issues/62958

Also relates to https://github.com/rust-lang/rust/pull/107500

FYI @oli-obk @pnkfelix

### DIFF
--- a/tests/pass/move-data-across-await-point.rs
+++ b/tests/pass/move-data-across-await-point.rs
@@ -1,0 +1,81 @@
+use std::future::Future;
+use std::ptr;
+
+// This test:
+// - Compares addresses of non-Copy data before and after moving it
+// - Writes to the pointer after it has moved across the await point
+//
+// This is only meant to assert current behavior, not guarantee that this is
+// how it should work in the future. In fact, upcoming changes to rustc
+// *should* break these tests.
+// See: https://github.com/rust-lang/rust/issues/62958
+async fn data_moved_async() {
+    async fn helper(mut data: Vec<u8>, raw_pointer: *mut Vec<u8>) {
+        let raw_pointer2 = ptr::addr_of_mut!(data);
+        // `raw_pointer` points to the original location where the Vec was stored in the caller.
+        // `data` is where that Vec (to be precise, its ptr+capacity+len on-stack data)
+        // got moved to. Those will usually not be the same since the Vec got moved twice
+        // (into the function call, and then into the generator upvar).
+        assert_ne!(raw_pointer, raw_pointer2);
+        unsafe {
+            // This writes into the `x` in `data_moved_async`, re-initializing it.
+            std::ptr::write(raw_pointer, vec![3]);
+        }
+    }
+    // Vec<T> is not Copy
+    let mut x: Vec<u8> = vec![2];
+    let raw_pointer = ptr::addr_of_mut!(x);
+    helper(x, raw_pointer).await;
+    unsafe {
+        assert_eq!(*raw_pointer, vec![3]);
+        // Drop to prevent leak.
+        std::ptr::drop_in_place(raw_pointer);
+    }
+}
+
+// Same thing as above, but non-async.
+fn data_moved() {
+    fn helper(mut data: Vec<u8>, raw_pointer: *mut Vec<u8>) {
+        let raw_pointer2 = ptr::addr_of_mut!(data);
+        assert_ne!(raw_pointer, raw_pointer2);
+        unsafe {
+            std::ptr::write(raw_pointer, vec![3]);
+        }
+    }
+
+    let mut x: Vec<u8> = vec![2];
+    let raw_pointer = ptr::addr_of_mut!(x);
+    helper(x, raw_pointer);
+    unsafe {
+        assert_eq!(*raw_pointer, vec![3]);
+        std::ptr::drop_in_place(raw_pointer);
+    }
+}
+
+fn run_fut<T>(fut: impl Future<Output = T>) -> T {
+    use std::sync::Arc;
+    use std::task::{Context, Poll, Wake, Waker};
+
+    struct MyWaker;
+    impl Wake for MyWaker {
+        fn wake(self: Arc<Self>) {
+            unimplemented!()
+        }
+    }
+
+    let waker = Waker::from(Arc::new(MyWaker));
+    let mut context = Context::from_waker(&waker);
+
+    let mut pinned = Box::pin(fut);
+    loop {
+        match pinned.as_mut().poll(&mut context) {
+            Poll::Pending => continue,
+            Poll::Ready(v) => return v,
+        }
+    }
+}
+
+fn main() {
+    run_fut(data_moved_async());
+    data_moved();
+}


tests/pass/move-data-across-await-point.rs
```diff
@@ -0,0 +1,78 @@
...
+    let raw_pointer2 = &mut data as *mut Vec<u8>;
+    // Addresses are different because the generator upvar for `data` is a copy.
+    // To be more precise, there is a `move` happening in the MIR of the
+    // generator closure, which copies the pointer.
+    //
+    // When copy propagation is enabled for generator upvars, the pointer addresses
+    // here should be equal.
+    assert_ne!(raw_pointer, raw_pointer2);
+    unsafe {
+        std::ptr::write(raw_pointer, vec![3]);
```
RalfJung, line 28:
```suggestion
        // This writes into the `x` in `data_moved_async`, re-initializing it.
        std::ptr::write(raw_pointer, vec![3]);
```

tests/pass/move-data-across-await-point.rs
```diff
@@ -0,0 +1,78 @@
...
+    let raw_pointer = &mut x as *mut Vec<u8>;
+    helper_async(x, raw_pointer).await;
+    unsafe {
+        assert_eq!(*raw_pointer, vec![3]);
+        // Drop to prevent leak.
+        std::ptr::drop_in_place(raw_pointer);
+    }
+}
+
+async fn helper_async(mut data: Vec<u8>, raw_pointer: *mut Vec<u8>) {
```
RalfJung, line 18: Please move this function to inside `data_moved_async`, they seem tightly coupled.

RalfJung:
Thanks! These are interesting tests.

However, I don't think we are *guaranteeing* that this works.  Specifically, the backing store for `x` could be deallocated earlier.

tests/pass/move-data-across-await-point.rs
```diff
@@ -0,0 +1,78 @@
...
+    let raw_pointer2 = &mut data as *mut Vec<u8>;
+    // Addresses are different because the generator upvar for `data` is a copy.
+    // To be more precise, there is a `move` happening in the MIR of the
+    // generator closure, which copies the pointer.
```
RalfJung, line 22: "the pointer"? It copies the `Vec`, right?

oli-obk:
> However, I don't think we are _guaranteeing_ that this works.

In fact, these tests are only for checking the current behaviour so that when we inevitably (as re-using these memory slots is being worked on) break these tests, that does not happen silently. We really want to break these tests, as they can cause immense wastes of space if enough async function calls are nested.

RalfJung:
If we want to break them we have to first make them UB...

But yeah that's what I thought. So there should be a comment explaining this.

oli-obk:
> If we want to break them we have to first make them UB...

well... the equivalent non-async code is UB already.

> But yeah that's what I thought. So there should be a comment explaining this.

agreed

RalfJung:
> well... the equivalent non-async code is UB already.

That doesn't seem true? This PR adds a non-async version as well and it is a 'pass' test.

tests/pass/move-data-across-await-point.rs
bryangarza: What I was trying to explain is that the ptr+capacity+len that is stored on the stack for that Vec is copied to the generator upvar so that's why `raw_pointer2` has a different address than `raw_pointer`. I'll see if there's a better way to phrase it in the comment

tests/pass/move-data-across-await-point.rs
```diff
@@ -0,0 +1,84 @@
...
+// This is only meant to assert current behavior, not guarantee that this is
+// how it should work in the future. In fact, upcoming changes to rustc
+// *should* break these tests.
+// See: https://github.com/rust-lang/rust/issues/62958
+async fn data_moved_async() {
+    async fn helper(mut data: Vec<u8>, raw_pointer: *mut Vec<u8>) {
+        let raw_pointer2 = &mut data as *mut Vec<u8>;
+        // When the Vec is moved into the generator upvar, that copies the Vec's
+        // ptr+capacity+len that is on the stack. Because of that, the value of
+        // `raw_pointer2` is different than that of `raw_pointer`.
```
RalfJung, line 16: I am confused by the mention of upvars here.  Just passing it to `helper` moves it to a new stack slot, right?

JakobDegen: Yes. It makes sense if you think of `helper(foo)` just being shorthand for a suitable `StatementKind::Aggregate` that creates the generator (I believe in practice creating the generator is indeed behind a function)

RalfJung, line 16:
```suggestion
        // `raw_pointer` points to the original location where the Vec was stored in the caller.
        // `data` is where that Vec (to be precise, its ptr+capacity+len on-stack data)
        // got moved to. Those will usually not be the same since the Vec got moved twice
        // (into the function call, and then into the generator upvar).
```

tests/pass/move-data-across-await-point.rs
```diff
@@ -0,0 +1,84 @@
...
+        //
+        // When copy propagation is enabled for generator upvars, the pointer
+        // addresses should be equal, becouse there will no longer be a
+        // generator upvar; instead, the function's `data` argument will be
+        // used directly.
```
RalfJung, line 21: The function's `data` argument is still different from where `raw_pointer` points to though -- it points to the caller's `x`. So they would still be inequal I think?

tests/pass/move-data-across-await-point.rs
```diff
@@ -0,0 +1,84 @@
...
+        assert_ne!(raw_pointer, raw_pointer2);
+        unsafe {
+            // This writes into the `x` in `data_moved_async`, re-initializing it.
+            std::ptr::write(raw_pointer, vec![3]);
+        }
+    }
+    // Vec<T> is not Copy
+    let mut x: Vec<u8> = vec![2];
+    let raw_pointer = &mut x as *mut Vec<u8>;
```
RalfJung, line 30:
```suggestion
    let raw_pointer = ptr::addr_of_mut!(x);
```
Also needs a `use std::ptr;`. Should be done for all the `&mut _ as *mut _` in this file.

pnkfelix:
To be honest, the kinds of corner cases that I think I'm worried about for https://github.com/rust-lang/rust/issues/62958 are more things that can only be expressed either in MIR directly or via things like:

```rust
fn manual_async_fn(param: MaybeBig) -> impl Future {
    async {
        let local = param;
        let borrow_the_upvar = &param as *const _;
        // uh oh, it is probably a bad idea to have `local` occupy same storage as`param` *now*
    }
}
```

I do not think you'll be able to express cases like the above via a normal `async fn`, but they are the cases I am worried about mishandling in my analysis and transformation.

(Furthermore, I do not think the generator upvar optimizations I am writing will actually attempt to be so aggressive as to do the kind of coalescing that is described in the test written here. But then again, I haven't tried it out on the example yet, so maybe I will end up surprising myself.)

tests/pass/move-data-across-await-point.rs
bryangarza: oh, yeah I think you are right. `data` itself is stored in another place than `x` even though the underlying Vec in the heap is the same. I'll remove that comment

bryangarza:
@pnkfelix 
> To be honest, the kinds of corner cases that I think I'm worried about for [rust-lang/rust#62958](https://github.com/rust-lang/rust/issues/62958) are more things that can only be expressed either in MIR directly or via things like:
> 
> ```rust
> fn manual_async_fn(param: MaybeBig) -> impl Future {
>     async {
>         let local = param;
>         let borrow_the_upvar = &param as *const _;
>         // uh oh, it is probably a bad idea to have `local` occupy same storage as`param` *now*
>     }
> }
> ```
> 
> I do not think you'll be able to express cases like the above via a normal `async fn`, but they are the cases I am worried about mishandling in my analysis and transformation.

Gotcha -- we can do a separate PR for the kind of tests you described above.

> (Furthermore, I do not think the generator upvar optimizations I am writing will actually attempt to be so aggressive as to do the kind of coalescing that is described in the test written in this Pull Request. But then again, I haven't tried my prototype out on the example yet, so maybe I will end up surprising myself.)

FYI, I ran these new miri tests on [your branch](https://github.com/pnkfelix/rust/tree/copy-prop-upvar-to-elim-local-for-62958) of rustc, and they pass  as you expected 🙂

PR #2785, opened 2023-02-13
Tree borrows

Vanille-N:
This PR implements the experimental Tree Borrows (TB) rules for tracking reference aliasing, as an optional alternative to Stacked Borrows (SB).

SB and TB are mutually exclusive. Using `-Zmiri-tree-borrows` replaces every invocation of SB with the equivalent TB procedure.

A detailed explanation of the TB rules is under review and not included here ;-)

This PR does NOT include
- enough `fail` tests for TB (although TB is less reliant than SB on `fail` tests to check that the implementation matches the design due to `pass` tests being more strict)
- good diagnostics for TB violations


src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -0,0 +1,548 @@
...
+// In the latter case any protector would not trigger.
+// FIXME: rethink this
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(super) struct AccessChk<T> {
+    /// The wrapped value
+    pub data: T,
+    /// Whether the value was ever accessed
+    pub accessed: bool,
+    /// Latest transition that was applied to this node:
+    /// Some(b) if foreign, with b=true for foreign write and b=false for foreign read;
```
RalfJung, line 26: This sounds like it should use an `enum`, not `bool`. We do have an `AccessKind` enum somewhere.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -0,0 +1,548 @@
...
+                                perm.latest_applied_was_write = Some(T::IS_WRITE);
+                            }
+                        } else {
+                            perm.latest_applied_was_write = None;
+                        }
+                        let old_perm = *perm;
+                        let new_perm = T::after_access(rel_pos, old_perm.data, protected)?;
+                        if old_perm.accessed {
+                            protector_allows_transition(global, node.tag, old_perm.data, new_perm)?
+                        }
```
RalfJung, line 393: Surprising that `protected` is not relevant here.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -0,0 +1,548 @@
...
+}
+
+/// Integration with the BorTag garbage collector
+impl Tree {
+    pub fn remove_unreachable_tags(&mut self, live_tags: &FxHashSet<BorTag>) {
+        assert!(self.keep_only_needed(self.root, live_tags)); // root can't be removed
+    }
+
+    /// Traversal function: returns true iff the tag is necessary. A tag is necessary
+    /// if it is either live or it has necessary children.
```
RalfJung, line 426: Judging from the name this doesn't just return a bool though it also has some side-effects that might be relevant...

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -0,0 +1,548 @@
...
+    fn keep_only_needed(&mut self, idx: UniIndex, live: &FxHashSet<BorTag>) -> bool {
+        let node = self.nodes.get(idx).unwrap();
+        let children: SmallVec<_> = node
+            .children
+            .clone()
+            .into_iter()
```
RalfJung, line 432: `.iter().cloned()` is probably more efficient as it avoids creating an intermediate `SmallVec`. Or can this even be `.copied()`? If not what is being cloned here?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -0,0 +1,548 @@
...
+            .filter(|child| self.keep_only_needed(*child, live))
+            .collect();
+        let no_children = children.is_empty();
+        let node = self.nodes.get_mut(idx).unwrap();
+        node.children = children;
+        // All of the children are unreachable
+        if !live.contains(&node.tag) && no_children {
```
RalfJung, line 439:
```suggestion
        if !live.contains(&node.tag) && no_children {
          // All of the children and this node are unreachable.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -0,0 +1,548 @@
...
+            // Important: make sure to remove _all_ of them because we're going to forget the index
+            // afterwards.
```
RalfJung, line 441: Mysterious "them"

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -0,0 +1,548 @@
...
+            false
+        } else {
+            true
+        }
+    }
+}
+
+impl VisitTags for Tree {
+    fn visit_tags(&self, visit: &mut dyn FnMut(BorTag)) {
+        visit(self.nodes.get(self.root).unwrap().tag)
```
RalfJung, line 457: Curious, why specifically and only the root?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -0,0 +1,548 @@
+use smallvec::SmallVec;
```
RalfJung, line 1: Please start the file with a doc comment summarizing the key things implemented here

RalfJung (commented):
I didn't get to the tests yet, but here's the comments for the code side of things. It's a lot of comments but they are mostly minor -- overall this looks great!

RalfJung (commented):
I didn't get to the tests yet, but here's the comments for the code side of things. It's a lot of comments but they are mostly minor -- overall this looks great!

tests/compiletest.rs
```diff
@@ -141,6 +141,7 @@ regexes! {
     r"\\"                           => "/",
     // erase Stacked Borrows tags
```
RalfJung, line 142:
```suggestion
    // erase borrow tags
```

tests/compiletest.rs
```diff
@@ -151,6 +152,7 @@ regexes! {
     "alloc[0-9]+"                    => "ALLOC",
     // erase Stacked Borrows tags
```
RalfJung, line 153:
```suggestion
    // erase borrow tags
```

tests/fail/tree-borrows/outside-range.rs
```diff
@@ -0,0 +1,21 @@
...
+    }
+}
+
+unsafe fn ok(x: &mut u64) {
+    let _val = *(x as *mut u64).add(1);
+}
+
+unsafe fn ko(x: &mut u64, y: *mut u64) {
+    let _val = *(x as *mut u64).add(1);
+    *y = 100; //~ ERROR: Protector
```
RalfJung, line 20: This (and other) fail tests also needs a comment at the top explaining what is being tested.

tests/fail/tree-borrows/propagation/unknown-size-protected.stderr
```diff
@@ -0,0 +1,59 @@
...
+|----| Re*|      └─┬──<TAG=data>
+|----| Re*|        ├─┬──<TAG=u>
+|----| Re*|        │ └────<TAG=arg:u> Strongly protected
+|----|?Re*|        └────<TAG=v,arg:v>
+──────────────────────────────────────────────────────────────────────
+error: Undefined Behavior: Protector triggered on <TAG> whose permissions were updated from Permission(Reserved { ty_is_freeze: false }) to Permission(Disabled). Caused by a foreign Write through <TAG>.
+  --> $DIR/../../../utils/tb/mod.rs:LL:CC
+   |
+LL |             *ptr.add($offset) = 0;
+   |             ^^^^^^^^^^^^^^^^^^^^^ Protector triggered on <TAG> whose permissions were updated from Permission(Reserved { ty_is_freeze: false }) to Permission(Disabled). Caused by a foreign Write through <TAG>.
```
RalfJung, line 41: What about

"write access through \<TAG> is forbidden because it is a foreign tag for \<TAG>, which would hence change from reserved to disabled, but \<TAG> is protected"

tests/fail/tree-borrows/reserved/int-protected-read.rs
```diff
@@ -0,0 +1,31 @@
...
+            name!(x, "callee:x");
+            name!(x=>1, "caller:x");
+            name!(y, "callee:y");
+            name!(y, "caller:y");
+            access!(Read y);
+            fmt_tree!(alloc_id);
+            access!(Write x); //~ ERROR: Impossible to perform a child Write access
+        }
+        // We expect callee:x to become Frz to respect [noalias]
+        // x and caller:x can stay Res
```
RalfJung, line 29:
```suggestion
            // We expect callee:x to become Frz to respect [noalias]:
            // noalias can reorder the read above and the write below, but they alias,
            // so this program must be UB.
            // x and caller:x can stay Res.
            access!(Write x); //~ ERROR: Impossible to perform a child Write access
        }
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -0,0 +1,548 @@
...
+            // as input to the children calls.
+            while let Some((parent_desc, next_in_path)) = path_ascend.pop() {
+                // next_in_path is excluded from here because it will be handled
+                // in the next iteration of the loop. It also needs a different
+                // AccessRelatedness.
+                stack.exec_and_visit(
+                    &mut self,
+                    parent_desc,
+                    Some(next_in_path),
+                    AccessRelatedness::StrictChildAccess,
```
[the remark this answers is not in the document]
Vanille-N: Hehe it actually meant "desc"ending...

src/shims/foreign_items.rs
```diff
@@ -434,6 +434,32 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
                 }
             }
+            "miri_print_borrow_tree" => {
+                // Just print the tags and the states in a nice way
+                let [id, only_named] = this.check_shim(abi, Abi::Rust, link_name, args)?;
+                // This is the alloc id
+                let id = this.read_scalar(id)?.to_u64()?;
+                let only_named = this.read_scalar(only_named)?.to_bool()?;
+                if let Some(id) = std::num::NonZeroU64::new(id) {
+                    this.tb_print_tree(AllocId(id), only_named)?;
```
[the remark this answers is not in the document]
Vanille-N: Yes, and `print_stacks` also goes wrong when TB runs.

How should I handle the extra argument ?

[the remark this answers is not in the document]
RalfJung: I didn't realize the signature is different.... not sure. We could just say that it is ignored under SB?

src/borrow_tracker/tree_borrows/tree.rs
Vanille-N: Right now `.iter().clone()` doesn't compile as it blocks the recursive mutable borrow. I'll try but it will require restructuring the entire function to make two passes instead of one.

src/borrow_tracker/tree_borrows/tree.rs
Vanille-N: Because I have taken no steps to handle a tree that doesn't have a root. `node.parent` is an `Option<tag>`, but `tree.root` is a `Tag`.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -0,0 +1,570 @@
...
+use crate::*;
+
+mod diagnostics;
+mod perms;
+mod tree;
+mod unimap;
+use perms::Permission;
+pub use tree::Tree;
+
+// FIXME: Tag gc integration should probably be in _this_ file ?
```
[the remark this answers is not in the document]
Vanille-N: It's in `tree.rs`, but SB has it in `mod.rs`

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: I think you skipped the part of the answer I was looking for. ;)

You want to ensure that the root never get GC'd, right? If so then please add a comment here explaining that.

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: Oh I see, `self` aliases with `node.children`...

In that case you can also leave a comment noting to explore in the future whether there's some way to optimize this and avoid the `clone`. If you do try the 2-pass approach, make sure to benchmark to check if it actually is faster.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -0,0 +1,548 @@
...
+            for perms in self.rperms.iter_mut_all() {
+                perms.remove(idx);
+            }
+            self.tag_mapping.forget(&tag);
+            false
+        } else {
+            true
+        }
+    }
+}
```
RalfJung, line 453: Might be worth noting somewhere that this leave cleanup potential on the table: if a dead tag is in the middle of the tree, we might still be able to "compact" the tree and re-assign its children to directly link to its parent. It's subtle though since we must make sure that this doesn't lead to any UB being missed.

src/borrow_tracker/tree_borrows/mod.rs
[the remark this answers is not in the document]
RalfJung: No need to religiously follow the SB structure.

tests/utils/tb/shims.rs
```diff
@@ -0,0 +1,8 @@
+#![allow(dead_code)]
+
+extern "Rust" {
+    pub fn miri_get_alloc_id(ptr: *const u8) -> u64;
+    pub fn miri_print_borrow_tree(alloc_id: u64, only_named: bool);
+    pub fn miri_pointer_name(ptr: *const u8, nth_parent: u8, name: &[u8]);
+    pub fn miri_write_to_stderr(msg: &[u8]);
```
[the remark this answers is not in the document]
Vanille-N: Should I also make it so that the TB tests call these magic functions from the same place ?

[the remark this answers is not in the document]
RalfJung: Not quite sure what you mean... the tests should call the functions by importing `miri_extern`. Eventually we'll want to also adjust the existing tests to do that but that can be a separate PR.

tests/pass/box.rs
```diff
@@ -1,3 +1,5 @@
+//@revisions: stack tree
+//@[tree]compile-flags: -Zmiri-tree-borrows
```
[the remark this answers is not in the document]
oli-obk: this is still unaddressed (there are also other occurrences of the same situation)

tests/pass/tree-borrows/cell/cell-alternate-writes.rs
```diff
@@ -0,0 +1,27 @@
+//@compile-flags: -Zmiri-tree-borrows -Zmiri-tag-gc=0
+#[path = "../../../utils/mod.rs"]
+mod utils;
+use utils::macros::*;
```
oli-obk, line 4: could also add a crate and make it a dependency of https://github.com/rust-lang/miri/blob/master/test_dependencies 

that way we can reuse it in all tests easily without having to fiddle with path imports

tests/pass/box.rs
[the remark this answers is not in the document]
Vanille-N: But then the test fails with "error: unsupported operation: integer-to-pointer casts and `ptr::from_exposed_addr` are not supported with `-Zmiri-strict-provenance`"

Did I misunderstand the comment ? At least the "suggested change" by itself is not sufficient.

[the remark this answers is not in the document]
RalfJung: Sorry, I meant `-Zmiri-permissive-provenance`

src/borrow_tracker/tree_borrows/unimap.rs
```diff
@@ -27,9 +27,7 @@ pub struct UniKeyMap<K> {
     mapping: FxHashMap<K, usize>,
     /// Indexes that can be reused: memory gain when the map gets sparse
     /// due to many deletions.
-    unassigned: Vec<usize>,
-    /// Number of assigned indexes, i.e. next unique identifier to assign.
-    capacity: usize,
+    deassigned: Vec<usize>,
```
RalfJung: Needs a comment explaining the key invariant: that the indices are contiguous, 0..N, except for what is deassigned

src/bin/miri.rs
```diff
@@ -315,7 +315,9 @@ fn main() {
         } else if arg == "-Zmiri-disable-stacked-borrows" {
             miri_config.borrow_tracker = None;
         } else if arg == "-Zmiri-tree-borrows" {
-            miri_config.borrow_tracker = Some(BorrowTrackerMethod::TreeBorrows);
+            if miri_config.borrow_tracker.is_some() {
+                miri_config.borrow_tracker = Some(BorrowTrackerMethod::TreeBorrows);
+            }
```
RalfJung, line 320: I think this should just always set borrow_tracker.

The docs should explain that disable-stacked-borrows disables both of them (but later flags take precedence)

tests/pass/tree-borrows/propagation/nontrivial-structure.rs
```diff
@@ -1,44 +1,27 @@
...
-        name!(xcab);
-        let xcba = child!(xcb => &);
-        name!(xcba);
-        let xcbb = child!(xcb => &);
-        name!(xcbb);
-        let alloc_id = alloc_id!(x);
-        fmt_tree!(alloc_id);
-    }
-}
+// Pushing the limits of the formating function that prints the tree structure
```
RalfJung, line 6: Does this test formatting or propagataion (the latter is indicated by the folder name)?

Sounds like this could be merged with formatting.rs.

tests/pass/tree-borrows/formating.rs
```diff
@@ -1,23 +1,26 @@
 //@compile-flags: -Zmiri-tree-borrows -Zmiri-tag-gc=0
 #![feature(extern_types)]
-#[path = "../../utils/mod.rs"]
+#[path = "../../utils.rs"]
 mod utils;
-use utils::tb::num::*;
+use utils::*;
 
+// Alignment check for the formating function: we split the array
```
RalfJung, line 7: "formatting" (also in the filename)

tests/utils.rs
```diff
@@ -0,0 +1,47 @@
...
+
+#[allow(dead_code)]
+extern "Rust" {
+    pub fn miri_get_alloc_id(ptr: *const u8) -> u64;
+    pub fn miri_print_borrow_tree(alloc_id: u64, only_named: bool);
+    pub fn miri_pointer_name(ptr: *const u8, nth_parent: u8, name: &[u8]);
+    pub fn miri_write_to_stderr(msg: &[u8]);
+}
+
+macro_rules! alloc_id {
```
RalfJung: These should all have doc comments -- this is a library, after all

src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -105,7 +105,7 @@ impl TbError {
         err_tb_ub(match self.error_kind {
             CannotAccess(perm) => {
                 format!(
-                    "Impossible to perform a {is_foreign} {kind} access through {initial:?} because {current:?} has permissions {perm:?}.",
+                    "Impossible to perform a {is_foreign} {kind} through {initial:?} because {current:?} has permissions {perm:?}.",
```
RalfJung: I think our messages usually start lower-case (check the code elsewhere in Miri)

tests/fail/tree-borrows/alternate-read-write.stderr
```diff
@@ -0,0 +1,14 @@
+error: Undefined Behavior: write access through <TAG> is forbidden because it is a child of <TAG> which is Permission(Frozen).
```
RalfJung, line 1: `Permission(Frozen)` looks like it is still printing debug representations

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -214,7 +214,7 @@ trait EvalContextPrivExt<'mir: 'ecx, 'tcx: 'mir, 'ecx>: crate::MiriInterpCxExt<'
     fn tb_reborrow(
         &mut self,
         place: &MPlaceTy<'tcx, Provenance>, // parent tag contained here
-        size: Size,
+        ptr_size: Size,
```
RalfJung, line 173: I am confused by this name. The doc comment doesn't explain it either...

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -0,0 +1,555 @@
...
+                    &mut self,
+                    ancestor,
+                    Some(next_in_path),
+                    AccessRelatedness::StrictChildAccess,
+                )?;
+            }
+        };
+        // All (potentially zero) ancestors have been explored, call f_propagate on start
+        stack.exec_and_visit(&mut self, start_idx, None, AccessRelatedness::This)?;
+        // up to this point we have never popped from stack, hence if the
```
RalfJung, line 245:
```suggestion
        // up to this point we have never popped from `stack`, hence if the
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -0,0 +1,555 @@
...
+/// Internal contents of `Tree` with the minimum of mutable access for
+/// the purposes of the tree traversal functions: the permissions (`perms`) can be
+/// updated but not the tree structure (`tag_mapping` and `nodes`)
+struct TreeVisitor<'tree> {
+    tag_mapping: &'tree UniKeyMap<BorTag>,
+    nodes: &'tree UniValMap<Node>,
+    perms: &'tree mut UniValMap<LocationState>,
+}
+
+/// Whether to continue exploring the chilrdren recursively or not.
```
RalfJung, line 145:
```suggestion
/// Whether to continue exploring the children recursively or not.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -458,8 +456,23 @@ impl Tree {
...
     /// and removes from the tree all tags that have no live children.
+    ///
+    /// NOTE: This leaves in the middle of the tree tags that are unreachable but have
+    /// reachable children. There is a potential for compacting the tree by reassigning
+    /// children of dead tags to the nearest live parent, but it must be done with care
+    /// not to remove UB.
+    ///
+    /// Example: Consider the tree `root - parent - child`, with `parent: Frozen` and
+    /// `child: Reserved`. This tree can exist. If we blindly delete `parent` and reassign
+    /// `child` to be a direct child of `root` then Reads to `child` are now permitted
```
RalfJung, line 470: I assume you mean Writes, not Reads?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -0,0 +1,555 @@
...
+enum ContinueTraversal {
+    Recurse,
+    SkipChildren,
+}
+
+impl<'tree> TreeVisitor<'tree> {
+    // Applies `f_propagate` to every vertex of the tree top-down in the following order: first
+    // all ancestors of `start`, then `start` itself, then children of `start`, then the rest.
+    // This ensures that errors are triggered in the following order
+    // - first invalid accesses with insufficient permissions, oldest pointer first
```
RalfJung, line 155: What does "oldest pointer" mean here?

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -276,8 +276,10 @@ trait EvalContextPrivExt<'mir: 'ecx, 'tcx: 'mir, 'ecx>: crate::MiriInterpCxExt<'
         let mut tree_borrows = alloc_extra.borrow_tracker_tb().borrow_mut();
         if new_perm.perform_read_access {
             let global = &this.machine.borrow_tracker.as_ref().unwrap();
-            // FIXME: make sure this read counts towards race_condition
             tree_borrows.perform_access(AccessKind::Read, orig_tag, range, global)?;
+            if let Some(data_race) = alloc_extra.data_race.as_ref() {
+                data_race.read(alloc_id, range, &this.machine)?;
+            }
```
RalfJung, line 283: It seems concerning that no test was updated with this rather fundamental change in behavior... there should be a 'fail' test checking this for SB; can you copy that for TB?

README.md
```diff
@@ -551,14 +558,35 @@ extern "Rust" {
...
-    fn miri_print_borrow_stacks(alloc_id: u64);
+    fn miri_print_borrow_state(alloc_id: u64, show_unnamed: bool);
+
+    /// Miri-provided extern function to associate a name to the nth parent of a tag.
+    /// Typically the name given would be the name of the program variable that holds the pointer.
+    /// Unreachable tags can still be named by using nonzero `nth_parent` and a child tag.
+    ///
+    /// This function does nothing under Stacked Borrows, since Stacked Borrows's implementation
+    /// of `miri_print_borrow_state` does not show the names.
+    fn miri_pointer_name(ptr: *const (), nth_parent: u8, name: &[u8]);
```
RalfJung, line 589: This is now entirely duplicated in tests/util, right? Might be better to just change the readme to link to that file then, to avoid redundancy.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -257,9 +249,10 @@ trait EvalContextPrivExt<'mir: 'ecx, 'tcx: 'mir, 'ecx>: crate::MiriInterpCxExt<'
         let Some(new_perm) = new_perm else { return Ok(Some((alloc_id, orig_tag))); };
 
         if let Some(protect) = new_perm.protector {
-            // See comment in `Stack::item_invalidated` for why we store the tag twice.
-            // FIXME: because of pointers that use parent tag, we need to handle items
-            // protected multiple times
+            // FIXME: because of pointers that use parent tag, we may need to handle items
+            // protected multiple times. For now this is fine since pointers that use the parent
+            // tag do not get a protector, but if we change the policy in
+            // `NewPermission::from_ref_ty` then we will need to also update it here.
```
RalfJung, line 258: Currently the type of `new_perm` cannot even represent "old tag but please protect", so this comment is a bit confusing. I am not sure if it is actually necessary/helpful?

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -0,0 +1,544 @@
...
+        let this = self.eval_context_mut();
+        let new_perm = if let &ty::Ref(_, pointee, mutability) = val.layout.ty.kind() {
+            NewPermission::from_ref_ty(pointee, mutability, kind, this)
+        } else {
+            None
+        };
+        this.tb_retag_reference(val, new_perm)
+    }
+
+    /// Retagging procedure when nontrivial types are involved.
```
RalfJung, line 360:
```suggestion
    /// Retag all pointers that are stored in this place.
```

tests/fail/tree-borrows/reserved/cell-protected-write.stderr
```diff
@@ -7,11 +7,11 @@ Warning: this tree is indicative only. Some tags may have been hidden.
 | Re*|      │   └────<TAG=callee:x> Strongly protected
 | Re*|      └────<TAG=y,callee:y,caller:y>
 ──────────────────────────────────────────────────────────────────────
-error: Undefined Behavior: write access through <TAG> is forbidden because it is a foreign tag for <TAG>, which would hence change from Permission(Reserved { ty_is_freeze: false }) to Permission(Disabled), but <TAG> is protected
+error: Undefined Behavior: write access through named pointer 'y,callee:y,caller:y' is forbidden because it is a foreign tag for named pointer 'callee:x', which would hence change from Reserved to Disabled, but named pointer 'callee:x' is protected
   --> $DIR/cell-protected-write.rs:LL:CC
    |
 LL |             *y = 1;
-   |             ^^^^^^ write access through <TAG> is forbidden because it is a foreign tag for <TAG>, which would hence change from Permission(Reserved { ty_is_freeze: false }) to Permission(Disabled), but <TAG> is protected
+   |             ^^^^^^ write access through named pointer 'y,callee:y,caller:y' is forbidden because it is a foreign tag for named pointer 'callee:x', which would hence change from Reserved to Disabled, but named pointer 'callee:x' is protected
```
RalfJung, line 14: Oh, so now the names are also used for errors? Makes sense, but then the docs of the miri extern fn should explain that. Also it might be good to print both the tag and the name? The current wording sounds as if "tagged pointer" and "named pointer" are two fundamental types of pointers. So e.g. "write access through tag <TAG>" and then optionally `(named 'y,callee:y,caller:y')`.

tests/utils/macros.rs
```diff
@@ -0,0 +1,61 @@
...
+    ($alloc_id:expr, $show:expr) => {
+        crate::utils::miri_extern::miri_print_borrow_state($alloc_id, $show);
+    };
+}
+
+/// `name!(ptr => nth_parent, name)`: associate `name` to the `nth_parent` of `ptr`.
+///
+/// `ptr` should be any pointer or reference that can be converted with `_ as *const u8`.
+///
+/// `nth_parent` is an optional `u8` that defaults to 0i. The corresponding ancestor
```
RalfJung, line 36:
```suggestion
/// `nth_parent` is an optional `u8` that defaults to 0. The corresponding ancestor
```

src/shims/foreign_items.rs
```diff
@@ -455,8 +455,11 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 let ptr = this.read_pointer(ptr)?;
                 let nth_parent = this.read_scalar(nth_parent)?.to_u8()?;
                 let name = this.read_byte_slice(name)?;
-                let name = String::from_utf8_lossy(name);
-                this.give_pointer_debug_name(ptr, nth_parent, &name.to_string())?;
+                // No we can't remove the `to_string` here because we need to
+                // end the shared borrow from `read_byte_slice` before we can
+                // start the mutable borrow for `give_pointer_debug_name`.
+                let name = String::from_utf8_lossy(name).to_string();
```
RalfJung, line 461: Sounds like you should be calling `to_owned`, not `to_string`.

RalfJung:
Awesome, let's land this. :) Thanks for your patience with the review!

@bors r+

PR #2787, opened 2023-02-13
move reject with isolation logic in fcntl

DebugSteven:
This PR moves the block of logic inside `fcntl` to reject if isolation is enabled into the branch checking if the command is `F_FULLFSYNC` on apple. This allows `fcntl` to duplicate file descriptors while using isolation.

This means we can now run the tokio tests with isolation.

This PR allows moves the now passing `fcntl` logic from `libc-fs-with-isolation.rs` into two different tests: 
- `fcntl(1, libc::F_DUPFD, 0)` succeeds with isolation 
- `fcntl(1, libc::F_FULLFSYNC, 0)` fails with isolation on MacOS

### DIFF
--- a/tests/pass-dep/shims/fcntl_f-fullfsync_apple.rs
+++ b/tests/pass-dep/shims/fcntl_f-fullfsync_apple.rs
@@ -1,0 +1,12 @@
+//@only-target-apple: F_FULLFSYNC only on apple systems
+//@compile-flags: -Zmiri-isolation-error=warn-nobacktrace
+
+use std::io::Error;
+
+fn main() {
+    // test `fcntl(F_FULLFSYNC)`
+    unsafe {
+        assert_eq!(libc::fcntl(1, libc::F_FULLFSYNC, 0), -1);
+        assert_eq!(Error::last_os_error().raw_os_error(), Some(libc::EPERM));
+    }
+}
--- a/tests/pass-dep/shims/libc-fs-with-isolation.rs
+++ b/tests/pass-dep/shims/libc-fs-with-isolation.rs
@@ -7,10 +7,9 @@ use std::fs;
 use std::io::{Error, ErrorKind};
 
 fn main() {
-    // test `fcntl`
+    // test `fcntl(F_DUPFD): should work even with isolation.`
     unsafe {
-        assert_eq!(libc::fcntl(1, libc::F_DUPFD, 0), -1);
-        assert_eq!(Error::last_os_error().raw_os_error(), Some(libc::EPERM));
+        assert!(libc::fcntl(1, libc::F_DUPFD, 0) >= 0);
     }
 
     // test `readlink`
--- a/tests/pass-dep/tokio/sleep.rs
+++ b/tests/pass-dep/tokio/sleep.rs
@@ -1,4 +1,4 @@
-//@compile-flags: -Zmiri-disable-isolation -Zmiri-permissive-provenance -Zmiri-backtrace=full
+//@compile-flags: -Zmiri-permissive-provenance -Zmiri-backtrace=full
 //@only-target-x86_64-unknown-linux: support for tokio only on linux and x86
 
 use tokio::time::{sleep, Duration, Instant};
@@ -7,8 +7,6 @@ use tokio::time::{sleep, Duration, Instant};
 async fn main() {
     let start = Instant::now();
     sleep(Duration::from_secs(1)).await;
-    // It takes 96 millisecond to sleep for 1 millisecond
-    // It takes 1025 millisecond to sleep for 1 second
     let time_elapsed = &start.elapsed().as_millis();
-    assert!(time_elapsed > &1000, "{}", time_elapsed);
+    assert!((1000..1100).contains(time_elapsed), "{}", time_elapsed);
 }
--- a/tests/pass-dep/tokio/tokio_mvp.rs
+++ b/tests/pass-dep/tokio/tokio_mvp.rs
@@ -1,5 +1,5 @@
 // Need to disable preemption to stay on the supported MVP codepath in mio.
-//@compile-flags: -Zmiri-disable-isolation -Zmiri-permissive-provenance
+//@compile-flags: -Zmiri-permissive-provenance
 //@only-target-x86_64-unknown-linux: support for tokio exists only on linux and x86
 
 #[tokio::main]
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -628,13 +628,6 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         let fd = this.read_scalar(&args[0])?.to_i32()?;
         let cmd = this.read_scalar(&args[1])?.to_i32()?;
 
-        // Reject if isolation is enabled.
-        if let IsolatedOp::Reject(reject_with) = this.machine.isolated_op {
-            this.reject_in_isolation("`fcntl`", reject_with)?;
-            this.set_last_error_from_io_error(ErrorKind::PermissionDenied)?;
-            return Ok(-1);
-        }
-
         // We only support getting the flags for a descriptor.
         if cmd == this.eval_libc_i32("F_GETFD") {
             // Currently this is the only flag that `F_GETFD` returns. It is OK to just return the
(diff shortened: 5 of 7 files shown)


tests/pass-dep/tokio/sleep.rs
```diff
@@ -1,14 +1,17 @@
...
 
 use tokio::time::{sleep, Duration, Instant};
 
 #[tokio::main]
 async fn main() {
+    // We are using a very short time for sleep, 1 millisecond, (and ensure sleep waits for at
+    // least 1000 microseconds, the same as 1 millisecond) because this test is using miri's
+    // virtual clock due to isolation. The virtual clock works by using 1 code block passed as 10
+    // nanoseconds. While writing this test, I saw 1,000,000 basic blocks had been executed pretty
+    // quickly which would be the equivalent of 1 millisecond.
```
oli-obk, line 12: @pvdrz I think we may want to look into increasing the tick size in isolation mode. A second of isolation mode sleep time should probably be not more than 10s real time on a regular desktop CPU.

tests/pass-dep/tokio/sleep.rs
```diff
@@ -1,14 +1,17 @@
...
     let start = Instant::now();
-    sleep(Duration::from_secs(1)).await;
-    // It takes 96 millisecond to sleep for 1 millisecond
-    // It takes 1025 millisecond to sleep for 1 second
-    let time_elapsed = &start.elapsed().as_millis();
+    sleep(Duration::from_millis(1)).await;
+    let time_elapsed = &start.elapsed().as_micros();
     assert!(time_elapsed > &1000, "{}", time_elapsed);
```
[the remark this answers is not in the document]
DebugSteven: Testing this locally I saw the time elapsed was 1,666 microseconds. I picked 2,000 as the upper end. Let me know if I should adjust that.

tests/pass-dep/tokio/sleep.rs
```diff
@@ -1,14 +1,17 @@
-//@compile-flags: -Zmiri-disable-isolation -Zmiri-permissive-provenance -Zmiri-backtrace=full
+//@compile-flags: -Zmiri-isolation-error=warn -Zmiri-permissive-provenance -Zmiri-backtrace=full
```
RalfJung, line 1: What is the point of the `warn` setting here?

tests/pass-dep/shims/fcntl.rs
```diff
@@ -0,0 +1,14 @@
...
+
+fn main() {
+    // test `fcntl`
+    // This should succeed for everything.
+    unsafe {
+        assert_eq!(libc::fcntl(1, libc::F_DUPFD, 0), 3);
+        // Since there's no error, this might not have the right behavior.
+        assert_eq!(Error::last_os_error().raw_os_error(), Some(0));
+    }
+}
```
RalfJung, line 14: Why a new test? Seems fine to keep this in libc-fs-with-isolation, but assert that it returns 0 and no error?

Also why do you check last_os_error when there was no error...?

tests/pass-dep/tokio/sleep.rs
DebugSteven: I left it in while I was still developing & forgot about it since it’s no longer needed. It’s removed now.

tests/pass-dep/shims/fcntl.rs
DebugSteven: I wasn’t sure the right route to go for this test case. I moved it to a different file since it didn’t seem like it fit with the other functions erroring in libc-fs-with-isolation. I moved it back.

How would I assert there’s no error without checking last_os_error? The function doesn’t return 0 because it is returning a duplicated file descriptor.

RalfJung:
> How would I assert there’s no error without checking last_os_error?

Check the fcntl docs, the return value will somehow indicate whether there was an error.
One never checks last_os_error unless one already knows that there was an error and wants to learn what the error was.

oli-obk:
> One never checks last_os_error unless one already knows that there was an error and wants to learn what the error was.

During review I said to leave it in, because it doesn't hurt. What's the worst case? We notice when the default value changes?

We could move it to a separate test tho.

tests/pass-dep/shims/libc-fs-with-isolation.rs
```diff
@@ -9,8 +9,7 @@ use std::io::{Error, ErrorKind};
 fn main() {
     // test `fcntl`
```
RalfJung, line 10:
```suggestion
    // test `fcntl(F_DUPFD)`: should work even with isolation.
```

tests/pass-dep/shims/fcntl_f-fullfsync_apple.rs
```diff
@@ -0,0 +1,12 @@
+//@only-target-apple: F_FULLFSYNC only on apple systems
+//@compile-flags: -Zmiri-isolation-error=warn-nobacktrace
+
+use std::io::Error;
+
+fn main() {
+    // test `fcntl`
```
RalfJung, line 7:
```suggestion
    // test `fcntl(F_FULLFSYNC)`
```

tests/pass-dep/tokio/tokio_mvp.rs
```diff
@@ -1,5 +1,5 @@
 // Need to disable preemption to stay on the supported MVP codepath in mio.
-//@compile-flags: -Zmiri-disable-isolation -Zmiri-permissive-provenance
+//@compile-flags: -Zmiri-isolation-error=warn -Zmiri-permissive-provenance
```
RalfJung, line 2: The question I raised about this flag for the other test also applies here. I usually avoid repeating the same question multiple times, you can assume it applies to all occurrences of the same pattern.

tests/pass-dep/tokio/sleep.rs
pvdrz: are you telling me that you want `miri` to be slower? :trollface:

RalfJung: So instead of 10ns we should make it 100ns? Sure, feel free to make a PR.

oli-obk: The thing is that if you sleep for 1s without isolation, the test runs in user-visible 1s (or close enough). If you sleep for 1s in isolation, it takes minutes. I think isolation should at least be in the right ballpark.

> So instead of 10ns we should make it 100ns? Sure, feel free to make a PR.

more like 1000ns, but yes

RalfJung: > The thing is that if you sleep for 1s without isolation, the test runs in user-visible 1s (or close enough). If you sleep for 1s in isolation, it takes minutes. I think isolation should at least be in the right ballpark.

Oh, but something here doesn't make sense. With isolation, it should always take the same amount of real time, no matter how long you sleep. Sleeping 60s should still only take a few ms of real time at most. And for sure changing the clock timer should not change how much real time a program takes. Something else is wrong here and this needs to be investigated.

tests/pass-dep/tokio/sleep.rs
```diff
@@ -1,14 +1,17 @@
...
+    // virtual clock due to isolation. The virtual clock works by using 1 code block passed as 10
+    // nanoseconds. While writing this test, I saw 1,000,000 basic blocks had been executed pretty
+    // quickly which would be the equivalent of 1 millisecond.
     let start = Instant::now();
-    sleep(Duration::from_secs(1)).await;
-    // It takes 96 millisecond to sleep for 1 millisecond
-    // It takes 1025 millisecond to sleep for 1 second
-    let time_elapsed = &start.elapsed().as_millis();
-    assert!(time_elapsed > &1000, "{}", time_elapsed);
+    sleep(Duration::from_millis(1)).await;
```
RalfJung, line 14: Which underlying sleeping primitive does tokio even use here?

oli-obk: It doesn't use any sleeping yet at all. Tokio itself does, but the current miri hook implementation just "wakes up early" (immediately), and then tokio tries to sleep again until the specified time is over. So we're busy waiting until the timer is done.

tests/pass-dep/tokio/sleep.rs
oli-obk: This was resolved in https://github.com/rust-lang/miri/pull/2816

tests/pass-dep/tokio/sleep.rs
RalfJung: Ah, so make this part of my complaint around incomplete shim implementations that dont document/check their limitations. ;)

PR #2794, opened 2023-02-21
Rustup

RalfJung:


RalfJung:
rustfmt is completely destroying the `dyn*` test -- it removes `dyn*` entirely. Shouldn't it ignore syntax it does not understand?

We'll need some way to exempt that file from formatting.

oli-obk:
A bit roundabout as it needs another feature to allow rustfmt::skip to be a top level attribute, but 🤷

RalfJung:
Ah, nice!
@bors r+

PR #2801, opened 2023-02-24
a bit of FileDescriptor trait cleanup

RalfJung:
- add default impl for `is_tty`
- `as_epoll_handle` was just dyn downcasting in disguise; this can be done more generally

### DIFF
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -1,5 +1,6 @@
 pub mod convert;
 
+use std::any::Any;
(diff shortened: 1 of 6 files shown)


oli-obk:
> * `as_epoll_handle` was just dyn downcasting in disguise; this can be done more generally

the same is true for `as_file_handle`

RalfJung:
Good point! I got rid of that function as well.

Issue #2800, opened 2023-02-24
epoll_wait implementation can silently produce wrong results


RalfJung:
This is to track the concerns I raised in https://github.com/rust-lang/miri/pull/2764#discussion_r1105055481: currently it is possible to run Rust code in Miri that calls `epoll_wait` and the call returns, but its behavior is just wrong. That should never happen. The call should either halt execution (e.g. via `throw_unsup`) or return with correct behavior. It's okay to only support a tiny fragment of what epoll_wait can do, but in that case it is IMO mandatory to fail loudly when the code being run falls outside that fragment.

This looks like it might also affect the `write` implementation for `Event`.

@DebugSteven can you look into adding the necessary checks in `epoll_wait` and `Event::write`?

DebugSteven:
Yeah I can do that. I've been working on checks in `Event::write` already. I'll put a PR with that progress in a few hours & work on adding checks for `epoll_wait` today.

RalfJung:
https://github.com/rust-lang/miri/pull/2838 fixes the immediate issue here, I think.

`Event::write` still seems wrong (it uses `from_be_bytes` instead of the current target endianess), but I don't know if there is any way that programs could observe this.

Antiarchitect:
Hi,
I don't know if this belongs here for sure but when I try to `cargo +nightly miri run` my project I face this backtrace:

```
Preparing a sysroot for Miri (target: x86_64-unknown-linux-gnu)... done
WARNING: Ignoring `RUSTC_WRAPPER` environment variable, Miri does not support wrapping.
    Finished dev [unoptimized + debuginfo] target(s) in 0.16s
     Running `/home/andrey/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/cargo-miri runner target/miri/x86_64-unknown-linux-gnu/debug/mqtt-proxy`
error: unsupported operation: returning ready events from epoll_wait is not yet implemented
   --> /home/andrey/.cargo/registry/src/index.crates.io-6f17d22bba15001f/mio-0.8.10/src/sys/unix/selector/epoll.rs:97:9
    |
97  | /         syscall!(epoll_wait(
98  | |             self.ep,
99  | |             events.as_mut_ptr(),
100 | |             events.capacity() as i32,
101 | |             timeout,
102 | |         ))
    | |__________^ returning ready events from epoll_wait is not yet implemented
    |
    = help: this is likely not a bug in the program; it indicates that the program performed an operation that the interpreter does not support
[...]

PR #2802, opened 2023-02-25
prevent epoll_wait from failing silently

DebugSteven:
This is a WIP PR to handle issues raised in https://github.com/rust-lang/miri/issues/2800.

So far, this handles checking `Event::write` to ensure we get 8 bytes or appropriately fail & that the `checked_add` does not exceed the max u64 value.

This still needs to handle `epoll_wait` correctly.

### DIFF
--- a/src/shims/unix/linux/fd/event.rs
+++ b/src/shims/unix/linux/fd/event.rs
@@ -42,10 +42,12 @@ impl FileDescriptor for Event {
     /// A write call adds the 8-byte integer value supplied in
     /// its buffer to the counter.  The maximum value that may be
     /// stored in the counter is the largest unsigned 64-bit value
-    /// minus 1 (i.e., 0xfffffffffffffffe).  If the addition would
+    /// minus 1 (i.e., 0xfffffffffffffffe).
+    ///
+    /// When write is supported in eventfd, if the addition would
     /// cause the counter's value to exceed the maximum, then the
-    /// write either blocks until a read is performed on the
-    /// file descriptor, or fails with the error EAGAIN if the
+    /// write should either block until a read is performed on the
+    /// file descriptor, or fail with the error EAGAIN if the
     /// file descriptor has been made nonblocking.
 
     /// A write fails with the error EINVAL if the size of the
--- a/src/shims/unix/linux/fd/event.rs
+++ b/src/shims/unix/linux/fd/event.rs
@@ -62,9 +62,13 @@ impl FileDescriptor for Event {
         let bytes = bytes
             .try_into()
             .map_err(|_| err_unsup_format!("we expected 8 bytes and got {}", bytes.len()))?;
-        // FIXME handle blocking when addition results in exceeding the max u64 value
-        // or fail with EAGAIN if the file descriptor is nonblocking.
-        let v2 = v1.checked_add(u64::from_be_bytes(bytes)).unwrap();
+        let v2 = v1.checked_add(u64::from_be_bytes(bytes)).ok_or_else(|| {
+            err_unsup_format!(
+                "handle blocking when addition results \
+                in exceeding the max u64 value or fail with \
+                EAGAIN if the file descriptor is nonblocking."
+            )
+        })?;
         self.val.set(v2);
         Ok(Ok(8))
     }
--- a/src/shims/unix/linux/fd/event.rs
+++ b/src/shims/unix/linux/fd/event.rs
@@ -1,6 +1,7 @@
 use crate::shims::unix::fs::FileDescriptor;
 
 use rustc_const_eval::interpret::InterpResult;
+use rustc_target::abi::Endian;
 
 use std::cell::Cell;
 use std::io;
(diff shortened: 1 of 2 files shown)


src/shims/unix/linux/fd/event.rs
```diff
@@ -59,11 +61,17 @@ impl FileDescriptor for Event {
...
-        // or fail with EAGAIN if the file descriptor is nonblocking.
-        let v2 = v1.checked_add(u64::from_be_bytes(bytes.try_into().unwrap())).unwrap();
+        let bytes = bytes
+            .try_into()
+            .map_err(|_| err_unsup_format!("we expected 8 bytes and got {}", bytes.len()))?;
+        let v2 = v1.checked_add(u64::from_be_bytes(bytes)).ok_or_else(|| {
+            err_unsup_format!(
+                "handle blocking when addition results \
+                in exceeding the max u64 value or fail with \
+                EAGAIN if the file descriptor is nonblocking."
```
DebugSteven, line 71: Should this say something more like "overflow not implemented", if it's left this way? That way the message isn't aspiration & reflects the current state. I could put the `FIXME` comment back so it's known what needs to be implemented still.

saethlin: The error message should be useful to a user seeing this come out of Miri for the first time, and I think it should be worded as if you're saying something to the user.

Currently this is worded more like a note to you/us. I'd go for something like 

> Miri currently has an incomplete epoll implementation. This operation would overflow (please fill in what exactly is being overflowed here, the details of this are a bit beyond me at the moment), which is not yet implemented.

src/shims/unix/linux/fd/event.rs
```diff
@@ -59,11 +61,17 @@ impl FileDescriptor for Event {
         bytes: &[u8],
```
RalfJung, line 61: There's a FIXME a few lines up, what's up with that?

I assume this should use the endianess of the target. Currently you are using `from_be_bytes`. That is strange since x86 is little-endian?

src/shims/unix/linux/fd/event.rs
```diff
@@ -59,11 +61,17 @@ impl FileDescriptor for Event {
...
     ) -> InterpResult<'tcx, io::Result<usize>> {
         let v1 = self.val.get();
-        // FIXME handle blocking when addition results in exceeding the max u64 value
-        // or fail with EAGAIN if the file descriptor is nonblocking.
-        let v2 = v1.checked_add(u64::from_be_bytes(bytes.try_into().unwrap())).unwrap();
+        let bytes = bytes
```
RalfJung, line 64:
```suggestion
        let bytes: [u8; 8] = bytes
```
This type inference is a bit too much magic for my taste.

src/shims/unix/linux/fd/event.rs
oli-obk: I confused myself on the order of the bytes. These bytes are written in native endianess (see an example usage) https://github.com/tokio-rs/mio/blob/master/src/sys/unix/waker.rs#L31

RalfJung: > These bytes are written in native endianess 

All right, that's what I expected. So this code should match on the target endianess (in `sess.target.endian`), and call `from_be_bytes`/`from_le_bytes` as appropriate. Might make sense to have a helper function `from_target_endianess_bytes`.

The annoying part is that we already have exactly this logic in `Allocation::read_scalar`. But here we get the already-read buffer of `u8`... if we got the original pointer to the user-provided buffer instead, we could just do `read_scalar`... well we'd also have to reset the alignment requirement to 1... not sure if that would be any better in the end.

RalfJung: Yeah let's stick with the current approach for now.

src/shims/unix/linux/fd/event.rs
```diff
@@ -17,6 +18,9 @@ pub struct Event {
     /// The object contains an unsigned 64-bit integer (uint64_t) counter that is maintained by the
     /// kernel. This counter is initialized with the value specified in the argument initval.
     pub val: Cell<u64>,
+    /// We don't have access to interpcx in the file descriptor method, so we use this for passing
```
RalfJung, line 21:
```suggestion
    /// FIXME HACK: We don't have access to interpcx in the file descriptor method, so we use this for passing
```

RalfJung:
This is still marked as a draft - @DebugSteven what is missing from your side here?

RalfJung:
@DebugSteven could you give us a status update here? I think it is quite important that we fix the currently dangerously wrong epoll_wait implementation. If we can't get this done in a timely manner, I would prefer to land a PR that simply disables epoll_wait entirely, which is still better than exposing users to wrong behavior.

DebugSteven:
Sorry for the delay on this.

I intended to mark this ready when `epoll_wait` returns the correct values. I was making pretty decent progress on that this last week by looking at the kernel implementation. I'm not at a point where I can promise I'll get it working, but I'm feeling hopeful about it.

I agree it wasn't good to leave `epoll_wait` in a state where it always returns 0 & misleads users. Here's a PR that [throws unsupported for `epoll_wait`](https://github.com/rust-lang/miri/pull/2838). Please let me know if you had something else in mind.

DebugSteven:
Implementation is ongoing. I’m looking into how the linux kernel does it as a reference for moving forward.

RalfJung:
@DebugSteven I'm going to close this PR as being inactive. Feel free to reopen if you want to get back to this! Make sure to reopen *before* pushing to the branch, otherwise github won't let you reopen.

PR #2804, opened 2023-03-05
Rustup

saethlin:


saethlin:
Ah, I should probably do this with a josh sync. I'll do that in the morning.

RalfJung:
Thanks!
@bors r+

PR #2430, opened 2022-07-23
attempt to overwrite globally set build.rustc-wrapper

RalfJung:
Hopefully fixes https://github.com/rust-lang/miri/issues/2429

PR #2806, opened 2023-03-07
Install binaries to the miri toolchain's sysroot

saethlin:
The default install produces this behavior:
```
$ cargo +miri miri --version
miri 0.1.0 (0ba1f4a0 2023-03-05)
$ cargo +nightly miri --version
miri 0.1.0 (0ba1f4a0 2023-03-05)
```
Which is not good. We've effectively erased the toolchain-selecting behavior.

After this change, we now get this:
```
$ cargo +miri miri --version
miri 0.1.0 (0ba1f4a0 2023-03-05)
$ cargo +nightly miri --version
miri 0.1.0 (f63ccaf 2023-03-06)
```

Thanks @jyn514 who all but wrote this for me.


saethlin:
That is very strange. I can't reproduce the CI failure on my desktop, and I also can't reproduce it in a fresh `ubuntu:latest` container. Here's my shell history from that container:
```
    1  cd
    2  git
    3  apt-get update
    4  apt-get install -y git gcc make
    5  apt-get install -y build-essential
    6  git clone https://github.com/saethlin/miri
    7  cd miri/
    8  curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
    9  apt-get install -y curl
   10  curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
   11  ./miri toolchain
   12  apt-get install -y python3
   13  ./miri toolchain
   14  source ~/.cargo/env 
   15  cargo install rustup-toolchain-install-master
   16  apt-get install -y pkg-config
   17  cargo install rustup-toolchain-install-master
   18  apt-get install -y libssl-devel
   19  apt-get install -y libssl-deve
   20  apt-get install -y libssl-dev
   21  apt-get install -y pkg-config
   22  cargo install rustup-toolchain-install-master
   23  ./miri toolchain
   24  ./miri test
   25  python3 test-cargo-miri/run-test.py 
   26  ./miri install
   27  python3 test-cargo-miri/run-test.py 
   28  history
```

jyn514:
@saethlin is there a way to see what command it's trying to run that doesn't exist? I see `echo 'build.rustc-wrapper = "thisdoesnotexist"'` on the line above which looks pretty suspicious.

miri
```diff
@@ -281,8 +282,10 @@ find_sysroot() {
 case "$COMMAND" in
 install)
     # "--locked" to respect the Cargo.lock file if it exists.
-    $CARGO install $CARGO_EXTRA_FLAGS --path "$MIRIDIR" --force --locked "$@"
-    $CARGO install $CARGO_EXTRA_FLAGS --path "$MIRIDIR"/cargo-miri --force --locked "$@"
+    # Install binaries to the miri toolchain's sysroot so they do not interact with other toolchains
+    TOOLCHAIN_SYSROOT=$($RUSTC --print sysroot)
```
RalfJung, line 286: We already have a `SYSROOT` variable which I think you can just use here.

RalfJung:
Ah, nice idea.

Is it even using the right binaries on CI still?

ci.sh
```diff
@@ -62,8 +62,7 @@ function run_tests {
   if [ "$HOST_TARGET" = x86_64-unknown-linux-gnu ]; then
     # These act up on Windows (`which miri` produces a filename that does not exist?!?),
     # so let's do this only on Linux. Also makes sure things work without these set.
-    export RUSTC=$(which rustc)
-    export MIRI=$(which miri)
+    #export RUSTC=$(which rustc) # Produces a warning unless we also set MIRI
```
jyn514, line 65: need at least one command in this block
```suggestion
    #export RUSTC=$(which rustc) # Produces a warning unless we also set MIRI
    :
```

rbtcollins:
Looks sensible to me (wearing my rustup hat).

saethlin:
> Is it even using the right binaries on CI still?

Yes that was the problem. CI caches the contents of `~/.cargo/bin/`, so even though we are no longer installing there, we get a non-functional `miri` and `cargo-miri` dropped in there by the CI cache.

I don't know what to do about this. Perhaps we remove that directory from the cache, land this PR, then add them back? Just to prove that this PR works and cache is the problem, I deleted the MacOS cache via GitHub and re-ran.

saethlin:
What is this suppposed to test? https://github.com/rust-lang/miri/blob/0ba1f4a0a62f02cefaeaa15bf55b1d890fb60a68/ci.sh#L61-L66

The comment just says "confuse" and I'm not even sure what could be confusing about this- we set `RUSTC` and `MIRI` to the executables we're trying to run, right?

RalfJung:
Yeah but them being set caused bugs in the past... see https://github.com/rust-lang/miri/pull/2430.

miri
```diff
@@ -281,8 +281,10 @@ find_sysroot() {
 case "$COMMAND" in
 install)
     # "--locked" to respect the Cargo.lock file if it exists.
-    $CARGO install $CARGO_EXTRA_FLAGS --path "$MIRIDIR" --force --locked "$@"
-    $CARGO install $CARGO_EXTRA_FLAGS --path "$MIRIDIR"/cargo-miri --force --locked "$@"
+    # Install binaries to the miri toolchain's sysroot so they do not interact with other toolchains
+    echo Binaries should be installed to $SYSROOT
```
RalfJung, line 285: "should be"? This sounds like you are telling the user what they should do, but that's probably not the intent here? "will be installed" sounds more clear?

saethlin: Oops that's from debugging and I didn't delete it

saethlin:
> see #2430.

I see the motivation to test `build.rustc-wrapper` there, that makes sense. Maybe I'm just being dense, but I still don't understand `RUSTC` and `MIRI`?

RalfJung:
Setting RUSTC_WRAPPER and RUSTC leads to very interesting cargo behavior, and then while I was at it I just set MIRI as well for good measure.

saethlin:
> So if setting RUSTC now causes CI failure, that's not a good sign. What happens when you bring back those env vars?

If I bring back only RUSTC, we fail CI because we get a warning. If I bring back both RUSTC and MIRI, while fixing the path for MIRI to be the interpreter we are trying to run, everything passes.

miri
```diff
@@ -281,8 +281,9 @@ find_sysroot() {
 case "$COMMAND" in
 install)
     # "--locked" to respect the Cargo.lock file if it exists.
-    $CARGO install $CARGO_EXTRA_FLAGS --path "$MIRIDIR" --force --locked "$@"
-    $CARGO install $CARGO_EXTRA_FLAGS --path "$MIRIDIR"/cargo-miri --force --locked "$@"
+    # Install binaries to the miri toolchain's sysroot so they do not interact with other toolchains
```
RalfJung, line 284:
```suggestion
    # Install binaries to the miri toolchain's sysroot so they do not interact with other toolchains.
```

RalfJung:
Nice, so looks like this approach works!

I think this means that the documentation at the top of the file, and in the readme, should be updated? The part about the toolchain staying the same is no longer needed, the docs should say where the binaries go, and we can remove the docs about how to reset rustup after `./miri install`.

CONTRIBUTING.md
```diff
@@ -129,18 +129,16 @@ development version of Miri using
...
 
-and then you can use it as if it was installed by `rustup`.  Make sure you use
-the same toolchain when calling `cargo miri` that you used when installing Miri!
-Usually this means you have to write `cargo +miri miri ...` to select the `miri`
-toolchain that was installed by `./miri toolchain`.
+and then you can use it as if it was installed by `rustup` as a component of the
+`miri` toolchain. Note that the `miri` and `cargo-miri` executables are placed
+in the `miri` toolchain's sysroot to prevent conflicts with other toolchains which
+would arise if they were placed in the usual `cargo` bin directory
+(usually ~/.cargo/bin`).
```
RalfJung, line 136:
```suggestion
in the `miri` toolchain's sysroot to prevent conflicts with other toolchains.
The Miri binaries in the `cargo` bin directory (usually `~/.cargo/bin`) are managed by rustup.
```

RalfJung:
Thanks! This elegantly solves an ancient issue when working on Miri. :)
@bors r+

saethlin:
Gah. I guess I was bamboozled by the CI cache after all because I never took your advice to add a step to delete the old binaries. I'll debug this tomorrow.

RalfJung:
Now your deletion code deletes the rustup-provided `cargo-miri`. We probably need a `rustup update` as well.

Or maybe the better idea is the `key` (and `restore-keys`) such that we just don't pick up the old cache any more.

saethlin:
When I've tried it, `rustup update` will only rewrite the rustup shims if it actually does an update.

Debugging this is also so frustrating because I can't coax any of the failures out locally. Even `HOST_TARGET=x86_64-unknown-linux-gnu ./ci.sh` works fine locally.

saethlin:
Yep, changing the cache key seems to get CI working again.

.github/workflows/ci.yml
```diff
@@ -54,8 +54,8 @@ jobs:
             # contains package information of crates installed via `cargo install`.
             ~/.cargo/.crates.toml
             ~/.cargo/.crates2.json
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: ${{ runner.os }}-cargo
+          key: ${{ runner.os }}-cargo-installfix-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: ${{ runner.os }}-cargo-installfix
```
RalfJung, line 58:
```suggestion
          key: ${{ runner.os }}-cargo-reset20230315-${{ hashFiles('**/Cargo.lock') }}
          restore-keys: ${{ runner.os }}-cargo-reset20230315
```

.github/workflows/ci.yml
```diff
@@ -106,8 +106,8 @@ jobs:
...
```
RalfJung, line 110:
```suggestion
          key: ${{ runner.os }}-cargo-reset20230315-${{ hashFiles('**/Cargo.lock') }}
          restore-keys: ${{ runner.os }}-cargo-reset20230315
```

Issue #2797, opened 2023-02-22
running miri in release mode should disable debug assertions


lukas-code:
### Code

```rust
fn main() {
    debug_assert!(false);
}
```

This currently passes with `cargo run --release`, but panics with `cargo miri run --release`.

Since debug assertions can actually change the behavior of a program, I think it would be useful to allow running miri with debug assertions disabled.

RalfJung:
This is currently a deliberate choice -- unfortunately undocumented. We should definitely at least fix that part. The reasoning is that Miri is anyway so slow that the benefit from the extra sanity checks outweighs the costs. OTOH I do agree that it can be confusing. I think we definitely want to keep overflow checks enabled (their cost in Miri is definitely negligible and this can find real bugs), I am less sure about regular debug assertions (which can actually be quite expensive).

You can force debug assertions to be disabled with `-Cdebug-assertions=off`

saethlin:
So are you advocating for Miri not setting any values here? What do you think about this consideration: https://github.com/rust-lang/miri/issues/2497

LegionMammal978:
I'm fine with Miri setting whatever it wants for `std`, `core`, etc.; it's always taken special incantations to fiddle with their `RUSTFLAGS`, and their behavior is unlikely to change much between them (unless the standard library is what is being debugged). What I'm more concerned about is the `RUSTFLAGS` of the user crate being compiled, since changes to the profile can sometimes have far more important effects to their behavior and control flow.

RalfJung:
> I just ran into this today when writing a reproducer for a reachable unreachable_unchecked() in cipher (https://github.com/RustCrypto/traits/issues/1275). To be defensive, the crate precedes the unreachable_unchecked() with a debug_assert!(false). I've also seen a few other crates, e.g., parking_lot, that switch between unreachable!() and unreachable_unchecked() based on cfg!(debug_assertions).

There's no problem here though, is there? Reaching the `unreachable` will still be reported.

> Soundness can often be sensitive to the particular profile settings. I've seen some crates which are unsound only when overflow_checks are disabled, 

With overflow checks enabled, those crates then raise panics, which are still clearly pointing out there is a bug.

>  It can easily make these edge cases very confusing to debug when Miri's behavior differs from the compiler's default behavior, given the same profile settings in Cargo.toml.

[...]

PR #2814, opened 2023-03-12
Emit a warning when Miri is used with optimizations

saethlin:
This may address https://github.com/rust-lang/miri/issues/2797, by clarifying to the user what is going on and what the consequences of their choices are.

Fixes https://github.com/rust-lang/miri/issues/2797

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -43,7 +43,9 @@ function run_tests {
     # optimizations up all the way, too).
     # Optimizations change diagnostics (mostly backtraces), so we don't check
     # them. Also error locations change so we don't run the failing tests.
-    MIRIFLAGS="${MIRIFLAGS:-} -O -Zmir-opt-level=4" MIRI_SKIP_UI_CHECKS=1 ./miri test -- tests/{pass,panic}
+    # We explicitly enable debug-assertions here, they are disabled by -O but we have tests
+    # which exist to check that we panic on debug assertion failures.
+    MIRIFLAGS="${MIRIFLAGS:-} -O -Zmir-opt-level=4 -Cdebug-assertions=yes" MIRI_SKIP_UI_CHECKS=1 ./miri test -- tests/{pass,panic}
 
     # Also run some many-seeds tests. 64 seeds means this takes around a minute per test.
     for FILE in tests/many-seeds/*.rs; do
--- a/src/bin/miri.rs
+++ b/src/bin/miri.rs
@@ -30,6 +30,8 @@ use rustc_middle::{
     },
     ty::{query::ExternProviders, TyCtxt},
 };
+use rustc_session::config::OptLevel;
+
 use rustc_session::{config::CrateType, search_paths::PathKind, CtfeBacktrace};
 
 use miri::{BacktraceStyle, BorrowTrackerMethod, ProvenanceMode, RetagFields};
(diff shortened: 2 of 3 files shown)


oli-obk:
Should we also tell the user that this will not significantly  improve the performance of running miri except in microbenchmarks?

ci.sh
```diff
@@ -43,7 +43,7 @@ function run_tests {
     # optimizations up all the way, too).
     # Optimizations change diagnostics (mostly backtraces), so we don't check
     # them. Also error locations change so we don't run the failing tests.
-    MIRIFLAGS="${MIRIFLAGS:-} -O -Zmir-opt-level=4" MIRI_SKIP_UI_CHECKS=1 ./miri test -- tests/{pass,panic}
+    MIRIFLAGS="${MIRIFLAGS:-} -O -Zmir-opt-level=4 -Cdebug-assertions=yes" MIRI_SKIP_UI_CHECKS=1 ./miri test -- tests/{pass,panic}
```
RalfJung, line 46: Why add the flag back here? I guess some tests rely on it? Please add a comment.

src/bin/miri.rs
```diff
@@ -82,6 +84,15 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
...
             }
 
+            if tcx.sess.opts.optimize != OptLevel::No {
+                tcx.sess.warn("Miri does not support optimizations.");
+            } else if tcx.sess.mir_opt_level() > 0 {
+                tcx.sess.warn("You have explicitly enabled MIR optimizations, overriding Miri's default \
+                    which is to completely disable them. Any optimizations may hide UB that Miri would \
+                    otherwise detect, and it is not necessarily possible to predict what kind of UB will \
+                    be missed.");
+            }
```
RalfJung, line 94: So if the user says `-O -Zmir-opt-level=3`, they will get the first warning, not the second? That seems wrong?

ci.sh
saethlin: Yeah. I considered doing something more clever, such `cfg(not(debug_assertions))` on the test cases, but the tests pretty explicitly call `debug_assert!` so it seemed very circular.

src/bin/miri.rs
saethlin: That was definitely wrong.

src/bin/miri.rs
```diff
@@ -82,6 +84,21 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
...
+                    such as whether debug assertions and overflow checks are enabled, those settings are \
+                    still applied.");
+            }
+            if tcx.sess.mir_opt_level() > 0 {
+                tcx.sess.warn("You have explicitly enabled MIR optimizations, overriding Miri's default \
+                    which is to completely disable them. Any optimizations may hide UB that Miri would \
+                    otherwise detect, and it is not necessarily possible to predict what kind of UB will \
+                    be missed. If you are enabling optimizations to make Miri run faster, we advise using \
+                    cfg(miri) to shrink your workload instead. The impact of enabling MIR optimizations is \
+                    usually marginal at best.");
```
RalfJung, line 99:
```suggestion
                    cfg(miri) to shrink your workload instead. The performance benefit of enabling MIR \
                    optimizations is usually marginal at best.");
```

PR #2816, opened 2023-03-15
Update the virtual clock in isolation mode to step forward with around the same speed as the host system.

oli-obk:
Before this, the 1s sleep test took around 4 minutes on my machine.

### DIFF
--- a/src/clock.rs
+++ b/src/clock.rs
@@ -3,7 +3,10 @@ use std::time::{Duration, Instant as StdInstant};
 
 /// When using a virtual clock, this defines how many nanoseconds we pretend are passing for each
 /// basic block.
-const NANOSECONDS_PER_BASIC_BLOCK: u64 = 10;
+/// This number is pretty random, but it has been shown to approximately cause
+/// some sample programs to run within an order of magnitude of real time on desktop CPUs.
+/// (See `tests/pass/shims/time-with-isolation*.rs`.)
+const NANOSECONDS_PER_BASIC_BLOCK: u64 = 5000;
 
 #[derive(Debug)]
 pub struct Instant {
--- a/tests/pass/concurrency/thread_park_isolated.rs
+++ b/tests/pass/concurrency/thread_park_isolated.rs
@@ -7,6 +7,6 @@ fn main() {
 
     thread::park_timeout(Duration::from_millis(200));
 
-    // Thanks to deterministic execution, this will wiat *exactly* 200ms (rounded to 1ms).
-    assert!((200..201).contains(&start.elapsed().as_millis()));
+    // Thanks to deterministic execution, this will wait *exactly* 200ms, plus the time for the surrounding code.
+    assert!((200..210).contains(&start.elapsed().as_millis()), "{}", start.elapsed().as_millis());
 }
--- a/tests/pass/shims/time-with-isolation.rs
+++ b/tests/pass/shims/time-with-isolation.rs
@@ -22,14 +22,23 @@ fn test_time_passes() {
     let diff = now2.duration_since(now1);
     assert_eq!(now1 + diff, now2);
     assert_eq!(now2 - diff, now1);
-    // The virtual clock is deterministic and I got 29us on a 64-bit Linux machine. However, this
+    // The virtual clock is deterministic and I got 15ms on a 64-bit Linux machine. However, this
     // changes according to the platform so we use an interval to be safe. This should be updated
     // if `NANOSECONDS_PER_BASIC_BLOCK` changes.
-    assert!(diff.as_micros() > 10);
-    assert!(diff.as_micros() < 40);
+    assert!(diff.as_millis() > 5);
+    assert!(diff.as_millis() < 20);
+}
+
+fn test_block_for_one_second() {
+    let end = Instant::now() + Duration::from_secs(1);
+    // This takes a long time, as we only increment when statements are executed.
+    // When we sleep on all threads, we fast forward to the sleep duration, which we can't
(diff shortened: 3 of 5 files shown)


tests/pass/shims/time-with-isolation.rs
```diff
@@ -25,11 +25,20 @@ fn test_time_passes() {
...
+    assert!(diff.as_millis() > 5);
+    assert!(diff.as_millis() < 20);
+}
+
+fn test_block_for_one_second() {
+    let end = Instant::now() + Duration::from_secs(1);
+    // This takes a long time, as we only increment when statements are executed.
+    // When we sleep on all threads, we fast forward to the sleep duration, which we can't
+    // do with busy waiting.
+    while Instant::now() < end {}
```
RalfJung, line 37: Oh this is an interesting test. It's not quite what I had in mind when talking about syncing up the two clocks.

What happens for a program like this?
```rust
let begin = Instant::now();
for _ in 0..100_000 {}
let time = begin.elapsed();
println!("The loop took {time:?}");
```
What does this print in regular execution (with debug mode and `-Zmir-opt-level=0`), vs Miri with isolation?

tests/pass/shims/time-with-isolation.rs
```diff
@@ -25,11 +25,20 @@ fn test_time_passes() {
     // The virtual clock is deterministic and I got 29us on a 64-bit Linux machine. However, this
```
RalfJung, line 25: This comment is surely outdated now.

tests/pass/concurrency/thread_park_isolated.rs
```diff
@@ -8,5 +8,5 @@ fn main() {
     thread::park_timeout(Duration::from_millis(200));
 
     // Thanks to deterministic execution, this will wiat *exactly* 200ms (rounded to 1ms).
```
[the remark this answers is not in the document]
RalfJung: 2ms overhead for the surrounding code seems like a lot though, maybe 5 microseconds per basic block is a bit too much?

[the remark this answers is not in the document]
RalfJung: Not sure what you mean, but I think this boils down to a call to futex_wait so the 2 extra milliseconds are the time it takes to get to that call, and the time taken after that call returns. That's the 'overhead' I am referring to.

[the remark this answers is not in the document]
oli-obk: Oh right. Yea so it's just counting that overhead as more time than strictly necessary.

src/clock.rs
```diff
@@ -3,7 +3,7 @@ use std::time::{Duration, Instant as StdInstant};
 
 /// When using a virtual clock, this defines how many nanoseconds we pretend are passing for each
 /// basic block.
-const NANOSECONDS_PER_BASIC_BLOCK: u64 = 10;
+const NANOSECONDS_PER_BASIC_BLOCK: u64 = 5000;
```
saethlin, line 6: All that this PR does is change a magic number. There's no particular reason to expect that basic blocks are uniform in size, or that the interpreter or typical host CPUs won't get faster or slower. Are we just expecting to tweak this number from time to time?

tests/pass/shims/time-with-isolation.rs
oli-obk: 18s under isolation, between 1 and 2 seconds on regular execution, opt level has no measurable effect.

RalfJung: > 36ms under isolation (on master)

Okay yeah the current number is way too small then.^^

Looks like this PR is somewhat overshooting though. Should we go for 1000ns maybe?

oli-obk: Whatever we choose, it will be odd as it depends on basic block counts which have no mapping to actual cost. Overshooting will make tests like the one I wrote finish faster, while producing somewhat odd values for tests like the one you wrote, without actually affecting the host wall time of miri interpreting your test.

RalfJung: How long does your test take with 1000ns?

oli-obk: 5s

RalfJung: Factor 5 on both ends, nice.

:shrug: no strong opinion, if you want to stick with 5 microseconds then I won't fight that.
The PR has some other unresolved threads though. :)

src/clock.rs
oli-obk: I don't think the number itself matters that much. The order of magnitude does. I'll leave a comment about the magicness of it.

If we have reason to tune it again in the future, we should find a better metric.

src/clock.rs
```diff
@@ -3,7 +3,9 @@ use std::time::{Duration, Instant as StdInstant};
 
 /// When using a virtual clock, this defines how many nanoseconds we pretend are passing for each
 /// basic block.
-const NANOSECONDS_PER_BASIC_BLOCK: u64 = 10;
+/// This number is pretty random, but it has been shown to approximately cause
+/// some sample programs to run within an order of magnitude of real time on desktop CPUs.
```
RalfJung, line 7:
```suggestion
/// some sample programs to run within an order of magnitude of real time on desktop CPUs.
/// (See `tests/pass/shims/time-with-isolation*.rs`.)
```
These are the sample programs you are referring to, right?

RalfJung:
r=me with "some sample programs" made a bit less vague.

PR #2826, opened 2023-03-29
Add link for tree borrows similar to stacked borrows

yerke:
@Vanille-N does it look ok to you?


Vanille-N:
I was just thinking that these were missing, thank you for adding them.
Looks great !

PR #2827, opened 2023-03-29
add `memcpy` and `strcpy` shims

kiscad:


src/shims/foreign_items.rs
```diff
@@ -744,6 +744,17 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
             }
+            "memcpy" => {
+                let [ptr_dest, ptr_src, n] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let ptr_dest = this.read_pointer(ptr_dest)?;
+                let ptr_src = this.read_pointer(ptr_src)?;
+                let n = this.read_target_usize(n)?;
+                let src_bytes =
+                    this.read_bytes_ptr_strip_provenance(ptr_src, Size::from_bytes(n))?.to_owned();
+                this.write_bytes_ptr(ptr_dest, src_bytes)?;
```
oli-obk, line 755: We have a `mem_copy` method that does both of these at the same time https://doc.rust-lang.org/nightly/nightly-rustc/rustc_const_eval/interpret/eval_context/struct.InterpCx.html#method.mem_copy

I believe we should not strip provenance, but just copy everything over including provenance.

oli-obk (changes requested):
Please also add tests that end up copying pointers

src/shims/foreign_items.rs
```diff
@@ -744,6 +744,54 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                    true,
+                )?;
+                this.write_pointer(ptr_dest, dest)?;
+            }
+            "strncpy" => {
+                let [ptr_dest, ptr_src, n] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let ptr_dest = this.read_pointer(ptr_dest)?;
+                let ptr_src = this.read_pointer(ptr_src)?;
+                let n = this.read_target_usize(n)?;
```
oli-obk, line 784: This is wrong. `strncpy` copies *up to* `n` bytes, but it will also stop earlier if the cstr is shorter. I don't remember if that includes the trailing null or not. Please check the `strncpy` docs and add tests for various edge cases.

oli-obk (commented):
oops forgot to send this review :/

src/shims/foreign_items.rs
```diff
@@ -744,6 +744,60 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                let ptr_src = this.read_pointer(ptr_src)?;
+                let n = this.read_c_str(ptr_src)?.len().saturating_add(1);
+                this.mem_copy(
+                    ptr_src,
+                    Align::ONE,
+                    ptr_dest,
+                    Align::ONE,
+                    Size::from_bytes(n),
+                    true,
+                )?;
```
RalfJung, line 776: This will preserve provenance in the copy. Is that really what we want?

Might make more sense to to `read_os_str_from_c_str` and then `write_os_str_to_c_str`.

tests/pass-dep/shims/libc-misc.rs
```diff
@@ -302,6 +302,77 @@ fn test_posix_mkstemp() {
...
+
+    unsafe {
+        let src = 123_i32;
+        let mut dest = 0_i32;
+        libc::memcpy(
+            &mut dest as *mut i32 as *mut libc::c_void,
+            &src as *const i32 as *const libc::c_void,
+            std::mem::size_of::<i32>(),
+        );
+        assert_eq!(dest, src);
```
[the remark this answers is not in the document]
RalfJung: This doesn't test that provenance actually got copied though. That needs `assert_eq!(*dest, 123);`.

src/shims/foreign_items.rs
saethlin: It's not true that they only deal with strings. Even if you are thinking based on strict aliasing, these functions deal in `char *` which means they can copy any object.

Whether using these functions like that is a _good idea_ is a separate question.

RalfJung: Hm yeah seems okay to make it public.

I guess we can also allow copying pointers with strcpy, if you think that's less confusing :shrug:

src/shims/foreign_items.rs
```diff
@@ -744,6 +744,62 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                this.write_pointer(ptr_dest, dest)?;
+            }
+            "strncpy" => {
+                let [ptr_dest, ptr_src, n] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let ptr_dest = this.read_pointer(ptr_dest)?;
+                let ptr_src = this.read_pointer(ptr_src)?;
+                let src_size = this.read_c_str(ptr_src)?.len().checked_add(1).unwrap();
```
RalfJung, line 784: This is not right: this will cause UB if there is no null terminator, but the search should stop after n bytes.

I don't think you can use `read_c_str` for `strncpy`.

tests/pass-dep/shims/libc-misc.rs
```diff
@@ -302,6 +302,125 @@ fn test_posix_mkstemp() {
...
+        let src = CString::new("rust").unwrap();
+        let size = src.as_bytes_with_nul().len();
+        let dest = libc::malloc(size + 1);
+        libc::strcpy(dest as *mut libc::c_char, src.as_ptr());
+        assert_eq!(CStr::from_ptr(dest as *const libc::c_char), src.as_ref());
+        libc::free(dest);
+    }
+}
+
+fn test_strncpy() {
```
RalfJung, line 382: Please also add a test to copy an array that does not have a null byte.

src/shims/foreign_items.rs
```diff
@@ -744,6 +744,62 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                )?;
+                this.write_pointer(ptr_dest, dest)?;
+            }
+            "strcpy" => {
+                let [ptr_dest, ptr_src] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let ptr_dest = this.read_pointer(ptr_dest)?;
+                let ptr_src = this.read_pointer(ptr_src)?;
+                let n = this.read_c_str(ptr_src)?.len().checked_add(1).unwrap();
+                this.mem_copy(
```
RalfJung, line 769:
```suggestion
                // We use `read_c_str` to determine the amount of data to copy,
                // and then use `mem_copy` for the actual copy. This means
                // pointer provenance is preserved by this implementation of `strcpy`.
                // That is properly overly cautious, but there also is no fundamental
                // reason to have `strcpy` destroy pointer provenance.
                let n = this.read_c_str(ptr_src)?.len().checked_add(1).unwrap();
                this.mem_copy(
```

src/shims/foreign_items.rs
```diff
@@ -744,6 +744,72 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+            "strncpy" => {
+                let [ptr_dest, ptr_src, n] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let ptr_dest = this.read_pointer(ptr_dest)?;
+                let ptr_src = this.read_pointer(ptr_src)?;
+                let n = usize::try_from(this.read_target_usize(n)?).unwrap();
+                let src_size = match this.read_c_str(ptr_src) {
+                    Ok(cstr) => cstr.len().checked_add(1).unwrap(),
+                    _ => n,
+                };
```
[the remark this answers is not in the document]
oli-obk: > It's difficult for user to set a max_length. Even in that case, users would usually set a very large max_length, which is not helpful.

Ah, I don't mean Rust users, but the call sites inside miri the tool. This is nothing users would see.

> In read_c_str, when the loop reaches the AllocRange boundary or it accesses the uninitialized bytes, the loop will immediately stop and return error

Yes, but that means if the first byte is an uninit byte, you ignore that and assume you got n bytes and copy n bytes afaict.

tests/pass-dep/shims/libc-misc.rs
```diff
@@ -429,6 +429,19 @@ fn test_strncpy() {
...
         libc::free(dest);
     }
+
+    // case: bytes without `\0` ending, and followed by uninitialized bytes
+    unsafe {
+        let slc = [1i8, 2, 3];
+        let src = libc::malloc(5);
+        libc::memcpy(src, slc.as_ptr() as *const libc::c_void, 3);
+        let dest = libc::calloc(4, 1);
+        libc::strncpy(dest as *mut libc::c_char, src as *const libc::c_char, 3);
```
oli-obk, line 439: The issue I'm worried about is if the `n` here is larger . So like if you use 10 in strncpy, but the 4th byte is uninit

src/shims/foreign_items.rs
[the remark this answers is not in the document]
RalfJung: Should we split strncpy into a separate PR? It is clearly harder, since fundamentally it cannot use `read_c_str` in its current form.

oli-obk:
@bors r+

Thanks for bearing with us through the various changes we asked for!

src/shims/foreign_items.rs
```diff
@@ -744,6 +744,44 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+            "strcpy" => {
+                let [ptr_dest, ptr_src] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let ptr_dest = this.read_pointer(ptr_dest)?;
+                let ptr_src = this.read_pointer(ptr_src)?;
+
+                // We use `read_c_str` to determine the amount of data to copy,
+                // and then use `mem_copy` for the actual copy. This means
+                // pointer provenance is preserved by this implementation of `strcpy`.
+                // That is properly overly cautious, but there also is no fundamental
```
RalfJung, line 772:
```suggestion
                // That is probably overly cautious, but there also is no fundamental
```
*oops* there was a typo in my suggestion...

PR #2828, opened 2023-03-29
Tree Borrows: improved diagnostics

Vanille-N:
Better diagnostics for Tree Borrows violations.

- Shows where the conflicting tags (the one that was accessed and the one that had a permission or protector that caused UB) were reborrowed, which is more readable than only `<TAG>`
- Shows a small history of what happened for the faulty tag to get there (which lines caused it to lose read/write permissions)
- Explains permissions and transitions in natural language (e.g. "does not have read permissions" instead of "is Disabled")

Not perfect, but at least testing TB will be less confusing.

Lack of range information from `RangeMap` makes selection of relevant events nontrivial: we reconstruct history from knowledge of `(initial, final)` and `(offset, pi, p'i)` so that `initial -> final = p1 -> p1' = p2 -> p2' = p3 -> ... = final `

### DIFF
--- a/src/borrow_tracker/tree_borrows/diagnostics.rs
+++ b/src/borrow_tracker/tree_borrows/diagnostics.rs
@@ -1,13 +1,89 @@
-use rustc_data_structures::fx::FxHashMap;
-
 use std::fmt;
 use std::ops::Range;
 
+use rustc_data_structures::fx::FxHashMap;
+use rustc_span::{Span, SpanData};
+use rustc_target::abi::Size;
+
 use crate::borrow_tracker::tree_borrows::{
-    err_tb_ub, perms::Permission, tree::LocationState, unimap::UniIndex,
+    perms::{PermTransition, Permission},
+    tree::LocationState,
+    unimap::UniIndex,
 };
 use crate::borrow_tracker::{AccessKind, ProtectorKind};
 use crate::*;
+
+/// Complete data for an event:
+/// - `kind` is what happened to the permissions
+/// - `access_kind` and `access_range` describe the access that caused the event
+/// - `offset` allows filtering only the relevant events for a given memory location
+/// (see how we perform the filtering in `History::extract_relevant`.
+/// - `span` is the line of code in question
+#[derive(Clone, Debug)]
+pub struct Event {
+    pub transition: PermTransition,
+    pub access_kind: AccessKind,
+    pub is_foreign: bool,
+    pub access_range: AllocRange,
+    pub offset: Size,
+    pub span: Span,
+}
+
+/// List of all events that affected a tag.
+/// NOTE: not all of these events are relevant for a particular location,
+/// the events should be filtered before the generation of diagnostics.
+/// Available filtering methods include `History::forget` and `History::extract_relevant`.
+#[derive(Clone, Debug)]
+pub struct History {
+    pub tag: BorTag,
+    pub created: (Span, Permission),
+    pub events: Vec<Event>,
+}
+
+/// History formatted for use by `src/diagnostics.rs`.
+///
+/// NOTE: needs to be `Send` because of a bound on `MachineStopType`, hence
+/// the use of `SpanData` rather than `Span`.
+#[derive(Debug, Clone, Default)]
+pub struct HistoryData {
+    pub events: Vec<(Option<SpanData>, String)>, // includes creation
+}
+
+impl History {
+    /// Record an additional event to the history.
+    pub fn push(&mut self, event: Event) {
+        self.events.push(event);
+    }
+}
+
+impl HistoryData {
+    // Format events from `new_history` into those recorded by `self`.
+    //
+    // NOTE: also converts `Span` to `SpanData`.
+    pub fn extend(
+        &mut self,
+        new_history: History,
+        tag_name: &'static str,
+        show_initial_state: bool,
+    ) {
+        let History { tag, created, events } = new_history;
+        let this = format!("the {tag_name} tag {tag:?}");
+        let msg_initial_state = format!(", in the initial state {}", created.1);
+        let msg_creation = format!(
+            "{this} was created here{maybe_msg_initial_state}.",
+            maybe_msg_initial_state = if show_initial_state { &msg_initial_state } else { "" },
+        );
+
+        self.events.push((Some(created.0.data()), msg_creation));
+        for &Event { transition, access_kind, is_foreign, access_range, span, offset: _ } in &events
+        {
+            // NOTE: `offset` is explicitly absent from the error message, it has no significance
+            // to the user. The meaningful one is `access_range`.
+            self.events.push((Some(span.data()), format!("{this} then transitioned {transition} due to a {rel} {access_kind} at offsets {access_range:?}.", rel = if is_foreign { "foreign" } else { "child" })));
+            self.events.push((None, format!("this corresponds to {}.", transition.summary())));
+        }
+    }
+}
 
 /// Some information that is irrelevant for the algorithm but very
 /// convenient to know about a tag for debugging and testing.
@@ -20,18 +96,29 @@ pub struct NodeDebugInfo {
     /// pointer in the source code.
     /// Helps match tag numbers to human-readable names.
     pub name: Option<String>,
+    /// Notable events in the history of this tag, used for
+    /// diagnostics.
+    ///
+    /// NOTE: by virtue of being part of `NodeDebugInfo`,
+    /// the history is automatically cleaned up by the GC.
+    /// NOTE: this is `!Send`, it needs to be converted before displaying
+    /// the actual diagnostics because `src/diagnostics.rs` requires `Send`.
+    pub history: History,
 }
+
 impl NodeDebugInfo {
-    /// New node info with a name.
-    pub fn new(tag: BorTag) -> Self {
-        Self { tag, name: None }
+    /// Information for a new node. By default it has no
+    /// name and an empty history.
+    pub fn new(tag: BorTag, initial: Permission, span: Span) -> Self {
+        let history = History { tag, created: (span, initial), events: Vec::new() };
+        Self { tag, name: None, history }
     }
 
     /// Add a name to the tag. If a same tag is associated to several pointers,
     /// it can have several names which will be separated by commas.
-    fn add_name(&mut self, name: &str) {
+    pub fn add_name(&mut self, name: &str) {
         if let Some(ref mut prev_name) = &mut self.name {
-            prev_name.push(',');
+            prev_name.push_str(", ");
             prev_name.push_str(name);
         } else {
             self.name = Some(String::from(name));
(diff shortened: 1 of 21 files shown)


src/diagnostics.rs
```diff
@@ -219,10 +221,27 @@ pub fn report_error<'tcx, 'mir>(
                 }
                 helps
             },
-            TreeBorrowsUb { .. } => {
-                let helps = vec![
+            TreeBorrowsUb { trigger_history, faulty_history, .. } => {
+                let url = "https://perso.crans.org/vanille/treebor";
```
saethlin, line 225: I agree that we should link people to something helpful, and also that your write-up is helpful, but I'm not excited about linking someone to a site that's not managed by the Rust project.

Vanille-N: Very understandable. I'll remove that.

tests/fail/tree-borrows/error-range.stderr
```diff
@@ -0,0 +1,39 @@
+error: Undefined Behavior: read access through <TAG> is forbidden because it is a child of <TAG> which is Disabled.
```
saethlin, line 1: "which is disabled" clashes with the rest of the terminology in this diagnostic now. Where was it disabled? Nothing in this diagnostic mentions a disable.

Not to open up the terminology question again, but I feel like in light of the other details given below it makes more sense to say "because it is a child of {tag} which lost all permissions". It seems like we have two sets of terminology here, and the one used in the helps requires a few more words but is more obvious to a reader.

Vanille-N: I'll update the diagnostic for protector violations too, it currently says "change from Active to Frozen"

src/diagnostics.rs
RalfJung: FWIW the readme still links there.

We could do what we did with SB, and have a page in https://github.com/rust-lang/unsafe-code-guidelines/tree/master/wip and link there. For now that page would basically just link to Neven's blog, but in the future it could contain its own concise description of the rules (like we do for SB).

RalfJung:
> This required getting some more information from RangeMap: from just the offset to the actual range.

Please undo that, the range is meaningless. Instead do what we did for SB: the offset is the first byte where a conflict happens; no statement is made about later bytes. Conceptually this code iterates over bytes and checks each byte separately; the fact that RangeMap optimizes this is an implementation detail that should not be observable.

Vanille-N:
Yeah I had specifically tried to avoid SB's way of doing this, which is to reconstruct the previous range from the current offset, which led me to think I needed the range from `RangeMap`.
It is in fact not necessary, but I'll let you decide if you find the added complexity of `extract_relevant` worth it.

src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -1,14 +1,53 @@
...
+    pub access_range: AllocRange,
+    pub offset: Size,
+    pub span: S,
+}
+
+/// List of all events that affected a tag.
+/// Many different offsets of memory are mixed together, so make sure
+/// to filter out just the relevant ones when printing.
+#[derive(Clone, Debug)]
+pub struct History<S> {
```
RalfJung, line 36: Please add a comment explaining what `S` is

tests/fail/tree-borrows/error-range.stderr
RalfJung: I agree we should use the same set of terminology everywhere, but I think it should be the state names. Reserved, Active, Frozen, Disabled seems much clearer to me than having to talk about "future write permissions".

tests/fail/tree-borrows/write-during-2phase.stderr
```diff
@@ -1,11 +1,28 @@
...
+LL |       let _res = f.add(unsafe {
+   |  ________________^
+LL | |         let n = f.0;
+LL | |         // This is the access at fault, but it's not immediately apparent because
+LL | |         // the reference that got invalidated is not under a Protector.
+LL | |         *inner = 42;
+LL | |         n
+LL | |     });
+   | |______^
+help: <TAG> lost read and write permissions due to a write access at offsets [0x0..0x8]
```
RalfJung, line 20: This should tell a story, e.g.
- TAG was created here in Reserved state by a reborrow (name parent tag, maybe?)
- TAG later changed to Disabled state due to a Foreign write access at offsets...

I guess @saethlin will want us to avoid using special Tree Borrows terminology here, but I really don't see a good way to avoid that. Replacing state names by long permission descriptions doesn't really explain that much IMO, and this here currently describes Reserved as "with read and write permissions" which is confusing since that sounds like Active.

Maybe we have the `help` line be a short one using special terms, and then have a more detailed `note`/`help` below (or somewhere) that explains a bit more, and that people familiar with TB can skip?

saethlin: > Replacing state names by long permission descriptions doesn't really explain that much IMO

🤷 I think it helps. It also gets us out of discussing disabled children.

Really we have a Stroustrup's Rule problem here: Everyone finds the new terminology confusing _at first_, but then later they want a terse notation for it. I've noticed that the pickup rate for changes in Miri can be amazingly slow, some rather experienced unsafe code users didn't notice the new span-based diagnostics until 6 months after they released.

So however you want to navigate all that is good with me. If we're going to put up all the Tree Borrows terminology, there needs to be an accessible way for people to learn what those terms mean. https://perso.crans.org/vanille/treebor/ is a great breakdown of the whole model, but it is quite intimidating for a newcomer who is just trying to figure out why Miri doesn't like their new data structure idea.

I've long wanted a sort of glossary built from common patterns that explains why they are invalid and how to fix them. https://miri.saethlin.dev/tree-borrows/ub exists now and will be complete once the upload finishes. I will work on such a glossary eventually if someone else doesn't beat me to it.

RalfJung: > Stroustrup's Rule

Definitely. But we also have the problem that there's a lot going on and trying to cram that all into a single sentence will not result in anything readable.

If you have good suggestions for how to avoid the state names, I am all ears. But the PR as-is describes Reserved as "having read and write permissions" and that's just misleading -- and the more correct "having read and future-write permissions" is no better than just saying "Reserved".

saethlin: > If you have good suggestions for how to avoid the state names, I am all ears.

I don't. I wish I did.

I was a bit surprised by the direction @Vanille-N took my comment, but as I dislike terse notation (because with my route through grad school I was constantly reading papers outside my subfield) I am not inclined to get in anyone's way when they are removing it. My initial comment was focused on the consistency or logical flow of terminology/phrasing because that's a clearer topic. If it turns out that addressing my comment that requires opening up a lot more lines of discussion let's do the incremental thing and add all the extra notes now and decide on the English that surrounds them later.

> We could also consider renaming Disabled to Inactive, Invalid, Deactivated, something like that.

Please not _another_ use of Invalid :p 

I don't particularly care what color the shed is, but I want us to be happy with the term before Tree Borrows becomes a default.

tests/fail/tree-borrows/error-range.stderr
```diff
@@ -0,0 +1,44 @@
...
+   |                  ^^^^^^^
+   = help: this corresponds to a loss of write permissions.
+help: the conflicting tag <TAG> then transitioned from Frozen to Disabled due to a foreign write access at offsets [0x5..0x6]
+  --> $DIR/error-range.rs:LL:CC
+   |
+LL |         data[5] = 1;
+   |         ^^^^^^^^^^^
+   = help: this corresponds to a loss of read permissions.
+   = note: BACKTRACE (of the first span):
+   = note: inside `main` at $DIR/error-range.rs:LL:CC
```
RalfJung, line 39: I love this error message!

RalfJung (commented):
Very nice, just some nits for the messages!

And one more concern: could you benchmark this against tree-borrows without the history tracking? In SB we had to work a bit to avoid regressing perf too much.

tests/fail/tree-borrows/read-to-local.stderr
```diff
@@ -1,12 +1,32 @@
-error: Undefined Behavior: write access through <TAG> is forbidden because it is a child of <TAG> which is Frozen.
+error: Undefined Behavior: write access through <TAG> is forbidden.
   --> $DIR/read-to-local.rs:LL:CC
    |
 LL |         *ptr = 0;
-   |         ^^^^^^^^ write access through <TAG> is forbidden because it is a child of <TAG> which is Frozen.
+   |         ^^^^^^^^ write access through <TAG> is forbidden.
    |
+   = help: the accessed tag <TAG> is also the conflicting tag.
+   = help: the conflicting tag <TAG> has state Frozen which forbids child write accesses.
```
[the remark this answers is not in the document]
Vanille-N: Yes I agree that this sounds weird. I'll try to find a solution that doesn't break too much the general structure of [creation of accessed - creation of conflicting - history of conflicting] with hard-to-follow branches.

Vanille-N:
> And one more concern: could you benchmark this against tree-borrows without the history tracking? In SB we had to work a bit to avoid regressing perf too much.

Tested just now. The easy comparison is between recording or ignoring the events: the only diagnostics-related operation that is done more than once per allocation is pushing an `Event` to the history.

With this measure, the two versions of the code with and without diagnostics have their mean ± σ that intersect on every single `./miri benchmark`, so the cost of recording every single event is no more than 2%.

RalfJung:
Are there any situations where we can avoid storing an event, or discard events that we won't need any more?

Vanille-N:
Discard, yes. Events are garbage collected with the tag they are relevant to.

Avoid storing, no. An event is _not_ an access, `Event = (access, range, transition) where !transition.is_noop()`, so there are by design at most 4 events (1 creation + ≤3 transitions) per tag per location.

All events can possibly be relevant to reconstruct the history of a tag, unless we're fine with having holes in the history, such as if we don't care where a tag went from `Active` to `Frozen` if the error is triggered because it is now `Disabled`.
I like having the current complete history of the permission, but it is true that we don't need all of it.

RalfJung:
Well I guess let's just go with this for now and maybe if @saethlin optimizes TB and this becomes an actual bottleneck we decide to only keep some of that history around. ;)

saethlin:
FWIW in Stacked Borrows I was rather shocked by how cheap the history is compared to the actual stacks. I think the reason this is the case is that these events are not split up per byte like aliasing properties are. In real programs I think a lot of the reborrows are _many_ bytes in size and will not be split no matter what aliasing activity is below them.

tests/fail/tree-borrows/reserved/cell-protected-write.stderr
```diff
@@ -15,7 +15,7 @@ LL |             *y = 1;
    |             ^^^^^^ write access through <TAG> (y, callee:y, caller:y) is forbidden.
    |
    = help: the accessed tag <TAG> (y, callee:y, caller:y) is a foreign tag for the protected tag <TAG> (callee:x).
```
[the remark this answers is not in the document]
RalfJung: Yeah I think it is a distraction, to understand the error it is not relevant that the pointer is foreign.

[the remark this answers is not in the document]
RalfJung: Hm okay that also reads awkwardly... I have tweaked the wording a bit more, time to land this. :)

PR #2833, opened 2023-04-06
Update to new ui_test crate

oli-obk:


tests/pass/0weak_memory_consistency.rs
```diff
@@ -61,17 +61,17 @@ fn test_corr() {
...
+        y.store(1, Release); //: ---------------------+                    |
+        r2 //:                                        |                    |
+    }); //:                                           |                    |
+    #[rustfmt::skip] //:                              |synchronizes-with   |happens-before
+    let j3 = spawn(move || { //:                      |                    |
+        acquires_value(&y, 1); //: <------------------+                    |
+        x.load(Relaxed) //: <----------------------------------------------+
+        //: The two reads on x are ordered by hb, so they cannot observe values
+        //: differently from the modification order. If the first read observed
+        //: 2, then the second read must observe 2 as well.
```
oli-obk, line 74: These and similar changes are for our new heuristic that detects things that are suspicious. It's obviously a bit overeager and detects `//                 | stuff`, so we'll need to reiterate a bit on https://github.com/oli-obk/ui_test/issues/46, but for now the fallout looks small enough to be ok to me

saethlin: I'd rather we be really picky about the magic comment syntax than have to tweak how we write comments elsewhere. I think that's what is on the table?

oli-obk: > I feel like the heuristic should never kick in with `// ` -- once there is a space there, that should be a clear signal, no?

The problem is that we have seen accidental spaces followed by commands in the rustc test suite, and then those commands were ignored. This is more an issue with `// ~` and `// [thing]` than `// @` though, as the @ is not needed in the rustc test suite and a leading space is in fact ok there

saethlin: I don't think those spaces are accidental. I have used them deliberately because I prefer the style. Such as `// compile-flags: -Zmir-opt-level=2`. That's what you are talking about right?

oli-obk: I get all that, I'm worried about `// @` or `// ~` being silently ignored. I can change the parser to accept spaces between `//` and the sigil, but I'd prefer to stay strict and just forbid `// @` entirely

oli-obk: So... one thing I didn't state clearly: this fallout is because of a heuristic *separate* from the parser. The strict parser only accepts `//@` and `//~` and ignores everything else. Then a second parser runs that checks for things that could be mistakes and complains about that without doing any processing.

RalfJung: Yeah and I think this diff shows that that heuristic is too loose. The entire point of ui-test is to make `// ` never relevant for the parser. I could be convinced that we should warn about `// @`, but I am not convinced about `//                                 @` and fully unconvinced about `//        |`.

oli-obk: I changed it to only consider a single space too close for comfort.

RalfJung:
Thanks!
@bors r+

PR #2835, opened 2023-04-10
rustup

saethlin:
@oli-obk I think this is the resolution we agreed to on Zulip?


src/lib.rs
```diff
@@ -125,10 +124,13 @@ pub use crate::tag_gc::{EvalContextExt as _, VisitTags};
...
 /// set per default, for maximal validation power.
+/// Also disable the MIR pass that inserts an alignment check on every pointer dereference. Miri
+/// does that too, and with a better error message.
 pub const MIRI_DEFAULT_ARGS: &[&str] = &[
     "--cfg=miri",
     "-Zalways-encode-mir",
     "-Zextra-const-ub-checks",
     "-Zmir-emit-retag",
     "-Zmir-opt-level=0",
+    "-Zmir-enable-passes=-CheckAlignment",
```
RalfJung, line 135: So this is not implied by mir-opt-level=0? Curious. Instead of listing all such passes we probably want a flag that says "please preserve all UB thank you" or so?

saethlin: Yeah, I will for sure add a flag that turns off the alignment checks and the niche checks when I get around to finishing implementing the niche checks.

mir-opt-level=0 turns off optimization passes because their `is_enabled` consults `tcx.sess.mir_opt_level`. CheckAlignment doesn't because well it isn't an optimization.

Would anyone else use a preserve-all-ub flag? Other checkers?

RalfJung: > CheckAlignment doesn't because well it isn't an optimization.

Yeah... I always considered mir-opt-level=0 to be the flag for "give me the original MIR without and extra fuzz", but indeed the name very clearly refers to optimizations. So we'd need a different flag name for "original unaltered MIR please".

PR #2838, opened 2023-04-11
throw unsupported for `epoll_wait`

DebugSteven:
This PR throws unsupported to indicate miri doesn't yet return ready events. Previously it always returned 0, indicating no ready events, even if events for the epoll file descriptor may have been ready.

### DIFF
--- a/tests/fail/tokio/sleep.stderr
+++ b/tests/fail/tokio/sleep.stderr
@@ -1,0 +1,68 @@
+error: unsupported operation: returning ready events from epoll_wait is not yet implemented
+  --> CARGO_REGISTRY/.../epoll.rs:LL:CC
+   |
+LL | /         syscall!(epoll_wait(
+LL | |             self.ep,
+LL | |             events.as_mut_ptr(),
+LL | |             events.capacity() as i32,
+LL | |             timeout,
+LL | |         ))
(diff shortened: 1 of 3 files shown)


tests/fail/tokio/sleep.stderr
```diff
@@ -0,0 +1,68 @@
...
+LL | |             self.ep,
+LL | |             events.as_mut_ptr(),
+LL | |             events.capacity() as i32,
+LL | |             timeout,
+LL | |         ))
+   | |__________^ returning ready events from epoll_wait is not yet implemented
+   |
+   = help: this is likely not a bug in the program; it indicates that the program performed an operation that the interpreter does not support
+   = note: BACKTRACE:
+   = note: inside `mio::sys::PLATFORM::selector::epoll::Selector::select` at CARGO_REGISTRY/.../mod.rs:LL:CC
```
oli-obk, line 14: Please filter out allthe notes. While we only run these tests on linux hosts,and thus don't suffer from the usual platform differences that other backtraces have, they convey no useful information for us to track. There should be examples of existing filters that you can copy here  if you grep for `//@.*= note`

DebugSteven: I didn't find this, but I got those lines filtered out. Let me know if it should be done differently.

PR #2841, opened 2023-04-13
compiletest: complain about unknown flags

RalfJung:
This would have avoided https://github.com/rust-lang/rust/issues/110102


oli-obk:
uh....

@bors ping

PR #2858, opened 2023-04-27
Bump libffi-sys to 2.3.0

heiher:
Bump libffi-sys to 2.3.0 that includes LoongArch support.

Thanks


RalfJung:
@bors r+

Thanks!
However note that this will not affect the Miri shipped with rustc, since our local lockfile is ignored in the rustc workspace.

PR #2867, opened 2023-05-01
Simplify event selection in TB diagnostics

Vanille-N:
As discussed previously, getting the range from `RangeMap` can make the filtering of events much simpler without any user-visible diff.

See minor exception in [<9d8fc00>](https://github.com/rust-lang/miri/commit/9d8fc00a4a57dde06a9c42b809afed827c3b0d5c) and decide how to resolve it
- add a boolean flag not to record events produced by deallocations ?
- add a `help: deallocation counts as an implicit write` ? (Note: could be generalized to also include `help: reborrow counts as an implicit read`)
- not bother and keep as-is ?
- something else ?

### DIFF
--- a/src/borrow_tracker/stacked_borrows/mod.rs
+++ b/src/borrow_tracker/stacked_borrows/mod.rs
@@ -459,7 +459,7 @@ impl<'tcx> Stack {
 impl Stacks {
     pub fn remove_unreachable_tags(&mut self, live_tags: &FxHashSet<BorTag>) {
         if self.modified_since_last_gc {
-            for stack in self.stacks.iter_mut_all() {
+            for (_stack_range, stack) in self.stacks.iter_mut_all() {
                 if stack.len() > 64 {
                     stack.retain(live_tags);
                 }
@@ -511,8 +511,8 @@ impl<'tcx> Stacks {
         ) -> InterpResult<'tcx>,
     ) -> InterpResult<'tcx> {
         self.modified_since_last_gc = true;
-        for (offset, stack) in self.stacks.iter_mut(range.start, range.size) {
-            let mut dcx = dcx_builder.build(&mut self.history, offset);
+        for (stack_range, stack) in self.stacks.iter_mut(range.start, range.size) {
+            let mut dcx = dcx_builder.build(&mut self.history, Size::from_bytes(stack_range.start));
             f(stack, &mut dcx, &mut self.exposed_tags)?;
             dcx_builder = dcx.unbuild();
         }
--- a/src/borrow_tracker/tree_borrows/diagnostics.rs
+++ b/src/borrow_tracker/tree_borrows/diagnostics.rs
@@ -3,7 +3,6 @@ use std::ops::Range;
 
 use rustc_data_structures::fx::FxHashMap;
 use rustc_span::{Span, SpanData};
-use rustc_target::abi::Size;
 
 use crate::borrow_tracker::tree_borrows::{
     perms::{PermTransition, Permission},
@@ -14,18 +13,30 @@ use crate::borrow_tracker::{AccessKind, ProtectorKind};
 use crate::*;
 
 /// Complete data for an event:
-/// - `kind` is what happened to the permissions
-/// - `access_kind` and `access_range` describe the access that caused the event
-/// - `offset` allows filtering only the relevant events for a given memory location
-/// (see how we perform the filtering in `History::extract_relevant`.
-/// - `span` is the line of code in question
 #[derive(Clone, Debug)]
 pub struct Event {
+    /// Transformation of permissions that occured because of this event
     pub transition: PermTransition,
+    /// Kind of the access that triggered this event
     pub access_kind: AccessKind,
+    /// Relative position of the tag to the one used for the access
     pub is_foreign: bool,
+    /// User-visible range of the access
     pub access_range: AllocRange,
-    pub offset: Size,
+    /// The transition recorded by this event only occured on a subrange of
+    /// `access_range`: a single access on `access_range` triggers several events,
+    /// each with their own mutually disjoint `transition_range`. No-op transitions
+    /// should not be recorded as events, so the union of all `transition_range` is not
+    /// necessarily the entire `access_range`.
+    ///
+    /// No data from any `transition_range` should ever be user-visible, because
+    /// both the start and end of `transition_range` are entirely dependent on the
+    /// internal representation of `RangeMap` which is supposed to be opaque.
+    /// What will be shown in the error message is the first byte `error_offset` of
+    /// the `TbError`, which should satisfy
+    /// `event.transition_range.contains(error.error_offset)`.
+    pub transition_range: Range<u64>,
+    /// Line of code that triggered this event
     pub span: Span,
 }
 
@@ -35,9 +46,9 @@ pub struct Event {
 /// Available filtering methods include `History::forget` and `History::extract_relevant`.
 #[derive(Clone, Debug)]
 pub struct History {
-    pub tag: BorTag,
-    pub created: (Span, Permission),
-    pub events: Vec<Event>,
+    tag: BorTag,
+    created: (Span, Permission),
+    events: Vec<Event>,
 }
 
 /// History formatted for use by `src/diagnostics.rs`.
@@ -60,12 +71,7 @@ impl HistoryData {
     // Format events from `new_history` into those recorded by `self`.
     //
     // NOTE: also converts `Span` to `SpanData`.
-    pub fn extend(
-        &mut self,
-        new_history: History,
-        tag_name: &'static str,
-        show_initial_state: bool,
-    ) {
+    fn extend(&mut self, new_history: History, tag_name: &'static str, show_initial_state: bool) {
         let History { tag, created, events } = new_history;
         let this = format!("the {tag_name} tag {tag:?}");
         let msg_initial_state = format!(", in the initial state {}", created.1);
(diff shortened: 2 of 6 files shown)


src/range_map.rs
```diff
@@ -63,7 +63,7 @@ impl<T> RangeMap<T> {
     /// through interior mutability.
     ///
     /// The iterator also provides the offset of the given element.
```
RalfJung, line 65:
```suggestion
    /// The iterator also provides the range of the given element.
    /// How exactly the ranges are split can differ even for otherwise identical
    /// maps, so user-visible behavior should never depend on the exact range.
```

src/range_map.rs
```diff
@@ -84,11 +84,11 @@ impl<T> RangeMap<T> {
         slice
             .iter()
             .take_while(move |elem| elem.range.start < end)
-            .map(|elem| (Size::from_bytes(elem.range.start), &elem.data))
+            .map(|elem| (elem.range.clone(), &elem.data))
     }
 
-    pub fn iter_mut_all(&mut self) -> impl Iterator<Item = &mut T> {
-        self.v.iter_mut().map(|elem| &mut elem.data)
+    pub fn iter_mut_all(&mut self) -> impl Iterator<Item = (ops::Range<u64>, &mut T)> {
```
RalfJung, line 90: Looks like no user of iter_mut_all actually needs the range, so maybe let's keep this how it used to be?

src/range_map.rs
RalfJung: (same for `iter_mut`)

RalfJung:
> See minor exception in [<9d8fc00>](https://github.com/rust-lang/miri/commit/9d8fc00a4a57dde06a9c42b809afed827c3b0d5c) and decide how to resolve it

I don't entirely understand what is going on here. Why does behavior change?

And is this state transition even relevant for that example? Strongly deallocated items always prohibit deallocation no matter their state (right?), so it seems like listing the state transitions makes sense for `the allocation of the accessed tag <TAG> also contains the strongly protected tag <TAG>` messages?

Vanille-N:
> And is this state transition even relevant for that example? Strongly deallocated items always prohibit deallocation no matter their state (right?), so it seems like listing the state transitions makes sense for `the allocation of the accessed tag <TAG> also contains the strongly protected tag <TAG>` messages?

Correct, that transition is not relevant at all to the error, which is why I hadn't noticed that it was missing.

We have other transitions that are shown yet not useful, like `Active -> Frozen` if the error is an attempted Read through `Disabled`, I can take the opportunity to more strictly filter the relevant events.

RalfJung:
That makes sense, but sounds orthogonal to this PR. And turns out `this corresponds to an activation` is also old, I just hadn't noticed it before. So let's leave those parts to a separate PR.

src/range_map.rs
Vanille-N: More accurately,
- callers of `iter_mut_all` need nothing
- callers of `iter_all` need the offset
- callers of `iter_mut` need the range
- callers of `iter` need nothing

Should we go for the minimum signature for each, or something more consistent across the 4 iterators ?

Vanille-N:
There are functions in concurrency that contain 3 ranges in the same scope, I hope I didn't misinterpret the meaning of each.

src/borrow_tracker/stacked_borrows/mod.rs
```diff
@@ -511,8 +511,8 @@ impl<'tcx> Stacks {
         ) -> InterpResult<'tcx>,
     ) -> InterpResult<'tcx> {
         self.modified_since_last_gc = true;
-        for (offset, stack) in self.stacks.iter_mut(range.start, range.size) {
-            let mut dcx = dcx_builder.build(&mut self.history, offset);
+        for (subrange, stack) in self.stacks.iter_mut(range.start, range.size) {
```
RalfJung, line 514: This could be `stack_range`

src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -14,18 +13,20 @@ use crate::borrow_tracker::{AccessKind, ProtectorKind};
...
+    /// Kind of the access that triggered this event
     pub access_kind: AccessKind,
+    /// Relative position of the tag to the one used for the access
     pub is_foreign: bool,
+    /// User-visible range of the access
     pub access_range: AllocRange,
-    pub offset: Size,
+    /// Internal only, used for filtering relevant events
+    /// (see `History::extract_relevant`)
+    pub access_subrange: Range<u64>,
```
RalfJung, line 28: If it's internal, why is it public?

Also this should be easy to explain, no? It is the range of offsets that the event refers to.

src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -243,7 +228,7 @@ pub(super) struct TbError<'node> {
     /// What failure occurred.
     pub error_kind: TransitionError,
     /// The byte at which the conflict occured.
```
RalfJung, line 230:
```suggestion
    /// The offset (into the allocation) at which the conflict occurred.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -349,7 +349,7 @@ impl<'tcx> Tree {
         span: Span, // diagnostics
     ) -> InterpResult<'tcx> {
         self.perform_access(AccessKind::Write, tag, range, global, span)?;
-        for (offset, perms) in self.rperms.iter_mut(range.start, range.size) {
+        for (subrange, perms) in self.rperms.iter_mut(range.start, range.size) {
```
RalfJung, line 352: What about `perms_range`?

src/concurrency/data_race.rs
```diff
@@ -851,25 +851,29 @@ impl VClockAlloc {
...
     ) -> InterpResult<'tcx> {
         let current_span = machine.current_span();
         let global = machine.data_race.as_ref().unwrap();
         if global.race_detecting() {
             let (index, mut clocks) = global.current_thread_state_mut(&machine.threads);
             let mut alloc_ranges = self.alloc_ranges.borrow_mut();
-            for (offset, range) in alloc_ranges.iter_mut(range.start, range.size) {
-                if let Err(DataRace) = range.read_race_detect(&mut clocks, index, current_span) {
+            for (access_subrange, other_range) in
+                alloc_ranges.iter_mut(access_range.start, access_range.size)
```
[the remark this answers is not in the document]
RalfJung: Ah dang, `clocks` already exists here. I think I'll make a PR renaming things in data_race, this is not your problem -- but it will conflict with your PR.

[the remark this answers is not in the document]
RalfJung: Please rebase onto https://github.com/rust-lang/miri/pull/2876.

src/range_map.rs
RalfJung: Marking as unresolved since the extended comments I asked for are not present in the current PR.

src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -14,18 +13,22 @@ use crate::borrow_tracker::{AccessKind, ProtectorKind};
...
     pub access_kind: AccessKind,
+    /// Relative position of the tag to the one used for the access
     pub is_foreign: bool,
+    /// User-visible range of the access
     pub access_range: AllocRange,
-    pub offset: Size,
+    /// The transition recorded by this event only occured on a subrange of
+    /// `access_range`: an access on `access_range` triggers an event
+    /// on `transition_range`.
+    /// Should never be user-visible.
```
RalfJung, line 29:
```suggestion
    /// The full range should never be user-visible; its `start` however is the
    /// first byte affected by this transition and okay to display in error messages.
```

src/borrow_tracker/stacked_borrows/mod.rs
```diff
@@ -459,7 +459,7 @@ impl<'tcx> Stack {
 impl Stacks {
     pub fn remove_unreachable_tags(&mut self, live_tags: &FxHashSet<BorTag>) {
         if self.modified_since_last_gc {
-            for stack in self.stacks.iter_mut_all() {
+            for (_, stack) in self.stacks.iter_mut_all() {
```
RalfJung, line 462:
```suggestion
            for (_stack_range, stack) in self.stacks.iter_mut_all() {
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -332,7 +332,7 @@ impl<'tcx> Tree {
         self.nodes.get_mut(parent_idx).unwrap().children.push(idx);
         // Initialize perms
         let perm = LocationState::new(default_initial_perm).with_access();
-        for (_range, perms) in self.rperms.iter_mut(range.start, range.size) {
+        for (_, perms) in self.rperms.iter_mut(reborrow_range.start, reborrow_range.size) {
```
RalfJung, line 335:
```suggestion
        for (_perms_range, perms) in self.rperms.iter_mut(reborrow_range.start, reborrow_range.size) {
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -530,7 +530,7 @@ impl Tree {
             // the tag from the mapping.
             let tag = node.tag;
             self.nodes.remove(idx);
-            for perms in self.rperms.iter_mut_all() {
+            for (_, perms) in self.rperms.iter_mut_all() {
```
RalfJung, line 533:
```suggestion
            for (_perms_range, perms) in self.rperms.iter_mut_all() {
```

RalfJung (commented):
just a few final nits :)

RalfJung:
Oh also this is still marked as a draft.  Did you even want another review?

RalfJung:
Okay, please squash this into one commit and then we can land this. :)

Vanille-N:
Note that the comment you suggested on `Event.transition_range` was inaccurate, so I fixed it in addition to squashing.

src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -14,18 +13,30 @@ use crate::borrow_tracker::{AccessKind, ProtectorKind};
...
+    /// each with their own mutually disjoint `transition_range`. No-op transitions
+    /// should not be recorded as events, so the union of all `transition_range` is not
+    /// necessarily the entire `access_range`.
+    ///
+    /// No data from any `transition_range` should ever be user-visible, because
+    /// both the start and end of `transition_range` are entirely dependent on the
+    /// internal representation of `RangeMap` which is supposed to be opaque.
+    /// What will be shown in the error message is the first byte `error_offset` of
+    /// the `TbError`, which should satisfy
+    /// `event.transition_range.start <= error.error_offset < event.transition_range.end`.
```
RalfJung, line 37:
```suggestion
    /// `event.transition_range.contains(error.error_offset)`.
```

RalfJung:
Ah, thanks for pointing that out!

src/borrow_tracker/tree_borrows/diagnostics.rs
Vanille-N: Oh nice. Then let's also use `.contains` in the `filter`.

Issue #2874, opened 2023-05-05
proc_macro crate missing from miri sysroot beginning nightly-2023-05-05


dtolnay:
```rust
// src/main.rs

extern crate proc_macro;

fn main() {}
```

```console
$ cargo +nightly-2023-05-04 miri run  # works
```

```console
$ cargo +nightly-2023-05-05 miri run
   Compiling testing v0.0.0
    Finished dev [unoptimized + debuginfo] target(s) in 0.11s
     Running `~/.rustup/toolchains/nightly-2023-05-05-x86_64-unknown-linux-gnu/bin/cargo-miri runner target/miri/x86_64-unknown-linux-gnu/debug/testing`

error[E0463]: can't find crate for `proc_macro`
 --> src/main.rs:1:1
  |
1 | extern crate proc_macro;
  | ^^^^^^^^^^^^^^^^^^^^^^^^ can't find crate

error: aborting due to previous error

For more information about this error, try `rustc --explain E0463`.
```

Rust-lang/rust commit range is https://github.com/rust-lang/rust/compare/473f916d8...74c482104. Probably https://github.com/rust-lang/rust/pull/108865 (FYI @Zoxc @jyn514)

jyn514:
Miri probably needs to change from building `test` to the new `sysroot` crate.

saethlin:
Ouch. I might look into this tomorrow, but for anyone else interested in fixing this, Miri uses this crate to build the sysroot: https://github.com/RalfJung/rustc-build-sysroot

RalfJung:
Yeah... that is the easy part though: https://github.com/RalfJung/rustc-build-sysroot/pull/6. The hard part is that this crate would like to support a range of rustc versions, not just the latest nightly.

workingjubilee:
I've posted https://github.com/rust-lang/rust/pull/111234 to hopefully give a bit more time to develop a migration plan.

RalfJung:
For Miri at least I think the migration is done when https://github.com/rust-lang/rust/pull/111236 lands.

PR #2875, opened 2023-05-05
Rustup

RalfJung:
This *should* fail because of https://github.com/rust-lang/miri/issues/2874 but it seems we don't actually cover that in our tests...


RalfJung:
All right, now there are tests.

@bors r+

PR #2877, opened 2023-05-05
miri-script: Transform Windows paths to Unix.

osiewicz:
python3 snippet used to fill $MIRIDIR variable returns native paths. Down the line in `bench` subcommand this leads to benchmark setup failure, preventing contributors from running benchmarks on Windows hosts.

This commit replaces usage of native os.path module with pathlib, which explicitly converts paths to Unix flavour.

pathlib is available from Python 3.4. There's also [a nice comparison of features](https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module) with os module.

On current master, running `./miri bench` under MinGW results in an error from cargo-metadata about path format.
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: CargoMetadata { stderr: "error: manifest path `D:Programmingmiri/bench-cargo-miri/backtraces/Cargo.toml` does not exist\n" }', src\util.rs:224:80
```
With this commit, it is not an issue anymore.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,6 +78,9 @@ jobs:
           rustc -Vv
           cargo -V
 
+      - name: Install hyperfine
+        run: cargo install hyperfine
+
       - name: Test
         run: bash ./ci.sh
 
--- a/ci.sh
+++ b/ci.sh
@@ -83,7 +83,7 @@ function run_tests {
   fi
   # Ensure that `./miri bench` works on all relevant platforms.
   # `slice-get-unchecked` is the least complex of all available benchmarks.
-  ./miri bench slice-get-unchecked
+  CARGO_EXTRA_FLAGS="" ./miri bench slice-get-unchecked
 
   endgroup
 }
--- a/ci.sh
+++ b/ci.sh
@@ -51,6 +51,11 @@ function run_tests {
     for FILE in tests/many-seeds/*.rs; do
       MIRI_SEEDS=64 CARGO_EXTRA_FLAGS="$CARGO_EXTRA_FLAGS -q" ./miri many-seeds ./miri run "$FILE"
     done
+
+    # Ensure that `./miri bench` works on all relevant platforms.
+    # `slice-get-unchecked` is the least complex of all available benchmarks.
+    # CARGO_EXTRA_FLAGS are cleared as `--locked` is already passed internally by miri bench.
+    CARGO_EXTRA_FLAGS="" ./miri bench slice-get-unchecked
   fi
 
   ## test-cargo-miri
@@ -81,9 +86,6 @@ function run_tests {
       cargo miri run --manifest-path bench-cargo-miri/$BENCH/Cargo.toml
     done
   fi
-  # Ensure that `./miri bench` works on all relevant platforms.
-  # `slice-get-unchecked` is the least complex of all available benchmarks.
-  CARGO_EXTRA_FLAGS="" ./miri bench slice-get-unchecked
 
   endgroup
 }
--- a/ci.sh
+++ b/ci.sh
@@ -81,6 +81,9 @@ function run_tests {
       cargo miri run --manifest-path bench-cargo-miri/$BENCH/Cargo.toml
     done
   fi
+  # Ensure that `./miri bench` works on all relevant platforms.
+  # `slice-get-unchecked` is the least complex of all available benchmarks.
+  ./miri bench slice-get-unchecked
 
   endgroup
 }
--- a/miri
+++ b/miri
@@ -77,7 +77,7 @@ if [ -z "$COMMAND" ]; then
 fi
 shift
 # macOS does not have a useful readlink/realpath so we have to use Python instead...
-MIRIDIR=$(python3 -c 'import os, sys; print(os.path.dirname(os.path.realpath(sys.argv[1])))' "$0")
+MIRIDIR=$(python3 -c 'import pathlib, sys; print(pathlib.Path(sys.argv[1]).resolve().parent.as_posix())' "$0")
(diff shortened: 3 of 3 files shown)


oli-obk:
Thanks! Could we test a trivial fast benchmark in CI just to make sure it works? Or will that take too long due to compiling in release mode?

osiewicz:
Running cargo clean followed by `./miri bench slice-get-unchecked` takes about 90 seconds on my PC, though as you've said, majority of that time is spent compiling Miri. CI run will take a bit longer, but IMHO it looks feasible to do that anyways.
Would you like `./miri bench` to run as part of `build` pipeline or do you have something else in mind?

oli-obk:
Yea, just part of the build pipeline makes sense. We'll see on the PR CI if there's a relevant increase in time

osiewicz:
So it seems like current hyperfine release doesn't work on i686-pc-windows-msvc target. The fix is available, but not released (yet). `cargo install --target i686-pc-windows-msvc hyperfine` fails and `cargo install --target i686-pc-windows-msvc --git https://github.com/sharkdp/hyperfine.git` works just fine. 
I will investigate failures from linux/macos. It seems like this might be blocked by hyperfine release anyways.

ci.sh
```diff
@@ -81,6 +81,9 @@ function run_tests {
       cargo miri run --manifest-path bench-cargo-miri/$BENCH/Cargo.toml
     done
   fi
+  # Ensure that `./miri bench` works on all relevant platforms.
+  # `slice-get-unchecked` is the least complex of all available benchmarks.
+  CARGO_EXTRA_FLAGS="" ./miri bench slice-get-unchecked
```
RalfJung, line 86: This is the wrong place to put the check. This here will be run *for each target that we check*. It needs to either be gated by `-z "${MIRI_TEST_TARGET+exists}"` (to only run on the host) or be moved elsewhere.

RalfJung: Also why the `CARGO_EXTRA_FLAGS=""`?

.github/workflows/ci.yml
```diff
@@ -78,6 +78,9 @@ jobs:
           rustc -Vv
           cargo -V
 
+      - name: Install hyperfine
+        run: cargo install hyperfine
```
RalfJung, line 82: This looks like it will cost us a minute. :/ Are there binary releases we could install?

Also the Linux run in this PR took almost 1h which is a lot slower than what I remember... is that a coincidence or related to this PR?

ci.sh
osiewicz: > Also why the `CARGO_EXTRA_FLAGS=""`?

Since CARGO_EXTRA_FLAGS are set to `--locked`, I've cleared them to avoid issue with `--locked` being passed twice (which manifested itself [in CI run of ca60799 commit](https://github.com/rust-lang/miri/actions/runs/4896227247/jobs/8742786560#step:9:1533)). Is there a better way to do that? Perhaps I could document why these flags are cleared?

.github/workflows/ci.yml
osiewicz: I'm not aware of binary distributions of hyperfine.

Perhaps we could install `dev` version, as we don't really care about measurements?
Though I'd expect hyperfine to be quick to build.

osiewicz: It looks like Linux run is back to normal. Perhaps running `miri bench` only for host targets is a positive factor?

miri
```diff
@@ -77,7 +77,7 @@ if [ -z "$COMMAND" ]; then
 fi
 shift
 # macOS does not have a useful readlink/realpath so we have to use Python instead...
-MIRIDIR=$(python3 -c 'import os, sys; print(os.path.dirname(os.path.realpath(sys.argv[1])))' "$0")
+MIRIDIR=$(python3 -c 'import pathlib, sys; print(pathlib.Path(sys.argv[1]).resolve().parent.as_posix())' "$0")
```
osiewicz, line 80: By the way, what do you think about detecting python binary to run (similar to [what's done in ci.sh](https://github.com/rust-lang/miri/blob/master/ci.sh#L57))? On my Windows machine `python3` does not work out of the box, so I guess it might be worthwhile to do that.

ci.sh
RalfJung: Ah, that sounds like a bug in our miri script... maybe we just shouldn't have the `--locked` in the `install` command.

.github/workflows/ci.yml
RalfJung: It's about a minute to build. So not too bad I guess...

OTOH we don't actually care about the benchmarking part on CI, so we could as a hack also have CI create some `hyperfine` shell script that just executes its last argument. At least that would work on Linux. Whether it makes sense on Windows I cannot say.

miri
RalfJung: Oh no why is everything so terrible. :(

We don't have to use python for this but I found no better way to make it work on macOS.

Really we should rewrite the entire damn thing in Rust I guess if we truly want it to be portable...

osiewicz: > Really we should rewrite the entire damn thing in Rust I guess if we truly want it to be portable...

I'd be happy to tackle that in the future PR. Is there a tracking issue for that effort?

RalfJung:
Looks like hyperfine doesn't even build on Windows currently?

I have implemented the alternative sketched above that avoids the need for hyperfine on CI in https://github.com/rust-lang/miri/pull/2908. I'll add your commit fixing Windows to that PR, and close this one. Thanks a lot for your help!

osiewicz:
New version of hyperfine just dropped (with proper Windows support): https://github.com/sharkdp/hyperfine/releases/tag/v0.17.0

PR #2879, opened 2023-05-07
Include the current Crate name in the measureme output name

saethlin:
See https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/measureme.20flamegraph.20panics/near/356367013
cc @andjo403

Currently, attempting to use `MIRIFLAGS=-Zmiri-measureme=miri cargo miri test` on a crate with multiple test targets (which is very common) will produce a corrupted measureme output file, because the various interpreter processes will stomp each other's output.

This change does not entirely prevent this, but the various test targets seem to always have different crate names, so if nothing else this will make the broken measureme files much harder to encounter by accident, while also making it clear what they are all for.

### DIFF
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -4,6 +4,8 @@
 use std::borrow::Cow;
 use std::cell::RefCell;
 use std::fmt;
+use std::path::Path;
+use std::process;
 
 use rand::rngs::StdRng;
 use rand::SeedableRng;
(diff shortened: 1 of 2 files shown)


RalfJung:
Does this need a note in the docs somewhere?

bjorn3:
Rustc includes it's own pid in the filename. This prevents overwriting files for identically named crates (like build_script_build) and profiles from previous runs.

src/machine.rs
```diff
@@ -498,7 +500,17 @@ impl<'mir, 'tcx> MiriMachine<'mir, 'tcx> {
...
-            measureme::Profiler::new(out).expect("Couldn't create `measureme` profiler")
+            let crate_name = layout_cx
+                .tcx
+                .sess
+                .opts
+                .crate_name
+                .clone()
+                .unwrap_or_else(|| "unknown-crate".to_string());
+            let pid = process::id();
+            let filename = format!("{crate_name}-{pid:07}");
```
RalfJung, line 511: Why are you padding the PID?

(Other tools using the PID in the filename, such as callgrind, AFAIK do not pad it.)

RalfJung: Then please also copy the comment from rustc that explains this. :)

saethlin: The justification in rustc does not apply to Miri. It is explained in this hackmd: https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#ASLR-it%E2%80%99s-free-entropy The comment in the source assumes you know the context in the hackmd, which I didn't know until I asked other people.

We aren't doing high-precision instruction counting with ASLR off. I copied the implementation from rustc just for consistency, nothing else.

RalfJung:
Makes sense, thanks!
@bors r+

PR #2884, opened 2023-05-09
Update to latest ui_test crate version.

oli-obk:
Also stops using github actions groups that conflict with our groups as github does not nest them

### DIFF
--- a/tests/compiletest.rs
+++ b/tests/compiletest.rs
@@ -46,19 +46,9 @@ fn build_so_for_c_ffi_tests() -> PathBuf {
 }
 
 fn run_tests(mode: Mode, path: &str, target: &str, with_dependencies: bool) -> Result<()> {
-    let mut config = Config {
-        target: Some(target.to_owned()),
-        stderr_filters: STDERR.clone(),
-        stdout_filters: STDOUT.clone(),
-        root_dir: PathBuf::from(path),
-        mode,
-        program: CommandBuilder::rustc(),
-        quiet: false,
-        edition: Some("2021".into()),
-        ..Config::default()
-    };
-
-    config.program.program = miri_path();
+    // Miri is rustc-like, so we create a default builder for rustc and modify it
+    let mut program = CommandBuilder::rustc();
+    program.program = miri_path();
 
     let in_rustc_test_suite = option_env!("RUSTC_STAGE").is_some();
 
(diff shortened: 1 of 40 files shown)


tests/compiletest.rs
```diff
@@ -50,33 +52,35 @@ fn run_tests(mode: Mode, path: &str, target: &str, with_dependencies: bool) -> R
         stdout_filters: STDOUT.clone(),
         root_dir: PathBuf::from(path),
         mode,
-        program: miri_path(),
+        program: CommandBuilder::rustc(),
```
RalfJung, line 55: Might be worth a comment that this is fixed up to use miri later?

Does this indicate a missing API for constructing a rustc-like CommandBuilder with a custom path?

RalfJung: Well I guess we are fixing up the args as well so :shrug:

RalfJung, line 55:
```suggestion
        // Will be set up to call Miri below.
        program: CommandBuilder::rustc(),
```

tests/compiletest.rs
```diff
@@ -135,7 +139,12 @@ fn run_tests(mode: Mode, path: &str, target: &str, with_dependencies: bool) -> R
             "run".into(), // There is no `cargo miri build` so we just use `cargo miri run`.
         ];
     }
-    ui_test::run_tests(config)
+    ui_test::run_tests_generic(
+        config,
+        |path| path.extension().map(|ext| ext == "rs").unwrap_or(false),
```
RalfJung, line 144: Could use `is_Some_and` :)

tests/compiletest.rs
```diff
@@ -135,7 +139,12 @@ fn run_tests(mode: Mode, path: &str, target: &str, with_dependencies: bool) -> R
...
+        |_, _| None,
```
RalfJung, line 145: Here I wish we had named arguments...

tests/compiletest.rs
```diff
@@ -235,3 +244,43 @@ fn main() -> Result<()> {
 
     Ok(())
 }
+
+struct TextAndGha;
```
RalfJung, line 248: This should have a comment explaining what this is about.

tests/compiletest.rs
RalfJung: For now, I think comments about what the arguments here are doing are sufficient.

PR #2887, opened 2023-05-11
TB: more fail tests (mostly shared with SB)

Vanille-N:
Although it was not in the tests, `mem::transmute` works for `UnsafeCell -> &` as well.

Draft: will also introduce more test cases for cases that fail.
Draft: depends on the new error messages from #2888


tests/pass/tree-borrows/transmute-unsafecell.rs
```diff
@@ -1,13 +1,32 @@
...
+//! that violate immutability.
+
 use core::cell::UnsafeCell;
 use core::mem;
 
 fn main() {
     unsafe {
-        let x = &0i32;
-        // As long as we only read, transmuting this to UnsafeCell should be fine.
-        let cell_x: &UnsafeCell<i32> = mem::transmute(&x);
```
oli-obk, line 10: O_o this was using a transmute on `&&i32` to `&UnsafeCell<i32>` whoops

tests/pass/tree-borrows/transmute-unsafecell.rs
```diff
@@ -1,13 +1,32 @@
...
-        let _val = *cell_x.get();
+        ref_to_cell();
+        cell_to_ref();
     }
 }
+
+// Pretend that the reference has interior mutability.
+// Don't actually mutate it though, it will fail because it has a Frozen parent.
```
oli-obk, line 18: do we have tests for that? Similarly for the other direction, mutating the unsafe cell should invalidate reading from the direct reference

tests/pass/tree-borrows/transmute-unsafecell.rs
Vanille-N: No comment. That's what the `assert!` was added for ;)

Vanille-N:
Selection of `fail` tests to run both TB and SB is done, I'll push when I've verified all the errors

RalfJung:
Oh, you are doing this for all fail tests now, wow. Time to rename the PR I guess? :D

Vanille-N:
Error messages updated after the merge of #2888.

Waiting for comments on
- structure (I picked `tests/fail/{tree,stacked,both}_borrows` which minimizes files moved around, but we could also replicate the `src` structure with `tests/fail/borrow_tracker/{tree_borrows,stacked_borrows,}`
- coverage (is this enough tests ? are there more things that should be tested ?)
- other wording details

Vanille-N:
Some renames have been wrongly detected as deletion+creation. I can eventually make sure that git properly sees that it's all already existing files

RalfJung:
> Some renames have been wrongly detected as deletion+creation. I can eventually make sure that git properly sees that it's all already existing files

git decides this via some heuristic, the data model doesn't even distinguish these cases so this is `git diff` making the decisions. I am not aware of a way to fix this. It's also not a big deal though.

RalfJung:
> structure (I picked tests/fail/{tree,stacked,both}_borrows which minimizes files moved around, but we could also replicate the src structure with tests/fail/borrow_tracker/{tree_borrows,stacked_borrows,}

Hm... in `pass` we have `{tree,stacked}-borrows` so let's match that I guess? Though `_` seems more consistent indeed.

`both_borrows` sounds a bit odd but I also can't think of anything better. :shrug: 

> coverage (is this enough tests ? are there more things that should be tested ?)

Coverage looks pretty good to me. Which tests did you not copy over from fail/stacked_borrows?

Vanille-N:
Oh no, need to rebase.
[`Fixes (2ebc752)`](https://github.com/rust-lang/miri/pull/2887/commits/2ebc7525147fbcd12c399284307b20bff1c60d23) is the first commit after we talked yesterday.

The only conflict seems to be the addition of `#![allow(cast_ref_to_mut)]` in `tests/fail/shr_frozen_violation1.rs`, it has been replicated in `tests/fail/both_borrows/shr_frozen_violation1.rs`. The newly added `tests/fail/tree_borrows/write_to_shr.rs` exhibits exactly the same issue and has also received an `#[allow(_)]`.

tests/fail/tree_borrows/parent_write_invalidates_read.rs
```diff
@@ -0,0 +1,14 @@
+//@compile-flags: -Zmiri-tree-borrows
+
+// Read to local variable kills reborrows *and* raw pointers derived from them.
```
RalfJung, line 3: "read *from* local variable"

tests/fail/tree_borrows/raw_write_invalidates_read.rs
```diff
@@ -0,0 +1,12 @@
...
+// Write to raw pointer invalidates all shared references derived from it.
+fn main() {
+    let mut root = 6u8;
+    let mref = &mut root;
+    let ptr = mref as *mut u8;
+
+    let shr1 = unsafe { &*ptr };
+    unsafe { *ptr = 2; } // Parent Write
+    let _v = *shr1; //~ ERROR: /read access through .* is forbidden/
+}
```
RalfJung, line 14: I like the test but not the filename. "raw write" isn't the key part... and it's also "parent write invalidates read" but that file already exists. ;)

What about parent_write_invalidates_mut and parent_write_invalidates_shr?

tests/fail/tree_borrows/write_to_shr.rs
```diff
@@ -0,0 +1,11 @@
...
+
+// Very simple test that a shared reference without interior mutability cannot
+// be written to.
+
+pub fn main() {
+    let x = &mut 0u64;
+    let xref = unsafe { &*(x as *mut u64) };
+    let xmut = unsafe { &mut*(xref as *const u64 as *mut u64) };
+    *xmut = 31; //~ ERROR: /write access through .* is forbidden/
+}
```
RalfJung, line 13: This is a duplicate of shr_frozen_violation1, isn't it?

tests/fail/tree_borrows/raw_write_invalidates_read.rs
RalfJung: In fact isn't this test here a duplicate of shr_frozen_violation2?

tests/fail/tree_borrows/parent_write_invalidates_read.rs
RalfJung: The code does a write though, not a read...

src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -21,6 +38,8 @@ pub struct Event {
     pub access_kind: AccessKind,
     /// Relative position of the tag to the one used for the access
     pub is_foreign: bool,
+    /// Whether this access was explicit or inserted implicitly by Tree Borrows.
+    pub access_cause: AccessCause,
```
RalfJung, line 42: I think I would prefer to leave the diagnostic changes to another PR, this one is getting too big... also then the diff in that PR makes it much easier to evaluate the change.

tests/fail/tree_borrows/parent_write_invalidates_read.rs
Vanille-N: Oh no, copy-pasted too eagerly

tests/fail/tree_borrows/raw_write_invalidates_read.rs
Vanille-N: > `parent_write_invalidates_mut_raw`, `parent_write_invalidates_shr`

Not sure about those names. It's not a parent write that invalidates a raw mutable pointer, it's a write _to a raw mutable parent_ that invalidates a shr.

It's also not quite the same as `shr_frozen_violation2`:
```
parent_write_invalidates_shr:
    parent                                    <--- (1) write
         |--- shr child
               = raw child                 <---  (2) read
         
raw_write_invalidates_read:
    parent
    = raw parent                      <--- (1) write
         |--- shr child                 <--- (2) read

shr_frozen_violation2:
    parent                                <--- (1) write
    = raw parent
        |--- shr child                 <--- (2) read
```
so on the surface those three tests all check that a parent write invalidates a child read, but also they check that the parent-child link is in the right place in a chain of reference-raw-reference.

I can accept though that this might not be the place to test that and I'm not opposed to deleting `raw_write_invalidates_read` in favor of `shr_frozen_violation2`.

tests/fail/tree_borrows/write_to_shr.rs
Vanille-N: Indeed

tests/fail/tree_borrows/raw_write_invalidates_read.rs
RalfJung: Given that raw ptrs inherit their parent tag, those all look exactly identical? Write to parent, then read from frozen child?

RalfJung:
That is everything we discussed taken care of, I assume?
All right, then please squash the history a bit (or entirely). Then you can run `bors r=RalfJung`.

@bors delegate

RalfJung:
Or you could if I could type the right command...
@bors delegate+

RalfJung:
This is ready to land now anyway. :shrug: Let's see if that works...
@bors r+

PR #2888, opened 2023-05-11
TB diagnostics: avoid printing irrelevant events

Vanille-N:
History contains some events that are relevant to the location but not useful to understand the error.
We can make the selection of events more precise, from only "does it affect the location" to also "is it relevant for this kind of error"

This is also the occasion to fix https://github.com/rust-lang/miri/pull/2867#issuecomment-1530065511

[Solved] Draft: find a way for blanks in the history to not be confusing, as with the current version the history can show the creation as `Reserved` then show where it transitioned from `Frozen` to `Disabled`, but it will say nothing of the `Reserved -> Frozen` leading up to it.

### DIFF
--- a/src/borrow_tracker/tree_borrows/perms.rs
+++ b/src/borrow_tracker/tree_borrows/perms.rs
@@ -1,6 +1,7 @@
 use std::cmp::{Ordering, PartialOrd};
 use std::fmt;
 
+use crate::borrow_tracker::tree_borrows::diagnostics::TransitionError;
 use crate::borrow_tracker::tree_borrows::tree::AccessRelatedness;
 use crate::borrow_tracker::AccessKind;
 
@@ -115,26 +116,31 @@ mod transition {
 /// Public interface to the state machine that controls read-write permissions.
 /// This is the "private `enum`" pattern.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct Permission(PermissionPriv);
+pub struct Permission {
+    inner: PermissionPriv,
+}
 
 /// Transition from one permission to the next.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct PermTransition(PermissionPriv, PermissionPriv);
+pub struct PermTransition {
+    from: PermissionPriv,
+    to: PermissionPriv,
+}
 
 impl Permission {
     /// Default initial permission of the root of a new tree.
     pub fn new_root() -> Self {
-        Self(Active)
+        Self { inner: Active }
     }
 
     /// Default initial permission of a reborrowed mutable reference.
     pub fn new_unique_2phase(ty_is_freeze: bool) -> Self {
-        Self(Reserved { ty_is_freeze })
+        Self { inner: Reserved { ty_is_freeze } }
     }
 
     /// Default initial permission of a reborrowed shared reference
     pub fn new_frozen() -> Self {
-        Self(Frozen)
+        Self { inner: Frozen }
     }
 
     /// Apply the transition to the inner PermissionPriv.
@@ -144,9 +150,9 @@ impl Permission {
         old_perm: Self,
         protected: bool,
     ) -> Option<PermTransition> {
-        let old_state = old_perm.0;
+        let old_state = old_perm.inner;
         transition::perform_access(kind, rel_pos, old_state, protected)
-            .map(|new_state| PermTransition(old_state, new_state))
+            .map(|new_state| PermTransition { from: old_state, to: new_state })
     }
 }
 
@@ -155,26 +161,27 @@ impl PermTransition {
     /// should be possible, but the same is not guaranteed by construction of
     /// transitions inferred by diagnostics. This checks that a transition
     /// reconstructed by diagnostics is indeed one that could happen.
-    fn is_possible(old: PermissionPriv, new: PermissionPriv) -> bool {
-        old <= new
+    fn is_possible(self) -> bool {
+        self.from <= self.to
     }
 
-    pub fn from(old: Permission, new: Permission) -> Option<Self> {
-        Self::is_possible(old.0, new.0).then_some(Self(old.0, new.0))
+    pub fn from(from: Permission, to: Permission) -> Option<Self> {
+        let t = Self { from: from.inner, to: to.inner };
+        t.is_possible().then_some(t)
     }
 
     pub fn is_noop(self) -> bool {
-        self.0 == self.1
+        self.from == self.to
     }
 
     /// Extract result of a transition (checks that the starting point matches).
     pub fn applied(self, starting_point: Permission) -> Option<Permission> {
-        (starting_point.0 == self.0).then_some(Permission(self.1))
+        (starting_point.inner == self.from).then_some(Permission { inner: self.to })
     }
 
     /// Extract starting point of a transition
     pub fn started(self) -> Permission {
-        Permission(self.0)
+        Permission { inner: self.from }
     }
 
     /// Determines whether a transition that occured is compatible with the presence
@@ -190,10 +197,9 @@ impl PermTransition {
     /// };
     /// ```
     pub fn is_allowed_by_protector(&self) -> bool {
-        let &Self(old, new) = self;
-        assert!(Self::is_possible(old, new));
-        match (old, new) {
-            _ if old == new => true,
+        assert!(self.is_possible());
+        match (self.from, self.to) {
+            _ if self.from == self.to => true,
             // It is always a protector violation to not be readable anymore
             (_, Disabled) => false,
             // In the case of a `Reserved` under a protector, both transitions
@@ -204,15 +210,8 @@ impl PermTransition {
             (Reserved { .. }, Active) | (Reserved { .. }, Frozen) => true,
             // This pointer should have stayed writeable for the whole function
             (Active, Frozen) => false,
-            _ => unreachable!("Transition from {old:?} to {new:?} should never be possible"),
+            _ => unreachable!("Transition {} should never be possible", self),
         }
-    }
-
-    /// Composition function: get the transition that can be added after `app` to
-    /// produce `self`.
-    pub fn apply_start(self, app: Self) -> Option<Self> {
-        let new_start = app.applied(Permission(self.0))?;
-        Self::from(new_start, Permission(self.1))
     }
 }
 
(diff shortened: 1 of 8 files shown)


RalfJung:
You wrote
> raft because I think the current version of error messages can still be confusing but I'm not sure how best to resolve that

What is an example that you still consider confusing?

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -271,7 +343,7 @@ pub mod diagnostics {
         pub fn summary(&self) -> &'static str {
             assert!(Self::is_possible(self.0, self.1));
             match (self.0, self.1) {
-                (_, Active) => "an activation",
+                (_, Active) => "the first write to a 2-phase borrowed pointer",
```
RalfJung, line 346: "pointer" often indicates "raw pointer" so we should probably avoid it here.
```suggestion
                (_, Active) => "the first write to a 2-phase borrowed mutable reference",
```

Vanille-N:
> You wrote
> 
> > raft because I think the current version of error messages can still be confusing but I'm not sure how best to resolve that
> 
> What is an example that you still consider confusing?

The pattern occurs in `error-range.rs`:
```
help: the conflicting tag <TAG> was created here, in the initial state Reserved
help: the conflicting tag <TAG> later transitioned from Frozen to Disabled due to a foreign write access at offsets [0x5..0x6]
```
I think it should acknowledge somehow that at some point the `Reserved -> Frozen` transition occurred, if only to tell the user that it's definitely not relevant to the error.

RalfJung:
It could just say "transitioned to Disabled" and thus avoid the mention of the (irrelevant) frozen.

Or you could add a line like
```
help: (irrelevant transitions omitted)
```

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -271,14 +265,93 @@ pub mod diagnostics {
...
+
+        /// Not all transitions are relevant to all errors, this function filters those out.
+        ///
+        /// For example we don't care about the activation of a pointer when the error is
+        /// because it is not readable. All we care about is when it lost read permissions.
+        /// Return value to be interpreted as
+        /// - `None`: not a relevant transition
+        /// - `Some(v)`: relevant transition, with the additional help messages `v`
+        ///
+        /// Are considered unreachable
```
RalfJung, line 285: Why 'unreachable'? should this be 'irrelevant'?

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -271,14 +265,93 @@ pub mod diagnostics {
...
+        /// - any errors that involve noop transitions, as those can never be the cause of the error
+        /// - other configurations that are known to never cause errors (e.g. `Active` cannot cause a
+        /// `ChildAccessForbidden` error because all child accesses are allowed on `Active`)
+        /// - violations of `<` for permissions: `self` is assumed to be anterior to `err`
```
RalfJung, line 289: What does 'anterior' mean here?

Vanille-N: The transition represented by `self` should occur before the error, so we're never considering things like `self = (Frozen, Disabled)` while `err = ProtectedTransition(Active, Frozen)` because that would imply that somewhere a permission did `Frozen -> Disabled -> ??? -> Active -> Frozen` which is impossible.

RalfJung:
Still has the draft label, so @rustbot author

Vanille-N:
It's probably better to get this one before #2887 , that way we'll have the correct error messages directly.

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -271,14 +265,99 @@ pub mod diagnostics {
...
+            if self.is_noop() {
+                return None;
+            }
+            match err {
+                TransitionError::ChildAccessForbidden(insufficient) => {
+                    // Show where the permission was gained then lost,
+                    // but ignore unrelated permissions.
+                    // This eliminates transitions like `Active -> Frozen`
+                    // when the error is a failed `Read`.
+                    match (self.0, self.1, insufficient.0) {
```
RalfJung, line 315: `self.0` is never actually matched here, is it?

Also I have no idea what all these permissions mean.^^ I guess self.1 is the `to` permission of this transition. What's `insufficient.0`?

(That is the problem with tuple structs: the fields have no names that could explain their meaning.)

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -271,14 +265,99 @@ pub mod diagnostics {
...
             }
         }
+
+        /// Not all transitions are relevant to all errors, this function filters those out.
+        ///
+        /// For example we don't care about the activation of a pointer when the error is
+        /// because it is not readable. All we care about is when it lost read permissions.
+        /// Return value to be interpreted as
+        /// - `None`: not a relevant transition
+        /// - `Some(v)`: relevant transition, with the additional help messages `v`
```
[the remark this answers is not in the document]
Vanille-N: None for now, at the time of writing the code I thought we might want some. Any ideas, or should I remove this ?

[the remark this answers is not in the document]
RalfJung: Seems better to just return a `bool` for now then.

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -312,17 +317,23 @@ pub mod diagnostics {
...
+                        (Active, Frozen) => Some(vec![]),
+                        (Disabled, Disabled) => Some(vec![]),
+                        // `Active` and `Reserved` have all permissions, so a
+                        // `ChildAccessForbidden(Reserved | Active)` can never exist.
+                        (_, Active) | (_, Reserved { .. }) => unreachable!("cannot cause an error"),
+                        // No transition has `Reserved` as its `.to` unless it's a noop.
+                        (Reserved { .. }, _) => unreachable!("noop transition"),
+                        // A pointer being `Disabled` is a strictly stronger source of
+                        // errors than it being `Frozen`. If we try to access a `Disabled`,
+                        // then where it became `Frozen` is the least of our concerns for now.
```
RalfJung, line 331:
```suggestion
                        // errors than it being `Frozen`. If we try to access a `Disabled`,
                        // then where it became `Frozen` (or `Active`) is the least of our concerns for now.
```

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -269,16 +268,119 @@ pub mod diagnostics {
...
+                (_, Active) => "the first write to a 2-phase borrowed mutable reference",
                 (_, Frozen) => "a loss of write permissions",
                 (Frozen, Disabled) => "a loss of read permissions",
                 (_, Disabled) => "a loss of read and write permissions",
                 (old, new) =>
                     unreachable!("Transition from {old:?} to {new:?} should never be possible"),
             }
         }
+
+        /// Not all transitions are relevant to all errors, this function filters those out.
```
RalfJung, line 282:
```suggestion
        /// Determines whether `self` is a relevant transition for the error `err`.
        /// Not all transitions are relevant to all errors, this function filters those out.
```

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -269,16 +268,115 @@ pub mod diagnostics {
...
+                        (_, Disabled, _) => unreachable!("noop transition"),
+                        // These imply that the permission became `Disabled`
+                        // (resp. `Frozen`, `Active`) and was later found `Reserved | Active | Frozen`
+                        // (resp. `Reserved | Active`, `Reserved`), which is impossible because
+                        // permissions only evolve in the order `Reserved -> Active -> Frozen -> Disabled`.
+                        (Disabled, Reserved { .. } | Active | Frozen, _)
+                        | (Frozen, Reserved { .. } | Active, _)
+                        | (Active, Reserved { .. }, _) =>
+                            unreachable!("violates monotony of permissions"),
+                        // If the error is a transition `Frozen -> Disabled`, then we don't really
```
RalfJung, line 358: This says `Frozen -> Disabled` but the code actually checks `Frozen -> _`. Or am I missing something?

Vanille-N: Would you prefer an explicit
```rs
| (Active, Frozen, Disabled) => false,
| (Active, Frozen, _) => unreachable!(),
```

or just a comment that the `_` is guaranteed to be `Disabled` ?

Vanille-N: Related: is it better if there is complete coverage of the `unreachable!()` cases, or is it fine if some of them would actually pass if accidentally reachable ?

RalfJung: I'd prefer the former, given that the function already has tons of panics asserting the transition makes sense.

> Related: is it better if there is complete coverage of the unreachable!() cases, or is it fine if some of them would actually pass if accidentally reachable ?


Depends on how complex the code becomes if it wants to cover everything unreachable. E.g. I wouldn't want to `match` on fields that are currently entirely ignored just to add some assertions.

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -317,19 +317,21 @@ pub mod diagnostics {
...
                         (Active | Frozen, Disabled) => false,
-                        // Implies that the permission became Disabled and was _later_
-                        // found Frozen, which is impossible because permissions only
+
+                        // `Active` and `Reserved` have all permissions, so a
+                        // `ChildAccessForbidden(Reserved | Active)` can never exist.
+                        (_, Active) | (_, Reserved { .. }) => unreachable!("cannot cause an error"),
+                        // No transition has `Reserved` as its `.to` unless it's a noop.
+                        (Reserved { .. }, _) => unreachable!("self is a noop transition"),
                         // evolve in the order `Reserved -> Active -> Frozen -> Disabled`.
```
RalfJung, line 330: Something went wrong here, this sentence lost its beginning?

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -339,30 +341,37 @@ pub mod diagnostics {
...
+                        (_, Disabled, Disabled) | (_, Frozen, Frozen) | (_, Active, Active) =>
+                            unreachable!("err contains a noop transition"),
+
+                        // Permissions only evolve in the order `Reserved -> Active -> Frozen -> Disabled`,
+                        // so permissions found must be increasing in the order
+                        // `self.from < self.to <= forbidden.from < forbidden.to`.
+                        (Disabled, Reserved { .. } | Active | Frozen, _)
+                        | (Frozen, Reserved { .. } | Active, _)
+                        | (Active, Reserved { .. }, _) =>
+                            unreachable!("violates monotony of permissions between self and err"),
```
RalfJung, line 370: What is "monotony of permissions"? I think "monotonicity" would be grammatically correct but that is still a strange use of the term I think. This is just describing the allowed transitions between states, right?

RalfJung:
Thanks! Can you squash the history a bit? Then we can land this.

Vanille-N:
@RalfJung bump ?

RalfJung:
Sorry I didn't see this -- GH doesn't reliably send notifications on force-pushes, so please always leave a message when the next round of review is due. :)

@bors r+

PR #2890, opened 2023-05-11
Add `./miri run-dep` for running a file with test dependencies available

### LINKED ISSUE #2443: Provide a way to run a test that needs an external dependency
It would be pretty neat if we had a way to have `./miri run` work on tests that need external dependencies -- right now, these are a bit annoying to debug (and for tests that call into libc, this is a regression compared to the state before we added support for test dependencies). This should probably be a separate `./miri run-dep` or so, and as much as possible it should share the test-dependencies infrastructure with ui_test/compiletest.

oli-obk:
fixes #2443

### DIFF
--- a/miri
+++ b/miri
@@ -330,7 +330,7 @@ run|run-dep)
     # Then run the actual command.
     
     if [ "$COMMAND" = "run-dep" ]; then
-        exec $CARGO test --test compiletest -- miri-run-dep-mode $CARGO_EXTRA_FLAGS --manifest-path "$MIRIDIR"/Cargo.toml -- $MIRIFLAGS "$@"
+        exec $CARGO test --test compiletest $CARGO_EXTRA_FLAGS --manifest-path "$MIRIDIR"/Cargo.toml -- --miri-run-dep-mode $MIRIFLAGS "$@"
     else
         exec $CARGO run $CARGO_EXTRA_FLAGS --manifest-path "$MIRIDIR"/Cargo.toml -- $MIRIFLAGS "$@"
     fi
--- a/miri
+++ b/miri
@@ -306,7 +306,7 @@ test|bless)
     # Only in root project as `cargo-miri` has no tests.
     $CARGO test $CARGO_EXTRA_FLAGS --manifest-path "$MIRIDIR"/Cargo.toml "$@"
     ;;
-run)
+run|run-dep)
     # Scan for "--target" to overwrite the "MIRI_TEST_TARGET" env var so
     # that we set the MIRI_SYSROOT up the right way.
     FOUND_TARGET_OPT=0
@@ -323,11 +323,17 @@ run)
         # Make sure Miri actually uses this target.
         MIRIFLAGS="$MIRIFLAGS --target $MIRI_TEST_TARGET"
     fi
+
     # First build and get a sysroot.
     $CARGO build $CARGO_EXTRA_FLAGS --manifest-path "$MIRIDIR"/Cargo.toml
     find_sysroot
     # Then run the actual command.
-    exec $CARGO run $CARGO_EXTRA_FLAGS --manifest-path "$MIRIDIR"/Cargo.toml -- $MIRIFLAGS "$@"
+    
+    if [ "$COMMAND" = "run-dep" ]; then
+        exec $CARGO test --test compiletest -- miri-run-dep-mode $CARGO_EXTRA_FLAGS --manifest-path "$MIRIDIR"/Cargo.toml -- $MIRIFLAGS "$@"
+    else
+        exec $CARGO run $CARGO_EXTRA_FLAGS --manifest-path "$MIRIDIR"/Cargo.toml -- $MIRIFLAGS "$@"
+    fi
     ;;
 fmt)
     find "$MIRIDIR" -not \( -name target -prune \) -name '*.rs' \
--- a/tests/compiletest.rs
+++ b/tests/compiletest.rs
@@ -242,7 +242,7 @@ fn main() -> Result<()> {
     let target = get_target();
 
     if let Some(first) = std::env::args().nth(1) {
-        if first == "miri-run-dep-mode" {
+        if first == "--miri-run-dep-mode" {
             return run_dep_mode(target);
         }
     }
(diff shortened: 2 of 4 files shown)


tests/compiletest.rs
```diff
@@ -45,7 +46,13 @@ fn build_so_for_c_ffi_tests() -> PathBuf {
     so_file_path
 }
 
-fn run_tests(mode: Mode, path: &str, target: &str, with_dependencies: bool) -> Result<()> {
+fn run_test_config(
```
RalfJung, line 49: This doesn't run the config, right? So `run` shouldn't be in the fn name IMO.

tests/compiletest.rs
```diff
@@ -226,8 +238,15 @@ fn get_target() -> String {
 
 fn main() -> Result<()> {
     ui_test::color_eyre::install()?;
+
     let target = get_target();
 
+    if let Some(first) = std::env::args().nth(1) {
+        if first == "miri-run-dep-mode" {
```
RalfJung, line 245: Should we make this a flag, `--miri-run-dep-mode`? We already don't allow filters starting with `-` before `--` anyway so this avoids any potential conflicts.

tests/compiletest.rs
```diff
@@ -250,6 +269,21 @@ fn main() -> Result<()> {
     Ok(())
 }
 
+fn run_dep_mode(target: String) -> Result<()> {
+    let files = std::env::args().skip_while(|arg| arg != "--").skip(1);
```
RalfJung, line 273: What's going on with the arguments here? Why skip to after the `--`?

tests/compiletest.rs
```diff
@@ -250,6 +269,21 @@ fn main() -> Result<()> {
...
+    for path in files {
+        let mut config = run_test_config(std::iter::empty(), &target, &path, Mode::Yolo, true);
+        config.program.args.remove(0); // remove the `--error-format=json` argument
+        config.program.args.push("--color".into());
+        config.program.args.push("always".into());
+        let output = ui_test::run_file(config, Path::new(&path))?;
+        std::io::stderr().write_all(&output.stderr)?;
+        std::io::stdout().write_all(&output.stdout)?;
```
RalfJung, line 281: Is there any way we can forward the output live, rather than buffer and re-emit? In particular this will print the wrong thing when the program prints into both stderr and stdout in an interleaved way.

tests/compiletest.rs
oli-obk: Yea.... "run test"-config vs run "test config"

tests/compiletest.rs
oli-obk: There is no support for that yet, and I'm not sure we want it? Fwiw: stderr and stdout aren't really reliably interleaved on hosts either

RalfJung: > stderr and stdout aren't really reliably interleaved on hosts either

I have never seen that not work. On a typical Linux terminal they are both sending things to the same output and they are properly interleaved. It is AFAIK very common for programs to rely on this.

miri
```diff
@@ -323,11 +323,17 @@ run)
...
     fi
+
     # First build and get a sysroot.
     $CARGO build $CARGO_EXTRA_FLAGS --manifest-path "$MIRIDIR"/Cargo.toml
     find_sysroot
     # Then run the actual command.
-    exec $CARGO run $CARGO_EXTRA_FLAGS --manifest-path "$MIRIDIR"/Cargo.toml -- $MIRIFLAGS "$@"
+    
+    if [ "$COMMAND" = "run-dep" ]; then
+        exec $CARGO test --test compiletest -- miri-run-dep-mode $CARGO_EXTRA_FLAGS --manifest-path "$MIRIDIR"/Cargo.toml -- $MIRIFLAGS "$@"
```
RalfJung, line 333: The `$CARGO_EXTRA_FLAGS --manifest-path "$MIRIDIR"/Cargo.toml` part here is odd. This should be parameters for `cargo test`, right? The goal is to make this command work even if your working directory is elsewhere.

tests/compiletest.rs
oli-obk: I'll have to figure out how to do that, so this won't get fixed in this PR

bjorn3: That is because terminals use a single pipe for stdout and stderr and thus are able to preserve relative ordering but can't distinguish between stdout and stderr. You could make a pipe and then pass it to `.stdout()` and `.stderr()` on the `Command` to have the same correct interleaving. `.stdout()` and `.stderr()` accept an `impl Into<Stdio>` which `File` and `OwnedFd` implement.

oli-obk: This will get addressed with https://github.com/oli-obk/ui_test/pull/68

tests/compiletest.rs
```diff
@@ -250,6 +269,21 @@ fn main() -> Result<()> {
     Ok(())
 }
 
+fn run_dep_mode(target: String) -> Result<()> {
+    let files = std::env::args().skip(2);
+    for path in files {
```
RalfJung, line 274: I would expect `./miri run-dep file.rs foo bar` to be interpreted as running `file.rs` with arguments `foo bar`. That's what `./miri run` does. This seems to interpret it as running `foo` and `bar` as well?

oli-obk: 😆 whoops.

tests/compiletest.rs
oli-obk: It wasn't possible to pass arguments without it anyway, so I fixed both this and argument passing in one go.

tests/compiletest.rs
```diff
@@ -250,6 +266,19 @@ fn main() -> Result<()> {
...
 }
 
+fn run_dep_mode(target: String, mut args: impl Iterator<Item = OsString>) -> Result<()> {
+    let path = args.next().expect("./miri run-dep must be followed by a file name");
+    let mut config = test_config(&target, "", Mode::Yolo, true);
+    config.program.args.remove(0); // remove the `--error-format=json` argument
+    config.program.args.push("--color".into());
+    config.program.args.push("always".into());
+    let mut cmd = ui_test::test_command(config, Path::new(&path))?;
+    cmd.arg("--");
```
RalfJung, line 276: Who is this `--` for? Separating `cargo run` arguments from arguments passed to the program?

tests/compiletest.rs
```diff
@@ -250,6 +266,19 @@ fn main() -> Result<()> {
     Ok(())
 }
 
+fn run_dep_mode(target: String, mut args: impl Iterator<Item = OsString>) -> Result<()> {
+    let path = args.next().expect("./miri run-dep must be followed by a file name");
+    let mut config = test_config(&target, "", Mode::Yolo, true);
```
RalfJung, line 271:
```suggestion
    let mut config = test_config(&target, "", Mode::Yolo, /* with_dependencies */ true);
```

RalfJung:
r=me with comments resolving the last 2 nits

Issue #2734, opened 2022-12-20
simd_bitmask ICEs when result type is array


matthiaskrgr:
from `src/test/ui/simd/simd-bitmask.rs`
````rust
//run-pass
//ignore-endian-big behavior of simd_select_bitmask is endian-specific
#![feature(repr_simd, platform_intrinsics)]

extern "platform-intrinsic" {
    fn simd_bitmask<T, U>(v: T) -> U;
}

#[derive(Copy, Clone)]
#[repr(simd)]
struct Simd<T, const N: usize>([T; N]);

fn main() {
    unsafe {
        let v = Simd::<i8, 4>([-1, 0, -1, 0]);
        let _a: [u8; 1] = simd_bitmask(v);
    }

}

````

````
thread 'rustc' panicked at 'assertion failed: dest.layout.ty.is_integral()', src/tools/miri/src/shims/intrinsics/simd.rs:550:17
stack backtrace:
   0:     0x7efe47365a7a - std::backtrace_rs::backtrace::libunwind::trace::h06c3ddb908575635
                               at /rustc/935dc07218b4bf6e20231e44eb9263b612fd649b/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:     0x7efe47365a7a - std::backtrace_rs::backtrace::trace_unsynchronized::h8bfc3753e1cc3345
                               at /rustc/935dc07218b4bf6e20231e44eb9263b612fd649b/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x7efe47365a7a - std::sys_common::backtrace::_print_fmt::h3c304feda98d6679
                               at /rustc/935dc07218b4bf6e20231e44eb9263b612fd649b/library/std/src/sys_common/backtrace.rs:65:5
   3:     0x7efe47365a7a - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h98af9289a7a29dc2
                               at /rustc/935dc07218b4bf6e20231e44eb9263b612fd649b/library/std/src/sys_common/backtrace.rs:44:22
   4:     0x7efe473c87ae - core::fmt::write::h02538848739b7e2a
                               at /rustc/935dc07218b4bf6e20231e44eb9263b612fd649b/library/core/src/fmt/mod.rs:1208:17
   5:     0x7efe47355f45 - std::io::Write::write_fmt::h4aac2f2c3f0128af
[...]

PR #2895, opened 2023-05-13
support array return types in simd_bitmask

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/2734

As usual I am stomped by the simd_bitmask behavior wrt endianess. I confirmed that for little endian, Miri matches what rustc currently does, but I can't test rustc on big endian. @workingjubilee @calebzulawski could you have a look whether those new tests make sense?


calebzulawski:
This looks correct--the endianness only affects each separate element and not the array order (in std::simd, we reverse each array element on big endian)

RalfJung:
Ah, I didn't even realize that that is what happens -- thanks!

@bors r+

PR #2899, opened 2023-05-19
Try to ensure measureme output is valid when the interpreter is interrupted

saethlin:
Dropping our `measureme::Profiler` will have the profiler flush its output buffer and write out its string table. If the profiler's `Drop` impl does not run, the output file is not usable.

This approach has zero runtime overhead.

### DIFF
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -3,6 +3,7 @@
 use std::cell::RefCell;
 use std::collections::hash_map::Entry;
 use std::num::TryFromIntError;
+use std::sync::atomic::{AtomicBool, Ordering::Relaxed};
 use std::task::Poll;
 use std::time::{Duration, SystemTime};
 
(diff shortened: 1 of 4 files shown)


src/concurrency/thread.rs
```diff
@@ -1012,8 +1013,24 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
     /// Run the core interpreter loop. Returns only when an interrupt occurs (an error or program
     /// termination).
     fn run_threads(&mut self) -> InterpResult<'tcx, !> {
+        static SIGNALED: AtomicBool = AtomicBool::new(false);
+        ctrlc::set_handler(move || {
+            // Indicate that we have ben signaled to stop. If we were already signaled, exit
+            // immediately. In our interpreter loop we try to consult this value often, but if for
+            // whatever reason we don't get to that check or the cleanup we do upon finding that
+            // this bool has become true takes a long time, the exit here will promptly exit the
+            // process.
```
RalfJung, line 1022:
```suggestion
            // process on the 2nd Ctrl-C.
```
That's assuming I understood the code correctly

src/concurrency/thread.rs
```diff
@@ -1012,8 +1013,24 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+            if SIGNALED.swap(true, Relaxed) {
+                std::process::exit(1);
+            }
+        })
+        .unwrap();
         let this = self.eval_context_mut();
         loop {
+            if SIGNALED.load(Relaxed) {
+                this.machine.handle_abnormal_termination();
+                std::process::exit(1);
```
RalfJung, line 1032: Is it worth using a different exit code here? No idea what the convention would be.

saethlin: I also do not know what the convention would be for exit codes. At a guess, the signal number would be used, but the ctrlc crate hides that from us and moreover I have no idea what you'd do on Windows.

src/machine.rs
```diff
@@ -713,6 +713,11 @@ impl<'mir, 'tcx> MiriMachine<'mir, 'tcx> {
         let def_id = frame.instance.def_id();
         def_id.is_local() || self.local_crates.contains(&def_id.krate)
     }
+
+    /// Called when the interpreter is going to shut down abnormally, such as due to a ctrl+c.
+    pub(crate) fn handle_abnormal_termination(&mut self) {
+        drop(self.profiler.take());
```
RalfJung, line 719: Would be good to have a comment explaining why this is important.

src/concurrency/thread.rs
RalfJung: Fair. Fine to go with 1 for now then.

src/concurrency/thread.rs
saethlin: Yup.

RalfJung:
r=me after squashing

PR #2901, opened 2023-05-23
Hide full miri command line in `./miri run-dep`

### LINKED ISSUE #2443: Provide a way to run a test that needs an external dependency

oli-obk:
fixes #2443


RalfJung:
@bors r+
Thanks!

PR #2902, opened 2023-05-26
rustup

saethlin:
I don't know what happened here: https://github.com/rust-lang/rust/commit/efb9c30ed2bef28812f47914dea3ae4f703e6870#r115119947 that commit broke our tests I think. I don't know why it didn't break on the rustc side.


RalfJung:
In rustc our tests run with "-A unused" so that it breaks less often.

I guess now with the subtree we don't need that any more.

PR #2908, opened 2023-05-31
CI: test ./miri bench

RalfJung:


RalfJung:
Also I'd expect this to break on Windows since we don't have https://github.com/rust-lang/miri/pull/2877 yet.

RalfJung:
Yes, there we go.

@bors r=oli-obk

PR #2909, opened 2023-05-31
Rewrite miri script in rust

### LINKED ISSUE #2883: Rewrite miri script in Rust
Per https://github.com/rust-lang/miri/pull/2381#pullrequestreview-1045199136 and https://github.com/rust-lang/miri/pull/2877, a [miri script](https://github.com/rust-lang/miri/commit/798e6e13b6122c85599790a49f076221aad8849b) written in Rust would be more portable.

osiewicz:
This is a sketch of a rewrite of miri script in Rust. It does not include changes made in https://github.com/rust-lang/miri/pull/2908 yet. Environment variables are not properly propagated yet, which is something I plan to address.

This PR is mostly a heads-up about the ongoing effort and it's state.
It's definitely not the cleanest code I've seen in my life, but my first goal was feature/interface parity. I will iterate on it a bit before marking it as ready.

I wonder though how this should be integrated/tested. Are you aware of anyone using `./miri` in their scripts?
I guess we should keep existing `miri` script in place and let it run miri-script package directly?

CI should probably `cargo check` this package as well.

Fixes #2883


RalfJung:
Thanks for the PR! Someone will take a look and review this eventually, we're just all busy currently -- sorry for the delay.

osiewicz:
Hey, 
No problem - that's totally understandable. I intend to polish this before marking this PR as ready for review. Sadly I couldn't find time for that (or anything, really) recently.

Miri script works as is, we don't need to rush this in.

saethlin:
> I wonder though how this should be integrated/tested. 

FWIW I'm not particularly worried about testing this code. Obviously it _would be nice_ if it had good tests, but a lot of this is mucking about with the state of "the system", and the parts of this that really matter will get exercised regularly by us. Write tests if they help you build confidence that this works. But I think it would be easy to go overboard and spend more effort on the test suite than the tool.

> Are you aware of anyone using `./miri` in their scripts?

I don't think this is particularly worth worrying about.

> I guess we should keep existing `miri` script in place and let it run miri-script package directly?

Yes. `./miri` is much shorter than `cargo run --manifest-path=miri-script/Cargo.toml -- `. It would be neat if the script were turned into the thinnest wrapper around that. User workflows could be unaffected, the CI setup could stay as-is, and so on.

miri-script/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+authors = ["Miri Team"]
+description = "Helpers for miri maintenance"
+license = "MIT OR Apache-2.0"
+name = "miri"
```
saethlin, line 5: If someone `cargo install`s this, unless they have a creative setup, this will stomp over the rustup proxy `miri` executable and break their install. The bin target here should be called something else, or otherwise prevent that situation.

miri-script/src/arg.rs
```diff
@@ -0,0 +1,112 @@
...
+#[derive(Subcommand, Clone, Debug)]
+pub enum Subcommands {
+    /// Installs the miri driver and cargo-miri.
+    /// Sets up the rpath such that the installed binary should work in any
+    /// working directory. Note that the binaries are placed in the `miri` toolchain
+    /// sysroot, to prevent conflicts with other toolchains.
+    Install {
+        /// Flags that are passed through to `cargo install`.
+        #[arg(trailing_var_arg = true)]
+        flags: Vec<String>,
```
saethlin, line 19: Is `String` the right type to use for a passed-through CLI argument? If the arguments are literally just passed through, I think `OsString` is better.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+use path_macro::path;
+use walkdir::WalkDir;
+
+use crate::arg::Subcommands;
+
+const JOSH_FILTER: &str =
+    ":rev(75dd959a3a40eb5b4574f8d2e23aa6efbeb33573:prefix=src/tools/miri):/src/tools/miri";
+
+fn detect_miri_dir() -> std::io::Result<PathBuf> {
+    const MIRI_SCRIPT_ROOT_DIR: &str = env!("CARGO_MANIFEST_DIR");
```
saethlin, line 17: Should this be `const`? I feel like we want this at runtime, no?

miri-script/Cargo.toml
osiewicz: Great point, I'll rename binary to `miri-script` then.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+        Ok(Self { libdir, sysroot })
+    }
+}
+
+impl<'a> MiriRunner<'a> {
+    pub(super) fn run(command: &'a super::Subcommands) -> Result<()> {
+        let miri_dir = detect_miri_dir()?;
+        let active_toolchain = get_current_toolchain(&miri_dir)?;
+        let config = command.get_auto_config(&miri_dir);
+        let cargo_extra_flags = std::env::var("CARGO_EXTRA_FLAGS").unwrap_or_default();
```
saethlin, line 79: I think this ends up passing an empty string to all the attempts to run `cargo`. At least it does on Linux, and so none of this actually works for me :upside_down_face: I was so excited to try this out and I definitely will when you figure out a solution to this. I thought I could hack around it quickly and I can't.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+    }
+
+    fn toolchain(&self, flags: &[String]) -> Result<()> {
+        let new_commit =
+            std::fs::read_to_string(&path!(self.miri_dir / "rust-version"))?.trim().to_owned();
+        // Make sure rustup-toolchain-install-master is installed.
+        which::which("rustup-toolchain-install-master").expect("Please install rustup-toolchain-install-master by running 'cargo install rustup-toolchain-install-master'");
+        let current_commit = {
+            let rustc_info =
+                cmd("rustc", ["+miri", "--version", "-v", "2"]).dir(&self.miri_dir).read()?;
```
saethlin, line 239: What is this `"2"` for?

miri-script/src/commands.rs
osiewicz: Yeah, env variables are not propagated properly (yet). IIRC `apply_env_variables` can help with that.

miri-script/src/commands.rs
osiewicz: Yeah, it's not necessary; It's essentially a copy-paste error. :/

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+            std::process::exit(2);
+        }
+        ret.extra_vars = extra_vars;
+        // Share target dir between `miri` and `cargo-miri`.
+        let target_dir = std::env::var("CARGO_TARGET_DIR")
+            .ok()
+            .filter(|val| !val.is_empty())
+            .unwrap_or_else(|| {
+                let target_dir = path!(ret.miri_dir / "target");
+                target_dir.to_string_lossy().to_string()
```
saethlin, line 121: Same comment on all the use of `to_string_lossy` as I had on pass-through CLI arguments. Perhaps that means we also need `env::var_os`?

saethlin:
I'm done leaving comments. I think this is an _amazing_ start, and I'd love to try it out when the problem with unset `CARGO_EXTRA_FLAGS` is addressed.

miri-script/src/commands.rs
RalfJung: I don't think we support putting this binary anywhere else, so compile-time seems fine, no?

miri-script/src/commands.rs
RalfJung: CARGO_EXTRA_FLAGS should be split into individual arguments according to shell rules, so the type here should be `Vec<String/OsString>`.

miri-script/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+clap = {version = "4.2", features = ["derive", "env"]}
+duct = "0.13"
+which = "4.4"
+walkdir = "2.3"
+itertools = "0.10"
+path_macro = "1.0"
```
RalfJung, line 19: missing newline at end of file

miri-script/src/arg.rs
```diff
@@ -0,0 +1,112 @@
...
+        /// Flags that are passed through to `cargo test`.
+        #[arg(trailing_var_arg = true)]
+        flags: Vec<String>,
+    },
+    /// Build miri, set up a sysroot and then run the test suite.
+    Bless {
+        /// Flags that are passed through to `cargo test`.
+        #[arg(trailing_var_arg = true)]
+        flags: Vec<String>,
+    },
```
RalfJung, line 44: FWIW, what we really want is `./miri test --bless`; that just was too annoying to implement with the shell script.

Feel free to just leave a FIXME to do this later.

miri-script/src/arg.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// Build miri, set up a sysroot and then run the driver with the given <flags>.
+    /// (Also respects MIRIFLAGS environment variable.)
+    Run {
+        /// Flags that are passed through to `miri`
+        #[arg(trailing_var_arg = true)]
+        flags: Vec<String>,
+    },
+    /// Build miri, set up a sysroot and then run the driver with the given <flags>.
+    /// (Also respects MIRIFLAGS environment variable.)
+    RunDep {
```
RalfJung, line 54: This should probably also eventually become `./miri run --dep` or so.

miri-script/src/main.rs
```diff
@@ -0,0 +1,63 @@
...
+            | RunDep { .. }
+            | Bless { .. }
+            | Fmt { .. }
+            | Clippy { .. }
+            | Cargo { .. } => true,
+        }
+    }
+    fn get_auto_config(&self, miri_dir: &Path) -> Option<AutoConfig> {
+        use std::env::VarError;
+        if let Err(VarError::NotPresent) = std::env::var("MIRI_AUTO_OPS") {
```
RalfJung, line 42: Just use `std::env::var_os(...).is_none()`

miri-script/src/main.rs
```diff
@@ -0,0 +1,63 @@
...
+            std::env::set_var("MIRI_AUTO_OPS", "42");
```
RalfJung, line 43: This shouldn't be needed any more; the shell script did this because it recursively invoked itself to run the auto-ops, but we do no such nonsense here.

miri-script/src/main.rs
```diff
@@ -0,0 +1,63 @@
...
+use arg::Subcommands;
+
+struct AutoConfig {
+    toolchain: bool,
+    fmt: bool,
+    clippy: bool,
+}
+
+impl Subcommands {
+    fn supports_auto_things(&self) -> bool {
```
RalfJung, line 19: `run_auto_things` seems like a more apt name

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+    variables: &[(String, String)],
+) -> &'a mut duct::Expression {
+    for (key, value) in variables {
+        *cmd = cmd.env(key, value);
+    }
+    cmd
+}
+#[derive(Clone, Debug)]
+pub(super) struct MiriRunner<'a> {
+    /// miri_dir is based on build directory of miri-script.
```
RalfJung, line 37: This doesn't tell us what directory this is.
```suggestion
    /// miri_dir is the root of the miri repository checkout we are working in.
```

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+struct Extras {
+    sysroot: PathBuf,
+    libdir: PathBuf,
+}
+
+impl Extras {
+    fn new(toolchain: &str) -> Result<Self> {
+        // Determine toolchain properties.
+        let toolchain_prop = &format!("+{toolchain}");
+        let target_output = cmd("rustc", [&toolchain_prop, "--version", "--verbose"]).read()?;
```
RalfJung, line 59: This seems like a good use for rustc_version

miri-script/src/main.rs
```diff
@@ -0,0 +1,63 @@
...
+            | Cargo { .. } => true,
+        }
+    }
+    fn get_auto_config(&self, miri_dir: &Path) -> Option<AutoConfig> {
+        use std::env::VarError;
+        if let Err(VarError::NotPresent) = std::env::var("MIRI_AUTO_OPS") {
+            std::env::set_var("MIRI_AUTO_OPS", "42");
+            if !self.supports_auto_things() {
+                return None;
+            }
```
RalfJung, line 46: IMO that logic would make more sense in `MiriRuner::run`

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+                ret.fmt(&[])?;
+            }
+            if config.clippy {
+                ret.clippy(&["--".to_owned(), "-D".to_owned(), "warnings".to_owned()])?;
+            }
+            if config.toolchain {
+                // Let's make sure to actually use that toolchain, too.
+                // We do it after running all the auto things though to use a detected toolchain.
+                ret.active_toolchain = "miri".to_owned();
+            }
```
RalfJung, line 105: This should be inside `if config.toolchain` to match the shell script behavior (in particular it must run before fmt/clippy)

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+            .filter(|val| !val.is_empty())
+            .unwrap_or_else(|| "2".to_owned());
+        ret.extra_flags.insert("CARGO_PROFILE_DEV_OPT_LEVEL".to_owned(), devel_opt_level);
+
+        let rustflags = std::env::var("RUSTFLAGS").unwrap_or_default();
+        let rustflags = format!(
+            "-Zunstable-options -Wrustc::internal -Wrust_2018_idioms -Wunused_lifetimes -Wsemicolon_in_expressions_from_macros {rustflags}"
+        );
+        let rustflags =
+            format!("-C link-args=-Wl,-rpath,{} {rustflags}", ret.extra_vars.libdir.display());
```
RalfJung, line 137: Please port over the comments from the shell script as well (also applies anywhere else if you dropped any comments)

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+
+        if !current_sysroot.is_empty() {
+            // Sysroot already set, use that.
+            let current_value =
+                self.extra_flags.insert("MIRI_SYSROOT".to_string(), current_sysroot.clone());
+            assert!(current_value.is_none() || current_value.unwrap() == current_sysroot);
+            return;
+        }
+        let miri_test_target = std::env::var("MIRI_TEST_TARGET").unwrap_or_default();
+        let target = (!miri_test_target.is_empty()).then_some(miri_test_target.as_str());
```
RalfJung, line 171: This looks like a strange way to go from `Option<_>` to `String` and then back to `Option<_>`. Please just stay inside the option.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+            Subcommands::Toolchain { flags } => self.toolchain(&flags),
+            Subcommands::RustcPull { commit } => self.rustc_pull(commit.clone()),
+            Subcommands::RustcPush { github_user, branch } =>
+                self.rustc_push(&github_user, &branch),
+        }
+    }
+    fn find_sysroot(&mut self) {
+        let current_sysroot = std::env::var("MIRI_SYSROOT").unwrap_or_default();
+
+        if !current_sysroot.is_empty() {
```
RalfJung, line 163: You are using `unwrap_or_default` a lot; this needs careful scrutiny.  Generally we should not just drop the `Option` like that.
For example here it should be
```suggestion
        let current_sysroot = std::env::var("MIRI_SYSROOT");

        if let Some(current_sysroot) = current_sysroot {
```

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+        }
+        let miri_test_target = std::env::var("MIRI_TEST_TARGET").unwrap_or_default();
+        let target = (!miri_test_target.is_empty()).then_some(miri_test_target.as_str());
+        let sysroot = self.build_sysroot(target);
+        self.extra_flags.insert("MIRI_SYSROOT".into(), sysroot);
+    }
+
+    fn build_sysroot(&self, target: Option<&str>) -> String {
+        let manifest_path = path!(self.miri_dir / "cargo-miri" / "Cargo.toml");
+        let manifest_path = manifest_path.to_string_lossy().to_string();
```
RalfJung, line 178:
```suggestion
        let manifest_path = manifest_path.to_string_lossy().into_owned();
```
applies to all `Cow` (`to_string` goes through `Display` so this is extremely generic and not usually what should be used IMO)

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+            &toolchain,
+            &self.cargo_extra_flags,
+            false,
+            &args,
+        )?;
+        Ok(())
+    }
+
+    /// Returns a toolchain override for use in command line (`+{toolchain_name}`).
+    fn get_toolchain_override(&self) -> String {
```
RalfJung, line 545: Please don't have this in the middle of the subcommand functions; move it to the top

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+            .run()?;
+            Ok(())
+        }
+        // Check, and let caller control flags.
+        let toolchain = self.get_toolchain_override();
+        let miri_manifest = path!(self.miri_dir / "Cargo.toml");
+        let cargo_miri_manifest = path!(self.miri_dir / "cargo-miri" / "Cargo.toml");
+        let args = flags.iter().map(String::as_str).collect::<Vec<_>>();
+        check_package(
+            &miri_manifest.to_string_lossy().to_owned(),
```
RalfJung, line 528:
```suggestion
            &miri_manifest.to_string_lossy().into_owned(),
```
also applies for other `Cow`s

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+    }
+
+    /// Returns a toolchain override for use in command line (`+{toolchain_name}`).
+    fn get_toolchain_override(&self) -> String {
+        format!("+{}", &self.active_toolchain)
+    }
+    fn test(&mut self, flags: &[String]) -> Result<()> {
+        let bless = matches!(self.command, Subcommands::Bless { .. });
+        let miri_manifest = path!(self.miri_dir / "Cargo.toml");
+        let miri_manifest = miri_manifest.to_string_lossy().to_owned();
```
RalfJung, line 551: As @saethlin noted, ultimately we are doing to pass this to a command-running thing so we shouldn't need this.

The standard `Command` works with all `AsRef<OsStr>`, so a `PathBuf` like this could just be used without any further conversion. Is it possible to achieve the same here?

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+        let cargo_miri_manifest = path!(self.miri_dir / "cargo-miri" / "Cargo.toml");
+        let args = flags.iter().map(String::as_str).collect::<Vec<_>>();
+        check_package(
+            &miri_manifest.to_string_lossy().to_owned(),
+            &toolchain,
+            &self.cargo_extra_flags,
+            true,
+            &args,
+        )?;
+        check_package(
```
RalfJung, line 534: I think it could be useful to have a "run cargo comment "helper that also takes care of the toolchain and the extra flags. That could help make this all less verbose.

```rust
self.cargo_command(
  "check",
  path!(self.miri_dir / "cargo-miri" / "Cargo.toml"),
  args
);
```

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+        )
+        .run()?;
+        Ok(())
+    }
+
+    fn run_(&mut self, flags: &[String]) -> Result<()> {
+        use itertools::Itertools;
+        let target = flags.iter().tuple_windows().find(|(first, _)| first.as_str() == "--target");
+        if let Some(target) = target {
+            self.extra_flags.insert("MIRI_TEST_TARGET".to_owned(), target.1.clone());
```
RalfJung, line 577:
```suggestion
        if let Some((_, target)) = target {
            self.extra_flags.insert("MIRI_TEST_TARGET".to_owned(), target.clone());
```
Probably we want a `self.set_env` that takes `AsRef<OsStr>` to have less conversion noise everywhere

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+#[derive(Clone, Debug)]
+pub(super) struct MiriRunner<'a> {
+    /// miri_dir is based on build directory of miri-script.
+    miri_dir: PathBuf,
+    /// active_toolchain is passed as `+toolchain` argument to cargo/rustc invocations.
+    active_toolchain: String,
+    cargo_extra_flags: String,
+    command: &'a super::Subcommands,
+    /// Environment variables passed to child processes.
+    extra_flags: BTreeMap<String, String>,
```
RalfJung, line 44: These aren't flags, they are env vars, as the comment says. So please rename to `env` or so.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+        // First build and get a sysroot.
+        let miri_manifest = path!(self.miri_dir / "Cargo.toml");
+        let miri_manifest = miri_manifest.to_string_lossy();
+        let toolchain = self.get_toolchain_override();
+        Self::build_package(&miri_manifest, &toolchain, &self.cargo_extra_flags, &[])?;
+        self.find_sysroot();
+        // Then run the actual command.
+        let miri_flags = self.extra_flags.get("MIRIFLAGS").cloned();
+        let miri_flags = miri_flags.as_deref();
+        let args = flags.iter().map(String::as_str).collect::<Vec<_>>();
```
RalfJung, line 596: Does `chain!` really force us to use `Vec<&str>`? Would be much more ergonomic if it worked more like `Command::args` where we can give any iterator with element type `AsRef<OsStr>`, then we would have less conversion noise here

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+        Ok(())
+    }
+    fn many_seeds(&self, command: &[String], seed_start: u64, seed_count: u64) -> Result<()> {
+        let seed_end = seed_start + seed_count;
+        assert!(!command.is_empty());
+        for seed in seed_start..seed_end {
+            println!("Trying seed: {seed}");
+            let current_miriflags = std::env::var("MIRIFLAGS").unwrap_or_default();
+            let miriflags = format!("{current_miriflags} -Zlayout-seed={seed} -Zmiri-seed={seed}");
+            let args = if command.len() == 1 { &[] } else { &command[1..] };
```
RalfJung, line 701: `command.split_first()` would be useful here

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+        } else {
+            benches.to_owned()
+        };
+        for bench in benches {
+            let current_bench_dir = path!(benches_dir / bench / "Cargo.toml");
+            let current_bench_dir = current_bench_dir.to_string_lossy().into_owned();
+            let toolchain = self.get_toolchain_override();
+            let command =
+                format!("cargo {} miri run --manifest-path {current_bench_dir}", toolchain);
+            cmd("hyperfine", ["-w", "1", "-m", "5", "--shell=none", &command]).run()?;
```
RalfJung, line 227: Note that this changed in the shell script recently, so the rust version needs updating

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+            let commit_hash_line = rustc_info
+                .lines()
+                .find(|line| line.starts_with("commit-hash: "))
+                .expect("rustc output does not contain commit-hash line");
+            commit_hash_line
+                .split_whitespace()
+                .nth(1)
+                .expect("rustc commit-hash line is malformed")
+                .trim()
+                .to_owned()
```
RalfJung, line 249: I think `rustc_version` can replace this ad-hoc parser

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+        // Fetch given rustc commit and note down which one that was
+        let fetch_url =
+            format!("http://localhost:8000/rust-lang/rust.git@{commit}{JOSH_FILTER}.git");
+        cmd("git", ["fetch", &fetch_url])
+            .run()
+            .expect("FAILED to fetch new commits, something went wrong");
+        const MERGE_COMMIT_MESSAGE: &str = "Merge from rustc";
+        cmd("git", ["merge", "FETCH_HEAD", "--no-ff", "-m", MERGE_COMMIT_MESSAGE])
+            .run()
+            .expect("FAILED to merge new commits, something went wrong");
```
RalfJung, line 323: This is using `expect` a lot for things that can totally actually happen, e.g. a merge conflict. Please use proper error reporting instead of panics. Probably this should be using anyhow and then we can write
```suggestion
            .context("FAILED to merge new commits, something went wrong")?;
```

miri-script/src/commands.rs
saethlin: This program looks like our version of `x`. So I think a user would be justified in assuming that they should `cargo install` it then be able to cd around or run this in multiple clones or a worktree just fine.

But it seems like it is not our version of `x`, so it would be nice to either prevent that entirely (can we make `cargo install` an error? I don't think so?) or if it's not that hard, handle it.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,711 @@
...
+        let sysroot = cmd("rustc", [&toolchain_prop, "--print", "sysroot"]).read()?;
+
+        let sysroot = PathBuf::from(sysroot);
+        let libdir = path!(sysroot / "lib" / "rustlib" / target / "lib");
+        Ok(Self { libdir, sysroot })
+    }
+}
+
+impl<'a> MiriRunner<'a> {
+    pub(super) fn run(command: &'a super::Subcommands) -> Result<()> {
```
RalfJung, line 75: If you rename this to `exec`, we don't need the `run_` hack :)

RalfJung (commented):
This is a great start indeed, thanks a lot. :) I left a bunch of notes from a first pass. My main concern is verbosity: this is a lot longer than the shell script it is porting. I think this can be reduced be avoiding all the conversion from paths to strings and whatnot -- no such conversion is needed when working with `Command` directly, I hope we can also make that work here.

Also command.rs could use some splitting up into multiple files: all the cargo-wrapping stuff, the pull/push stuff, and the other stuff (fmt, bench, many-seeds, toolchain)

As usual please don't force-push this branch, but add new commits, so that I can see more easily what changed compared to last time.

miri-script/src/commands.rs
RalfJung: The entrypoint is still `./miri`, right? So users shouldn't ever even see this crate unless they go look for it. We can have a comment in the Cargo.toml saying that this is not meant to be installed.

saethlin: I don't know what the entrypoint is meant to be :p

So long as we have a comment or something about this situation, I'm willing to wait if/until someone files an issue because they tried to `cargo install` this crate.

RalfJung: > I don't know what the entrypoint is meant to be :p

Well we certainly don't want `cargo run --manifest-path=...`.

OTOH, how is that `./miri` wrapper supposed to be able to know where the manifest is? That needs the readdir logic and an issue with exactly that logic is what triggered the rewrite in the first place... maybe `$(dirname "$0")/miri-script/Cargo.toml` is good enough (but then I wonder why we had that normalization logic in there in the first place :shrug: )

miri-script/src/arg.rs
osiewicz: Nah, it's fine - I've thought that `--bless` would be better than having a separate command for it, but I wanted to stay as close to script as possible. If we're fine with such "breaking" changes, that's fine with me.

miri-script/src/commands.rs
osiewicz: I assume you have [this crate](https://crates.io/crates/rustc_version) in mind, right?

miri-script/src/commands.rs
osiewicz: I agree `unwrap_or_default` is a bit overused, but once again, this was a matter of parity with current Miri script. In some places it checks for variable's presence whereas in others it also checks whether it is not an empty string too.

miri-script/src/commands.rs
```diff
@@ -227,8 +225,17 @@ impl MiriRunner<'_> {
         .run()?;
         Ok(())
     }
+}
+
+impl MiriRunner<'_> {
     fn bench(&self, benches: &[String]) -> Result<()> {
         // Make sure we have an up-to-date Miri installed
+        let hyperfine = std::env::var("HYPERFINE");
+        let hyperfine = hyperfine.as_deref().unwrap_or("hyperfine -w 1 -m 5 --shell=none");
```
saethlin, line 247: This is now verbose in a different direction but it would be _very cool_ if we could write shell commands like this in just one string literal.

Even cooler if we could stop shelling out all the time, but you take what you can get.

miri-script/src/commands.rs
osiewicz: Uhh, I'm not sure; correct me if I'm wrong, but [TOOLCHAIN variable is not exported](https://github.com/rust-lang/miri/blob/master/miri#L212), so it shouldn't affect child shells. 

On the other hand, `toolchain` overrides toolchain in current directory, so it might be the case that fmt and clippy use `miri` toolchain indirectly anyways. 

Okay, let's just change this for now.

miri-script/src/commands.rs
saethlin: (you reacted with confused so I'm going to try to clarify)

I find the vertical arrangement of these short CLI arguments to be a bit hard to read. The one for rustup-toolchain-install-master is a bunch of very short arguments each on their own line, and between all the multiple lines and all the string literal delimiters that's just not what it looks like in my terminal when I run rustup-toolchain-install-master. It would be very neat if we could have Rust code invoking CLI tools in a way that looks similar to when I run them by hand. I didn't even think of what you're doing here as a way to make that happen. It's interesting.

What would be _really_ nice is if we could do all of the tasks this program needs to do without spawning subprocesses, but that's probably just a dream, and just migrating this to Rust is a huge benefit in itself.

osiewicz: Hey, sorry for the confusion; I didn't actually notice that it's a confused emoji - I thought it's a sad one, as I also find it a bit too verbose for my liking. I appreciate that you've reacted to it though =)
Emojis aside, it kinda sucks that we need to worry about knowing which part of `hyperfine` we should interpret as a command and which of them are actually parameters.

miri-script/src/commands.rs
osiewicz: Noted, I've kind of addressed that in follow-up commits. The issue I am facing is that when I'm passing args to `cmd`, some of them are string literals, whereas others are `OsStr`s already. This forces me to convert some arguments and make all of this unnecessarily noisy. I wish there was a way to e.g. accept a tuple of types that all implement `AsRef<OsStr` in chain :/

miri-script/src/commands.rs
RalfJung: Yes.

miri-script/src/commands.rs
RalfJung: Yeah that's because it's really hard to do otherwise in a shell script.

miri-script/src/commands.rs
RalfJung: > [TOOLCHAIN variable is not exported](https://github.com/rust-lang/miri/blob/master/miri#L212), so it shouldn't affect child shells.

Ah, fair.

> On the other hand, toolchain overrides toolchain in current directory, so it might be the case that fmt and clippy use miri toolchain indirectly anyways.

Indeed that will be the case.

miri-script/src/commands.rs
RalfJung: Yeah I was hoping `duct` would provide that...

Can't we have our own `chain_cmd` function with a signature like this?
```rust
fn chain_cmd(a: impl IntoIterator<Item = impl AsRef<OsStr>>, b: impl IntoIterator<Item = impl AsRef<OsStr>>) -> impl Iterator<Item=&OsStr>
```

miri-script/src/commands.rs
```diff
@@ -19,6 +20,7 @@ fn detect_miri_dir() -> std::io::Result<PathBuf> {
     Ok(PathBuf::from(MIRI_SCRIPT_ROOT_DIR).canonicalize()?.parent().unwrap().into())
 }
 
+/// Needed for `./miri bench`.
 fn get_current_toolchain(miri_dir: &Path) -> std::io::Result<String> {
```
RalfJung: get_current_toolchain isn't just used in `./miri bench`, something went wrong with comment porting here

miri-script/src/main.rs
```diff
@@ -38,21 +38,15 @@ impl Subcommands {
         }
     }
     fn get_auto_config(&self, miri_dir: &Path) -> Option<AutoConfig> {
-        use std::env::VarError;
-        if let Err(VarError::NotPresent) = std::env::var("MIRI_AUTO_OPS") {
```
RalfJung: Removing this entirely is not what I said. We do need to check if it is set, and skip the auto-ops if so.

We just don't have to set it ourselves.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,757 @@
...
+}
+
+impl MiriRunner<'_> {
+    fn bench(&self, benches: &[OsString]) -> Result<()> {
+        // Make sure we have an up-to-date Miri installed
+        let hyperfine = std::env::var("HYPERFINE");
+        let hyperfine = hyperfine.as_deref().unwrap_or("hyperfine -w 1 -m 5 --shell=none");
+        let hyperfine = shell_words::split(hyperfine).unwrap();
+        let Some((command, args)) = hyperfine.split_first() else {
+            todo!();
```
RalfJung, line 260: This should probably `bail!` (an anyhow macro)

miri-script/src/commands.rs
```diff
@@ -181,32 +190,32 @@ impl<'a> MiriRunner<'a> {
...
-            "--",
-            "miri",
-            "setup",
+            active_toolchain.as_os_str(),
+            &self.cargo_extra_flags.as_ref(),
+            "--manifest-path".as_ref(),
+            manifest_path.as_ref(),
+            "--".as_ref(),
+            "miri".as_ref(),
+            "setup".as_ref(),
```
RalfJung, line 213: What an eyesore. :/ We need to find a way to make this not look terrible (as this happens all over the codebase)... isn't duct supposed to be useful for exactly things like this?

miri-script/src/commands.rs
```diff
@@ -155,13 +155,15 @@ impl<'a> MiriRunner<'a> {
             Subcommands::Fmt { flags } => self.fmt(&flags),
             Subcommands::Clippy { flags } => self.clippy(&flags),
             Subcommands::Cargo { flags } => self.cargo(&flags),
-            Subcommands::ManySeeds { command, seed_start, seeds } =>
-                self.many_seeds(&command, *seed_start, *seeds),
+            Subcommands::ManySeeds { command, seed_start, seeds } => {
+                self.many_seeds(&command, *seed_start, *seeds)
+            }
```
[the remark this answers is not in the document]
osiewicz: Yup, I use multiple setups so it might be the case. I am sorry, I'll try to keep it consistent.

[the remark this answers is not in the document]
saethlin: Whatever you are using to format your checkout should be able to pull in Miri's rustfmt.toml. Sometimes editors have difficulty with Miri or rustc because they aren't just standard Cargo projects. Can you take a skim over the CONTRIBUTING.md and see if anything there looks helpful? If there isn't, can you maybe open an issue or start a topic on the Zulip so we can figure out what Miri can do to improve this?

miri-script/src/commands.rs
osiewicz: Yeah, we can probably use `duct` macro here, though it doesn't help with parts where we add some of the arguments conditionally.

miri-script/src/commands.rs
osiewicz: Yep; the original comment was for a TOOLCHAIN variable. I guess we could document `active_toolchain` member variable or scrap this altogether?

RalfJung: The original comment was for evaluation order: `TOOLCHAIN` needs to be set before the early block of commands, because of `bench`. The comment doesn't document the variable, it documes where the variable is computed.

With the Rust version, each command just queries the toolchain as it sees fit (right?), so this entire order issue disappears.

Having a doc comment that describes what this fn does would be useful though.

miri-script/src/commands.rs
osiewicz: Ugh. There are two ways to create a command with duct; wiith `cmd` [macro](https://docs.rs/duct/latest/duct/macro.cmd.html) and `cmd` [function](https://docs.rs/duct/latest/duct/fn.cmd.html). The former takes a fixed-size list of arguments that implement `Into<OsString>`, whereas the latter takes an iterable. Neither is ideal: the former takes literals/variable names - so if any of the options is appended conditionally, it won't work - whereas the latter is more flexible and more noisy at the same time.

I honestly think we might as well roll back to std processes.

osiewicz: https://github.com/matklad/xshell This lib might be worth a shot as well.

miri-script/src/commands.rs
osiewicz: Do we have any consensus here? I'd personally lean towards detecting Miri dir at compile time.

miri-script/src/commands.rs
osiewicz: I guess this is not valid anymore after a move to `xshell`?

miri-script/src/commands.rs
osiewicz: Chain required Item types of chained iterators to match, so there was a need for conversion. I guess it's redundant now?

miri-script/src/commands.rs
osiewicz: Seems like [xshell](https://github.com/matklad/xshell) is just what we wanted. :)

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+                target_dir.into()
+            });
+        ret.set_env("CARGO_TARGET_DIR", target_dir);
+
+        // We configure dev builds to not be unusably slow.
+        let devel_opt_level = std::env::var_os("CARGO_PROFILE_DEV_OPT_LEVEL")
+            .filter(|val| !val.is_empty())
+            .unwrap_or_else(|| "2".into());
+        ret.set_env("CARGO_PROFILE_DEV_OPT_LEVEL", devel_opt_level);
+        // Enable rustc-specific lints (ignored without `-Zunstable-options`).
```
RalfJung, line 144:
```suggestion
```

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+    /// Build a sysroot and set MIRI_SYSROOT to use it. Arguments are passed to `cargo miri setup`.
+    fn build_sysroot(&self, target: Option<&OsStr>) -> Result<String> {
+        let manifest_path = path!(self.miri_dir / "cargo-miri" / "Cargo.toml");
+        let Self { active_toolchain, cargo_extra_flags, .. } = &self;
+        let target_prefix: Option<&OsStr> = target.map(|_| "--target".as_ref());
+        let sh = self.shell_with_miri_env()?;
+        let output = cmd!(sh, "cargo +{active_toolchain} run {cargo_extra_flags...} --manifest-path {manifest_path} -- miri setup --print-sysroot {target_prefix...} {target...}").read();
+        if output.is_err() {
+            // Run it again so the user can see the error.
+            // This time without printing a sysroot though.
```
RalfJung, line 217:
```suggestion
            // Run it again (without `--print-sysroot`) so the user can see the error.
```

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+            let _ = cmd!(sh, "cargo +{active_toolchain} run {cargo_extra_flags...} --manifest-path {manifest_path} -- miri setup {target_prefix...} {target...}").run().with_context(|| "`cargo miri setup` failed")?;
+        }
+
+        Ok(output?)
+    }
+    /// Returns a toolchain override for use in command line (`+{toolchain_name}`).
+    fn get_toolchain_override(&self) -> OsString {
+        format!("+{}", &self.active_toolchain).into()
+    }
```
RalfJung, line 226: This doesn't seem worth it -- `build_sysroot` looks just fine without this helper?

If we want a helper function it would be for invoking `cargo +toolchain <subcommand>`, but it seems with xshell that isn't really needed any more.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+        toolchain: &OsStr,
+        extra_flags: &[String],
+        args: impl IntoIterator<Item = impl AsRef<OsStr>>,
+    ) -> Result<()> {
+        let sh = Shell::new()?;
+        cmd!(sh, "cargo {toolchain} build {extra_flags...} --manifest-path {path} {args...}")
+            .run()?;
+        Ok(())
+    }
+    fn shell_with_miri_env(&self) -> Result<Shell> {
```
RalfJung, line 239: Please move these helper fn up (currently they are called before they are defined)

Just `shell` is probably a good name; it is already a fn on `MiriRunner` so the "for_miri" part is redundant.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+        format!("+{}", &self.active_toolchain).into()
+    }
+    fn build_package(
+        // Path to Cargo.toml file of a package to build.
+        path: &OsStr,
+        toolchain: &OsStr,
+        extra_flags: &[String],
+        args: impl IntoIterator<Item = impl AsRef<OsStr>>,
+    ) -> Result<()> {
+        let sh = Shell::new()?;
```
RalfJung, line 234: Not the Miri shell?

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+    fn bench(&self, benches: &[OsString]) -> Result<()> {
+        // The hyperfine to use
+        let hyperfine = std::env::var("HYPERFINE");
+        let hyperfine = hyperfine.as_deref().unwrap_or("hyperfine -w 1 -m 5 --shell=none");
+        let hyperfine = shell_words::split(hyperfine).unwrap();
+        let Some((program_name, args)) = hyperfine.split_first() else {
+            bail!("Expected HYPERFINE environment variable to be non-empty");
+        };
+        // Make sure we have an up-to-date Miri installed
+        Self::exec_inner(&Subcommands::Install { flags: vec![] }, false)?;
```
RalfJung, line 259: As before I hope we can avoid this `exec_inner`

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+            benches.to_owned()
+        };
+        let sh = shell_with_parent_env()?;
+        let toolchain = &self.active_toolchain;
+        // Run the requested benchmarks
+        for bench in benches {
+            let current_bench_dir = path!(benches_dir / bench / "Cargo.toml");
+            cmd!(
+                sh,
+                "{program_name} {args...} 'cargo +'{toolchain}' miri run --manifest-path '{current_bench_dir}''"
```
RalfJung, line 279: What is going on with the `'` here?

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+        Ok(())
+    }
+
+    fn toolchain(&self, flags: &[OsString]) -> Result<()> {
+        // Make sure rustup-toolchain-install-master is installed.
+        which::which("rustup-toolchain-install-master").context("Please install rustup-toolchain-install-master by running 'cargo install rustup-toolchain-install-master'")?;
+        let sh = shell_with_parent_env()?;
+        sh.change_dir(&self.miri_dir);
+        let new_commit =
+            Some(sh.read_file(path!(self.miri_dir / "rust-version"))?.trim().to_owned());
```
RalfJung, line 292:
```suggestion
            Some(sh.read_file("rust-version")?.trim().to_owned());
```
why use `sh.read_file` if you don't make use of the relative working dir...?

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+            .context("FAILED to fetch new commits, something went wrong")?;
+        const MERGE_COMMIT_MESSAGE: &str = "Merge from rustc";
+        cmd!(sh, "git merge FETCH_HEAD --no-ff -m {MERGE_COMMIT_MESSAGE}")
+            .run()
+            .context("FAILED to merge new commits, something went wrong")?;
+        Ok(())
+    }
+
+    fn rustc_push(&self, github_user: &str, branch: &str) -> Result<()> {
+        let rustc_git = std::env::var_os("RUSTC_GIT").filter(|value| !value.is_empty());
```
RalfJung, line 358:
```suggestion
        let rustc_git = std::env::var_os("RUSTC_GIT");
```

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+        let working_directory = rustc_git.unwrap_or_else(|| {
+            // If rustc_git is `Some`, we'll use an existing fork for the branch updates.
+            // Otherwise, do this in the local Miri repo.
+            println!("This will pull a copy of the rust-lang/rust history into this Miri checkout, growing it by about 1GB.");
+            println!("To avoid that, abort now and set the RUSTC_GIT environment variable to an existing rustc checkout. Proceed? [y/N] ");
+            let mut answer = String::new();
+            let _ = std::io::stdin().read_line(&mut answer);
```
RalfJung, line 365: Why ignore errors here?

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+            // Path to Cargo.toml file of a package to check.
+            path: &OsStr,
+            toolchain: &OsStr,
+            extra_flags: &[String],
+            all_targets: bool,
+            args: impl IntoIterator<Item = impl AsRef<OsStr>>,
+        ) -> Result<()> {
+            let all_targets: Option<&OsStr> = all_targets.then_some("--all-targets".as_ref());
+            let sh = Shell::new()?;
+            cmd!(sh, "cargo {toolchain} check {extra_flags...} --manifest-path {path} {all_targets...} {args...}").run()?;
```
RalfJung, line 481:
```suggestion
            cmd!(sh, "cargo +{toolchain} check {extra_flags...} --manifest-path {path} {all_targets...} {args...}").run()?;
```
Missing `+`?

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+    }
+
+    fn test(&mut self, bless: bool, flags: &[OsString]) -> Result<()> {
+        let miri_manifest = path!(self.miri_dir / "Cargo.toml");
+        let toolchain = self.get_toolchain_override();
+        // First build and get a sysroot.
+        Self::build_package(
+            miri_manifest.as_ref(),
+            &toolchain,
+            &self.cargo_extra_flags,
```
RalfJung, line 506: If you make this a method (`self.build_package`), you don't have to pass `toolchain` and `cargo_extra_flags`.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+            std::iter::empty::<OsString>(),
+        )?;
+        self.find_sysroot()?;
+        if bless {
+            self.set_env("MIRI_BLESS", "Gesundheit");
```
RalfJung, line 511: Seems better to just use `sh.set_env` here, and keep this local to this `test` invocation.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+                break;
+            }
+        }
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use shell_words::split;
```
RalfJung, line 632: Why are you testing another crate's function here...?

RalfJung (commented):
xshell was a great choice, this looks a lot better. :)

miri-script/src/commands.rs
osiewicz: Hm, this does make use of relative working dir according to [xshell's doc](https://docs.rs/xshell/latest/xshell/).
Specifically:
`[Shell::read_file](https://docs.rs/xshell/latest/xshell/struct.Shell.html#method.read_file) works like [std::fs::read_to_string](https://doc.rust-lang.org/nightly/std/fs/fn.read_to_string.html), but paths are relative to the current directory of the [Shell](https://docs.rs/xshell/latest/xshell/struct.Shell.html).`

miri-script/src/commands.rs
osiewicz: I've left that in while looking for a way to split the shell words. I'll remove that, I am sorry. :(

miri-script/src/commands.rs
RalfJung: Yeah but you used `self.miri_dir / "rust-version"`, which is an absolute path, so the `sh` working dir doesn't matter. With my patch it is being used.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+    fn exec_inner(command: &super::Subcommands, run_auto_things: bool) -> Result<()> {
+        let miri_dir = detect_miri_dir()?;
+        let active_toolchain = get_active_toolchain(&miri_dir)?;
+        let config = command.get_auto_config(&miri_dir).filter(|_| run_auto_things);
+        // CARGO_EXTRA_FLAGS do not have to be a valid UTF-8, but that's what shell_words' expects.
+        let cargo_extra_flags = std::env::var("CARGO_EXTRA_FLAGS").unwrap_or_default();
+        let cargo_extra_flags = shell_words::split(&cargo_extra_flags)?;
+        let env = BTreeMap::new();
+
+        let mut ret = MiriRunner {
```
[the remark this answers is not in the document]
RalfJung: No need to be sorry. :)  This is shaping up great, so I started nitpicking! My comments are terse but please don't take that for grumpiness or anything like that.

miri-script/src/commands.rs
osiewicz: Oh, you're right. 😅

miri-script/src/commands.rs
osiewicz: Uh, I guess I did it out of laziness (not having to deal with type conversions to satisfy `unwrap_or_else` et al was nice). 
I've moved this snippet to `if let` now.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+    cargo_extra_flags: Vec<String>,
+    command: &'a super::Subcommands,
+    /// Environment variables passed to child processes.
+    env: BTreeMap<OsString, OsString>,
+    /// Additional variables used by environment-altering commands.
+    extra_vars: Extras,
+}
+
+#[derive(Clone, Debug, Default)]
+struct Extras {
```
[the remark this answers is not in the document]
osiewicz: I did it to essentially ensure that these variables are always available in env - and not have to deal with environment lookup in code paths that do use it. My stream of thought was "if we have a valid `Extras` object, sys root and co were obtained successfully".

And yeah, the name sucks.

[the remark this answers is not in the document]
osiewicz: OTOH, `MiriRunner` uses `Extras::default` in its `new` function, which sucks even more. Some of the member functions might see an empty sys root/libdir.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+                "-Zunstable-options -Wrustc::internal -Wrust_2018_idioms -Wunused_lifetimes -Wsemicolon_in_expressions_from_macros ",
+            );
+            if let Some(value) = env {
+                flags_with_warnings.push(value);
+            }
+            // We set the rpath so that Miri finds the private rustc libraries it needs.
+            let mut flags_with_compiler_settings = OsString::from("-C link-args=-Wl,-rpath,");
+            flags_with_compiler_settings.push(&ret.extra_vars.libdir);
+            flags_with_compiler_settings.push(flags_with_warnings);
+            flags_with_compiler_settings
```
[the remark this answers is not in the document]
osiewicz: Ya, I believe I've tried to keep argument order from bash script. I agree that this is cleaner.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,637 @@
...
+        &mut self,
+        key: impl Into<OsString>,
+        value: impl Into<OsString>,
+    ) -> Option<OsString> {
+        self.env.insert(key.into(), value.into())
+    }
+
+    /// Prepare and set MIRI_SYSROOT. Respects `MIRI_TEST_TARGET` and takes into account
+    /// locally built vs. distributed rustc.
+    fn find_sysroot(&mut self) -> Result<()> {
```
RalfJung, line 191: Should probably be `find_miri_sysroot` to avoid confusion with `extras.sysroot`.

Same for `build_sysroot` → `build_miri_sysroot`

miri-script/src/commands.rs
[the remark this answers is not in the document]
osiewicz: Yup, these are good ideas. And yeah, I agree that neither sysroot nor libdir should be in env.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+            extra_vars: Default::default(),
+            cargo_extra_flags,
+        };
+        // Run the auto-things.
+        if let Some(config) = config {
+            if config.toolchain {
+                // Run this first, so that the toolchain doesn't change after
+                // other code has run.
+                let command = Subcommands::Toolchain { flags: vec![] };
+                Self::exec_inner(&command, false)?;
```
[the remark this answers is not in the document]
osiewicz: I've figured that `exec_inner` resembles old script a bit better (we start off a fresh `MiriRunner` instance), but I'd be happy to change it.

osiewicz:
@RalfJung it looks like there's an issue with `./miri toolchain` where on Windows setting an override for a directory doesn't affect subsequent call to `rustc --print sysroot`. Using `dustup run rustc --print sysroot` does propagate the override properly. I'm not sure what I'm doing wrong; I've pulled so much hair on this that at this point I might have to see a barber. :/

RalfJung:
>  it looks like there's an issue with ./miri toolchain where on Windows setting an override for a directory doesn't affect subsequent call to rustc --print sysroot. Using rustup run rustc --print sysroot does respect the override.

I assume the difference is that miri-script itself is run inside a `cargo run` which runs in rustup. So some rustup env vars are set which take precedence over the toolchain file. We should probably unset them / not forward them to programs we invoke.

miri-script/src/commands.rs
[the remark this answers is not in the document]
RalfJung: The control flow in the script was terrible though.^^ We should initialize things once and then in `bench` we know things are initialized enough for `install`.

osiewicz:
Hm, that sounds plausible. What do you say to having something akin to:
`cargo build --manifest-path "$(dirname "$0")"/miri-script/Cargo.toml && "$(dirname "$0")"/miri-script/target/debug/miri-script $@`
instead of:
`cargo run --manifest-path "$(dirname "$0")"/miri-script/Cargo.toml -- $@`
The former should not go through rustup, so it won't set its environment variables - but it will still be possible to set them just in case someone wants to do that.

RalfJung:
Hm, then we have to make assumptions about where cargo puts the build... But that might still be the better option than fiddling with the environment.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,652 @@
...
+    pub(super) fn exec(command: &super::Subcommands) -> Result<()> {
+        Self::exec_inner(command, true)
+    }
+    fn exec_inner(command: &super::Subcommands, run_auto_things: bool) -> Result<()> {
+        let miri_dir = detect_miri_dir()?;
+        let active_toolchain = get_active_toolchain(&miri_dir)?;
+        let config = command.get_auto_config(&miri_dir).filter(|_| run_auto_things);
+        // CARGO_EXTRA_FLAGS do not have to be a valid UTF-8, but that's what shell_words' expects.
+        let cargo_extra_flags = std::env::var("CARGO_EXTRA_FLAGS").unwrap_or_default();
+        let cargo_extra_flags = shell_words::split(&cargo_extra_flags)?;
```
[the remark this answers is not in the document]
osiewicz: https://github.com/rust-lang/cargo/blob/master/src/cargo/core/compiler/build_context/target_info.rs#L761
Yeah, it doesn't. This should simplify things a bit.

miri-script/src/commands.rs
osiewicz: It is a special syntax from xshell; https://docs.rs/xshell/latest/xshell/macro.cmd.html It helps with escaping a bunch.

RalfJung:
I'm going to take over this PR -- at this point I think it is easier if I just implement things the way I prefer, rather than trying to explain to you how I would like this to be changed. :)  Your groundwork has been extremely helpful for this, thanks a ton!

RalfJung:
Looks like there is a problem (which the old code work around in an incorrect way)... for things like `./miri clippy -- -D warnings`, we want all of `-- -D warnings` to be forwarded to clippy. That is, we *don't* want the clap parser to intercept this `--` at all.

Is that possible?

osiewicz:
Hi,
I am sorry for not carrying this further - I've had less time recently for OS work. :/
As for the "--" forwarding issue, I believe I've annotated `clippy` and `fmt` subcommands with `allow_leading_hyphen`. I recall that there were still issues with it tho.
Another problem with what I've had up to this point was that new `fmt` was noticeably slower than the old one. I suppose it came down to formatting files sequentially Vs using `xargs` in a shell script - which could've parallelized different rustfmt instances under the hood.

RalfJung:
> Another problem with what I've had up to this point was that new fmt was noticeably slower than the old one.

I fixed that by running rustfmt a single time on all the files, instead of having a new rustfmt invocation for each file. (That matches what the old script does.)

> As for the "--" forwarding issue, I believe I've annotated clippy and fmt subcommands with allow_leading_hyphen. I recall that there were still issues with it tho.

That's not sufficient -- as the clap docs for `trailing_var_arg` say, "users still have the option to explicitly escape ambiguous arguments with --". I don't know if clap has a way to to disable this. If not, we won't be able to use clap.

All the CI failures currently are caused by this.

miri-script/src/commands.rs
```diff
@@ -0,0 +1,629 @@
...
+        // Then test, and let caller control flags.
+        // Only in root project as `cargo-miri` has no tests.
+        let sh = self.shell()?;
+        if bless {
+            sh.set_var("MIRI_BLESS", "Gesundheit");
+        }
+        let toolchain: &OsStr = self.active_toolchain.as_ref();
+        cmd!(
+            sh,
+            "cargo +{toolchain} test {extra_flags...} --manifest-path {miri_manifest} -- {flags...}"
```
RalfJung: The `-- {flags..}` here is what got CI to work despite the parser issue, but it breaks other things -- for instance, `./miri test --all-features` should pass that flag to cargo (before the `--`), not to the test runner. (Same for clippy.)

PR #2916, opened 2023-06-04
Remove rustc-workspace-hack

ehuss:
The `rustc-workspace-hack` dependency was removed in https://github.com/rust-lang/rust/pull/109133 and should no longer be needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -442,7 +442,6 @@ dependencies = [
  "measureme",
  "rand",
  "regex",
- "rustc-workspace-hack",
  "rustc_version",
  "smallvec",
  "ui_test",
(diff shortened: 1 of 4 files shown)


RalfJung:
This sounds like a PR that would be better to land in rustc first, not in the subrepo? That's where it will lead to build errors if things go wrong, after all.

ehuss:
I can do it there if you want, but I don't suspect there will be any issues. `rustc-workspace-hack` is an empty dependency with no dependencies itself (it really is a no-op). All the support infrastructure in rust-lang/rust has been removed.

RalfJung:
Ah, fair.

cargo-miri/Cargo.lock will probably also need updating.

ehuss:
The lock should be updated now.

PR #2918, opened 2023-06-05
TB: improve error messages (distinguish between accesses and reborrows)

Vanille-N:
I don't remember where, but there was a complaint that while SB's error messages distinguish between "invalidated by an access" and "invalidated by a reborrow", TB's error messages do not.

### DIFF
--- a/src/borrow_tracker/tree_borrows/diagnostics.rs
+++ b/src/borrow_tracker/tree_borrows/diagnostics.rs
@@ -12,16 +12,46 @@ use crate::borrow_tracker::tree_borrows::{
 use crate::borrow_tracker::{AccessKind, ProtectorKind};
 use crate::*;
 
+/// Cause of an access: either a real access or one
+/// inserted by Tree Borrows due to a reborrow or a deallocation.
+#[derive(Clone, Copy, Debug)]
+pub enum AccessCause {
+    Explicit(AccessKind),
+    Reborrow,
+    Dealloc,
+}
+
+impl fmt::Display for AccessCause {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Explicit(kind) => write!(f, "{kind}"),
+            Self::Reborrow => write!(f, "reborrow"),
+            Self::Dealloc => write!(f, "deallocation"),
+        }
+    }
+}
+
+impl AccessCause {
+    fn print_as_access(self, is_foreign: bool) -> String {
+        let rel = if is_foreign { "foreign" } else { "child" };
+        match self {
+            Self::Explicit(kind) => format!("{rel} {kind}"),
+            Self::Reborrow => format!("reborrow (acting as a {rel} read access)"),
+            Self::Dealloc => format!("deallocation (acting as a {rel} write access)"),
+        }
+    }
+}
+
 /// Complete data for an event:
 #[derive(Clone, Debug)]
 pub struct Event {
     /// Transformation of permissions that occured because of this event
     pub transition: PermTransition,
     /// Kind of the access that triggered this event
-    pub access_kind: AccessKind,
+    pub access_cause: AccessCause,
     /// Relative position of the tag to the one used for the access
     pub is_foreign: bool,
-    /// User-visible range of the access
+    /// Whether this access was explicit or inserted implicitly by Tree Borrows.
     pub access_range: AllocRange,
     /// The transition recorded by this event only occured on a subrange of
     /// `access_range`: a single access on `access_range` triggers several events,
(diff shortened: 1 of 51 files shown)


tests/fail/both_borrows/aliasing_mut1.tree.stderr
```diff
@@ -16,7 +16,8 @@ help: the accessed tag <TAG> later transitioned to Frozen due to a foreign read
    |
 LL | pub fn safe(x: &mut i32, y: &mut i32) {
    |                          ^
-   = help: this corresponds to a loss of write permissions
+   = help: this transition corresponds to a loss of write permissions
+   = help: the read access that caused this transition was inserted implicitly due to a reborrow
```
RalfJung, line 20: I think what I'd like to see here is a change already further up in this message:
```
help: the accessed tag <TAG> later transitioned to Frozen due to a foreign reborrow at offsets [0x0..0x4]
  --> $DIR/aliasing_mut1.rs:LL:CC
   |
LL | pub fn safe(x: &mut i32, y: &mut i32) {
   |                          ^
   = help: this transition corresponds to a loss of write permissions
```
Maybe with another line saying
```
   = help: reborrows act like read access for the purpose of the aliasing rules
```
though I am not sure if that is needed (we don't have it for SB)

tests/fail/both_borrows/newtype_pair_retagging.tree.stderr
```diff
@@ -7,7 +7,8 @@ LL |     unsafe { __rust_dealloc(ptr, layout.size(), layout.align()) }
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Tree Borrows rules it violated are still experimental
    = help: the accessed tag <TAG> is foreign to the protected tag <TAG> (i.e., it is not a child)
    = help: the access would cause the protected tag <TAG> to transition from Frozen to Disabled
-   = help: this is a loss of read permissions, which is not allowed for protected tags
+   = help: this transition would be a loss of read permissions, which is not allowed for protected tags
+   = help: the write access that caused this error was inserted implicitly due to a deallocation
```
RalfJung, line 11: Correspondingly, here the message should say "deallocation through \<TAG> forbidden"

tests/fail/both_borrows/aliasing_mut1.tree.stderr
Vanille-N: Current phrasing gives both: "the accessed tag <TAG> later transitioned to Frozen due to a reborrow (acting as a foreign read access) at offsets [0x0..0x4]"

src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -250,46 +281,46 @@ impl TbError<'_> {
...
                 let conflicting_tag_name =
                     if accessed_is_conflicting { "accessed" } else { "conflicting" };
-                let title = format!("{kind} through {accessed} is forbidden");
                 let mut details = Vec::new();
                 if !accessed_is_conflicting {
                     details.push(format!(
                         "the accessed tag {accessed} is a child of the conflicting tag {conflicting}"
                     ));
                 }
+                let access = cause.print_as_access(false);
```
RalfJung, line 299:
```suggestion
                let access = cause.print_as_access(/* is_foreign*/ false);
```

src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -250,46 +281,46 @@ impl TbError<'_> {
...
                 details.push(format!(
-                    "the {conflicting_tag_name} tag {conflicting} has state {perm} which forbids child {kind}es"
+                    "the {conflicting_tag_name} tag {conflicting} has state {perm} which forbids this {access}"
                 ));
                 (title, details, conflicting_tag_name)
             }
             ProtectedTransition(transition) => {
                 let conflicting_tag_name = "protected";
-                let title = format!("{kind} through {accessed} is forbidden");
+                let access = cause.print_as_access(true);
```
RalfJung, line 307:
```suggestion
                let access = cause.print_as_access(/* is_foreign*/ true);
```

RalfJung (approved):
I like it. :) Just a comment nit. Also please squash.

src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -12,16 +12,46 @@ use crate::borrow_tracker::tree_borrows::{
...
 pub struct Event {
     /// Transformation of permissions that occured because of this event
     pub transition: PermTransition,
     /// Kind of the access that triggered this event
-    pub access_kind: AccessKind,
+    pub access_cause: AccessCause,
     /// Relative position of the tag to the one used for the access
     pub is_foreign: bool,
-    /// User-visible range of the access
+    /// Whether this access was explicit or inserted implicitly by Tree Borrows.
```
deltragon, line 54: Was the changing of this comment on purpose?

Vanille-N: No, this is definitely a mistake. Two mistakes in fact, since this is not a comment changed but rather one comment removed that should have been kept and an unrelated comment that should have been removed.

PR #2931, opened 2023-06-16
Treat thread-local statics on main thread as static roots for leakage analysis

### LINKED ISSUE #2881: Data reachable from thread-local storage of the main thread should not be considered leaked
Miri does not detect any leak for the following code since the reference is stored in a `static` variable and is therefore accessible for the whole duration of the program.
```rust
use std::sync::Mutex;

static REF: Mutex<Option<&'static i32>> = Mutex::new(None);

pub fn main() {
    let a = 123;
    let b = Box::new(a);
    let r = Box::leak(b);
    println!("{r}");
    *REF.lock().unwrap() = Some(r);
}
```

However, replacing the `static` with a `thread_local!`, Miri complains about a memory leak:
```rust
use std::cell::Cell;

// static REF: Mutex<Option<&'static i32>> = Mutex::new(None);

pub fn main() {
    let a = 123;
    let b = Box::new(a);
    let r = Box::leak(b);
    println!("{r}");

    thread_local! {
        static REF: Cell<Option<&'static i32>> = Cell::new(None);
    }

    REF.with(|cell| {
        cell.set(Some(r));
    })
}
```

max-heller:
Miri currently treats allocations as leaked if they're only referenced in thread-local statics. For threads other than the main thread, this is correct, since the thread can terminate before the program does, but references in the main thread's locals should be treated as living for the duration of the program since the thread lives for the duration of the program.

This PR adds thread-local statics and TLS keys as "static roots" for leakage analysis, but does not yet bless the example program from #2881. See https://github.com/rust-lang/miri/issues/2881#issuecomment-1585666652

Closes #2881

### DIFF
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -33,6 +33,15 @@ enum SchedulingAction {
     Sleep(Duration),
 }
 
+/// What to do with TLS allocations from terminated threads
+pub enum TlsAllocAction {
+    /// Deallocate backing memory of thread-local statics as usual
+    Deallocate,
+    /// Skip deallocating backing memory of thread-local statics and consider all memory reachable
+    /// from them as "allowed to leak" (like global `static`s).
+    Leak,
+}
+
 /// Trait for callbacks that can be executed when some event happens, such as after a timeout.
 pub trait MachineCallback<'mir, 'tcx>: VisitTags {
     fn call(&self, ecx: &mut InterpCx<'mir, 'tcx, MiriMachine<'mir, 'tcx>>) -> InterpResult<'tcx>;
@@ -1051,7 +1060,8 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                         // See if this thread can do something else.
                         match this.run_on_stack_empty()? {
                             Poll::Pending => {} // keep going
-                            Poll::Ready(()) => this.terminate_active_thread()?,
+                            Poll::Ready(()) =>
+                                this.terminate_active_thread(TlsAllocAction::Deallocate)?,
                         }
                     }
                 }
@@ -1066,21 +1076,29 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
     }
 
     /// Handles thread termination of the active thread: wakes up threads joining on this one,
-    /// and deallocated thread-local statics.
+    /// and deals with the thread's thread-local statics according to `tls_alloc_action`.
     ///
     /// This is called by the eval loop when a thread's on_stack_empty returns `Ready`.
     #[inline]
-    fn terminate_active_thread(&mut self) -> InterpResult<'tcx> {
+    fn terminate_active_thread(&mut self, tls_alloc_action: TlsAllocAction) -> InterpResult<'tcx> {
         let this = self.eval_context_mut();
         let thread = this.active_thread_mut();
         assert!(thread.stack.is_empty(), "only threads with an empty stack can be terminated");
         thread.state = ThreadState::Terminated;
 
         let current_span = this.machine.current_span();
-        for ptr in
-            this.machine.threads.thread_terminated(this.machine.data_race.as_mut(), current_span)
-        {
-            this.deallocate_ptr(ptr.into(), None, MiriMemoryKind::Tls.into())?;
+        let thread_local_allocations =
+            this.machine.threads.thread_terminated(this.machine.data_race.as_mut(), current_span);
+        for ptr in thread_local_allocations {
+            match tls_alloc_action {
+                TlsAllocAction::Deallocate =>
+                    this.deallocate_ptr(ptr.into(), None, MiriMemoryKind::Tls.into())?,
+                TlsAllocAction::Leak =>
+                    if let Some(alloc) = ptr.provenance.get_alloc_id() {
+                        trace!("Thread-local static leaked and stored as static root: {:?}", alloc);
+                        this.machine.static_roots.push(alloc);
+                    },
+            }
         }
         Ok(())
     }
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -11,6 +11,7 @@ use log::info;
 use rustc_middle::ty::Ty;
 
 use crate::borrow_tracker::RetagFields;
+use crate::concurrency::thread::TlsAllocAction;
 use crate::diagnostics::report_leaks;
 use rustc_data_structures::fx::FxHashSet;
 use rustc_hir::def::Namespace;
@@ -244,9 +245,9 @@ impl MainThreadState {
                 // Figure out exit code.
                 let ret_place = this.machine.main_fn_ret_place.clone().unwrap();
                 let exit_code = this.read_target_isize(&ret_place)?;
-                // Need to call this ourselves since we are not going to return to the scheduler
-                // loop, and we want the main thread TLS to not show up as memory leaks.
-                this.terminate_active_thread()?;
+                // Deal with our thread-local memory. We do *not* want to actually free it, instead we consider TLS
+                // to be like a global `static`, so that all memory reached by it is considered to "not leak".
+                this.terminate_active_thread(TlsAllocAction::Leak)?;
                 // Stop interpreter loop.
                 throw_machine_stop!(TerminationInfo::Exit { code: exit_code, leak_check: true });
             }
--- a/tests/pass/concurrency/tls_leak_main_thread_allowed.rs
+++ b/tests/pass/concurrency/tls_leak_main_thread_allowed.rs
@@ -1,0 +1,21 @@
+//@ignore-target-windows: Windows uses a different mechanism for `thread_local!`
+#![feature(thread_local)]
+
+use std::cell::Cell;
+
+// Thread-local variables in the main thread are basically like `static` (they live
+// as long as the program does), so make sure we treat them the same for leak purposes.
+pub fn main() {
+    thread_local! {
+        static TLS_KEY: Cell<Option<&'static i32>> = Cell::new(None);
+    }
+
+    TLS_KEY.with(|cell| {
+        cell.set(Some(Box::leak(Box::new(123))));
+    });
+
+    #[thread_local]
+    static TLS: Cell<Option<&'static i32>> = Cell::new(None);
+
+    TLS.set(Some(Box::leak(Box::new(123))));
+}
--- a/tests/fail/leak_in_lib_tls.rs
+++ b/tests/fail/leak_in_lib_tls.rs
@@ -1,0 +1,21 @@
+//@error-in-other-file: memory leaked
+//@normalize-stderr-test: ".*│.*" -> "$$stripped$$"
+
+use std::cell::Cell;
+
+pub fn main() {
+    thread_local! {
+        static TLS: Cell<Option<&'static i32>> = Cell::new(None);
+    }
+
+    std::thread::spawn(|| {
+        TLS.with(|cell| {
+            cell.set(Some(Box::leak(Box::new(123))));
+        });
+    })
+    .join()
+    .unwrap();
+
+    // Imagine the program running for a long time while the thread is gone
+    // and this memory still sits around, unused -- leaked.
+}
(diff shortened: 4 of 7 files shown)


oli-obk:
I did some extra logging, and the issue turns out to be that all the static roots you added are deallocated at the end of the program, causing the leaked allocations within to not be detected anymore.

If you do not leak, then the thread local statics (even of the main thread) get freed. As an example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3c4634970fddcc8784f4376a8e1241fd passes without leaking.

This means we cannot just apply what was proposed in https://github.com/rust-lang/miri/issues/2881#issuecomment-1579988825

Knowing this, do you still think it's worth it to add this extra logic for something that goes away if the main thread TLS actually cleans up its memory?

RalfJung:
Oh right, we'd have to also skip deallocating main thread TLS.

max-heller:
> Knowing this, do you still think it's worth it to add this extra logic for something that goes away if the main thread TLS actually cleans up its memory?

Does this mean the value actually doesn't live for the lifetime of the program, or is it still valid to consider the program as ended before the main thread cleans up its TLS?

> Oh right, we'd have to also skip deallocating main thread TLS.

Would this mean missing out on analysis of main thread TLS destructors? E.g. if a destructor contained UB we'd no longer be able to detect it?

Instead of skipping deallocation, could we run the leakage check before the main thread cleans up?

RalfJung:
We should definitely run the destructor. We just shouldn't deallocate the backing storage.

tests/fail/leak_in_tls.rs
```diff
@@ -0,0 +1,24 @@
+//@error-in-other-file: memory leaked
+//@normalize-stderr-test: ".*│.*" -> "$$stripped$$"
+
+use std::cell::Cell;
+
+pub fn main() {
+    thread_local! {
+        static REF: Cell<Option<&'static i32>> = Cell::new(None);
+    }
```
RalfJung, line 9: So this will test that we allow leaks in one of the 2 kinds of TLS (TLS statics vs OS-supported TLS keys). We should make sure to have tests covering both (by directly using the underlying mechanisms).

oli-obk:
looks like the msvc thread locals need a similar treatment. I think they have a completely separate code path

tests/fail/leak_in_tls.rs
max-heller: Added `#[thread_local]`s to the tests

max-heller:
> looks like the msvc thread locals need a similar treatment. I think they have a completely separate code path

separate from this path, you mean?

https://github.com/rust-lang/miri/blob/6f688d5907b93c0fad3664ec0e113328732a6290/src/shims/windows/foreign_items.rs#L241-L269

tests/fail/leak_in_tls.rs
RalfJung: I think we want one test that uses `#[thread_local]` and one that uses pthread TLS keys directly.

Or we only treat `#[thread_local]` as static roots? Then we cover all targets with `target_thread_local`. Doing special magic to library stuff like the pthread keys is somewhat dubious anyway...

src/concurrency/thread.rs
```diff
@@ -1070,10 +1070,26 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
-        for ptr in
-            this.machine.threads.thread_terminated(this.machine.data_race.as_mut(), current_span)
-        {
-            this.deallocate_ptr(ptr.into(), None, MiriMemoryKind::Tls.into())?;
+        let thread_local_allocations =
+            this.machine.threads.thread_terminated(this.machine.data_race.as_mut(), current_span);
+        match this.get_active_thread() {
+            // For main thread, add thread-local statics to static roots and skip deallocating
+            // backing memory
+            ThreadId(0) =>
```
RalfJung, line 1078: Instead of matching here, it makes more sense to change the code here:

https://github.com/rust-lang/miri/blob/8bbc93ca31f29b10168919f241d9c9be6b69f77e/src/eval.rs#L250-L252

the main thread is already special, so instead of calling to common code and then special-casing again -- just don't call to common code. :)

RalfJung:
Sorry for pointing you to `shims/tls.rs`, that was probably the wrong starting point. I think in a first step we just want to replace this here

https://github.com/rust-lang/miri/blob/8bbc93ca31f29b10168919f241d9c9be6b69f77e/src/eval.rs#L250-L252

by code that instead marks the `thread_local` statics of the main thread as "allowed to leak".
Then we should have tests:
- for having some leaked memory in a `#[thread_local] static`.
- for having some leaked memory in a `tread_local! { static }`. This test should `ignore-windows` since some Windows targets don't use `#[thread_local]` for their ``tread_local!`.

tests/fail/leak_in_os_tls.rs
RalfJung, line 1: `leak_in_lib_tls` would be a better name I think

tests/fail/leak_in_tls.rs
RalfJung, line 1: `leak_in_static_tls`, maybe?

In fact this is the only new fail test we really need, I think.

tests/fail/leak_in_tls.rs
```diff
@@ -0,0 +1,20 @@
+//@error-in-other-file: memory leaked
+//@normalize-stderr-test: ".*│.*" -> "$$stripped$$"
+
+#![feature(thread_local)]
+
+use std::cell::Cell;
+
+pub fn main() {
```
RalfJung, line 8:
```suggestion
/// Ensure that leaks through `thread_local` statics *not in the main thread*
/// are detected.
pub fn main() {
```

tests/pass/issues/issue-miri-2881.rs
RalfJung, line 1: Please call this something like `pass/concurrency/tls_leak_main_thread_allowed.rs`

src/shims/tls.rs
```diff
@@ -105,12 +105,25 @@ impl<'tcx> TlsData<'tcx> {
...
+                    if thread_id.to_u32() == 0 {
+                        if let Some(alloc) = (new_data.to_pointer(cx).ok())
+                            .and_then(|ptr| ptr.provenance.and_then(|prov| prov.get_alloc_id()))
+                        {
+                            trace!(
+                                "TLS key {} for main thread stored as static root: {:?}",
+                                key,
+                                alloc,
+                            );
+                            static_roots.push(alloc);
```
RalfJung, line 124: This is definitely not what we want -- if a value is set for some TLS variable and then changed again, only the final value should be allowed to leak.

So if you want to also tackle library TLS state (which will be needed to make the test pass on Windows), that must happen around here

https://github.com/rust-lang/miri/blob/fcabf757701254a4d25e7ea6dc91f445f0eed14c/src/shims/tls.rs#L274

The `on_stack_empty` function should receive a flag indicating whether it should in the end leak or free the remaining thread-local state. [Here](https://github.com/rust-lang/miri/blob/8bbc93ca31f29b10168919f241d9c9be6b69f77e/src/eval.rs#L219) we'd pass in "leak" but [here](https://github.com/rust-lang/miri/blob/8bbc93ca31f29b10168919f241d9c9be6b69f77e/src/concurrency/thread.rs#L825) it should be "free".

This is a stretch goal if you want to tackle the Windows issue in the same PR. It's also fine to leave this to a future PR.

src/concurrency/thread.rs
max-heller: I tried doing this, but doing so causes many tests to fail since we do still want to run some of the other logic in `terminate_active_thread()` (e.g. waiting on other threads to finish):
> Error: the main thread terminated without waiting for all remaining threads

RalfJung: Ah, fair. We  probably need a flag to be passed to `terminate_active_thread` then indicating whether the thread-locals should be freed or leaked.

RalfJung:
@max-heller this is waiting for you to resolve the remaining comments. See the open threads above, in particular [here](https://github.com/rust-lang/miri/pull/2931#discussion_r1299207882) and [here](https://github.com/rust-lang/miri/pull/2931#discussion_r1284064104).

Once this is ready for review again, please use `@rustbot ready`.

src/concurrency/thread.rs
max-heller: Added a `leak_tls` flag

src/shims/tls.rs
max-heller: Removed the partial implementation to leave Windows for a future PR

max-heller:
Addressed the open requests, sorry for the delay.
@rustbot ready

RalfJung:
No worries, it's not like I had fast reaction times on this PR in the past. :)

src/eval.rs
```diff
@@ -246,7 +246,7 @@ impl MainThreadState {
                 let exit_code = this.read_target_isize(&ret_place)?;
                 // Need to call this ourselves since we are not going to return to the scheduler
                 // loop, and we want the main thread TLS to not show up as memory leaks.
-                this.terminate_active_thread()?;
+                this.terminate_active_thread(true)?;
```
RalfJung, line 249:
```suggestion
                this.terminate_active_thread(/* leak_tls */ true)?;
```
anonymous bools are an anti-pattern; either use an enum or at least add comments that indicate their meaning

src/concurrency/thread.rs
```diff
@@ -1066,21 +1066,33 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
         thread.state = ThreadState::Terminated;
 
         let current_span = this.machine.current_span();
-        for ptr in
-            this.machine.threads.thread_terminated(this.machine.data_race.as_mut(), current_span)
-        {
-            this.deallocate_ptr(ptr.into(), None, MiriMemoryKind::Tls.into())?;
+        let thread_local_allocations =
+            this.machine.threads.thread_terminated(this.machine.data_race.as_mut(), current_span);
+        if leak_tls {
```
RalfJung, line 1083: Please move the `if` inside the loop. No reason to duplicate that logic.

src/eval.rs
```diff
@@ -246,7 +246,7 @@ impl MainThreadState {
                 let exit_code = this.read_target_isize(&ret_place)?;
                 // Need to call this ourselves since we are not going to return to the scheduler
                 // loop, and we want the main thread TLS to not show up as memory leaks.
```
RalfJung, line 248:
```suggestion
                // Deal with our thread-local memory. We do *not* want to actually free it, instead we consider TLS
                // to be like a global `static`, so that all memory reached by it is considered to "not leak".
```

tests/pass/concurrency/tls_leak_main_thread_allowed.rs
```diff
@@ -0,0 +1,19 @@
+//@ignore-target-windows: Windows uses a different mechanism for `thread_local!`
+#![feature(thread_local)]
+
+use std::cell::Cell;
+
+pub fn main() {
```
RalfJung, line 6:
```suggestion
// Thread-local variables in the main thread are basically like `static` (they live
// as long as the program does), so make sure we treat them the same for leak purposes.
pub fn main() {
```

RalfJung:
@rutsbot author

src/concurrency/thread.rs
```diff
@@ -1066,22 +1067,39 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
         Ok(())
     }
 }
+
+/// What to do with TLS allocations from terminated threads
+pub enum TerminatedThreadTls {
+    /// Deallocate backing memory of thread-local statics as usual
+    Deallocate,
+    /// Skip deallocating backing memory of thread-local statics and consider all memory reachable
+    /// from them as not leaked (like global `static`s).
```
RalfJung, line 1103:
```suggestion
    /// from them as "allowed to leak" (like global `static`s).
```

src/concurrency/thread.rs
```diff
@@ -1066,22 +1067,39 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                        this.machine.static_roots.push(alloc);
+                    },
+            }
         }
         Ok(())
     }
 }
+
+/// What to do with TLS allocations from terminated threads
+pub enum TerminatedThreadTls {
```
RalfJung, line 1099: Please move this to the top of the file. Type declarations should generally be at the top.

Also, the type name isn't exactly informative... maybe let's call it `TlsAllocAction`?

RalfJung:
@rustbot author
I assume this is the last round, so please also squash this into a single commit.

RalfJung:
Awesome, thanks for staying with us until this is finally done. :)
@bors r+

PR #2935, opened 2023-06-18
Rustup

RalfJung:


saethlin:
Uh-oh. Tests shouldn't need to be blessed when syncing back right?

oli-obk:
sometimes they do, like if we added new tests on the miri side, then the rustc side won't know about them yet, and can't have blessed them.

saethlin:
The diff that is showing up when running the tests is
```diff
-   = note: inside `alloc::alloc::box_free::<i32, std::alloc::Global>` at RUSTLIB/alloc/src/alloc.rs:LL:CC
+   = note: inside `<std::boxed::Box<i32> as std::ops::Drop>::drop` at RUSTLIB/alloc/src/boxed.rs:LL:CC
```
But this is bringing in a PR that removes `box_free`. So I feel like this diff should have already been applies when the rust-lang/rust PR landed.

RalfJung:
These tests don't exist in the rustc repo yet, so they can't be blessed there.

PR #2936, opened 2023-06-19
Optional semantics for `Unique`

Vanille-N:
Use with `-Zmiri-unique-is-unique`, makes `core::ptr::Unique` get a reborrow with its own permissions.


src/bin/miri.rs
```diff
@@ -557,6 +559,14 @@ fn main() {
             rustc_args.push(arg);
         }
     }
+    // `-Zmiri-unique-is-unique` should only be used with `-Zmiri-tree-borrows`
+    if miri_config.unique_is_unique
+        && !matches!(miri_config.borrow_tracker, Some(BorrowTrackerMethod::TreeBorrows))
+    {
+        eprintln!(
+            "WARNING: -Zmiri-unique-is-unique only has an effect when -Zmiri-tree-borrows is also used"
+        );
```
RalfJung, line 568:
```suggestion
        show_error!(
            "-Zmiri-unique-is-unique only has an effect when -Zmiri-tree-borrows is also used"
        );
```

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -407,8 +430,16 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 self.ecx
             }
 
+            // Regardless of whether special behavior is enforced for `Unique`
+            // (`unique_is_unique`), we always treat `Box`es the same way, though they get an extra
+            // read access.
```
RalfJung, line 435: What is the extra read access?

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -552,3 +592,28 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         tree_borrows.give_pointer_debug_name(tag, nth_parent, name)
     }
 }
+
+/// Takes a place for a `Unique` and turns it into a place for a dereferenceable
+/// pointer.
+/// I.e. input is what you get from the visitor upon encountering an `adt` that is `Unique`,
+/// and output can be used by `retag_ptr_inplace`.
+fn inner_place_of_unique<'tcx>(
```
RalfJung, line 600:
```suggestion
/// Takes a place for a `Unique` and turns it into a place with the inner raw pointer.
/// I.e. input is what you get from the visitor upon encountering an `adt` that is `Unique`,
/// and output can be used by `retag_ptr_inplace`.
fn inner_ptr_of_unique<'tcx>(
```

src/borrow_tracker/tree_borrows/mod.rs
Vanille-N: My bad, this is outdated. It should say instead that `Unique` gets a `zero_size` while `Box` has an actual non-zero-sized reborrow.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -374,15 +382,30 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+            (false, _) => None,             // No special handling for `Unique`
+            (true, Some(did)) => Some(did), // Found the `DefId` that requires special handling.
+            (true, None) => {
+                // No `Unique` defined or no efficient way to identify it.
+                // Ignore the command-line flag after a Warning.
+                eprintln!(
+                    r#"WARNING: `-Zmiri-unique-is-unique` has no effect if no `Unique` pointer exists and will be ignored. Use `#[lang = "ptr_unique"]` to declare the struct that should be treated as `Unique`"#
+                );
+                options.unique_is_unique = false;
+                None
```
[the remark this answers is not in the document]
RalfJung: Yeah, elsewhere we just do `Some(adt.did()) == self.ecx.tcx.lang_items().unsafe_cell_type()`. :shrug: fine for me either way

oli-obk:
did some discussions related to this happen somewhere? I'm not against this change, just curious since it came out of nowhere from my perspective :D

RalfJung:
No, this is an experiment Neven and me discussed.

`Unique` having semantics came up in https://github.com/rust-lang/unsafe-code-guidelines/issues/384, https://github.com/rust-lang/unsafe-code-guidelines/issues/326.

tests/fail/tree_borrows/unique.uniq.stderr
```diff
@@ -0,0 +1,38 @@
...
+help: the accessed tag <TAG> was created here
+  --> $DIR/unique.rs:LL:CC
+   |
+LL |     let uniq = Unique::new(rawptr).unwrap();
+   |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+help: the conflicting tag <TAG> was created here, in the initial state Reserved
+  --> $DIR/unique.rs:LL:CC
+   |
+LL |     let uniq = Unique::new(rawptr).unwrap();
+   |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
RalfJung, line 19: This looks a bit strange, the accessed and conflict tag seem to be the same?

Is this a new thing with unique or an existing diagnostics issue?

tests/pass/tree_borrows/tree-borrows.rs
```diff
@@ -1,4 +1,6 @@
+//@revisions: default uniq
 //@compile-flags: -Zmiri-tree-borrows
+//@[uniq]compile-flags: -Zmiri-unique-is-unique
```
RalfJung, line 3: Nothing here actually uses unique, or does it?

tests/pass/tree_borrows/unique.rs
```diff
@@ -0,0 +1,37 @@
...
+        name!(raw);
+
+        // We create a Unique and expect it to have a fresh tag
+        // and uninitialized permissions.
+        let uniq = Unique::new(raw).unwrap();
+        name!(uniq.as_ptr(), "uniq");
+        print_state!(alloc_id);
+
+        // We can activate the Unique and use it mutably.
+        *uniq.as_ptr() = 42;
```
RalfJung, line 31: Might be good to also print the state here?

RalfJung (commented):
Very nice, just some nits :)

tests/pass/tree_borrows/tree-borrows.rs
Vanille-N: There are `Vec`s and `Box`es in the test.

RalfJung: Fair.

tests/fail/tree_borrows/unique.uniq.stderr
Vanille-N: These are in fact not the same, there's a special case in the diagnostics when they are (the line "`= help: the accessed tag <TAG> is a child of the conflicting tag <TAG>`" would not be printed if they were equal).
Those are two different tags, but the `Span` for the location where they were created is the same.

RalfJung: Ah, I guess there are two retags here, due to the `unwrap`...

Maybe use `new_unchecked` instead? That doesn't solve the general problem (when there are many conflicting pointers, maybe we shouldn't always report the closest like here), but could at least help for this test.

Vanille-N: There might be a bug somewhere else in the retagging, `new_unchecked` doesn't seem to do *any* retagging.

Vanille-N: If I use `new_unchecked` then there's no difference in this test between having `-Zmiri-unique-is-unique` or not.

RalfJung: Ah, that's a problem.

You might have to adjust this function: https://github.com/rust-lang/rust/blob/08384b854f4bcbba9459f22665063e7c45877dfe/compiler/rustc_mir_transform/src/add_retag.rs#L13-L15.

tests/pass/tree_borrows/tree-borrows.rs
saethlin: Will we get any test breakage if libs decides to remove `ptr::Unique` from `Vec`? Or would that be just okay?

Vanille-N:
Rebased at [README: -Zmiri-unique-is-unique](https://github.com/rust-lang/miri/pull/2936/commits/8ea5c3f6109520533ec1f788718e4d9b9510181f)

Tests will fail.

RalfJung:
If you rebase over latest Miri master (and run `./miri toolchain`) you should get a rustc with your retag PR.

Vanille-N:
The problem here is that `Unique::as_ptr` reborrows and thus does not hand out the actual pointer inside the `Unique`.

### Problem

The following diff illustrates the issue:
```diff
--- tests/pass/tree_borrows/unique.uniq.stderr
 | Act|      └─┬──<TAG=base>
 | Act|        └─┬──<TAG=raw>
-| Act|          └────<TAG=uniq, uniq>
+|?Dis|          ├────<TAG=uniq>
+|?Dis|          └────<TAG=uniq>
```
`uniq` was named through `_.as_ptr()`, and independent invocations of `as_ptr` hand out cousin pointers.

This part of the diff shows that the wrong pointer is being named, since it gets invalidates when the pointer inside `Unique` should not be.
```diff
-| Act|
+|?Dis|
```

This part of the diff shows that pointers obtained from `as_ptr` are cousins instead of being identical.
```
└────<TAG=uniq, uniq>
├────<TAG=uniq>
└────<TAG=uniq>
```

### Solution

In the long term, if we find out that `Unique` cannot be used for `Vec` or that there is too much breakage from `-Zmiri-unique-is-unique`, this is likely to be a cause.
We will have to decide (1) whether that is indented behavior, and (2) if it is then is it the fault of TB or the fault of the implementation of `as_ptr` ?

Right now, this illustrates that `miri_pointer_name` is not capable of naming the tags that we want to name without breaking upon `-Zmiri-unique-is-unique`, see [TODO] for an idea of a solution.

RalfJung:
Isn't this a bigger problem than just the ui tests? This means we cannot do the equivalent of `addr_of_mut!(*mutable_ref)`, i.e. create a raw pointer with the tag of the `noalias` pointer (or an equivalent tag).

I guess if the tests still pass we can keep going with the experiment, but eventually we might want a new borrow state for this, one where alias requirements are only enforced while the tag is protected.

RalfJung:

FWIW github doesn't send notification for this so it's pretty much useless. Thankfully we have a bot that helps: @rustbot ready. :)

tests/pass/tree_borrows/tree-borrows.rs
RalfJung: We will get test breakage but that seems fine for me, we can then just adjust the tests.

tests/pass/tree_borrows/unique.rs
```diff
@@ -0,0 +1,40 @@
...
+        let alloc_id = alloc_id!(base);
+        name!(base);
+
+        let raw = &mut *base as *mut u8;
+        name!(raw);
+
+        // We create a `Unique` and expect it to have a fresh tag
+        // and uninitialized permissions.
+        let uniq = Unique::new_unchecked(raw);
+        let intermediate = if cfg!(uniq) { 2 } else { 0 };
```
RalfJung, line 27: `intermediate`? I don't understand the name. This is about how many parents we need to walk up for to name the pointer, right?

In the macro docs this is called `nth_parent`, so maybe also use that name here, and explain why it needs an `if`.
(Also I just noticed in the macro sources it is called `nb`, whatever that means. Can you change that to `nth_parent` while we are at it?)

tests/pass/tree_borrows/vec_unique.rs
```diff
@@ -0,0 +1,48 @@
...
+    unsafe {
+        let base = vec![0u8, 1];
+        let alloc_id = alloc_id!(base.as_ptr());
+
+        // Doing this twice to illustrate how treating `Unique` differently
+        // changes `as_ptr`:
+        // - with `default`, `as_ptr` always returns the same pointer without
+        //   extra reborrow: two `as_ptr` from the same `Vec` retun the *same pointer*,
+        // - with `uniq` however, the two returned pointers are now cousins.
+        // This is probably problematic.
```
RalfJung, line 25: In the tree dump it looks like the 2 returned pointers are the same?

tests/fail/tree_borrows/kill-copies.rs
```diff
@@ -0,0 +1,43 @@
...
+        unique_keeps_copies(Unique::new_unchecked(raw));
+
+        // Ultimately the intended behavior is that both above tests would
+        // succeed.
+        std::hint::unreachable_unchecked();
+        //~[default]^ ERROR: entering unreachable code
+    }
+}
+
+unsafe fn ref_mut_keeps_copies(x: &mut u8) {
```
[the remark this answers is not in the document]
Vanille-N: (`cp` = copy :wink: )
But I agree that the test is better with `child1` and `child2`.

tests/fail/tree_borrows/children-can-alias.rs
```diff
@@ -0,0 +1,56 @@
...
+    child1.write(1);
+    child2.write(2);
+}
+
+unsafe fn raw_children_of_unique_can_alias(x: Unique<u8>) {
+    let child1 = x.as_ptr();
+    let child2 = x.as_ptr();
+    // Under `-Zmiri-unique-is-unique`, `Unique` accidentally offers more guarantees
+    // than `&mut`. Not because it responds differently to accesses but because
+    // there is no way to obtain a copy with the same tag.
```
RalfJung, line 44:
```suggestion
    // there is no easy way to obtain a copy with the same tag.
```
There is always a way: `transmute_copy`. ;)

RalfJung (commented):
Some tweaks for the comments -- please apply the same changes to all copies of these comments.

Then please squash the commits and we can land this. :)

PR #2939, opened 2023-06-22
Provenance forging for `miri_pointer_name`

Vanille-N:
Shims to give `miri_pointer_name` a tag in a non-hacky way that is compatible with current `Unique` experiments as well as possible future GC developments.

Previous way of naming an unreachable tag was to refer to the nth parent of a reachable tag. This is very unreliable in the presence of `-Zmiri-unique-is-unique` which changes the number of intermediate tags between reachable tags (`unique.as_ptr()`) and interesting tags (`unique.ptr`).

A more reliable method is proposed here, with
- no more nth parent handling in `miri_pointer_name` itself,
- nth parent accessor still available but extracted to its own function and discouraged for n > 1 (only a macro for the direct parent),
- refering to a tag as the nearest common ancestor of two reachable tags.

This will provide the tools to easily solve the diagnostics aspect of the recent blocker for #2936

### DIFF
--- a/src/borrow_tracker/mod.rs
+++ b/src/borrow_tracker/mod.rs
@@ -318,7 +318,6 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
     fn give_pointer_debug_name(
         &mut self,
         ptr: Pointer<Option<Provenance>>,
-        nth_parent: u8,
         name: &str,
     ) -> InterpResult<'tcx> {
         let this = self.eval_context_mut();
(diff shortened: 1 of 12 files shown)


RalfJung:
How is `miri_tree_nth_parent` less hacky than the previous approach? Previously at least this "nth parent" thing was only useful for naming pointers, now it lets you actually obtain that provenance in the code, which seems so much worse to me.

If the goal is just to get https://github.com/rust-lang/miri/pull/2936 to land, why not just use a different depth for the two variants of the test? `if cfg!(uniq) { 2 } else { 1 }` or something like that should do it.

Issue #2917, opened 2023-06-04
Indicate where an allocation was created when reporting a use-after-free


saethlin:
Use-after-free is detected in the core interpreter. How would Miri add its span information to the use-after-free diagnostics?

RalfJung:
We do something like that for `InvalidUninitBytes`, the same could work here?

https://github.com/rust-lang/miri/blob/9fa049f918fdac006a16f560f253f3ba07e415a2/src/diagnostics.rs#L379

RalfJung:
I think this got fixed by https://github.com/rust-lang/rust/pull/115184.

PR #2940, opened 2023-06-25
When reporting a heap use-after-free, say where the allocation was allocated and deallocated

saethlin:
This is a partial solution to: https://github.com/rust-lang/miri/issues/2917

Currently in the interpreter, we only have accurate information for where heap allocations are allocated and deallocated (see https://github.com/rust-lang/miri/pull/2940#discussion_r1243559711). So this just implements support for allocations where the information is already available, and the full support will require more interpreter tweaks.

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -304,11 +304,21 @@ pub fn report_error<'tcx, 'mir>(
                     (None, format!("this usually indicates that your program performed an invalid operation and caused Undefined Behavior")),
                     (None, format!("but due to `-Zmiri-symbolic-alignment-check`, alignment errors can also be false positives")),
                 ],
-            UndefinedBehavior(_) =>
-                vec![
+            UndefinedBehavior(info) => {
+                let mut helps = vec![
                     (None, format!("this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior")),
                     (None, format!("see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information")),
-                ],
+                ];
+                if let UndefinedBehaviorInfo::PointerUseAfterFree(alloc_id, _) = info {
+                    if let Some(span) = ecx.machine.allocated_span(*alloc_id) {
+                        helps.push((Some(span), format!("{:?} was allocated here:", alloc_id)));
+                    }
+                    if let Some(span) = ecx.machine.deallocated_span(*alloc_id) {
+                        helps.push((Some(span), format!("{:?} was deallocated here:", alloc_id)));
+                    }
+                }
+                helps
+            }
             InvalidProgram(
                 InvalidProgramInfo::AlreadyReported(_)
             ) => {
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -25,7 +25,7 @@ use rustc_middle::{
     },
 };
 use rustc_span::def_id::{CrateNum, DefId};
-use rustc_span::Symbol;
+use rustc_span::{Span, SpanData, Symbol};
 use rustc_target::abi::{Align, Size};
 use rustc_target::spec::abi::Abi;
 
@@ -131,6 +131,17 @@ impl MayLeak for MiriMemoryKind {
         match self {
             Rust | Miri | C | WinHeap | Runtime => false,
             Machine | Global | ExternStatic | Tls | Mmap => true,
+        }
+    }
+}
+
+impl MiriMemoryKind {
+    /// Whether we have a useful allocation span for an allocation of this kind.
+    fn should_save_allocation_span(self) -> bool {
+        use self::MiriMemoryKind::*;
+        match self {
+            Rust | Miri | C | Mmap => true,
+            Machine | Global | ExternStatic | Tls | WinHeap | Runtime => false,
         }
     }
 }
@@ -497,6 +508,10 @@ pub struct MiriMachine<'mir, 'tcx> {
 
     /// Whether to collect a backtrace when each allocation is created, just in case it leaks.
     pub(crate) collect_leak_backtraces: bool,
+
+    /// The spans we will use to report where an allocation was created and deallocated in
+    /// diagnostics.
+    pub(crate) allocation_spans: RefCell<FxHashMap<AllocId, (Option<Span>, Option<Span>)>>,
 }
 
 impl<'mir, 'tcx> MiriMachine<'mir, 'tcx> {
@@ -621,6 +636,7 @@ impl<'mir, 'tcx> MiriMachine<'mir, 'tcx> {
             stack_addr,
             stack_size,
             collect_leak_backtraces: config.collect_leak_backtraces,
+            allocation_spans: RefCell::new(FxHashMap::default()),
         }
     }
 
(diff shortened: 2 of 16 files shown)


RalfJung:
Wouldn't it make more sense to point to where the allocation was freed?

saethlin:
The primary response to the current diagnostic is "great, now what even is alloc1234" and this change answers that question. I agree that the next question is where the allocation is deallocated, I'll add that.

RalfJung:
I see, makes sense. I guess we can do that in a future PR.

However what would make sense for this PR is to also display the allocation span for out-of-bounds accesses.

src/diagnostics.rs
```diff
@@ -303,6 +303,12 @@ pub fn report_error<'tcx, 'mir>(
                     (None, format!("this usually indicates that your program performed an invalid operation and caused Undefined Behavior")),
                     (None, format!("but due to `-Zmiri-symbolic-alignment-check`, alignment errors can also be false positives")),
                 ],
+            UndefinedBehavior(UndefinedBehaviorInfo::PointerUseAfterFree(alloc_id)) =>
+                vec![
+                    (None, format!("this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior")),
+                    (None, format!("see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information")),
```
RalfJung, line 309: Could you avoid duplicating this part?

src/machine.rs
```diff
@@ -1046,6 +1054,9 @@ impl<'mir, 'tcx> Machine<'mir, 'tcx> for MiriMachine<'mir, 'tcx> {
             },
             |ptr| ecx.global_base_pointer(ptr),
         )?;
+
+        ecx.machine.allocation_spans.borrow_mut().insert(id, ecx.machine.current_span());
```
RalfJung, line 1058: I'm not sure this will give us the right span... that's not exactly what `adjust_allocation` is meant for. For global allocations (consts, statics) `adjust_allocation` is called lazily the first time they are accessed... I guess for your error this is fine since these allocations will never be freed, but it does mean `allocation_spans` has the wrong span for all of them.

Worse, for stack allocations I think it will also be wrong. With code like
```rust
let x = 42;
let ptr = &x;
```
the allocation will be created when `&x` is run; we avoid creating an allocation for locals of scalar type that never have their address taken. I think this can lead to wrong spans being shown in use-after-free messages for such stack-allocated variables.

tests/fail/concurrency/thread_local_static_dealloc.stderr
```diff
@@ -6,7 +6,12 @@ LL |         let _val = *dangling_ptr.0;
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
-   = note: BACKTRACE:
+help: ALLOC was allocated here:
+  --> $DIR/thread_local_static_dealloc.rs:LL:CC
+   |
+LL |         let dangling_ptr = std::thread::spawn(|| SendRaw(&TLS as *const u8)).join().unwrap();
```
RalfJung, line 12: This is a slightly suspicious span, though I guess for thread-locals it is unclear what the span *should* be.

saethlin: Yup. No idea what it should be for allocation _or_ deallocation.

src/machine.rs
saethlin: You're right, we get the wrong span in that case.

I'm not sure what to do about this. On the one hand, I think one could justify merging this because it produces exactly the same quality of output as the only other way to dig this information out of Miri. Currently my advice to users who want to debug a use-after-free is to use `-Zmiri-track-alloc-id`. So at least this PR would get the same wrong answer faster. :shrug: 

I'm also sure it's possible to fix this. If nothing else, we do have information to the right span in the interpreter and I could totally pass an `Option<Span>` down through `adjust_allocation`. That feels hacky but it would work.

RalfJung: I feel like we probably want to treat different kinds of allocations pretty differently:
- for heap allocations, keep a span pointing to where they were created
- for local variables, keep track of which local variable they correspond to (maybe point to the span of the `let x` or so?)
- for thread-local storage and globals (static/const), point to the declaration of the variable

You do have `kind` so what would be easy to do is to recognize heap allocations and just not record a span for anything else. For stack variables we'd probably need a new hook to be called from `force_allocation` to record the association between an AllocId and a particular local.

Global variables need their own special handling; they are probably also less important to cover. `thread_local_static_base_pointer` could store the association between thread-locals and AllocId, and `before_access_global` could do that for regular statics. For other globals this is a lot more tricky.

RalfJung:
@saethlin what are your plans for this PR?

An MVP of this might be to only track spans for heap allocations for now?

saethlin:
MVP is a good idea. I'll strip this down into a heap-only MVP then probably open an issue to permit this to work for all allocations.

src/machine.rs
```diff
@@ -135,6 +135,17 @@ impl MayLeak for MiriMemoryKind {
     }
 }
 
+impl MiriMemoryKind {
+    /// Whether we have a useful allocation span for an allocation of this kind.
+    fn should_save_allocation_span(self) -> bool {
+        use self::MiriMemoryKind::*;
+        match self {
+            Rust | Miri | C | Mmap => true,
+            Machine | Global | ExternStatic | Tls | WinHeap | Runtime => false,
```
RalfJung, line 144: Why is `WinHeap` treated different from the other heap allocations?

src/machine.rs
```diff
@@ -135,6 +135,17 @@ impl MayLeak for MiriMemoryKind {
     }
 }
 
+impl MiriMemoryKind {
+    /// Whether we have a useful allocation span for an allocation of this kind.
+    fn should_save_allocation_span(self) -> bool {
+        use self::MiriMemoryKind::*;
+        match self {
+            Rust | Miri | C | Mmap => true,
```
RalfJung, line 143:
```suggestion
            // Heap allocations are fine since the `Allocation` is created immediately.
            Rust | Miri | C | Mmap => true,
            // Everything else is unclear, let's not show potentially confusing spans.
```

src/machine.rs
```diff
@@ -1051,6 +1084,16 @@ impl<'mir, 'tcx> Machine<'mir, 'tcx> for MiriMachine<'mir, 'tcx> {
             },
             |ptr| ecx.global_base_pointer(ptr),
         )?;
+
+        if let MemoryKind::Machine(kind) = kind {
+            if kind.should_save_allocation_span() {
```
RalfJung, line 1089: `if matches!(kind, MemoryKind::Machine(kind) if kind.should_save_allocation_span())` could compress these into one off.

src/machine.rs
```diff
@@ -1051,6 +1084,16 @@ impl<'mir, 'tcx> Machine<'mir, 'tcx> for MiriMachine<'mir, 'tcx> {
...
+                ecx.machine
+                    .allocation_spans
+                    .borrow_mut()
+                    .insert(id, (Some(ecx.machine.current_span()), None));
```
RalfJung, line 1093: The first component seems to always be `Some`, so it doesn't have to be an `Option`.

RalfJung (commented):
Looks good, just some nits!

src/machine.rs
saethlin: I probably pulled it into the false list then never really noticed the mistake... do we have any tests that would show a use-after-free span for WinHeap? I don't think we actually do.

RalfJung: I don't think we have such a test, no.

RalfJung:
Thanks! A nice improvement. :)
@bors r+

PR #2941, opened 2023-06-26
Update ui test crate

oli-obk:
The main changes:

* we don't need to roll our own `status_emitter` anymore 
* there's only one way to filter now: with the closure passed to `run_tests_generic`

### DIFF
--- a/tests/compiletest.rs
+++ b/tests/compiletest.rs
@@ -3,7 +3,6 @@ use regex::bytes::Regex;
 use std::ffi::OsString;
 use std::path::{Path, PathBuf};
 use std::{env, process::Command};
-use ui_test::status_emitter::StatusEmitter;
 use ui_test::CommandBuilder;
 use ui_test::{color_eyre::Result, Config, Match, Mode, OutputConflictHandling};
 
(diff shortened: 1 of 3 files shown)


tests/compiletest.rs
```diff
@@ -113,39 +110,48 @@ fn test_config(target: &str, path: &str, mode: Mode, with_dependencies: bool) ->
...
-        }
-        match &**arg {
-            "--quiet" => {
-                config.quiet = true;
-                false
-            }
-            "--" => {
-                after_dashdash = true;
-                false
+    let mut quiet = false;
```
RalfJung, line 117: I can't find any place that this variable s actually used?

tests/compiletest.rs
```diff
@@ -113,39 +110,48 @@ fn test_config(target: &str, path: &str, mode: Mode, with_dependencies: bool) ->
...
+        |path| {
+            path.extension().is_some_and(|ext| ext == "rs")
+                && (filters.is_empty() || filters.iter().any(|f| path.starts_with(f)))
+        },
         // This could be used to overwrite the `Config` on a per-test basis.
         |_, _| None,
-        TextAndGha,
+        (
+            ui_test::status_emitter::Text,
+            ui_test::status_emitter::Gha::<false> { name: format!("{mode:?} {path} ({target})") },
```
RalfJung, line 153: What does the `false` here mean?

oli-obk: Turns off gha groups

RalfJung:
```suggestion
            ui_test::status_emitter::Gha::</* GHA groups */ false> { name: format!("{mode:?} {path} ({target})") },
```

tests/compiletest.rs
oli-obk: oh no rustc, why hast thou forsaken me

PR #2945, opened 2023-06-28
Try running a sync automatically

oli-obk:
absolutely no clue if this works, but it happens after the zulip message, so worst case we'll see the failure after the cron job did everything it currently does.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,13 +156,13 @@ jobs:
       - name: mark the job as a failure
         run: exit 1
 
-  # Send a Zulip notification when a cron job fails
   cron-fail-notify:
     name: cronjob failure notification
     runs-on: ubuntu-latest
     needs: [build, style]
     if: github.event_name == 'schedule' && (failure() || cancelled())
     steps:
+      # Send a Zulip notification
       - name: Install zulip-send
         run: pip3 install zulip
       - name: Send Zulip notification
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -185,3 +185,22 @@ jobs:
...
+          git config --global user.name 'The Miri Conjob Bot'
+          git config --global user.email 'miri@cron.bot'
+      - name: get changes from rustc
+        run: ./miri rustc-pull
+      - name: Push changes to a branch
+        run: |
+          git switch -c "rustup$(date -u +%Y-%m)"
+          git push
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v5
```
RalfJung, line 203: Looks like this can now be done without third-party actions: https://stackoverflow.com/a/73340290.

Also would be better to avoid computing the branch name twice.

.github/workflows/ci.yml
```diff
@@ -185,3 +185,22 @@ jobs:
...
+      - name: install josh-proxy
+        run: cargo +stable install josh-proxy --git https://github.com/josh-project/josh --tag r22.12.06
+      - name: start josh-proxy
+        run: josh-proxy --local=$HOME/.cache/josh --remote=https://github.com --no-background &
+      - name: setup bot git name and email
+        run: |
+          git config --global user.name 'The Miri Conjob Bot'
+          git config --global user.email 'miri@cron.bot'
+      - name: get changes from rustc
+        run: ./miri rustc-pull
```
RalfJung, line 197: Potential improvement (but we can leave that for later):
```
./miri toolchain
./miri fmt
git commit -am "rustfmt" || true # allowed to fail since there might be nothing to commit
```

.github/workflows/ci.yml
oli-obk: Ah yea, I just picked the first answer to that question and didn't read on :D

.github/workflows/ci.yml
oli-obk: Yea, tho I was thinking we could also try to make the regular CI job do that automatically if the fmt job fails. Like for every PR, not just on the sync PR.

RalfJung: That sounds fancy... not sure how I feel about auto-pushing to other people's branches, but it might be worth a shot.

.github/workflows/ci.yml
```diff
@@ -185,3 +185,26 @@ jobs:
...
+      - name: format changes (if any)
+        run: |
+          ./miri toolchain
+          ./miri fmt --check || ./miri fmt && git commit -am "fmt"
```
RalfJung, line 201:
```suggestion
          ./miri fmt --check || (./miri fmt && git commit -am "fmt")
```

.github/workflows/ci.yml
```diff
@@ -185,3 +185,26 @@ jobs:
           Sincerely,
           The Miri Cronjobs Bot' \
             --user $ZULIP_BOT_EMAIL --api-key $ZULIP_API_TOKEN --site https://rust-lang.zulipchat.com
+      - name: install josh-proxy
```
RalfJung, line 188:
```suggestion
      # Attempt to auto-sync with rustc
      - name: install josh-proxy
```

RalfJung (commented):
Just some final nits, r=me with those resolves. I am excited to see this in action!

.github/workflows/ci.yml
```diff
@@ -185,3 +185,27 @@ jobs:
...
+        run: josh-proxy --local=$HOME/.cache/josh --remote=https://github.com --no-background &
+      - name: setup bot git name and email
+        run: |
+          git config --global user.name 'The Miri Conjob Bot'
+          git config --global user.email 'miri@cron.bot'
+      - name: get changes from rustc
+        run: ./miri rustc-pull
+      - name: format changes (if any)
+        run: |
+          ./miri toolchain
```
RalfJung, line 201: Is rustup-toolchain-install-master installed here?

PR #2947, opened 2023-06-29
Try to fix cronjob PR creation

oli-obk:
The main tests use `bash` several times, maybe that is the issue here, too?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,10 +198,7 @@ jobs:
       - name: get changes from rustc
         run: ./miri rustc-pull
       - name: Install rustup-toolchain-install-master
-        if: ${{ steps.cache.outputs.cache-hit != 'true' }}
-        shell: bash
-        run: |
-          cargo install -f rustup-toolchain-install-master
+        run: cargo install -f rustup-toolchain-install-master
       - name: format changes (if any)
         run: |
           ./miri toolchain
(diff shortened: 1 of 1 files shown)


RalfJung:
Lol, we perfectly raced. ;) but https://github.com/rust-lang/miri/pull/2946 was a bit earlier.
@bors r-

.github/workflows/ci.yml
```diff
@@ -195,8 +195,10 @@ jobs:
           git config --global user.name 'The Miri Conjob Bot'
           git config --global user.email 'miri@cron.bot'
       - name: get changes from rustc
+        shell: bash
         run: ./miri rustc-pull
```
RalfJung, line 199: I think the issue is that the Miri repo is not checked out in this job.

Also, don't you have to install rustup-toolchain-install-master somewhere?

oli-obk: force pushed a change that should do everything needed

.github/workflows/ci.yml
```diff
@@ -196,6 +197,11 @@ jobs:
           git config --global user.email 'miri@cron.bot'
       - name: get changes from rustc
         run: ./miri rustc-pull
+      - name: Install rustup-toolchain-install-master
+        if: ${{ steps.cache.outputs.cache-hit != 'true' }}
+        shell: bash
+        run: |
+          cargo install -f rustup-toolchain-install-master
```
RalfJung, line 204:
```suggestion
        run: cargo install -f rustup-toolchain-install-master
```
we don't have a cache here so the rest doesn't do much (and no idea why we set `shell`...)

PR #2950, opened 2023-07-02
cargo-miri: better error message when RUSTC is not set

icewind1991:
Currently, when running `cargo-miri` instead of `cargo miri` you get a very confusing.

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: NotPresent', src/tools/miri/cargo-miri/src/main.rs:83:41
```

error. This replaces that with a message telling them to use `cargo miri`


RalfJung:
Thanks, that should certainly be better!

We will still do something very confusing if people happen to have set RUSTC in their environment, but that is less likely I guess.

@bors r+

PR #2956, opened 2023-07-03
Switch over to rustc's `tracing` crate instead of using our own `log` crate

oli-obk:
disadvantage: to turn on debug/trace logging, you need a rustc built with it
advantage: sharing our logging scheme with rustc, so our logs get nested correctly.

### DIFF
--- a/src/borrow_tracker/mod.rs
+++ b/src/borrow_tracker/mod.rs
@@ -2,7 +2,6 @@ use std::cell::RefCell;
 use std::fmt;
 use std::num::NonZeroU64;
 
-use log::trace;
 use smallvec::SmallVec;
 
 use rustc_data_structures::fx::{FxHashMap, FxHashSet};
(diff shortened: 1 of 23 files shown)


src/borrow_tracker/mod.rs
```diff
@@ -1,10 +1,8 @@
+use smallvec::SmallVec;
 use std::cell::RefCell;
 use std::fmt;
 use std::num::NonZeroU64;
```
RalfJung, line 4: Please keep `std` and crate imports separate, as before

RalfJung:
> disadvantage: to turn on debug/trace logging, you need a rustc built with it

Hm, that is unfortunate... but probably worth it to get the proper nesting.
Are there docs that need/should be updated for this?

RalfJung:
@oli-obk what is the status of this PR? Is it worth keeping it open?
If you don't plan to get back to it in the near future, IMO it'd be better to close (and then reopen if/when you have time again).

RalfJung:
Thanks. :)
@bors r+

PR #2969, opened 2023-07-06
Stop parsing ui_test annotations in `run-dep` mode

### LINKED ISSUE #2967: `./miri run-dep` parses ui-test comments
Title says it all: running `./miri run-dep` shouldn't care about whether the ui-test comments in the file make sense, they should just all be ignored. But currently when working on a test I am getting errors like

```
$ ./miri run-dep tests/fail/shims/memcpy_zero.rs
    Finished dev [optimized + debuginfo] target(s) in 0.01s
    Finished dev [optimized + debuginfo] target(s) in 0.01s
     Running `target/debug/cargo-miri miri setup --print-sysroot`
    Finished test [optimized + debuginfo] target(s) in 0.02s
     Running tests/compiletest.rs (target/debug/deps/compiletest-abb37bc11260f53d)
   Building test dependencies...
Error: 
   0: [
          InvalidComment {
              msg: "no pattern specified",
              line: 9,
          },
      ]

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
error: test failed, to rerun pass `--test compiletest`
```

oli-obk:
fixes #2967

### DIFF
--- a/tests/compiletest.rs
+++ b/tests/compiletest.rs
@@ -277,11 +277,6 @@ fn run_dep_mode(target: String, mut args: impl Iterator<Item = OsString>) -> Res
     config.program.args.clear(); // We want to give the user full control over flags
     config.build_dependencies_and_link_them()?;
 
-    if let Ok(extra_flags) = env::var("MIRIFLAGS") {
-        for flag in extra_flags.split_whitespace() {
-            config.program.args.push(flag.into());
-        }
-    }
     let mut cmd = config.program.build(&config.out_dir);
 
     cmd.arg(path);
--- a/tests/compiletest.rs
+++ b/tests/compiletest.rs
@@ -274,13 +274,20 @@ fn main() -> Result<()> {
 fn run_dep_mode(target: String, mut args: impl Iterator<Item = OsString>) -> Result<()> {
     let path = args.next().expect("./miri run-dep must be followed by a file name");
     let mut config = test_config(&target, "", Mode::Yolo, /* with dependencies */ true);
-    config.program.args.remove(0); // remove the `--error-format=json` argument
-    config.program.args.push("--color".into());
-    config.program.args.push("always".into());
-    let mut cmd = ui_test::test_command(config, Path::new(&path))?;
-    // Separate the arguments to the `cargo miri` invocation from
-    // the arguments to the interpreted prog
-    cmd.arg("--");
+    config.program.args.clear(); // We want to give the user full control over flags
+    config.build_dependencies_and_link_them()?;
+
+    if let Ok(extra_flags) = env::var("MIRIFLAGS") {
+        for flag in extra_flags.split_whitespace() {
+            config.program.args.push(flag.into());
+        }
+    }
+    let mut cmd = config.program.build(&config.out_dir);
+
+    cmd.arg("--color").arg("always");
+    cmd.arg(path);
+
     cmd.args(args);
+    println!("Running {cmd:?}");
     if cmd.spawn()?.wait()?.success() { Ok(()) } else { std::process::exit(1) }
 }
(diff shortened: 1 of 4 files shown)


tests/compiletest.rs
```diff
@@ -274,10 +274,16 @@ fn main() -> Result<()> {
...
-    config.program.args.push("always".into());
-    let mut cmd = ui_test::test_command(config, Path::new(&path))?;
+    let mut cmd = config.program.build(&config.out_dir);
+
+    cmd.arg("--color").arg("always");
+    cmd.arg(path);
+
+    if let Some(edition) = config.edition {
+        cmd.arg("--edition").arg(edition);
+    }
```
RalfJung, line 286: Does this mean `./miri run-dep`  defaults to edition 2021? Should we make `./miri run` also add `--edition 2021`?

oli-obk: probably 😆 Seems like a good default to have

oli-obk: But it's then hard to overwrite... so maybe let's just remove it from run-dep

RalfJung: Will `./miri run-dep file.rs --edition 2021` work?

oli-obk: It works now 😁 I tested that the behaviour around `--` is now the same as with `./miri run`.

tests/compiletest.rs
```diff
@@ -274,13 +274,20 @@ fn main() -> Result<()> {
...
+            config.program.args.push(flag.into());
+        }
+    }
+    let mut cmd = config.program.build(&config.out_dir);
+
+    cmd.arg("--color").arg("always");
+    cmd.arg(path);
+
     cmd.args(args);
+    println!("Running {cmd:?}");
```
oli-obk, line 291: I re-added this, as this is what `./miri run` also prints

RalfJung: Well, for `./miri run` it doesn't contain 3 lines of linker junk though.^^

tests/compiletest.rs
```diff
@@ -274,13 +274,20 @@ fn main() -> Result<()> {
...
-    // the arguments to the interpreted prog
-    cmd.arg("--");
+    config.program.args.clear(); // We want to give the user full control over flags
+    config.build_dependencies_and_link_them()?;
+
+    if let Ok(extra_flags) = env::var("MIRIFLAGS") {
+        for flag in extra_flags.split_whitespace() {
+            config.program.args.push(flag.into());
+        }
+    }
```
RalfJung, line 284: I'd rather have this done in `./miri` than here (then it is nicely in the same place for both run and run-dep)

tests/compiletest.rs
```diff
@@ -274,13 +274,20 @@ fn main() -> Result<()> {
...
+    let mut cmd = config.program.build(&config.out_dir);
+
+    cmd.arg("--color").arg("always");
```
RalfJung, line 287: Do we still need this?

tests/compiletest.rs
oli-obk: I could print before adding that 😆

RalfJung: `./miri run` prints only because that is forced behavior by cargo, I don't think we want to reproduce that here... maybe just add a `-q` in `./miri run`?

tests/compiletest.rs
oli-obk: it was already done 😆 I ended up doing it twice

tests/compiletest.rs
oli-obk: Did that for both `run` and `run-dep`

RalfJung:
Looks good, thanks!
@bors r+

RalfJung:
Output is not fully identical yet:
```rust
$ ./miri run-dep tests/pass/hello.rs
   Building test dependencies...
Hello, world!
```
Should we hide that as well?

oli-obk:
Yea, but that has to happen on the ui_test side. I'm working on simplifying the API (now that I got a bunch of users of it), so that `--quiet` and friends are automatically handled and affect these kind of outputs

PR #2972, opened 2023-07-09
C `mem` function shims: consistently treat "invalid" pointers as UB

RalfJung:
Depends on https://github.com/rust-lang/rust/pull/113435.


RalfJung:
https://github.com/rust-lang/rust/pull/114382 has landed, so this should be good to go.
@bors r+

PR #2975, opened 2023-07-11
remove compile-flags that are no longer needed

RalfJung:
We stopped running fail-tests with optimizations a while ago, so we don't need to explicitly set the opt-level to 0 any more.


RalfJung:
Let's land https://github.com/rust-lang/miri/pull/2977 first.
@bor r-

PR #2977, opened 2023-07-12
Rustup

RalfJung:


RalfJung:
What the heck? Some tree borrows tests failing but only on Linux CI and I cannot reproduce locally...

RalfJung:
Ah, we are probably GC'ing the protected tags, removing the protection as a side-effect.

But why does this only affect Tree Borrows? Cc @Vanille-N

RalfJung:
Oh I think I know why SB was not affected... We only GC stacks that are st least 64 elements large, so this example never gets there.

PR #2985, opened 2023-07-17
make full field retagging the default

RalfJung:
The 'scalar' field retagging mode is clearly a hack -- it mirrors details of the codegen backend and how various structs are represented in LLVM. This means whether code has UB or not depends on surprising aspects, such as whether a struct has 2 or 3 (non-zero-sized) fields. Now that both hashbrown and scopeguard have released fixes to be compatible with field retagging, I think it is time to enable full field retagging by default.

@saethlin do you have an idea of how much fallout enabling full field retagging by default will cause? Do you have objections to enabling it by default?

Fixes https://github.com/rust-lang/miri/issues/2528

### DIFF
--- a/src/borrow_tracker/stacked_borrows/diagnostics.rs
+++ b/src/borrow_tracker/stacked_borrows/diagnostics.rs
@@ -61,12 +61,15 @@ struct Invalidation {
 #[derive(Clone, Debug)]
 enum InvalidationCause {
     Access(AccessKind),
-    Retag(Permission, RetagCause),
+    Retag(Permission, RetagInfo),
 }
 
 impl Invalidation {
     fn generate_diagnostic(&self) -> (String, SpanData) {
-        let message = if let InvalidationCause::Retag(_, RetagCause::FnEntry) = self.cause {
+        let message = if matches!(
+            self.cause,
+            InvalidationCause::Retag(_, RetagInfo { cause: RetagCause::FnEntry, .. })
+        ) {
             // For a FnEntry retag, our Span points at the caller.
             // See `DiagnosticCx::log_invalidation`.
             format!(
@@ -87,8 +90,8 @@ impl fmt::Display for InvalidationCause {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             InvalidationCause::Access(kind) => write!(f, "{kind}"),
-            InvalidationCause::Retag(perm, kind) =>
-                write!(f, "{perm:?} {retag}", retag = kind.summary()),
+            InvalidationCause::Retag(perm, info) =>
+                write!(f, "{perm:?} {retag}", retag = info.summary()),
         }
     }
 }
@@ -129,13 +132,13 @@ impl<'ecx, 'mir, 'tcx> DiagnosticCxBuilder<'ecx, 'mir, 'tcx> {
 
     pub fn retag(
         machine: &'ecx MiriMachine<'mir, 'tcx>,
-        cause: RetagCause,
+        info: RetagInfo,
         new_tag: BorTag,
         orig_tag: ProvenanceExtra,
         range: AllocRange,
     ) -> Self {
         let operation =
-            Operation::Retag(RetagOp { cause, new_tag, orig_tag, range, permission: None });
+            Operation::Retag(RetagOp { info, new_tag, orig_tag, range, permission: None });
 
         DiagnosticCxBuilder { machine, operation }
     }
@@ -179,11 +182,17 @@ enum Operation {
 
 #[derive(Debug, Clone)]
 struct RetagOp {
-    cause: RetagCause,
+    info: RetagInfo,
     new_tag: BorTag,
     orig_tag: ProvenanceExtra,
     range: AllocRange,
     permission: Option<Permission>,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct RetagInfo {
+    pub cause: RetagCause,
+    pub in_field: bool,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq)]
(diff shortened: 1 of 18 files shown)


saethlin:
I have quality-of-implementation sort of concerns, but I don't think anything that would make me object to this? I do not have a concrete sense of the fallout. I suspect this will primarily make self-referential things explode that didn't before because the self-referential struct was too large to be Scalar/ScalarPair.

I agree with the zero-cost abstraction argument for why we should have full field retagging. But at the same time I'm not excited about sanitizing against UB that a user can prove we do not optimize on. I'd be a lot happier if we relied on full field retagging for optimization(s) or at least had a concrete plan to do so. It would be a real shame if there is some compiler design reason that this optimization may never be worthwhile.

Field retagging is pretty subtle and I'm not sure our diagnostics do a good job of explaining it. When the default changes, some people's code will suddenly start reporting UB where it didn't before. It would be really good to add a note in our diagnostics that a retag was caused by field retagging. For Stack Borrows we already have `RetagCause` which carries this information, we'd just need a new variant I think. I'm not entirely sure how I'd slot this into Tree Borrows.

RalfJung:
> I'm not excited about sanitizing against UB that a user can prove we do not optimize on. I'd be a lot happier if we relied on full field retagging for optimization(s) or at least had a concrete plan to do so. It would be a real shame if there is some compiler design reason that this UB may never be useful.

We have a ton of UB of that sort. In fact we are actively avoiding exploiting e.g. Stacked Borrows (the parts that go beyond `noalias`) for optimizations. We'll usually want UB reported in Miri before exploiting it in optimizations to give people at least some chance of being warned about it.

So, I don't think "the UB should be exploited by the compiler" is a standard we can reasonably satisfy.

What I am more concerned about is that some situations where this UB arises don't have a good alternative. They would need something like https://github.com/rust-lang/rfcs/pull/3336. I'm not sure how to make progress on that RFC. Maybe we'll get some more motivating examples when Miri starts reporting more UB for self-referential code...

> It would be really good to add a note in our diagnostics that a retag was caused by field retagging. For Stack Borrows we already have RetagCause which carries this information, we'd just need a new variant I think. I'm not entirely sure how I'd slot this into Tree Borrows.

That is a good point. I am less concerned about Tree Borrows since it is opt-in anyway.

saethlin:
> Maybe we'll get some more motivating examples when Miri starts reporting more UB for self-referential code...

Hm, maybe we should amend our output to encourage people to reach out to us if they find the UB report troubling. Currently, we do not.

RalfJung:
> It would be really good to add a note in our diagnostics that a retag was caused by field retagging. 

Done.

> Hm, maybe we should amend our output to encourage people to reach out to us if they find the UB report troubling. Currently, we do not.

I added a message along those lines for all SB UB. Or did you mean only for field retagging? Or all UB? I'm a bit worried about being overwhelmed, we can't offer a general "please help my code is UB" hotline.^^

src/diagnostics.rs
```diff
@@ -224,11 +224,11 @@ pub fn report_error<'tcx, 'mir>(
                     (None, format!("or pass `-Zmiri-isolation-error=warn` to configure Miri to return an error code from isolated operations (if supported for that operation) and continue with a warning")),
                 ],
             StackedBorrowsUb { help, history, .. } => {
-                let url = "https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md";
                 msg.extend(help.clone());
                 let mut helps = vec![
                     (None, format!("this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental")),
-                    (None, format!("see {url} for further information")),
+                    (None, format!("see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information")),
+                    (None, format!("please reach out to us (e.g. at <https://rust-lang.zulipchat.com/#narrow/stream/269128-miri>) if you find this troubling")),
```
saethlin, line 231: Do the `< >` have any particular meaning here? I can only think they'd make it harder to copy this into a URL bar.

RalfJung: They are commonly used in email to mark URLs, so that it is easy to tell where it begins and ends and to deal with linebreaks that have been introduced.

In this case there is a `)` after the URL and I was not sure if that woukd split correctly in all terminal emulators.

saethlin: That makes sense.

saethlin:
> I added a message along those lines for all SB UB. Or did you mean only for field retagging? Or all UB? I'm a bit worried about being overwhelmed, we can't offer a general "please help my code is UB" hotline.^^

At the time I was thinking of field retagging, because that's where you said we'd it would be good to have motivating examples. I think for Stacked Borrows generally, we already have a lot of motivating examples? So outside of field retagging the message might generate more noise than signal.

RalfJung:
Okay I think I managed to get that. It's a bit hard to see through the error reporting layers.^^

src/borrow_tracker/stacked_borrows/diagnostics.rs
```diff
@@ -372,9 +389,13 @@ impl<'history, 'ecx, 'mir, 'tcx> DiagnosticCx<'history, 'ecx, 'mir, 'tcx> {
             self.history.id,
             self.offset.bytes(),
         );
+        let mut helps = vec![operation_summary(&op.info.summary(), self.history.id, op.range)];
+        if op.info.in_field {
+            helps.push(format!("errors for retagging in fields are fairly new; please reach out to us (e.g. at <https://rust-lang.zulipchat.com/#narrow/stream/269128-miri>) if you find this troubling"));
```
saethlin, line 394:
```suggestion
            helps.push(format!("errors for retagging in fields are fairly new; please reach out to us (e.g. at <https://rust-lang.zulipchat.com/#narrow/stream/269128-miri>) if you find this error troubling"));
```

saethlin (commented):
r=me with English nit

lopopolo:
Hi folks, I wanted to reach out here and say thank you for the care that went into this. My weekly Miri CI failed as a result of this change and I was largely able to diagnose and fix the issues the new sanitizer caught on my own.

If you're curious:

- https://github.com/artichoke/intaglio/issues/235
- https://github.com/artichoke/intaglio/pull/236
- https://rustsec.org/advisories/RUSTSEC-2023-0048.html

PR #2989, opened 2023-07-22
miri: implement some `llvm.x86.sse.*` intrinsics and add tests

eduardosm:
PR moved from https://github.com/rust-lang/rust/pull/113932.

Implements LLVM intrisics needed to run most SSE functions from `core::arch::x86{,_64}`.

Also adds miri tests for those functions (mostly copied from core_arch tests).

r? @RalfJung 

The first commit is the same that the commit in the PR I had opened in the Rust repository. I addressed review comments in additional commits to make it easier to review. I also fixed formatting and clippy warnings.

### DIFF
--- a/src/shims/foreign_items.rs
+++ b/src/shims/foreign_items.rs
@@ -918,6 +918,33 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 this.write_scalar(Scalar::from_f64(res), dest)?;
             }
 
+            "llvm.prefetch" => {
+                let [p, rw, loc, ty] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let _ = this.read_pointer(p)?;
+                let rw = this.read_scalar(rw)?.to_i32()?;
+                let loc = this.read_scalar(loc)?.to_i32()?;
+                let ty = this.read_scalar(ty)?.to_i32()?;
+
+                if ty == 1 {
+                    // Data cache prefetch.
+                    // Notably, we do not have to check the pointer, this operation is never UB!
+
+                    if !matches!(rw, 0 | 1) {
+                        throw_unsup_format!("invalid `rw` value passed to `llvm.prefetch`: {}", rw);
+                    }
+                    if !matches!(loc, 0..=3) {
+                        throw_unsup_format!(
+                            "invalid `loc` value passed to `llvm.prefetch`: {}",
+                            loc
+                        );
+                    }
+                } else {
+                    throw_unsup_format!("unsupported `llvm.prefetch` type argument: {}", ty);
+                }
+            }
+
             // Architecture-specific shims
             "llvm.x86.addcarry.64" if this.tcx.sess.target.arch == "x86_64" => {
                 // Computes u8+u64+u64, returning tuple (u8,u64) comprising the output carry and truncated sum.
@@ -968,6 +995,12 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                         throw_unsup_format!("unsupported llvm.arm.hint argument {}", arg);
                     }
                 }
+            }
+
+            name if name.starts_with("llvm.x86.sse.") => {
+                return shims::x86::sse::EvalContextExt::emulate_x86_sse_intrinsic(
+                    this, link_name, abi, args, dest,
+                );
             }
 
             // Platform-specific shims
--- a/src/shims/x86/sse.rs
+++ b/src/shims/x86/sse.rs
@@ -1,0 +1,609 @@
+use rustc_apfloat::{ieee::Single, Float as _};
+use rustc_middle::mir;
+use rustc_span::Symbol;
+use rustc_target::spec::abi::Abi;
+
+use rand::Rng as _;
+
+use crate::*;
+use shims::foreign_items::EmulateByNameResult;
+
+impl<'mir, 'tcx: 'mir> EvalContextExt<'mir, 'tcx> for crate::MiriInterpCx<'mir, 'tcx> {}
+pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
+    fn emulate_x86_sse_intrinsic(
+        &mut self,
+        link_name: Symbol,
+        abi: Abi,
+        args: &[OpTy<'tcx, Provenance>],
+        dest: &PlaceTy<'tcx, Provenance>,
+    ) -> InterpResult<'tcx, EmulateByNameResult<'mir, 'tcx>> {
+        let this = self.eval_context_mut();
+        // Prefix should have already been checked.
+        let unprefixed_name = link_name.as_str().strip_prefix("llvm.x86.sse.").unwrap();
+        // All these intrinsics operate on 128-bit (f32x4) SIMD vectors unless stated otherwise.
+        // Many intrinsic names are sufixed with "ps" (packed single) or "ss" (scalar single),
+        // where single means single precision floating point (f32). "ps" means thet the operation
+        // is performed on each element of the vector, while "ss" means that the operation is
+        // performed only on the first element, copying the remaining elements from the input
+        // vector (for binary operations, from the left-hand side).
+        match unprefixed_name {
+            // Used to implement _mm_{add,sub,mul,div,min,max}_ss functions.
+            // Performs the operations on the first component of `left` and
+            // `right` and copies the remaining components from `left`.
+            "add.ss" | "sub.ss" | "mul.ss" | "div.ss" | "min.ss" | "max.ss" => {
+                let [left, right] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let which = match unprefixed_name {
+                    "add.ss" => FloatBinOp::Arith(mir::BinOp::Add),
+                    "sub.ss" => FloatBinOp::Arith(mir::BinOp::Sub),
+                    "mul.ss" => FloatBinOp::Arith(mir::BinOp::Mul),
+                    "div.ss" => FloatBinOp::Arith(mir::BinOp::Div),
+                    "min.ss" => FloatBinOp::Min,
+                    "max.ss" => FloatBinOp::Max,
+                    _ => unreachable!(),
+                };
+
+                bin_op_ss(this, which, left, right, dest)?;
+            }
+            // Used to implement _mm_min_ps and _mm_max_ps functions.
+            // Note that the semantics are a bit different from Rust simd_min
+            // and simd_max intrinsics regarding handling of NaN and -0.0: Rust
+            // matches the IEEE min/max operations, while x86 has different
+            // semantics.
+            "min.ps" | "max.ps" => {
+                let [left, right] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let which = match unprefixed_name {
+                    "min.ps" => FloatBinOp::Min,
+                    "max.ps" => FloatBinOp::Max,
+                    _ => unreachable!(),
+                };
+
+                bin_op_ps(this, which, left, right, dest)?;
+            }
+            // Used to implement _mm_{sqrt,rcp,rsqrt}_ss functions.
+            // Performs the operations on the first component of `op` and
+            // copies the remaining components from `op`.
+            "sqrt.ss" | "rcp.ss" | "rsqrt.ss" => {
+                let [op] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let which = match unprefixed_name {
+                    "sqrt.ss" => FloatUnaryOp::Sqrt,
+                    "rcp.ss" => FloatUnaryOp::Rcp,
+                    "rsqrt.ss" => FloatUnaryOp::Rsqrt,
+                    _ => unreachable!(),
+                };
+
+                unary_op_ss(this, which, op, dest)?;
+            }
+            // Used to implement _mm_{sqrt,rcp,rsqrt}_ss functions.
+            // Performs the operations on all components of `op`.
+            "sqrt.ps" | "rcp.ps" | "rsqrt.ps" => {
+                let [op] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let which = match unprefixed_name {
+                    "sqrt.ps" => FloatUnaryOp::Sqrt,
+                    "rcp.ps" => FloatUnaryOp::Rcp,
+                    "rsqrt.ps" => FloatUnaryOp::Rsqrt,
+                    _ => unreachable!(),
+                };
+
+                unary_op_ps(this, which, op, dest)?;
+            }
+            // Used to implement the _mm_cmp_ss function.
+            // Performs a comparison operation on the first component of `left`
+            // and `right`, returning 0 if false or `u32::MAX` if true. The remaining
+            // components are copied from `left`.
+            "cmp.ss" => {
+                let [left, right, imm] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let which = match this.read_scalar(imm)?.to_i8()? {
+                    0 => FloatBinOp::Cmp(FloatCmpOp::Eq),
+                    1 => FloatBinOp::Cmp(FloatCmpOp::Lt),
+                    2 => FloatBinOp::Cmp(FloatCmpOp::Le),
+                    3 => FloatBinOp::Cmp(FloatCmpOp::Unord),
+                    4 => FloatBinOp::Cmp(FloatCmpOp::Neq),
+                    5 => FloatBinOp::Cmp(FloatCmpOp::Nlt),
+                    6 => FloatBinOp::Cmp(FloatCmpOp::Nle),
+                    7 => FloatBinOp::Cmp(FloatCmpOp::Ord),
+                    imm => {
+                        throw_unsup_format!(
+                            "invalid 3rd parameter of llvm.x86.sse.cmp.ps: {}",
+                            imm
+                        );
+                    }
+                };
+
+                bin_op_ss(this, which, left, right, dest)?;
+            }
+            // Used to implement the _mm_cmp_ps function.
+            // Performs a comparison operation on each component of `left`
+            // and `right`. For each component, returns 0 if false or u32::MAX
+            // if true.
+            "cmp.ps" => {
+                let [left, right, imm] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let which = match this.read_scalar(imm)?.to_i8()? {
+                    0 => FloatBinOp::Cmp(FloatCmpOp::Eq),
+                    1 => FloatBinOp::Cmp(FloatCmpOp::Lt),
+                    2 => FloatBinOp::Cmp(FloatCmpOp::Le),
+                    3 => FloatBinOp::Cmp(FloatCmpOp::Unord),
+                    4 => FloatBinOp::Cmp(FloatCmpOp::Neq),
+                    5 => FloatBinOp::Cmp(FloatCmpOp::Nlt),
+                    6 => FloatBinOp::Cmp(FloatCmpOp::Nle),
+                    7 => FloatBinOp::Cmp(FloatCmpOp::Ord),
+                    imm => {
+                        throw_unsup_format!(
+                            "invalid 3rd parameter of llvm.x86.sse.cmp.ps: {}",
+                            imm
+                        );
+                    }
+                };
+
+                bin_op_ps(this, which, left, right, dest)?;
+            }
+            // Used to implement _mm_{,u}comi{eq,lt,le,gt,ge,neq}_ps functions.
+            // Compares the first component of `left` and `right` and returns
+            // a scalar value (0 or 1).
+            "comieq.ss" | "comilt.ss" | "comile.ss" | "comigt.ss" | "comige.ss" | "comineq.ss"
+            | "ucomieq.ss" | "ucomilt.ss" | "ucomile.ss" | "ucomigt.ss" | "ucomige.ss"
+            | "ucomineq.ss" => {
+                let [left, right] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let (left, left_len) = this.operand_to_simd(left)?;
+                let (right, right_len) = this.operand_to_simd(right)?;
+
+                assert_eq!(left_len, right_len);
+
+                let left = this.read_scalar(&this.project_index(&left, 0)?)?.to_f32()?;
+                let right = this.read_scalar(&this.project_index(&right, 0)?)?.to_f32()?;
+                // The difference between the com* and *ucom variants is signaling
+                // of exceptions when either argument is a quiet NaN. We do not
+                // support accessing the SSE status register from miri (or from Rust,
+                // for that matter), so we treat equally both variants.
+                let res = match unprefixed_name {
+                    "comieq.ss" | "ucomieq.ss" => left == right,
+                    "comilt.ss" | "ucomilt.ss" => left < right,
+                    "comile.ss" | "ucomile.ss" => left <= right,
+                    "comigt.ss" | "ucomigt.ss" => left > right,
+                    "comige.ss" | "ucomige.ss" => left >= right,
+                    "comineq.ss" | "ucomineq.ss" => left != right,
+                    _ => unreachable!(),
+                };
+                this.write_scalar(Scalar::from_i32(i32::from(res)), dest)?;
+            }
+            // Use to implement _mm_cvtss_si32 and _mm_cvttss_si32.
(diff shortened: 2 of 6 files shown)


src/shims/x86/sse.rs
```diff
@@ -0,0 +1,568 @@
...
+                    "cvttss2si" => rustc_apfloat::Round::TowardZero,
+                    _ => unreachable!(),
+                };
+
+                let mut exact = false;
+                let cvt = op.to_i128_r(32, rnd, &mut exact);
+                let res = if cvt.status.intersects(
+                    rustc_apfloat::Status::INVALID_OP
+                        | rustc_apfloat::Status::OVERFLOW
+                        | rustc_apfloat::Status::UNDERFLOW,
```
RalfJung, line 188: So this ignores all other status flags -- is that correct? (Elsewhere in Miri we've used `res.status.is_empty()` on float2int casts)

src/shims/x86/sse.rs
```diff
@@ -0,0 +1,568 @@
...
+/// <https://www.felixcloutier.com/x86/cmpps>
+#[derive(Copy, Clone)]
+enum FloatCmpOp {
+    Eq,
+    Lt,
+    Le,
+    Unord,
+    Neq,
+    Nlt,
+    Nle,
```
RalfJung, line 321:
```suggestion
    /// Not less-than
    Nlt,
    /// Not less-or-equal
    Nle,
```

src/shims/x86/sse.rs
```diff
@@ -0,0 +1,568 @@
...
+    Ord,
```
RalfJung, line 322:
```suggestion
    /// Ordered, i.e. neither of them is NaN
    Ord,
```

src/shims/x86/sse.rs
```diff
@@ -0,0 +1,568 @@
...
+                FloatCmpOp::Nle => !matches!(left.partial_cmp(&right), Some(Equal | Less)),
+                FloatCmpOp::Ord => !left.is_nan() && !right.is_nan(),
+            };
+            Ok(Scalar::from_u32(if res { u32::MAX } else { 0 }))
+        }
+        FloatBinOp::Min => {
+            let left = left.to_scalar().to_f32()?;
+            let right = right.to_scalar().to_f32()?;
+            // SSE semantics to handle zero and NaN
+            if (left == Single::ZERO && right == Single::ZERO)
```
RalfJung, line 377: So the `==` here is a comparison that threads `-0` and `0` as equal? If so, please note this in a comment, as on first sight this is confusing. (Same for `Max` below.)

src/shims/x86/sse.rs
```diff
@@ -0,0 +1,568 @@
...
+        FloatUnaryOp::Sqrt => {
+            let op = op.to_scalar();
+            // FIXME using host floats
+            Ok(Scalar::from_u32(f32::from_bits(op.to_u32()?).sqrt().to_bits()))
+        }
+        FloatUnaryOp::Rcp => {
+            let op = op.to_scalar();
+            // Use 1.0001 as a crude way of simulating inaccuracy of rcp
+            // 1.0001 / x ~= 1 / x
+            let one = Single::from_bits(1.0001f32.to_bits().into());
```
RalfJung, line 501: What is going on here?

RalfJung (commented):
This looks very good, thanks a lot for all the comments. :) I left a bunch of notes, mostly minor.

I can't actually tell if these implementations are correct, I'll just have to trust you on that and wait for bug reports otherwise. ;)

One thing that might be useful is to briefly establish some terminology at the top of sse.rs. In particular "ss" and "ps" seem to be very common and I have no idea what these are for.

src/shims/x86/sse.rs
```diff
@@ -0,0 +1,568 @@
...
+                    "min.ss" => FloatBinOp::Min,
+                    "max.ss" => FloatBinOp::Max,
+                    _ => unreachable!(),
+                };
+
+                bin_op_ss(this, which, left, right, dest)?;
+            }
+            // Used to implement _mm_min_ps and _mm_max_ps functions.
+            // Note that the semantics are a bit different from Rust simd_min
+            // and simd_max intrinsics regarding handling of -0.0.
```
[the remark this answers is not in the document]
eduardosm: It is :)

src/shims/x86/sse.rs
```diff
@@ -0,0 +1,568 @@
...
+            let right = right.to_scalar().to_f32()?;
+            // FIXME: Make sure that these operations match the semantics of cmpps
+            use std::cmp::Ordering::*;
+            let res = match which {
+                FloatCmpOp::Eq => left == right,
+                FloatCmpOp::Lt => left < right,
+                FloatCmpOp::Le => left <= right,
+                FloatCmpOp::Unord => left.is_nan() || right.is_nan(),
+                FloatCmpOp::Neq => left != right,
+                FloatCmpOp::Nlt => !matches!(left.partial_cmp(&right), Some(Less)),
```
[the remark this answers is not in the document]
eduardosm: Because clippy complains.

https://rust-lang.github.io/rust-clippy/master/index.html#/neg_cmp_op_on_partial_ord

tests/pass/intrinsics-x86-sse.rs
```diff
@@ -0,0 +1,1082 @@
...
+        let a = _mm_setr_ps(-1.0, 5.0, 0.0, -10.0);
+        let b = _mm_setr_ps(-100.0, 20.0, 0.0, -5.0);
+        let r = _mm_min_ps(a, b);
+        assert_eq_m128(r, _mm_setr_ps(-100.0, 5.0, 0.0, -10.0));
+
+        // `_mm_min_ps` can **not** be implemented using the `simd_min` rust intrinsic. `simd_min`
+        // is lowered by the llvm codegen backend to `llvm.minnum.v*` llvm intrinsic. This intrinsic
+        // doesn't specify how -0.0 is handled. Unfortunately it happens to behave different from
+        // the `minps` x86 instruction on x86. The `llvm.minnum.v*` llvm intrinsic equals
+        // `r1` to `a` and `r2` to `b`.
```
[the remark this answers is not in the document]
eduardosm: I mostly copied these tests (and their comments) from core_arch. Shall I replace the comment with a more high-level explanation?

src/shims/x86/sse.rs
```diff
@@ -0,0 +1,568 @@
...
+                    0 => FloatBinOp::Cmp(FloatCmpOp::Eq),
+                    1 => FloatBinOp::Cmp(FloatCmpOp::Lt),
+                    2 => FloatBinOp::Cmp(FloatCmpOp::Le),
+                    3 => FloatBinOp::Cmp(FloatCmpOp::Unord),
+                    4 => FloatBinOp::Cmp(FloatCmpOp::Neq),
+                    5 => FloatBinOp::Cmp(FloatCmpOp::Nlt),
+                    6 => FloatBinOp::Cmp(FloatCmpOp::Nle),
+                    7 => FloatBinOp::Cmp(FloatCmpOp::Ord),
+                    imm => {
+                        throw_ub_format!("invalid 3rd parameter of llvm.x86.sse.cmp.ps: {}", imm);
```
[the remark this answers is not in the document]
eduardosm: I guess it is just unsupported/invalid

src/shims/x86/sse.rs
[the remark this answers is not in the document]
RalfJung: Then please allow the clippy lint.

tests/pass/intrinsics-x86-sse.rs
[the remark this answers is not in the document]
RalfJung: yeah, this should talk about how the x86 intrinsic is different from the simd intrinsic because the latter uses IEEE semantics.

I hope there are tests covering the differences?

src/shims/x86/sse.rs
```diff
@@ -0,0 +1,568 @@
...
+                let (left, left_len) = this.operand_to_simd(left)?;
+                let (right, right_len) = this.operand_to_simd(right)?;
+
+                assert_eq!(left_len, right_len);
+
+                let left = this.read_scalar(&this.project_index(&left, 0)?)?.to_f32()?;
+                let right = this.read_scalar(&this.project_index(&right, 0)?)?.to_f32()?;
+                // The difference between the com* and *ucom variants is signaling
+                // of exceptions when either argument is a quiet NaN. We do not
+                // support accessing the SSE status register from miri, so we treat
```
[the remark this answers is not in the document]
RalfJung: Lol, yeah using that makes no guarantees at all. LLVM will happily reorganize floating point operations with no regard for anyone potentially looking at the FP status register.

src/shims/foreign_items.rs
```diff
@@ -918,6 +918,29 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                let [p, rw, loc, ty] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let _ = this.read_pointer(p)?;
+                let rw = this.read_scalar(rw)?.to_i32()?;
+                let loc = this.read_scalar(loc)?.to_i32()?;
+                let ty = this.read_scalar(ty)?.to_i32()?;
+
+                if ty == 1 {
+                    // Data cache prefetch
```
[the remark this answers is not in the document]
eduardosm: Done

src/shims/foreign_items.rs
```diff
@@ -918,6 +918,29 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+
+                    if !matches!(rw, 0 | 1) {
+                        throw_ub_format!("invalid `rw` value passed to `llvm.prefetch`: {}", rw);
```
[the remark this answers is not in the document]
eduardosm: Replaced with `throw_unsup_format`

src/shims/x86/sse.rs
eduardosm: Done

src/shims/x86/sse.rs
eduardosm: Done

src/shims/x86/sse.rs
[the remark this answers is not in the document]
eduardosm: I'll update the comment then.

src/shims/intrinsics/mod.rs
```diff
@@ -120,6 +120,12 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 this.write_bytes_ptr(ptr, iter::repeat(val_byte).take(byte_count.bytes_usize()))?;
             }
 
+            "nontemporal_store" => {
+                let [place, dest] = check_arg_count(args)?;
+                let place = this.deref_operand(place)?;
+                this.copy_op(dest, &place, /*allow_transmute*/ false)?;
+            }
```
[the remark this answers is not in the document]
eduardosm: Done

src/shims/x86/sse.rs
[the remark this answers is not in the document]
eduardosm: Done

src/shims/x86/sse.rs
```diff
@@ -0,0 +1,568 @@
...
+                    imm => {
+                        throw_ub_format!("invalid 3rd parameter of llvm.x86.sse.cmp.ps: {}", imm);
+                    }
+                };
+
+                bin_op_ps(this, which, left, right, dest)?;
+            }
+            // Used to implement _mm_{,u}comi{eq,lt,le,gt,ge,neq}_ps functions.
+            // Compares the first component of `left` and `right` and returns
+            // an scalar value (0 or 1).
```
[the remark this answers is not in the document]
eduardosm: Done

src/shims/x86/sse.rs
RalfJung: Hm, why is it okay to not ignore that flags in the existing float2int code?

RalfJung:
Please never merge master into your PR, and only rebase when there are conflicts.

eduardosm:
> Please never merge master into your PR, and only rebase when there are conflicts or it is needed to resolve CI failures (which might be the case here due to the operand_deref rename).

I was avoiding force-pushing so everything appears a new commits to make it easier to review. I was planning on squashing everything together at the end.

RalfJung:
Yeah, keeping it in new commits is definitely appreciated, even across the rebase.

I guess the merge commits don't hurt if we remember to squash-before-merge.

src/shims/x86/sse.rs
eduardosm: It looks like the existing float2int code first truncates the floating point value and then performs the conversion, so rounding cannot happen in the second step.

```rust
    fn float_to_int_unchecked<F>(
        &self,
        f: F,
        dest_ty: Ty<'tcx>,
    ) -> InterpResult<'tcx, Scalar<Provenance>>
    where
        F: Float + Into<Scalar<Provenance>>,
    {

        // Step 1: cut off the fractional part of `f`. The result of this is
        // guaranteed to be precisely representable in IEEE floats.
        let f = f.round_to_integral(Round::TowardZero).value;

        // Step 2: Cast the truncated float to the target integer type and see if we lose any information in this step.
        Ok(match dest_ty.kind() {
            // Unsigned
            ty::Uint(t) => {
                let size = Integer::from_uint_ty(this, *t).size();
                let res = f.to_u128(size.bits_usize());
                if res.status.is_empty() {
                    // No status flags means there was no further rounding or other loss of precision.
                    Scalar::from_uint(res.value, size)
                } else {
                    // `f` was not representable in this integer type.
                    throw_ub_format!(
                        "`float_to_int_unchecked` intrinsic called on {f} which cannot be represented in target type `{dest_ty:?}`",
                    );
                }
            }
```

RalfJung: Ah, makes sense.

We should probably do this consistently, maybe with a helper function that returns `Option<Scalar>`. I don't have a strong preference whether we do truncate-then-cast like the old code or the direct cast like your new code.

src/shims/x86/sse.rs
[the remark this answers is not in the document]
eduardosm: Nope, I couldn't think of a sane way of implementing them

src/shims/x86/sse.rs
eduardosm: > We should probably do this consistently, maybe with a helper function that returns `Option<Scalar>`.

Agreed, although I'd prefer to do that in another PR after this one.

> Looks like SSE only provides casts to signed types, is that right?

It looks like that.

src/shims/x86/sse.rs
```diff
@@ -0,0 +1,578 @@
...
+                    // https://www.felixcloutier.com/x86/cvtss2si
+                    "cvtss2si" => rustc_apfloat::Round::NearestTiesToEven,
+                    // always truncate
+                    // https://www.felixcloutier.com/x86/cvttss2si
+                    "cvttss2si" => rustc_apfloat::Round::TowardZero,
+                    _ => unreachable!(),
+                };
+
+                let mut exact = false;
+                let cvt = op.to_i128_r(32, rnd, &mut exact);
```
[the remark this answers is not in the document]
eduardosm: But `to_i128` is missing the round direction argument.

src/shims/x86/sse.rs
RalfJung: > Agreed, although I'd prefer to do that in another PR after this one.

Fair.
For this PR, please explain in a comment that "INEXACT" is deliberately ignored.

RalfJung: How can INVALID_OP occur on a cast? If the input is NaN?

src/shims/x86/sse.rs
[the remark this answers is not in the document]
RalfJung: Ah, I spoke too soon.

I guess call you can do then is remove `exact` and use `&mut true` like `to_i128` does in its implementation, to avoid a needless binder.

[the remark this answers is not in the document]
RalfJung: But not sure if that even is an improvement. So yeah let's stick to what you got.

src/shims/x86/sse.rs
eduardosm: rcp (rcpps, rcpss) calculates an approximation of 1/x with a relative error of less than 1.5·2<sup>-12</sup>. As I couldn't find what algorithms are used to implement the operation in hardware, I simply simulated it with 1.0001/x.

Similar for rsqrt, which calculates an approximation of 1/sqrt(x)

src/shims/x86/sse.rs
eduardosm: > For this PR, please explain in a comment that "INEXACT" is deliberately ignored.

Done.

> How can INVALID_OP occur on a cast? If the input is NaN?

Exactly (I looked at the implementation of `to_u128_r` to be sure).

src/shims/x86/sse.rs
eduardosm: Done

tests/pass/intrinsics-x86-sse.rs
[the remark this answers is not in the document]
eduardosm: Done. Tests were missing for the `max` case, which I have just added.

src/shims/x86/sse.rs
RalfJung: But just using `1/x` would also be correct, right?

If the goal is to not have code rely on a particular result, then what about computing `1/x` and then using our pRNG to disturb the result a bit (less than the allowed error)?

eduardosm: Yeah, let's use a random error

src/shims/x86/sse.rs
```diff
@@ -19,6 +19,10 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         // Prefix should have already been checked.
         let unprefixed_name = link_name.as_str().strip_prefix("llvm.x86.sse.").unwrap();
         // All these intrinsics operate on 128-bit (f32x4) SIMD vectors unless stated otherwise.
+        // Many intrinsic names are sufixed with "ps" or "ss". "ps" means thet the operation is
+        // performed on each element of the vector, while "ss" means that the operation is
+        // performed only on the first element, copying the remaining elements from the input
+        // vector (for binary operations, from the left-hand side).
```
RalfJung, line 27: Is there a mnemonic for this, like "ss = scalar \<something with s>" and "ps = \<something p> \<something s>"?

src/shims/x86/sse.rs
```diff
@@ -0,0 +1,602 @@
...
+                let mut exact = false;
+                let cvt = op.to_i128_r(32, rnd, &mut exact);
+                let res = if cvt.status.intersects(
+                    rustc_apfloat::Status::INVALID_OP
+                        | rustc_apfloat::Status::OVERFLOW
+                        | rustc_apfloat::Status::UNDERFLOW,
+                ) {
+                    // Input is NaN (flagged with INVALID_OP) or does not fit
+                    // in an i32 (flagged with OVERFLOW or UNDERFLOW), fallback
+                    // to minimum acording to SSE semantics. The inexact INEXACT
```
RalfJung, line 206:
```suggestion
                    // to minimum acording to SSE semantics. The INEXACT flag
```

src/shims/x86/sse.rs
```diff
@@ -0,0 +1,602 @@
...
+                let mut exact = false;
+                let cvt = op.to_i128_r(64, rnd, &mut exact);
+                let res = if cvt.status.intersects(
+                    rustc_apfloat::Status::INVALID_OP
+                        | rustc_apfloat::Status::OVERFLOW
+                        | rustc_apfloat::Status::UNDERFLOW,
+                ) {
+                    // Input is NaN (flagged with INVALID_OP) or does not fit
+                    // in an i32 (flagged with OVERFLOW or UNDERFLOW), fallback
+                    // to minimum acording to SSE semantics. The inexact INEXACT
```
RalfJung, line 243:
```suggestion
                    // to minimum acording to SSE semantics. The INEXACT flag
```

src/shims/x86/sse.rs
```diff
@@ -518,27 +521,36 @@ fn unary_op_f32<'tcx>(
...
         FloatUnaryOp::Rcp => {
             let op = op.to_scalar();
-            // Use 1.0001 as a crude way of simulating inaccuracy of rcp
-            // 1.0001 / x ~= 1 / x
-            let one = Single::from_bits(1.0001f32.to_bits().into());
-            // floating point operation without side effects
-            #[allow(clippy::arithmetic_side_effects)]
+            // Calculate an approximation of 1 / x as (1 + err) / x, where
+            // err is a random value with a magnitude in the order of 2^-12
+            let one = (Single::from_u128(1).value + gen_random_float_error(this, -12)).value;
```
RalfJung, line 526: Shouldn't it apply the error *after* doing the computation?

src/shims/x86/sse.rs
```diff
@@ -518,27 +521,36 @@ fn unary_op_f32<'tcx>(
...
+fn gen_random_float_error<F: rustc_apfloat::Float>(
+    this: &mut crate::MiriInterpCx<'_, '_>,
+    scale: i32,
+) -> F {
+    let rng = this.machine.rng.get_mut();
+    // generates rand(0, 2^64) * 2^(scale - 64) = rand(0, 1) * 2^scale
+    let v = F::from_u128(rng.gen::<u64>().into()).value.scalbn(scale.checked_sub(64).unwrap());
+    // give it a random sign
+    #[allow(clippy::arithmetic_side_effects)] // floating point operation without side effects
+    if rng.gen::<bool>() { -v } else { v }
```
RalfJung, line 551: That clippy lint triggers on the float types...?!? This is an overloaded operator, I am quite surprised to see clippy complain about it.

src/shims/x86/sse.rs
eduardosm: Done (and the one below)

src/shims/x86/sse.rs
```diff
@@ -0,0 +1,602 @@
...
+/// Returns a random a value in the range (-2^scale, 2^scale)
+fn gen_random_float_error<F: rustc_apfloat::Float>(
+    this: &mut crate::MiriInterpCx<'_, '_>,
+    scale: i32,
+) -> F {
+    let rng = this.machine.rng.get_mut();
+    // generates rand(0, 2^64) * 2^(scale - 64) = rand(0, 1) * 2^scale
+    let v = F::from_u128(rng.gen::<u64>().into()).value.scalbn(scale.checked_sub(64).unwrap());
+    // give it a random sign
+    #[allow(clippy::arithmetic_side_effects)] // floating point operation without side effects
```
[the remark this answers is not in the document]
eduardosm: Done

src/shims/x86/sse.rs
eduardosm: Applying the error after doing the computation would result in an expression like `(1 / x) * (1 + err)`, which I simplified to `(1 + err) / x`.

src/shims/x86/sse.rs
eduardosm: I guess "ss = scalar single" and "ps = packed single" ("single" as in single precision FP) or something like that.

src/shims/x86/sse.rs
RalfJung: I'd prefer to do the obviously correct thing consistently (here and for sqrt) over being clever -- unless there is some good reason for having to be clever.

src/shims/x86/sse.rs
RalfJung: That makes sense, can you add it to the comment?

src/shims/x86/sse.rs
eduardosm: Done, there is not really a reason :)

src/shims/x86/sse.rs
eduardosm: Done

RalfJung:
I've added a commit that moves the entire "multiply by (1+err)" logic into the helper function, makes it much more clear what's going on IMO. I wonder if it wouldn't be easier to just XOR all but the first 11 or 12 bits of the mantissa with random garbage, but I don't know how to do that with apfloat so :shrug: .

LGTM, thanks a lot for enduring the review! Can you please squash?

eduardosm:
Squashed! Thanks for the thorough review :)

RalfJung:
This test failure here is concerning: https://github.com/rust-lang/miri/actions/runs/5832262620/job/15817332780

Is it possible that the tests are expecting more precision from RCP/RSQRT than they provide?

eduardosm:
> This test failure here is concerning: https://github.com/rust-lang/miri/actions/runs/5832262620/job/15817332780
> 
> This failed [here](https://github.com/rust-lang/miri/blob/277e65060553708d74e10cfac623a88105b9eac8/tests/pass/intrinsics-x86-sse.rs#L136). Is it possible that the tests are expecting more precision from RCP/RSQRT than they provide?

Nah, that's because it was calculating sqrt(x) instead of 1/sqrt(x).

https://github.com/rust-lang/miri/blob/31ed6d098e8e2a614f7bd01966cfb411b9934afe/src/shims/x86/sse.rs#L531-L539

 I fixed it before squashing.

RalfJung:
>  Nah, that's because it was calculating sqrt(x) instead of 1/sqrt(x).

Oh, I screwed up then. *oops* Thanks for taking care of it!

PR #2993, opened 2023-07-26
TB: Redefine trigger condition for protectors

Vanille-N:
The Coq formalization revealed that as currently implemented, read accesses did not always commute.
Indeed starting from a lazily initialized `Active` protected tag, applying a foreign read then a child read produces `Frozen`, but child read then foreign read triggers UB (because the child read initializes _before_ the `Active -> Frozen`).

This reformulation of when protectors trigger fixes that issue:
- instead of `Active + foreign read -> Frozen` and `Active -> Frozen` when protected is UB
- we do `Active + foreign read -> if protected { Disabled } else { Frozen }`

There is already precedent for transitions being dependent on the presence of a protector (`Reserved + foreign read -> if protected { Frozen } else { Reserved }`), and this has the nice side-effect of simplifying the protector trigger condition to just an equality check against `Disabled` since now there is protector UB iff a protected tag becomes `Disabled`.

In order not to introduce an extra `if`, it was decided that `Disabled -> Disabled` would be UB when protected, which was not the case previously. This is merely a theoretical for now because a protected `Disabled` is unreachable in the first place.

The extra test is not directly related to this modification, but also checks things related to protectors and lazy initialization.


tests/fail/both_borrows/aliasing_mut4.tree.stderr
```diff
@@ -6,8 +6,8 @@ LL |         ptr::write(dest, src);
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Tree Borrows rules it violated are still experimental
    = help: the accessed tag <TAG> is foreign to the protected tag <TAG> (i.e., it is not a child)
-   = help: this foreign write access would cause the protected tag <TAG> to transition from Frozen to Disabled
-   = help: this transition would be a loss of read permissions, which is not allowed for protected tags
+   = help: this foreign write access would cause the protected tag <TAG> currently Frozen to become Disabled
```
RalfJung, line 9:
```suggestion
   = help: this foreign write access would cause the protected tag <TAG> (currently Frozen) to become Disabled
```
this reads a bit strange without the parentheses

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -69,6 +70,74 @@ impl LocationState {
...
+                // (2) justified by `all_transitions_idempotent`
+                (Some(old), new) if old == new => true,
+                // In all other cases there has been a recent enough
+                // child access that the effects of the new foreign access
+                // need to be applied to this subtree.
+                _ => false,
+            };
+            if new_access_noop {
+                // Abort traversal if the new transition is indeed guaranteed
+                // to be noop.
```
RalfJung, line 126: Crucially, we do *not* update `latest_foreign_access` here. If the previous last access was a write and this one is a read, `latest_foreign_access` remains a write. That's what we want, it means the next write will get the optimization. But it does mean that `latest_foreign_access` is a bit if a misnomer since literally speaking, the last foreign access in this situation was a read.

Really `latest_foreign_access` stores the "highest access of the streak of preceding foreign accesses", where write is higher than read.

That seems worthy of a comment.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -69,6 +70,74 @@ impl LocationState {
...
+        rel_pos: AccessRelatedness,
+        protected: bool,
+    ) -> Result<PermTransition, TransitionError> {
+        let old_perm = self.permission;
+        let transition = Permission::perform_access(access_kind, rel_pos, old_perm, protected)
+            .ok_or(TransitionError::ChildAccessForbidden(old_perm))?;
+        if protected
+            // Can't trigger Protector on uninitialized locations
+            && self.initialized
+            && transition.produces_disabled()
```
RalfJung, line 90:
```suggestion
        // Why do only initialized locations cause protector errors?
        // Consider two mutable references `x`, `y` into disjoint parts of
        // the same allocation. A priori, these may actually both be used to
        // access the entire allocation, as long as only reads occur. However,
        // a write to `y` needs to somehow record that `x` can no longer be used
        // on that location at all. For these uninitialized locations (i.e., locations
        // that haven't been accessed with `x` yet), we track the "future initial state":
        // it defaults to whatever the initial state of the tag is,
        // but the access to `y` moves that "future initial state" of `x` to 'disabled'.
        // However, usually a reserved -> disabled transition would be UB due to the protector!
        // So clearly protectors shouldn't fire for such "future initial state" transitions.
        if protected
            && self.initialized
            && transition.produces_disabled()
```

RalfJung:
Thanks!

@bors r+

PR #3000, opened 2023-07-31
auto-detect no_std where possible

RalfJung:
r? @oli-obk 
Cc https://rust-lang.zulipchat.com/#narrow/stream/269128-miri/topic/restricted_std.20sysroot.3F

### DIFF
--- a/cargo-miri/src/setup.rs
+++ b/cargo-miri/src/setup.rs
@@ -74,7 +74,17 @@ pub fn setup(
         }
     };
     // Sysroot configuration and build details.
-    let sysroot_config = if std::env::var_os("MIRI_NO_STD").is_some() {
+    let no_std = match std::env::var_os("MIRI_NO_STD") {
+        None =>
+        // No-std heuristic taken from rust/src/bootstrap/config.rs
+        // (https://github.com/rust-lang/rust/blob/25b5af1b3a0b9e2c0c57b223b2d0e3e203869b2c/src/bootstrap/config.rs#L549-L555).
+            target.contains("-none")
+                || target.contains("nvptx")
+                || target.contains("switch")
+                || target.contains("-uefi"),
+        Some(val) => val != "0",
+    };
+    let sysroot_config = if no_std {
         SysrootConfig::NoStd
     } else {
         SysrootConfig::WithStd {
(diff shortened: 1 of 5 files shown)


ci.sh
```diff
@@ -112,7 +112,7 @@ case $HOST_TARGET in
     MIRI_TEST_TARGET=aarch64-linux-android run_tests_minimal hello integer vec panic/panic
     MIRI_TEST_TARGET=wasm32-wasi run_tests_minimal no_std integer strings
     MIRI_TEST_TARGET=wasm32-unknown-unknown run_tests_minimal no_std integer strings
-    MIRI_TEST_TARGET=thumbv7em-none-eabihf MIRI_NO_STD=1 run_tests_minimal no_std # no_std embedded architecture
+    MIRI_TEST_TARGET=thumbv7em-none-eabihf run_tests_minimal no_std # no_std embedded architecture
```
oli-obk, line 115: Weird, why aren't we detecting this with the `contains("-none")` check?

cargo-miri/src/setup.rs
```diff
@@ -67,7 +67,13 @@ pub fn setup(subcommand: &MiriCommand, target: &str, rustc_version: &VersionMeta
...
     };
     // Sysroot configuration and build details.
-    let sysroot_config = if std::env::var_os("MIRI_NO_STD").is_some() {
+    // No-std heuristic taken from rust/src/bootstrap/config.rs
+    // (https://github.com/rust-lang/rust/blob/25b5af1b3a0b9e2c0c57b223b2d0e3e203869b2c/src/bootstrap/config.rs#L549-L555).
+    let no_std = target.contains("-none")
+        || target.contains("nvptx")
+        || target.contains("switch")
+        || target.contains("-uefi");
+    let sysroot_config = if no_std || std::env::var_os("MIRI_NO_STD").is_some() {
```
oli-obk, line 76: Do we want to be able to disable the heuristic with `MIRI_NO_STD=0`?

RalfJung:
Ah, the test suite also checks the env var

https://github.com/rust-lang/miri/blob/defbb668ffe7c28af448cf8054ba100450e8b5b5/tests/compiletest.rs#L97

I don't think I want to repeat the target name check there, that seems a bit too ugly...

RalfJung:
I wonder if we could have ui_test only build the dependencies if at least one test matches the filter? That would elegantly avoid the need for the `MIRI_NO_STD` check.

oli-obk:
Yea, I'm working on this, but it's a bigger refactoring to do it nicely. If you want it fast I can build a hacky version of it.

RalfJung:
Nah, this is not urgent.

RalfJung:
Marking as blocked on https://github.com/rust-lang/miri/pull/3008

oli-obk:
> I wonder if we could have ui_test only build the dependencies if at least one test matches the filter? That would elegantly avoid the need for the `MIRI_NO_STD` check.

this is happening now

RalfJung:
Looks like that is working. :)
@rustbot ready

Issue #2783, opened 2023-02-10
Remove uses of `set_var`


RalfJung:
`set_var` [has a problem](https://github.com/rust-lang/rust/issues/90308) and there are plans to make it unsafe. So we should find replacements for our uses of it.

- https://github.com/rust-lang/miri/blob/d0fd429d46b39662b649aa523858c2bf794435fe/src/bin/miri.rs#L169
  Needs some way to pass this flag to the rustc logging machinery without an env var.
- https://github.com/rust-lang/miri/blob/d0fd429d46b39662b649aa523858c2bf794435fe/src/bin/miri.rs#L191-L198
  Needs a way to init rustc logging by giving it the *value* of the filter expression we want, instead of the name of an env var with the filter.
- https://github.com/rust-lang/miri/blob/d0fd429d46b39662b649aa523858c2bf794435fe/cargo-miri/src/setup.rs#L121
  This one we could already do: find some other way to propagate this information to all parts of the binary, and to all created sub-processes.

RalfJung:
https://github.com/rust-lang/miri/pull/3006 removes the one in cargo-miri.

The others need logging machinery extensions first, so this is blocked on there being other APIs to control rustc logging.

RalfJung:
All uses of `set_var` in Miri itself are gone now thanks to https://github.com/rust-lang/rust/pull/117814. All we have left is some uses in compiletest:

https://github.com/rust-lang/miri/blob/d4d85a8f7e72485c572fc0ed29ef279fe47489ed/tests/compiletest.rs#L235-L238

PR #3006, opened 2023-08-02
miri-script and cargo-miri cleanups

RalfJung:
At least in an initial test, `cmd!(Shell::new()?, "env")` does show the full environment. Let's see what CI says.

@osiewicz why did you add this code to forward the host environment?


osiewicz:
Hi,
Yeah, it might be redundant given that it was added quite early in the migration to `xshell` and not everything around environment variables worked fine back then.

RalfJung:
Okay then, let's get rid of it. :) Please let me know if there is any other cleanup you had pending. :D

Also, this finally gets rid of that pesky `set_var` in cargo-miri. It wasn't even that hard. ;) Cc https://github.com/rust-lang/miri/issues/2783

@bors r+

PR #3018, opened 2023-08-08
[nit][typo] Fix out of order words.

ttsugriy:


RalfJung:
Thanks for the PR!

However, as far as I know both of these are allowed?

saethlin:
I never expected that working on Miri would push my English skills like it does, but here we are.

It would also be valid English to say
> MIri currently does not check

All these orderings are valid, but they imply slightly different emphasis and the current wording just reads awkwardly. I think I would prefer that we just delete the "currently", because the current wording is just awkward (if I heard someone use it I would start to suspect they are not a native speaker), and with this PR the phrasing implies that we will add this check in the future. And I don't think there's any reason to suggest to the reader that we will.

RalfJung:
Works for me, let's just remove the `currently`.

@ttsugriy can you adjust the PR?

RalfJung:
Let's try this again.

@bors retry

PR #3019, opened 2023-08-08
[perf] Reserve capacity for argvs.

ttsugriy:
This avoids unnecessary allocations to grow vector to sufficient size.


RalfJung:
Thanks!

This code is totally not perf critical, so it's kind of unnecessary to optimize it (unless you can show a benchmark where this helps). But it doesn't hurt much either so *shrug* LGTM...

@bors r+

PR #3021, opened 2023-08-08
Use Vec's binary search instead of hand-written one.

ttsugriy:


src/concurrency/range_object_map.rs
```diff
@@ -42,30 +42,15 @@ impl<T> RangeObjectMap<T> {
...
-                return Err(left);
-            }
-            let candidate = left.checked_add(right).unwrap() / 2;
-            let elem = &self.v[candidate];
+        self.v.binary_search_by(|elem| -> std::cmp::Ordering {
             if offset < elem.range.start {
-                // We are too far right (offset is further left).
-                debug_assert!(candidate < right); // we are making progress
-                right = candidate;
+                std::cmp::Ordering::Greater
```
RalfJung, line 47: Please preserve the comments inside the `if`s here, they still seem helpful

src/concurrency/range_object_map.rs
```diff
@@ -42,30 +42,18 @@ impl<T> RangeObjectMap<T> {
...
-            if left == right {
-                // No element contains the given offset. But the
-                // position is where such element should be placed at.
-                return Err(left);
-            }
-            let candidate = left.checked_add(right).unwrap() / 2;
-            let elem = &self.v[candidate];
+        self.v.binary_search_by(|elem| -> std::cmp::Ordering {
             if offset < elem.range.start {
                 // We are too far right (offset is further left).
```
RalfJung, line 47:
```suggestion
                // We are too far right (offset is further left).
                // (`Greater` means that `elem` is greater than the desired target.)
```

RalfJung:
Thanks, this looks much cleaner!

@bors r+

PR #3022, opened 2023-08-10
replace AsAny hack by trait upcasting :)

RalfJung:


RalfJung:
Ah, this is actually wrong, I am upcasting the wrong thing. That's kind of subtle and easy to miss. (Cc @compiler-errors )

compiler-errors (approved):
r=me (if you're okay with non-miri approvals 😜)

RalfJung:
> not &&dyn Tr

They were `Box<dyn Tr>`, which is very close to `&dyn Tr` (no double pointers!) so I was quite surprised to need `&**` here.

src/shims/unix/fs.rs
```diff
@@ -72,6 +73,18 @@ pub trait FileDescriptor: std::fmt::Debug + helpers::AsAny {
...
+    #[inline(always)]
+    pub fn downcast_ref<T: Any>(&self) -> Option<&T> {
+        (self as &dyn Any).downcast_ref()
+    }
+
+    #[inline(always)]
+    pub fn downcast_mut<T: Any>(&mut self) -> Option<&mut T> {
+        (self as &mut dyn Any).downcast_mut()
+    }
+}
```
compiler-errors, line 86: Yeah this is the best approach!

PR #3023, opened 2023-08-10
add gamma function shims

RalfJung:
needs a rustup to fetch the new declarations in std


tests/pass/intrinsics-math.rs
```diff
@@ -1,5 +1,4 @@
-// SPDX-License-Identifier: MIT OR Apache-2.0
-// SPDX-FileCopyrightText: The Rust Project Developers (see https://thanks.rust-lang.org)
```
RalfJung, line 2: These were some old license headers that we just failed to remove when most of them got removed. The license and copyright holders are the same as for the main repo so they add no information.

PR #3025, opened 2023-08-11
Add checked float-to-int helper function

eduardosm:
As discussed in https://github.com/rust-lang/miri/pull/2989#discussion_r1287252367

### DIFF
--- a/src/shims/intrinsics/mod.rs
+++ b/src/shims/intrinsics/mod.rs
@@ -6,12 +6,12 @@ use std::iter;
 use log::trace;
 
 use rustc_apfloat::{Float, Round};
-use rustc_middle::ty::layout::{IntegerExt, LayoutOf};
+use rustc_middle::ty::layout::LayoutOf;
 use rustc_middle::{
     mir,
     ty::{self, FloatTy, Ty},
 };
-use rustc_target::abi::{Integer, Size};
+use rustc_target::abi::Size;
 
 use crate::*;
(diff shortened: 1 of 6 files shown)


src/shims/intrinsics/mod.rs
```diff
@@ -393,47 +393,10 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         F: Float + Into<Scalar<Provenance>>,
     {
```
RalfJung, line 394: This is not worth its own function any more (and having 2 functions with the same name is anyway potentially confusing).

eduardosm: Inlined the function

src/shims/intrinsics/simd.rs
```diff
@@ -420,7 +420,6 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                     }
                 }
             }
-            #[rustfmt::skip]
```
eduardosm, line 423: I wasn't sure about the purpose of this

RalfJung: It avoided a bunch of reformatting that I considered undesirable. But now the branches have fairly different sizes anyway so it seems okay.

RalfJung:
Thanks!
@bors r+

PR #3027, opened 2023-08-16
Avoid unnecessary Vec resize.

ttsugriy:
If `size > 0` current implementation will first create an empty vec and then push an element into it, which will cause a resize that can be easily avoided.

It's obviously not a big deal, but this also gets rid of `mut` local variable.

### DIFF
--- a/src/range_map.rs
+++ b/src/range_map.rs
@@ -27,11 +27,8 @@ impl<T> RangeMap<T> {
     #[inline(always)]
     pub fn new(size: Size, init: T) -> RangeMap<T> {
         let size = size.bytes();
-        let mut map = RangeMap { v: Vec::new() };
-        if size > 0 {
-            map.v.push(Elem { range: 0..size, data: init });
-        }
-        map
+        let v = if size > 0 { vec![Elem { range: 0..size, data: init }] } else { Vec::new() };
+        RangeMap { v }
     }
 
     /// Finds the index containing the given offset.


RalfJung:
`Vec::new` is completely free, so the old code isn't resizing either. It is just creating a new allocation in the `size > 0` case.

I like avoiding the `mut`, but this expression is getting a bit big. Can you let-bind the initial vector?

RalfJung:
> that's not entirely true and instead of arguing I'll just drop the [benchmark](https://gist.github.com/ttsugriy/5b4d3c3ea18bbd95371b8711bfb6f4e3) results below:

Hm, interesting. I'm a bit worried that the allocations are optimized away entirely, can you `drop(hint::black_box(v))` to make sure that does not happen?

> sorry, I'm not sure I understand. Are you suggesting something like

Yes.

ttsugriy:
> > that's not entirely true and instead of arguing I'll just drop the [benchmark](https://gist.github.com/ttsugriy/5b4d3c3ea18bbd95371b8711bfb6f4e3) results below:
> 
> Hm, interesting. I'm a bit worried that the allocations are optimized away entirely, can you `drop(hint::black_box(v))` to make sure that does not happen? Or does criterion do that automatically with the return value?

criterion does it automatically but it's even easier to understand the perf impact by looking at assembly for both versions side by side - https://godbolt.org/z/GoT4Kes5b

> > sorry, I'm not sure I understand. Are you suggesting something like
> 
> Yes.

I don't have a preference, so I've updated this PR with your suggestion :)

RalfJung:
Thanks! I highly doubt this will even be measurable in any Miri benchmark, but I like getting rid of the `mut`. :)

@bors r+

ttsugriy:
> Thanks! I highly doubt this will even be measurable in any Miri benchmark, but I like getting rid of the `mut`. :)

You are very welcome! I'm sure this change won't affect Miri's perf on the global scale, but since most of the code in the world is written by copy-pasting it's best to have more "better" examples :D

PR #3028, opened 2023-08-16
Replace hand-written binary search with Vec::binary_search_by.

ttsugriy:
It's similar to https://github.com/rust-lang/miri/pull/3021 and should improve maintainability.


RalfJung:
Makes sense, thanks!
@bors r+

PR #3030, opened 2023-08-16
on out-of-bounds error, show where the allocation was created

RalfJung:
This seems useful, I think?
r? @saethlin


saethlin:
I skimmed over a few examples of this diagnostic in the wild and I agree this will make them less mysterious. Also impressively little code to add this!

@bors r+

RalfJung:
> Also impressively little code to add this!

You did all the hard work. ;)

PR #3036, opened 2023-08-22
automatically start and stop josh in rustc-pull/push

RalfJung:
Let's make use of the fact that this is not a shell script any more. :)


oli-obk:
This is awesome! Thank you

PR #3046, opened 2023-08-31
Rustup

RalfJung:
also more ABI compat tests


RalfJung:
Ah, we're running into https://github.com/rust-lang/rust/issues/115404

@bors r+

PR #3047, opened 2023-08-31
update target support section

RalfJung:
The MIPS targets were demoted to tier 3, so let's recommend a different big-endian target.

Also instead of listing the targets we test individually, just say we test all tier 1 targets.

r? @rust-lang/miri


saethlin:
Yeah, I've been thinking about how we should probably use s390x more considering the amount of community/IBM support that target gets.

@bors r+

RalfJung:
Oh interesting, on `s390x-unknown-linux-gnu` the C ABI for i32 vs u32 differs. We do catch this since the `mode` differs, but given that -- maybe we shouldn't ignore the sign on any target, to ensure more portability? That'll have to be adjusted in a rustc PR though.

RalfJung:
>  considering the amount of community/IBM support that target gets.

I didn't even know about that, it's just the last big-endian tier 2 target left. ;)

@bors r=saethlin

PR #3054, opened 2023-09-07
Issue discovered in TB: spurious reads are not (yet) possible in a concurrent setting

Vanille-N:
We discovered a week ago that in general, the current model of TB does not allow spurious reads because although reads provably never invalidate other reads, they migh invalidate writes.

Consider the code
```rs
fn f1(x: &u8) {}
fn f2(y: &mut u8) -> &mut u8 { &mut *y }

let mut data = 0;
let _ = thread::spawn(|| {
    f1(&mut data)
};
let _ = thread::spawn(|| {
    let y = f2(&mut data);
    *y = 42;
});
```
of which one possible interleaving is
```rs
1: retag x (&, protect) // x: [P]Frozen
2: retag y (&mut, protect) // y: [P]Reserved, x: [P]Frozen
1: return f1 // x: [P]Frozen -> Frozen, y: [P]Reserved
2: return f2 // x: Frozen, y: [P]Reserved -> Reserved
2: write y // x: Disabled, y: Active
```
that does not have UB.

Assume enough barriers to force this specific interleaving, and consider that the compiler could choose to insert a spurious read throug `x` during the call to `f1` which would produce
```rs
1: retag x (&, protect) // x: [P]Frozen
2: retag y (&mut, protect) // y: [P]Reserved, x: [P]Frozen
1: spurious read x // x:  [P]Frozen, y: [P]Reserved -> [P]Frozen
1: return f1 // x: [P]Frozen -> Frozen, y: [P]Frozen
2: return f2 // x: Frozen, y: [P]Frozen -> Frozen
2: write y // UB
```
Thus the target of the optimization (with a spurious read) has UB when the source did not.

This is bad.

SB is not affected because the code would be UB as early as `retag y`, this happens because we're trying to be a bit more subtle than that, and because the effects of a foreign read on a protected `&mut` bleed outside of the boundaries of the protector. Fortunately we have a fix planned, but in the meantime here are some `#[should_panic]` exhaustive tests to illustrate the issue.

The error message printed by the `#[should_panic]` tests flags the present issue in slightly more general terms: it says that the sequence `retag x (&, protect); retag y (&mut, protect);` produces the configuration `C_source := x: [P]Frozen, x: [P]Reserved`, and that inserting a spurious read through `x` turns it into `C_target := x: [P]Frozen, y: [P]Reserved`.
It then says that `C_source` is distinguishable from `C_target`, which means that there exists a sequence of instructions applied to both that triggers UB in `C_target` but not in `C_source`.
It happens that one such sequence is `1: return f1; 2: return f2; 2: write y;` as shown above, but it is not the only one, as for example the interleaving `1: return f1; 2: write y;` is also problematic.


src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -442,30 +444,32 @@ mod propagation_optimization_checks {
...
     fn foreign_read_is_noop_after_foreign_write() {
         use transition::*;
         let old_access = AccessKind::Write;
         let new_access = AccessKind::Read;
-        for old in PermissionPriv::all() {
-            for (old_protected, new_protected) in [(true, true), (true, false), (false, false)] {
-                for rel_pos in AccessRelatedness::all().filter(|rel| rel.is_foreign()) {
+        for old in PermissionPriv::exhaustive() {
+            for [old_protected, new_protected] in <[bool; 2]>::exhaustive() {
+                precondition!(old_protected || !new_protected);
```
RalfJung, line 470: Why do we have that precondition?
(I realize this is not new, but it still seems worth explaining.)

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -658,25 +670,26 @@ impl AccessRelatedness {
     /// for the child node. This function assumes that we propagate away from the initial
     /// access.
     pub fn for_child(self) -> Self {
-        use AccessRelatedness::*;
+        use AccessRelatedness as Rel;
```
RalfJung, line 673: Why the renames? I find that more confusing than `::*` because it means one has to learn a new name.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -725,27 +736,480 @@ mod commutation_tests {
...
+                    assert!(
+                        kind1 == AccessKind::Read && kind2 == AccessKind::Read,
+                        "failed to enforce noalias between two accesses that are not both reads"
+                    );
+                }
+            }
+        }
+    }
+
+    // Possible pairs of `AccessRelatedness` that can occur
```
RalfJung, line 757: Please move these tests into a new file... this almost double the current file size.^^
(You can make it `tree/tests.rs` or so if you need access to private stuff of `tree`.)

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -725,27 +736,480 @@ mod commutation_tests {
...
+    // to two tags that are mutually foreign.
+    // The two tags are by convention named `x` and `y` throughout these tests.
+    // This includes everything except `(StrictChildAccess, StrictChildAccess)`
+    // because no tag can be a child of both.
```
RalfJung, line 761: This vaguely makes sense but I have no idea how the three items below relate to the pairs you are talking about here. Please add doc comments at the enum variants explaining that. After all there's more than 2 `AccessRelatedness`...

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -725,27 +736,480 @@ mod commutation_tests {
...
+            Box::new(
+                <(AccessRel2, AccessKind)>::exhaustive().map(|(rel2, kind)| Access2 { rel2, kind }),
+            )
+        }
+    }
+
+    #[derive(Clone, PartialEq, Eq, Hash)]
+    /// Our testing utilties include the protector bound to the state
+    /// rather than the tag. There is little room for errors since we only
+    /// do examples on single-byte locations anyway.
```
RalfJung, line 793: I don't understand this comment.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -725,27 +736,480 @@ mod commutation_tests {
...
+    struct LocStateProt {
+        state: LocationState,
+        prot: bool,
+    }
+
+    impl LocStateProt {
+        fn perform_access(&self, kind: AccessKind, rel: AccessRelatedness) -> Result<Self, ()> {
+            let mut state = self.state;
+            let _ = state.perform_access(kind, rel, self.prot).map_err(|_| ())?;
```
RalfJung, line 802:
```suggestion
            state.perform_access(kind, rel, self.prot).map_err(|_| ())?;
```
or is there a reason to explicitly use an `_` here?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -725,27 +736,480 @@ mod commutation_tests {
...
+            Ok(Self { state, prot: self.prot })
+        }
+
+        /// End of the protector.
+        fn release(&self) -> Self {
+            let state = self.state;
+            let prot = false;
+            Self { state, prot }
```
RalfJung, line 810:
```suggestion
            Self { prot: false, state: self.state }
```
The name `release` also isn't very clear. `end_protector` or so, maybe?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -725,27 +736,480 @@ mod commutation_tests {
...
+    /// could be removed in an arbitrary order.
+    struct LocStateProt2 {
+        x: LocStateProt,
+        y: LocStateProt,
+    }
+
+    impl LocStateProt2 {
+        fn perform_access2(&self, acc: Access2) -> Result<Self, ()> {
+            use AccessRel2 as Rel2;
+            use AccessRelatedness as Rel1;
```
RalfJung, line 848: Please don't use renames.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -725,27 +736,480 @@ mod commutation_tests {
...
+    /// A `(LocStateProt, LocStateProt)` has more degrees of freedom
+    /// than what we want to allow for two aliasing and mutually foreign
+    /// tags (e.g. `(Active, Active)` among many others is unreachable through
+    /// normal means), so a pair of tags gets its own methods.
+    /// We'll still generate these cases exhaustively, but it prevents silly mistakes.
+    /// Pointers `x` and `y` might be used in different threads, so their protectors
+    /// could be removed in an arbitrary order.
+    struct LocStateProt2 {
+        x: LocStateProt,
+        y: LocStateProt,
```
RalfJung, line 842: I'm confused, why are there two locations? There should be only one state, right, for the one location that everything points to?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -725,27 +736,480 @@ mod commutation_tests {
...
+        events: Vec<Option<Access2>>,
+    }
+
+    impl Exhaustive for Opaque2 {
+        fn exhaustive() -> Box<dyn Iterator<Item = Self>> {
+            // There are a lot of combinations (infinitely many in fact),
+            // so we're only generating a small subset and then testing through other means
+            // that we have a representative sample.
+            // The skips that we introduce here take advantage of
+            // - beyond length 4 there is a shorter sequence that has the same effect
```
RalfJung, line 904: Why?

Also the length you are suing in the end is 3, not 4. But the comment here indicates 4 ("3 or less" can be None, "more than 4" is not needed -> 4 is what we want)

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -725,27 +736,480 @@ mod commutation_tests {
...
+            // - since accesses are idempotent it's never useful to have the same twice in a row
+            // - sequences of length 3 or less might as well have their no-ops (`None`) at the
+            //   beginning.
+            // In addition to making the space to explore finite (when there are infinitely
+            // many possible sequences of accesses) this further reduces the space to explore
+            // by about a factor 3 compared to a naive implementation that would not
+            // filter on `second_no_weaker`, which is very much worth it.
+            fn second_no_weaker(acc1: &Option<Access2>, acc2: &Option<Access2>) -> bool {
```
RalfJung, line 912: What is `second_no_weaker` testing for and why is that sound? Needs comment.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -725,27 +736,480 @@ mod commutation_tests {
...
+                                "UB can be triggered from {reach} by {step:?}, should be recorded as reachable from {state}"
+                            ),
+                    }
+                }
+            }
+        }
+    }
+
+    impl LocStateProt2 {
+        fn perform_opaque2(mut self, code: &Opaque2) -> Result<Self, ()> {
```
RalfJung, line 976: Please move this up above the test. It's used in the test after all.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -725,27 +736,480 @@ mod commutation_tests {
...
+                write!(f, "{kind} {ptr}; ")?;
+            }
+            Ok(())
+        }
+    }
+
+    impl LocStateProt2 {
+        fn trivially_indistinguishable(&self, other: &Self) -> bool {
+            self == other
+        }
```
RalfJung, line 1007: Why is this a method?^^

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -725,27 +736,480 @@ mod commutation_tests {
...
+        /// that causes UB in the target but not in the source.
+        /// This (fast enough) implementation simply explores the reachable space
+        /// by all sequences of `Access2` or function returns.
+        fn distinguishable(&self, other: &Self) -> bool {
+            if self.trivially_indistinguishable(other) { return false; }
+            let mut states = vec![(self.clone(), other.clone())];
+            let mut seen = FxHashSet::default();
+            while let Some(state) = states.pop() {
+                if seen.contains(&state) { continue; }
+                seen.insert(state.clone());
```
RalfJung, line 1021: `seen.insert` returns whether the thing was in the set, so you don't need to do `contains` and then `insert` -- you can do this in a single step.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -725,27 +736,480 @@ mod commutation_tests {
...
+                let (source, target) = state;
+                for acc in Access2::exhaustive() {
```
RalfJung, line 1023:
```suggestion
                // Generate successor states through accesses.
                for acc in Access2::exhaustive() {
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -725,27 +736,480 @@ mod commutation_tests {
...
+            }
+        }
+
+        impl Pattern {
+            fn execute(&self, spurious_read: bool) -> Result<LocStateProt2, ()> {
+                let state = LocStateProt2 {
+                    x: LocStateProt { state: self.x_initial_state, prot: true },
+                    y: LocStateProt { state: self.y_initial_state, prot: self.y_protected },
+                };
+                // retag y
```
RalfJung, line 1152: The text also says there's a "retag x". Where is that?
And below is a read, not a retag... needs more explanation.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -725,27 +736,480 @@ mod commutation_tests {
...
+    /// two mutually foreign tags, we test if a spurious read through one
+    /// changes the outcome (introduces UB in the target when there is none
+    /// in the source).
+    fn spurious_read() {
+        // The general code pattern being exhaustively tested is
+        //     [thread 1]          || [thread 2]
+        //     retag x (protect)   || retag y (protect/noprotect)
+        //                    arbitrary code
+        //                         || optional return
+        //     <spurious read x>   ||
```
RalfJung, line 1083: Needs to explain that x must still be protected when we do the spurious read, that's why x is fixed as protected and there is no optional return here. (There is an optional return in the distinguishability test though.)

src/borrow_tracker/tree_borrows/tree.rs
Vanille-N: One location, but the two pointers each have their own state.
`LocStateProt` is per-pointer per-byte, so we need 2 for a 1-byte allocation with two pointers.

RalfJung: Oh I see, yes that makes sense.

So the first sentence of the docs should be: for a single location, this stores the state of two pointer tags.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -725,27 +736,480 @@ mod commutation_tests {
...
+    impl fmt::Display for LocStateProt2 {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            write!(f, "x:{}, y:{}", self.x, self.y)
+        }
+    }
+
+    // Arbitrary sequence of accesses, as experienced by two unrelated pointers.
+    #[derive(Debug, Clone)]
+    struct Opaque2 {
+        events: Vec<Option<Access2>>,
```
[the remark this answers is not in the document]
Vanille-N: There are other pointers in the allocation. The state e.g. `x: Disabled, y: Disabled` is not reachable from any initial state unless we have a way to create an access through neither `x` nor `y` but something foreign to both.

Maybe this doesn't actually change anything for the test, but that would need justification. With `AccessRel2` I can easily prove that all cases are covered without reasoning on the actual state machine: any pointer is either
- `x` or a child of `x`, in which case it cannot also be `y` or a child of `y` and it is covered by the `ChildX` variant;
- `y` or a child of `y`, in which case it cannot also be `x` or a child of `x` and it is covered by the `ChildY` variant;
- if it fits neither of the two disjoint categories, then it must be foreign for both `x` and `y` and it is then covered by the `Other` variant.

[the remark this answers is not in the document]
Vanille-N: > the relatedness of the pointers can change?

Ok if I understand correctly the misunderstanding here, I just might have very poorly explained what `AccessRel2` is.

It's not the position of `x` relative to `y`, it's the position of an arbitrary access relative to *both* `x` and `y`, assuming that `x` and `y` are mutually foreign.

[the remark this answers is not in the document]
RalfJung: Okay, so it's basically this?
```rust
/// Determines which pointer is used for a given memory access.
/// We assume by convention that we have two mutually foreign pointers `x` and `y`;
/// every access is either through one of these pointers or some unrelated pointers.
enum WhichPtr {
  /// x, or one of its children
  X
  /// y, or one of its children
  Y
  /// None of the above
  Other
}
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -725,27 +736,480 @@ mod commutation_tests {
...
+                if seen.contains(&state) { continue; }
+                seen.insert(state.clone());
+                let (source, target) = state;
+                for acc in Access2::exhaustive() {
+                    let Ok(new_source) = source.perform_access2(acc) else { continue; };
+                    let Ok(new_target) = target.perform_access2(acc) else { return true; };
+                    if new_source.trivially_indistinguishable(&new_target) { continue; }
+                    states.push((new_source, new_target));
+                }
+                if source.x.prot {
```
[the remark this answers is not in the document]
Vanille-N: Sure, and in fact restructuring the code a bit to incorporate that naturally leads to much more efficient and more complete checks.

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,671 @@
...
+}
+
+#[derive(Clone, PartialEq, Eq, Hash)]
+/// The states of two pointers to the same location.
+///
+/// Note that the two states may evolve non-independently because
+/// `x` and `y` of a single `LocStateProt2` have a static relative position.
+/// This makes `LocStateProt2` more meaningful than a simple
+/// `(LocStateProt, LocStateProt)` where the two states are not guaranteed
+/// to be updated at the same time.
```
RalfJung, line 320:
```suggestion
/// Note that the two states interact: using one pointer can have
/// an impact in the other.
/// This makes `LocStateProt2` more meaningful than a simple
/// `(LocStateProt, LocStateProt)` where the two states are not guaranteed
/// to be updated at the same time.
```

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,671 @@
...
+
+/// Possible pairs of `AccessRelatedness` that can occur
+/// to two tags if one is a child of the other.
+/// If there are two pointers `x` and `y` and `x` is a strict child of `y`,
+/// then any access is either
+/// - a child of both (`ChildBoth`)
+/// - a child of `y` but foreign for `x` (`ChildY`)
+/// - foreign for both (`ChildNeither`)
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum AccessRel2XChildY {
```
RalfJung, line 118: This is new, right? Why does it show up now?

src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -280,14 +280,14 @@ pub(super) struct TbError<'node> {
 impl TbError<'_> {
     /// Produce a UB error.
     pub fn build<'tcx>(self) -> InterpError<'tcx> {
-        use TransitionError::*;
+        use TransitionError as TErr;
```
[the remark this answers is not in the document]
Vanille-N: Sorry, I thought I had resolved this.

src/borrow_tracker/tree_borrows/tree/tests.rs
Vanille-N: Yes this is new. While explaining some things above I realized that I didn't prove that this configuration works. It did, but it wasn't tested.

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,671 @@
...
+/// (which also means they are distinct), then any access is either
+/// - a child of `x` but foreign for `y` (`ChildX`)
+/// - a child of `y` but foreign for `x` (`ChildY`)
+/// - foreign for both (`Other`)
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum AccessRel2MutuallyForeign {
+    ChildX,
+    ChildY,
+    Other,
+}
```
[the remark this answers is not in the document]
Vanille-N: You're right that's better. I'm splitting this into `enum PtrSelector { X, Y, Other }` and `enum RelPosXY { MutuallyForeign, XChildY }`.

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -75,151 +75,109 @@ fn protected_enforces_noalias() {
...
+    Y,
     Other,
 }
 
-impl Exhaustive for AccessRel2MutuallyForeign {
+/// Relative position of `x` and `y`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum RelPosXY {
+    MutuallyForeign,
+    XChildY,
```
RalfJung, line 90:
```suggestion
    /// X is a child of Y.
    XChildY,
```

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -75,151 +75,109 @@ fn protected_enforces_noalias() {
...
+}
```
RalfJung, line 91: What bout the obvious 3rd case, `YChildX`?

Is it because X is definitely getting a retag, and hence cannot have children elsewhere? Please explain that in the doc comment. Really, this means you need to have a comment before all this that explains the set of patterns you are going to model, so that the reader has any chance of having the right context available.

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -75,151 +75,109 @@ fn protected_enforces_noalias() {
...
-}
-
-impl Into2Rel for AccessRel2MutuallyForeign {
-    const LEGEND: &'static str = "x and y are mutually foreign";
-    const THROUGH_X: Self = Self::ChildX;
-    const THROUGH_Y: Self = Self::ChildY;
-    const ALLOWS_X_RETAG_FIRST: bool = true;
-    fn into_pair(self) -> (AccessRelatedness, AccessRelatedness) {
+impl RelPosXY {
+    fn allows_x_retag_first(self) -> bool {
```
RalfJung, line 127: What does this function do/mean? Please add a comment.

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -75,151 +75,109 @@ fn protected_enforces_noalias() {
...
-                (AccessRelatedness::AncestorAccess, AccessRelatedness::StrictChildAccess),
-            AccessRel2XChildY::ChildNeither =>
-                (AccessRelatedness::DistantAccess, AccessRelatedness::DistantAccess),
+impl PtrSelector {
+    fn rel_pair(self, xy_rel: RelPosXY) -> (AccessRelatedness, AccessRelatedness) {
+        match xy_rel {
+            RelPosXY::MutuallyForeign =>
+                match self {
+                    PtrSelector::X =>
+                        (AccessRelatedness::StrictChildAccess, AccessRelatedness::DistantAccess),
```
RalfJung, line 141: Why not `This`? Seems more logical -- you are asking how x is related to `self` which is `X`.

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -75,151 +75,109 @@ fn protected_enforces_noalias() {
...
-    fn into_pair(self) -> (AccessRelatedness, AccessRelatedness) {
-        match self {
-            AccessRel2XChildY::ChildBoth =>
-                (AccessRelatedness::StrictChildAccess, AccessRelatedness::StrictChildAccess),
-            AccessRel2XChildY::ChildY =>
-                (AccessRelatedness::AncestorAccess, AccessRelatedness::StrictChildAccess),
-            AccessRel2XChildY::ChildNeither =>
-                (AccessRelatedness::DistantAccess, AccessRelatedness::DistantAccess),
+impl PtrSelector {
+    fn rel_pair(self, xy_rel: RelPosXY) -> (AccessRelatedness, AccessRelatedness) {
```
RalfJung, line 136: Please add doc comment explaining what this does.

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -75,151 +75,109 @@ fn protected_enforces_noalias() {
...
-        write!(f, "{kind_text} {}", self.rel2)
-    }
-}
-
 /// Arbitrary access parametrized by the relative position of `x` and `y`
 /// to each other.
 #[derive(Debug, Clone, Copy)]
-struct Access2<Rel2> {
-    rel2: Rel2,
+struct Access2 {
```
RalfJung, line 163: I think we should get rid of all of these "2" suffices, they are confusing. What about `TestAccess` and `TestEvent`?

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -326,33 +280,34 @@ struct LocStateProt2 {
...
+        evt: &Event2<RetX, RetY>,
+        xy_rel: RelPosXY,
     ) -> Result<Self, ()> {
         match evt {
-            Event2::Access2(acc2) => self.perform_access2(acc2),
-            Event2::RetX(_) => self.end_protector_x(),
-            Event2::RetY(_) => self.end_protector_y(),
+            Event2::Access2(acc2) => self.perform_access2(acc2, xy_rel),
+            Event2::RetX(_) => self.end_protector_x(xy_rel),
+            Event2::RetY(_) => self.end_protector_y(xy_rel),
```
RalfJung, line 310:
```suggestion
            // The fields of the Ret variants just serve to make them impossible
            // via the `!` type; we can ignore their values here.
            Event2::RetX(_) => self.end_protector_x(xy_rel),
            Event2::RetY(_) => self.end_protector_y(xy_rel),
```

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -372,11 +327,11 @@ impl fmt::Display for LocStateProt2 {
 /// Arbitrary sequence of accesses, as experienced by two mutually foreign pointers
 /// to the same location.
 #[derive(Debug, Clone)]
-struct Opaque2<Rel2, RetX, RetY> {
-    events: Vec<Event2<Rel2, RetX, RetY>>,
+struct Opaque2<RetX, RetY> {
```
RalfJung, line 330: `TestEvents`, maybe? or `OpaqueCode`.

src/borrow_tracker/tree_borrows/tree/tests.rs
RalfJung: Hm no, y also always gets retagged. Which difference can XChildY make?

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,629 @@
...
+/// This test in a generalization of `reserved_frozen_protected_distinguishable`
+/// that more exhaustively explores the feasability of inserting a spurious read.
+/// For a number of interleavings of different pieces of code involving
+/// two mutually foreign tags, we test if a spurious read through one
+/// changes the outcome (introduces UB in the target when there is none
+/// in the source).
+fn spurious_read() {
+    // The general code pattern being exhaustively tested is
+    //     [thread 1]             || [thread 2]
+    //     retag x (protect)      || retag y (protect/noprotect)
```
RalfJung, line 443: Why is this not just
```
retag x
<arbitrary code, some of it potentially on another thread>
optional spurious read
```
If `y` gets retagged before, that's part of the initial state. If `y` gets retagged after, that's part of the arbitrary code.

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,629 @@
...
+                for (x_initial_state, y_initial_state) in
+                    <(LocationState, LocationState)>::exhaustive()
+                {
+                    // We can only do spurious reads for initialized locations anyway.
+                    precondition!(x_initial_state.initialized);
+                    // Moreover we don't create `Active` or `Disabled` directly on reborrow.
+                    precondition!(!x_initial_state.permission.is_disabled());
+                    precondition!(!y_initial_state.permission.is_disabled());
+                    precondition!(!x_initial_state.permission.is_active());
+                    precondition!(!y_initial_state.permission.is_active());
```
RalfJung, line 481: These are the states *before* the retags, right? Why can't those be active/disabled?

Vanille-N: Those are the states that `x` and `y` should have immediately after their (optional) retag.

src/borrow_tracker/tree_borrows/tree/tests.rs
Vanille-N: In spirit yes, but I can't think of a way to ensure that this is exhaustive and efficient other than the current solution of limiting the state to exactly two observed pointers ?

src/borrow_tracker/tree_borrows/tree/tests.rs
Vanille-N: `y` does not always get retagged, there's an `if state.y.state.initialized { ... }` that accounts for the fact that the creation of `y` might not involve a read to that location.

All cases are covered:
- `y` raw pointer to previously initialized location  -> `y` initialized retagged before `x`
- `y` raw pointer, out-of-bounds of latest retag -> `y` uninitialized not retagged
- `y` reference, retagge -> `y` initialized retagged either before or after `x`
- `y` reference, out-of-bounds of retag -> `y` uninitialized not retagged

src/borrow_tracker/tree_borrows/tree/tests.rs
RalfJung: Oh, so they are not "initial" then, they are after-retag... that should be better reflected in comments and code. :)

src/borrow_tracker/tree_borrows/tree/tests.rs
RalfJung: What I don't get is why RelPosXY is independent of y_retag.  If you already consider both MutuallyForeign and XChildY and should be enough, no?

> there's an if state.y.state.initialized { ... } that accounts for the fact that the creation of y might not involve a read to that location.

Reads are already part of the unknown code though, so this seems redundant.

Vanille-N: There are 3 relevant configurations:
- `XChildY`
- `MutuallyForeign` with `x` older than `y`
- `MutuallyForeign` with `y` older than `x`
the latter two are not trivially guaranteed to be equivalent, because some cases could be eliminated by a read-on-reborrow.

"Reads are already part of the unknown code" yes but the unknown code also includes sequences with no read at all that could appear in the test as false negatives, I think.

src/borrow_tracker/tree_borrows/tree/tests.rs
Vanille-N: Sure, I can add a comment to make it explicit, but after-retag *is* initial since `x` and `y` don't even exist before the retag...

RalfJung: From all the comments I was expecting "initial state" to refer to the state at the beginning of the example code. After all there's `before_arbitrary_code` to go from the initial state to the beginning of the arbitrary code.

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,674 @@
...
+        where
+            RetX: Exhaustive + Clone + 'static,
+            RetY: Exhaustive + Clone + 'static,
+        {
+            // (state, code) left to handle
+            let mut handle = vec![(self, OpaqueCode { events: Vec::new() })];
+            // code that can be applied to `self`, and final state
+            let mut paths = Vec::new();
+            // already explored states reachable from `self`
+            let mut seen = FxHashSet::default();
```
RalfJung, line 413: Please add a type annotation, it's a bit hard to figure out which kinds of elements this set has.

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,674 @@
...
+                write!(f, "{evt}; ")?;
+            }
+            Ok(())
+        }
+    }
+
+    impl LocStateProtPair {
+        /// List all sequences of operations that lead to non-UB configurations,
+        /// without duplicates (not guaranteed to be the shortest sequence of events).
+        fn applicable_opaque_code_without_ub<RetX, RetY>(
```
RalfJung, line 400: Is this replacing all that previous stuff about only having a maximum of 4 events or so? I was wondering where that had gone.

Not sure if I understand this fn correctly, is this accurate?
```suggestion
    impl LocStateProtPair {
        /// List all sequences of operations start at `self` and do not cause UB
        /// without duplicates (not guaranteed to be the shortest sequence of events).
        /// Yields the states it reaches, and the sequence of operations that got us there.
        fn applicable_opaque_code_without_ub<RetX, RetY>(
```

Maybe rename it so `all_states_reachable_via_opaque_code` or so?

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,674 @@
...
+        for pat in Pattern::exhaustive() {
+            precondition!(pat.xy_rel.allows_x_retag_first() || !pat.x_retag_first);
+            let Ok(common_0) = pat.before_arbitrary_code() else {
+                continue;
+            };
+            // `x` must stay protected, but the function of `y` might return here
+            for (common_1, opaque) in
+                common_0.applicable_opaque_code_without_ub::<!, ()>(pat.xy_rel)
+            {
+                let Ok(source_post) =
```
RalfJung, line 628:
```suggestion
                // Both executions are identical up to here. Now, in the source we
                // run `after_arbitrary_code` *without* a spurious read, and in the target
                // we do it *with* a spurious read. The we check if the resulting states
                // are distinguishable.
                let Ok(source_post) =
```

src/borrow_tracker/tree_borrows/tree/tests.rs
RalfJung: So what about this: `OpaqueCode` also gets a `initial_y_retag: bool` field, to make it clear that we are treating this as part of the unknown code. Then it becomes part of `applicable_opaque_code_without_ub` to try things with and without the initial retag of `y`.

Do we have some argument for why doing the retag right at the start (and not in the middle) is sufficient?

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,674 @@
...
+    }
+
+    #[derive(Clone, Debug)]
+    struct Pattern {
+        xy_rel: RelPosXY,
+        x_retag_first: bool,
+        x_initial_state: LocationState,
+        y_initial_state: LocationState,
+        y_protected: bool,
+        y_return: bool,
```
RalfJung, line 496: `y_return` seems to be completely ignored except for printing?

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,674 @@
...
+        let mut ok = 0;
+        let mut err = 0;
+        for pat in Pattern::exhaustive() {
+            precondition!(pat.xy_rel.allows_x_retag_first() || !pat.x_retag_first);
+            let Ok(common_0) = pat.before_arbitrary_code() else {
+                continue;
+            };
+            // `x` must stay protected, but the function of `y` might return here
+            for (common_1, opaque) in
+                common_0.applicable_opaque_code_without_ub::<!, ()>(pat.xy_rel)
```
RalfJung, line 626: So... is `RetX` always `!` and `RetY` always `()`? Then why make all that code generic? Also further down the look I see `::<(), ()>`, why does that use a different `RetX`? Definitely all of these need comments explaining wht we are picking the types we are picking. This is very implicitly making some super important choices!

src/borrow_tracker/tree_borrows/tree/tests.rs
Vanille-N: Yes, exactly.

This replaces the previous enumeration of sequences of 4 events because computing the reachable set from one `LocStateProtPair` is (unsurprisingly) much more efficient than returning a set of `OpaqueCode` that is known to reach all states from all starting configurations.

src/borrow_tracker/tree_borrows/tree/tests.rs
Vanille-N: Oh yeah, everything that `y_return` was in charge of is now handled by `RetY` on `distinguishable`.

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,681 @@
...
+        let mut ok = 0;
+        let mut err = 0;
+        for pat in Pattern::exhaustive() {
+            precondition!(pat.xy_rel.allows_x_retag_first() || !pat.x_retag_first);
+            let Ok(common_0) = pat.before_arbitrary_code() else {
+                continue;
+            };
+            // `x` must stay protected, but the function of `y` might return here
+            for (common_1, opaque) in
+                common_0.all_states_reachable_via_opaque_code::<NoRet, AllowRet>(pat.xy_rel)
```
RalfJung, line 629:
```suggestion
                common_0.all_states_reachable_via_opaque_code::</*RetX*/ NoRet, /*RetY*/ AllowRet>(pat.xy_rel)
```

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,681 @@
...
+                    eprintln!("  (arbitrary code instanciated with '{}')", opaque);
+                    eprintln!(
+                        "  (without the spurious read the final state was '{}')",
+                        source_post
+                    );
+                    err += 1;
+                    continue;
+                };
+                // Now `x` might lose its protector
+                if source_post.distinguishable::<AllowRet, AllowRet>(&target_post, pat.xy_rel) {
```
RalfJung, line 660:
```suggestion
                if source_post.distinguishable::</*RetX*/ AllowRet, /*RetX*/ AllowRet>(&target_post, pat.xy_rel) {
```

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -245,6 +242,7 @@ mod spurious_read {
         Access(TestAccess),
         RetX(RetX),
         RetY(RetY),
+        RetagY(LocStateProt),
```
RalfJung, line 245:
```suggestion
        /// LocStateProt must be an initial state (as per `state.is_initial()`).
        RetagY(LocStateProt),
```

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,645 @@
...
+            write!(f, "{}", self.state)?;
+            if self.prot {
+                write!(f, ", protect")?;
+            }
+            Ok(())
+        }
+    }
+
+    #[derive(Clone, PartialEq, Eq, Hash)]
+    /// The states of two pointers to the same location.
```
RalfJung, line 328:
```suggestion
    /// The states of two pointers to the same location,
    /// and their relationship to each other in the tree.
```

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,645 @@
...
+
+    impl LocStateProtPair {
+        fn perform_test_access(self, acc: &TestAccess) -> Result<Self, ()> {
+            let (xrel, yrel) = acc.ptr.rel_pair(self.xy_rel);
+            let x = self.x.perform_access(acc.kind, xrel)?;
+            let y = self.y.perform_access(acc.kind, yrel)?;
+            Ok(Self { x, y, ..self })
+        }
+
+        fn read_if_initialized(self, ptr: PtrSelector) -> Result<Self, ()> {
```
RalfJung, line 351: I think this should panic if called on `Other`, it makes little sense there. Also it needs a doc comment.
```
/// Perform a read on the given pointer if its state is `initialized`.
/// Must be called just after reborrowing a pointer.
```

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,645 @@
...
+            Ok(Self { x, ..self })
+        }
+
+        fn end_protector_y(self) -> Result<Self, ()> {
+            let y = self.y.end_protector();
+            Ok(Self { y, ..self })
+        }
+
+        fn retag_y(self, new_y: LocStateProt) -> Result<Self, ()> {
+            Self { y: new_y, xy_rel: RelPosXY::MutuallyForeign, ..self }
```
RalfJung, line 375:
```suggestion
            assert!(new_y.is_initial());
            // `xy_rel` changes to "mutually foreign" now: `y` can no longer be a parent of `x`.
            Self { y: new_y, xy_rel: RelPosXY::MutuallyForeign, ..self }
```

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,645 @@
...
+        }
+    }
+
+    impl fmt::Display for LocStateProtPair {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            write!(f, "x:{}, y:{}", self.x, self.y)
+        }
+    }
+
+    /// Arbitrary sequence of accesses, as experienced by two mutually foreign pointers
```
RalfJung, line 407:
```suggestion
    /// Arbitrary sequence of events, as experienced by two mutually foreign pointers
```

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,645 @@
...
+        let target = source.clone().perform_test_access(&acc).unwrap();
+        assert!(source.y.state.permission.is_reserved(None));
+        assert!(target.y.state.permission.is_frozen());
+        assert!(!source.distinguishable::<(), ()>(&target))
+    }
+
+    #[derive(Clone, Debug)]
+    struct Pattern {
+        /// The relative position of `x` and `y` at the beginning of the pattern.
+        /// Might change if the opaque code contains any `retag y`.
```
RalfJung, line 524:
```suggestion
        /// The relative position of `x` and `y` at the beginning of the arbitrary code
        /// (i.e., just after `x` got created).
        /// Might change if the opaque code contains any `retag y`.
```

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,645 @@
...
+        xy_rel: RelPosXY,
+        /// Permission that `x` will be retagged as (always protected until
```
RalfJung, line 526:
```suggestion
        /// Permission that `x` will be created as (always protected until
```

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,645 @@
...
+        /// a possible `ret x` is the second phase).
+        x_retag_perm: LocationState,
+        /// Permission that `y` has at the beginning of the pattern.
+        /// Might by reset if the opaque code contains any `retag y`.
```
RalfJung, line 530:
```suggestion
        /// Might by reset if the opaque code contains any `retag y`.
        /// Can be any state, not necessarily initial
        /// (since `y` exists already before the pattern starts).
```

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,645 @@
...
+                vec![(self, OpaqueCode { events: Vec::new() })];
+            // Code that can be applied to `self`, and final state.
+            let mut paths: Vec<(Self, OpaqueCode<_, _>)> = Default::default();
+            // Already explored states reachable from `self`
+            let mut seen: FxHashSet<Self> = Default::default();
+            // This is essentially just computing the transitive closure by `perform_test_event`,
+            // most of the work lies in remembering the path up to the current state.
+            while let Some((state, path)) = handle.pop() {
+                for evt in <TestEvent<RetX, RetY>>::exhaustive() {
+                    if let Ok(next) = state.clone().perform_test_event(&evt) {
```
RalfJung, line 450: This would work without the `clone()` because `LocStareRel` is `Copy`, right?

Please either remove the `Copy` and use `Clone` consistently, or just exploit the implicit copy.

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,645 @@
...
+        fn exhaustive() -> Box<dyn Iterator<Item = Self>> {
+            let mut v = Vec::new();
+            for xy_rel in RelPosXY::exhaustive() {
+                for (x_retag_perm, y_initial_state) in
+                    <(LocationState, LocationState)>::exhaustive()
+                {
+                    // We can only do spurious reads for initialized locations anyway.
+                    precondition!(x_retag_perm.initialized);
+                    // And we don't allow `x` to start as `Disabled` because it's protected.
+                    precondition!(!x_retag_perm.permission.is_disabled());
```
RalfJung, line 547: Why doesn't this require `x_retag_perm.is_initial()`?

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,645 @@
...
+    /// in the target but not in the source.
+    fn test_all_patterns() {
+        let mut ok = 0;
+        let mut err = 0;
+        for pat in Pattern::exhaustive() {
+            let Ok(initial_source) = pat.initial_state() else {
+                // Failed to retag `x` in the source (e.g. `y` was protected Active)
+                continue;
+            };
+            // `x` must stay protected, but the function of `y` might return here
```
RalfJung, line 601:
```suggestion
            // `x` must stay protected, but the function protecting `y` might return here
```

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -0,0 +1,645 @@
...
+    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+    enum PtrSelector {
+        X,
+        Y,
+        Other,
+    }
+
+    /// Relative position of `x` and `y`.
+    /// `y` cannot be a child of `x` because the compiler won't blindly
+    /// insert spurious reads through `x` if it has children.
```
RalfJung, line 128:
```suggestion
    /// `y` cannot be a child of `x` because `x` gets retagged as the first thing in the
    /// pattern, so it cannot have children.
```

src/borrow_tracker/tree_borrows/tree/tests.rs
Vanille-N: `LocationState` is `Copy`, but `LocStateProt` and `LocStateProtPair` are not.
This one is a `LocStateProtPair`.
Should I make them `Copy` ?

RalfJung: Ah, I misjudged the type.

No it's fine to clone explicitly (makes it more clear that we are not mutating `state`). Just don't call `clone()` on things that are `Copy`.

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -538,15 +552,14 @@ mod spurious_read {
...
             for xy_rel in RelPosXY::exhaustive() {
-                for (x_retag_perm, y_initial_state) in
-                    <(LocationState, LocationState)>::exhaustive()
+                for (x_retag_perm, y_current_perm) in <(LocationState, LocationState)>::exhaustive()
                 {
                     // We can only do spurious reads for initialized locations anyway.
                     precondition!(x_retag_perm.initialized);
-                    // And we don't allow `x` to start as `Disabled` because it's protected.
-                    precondition!(!x_retag_perm.permission.is_disabled());
+                    // And `x` just retagged must be initial.
```
RalfJung, line 559:
```suggestion
                    // And `x` just got retagged, so it must be initial.
```

RalfJung:
Just one last comment nit. :) Then please squash the commits and we're good to go.

RalfJung:
r=me after squashing and fixing that comment.
@bors delegate+

PR #3055, opened 2023-09-08
Implement some `llvm.x86.sse2.*` intrinsics and add tests 

eduardosm:
Continuation of https://github.com/rust-lang/miri/pull/2989 with SSE2 intrinsics.

Thankfully, a significant amount of SSE2 functions use `simd_*` intrinsics, which are already implemented in Miri.

### DIFF
--- a/src/shims/x86/sse2.rs
+++ b/src/shims/x86/sse2.rs
@@ -1,0 +1,982 @@
+use rustc_apfloat::{
+    ieee::{Double, Single},
+    Float as _, FloatConvert as _,
+};
+use rustc_middle::ty::layout::LayoutOf as _;
+use rustc_middle::ty::Ty;
+use rustc_span::Symbol;
+use rustc_target::abi::Size;
+use rustc_target::spec::abi::Abi;
+
+use super::FloatCmpOp;
+use crate::*;
+use shims::foreign_items::EmulateByNameResult;
+
+impl<'mir, 'tcx: 'mir> EvalContextExt<'mir, 'tcx> for crate::MiriInterpCx<'mir, 'tcx> {}
+pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
+    fn emulate_x86_sse2_intrinsic(
+        &mut self,
+        link_name: Symbol,
+        abi: Abi,
+        args: &[OpTy<'tcx, Provenance>],
+        dest: &PlaceTy<'tcx, Provenance>,
+    ) -> InterpResult<'tcx, EmulateByNameResult<'mir, 'tcx>> {
+        let this = self.eval_context_mut();
+        // Prefix should have already been checked.
+        let unprefixed_name = link_name.as_str().strip_prefix("llvm.x86.sse2.").unwrap();
+
+        // These intrinsics operate on 128-bit (f32x4, f64x2, i8x16, i16x8, i32x4, i64x2) SIMD
+        // vectors unless stated otherwise.
+        // Many intrinsic names are sufixed with "ps" (packed single), "ss" (scalar signle),
+        // "pd" (packed double) or "sd" (scalar double), where single means single precision
+        // floating point (f32) and double means double precision floating point (f64). "ps"
+        // and "pd" means thet the operation is performed on each element of the vector, while
+        // "ss" and "sd" means that the operation is performed only on the first element, copying
+        // the remaining elements from the input vector (for binary operations, from the left-hand
+        // side).
+        // Intrinsincs sufixed with "epiX" or "epuX" operate with X-bit signed or unsigned
+        // vectors.
+        match unprefixed_name {
+            // Used to implement the _mm_avg_epu8 function.
+            // Averages packed unsigned 8-bit integers in `left` and `right`.
+            "pavg.b" => {
+                let [left, right] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let (left, left_len) = this.operand_to_simd(left)?;
+                let (right, right_len) = this.operand_to_simd(right)?;
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                assert_eq!(dest_len, left_len);
+                assert_eq!(dest_len, right_len);
+
+                for i in 0..dest_len {
+                    let left = this.read_scalar(&this.project_index(&left, i)?)?.to_u8()?;
+                    let right = this.read_scalar(&this.project_index(&right, i)?)?.to_u8()?;
+                    let dest = this.project_index(&dest, i)?;
+
+                    // Values are expanded from u8 to u16, so adds cannot overflow.
+                    let res = u16::from(left)
+                        .checked_add(u16::from(right))
+                        .unwrap()
+                        .checked_add(1)
+                        .unwrap()
+                        / 2;
+                    this.write_scalar(Scalar::from_u8(res.try_into().unwrap()), &dest)?;
+                }
+            }
+            // Used to implement the _mm_avg_epu16 function.
+            // Averages packed unsigned 16-bit integers in `left` and `right`.
+            "pavg.w" => {
+                let [left, right] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let (left, left_len) = this.operand_to_simd(left)?;
+                let (right, right_len) = this.operand_to_simd(right)?;
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                assert_eq!(dest_len, left_len);
+                assert_eq!(dest_len, right_len);
+
+                for i in 0..dest_len {
+                    let left = this.read_scalar(&this.project_index(&left, i)?)?.to_u16()?;
+                    let right = this.read_scalar(&this.project_index(&right, i)?)?.to_u16()?;
+                    let dest = this.project_index(&dest, i)?;
+
+                    // Values are expanded from u16 to u32, so adds cannot overflow.
+                    let res = u32::from(left)
+                        .checked_add(u32::from(right))
+                        .unwrap()
+                        .checked_add(1)
+                        .unwrap()
+                        / 2;
+                    this.write_scalar(Scalar::from_u16(res.try_into().unwrap()), &dest)?;
+                }
+            }
+            // Used to implement the _mm_mulhi_epi16 function.
+            "pmulh.w" => {
+                let [left, right] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let (left, left_len) = this.operand_to_simd(left)?;
+                let (right, right_len) = this.operand_to_simd(right)?;
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                assert_eq!(dest_len, left_len);
+                assert_eq!(dest_len, right_len);
+
+                for i in 0..dest_len {
+                    let left = this.read_scalar(&this.project_index(&left, i)?)?.to_i16()?;
+                    let right = this.read_scalar(&this.project_index(&right, i)?)?.to_i16()?;
+                    let dest = this.project_index(&dest, i)?;
+
+                    // Values are expanded from i16 to i32, so multiplication cannot overflow.
+                    let res = i32::from(left).checked_mul(i32::from(right)).unwrap() >> 16;
+                    this.write_scalar(Scalar::from_int(res, Size::from_bits(16)), &dest)?;
+                }
+            }
+            // Used to implement the _mm_mulhi_epu16 function.
+            "pmulhu.w" => {
+                let [left, right] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let (left, left_len) = this.operand_to_simd(left)?;
+                let (right, right_len) = this.operand_to_simd(right)?;
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                assert_eq!(dest_len, left_len);
+                assert_eq!(dest_len, right_len);
+
+                for i in 0..dest_len {
+                    let left = this.read_scalar(&this.project_index(&left, i)?)?.to_u16()?;
+                    let right = this.read_scalar(&this.project_index(&right, i)?)?.to_u16()?;
+                    let dest = this.project_index(&dest, i)?;
+
+                    // Values are expanded from u16 to u32, so multiplication cannot overflow.
+                    let res = u32::from(left).checked_mul(u32::from(right)).unwrap() >> 16;
+                    this.write_scalar(Scalar::from_u16(res.try_into().unwrap()), &dest)?;
+                }
+            }
+            // Used to implement the _mm_mul_epu32 function.
+            // Multiplies the the low unsigned 32-bit integers from each packed
+            // 64-bit element and stores the result as 64-bit unsigned integers.
+            "pmulu.dq" => {
+                let [left, right] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let (left, left_len) = this.operand_to_simd(left)?;
+                let (right, right_len) = this.operand_to_simd(right)?;
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                // left and right are u32x4, dest is u64x2
+                assert_eq!(left_len, 4);
+                assert_eq!(right_len, 4);
+                assert_eq!(dest_len, 2);
+
+                for i in 0..dest_len {
+                    let op_i = i.checked_mul(2).unwrap();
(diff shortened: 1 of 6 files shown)


tests/pass/intrinsics-x86-sse2.rs
```diff
@@ -0,0 +1,742 @@
+//@only-target-x86_64
```
RalfJung, line 1: I now realize this also applies to your other test file, but -- shouldn't we also test those intrinsics on 32bit x86?

src/shims/x86/sse2.rs
```diff
@@ -0,0 +1,1026 @@
...
+                assert_eq!(dest_len, left_len);
+                assert_eq!(dest_len, right_len);
+
+                for i in 0..dest_len {
+                    let left = this.read_scalar(&this.project_index(&left, i)?)?.to_u8()?;
+                    let right = this.read_scalar(&this.project_index(&right, i)?)?.to_u8()?;
+                    let dest = this.project_index(&dest, i)?;
+
+                    // Values are expanded from u8 to u16, so adds cannot overflow.
+                    let res = u16::from(left).wrapping_add(u16::from(right)).wrapping_add(1) >> 1;
```
RalfJung, line 56: What's this `>> 1` doing? If this is division, please use `/ 2` to make the intent clear.

src/shims/x86/sse2.rs
```diff
@@ -0,0 +1,1026 @@
...
+                assert_eq!(dest_len, left_len);
+                assert_eq!(dest_len, right_len);
+
+                for i in 0..dest_len {
+                    let left = this.read_scalar(&this.project_index(&left, i)?)?.to_u16()?;
+                    let right = this.read_scalar(&this.project_index(&right, i)?)?.to_u16()?;
+                    let dest = this.project_index(&dest, i)?;
+
+                    // Values are expanded from u16 to u32, so adds cannot overflow.
+                    let res = u32::from(left).wrapping_add(u32::from(right)).wrapping_add(1) >> 1;
```
RalfJung, line 79: Whenever you are asserting that things cannot overflow, please use `checked_$OP().unwrap()` so that the code actually asserts that the comment is true.

src/shims/x86/sse2.rs
```diff
@@ -0,0 +1,1026 @@
...
+                    let right = this.read_scalar(&this.project_index(&right, i)?)?.to_i16()?;
+                    let dest = this.project_index(&dest, i)?;
+
+                    // Values are expanded from i16 to i32, so multiplication cannot overflow.
+                    let res = i32::from(left).wrapping_mul(i32::from(right)) >> 16;
+                    this.write_scalar(Scalar::from_int(res, Size::from_bits(16)), &dest)?;
+                }
+            }
+            // Used to implement the _mm_mulhi_epu16 function.
+            "pmulhu.w" => {
```
RalfJung, line 106: I wonder if you could de-duplicate some code here by sharing this with `pmulh.w`, and using operations on `ImmTy` (as produced by `read_immediate`) instead of host integer arithmetic? We already have all arithmetic operations implemented in `overflowing_binary_op` after all.

This applies throughout the entire file. There seems to be a lot of use of host integers here and that's generally surprising. (And same for floats, `overflowing_binary_op` will do the right thing on immediates of float type as well.)

tests/pass/intrinsics-x86-sse2.rs
eduardosm: Is there a directive to specify more than one target?

RalfJung: Hm, multiple "only" are interpreted conjunctively so that won't work.

But its' a substring test, so `only-target-x86` should work.

src/shims/x86/sse2.rs
RalfJung: Thanks for trying this out! Looks like the shared version is overall longer than the 2 separate versions... so I guess it's not worth it.

Are there other sets of intrinsics that are completely the same except for the types involved?

RalfJung: Like, for instance pavg also exists twice. With some helper methods for widening casts, could this be useful?

"No I think that would introduce more overhead than it helps" is a reasonable answer here, I'm just asking you since it's hard to say for myself without actually trying to write that code. :)

src/shims/x86/sse2.rs
```diff
@@ -0,0 +1,1042 @@
...
+                    Sra,
+                }
+                let which = match unprefixed_name {
+                    "psll.w" => ShiftOp::Sll,
+                    "psrl.w" => ShiftOp::Srl,
+                    "psra.w" => ShiftOp::Sra,
+                    _ => unreachable!(),
+                };
+
+                // Build the 64-bit shift operand from 16-bit elements
```
RalfJung, line 238: Is this equivalent to transmuting `right` to a `u64`? Or does endianess mean this is not the same?

src/shims/x86/sse2.rs
```diff
@@ -0,0 +1,1042 @@
...
+                    | (u64::from(this.read_scalar(&this.project_index(&right, 3)?)?.to_u16()?)
+                        << 48);
+
+                for i in 0..dest_len {
+                    let left = this.read_scalar(&this.project_index(&left, i)?)?.to_u16()?;
+                    let dest = this.project_index(&dest, i)?;
+
+                    let res = match which {
+                        ShiftOp::Sll =>
+                            left.checked_shl(shift.try_into().unwrap_or(u32::MAX)).unwrap_or(0),
```
RalfJung, line 253: The subexpression `shift.try_into().unwrap_or(u32::MAX)` occurs several times, that could be computed (and named) once before the loop.

src/shims/x86/sse2.rs
```diff
@@ -0,0 +1,1042 @@
...
+                        ShiftOp::Srl =>
+                            left.checked_shr(shift.try_into().unwrap_or(u32::MAX)).unwrap_or(0),
+                        #[allow(clippy::cast_possible_wrap, clippy::cast_sign_loss)]
+                        ShiftOp::Sra => {
+                            // Convert u16 to i16 to use arithmetic shift
+                            let left = left as i16;
+                            left.checked_shr(shift.try_into().unwrap_or(u32::MAX))
+                                .unwrap_or(left >> 15) as u16
```
RalfJung, line 261: What the `left >> 15` doing here? Is that the "copying the sign bit to the remaining bits"? Please add a comment, such random integer constants are always suspicious.

src/shims/x86/sse2.rs
```diff
@@ -0,0 +1,1042 @@
...
+            "sqrt.sd" => {
+                let [op] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let (op, op_len) = this.operand_to_simd(op)?;
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                assert_eq!(dest_len, op_len);
+
+                let op0 = this.read_scalar(&this.project_index(&op, 0)?)?.to_u64()?;
+                let res0 = Scalar::from_u64(f64::from_bits(op0).sqrt().to_bits());
```
RalfJung, line 555: Needs the host float FIXME.

src/shims/x86/sse2.rs
```diff
@@ -0,0 +1,1042 @@
...
+            }
+            // Used to implement the _mm_cmp*_sd function.
+            // Performs a comparison operation on the first component of `left`
+            // and `right`, returning 0 if false or `u64::MAX` if true. The remaining
+            // components are copied from `left`.
+            "cmp.sd" => {
+                let [left, right, imm] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let which = match this.read_scalar(imm)?.to_i8()? {
```
RalfJung, line 594: This exact match already exists for `cmp.ss`; looks like unnecessary duplication.

src/shims/x86/sse2.rs
```diff
@@ -0,0 +1,1042 @@
...
+
+                // op is f64x2, dest is f32x4
+                assert_eq!(op_len, 2);
+                assert_eq!(dest_len, 4);
+
+                for i in 0..op_len {
+                    let op = this.read_scalar(&this.project_index(&op, i)?)?.to_f64()?;
+                    let dest = this.project_index(&dest, i)?;
+
+                    let res = op.convert(/*loses_info*/ &mut false).value;
```
RalfJung, line 687: Can't this use `float_to_float_or_int`, and then the code is completely identical between cvtpd2ps and cvtps2pd?

src/shims/x86/sse2.rs
```diff
@@ -0,0 +1,1042 @@
...
+                    this.copy_op(
+                        &this.project_index(&left, i)?,
+                        &this.project_index(&dest, i)?,
+                        /*allow_transmute*/ false,
+                    )?;
+                }
+            }
+            // Used to implement the _mm_cvtss_sd function.
+            // Converts the first f64 from `right` to f32 and copies
+            // the remaining elements from `left`
```
RalfJung, line 838: This has the exact same comment as cvtsd2ss? There should be some difference, no?

src/shims/x86/sse2.rs
```diff
@@ -0,0 +1,1042 @@
...
+            "movmsk.pd" => {
+                let [op] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let (op, op_len) = this.operand_to_simd(op)?;
+
+                let mut res = 0;
+                for i in 0..op_len {
+                    let op = this.read_scalar(&this.project_index(&op, i)?)?;
+                    let op = op.to_u64()?;
+
+                    res |= (op >> 63) << i;
```
RalfJung, line 879: Isn't there a nicer way to say "highest bit"? At least please add a comment saying that this is the highest bit.

src/shims/x86/sse2.rs
```diff
@@ -0,0 +1,1042 @@
...
+            &this.project_index(&dest, i)?,
+            /*allow_transmute*/ false,
+        )?;
+    }
+
+    Ok(())
+}
+
+/// Performs `which` operation on each component of `left`, and
+/// `right` storing the result is stored in `dest`.
```
RalfJung, line 1017:
```suggestion
/// Performs `which` operation on each component of `left` and
/// `right`, storing the result is stored in `dest`.
```
(also in the SSE file, while we are at it)

src/shims/x86/sse2.rs
```diff
@@ -0,0 +1,1042 @@
...
+    /// Maximum value (with SSE semantics)
+    ///
+    /// <https://www.felixcloutier.com/x86/maxsd>
+    /// <https://www.felixcloutier.com/x86/maxpd>
+    Max,
+}
+
+/// Performs `which` scalar operation on `left` and `right` and returns
+/// the result.
+fn bin_op_f64<'tcx>(
```
RalfJung, line 927: I was hoping this could be shared with the f32 code, in a single function that is generic over the apfloat type. What do we need in terms of infrastructure for that? I think the one problem is converting from `ImmTy` to the float type, right? In binary_float_op in Miri, they come pre-converted. So maybe leave a FIXME, this could make a nice next PR: make sure we have `ImmTy: TryInto<Single>` and `ImmTy: TryInto<Double>`, and then use that to make the float handling more generic.

tests/pass/intrinsics-x86-sse2.rs
```diff
@@ -0,0 +1,742 @@
...
+    unsafe fn test_mm_max_pd() {
+        let a = _mm_setr_pd(1.0, 2.0);
+        let b = _mm_setr_pd(5.0, 10.0);
+        let r = _mm_max_pd(a, b);
+        assert_eq_m128d(r, _mm_setr_pd(5.0, 10.0));
+    }
+    test_mm_max_pd();
+
+    #[target_feature(enable = "sse2")]
+    unsafe fn test_mm_min_sd() {
```
RalfJung, line 257: Can you make sure the min/max functions have the same tests that the SSE(1) variants do, checking the handling of negative zero?

src/shims/x86/sse2.rs
eduardosm: It is equivalent to a transmute after discarding the high 64 bits of the SIMD vector. Is there a nicer way of doing that?

src/shims/x86/sse2.rs
eduardosm: It might be worth trying it out, almost half of the loop body is used by the widening casts.

But I would leave it for the future PR you mentioned below (the one about making `bin_op_f32`/`bin_op_f64` generic).

src/shims/x86/sse2.rs
RalfJung: Ah, it's only loading the first half? Hm... you could transmute to a `[u62; 2]` array, project to an index, and then load from that. Not sure if that's any better than what you currently do.

src/shims/x86/sse2.rs
```diff
@@ -243,22 +238,22 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                         << 32)
                     | (u64::from(this.read_scalar(&this.project_index(&right, 3)?)?.to_u16()?)
                         << 48);
+                // Convert it to the type expected by checked_{shl,shr}
+                let shift = shift.try_into().unwrap_or(u32::MAX);
```
RalfJung, line 242: Why is it okay to fall back to u32::MAX here? Please add a comment.
Do we have a test covering this case?

RalfJung:
@rutsbot author

tests/pass/intrinsics-x86-sse2.rs
eduardosm: Does the directive check the target triple (x86_64-unknown-linux-gnu, i686-unknown-linux-gnu) or the target arch name (x86_64, x86)?

RalfJung: It's a substring match on the entire triple.

src/shims/x86/sse2.rs
eduardosm: Tests could definitely be improved. I probably will contribute them also to stdarch.

src/shims/x86/sse2.rs
```diff
@@ -829,6 +829,23 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
     }
 }
 
+/// Takes a 128-bit vector, transmutes it to `[u64; 2]` and extracts
+/// the fiest value.
```
RalfJung, line 834:
```suggestion
/// the first value.
```

src/shims/x86/sse2.rs
```diff
@@ -829,6 +829,23 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+fn extract_first_u64<'tcx>(
+    this: &crate::MiriInterpCx<'_, 'tcx>,
+    op: &MPlaceTy<'tcx, Provenance>,
+) -> InterpResult<'tcx, u64> {
+    // Transmute vector to `[u64; 2]`
+    let u64_array_layout = this
+        .tcx
+        .layout_of(ty::ParamEnv::empty().and(Ty::new_array(this.tcx.tcx, this.tcx.types.usize, 2)))
```
RalfJung, line 842:
```suggestion
    let u64_array_layout = this
        .layout_of(Ty::new_array(this.tcx.tcx, this.tcx.types.usize, 2))
```

eduardosm: I knew I was making that too complex :)

src/shims/x86/sse2.rs
```diff
@@ -1,8 +1,9 @@
+use crate::rustc_middle::ty::layout::LayoutOf as _;
```
RalfJung, line 1: Please keep the imports sorted:
- std
- 3rd-party crates
- rustc crates
- local imports

tests/pass/intrinsics-x86-sse.rs
```diff
@@ -997,6 +997,7 @@ unsafe fn test_sse() {
     }
     test_mm_cvtsi32_ss();
 
+    #[cfg(target_arch = "x86_64")]
```
RalfJung, line 1000: Why are some of these only enabled on x86_64? Please add comments.

RalfJung:
All right, looking great overall! Please squash into <4 commits, then we can land this. :)

eduardosm:
Ready!

eduardosm:
Remainder to myself:

* Implement `bin_op_f32`/`bin_op_f64` with generics to reduce code duplication.
* Figure out a widening conversion helper function to implement `avg` and `mulh` without host integer arithmetic.

I will work on that once this lands in the main Rust repo because I expect to add helpers to `rustc_middle`.

RalfJung:
Thanks. :)
@bors r+

> I will work on that once this lands in the main Rust repo because I expect to add helpers to rustc_middle.

I expect it will be rustc_const_eval, but it will be on the rustc side either way.

RalfJung:
@eduardosm his is synced into rustc now.

PR #3056, opened 2023-09-08
Use `#!/usr/bin/env` shebang

Nilstrieb:
```
$ ls /bin/bash
ls: cannot access '/bin/bash': No such file or directory
```

On certain systems, `/bin` and `/usr/bin` are nothing but wastelands, with just `env`around as the last survivor of the great purge. The binaries have cowardly hidden away and only `env` can show us the way to greatness (bash). ❄️

### DIFF
--- a/miri
+++ b/miri
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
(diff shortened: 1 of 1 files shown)


RalfJung:
Ah yeah I keep forgetting about that. ^^ Thanks!
@bors r+

PR #3062, opened 2023-09-16
triagebot exclude_labels -> exclude_titles

pitaj:
https://github.com/rust-lang/triagebot/pull/1720

### DIFF
--- a/miri-script/src/commands.rs
+++ b/miri-script/src/commands.rs
@@ -339,9 +339,9 @@ impl Command {
             "Confirmed that the push round-trips back to Miri properly. Please create a rustc PR:"
         );
         println!(
-            // Open PR with `subtree-sync` label to satisfy the `no-merges` triagebot check
+            // Open PR with `subtree update` title to silence the `no-merges` triagebot check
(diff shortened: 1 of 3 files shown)


miri-script/src/commands.rs
```diff
@@ -339,9 +339,9 @@ impl Command {
             "Confirmed that the push round-trips back to Miri properly. Please create a rustc PR:"
         );
         println!(
-            // Open PR with `subtree-sync` label to satisfy the `no-merges` triagebot check
+            // Open PR with `subtree update` title to satisfy the `no-merges` triagebot check
```
RalfJung, line 342:
```suggestion
            // Open PR with `subtree update` title to silence the `no-merges` triagebot check
```

RalfJung:
Thanks!
Does this have to wait until https://github.com/rust-lang/triagebot/pull/1720 lands and gets deployed? (Not sure how triagebot deals with unknown options.)

pitaj:
It does not need to wait. Extra options in the config are ignored.

PR #3067, opened 2023-09-20
Continuation of #3054: enable spurious reads in TB

Vanille-N:
The last additions to the test suite of TB left some unresolved `#[should_panic]` that these new modifications solve.

## Problem

Recall that the issues were arising from the interleavings that follow.

### A. `Reserved -> Frozen` has visible effects after function exit

The transition `Reserved -> Frozen` irreversibly blocks write accesses to the tag, so in the interleaving below `y` initially `Reserved` becomes `Frozen` only in the target where a spurious read through `x` is inserted. This makes the later write through `y` UB only in the target and not in the source.
```
1: retag x (&, protect)
2: retag y (&mut, protect)
1: spurious read x
1: ret x
2: ret y
2: write y
```

### B. Protectors only announce their presence on retag

There is a read-on-reborrow for protected locations, but if the retag of `x` occurs before that of `y` and there is no explicit access through `x`, then `y` is unaware of the existence of `x`. This is problematic because a spurious read inserted through `x` between the retag of `y` and the return of the function protecting `x` is a noalias violation in the target without UB in the source.
```
1: retag x (&, protect)
2: retag y (&mut, protect)
1: spurious read x
1: ret x
2: write y
2: ret y
```

## Step 1: Finer behavior for `Reserved`

Since one problem is that `Reserved -> Frozen` has consequences beyond function exit, we decide to remove this transition entirely. To replace it we introduce a new subtype of `Reserved` with the extra boolean `aliased` set.
`Reserved { aliased: true }` forbids child accesses, but only temporarily: it has no effect on activation once the tag is no longer protected.
This makes the semantics of Tree Borrows slightly weaker in favor of being more similar to noalias.

This solves interleaving **A.**, but **B.** is still a problem and the exhaustive tests do not pass yet.

## Step 2: Read on function exit

Protected tags issue a "reminder" that they are protected until this instant inclusive, in the form of an implicit read (symmetrically to the implicit read on retag). This ensures that if the periods on which two tags `x` and `y` are protected overlap then no matter the interleaving of retags and returns, there is either a protector currently active or a read that has been emitted, both of which temporarily block activation.

This makes the exhaustive test designed previously pass, but it has an effect on the ability to return an activated pointer that I had not foreseen before implementing it.

## Step 2': Do not propagate to children

A naive implementation of **Step 2** makes the following code UB:
```rs
fn reborrow(x: &mut u8) -> &mut u8 {
    let y = &mut *x;
    *y = *y;
    y // callee returns `y: Active`...
}

let x = &mut 0u8;
let y = reborrow(x); // ... and caller receives `y: Frozen`
*y = 1; // UB
```
This is unacceptable, and a simple fix is to make this implicit read visible only to foreign tags.

We still lack hindsight on the ramifications of this decision, and the fact that the problematic pattern was only discovered because it occured in one completely unrelated test (with a cryptic error message) is worrying. We should be vigilant as to how this interacts with the rest of the model.

## TODO

As of commit #281c30, the data race model has not been fully updated.
We have removed the reborrow of mutable references counting as a write access, but we still need the implicit read of function exit to count as a read.


src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -71,24 +79,20 @@ mod transition {
...
-    /// A non-child node was read-accessed: noop on non-protected Reserved, advance to Frozen otherwise.
+    /// A non-child node was read-accessed: keep `Reserved` but mark it as `aliased` if it
+    /// is protected; invalidate `Active`.
     fn foreign_read(state: PermissionPriv, protected: bool) -> Option<PermissionPriv> {
         Some(match state {
             // Non-writeable states just ignore foreign reads.
             non_writeable @ (Frozen | Disabled) => non_writeable,
             // Writeable states are more tricky, and depend on whether things are protected.
             // The inner data `ty_is_freeze` of `Reserved` is always irrelevant for Read
             // accesses, since the data is not being mutated. Hence the `{ .. }`
```
RalfJung, line 90:
```suggestion
            // accesses, since the data is not being mutated. Hence the `{ .. }`.
```

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -117,7 +122,8 @@ mod transition {
     /// non-protected interior mutable `Reserved` which stay the same.
     fn foreign_write(state: PermissionPriv, protected: bool) -> Option<PermissionPriv> {
         Some(match state {
-            cell @ Reserved { ty_is_freeze: false } if !protected => cell,
+            Reserved { .. } if protected => Disabled,
+            cell @ Reserved { ty_is_freeze: false, .. } => cell,
```
RalfJung, line 126:
```suggestion
            res @ Reserved { ty_is_freeze: false, .. } => res,
```

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -177,14 +183,20 @@ impl Permission {
         matches!(self.inner, Active)
     }
 
-    // Leave `interior_mut` as `None` if interior mutability
-    // is irrelevant.
-    pub fn is_reserved(self, interior_mut: Option<bool>) -> bool {
-        match (interior_mut, self.inner) {
+    // Leave `ty_is_freeze` (resp. `aliased`) as `None` if interior mutability
+    // (resp. conflict detection) is irrelevant.
+    pub fn is_reserved(self, ty_is_freeze: Option<bool>, aliased: Option<bool>) -> bool {
```
RalfJung, line 188: Please add a FIXME to move this to tests.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -282,6 +282,56 @@ enum ContinueTraversal {
     SkipChildren,
 }
 
+struct TreeVisitAux<NodeApp, ErrHandler> {
```
RalfJung, line 285: Please add doc comment and ideally find a more descriptive name.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -400,6 +395,63 @@ where {
...
+                // It also needs a different `AccessRelatedness` than the other
+                // children of `ancestor`.
+                stack.exec_and_visit(
+                    &mut self,
+                    ancestor,
+                    Some(next_in_path),
+                    AccessRelatedness::StrictChildAccess,
+                )?;
+            }
+        };
```
RalfJung, line 438: This is also completely duplicated from `traverse_parents_this_children_others`, isn't it? Can this be factored into a shared helper function?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -530,56 +584,105 @@ impl<'tcx> Tree {
...
+        // - insert the permission if it does not exist
+        // - perform the access
+        // - record the transition
+        // to which some optimizations are added:
+        // - skip the traversal of the children in some cases
+        // - do not record noop transitions
+        //
+        // This needs some extra data: the range of the `RangeMap`
+        // (the one that is not visible in diagnostics, only used to record events)
+        // is known only at call site.
```
RalfJung, line 603: I don't understand this sentence. The range is known only at the call site? The call site of what? Why do you care?

Are you explaining why this is a closure? I don't think that's necessary. It took me a good minute to even realize that this might be what you are explaining here.^^

This is a closure, so the comment should be written like the doc comment for a `fn`: say what it does and what the arguments do.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -530,56 +584,105 @@ impl<'tcx> Tree {
...
+        let node_app = |perms_range: Range<u64>,
+                        args: NodeAppArgs<'_>|
+         -> Result<ContinueTraversal, TransitionError> {
+            let NodeAppArgs { node, mut perm, rel_pos } = args;
+
+            let old_state = perm.or_insert_with(|| LocationState::new(node.default_initial_perm));
```
RalfJung, line 609: The closure doesn't seem to actually do any computation, so `or_insert` is sufficient here.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -530,56 +584,105 @@ impl<'tcx> Tree {
...
+            Ok(ContinueTraversal::Recurse)
+        };
 
-                        match old_state.skip_if_known_noop(access_kind, rel_pos) {
-                            ContinueTraversal::SkipChildren =>
-                                return Ok(ContinueTraversal::SkipChildren),
-                            _ => {}
-                        }
+        // Error handler in case `node_app` goes wrong to wrap the faulty
+        // transition in more context for diagnostics.
```
RalfJung, line 634:
```suggestion
        // Error handler in case `node_app` goes wrong. Wraps the faulty
        // transition in more context for diagnostics.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -530,56 +584,105 @@ impl<'tcx> Tree {
...
+            // This is a special access through the entire allocation.
+            // It actually only affects `initialized` locations, so we need
+            // to filter on those before initiating the traversal.
+            // In addition this implicit access should not be visible to children,
+            // thus the use of `traverse_nonchildren`.
+            for (perms_range, perms) in self.rperms.iter_mut_all() {
+                let idx = self.tag_mapping.get(&tag).unwrap();
+                if let Some(p) = perms.get(idx) {
+                    // missing permissions cannot be initialized
+                    if p.initialized {
```
RalfJung, line 671:
```suggestion
                // Only visit initialized permissions.
                if let Some(p) = perms.get(idx) && p.initialized {
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -530,56 +584,105 @@ impl<'tcx> Tree {
...
+                        // apply the following traversal only to initialized permissions
```
RalfJung, line 672:
```suggestion
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -519,6 +571,8 @@ impl<'tcx> Tree {
 
     /// Map the per-node and per-location `LocationState::perform_access`
     /// to each location of `access_range`, on every tag of the allocation.
+    /// If `access_range` is `None`, the access will be applied to
+    /// all initialized locations of the allocation.
```
RalfJung, line 575: This must definitely mention that the `None` access is visible to children only!

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -530,56 +584,105 @@ impl<'tcx> Tree {
...
+                        |args| node_app(perms_range.clone(), args),
+                        |args| err_handler(perms_range.clone(), args),
+                    )?;
+            }
+        } else {
+            // This is a special access through the entire allocation.
+            // It actually only affects `initialized` locations, so we need
+            // to filter on those before initiating the traversal.
+            // In addition this implicit access should not be visible to children,
+            // thus the use of `traverse_nonchildren`.
```
RalfJung, line 666: This needs an explanation of *why* we skip children. Please mention a testcase that would fail otherwise. (We definitely should have such a testcase.)

tests/fail/tree_borrows/spurious_read.rs
```diff
@@ -0,0 +1,107 @@
...
+        let (thread_id, barrier) = &$thread;
+        eprintln!("Thread {} executing: {}", thread_id, $msg);
+        barrier.wait();
+    }};
+}
+
+// Interleaving:
+//   retag x (protect)
+//   retag y (protect)
+//   spurious read x
```
RalfJung, line 37:
```suggestion
//   target only: spurious read x
```

tests/fail/tree_borrows/spurious_read.rs
```diff
@@ -0,0 +1,107 @@
...
+//   ret x
+//   write y
+//   ret y
+//
+// This is an interleaving that will never *not* have UB in the target
+// (`noalias` violation on `y`).
+// For the spurious read to be allowed, we need to ensure there *is* UB
+// in the source.
+fn retagx_retagy_spuriousx_retx_writey_rety() {
```
RalfJung, line 46:
```suggestion
fn retagx_retagy_retx_writey_rety() {
```
After all this doesn't do the read...

tests/fail/tree_borrows/spurious_read.rs
```diff
@@ -0,0 +1,107 @@
...
+//   retag y (protect)
+//   spurious read x
+//   ret x
+//   write y
+//   ret y
+//
+// This is an interleaving that will never *not* have UB in the target
+// (`noalias` violation on `y`).
+// For the spurious read to be allowed, we need to ensure there *is* UB
+// in the source.
```
RalfJung, line 45:
```suggestion
// in the source (i.e., without the spurious read).
```

tests/fail/tree_borrows/spurious_read.rs
```diff
@@ -0,0 +1,107 @@
...
+    let thread_x = thread::spawn(move || {
+        let b = (1, bx);
+        synchronized!(b, "start");
+        let ptr = ptr;
+        synchronized!(b, "retag x (&mut, protect)");
+        fn as_mut(x: &mut u8, b: (usize, Arc<Barrier>)) -> *mut u8 {
+            synchronized!(b, "retag y (&mut, protect)");
+            synchronized!(b, "spurious read x");
+            // This is the source interleaving, we're not actually
+            // doing the spurious read, but this is where it would be.
```
RalfJung, line 67:
```suggestion
            synchronized!(b, "location where spurious read of x would happen in the target");
            // This is ensuring that we have UB *without* the spurious read, so we don't read here.
```

tests/fail/tree_borrows/spurious_read.rs
```diff
@@ -0,0 +1,107 @@
+//@compile-flags: -Zmiri-preemption-rate=0
```
RalfJung, line 1:
```suggestion
// We ensure a deterministic interleaving
//@compile-flags: -Zmiri-preemption-rate=0
```

tests/fail/tree_borrows/spurious_read.stderr
```diff
@@ -0,0 +1,44 @@
...
+   |                 ^^^^^^ write access through <TAG> is forbidden
+   |
+   = help: this indicates a potential bug in the program: it performed an invalid operation, but the Tree Borrows rules it violated are still experimental
+   = help: the accessed tag <TAG> has state Reserved (aliased) which forbids this child write access
+help: the accessed tag <TAG> was created here, in the initial state Reserved
+  --> $DIR/spurious_read.rs:LL:CC
+   |
+LL |         fn as_mut(y: &mut u8, b: (usize, Arc<Barrier>)) -> *mut u8 {
+   |                   ^
+help: the accessed tag <TAG> later transitioned to Reserved (aliased) due to a protector release (acting as a foreign read access) over the entire allocation
```
RalfJung, line 27: It's not the entire allocation though, only the parts that were previously accessed...
```suggestion
help: the accessed tag <TAG> later transitioned to Reserved (aliased) due to a protector release (acting as a foreign read access over everything that was previously accessed by this tag)
```

tests/pass/tree_borrows/spurious_read.rs
RalfJung, line 1: The comments from the `fail` test largely also apply here. Furthermore, each test should explain how it differs from the other (since at first glance they look basically the same).

tests/pass/tree_borrows/tree-borrows.rs
```diff
@@ -93,6 +94,21 @@ fn two_mut_protected_same_alloc() {
     write_second(&mut data.0, &mut data.1);
 }
 
+// This checks that an Active returned from a function is actually Active.
```
RalfJung, line 97:
```suggestion
// This checks that a reborrowed mutable reference returned from a function
// is actually writeable.
```

src/borrow_tracker/tree_borrows/tree.rs
Vanille-N: I'll take this as implicitly allowing me to add `#![feature(let_chains)]`

RalfJung: Yes, in Miri itself we're fairly liberal with (reasonably mature) nightly features. We are closely tied to internal rustc APIs anyway.

cargo-miri and miri-script should avoid nightly features.

src/machine.rs
```diff
@@ -1382,8 +1382,31 @@ impl<'mir, 'tcx> Machine<'mir, 'tcx> for MiriMachine<'mir, 'tcx> {
...
+                .borrow_tracker
+                .as_ref()
+                .expect("we should have borrow tracking data")
+                .protected_tags
+            {
+                // If this fails, *do not* return it as an error, it just
+                // means that the allocation no longer exists, and since this is
+                // only going to produce implicit reads and not actual ones, we
+                // don't want that to produce errors.
+                if let Ok(alloc_extra) = ecx.get_alloc_extra(*alloc_id) {
```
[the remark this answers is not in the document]
Vanille-N: This was intended as a protection against trying to access an allocation that was freed.
Just because there is a protector does not guarantee that the allocation exists (weak protector). What's the canonical way of checking if an allocation still exists ?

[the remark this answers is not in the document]
RalfJung: You can call `get_alloc_info`.

tests/fail/tree_borrows/spurious_read.rs
```diff
@@ -0,0 +1,116 @@
...
+//
+// This is an interleaving that will never *not* have UB in the target
+// (`noalias` violation on `y`).
+// For the spurious read to be allowed, we need to ensure there *is* UB
+// in the source (i.e., without the spurious read).
+//
+// The interleaving differs from the one in `tests/pass/tree_borrows/spurious_read.rs` only
+// in that it has the `write y` while `y` is still protected.
+// This test thus checks that we are not overly restricting the behavior of mutable references
+// beyond function boundaries.
```
RalfJung, line 54:
```suggestion
// in that it has the `write y` while `y` is still protected.
// When the write occurs after protection ends, both source and target are fine
// (checked by the `pass` test); when the write occurs during protection, both source
// and target are UB (checked by this test).
```

tests/fail/both_borrows/retag_data_race_protected_read.rs
```diff
@@ -1,29 +0,0 @@
-//@revisions: stack tree
-//@compile-flags: -Zmiri-preemption-rate=0
-//@[tree]compile-flags: -Zmiri-tree-borrows
```
RalfJung, line 3: We should have this as a `pass` test, with comments explaining which optimization is invalid because of this code being allowed.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -298,107 +405,40 @@ impl<'tree> TreeVisitor<'tree> {
...
-                }
-                Ok(())
-            }
-        }
+    ) -> Result<(), OutErr> {
+        let start_idx = self.tag_mapping.get(&start).unwrap();
+        let mut stack = TreeVisitorStack::new(start_idx, f_propagate, err_builder);
+        stack.push_and_visit_strict_ancestors(&mut self)?;
+        // All (potentially zero) ancestors have been explored,
+        // it's time to explore the initial tag.
```
RalfJung, line 413:
```suggestion
        // it's time to explore the `start` tag.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -298,107 +405,40 @@ impl<'tree> TreeVisitor<'tree> {
...
-        // - then the unexplored descendants of `parent(start)`;
-        // ...
-        // - until finally the unexplored descendants of `root`.
-        while let Some((tag, rel_pos)) = stack.pop() {
-            stack.exec_and_visit(&mut self, tag, None, rel_pos)?;
-        }
-        Ok(())
+        let mut stack = TreeVisitorStack::new(start_idx, f_propagate, err_builder);
+        stack.push_and_visit_strict_ancestors(&mut self)?;
+        // We *don't* visit the initial tag, and we don't push its children.
```
RalfJung, line 439:
```suggestion
        // We *don't* visit the `start` tag, and we don't push its children.
```

Vanille-N:
The `fail` test in the last commit is technically redundant since it's a subpattern of our spurious read test.
I added it because I think it's helpfull to `diff` it with the `pass` test to see the difference, but it shouldn't be required.

tests/fail/tree_borrows/read_retag_no_race.rs
```diff
@@ -0,0 +1,96 @@
...
+//      read y
+//      write x
+//
+// This is a clear violation of the `noalias` requirements on `x`,
+// but we are ensuring that the UB is detected by the aliasing model
+// and *not* by the data race model.
+// Although there is race-like behavior going on in this pattern
+// (where some but not all interleavings contain UB), making this an actual
+// data race has the confusing requirement of one single access being treated
+// as being of different `AccessKind`s by different parts of Miri.
```
RalfJung, line 34:
```suggestion
// as being of different `AccessKind`s by different parts of Miri
// (a read for the aliasing model, a write for the data race model).
```

tests/fail/tree_borrows/read_retag_no_race.rs
```diff
@@ -0,0 +1,96 @@
...
+
+// The other interleaving is featured in tests/pass/tree_borrows/read_retag_no_race.rs
```
RalfJung, line 36: Can't you have both in the same test (in a single function with a parameter that says which one to use), and then in the test do
```
fn main() {
  run_test_without_race();
  run_test_with_race();
}
```
and then the backtrace will show that only the 2nd test fails, which means the first one passed?

RalfJung: In fact I don't think I understand the point of this test... "not a write for the data race model" leads to less UB so I only expected a pass test, not a fail test?

Vanille-N: With write-on-retag for the data race model, both interleavings were UB. Not doing writes for the data race model has made exactly one of those two interleavings not contain UB anymore, so I was checking that.

As I mentioned, the fail test is also a subpattern of `tests/fail/tree_borrows/spurious_read.rs`: we already have a test that contains a `retag x; read y; write x` and checks that it is UB on the write (though it is only identical up to an α-renaming of `x`,`y` into `y`,`x` which is confusing and which I will fix), so I could have `tests/pass/tree_borrows/read_retag_no_race.rs` refer to `tests/fail/tree_borrows/spurious_read.rs` as its companion interleaving and remove `tests/fail/tree_borrows/read_retag_no_race.rs`.

There is however a bigger problem with `tests/pass/tree_borrows/read_retag_no_race.rs` test, which is that I've probably inserted too much synchonization that even if we re-enabled the data-race-write-on-retag it wouldn't be a race. I'll need to rework the entire test anyway.

tests/pass/tree_borrows/read_retag_no_race.rs
```diff
@@ -27,11 +31,24 @@ unsafe impl Send for SendPtr {}
...
+// (a retag would be always a read for the aliasing model, and sometimes a write
+// for the data race model).
 
-// The other interleaving is featured in tests/fail/tree_borrows/read_retag_no_race.rs
+// The other interleaving is a subsequence of `tests/fail/tree_borrows/spurious_read.rs`
+// which asserts that
+//      retag x
+//      read y
+//      write x
+// is UB.
```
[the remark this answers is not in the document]
Vanille-N: Oops, I was planning to make the names consistent across `read_retag_no_race` and `spurious_read`
(`x` in thread 1, `y` in thread 2, `x` gets read, `y` gets written), but it looks like I started in and didn't finish the process.

RalfJung:
Great, thanks. :)
Please rebase, fix the build, and squash.

RalfJung:
Looking great!
@bors r+

PR #1776, opened 2021-04-16
Implement calls to exported symbols (#1170)

hyd-dev:
Closes #1170.

PR #3071, opened 2023-09-21
deprecate -Zmiri-disable-abi-check

RalfJung:
This was added in https://github.com/rust-lang/miri/pull/1776 but I couldn't find any discussion or motivation.


RalfJung:
Using github search, the only uses of this flag that I found are copies/forks of Miri itself.

@saethlin I recall you said in the past you didn't like our `disable` flags, so I assume you are in favor?

@oli-obk what do you think?

oli-obk (approved):

saethlin (approved):
Yes I am in favor

Issue #2057, opened 2022-04-07
Shim wishlist


saethlin:
I run Miri on the most-downloaded crates, with the flags `-Zmiri-panic-on-unsupported` and `--no-fail-fast` which basically asks Miri to try to run as many tests as it can. Searching all my build logs, I find these are the most-encountered `can't call foreign function: ` errors:
```
15232 counts
(  1)     4329 (28.4%, 28.4%): epoll_create1
(  2)     1128 ( 7.4%, 35.8%): pipe2
(  3)      729 ( 4.8%, 40.6%): socket
(  4)      345 ( 2.3%, 42.9%): TLS_method
(  5)      294 ( 1.9%, 44.8%): __xpg_strerror_r
(  6)      279 ( 1.8%, 46.6%): cuInit
(  7)      257 ( 1.7%, 48.3%): realpath
(  8)      255 ( 1.7%, 50.0%): openat
(  9)      229 ( 1.5%, 51.5%): getaddrinfo
( 10)      205 ( 1.3%, 52.8%): Py_IsInitialized
( 11)      199 ( 1.3%, 54.2%): ruby_init
( 12)      198 ( 1.3%, 55.5%): localtime_r
( 13)      174 ( 1.1%, 56.6%): dlopen
( 14)      164 ( 1.1%, 57.7%): GEOS_init_r
( 15)      161 ( 1.1%, 58.7%): gnu_get_libc_version
( 16)      144 ( 0.9%, 59.7%): eventfd
( 17)      131 ( 0.9%, 60.5%): js_newstate
( 18)      130 ( 0.9%, 61.4%): Rf_initialize_R
( 19)      124 ( 0.8%, 62.2%): secp256k1_context_create
( 20)      111 ( 0.7%, 62.9%): ts_parser_new
( 21)      109 ( 0.7%, 63.6%): mmap
( 22)      106 ( 0.7%, 64.3%): getuid
( 23)      104 ( 0.7%, 65.0%): getpid
( 24)      104 ( 0.7%, 65.7%): lua_newstate
( 25)      103 ( 0.7%, 66.4%): blst_scalar_from_uint64
( 26)      103 ( 0.7%, 67.1%): ion_reader_open_buffer
( 27)      101 ( 0.7%, 67.7%): sigemptyset
( 28)       89 ( 0.6%, 68.3%): gst_init_check
( 29)       88 ( 0.6%, 68.9%): mdb_env_create
[...]

PR #3072, opened 2023-09-21
Implement `llvm.ctpop.v*` intrinsics

eduardosm:
Tested through x86 avx512vpopcntdq and avx512bitalg functions.

I picked them from https://github.com/rust-lang/miri/issues/2057#issuecomment-1716309674, which looked easy.

### DIFF
--- a/src/shims/foreign_items.rs
+++ b/src/shims/foreign_items.rs
@@ -1032,6 +1032,29 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 }
             }
 
+            // Used to implement the x86 `_mm{,256,512}_popcnt_epi{8,16,32,64}` and wasm
+            // `{i,u}8x16_popcnt` functions.
+            name if name.starts_with("llvm.ctpop.v") => {
+                let [op] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let (op, op_len) = this.operand_to_simd(op)?;
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                assert_eq!(dest_len, op_len);
+
+                for i in 0..dest_len {
+                    let op = this.read_immediate(&this.project_index(&op, i)?)?;
+                    // Use `to_uint` to get a zero-extended `u128`. Those
+                    // extra zeros will not affect `count_ones`.
+                    let res = op.to_scalar().to_uint(op.layout.size)?.count_ones();
+
+                    this.write_scalar(
+                        Scalar::from_uint(res, op.layout.size),
+                        &this.project_index(&dest, i)?,
+                    )?;
+                }
+            }
+
             name if name.starts_with("llvm.x86.sse.") => {
                 return shims::x86::sse::EvalContextExt::emulate_x86_sse_intrinsic(
                     this, link_name, abi, args, dest,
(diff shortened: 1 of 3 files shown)


src/shims/foreign_items.rs
```diff
@@ -1032,6 +1032,29 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
 
+            // Used to implement the x86 `_mm{,256,512}_popcnt_epi{8,16,32,64}` and wasm
+            // `{i,u}8x16_popcnt` functions.
+            name if name.starts_with("llvm.ctpop.v") => {
+                let [op] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let (op, op_len) = this.operand_to_simd(op)?;
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                assert_eq!(dest_len, op_len);
```
saethlin, line 1043: You have this check in here and a lot of other intrinsics. I think this is a good thing to check, but an ICE is probably the wrong failure mode; this assertion can be tripped if someone declares their own LLVM intrinsics but does it wrong.

For example, this program ICEs:
```rust
#![feature(link_llvm_intrinsics, simd_ffi)]
#[allow(improper_ctypes)]
use std::arch::x86_64::*;

extern "C" {
    #[link_name = "llvm.x86.sse2.sqrt.pd"]
    fn oof(x: __m128d) -> __m256d;
}

fn main() {
    unsafe {
        oof(_mm_set_pd(0.0, 0.0));
    };
}
```

saethlin: https://github.com/rust-lang/miri/issues/3076

saethlin:
Thanks!
@bors r+

tests/pass/intrinsics-x86-avx512bitalg.rs
RalfJung, line 1: These test files are getting a bit too fine-grained IMO. What about just using `intrinsics-x86-other` for "everything except the SSE/SSE2 intrinsics"? llvm.x86.addcarry.64 should also be tested there, probably.

src/shims/foreign_items.rs
RalfJung: We generally have the policy that ICEing when intrinsics are used incorrectly is fine. I think we should apply that to LLVM intrinsics as well. If https://github.com/rust-lang/rust/pull/116093 lands, people will get an appropriate warning when enabling the feature.

tests/pass/intrinsics-x86-avx512bitalg.rs
RalfJung: Ah, we already have tests/pass/intrinsics-x86.rs. So this could be added there?

saethlin: The tests are grouped by what CPU features they use, and the avx512 CPU features are significantly more fine-grained than the others. I wouldn't mind lumping all the avx512 intrinsics together, but I don't think they belong in "other" with `llvm.x86.rdtsc`.

RalfJung: Ah, that's fair. But the two test files added by this PR should IMO be merged.

@eduardosm can you wile a follow-up PR?

eduardosm: I will!

PR #3075, opened 2023-09-22
Move `llvm.x86.*` shims into `shims::x86` and implement `_addcarry_u32` and `_subborrow_u{32,64}`

eduardosm:
This PR moves all `llvm.x86.*` shims into `shims::x86` and adds `llvm.x86.addcarry.32`, `llvm.x86.subborrow.32` and `llvm.x86.subborrow.64`.

Additionally, it fixes the input carry semantics of `llvm.x86.addcarry.32`. The input carry is an 8-bit value that is interpreted as 1 when it is non-zero.

https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-8/addcarry-u32-addcarry-u64.html

### DIFF
--- a/src/shims/foreign_items.rs
+++ b/src/shims/foreign_items.rs
@@ -22,7 +22,7 @@ use rustc_target::{
 };
 
 use super::backtrace::EvalContextExt as _;
-use crate::helpers::{convert::Truncate, target_os_is_unix};
+use crate::helpers::target_os_is_unix;
 use crate::*;
 
 /// Returned by `emulate_foreign_item_by_name`.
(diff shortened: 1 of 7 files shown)


src/shims/foreign_items.rs
```diff
@@ -981,30 +981,6 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
-                let sum_field = this.project_field(dest, 1)?;
-                this.write_scalar(Scalar::from_u64(sum), &sum_field)?;
-            }
-            "llvm.x86.sse2.pause"
-                if this.tcx.sess.target.arch == "x86" || this.tcx.sess.target.arch == "x86_64" =>
-            {
-                let [] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
-                this.yield_active_thread();
-            }
             "llvm.aarch64.isb" if this.tcx.sess.target.arch == "aarch64" => {
```
RalfJung, line 984:
```suggestion
            // FIXME: Move these to an `arm` submodule.
            "llvm.aarch64.isb" if this.tcx.sess.target.arch == "aarch64" => {
```

src/shims/x86/mod.rs
```diff
@@ -1,11 +1,97 @@
...
+                    mir::BinOp::Sub,
+                    &sub,
+                    &ImmTy::from_uint(b_in, a.layout),
+                )?;
+                let b_out = overflow1 | overflow2;
+
+                this.write_scalar(Scalar::from_u8(b_out.into()), &this.project_field(dest, 0)?)?;
+                this.write_immediate(*sub, &this.project_field(dest, 1)?)?;
+            }
+            name if name.starts_with("sse.") => {
```
RalfJung, line 80:
```suggestion

            name if name.starts_with("sse.") => {
```

tests/pass/intrinsics-x86.rs
```diff
@@ -1,3 +1,51 @@
...
+        // SAFETY: There are no safety requirements for calling `_subborrow_u32`.
+        // It's just unsafe for API consistency with other intrinsics.
+        let b_out = unsafe { arch::_subborrow_u32(b_in, a, b, &mut sum) };
+        (b_out, sum)
+    }
+
+    pub fn main() {
+        assert_eq!(adc(0, 1, 1), (0, 2));
+        assert_eq!(adc(1, 1, 1), (0, 3));
+        assert_eq!(adc(2, 1, 1), (0, 3));
```
RalfJung, line 27:
```suggestion
        assert_eq!(adc(2, 1, 1), (0, 3)); // any non-zero carry acts as 1!
```

tests/pass/intrinsics-x86.rs
```diff
@@ -1,3 +1,51 @@
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
```
RalfJung, line 1: Ah, here's another way to handle tests that should run only on x86. Might be good to also port the other tests (that use the long `ignore-` list) to this pattern.

eduardosm: For SSE and SSE2 that will work fine because they are enabled by default. But other features require `-C target-feature=+feature`, which will generate warnings on non-x86.

Maybe a new directive could be added to compiletest? Something like `@only-x86-family`

eduardosm: I've done it for SSE and SSE2 for the time being.

saethlin:
It looks to me like you addressed Ralf's comments above but GitHub doesn't seem to have noticed :thinking:

saethlin:
@eduardosm Can you squash the commits down a bit? I think your original 4 commits are a good organization, but the rest don't stand on their own.

Then this LGTM (and I  would rather like to sync these changes into rust-lang/rust).

eduardosm:
Ready!

PR #3078, opened 2023-09-23
Support when `build-std` is set for Cargo

norepimorphism:
See #2705. This is my first contribution to a rust-lang project, and I am not extremely familiar with the internals of Miri and Cargo, so I apologize if there obvious deficiencies in this PR (besides the ones I am about to list, which are:)

- in the case that `build-std` is set, `cargo-miri` will not set `MIRI_SYSROOT`, which may be problematic for things that expect `MIRI_SYSROOT` to always be set after `cargo miri setup`
- `--print-syroot` is ignored by `cargo miri setup` when `build-std` is set (but what else can it do)
- the Miri driver parses the value of `--crate-name` from the CLI arguments using a dedicated function, but ideally we'd want to borrow functionality from the `arg` module in `cargo-miri`?
- it may be possible for a user to name their crate `std` or `core` or something, in which case, if `build-std` is also set, the Miri driver will naively assume that crate is really part of the standard library and disable debug assertions, enable overflow checks, etc. (I'm not actually sure if this can happen or not; someone please clarify if I'm wrong here)
- there may be some special flags we set for the Miri sysroot that I have forgotten to pass along in the `build-std` case

I have created a test crate with `build-std` set and a target spec for the PlayStation 2 CPU, with the expectation that Miri will see that `build-std` is set and skip building the sysroot. I am not actually able to test this at the moment.

### DIFF
--- a/cargo-miri/src/phases.rs
+++ b/cargo-miri/src/phases.rs
@@ -93,8 +93,30 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
     let target = get_arg_flag_value("--target");
     let target = target.as_ref().unwrap_or(host);
 
-    // We always setup.
-    let miri_sysroot = setup(&subcommand, target, &rustc_version, verbose);
+    let has_build_std = is_build_std_set();
+    if has_build_std {
+        if env::var_os("MIRI_SYSROOT").is_some() {
+            println!("WARNING: `MIRI_SYSROOT` has no effect when Cargo has `build-std` enabled.");
+        }
+        if env::var_os("MIRI_NO_STD").is_some() {
+            println!("WARNING: `MIRI_NO_STD` has no effect when Cargo has `build-std` enabled.");
+        }
+    }
+    let cargo_cmd = match subcommand {
+        MiriCommand::Forward(s) => s,
+        MiriCommand::Setup => {
+            if has_build_std {
+                println!(
+                    "NOTE: `cargo miri setup` is unnecessary when Cargo has `build-std` enabled. \
+                    libstd and friends will be built on-the-fly when `cargo miri run` or `cargo \
+                    miri test` is invoked."
+                );
+            } else {
+                setup(&subcommand, target, &rustc_version, verbose);
+            }
+            return; // `cargo miri setup` stops here.
+        }
+    };
 
     // Invoke actual cargo for the job, but with different flags.
     // We re-use `cargo test` and `cargo run`, which makes target and binary handling very easy but
@@ -107,10 +129,6 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
         .into_os_string()
         .into_string()
         .expect("current executable path is not valid UTF-8");
-    let cargo_cmd = match subcommand {
-        MiriCommand::Forward(s) => s,
-        MiriCommand::Setup => return, // `cargo miri setup` stops here.
-    };
     let metadata = get_cargo_metadata();
     let mut cmd = cargo();
     cmd.arg(&cargo_cmd);
@@ -159,8 +177,12 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
     // Forward all further arguments (not consumed by `ArgSplitFlagValue`) to cargo.
     cmd.args(args);
 
-    // Let it know where the Miri sysroot lives.
-    cmd.env("MIRI_SYSROOT", miri_sysroot);
+    if has_build_std {
+        cmd.env("MIRI_STD_AWARE_CARGO", "1");
+    } else {
+        // Let it know where the Miri sysroot lives.
+        cmd.env("MIRI_SYSROOT", setup(&subcommand, target, &rustc_version, verbose));
+    }
     // Set `RUSTC_WRAPPER` to ourselves.  Cargo will prepend that binary to its usual invocation,
     // i.e., the first argument is `rustc` -- which is what we use in `main` to distinguish
     // the two codepaths. (That extra argument is why we prefer this over setting `RUSTC`.)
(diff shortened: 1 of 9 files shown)


src/bin/miri.rs
```diff
@@ -242,21 +242,49 @@ fn run_compiler(
...
+                | "core"
+                | "alloc"
+                | "proc_macro"
+                | "panic_abort"
+                | "panic_unwind"
+                | "test"
+                | "compiler_builtins"
+            ) {
+                // We are compiling the standard library.
+                args.push("-Cdebug-assertions=off".into());
```
norepimorphism, line 264: I have also written a version of this which attempts to fix existing "-Cdebug-assertions", "-Coverflow-checks", etc. options if they are present (and also accounting for, e.g., if "-C" and "debug-assertions" are separate arguments), and only resorts to appending new arguments if the options are not present. However, I'm not sure if this extra complexity is warranted, because I think rustc can handle duplicate codegen options just fine, so probably the only benefit would be to make the rustc invocation "nicer" for someone debugging the interaction between miri and rustc

RalfJung:
Thanks for the PR! :) It will be reviewed eventually but that might take a bit at least I am rather busy currently.

I am quite concerned about that list of crate names in the Miri driver, that sounds like something that we really want to avoid.

cargo-miri/src/util.rs
```diff
@@ -252,3 +252,20 @@ pub fn debug_cmd(prefix: &str, verbose: usize, cmd: &Command) {
     }
     eprintln!("{prefix} running command: {cmd:?}");
 }
+
+/// Checks if Cargo's build-std setting is enabled for one or more crates.
+pub fn is_build_std_set() -> bool {
+    let cmd = cargo()
```
RalfJung, line 258: You need to pass `cargo_extra_flags()` as arguments to ensure that if the user sets `--cfg unstable.build-std` that will be respected. Also what about `-Zbuild-std`? I don't think I understand how these flags even interact.

It might be worth having a little helper function that invokes `cargo config` to get the value of an arbitrary key.

cargo-miri/src/util.rs
```diff
@@ -252,3 +252,20 @@ pub fn debug_cmd(prefix: &str, verbose: usize, cmd: &Command) {
...
+        // TODO: remove the -Z argument when this subcommand is stabilized;
+        // see <https://github.com/rust-lang/cargo/issues/9301>
+        .args(&["-Zunstable-options", "config", "get", "--format=json-value", "unstable.build-std"])
+        .output()
+        .expect("failed to get Cargo config");
+    // TODO: Cargo will exit with an error code if the `unstable.build-std` key is not set. Should
+    // we still return false if an unrelated error has occurred?
```
RalfJung, line 265: Ideally we'd distinguish these cases. Which status code does cargo return when the key is absent? The least we can do is raise an error on any unexpected status code.

cargo-miri/src/phases.rs
```diff
@@ -93,8 +93,30 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
...
+    let has_build_std = is_build_std_set();
+    if has_build_std {
+        if env::var_os("MIRI_SYSROOT").is_some() {
+            println!("WARNING: `MIRI_SYSROOT` has no effect when Cargo has `build-std` enabled.");
+        }
+        if env::var_os("MIRI_NO_STD").is_some() {
+            println!("WARNING: `MIRI_NO_STD` has no effect when Cargo has `build-std` enabled.");
+        }
+    }
+    let cargo_cmd = match subcommand {
```
RalfJung, line 105: Why did you move the `cargo_cmd` logic up? This is confusing now, since the `match` has too many jobs.

Also we want to call `setup` before any of the other logic here. I think this should be more like
```rust
let miri_sysroot = if has_build_std {
  Some(setup(...))
} else { None };
```

src/bin/miri.rs
```diff
@@ -242,21 +242,49 @@ fn run_compiler(
...
-        let sysroot_flag = "--sysroot";
-        if !args.iter().any(|e| e == sysroot_flag) {
-            // Using the built-in default here would be plain wrong, so we *require*
-            // the env var to make sure things make sense.
-            let miri_sysroot = env::var("MIRI_SYSROOT").unwrap_or_else(|_| {
-                show_error!(
-                    "Miri was invoked in 'target' mode without `MIRI_SYSROOT` or `--sysroot` being set"
-                    )
-            });
+        if env::var_os("MIRI_STD_AWARE_CARGO").is_some() {
```
RalfJung, line 245: I would strongly prefer if we could avoid having such hacks in the Miri driver.  This should all be taken care of by `cargo-miri`.

`cargo-miri` already injects itself into every invocation from cargo, so it should be entirely possible to do any further command-line argument adjustments there rather than doing them inside the Miri driver.

src/bin/miri.rs
RalfJung, line 264: Why do we want this? If the user has enabled debug assertions then we should honor that. We should behave like regular `cargo run`, which for build-std will also apply flags to the std build.

Miri no longer sets `-Cdebug-assertions=on` either, so I see no reason why we'd start messing with `-C` flags again.

src/bin/miri.rs
```diff
@@ -242,21 +242,49 @@ fn run_compiler(
...
+                | "compiler_builtins"
+            ) {
+                // We are compiling the standard library.
+                args.push("-Cdebug-assertions=off".into());
+                args.push("-Coverflow-checks=on".into());
+            }
+            // See `cargo_miri::phase_rustc`.
+            if crate_name == "panic_abort" {
+                args.push("-Cpanic=abort".into());
+            }
```
RalfJung, line 270: Is this really needed? If build-std with regular rustc doesn't need it then we shouldn't need it either. AFAIK build-std will simply only build the panic runtime that is actually needed for this crate and this avoids the issue that comes up when building the panic runtime for a sysroot.

test-cargo-miri/Cargo.toml
```diff
@@ -16,6 +16,7 @@ issue_1567 = { path = "issue-1567" }
 issue_1691 = { path = "issue-1691" }
 issue_1705 = { path = "issue-1705" }
 issue_1760 = { path = "issue-1760" }
+issue_2705 = { path = "issue-2705" }
```
RalfJung, line 19: I was expecting we would add something somewhere that calls `cargo miri run -Zbuild-std`; why is this a separate crate?

cargo-miri/src/util.rs
norepimorphism: Well, I think Cargo returns 101 for most errors (unless I am misunderstanding your question). This includes when the key is absent but also, e.g., when the key `a.b` cannot be resolved in the first place because `a` exists and is not a table. I'm sure there are many other error cases for `cargo config get` which return 101.

Relevant Cargo code: https://github.com/rust-lang/cargo/blob/f5418fcfebe7d01536b891c8b2ceb32c4eaf098c/src/cargo/util/errors.rs#L324C1-L341C2

RalfJung: Okay, that's unfortunate.

But still you can then treat only error 101 as "absent" and all other codes as actual errors.

PR #3086, opened 2023-09-26
Implement SSE3 and SSSE3 intrinsics

eduardosm:


src/shims/x86/mod.rs
```diff
@@ -286,3 +298,49 @@ fn bin_op_simd_float_all<'tcx, F: rustc_apfloat::Float>(
...
+                this.read_immediate(&this.project_index(&left, base_i)?)?,
+                this.read_immediate(&this.project_index(&left, base_i.checked_add(1).unwrap())?)?,
+            )
+        } else {
+            let base_i = i.checked_sub(middle).unwrap().checked_mul(2).unwrap();
+            (
+                this.read_immediate(&this.project_index(&right, base_i)?)?,
+                this.read_immediate(&this.project_index(&right, base_i.checked_add(1).unwrap())?)?,
+            )
+        };
```
saethlin, line 334: The actual difference here is whether we apply `.checked_sub(middle)` and whether we use `left` or `right` as our source vector. Can this be rewritten along those lines? I'd rather read something like this because it's more clear what's different where:
```rust
let (i, src) = if i < middle {
    (i, &left)
} else {
    (i.checked_sub(middle).unwrap(), &*right)
};
let base_i = i .checked_mul(2).unwrap();
let lhs = this.read_immediate...
```

eduardosm: Yeah, that's definitely better.

src/shims/x86/mod.rs
```diff
@@ -286,3 +298,42 @@ fn bin_op_simd_float_all<'tcx, F: rustc_apfloat::Float>(
...
+
+    assert_eq!(dest_len, left_len);
+    assert_eq!(dest_len, right_len);
+    assert_eq!(dest_len % 2, 0);
+
+    let middle = dest_len / 2;
+    for i in 0..dest_len {
+        let (j, src) =
+            if i < middle { (i, &left) } else { (i.checked_sub(middle).unwrap(), &right) };
+        let base_i = j.checked_mul(2).unwrap();
```
RalfJung, line 324: Please add comments explaining the meaning of "j" and "base_i"

src/shims/x86/mod.rs
```diff
@@ -286,3 +298,42 @@ fn bin_op_simd_float_all<'tcx, F: rustc_apfloat::Float>(
...
+        let lhs = this.read_immediate(&this.project_index(src, base_i)?)?;
+        let rhs = this.read_immediate(&this.project_index(src, base_i.checked_add(1).unwrap())?)?;
+
+        let res = if saturating {
+            Immediate::from(this.saturating_arith(which, &lhs, &rhs)?)
+        } else {
+            let (res, _overflow) = this.overflowing_binary_op(which, &lhs, &rhs)?;
```
RalfJung, line 331: If you ignore overflow, please use "wrapping_binary_op". (Applies elsewhere, too.)

src/shims/x86/sse3.rs
```diff
@@ -0,0 +1,92 @@
...
+        args: &[OpTy<'tcx, Provenance>],
+        dest: &PlaceTy<'tcx, Provenance>,
+    ) -> InterpResult<'tcx, EmulateByNameResult<'mir, 'tcx>> {
+        let this = self.eval_context_mut();
+        // Prefix should have already been checked.
+        let unprefixed_name = link_name.as_str().strip_prefix("llvm.x86.sse3.").unwrap();
+
+        match unprefixed_name {
+            // Used to implement the _mm_addsub_ps and _mm_addsub_pd functions.
+            // Alternatively add and subtract floating point (f32 or f64) from
```
RalfJung, line 27: Did you mean "alternatingly"?

src/shims/x86/sse3.rs
```diff
@@ -0,0 +1,92 @@
...
+            // Reads a 128-bit vector from an unaligned pointer. This intrinsic
+            // is expected to perform better than a regular unaligned read when
+            // the data crosses a cache line, but for Miri this is just a regular
+            // unaligned read.
+            "ldu.dq" => {
+                let [src_ptr] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let dest = dest.force_mplace(this)?;
+
+                let src_ptr = this.read_pointer(src_ptr)?;
+                let dest_ptr = this.read_pointer(&this.mplace_to_ref(&dest)?)?;
```
RalfJung, line 77: Isn't this just `dest.ptr()`?

src/shims/x86/ssse3.rs
```diff
@@ -0,0 +1,206 @@
...
+                let (op, op_len) = this.operand_to_simd(op)?;
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                assert_eq!(op_len, dest_len);
+
+                for i in 0..dest_len {
+                    let op = this.read_scalar(&this.project_index(&op, i)?)?;
+                    let dest = this.project_index(&dest, i)?;
+
+                    let res = op.to_int(dest.layout.size)?.unsigned_abs();
```
RalfJung, line 39:
```suggestion
                    // Converting to a host "i128" works since the input is always signed.
                    let res = op.to_int(dest.layout.size)?.unsigned_abs();
```

src/shims/x86/ssse3.rs
```diff
@@ -0,0 +1,206 @@
...
+                    } else {
+                        // If the highest bit in `right` is 1, write zero.
+                        Scalar::from_u8(0)
+                    };
+
+                    this.write_scalar(res, &dest)?;
+                }
+            }
+            // Used to implement the _mm_h{add,adds,sub}_epi{16,32} functions.
+            // Horizontally add/add with saturation/subtract adjacent 16/32-bit
```
RalfJung, line 74:
```suggestion
            // Horizontally add / add with saturation / subtract adjacent 16/32-bit
```

src/shims/x86/ssse3.rs
```diff
@@ -0,0 +1,206 @@
...
+            "pmadd.ub.sw.128" => {
+                let [left, right] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let (left, left_len) = this.operand_to_simd(left)?;
+                let (right, right_len) = this.operand_to_simd(right)?;
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                assert_eq!(left_len, right_len);
+                assert_eq!(dest_len, left_len / 2);
```
RalfJung, line 107:
```suggestion
                assert_eq!(2 * dest_len, left_len);
```
Let's avoid potential rounding pitfalls...

src/shims/x86/ssse3.rs
```diff
@@ -0,0 +1,206 @@
...
+                    let left1 = this.read_scalar(&this.project_index(&left, j1)?)?.to_u8()?;
+                    let right1 = this.read_scalar(&this.project_index(&right, j1)?)?.to_i8()?;
+
+                    let j2 = j1.checked_add(1).unwrap();
+                    let left2 = this.read_scalar(&this.project_index(&left, j2)?)?.to_u8()?;
+                    let right2 = this.read_scalar(&this.project_index(&right, j2)?)?.to_i8()?;
+
+                    let dest = this.project_index(&dest, i)?;
+
+                    let mul1 = i16::from(left1).checked_mul(right1.into()).unwrap();
```
RalfJung, line 120:
```suggestion
                    // Multiplication of a u8 and an i8 into an i16 cannot overflow.
                    let mul1 = i16::from(left1).checked_mul(right1.into()).unwrap();
```

src/shims/x86/ssse3.rs
```diff
@@ -0,0 +1,206 @@
...
+                    _ => unreachable!(),
+                };
+
+                horizontal_bin_op(this, which, saturating, left, right, dest)?;
+            }
+            // Used to implement the _mm_maddubs_epi16 function.
+            // Multiplies corresponding pairs of packed 8-bit unsigned integer
+            // values contained in the first source operand and packed 8-bit signed
+            // integer values contained in the second source operand, add pairs of
+            // contiguous products with signed saturation, and writes the 16-bit sums to
```
RalfJung, line 95: I am confused by the "contiguous" here. This might just need some math: we are adding the pairs at indices 2*i and 2*i+1 to compute the output index i.

src/shims/x86/ssse3.rs
```diff
@@ -0,0 +1,206 @@
...
+                    "phadd.sw.128" => (mir::BinOp::Add, true),
+                    "phsub.w.128" | "phsub.d.128" => (mir::BinOp::Sub, false),
+                    "phsub.sw.128" => (mir::BinOp::Sub, true),
+                    _ => unreachable!(),
+                };
+
+                horizontal_bin_op(this, which, saturating, left, right, dest)?;
+            }
+            // Used to implement the _mm_maddubs_epi16 function.
+            // Multiplies corresponding pairs of packed 8-bit unsigned integer
```
RalfJung, line 92: Please mention that this is widening multiplication into an i16.

src/shims/x86/ssse3.rs
```diff
@@ -0,0 +1,206 @@
...
+                    let mul2 = i16::from(left2).checked_mul(right2.into()).unwrap();
+                    let res = mul1.saturating_add(mul2);
+
+                    this.write_scalar(Scalar::from_i16(res), &dest)?;
+                }
+            }
+            // Used to implement the _mm_mulhrs_epi16 function.
+            // Multiplies packed 16-bit signed integer values, truncate the 32-bit
+            // product to the 18 most significant bits by right-shifting, round the
+            // truncated value by adding 1, and write bits `[16:1]` to the destination.
```
RalfJung, line 130: Wtf are those operations... this seems so arbitrary.^^

In which sense does adding 1 mean that something has been "rounded"?

What does "bits [16:1]" mean? Can you use Rust subslice syntax instead?

src/shims/x86/ssse3.rs
```diff
@@ -0,0 +1,206 @@
...
+                        .checked_add(1)
+                        .unwrap()
+                        >> 1;
+
+                    this.write_scalar(Scalar::from_i16(res.try_into().unwrap()), &dest)?;
+                }
+            }
+            // Used to implement the _mm_sign_epi{8,16,32} functions.
+            // Negates elements from `left` when the its corresponding from
+            // `right` is negative. If an element from `right` is zero, zero
```
RalfJung, line 158:
```suggestion
            // Negates elements from `left` when the its corresponding element
            // in `right` is negative. If an element from `right` is zero, zero
```

src/shims/x86/ssse3.rs
```diff
@@ -0,0 +1,206 @@
...
+            // is writen to the corresponding output element.
```
RalfJung, line 159:
```suggestion
            // is writen to the corresponding output element.
            // Basically, we multiply `left` with `right.signum()`.
```

src/shims/x86/ssse3.rs
```diff
@@ -0,0 +1,206 @@
...
+                    } else if right < 0 {
+                        if left == dest.layout.size.signed_int_min() {
+                            // Make sure `Scalar::from_int` does not get an overflowed value.
+                            //
+                            // For example, with 16-bit value -0x8000, negating after sign
+                            // extension to i128 will produce 0x8000. However, `Scalar::from_int`
+                            // will fail because 0x8000 is outside the range of signed 16-bit
+                            // integers. If the negation had been done directly with wrapping 16-bit
+                            // arithmetic, the result would be -0x8000, so we simulate that.
+                            left
```
RalfJung, line 191: Can you avoid that by using `wrapping_binary_op` instead, multiplying `left` with `Scalar::from_int(right.to_int().signum())`?

tests/pass/intrinsics-x86-sse3-ssse3.rs
```diff
@@ -0,0 +1,262 @@
...
+//@compile-flags: -C target-feature=+ssse3
+
+use core::mem::transmute;
+#[cfg(target_arch = "x86")]
+use std::arch::x86::*;
+#[cfg(target_arch = "x86_64")]
+use std::arch::x86_64::*;
+
+fn main() {
+    assert!(is_x86_feature_detected!("ssse3"));
```
RalfJung, line 20: Is there an sse3 target feature we can assert here?

src/shims/x86/ssse3.rs
```diff
@@ -0,0 +1,206 @@
...
+                    let left2 = this.read_scalar(&this.project_index(&left, j2)?)?.to_u8()?;
+                    let right2 = this.read_scalar(&this.project_index(&right, j2)?)?.to_i8()?;
+
+                    let dest = this.project_index(&dest, i)?;
+
+                    let mul1 = i16::from(left1).checked_mul(right1.into()).unwrap();
+                    let mul2 = i16::from(left2).checked_mul(right2.into()).unwrap();
+                    let res = mul1.saturating_add(mul2);
+
+                    this.write_scalar(Scalar::from_i16(res), &dest)?;
```
RalfJung, line 124: Are there some tests demonstrating that this will indeed saturate, and that multiplication is widening (rather than 8-bit-overflowing)?

src/shims/x86/ssse3.rs
```diff
@@ -0,0 +1,206 @@
...
+                for i in 0..dest_len {
+                    let right = this.read_scalar(&this.project_index(&right, i)?)?.to_u8()?;
+                    let dest = this.project_index(&dest, i)?;
+
+                    let res = if right & 0x80 == 0 {
+                        let j = right % 16; // index wraps around
+                        this.read_scalar(&this.project_index(&left, j.into())?)?
+                    } else {
+                        // If the highest bit in `right` is 1, write zero.
+                        Scalar::from_u8(0)
```
RalfJung, line 67: Is there a test for this?

src/shims/x86/ssse3.rs
```diff
@@ -0,0 +1,206 @@
...
+
+                assert_eq!(dest_len, left_len);
+                assert_eq!(dest_len, right_len);
+
+                for i in 0..dest_len {
+                    let right = this.read_scalar(&this.project_index(&right, i)?)?.to_u8()?;
+                    let dest = this.project_index(&dest, i)?;
+
+                    let res = if right & 0x80 == 0 {
+                        let j = right % 16; // index wraps around
```
RalfJung, line 63: Is there a test for this?

src/shims/x86/ssse3.rs
```diff
@@ -0,0 +1,206 @@
...
+                let [left, right] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let (which, saturating) = match unprefixed_name {
+                    "phadd.w.128" | "phadd.d.128" => (mir::BinOp::Add, false),
+                    "phadd.sw.128" => (mir::BinOp::Add, true),
+                    "phsub.w.128" | "phsub.d.128" => (mir::BinOp::Sub, false),
+                    "phsub.sw.128" => (mir::BinOp::Sub, true),
+                    _ => unreachable!(),
+                };
```
RalfJung, line 87: Are there tests that hit the wrapping / saturation cases?

src/shims/x86/sse3.rs
eduardosm: I did :)

src/shims/x86/sse3.rs
eduardosm: I didn't know about `ptr`. Thanks!

tests/pass/intrinsics-x86-sse3-ssse3.rs
eduardosm: SSSE3 implicitly enables SSE3. But it doesn't hurt to explicitly check it.

src/shims/x86/ssse3.rs
eduardosm: There is. The second element of `b` is 128 = 0x80 and the second element of `expected` is 0.

```rust
    #[target_feature(enable = "ssse3")]
    unsafe fn test_mm_shuffle_epi8() {
        let a = _mm_setr_epi8(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16);
        let b = _mm_setr_epi8(4, 128_u8 as i8, 4, 3, 24, 12, 6, 19, 12, 5, 5, 10, 4, 1, 8, 0);
        let expected = _mm_setr_epi8(5, 0, 5, 4, 9, 13, 7, 4, 13, 6, 6, 11, 5, 2, 9, 1);
        let r = _mm_shuffle_epi8(a, b);
        assert_eq_m128i(r, expected);
    }
    test_mm_shuffle_epi8();
```

src/shims/x86/ssse3.rs
eduardosm: There is now :)

src/shims/x86/ssse3.rs
eduardosm: There are now

src/shims/x86/sse3.rs
```diff
@@ -0,0 +1,92 @@
...
+                let [src_ptr] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let dest = dest.force_mplace(this)?;
+
+                let src_ptr = this.read_pointer(src_ptr)?;
+                let dest_ptr = dest.ptr();
+
+                this.mem_copy(
+                    src_ptr,
+                    Align::ONE,
+                    dest_ptr,
```
RalfJung, line 82:
```suggestion
                this.mem_copy(
                    src_ptr,
                    Align::ONE,
                    dest.ptr(),
```

eduardosm:
Some conversations were hidden by github and I have just seem them now 🤦🏻‍♂️

src/shims/x86/ssse3.rs
eduardosm: They are as arbitrary as intel decided to make them :)

After adding one, the lowest bit is dropped when taking bits `[16:1]` (aka `1..=16`). Dropping the lowest bit is like dividing by 2, so adding 1 before dividing by 2 is like adding 0.5 after the division and truncating the result.

RalfJung: Oh, okay. The text talks about rounding before it even mentions taking some of the bits so this wasn't clear at all. So please say something like: Multiplies packed 16-bit signed integer values, and then divides the 32-bit result by 2 (rounding up) by first adding 1 and then taking the bits `1..=16`.

eduardosm: It is rounding to nearest rather than rounding up. And we are still taking the highest 18 bits before dividing by 2.

That would leave us:
```rust
// Multiplies packed 16-bit signed integer values, truncate the 32-bit
// product to the 18 most significant bits by right-shifting, and then
// divides the 18-bit value by 2 (rounding to nearest) by first adding
// 1 and then taking the bits `1..=16`.
```

RalfJung: "truncate" → "truncates" but looks good otherwise!

eduardosm:
@rustbot ready

I also squashed everything into two commits

RalfJung:
Thanks a lot!
@bors r+

PR #3093, opened 2023-09-28
Implement the `llvm.x86.sse2.pmadd.wd` intrinsic

eduardosm:


src/shims/x86/sse2.rs
```diff
@@ -82,6 +82,39 @@ pub(super) trait EvalContextExt<'mir, 'tcx: 'mir>:
...
+
+                    let j2 = j1.checked_add(1).unwrap();
+                    let left2 = this.read_scalar(&this.project_index(&left, j2)?)?.to_i16()?;
+                    let right2 = this.read_scalar(&this.project_index(&right, j2)?)?.to_i16()?;
+
+                    let dest = this.project_index(&dest, i)?;
+
+                    let mul1 = i32::from(left1).checked_mul(right1.into()).unwrap();
+                    let mul2 = i32::from(left2).checked_mul(right2.into()).unwrap();
+                    let res = mul1.checked_add(mul2).unwrap();
```
saethlin, line 113: I think this ICEs if overflow happens during the multiply-add. Shouldn't this wrap on overflow?

src/shims/x86/sse2.rs
```diff
@@ -82,6 +82,39 @@ pub(super) trait EvalContextExt<'mir, 'tcx: 'mir>:
...
+            "pmadd.wd" => {
+                let [left, right] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let (left, left_len) = this.operand_to_simd(left)?;
+                let (right, right_len) = this.operand_to_simd(right)?;
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                assert_eq!(left_len, right_len);
+                assert_eq!(dest_len, left_len / 2);
```
RalfJung, line 98:
```suggestion
                assert_eq!(2 * dest_len, left_len);
```

src/shims/x86/sse2.rs
RalfJung: I assume this cannot overflow since the inputs are all `i16` that got casted to `i32`, multiplied, then added. But a comment along those lines would be good.

tests/pass/intrinsics-x86-sse2.rs
```diff
@@ -70,6 +70,16 @@ mod tests {
         }
         test_mm_avg_epu16();
 
+        #[target_feature(enable = "sse2")]
+        unsafe fn test_mm_madd_epi16() {
+            let a = _mm_setr_epi16(1, 2, 3, 4, 5, 6, 7, 8);
+            let b = _mm_setr_epi16(9, 10, 11, 12, 13, 14, 15, 16);
+            let r = _mm_madd_epi16(a, b);
+            let e = _mm_setr_epi32(29, 81, 149, 233);
+            assert_eq_m128i(r, e);
```
RalfJung, line 79: Can you add a test that exploits the fact that the i16 become i32 before being multiplied?

src/shims/x86/sse2.rs
eduardosm: It turns out it could overflow with `(-0x8000)*(-0x8000)+(-0x8000)*(-0x8000) = 0x80000000`.

RalfJung:
Thanks!
@bors r+

PR #3096, opened 2023-09-30
Bump libloading to 0.8

sthibaul:


sthibaul:
(libloading doesn't have a stable 0.7 branch, so we need to bump to the 0.8 series to get the patches for the newer OS ports)

RalfJung:
Makes sense, thanks.
@bors r+

PR #3098, opened 2023-10-01
Support getentropy on macOS as a foreign item

BlackHoleFox:
Prior this was always assumed to be accessed via `dlsym` shim, but in `std` I'm attempting to start [unconditionally linking](https://github.com/rust-lang/rust/pull/116319) to `getentropy` on macOS now that Rust's platform version support allows it.

This just moves the main logic of the previous `dlsym` handler into an eval context extension so it can be used via both call paths. The `dlsym` handler is still needed as `getrandom` uses it.

### DIFF
--- a/src/shims/unix/macos/dlsym.rs
+++ b/src/shims/unix/macos/dlsym.rs
@@ -2,6 +2,7 @@ use rustc_middle::mir;
 
 use log::trace;
 
+use super::foreign_items::EvalContextExt as _;
 use crate::*;
 use helpers::check_arg_count;
 
@@ -38,10 +39,8 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         match dlsym {
             Dlsym::getentropy => {
                 let [ptr, len] = check_arg_count(args)?;
-                let ptr = this.read_pointer(ptr)?;
-                let len = this.read_target_usize(len)?;
-                this.gen_random(ptr, len)?;
-                this.write_null(dest)?;
+                let result = this.getentropy(ptr, len)?;
+                this.write_scalar(result, dest)?;
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/shims/rand.rs
```diff
@@ -0,0 +1,19 @@
+use crate::*;
+
+impl<'mir, 'tcx: 'mir> EvalContextExt<'mir, 'tcx> for crate::MiriInterpCx<'mir, 'tcx> {}
+pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
+    fn getentropy(
```
RalfJung, line 5: Please move this to `macos/foreign_items` and call it from `dlsym`. It is macos-specific and should live in the macos module.

BlackHoleFox: Ah I can do that. I was following the pattern of `mach_timebase_info` but I guess there's not enough in this file for it to be worth existing.

RalfJung: Ah yeah for time we have all the shims for all OSes in one file. Those are old shims, we've been adjusting and organizing them differently more recently. (Also, following that pattern would mean putting all getrandom shims for all OSes into `shims/rand.v`.)

BlackHoleFox: Alrighty, its been moved to the macOS-specific module.

RalfJung:
As mentioned in the [other PR](https://github.com/rust-lang/rust/pull/116319#issuecomment-1742188338), you can do the Miri changes right then and there. I think that would be preferable -- you can then entirely remove the dlsym, supporting only the direct-linking case, and we can avoid having a shim that is not being tested.

RalfJung:
Ah never mind, `getrandom` sill uses `dlsym`, so we should probably support both.

BlackHoleFox:
~~I can't remove the `dlsym` functionality in these same changes, it breaks the tests that use `getrandom` (and any downstream crates also relying on that). So it is still being tested with this current changeset~~. Hah you typed out faster then me.

I'm good with moving the changes over there though since that appears easier for both of us. If you think what is in this branch right now is fine I'll close this PR and cherry-pick it over to `rust-lang/rust` instead.

RalfJung:
Right, the concern about tests was that the new shim (not via dlsym) is untested now. But I guess it will be tested soon and anyway it shares the code with the dlsym shim that we do test, so...

@bors r+

PR #3103, opened 2023-10-04
Extend the BorTag GC to AllocIds

saethlin:
As suggested in https://github.com/rust-lang/miri/issues/3080#issuecomment-1732505573

We previously solved memory growth issues associated with the Stacked Borrows and Tree Borrows runtimes with a GC. But of course we also have state accumulation associated with whole allocations elsewhere in the interpreter, and this PR starts tackling those.

To do this, we expand the visitor for the GC so that it can visit a BorTag or an AllocId. Then when the GC runs, we use the reachable set of AllocIds to remove elements from the intptrcast state.


src/tag_gc.rs
```diff
@@ -6,6 +6,7 @@ use crate::*;
 
 pub trait VisitTags {
     fn visit_tags(&self, visit: &mut dyn FnMut(BorTag));
+    fn visit_prov(&self, visit: &mut dyn FnMut(AllocId));
```
RalfJung, line 9: So now we have to copy-paste the traversal logic in every `impl VisitTags`, that's not great.

The alternative would be to make this `visit_provenance(&self, visit: &mut dyn FnMut(Provenance))`. But then the problem is that the SB stacks only have a `BorTag`, not an `AllocId`... and possibly somewhere we have an `AllocId` and not a `BorTag`? So what about
```
visit_provenance(&self, visit: &mut dyn FnMut(Option<AllocId>, Option<BorTag>))
```

saethlin: Originally I didn't want to do this because I thought we would have reason to run the AllocId collection much less often than the BorTag collection, but I think AllocIds are just so much more rare that it's not worth doing such an optimization.

src/tag_gc.rs
```diff
@@ -4,54 +4,62 @@ use rustc_data_structures::fx::FxHashSet;
 
 use crate::*;
 
+pub type TagVisitor<'a> = dyn FnMut(Option<AllocId>, Option<BorTag>) + 'a;
+
 pub trait VisitTags {
-    fn visit_tags(&self, visit: &mut dyn FnMut(BorTag));
+    fn visit_tags(&self, visit: &mut TagVisitor<'_>);
```
RalfJung, line 10:
```suggestion
    /// Will be called on all pointers in the state, as well as any
    /// other place where we store `AllocId` and/or `BorTag`.
    fn visit_tags(&self, visit: &mut TagVisitor<'_>);
```

src/tag_gc.rs
```diff
@@ -148,10 +156,13 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: MiriInterpCxExt<'mir, 'tcx> {
         }
 
         let mut tags = FxHashSet::default();
-        this.visit_tags(&mut |tag| {
-            tags.insert(tag);
+        let mut alloc_ids = FxHashSet::default();
+        this.visit_tags(&mut |id, tag| {
+            id.map(|id| alloc_ids.insert(id));
+            tag.map(|tag| tags.insert(tag));
```
RalfJung, line 162: I'm not a fan of calling `map` for side-effects only. Can you make this `if let`, or maybe `.iter().for_each` if you want to optimize for line numbers?

src/intptrcast.rs
```diff
@@ -45,8 +45,10 @@ pub struct GlobalStateInner {
 }
 
 impl VisitTags for GlobalStateInner {
-    fn visit_tags(&self, _visit: &mut dyn FnMut(BorTag)) {
-        // Nothing to visit here.
+    fn visit_tags(&self, visit: &mut TagVisitor<'_>) {
+        for id in &self.exposed {
```
RalfJung, line 49: Please add a comment explaining why we can skip `int_to_ptr_map` and `base_addr`.

src/intptrcast.rs
```diff
@@ -63,6 +65,11 @@ impl GlobalStateInner {
 }
 
 impl<'mir, 'tcx> GlobalStateInner {
+    pub fn remove_unreachable_allocs(&mut self, live_allocs: &FxHashSet<AllocId>) {
+        self.int_to_ptr_map.retain(|(_, id)| live_allocs.contains(id));
```
RalfJung, line 69: Removing things from `int_to_ptr_map` will mean we return a different `AllocId` when an address inside that allocation is looked up. Why does that not change the semantics?

src/tag_gc.rs
RalfJung, line 1: I think we want to name this file and the trait, since it's no longer just about tags. `provenance_gc`/`VisitProvenance`, maybe?

src/tag_gc.rs
```diff
@@ -4,54 +4,62 @@ use rustc_data_structures::fx::FxHashSet;
 
 use crate::*;
 
+pub type TagVisitor<'a> = dyn FnMut(Option<AllocId>, Option<BorTag>) + 'a;
```
RalfJung, line 7: This is not the visitor though, it is the function called by the visitor. I don't have immediate ideas for a better name but this name is potentially confusing.

tests/pass/ptr_int_casts.rs
```diff
@@ -46,6 +46,26 @@ fn ptr_int_casts() {
 
     // involving types other than usize
     assert_eq!((-1i32) as usize as *const i32 as usize, (-1i32) as usize);
+
+    // Stress-test that the GC doesn't delete context that would prevent us from casting from int
+    // to pointer correctly after the allocation is dead.
+    let mut ptrs: Vec<*const u8> = Vec::new();
+    let mut ints: Vec<usize> = Vec::new();
+
+    for _ in 0..1_000 {
```
RalfJung, line 55: Do we really need 1000 here or is 100 enough? We [run the GC after every single machine step on CI](https://github.com/rust-lang/miri/blob/f6c0b68dbbdf75b4e865f939407c142c4c6c5724/.github/workflows/ci.yml#L40) so we should be detecting issues with fairly low iteration counts, even just a single iteration in principle.

src/intptrcast.rs
RalfJung, line 1: If you are touching this file anyway, can I ask you to update the comment on int_to_ptr_map?
```rust
    /// This is used as a map between the address of each allocation and its `AllocId`. It is always
    /// sorted. We cannot use a `HashMap` since we can be given an address that is offset from the
    /// base address, and we need to find the `AllocId` it belongs to.
    int_to_ptr_map: Vec<(u64, AllocId)>,
```

tests/pass/ptr_int_casts.rs
saethlin: 100 is enough here. I just like powers of 1000 :p

saethlin: We don't actually run the GC after every machine step. I considered making the GC run every N machine steps, but it runs every N terminators because that's how preemption and the clock work. I'm definitely not opposed to shifting all such logic to machine steps.

src/intptrcast.rs
saethlin: I've been unable to craft an example that produces any problematic behavior. I think this data structure is only used when we are immediately using an absolute pointer, either to deref it or to reallocate/deallocate it. So we can fail to select an allocation instead of finding the dead allocation, which doesn't change whether the interpreter step is about to report UB.

But I've spent so much time thinking about this situation, this code is so complicated in how everything cooperates indirectly. I might see if there's a simpler way to write this that's more obviously consistent. I think we're going to have all these same problems when we try to reuse addresses.

tests/pass/ptr_int_casts.rs
RalfJung: Ah, every N blocks is probably better to avoid completely grinding Miri to a halt.^^

But even then, there's a bunch of function calls here, leading to tons of blocks. So we don't even need a loop, I think?

src/intptrcast.rs
RalfJung: >  I think this data structure is only used when we are immediately using an absolute pointer, either to deref it or to reallocate/deallocate it. So we can fail to select an allocation instead of finding the dead allocation, which doesn't change whether the interpreter step is about to report UB.

Yes I think that's true. But does "unreachable" imply "dead"? You're removing AllocId that aren't mentioned anywhere, but what if an actually existing allocation isn't mentioned anywhere?

Also, can it change which kind of UB is reported? That wouldn't be great either.

I think for now it'd be better to now remove things from `int_to_ptr_map` ever. Which probably means we should add the `AllocId` in that map to the live set.

> But I've spent so much time thinking about this situation, this code is so complicated in how everything cooperates indirectly. I might see if there's a simpler way to write this that's more obviously consistent. I think we're going to have all these same problems when we try to reuse addresses.

I did spend a lot of time on our integer address  system to make it as nice as I could, but if you have ideas for cleaning it up further then I'm curious to see them. :) The tricky part is having all the generic code in the interpreter also work for pointers with relative addresses (for const-eval).

saethlin: > But does "unreachable" imply "dead"?

No, but I've fixed this now by adding all live AllocIds to the set. Does that address your other concerns here? `GlobalStateInner::alloc_id_from_addr` checks that the AllocId it is going to return is live, so what we do with dead allocations should be irrelevant.

RalfJung: Ah, and `alloc_id_from_addr` also checks that we are in-bounds. So it's always an in-bounds address of a live allocation that we return, and therefore we can indeed remove dead allocations. Yes that should work.

RalfJung: Ah no, there's still one case where this changes behavior. If we have two allocations directly adjacent to each other, then the address "at the end" of the first equals the base address of the 2nd. Once the 2nd allocation becomes dead, as long as it remains in the `int_to_ptr_map`, `alloc_id_from_addr` will return `None` for the base address of the 2nd allocation. But if we remove the 2nd allocation from the `int_to_ptr_map`, then it will now return `Some(first_alloc_id)`.

tests/pass/ptr_int_casts.rs
```diff
@@ -4,6 +4,10 @@
 use std::mem;
 use std::ptr;
 
+extern "Rust" {
+    fn miri_run_provenance_gc();
+}
```
RalfJung, line 9: Please add this to tests/utils/miri_extern.rs instead

saethlin:
This isn't ready, I'm still thinking about your two comments I didn't mark as resolved.

src/intptrcast.rs
saethlin: We're now managing `into_to_ptr_map` without the GC. Do we still need to add all its entries to the reachable set?

RalfJung: The GC should probably entirely ignore `int_to_ptr_map` -- neither scan it, nor clean it up.

That said... do we have all live allocations in the reachable set? Cleaning up allocations that are still live seems a bit risky.^^ If we have all live allocations already in there, then scanning `int_to_ptr_map` anyway can't add anything new.

saethlin: Oh yeah 🤦 that's another thing I was going to fix

saethlin: I added some code to explicitly initialize the collection with all live allocation IDs.

src/provenance_gc.rs
```diff
@@ -0,0 +1,196 @@
...
+
+impl<'mir, 'tcx: 'mir> EvalContextExt<'mir, 'tcx> for crate::MiriInterpCx<'mir, 'tcx> {}
+pub trait EvalContextExt<'mir, 'tcx: 'mir>: MiriInterpCxExt<'mir, 'tcx> {
+    fn run_provenance_gc(&mut self) {
+        let this = self.eval_context_mut();
+
+        let mut tags = FxHashSet::default();
+
+        // Initially populate the set of reachable AllocIds with all live allocations. This ensures
+        // that we do not remove state associated with leaked allocations.
```
RalfJung, line 158:
```suggestion
        // that we do not remove state associated with leaked allocations.
        // Here we exploit that `adjust_allocation` always returns `Owned`, to all
        // tcx-managed allocations ever read or written will be copied in `alloc_map`.
```

src/provenance_gc.rs
```diff
@@ -0,0 +1,196 @@
...
+        let mut alloc_ids = FxHashSet::default();
+        this.memory.alloc_map().iter(|it| {
+            for (id, _) in it {
+                alloc_ids.insert(*id);
+            }
+        });
```
RalfJung, line 164: Shouldn't this be in `MiriInterpCx::visit_provenance`? You can merge it with the `self.memory.alloc_map().iter` that's already there.

tests/utils/miri_extern.rs
```diff
@@ -7,8 +7,7 @@ pub struct MiriFrame {
     // The size of filename of the function being executed, encoded in UTF-8
     pub filename_len: usize,
     // The line number currently being executed in `filename`, starting from '1'.
-    pub lineno: u32,
-    // The column number currently being executed in `filename`, starting from '1'.
+    pub lineno: u32, // The column number currently being executed in `filename`, starting from '1'.
```
RalfJung, line 10: This diff looks like a mistake?

tests/utils/miri_extern.rs
```diff
@@ -137,4 +136,9 @@ extern "Rust" {
         out: *mut std::ffi::c_char,
         out_size: usize,
     ) -> usize;
+
+    /// Run the provenance GC. The GC will run automatically at some cadence, but tests we want to
+    /// have control of when it runs so that we can run it for sure at certain points to make sure
+    /// that it doesn't break anything.
```
RalfJung, line 142:
```suggestion
    /// Run the provenance GC. The GC will run automatically at some cadence,
    /// but in tests we want to for sure run it at certain points to check
    /// that it doesn't break anything.
```

src/intptrcast.rs
```diff
@@ -46,9 +46,23 @@ pub struct GlobalStateInner {
...
+            exposed,
+            next_base_addr: _,
+            provenance_mode: _,
+        } = self;
+        // Though base_addr and int_to_ptr_map contain AllocIds, we do not want to visit them.
+        // We're trying to remove unused elements from base_addr, so visiting it would prevent
+        // removing anything. int_to_ptr_map is managed by free_alloc_id, and entries in it do not
+        // actually make allocation base addresses reachable so we don't need to visit it.
+        // But exposed tags do make base addresses reachable.
+        for id in exposed {
```
RalfJung, line 63: There's no point in visiting exposed since it contains only live allocations, and those have already been added.
At least that should be true, we remove the allocation from `exposed` in `free_alloc_id`... but currently we don't check that it is live in `expose_ptr` *oops*

Similarly, the argument for not visiting `int_to_ptr_map` should be that it only contains live allocations.

Only base_addr needs a different argument: that map is only ever relevant when we *have* a pointer of given `AllocId` and need to know its base address; if the AllocId exists nowhere else we can remove it here.

src/provenance_gc.rs
saethlin: Yes, it's already handled there :facepalm:

RalfJung: Well so far it only added the tags there, not the AllocIds, I think?

src/intptrcast.rs
saethlin: > but currently we don't check that it is live in `expose_ptr` _oops_

This sounds like a good thing to add a check for; but `MonoHashMap::contains_key` takes `&mut self` so the call is tricky to pull off here. I think the receiver should be changed to `&self`, but that would require changing the trait... I'll do that PR to the main repo.

RalfJung: I am confused, can't we just add a liveness check trivially around [here](https://github.com/rust-lang/miri/blob/622a4d35b4a3f2404caa6834429d1ff452f5b27d/src/intptrcast.rs#L186)?

saethlin: Maybe I'm just doing it wrong, but I tried to do `ecx.machine.alloc_map().contains_key`.

RalfJung: No, that's a *super* low-level function, don't use it.

`ecx.get_alloc_info(alloc_id)` returns `AllocKind::Dead` for dead allocations and something else for live allocations; you can use that.

saethlin: Of course, there are even other checks just like that in this file.

src/intptrcast.rs
```diff
@@ -46,9 +46,21 @@ pub struct GlobalStateInner {
...
+    fn visit_provenance(&self, _visit: &mut VisitWith<'_>) {
+        let GlobalStateInner {
+            int_to_ptr_map: _,
+            base_addr: _,
+            exposed: _,
+            next_base_addr: _,
+            provenance_mode: _,
+        } = self;
+        // Though base_addr, int_to_ptr_map, and exposed contain AllocIds, we do not want to visit them.
+        // int_to_ptr_map and exposed must contain live allocations, and those were already handled
```
RalfJung, line 59:
```suggestion
        // int_to_ptr_map and exposed must contain only live allocations, and those were already handled
```

src/provenance_gc.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+}
+
+impl VisitProvenance for crate::MiriInterpCx<'_, '_> {
+    fn visit_provenance(&self, visit: &mut VisitWith<'_>) {
+        // Visit the contents of the allocations and the IDs themselves, to account for all
+        // live allocation IDs and all provenance in the allocation bytes, even if they are leaked.
+        // Here we exploit that `adjust_allocation` always returns `Owned`, to all
+        // tcx-managed allocations ever read or written will be copied in `alloc_map`.
+        self.memory.alloc_map().iter(|it| {
```
RalfJung, line 141: We also have the secondary memory map in `tcx`. I guess we have to walk that one as well, judging from the test failures. The comment I suggested about not having to worry about tcx-managed allocation seems to be wrong.

In particular it is wrong for function and vtable allocations, which never get copied into the local `alloc_map`.

src/intptrcast.rs
```diff
@@ -62,6 +74,10 @@ impl GlobalStateInner {
             provenance_mode: config.provenance_mode,
         }
     }
+
+    pub fn remove_unreachable_allocs(&mut self, reachable_allocs: &FxHashSet<AllocId>) {
+        self.base_addr.retain(|id, _| reachable_allocs.contains(id));
```
RalfJung, line 79:
```suggestion
        // `exposed` and `int_to_ptr_map` are cleared immediately when an allocation
        // is freed, so `base_addr` is the only one we have to clean up based on the GC.
        self.base_addr.retain(|id, _| reachable_allocs.contains(id));
```

src/intptrcast.rs
```diff
@@ -179,6 +195,9 @@ impl<'mir, 'tcx: 'mir> EvalContextExt<'mir, 'tcx> for crate::MiriInterpCx<'mir,
 pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
     fn expose_ptr(&mut self, alloc_id: AllocId, tag: BorTag) -> InterpResult<'tcx> {
         let ecx = self.eval_context_mut();
+        // Make sure we aren't trying to expose a dead allocation; this is part of justifying the
+        // way we treat the exposed set in VisitProvenance.
+        assert!(!matches!(ecx.get_alloc_info(alloc_id).2, AllocKind::Dead));
```
RalfJung, line 200: Exposing a dead allocation is totally legal. It just shouldn't add the allocation to `exposed`, since you can't int2ptr cast it anyway.

src/provenance_gc.rs
```diff
@@ -0,0 +1,191 @@
...
+            }
+        });
+    }
+
+    fn remove_unreachable_allocs(&mut self, allocs: FxHashSet<AllocId>) {
+        let this = self.eval_context_mut();
+        this.machine.allocation_spans.borrow_mut().retain(|id, _| allocs.contains(id));
+        this.machine.intptrcast.borrow_mut().remove_unreachable_allocs(&allocs);
+        if let Some(borrow_tracker) = &this.machine.borrow_tracker {
+            borrow_tracker.borrow_mut().base_ptr_tags.retain(|id, _| allocs.contains(id));
```
RalfJung, line 188: When https://github.com/rust-lang/miri/pull/3163 landed you won't be able to access `base_ptr_tags` directly any more; please add a `remove_unreachable_allocs` method instead of directly accessing the struct fields.

src/intptrcast.rs
```diff
@@ -197,17 +197,20 @@ impl<'mir, 'tcx: 'mir> EvalContextExt<'mir, 'tcx> for crate::MiriInterpCx<'mir,
 pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
     fn expose_ptr(&mut self, alloc_id: AllocId, tag: BorTag) -> InterpResult<'tcx> {
         let ecx = self.eval_context_mut();
-        // Make sure we aren't trying to expose a dead allocation; this is part of justifying the
-        // way we treat the exposed set in VisitProvenance.
-        assert!(!matches!(ecx.get_alloc_info(alloc_id).2, AllocKind::Dead));
+        // Exposing a dead alloc is a no-op, because it's not possible to get a dead allocation
+        // via int2ptr.
+        if matches!(ecx.get_alloc_info(alloc_id).2, AllocKind::Dead) {
```
RalfJung, line 202: Please check this *after* `provenance_mode`. I doubt it matters, but still -- the cheap check should come first, and `get_alloc_info` can potentially require a query.

saethlin:
CI has been running for an hour. That's not good. I'll look into it.

saethlin:
The slow part of the CI is inserting all the live AllocIds into the reachable set. Aside from asking the interpreter to track a `FxHashSet<AllocId>` for us then hand it to the GC via clone I'm not sure what else to do other than run the GC less often, but the CI is slow because we're running the GC super often. Maybe trim down the size of that test?

The cde that's causing the GC to lose track of a pointer is here:
https://github.com/rust-random/getrandom/blob/bd14b7a243c7cda78f698742a84d1a0bbfcba972/src/macos.rs#L14-L21
```rust
type GetEntropyFn = unsafe extern "C" fn(*mut u8, libc::size_t) -> libc::c_int;

pub fn getrandom_inner(dest: &mut [u8]) -> Result<(), Error> {
    static GETENTROPY: Weak = unsafe { Weak::new("getentropy\0") };
    if let Some(fptr) = GETENTROPY.ptr() {
        let func: GetEntropyFn = unsafe { mem::transmute(fptr) };
        for chunk in dest.chunks_mut(256) {
            let ret = unsafe { func(chunk.as_mut_ptr(), chunk.len()) };
```

RalfJung:
> The slow part of the CI is inserting all the live AllocIds into the reachable set. Aside from asking the interpreter to track a FxHashSet<AllocId> for us then hand it to the GC via clone I'm not sure what else to do other than run the GC less often, but the CI is slow because we're running the GC super often. Maybe trim down the size of that test?

That test? We run *all* tests on Linux with "do GC after every step".

https://github.com/rust-lang/miri/blob/d36c3e7cac5d97aefd86f0a2ecb93e9b29f70324/.github/workflows/ci.yml#L38-L40

And that seems to be good, we caught a bug?

For performance, here's one idea: we don't actually need the set. We need a function `fn(AllocId) -> bool` that tells us whether the AllocId is used anywhere. So instead of adding all live allocations to the set, we can make that function a 3-way disjunction, where it checks the global alloc_map, the local alloc_map, and the collected set. That might help?

RalfJung:
> The cde that's causing the GC to lose track of a pointer is here, and it has a wild diagnostic:

Ah, that's the old macOS getrandom doing strange integer pointer shenanigans. If this triggers a GC bug then the first thing to do is extract it into a dedicated test so that we don't have to rely on a pass-dep test to catch this. It should then also reproduce on all targets; the reason it's macOS-only is that getrandom uses different code for different targets.

saethlin:
Here's a minimization:
```rust
#![feature(rustc_private)]
extern crate libc;

extern "Rust" {
    pub fn miri_run_provenance_gc();
}

type GetEntropyFn = unsafe extern "C" fn(*mut u8, libc::size_t) -> libc::c_int;

fn main() {
    let name = "getentropy\0";
    let addr = unsafe { libc::dlsym(libc::RTLD_DEFAULT, name.as_ptr() as *const _) as usize };
    unsafe {
        miri_run_provenance_gc();
    }

    let ptr = addr as *mut libc::c_void;
    let func: GetEntropyFn = unsafe { std::mem::transmute(ptr) };
    let dest = &mut [0u8];
    unsafe { func(dest.as_mut_ptr(), dest.len()) };
}
```
Run with `./miri run -Zmiri-tag-gc=0` to get the error we get in CI.

saethlin:
Note that this code runs fine without explicitly invoking the provenance GC, because in the minimization there is no basic block end in the important region of the program. I always add code to run the GC on every machine step while debugging to avoid getting tripped up by this.

saethlin:
Ah! I think the GC isn't picking up the contents of `extra_fn_ptr_map`: https://github.com/rust-lang/rust/blob/4d7f952a02d0bca67c98a6b74895b7e3fbe38341/compiler/rustc_const_eval/src/interpret/memory.rs#L106

saethlin:
I think this is going to be easier to do as a rustc PR, so that I can iterate on the interface between Miri and the interpreter core code. I'll set up that later today.

RalfJung:
> I think this is going to be easier to do as a rustc PR, so that I can iterate on the interface between Miri and the interpreter core code. I'll set up that later today.

However, rustc doesn't run the gc-interval=1 tests.

RalfJung:
> Ah! I think the GC isn't picking up the contents of `extra_fn_ptr_map`: https://github.com/rust-lang/rust/blob/4d7f952a02d0bca67c98a6b74895b7e3fbe38341/compiler/rustc_const_eval/src/interpret/memory.rs#L106

Aha!

If we follow my proposal, the API we need is an efficient "is this AllocId live". So that's a single `fn(&self, AllocId) -> bool`, I think. Basically a more efficient version of `get_alloc_info` that doesn't compute the size. Maybe we just want that function to be split into two, one for size+align and one for `AllocKind`?

saethlin:
Subsumed by https://github.com/rust-lang/rust/pull/118029

PR #3107, opened 2023-10-05
Update dependencies

eduardosm:


Cargo.toml
```diff
@@ -39,10 +39,10 @@ colored = "2"
 ui_test = "0.21.1"
 rustc_version = "0.4"
 # Features chosen to match those required by env_logger, to avoid rebuilds
-regex = { version = "1.5.5", default-features = false, features = ["perf", "std"] }
+regex = { version = "1.9.6", default-features = false, features = ["perf", "std"] }
 lazy_static = "1.4.0"
 # Require a version of serde without intransparent unreproducible binary blobs.
-serde = { version = "1.0.185", features = ["derive"] }
+serde = { version = "1.0.188", features = ["derive"] }
```
RalfJung, line 45: Why this change? This is still the same semver bracket.

(Also applies for some of the other toml file changes)

eduardosm: What is the criterion? Specify the latest version available at the time the dependency is added and keep it frozen forever in `Cargo.toml` and only update it in `Cargo.lock`?

RalfJung: That's what we did so far. If we want something more principled it should be the minimum number that works (and we should have a CI job that uses minimal-version resolution -- which might nor work since some widely used ecosystem crates historically didn't care for working builds under minimal versions).

eduardosm: So be it

RalfJung:
Thanks!
@bors r+

cargo-miri/Cargo.lock
```diff
@@ -73,33 +82,34 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
...
 [[package]]
 name = "dirs-sys"
-version = "0.3.7"
+version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1b1d1d91c932ef41c0f2663aa8b0ca0342d444d842c06914aa0a7e352d0bada6"
+checksum = "520f05a5cbd335fae5a99ff7a6ab8627577660ee5cfd6a94a6a929b52ff0321c"
 dependencies = [
  "libc",
+ "option-ext",
```
RalfJung, line 99: option-ext has a license that rustc's tidy doesn't like... I'm going to partially revert this PR until we figure that out.

PR #3110, opened 2023-10-06
Do not use host floats in `simd_{ceil,floor,round,trunc}`

eduardosm:


RalfJung:
Ah, that's how apfloat exposes these operations, nice .:) Thanks a lot!
@bors r+

PR #3111, opened 2023-10-06
downgrade directories to v4 again

RalfJung:
rustc tidy says
```
tidy error: invalid license `MPL-2.0` in `option-ext 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)`
```


RalfJung:
Ah looks like we already have some MPL-2.0 exceptions and it's fine as long as it's only in rustc tooling, not the runtime.
@bors r-

PR #3118, opened 2023-10-10
Implement `llvm.x86.sse41.*` intrinsics

eduardosm:


src/shims/x86/sse41.rs
```diff
@@ -0,0 +1,312 @@
...
+    ) -> InterpResult<'tcx, EmulateForeignItemResult> {
+        let this = self.eval_context_mut();
+        // Prefix should have already been checked.
+        let unprefixed_name = link_name.as_str().strip_prefix("llvm.x86.sse41.").unwrap();
+
+        match unprefixed_name {
+            // Used to implement the _mm_insert_ps function.
+            // Takes one element of `right` and inserts it into `left` and
+            // optionally zero some elements. Source index, destination index
+            // and zeroed indices are specified by `imm`.
```
RalfJung, line 28: How are they specified by `imm`? Seems to be some bitcoding magic?

src/shims/x86/sse41.rs
```diff
@@ -0,0 +1,312 @@
...
+                let dst_index = u64::from((imm >> 4) & 0b11);
+
+                for i in 0..dest_len {
+                    let dest = this.project_index(&dest, i)?;
+
+                    if imm & (1 << i) != 0 {
+                        // zeroed
+                        this.write_scalar(Scalar::from_u32(0), &dest)?;
+                    } else if i == dst_index {
+                        // copy from `right`
```
RalfJung, line 51: So if the `i`th bit is set then "zeroed" takes precedence over "dst_index"? Do we have a test checking this?

src/shims/x86/sse41.rs
```diff
@@ -0,0 +1,312 @@
...
+                        this.copy_op(
+                            &this.project_index(&right, src_index)?,
```
RalfJung, line 53: Might be better to `read_immediate` this before the loop, to ensure it always gets read and validated even if `dst_index` is being zeroed

src/shims/x86/sse41.rs
```diff
@@ -0,0 +1,312 @@
...
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                assert_eq!(left_len, right_len);
+                assert_eq!(dest_len, left_len.checked_mul(2).unwrap());
+
+                for i in 0..left_len {
+                    let left = this.read_scalar(&this.project_index(&left, i)?)?.to_i32()?;
+                    let right = this.read_scalar(&this.project_index(&right, i)?)?.to_i32()?;
+                    let left_dest = this.project_index(&dest, i)?;
+                    let right_dest = this.project_index(&dest, i.checked_add(left_len).unwrap())?;
```
RalfJung, line 85: Ah so they don't get interleaved, they get concatenated?

src/shims/x86/sse41.rs
```diff
@@ -0,0 +1,312 @@
...
+            // 4 bits of `imm`.
+            "dpps" | "dppd" => {
+                let [left, right, imm] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let (left, left_len) = this.operand_to_simd(left)?;
+                let (right, right_len) = this.operand_to_simd(right)?;
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                assert_eq!(left_len, right_len);
```
RalfJung, line 109: Is there no constraint at all on dest_len?

src/shims/x86/sse41.rs
```diff
@@ -0,0 +1,312 @@
...
+                assert_eq!(dest_len, right_len);
+
+                let imm = this.read_scalar(imm)?.to_u8()?;
+                let src_index = u64::from((imm >> 6) & 0b11);
+                let dst_index = u64::from((imm >> 4) & 0b11);
+
+                for i in 0..dest_len {
+                    let dest = this.project_index(&dest, i)?;
+
+                    if imm & (1 << i) != 0 {
```
RalfJung, line 47: I assume here we have `i < 4`?

src/shims/x86/sse41.rs
```diff
@@ -0,0 +1,312 @@
...
+                        &this.project_index(&left, i)?,
+                        &this.project_index(&dest, i)?,
+                        /*allow_transmute*/ false,
+                    )?;
+                }
+            }
+            // Used to implement the _mm_floor_sd, _mm_ceil_sd and _mm_round_sd
+            // functions. Rounds the first element of `right` according to `rounding`
+            // and copies the remaining from `left`.
+            "round.sd" => {
```
RalfJung, line 177: This looks like it could share the implementation with the float ("s") version.

src/shims/x86/sse41.rs
```diff
@@ -0,0 +1,312 @@
...
+
+                let (op, op_len) = this.operand_to_simd(op)?;
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                // Find minimum
+                let mut min_value = u16::MAX;
+                let mut min_index = 0;
+                for i in 0..op_len {
+                    let op = this.read_scalar(&this.project_index(&op, i)?)?.to_u16()?;
+                    if op < min_value {
```
RalfJung, line 222: Do we have a test that distinguishes `<` from `<=` here? That obviously returns the same value, but not the same index.

src/shims/x86/sse41.rs
```diff
@@ -0,0 +1,312 @@
...
+
+                let (left, left_len) = this.operand_to_simd(left)?;
+                let (right, right_len) = this.operand_to_simd(right)?;
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                assert_eq!(left_len, right_len);
+                assert_eq!(left_len, dest_len.checked_mul(2).unwrap());
+
+                let imm = this.read_scalar(imm)?.to_u8()?;
+                let left_offset = u64::from((imm >> 2) & 1).checked_mul(4).unwrap();
```
RalfJung, line 257: `((imm >> 2) & 1) * 4` definitely needs a comment explaining what is being decoded here.

src/shims/x86/sse41.rs
eduardosm: I will extend the comment.

src/shims/x86/sse41.rs
eduardosm: There will be.

src/shims/x86/sse41.rs
eduardosm: I will add a `<= 4` constraint.

src/shims/x86/sse41.rs
eduardosm: There will be.

RalfJung:
The question about the "imm" bitcoding applies to all the intrinsics that use bitmask operations on "imm", and same for the size constraint - if the code only makes sense for size 4, then that should be asserted.

src/shims/x86/sse41.rs
```diff
@@ -0,0 +1,310 @@
...
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                assert_eq!(left_len, right_len);
+                assert_eq!(left_len, dest_len.checked_mul(2).unwrap());
+
+                let imm = this.read_scalar(imm)?.to_u8()?;
+                // Bit 2 of `imm` specifies the offset for indices of `left`.
+                // The offset is 0 when the bit is 0 or 4 when the bit is 1.
+                let left_offset = u64::from((imm >> 2) & 1).checked_mul(4).unwrap();
+                // Bits 0..=1 of `imm` specifies the offset for indices of
```
RalfJung, line 212:
```suggestion
                // Bits 0..=1 of `imm` specify the offset for indices of
```

src/shims/x86/sse41.rs
```diff
@@ -0,0 +1,310 @@
...
+                this.write_scalar(Scalar::from_i32(all_zero.into()), dest)?;
+            }
+            _ => return Ok(EmulateForeignItemResult::NotSupported),
+        }
+        Ok(EmulateForeignItemResult::NeedsJumping)
+    }
+}
+
+// Rounds the first element of `right` according to `rounding`
+// and copies the remaining from `left`.
```
RalfJung, line 271:
```suggestion
// and copies the remaining elements from `left`.
```

RalfJung:
Looks great, thanks. :-)
@bors r+

PR #3122, opened 2023-10-14
intptrcast: remove information about dead allocations

RalfJung:
The discussion in https://github.com/rust-lang/miri/pull/3103 convinced me we don't have to keep `int_to_ptr_map` around for dead allocations. But we should not make that depend on the GC, we can just tie it to when the allocation gets freed. That means everything still behaves deterministically, if anything weird happens (but it shouldn't).

r? @saethlin 
Only the first and last commit contain logic changes, the 2nd commit just moves code around a bit.

### DIFF
--- a/src/intptrcast.rs
+++ b/src/intptrcast.rs
@@ -82,9 +82,12 @@ impl<'mir, 'tcx> GlobalStateInner {
                 let (glb, alloc_id) = global_state.int_to_ptr_map[pos - 1];
                 // This never overflows because `addr >= glb`
                 let offset = addr - glb;
-                // If the offset exceeds the size of the allocation, don't use this `alloc_id`.
+                // We require this to be strict in-bounds of the allocation. This arm is only
+                // entered for addresses that are not the base address, so even zero-sized
+                // allocations will get recognized at their base address -- but all other
+                // allocations will *not* be recognized at their "end" address.
                 let size = ecx.get_alloc_info(alloc_id).0;
-                if offset <= size.bytes() { Some(alloc_id) } else { None }
+                if offset < size.bytes() { Some(alloc_id) } else { None }
             }
         }?;
 
--- a/src/intptrcast.rs
+++ b/src/intptrcast.rs
@@ -62,10 +62,21 @@ impl GlobalStateInner {
     }
 }
 
-impl<'mir, 'tcx> GlobalStateInner {
+/// Shifts `addr` to make it aligned with `align` by rounding `addr` to the smallest multiple
+/// of `align` that is larger or equal to `addr`
+fn align_addr(addr: u64, align: u64) -> u64 {
+    match addr % align {
+        0 => addr,
+        rem => addr.checked_add(align).unwrap() - rem,
+    }
+}
+
+impl<'mir, 'tcx: 'mir> EvalContextExtPriv<'mir, 'tcx> for crate::MiriInterpCx<'mir, 'tcx> {}
+trait EvalContextExtPriv<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
     // Returns the exposed `AllocId` that corresponds to the specified addr,
     // or `None` if the addr is out of bounds
-    fn alloc_id_from_addr(ecx: &MiriInterpCx<'mir, 'tcx>, addr: u64) -> Option<AllocId> {
+    fn alloc_id_from_addr(&self, addr: u64) -> Option<AllocId> {
+        let ecx = self.eval_context_ref();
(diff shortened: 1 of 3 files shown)


RalfJung:
Uh, what... I was sure I ran the tests locally...

RalfJung:
Okay, `base_addr` needs to stay, but ``int_to_ptr_map`` can be removed.

saethlin:
> Okay, `base_addr` needs to stay, but `int_to_ptr_map` can be removed.

That's because we need the GC to be able to delete entries from `base_addr`, right?

RalfJung:
Yes.

src/intptrcast.rs
```diff
@@ -260,14 +261,17 @@ impl<'mir, 'tcx> GlobalStateInner {
...
-            rem => addr.checked_add(align).unwrap() - rem,
-        }
+impl GlobalStateInner {
+    pub fn free_alloc_id(&mut self, dead_id: AllocId) {
+        // We can *not* remove this from `base_addr`, since `addr_from_alloc_id` is called on each
+        // attempt at a memory access to determine the allocation ID and offset -- and there can
+        // still be pointers with `dead_id` that one can attempt to use for a memory access.
+        // However, we *can* remove it from `int_to_ptr_map`, since any wildcard pointers that exist
+        // can no longer actually be accessing that address. This ensures `alloc_id_from_addr` never
+        // returns a dead allocation.
```
saethlin, line 273: Can we also remove the AllocId from `exposed`?

src/intptrcast.rs
```diff
@@ -260,14 +261,17 @@ impl<'mir, 'tcx> GlobalStateInner {
...
-    fn align_addr(addr: u64, align: u64) -> u64 {
-        match addr % align {
-            0 => addr,
-            rem => addr.checked_add(align).unwrap() - rem,
-        }
+impl GlobalStateInner {
+    pub fn free_alloc_id(&mut self, dead_id: AllocId) {
+        // We can *not* remove this from `base_addr`, since `addr_from_alloc_id` is called on each
+        // attempt at a memory access to determine the allocation ID and offset -- and there can
+        // still be pointers with `dead_id` that one can attempt to use for a memory access.
```
saethlin, line 270: I don't understand why this is problematic for accesses. Even if we remove an allocation which still has pointers to it, the pointer is still dangling right? I expected the problem here to be converting pointers to freed allocations to integers.

RalfJung: When such a dangling pointer gets used, the first thing that happens is that we call `ptr_get_alloc`, which is expected to return the `AllocId` and the offset inside the allocation. Once we remove an allocation from `base_addr` we can no longer determine the offset, so we have a problem.

src/intptrcast.rs
RalfJung: Yes we can. I added that.

src/intptrcast.rs
saethlin: In what case would we need to know the offset inside the allocation and also be okay with using a pointer to a deallocated allocation?

RalfJung: You are implicitly suggesting a rework of the `Machine` trait here. And it's not super trivial since in error messages we want to distinguish "ptr with no provenance" from "ptr with provenance of a dangling allocation", so just returning `None` when asked for "what allocid + offset is this provenance" would not work.

So this can probably be done but it's a bigger change than what this PR does. So far the API is designed around the idea that "get allocid + offset" and "determine whether alloc is live and offset inbounds" are two separate independent steps.

RalfJung:
@saethlin any objections to landing this, any remaining ideas for how to improve this?

I think making `base_addr` cleanable is a larger refactor and I'm not yet convinced it's worth the impact on the rest of the code. It's definitely not something I want to do in this PR.

src/intptrcast.rs
saethlin: > You are implicitly suggesting a rework of the Machine trait here.

I didn't mean to suggest that. I think this whole situation you're describing is part of why I was so confused, I don't know the interpreter internals well enough to reason from them as opposed to from the language. And to this specific question of mine above, I was half expecting you to produce some obscure Rust code in answer.

Can you add a short comment to indicate that the design of the interpreter requires us to be able to do "get alloc_id + offset" _before_ we do UB detection? Perhaps something like

```rust
        // We can *not* remove this from `base_addr`, since the interpreter design requires that we be able to retrieve
        // an AllocId + offset for any memory access *before* we check if the access is valid.
```

saethlin: r=me with any such tweak

RalfJung: Ah okay. :) I incorporated that into the comment.

PR #3124, opened 2023-10-16
Fix rounding mode check in SSE4.1 round functions

eduardosm:
Now it masks out the correct bit and adds some explanatory comments. Also extends the tests.

### DIFF
--- a/src/shims/x86/sse41.rs
+++ b/src/shims/x86/sse41.rs
@@ -283,11 +283,20 @@ fn round_first<'tcx, F: rustc_apfloat::Float>(
     assert_eq!(dest_len, left_len);
     assert_eq!(dest_len, right_len);
 
-    let rounding = match this.read_scalar(rounding)?.to_i32()? & !0x80 {
-        0x00 => rustc_apfloat::Round::NearestTiesToEven,
-        0x01 => rustc_apfloat::Round::TowardNegative,
-        0x02 => rustc_apfloat::Round::TowardPositive,
-        0x03 => rustc_apfloat::Round::TowardZero,
+    // The fourth bit of `rounding` only affects the SSE status
+    // register, which cannot be accessed from Miri (or from Rust,
+    // for that matter), so we can ignore it.
+    let rounding = match this.read_scalar(rounding)?.to_i32()? & !0b1000 {
+        // When the third bit is 0, the rounding mode is determined by the
+        // first two bits.
+        0b000 => rustc_apfloat::Round::NearestTiesToEven,
+        0b001 => rustc_apfloat::Round::TowardNegative,
+        0b010 => rustc_apfloat::Round::TowardPositive,
+        0b011 => rustc_apfloat::Round::TowardZero,
+        // When the third bit is 1, the rounding mode is determined by the
+        // SSE status register. Since we do not support modifying it from
+        // Miri (or Rust), we assume it to be at its default mode (round-to-nearest).
+        0b100..=0b111 => rustc_apfloat::Round::NearestTiesToEven,
         rounding => throw_unsup_format!("unsupported rounding mode 0x{rounding:02x}"),
     };
 
--- a/tests/pass/intrinsics-x86-sse41.rs
+++ b/tests/pass/intrinsics-x86-sse41.rs
@@ -119,6 +119,31 @@ unsafe fn test_sse41() {
         let r = _mm_round_sd::<_MM_FROUND_TO_NEAREST_INT>(a, b);
         let e = _mm_setr_pd(-2.0, 3.5);
         assert_eq_m128d(r, e);
+
+        let a = _mm_setr_pd(1.5, 3.5);
+        let b = _mm_setr_pd(-2.5, -4.5);
+        let r = _mm_round_sd::<_MM_FROUND_TO_NEG_INF>(a, b);
+        let e = _mm_setr_pd(-3.0, 3.5);
+        assert_eq_m128d(r, e);
+
+        let a = _mm_setr_pd(1.5, 3.5);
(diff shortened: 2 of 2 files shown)


src/shims/x86/sse41.rs
```diff
@@ -283,11 +283,18 @@ fn round_first<'tcx, F: rustc_apfloat::Float>(
...
-        0x02 => rustc_apfloat::Round::TowardPositive,
-        0x03 => rustc_apfloat::Round::TowardZero,
+    // The fourth bit of `rounding` only affects the SSE status
+    // register, which cannot be accessed from Miri (or from Rust,
+    // for that matter), so we can ignore it.
+    let rounding = match this.read_scalar(rounding)?.to_i32()? & !0b1000 {
+        // When the third bit is 0, the rounding mode is determined by the
+        // first two bits. When the third bit is 1, the rounding mode is
+        // determined by the SSE status register. We do not support that case,
+        // since modifying it is unsupported in Miri (and Rust).
```
RalfJung, line 293: Is the SSE status register different from the float status register?

For the float status register, we just assume it is always in NearestTiesToEven mode... but that's the rounding mode for float-to-float operations, right? Is the integer rounding mode even controlled in a status register for regular floats?

So for the SSE status register, I assume we'd just say that it is always assumed to have a particular rounding mode. But I don't know which rounding mode e.g. LLVM will assume here.

eduardosm: Yes, it is the same register that specifies rounding mode for arithmetic operations, so I guess we can assume to be round-to-nearest (we already do that for float-to-int conversions).

It looks like LLVM does not const-fold it, even when the rounding mode does not depend on the status register.
 https://godbolt.org/z/zdh3McGcb

RalfJung: Wait, don't the normal float-to-int cast do a round-to-zero?

eduardosm: SSE has two version: one that rounds to zero (`_mm_cvttss_si32`) and one that uses the rounding mode of the status register (`_mm_cvtss_si32`). We assume round-to-nearest for the later.

https://github.com/rust-lang/miri/blob/c8d4e8356525905dc699e38d9ecf35bbd46af2a2/src/shims/x86/sse.rs#L168-L176

RalfJung: > SSE has two version: one that rounds to zero (_mm_cvttss_si32) and one that uses the rounding mode of the status register (_mm_cvtss_si32). We assume round-to-nearest for the later.

Ah okay, let's also do that here then.

But independently of that -- `f32 as i32` casts also always round to zero, so I supposed they also ignore the status register? IOW, that status register really only affects SSE operations, not scalar float operations?

src/shims/x86/sse41.rs
```diff
@@ -283,11 +283,20 @@ fn round_first<'tcx, F: rustc_apfloat::Float>(
...
+        // first two bits.
+        0b000 => rustc_apfloat::Round::NearestTiesToEven,
+        0b001 => rustc_apfloat::Round::TowardNegative,
+        0b010 => rustc_apfloat::Round::TowardPositive,
+        0b011 => rustc_apfloat::Round::TowardZero,
+        // When the third bit is 1, the rounding mode is determined by the
+        // SSE status register. Since we do not support modifying it from
+        // Miri (or Rust), we assume to be at its default mode (round-to-nearest).
+        0b100..=0b111 => rustc_apfloat::Round::NearestTiesToEven,
         rounding => throw_unsup_format!("unsupported rounding mode 0x{rounding:02x}"),
```
RalfJung, line 300:
```suggestion
        _ => unreachable!(),
```
All cases are handled now, right?

src/shims/x86/sse41.rs
eduardosm: Yes, Rust's built-in cast will use the same instruction as `_mm_cvttss_si32`.

src/shims/x86/sse41.rs
eduardosm: All valid cases are handled but there are 4 extra bits. However, the stdarch function will check that those extra bits are zero before calling the instrinsic, so I'll change that to `unreachable!()` if you prefer it.

RalfJung: Ah you're only masking off bit 4, not all the higher bits.

What does Intel say happens when those are non-zero?

eduardosm: > What does Intel say happens when those are non-zero?

Does not say much, Rust implementation but it will cause a compilation error

RalfJung: Okay, "unsupported" seems fine then.

src/shims/x86/sse41.rs
```diff
@@ -283,11 +283,20 @@ fn round_first<'tcx, F: rustc_apfloat::Float>(
...
+    let rounding = match this.read_scalar(rounding)?.to_i32()? & !0b1000 {
+        // When the third bit is 0, the rounding mode is determined by the
+        // first two bits.
+        0b000 => rustc_apfloat::Round::NearestTiesToEven,
+        0b001 => rustc_apfloat::Round::TowardNegative,
+        0b010 => rustc_apfloat::Round::TowardPositive,
+        0b011 => rustc_apfloat::Round::TowardZero,
+        // When the third bit is 1, the rounding mode is determined by the
+        // SSE status register. Since we do not support modifying it from
+        // Miri (or Rust), we assume to be at its default mode (round-to-nearest).
```
RalfJung, line 298:
```suggestion
        // Miri (or Rust), we assume it to be at its default mode (round-to-nearest).
```

RalfJung:
r=me with the last typo fixed.
@bors delegate+

PR #3134, opened 2023-10-22
avoid a linear scan over the entire int_to_ptr_map on each deallocation

RalfJung:


saethlin:
Oof. Yeah this probably matters a fair bit if it's not done by the GC.

@bors r+

Issue #2303, opened 2022-07-02
Detect non-perfectly-overlapping unordered atomic accesses as UB?


RalfJung:
See https://github.com/rust-lang/unsafe-code-guidelines/issues/345 for the details. This issue is about actually detecting the UB described there in Miri. In particular, this should be UB:
```rust
use std::sync::atomic::{AtomicU8, AtomicU16, Ordering};
use std::thread;

fn convert(a: &AtomicU16) -> &[AtomicU8; 2] {
    unsafe { std::mem::transmute(a) }
}

fn main() {
    let a = AtomicU16::new(0);
    let a16 = &a;
    let a8 = convert(a16);
    
    thread::scope(|s| {
        s.spawn(|| {
            a16.store(1, Ordering::SeqCst);
        });
        s.spawn(|| {
            a8[0].load(Ordering::SeqCst);
        });
    });
}
```

So, I wonder can we make the Miri data race detector detect this?
@cbeuw and me discussed these things in detail for https://github.com/rust-lang/miri/pull/1963, because the weak memory layer inevitably notices this. However I think we should detect this even with `-Zmiri-disable-weak-memory-emulation`, since this is more of a data race thing than a weak memory thing. And in principle, the vector clocks should have enough information for that, shouldn't they?
Cc @JCTyblaidd

RalfJung:
A concern here is that this would have to duplicate a bunch of logic from weak_memory_emulation in data_race, which would not be great. Or maybe they could somehow share that logic.

Alternatively, we would have to rephrase the documentation of `-Zmiri-disable-weak-memory-emulation` (and maybe rename the flag?) to indicate that setting this flag could miss some UB.

Or can we run just enough of weak memory emulation to detect this UB, but do no outdated reads and also mitigate most of the performance cost of weak memory emulation?

RalfJung:
We should also detect racing atomic and non-atomic accesses, even if both are reads -- C++ does not allow them so Rust should conservatively also not allow them. Here is a test:
```rust
use std::sync::atomic::{AtomicU16, Ordering};
use std::thread;

fn main() {
    let a = AtomicU16::new(0);
    
    thread::scope(|s| {
        s.spawn(|| {
            let ptr = &a as *const AtomicU16 as *mut u16;
            unsafe { ptr.read(); }
        });
        s.spawn(|| {
            a.load(Ordering::SeqCst);
        });
    });
}
```

taiki-e:
> even if both are reads

[...]

PR #3137, opened 2023-10-23
Detect mixed-size and mixed-atomicity non-synchronized accesses

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/2303


oli-obk (approved):

tests/pass/concurrency/simple.rs
```diff
@@ -62,6 +62,23 @@ fn panic_named() {
         .unwrap_err();
 }
 
+// THis is not a data race!
```
taiki-e, line 65:
```suggestion
// This is not a data race!
```

tests/pass-dep/concurrency/linux-futex.rs
```diff
@@ -267,7 +267,7 @@ fn concurrent_wait_wake() {
         FUTEX.store(FREE, Ordering::Relaxed);
         unsafe {
             DATA = 1;
-            libc::syscall(libc::SYS_futex, &FUTEX as *const AtomicI32, libc::FUTEX_WAKE, 1);
+            libc::syscall(libc::SYS_futex, addr_of!(FUTEX), libc::FUTEX_WAKE, 1);
```
taiki-e, line 270: I'm confused about the changes to this file. Does this mean that creating a reference *to an atomic type* is considered a data race?

tests/pass/concurrency/simple.rs
RalfJung: Thanks!

PR #3143, opened 2023-10-26
freebsd interceptions update proposal

devnexen:


RalfJung:
Thanks for the PR!

However, new functions without tests are generally not a good idea. Why is the `pthread_setname_np` alias needed? And for `pthread_get_name_np`, we have a test in tests/pass-dep/shims/pthreads.rs. You could split out the thread name part of the test and enable it on freebsd (by editing `ci.sh`).

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -26,17 +26,29 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
                 this.write_null(dest)?;
             }
-            "pthread_set_name_np" => {
+            // Old vs new name
+            "pthread_set_name_np" | "pthread_setname_np" => {
                 let [thread, name] =
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
-                let max_len = usize::MAX; // freebsd does not seem to have a limit.
+                let max_len = 20; // freebsd silently truncates to MAXCOMLEN (+ 1).
```
RalfJung, line 33: And MAXCOMLEN is 19 then, I assume?

tests/pass-dep/shims/pthreads.rs
```diff
@@ -147,9 +161,14 @@ fn test_named_thread_truncation() {
...
+        #[cfg(not(target_os = "freebsd"))]
         unsafe {
             libc::pthread_getname_np(libc::pthread_self(), buf.as_mut_ptr().cast(), buf.len())
         };
+        #[cfg(target_os = "freebsd")]
+        unsafe {
+            libc::pthread_get_name_np(libc::pthread_self(), buf.as_mut_ptr().cast(), buf.len())
+        };
         let cstr = CStr::from_bytes_until_nul(&buf).unwrap();
         assert!(cstr.to_bytes().len() >= 15); // POSIX seems to promise at least 15 chars
```
RalfJung, line 173: This assertion fails. Does it also fail locally when you do `./miri run --target x86_64-unknown-freebsd tests/pass-dep/shims/pthreads.rs`? If yes then you have a shot at debugging this. Having the `assert!` print the actual length on failure would probably be a good start.

```suggestion
        assert!(cstr.to_bytes().len() >= 15, "name is too short: len={}", cstr.to_bytes().len()); // POSIX seems to promise at least 15 chars
```

src/shims/unix/freebsd/foreign_items.rs
devnexen: Yes

tests/pass-dep/shims/pthreads.rs
```diff
@@ -137,6 +145,12 @@ fn test_named_thread_truncation() {
     fn set_thread_name(name: &CStr) -> i32 {
         #[cfg(target_os = "linux")]
         return unsafe { libc::pthread_setname_np(libc::pthread_self(), name.as_ptr().cast()) };
+        #[cfg(target_os = "freebsd")]
+        unsafe {
+            libc::pthread_set_name_np(libc::pthread_self(), name.as_ptr().cast())
+        };
+        #[cfg(target_os = "freebsd")]
+        return 0;
```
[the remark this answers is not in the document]
RalfJung: That's not what you implemented though. In src/shims/unix/freebsd/foreign_items.rs you're returning errors the same way Linux would.

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -26,17 +26,29 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
                 let res = this.pthread_setname_np(
                     this.read_scalar(thread)?,
                     this.read_scalar(name)?,
                     max_len,
                 )?;
                 this.write_scalar(res, dest)?;
```
RalfJung, line 39: I'm very surprised that this does not explode. Something is weird. This function returns nothing, and here we are assuming it returns an `i32`, so that should fail somewhere...

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -26,17 +26,29 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
                 this.write_null(dest)?;
             }
-            "pthread_set_name_np" => {
+            // Old vs new name
+            "pthread_set_name_np" | "pthread_setname_np" => {
```
RalfJung, line 30: According to the manpage these are not the same, they have different return types. So the implementation will also have to be different, and both variants need to be tested.

tests/pass-dep/shims/pthreads.rs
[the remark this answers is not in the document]
devnexen: Thanks for the suggestions.

src/shims/unix/freebsd/foreign_items.rs
RalfJung: Is there somewhere you can link to that documents this?

src/shims/unix/freebsd/foreign_items.rs
devnexen: In fact I ll remove the new names for now and focus on the current ones, not even libc crate have them.

src/shims/unix/freebsd/foreign_items.rs
devnexen: Sure
https://github.com/freebsd/freebsd-src/blob/314542de6d0a044640d349d68b2d1a1fe5f427ff/sys/sys/param.h#L127

here the syscall doc

https://github.com/freebsd/freebsd-src/blob/314542de6d0a044640d349d68b2d1a1fe5f427ff/lib/libc/sys/thr_set_name.2#L47

RalfJung: Okay, please add those links to the code. :)

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -26,16 +26,26 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
                 this.write_null(dest)?;
             }
+            // Old vs new name
```
RalfJung, line 29: This comment seems outdated?

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -26,16 +26,26 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
             "pthread_set_name_np" => {
                 let [thread, name] =
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
-                let max_len = usize::MAX; // freebsd does not seem to have a limit.
```
RalfJung, line 32: Why did you just remove this comment?

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -26,16 +26,26 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
                     this.read_scalar(name)?,
                     max_len,
                 )?;
-                this.write_scalar(res, dest)?;
+            }
+            // Old vs new name
+            "pthread_get_name_np" => {
+                let [thread, name, len] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                this.pthread_getname_np(
```
RalfJung, line 44:
```suggestion
                // FreeBSD pthread_get_name_np does not return anything.
                this.pthread_getname_np(
```

tests/pass-dep/shims/pthreads.rs
```diff
@@ -1,18 +1,25 @@
...
 fn main() {
+    #[cfg(not(target_os = "freebsd"))]
     test_mutex_libc_init_recursive();
+    #[cfg(not(target_os = "freebsd"))]
     test_mutex_libc_init_normal();
+    #[cfg(not(target_os = "freebsd"))]
     test_mutex_libc_init_errorcheck();
+    #[cfg(not(target_os = "freebsd"))]
     test_rwlock_libc_static_initializer();
     test_named_thread_truncation();
```
RalfJung, line 16:
```suggestion

    test_named_thread_truncation();
```

tests/pass-dep/shims/pthreads.rs
```diff
@@ -137,6 +147,12 @@ fn test_named_thread_truncation() {
     fn set_thread_name(name: &CStr) -> i32 {
         #[cfg(target_os = "linux")]
         return unsafe { libc::pthread_setname_np(libc::pthread_self(), name.as_ptr().cast()) };
+        #[cfg(target_os = "freebsd")]
+        unsafe {
+            libc::pthread_set_name_np(libc::pthread_self(), name.as_ptr().cast())
+        };
+        #[cfg(target_os = "freebsd")]
+        return 0;
```
RalfJung, line 155:
```suggestion
        unsafe {
            libc::pthread_set_name_np(libc::pthread_self(), name.as_ptr().cast());
            // pthread_set_name_np does not return anything, just assume it went well.
            return 0;
        };
```

tests/pass-dep/shims/pthreads.rs
```diff
@@ -147,16 +163,22 @@ fn test_named_thread_truncation() {
...
+            libc::pthread_get_name_np(libc::pthread_self(), buf.as_mut_ptr().cast(), buf.len())
+        };
         let cstr = CStr::from_bytes_until_nul(&buf).unwrap();
-        assert!(cstr.to_bytes().len() >= 15); // POSIX seems to promise at least 15 chars
+        assert!(cstr.to_bytes().len() >= 15, "name is too short: len={}", cstr.to_bytes().len()); // POSIX seems to promise at least 15 chars
         assert!(long_name.as_bytes().starts_with(cstr.to_bytes()));
 
         // Also test directly calling pthread_setname to check its return value.
         assert_eq!(set_thread_name(&cstr), 0);
         // But with a too long name it should fail.
```
RalfJung, line 180:
```suggestion
        // But with a too long name it should fail (except on FreeBSD where the function
        // has no return value and hence cannot indicate failure).
```

RalfJung (commented):
Sorry for the silence, I didn't realize this was ready for another round of review. Please don't just push and assume we'll get notified; github does not always send notifications for force pushes and it's never clear whether such a push is meant to resolve all prior discussions or whether more pushes will come.

@rustbot author
(write `@rustbot review` when this is ready for review again)

RalfJung:
Looking good, thanks! I just did some tweaks.

@bors r+

PR #3144, opened 2023-10-27
add aarch64-apple-darwin to list of supported targets

RalfJung:
I did not realize that aarch64-apple-darwin is a tier 2 target. We are testing it properly, even on rustc CI, so we might as well list it as fully supported here.


README.md
```diff
@@ -223,7 +223,8 @@ degree documented below):
 - All Rust [Tier 1 targets](https://doc.rust-lang.org/rustc/platform-support.html) are supported by
   Miri. They are all checked on Miri's CI, and some (at least one per OS) are even checked on every
   Rust PR, so the shipped Miri should always work on these targets.
-- We also support `s390x-unknown-linux-gnu` as our "big-endian target of choice".
```
oli-obk, line 226: Maybe this should be a list (with the endian note after the s390x item)?

Issue #3142, opened 2023-10-25
Show targeted message on read-read races


RalfJung:
We now consider atomic reads racing with non-atomic reads to be UB. We should probably print a targeted message for that case since it's rather surprising that this would be disallowed.

PR #3145, opened 2023-10-27
give some more help for the unusual data races

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/3142

### DIFF
--- a/src/concurrency/data_race.rs
+++ b/src/concurrency/data_race.rs
@@ -41,7 +41,6 @@
 //! on the data-race detection code.
 
 use std::{
-    borrow::Cow,
     cell::{Cell, Ref, RefCell, RefMut},
     fmt::Debug,
     mem,
@@ -199,7 +198,7 @@ struct AtomicMemoryCellClocks {
 /// are all treated as writes for the purpose
 /// of the data-race detector.
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
-enum WriteType {
+enum NaWriteType {
     /// Allocate memory.
     Allocate,
 
@@ -212,12 +211,41 @@ enum WriteType {
     /// (Same for `Allocate` above.)
     Deallocate,
 }
-impl WriteType {
-    fn get_descriptor(self) -> &'static str {
+
+impl NaWriteType {
+    fn description(self) -> &'static str {
         match self {
-            WriteType::Allocate => "Allocate",
-            WriteType::Write => "Write",
-            WriteType::Deallocate => "Deallocate",
+            NaWriteType::Allocate => "creating a new allocation",
+            NaWriteType::Write => "non-atomic write",
+            NaWriteType::Deallocate => "deallocation",
+        }
+    }
+}
+
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+enum AccessType {
+    NaRead,
+    NaWrite(NaWriteType),
+    AtomicLoad,
+    AtomicStore,
+    AtomicRmw,
+}
+
+impl AccessType {
+    fn description(self) -> &'static str {
+        match self {
+            AccessType::NaRead => "non-atomic read",
+            AccessType::NaWrite(w) => w.description(),
+            AccessType::AtomicLoad => "atomic load",
+            AccessType::AtomicStore => "atomic store",
+            AccessType::AtomicRmw => "atomic read-modify-write",
+        }
+    }
+
+    fn is_atomic(self) -> bool {
+        match self {
+            AccessType::AtomicLoad | AccessType::AtomicStore | AccessType::AtomicRmw => true,
+            AccessType::NaRead | AccessType::NaWrite(_) => false,
         }
     }
 }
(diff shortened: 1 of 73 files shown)


src/concurrency/data_race.rs
```diff
@@ -892,18 +924,39 @@ impl VClockAlloc {
...
+        let involves_non_atomic = !access.is_atomic() || !other_access.is_atomic();
 
         // Throw the data-race detection.
+        let extra = if other_size.is_some() {
+            assert!(!involves_non_atomic);
+            Some("overlapping unsynchronized atomic accesses must use the same access size")
+        } else if access.is_read() && other_access.is_read() {
+            assert!(involves_non_atomic);
+            Some(
+                "overlapping atomic and non-atomic accesses must be synchronized, even if both are read-only",
```
RalfJung, line 936: @saethlin I have been deliberating how to best express this... is "must be synchronized" clear enough? Would "must be ordered" better? Any other ideas?

saethlin: Just to be sure, is this an example of the UB you are talking about?
```
error: Undefined Behavior: Data race detected between (1) Atomic Load on thread `tokio-runtime-w` and (2) Read on thread `tokio-runtime-w` at alloc31996+0x20. (2) just happened here
   --> /home/ben/tokio/tokio/src/loom/std/atomic_u32.rs:26:9
    |
26  |         core::ptr::read(self.inner.get() as *const u32)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Data race detected between (1) Atomic Load on thread `tokio-runtime-w` and (2) Read on thread `tokio-runtime-w` at alloc31996+0x20. (2) just happened here
    |
   --> /home/ben/tokio/tokio/src/runtime/scheduler/multi_thread/queue.rs:455:28
    |
455 |             let src_tail = self.0.tail.load(Acquire);
    |                            ^^^^^^^^^^^^^^^^^^^^^^^^^
```

RalfJung: Yes.

saethlin: I think I prefer "ordered"? At least the way I'd hand-wave the difference between those terms in English is to say that atomic operations provide orderings, then synchronization tools like locks can be built on those. Synchronization sounds like a stronger term to me :shrug:

But my biggest question/concern with the diagnostic here is that this being UB will probably be surprising, so I think it would be best if we could preempt people asking questions about why. Is there a short version explanation? Or maybe we could link to https://github.com/rust-lang/unsafe-code-guidelines/issues/345? The Twitter link in the issue is already broken, so the story is already a little confusing. My best reading of that issue is that this is UB because we mostly punt to C++ for our atomic semantics, and C++ seems to make this UB, though it's not clear why, and x86 really wants mixed-size atomics to be UB.

RalfJung: > Is there a short version explanation? 

"There's no way to write code that does this in C++ without causing UB, and for now we don't want to invent our own memory model."

I am not sure if C++ has a very good reason for making this UB... x86 says it's disallowed but also Linux and Wine do it so it's not like they will be able to break this.

RalfJung: Hm on second though... AFAIK describing mixed-size accesses in these C++-style models is just very hard / hasn't been figured out yet. Suddenly a single read can read-from multiple writes and you have to consider the total coherence order per-location rather than "per atomic object", and that doesn't seem to "just work".

For races between atomic and non-atomic reads, I don't think there is any good reason to disallow them, IMO that's an oversight in C++. Formal models of the C++ memory model generally allow them.

saethlin: > AFAIK describing mixed-size accesses in these C++-style models is just very hard / hasn't been figured out yet

That's definitely how it seems to me. It would be unfortunate for us to say that atomic and non-atomic reads can produce a data race, especially if it's known to be permissible in formal models.

We've been not reporting this scenario as UB for a long time, so it would be kind of silly to start defensively reporting UB when we're sure that this will be made well-defined in the future.

RalfJung: > We've been not reporting this scenario as UB for a long time,

We've not supported threads for a long time, and when we did we've not reported any data races for a long time. The fact that we allowed racing atomic and non-atomic accesses was an oversight; I didn't realize that this is not possible in C++.

We have documented for a looooong time that we are using the C++ memory model. Until recently we didn't have any APIs that would anyone write racing atomic and non-atomic accesses (without brute-force operations like `transmute`). https://github.com/rust-lang/rust/pull/115719 stabilized `AtomicX::from_ptr` which made this possible for the first time, and as part of stabilization we also clarified the docs to say explicitly that mixing atomic and non-atomic accesses is not allowed. Miri should detect such bugs.

saethlin: Are you suggesting we link to the docs? People tend to behave as if the docs are stable, they read them perhaps a few times when they are learning APIs then just use them. Unless we do something to prompt them, I do not think it is reasonable to expect an experienced user to notice that the docs have changed.

I'm not opposing reporting UB here, I just expect this one to be very odd to explain to people. And the more we can head that off when they see the UB report, the better.

RalfJung: Sure, I can add links to the docs.

RalfJung: But that is blocked on https://github.com/rust-lang/rust/pull/116762, which puts the relevant docs in a more central place.

RalfJung:
I'll add a link to the docs once https://github.com/rust-lang/rust/pull/116762 lands. Meanwhile I think this is still an improvement, we can always further improve the message later.

@bors r+

PR #3161, opened 2023-11-12
freebsd adding getentropy interception support

devnexen:


tests/pass-dep/shims/libc-getentropy.rs
```diff
@@ -0,0 +1,11 @@
+//@only-target-freebsd
```
RalfJung, line 1: Why should this test run only on freebsd? The function seems to exist on other Unixes as well.

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -47,6 +47,21 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                     this.read_scalar(len)?,
                 )?;
             }
+            "getentropy" => {
+                let [buf, bufsize] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let buf = this.read_pointer(buf)?;
+                let bufsize = this.read_target_usize(bufsize)?;
+
+                if bufsize > 256 {
```
RalfJung, line 56: Where does the magic number 256 come from? Please point to documentation.

RalfJung: Interesting, even the Linux and macOS version seem to have the 256 limit (judging from my local man page and macOS man pages I found online). We forgot to check this in the macos version...

tests/pass-dep/shims/libc-getentropy.rs
devnexen: because I was focusing on freebsd only. we might be able to expand it in a later PR.

src/shims/unix/freebsd/foreign_items.rs
devnexen: > Where does the magic number 256 come from? Please point to documentation.

sure will do.

RalfJung:
Looks good, thanks!
@bors r+

tests/pass-dep/shims/libc-getentropy.rs
RalfJung: I've done this in https://github.com/rust-lang/miri/pull/3162

PR #3164, opened 2023-11-13
freebsd adding getrandom interception.

devnexen:
note that os support was added in same time as getentropy.

### DIFF
--- a/src/shims/unix/freebsd/foreign_items.rs
+++ b/src/shims/unix/freebsd/foreign_items.rs
@@ -47,6 +47,21 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                     this.read_scalar(len)?,
                 )?;
             }
+            "getrandom" => {
+                let [ptr, len, flags] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let ptr = this.read_pointer(ptr)?;
+                let len = this.read_target_usize(len)?;
+                let _flags = this.read_scalar(flags)?.to_i32()?;
+                // flags on freebsd does not really matter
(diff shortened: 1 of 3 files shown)


src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -47,6 +47,20 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                     this.read_scalar(len)?,
                 )?;
             }
+            "getrandom" => {
+                let [ptr, len, _flags] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let ptr = this.read_pointer(ptr)?;
+                let len = this.read_target_usize(len)?;
+                // flags on freebsd does not really matter
```
RalfJung, line 55: Even when flags are ignored, they must still be `read_` at the appropriate type to ensure some initialized argument of the right size was provided.

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -47,6 +47,21 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                     this.read_scalar(len)?,
                 )?;
             }
+            "getrandom" => {
+                let [ptr, len, _flags] =
```
RalfJung, line 51: This variable is used now, please rename it to `flags`.

devnexen: ok

devnexen: wait ... now the build complains if I rename it.

RalfJung:
Looking good, thanks!
@bors r+

PR #3166, opened 2023-11-15
reallocarray shim linux/freebsd support proposal.

devnexen:


RalfJung:
Thanks!

However, please share the code between the linux and freebsd implementations. For instance you could have this in the shared unix/foreign_items module with a guard throwing 'unsupported' when the OS is not Linux or FreeBSD.

src/shims/unix/foreign_items.rs
```diff
@@ -252,6 +252,38 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 this.write_scalar(result, dest)?;
             }
 
+            "reallocarray" => {
+                // TODO: Since macOs tends to grab FreeBSD's userland pieces,
+                // it might get support at some point.
```
RalfJung, line 257:
```suggestion
                // Currently this function does not exist on all Unixes, e.g. on macOS.
```
It's not really our concern what Apple does with macOS. We'll add it if that happens, but until then it's not a TODO on our side.

RalfJung:
> error: pass test got signal: 6 (SIGABRT), but expected 0

No idea why this is happening, but let's hope it does not happen again.^^
@bors r+

PR #3170, opened 2023-11-16
a bit of libc test reorganization

RalfJung:


RalfJung:
Hm, seems it did not like that.
@bors r+

PR #3176, opened 2023-11-18
Implement all 16 AVX compare operators for 128-bit SIMD vectors

eduardosm:
`_mm_cmp_{ss,ps,sd,pd}` functions are AVX functions that use `llvm.x86.sse{,2}.` prefixed intrinsics, so they were "accidentally" partially implemented when SSE and SSE2 intrinsics were implemented.

The 16 AVX compare operators are now implemented and tested.

### DIFF
--- a/src/shims/x86/mod.rs
+++ b/src/shims/x86/mod.rs
@@ -119,53 +119,32 @@ pub(super) trait EvalContextExt<'mir, 'tcx: 'mir>:
     }
 }
 
-/// Floating point comparison operation
-///
-/// <https://www.felixcloutier.com/x86/cmpss>
-/// <https://www.felixcloutier.com/x86/cmpps>
-/// <https://www.felixcloutier.com/x86/cmpsd>
-/// <https://www.felixcloutier.com/x86/cmppd>
-#[derive(Copy, Clone)]
-enum FloatCmpOp {
-    Eq,
-    Lt,
-    Le,
-    Unord,
-    Neq,
-    /// Not less-than
-    Nlt,
-    /// Not less-or-equal
-    Nle,
-    /// Ordered, i.e. neither of them is NaN
-    Ord,
-}
-
-impl FloatCmpOp {
-    /// Convert from the `imm` argument used to specify the comparison
-    /// operation in intrinsics such as `llvm.x86.sse.cmp.ss`.
-    fn from_intrinsic_imm(imm: i8, intrinsic: &str) -> InterpResult<'_, Self> {
-        match imm {
-            0 => Ok(Self::Eq),
-            1 => Ok(Self::Lt),
-            2 => Ok(Self::Le),
-            3 => Ok(Self::Unord),
-            4 => Ok(Self::Neq),
-            5 => Ok(Self::Nlt),
-            6 => Ok(Self::Nle),
-            7 => Ok(Self::Ord),
-            imm => {
-                throw_unsup_format!("invalid `imm` parameter of {intrinsic}: {imm}");
-            }
-        }
-    }
-}
-
 #[derive(Copy, Clone)]
 enum FloatBinOp {
     /// Arithmetic operation
     Arith(mir::BinOp),
     /// Comparison
-    Cmp(FloatCmpOp),
+    ///
+    /// The semantics of this operator is a case distinction: we compare the two operands,
+    /// and then we return one of the four booleans `gt`, `lt`, `eq`, `unord` depending on
+    /// which class they fall into.
+    ///
+    /// AVX supports all 16 combinations, SSE only a subset
+    ///
+    /// <https://www.felixcloutier.com/x86/cmpss>
+    /// <https://www.felixcloutier.com/x86/cmpps>
+    /// <https://www.felixcloutier.com/x86/cmpsd>
+    /// <https://www.felixcloutier.com/x86/cmppd>
+    Cmp {
+        /// Result when lhs < rhs
+        gt: bool,
+        /// Result when lhs > rhs
+        lt: bool,
+        /// Result when lhs == rhs
+        eq: bool,
+        /// Result when lhs is NaN or rhs is NaN
+        unord: bool,
+    },
     /// Minimum value (with SSE semantics)
     ///
     /// <https://www.felixcloutier.com/x86/minss>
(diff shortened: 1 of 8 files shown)


src/shims/x86/mod.rs
```diff
@@ -119,53 +119,27 @@ pub(super) trait EvalContextExt<'mir, 'tcx: 'mir>:
...
+    Cmp {
+        /// Result when lhs < rhs
+        gt: bool,
+        /// Result when lhs > rhs
+        lt: bool,
+        /// Result when lhs == rhs
+        eq: bool,
+        /// Result when lhs is NaN or rhs is NaN
+        unord: bool,
+    },
```
RalfJung, line 142: What does this mean though when I have e.g. `Cmp { gt: true, lt: true, eq: false, unord: false }`? This gives me 2 booleans, how are they combined? Probably either "and" or "or"?

src/shims/x86/mod.rs
```diff
@@ -182,6 +156,53 @@ enum FloatBinOp {
...
 
+impl FloatBinOp {
+    /// Convert from the `imm` argument used to specify the comparison
+    /// operation in intrinsics such as `llvm.x86.sse.cmp.ss`.
+    fn cmp_from_imm(imm: i8, intrinsic: &str) -> InterpResult<'_, Self> {
+        // Bit 4 specifies whether the operation is quiet or signaling, which
+        // we do not care in Miri.
+        match imm & !0x10 {
+            // EQ_O
+            0x0 => Ok(Self::Cmp { gt: false, lt: false, eq: true, unord: false }),
```
RalfJung, line 167: This sounds like it could be written with less redundacny by having a little helper `get_bit` and then using `lt: get_bit(0)` and so on?

src/shims/x86/mod.rs
```diff
@@ -195,20 +216,15 @@ fn bin_op_float<'tcx, F: rustc_apfloat::Float>(
...
-                FloatCmpOp::Neq => left != right,
-                FloatCmpOp::Nlt => !(left < right),
-                FloatCmpOp::Nle => !(left <= right),
-                FloatCmpOp::Ord => !left.is_nan() && !right.is_nan(),
+
+            let res = match left.partial_cmp(&right) {
+                None => unord,
+                Some(std::cmp::Ordering::Less) => lt,
+                Some(std::cmp::Ordering::Equal) => eq,
+                Some(std::cmp::Ordering::Greater) => gt,
```
RalfJung, line 227: Oh, *those* are the semantics... yeah that must be documented properly in `FloatBinOp`, this is not obvious at all.

src/shims/x86/mod.rs
```diff
@@ -119,53 +119,27 @@ pub(super) trait EvalContextExt<'mir, 'tcx: 'mir>:
...
-            }
-        }
-    }
-}
-
 #[derive(Copy, Clone)]
 enum FloatBinOp {
     /// Arithmetic operation
     Arith(mir::BinOp),
     /// Comparison
```
RalfJung, line 126:
```suggestion
    /// Comparison
    ///
    /// The semantics of this operator is a case distinction: we compare the two operands,
    /// and then we return one of the four booleans `gt`, `lt`, `eq`, `unord` depending on
    /// which class they fall into.
    ///
```

src/shims/x86/sse2.rs
```diff
@@ -461,33 +461,41 @@ pub(super) trait EvalContextExt<'mir, 'tcx: 'mir>:
...
                 bin_op_simd_float_first::<Double>(this, which, left, right, dest)?;
             }
             // Used to implement the _mm_cmp*_pd functions.
             // Performs a comparison operation on each component of `left`
             // and `right`. For each component, returns 0 if false or `u64::MAX`
             // if true.
+            // _mm_cmp_pd is actually an AVX function where the operation is specified
+            // by a const parameter.
+            // _mm_cmp{eq,lt,le,gt,ge,neq,nlt,nle,ngt,nge,ord,unord}_pd are SSE2 functions
+            // with hard-coded operations.
```
RalfJung, line 490: So should we support the AVX part of this only sometimes, when AVX is actually available, or something like that?

tests/pass/intrinsics-x86-avx.rs
```diff
@@ -0,0 +1,162 @@
+// Ignore everything except x86 and x86_64
+// Any additional target are added to CI should be ignored here
```
RalfJung, line 2:
```suggestion
// Any new targets that are added to CI should be ignored here.
```

src/shims/x86/sse2.rs
eduardosm: Currently, we are not checking whether the target feature required by a LLVM intrinsic is enabled or not (at least not for x86). We assume that it is called from a stdarch function with the corresponding `#[target_feature(enable = "...")]`.

RalfJung: Oh, interesting. Can you open an issue for that? We should probably trigger UB if the target feature is missing...

tests/pass/intrinsics-x86-avx.rs
RalfJung: If you could fix this in all test files while you are at it, that would be great. :)

src/shims/x86/mod.rs
eduardosm: I reduced the table to half. It can be further reduced with some boolean logic, but I think it would make it more confusing.

src/shims/x86/sse2.rs
eduardosm: Done: https://github.com/rust-lang/miri/issues/3178

I will work on that after this PR is landed.

src/shims/x86/mod.rs
```diff
@@ -182,6 +161,42 @@ enum FloatBinOp {
...
+    /// operation in intrinsics such as `llvm.x86.sse.cmp.ss`.
+    fn cmp_from_imm(imm: i8, intrinsic: &str) -> InterpResult<'_, Self> {
+        // Only bits 0..=4 are used, remaining should be zero.
+        if imm & !0b1_1111 != 0 {
+            throw_unsup_format!("invalid `imm` parameter of {intrinsic}: 0x{imm:x}");
+        }
+        // Bit 4 specifies whether the operation is quiet or signaling, which
+        // we do not care in Miri.
+        // Bits 0..=2 specifies the operation.
+        let (gt, lt, eq, unord) = match imm & 0b111 {
```
RalfJung, line 175:
```suggestion
        // `gt` indicates the result to be returned when the LHS is strictly greater than the RHS, and so on.
        let (gt, lt, eq, unord) = match imm & 0b111 {
```

src/shims/x86/mod.rs
RalfJung: Oh, that's not nearly as nice as I had hoped... oh well, I guess I was expecting too much from Intel.^^

RalfJung:
Awesome, thanks!
@bors r+

PR #3177, opened 2023-11-18
libc-misc test freebsd fixes attempt

devnexen:


tests/pass-dep/shims/libc-misc.rs
```diff
@@ -417,7 +416,7 @@ fn main() {
 
     test_isatty();
 
-    #[cfg(not(target_os = "freebsd"))] // FIXME we should support this on FreeBSD as well
+    //#[cfg(not(target_os = "freebsd"))] // FIXME we should support this on FreeBSD as well
```
RalfJung, line 419: Please don't leave commented-out `cfg` behind

tests/pass-dep/shims/libc-misc.rs
```diff
@@ -406,7 +405,7 @@ fn test_reallocarray() {
 fn main() {
     test_posix_gettimeofday();
 
-    #[cfg(not(target_os = "freebsd"))] // FIXME we should support this on FreeBSD as well
+    #[cfg(not(all(target_os = "freebsd", pointer_width = "32")))]
```
RalfJung, line 408: We don't run these tests on 32bit freebsd so this is unnecessary

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -62,6 +63,15 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 this.gen_random(ptr, len)?;
                 this.write_scalar(Scalar::from_target_usize(len, this), dest)?;
             }
+            "ftruncate" => {
+                if this.tcx.sess.target.pointer_width == 32 {
```
RalfJung, line 67: Does this function really only exist on 64bit FreeBSD? That's very strange...

RalfJung: Oh I guess the problem is that the second argument is actually `off_t`, which I assume is `isize`, not `i64`?

having functions only exist on some bitwidths is too surprising IMO. Please change `ftruncate64` to take `fd: i32, length: i64` rather than taking `OpTy`. Then Linux and FreeBSD can share the implementation.

src/shims/unix/foreign_items.rs
```diff
@@ -163,9 +163,23 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
             }
+            "ftruncate" => {
+                let [fd, length] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let fd = this.read_scalar(fd)?.to_i32()?;
+                let length_type: i64 = if this.tcx.sess.target.pointer_width == 64 {
+                    this.read_scalar(length)?.to_i64()?
+                } else {
+                    i64::from(this.read_scalar(length)?.to_i32()?)
+                };
```
RalfJung, line 179: Please use `read_target_isize` instead. No need to test the pointer_width.

src/shims/unix/foreign_items.rs
```diff
@@ -163,9 +163,23 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
             "ftruncate64" => {
                 let [fd, length] =
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let fd = this.read_scalar(fd)?.to_i32()?;
+                let length = this.read_scalar(length)?.to_i64()?;
                 let result = this.ftruncate64(fd, length)?;
                 this.write_scalar(result, dest)?;
             }
+            "ftruncate" => {
```
RalfJung, line 171: If you're adding this for all targets now we should probably have a test for it... `libc-fs.rs` sound like the right file.

We've historically been pretty bad with libc tests for our FS APIs though. And `test_posix_mkstemp` should probably be in libc-fs.rs as well. So if you prefer we can open an issue to track fixing this in the future.

devnexen:
@rustbost review

src/shims/unix/foreign_items.rs
RalfJung: There's still no test for this function, is there?

tests/pass-dep/shims/libc-fs.rs
```diff
@@ -93,6 +97,7 @@ fn test_canonicalize_too_long() {
     assert!(canonicalize(too_long).is_err());
 }
 
+#[cfg(not(target_os = "freebsd"))]
```
RalfJung, line 100: This test seems to work fine on FreeBSD?

tests/pass-dep/shims/libc-fs-with-isolation.rs
```diff
@@ -1,4 +1,5 @@
 //@ignore-target-windows: no libc on Windows
+//@ignore-target-freebsd
```
RalfJung, line 2:
```suggestion
//@ignore-target-freebsd: FIXME needs foreign function `stat@FBSD_1.0`
```

src/shims/unix/foreign_items.rs
devnexen: I did not add purposely, can be done separately.

tests/pass-dep/shims/libc-fs-with-isolation.rs
devnexen: Note could be fixed in "all fs" next PR eventually.

src/shims/unix/foreign_items.rs
RalfJung: I said it should either be done here or have an issue to track it. :)
Here's an issue: https://github.com/rust-lang/miri/issues/3179

RalfJung:
Looks good, thanks!
@bors r+

Issue #3178, opened 2023-11-19
Check for required target features when evaluating LLVM intrinsics


eduardosm:
Many LLVM intrinsics require some target features (e.g., `llvm.x86.sse.*` intrinsics generally require the x86 SSE feature)-

We are not currently checking whether they are enabled or not, since intrinsics are not defined with a `#[target_feature(enable = "...")]` attribute.

Noticed in https://github.com/rust-lang/miri/pull/3176#discussion_r1398421309.

PR #3180, opened 2023-11-20
Check that target features required by LLVM intrinsics are enabled

eduardosm:
Fixes https://github.com/rust-lang/miri/issues/3178

### DIFF
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -1063,6 +1063,24 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
             _ => span_bug!(this.cur_span(), "unexpected type: {ty:?}"),
         }
     }
+
+    /// Checks that target feature `target_feature` is enabled.
+    ///
+    /// If not enabled, emits an UB error that states that the feature is
+    /// required by `intrinsic`.
+    fn expect_target_feature_for_intrinsic(
+        &self,
+        intrinsic: Symbol,
+        target_feature: &str,
+    ) -> InterpResult<'tcx, ()> {
+        let this = self.eval_context_ref();
+        if !this.tcx.sess.unstable_target_features.contains(&Symbol::intern(target_feature)) {
+            throw_ub_format!(
+                "attempted to call intrinsic `{intrinsic}` that requires missing target feature {target_feature}"
+            );
+        }
+        Ok(())
+    }
 }
 
 impl<'mir, 'tcx> MiriMachine<'mir, 'tcx> {
--- a/src/shims/foreign_items.rs
+++ b/src/shims/foreign_items.rs
@@ -1008,9 +1008,11 @@ trait EvalContextExtPriv<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
             "llvm.arm.hint" if this.tcx.sess.target.arch == "arm" => {
                 let [arg] = this.check_shim(abi, Abi::Unadjusted, link_name, args)?;
                 let arg = this.read_scalar(arg)?.to_i32()?;
+                // Note that different arguments might have different target feature requirements.
                 match arg {
                     // YIELD
                     1 => {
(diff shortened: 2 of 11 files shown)


eduardosm:
I'm not sure what's going on. `llvm.arm.hint` is called through `std::arch::arm::__yield`, which requires `target_feature = "v6"`.

```rust
#[cfg(any(target_feature = "v6", target_arch = "aarch64", doc))]
#[inline(always)]
pub unsafe fn __yield() {
    hint(HINT_YIELD);
}

...

extern "unadjusted" {
    #[cfg_attr(target_arch = "aarch64", link_name = "llvm.aarch64.hint")]
    #[cfg_attr(target_arch = "arm", link_name = "llvm.arm.hint")]
    fn hint(_: i32);
}
```

src/helpers.rs
```diff
@@ -1063,6 +1063,24 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+    /// required by `intrinsic`.
+    fn expect_target_feature_for_intrinsic(
+        &self,
+        intrinsic: Symbol,
+        target_feature: Symbol,
+    ) -> InterpResult<'tcx, ()> {
+        let this = self.eval_context_ref();
+        if !this.tcx.sess.unstable_target_features.contains(&target_feature) {
+            throw_ub_format!(
+                "attempted to call intrinsic `{intrinsic}` that requires missing {target_feature} target feature"
```
RalfJung, line 1079:
```suggestion
                "attempted to call intrinsic `{intrinsic}` that requires missing target feature `{target_feature}`"
```

src/shims/foreign_items.rs
```diff
@@ -1008,9 +1008,11 @@ trait EvalContextExtPriv<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
             "llvm.arm.hint" if this.tcx.sess.target.arch == "arm" => {
                 let [arg] = this.check_shim(abi, Abi::Unadjusted, link_name, args)?;
                 let arg = this.read_scalar(arg)?.to_i32()?;
+                // Note that different arguments might have different target feature requirements.
                 match arg {
                     // YIELD
                     1 => {
+                        this.expect_target_feature_for_intrinsic(link_name, Symbol::intern("v6"))?;
```
RalfJung, line 1015: Looks like `expect_target_feature_for_intrinsic` should take `&str` as its second argument?

RalfJung (commented):
Implementation looks good, I just hope we got them all. :)

Could you add a test to ensure the UB indeed does get reported properly?

src/shims/foreign_items.rs
eduardosm: Yeah, since we only have `sse` and `avx` in `sym` anyways

eduardosm:
Oh right, SSE4.1 is enabled by default on macOS

tests/fail/shims/intrinsic_target_feature.rs
```diff
@@ -0,0 +1,42 @@
...
+#[cfg(target_arch = "x86_64")]
+use std::arch::x86_64::*;
+
+fn main() {
+    assert!(is_x86_feature_detected!("sse"));
+    assert!(!is_x86_feature_detected!("sse4.1"));
+
+    unsafe {
+        // Pass, since SSE is enabled
+        let _ = addss(_mm_setzero_ps(), _mm_setzero_ps());
```
RalfJung, line 27: Are these must_use, or why the `let _`?

eduardosm: Simply to make it explicit that we don't care about the returned value

RalfJung: I find this more confusing than helpful, TBH. I'd prefer if you removed it.

eduardosm: So be it

RalfJung:
Thanks for the PR!
@bors r+

PR #3181, opened 2023-11-20
freebsd add *stat calls interception support

devnexen:


src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -63,6 +64,25 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 this.write_scalar(Scalar::from_target_usize(len, this), dest)?;
             }
 
+            // File related shims
+            "stat" | "stat@FBSD_1.0" => {
```
RalfJung, line 68: So both of these exist? I would have expected only the `stat@FBSD_1.0` symbol.

src/shims/unix/fs.rs
```diff
@@ -911,13 +911,16 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+    fn macos_fbsd_stat(
         &mut self,
         path_op: &OpTy<'tcx, Provenance>,
         buf_op: &OpTy<'tcx, Provenance>,
     ) -> InterpResult<'tcx, Scalar<Provenance>> {
         let this = self.eval_context_mut();
-        this.assert_target_os("macos", "stat");
+
+        if !matches!(&*this.tcx.sess.target.os, "macos" | "freebsd") {
+            throw_unsup_format!("macos_fbsd_stat not meant for os {}", this.tcx.sess.target.os);
```
RalfJung, line 922: This shouldn't throw a nice error, it should panic. There's a bug in Miri when this happens.

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -63,6 +64,25 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                let [path, buf] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let result = this.macos_fbsd_stat(path, buf)?;
+                this.write_scalar(result, dest)?;
+            }
+            "lstat" | "lstat@FBSD_1.0" => {
```
RalfJung, line 74: Is there any test even calling this function?

Same question for `fstat`.

Usually these are tested by tests/pass/shims/fs.rs but I think that test is not run on FreeBSD yet.

Cargo.lock
```diff
@@ -30,18 +30,18 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.1"
+version = "1.1.2"
```
RalfJung, line 33: Please revert the lockfile changes, no reason to do that in the same PR as the rest of this.

src/shims/unix/foreign_items.rs
```diff
@@ -157,6 +157,17 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                let offset = this.read_scalar(offset)?.to_i64()?;
+                let whence = this.read_scalar(whence)?.to_i32()?;
+                let result = this.lseek64(fd, offset, whence)?;
+                this.write_scalar(result, dest)?;
+            }
+            "lseek" => {
+                let [fd, offset, whence] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let fd = this.read_scalar(fd)?.to_i32()?;
+                let offset = this.read_scalar(offset)?.to_i64()?;
+                let whence = this.read_scalar(whence)?.to_i32()?;
```
RalfJung, line 170: This has the *exact* same code as `lseek64`. That does not seem right? `offset` should probably use `read_target_isize`.

src/shims/unix/freebsd/foreign_items.rs
RalfJung: You have not responded to or resolved this comment.

src/shims/unix/fs.rs
```diff
@@ -940,13 +939,16 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+    fn macos_fbsd_lstat(
         &mut self,
         path_op: &OpTy<'tcx, Provenance>,
         buf_op: &OpTy<'tcx, Provenance>,
     ) -> InterpResult<'tcx, Scalar<Provenance>> {
         let this = self.eval_context_mut();
-        this.assert_target_os("macos", "lstat");
+
+        if !matches!(&*this.tcx.sess.target.os, "macos" | "freebsd") {
+            panic!("macos_fbsd_lstat not meant for os {}", this.tcx.sess.target.os);
```
RalfJung, line 950: There's a verb missing in this sentence.
```suggestion
            panic!("`macos_fbsd_lstat` should not be called on {}", this.tcx.sess.target.os);
```
(Do this everywhere you have the new panic.)

src/shims/unix/fs.rs
```diff
@@ -1385,15 +1389,20 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
     ) -> InterpResult<'tcx, Scalar<Provenance>> {
         let this = self.eval_context_mut();
 
-        this.assert_target_os("macos", "readdir_r");
+        if !matches!(&*this.tcx.sess.target.os, "macos" | "freebsd") {
+            panic!("macos_fbsd_readdir_r not meant for os {}", this.tcx.sess.target.os);
+        }
+
+        // FIXME?: oddily it seems to pick up the old libc crate's freebsd11 definition instead of freebsd13, should be 6
+        let name_offset = if this.tcx.sess.target.os == "macos" { 5 } else { 4 };
```
RalfJung, line 1405: Not sure why you mention libc here, this is a hard-coded offset, how should that pick up anything...?

Also, did that offset change between different versions of freebsd?!? How are we supposed to handle that, not knowing which version the program is built for?

RalfJung: Oh wait, the name is just confusing. This is not an offset! It is a field index.

This should pick up the version of libc that comes with the standard library.

src/shims/unix/fs.rs
```diff
@@ -1424,7 +1433,7 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 // }
 
                 let entry_place = this.deref_pointer_as(entry_op, this.libc_ty_layout("dirent"))?;
-                let name_place = this.project_field(&entry_place, 5)?;
+                let name_place = this.project_field(&entry_place, name_offset)?;
```
RalfJung, line 1436: Please use project_field_named, that should avoid this index hard-coding.

tests/pass/shims/fs-with-isolation.rs
```diff
@@ -1,4 +1,5 @@
 //@ignore-target-windows: File handling is not implemented yet
+//@ignore-target-freebsd
```
RalfJung, line 2: what goes wrong when this runs on freebsd? please add a FIXME

tests/pass/shims/fs.rs
```diff
@@ -4,12 +4,15 @@
 #![feature(io_error_more)]
 #![feature(io_error_uncategorized)]
 
+#[cfg(not(target_os = "freebsd"))]
 use std::collections::HashMap;
+#[cfg(not(target_os = "freebsd"))]
 use std::ffi::OsString;
```
RalfJung, line 10: Please move these imports to inside `test_directory` rather than having such global `cfg`

tests/pass/shims/fs.rs
```diff
@@ -296,6 +300,7 @@ fn test_canonicalize() {
     remove_dir_all(&dir_path).unwrap();
 }
 
+#[cfg(not(target_os = "freebsd"))]
```
RalfJung, line 303:
```suggestion
#[cfg(not(target_os = "freebsd"))] // FIXME freebsd does not support readdir yet
```

src/shims/unix/fs.rs
```diff
@@ -911,13 +907,16 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         this.try_unwrap_io_result(result)
     }
 
-    fn macos_stat(
+    fn macos_fbsd_stat(
```
RalfJung, line 910: All of these will need careful auditing since macos is 64bit-only. It's unclear whether the existing code makes sense for other OSes.

RalfJung:
Can you add a 32bit freebsd target to CI? With this exposure of previously 64bit-only code to 32bit targets, we definitely need that test coverage.

tests/pass/shims/fs-with-isolation.rs
devnexen: no I just forgot to enable it again..

src/shims/unix/foreign_items.rs
devnexen: true true I just forgot.

Cargo.lock
devnexen: was not intended.

src/shims/unix/fs.rs
RalfJung: And so the Rust libc crate then got that update as well at time point? Which libc crate version is that?

This [file here](https://github.com/rust-lang/rust/blob/91f7f266ce973a08e6e700f9ce815f7b61d347e9/Cargo.lock#L2167) should be defining which libc is picked up by Miri here. It's the latest version currently.

devnexen: in fact freebsd is one of the few oses in the libc's crate which maintains several os releases versions e.g. freebsd 11 up to 14 having their own definitions when there are changes.

src/shims/unix/freebsd/foreign_items.rs
devnexen: I m covering both cases like macos does.

src/shims/unix/fs.rs
RalfJung: I don't quite follow, can you point me at the code for that in the libc crate? Which type does one get when asking for `libc::dirent`?

devnexen: Sure
[freebsd 11](https://github.com/rust-lang/libc/blob/main/src/unix/bsd/freebsdlike/freebsd/freebsd11/mod.rs#L215) and [later](https://github.com/rust-lang/libc/blob/main/src/unix/bsd/freebsdlike/freebsd/freebsd13/mod.rs#L234)

src/shims/unix/freebsd/foreign_items.rs
RalfJung: In macOS we did this because those symbols actually existed, but are deprecated.

So copying this only makes sense if the symbols also exist(ed) on FreeBSD.

devnexen: they do exist but from freebsd 12 and onwards.

RalfJung: Okay, sounds good then. Please explain that in a comment.

src/shims/unix/fs.rs
RalfJung: Ah, I think I found it... [here](https://github.com/rust-lang/libc/blob/874e5277f968ca18420e121d1bd296cbefc47b1d/build.rs#L67). The libc build script runs some code to determine the FreeBSD version. When the FreeBSD standard library is built on a different OS (like e.g. when Miri CI tests the freebsd target on a Linux host), that will fail, and it defaults to freebsd11. But if you are using Miri on a freebsd host, it will use the host freebsd version.

Gosh, what a nightmare.^^ This is impossible to test properly.

RalfJung: Okay so what this means for Miri is we have to support all versions of freebsd since we can't know what the build script picks. But we can only test one of them. Amazing. Maybe long-term we want to work with the libc crate to overwrite the version detection inside Miri, though maybe that has other issues.

For this PR, it means:
- definitely use `project_field_named` to get the name_place, that part should be easy
- for writing all the other fields, do a dynamic check whether the `d_off` field is present. We can't test on "is this freebsd or not" since the type is different for different versions of freebsd, we need to check the presence of this field. You will need to add a new helper function that checks whether a field of a given name exists.

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -63,6 +64,33 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 this.write_scalar(Scalar::from_target_usize(len, this), dest)?;
             }
 
+            // File related shims
+            // For those, we both intercept `call` and `call@FBSD_1.0` symbols cases
+            // since freebsd 12 they can be expected
```
RalfJung, line 69:
```suggestion
            // For those, we both intercept `func` and `func@FBSD_1.0` symbols cases
            // since freebsd 12 they can be expected
```

I don't understand the second line. Who can expect what since freebsd 12?

ci.sh
```diff
@@ -108,7 +108,8 @@ case $HOST_TARGET in
     MIRI_TEST_TARGET=aarch64-unknown-linux-gnu run_tests
     MIRI_TEST_TARGET=aarch64-apple-darwin run_tests
     MIRI_TEST_TARGET=i686-pc-windows-gnu run_tests
-    MIRI_TEST_TARGET=x86_64-unknown-freebsd run_tests_minimal hello integer vec panic/panic concurrency/simple pthread-threadname libc-getentropy libc-getrandom libc-misc libc-fs atomic env align
+    MIRI_TEST_TARGET=x86_64-unknown-freebsd run_tests_minimal hello integer vec panic/panic concurrency/simple pthread-threadname libc-getentropy libc-getrandom libc-misc libc-fs fs atomic env align
+    MIRI_TEST_TARGET=i686-unknown-freebsd run_tests_minimal hello integer vec panic/panic concurrency/simple pthread-threadname libc-getentropy libc-getrandom libc-misc libc-fs fs atomic env align
```
RalfJung, line 112:
```suggestion
    MIRI_TEST_TARGET=x86_64-unknown-freebsd run_tests_minimal hello integer vec panic/panic concurrency/simple pthread-threadname libc-getentropy libc-getrandom libc-misc fs atomic env align
    MIRI_TEST_TARGET=i686-unknown-freebsd run_tests_minimal hello integer vec panic/panic concurrency/simple pthread-threadname libc-getentropy libc-getrandom libc-misc fs atomic env align
```
"fs" also enables "libc-fs"

src/helpers.rs
```diff
@@ -224,6 +224,20 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         bug!("No field named {} in type {}", name, base.layout().ty);
     }
 
+    fn project_field_hasname<P: Projectable<'tcx, Provenance>>(
+        &self,
+        base: &P,
+        name: &str,
+    ) -> bool {
```
RalfJung, line 231: This doesn't project anything, so the current name is confusing. What about `projectable_has_field`?

src/shims/unix/fs.rs
```diff
@@ -856,7 +852,7 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
             if let Some(file_descriptor) = this.machine.file_handler.handles.get_mut(&fd) {
                 let result = file_descriptor
                     .seek(communicate, seek_from)?
-                    .map(|offset| i64::try_from(offset).unwrap());
```
RalfJung, line 859: I actually liked the explicit type here, makes it a bit easier to follow what happens. So if `i64::try_from` still works please keep it.

src/shims/unix/fs.rs
```diff
@@ -1448,16 +1454,41 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
-                    &[
-                        ("d_ino", ino.into()),
-                        ("d_seekoff", 0),
-                        ("d_reclen", 0),
-                        ("d_namlen", file_name_len.into()),
-                        ("d_type", file_type.into()),
-                    ],
-                    &entry_place,
-                )?;
+                if this.tcx.sess.target.os == "macos" {
```
RalfJung, line 1457: Please don't switch on the OS at all, instead just do the `if this.project_field_hasname(&entry_place, "d_off")` part.

Also add a comment explaining why we need to do this.

tests/pass-dep/shims/libc-fs.rs
```diff
@@ -23,6 +23,7 @@ fn main() {
     test_file_open_unix_extra_third_arg();
     #[cfg(target_os = "linux")]
     test_o_tmpfile_flag();
+    #[cfg(not(all(target_os = "freebsd", target_pointer_width = "32")))]
```
RalfJung, line 26: Why does this fail on 32bit freebsd?

tests/pass/shims/fs.rs
```diff
@@ -4,30 +4,31 @@
...
+use std::fs::{canonicalize, create_dir, remove_dir_all, remove_file, rename, File, OpenOptions};
+use std::io::{Error, ErrorKind, Result, Write};
 use std::path::{Path, PathBuf};
 
 #[path = "../../utils/mod.rs"]
 mod utils;
 
 fn main() {
     test_path_conversion();
+    #[cfg(not(all(target_os = "freebsd", target_pointer_width = "32")))]
```
RalfJung, line 16: This is not looking good. The entire point of adding the 32bit test was to ensure that the code you are touching actually works on 32bit. If these tests are failing then looks like the code does not work on 32bit. That means there is a bug. Please fix the bug instead of disabling the tests.

You are taking code that was only ever tested on 64bit and exposing it to 32bit. It's not surprising there are bugs. But It's not okay to just hide them!

tests/pass/shims/fs.rs
```diff
@@ -296,7 +309,11 @@ fn test_canonicalize() {
     remove_dir_all(&dir_path).unwrap();
 }
 
+#[cfg(not(target_os = "freebsd"))] // FIXME: freebsd does not support readdir yet
```
RalfJung, line 312: Hm I just saw you implemented readdir. Why does this test fail?

RalfJung:
You need to make sure that *every single function* that you add for freebsd is actually called by some test, on 32bit and 64bit. This is the standard we expect for all our targets. Don't just disable tests, that's now how one develops reliable software.

@rustbot review

RalfJung:
Eh I meant @rustbot author.^^

Btw I am looking into why the existing libc-fs test is already failing on FreeBSD. Is it possible that `ftruncate` on 32bit FreeBSD takes a 64bit argument...?

RalfJung:
Yes, that does [seem to be the case on all BSDs](https://github.com/rust-lang/libc/blob/c461e304db322c53e8d0b28ae9192e9075b54a03/src/unix/bsd/mod.rs#L1). Oh wow.

src/helpers.rs
```diff
@@ -225,6 +225,21 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         bug!("No field named {} in type {}", name, base.layout().ty);
     }
 
+    /// Search if Project (which must be a struct or union type) contains the `name` field.
+    fn project_field_has_field<P: Projectable<'tcx, Provenance>>(
```
RalfJung, line 229:
```suggestion
    fn projectable_has_field<P: Projectable<'tcx, Provenance>>(
```

src/shims/unix/fs.rs
```diff
@@ -1448,16 +1454,43 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                            ("d_seekoff", 0),
+                            ("d_reclen", 0),
+                            ("d_namlen", file_name_len.into()),
+                            ("d_type", file_type.into()),
+                        ],
+                        &entry_place,
+                    )?;
+                } else {
+                    // freebsd 12 and onwards had added the d_off field
+                    if this.project_field_has_field(&entry_place, "d_off") {
```
RalfJung, line 1471: Please collapse this into `} else if {`

tests/pass/shims/fs.rs
```diff
@@ -123,7 +120,9 @@ fn test_file_create_new() {
     remove_file(&path).unwrap();
 }
 
+#[cfg(not(all(target_os = "freebsd", target_pointer_width = "32")))]
```
RalfJung, line 123: There's still something wrong here, why is this disabled on 32bit?
I said it multiple times: please don't just disable tests when they fail! Sometimes it is okay to cfg-out a particular test on a particular target, but that *always* needs a comment explaining why it is done. And here I would expect the test should pass on freebsd so when it fails that needs investigation, not `cfg`.

tests/pass/shims/fs.rs
```diff
@@ -64,6 +59,7 @@ fn test_path_conversion() {
 }
 
 fn test_file() {
+    use std::io::{IsTerminal, Read};
```
RalfJung, line 62: Seems unnecessary to do this for functions that are always enabled? This is only needed for cfg-gated functions.

RalfJung (commented):
We're getting there I think. :)

tests/pass/shims/fs.rs
devnexen: I forgot to reenable this one (especially there is no cfg upon test_seek above), no need to assume bad intentions ..

devnexen:
@saethlin I think I ve done all was asked can you have a look at it pls ?

src/helpers.rs
```diff
@@ -225,6 +225,21 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         bug!("No field named {} in type {}", name, base.layout().ty);
     }
 
+    /// Search if Project (which must be a struct or union type) contains the `name` field.
```
RalfJung, line 228:
```suggestion
    /// Search if `base` (which must be a struct or union type) contains the `name` field.
```

src/shims/unix/fs.rs
```diff
@@ -1448,16 +1454,41 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                            ("d_seekoff", 0),
+                            ("d_reclen", 0),
+                            ("d_namlen", file_name_len.into()),
+                            ("d_type", file_type.into()),
+                        ],
+                        &entry_place,
+                    )?;
+                    // freebsd 12 and onwards had added the d_off field
+                } else if this.projectable_has_field(&entry_place, "d_off") {
+                    this.write_int_fields_named(
```
RalfJung, line 1471:
```suggestion
                } else if this.projectable_has_field(&entry_place, "d_off") {
                    // freebsd 12 and onwards had added the d_off field
                    this.write_int_fields_named(
```

RalfJung:
Sorry for the long delay!
@bors r+

PR #3185, opened 2023-11-23
Refactor `float_to_int_checked` to remove its generic parameter and reduce code duplication a bit

eduardosm:


tests/fail/intrinsics/float_to_int_32_too_big1.stderr
```diff
@@ -1,8 +1,8 @@
-error: Undefined Behavior: `float_to_int_unchecked` intrinsic called on 2.14748365E+9 which cannot be represented in target type `i32`
+error: Undefined Behavior: `float_to_int_unchecked` intrinsic called on 2.14748365E+9f32 which cannot be represented in target type `i32`
   --> $DIR/float_to_int_32_too_big1.rs:LL:CC
    |
 LL |         float_to_int_unchecked::<f32, i32>(2147483648.0f32);
-   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `float_to_int_unchecked` intrinsic called on 2.14748365E+9 which cannot be represented in target type `i32`
+   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `float_to_int_unchecked` intrinsic called on 2.14748365E+9f32 which cannot be represented in target type `i32`
```
eduardosm, line 5: This happens because we are now printing the `ImmTy` instead of the apfloat

src/helpers.rs
```diff
@@ -986,65 +987,83 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
-                } else {
-                    // Floating point value can be represented by the integer type after rounding.
-                    // The INEXACT flag is ignored on purpose to allow rounding.
-                    Scalar::from_uint(res.value, size)
-                }
+        let int_size = cast_to.layout.size;
+        let (val, status) = match (src.layout.ty.kind(), cast_to.ty.kind()) {
+            // f32 to unsigned
+            (ty::Float(FloatTy::F32), ty::Uint(_)) => {
+                let res = src.to_scalar().to_f32().unwrap().to_u128_r(
```
RalfJung, line 1005: The `unwrap` here is suspicious, this should probably be propagated properly.

src/helpers.rs
```diff
@@ -986,65 +987,83 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                span_bug!(
+                    this.cur_span(),
+                    "attempted float-to-int conversion with non-float input type {}",
+                    cast_to.ty,
+                ),
             _ =>
                 span_bug!(
                     this.cur_span(),
                     "attempted float-to-int conversion with non-int output type {}",
                     cast_to.ty,
```
RalfJung, line 1050: Just merge these two arms, no need to give such detailed errors when there's a bug Miri.

src/helpers.rs
```diff
@@ -986,65 +987,83 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                (Scalar::from_int(res.value, int_size), res.status)
+            }
+            // f64 to unsigned
+            (ty::Float(FloatTy::F64), ty::Uint(_)) => {
+                let res = src.to_scalar().to_f64().unwrap().to_u128_r(
+                    int_size.bits_usize(),
+                    round,
+                    &mut false,
+                );
+                (Scalar::from_uint(res.value, int_size), res.status)
```
RalfJung, line 1028: If we ever have more than 2 float types I think a method with a generic parameter would still help reduce code size here... but as things stand now, it's just duplicating 4 lines of code, and the overhead of the separate methods is probably bigger than that.

src/shims/x86/mod.rs
```diff
@@ -331,6 +331,40 @@ fn bin_op_simd_float_all<'tcx, F: rustc_apfloat::Float>(
     Ok(())
 }
 
+/// Converts each element of `op` from floating point to signed integer.
+///
+/// When the input value is NaN or out of range, fall back to minimum value.
+///
+/// If `op` has more elements than `dest`, extra elements are ignored. If `op`
+/// has less elements than `dest`, remaining is filled with zeros.
```
RalfJung, line 339:
```suggestion
/// has less elements than `dest`, the rest is filled with zeros.
```

src/shims/x86/mod.rs
```diff
@@ -331,6 +331,40 @@ fn bin_op_simd_float_all<'tcx, F: rustc_apfloat::Float>(
...
+        let op = this.read_immediate(&this.project_index(&op, i)?)?;
+        let dest = this.project_index(&dest, i)?;
+
+        let res = this.float_to_int_checked(&op, dest.layout, rnd).unwrap_or_else(|| {
+            // Fallback to minimum acording to SSE/AVX semantics.
+            ImmTy::from_int(dest.layout.size.signed_int_min(), dest.layout)
+        });
+        this.write_immediate(*res, &dest)?;
+    }
+    // Fill remaining with zeros
```
RalfJung, line 359:
```suggestion
    // Fill remainder with zeros
```

src/shims/x86/mod.rs
```diff
@@ -331,6 +331,40 @@ fn bin_op_simd_float_all<'tcx, F: rustc_apfloat::Float>(
...
+    let (op, op_len) = this.operand_to_simd(op)?;
+    let (dest, dest_len) = this.place_to_simd(dest)?;
+
+    for i in 0..op_len.min(dest_len) {
+        let op = this.read_immediate(&this.project_index(&op, i)?)?;
+        let dest = this.project_index(&dest, i)?;
+
+        let res = this.float_to_int_checked(&op, dest.layout, rnd).unwrap_or_else(|| {
+            // Fallback to minimum acording to SSE/AVX semantics.
+            ImmTy::from_int(dest.layout.size.signed_int_min(), dest.layout)
```
RalfJung, line 355: Might be good to have an `assert!(matches!(dest.layout.ty, ty::Int(_)))` to make sure this is truly a signed output type.

tests/fail/intrinsics/float_to_int_32_inf1.stderr
```diff
@@ -1,8 +1,8 @@
-error: Undefined Behavior: `float_to_int_unchecked` intrinsic called on +Inf which cannot be represented in target type `i32`
+error: Undefined Behavior: `float_to_int_unchecked` intrinsic called on +Inff32 which cannot be represented in target type `i32`
```
RalfJung, line 1: `Inff32` looks a bit strange... that's probably a bug in immediate printing. Can you file an issue to track this?

src/shims/x86/sse2.rs
```diff
@@ -260,37 +260,25 @@ pub(super) trait EvalContextExt<'mir, 'tcx: 'mir>:
...
                     // "current SSE rounding mode", assume nearest
                     // https://www.felixcloutier.com/x86/cvtps2dq
-                    "cvtps2dq" => rustc_apfloat::Round::NearestTiesToEven,
+                    // https://www.felixcloutier.com/x86/cvtpd2dq
+                    "cvtps2dq" | "cvtpd2dq" => rustc_apfloat::Round::NearestTiesToEven,
                     // always truncate
                     // https://www.felixcloutier.com/x86/cvttps2dq
-                    "cvttps2dq" => rustc_apfloat::Round::TowardZero,
+                    // https://www.felixcloutier.com/x86/cvttpd2dq
+                    "cvttps2dq" | "cvttpd2dq" => rustc_apfloat::Round::TowardZero,
```
RalfJung, line 277: This looks odd, it seems as if two of the intrinsics have the exact same semantics. Can you add back some assertion about the vector lengths?

RalfJung:
Love to see a cleanup, thanks a lot. :)

src/shims/x86/sse2.rs
eduardosm: Done: https://github.com/rust-lang/rust/issues/118221

tests/fail/intrinsics/float_to_int_32_inf1.stderr
eduardosm: Done: https://github.com/rust-lang/rust/issues/118221

src/helpers.rs
eduardosm: I ended adding an inner function with a generic parameter. It did not make a significant difference in code size, but I think it looks cleaner. But that means we cannot merge the `span_bug!`s

src/shims/x86/mod.rs
```diff
@@ -331,6 +331,42 @@ fn bin_op_simd_float_all<'tcx, F: rustc_apfloat::Float>(
...
+fn convert_float_to_int<'tcx>(
+    this: &mut crate::MiriInterpCx<'_, 'tcx>,
+    op: &OpTy<'tcx, Provenance>,
+    rnd: rustc_apfloat::Round,
+    dest: &PlaceTy<'tcx, Provenance>,
+) -> InterpResult<'tcx, ()> {
+    let (op, op_len) = this.operand_to_simd(op)?;
+    let (dest, dest_len) = this.place_to_simd(dest)?;
+
+    assert!(matches!(dest.layout.field(this, 0).ty.kind(), ty::Int(_)));
```
RalfJung, line 349:
```suggestion
    // Output must be *signed* integers.
    assert!(matches!(dest.layout.field(this, 0).ty.kind(), ty::Int(_)));
```

RalfJung:
Looks good, just one more comment nit. :) Then r=me.

@bors delegate+

PR #3189, opened 2023-11-25
run the provenance-gc=1 test on all targets, but only for the host tests

RalfJung:
No need to slow down *all those tests* running on the Linux host... but lets cover each major OS at least once. We've had bugs that only some macOS-specific code in `getrandom` found, after all.

Let's see how much this affects timing on the macOS / Windows runners.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,10 +35,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Set the tag GC interval to 1 on linux
-        if: runner.os == 'Linux'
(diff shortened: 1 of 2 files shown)


RalfJung:
On macOS this seems to be fine, though times are so noisy there it's hard to tell.

On Windows this does seem like a noticeable slowdown.

RalfJung:
Locally,  `./miri test` goes from 16s to 1m9s with this flag. So yeah it's very noticeable.

RalfJung:
This seems like a reasonable trade-off

    before: only on Linux host, all tests
    after: only the test suite itself (not cargo-miri or the mir-opt-level=4 run),
    on all hosts for the host target and on Linux for all "full" targets.


@bors r+

PR #3192, opened 2023-11-26
Implement x86 AVX intrinsics

eduardosm:
~Blocked on <https://github.com/rust-lang/miri/pull/3214>~

### DIFF
--- a/src/shims/x86/avx.rs
+++ b/src/shims/x86/avx.rs
@@ -1,0 +1,417 @@
+use rustc_apfloat::{ieee::Double, ieee::Single};
+use rustc_middle::mir;
+use rustc_middle::ty::layout::LayoutOf as _;
+use rustc_middle::ty::Ty;
+use rustc_span::Symbol;
+use rustc_target::spec::abi::Abi;
+
+use super::{
+    bin_op_simd_float_all, conditional_dot_product, convert_float_to_int, horizontal_bin_op,
+    round_all, test_bits_masked, test_high_bits_masked, unary_op_ps, FloatBinOp, FloatUnaryOp,
+};
+use crate::*;
+use shims::foreign_items::EmulateForeignItemResult;
+
+impl<'mir, 'tcx: 'mir> EvalContextExt<'mir, 'tcx> for crate::MiriInterpCx<'mir, 'tcx> {}
+pub(super) trait EvalContextExt<'mir, 'tcx: 'mir>:
+    crate::MiriInterpCxExt<'mir, 'tcx>
+{
+    fn emulate_x86_avx_intrinsic(
+        &mut self,
+        link_name: Symbol,
+        abi: Abi,
+        args: &[OpTy<'tcx, Provenance>],
+        dest: &PlaceTy<'tcx, Provenance>,
+    ) -> InterpResult<'tcx, EmulateForeignItemResult> {
+        let this = self.eval_context_mut();
+        this.expect_target_feature_for_intrinsic(link_name, "avx")?;
+        // Prefix should have already been checked.
+        let unprefixed_name = link_name.as_str().strip_prefix("llvm.x86.avx.").unwrap();
+
+        match unprefixed_name {
+            // Used to implement _mm256_min_ps and _mm256_max_ps functions.
+            // Note that the semantics are a bit different from Rust simd_min
+            // and simd_max intrinsics regarding handling of NaN and -0.0: Rust
+            // matches the IEEE min/max operations, while x86 has different
+            // semantics.
+            "min.ps.256" | "max.ps.256" => {
+                let [left, right] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let which = match unprefixed_name {
+                    "min.ps.256" => FloatBinOp::Min,
+                    "max.ps.256" => FloatBinOp::Max,
+                    _ => unreachable!(),
+                };
+
+                bin_op_simd_float_all::<Single>(this, which, left, right, dest)?;
+            }
+            // Used to implement _mm256_min_pd and _mm256_max_pd functions.
+            "min.pd.256" | "max.pd.256" => {
+                let [left, right] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let which = match unprefixed_name {
+                    "min.pd.256" => FloatBinOp::Min,
+                    "max.pd.256" => FloatBinOp::Max,
+                    _ => unreachable!(),
+                };
+
+                bin_op_simd_float_all::<Double>(this, which, left, right, dest)?;
+            }
+            // Used to implement the _mm256_round_ps function.
+            // Rounds the elements of `op` according to `rounding`.
+            "round.ps.256" => {
+                let [op, rounding] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                round_all::<rustc_apfloat::ieee::Single>(this, op, rounding, dest)?;
+            }
+            // Used to implement the _mm256_round_pd function.
+            // Rounds the elements of `op` according to `rounding`.
+            "round.pd.256" => {
+                let [op, rounding] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                round_all::<rustc_apfloat::ieee::Double>(this, op, rounding, dest)?;
+            }
+            // Used to implement _mm256_{sqrt,rcp,rsqrt}_ps functions.
+            // Performs the operations on all components of `op`.
+            "sqrt.ps.256" | "rcp.ps.256" | "rsqrt.ps.256" => {
+                let [op] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let which = match unprefixed_name {
+                    "sqrt.ps.256" => FloatUnaryOp::Sqrt,
+                    "rcp.ps.256" => FloatUnaryOp::Rcp,
+                    "rsqrt.ps.256" => FloatUnaryOp::Rsqrt,
+                    _ => unreachable!(),
+                };
+
+                unary_op_ps(this, which, op, dest)?;
+            }
+            // Used to implement the _mm256_dp_ps function.
+            "dp.ps.256" => {
+                let [left, right, imm] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                conditional_dot_product(this, left, right, imm, dest)?;
+            }
+            // Used to implement the _mm256_h{add,sub}_p{s,d} functions.
+            // Horizontally add/subtract adjacent floating point values
+            // in `left` and `right`.
+            "hadd.ps.256" | "hadd.pd.256" | "hsub.ps.256" | "hsub.pd.256" => {
+                let [left, right] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let which = match unprefixed_name {
+                    "hadd.ps.256" | "hadd.pd.256" => mir::BinOp::Add,
+                    "hsub.ps.256" | "hsub.pd.256" => mir::BinOp::Sub,
+                    _ => unreachable!(),
+                };
+
+                horizontal_bin_op(this, which, /*saturating*/ false, left, right, dest)?;
+            }
+            // Used to implement the _mm256_cmp_ps function.
+            // Performs a comparison operation on each component of `left`
+            // and `right`. For each component, returns 0 if false or u32::MAX
+            // if true.
+            "cmp.ps.256" => {
+                let [left, right, imm] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let which =
+                    FloatBinOp::cmp_from_imm(this, this.read_scalar(imm)?.to_i8()?, link_name)?;
+
+                bin_op_simd_float_all::<Single>(this, which, left, right, dest)?;
+            }
+            // Used to implement the _mm256_cmp_pd function.
+            // Performs a comparison operation on each component of `left`
+            // and `right`. For each component, returns 0 if false or u64::MAX
+            // if true.
+            "cmp.pd.256" => {
+                let [left, right, imm] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let which =
+                    FloatBinOp::cmp_from_imm(this, this.read_scalar(imm)?.to_i8()?, link_name)?;
+
+                bin_op_simd_float_all::<Double>(this, which, left, right, dest)?;
+            }
+            // Used to implement the _mm256_cvtps_epi32, _mm256_cvttps_epi32, _mm256_cvtpd_epi32
+            // and _mm256_cvttpd_epi32 functions.
+            // Converts packed f32/f64 to packed i32.
+            "cvt.ps2dq.256" | "cvtt.ps2dq.256" | "cvt.pd2dq.256" | "cvtt.pd2dq.256" => {
+                let [op] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let rnd = match unprefixed_name {
+                    // "current SSE rounding mode", assume nearest
+                    "cvt.ps2dq.256" | "cvt.pd2dq.256" => rustc_apfloat::Round::NearestTiesToEven,
+                    // always truncate
+                    "cvtt.ps2dq.256" | "cvtt.pd2dq.256" => rustc_apfloat::Round::TowardZero,
+                    _ => unreachable!(),
+                };
+
+                convert_float_to_int(this, op, rnd, dest)?;
+            }
+            // Used to implement the _mm_permutevar_ps and _mm256_permutevar_ps functions.
+            // Shuffles 32-bit floats from `data` using `control` as control. Each 128-bit
+            // chunk is shuffled independently: this means that we view the vector as a
+            // sequence of 4-element arrays, and we shuffle each of these arrays, where
+            // `control` determines which element of the current `data` array is written.
+            "vpermilvar.ps" | "vpermilvar.ps.256" => {
+                let [data, control] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let (data, data_len) = this.operand_to_simd(data)?;
+                let (control, control_len) = this.operand_to_simd(control)?;
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                assert_eq!(dest_len, data_len);
+                assert_eq!(dest_len, control_len);
+
+                for i in 0..dest_len {
+                    let control = this.project_index(&control, i)?;
+
+                    // Each 128-bit chunk is shuffled independently. Since each chunk contains
+                    // four 32-bit elements, only two bits from `control` are used. To read the
+                    // value from the current chunk, add the destination index truncated to a multiple
+                    // of 4.
+                    let chunk_base = i & !0b11;
+                    let src_i = u64::from(this.read_scalar(&control)?.to_u32()? & 0b11)
+                        .checked_add(chunk_base)
+                        .unwrap();
+
+                    this.copy_op(
+                        &this.project_index(&data, src_i)?,
+                        &this.project_index(&dest, i)?,
+                    )?;
+                }
+            }
+            // Used to implement the _mm_permutevar_pd and _mm256_permutevar_pd functions.
+            // Shuffles 64-bit floats from `left` using `right` as control. Each 128-bit
+            // chunk is shuffled independently: this means that we view the vector as
+            // a sequence of 2-element arrays, and we shuffle each of these arrays,
+            // where `right` determines which element of the current `left` array is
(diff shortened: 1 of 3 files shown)


RalfJung:
Thanks for the PR! Unfortunately I am exceptionally busy right now and honestly I am not sure if I'll be able to review a PR of this size before the Christmas break. Sorry for that.

workingjubilee:
Perhaps if @eduardosm splits out the "mere code movement" parts with rounding and sqrt into a separate PR so that the maskload, etc. stuff can be reviewed separately?

RalfJung:
Splitting up is always helpful, yeah. Reviewing has a strictly superlinear time complexity, I suspect somewhere between `O(n * log n)` and `O(n^2)`, so two smaller PRs are less work to review than a single PR with twice the size.

workingjubilee:
Yeah. I was also thinking "hm, I could help examine the tricky parts and doublecheck stuff, but Ralf knows how stuff is/should be organized better than me (obviously), so I have no comment on 'where should `enum FloatUnaryOp` live, and/or does this abstraction need to be changed now that it's seeing more use?"

eduardosm:
Done, created https://github.com/rust-lang/miri/pull/3214 with the "just moving things" part

src/shims/x86/avx.rs
```diff
@@ -0,0 +1,413 @@
...
+            "cmp.pd.256" => {
+                let [left, right, imm] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let which =
+                    FloatBinOp::cmp_from_imm(this, this.read_scalar(imm)?.to_i8()?, link_name)?;
+
+                bin_op_simd_float_all::<Double>(this, which, left, right, dest)?;
+            }
+            "cvt.ps2dq.256" | "cvtt.ps2dq.256" | "cvt.pd2dq.256" | "cvtt.pd2dq.256" => {
```
RalfJung, line 140: This doesn't seem to have a comment?

src/shims/x86/avx.rs
```diff
@@ -0,0 +1,413 @@
...
+                    "cvt.ps2dq.256" | "cvt.pd2dq.256" => rustc_apfloat::Round::NearestTiesToEven,
+                    // always truncate
+                    "cvtt.ps2dq.256" | "cvtt.pd2dq.256" => rustc_apfloat::Round::TowardZero,
+                    _ => unreachable!(),
+                };
+
+                convert_float_to_int(this, op, rnd, dest)?;
+            }
+            // Used to implement the _mm256_zeroall and _mm256_zeroupper functions.
+            // They are use to fully or partially zero YMM registers. Register holding
```
RalfJung, line 154:
```suggestion
            // They are used to fully or partially zero YMM registers. Register holding
```

src/shims/x86/avx.rs
```diff
@@ -0,0 +1,413 @@
...
+            // useful data are temporarilly saved to stack, so they do not cause any
+            // effect visible from Rust, making them a no-op for Miri.
+            "vzeroall" | "vzeroupper" => {
+                let [] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+            }
```
RalfJung, line 159: Can these even be used reliably from Rust? The register allocator could clobber those registers again immediately after this intrinsic is done, I think?

src/shims/x86/avx.rs
```diff
@@ -0,0 +1,413 @@
...
+    let high_bit_offset = mask_item_size.bits().checked_sub(1).unwrap();
+
+    let ptr = this.read_pointer(ptr)?;
+    for i in 0..dest_len {
+        let mask = this.project_index(&mask, i)?;
+        let dest = this.project_index(&dest, i)?;
+
+        if this.read_scalar(&mask)?.to_uint(mask_item_size)? >> high_bit_offset != 0 {
+            // Size * u64 is implemented as always checked
+            #[allow(clippy::arithmetic_side_effects)]
```
RalfJung, line 370: I'm surprised the lint fires here? AFAIK we use this multiplication in quite a few places?

src/shims/x86/avx.rs
```diff
@@ -0,0 +1,413 @@
...
+            let ptr = ptr.offset(dest.layout.size * i, &this.tcx)?;
+            let src = this.ptr_to_mplace(ptr, dest.layout);
+            this.copy_op(&src, &dest, /*allow_transmute*/ false)?;
```
RalfJung, line 373: So this will use require the pointer to be aligned to lane/element type. Is that the correct behavior?

RalfJung (commented):
Thanks! Mostly looks good, I have some nits and questions though.

I didn't look at the tests (yet); I assume they cover all the new intrinsics and in particular the corner cases.

tests/pass/intrinsics-x86-avx.rs
```diff
@@ -159,4 +724,511 @@ unsafe fn test_avx() {
...
+    #[target_feature(enable = "avx")]
+    unsafe fn test_mm_maskload_ps() {
+        let a = &[1.0f32, 2., 3., 4.];
+        let p = a.as_ptr();
+        let mask = _mm_setr_epi32(0, !0, 0, !0);
+        let r = _mm_maskload_ps(black_box(p), mask);
+        let e = _mm_setr_ps(0., 2., 0., 4.);
+        assert_eq_m128(r, e);
+    }
+    test_mm_maskload_ps();
```
RalfJung, line 901: Could you add maskload/maskstore tests where one of the masked-off elements is OOB, i.e., actually loading it would be UB?

src/shims/x86/avx.rs
eduardosm: Yeah, inline assembly should be used instead if you need this. I implemented them since they were quite easy.

src/shims/x86/avx.rs
eduardosm: The [arithmetic_side_effects](https://rust-lang.github.io/rust-clippy/master/index.html#/arithmetic_side_effects) lint is enabled by default for custom types.

src/shims/x86/avx.rs
eduardosm: Hmmm, according to the [intel manual](https://software.intel.com/en-us/download/intel-64-and-ia-32-architectures-sdm-combined-volumes-1-2a-2b-2c-2d-3a-3b-3c-3d-and-4), page 2357:

> See Table 2-23, “Type 6 Class Exception Conditions” (No AC# reported for any mask bit combinations).

(AC stands for "Alignment Check")

So I believe those can actually be unaligned.

src/shims/x86/avx.rs
RalfJung: Sounds similar to the "read floating point status register" intrinsics then. Maybe it'd be better not to implement this for now.

src/shims/x86/avx.rs
RalfJung: Oh, interesting. 

Is this the first operation in the SSE/AVX family that loads things from a pointer, where alignment questions can come up?

RalfJung: Please add a test for alignment as well then.

src/shims/x86/avx.rs
```diff
@@ -0,0 +1,414 @@
...
+
+    let ptr = this.read_pointer(ptr)?;
+    for i in 0..dest_len {
+        let mask = this.project_index(&mask, i)?;
+        let dest = this.project_index(&dest, i)?;
+
+        if this.read_scalar(&mask)?.to_uint(mask_item_size)? >> high_bit_offset != 0 {
+            // Size * u64 is implemented as always checked
+            #[allow(clippy::arithmetic_side_effects)]
+            let ptr = ptr.offset(dest.layout.size * i, &this.tcx)?;
```
RalfJung, line 374: FWIW this is a *non-inbounds* offset that is UB only if it overflows the address space. That might well be what we want but it'd be worth stating this explicitly.

RalfJung: Actually we probably want `wrapping_offset` I think?

tests/pass/intrinsics-x86-avx.rs
RalfJung: Similarly, it'd be good to have tests where the mask is `[0, 0, 0, !0]` and only the last element actually exists, the rest would be OOB.

src/shims/x86/avx.rs
eduardosm: To allow something like this?
```rust
let a = &[2.0f32];
let p = a.as_ptr().wrapping_sub(3); // out of bounds
let mask = _mm_setr_epi32(0, 0, 0, !0);
let r = _mm_maskload_ps(p, mask);
```

RalfJung: Yes, exactly.

src/shims/x86/mod.rs
```diff
@@ -575,8 +583,54 @@ fn convert_float_to_int<'tcx>(
...
+
+    Ok((num_chunks, items_per_chunk, left, right, dest))
+}
+
 /// Horizontaly performs `which` operation on adjacent values of
 /// `left` and `right` SIMD vectors and stores the result in `dest`.
+///
+/// Each 128-bit chunk is treated independently (i.e., the value for
+/// the is i-th 128-bit chunk of `dest` is calculated with the i-th
+/// 128-bit blocks of `left` and `right`).
```
RalfJung, line 633: Doesn't that mean the output should be shorter than the input, since each 128bit chunk produces one result?

src/shims/x86/mod.rs
```diff
@@ -575,8 +583,54 @@ fn convert_float_to_int<'tcx>(
...
+///
+/// `left`, `right` and `dest` cannot have different types.
+fn split_simd_to_128bit_lanes<'tcx>(
+    this: &mut crate::MiriInterpCx<'_, 'tcx>,
+    left: &OpTy<'tcx, Provenance>,
+    right: &OpTy<'tcx, Provenance>,
+    dest: &PlaceTy<'tcx, Provenance>,
+) -> InterpResult<
+    'tcx,
+    (u64, u64, MPlaceTy<'tcx, Provenance>, MPlaceTy<'tcx, Provenance>, MPlaceTy<'tcx, Provenance>),
```
RalfJung, line 597: The doc comment should explain the return value.

src/shims/x86/mod.rs
```diff
@@ -620,46 +676,54 @@ fn horizontal_bin_op<'tcx>(
...
-    for i in 0..left_len {
-        if imm & (1 << i.checked_add(4).unwrap()) != 0 {
-            let left = this.read_immediate(&this.project_index(&left, i)?)?;
-            let right = this.read_immediate(&this.project_index(&right, i)?)?;
-
-            let mul = this.wrapping_binary_op(mir::BinOp::Mul, &left, &right)?;
-            sum = this.wrapping_binary_op(mir::BinOp::Add, &sum, &mul)?;
+        // Calculate dot product
+        // Elements are floating point numbers, but we can use `from_int`
+        // because the representation of 0.0 is all zero bits.
```
RalfJung, line 706:
```suggestion
        // Elements are floating point numbers, but we can use `from_int`
        // for the initial value because the representation of 0.0 is all zero bits.
```

RalfJung (commented):
All right, finally went through all the ops and comments. Sorry for the long wait!

src/shims/x86/avx.rs
```diff
@@ -0,0 +1,407 @@
...
+    let ptr = this.read_pointer(ptr)?;
+    for i in 0..dest_len {
+        let mask = this.project_index(&mask, i)?;
+        let dest = this.project_index(&dest, i)?;
+
+        if this.read_scalar(&mask)?.to_uint(mask_item_size)? >> high_bit_offset != 0 {
+            // Size * u64 is implemented as always checked
+            #[allow(clippy::arithmetic_side_effects)]
+            let ptr = ptr.wrapping_offset(dest.layout.size * i, &this.tcx);
+            this.mem_copy(ptr, dest.ptr(), dest.layout.size, /*nonoverlapping*/ true)?;
```
RalfJung, line 368:
```suggestion
            // Unaligned copy, which is what we want.
            this.mem_copy(ptr, dest.ptr(), dest.layout.size, /*nonoverlapping*/ true)?;
```

src/shims/x86/avx.rs
```diff
@@ -0,0 +1,407 @@
...
+    let ptr = this.read_pointer(ptr)?;
+    for i in 0..value_len {
+        let mask = this.project_index(&mask, i)?;
+        let value = this.project_index(&value, i)?;
+
+        if this.read_scalar(&mask)?.to_uint(mask_item_size)? >> high_bit_offset != 0 {
+            // Size * u64 is implemented as always checked
+            #[allow(clippy::arithmetic_side_effects)]
+            let ptr = ptr.wrapping_offset(value.layout.size * i, &this.tcx);
+            this.mem_copy(value.ptr(), ptr, value.layout.size, /*nonoverlapping*/ true)?;
```
RalfJung, line 402:
```suggestion
            // Unaligned copy, which is what we want.
            this.mem_copy(value.ptr(), ptr, value.layout.size, /*nonoverlapping*/ true)?;
```

src/shims/x86/avx.rs
```diff
@@ -0,0 +1,407 @@
...
+            // Reads a 256-bit vector from an unaligned pointer. This intrinsic
+            // is expected to perform better than a regular unaligned read when
+            // the data crosses a cache line, but for Miri this is just a regular
+            // unaligned read.
+            "ldu.dq.256" => {
+                let [src_ptr] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let src_ptr = this.read_pointer(src_ptr)?;
+                let dest = dest.force_mplace(this)?;
+
+                this.mem_copy(src_ptr, dest.ptr(), dest.layout.size, /*nonoverlapping*/ true)?;
```
RalfJung, line 294:
```suggestion
                // Unaligned copy, which is what we want.
                this.mem_copy(src_ptr, dest.ptr(), dest.layout.size, /*nonoverlapping*/ true)?;
```

src/shims/x86/avx.rs
```diff
@@ -0,0 +1,407 @@
...
+                    // always truncate
+                    "cvtt.ps2dq.256" | "cvtt.pd2dq.256" => rustc_apfloat::Round::TowardZero,
+                    _ => unreachable!(),
+                };
+
+                convert_float_to_int(this, op, rnd, dest)?;
+            }
+            // Used to implement the _mm_permutevar_ps and _mm256_permutevar_ps functions.
+            // Shuffles 32-bit floats from `left` using `right` as control. Each 128-bit
+            // lane is shuffled independently.
```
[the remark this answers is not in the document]
RalfJung: Yeah, "chunk" or "block" would be better I think. "chunk" matches what the standard library calls this.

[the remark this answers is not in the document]
RalfJung: You added this comment for the pd intrinsics but not the ps ones.

src/shims/x86/avx.rs
```diff
@@ -0,0 +1,407 @@
...
+                    this.copy_op(
+                        &this.project_index(&left, src_i)?,
+                        &this.project_index(&dest, i)?,
+                        /*allow_transmute*/ false,
+                    )?;
+                }
+            }
+            // Used to implement the _mm256_permute2f128_ps, _mm256_permute2f128_pd and
+            // _mm256_permute2f128_si256 functions.
+            // For each 128-bit lane of `dest`, copies one of the lanes of `left` or `right`,
```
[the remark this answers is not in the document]
RalfJung: Here it still seems easier to consider this an array of 128bit elements, I think? There is no smaller element type of which we are considering 128bit array chunks, is there?

src/shims/x86/avx.rs
```diff
@@ -0,0 +1,407 @@
...
+                            _ => unreachable!(),
+                        };
+                        this.copy_op(&src, &dest, /*allow_transmute*/ false)?;
+                    }
+                }
+            }
+            // Used to implement the _mm_maskload_ps, _mm_maskload_pd, _mm256_maskload_ps
+            // and _mm256_maskload_pd functions.
+            // For each element, if high bit of the matching element of `mask`
+            // is one, it is loaded from `ptr`, otherwise zero is loaded.
```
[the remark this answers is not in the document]
eduardosm: It is, I will update the comment

src/shims/x86/mod.rs
eduardosm: The output has the same size as each input, since each 128-bit chunk of the input produces a 128-bit chunk for the output.

RalfJung:
(FYI, I will assume the ball is in your court until you do `rustbot ready`)

src/shims/x86/avx.rs
```diff
@@ -0,0 +1,418 @@
...
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                assert_eq!(dest_len, data_len);
+                assert_eq!(dest_len, control_len);
+
+                for i in 0..dest_len {
+                    let control = this.project_index(&control, i)?;
+
+                    // Each 128-bit lane is shuffled independently. Since each lane contains
+                    // two 64-bit elements, only the second bit from `right` is used (yes, the
```
RalfJung, line 212: `right` does not exist any more, should probably be `control` now?

src/shims/x86/avx.rs
```diff
@@ -0,0 +1,418 @@
...
+                    // second instead of the first, ask Intel). To read the value from the current
+                    // lane, add the destination index truncated to a multiple of 2.
+                    let src_i = ((this.read_scalar(&control)?.to_u64()? >> 1) & 1)
+                        .checked_add(i & !1)
```
RalfJung, line 216: Similar to above, `i & !1` is the `chunk_base`, isn't it?

src/shims/x86/avx.rs
```diff
@@ -0,0 +1,418 @@
...
+                let (control, control_len) = this.operand_to_simd(control)?;
+                let (dest, dest_len) = this.place_to_simd(dest)?;
+
+                assert_eq!(dest_len, data_len);
+                assert_eq!(dest_len, control_len);
+
+                for i in 0..dest_len {
+                    let control = this.project_index(&control, i)?;
+
+                    // Each 128-bit lane is shuffled independently. Since each lane contains
```
RalfJung, line 211: Should these `lane` be `chunk`?

src/shims/x86/mod.rs
RalfJung: I guess I don't understand what is meant by performing an operation "horizontally".

Given two vectors, I can see how can one do an operation "pointwise" (that's what it would be called in mathematics, anyway): the i-th output element is computed as `op(left[i], right[i])`. Is that what you mean by "horizontally"? If not, what do you mean?

RalfJung (commented):
@rustbot author

src/shims/x86/mod.rs
RalfJung: Okay I stored at the code some more, and I think it's composing neighboring elements?
- concatenate `left ++ right` to form `input`
- then compute `output[i]` as `op(input[2*i], input[2*i + 1])`

Is that correct?

src/shims/x86/avx.rs
eduardosm: It is, since each element is 64 bits, and each chunk is 128 bits, a chunk has two elements, so we chop off the lowest bit to get the chunk base.

src/shims/x86/avx.rs
eduardosm: Oops, I missed those

src/shims/x86/mod.rs
eduardosm: That's correct

src/shims/x86/avx.rs
RalfJung: Okay, then let-bind it like above please. :)

src/shims/x86/mod.rs
```diff
@@ -575,8 +583,65 @@ fn convert_float_to_int<'tcx>(
...
+    let dest = dest.transmute(chunked_layout, this)?;
+
+    Ok((num_chunks, items_per_chunk, left, right, dest))
+}
+
 /// Horizontaly performs `which` operation on adjacent values of
 /// `left` and `right` SIMD vectors and stores the result in `dest`.
+/// "Horizontal" means that the operation the i-th element is calculated
+/// from the elements 2*i and 2*i+1 of the concatenation of `left` and
+/// `right`.
```
RalfJung, line 640:
```suggestion
/// "Horizontal" means that the i-th output element is calculated
/// from the elements 2*i and 2*i+1 of the concatenation of `left` and
/// `right`.
```

src/shims/x86/avx.rs
```diff
@@ -0,0 +1,419 @@
...
+                    let chunk_base = i & !0b11;
+                    let src_i = u64::from(this.read_scalar(&control)?.to_u32()? & 0b11)
+                        .checked_add(chunk_base)
+                        .unwrap();
+
+                    this.copy_op(
+                        &this.project_index(&data, src_i)?,
+                        &this.project_index(&dest, i)?,
+                        /*allow_transmute*/ false,
+                    )?;
```
RalfJung, line 188: The signature of copy_op changed, this needs a rebase.

PR #3198, opened 2023-12-01
illumos support

devnexen:


RalfJung:
@devnexen what is the status of this PR? You opened it as a draft, so I assume you are not currently looking for feedback. If you don't plan to work on it in the near future, it'd be better to close the PR -- this way the open PRs reflect things that people are actively working on. We can always reopen when you have time to get back to this!

RalfJung:

Was this meant to ask for review? Github doesn't seem to send notifications on this event, so I missed this -- sorry. In the future, please just post a comment when a PR is ready. You can write `@rustbot ready` to automatically update the labels as well.

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -0,0 +1,70 @@
...
+                let [thread, name, len] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let res = this.pthread_getname_np(
+                    this.read_scalar(thread)?,
+                    this.read_scalar(name)?,
+                    this.read_scalar(len)?,
+                )?;
+                this.write_scalar(res, dest)?;
+            }
+            "getrandom" => {
```
RalfJung, line 54: You left this in the "Threading" section of shims. Please add a separate section, like we do for the other OSes.

ci/ci.sh
```diff
@@ -123,7 +123,7 @@ case $HOST_TARGET in
     # Some targets are only partially supported.
     MIRI_TEST_TARGET=x86_64-unknown-freebsd run_tests_minimal hello integer vec panic/panic concurrency/simple pthread-threadname libc-getentropy libc-getrandom libc-misc libc-fs atomic env align num_cpus
     MIRI_TEST_TARGET=i686-unknown-freebsd run_tests_minimal hello integer vec panic/panic concurrency/simple pthread-threadname libc-getentropy libc-getrandom libc-misc libc-fs atomic env align num_cpus
-
+    MIRI_TEST_TARGET=x86_64-unknown-illumos run_tests_minimal no_std
```
RalfJung, line 126: This won't test most of the code you added. Please make sure we don't have dead code.

Since this is not a no_std target, please definitely add `hello`.

For the threading functions, you probably should add `concurrency/simple concurrency/thread_name pthread-threadname`.

For errno and since you edited that test, `libc-misc` should probably also be added. However libc-misc has a bunch of stuff so I'm not sure if it will fully work.

For getrandom/getentropy, add `libc-getrandom libc-getentropy`.

RalfJung (commented):
The code mostly looks good, but is lacking test coverage.

Also, it will need a rebase to resolve conflicts.

RalfJung:
Also, after rebasing, please add solaris/illumos to the unofficial target list in the README and set yourself as maintainer.

ci/ci.sh
```diff
@@ -148,6 +148,7 @@ case $HOST_TARGET in
     MIRI_TEST_TARGET=aarch64-linux-android  run_tests_minimal $VERY_BASIC hello panic/panic
     MIRI_TEST_TARGET=wasm32-wasi run_tests_minimal $VERY_BASIC wasm
     MIRI_TEST_TARGET=wasm32-unknown-unknown run_tests_minimal $VERY_BASIC wasm
+    MIRI_TEST_TARGET=x86_64-unknown-illumos run_tests_minimal libc-misc concurrency/simple pthread-threadname libc-getentropy libc-getrandom
```
RalfJung, line 151:
```suggestion
    MIRI_TEST_TARGET=x86_64-unknown-illumos run_tests_minimal $VERY_BASIC hello libc-misc concurrency/simple threadname libc-getentropy libc-getrandom
```
VERY_BASIC should always be tested, and the docs say `hello` should work so let's test that.

Also can you move this up next to Android?

RalfJung:
Looks like something changed in `std`, and `pthread_condattr_init` is now required to make it into `main`.

RalfJung:
> Looks like something changed in `std`, and `pthread_condattr_init` is now required to make it into `main`.

Or, rather, previously only no_std was tested so the pre-main code was not even executed.

I think the PR should be adjusted to add only what is needed for `main`. getentropy, getrandom, threadname... those are all nice things to have eventually, but we have to walk before we can run.

The goal should be for `MIRI_TEST_TARGET=x86_64-unknown-illumos ./miri test vec` to pass.

devnexen:
> Looks like something changed in `std`, and `pthread_condattr_init` is now required to make it into `main`.

Do I need to implement something in particular to fix the out-of-bounds issue ?

RalfJung:
That looks like the pthread offsets don't work for Solarish.

This seems to be a mutex, so this is the relevant comment:

https://github.com/rust-lang/miri/blob/baf32fd5c019785c26836145a4b0eca54326217c/src/shims/unix/sync.rs#L61-L68

It looks like something else is writing to the field that Miri uses to store the MutexId, and then Miri doesn't deal well with the invalid MutexId.

RalfJung:
Ah no, it's not something else writing... the problem is the layout needs to be compatible with the PTHREAD_MUTEX_INITIALIZER. For Solarish that's defined [here](https://github.com/rust-lang/libc/blob/79c1a811dbb6f46203a2ed65ddc9095d0e6a5a9f/src/unix/solarish/mod.rs#L2124-L2132). This is the type declaration:
```rust
    pub struct pthread_mutex_t {
        __pthread_mutex_flag1: u16,
        __pthread_mutex_flag2: u8,
        __pthread_mutex_ceiling: u8,
        __pthread_mutex_type: u16,
        __pthread_mutex_magic: u16,
        __pthread_mutex_lock: u64,
        __pthread_mutex_data: u64
    }
```

All of the fields Miri needs need to be initialized to 0. However, bytes 4-7 (where we store the Mutex ID) contains this __pthread_mutex_magic field that is not zero-initialized.

So we'll need to pick a different offset for the Mutex ID on solarish. The offset is currently duplicated in mutex_get_id and mutex_reset_id; the `4` there needs to become something else on solarish. It looks like `0` should work.

src/shims/unix/sync.rs
```diff
@@ -279,7 +279,7 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
     ) -> InterpResult<'tcx, i32> {
         let this = self.eval_context_mut();
 
-        if !matches!(&*this.tcx.sess.target.os, "linux" | "macos") {
+        if !matches!(&*this.tcx.sess.target.os, "linux" | "macos" | "solaris" | "illumos") {
```
RalfJung, line 282: If you enable all these things, you definitely need to also add the `pthread-sync` test to CI to ensure they actually work.

devnexen:
cool with your suggestion changes vec tests pass now.

src/shims/unix/sync.rs
```diff
@@ -16,6 +16,11 @@ use crate::*;
 /// in `pthread_mutexattr_settype` function.
 const PTHREAD_MUTEX_NORMAL_FLAG: i32 = 0x8000000;
 
+#[cfg(not(target_os = "macos"))]
+const PTHREAD_MUTEX_ID: u64 = 0;
+#[cfg(target_os = "macos")]
+const PTHREAD_MUTEX_ID: u64 = 4;
```
RalfJung, line 22: I'm afraid this will not work. You are testing the host OS Miri is running on, but you need to test the target OS that the interpreted program is running for. These are entirely independent.

You'll have to define a function, `mutex_id_offset`, that takes as argument an `ecx: &MiriInterpCx<'mir, 'tcx>` and checks the `ecx.tcx.sess.target.os` to determine the offset.

src/shims/unix/sync.rs
```diff
@@ -80,10 +85,10 @@ fn mutex_reset_id<'mir, 'tcx: 'mir>(
 ) -> InterpResult<'tcx, ()> {
     ecx.deref_pointer_and_write(
         mutex_op,
-        4,
+        PTHREAD_MUTEX_ID,
         Scalar::from_i32(0),
         ecx.libc_ty_layout("pthread_mutex_t"),
-        ecx.machine.layouts.u32,
+        ecx.machine.layouts.i32,
```
RalfJung, line 91: No this should be u32, we interpret this as an unsigned number. It doesn't really matter what the libc type does here.

devnexen: sorry I had a crash without this change but then now it works ... will roll back

README.md
```diff
@@ -227,6 +227,7 @@ degree documented below):
 - We have unofficial support (not maintained by the Miri team itself) for some further operating systems.
   - `freebsd`: **maintainer wanted**. Supports `std::env` and parts of `std::{thread, fs}`, but not `std::sync`.
   - `android`: **maintainer wanted**. Support very incomplete, but a basic "hello world" works.
+  - `solaris/illumos`: @devnexen. Support very incomplete, but a basic "hello world" works.
```
RalfJung, line 230:
```suggestion
  - `solaris/illumos`: maintained by @devnexen. Support very incomplete, but a basic "hello world" works.
```

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -0,0 +1,47 @@
...
+    ) -> InterpResult<'tcx, EmulateForeignItemResult> {
+        let this = self.eval_context_mut();
+        match link_name.as_str() {
+            // errno
+            "__errno" => {
+                let [] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let errno_place = this.last_error_place()?;
+                this.write_scalar(errno_place.to_ref(this).to_scalar(), dest)?;
+            }
+            // Threading
```
RalfJung, line 29:
```suggestion

            // Threading
```
just to match our usual style of having empty lines between the "groups" of shims

tests/pass-dep/shims/libc-misc.rs
RalfJung, line 1: If you are keeping this diff, please also add libc-misc to CI.

But probably it's better to drop this diff for now, and focus on the core that's needed to get started.

src/shims/unix/sync.rs
```diff
@@ -704,7 +724,7 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
     ) -> InterpResult<'tcx, i32> {
         let this = self.eval_context_mut();
 
-        if !matches!(&*this.tcx.sess.target.os, "linux" | "macos") {
+        if !matches!(&*this.tcx.sess.target.os, "linux" | "macos" | "solaris" | "illumos") {
```
RalfJung, line 727: You'll need to do the same in pthread_condattr_setclock, pthread_condattr_getclock.

(Also I just realized the function name there is wrong... while at it, could you fix it?)

RalfJung:
Almost there. :)  There's something wrong with the signature of pthread_join... I assume `pthread_t` has a different size on this platform than on others. We usually assume it is `usize`, but on Solarish it is `u32`.

So here

https://github.com/rust-lang/miri/blob/af37aca6c05e13afa0acfbed53bda24dea8b7526/src/shims/unix/thread.rs#L45

you need to us something like this instead
```rust
let thread_id = this.read_scalar(thread)?.to_int(this.libc_ty_layout("pthread_t").size)?;
```

RalfJung:
Also, is it possible there's something wrong with your system clock? Your commits show up as coming 20-50minutes from the future...^^

ci/ci.sh
```diff
@@ -146,6 +146,7 @@ case $HOST_TARGET in
     MIRI_TEST_TARGET=x86_64-unknown-freebsd run_tests_minimal $BASIC panic/panic concurrency/simple atomic threadname libc-getentropy libc-getrandom libc-misc fs env num_cpus
     MIRI_TEST_TARGET=i686-unknown-freebsd   run_tests_minimal $BASIC panic/panic concurrency/simple atomic threadname libc-getentropy libc-getrandom libc-misc fs env num_cpus
     MIRI_TEST_TARGET=aarch64-linux-android  run_tests_minimal $VERY_BASIC hello panic/panic
+    MIRI_TEST_TARGET=x86_64-unknown-illumos run_tests_minimal $VERY_BASIC hello panic/panic pthread-sync
```
RalfJung, line 149: Should we also run tests with `x86_64-pc-solaris`? You've added two OSes everywhere but we're only testing one of them.

README.md
```diff
@@ -227,6 +227,7 @@ degree documented below):
 - We have unofficial support (not maintained by the Miri team itself) for some further operating systems.
   - `freebsd`: **maintainer wanted**. Supports `std::env` and parts of `std::{thread, fs}`, but not `std::sync`.
   - `android`: **maintainer wanted**. Support very incomplete, but a basic "hello world" works.
+  - `solaris/illumos`: maintained by @devnexen. Support very incomplete, but a basic "hello world" works.
```
RalfJung, line 230:
```suggestion
  - `solaris`/`illumos`: maintained by @devnexen. Support very incomplete, but a basic "hello world" works.
```

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -0,0 +1,48 @@
...
+        &mut self,
+        link_name: Symbol,
+        abi: Abi,
+        args: &[OpTy<'tcx, Provenance>],
+        dest: &MPlaceTy<'tcx, Provenance>,
+    ) -> InterpResult<'tcx, EmulateItemResult> {
+        let this = self.eval_context_mut();
+        match link_name.as_str() {
+            // errno
+            "__errno" => {
```
RalfJung, line 24: Is `__errno` (with two `_` right)? libc and std seem to use triple-underscore `___errno`.

If this is dead code (i.e., not needed for these tests), then please remove it entirely.

src/concurrency/thread.rs
```diff
@@ -90,6 +97,12 @@ impl From<ThreadId> for u64 {
     }
 }
 
+impl From<ThreadId> for i128 {
+    fn from(t: ThreadId) -> Self {
+        t.0.into()
+    }
+}
```
RalfJung, line 104: This is not used, is it?

RalfJung (commented):
Thanks! This looks good, apart from a few nits. Why is it still marked WIP? :)

RalfJung:
Ah, Solaris in `std` has support for the stack guard, which seems to be disabled on illumos. That calls a bunch of very platform-specific functions (see [this file](https://github.com/rust-lang/rust/blob/5b9d7ab558fe6ee53bd40e16a20c25716f3c9e56/library/std/src/sys/pal/unix/stack_overflow.rs)).

This can be quite the rabbit hole of little functions that need to be added -- usually with [stubs like this](https://github.com/rust-lang/miri/blob/100c718ee7bf479df7763f78b4c268a05a109333/src/shims/unix/linux/foreign_items.rs#L190-L196) that don't actually do anything.

It's up to you whether you want to add that, or restrict the PR to illumos.

RalfJung:
You will probably have to extend this hack to also allow solaris, not just macos:

https://github.com/rust-lang/miri/blob/10e8bb87fe416bc190426789032594aa1aa311f2/src/shims/unix/mem.rs#L47-L49

README.md
```diff
@@ -227,6 +227,7 @@ degree documented below):
 - We have unofficial support (not maintained by the Miri team itself) for some further operating systems.
   - `freebsd`: **maintainer wanted**. Supports `std::env` and parts of `std::{thread, fs}`, but not `std::sync`.
   - `android`: **maintainer wanted**. Support very incomplete, but a basic "hello world" works.
+  - `solaris` / `illumos`: maintained by @devnexen. Support very incomplete, but a basic "hello world" works.
```
RalfJung, line 230:
```suggestion
  - `illumos`: maintained by @devnexen. Support very incomplete, but a basic "hello world" works.
```

RalfJung:
Great, thanks!
@bors r+

PR #3205, opened 2023-12-03
also test simd_select_bitmask on arrays for less than 8 elements

RalfJung:


RalfJung:
@workingjubilee @calebzulawski this code is causing a Miri UB error when run on a big-endian target:
```rust
    let bitmask = u8x4::from_array([0b00001101, 0, 0, 0]);
    assert_eq!(
        mask32x4::from_bitmask_vector(bitmask),
        mask32x4::from_array([true, false, true, true]),
    );
```
It seems to be calling `simd_select_bitmask` with `10110000`.  My interpretation was that even on big-endian, it is the N *least* significant bits that carry the mask when N<8, so this is wrong?

This is basically https://github.com/rust-lang/portable-simd/pull/267 again, but for simd_select_bitmask rather than simd_bitmask.

RalfJung:
Filed an issue: https://github.com/rust-lang/portable-simd/issues/379

I think what Miri does here is right, or at least it is consistent with simd_bitmask and with https://github.com/rust-lang/portable-simd/pull/267. So I'll land this.

@bors r+

PR #3212, opened 2023-12-05
Don't explicitly warn against `semicolon_in_expressions_from_macros`

Xanewok:
This warns-by-default since 2 years and already has been added to the future-incompat group since Rust 1.68.

See https://github.com/rust-lang/rust/issues/79813 for the tracking issue.

Just something I noticed when trying to wrap my head around lints warned against in the rust-lang/rust and noticed it's not warned against anymore; let me know if the change makes sense :pray:

### DIFF
--- a/miri-script/src/util.rs
+++ b/miri-script/src/util.rs
@@ -73,7 +73,9 @@ impl MiriEnv {
             flags.push("-C link-args=-Wl,-rpath,");
             flags.push(libdir);
             // Enable rustc-specific lints (ignored without `-Zunstable-options`).
-            flags.push(" -Zunstable-options -Wrustc::internal -Wrust_2018_idioms -Wunused_lifetimes");
+            flags.push(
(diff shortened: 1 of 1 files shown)


RalfJung:
Thanks for the PR!  Funny that it insists on formatting only after the line got shorter... that makes no sense, does it? Looks like a rustfmt bug to me.
@bors r+

This is taken [directly from bootstrap](https://github.com/rust-lang/rust/blob/b7e6da80af0d316b966940368abf073c89abdce9/src/bootstrap/src/core/builder.rs#L1873-L1875), I guess they don't need this flag any more either then?

Xanewok:
It was weird, indeed! Maybe it only checks for the modified lines? In any case, didn't want to wrestle Rustfmt here :smile: 

I've also opened https://github.com/rust-lang/rust/pull/118642 to remove the warning argument in bootstrap.

RalfJung:
No, rustfmt will always check the entire file. :shrug:

PR #3213, opened 2023-12-06
remove unnecesary `-Zunstable-options`

weihanglo:
AFAIK `-Zunstable-options` is for `cargo --config` CLI, which was stabilized in 1.63

https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#config-cli


RalfJung:
Ah, nice. :)
@bors r+

PR #3214, opened 2023-12-07
Move some x86 intrinsics code to helper functions in `shims::x86`

eduardosm:
To make them reusable for intrinsics of other x86 features.

Splitted from https://github.com/rust-lang/miri/pull/3192

### DIFF
--- a/src/shims/x86/mod.rs
+++ b/src/shims/x86/mod.rs
@@ -616,6 +616,56 @@ fn horizontal_bin_op<'tcx>(
     Ok(())
 }
 
+/// Conditionally multiplies the packed floating-point elements in
+/// `left` and `right` using the high 4 bits in `imm`, sums the calculated
+/// products (up to 4), and conditionally stores the sum in `dest` using
+/// the low 4 bits of `imm`.
+fn conditional_dot_product<'tcx>(
+    this: &mut crate::MiriInterpCx<'_, 'tcx>,
+    left: &OpTy<'tcx, Provenance>,
+    right: &OpTy<'tcx, Provenance>,
+    imm: &OpTy<'tcx, Provenance>,
+    dest: &PlaceTy<'tcx, Provenance>,
+) -> InterpResult<'tcx, ()> {
+    let (left, left_len) = this.operand_to_simd(left)?;
+    let (right, right_len) = this.operand_to_simd(right)?;
+    let (dest, dest_len) = this.place_to_simd(dest)?;
+
+    assert_eq!(left_len, right_len);
+    assert!(dest_len <= 4);
+
+    let imm = this.read_scalar(imm)?.to_u8()?;
+
+    let element_layout = left.layout.field(this, 0);
+
+    // Calculate dot product
+    // Elements are floating point numbers, but we can use `from_int`
+    // because the representation of 0.0 is all zero bits.
+    let mut sum = ImmTy::from_int(0u8, element_layout);
+    for i in 0..left_len {
+        if imm & (1 << i.checked_add(4).unwrap()) != 0 {
+            let left = this.read_immediate(&this.project_index(&left, i)?)?;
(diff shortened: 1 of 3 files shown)


src/shims/x86/mod.rs
```diff
@@ -408,3 +615,84 @@ fn horizontal_bin_op<'tcx>(
 
     Ok(())
 }
+
+/// Conditionally multiplies the packed floating-point elements in
+/// `left` and `right` using the high 4 bits in `imm`, sums the four
+/// products, and conditionally stores the sum in `dest` using the low
```
RalfJung, line 621: Since this is "conditional", I assume these are not actually four products but up to four products?

src/shims/x86/mod.rs
```diff
@@ -408,3 +615,84 @@ fn horizontal_bin_op<'tcx>(
...
+    let mut all_zero = true;
+    let mut masked_set = true;
+    for i in 0..op_len {
+        let op = this.project_index(&op, i)?;
+        let mask = this.project_index(&mask, i)?;
+
+        let op = this.read_scalar(&op)?.to_uint(op.layout.size)?;
+        let mask = this.read_scalar(&mask)?.to_uint(mask.layout.size)?;
+        all_zero &= op & mask == 0;
+        masked_set &= op & mask == mask;
```
RalfJung, line 694:
```suggestion
        masked_set &= (op & mask) == mask;
```
I first read this as `op & (mask == mask)` and was quite confused...

src/shims/x86/mod.rs
```diff
@@ -408,3 +615,85 @@ fn horizontal_bin_op<'tcx>(
...
+        let op = this.project_index(&op, i)?;
+        let mask = this.project_index(&mask, i)?;
+
+        let op = this.read_scalar(&op)?.to_uint(op.layout.size)?;
+        let mask = this.read_scalar(&mask)?.to_uint(mask.layout.size)?;
+        all_zero &= (op & mask) == 0;
+        masked_set &= (op & mask) == mask;
+    }
+
+    Ok(f(all_zero, masked_set))
```
RalfJung, line 698: No that's not what I meant... what I mean was to do `acc = f(acc, op, mask)` inside the loop. That way the caller then decides how the results get accumulated.

Basically, this is folding in some way over the two vectors, and so the natural general interface is to allow any kind of folding.

eduardosm: > No that's not what I meant... what I mean was to do `acc = f(acc, op, mask)` inside the loop. That way the caller then decides how the results get accumulated.

That would a `fn(bool, Scalar, Scalar) -> bool`, wouldn't it? Not a `fn(bool, bool) -> bool`.

RalfJung: Ah yeah -- I might have gotten the signature wrong.

Basically, whatever the type of  `f` is in the current code. Or maybe something a bit more general.

src/shims/x86/sse41.rs
```diff
@@ -257,112 +222,34 @@ pub(super) trait EvalContextExt<'mir, 'tcx: 'mir>:
...
+                            |acc, op, mask| {
+                                let op = op.to_scalar().to_uint(op.layout.size)?;
+                                let mask = mask.to_scalar().to_uint(mask.layout.size)?;
+                                Ok(acc && (op & mask) == mask)
+                            },
+                        "ptestnzc" =>
+                            |acc, op, mask| {
+                                let op = op.to_scalar().to_uint(op.layout.size)?;
+                                let mask = mask.to_scalar().to_uint(mask.layout.size)?;
+                                Ok(acc && (op & mask) != 0 && (op & mask) != mask)
```
RalfJung, line 243: I think the code will be overall shorter if you move the match down inside the closure; then the `let op` and `let mask` can be outside the `match`.

RalfJung:
Thanks a lot. :)
@bors r+

PR #3216, opened 2023-12-08
Fix x86 SSE4.1 ptestnzc

eduardosm:
Fixes ptestnzc by bringing back the original implementation of https://github.com/rust-lang/miri/pull/3214.

`(op & mask) != 0 && (op & mask) == !ask` need to be calculated for the whole vector. It cannot be calculated for each element and then folded.

For example, given
* `op = [0b100, 0b010]`
* `mask = [0b100, 0b110]`

The correct result would be:
* `op & mask = [0b100, 0b010]`
Comparisons are done on the vector as a whole:
* `all_zero = (op & mask) == [0, 0] = false`
* `masked_set = (op & mask) == mask = false`
* `!all_zero && !masked_set = true` correct result

The previous method:
* `op & mask = [0b100, 0b010]`
Comparisons are done element-wise:
* `all_zero = (op & mask) == [0, 0] = [true, true]`
* `masked_set = (op & mask) == mask = [true, false]`
* `!all_zero && !masked_set = [true, false]`
 After folding with AND, the final result would be `false`, which is incorrect.

### DIFF
--- a/src/shims/x86/mod.rs
+++ b/src/shims/x86/mod.rs
@@ -666,30 +666,33 @@ fn conditional_dot_product<'tcx>(
     Ok(())
 }
 
-/// Folds SIMD vectors `lhs` and `rhs` into a value of type `T` using `f`.
-fn bin_op_folded<'tcx, T>(
+/// Calculates two booleans.
+///
+/// The first is true when all the bits of `op & mask` are zero.
+/// The second is true when `(op & mask) == mask`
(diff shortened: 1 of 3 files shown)


RalfJung:
Ah... unfortunate. I guess we could still express this as a fold if we wanted (by folding over a pair of booleans), but it doesn't seem worth it.

@bors r+

Issue #3218, opened 2023-12-10
mmap failure returning 0x00 instead of MAP_FAILED


gimbling-away:
Upon failure, mmap is supposed to return MAP_FAILED, not nullptr.
https://github.com/rust-lang/miri/blob/12c868ad506dde09a0f842e747e03a632dcb65a5/src/shims/unix/mem.rs#L75
Currently mmap returns null pointer in miri.

Is this intentional or accidental?

PR #3219, opened 2023-12-10
Return MAP_FAILED when mmap fails

saethlin:
I don't properly remember why we ended up with a hodgepodge of return values, but https://github.com/rust-lang/miri/issues/3218 correctly points out that we are supposed to return `MAP_FAILED`. This should fix that return value and also add sufficient tests to prevent making a similar mistake.


RalfJung:
Thanks, yes this makes a lot more sense. :)
@bors r+

Issue #3041, opened 2023-08-28
Unix: `munmap` emits "warning: integer-to-pointer cast"


morrisonlevi:
On both Linux and Mac, I get this warning for `munmap`:

```
test tests::it_works ... warning: integer-to-pointer cast
  --> src/lib.rs:39:15
   |
39 |     let foo = munmap(addr, len);
   |               ^^^^^^^^^^^^^^^^^ integer-to-pointer cast
   |
   = help: This program is using integer-to-pointer casts or (equivalently) `ptr::from_exposed_addr`,
   = help: which means that Miri might miss pointer bugs in this program.
   = help: See https://doc.rust-lang.org/nightly/std/ptr/fn.from_exposed_addr.html for more details on that operation.
   = help: To ensure that Miri does not miss bugs in your program, use Strict Provenance APIs (https://doc.rust-lang.org/nightly/std/ptr/index.html#strict-provenance, https://crates.io/crates/sptr) instead.
   = help: You can then pass the `-Zmiri-strict-provenance` flag to Miri, to ensure you are not relying on `from_exposed_addr` semantics.
   = help: Alternatively, the `-Zmiri-permissive-provenance` flag disables this warning.
   = note: BACKTRACE:
   = note: inside `alloc_dealloc` at src/lib.rs:39:15: 39:32
```

I think the [miri munmap shim](https://github.com/rust-lang/miri/blob/09704219d635bf3457da371a4186709723cb5e51/src/shims/unix/mem.rs#L109-L158) itself is triggering this warning. It looks like it turns the pointer into a usize on line 116 so it can check that it's aligned, and then turns it back into a pointer on 128 instead of using the original pointer.

For a quick reproduction, the code below passes the address directly returned from `mmap` back into `munmap`, without doing any integer-to-pointer casts:

```rs
use libc::*;
use std::io;

pub unsafe fn page_size() -> io::Result<size_t> {
    let page_size = sysconf(_SC_PAGESIZE);
    if page_size == -1 {
        return Err(io::Error::last_os_error());
    }
    match size_t::try_from(page_size) {
        Ok(len) => {
            if len.is_power_of_two() {
                Ok(len)
            } else {
                return Err(io::Error::new(
                    io::ErrorKind::Other,
                    "page size must be a power of two",
                ));
            }
        }
        Err(err) => return Err(io::Error::new(io::ErrorKind::Other, err)),
    }
}

pub unsafe fn alloc_dealloc() -> io::Result<()> {
[...]

PR #3220, opened 2023-12-10
Make mmap not use expose semantics

saethlin:
Fixes https://github.com/rust-lang/miri/issues/3041 per https://github.com/rust-lang/miri/issues/3041#issuecomment-1711235196


src/shims/unix/mem.rs
```diff
@@ -113,21 +111,19 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
             this.set_last_error(Scalar::from_i32(this.eval_libc_i32("EINVAL")))?;
             return Ok(Scalar::from_i32(-1));
         }
 
         let length = round_to_next_multiple_of(length, this.machine.page_size);
 
-        let ptr = Machine::ptr_from_addr_cast(this, addr)?;
-
-        let Ok(ptr) = ptr.into_pointer_or_addr() else {
+        let Ok(ptr) = addr.into_pointer_or_addr() else {
```
RalfJung, line 126: `addr` is already a `Pointer`. Why is this still needed?

Instead of `Machine::ptr_get_alloc`, you should call `self.ptr_get_alloc_id`.

saethlin: Actually I think we don't even want to do that much. I think we want to report UB in all the cases where we threw an unsupported error, because LLVM's semantics do make these UB.

RalfJung: You need to get the alloc_id though and that is the method to get it.

RalfJung: Oh, `deallocate_ptr` already does that. Yeah that also works.

src/shims/unix/mem.rs
```diff
@@ -113,43 +111,20 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
-        let (_kind, alloc) = this.memory.alloc_map().get(alloc_id).unwrap();
-        if offset != Size::ZERO || alloc.len() as u64 != length {
-            throw_unsup_format!(
-                "Miri only supports munmap calls that exactly unmap a region previously returned by mmap"
-            );
-        }
-
-        let len = Size::from_bytes(alloc.len() as u64);
+        let length = Size::from_bytes(round_to_next_multiple_of(length, this.machine.page_size));
         this.deallocate_ptr(
```
RalfJung, line 125: Given that there was quite a bit of discussion that led to this implementation, please add a comment explaining that.

saethlin: Good point. I've updated the comment at the top of the module, and also nuked the test that was checking for this. Does that cover your request, or should there be more explanation in `munmap` itself?

saethlin: I force-pushed to fix the failing tests and didn't comment. So I'm commenting now.

RalfJung:
Looks good, thanks. :)
@bors r+

PR #3226, opened 2023-12-14
Add a regression test for rust#115145

saethlin:
Per https://github.com/rust-lang/rust/pull/118805#issuecomment-1855264860


tests/pass/async-fn.rs
```diff
@@ -58,6 +58,20 @@ async fn hello_world() {
     read_exact(&mut reader, &mut marker).await.unwrap();
 }
 
+async fn uninhabited_variant() {
```
RalfJung, line 61: Please add a link to the issue where this came up.

Then r=me.

Issue #3228, opened 2023-12-16
Provenance GC causes dead_alloc_map ICE when a protector is the last mention of an AllocId


saethlin:
This program will crash Miri:
```rust
extern "Rust" {
    pub fn miri_run_provenance_gc();
}

fn oof(mut b: Box<u8>) {
    drop(b);
    b = Box::new(0u8);
    unsafe { miri_run_provenance_gc() };
}

fn main() {
    oof(Box::new(0u8));
}
```

We create a `Box`, and immediately pass it to a function which makes it weakly protected.
This function argument is the only memory with the relevant AllocId.
Then we deallocate the `Box`, and at the same time overwrite the function argument.
_One GC run later_

We return from the function and execute this code to check if the allocation is still alive and :boom:
https://github.com/rust-lang/miri/blob/e6154269ef0028c7bcecb2dce93bd35da5d7cf26/src/borrow_tracker/mod.rs#L378-L384

I'm not entirely sure what the best way to fix this is. The obvious way would be to visit the AllocIds in `borrow_tracker::FrameExtra` in the GC. But the comments we left around there seemed pretty confident that wouldn't be required, so I'm not totally sure there isn't another sort of design flaw here.

PR #3229, opened 2023-12-16
Visit the AllocIds and BorTags in borrow state FrameExtra

saethlin:
Fixes https://github.com/rust-lang/miri/issues/3228

I said

> The obvious way would be to visit the AllocIds in borrow_tracker::FrameExtra in the GC.

Since I have had no new ideas, that's what this does.

### DIFF
--- a/src/borrow_tracker/mod.rs
+++ b/src/borrow_tracker/mod.rs
@@ -65,7 +65,7 @@ pub struct FrameState {
     /// incremental updates of the global list of protected tags stored in the
     /// `stacked_borrows::GlobalState` upon function return, and if we attempt to pop a protected
     /// tag, to identify which call is responsible for protecting the tag.
-    /// See `Stack::item_popped` for more explanation.
+    /// See `Stack::item_invalidated` for more explanation.
     /// Tree Borrows also needs to know which allocation these tags
     /// belong to so that it can perform a read through them immediately before
     /// the frame gets popped.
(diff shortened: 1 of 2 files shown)


src/borrow_tracker/mod.rs
```diff
@@ -76,8 +76,10 @@ pub struct FrameState {
 }
 
 impl VisitProvenance for FrameState {
-    fn visit_provenance(&self, _visit: &mut VisitWith<'_>) {
-        // `protected_tags` are already recorded by `GlobalStateInner`.
+    fn visit_provenance(&self, visit: &mut VisitWith<'_>) {
+        for (id, tag) in &self.protected_tags {
```
RalfJung, line 80: Every tag in here should also be in `GlobalStateInner::protected_tags`. That's why we had the comment.

However it looks like that's only true for the tags, not the Alloc ID. Good catch!
But this should mean we can stop traversing `GlobalStateInner::protected_tags`, since we are now visiting both the tags and alloc IDs here in `FrameState`.

tests/pass/protector-gc.rs
RalfJung, line 1: Please add a link to the issue that explains why this is tested.

src/borrow_tracker/mod.rs
saethlin: Yup, I think I did that?

tests/pass/protector-gc.rs
saethlin: Done

PR #3233, opened 2023-12-21
Add portable-atomic-util bug to "bugs found" list

RalfJung:
At least, reading https://notgull.net/cautionary-unsafe-tale/ it seems fair to say Miri found this bug. @notgull please let me know if you are okay with having this listed here.


taiki-e:
It is not portable-atomic bug, and it is portable-atomic-util bug. They are not the same crate and have different actual authors.

RalfJung:
Ah, sorry for that. Fixed.

PR #3199, opened 2023-12-02
Fix miri script target dir and update doc link

Jefffrey:
Fix ui_test doc link in `CONTRIBUTING.md`

Explicitly set `--target-dir` in `miri` script when building to expected location

- I have `build.target-dir` cargo config set to be some other location so miri script was erroring out since compiled binary was not in expected location

PR #3236, opened 2023-12-22
accommodate CARGO_TARGET_DIR being set

Dante-Broggi:
Hello, when looking to locally build Miri, I found that setting CARGO_TARGET_DIR, which I set globally to centralize build directories, causes miri-script to fail, as it assumes cargo uses the default target dir and thus uses the contents of the "target" directory. I resolved this by explicitly setting the target-dir build option in Cargo.toml, which is preferred over CARGO_TARGET_DIR.

It may be preferable to query where cargo is putting build products and use that location in miri-script, but that would require modifying miri-script itself to query cargo, which seems unnecessary.

### DIFF
--- a/miri-script/Cargo.toml
+++ b/miri-script/Cargo.toml
@@ -21,3 +21,6 @@ xshell = "0.2"
 rustc_version = "0.4"
 dunce = "1.0.4"
 directories = "5"
+
+[build]
+target-dir = "target"


RalfJung:
Didn't https://github.com/rust-lang/miri/pull/3199 fix this?

RalfJung:
@Dante-Broggi it is unclear what you are trying to fix with this PR. As far as I know, https://github.com/rust-lang/miri/pull/3199 fixed the issues we had with CARGO_TARGET_DIR around miri-script. Why is that existing patch not sufficient?

Dante-Broggi:
Sorry, I forgot about this, the other commit fixed this issue.

PR #3239, opened 2023-12-26
Use `windows-sys` in windows tests

beepster4096:
This PR adds `windows-sys` to `test_dependencies` so that we don't have to write out windows api bindings for each test.


tests/pass-dep/concurrency/windows_condvar_shared.rs
```diff
@@ -2,42 +2,29 @@
...
-
-const CONDITION_VARIABLE_LOCKMODE_SHARED: u32 = 1;
-const INFINITE: u32 = u32::MAX;
-
 /// threads should be able to reacquire the lock while it is locked by multiple other threads in shared mode
 fn all_shared() {
     println!("all_shared");
 
-    let mut lock = null_mut();
-    let mut condvar = null_mut();
```
RalfJung, line 37: You also changed this from null ptrs to pointers to some actual memory. Why that?

RalfJung (commented):
This is great, thanks! I got one question though, see the review comment.

tests/pass-dep/concurrency/windows_condvar_shared.rs
beepster4096: They're pointer sized structures. In the original tests, I just made them pointers for simplicity. `windows-sys` uses separate types for them, so I thought this was the simplest way to zero initialize them. `mem::zeroed` would also work here, but requires more unsafe blocks.

beepster4096: Oh and `SRWLOCK/CONDITION_VARIABLE/INIT_ONCE { Ptr: null_mut() }` also works too, but is bigger looking and requires more imports.

RalfJung: Ah I see, the tests previously used imprecise types.

According to [these docs](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-sleepconditionvariablesrw) what we do here is anyway questionable...
"This variable must be initialized using the InitializeConditionVariable function."

I guess Miri doesn't currently check that the variable is initialized at all. But that's a pre-existing issue, nothing to do with this PR.

tests/pass-dep/concurrency/windows_condvar_shared.rs
```diff
@@ -2,42 +2,29 @@
...
+    let mut lock = MaybeUninit::zeroed();
```
RalfJung, line 23: I think it'd be better to indicate the type of what we are allocating here. Type inference is great but it also makes this code pretty hard to follow if one dos not know the Windows API types by heart.

This applies everywhere in this PR.

test_dependencies/Cargo.lock
```diff
@@ -84,9 +84,9 @@ dependencies = [
 
 [[package]]
 name = "getrandom"
-version = "0.2.10"
+version = "0.2.11"
```
RalfJung, line 87: Please avoid unrelated lockfile changes.

tests/pass-dep/concurrency/windows_condvar_shared.rs
beepster4096: There are constants to statically initialize these types mentioned in [these docs](https://learn.microsoft.com/en-us/windows/win32/sync/slim-reader-writer--srw--locks)
> The caller must allocate an SRWLOCK structure and initialize it by either calling [InitializeSRWLock](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-initializesrwlock) (to initialize the structure dynamically) or assign the constant SRWLOCK_INIT to the structure variable (to initialize the structure statically).

`SRWLOCK_INIT`, `CONDITION_VARIABLE_INIT` and `INIT_ONCE_STATIC_INIT` are all equal to `{0}`. They aren't in `windows-sys` though.

RalfJung:
Looks good, thanks :)
@bors r+

RalfJung:
looks like macos being macos...
@bors retry

PR #3240, opened 2023-12-26
Support for tempfile crate on UNIX hosts

Jefffrey:
Reviving old PR: https://github.com/rust-lang/miri/pull/2720

Attempted to apply the changes as suggested by https://github.com/rust-lang/miri/pull/2720#issuecomment-1347867264

To fix tempfile to work for UNIX targets only and fall back to previous behaviour of only supporting default mode for Windows targets

### DIFF
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -559,8 +559,21 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 );
             };
 
-            if mode != 0o666 {
-                throw_unsup_format!("non-default mode 0o{:o} is not supported", mode);
+            #[cfg(unix)]
+            {
+                // Support all modes on UNIX host
+                use std::os::unix::fs::OpenOptionsExt;
+                options.mode(mode);
+            }
+            #[cfg(not(unix))]
+            {
+                // Only support default mode for non-UNIX (i.e. Windows) host
+                if mode != 0o666 {
+                    throw_unsup_format!(
+                        "non-default mode 0o{:o} is not supported on non-Unix hosts",
+                        mode
+                    );
+                }
             }
 
             mirror |= o_creat;
--- a/test_dependencies/Cargo.lock
+++ b/test_dependencies/Cargo.lock
@@ -45,6 +45,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
+name = "bitflags"
+version = "2.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "327762f6e5a765692301e5bb513e0d9fef63be86bbc14528052b1cd3e6f03e07"
+
+[[package]]
 name = "bumpalo"
 version = "3.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


tests/pass-dep/tempfile.rs
```diff
@@ -0,0 +1,20 @@
+//@ignore-target-windows: no libc on Windows
```
RalfJung, line 1: This comment seems wrong, the test doesn't use the libc.
```suggestion
//@ignore-target-windows: File handling is not implemented yet
```

test_dependencies/Cargo.lock
```diff
@@ -352,6 +389,20 @@ dependencies = [
  "unicode-ident",
 ]
 
+[[package]]
+name = "tempfile"
+version = "3.3.0"
```
RalfJung, line 394: Any reason you're not adding the latest version, 3.8.1?

src/shims/unix/fs.rs
```diff
@@ -559,8 +559,14 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
-            if mode != 0o666 {
-                throw_unsup_format!("non-default mode 0o{:o} is not supported", mode);
+            match this.tcx.sess.target.os.as_ref() {
+                // Support all modes on UNIX host
+                "linux" | "freebsd" | "android" | "macos" => {}
+                // Non-UNIX (aka Windows) only support default mode
+                _ =>
+                    if mode != 0o666 {
+                        throw_unsup_format!("non-default mode 0o{:o} is not supported", mode);
+                    },
```
RalfJung, line 569: Now you are just completely ignoring the mode. This is *not* what was suggested [here](https://github.com/rust-lang/miri/pull/2720#issuecomment-1347867264). Instead you should
- support all modes on a Unix host, just forwarding the flag
- throw_unsup on a Windows host for non-default flags

src/shims/unix/fs.rs
```diff
@@ -559,8 +559,14 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 );
             };
 
-            if mode != 0o666 {
-                throw_unsup_format!("non-default mode 0o{:o} is not supported", mode);
+            match this.tcx.sess.target.os.as_ref() {
+                // Support all modes on UNIX host
```
RalfJung, line 563: Comment and code do not match. The comment tests whether the *target* is unix or Windows. It will always be Unix here.

You need to check the *host*, via `#[cfg]`/`cfg!`.

test_dependencies/Cargo.lock
Jefffrey: Came through as part of the rebase, I've bumped the version now

src/shims/unix/fs.rs
Jefffrey: My mistake, I confused the host with the target. Hopefully fixed now, and also setting the mode on `options` too

src/shims/unix/fs.rs
```diff
@@ -559,8 +559,18 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+            {
+                // Support all modes on UNIX host
+                use std::os::unix::fs::OpenOptionsExt;
+                options.mode(mode);
+            }
+            #[cfg(not(unix))]
+            {
+                // Only support default mode for non-UNIX (i.e. Windows) host
+                if mode != 0o666 {
+                    throw_unsup_format!("non-default mode 0o{:o} is not supported", mode);
```
RalfJung, line 572:
```suggestion
                    throw_unsup_format!("non-default mode 0o{:o} is not supported on non-Unix hosts", mode);
```

RalfJung:
Looks good, thanks!

Could you please squash this into a single commit? (Or two commits, one by Pointerbender and one by you)

Jefffrey:
Squashed to single commit

RalfJung:
Thanks. :)

@bors r+

PR #3246, opened 2023-12-28
Fix integer overflow ICEs from round_up_to_next_multiple_of

saethlin:
Turns out the standard library _does_ have the function we need. So I swapped us to using the checked version in mmap/munmap where we can return an error, and we're still using the ICEy version in SIMD.

I found one of the ICE cases by running this test: https://github.com/wbcchsyn/rust-mmap-allocator/blob/765bcaab6e3bfd1cb1e6eaac80ac7e821fb5979b/src/mmap_allocator.rs#L195-L210


saethlin:
Hunh. Tests passed on s390x-unknown-linux-gnu. Odd...

RalfJung:
Ah nice :)
@bors r+

PR #3256, opened 2024-01-06
implement the rounding intrinsics using apfloat rounding

RalfJung:
No reason to use host floats for these. Also merge two files that were both testing various float things.


RalfJung:
@bors retry
More text since the email is otherwise discarded

PR #3257, opened 2024-01-06
./miri build: also build tests, to avoid rebuilds later

RalfJung:


RalfJung:
@bors retry
More text since the email is otherwise sometimes discarded

PR #3260, opened 2024-01-07
Check if tier 2 targets build in the nightly cron job

saethlin:
This PR adds a CI job that only runs nightly which will install Miri built from the latest commit, and try to build every Tier 2 without host tools target, as documented on https://doc.rust-lang.org/nightly/rustc/platform-support.html. 

I'm not really excited about the idea of scraping the tier 2 without host tools list, but also keeping the list up-to-date by hand seems prone to forgetting to update it. And that update seems like the sort of manual maintenance we should automate.


.github/workflows/ci.yml
```diff
@@ -128,11 +128,53 @@ jobs:
       - name: rustdoc
         run: RUSTDOCFLAGS="-Dwarnings" ./miri cargo doc --document-private-items
 
+  sysroots:
```
RalfJung, line 131: Would it make sense to use a separate workflow (in a separate yml file) for this? It doesn't seem to share anything with the main workflow except the time of the cronjob.

ci/build-all-targets.sh
RalfJung, line 1: If we have a `ci/` folder, we should move `ci.sh` into it.

cargo-miri/Cargo.lock
```diff
@@ -212,9 +212,9 @@ dependencies = [
 
 [[package]]
 name = "rustc-build-sysroot"
-version = "0.4.2"
+version = "0.4.3"
```
RalfJung, line 215: I released 0.4.4 just a few minutes later since clippy pointed out a potential issue. Can you update to that?

ci/scrape-targets.py
```diff
@@ -0,0 +1,16 @@
+from bs4 import BeautifulSoup
+import requests
+
+url = "https://doc.rust-lang.org/nightly/rustc/platform-support.html"
```
RalfJung, line 4: Is there a way we can get this from the installed rustc rather than an online source?

ci/scrape-targets.py
```diff
@@ -0,0 +1,16 @@
...
+body = requests.get(url).text
+soup = BeautifulSoup(body, features="html.parser")
+# The tables are:
+# Tier 1
+# Tier 2 with host tools
+# Tier 2 without host tools <-- we want this one
```
RalfJung, line 10: Why only tier 2 without host tools, and not also tier 2 with host tools?

ci/scrape-targets.py
saethlin: It looks like we can install the docs and scrape a local HTML file? Trying this, it's a bit fiddly to test locally because `./miri toolchain -c rust-docs` doesn't add them to the toolchain, but if the `miri` toolchain doesn't exist yet it looks like the component is installed.

ci/scrape-targets.py
saethlin: :thinking: At one point I was thinking that the "with host tools" part of the build would ensure that the sysroot builds, but I suppose if we never test the host tools, they might not build.

.github/workflows/ci.yml
saethlin: I tested this PR in my fork and ironed out a few issues. Let me see if I can cut-and-paste without breaking this...

saethlin: I don't see any evidence that GitHub has picked up the new workflow. But it seems to have been picked up in my own repo? https://github.com/saethlin/miri/pull/1

ci/scrape-targets.py
RalfJung: Ah... well if it's complicated we can also stick to the URL and hope we don't get spurious failures from that.

ci/scrape-targets.py
RalfJung: Also this doesn't imply that the sysroot builds in Miri.

ci/scrape-targets.py
```diff
@@ -0,0 +1,15 @@
+import sys
+from bs4 import BeautifulSoup
+
+html = open(sys.argv[1], 'r').read()
+soup = BeautifulSoup(html, features="html.parser")
+# The tables are:
+# Tier 1
```
RalfJung, line 7:
```suggestion
# Tier 1          <-- this is already checked by main CI, so we ignore it here
```

ci/scrape-targets.py
```diff
@@ -0,0 +1,15 @@
...
+# Tier 2 with host tools     <-- we want this one
+# Tier 2 without host tools  <-- and also this
+# Tier 3
+for table in soup.find_all("table")[2:4]
```
RalfJung, line 11: So `2:4` means "indices 1 and 2"? Or is there something else going on here that I don't understand?

.github/workflows/ci.yml
RalfJung: It may be the case that it only gets picked up once it exists in master.

ci/scrape-targets.py
RalfJung: > ./miri toolchain -c rust-docs doesn't add them to the toolchain

So this worked after all?

saethlin: The CI job works because there is no pre-existing `miri` toolchain when we run `./miri toolchain -c rust-docs`. It's functional, but awkward to test manually.

RalfJung: Ah that's what you mean.

I am not sure if RTIM supports adding components to an already installed toolchain.

ci/scrape-targets.py
saethlin: Nope, I simply typed the wrong numbers

.github/workflows/ci.yml
saethlin: That's what I thought too, but in my fork the workflow doesn't exist on master.

bjorn3: I believe for your own repo it will apply every workflow found on a local branch, but for a PR it will only apply those found on the target branch.

saethlin: If you click the link you'll see the sysroots job running in a PR, no?

RalfJung: You are the owner of that repo, so your PRs are not regular PRs.

RalfJung:
LGTM, r=me after squashing.

.github/workflows/ci.yml
saethlin: I did not know that. That is very annoying.

ci/scrape-targets.py
bjorn3: Apart from scraping platform-support.html you could also look in https://static.rust-lang.org/dist/channel-rust-nightly.toml to find all targets for which we ship libstd (all tier 1 and tier 2 targets) In case of a rustup installed toolchain it is available locally under `$(rustc --print sysroot)/lib/rustlib/multirust-channel-manifest.toml`. This doesn't distinguish between tier 1 and tier 2 though, but you could distinguish between with and without host tools by looking if rustc is available for the target or not.

RalfJung: Or we just test tier 1 as well, it should only take a few extra minutes.

That would probably be more reliable than scraping html, yeah. Not a blocker to landing the initial experiment though.

PR #3279, opened 2024-01-25
bump ui_test to 0.22

oli-obk:


tests/compiletest.rs
```diff
@@ -86,17 +85,20 @@ fn test_config(target: &str, path: &str, mode: Mode, with_dependencies: bool) ->
...
         program,
         out_dir: PathBuf::from(std::env::var_os("CARGO_TARGET_DIR").unwrap()).join("ui"),
-        edition: Some("2021".into()), // keep in sync with `./miri run`
         threads: std::env::var("MIRI_TEST_THREADS")
             .ok()
             .map(|threads| NonZeroUsize::new(threads.parse().unwrap()).unwrap()),
         ..Config::rustc(path)
     };
+    // keep in sync with `./miri run`
+    config.comment_defaults.base().edition = Some(Spanned::dummy("2021".into())).into();
```
RalfJung, line 96: It's a bit unclear what should be kept in sync.
```suggestion
    let edition = "2021"; // keep in sync with `./miri run`
    config.comment_defaults.base().edition = Some(Spanned::dummy(edition.into())).into();
```
But also... this is quite confusing, why do we need to `Spanned` everything?

tests/compiletest.rs
```diff
@@ -66,8 +67,6 @@ fn test_config(target: &str, path: &str, mode: Mode, with_dependencies: bool) ->
         }
     }
     program.args.push("-Zui-testing".into());
-    program.args.push("--target".into());
-    program.args.push(target.into());
```
RalfJung, line 70: From the CI errors, removing `--target` doesn't seem to be right.

oli-obk: yea ^^ looks like it

RalfJung:
@oli-obk what's the plan with this PR? Is ui_test 0.23 ready to be used by Miri?

oli-obk:
still working on this now and then, keeping it open is not useful

PR #3283, opened 2024-01-29
moving out sched_getaffinity interception from linux'shim, FreeBSD su…

devnexen:
…pporting it too.


RalfJung:
I think it's the num_cpus test (tests/pass-dep/num_cpus.rs) that tests this function on Linux.

devnexen:
done :)

RalfJung:
Actually turns out this is still entirely untested on freeBSD. :(

RalfJung:
It's also not very useful, since all it does is return EINVAL.

PR #3285, opened 2024-01-30
Implement the `mmap64` foreign item.

sunfishcode:
`mmap64` is like `mmap` but uses a 64-bit integer instead of `off_t` for the offset parameter.

### DIFF
--- a/src/shims/unix/linux/foreign_items.rs
+++ b/src/shims/unix/linux/foreign_items.rs
@@ -44,6 +44,14 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 let result = this.linux_readdir64(dirp)?;
                 this.write_scalar(result, dest)?;
             }
+            "mmap64" => {
+                let [addr, length, prot, flags, fd, offset] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let offset = this.read_scalar(offset)?.to_i64()?;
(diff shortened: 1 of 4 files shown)


saethlin:
Can we avoid a copy+paste of all the mmap tests by doing something like `test_mmap(libc::mmap); test_mmap(libc::mmap64);`? I've had to update those tests are they're already a lot of text.

tests/pass-dep/shims/mmap.rs
```diff
@@ -163,7 +172,9 @@ fn test_mremap() {
 }
 
 fn main() {
-    test_mmap();
+    test_mmap(libc::mmap);
+    #[cfg(target_os = "linux")]
+    test_mmap(libc::mmap64);
```
RalfJung, line 177: You implemented this shim for all unices, but you are testing it only for Linux -- why?

src/shims/unix/linux/foreign_items.rs
```diff
@@ -217,6 +218,14 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 this.write_null(dest)?;
             }
 
+            "mmap64" => {
```
RalfJung, line 221: You put this into the category of
```
// Incomplete shims that we "stub out" just to get pre-main initialization code to work.
// These shims are enabled only when the caller is in the standard library.
```
That doesn't seem right.

sunfishcode: I didn't notice that comment :-). I've now moved it to "File related shims".

RalfJung:
Thanks!
@bors r+

PR #3287, opened 2024-02-01
macOs intercepting CCRandomGenerateBytes.

devnexen:


RalfJung:
Thanks for the PR!

IIRC we had that function, but removed it since it was not needed for the standard library nor for getrandom.

What is your motivation for adding it?

devnexen:
Do you mean [rust std](https://github.com/rust-lang/rust/blob/11f32b73e0dc9287e305b5b9980d24aecdc8c17f/library/std/src/sys/pal/unix/rand.rs#L186) ?

saethlin:
That's in a function with `#[cfg(not(target_os = "macos"))]`, so why would you want to shim this function on MacOS if the standard library deliberately avoids calling it?

oli-obk:
It's specifically for iOS, not macos. I don't think we have any support for iOS in miri yet.

devnexen:
> It's specifically for iOS, not macos. I don't think we have any support for iOS in miri yet.

ah good to know, no need indeed.

PR #3290, opened 2024-02-04
Stop recommending cargo clean in README

### LINKED ISSUE #3289: Stop recommending cargo clean?
The README says

> Now you can run your project in Miri:
>
> 1. Run cargo clean to eliminate any cached dependencies. Miri needs your dependencies to be compiled the right way, that would not happen if they have previously already been compiled.

But apparently Miri has used `target/miri` since 2021 to keep itself separate from non-miri builds:

https://github.com/rust-lang/miri/pull/1842
https://github.com/rust-lang/miri/commit/6a18683d09f74ead9438aa7fdd323b18821f7709

Would it be safe to simplify the README and save people & computers time by not asking for `cargo clean`?

https://github.com/rust-lang/miri/commit/6a18683d09f74ead9438aa7fdd323b18821f7709

tv42:
Miri has used the `target/miri` subdirectory since 2021 to keep itself separate from non-miri builds, so this should not be necessary.

See commit 6a18683d09f74ead9438aa7fdd323b18821f7709

Since the items are no longer a sequence of steps to do in order ("first, make sure that ..."), switch to an unordered list while we're at it.

Closes #3289

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,11 +108,8 @@ assume the right toolchain is pinned via `rustup override set nightly` or
 
 Now you can run your project in Miri:
 
-1. Run `cargo clean` to eliminate any cached dependencies. Miri needs your
-   dependencies to be compiled the right way, that would not happen if they have
-   previously already been compiled.
-2. To run all tests in your project through Miri, use `cargo miri test`.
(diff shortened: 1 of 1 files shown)


RalfJung:
Thanks :)

PR #3309, opened 2024-02-21
Test windows random shims

### LINKED ISSUE #3308: Add direct tests for the Windows RNG functions
Once https://github.com/rust-lang/rust/pull/121337 lands we'll support 3 RNG functions on Windows. I don't know which of these we are testing at all, since we're only testing them indirectly via std and getrandom. Would be good to have a test that calls them directly, just to make sure that they remain covered.

@ChrisDenton if you could help write such a test that would be great. :)

ChrisDenton:
This adds tests for `BCryptGenRandom` and `RtlGenRandom` (aka `SystemFunction036`). Note that `BCryptGenRandom` comes in two flavours:

```rust
BCryptGenRandom(null_mut(), key.as_mut_ptr(), len, BCRYPT_USE_SYSTEM_PREFERRED_RNG)
BCryptGenRandom(BCRYPT_RNG_ALG_HANDLE, key.as_mut_ptr(), len, 0)
```

Fixes #3308

### DIFF
--- a/tests/pass/shims/windows-rand.rs
+++ b/tests/pass/shims/windows-rand.rs
@@ -1,0 +1,41 @@
+//@only-target-windows: this directly tests windows only random functions
+use core::ffi::c_void;
+use core::mem::size_of_val;
+use core::ptr::null_mut;
+
+// Windows API definitions.
+type NTSTATUS = i32;
+type BOOLEAN = u8;
+const BCRYPT_USE_SYSTEM_PREFERRED_RNG: u32 = 0x00000002;
+const BCRYPT_RNG_ALG_HANDLE: *mut c_void = 0x81 as *mut c_void;
+#[link(name = "bcrypt")]
+extern "system" {
+    fn BCryptGenRandom(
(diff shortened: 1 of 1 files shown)


tests/pass/shims/windows-rand.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    let ret = unsafe { BCryptGenRandom(BCRYPT_RNG_ALG_HANDLE, key.as_mut_ptr(), len, 0) };
+    assert!(ret >= 0);
+
+    let ret = unsafe { RtlGenRandom(key.as_mut_ptr(), len) };
+    // RtlGenRandom returns a BOOLEAN where 0 indicates an error
+    assert_ne!(ret, 0);
+}
+
+// Windows API definitions.
```
RalfJung, line 23: Can you put the definitions above the code where they are used?

Other than that, looks good to me, thanks. :)

ChrisDenton: Done.

PR #3312, opened 2024-02-23
Add "cargo miri clean" command

RossSmyth:
My first reaction when my miri cache was messed up was to attempt run this, which obviously failed. This helps paper over platform differences and such.

### DIFF
--- a/cargo-miri/src/phases.rs
+++ b/cargo-miri/src/phases.rs
@@ -20,6 +20,7 @@ Subcommands:
     test, t                  Run tests
     nextest                  Run tests with nextest (requires cargo-nextest installed)
     setup                    Only perform automatic setup, but without asking questions (for getting a proper libstd)
+    clean                    Clean the Miri cache & target directory
 
 The cargo options are exactly the same as for `cargo run` and `cargo test`, respectively.
 
@@ -74,14 +75,15 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
     // We cannot know which of those flags take arguments and which do not,
     // so we cannot detect subcommands later.
     let Some(subcommand) = args.next() else {
-        show_error!("`cargo miri` needs to be called with a subcommand (`run`, `test`)");
+        show_error!("`cargo miri` needs to be called with a subcommand (`run`, `test`, `clean`)");
     };
     let subcommand = match &*subcommand {
         "setup" => MiriCommand::Setup,
         "test" | "t" | "run" | "r" | "nextest" => MiriCommand::Forward(subcommand),
+        "clean" => MiriCommand::Clean,
         _ =>
             show_error!(
-                "`cargo miri` supports the following subcommands: `run`, `test`, `nextest`, and `setup`."
+                "`cargo miri` supports the following subcommands: `run`, `test`, `nextest`, `clean`, and `setup`."
             ),
     };
     let verbose = num_arg_flag("-v");
@@ -92,6 +94,16 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
     let host = &rustc_version.host;
     let target = get_arg_flag_value("--target");
     let target = target.as_ref().unwrap_or(host);
+
+    // If cleaning the the target directory & sysroot cache,
+    // delete them then exit. There is no reason to setup a new
+    // sysroot in this execution.
+    if let MiriCommand::Clean = subcommand {
+        let metadata = get_cargo_metadata();
+        clean_target_dir(&metadata);
+        clean_sysroot();
+        return;
+    }
 
     // We always setup.
     let miri_sysroot = setup(&subcommand, target, &rustc_version, verbose);
@@ -110,6 +122,7 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
     let cargo_cmd = match subcommand {
         MiriCommand::Forward(s) => s,
         MiriCommand::Setup => return, // `cargo miri setup` stops here.
+        MiriCommand::Clean => unreachable!(),
     };
     let metadata = get_cargo_metadata();
     let mut cmd = cargo();
@@ -142,11 +155,7 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
         .arg(format!("target.'cfg(all())'.runner=[{cargo_miri_path_for_toml}, 'runner']"));
 
     // Set `--target-dir` to `miri` inside the original target directory.
-    let mut target_dir = match get_arg_flag_value("--target-dir") {
-        Some(dir) => PathBuf::from(dir),
-        None => metadata.target_directory.clone().into_std_path_buf(),
-    };
-    target_dir.push("miri");
+    let target_dir = get_target_dir(&metadata);
     cmd.arg("--target-dir").arg(target_dir);
 
     // *After* we set all the flags that need setting, forward everything else. Make sure to skip
--- a/cargo-miri/src/setup.rs
+++ b/cargo-miri/src/setup.rs
@@ -67,13 +67,8 @@ pub fn setup(
     }
 
     // Determine where to put the sysroot.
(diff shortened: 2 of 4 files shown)


RalfJung:
This makes a lot of sense, thanks!

I wonder, though... should it maybe also do what `cargo clean` does, i.e., clean the target dir? In particular since Miri has a separate target dir, I don't know if `cargo clean` would even clean that one.

cargo-miri/src/util.rs
```diff
@@ -249,3 +251,26 @@ pub fn debug_cmd(prefix: &str, verbose: usize, cmd: &Command) {
...
+}
+
+/// Deletes the Miri sysroot cache directory.
+pub fn clean_sysroot() {
+    let sysroot_dir = get_sysroot();
+
+    eprint!("[cargo-miri] clearing cache at {}", sysroot_dir.display());
+
+    // Keep it simple
+    std::fs::remove_dir_all(sysroot_dir).unwrap();
```
RalfJung, line 275: I wonder if we should refuse to do this when MIRI_SYSROOT is set. Otherwise there's a risk that it is set incorrect and then we'll delete the wrong thing...

RossSmyth:
@saethlin #3313

cargo-miri/src/util.rs
```diff
@@ -249,3 +251,26 @@ pub fn debug_cmd(prefix: &str, verbose: usize, cmd: &Command) {
...
+            user_dirs.cache_dir().to_owned()
+        }
+    }
+}
+
+/// Deletes the Miri sysroot cache directory.
+pub fn clean_sysroot() {
+    let sysroot_dir = get_sysroot();
+
+    eprint!("[cargo-miri] clearing cache at {}", sysroot_dir.display());
```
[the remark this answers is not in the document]
RossSmyth: Oh yeah I'll switch that. I chose eprint[ln]!() because it is used elsewhere in cargo-miri for outputs.

https://github.com/rust-lang/miri/blob/61af95673d0d9df50052c43e4ae2302824ef1089/cargo-miri/src/util.rs#L191
https://github.com/rust-lang/miri/blob/61af95673d0d9df50052c43e4ae2302824ef1089/cargo-miri/src/util.rs#L250
https://github.com/rust-lang/miri/blob/61af95673d0d9df50052c43e4ae2302824ef1089/cargo-miri/src/phases.rs#L299-L306
and others. Unless I misunderstood what those are for. But I'll switch it to eprintln!()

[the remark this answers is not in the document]
RalfJung: You'll notice that every single one of them is under an `if` that makes it only show in "verbose" mode.

RossSmyth:
> This makes a lot of sense, thanks!
> 
> I wonder, though... should it maybe also do what `cargo clean` does, i.e., clean the target dir? In particular since Miri has a separate target dir, I don't know if `cargo clean` would even clean that one.


Yeah it could do that as well. When I use `cargo clean` I mainly expect it to nuke any build-releated caches, so I would include the sysroot cache in "things that may effect the build". I can add cleaning the target directory as well though.

cargo-miri/src/util.rs
RossSmyth: Yeah that's reasonable.

cargo-miri/src/util.rs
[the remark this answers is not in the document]
RossSmyth: I do not believe that is true for all of them, mainly in setup.rs

https://github.com/rust-lang/miri/blob/61af95673d0d9df50052c43e4ae2302824ef1089/cargo-miri/src/setup.rs#L147-L157
https://github.com/rust-lang/miri/blob/61af95673d0d9df50052c43e4ae2302824ef1089/cargo-miri/src/setup.rs#L177-L187

But regardless, it doesn't make a difference to me. Should these be println!() instead?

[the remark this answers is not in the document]
RalfJung: Ah true, those are not "verbose" mode.

It seems like cargo prints its stuff to stderr as well, so `eprintln!` should be fine. The `[cargo-miri rustc]` prefix makes no sense here though, that's indicating where in cargo-miri certain verbose debug output comes from.

RossSmyth:
I just remembered that the `remove_dir_all` method errors if the directory does not exist. To keep in line with `cargo clean` this should not panic in that case.

cargo-miri/src/setup.rs
```diff
@@ -67,13 +67,8 @@ pub fn setup(
...
 
     // Determine where to put the sysroot.
-    let sysroot_dir = match std::env::var_os("MIRI_SYSROOT") {
-        Some(dir) => PathBuf::from(dir),
-        None => {
-            let user_dirs = directories::ProjectDirs::from("org", "rust-lang", "miri").unwrap();
-            user_dirs.cache_dir().to_owned()
-        }
-    };
+    let sysroot_dir = get_sysroot();
```
RalfJung, line 70: In symmetry with other function names, "get_sysroot_dir" might be a better name.

cargo-miri/src/util.rs
```diff
@@ -249,3 +251,62 @@ pub fn debug_cmd(prefix: &str, verbose: usize, cmd: &Command) {
...
+pub fn clean_sysroot() {
+    if std::env::var_os("MIRI_SYSROOT").is_some() {
+        show_error!(
+            "MIRI_SYSROOT is set. Aborting clean, please clean your custom cache directory manually."
+        )
+    }
+
+    let sysroot_dir = get_sysroot();
+
+    eprintln!("Cleaning cache at {}", sysroot_dir.display());
```
RalfJung, line 290:
```suggestion
    eprintln!("Cleaning sysroot cache at {}", sysroot_dir.display());
```

cargo-miri/src/util.rs
```diff
@@ -249,3 +251,62 @@ pub fn debug_cmd(prefix: &str, verbose: usize, cmd: &Command) {
...
+    let target_dir = get_target_dir(meta);
+
+    eprintln!("Cleaning target directory at {}", target_dir.display());
+
+    match std::fs::remove_dir_all(target_dir) {
+        Ok(_) => (),
+        // If the directory doesn't exist, it is still a success.
+        Err(err) if err.kind() == io::ErrorKind::NotFound => (),
+        Err(err) => show_error!("{}", err),
+    }
```
RalfJung, line 311: This snippet now exists twice, seems worth factoring into a helper function.

cargo-miri/src/util.rs
```diff
@@ -249,3 +251,62 @@ pub fn debug_cmd(prefix: &str, verbose: usize, cmd: &Command) {
...
+        }
+    }
+}
+
+/// Deletes the Miri sysroot cache
+pub fn clean_sysroot() {
+    if std::env::var_os("MIRI_SYSROOT").is_some() {
+        show_error!(
+            "MIRI_SYSROOT is set. Aborting clean, please clean your custom cache directory manually."
+        )
```
RalfJung, line 285: I'd say this should still clean the target dir, just not the sysroot cache, with a note saying why.

cargo-miri/src/phases.rs
```diff
@@ -93,6 +95,17 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
...
     let target = target.as_ref().unwrap_or(host);
 
+    // If cleaning the the target directory & sysroot cache,
+    // delete them then exit. There is no reason to setup a new
+    // sysroot in this execution.
+    if let MiriCommand::Clean = subcommand {
+        let metadata = get_cargo_metadata();
+        clean_sysroot();
+        clean_target_dir(&metadata);
+        eprintln!("Cleaning successful.");
```
RalfJung, line 105:
```suggestion
        clean_target_dir(&metadata);
```
I don't think we usually have such success messages for such simple operations. Absence of an error indicates success. :)

RalfJung:
LGTM, thanks!
@bors r+

PR #3315, opened 2024-02-23
Force panic=unwind for sysroot

### LINKED ISSUE #3313: Broken sysroot when custom "release" profile sets `panic="abort"`
Hello,

When attempting to use miri today I did the normal `cargo +nightly miri test`. But it failed to build the project with a mysterious error: 
> error: the crate `panic_unwind` does not have the panic strategy `unwind`

Eventually it was tracked down to me having a custom "release" profile, because I basically only use that profile for installing via `cargo install`. The key at fault is the `panic = "abort"` key. 

This is quite easy to replicate. Set the envar `CARGO_PROFILE_RELEASE_PANIC="abort"` then try to run anything with miri. For example I build the miri test suite and every test failed with the error above. Then since the sysroot cache is polluted you have to delete it to fix it hence my open PR :). 

cc @Nemo157 

Nemo described it as "I'm not even seeing panic=abort in the rustc commands"

RossSmyth:
This fixed #3313 partially for me. Previously, none of the miri tests could even build. Now they all pass :). 

There is a lingering issue that I don't quite understand yet, which is that attempting to do `miri test` with another crate can still having aborting behavior. 

The motivating example I have is https://github.com/PSeitz/lz4_flex, and how I found the issue to begin with. I can now build the test binaries, but now it immediately fails upon test execution with `(exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)` (Windows). 

It appears something is still being build with `panic="abort"` somewhere, but adding it to the other cargo invocation did not seem to solve it
https://github.com/rust-lang/miri/blob/92110c2c9353d70db7146c661b087f53c033075d/cargo-miri/src/phases.rs#L115 somewhere after here I unconditionally put `-Cpanic=unwind` as a test but the abort still occurred. And to note, the tests run fine without touching the cargo config and running `cargo test`.

### DIFF
--- a/cargo-miri/src/phases.rs
+++ b/cargo-miri/src/phases.rs
@@ -415,6 +415,9 @@ pub fn phase_rustc(mut args: impl Iterator<Item = String>, phase: RustcPhase) {
             && get_arg_flag_value("--crate-name").as_deref() == Some("panic_abort")
         {
             cmd.arg("-C").arg("panic=abort");
+        } else if phase == RustcPhase::Setup {
+            // If the user has a custom profile, this overrides it so that the sysroot is built properly.
+            cmd.arg("-C").arg("panic=unwind");
         }
     } else {
         // For host crates (but not when we are just printing some info),


RalfJung:
Can we set of build profile of "miri_sysroot" so that if people changed things about their "release" profile that just doesn't affect us?

cargo-miri/src/phases.rs
```diff
@@ -415,6 +415,9 @@ pub fn phase_rustc(mut args: impl Iterator<Item = String>, phase: RustcPhase) {
             && get_arg_flag_value("--crate-name").as_deref() == Some("panic_abort")
         {
             cmd.arg("-C").arg("panic=abort");
+        } else {
+            // If the user has a custom profile, this overrides it so that the sysroot is built properly.
+            cmd.arg("-C").arg("panic=unwind");
```
RalfJung, line 420: This now runs also outside `phase == RustcPhase::Setup`, that does not seem right.

RossSmyth:
> Can we set of build profile of "miri_sysroot" so that if people changed things about their "release" profile that just doesn't affect us?

Probably? Now that you say that I think this PR probably isn't the way to go.

PR #3316, opened 2024-02-24
Windows miri-script execution egronomics

RossSmyth:
This allows for Windows users to use miri-script without pain. As working on miri earlier I was doing
`.\miri-script\target\debug\miri-script.exe { install | build | ... }` which wasn't fun.

### DIFF
--- a/miri.bat
+++ b/miri.bat
@@ -1,0 +1,10 @@
+:: This makes execution of ./miri on Linux and Windows the same.
+:: Windows will not execute the bash script, and select this.
+@echo off
+set MIRI_SCRIPT_TARGET_DIR=%0\..\miri-script\target
+cargo build %CARGO_EXTRA_FLAGS% -q --target-dir %MIRI_SCRIPT_TARGET_DIR% --manifest-path %0\..\miri-script\Cargo.toml
+
+:: Forwards all arguments to this file to the executable.
+:: We invoke the binary directly to avoid going through rustup, which would set some extra
+:: env vars that we do not want.
+%MIRI_SCRIPT_TARGET_DIR%\debug\miri-script %*


miri.bat
```diff
@@ -0,0 +1,8 @@
+:: This makes execution of ./miri on Linux and Windows the same.
+:: Windows will not executable the bash script, and select this.
```
RalfJung, line 2:
```suggestion
:: Windows will not execute the bash script, and select this.
```
Does this depend on the shell ones uses (old cmd / powershell / not sure if there's anything else)?

miri.bat
```diff
@@ -0,0 +1,8 @@
...
+@echo off
+set MIRI_SCRIPT_TARGET_DIR=%0\..\miri-script\target
+cargo build %CARGO_EXTRA_FLAGS% -q --target-dir %MIRI_SCRIPT_TARGET_DIR% --manifest-path %0\..\miri-script\Cargo.toml
+
+:: Forwards all arguments to this file to the executable.
+%MIRI_SCRIPT_TARGET_DIR%\debug\miri-script %*
```
RalfJung, line 8:
```suggestion
:: We invoke the binary directly to avoid going through rustup, which would set some extra
:: env vars that we do not want.
%MIRI_SCRIPT_TARGET_DIR%\debug\miri-script %*
```

RalfJung:
Thanks! None of us on the team use Windows so it's hard for us to even know where the issues are, let alone how to fix them. :)

In the future it'd be good to actually use this on CI (CI is a bash script even on Windows so I assume this script is not used there), but that doesn't have to happen for this PR.

miri.bat
RossSmyth: No it does not matter

RossSmyth: Unless they modify their PATHEXT env var themselves, which is on them.

RossSmyth:
Good idea, I'll open a PR with CI later.

RalfJung:
@bors r+
Thanks!

Issue #3297, opened 2024-02-12
"./miri test" is very slow on repeated executions


RalfJung:
Running `./miri run --dep` multiple times is very slow. The first time I guess it is building parts of Miri or so (`./miri run` doesn't print *any* updates about what it does, making this hard to figure out), but the 2nd time it shouldn't have to do any of that -- and yet, it is slow.

Needless to say, this is quite painful when debugging a pass-dep test.

RalfJung:
It's not just "./miri run". "./miri test" and "./miri build" have the same issue.

This was fine until recently, so something must have changed somewhere to break this... *sigh*

RalfJung:
I think it has to do with ui_test dependency building.

Repeating "./miri test hello" is fine, but repeating a pass-dep test rebuilds a lot of stuff.

Cc @oli-obk -- somehow it seems like ui_test dependency building is now polluting Miri's own build cache?

oli-obk:
Huh. We're still on a version from October. I'll investigate

RalfJung:
Turns out this has been ongoing for a while.
f006d42618a038f7e38d2b59d1b0664727e51382 is still bad.

RalfJung:
Bisecting says this is the bad PR: https://github.com/rust-lang/miri/pull/3252.

That did indeed touch the "deps build" logic, by adding an env var.

bjorn3:
As far as I can tell the `RUSTFLAGS` are passed to a `cargo run` of cargo-miri with the same target dir as miri. As such any shared dependencies will see a different `RUSTFLAGS` between building miri / the regular build of cargo-miri and the `cargo run` of cargo-miri for the purpose of building the test dependencies. I think it would be possible to fix this by directly running the previously built cargo-miri executable rather than rebuilding it with empty `RUSTFLAGS`.

RalfJung:
Oh I see, so the issue isn't that dep builds get mixed with the other builds but that we're re-building `cargo-miri` inside the dep build...

PR #3318, opened 2024-02-24
compiletest: call cargo-miri directly rather than via 'cargo run'

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/3297. Thanks to @bjorn3 for figuring out the cause of this.

r? @oli-obk

### DIFF
--- a/tests/compiletest.rs
+++ b/tests/compiletest.rs
@@ -99,18 +99,15 @@ fn test_config(target: &str, path: &str, mode: Mode, with_dependencies: bool) ->
     };
 
     if with_dependencies {
+        // Set the `cargo-miri` binary, which we expect to be in the same folder as the `miri` binary.
+        // (It's a separate crate, so we don't get an env var from cargo.)
+        let mut prog = miri_path();
+        prog.set_file_name("cargo-miri");
+        config.dependency_builder.program = prog;
+        let builder_args = ["miri", "run"]; // There is no `cargo miri build` so we just use `cargo miri run`.
+        config.dependency_builder.args = builder_args.into_iter().map(Into::into).collect();
         config.dependencies_crate_manifest_path =
             Some(Path::new("test_dependencies").join("Cargo.toml"));
-        let mut builder_args = vec!["run".into()];
-        builder_args.extend(flagsplit(&env::var("CARGO_EXTRA_FLAGS").unwrap_or_default()));
-        builder_args.extend([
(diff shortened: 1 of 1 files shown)


tests/compiletest.rs
```diff
@@ -99,18 +99,15 @@ fn test_config(target: &str, path: &str, mode: Mode, with_dependencies: bool) ->
     };
 
     if with_dependencies {
+        // Set the `cargo-miri` binary, which we expect to be in the same folder as the `miri` binary.
+        // (It's a separate crate, so we don't get an env var from cargo.)
+        let mut prog = miri_path();
+        prog.set_file_name("cargo-miri");
```
bjorn3, line 105: This needs the `.exe` extension on Windows, right?

RalfJung: Does it?

I guess CI will tell...

RossSmyth: .exe is in the PATHEXT env var so it should just execute (that's how my miri.bat file works). Though since this is executed in bash on msys2 I'm not sure if that applies or not as I do not use that.

bjorn3: I thought PATHEXT only applied to cmd.exe and `ShellExecute` and not `CreateProcess`. `std::process::Command` uses `CreateProcessW`. The documentation of `CreateProcessW` doesn't mention whether or not it does apply

RalfJung: Windows CI seems to be working so it does seem to add the `.exe`. :shrug:

ChrisDenton: `Command`, for security reasons, uses its own logic for finding the exe and for compatibility reasons it always assumes `.exe` was intended if no extension is present.

RalfJung: @ChrisDenton do you always show up when someone says the word "Windows" in a Rust repo? ;)

Anyway, thanks for the explanation! :)

ChrisDenton: Yes, but only if the proper rituals are observed

oli-obk:
Will that work on a clean build running `./miri test tests/pass-dep`? If not, we should at least provide some useful diagnostic to the user

bjorn3:
It should. miri-script already builds cargo-miri before running the test suite.

RalfJung:
Yes, `cargo clean && ./miri test tests/pass-dep` still works.

PR #3319, opened 2024-02-24
Directly implement native exception raise methods in miri

bjorn3:
This implements the `_Unwind_RaiseException` function used on pretty much every unix system for starting unwinding. This allows removing the miri special case from libpanic_unwind for unix.

Windows still needs `miri_start_unwind` as SEH unwinding isn't supported by miri. Unlike DWARF unwinding, SEH preserves all stack frames until right after the do_catch function has executed. Because of this panic_unwind stack allocates the exception object. Miri can't currently model unwinding without destroying stack frames and as such will report a use-after-free of the exception object.

### DIFF
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -640,14 +640,28 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 this.write_scalar(Scalar::from_target_usize(len, this), dest)?;
             }
             "_Unwind_RaiseException" => {
-                trace!("_Unwind_RaiseException: {:?}", this.frame().instance);
-
-                // Get the raw pointer stored in arg[0] (the panic payload).
+                // This is not formally part of POSIX, but it is very wide-spread on POSIX systems.
+                // It was originally specified as part of the Itanium C++ ABI:
+                // https://itanium-cxx-abi.github.io/cxx-abi/abi-eh.html#base-throw.
+                // On Linux it is
+                // documented as part of the LSB:
+                // https://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/baselib--unwind-raiseexception.html
+                // Basically every other UNIX uses the exact same api though. Arm also references
+                // back to the Itanium C++ ABI for the definition of `_Unwind_RaiseException` for
+                // arm64:
+                // https://github.com/ARM-software/abi-aa/blob/main/cppabi64/cppabi64.rst#toc-entry-35
+                // For arm32 they did something custom, but similar enough that the same
+                // `_Unwind_RaiseException` impl in miri should work:
+                // https://github.com/ARM-software/abi-aa/blob/main/ehabi32/ehabi32.rst
+                if !matches!(&*this.tcx.sess.target.os, "linux" | "freebsd" | "illumos" | "solaris" | "android" | "macos") {
+                    throw_unsup_format!(
+                        "`_Unwind_RaiseException` is not supported on {}",
+                        this.tcx.sess.target.os
+                    );
+                }
+                // This function looks and behaves excatly like miri_start_unwind.
                 let [payload] = this.check_shim(abi, Abi::C { unwind: true }, link_name, args)?;
-                let payload = this.read_scalar(payload)?;
-                let thread = this.active_thread_mut();
-                thread.panic_payloads.push(payload);
-
+                this.handle_miri_start_unwind(payload)?;
                 return Ok(EmulateItemResult::NeedsUnwind);
             }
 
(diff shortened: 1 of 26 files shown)


RalfJung:
I was confused by the context here, since it's unclear what issue you are fixing, and what "old custom ABI" is.

After reading the PR code and mulling over this for a bit, I think "old custom ABI" is the `miri_start_panic` way of doing panics? I'd never call this an ABI so I didn't think of this at all when reading the PR description.^^ And the entire PR description is about what the PR does *not* do, rather than what the PR actually does, right? Would be good to update it to make it less confusing. :)

src/shims/unix/foreign_items.rs
```diff
@@ -729,6 +729,18 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 this.write_scalar(Scalar::from_i32(-1), dest)?;
             }
 
+            "_Unwind_RaiseException" => {
```
RalfJung, line 732: So is this a standard Unix function? What are the documented guarantees for how it behaves? We should be careful to not implement it in a way that provides fewer or more guarantees than the official spec.

bjorn3: It was originally specified as part of the Itanium C++ ABI afaik: https://itanium-cxx-abi.github.io/cxx-abi/abi-eh.html#base-throw On Linux it is documented as part of the LSB: https://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/baselib--unwind-raiseexception.html Basically every other UNIX uses the exact same api though. Arm also references back to the Itanium C++ ABI for the definition of `_Unwind_RaiseException` for arm64: https://github.com/ARM-software/abi-aa/blob/main/cppabi64/cppabi64.rst#toc-entry-35 For arm32 they did something custom, but similar enough that the same `_Unwind_RaiseException` impl in miri should work: https://github.com/ARM-software/abi-aa/blob/main/ehabi32/ehabi32.rst

bjorn3:
> I was confused by the context here, since it's unclear what issue you are fixing, and what "old custom ABI" is.

My bad. I was indeed referring to `miri_start_panic`.

src/shims/unix/foreign_items.rs
RalfJung: Thanks! So this is way more standard than I thought.

This seems a bit asymmetric though, one can now start unwinding with a standard function but how does one catch it? Is it normal that that still requires a magic builtin? Hm I guess that's not even a Miri magic function, that's an intrinsic.

Final point, how does this work when different languages do unwinding? My understanding is that this is somehow distinguished, there is such a thing as "foreign unwinding" coming into Rust code which is different from a Rust panic -- but how is it different? Just different payload behind this pointer? Is that why the payload must have type `_Unwind_Exception`? Maybe Miri should check that, as [these docs](https://itanium-cxx-abi.github.io/cxx-abi/abi-eh.html#base-throw) indicate, the exception_class and exception_cleanup fields are set? And Miri assumes all unwinding is Rust unwinding so exception_class must be set to whatever the magic value is that Rust uses there.

RalfJung: > The current std::panic::catch_unwind implementation does abort when catching a foreign exception by calling __rust_foreign_exception()

But nothing in Miri sets this? So with `miri_start_panic` does the caller also set this  "M O Z \0 R U S T" class?

RalfJung: Oh I get it, `catch_unwind` only does that when using the "gcc" unwinding implementation. Basically the Miri unwinding "ABI" doesn't currently have support for indicating the exception class.

But this means it should be possible to throw a foreign exception with `_Unwind_Exception` and then catch it with `try` (what a terrible name^^) and observe the payload... Oh wait, that's exactly what your test is doing.^^

tests/pass/panic/unwind_dwarf.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    let exception = exception.cast::<Exception>();
+
+    let exception = Box::from_raw(exception as *mut Exception);
+    exception.cause
+}
+
+fn rust_exception_class() -> uw::_Unwind_Exception_Class {
+    // M O Z \0  R U S T -- vendor, language
+    0x4d4f5a_00_52555354
```
RalfJung, line 52: Maybe set this to something else since we're testing the unwinder mechanism in general, not specifically Rust panics?

RalfJung:
For miri_start_panic we have some interesting tests: miri_bad_start_panic, unwind_panic_abort. We probably want to have similar tests for this as well. Other than that, LGTM.

RalfJung:
@bjorn3 what is the status of this? This seems basically done, just needs more tests.

RalfJung:
I did a bit of refactoring to finish the PR. Thanks for the initial idea!
@bors r+

PR #3330, opened 2024-02-26
Fix .\miri fmt on Windows

### LINKED ISSUE #3317: miri fmt failing on Windows
One issue I am still having is rustfmt isn't working. I had that issue earlier as well. It just says 
```ps
PS C:\Users\Ross\Documents\miri> .\miri fmt -v
$ rustfmt +miri --edition=2021 --config-path C:\Users\Ross\Documents\miri\rustfmt.toml --unstable-features --skip-children -v ...
Error: `rustfmt` failed
```
Adding or removing `-v` doesn't change the output at all.

RossSmyth:
This allows .\miri fmt to work on Windows. Closes #3317.

To reiterate, the problem with using `miri fmt` on Windows is that the CLI arguments to rustfmt are too long. Currently over 65,000 characters are used in the call to rustfmt, [which is incompatible with Windows](https://devblogs.microsoft.com/oldnewthing/20031210-00/?p=41553) as it is limited to (2^15 - 1) for all arguments plus all env vars.

Two things are done do get around this limit:

1. Call out to cargo-fmt for the crates that exist. 
2. Batch rustfmt calls by length

Another alternative would be to just use rustfmt for everything and don't use cargo-fmt for the crates.

I don't know how much you guys may care about `miri fmt` time to run. I don't know the diff as it did not work before on my computer. 

[I have another branch that solves this, but in a less permanent way](RossSmyth/miri/tree/windows-fmt). That was my initial attempt, and I learned that even with cargo-fmt and relative paths, the rustfmt call still has 27k characters. This is closer to the limit than I expected, so it would not be a permanent solution. So I went back to absolute paths & batching.

### DIFF
--- a/miri-script/src/commands.rs
+++ b/miri-script/src/commands.rs
@@ -526,37 +526,27 @@ impl Command {
     }
 
     fn fmt(flags: Vec<OsString>) -> Result<()> {
+        use itertools::Itertools;
+
         let e = MiriEnv::new()?;
-        let toolchain = &e.toolchain;
         let config_path = path!(e.miri_dir / "rustfmt.toml");
 
-        let mut cmd = cmd!(
-            e.sh,
-            "rustfmt +{toolchain} --edition=2021 --config-path {config_path} --unstable-features --skip-children {flags...}"
-        );
-        eprintln!("$ {cmd} ...");
+        // Collect each rust file in the miri repo.
+        let files = WalkDir::new(&e.miri_dir)
+            .into_iter()
+            .filter_entry(|entry| {
+                let name = entry.file_name().to_string_lossy();
+                let ty = entry.file_type();
+                if ty.is_file() {
+                    name.ends_with(".rs")
+                } else {
+                    // dir or symlink. skip `target` and `.git`.
+                    &name != "target" && &name != ".git"
+                }
+            })
+            .filter_ok(|item| item.file_type().is_file())
+            .map_ok(|item| item.into_path());
 
-        // Add all the filenames to the command.
-        // FIXME: `rustfmt` will follow the `mod` statements in these files, so we get a bunch of
-        // duplicate diffs.
-        for item in WalkDir::new(&e.miri_dir).into_iter().filter_entry(|entry| {
-            let name = entry.file_name().to_string_lossy();
-            let ty = entry.file_type();
-            if ty.is_file() {
-                name.ends_with(".rs")
-            } else {
-                // dir or symlink. skip `target` and `.git`.
-                &name != "target" && &name != ".git"
-            }
-        }) {
-            let item = item?;
-            if item.file_type().is_file() {
-                cmd = cmd.arg(item.into_path());
-            }
-        }
-
-        // We want our own error message, repeating the command is too much.
-        cmd.quiet().run().map_err(|_| anyhow!("`rustfmt` failed"))?;
-        Ok(())
+        e.format_files(files, &e.toolchain[..], &config_path, &flags[..])
     }
 }
--- a/miri-script/src/util.rs
+++ b/miri-script/src/util.rs
@@ -148,7 +148,7 @@ impl MiriEnv {
 
     /// Receives an iterator of files.
     /// Will format each file with the miri rustfmt config.
-    /// Does not follow module relationships.
+    /// Does not recursively format modules.
     pub fn format_files(
         &self,
         files: impl Iterator<Item = Result<PathBuf, walkdir::Error>>,
@@ -160,7 +160,7 @@ impl MiriEnv {
 
         let mut first = true;
 
-        // Format in batches as not all out files fit into Windows' command argument limit.
+        // Format in batches as not all our files fit into Windows' command argument limit.
         for batch in &files.chunks(256) {
             // Build base command.
             let mut cmd = cmd!(
@@ -168,6 +168,7 @@ impl MiriEnv {
                 "rustfmt +{toolchain} --edition=2021 --config-path {config_path} --unstable-features --skip-children {flags...}"
             );
             if first {
+                // Log an abbreviating command, and only once.
                 eprintln!("$ {cmd} ...");
                 first = false;
             }
@@ -181,7 +182,7 @@ impl MiriEnv {
                 cmd = cmd.arg(file);
             }
 
-            // Run commands.
+            // Run rustfmt.
             // We want our own error message, repeating the command is too much.
             cmd.quiet().run().map_err(|_| anyhow!("`rustfmt` failed"))?;
         }
--- a/miri-script/src/util.rs
+++ b/miri-script/src/util.rs
@@ -1,7 +1,7 @@
 use std::ffi::{OsStr, OsString};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
-use anyhow::{Context, Result};
+use anyhow::{anyhow, Context, Result};
 use dunce::canonicalize;
 use path_macro::path;
 use xshell::{cmd, Shell};
@@ -143,6 +143,49 @@ impl MiriEnv {
             "cargo +{toolchain} test {cargo_extra_flags...} --manifest-path {manifest_path} {args...}"
         )
         .run()?;
+        Ok(())
+    }
+
+    /// Receives an iterator of files.
+    /// Will format each file with the miri rustfmt config.
+    /// Does not follow module relationships.
+    pub fn format_files(
+        &self,
+        files: impl Iterator<Item = Result<PathBuf, walkdir::Error>>,
+        toolchain: &str,
+        config_path: &Path,
+        flags: &[OsString],
+    ) -> anyhow::Result<()> {
+        use itertools::Itertools;
+
+        let mut first = true;
+
+        // Format in batches as not all out files fit into Windows' command argument limit.
+        for batch in &files.chunks(256) {
+            // Build base command.
+            let mut cmd = cmd!(
+                self.sh,
+                "rustfmt +{toolchain} --edition=2021 --config-path {config_path} --unstable-features --skip-children {flags...}"
+            );
+            if first {
+                eprintln!("$ {cmd} ...");
+                first = false;
+            }
+            // Add files.
+            for file in batch {
+                // Make it a relative path so that on platforms with extremely tight argument
+                // limits (like Windows), we become immune to someone cloning the repo
+                // 50 directories deep.
+                let file = file?;
+                let file = file.strip_prefix(&self.miri_dir)?;
+                cmd = cmd.arg(file);
+            }
+
+            // Run commands.
+            // We want our own error message, repeating the command is too much.
+            cmd.quiet().run().map_err(|_| anyhow!("`rustfmt` failed"))?;
+        }
+
         Ok(())
     }
 }


miri-script/src/commands.rs
```diff
@@ -526,37 +526,56 @@ impl Command {
...
+                if ty.is_file() {
+                    name.ends_with(".rs")
+                } else {
+                    // dir or symlink. skip `target` and `.git`.
+                    &name != "target" && &name != ".git"
+                }
+            })
+            .filter_ok(|item| item.file_type().is_file())
+            .map_ok(|item| item.path().to_owned())
+            .map(Result::unwrap);
```
RalfJung, line 575: I'd much prefer if we could avoid having to list every single crate in the repo. Just formatting all .rs files means we don't have to worry about keeping this list up-to-date. So can you please go back to the old structure here, and just do the batching?

RalfJung:
> I don't know how much you guys may care about miri fmt time to run. I don't know the diff as it did not work before on my computer.

I run this in my git hook, so I care about it being basically instantaneous as it is currently. This will have to be benchmarked before it can land.

miri-script/src/commands.rs
RossSmyth: Yeah thinking about it further last night having two different ways doesn't make too much sense.

RossSmyth:
On my computer, which is basically worst case right now as it is Windows, this is what I measured.
```powershell
C:\Users\rsmyth\Documents\miri> Measure-Command { .\miri fmt }


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 2
Milliseconds      : 119
Ticks             : 21195402
TotalDays         : 2.45317152777778E-05
TotalHours        : 0.000588761166666667
TotalMinutes      : 0.03532567
TotalSeconds      : 2.1195402
TotalMilliseconds : 2119.5402
```

miri-script/src/util.rs
```diff
@@ -146,3 +146,84 @@ impl MiriEnv {
...
+            // Peek and check to see if we go above the threshold with the
+            // next path
+            if let Some(path) = file_iter.peek() {
+                // This checks to make sure the current Cmd's arguments are
+                // in-bound of the OS's maximum string length.
+                //
+                // I tried manually counting in a previous revision, but that
+                // turns out to be non-trival due to escaping.
+                if cmd.to_string().len() + path.as_os_str().len() + env_length
+                    >= MAX_ARGUMENT_LENGTH
```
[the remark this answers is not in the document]
RossSmyth: I mainly did it to have the least batches, but I guess it doesn't really matter.

miri-script/src/commands.rs
```diff
@@ -526,37 +526,28 @@ impl Command {
...
+                }
+            })
+            .filter_ok(|item| item.file_type().is_file())
+            .map_ok(|item| item.path().to_owned());
 
-        // We want our own error message, repeating the command is too much.
-        cmd.quiet().run().map_err(|_| anyhow!("`rustfmt` failed"))?;
-        Ok(())
+        format_files(&e.sh, files, &e.toolchain[..], &config_path, &flags[..])
+            .map_err(|_| anyhow!("`rustfmt` failed"))
```
RalfJung, line 551: Please don't `map_err` here, there's no reason to do this.

miri-script/src/util.rs
```diff
@@ -146,3 +146,43 @@ impl MiriEnv {
...
+        if first {
+            eprintln!("$ {cmd} ...");
+            first = false;
+        }
+        // Add files.
+        for file in batch {
+            // Make it a relative path so that we become immune to someone cloning the repo
+            // 50 directories deep
+            let file = file?;
+            let file = file.strip_prefix(shell.current_dir())?;
```
RalfJung, line 178:
```suggestion
            let file = file?.strip_prefix(shell.current_dir())?;
```

miri-script/src/util.rs
```diff
@@ -146,3 +146,43 @@ impl MiriEnv {
...
+            "rustfmt +{toolchain} --edition=2021 --config-path {config_path} --unstable-features --skip-children {flags...}"
+        );
+        if first {
+            eprintln!("$ {cmd} ...");
+            first = false;
+        }
+        // Add files.
+        for file in batch {
+            // Make it a relative path so that we become immune to someone cloning the repo
+            // 50 directories deep
```
RalfJung, line 176:
```suggestion
            // Make it a relative path so that on platforms with extremely tight argument
            // limits (like Windows), we become immune to someone cloning the repo
            // 50 directories deep.
```

miri-script/src/util.rs
```diff
@@ -146,3 +146,43 @@ impl MiriEnv {
...
+    shell: &Shell,
+    files: impl Iterator<Item = Result<PathBuf, walkdir::Error>>,
+    toolchain: &str,
+    config_path: &Path,
+    flags: &[OsString],
+) -> anyhow::Result<()> {
+    use itertools::Itertools;
+
+    let mut first = true;
+    for batch in &files.chunks(128) {
```
RalfJung, line 163:
```suggestion
    // Format in batches as not all out files fit into Windows' command argument limit.
    for batch in &files.chunks(128) {
```

RalfJung (commented):
Code mostly LGTM, but this is a 20% perf regression on my system when all files are in the cache. I'm not sure why that is, process spawning shouldn't be *so* expensive...

miri-script/src/commands.rs
```diff
@@ -526,37 +526,28 @@ impl Command {
...
+                let ty = entry.file_type();
+                if ty.is_file() {
+                    name.ends_with(".rs")
+                } else {
+                    // dir or symlink. skip `target` and `.git`.
+                    &name != "target" && &name != ".git"
+                }
+            })
+            .filter_ok(|item| item.file_type().is_file())
+            .map_ok(|item| item.path().to_owned());
```
RalfJung, line 548:
```suggestion
            .map_ok(|item| item.into_path());
```
avoids some unnecessary cloning

RalfJung:
Looks like process spawning *is* that expensive. Going to chunks of size 512 largely fixes the perf regression. It's like an extra 10ms per spawned process, and we have around 900 files.

We could probably offset this by doing things in parallel, but then we need crossbeam's queue -- not sure if that's worth it. Also xshell is `!Sync` so we can't easily spawn multiple rustfmt in parallel.

Maybe let's just go for 256 files per chunk for now and call it a day.

miri-script/src/util.rs
RossSmyth: This doesn't exactly work because `strip_prefix()` returns a `&Path`, not a `PathBuf` so the PathBuf is dropped at the end of the statement.

RossSmyth:
[Just as an experiment I made a naive parallel `miri fmt` with Rayon](https://github.com/RossSmyth/miri/blob/2936268be88588ff9b7b70d4c60bbe637e9f7296/miri-script/src/util.rs#L162-L202). Biggest pain point is that the xshell Shell type is not Clone. The results (both with files in cache):

The current commit on this PR:
```ps
Minutes           : 0
Seconds           : 1
Milliseconds      : 861
Ticks             : 18612607
TotalDays         : 2.1542369212963E-05
TotalHours        : 0.000517016861111111
TotalMinutes      : 0.0310210116666667
TotalSeconds      : 1.8612607
TotalMilliseconds : 1861.2607
```

Parallel:
```ps
Minutes           : 0
Seconds           : 0
Milliseconds      : 984
Ticks             : 9844392
TotalDays         : 1.13939722222222E-05
TotalHours        : 0.000273455333333333
TotalMinutes      : 0.01640732
TotalSeconds      : 0.9844392
TotalMilliseconds : 984.4392
```

Though if you compile `miri-script` with `--release` (is that something anyone does?) the results of the parallel impl decrease by 10-20ms, while the serial impl decrease by about 900ms so that it is reliably within 20ms of the parallel impl. 

A better impl could be made but this was just a nerdsnipe since I was staring at it and Rayon is easy.

miri-script/src/commands.rs
RalfJung: Why did you resolve this? The `map_err` is still present!

miri-script/src/util.rs
```diff
@@ -146,3 +146,46 @@ impl MiriEnv {
         Ok(())
     }
 }
+
+/// Receives an iterator of files.
+/// Will format each file with the miri rustfmt config.
+/// Does not follow module relationships.
+pub fn format_files(
```
RalfJung, line 153: Please put this into the `impl MiriEnv`, then you also need to pass fewer parameters.

miri-script/src/util.rs
```diff
@@ -146,3 +146,46 @@ impl MiriEnv {
...
+            // limits (like Windows), we become immune to someone cloning the repo
+            // 50 directories deep.
+            let file = file?;
+            let file = file.strip_prefix(shell.current_dir())?;
+            cmd = cmd.arg(file);
+        }
+
+        // Run commands.
+        // We want our own error message, repeating the command is too much.
+        cmd.quiet().run()?;
```
RalfJung, line 187: This is where the `map_err` should go.

RalfJung:
> [Just as an experiment I made a naive parallel miri fmt with Rayon](https://github.com/RossSmyth/miri/blob/2936268be88588ff9b7b70d4c60bbe637e9f7296/miri-script/src/util.rs#L162-L202). Biggest pain point is that the xshell Shell type is not Clone. The results (both with files in cache):

Yeah you are now running `MiriEnv::new()` over and over again, I am not surprised that this doesn't go very fast. That function is expensive.

Given my measurements above I am okay with not parallelizing this for now.

miri-script/src/util.rs
```diff
@@ -145,4 +145,47 @@ impl MiriEnv {
         .run()?;
         Ok(())
     }
+
+    /// Receives an iterator of files.
+    /// Will format each file with the miri rustfmt config.
+    /// Does not follow module relationships.
```
RalfJung, line 151:
```suggestion
    /// Does not recursively format modules.
```

miri-script/src/util.rs
```diff
@@ -145,4 +145,47 @@ impl MiriEnv {
...
+        files: impl Iterator<Item = Result<PathBuf, walkdir::Error>>,
+        toolchain: &str,
+        config_path: &Path,
+        flags: &[OsString],
+    ) -> anyhow::Result<()> {
+        use itertools::Itertools;
+
+        let mut first = true;
+
+        // Format in batches as not all out files fit into Windows' command argument limit.
```
RalfJung, line 163:
```suggestion
        // Format in batches as not all our files fit into Windows' command argument limit.
```

miri-script/src/util.rs
```diff
@@ -145,4 +145,47 @@ impl MiriEnv {
...
+        for batch in &files.chunks(256) {
+            // Build base command.
+            let mut cmd = cmd!(
+                self.sh,
+                "rustfmt +{toolchain} --edition=2021 --config-path {config_path} --unstable-features --skip-children {flags...}"
+            );
+            if first {
+                eprintln!("$ {cmd} ...");
```
RalfJung, line 171:
```suggestion
                // Log an abbreviating command, and only once.
                eprintln!("$ {cmd} ...");
```

miri-script/src/util.rs
```diff
@@ -145,4 +145,47 @@ impl MiriEnv {
...
+            for file in batch {
+                // Make it a relative path so that on platforms with extremely tight argument
+                // limits (like Windows), we become immune to someone cloning the repo
+                // 50 directories deep.
+                let file = file?;
+                let file = file.strip_prefix(&self.miri_dir)?;
+                cmd = cmd.arg(file);
+            }
+
+            // Run commands.
```
RalfJung, line 184:
```suggestion
            // Run rustfmt.
```

PR #3339, opened 2024-03-02
remove some flags that haven't had an effect in a while

RalfJung:
@rust-lang/miri any objections to this?


saethlin:
Nope. Happy to see it.

RalfJung:
All right let's land this then.
@bors r=saethlin

PR #3345, opened 2024-03-03
Windows: support getting the thread name

RalfJung:
Also organize the thread name tests a bit.

### DIFF
--- a/src/shims/windows/foreign_items.rs
+++ b/src/shims/windows/foreign_items.rs
@@ -5,6 +5,7 @@ use rustc_span::Symbol;
 use rustc_target::abi::Size;
 use rustc_target::spec::abi::Abi;
 
+use crate::shims::os_str::bytes_to_os_str;
 use crate::*;
 use shims::foreign_items::EmulateForeignItemResult;
 use shims::windows::handle::{EvalContextExt as _, Handle, PseudoHandle};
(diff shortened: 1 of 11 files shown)


src/shims/windows/foreign_items.rs
```diff
@@ -412,7 +422,31 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+
+                let handle = this.read_scalar(handle)?;
+                let name_ptr = this.deref_pointer(name_ptr)?; // the pointer where we should store the ptr to the name
+
+                let thread = match Handle::from_scalar(handle, this)? {
+                    Some(Handle::Thread(thread)) => thread,
+                    Some(Handle::Pseudo(PseudoHandle::CurrentThread)) => this.get_active_thread(),
+                    _ => this.invalid_handle("SetThreadDescription")?,
+                };
+                // FIXME what is the right default thread name? Or should we raise an error?
```
ChrisDenton, line 442: The right default would be an empty string. Though in std that would be changed to `<unnamed>` anyways.

RalfJung: So on actual Windows it succeeds and returns an empty string when no name was that?
That's easy enough to do. Is this documented anywhere?

Yeah I know this is not observable on std.

ChrisDenton: > Is this documented anywhere?

No.

Issue #3350, opened 2024-03-03
Disable diagnostic duplication


RalfJung:
We'll want to disable diagnostic duplication in Miri, at least when "tracking" of some sort is enabled. I just wasted literal hours debugging why a certain load event did not happen that was supposed to happen, when actually the output just got deduplicated away...

PR #3351, opened 2024-03-03
disable diagnostic deduplication

RalfJung:
@oli-obk is there a better way to do this? Ideally we'd only set this when interpretation starts but the value in the compiler session seems to be immutable. I assume people will do `cargo check` before `cargo miri` so hopefully this won't lead to too much confusion.

Fixes https://github.com/rust-lang/miri/issues/3350


RalfJung:
Ah... some errors we do want to get deduplicated. Looks like we'll have to do our own deduplication for those.

PR #3370, opened 2024-03-09
windows: remove support for slim rwlock

RalfJung:
Since https://github.com/rust-lang/rust/pull/121956 we don't need it any more, and we are generally short on Windows staff so reducing the amount of code we have to test and maintain sounds like a good idea.

The InitOnce stuff is still used by `thread_local_key::StaticKey` on 64bit windows-gnu.


RalfJung:
@rust-lang/miri any objections to this?

saethlin:
No objection. I don't see any reason to support things like this that aren't used in the standard library anymore. If this is a problem for someone, hopefully they file an issue so we can at least know that such usage exists.

PR #3376, opened 2024-03-14
Improve sysroots notification

saethlin:
* Removes the T-miri ping (because these failures can't be acted on quickly, ergo they cannot be urgent)
* Puts the message in a separate Zulip topic
* Lists which targets failed to build in the Zulip message


saethlin:
I tested this on my fork, and it seemed to work as intended, with the exception that I can't actually send the Zulip message. But I checked that the tarball extracts fine and contains a file named after the SGX target.

I'm merging this now so that we get the improved failure tonight. Happy to continue iterating on the setup here.

@bors r+

Issue #3377, opened 2024-03-14
time-with-isolation test is drifting and often confuses people


saethlin:
On my machine, with a silly `for _ in 0..100_000 {}`, the isolated clock already runs at about half the speed of the actual clock. That's probably somewhat tolerable, but https://github.com/rust-lang/rust/issues/122420 is going to push us farther away from reality.

We might want to re-calibrate this magic value: https://github.com/rust-lang/miri/blob/b228407aae03eb01c658737a02caba925a6a6118/src/clock.rs#L4-L9

Or if our goal is really just an order of magnitude, we should make the test failure a lot more clear so reviewers/contributors are less confused by it.

PR #3379, opened 2024-03-14
explain time-with-isolation test better

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/3377

@saethlin do you think this is better?

### DIFF
--- a/tests/pass/shims/time-with-isolation.stdout
+++ b/tests/pass/shims/time-with-isolation.stdout
@@ -1,0 +1,2 @@
+The loop took around 7s
+(It's fine for this number to change when you `--bless` this test.)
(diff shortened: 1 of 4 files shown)


tests/pass/shims/time-with-isolation.stdout
```diff
@@ -0,0 +1,2 @@
+The loop took around 7s
+(It's fine for this number to change a bit when you `--bless` this test.)
```
saethlin, line 2: How much is "a bit"? The recent rust-lang/rust PR changed this number from 7 to 12 which is quite a large swing normally, but per the comment in the source code within an order of magnitude is fine.

So probably we should say "within an order of magnitude" here if that is our expectation.

RalfJung: We can also just say "it is okay to change"; we'll get pinged anyway when it happens so we can intervene if something seems to go terribly wrong.

saethlin: Works for me!

saethlin:
r=me

RalfJung:
@bors r=saethlin

(Btw, you should have "r+" powers here. :)

PR #3385, opened 2024-03-17
Report retags as distinct from real memory accesses for data races

Zoxc:
This changes the error reporting for data races such that reference invariants are no longer reported as real read and writes.

Before:
```
Data race detected between (1) non-atomic write on thread `unnamed-6` and (2) non-atomic read on thread `unnamed-5` at alloc1034971+0x10c. (2) just happened here
```

After:
```
Data race detected between (1) non-atomic write on thread `unnamed-8` and (2) shared reference invariant on thread `unnamed-6` at alloc1018329+0x190. (2) just happened here
```

Non-atomic read accesses from the *other* thread don't have this information tracked so those are called `some potential non-atomic read access` here.


Zoxc:
Should I change the race tests to look for `race detected`? Matching the entire string as they currently do seems quite redundant.

RalfJung:
This is great, I was just thinking that we should finally tackle this issue. :)

In terms of interface, I would suggest to simply extend the existing `read` with an `NaReadType` parameter. I don't quite see why we'd want functions like `assert_immutable_memory`, that just makes the API surface one has to understand so much bigger. (I see for `write` we have this internal `unique_access` function; I think it'd be fine to remove this and just rename `unique_access` to `write` and expose that. In fact `na_read` and `na_write` or `non_atomic_read`/`non_atomic_write` would probably be better.)

In terms of wording, "shared reference invariant on thread" does not make a lot of sense to me. I would say something like "implicit non-atomic read due to retag". (Cc @saethlin in case you have better ideas)

> Should I change the race tests to look for race detected? Matching the entire string as they currently do seems quite redundant.

The point of these tests is to ensure that we call out the right thread names and access types. So just saying "race detected" would risk regressions of those aspects.

I am surprised that there are so many test failures, even for tests that involve actual accesses.  Take for instance the classic `read_write_race.rs`:
```
  -error: Undefined Behavior: Data race detected between (1) non-atomic read on thread `unnamed-ID` and (2) non-atomic write on thread `unnamed-ID` at ALLOC. (2) just happened here
  +error: Undefined Behavior: Data race detected between (1) some potential non-atomic read access on thread `unnamed-ID` and (2) 4-byte non-atomic write on thread `unnamed-ID` at ALLOC. (2) just happened here
```
That is a completely different error. It now started mentioning the access size as well, and what is "some potential access"? There's a very clear non-atomic read access in the source. So something is definitely wrong with the diagnostics here.

Zoxc:
> implicit non-atomic read due to retag

I do find calling it a read quite confusing. I did initially go looking for a *real* read operation when I found a race, which is what promted the PR.

> The point of these tests is to ensure that we call out the right thread names and access types. So just saying "race detected" would risk regressions of those aspects.

The usage in rustc is more to have the `.stderr` catch such regressions and the test source code just ensures the right kind and location.

> It now started mentioning the access size as well

I fixed that now.

> and what is "some potential access"

Maybe calling it "read or retag" is better? Tracking the information on reads would be nicer, but I don't know where to do that.

RalfJung:
> I do find calling it a read quite confusing. I did initially go looking for a real read operation when I found a race, which is what promted the PR.

This is the common terminology. These *are* reads, they are just implicit side-effects of retags rather than explicit in the source.

Calling it an "invariant" doesn't even make semantic sense. An invariant is a property that is maintained throughout execution.  We're not talking about a property here but about an operation.

> The usage in rustc is more to have the .stderr catch such regressions and the test source code just ensures the right kind and location.

I don't agree, in rustc we alsu use the annotations to ensure that the right things show up in the message. Otherwise we'd just use `//~ERROR` and never `//~ERROR: message`.

> Maybe calling it "read or retag" is better? Tracking the information on reads would be nicer, but I don't know where to do that.

Why does "unknown" even happen? "read or retag" will be even more confusing for the common case where it is a regular read.

Zoxc:
> they are just implicit side-effects of retags rather than explicit in the source.

Maybe calling them *speculative reads* would be better?

> Why does "unknown" even happen?

We only know the read type on the current thread, not on the other conflicting operation, so I wanted a hint to the user that it could be either.

RalfJung:
> Maybe calling them speculative reads would be better?

Hm... that sounds a bit like what CPUs are doing, which is unrelated. OTOH these reads *are* indeed related to the compiler's ability to insert read (or write) operations. So maybe the term makes sense. But then we should also use it for the aliasing errors (eventually, not in this PR).

@rust-lang/miri any opinions on terminology here?

> We only know the read type on the current thread, not on the other conflicting operation, so I wanted a hint to the user that it could be either.

That sounds like it could cause a lot of confusion. Is there any way we can store that one bit of information (implicit/speculative read vs explicit read) somehow so that it is available later when computing the error? We already store an entire span, storing one more bit can't be so bad...

Zoxc:
> Hm... that sounds a bit like what CPUs are doing, which is unrelated. OTOH these reads _are_ indeed related to the compiler's ability to insert read (or write) operations. So maybe the term makes sense

*Speculative retag reads* or *retag reads* could also be options that relates them back more to the borrow model.

> Is there any way we can store that one bit of information (implicit/speculative read vs explicit read) somehow so that it is available later when computing the error?

Very likely, but I wasn't able to locate where it would go. The span logic seemed quite weird to me with it's handling of dummy spans too.

RalfJung:
"retag read" also works for me, though using "retag" as an adjective seems a bit odd.

> Very likely, but I wasn't able to locate where it would go. The span logic seemed quite weird to me with it's handling of dummy spans too.

`VTimestamp` probably needs a `bool` that indicates whether the read is "real" or not. That should likely be packed into `time` to avoid making it 4 bytes bigger.

saethlin:
I think our diagnostics should be informative for people who were completely unaware of this property of references until seeing the error. If we call them something like "a speculative retag read" we should provide a link or some other kind of reference to documentation that explains what that terminology refers to, or a separate `help:` message that says something like:
```
help: Retags permit optimizations that insert speculative (or hoisted) reads/writes. Therefore from the perspective of aliasing and data race invariants, a retag has the same implications as a read/write.
```

RalfJung:
Tweaking the wording of that message:

> help: Retags permit optimizations that insert speculative (or loop-hoisted) reads/writes. Therefore from the perspective of aliasing and data race models, a retag has the same implications as a read/write.

Or maybe "restrictions" instead of "models"? I don't like the use of "invariants" there, that's a specific technical term and this is too far outside its regular meaning.

Zoxc:
> "retag read" also works for me, though using "retag" as an adjective seems a bit odd.

Could we put an adjective in front of "retag" instead? "shared retag" and "unique retag"? That makes it more clear that there's no correspondence to reads in the source code.

RalfJung:
Note that whether unique retags are reads or writes depends on the aliasing model (in SB they are writes, in TB reads).

Is the idea that you would the entirely relegate the explanation that "shared retag" means a form of read into the `help`? That would also work for me.

Further editing on the help text, since it is specifically for data races so I think we can remove the "aliasing model" part:

> help: Retags permit optimizations that insert speculative (or loop-hoisted) reads/writes. Therefore from the perspective of data races, a retag has the same implications as a read/write.

Zoxc:
> Note that whether unique retags are reads or writes depends on the aliasing model (in SB they are writes, in TB reads).

I stuck with `retag read` and `retag write` then. It's useful from the data race perspective to know if they're reads or writes.

saethlin:
I like the current wording.

tests/fail/both_borrows/retag_data_race_write.stack.stderr
```diff
@@ -1,14 +1,15 @@
...
 LL |         *p = 5;
-   |         ^^^^^^ Data race detected between (1) non-atomic write on thread `unnamed-ID` and (2) non-atomic write on thread `unnamed-ID` at ALLOC. (2) just happened here
+   |         ^^^^^^ Data race detected between (1) retag write on thread `unnamed-ID` and (2) non-atomic write on thread `unnamed-ID` at ALLOC. (2) just happened here
    |
 help: and (1) occurred earlier here
   --> $DIR/retag_data_race_write.rs:LL:CC
    |
 LL |         let _r = &mut *p;
    |                  ^^^^^^^
+   = help: retags permit optimizations that insert speculative (or loop-hoisted) reads/writes. Therefore from the perspective of data races, a retag has the same implications as a read/write.
```
RalfJung, line 12: One possible issue here is that the user probably won't know what a "retag" is. For aliasing model messages at least there is context implying it's probably an aliasing thing, but here this is a data race message and people won't expect aliasing model terms.

Not sure what to do about that. We could either use a less precise but more common term, such as "reborrow", or we have to say at least a few words about what a retag is.

src/concurrency/vector_clock.rs
```diff
@@ -50,21 +52,39 @@ const SMALL_VECTOR: usize = 4;
 /// so that diagnostics can report what code was responsible for an operation.
 #[derive(Clone, Copy, Debug)]
 pub struct VTimestamp {
-    time: u32,
+    // The lowest bit indicates read type, the rest is the time.
```
RalfJung, line 55:
```suggestion
    /// The lowest bit indicates read type, the rest is the time.
    /// `1` indicates a retag read, `0` a regular read.
```

src/concurrency/vector_clock.rs
```diff
@@ -50,21 +52,39 @@ const SMALL_VECTOR: usize = 4;
...
         self.span.data()
     }
+
+    pub fn read_type(&self) -> NaReadType {
+        if self.time_and_read_type & 1 == 0 { NaReadType::Read } else { NaReadType::Retag }
+    }
+
+    pub fn set_read_type(&mut self, read_type: NaReadType) {
+        self.time_and_read_type = (self.time_and_read_type & !1) | read_type as u32
+    }
```
RalfJung, line 82: This seems to make some non-local assumptions about `Read` casting to 0 and `Retag` to 1? IMO it'd be better to keep the logic for int-to-NaReadType and the other direction next to each other to reduce the risk of them getting out-of-sync.

src/concurrency/vector_clock.rs
```diff
@@ -50,21 +52,39 @@ const SMALL_VECTOR: usize = 4;
...
-    pub const ZERO: VTimestamp = VTimestamp { time: 0, span: DUMMY_SP };
+    pub const ZERO: VTimestamp = VTimestamp { time_and_read_type: 0, span: DUMMY_SP };
+
+    fn time(&self) -> u32 {
+        self.time_and_read_type.shr(1)
+    }
+
+    fn set_time(&mut self, time: u32) {
+        self.time_and_read_type =
+            (self.time_and_read_type & 1) | time.checked_shl(1).expect("Vector clock overflow")
```
RalfJung, line 69: This just checks whether `1 >= 32`. What we actually want to check is whether we are losing a bit here.

src/concurrency/vector_clock.rs
```diff
@@ -448,7 +467,13 @@ mod tests {
         while let Some(0) = slice.last() {
             slice = &slice[..slice.len() - 1]
         }
-        VClock(slice.iter().copied().map(|time| VTimestamp { time, span: DUMMY_SP }).collect())
+        VClock(
+            slice
+                .iter()
+                .copied()
+                .map(|time| VTimestamp { time_and_read_type: time.shl(1), span: DUMMY_SP })
```
RalfJung, line 474: Please don't duplicate the encoding logic here, add a constructor instead or use `set_time`.

src/concurrency/data_race.rs
```diff
@@ -205,13 +205,36 @@ enum AtomicAccessType {
 /// are all treated as writes for the purpose
 /// of the data-race detector.
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
-enum NaWriteType {
+pub enum NaReadType {
+    /// Standard unsynchronized write.
+    Read,
+
+    Retag,
```
RalfJung, line 212: The doc comment for `Read` seems to have a copy-paste mistake, and please also add one for `Retag`.

src/concurrency/data_race.rs
```diff
@@ -205,13 +205,36 @@ enum AtomicAccessType {
 /// are all treated as writes for the purpose
 /// of the data-race detector.
```
RalfJung, line 206: That's the NaWriteType doc comment which is now attached to NaReadType.

src/concurrency/data_race.rs
```diff
@@ -205,13 +205,36 @@ enum AtomicAccessType {
...
+/// of the data-race detector.
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub enum NaWriteType {
     /// Allocate memory.
     Allocate,
 
     /// Standard unsynchronized write.
     Write,
 
+    Retag,
```
RalfJung, line 236: Please add doc comment

src/concurrency/data_race.rs
```diff
@@ -885,6 +930,7 @@ impl VClockAlloc {
         access: AccessType,
         access_size: Size,
         ptr_dbg: Pointer<AllocId>,
+        ty: Option<Ty<'_>>,
```
RalfJung, line 933: Please extend the doc comment to explain the meaning of `ty`.

src/concurrency/data_race.rs
```diff
@@ -981,6 +1020,8 @@ impl VClockAlloc {
         &self,
         alloc_id: AllocId,
         access_range: AllocRange,
+        read_type: NaReadType,
+        ty: Option<Ty<'_>>,
```
RalfJung, line 1024: Please explain the meaning of this parameter in the doc comment, and explicitly say that it is for diagnostics only.

src/concurrency/data_race.rs
```diff
@@ -1012,12 +1054,16 @@ impl VClockAlloc {
         }
     }
 
-    // Shared code for detecting data-races on unique access to a section of memory
-    fn unique_access<'tcx>(
+    /// Detect data-races for an unsynchronized write operation, will not perform
+    /// data-race threads if `race_detecting()` is false, either due to no threads
```
RalfJung, line 1058:
```suggestion
    /// Detect data-races for an unsynchronized write operation. Will not perform
    /// data-race detection if `race_detecting()` is false, either due to no threads
```

tests/fail/stacked_borrows/retag_data_race_protected_read.stderr
```diff
@@ -1,14 +1,15 @@
-error: Undefined Behavior: Data race detected between (1) non-atomic read on thread `main` and (2) non-atomic write on thread `unnamed-ID` at ALLOC. (2) just happened here
+error: Undefined Behavior: Data race detected between (1) non-atomic read on thread `main` and (2) retag write with type `i32` on thread `unnamed-ID` at ALLOC. (2) just happened here
   --> $DIR/retag_data_race_protected_read.rs:LL:CC
    |
 LL |         retag(unsafe { &mut *ptr.0 });
-   |                        ^^^^^^^^^^^ Data race detected between (1) non-atomic read on thread `main` and (2) non-atomic write on thread `unnamed-ID` at ALLOC. (2) just happened here
+   |                        ^^^^^^^^^^^ Data race detected between (1) non-atomic read on thread `main` and (2) retag write with type `i32` on thread `unnamed-ID` at ALLOC. (2) just happened here
```
RalfJung, line 5: Should it maybe say "of type" instead of "with type"? That sounds more like proper grammar to me, but I'm not entirely sure.

RalfJung (commented):
This looks great, thanks! We're getting into the fine-tuning phase.

src/concurrency/data_race.rs
Zoxc: What should I add here? I'm not too familiar with retags.

RalfJung: Something like "An implicit write generated by a retag" would work.

src/concurrency/data_race.rs
```diff
@@ -200,18 +200,40 @@ enum AtomicAccessType {
     Rmw,
 }
 
+/// Type of read operation
```
RalfJung, line 203:
```suggestion
/// Type of a non-atomic read operation.
```

src/concurrency/vector_clock.rs
```diff
@@ -52,25 +52,33 @@ const SMALL_VECTOR: usize = 4;
...
     }
 
     fn set_time(&mut self, time: u32) {
         self.time_and_read_type =
-            (self.time_and_read_type & 1) | time.checked_shl(1).expect("Vector clock overflow")
-    }
-
-    pub fn span_data(&self) -> SpanData {
-        self.span.data()
+            (self.time_and_read_type & 1) | time.checked_mul(2).expect("Vector clock overflow")
```
RalfJung, line 81: That's a bit too clever IMO. At least it needs clear comments.

src/concurrency/vector_clock.rs
```diff
@@ -52,25 +52,33 @@ const SMALL_VECTOR: usize = 4;
...
+
+    const fn new(time: u32, read_type: NaReadType, span: Span) -> Self {
+        let read_type = match read_type {
+            NaReadType::Read => 0,
+            NaReadType::Retag => 1,
+        };
+        Self {
+            time_and_read_type: read_type | time.checked_mul(2).expect("Vector clock overflow"),
+            span,
+        }
```
RalfJung, line 72: Is there any way to avoid duplicating the logic with the `set` functions?

Zoxc: I could call `new` in them. It might result in code for dealing with the span though.

src/concurrency/vector_clock.rs
```diff
@@ -448,7 +480,13 @@ mod tests {
         while let Some(0) = slice.last() {
             slice = &slice[..slice.len() - 1]
         }
-        VClock(slice.iter().copied().map(|time| VTimestamp { time, span: DUMMY_SP }).collect())
+        VClock(
+            slice
+                .iter()
+                .copied()
+                .map(|time| VTimestamp::new(time, NaReadType::Read, DUMMY_SP))
```
RalfJung, line 487: Maybe you can avoid even having `new` by doing something like
```rust
let mut clock = VTimestamp::ZERO;
clock.set_time(time);
clock
```

src/concurrency/data_race.rs
```diff
@@ -219,10 +219,8 @@ impl NaReadType {
     }
 }
 
-/// Type of write operation: allocating memory
-/// non-atomic writes and deallocating memory
-/// are all treated as writes for the purpose
-/// of the data-race detector.
+/// Type of a non-atomic of write operation: allocating memory non-atomic writes and
+/// deallocating memory are all treated as writes for the purpose of the data-race detector.
```
RalfJung, line 223:
```suggestion
/// Type of a non-atomic write operation: allocating memory, non-atomic writes, and
/// deallocating memory are all treated as writes for the purpose of the data-race detector.
```

src/concurrency/vector_clock.rs
```diff
@@ -77,16 +79,15 @@ impl VTimestamp {
     }
 
     fn set_time(&mut self, time: u32) {
-        self.time_and_read_type =
-            (self.time_and_read_type & 1) | time.checked_mul(2).expect("Vector clock overflow")
+        self.time_and_read_type = Self::new(time, self.read_type(), DUMMY_SP).time_and_read_type;
```
RalfJung, line 82: Please define an `encode_time_and_read_type` function, that seems better than creating a full `VTimestamp` and then only reading one field and avoids the dummy span.

tests/fail/both_borrows/retag_data_race_write.stack.stderr
Zoxc: I'd probably add a short explanation of what a retag is. When would they not correspond to a reborrow?

RalfJung: Retags happen (almost) every time a value of reference is assigned. So e.g. if `x` has type `(&i32, &i32)` then passing that to a function retags both references. If `y` has type `&(&i32, &i32)` then `let x = *y;` also retags both references in the pair.

src/concurrency/vector_clock.rs
```diff
@@ -61,33 +61,33 @@ pub struct VTimestamp {
 impl VTimestamp {
     pub const ZERO: VTimestamp = VTimestamp::new(0, NaReadType::Read, DUMMY_SP);
 
-    const fn new(time: u32, read_type: NaReadType, span: Span) -> Self {
-        let read_type = match read_type {
+    const fn encode_time_and_read_type(time: u32, read_type: NaReadType) -> u32 {
```
RalfJung, line 64: May be worth marking this as `#[inline]` just to be safe...

tests/fail/both_borrows/retag_data_race_write.stack.stderr
Zoxc: Maybe something like `retags happens due to explicit or implicit borrowing`?

Does the initial borrow also result in a retag? Presumably those could be involved in data races too.

RalfJung: Yes, every execution of `&`/`&mut` expressions results in a retag.

RalfJung: So... maybe you can just add another line which says something like
```
help: retags occur on all (re)borrows and as well as when references are copied/moved
```

src/diagnostics.rs
```diff
@@ -271,7 +271,9 @@ pub fn report_error<'tcx, 'mir>(
                     helps.push((None, format!("see https://doc.rust-lang.org/nightly/std/sync/atomic/index.html#memory-model-for-atomic-accesses for more information about the Rust memory model")));
                 }
                 if *retag_explain {
-                    helps.push((None, "retags permit optimizations that insert speculative (or loop-hoisted) reads/writes. Therefore from the perspective of data races, a retag has the same implications as a read/write.".to_owned()));
+                    helps.push((None, "retags occur on all (re)borrows and as well as when references are copied or moved. \
+                        Retags permit optimizations that insert speculative reads or writes. \
+                        Therefore from the perspective of data races, a retag has the same implications as a read or write.".to_owned()));
```
RalfJung, line 276:
```suggestion
                    helps.push((None, "retags permit optimizations that insert speculative reads or writes. \
                        Therefore, a retag has the same implications as a read or write from the perspective of data races.".to_owned()));
                    helps.push((None, "retags occur on all (re)borrows and as well as when references are copied or moved".to_owned()));
```

RalfJung:
Just one final tweak of the message. Then please squash all the commits.

src/diagnostics.rs
Zoxc: Should the lines be split to one help message each? Why would `why is a retag?` be answered before `what is a retag?`?

RalfJung: I was thinking we would start with "why is this a race" -- because of speculative reads/writes. Then we can explain where those come from. But you are right, the other order would also make sense.

Splitting into 3 messages also works for me, I just felt it was too long for 1.^^

RalfJung:
Looks good, thanks for working through all these revisions with me. :)
Please squash, then we can land this.

PR #3387, opened 2024-03-17
make 'invalidate' benchmark shorter

RalfJung:
This is currently by far the slowest benchmark in our suite, taking >9s, when the second slowest takes 2.7s. So let's speed this up to 2.3s, making it still the second-slowest in the benchmark suite.

@saethlin any objections? Also, why is this called "invalidate"? It got added in https://github.com/rust-lang/miri/pull/3083 but I can't figure out the point of that name even after looking at the PR.^^ There should be a comment in the benchmark explaining what it is testing.

### DIFF
--- a/bench-cargo-miri/invalidate/src/main.rs
+++ b/bench-cargo-miri/invalidate/src/main.rs
@@ -1,4 +1,4 @@
 fn main() {
     // The end of the range is just chosen to make the benchmark run for a few seconds.
-    for _ in 0..200_000 {}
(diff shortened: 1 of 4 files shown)


saethlin:
:thinking: We should probably just rename the benchmark something like `range-iteration`. I think the name is odd because I reduced it from something else. The benchmark exists now to generate a lot of work for the AllocId part of the GC.

RalfJung:
Makes sense, thanks!

@bors r+

Issue #2423, opened 2022-07-23
`compile_fail` doctests that rely on post-monomorphization errors don't pass `miri test`


AngelicosPhosphoros:
Example code:

```rust
/// You cannot send arrays with different sizes:
/// ```compile_fail
/// // You need to use your package name.
/// use reproduce_miri::G;
/// G::i_am_break_on_different_array_sizes([0u8;5], [0u32;3]);
/// ```
pub struct G;

pub trait IAmArray{
    const SIZE: usize;
}

impl<T, const N: usize> IAmArray for [T; N]{
    const SIZE: usize = N;
}

impl G{
    pub fn i_am_break_on_different_array_sizes<A, B>(_a: A, _b: B)
        where A: IAmArray,
            B: IAmArray,
    {
        trait CompileAssert{
            const TRIGGER: ();
        }
        impl<A, B> CompileAssert for (A, B)
            where A: IAmArray,
                B: IAmArray,
        {
            const TRIGGER: () = if A::SIZE == B::SIZE {} else {panic!("You must provide arrays of same length")};
        }
        
        let _ = <(A, B) as CompileAssert>::TRIGGER;
    }
}
```

Using `cargo +nightly test`:
<details><summary>Output</summary>

```
> cargo +nightly  test
   Compiling reproduce_miri v0.1.0 (E:\Programs\Rust\reproduce_miri)
    Finished test [unoptimized + debuginfo] target(s) in 0.72s
     Running unittests src\lib.rs (target\debug\deps\reproduce_miri-632b99990da55433.exe)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

   Doc-tests reproduce_miri

running 1 test
test src\lib.rs - G (line 2) - compile fail ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.10s
```

</details>

Using `cargo +nightly miri test`
<details><summary>Output</summary>

```
> cargo +nightly miri test
   Compiling reproduce_miri v0.1.0 (E:\Programs\Rust\reproduce_miri)
    Finished test [unoptimized + debuginfo] target(s) in 0.17s
     Running unittests src\lib.rs (target\miri\x86_64-pc-windows-msvc\debug\deps\reproduce_miri-40824facde3e8e37.exe)

running 0 tests

[...]

PR #3392, opened 2024-03-19
fix compile_fail doctests with post-mono errors

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/2423

### DIFF
--- a/src/bin/miri.rs
+++ b/src/bin/miri.rs
@@ -179,6 +179,26 @@ impl rustc_driver::Callbacks for MiriBeRustCompilerCalls {
             });
         }
     }
+
+    fn after_analysis<'tcx>(
+        &mut self,
+        _: &rustc_interface::interface::Compiler,
+        queries: &'tcx rustc_interface::Queries<'tcx>,
+    ) -> Compilation {
+        queries.global_ctxt().unwrap().enter(|tcx| {
+            if self.target_crate {
+                // cargo-miri has patched the compiler flags to make these into check-only builds,
+                // but we are still emulating regular rustc builds, which would perform post-mono
+                // const-eval during collection. So let's also do that here, even if we might be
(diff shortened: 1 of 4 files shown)


src/bin/miri.rs
```diff
@@ -179,6 +196,21 @@ impl rustc_driver::Callbacks for MiriBeRustCompilerCalls {
...
+    fn after_analysis<'tcx>(
+        &mut self,
+        _: &rustc_interface::interface::Compiler,
+        queries: &'tcx rustc_interface::Queries<'tcx>,
+    ) -> Compilation {
+        queries.global_ctxt().unwrap().enter(|tcx| {
+            // Make sure compile_fail tests with post-monomorphization const-eval errors work as
+            // intended in rustdoc mode.
+            if self.mode == MiriBeRustcMode::Rustdoc {
+                let _ = tcx.collect_and_partition_mono_items(());
```
RalfJung, line 209: So... should we maybe run this always, not just when we are rustdoc? Maybe even when we are Miri and running the program? That makes our builds slower, but it also makes us behave more like real rustc.

RalfJung: I can't measure a difference for sysroot builds, even though those are non-trivial crates. So I propose we run `collect_and_partition_mono_items` at least for MIRI_BE_RUSTC mode.

I'm on the fence about whether the interpreter driver itself should do this as well.

RalfJung:
Hm, judging from the failure in https://github.com/rust-lang/rust/pull/122744, it seems like calling `collect_and_partition_mono_items` when the dependencies are metadata-only may run into problems...

RalfJung:
> We won't run into that specific problem as we -Zalways-emit-mir

Ah!

But only for target crates. Let me adjust the PR.
@bors r-

PR #3393, opened 2024-03-21
Allow `llvm.x86.sse2.pause` instrinsic to be called without SSE2

eduardosm:
The instrinsic is compiled to a `pause` instruction, which behaves like a no-op when SSE2 is not available.

https://www.felixcloutier.com/x86/pause.html

### DIFF
--- a/tests/pass/intrinsics-x86-pause-without-sse2.rs
+++ b/tests/pass/intrinsics-x86-pause-without-sse2.rs
@@ -1,0 +1,25 @@
+// Ignore everything except x86 and x86_64
+// Any new targets that are added to CI should be ignored here.
+// (We cannot use `cfg`-based tricks here since the `target-feature` flags below only work on x86.)
+//@ignore-target-aarch64
+//@ignore-target-arm
+//@ignore-target-avr
+//@ignore-target-s390x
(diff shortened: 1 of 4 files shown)


eduardosm:
This probably had regressed with https://github.com/rust-lang/miri/pull/3180

tests/pass/intrinsics-x86-pause-without-sse2.rs
```diff
@@ -0,0 +1,26 @@
...
+// Any new targets that are added to CI should be ignored here.
+// (We cannot use `cfg`-based tricks here since the `target-feature` flags below only work on x86.)
+//@ignore-target-aarch64
+//@ignore-target-arm
+//@ignore-target-avr
+//@ignore-target-s390x
+//@ignore-target-thumbv7em
+//@ignore-target-wasm32
+// SSSE3 implicitly enables SSE3
+//@compile-flags: -C target-feature=-sse2
```
RalfJung, line 11: What's the connection between the comment and the line that follows?

eduardosm: copy-paste leftover

RalfJung:
Thanks!
@bors r+

Issue #3179, opened 2023-11-20
Add direct tests for libc FS functions like "ftruncate"


RalfJung:
"ftruncate" (and really, most of the file system APIs) is currently only tested via the libstd functions, but not via direct libc calls. With PRs like https://github.com/rust-lang/miri/pull/3177 that add support for ftruncate on targets where libstd uses other functions (like ftruncate64), it'd be good to have direct tests as well.

RalfJung:
Right, basically these tests should be fairly similar to [this one](https://github.com/rust-lang/miri/blob/master/tests/pass/shims/fs.rs) but directly using the libc APIs rather than calling Rust's `std::fs` functions.

You can use the `CString` type to generate C-compatible strings, no need to manually push `\0`.

RalfJung:
Support for extra modes in `open` (as implemented in https://github.com/rust-lang/miri/pull/3240) should also be covered by such a test.

eduardorittner:
I looked over all `foreign_item.rs`s and for each function there looked for corresponding direct libc tests, these are the functions which do not have any direct tests associated, categorized by where they are located in the source code. If I found a single instance of a direct function call in tests, I immediately discarded it, so it is possible for functions not listed here to have been called only once in a test and not have sufficient test coverage. I also did not include any functions which check for `this.frame_in_std()`.

# Shared (`foreign_items.rs`)
- [ ] `exit`
- [ ] `abort`
- [x] `memcmp`

# Unix (`unix/foreign_items.rs`)

- [x] `getenv`
- [x] `setenv`
- [x] `unsetenv`
- [ ] `getcwd`
- [ ] `chdir`
- [ ] `realpath`
- [x] `pread`
- [x] `pwrite`
- [ ] `lseek`
- [x] `fcntl`
- [x] `fstat`
- [x] `stat`
- [x] `lstat`
- [ ] `unlink`
- [ ] `symlink`
- [ ] `mkdir`
- [ ] `rmdir`
- [ ] `opendir`
- [ ] `readdir`
- [ ] `closedir`
- [ ] `fsync`
- [ ] `fdatasync`
- [ ] `aligned_alloc`
- [ ] `pthread_key_delete`
- [ ] `pthread_getspecific`
- [ ] `pthread_setspecific`
- [ ] `sched_getaffinity`
- [ ] `pthread_atfork`
- [x] `arc4random_buf`
- [x] `_Unwind_RaiseException`

## Android (`unix/android/foreign_items.rs`)

[...]

PR #3398, opened 2024-03-23
Add libc direct test for shims and update CONTRIBUTING.md

tiif:
Add more ``libc`` direct test for shims as mentioned in #3179.

Changes:
- Added ``libc`` direct tests for ``rename`` and ``ftruncate64``
- Added the command for running ``pass-dep`` test in ``CONTRIBUTING.md``

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,19 +64,22 @@ For example, you can (cross-)run the driver on a particular file by doing
 ./miri run tests/pass/hello.rs --target i686-unknown-linux-gnu
 ```
 
-and you can (cross-)run the entire test suite using:
+Tests in ``pass-dep`` need to be run using ``./miri run --dep <filename>``.  
+For example:
+```sh
+./miri run --dep tests/pass-dep/shims/libc-fs.rs
+```
+
+You can (cross-)run the entire test suite using:
 
 ```
 ./miri test
 MIRI_TEST_TARGET=i686-unknown-linux-gnu ./miri test
 ```
 
-If your target doesn't support libstd that should usually just work. However, if you are using a
-custom target file, you might have to set `MIRI_NO_STD=1`.
-
 `./miri test FILTER` only runs those tests that contain `FILTER` in their filename (including the
 base directory, e.g. `./miri test fail` will run all compile-fail tests). These filters are passed
-to `cargo test`, so for multiple filers you need to use `./miri test -- FILTER1 FILTER2`.
+to `cargo test`, so for multiple filters you need to use `./miri test -- FILTER1 FILTER2`.
 
 #### Fine grained logging
 
--- a/tests/pass-dep/shims/libc-fs.rs
+++ b/tests/pass-dep/shims/libc-fs.rs
@@ -16,6 +16,10 @@ mod utils;
 fn main() {
     test_dup_stdout_stderr();
     test_canonicalize_too_long();
+    test_rename();
+    test_ftruncate::<libc::off_t>(libc::ftruncate);
+    #[cfg(target_os = "linux")]
+    test_ftruncate::<libc::off64_t>(libc::ftruncate64);
     test_readlink();
     test_file_open_unix_allow_two_args();
     test_file_open_unix_needs_three_args();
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -71,6 +71,12 @@ and you can (cross-)run the entire test suite using:
 MIRI_TEST_TARGET=i686-unknown-linux-gnu ./miri test
 ```
 
+Tests in ``pass-dep`` need to be run using ``./miri run --dep <filename>``.  
+For example:
+```sh
+./miri run --dep tests/pass-dep/shims/libc-fs.rs
```
RalfJung, line 77: In terms of ordering this now first explains `./miri run`, then `./miri test`, then `./miri run --dep`. That makes little sense. Can you move the new paragraph up so that it is right after `./miri run`?

CONTRIBUTING.md
```diff
@@ -71,6 +71,12 @@ and you can (cross-)run the entire test suite using:
...
+```
+
 If your target doesn't support libstd that should usually just work. However, if you are using a
 custom target file, you might have to set `MIRI_NO_STD=1`.
```
RalfJung, line 81: This can be removed, no-std is determined automatically these days.

CONTRIBUTING.md
tiif: Sure, this is indeed easier to read.

tests/pass-dep/shims/libc-fs.rs
```diff
@@ -133,6 +135,62 @@ fn test_readlink() {
...
+
+    // Renaming a nonexistent file should fail
+    let res = unsafe { libc::rename(c_path1.as_ptr(), c_path2.as_ptr()) };
+    assert_eq!(res, -1);
+    assert_eq!(Error::last_os_error().kind(), ErrorKind::NotFound);
+
+    remove_file(&path2).unwrap();
+}
+
+fn test_ftruncate() {
```
RalfJung, line 163: To test both ftruncate and ftruncate64 without duplicating the test, you can follow the same approach as `tests/pass-dep/shims/mmap.rs` and give this function a fn ptr of what function it should call to truncate.

tiif: Oh, that's neat! I assume this approach can also be used to prevent macos from running ``ftruncate64``. 
```rust
test_mmap(libc::mmap);
#[cfg(target_os = "linux")]
test_mmap(libc::mmap64);

```

RalfJung: Yes, exactly.

tiif:
May I know how to generate the error presented in CI locally? I tested with the command below and no error is shown.  
```
./miri run --dep ./tests/pass-dep/shims/libc-fs.rs -Zmiri-disable-isolation --target x86_64-apple-darwin
./miri run --dep ./tests/pass-dep/shims/libc-fs.rs -Zmiri-disable-isolation --target x86_64-unknown-linux-gnu

```

tests/pass-dep/shims/libc-fs.rs
```diff
@@ -160,7 +162,7 @@ fn test_rename() {
     remove_file(&path2).unwrap();
 }
 
-fn test_ftruncate() {
+fn test_ftruncate(ftruncate: unsafe extern "C" fn(fd: libc::c_int, length: i64) -> libc::c_int) {
```
RalfJung, line 165: You need to make the `length` type generic; it is not the same between the two functions.

tiif: I see, thanks!

RalfJung:
> May I know how to generate the error presented in CI locally?

You have to check which target the tests are failing for. If you scroll up from the test failure you will see
```
     Compiler: "MIRI_ENV_VAR_TEST"="0" "MIRI_TEMP"="/tmp/miri-uitest-1hWALp" /home/runner/work/miri/miri/target/debug/miri "--error-format=json" "-Dwarnings" "-Dunused" "-Ainternal_features" "-Zmiri-provenance-gc=1" "-Zui-testing" "--target" "i686-unknown-linux-gnu" "--out-dir" OUT_DIR
```
So the target is `i686-unknown-linux-gnu`.

tests/pass-dep/shims/libc-fs.rs
tiif: I might need more help on this. I am not too sure how can I make length generic. ``length`` can be either ``libc::off_t`` or ``libc::off64_t``. If I use:
```rust
fn test_ftruncate<T>(ftruncate: unsafe extern "C" fn(fd: libc::c_int, length: T) -> libc::c_int) {
```
then I will face problem casting ``i64`` to ``libc::off_t`` or ``libc::off64_t`` because the code below won't work. 
```rust
length:i64 = 5;
length_casted = 5 as T;
```
``Default`` used by ``tests/pass-dep/shims/mmap.rs`` is not applicable too as the value of length needs to be customised in the test.

RalfJung: `T: From<i64>` or `T: TryFrom<i64>` should work

tiif:
@RalfJung 
Ok, the CI is finally green. Please take a look whenever you find it convenient, thanks!

tests/pass-dep/shims/libc-fs.rs
```diff
@@ -193,3 +258,20 @@ fn test_posix_mkstemp() {
         assert_eq!(e.kind(), std::io::ErrorKind::InvalidInput);
     }
 }
+
+// Trait implemented for test_ftruncate to convert i64 to the provided type
+trait Length: Sized {
```
RalfJung, line 263: Why are you defining a new trait for this? The standard library has `From<i32>` that you should be able to directly use.

tiif: Yes, ``From<i32>`` works. I haven't found a correct way to cast ``i64`` to ``i32`` so I resorted to implementing traits. Thanks for pointing this out.

RalfJung:
LGTM, thanks!
Please squash the commits.

tiif:
Sorry for doing force push, this is my first time using squash, so I am not sure why the push is rejected after squashing. Thank you for your guidance and multiple rounds of reviews for this PR, I really learned a lot! I will work on implementing some easier shim next.

RalfJung:
Force-pushing is expected after squashing. :) This looks right, thanks for the PR!
@bors r+

PR #3399, opened 2024-03-23
make CI failures easier to interpret

RalfJung:
RUST_BACKTRACE=1 means we show 2 backtraces that are almost always irrelevant: from ui_test and from the miri script. Instead let's set the env var inside the test harness so we see backtraces of Miri ICEing but nothing else.

### DIFF
--- a/tests/ui.rs
+++ b/tests/ui.rs
@@ -4,6 +4,7 @@ use std::ffi::OsString;
 use std::num::NonZeroUsize;
 use std::path::{Path, PathBuf};
 use std::{env, process::Command};
+use ui_test::color_eyre::eyre::Context;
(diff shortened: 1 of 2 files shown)


RalfJung:
Oh lol, I force-pushed after approval and bors picked that up...
just in case: @bors r+

tests/ui.rs
```diff
@@ -223,14 +227,15 @@ fn ui(
     with_dependencies: Dependencies,
     tmpdir: &Path,
 ) -> Result<()> {
-    let msg = format!("## Running ui tests in {path} against miri for {target}");
+    let msg = format!("## Running ui tests in {path} for {target}");
```
RalfJung, line 230: The "against miri" part is a leftover from ancient times when we also ran these tests in rustc.

PR #3402, opened 2024-03-24
miri script: build with stable toolchain

RalfJung:
`./miri toolchain` sets up a `rustup override miri`. But then if something goes wrong and the `miri` toolchain doesn't work, one can't even run `./miri toolchain` again as building miri-script needs a toolchain...

So let's always use stable to build miri-script, making it override-independent. I assume everyone will have that installed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,12 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Show Rust version (stable toolchain)
+        run: |
+          rustup show
+          rustc -Vv
+          cargo -V
+
       # Cache the global cargo directory, but NOT the local `target` directory which
       # we cannot reuse anyway when the nightly changes (and it grows quite large
       # over time).
@@ -57,7 +63,7 @@ jobs:
         if: ${{ steps.cache.outputs.cache-hit != 'true' }}
         run: cargo install -f rustup-toolchain-install-master
 
-      - name: Install "master" toolchain
+      - name: Install miri toolchain
         run: |
           if [[ ${{ github.event_name }} == 'schedule' ]]; then
             echo "Building against latest rustc git version"
@@ -65,13 +71,13 @@ jobs:
           fi
           ./miri toolchain --host ${{ matrix.host_target }}
 
-      - name: Show Rust version
+      - name: Show Rust version (miri toolchain)
         run: |
           rustup show
           rustc -Vv
           cargo -V
 
-      - name: Test
+      - name: Test Miri
         run: ./ci/ci.sh
 
   style:
--- a/ci/ci.sh
+++ b/ci/ci.sh
@@ -13,6 +13,15 @@ function endgroup {
 
 begingroup "Building Miri"
 
+# Special Windows hacks
+if [ "$HOST_TARGET" = i686-pc-windows-msvc ]; then
+  # The $BASH variable is `/bin/bash` here, but that path does not actually work. There are some
+  # hacks in place somewhere to try to paper over this, but the hacks dont work either (see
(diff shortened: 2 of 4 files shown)


RalfJung:
Why would this PR cause *that* test to fail on Windows...?!?
```
  $ bash -c "cargo miri run --locked --manifest-path \"D:\\a\\miri\\miri\\bench-cargo-miri\\backtraces\\Cargo.toml\""
  Windows Subsystem for Linux has no installed distributions.
  
  Distributions can be installed by visiting the Microsoft Store:
  
  https://aka.ms/wslstore
  
  Error: command exited with non-zero code `bash -c "cargo miri run --locked --manifest-path \"D:\\a\\miri\\miri\\bench-cargo-miri\\backtraces\\Cargo.toml\""`: 1
```
We are running our entire CI in bash.^^ But now somehow this one `bash` call is trying to invoke WSL?!?
Cc @ChrisDenton

ChrisDenton:
That could happen if it's picking up the bash in the system directory.

Maybe use `$BASH` instead?

RalfJung:
I'll try that, thanks. But why does adding `+stable` change anything?

ChrisDenton:
> I'll try that, thanks. But why does adding `+stable` change anything? are we getting an old version of Rust where `Command` behaved differently?

`Command` on Windows has some weirdly inconsistent behaviour depending on if `Command::env` is used or not. The short version is that in Rust's early history, an incomplete attempt was made to emulate Linux behaviour and over time this code rotted further as people refactored the code for modern idioms (lack of testing didn't help). I've improved the code since I started helping to maintain std but actually fixing the behaviour needs a decision on what exactly the new behaviour should be.

RalfJung:
What is even going on? `"$BASH"` expands to `/bin/bash`  but then magically that becomes `C:/Program Files/Git/usr/bin/bash` which causes a problem because of the space. So I use `"'$BASH'"` instead but now the magic transformation does not happen and it tries to run `/bin/bash` which does not exist.

RalfJung:
Looks like they are using 1.76. Let's see if 1.77 fixes this... after all we've been using miri-script for a while now (but with nightly Rust) and it always worked on Windows.

@bors r+

RalfJung:
Nope, latest stable is still affected.
We may just have to disable this test on Windows for now...

RalfJung:
Hm, something doesn't make sense. We've been using this for at least 7 months now. Whatever was the nightly back then, has long reached stable.

So this can't be just a difference between Rust stable vs nightly. It has to be that somehow adding the `+stable` itself is the problem...?!?

RalfJung:
Well, hard-coding the path seems to work. :shrug: 

@bors r+

RalfJung:
Yeah something is definitely patching the env var -- but only if it is *not* surrounded by `'` or `"`, which leads to very bad behavior. What a terrible hack, and I don't even know where it lives.
```
2024-03-24T14:59:55.2091474Z + HYPERFINE='/bin/bash -c'
2024-03-24T14:59:55.2091893Z + ./miri bench
2024-03-24T14:59:55.2092838Z ##[group]Testing host architecture
2024-03-24T14:59:55.3781009Z [src\commands.rs:399:9] program_name = "C:/Program"
2024-03-24T14:59:55.3781848Z [src\commands.rs:399:9] args = [
2024-03-24T14:59:55.3782357Z     "Files/Git/usr/bin/bash",
2024-03-24T14:59:55.3782915Z     "-c",
2024-03-24T14:59:55.3783142Z ]
```

RalfJung:
Seems to be [this functionality](https://stackoverflow.com/questions/7250130/how-to-stop-mingw-and-msys-from-mangling-path-names-given-at-the-command-line/34386471) that reportedly is specific to Git-for-Windows.

What remains completely unclear is why adding the `+stable` changed anything...

Issue #3405, opened 2024-03-24
Are our many-seeds tests failing on Windows?


RalfJung:
I just noticed this in our Windows CI logs:
```
  + ./miri many-seeds ./miri run tests/many-seeds/scoped-thread-leak.rs
  Trying seed: 0
  Failing seed: 0
```
There seem to be multiple things wrong here:
- Why is the test failing?
- Why is there no error message?
- Why does CI succeed anyway?

PR #3406, opened 2024-03-24
many-seeds: propagate failure properly

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/3405.


RalfJung:
Okay so now it indeed does fail on Windows. But not in a useful way:
```
 Error: io error when running command `./miri run tests/many-seeds/scoped-thread-leak.rs`: %1 is not a valid Win32 application. (os error 193)
```
@ChrisDenton can `Command` not spawn scripts on Windows? Would it work to run the batch script instead of the bash script?

ChrisDenton:
On Windows you can only reliably run exe files.

https://github.com/rust-lang/rust/issues/94743 has some more details (though ultimately adding support for scripts was not popular)

RalfJung:
Then either we can get bash to launch the script for us or we'll have to disable the test on Windows. I don't want to duplicate the logic from './miri' that computes where the binary is.

RalfJung:
This might work...
@bors r+

Issue #3404, opened 2024-03-24
`cargo miri test` is broken for doctests (Option 'sysroot' given more than once)


RalfJung:
https://github.com/rust-lang/rust/pull/122840 changed the way rustdoc passes arguments to rustc for doctest builds, which causes `cargo miri test` to break. The first symptom is output like
```
---- src/lib.rs - my_fn (line 37) stdout ----
error: Option 'sysroot' given more than once

Couldn't compile the test.
```
The deeper cause is that Miri does a lot of scanning and patching of rustc command-line arguments, and if those arguments are hidden in a file then everything breaks.

RalfJung:
`cargo miri test -v` shows
```
[cargo-miri rustc inside rustdoc] going to run:
MIRI_BE_RUSTC="target" "/home/r/.rustup/toolchains/miri/bin/miri" "@/tmp/rustdoctestAMWaIS/rustdoc-cfgs" "--edition" "2018" "-o" "/tmp/rustdoctestyUJi8S/rust_out.miri" "--emit=metadata" "--target" "x86_64-unknown-linux-gnu" "--color" "always" "-"
```

That arg file contains
```
--crate-type=bin
--cfg=miri
--sysroot=/home/r/.cache/miri
-Ldependency=/home/r/src/rust/miri/test-cargo-miri/target/miri/x86_64-unknown-linux-gnu/debug/deps
-Ldependency=/home/r/src/rust/miri/test-cargo-miri/target/miri/debug/deps
--extern=byteorder_2=/home/r/src/rust/miri/test-cargo-miri/target/miri/x86_64-unknown-linux-gnu/debug/deps/libbyteorder-ee60f6ecbd530bd3.rmeta
--extern=byteorder=/home/r/src/rust/miri/test-cargo-miri/target/miri/x86_64-unknown-linux-gnu/debug/deps/libbyteorder-f178d57ea71fe0c8.rmeta
--extern=cargo_miri_test=/home/r/src/rust/miri/test-cargo-miri/target/miri/x86_64-unknown-linux-gnu/debug/deps/libcargo_miri_test-abd1c36147f6f274.rmeta
--extern=exported_symbol=/home/r/src/rust/miri/test-cargo-miri/target/miri/x86_64-unknown-linux-gnu/debug/deps/libexported_symbol-6d3a18b7c769098c.rmeta
[...]

PR #3409, opened 2024-03-25
Rustup

RalfJung:
This uses a different approach to resolve https://github.com/rust-lang/miri/issues/3404: we entirely move the responsibility of setting miri-sysroot to whatever *invokes* the Miri driver.  cargo-miri knows whether it is inside rustdoc or not and can adjust accordingly. I previously avoided doing that because there are a bunch of places that are invoking the driver (cargo-miri, the ui test suite, `./miri run`, `./x.py run miri`) and they all need to be adjusted now. But it is also somewhat less fragile as we usually have more information there -- and we can just decide that `./miri run file.rs --sysroot path` is not supported.


RalfJung:
Closing in favor of https://github.com/rust-lang/miri/pull/3411 which is just the cargo-miri changes without the rustup. (That won't make it work, but it makes more sense to keep that around as a draft.)

PR #3411, opened 2024-03-25
Handle Miri sysroot entirely outside the Miri driver

RalfJung:
(Extracted from https://github.com/rust-lang/miri/pull/3409)

This entirely moves the responsibility of setting miri-sysroot to whatever *invokes* the Miri driver.  cargo-miri knows whether it is inside rustdoc or not and can adjust accordingly. I previously avoided doing that because there are a bunch of places that are invoking the driver (cargo-miri, the ui test suite, `./miri run`, `./x.py run miri`) and they all need to be adjusted now. But it is also somewhat less fragile as we usually have more information there -- and we can just decide that `./miri run file.rs --sysroot path` is not supported. The advantage of this is that the driver is reasonably clean and doesn't need magic environment variables like MIRI_SYSROOT, and we don't have to fight rustc_driver to use a different default sysroot. Everything is done in cargo-miri (and the other much simpler driver wrappers) where it can hopefully be debugged much better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,10 +464,9 @@ Moreover, Miri recognizes some environment variables:
   standard library that it will build and use for interpretation. This directory
   must point to the `library` subdirectory of a `rust-lang/rust` repository
   checkout.
-* `MIRI_SYSROOT` (recognized by `cargo miri` and the Miri driver) indicates the sysroot to use. When
+* `MIRI_SYSROOT` (recognized by `cargo miri` and the test suite) indicates the sysroot to use. When
   using `cargo miri`, this skips the automatic setup -- only set this if you do not want to use the
-  automatically created sysroot. For directly invoking the Miri driver, this variable (or a
-  `--sysroot` flag) is mandatory. When invoking `cargo miri setup`, this indicates where the sysroot
+  automatically created sysroot. When invoking `cargo miri setup`, this indicates where the sysroot
   will be put.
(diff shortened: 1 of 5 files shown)


RalfJung:
autocfg got patched. :) So this should pass CI now.

It's still a bit fragile though due to https://github.com/rust-lang/cargo/issues/10885. Also if we land this now cross-builds will break for anyone using autocfg older than 1.2. (But host builds should work, autocfg probes will just use the wrong sysroot.)

RalfJung:
Ah the eyre build probe also still fails to apply RUSTC_WRAPPER. So landing this would break everything that uses eyre...

RalfJung:
Also this should only break with old autocfg when using `--target`; plain `cargo miri test` should be unaffected.

saethlin:
> how long do you think we should wait before rolling this out?

Based on the downloads graph on crates.io, 1.2 seems like just over a third of downloads. I'm concerned we have a surprising amount of --target usage due to our much better shim support on x86_64-unknown-linux-gnu, so I would prefer to wait a week from now.

RalfJung:
>  I would prefer to wait a week from now.

We have waited 2 weeks since then. Download stats for autocfg haven't really changed.

Let's ship this; we can always revert if there are too many problems.
@bors r+

RalfJung:
The build has long finished, but somehow bors didn't notice...?
@bors retry

PR #3415, opened 2024-03-25
Tree Borrows: Make tree root always be initialized

JoJoDeveloping:
This PR fixes a slight annoyance we discovered while formally proving that certain optimizations are sound with Tree Borrows. In particular... (copied from the commit message):

There should never be an `Active` but not initialized node in the borrow tree. If such a node exists, and if it has a protector, then on a foreign read, it could become disabled. This leads to some problems when formally proving that read moving optimizations are sound.

The root node is the only node for which this can happen, since all other nodes can only become `Active` when actually used. But special-casing the root node here is annoying to reason about, everything becomes much nicer if we can simply say that *all* `Active` nodes must be initialized. This requires making the root node default-initialized.

This is also more intuitive, since the root arguably becomes initialized during the allocation, which can be considered a write.

### DIFF
--- a/src/borrow_tracker/tree_borrows/tree.rs
+++ b/src/borrow_tracker/tree_borrows/tree.rs
@@ -42,6 +42,7 @@ pub(super) struct LocationState {
     /// protected, that is *not* the case for uninitialized locations. Instead we just have a latent
     /// "future initial permission" of `Disabled`, causing UB only if an access is ever actually
     /// performed.
+    /// Note that the tree root is also always initialized, as if the allocation was a write access.
     initialized: bool,
     /// This pointer's current permission / future initial permission.
     permission: Permission,
@@ -55,17 +56,18 @@ pub(super) struct LocationState {
 }
 
 impl LocationState {
-    /// Default initial state has never been accessed and has been subjected to no
-    /// foreign access.
-    fn new(permission: Permission) -> Self {
+    /// Constructs a new initial state. It has neither been accessed, nor been subjected
+    /// to any foreign access yet.
+    /// The permission is not allowed to be `Active`.
+    fn new_uninit(permission: Permission) -> Self {
+        assert!(permission.is_initial() || permission.is_disabled());
         Self { permission, initialized: false, latest_foreign_access: None }
     }
 
-    /// Record that this location was accessed through a child pointer by
-    /// marking it as initialized
-    fn with_access(mut self) -> Self {
-        self.initialized = true;
-        self
+    /// Constructs a new initial state. It has not yet been subjected
+    /// to any foreign access. However, it is already marked as having been accessed.
+    fn new_init(permission: Permission) -> Self {
+        Self { permission, initialized: true, latest_foreign_access: None }
     }
 
     /// Check if the location has been initialized, i.e. if it has
@@ -238,8 +240,10 @@ pub(super) struct Node {
     /// If the pointer was reborrowed, it has children.
     // FIXME: bench to compare this to FxHashSet and to other SmallVec sizes
     pub children: SmallVec<[UniIndex; 4]>,
-    /// Either `Reserved` or `Frozen`, the permission this tag will be lazily initialized
-    /// to on the first access.
+    /// Either `Reserved`,  `Frozen`, or `Disabled`, it is the permission this tag will
+    /// lazily be initialized to on the first access.
+    /// It is only ever `Disabled` for a tree root, since the root is initialized to `Active` by
+    /// its own separate mechanism.
     default_initial_perm: Permission,
     /// Some extra information useful only for debugging purposes
     pub debug_info: NodeDebugInfo,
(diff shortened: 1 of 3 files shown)


src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -57,8 +58,8 @@ pub(super) struct LocationState {
 impl LocationState {
     /// Default initial state has never been accessed and has been subjected to no
     /// foreign access.
```
RalfJung, line 60: The comment above seems outdated now.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -240,7 +241,12 @@ pub(super) struct Node {
     pub children: SmallVec<[UniIndex; 4]>,
     /// Either `Reserved` or `Frozen`, the permission this tag will be lazily initialized
     /// to on the first access.
+    /// Can also be `Active` for the root node, in which case we consider every node
+    /// to be already initialized (see `is_default_initialized`).
     default_initial_perm: Permission,
+    /// Whether this node is always considered to be initialized. This is `true` for the
+    /// root node, which is also the only by-default `Active` node.
+    is_default_initialized: bool,
```
RalfJung, line 249: Why do you need this?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -599,7 +609,10 @@ impl<'tcx> Tree {
          -> Result<ContinueTraversal, TransitionError> {
             let NodeAppArgs { node, mut perm, rel_pos } = args;
 
-            let old_state = perm.or_insert(LocationState::new(node.default_initial_perm));
+            let old_state = perm.or_insert(LocationState::new(
+                node.is_default_initialized,
```
RalfJung, line 613: I think this can always be `false` as it can never be reached for the root node -- the root node is immediately initialized for the entire range it can ever be used for.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -57,8 +58,8 @@ pub(super) struct LocationState {
 impl LocationState {
     /// Default initial state has never been accessed and has been subjected to no
     /// foreign access.
-    fn new(permission: Permission) -> Self {
-        Self { permission, initialized: false, latest_foreign_access: None }
+    fn new(initialized: bool, permission: Permission) -> Self {
+        Self { permission, initialized, latest_foreign_access: None }
```
RalfJung, line 62: May be worth a `debug_assert!` to ensure that if `permission` is `Active` then `initialized` is `true`. After all we rely on this in the proof...

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -516,11 +516,11 @@ mod spurious_read {
         let source = LocStateProtPair {
             xy_rel: RelPosXY::MutuallyForeign,
             x: LocStateProt {
-                state: LocationState::new(Permission::new_frozen()).with_access(),
+                state: LocationState::new(false, Permission::new_frozen()).with_access(),
```
RalfJung, line 519: Bare booleans are hard to read... maybe a `new_uninit` constructor would be useful?

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: Turns out it is not necessary. Instead, we've decided to implement a different approach (see comment below)

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: I've added a debug assert.

src/borrow_tracker/tree_borrows/tree/tests.rs
JoJoDeveloping: There are now two constructors, one for `init` and one for `uninit`

JoJoDeveloping:
After some further thought, we decided to make the root note default `Disabled` and to manually set it to `Active` initialized on all in-bounds positions. This way, out-of-bounds accesses are also UB in TB.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -458,15 +464,18 @@ impl Tree {
...
                 },
             );
             nodes
         };
         let rperms = {
             let mut perms = UniValMap::default();
-            perms.insert(root_idx, LocationState::new(root_perm).with_access());
+            // we manually set it to `Active` on all in-bounds positions.
+            // we also ensure that it is initalized, so that no `Active` but
+            // not yet initialized nodes exist.
```
RalfJung, line 477:
```suggestion
            // We manually set it to `Active` on all in-bounds positions.
            // We also ensure that it is initalized, so that no `Active` but
            // not yet initialized nodes exist. Essentially, we pretend there
            // was a write that initialized these to `Active`.
```

RalfJung:
LGTM, just one comment nit. :)

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: Done

RalfJung: Not dropped part of my diff, I had also fixed capitalization in the first line.

That's why using the github button is a good idea. ;)  (You can than still pull, squash, force-push if you want to avoid the extra commit.)

JoJoDeveloping: Ah, alternatively one can also not be blind :see_no_evil: 
I've also fixed capitalization in other places.

RalfJung:
Thanks!
@bors r+

PR #3423, opened 2024-03-26
test-cargo-miri: add proc-macro2

RalfJung:
This is already in the dependency tree of `serde_derive`, but I guess there is is a host dependency, here it is a target dependency.

The logic is presumably the same as in anyhow, so we don't need both; let's test the one that is more widely used.


test-cargo-miri/Cargo.toml
```diff
@@ -12,20 +12,20 @@ edition = "2018"
...
+proc_macro_crate = { path = "proc-macro-crate" }
 issue_1567 = { path = "issue-1567" }
 issue_1691 = { path = "issue-1691" }
 issue_1705 = { path = "issue-1705" }
-issue_1760 = { path = "issue-1760" }
 issue_rust_86261 = { path = "issue-rust-86261" }
 
 [dev-dependencies]
 byteorder_2 = { package = "byteorder", version = "0.5" } # to test dev-dependencies behave as expected, with renaming
-# Not actually used, but exercises some unique code path (`--extern` .so file).
```
RalfJung, line 23: Turns out the unique code path is "being a proc macro", which the issue_1760 already covers.
So we can remove serde_derive (and therefore syn, by far the slowest dependency of test-cargo-miri).

RalfJung:
We should test a sufficiently recent proc-macro2 that even includes a probe to begin with...
@bors r+

PR #3426, opened 2024-03-29
Implement minimal socket shim

ZoeS17:
I vaguely followed socketpair as to implementation details. I attempted to make a test too but there's not exactly a good way, that I can figure out, besides simply checking that an file descriptor gets made. I dd so for both the libc crate and the version I implemented.

Free feel to ask question and/or adjust if needed.

### DIFF
--- a/src/shims/unix/linux/foreign_items.rs
+++ b/src/shims/unix/linux/foreign_items.rs
@@ -86,6 +86,12 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 let ptr = this.mremap(old_address, old_size, new_size, flags)?;
                 this.write_scalar(ptr, dest)?;
             }
+            "socket" => {
+                let [domain, type_, protocol] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let result = this.socket(domain, type_, protocol)?;
+                this.write_scalar(result, dest)?;
+            }
             "socketpair" => {
                 let [domain, type_, protocol, sv] =
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
--- a/src/shims/unix/linux/epoll.rs
+++ b/src/shims/unix/linux/epoll.rs
@@ -196,4 +196,3 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
             Ok(Scalar::from_i32(this.fd_not_found()?))
         }
     }
-}
--- a/src/shims/unix/linux/fd/socket.rs
+++ b/src/shims/unix/linux/fd/socket.rs
@@ -1,0 +1,28 @@
+use crate::*;
+
+use crate::shims::unix::fs::FileDescriptor;
+
+use std::io;
+
+/// Socket
+///
+/// We currently don't allow sending any data through this socket, so this can be just a dummy.
+#[derive(Debug)]
+pub struct Socket;
+
+impl FileDescriptor for Socket {
+    fn name(&self) -> &'static str {
+        "socket"
+    }
+
+    fn dup(&mut self) -> io::Result<Box<dyn FileDescriptor>> {
+        Ok(Box::new(Socket))
+    }
+
+    fn close<'tcx>(
+        self: Box<Self>,
+        _communicate_allowed: bool,
+    ) -> InterpResult<'tcx, io::Result<i32>> {
+        Ok(Ok(0))
+    }
+}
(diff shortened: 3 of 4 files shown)


src/shims/unix/linux/fd.rs
```diff
@@ -208,6 +210,39 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        &mut self,
+        domain: &OpTy<'tcx, Provenance>,
+        type_: &OpTy<'tcx, Provenance>,
+        protocol: &OpTy<'tcx, Provenance>,
+    ) -> InterpResult<'tcx, Scalar<Provenance>> {
+        let this = self.eval_context_mut();
+
+        let _domain = this.read_scalar(domain)?.to_i32()?;
+        let _type_ = this.read_scalar(type_)?.to_i32()?;
+        let _protocol = this.read_scalar(protocol)?.to_i32()?;
```
RalfJung, line 238: Unless we intend to support *all* values here, we should `throw_unsupported!` for socket types we do not support.

RalfJung (commented):
Thanks for the PR!

However, as-is I don't think landing this would be very useful. Does it even make sense to support `socket` alone? As you said we can't really write any tests, so this doesn't actually make any program work with Miri. Likely this entire function will have to be rewritten as we discover how sockets can actually be implemented.

I think we need a plan for what the support for socket + bind + connect looks like. Once we have a plan we can land it step-by-step, but it makes no sense to start going without a plan for where we want to go. In particular, clearly the `Socket` type can't just be a unit struct. We need a sketch for what that type should look like. Then we can land this in multiple PRs, but at least there will be a perspective for where we are going.

If we land this as-is it will just be another useless stub next to socketpair, a half-finished construction site without any plan or vision or idea for what to do next -- that's not really useful for anything. I don't like the existing construction site, I regret that we landed it.  I don't want us to add another one.

src/shims/unix/linux/foreign_items.rs
```diff
@@ -88,6 +88,12 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 let ptr = this.mremap(old_address, old_size, new_size, flags)?;
                 this.write_scalar(ptr, dest)?;
             }
+            "socket" => {
```
RalfJung, line 91: `socket` is a standard Unix function, why is this in the "linux" module?

RalfJung:
IOW, I think this should follow the proposed "project" process: https://github.com/rust-lang/miri/issues/3443.
We're happy to help you through that process! This is probably best done on [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/269128-miri), but you can also open an issue dedicated to the project and we can discuss there.

RalfJung:
I have opened an issue to track this: https://github.com/rust-lang/miri/issues/3449.

I think blocking sockets are actually highly non-trivial to implement.

RalfJung:
So to be clear I think before any implementation work is done, we need a clear description of which part of the socket API is intended to be implemented (TCP/UDP/..., blocking/non-blocking, which shims are involved). And since this will be non-trivial to implement, we also need a description of what the implementation plan is -- which new state is needed in Miri, how do we handle waking up the right threads when the sockets they are blocked on get ready, things like that. I laid out some of the complications in https://github.com/rust-lang/miri/issues/3449.

Adding sockets is *not* an easy project. And with projects of this scale it's not a good idea to just go implement something -- that has a pretty high risk that we'll ask you to re-write it entirely as we see flaws with the fundamental structure of the approach, and it also makes review much harder when we have to reverse engineer the high-level approach from the code.

@ZoeS17 what *is* your goal with this PR? Is there a motivation for wanting to land just the `socket` shim without actually supporting sockets at all?

ZoeS17:
> Adding sockets is _not_ an easy project. And with projects of this scale it's not a good idea to just go implement something -- that has a pretty high risk that we'll ask you to re-write it entirely as we see flaws with the fundamental structure of the approach, and it also makes review much harder when we have to reverse engineer the high-level approach from the code.

Agreed.

> what _is_ your goal with this PR? Is there a motivation for wanting to land just the `socket` shim without actually supporting sockets at all?

My intent was originally to land this as a shim without supporting sockets at all but given what has been said in the #3449 I agree that it seems foolish to continue as a "hollow" implementation. I originally come upon the desire to pick up this work because I was attempting to run miri against a program I was building that utilizes sockets to connect to IRC. Since obviously that needs a deeper implementation then I originally understood I'm up for closing this PR with an understanding that I'm willing to admit I may be out of my depth here.

RalfJung:
>  I originally come upon the desire to pick up this work because I was attempting to run miri against a program I was building that utilizes sockets to connect to IRC. 

Ah, yes makes sense. I'm afraid that's a pretty hard problem and quite far off currently, unfortunately.

I'll close this PR then. Thanks anyway!

PR #3429, opened 2024-03-30
De-duplicate SSE2 sll/srl/sra code

eduardosm:


src/shims/x86/mod.rs
```diff
@@ -468,6 +468,79 @@ fn unary_op_ps<'tcx>(
...
+    this: &mut crate::MiriInterpCx<'_, 'tcx>,
+    left: &OpTy<'tcx, Provenance>,
+    right: &OpTy<'tcx, Provenance>,
+    which: ShiftOp,
+    dest: &MPlaceTy<'tcx, Provenance>,
+) -> InterpResult<'tcx, ()> {
+    let (left, left_len) = this.operand_to_simd(left)?;
+    let (dest, dest_len) = this.mplace_to_simd(dest)?;
+
+    assert_eq!(dest_len, left_len);
```
eduardosm, line 494: I did not include `assert_eq!(dest_len, right_len);` because in AVX2 (not yet implemented), `left` is 256-bit and `right` is 128-bit.

src/shims/x86/mod.rs
```diff
@@ -468,6 +468,79 @@ fn unary_op_ps<'tcx>(
     Ok(())
 }
 
+enum ShiftOp {
```
saethlin, line 471: What about Srl? I suppose it doesn't make sense for some reason?

src/shims/x86/mod.rs
RalfJung: `len` is the number of elements here, not the bitwidth. In AVX2 do the two sides have a different number of elements...?

eduardosm: Maybe I should compare `left.layout` and `dest.layout`? `left` has twice elements in AVX2. `right` is always a 128-bit SIMD vector. We do not care about its exact layout, since we transmute it to `[u64; 2]` and take the first element.

src/shims/x86/mod.rs
eduardosm: Sla would be identical to Sll, since there isn't any difference to left-shifting signed and unsigned numbers

src/shims/x86/mod.rs
RalfJung: > since we transmute it to [u64; 2] and take the first element.

That sounds like a cursed API. Can't callers use a proper Rust type reflecting this (e.g. `u64x2`)?

eduardosm: I guess that's up to LLVM

RalfJung: Oh right I forgot these are not our nice SIMD intrinsics.

But in the end it's about how Rust code uses them, no? If all stdarch functions call this with u64x2 as their second type we'd be good. What *do* they call it as?

eduardosm: The intrinsics are defined as if the element type of the second operand was the same as the first's.

```rust
pub unsafe fn _mm_sll_epi32(a: __m128i, count: __m128i) -> __m128i {
    transmute(pslld(a.as_i32x4(), count.as_i32x4()))
}
```

```rust
#[link_name = "llvm.x86.sse2.psll.d"]
fn pslld(a: i32x4, count: i32x4) -> i32x4;
```

```rust
pub unsafe fn _mm256_sll_epi16(a: __m256i, count: __m128i) -> __m256i {
    transmute(psllw(a.as_i16x16(), count.as_i16x8()))
}
```

```rust
#[link_name = "llvm.x86.avx2.psll.w"]
fn psllw(a: i16x16, count: i16x8) -> i16x16;
```

RalfJung: How does that even make sense? In particular for pslld and psllw, those look like the actual shift amount will be determined by the first 2 or 4 elements of the `count` vector...?

src/shims/x86/mod.rs
RalfJung, line 471: Please add doc comments explaining these. "Sll" is not exactly a self-explaining acronym.

src/shims/x86/mod.rs
```diff
@@ -468,6 +468,79 @@ fn unary_op_ps<'tcx>(
...
+    Sll,
+    Srl,
+    Sra,
+}
+
+/// Shifts each element of `left` by a scalar amount. The shift amount
+/// is taken `right`, which is a 128-bit vector, only the lowest 64 bits
+/// are used.
```
RalfJung, line 479:
```suggestion
/// is determined by the lowest 64bits of `right` (which is a 128-bit vector).
```

src/shims/x86/mod.rs
```diff
@@ -468,6 +468,86 @@ fn unary_op_ps<'tcx>(
     Ok(())
 }
 
+enum ShiftOp {
+    /// Shift left logic
+    Sll,
+    /// Shift right logic (shift in zeros)
+    Srl,
+    /// Shift right arithmetic (shift in sign)
+    Sra,
```
RalfJung, line 477:
```suggestion
    /// Shift left, logically (shift in zeros) -- same as shift left, arithmetically
    Sll,
    /// Shift right, logically (shift in zeros)
    Srl,
    /// Shift right, arithmetically (shift in sign)
    Sra,
```
Would it make sense to rename `Sll` to `Sl` since there is only one kind of left-shift?

eduardosm: What about?

```rust
enum ShiftOp {
    Left,
    RightLogic,
    RightArith,
}
```

Since the S is actually redundant

RalfJung: Sounds good :)

RalfJung:
Thanks. :)
@bors r+

PR #3432, opened 2024-03-30
run GC stress test only for host tests

RalfJung:
I suspect these are a significant contributor to our Linux CI job being by far the slowest currently. Let's see.

We have Linux, Windows, and macOS hosts so all major OSes are still covered.

### DIFF
--- a/ci/ci.sh
+++ b/ci/ci.sh
@@ -130,6 +130,8 @@ case $HOST_TARGET in
     MIRI_TEST_TARGET=aarch64-unknown-linux-gnu run_tests
     MIRI_TEST_TARGET=aarch64-apple-darwin run_tests
     MIRI_TEST_TARGET=i686-pc-windows-gnu run_tests
+    MIRI_TEST_TARGET=x86_64-pc-windows-gnu run_tests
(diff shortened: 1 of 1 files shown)


RalfJung:
Oh wow yes, this cuts Linux CI time from ~53 to 23 minutes, so what was previously the slowest builder is now the fastest.^^

RalfJung:
@rust-lang/miri any objections to this? The macOS run was unusually slow in this PR, usually it's below 30m.

The Windows run is often close to 50min, that is the bottleneck. We currently test 3 foreign targets on Windows, it probably makes sense to move 2 of them to the Linux runner. (I'd like to have one 64bit Unix target still there to test "64bit target on 32bit host" and "Unix target on Windows host".)

saethlin:
This change seems quite reasonable to me.

RalfJung:
Windows still took 42 minutes... if it remains the bottleneck I'm considering just not doing the `many-seeds` and mir-opt-level=4 tests on Windows; doing those on macOS and Linux should suffice.

But for now let's land this.
@bors r+

PR #3433, opened 2024-03-30
experiment with macOS M1 runners

RalfJung:
Try running our macOS CI on the new free M1 runners. This would mean that our 3 jobs run 3 different OSes and 3 different host CPU architectures. :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,15 +49,16 @@ jobs:
         with:
           path: |
             # Taken from <https://doc.rust-lang.org/nightly/cargo/guide/cargo-home.html#caching-the-cargo-home-in-ci>.
-            ~/.cargo/bin
+            # Cache package/registry information
             ~/.cargo/registry/index
             ~/.cargo/registry/cache
             ~/.cargo/git/db
-            # contains package information of crates installed via `cargo install`.
+            # Cache installed binaries
+            ~/.cargo/bin
             ~/.cargo/.crates.toml
             ~/.cargo/.crates2.json
-          key: ${{ runner.os }}-cargo-reset20230315-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: ${{ runner.os }}-cargo-reset20230315
+          key: cargo-${{ runner.os }}-reset20240331-${{ hashFiles('**/Cargo.lock') }}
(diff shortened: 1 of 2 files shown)


RalfJung:
Hm, that doesn't seem to work -- they don't have cargo?
According to [this](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md) Ruts is supposed to be installed.

RalfJung:
Oh so the problem was the cache, interesting. No idea how a bad cache would lead to that kind of error, probably because some binaries have the wrong architecture or so?

And the build was very fast, too.
@rust-lang/miri what do you think, shall we give these runners a try? It's easy to switch back to the old runners if we're getting problems.

RalfJung:
One point I didn't realize is that this is still a tier 2 target. So there's some risk things break as rustc CI doesn't cover everything.

But it's also easy to revert this PR in the future, should it become a problem.

RalfJung:
Judging from [this GH announcement](https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/), `macos-latest` will switch to M1 machines some time soon (April - June). So we have to either do something like this PR or set the runner to `macos-12` to get more time to migrate.

saethlin:
I do not think the risk is very high. If aarch64 MacOS breaks, too many of us will know because we develop on it. Speed is nice. No objection to merging this.

RalfJung:
> I do not think the risk is very high. If aarch64 MacOS breaks, too many of us will know because we develop on it. Speed is nice. No objection to merging this.

Yeah, makes sense. And additionally we can serve as a canary for Rust compiler stuff breaking on M1 hosts, which is a good service to the compiler team. This target is clearly going to become tier 1, so we may as well be part of slowly increasing test coverage.

@bors r+

Cc @shepmaster since you seem to be tracking M1 progress. :)

shepmaster:
> this is still a tier 2 target

To the best of my knowledge, this is in name only. Right when the M1 runners came out we enabled tests and other things that the tier 1 platforms do. It’s mostly waiting on me to do the paperwork for the promotion and I’m just waiting for a bit of stability on GitHub’s side over a medium-length time period. 

> switch to M1 machines some time soon (April - June)

I had missed that, so that might be the sign I was waiting for…

PR #3436, opened 2024-03-31
speed up Windows CI

RalfJung:
The many-seeds test is taking 15 minutes. Let's just run that only once instead of many times on Windows.

Also refactor the CI script to make the caller control which tests are being run.


RalfJung:
I would say that worked, they are all below 50 minutes now. :)

@bors r+

PR #3438, opened 2024-03-31
update josh version and guidance

RalfJung:


CONTRIBUTING.md
```diff
@@ -266,12 +268,6 @@ needed.
 
 ### Exporting changes to the rustc repo
 
-Keep in mind that pushing is the most complicated job that josh has to do -- pulling just filters
-the rustc history, but pushing needs to construct a new rustc history that would filter to the given
-Miri history! To avoid problems, it is a good idea to always pull immediately before you push. If
-you are getting strange errors, chances are you are running into [this josh
-bug](https://github.com/josh-project/josh/issues/998). In that case, please get in touch on Zulip.
```
RalfJung, line 273: The new version of josh includes a patch that fixes many cases of this issue. I haven't seen problems since I just josh with that patch, so I think we can remove this paragraph.

PR #3446, opened 2024-04-03
eventfd: fix flag check and note a FIXME

RalfJung:


RalfJung:
There's no logs for the failed jobs? What's going on there...?

PR #3461, opened 2024-04-11
Add localtime_r shim

tiif:
- Implement ``localtime_r`` shim as mentioned in #2057 

Note:
- ``tm_zone``, ``tm_gmtoff`` might not be consistent with ``libc::localtime_r`` as custom implementation is provided through ``chrono``. Due to the lack of daylight saving information in ``chrono``, ``is_dst`` value will always be ``-1``.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ smallvec = "1.7"
 aes = { version = "0.8.3", features = ["hazmat"] }
 measureme = "11"
 ctrlc = "3.2.5"
+chrono = { version = "0.4.38", default-features = false, features = ["clock"] }
 
 # Copied from `compiler/rustc/Cargo.toml`.
 # But only for some targets, it fails for others. Rustc configures this in its CI, but we can't
--- a/src/shims/time.rs
+++ b/src/shims/time.rs
@@ -1,4 +1,8 @@
+use std::ffi::OsString;
+use std::fmt::Write;
 use std::time::{Duration, SystemTime};
+
+use chrono::{DateTime, Datelike, Local, Timelike, Utc};
 
 use crate::concurrency::thread::MachineCallback;
 use crate::*;
@@ -107,6 +111,80 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         Ok(0)
     }
 
+    // The localtime() function shall convert the time in seconds since the Epoch pointed to by
+    // timer into a broken-down time, expressed as a local time.
+    // https://linux.die.net/man/3/localtime_r
+    fn localtime_r(
+        &mut self,
+        timep: &OpTy<'tcx, Provenance>,
+        result_op: &OpTy<'tcx, Provenance>,
+    ) -> InterpResult<'tcx, Pointer<Option<Provenance>>> {
+        let this = self.eval_context_mut();
+
+        this.assert_target_os_is_unix("localtime_r");
+        this.check_no_isolation("`localtime_r`")?;
+
+        let timep = this.deref_pointer(timep)?;
+        let result = this.deref_pointer_as(result_op, this.libc_ty_layout("tm"))?;
+
+        // The input "represents the number of seconds elapsed since the Epoch,
+        // 1970-01-01 00:00:00 +0000 (UTC)".
+        let sec_since_epoch: i64 = this
+            .read_scalar(&timep)?
+            .to_int(this.libc_ty_layout("time_t").size)?
+            .try_into()
+            .unwrap();
+        let dt_utc: DateTime<Utc> =
+            DateTime::from_timestamp(sec_since_epoch, 0).expect("Invalid timestamp");
+        // Convert that to local time, then return the broken-down time value.
+        let dt: DateTime<Local> = DateTime::from(dt_utc);
+
+        // This value is always set to -1, because there is no way to know if dst is in effect with
+        // chrono crate yet.
+        // This may not be consistent with libc::localtime_r's result.
+        let tm_isdst = -1;
+
+        // tm_zone represents the timezone value in the form of: +0730, +08, -0730 or -08.
+        // This may not be consistent with libc::localtime_r's result.
+        let offset_in_second = Local::now().offset().local_minus_utc();
+        let tm_gmtoff = offset_in_second;
+        let mut tm_zone = String::new();
+        if offset_in_second < 0 {
+            tm_zone.push('-');
+        } else {
+            tm_zone.push('+');
+        }
+        let offset_hour = offset_in_second.abs() / 3600;
+        write!(tm_zone, "{:02}", offset_hour).unwrap();
+        let offset_min = (offset_in_second.abs() % 3600) / 60;
+        if offset_min != 0 {
+            write!(tm_zone, "{:02}", offset_min).unwrap();
+        }
+
+        // FIXME: String de-duplication is needed so that we only allocate this string only once
+        // even when there are multiple calls to this function.
+        let tm_zone_ptr =
+            this.alloc_os_str_as_c_str(&OsString::from(tm_zone), MiriMemoryKind::Machine.into())?;
+
+        this.write_pointer(tm_zone_ptr, &this.project_field_named(&result, "tm_zone")?)?;
+        this.write_int_fields_named(
+            &[
+                ("tm_sec", dt.second().into()),
+                ("tm_min", dt.minute().into()),
+                ("tm_hour", dt.hour().into()),
+                ("tm_mday", dt.day().into()),
+                ("tm_mon", dt.month0().into()),
+                ("tm_year", dt.year().checked_sub(1900).unwrap().into()),
+                ("tm_wday", dt.weekday().num_days_from_sunday().into()),
+                ("tm_yday", dt.ordinal0().into()),
+                ("tm_isdst", tm_isdst),
+                ("tm_gmtoff", tm_gmtoff.into()),
+            ],
+            &result,
+        )?;
+
+        Ok(result.ptr())
+    }
     #[allow(non_snake_case, clippy::arithmetic_side_effects)]
     fn GetSystemTimeAsFileTime(
         &mut self,
--- a/tests/pass-dep/shims/libc-misc.rs
+++ b/tests/pass-dep/shims/libc-misc.rs
@@ -213,6 +213,50 @@ fn test_posix_gettimeofday() {
     assert_eq!(is_error, -1);
 }
 
+fn test_localtime_r() {
+    use std::ffi::CStr;
+    use std::{env, ptr};
+
+    // Set timezone to GMT.
+    let key = "TZ";
+    env::set_var(key, "GMT");
+
+    const TIME_SINCE_EPOCH: libc::time_t = 1712475836;
+    let custom_time_ptr = &TIME_SINCE_EPOCH;
+    let mut tm = libc::tm {
+        tm_sec: 0,
+        tm_min: 0,
+        tm_hour: 0,
+        tm_mday: 0,
+        tm_mon: 0,
+        tm_year: 0,
+        tm_wday: 0,
+        tm_yday: 0,
+        tm_isdst: 0,
+        tm_gmtoff: 0,
+        tm_zone: std::ptr::null_mut::<libc::c_char>(),
+    };
+    let res = unsafe { libc::localtime_r(custom_time_ptr, &mut tm) };
+
+    assert_eq!(tm.tm_sec, 56);
+    assert_eq!(tm.tm_min, 43);
+    assert_eq!(tm.tm_hour, 7);
+    assert_eq!(tm.tm_mday, 7);
+    assert_eq!(tm.tm_mon, 3);
+    assert_eq!(tm.tm_year, 124);
+    assert_eq!(tm.tm_wday, 0);
+    assert_eq!(tm.tm_yday, 97);
+    assert_eq!(tm.tm_isdst, -1);
+    assert_eq!(tm.tm_gmtoff, 0);
+    unsafe { assert_eq!(CStr::from_ptr(tm.tm_zone).to_str().unwrap(), "+00") };
+
+    // The returned value is the pointer passed in.
+    assert!(ptr::eq(res, &mut tm));
+
+    //Remove timezone setting.
+    env::remove_var(key);
+}
+
 fn test_isatty() {
     // Testing whether our isatty shim returns the right value would require controlling whether
     // these streams are actually TTYs, which is hard.
@@ -365,6 +409,7 @@ fn main() {
     test_posix_realpath_errors();
 
     test_thread_local_errno();
+    test_localtime_r();
 
     test_isatty();
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,6 +38,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "android-tzdata"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
+
+[[package]]
+name = "android_system_properties"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "annotate-snippets"
 version = "0.9.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -107,6 +122,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bumpalo"
+version = "3.16.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "79296716171880943b8470b5f8d03aa55eb2e645a4874bdbb28adb49162e012c"
+
+[[package]]
 name = "camino"
 version = "1.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -149,6 +170,18 @@ name = "cfg-if"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
+
+[[package]]
+name = "chrono"
+version = "0.4.38"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a21f936df1771bf62b77f047b726c4625ff2e8aa607c01ec06e5a05bd8463401"
+dependencies = [
+ "android-tzdata",
+ "iana-time-zone",
+ "num-traits",
+ "windows-targets 0.52.3",
+]
 
 [[package]]
 name = "cipher"
@@ -215,6 +248,12 @@ dependencies = [
  "unicode-width",
  "windows-sys 0.52.0",
 ]
+
+[[package]]
+name = "core-foundation-sys"
+version = "0.8.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "06ea2b9bc92be3c2baa9334a323ebca2d6f074ff852cd1d7b11064035cd3868f"
 
 [[package]]
 name = "cpufeatures"
@@ -320,6 +359,29 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4271d37baee1b8c7e4b708028c57d816cf9d2434acb33a549475f78c181f6253"
 
 [[package]]
+name = "iana-time-zone"
+version = "0.1.60"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e7ffbb5a1b541ea2561f8c41c087286cc091e21e556a4f09a8f6cbf17b69b141"
+dependencies = [
+ "android_system_properties",
+ "core-foundation-sys",
+ "iana-time-zone-haiku",
+ "js-sys",
+ "wasm-bindgen",
+ "windows-core",
+]
+
+[[package]]
+name = "iana-time-zone-haiku"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f31827a206f56af32e590ba56d5d2d085f558508192593743f16b2306495269f"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "indenter"
 version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -370,6 +432,15 @@ checksum = "ac6c1946e1cea1788cbfde01c993b52a10e2da07f4bac608228d1bed20bfebf2"
 dependencies = [
  "cc",
  "libc",
+]
+
+[[package]]
+name = "js-sys"
+version = "0.3.69"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "29c15563dc2726973df627357ce0c9ddddbea194836909d655df6a75d2cf296d"
+dependencies = [
+ "wasm-bindgen",
 ]
 
 [[package]]
@@ -484,6 +555,7 @@ name = "miri"
 version = "0.1.0"
 dependencies = [
  "aes",
+ "chrono",
  "colored",
  "ctrlc",
  "getrandom",
@@ -510,6 +582,15 @@ dependencies = [
  "bitflags 2.4.2",
  "cfg-if",
  "libc",
+]
+
+[[package]]
+name = "num-traits"
+version = "0.2.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "da0df0e5185db44f69b44f26786fe401b6c293d1907744beaa7fa62b2e5a517a"
+dependencies = [
+ "autocfg",
 ]
 
 [[package]]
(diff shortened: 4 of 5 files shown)


tests/pass-dep/shims/libc-misc.rs
```diff
@@ -213,6 +213,62 @@ fn test_posix_gettimeofday() {
...
+    // Set timezone to GMT
+    //let key = "TZ";
+    //env::set_var(key, "GMT");
+    //assert_eq!(env::var(key), Ok("GMT".to_string()));
+
+    const custom_time: i64 = 1712475836;
+    let custom_time_ptr: *const i64 = &custom_time;
+
+    //TODO: This causes a leak
+    let zone = CString::new("").expect("Cstring new failed").into_raw();
```
tiif, line 229: Miri complains about a memory leak here, what is the proper way to assign a raw pointer for ``*const c_char`` so it will be freed afterwards?

src/shims/time.rs
```diff
@@ -107,6 +109,93 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+            tm_zone_str.push_str(&offset_min.to_string());
+        }
+
+        // Allocate pointer for tm_zone string
+        let byte_str = tm_zone_str.as_bytes();
+        let size: u64 = byte_str.len().try_into().unwrap();
+        let kind = MiriMemoryKind::C;
+        let ptr = this
+            .allocate_ptr(Size::from_bytes(size + 1), Align::from_bytes(2).unwrap(), kind.into())
+            .unwrap();
```
oli-obk, line 173: This is your memory leak. You want to use `MiriMemoryKind::Machine`. Also there's a convenience helper called `allocate_str` that does the whole "allocate then write" dance for you

src/shims/time.rs
```diff
@@ -107,6 +109,93 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         Ok(0)
     }
 
+    // The localtime() function shall convert the time in seconds since the Epoch pointed to by
+    // timer into a broken-down time, expressed as a local time.
+    // https://linux.die.net/man/3/localtime_r
+    #[allow(clippy::arithmetic_side_effects)]
```
RalfJung, line 115: Why allow this?

src/shims/time.rs
RalfJung: How is the memory for the time zone string managed in the real implementation?

We shouldn't allocate a new timezone string that is then lekaed each time this function is called. So if this becomes `Machine` memory then it should be allocated once at program startup.

oli-obk: This is what we do for `ERRNO` and panic messages, too. If we want to deduplicate things, we should add a general mechanism. It's not something really new from this PR.

https://www.man7.org/linux/man-pages/man3/tm.3type.html says 

> tm_zone points to static storage and may be overridden on
       subsequent calls to [localtime(3)](https://www.man7.org/linux/man-pages/man3/localtime.3.html) and similar functions (however,
       this never happens under glibc).

it also says

> tm_zone is the equivalent of [tzname(3)](https://www.man7.org/linux/man-pages/man3/tzname.3.html) for the timezone
       represented by this broken-down time.

So we could teach miri about `tmzone` and always fetch it to use its data, but it's not really clear to me what the right behaviour here is

Most `tm` man pages don't even contain this extra field at all.

tiif:
@RalfJung @oli-obk If memory leak in tm_zone string is a concern, would it be better if we just don't update the ``tm_zone`` field at all? The implementation provided is not accurate most of the time anyway since I didn't manage to use timezone abbreviation, I used put something like ``+08`` instead.

src/shims/time.rs
RalfJung: > This is what we do for ERRNO and panic messages, too. 

Not for errno, those are written to a caller-provided buffer.

But for panic messages, indeed. We should probably make this stack-allocated, the panic handler will make a copy of this anyway so the panic message should be freed when the panic handler returns. Or we add a de-duplicating "interner" that makes sure we allocate each string only once (i.e., this acts as-if it was a static string embedded in the binary).

> may be overridden on subsequent calls  to [localtime(3)](https://www.man7.org/linux/man-pages/man3/localtime.3.html)

But crucially, not for `localtime_r` as that would not be thread-safe. Would be interesting to figure out what glibc does here. Probably the parse the timezone info into a single global variable once and then just return pointers to that.

I'm okay with leaving this as a FIXME, and opening an issue to do string de-duplication.

Cargo.toml
```diff
@@ -24,6 +24,7 @@ smallvec = "1.7"
 aes = { version = "0.8.3", features = ["hazmat"] }
 measureme = "11"
 ctrlc = "3.2.5"
+chrono = "0.4.37"
```
RalfJung, line 27: Chrono has a bunch of default features, which of these do we really need?

src/shims/time.rs
```diff
@@ -107,6 +109,93 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        }
+
+        // Allocate pointer for tm_zone string
+        let byte_str = tm_zone_str.as_bytes();
+        let size: u64 = byte_str.len().try_into().unwrap();
+        let kind = MiriMemoryKind::C;
+        let ptr = this
+            .allocate_ptr(Size::from_bytes(size + 1), Align::from_bytes(2).unwrap(), kind.into())
+            .unwrap();
+        let Ok((_, _)) = this.write_c_str(byte_str, ptr.into(), size + 1) else {
```
RalfJung, line 174: Please never ever match on an `InterpResult` (like your `let ... else` here). Just use `?` instead.

src/shims/time.rs
```diff
@@ -107,6 +109,93 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                ("tm_wday", tm_wday.into()),
+                ("tm_yday", tm_yday.into()),
+                ("tm_isdst", tm_isdst),
+                ("tm_gmtoff", tm_gmtoff.into()),
+            ],
+            &result,
+        )?;
+
+        // Write to tm_zone field
+        this.write_pointer(ptr, &this.project_field(&result, 10)?)?;
```
RalfJung, line 195: Please do this with `project_field_named` instead.

src/shims/time.rs
```diff
@@ -107,6 +109,93 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        timep: &OpTy<'tcx, Provenance>,
+        result_op: &OpTy<'tcx, Provenance>,
+    ) -> InterpResult<'tcx, Pointer<Option<Provenance>>> {
+        let this = self.eval_context_mut();
+
+        this.assert_target_os_is_unix("localtime_r");
+        this.check_no_isolation("`localtime_r`")?;
+
+        let result = this.deref_pointer_as(result_op, this.libc_ty_layout("tm"))?;
+        let result_ptr = this.read_pointer(result_op)?;
```
RalfJung, line 127: Please follow the consistent pattern for our shims: the first thing we do is turn all argument `OpTy` into something of the right type.

So, `let timep = this.deref_pointer(timep)?;` should come before `let result`.

src/shims/time.rs
```diff
@@ -1,4 +1,6 @@
-use std::time::{Duration, SystemTime};
+use chrono::{DateTime, Datelike, Local, Timelike};
+use rustc_target::abi::{Align, Size};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
```
RalfJung, line 3:
```suggestion
use std::time::{Duration, SystemTime, UNIX_EPOCH};

use chrono::{DateTime, Datelike, Local, Timelike};

use rustc_target::abi::{Align, Size};
```

src/shims/time.rs
```diff
@@ -107,6 +109,93 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        let tm_hour = dt.hour();
+        let tm_mday = dt.day();
+        let tm_mon = dt.month0();
+        let tm_year = dt.year() - 1900;
+        let tm_wday = dt.weekday().num_days_from_sunday();
+        let tm_yday = dt.ordinal0();
+
+        // This value is always set to -1, because there is no way to know if dst is in effect with
+        // chrono crate yet
+        // This may not be consistent with libc::localtime_r's result
```
RalfJung, line 147:
```suggestion
        // chrono crate yet.
        // This may not be consistent with libc::localtime_r's result.
```

src/shims/time.rs
```diff
@@ -107,6 +109,93 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        let tm_isdst = -1;
+
+        // tm_zone represents the timezone value in the form of: +0730, +08, -0730 or -08
+        // This may not be consistent with libc::localtime_r's result
```
RalfJung, line 151:
```suggestion
        // tm_zone represents the timezone value in the form of: +0730, +08, -0730 or -08.
        // This may not be consistent with libc::localtime_r's result.
```
Also in the rest of the PR, please make sure that all sentences are properly ended with a `.`. Reading these comments is a lot harder when you can't tell where the sentences end.

src/shims/time.rs
```diff
@@ -107,6 +109,93 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        if offset_in_second < 0 {
+            tm_zone_str.push('-');
+        } else {
+            tm_zone_str.push('+');
+        }
+        let offset_hour = format!("{:0>2}", (offset_in_second.abs() / 3600).to_string());
+        tm_zone_str.push_str(&offset_hour);
+        let offset_min = (offset_in_second % 3600) / 60;
+        if offset_min != 0 {
+            tm_zone_str.push_str(&offset_min.to_string());
```
RalfJung, line 164: This should also pad to two digits.

src/shims/time.rs
```diff
@@ -107,6 +109,93 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        // This may not be consistent with libc::localtime_r's result
+        let offset_in_second = Local::now().offset().local_minus_utc();
+        let tm_gmtoff = offset_in_second;
+        let mut tm_zone_str = String::new();
+        if offset_in_second < 0 {
+            tm_zone_str.push('-');
+        } else {
+            tm_zone_str.push('+');
+        }
+        let offset_hour = format!("{:0>2}", (offset_in_second.abs() / 3600).to_string());
```
RalfJung, line 160: The `>` is unnecessary I think? The default for numeric types is right-aligned.

src/shims/time.rs
```diff
@@ -107,6 +109,93 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        tm_zone_str.push_str(&offset_hour);
```
RalfJung, line 161: You can use `write!(tm_zone_str, ...)` instead of using an intermediate `format!`.

src/shims/time.rs
```diff
@@ -107,6 +109,93 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        let offset_min = (offset_in_second % 3600) / 60;
```
RalfJung, line 162: You didn't use `abs` here, that seems wrong?

tiif:
Thanks for the thorough review! I need a little bit more time to fully address them, and will mark this pr as ready after that.

tiif:
Sorry for submitting so many comments one go, I didn't notice that these comments is under pending, and only visible to me :(

Cargo.toml
```diff
@@ -24,6 +24,7 @@ smallvec = "1.7"
 aes = { version = "0.8.3", features = ["hazmat"] }
 measureme = "11"
 ctrlc = "3.2.5"
+chrono = { version = "0.4.38", features = ["std"] }
```
oli-obk, line 27: you need to set `no-default-features=true`, otherwise all default features are still enabled, not just the ones you explicitly turned on

src/shims/time.rs
```diff
@@ -107,6 +112,87 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        let this = self.eval_context_mut();
+
+        this.assert_target_os_is_unix("localtime_r");
+        this.check_no_isolation("`localtime_r`")?;
+
+        let timep = this.deref_pointer(timep)?;
+        let result = this.deref_pointer_as(result_op, this.libc_ty_layout("tm"))?;
+        let result_ptr = this.read_pointer(result_op)?;
+
+        let sec_since_epoch: i64 = this.read_scalar(&timep)?.to_i64()?;
```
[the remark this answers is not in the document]
oli-obk: is it guaranteed to be pointer sized on all platforms? if so, you can use `read_target_isize` instead. Otherwise you can use `to_bits` instead of `to_i64`, but then you need to provide the size, which you can get from `timep.layout.size()`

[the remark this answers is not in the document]
tiif: I am not sure how verify if it is pointer size in all platforms, but using ``read_target_isize`` did pass all the tests for tier 1 linux platform locally.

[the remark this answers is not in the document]
RalfJung: For off_t we use something like
```rust
let offset = this.read_scalar(offset)?.to_int(this.libc_ty_layout("off_t").size)?;
```
Here it's `time_t`, I think; would make sense to use something similar.

src/shims/time.rs
RalfJung: `{:02}` should also result in `08`, doesn't it?

src/shims/time.rs
RalfJung: If you learned anything from https://news.ycombinator.com/item?id=2725015 it should be to not make assumptions like this. ;)

src/shims/time.rs
RalfJung: > It seems like they malloc it, but I didn't see it being freed anywhere yet.

They malloc it once and then keep a global cache.

> This issue is tracked under https://github.com/rust-lang/miri/issues/3470.

Thanks! That works for now.

src/shims/time.rs
tiif: Yes, I started to be wary of how weird timezones can behave after going through the rabbit hole. :')

src/shims/time.rs
tiif: Oh I see, if I use ``{:02}``, it will still result in ``+8``. I think it only works for int but not string. 

```rust
// This will print "7"
 println!("{:02}", "7");
// This will print "07"
 println!("{:02}", 7);

```

RalfJung: Ah yeah -- why are you calling `to_string`? I didn't even see that.^^ Please remove that call.

tiif: Sure, I actually forgot why I added a ``to_string()`` there...

src/shims/time.rs
```diff
@@ -107,6 +112,91 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        result_op: &OpTy<'tcx, Provenance>,
+    ) -> InterpResult<'tcx, Pointer<Option<Provenance>>> {
+        let this = self.eval_context_mut();
+
+        this.assert_target_os_is_unix("localtime_r");
+        this.check_no_isolation("`localtime_r`")?;
+
+        let timep = this.deref_pointer(timep)?;
+        let result = this.deref_pointer_as(result_op, this.libc_ty_layout("tm"))?;
+        let result_ptr = this.read_pointer(result_op)?;
```
RalfJung, line 130: You can get rid of `result_ptr` and use `result.ptr()` instead

src/shims/time.rs
```diff
@@ -107,6 +112,91 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+
+        let sec_since_epoch: i64 = this
```
RalfJung, line 132:
```suggestion
        // The input "represents the number of seconds elapsed since the Epoch,
        // 1970-01-01 00:00:00 +0000 (UTC)".
        let sec_since_epoch: i64 = this
```

src/shims/time.rs
```diff
@@ -107,6 +112,91 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+            .read_scalar(&timep)?
+            .to_int(this.libc_ty_layout("time_t").size)?
+            .try_into()
+            .unwrap();
+        let dt_utc: DateTime<Utc> =
+            DateTime::from_timestamp(sec_since_epoch, 0).expect("Invalid timestamp");
+        let dt: DateTime<Local> = DateTime::from(dt_utc);
+
```
RalfJung, line 140:
```suggestion

        
        // Convert that to local time, then return the broken-down time value.
        let dt: DateTime<Local> = DateTime::from(dt_utc);
```

src/shims/time.rs
```diff
@@ -107,6 +112,91 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        let tm_sec = dt.second();
+        let tm_min = dt.minute();
+        let tm_hour = dt.hour();
+        let tm_mday = dt.day();
+        let tm_mon = dt.month0();
+        let tm_year = dt.year().saturating_sub(1900);
+        let tm_wday = dt.weekday().num_days_from_sunday();
+        let tm_yday = dt.ordinal0();
```
RalfJung, line 148: Is it really worth let-binding all these variables? Seems just as clear to just inline them all below.

src/shims/time.rs
```diff
@@ -107,6 +112,91 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        let dt_utc: DateTime<Utc> =
+            DateTime::from_timestamp(sec_since_epoch, 0).expect("Invalid timestamp");
+        let dt: DateTime<Local> = DateTime::from(dt_utc);
+
+        let tm_sec = dt.second();
+        let tm_min = dt.minute();
+        let tm_hour = dt.hour();
+        let tm_mday = dt.day();
+        let tm_mon = dt.month0();
+        let tm_year = dt.year().saturating_sub(1900);
```
RalfJung, line 146: Is it even possible for this to overflow? Why use saturating arithmetic?

src/shims/time.rs
```diff
@@ -107,6 +112,91 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        let offset_in_second = Local::now().offset().local_minus_utc();
+        let tm_gmtoff = offset_in_second;
+        let mut tm_zone = String::new();
+        if offset_in_second < 0 {
+            tm_zone.push('-');
+        } else {
+            tm_zone.push('+');
+        }
+        let offset_hour = offset_in_second.abs() / 3600;
+        write!(&mut tm_zone, "{:02}", offset_hour)
```
RalfJung, line 166:
```suggestion
        write!(tm_zone, "{:02}", offset_hour)
```

src/shims/time.rs
```diff
@@ -107,6 +112,91 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+            .expect("Error occured while writing to tm_zone_str");
```
RalfJung, line 167:
```suggestion
            .unwrap();
```
This can never fail (guaranteed by `fmt` trait contracts)

src/shims/time.rs
```diff
@@ -107,6 +112,91 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        let offset_min = (offset_in_second.abs() % 3600) / 60;
+        if offset_min != 0 {
+            write!(&mut tm_zone, "{:02}", offset_min)
+                .expect("Error occured while writing to tm_zone_str");
```
RalfJung, line 171:
```suggestion
            write!(tm_zone, "{:02}", offset_min).unwrap();
```

src/shims/time.rs
```diff
@@ -107,6 +112,91 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        }
+        tm_zone.push('\0');
+
+        // FIXME: String de-duplication is needed so that we only allocate this string only once
+        // even when there are multiple calls to this function.
+        let tmzone_alloc =
```
RalfJung, line 177: Please use `alloc_os_str_as_c_str`, then you can also avoid dealing with the null terminator.

src/shims/time.rs
```diff
@@ -107,6 +112,91 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+            this.allocate_str(&tm_zone, MiriMemoryKind::Machine.into(), Mutability::Not)?;
+        let tm_zone_ptr = tmzone_alloc.ptr();
```
RalfJung, line 179: This let-binding can also be inlined without reducing clarity.

tests/pass-dep/shims/libc-misc.rs
```diff
@@ -213,6 +213,64 @@ fn test_posix_gettimeofday() {
...
+    unsafe { assert_eq!((*res).tm_min, 43) };
+    unsafe { assert_eq!((*res).tm_hour, 7) };
+    unsafe { assert_eq!((*res).tm_mday, 7) };
+    unsafe { assert_eq!((*res).tm_mon, 3) };
+    unsafe { assert_eq!((*res).tm_year, 124) };
+    unsafe { assert_eq!((*res).tm_wday, 0) };
+    unsafe { assert_eq!((*res).tm_yday, 97) };
+    unsafe { assert_eq!((*res).tm_isdst, -1) };
+    unsafe { assert_eq!((*res).tm_gmtoff, 0) };
+    unsafe { assert_eq!(CStr::from_ptr((*res).tm_zone).to_str().unwrap(), "+00") };
```
RalfJung, line 267: Please make this one big unsafe block rather than a lot of small ones. Or better, convert `res` to a reference so most of this becomes safe.

tests/pass-dep/shims/libc-misc.rs
```diff
@@ -213,6 +213,64 @@ fn test_posix_gettimeofday() {
...
 }
 
+fn test_localtime_r() {
+    use std::env;
+    use std::ffi::CStr;
+
+    // Set timezone to GMT.
+    let key = "TZ";
+    env::set_var(key, "GMT");
+    assert_eq!(env::var(key), Ok("GMT".to_string()));
```
RalfJung, line 223: No need to double-check that the `env` module works, we have tests for that elsewhere. (Same below for `env::remove_var`.)

tests/pass-dep/shims/libc-misc.rs
```diff
@@ -213,6 +213,64 @@ fn test_posix_gettimeofday() {
...
+        tm_min: 0,
+        tm_hour: 0,
+        tm_mday: 0,
+        tm_mon: 0,
+        tm_year: 0,
+        tm_wday: 0,
+        tm_yday: 0,
+        tm_isdst: 0,
+        tm_gmtoff: 0,
+        tm_zone: zone,
```
RalfJung, line 239:
```suggestion
    let mut tm = libc::tm {
        tm_sec: 0,
        tm_min: 0,
        tm_hour: 0,
        tm_mday: 0,
        tm_mon: 0,
        tm_year: 0,
        tm_wday: 0,
        tm_yday: 0,
        tm_isdst: 0,
        tm_gmtoff: 0,
        tm_zone: std::ptr::null_mut::<libc::c_char>(),
```

tests/pass-dep/shims/libc-misc.rs
```diff
@@ -213,6 +213,64 @@ fn test_posix_gettimeofday() {
...
+    use std::env;
+    use std::ffi::CStr;
+
+    // Set timezone to GMT.
+    let key = "TZ";
+    env::set_var(key, "GMT");
+    assert_eq!(env::var(key), Ok("GMT".to_string()));
+
+    const TIME_SINCE_EPOCH: i32 = 1712475836;
+    let custom_time_ptr = &(TIME_SINCE_EPOCH.into()) as *const libc::time_t;
```
RalfJung, line 226:
```suggestion
    const TIME_SINCE_EPOCH: libc::time_t = 1712475836;
    let custom_time_ptr = &TIME_SINCE_EPOCH;
```

tests/pass-dep/shims/libc-misc.rs
```diff
@@ -213,6 +213,64 @@ fn test_posix_gettimeofday() {
...
+        tm_mon: 0,
+        tm_year: 0,
+        tm_wday: 0,
+        tm_yday: 0,
+        tm_isdst: 0,
+        tm_gmtoff: 0,
+        tm_zone: zone,
+    };
+    let tm_ptr: *mut libc::tm = &mut tm as *mut libc::tm;
+    let res = unsafe { libc::localtime_r(custom_time_ptr, tm_ptr) };
```
RalfJung, line 242:
```suggestion
    let res = unsafe { libc::localtime_r(custom_time_ptr, &mut tm) };
```

RalfJung (commented):
Thanks! We're entering the final polishing stage of this PR. :)

@rustbot author

src/shims/time.rs
tiif: Since ``i32`` maximum value is ``2147483647``, should be impossible to overflow, at least not soon :). But I use this because the compiler will produce arithmetic side effect warning if I do 
```rust
let tm_year = dt.year() - 1900;
```
So this is mainly to eliminate the warning.

RalfJung: The intended way to eliminate the warning is `checked_sub().unwrap()`

src/shims/time.rs
```diff
@@ -107,6 +111,82 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+            .unwrap();
+        let dt_utc: DateTime<Utc> =
+            DateTime::from_timestamp(sec_since_epoch, 0).expect("Invalid timestamp");
+        // Convert that to local time, then return the broken-down time value.
+        let dt: DateTime<Local> = DateTime::from(dt_utc);
+
+        // This value is always set to -1, because there is no way to know if dst is in effect with
+        // chrono crate yet.
+        // This may not be consistent with libc::localtime_r's result.
+        let tm_isdst = -1;
```
tiif, line 145: This field can be inlined, but I think the comment is clearer when it is not inlined.

RalfJung: Agreed.

tests/pass-dep/shims/libc-misc.rs
```diff
@@ -239,30 +239,34 @@ fn test_localtime_r() {
...
+    assert_eq!(tm.tm_yday, 97);
+    assert_eq!(tm.tm_isdst, -1);
+    assert_eq!(tm.tm_gmtoff, 0);
+    unsafe { assert_eq!(CStr::from_ptr(tm.tm_zone).to_str().unwrap(), "+00") };
+
+    let res_tm: libc::tm;
+    unsafe {
+        res_tm = *res;
+    }
     // The returned value should match the pointer passed in.
```
RalfJung, line 257: Instead of comparing everything a second time, can't you just check that `res` is equal to `&mut tm` (as in, `ptr::eq`)?

src/shims/time.rs
```diff
@@ -107,6 +111,82 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        write!(tm_zone, "{:02}", offset_hour).unwrap();
+        let offset_min = (offset_in_second.abs() % 3600) / 60;
+        if offset_min != 0 {
+            write!(tm_zone, "{:02}", offset_min).unwrap();
+        }
+
+        // FIXME: String de-duplication is needed so that we only allocate this string only once
+        // even when there are multiple calls to this function.
+        let tm_zone_ptr = this.alloc_os_str_as_c_str(
+            OsString::from(tm_zone).as_os_str(),
```
RalfJung, line 167:
```suggestion
            &OsString::from(tm_zone),
```

tests/pass-dep/shims/libc-misc.rs
tiif: Yeah, this makes sense.

RalfJung:
All right, looks good. Thanks! Please rebase and squash.

tiif:
Thanks @RalfJung and @oli-obk for mentoring this PR! I really learned a lot from the detailed suggestion and discovered a lot of new Rust syntax that I never used before. I will be away preparing for final exams until 1 May, and I will start working on  ``socketpair`` by then.

PR #3462, opened 2024-04-14
Make `split_simd_to_128bit_chunks` take only one operand

eduardosm:
It will allow more flexible uses in the future. This makes `split_simd_to_128bit_chunks` simpler, moving some of the complexity to its callers.

### DIFF
--- a/src/shims/x86/mod.rs
+++ b/src/shims/x86/mod.rs
@@ -664,54 +664,35 @@ fn convert_float_to_int<'tcx>(
     Ok(())
 }
 
-/// Splits `left`, `right` and `dest` (which must be SIMD vectors)
-/// into 128-bit chuncks.
-///
-/// `left`, `right` and `dest` cannot have different types.
+/// Splits `op` (which must be a SIMD vector) into 128-bit chuncks.
 ///
 /// Returns a tuple where:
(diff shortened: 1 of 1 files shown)


src/shims/x86/mod.rs
```diff
@@ -664,54 +664,35 @@ fn convert_float_to_int<'tcx>(
...
-///
-/// `left`, `right` and `dest` cannot have different types.
+/// Splits `op` (which must be a SIMD vector) into 128-bit chuncks.
 ///
 /// Returns a tuple where:
 /// * The first element is the number of 128-bit chunks (let's call it `N`).
 /// * The second element is the number of elements per chunk (let's call it `M`).
-/// * The third element is the `left` vector split into chunks, i.e, it's
+/// * The third element is the `op` vector split into chunks, i.e, it's
 ///   type is `[[T; M]; N]`.
```
RalfJung, line 673:
```suggestion
///   type is `[[T; M]; N]` where `T` is the element type of `op`.
```

RalfJung:
r=me with the comment improved.

@bors delegate+

RalfJung:
Yeah bors is on strike... Again. We can just hope it sorts itself out...

PR #3467, opened 2024-04-15
Bump rustc-build-sysroot to 0.4.6

### LINKED ISSUE #3313: Broken sysroot when custom "release" profile sets `panic="abort"`

RossSmyth:
Fixes #3313

### DIFF
--- a/cargo-miri/Cargo.lock
+++ b/cargo-miri/Cargo.lock
@@ -194,9 +194,9 @@ dependencies = [
 
 [[package]]
 name = "rustc-build-sysroot"
-version = "0.4.5"
+version = "0.4.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


RalfJung:
@bors r+

Thanks!

PR #3474, opened 2024-04-16
Implement wcslen

tyilo:


src/helpers.rs
```diff
@@ -1017,6 +1017,35 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        let wchar_t = this.libc_ty_layout("wchar_t");
+
+        this.check_ptr_align(ptr, wchar_t.align.abi)?;
+
+        let mut wchars = Vec::new();
+        loop {
+            // FIXME: We are re-getting the allocation each time around the loop.
+            // Would be nice if we could somehow "extend" an existing AllocRange.
+            let alloc = this.get_ptr_alloc(ptr, wchar_t.size)?.unwrap(); // not a ZST, so we will get a result
+            let wchar = alloc.read_integer(alloc_range(Size::ZERO, wchar_t.size))?.to_u32()?;
```
oli-obk, line 1037: the `to_u32` expects 4 bytes, but on windows it only has 2

> type for wide character representation (see [wide strings](https://en.cppreference.com/w/cpp/string/wide)). It has the same size, signedness, and alignment as one of the integer types, but is a distinct type. In practice, it is 32 bits and holds UTF-32 on Linux and many other non-Windows systems, but 16 bits and holds UTF-16 code units on Windows. The standard used to require wchar_t to be large enough to represent any supported character code point. However, such requirement cannot be fulfilled on Windows, and thus it is considered as a [defect](https://en.cppreference.com/w/cpp/language/types#Defect_reports) and removed.

you'll need to use `to_bits`

RalfJung: You can grep for `off_t` to find how existing code handles "read an integer type that has platform-defined size".

src/helpers.rs
```diff
@@ -1017,6 +1017,35 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         Ok((true, string_length))
     }
 
+    /// Read a sequence of wchar_t until the first null terminator.
+    /// Always returns a `Vec<u32>` no matter the size of `wchar_t`.
+    fn read_wchar_t_str(
```
RalfJung, line 1022: Would it make sense to share the code with `read_wide_str`, and have a single function that can read a sequence of integers until hitting a 0?

tyilo: Definitely. I will try do that when the tests succeeds.

src/helpers.rs
```diff
@@ -912,30 +912,57 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         })
     }
 
-    /// Read a sequence of bytes until the first null terminator.
-    fn read_c_str<'a>(&'a self, ptr: Pointer<Option<Provenance>>) -> InterpResult<'tcx, &'a [u8]>
```
RalfJung, line 916: `read_c_str` is different than the others, since it can directly return `&'a [u8]`. It probably shouldn't be part of the unified `read_any_str` function.

src/helpers.rs
```diff
@@ -912,30 +912,57 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
-    where
-        'tcx: 'a,
-        'mir: 'a,
-    {
+    /// Generic version of strlen to support multiple character sizes.
```
RalfJung, line 915: I think you can have a generic read_str, no need to add strlen functions here.

The function should take
- a `TyAndLayout` describing the type to do the loads with
- a generic `T: TryFrom<u128>` to use for the vector that it returns

src/helpers.rs
tyilo: Isn't the same true for `read_wide_str`? My new implementation returns `&'a [u16]` instead of the previous `Vec<u16>`.

src/helpers.rs
tyilo: The point is that we only need `strlen` for implementing `strlen` and `strcpy`.

Furthermore `wcslen` only needs `wchar_t_strlen` and not some `read_w_char_t_str`.

src/helpers.rs
RalfJung: It's not possible to return `&'a [u16]`, I think. Due to endianess the data we return may not match what exists in memory.

src/helpers.rs
RalfJung: Sure, but we do need `read_wide_str`, and if we can just make that generic then it can be used by `wcslen` and the existing os_str helpers, so we have less code to maintain. :)

src/helpers.rs
tyilo: Makes sense. Do we have any big endian miri targets, so that we can test that the miri tests also works on those?

RalfJung: Yeah, CI runs the tests on a big-endian target.

tyilo: Hmm, but then the test succeeds with returning `&[u16]`. So does that mean the tests aren't good enough or that it is ok to return `&[u16]`?

src/helpers.rs
```diff
@@ -977,29 +977,42 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        this.check_ptr_align(ptr, align)?;
 
         let mut wchars = Vec::new();
         loop {
             // FIXME: We are re-getting the allocation each time around the loop.
             // Would be nice if we could somehow "extend" an existing AllocRange.
-            let alloc = this.get_ptr_alloc(ptr, size2)?.unwrap(); // not a ZST, so we will get a result
-            let wchar = alloc.read_integer(alloc_range(Size::ZERO, size2))?.to_u16()?;
-            if wchar == 0 {
+            let alloc = this.get_ptr_alloc(ptr, size)?.unwrap(); // not a ZST, so we will get a result
```
RalfJung, line 998: The function should assert that `size` is not 0, to justify this comment.

src/helpers.rs
```diff
@@ -977,29 +977,42 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+            let i = alloc.read_integer(alloc_range(Size::ZERO, size))?.to_bits(size)?;
```
RalfJung, line 999: `i` sounds like an index, but this is the character. So `wchar` seems like the better name, though I am also open to other options.

src/helpers.rs
```diff
@@ -977,29 +977,42 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         Ok((true, string_length))
     }
 
-    /// Read a sequence of u16 until the first null terminator.
-    fn read_wide_str(&self, mut ptr: Pointer<Option<Provenance>>) -> InterpResult<'tcx, Vec<u16>> {
+    fn read_c_str_with_char_size<T>(
```
RalfJung, line 980:
```suggestion
    /// Helper function to read a sequence of unsigned integers of the given size and alignment
    /// until the first null terminator.
    fn read_c_str_with_char_size<T>(
```

RalfJung (commented):
Thanks! We're entering the nit-picking stage of the review process. :)

@rustbot author

src/helpers.rs
tyilo: I went with `wchar_int` as it is a different type than `T` and `wchars` has type `Vec<T>`.

RalfJung:
Thanks, looks good! Please squash the commits.

RalfJung:
Looks good though. Thanks for the PR!

@bors r+

Issue #3439, opened 2024-04-02
can't call (diverging) foreign function: __rust_alloc_error_handler


RalfJung:
We don't support whatever magic links in the alloc error handler, it seems:
```rust
#![feature(alloc_error_hook, allocator_api)]

use std::alloc::*;
use std::ptr::NonNull;

struct BadAlloc;

unsafe impl Allocator for BadAlloc {
    fn allocate(&self, _l: Layout) -> Result<NonNull<[u8]>, AllocError> {
        Err(AllocError)
    }
    
    unsafe fn deallocate(&self, _ptr: NonNull<u8>, _layout: Layout) {
        unreachable!();
    }
}

fn main() {
    set_alloc_error_hook(|_| panic!("alloc error"));
    let _b = Box::new_in(0, BadAlloc);
}
```

```
error: unsupported operation: can't call (diverging) foreign function: __rust_alloc_error_handler
   --> /playground/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/alloc.rs:385:13
    |
385 |             __rust_alloc_error_handler(layout.size(), layout.align());
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't call (diverging) foreign function: __rust_alloc_error_handler
    |
    = help: this is likely not a bug in the program; it indicates that the program performed an operation that the interpreter does not support
    = note: BACKTRACE:
    = note: inside `std::alloc::handle_alloc_error::rt_error` at /playground/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/alloc.rs:385:13: 385:70
    = note: inside `std::alloc::handle_alloc_error` at /playground/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/alloc.rs:391:9: 391:75
    = note: inside `std::boxed::Box::<i32, BadAlloc>::new_uninit_in` at /playground/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/boxed.rs:451:23: 451:49
[...]

PR #3478, opened 2024-04-16
implement support for __rust_alloc_error_handler

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/3439

### DIFF
--- a/tests/fail/alloc/alloc_error_handler.rs
+++ b/tests/fail/alloc/alloc_error_handler.rs
@@ -1,0 +1,25 @@
+//@error-in-other-file: aborted
+//@normalize-stderr-test: "unsafe \{ libc::abort\(\) \}|crate::intrinsics::abort\(\);" -> "ABORT();"
+//@normalize-stderr-test: "\| +\^+" -> "| ^"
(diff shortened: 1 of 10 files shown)


tests/pass/alloc-access-tracking.stderr
saethlin, line 1: Do we care to change the tracked ID to keep this stderr file intact?

RalfJung: Ah, oops

tests/fail/alloc/alloc_error_handler.rs
```diff
@@ -0,0 +1,25 @@
...
+//@normalize-stderr-test: "\| +\^+" -> "| ^"
+#![feature(allocator_api)]
+
+use std::alloc::*;
+use std::ptr::NonNull;
+
+struct BadAlloc;
+
+// Create a failing allocator; Miri's native allocator never fails so this is the only way to
+// actually call the alloc error handler.
```
bjorn3, line 12: User code can directly call `std::alloc::handle_alloc_error(layout)` on stable to call the alloc error handler.

Issue #3397, opened 2024-03-23
teach miri to use intrinsics' fallback bodies


oli-obk:
And then replace lots of our SIMD and numeric  shims with fallback bodies by adding said fallback bodies to libcore.

Note that we need to take lots of care to preserve UB.

This may regress diagnostics quality, so maybe we need some sort of `unreachable_unchecked` intrinsic equivalent which can be passed a string that explains the UB?

RalfJung:
I think we may want to require some sort of opt-in here, like `#[rustc_intrinsic_fallback_fit_for_miri]` or so, so that one can locally review in the intrinsic body whether the requirement is upheld.

RalfJung:
Isn't this a duplicate of https://github.com/rust-lang/miri/issues/3302?

PR #3492, opened 2024-04-19
Implement LLVM x86 AVX2 intrinsics

eduardosm:


oli-obk:
My hope was that we'd implement them in libstd as fallback bodies instead (https://github.com/rust-lang/miri/issues/3397), but I guess now that we have a shim impl and tests, doing that would be simpler as we have some help against messing up the fallback bodies

src/shims/x86/mod.rs
```diff
@@ -534,6 +540,61 @@ fn shift_simd_by_scalar<'tcx>(
     Ok(())
 }
 
+/// Shifts each element of `left` by the corresponding element of `right`.
+///
+/// For logic shifts, when right is larger than BITS - 1, zero is produced.
+/// For arithmetic right-shifts, when right is larger than BITS - 1, the sign
+/// bit is copied to remaining bits.
```
oli-obk, line 547: which remaining bits? Isn't the sign bit right shifted together with the other bits, so right shifts larger than `BITS - 1` are equivalent to a right shift of `BITS - 1`?

src/shims/x86/mod.rs
```diff
@@ -664,6 +725,30 @@ fn convert_float_to_int<'tcx>(
     Ok(())
 }
 
+/// Calculates absolute value of integers in `op` and stores the result in `dest`.
```
oli-obk, line 728: Leave a comment about the behaviour when computing the `abs` of the negative most value representable by the type

src/shims/x86/mod.rs
```diff
@@ -874,3 +959,320 @@ fn test_high_bits_masked<'tcx>(
...
+    let high_bit_offset = mask_item_size.bits().checked_sub(1).unwrap();
+
+    let ptr = this.read_pointer(ptr)?;
+    for i in 0..dest_len {
+        let mask = this.project_index(&mask, i)?;
+        let dest = this.project_index(&dest, i)?;
+
+        if this.read_scalar(&mask)?.to_uint(mask_item_size)? >> high_bit_offset != 0 {
+            // Size * u64 is implemented as always checked
+            #[allow(clippy::arithmetic_side_effects)]
```
oli-obk, line 986: Could add a `clippy.toml` and disable this lint for all `Size` math ops: https://doc.rust-lang.org/clippy/lint_configuration.html#arithmetic-side-effects-allowed

src/shims/x86/avx2.rs
```diff
@@ -0,0 +1,444 @@
...
+                let actual_len = dest_len.min(offsets_len);
+
+                assert_eq!(dest_len, mask_len);
+
+                let mask_item_size = mask.layout.field(this, 0).size;
+                let high_bit_offset = mask_item_size.bits().checked_sub(1).unwrap();
+
+                let scale = this.read_scalar(scale)?.to_i8()?;
+                if !matches!(scale, 1 | 2 | 4 | 8) {
+                    throw_unsup_format!("invalid gather scale {scale}");
```
oli-obk, line 83: are these absolutely illegal? or just not supported by llvm right now? If this is entirely wrong, it should probably be `ub` instead of `unsup`

tests/pass/intrinsics-x86-avx2.rs
```diff
@@ -0,0 +1,1613 @@
+// Ignore everything except x86 and x86_64
+// Any new targets that are added to CI should be ignored here.
+// (We cannot use `cfg`-based tricks here since the `target-feature` flags below only work on x86.)
+//@ignore-target-aarch64
+//@ignore-target-arm
+//@ignore-target-avr
+//@ignore-target-s390x
+//@ignore-target-thumbv7em
+//@ignore-target-wasm32
```
oli-obk, line 9: can we use `only-target-x86` instead?

src/shims/x86/avx2.rs
RalfJung: I think they just don't exist on the hardware -- these are basically ill-formed programs. Maybe we should have `throw_inval_format!`?

tests/pass/intrinsics-x86-avx2.rs
eduardosm: Does `only-target-x86` cover i686?

src/shims/x86/avx2.rs
eduardosm: There is `throw_inval!` in rustc_middle, but not `throw_inval_format!`. We would have to add a variant to `InvalidProgramInfo` to use it.

src/shims/x86/mod.rs
eduardosm: I meant all bits, fixed

src/shims/x86/avx2.rs
RalfJung: Indeed, that's what I was proposing.

But so far we've usually been using "unsup" for "inval" so it's fine for me to keep doing that for now.

RalfJung: Since this is an intrinsic, the other totally reasonable alternative is to use `span_bug!`. We have plenty of assertions already that just cause a panic on invalid use of intrinsics.

tests/pass/intrinsics-x86-avx2.rs
oli-obk: hmno. we need https://github.com/oli-obk/ui_test/issues/162 first

src/shims/x86/mod.rs
```diff
@@ -664,6 +725,33 @@ fn convert_float_to_int<'tcx>(
...
+    let (dest, dest_len) = this.mplace_to_simd(dest)?;
+
+    assert_eq!(op_len, dest_len);
+
+    for i in 0..dest_len {
+        let op = this.read_scalar(&this.project_index(&op, i)?)?;
+        let dest = this.project_index(&dest, i)?;
+
+        // Converting to a host "i128" works since the input is always signed.
+        let res = op.to_int(dest.layout.size)?.unsigned_abs();
```
RalfJung, line 747: This may be worth an assertion.

Alternatively, a moire clear implementation might avoid using host integers entirely and just use `binop_with_overflow`.

src/shims/x86/mod.rs
```diff
@@ -874,3 +962,316 @@ fn test_high_bits_masked<'tcx>(
...
+        #[allow(clippy::cast_possible_truncation)]
+        let res = res as i16;
+
+        this.write_scalar(Scalar::from_i16(res), &dest)?;
+    }
+
+    Ok(())
+}
+
+fn pack_generic<'tcx>(
```
RalfJung, line 1130: A doc comment would be good.

src/shims/x86/avx2.rs
```diff
@@ -0,0 +1,444 @@
...
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                assert_eq!(dest.layout, src.layout);
+
+                let (src, _) = this.operand_to_simd(src)?;
+                let (offsets, offsets_len) = this.operand_to_simd(offsets)?;
+                let (mask, mask_len) = this.operand_to_simd(mask)?;
+                let (dest, dest_len) = this.mplace_to_simd(dest)?;
+
+                // There are cases like dest: i32x4, offsets: i64x2
```
RalfJung, line 73: What happens if the offsets are *longer* than the dest?

src/shims/x86/mod.rs
eduardosm: I'm not sure what to check with an assertion. How would you implement this with `binop_with_overflow`?

src/shims/x86/avx2.rs
eduardosm: Extra elements are ignored.

src/shims/x86/mod.rs
RalfJung: The assertion could check that the element type is signed, no?

>  How would you implement this with binop_with_overflow?

Sorry, I meant `wrapping_binary_op`.

Basically run the equivalent of `if op < 0 { -op } else { op }`. It's one `BinOp::lt` and one... okay you need `wrapping_unary_op` as well, `UnOp::Neg`.

src/shims/x86/avx2.rs
RalfJung: And that's how the intrinsic is intended to work? Wow.

Would be good to have a comment explaining that.

PR #3493, opened 2024-04-19
Add `-Zmiri-env-set` to set environment variables without modifying the host environment

eduardosm:
This option allows to pass environment variables to the interpreted program without needing to modify the host environment (which may have undesired effects in some cases).

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,7 +311,7 @@ environment variable. We first document the most relevant and most commonly used
 * `-Zmiri-env-forward=<var>` forwards the `var` environment variable to the interpreted program. Can
   be used multiple times to forward several variables. Execution will still be deterministic if the
   value of forwarded variables stays the same. Has no effect if `-Zmiri-disable-isolation` is set.
-* `-Zmiri-env-set=<var>=<value>` sets the `var` environment variable to `value` in the interpreted.
+* `-Zmiri-env-set=<var>=<value>` sets the `var` environment variable to `value` in the interpreted program.
   It can be used to pass environment variables without needing to alter the host environment. It can
   be used multiple times to set several variables. If `-Zmiri-disable-isolation` or `-Zmiri-env-forward`
   is set, values set with this option will have priority over values from the host environment.
--- a/README.md
+++ b/README.md
@@ -311,6 +311,10 @@ environment variable. We first document the most relevant and most commonly used
 * `-Zmiri-env-forward=<var>` forwards the `var` environment variable to the interpreted program. Can
   be used multiple times to forward several variables. Execution will still be deterministic if the
   value of forwarded variables stays the same. Has no effect if `-Zmiri-disable-isolation` is set.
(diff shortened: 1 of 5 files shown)


RalfJung:
Doesn't rustc have some flag to sent env vars nowadays? Should we just reuse that instead of adding our own?

eduardosm:
I'm not aware of such flag

RalfJung:
https://github.com/rust-lang/compiler-team/issues/653 was the decision to add it. Tracking issue is at https://github.com/rust-lang/rust/issues/118372, so it seems to have been implemented.

eduardosm:
Do we want the same flag for compile-time `env!` and run-time `std::env::var`?

README.md
```diff
@@ -311,6 +311,10 @@ environment variable. We first document the most relevant and most commonly used
 * `-Zmiri-env-forward=<var>` forwards the `var` environment variable to the interpreted program. Can
   be used multiple times to forward several variables. Execution will still be deterministic if the
   value of forwarded variables stays the same. Has no effect if `-Zmiri-disable-isolation` is set.
+* `-Zmiri-env-set=<var>=<value>` sets the `var` environment variable to `value` in the interpreted.
+  It can be used to pass environment variables without needing to alter the host environment. It can
+  be used multiple times to set several variables. If `-Zmiri-disable-isolation` is set, values set
```
RalfJung, line 316:
```suggestion
  be used multiple times to set several variables. If `-Zmiri-disable-isolation` or `-Zmiri-env-forward` is set, values set
```

tests/pass/shims/env/var-set.rs
```diff
@@ -0,0 +1,7 @@
+// Test both a value set and unset in the host.
```
RalfJung, line 1:
```suggestion
// Test a value set on the host (MIRI_ENV_VAR_TEST) and one that is not.
```

RalfJung:
> Do we want the same flag for compile-time `env!` and run-time `std::env::var`?

Fair, I guess it makes sense to treat them separately.

README.md
```diff
@@ -311,6 +311,10 @@ environment variable. We first document the most relevant and most commonly used
 * `-Zmiri-env-forward=<var>` forwards the `var` environment variable to the interpreted program. Can
   be used multiple times to forward several variables. Execution will still be deterministic if the
   value of forwarded variables stays the same. Has no effect if `-Zmiri-disable-isolation` is set.
+* `-Zmiri-env-set=<var>=<value>` sets the `var` environment variable to `value` in the interpreted.
```
oli-obk, line 314:
```suggestion
* `-Zmiri-env-set=<var>=<value>` sets the `var` environment variable to `value` in the interpreted program.
```

oli-obk (changes requested):

oli-obk (approved):

Issue #3499, opened 2024-04-22
Windows home_dir fallback when HOME is not set not supported


RalfJung:
```
  0.159555   error: unsupported operation: can't call foreign function `GetUserProfileDirectoryW` on OS `windows`
  0.000051      --> /home/runner/work/miri-test-libstd/miri-test-libstd/rust-src-patched/library/std/src/sys/pal/windows/os.rs:329:23
  0.000008       |
  0.000008   329 |                   match c::GetUserProfileDirectoryW(
  0.000005       |  _______________________^
  0.000006   330 | |                     ptr::without_provenance_mut(CURRENT_PROCESS_TOKEN),
  0.000006   331 | |                     buf,
  0.000005   332 | |                     &mut sz,
  0.000009   333 | |                 ) {
  0.000010       | |_________________^ can't call foreign function `GetUserProfileDirectoryW` on OS `windows`
```

PR #3502, opened 2024-04-23
windows: basic support for GetUserProfileDirectoryW

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/3499

### DIFF
--- a/src/shims/env.rs
+++ b/src/shims/env.rs
@@ -178,7 +178,6 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                     &var,
                     buf_ptr,
                     buf_size.into(),
-                    /*truncate*/ false,
                 )?))
                 // This can in fact return 0. It is up to the caller to set last_error to 0
                 // beforehand and check it afterwards to exclude that case.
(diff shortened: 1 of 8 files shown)


src/shims/env.rs
```diff
@@ -482,9 +482,65 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                // a different method for dealing with a too-small buffer than the other functions...
+                let (success, len) = this.write_path_to_wide_str(
+                    home,
+                    buf,
+                    size_avail.into(),
+                    /*truncate*/ false,
+                )?;
+                // The Windows docs just say that this is written on failure. But std
+                // seems to rely on it always being written.
+                this.write_scalar(Scalar::from_u32(len.try_into().unwrap()), &size)?;
```
RalfJung, line 530: @ChrisDenton something seems to be odd here, or I am misunderstanding something. The [docs](https://learn.microsoft.com/en-us/windows/win32/api/userenv/nf-userenv-getuserprofiledirectoryw) say about the size pointer

> If the buffer specified by lpProfileDir is not large enough or lpProfileDir is NULL, the function fails and this parameter receives the necessary buffer size, including the terminating null character.

So in particular, if the buffer is large enough, it doesn't say that the pointer is changed at all.

But the way std calls this function, it assumes that on success, `sz` is updated to the actual written size *including* the null terminator.

ChrisDenton: You're right. We shouldn't be doing that if the behaviour isn't documented.

Given that's how it's worked in practice and Rust has apparently been relying on this for a long time, I do think updating the API docs is warranted. However, unless or until that actually happens we should act as if such a documentation change would be rejected.

RalfJung: FWIW wine also does the always-update-size thing.

Thanks for confirming! I will then land this PR (if I stopped updating the size the tests would fail). I opened an issue to track this: https://github.com/rust-lang/rust/issues/124325.

PR #3511, opened 2024-04-24
Upgrade to `actions/checkout@v4` in `ci.yml`.

obi1kenobi:
This is a newer version of the same action. None of the uses here were particularly special (no complex features of v3 were used) so this is a straightforward as-is upgrade.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
     env:
       HOST_TARGET: ${{ matrix.host_target }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Show Rust version (stable toolchain)
         run: |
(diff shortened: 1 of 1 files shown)


RalfJung:
Thanks! I don't understand why github makes everyone do this every other year or so... oh well.

@bors r+

obi1kenobi:
There's technically a breaking change between v3 and v4, just one that affects very few use cases in practice. So I think they did the right thing, even though it's a bit annoying that almost everyone has to actively upgrade even though they aren't affected.

Btw, GitHub's Dependabot can keep actions updated to the latest versions -- I have that configured [in cargo-semver-checks](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/.github/dependabot.yml) and it's worked well so far. If you'd like, I'd be happy to add a similar config here: once a week, it will open PRs with any upgraded actions versions that are available.

RalfJung:
Thanks! I think we'll wait for other rust-lang projects to do something like that before considering it.

obi1kenobi:
Makes sense. Open offer, ping me anytime if you or someone in another rust-lang project wants to try it. Happy to help.

Issue #3522, opened 2024-04-27
localtime_r shim uses TZ variable from the host, not the interpreted program


saethlin:
Currently the test suite may fail to pass, unless you run it with `TZ=UTC ./miri test`. This is because the `localtime_r` shim reads the `TZ` variable from the host on this line: https://github.com/rust-lang/miri/blob/f26bd28c4d0e96a3b855b7f544030ee593b3090a/src/shims/time.rs#L139-L140

In other words, this test code is ineffective: https://github.com/rust-lang/miri/blob/f26bd28c4d0e96a3b855b7f544030ee593b3090a/tests/pass-dep/shims/libc-misc.rs#L220-L222

I'm going to look into this.

cc @tiif @eduardosm

tiif:
Thanks for looking into this, and sorry for causing this. I ran into this issue while testing and discussed with @oli-obk before, but I thought there is something wrong with my machine, because  ``env::set_var`` works when I isolated it out in a test.
 
This is how the test looks like: 
In ``tests/pass-dep/shims/libc-misc.rs``
```rust
fn test_tz_var_setting() {
    use chrono::Local;
    use std::env;

    let key = "TZ";
    env::set_var(key, "GMT");
    assert_eq!(env::var(key), Ok("GMT".to_string()));
    let offset_in_second = Local::now().offset().local_minus_utc();
    // If the timezone is GMT, tm_gmtoff will be 0
    let tm_gmtoff = offset_in_second;
    assert_eq!(tm_gmtoff, 0);
    env::remove_var(key);
    assert!(env::var(key).is_err());
}
```
In ``test_dependencies/Cargo.toml``, add ``chrono`` to dependencies: 
```rust
[dependencies]
# all dependencies (and their transitive ones) listed here can be used in `tests/`.
libc = "0.2"
num_cpus = "1.10.1"
tempfile = "3"
chrono = "0.4.37"
```

This test still pass after I merged the latest update from master (commit ``f26bd28c``).

RalfJung:
> This is how the test looks like:

That's using `chrono::Local` *inside* the program, so when it does `env::get` it sees the variables of the interpreted program.

However, the shim uses `chrono::Local` on the host, so that will use the host's `env::get` instead of the one of the interpreted program.

tiif:
> That's using `chrono::Local` _inside_ the program, so when it does `env::get` it sees the variables of the interpreted program.

[...]

PR #3523, opened 2024-04-27
Use the interpreted program's TZ variable in localtime_r

saethlin:
This requires a bit of wiring and a new dependency, but the tests should correctly pass now regardless of what the host's time zone is.

Fixes https://github.com/rust-lang/miri/issues/3522

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,21 +38,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
-
-[[package]]
-name = "android_system_properties"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
-dependencies = [
- "libc",
-]
-
-[[package]]
 name = "annotate-snippets"
 version = "0.9.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -122,12 +107,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bumpalo"
-version = "3.16.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "79296716171880943b8470b5f8d03aa55eb2e645a4874bdbb28adb49162e012c"
-
-[[package]]
 name = "camino"
 version = "1.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -177,10 +156,29 @@ version = "0.4.38"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a21f936df1771bf62b77f047b726c4625ff2e8aa607c01ec06e5a05bd8463401"
 dependencies = [
- "android-tzdata",
- "iana-time-zone",
  "num-traits",
- "windows-targets 0.52.3",
+]
+
+[[package]]
+name = "chrono-tz"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "93698b29de5e97ad0ae26447b344c482a7284c737d9ddc5f9e52b74a336671bb"
+dependencies = [
+ "chrono",
+ "chrono-tz-build",
+ "phf",
+]
+
+[[package]]
+name = "chrono-tz-build"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0c088aee841df9c3041febbb73934cfc39708749bf96dc827e3359cd39ef11b1"
+dependencies = [
+ "parse-zoneinfo",
+ "phf",
+ "phf_codegen",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


Cargo.lock
```diff
@@ -690,6 +712,15 @@ dependencies = [
  "windows-targets 0.48.5",
 ]
 
+[[package]]
+name = "parse-zoneinfo"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f4ee19a3656dadae35a33467f9714f1228dd34766dbe49e10e656b5296867aea"
+dependencies = [
+ "regex 0.2.11",
```
RalfJung, line 721: This is an ancient version of the regex crate -- is parse-zoneinfo even still maintained?

I don't think it would be good to bring such an ancient regex back into the rustc workspace.

RalfJung: Ah the main issue is that this version of parse-zoneinfo is from 6 years ago. There's version 0.3.0 from 4 years ago that uses a more recent `regex`.

I don't think we should add a dependency on regex 0.2. That's just too old.

Cargo.toml
```diff
@@ -25,6 +25,7 @@ aes = { version = "0.8.3", features = ["hazmat"] }
 measureme = "11"
 ctrlc = "3.2.5"
 chrono = { version = "0.4.38", default-features = false, features = ["clock"] }
+chrono-tz = "0.4"
```
RalfJung, line 28: Why did you add such an ancient version of chrono-tz? The current version is 0.9.

src/shims/time.rs
```diff
@@ -136,8 +138,16 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
             .unwrap();
         let dt_utc: DateTime<Utc> =
             DateTime::from_timestamp(sec_since_epoch, 0).expect("Invalid timestamp");
+
+        // Figure out what time zone is in use
+        let tz = this.get_var(OsStr::new("TZ"))?.unwrap_or_else(|| OsString::from("UTC"));
+        let tz = match tz.into_string() {
+            Ok(tz) => Tz::from_str(&tz).unwrap_or(Tz::Zulu),
+            _ => Tz::Zulu,
```
RalfJung, line 146: Why is Zulu the fallback, not UTC...?

src/shims/time.rs
```diff
@@ -146,17 +156,17 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
 
         // tm_zone represents the timezone value in the form of: +0730, +08, -0730 or -08.
         // This may not be consistent with libc::localtime_r's result.
-        let offset_in_second = Local::now().offset().local_minus_utc();
-        let tm_gmtoff = offset_in_second;
+        let offset_in_seconds = Utc::now().with_timezone(&tz).offset().fix().local_minus_utc();
```
RalfJung, line 159: Can't `dt` be used for this, why do we need `Utc::now`?

Cargo.toml
```diff
@@ -25,6 +25,7 @@ aes = { version = "0.8.3", features = ["hazmat"] }
 measureme = "11"
 ctrlc = "3.2.5"
 chrono = { version = "0.4.38", default-features = false, features = ["clock"] }
```
RalfJung, line 27: If I understand the docs correctly, we don't need the `clock` feature any more?

Cargo.toml
saethlin: Probably because 4 looks like 9.

src/shims/unix/env.rs
```diff
@@ -70,6 +70,27 @@ impl<'tcx> UnixEnvVars<'tcx> {
...
+        let _vars_ptr = ecx.read_pointer(&self.environ)?;
+        let Some(var_ptr) = self.map.get(name) else {
+            return Ok(None);
+        };
+        // The offset is used to strip the "{name}=" part of the string.
+        let var_ptr = var_ptr.offset(
+            Size::from_bytes(u64::try_from(name.len()).unwrap().checked_add(1).unwrap()),
+            ecx,
+        )?;
+        ecx.read_os_str_from_c_str(var_ptr).map(|s| Some(s.to_owned()))
```
RalfJung, line 92: This duplicates a bunch of code in `getenv`, doesn't it? Can the code be shared, e.g. by having getenv call this?

src/shims/env.rs
```diff
@@ -99,4 +99,15 @@ impl<'tcx> EnvVars<'tcx> {
 }
 
 impl<'mir, 'tcx: 'mir> EvalContextExt<'mir, 'tcx> for crate::MiriInterpCx<'mir, 'tcx> {}
-pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {}
+pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
+    /// Try to get an environment variable from the interpreted program's environment. This is
+    /// useful for implementing shims which are documented to read from the environment.
+    fn get_var(&mut self, name: &OsStr) -> InterpResult<'tcx, Option<OsString>> {
```
RalfJung, line 105: `get_env_var` may be better since this is a global namespace.

src/shims/unix/env.rs
saethlin: It does duplicate some code. `getenv` can't call it as-written but with some rearrangement it's possible to share the logic.

PR #3526, opened 2024-04-28
Adjust Allocation Bytes used by Miri to custom MiriAllocBytes

Strophox:
Previously, the `MiriMachine` used `type Bytes = Box<[u8]>` for its allocations.
This PR swaps this out for a custom `MiriAllocBytes` type implemented in `alloc_bytes.rs`.
This is in anticipation of an extension to Miri's FFI, which will require its allocations to take care of alignment (the methods in `impl AllocBytes for Box<[u8]>` ignore this `_align: Align` argument).

Needs https://github.com/rust-lang/rust/pull/124492

### DIFF
--- a/src/alloc_bytes.rs
+++ b/src/alloc_bytes.rs
@@ -1,0 +1,109 @@
+use std::alloc;
+use std::alloc::Layout;
+use std::borrow::Cow;
+use std::slice;
+
+use rustc_middle::mir::interpret::AllocBytes;
+use rustc_target::abi::{Align, Size};
+
+/// Allocation bytes that explicitly handle the layout of the data they're storing.
+/// This is necessary to interface with native code that accesses the program store in Miri.
+#[derive(Debug)]
+pub struct MiriAllocBytes {
+    /// Stored layout information about the allocation.
+    layout: alloc::Layout,
+    /// Pointer to the allocation contents.
+    /// Invariant:
+    /// * If `self.layout.size() == 0`, then `self.ptr` is some suitably aligned pointer
+    ///   without provenance (and no actual memory was allocated).
+    /// * Otherwise, `self.ptr` points to memory allocated with `self.layout`.
+    ptr: *mut u8,
+}
+
+impl Clone for MiriAllocBytes {
+    fn clone(&self) -> Self {
+        let bytes: Cow<'_, [u8]> = Cow::Borrowed(self);
+        let align = Align::from_bytes(self.layout.align().try_into().unwrap()).unwrap();
+        MiriAllocBytes::from_bytes(bytes, align)
+    }
+}
+
+impl Drop for MiriAllocBytes {
+    fn drop(&mut self) {
+        if self.layout.size() != 0 {
+            // SAFETY: Invariant, `self.ptr` points to memory allocated with `self.layout`.
+            unsafe { alloc::dealloc(self.ptr, self.layout) }
+        }
+    }
+}
+
+impl std::ops::Deref for MiriAllocBytes {
+    type Target = [u8];
+
+    fn deref(&self) -> &Self::Target {
+        // SAFETY: `ptr` is non-null, properly aligned, and valid for reading out `self.layout.size()`-many bytes.
+        // Note that due to the invariant this is true even if `self.layout.size() == 0`.
+        unsafe { slice::from_raw_parts(self.ptr, self.layout.size()) }
+    }
+}
+
+impl std::ops::DerefMut for MiriAllocBytes {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        // SAFETY: `ptr` is non-null, properly aligned, and valid for reading out `self.layout.size()`-many bytes.
+        // Note that due to the invariant this is true even if `self.layout.size() == 0`.
+        unsafe { slice::from_raw_parts_mut(self.ptr, self.layout.size()) }
+    }
+}
+
+impl MiriAllocBytes {
+    /// This method factors out how a `MiriAllocBytes` object is allocated,
+    /// specifically given an allocation function `alloc_fn`.
+    /// `alloc_fn` is only used if `size != 0`.
+    /// Returns `Err(layout)` if the allocation function returns a `ptr` that is `ptr.is_null()`.
+    fn alloc_with(
+        size: usize,
+        align: usize,
+        alloc_fn: impl FnOnce(Layout) -> *mut u8,
+    ) -> Result<MiriAllocBytes, Layout> {
+        let layout = Layout::from_size_align(size, align).unwrap();
+        let ptr = if size == 0 {
+            std::ptr::without_provenance_mut(align)
+        } else {
+            let ptr = alloc_fn(layout);
+            if ptr.is_null() {
+                return Err(layout);
+            }
+            ptr
+        };
+        // SAFETY: All `MiriAllocBytes` invariants are fulfilled.
+        Ok(Self { ptr, layout })
+    }
+}
+
+impl AllocBytes for MiriAllocBytes {
+    fn from_bytes<'a>(slice: impl Into<Cow<'a, [u8]>>, align: Align) -> Self {
+        let slice = slice.into();
+        let size = slice.len();
+        let align = align.bytes_usize();
+        // SAFETY: `alloc_fn` will only be used if `size != 0`.
+        let alloc_fn = |layout| unsafe { alloc::alloc(layout) };
+        let alloc_bytes = MiriAllocBytes::alloc_with(size, align, alloc_fn)
+            .unwrap_or_else(|layout| alloc::handle_alloc_error(layout));
+        // SAFETY: `alloc_bytes.ptr` and `slice.as_ptr()` are non-null, properly aligned
+        // and valid for the `size`-many bytes to be copied.
+        unsafe { alloc_bytes.ptr.copy_from(slice.as_ptr(), size) };
+        alloc_bytes
+    }
+
+    fn zeroed(size: Size, align: Align) -> Option<Self> {
+        let size = size.bytes_usize();
+        let align = align.bytes_usize();
+        // SAFETY: `alloc_fn` will only be used if `size != 0`.
+        let alloc_fn = |layout| unsafe { alloc::alloc_zeroed(layout) };
+        MiriAllocBytes::alloc_with(size, align, alloc_fn).ok()
+    }
+
+    fn as_mut_ptr(&mut self) -> *mut u8 {
+        self.ptr
+    }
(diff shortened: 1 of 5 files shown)


src/alloc_bytes.rs
```diff
@@ -0,0 +1,113 @@
+use std::alloc;
+use std::alloc::Layout;
+use std::slice;
+use std::borrow::Cow;
+
+use rustc_target::abi::{Align, Size};
+use rustc_middle::mir::interpret::AllocBytes;
+
+/// Allocation bytes that explicitly handle the layout of the data they're storing.
+/// This is necessary to interface with native code that accesses the program store in Miri.
```
RalfJung, line 10:
```suggestion
/// The bytes of a Miri allocation, allocated using the size and alignment requested
/// by the interpreted program.
/// This is necessary to interface with native code that exchanges pointers with the
/// interpreted program.
```

src/alloc_bytes.rs
```diff
@@ -0,0 +1,113 @@
...
+#[derive(Debug)]
+#[derive(PartialEq, Eq, Hash)] // TODO Necessary?
```
RalfJung, line 12: They should not be necessary. What happens if you remove them?

IIRC we said that https://github.com/rust-lang/rust/pull/124492 should change [this line](https://github.com/rust-lang/rust/blob/7be47b219df42cba96560a97de707967cca8c762/compiler/rustc_middle/src/mir/interpret/allocation.rs#L32) to remove the `Eq + PartialEq + Hash`.

src/alloc_bytes.rs
```diff
@@ -0,0 +1,113 @@
...
+pub struct MiriAllocBytes {
+    /// Stored layout information about the allocation.
+    layout: alloc::Layout,
+    /// Pointer to the first byte of the allocation.
+    /// Invariant: if `self.layout.size() == 0`
+    /// * then `self.ptr == self.layout.align()`
+    /// * else `self.ptr` points to memory allocated with `self.layout`.
```
RalfJung, line 19:
```suggestion
    /// Invariant: 
    /// * If `self.layout.size() == 0`, then `self.ptr` is some suitably aligned pointer
    ///   without provenance (and no actual memory was allocated).
    /// * Otherwise, `self.ptr` points to memory allocated with `self.layout`. 
```

src/alloc_bytes.rs
```diff
@@ -0,0 +1,113 @@
...
+use rustc_middle::mir::interpret::AllocBytes;
+
+/// Allocation bytes that explicitly handle the layout of the data they're storing.
+/// This is necessary to interface with native code that accesses the program store in Miri.
+#[derive(Debug)]
+#[derive(PartialEq, Eq, Hash)] // TODO Necessary?
+pub struct MiriAllocBytes {
+    /// Stored layout information about the allocation.
+    layout: alloc::Layout,
+    /// Pointer to the first byte of the allocation.
```
RalfJung, line 16:
```suggestion
    /// Pointer to the allocation contents.
```

src/alloc_bytes.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl MiriAllocBytes {
+    /// This method factors out how a `MiriAllocBytes` object is allocated,
+    /// specifically given an allocation function `alloc_fn`.
+    /// `alloc_fn` is only used if `size != 0`.
+    /// Returns `Err(layout)` if the allocation function returns a `ptr` that is `ptr.is_null()`.
+    fn alloc_with(size: usize, align: usize, alloc_fn: impl FnOnce(Layout) -> *mut u8) -> Result<MiriAllocBytes, Layout> {
+        let layout = Layout::from_size_align(size, align).unwrap();
+        let ptr = if size == 0 {
+            align as *mut u8 // std::ptr::without_provenance_mut(align) : "use of unstable library feature 'strict_provenance'"
```
RalfJung, line 73: You can just use that unstable feature, it's fine and expresses the intent better.

src/alloc_bytes.rs
```diff
@@ -0,0 +1,113 @@
...
+        } else {
+            alloc_fn(layout)
+        };
+        if ptr.is_null() {
```
RalfJung, line 77: You're now checking this even in the `size == 0` case, which is unnecessary. You can move this `if` inside the `else { ... }` above.

src/alloc_bytes.rs
```diff
@@ -0,0 +1,113 @@
...
+    layout: alloc::Layout,
+    /// Pointer to the first byte of the allocation.
+    /// Invariant: if `self.layout.size() == 0`
+    /// * then `self.ptr == self.layout.align()`
+    /// * else `self.ptr` points to memory allocated with `self.layout`. 
+    ptr: *mut u8,
+}
+
+impl Clone for MiriAllocBytes {
+    fn clone(&self) -> Self {
```
RalfJung, line 24: Can you implement this using `MiriAllocBytes:: from_bytes(Cow::Borrowed(&*self), ...)`? Then you don't need unsafe here. :)

src/alloc_bytes.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn from_bytes<'a>(slice: impl Into<Cow<'a, [u8]>>, align: Align) -> Self {
+        let slice = slice.into();
+        let size = slice.len();
+        let align = align.bytes().try_into().unwrap();
+        // SAFETY: `alloc_fn` will only be used if `size != 0`.
+        let alloc_fn = |layout| unsafe { alloc::alloc(layout) };
+        let alloc_bytes = MiriAllocBytes::alloc_with(size, align, alloc_fn)
+            .unwrap_or_else(|layout| alloc::handle_alloc_error(layout));
+        // SAFETY: `alloc_bytes.ptr` and `slice.as_ptr()` are non-null, properly aligned
+        // and valid for the `slice.len()`-many bytes to be copied.
```
RalfJung, line 97:
```suggestion
        // and valid for the `size`-many bytes to be copied.
```

src/alloc_bytes.rs
```diff
@@ -0,0 +1,113 @@
...
+        let align = align.bytes().try_into().unwrap();
+        // SAFETY: `alloc_fn` will only be used if `size != 0`.
+        let alloc_fn = |layout| unsafe { alloc::alloc_zeroed(layout) };
+        MiriAllocBytes::alloc_with(size, align, alloc_fn).ok()
+    }
+
+    fn as_mut_ptr(&mut self) -> *mut u8 {
+        self.ptr
+    }
+}
```
RalfJung, line 113: Please add a newline at the end. (`./miri fmt` should do that automatically, and also otherwise will make sure you use standard Rust formatting.)

src/provenance_gc.rs
```diff
@@ -122,7 +122,7 @@ impl VisitProvenance for OpTy<'_, Provenance> {
     }
 }
 
-impl VisitProvenance for Allocation<Provenance, AllocExtra<'_>> {
+impl<Bytes: AllocBytes> VisitProvenance for Allocation<Provenance, AllocExtra<'_>, Bytes> {
```
RalfJung, line 125:
```suggestion
impl VisitProvenance for Allocation<Provenance, AllocExtra<'_>, MiriAllocBytes> {
```

RalfJung (commented):
Thanks, this is a good start!

src/alloc_bytes.rs
```diff
@@ -0,0 +1,113 @@
...
+            Ok(alloc_bytes)
+        }
+    }
+}
+
+impl AllocBytes for MiriAllocBytes {
+    fn from_bytes<'a>(slice: impl Into<Cow<'a, [u8]>>, align: Align) -> Self {
+        let slice = slice.into();
+        let size = slice.len();
+        let align = align.bytes().try_into().unwrap();
```
RalfJung, line 91: After rebasing over the latest Miri master, you should be able to use `align.bytes_usize()` here.

src/alloc_bytes.rs
Strophox: Currently when I remove them it complains about it at `impl AllocBytes for MiriAllocBytes`. I'll remove this line and incorporate the change of `AllocBytes` inside Rust itself in the other PR.

src/alloc_bytes.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl Clone for MiriAllocBytes {
+    fn clone(&self) -> Self {
+        let size = self.layout.size();
+        let align = self.layout.align();
+        // SAFETY: `alloc_fn` will only be used if `size != 0`.
+        let alloc_fn = |layout| unsafe { alloc::alloc(layout) };
+        let alloc_clone = MiriAllocBytes::alloc_with(size, align, alloc_fn)
+            .unwrap_or_else(|layout| alloc::handle_alloc_error(layout));
+        // SAFETY: `alloc_bytes.ptr` and `slice.as_ptr()` are non-null, properly aligned
```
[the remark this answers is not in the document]
Strophox: Indeed; I've adapted the comment to the modifications I had made to the code now.

src/alloc_bytes.rs
RalfJung: Yeah, this will have to wait until https://github.com/rust-lang/rust/pull/124492 lands. So let's finish that PR first. :)

RalfJung:
If you rebase over the latest Miri master, you should get the updated rustc that includes your PR, so that this one can finally build. :)

src/alloc_bytes.rs
Strophox: Forgot to run `./miri fmt` before this, will be included in the following commits!

src/alloc_bytes.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// Invariant:
+    /// * If `self.layout.size() == 0`, then `self.ptr` is some suitably aligned pointer
+    ///   without provenance (and no actual memory was allocated).
+    /// * Otherwise, `self.ptr` points to memory allocated with `self.layout`.
+    ptr: *mut u8,
+}
+
+impl Clone for MiriAllocBytes {
+    fn clone(&self) -> Self {
+        let bytes: Cow<'_, [u8]> = Cow::Borrowed(&*self); // TODO
```
RalfJung, line 25:
```suggestion
        let bytes: Cow<'_, [u8]> = Cow::Borrowed(self);
```
Clippy notes that this can be simplified

RalfJung:
Looks good, thanks!

Please squash this into a single commit.

RalfJung:
Thanks!

Force-pushes do not always trigger a notification on github, so you have to let the reviewer know that the PR is ready again. In Rust we have the `@rustbot ready` command for that.

@bors r+

PR #3533, opened 2024-05-02
Make file descriptors into refcount references

### LINKED ISSUE #3525: Make file descriptors into refcounted references to "file descriptions"
Currently, every file descriptor has to implement `dup`, and some do that incorrectly. The issue is that file descriptors are more like `Arc` than like `Box`, so all state should be shared among all the duplicates, which typically requires an extra level of indirection -- e.g. in eventfd or epoll, where duplicate file descriptors should reference the same underlying state.

Instead of making each FD type handle that, we should remove `dup` from the `FileDescriptor` trait and handle this ourselves. I see two ways to do that:
- Literally use `Arc<dyn FileDescription>` instead of `Box<dyn FileDescriptor>`, so `Arc` handles the refcounting for us for free. However this will require interior mutability if the file description tracks mutable state (as many of the "special" FDs will).
- Add some sort of `FxHashMap<FileDescriptionId, (usize, Box<dyn FileDescription>)>` (where the `usize` is a refcount) and then make the FD table simply reference `FileDescriptionId`. This means we have to do refcounting ourselves but we can get exclusive access to the data inside a file description.

Interior mutability is annoying, but then there is some chance we'll need to give the `read`/`write` methods access to the machine (even mutable access!) and then it seems hard to avoid some form of interior mutability. So currently my gut feeling is that manual refcounting isn't worth it.

Luv-Ray:
fixes #3525 

Remove `fn dup` in `trait FileDescription`, define `struct FileDescriptor(Rc<RefCell<dyn FileDescription>>)`, and use `BTreeMap<i32, FileDescriptor>` in `FdTable`.

---

There are some refactors similar to the following form:
```rust
{  // origin:
    if let Some(file_descriptor) = this.machine.fds.get_mut(fd) {
        // write file_descriptor
    } else {
        this.fd_not_found()
    }
}
{  // now:
    let Some(mut file_descriptor) = this.machine.fds.get_mut(fd) else {
        return this.fd_not_found();
    };
    // write file_descriptor
    drop(file_descriptor);
}
```
The origin form can't compile because as using `RefCell` to get interior mutability, `fn get_mut` return `Option<std::cell::RefMut<'_, dyn FileDescription>>` instead of `Option<&mut dyn FileDescription>` now, and the `deref_mut` on `file_descriptor: RefMut` will cause borrow `this` as mutable more than once at a time. 
So this form of refactors and manual drops are are implemented to avoid borrowing `this` at the same time.

### DIFF
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -2,8 +2,10 @@
 //! standard file descriptors (stdin/stdout/stderr).
 
 use std::any::Any;
+use std::cell::{Ref, RefCell, RefMut};
 use std::collections::BTreeMap;
 use std::io::{self, ErrorKind, IsTerminal, Read, SeekFrom, Write};
+use std::rc::Rc;
 
 use rustc_middle::ty::TyCtxt;
 use rustc_target::abi::Size;
@@ -12,7 +14,7 @@ use crate::shims::unix::*;
 use crate::*;
 
 /// Represents an open file descriptor.
-pub trait FileDescriptor: std::fmt::Debug + Any {
+pub trait FileDescription: std::fmt::Debug + Any {
     fn name(&self) -> &'static str;
 
     fn read<'tcx>(
@@ -44,12 +46,9 @@ pub trait FileDescriptor: std::fmt::Debug + Any {
     fn close<'tcx>(
         self: Box<Self>,
         _communicate_allowed: bool,
-    ) -> InterpResult<'tcx, io::Result<i32>> {
+    ) -> InterpResult<'tcx, io::Result<()>> {
         throw_unsup_format!("cannot close {}", self.name());
     }
-
-    /// Return a new file descriptor *that refers to the same underlying object*.
-    fn dup(&mut self) -> io::Result<Box<dyn FileDescriptor>>;
 
     fn is_tty(&self, _communicate_allowed: bool) -> bool {
         // Most FDs are not tty's and the consequence of a wrong `false` are minor,
@@ -58,7 +57,7 @@ pub trait FileDescriptor: std::fmt::Debug + Any {
     }
 }
 
-impl dyn FileDescriptor {
+impl dyn FileDescription {
     #[inline(always)]
     pub fn downcast_ref<T: Any>(&self) -> Option<&T> {
         (self as &dyn Any).downcast_ref()
@@ -70,7 +69,7 @@ impl dyn FileDescriptor {
     }
 }
 
-impl FileDescriptor for io::Stdin {
+impl FileDescription for io::Stdin {
     fn name(&self) -> &'static str {
         "stdin"
     }
@@ -88,16 +87,12 @@ impl FileDescriptor for io::Stdin {
         Ok(Read::read(self, bytes))
     }
 
-    fn dup(&mut self) -> io::Result<Box<dyn FileDescriptor>> {
-        Ok(Box::new(io::stdin()))
-    }
-
     fn is_tty(&self, communicate_allowed: bool) -> bool {
         communicate_allowed && self.is_terminal()
     }
 }
 
-impl FileDescriptor for io::Stdout {
+impl FileDescription for io::Stdout {
     fn name(&self) -> &'static str {
         "stdout"
     }
@@ -120,16 +115,12 @@ impl FileDescriptor for io::Stdout {
         Ok(result)
     }
 
-    fn dup(&mut self) -> io::Result<Box<dyn FileDescriptor>> {
-        Ok(Box::new(io::stdout()))
-    }
-
     fn is_tty(&self, communicate_allowed: bool) -> bool {
         communicate_allowed && self.is_terminal()
     }
 }
 
-impl FileDescriptor for io::Stderr {
+impl FileDescription for io::Stderr {
     fn name(&self) -> &'static str {
         "stderr"
     }
@@ -145,10 +136,6 @@ impl FileDescriptor for io::Stderr {
         Ok(Write::write(&mut { self }, bytes))
     }
 
-    fn dup(&mut self) -> io::Result<Box<dyn FileDescriptor>> {
-        Ok(Box::new(io::stderr()))
-    }
-
     fn is_tty(&self, communicate_allowed: bool) -> bool {
         communicate_allowed && self.is_terminal()
     }
@@ -158,7 +145,7 @@ impl FileDescriptor for io::Stderr {
 #[derive(Debug)]
 pub struct NullOutput;
 
-impl FileDescriptor for NullOutput {
+impl FileDescription for NullOutput {
     fn name(&self) -> &'static str {
         "stderr and stdout"
     }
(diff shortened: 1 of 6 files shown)


RalfJung:
Thanks for the PR! That was quick. :)

> define struct FileDescription(Rc<RefCell<dyn FileDescriptor>>)

This sounds the wrong way around. Many file descriptors map to fewer file descriptions, not the other way around.

src/shims/unix/fd.rs
```diff
@@ -41,24 +43,19 @@ pub trait FileDescriptor: std::fmt::Debug + Any {
         throw_unsup_format!("cannot seek on {}", self.name());
     }
 
-    fn close<'tcx>(
-        self: Box<Self>,
-        _communicate_allowed: bool,
-    ) -> InterpResult<'tcx, io::Result<i32>> {
+    /// Note: Please ensure `Rc::strong_count(&fd) == 1` before calling this fuction.
```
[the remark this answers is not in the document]
oli-obk: ah, but that's what you meant 🤦 yea, we should only call this once

src/shims/unix/fd.rs
```diff
@@ -329,7 +313,8 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         let fd = this.read_scalar(fd_op)?.to_i32()?;
 
         Ok(Scalar::from_i32(if let Some(file_descriptor) = this.machine.fds.remove(fd) {
-            let result = file_descriptor.close(this.machine.communicate())?;
+            let result =
+                file_descriptor.0.as_ref().borrow_mut().close(this.machine.communicate())?;
```
oli-obk, line 317: so.. we'll want to check the refcount here first, and only call close if it's 1

Luv-Ray: refactored the close check to another function, thanks for the review :)

src/shims/unix/fd.rs
```diff
@@ -41,24 +43,19 @@ pub trait FileDescriptor: std::fmt::Debug + Any {
         throw_unsup_format!("cannot seek on {}", self.name());
     }
 
-    fn close<'tcx>(
-        self: Box<Self>,
-        _communicate_allowed: bool,
-    ) -> InterpResult<'tcx, io::Result<i32>> {
+    /// Note: This function should not be called except [`FileDescriptor::close`].
+    fn close<'tcx>(&mut self, _communicate_allowed: bool) -> InterpResult<'tcx, io::Result<i32>> {
```
RalfJung, line 47: Can you make this function take `self`? Then we can be sure it is called only once.

We can use `Rc::into_inner` to get ownership of `self` for this.

RalfJung (commented):
Nice! I quite like the `let ... else` pattern, it avoids rightward drift of the actual core logic of these operations.

However, I think we can do `close` a bit nicer, without manually looking at the `strong_count`.

src/shims/unix/fd.rs
Luv-Ray: That's correct, using `self` as a parameter for `close` would be more natural. However, it would result in an error:

>  the size for values of type `Self` cannot be known at compilation time

An alternative approach is to take `Box<self>` as parameter and `FileDescriptor` will become `Rc<RefCell<Box<dyn FileDescription>>>`. What do you think about this?

RalfJung: Ah, right, it must be object safe... the double pointer indirection is a bit unfortunate,y but yeah making it `Box<self>` sounds like a good idea to me.

Luv-Ray: done :)

src/shims/unix/fd.rs
```diff
@@ -172,16 +159,28 @@ impl FileDescriptor for NullOutput {
...
+
+impl FileDescriptor {
+    pub fn new<T: FileDescription>(fd: T) -> Self {
+        FileDescriptor(Rc::new(RefCell::new(Box::new(fd))))
+    }
 
-    fn dup(&mut self) -> io::Result<Box<dyn FileDescriptor>> {
-        Ok(Box::new(NullOutput))
+    pub fn close<'ctx>(self, communicate_allowed: bool) -> InterpResult<'ctx, io::Result<i32>> {
+        match Rc::into_inner(self.0) {
```
RalfJung, line 173:
```suggestion
        // Destroy this `Rc` using `into_inner` so we can call `close` instead of
        // implicitly running the destructor of the file description.
        match Rc::into_inner(self.0) {
```

src/shims/unix/fd.rs
```diff
@@ -43,8 +43,10 @@ pub trait FileDescription: std::fmt::Debug + Any {
         throw_unsup_format!("cannot seek on {}", self.name());
     }
 
-    /// Note: This function should not be called except [`FileDescriptor::close`].
-    fn close<'tcx>(&mut self, _communicate_allowed: bool) -> InterpResult<'tcx, io::Result<i32>> {
+    fn close<'tcx>(
+        self: Box<Self>,
+        _communicate_allowed: bool,
+    ) -> InterpResult<'tcx, io::Result<i32>> {
```
RalfJung, line 49: Do you know why `close` returns an `i32`? Shouldn't this be `InterpResult<'tcx, io::Result<()>>`?

src/shims/unix/fd.rs
```diff
@@ -233,12 +234,12 @@ impl FdTable {
         new_fd
     }
 
-    pub fn get(&self, fd: i32) -> Option<Ref<'_, dyn FileDescription>> {
```
RalfJung: You can probably keep the `Option<Ref<'_, dyn FileDescription>>` return type by using `Ref::map`.

Luv-Ray: learned a lot :)

src/shims/unix/fd.rs
Luv-Ray: As `EvalContextExt::close` returns `InterpResult<'tcx, Scalar<Provenance>>`, we can return `InterpResult<'tcx, io::Result<()>>` in `FileDescription::close` and then map the `io::Result<()>` to `io::Result<i32>`

src/shims/unix/fs.rs
```diff
@@ -17,15 +17,17 @@ use crate::shims::unix::*;
...
 struct FileHandle {
     file: File,
     writable: bool,
 }
 
-impl FileDescriptor for FileHandle {
+impl FileDescription for FileHandle {
     fn name(&self) -> &'static str {
-        "FILE"
+        "file"
```
Luv-Ray, line 30: All other `FileDescription`'s name are lower case, so I changed it to keep the same.

src/shims/unix/fd.rs
```diff
@@ -330,6 +325,8 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
 
         Ok(Scalar::from_i32(if let Some(file_descriptor) = this.machine.fds.remove(fd) {
             let result = file_descriptor.close(this.machine.communicate())?;
+            // return `0` if close is successful
+            let result = result.map(|_| 0i32);
```
Luv-Ray, line 329: let `FileDescription::close` return `InterpResult<'tcx, io::Result<()>>` and map the `()` to `0i32` here

src/shims/unix/fd.rs
```diff
@@ -169,10 +169,12 @@ impl FileDescriptor {
         FileDescriptor(Rc::new(RefCell::new(Box::new(fd))))
     }
 
-    pub fn close<'ctx>(self, communicate_allowed: bool) -> InterpResult<'ctx, io::Result<i32>> {
+    // Destroy this `Rc` using `into_inner` so we can call `close` instead of
+    // implicitly running the destructor of the file description.
```
RalfJung, line 173: The comment was meant to go inside the function, not outside it.

Github has a handy button to apply these suggestions so you don;t have to copy-paste them. :)

src/shims/unix/fd.rs
RalfJung, line 329:
```suggestion
            let result = result.map(|()| 0i32);
```
This makes sure we notice if the argument type becomes something else that we do not want to throw away

src/shims/unix/fd.rs
```diff
@@ -234,14 +236,14 @@ impl FdTable {
         new_fd
     }
 
-    pub fn get(&self, fd: i32) -> Option<Ref<'_, Box<dyn FileDescription>>> {
+    pub fn get(&self, fd: i32) -> Option<Ref<'_, dyn FileDescription>> {
         let fd = self.fds.get(&fd)?;
-        Some(fd.0.as_ref().borrow())
+        Some(Ref::map(fd.0.as_ref().borrow(), |fd| fd.as_ref()))
```
RalfJung, line 241: The first `as_ref` should be unnecessary, `Rc` auto-derefs

RalfJung:
Awesome, I think we're done. :)

Please squash the commits together.

RalfJung:
Thanks a lot!

@bors r+

RalfJung:
@Luv-Ray a possible next step, if you want, would be to give `read` and `write` access to `ecx: &mut InterpCx<'mir, 'tcx, MiriMachine<'mir, 'tcx>>`. This will be needed in the future to support blocking on these operations.

To make that work out in terms of borrowing, I think what we will have to do is clone the `Rc` in the `read`/`write` functions in `unix/fd.rs`, so that we have a fully owned `FileDescriptor` next to a uniquely borrowed `InterpCx`.

Luv-Ray:
Thanks for review! I think I can give it a try on the next step.

PR #3557, opened 2024-05-04
Move sched_getaffinity back to Linux

RalfJung:
This reverts commit c1a3f8576ea12b0bed68ad3dedf4069ca3d9816f. The shim isn't actually useful for anything, and it is untested on FreeBSD. On Linux it exists solely because std and num_cpus are trying this before falling back to `sysconf`, but on FreeBSD that's not how they work, so there's no reason I can see to have this stub shim on FreeBSD.

### DIFF
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -724,25 +724,6 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 }
             }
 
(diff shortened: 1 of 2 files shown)


RalfJung:
@devnexen do you have a use-case where adding this function (always returning EINVAL) actually helps? If so, we should add a test, but otherwise I'd prefer to revert.

devnexen:
might be wiser to revert we can always add back if context changes.

RalfJung:
Okay, thanks for clarifying!
@bors r+

PR #3560, opened 2024-05-04
sync: better error in invalid synchronization primitive ID

RalfJung:
@devnexen this should fix the ICE in your PR (but it won't fix the code, it will just report proper UB instead).


src/shims/unix/sync.rs
```diff
@@ -719,6 +727,14 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
     ) -> InterpResult<'tcx, Scalar<Provenance>> {
         let this = self.eval_context_mut();
 
+        // Does not exist on macOS!
+        if !matches!(&*this.tcx.sess.target.os, "linux") {
```
RalfJung, line 731: @devnexen when this lands you'll have to rebase and add the solarish OSes here

Issue #3566, opened 2024-05-05
Solaris: add stub shims so that pre-main code works


RalfJung:
Currently, even an empty `main` fails on Solaris as we're hitting some unsupported functions in the pre-main `std` initialization code -- specifically, the guard page setup.

On other OSes we have some shims that don't even try to implement their proper functionality ([like this](https://github.com/rust-lang/miri/blob/100c718ee7bf479df7763f78b4c268a05a109333/src/shims/unix/linux/foreign_items.rs#L205)), solely for the purpose of making the guard page work. We likely need something similar here.

We will probably have to extend this hack to also allow solaris, not just macos:

https://github.com/rust-lang/miri/blob/10e8bb87fe416bc190426789032594aa1aa311f2/src/shims/unix/mem.rs#L47-L49

We also need a `stack_getbounds` shim. It looks like `stack_getbounds` can just do nothing, but you're going to have to try this out. If not, `this.machine.stack_addr` and `this.machine.stack_size` are the fake values we are using elsewhere.

Cc @devnexen

PR #3570, opened 2024-05-05
Solaris: make pre-main code work

devnexen:
Fixes https://github.com/rust-lang/miri/issues/3566

### DIFF
--- a/src/shims/unix/mem.rs
+++ b/src/shims/unix/mem.rs
@@ -42,9 +42,12 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         let map_shared = this.eval_libc_i32("MAP_SHARED");
         let map_fixed = this.eval_libc_i32("MAP_FIXED");
 
-        // This is a horrible hack, but on MacOS the guard page mechanism uses mmap
+        // This is a horrible hack, but on MacOS and Solaris the guard page mechanism uses mmap
         // in a way we do not support. We just give it the return value it expects.
-        if this.frame_in_std() && this.tcx.sess.target.os == "macos" && (flags & map_fixed) != 0 {
+        if this.frame_in_std()
+            && matches!(&*this.tcx.sess.target.os, "macos" | "solaris")
+            && (flags & map_fixed) != 0
+        {
             return Ok(Scalar::from_maybe_pointer(Pointer::from_addr_invalid(addr), this));
         }
 
--- a/src/shims/unix/solarish/foreign_items.rs
+++ b/src/shims/unix/solarish/foreign_items.rs
@@ -2,7 +2,6 @@ use rustc_span::Symbol;
 use rustc_target::spec::abi::Abi;
 
 use crate::*;
-use shims::EmulateItemResult;
 
 pub fn is_dyn_sym(_name: &str) -> bool {
     false
@@ -24,6 +23,24 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 let [] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
                 let errno_place = this.last_error_place()?;
                 this.write_scalar(errno_place.to_ref(this).to_scalar(), dest)?;
+            }
+
+            "stack_getbounds" => {
+                let [stack] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let stack = this.deref_pointer_as(stack, this.libc_ty_layout("stack_t"))?;
+
+                this.write_int_fields_named(
+                    &[
+                        ("ss_sp", this.machine.stack_addr.into()),
+                        ("ss_size", this.machine.stack_size.into()),
+                        // field set to 0 means not in an alternate signal stack
+                        // https://docs.oracle.com/cd/E86824_01/html/E54766/stack-getbounds-3c.html
+                        ("ss_flags", 0),
+                    ],
+                    &stack,
+                )?;
+
+                this.write_null(dest)?;
             }
 
             _ => return Ok(EmulateItemResult::NotSupported),
--- a/README.md
+++ b/README.md
@@ -227,7 +227,7 @@ degree documented below):
 - We have unofficial support (not maintained by the Miri team itself) for some further operating systems.
   - `freebsd`: **maintainer wanted**. Supports `std::env` and parts of `std::{thread, fs}`, but not `std::sync`.
   - `android`: **maintainer wanted**. Support very incomplete, but a basic "hello world" works.
-  - `illumos`: maintained by @devnexen. Support very incomplete, but a basic "hello world" works.
+  - `solaris` / `illumos`: maintained by @devnexen. Support very incomplete, but a basic "hello world" works.
   - `wasm`: **maintainer wanted**. Support very incomplete, not even standard output works, but an empty `main` function works.
 - For targets on other operating systems, Miri might fail before even reaching the `main` function.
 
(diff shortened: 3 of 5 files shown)


src/machine.rs
```diff
@@ -986,7 +986,7 @@ impl<'mir, 'tcx> Machine<'mir, 'tcx> for MiriMachine<'mir, 'tcx> {
         dest: &MPlaceTy<'tcx, Provenance>,
         ret: Option<mir::BasicBlock>,
         unwind: mir::UnwindAction,
-    ) -> InterpResult<'tcx, Option<ty::Instance<'tcx>>> {
+    ) -> InterpResult<'tcx> {
```
RalfJung, line 989: Not sure what this is about, but this can't be right. Are you maybe using an outdated local toolchain? I recommend doing `touch .auto-toolchain` in your Miri checkout so that all `./miri` command will automatically install the correct toolchain when it is outdated.

src/shims/unix/mem.rs
```diff
@@ -44,7 +44,9 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
 
         // This is a horrible hack, but on MacOS the guard page mechanism uses mmap
         // in a way we do not support. We just give it the return value it expects.
-        if this.frame_in_std() && this.tcx.sess.target.os == "macos" && (flags & map_fixed) != 0 {
+        if this.frame_in_std() && this.tcx.sess.target.os == "macos"
+            || this.tcx.sess.target.os == "solaris" && (flags & map_fixed) != 0
```
RalfJung, line 48: This will need parentheses around the `... || ...`

src/machine.rs
devnexen: I see was not sure, why I kept in separate commit. Will fix later.

RalfJung: If you are having trouble running Miri on your own system, feel free to ask [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/269128-miri); there's people there that can help. You shouldn't have to always wait for CI to see if anything worked. :)

You don't even need a Solaris machine; `./miri run -v --target x86_64-pc-solaris tests/pass/vec.rs` should do it. (This is assuming you already did `touch .auto-toolchain`.)

RalfJung:
You lost the changes in `ci.sh` so currently this won't be tested.

RalfJung:
Also your commits are still coming from the future. ;)

image

This seems to confuse github as well (when I posted a comment, often it doesn't appear until I reload the page), so it would be good if you could fix your system clock.

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -19,6 +18,11 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
     ) -> InterpResult<'tcx, EmulateItemResult> {
         let this = self.eval_context_mut();
         match link_name.as_str() {
+            "stack_getbounds" => {
```
RalfJung, line 21: Please match the structure [used here](https://github.com/rust-lang/miri/blob/100c718ee7bf479df7763f78b4c268a05a109333/src/shims/unix/linux/foreign_items.rs#L205): use the same comment, put the shim at the end of the list, and guard it with `if this.frame_in_std()`

src/shims/unix/mem.rs
```diff
@@ -44,8 +44,9 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
 
         // This is a horrible hack, but on MacOS the guard page mechanism uses mmap
         // in a way we do not support. We just give it the return value it expects.
-        if this.frame_in_std() && this.tcx.sess.target.os == "macos"
-            || this.tcx.sess.target.os == "solaris" && (flags & map_fixed) != 0
+        if this.frame_in_std()
+            && (this.tcx.sess.target.os == "macos" || this.tcx.sess.target.os == "solaris")
```
RalfJung, line 48:
```suggestion
            && matches!(&*this.tcx.sess.target.os, "macos" | "solaris")
```

src/shims/unix/mem.rs
```diff
@@ -44,8 +44,9 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
 
         // This is a horrible hack, but on MacOS the guard page mechanism uses mmap
```
RalfJung, line 45:
```suggestion
        // This is a horrible hack, but on MacOS and Solaris the guard page mechanism uses mmap
```

RalfJung:
Okay looks like just leaving it at 0 doesn't work.^^ It triggers an underflow [here](https://github.com/rust-lang/rust/blob/5b9d7ab558fe6ee53bd40e16a20c25716f3c9e56/library/std/src/sys/pal/unix/stack_overflow.rs#L432).

So stack_getbounds will have to actually store the stack address and size in the appropriate fields of the struct. You can use [this shim](https://github.com/rust-lang/miri/blob/c3ee0077a3e43b01eaa02985b4946b74be9b8bc0/src/shims/windows/foreign_items.rs#L293-L313) as a template.

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -28,7 +29,22 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
             // Incomplete shims that we "stub out" just to get pre-main initialization code to work.
             // These shims are enabled only when the caller is in the standard library.
             "stack_getbounds" if this.frame_in_std() => {
-                let [_stack] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let [stack] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let stack = this.deref_pointer_as(stack, this.libc_ty_layout("stack_t"))?;
+                // Not sure we can figure out ss_flags field ?
```
RalfJung, line 34: Is there documentation somewhere on what that field means?

devnexen: [yes](https://docs.oracle.com/cd/E36784_01/html/E36874/stack-getbounds-3c.html)

RalfJung: Ah, [found it](https://docs.oracle.com/cd/E36784_01/html/E36874/stack-getbounds-3c.html)

> The ss_flags member is set to SS_ONSTACK if the calling thread is executing on its alternate signal stack, and zero otherwise.

So, you can just set it to 0, since we don't support signals. (But add a comment explaining why we set it to 0.)

Then you can also remove the `write_bytes_ptr` as we are then setting all fields.

And you can remove the `if this.frame_in_std()` as this is actually doing its job then. :D  (okay the stack address is nonsense but so far that hasn't caused issues)

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -26,6 +25,24 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 this.write_scalar(errno_place.to_ref(this).to_scalar(), dest)?;
             }
 
+            // Incomplete shims that we "stub out" just to get pre-main initialization code to work.
+            // These shims are enabled only when the caller is in the standard library.
+            "stack_getbounds" => {
```
RalfJung, line 30:
```suggestion
            }
            "stack_getbounds" => {
```

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -26,6 +25,24 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                let [stack] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let stack = this.deref_pointer_as(stack, this.libc_ty_layout("stack_t"))?;
+
+                this.write_int_fields_named(
+                    &[
+                        ("ss_sp", this.machine.stack_addr.into()),
+                        ("ss_size", this.machine.stack_size.into()),
+                        ("ss_flags", 0),
```
RalfJung, line 38:
```suggestion
                        // 0 means we are not in the alternate signal stack
                        ("ss_flags", 0),
```
As I said, there should be a comment.

RalfJung:
Looks good, thanks!

Please squash the commits.

RalfJung:
Awesome. :)
@bors r+

Issue #3567, opened 2024-05-05
Illumos: make basic std facilities work


RalfJung:
We're currently lacking some basics on Illumos:
- [ ] high-alignment heap allocations (`MIRI_TEST_TARGET=x86_64-unknown-illumos ./miri test alloc`): needs support for `libc::memalign`.
- [ ] HashMap random initialization  (`MIRI_TEST_TARGET=x86_64-unknown-illumos ./miri test hashmap`): this complains about `open64` in isolation; the issue is that std falls back to file-based randomness, but we want it to use one of the native randomness APIs. Apparently this will need a patch for the standard library; currently it has no support for [`getrandom` on Solarish](https://github.com/rust-lang/rust/blob/d9c0eb8084deb65de69b23a7714b66efbff7eb03/library/std/src/sys/pal/unix/rand.rs#L92-L101). Solarish has `libc::getrandom` so it should be trivial to extend [this](https://github.com/rust-lang/rust/blob/d9c0eb8084deb65de69b23a7714b66efbff7eb03/library/std/src/sys/pal/unix/rand.rs#L62-L65) to also cover these OSes.

Cc @devnexen

This issue has been assigned to @devnexen via [this comment](https://github.com/rust-lang/miri/issues/3567#issuecomment-2095785240).

RalfJung:
@rustbot assign @devnexen
you seem to be working on this?

PR #3575, opened 2024-05-05
Illumos part3

devnexen:
Fixes https://github.com/rust-lang/miri/issues/3567

### DIFF
--- a/ci/ci.sh
+++ b/ci/ci.sh
@@ -148,8 +148,8 @@ case $HOST_TARGET in
     BASIC="$VERY_BASIC hello hashmap alloc align" # ensures we have the shims for stdout and basic data structures
     TEST_TARGET=x86_64-unknown-freebsd run_tests_minimal $BASIC panic/panic concurrency/simple atomic threadname libc-mem libc-misc libc-random libc-time fs env num_cpus
     TEST_TARGET=i686-unknown-freebsd   run_tests_minimal $BASIC panic/panic concurrency/simple atomic threadname libc-mem libc-misc libc-random libc-time fs env num_cpus
-    TEST_TARGET=x86_64-unknown-illumos run_tests_minimal $VERY_BASIC hello panic/panic concurrency/simple pthread-sync libc-mem libc-misc libc-random
-    TEST_TARGET=x86_64-pc-solaris run_tests_minimal $VERY_BASIC hello panic/panic concurrency/simple pthread-sync libc-mem libc-misc libc-random
+    TEST_TARGET=x86_64-unknown-illumos run_tests_minimal $BASIC hello panic/panic concurrency/simple pthread-sync libc-mem libc-misc libc-random env
+    TEST_TARGET=x86_64-pc-solaris run_tests_minimal $BASIC hello panic/panic concurrency/simple pthread-sync libc-mem libc-misc libc-random env
     TEST_TARGET=aarch64-linux-android  run_tests_minimal $VERY_BASIC hello panic/panic
     TEST_TARGET=wasm32-wasi run_tests_minimal $VERY_BASIC wasm
     TEST_TARGET=wasm32-unknown-unknown run_tests_minimal $VERY_BASIC wasm
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -714,8 +714,9 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 this.write_int(super::UID, dest)?;
             }
 
-            "getpwuid_r"
+            "getpwuid_r" | "__posix_getpwuid_r"
             if this.frame_in_std() => {
+                // getpwuid_r is the standard name, __posix_getpwuid_r is used on solarish
                 let [uid, pwd, buf, buflen, result] =
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
                 this.check_no_isolation("`getpwuid_r`")?;
--- a/tests/fail/environ-gets-deallocated.rs
+++ b/tests/fail/environ-gets-deallocated.rs
@@ -1,6 +1,11 @@
 //@ignore-target-windows: Windows does not have a global environ list that the program can access directly
 
-#[cfg(any(target_os = "linux", target_os = "freebsd"))]
+#[cfg(any(
+    target_os = "linux",
+    target_os = "freebsd",
+    target_os = "solaris",
+    target_os = "illumos"
+))]
 fn get_environ() -> *const *const u8 {
     extern "C" {
         static mut environ: *const *const u8;
--- a/src/shims/extern_static.rs
+++ b/src/shims/extern_static.rs
@@ -76,6 +76,10 @@ impl<'mir, 'tcx> MiriMachine<'mir, 'tcx> {
                 Self::null_ptr_extern_statics(this, &["bsd_signal"])?;
                 Self::weak_symbol_extern_statics(this, &["signal"])?;
             }
+            "solaris" | "illumos" => {
+                let environ = this.machine.env_vars.unix().environ();
+                Self::add_extern_static(this, "environ", environ);
+            }
             "windows" => {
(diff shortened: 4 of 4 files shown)


src/shims/unix/solarish/foreign_items.rs
```diff
@@ -18,6 +19,28 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+            // Allocation
+            "memalign" => {
+                let [align, size] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let align = this.read_target_usize(align)?;
+                let size = this.read_target_usize(size)?;
+
+                // memalign requires the alignment to be a power of 2
+                // and to be, at least, of word size.
+                // http://docs.oracle.com/cd/E88353_01/html/E37743/memalign-3c.html
```
RalfJung, line 31: This URL shows an error 404 for me.

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -18,6 +19,28 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                if !align.is_power_of_two() || align < this.pointer_size().bytes() || size == 0 {
```
RalfJung, line 32:
```suggestion
                // The docs also guarantee a null pointer is returned when the size is 0.
                if !align.is_power_of_two() || align < this.pointer_size().bytes() || size == 0 {
```

src/shims/unix/solarish/foreign_items.rs
RalfJung: I found it at https://docs.oracle.com/cd/E88353_01/html/E37843/memalign-3c.html.

It doesn't say what happens when the alignment is violated though -- is it UB or a safe error?

RalfJung: Ah no it does say something:

> The memalign() function may also fail if:
>
> EINVAL
>
>    The value of the alignment parameter is not a power of two multiple of sizeof(void *), or the size parameter is zero, or the combination of the two would lead to an integer overflow.

But how is the error returned...?
This is where it'd probably be good to experiment with an actual Solaris/Illumos system and see what happens on a bad alignment.^^

devnexen: I just tried on my illumos instance, it returns null and emits EINVAL.

RalfJung:
That CI failure is actually a legit bug in std, I think: https://github.com/rust-lang/rust/issues/124787.

src/shims/unix/solarish/foreign_items.rs
RalfJung: What does "emit" mean"?

In [this document](https://open-std.org/JTC1/SC22/WG14/www/docs/summary.htm#dr_460) and in `man memalign` on my system it says that `memalign` returns NULL and sets errno on error.

devnexen:
> That CI failure is actually a legit bug in std, I think: [rust-lang/rust#124787](https://github.com/rust-lang/rust/issues/124787).

alright .. same will wait the getrandom part being merged to pick up this PR.

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -18,6 +19,30 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let align = this.read_target_usize(align)?;
+                let size = this.read_target_usize(size)?;
+
+                // memalign requires the size to be greater than 0, the alignment to be a power of 2
+                // to be, at least, of word size in which EINVAL is emitted
+                // and a null pointer is returned.
+                // https://docs.oracle.com/cd/E88353_01/html/E37843/memalign-3c.html
+                if !align.is_power_of_two() || align < this.pointer_size().bytes() || size == 0 {
+                    this.set_last_error(this.eval_libc("EINVAL"))?;
```
RalfJung, line 34: I don't think size 0 should lead to EINVAL.

RalfJung: Ah, it is listed in the error conditions below: "or the size parameter is zero". But it's not listed above. Wow that's bad documentation.^^

Please point this out in the comments.

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -18,6 +19,30 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
     ) -> InterpResult<'tcx, EmulateItemResult> {
         let this = self.eval_context_mut();
         match link_name.as_str() {
+            // Allocation
+            "memalign" => {
```
RalfJung, line 23: With https://github.com/rust-lang/rust/pull/124798, `memalign` will not be called on Solarish any more.

You can still add it here if you want, but then you should also add a test that calls this function (directly via `libc`).

RalfJung:
CI fails because https://github.com/rust-lang/rust/pull/124766 isn't picked up automatically; I have to do a rustc-pull for that.

But let's wait for https://github.com/rust-lang/rust/pull/124798 to land first, so we can do it all at once.

src/shims/unix/solarish/foreign_items.rs
RalfJung: This comment is not resolved yet: you're adding a new shim that's entirely untested. That's not going to fly.

ci/ci.sh
```diff
@@ -148,8 +148,8 @@ case $HOST_TARGET in
     BASIC="$VERY_BASIC hello hashmap alloc align" # ensures we have the shims for stdout and basic data structures
     TEST_TARGET=x86_64-unknown-freebsd run_tests_minimal $BASIC panic/panic concurrency/simple atomic threadname libc-mem libc-misc libc-random libc-time fs env num_cpus
     TEST_TARGET=i686-unknown-freebsd   run_tests_minimal $BASIC panic/panic concurrency/simple atomic threadname libc-mem libc-misc libc-random libc-time fs env num_cpus
-    TEST_TARGET=x86_64-unknown-illumos run_tests_minimal $VERY_BASIC hello panic/panic concurrency/simple pthread-sync libc-mem libc-misc libc-random
-    TEST_TARGET=x86_64-pc-solaris run_tests_minimal $VERY_BASIC hello panic/panic concurrency/simple pthread-sync libc-mem libc-misc libc-random
+    TEST_TARGET=x86_64-unknown-illumos run_tests_minimal $VERY_BASIC hello panic/panic concurrency/simple pthread-sync libc-mem libc-misc libc-random alloc hashmap
```
RalfJung, line 151:
```suggestion
    TEST_TARGET=x86_64-unknown-illumos run_tests_minimal $BASIC hello panic/panic concurrency/simple pthread-sync libc-mem libc-misc libc-random
```
same for solaris

tests/fail/environ-gets-deallocated.rs
RalfJung, line 1: This test isn't actually run by this PR, is it? same for the `environ` change. Did you mean to add `env` to the CI filters?

src/shims/unix/foreign_items.rs
```diff
@@ -714,7 +714,7 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 this.write_int(super::UID, dest)?;
             }
 
-            "getpwuid_r"
+            "getpwuid_r" | "__posix_getpwuid_r"
             if this.frame_in_std() => {
                 let [uid, pwd, buf, buflen, result] =
```
RalfJung, line 719:
```suggestion
                // getpwuid_r is the standard name, __posix_getpwuid_r is used on Solarish
                let [uid, pwd, buf, buflen, result] =
```

RalfJung:
Looks good, thanks!
@bors r+

Issue #3576, opened 2024-05-06
Make `malloc(0)` and `posix_memalign(0)` return a non-null pointer


RalfJung:
Currently we just return a null pointer, but we're also allowed to return a unique non-zero address. That would have the benefit that we can detect leaks (not calling free) or double-free.

RalfJung:
> This is interesting. I've had this question on my mind for quite a while, how will malloc(0) cause a memory leak if the memory allocated is 0 byte?

There can be metadata associated with this allocation, managed by the allocator, that would be leaked.

> Besides, the return value of malloc(0) seems to be implementation-defined[1](https://github.com/rust-lang/miri/issues/3576#user-content-fn-1-c1e2d588d114622dea076268d4b2e4a5), and may return null or a valid pointer. If miri only return a non-zero address, and the user assumed malloc(0) return null (hence they don't free it), would it be considered a false positive?

Implementation-defined means that the *implementation* gets to decide. If the user assumes that `malloc(0)` returns null, then they are making a false assumption, since it is not necessarily true that the implementation will return null.

RalfJung:
The null pointer has no AllocId, so yeah this is expected.

Why are you changing realloc at all? It doesn't seem to me like it has to change.

tiif:
> Why are you changing realloc at all? It doesn't seem to me like it has to change.

Yea, I don't think anything needs to be change in ``realloc``. I just had a memory leak that I thought it was coming from the ``realloc`` implementation.

tiif:
I wonder if we should do the same for ``posix_memalign``? It is still returning a null pointer currently. 

> If size is 0, then the value placed in *memptr is either NULL or a unique pointer value.
https://man7.org/linux/man-pages/man3/posix_memalign.3.html

RalfJung:
Yeah, posix_memalign shod also return non-null on size 0. (But that can be a separate PR.)

One of the key parts of the PR will be two tests to ensure we detect double free and use after free of malloc(0).

RalfJung:
https://github.com/rust-lang/miri/pull/3580 implements this for malloc; posix_memalign could get the same treatment as discussed above.

PR #3580, opened 2024-05-07
Implement non-null pointer for malloc(0)

tiif:
Use non-null pointer for malloc(0) as mentioned in  #3576 to detect leaks and double free of ``malloc(0)`` addresses.

### DIFF
--- a/src/shims/alloc.rs
+++ b/src/shims/alloc.rs
@@ -36,6 +36,9 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         if kind == MiriMemoryKind::WinHeap || size >= min_align {
             return Align::from_bytes(min_align).unwrap();
         }
+        if size == 0 {
+            return Align::ONE;
+        }
         // We have `size < min_align`. Round `size` *down* to the next power of two and use that.
         fn prev_power_of_two(x: u64) -> u64 {
             let next_pow2 = x.next_power_of_two();
@@ -85,21 +88,17 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         kind: MiriMemoryKind,
     ) -> InterpResult<'tcx, Pointer<Option<Provenance>>> {
         let this = self.eval_context_mut();
-        if size == 0 {
-            Ok(Pointer::null())
-        } else {
-            let align = this.min_align(size, kind);
-            let ptr = this.allocate_ptr(Size::from_bytes(size), align, kind.into())?;
-            if zero_init {
-                // We just allocated this, the access is definitely in-bounds and fits into our address space.
-                this.write_bytes_ptr(
-                    ptr.into(),
-                    iter::repeat(0u8).take(usize::try_from(size).unwrap()),
-                )
-                .unwrap();
-            }
-            Ok(ptr.into())
+        let align = this.min_align(size, kind);
+        let ptr = this.allocate_ptr(Size::from_bytes(size), align, kind.into())?;
+        if zero_init {
+            // We just allocated this, the access is definitely in-bounds and fits into our address space.
+            this.write_bytes_ptr(
+                ptr.into(),
+                iter::repeat(0u8).take(usize::try_from(size).unwrap()),
+            )
+            .unwrap();
         }
+        Ok(ptr.into())
     }
 
     fn free(
--- a/tests/pass-dep/libc/libc-mem.rs
+++ b/tests/pass-dep/libc/libc-mem.rs
@@ -110,9 +110,10 @@ fn test_malloc() {
     }
 
     unsafe {
-        // Realloc with size 0 is okay for the null pointer
+        // Realloc with size 0 is okay for the null pointer (and acts like `malloc(0)`)
         let p2 = libc::realloc(ptr::null_mut(), 0);
-        assert!(p2.is_null());
+        assert!(!p2.is_null());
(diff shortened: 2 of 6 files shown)


RalfJung:
Thanks!

As just mentioned in the issue, this needs two new "fail-dep" tests: a double free of malloc(0), and a use after free.

src/shims/alloc.rs
```diff
@@ -85,21 +85,18 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
-                    ptr.into(),
-                    iter::repeat(0u8).take(usize::try_from(size).unwrap()),
-                )
-                .unwrap();
-            }
-            Ok(ptr.into())
+        let align = this.min_align(size, kind);
+        let ptr = this.allocate_ptr(Size::from_bytes(size), align, kind.into())?;
+        // Do not write to the pointer if the size of malloc is 0.
+        if zero_init && size != 0 {
```
RalfJung, line 91: The size check here should not be needed. Writing zero bytes to a zero-sized allocation should be fine.

src/shims/alloc.rs
```diff
@@ -85,21 +85,18 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
-            if zero_init {
-                // We just allocated this, the access is definitely in-bounds and fits into our address space.
-                this.write_bytes_ptr(
-                    ptr.into(),
-                    iter::repeat(0u8).take(usize::try_from(size).unwrap()),
-                )
-                .unwrap();
-            }
-            Ok(ptr.into())
+        let align = this.min_align(size, kind);
```
RalfJung, line 88: Can you update `fn prev_power_of_two` inside `min_align` so that if the input is 0, it returns 1? It seems to work currently despite this because `Align::from_bytes(0)` returns `Align::ONE` but that's not a nice thing to rely on IMO.

tests/pass-dep/libc/libc-mem.rs
```diff
@@ -112,7 +112,8 @@ fn test_malloc() {
     unsafe {
         // Realloc with size 0 is okay for the null pointer
```
RalfJung, line 113:
```suggestion
        // Realloc with size 0 is okay for the null pointer (and acts like `malloc(0)`)
```

tiif:
I am trying to create a use-after-free test, but the test below surprisingly not throwing any error? 
```rust
    unsafe {
        let ptr = libc::malloc(0);
        libc::free(ptr);
        let _ = *ptr;
    }
```

RalfJung:
Yes, that's how `let _` behaves. It creates the place but doesn't read from it.

Use `let _val = *ptr;`.

tiif:
I see, right now this error shows up, how is pointer dereference related to Copy?
```
error[E0507]: cannot move out of `*ptr` which is behind a raw pointer
 --> ./tests/fail-dep/libc/malloc_zero_use_after_free.rs:5:20
  |
5 |         let _val = *ptr;
  |                    ^^^^ move occurs because `*ptr` has type `libc::c_void`, which does not implement the `Copy` trait
  |
help: consider removing the dereference here
  |
5 -         let _val = *ptr;
5 +         let _val = ptr;
  |

For more information about this error, try `rustc --explain E0507`.

```
full test:
```rust
fn main() {
    unsafe {
        let ptr = libc::malloc(0);
        libc::free(ptr);
        let _val = *ptr;
    }
}

```

RalfJung:
c_void is not Copy so you cannot just do a deref.

But anyway this will not give the desired error as ZST accesses are NOPs. I should not have said use after free, sorry. I meant memory leak and double free.

tiif:
> c_void is not Copy so you cannot just do a deref. But anyway this will not give the desired error as ZST accesses are NOPs. I should not have said use after free, sorry. I meant memory leak and double free.

Ah I see, thanks!

src/shims/alloc.rs
```diff
@@ -38,6 +38,9 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         }
         // We have `size < min_align`. Round `size` *down* to the next power of two and use that.
         fn prev_power_of_two(x: u64) -> u64 {
+            if x == 0 {
+                return 1;
+            }
```
tiif, line 43: I wonder if it is better to do it outside ``prev_power_of_two`` as 0 is not the previous power of two.

src/shims/alloc.rs
```diff
@@ -47,6 +47,9 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 next_pow2 / 2
             }
         }
+        if size == 0 {
+            return Align::from_bytes(1).unwrap();
+        }
```
RalfJung, line 52: This works, but please move it up above the comment that says we will round down.

Also, you can use `Align::ONE` to avoid an unwrap here.

RalfJung:
Awesome. :) Please squash the commits.

RalfJung:
I just realized this also affects the Windows HeapAlloc function. @ChrisDenton I didn't find anything in the docs about HeapAlloc with size 0... is it okay for that to return a non-null pointer that must be freed later?

ChrisDenton:
Sure that's fine. There's no documented restriction on allocating zero bytes and `HeapAlloc` in practice does in fact return a non-null pointer if you ask for 0 bytes. Though tbh it's not something I'd be in the habit of doing.

Issue #2182, opened 2022-06-03
Integers and pointers sometimes incorrectly preserve provenance on typed copies


RalfJung:
Miri currently does not properly implement what happens to provenance during a typed copy:
- When doing a copy at integer type, provenance should be stripped. Instead, we will currently complain during validation that provenance on an integer is UB.
- When doing a copy at pointer type, if the provenance is not the same for all bytes of a pointer, it should be stripped. Instead, we will currently just preserve it.

The best way I see to solve this is to make validation mutating, so that during validation we can adjust integer and pointer values following these rules. Mutating validation is also needed for https://github.com/rust-lang/miri/issues/845.

a-b-c-1-2-3:
Seems like testcase #1 currently passes but I didn't dig too deep to see why/since when; maybe this issue is worth updating for anyone coming to work on it in the future?

RalfJung:
Ah right that got fixed by https://github.com/rust-lang/rust/pull/104054.

Issue #845, opened 2019-07-17
Miri preserves padding and partial initialization on copies


RalfJung:
Miri should be able to detect that the following is UB because it prints uninitialized memory:
```rust
use std::mem;

#[repr(C)]
struct Pair(u8, u16);

fn main() { unsafe {
    let p: Pair = mem::transmute(0u32); // The copy when `Pair` is returned from `transmute` should destroy padding.
    let c = &p as *const _ as *const u8;
    println!("{}", *c.add(1)); // Print the padding byte.
} }
```
However, currently assignment is just implemented as an untyped `memcpy`, so we incorrectly preserve padding.

RalfJung:
A related issue is that when copying, e.g., `[usize; 3]`, Miri will *not* perform a "typed copy": a typed copy would make the entire array element "undefined" if any of its bytes is undefined, whereas Miri will just copy the underlying bytes. Also see [this stackoverflow question](https://stackoverflow.com/questions/61114026/does-stdptrwrite-transfer-the-uninitialized-ness-of-the-bytes-it-writes).

sporksmith:
Another example where Miri may be missing UB. I was attempting to use Miri to remind myself of how to safely work with libc style socket unions, and mistakenly (?) convinced myself that that this is sound:

```
#[repr(C)]
#[derive(Debug, Copy, Clone)]
struct Foo {
    val16: u16,
    // Padding bytes go here!
    val32: u32,
}

#[repr(C)]
#[derive(Debug, Copy, Clone)]
struct Bar {
    bytes: [u8; 8],
}

union FooBar {
    foo: Foo,
    bar: Bar,
}

pub fn main() {
    // Initialize as u8 to ensure padding bytes are zeroed.
    let mut foobar = FooBar { bar: Bar { bytes: [0u8; 8] } };
    // Reading either field is ok.
    println!("{:?} {:?}", unsafe { foobar.foo }, unsafe { foobar.bar });
    // Does this assignment copy the uninitialized padding bytes
    // over the initialized padding bytes? miri doesn't seem to think so.
    foobar.foo = Foo { val16: 1, val32: 2 };
    // Still OK to read.
    println!("{:?} {:?}", unsafe { foobar.foo }, unsafe { foobar.bar });
}
```

IIUC the line `foobar.foo = Foo { val16: 1, val32: 2 };` is allowed to overwrite the padding bytes in `foobar.foo`, but no UB is detected when those bytes are read on the following line.

[...]

PR #3583, opened 2024-05-07
README: update introduction

RalfJung:
For some time now, we have (to my knowledge) been able to detect all the UB that rustc actually exploits. I think it's time to advertise that.

We also haven't had the problem of "not every nightly has Miri" for a while.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,31 +1,26 @@
 # Miri
 
-An experimental interpreter for [Rust][rust]'s
-[mid-level intermediate representation][mir] (MIR). It can run binaries and
-test suites of cargo projects and detect certain classes of
-[undefined behavior](https://doc.rust-lang.org/reference/behavior-considered-undefined.html),
-for example:
+Miri is an [Undefined Behavior][reference-ub] detection tool for Rust. It can run binaries and test
+suites of cargo projects and detect unsafe code that fails to uphold its safety requirements. For
+instance:
 
 * Out-of-bounds memory accesses and use-after-free
 * Invalid use of uninitialized data
 * Violation of intrinsic preconditions (an [`unreachable_unchecked`] being
   reached, calling [`copy_nonoverlapping`] with overlapping ranges, ...)
 * Not sufficiently aligned memory accesses and references
-* Violation of *some* basic type invariants (a `bool` that is not 0 or 1, for example,
+* Violation of basic type invariants (a `bool` that is not 0 or 1, for example,
   or an invalid enum discriminant)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,33 +40,36 @@ clocks, are replaced by deterministic "fake" implementations. Set
 (In particular, the "fake" system RNG APIs make Miri **not suited for
 cryptographic use**! Do not generate keys using Miri.)
 
-All that said, be aware that Miri will **not catch all cases of undefined
-behavior** in your program, and cannot run all programs:
+All that said, be aware that Miri does **not catch every violation of the Rust
+specification** in your program, not least because there is no such specification.
```
saethlin, line 44: Can you explain how this wording change relates to https://github.com/rust-lang/miri/issues/845 and https://github.com/rust-lang/miri/issues/2182?

RalfJung: Hm, to me these are "just" bugs. I don't expect [this list](https://github.com/rust-lang/miri/labels/I-misses-UB) to generally be empty -- just like Rust has soundness bugs and still claims type safety.

I am not sure where we should put the threshold for prominently pointing out such bugs in our README.

saethlin: I have actually seen someone run into https://github.com/rust-lang/miri/issues/845 and the conversation rapidly devolved into a bunch of very wrong speculation about how Rust works until found the conversation and linked the issue.

I agree that it is very hard to know where to draw the line, but I also think Miri is in a bit of a different position. People expect Miri to be the arbiter of when UB is executed, so when Miri is silent on a program that executes UB it's hard for people to know what to do. When rustc accepts your program then mangles it in codegen because of a soundness hole, you know something is afoot. But Miri never gives any indication that is is behaving buggily in this sense.

I also think it's a bit sketchy that an I-misses-ub has been open for (almost) 5 years. I know I looked over the implementation that would have to be done to fix it at one point, maybe I'll look again.

I like the mention of the issue label that you added. Advertising our errata list puts us a step ahead of some other parts of the Rust project :p

RalfJung: > When rustc accepts your program then mangles it in codegen because of a soundness hole, you know something is afoot.

rustc may also accept your program and happen to not exploit the UB. Then there's no indication that it behaves buggily.

> I also think it's a bit sketchy that an I-misses-ub has been open for (almost) 5 years. I know I looked over the implementation that would have to be done to fix it at one point, maybe I'll look again.

Yeah, it's a tricky one. I was planning to work on it this summer.

Issue #3577, opened 2024-05-06
Implement aligned_alloc


RalfJung:
C11 has a new allocation function: aligned_alloc. We should implement it in Miri. Docs can be found [here](https://en.cppreference.com/w/cpp/memory/c/aligned_alloc).

RalfJung:
Ah, but the function spec is completely terrible of course...

> alignment 	- 	specifies the alignment. Must be a valid alignment supported by the implementation. 

There is no guarantee at all for what is a "valid alignment supported by the implementation". So this can basically always return null, great.

Given the comments in the docs, it seems like a reasonable choice of "valid alignment supported by the implementation" is to require the alignment to be at least the size of a pointer. I have no clue why an implementation would reject smaller alignment, but it is what posix_memalign does so it seems to be common... somehow...

workingjubilee:
indeed, `void*` minimum in practice, it seems: https://www.gnu.org/software/gnulib/manual/html_node/aligned_005falloc.html

workingjubilee:
though asking for too-large alignments is also bad: https://github.com/rust-lang/rust/pull/124788/commits/53bd38b7c5d5dc43feced7cf76904fa5b788de0c

RalfJung:
> indeed, `void*` minimum in practice, it seems: https://www.gnu.org/software/gnulib/manual/html_node/aligned_005falloc.html

That's for that one implementation. But who knows about others...



> though asking for too-large alignments is also bad: [rust-lang/rust@53bd38b](https://github.com/rust-lang/rust/commit/53bd38b7c5d5dc43feced7cf76904fa5b788de0c)

That's "obviously" a platform bug though, I don't think Miri should model that.
(Returning NULL would be fine, but returning an unaligned value is clearly violating the spec.)

RalfJung:
> indeed, `void*` minimum in practice, it seems: https://www.gnu.org/software/gnulib/manual/html_node/aligned_005falloc.html

Looking at the spec again, it says fundamental alignments must be supported. So I think a `void*` minimum is actually non-compliant.

RalfJung:
:joy: turns out libc doesn't have this function yet... https://github.com/rust-lang/libc/issues/3689. That certainly makes testing harder.

workingjubilee:
> Looking at the spec again, it says fundamental alignments must be supported. So I think a void* minimum is actually non-compliant. macOS (and Illumos) implementations of this function are buggy.

[...]

PR #3585, opened 2024-05-07
support aligned_alloc for unixes.

devnexen:
Fixes https://github.com/rust-lang/miri/issues/3577

### DIFF
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -295,6 +295,17 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                     }
                 }
             }
+            "aligned_alloc" => {
+                // This is a C11 function, we assume all Unixes have it.
+                // (MSVC explicitly does not support this.)
+                let [align, size] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let align = this.read_target_usize(align)?;
+                let size = this.read_target_usize(size)?;
+
+                let res = this.aligned_alloc(align, size)?;
+                this.write_pointer(res, dest)?;
+            }
 
             // Dynamic symbol loading
             "dlsym" => {
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -300,9 +300,6 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 // (MSVC explicitly does not support this.)
                 let [align, size] =
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
-                let align = this.read_target_usize(align)?;
-                let size = this.read_target_usize(size)?;
-
                 let res = this.aligned_alloc(align, size)?;
                 this.write_pointer(res, dest)?;
             }
--- a/tests/fail-dep/libc/aligned_alloc_size_zero_leak.rs
+++ b/tests/fail-dep/libc/aligned_alloc_size_zero_leak.rs
@@ -1,0 +1,15 @@
+//@ignore-target-windows: Windows does not support the standard C11 aligned_alloc.
+
+fn main() {
+    // libc doesn't have this function (https://github.com/rust-lang/libc/issues/3689),
+    // so we declare it ourselves.
+    extern "C" {
+        fn aligned_alloc(alignment: libc::size_t, size: libc::size_t) -> *mut libc::c_void;
+    }
+
+    // Make sure even zero-sized allocations need to be freed.
+
+    unsafe {
+        aligned_alloc(2, 0); //~ERROR: memory leaked
+    }
+}
--- a/tests/pass-dep/libc/libc-mem.rs
+++ b/tests/pass-dep/libc/libc-mem.rs
@@ -148,53 +148,55 @@ fn test_calloc() {
 
 #[cfg(not(target_os = "windows"))]
 fn test_memalign() {
-    // A normal allocation.
-    unsafe {
-        let mut ptr: *mut libc::c_void = ptr::null_mut();
-        let align = 8;
-        let size = 64;
-        assert_eq!(libc::posix_memalign(&mut ptr, align, size), 0);
-        assert!(!ptr.is_null());
-        assert!(ptr.is_aligned_to(align));
-        ptr.cast::<u8>().write_bytes(1, size);
-        libc::free(ptr);
-    }
+    for _ in 0..16 {
+        // A normal allocation.
+        unsafe {
+            let mut ptr: *mut libc::c_void = ptr::null_mut();
+            let align = 8;
+            let size = 64;
+            assert_eq!(libc::posix_memalign(&mut ptr, align, size), 0);
+            assert!(!ptr.is_null());
+            assert!(ptr.is_aligned_to(align));
+            ptr.cast::<u8>().write_bytes(1, size);
+            libc::free(ptr);
+        }
 
-    // Align > size.
-    unsafe {
-        let mut ptr: *mut libc::c_void = ptr::null_mut();
-        let align = 64;
-        let size = 8;
-        assert_eq!(libc::posix_memalign(&mut ptr, align, size), 0);
-        assert!(!ptr.is_null());
-        assert!(ptr.is_aligned_to(align));
-        ptr.cast::<u8>().write_bytes(1, size);
-        libc::free(ptr);
-    }
+        // Align > size.
+        unsafe {
+            let mut ptr: *mut libc::c_void = ptr::null_mut();
+            let align = 64;
+            let size = 8;
+            assert_eq!(libc::posix_memalign(&mut ptr, align, size), 0);
+            assert!(!ptr.is_null());
+            assert!(ptr.is_aligned_to(align));
+            ptr.cast::<u8>().write_bytes(1, size);
+            libc::free(ptr);
+        }
 
-    // Size not multiple of align
-    unsafe {
-        let mut ptr: *mut libc::c_void = ptr::null_mut();
-        let align = 16;
-        let size = 31;
-        assert_eq!(libc::posix_memalign(&mut ptr, align, size), 0);
-        assert!(!ptr.is_null());
-        assert!(ptr.is_aligned_to(align));
-        ptr.cast::<u8>().write_bytes(1, size);
-        libc::free(ptr);
-    }
+        // Size not multiple of align
+        unsafe {
+            let mut ptr: *mut libc::c_void = ptr::null_mut();
+            let align = 16;
+            let size = 31;
+            assert_eq!(libc::posix_memalign(&mut ptr, align, size), 0);
+            assert!(!ptr.is_null());
+            assert!(ptr.is_aligned_to(align));
+            ptr.cast::<u8>().write_bytes(1, size);
+            libc::free(ptr);
+        }
 
-    // Size == 0
-    unsafe {
-        let mut ptr: *mut libc::c_void = ptr::null_mut();
-        let align = 64;
-        let size = 0;
-        assert_eq!(libc::posix_memalign(&mut ptr, align, size), 0);
-        // Non-null pointer is returned if size == 0.
-        // (This is not a guarantee, it just reflects our current behavior.)
-        assert!(!ptr.is_null());
-        assert!(ptr.is_aligned_to(align));
-        libc::free(ptr);
+        // Size == 0
+        unsafe {
+            let mut ptr: *mut libc::c_void = ptr::null_mut();
+            let align = 64;
+            let size = 0;
+            assert_eq!(libc::posix_memalign(&mut ptr, align, size), 0);
+            // Non-null pointer is returned if size == 0.
+            // (This is not a guarantee, it just reflects our current behavior.)
+            assert!(!ptr.is_null());
+            assert!(ptr.is_aligned_to(align));
+            libc::free(ptr);
+        }
     }
 
     // Non-power of 2 align
@@ -260,20 +262,20 @@ fn test_aligned_alloc() {
         assert_eq!(p, ptr::null_mut());
     }
 
-    // alignment lesser than a word but still a successful allocation
-    unsafe {
-        let p = aligned_alloc(1, 4);
-        assert!(!p.is_null());
-        assert!(p.is_aligned_to(4));
-        libc::free(p);
-    }
-
     // repeated tests on correct alignment/size
     for _ in 0..16 {
+        // alignment 1, size 4 should succeed and actually must align to 4 (because C says so...)
         unsafe {
-            let p = aligned_alloc(16, 16);
+            let p = aligned_alloc(1, 4);
             assert!(!p.is_null());
-            assert!(p.is_aligned_to(16));
+            assert!(p.is_aligned_to(4));
+            libc::free(p);
+        }
+
+        unsafe {
+            let p = aligned_alloc(64, 64);
+            assert!(!p.is_null());
+            assert!(p.is_aligned_to(64));
             libc::free(p);
         }
     }
--- a/tests/pass-dep/libc/libc-mem.rs
+++ b/tests/pass-dep/libc/libc-mem.rs
@@ -241,6 +241,44 @@ fn test_reallocarray() {
     }
 }
 
+#[cfg(not(target_os = "windows"))]
+fn test_aligned_alloc() {
+    // libc doesn't have this function (https://github.com/rust-lang/libc/issues/3689),
+    // so we declare it ourselves.
+    extern "C" {
+        fn aligned_alloc(alignment: libc::size_t, size: libc::size_t) -> *mut libc::c_void;
+    }
+    // size not a multiple of the alignment
+    unsafe {
+        let p = aligned_alloc(16, 3);
+        assert_eq!(p, ptr::null_mut());
+    }
+
+    // alignment not power of 2
+    unsafe {
+        let p = aligned_alloc(63, 8);
+        assert_eq!(p, ptr::null_mut());
+    }
+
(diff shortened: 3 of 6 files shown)


RalfJung:
This is for https://github.com/rust-lang/miri/issues/3577, right? When a PR works on an issue, please mention the issue in the PR description. Generally, PR descriptions should hardly ever be empty. Please spend a bit of time giving context to the change -- which issue it resolves, which new feature it implements and how, things like that. It makes review a lot easier.

src/shims/unix/foreign_items.rs
```diff
@@ -315,6 +315,41 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                let [align, size] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let align = this.read_target_usize(align)?;
+                let size = this.read_target_usize(size)?;
+
+                // The size must be a power of the alignment.
+                // The specs says nothing about error handling other than null
+                // pointer but aligned_alloc Linux/macOs implementations set EINVAL. Note that
+                // on FreeBSD setting a size not being a power of alignment is UB but we do not
+                // want it.
```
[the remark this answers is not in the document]
RalfJung: Ah no actually it can't be UB

> Passing a size which is not an integral multiple of alignment or an alignment which is not valid or not supported by the implementation causes the function to fail and return a null pointer

Is it truly UB in FreeBSD? That's a critical bug in their implementation then, may be worth reporting.

But, we shouldn't set_last_error, as the spec says nothing like that.

src/shims/unix/foreign_items.rs
```diff
@@ -315,6 +315,41 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                //
+                // http://en.cppreference.com/w/cpp/memory/c/aligned_alloc
+                // Linux: https://linux.die.net/man/3/aligned_alloc
+                // FreeBSD: https://man.freebsd.org/cgi/man.cgi?query=aligned_alloc&apropos=0&sektion=3&manpath=FreeBSD+9-current&format=html
+                match size.checked_rem(align) {
+                    Some(0) => {
```
RalfJung, line 334:
```suggestion
                    Some(0) if align.is_power_of_two() && align >= this.pointer_size().bytes() => {
```
Then you need only one error path

tests/pass-dep/libc/libc-mem.rs
```diff
@@ -237,6 +237,32 @@ fn test_reallocarray() {
     }
 }
 
+#[cfg(not(target_os = "windows"))]
+fn test_aligned_alloc() {
+    extern "C" {
+        fn aligned_alloc(_: libc::size_t, _: libc::size_t) -> *mut libc::c_void;
+    }
+    // wrong size
```
RalfJung, line 245:
```suggestion
    // wrong size: not a multiple of the alignment
```

src/shims/unix/foreign_items.rs
```diff
@@ -315,6 +315,41 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+
+                // The size must be a power of the alignment.
+                // The specs says nothing about error handling other than null
+                // pointer but aligned_alloc Linux/macOs implementations set EINVAL. Note that
+                // on FreeBSD setting a size not being a power of alignment is UB but we do not
+                // want it.
+                //
+                // http://en.cppreference.com/w/cpp/memory/c/aligned_alloc
+                // Linux: https://linux.die.net/man/3/aligned_alloc
+                // FreeBSD: https://man.freebsd.org/cgi/man.cgi?query=aligned_alloc&apropos=0&sektion=3&manpath=FreeBSD+9-current&format=html
```
RalfJung, line 332:
```suggestion
                // The size must be a multiple of the alignment.
                // Alignment must be a power of 2, and "supported by the implementation".
                // If any of these are violated, we have to return NULL.
                //
                // Linux also sets EINVAL, but that's non-standard behavior that we do not emulate.
                // FreeBSD says some of these cases are UB but that's violating the C standard.
                //
                // http://en.cppreference.com/w/cpp/memory/c/aligned_alloc
                // Linux: https://linux.die.net/man/3/aligned_alloc
                // FreeBSD: https://man.freebsd.org/cgi/man.cgi?query=aligned_alloc&apropos=0&sektion=3&manpath=FreeBSD+9-current&format=html
```

tests/pass-dep/libc/libc-mem.rs
```diff
@@ -237,6 +237,32 @@ fn test_reallocarray() {
     }
 }
 
+#[cfg(not(target_os = "windows"))]
+fn test_aligned_alloc() {
+    extern "C" {
+        fn aligned_alloc(_: libc::size_t, _: libc::size_t) -> *mut libc::c_void;
+    }
```
RalfJung, line 244: Please import the function from `libc` rather than declaring it here.

devnexen: it is not in libc but only for a handful of platforms
https://github.com/search?q=repo%3Arust-lang%2Flibc%20aligned_alloc&type=code

RalfJung:
On size 0 this should probably behave like https://github.com/rust-lang/miri/pull/3580. That's already what you implemented now, but there should be tests ensuring we detect the memory leak when a size-zero allocation was not freed, and size-0-double-free.

src/shims/unix/foreign_items.rs
RalfJung: Probably the better way to do this is a little helper function `is_align_supported` or so.

RalfJung:
I realized I missed an important part of the spec.

> Fundamental alignments are always supported.

So... I think we have to support small alignments as well.

I think the point of the posix_memalign example is to mention that the alignment must be a power of two?

RalfJung:
@rustbot author
(please do `rustbot ready` when this is ready for the next round of review)

RalfJung:
Looking at the libc crate, it doesn't seem like it has this function on any target we support. :joy: 

Okay, so... how do we find out which targets support this function? Linux definitely does, at least I have a manpage for it.

tests/pass-dep/libc/libc-mem.rs
RalfJung: Good point, I reported that as an issue: https://github.com/rust-lang/libc/issues/3689

RalfJung:
I did a bit of googling and found evidence that it exists for all platforms we support (see the links in https://github.com/rust-lang/libc/issues/3689). For Iluumos the man page says

>  The value of alignment is constrained, it
     must be a power of two and it must be greater than or equal to the size of
     a word on the platform.

That is clearly a violation of the C standard, and hence a platform bug. I don't think  we should emulate that.

So, yeah seems fine to just implement it on all targets then (with it being a C11 standard function -- that's more than 10 years old at this point), and declare it in the test like you did.

src/shims/unix/foreign_items.rs
```diff
@@ -315,6 +315,41 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                     }
                 }
             }
+            "aligned_alloc" => {
+                let [align, size] =
```
RalfJung, line 319:
```suggestion
                // This is a C11 function, we assume all Unices have it.
                let [align, size] =
```

tests/pass-dep/libc/libc-mem.rs
```diff
@@ -237,6 +237,32 @@ fn test_reallocarray() {
     }
 }
 
+#[cfg(not(target_os = "windows"))]
+fn test_aligned_alloc() {
+    extern "C" {
```
RalfJung, line 242:
```suggestion
    // libc doesn't have this function (https://github.com/rust-lang/libc/issues/3689),
    // so we declare it ourselves.
    extern "C" {
```

tests/pass-dep/libc/libc-mem.rs
```diff
@@ -237,6 +237,32 @@ fn test_reallocarray() {
...
+        fn aligned_alloc(_: libc::size_t, _: libc::size_t) -> *mut libc::c_void;
```
RalfJung, line 243:
```suggestion
        fn aligned_alloc(alignment: libc::size_t, size: libc::size_t) -> *mut libc::c_void;
```

src/shims/unix/foreign_items.rs
```diff
@@ -315,6 +315,41 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                // FreeBSD: https://man.freebsd.org/cgi/man.cgi?query=aligned_alloc&apropos=0&sektion=3&manpath=FreeBSD+9-current&format=html
+                match size.checked_rem(align) {
+                    Some(0) => {
+                        if !align.is_power_of_two() || align < this.pointer_size().bytes() {
+                            let _ = this.set_last_error(this.eval_libc("EINVAL"));
+                            this.write_null(dest)?;
+                        } else {
+                            let ptr = this.allocate_ptr(
+                                Size::from_bytes(size),
+                                Align::from_bytes(align).unwrap(),
```
RalfJung, line 341: Looking at the actual C23 standard, things are even more complicated.^^ We have to honor `min_align` even if the requested alignment is smaller. So e.g. `aligned_alloc(1, 4)` has to actually return a 4-aligned allocation even though only alignment 1 was requested.

src/shims/unix/foreign_items.rs
devnexen: well windows kind of support it but it s different. `_aligned_alloc` but you need to call `_aligned_free`, kind of complicated ..

RalfJung: Yeah, so they don't support the official aligned_alloc which says you can free with `free`. "kind of" support isn't support. ;)

src/shims/unix/foreign_items.rs
RalfJung: What about this suggestion? There are still some mistakes in your text, like saying the size must be a "power" of the alignment, which is not correct.

src/shims/unix/foreign_items.rs
```diff
@@ -311,6 +311,48 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                            let ptr = this.allocate_ptr(
+                                Size::from_bytes(size),
+                                align,
+                                MiriMemoryKind::C.into(),
+                            )?;
+                            this.write_pointer(ptr, dest)?;
+                        }
+                    },
+                    _ => {
+                        let _ = this.set_last_error(this.eval_libc("EINVAL"));
```
RalfJung, line 351: I also wrote some comments saying to not set errno please.

RalfJung:
@rustbot author
please go over all my comments from the previous review and either make sure they are handled, or argue why you think they shouldn't be handled.

src/shims/unix/foreign_items.rs
```diff
@@ -295,6 +296,44 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                //
+                // Linux also sets errno to EINVAL, but that's non-standard behavior that we do not
+                // emulate/
+                // FreeBSD says some of these cases are UB but that's violating the C standard.
+                // http://en.cppreference.com/w/cpp/memory/c/aligned_alloc
+                // Linux: https://linux.die.net/man/3/aligned_alloc
+                // FreeBSD: https://man.freebsd.org/cgi/man.cgi?query=aligned_alloc&apropos=0&sektion=3&manpath=FreeBSD+9-current&format=html
+                match size.checked_rem(align) {
+                    Some(0) if align.is_power_of_two() => {
+                            let align = this.min_align(align, size);
```
RalfJung, line 324: You can just use `align.max(this.malloc_align(size))`, no need to write a new function for that

src/shims/unix/foreign_items.rs
```diff
@@ -295,6 +296,44 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+            "aligned_alloc" => {
+                // This is a C11 function, we assume all Unixes have it.
+                // (MSVC explicitly does not support this.)
+                let [align, size] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let align = this.read_target_usize(align)?;
+                let size = this.read_target_usize(size)?;
+
+                // The size must be a multiple of the alignment.
+                // Alignment must be a power of 2, and "supported by the implementation".
```
RalfJung, line 308:
```suggestion
                // Alignment must be a power of 2, and "supported by the implementation".
                // We decide that "supported by the implementation" means that the
                // size must be a multiple of the alignment. (This restriction seems common
                // enough that it is stated on <https://en.cppreference.com/w/c/memory/aligned_alloc>
                // as a general rule, but the actual standard has no such rule.)
```

src/shims/unix/foreign_items.rs
```diff
@@ -295,6 +296,44 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                // If any of these are violated, we have to return NULL.
+                // All fundamental alignments must be supported.
+                // We decide that our implementation supports all alignments.
```
RalfJung, line 311:
```suggestion
```

src/shims/unix/foreign_items.rs
```diff
@@ -295,6 +296,44 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                //
+                // macOS and Illumos are buggy in that they require the alignment
+                // to be at least the size of a pointer. We do not emulate those platform bugs.
```
RalfJung, line 314:
```suggestion
                // to be at least the size of a pointer, so they do not support all fundamental
                // alignments. We do not emulate those platform bugs.
```

src/shims/unix/foreign_items.rs
```diff
@@ -295,6 +296,44 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                //
+                // Linux also sets errno to EINVAL, but that's non-standard behavior that we do not
+                // emulate/
```
RalfJung, line 317:
```suggestion
                // emulate.
```

tests/fail-dep/libc/aligned_alloc.rs
RalfJung, line 1: Please name the test aligned_alloc_leak or so to indicate what it tests.

tests/pass-dep/libc/libc-mem.rs
```diff
@@ -241,6 +241,42 @@ fn test_reallocarray() {
     }
 }
 
+#[cfg(not(any(target_os = "windows", target_os = "wasi")))]
```
RalfJung, line 244: Why not on wasi?

tests/pass-dep/libc/libc-mem.rs
```diff
@@ -241,6 +241,42 @@ fn test_reallocarray() {
...
+        libc::free(p);
+    }
+
+    // finally ..
+    unsafe {
+        let p = aligned_alloc(16, 16);
+        assert!(!p.is_null());
+        assert!(p.is_aligned_to(16));
+        libc::free(p);
+    }
```
RalfJung, line 277: Maybe put these in a `for _ in 0..16` or so, because the actual alignment is random so we want to get some extra test coverage.

tests/pass-dep/libc/libc-mem.rs
RalfJung: Where have you seen that? libc even declares aligned_alloc for wasi so I'd think that it exists there.

devnexen:
alright .. going to reactive wasi then, we shall see.

RalfJung:
Oh d'oh, wasi is not a unix, I keep forgetting that. Sorry!

We shouldn't duplicate the function. Please either move the function to alloc.rs so it can be shared - or it's also okay to remove wasi again and leave it to future work.

tests/pass-dep/libc/libc-mem.rs
```diff
@@ -241,6 +241,44 @@ fn test_reallocarray() {
...
+    unsafe {
+        let p = aligned_alloc(1, 4);
+        assert!(!p.is_null());
+        assert!(p.is_aligned_to(4));
+        libc::free(p);
+    }
+
+    for i in 0..16 {
+        unsafe {
+            let p = aligned_alloc(i, i);
```
RalfJung, line 273: I don't understand the point of this -- most of these numbers aren't even valid alignments?

tests/fail-dep/libc/aligned_alloc_size_zero_leak.rs
```diff
@@ -0,0 +1,15 @@
...
+
+fn main() {
+    // libc doesn't have this function (https://github.com/rust-lang/libc/issues/3689),
+    // so we declare it ourselves.
+    extern "C" {
+        fn aligned_alloc(alignment: libc::size_t, size: libc::size_t) -> *mut libc::c_void;
+    }
+
+    // Make sure event zero-sized allocations need to be freed.
+
```
RalfJung, line 11:
```suggestion
    // Make sure even zero-sized allocations need to be freed.
```

tests/pass-dep/libc/libc-mem.rs
devnexen: ah ok I did not get you.

RalfJung:
I assume the "WIP" in the PR description can be removed now.

RalfJung:
Thanks for bearing with me through all these rounds of review. :)
@bors r+

PR #3588, opened 2024-05-08
Allow test targets to be set via CLI args

### LINKED ISSUE #3584: Cannot add new targets after installation of toolchain
Basically the annoyance is:

1. Run `.\miri toolchain`
2. Later, need more targets for the toolchain (Linux, OSX, ...)
3. Run `.\miri toolchain -t x86_64-unknown-linux-gnu aarch64-apple-darwin`
4. Instantly returns without doing anything

I believe this is a combination of factors. One being this portion of miri-script ignoring components & targets:

https://github.com/rust-lang/miri/blob/cf2df2d7d84c8ac06d3ab8200f290e9ec8128951/miri-script/src/commands.rs#L184-L203

Which doesn't pass the command to `rustup-toolchain-install-master` at all, the also the fact that `rustup-toolchain-install-master` doesn't add new targets to installed toolchains anyways. 

The solution is to delete the toolchain with rustup then rerun `.\miri toolchain -t whatever` right now.

kennytm/rustup-toolchain-install-master#19 semi-related

Since the upstream tool seems to be mostly inactive I'm not going to work on this unless there is any indication of activity upstream.

RossSmyth:
Fixes #3584

I'm not a pro shell script reader as I am a Windows user, but we shall see if the CI script broke.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,6 +73,7 @@ For example:
 You can (cross-)run the entire test suite using:
 
 ```sh
+./miri test
 ./miri test --target i686-unknown-linux-gnu
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,14 +72,13 @@ For example:
 
 You can (cross-)run the entire test suite using:
 
-```
-./miri test
-MIRI_TEST_TARGET=i686-unknown-linux-gnu ./miri test
+```sh
+./miri test --target i686-unknown-linux-gnu
 ```
 
 `./miri test FILTER` only runs those tests that contain `FILTER` in their filename (including the
-base directory, e.g. `./miri test fail` will run all compile-fail tests). These filters are passed
-to `cargo test`, so for multiple filters you need to use `./miri test -- FILTER1 FILTER2`.
+base directory, e.g. `./miri test fail` will run all compile-fail tests). Multiple filters 
+are supported: `./miri test FILTER1 FILTER2`.
 
 #### Fine grained logging
 
--- a/miri-script/src/commands.rs
+++ b/miri-script/src/commands.rs
@@ -23,7 +23,9 @@ const JOSH_PORT: &str = "42042";
 
 impl MiriEnv {
     /// Returns the location of the sysroot.
-    fn build_miri_sysroot(&mut self, quiet: bool) -> Result<PathBuf> {
+    ///
+    /// If the target is None the sysroot will be built for the host machine.
+    fn build_miri_sysroot(&mut self, quiet: bool, target: Option<&str>) -> Result<PathBuf> {
         if let Some(miri_sysroot) = self.sh.var_os("MIRI_SYSROOT") {
             // Sysroot already set, use that.
             return Ok(miri_sysroot.into());
@@ -35,26 +37,27 @@ impl MiriEnv {
         self.build(path!(self.miri_dir / "Cargo.toml"), &[], quiet)?;
         self.build(&manifest_path, &[], quiet)?;
 
-        let target = &match self.sh.var("MIRI_TEST_TARGET") {
-            Ok(target) => vec!["--target".into(), target],
-            Err(_) => vec![],
-        };
+        let target_flag =
+            &if let Some(target) = target { vec!["--target", target] } else { vec![] };
+
         if !quiet {
-            match self.sh.var("MIRI_TEST_TARGET") {
-                Ok(target) => eprintln!("$ (building Miri sysroot for {target})"),
-                Err(_) => eprintln!("$ (building Miri sysroot)"),
+            if let Some(target) = target {
+                eprintln!("$ (building Miri sysroot for {target})");
+            } else {
+                eprintln!("$ (building Miri sysroot)");
             }
         }
+
         let output = cmd!(self.sh,
             "cargo +{toolchain} --quiet run {cargo_extra_flags...} --manifest-path {manifest_path} --
-             miri setup --print-sysroot {target...}"
+             miri setup --print-sysroot {target_flag...}"
         ).read();
         let Ok(output) = output else {
             // Run it again (without `--print-sysroot` or `--quiet`) so the user can see the error.
             cmd!(
                 self.sh,
                 "cargo +{toolchain} run {cargo_extra_flags...} --manifest-path {manifest_path} --
-                miri setup {target...}"
+                miri setup {target_flag...}"
             )
             .run()
             .with_context(|| "`cargo miri setup` failed")?;
@@ -161,7 +164,7 @@ impl Command {
             Command::Install { flags } => Self::install(flags),
             Command::Build { flags } => Self::build(flags),
             Command::Check { flags } => Self::check(flags),
-            Command::Test { bless, flags } => Self::test(bless, flags),
+            Command::Test { bless, flags, target } => Self::test(bless, flags, target),
             Command::Run { dep, verbose, many_seeds, flags } =>
                 Self::run(dep, verbose, many_seeds, flags),
             Command::Fmt { flags } => Self::fmt(flags),
@@ -446,16 +449,23 @@ impl Command {
         Ok(())
     }
 
-    fn test(bless: bool, flags: Vec<OsString>) -> Result<()> {
+    fn test(bless: bool, flags: Vec<OsString>, target: Option<String>) -> Result<()> {
         let mut e = MiriEnv::new()?;
+
+        if let Some(target) = target.as_deref() {
+            // Tell the sysroot which target to test.
+            e.sh.set_var("MIRI_TEST_TARGET", target);
+        }
+
         // Prepare a sysroot.
-        e.build_miri_sysroot(/* quiet */ false)?;
+        e.build_miri_sysroot(/* quiet */ false, target.as_deref())?;
 
         // Then test, and let caller control flags.
         // Only in root project as `cargo-miri` has no tests.
         if bless {
             e.sh.set_var("RUSTC_BLESS", "Gesundheit");
         }
+
         e.test(path!(e.miri_dir / "Cargo.toml"), &flags)?;
         Ok(())
     }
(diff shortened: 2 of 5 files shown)


miri-script/src/commands.rs
```diff
@@ -35,26 +37,27 @@ impl MiriEnv {
         self.build(path!(self.miri_dir / "Cargo.toml"), &[], quiet)?;
         self.build(&manifest_path, &[], quiet)?;
 
-        let target = &match self.sh.var("MIRI_TEST_TARGET") {
-            Ok(target) => vec!["--target".into(), target],
-            Err(_) => vec![],
-        };
+        let target_flag =
+            &if let Some(target) = target { vec!["--target", target] } else { vec![] }[..];
```
RalfJung, line 41: Why is `[..]` needed now? It wasn't needed before

miri-script/src/commands.rs
```diff
@@ -446,10 +449,28 @@ impl Command {
...
+            e.sh.set_var("MIRI_TEST_TARGET", target);
+
+            if has_flag(&flags, "target") {
+                // We must remove the flag if it came from the CLI because test driver has no idea
+                // what the target flag does.
+                let position = flags.iter().position(|flag| flag == "--target").unwrap();
+
+                // Remove both the key and the value
+                flags.remove(position);
+                flags.remove(position);
```
RalfJung, line 468: Yeah sorry that's exactly what I said I didn't like... randomly mutating things in the middle of the CLI argument stream.

I think the better approach is to
- make `Test` [here](https://github.com/rust-lang/miri/blob/f399ed23678ff15831588272cc1343a5398ff7be/miri-script/src/main.rs#L31-L36) have an extra `target: Option<String>`
- Adjust the parsing [here](https://github.com/rust-lang/miri/blob/f399ed23678ff15831588272cc1343a5398ff7be/miri-script/src/main.rs#L149) to be more like `run` (i.e., with a loop) to extract leading `--bless` and `--target` flags

miri-script/src/util.rs
```diff
@@ -143,7 +173,7 @@ impl MiriEnv {
         let MiriEnv { toolchain, cargo_extra_flags, .. } = self;
         cmd!(
             self.sh,
-            "cargo +{toolchain} test {cargo_extra_flags...} --manifest-path {manifest_path} {args...}"
+            "cargo +{toolchain} test {cargo_extra_flags...} --manifest-path {manifest_path} -- {args...}"
```
RalfJung, line 176: This should probably be done by the caller, not here. Here it's somewhat inconsistent.

Then the docs will also need adjusting, and `ci.sh` as well... seems like passing two `--` actually works but it shouldn't be necessary. The README also needs updating.

miri-script/src/commands.rs
RossSmyth: Not needed, I just like slices better

RossSmyth:
Ah I forgot about the env vars.

CONTRIBUTING.md
```diff
@@ -72,14 +72,13 @@ For example:
...
 
-```
-./miri test
-MIRI_TEST_TARGET=i686-unknown-linux-gnu ./miri test
+```sh
+./miri test --target i686-unknown-linux-gnu
 ```
 
 `./miri test FILTER` only runs those tests that contain `FILTER` in their filename (including the
 base directory, e.g. `./miri test fail` will run all compile-fail tests). These filters are passed
```
RalfJung, line 80:
```suggestion
base directory, e.g. `./miri test fail` will run all compile-fail tests). Multiple filters
```

CONTRIBUTING.md
```diff
@@ -72,14 +72,13 @@ For example:
...
-to `cargo test`, so for multiple filters you need to use `./miri test -- FILTER1 FILTER2`.
+to the test driver, so for multiple filters you can use `./miri test FILTER1 FILTER2`.
```
RalfJung, line 81:
```suggestion
are supported: `./miri test FILTER1 FILTER2`.
```

miri-script/src/commands.rs
RalfJung: I find it to be just syntactic noise, please remove it.

miri-script/src/commands.rs
```diff
@@ -446,16 +449,23 @@ impl Command {
         Ok(())
     }
 
-    fn test(bless: bool, flags: Vec<OsString>) -> Result<()> {
+    fn test(bless: bool, flags: Vec<OsString>, target: Option<String>) -> Result<()> {
         let mut e = MiriEnv::new()?;
+
+        if let Some(target) = target.as_deref() {
+            // This must be set, it is what makes miri actually cross-interpret.
```
RalfJung, line 456:
```suggestion
            // Tell the sysroot which target to test.
```

miri-script/src/commands.rs
```diff
@@ -470,29 +480,26 @@ impl Command {
...
-        } else if let Ok(target) = std::env::var("MIRI_TEST_TARGET") {
-            // Convert `MIRI_TEST_TARGET` into `--target`.
-            flags.push("--target".into());
-            flags.push(target.into());
+
+            // Set target if not in the arguments
+            if !has_flag(&flags, "target") {
+                flags.push("--target".into());
+                flags.push(target.into());
+            }
```
RalfJung, line 493: I don't see the point of this part of the PR.

miri-script/src/main.rs
```diff
@@ -147,12 +150,30 @@ fn main() -> Result<()> {
...
+                        target = args
+                            .peek()
+                            .map(|s| s.clone().into_string())
+                            .transpose()
+                            .map_err(|os_str| anyhow!("Invalid target string: {:?}", os_str))?;
+                    }
+                    _ => break,
+                }
+
+                args.next();
```
RalfJung, line 171:
```suggestion
                // Consume the flag, look at the next one.
                args.next().unwrap();
```

miri-script/src/main.rs
```diff
@@ -147,12 +150,30 @@ fn main() -> Result<()> {
...
-                args.next().unwrap();
+            let mut target = std::env::var("MIRI_TEST_TARGET").ok();
+            let mut bless = false;
+
+            while let Some(arg) = args.peek().and_then(|s| s.to_str()) {
+                match arg {
+                    "--bless" => bless = true,
+                    "--target" => {
+                        // Skip "--target"
+                        args.next();
```
RalfJung, line 161:
```suggestion
                        args.next().unwrap();
```

miri-script/src/main.rs
```diff
@@ -147,12 +150,30 @@ fn main() -> Result<()> {
...
+                        target = args
+                            .peek()
+                            .map(|s| s.clone().into_string())
+                            .transpose()
+                            .map_err(|os_str| anyhow!("Invalid target string: {:?}", os_str))?;
```
RalfJung, line 166: This looks unnecessarily complicated... if `args.peek()` is `None`, you can just throw an error.

miri-script/src/commands.rs
RossSmyth: It preserves what it was doing before, keeping that environment variable and CLI args in sync.

RalfJung: We already did that before though -- we supported both the env var and the CLI arg. I don't think any change is needed in `fn run`.

RossSmyth: There is no semantic change. It does what the current HEAD does. In current head, if the enviroment variable was set, but there is not a `--target` flag, it inserts them.

https://github.com/rust-lang/miri/blob/d9188da764dafd6f46affeef03ec8d4f96e9640c/miri-script/src/commands.rs#L482-L486

I can test if it works without keeping them in sync though.

RalfJung: > There is no semantic change.

Right, so the code looks more complicated than before and there's no semantic change... so why do this...?

RossSmyth: Because it was done in multiple places previously so I factored it to its own function. I will revert it.

RalfJung: I don't think it was? it was multiple places with your first attempt for `fn test` but we changed that and now it's just `fn run`.

FWIW I am planning a follow-up PR that entirely removes MIRI_TEST_TARGET from the public interface. I never liked having two ways of setting the target for `./miri run`, and now that you're adding `--target` support for `./miri test` we can just use `--target` everywhere.

RossSmyth: Yeah, I meant in a previous attempt. 

I made the changes for removing MIRI_TEST_TARGET then got to the ci script and realized I don't know bash well enough to do those changes.

CONTRIBUTING.md
```diff
@@ -72,14 +72,13 @@ For example:
 
 You can (cross-)run the entire test suite using:
 
-```
-./miri test
-MIRI_TEST_TARGET=i686-unknown-linux-gnu ./miri test
+```sh
+./miri test --target i686-unknown-linux-gnu
```
RalfJung, line 76:
```suggestion
./miri test
./miri test --target i686-unknown-linux-gnu
```
not sure why you removed that first line

miri-script/src/commands.rs
```diff
@@ -476,14 +486,26 @@ impl Command {
...
-        if let Some((_, target)) = target {
+
+        let target_triple = if let Some((_, target)) = target {
             // Found it!
             e.sh.set_var("MIRI_TEST_TARGET", target);
+
+            let triple = target
+                .clone()
+                .into_string()
+                .map_err(|_| anyhow!("invalid target triple encoding"))?;
```
RalfJung, line 497:
```suggestion
                .map_err(|_| anyhow!("target triple is not UTF-8"))?;
```

miri-script/src/main.rs
```diff
@@ -84,9 +87,9 @@ Just build miri. <flags> are passed to `cargo build`.
 ./miri check <flags>:
 Just check miri. <flags> are passed to `cargo check`.
 
-./miri test [--bless] <flags>:
+./miri test [--bless] [--target] <flags>:
```
RalfJung, line 90:
```suggestion
./miri test [--bless] [--target <target>] <flags>:
```

miri-script/src/main.rs
```diff
@@ -147,12 +150,38 @@ fn main() -> Result<()> {
...
+                    "--target" => {
+                        // Skip "--target"
+                        args.next().unwrap();
+
+                        // Check that there is a target triple, and that it is unicode.
+                        target = if let Some(value) = args.peek() {
+                            let target_str = value
+                                .clone()
+                                .into_string()
+                                .map_err(|_| anyhow!("invalid target triple encoding"))?;
```
RalfJung, line 168:
```suggestion
                                .map_err(|_| anyhow!("target triple is not UTF-8"))?;
```

Issue #3555, opened 2024-05-04
Use non-deterministic precision for float operations that do not have guaranteed precision


RalfJung:
As per https://github.com/rust-lang/rust/pull/124609, these operations behave non-deterministically, so we should implement that in Miri. The `apply_random_float_error` function (currently only used for x86 intrinsics) could be used to do that.

RalfJung:
This should also be used for fast-math and "algebraic" float operations.

LorrensP-2158466:
Hi, i would like to try this. I have 2 questions though. Do you mean that we should add the float_error to this part of the code?

https://github.com/rust-lang/miri/blob/95558badc942353582b742b30950412d46d69a06/src/intrinsics/mod.rs#L247C1-L263C18

And should we use the same magnitude as the x86 instrinics? https://github.com/rust-lang/miri/blob/95558badc942353582b742b30950412d46d69a06/src/shims/x86/mod.rs#L413

RalfJung:
Yes, all of those there except for sqrt should apply a float error.

There's no fixed value for the magnitude of the error so we might as well pick the same as the x86 intrinsic.

LorrensP-2158466:
Hi, I think I successfully implemented it for all but 2 match arms because I don't know how to do it nicely. It is about the `f*_algebraic` operations:
```rust
  | "fadd_algebraic"
  | "fsub_algebraic"
  | "fmul_algebraic"
  | "fdiv_algebraic"
  | "frem_algebraic"
  => {
      let [a, b] = check_arg_count(args)?;
      let a = this.read_immediate(a)?;
      let b = this.read_immediate(b)?;
      let op = match intrinsic_name {
          "fadd_algebraic" => mir::BinOp::Add,
          "fsub_algebraic" => mir::BinOp::Sub,
          "fmul_algebraic" => mir::BinOp::Mul,
          "fdiv_algebraic" => mir::BinOp::Div,
          "frem_algebraic" => mir::BinOp::Rem,
          _ => bug!(),
      };
      let res = this.binary_op(op, &a, &b)?;
      // `binary_op` already called `generate_nan` if necessary.
      this.write_immediate(*res, dest)?;
  }
```
I don't know how I can nicely implement this.
I thought about converting to a scalar int `to_scalar_int()` and then matching on the float type converting to that type, applying the error, and then writing the scalar. But maybe there is a better way?

And the `f*_fast` operation. It says this in the code:
[...]

PR #3596, opened 2024-05-11
support `f*_algebraic`

bend-n:
supports the [`f*_algebraic`](https://doc.rust-lang.org/std/intrinsics/fn.fadd_algebraic.html) intrinsics.

### DIFF
--- a/src/intrinsics/mod.rs
+++ b/src/intrinsics/mod.rs
@@ -256,6 +256,28 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 let res = this.adjust_nan(res, &[f]);
                 this.write_scalar(res, dest)?;
             }
+            #[rustfmt::skip]
+            | "fadd_algebraic"
+            | "fsub_algebraic"
+            | "fmul_algebraic"
+            | "fdiv_algebraic"
+            | "frem_algebraic"
+            => {
+                let [a, b] = check_arg_count(args)?;
+                let a = this.read_immediate(a)?;
+                let b = this.read_immediate(b)?;
+                let op = match intrinsic_name {
+                    "fadd_algebraic" => mir::BinOp::Add,
+                    "fsub_algebraic" => mir::BinOp::Sub,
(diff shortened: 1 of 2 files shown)


RalfJung:
Thanks!

This should use `adjust_nan` to avoid making any guarantees about the NaNs being returned.

Ideally, we'd also not return the *exact* result but perturb it a bit, to drive home the point that these operations are non-deterministic. But the existing fast-math intrinsics don't do that either so we can leave that to https://github.com/rust-lang/miri/issues/3555.

tests/pass/float_algebraic_math.rs
RalfJung, line 1: Please put the test into the `tests/pass/intrinsics` folder.

bend-n:
> This should use `adjust_nan` to avoid making any guarantees about the NaNs being 

the comment on `f*_fast` says that `wrapping_binary_op` already calls `generate_nan`?

RalfJung:
>      the comment on f*_fast says that wrapping_binary_op already calls generate_nan?

Ah, you're right, thanks!

src/intrinsics/mod.rs
```diff
@@ -256,6 +256,27 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                let b = this.read_immediate(b)?;
+                let op = match intrinsic_name {
+                    "fadd_algebraic" => mir::BinOp::Add,
+                    "fsub_algebraic" => mir::BinOp::Sub,
+                    "fmul_algebraic" => mir::BinOp::Mul,
+                    "fdiv_algebraic" => mir::BinOp::Div,
+                    "frem_algebraic" => mir::BinOp::Rem,
+                    _ => bug!(),
+                };
+                let res = this.wrapping_binary_op(op, &a, &b)?;
```
RalfJung, line 277:
```suggestion
                let res = this.wrapping_binary_op(op, &a, &b)?;
                // `wrapping_binary_op` already called `generate_nan` if needed.
```

RalfJung:
Thanks, looks good. :)

Please squash the commits.

PR #3599, opened 2024-05-11
Don't print unnecessary sysroot messages

saethlin:
Currently, when running `cargo miri setup`, we always print that a sysroot is being prepared, even if we just bail out of building after checking the hash. So that message is wrong; we didn't actually prepare a sysroot.

We also always print the preparing message for `cargo miri run`, even if no sysroot is prepared.

With this PR, `cargo miri run` prints no sysroot messages when an existing one is reused, and when a redundant `cargo miri setup` is requested, we print:
```
A sysroot for Miri is already available in `/home/ben/.cache/miri`.
```

### DIFF
--- a/cargo-miri/src/setup.rs
+++ b/cargo-miri/src/setup.rs
@@ -6,7 +6,7 @@ use std::fmt::Write;
 use std::path::PathBuf;
 use std::process::{self, Command};
 
-use rustc_build_sysroot::{BuildMode, SysrootBuilder, SysrootConfig};
+use rustc_build_sysroot::{BuildMode, SysrootBuilder, SysrootConfig, SysrootStatus};
 use rustc_version::VersionMeta;
 
 use crate::util::*;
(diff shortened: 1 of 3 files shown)


cargo-miri/src/setup.rs
```diff
@@ -137,32 +137,51 @@ pub fn setup(
...
-        // Be silent.
-    } else {
-        let mut msg = String::new();
-        write!(msg, "Preparing a sysroot for Miri (target: {target})").unwrap();
-        if verbose > 0 {
-            write!(msg, " in {}", sysroot_dir.display()).unwrap();
-        }
-        write!(msg, "...").unwrap();
-        if only_setup {
+    let mut msg = String::new();
```
RalfJung, line 140: This is only needed inside the closure, it seems? Does it have to be computed outside?

cargo-miri/src/setup.rs
```diff
@@ -137,32 +137,51 @@ pub fn setup(
...
         .sysroot_config(sysroot_config)
         .rustflags(rustflags)
         .cargo(cargo_cmd)
-        .build_from_source(&rust_src)
-        .unwrap_or_else(|err| {
+        .when_build_required(notify)
+        .build_from_source(&rust_src);
+    match status {
+        Ok(SysrootStatus::AlreadyCached) =>
+            if only_setup {
```
RalfJung, line 169:
```suggestion
            if only_setup  && !(print_sysroot || quiet) {
```

cargo-miri/src/setup.rs
saethlin: Constructing the message inside just requires giving the hook a lifetime that's not `'static` inside `rustc-build-sysroot.

saethlin:
Ah, fmt.

RalfJung:
I published a new rustc-build-sysroot, can you update the PR?

RalfJung:
LGTM, thanks!
@bors r+

PR #3600, opened 2024-05-11
Use non-null pointer for size 0 posix memalign

### LINKED ISSUE #3576: Make `malloc(0)` and `posix_memalign(0)` return a non-null pointer

tiif:
Fixes #3576

### DIFF
--- a/tests/pass-dep/libc/libc-mem.rs
+++ b/tests/pass-dep/libc/libc-mem.rs
@@ -190,11 +190,10 @@ fn test_memalign() {
         let align = 64;
         let size = 0;
         assert_eq!(libc::posix_memalign(&mut ptr, align, size), 0);
-        // We are not required to return null if size == 0, but we currently do.
-        // It's fine to remove this assert if we start returning non-null pointers.
-        assert!(ptr.is_null());
+        // Non-null pointer is returned if size == 0.
+        // (This is not a guarantee, it just reflects our current behavior.)
+        assert!(!ptr.is_null());
         assert!(ptr.is_aligned_to(align));
-        // Regardless of what we return, it must be `free`able.
         libc::free(ptr);
     }
 
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -259,16 +259,12 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                     let einval = this.eval_libc_i32("EINVAL");
                     this.write_int(einval, dest)?;
                 } else {
-                    if size == 0 {
-                        this.write_null(&ret)?;
-                    } else {
-                        let ptr = this.allocate_ptr(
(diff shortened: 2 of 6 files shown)


tests/pass-dep/libc/libc-mem.rs
```diff
@@ -190,11 +190,10 @@ fn test_memalign() {
         let align = 64;
         let size = 0;
         assert_eq!(libc::posix_memalign(&mut ptr, align, size), 0);
-        // We are not required to return null if size == 0, but we currently do.
-        // It's fine to remove this assert if we start returning non-null pointers.
-        assert!(ptr.is_null());
+        // Non-null pointer is returned if size == 0
+        assert!(!ptr.is_null());
+        //TODO: The line below has caused a leak?
```
tiif, line 195: The test will pass without error when line 196 is commented out, and memory leak is reported when it is included. But I don't think a memory leak should be reported since ``ptr`` is freed afterwards? 

Command used: ``./miri run --dep ./tests/pass-dep/libc/libc-mem.rs``

tests/pass-dep/libc/libc-mem.rs
```diff
@@ -190,11 +190,10 @@ fn test_memalign() {
...
         assert!(ptr.is_aligned_to(align));
-        // Regardless of what we return, it must be `free`able.
         libc::free(ptr);
```
tiif, line 197: The test will pass without error when line 196 is commented out, and memory leak is reported when it is included. But I don't think a memory leak should be reported since ``ptr`` is freed afterwards? 

Command used: ``./miri run --dep ./tests/pass-dep/libc/libc-mem.rs``

tiif:
Wait... I see the problem. ``Align::ONE`` shouldn't be used here.

tests/pass-dep/libc/libc-mem.rs
tiif: Resolved: Align::ONE shouldn't be used here.

tests/pass-dep/libc/libc-mem.rs
```diff
@@ -190,11 +190,9 @@ fn test_memalign() {
         let align = 64;
         let size = 0;
         assert_eq!(libc::posix_memalign(&mut ptr, align, size), 0);
-        // We are not required to return null if size == 0, but we currently do.
-        // It's fine to remove this assert if we start returning non-null pointers.
-        assert!(ptr.is_null());
+        // Non-null pointer is returned if size == 0
```
RalfJung, line 193:
```suggestion
        // Non-null pointer is returned if size == 0.
        // (This is not a guarantee, it just reflects our current behavior.)
```

RalfJung:
LGTM, thanks, just one comment nit.

Then please squash the commits.

tiif:
As always, thanks for the prompt review :)

PR #3603, opened 2024-05-13
Give `FileDescription::{read, write}` access to the `MiriInterpCx `

### LINKED ISSUE #3572: Give `FileDescription::{read, write}` access to the `InterpCx`
The `read` and `write` methods of `FileDescriptor` should have access to `ecx: &mut InterpCx<'mir, 'tcx, MiriMachine<'mir, 'tcx>>`. This will be needed in the future to support blocking on these operations.

This is a prerequisite for
- https://github.com/rust-lang/miri/issues/3442
- https://github.com/rust-lang/miri/issues/3445

To make that work out in terms of borrowing, I think what we will have to do is clone the `Rc` in the `read`/`write` functions in `unix/fd.rs`, so that we have a fully owned `FileDescriptor` next to a uniquely borrowed `InterpCx`.

Cc @tiif @Luv-Ray
(AFAIK @tiif is still in the design stage so this is not a blocker yet)

Luv-Ray:
fixes #3572

### DIFF
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -7,7 +7,6 @@ use std::collections::BTreeMap;
 use std::io::{self, ErrorKind, IsTerminal, Read, SeekFrom, Write};
 use std::rc::Rc;
 
-use rustc_middle::ty::TyCtxt;
 use rustc_target::abi::Size;
 
 use crate::shims::unix::*;
@@ -22,7 +21,7 @@ pub trait FileDescription: std::fmt::Debug + Any {
         &mut self,
         _communicate_allowed: bool,
         _bytes: &mut [u8],
-        _tcx: TyCtxt<'tcx>,
+        _ecx: &mut MiriInterpCx<'_, 'tcx>,
     ) -> InterpResult<'tcx, io::Result<usize>> {
         throw_unsup_format!("cannot read from {}", self.name());
     }
@@ -32,7 +31,7 @@ pub trait FileDescription: std::fmt::Debug + Any {
         &mut self,
         _communicate_allowed: bool,
         _bytes: &[u8],
-        _tcx: TyCtxt<'tcx>,
+        _ecx: &mut MiriInterpCx<'_, 'tcx>,
     ) -> InterpResult<'tcx, io::Result<usize>> {
         throw_unsup_format!("cannot write to {}", self.name());
     }
@@ -82,7 +81,7 @@ impl FileDescription for io::Stdin {
         &mut self,
         communicate_allowed: bool,
         bytes: &mut [u8],
-        _tcx: TyCtxt<'tcx>,
+        _ecx: &mut MiriInterpCx<'_, 'tcx>,
     ) -> InterpResult<'tcx, io::Result<usize>> {
         if !communicate_allowed {
             // We want isolation mode to be deterministic, so we have to disallow all reads, even stdin.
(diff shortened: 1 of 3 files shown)


src/shims/unix/fd.rs
```diff
@@ -383,9 +390,9 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         // `File::read` never returns a value larger than `count`,
         // so this cannot fail.
         let result = file_descriptor
-            .read(communicate, &mut bytes, *this.tcx)?
+            .borrow_mut()
+            .read(communicate, &mut bytes, this)?
             .map(|c| i64::try_from(c).unwrap());
-        drop(file_descriptor);
```
RalfJung, line 388: I think its still good to keep the FD for as short as possible, so please keep the `drop` around.

src/shims/unix/fd.rs
```diff
@@ -421,14 +428,14 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         let communicate = this.machine.communicate();
 
         let bytes = this.read_bytes_ptr_strip_provenance(buf, Size::from_bytes(count))?.to_owned();
```
RalfJung, line 430: Cloning the bytes should not be needed any more now that we `dup` the FD, I think?

src/shims/unix/fd.rs
```diff
@@ -370,7 +377,7 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
             .min(u64::try_from(isize::MAX).unwrap());
         let communicate = this.machine.communicate();
 
-        let Some(mut file_descriptor) = this.machine.fds.get_mut(fd) else {
+        let Some(file_descriptor) = this.machine.fds.dup(fd) else {
```
RalfJung, line 380:
```suggestion
        // We temporarily dup the FD to be able to retain mutable access to `this`.
        let Some(file_descriptor) = this.machine.fds.dup(fd) else {
```

src/shims/unix/fd.rs
```diff
@@ -421,14 +428,14 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
         let communicate = this.machine.communicate();
 
         let bytes = this.read_bytes_ptr_strip_provenance(buf, Size::from_bytes(count))?.to_owned();
-        let Some(mut file_descriptor) = this.machine.fds.get_mut(fd) else {
+        let Some(file_descriptor) = this.machine.fds.dup(fd) else {
```
RalfJung, line 431:
```suggestion
        // We temporarily dup the FD to be able to retain mutable access to `this`.
        let Some(file_descriptor) = this.machine.fds.dup(fd) else {
```

RalfJung (commented):
Looks good overall, thanks! Just some nits.

src/shims/unix/fd.rs
Luv-Ray: We needs to call `write(communicate, &bytes, this)` later, seems that there is still a conflict if `bytes` holds a 
immutable borrow to `this` :(

src/shims/unix/fd.rs
Luv-Ray: That's right, fixed.

src/shims/unix/fd.rs
RalfJung: Ah, right.

RalfJung:
Looks good! Please squash the commits.

Luv-Ray:
poll :)
Is there anything left for me to do?

PR #3609, opened 2024-05-16
Add socketpair shim

### LINKED ISSUE #3442: Complete socketpair support
We pretend to implement `socketpair`, but that's a complete lie. Completing this however doesn't need any involvement with the rest of the epoll plans. We "just" need to allocate somewhere two buffers for the data to be sent between these two endpoints, and then wire up the read/write of the two sockets with the right buffers. This can be tested by directly using `socketpair`/`read`/`write`.

We can provide more detailed mentoring instructions if needed.

tiif:
Fixes #3442 
Design proposal: https://hackmd.io/@tiif/Skhc1t0-C

### DIFF
--- a/src/shims/unix/socket.rs
+++ b/src/shims/unix/socket.rs
@@ -1,15 +1,38 @@
+use std::cell::RefCell;
+use std::collections::VecDeque;
 use std::io;
+use std::io::{Error, ErrorKind, Read};
+use std::rc::{Rc, Weak};
 
 use crate::shims::unix::*;
-use crate::*;
+use crate::{concurrency::VClock, *};
 
 use self::fd::FileDescriptor;
 
+/// The maximum capacity of the socketpair buffer in bytes.
+/// This number is arbitrary as the value can always
+/// be configured in the real system.
+const MAX_SOCKETPAIR_BUFFER_CAPACITY: usize = 212992;
+
 /// Pair of connected sockets.
-///
-/// We currently don't allow sending any data through this pair, so this can be just a dummy.
 #[derive(Debug)]
-struct SocketPair;
+struct SocketPair {
+    // By making the write link weak, a `write` can detect when all readers are
+    // gone, and trigger EPIPE as appropriate.
+    writebuf: Weak<RefCell<Buffer>>,
+    readbuf: Rc<RefCell<Buffer>>,
+    is_nonblock: bool,
+}
+
+#[derive(Debug)]
+struct Buffer {
+    buf: VecDeque<u8>,
+    clock: VClock,
+    /// Indicates if there is at least one active writer to this buffer.
+    /// If all writers of this buffer are dropped, buf_has_writer becomes false and we
+    /// indicate EOF instead of blocking.
+    buf_has_writer: bool,
+}
 
 impl FileDescription for SocketPair {
     fn name(&self) -> &'static str {
@@ -20,17 +43,102 @@ impl FileDescription for SocketPair {
         self: Box<Self>,
         _communicate_allowed: bool,
     ) -> InterpResult<'tcx, io::Result<()>> {
+        // This is used to signal socketfd of other side that there is no writer to its readbuf.
+        // If the upgrade fails, there is no need to update as all read ends have been dropped.
+        if let Some(writebuf) = self.writebuf.upgrade() {
+            writebuf.borrow_mut().buf_has_writer = false;
+        };
         Ok(Ok(()))
+    }
+
+    fn read<'tcx>(
+        &mut self,
+        _communicate_allowed: bool,
+        bytes: &mut [u8],
+        ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, io::Result<usize>> {
+        let request_byte_size = bytes.len();
+        let mut readbuf = self.readbuf.borrow_mut();
+
+        // Always succeed on read size 0.
+        if request_byte_size == 0 {
+            return Ok(Ok(0));
+        }
+
+        if readbuf.buf.is_empty() {
+            if !readbuf.buf_has_writer {
+                // Socketpair with no writer and empty buffer.
+                // 0 bytes successfully read indicates end-of-file.
+                return Ok(Ok(0));
+            } else {
+                if self.is_nonblock {
+                    // Non-blocking socketpair with writer and empty buffer.
+                    // https://linux.die.net/man/2/read
+                    // EAGAIN or EWOULDBLOCK can be returned for socket,
+                    // POSIX.1-2001 allows either error to be returned for this case.
+                    // Since there is no ErrorKind for EAGAIN, WouldBlock is used.
+                    return Ok(Err(Error::from(ErrorKind::WouldBlock)));
+                } else {
+                    // Blocking socketpair with writer and empty buffer.
+                    // FIXME: blocking is currently not supported
+                    throw_unsup_format!("socketpair read: blocking isn't supported yet");
+                }
+            }
+        }
+
+        // Synchronize with all previous writes to this buffer.
+        // FIXME: this over-synchronizes; a more precise approach would be to
+        // only sync with the writes whose data we will read.
+        ecx.acquire_clock(&readbuf.clock);
+        // Do full read / partial read based on the space available.
+        // Conveniently, `read` exists on `VecDeque` and has exactly the desired behavior.
+        let actual_read_size = readbuf.buf.read(bytes).unwrap();
+        return Ok(Ok(actual_read_size));
+    }
+
+    fn write<'tcx>(
+        &mut self,
+        _communicate_allowed: bool,
+        bytes: &[u8],
+        ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, io::Result<usize>> {
+        let write_size = bytes.len();
+        // Always succeed on write size 0.
+        // ("If count is zero and fd refers to a file other than a regular file, the results are not specified.")
+        if write_size == 0 {
+            return Ok(Ok(0));
+        }
+
+        let Some(writebuf) = self.writebuf.upgrade() else {
+            // If the upgrade from Weak to Rc fails, it indicates that all read ends have been
+            // closed.
+            return Ok(Err(Error::from(ErrorKind::BrokenPipe)));
+        };
+        let mut writebuf = writebuf.borrow_mut();
+        let data_size = writebuf.buf.len();
+        let available_space = MAX_SOCKETPAIR_BUFFER_CAPACITY.checked_sub(data_size).unwrap();
+        if available_space == 0 {
+            if self.is_nonblock {
+                // Non-blocking socketpair with a full buffer.
+                return Ok(Err(Error::from(ErrorKind::WouldBlock)));
+            } else {
+                // Blocking socketpair with a full buffer.
+                throw_unsup_format!("socketpair write: blocking isn't supported yet");
+            }
+        }
+        // Remember this clock so `read` can synchronize with us.
+        if let Some(clock) = &ecx.release_clock() {
+            writebuf.clock.join(clock);
+        }
+        // Do full write / partial write based on the space available.
+        let actual_write_size = write_size.min(available_space);
+        writebuf.buf.extend(&bytes[..actual_write_size]);
+        return Ok(Ok(actual_write_size));
     }
 }
 
 impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
-    /// Currently this function this function is a stub. Eventually we need to
-    /// properly implement an FD type for sockets and have this function create
-    /// two sockets and associated FDs such that writing to one will produce
-    /// data that can be read from the other.
-    ///
     /// For more information on the arguments see the socketpair manpage:
     /// <https://linux.die.net/man/2/socketpair>
     fn socketpair(
@@ -42,17 +150,80 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     ) -> InterpResult<'tcx, Scalar> {
         let this = self.eval_context_mut();
 
-        let _domain = this.read_scalar(domain)?.to_i32()?;
-        let _type_ = this.read_scalar(type_)?.to_i32()?;
-        let _protocol = this.read_scalar(protocol)?.to_i32()?;
+        let domain = this.read_scalar(domain)?.to_i32()?;
+        let mut type_ = this.read_scalar(type_)?.to_i32()?;
+        let protocol = this.read_scalar(protocol)?.to_i32()?;
         let sv = this.deref_pointer(sv)?;
 
-        // FIXME: fail on unsupported inputs
+        let mut is_sock_nonblock = false;
+
+        // Parse and remove the type flags that we support. If type != 0 after removing,
+        // unsupported flags are used.
+        if type_ & this.eval_libc_i32("SOCK_STREAM") == this.eval_libc_i32("SOCK_STREAM") {
+            type_ &= !(this.eval_libc_i32("SOCK_STREAM"));
+        }
+
+        // SOCK_NONBLOCK only exists on Linux.
+        if this.tcx.sess.target.os == "linux" {
+            if type_ & this.eval_libc_i32("SOCK_NONBLOCK") == this.eval_libc_i32("SOCK_NONBLOCK") {
+                is_sock_nonblock = true;
+                type_ &= !(this.eval_libc_i32("SOCK_NONBLOCK"));
+            }
+            if type_ & this.eval_libc_i32("SOCK_CLOEXEC") == this.eval_libc_i32("SOCK_CLOEXEC") {
+                type_ &= !(this.eval_libc_i32("SOCK_CLOEXEC"));
+            }
+        }
+
+        // Fail on unsupported input.
+        // AF_UNIX and AF_LOCAL are synonyms, so we accept both in case
+        // their values differ.
+        if domain != this.eval_libc_i32("AF_UNIX") && domain != this.eval_libc_i32("AF_LOCAL") {
+            throw_unsup_format!(
+                "socketpair: Unsupported domain {:#x} is used, only AF_UNIX \
+                                 and AF_LOCAL are allowed",
+                domain
+            );
+        } else if type_ != 0 {
+            throw_unsup_format!(
+                "socketpair: Unsupported type {:#x} is used, only SOCK_STREAM, \
+                                 SOCK_CLOEXEC and SOCK_NONBLOCK are allowed",
+                type_
+            );
+        } else if protocol != 0 {
+            throw_unsup_format!(
+                "socketpair: Unsupported socket protocol {protocol} is used, \
+                                 only 0 is allowed",
+            );
+        }
+
+        let buffer1 = Rc::new(RefCell::new(Buffer {
+            buf: VecDeque::new(),
+            clock: VClock::default(),
+            buf_has_writer: true,
+        }));
+
+        let buffer2 = Rc::new(RefCell::new(Buffer {
+            buf: VecDeque::new(),
+            clock: VClock::default(),
+            buf_has_writer: true,
+        }));
+
+        let socketpair_0 = SocketPair {
+            writebuf: Rc::downgrade(&buffer1),
+            readbuf: Rc::clone(&buffer2),
+            is_nonblock: is_sock_nonblock,
+        };
+
+        let socketpair_1 = SocketPair {
+            writebuf: Rc::downgrade(&buffer2),
+            readbuf: Rc::clone(&buffer1),
+            is_nonblock: is_sock_nonblock,
+        };
 
         let fds = &mut this.machine.fds;
-        let sv0 = fds.insert_fd(FileDescriptor::new(SocketPair));
+        let sv0 = fds.insert_fd(FileDescriptor::new(socketpair_0));
         let sv0 = Scalar::try_from_int(sv0, sv.layout.size).unwrap();
-        let sv1 = fds.insert_fd(FileDescriptor::new(SocketPair));
+        let sv1 = fds.insert_fd(FileDescriptor::new(socketpair_1));
         let sv1 = Scalar::try_from_int(sv1, sv.layout.size).unwrap();
 
         this.write_scalar(sv0, &sv)?;
--- a/tests/fail-dep/libc/socketpair_read_blocking.rs
+++ b/tests/fail-dep/libc/socketpair_read_blocking.rs
@@ -1,0 +1,12 @@
+//@ignore-target-windows: no libc socketpair on Windows
+
+// This is temporarily here because blocking on fd is not supported yet.
+// When blocking is eventually supported, this will be moved to pass-dep/libc/libc-socketpair
+
+fn main() {
+    let mut fds = [-1, -1];
+    let _ = unsafe { libc::socketpair(libc::AF_UNIX, libc::SOCK_STREAM, 0, fds.as_mut_ptr()) };
+    // The read below will be blocked because the buffer is empty.
+    let mut buf: [u8; 3] = [0; 3];
+    let _res = unsafe { libc::read(fds[1], buf.as_mut_ptr().cast(), buf.len() as libc::size_t) }; //~ERROR: blocking isn't supported
+}
--- a/tests/pass-dep/libc/libc-socketpair.rs
+++ b/tests/pass-dep/libc/libc-socketpair.rs
@@ -1,0 +1,124 @@
+//@ignore-target-windows: No libc socketpair on Windows
+// test_race depends on a deterministic schedule.
+//@compile-flags: -Zmiri-preemption-rate=0
+use std::thread;
+fn main() {
+    test_socketpair();
+    test_socketpair_threaded();
+    test_race();
+}
+
+fn test_socketpair() {
+    let mut fds = [-1, -1];
+    let mut res =
+        unsafe { libc::socketpair(libc::AF_UNIX, libc::SOCK_STREAM, 0, fds.as_mut_ptr()) };
+    assert_eq!(res, 0);
+
+    // Read size == data available in buffer.
+    let data = "abcde".as_bytes().as_ptr();
+    res = unsafe { libc::write(fds[0], data as *const libc::c_void, 5).try_into().unwrap() };
+    assert_eq!(res, 5);
+    let mut buf: [u8; 5] = [0; 5];
+    res = unsafe {
+        libc::read(fds[1], buf.as_mut_ptr().cast(), buf.len() as libc::size_t).try_into().unwrap()
+    };
+    assert_eq!(res, 5);
+    assert_eq!(buf, "abcde".as_bytes());
+
+    // Read size > data available in buffer.
+    let data = "abc".as_bytes().as_ptr();
+    res = unsafe { libc::write(fds[0], data as *const libc::c_void, 3).try_into().unwrap() };
+    assert_eq!(res, 3);
+    let mut buf2: [u8; 5] = [0; 5];
+    res = unsafe {
+        libc::read(fds[1], buf2.as_mut_ptr().cast(), buf2.len() as libc::size_t).try_into().unwrap()
+    };
+    assert_eq!(res, 3);
+    assert_eq!(&buf2[0..3], "abc".as_bytes());
+
+    // Test read and write from another direction.
+    // Read size == data available in buffer.
+    let data = "12345".as_bytes().as_ptr();
+    res = unsafe { libc::write(fds[1], data as *const libc::c_void, 5).try_into().unwrap() };
+    assert_eq!(res, 5);
+    let mut buf3: [u8; 5] = [0; 5];
+    res = unsafe {
+        libc::read(fds[0], buf3.as_mut_ptr().cast(), buf3.len() as libc::size_t).try_into().unwrap()
+    };
+    assert_eq!(res, 5);
+    assert_eq!(buf3, "12345".as_bytes());
+
+    // Read size > data available in buffer.
+    let data = "123".as_bytes().as_ptr();
+    res = unsafe { libc::write(fds[1], data as *const libc::c_void, 3).try_into().unwrap() };
+    assert_eq!(res, 3);
+    let mut buf4: [u8; 5] = [0; 5];
+    res = unsafe {
+        libc::read(fds[0], buf4.as_mut_ptr().cast(), buf4.len() as libc::size_t).try_into().unwrap()
+    };
+    assert_eq!(res, 3);
+    assert_eq!(&buf4[0..3], "123".as_bytes());
+}
+
+fn test_socketpair_threaded() {
+    let mut fds = [-1, -1];
+    let mut res =
+        unsafe { libc::socketpair(libc::AF_UNIX, libc::SOCK_STREAM, 0, fds.as_mut_ptr()) };
+    assert_eq!(res, 0);
+
+    let data = "abcde".as_bytes().as_ptr();
+    res = unsafe { libc::write(fds[0], data as *const libc::c_void, 5).try_into().unwrap() };
+    assert_eq!(res, 5);
+    let thread1 = thread::spawn(move || {
+        let mut buf: [u8; 5] = [0; 5];
+        let res: i64 = unsafe {
+            libc::read(fds[1], buf.as_mut_ptr().cast(), buf.len() as libc::size_t)
+                .try_into()
+                .unwrap()
+        };
+        assert_eq!(res, 5);
+        assert_eq!(buf, "abcde".as_bytes());
+    });
+    thread1.join().unwrap();
+
+    // Read and write from different direction
+    let thread2 = thread::spawn(move || {
+        let data = "12345".as_bytes().as_ptr();
+        let res: i64 =
+            unsafe { libc::write(fds[0], data as *const libc::c_void, 5).try_into().unwrap() };
+        assert_eq!(res, 5);
+    });
+    thread2.join().unwrap();
+    let mut buf: [u8; 5] = [0; 5];
+    res = unsafe {
+        libc::read(fds[1], buf.as_mut_ptr().cast(), buf.len() as libc::size_t).try_into().unwrap()
+    };
+    assert_eq!(res, 5);
+    assert_eq!(buf, "12345".as_bytes());
+}
+fn test_race() {
+    static mut VAL: u8 = 0;
+    let mut fds = [-1, -1];
+    let mut res =
+        unsafe { libc::socketpair(libc::AF_UNIX, libc::SOCK_STREAM, 0, fds.as_mut_ptr()) };
+    assert_eq!(res, 0);
+    let thread1 = thread::spawn(move || {
+        let mut buf: [u8; 1] = [0; 1];
+        // write() from the main thread will occur before the read() here
+        // because preemption is disabled and the main thread yields after write().
+        let res: i32 = unsafe {
+            libc::read(fds[1], buf.as_mut_ptr().cast(), buf.len() as libc::size_t)
+                .try_into()
+                .unwrap()
+        };
+        assert_eq!(res, 1);
+        assert_eq!(buf, "a".as_bytes());
+        unsafe { assert_eq!(VAL, 1) };
+    });
+    unsafe { VAL = 1 };
+    let data = "a".as_bytes().as_ptr();
+    res = unsafe { libc::write(fds[0], data as *const libc::c_void, 1).try_into().unwrap() };
+    assert_eq!(res, 1);
+    thread::yield_now();
+    thread1.join().unwrap();
+}
--- a/tests/fail-dep/libc/socketpair_read_blocking.stderr
+++ b/tests/fail-dep/libc/socketpair_read_blocking.stderr
@@ -1,0 +1,14 @@
+error: unsupported operation: socketpair read: blocking isn't supported yet
+  --> $DIR/socketpair_read_blocking.rs:LL:CC
+   |
+LL |     let _res = unsafe { libc::read(fds[1], buf.as_mut_ptr().cast(), buf.len() as libc::size_t) };
+   |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ socketpair read: blocking isn't supported yet
+   |
+   = help: this is likely not a bug in the program; it indicates that the program performed an operation that Miri does not support
+   = note: BACKTRACE:
+   = note: inside `main` at $DIR/socketpair_read_blocking.rs:LL:CC
+
+note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace
+
+error: aborting due to 1 previous error
+
--- a/tests/fail-dep/libc/socketpair_write_blocking.rs
+++ b/tests/fail-dep/libc/socketpair_write_blocking.rs
@@ -1,0 +1,16 @@
+//@ignore-target-windows: no libc socketpair on Windows
+// This is temporarily here because blocking on fd is not supported yet.
+// When blocking is eventually supported, this will be moved to pass-dep/libc/libc-socketpair
+fn main() {
+    let mut fds = [-1, -1];
+    let _ = unsafe { libc::socketpair(libc::AF_UNIX, libc::SOCK_STREAM, 0, fds.as_mut_ptr()) };
+    // Write size > buffer capacity
+    // Used up all the space in the buffer.
+    let arr1: [u8; 212992] = [1; 212992];
+    let _ = unsafe { libc::write(fds[0], arr1.as_ptr() as *const libc::c_void, 212992) };
+    let data = "abc".as_bytes().as_ptr();
+    // The write below will be blocked as the buffer is full.
+    let _ = unsafe { libc::write(fds[0], data as *const libc::c_void, 3) }; //~ERROR: blocking isn't supported
+    let mut buf: [u8; 3] = [0; 3];
+    let _res = unsafe { libc::read(fds[1], buf.as_mut_ptr().cast(), buf.len() as libc::size_t) };
+}
--- a/tests/fail-dep/libc/socketpair_write_blocking.stderr
+++ b/tests/fail-dep/libc/socketpair_write_blocking.stderr
@@ -1,0 +1,14 @@
+error: unsupported operation: socketpair write: blocking isn't supported yet
+  --> $DIR/socketpair_write_blocking.rs:LL:CC
+   |
+LL |     let _ = unsafe { libc::write(fds[0], data as *const libc::c_void, 3) };
+   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ socketpair write: blocking isn't supported yet
+   |
+   = help: this is likely not a bug in the program; it indicates that the program performed an operation that Miri does not support
+   = note: BACKTRACE:
+   = note: inside `main` at $DIR/socketpair_write_blocking.rs:LL:CC
+
+note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace
+
+error: aborting due to 1 previous error
+


tests/pass-dep/libc/libc-socketpair.rs
```diff
@@ -0,0 +1,120 @@
...
+    let _ = thread1.join();
+    _ = thread2.join();
+    let mut buf: [u8; 200000] = [2; 200000];
+    _ = unsafe { libc::read(fds[1], buf.as_mut_ptr().cast(), buf.len() as libc::size_t) };
+    for byte in buf.iter().take(100000) {
+        if *byte == b'0' {
+            println!("race happened");
+        }
+    }
+}
```
tiif, line 120: I suspect a data race can happen since there is no synchronization for the socketpair buffer yet, but currently, data race is not detected in this test. @RalfJung please let me know your thoughts on this. :)

tests/pass-dep/libc/libc-socketpair.rs
```diff
@@ -0,0 +1,121 @@
...
+    _ = thread2.join();
+    let mut buf: [u8; 200000] = [2; 200000];
+    _ = unsafe { libc::read(fds[1], buf.as_mut_ptr().cast(), buf.len() as libc::size_t) };
+    for byte in buf.iter().take(100000) {
+        if *byte != 0 {
+            // The first 100000 bytes should all be 0.
+            println!("race happened");
+        }
+    }
+}
```
tiif, line 121: Currently there is no synchronization on the socketpair buffers yet. I suspect a data race can happen, but it is not detected in this test. @RalfJung please let me know your thoughts on this. :)

tiif: When a race happened, a data pattern of 00100 might appear in the buffer because the write of thread2 started before the write of thread1 finished.

tiif:
This is not entirely ready, and still waiting for discussion to happen, but I will tag it with ``waiting for review`` for more visibility. 

@rustbot ready

src/shims/unix/socket.rs
```diff
@@ -1,15 +1,22 @@
...
+use rustc_middle::ty::TyCtxt;
+
 /// Pair of connected sockets.
-///
-/// We currently don't allow sending any data through this pair, so this can be just a dummy.
 #[derive(Debug)]
-struct SocketPair;
+struct SocketPair {
+    readbuf: Rc<RefCell<VecDeque<u8>>>,
+    writebuf: Rc<RefCell<VecDeque<u8>>>,
```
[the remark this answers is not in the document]
oli-obk: no, miri is entirely single threaded, even if it emulates multithreading

src/shims/unix/socket.rs
```diff
@@ -42,17 +105,40 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
-        // FIXME: fail on unsupported inputs
+        // Fail on unsupported input
+        if domain != this.eval_libc_i32("AF_UNIX") && domain != this.eval_libc_i32("AF_LOCAL") {
+            throw_unsup_format!("socketpair: Only AF_UNIX and AF_LOCAL are supported");
+        } else if type_ != this.eval_libc_i32("SOCK_STREAM")
+            && type_ != this.eval_libc_i32("SOCK_DGRAM")
+        {
+            throw_unsup_format!("socketpair: Only SOCK_STREAM and SOCK_DGRAM are suppported");
+        } else if protocol != 0 {
+            throw_unsup_format!("socketpair: Unsupported socket protocol is used");
```
oli-obk, line 121: report the protocol id in the message (same for `domain` and `type_`)

src/shims/unix/socket.rs
```diff
@@ -42,17 +105,40 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+        }
+
+        let buf1 = Rc::new(RefCell::new(VecDeque::with_capacity(3)));
+        let buf2 = Rc::new(RefCell::new(VecDeque::with_capacity(3)));
+
+        let socketpair_0 = SocketPair {
+            writebuf: Rc::clone(&buf1),
+            readbuf: Rc::clone(&buf2),
+            max_capacity: 212992,
```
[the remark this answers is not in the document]
oli-obk: iirc we have similar magic numbers (e.g. page sizes) in miri elswhere, check how they are defined (probably with named constants and some documentation).

src/shims/unix/socket.rs
```diff
@@ -22,15 +29,71 @@ impl FileDescription for SocketPair {
...
+        let write_size = bytes.len();
+
+        // If there is no space left in the buffer, we will block until there is available space.
+        // FIXME: blocking is currently not supported.
+        if available_space == 0 {
+            throw_unsup_format!("socketpair write: blocking isn't supported yet");
+        }
+        if write_size == 0 {
+            return Ok(Ok(0));
+        }
```
oli-obk, line 79: shouldn't these two checks be swapped? writing zero bytes into a full buffer is fine

src/shims/unix/socket.rs
```diff
@@ -22,15 +29,71 @@ impl FileDescription for SocketPair {
...
+            for byte in bytes.iter().take(available_space) {
+                self.writebuf.borrow_mut().push_back(*byte);
+            }
+            return Ok(Ok(available_space));
+        } else {
+            for byte in bytes.iter() {
+                self.writebuf.borrow_mut().push_back(*byte);
+            }
+            return Ok(Ok(write_size));
+        }
```
oli-obk, line 91: instead of having two code paths, compute the actual write size (`actual_write_size = write_size.min(available_space)`), and then unconditionally do something like `self.write_buf.borrow_mut().extend(&bytes[..actual_write_size])`

src/shims/unix/socket.rs
```diff
@@ -22,15 +29,71 @@ impl FileDescription for SocketPair {
...
+        bytes: &mut [u8],
+        _tcx: TyCtxt<'tcx>,
+    ) -> InterpResult<'tcx, io::Result<usize>> {
+        let request_byte_size = bytes.len();
+        let data_size = self.readbuf.borrow_mut().len();
+
+        // FIXME: blocking is currently not supported
+        if data_size == 0 {
+            throw_unsup_format!("socketpair read: blocking isn't supported yet");
+        }
```
oli-obk, line 45: as Ralf mentioned in the corresponding issue

> store during socket creation whether the socket is blocking, and then in read/write, if we would have to block, throw_unsup_format!.

This should only bail out as unsupported for blocking sockets.

src/shims/unix/socket.rs
```diff
@@ -22,15 +29,71 @@ impl FileDescription for SocketPair {
...
+        if request_byte_size == 0 {
+            return Ok(Ok(0));
+        }
```
oli-obk, line 48: do this before the available bytes check, because reading zero bytes is always fine.

src/shims/unix/socket.rs
```diff
@@ -22,15 +29,71 @@ impl FileDescription for SocketPair {
...
+            for byte in bytes.iter_mut().take(data_size) {
+                *byte = self.readbuf.borrow_mut().pop_front().unwrap();
+            }
+            return Ok(Ok(data_size));
+        } else {
+            for byte in bytes.iter_mut() {
+                *byte = self.readbuf.borrow_mut().pop_front().unwrap();
+            }
+            return Ok(Ok(request_byte_size));
+        }
```
oli-obk, line 60: do something similar as proposed in `write`, branchless writing of the slice

oli-obk (changes requested):
This will also need to handle the case where the other socket was dropped. What do the manpages say about that case? Will we need to report https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.ConnectionAborted or something?

tests/pass-dep/libc/libc-socketpair.rs
oli-obk: Yea, our write/read operations are atomic if they succeed fully. So if the returned value from `write` is `100000`, then all of the bytes were written, and can't have been interleaved with the other thread's written data.

The only thing that could happen is that the first 100k bytes are `1`, because it's not guaranteed which thread writes first. But there will be no interleaving.

What is the expected behaviour on a real system? Is this going to race? If so, we'll need to teach our data race detector about file descriptors.

src/shims/unix/socket.rs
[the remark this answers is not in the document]
RalfJung: We don't have to exactly match any particular kernel here as this is a system configuration knob. But yeah magic values should definitely live in constants.

src/shims/unix/socket.rs
```diff
@@ -1,15 +1,22 @@
...
+
 /// Pair of connected sockets.
-///
-/// We currently don't allow sending any data through this pair, so this can be just a dummy.
 #[derive(Debug)]
-struct SocketPair;
+struct SocketPair {
+    readbuf: Rc<RefCell<VecDeque<u8>>>,
+    writebuf: Rc<RefCell<VecDeque<u8>>>,
+    max_capacity: usize,
```
RalfJung, line 18: Isn't max_capacity a property of the buffer, rather than the socket pair?

Currently it is anyway a constant so there's little point in carrying it here as a field, is there?

src/shims/unix/socket.rs
```diff
@@ -22,15 +29,71 @@ impl FileDescription for SocketPair {
...
     }
+
+    fn read<'tcx>(
+        &mut self,
+        _communicate_allowed: bool,
+        bytes: &mut [u8],
+        _tcx: TyCtxt<'tcx>,
+    ) -> InterpResult<'tcx, io::Result<usize>> {
+        let request_byte_size = bytes.len();
+        let data_size = self.readbuf.borrow_mut().len();
```
RalfJung, line 40: Please only call `borrow_mut` once in this function.

src/shims/unix/socket.rs
```diff
@@ -42,17 +105,40 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
         let sv = this.deref_pointer(sv)?;
 
-        // FIXME: fail on unsupported inputs
+        // Fail on unsupported input
+        if domain != this.eval_libc_i32("AF_UNIX") && domain != this.eval_libc_i32("AF_LOCAL") {
+            throw_unsup_format!("socketpair: Only AF_UNIX and AF_LOCAL are supported");
+        } else if type_ != this.eval_libc_i32("SOCK_STREAM")
+            && type_ != this.eval_libc_i32("SOCK_DGRAM")
+        {
+            throw_unsup_format!("socketpair: Only SOCK_STREAM and SOCK_DGRAM are suppported");
```
RalfJung, line 119: We don't support DGRAM, so I think that one should be removed.

src/shims/unix/socket.rs
```diff
@@ -42,17 +105,40 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
-        let _type_ = this.read_scalar(type_)?.to_i32()?;
-        let _protocol = this.read_scalar(protocol)?.to_i32()?;
+        let domain = this.read_scalar(domain)?.to_i32()?;
+        let type_ = this.read_scalar(type_)?.to_i32()?;
+        let protocol = this.read_scalar(protocol)?.to_i32()?;
         let sv = this.deref_pointer(sv)?;
 
-        // FIXME: fail on unsupported inputs
+        // Fail on unsupported input
+        if domain != this.eval_libc_i32("AF_UNIX") && domain != this.eval_libc_i32("AF_LOCAL") {
```
RalfJung, line 114:
```suggestion
        // AF_UNIX and AF_LOCAL are synonyms, so we accept both in case
        // their values differ.
        if domain != this.eval_libc_i32("AF_UNIX") && domain != this.eval_libc_i32("AF_LOCAL") {
```

src/shims/unix/socket.rs
```diff
@@ -42,17 +105,40 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+            throw_unsup_format!("socketpair: Only AF_UNIX and AF_LOCAL are supported");
+        } else if type_ != this.eval_libc_i32("SOCK_STREAM")
+            && type_ != this.eval_libc_i32("SOCK_DGRAM")
+        {
+            throw_unsup_format!("socketpair: Only SOCK_STREAM and SOCK_DGRAM are suppported");
+        } else if protocol != 0 {
+            throw_unsup_format!("socketpair: Unsupported socket protocol is used");
+        }
+
+        let buf1 = Rc::new(RefCell::new(VecDeque::with_capacity(3)));
```
RalfJung, line 124: Why `3`?

src/shims/unix/socket.rs
RalfJung: Currently all sockets are blocking as we're rejecting `type` set to `... | SOCK_NONBLOCK`.

However, @tiif said elsewhere tokio needs non-blocking sockets, and yet the tokio tests still pass despite non-blocking sockets now being rejected, so something is odd.

tests/pass-dep/libc/libc-socketpair.rs
```diff
@@ -0,0 +1,121 @@
...
+    };
+    assert_eq!(res, 5);
+    assert_eq!(buf, "12345".as_bytes());
+}
+fn test_write_race() {
+    use std::thread;
+    let mut fds = [-1, -1];
+    let _res = unsafe { libc::socketpair(libc::AF_UNIX, libc::SOCK_STREAM, 0, fds.as_mut_ptr()) };
+    let thread1 = thread::spawn(move || {
+        let arr0: [u8; 100000] = [0; 100000];
```
RalfJung, line 104: These are very big buffers, and Miri is kind of slow... how long do these tests take?

tests/pass-dep/libc/libc-socketpair.rs
```diff
@@ -0,0 +1,121 @@
...
+        let _ = unsafe { libc::write(fds[0], arr0.as_ptr() as *const c_void, 100000) };
+    });
+    let thread2 = thread::spawn(move || {
+        let arr1: [u8; 100000] = [1; 100000];
+        let _ = unsafe { libc::write(fds[0], arr1.as_ptr() as *const c_void, 100000) };
+    });
+    let _ = thread1.join();
+    _ = thread2.join();
+    let mut buf: [u8; 200000] = [2; 200000];
+    _ = unsafe { libc::read(fds[1], buf.as_mut_ptr().cast(), buf.len() as libc::size_t) };
```
RalfJung, line 114: You should never ignore the return value of the read, it tells you how much has actually been read!

tests/pass-dep/libc/libc-socketpair.rs
```diff
@@ -0,0 +1,121 @@
...
+    let thread1 = thread::spawn(move || {
+        let arr0: [u8; 100000] = [0; 100000];
+        let _ = unsafe { libc::write(fds[0], arr0.as_ptr() as *const c_void, 100000) };
+    });
+    let thread2 = thread::spawn(move || {
+        let arr1: [u8; 100000] = [1; 100000];
+        let _ = unsafe { libc::write(fds[0], arr1.as_ptr() as *const c_void, 100000) };
+    });
+    let _ = thread1.join();
+    _ = thread2.join();
```
RalfJung, line 112:
```suggestion
    thread2.join().unwrap();
```
same for the other `join` calls

tests/pass-dep/libc/libc-socketpair.rs
RalfJung: I think `write` calls are atomic in real systems as well.

However, for this test, can't the second thread go first so we get the 1s before the 0s?

RalfJung (commented):
Thanks! I think the general strategy works, I left some comments to complement Oli's feedback.

There is one issue we need to still consider: the interaction with data race detection. Since this can be used to send data from one thread to another, arguably a `write` to a buffer should be considered a "release" operation, and a `read` should be considered an "acquire". The way to test for this is to have a test with `-Zmiri-preemption-rate=0` and then
- have a global `static mut VAL: u8 = 0;`
- in main, create a socketpair and then spawn a thread
- in the main thread, go on by setting `VAL` to 1 and then sending a byte onto one socket. then call `thread::yield_now`.
- in the other thread, read one byte from the socket, then `assert_eq!(VAL, 1)`.

In your PR as-is, this should report UB due to a data race. However, arguably there is no data race here, since the other thread waited to receive a signal from the main thread.

To fix that, you should move the `VecDeque<u8>` into a new type, `Buffer`. This buffer needs a second field `clock: VClock`.
- When doing a `write`, you need to do something like `clock.join(data_race.release_clock(ecx.get_active_thread(), ecx.machine.current_span()))`, where `data_race` is in `ecx.machine.data_race` but it can be `None` (in which case you don't need to do anything).
- When doing a read, you need to `data_race.acquire_clock(&buffer.clock, ecx.get_active_thread())`.

(We should really have some nicer wrapper functions for this...)

tests/pass-dep/libc/libc-socketpair.rs
tiif: > But there will be no interleaving.

Yes, this is good enough. 

> I think write calls are atomic in real systems as well.

I haven't found the specification for ``write`` on ``socketpair``, but at least for ``pipes`` or ``FIFO``, ``write`` is only guranteed to be atomic when the ``bytes written`` <= ``buffer size``

> can't the second thread go first so we get the 1s before the 0s?

Yes, it is still correct even if second thread goes first. It is fine as long as the ``write`` of both threads are not interleaved (not having the ``00100`` pattern)

src/shims/unix/socket.rs
tiif: Oops, that's a typo. I intended to put ``with_capacity(212992)``, which is the magic number that is going to be a named constant soon.

tests/pass-dep/libc/libc-socketpair.rs
RalfJung: The test checks that the first 10k bytes are 0, doesn't it? So if the 2nd thread goes first it will fail? I must be missing something.

src/shims/unix/socket.rs
RalfJung: Should we really consume 400k RAM for each `socketpair`?
Does the kernel pre-allocate the entire buffer or grow it up to the given size as needed?

tests/pass-dep/libc/libc-socketpair.rs
oli-obk: yea, the test will fail under some seeds and preemption rates

tiif: Second thread goes first is a correct behaviour to have in a real system, but I didn't consider it in this test. Don't worry, I wrote this test is just to discuss if interleaving write is possible, I didn't plan to keep this test.

src/shims/unix/socket.rs
RalfJung: Oh strange, we can remove the `socketpair` shim entirely and our test suite still passes. Looks like that shim got added without *any* test coverage. That would explain the confusion.

src/shims/unix/socket.rs
oli-obk: We can safely just use an empty VecDeque. It will grow as it needs more size, and we have a hard limit on the 212992 due to the separate check for the length.

src/shims/unix/socket.rs
tiif: I think to actually invoke the ``libc::socketpair`` from tokio, we need to use ``UnixStream::pair``, This uses the wrapper function from ``mio`` that calls ``libc::socketpair``. I can add this to the test. 

```rust
use tokio::net::UnixStream;

#[tokio::main]
async fn main() {
    let _a = UnixStream::pair();
}
```
According to strace, the test above invoked 
```
socketpair(AF_UNIX, SOCK_STREAM|SOCK_CLOEXEC|SOCK_NONBLOCK, 0, [6, 7]) = 0
socketpair(AF_UNIX, SOCK_STREAM|SOCK_CLOEXEC|SOCK_NONBLOCK, 0, [9, 10]) = 0
```

But strace also says the test below invoked ``socketpair`` too, so I am not sure why ``socketpair`` has no test coverage
```rust
#[tokio::main]
async fn main() {
}
```
strace output for the test above:
```
socketpair(AF_UNIX, SOCK_STREAM|SOCK_CLOEXEC|SOCK_NONBLOCK, 0, [6, 7]) = 0

```

RalfJung: You're adding tests that directly call `socketpair`, that's enough. No need to worry about also hitting this via tokio (yet).

I would assume tokio uses socketpair somewhere in its runtime internally, and that's why it came up... but maybe it doesn't always do that, or so?

src/shims/unix/socket.rs
tiif: Yup you are right, it should be a constant for buffer instead of a field.

tests/pass-dep/libc/libc-socketpair.rs
tiif: I am going to remove this test soon, but if you are curious, ``test_write_race`` takes roughly 11s :)

src/shims/unix/socket.rs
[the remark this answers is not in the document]
tiif: It seems like 64-bits Linux uses 212992 where 32-bits Linux uses 163840[^1]. Does it matter if the buffer size is too big or too small for the host system? 

[^1]: https://unix.stackexchange.com/questions/424380/what-values-may-linux-use-for-the-default-unix-socket-buffer-size

src/shims/unix/socket.rs
tiif: Thanks, I really like this suggestion! I have been thinking how to make this ``if..else`` chunk better but couldn't come up with anything.

src/shims/unix/socket.rs
oli-obk: yea, it doesn't always invoke that. I think it depends on thread timings.

tiif:
Hmm, it seems like I got into some rebase trouble.

src/shims/unix/socket.rs
tiif: I wonder if we should report the exact flag name, because when I did the test 
```rust
    let flag = libc::SOCK_STREAM | libc::SOCK_NONBLOCK | libc::SOCK_CLOEXEC;
    let mut fds = [-1, -1];
    let mut res = unsafe { libc::socketpair(libc::AF_UNIX, flag, 0, fds.as_mut_ptr()) };
```
and the error shown is not very descriptive
```bash
error: unsupported operation: socketpair: Unsupported type 526337 is used
   --> ./tests/pass-dep/libc/libc-socketpair.rs:104:28
    |
104 | ... { libc::socketpair(libc::AF_UNIX, flag, 0, fds.as_mut_ptr()) };
    |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ socketpair: Unsupported type 526337 is used
```

tiif: For ``socketpair``, it is still manageable to parse the flags and report corresponding unsup error for each flag, but for other bigger functions like ``epoll`` that can have up to ``12`` flags, I am not sure if it is still a good idea to do this.

oli-obk: I think it's ok to just report the integer. It can be looked up manually. If there's an easy way to describe the value, we can use that, but otherwise any value is better than none in my book

saethlin: In my experience, it is very very rare to be familiar with the integer values, because they are not specified. People learn the names because that is what is portable.

In an effort to be helpful, can we mention _is_ allowed?

RalfJung: It helps a bit to print the integer value in hex, since there's usually bitwise combination of flags involved.

But decoding the flag and showing it with the names is quite a bit of work and requires knowing the names and values of all the relevant bits. If we want to do that I suggest it should be a helper function we can use in multiple places, and a separate PR.

src/shims/unix/socket.rs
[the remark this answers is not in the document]
RalfJung: As I said, no the exact value doesn't really matter.

RalfJung:
@rustbot author
(at least I think that reflects the current state)

src/shims/unix/socket.rs
RalfJung: Turns out we did have test coverage but I accidentally removed it... oopsie.
Will be fixed by https://github.com/rust-lang/miri/pull/3628.

src/shims/unix/socket.rs
```diff
@@ -1,15 +1,24 @@
...
 /// Pair of connected sockets.
-///
-/// We currently don't allow sending any data through this pair, so this can be just a dummy.
 #[derive(Debug)]
-struct SocketPair;
+struct SocketPair {
+    readbuf: Rc<RefCell<VecDeque<u8>>>,
+    writebuf: Rc<RefCell<VecDeque<u8>>>,
+    is_nonblock: bool,
+    _is_cloexec: bool,
```
RalfJung, line 20: There's no need for this field since we don't support `exec`. Elsewhere in Miri we discard CLOEXEC flags entirely so I think we should do the same here.

src/shims/unix/socket.rs
tiif: Combining all the ideas, we can do print value in hex + mention what is allowed, for example:

```rust
throw_unsup_format!("socketpair: type {:x} is not supported, only SOCK_STREAM, SOCK_CLOEXEC and SOCK_NONBLOCK are allowed", _type);
```

It's a bit lengthy, but this is the longest that it can get for ``socketpair``

src/shims/unix/socket.rs
```diff
@@ -22,15 +31,62 @@ impl FileDescription for SocketPair {
     ) -> InterpResult<'tcx, io::Result<()>> {
         Ok(Ok(()))
     }
+
+    fn read<'tcx>(
```
RalfJung, line 35: Don't we have to indicate EOF at some point in `read`? Like, when the other side got closed and all data has been read, or so?

src/shims/unix/socket.rs
RalfJung: Sounds good, except please add the leading `0x` to the hex output. (`{:#x}` will do that).

tiif:
It turns out that ``x86_64-apple-darwin`` does not support ``SOCK_NONBLOCK`` and ``SOCK_CLOEXEC``. Details is posted here https://github.com/rust-lang/miri/issues/3442#issuecomment-2130530963

But we have quite a few tasks that should be handled first, maybe we can delay the support ``x86_64-apple-darwin`` to another PR?

src/shims/unix/socket.rs
```diff
@@ -22,15 +31,62 @@ impl FileDescription for SocketPair {
...
+        // TODO: add support for clooexec and nonblock
+        if data_size == 0 && !self.is_nonblock {
+            throw_unsup_format!("socketpair read: blocking isn't supported yet");
+        }
+        // If the current data size in buffer < size requested, we only read up to data_size in buffer.
+        // Else, we will read the size requested.
+        let actual_read_size = request_byte_size.min(data_size);
+        for byte in bytes.iter_mut().take(actual_read_size) {
+            *byte = self.readbuf.borrow_mut().pop_front().unwrap();
+        }
```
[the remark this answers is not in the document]
tiif: Thanks, this is another neat rust tip that I didn't know before! :)

tiif:
> Don't we have to indicate EOF at some point in read? Like, when the other side got closed and all data has been read, or so?

> This will also need to handle the case where the other socket was dropped. What do the manpages say about that case? Will we need to report https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.ConnectionAborted or something?

Yes we need special handling for these cases[^1], this is the new specification: 
### For non-blocking 
- read
    - if write end open but no data available 
        - fail with ``EAGAIN``
    - if write end close
        - return 0 (because ``EOF``)
    - if read size < available data
        - read as requested and return
    - if read size  >= available data
        - read as much as we can and return (partial read)
- write
    - if read end open
        - if write size <= buffer size
            - if there is available space to write **all** requested bytes
                - write and return 
            - else (no available space to write **all** requested bytes)
                - fail with ``EAGAIN``
        - if write size > buffer size
            - if available space in buffer is 0
                - fail with ``EAGAIN``
            - else (available space in buffer != 0)
                - write as much as we can (partial write)
    - if read end closed
        - fail with ``EPIPE``
### For blocking
- read
    - if write end open but no data available 
        - block 
    - if write end close
       - return 0 (because ``EOF``)
    - if read size < available data
        - read as requested and return
    - if read size  >= available data
        - read as much as we can and return 
- write
    - if read end open
        - if write size <= buffer size
            - block until there is sufficient space
        - if write size > buffer size
            - block until there is sufficient space
    - if read end close
        - fail with ``EPIPE``

[^1]: https://man7.org/linux/man-pages/man7/pipe.7.html ``PIPE_BUF`` section, I can't find exact specification for sockets, I assumed behaviour of NONBLOCKING file descriptors (sockets/pipe/FIFO) are the same here

tiif:
We also need to implement ``close``.


We can add in a new field that is being shared between 2 socketpair struct:
```rust
struct SocketPair {
    readbuf: Rc<RefCell<VecDeque<u8>>>,
    writebuf: Rc<RefCell<VecDeque<u8>>>,
    is_nonblock: bool,
    both_end_open: Rc<RefCell<bool>>,
}
```
This is the initial state where both side is opened,
```
write to fd[0]  -----> write to buffer0 ----> read from fd[1] 
read from fd[0] <----- write to buffer1 <---- write to fd[1]
```
When ``fd[1]`` is closed, this will be the state of the socketpair:
```
write to fd[0]  -----> write to buffer0 ----> X
read from fd[0] <----- write to buffer1 <---- X
```
``fd[1]`` should set ``both_end_open`` to ``false`` to notify ``fd[0]`` that the other end has been dropped.

oli-obk:
How is nonblocking EOF differentiated from no available bytes?



> ```rust
> RefCell<bool>
> ```

Prefer `Cell` when the wrapped type is simple and Copy like bool

oli-obk:
Blocking write/read will require some refactorings to allow us to park the thread and continue later. We already have the concept for `sleep`, something similar will be needed here. I would recommend doing so in a separate PR and just keep emitting unsupported where it would block

tiif:
> How is nonblocking EOF differentiated from no available bytes?

I am not sure if I fully understand your question, please clarify if needed. For nonblocking, ``return 0`` is used to indicate ``EOF`` when the write end is closed. In this case, there is no input, and there won't be any input in future. For write end opened but no available bytes, ``EAGAIN`` is thrown to express "there is no available bytes right now, please try again later". (note: I didn't come out with this specification, I only transcribed what I read)

oli-obk:
Oh yea that makes sense. I confused myself with the "bytes available" bullet points

tiif:
> What we could do instead is to use Weak instead of Rc for the writer. Once the corresponding reader is dropped, the weak cannot be upgraded to an rc anymore, signalling that the other end is gone.

Ah, that's neat! Since not being able to upgrade to ``Rc`` means the other side is dropped, we only need 
```rust
struct SocketPair {
    readbuf: Weak<RefCell<VecDeque<u8>>>,
    writebuf: Weak<RefCell<VecDeque<u8>>>,
    is_nonblock: bool,
}
```

oli-obk:
readbuf needs to be Rc, not Weak, one side always needs to be Rc, otherwise both Weaks are dangling and can't be upgraded. When thr last Rc gets dropped, all weak become invalid

tiif:
> readbuf needs to be Rc, not Weak, one side always needs to be Rc, otherwise both Weaks are dangling and can't be upgraded.

I see. 

Let's say ``fd[1]`` is dropped, 
```
(type: Weak) write to fd[0]  -----> write to buffer0 ----> X
(type: Rc) read from fd[0] <----- write to buffer1 <----  X <-- consider this direction
```
we could use ``Rc::weak_count`` to track if the write end is dropped?

oli-obk:
> we could use `Rc::weak_count` to track if the write end is dropped?

Yea, that would work

RalfJung:
@tiif your list above is not entirely right -- when all writers are closed but there is data left in the buffer, that data should still be returned. Only *then* should you indicate EOF.

RalfJung:
Checking the `count`s is suspicious, those are pretty hard to use as a signal for anything... if we ever add some other reference to this buffer then suddenly this logic will break.

Maybe it's better to just have our own `num_writers` counter that we track explicitly?

tiif:
> when all writers are closed but there is data left in the buffer, that data should still be returned. 

Yes, I didn't consider this case. The list is now updated. 

> Maybe it's better to just have our own has_writer flag that we track explicitly? close on a Socketpair can then set has_writer on its write buffer to false. (close will only be called when the last FD for this file description is closed.)

I have been thinking about this for a whlie, but not sure how this is going to work yet. 

Initial state, ``fd[0]`` is the writer of its writebuf, and ``fd[1]`` is the writer of ``fd[0]``'s readbuf. 
```
write to fd[0]  -----> write to buffer0 ----> read from fd[1] 
read from fd[0] <----- write to buffer1 <---- write to fd[1]
```
If we ``dup fd[0] ``,  and only consider ``buffer0``
```
write to fd[0]  ----->    |----------------------|
                          | buffer0              |----> read from fd[1] 
write to newfd  ----->    |----------------------|
```
Closing ``fd[0]`` is a valid action and it will just make the state transit to:
```
write to newfd  -----> write to buffer0 ----> read from fd[1] 
```
So I am not sure how ``close`` can only be called when the last FD for this file description is closed.

I am thinking of letting the ``socketpair`` having refcount for writers of its ``writebuf`` and ``readbuf``. 
```rust
struct SocketPair {
    readbuf: Rc<RefCell<VecDeque<u8>>>,
    writebuf: Weak<RefCell<VecDeque<u8>>>,
    is_nonblock: bool,
    // number of writer for readbuf
    readbufWriterCount: Rc<Cell<i64>>,
    // number of writer for writebuf
    writebufWriterCount: Rc<Cell<i64>>
}
```
```
write to fd[0]  ----->    |----------------------|
                          | buffer0              |----> read from fd[1] 
write to newfd  ----->    |----------------------|
```

For this case, ``writebufWriterCount`` of ``fd[0]`` and ``newfd`` hold the same reference as ``readbufWriterCount`` of ``fd[1]`` 
When ``fd[0]`` is  ``dup`` to ``newfd``, ``writebufWriterCount`` of ``fd[0]`` and ``newfd`` will increase by 1 (So ``readbufWriterCount`` of ``fd[1]`` will increase by 1 too). To check if all the write end of the ``readbuf`` is dropped, we only need to check if the value of ``readbufWriterCount`` is 0. 

Since these counters is only used to track writer of buffers, it won't break when the buffer is referred for other purpose. 

Not sure if I over-complicated this problem, I feel a more elegant solution exists somewhere.

RalfJung:
> So I am not sure how close can only be called when the last FD for this file description is closed.

This is already the case, the general file descriptor machinery takes care of that.

`close` on a file *description* is called exactly once, and each buffer is a write buffer for exactly one file *description*, so you can use the closing of that file description as the indication for "all writers are gone".

tiif:
> close on a file description is called exactly once (even if, using dup, there are many file descriptors for the same file description), and each buffer is a write buffer for exactly one file description, so you can use the closing of that file description as the indication for "all writers are gone".

Ah now I see how it works now, yes this makes sense. I confused ``close`` for file descriptor and file description.

tests/pass-dep/libc/libc-socketpair.rs
```diff
@@ -1,4 +1,5 @@
 //@ignore-target-windows: No libc socketpair on Windows
+//@compile-flags: -Zmiri-preemption-rate=0
```
RalfJung, line 2: Why would this test disable preemption...?

RalfJung: Oh, it contains the race test as well.
```suggestion
// test_race depends on a deterministic schedule.
//@compile-flags: -Zmiri-preemption-rate=0
```

src/shims/unix/socket.rs
```diff
@@ -20,17 +38,122 @@ impl FileDescription for SocketPair {
...
+                throw_unsup_format!("socketpair read: blocking isn't supported yet");
+            }
+        }
+        if let Some(data_race) = &ecx.machine.data_race {
+            data_race.acquire_clock(&self.readbuf.clone().borrow_mut().clock, &ecx.machine.threads);
+        }
+        // Do full read / partial read based on the space available.
+        let actual_read_size = request_byte_size.min(data_size);
+        self.readbuf.borrow_mut().buf.read_exact(&mut bytes[..actual_read_size]).unwrap();
+        return Ok(Ok(actual_read_size));
```
oli-obk, line 96: I wonder if we can delegate the size check to the `VecDeque`

```suggestion
        let actual_read_size = self.readbuf.borrow_mut().buf.read(bytes).unwrap();
        return Ok(Ok(actual_read_size));
```

src/shims/unix/socket.rs
```diff
@@ -20,17 +38,122 @@ impl FileDescription for SocketPair {
...
+        }
+
+        let data_size = writebuf.clone().unwrap().borrow().buf.len();
+        let available_space = MAX_SOCKETPAIR_BUFFER_CAPACITY.checked_sub(data_size).unwrap();
+        let actual_write_size = write_size.min(available_space);
+        if self.is_nonblock {
+            if write_size <= MAX_SOCKETPAIR_BUFFER_CAPACITY {
+                if available_space < write_size {
+                    // Non-blocking socketpair with write_size <= buffer_size and has
+                    // insufficient space to writer all the data.
```
oli-obk, line 125:
```suggestion
                    // insufficient space to write all the data.
```

src/shims/unix/socket.rs
```diff
@@ -20,17 +38,122 @@ impl FileDescription for SocketPair {
...
+                    return Ok(Err(Error::from(ErrorKind::WouldBlock)));
+                }
+            } else if available_space == 0 {
```
oli-obk, line 128: Is this reachable? If `available_space == 0`, then the only way not to go into the `if available_space < write_size` is if `write_size == 0`, which we already checked at the start of this function

src/shims/unix/socket.rs
```diff
@@ -42,17 +165,82 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                type_ &= !(this.eval_libc_i32("SOCK_CLOEXEC"));
+            }
+        }
+
+        // Fail on unsupported input.
+        // AF_UNIX and AF_LOCAL are synonyms, so we accept both in case
+        // their values differ.
+        if domain != this.eval_libc_i32("AF_UNIX") && domain != this.eval_libc_i32("AF_LOCAL") {
+            throw_unsup_format!(
+                "socketpair: Unsupported domain {:#x} is used, only AF_UNIX
```
oli-obk, line 198:
```suggestion
                "socketpair: Unsupported domain {:#x} is used, only AF_UNIX \
```

otherwise all the whitespace will be included in the message. Same below

src/shims/unix/socket.rs
```diff
@@ -42,17 +165,82 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            throw_unsup_format!(
+                "socketpair: Unsupported type {:#x} is used, only SOCK_STREAM,
+                                 SOCK_CLOEXEC and SOCK_NONBLOCK are allowed",
+                type_
+            );
+        } else if protocol != 0 {
+            throw_unsup_format!(
+                "socketpair: Unsupported socket protocol {} is used,
+                                 only 0 is allowed",
+                protocol
```
oli-obk, line 212: you can inline simple variables that get printed

```suggestion
                "socketpair: Unsupported socket protocol {protocol} is used,
                                 only 0 is allowed",
```

src/shims/unix/socket.rs
tiif: I think this works, but we might need to add ``&mut`` in front of ``bytes``
```rust
let actual_read_size = self.readbuf.borrow_mut().buf.read(&mut bytes).unwrap();
return Ok(Ok(actual_read_size));
```

src/shims/unix/socket.rs
tiif: Nice catch!

src/shims/unix/socket.rs
tiif: That branch will hit only if:
1. write_size != 0 (checked at the start of the program)
2. socketpair is ``non-blocking``
3. write_size > MAX_SOCKETPAIR_BUFFER_CAPACITY
4. available space == 0

To go into ``if available_space < write_size`` above, it needs to be ``write_size <= MAX_SOCKETPAIR_BUFFER_CAPACITY``; To go into ``available_size <= write_size`` below, it needs to be be ``blocking`` socketpair. 

Let me know if I missed anything. There are a lot of branches and I am not sure if I got them correctly even after multiple checks.

oli-obk: Why is `write_size > MAX_SOCKETPAIR_BUFFER_CAPACITY` something useful to check? I would have assumed the `available_space < write_size` check to suffice. Basically the only difference between blocking and nonblocking is the behaviour if we'd block, not how we check whether we block.

src/shims/unix/socket.rs
```diff
@@ -20,17 +38,121 @@ impl FileDescription for SocketPair {
...
+                if data_size == 0 {
+                    // Blocking socketpair with no writer and empty buffer.
+                    return Ok(Ok(0));
+                }
+            } else if data_size == 0 {
+                // Blocking socketpair with writer and empty buffer.
+                // FIXME: blocking is currently not supported
+                throw_unsup_format!("socketpair read: blocking isn't supported yet");
+            }
+        }
```
oli-obk, line 89: you can deduplicate these arms by moving the `self.is_nonblock` check into the `data_size == 0` arm

oli-obk (approved):

oli-obk:
beyond the two cleanups, this lgtm now

src/shims/unix/socket.rs
```diff
@@ -20,17 +38,121 @@ impl FileDescription for SocketPair {
...
+        if let Some(data_race) = &ecx.machine.data_race {
+            data_race.acquire_clock(&self.readbuf.clone().borrow_mut().clock, &ecx.machine.threads);
+        }
```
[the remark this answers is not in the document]
oli-obk: oh heh, yea. Your logic is just the convenience method inlined.

tiif:
There is a minor change in non-blocking write specification. Basically it is not necessary to have ``write_size <= buffer_size`` and ``write_size > buffer_size`` check, so for write, so it'd be:
for blocking:
```
    - if read end open
        - if available_space == 0
            - block
        - else 
            - write as much as we can then return number of bytes written. 
    - if read end close
        - fail with ``EPIPE``
```

for non-block: 
```
    - if read end open
        - if available_space == 0
            - fail with ``EWOULDBLOCK``
        - else
            - write as much as we can and return number of bytes written
    - if read end closed
        - fail with ``EPIPE``
```

<details>
  <summary>Full specification for reference</summary>
  
### For non-blocking 
- read
    - if write end open but no data available 
        - fail with ``EWOULDBLOCK``
    - if write end close
       - if buffer is not empty
          - read as much as we need/can and return number of bytes read
       - else (buffer is empty )
          - return 0
    - if write end open with data
        - if read size < available data
            - read as requested and return
        - if read size  >= available data
            - read as much as we can and return (partial read)
- write
    - if read end open
        - if available_space == 0
            - fail with ``EWOULDBLOCK``
        - else
            - write as much as we can and return number of bytes written
    - if read end closed
        - fail with ``EPIPE``
### For blocking
- read
    - if write end open but no data available 
        - block 
    - if write end close
       - if buffer is not empty
          - read as much as we need/can and return number of bytes read
       - else (buffer is empty )
          - return 0
    - if write end open with data
        - if read size < available data
            - read as requested and return
        - if read size  >= available data
            - read as much as we can and return 
- write
    - if read end open
        - if available_space == 0
            - block
        - else 
            - write as much as we can then return number of bytes written. 
    - if read end close
        - fail with ``EPIPE``



</details>

src/shims/unix/socket.rs
tiif: Specification updated in the latest comment below https://github.com/rust-lang/miri/pull/3609#issuecomment-2142513146

oli-obk (approved):
@RalfJung this looks ready to me now. Once you've approved, too, let's squash all commits into one

src/shims/unix/socket.rs
```diff
@@ -1,15 +1,33 @@
...
+    writebuf: Weak<RefCell<Buffer>>,
+    readbuf: Rc<RefCell<Buffer>>,
+    is_nonblock: bool,
+}
+
+#[derive(Debug)]
+struct Buffer {
+    buf: VecDeque<u8>,
+    clock: VClock,
+    buf_has_writer: bool,
```
RalfJung, line 29: This field is also interesting enough that it should have a comment.

src/shims/unix/socket.rs
```diff
@@ -20,17 +38,97 @@ impl FileDescription for SocketPair {
         self: Box<Self>,
         _communicate_allowed: bool,
     ) -> InterpResult<'tcx, io::Result<()>> {
+        // This is used to signal socketfd of other side that there is no writer to it's readbuf.
```
RalfJung, line 41:
```suggestion
        // This is used to signal socketfd of other side that there is no writer to its readbuf.
```

src/shims/unix/socket.rs
```diff
@@ -20,17 +38,97 @@ impl FileDescription for SocketPair {
...
+        // If the upgrade fails, there is no need to update as all read ends has been dropped.
```
RalfJung, line 42:
```suggestion
        // If the upgrade fails, there is no need to update as all read ends have been dropped.
```

src/shims/unix/socket.rs
```diff
@@ -20,17 +38,97 @@ impl FileDescription for SocketPair {
...
+
+        // Always succeed on read size 0.
+        if request_byte_size == 0 {
+            return Ok(Ok(0));
+        }
+
+        if data_size == 0 {
+            if !readbuf_has_writer {
+                // Socketpair with no writer and empty buffer.
+                return Ok(Ok(0));
```
RalfJung, line 67:
```suggestion
                // 0 bytes successfully read indicates end-of-file.
                return Ok(Ok(0));
```

src/shims/unix/socket.rs
```diff
@@ -20,17 +38,97 @@ impl FileDescription for SocketPair {
...
+                    return Ok(Err(Error::from(ErrorKind::WouldBlock)));
+                } else {
+                    // Blocking socketpair with writer and empty buffer.
+                    // FIXME: blocking is currently not supported
+                    throw_unsup_format!("socketpair read: blocking isn't supported yet");
+                }
+            }
+        }
+
+        // Prevent false alarm in data race detection when doing synchronisation via socketpair.
```
RalfJung, line 84:
```suggestion
        // Synchronize with all previous writes to this buffer.
        // FIXME: this over-synchronizes; a more precise approach would be to
        // only sync with the writes whose data we will read.
```

src/shims/unix/socket.rs
```diff
@@ -20,17 +38,97 @@ impl FileDescription for SocketPair {
...
+        ecx.acquire_clock(&self.readbuf.clone().borrow_mut().clock);
```
RalfJung, line 85: It should not be necessary to do a `clone` of the `Rc` just to acquire the clock.

src/shims/unix/socket.rs
```diff
@@ -20,17 +38,97 @@ impl FileDescription for SocketPair {
...
+        // Do full read / partial read based on the space available.
```
RalfJung, line 86:
```suggestion
        // Do full read / partial read based on the space available.
        // Conveniently, `read` exists on `VecDeque` and has exactly the desired behavior.
```

src/shims/unix/socket.rs
```diff
@@ -1,15 +1,33 @@
...
+/// be configured in the real system.
+const MAX_SOCKETPAIR_BUFFER_CAPACITY: usize = 212992;
+
 /// Pair of connected sockets.
-///
-/// We currently don't allow sending any data through this pair, so this can be just a dummy.
 #[derive(Debug)]
-struct SocketPair;
+struct SocketPair {
+    writebuf: Weak<RefCell<Buffer>>,
```
RalfJung, line 20:
```suggestion
    // By making the write link weak, a `write` can detect when all readers are
    // gone, and trigger EPIPE as appropriate.
    writebuf: Weak<RefCell<Buffer>>,
```

src/shims/unix/socket.rs
```diff
@@ -20,17 +38,97 @@ impl FileDescription for SocketPair {
...
+    }
+
+    fn write<'tcx>(
+        &mut self,
+        _communicate_allowed: bool,
+        bytes: &[u8],
+        ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, io::Result<usize>> {
+        let write_size = bytes.len();
+        // Always succeed on write size 0.
```
RalfJung, line 98:
```suggestion
        // Always succeed on write size 0.
        // ("If count is zero and fd refers to a file other than a regular file, the results are not specified.")
```

src/shims/unix/socket.rs
```diff
@@ -20,17 +38,97 @@ impl FileDescription for SocketPair {
...
+        }
+
+        let writebuf = self.writebuf.upgrade();
+        if writebuf.is_none() {
+            // If the upgrade from Weak to Rc fails, it indicates that all read ends have been
+            // closed.
+            return Ok(Err(Error::from(ErrorKind::BrokenPipe)));
+        }
+
+        let data_size = writebuf.clone().unwrap().borrow().buf.len();
```
RalfJung, line 110: You shouldn't have to `clone` the Rc just to determine the length.

Ideally call `borrow_mut` exactly once and use it for the entire operation.

src/shims/unix/socket.rs
```diff
@@ -20,17 +38,97 @@ impl FileDescription for SocketPair {
...
+        if available_space == 0 {
+            if self.is_nonblock {
+                // Non-blocking socketpair with a full buffer.
+                return Ok(Err(Error::from(ErrorKind::WouldBlock)));
+            } else {
+                // Blocking socketpair with a full buffer.
+                throw_unsup_format!("socketpair write: blocking isn't supported yet");
+            }
+        }
+        // Prevent false alarm in data race detection when doing synchronisation via socketpair.
```
RalfJung, line 122:
```suggestion
        // Remember this clock so `read` can synchronize with us.
```

src/shims/unix/socket.rs
```diff
@@ -20,17 +38,97 @@ impl FileDescription for SocketPair {
...
+        writebuf.clone().unwrap().borrow_mut().clock.join(&ecx.release_clock().unwrap());
```
RalfJung, line 123: You cannot `unwrap` this, that can ICE. Instead do something like
```rust
if let Some(clock) = ecx.release_clock() {
  // ...
}
```

src/shims/unix/socket.rs
```diff
@@ -42,17 +140,81 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let mut type_ = this.read_scalar(type_)?.to_i32()?;
+        let protocol = this.read_scalar(protocol)?.to_i32()?;
         let sv = this.deref_pointer(sv)?;
 
-        // FIXME: fail on unsupported inputs
+        let mut is_sock_nonblock = false;
+
+        // Parse and remove the type flags that we support. If type != 0 after removing,
+        // unsupported flags are used.
+        // FIXME: SOCK_DRAM that supposed to have message boundaries is not supported
```
RalfJung, line 152:
```suggestion
```
This is more like "TODO: support SOCK_DGRAM", but I don't think that's something we want/need in the code. It can be an issue if needed.

src/shims/unix/socket.rs
```diff
@@ -42,17 +140,81 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        if type_ & this.eval_libc_i32("SOCK_STREAM") == this.eval_libc_i32("SOCK_STREAM") {
+            type_ &= !(this.eval_libc_i32("SOCK_STREAM"));
+        }
+        // FIXME: To support macos for tokio socketpair, extra handling on the libc::O_NONBLOCK
+        // flag is needed because macos doesn't have SOCK_CLOEXEC and SOCK_NONBLOCK flag.
```
RalfJung, line 157:
```suggestion
        // SOCK_NONBLOCK only exists on Linux.
```
Please also then adjust the check to be ` == "linux"`. Currently you re saying that *all OSes* except macOS support SOCK_NONBLOCK. Have you checked that FreeBSD, Solaris, OpenBSD, and so on all support it? If not, then please only add support on OSes where we are use support exists. :)

tests/fail-dep/libc/socketpair_read_blocking.rs
RalfJung, line 1: FWIW I don't think these tests are necessary. We don't have tests for *all* unsupported situations, that would quickly get out of hand. OTOH now that you already wrote the tests we may as well keep them. :)

RalfJung (commented):
Thanks a lot! I have a bunch of comments but they're mostly minor. :)

tests/fail-dep/libc/socketpair_read_blocking.rs
tiif: Ah I see, sure :D

src/shims/unix/socket.rs
```diff
@@ -1,15 +1,38 @@
...
+    is_nonblock: bool,
+}
+
+#[derive(Debug)]
+struct Buffer {
+    buf: VecDeque<u8>,
+    clock: VClock,
+    // Indicates if there is at least one active writer to this buffer.
+    // If all writer of this buffer is dropped, buf_has_writer becomes false and hit
+    // the 'no writer' branch.
```
RalfJung, line 33:
```suggestion
    /// Indicates if there is at least one active writer to this buffer.
    /// If all writers of this buffer are dropped, buf_has_writer becomes false and we
    /// indicate EOF instead of blocking.
```

src/shims/unix/socket.rs
```diff
@@ -20,17 +43,107 @@ impl FileDescription for SocketPair {
...
+    fn read<'tcx>(
+        &mut self,
+        _communicate_allowed: bool,
+        bytes: &mut [u8],
+        ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, io::Result<usize>> {
+        let request_byte_size = bytes.len();
+        let data_size: usize = self.readbuf.borrow().buf.len();
+        let readbuf_has_writer = self.readbuf.borrow().buf_has_writer;
+        let mut mut_readbuf = self.readbuf.borrow_mut();
```
RalfJung, line 63: You are still borrowing the same RefCell 3 times in a row here. There's no need for that, borrowing once should suffice.

src/shims/unix/socket.rs
```diff
@@ -20,17 +43,107 @@ impl FileDescription for SocketPair {
...
+        ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, io::Result<usize>> {
+        let write_size = bytes.len();
+        // Always succeed on write size 0.
+        // ("If count is zero and fd refers to a file other than a regular file, the results are not specified.")
+        if write_size == 0 {
+            return Ok(Ok(0));
+        }
+
+        let writebuf = match self.writebuf.upgrade() {
```
RalfJung, line 114: Use `let Some(writebuf) = ... else { ... }` here, that's more compact and more readable.

RalfJung (commented):
We're getting there. :)

src/shims/unix/socket.rs
```diff
@@ -20,17 +43,107 @@ impl FileDescription for SocketPair {
...
+            Some(writebuf) => writebuf,
+            None => {
+                // If the upgrade from Weak to Rc fails, it indicates that all read ends have been
+                // closed.
+                return Ok(Err(Error::from(ErrorKind::BrokenPipe)));
+            }
+        };
+        let data_size = writebuf.borrow().buf.len();
+        let mut mut_writebuf = writebuf.borrow_mut();
```
RalfJung, line 123: This is also still borrowing the same RefCell twice in a row.

src/shims/unix/socket.rs
tiif: Right, thanks! I thought mutable borrow and immutable borrow need to be separated.

tiif:
Thanks for the thorough review ;)

@rustbot ready

src/shims/unix/socket.rs
```diff
@@ -58,9 +58,9 @@ impl FileDescription for SocketPair {
         ecx: &mut MiriInterpCx<'tcx>,
     ) -> InterpResult<'tcx, io::Result<usize>> {
         let request_byte_size = bytes.len();
-        let data_size: usize = self.readbuf.borrow().buf.len();
-        let readbuf_has_writer = self.readbuf.borrow().buf_has_writer;
-        let mut mut_readbuf = self.readbuf.borrow_mut();
+        let mut readbuf = self.readbuf.borrow_mut();
+        let data_size: usize = readbuf.buf.len();
+        let readbuf_has_writer = readbuf.buf_has_writer;
```
RalfJung, line 63: Both of these let-bindings are used exactly once. Are they worth having as separate variables here or does the code become cleaner when they get inlined?

In fact data_size is just used in a `== 0` check so that could become `if readbuf.buf.is_empty()`.

src/shims/unix/socket.rs
```diff
@@ -119,8 +119,8 @@ impl FileDescription for SocketPair {
                 return Ok(Err(Error::from(ErrorKind::BrokenPipe)));
             }
         };
-        let data_size = writebuf.borrow().buf.len();
-        let mut mut_writebuf = writebuf.borrow_mut();
+        let mut writebuf = writebuf.borrow_mut();
+        let data_size = writebuf.buf.len();
         let available_space = MAX_SOCKETPAIR_BUFFER_CAPACITY.checked_sub(data_size).unwrap();
         let actual_write_size = write_size.min(available_space);
```
RalfJung, line 122: This `let` can be moved down, it's not used until the `extend` below.

RalfJung:
Just two more nits. :)

Please squash after resolving them.

tiif:
Thanks, just a heads-up, I did rebase + squash again because the git log is quite a mess, I will find a better workflow soon.

RalfJung:
@bors r+

> Thanks, just a heads-up, I did rebase + squash again because the git log is quite a mess, I will find a better workflow soon.

What I usually do is look at `git log` to find the most recent bors commit, copy-paste that commit ID, and then do `git rebase --interactive <that commit id>` to squash.

tiif:
Thanks, I will try that next time.

PR #3610, opened 2024-05-17
Use `throw_unsup_format!` instead of returning `ENOTSUP` in the mmap shim

marc0246:
I noticed this while trying to use `mmap` with `PROT_NONE`, which resulted in this error message:
```
Os { code: 95, kind: Uncategorized, message: "<unknown errnum in strerror_r: 95>" }
```
cc @saethlin

### DIFF
--- a/src/shims/unix/mem.rs
+++ b/src/shims/unix/mem.rs
@@ -71,24 +71,27 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
             throw_unsup_format!("Miri does not support file-backed memory mappings");
         }
 
-        // POSIX says:
-        // [ENOTSUP]
-        // * MAP_FIXED or MAP_PRIVATE was specified in the flags argument and the implementation
-        // does not support this functionality.
-        // * The implementation does not support the combination of accesses requested in the
-        // prot argument.
-        //
-        // Miri doesn't support MAP_FIXED or any any protections other than PROT_READ|PROT_WRITE.
-        if flags & map_fixed != 0 || prot != prot_read | prot_write {
-            this.set_last_error(this.eval_libc("ENOTSUP"))?;
-            return Ok(this.eval_libc("MAP_FAILED"));
+        // Miri doesn't support MAP_FIXED.
+        if flags & map_fixed != 0 {
+            throw_unsup_format!(
+                "Miri does not support calls to mmap with MAP_FIXED as part of the flags argument",
+            );
+        }
+
+        // Miri doesn't support protections other than PROT_READ|PROT_WRITE.
+        if prot != prot_read | prot_write {
+            throw_unsup_format!(
+                "Miri does not support calls to mmap with protections other than \
+                 PROT_READ|PROT_WRITE",
+            );
         }
 
         // Miri does not support shared mappings, or any of the other extensions that for example
         // Linux has added to the flags arguments.
         if flags != map_private | map_anonymous {
             throw_unsup_format!(
-                "Miri only supports calls to mmap which set the flags argument to MAP_PRIVATE|MAP_ANONYMOUS"
+                "Miri only supports calls to mmap which set the flags argument to \
+                 MAP_PRIVATE|MAP_ANONYMOUS",
             );
         }
 
--- a/tests/pass-dep/libc/mmap.rs
+++ b/tests/pass-dep/libc/mmap.rs
@@ -69,36 +69,6 @@ fn test_mmap<Offset: Default>(
     assert_eq!(ptr, libc::MAP_FAILED);
     assert_eq!(Error::last_os_error().raw_os_error().unwrap(), libc::EINVAL);
 
-    let ptr = unsafe {
-        mmap(
-            ptr::without_provenance_mut(page_size * 64),
-            page_size,
-            libc::PROT_READ | libc::PROT_WRITE,
-            // We don't support MAP_FIXED
-            libc::MAP_PRIVATE | libc::MAP_ANONYMOUS | libc::MAP_FIXED,
-            -1,
-            Default::default(),
-        )
-    };
-    assert_eq!(ptr, libc::MAP_FAILED);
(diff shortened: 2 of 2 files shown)


src/helpers.rs
```diff
@@ -75,12 +74,15 @@ const UNIX_IO_ERROR_TABLE: &[(&str, std::io::ErrorKind)] = {
...
-        // first one will be used for the backwards mapping.
+        // The following have multiple valid options. We have all of them for the forwards mapping;
+        // only the first one will be used for the backwards mapping.
         ("EPERM", PermissionDenied),
         ("EACCES", PermissionDenied),
         ("EWOULDBLOCK", WouldBlock),
         ("EAGAIN", WouldBlock),
+        ("ENOTSUP", Unsupported),
+        ("EOPNOTSUPP", Unsupported),
+        ("ENOSYS", Unsupported),
```
saethlin, line 85: This is a really unfortunate collision, but we are not the first to notice it. It was discussed on the libs PR that added `ErrorKind::Unsupported`: https://github.com/rust-lang/rust/pull/78880#issuecomment-777623233

Due to this degeneracy in the mapping, I think we should change our `strerror` mapping to not convert from libc constants to `ErrorKind`.

saethlin:
@marc0246 Can you add a test that verifies that a program which encounters `ENOTSUP` can correctly print `last_os_error`?

The precipitating program is this:
```rust
let ptr = unsafe {
    libc::mmap(
        std::ptr::null_mut(),
        4096,
        libc::PROT_NONE,
        libc::MAP_PRIVATE | libc::MAP_ANONYMOUS,
        -1,
        0,
    )
};

if ptr == libc::MAP_FAILED {
    Err::<(), _>(std::io::Error::last_os_error()).unwrap();
}
```
Which before this PR prints:
```
thread 'main' panicked at src/main.rs:1:265:
called `Result::unwrap()` on an `Err` value: Os { code: 95, kind: Uncategorized, message: "<unknown errnum in strerror_r: 95>" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
note: in Miri, you may have to set `-Zmiri-env-forward=RUST_BACKTRACE` for the environment variable to have an effect
```

This scenario is close to a viable test case, the only complaint I'd have about this as a test is that it relies on the details of our `mmap` implementation.

RalfJung:
The program could set the last error itself and then print it, instead of relying on `mmap`?

src/helpers.rs
RalfJung, line 85: Previously we had `Unsupported` map back t `ENOSYS`. This changes that to `ENOTSUP`. Why is that the better choice?

src/helpers.rs
```diff
@@ -75,12 +74,15 @@ const UNIX_IO_ERROR_TABLE: &[(&str, std::io::ErrorKind)] = {
...
-        // The following have two valid options. We have both for the forwards mapping; only the
-        // first one will be used for the backwards mapping.
+        // The following have multiple valid options. We have all of them for the forwards mapping;
+        // only the first one will be used for the backwards mapping.
         ("EPERM", PermissionDenied),
         ("EACCES", PermissionDenied),
         ("EWOULDBLOCK", WouldBlock),
         ("EAGAIN", WouldBlock),
+        ("ENOTSUP", Unsupported),
+        ("EOPNOTSUPP", Unsupported),
```
RalfJung, line 84: Further up the comment says this mapping should be in sync with the one in the standard library. The new ones you are adding here are not present in the standard library. Why should we add them here?

src/helpers.rs
saethlin: Ah, right. The library decided to _not_ have this mapping farther down the thread: https://github.com/rust-lang/rust/pull/78880#issuecomment-778495108. I must have remembered this because of the `ENOSYS => ErrorKind::Unsupported` mapping that exists in the library.

src/helpers.rs
saethlin: > Why should we add them here?

I would prefer that we do _something_ in this situation. The motivating example is the `mmap` call I pasted above, which reports an error string with wording that suggests there is a bug in Miri.

RalfJung: Maybe Miri's mmap shouldn't set `ENOTSUP` if that error code is not supported (heh) by Miri itself.

RalfJung: Though usually what we do in that case is just `throw_unsup_format!`, which can give a lot more context and so be much more informative. Is there a reason mmap doesn't do that?

src/helpers.rs
saethlin: We're having the same conversation on about 4 different threads. I'm going to reply on the PR as a whole instead of on a diff.

saethlin:
> Do you think I should close the PR against std then?

libs decided not to FCP a colliding conversion from errno values to `ErrorKind` so I think we shouldn't suggest adding one.

About this PR as a whole, we definitely need a better message for this `mmap` misuse overall. Ralf said:

> Though usually what we do in that case is just `throw_unsup_format!`, which can give a lot more context and so be much more informative. Is there a reason mmap doesn't do that?

There is no good reason. I waffled a lot while implementing the `mmap` shim about whether we should return an error or `throw_unsup_format!`. The core tension here is that if we return an error, callers can use some kind of fallback behavior. If we use `throw_unsup_format!`, we halt the interpreter and the caller needs to add a `cfg(miri)` to their code to make progress.

I suspect that given your experience and given that I've now been corrected that the standard library does not use `ENOTSUP` anywhere, it seems dubious to return `ENOTSUP` for these calls. I think now I'd prefer to `throw_unsup_format!` with a descriptive message instead. We should probably provide a different string for use of `MAP_FIXED` and use of protections other than `PROT_READ | PROT_WRITE`. Here's the relevant code: https://github.com/rust-lang/miri/blob/e1099c653869d18d2590bc3366b74b96a70997b1/src/shims/unix/mem.rs#L83-L84

RalfJung:
Okay, so let's bring mmap in sync with all other Miri shims then by defaulting to throw_unsup_format.

@marc0246 can you update the PR to do that instead?
(I don't think it needs a test, we don't usually test the "unsupported" paths.)

RalfJung:
Looks good, thanks!

@bors r+

PR #3611, opened 2024-05-17
Ignore the Helix configuration directory

marc0246:
For us Helix users.


saethlin:
Ah I see we already have an ignore for vscode. If you give a moose a muffin...

@bors r+

PR #3614, opened 2024-05-18
solaris/illumos localtime_r / clock_getime support enabled.

devnexen:
clock_gettime support CLOCK_REALTIME/CLOCK_MONOTONIC clockid_t. localtime_r is supported only tm struct is more limited than other
 supported platforms.

### DIFF
--- a/src/shims/time.rs
+++ b/src/shims/time.rs
@@ -62,6 +62,16 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 // We need to support it because std uses it.
                 relative_clocks.push(this.eval_libc_i32("CLOCK_UPTIME_RAW"));
             }
+            "solaris" | "illumos" => {
+                // The REALTIME clock returns the actual time since the Unix epoch.
+                absolute_clocks = vec![this.eval_libc_i32("CLOCK_REALTIME")];
+                // MONOTONIC, in the other hand, is the high resolution, non-adjustable
+                // clock from an arbitrary time in the past.
+                // Note that the man page mentions HIGHRES but it is just
+                // an alias of MONOTONIC and the libc crate does not expose it anyway.
+                // https://docs.oracle.com/cd/E23824_01/html/821-1465/clock-gettime-3c.html
+                relative_clocks = vec![this.eval_libc_i32("CLOCK_MONOTONIC")];
+            }
             target => throw_unsup_format!("`clock_gettime` is not supported on target OS {target}"),
         }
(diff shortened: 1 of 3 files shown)


src/shims/time.rs
```diff
@@ -154,46 +164,73 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
-        }
-        let offset_hour = offset_in_seconds.abs() / 3600;
-        write!(tm_zone, "{:02}", offset_hour).unwrap();
-        let offset_min = (offset_in_seconds.abs() % 3600) / 60;
-        if offset_min != 0 {
-            write!(tm_zone, "{:02}", offset_min).unwrap();
+        match this.tcx.sess.target.os.as_ref() {
+            "solaris" | "illumos" => {
+                // solaris/illumos system tm struct does not have
+                // the additional tm_zone/tm_gmtoff fields.
```
oli-obk, line 170: instead of duplicating the core part, can you skip the parts that write the extra fields?

oli-obk (changes requested):

tests/pass-dep/libc/libc-time.rs
```diff
@@ -1,5 +1,6 @@
 //@ignore-target-windows: no libc time APIs on Windows
 //@compile-flags: -Zmiri-disable-isolation
+#[cfg(any(target_os = "linux", target_os = "macos", target_os = "freebsd"))]
```
RalfJung, line 3: Instead of making the import conditional, please remove it entirely and use the fully expanded type name below where `CStr` shows up.

src/shims/time.rs
```diff
@@ -188,11 +174,42 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+            }
+
+            // FIXME: String de-duplication is needed so that we only allocate this string only once
+            // even when there are multiple calls to this function.
+            let tm_zone_ptr = this
+                .alloc_os_str_as_c_str(&OsString::from(tm_zone), MiriMemoryKind::Machine.into())?;
+
+            this.write_pointer(tm_zone_ptr, &this.project_field_named(&result, "tm_zone")?)?;
+            this.write_int_fields_named(
+                &[("tm_isdst", tm_isdst), ("tm_gmtoff", tm_gmtoff.into())],
```
oli-obk, line 209:
```suggestion
                &[("tm_gmtoff", tm_gmtoff.into())],
```

tm_isdst was now written twice

PR #3622, opened 2024-05-19
Implement LLVM x86 SSE4.2 intrinsics

TDecking:
SSE4.2 is arguably the least important SIMD extension for the x86 ISA, but it should still be supported for the sake of completeness.

### DIFF
--- a/src/shims/x86/sse42.rs
+++ b/src/shims/x86/sse42.rs
@@ -1,0 +1,500 @@
+use rustc_middle::mir;
+use rustc_middle::ty::layout::LayoutOf as _;
+use rustc_middle::ty::Ty;
+use rustc_span::Symbol;
+use rustc_target::abi::Size;
+use rustc_target::spec::abi::Abi;
+
+use crate::*;
+
+/// A bitmask constant for scrutinizing the immediate byte provided
+/// to the string comparison intrinsics. It distinuishes between
+/// 16-bit integers and 8-bit integers. See [`compare_strings`]
+/// for more details about the immediate byte.
+const USE_WORDS: u8 = 1;
+
+/// A bitmask constant for scrutinizing the immediate byte provided
+/// to the string comparison intrinsics. It distinuishes between
+/// signed integers and unsigned integers. See [`compare_strings`]
+/// for more details about the immediate byte.
+const USE_SIGNED: u8 = 2;
+
+/// The main worker for the string comparison intrinsics, where the given
+/// strings are analyzed according to the given immediate byte.
+///
+/// # Arguments
+///
+/// * `str1` - The first string argument. It is always a length 16 array of bytes
+///   or a length 8 array of two-byte words.
+/// * `str2` - The second string argument. It is always a length 16 array of bytes
+///   or a length 8 array of two-byte words.
+/// * `len` is the length values of the supplied strings. It is distinct from the operand length
+///   in that it describes how much of `str1` and `str2` will be used for the calculation and may
+///   be smaller than the array length of `str1` and `str2`. The string length is counted in bytes
+///   if using byte operands and in two-byte words when using two-byte word operands.
+///   If the value is `None`, the length of a string is determined by the first
+///   null value inside the string.
+/// * `imm` is the immediate byte argument supplied to the intrinsic. The byte influences
+///   the operation as follows:
+///
+///   ```text
+///   0babccddef
+///     || | |||- Use of bytes vs use of two-byte words inside the operation.
+///     || | ||
+///     || | ||- Use of signed values versus use of unsigned values.
+///     || | |
+///     || | |- The comparison operation performed. A total of four operations are available.
+///     || |    * Equal any: Checks which characters of `str2` are inside `str1`.
+///     || |    * String ranges: Check if characters in `str2` are inside the provided character ranges.
+///     || |      Adjacent characters in `str1` constitute one range.
+///     || |    * String comparison: Mark positions where `str1` and `str2` have the same character.
+///     || |    * Substring search: Mark positions where `str1` is a substring in `str2`.
+///     || |
+///     || |- Result Polarity. The result bits may be subjected to a bitwise complement
+///     ||    if these bits are set.
+///     ||
+///     ||- Output selection. This bit has two meanings depending on the instruction.
+///     |   If the instruction is generating a mask, it distinguishes between a bit mask
+///     |   and a byte mask. Otherwise it distinguishes between the most significand bit
+///     |   and the least significand bit when generating an index.
+///     |
+///     |- This bit is ignored. It is expected that this bit is set to zero, but it is
+///        not a requirement.
+///   ```
+///
+/// # Returns
+///
+/// A result mask. The bit at index `i` inside the mask is set if 'str2' starting at `i`
+/// fulfills the test as defined inside the immediate byte.
+/// The mask may be negated if negation flags inside the immediate byte are set.
+///
+/// For more information, see the Intel Software Developer's Manual, Vol. 2b, Chapter 4.1.
+#[allow(clippy::arithmetic_side_effects)]
+fn compare_strings<'tcx>(
+    this: &mut MiriInterpCx<'tcx>,
+    str1: &OpTy<'tcx>,
+    str2: &OpTy<'tcx>,
+    len: Option<(u64, u64)>,
+    imm: u8,
+) -> InterpResult<'tcx, i32> {
+    let default_len = default_len::<u64>(imm);
+    let (len1, len2) = if let Some(t) = len {
+        t
+    } else {
+        let len1 = implicit_len(this, str1, imm)?.unwrap_or(default_len);
+        let len2 = implicit_len(this, str2, imm)?.unwrap_or(default_len);
+        (len1, len2)
+    };
+
+    let mut result = 0;
+    match (imm >> 2) & 3 {
+        0 => {
+            // Equal any: Checks which characters of `str2` are inside `str1`.
+            for i in 0..len2 {
+                let ch2 = this.read_immediate(&this.project_index(str2, i)?)?;
+
+                for j in 0..len1 {
+                    let ch1 = this.read_immediate(&this.project_index(str1, j)?)?;
+
+                    let eq = this.binary_op(mir::BinOp::Eq, &ch1, &ch2)?;
+                    if eq.to_scalar().to_bool()? {
+                        result |= 1 << i;
+                        break;
+                    }
+                }
+            }
+        }
+        1 => {
+            // String ranges: Check if characters in `str2` are inside the provided character ranges.
+            // Adjacent characters in `str1` constitute one range.
+            let len1 = len1 - (len1 & 1);
+            let get_ch = |ch: Scalar| -> InterpResult<'tcx, i32> {
+                let result = match (imm & USE_WORDS != 0, imm & USE_SIGNED != 0) {
+                    (true, true) => i32::from(ch.to_i16()?),
+                    (true, false) => i32::from(ch.to_u16()?),
+                    (false, true) => i32::from(ch.to_i8()?),
+                    (false, false) => i32::from(ch.to_u8()?),
+                };
+                Ok(result)
+            };
+
+            for i in 0..len2 {
+                for j in (0..len1).step_by(2) {
+                    let ch2 = get_ch(this.read_scalar(&this.project_index(str2, i)?)?)?;
+                    let ch1_1 = get_ch(this.read_scalar(&this.project_index(str1, j)?)?)?;
+                    let ch1_2 = get_ch(this.read_scalar(&this.project_index(str1, j + 1)?)?)?;
+
+                    if ch1_1 <= ch2 && ch2 <= ch1_2 {
+                        result |= 1 << i;
+                    }
+                }
+            }
+        }
+        2 => {
+            // String comparison: Mark positions where `str1` and `str2` have the same character.
+            result = (1 << default_len) - 1;
+            result ^= (1 << len1.max(len2)) - 1;
+
+            for i in 0..len1.min(len2) {
+                let ch1 = this.read_immediate(&this.project_index(str1, i)?)?;
+                let ch2 = this.read_immediate(&this.project_index(str2, i)?)?;
+                let eq = this.binary_op(mir::BinOp::Eq, &ch1, &ch2)?;
+                result |= i32::from(eq.to_scalar().to_bool()?) << i;
+            }
+        }
+        3 => {
+            // Substring search: Mark positions where `str1` is a substring in `str2`.
+            if len1 == 0 {
+                result = (1 << default_len) - 1;
+            } else if len1 <= len2 {
+                for i in 0..len2 {
+                    if len1 > len2 - i {
+                        break;
+                    }
+
+                    result |= 1 << i;
+
+                    for j in 0..len1 {
+                        let k = i + j;
+
+                        if k >= default_len {
+                            break;
+                        } else {
+                            let ch1 = this.read_immediate(&this.project_index(str1, j)?)?;
+                            let ch2 = this.read_immediate(&this.project_index(str2, k)?)?;
+                            let ne = this.binary_op(mir::BinOp::Ne, &ch1, &ch2)?;
+
+                            if ne.to_scalar().to_bool()? {
+                                result &= !(1 << i);
+                                break;
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        _ => unreachable!(),
+    }
+
+    // Polarity: Possibly perform a bitwise complement on the result.
+    match (imm >> 4) & 3 {
+        3 => result ^= (1 << len1) - 1,
+        1 => result ^= (1 << default_len) - 1,
+        _ => (),
+    }
+
+    Ok(result)
+}
+
+/// Obtain the arguments of the intrinsic based on its name.
+/// The result is a tuple with the following values:
+/// * The first string argument.
+/// * The second string argument.
+/// * The string length values, if the intrinsic requires them.
+/// * The immediate instruction byte.
+///
+/// The string arguments will be transmuted into arrays of bytes
+/// or two-byte words, depending on the value of the immediate byte.
+/// Originally, they are [__m128i](https://doc.rust-lang.org/stable/core/arch/x86_64/struct.__m128i.html) values
+/// corresponding to the x86 128-bit integer SIMD type.
+fn deconstruct_args<'tcx>(
+    unprefixed_name: &str,
+    this: &mut MiriInterpCx<'tcx>,
+    link_name: Symbol,
+    abi: Abi,
+    args: &[OpTy<'tcx>],
+) -> InterpResult<'tcx, (OpTy<'tcx>, OpTy<'tcx>, Option<(u64, u64)>, u8)> {
+    let array_layout_fn = |this: &mut MiriInterpCx<'tcx>, imm: u8| {
+        if imm & USE_WORDS != 0 {
+            this.layout_of(Ty::new_array(this.tcx.tcx, this.tcx.types.u16, 8))
+        } else {
+            this.layout_of(Ty::new_array(this.tcx.tcx, this.tcx.types.u8, 16))
+        }
+    };
+
+    // The fourth letter of each string comparison intrinsic is either 'e' for "explicit" or 'i' for "implicit".
+    // The distinction will correspond to the intrinsics type signature. In this constext, "explicit" and "implicit"
+    // refer to the way the string length is determined. The length is either passed explicitly in the "explicit"
+    // case or determined by a null terminator in the "implicit" case.
+    let is_explicit = match unprefixed_name.as_bytes().get(4) {
+        Some(&b'e') => true,
+        Some(&b'i') => false,
+        _ => unreachable!(),
+    };
+
+    if is_explicit {
+        let [str1, len1, str2, len2, imm] =
+            this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+        let imm = this.read_scalar(imm)?.to_u8()?;
+
+        let default_len = default_len::<u32>(imm);
+        let len1 = u64::from(this.read_scalar(len1)?.to_u32()?.min(default_len));
+        let len2 = u64::from(this.read_scalar(len2)?.to_u32()?.min(default_len));
+
+        let array_layout = array_layout_fn(this, imm)?;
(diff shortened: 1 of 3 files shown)


RalfJung:
@eduardosm you've done all the other x86 intrinsics -- can you help review this one? That would be great. :)
(No pressure, and no rush either.)

RalfJung:
@TDecking
Thanks for the PR!

I see you've done a whole lot of back-and-forth with CI. Is running tests locally not working for you? We have instructions for that [here](https://github.com/rust-lang/miri/blob/master/CONTRIBUTING.md). If that does not work, please describe our problem [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/269128-miri); having to wait for CI adds a lot of friction so we should figure out how to get it all running on your machine.

eduardosm (commented):
Thanks! I never had the nerve to implement SSE4.2 string manipulation intrinsics.

I believe this is still work-in-progress, so I left some comments that I hope you find useful to get things working :)

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,411 @@
...
+                        let lo = lo.wrapping_sub(SUB) & !lo & MASK != 0;
+                        let hi = hi.wrapping_sub(SUB) & !hi & MASK != 0;
+                        lo || hi
+                    } else {
+                        const MASK: u64 = 0x80808080_80808080;
+                        const SUB: u64 = 0x01010101_01010101;
+                        let lo = lo.wrapping_sub(SUB) & !lo & MASK != 0;
+                        let hi = hi.wrapping_sub(SUB) & !hi & MASK != 0;
+                        lo || hi
+                    }
```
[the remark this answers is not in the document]
TDecking: I ended up bundeling the string length code into its own function and decided to also use to check for the string termination. This bit hack is now gone.

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,411 @@
...
+
+#[allow(clippy::arithmetic_side_effects)]
+fn compare_strings<'mir, 'tcx: 'mir>(
+    this: &mut InterpCx<'mir, 'tcx, MiriMachine<'mir, 'tcx>>,
+    str1: &MPlaceTy<'tcx, Provenance>,
+    str2: &MPlaceTy<'tcx, Provenance>,
+    len: Option<(u64, u64)>,
+    imm: u8,
+) -> InterpResult<'tcx, i32> {
+    let use_x16 = imm & 1 != 0;
```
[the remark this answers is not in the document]
TDecking: At least not where `use_x16` is defined.

TDecking:
@eduardosm Thanks for your input. The implementation is fixed and ready.

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,426 @@
...
+                    } else {
+                        let (dest, _dest_len) = this.mplace_to_simd(dest)?;
+                        for i in 0..16 {
+                            let result = if mask & (1 << i) != 0 { u8::MAX } else { 0 };
+                            this.write_scalar(
+                                Scalar::from_u8(result),
+                                &this.project_index(&dest, i)?,
+                            )?;
+                        }
+                    }
```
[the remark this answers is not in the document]
TDecking: The loops are deduplicated.

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,426 @@
...
+            result = (1 << default_len) - 1;
+            result ^= (1 << len1.max(len2)) - 1;
+
+            for i in 0..len1.min(len2) {
+                let ch1 = this.read_scalar(&this.project_index(str1, i)?)?;
+                let ch1 = if use_x16 { ch1.to_u16()? } else { u16::from(ch1.to_u8()?) };
+
+                let ch2 = this.read_scalar(&this.project_index(str2, i)?)?;
+                let ch2 = if use_x16 { ch2.to_u16()? } else { u16::from(ch2.to_u8()?) };
+                result |= i32::from(ch1 == ch2) << i;
```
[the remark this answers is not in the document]
TDecking: Done.

TDecking:
@eduardosm Ready.

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,406 @@
...
+            }
+
+            // Used to implement the `_mm_cmpestrm` and the `_mm_cmpistrm` functions.
+            // These functions compare the input strings and return the resulting mask.
+            "pcmpistrm128" | "pcmpestrm128" => {
+                let (str1, str2, len, imm) =
+                    check_shim(unprefixed_name, this, link_name, abi, args)?;
+                let mask = compare_strings(this, &str1, &str2, len, imm)?;
+
+                if imm & 0b100_0000 != 0 {
```
RalfJung, line 92: What is this comparison about?

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,406 @@
...
+                    Scalar::from_u64(u64::from(result))
+                } else {
+                    Scalar::from_u32(result)
+                };
+
+                this.write_scalar(result, dest)?;
+            }
+
+            // Used to implement the `_mm_cmpestrm` and the `_mm_cmpistrm` functions.
+            // These functions compare the input strings and return the resulting mask.
```
RalfJung, line 86: What is the "resulting mask"?

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,406 @@
...
+                )?;
+            }
+            _ => return Ok(EmulateItemResult::NotSupported),
+        }
+        Ok(EmulateItemResult::NeedsReturn)
+    }
+}
+
+/// The main worker for the string comparison intrinsics, where the given
+/// strings are analyzed according to the given immediate byte.
```
RalfJung, line 216: Would be good to briefly say what the arguments mean, and link to the full documentation.

Consider that this will have to be maintained by people that have never seen any of these intrinsics ever before. :) Please help them by leaving comments they can use as starting point should there ever be bugs or other questions around these intrinsics.

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,406 @@
...
+    match (imm >> 4) & 3 {
+        3 => result ^= (1 << len1) - 1,
+        1 => result ^= (1 << default_len) - 1,
+        _ => (),
+    }
+
+    Ok(result)
+}
+
+fn check_shim<'mir, 'tcx: 'mir>(
```
RalfJung, line 333: Given that we have a global method called `check_shim`, please call this something more specific. Also add a doc comment saying what it does, and what all the components in the return type mean.

This seems to be used by all the string functions. So maybe this is a good place to explain what an "input string" even is -- how is the string buffer passed to the function, how is the length determined, etc. Again, assume the reader knows *nothing* about SSE4.2 (or really, SSE in general), and add enough comments that would give them a basic idea of the structure and goals of the code.

tests/pass/shims/x86/intrinsics-x86-sse42.rs
```diff
@@ -0,0 +1,288 @@
...
+unsafe fn test_sse42() {
+    // Mostly copied from library/stdarch/crates/core_arch/src/x86/sse42.rs
+
+    test_crc();
+    test_cmp();
+    test_str();
+}
+
+#[target_feature(enable = "sse4.2")]
+unsafe fn test_crc() {
```
RalfJung, line 36: Is there an easily accessible set of test vectors we could use to increase coverage here?

tests/pass/shims/x86/intrinsics-x86-sse42.rs
```diff
@@ -0,0 +1,288 @@
...
+#[target_feature(enable = "sse4.2")]
+unsafe fn test_cmp() {
+    let a = _mm_set_epi64x(0x2a, 0);
+    let b = _mm_set1_epi64x(0x00);
+    let i = _mm_cmpgt_epi64(a, b);
+    assert_eq_m128i(i, _mm_set_epi64x(0xffffffffffffffffu64 as i64, 0x00));
+}
+
+#[target_feature(enable = "sse4.2")]
+unsafe fn test_str() {
```
RalfJung, line 86: Having some tests is a great way to understand what an API does... but here the tests have absolutely no comments so it's not super helpful unfortunately.  Is it possible to add some comments that explain what is being tested (at a high level, i.e. for a reader that doesn't know the SSE4.2 functions)?

RalfJung (commented):
Thanks!

I can't really review the actual implementations here, as I don't have any clue what they are even trying to implement. :joy:  I think in general that's fine, since we have tests -- but to make future maintenance possible here, I ask you to please add a lot more comments. There's a bunch of concepts that are just assumed to be known here, which you have to explain -- like how a string is passed to these functions, or what the difference is between "implicit" and "explicit" (or what these terms even mean).

Also, how good is the test coverage? If someone added a random bug in the middle of one of your functions, would the tests catch that? If not, then it'd be great if you could add more tests, so that when we have to adjust this code in the future (e.g. for Miri API changes) we don't accidentally break it.

RalfJung:
@eduardosm thanks a lot for your help with the initial review here! :yellow_heart:

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,406 @@
...
+        if is_zero.to_scalar().to_bool()? {
+            result = Some(i);
+            break;
+        }
+    }
+    Ok(result)
+}
+
+#[inline]
+#[allow(clippy::arithmetic_side_effects)]
```
eduardosm, line 403: This `allow` is not needed anymore.

tests/pass/shims/x86/intrinsics-x86-sse42.rs
TDecking: I increased the test coverage of the string analysis intrinsics to cover the remaining features offered by them.
These as well as the remaining tests have comments attached to them.

tests/pass/shims/x86/intrinsics-x86-sse42.rs
TDecking: Not that I am aware of it. I decided to attach an additional test case for each of the intrinsics just in case.

src/shims/x86/sse42.rs
TDecking: Distinguishing between a bit mask and a byte mask. This comparison is now commented.

TDecking:
@rustbot ready.

@RalfJung I've added some additional test cases, which should cover every possible code path now.
I've also improved the documentation, but I'm not certain wether it is sufficient. To what extent may I assume that a maintainer
will google the intrinsics when performing a major refactor?

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,454 @@
...
+use rustc_middle::ty::Ty;
+use rustc_span::Symbol;
+use rustc_target::abi::Size;
+use rustc_target::spec::abi::Abi;
+
+use crate::*;
+
+/// A constant for scrutinized the immediate byte provided
+/// to the string comparison intrinsics. It distinuishes between
+/// 16-bit integers and 8-bit integers.
```
RalfJung, line 12: I can't parse the first sentence here... what are you trying to say?
From the use of the constant it seems like this is a bitmask? That would be worth saying.

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,454 @@
...
+                    64 => u128::from(right.to_u64()?.reverse_bits()),
+                    _ => unreachable!(),
+                };
+
+                let mut dividend = (crc << bit_size) ^ (v << 32);
+                const POLYNOMIAL: u128 = {
+                    let polynomial = 0x11EDC6F41u128;
+                    polynomial << polynomial.leading_zeros()
+                };
+                while dividend.leading_zeros() < 96 {
```
RalfJung, line 78: Why 96? Above there's a shift by `polynomial.leading_zeros()`  which I think is also 96 -- if it's important that those are the same number, that should be clarified.

src/shims/x86/sse42.rs
RalfJung: Maybe move `compare_strings` and the other helpers up above `emulate_x86_sse42_intrinsic`, that makes it easier to follow what happens when reading the file top-to-bottom.

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,454 @@
...
+        Ok(EmulateItemResult::NeedsReturn)
+    }
+}
+
+/// The main worker for the string comparison intrinsics, where the given
+/// strings are analyzed according to the given immediate byte.
+///
+/// # Arguments
+///
+/// * `this` - The context under which interpretation is happening.
```
RalfJung, line 234:
```suggestion
```
This one is clear, it's everywhere. :)

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,454 @@
...
+/// * `str1` - The first string argument. It is always a length 16 array of bytes
+///   or a length 8 array of words.
```
RalfJung, line 236: Is a "word" 2 bytes here? "Word" means "pointer-sized" elsewhere so I am a bit confused.

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,454 @@
...
+/// * `str2` - The second string argument. It is always a length 16 array of bytes
+///   or a length 8 array of words.
+/// * `len` is the length values of the supplied strings.
+///   If the value is `None`, the length of a string is determined by the first
+///   null value inside the string.
```
RalfJung, line 241: I thought the strings have length 16...? That's what the docs for `str1`/`str2` seem to say.

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,454 @@
...
+/// * `imm` is the immediate byte argument supplied to the intrinsic.
```
RalfJung, line 242: What does this immediate byte *mean*?

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,454 @@
...
+            this.layout_of(Ty::new_array(this.tcx.tcx, this.tcx.types.u16, 8))
+        } else {
+            this.layout_of(Ty::new_array(this.tcx.tcx, this.tcx.types.u8, 16))
+        }
+    };
+
+    // The fourth letter of each string comparison intrinsic is either 'e' for "explicit" or 'i' for "implicit".
+    // The distinction will correspond to the intrinsics type signature. In this constext, "explicit" and "implicit"
+    // refer to the way the string length is determined. The length is either passed explicitly in the "explicit"
+    // case or determined by a null-terminator in the "implicit" case.
```
RalfJung, line 397:
```suggestion
    // case or determined by a null terminator in the "implicit" case.
```

RalfJung (commented):
Thanks, this helped! It does create new questions though. :)

> I've also improved the documentation, but I'm not certain wether it is sufficient. To what extent may I assume that a maintainer
will google the intrinsics when performing a major refactor?

Ideally searching the web won't be needed as the code contains URLs where decent documentation can be found.

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,454 @@
...
+
+                    for i in 0..default_len::<u64>(imm) {
+                        let result = helpers::bool_to_simd_element(mask & (1 << i) != 0, size);
+                        this.write_scalar(result, &this.project_index(&dest, i)?)?;
+                    }
+                } else {
+                    let layout = this.layout_of(this.tcx.types.u128)?;
+                    let dest = dest.transmute(layout, this)?;
+                    // FIXME: Use `Scalar::from_i128` once it exists.
+                    this.write_scalar(Scalar::from_u128(u128::try_from(mask).unwrap()), &dest)?;
```
RalfJung, line 120: I made a PR for that: https://github.com/rust-lang/rust/pull/126157.

RalfJung: With https://github.com/rust-lang/miri/pull/3656 Miri can now use that function; please rebase.

TDecking: Done.

src/shims/x86/sse42.rs
TDecking: Fixed.

src/shims/x86/sse42.rs
TDecking: Yes. All the references of "words" inside the comments have been replaced with "two-byte words", which hopefully clears things up.

src/shims/x86/sse42.rs
TDecking: Fixed.

src/shims/x86/sse42.rs
TDecking: Yes, these constats are bitmasks. This is now documented.

src/shims/x86/sse42.rs
TDecking: There are two length types at play here. One the one hand, there is the SIMD length (16 or 8) and the string length (how much of the input value is used). I've clarified the comment.

src/shims/x86/sse42.rs
TDecking: I've reordered the file.

src/shims/x86/sse42.rs
RalfJung: Okay so these intrinsics work for strings of length "up to 16"? (or 8, for the two-byte mode)

RalfJung: In word mode, I assume `len` is in units of words, not bytes? That would be good to clarify.

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,480 @@
...
+///     || |- Result Polarity. The result bits may be subjected to a bitwise complement
+///     ||    if these bits are set.
+///     ||
+///     ||- Output selection. This bit has two meanings depending on the instruction.
+///     |   If the instruction is generating a mask, it distinguishes between a bit mask
+///     |   and a byte mask. Otherwise it distinguishes between the most significand bit
+///     |   and the least significand bit when generating an index.
+///     |
+///     |- This bit is ignored. It is expected that this bit is set to zero, but it is
+///        not a requirement.
```
RalfJung, line 61: Thanks a ton for this super detailed comment!

src/shims/x86/sse42.rs
TDecking: Yes. For example, if you want to use use the substring functionality and search for the word "Rust", you'll need a way to specify that only the first four characters inside the argument are relevant and the remaining 12/4 are not.

src/shims/x86/sse42.rs
```diff
@@ -0,0 +1,480 @@
...
+                    _ => unreachable!(),
+                };
+
+                // Perform polynomial division modulo 2.
+                let mut dividend = (crc << bit_size) ^ (v << 32);
+                const POLYNOMIAL: u128 = 0x11EDC6F41;
+                let leading_zeros = dividend.leading_zeros();
+                while leading_zeros <= POLYNOMIAL.leading_zeros() {
+                    dividend ^= (POLYNOMIAL << POLYNOMIAL.leading_zeros()) >> leading_zeros;
+                }
```
RalfJung, line 465: This looks like dark magic to me.^^ Shifting by the number of leading zeroes of `(crc << bit_size) ^ (v << 32)`... now my head hurts. ;)

I assume this is a standard algorithm; is there a nice website we can link to that documents/explains these shifts?

src/shims/x86/sse42.rs
TDecking: You're welcome :)

src/shims/x86/sse42.rs
TDecking: The intrinsics guide goes into the previous steps and mentions them. _Why_ they are done I don't know either, for that my grasp on CRC is too weak.

As for the loop, its simple schoolbook division, but it's in GF(2), so addition/subtraction is done using a bitwise XOR.

src/shims/x86/sse42.rs
```diff
@@ -459,9 +460,8 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // Perform polynomial division modulo 2.
                 let mut dividend = (crc << bit_size) ^ (v << 32);
                 const POLYNOMIAL: u128 = 0x11EDC6F41;
-                let leading_zeros = dividend.leading_zeros();
-                while leading_zeros <= POLYNOMIAL.leading_zeros() {
-                    dividend ^= (POLYNOMIAL << POLYNOMIAL.leading_zeros()) >> leading_zeros;
+                while dividend.leading_zeros() <= POLYNOMIAL.leading_zeros() {
+                    dividend ^= (POLYNOMIAL << POLYNOMIAL.leading_zeros()) >> dividend.leading_zeros();
```
RalfJung: This diff is a functional change, right? Previously `leading_zeros` was a constant, not it is recomputed based on the current `dividend` each round.

Which one is right? Do the tests catch the wrong one?

src/shims/x86/sse42.rs
RalfJung: The intrinsic guide being [this](https://datatracker.ietf.org/doc/html/rfc3720#section-12.1)? It doesn't seem to say anything specific about this intrinsic though, just CRC in general.

> As for the loop, its simple schoolbook division, but it's in GF(2), so addition/subtraction is done using a bitwise XOR.

I believe that it is standard, but having a link to somewhere that explains the concrete algorithm, and in particular the loop condition involving `leading_zeros`, would still be good.

src/shims/x86/sse42.rs
TDecking: Yes. I tried to remove one call to `leading_zeros`, but the transformation was incorrect. I had to revert it.

src/shims/x86/sse42.rs
TDecking: I didn't find any satisfactory explanation online so I hope the my custom explanation is sufficient.

I also decided to extract the algorithm and exhaustively compare it with the machine results for the `_mm_crc32_u8` intrinsic. The algorithm is _definitely_ correct.

RalfJung: Oh, I thought you had copied the algorithm from somewhere and it'd be easier to link to that. I didn't realize you just implemented it based on the abstract algebraic spec yourself. :) Thanks for adding the comment!

RalfJung:
I think this is good to go... I can't actually check whether all these intrinsics behave the way they should, but I guess people will complain if they see wrong results. ;)

Please squash the commits.

TDecking:
@RalfJung done.

RalfJung:
Thanks a lot for the PR and for staying with us through all the rounds of review! :-)

@bors r+

PR #3625, opened 2024-05-23
Bugfix `MiriAllocBytes` to guarantee different addresses

Strophox:
Fix in `alloc_bytes.rs` following https://github.com/rust-lang/miri/pull/3526

Currently when an allocation of `size == 0` is requested we return a `std::ptr::without_provenance_mut(align)`, but this means returned `ptr`s may overlap, which breaks things.

### DIFF
--- a/src/alloc_bytes.rs
+++ b/src/alloc_bytes.rs
@@ -14,8 +14,7 @@ pub struct MiriAllocBytes {
     layout: alloc::Layout,
     /// Pointer to the allocation contents.
     /// Invariant:
-    /// * If `self.layout.size() == 0`, then `self.ptr` is some suitably aligned pointer
-    ///   that was allocated with the same layout but `size == 1`.
+    /// * If `self.layout.size() == 0`, then `self.ptr` was allocated with the equivalent layout with size 1.
     /// * Otherwise, `self.ptr` points to memory allocated with `self.layout`.
     ptr: *mut u8,
 }
@@ -30,6 +29,8 @@ impl Clone for MiriAllocBytes {
 
 impl Drop for MiriAllocBytes {
     fn drop(&mut self) {
+        // We have to reconstruct the actual layout used for allocation.
+        // (`Deref` relies on `size` so we can't just always set it to at least 1.)
         let alloc_layout = if self.layout.size() == 0 {
             Layout::from_size_align(1, self.layout.align()).unwrap()
         } else {
--- a/src/alloc_bytes.rs
+++ b/src/alloc_bytes.rs
@@ -13,10 +13,7 @@ pub struct MiriAllocBytes {
     /// Stored layout information about the allocation.
     layout: alloc::Layout,
     /// Pointer to the allocation contents.
-    /// Invariant:
-    /// * If `self.layout.size() == 0`, then `self.ptr` is some suitably aligned pointer
-    ///   without provenance (and no actual memory was allocated).
-    /// * Otherwise, `self.ptr` points to memory allocated with `self.layout`.
+    /// Invariant: `self.ptr` points to memory allocated with `self.layout`.
     ptr: *mut u8,
 }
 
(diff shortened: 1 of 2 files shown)


RalfJung:
> but this means returned ptrs may overlap

"overlap" isn't quite right, since arguably ranges with with size 0 can never overlap.

But two different allocations can end up at the same address, and that's a problem.

src/alloc_bytes.rs
```diff
@@ -58,25 +53,23 @@ impl std::ops::DerefMut for MiriAllocBytes {
...
-    /// `alloc_fn` is only used if `size != 0`.
-    /// Returns `Err(layout)` if the allocation function returns a `ptr` that is `ptr.is_null()`.
+    /// `alloc_fn` is only used with `size != 0`.
+    /// Returns `Err(layout)` if the allocation function returns a `ptr` where `ptr.is_null()`.
     fn alloc_with(
         size: usize,
         align: usize,
         alloc_fn: impl FnOnce(Layout) -> *mut u8,
     ) -> Result<MiriAllocBytes, Layout> {
+        // When size is 0 we allocate 1 byte anyway, so addresses don't possibly overlap.
```
RalfJung, line 63:
```suggestion
        // When size is 0 we allocate 1 byte anyway, to ensure each allocation has a unique address.
```

src/alloc_bytes.rs
```diff
@@ -13,10 +13,7 @@ pub struct MiriAllocBytes {
     /// Stored layout information about the allocation.
```
RalfJung: May be worth mentioning that the size here is never 0.

src/alloc_bytes.rs
```diff
@@ -30,10 +30,13 @@ impl Clone for MiriAllocBytes {
 
 impl Drop for MiriAllocBytes {
     fn drop(&mut self) {
-        if self.layout.size() != 0 {
-            // SAFETY: Invariant, `self.ptr` points to memory allocated with `self.layout`.
-            unsafe { alloc::dealloc(self.ptr, self.layout) }
-        }
+        let alloc_layout = if self.layout.size() == 0 {
```
RalfJung, line 33: Can't you store the changed layout with size 1 in the `layout`? Then you don't need to duplicate the logic here.

src/alloc_bytes.rs
Strophox: I had to change some things around but the relevant part of this comment is in the function documentation now.

src/alloc_bytes.rs
Strophox: I initially did it where if `size == 0` was requested I simply changed `layout.size() == 1`, but this surprisingly failed all tests:
Here I thought it would be better to ensure *allocating* (and deallocating) at least one byte, but leave the remaining part of a `MiriAllocBytes` object with `size == 0` intact.

RalfJung:
> but this surprisingly failed all tests:

Why did they fail? Setting the size to 1 should work and avoids duplicating logic so please figure out why it did not work.

src/alloc_bytes.rs
```diff
@@ -15,7 +15,7 @@ pub struct MiriAllocBytes {
     /// Pointer to the allocation contents.
     /// Invariant:
     /// * If `self.layout.size() == 0`, then `self.ptr` is some suitably aligned pointer
```
RalfJung, line 17:
```suggestion
    /// * If `self.layout.size() == 0`, then `self.ptr` was allocated with the equivalent layout with size 1.
```

src/alloc_bytes.rs
```diff
@@ -15,7 +15,7 @@ pub struct MiriAllocBytes {
...
-    ///   without provenance (and no actual memory was allocated).
+    ///   that was allocated with the same layout but `size == 1`.
```
RalfJung, line 18:
```suggestion
```

PR #3629, opened 2024-05-24
solarish platform: add support for available-parallelism.

devnexen:


src/shims/unix/solarish/foreign_items.rs
```diff
@@ -69,6 +69,18 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                let [_pset, _type, cpus, _list] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                // We do not need to handle the current process cpu mask, available_parallelism
+                // implementation pass null anyway. We only care for the number of
+                // cpus.
+                // https://docs.oracle.com/cd/E88353_01/html/E37841/pset-info-2.html
+                let cpus = this.deref_pointer(cpus)?;
+                this.write_scalar(Scalar::from_u32(this.machine.num_cpus), &cpus)?;
+                this.write_null(dest)?;
+            }
```
RalfJung, line 82: This shim could be called by other code outside of std, so it should do something reasonable also for the other parameters -- or it should be only used when `frame_in_std` is true.

saethlin: In general I would prefer if we didn't do the `frame_in_std` trick here. If someone gets an unsupported error in their code but their implementation is a copy of the standard library code, perhaps because it is a polyfill... That seems unnecessarily confusing. Can we `throw_unsup_format` except for the calls we need for the standard library? It sounds to me like we properly implement some subset of calls, unlike the `frame_in_std` we use for thread startup code.

RalfJung: Seems like we'd have to check that type and list are null. And pset must be PS_MYID.

RalfJung: Though FWIW `frame_in_std` is not just used for pre-init code. It also guards things like `getpwuid_r` as well; as a whole bunch of stuff on Windows. I agree it's potentially confusing, but IMO it's still better than no support at all.

saethlin: I agree that it's better than no support at all. I'll look at the other calls later to see if we're consistent with what I asked for above.

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -69,6 +69,31 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+            "pset_info" if this.frame_in_std() => {
+                let [pset, tpe, cpus, list] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                // We do not need to handle the current process cpu mask, available_parallelism
+                // implementation pass null anyway. We only care for the number of
+                // cpus.
+                // https://docs.oracle.com/cd/E88353_01/html/E37841/pset-info-2.html
+
+                if this.eval_libc_i32("PS_MYID") != this.read_scalar(pset)?.to_i32()? {
+                    throw_unsup_format!("Miri supports pset_info only with pset==PS_MYID");
```
RalfJung, line 81:
```suggestion
                    throw_unsup_format!("pset_info is only supported with pset==PS_MYID");
```

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -69,6 +69,31 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 this.write_null(dest)?;
             }
 
+            "pset_info" if this.frame_in_std() => {
```
RalfJung, line 72:
```suggestion
            "pset_info" => {
```
Now that you implemented all the checks, this can be always enabled. :)

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -69,6 +69,31 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                // implementation pass null anyway. We only care for the number of
+                // cpus.
+                // https://docs.oracle.com/cd/E88353_01/html/E37841/pset-info-2.html
+
+                if this.eval_libc_i32("PS_MYID") != this.read_scalar(pset)?.to_i32()? {
+                    throw_unsup_format!("Miri supports pset_info only with pset==PS_MYID");
+                }
+
+                if !this.ptr_is_null(this.read_pointer(tpe)?)? {
+                    throw_unsup_format!("Miri supports pset_info only with type==NULL");
```
RalfJung, line 85:
```suggestion
                    throw_unsup_format!("pset_info is only supported with type==NULL");
```

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -69,6 +69,31 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                }
+
+                if !this.ptr_is_null(this.read_pointer(list)?)? {
+                    throw_unsup_format!("Miri supports pset_info only with list==NULL");
```
RalfJung, line 89:
```suggestion
                    throw_unsup_format!("pset_info is only supported with list==NULL");
```

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -69,6 +69,31 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
                 this.write_null(dest)?;
             }
 
+            "pset_info" if this.frame_in_std() => {
+                let [pset, tpe, cpus, list] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
```
RalfJung, line 74: Please write this the way we write all our shims: with a line like `let pset = this.read_scalar(pset)?.to_i32()?;` etc for each argument. The "turning arguments into a typed representation" should be separate from all further processing.

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -69,6 +69,36 @@ pub trait EvalContextExt<'mir, 'tcx: 'mir>: crate::MiriInterpCxExt<'mir, 'tcx> {
...
+                // implementation pass null anyway. We only care for the number of
+                // cpus.
+                // https://docs.oracle.com/cd/E88353_01/html/E37841/pset-info-2.html
+
+                let ps_myid = this.eval_libc_i32("PS_MYID");
+                let pset = this.read_scalar(pset)?.to_i32()?;
+                let tpe = this.read_pointer(tpe)?;
+                let list = this.read_pointer(list)?;
+
+                if ps_myid != pset {
```
RalfJung, line 85:
```suggestion
                let pset = this.read_scalar(pset)?.to_i32()?;
                let tpe = this.read_pointer(tpe)?;
                let list = this.read_pointer(list)?;

                let ps_myid = this.eval_libc_i32("PS_MYID");
                if ps_myid != pset {
```

RalfJung:
Thanks! I've fixed a typo in the PR title as well, should be good to go now. :)
@bors r+

Issue #3637, opened 2024-05-27
Big `Vec::try_reserve` OOMs Miri


DaniPopes:
On both GitHub CI default linux runner (should have 16GB RAM), and locally (also linux, 64GB RAM), the following test will run out of memory in Miri, but panic in normal `cargo test`:

```rust
#[test]
fn try_allocate_a_lot() {
    let mut v = Vec::<u8>::new();
    v.try_reserve(128 * 1024 * 1024 * 1024).unwrap();
    let _ = v;
}
```

- `cargo test` panics: ``called `Result::unwrap()` on an `Err` value: TryReserveError { kind: AllocError { layout: Layout { size: 137438953472, align: 1 (1 << 0) }, non_exhaustive: () } }``
- `cargo miri test` will try allocating and get killed for OOM, CTRL+C does not stop it

Apologies if this has been reported before, I could not find another issue for this.

bjorn3:
> CTRL+C does not stop it

Does a second ctrl-c stop it? Also since https://github.com/rust-lang/rust/pull/125523 it should exit at most 100ms after a ctrl-c.

RalfJung:
Yeah this is pretty much known, see https://github.com/rust-lang/miri/issues/613. We don't make any attempt at recovering from allocation failure.

RalfJung:
However, what I would expect to happen is that Miri itself aborts due to OOM. Maybe the reason this does not happen is that Miri uses jemalloc under the hood, whereas `cargo test` will use the system allocator. So maybe `malloc` fails immediately on an allocation of that size but jemalloc tries anyway, and then your system grinds to a halt due to swapping.

RalfJung:
FWIW when I try to run that function locally, the program gets swiftly killed. So I can't reproduce the hang you seem to be seeing.

DaniPopes:
> Does a second ctrl-c stop it? Also since https://github.com/rust-lang/rust/pull/125523 it should exit at most 100ms after a ctrl-c.

I was on `1.80.0-nightly (1ba35e9bb 2024-05-25)`, CTRL+C would exit cargo but miri would still be in the background allocating, forcing me to kill it manually.
I just updated to `rustc 1.80.0-nightly (bdbbb6c6a 2024-05-26)` and now it does exit fully.

saethlin:
[...]

PR #3638, opened 2024-05-27
Add a benchmark for creating large uninit allocations

saethlin:
Extracted from https://github.com/rust-lang/miri/issues/3637

I used this program to confirm that https://github.com/rust-lang/rust/pull/125633 has the desired effect.


saethlin:
That is strange. The only docs I can find suggest this should fit comfortably on a GitHub-hosted runner: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

RalfJung:
Note that on Windows we're using a 32bit target as the host. Not sure how Windows manages the address space there, maybe it keeps half of it for the kernel so there's only 2GB total for the program?

saethlin:
Right. I was trying to avoid doing that because the effect size shrinks, but of course a little loop fixes that.

RalfJung:
How long does the benchmark take (with the fix)?

RalfJung:
https://github.com/rust-lang/rust/pull/125633 landed, but a `./miri rustc-pull` will be needed before we have it available in the repo.

saethlin:
> How long does the benchmark take (with the fix)?

200 ms per iteration with the fix. Without the fix, it's 800 ms per iteration. So I think it would be fair to land this without the fix, because it's already one of our faster benchmarks.

RalfJung:
Makes sense, thanks!
@bors r+

PR #3640, opened 2024-05-28
add support for `pclmulqdq` intrinsic

folkertdev:
This instruction is required in fast implementations of the crc32 checksum algorithm, and used in the https://crates.io/crates/crc32fast and https://crates.io/crates/zlib-rs crates.

Some questions from my side

- is my method for decomposing a `__m128i` into two separate `i64` values allright?

### DIFF
--- a/src/shims/x86/mod.rs
+++ b/src/shims/x86/mod.rs
@@ -105,6 +105,13 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 }
             }
 
+            "pclmulqdq" => {
+                let [left, right, imm] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                pclmulqdq(this, left, right, imm, dest)?;
+            }
+
             name if name.starts_with("sse.") => {
                 return sse::EvalContextExt::emulate_x86_sse_intrinsic(
                     this, link_name, abi, args, dest,
@@ -1129,6 +1136,68 @@ fn pmulhrsw<'tcx>(
 
         this.write_scalar(Scalar::from_i16(res), &dest)?;
     }
+
+    Ok(())
+}
+
+/// Perform a carry-less multiplication of two 64-bit integers, selected from left and right according to imm8,
+/// and store the results in dst.
+///
+/// Left and right are both vectors of type 2 x i64. Only bits 0 and 4 of imm8 matter;
+/// they select the element of left and right, respectively.
+///
+/// <https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_clmulepi64_si128>
+fn pclmulqdq<'tcx>(
+    this: &mut crate::MiriInterpCx<'tcx>,
+    left: &OpTy<'tcx>,
+    right: &OpTy<'tcx>,
+    imm8: &OpTy<'tcx>,
+    dest: &MPlaceTy<'tcx>,
+) -> InterpResult<'tcx, ()> {
+    assert_eq!(left.layout, right.layout);
+    assert_eq!(left.layout.size, dest.layout.size);
+
+    // Transmute to `[u64; 2]`
+
+    let array_layout = this.layout_of(Ty::new_array(this.tcx.tcx, this.tcx.types.u64, 2))?;
+    let left = left.transmute(array_layout, this)?;
+    let right = right.transmute(array_layout, this)?;
+    let dest = dest.transmute(array_layout, this)?;
+
+    let imm8 = this.read_scalar(imm8)?.to_u8()?;
+
+    // select the 64-bit integer from left that the user specified (low or high)
+    let index = if (imm8 & 0x01) == 0 { 0 } else { 1 };
+    let left = this.read_scalar(&this.project_index(&left, index)?)?.to_u64()?;
+
+    // select the 64-bit integer from right that the user specified (low or high)
+    let index = if (imm8 & 0x10) == 0 { 0 } else { 1 };
+    let right = this.read_scalar(&this.project_index(&right, index)?)?.to_u64()?;
+
+    // Perform carry-less multiplication
+    //
+    // This operation is like long multiplication, but ignores all carries.
+    // That idea corresponds to the xor operator, which is used in the implementation.
+    //
+    // Wikipedia has an example https://en.wikipedia.org/wiki/Carry-less_product#Example
+    let mut result: u128 = 0;
+
+    for i in 0..64 {
+        // if the i-th bit in right is set
+        if (right & (1 << i)) != 0 {
+            // xor result with `left` shifted to the left by i positions
+            result ^= (left as u128) << i;
+        }
+    }
+
+    let result_low = (result & 0xFFFF_FFFF_FFFF_FFFF) as u64;
+    let result_high = (result >> 64) as u64;
+
+    let dest_low = this.project_index(&dest, 0)?;
+    this.write_scalar(Scalar::from_u64(result_low), &dest_low)?;
+
+    let dest_high = this.project_index(&dest, 1)?;
+    this.write_scalar(Scalar::from_u64(result_high), &dest_high)?;
 
     Ok(())
 }
--- a/src/shims/x86/mod.rs
+++ b/src/shims/x86/mod.rs
@@ -1140,15 +1140,15 @@ fn pmulhrsw<'tcx>(
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


src/shims/x86/mod.rs
```diff
@@ -146,6 +146,13 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 );
             }
 
+            "pclmulqdq" => {
```
RalfJung, line 149: Please move this up to the other concrete intrinsic implementations. Down here we only dispatch entire intrinsic groups to another file.

src/shims/x86/mod.rs
```diff
@@ -1133,6 +1140,58 @@ fn pmulhrsw<'tcx>(
     Ok(())
 }
 
+/// Perform a carry-less multiplication of two 64-bit integers, selected from a and b according to imm8, and store the results in dst.
```
RalfJung, line 1143:
```suggestion
/// a, b are both vectors of 2 x i64.
/// Perform a carry-less multiplication of two 64-bit integers, selected from a and b according to imm8, and store the results in dst.
```

src/shims/x86/mod.rs
```diff
@@ -1133,6 +1140,58 @@ fn pmulhrsw<'tcx>(
...
+    let imm8 = this.read_scalar(imm8)?.to_u8()?;
+
+    let left_low = this.read_scalar(&this.project_index(&left, 0)?)?.to_u64()?;
+    let left_high = this.read_scalar(&this.project_index(&left, 1)?)?.to_u64()?;
+
+    let right_low = this.read_scalar(&this.project_index(&right, 0)?)?.to_u64()?;
+    let right_high = this.read_scalar(&this.project_index(&right, 1)?)?.to_u64()?;
+
+    let temp1 = if (imm8 & 0x01) == 0 { left_low } else { left_high };
+    let temp2 = if (imm8 & 0x10) == 0 { right_low } else { right_high };
```
RalfJung, line 1172: Is it possible to pick more descriptive names here?

src/shims/x86/mod.rs
```diff
@@ -1133,6 +1140,58 @@ fn pmulhrsw<'tcx>(
...
+
+    // Perform carry-less multiplication
```
RalfJung, line 1174: What exactly is "carry-less multiplication", and why does it involve XOR...?

I've read the linked docs but that looks even more complicated and not even obviously equivalent -- there are 2 loops here, and a funny `TEMP1[0] and TEMP2[i]` initialization.

src/shims/x86/mod.rs
```diff
@@ -1133,6 +1140,58 @@ fn pmulhrsw<'tcx>(
...
+    let mut result: u128 = 0;
+
+    for i in 0..64 {
+        if (temp2 & (1 << i)) != 0 {
+            result ^= (temp1 as u128) << i;
+        }
+    }
+
+    let low_result = (result & 0xFFFF_FFFF_FFFF_FFFF) as u64;
+    let high_result = (result >> 64) as u64;
```
RalfJung, line 1184: Naming is a bit consistent; matching the above it should be "result_{low,high}"

tests/pass/shims/x86/intrinsics-x86-pclmulqdq.rs
```diff
@@ -0,0 +1,48 @@
...
+    assert!(is_x86_feature_detected!("pclmulqdq"));
+
+    let a = (0x7fffffffffffffff, 0x4317e40ab4ddcf05);
+    let b = (0xdd358416f52ecd34, 0x633d11cc638ca16b);
+
+    unsafe {
+        assert_eq!(clmulepi64_si128::<0x00>(a, b), (13036940098130298092, 2704901987789626761));
+        assert_eq!(clmulepi64_si128::<0x01>(a, b), (6707488474444649956, 3901733953304450635));
+        assert_eq!(clmulepi64_si128::<0x10>(a, b), (11607166829323378905, 1191897396234301548));
+        assert_eq!(clmulepi64_si128::<0x11>(a, b), (7731954893213347271, 1760130762532070957));
```
RalfJung, line 27: What happens if the imm8 is not one of these 4 values?

RalfJung (commented):
Thanks for the PR!

> is my method for decomposing a __m128i into two separate i64 values allright?

A transmute is one option; just accessing the two fields would also work but would rely on the way that type is declared. I think we already have similar transmutes for other x86 intrinsics, so this is fine.

src/shims/x86/mod.rs
RalfJung: That's a good reference, can you add that to the code?

tests/pass/shims/x86/intrinsics-x86-pclmulqdq.rs
RalfJung: Does intel *guarantee* that e.g. 0x01 and 0x03 behave the same?
Or should we make Miri halt execution when imm8 is not one of the 4 values here?

folkertdev: well based on [the docs](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_clmulepi64_si128&ig_expand=754) it just checks for whether the bit at index 0 and 4 is set, corresponding with the bitwise and above

```
IF (imm8[0] == 0)
	TEMP1 := a[63:0]
ELSE
	TEMP1 := a[127:64]
FI 
IF (imm8[4] == 0)
	TEMP2 := b[63:0]
ELSE 
	TEMP2 := b[127:64]
FI
```

using some other value like `0x0F` also works fine on the rust playground anyway.

RalfJung: Okay, thanks for checking.

src/shims/x86/mod.rs
```diff
@@ -1133,6 +1140,65 @@ fn pmulhrsw<'tcx>(
     Ok(())
 }
 
+/// a, b are both vectors of 2 x i64.
+/// Perform a carry-less multiplication of two 64-bit integers, selected from a and b according to imm8, and store the results in dst.
```
RalfJung, line 1144:
```suggestion
/// Perform a carry-less multiplication of two 64-bit integers, selected from a and b according to imm8, and store the results in dst.
/// Only bits 0 and 4 of imm8 matter; they select the element of a and b, respectively.
```

src/shims/x86/mod.rs
```diff
@@ -1133,6 +1140,65 @@ fn pmulhrsw<'tcx>(
     Ok(())
 }
 
+/// a, b are both vectors of 2 x i64.
```
RalfJung, line 1143: The function now says left/right; please update the comment to match.

RalfJung:
In future please avoid force-pushing (except when there are conflicts that need resolving) until the PR is done. Github is pretty bad at keeping track of force-pushes, making it basically impossible to tell what changed since my last review.

src/shims/x86/mod.rs
```diff
@@ -1133,6 +1140,68 @@ fn pmulhrsw<'tcx>(
...
 
+/// Perform a carry-less multiplication of two 64-bit integers, selected from left and right according to imm8,
+/// and store the results in dst.
+///
+/// Left and right are both vectors of type 2 x i64. Only bits 0 and 4 of imm8 matter;
+/// they select the element of left and right, respectively.
+///
+/// <https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_clmulepi64_si128>
+fn pclmulqdq<'tcx>(
+    this: &mut crate::MiriInterpCx<'tcx>,
```
RalfJung, line 1151:
```suggestion
    this: &mut MiriInterpCx<'tcx>,
```

src/shims/x86/mod.rs
```diff
@@ -1133,6 +1140,68 @@ fn pmulhrsw<'tcx>(
     Ok(())
 }
 
+/// Perform a carry-less multiplication of two 64-bit integers, selected from left and right according to imm8,
+/// and store the results in dst.
+///
+/// Left and right are both vectors of type 2 x i64. Only bits 0 and 4 of imm8 matter;
+/// they select the element of left and right, respectively.
```
RalfJung, line 1147:
```suggestion
/// Perform a carry-less multiplication of two 64-bit integers, selected from `left` and `right` according to `imm8`,
/// and store the results in `dst`.
///
/// `left` and `right` are both vectors of type 2 x i64. Only bits 0 and 4 of `imm8` matter;
/// they select the element of `left` and `right`, respectively.
```

RalfJung:
Thanks. :-)

@bors r+

PR #3649, opened 2024-06-05
Draft: Create _umtx_op shim for FreeBSD

b-ncMN:
This PR lays the basic works for the _umtx_op shim for freebsd

I have implemented this by taking inspiration from the linux futex shim, so it may not be perfect and I am looking forward to criticism.

I will continue that on later commits because the tests do not currently pass

### DIFF
--- a/src/shims/unix/freebsd/sync.rs
+++ b/src/shims/unix/freebsd/sync.rs
@@ -1,16 +1,27 @@
 use crate::*;
+use rustc_middle::ty::Ty;
+use rustc_middle::ty::Mutability;
+use rustc_middle::ty::layout::LayoutOf;
 
 pub fn futex<'tcx>(
     this: &mut MiriInterpCx<'tcx>,
     args: &[OpTy<'tcx>],
     _dest: &MPlaceTy<'tcx>,
 ) -> InterpResult<'tcx> {
+    //Checking for the minimal amount of arguments required to proceed with
+    if args.len() < 3 {
+        throw_ub_format!(
+            "incorrect number of arguments for `futex` syscall: got {}, expected at least 3",
+            args.len()
+        );
+    }
+
     //void *obj
-    let _obj = this.read_pointer(&args[0])?;
+    let obj = this.read_pointer(&args[0])?;
     //int op
     let op = this.read_scalar(&args[1])?.to_i32()?;
     //u_long val
-    let _val = this.read_scalar(&args[2])?.to_u64()?;
+    let val = this.read_scalar(&args[2])?.to_i64()?;
     //void *uaddr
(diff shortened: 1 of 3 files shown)


RalfJung:
Just one quick nit because I saw the notification, the PR description and commit message could be a bit more descriptive. :) They should mention `_umtx_op`, and probably FreeBSD.

b-ncMN:
I think all of these commits will be merged into one before pulling this in main.
I tried to keep every commits as clean as possible. But it seems I am stuck on a clippy warning on HEAD, is it worth for me to fix it in order to keep the tree clean or is that not really a problem? Because further commit will override that faulty code and it will be all in one commit after.

RalfJung:
When you mark this as ready for review, CI must be green. We won't land code that has clippy warnings.

For intermediate commits it doesn't matter.

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,57 @@
...
+    let _uaddr2 = this.read_pointer(&args[4])?;
+
+    let umtx_op_wait_uint_private = this.eval_libc_i32("UMTX_OP_WAIT_UINT_PRIVATE");
+    let umtx_op_wake_private = this.eval_libc_i32("UMTX_OP_WAKE_PRIVATE");
+    let umtx_op_nwake_private = this.eval_libc_i32("UMTX_OP_NWAKE_PRIVATE");
+
+    let umtx_op_wait = this.eval_libc_i32("UMTX_OP_WAIT");
+
+    //FUTEX private operations are not supported by miri
+    match op & !umtx_op_wait_uint_private & !umtx_op_wake_private & !umtx_op_nwake_private {
```
tavianator, line 37: Hmm?  `UMTX_OP_WAIT_UINT_PRIVATE` etc are not bitflags: https://github.com/freebsd/freebsd-src/blob/d19851f002862a5510bf31fae4083fab979258be/sys/sys/umtx.h#L76-L105

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,57 @@
...
+        umtx_op_wait => {
```
tavianator, line 38: Since `umtx_op_wait` is not a constant, I think this is the same as

```rust
match foo {
    bar => todo!(),
}
```

i.e. it matches anything.

RalfJung:
@b-ncMN what is your plan for this? If this is unlikely to be ready for at least some early feedback within the next 2 weeks, I'd prefer to close the PR so that we can keep the PR list for things that are actively being worked on. If you can out of time, you can always reopen the PR later. :)

RalfJung:
I'll close the PR due to inactivity, feel free to pick it up again any time. :)

PR #3654, opened 2024-06-07
Fix stage in contributing

Nilstrieb:
0 errors now because they were moved by 1.


RalfJung:
Stage 1 is the default for "test" so there it can be removed entirely.

Nilstrieb:
makes sense

RalfJung:
Thanks!
@bors r+

PR #3655, opened 2024-06-08
portable-simd: add test for non-power-of-2 bitmask

RalfJung:
@calebzulawski is that the intended behavior? Specifically for arrays, the bitmask `[1, 0, 0, 1, 0, 0, 1, 0, 1, 0]` becomes
- `[0b01001001, 0b01]` on little endian
- `[0b10010010, 0b10]` on big endian

### DIFF
--- a/tests/pass/intrinsics/portable-simd.rs
+++ b/tests/pass/intrinsics/portable-simd.rs
@@ -1,5 +1,5 @@
 //@compile-flags: -Zmiri-strict-provenance
-#![feature(portable_simd, adt_const_params, core_intrinsics)]
+#![feature(portable_simd, adt_const_params, core_intrinsics, repr_simd)]
 #![allow(incomplete_features, internal_features)]
 use std::intrinsics::simd as intrinsics;
 use std::ptr;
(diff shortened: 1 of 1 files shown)


calebzulawski:
Yes, I think that's expected (we reverse the bitmasks in std::simd)

FYI, we actually just removed use of that variation of the intrinsic in https://github.com/rust-lang/portable-simd/pull/423

RalfJung:
> FYI, we actually just removed use of that variation of the intrinsic in https://github.com/rust-lang/portable-simd/pull/423

The intrinsic is still called in other places though. In which sense is it "removed"?

RalfJung:
> Yes, I think that's expected (we reverse the bitmasks in std::simd)

Great, thanks. :)
@bors r+

calebzulawski:
> > FYI, we actually just removed use of that variation of the intrinsic in [rust-lang/portable-simd#423](https://github.com/rust-lang/portable-simd/pull/423)
> 
> The intrinsic is still called in other places though. In which sense is it "removed"?

We still use the intrinsic, but currently (not yet synced to rust-lang/rust) we only use integer return types

RalfJung:
Ah, this seems to fail on big-endian...

> We still use the intrinsic, but currently (not yet synced to rust-lang/rust) we only use integer return types

I see. If the array support ever gets removed entirely, please ping me.

RalfJung:
The docs say

> No matter whether the output is an array or an unsigned integer, it is treated as a single contiguous list of bits.

On big-endian, the bitlist `0b1001001010u16` corresponds to the array `[0b10, 0b01001010]`.
So either the docs are wrong or my examples listed above are wrong.

Maybe that explains why you had trouble with this in the non-power-of-2 PR?

PR #3661, opened 2024-06-09
Follow up PR for eventfd shim

tiif:
Follow up of https://github.com/rust-lang/miri/pull/3650

### DIFF
--- a/src/shims/unix/linux/eventfd.rs
+++ b/src/shims/unix/linux/eventfd.rs
@@ -1,6 +1,7 @@
 //! Linux `eventfd` implementation.
 use std::io;
 use std::io::{Error, ErrorKind};
+use std::mem;
 
 use rustc_target::abi::Endian;
 
@@ -9,8 +10,8 @@ use crate::{concurrency::VClock, *};
 
 use self::shims::unix::fd::FileDescriptor;
 
-/// Minimum size of u8 array to hold u64 value.
-const U64_MIN_ARRAY_SIZE: usize = 8;
+// We'll only do reads and writes in chunks of size u64.
+const U64_ARRAY_SIZE: usize = mem::size_of::<u64>();
 
 /// Maximum value that the eventfd counter can hold.
 const MAX_COUNTER: u64 = u64::MAX - 1;
@@ -51,7 +52,7 @@ impl FileDescription for Event {
         ecx: &mut MiriInterpCx<'tcx>,
     ) -> InterpResult<'tcx, io::Result<usize>> {
         // Check the size of slice, and return error only if the size of the slice < 8.
-        let Some(bytes) = bytes.first_chunk_mut::<U64_MIN_ARRAY_SIZE>() else {
+        let Some(bytes) = bytes.first_chunk_mut::<U64_ARRAY_SIZE>() else {
             return Ok(Err(Error::from(ErrorKind::InvalidInput)));
         };
         // Block when counter == 0.
@@ -63,7 +64,7 @@ impl FileDescription for Event {
                 throw_unsup_format!("eventfd: blocking is unsupported");
             }
         } else {
-            // Prevent false alarm in data race detection when doing synchronisation via eventfd.
+            // Synchronize with all prior `write` calls to this FD.
             ecx.acquire_clock(&self.clock);
             // Return the counter in the host endianness using the buffer provided by caller.
             *bytes = match ecx.tcx.sess.target.endian {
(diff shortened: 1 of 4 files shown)


src/shims/unix/linux/eventfd.rs
```diff
@@ -9,8 +10,8 @@ use crate::{concurrency::VClock, *};
 
 use self::shims::unix::fd::FileDescriptor;
 
-/// Minimum size of u8 array to hold u64 value.
-const U64_MIN_ARRAY_SIZE: usize = 8;
+// Size of u8 array to hold u64.
+const U64_ARRAY_SIZE: usize = mem::size_of::<u64>();
```
tiif, line 14: Is it ok to leave this as const? ``bytes.first_chunk_mut::<U64_ARRAY_SIZE>()`` can only take in const for ``U64_ARRAY_SIZE``.

RalfJung: Yeah a const is fine.

src/shims/unix/linux/eventfd.rs
```diff
@@ -9,8 +10,8 @@ use crate::{concurrency::VClock, *};
 
 use self::shims::unix::fd::FileDescriptor;
 
-/// Minimum size of u8 array to hold u64 value.
-const U64_MIN_ARRAY_SIZE: usize = 8;
+// Size of u8 array to hold u64.
```
RalfJung, line 13:
```suggestion
// We'll only do reads and writes in chunks of size u64.
```

src/shims/unix/linux/eventfd.rs
```diff
@@ -63,15 +64,15 @@ impl FileDescription for Event {
                 throw_unsup_format!("eventfd: blocking is unsupported");
             }
         } else {
-            // Prevent false alarm in data race detection when doing synchronisation via eventfd.
+            // Future `read` calls will synchronize with this write, so update the FD clock.
```
RalfJung, line 67: This is a "read", I think you mixed up the 2 comments?
```suggestion
            // Synchronize with past `write` calls.
```

src/shims/unix/linux/eventfd.rs
```diff
@@ -110,8 +111,10 @@ impl FileDescription for Event {
         // Else, block.
         match self.counter.checked_add(num) {
             Some(new_count @ 0..=MAX_COUNTER) => {
-                // Prevent false alarm in data race detection when doing synchronisation via eventfd.
-                self.clock.join(&ecx.release_clock().unwrap());
+                // Synchronize with all prior `write` calls to this FD.
```
RalfJung, line 114:
```suggestion
                // Future `read` calls will synchronize with this write, so update the FD clock.
```

src/shims/unix/linux/eventfd.rs
```diff
@@ -163,19 +166,21 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             flags &= !efd_nonblock;
             is_nonblock = true;
         }
         if flags != 0 {
-            let einval = this.eval_libc("EINVAL");
-            this.set_last_error(einval)?;
-            return Ok(Scalar::from_i32(-1));
+            throw_unsup_format!(
+                "eventfd: Flags other than EFD_CLOEXEC, EFD_NONBLOCK \
+            and EFD_SEMAPHORE is not supported."
```
RalfJung, line 182: EFD_SEMAPHORE is also not supported...
```suggestion
                "eventfd: encountered unknown unsupported flags {:#x}"
```

src/shims/unix/linux/eventfd.rs
tiif: Ah yes, thanks for catching this. It is likely that my eyes parse the word ``read`` and decided to put it in ``read`` function ><

src/shims/unix/linux/eventfd.rs
tiif: Oh yea this gave the illusion of we actually support EFD_SEMAPHORE when in fact, they will hit the unsup EFD_SEMAPHORE branch.

RalfJung:
Looks good, thanks. :)
There's no need to rebase when there are no conflicts, and github is terrible at dealing with foce-pushes, so please only do that when needed.

Please squash.

RalfJung:
Hm, that force-push has a non-empty diff? Did you do squash+rebase?

RalfJung:
PR looks the same though.
For reviewing it's easier if you don't do rebase unless needed -- including for the final squash.
(One day maybe github can do git-range-diff, but until then we'll all spend a lot of time working around this...)

@bors r+

tiif:
Ah I see. I did rebase on upstream before squash, will prevent doing that next time.

Issue #3546, opened 2024-05-03
Add support for many-seeds mode to `cargo miri`


RalfJung:
We have `./miri many-seeds` for our own development and testing purposes, but this could also be useful for end-users. I am imagining something like `cargo miri --many-seeds=0..256 test ...` or so, which would run the test 256 times with different seeds.

PR #3672, opened 2024-06-13
cargo miri: add support for '--many-seeds'

RalfJung:
to run the program / tests many times with different seeds: `cargo miri run --many-seeds` / `cargo miri test --many-seeds`.

@rust-lang/miri any opinion on the flag name here? Should it be `-Zmiri-many-seeds` or is `--many-seeds` fine?

Fixes https://github.com/rust-lang/miri/issues/3546

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,21 @@ platform. For example `cargo miri test --target s390x-unknown-linux-gnu`
 will run your test suite on a big-endian target, which is useful for testing
 endian-sensitive code.
 
+### Testing multiple different executions
+
+Certain parts of the execution are picked randomly by Miri, such as the exact base address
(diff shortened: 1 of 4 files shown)


saethlin:
I think `--many-seeds` is better. Honestly I don't see any reason to claim this is unstable/experimental, we've been doing this many seeds thing for years and I think it has proven its value.

RalfJung:
All right, let's land this then. We can always adjust this later.

@bors r+

ibraheemdev:
`--many-seeds` doesn't seem to run integration tests, was this intentional?

RalfJung:
I am not sure what you mean. Test selection is independent of this flag.

Please file an issue with reproducible instructions.

PR #3674, opened 2024-06-15
Implement LLVM x86 bmi intrinsics

TDecking:
This implements the intrinsics for both the bmi1 and bmi2 ISA extensions. All of these intrinsics live inside the same namespace as far as LLVM is concerned, which is why it is arguably better to bundle the implementations of these two extensions.

### DIFF
--- a/src/shims/x86/bmi.rs
+++ b/src/shims/x86/bmi.rs
@@ -1,0 +1,108 @@
+use rustc_span::Symbol;
+use rustc_target::spec::abi::Abi;
+
+use crate::*;
+
+impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
+pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    fn emulate_x86_bmi_intrinsic(
+        &mut self,
+        link_name: Symbol,
+        abi: Abi,
+        args: &[OpTy<'tcx>],
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx, EmulateItemResult> {
+        let this = self.eval_context_mut();
+
+        // Prefix should have already been checked.
+        let unprefixed_name = link_name.as_str().strip_prefix("llvm.x86.bmi.").unwrap();
+
+        // The intrinsics are suffixed with the bit size of their operands.
+        let (is_64_bit, unprefixed_name) = if unprefixed_name.ends_with("64") {
+            (true, unprefixed_name.strip_suffix(".64").unwrap_or(""))
+        } else {
+            (false, unprefixed_name.strip_suffix(".32").unwrap_or(""))
+        };
+
+        // All intrinsics of the "bmi" namespace belong to the "bmi2" ISA extension.
+        // The exception is "bextr", which belongs to "bmi1".
+        let target_feature = if unprefixed_name == "bextr" { "bmi1" } else { "bmi2" };
+        this.expect_target_feature_for_intrinsic(link_name, target_feature)?;
+
+        if is_64_bit && this.tcx.sess.target.arch != "x86_64" {
+            return Ok(EmulateItemResult::NotSupported);
+        }
+
+        let [left, right] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+        let left = this.read_scalar(left)?;
+        let right = this.read_scalar(right)?;
+
+        let left = if is_64_bit { left.to_u64()? } else { u64::from(left.to_u32()?) };
+        let right = if is_64_bit { right.to_u64()? } else { u64::from(right.to_u32()?) };
+
+        let result = match unprefixed_name {
+            // Extract a contigous range of bits from an unsigned integer.
+            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_bextr_u32
+            "bextr" => {
+                let start = u32::try_from(right & 0xff).unwrap();
+                let len = u32::try_from((right >> 8) & 0xff).unwrap();
+                let shifted = left.checked_shr(start).unwrap_or(0);
+                // Keep the `len` lowest bits of `shifted`, or all bits if `len` is too big.
+                if len >= 64 { shifted } else { shifted & 1u64.wrapping_shl(len).wrapping_sub(1) }
+            }
+            // Create a copy of an unsigned integer with bits above a certain index cleared.
+            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_bzhi_u32
+            "bzhi" => {
+                let index = u32::try_from(right & 0xff).unwrap();
+                // Keep the `index` lowest bits of `left`, or all bits if `index` is too big.
+                if index >= 64 { left } else { left & 1u64.wrapping_shl(index).wrapping_sub(1) }
+            }
+            // Extract bit values of an unsigned integer at positions marked by a mask.
+            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_pext_u32
+            "pext" => {
+                let mut mask = right;
+                let mut i = 0u32;
+                let mut result = 0;
+                // Iterate over the mask one 1-bit at a time, from
+                // the least significant bit to the most significant bit.
+                while mask != 0 {
+                    // Extract the bit marked by the mask's least significant set bit
+                    // and put it at position `i` of the result.
+                    result |= u64::from(left & (1 << mask.trailing_zeros()) != 0) << i;
+                    i = i.wrapping_add(1);
+                    // Clear the least significant set bit.
+                    mask &= mask.wrapping_sub(1);
+                }
+                result
+            }
+            // Deposit bit values of an unsigned integer to positions marked by a mask.
+            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_pdep_u32
+            "pdep" => {
+                let mut mask = right;
+                let mut set = left;
+                let mut result = 0;
+                // Iterate over the mask one 1-bit at a time, from
+                // the least significant bit to the most significant bit.
+                while mask != 0 {
+                    // Put rightmost bit of `set` at the position of the current `mask` bit.
+                    result |= (set & 1) << mask.trailing_zeros();
+                    // Go to next bit of `set`.
+                    set >>= 1;
+                    // Clear the least significant set bit.
+                    mask &= mask.wrapping_sub(1);
+                }
+                result
+            }
+            _ => return Ok(EmulateItemResult::NotSupported),
+        };
+
+        let result = if is_64_bit {
+            Scalar::from_u64(result)
+        } else {
+            Scalar::from_u32(u32::try_from(result).unwrap())
+        };
+        this.write_scalar(result, dest)?;
+
+        Ok(EmulateItemResult::NeedsReturn)
+    }
+}
--- a/tests/pass/shims/x86/intrinsics-x86-bmi.rs
+++ b/tests/pass/shims/x86/intrinsics-x86-bmi.rs
@@ -1,0 +1,216 @@
+// Ignore everything except x86 and x86_64
+// Any new targets that are added to CI should be ignored here.
+// (We cannot use `cfg`-based tricks here since the `target-feature` flags below only work on x86.)
+//@ignore-target-aarch64
+//@ignore-target-arm
+//@ignore-target-avr
+//@ignore-target-s390x
+//@ignore-target-thumbv7em
+//@ignore-target-wasm32
+//@compile-flags: -C target-feature=+bmi1,+bmi2
+
+#[cfg(target_arch = "x86")]
+use std::arch::x86::*;
+#[cfg(target_arch = "x86_64")]
+use std::arch::x86_64::*;
+
+fn main() {
+    // BMI1 and BMI2 are independent from each other, so both must be checked.
+    assert!(is_x86_feature_detected!("bmi1"));
+    assert!(is_x86_feature_detected!("bmi2"));
+
+    unsafe {
+        test_bmi_32();
+        test_bmi_64();
+    }
+}
+
+/// Test the 32-bit variants of the intrinsics.
+unsafe fn test_bmi_32() {
+    unsafe fn test_bextr_u32() {
+        let r = _bextr_u32(0b0101_0000u32, 4, 4);
+        assert_eq!(r, 0b0000_0101u32);
+
+        for i in 0..16 {
+            assert_eq!(_bextr_u32(u32::MAX, i, 4), 0b1111);
+            assert_eq!(_bextr_u32(u32::MAX, 4, i), (1 << i) - 1);
+        }
+
+        // Ensure that indices larger than the bit count are covered.
+        // It is important to go above 32 in order to verify the bit selection
+        // of the instruction.
+
+        for i in 0..256 {
+            // If the index is out of bounds, the original input won't be changed, thus the `min(32)`.
+            assert_eq!(_bextr_u32(u32::MAX, 0, i).count_ones(), i.min(32));
+        }
+
+        for i in 0..256 {
+            assert_eq!(_bextr_u32(u32::MAX, i, 0), 0);
+        }
+
+        // Test cases with completly random values. These cases also test
+        // that the function works even if upper bits of the control value are set.
+        assert_eq!(_bextr2_u32(0x7408a392, 0x54ef705), 0x3a0451c);
+        assert_eq!(_bextr2_u32(0xbc5a3494, 0xdd193203), 0x178b4692);
+        assert_eq!(_bextr2_u32(0xc0332325, 0xf96e207), 0x1806646);
+    }
+    test_bextr_u32();
+
+    unsafe fn test_pext_u32() {
+        let n = 0b1011_1110_1001_0011u32;
+
+        let m0 = 0b0110_0011_1000_0101u32;
+        let s0 = 0b0000_0000_0011_0101u32;
+
+        let m1 = 0b1110_1011_1110_1111u32;
+        let s1 = 0b0001_0111_0100_0011u32;
+
+        // Testing of random values.
+        assert_eq!(_pext_u32(n, m0), s0);
+        assert_eq!(_pext_u32(n, m1), s1);
+        assert_eq!(_pext_u32(0x12345678, 0xff00fff0), 0x00012567);
+
+        // Testing of various identities.
+        assert_eq!(_pext_u32(u32::MAX, u32::MAX), u32::MAX);
+        assert_eq!(_pext_u32(u32::MAX, 0), 0);
+        assert_eq!(_pext_u32(0, u32::MAX), 0);
+    }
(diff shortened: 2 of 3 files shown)


tests/pass/shims/x86/intrinsics-x86-bmi.rs
```diff
@@ -0,0 +1,146 @@
...
+        let n = 0b1011_1110_1001_0011u32;
+
+        let m0 = 0b0110_0011_1000_0101u32;
+        let s0 = 0b0000_0000_0011_0101u32;
+
+        let m1 = 0b1110_1011_1110_1111u32;
+        let s1 = 0b0001_0111_0100_0011u32;
+
+        assert_eq!(_pext_u32(n, m0), s0);
+        assert_eq!(_pext_u32(n, m1), s1);
```
folkertdev, line 46: the wikipedia has some examples that I think are a bit easier to grasp

```rust
    assert_eq!(_pdep_u32(0x00012567, 0xff00fff0), 0x12005670);
    assert_eq!(_pext_u32(0x12345678, 0xff00fff0), 0x00012567);
```

https://en.wikipedia.org/wiki/X86_Bit_manipulation_instruction_set#Parallel_bit_deposit_and_extract

TDecking: I've included them.

src/shims/x86/bmi.rs
```diff
@@ -0,0 +1,96 @@
...
+                if index >= 64 { left } else { left & 1u64.wrapping_shl(index).wrapping_sub(1) }
+            }
+            // Extract bit values of an unsigned integer at positions marked by a mask.
+            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_pext_u32
+            "pext" => {
+                let mut mask = right;
+                let mut i = 0u32;
+                let mut result = 0;
+                while mask != 0 {
+                    result |= u64::from(left & 1 << mask.trailing_zeros() != 0) << i;
```
RalfJung, line 65: Please add parentheses to clarify operator precedence, I can't even parse this.^^ I think it's this?
```suggestion
                    result |= u64::from(left & (1 << mask.trailing_zeros()) != 0) << i;
```

src/shims/x86/bmi.rs
```diff
@@ -0,0 +1,96 @@
...
+        let right = if is_64_bit { right.to_u64()? } else { u64::from(right.to_u32()?) };
+
+        let result = match unprefixed_name {
+            // Extract a contigous range of bits from an unsigned integer.
+            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_bextr_u32
+            "bextr" => {
+                let start = u32::try_from(right & 0xff).unwrap();
+                let len = u32::try_from((right >> 8) & 0xff).unwrap();
+                left.checked_shr(start).unwrap_or(0)
+                    & 1u64.checked_shl(len).map(|x| x.wrapping_sub(1)).unwrap_or(u64::MAX)
```
RalfJung, line 50: Are there tests covering the edge cases here (start and/or len being bigger than the size of the type in bits)?
Also, on 32bits for values between 32 and 64 this will *not* use the `unwrap_or` case but instead use extra bits that wouldn't even exist in hardware -- please make sure that is covered by tests.

tests/pass/shims/x86/intrinsics-x86-bmi.rs
```diff
@@ -0,0 +1,154 @@
...
+//@ignore-target-thumbv7em
+//@ignore-target-wasm32
+//@compile-flags: -C target-feature=+bmi1,+bmi2
+
+#[cfg(target_arch = "x86")]
+use std::arch::x86::*;
+#[cfg(target_arch = "x86_64")]
+use std::arch::x86_64::*;
+
+fn main() {
```
RalfJung, line 17: Usually these tests have a `assert!(is_x86_feature_detected!("..."));` to make sure they are safe to copy-paste and run in rustc as well. Please add that here.

tests/pass/shims/x86/intrinsics-x86-bmi.rs
```diff
@@ -0,0 +1,154 @@
...
+    unsafe {
+        test_bmi_32();
+        test_bmi_64();
+    }
+}
+
+unsafe fn test_bmi_32() {
```
RalfJung, line 24: The tests have zero comments. Is there anything to say about the values being tested or so?

src/shims/x86/bmi.rs
```diff
@@ -0,0 +1,96 @@
...
+            }
+            // Create a copy of an unsigned integer with bits above a certain index cleared.
+            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_bzhi_u32
+            "bzhi" => {
+                let index = u32::try_from(right & 0xff).unwrap();
+                if index >= 64 { left } else { left & 1u64.wrapping_shl(index).wrapping_sub(1) }
```
RalfJung, line 56: This should also have a test where `index` goes OOB. Why is it `wrapping` here but `unwrap_or(MAX)` below?

RalfJung (commented):
Thanks!
The world of Intel intrinsics is quite bizarre form an outside perspective. ;)

src/shims/x86/bmi.rs
folkertdev, line 50:
```suggestion
                    & 1u64 << (len & (64 - 1)).wrapping_sub(1);
```

I believe this is equivalent, but it's [shorter assembly](https://godbolt.org/z/69dGffEKq). The checked/unwrap stuff kind of obfuscates the logic for me here.

src/shims/x86/bmi.rs
folkertdev: I think `1u64 << (len & (64 - 1)).wrapping_sub(1)` or similar works better again

src/shims/x86/bmi.rs
TDecking: Done.

src/shims/x86/bmi.rs
TDecking: There are now.

tests/pass/shims/x86/intrinsics-x86-bmi.rs
TDecking: Not really. I've added some comments but there is not much worth explaining.

tests/pass/shims/x86/intrinsics-x86-bmi.rs
TDecking: Done.

src/shims/x86/bmi.rs
TDecking: I you refer to the different masking approach between the `bextr` and `bzhi` implementations, I've unified those.
As for large indices, there are now tests.

tests/pass/shims/x86/intrinsics-x86-bmi.rs
```diff
@@ -30,6 +35,16 @@ unsafe fn test_bmi_32() {
             assert_eq!(_bextr_u32(u32::MAX, i, 4), 0b1111);
             assert_eq!(_bextr_u32(u32::MAX, 4, i), (1 << i) - 1);
         }
+
+        // Ensure that indices larger than the bit count are covered.
+
+        for i in 0..1000 {
+            assert_eq!(_bextr_u32(u32::MAX, 0, i).count_ones(), (i & 0xff).min(32));
```
RalfJung, line 42: Why `(i & 0xff).min(32)`...?

tests/pass/shims/x86/intrinsics-x86-bmi.rs
```diff
@@ -78,8 +95,12 @@ unsafe fn test_bmi_32() {
         let s = 0b0001_0010u32;
         assert_eq!(_bzhi_u32(n, 5), s);
 
-        for i in 0..=32 {
-            assert_eq!(_bzhi_u32(u32::MAX, i), u32::MAX.checked_shr(32 - i).unwrap_or(0),);
+        // Ensure that indices larger than the bit count are covered.
+        for i in 0..1000 {
```
RalfJung, line 99: How long does this test take now? 1000 loop iterations can become somewhat slow in Miri...

tests/pass/shims/x86/intrinsics-x86-bmi.rs
TDecking: The instruction only takes the lowest eight bits to generate the index, thus the mask. If the index is out of bounds, the original input won't be changed, thus the `min(32)`.

tests/pass/shims/x86/intrinsics-x86-bmi.rs
TDecking: On my system, the execution was pretty much instant.

tests/pass/shims/x86/intrinsics-x86-bmi.rs
RalfJung: Okay, that seems worth a short comment.

TDecking: Done.

tests/pass/shims/x86/intrinsics-x86-bmi.rs
RalfJung: True, this is faster than I expected...
Still, should be enough to cover `0..=512`. That also looks less arbitrary than 1000. ;)

tests/pass/shims/x86/intrinsics-x86-bmi.rs
```diff
@@ -0,0 +1,196 @@
...
+
+        for i in 0..16 {
+            assert_eq!(_bextr_u32(u32::MAX, i, 4), 0b1111);
+            assert_eq!(_bextr_u32(u32::MAX, 4, i), (1 << i) - 1);
+        }
+
+        // Ensure that indices larger than the bit count are covered.
+
+        for i in 0..1000 {
+            // The instruction only takes the lowest eight bits to generate the index, thus the mask.
```
RalfJung, line 42:
```suggestion
            // The instruction only takes the lowest eight bits to generate the index, hence `i & 0xff`.
```

src/shims/x86/bmi.rs
```diff
@@ -0,0 +1,96 @@
...
+            // Extract bit values of an unsigned integer at positions marked by a mask.
+            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_pext_u32
+            "pext" => {
+                let mut mask = right;
+                let mut i = 0u32;
+                let mut result = 0;
+                while mask != 0 {
+                    result |= u64::from(left & (1 << mask.trailing_zeros()) != 0) << i;
+                    i = i.wrapping_add(1);
+                    mask &= mask.wrapping_sub(1);
```
RalfJung, line 67: I can't follow the bit magic here.^^ Why is the mask and-ed with `mask-1`...?
Please add a comment.

src/shims/x86/bmi.rs
```diff
@@ -0,0 +1,96 @@
...
+                let index = u32::try_from(right & 0xff).unwrap();
+                if index >= 64 { left } else { left & 1u64.wrapping_shl(index).wrapping_sub(1) }
+            }
+            // Extract bit values of an unsigned integer at positions marked by a mask.
+            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_pext_u32
+            "pext" => {
+                let mut mask = right;
+                let mut i = 0u32;
+                let mut result = 0;
+                while mask != 0 {
```
RalfJung, line 64:
```suggestion
                // Iterate over the mask one 1-bit at a time, right-to-left.
                while mask != 0 {
```

src/shims/x86/bmi.rs
```diff
@@ -0,0 +1,96 @@
...
+                }
+                result
+            }
+            // Deposit bit values of an unsigned integer to positions marked by a mask.
+            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_pdep_u32
+            "pdep" => {
+                let mut mask = right;
+                let mut set = left;
+                let mut result = 0;
+                while mask != 0 {
```
RalfJung, line 77:
```suggestion
                // Iterate over the mask one 1-bit at a time, right-to-left.
                while mask != 0 {
```

src/shims/x86/bmi.rs
```diff
@@ -0,0 +1,96 @@
...
+                    result |= (set & 1) << mask.trailing_zeros();
```
RalfJung, line 78:
```suggestion
                    // Put rightmost bit of `set` at the position of the current `mask` bit.
                    result |= (set & 1) << mask.trailing_zeros();
                    // Go to next bit of `set`.
```

src/shims/x86/bmi.rs
```diff
@@ -0,0 +1,96 @@
...
+        let right = if is_64_bit { right.to_u64()? } else { u64::from(right.to_u32()?) };
+
+        let result = match unprefixed_name {
+            // Extract a contigous range of bits from an unsigned integer.
+            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_bextr_u32
+            "bextr" => {
+                let start = u32::try_from(right & 0xff).unwrap();
+                let len = u32::try_from((right >> 8) & 0xff).unwrap();
+                let shifted = left.checked_shr(start).unwrap_or(0);
+                if len >= 64 { shifted } else { shifted & 1u64.wrapping_shl(len).wrapping_sub(1) }
```
RalfJung, line 50:
```suggestion
                // Keep the `len` lowest bits of `shifted`, or all bits if `len` is too big.
                if len >= 64 { shifted } else { shifted & 1u64.wrapping_shl(len).wrapping_sub(1) }
```

src/shims/x86/bmi.rs
```diff
@@ -0,0 +1,96 @@
...
+            }
+            // Create a copy of an unsigned integer with bits above a certain index cleared.
+            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_bzhi_u32
+            "bzhi" => {
+                let index = u32::try_from(right & 0xff).unwrap();
+                if index >= 64 { left } else { left & 1u64.wrapping_shl(index).wrapping_sub(1) }
```
RalfJung, line 56:
```suggestion
                // Keep the `index` lowest bits of `shifted`, or all bits if `index` is too big.
                if index >= 64 { left } else { left & 1u64.wrapping_shl(index).wrapping_sub(1) }
```

src/shims/x86/bmi.rs
```diff
@@ -0,0 +1,96 @@
...
+
+        let left = if is_64_bit { left.to_u64()? } else { u64::from(left.to_u32()?) };
+        let right = if is_64_bit { right.to_u64()? } else { u64::from(right.to_u32()?) };
+
+        let result = match unprefixed_name {
+            // Extract a contigous range of bits from an unsigned integer.
+            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_bextr_u32
+            "bextr" => {
+                let start = u32::try_from(right & 0xff).unwrap();
+                let len = u32::try_from((right >> 8) & 0xff).unwrap();
```
RalfJung, line 48: So any higher bits in `right` are entirely ignored, is that right? Do we have tests for that?

src/shims/x86/bmi.rs
```diff
@@ -0,0 +1,105 @@
...
+            // Extract bit values of an unsigned integer at positions marked by a mask.
+            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_pext_u32
+            "pext" => {
+                let mut mask = right;
+                let mut i = 0u32;
+                let mut result = 0;
+                // Iterate over the mask one 1-bit at a time, from
+                // the least significant bit to the most significant bit.
+                while mask != 0 {
+                    // Extract the bit marked by the masks least significant set bit
```
RalfJung, line 67:
```suggestion
                    // Extract the bit marked by the mask's least significant set bit
```

src/shims/x86/bmi.rs
```diff
@@ -0,0 +1,105 @@
...
+            "pdep" => {
+                let mut mask = right;
+                let mut set = left;
+                let mut result = 0;
+                // Iterate over the mask one 1-bit at a time, from
+                // the least significant bit to the most significant bit.
+                while mask != 0 {
+                    // Put rightmost bit of `set` at the position of the current `mask` bit.
+                    result |= (set & 1) << mask.trailing_zeros();
+                    set >>= 1;
```
RalfJung, line 87:
```suggestion
                    // Go to next bit of `set`.
                    set >>= 1;
```

src/shims/x86/bmi.rs
TDecking: We do. It's one of the reasons why the loop bbounds were set to 1000, to ensure the coverage of this case.

TDecking:
@RalfJung All the comments are added.

src/shims/x86/bmi.rs
RalfJung: It's impossible to cover that case by calling `_bextr_u64`, you have to call `_bextr2_u64` directly.

TDecking: I've added test cases using the `_bextr2` variants.
The [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d6d63b2a0b327b49bd87913132f5f57e) also verifies these cases.

tests/pass/shims/x86/intrinsics-x86-bmi.rs
```diff
@@ -0,0 +1,224 @@
...
+        for i in 0..=1024 {
+            let start = i & 0xff;
+            let len = (i >> 8) & 0xff;
+            let expected = {
+                let shifted = u32::MAX.checked_shr(start).unwrap_or(0);
+                if len >= 32 { shifted } else { shifted & ((1u32 << len) - 1) }
+            };
+            let actual = _bextr2_u32(u32::MAX, i);
+            assert_eq!(expected, actual);
+        }
```
RalfJung, line 62: Sorry, but this doesn't make a ton of sense to me. This is a lot of loop iterations and still not covering some interesting cases, like when `i` is above 0xffff (and hence has some bits that are ignored entirely). Also this goes to show that the first loop, that goes until 512, is kind of pointless since `i` is already going to be masked before it even reaches the intrinsic.

I think it makes more sense to just test a few hand-picked cases that call the intrinsic directly with interesting values, and hard-code their results, with some comments explaining things.

TDecking: Done.

RalfJung: > Also this goes to show that the first loop, that goes until 512, is kind of pointless since i is already going to be masked before it even reaches the intrinsic.

This part has not been resolved. It makes no sense for the loop to go beyond 255.

TDecking: The loop bounds are adjusted now.

RalfJung:
Thanks, looks good! Please squash the commits.

TDecking:
Everything is ready.

RalfJung:
Awesome, thanks for the PR and the adjustments during review. :)

@bors r+

PR #3675, opened 2024-06-15
disable tls with Box::leak test on solaris/illumos.

devnexen:
on these platforms, thread_local is disabled as there is no support for cleanup callback, the test is leaking here.

### DIFF
--- a/tests/pass/tls/tls_leak_main_thread_allowed.rs
+++ b/tests/pass/tls/tls_leak_main_thread_allowed.rs
@@ -1,4 +1,6 @@
 //@ignore-target-windows: Windows uses a different mechanism for `thread_local!`
+//@revisions: default tls
+//@[tls]compile-flags: -Zmiri-ignore-leaks
 #![feature(thread_local)]
 
 use std::cell::Cell;


saethlin:
Instead of ignoring the test on the target, can you add `-Zmiri-ignore-leaks` using the revisions system such as done in this test: 
https://github.com/rust-lang/miri/blob/60a720040d6c60656ab9cac0980e587d19d9c07a/tests/pass/tree_borrows/tree-borrows.rs#L1-L3

saethlin:
I think you need multiple revisions as in the example. Does the test now fail on Linux if you add a leak to it?

RalfJung:
I don't see why this test should be ignored on Solarish. It shows that we are not properly recognizing main thread TLS variables as being allowed-to-leak.

RalfJung:
Also, other tests like tls_macro_drop work fine. "on these platforms, thread_local is disabled" does not seem to be correct. The cleanup callback *is* called when it is needed. So I don't follow your PR description either.

RalfJung:
@b-ncMN

RalfJung:
@devnexen what are your plans for this PR?

Looking at the test, what happens here is that the logic for ignoring memory leaks in the main thread TLS state does not work on Solarish. We already have `ignore-target-windows`, so adding more `ignore-target` with an explanation for why they are ignored seems fine. I realize that's what you did originally, but you didn't explain to us why you did what we did so we weren't able to give you proper advice. I now spent the effort of digging into the problem myself, but often I won't have the time for that and we rely on contributors to explain to us what it is they are doing and why.

So to move this forward, please update not only the patch but also the PR description. The explanation for why we want to ignore some targets is: `thread_local!` doesn't use `#[thread_local]` statics on these targets.

Also, if the goal of the PR is to make more tests pass on Solarish, please adjust `ci.sh` and add `tls` to the list of tests being run.

RalfJung:
Thinking about this some more, it's probably best if that test just checks `cfg(target_thread_local)` directly. That makes it then work on Solarish without any further changes. :)

I implemented that in https://github.com/rust-lang/miri/pull/3822, so that supersedes this PR. Thanks for looking into this!

devnexen:
Thanks. Sorry I put it aside I no longer access my illumos VM at the moment.

PR #3676, opened 2024-06-15
add the `pext` and `pdep` intrinsics

folkertdev:
Adds support for parallel bit extraction and deposition. 

These are useful in low-level parsing code.

Wikipedia has [an explanation](https://en.wikipedia.org/wiki/X86_Bit_manipulation_instruction_set#Parallel_bit_deposit_and_extract) of what these do.

### DIFF
--- a/src/shims/x86/bmi.rs
+++ b/src/shims/x86/bmi.rs
@@ -82,8 +82,8 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 ///
 /// See also
 ///
-/// - https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_pdep_u64
-/// - https://en.wikipedia.org/wiki/X86_Bit_manipulation_instruction_set#Parallel_bit_deposit_and_extract
+/// - <https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_pdep_u64>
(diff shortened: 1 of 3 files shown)


src/shims/x86/bmi.rs
```diff
@@ -0,0 +1,126 @@
...
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx, EmulateItemResult> {
+        let this = self.eval_context_mut();
+        // Prefix should have already been checked.
+        let unprefixed_name = link_name.as_str().strip_prefix("llvm.x86.bmi.").unwrap();
+
+        match unprefixed_name {
+            "pdep.32" => {
+                // <https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_pdep_u32>
+                this.expect_target_feature_for_intrinsic(link_name, "bmi2")?;
```
folkertdev, line 22: llvm unhelpfully does not make a distinction between `bmi1` and `bmi2`: all bmi intrinsics are in the `bmi` namespace. So I don't see another way than to in every branch check for the relevant target features, even though it is error-prone.

folkertdev:
whoops, https://github.com/rust-lang/miri/pull/3674 exists

RalfJung:
Ah, that is an unfortunate coincidence that you both wrote the same PR at the same time...

If you could take a look at the other PR, some help with reviewing is always appreciated. :)

PR #3686, opened 2024-06-19
bump ui_test to 0.22.2

klensy:
This is almost #3279, but working?

Can't currently bump to ui_test 0.22.3, as tests start failing (but contains https://github.com/oli-obk/ui_test/pull/196, which i kind-of manually included here).

Interaction with `target` around is annoying and I'm not sure that this working as intended.

### DIFF
--- a/tests/ui.rs
+++ b/tests/ui.rs
@@ -7,6 +7,7 @@ use std::{env, process::Command};
 use colored::*;
 use regex::bytes::Regex;
 use ui_test::color_eyre::eyre::{Context, Result};
+use ui_test::spanned::Spanned;
 use ui_test::{
     status_emitter, CommandBuilder, Config, Format, Match, Mode, OutputConflictHandling,
     RustfixMode,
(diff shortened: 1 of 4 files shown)


tests/fail-dep/concurrency/windows_join_detached.stderr
```diff
@@ -11,7 +11,7 @@ LL |         let rc = unsafe { c::WaitForSingleObject(self.handle.as_raw_handle(
    = note: inside `std::thread::JoinInner::<'_, ()>::join` at RUSTLIB/std/src/thread/mod.rs:LL:CC
    = note: inside `std::thread::JoinHandle::<()>::join` at RUSTLIB/std/src/thread/mod.rs:LL:CC
 note: inside `main`
-  --> $DIR/windows_join_detached.rs:LL:CC
```
RalfJung, line 14: Why is this not being normalized any more?

tests/ui.rs
```diff
@@ -130,9 +149,11 @@ fn run_tests(
         }
     }
     config.program.args.push("-Zui-testing".into());
-    config.program.args.push("--target".into());
-    config.program.args.push(target.into());
 
+    if config.host != config.target {
```
RalfJung, line 153: Why can't we always set `--target`?

klensy: `*-dep` tests start passing double `--target` args to rustc which fails because of that. Better solution is update to 0.22.3 (which currently blocked by other error)

tests/fail-dep/concurrency/windows_join_detached.stderr
klensy: `https://github.com/oli-obk/ui_test/blob/0.22.0/CHANGELOG.md#removed` https://github.com/oli-obk/ui_test/pull/180 but probably can be restored.

RalfJung: Yeah we probably want to add that normalization ourselves then.

tests/ui.rs
RalfJung: But why is that only a problem with `host == target`...? Very strange.

Anyway, please add a comment in the code.

tests/fail-dep/concurrency/windows_join_detached.stderr
klensy: Added workaround, now it's works almost as before.

tests/ui.rs
RalfJung: It seems to be a problem with `host != target` as well, judging from the CI failures
```
error: Option 'target' given more than once
```

I think ui_test 0.22 may be too buggy for us, unfortunately. We'll probably have to wait for 0.24 as 0.23 suffers from https://github.com/oli-obk/ui_test/issues/238.

RalfJung:
@klensy do you still plan to get back to this PR?
It should probably update ui_test to 0.24, since 0.22 has some issue with `--target`.

RalfJung:
I'll close this PR due to inactivity. Feel free to pick it up again any time -- probably best to update to version 0.25, then. :)

PR #3688, opened 2024-06-20
Refactor float casting tests, add `f16` and `f128`

tgross35:
This is an attempt to remove the magic from a lot of the numbers tested, which makes things easier when adding `f16` and `f128`. A nice side effect is that these tests now cover all int <-> float conversions with the same amount of tests.

### DIFF
--- a/tests/pass/float.rs
+++ b/tests/pass/float.rs
@@ -4,8 +4,11 @@
 #![feature(f128)]
 #![feature(f16)]
 #![allow(arithmetic_overflow)]
+#![allow(internal_features)]
 
-use std::fmt::Debug;
+use std::any::type_name;
+use std::cmp::min;
+use std::fmt::{Debug, Display, LowerHex};
 use std::hint::black_box;
 use std::{f32, f64};
 
@@ -29,14 +32,38 @@ fn main() {
     test_algebraic();
 }
 
-// Helper function to avoid promotion so that this tests "run-time" casts, not CTFE.
-// Doesn't make a big difference when running this in Miri, but it means we can compare this
-// with the LLVM backend by running `rustc -Zmir-opt-level=0 -Zsaturating-float-casts`.
-#[track_caller]
-#[inline(never)]
-fn assert_eq<T: PartialEq + Debug>(x: T, y: T) {
-    assert_eq!(x, y);
+trait Float: Copy + PartialEq + Debug {
+    /// The unsigned integer with the same bit width as this float
+    type Int: Copy + PartialEq + LowerHex + Debug;
+    const BITS: u32 = size_of::<Self>() as u32 * 8;
+    const EXPONENT_BITS: u32 = Self::BITS - Self::SIGNIFICAND_BITS - 1;
+    const SIGNIFICAND_BITS: u32;
+
+    /// The saturated (all ones) value of the exponent (infinity representation)
+    const EXPONENT_SAT: u32 = (1 << Self::EXPONENT_BITS) - 1;
+
+    /// The exponent bias value (max representable positive exponent)
+    const EXPONENT_BIAS: u32 = Self::EXPONENT_SAT >> 1;
+
+    fn to_bits(self) -> Self::Int;
 }
+
+macro_rules! impl_float {
+    ($ty:ty, $ity:ty) => {
+        impl Float for $ty {
+            type Int = $ity;
+            // Just get this from std's value, which includes the implicit digit
+            const SIGNIFICAND_BITS: u32 = <$ty>::MANTISSA_DIGITS - 1;
+
+            fn to_bits(self) -> Self::Int {
+                self.to_bits()
+            }
+        }
+    };
+}
+
+impl_float!(f32, u32);
+impl_float!(f64, u64);
 
 trait FloatToInt<Int>: Copy {
     fn cast(self) -> Int;
@@ -64,13 +91,53 @@ float_to_int!(f64 => i8, u8, i16, u16, i32, u32, i64, u64, i128, u128);
 /// Test this cast both via `as` and via `approx_unchecked` (i.e., it must not saturate).
 #[track_caller]
 #[inline(never)]
-fn test_both_cast<F, I>(x: F, y: I)
+fn test_both_cast<F, I>(x: F, y: I, msg: impl Display)
 where
     F: FloatToInt<I>,
     I: PartialEq + Debug,
 {
-    assert_eq!(x.cast(), y);
-    assert_eq!(unsafe { x.cast_unchecked() }, y);
+    let f_tname = type_name::<F>();
+    let i_tname = type_name::<I>();
+    assert_eq!(x.cast(), y, "{f_tname} -> {i_tname}: {msg}");
+    assert_eq!(unsafe { x.cast_unchecked() }, y, "{f_tname} -> {i_tname}: {msg}",);
+}
+
+/// Helper function to avoid promotion so that this tests "run-time" casts, not CTFE.
+/// Doesn't make a big difference when running this in Miri, but it means we can compare this
+/// with the LLVM backend by running `rustc -Zmir-opt-level=0 -Zsaturating-float-casts`.
+#[track_caller]
+#[inline(never)]
+fn assert_eq<T: PartialEq + Debug>(x: T, y: T) {
+    assert_eq!(x, y);
+}
+
+/// The same as `assert_eq` except prints a specific message on failure
+#[track_caller]
+#[inline(never)]
+fn assert_eq_msg<T: PartialEq + Debug>(x: T, y: T, msg: impl Display) {
+    assert_eq!(x, y, "{msg}");
+}
+
+/// Check that floats have bitwise equality
+fn assert_biteq<F: Float>(a: F, b: F, msg: impl Display) {
+    let ab = a.to_bits();
+    let bb = b.to_bits();
+    let tname = type_name::<F>();
+    let width = (2 + F::BITS / 4) as usize;
+    assert_eq_msg::<F::Int>(
+        ab,
+        bb,
+        format_args!("({ab:#0width$x} != {bb:#0width$x}) {tname}: {msg}"),
+    );
+}
+
+/// Check that two floats have equality
+fn assert_feq<F: Float>(a: F, b: F, msg: impl Display) {
+    let ab = a.to_bits();
+    let bb = b.to_bits();
+    let tname = type_name::<F>();
+    let width = (2 + F::BITS / 4) as usize;
+    assert_eq_msg::<F>(a, b, format_args!("({ab:#0width$x} != {bb:#0width$x}) {tname}: {msg}"));
 }
 
 fn basic() {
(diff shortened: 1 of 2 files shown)


tests/pass/float.rs
```diff
@@ -148,155 +260,250 @@ fn basic() {
...
+
+        #[allow(overflowing_literals)]
+        let imin_f = $imin_f;
+        #[allow(overflowing_literals)]
+        let imax_f = $imax_f;
+
+        // Skip unchecked cast if negative numbers are unrepresentable
+        let assert_ftoi_neg = if isigned { assert_ftoi } else { assert_ftoi_unrep };
+        // Skip unchecked cast when int min/max would be unrepresentable
+        let assert_ftoi_big = if ibits < fbits { assert_ftoi } else { assert_ftoi_unrep };
```
RalfJung, line 315: This is not a strict rule, is it? A hypothetical f33 type could probably not represent all u32 even though it is "bigger", since bits are lost to exponent and sign?
I'm okay with having a heuristic here, but there should be comments explaining that.

tests/pass/float.rs
```diff
@@ -148,155 +260,250 @@ fn basic() {
...
+        assert_ftoi(-0.0, 0, "negative zero");
+        assert_ftoi_neg(-0.99999999999999999999999999, <$ity>::NEG_ONE, "near -1");
+        assert_ftoi(<$fty>::from_bits(0x1), 0, "min subnormal");
+        assert_ftoi(<$fty>::from_bits(0x1 | 1 << (fbits - 1)), 0, "min neg subnormal");
+
+        // Spot checks
+        assert_ftoi(0.9, 0, "0.9");
+        assert_ftoi_neg(-0.9, 0, "-0.9");
+        assert_ftoi(1.1, 1, "1.1");
+        assert_ftoi_neg(-1.1, <$ity>::NEG_ONE, "-1.1");
```
RalfJung, line 329: Why can floats use literals but integers need to use constants?

RalfJung (commented):
Running out of time now, will go on later this week hopefully

tests/pass/float.rs
```diff
@@ -29,15 +32,78 @@ fn main() {
...
 
-// Helper function to avoid promotion so that this tests "run-time" casts, not CTFE.
-// Doesn't make a big difference when running this in Miri, but it means we can compare this
-// with the LLVM backend by running `rustc -Zmir-opt-level=0 -Zsaturating-float-casts`.
-#[track_caller]
-#[inline(never)]
-fn assert_eq<T: PartialEq + Debug>(x: T, y: T) {
-    assert_eq!(x, y);
+trait Float: Copy + PartialEq + Debug {
+    type Int: Int;
```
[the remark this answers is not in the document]
tgross35: Unsigned and the same size, I'll update the docs and bound

tests/pass/float.rs
tgross35: I think the actual rule is `F::SIGNIFICAND_BITS + 1 >= I::BITS` to have a guaranteed representation for all values, since the entire integer can be contained without truncation. I'll just change this since it may come up for something like bf16 if that winds up being added.

tests/pass/float.rs
tgross35: Just an easy way to have a different value depending on sign - it's called `NEG_ONE` but the value is 0 on unsigned numbers https://github.com/rust-lang/miri/pull/3688/files#diff-64fd96f1f6fb00d3a7940a8e5c347ce262ea751844538a2c87d4b17574a01f51R73-R76

tests/pass/float.rs
RalfJung: Why `+ 1`?

tests/pass/float.rs
RalfJung: Call it `SATURATED_NEG_ONE` then maybe?
And we can't do this with literals? I guess `if signed { -1 } else { 0 }` will cause errors since `-1` is not a valid literal for an unsigned type? Maybe `if signed { -1i32 as $ity } else { 0 }`  or so?

tgross35: `SATURATING_*` would be a better name, but that reminded me that `zero.saturating_sub(1) // or 5` also works. Think that's a bit cleaner without the `if`, so I updated to that

tests/pass/float.rs
```diff
@@ -148,155 +253,256 @@ fn basic() {
...
+        assert_ftoi(1.9, 1, "1.9");
+        assert_ftoi_neg(-1.9, izero.saturating_sub(1), "-1.9");
+        assert_ftoi(5.0, 5, "5.0");
+        assert_ftoi_neg(-5.0, izero.saturating_sub(5), "-5.0");
+        assert_ftoi(5.9, 5, "5.0");
+        assert_ftoi_neg(-5.9, izero.saturating_sub(5), "-5.0");
+
+        // Half representations should always fit perfectly
+        let half_i_max =
+            <$ity>::try_from((!(0 as <$ity as Int>::Unsigned) >> (ibits / 2)) + 1).unwrap();
```
RalfJung, line 338: Can't this unconditionally use `u128`?
I also don't understand what a "half representation" is. What is this computing?

tests/pass/float.rs
```diff
@@ -148,155 +253,256 @@ fn basic() {
...
+        fn assert_bidir_unrep(f: $fty, i: $ity, msg: &str) {
+            assert_ftoi_unrep(f, i, msg);
+            assert_itof(i, f, msg);
+        }
+
+        let fbits = <$fty>::BITS;
+        let ibits = <$ity>::BITS;
+        let imax = <$ity>::MAX;
+        let imin = <$ity>::MIN;
+        let izero = <$ity>::ZERO;
```
RalfJung, line 298: I am still a bit confused by this mix of macros and traits. Since this is a macro, you can use `0 as $ity`, `$ity::MAX` etc, can't you? `BITS` is `8 * size_of::<$ity>()`. `signed` is the only tricky one, but `0.saturating_sub(1) < 0` should work.

tests/pass/float.rs
```diff
@@ -148,155 +253,256 @@ fn basic() {
...
+        let isigned = <$ity>::SIGNED;
+
+        #[allow(overflowing_literals)]
+        let imin_f = $imin_f;
+        #[allow(overflowing_literals)]
+        let imax_f = $imax_f;
```
RalfJung, line 304: Should these get type annotations to be sure they are used at the right type?

tests/pass/float.rs
```diff
@@ -29,15 +32,71 @@ fn main() {
...
-#[track_caller]
-#[inline(never)]
-fn assert_eq<T: PartialEq + Debug>(x: T, y: T) {
-    assert_eq!(x, y);
+trait Float: Copy + PartialEq + Debug {
+    /// The unsigned integer with the same bit width as this float
+    type Int: Int;
+    const BITS: u32;
+    const EXPONENT_BITS: u32;
+    const SIGNIFICAND_BITS: u32 = Self::BITS - Self::EXPONENT_BITS - 1;
```
RalfJung, line 40: libcore has MANTISSA_DIGITS; this can be computed from them, can't it?

tests/pass/float.rs
```diff
@@ -148,155 +253,256 @@ fn basic() {
...
+
+            if isigned {
+                let min_ge_imin = imin + (exp_overflow as $ity);
+                assert_bidir(min_ge_imin as $fty, imin, "min representable near int min");
+            }
+        }
+
+        if fbits >= ibits {
+            // Max representable power of 10
+            let pow10_max = (izero + 10).pow(imax.ilog10());
```
RalfJung, line 376:
```suggestion
            let pow10_max = (10 as $ity).pow(imax.ilog10());
```

tests/pass/float.rs
tgross35: Good call, added these and a few other places

tests/pass/float.rs
tgross35: Even easier, I can get the same pattern by shifting a single 0x1. I think from the original test this was just meant to test the middle of the integer's dynamic range, updated the comment to make this more clear

tests/pass/float.rs
tgross35: Updated this

tests/pass/float.rs
tgross35: Yes, updated to use that

tests/pass/float.rs
```diff
@@ -148,155 +237,258 @@ fn basic() {
...
+        assert_ftoi_neg(-1.1, izero.saturating_sub(1), "-1.1");
+        assert_ftoi(1.9, 1, "1.9");
+        assert_ftoi_neg(-1.9, izero.saturating_sub(1), "-1.9");
+        assert_ftoi(5.0, 5, "5.0");
+        assert_ftoi_neg(-5.0, izero.saturating_sub(5), "-5.0");
+        assert_ftoi(5.9, 5, "5.0");
+        assert_ftoi_neg(-5.9, izero.saturating_sub(5), "-5.0");
+
+        // Exercise the middle of the integer's bit range. A single digit should always
+        // fit perfectly.
```
RalfJung, line 321: Why should this work? The middle of u128 is around 2^64 and I doubt that fits into an f16?

tests/pass/float.rs
```diff
@@ -64,13 +105,59 @@ float_to_int!(f64 => i8, u8, i16, u16, i32, u32, i64, u64, i128, u128);
...
+        width = (2 + F::Int::BITS / 4) as usize
+    );
+    assert_eq_msg::<F::Int>(ab, bb, &msg);
+}
+
+/// Check that two floats have equality
+fn assert_feq<F: Float>(a: F, b: F, msg: &str) {
+    let ab = a.to_bits();
+    let bb = b.to_bits();
+    let msg = format!(
```
RalfJung, line 155: There's a lot of `format!` happening for strings that will never be shown... I guess even Miri is fast enough for this but I still find it mildly displeasing.^^

Does it work to make `assert_eq_msg` take `impl Display` or so?

tests/pass/float.rs
tgross35: The trait originally had a few more things because I was hoping to use a generic function rather than a macro, that just unfortunately fell apart because of the need for literals (at least without relying on float to float casting, which is also being tested). I think keeping the trait for the last few things just makes it easier to use in multiple macros (at least for `Float`).

RalfJung: Yeah I think keeping the Float trait is fine, but the Int trait IMO isn't carrying its weight.

tests/pass/float.rs
tgross35: Works as long as the exponent can reach +64... which you are correct does not fit in f16's 5 bits of exponent. Any suggestions for how to tweak it?

RalfJung: Maybe shift by `ibits.min(EXPONENT_BITS) - 1`, the largest power of 2 that fits both?

tests/pass/float.rs
tgross35: That does indeed work, updated

tests/pass/float.rs
tgross35: Alright, it's gone

tests/pass/float.rs
tgross35: Adjusted this to what I _think_ is the correct calculation

tests/pass/float.rs
```diff
@@ -148,155 +212,266 @@ fn basic() {
...
+        assert_ftoi_neg(-1.9, izero.saturating_sub(1), "-1.9");
+        assert_ftoi(5.0, 5, "5.0");
+        assert_ftoi_neg(-5.0, izero.saturating_sub(5), "-5.0");
+        assert_ftoi(5.9, 5, "5.0");
+        assert_ftoi_neg(-5.9, izero.saturating_sub(5), "-5.0");
+
+        // Exercise the middle of the integer's bit range. A single digit fits as long as the
+        // exponent can fit its offset, so cap at the largest shift representable by the
+        // exponent.
+        let half_i_max: $ity = 1 << (ibits / 2).min((1 << (<$fty>::EXPONENT_BITS - 1)) - 1);
```
RalfJung, line 305: I am quite confused by this... this is `1 << min(ibits / 2, (1 << (<$fty>::EXPONENT_BITS - 1)) - 1)`, right? The LHS is still the log but the RHS is already the power...?

I would have expected something more like `1 << min(ibits / 2, <$fty>::EXPONENT_BITS - 1)`.

tests/pass/float.rs
```diff
@@ -148,155 +212,266 @@ fn basic() {
...
+        assert_ftoi(1.1, 1, "1.1");
+        assert_ftoi_neg(-1.1, izero.saturating_sub(1), "-1.1");
+        assert_ftoi(1.9, 1, "1.9");
+        assert_ftoi_neg(-1.9, izero.saturating_sub(1), "-1.9");
+        assert_ftoi(5.0, 5, "5.0");
+        assert_ftoi_neg(-5.0, izero.saturating_sub(5), "-5.0");
+        assert_ftoi(5.9, 5, "5.0");
+        assert_ftoi_neg(-5.9, izero.saturating_sub(5), "-5.0");
+
+        // Exercise the middle of the integer's bit range. A single digit fits as long as the
```
RalfJung, line 302:
```suggestion
        // Exercise the middle of the integer's bit range. A power of 2 fits as long as the
```

tests/pass/float.rs
```diff
@@ -148,155 +212,266 @@ fn basic() {
...
+        // exponent can fit its offset, so cap at the largest shift representable by the
```
RalfJung, line 303:
```suggestion
        // exponent can fit its log, so cap at the largest shift representable by the
```

tests/pass/float.rs
```diff
@@ -148,155 +212,266 @@ fn basic() {
...
+        assert_bidir_big(imin_f, imin, "i min");
+        assert_ftoi_big(imax_f + 0.99, <$ity>::MAX, "slightly above i max");
+        assert_ftoi_neg(imin_f - 0.99, <$ity>::MIN, "slightly below i min");
+
+        if fbits == ibits {
+            // Signed integers have one fewer bit of range
+            let adj = if isigned { 1 } else { 0 };
+
+            let exp_overflow = 1 << (<$fty>::EXPONENT_BITS - 1 - adj);
+            let exp_max = (1 << <$fty>::EXPONENT_BITS - adj) - 1;
```
RalfJung, line 321: I have absolutely no idea what this does

tests/pass/float.rs
tgross35: `1 << (EXPONENT_BITS - 1)` would be `1 << 4`, which is the max value that can fit in an exponent but not the max value the exponent can represent. However, turns out I was just recalculating `EXPONENT_BIAS`, so I can use that constant instead.

tests/pass/float.rs
tgross35: Added

tests/pass/float.rs
tgross35: I am not sure how to explain it concisely since I don't entirely understand it either, could use some suggestions. This comes from the "rounding loss" magic number tests:

```rust
test_both_cast::<f32, u32>((u32::MAX - 128) as f32, u32::MAX - 255); // rounding loss
assert_eq::<u32>((u32::MAX - 127) as f32 as u32, u32::MAX); // rounds up and then becomes unrepresentable

test_both_cast::<f64, u64>((u64::MAX - 1024) as f64, u64::MAX - 2047); // rounding loss
assert_eq::<u64>((u64::MAX - 1023) as f64 as u64, u64::MAX); // rounds up and then becomes unrepresentable
```

Basically it calculates the maximum value that will get rounded down to an integer less than `Int::MAX` (that exact value being `u32::MAX - 255` / `u64::MAX - 2047`), and checks the result. Then any number greater will get rounded up to `Int::MAX` or above.

I had to reverse engineer the pattern, `(exp_overflow, exp_max)` are my bad names for `(128, 255)` and `(1024, 2047)`. I renamed them to `(nearest_rep_offset, imax_offset)` but even those don't seem great. Signed numbers aren't tested in the original but the pattern holds if you shift the bits an extra position (as I do with `adj`).

RalfJung: I think I added these tests. Basically I was looking for the largest integer that does round-trip properly with a float. For f32 and u32, that's `u32::MAX - 128` -- once you get any bigger, it becomes a float that saturates to `u32:MAX` (and causes UB with the unchecked casts).

I have no idea how to compute these numbers I found them experimentally.

tests/pass/float.rs
RalfJung: Ah, yes makes sense, thanks. I forgot that the exponent is one more log level away, of course it is.

tests/pass/float.rs
tgross35: I tested this pattern with f16 and f128 before and I _think_ it holds true for those as well, but don't remember. https://github.com/rust-lang/rust/pull/126608 should hopefully be in pretty soon so I'll probably just wait for that then update this PR, to get the extra two cases to test the tests.

RalfJung: If you want to stick to computing these numbers, then please add comments explaining what you are computing and how.

Is there a reason we can only have these tests when `fbits == ibits`? I couldn't be bothered to figure out the largest representable number for all combinations of types, but if there is a way to compute this automatically then that should be doable, right?

tests/pass/float.rs
```diff
@@ -148,155 +213,268 @@ fn basic() {
...
+            assert_ftoi_unrep(<$fty>::MAX, imax, "f max");
+            assert_ftoi_unrep(<$fty>::MIN, imin, "f min");
+        }
+
+        if ibits > fbits * 2 && !isigned {
+            // Unsigned max is enough to saturate only when more than 2x bit width
+            assert_itof(imax, <$fty>::INFINITY, "imax, ibits > fbits * 2");
+        } else {
+            // Signed values and anything smaller still fit
+            assert_itof(imax, imax_f, "imax, ibits > fbits * 2");
```
RalfJung, line 359: Why does this say `ibits > fbits * 2`? It's in the `else` branch!

Also why can things never saturate for `signed`?

tests/pass/float.rs
```diff
@@ -148,155 +213,268 @@ fn basic() {
...
+                assert_bidir((min_ge_imin - 1) as $fty, imin, "min representable near int min");
+            }
+        }
+
+        if fbits >= ibits {
+            // Max representable power of 10
+            let pow10_max = (10 as $ity).pow(imax.ilog10());
+            assert_bidir(pow10_max as $fty, pow10_max, "pow10 max");
+
+            // Float limits
```
RalfJung, line 349:
```suggestion
            // Float limits saturate when the float type is at least as big as the int type.
```

tests/pass/float.rs
```diff
@@ -148,155 +213,268 @@ fn basic() {
...
+            if isigned {
+                // Min representable value that is <= I::MIN
+                let min_ge_imin = imin + nearest_rep_offset;
+                assert_bidir(min_ge_imin as $fty, imin, "min representable near int min");
+                assert_bidir((min_ge_imin - 1) as $fty, imin, "min representable near int min");
+            }
+        }
+
+        if fbits >= ibits {
+            // Max representable power of 10
```
RalfJung, line 345:
```suggestion
            // Max representable power of 10 in the int type must be exactly representable
            // in the float type.
```

tests/pass/float.rs
```diff
@@ -148,155 +213,268 @@ fn basic() {
...
+
+            if isigned {
+                // Min representable value that is <= I::MIN
+                let min_ge_imin = imin + nearest_rep_offset;
+                assert_bidir(min_ge_imin as $fty, imin, "min representable near int min");
+                assert_bidir((min_ge_imin - 1) as $fty, imin, "min representable near int min");
+            }
+        }
+
+        if fbits >= ibits {
```
RalfJung, line 344:
```suggestion
        // Some tests for when the float type has (much) larger range than the int type.
        if fbits >= ibits {
```

tests/pass/float.rs
tgross35: I guess this `if` branch isn't even needed, if `imax as F` saturates then the literal is also saturated. Removed

tests/pass/float.rs
```diff
@@ -148,155 +213,265 @@ fn basic() {
...
+            // in the float type.
+            let pow10_max = (10 as $ity).pow(imax.ilog10());
+            assert_bidir(pow10_max as $fty, pow10_max, "pow10 max");
+
+            // Float limits saturate.
+            assert_ftoi_unrep(<$fty>::MAX, imax, "f max");
+            assert_ftoi_unrep(<$fty>::MIN, imin, "f min");
+        }
+
+        // Check potentially saturating int ranges.
```
RalfJung, line 356:
```suggestion
        // Check potentially saturating int ranges.
        // (`imax_f` here will be `$fty::INFINITY` if it cannot be represented as a finite value.)
```

tgross35:
Ah I thought https://github.com/rust-lang/miri/pull/3721 might have gotten it, but it looks like https://github.com/rust-lang/rust/pull/127032 still didn't get synced yet.

RalfJung:
I did another rustup, if you rebase again it should work.

tgross35:
Thanks for the update. Great news is that most of these pass, but I need to fix how I restrict a couple of them. And wrap up the remaining comments of course.

tests/pass/float.rs
tgross35: Finally think I wrapped my head around this - the max integer that can be represented is a saturated mantissa (including the implicit bit), shifted to fully to the left in the integer. IOW, zero out the last F::EXPONENT_BITS of I::MAX if they have the same size, which is what subtracting 255 does (255 = 0b11111111, which is a f32's exponent mask). So `x = u32::MAX - 255`

Then the cutoff point between rounding to x and rounding to infinity will be midway between x and the integer's max, or adding back `x + 255/2`. `-255 + (255 / 2) = `-128`, hence that number. Signed integers don't use the top bit for their maximum, so they get one extra bit in their lower range compared to their same-sized unsigned integer. This is what I was representing with `adj`, without really understanding it.

I think this should be generalizable to anything where:

- `F::EXPONENT_BIAS >= I::BITS`, because it needs to be able to represent the magnitude of an integer with a learning `1`. Basically no f16.
- `I::BITS > (F::MANTISSA_BITS + 1)`, because if any integer bitpattern fits in the mantissa then this is pointless.

(This isn't so much an update for you, mostly just writing down my informal proof so I remember it :). [scratchpad](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d5ba0abff91d84238da3789ef0f90c6a) )

tgross35:
🎉 with a better understanding of the tests in https://github.com/rust-lang/miri/pull/3688#discussion_r1664895098, everything seems to be passing for all four float types. Ready for a (hopefully) final look.

@rustbot ready

tests/pass/float.rs
```diff
@@ -148,155 +219,367 @@ fn basic() {
...
+            let pow10_max = (10 as $ity).pow(imax.ilog10());
+
+            // If the power of 10 should be representable (fits in a mantissa), check it
+            if ibits - pow10_max.leading_zeros() - pow10_max.trailing_zeros() <= fsig_bits + 1 {
+                assert_bidir(pow10_max as $fty, pow10_max, "pow10 max");
+            }
+        }
+
+        // Tests for when there are some integers that do not have an exact representation
+        if int_range_rep && !all_ints_exact_rep {
```
RalfJung, line 361: What goes wrong with this test if `all_ints_exact_rep` is true?

tests/pass/float.rs
```diff
@@ -148,155 +219,367 @@ fn basic() {
...
+            let max_rep = sat_mantissa << (sat_mantissa.leading_zeros() - adj);
+
+            // This value should roundtrip exactly
+            assert_bidir(max_rep as $fty, max_rep, "max representable int");
+
+            // The cutoff for where to round to infinity is halfway between the maximum
+            // exactly representable integer and the integer's max. This should round to
+            // the maximum representation.
+            let max_non_inf = max_rep + ((imax - max_rep) / 2);
+            assert_bidir(max_non_inf as $fty, max_rep, "max non infinite int");
```
RalfJung, line 377: I am confused how both of these assert_bidir can pass... max_rep cast to the float type can only return one of max_rep and max_non_inf, right?

tests/pass/float.rs
```diff
@@ -148,155 +219,367 @@ fn basic() {
...
+
+            // So the next value up should round to infinity
+            assert_bidir_unrep((max_non_inf + 1) as $fty, imax, "min infinite int");
```
RalfJung, line 380: There's no infinity here...?

RalfJung (commented):
Great job at figuring this out! I am following... maybe half of this.^^ But you left a ton of comments so I hope if someone ever needs to dig into this again they will be helpful...

I do have some kind of naive questions though.

tests/pass/float.rs
tgross35: The math just doesn't work out as nicely because it hits more edge cases, e.g. shifts overflow (I was kind of surprised we don't have a saturating_shl). It could probably be adjusted to make it work, but doesn't seem purposeful because this tests rounding. I'll update the comment to mention rounding.

tests/pass/float.rs
tgross35: Round to the float's infinity in the `as` cast, which then saturates to the integer's max. I'll make this more clear.

tgross35: Sorry no that's wrong, you're right I just meant the integer's max. Will update the comment.

tests/pass/float.rs
RalfJung: Ah okay so really `max_rep as $fty == max_non_sat as $fty` and then the rest follows from that?

tests/pass/float.rs
```diff
@@ -148,155 +219,366 @@ fn basic() {
...
+            // Positive signed integers never use their top bit, so shift by one bit fewer.
+            let sat_mantissa: $ity = (1 << (fsig_bits + 1)) - 1;
+            let adj = if isigned { 1 } else { 0 };
+            let max_rep = sat_mantissa << (sat_mantissa.leading_zeros() - adj);
+
+            // This value should roundtrip exactly
+            assert_bidir(max_rep as $fty, max_rep, "max representable int");
+
+            // The cutoff for where to round to `imax` is halfway between the maximum exactly
+            // representable integer and `imax`. This should round down (to `max_rep`).
```
RalfJung, line 374:
```suggestion
            // representable integer and `imax`. This should round down (to `max_rep`),
            // i.e., `max_rep as $fty == max_non_sat as $fty`.
```

RalfJung:
In future please avoid rebases unless here are conflicts... it seems I now have to way to see the diff of what you changed since my earlier review. :/

tests/pass/float.rs
```diff
@@ -148,155 +219,366 @@ fn basic() {
...
+            // Max representable power of 10
+            let pow10_max = (10 as $ity).pow(imax.ilog10());
+
+            // If the power of 10 should be representable (fits in a mantissa), check it
+            if ibits - pow10_max.leading_zeros() - pow10_max.trailing_zeros() <= fsig_bits + 1 {
+                assert_bidir(pow10_max as $fty, pow10_max, "pow10 max");
+            }
+        }
+
+        // Test rounding when some integers that do not have an exact representation.
```
RalfJung, line 360:
```suggestion
        // Test rounding the largest and smallest integers, but skip this when
        // all integers have an exact representation (it's less interesting then and the arithmetic gets more complicated).
```

tests/pass/float.rs
```diff
@@ -148,155 +219,366 @@ fn basic() {
...
+                    << <$f2>::SIGNIFICAND_BITS,
+            );
+            assert_biteq(min_sub_casted, sub_casted, "min subnormal");
+            assert_biteq(min_neg_sub_casted, -sub_casted, "min neg subnormal");
+        }
+    }};
+}
+
 /// Many of these test values are taken from
 /// https://github.com/WebAssembly/testsuite/blob/master/conversions.wast.
```
RalfJung, line 470: Are these values still present? The comment as it is doesn't seem right any more, either this should be moved down to the spot checks or... not sure where.

tests/pass/float.rs
tgross35: Yes, that is correct

tests/pass/float.rs
tgross35: I just updated it to say the patterns were taken from the values there, which is still true since I just made the cases generic.

RalfJung:
All right, I think we're good to go. Thanks for your patience through all these reviews!

Please squash the commits (ideally in a way that the force push has no diff).

tgross35:
> In future please avoid rebases unless there are conflicts or other reasons that _require_ a rebase... it seems I now have no way to see the diff of what you changed since my earlier review. :/

Do you mean a squashes or specifically rebase onto latest? I'm tend to aggressively squash / fixup to keep commit history clean, because I commit frequently and otherwise sometimes things get merged with a messy log (drives me crazy...). The actual change of base wasn't intentional, I just did it from a different checkout (need to get in the habit of merge-base, but I can never remember the exact incantation to get things right).

In any case the change from my previous push is in https://github.com/rust-lang/miri/compare/dfddc6856e10ecf25f9808d83f5818430fc0ce67..bbfe57710ce7072ef7417188f2d3b1891382ea01#diff-64fd96f1f6fb00d3a7940a8e5c347ce262ea751844538a2c87d4b17574a01f51. Wish GH was just better hiding change-of-base differences.

So I'll just wait until you're ready and then squash one last time 😆

tgross35:
Okiedoke, squashed. Thanks for all the reviews!

RalfJung:
Force-pushing during review in general isn't great since github tends to lose diff then, but change-of-base makes it a *lot* worse. If you commit frequently you can still squash the *new* commits before you push to avoid force-pushes. :D

@bors r+

PR #3689, opened 2024-06-20
Fix ICE caused by seeking past `i64::MAX`

### LINKED ISSUE #3680: Trying to SeekFrom::Start past i64::MAX ICEs Miri
This program (reduced from an ICE encountered when running the tests for https://crates.io/crates/tempfile):
```rust
use std::io::Seek;

fn main() {
    let mut f = std::fs::File::create("test").unwrap();
    f.seek(std::io::SeekFrom::Start(i64::MAX as u64 + 1)).unwrap();
}
```
will ICE with:
```
thread 'rustc' panicked at src/tools/miri/src/shims/unix/fs.rs:398:51:
called `Result::unwrap()` on an `Err` value: TryFromIntError(())
```
The problem is the standard library casts our `u64` `SeekFrom` to an `i64` (`off64_t` is `i64`):
https://github.com/rust-lang/rust/blob/55cac26a9ef17da1c9c77c0816e88e178b7cc5dd/library/std/src/sys/pal/unix/fs.rs#L1296-L1302
```rust
            // Casting to `i64` is fine, too large values will end up as
            // negative which will cause an error in `lseek64`.
            SeekFrom::Start(off) => (libc::SEEK_SET, off as i64),
            SeekFrom::End(off) => (libc::SEEK_END, off),
            SeekFrom::Current(off) => (libc::SEEK_CUR, off),
        };
        let n = cvt(unsafe { lseek64(self.as_raw_fd(), pos as off64_t, whence) })?;
```
But we just convert the `i64` to an `i128` (with `into`):
https://github.com/rust-lang/miri/blob/60a720040d6c60656ab9cac0980e587d19d9c07a/src/shims/unix/foreign_items.rs#L159-L161

Then we try to convert the `i128` numerically into a `u64`:
https://github.com/rust-lang/miri/blob/60a720040d6c60656ab9cac0980e587d19d9c07a/src/shims/unix/fs.rs#L392-L398
... even though the standard library expects us to just return an error for a negative seek offset. We should do that.

adwinwhite:
Make Miri behave the same as standard library on file seeking offset.

Fixes #3680.

### DIFF
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -395,7 +395,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // Isolation check is done via `FileDescriptor` trait.
 
         let seek_from = if whence == this.eval_libc_i32("SEEK_SET") {
-            SeekFrom::Start(u64::try_from(offset).unwrap())
+            if offset < 0 {
+                // Negative offsets return `EINVAL`.
+                let einval = this.eval_libc("EINVAL");
+                this.set_last_error(einval)?;
+                return Ok(Scalar::from_i64(-1));
+            } else {
+                SeekFrom::Start(u64::try_from(offset).unwrap())
+            }
         } else if whence == this.eval_libc_i32("SEEK_CUR") {
             SeekFrom::Current(i64::try_from(offset).unwrap())
         } else if whence == this.eval_libc_i32("SEEK_END") {
--- a/tests/pass/issues/issue-miri-3680.rs
+++ b/tests/pass/issues/issue-miri-3680.rs
@@ -1,0 +1,21 @@
+//@ignore-target-windows: File handling is not implemented yet
+//@compile-flags: -Zmiri-disable-isolation
+
+use std::fs::remove_file;
+use std::io::{ErrorKind, Seek};
+
+#[path = "../../utils/mod.rs"]
+mod utils;
+
+fn main() {
+    let path = utils::prepare("miri_test_fs_seek_i64_max_plus_1.txt");
+
+    let mut f = std::fs::File::create(&path).unwrap();
+    let error = f.seek(std::io::SeekFrom::Start(i64::MAX as u64 + 1)).unwrap_err();
+
+    // It should be error due to negative offset.
+    assert_eq!(error.kind(), ErrorKind::InvalidInput);
+
+    // Cleanup
+    remove_file(&path).unwrap();
+}


src/shims/unix/fs.rs
```diff
@@ -395,7 +395,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // Isolation check is done via `FileDescriptor` trait.
 
         let seek_from = if whence == this.eval_libc_i32("SEEK_SET") {
-            SeekFrom::Start(u64::try_from(offset).unwrap())
+            // Be consistent with standard library.
+            #[allow(clippy::cast_sign_loss)]
+            SeekFrom::Start(i64::try_from(offset).unwrap() as u64)
```
RalfJung, line 400: We're implementing POSIX lseek here, it doesn't matter what the Rust standard library does.

The real question is what does POSIX say should happen when `offset` is negative in SEEK_SET mode?

tests/pass/issues/issue-miri-3680.rs
```diff
@@ -0,0 +1,24 @@
...
+
+#[path = "../../utils/mod.rs"]
+mod utils;
+
+fn main() {
+    let path = utils::prepare("miri_test_fs_seek_i64_max_plus_1.txt");
+
+    let result = catch_unwind(|| {
+        let mut f = std::fs::File::create(&path).unwrap();
+        f.seek(std::io::SeekFrom::Start(i64::MAX as u64 + 1)).unwrap();
```
RalfJung, line 16: Looks like on a Windows host, we are getting "unsupported io error". Curious that this doesn't happen on Linux hosts, I guess they generate different `io::ErrorKind` for this case?

RalfJung: There's no named constant for error 131, so on windows this becomes `ErrorKind::Other` and then we can't map it to a target error.

I think if the offset is negative we probably don't want to call `file_descriptor.seek` at all, and instead return EINVAL directly ourselves.

adwinwhite: Indeed. So I guess test should only be run for unix host. No idea how to configure this, but `@only-host-linux` is good enough?

Sounds good.

RalfJung: No it should run on all hosts, Miri should just behave consistently.

adwinwhite: Oh, I got it. We only need to know Miri can interpret this operation successfully without ICE so the stderr output doesn't matter.
How can I configure the test header for this?  I tried `@rustc-env: MIRI_SKIP_UI_CHECKS=1` and it still compares output.

RalfJung: No stderr output should be checked.

If you do this...

> I think if the offset is negative we probably don't want to call file_descriptor.seek at all, and instead return EINVAL directly ourselves.

... then I think all the troubles with the test will disappear.

adwinwhite: Understood now. No need to panic on purpose thus no stderr is emitted.

RalfJung: No, that's not what I meant. The error code will be the same across all targets if you  implement what I quoted above, so the stderr will be the same.

src/shims/unix/fs.rs
```diff
@@ -395,7 +395,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // Isolation check is done via `FileDescriptor` trait.
 
         let seek_from = if whence == this.eval_libc_i32("SEEK_SET") {
-            SeekFrom::Start(u64::try_from(offset).unwrap())
+            if offset < 0 {
+                let einval = this.eval_libc("EINVAL");
```
RalfJung, line 399:
```suggestion
                // Negative offsets return `EINVAL`.
                let einval = this.eval_libc("EINVAL");
```

tests/pass/issues/issue-miri-3680.rs
```diff
@@ -0,0 +1,21 @@
...
+mod utils;
+
+fn main() {
+    let path = utils::prepare("miri_test_fs_seek_i64_max_plus_1.txt");
+
+    let mut f = std::fs::File::create(&path).unwrap();
+    let result = f.seek(std::io::SeekFrom::Start(i64::MAX as u64 + 1));
+
+    // It should be error due to negative offset.
+    assert!(result.is_err());
```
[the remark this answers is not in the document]
adwinwhite: But for windows, it's `ErrorKind::Other`. Should I write platform-specific check?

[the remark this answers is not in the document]
adwinwhite: Sorry, I misunderstood the codebase. I thought `miri/src/shims/unix/fs.rs` was platform-specific code and would be used for unix host only.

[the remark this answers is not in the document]
RalfJung: It is used for Unix *targets* only :)

RalfJung:
Looks good, thanks!

@bors r+

PR #3690, opened 2024-06-20
Implement LLVM x86 adx intrinsics

TDecking:
See title. It also explots a small opportunity to deduplicate a bit of intrinsics code.

### DIFF
--- a/src/shims/x86/mod.rs
+++ b/src/shims/x86/mod.rs
@@ -64,15 +64,20 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             "addcarryx.u32" | "addcarryx.u64" => {
                 this.expect_target_feature_for_intrinsic(link_name, "adx")?;
 
-                if unprefixed_name.ends_with("64") && this.tcx.sess.target.arch != "x86_64" {
+                let is_u64 = unprefixed_name.ends_with("64");
+                if is_u64 && this.tcx.sess.target.arch != "x86_64" {
                     return Ok(EmulateItemResult::NotSupported);
                 }
 
                 let [c_in, a, b, out] = this.check_shim(abi, Abi::Unadjusted, link_name, args)?;
+                let out = this.deref_pointer_as(
+                    out,
+                    if is_u64 { this.machine.layouts.u64 } else { this.machine.layouts.u32 },
+                )?;
 
                 let (sum, c_out) = carrying_add(this, c_in, a, b, mir::BinOp::AddWithOverflow)?;
                 this.write_scalar(c_out, dest)?;
-                this.write_immediate(*sum, &this.deref_pointer_as(out, sum.layout)?)?;
+                this.write_immediate(*sum, &out)?;
             }
 
             // Used to implement the `_mm_pause` function.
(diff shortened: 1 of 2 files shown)


TDecking:
This is likely the last extension I can develop for a while. The remaining ones are either not supprted by my computer or implemented in stdarch using inline assembly.

On a related note, I managed to rewrite some intrinsics in stdarch using generic intrinsics, which means that the next stdarch submodule update over at rust-lang will come with free and full support for x86 fma.

RalfJung:
Oh that's cool, does that mean we can remove some of the intrinsics in Miri as they can no longer be reached via stdarch? :D

TDecking:
The number is small, but yes.

RalfJung:
Good to know, I opened https://github.com/rust-lang/miri/issues/3691 to track that.

src/shims/x86/mod.rs
```diff
@@ -35,63 +35,67 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-            // Used to implement the `_subborrow_u32` and `_subborrow_u64` functions.
-            // Computes a - b with input and output borrow. The input borrow is an 8-bit
-            // value, which is interpreted as 1 if it is non-zero. The output borrow is
-            // an 8-bit value that will be 0 or 1.
-            // https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-8/subborrow-u32-subborrow-u64.html
-            "subborrow.32" | "subborrow.64" => {
-                if unprefixed_name == "subborrow.64" && this.tcx.sess.target.arch != "x86_64" {
+
+            // Used to implement the `_addcarryx_u{32, 64}` functions. They are semantically identical with the `_addcarry_u{32, 64}` functions,
+            // except for a slightly different type signature and the requirement for the "adx" target feature.
```
RalfJung, line 69: Sounds like you could have a helper function that takes the two places to write the result to, and call that in both cases?

src/shims/x86/mod.rs
```diff
@@ -35,63 +35,67 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                     .to_pair(this);
-                let b_out = overflow1.to_scalar().to_bool()? | overflow2.to_scalar().to_bool()?;
+                let c_out = overflow1.to_scalar().to_bool()? | overflow2.to_scalar().to_bool()?;
+
+                this.write_scalar(Scalar::from_u8(c_out.into()), dest)?;
 
-                this.write_scalar(Scalar::from_u8(b_out.into()), &this.project_field(dest, 0)?)?;
-                this.write_immediate(*sub, &this.project_field(dest, 1)?)?;
+                let out = this.read_pointer(out)?;
+                let out = this.ptr_to_mplace(out, sum.layout);
```
RalfJung, line 97: This looks like you can use `deref_pointer_as`

TDecking: I was hoping for this kind of function. Thanks!

src/shims/x86/mod.rs
TDecking: This pattern is too small and does not appear often enough to justify it IMHO. Adding such a function now would increase the number of lines a reader needs to parse.

RalfJung: It's more lines but still less mental effort overall as it avoids having to parse the same logic twice.

TDecking: Ok. Ive added the function.

RalfJung: I was talking about sharing the entire add-with-carry logic, not just writing two values to two places.

TDecking: The addition logic can be moved out, that has been done. Actually unifiying the result writing logic did not look good after one attempt, so I'd prefer if that was left alone.

RalfJung:
That works, thanks. :)

@bors r+

PR #3696, opened 2024-06-21
CI: try to share setup code across actions

RalfJung:


oli-obk:
Oh that failure was a caching issue, that makes sense

@bors r+

Issue #2749, opened 2023-01-06
can't call foreign function: sched_setaffinity


vlovich:
Running a basic test program against glommio:

```
$ MIRIFLAGS="-Zmiri-disable-isolation" cargo +nightly miri run --all-features --verbose
error: unsupported operation: can't call foreign function: sched_setaffinity
   --> /home/vlovich/.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.23.2/src/sched.rs:175:13
    |
175 | /             libc::sched_setaffinity(
176 | |                 pid.into(),
177 | |                 mem::size_of::<CpuSet>() as libc::size_t,
178 | |                 &cpuset.cpu_set,
179 | |             )
    | |_____________^ can't call foreign function: sched_setaffinity
    |
    = help: this is likely not a bug in the program; it indicates that the program performed an operation that the interpreter does not support
    = note: BACKTRACE:
    = note: inside `nix::sched::sched_linux_like::sched_setaffinity` at /home/vlovich/.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.23.2/src/sched.rs:175:13: 179:14
    = note: inside `glommio::executor::bind_to_cpu_set::<std::iter::Map<glommio::executor::placement::CpuIter, [closure@glommio::executor::placement::CpuIter::cpu_binding::{closure#0}]>>` at /home/vlovich/.cargo/registry/src/github.com-1ecc6299db9ec823/glommio-0.7.0/src/executor/mod.rs:220:5: 220:48
```

Not sure if it's expected to not work but figured I'd report it. I'm sure there may be others.

RalfJung:
Yeah that function is not currently implemented by Miri. We could probably give a trivial implementation that does nothing, so this should be fairly easy to do.

PR #3698, opened 2024-06-21
implement `libc::sched_setaffinity` on linux

folkertdev:
fixes https://github.com/rust-lang/miri/issues/2749

the implementation, like `libc::sched_getaffinity`, just always returns `EINVAL`, which kind of simulates a device with zero cpus. I believe the idea is that callers of this function always do it to optimize, so they are likely to gracefully recover from this function returning an error. 

based on the libc crate, these functions are also available on android and freebsd (but not on macos or windows). So should the implementation of the `sched_*` functions just be copied to the android and freebsd shims?

### DIFF
--- a/src/bin/miri.rs
+++ b/src/bin/miri.rs
@@ -592,6 +592,9 @@ fn main() {
             let num_cpus = param
                 .parse::<u32>()
                 .unwrap_or_else(|err| show_error!("-Zmiri-num-cpus requires a `u32`: {}", err));
+            if !(1..=miri::MAX_CPUS).contains(&usize::try_from(num_cpus).unwrap()) {
+                show_error!("-Zmiri-num-cpus must be in the range 1..={}", miri::MAX_CPUS);
+            }
             miri_config.num_cpus = num_cpus;
         } else if let Some(param) = arg.strip_prefix("-Zmiri-force-page-size=") {
             let page_size = param.parse::<u64>().unwrap_or_else(|err| {
--- a/src/concurrency/cpu_affinity.rs
+++ b/src/concurrency/cpu_affinity.rs
@@ -1,5 +1,7 @@
-use crate::bug;
+use rustc_middle::ty::layout::LayoutOf;
 use rustc_target::abi::Endian;
+
+use crate::*;
 
 /// The maximum number of CPUs supported by miri.
 ///
@@ -19,41 +21,34 @@ pub(crate) struct CpuAffinityMask([u8; Self::CPU_MASK_BYTES]);
 impl CpuAffinityMask {
     pub(crate) const CPU_MASK_BYTES: usize = MAX_CPUS / 8;
 
-    pub fn new(target: &rustc_target::spec::Target, cpu_count: u32) -> Self {
+    pub fn new<'tcx>(cx: &impl LayoutOf<'tcx>, cpu_count: u32) -> Self {
         let mut this = Self([0; Self::CPU_MASK_BYTES]);
 
         // the default affinity mask includes only the available CPUs
         for i in 0..cpu_count as usize {
-            this.set(target, i);
+            this.set(cx, i);
         }
 
         this
     }
 
-    pub fn chunk_size(target: &rustc_target::spec::Target) -> u64 {
-        // The actual representation of the CpuAffinityMask is [c_ulong; _], in practice either
-        //
-        // - [u32; 32] on 32-bit platforms
-        // - [u64; 16] everywhere else
-
-        // FIXME: this should be `size_of::<core::ffi::c_ulong>()`
-        u64::from(target.pointer_width / 8)
+    pub fn chunk_size<'tcx>(cx: &impl LayoutOf<'tcx>) -> u64 {
+        // The actual representation of the CpuAffinityMask is [c_ulong; _].
+        let ulong = helpers::path_ty_layout(cx, &["core", "ffi", "c_ulong"]);
+        ulong.size.bytes()
     }
 
-    fn set(&mut self, target: &rustc_target::spec::Target, cpu: usize) {
+    fn set<'tcx>(&mut self, cx: &impl LayoutOf<'tcx>, cpu: usize) {
         // we silently ignore CPUs that are out of bounds. This matches the behavior of
         // `sched_setaffinity` with a mask that specifies more than `CPU_SETSIZE` CPUs.
         if cpu >= MAX_CPUS {
             return;
         }
 
-        // The actual representation of the CpuAffinityMask is [c_ulong; _], in practice either
-        //
-        // - [u32; 32] on 32-bit platforms
-        // - [u64; 16] everywhere else
-        //
+        // The actual representation of the CpuAffinityMask is [c_ulong; _].
         // Within the array elements, we need to use the endianness of the target.
-        match Self::chunk_size(target) {
+        let target = &cx.tcx().sess.target;
+        match Self::chunk_size(cx) {
             4 => {
                 let start = cpu / 32 * 4; // first byte of the correct u32
                 let chunk = self.0[start..].first_chunk_mut::<4>().unwrap();
@@ -72,7 +67,7 @@ impl CpuAffinityMask {
                     Endian::Big => (u64::from_be_bytes(*chunk) | 1 << offset).to_be_bytes(),
                 };
             }
-            other => bug!("other chunk sizes are not supported: {other}"),
+            other => bug!("chunk size not supported: {other}"),
         };
     }
 
@@ -80,13 +75,13 @@ impl CpuAffinityMask {
         self.0.as_slice()
     }
 
-    pub fn from_array(
-        target: &rustc_target::spec::Target,
+    pub fn from_array<'tcx>(
+        cx: &impl LayoutOf<'tcx>,
         cpu_count: u32,
         bytes: [u8; Self::CPU_MASK_BYTES],
     ) -> Option<Self> {
         // mask by what CPUs are actually available
-        let default = Self::new(target, cpu_count);
+        let default = Self::new(cx, cpu_count);
         let masked = std::array::from_fn(|i| bytes[i] & default.0[i]);
 
         // at least one thread must be set for the input to be valid
--- a/src/concurrency/cpu_affinity.rs
+++ b/src/concurrency/cpu_affinity.rs
@@ -1,0 +1,95 @@
+use crate::bug;
+use rustc_target::abi::Endian;
+
+/// The maximum number of CPUs supported by miri.
+///
+/// This value is compatible with the libc `CPU_SETSIZE` constant and corresponds to the number
+/// of CPUs that a `cpu_set_t` can contain.
+///
+/// Real machines can have more CPUs than this number, and there exist APIs to set their affinity,
+/// but this is not currently supported by miri.
+pub const MAX_CPUS: usize = 1024;
+
+/// A thread's CPU affinity mask determines the set of CPUs on which it is eligible to run.
+// the actual representation depends on the target's endianness and pointer width.
+// See CpuAffinityMask::set for details
+#[derive(Clone)]
+pub(crate) struct CpuAffinityMask([u8; Self::CPU_MASK_BYTES]);
+
+impl CpuAffinityMask {
+    pub(crate) const CPU_MASK_BYTES: usize = MAX_CPUS / 8;
+
+    pub fn new(target: &rustc_target::spec::Target, cpu_count: u32) -> Self {
+        let mut this = Self([0; Self::CPU_MASK_BYTES]);
+
+        // the default affinity mask includes only the available CPUs
+        for i in 0..cpu_count as usize {
+            this.set(target, i);
+        }
+
+        this
+    }
+
+    pub fn chunk_size(target: &rustc_target::spec::Target) -> u64 {
+        // The actual representation of the CpuAffinityMask is [c_ulong; _], in practice either
+        //
+        // - [u32; 32] on 32-bit platforms
+        // - [u64; 16] everywhere else
+
+        // FIXME: this should be `size_of::<core::ffi::c_ulong>()`
+        u64::from(target.pointer_width / 8)
+    }
+
+    fn set(&mut self, target: &rustc_target::spec::Target, cpu: usize) {
+        // we silently ignore CPUs that are out of bounds. This matches the behavior of
+        // `sched_setaffinity` with a mask that specifies more than `CPU_SETSIZE` CPUs.
+        if cpu >= MAX_CPUS {
+            return;
+        }
+
+        // The actual representation of the CpuAffinityMask is [c_ulong; _], in practice either
+        //
+        // - [u32; 32] on 32-bit platforms
+        // - [u64; 16] everywhere else
+        //
+        // Within the array elements, we need to use the endianness of the target.
+        match Self::chunk_size(target) {
+            4 => {
+                let start = cpu / 32 * 4; // first byte of the correct u32
+                let chunk = self.0[start..].first_chunk_mut::<4>().unwrap();
+                let offset = cpu % 32;
+                *chunk = match target.options.endian {
+                    Endian::Little => (u32::from_le_bytes(*chunk) | 1 << offset).to_le_bytes(),
+                    Endian::Big => (u32::from_be_bytes(*chunk) | 1 << offset).to_be_bytes(),
+                };
+            }
+            8 => {
+                let start = cpu / 64 * 8; // first byte of the correct u64
+                let chunk = self.0[start..].first_chunk_mut::<8>().unwrap();
+                let offset = cpu % 64;
+                *chunk = match target.options.endian {
+                    Endian::Little => (u64::from_le_bytes(*chunk) | 1 << offset).to_le_bytes(),
+                    Endian::Big => (u64::from_be_bytes(*chunk) | 1 << offset).to_be_bytes(),
+                };
+            }
+            other => bug!("other chunk sizes are not supported: {other}"),
+        };
+    }
+
+    pub fn as_slice(&self) -> &[u8] {
+        self.0.as_slice()
+    }
+
+    pub fn from_array(
+        target: &rustc_target::spec::Target,
+        cpu_count: u32,
+        bytes: [u8; Self::CPU_MASK_BYTES],
+    ) -> Option<Self> {
+        // mask by what CPUs are actually available
+        let default = Self::new(target, cpu_count);
+        let masked = std::array::from_fn(|i| bytes[i] & default.0[i]);
+
+        // at least one thread must be set for the input to be valid
+        masked.iter().any(|b| *b != 0).then_some(Self(masked))
+    }
+}
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -936,6 +936,11 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // After this all accesses will be treated as occurring in the new thread.
         let old_thread_id = this.machine.threads.set_active_thread_id(new_thread_id);
 
+        // The child inherits its parent's cpu affinity.
+        if let Some(cpuset) = this.machine.thread_cpu_affinity.get(&old_thread_id).cloned() {
+            this.machine.thread_cpu_affinity.insert(new_thread_id, cpuset);
+        }
+
         // Perform the function pointer load in the new thread frame.
         let instance = this.get_ptr_fn(start_routine)?.as_instance()?;
 
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -30,6 +30,7 @@ use rustc_target::spec::abi::Abi;
 
 use crate::{
     concurrency::{
+        cpu_affinity::{self, CpuAffinityMask},
         data_race::{self, NaReadType, NaWriteType},
         weak_memory,
     },
@@ -471,6 +472,12 @@ pub struct MiriMachine<'tcx> {
 
     /// The set of threads.
     pub(crate) threads: ThreadManager<'tcx>,
+
+    /// Stores which thread is eligible to run on which CPUs.
+    /// This has no effect at all, it is just tracked to produce the correct result
+    /// in `sched_getaffinity`
+    pub(crate) thread_cpu_affinity: FxHashMap<ThreadId, CpuAffinityMask>,
+
     /// The state of the primitive synchronization objects.
     pub(crate) sync: SynchronizationObjects,
 
@@ -627,6 +634,20 @@ impl<'tcx> MiriMachine<'tcx> {
         let stack_addr = if tcx.pointer_size().bits() < 32 { page_size } else { page_size * 32 };
         let stack_size =
             if tcx.pointer_size().bits() < 32 { page_size * 4 } else { page_size * 16 };
+        assert!(
+            usize::try_from(config.num_cpus).unwrap() <= cpu_affinity::MAX_CPUS,
+            "miri only supports up to {} CPUs, but {} were configured",
+            cpu_affinity::MAX_CPUS,
+            config.num_cpus
+        );
+        let threads = ThreadManager::default();
+        let mut thread_cpu_affinity = FxHashMap::default();
+        if matches!(&*tcx.sess.target.os, "linux" | "freebsd" | "android") {
+            thread_cpu_affinity.insert(
+                threads.active_thread(),
+                CpuAffinityMask::new(&tcx.sess.target, config.num_cpus),
+            );
+        }
         MiriMachine {
             tcx,
             borrow_tracker,
@@ -644,7 +665,8 @@ impl<'tcx> MiriMachine<'tcx> {
             fds: shims::FdTable::new(config.mute_stdout_stderr),
             dirs: Default::default(),
             layouts,
-            threads: ThreadManager::default(),
+            threads,
+            thread_cpu_affinity,
             sync: SynchronizationObjects::default(),
             static_roots: Vec::new(),
             profiler,
@@ -765,6 +787,7 @@ impl VisitProvenance for MiriMachine<'_> {
         #[rustfmt::skip]
         let MiriMachine {
             threads,
+            thread_cpu_affinity: _,
             sync: _,
             tls,
             env_vars,
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -643,10 +643,8 @@ impl<'tcx> MiriMachine<'tcx> {
         let threads = ThreadManager::default();
         let mut thread_cpu_affinity = FxHashMap::default();
         if matches!(&*tcx.sess.target.os, "linux" | "freebsd" | "android") {
-            thread_cpu_affinity.insert(
-                threads.active_thread(),
-                CpuAffinityMask::new(&tcx.sess.target, config.num_cpus),
-            );
+            thread_cpu_affinity
+                .insert(threads.active_thread(), CpuAffinityMask::new(&layout_cx, config.num_cpus));
         }
         MiriMachine {
             tcx,
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -594,11 +594,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     _ => throw_unsup_format!("`sched_getaffinity` is only supported with a pid of 0 (indicating the current thread)"),
                 };
 
-                // The actual representation of the CpuAffinityMask is [c_ulong; _], in practice either
-                //
-                // - [u32; 32] on 32-bit platforms
-                // - [u64; 16] everywhere else
-                let chunk_size = CpuAffinityMask::chunk_size(&this.tcx.sess.target);
+                // The mask is stored in chunks, and the size must be a whole number of chunks.
+                let chunk_size = CpuAffinityMask::chunk_size(this);
 
                 if this.ptr_is_null(mask)? {
                     let einval = this.eval_libc("EFAULT");
@@ -643,7 +640,6 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     _ => throw_unsup_format!("`sched_setaffinity` is only supported with a pid of 0 (indicating the current thread)"),
                 };
 
-                #[allow(clippy::map_entry)]
                 if this.ptr_is_null(mask)? {
                     let einval = this.eval_libc("EFAULT");
                     this.set_last_error(einval)?;
@@ -652,9 +648,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     // NOTE: cpusetsize might be smaller than `CpuAffinityMask::CPU_MASK_BYTES`
                     let bits_slice = this.read_bytes_ptr_strip_provenance(mask, Size::from_bytes(cpusetsize))?;
                     // This ignores the bytes beyond `CpuAffinityMask::CPU_MASK_BYTES`
-                    let bits_array: [u8;CpuAffinityMask::CPU_MASK_BYTES] =
+                    let bits_array: [u8; CpuAffinityMask::CPU_MASK_BYTES] =
                         std::array::from_fn(|i| bits_slice.get(i).copied().unwrap_or(0));
-                    match CpuAffinityMask::from_array(&this.tcx.sess.target, this.machine.num_cpus, bits_array) {
+                    match CpuAffinityMask::from_array(this, this.machine.num_cpus, bits_array) {
                         Some(cpuset) => {
                             this.machine.thread_cpu_affinity.insert(thread_id, cpuset);
                             this.write_scalar(Scalar::from_i32(0), dest)?;
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -3,8 +3,10 @@ use std::str;
 
 use rustc_middle::ty::layout::LayoutOf;
 use rustc_span::Symbol;
+use rustc_target::abi::Size;
 use rustc_target::spec::abi::Abi;
 
+use crate::concurrency::cpu_affinity::CpuAffinityMask;
 use crate::shims::alloc::EvalContextExt as _;
 use crate::shims::unix::*;
 use crate::*;
@@ -570,6 +572,101 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let [req, rem] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
                 let result = this.nanosleep(req, rem)?;
                 this.write_scalar(Scalar::from_i32(result), dest)?;
+            }
+            "sched_getaffinity" => {
+                // Currently this function does not exist on all Unixes, e.g. on macOS.
+                if !matches!(&*this.tcx.sess.target.os, "linux" | "freebsd" | "android") {
+                    throw_unsup_format!(
+                        "`sched_getaffinity` is not supported on {}",
+                        this.tcx.sess.target.os
+                    );
+                }
+
+                let [pid, cpusetsize, mask] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let pid = this.read_scalar(pid)?.to_u32()?;
+                let cpusetsize = this.read_target_usize(cpusetsize)?;
+                let mask = this.read_pointer(mask)?;
+
+                // TODO: when https://github.com/rust-lang/miri/issues/3730 is fixed this should use its notion of tid/pid
+                let thread_id = match pid {
+                    0 => this.active_thread(),
+                    _ => throw_unsup_format!("`sched_getaffinity` is only supported with a pid of 0 (indicating the current thread)"),
+                };
+
+                // The actual representation of the CpuAffinityMask is [c_ulong; _], in practice either
+                //
+                // - [u32; 32] on 32-bit platforms
+                // - [u64; 16] everywhere else
+                let chunk_size = CpuAffinityMask::chunk_size(&this.tcx.sess.target);
+
+                if this.ptr_is_null(mask)? {
+                    let einval = this.eval_libc("EFAULT");
+                    this.set_last_error(einval)?;
+                    this.write_scalar(Scalar::from_i32(-1), dest)?;
+                } else if cpusetsize == 0 || cpusetsize.checked_rem(chunk_size).unwrap() != 0 {
+                    // we only copy whole chunks of size_of::<c_ulong>()
+                    let einval = this.eval_libc("EINVAL");
+                    this.set_last_error(einval)?;
+                    this.write_scalar(Scalar::from_i32(-1), dest)?;
+                } else if let Some(cpuset) = this.machine.thread_cpu_affinity.get(&thread_id) {
+                    let cpuset = cpuset.clone();
+                    // we only copy whole chunks of size_of::<c_ulong>()
+                    let byte_count = Ord::min(cpuset.as_slice().len(), cpusetsize.try_into().unwrap());
+                    this.write_bytes_ptr(mask, cpuset.as_slice()[..byte_count].iter().copied())?;
+                    this.write_scalar(Scalar::from_i32(0), dest)?;
+                } else {
+                    // The thread whose ID is pid could not be found
+                    let einval = this.eval_libc("ESRCH");
+                    this.set_last_error(einval)?;
+                    this.write_scalar(Scalar::from_i32(-1), dest)?;
+                }
+            }
+            "sched_setaffinity" => {
+                // Currently this function does not exist on all Unixes, e.g. on macOS.
+                if !matches!(&*this.tcx.sess.target.os, "linux" | "freebsd" | "android") {
+                    throw_unsup_format!(
+                        "`sched_setaffinity` is not supported on {}",
+                        this.tcx.sess.target.os
+                    );
+                }
+
+                let [pid, cpusetsize, mask] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let pid = this.read_scalar(pid)?.to_u32()?;
+                let cpusetsize = this.read_target_usize(cpusetsize)?;
+                let mask = this.read_pointer(mask)?;
+
+                // TODO: when https://github.com/rust-lang/miri/issues/3730 is fixed this should use its notion of tid/pid
+                let thread_id = match pid {
+                    0 => this.active_thread(),
+                    _ => throw_unsup_format!("`sched_setaffinity` is only supported with a pid of 0 (indicating the current thread)"),
+                };
+
+                #[allow(clippy::map_entry)]
+                if this.ptr_is_null(mask)? {
+                    let einval = this.eval_libc("EFAULT");
+                    this.set_last_error(einval)?;
+                    this.write_scalar(Scalar::from_i32(-1), dest)?;
+                } else {
+                    // NOTE: cpusetsize might be smaller than `CpuAffinityMask::CPU_MASK_BYTES`
+                    let bits_slice = this.read_bytes_ptr_strip_provenance(mask, Size::from_bytes(cpusetsize))?;
+                    // This ignores the bytes beyond `CpuAffinityMask::CPU_MASK_BYTES`
+                    let bits_array: [u8;CpuAffinityMask::CPU_MASK_BYTES] =
+                        std::array::from_fn(|i| bits_slice.get(i).copied().unwrap_or(0));
+                    match CpuAffinityMask::from_array(&this.tcx.sess.target, this.machine.num_cpus, bits_array) {
+                        Some(cpuset) => {
+                            this.machine.thread_cpu_affinity.insert(thread_id, cpuset);
+                            this.write_scalar(Scalar::from_i32(0), dest)?;
+                        }
+                        None => {
+                            // The intersection between the mask and the available CPUs was empty.
+                            let einval = this.eval_libc("EINVAL");
+                            this.set_last_error(einval)?;
+                            this.write_scalar(Scalar::from_i32(-1), dest)?;
+                        }
+                    }
+                }
             }
 
             // Miscellaneous
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -645,7 +645,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     this.set_last_error(einval)?;
                     this.write_scalar(Scalar::from_i32(-1), dest)?;
                 } else {
-                    // NOTE: cpusetsize might be smaller than `CpuAffinityMask::CPU_MASK_BYTES`
+                    // NOTE: cpusetsize might be smaller than `CpuAffinityMask::CPU_MASK_BYTES`.
+                    // Any unspecified bytes are treated as zero here (none of the CPUs are configured).
+                    // This is not exactly documented, so we assume that this is the behavior in practice.
                     let bits_slice = this.read_bytes_ptr_strip_provenance(mask, Size::from_bytes(cpusetsize))?;
                     // This ignores the bytes beyond `CpuAffinityMask::CPU_MASK_BYTES`
                     let bits_array: [u8; CpuAffinityMask::CPU_MASK_BYTES] =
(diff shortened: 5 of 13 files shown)


src/shims/unix/linux/foreign_items.rs
```diff
@@ -191,6 +191,21 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.set_last_error(einval)?;
                 this.write_scalar(Scalar::from_i32(-1), dest)?;
             }
+            "sched_setaffinity" => {
+                // a no-op. The assumption is that application code should be able to recover from
+                // this operation failing with EINVAL, which in this case indicates that the
+                // affinity bit mask does not contain any processors that are physically on the
+                // system.
```
saethlin, line 198: I don't think this is a valid assumption; wouldn't application code using an API based on `sched_setaffinity` have just queried which processors are on the system, and thus be confident that this error case won't be hit?

saethlin: We have a a flag `-Zmiri-num-cpus`, which defaults to `1` (and I for one routinely set to large numbers). I'd prefer if the behavior of this shim were based on that flag.

folkertdev: well, conveniently miri's `libc::sched_getaffinity` also always returns `EINVAL`, so yeah in proper code I guess  you'd never actually reach the `libc::sched_setaffinity`.

I think the real reason for this implementation is that miri doesn't want to simulate this function properly, so returning an error is the only thing that is left to be done.

saethlin: Ouch. I see your line of thinking now.

RalfJung:
The original motivation stated in the issue was a call site in nix-0.23.2/src/sched.rs:175. What does that code do when EINVAL is returned? Does this help Miri do anything useful in that code?

folkertdev:
Allright, I dug into `glommi`'s usage os these functions. for instance they have 

```rust 
#[test]
fn blocking_function_placement_independent_of_executor_placement() {
    let affinity = nix::sched::sched_getaffinity(nix::unistd::Pid::from_raw(0)).unwrap();
```

this test will just fail on miri because `sched_getaffinity` always returns EINVAL. The only benefit that you get from miri supporting this function at all is that the arguments to `libc::sched_getaffinity` are checked (e.g. if you pass in a NULL pointer miri would yell at you about that). The `executor_pool_builder_placements` test would also hit an unwrap on the result of `sched_getaffinity`, but it errors earlier on an unsupported syscall:

```
error: unsupported operation: can't execute syscall with ID 324
    --> glommio/src/sys/membarrier.rs:67:18
     |
67   |         unsafe { libc::syscall(libc::SYS_membarrier, cmd as libc::c_int, 0 as libc::c_int) }
     |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't execute syscall with ID 324
```

That's a separate issue though (no idea how miri should handle that).

---

Then `sched_setaffinity`. In `glommio` it is wrapped in this function


```rust
pub(crate) fn bind_to_cpu_set(cpus: impl IntoIterator<Item = usize>) -> Result<()> {
    let mut cpuset = nix::sched::CpuSet::new();
    for cpu in cpus {
        cpuset.set(cpu).map_err(|e| to_io_error!(e))?;
    }
    let pid = nix::unistd::Pid::from_raw(0);
    nix::sched::sched_setaffinity(pid, &cpuset).map_err(|e| Into::into(to_io_error!(e)))
}
```

which is tested here 

```rust
#[test]
fn bind_to_cpu_set_range() {
    // libc supports cpu ids up to 1023 and will use the intersection of values
    // specified by the cpu mask and those present on the system
    // https://man7.org/linux/man-pages/man2/sched_setaffinity.2.html#NOTES
    assert!(bind_to_cpu_set(vec![0, 1, 2, 3]).is_ok());
    assert!(bind_to_cpu_set(0..1024).is_ok());
    assert!(bind_to_cpu_set(0..1025).is_err());
}
```

So this will just fail with miri. Based on [the man page](https://man7.org/linux/man-pages/man2/sched_setaffinity.2.html) this will also fail in general (i.e. even without miri) when

> The affinity bit mask mask contains no processors that are currently physically on the system and permitted to the thread according to any restrictions that may be imposed by cpuset cgroups or the "cpuset" mechanism described in cpuset(7).

but this test seems to just assume that that did not happen, and up to 1024 cpus can be configured for the current thread.

---

I think in general that means that the implementations of `{g, s}et_affinity` are not that useful for glommio at least, besided checking some safety properties on the arguments.

I'm not sure how miri could do better though. It would have to keep track of the affinity (1024 bits per thread id, with pid 0 mapping to the current thread) in some piece of global state?

RalfJung:
We can have a simple-but-wrong implementation that happily says "ok" on setaffinity but just ignores this and returns the full mask on getaffinity. Or we track the per-thread mask in some per-thread state, so that getaffinity can return the right result. Either way it would not have any other effect. I don't have a strong preference either way, but maybe some apps get confused when their setaffinity is not reflected in getaffinity.

> based on the libc crate, these functions are also available on android and freebsd (but not on macos or windows). So should the implementation of the sched_* functions just be copied to the android and freebsd shims?

No copy-paste please. You can do it like `reallocarray` -- have the impl in the generic unix file and guard it on what the concrete OS is.

folkertdev:
are there examples of such per-thread state?

I don't think the simple-but-wrong approach is really more useful than the status quo.

RalfJung:
I'd suggest you add a new `FxHashMap<ThreadId, ...>` to `MiriMachine`.

folkertdev:
@rustbot ready

allright, the implementation now simulates the logic in that setting and then getting the mask returns the updated value. the initial value of the mask is based on `-Zmiri-num-cpus`.

folkertdev:
apparently for this target no CPUs are configured? specifically the `s390x-unknown-linux-gnu` target fails when executed on macos?!

```
  Error: 
     0: ui tests in tests/pass-dep for s390x-unknown-linux-gnu failed
     1: tests failed

  full stderr:
  thread 'main' panicked at tests/pass-dep/libc/libc-misc.rs:80:5:
  assertion failed: unsafe { libc::CPU_ISSET(0, &cpuset) }
  stack backtrace:
     0: std::panicking::begin_panic_handler
               at /Users/runner/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/panicking.rs:661:5
     1: core::panicking::panic_fmt
               at /Users/runner/.rustup/toolchains/miri/lib/rustlib/src/rust/library/core/src/panicking.rs:74:14
     2: core::panicking::panic
               at /Users/runner/.rustup/toolchains/miri/lib/rustlib/src/rust/library/core/src/panicking.rs:148:5
     3: test_affinity
               at tests/pass-dep/libc/libc-misc.rs:80:5
     4: main
               at tests/pass-dep/libc/libc-misc.rs:141:9
     5: <fn() as std::ops::FnOnce<()>>::call_once - shim(fn())
               at /Users/runner/.rustup/toolchains/miri/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
  note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

https://github.com/rust-lang/miri/actions/runs/9627551162/job/26554759066?pr=3698

we can just skip the test on this platform but it is weird.

RalfJung:
s390x is a big-endian architecture, so likely your code is simply not correct for big-endian systems.

src/concurrency/thread.rs
```diff
@@ -924,6 +924,12 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             data_race.thread_created(&this.machine.threads, new_thread_id, current_span);
         }
 
+        // the child inherits its parent's cpu affinity
+        let parent = this.active_thread();
```
RalfJung, line 928: This is the same as `old_thread_id` below, isn't it?

src/machine.rs
```diff
@@ -420,6 +420,34 @@ impl<'tcx> PrimitiveLayouts<'tcx> {
     }
 }
 
+/// A thread's CPU affinity mask determines the set of CPUs on which it is eligible to run.
```
RalfJung, line 423: What's the format for this? 1024 bits, each encoding one core? Why 1024?

Also, how does endianess factor in? From the comments below it seems like on little-endian, CPU N is encoded in element `N / 8`, bit `N % 8`, counting bits from least significant to most significant.

src/machine.rs
```diff
@@ -420,6 +420,34 @@ impl<'tcx> PrimitiveLayouts<'tcx> {
...
+        let mut bits = [0; 128];
+
+        // set the first cpu_count bits. The format is as follows:
+        //
+        // cpu 0    : [ 0b0000_0001, 0, 0, ..., 0 ]
+        // cpu 0, 1 : [ 0b0000_0011, 0, 0, ..., 0 ]
+        // cpu 0, 8 : [ 0b0000_0001, 0b0000_0001, 0, ..., 0 ]
+        for byte in bits.iter_mut() {
+            // set the lowest N bits of a byte
+            *byte = (0xFFu16 << Ord::min(8, cpu_count) >> 8) as u8;
```
RalfJung, line 439: Please add parentheses, I can't even parse this.

What's N here? cpu_count? No I guess it must be `Ord::min(8, cpu_count)`? Maybe let-bind that first.

src/machine.rs
```diff
@@ -471,6 +499,10 @@ pub struct MiriMachine<'tcx> {
 
     /// The set of threads.
     pub(crate) threads: ThreadManager<'tcx>,
+
+    /// On which CPUs each thread is eligible to run
```
RalfJung, line 503:
```suggestion
    /// On which CPUs each thread is eligible to run.
    /// This has no effect at all, it is just tracked to produce the correct result
    /// in sched_getaffinity.
```

src/shims/unix/foreign_items.rs
```diff
@@ -267,6 +269,73 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.write_scalar(result, dest)?;
             }
 
+            "sched_getaffinity" => {
```
RalfJung, line 272: You added this in the "allocation" group of shims. That does not seem right.  "Threading" seems best.

src/shims/unix/foreign_items.rs
```diff
@@ -267,6 +269,73 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let pid = this.read_scalar(pid)?.to_u32()?;
+                let cpusetsize = this.read_target_usize(cpusetsize)?;
+                let mask = this.read_pointer(mask)?;
+
+                let thread_id = match pid {
+                    0 => this.active_thread(),
+                    _ => ThreadId::from(pid),
+                };
+
+                if cpusetsize != std::mem::size_of::<CpuAffinityMask>() as u64 {
```
RalfJung, line 292: The man page says "smaller then", so shouldn't this be `<`?
If it's actually bigger you should probably initialize the remaining bytes to zero... though the manpage doesn't really say what happens.

We generally avoid `as` casts, please use `try_into().unwrap()`.

src/shims/unix/foreign_items.rs
```diff
@@ -267,6 +269,73 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let cpusetsize = this.read_target_usize(cpusetsize)?;
+                let mask = this.read_pointer(mask)?;
+
+                let thread_id = match pid {
+                    0 => this.active_thread(),
+                    _ => ThreadId::from(pid),
+                };
+
+                if cpusetsize != std::mem::size_of::<CpuAffinityMask>() as u64  {
+                    // apparently this can happen for older kernels
```
RalfJung, line 327: But why would we implement the behavior of older kernels...?

src/shims/unix/foreign_items.rs
```diff
@@ -267,6 +269,73 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let cpusetsize = this.read_target_usize(cpusetsize)?;
+                let mask = this.read_pointer(mask)?;
+
+                let thread_id = match pid {
+                    0 => this.active_thread(),
+                    _ => ThreadId::from(pid),
+                };
+
+                if cpusetsize != std::mem::size_of::<CpuAffinityMask>() as u64 {
+                    // apparently this can happen for older kernels
```
RalfJung, line 293: The manpage seems to say that for getaffinity this is still current behavior.

tests/pass-dep/libc/libc-misc.rs
```diff
@@ -63,6 +63,60 @@ fn test_sigrt() {
     assert!(max - min >= 8)
 }
 
+#[cfg(target_os = "linux")]
```
RalfJung, line 66: Please make sure this is tested for all OSes that you added support for.

src/machine.rs
RalfJung: Reading the man page, the internal kernel structure for this might be bigger. So an alternative would be to dynamically allocate something big enough for num_cpus.

I am fine with a static size as well, but then we should limit num_cpus so that, too. So there should be a `const MAX_CPUS = 1024` and  everything is computed from that.

src/machine.rs
```diff
@@ -420,6 +420,70 @@ impl<'tcx> PrimitiveLayouts<'tcx> {
...
+
+        this
+    }
+
+    fn set(&mut self, target: &rustc_target::spec::Target, cpu: usize) {
+        // we silently ignore CPUs that are out of bounds. Linux doesn't support more than 1024
+        // CPUs apparently, so neither do we.
+        if cpu > 8 * self.0.len() {
+            return;
+        }
```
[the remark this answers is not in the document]
folkertdev: the equivalent `libc::CPU_SET` implementation will panic https://docs.rs/libc/latest/src/libc/unix/linux_like/linux/mod.rs.html#4703 with an out of bounds error. 

but this function cannot be accessed directly by users. The only case where that branch gets hit is if a user configures miri with 1024 or more CPUs.

src/machine.rs
folkertdev: well I'm basing the implementation in this PR on the logic around `cpu_set_t` in the rust libc crate. Miri should at least be compatible with that I'd think.

I'll rework the logic with that constant.

src/machine.rs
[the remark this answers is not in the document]
RalfJung: I asked what Linux does, the kernel / glibc, not the Rust crate. :)

> but this set function here cannot be accessed directly by users. 

They can call setaffinity with a larger-than-1024bit cpu set. What does that do?

src/machine.rs
RalfJung: > well I'm basing the implementation in this PR on the logic around cpu_set_t in the rust libc crate. 

Then you need some FIXMEs in the implementation, since what this *should* implement is the logic of the glibc function with that name. It seems like the libc crate only exposes a part of that.

folkertdev: well, the [man page](https://linux.die.net/man/3/cpu_alloc_size) does currently bound the cpu set at 1024 elements:

> The first available CPU on the system corresponds to a cpu value of 0, the next CPU corresponds to a cpu value of 1, and so on. The constant CPU_SETSIZE (currently 1024) specifies a value one greater than the maximum CPU number that can be stored in cpu_set_t.

is that not sufficient? I guess miri could assert on that constant being 1024 in practice?

src/machine.rs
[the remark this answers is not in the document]
folkertdev: based on a C program run with valgrind, it just ignores the extra bytes (and does not read the extra bytes that the caller says are available). There is now a test for this case

src/machine.rs
RalfJung: "the maximum CPU number that can be stored in cpu_set_t" is not the same as "the maximum number of CPUs supported by these APIs". The man page explicitly discusses this.

```
   Handling systems with large CPU affinity masks
       The underlying system calls (which represent CPU masks as bit masks of type unsigned long *) impose no restriction on the size of the CPU mask.  However, the cpu_set_t data type used by glibc has a fixed
       size of 128 bytes, meaning that the maximum CPU number that can be represented is 1023.  If the kernel CPU affinity mask is larger than 1024, then calls of the form:

           sched_getaffinity(pid, sizeof(cpu_set_t), &mask);

       fail with the error EINVAL, the error produced by the underlying system call for the case where the mask size specified in cpusetsize is smaller than the size of the affinity mask  used  by  the  kernel.
       (Depending on the system CPU topology, the kernel affinity mask can be substantially larger than the number of active CPUs in the system.)

       When working on systems with large kernel CPU affinity masks, one must dynamically allocate the mask argument (see CPU_ALLOC(3)).  Currently, the only way to do this is by probing for the size of the re‐
       quired mask using sched_getaffinity() calls with increasing mask sizes (until the call does not fail with the error EINVAL).

       Be aware that CPU_ALLOC(3) may allocate a slightly larger CPU set than requested (because CPU sets are implemented as bit masks allocated in units of sizeof(long)).  Consequently, sched_getaffinity() can
       set  bits  beyond  the requested allocation size, because the kernel sees a few additional bits.  Therefore, the caller should iterate over the bits in the returned set, counting those which are set, and
       stop upon reaching the value returned by CPU_COUNT(3) (rather than iterating over the number of bits requested to be allocated).
```

src/machine.rs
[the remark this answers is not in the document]
RalfJung: Wait, you are saying that `1` bits beyond the first 1024 bits are ignored? That contradicts the documentation. I am not sure how you are concluding that though, presumably your system has fewer than 1024 CPU cores so the bit has no effect.

It doesn't seem to return EINVAL, is the key point. So the Miri implementation should read all the bytes. If none of the cores that actually "exist" (according to num_cpus) is enabled, it should return EINVAL. The one thing I am not sure about is what happens when you do setaffinity with an all-1 mask, and then getaffinity -- do you get the all-1 mask, or do you get just the CPU cores that actually exist on your system?

src/machine.rs
```diff
@@ -420,6 +420,81 @@ impl<'tcx> PrimitiveLayouts<'tcx> {
     }
 }
 
+/// A thread's CPU affinity mask determines the set of CPUs on which it is eligible to run.
```
RalfJung, line 423: Please move this to another file, `machine.rs` is already way too big. Since it is related to threads, a new file in `src/concurrency` is probably a good choice.

src/machine.rs
```diff
@@ -420,6 +420,81 @@ impl<'tcx> PrimitiveLayouts<'tcx> {
...
+// the actual representation depends on the target's endianness and pointer width.
+// See CpuAffinityMask::set for details
+#[derive(Clone)]
+pub(crate) struct CpuAffinityMask([u8; Self::MAX_CPUS / 8]);
+
+impl CpuAffinityMask {
+    // this is compatible with the libc `CPU_SETSIZE` constant and corresponds to the number
+    // of CPUs that a `cpu_set_t` can contain.
+    const MAX_CPUS: usize = 1024;
```
RalfJung, line 432: Somewhere in machine initialization, this constant should be used to throw an error when `Zmiri-num-cpus` is set to a larger value.

src/machine.rs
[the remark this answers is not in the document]
folkertdev: you are correct, and valgrind needs a bit more help to see that

<details>

```c
#define _GNU_SOURCE
#include <stdio.h>
#include <stdlib.h>
#include <sched.h>
#include <unistd.h>
#include <errno.h>
#include <string.h>

int main(void) {
    pid_t pid = 0;
    cpu_set_t *mask = malloc(sizeof(cpu_set_t));

    // Initialize the CPU set
    CPU_ZERO(mask);

    for (int i = 0; i < 8; ++i) {
        CPU_SET(i, mask);
    }

    if (sched_setaffinity(pid, sizeof(cpu_set_t) + 64, mask) != 0) {
        perror("sched_setaffinity");
        exit(EXIT_FAILURE);
    }

    return 0;
}
```

gives this error

```
==45186== Syscall param sched_setaffinity(mask) points to unaddressable byte(s)
==45186==    at 0x4D7784B: sched_setaffinity@@GLIBC_2.3.4 (sched_setaffinity.c:35)
==45186==    by 0x40122F: main (in /home/folkertdev/c/affinity)
==45186==  Address 0x4e6a0c0 is 0 bytes after a block of size 128 alloc'd
==45186==    at 0x4A390C5: malloc (vg_replace_malloc.c:442)
==45186==    by 0x40118F: main (in /home/folkertdev/c/affinity)
```

</details>

There is now a test that verifies miri catches such invalid input.

> do you get the all-1 mask, or do you get just the CPU cores that actually exist on your system?

you get a mask with just the cores that exist on your system. That is why `sched_getaffinity` can be used to determine available parallelism.

src/machine.rs
folkertdev: there is now an assert there, but I also added validation where the miriflags get parsed

RalfJung:
Instead of repeating the same cfg over and over, please move these "use" inside the function.

folkertdev:
does 

```
//@ignore-target-windows: only very limited libc on Windows
//@ignore-target-apple: `sched_setaffinity` is not supported on macOS
//@ignore-target-wasm32
```

let through the same targets as 

```
    #[cfg(any(target_os = "linux", target_os = "freebsd", target_os = "android"))]
```

it's close but there's probably something I'm missing

RalfJung:
You don't need the wasm line, we only run very few tests on wasm as it is not fully supported by Miri.

tests/fail-dep/libc/affinity.rs
```diff
@@ -0,0 +1,17 @@
...
+    use libc::{cpu_set_t, sched_setaffinity};
+
+    use std::mem::size_of;
+
+    // If pid is zero, then the calling thread is used.
+    const PID: i32 = 0;
+
+    let mut cpuset: cpu_set_t = unsafe { core::mem::MaybeUninit::zeroed().assume_init() };
+
+    let err = unsafe { sched_setaffinity(PID, size_of::<cpu_set_t>() + 1, &mut cpuset) }; //~ ERROR: memory access failed
```
RalfJung, line 15:
```suggestion
    let err = unsafe { sched_setaffinity(PID, size_of::<cpu_set_t>() + 1, &cpuset) }; //~ ERROR: memory access failed
```

tests/pass-dep/libc/libc-affinity.rs
```diff
@@ -0,0 +1,183 @@
...
+    assert_eq!(std::io::Error::last_os_error().kind(), std::io::ErrorKind::InvalidInput);
+
+    // at the start, thread 1 should be set
+    let err = unsafe { sched_getaffinity(PID, size_of::<cpu_set_t>(), &mut cpuset) };
+    assert_eq!(err, 0);
+    assert!(unsafe { libc::CPU_ISSET(1, &cpuset) });
+
+    // make a valid mask
+    unsafe { libc::CPU_ZERO(&mut cpuset) };
+    unsafe { libc::CPU_SET(0, &mut cpuset) };
```
RalfJung, line 111: This is also a key test for endianess and pointer size shenanigans, isn't it? CPU_SET etc are in libc and so if your mapping of which bit is stored where would be wrong, this is the test that would notice.

We should in particular make sure that this checks that we are doing the chunk size and endianess correctly. I think chunk size can only make a difference on a big-endian system? On little-endian, the first byte is always going to store the lowest 8 CPUs, no matter the chunk size. We run CI on a 64bit big-endian target as well though.

So, we're probably good on the testing side, but it is confusing to have this in a test called `lying_about_size`.

RalfJung (commented):
Thanks!

This is such an annoying API to implement and for so little benefit... I am beginning to regret going down this rabbit hole. :/ But I guess now we should finish what we started. ;)

src/concurrency/cpu_affinity.rs
```diff
@@ -0,0 +1,84 @@
...
+    fn set(&mut self, target: &rustc_target::spec::Target, cpu: usize) {
+        // we silently ignore CPUs that are out of bounds. This matches the behavior of
+        // `sched_setaffinity` with a mask that specifies more than `CPU_SETSIZE` CPUs.
+        if cpu >= MAX_CPUS {
+            return;
+        }
+
+        // the actual representation of the bytes array is either
+        //
+        // - [u32; 32] on 32-bit platforms except x86_64
```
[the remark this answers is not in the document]
folkertdev: yeah, that I what I thought too, but this is what rust libc seems to do?

https://docs.rs/libc/latest/src/libc/unix/linux_like/linux/mod.rs.html#150-157

```rust
    pub struct cpu_set_t {
        #[cfg(all(target_pointer_width = "32",
                  not(target_arch = "x86_64")))]
        bits: [u32; 32],
        #[cfg(not(all(target_pointer_width = "32",
                      not(target_arch = "x86_64"))))]
        bits: [u64; 16],
    }
```

could this have something to do with 64-bit systems that can still run 32-bit binaries?

src/shims/unix/foreign_items.rs
```diff
@@ -553,6 +555,94 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let thread_id = match pid {
+                    0 => this.active_thread(),
+                    _ => ThreadId::from(pid),
+                };
+
+                if this.ptr_is_null(mask)? {
+                    let einval = this.eval_libc("EFAULT");
+                    this.set_last_error(einval)?;
+                    this.write_scalar(Scalar::from_i32(-1), dest)?;
+                } else if cpusetsize < std::mem::size_of::<CpuAffinityMask>().try_into().unwrap() {
```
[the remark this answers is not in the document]
folkertdev: I have checked this, and this is documented behavior for `sched_getaffinity`. For `sched_setaffinity` it only looks at the number of bytes you provide, but `sched_getaffinity` will error if you provide it less that `sizeof(cpuset_t)` bytes.

src/bin/miri.rs
```diff
@@ -593,6 +593,9 @@ fn main() {
             let num_cpus = param
                 .parse::<u32>()
                 .unwrap_or_else(|err| show_error!("-Zmiri-num-cpus requires a `u32`: {}", err));
+            if !(1..=miri::MAX_CPUS).contains(&usize::try_from(num_cpus).unwrap()) {
```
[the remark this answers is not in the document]
folkertdev: but `num_cpus` is a u32 anyway? and `miri::MAX_CPUS` is just 1024 right now, so I don't think there is an actual risk of this line causing issues

tests/pass-dep/libc/libc-affinity.rs
folkertdev: yes, invalid endianness logic will trigger errors in the test suite on CI.

src/concurrency/cpu_affinity.rs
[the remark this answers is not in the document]
RalfJung: Maybe it's about the x32 target. I'd say ignore this for now, just use the ptr width.

src/bin/miri.rs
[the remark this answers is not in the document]
RalfJung: Ah, I had missed that num_cpus is a u32.

tests/pass-dep/libc/libc-affinity.rs
folkertdev: I renamed the test

folkertdev (commented):
@rustbot ready

src/concurrency/cpu_affinity.rs
```diff
@@ -13,14 +13,15 @@ pub const MAX_CPUS: usize = 1024;
...
 // See CpuAffinityMask::set for details
 #[derive(Clone)]
-pub(crate) struct CpuAffinityMask([u8; MAX_CPUS / 8]);
+pub(crate) struct CpuAffinityMask([u8; Self::CPU_MASK_BYTES]);
 
 impl CpuAffinityMask {
+    const CPU_MASK_BYTES: usize = MAX_CPUS / 8;
     // code depends on the exact size of this type
-    const _SIZE_ASSERT: () = assert!(std::mem::size_of::<Self>() == 128);
+    const _SIZE_ASSERT: () = assert!(std::mem::size_of::<Self>() == Self::CPU_MASK_BYTES);
```
RalfJung, line 21: I don't get the point of this assertion, it is literally defined as an array of CPU_MASK_BYTES bytes... and we're not transmuting `CpuAffinityMask` around so its underlying representation does not even matter.

folkertdev: we're not transmuting but the logic does use `size_of::<CpuAffinityMask>` so changing its size (by adding extra fields) would break logic in perhaps non-obvious ways. That is why this asserted agains the value 128 initially: that is just the number of bytes that `cpu_size_t` is supposed to be according to libc.

I think the better approach now is to expose the `CPU_MASK_BYTES`, and use that in the logic instead of the size of the type.

RalfJung: > I think the better approach now is to expose the CPU_MASK_BYTES, and use that in the logic instead of the size of the type.

:+1:

src/shims/unix/foreign_items.rs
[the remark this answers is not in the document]
RalfJung: The only docs I found is "cpusetsize is smaller than the size of the affinity mask used by the kernel". It is unclear whether "the size of the affinity mask used by the kernel" refers to the actual mask used on this system or to cpuset_t.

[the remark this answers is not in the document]
folkertdev: It's imperfect, but 

https://android.googlesource.com/kernel/mediatek/+/android-mediatek-sprout-3.4-kitkat-mr2/include/linux/cpumask.h

says that `cpumask_t` is just a bitset with `NR_CPUS` bits (rounded up to a mulitple of `sizeof(unsigned long)`. Given that we now explicitly define miri's `NR_CPUS` to be 1024, I think that means it is correct to return an error if a buffer smaller than `1024 / 8 = 128` bytes is provided.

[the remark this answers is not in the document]
RalfJung: As I already said yesterday, I think the easiest way forward is to just do an experiment. What happens when you call `sched_getaffinity` on a cpuset that is "just" 512 bits large?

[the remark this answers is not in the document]
folkertdev: wild. 

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9b6f5e563113cc5166c7658767d91457

```rust
use libc::{cpu_set_t, sched_getaffinity};

const PID: i32 = 0;

fn main() {
    let mut cpuset: cpu_set_t = unsafe { core::mem::MaybeUninit::zeroed().assume_init() };

    dbg!(core::mem::size_of::<libc::c_long>());

    for i in (0..=1024 / 8).step_by(8) {
        for j in 1..8 {
            let err = unsafe { sched_getaffinity(PID, i + j, &mut cpuset) };
            assert_eq!(err, -1, "fail for {}", i + j);
            assert_eq!(
                std::io::Error::last_os_error().kind(),
                std::io::ErrorKind::InvalidInput
            );
        }
    }

    for i in (8..=1024 / 8).step_by(8) {
        dbg!(i);
        let err = unsafe { sched_getaffinity(PID, i, &mut cpuset) };
        assert_eq!(err, 0, "fail for {i}");
    }
}
```

this program succeeds on x86_64 linux. So multiples of 8 (besides 0) return a result, but anything other than that does not. I suspect that this is related to the size of the `unsigned long` type used to represent the underlying storage, but I don't have a system where that is a value besides 8 (and implements these APIs).

I also don't really see how this relates to the man page text. But anyway, this is the behavior in practice so I guess we just mirror that?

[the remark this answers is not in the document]
RalfJung: Yeah, seems like the mask has to be a multiple of the "chunk size" then. That actually kind of makes sense, doesn't it?

src/concurrency/cpu_affinity.rs
```diff
@@ -29,21 +29,32 @@ impl CpuAffinityMask {
         this
     }
 
+    pub fn chunk_size(target: &rustc_target::spec::Target) -> usize {
+        // The actual representation of the CpuAffinityMask is [c_ulong; _], in practice either
+        //
+        // - [u32; 32] on 32-bit platforms
+        // - [u64; 16] everywhere else
+        match target.pointer_width {
```
RalfJung, line 37: Isn't this just `pointer_width.bytes()`?

Then the natural return type becomes `u64` which will also work better in sched_getaffinity.

src/concurrency/cpu_affinity.rs
```diff
@@ -0,0 +1,94 @@
...
+            4 => {
+                let start = cpu / 32 * 4; // first byte of the correct u32
+                let chunk = self.0[start..].first_chunk_mut::<4>().unwrap();
+                let offset = cpu % 32;
+                *chunk = match target.options.endian {
+                    Endian::Little => (u32::from_le_bytes(*chunk) | 1 << offset).to_le_bytes(),
+                    Endian::Big => (u32::from_be_bytes(*chunk) | 1 << offset).to_be_bytes(),
+                };
+            }
+            _ => {
```
RalfJung, line 66:
```suggestion
            8 => {
```
and then `bug!` in the fallback arm

src/shims/unix/foreign_items.rs
```diff
@@ -575,18 +575,26 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    // we only copy whole chunks of size_of::<c_ulong>()
                     let einval = this.eval_libc("EINVAL");
                     this.set_last_error(einval)?;
                     this.write_scalar(Scalar::from_i32(-1), dest)?;
                 } else if let Some(cpuset) = this.machine.thread_cpu_affinity.get(&thread_id) {
                     let cpuset = cpuset.clone();
-                    this.write_bytes_ptr(mask, cpuset.as_slice().iter().copied())?;
+                    // we only copy whole chunks of size_of::<c_ulong>()
+                    let byte_count = chunk_size * (cpuset.as_slice().len() / chunk_size);
+                    this.write_bytes_ptr(mask, cpuset.as_slice()[..byte_count].iter().copied())?;
```
RalfJung, line 597: We've already checked that the size is a multiple of chunk_size... so isn't this equivalent to passing the entire `cpuset`?

tests/pass-dep/libc/libc-affinity.rs
```diff
@@ -89,18 +89,37 @@ fn large_set() {
...
+    let mut cpuset: cpu_set_t = unsafe { core::mem::MaybeUninit::zeroed().assume_init() };
+
+    // should be 4 on 32-bit systems and 8 otherwise for systems that implement sched_getaffinity
+    let step = size_of::<std::ffi::c_ulong>();
+
+    // currently the maximum number of CPUs that miri supports, and also the number that a
+    // libc::cpu_set_t can hold.
+    let max_cpus = 1024;
+
+    for i in (0..=max_cpus / step).step_by(step) {
```
RalfJung, line 102: I don't think we have to exhaustively test this all the way up to 1024. `for i in `(0..=2).map(|x| x*step)` should be enough.

tests/pass-dep/libc/libc-affinity.rs
```diff
@@ -89,18 +89,37 @@ fn large_set() {
...
+        // a whole multiple of the size of c_ulong works
+        let err = unsafe { sched_getaffinity(PID, i, &mut cpuset) };
+        assert_eq!(err, 0, "fail for {i}");
```
RalfJung, line 105: Shouldn't it fail for 0?

src/shims/unix/foreign_items.rs
```diff
@@ -585,15 +585,21 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                     let cpuset = cpuset.clone();
                     // we only copy whole chunks of size_of::<c_ulong>()
-                    let byte_count = chunk_size * (cpuset.as_slice().len() / chunk_size);
+                    let byte_count = cpuset
+                        .as_slice()
+                        .len()
+                        .checked_div(chunk_size)
+                        .unwrap()
+                        .checked_mul(chunk_size)
+                        .unwrap();
```
RalfJung, line 602: `strict_mul`/`strict_div` can be used to avoid the `unwrap`

src/concurrency/cpu_affinity.rs
RalfJung: Oh sorry, it's not a `Size`.

`pointer_width * 8` then.

Please also add a FIXME to look up the actual size of `core::ffi::c_ulong` in the future.

src/concurrency/cpu_affinity.rs
folkertdev: That's useful, because it fails for the `avr.json` case, which appears to be 16-bit system which we haven't considered yet. We could either

- make the logic work on that system (so we simulate a `[u16; 64]` array)
- add a `2 => { /* do nothing */ }`  branch, because the `sched_{g, s}etaffinity` APIs will not be available on 16-bit systems anyway

do you have a preference?

RalfJung: The avr test isn't even running Unix. So I think we should avoid doing anything with this CpuSet map on non-Unix systems.

src/machine.rs
```diff
@@ -627,6 +634,18 @@ impl<'tcx> MiriMachine<'tcx> {
...
+            "miri only supports up to {} CPUs, but {} were configured",
+            cpu_affinity::MAX_CPUS,
+            config.num_cpus
+        );
+        let threads = ThreadManager::default();
+        let mut thread_cpu_affinity = FxHashMap::default();
+        thread_cpu_affinity.insert(
+            threads.active_thread(),
+            CpuAffinityMask::new(&tcx.sess.target, config.num_cpus),
+        );
```
RalfJung, line 648: I was actually thinking this should be skipped on non-Unix.

src/shims/unix/foreign_items.rs
```diff
@@ -553,6 +555,103 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+                let [pid, cpusetsize, mask] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let pid = this.read_scalar(pid)?.to_u32()?;
+                let cpusetsize = this.read_target_usize(cpusetsize)?;
+                let mask = this.read_pointer(mask)?;
+
+                let thread_id = match pid {
+                    0 => this.active_thread(),
+                    _ => ThreadId::from(pid),
```
RalfJung, line 575: I think this is the wrong ID... `ThreadId` corresponds to the pthtread ID, not the kernel PID/TID. Miri currently does not have a notion of PID/TID for threads (see https://github.com/rust-lang/miri/issues/3730).

Would it suffice to only support pid 0 for now?

src/shims/unix/foreign_items.rs
```diff
@@ -553,6 +555,103 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                } else if !this.machine.thread_cpu_affinity.contains_key(&thread_id) {
+                    // The thread whose ID is pid could not be found
+                    let einval = this.eval_libc("ESRCH");
+                    this.set_last_error(einval)?;
+                    this.write_scalar(Scalar::from_i32(-1), dest)?;
+                } else {
+                    // NOTE: cpusetsize might be smaller than `CpuAffinityMask::CPU_MASK_BYTES`
+                    let bits_slice = this.read_bytes_ptr_strip_provenance(mask, Size::from_bytes(cpusetsize))?;
+                    // This ignores the bytes beyond `CpuAffinityMask::CPU_MASK_BYTES`
+                    let bits_array = std::array::from_fn(|i| bits_slice.get(i).copied().unwrap_or(0));
```
RalfJung, line 640: Please add a type annotation here so that one can locally see what kind of array is created here -- right now the array length isn't even set anywhere in this module.

src/shims/unix/foreign_items.rs
saethlin: On MacOS, `man getpid` documents

>      getpid() returns the process ID of the calling process.  The ID is
>      guaranteed to be unique and is useful for constructing temporary file
>      names.

So I don't think always returning 0 would be a valid implementation.

saethlin: (not that I have any particular advice on what else to do)

RalfJung: We're not returning a PID here, we are taking a PID as argument, and 0 is specified to mean "current thread". So I think we can declare that we don't support any other value.

saethlin: Ah! Yes.

folkertdev: > Would it suffice to only support pid 0 for now? The case of setting/getting another thread's mask anyway seems to be untested.

Yes, that covers all actual rust examples I've seen so far, and it's untested exactly because `getttid` is not implemented currently.

src/shims/unix/foreign_items.rs
```diff
@@ -590,6 +590,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     let einval = this.eval_libc("EINVAL");
                     this.set_last_error(einval)?;
                     this.write_scalar(Scalar::from_i32(-1), dest)?;
+                } else if pid != 0 {
+                    // TODO: when https://github.com/rust-lang/miri/issues/3730 is fixed this should use its notion of tid/pid
+
+                    // The thread whose ID is pid could not be found
+                    let einval = this.eval_libc("ESRCH");
+                    this.set_last_error(einval)?;
+                    this.write_scalar(Scalar::from_i32(-1), dest)?;
```
RalfJung, line 599: This looks like the program did something wrong, which is not the case. Also you still have the `ThreadId::from(pid)` above which is a conversion that makes no sense.

Instead of this, please make the `_` arm in the `match pid` above do something like
```rust
throw_unsup_format!("`sched_getaffinity` is only supported with a pid of 0 (indicating the current thread)");
```

RalfJung:
That was the last of my comments, thanks and congrats. :)

Please squash it all into one or a few commits (ideally with a force-push that has no diff)

folkertdev:
@rustbot ready

yeah lol this "good first issue" escalated quickly

RalfJung:
That force-push has quite a big diff unfortunately... you rebased over the latest master instead of keeping the same base commit, didn't you? That makes reviewing a lot harder. :/

folkertdev:
ah you meant no diff versus the branch (and not versus the master branch). I can probably revert it back to 467f5e1eb7db71667589ff1f2eedf63163d3a19d again if that is easier

RalfJung:
Well there's always a diff to the master branch, it's all your changes. :)
I guess I wasn't clear enough.

RalfJung:
I made it lookup the c_ulong type from `core` to avoid having to hard-code that.

However, in doing that I noticed something odd -- `sched_setaffinity` still allows sizes that are not a multiple of the chunk size. That can't be right, can it?

folkertdev:
For `sched_setaffinity` an arbitrary size does work, except that the current implementation does not handle the `cpusetsize == 0` case

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b861ccd6ea9d75a3b1146c38f2ff2926

this runs  succesfully

```rust
use libc::{cpu_set_t, sched_setaffinity};

const PID: i32 = 0;

fn main() {
    let mut cpuset: cpu_set_t = unsafe { core::mem::MaybeUninit::zeroed().assume_init() };

    unsafe { libc::CPU_SET(1, &mut cpuset) };

    let err = unsafe { sched_setaffinity(PID, 0, &cpuset) };
    assert_eq!(err, -1, "fail for {}", 0);
    assert_eq!(
        std::io::Error::last_os_error().kind(),
        std::io::ErrorKind::InvalidInput
    );

    for i in 1..1024 {
        let err = unsafe { sched_setaffinity(PID, i, &cpuset) };
        assert_eq!(err, 0, "fail for {i}");
    }
}
```

I'll fix that

RalfJung:
Interesting. And the semantics is that it works just as if all the remaining bytes are zero?

src/shims/unix/foreign_items.rs
```diff
@@ -571,6 +573,97 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                };
+
+                if this.ptr_is_null(mask)? {
+                    let einval = this.eval_libc("EFAULT");
+                    this.set_last_error(einval)?;
+                    this.write_scalar(Scalar::from_i32(-1), dest)?;
+                } else {
+                    // NOTE: cpusetsize might be smaller than `CpuAffinityMask::CPU_MASK_BYTES`
+                    let bits_slice = this.read_bytes_ptr_strip_provenance(mask, Size::from_bytes(cpusetsize))?;
+                    // This ignores the bytes beyond `CpuAffinityMask::CPU_MASK_BYTES`
```
RalfJung, line 650:
```suggestion
                    // This ignores the bytes beyond `CpuAffinityMask::CPU_MASK_BYTES`, and fills
                    // short arrays with 0s until they have the right length.
```

folkertdev:
that is what I assume but I can't really test that because I only have 8 physical cores on my system. the man pages say this 

```
       sched_setaffinity()  sets the CPU affinity mask of the thread whose ID is pid to the value specified by mask.  If pid is zero, then the
       calling thread is used.  The argument cpusetsize is the length (in bytes) of the data pointed to by mask.  Normally this argument would
       be specified as sizeof(cpu_set_t).
```

so the bytes thing is documented. What happens to any unspecified bytes isn't but assuming zeros seems like the logical choice.

RalfJung:
Makes sense. Maybe leave a comment in the source saying that this part is guesswork.

src/shims/unix/foreign_items.rs
```diff
@@ -645,7 +645,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     this.set_last_error(einval)?;
                     this.write_scalar(Scalar::from_i32(-1), dest)?;
                 } else {
-                    // NOTE: cpusetsize might be smaller than `CpuAffinityMask::CPU_MASK_BYTES`
+                    // NOTE: cpusetsize might be smaller than `CpuAffinityMask::CPU_MASK_BYTES`.
+                    // Any unspecified bytes are treated as zero here (none of the CPUs are configured).
+                    // This is not exactly documented, so we assume that this is the behavior in practice
```
RalfJung, line 650:
```suggestion
                    // This is not exactly documented, so we assume that this is the behavior in practice.
```

RalfJung:
> yeah lol this "good first issue" escalated quickly

It did.^^ It's not always easy to predict how much research will be required to implement some API...
Thanks for seeing this through!

@bors r+

folkertdev:
that test failure is legit, in that on a big-endian system like `s390x-unknown-linux-gnu` giving it one byte of the cpu mask codes for CPUs 56..64 which are not available on the system... so in practice you'd likely want to work in multiples of the chunk size anyway, at least on BE systems.

So if we make the loop bound here start at 8 on BE systems things work as expected again

```rust
    // any other number of bytes (at least up to `size_of<cpu_set_t>()` will work
    for i in 1..24 {
        let err = unsafe { sched_setaffinity(PID, i, &cpuset) };
        assert_eq!(err, 0, "fail for {i}");
    }
```

tests/pass-dep/libc/libc-affinity.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    // setting a mask of size 0 is invalid
+    let err = unsafe { sched_setaffinity(PID, 0, &cpuset) };
+    assert_eq!(err, -1);
+    assert_eq!(std::io::Error::last_os_error().kind(), std::io::ErrorKind::InvalidInput);
+
+    // any other number of bytes (at least up to `size_of<cpu_set_t>()` will work
+    for i in 1..24 {
+        let err = unsafe { sched_setaffinity(PID, i, &cpuset) };
+        assert_eq!(err, 0, "fail for {i}");
```
RalfJung, line 122: We're getting
```
   thread 'main' panicked at tests/pass-dep/libc/libc-affinity.rs:122:9:
  assertion `left == right` failed: fail for 1
    left: -1
   right: 0
```
on s390x-unknown-linux-gnu (big-endian)

RalfJung: which makes sense, the first 7 bytes are 0 in the mask for that target, so this is setting an all-0 mask

RalfJung:
> So if we make the loop bound here start at 8 on BE systems things work as expected again

Yeah, something like that.

PR #3700, opened 2024-06-22
./miri: nicer error when building miri-script fails

RalfJung:


RalfJung:
@RossSmyth @ChrisDenton if you could do the same for `miri.bat` that would be great :)

PR #3701, opened 2024-06-22
show warning when Stacked Borrows skips a reborrow due to 'extern type'

RalfJung:
When this happens, we can't actually be sure to catch all bugs -- LLVM will still get a `noalias` but Miri can't do reborrowing. That's not good.

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -130,6 +130,7 @@ pub enum NonHaltingDiagnostic {
     WeakMemoryOutdatedLoad {
         ptr: Pointer,
     },
+    ExternTypeReborrow,
 }
 
 /// Level of Miri specific diagnostics
@@ -593,6 +594,8 @@ impl<'tcx> MiriMachine<'tcx> {
             RejectedIsolatedOp(_) =>
                 ("operation rejected by isolation".to_string(), DiagLevel::Warning),
             Int2Ptr { .. } => ("integer-to-pointer cast".to_string(), DiagLevel::Warning),
+            ExternTypeReborrow =>
+                ("reborrow of reference to `extern type`".to_string(), DiagLevel::Warning),
             CreatedPointerTag(..)
             | PoppedPointerTag(..)
             | CreatedCallId(..)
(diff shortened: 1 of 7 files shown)


tests/pass/intrinsics/intrinsics.stderr
RalfJung, line 1: This one is problematic, I need to see whether we can avoid this warning.

RalfJung:
With https://github.com/rust-lang/rust/pull/127859, this doesn't show any warnings for standard library code any more. :)

src/diagnostics.rs
```diff
@@ -647,34 +652,50 @@ impl<'tcx> MiriMachine<'tcx> {
...
+                            "alternatively, `MIRIFLAGS=-Zmiri-permissive-provenance` disables this warning"
                         ),
                     ),
                 ],
+            ExternTypeReborrow => {
+                vec![
+                    (
+                        None,
+                        format!(
+                            "`extern type` are not compatible with Miri's Stacked Borrows aliasing checks; Miri may miss bugs in this code"
```
saethlin, line 688: I would rather not just call these "checks" because I think it gives the wrong impression. Our aliasing model checking isn't a bunch of separate checks like Ubsan, it's a runtime model. The difference between just saying "checks" and something like "model checking" is more clear in the line below that I'm not commenting on. But I don't want to obsess over the exact wording, just having this warning at all is very nice.
```suggestion
                            "`extern type` are not compatible with Miri's Stacked Borrows aliasing model checking; Miri may miss bugs in this code"
```

saethlin:
I think you want to update the PR description a bit. I support having a warning for this, and won't complain much if you prefer your current wording :p

src/diagnostics.rs
RalfJung: We could also say something like "with the Stacked Borrows aliasing model implemented by Miri"?

saethlin: That wording sounds good to me

RalfJung:
All right, let's go with this then. :)

@bors r=saethlin

Issue #3460, opened 2024-04-09
cargo-miri: incorrect behavior when `--target` is given multiple times


RalfJung:
When `--target` is given multiple times, Miri thinks we are building the first target, but actually cargo will build the last target.

bjorn3:
If you specify `--target` multiple times, cargo will actually build for every specifies target at the same time, except when doing `cargo run` where passing multiple `--target` errors out.

RalfJung:
Ah true, `cargo test` actually runs the tests multiple times.

So yeah Miri should support that as well then.

PR #3702, opened 2024-06-22
Run tests for all specified targets

Mandragorian:
Currently cargo-miri uses the first target specified in the command line. However, when multiple targets are specified in a `cargo build` invocation, cargo will build for all of them.

Miri should match this behaviour to reduce surprises.

Fixes: #3460

### DIFF
--- a/cargo-miri/src/phases.rs
+++ b/cargo-miri/src/phases.rs
@@ -104,9 +104,17 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
             miri_for_host()
         )
     });
-    let host = &rustc_version.host;
-    let target = get_arg_flag_value("--target");
-    let target = target.as_ref().unwrap_or(host);
+    let mut targets = get_arg_flag_values("--target").collect::<Vec<_>>();
+    // If `targets` is empty, we need to add a `--target $HOST` flag ourselves, and also ensure
+    // that the host target is indeed setup.
+    let target_flag = if targets.is_empty() {
+        let host = &rustc_version.host;
+        targets.push(host.clone());
+        Some(host)
+    } else {
+        // We don't need to add a `--target` flag, we just forward the user's flags.
+        None
+    };
 
     // If cleaning the target directory & sysroot cache,
     // delete them then exit. There is no reason to setup a new
@@ -118,8 +126,11 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
         return;
     }
 
-    // We always setup.
-    let miri_sysroot = setup(&subcommand, target, &rustc_version, verbose, quiet);
+    for target in &targets {
+        // We always setup.
+        setup(&subcommand, target.as_str(), &rustc_version, verbose, quiet);
+    }
+    let miri_sysroot = get_sysroot_dir();
 
     // Invoke actual cargo for the job, but with different flags.
     // We re-use `cargo test` and `cargo run`, which makes target and binary handling very easy but
@@ -155,10 +166,9 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
     // This is needed to make the `target.runner` settings do something,
     // and it later helps us detect which crates are proc-macro/build-script
     // (host crates) and which crates are needed for the program itself.
-    if get_arg_flag_value("--target").is_none() {
-        // No target given. Explicitly pick the host.
+    if let Some(target_flag) = target_flag {
         cmd.arg("--target");
-        cmd.arg(host);
+        cmd.arg(target_flag);
     }
 
     // Set ourselves as runner for al binaries invoked by cargo.
--- a/test-cargo-miri/run-test.py
+++ b/test-cargo-miri/run-test.py
@@ -22,12 +22,17 @@ def fail(msg):
     print("\nTEST FAIL: {}".format(msg))
     sys.exit(1)
 
-def cargo_miri(cmd, quiet = True):
+def cargo_miri(cmd, quiet = True, targets = None):
     args = ["cargo", "miri", cmd] + CARGO_EXTRA_FLAGS
     if quiet:
         args += ["-q"]
-    if ARGS.target:
+
+    if targets is not None:
+        for target in targets:
+            args.extend(("--target", target))
+    elif ARGS.target is not None:
         args += ["--target", ARGS.target]
+
     return args
 
 def normalize_stdout(str):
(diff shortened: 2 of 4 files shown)


cargo-miri/src/phases.rs
```diff
@@ -105,149 +105,148 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
...
+        // interpreting things as flags when it sees the first positional argument.
+
+        // Make sure the build target is explicitly set.
+        // This is needed to make the `target.runner` settings do something,
+        // and it later helps us detect which crates are proc-macro/build-script
+        // (host crates) and which crates are needed for the program itself.
         cmd.arg("--target");
-        cmd.arg(host);
-    }
+        cmd.arg(target);
```
RalfJung, line 160: This does not seem right -- won't this pass `--target` twice if there's also a target set in `args`?

CI passes though so maybe cargo just ignores duplicate targets?

RalfJung (commented):
Thanks for the PR!

Given that cargo supports multiple tagets itself, why do we even need to have a loop ourselves? I mean, we need a loop for `setup`,  but then IMO we should just do one big cargo invocation with all the `--target`, and let cargo do the looping.

Also, please add a test.

saethlin:
> I mean, we need a loop for `setup`

I think we actually need support in rustc-build-sysroot for passing through the full list of targets to Cargo, not a loop.

RalfJung:
Why should calling `setup` in a loop not work?

saethlin:
Ah, bad phrasing. It will work, but we don't _need_ to use a loop. It would be much faster to let Cargo run all the sysroot builds in parallel.

RalfJung:
Ah, that's what you mean. Yeah makes sense. It also complicates the copying and printing logic, though.

We can leave that as a future possibility though, it doesn't have to happen in this PR.

Mandragorian:
Looking into the code more, I see that if I don't loop I get an error that `error: only one `--target` argument is supported`, which I think originates from `cargo` and not `miri`. 

I also realised that the loop as I had set it up wouldn't work since `exec(cmd)` never returns. Multiple targets are only supported for the `build` subcommand. So I think that if we want to support running/testing for all targets specified in miri, we might need to spawn the command in a loop afte rall.

RalfJung:
> Multiple targets are only supported for the build subcommand. 

Wait, really? We intend to mirror `cargo test`. If `cargo test` does not support multiple targets then I don't think we should, either.

Mandragorian:
ah my bad. I actually run `cargo miri run` instead of `test`. sorry for the confusion.

cargo-miri/src/phases.rs
```diff
@@ -105,149 +105,148 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
...
+            if arg == "--many-seeds" {
+                many_seeds = Some(DEFAULT_MANY_SEEDS.to_owned());
+            } else if let Some(val) = arg.strip_prefix("--many-seeds=") {
+                many_seeds = Some(val.to_owned());
+            } else {
+                cmd.arg(arg);
+            }
+        }
+        // Forward all further arguments after `--` (not consumed by `ArgSplitFlagValue`) to cargo.
+        cmd.args(args);
```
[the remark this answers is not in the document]
Mandragorian: Since this is no longer in the loop the first issue should be addressed.

I added a test that passes 2 targets to miri and expects to see 2 executions of the tests. Is this what you had in mind?

I find the way I did it not so elegant. I am not sure if there is a better way.

If the `cargo_miri` method in the test suite didn't always forward the `--target` flag from its command line to miri, the code might look a bit less complex. But I am not sure if this change is wanted, or if it should be included in this PR.

cargo-miri/src/phases.rs
Mandragorian: I changed the code to only pass `--target` if it was not specified at all.

cargo-miri/src/phases.rs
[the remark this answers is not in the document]
RalfJung: We must always set `--target`, things will break horribly if we invoke cargo without `--target`.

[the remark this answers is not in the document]
Mandragorian: Sorry, let me rephrase. We should set `--target`, yes. I wanted to say that the `cargo_miri` function should not always pass the value of `ARGS.target` to pass the `--target` flag. Maybe it would be simpler if `cargo_miri` accepted the targets as an argument.

Then we could specify exactly which target(s) we want, without having to do all the manipulation I do to ensure that exactly two targets are passed to miri in the tests.

Perhaps we can add a `targets` argument with a default None value. If the value is none, only then we try to use `ARGS.target`, otherwise we use the explicitly provided argument.

cargo-miri/src/phases.rs
```diff
@@ -104,9 +104,14 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
             miri_for_host()
         )
     });
-    let host = &rustc_version.host;
-    let target = get_arg_flag_value("--target");
-    let target = target.as_ref().unwrap_or(host);
+    let mut targets = get_arg_flag_values("--target").collect::<Vec<_>>();
+    let explicit_target = if targets.is_empty() {
```
RalfJung, line 108:
```suggestion
    // If `targets` is empty, we need to add a `--target $HOST` flag ourselves, and also ensure
    // that the host target is indeed setup.
    let target_flag = if targets.is_empty() {
```

cargo-miri/src/phases.rs
```diff
@@ -104,9 +104,14 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
...
+        let host = &rustc_version.host;
+        targets.push(host.clone());
+        Some(host)
+    } else {
+        None
```
RalfJung, line 113:
```suggestion
        // We don't need to add a `--target` flag, we just forward the user's flags.
        None
```

cargo-miri/src/phases.rs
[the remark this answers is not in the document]
RalfJung: Ah, you are talking about run_test.py, not cargo-miri. That context was missing. :)

Yes, please make cargo_miri take a "targets" parameter that defaults to `None`, in which case it uses ARGS.target.

RalfJung (commented):
Thanks! Rust side looks good, just some nits.
Test still needs a bit of work.

@rustbot author

cargo-miri/src/phases.rs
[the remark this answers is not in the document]
Mandragorian: yes, i was talking about `run_test.py`.

I changed the code as discussed. Let me know if that works!

test-cargo-miri/run-test.py
```diff
@@ -180,6 +185,11 @@ def test_cargo_miri_test():
         default_ref, "test.stderr-empty.ref",
         env={'MIRIFLAGS': "-Zmiri-permissive-provenance"},
     )
+    test("`cargo miri test` (multiple targets)",
+        cargo_miri("test", targets = ["aarch64-unknown-linux-gnu", "s390x-unknown-linux-gnu"]),
+        "test.multiple_targets.stdout.ref", "test.stderr-empty.ref",
+        env={'MIRIFLAGS': "-Zmiri-permissive-provenance"},
+    )
```
RalfJung, line 192: This will make `./run-test.py` take quite a bit longer since it needs to build 2 more sysroots... and also on CI we run this many times with different `--target` but this test here is going to always be the same.

Maybe we should have another flag, like `--multi-target`, and only run the test when the flag is set? Then we can set it on CI for the first `run-test.py` invocation and avoid repeating it later.

Mandragorian: I added a `--multi-target` flag and made it so that the CI only uses it when testing the host target.

My shell skills are very bad so hopefully the code looks ok, but let me know.

RalfJung:
This looks great, thanks. :)

Can you please squash this into a single commit? (ideally while keeping the same base commit so that the force-push has an empty diff)

Mandragorian:
> This looks great, thanks. :)
> 
> Can you please squash this into a single commit? (ideally while keeping the same base commit so that the force-push has an empty diff)

done!

RalfJung:
Great, thanks!

@bors r+

PR #3703, opened 2024-06-23
nicer batch file error when building miri-script fails

RossSmyth:
https://github.com/rust-lang/miri/pull/3700#issuecomment-2184026431


RalfJung:
Thanks :)
@bors r+

PR #3707, opened 2024-06-24
Add syscall `dup()` for unix target

### LINKED ISSUE #3454: `dup()` not supported
We support duplicating file descriptors via `fcntl(F_DUPFD)`, but nut via `dup`. That's a bit silly. ;)
And we should probably support `dup2`, too.

(It also shows up in a standard library test that directly calls `dup`.)

adwinwhite:
Add support for `dup()` and `dup2()`.

Fixes #3454

### DIFF
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -273,6 +273,34 @@ impl FdTable {
 
 impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    fn dup(&mut self, old_fd: i32) -> InterpResult<'tcx, i32> {
+        let this = self.eval_context_mut();
+
+        match this.machine.fds.dup(old_fd) {
+            Some(dup_fd) => Ok(this.machine.fds.insert_fd_with_min_fd(dup_fd, 0)),
+            None => this.fd_not_found(),
+        }
+    }
+
+    fn dup2(&mut self, old_fd: i32, new_fd: i32) -> InterpResult<'tcx, i32> {
+        let this = self.eval_context_mut();
+
+        match this.machine.fds.dup(old_fd) {
+            Some(dup_fd) => {
+                if new_fd != old_fd {
+                    // Close new_fd if it is previously opened.
+                    // If old_fd and new_fd point to the same description, then `dup_fd` ensures we keep the underlying file description alive.
+                    if let Some(file_descriptor) = this.machine.fds.fds.insert(new_fd, dup_fd) {
+                        // Ignore close error (not interpreter's) according to dup2() doc.
+                        file_descriptor.close(this.machine.communicate())?.ok();
+                    }
+                }
+                Ok(new_fd)
+            }
+            None => this.fd_not_found(),
+        }
+    }
+
     fn fcntl(&mut self, args: &[OpTy<'tcx>]) -> InterpResult<'tcx, i32> {
         let this = self.eval_context_mut();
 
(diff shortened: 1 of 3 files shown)


src/shims/unix/fd.rs
```diff
@@ -273,6 +273,35 @@ impl FdTable {
...
+        let this = self.eval_context_mut();
+
+        match this.machine.fds.dup(old_fd) {
+            Some(dup_fd) => {
+                if new_fd != old_fd {
+                    // Close new_fs if it is previously opened.
+                    if let Some(file_descriptor) = this.machine.fds.remove(new_fd) {
+                        // Ignore close error according to dup2() doc.
+                        file_descriptor.close(this.machine.communicate()).ok();
+                    }
```
tiif, line 295: I wonder what would happen if ``old_fd`` and ``new_fd`` point to the same file description. If ``new_fd`` is close, would ``old_fd`` point to a destructed file description? 

It seems to be a valid operation to do this:
```c
#include <sys/epoll.h>
#include <stdio.h>
#include <fcntl.h>
#include <unistd.h>
#include <sys/socket.h>
#include <stdlib.h>

int main(int argc, char *argv[]) {

  // Just an example of file description 
  int sv[2];   
  if (socketpair(AF_UNIX, SOCK_STREAM | SOCK_NONBLOCK, 0, sv) == -1) {
    perror("socketpair");
    exit(EXIT_FAILURE);
  }

  // fd2 point to sv[0] file description too. 
  int fd2 = dup(sv[0]);
  printf("fd2 is: %d\n", fd2);
  // call dup2 again.  
  int res = dup2(sv[0], fd2);
  // It is expected to be successful.
  printf("result is: %d \n", res);
}
```

adwinwhite: If `new_fd` and `old_fd` point to the same open file description, then `close()` one of them should merely decrease the ref count by 1.
Miri does implement [this](https://github.com/rust-lang/miri/blob/3f392aa36197496427c6dd933a6a212ded627106/src/shims/unix/fd.rs#L186) correctly. `Rc::into_inner()` returns `None` unless the ref count is 1.

src/shims/unix/fd.rs
```diff
@@ -273,6 +273,35 @@ impl FdTable {
...
+        }
+    }
+
+    fn dup2(&mut self, old_fd: i32, new_fd: i32) -> InterpResult<'tcx, i32> {
+        let this = self.eval_context_mut();
+
+        match this.machine.fds.dup(old_fd) {
+            Some(dup_fd) => {
+                if new_fd != old_fd {
+                    // Close new_fs if it is previously opened.
```
RalfJung, line 291:
```suggestion
                    // Close new_fs if it is previously opened.
                    // (If old == new, then `dup_fd` ensure we keep the underlying file description alive.)
```

src/shims/unix/fd.rs
```diff
@@ -273,6 +273,35 @@ impl FdTable {
...
+                    if let Some(file_descriptor) = this.machine.fds.remove(new_fd) {
+                        // Ignore close error according to dup2() doc.
+                        file_descriptor.close(this.machine.communicate()).ok();
```
RalfJung, line 294: Nono, we can never ignore this. This could ignore UB!

You can ignore the `Scalar`, but the `?` that propagates interpreter errors must be present.

src/shims/unix/fd.rs
```diff
@@ -273,6 +273,35 @@ impl FdTable {
...
+                    }
+
+                    this.machine.fds.fds.insert(new_fd, dup_fd);
```
RalfJung, line 297: `insert` returns the old value if any, so instead of calling `remove` first you can just use that.

RalfJung (commented):
Thanks for the PR!

src/shims/unix/fd.rs
adwinwhite: Totally. I didn't distinguish interpreter's error from program's error back then.

RalfJung:
Looks good, thanks a lot :)
@bors r+

PR #3709, opened 2024-06-24
clarify the warning shown when optimizations are enabled, and the status of Tree Borrows

RalfJung:


src/bin/miri.rs
```diff
@@ -98,10 +98,9 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
             }
 
             if tcx.sess.opts.optimize != OptLevel::No {
-                tcx.dcx().warn("Miri does not support optimizations. If you have enabled optimizations \
-                    by selecting a Cargo profile (such as --release) which changes other profile settings \
-                    such as whether debug assertions and overflow checks are enabled, those settings are \
-                    still applied.");
+                tcx.dcx().warn("Miri does not support optimizations: the `-O` flag is ignored. The only effect \
```
saethlin, line 101: Why is it useful to mention -O? I've only seen that flag used internally, mostly in codegen tests.

RalfJung:
We can use some other term,  but *something* is ignored and I think we should say that. I think cargo calls this "opt-level"?

saethlin:
Yeah, Cargo and rustc both call it opt-level. `-O` is `opt-level=2`, but in GCC and clang it's equivalent to `-O1` and I strongly suspect that the most used nonzero opt level in Rust is 3 because that's the default for Cargo's `--release`.

I remain surprised that this chaos hasn't caused issues for the compiler's codegen tests.

src/bin/miri.rs
RalfJung: All right, it's "opt-level" now.

saethlin (approved):
New state looks good to me

PR #3718, opened 2024-06-28
readme: tweak wording around soundness

RalfJung:
Miri *can* tell you whether your code is sound when it finds UB -- it's not sound in that case. It can give negative answers, just not positive ones.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,11 +72,13 @@ Further caveats that Miri users should be aware of:
   when `SeqCst` fences are used that are not actually permitted by the Rust memory model, and it
   cannot produce all behaviors possibly observable on real hardware.
 
-Moreover, Miri fundamentally cannot tell you whether your code is *sound*. [Soundness] is the property
-of never causing undefined behavior when invoked from arbitrary safe code, even in combination with
+Moreover, Miri fundamentally cannot ensure that your code is *sound*. [Soundness] is the property of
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,11 +72,12 @@ Further caveats that Miri users should be aware of:
...
-Moreover, Miri fundamentally cannot tell you whether your code is *sound*. [Soundness] is the property
-of never causing undefined behavior when invoked from arbitrary safe code, even in combination with
+Moreover, Miri fundamentally cannot ensure that your code is *sound*. [Soundness] is the property of
+never causing undefined behavior when invoked from arbitrary safe code, even in combination with
 other sound code. In contrast, Miri can just tell you if *a particular way of interacting with your
-code* (e.g., a test suite) causes any undefined behavior. It is up to you to ensure sufficient
-coverage.
+code* (e.g., a test suite) causes any undefined behavior *in a particular execution* (of which there
+may be many, e.g. when concurrency is involved). When Miri finds UB, your code is definitely
+unsound, but when Miri does not find UB, that may just be a case of insufficient coverage.
```
oli-obk, line 80: > Or arbitrary choices like when a thread switch occurs or simply real random numbers are obtained from the operating system.

It may not be clear otherwise that even 100% coverage (not just every code path being tested, but all combinations of code paths) may have UB left over

RalfJung: We already talk about that in the previous sentence though...

PR #3720, opened 2024-06-29
Fix libc::read shim: make it write to a buffer correct amount of bytes. Add tests for new behavior

safinaskar:
libc::read shim had a bug: if underlying real call libc::read(fd, buf, N) returns M, then
libc::read shim writes N bytes to buf instead of M. Remaining N - M bytes are filled with zeros.
This commit fixes this bug and adds tests for new behavior

### DIFF
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -394,7 +394,12 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         match result {
             Ok(read_bytes) => {
                 // If reading to `bytes` did not fail, we write those bytes to the buffer.
-                this.write_bytes_ptr(buf, bytes)?;
+                // Crucially, if fewer than `bytes.len()` bytes were read, only write
+                // that much into the output buffer!
+                this.write_bytes_ptr(
+                    buf,
+                    bytes[..usize::try_from(read_bytes).unwrap()].iter().copied(),
+                )?;
                 Ok(read_bytes)
             }
             Err(e) => {
--- a/tests/fail-dep/libc/libc-read-and-uninit-premature-eof.rs
+++ b/tests/fail-dep/libc/libc-read-and-uninit-premature-eof.rs
@@ -1,0 +1,27 @@
+//! We test that if we requested to read 4 bytes, but actually read 3 bytes,
+//! then 3 bytes (not 4) will be initialized.
+//@ignore-target-windows: no file system support on Windows
+//@compile-flags: -Zmiri-disable-isolation
+
+use std::ffi::CString;
+use std::fs::remove_file;
+use std::mem::MaybeUninit;
+
+#[path = "../../utils/mod.rs"]
+mod utils;
+
+fn main() {
+    let path =
+        utils::prepare_with_content("fail-libc-read-and-uninit-premature-eof.txt", &[1u8, 2, 3]);
+    let cpath = CString::new(path.clone().into_os_string().into_encoded_bytes()).unwrap();
+    unsafe {
+        let fd = libc::open(cpath.as_ptr(), libc::O_RDONLY);
+        assert_ne!(fd, -1);
+        let mut buf: MaybeUninit<[u8; 4]> = std::mem::MaybeUninit::uninit();
+        // Read 4 bytes from a 3-byte file.
+        assert_eq!(libc::read(fd, buf.as_mut_ptr().cast::<std::ffi::c_void>(), 4), 3);
+        buf.assume_init(); //~ERROR: Undefined Behavior: constructing invalid value at .value[3]: encountered uninitialized memory, but expected an integer
+        assert_eq!(libc::close(fd), 0);
+    }
+    remove_file(&path).unwrap();
+}
--- a/tests/pass-dep/libc/libc-fs.rs
+++ b/tests/pass-dep/libc/libc-fs.rs
@@ -35,6 +35,7 @@ fn main() {
     #[cfg(target_os = "linux")]
     test_sync_file_range();
     test_isatty();
+    test_read_and_uninit();
 }
 
 fn test_file_open_unix_allow_two_args() {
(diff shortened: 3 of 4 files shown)


RalfJung:
Thanks!
Could you please add a PR description explaining the goal of the PR? We already have tests of the file system APIs and if `read` did not initialize the buffer they would obviously fail.

It would be nice to have some tests that check the libc APIs (open/read/write/close) directly, basically a libc version of `src/shims/unix/fs.rs` -- see https://github.com/rust-lang/miri/issues/3179. If that is the goal of this PR then the filename and test content should be adjusted accordingly.

safinaskar:
@RalfJung , okay. I will remove draft status, when I'm done

RalfJung:
I am not sure if that generates a notification... please write `@rustbot ready` when it is ready.

@rustbot author

safinaskar:
@RalfJung , okay

safinaskar:
@RalfJung
@rustbot ready
See updated title and description

src/shims/unix/fd.rs
```diff
@@ -394,7 +394,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         match result {
             Ok(read_bytes) => {
                 // If reading to `bytes` did not fail, we write those bytes to the buffer.
-                this.write_bytes_ptr(buf, bytes)?;
+                this.write_bytes_ptr(buf, bytes[..usize::try_from(read_bytes).unwrap()].to_vec())?;
```
RalfJung, line 397:
```suggestion
                // Crucially, if fewer than `bytes.len()` bytes were read, only write
                // that much into the output buffer!
                this.write_bytes_ptr(buf, bytes[..usize::try_from(read_bytes).unwrap()].to_vec())?;
```
Also, `to_vec` should not be needed. `.iter().copied()` should work.

tests/fail-dep/libc/libc-read-and-uninit-premature-eof.rs
```diff
@@ -0,0 +1,25 @@
+// We test that if we requested to read 4 bytes, but actually read 3 bytes, then 3 bytes (not 4) will be initialized
+//@ignore-target-windows: we have no file deletion on Windows
```
RalfJung, line 2:
```suggestion
//@ignore-target-windows: no file system support on Windows
```

tests/fail-dep/libc/libc-read-and-uninit-premature-eof.rs
```diff
@@ -0,0 +1,25 @@
...
+
+fn main() {
+    let path =
+        utils::prepare_with_content("fail-libc-read-and-uninit-premature-eof.txt", &[1u8, 2, 3]);
+    let cpath = CString::new(path.clone().into_os_string().into_encoded_bytes()).unwrap();
+    unsafe {
+        let fd = libc::open(cpath.as_ptr(), libc::O_RDONLY);
+        assert_ne!(fd, -1);
+        let mut buf: MaybeUninit<[u8; 4]> = std::mem::MaybeUninit::uninit();
+        assert_eq!(libc::read(fd, buf.as_mut_ptr().cast::<std::ffi::c_void>(), 4), 3);
```
RalfJung, line 20:
```suggestion
        // Read 4 byte from a 3-byte file.
        assert_eq!(libc::read(fd, buf.as_mut_ptr().cast::<std::ffi::c_void>(), 4), 3);
```

tests/fail-dep/libc/libc-read-and-uninit.rs
RalfJung, line 1: What is the point of this test? I don't think this is a likely enough error to justify spending CI time on.

tests/pass-dep/libc/libc-read-and-uninit.rs
RalfJung, line 1: Please move these into libc-fs.rs

RalfJung (commented):
Ah, that's what this is about. :)  Good catch, and thanks for the PR!

safinaskar:
@RalfJung
@rustbot ready
I made all requested changes

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -362,3 +363,36 @@ fn test_isatty() {
...
+            let mut buf: MaybeUninit<[u8; 2]> = std::mem::MaybeUninit::uninit();
+            assert_eq!(libc::read(fd, buf.as_mut_ptr().cast::<std::ffi::c_void>(), 2), 2);
+            let buf = buf.assume_init();
+            assert_eq!(buf, [1, 2]);
+            assert_eq!(libc::close(fd), 0);
+        }
+        remove_file(&path).unwrap();
+    }
+    {
+        // We test that if we requested to read 4 bytes, but actually read 3 bytes, then 3 bytes (not 4) will be overwritten, and remaining byte will be left as is
```
RalfJung, line 385:
```suggestion
        // We test that if we requested to read 4 bytes, but actually read 3 bytes, then
        // 3 bytes (not 4) will be overwritten, and the remaining byte will be left as-is.
```

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -362,3 +363,36 @@ fn test_isatty() {
         remove_file(&path).unwrap();
     }
 }
+
+fn test_read_and_uninit() {
+    use std::mem::MaybeUninit;
+    {
+        // We test that libc::read initializes its buffer
```
RalfJung, line 370:
```suggestion
        // We test that libc::read initializes its buffer.
```

tests/fail-dep/libc/libc-read-and-uninit-premature-eof.rs
```diff
@@ -0,0 +1,26 @@
+// We test that if we requested to read 4 bytes, but actually read 3 bytes, then 3 bytes (not 4) will be initialized
```
RalfJung, line 1:
```suggestion
//! We test that if we requested to read 4 bytes, but actually read 3 bytes,
//! then 3 bytes (not 4) will be initialized.
```

RalfJung:
Just some final nits, please make these comments proper sentences and line-wrap them.

Usually I'd just apply these patches myself, but you seem to have disabled the option in the PR to let maintainers push to your branch. I can only recommend to enable that option, it saves you some work that maintainers can do for you. :)

@rustbot author

safinaskar:
@RalfJung
@rustbot ready
I made all requested changes

RalfJung:
Thanks, looks good!
@bors r+

PR #3724, opened 2024-07-01
Use the symbol_name query instead of trying to infer from the link_name attribute

bjorn3:
This prevents the calculated name from going out of sync with exported_symbols. It also avoids having to special case the panic_impl lang item.

It also makes it easier to fix miri with https://github.com/rust-lang/rust/pull/127173.

### DIFF
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -954,7 +954,7 @@ impl<'tcx> Machine<'tcx> for MiriMachine<'tcx> {
             // foreign function
             // Any needed call to `goto_block` will be performed by `emulate_foreign_item`.
             let args = ecx.copy_fn_args(args); // FIXME: Should `InPlace` arguments be reset to uninit?
-            let link_name = ecx.item_link_name(instance.def_id());
+            let link_name = Symbol::intern(ecx.tcx.symbol_name(instance).name);
             return ecx.emulate_foreign_item(link_name, abi, &args, dest, ret, unwind);
         }
 
@@ -1050,7 +1050,7 @@ impl<'tcx> Machine<'tcx> for MiriMachine<'tcx> {
         ecx: &MiriInterpCx<'tcx>,
         def_id: DefId,
     ) -> InterpResult<'tcx, StrictPointer> {
-        let link_name = ecx.item_link_name(def_id);
+        let link_name = Symbol::intern(ecx.tcx.symbol_name(Instance::mono(*ecx.tcx, def_id)).name);
         if let Some(&ptr) = ecx.machine.extern_statics.get(&link_name) {
             // Various parts of the engine rely on `get_alloc_info` for size and alignment
             // information. That uses the type information of this static.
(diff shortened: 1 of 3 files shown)


src/shims/foreign_items.rs
```diff
@@ -46,24 +46,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         unwind: mir::UnwindAction,
     ) -> InterpResult<'tcx, Option<(&'tcx mir::Body<'tcx>, ty::Instance<'tcx>)>> {
         let this = self.eval_context_mut();
-        let tcx = this.tcx.tcx;
 
         // Some shims forward to other MIR bodies.
         match link_name.as_str() {
-            // This matches calls to the foreign item `panic_impl`.
-            // The implementation is provided by the function with the `#[panic_handler]` attribute.
-            "panic_impl" => {
```
RalfJung, line 55: I am confused, why is this not needed any more?

src/machine.rs
```diff
@@ -954,7 +954,7 @@ impl<'tcx> Machine<'tcx> for MiriMachine<'tcx> {
             // foreign function
             // Any needed call to `goto_block` will be performed by `emulate_foreign_item`.
             let args = ecx.copy_fn_args(args); // FIXME: Should `InPlace` arguments be reset to uninit?
-            let link_name = ecx.item_link_name(instance.def_id());
+            let link_name = Symbol::intern(ecx.tcx.symbol_name(instance).name);
             return ecx.emulate_foreign_item(link_name, abi, &args, dest, ret, unwind);
```
RalfJung, line 958: emulate_foreign_item anyway makes this an `&str`... so the interning seems unnecessary here? It's probably overall easier to pass around `&str`.

src/shims/foreign_items.rs
bjorn3: The previous code would incorrectly infer the symbol name of this function as panic_impl rather than rusy_begin_unwind, necessitating a manual forward to the real implementation.

src/machine.rs
bjorn3: `symbol_name` returns a `SymbolName<'tcx>` which is a wrapper around `&'tcx str`. The extra lifetime is giving me quite a bit of trouble. I could pass around String instead, but that is likely less efficient.

RalfJung: I thought it'd be trivial to just make this `&str` everywhere (so `emulate_foreign_item` would not even know that the lifetime is `tcx`)... but I guess it is not that simple, then?

bjorn3: There are a couple of types without a 'tcx lifetime, some functions return a symbol name as `String` and I also saw a bunch of borrowck errors for whatever reason.

RalfJung: :shrug: okay not sure what happens and I don't have the time to take a look myself right now

PR #3726, opened 2024-07-01
Implement the `_mm256_zeroupper` and `_mm256_zeroall` intrinsics

TDecking:
These two intrinsics were missing from the original implementation of the AVX intrinsics.
Fortunately their implementation is trivial.

### DIFF
--- a/src/shims/x86/avx.rs
+++ b/src/shims/x86/avx.rs
@@ -338,6 +338,17 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
                 this.write_scalar(Scalar::from_i32(res.into()), dest)?;
             }
+            // Used to implement the `_mm256_zeroupper` and `_mm256_zeroall` functions.
+            // These function clear out the upper 128 bits of all avx registers or
+            // zero out all avx registers respectively.
+            "vzeroupper" | "vzeroall" => {
+                // These functions are purely a performance hint for the CPU.
+                // Any registers currently in use will be saved beforehand by the
+                // compiler, making these functions no-ops.
+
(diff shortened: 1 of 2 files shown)


RalfJung:
These did come up before, and we decided to not have them because we were concerned about the semantics...

So you are sure it is correct to insert these *anywhere at all in a Rust program* and that will never change program behavior?
Cc @workingjubilee @Amanieu

Amanieu:
Yes, these are hints to the compiler and have no effect otherwise.

src/shims/x86/avx.rs
```diff
@@ -338,6 +338,14 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
                 this.write_scalar(Scalar::from_i32(res.into()), dest)?;
             }
+            // Used to implement the `_mm256_zeroupper` and `_mm256_zeroall` functions.
+            // These function clear out the upper 128 bits of all avx registers or
+            // zero out all avx registers respectively.
+            "vzeroupper" | "vzeroall" => {
+                // These functions are purely a performance hint for the CPU.
+                // Any registers currently in use will be saved beforehand by the
+                // compiler, making these functions no-ops.
```
RalfJung, line 347: This needs something like `let [] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;` to make sure it gets called with the right ABI and the right number of arguments.

TDecking: Done.

RalfJung:
> Yes, these are hints to the compiler and have no effect otherwise.

All right, seems good to go then. Thanks!

@bors r+

PR #3732, opened 2024-07-04
TB: Refine protector end semantics

JoJoDeveloping:
Tree Borrows has protector end tag semantics, namely that protectors ending cause a [special implicit read](https://perso.crans.org/vanille/treebor/diff.0.html) on all locations protected by that protector that have actually been accessed. See also #3067.

While this is enough for ensuring protectors allow adding/reordering reads, it does not prove that one can reorder writes. For this, we need to make this stronger, by making this implicit read be a write in cases when there was a write to the location protected by that protector, i.e. if the permission is `Active`.

There is a test that shows why this behavior is necessary, see `tests/fail/tree_borrows/protector-write-lazy.rs`.

### DIFF
--- a/src/borrow_tracker/tree_borrows/diagnostics.rs
+++ b/src/borrow_tracker/tree_borrows/diagnostics.rs
@@ -19,7 +19,7 @@ pub enum AccessCause {
     Explicit(AccessKind),
     Reborrow,
     Dealloc,
-    FnExit,
+    FnExit(AccessKind),
 }
 
 impl fmt::Display for AccessCause {
@@ -28,7 +28,8 @@ impl fmt::Display for AccessCause {
             Self::Explicit(kind) => write!(f, "{kind}"),
             Self::Reborrow => write!(f, "reborrow"),
             Self::Dealloc => write!(f, "deallocation"),
-            Self::FnExit => write!(f, "protector release"),
+            Self::FnExit(AccessKind::Read) => write!(f, "protector release read"),
+            Self::FnExit(AccessKind::Write) => write!(f, "protector release write"),
         }
     }
 }
@@ -40,7 +41,7 @@ impl AccessCause {
             Self::Explicit(kind) => format!("{rel} {kind}"),
             Self::Reborrow => format!("reborrow (acting as a {rel} read access)"),
             Self::Dealloc => format!("deallocation (acting as a {rel} write access)"),
-            Self::FnExit => format!("protector release (acting as a {rel} read access)"),
+            Self::FnExit(kind) => format!("protector release (acting as a {rel} {kind})"),
         }
     }
 }
--- a/src/borrow_tracker/tree_borrows/diagnostics.rs
+++ b/src/borrow_tracker/tree_borrows/diagnostics.rs
@@ -28,8 +28,11 @@ impl fmt::Display for AccessCause {
             Self::Explicit(kind) => write!(f, "{kind}"),
             Self::Reborrow => write!(f, "reborrow"),
             Self::Dealloc => write!(f, "deallocation"),
-            Self::FnExit(AccessKind::Read) => write!(f, "protector release read"),
-            Self::FnExit(AccessKind::Write) => write!(f, "protector release write"),
+            // This is dead code, since the protector release access itself can never
+            // cause UB (while the protector is active, if some other access invalidates
+            // further use of the protected tag, that is immediate UB).
+            // Describing the cause of UB is the only time this function is called.
+            Self::FnExit(_) => unreachable!("protector accesses can never be the source of UB"),
         }
     }
 }
--- a/tests/fail/tree_borrows/protector-write-lazy.rs
+++ b/tests/fail/tree_borrows/protector-write-lazy.rs
@@ -1,0 +1,35 @@
+//@compile-flags: -Zmiri-tree-borrows
+// This test tests that TB's protector end semantics correctly ensure
+// that protected activated writes can be reordered.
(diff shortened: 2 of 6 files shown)


tests/fail/tree_borrows/protector-write-lazy.rs
```diff
@@ -0,0 +1,35 @@
...
+    // We need to ensure that the `funky_ref` is unusable even if the write to `ref_to_fst_elem`
+    // happens before the retag.
+    *ref_to_fst_elem = 42;
+    let funky_ref_lazy_on_fst_elem;
+    {
+        // this creates a reference that is Reserved Lazy on the first element (offset 0).
+        // It does so by doing a proper retag on the second element (offset 1), which is fine
+        // since nothing else happens at that offset, but the lazy init mechanism means it's
+        // also reserved at offset 0, but not initialized.
+        funky_ref_lazy_on_fst_elem =
```
RalfJung, line 18: Is there a reason you did not make this `let funky_ref_lazy_on_fst_elem = ...`?

tests/fail/tree_borrows/protector-write-lazy.rs
```diff
@@ -0,0 +1,35 @@
...
+            unsafe { (&mut *(ptr_to_vec.wrapping_add(1))) as *mut i32 }.wrapping_sub(1);
+    }
+    return funky_ref_lazy_on_fst_elem;
```
RalfJung, line 21:
```suggestion
    // If we write to `ref_to_fst_elem` here, then any further access to `funky_ref_lazy_on_fst_elem` would
    // definitely be UB. Since the compiler ought to be able to reorder the write of `42` above down to
    // here, that means we want this program to also be UB.
    return funky_ref_lazy_on_fst_elem;
```

src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -28,7 +28,7 @@ impl fmt::Display for AccessCause {
             Self::Explicit(kind) => write!(f, "{kind}"),
             Self::Reborrow => write!(f, "reborrow"),
             Self::Dealloc => write!(f, "deallocation"),
-            Self::FnExit => write!(f, "protector release"),
+            Self::FnExit(kind) => write!(f, "protector release ({kind})"),
```
RalfJung, line 31: "protector release read" and "protector release write" would be a bit more concise than "protector release (read access)".
```suggestion
            Self::FnExit(AccessKind::Read) => write!(f, "protector release read"),
            Self::FnExit(AccessKind::Write) => write!(f, "protector release write"),
```
But also we don't seem to have any test that even uses this string? That's a bit unfortunate. Oh well.

JoJoDeveloping: Indeed, I could not figure out where this is used so I didn't think about how it'd look in context.

RalfJung: It seems the only place it is used is [here](https://github.com/rust-lang/miri/blob/81f85ec72e1a98c244838a2a3d5adc5a4a75276d/src/borrow_tracker/tree_borrows/diagnostics.rs#L306)

JoJoDeveloping: I think this can never happen..

RalfJung: Oh, really? Why?

RalfJung: Ah I guess the protectors ensure these accesses would be allowed, that's the entire point.

JoJoDeveloping: well, you would need a protected initialized tag that's a cousin to another protected initialized tag. This means none of the tags can be `Active` since it's impossible to create that situation. It is possible for them to be `Frozen` or `Reserved`, and then they survive the foreign read emitted by the protector end semantics.

RalfJung, line 31:
```suggestion
            // These are dead code, since the protector release access itself can never
            // cause UB (while the protector is active, if some other access invalidates
            // further use of the protected tag, that is immediate UB).
            // Describing the cause of UB is the only time this function is called.
            Self::FnExit(_) => unreachable!("protector accesses can never be the source of UB"),
```
This way if someone out there happens to somehow trigger this, we'll know. :)

JoJoDeveloping: Ah, turns out this code path is triggered, for instance it produces [this](https://github.com/rust-lang/miri/blob/2982438e8970b194fb4cdd2e022c3a4356c7dfec/tests/fail/tree_borrows/spurious_read.stderr#L27) diagnostic.

tests/fail/tree_borrows/protector-write-lazy.rs
JoJoDeveloping: No reason, just being derpy. :slightly_smiling_face:

tests/fail/tree_borrows/protector-write-lazy.rs
JoJoDeveloping: Added!

src/borrow_tracker/tree_borrows/diagnostics.rs
JoJoDeveloping: I've added it, and all tests still passed.

RalfJung:
Looks good, thanks. :)

@bors r+

PR #3739, opened 2024-07-07
Implement support for multiple TLS destructors on macOS

joboet:
I want to get rid of [this `#[cfg]` block](https://github.com/rust-lang/rust/blob/98dcbae5c9ac615d5acfbf42d42b19a77cb9ec11/library/std/src/thread/mod.rs#L195-L211) in `std`, but it is currently required for miri, as it does not support multiple macOS TLS destructors. This is not true for the platform itself, however, as can be observed in the [implementation](https://github.com/apple-oss-distributions/dyld/blob/d552c40cd1de105f0ec95008e0e0c0972de43456/dyld/DyldRuntimeState.cpp#L2239).

### DIFF
--- a/src/shims/tls.rs
+++ b/src/shims/tls.rs
@@ -36,9 +36,9 @@ pub struct TlsData<'tcx> {
     /// pthreads-style thread-local storage.
     keys: BTreeMap<TlsKey, TlsEntry<'tcx>>,
 
-    /// A single per thread destructor of the thread local storage (that's how
-    /// things work on macOS) with a data argument.
-    macos_thread_dtors: BTreeMap<ThreadId, (ty::Instance<'tcx>, Scalar)>,
+    /// On macOS, each thread holds a list of destructor functions with their
+    /// respective data arguments.
+    macos_thread_dtors: BTreeMap<ThreadId, Vec<(ty::Instance<'tcx>, Scalar)>>,
 }
 
 impl<'tcx> Default for TlsData<'tcx> {
@@ -119,26 +119,15 @@ impl<'tcx> TlsData<'tcx> {
         }
     }
 
-    /// Set the thread wide destructor of the thread local storage for the given
-    /// thread. This function is used to implement `_tlv_atexit` shim on MacOS.
-    ///
-    /// Thread wide dtors are available only on MacOS. There is one destructor
-    /// per thread as can be guessed from the following comment in the
-    /// [`_tlv_atexit`
-    /// implementation](https://github.com/opensource-apple/dyld/blob/195030646877261f0c8c7ad8b001f52d6a26f514/src/threadLocalVariables.c#L389):
-    ///
-    /// NOTE: this does not need locks because it only operates on current thread data
-    pub fn set_macos_thread_dtor(
+    /// Add a thread local storage destructor for the given thread. This function
+    /// is used to implement the `_tlv_atexit` shim on MacOS.
+    pub fn add_macos_thread_dtor(
         &mut self,
         thread: ThreadId,
         dtor: ty::Instance<'tcx>,
         data: Scalar,
     ) -> InterpResult<'tcx> {
-        if self.macos_thread_dtors.insert(thread, (dtor, data)).is_some() {
-            throw_unsup_format!(
-                "setting more than one thread local storage destructor for the same thread is not supported"
-            );
-        }
+        self.macos_thread_dtors.entry(thread).or_default().push((dtor, data));
         Ok(())
     }
 
@@ -202,6 +191,10 @@ impl<'tcx> TlsData<'tcx> {
         for TlsEntry { data, .. } in self.keys.values_mut() {
             data.remove(&thread_id);
         }
+
+        if let Some(dtors) = self.macos_thread_dtors.remove(&thread_id) {
+            assert!(dtors.is_empty(), "the destructors should have already been run");
+        }
     }
 }
 
@@ -212,7 +205,7 @@ impl VisitProvenance for TlsData<'_> {
         for scalar in keys.values().flat_map(|v| v.data.values()) {
             scalar.visit_provenance(visit);
         }
-        for (_, scalar) in macos_thread_dtors.values() {
+        for (_, scalar) in macos_thread_dtors.values().flatten() {
             scalar.visit_provenance(visit);
         }
     }
@@ -225,6 +218,7 @@ pub struct TlsDtorsState<'tcx>(TlsDtorsStatePriv<'tcx>);
 enum TlsDtorsStatePriv<'tcx> {
     #[default]
     Init,
+    MacOsDtors,
     PthreadDtors(RunningDtorState),
     /// For Windows Dtors, we store the list of functions that we still have to call.
     /// These are functions from the magic `.CRT$XLB` linker section.
@@ -243,11 +237,10 @@ impl<'tcx> TlsDtorsState<'tcx> {
                 Init => {
                     match this.tcx.sess.target.os.as_ref() {
                         "macos" => {
-                            // The macOS thread wide destructor runs "before any TLS slots get
-                            // freed", so do that first.
-                            this.schedule_macos_tls_dtor()?;
-                            // When that destructor is done, go on with the pthread dtors.
-                            break 'new_state PthreadDtors(Default::default());
+                            // macOS has a _tlv_atexit function that allows
+                            // registering destructors without associated keys.
+                            // These are run first.
+                            break 'new_state MacOsDtors;
                         }
                         _ if this.target_os_is_unix() => {
                             // All other Unixes directly jump to running the pthread dtors.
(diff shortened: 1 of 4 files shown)


RalfJung:
It *is* however UB to set another TLS dtor while TLS dtors are running. So Miri should probably detect that.

Curiously this could also mean one can cause UB in the std test suite, once that `cfg` is removed, by having both a `realstd::thread_local` and a `std::thread_local`, and then initializing one in the dtor of the other.

src/shims/tls.rs
```diff
@@ -119,26 +119,18 @@ impl<'tcx> TlsData<'tcx> {
...
-    /// per thread as can be guessed from the following comment in the
-    /// [`_tlv_atexit`
-    /// implementation](https://github.com/opensource-apple/dyld/blob/195030646877261f0c8c7ad8b001f52d6a26f514/src/threadLocalVariables.c#L389):
-    ///
-    /// NOTE: this does not need locks because it only operates on current thread data
-    pub fn set_macos_thread_dtor(
+    /// Add a thread local storage for the given thread. This function is used to
+    /// implement `_tlv_atexit` shim on MacOS.
+    /// FIXME: also implement `__cxa_thread_atexit_impl` for Linux using this, see
+    /// <https://sourceware.org/glibc/wiki/Destructor%20support%20for%20thread_local%20variables>
```
RalfJung, line 125: This seems better suited for an issue rather than a FIXME in the code.

src/shims/tls.rs
```diff
@@ -328,21 +320,26 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-            )?;
+        if let Some(dtors) = this.machine.tls.macos_thread_dtors.remove(&thread_id) {
+            // macOS runs TLS destructors in the reverse order of their registration.
+            // See https://github.com/apple-oss-distributions/dyld/blob/d552c40cd1de105f0ec95008e0e0c0972de43456/dyld/DyldRuntimeState.cpp#L2277
+            // FIXME: Old macOS versions had a use-after-free bug that triggered when
+            // _tlv_atexit was run inside a destructor. Maybe miri should catch that?
+            for (instance, data) in dtors.into_iter().rev() {
+                trace!("Running macos dtor {:?} on {:?} at {:?}", instance, data, thread_id);
+
+                this.call_function(
```
RalfJung, line 335: No this is not right. `call_function` just pushes the stack frame, so what this does is push a stack frame for each of these functions and then run them all, making the stack look as if the 2nd dtor was called by the first.

There's a big surrounding interpreter loop here, and the next dtor needs to be started only once this dtor finishes running, when the stack is empty again.  See what we do for `WindowsDtors` for how that can be implemented.

RalfJung (commented):
Please also add a test that directly calls `_tlv_atexit`, twice, and ensures both get run.

joboet:
> It _is_ however UB to set another TLS dtor while TLS dtors are running, as far as I know. So Miri should probably detect that.

It isn't in newer versions of macOS, they [fixed that bug](https://github.com/apple-oss-distributions/dyld/blob/d552c40cd1de105f0ec95008e0e0c0972de43456/dyld/DyldRuntimeState.cpp#L2284).

RalfJung:
And that is an intentional guarantee?

Seems like the behavior is to run the newest things first. So probably what should happen is quite different from `WindowsDtors` then. We have a new `MacosDtors` state that we jump to on macOS targets. It calls `schedule_macos_tls_dtor`, which should take the latest dtor from the list and run it, but we stay in the  `MacosDtors` state. So when that dtor is done, we again take the latest off the list (including when more were added now). This repeats until we `on_stack_empty` in the `MacosDtors` state sees that the macOS dtor list is empty; then we move on to `PthreadDtors(Default::default())`.

src/shims/tls.rs
RalfJung: To give a bit of context here, each thread has a `TlsDtorsState` and each time that thread's stack is empty, the main loop calls `on_stack_empty` on it. If that returns `Pending`, the thread sticks around. This *may* also push a new stack frame, in which case that now gets executed, until the stack is empty and `on_stack_empty` is called again. This repeats until `on_stack_empty` returns `Ready`, at which point the thread is fully terminated and never scheduled again.

joboet:
It seems very unlikely that they would remove that fix. But of course, this is Apple we're talking about, so I don't think we'll ever get a proper guarantee. It doesn't matter too much for us anyway, since we still support the older platforms that crash.

src/shims/tls.rs
```diff
@@ -202,6 +191,8 @@ impl<'tcx> TlsData<'tcx> {
         for TlsEntry { data, .. } in self.keys.values_mut() {
             data.remove(&thread_id);
         }
+
+        self.macos_thread_dtors.remove(&thread_id);
```
RalfJung, line 195: Aren't these always empty because we ran the destructors?

tests/pass/tls/macos_tlv_atexit.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+extern "C" fn run(_: *mut u8) {
+    println!("dtor was called");
+}
+
+fn main() {
+    thread::spawn(|| {
+        register(run);
+        register(run);
```
RalfJung, line 28: This does not ensure that they are run in the right order, also wrt running newly registered dtors immediately.

src/shims/tls.rs
```diff
@@ -328,12 +327,15 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-    fn schedule_macos_tls_dtor(&mut self) -> InterpResult<'tcx> {
+    /// Schedule the macOS thread local storage destructors to be executed.
+    fn schedule_macos_tls_dtor(&mut self) -> InterpResult<'tcx, Poll<()>> {
         let this = self.eval_context_mut();
         let thread_id = this.active_thread();
-        if let Some((instance, data)) = this.machine.tls.macos_thread_dtors.remove(&thread_id) {
+        // macOS keeps track of TLS destructors in a stack. If a destructor
+        // registers another destructor, it will be run next.
+        // See https://github.com/apple-oss-distributions/dyld/blob/d552c40cd1de105f0ec95008e0e0c0972de43456/dyld/DyldRuntimeState.cpp#L2277
+        let dtor = this.machine.tls.macos_thread_dtors.get_mut(&thread_id).and_then(Vec::pop);
```
RalfJung, line 337: Seems like a good place to use `let Some(dtor) = ... else { ... }` and avoid rightwards drift.

joboet: It's only one indent, and the current version mirrors the UNIX one, so I'm inclined not to change it, unless you insist.

tests/pass/tls/macos_tlv_atexit.rs
joboet: Sure!

src/shims/tls.rs
joboet: Yes, but it seems neater to free the memory of the `Vec` here.

RalfJung: Can you add an assertion ensuring the vector we pop is empty?

tests/pass/tls/macos_tlv_atexit.rs
```diff
@@ -1,34 +1,40 @@
...
 }
 
-fn register(dtor: extern "C" fn(*mut u8)) {
-    unsafe {
-        _tlv_atexit(dtor, ptr::null_mut());
+fn register<F>(f: F)
+where
+    F: FnOnce() + 'static,
+{
+    unsafe extern "C" fn run<F>(data: *mut u8)
```
RalfJung, line 13: Please add a comment saying that `data` is expected to actually be a `Box<F>`

tests/pass/tls/macos_tlv_atexit.rs
```diff
@@ -1,34 +1,40 @@
...
+    where
+        F: FnOnce() + 'static,
+    {
+        unsafe { (*Box::from_raw(data as *mut F))() }
```
RalfJung, line 17: It took me a bit to realize you are also calling the function here. Please let-bind the `Box::from_raw(data as *mut F)` so that this becomes more clear.

tests/pass/tls/macos_tlv_atexit.rs
```diff
@@ -1,34 +1,40 @@
...
     })
     .join()
     .unwrap();
 
-    register(register_another);
+    println!("exiting main");
+    register(|| println!("dtor 5"));
+    register(|| {
+        println!("registering dtor in dtor 4");
+        register(|| println!("dtor 3"));
```
RalfJung, line 38: Shouldn't 3 and 4 be swapped here?

joboet: Oh yeah, whoops!

PR #3743, opened 2024-07-12
Add `pread` and `pwrite` shims

newpavlov:
Requested in #2057.

### DIFF
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -36,6 +36,30 @@ pub trait FileDescription: std::fmt::Debug + Any {
         throw_unsup_format!("cannot write to {}", self.name());
     }
 
+    /// Reads as much as possible into the given buffer from a given offset,
+    /// and returns the number of bytes read.
+    fn pread<'tcx>(
+        &mut self,
+        _communicate_allowed: bool,
+        _bytes: &mut [u8],
+        _offset: u64,
+        _ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, io::Result<usize>> {
+        throw_unsup_format!("cannot pread from {}", self.name());
+    }
+
+    /// Writes as much as possible from the given buffer starting at a given offset,
+    /// and returns the number of bytes written.
+    fn pwrite<'tcx>(
+        &mut self,
+        _communicate_allowed: bool,
+        _bytes: &[u8],
+        _offset: u64,
+        _ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, io::Result<usize>> {
+        throw_unsup_format!("cannot pwrite to {}", self.name());
+    }
+
     /// Seeks to the given offset (which can be relative to the beginning, end, or current position).
     /// Returns the new position from the start of the stream.
     fn seek<'tcx>(
@@ -380,7 +404,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         Ok((-1).into())
     }
 
-    fn read(&mut self, fd: i32, buf: Pointer, count: u64) -> InterpResult<'tcx, i64> {
+    /// Read data from `fd` into buffer specified by `buf` and `count`.
+    ///
+    /// If `offset` is `None`, reads data from current cursor position associated with `fd`
+    /// and updates cursor position on completion. Otherwise, reads from the specified offset
+    /// and keeps the cursor unchanged.
+    fn read(
+        &mut self,
+        fd: i32,
+        buf: Pointer,
+        count: u64,
+        offset: Option<i128>,
+    ) -> InterpResult<'tcx, i64> {
         let this = self.eval_context_mut();
 
         // Isolation check is done via `FileDescriptor` trait.
@@ -398,25 +433,31 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let communicate = this.machine.communicate();
 
         // We temporarily dup the FD to be able to retain mutable access to `this`.
-        let Some(file_descriptor) = this.machine.fds.dup(fd) else {
+        let Some(fd) = this.machine.fds.dup(fd) else {
             trace!("read: FD not found");
             return this.fd_not_found();
         };
 
-        trace!("read: FD mapped to {:?}", file_descriptor);
+        trace!("read: FD mapped to {fd:?}");
         // We want to read at most `count` bytes. We are sure that `count` is not negative
         // because it was a target's `usize`. Also we are sure that its smaller than
         // `usize::MAX` because it is bounded by the host's `isize`.
         let mut bytes = vec![0; usize::try_from(count).unwrap()];
-        // `File::read` never returns a value larger than `count`,
-        // so this cannot fail.
-        let result = file_descriptor
-            .borrow_mut()
-            .read(communicate, &mut bytes, this)?
-            .map(|c| i64::try_from(c).unwrap());
-        drop(file_descriptor);
+        let result = match offset {
+            None => fd.borrow_mut().read(communicate, &mut bytes, this),
+            Some(offset) => {
+                let Ok(offset) = u64::try_from(offset) else {
+                    let einval = this.eval_libc("EINVAL");
+                    this.set_last_error(einval)?;
+                    return Ok(-1);
+                };
+                fd.borrow_mut().pread(communicate, &mut bytes, offset, this)
+            }
+        };
+        drop(fd);
 
-        match result {
+        // `File::read` never returns a value larger than `count`, so this cannot fail.
+        match result?.map(|c| i64::try_from(c).unwrap()) {
             Ok(read_bytes) => {
                 // If reading to `bytes` did not fail, we write those bytes to the buffer.
                 // Crucially, if fewer than `bytes.len()` bytes were read, only write
@@ -434,7 +475,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         }
     }
 
-    fn write(&mut self, fd: i32, buf: Pointer, count: u64) -> InterpResult<'tcx, i64> {
+    fn write(
+        &mut self,
+        fd: i32,
+        buf: Pointer,
+        count: u64,
+        offset: Option<i128>,
+    ) -> InterpResult<'tcx, i64> {
         let this = self.eval_context_mut();
 
         // Isolation check is done via `FileDescriptor` trait.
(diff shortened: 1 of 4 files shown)


src/shims/unix/fd.rs
```diff
@@ -36,6 +36,30 @@ pub trait FileDescription: std::fmt::Debug + Any {
...
+    /// Reads as much as possible into the given buffer from a given offset,
+    /// and returns the number of bytes read.
+    fn pread<'tcx>(
+        &mut self,
+        _communicate_allowed: bool,
+        _bytes: &mut [u8],
+        _offset: u64,
+        _ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, io::Result<usize>> {
+        throw_unsup_format!("cannot pread from {}", self.name());
```
saethlin, line 48: We _specifically_ try to avoid this behavior. Miri is a cross-interpreter; we do everything reasonable to ensure that shims behave the same no matter what host is selected.

In this case, I think these shims can be implemented so that they work across targets by doing an explicit seek first; this is definitely slower but it is portable and should fulfill the required API.

newpavlov: IIUC Windows simply does not have a similar syscall. This is why `FileExt` is gated on `unix` in the standard library. See this [discussion](https://internals.rust-lang.org/t/19649) for more information. It may be possible to emulate it with `Seek`, but it would require either locking, or file descriptor duplication.

newpavlov: I would prefer for this PR to be merged as-is and to discuss Windows emulation in a separate PR/issue.

newpavlov:
Any updates on this? Lack of these shims is the main blocker for testing our code with MIRI.

RalfJung:
This is in my review queue, but unfortunately I currently don't have a lot of time for reviewing -- sorry.

In terms of the general approach, as Ben said this is quite the departure from how all file system code works in Miri so far -- which is via a a portable implementation on top of `std`, using platform-specific functions only when absolutely needed. In that sense I think I would strongly prefer exploring an approach that only needs `std` APIs. Miri is entirely sequential, so something like "record old position in file, seek, read, seek back to old position" should be a correct emulation I think?

src/shims/unix/foreign_items.rs
```diff
@@ -105,6 +105,70 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // Now, `result` is the value we return back to the program.
                 this.write_scalar(Scalar::from_target_isize(result, this), dest)?;
             }
+            "pread" => {
+                if matches!(&*this.tcx.sess.target.os, "windows") {
```
RalfJung, line 109: This is in the "unix" file, so "windows" is an impossible value here -- the check is not needed.

src/shims/unix/fs.rs
```diff
@@ -61,7 +57,22 @@ impl FileDescription for FileHandle {
...
     ) -> InterpResult<'tcx, io::Result<usize>> {
         assert!(communicate_allowed, "isolation should have prevented even opening a file");
-        Ok(self.file.read_at(bytes, offset))
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::FileExt;
+            Ok(self.file.read_at(bytes, offset))
+        }
+        // Emulate pread using seek + read + seek to restore cursor position.
+        // Correctness of this emulation relies on sequential nature of Miri execution.
```
saethlin, line 66: What exactly is the problem with always using the portable emulation?

newpavlov: Why use a subpar emulation when we have access to the original?

RalfJung: One shared codepath for everything means better test coverage. The possibility of subtle breakage is an argument *for* a single implementation.

In which sense is the Unix user experience pessimized? The number of syscalls is really irrelevant, Miri is so much slower than native execution that performance is not a valid argument.

RalfJung: >  I do not care about Windows targets at all

This is about windows *hosts*. Windows targets don't have any FS emulation, so they must currently use Linux targets when using Miri with programs that access the file system. That makes it even more relevant to have a maximally portable implementation in Miri.

newpavlov: Personally, I would prefer to not have this emulation in the first place. The `is_symlink` emulation is not 100% equivalent to the `O_NOFOLLOW` flag, e.g. `is_symlink` may return `true` in case of an IO error, while `open` will succeed. In my opinion, it's better to to deal with reliable and straightforward shims instead of potentially problematic emulations.

If you insist on using the emulation for all hosts, I will do it.

>In which sense is the Unix user experience pessimized?

The requirement of having a Windows emulation means that it's not trivial to add UNIX APIs which do not have direct analogs in Windows. It means that it requires much more effort to implement a shim, or that it will not be implemented at all, thus limiting Miri usefulness for UNIX users.

Arguably, the burden of implementing emulation should be on users who need it. Lack of emulation does not mean that Miri is uselles, only that you need a UNIX host (or WSL) to run Miri tests for trickier programs.

RalfJung: It seems you partially mixed this up with the other PR? There's no symlinks here...

> The requirement of having a Windows emulation means that it's not trivial to add UNIX APIs which do not have direct analogs in Windows. It means that it requires much more effort to implement a shim, or that it will not be implemented at all, thus limiting Miri usefulness for UNIX users.

Cross-target-support is a pretty nifty feature of Miri and increases its usefulness. So there's a tradeoff here. If something is impossible to implement or requires unreasonable effort, we have [accepted](https://github.com/rust-lang/miri/blob/c9c887b7170ef3fdc608ae7f876dac7782e92cf0/src/shims/unix/fs.rs#L327-L332) `cfg(unix)` code in the past and will do so again. But if it is relatively simple then we take the portable route.

newpavlov: >It seems you partially mixed this up with the other PR? There's no symlinks here...

Yes, my mistake. But the same applies to the seek-based emulation. For example, because of race conditions it may fail where `pread`/`pwrite` would've succeeded or return error codes may be different. There is less probability for something going wrong with a straightforward shim. The testing argument means that you trade experience of most users for a probability of improving experience for a minority.

Do you insist on using the emulation for all hosts (in this and in the `O_NOFOLLOW` PR)?

>Cross-target-support is a pretty nifty feature of Miri and increases its usefulness.

I do not argue against its usefulness. My point is that such emulations should be done by users who need them. I find it unlikely that Windows developers will test programs which rely on `pread`/`pwrite` or `O_NOFOLLOW`.

RalfJung: > because of race conditions

Which race conditions? Miri is sequential.

> Do you insist on using the emulation for all hosts (in this and in the O_NOFOLLOW PR)?

This is a case-by-case thing. Currently I am trying to understand the tradeoff for this PR.

saethlin: > because of race conditions

Are you trying to talk about a race on the file descriptor's state?

newpavlov: >Which race conditions? Miri is sequential.

With outside environment. For example, if we test a program which works in tandem with other program.

In general, I do not want to think about potential ways for the emulation to break or have an observably different behavior. But it's more important for me to have `pread`/`pwrite` support in Miri, so if using the emulation for all hosts will make acceptance of this PR easier and quicker, I am fine with changing the code accordingly. I will create a commit with such changes, if you'll decide that we do not need emulation on UNIX hosts after all, we can revert it.

RalfJung: I am trying to understand which downsides you are seeing. The outside environment cannot observe or change our file descriptor state (Miri doesn't do FD sharing with other processes) so I fail to see the possible problems there.

But if you prefer we can also skip this and just go with seek-based emulation on all hosts.

src/shims/unix/fd.rs
```diff
@@ -463,4 +487,116 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
         this.try_unwrap_io_result(result)
     }
+
+    fn pread(
+        &mut self,
+        fd: i32,
+        buf: Pointer,
+        count: u64,
+        offset: i128,
+    ) -> InterpResult<'tcx, i64> {
```
RalfJung, line 497: This is largely a duplicate of `read`, right? Instead of duplicating all that code, would it be worth making this a single function? It could take an `Option` offset to indicate which mode to use.

src/shims/unix/fs.rs
```diff
@@ -49,6 +49,48 @@ impl FileDescription for FileHandle {
...
+    ) -> InterpResult<'tcx, io::Result<usize>> {
+        assert!(communicate_allowed, "isolation should have prevented even opening a file");
+        // Emulates pread using seek + read + seek to restore cursor position.
+        // Correctness of this emulation relies on sequential nature of Miri execution.
+        let mut f = || {
+            let cursor_pos = self.file.stream_position()?;
+            self.file.seek(SeekFrom::Start(offset))?;
+            let res = self.file.read(bytes);
+            // Attempt to restore cursor position even if read has failed
+            self.file.seek(SeekFrom::Start(cursor_pos))?;
```
RalfJung, line 67:
```suggestion
            self.file.seek(SeekFrom::Start(cursor_pos)).expect("failed to restore file position, this shouldn't be possible");
```
same for `pwrite`

src/shims/unix/fs.rs
```diff
@@ -49,6 +49,48 @@ impl FileDescription for FileHandle {
...
+            res
+        };
+        Ok(f())
```
RalfJung, line 70: Why is this wrapped in a closure?

tests/pass-dep/tempfile.rs
```diff
@@ -9,6 +9,7 @@ mod utils;
 fn main() {
     test_tempfile();
     test_tempfile_in();
+    test_tempfile_ext().unwrap();
```
RalfJung, line 12: This will make it quite hard to figure out which operation failed, won't it?

tests/pass-dep/tempfile.rs
```diff
@@ -19,3 +20,24 @@ fn test_tempfile_in() {
...
+    use std::io::Write;
+    use std::os::unix::fs::FileExt;
+
+    let mut f = tempfile::tempfile()?;
+    f.write_all(b"hello world")?;
+    let mut buf = [0u8; 3];
+    f.read_exact_at(&mut buf, 2)?;
+    assert_eq!(&buf, b"llo");
+    f.read_exact_at(&mut buf, 6)?;
+    assert_eq!(&buf, b"wor");
```
RalfJung, line 34: Can you add some normal `read`s to ensure that read_exact_at leaves the file offset unaffected?

tests/pass-dep/tempfile.rs
```diff
@@ -19,3 +20,24 @@ fn test_tempfile_in() {
     let dir_path = utils::tmp();
     tempfile::tempfile_in(dir_path).unwrap();
 }
+
+fn test_tempfile_ext() -> std::io::Result<()> {
```
RalfJung, line 24: Why is the only test for this here? The point of this file is to test the `tempfile` crate.

Please add a test in tests/pass/shims/fs.rs. I don't think it makes sense to also have the test here -- is there anything tempfile-specific going on?

RalfJung (commented):
Thanks for the PR! I still have some comments/questions.

src/shims/unix/fs.rs
newpavlov: To allow use of `?`. Ideally, I would use a `try` block here.

tests/pass-dep/tempfile.rs
newpavlov: I didn't know about `utils::prepare`. I will move the test to `fs.rs` a bit later.

src/shims/unix/fs.rs
RalfJung: Ah, those `?` are for `io::Result`, not `InterpResult`. That's worth a comment.

RalfJung:
Looks good, thanks. :)

@bors r+

newpavlov:
@RalfJung 
Hi! Do you know when approximately we can expect for these changes to land on the Nightly toolchain? I tried to run tests on `miri 0.1.0 (7c2012d 2024-07-26)`, but they still fail with:
>unsupported operation: can't call foreign function `pread64` on OS `linux`

RalfJung:
I usually do a sync every weekend but this weekend I had other things to do.

PR #3744, opened 2024-07-12
Add `O_NOFOLLOW` flag support

newpavlov:


src/shims/unix/fs.rs
```diff
@@ -366,6 +366,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 return Ok(-1);
             }
         }
+
+        #[cfg(unix)]
```
saethlin, line 370: This means that Miri running on Linux will have different support for `x86_64-unknown-linux-gnu` than Miri running on Windows. I do not think we want that. Can we support this on all host platforms?

newpavlov: I am not very familiar with Windows, so I am not sure how to properly emulate it and whether it's possible in the first place. Maybe we could simply do nothing on Windows for this flag?

saethlin: You don't need to know Windows, all you need to do is figure out how to implement this flag with portable standard library APIs, without passing through to the host. Based on the documentation of `O_NOFOLLOW` this seems quite possible.

newpavlov: You mean by adding a default impl which uses [`Path::is_symlink`](https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.is_symlink)?

newpavlov: I decided to add a warning that `O_NOFOLLOW` is ignored on non-UNIX targets. There is already a bunch of code gated on `unix` in this module, so it's not like this is a sole exception. A `Path`-based emulation can be added in a later PR, I can create an issue for it later.

src/shims/unix/fs.rs
```diff
@@ -366,14 +367,33 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 return Ok(-1);
             }
         }
+
+        let o_nofollow = this.eval_libc_i32("O_NOFOLLOW");
+        if flag & o_nofollow == o_nofollow {
+            #[cfg(unix)]
+            {
+                use std::os::unix::fs::OpenOptionsExt;
+                options.custom_flags(o_nofollow);
```
RalfJung, line 376: This uses the *target* value for O_NOFOLLOW to set the flag on the host. Unless all unixes have the same value here, that's wrong -- `custom_flags` should probably use `libc::O_NOFOLLOW` or something like that.

src/shims/unix/fs.rs
```diff
@@ -366,14 +367,33 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            }
+            // Strictly speaking, this emulation is not equivalent to the O_NOFOLLOW flag behavior,
+            // but it's good enough for Miri purposes.
+            #[cfg(not(unix))]
+            {
+                if path.is_symlink() {
```
RalfJung, line 382: What will this do with a path like `/path/symlink-to-dir/file-in-dir`? AFAIK `is_symlink` only tests the top-level path, right? So this would have to check all prefixes of this path as well.

RalfJung (commented):
Thanks for the PR!

Please also add a test (pass-dep/libc/libc-fs.rs seems like the right place).

src/shims/unix/fs.rs
RalfJung: Oh, reading the docs I learned O_NOFOLLOW also just checks the last component. Then yeah this emulation seems right.

src/shims/unix/fs.rs
```diff
@@ -366,14 +367,33 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        let o_nofollow = this.eval_libc_i32("O_NOFOLLOW");
+        if flag & o_nofollow == o_nofollow {
+            #[cfg(unix)]
+            {
+                use std::os::unix::fs::OpenOptionsExt;
+                options.custom_flags(o_nofollow);
+            }
+            // Strictly speaking, this emulation is not equivalent to the O_NOFOLLOW flag behavior,
+            // but it's good enough for Miri purposes.
```
RalfJung, line 379:
```suggestion
            // Strictly speaking, this emulation is not equivalent to the O_NOFOLLOW flag behavior:
            // the path could change between us checking it here and the later call to `open`.
            // But it's good enough for Miri purposes.
```

src/shims/unix/fs.rs
RalfJung, line 382:
```suggestion
                // O_NOFOLLOW only fails when the trailing component is a symlink;
                // the entire rest of the path can still contain symlinks.
                if path.is_symlink() {
```

RalfJung:
> Please also add a test (pass-dep/libc/libc-fs.rs seems like the right place).

Specifically, that file can have a test where the file is *not* a symlink and we just check that the flag is supported.

The test where the file *is* a symlink should go into tests/pass-dep/libc/libc-fs-readlink.rs (which we should probably rename to `libc-fs-symlinks`); that file is skipped on Windows hosts since on Windows we can't create symlinks when we're not admins...

src/shims/unix/fs.rs
newpavlov: Ah, indeed. Fixed.

RalfJung:
Looks good, thanks. :)
@bors r+

PR #3745, opened 2024-07-13
Implement the `os_unfair_lock` functions on macOS

joboet:
These are needed for rust-lang/rust#122408. See the documentation [here](https://developer.apple.com/documentation/os/synchronization?language=objc) and the implementation [here](https://github.com/apple-oss-distributions/libplatform/blob/a00a4cc36da2110578bcf3b8eeeeb93dcc7f4e11/src/os/lock.c#L645).

### DIFF
--- a/src/concurrency/sync.rs
+++ b/src/concurrency/sync.rs
@@ -269,7 +269,7 @@ pub(super) trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let this = self.eval_context_mut();
         if this.mutex_is_locked(mutex) {
             assert_ne!(this.mutex_get_owner(mutex), this.active_thread());
-            this.mutex_enqueue_and_block(mutex, retval, dest);
+            this.mutex_enqueue_and_block(mutex, Some((retval, dest)));
         } else {
             // We can have it right now!
             this.mutex_lock(mutex);
@@ -390,9 +390,15 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     }
 
     /// Put the thread into the queue waiting for the mutex.
-    /// Once the Mutex becomes available, `retval` will be written to `dest`.
+    ///
+    /// Once the Mutex becomes available and if it exists, `retval_dest.0` will
+    /// be written to `retval_dest.1`.
     #[inline]
-    fn mutex_enqueue_and_block(&mut self, id: MutexId, retval: Scalar, dest: MPlaceTy<'tcx>) {
+    fn mutex_enqueue_and_block(
+        &mut self,
+        id: MutexId,
+        retval_dest: Option<(Scalar, MPlaceTy<'tcx>)>,
+    ) {
         let this = self.eval_context_mut();
         assert!(this.mutex_is_locked(id), "queing on unlocked mutex");
         let thread = this.active_thread();
@@ -403,13 +409,16 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             callback!(
                 @capture<'tcx> {
                     id: MutexId,
-                    retval: Scalar,
-                    dest: MPlaceTy<'tcx>,
+                    retval_dest: Option<(Scalar, MPlaceTy<'tcx>)>,
                 }
                 @unblock = |this| {
                     assert!(!this.mutex_is_locked(id));
                     this.mutex_lock(id);
-                    this.write_scalar(retval, &dest)?;
+
+                    if let Some((retval, dest)) = retval_dest {
+                        this.write_scalar(retval, &dest)?;
+                    }
+
                     Ok(())
                 }
             ),
--- a/src/provenance_gc.rs
+++ b/src/provenance_gc.rs
@@ -30,6 +30,17 @@ impl<T: VisitProvenance> VisitProvenance for Option<T> {
     }
 }
 
+impl<A, B> VisitProvenance for (A, B)
+where
+    A: VisitProvenance,
+    B: VisitProvenance,
+{
+    fn visit_provenance(&self, visit: &mut VisitWith<'_>) {
+        self.0.visit_provenance(visit);
+        self.1.visit_provenance(visit);
+    }
+}
+
 impl<T: VisitProvenance> VisitProvenance for std::cell::RefCell<T> {
     fn visit_provenance(&self, visit: &mut VisitWith<'_>) {
         self.borrow().visit_provenance(visit)
--- a/src/shims/unix/macos/foreign_items.rs
+++ b/src/shims/unix/macos/foreign_items.rs
@@ -1,6 +1,7 @@
 use rustc_span::Symbol;
 use rustc_target::spec::abi::Abi;
 
+use super::sync::EvalContextExt as _;
 use crate::shims::unix::*;
 use crate::*;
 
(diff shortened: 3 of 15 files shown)


src/concurrency/sync.rs
```diff
@@ -392,7 +392,12 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     /// Put the thread into the queue waiting for the mutex.
     /// Once the Mutex becomes available, `retval` will be written to `dest`.
```
RalfJung, line 393: Seems like the comment needs updating to explain what happens when `retval` is `None`. Seems like `dest` is entirely ignored then? Maybe make them optional together in that case?

src/shims/unix/macos/foreign_items.rs
```diff
@@ -174,6 +174,70 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    if this.mutex_get_owner(id) == this.active_thread() {
+                        throw_machine_stop!(TerminationInfo::Abort(
+                            "attempted to lock an os_unfair_lock that is already locked by the current thread".to_owned()
+                        ));
+                    }
+
+                    this.mutex_enqueue_and_block(id, None, dest.clone());
+                } else {
+                    this.mutex_lock(id);
+                }
```
RalfJung, line 192: Please create a file `macos/sync.rs` and move these implementations there. `foreign_items` should then do the `check_shim` and call the logic in the other file.

src/shims/unix/macos/foreign_items.rs
```diff
@@ -174,6 +174,70 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             }
 
+            "os_unfair_lock_lock" => {
+                let [lock_op] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let id =
+                    this.mutex_get_or_create_id(lock_op, this.libc_ty_layout("os_unfair_lock"), 0)?;
+
+                if this.mutex_is_locked(id) {
+                    if this.mutex_get_owner(id) == this.active_thread() {
+                        throw_machine_stop!(TerminationInfo::Abort(
```
RalfJung, line 184:
```suggestion
                        // Matching the current macOS implementation: abort on reentrant locking.
                        throw_machine_stop!(TerminationInfo::Abort(
```

src/shims/unix/macos/foreign_items.rs
```diff
@@ -174,6 +174,70 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    this.mutex_lock(id);
+                }
+            }
+            "os_unfair_lock_trylock" => {
+                let [lock_op] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let id =
+                    this.mutex_get_or_create_id(lock_op, this.libc_ty_layout("os_unfair_lock"), 0)?;
+
+                if this.mutex_is_locked(id) {
+                    this.write_scalar(Scalar::from_bool(false), dest)?;
```
RalfJung, line 200: So this does *not* detect reentrancy?

tests/pass-dep/concurrency/apple-os-unfair-lock.rs
```diff
@@ -0,0 +1,21 @@
...
+    unsafe {
+        libc::os_unfair_lock_lock(lock.get());
+        libc::os_unfair_lock_assert_owner(lock.get());
+        assert!(!libc::os_unfair_lock_trylock(lock.get()));
+        libc::os_unfair_lock_unlock(lock.get());
+
+        libc::os_unfair_lock_assert_not_owner(lock.get());
+    }
+
+    // `os_unfair_lock`s can be moved and leaked:
```
RalfJung, line 17: What if they get moved while being locked?

RalfJung (commented):
Thanks, this looks quite good! See the comments below.

> See the documentation [here](https://developer.apple.com/documentation/os/synchronization?language=objc) and the implementation [here](https://github.com/apple-oss-distributions/libplatform/blob/a00a4cc36da2110578bcf3b8eeeeb93dcc7f4e11/src/os/lock.c#L645).

Please add links to this in the code as well.

src/concurrency/sync.rs
joboet: Done.

src/shims/unix/macos/foreign_items.rs
joboet: Sure!

src/shims/unix/macos/foreign_items.rs
joboet: No, it doesn't, it simply returns `false` in all cases where the lock couldn't be acquired.

tests/pass-dep/concurrency/apple-os-unfair-lock.rs
joboet: Nothing, the lock continues to function. We can in all likelihood rely on this, because the lock is unlocked in userspace before entering the kernel, which means the kernel cannot rely on the lock being present when doing the wake, as another thread might lock, unlock and deallocate it immediately. And as the userspace code cannot rely on the lock being fixed either – the structure is too small to contain a pointer – we can do all the moving and leaking we like.

src/provenance_gc.rs
```diff
@@ -162,6 +162,17 @@ impl VisitProvenance for crate::MiriInterpCx<'_> {
     }
 }
 
+impl<A, B> VisitProvenance for (A, B)
+where
```
RalfJung, line 166: Please move this up to `Option<T>` so that we keep the impls for non-Miri types together.

tests/pass-dep/concurrency/apple-os-unfair-lock.rs
```diff
@@ -0,0 +1,23 @@
...
+        libc::os_unfair_lock_lock(lock.get());
+        libc::os_unfair_lock_assert_owner(lock.get());
+        assert!(!libc::os_unfair_lock_trylock(lock.get()));
+        libc::os_unfair_lock_unlock(lock.get());
+
+        libc::os_unfair_lock_assert_not_owner(lock.get());
+    }
+
+    // `os_unfair_lock`s can be moved and leaked, even when locked,
+    // as long as no operation is occurring on the lock:
```
RalfJung, line 18: No operation occurs -- where/when? Concurrently with the move? The move is just copying some bytes around, so not sure what that would even mean. Or is it about having concurrent operations on different "copies" of the same lock? Miri would currently accept that without complaining, right?

(Miri also accepts pthread locks being moved; we don't currently attempt to fully model all the UB here.)

joboet: I meant concurrently: you shouldn't invalidate the pointer to the lock while it is being used in `os_unfair_*` in another thread.

RalfJung: But if I make a copy of the lock value while it is locked, and then call `unlock` on both the old and new value -- that can't be well-defined, can it?

joboet: Funnily enough, that actually works with the current macOS implementation, but fails in miri. I guess I could emulate that 🤷. The way `os_unfair_lock` works is that it uses the thread ID of the owning thread as a "locked"-value of an atomic variable, and then calls the macOS equivalent of futex (called `ulock`) to handle blocking and priority inheritance in the contended case. The [Fuchsia mutex](https://github.com/rust-lang/rust/blob/25acbbd12f87ed9e3eef13c095083c4693da2e8f/library/std/src/sys/sync/mutex/fuchsia.rs) in `std` (itself derived from Fuchsia's one) works exactly the same way, just with `panic!`s instead of `SIGTRAP`s.

RalfJung: So... what would it do? Both unlocks succeed, and they "wake up" everyone waiting on the respective address?

No need to implement that in Miri, but having a note somewhere that we don't implement this fully correctly seems good.

RalfJung, line 18:
```suggestion
    // `os_unfair_lock`s can be moved and leaked.
    // In the real implementation, even moving it while locked is possible
    // (and "forks" the lock, i.e. old and new location have independent wait queues);
    // Miri behavior differs here and anyway none of this is documented.
```

RalfJung:
Looks good, thanks a lot! I also filed https://github.com/rust-lang/miri/issues/3749 to track adjusting our shared lock implementations so that they can be address-sensitive the way they should be.

Please squash the commits.

joboet:
> I also filed #3749 to track adjusting our shared lock implementations so that they can be address-sensitive the way they should be.

I could change this to use the futex API like the real implementation, which would make this behave exactly the same way.

RalfJung:
@bors r+

> I could change this to use the futex API like the real implementation, which would make this behave exactly the same way.

That seems a bit awkward to implement inside Miri... also, we'd want to raise UB when doing things that are not guaranteed by the implementation, so having exactly the same implementation does not seem quite right.

Issue #3691, opened 2024-06-20
Remove dead Intel intrinsics


RalfJung:
Since https://github.com/rust-lang/stdarch/pull/1584 there are some LLVM intel intrinsics that stdarch does not use any more, making them dead code in Miri. Might be a good idea to clean that up. :)

(I also wonder whether there's some good way to check, in the future, that we don't have other dead untested vendor intrinsics... maybe something that checks line coverage or so?)

PR #3747, opened 2024-07-13
remove some SSE/SSE2 intrinsics that are no longer used by stdarch

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/3691

### DIFF
--- a/src/shims/x86/avx.rs
+++ b/src/shims/x86/avx.rs
@@ -73,13 +73,12 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
                 round_all::<rustc_apfloat::ieee::Double>(this, op, rounding, dest)?;
             }
-            // Used to implement _mm256_{sqrt,rcp,rsqrt}_ps functions.
+            // Used to implement _mm256_{rcp,rsqrt}_ps functions.
             // Performs the operations on all components of `op`.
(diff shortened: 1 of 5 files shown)


RalfJung:
Blocked on https://github.com/rust-lang/stdarch/pull/1584 propagating to the rustc repo.

TDecking:
The stdarch patch will also ensure that the "sqrt.ps.256" case in https://github.com/rust-lang/miri/blob/master/src/shims/x86/avx.rs#L78 will never be reached. This case is still present here.

RalfJung:
Yeah, but sqrt is implemented together with two other functions there so there's not really any code saved by removing that... :shrug: 

The same goes for sse.sqrt.ss, sse.sqrt.ps.

RalfJung:
Hm I guess this removes all uses of `FloatUnaryOp::Sqrt`. So maybe it's worth it after all.

TDecking:
@RalfJung https://github.com/rust-lang/rust/pull/128466 got merged. This PR is no longer blocked.

RalfJung:
Ah, good point, thanks! :)

@bors r+

RalfJung:
Ah, maybe we need the rustup to land first.

RalfJung:
Hm, no... we had a rustup yesterday, so https://github.com/rust-lang/rust/pull/128466 should have been picked up.

RalfJung:
D'oh, this is from a Miri test directly importing the intrinsic.^^

PR #3752, opened 2024-07-17
Implement SHA256 SIMD intrinsics on x86

Kixunil:
Disclaimer: this is my first contribution to `miri`'s code. It's quite possible I'm missing something. This code works but may not be the cleanest/best possible.

It'd be useful to be able to verify code implementing SHA256 using SIMD since such code is a bit more complicated and at some points requires use of pointers. Until now `miri` didn't support x86 SHA256 intrinsics. This commit implements them.

### DIFF
--- a/src/shims/x86/sha.rs
+++ b/src/shims/x86/sha.rs
@@ -1,0 +1,209 @@
+use rustc_span::Symbol;
+use rustc_target::spec::abi::Abi;
+
+use crate::*;
+
+impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
+pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    fn emulate_x86_sha_intrinsic(
+        &mut self,
+        link_name: Symbol,
+        abi: Abi,
+        args: &[OpTy<'tcx>],
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx, EmulateItemResult> {
+        let this = self.eval_context_mut();
+        this.expect_target_feature_for_intrinsic(link_name, "sha")?;
+        // Prefix should have already been checked.
+        let unprefixed_name = link_name.as_str().strip_prefix("llvm.x86.sha").unwrap();
+
+        match unprefixed_name {
+            // Used to implement the _mm_sha256rnds2_epu32 function.
+            "256rnds2" => {
+                let [a, b, k] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let (a_reg, a_len) = this.operand_to_simd(a)?;
+                let (b_reg, b_len) = this.operand_to_simd(b)?;
+                let (k_reg, k_len) = this.operand_to_simd(k)?;
+                let (dest, dest_len) = this.mplace_to_simd(dest)?;
+
+                assert_eq!(a_len, 4);
+                assert_eq!(b_len, 4);
+                assert_eq!(k_len, 4);
+                assert_eq!(dest_len, 4);
+
+                let mut a = [0; 4];
+                for (i, dst) in a.iter_mut().enumerate() {
+                    *dst = this.read_scalar(&this.project_index(&a_reg, i as u64)?)?.to_u32()?
+                }
+                let mut b = [0; 4];
+                for (i, dst) in b.iter_mut().enumerate() {
+                    *dst = this.read_scalar(&this.project_index(&b_reg, i as u64)?)?.to_u32()?
+                }
+                let mut k = [0; 4];
+                for (i, dst) in k.iter_mut().enumerate() {
+                    *dst = this.read_scalar(&this.project_index(&k_reg, i as u64)?)?.to_u32()?
+                }
+
+                let result = sha256rnds2_epu32(a, b, k);
+                for (i, part) in result.into_iter().enumerate() {
+                    this.write_scalar(Scalar::from_u32(part), &this.project_index(&dest, i as u64)?)?;
+                }
+            }
+            // Used to implement the _mm_sha256msg1_epu32 function.
+            "256msg1" => {
+                let [a, b] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let (a_reg, a_len) = this.operand_to_simd(a)?;
+                let (b_reg, b_len) = this.operand_to_simd(b)?;
+                let (dest, dest_len) = this.mplace_to_simd(dest)?;
+
+                assert_eq!(a_len, 4);
+                assert_eq!(b_len, 4);
+                assert_eq!(dest_len, 4);
+
+                let mut a = [0; 4];
+                for (i, dst) in a.iter_mut().enumerate() {
+                    *dst = this.read_scalar(&this.project_index(&a_reg, i as u64)?)?.to_u32()?
+                }
+                // least significant part of b
+                let b_ls = this.read_scalar(&this.project_index(&b_reg, 0)?)?.to_u32()?;
+
+                let result = sha256msg1_epu32(a, b_ls);
+                for (i, part) in result.into_iter().enumerate() {
+                    this.write_scalar(Scalar::from_u32(part), &this.project_index(&dest, i as u64)?)?;
+                }
+            }
+            // Used to implement the _mm_sha256msg2_epu32 function.
+            "256msg2" => {
+                let [a, b] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                let (a_reg, a_len) = this.operand_to_simd(a)?;
+                let (b_reg, b_len) = this.operand_to_simd(b)?;
+                let (dest, dest_len) = this.mplace_to_simd(dest)?;
+
+                assert_eq!(a_len, 4);
+                assert_eq!(b_len, 4);
+                assert_eq!(dest_len, 4);
+
+                let mut a = [0; 4];
+                for (i, dst) in a.iter_mut().enumerate() {
+                    *dst = this.read_scalar(&this.project_index(&a_reg, i as u64)?)?.to_u32()?
+                }
+                // most significant part of b
+                let mut b_ms = [0; 2];
+                for (i, dst) in b_ms.iter_mut().enumerate() {
+                    *dst = this.read_scalar(&this.project_index(&b_reg, i as u64 + 2)?)?.to_u32()?
+                }
+
+                let result = sha256msg2_epu32(a, b_ms);
+                for (i, part) in result.into_iter().enumerate() {
+                    this.write_scalar(Scalar::from_u32(part), &this.project_index(&dest, i as u64)?)?;
+                }
+            }
+            _ => return Ok(EmulateItemResult::NotSupported),
+        }
+        Ok(EmulateItemResult::NeedsReturn)
+    }
+}
+
+#[allow(non_snake_case)]
+fn sha256rnds2_epu32(a: [u32; 4], b: [u32; 4], k: [u32; 4]) -> [u32; 4] {
+    // Translated from the Intel's documentation by Chat GPT.
+    // It works, don't ask why it looks horrible.
+    let mut A: [u32; 3] = [0; 3];
+    let mut B: [u32; 3] = [0; 3];
+    let mut C: [u32; 3] = [0; 3];
+    let mut D: [u32; 3] = [0; 3];
+    let mut E: [u32; 3] = [0; 3];
+    let mut F: [u32; 3] = [0; 3];
+    let mut G: [u32; 3] = [0; 3];
+    let mut H: [u32; 3] = [0; 3];
+    let mut W_K: [u32; 2] = [0; 2];
+
+    A[0] = b[3];
+    B[0] = b[2];
+    C[0] = a[3];
+    D[0] = a[2];
+    E[0] = b[1];
+    F[0] = b[0];
+    G[0] = a[1];
+    H[0] = a[0];
+
+    W_K[0] = k[0];
+    W_K[1] = k[1];
+
+    for i in 0..2 {
+        A[i + 1] = ch(E[i], F[i], G[i])
+            .wrapping_add(sum1(E[i]))
+            .wrapping_add(W_K[i])
+            .wrapping_add(H[i])
+            .wrapping_add(maj(A[i], B[i], C[i]))
+            .wrapping_add(sum0(A[i]));
+
+        B[i + 1] = A[i];
+        C[i + 1] = B[i];
+        D[i + 1] = C[i];
+
+        E[i + 1] = ch(E[i], F[i], G[i])
+            .wrapping_add(sum1(E[i]))
+            .wrapping_add(W_K[i])
+            .wrapping_add(H[i])
+            .wrapping_add(D[i]);
+
+        F[i + 1] = E[i];
+        G[i + 1] = F[i];
+        H[i + 1] = G[i];
+    }
+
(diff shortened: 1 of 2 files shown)


Kixunil:
ow do I properly convey that I'm sure side effects don't happen? Just `allow` the lint?

RalfJung:
Thanks for the PR! There are some other PRs we have to review before we get to this one, so unfortunately it may take a bit -- sorry for that.

From a quick glance, it seems like this PR actually implements sha256 in Miri. Ideally I'd like to avoid being responsible for cryptographic code. Is there a small crate, akin to `aes`, that has a suitable license for us to be able to use it here as a sha256 implementation?

If we do carry a sha256 implementation, I would prefer one that was not written by ChatGPT, but by a human that actually understands sha256. Maybe there's a larger Rust crypto crate that we can copy an implementation from.

apoelstra:
@RalfJung there is the [bitcoin_hashes](https://crates.io/crates/bitcoin_hashes) crate which you're welcome to steal code from (it is CC0 licensed) which is written by professional human cryptographers. You probably don't want to depend on it because it does a bunch of other stuff (and you'll get complaints about having `bitcoin` in a dependency name).

There is also the [sha2](https://crates.io/crates/sha2) crate written by a different set of professional human cryptographers, which is MIT/Apache licensed. And you could use it as a direct dependency (there is a good chance it is already in your dep tree somewhere) since it's narrowly focused and maintained. Though again, it does more than SHA256 which is really the only thing you want.

apoelstra:
I'll also add, that while your aversion to carrying/maintaining crypto code is well-placed, hashes are a low-burden thing to maintain for a few reasons. By their nature you don't need to worry about constant-timeness so you can write them once and forget about the code forever (especially in something like Miri where you aren't going to want to go microoptimizing this). And also by their nature it's super hard to have subtle correctness bugs; pretty-much any bug will result in every single output being visibly totally wrong. Finally hash functions are pretty small self-contained pure algorithms.

RalfJung:
Thanks, those are helpful comments!

Kixunil:
> There are some other PRs we have to review before we get to this one, so unfortunately it may take a bit -- sorry for that.

No problem! There's no rush, we can run my fork for now.

> it seems like this PR actually implements sha256 in Miri

Not really, just the three primitives that need to be used together with a bunch of other code to actually implement sha256. I also suspect it's not trivial to just pull out these from existing impls - at least I tried and the code looked so different that I couldn't do it. That's why I gave up and translated the Intel's psueudocode which literally describe how the specific instructions work.

> Ideally I'd like to avoid being responsible for cryptographic code.

Isn't `miri` only supposed to be used for testing (or rather sanitizing) the code? I don't believe anyone would ever run this in production where it would actually matter. The SHA256 tests that take under a second to run in *debug* mode run several minutes under `miri`. Further we can add a big fat warning about this.

Anyway, everything @apoelstra said is true. The code is constant-time and any single tiny mistake would just completely break it - as I experienced this first hand. It also happens to make debugging the code incredibly hard. (I had to put `dbg!` statements both into `miri` and into the tested code to figure out what was wrong.)

RalfJung:
> Not really, just the three primitives that need to be used together with a bunch of other code to actually implement sha256. I also suspect it's not trivial to just pull out these from existing impls - at least I tried and the code looked so different that I couldn't do it. That's why I gave up and translated the Intel's psueudocode which literally describe how the specific instructions work.

So the sha2 crate wouldn't be useful? That seems like the closest thing.
Cc @newpavlov -- could I ask for your help with this PR?

> Isn't miri only supposed to be used for testing (or rather sanitizing) the code? I don't believe anyone would ever run this in production where it would actually matter. The SHA256 tests that take under a second to run in debug mode run several minutes under miri. Further we can add a big fat warning about this.

Yes, but that doesn't change the fact that Miri has a small set of maintainers, and we can't just ship code we all have no clue about -- that's not living up to our responsibility as maintainers. It's not about performance, it's about the fact that *I can't check whether your code makes any sense* and I don't have the time to become an expert in every problem domain that is covered by an Intel intrinsic.

"Code was translated by Chat GPT" also raises some very big alarm bells, since it seems to indicate that you also didn't check whether the code makes any sense. At the very least I would expect a link to the Intel docs, and a statement by you as PR author that you are sure that the Chat GPT translation is correct. Someone from the Miri team will have to double-check this before we can land this (if we can't find an alternative to carrying this code ourselves), but it's not reasonable to ask us to be the first humans to check this.

And of course the PR needs to come with some tests.

src/shims/x86/sha.rs
```diff
@@ -0,0 +1,209 @@
...
+                let (dest, dest_len) = this.mplace_to_simd(dest)?;
+
+                assert_eq!(a_len, 4);
+                assert_eq!(b_len, 4);
+                assert_eq!(dest_len, 4);
+
+                let mut a = [0; 4];
+                for (i, dst) in a.iter_mut().enumerate() {
+                    *dst = this.read_scalar(&this.project_index(&a_reg, i as u64)?)?.to_u32()?
+                }
```
RalfJung, line 95: I'm seeing many instances of this same loop; seems worth putting that into a helper method?

src/shims/x86/sha.rs
```diff
@@ -0,0 +1,209 @@
...
+                // most significant part of b
```
RalfJung, line 96:
```suggestion
                // b_ms becomes the most significant part of b, i.e., b[2..4].
```
I first thought this would be the most significant half of every element or so. It's just the second half of the array, quite confusing to call that "most significant".

newpavlov:
We probably could handle emulation of intrinsics on the RustCrypto side. For example, we have a similar request for AES: https://github.com/RustCrypto/block-ciphers/issues/389

But, personally, I would prefer to do it in separate crates, since having emulation of SHA2 hardware intrinsics in the `sha2` crate is not worthwhile in my opinion.

@tarcieri WDYT?

Kixunil:
> So the sha2 crate wouldn't be useful?

Looking at its code it does actually have the same primitives inside so copying them would be arguably better but they still need a bit of wiring code - the same there is at the beginning that just shuffles values in arrays.

> _I can't check whether your code makes any sense_ and I don't have the time to become an expert in every problem domain that is covered by an Intel intrinsic.

OK, fair. How can we solve it then? The code should exactly (bug-for-bug) emulate the Intel's intrinsic and the intrinsic has documented pseudo code implementation should just comparing the implementation be enough? Or is there anything else I can do?

> "Code was translated by Chat GPT" also raises some very big alarm bells, since it seems to indicate that you also didn't check whether the code makes any sense.

I did test it by running `bitcoin_hashes` test suite in `miri` which passed perfectly and due to the property of hashes that any tiny mistake causes all the tests to reliably fail I'm confident it has the same behavior as the real instructions. The only thing I didn't yet check is if there are performance optimizations left. I thought this is not urgent.

If this is not sufficient please suggest specific steps I could do and I'll do it.

> but it's not reasonable to ask us to be the first humans to check this.

That's absolutely something I wouldn't make you do, I'm sorry if it looked like I did. That's what I meant by "this code works". The things I'm unsure about is the `miri` part - does the way I do loads and stores make sense? I have no clue about `miri` internals, I just looked how other intrinsics do it and copied it. This is something I need guidance in.

> And of course the PR needs to come with some tests.

Of course, I first wanted to see if my approach makes sense so that I wouldn't need to refactor the test (running SHA256 is the ultimate test). I'm also not sure what kind of testing strategy would be most appropriate. I kinda wish to just run the real instructions on real hardware with random inputs and compare the results but I don't know if this is a viable solution. Or just bake in the real SHA256 and run it on a few inputs. Of course I can also grab a bunch of random vectors and bake them in.

src/shims/x86/sha.rs
Kixunil: Yeah, the whole thing is quite confusing because the registers are little endian. I guess I should mention it in the code.

src/shims/x86/sha.rs
Kixunil: Yes, I was thinking about it, just wanted to know if what I'm doing here makes sense from `miri` perspective before I put in any more work that might be thrown away.

RalfJung:
> Looking at its code it does actually have the same primitives inside so copying them would be arguably better but they still need a bit of wiring code - the same there is at the beginning that just shuffles values in arrays.

Sounds like a good idea.

> OK, fair. How can we solve it then? The code should exactly (bug-for-bug) emulate the Intel's intrinsic and the intrinsic has documented pseudo code implementation should just comparing the implementation be enough? Or is there anything else I can do?

Manually comparing Intel's pseudocode with the Miri code would be a way to get additional confidence on top of the testing.

> The things I'm unsure about is the miri part - does the way I do loads and stores make sense? I have no clue about miri internals, I just looked how other intrinsics do it and copied it. This is something I need guidance in.

We don't usually load things into arrays on the host side, we just process them element-by-element, but here that doesn't seem to make sense so what you did looks reasonable.

> Of course, I first wanted to see if my approach makes sense so that I wouldn't need to refactor the test (running SHA256 is the ultimate test). I'm also not sure what kind of testing strategy would be most appropriate. I kinda wish to just run the real instructions on real hardware with random inputs and compare the results but I don't know if this is a viable solution. Or just bake in the real SHA256 and run it on a few inputs. Of course I can also grab a bunch of random vectors and bake them in.

We should definitely have some test vectors directly checking the intrinsic, yes. If there are any corner cases, they should be covered here.

How big is a (not performance-oriented) sha256 implementation on top of these intrinsics? If that's small enough to put in a test, then we can also have that in addition.

Kixunil:
OK, I'll look into it later, I have currently some urgent things to do.

RalfJung:
@rustbot author
based on the last comment

Regarding the decision about how to implement these intrinsics; @rust-lang/miri see the discussion above -- what do you think?

Kixunil:
@rustbot ready

Funny how you've looked at it at around the same time I did. :)

I've pushed an updated version that copies stuff from RustCrypto. I have addressed all your review comments including adding tests. The tests are somewhat large so LMK if I should remove anything.

RalfJung:
Thanks!

Regarding the tests, it seems I should have been more clear -- what I meant is tests in `tests/pass` that invoke the intrinsics from a Rust program that is interpreted by Miri. We want to know that the intrinsic actually works, after all. :) Unit tests inside `src` are not really necessary.

Kixunil:
I felt like something is off... Anyway, it's moved now.

src/shims/x86/sha.rs
```diff
@@ -0,0 +1,219 @@
...
+                let result = sha256msg2_epu32(a, b);
+                write(this, &dest, result)?;
+            }
+            _ => return Ok(EmulateItemResult::NotSupported),
+        }
+        Ok(EmulateItemResult::NeedsReturn)
+    }
+}
+
+#[inline(always)]
```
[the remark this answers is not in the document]
Kixunil: Do you prefer link to github or docs.rs? I could also link the crate at crates.io but that would force people to manually search the `soft.rs` module.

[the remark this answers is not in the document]
RalfJung: A github link to the file (or relevant lines in the file) would be best IMO.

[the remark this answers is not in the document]
Kixunil: Done.

saethlin:
> @rust-lang/miri see the discussion above -- what do you think?

I am similarly opposed to checking in any code that comes from ChatGPT. In addition to the issues with its accuracy and our ability to read the code we are checking in, it seems like the project has no regard for the licensing of its training set.

As to reviewing the implementation, I also cannot vouch for its correctness. I might be able to, given some time to learn SHA256.

That being said, I think there's a pretty low chance that the implementation here is subtly broken if it passes a few tests. The implementation has no branches and I don't see any edge cases, so just a handful of test cases should provide full coverage.

Kixunil:
> I am similarly opposed to checking in any code that comes from ChatGPT.

This is no longer relevant, the implementation now copies `sha2` crate.

> As to reviewing the implementation, I also cannot vouch for its correctness.

Since it's a copy of `sha2` we can just rely on them.

> That being said, I think there's a pretty low chance that the implementation here is subtly broken if it passes a few tests.

It's pretty much impossible to write an incorrect sha256 since any mistake will produce a completely different output. It's the same property as for the input.

src/shims/x86/sha.rs
```diff
@@ -0,0 +1,221 @@
...
+        a[2].wrapping_add(b[2]),
+        a[3].wrapping_add(b[3]),
+    ]
+}
+
+fn sha256load(v2: [u32; 4], v3: [u32; 4]) -> [u32; 4] {
+    [v3[3], v2[0], v2[1], v2[2]]
+}
+
+fn sha256rnds2_epu32(cdgh: [u32; 4], abef: [u32; 4], wk: [u32; 4]) -> [u32; 4] {
```
RalfJung, line 149: This seems to be called `sha256_digest_round_x2` in `sha2`, can you follow their naming?

src/shims/x86/sha.rs
```diff
@@ -0,0 +1,221 @@
...
+        let t1 = or(shr(x, 7), shl(x, 25));
+        let t2 = or(shr(x, 18), shl(x, 14));
+        let t3 = shr(x, 3);
+        xor(xor(t1, t2), t3)
+    }
+
+    add(v0, sigma0x4(sha256load(v0, v1)))
+}
+
+fn sha256msg2_epu32(v4: [u32; 4], v3: [u32; 4]) -> [u32; 4] {
```
RalfJung, line 205: Please rename to `sha256msg2`.

src/shims/x86/sha.rs
```diff
@@ -0,0 +1,221 @@
...
+    let x1 =
+        big_sigma1!(e1).wrapping_add(bool3ary_202!(e1, f1, g1)).wrapping_add(wk1).wrapping_add(h1);
+    let y1 = big_sigma0!(a1).wrapping_add(bool3ary_232!(a1, b1, c1));
+    let (a2, b2, _, _, e2, f2, _, _) =
+        (x1.wrapping_add(y1), a1, b1, c1, x1.wrapping_add(d1), e1, f1, g1);
+
+    [a2, b2, e2, f2]
+}
+
+fn sha256msg1_epu32(v0: [u32; 4], v1: [u32; 4]) -> [u32; 4] {
```
RalfJung, line 192: Please rename to `sha256msg1`.

tests/pass/shims/x86/intrinsics-sha.rs
```diff
@@ -0,0 +1,270 @@
...
+    assert!(is_x86_feature_detected!("sha"));
+    assert!(is_x86_feature_detected!("sse2"));
+    assert!(is_x86_feature_detected!("ssse3"));
+    assert!(is_x86_feature_detected!("sse4.1"));
+
+    unsafe {
+        test_sha256rnds2();
+        test_sha256();
+        test_sha256msg1();
+        test_sha256msg2();
```
RalfJung, line 49:
```suggestion
        test_sha256rnds2();
        test_sha256msg1();
        test_sha256msg2();
        test_sha256();
```
Going from lower-level to higher-level

RalfJung (commented):
LGTM, I just got some nits. :)

src/shims/x86/sha.rs
Kixunil: Sure, I wanted to follow the naming of intrinsic but I don't really mind.

Kixunil:
Addressed the nits.

RalfJung:
Looks good, thanks. :)

@bors r+

Kixunil:
I can't find any information about release schedule will this be available in the next nightly or will I have to wait longer?

RalfJung:
It will be available once Miri get synced into the rustc repo. I usually do this each weekend, though this weekend I am traveling so it might happen later.

PR #3756, opened 2024-07-20
Add `gettid` support

### LINKED ISSUE #3730: Implement `gettid`
Miri supports `getpid` but not `gettid`. It is tempting to implement `gettid` as just returning the internal thread ID, but we have to be careful about this requirement:

> In a single-threaded process, the thread ID is equal to the process ID (PID, as returned by getpid(2)).

Mandragorian:
Add support for `gettid` in miri.

To ensure that the requirement that  `getpid() == gettdi()` for the main thread, we use the value returned by `getpid` and add to it the internal thread index.

Since `getpid` is only supported when isolation is disabled, and we want `gettid` to be used both in isolated and non-isolated executions, we modify `getpid` to return a hardcoded value (1000) when running in isolation mode.

Fixes #3730

### DIFF
--- a/src/shims/env.rs
+++ b/src/shims/env.rs
@@ -108,4 +108,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             EnvVars::Windows(vars) => vars.get(name),
         }
     }
+
+    fn get_pid(&self) -> u32 {
+        let this = self.eval_context_ref();
+        if this.machine.communicate() { std::process::id() } else { 1000 }
+    }
 }
--- a/src/shims/unix/env.rs
+++ b/src/shims/unix/env.rs
@@ -274,12 +274,23 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let this = self.eval_context_mut();
         this.assert_target_os_is_unix("getpid");
 
-        this.check_no_isolation("`getpid`")?;
-
         // The reason we need to do this wacky of a conversion is because
         // `libc::getpid` returns an i32, however, `std::process::id()` return an u32.
         // So we un-do the conversion that stdlib does and turn it back into an i32.
         #[allow(clippy::cast_possible_wrap)]
-        Ok(std::process::id() as i32)
+        Ok(this.get_pid() as i32)
+    }
+
+    fn linux_gettid(&mut self) -> InterpResult<'tcx, i32> {
+        let this = self.eval_context_ref();
+        this.assert_target_os("linux", "gettid");
+
+        let index = this.machine.threads.active_thread().to_u32();
+
+        // Compute a TID for this thread, ensuring that the main thread has PID == TID.
+        let tid = this.get_pid().strict_add(index);
+
+        #[allow(clippy::cast_possible_wrap)]
+        Ok(tid as i32)
     }
 }
--- a/tests/pass-dep/libc/gettid.rs
+++ b/tests/pass-dep/libc/gettid.rs
@@ -1,0 +1,22 @@
+//@only-target-linux
+//@revisions: with_isolation without_isolation
+//@[without_isolation] compile-flags: -Zmiri-disable-isolation
+
+use libc::{getpid, gettid};
+use std::thread;
+
+fn main() {
+    thread::spawn(|| {
+        // Test that in isolation mode a deterministic value will be returned.
+        // The value 1001 is not important, we only care that whatever the value
+        // is, won't change from execution to execution.
+        #[cfg(with_isolation)]
+        assert_eq!(unsafe { gettid() }, 1001);
+
+        assert_ne!(unsafe { gettid() }, unsafe { getpid() });
+    });
+
+    // Test that the thread ID of the main thread is the same as the process
+    // ID.
+    assert_eq!(unsafe { gettid() }, unsafe { getpid() });
+}
--- a/src/shims/unix/linux/foreign_items.rs
+++ b/src/shims/unix/linux/foreign_items.rs
@@ -94,6 +94,11 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 )?;
                 this.write_scalar(res, dest)?;
             }
+            "gettid" => {
+                let [] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                let result = this.linux_gettid()?;
+                this.write_scalar(Scalar::from_i32(result), dest)?;
+            }
 
             // Dynamically invoked syscalls
             "syscall" => {
--- a/src/shims/windows/env.rs
+++ b/src/shims/windows/env.rs
@@ -200,9 +200,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     fn GetCurrentProcessId(&mut self) -> InterpResult<'tcx, u32> {
         let this = self.eval_context_mut();
         this.assert_target_os("windows", "GetCurrentProcessId");
-        this.check_no_isolation("`GetCurrentProcessId`")?;
 
-        Ok(std::process::id())
+        Ok(this.get_pid())
     }
 
     #[allow(non_snake_case)]
--- a/tests/pass/getpid.rs
+++ b/tests/pass/getpid.rs
@@ -1,9 +1,20 @@
-//@compile-flags: -Zmiri-disable-isolation
+//@revisions: with_isolation without_isolation
+//@[without_isolation] compile-flags: -Zmiri-disable-isolation
 
 fn getpid() -> u32 {
     std::process::id()
 }
 
 fn main() {
-    getpid();
+    let pid = getpid();
+
+    std::thread::spawn(move || {
+        assert_eq!(getpid(), pid);
+    });
+
+    // Test that in isolation mode a deterministic value will be returned.
+    // The value 1000 is not important, we only care that whatever the value
+    // is, won't change from execution to execution.
+    #[cfg(with_isolation)]
+    assert_eq!(pid, 1000);
 }


Mandragorian:
Not sure yet why the tests are failing. When I run them locally with `./miri test` it succeeds. I think the `libc` version in `test-dependencies` has the `gettid` function. I will have time to investigate later today or tomorrow.

RalfJung:
The tests are failing on x86_64-apple-darwin, so the command for that would be `./miri test --target x86_64-apple-darwin`. it will work on any host, you don't need a macOS system for this.

tests/pass/shims/env/getpid_with_isolation.rs
```diff
@@ -0,0 +1,7 @@
+//@only-target-linux
```
saethlin, line 1: This test only uses std so it should work on all targets, no?

tests/pass-dep/libc/gettid.rs
```diff
@@ -0,0 +1,9 @@
+//@only-target-linux
+
+use libc::{getpid, gettid};
+use std::thread;
+
+fn main() {
+    thread::spawn(|| assert_eq!(unsafe { gettid() }, 1001));
```
saethlin, line 7: What is this testing? Or rather, why are we checking that gettid returns this specific value?

tests/pass/shims/env/getpid_with_isolation.rs
Mandragorian: I was about to write a comment explaining this (I changed the test to only skip windows now):

Additionally the `getpid_with_isolation` test was failing for window targets since my PR only changed the `getpid` function, but windows. So I skip the test for windows. However it would also make sense to make similar changes to support `GetCurrentProcessId` in isolation mode, but I don't know if that is wanted. Let me know.

tests/pass-dep/libc/gettid.rs
Mandragorian: I am also not sure if assertion provides much value. I wanted to test that `gettid` will always return the same value in isolation mode, since determinism is expected. I don't have strong feelings about it, I can remove it if you feel this isn't wanted.

tests/pass/shims/env/getpid_with_isolation.rs
saethlin: If it's a small change it would be rather nice to have this test run on all our tested targets :)

Mandragorian: done!

tests/pass-dep/libc/gettid.rs
saethlin: That's a fair idea, but I wouldn't mind a comment that explains that the only point is to ensure that we always return the same value, and that the specific value 1001 is not important per se.

Mandragorian: done!

Mandragorian: actually I added a similar comment in the `getpid_with_isolation`  test.

src/shims/env.rs
```diff
@@ -109,3 +109,10 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         }
     }
 }
+
+pub(crate) fn get_pid<'tcx>(ecx: &InterpCx<'tcx, MiriMachine<'tcx>>, item_name: &str) -> u32 {
```
RalfJung, line 113: Please don't use free functions when this can easily become a method on the extension trait.

src/shims/env.rs
```diff
@@ -109,3 +109,10 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    match ecx.check_no_isolation(item_name) {
```
RalfJung, line 114: We never ever match on `InterpResult`. Instead of this, just check `machine.communicate()` directly.

src/shims/unix/env.rs
```diff
@@ -274,12 +275,23 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        Ok(get_pid(this, "`getpid`") as i32)
+    }
+
+    fn gettid(&mut self) -> InterpResult<'tcx, i32> {
+        let this = self.eval_context_ref();
+        this.assert_target_os_is_unix("gettid");
+
+        let index = this.machine.threads.active_thread().to_u32();
+
+        #[allow(clippy::arithmetic_side_effects)]
```
RalfJung, line 291: This is suspicious, why did you allow the lint? It exists for a reason!

tests/pass-dep/libc/gettid.rs
```diff
@@ -0,0 +1,15 @@
+//@only-target-linux
```
RalfJung, line 1: You added this function for all Unix hosts, so just testing it on Linux is not sufficient.

`ignore-target-windows` is probably better.

tests/pass-dep/libc/gettid.rs
```diff
@@ -0,0 +1,15 @@
...
+
+fn main() {
+    // Test that in isolation mode a deterministic value will be returned.
+    // The value 1001 is not important, we only care that whatever the value
+    // is, won't change from execution to execution.
+    thread::spawn(|| assert_eq!(unsafe { gettid() }, 1001));
+
+    // Test that the thread ID of the main thread is the same as the process
+    // ID.
+    assert_eq!(unsafe { gettid() }, unsafe { getpid() });
```
RalfJung, line 14: Please also test that it is *different* from what we got in the child thread.

src/shims/unix/env.rs
Mandragorian: From what I understand [arithmetic_side_effects](https://rust-lang.github.io/rust-clippy/master/index.html#/arithmetic_side_effects) this lint tries to warn about potential panics or overflows. Since From what I understand the thread IDs and the process IDs come from the same namespace, if it is safe to cast a pid from u32 to i32, `pid + thread_id` would not overflow a u32.

tests/pass-dep/libc/gettid.rs
Mandragorian: It seems that macos doesn't have a `gettid` function in its libc. In the libc crate there is a `gettid` function in `unix/linux-like` and `unix/nto`. Would it make more sense to only support `gettid` for linux targets instead?

src/shims/unix/env.rs
RalfJung: We always use `strict_add` (or the equivalent) in Miri unless there's a good reason to use anything else.

tests/pass-dep/libc/gettid.rs
RalfJung: Yes in that case please move the gettid shim to `linux/foreign_items.rs` and also add an assertion inside `fn gettid` to ensure that it is only used on Linux.

tests/pass-dep/libc/gettid.rs
Mandragorian: done.

tests/pass-dep/libc/gettid.rs
Mandragorian: done.

src/shims/unix/env.rs
Mandragorian: done.

src/shims/env.rs
Mandragorian: done.

Could you explain a bit why we should never match `InterpResult`?

src/shims/env.rs
Mandragorian: done.

Is there a reason methods in extension trait are preferred? To not pollute the namespace?

src/shims/env.rs
RalfJung: InterpResult is constructed when we want to abort execution, e.g. on UB. It is always a bug to do anything but abort execution at that point.

RalfJung: I wish we had a way to ensure that `InterpResult` is never "caught" in this way...

Mandragorian: I have never done this, so I don't know how easy it is, but can't some lint be created for that?

RalfJung: Yeah, but clippy would have to know which error types are okay with a `Result<T, E>` to be `match`ed on and which not. 

Anyway, that's off-topic here.

src/shims/env.rs
RalfJung: To avoid having to import each function separately, and to keep things consistent.

tests/pass-dep/libc/gettid.rs
RalfJung, line 1: Thinking about this test a bit more, we currently have no test coverage for `gettid` without isolation.

So maybe make this test have two revisions
```
//@revisions: with_isolation without_isolation
//@[without_isolation] compiler-flags: -Zmiri-disable-isolation
```
Then some thing can be tested for all threads, and the check that for `1001` can be guarded by `if cfg!(with_isolation)`.

tests/pass/shims/env/getpid_with_isolation.rs
RalfJung, line 1: Please use revisions so that this can be in the same file as `getpid.rs`.

Mandragorian: done.

tests/pass-dep/libc/gettid.rs
Mandragorian: done

src/shims/unix/env.rs
```diff
@@ -274,12 +274,22 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        Ok(this.get_pid() as i32)
+    }
+
+    fn linux_gettid(&mut self) -> InterpResult<'tcx, i32> {
+        let this = self.eval_context_ref();
+        this.assert_target_os("linux", "gettid");
+
+        let index = this.machine.threads.active_thread().to_u32();
+
+        let tid = this.get_pid().strict_add(index);
```
RalfJung, line 290:
```suggestion
        // Compute a TID for this thread, ensuring that the main thread has PID == TID.
        let tid = this.get_pid().strict_add(index);
```

RalfJung:
Just one final comment nit. :)

Then please squash the commits, and we are good to go!
@rustbot author

RalfJung:
Thanks a lot for the PR!

@bors r+

Issue #3449, opened 2024-04-04
Implement basic support for sockets


RalfJung:
Miri can access files but not sockets currently.

The main issue with implementing blocking sockets is that they would provide *the first source of blocking in Miri that is controlled by the outside world* (except for time but we've got that handled pretty well). That will be quite non-trivial to implement as Miri needs to basically become or import an async runtime. We'll have a concept of threads being blocked on a socket. When all threads are blocked and Miri goes to sleep to wait for a timeout to pass, it needs to be able to wait on "either the timeout passes or all the sockets any thread is blocked on".  When the socket unblocks we should wake up the thread, even if we never reach the "all threads are blocked" state. Both of these are exactly the core operations of an async runtime.

Somewhat surprisingly, I do *not* think that epoll makes this a lot more complicated. Even without epoll, we could have 5 threads waiting on a different socket each, so Miri basically needs epoll on the host even if Miri programs do not have epoll available. Having epoll just means that a single Miri thread can wait on more than one socket, but since Miri anyway has to support many threads that doesn't add significant new complications.

Non-blocking sockets would be a lot simpler, but probably also a lot less interesting.

RalfJung:
As [discussed on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/269128-miri/topic/socket.20support), it would probably make sense to use `mio` on the Miri side here. Fundamentally, the primitives we need are:
- each time the active thread yields, do a non-blocking poll on a set of host sockets -- that's generalizing what we already do [here](https://github.com/rust-lang/miri/blob/ec51a35279fcbd73f352986acffbc3a3acb042af/src/concurrency/thread.rs#L748-L752) where we check whether there's a timer we need to fire
- if all Miri threads are blocked, do a blocking poll on a set of host sockets (potentially with a timeout) -- that's generalizing what we do [here](https://github.com/rust-lang/miri/blob/de5464233a0e430943a9e886817efefe81f72e90/src/concurrency/thread.rs#L792-L796) and [here](https://github.com/rust-lang/miri/blob/de5464233a0e430943a9e886817efefe81f72e90/src/concurrency/thread.rs#L1103-L1105) where we just sleep until the next timeout expires

[...]

PR #3759, opened 2024-07-22
Add `flock` shim

newpavlov:


src/shims/unix/fs.rs
```diff
@@ -81,6 +81,48 @@ impl FileDescription for FileHandle {
...
+                -1 => Err(io::Error::last_os_error()),
+                ret => panic!("Unexpected return value from flock: {ret}"),
+            };
+            Ok(res)
+        }
+
+        #[cfg(not(unix))]
+        {
+            let _ = op;
+            throw_unsup_format!("flock emulation is not supported on non-UNIX hosts");
```
newpavlov, line 122: It may be possible to emulate `flock` on Windows using [`LockFileEx`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-lockfileex), but I am not familiar with this API, so I do not plan to implement it.

oli-obk: Instead of doing this work ourselves in miri, is https://crates.io/crates/fs4 tunable enough to emulate this correctly?

newpavlov: I think it's a small enough code to include it into Miri directly. I will try to write it based on the `fs4` code.

src/shims/unix/fs.rs
```diff
@@ -81,6 +81,89 @@ impl FileDescription for FileHandle {
...
+            let res = match ret {
+                TRUE => Ok(()),
+                FALSE => {
+                    let mut err = io::Error::last_os_error();
+                    let code: u32 = err.raw_os_error().unwrap().try_into().unwrap();
+                    // Replace error with a custom WouldBlock error, which later will be mapped
+                    // in the `helpers` module
+                    if lock_nb && matches!(code, ERROR_IO_PENDING | ERROR_LOCK_VIOLATION) {
+                        let desc = format!("LockFileEx wouldblock error: {err}");
+                        err = io::Error::new(io::ErrorKind::WouldBlock, desc);
```
newpavlov, line 151: This remapping is a bit hacky, but I couldn't find a better solution. Mapping it here or in `fd.rs` conflicts with mapping in `helpers.rs`.

Also, according to `LockFileEx` non-blocking call should return `ERROR_IO_PENDING`, but in practice we get `ERROR_LOCK_VIOLATION`, so to be safe I remap both of those.

tests/pass-dep/libc/libc-fs-flock.rs
```diff
@@ -0,0 +1,71 @@
+// Flock tests are separate since they don't in general work on a Windows host.
+//@ignore-target-windows: File handling is not implemented yet
```
oli-obk, line 2: Was this meant to filter the host? Or is there an issue with windows targets if emulated from linux?

src/shims/unix/fs.rs
```diff
@@ -127,6 +129,89 @@ impl FileDescription for FileHandle {
...
+                }
+                _ => panic!("Unexpected return value: {ret}"),
+            };
+            Ok(res)
+        }
+
+        #[cfg(not(any(target_family = "unix", target_family = "windows")))]
+        {
+            let _ = op;
+            throw_unsup_format!("flock is supported only on UNIX and Windows hosts");
```
oli-obk, line 211: Do we even have other hosts? We could probably just `compile_error!` here

RalfJung: We have a bunch of code with `cfg(unix)` and `cfg(windows)` and nothing else. Miri does currently not build on other hosts.

src/shims/unix/fs.rs
```diff
@@ -127,6 +129,89 @@ impl FileDescription for FileHandle {
...
+            let host_op = match op {
+                SharedLock { nonblocking: false } => libc::LOCK_SH,
+                SharedLock { nonblocking: true } => libc::LOCK_SH | libc::LOCK_NB,
+                ExclusiveLock { nonblocking: false } => libc::LOCK_EX,
+                ExclusiveLock { nonblocking: true } => libc::LOCK_EX | libc::LOCK_NB,
+                Unlock => libc::LOCK_UN,
+            };
+
+            let fd = self.file.as_raw_fd();
+            let ret = unsafe { libc::flock(fd, host_op) };
```
oli-obk, line 153: in the blocking modes this will just block the interpreter, even though it may have other interpreted threads that it could make progress on. I think we'll need to do some emulation of blocking mode by using the nonblocking operations and blocking the interpreter thread only

tests/pass-dep/libc/libc-fs-flock.rs
newpavlov: No, it's about Windows targets, not hosts. This test is being executed on Windows host for Linux target.

src/shims/unix/fs.rs
newpavlov: Lately, I prefer to use `target_family` instead of `cfg(unix/windows)`. I think having this future-proofing is fine.

src/shims/unix/fs.rs
RalfJung: Ouch, yeah, good catch -- we can't call blocking host operations in Miri. We shouldn't expose blocking `flock` at all for now; some significant design work is needed before we can do this. See also https://github.com/rust-lang/miri/issues/3449.

newpavlov: Maybe we could use non-blocking `flock` even when blocking was requested and raise an error in case if it would block? I think it should be a good enough workaround for the time being.

oli-obk: We already support putting threads to sleep. I think this could be a good candidate for implementing #3449 without having to go all the way to arbitrary socket support.

newpavlov: So you suggest to use non-blocking `flock` and if it fails put thread to sleep? Can you point to a relevant piece of code, which I could use as a reference?

oli-obk: > Maybe we could use non-blocking `flock` even when blocking was requested and raise an error in case if it would block? I think it should be a good enough workaround for the time being.

that may be the better immediate step to land this PR, then the "put thread to sleep" logic can be implemented in a separate PR as there's a bit of design discussions that have to happen first

RalfJung: I also think it'd be better to just do non-blocking in this PR, and leave blocking for later.

> Can you point to a relevant piece of code, which I could use as a reference?

That's the thing, there's no code like that in Miri yet, we don't support any operations that block on the host. All blocking so far is entirely controlled by Miri itself (like blocking on a futex that is managed by Miri).

newpavlov: >I also think it'd be better to just do non-blocking in this PR, and leave blocking for later.

I would prefer to have a rudimentary blocking `flock` support. How about my suggestion above to throw an error when blocking `flock` is requested, but non-blocking `flock` has failed?

RalfJung: > Maybe we could use non-blocking flock even when blocking was requested and raise an error in case if it would block? I think it should be a good enough workaround for the time being.

Yes that sounds good, that's pretty much what I meant by not supporting blocking for now.

tests/pass-dep/libc/libc-fs-flock.rs
oli-obk, line 1: Can this test fail spuriously depend on host file system usage? Or is the way you're using it guaranteed to never block?

newpavlov: According to the [man page](https://www.man7.org/linux/man-pages/man2/flock.2.html), I don't think that spurious failures should be possible. `EINTR` should happen only in contended cases (i.e. when we wait for a lock to be released), while `ENOLCK` should happen only in extreme cases.

oli-obk (approved):

tests/pass-dep/libc/libc-fs-flock.rs
```diff
@@ -0,0 +1,71 @@
+// Flock tests are separate since they don't in general work on a Windows host.
```
RalfJung, line 1: The comment is confusing... CI seems to indicate that it works just fine on a windows host?

newpavlov: Yeah, it was copied from another test and I forgot to update it after the emulation was implemented.

PR #3712, opened 2024-06-26
Implement epoll shim

tiif:
This PR:
- implemented non-blocking ``epoll`` for #3448 . The design for this PR is documented in https://hackmd.io/@tiif/SJatftrH0 . 
- renamed FileDescriptor to FileDescriptionRef
- assigned an ID to every file description

PR #3763, opened 2024-07-25
Insert FileDescription instead of FileDescriptor in ``insert_fd``

tiif:
This PR moves the creation of ``FileDescriptor`` inside ``insert_fd``, and ``insert_fd`` now takes in ``FileDescription`` instead of ``FileDescriptor``.  This change is needed by #3712.

### DIFF
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -192,10 +192,6 @@ impl FileDescription for NullOutput {
 pub struct FileDescriptor(Rc<RefCell<Box<dyn FileDescription>>>);
 
 impl FileDescriptor {
-    pub fn new<T: FileDescription>(fd: T) -> Self {
-        FileDescriptor(Rc::new(RefCell::new(Box::new(fd))))
-    }
-
     pub fn borrow(&self) -> Ref<'_, dyn FileDescription> {
         Ref::map(self.0.borrow(), |fd| fd.as_ref())
     }
@@ -227,20 +223,25 @@ impl VisitProvenance for FdTable {
 }
 
 impl FdTable {
-    pub(crate) fn new(mute_stdout_stderr: bool) -> FdTable {
-        let mut fds: BTreeMap<_, FileDescriptor> = BTreeMap::new();
-        fds.insert(0i32, FileDescriptor::new(io::stdin()));
+    fn new() -> Self {
+        FdTable { fds: BTreeMap::new() }
+    }
+    pub(crate) fn init(mute_stdout_stderr: bool) -> FdTable {
+        let mut fds = FdTable::new();
+        fds.insert_fd(io::stdin());
         if mute_stdout_stderr {
-            fds.insert(1i32, FileDescriptor::new(NullOutput));
-            fds.insert(2i32, FileDescriptor::new(NullOutput));
+            assert_eq!(fds.insert_fd(NullOutput), 1);
+            assert_eq!(fds.insert_fd(NullOutput), 2);
         } else {
-            fds.insert(1i32, FileDescriptor::new(io::stdout()));
-            fds.insert(2i32, FileDescriptor::new(io::stderr()));
+            assert_eq!(fds.insert_fd(io::stdout()), 1);
+            assert_eq!(fds.insert_fd(io::stderr()), 2);
         }
-        FdTable { fds }
+        fds
     }
 
-    pub fn insert_fd(&mut self, file_handle: FileDescriptor) -> i32 {
+    /// Insert a file descriptor to the FdTable.
+    pub fn insert_fd<T: FileDescription>(&mut self, fd: T) -> i32 {
+        let file_handle = FileDescriptor(Rc::new(RefCell::new(Box::new(fd))));
         self.insert_fd_with_min_fd(file_handle, 0)
     }
 
--- a/src/shims/unix/linux/epoll.rs
+++ b/src/shims/unix/linux/epoll.rs
@@ -5,8 +5,6 @@ use rustc_data_structures::fx::FxHashMap;
 use crate::shims::unix::*;
 use crate::*;
 
-use self::shims::unix::fd::FileDescriptor;
-
 /// An `Epoll` file descriptor connects file handles and epoll events
 #[derive(Clone, Debug, Default)]
 struct Epoll {
@@ -66,7 +64,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             );
         }
 
-        let fd = this.machine.fds.insert_fd(FileDescriptor::new(Epoll::default()));
+        let fd = this.machine.fds.insert_fd(Epoll::default());
         Ok(Scalar::from_i32(fd))
     }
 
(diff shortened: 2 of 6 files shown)


src/shims/unix/fd.rs
```diff
@@ -218,6 +226,8 @@ impl FileDescriptor {
 #[derive(Debug)]
 pub struct FdTable {
     pub fds: BTreeMap<i32, FileDescriptor>,
+    /// Unique identifier for file description, used to differentiate between various file description.
+    next_file_description_id: usize,
```
oli-obk, line 230: This could have been kept in the other PR, as it is not necessary on master, and fairly simple to add after the other changes of this PR

src/shims/unix/fd.rs
```diff
@@ -317,7 +340,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             // If old_fd and new_fd point to the same description, then `dup_fd` ensures we keep the underlying file description alive.
             if let Some(file_descriptor) = this.machine.fds.fds.insert(new_fd, dup_fd) {
                 // Ignore close error (not interpreter's) according to dup2() doc.
-                file_descriptor.close(this.machine.communicate())?.ok();
+                file_descriptor.close(this, this.machine.communicate())?.ok();
```
oli-obk, line 343: A bunch of unrelated changes were left in this PR

src/shims/unix/linux/epoll.rs
```diff
@@ -39,6 +37,7 @@ impl FileDescription for Epoll {
     fn close<'tcx>(
         self: Box<Self>,
         _communicate_allowed: bool,
+        _ecx: &mut MiriInterpCx<'tcx>,
```
oli-obk, line 40: I'd prefer to just leave it out and add it in the other PR. After removing the integer ids from this PR there should not be too many conflicts

src/shims/unix/fd.rs
```diff
@@ -187,28 +188,35 @@ impl FileDescription for NullOutput {
         Ok(Ok(bytes.len()))
     }
 }
+/// Structure contains both the file description and its unique identifier.
+#[derive(Clone, Debug)]
+pub struct FileDescWithID<T: FileDescription + ?Sized> {
+    #[allow(dead_code)]
+    id: usize,
+    file_description: RefCell<Box<T>>,
+}
```
oli-obk, line 197: Similary, this one is not needed here

tiif: You mean the ``id`` field or the entire ``FileDescWithID`` struct? I currently only remove the ``id`` field but let me know if you mean the latter.

src/shims/unix/fd.rs
```diff
@@ -227,20 +223,29 @@ impl VisitProvenance for FdTable {
...
+    }
+    pub(crate) fn init(mute_stdout_stderr: bool) -> FdTable {
+        let mut fds = FdTable::new();
+        fds.insert_fd(io::stdin());
         if mute_stdout_stderr {
-            fds.insert(1i32, FileDescriptor::new(NullOutput));
-            fds.insert(2i32, FileDescriptor::new(NullOutput));
+            let mut fd = fds.insert_fd(NullOutput);
+            assert_eq!(fd, 1);
+            fd = fds.insert_fd(NullOutput);
```
oli-obk, line 235: Use separate let bindings instead of mutation or even move the entire method call directly into the assert

oli-obk (approved):
Just a nit then we can land this

src/shims/unix/fd.rs
```diff
@@ -227,20 +223,25 @@ impl VisitProvenance for FdTable {
...
-            fds.insert(2i32, FileDescriptor::new(io::stderr()));
+            assert_eq!(fds.insert_fd(io::stdout()), 1);
+            assert_eq!(fds.insert_fd(io::stderr()), 2);
         }
-        FdTable { fds }
+        fds
     }
 
-    pub fn insert_fd(&mut self, file_handle: FileDescriptor) -> i32 {
+    /// Insert a file descriptor to the FdTable.
```
RalfJung, line 242: The comment seems to contradict the signature? This is a file description, isn't it?

src/shims/unix/fd.rs
```diff
@@ -227,20 +223,25 @@ impl VisitProvenance for FdTable {
...
+    pub fn insert_fd<T: FileDescription>(&mut self, fd: T) -> i32 {
```
RalfJung, line 243: I'm not sure what the motivation is for this API change, but it seems odd to me. Note that `insert_fd_with_min_fd` wasn't changed, so now we have two equally-named APIs doing very different things. And `insert_fd_with_min_fd` can't be changed since indeed it is used to insert duplicates of existing file descriptions that shouldn't be wrapped in a new `Rc`.

> This change is needed by https://github.com/rust-lang/miri/pull/3712.

I don't understand how it is possible for this change to be *needed*. This change means that the API is now strictly less powerful than it was before. The only argument I can currently imagine for this change is ergonomics, and I don't agree that's worth it given the API inconsistencies this introduces.

IMO this should be reverted, both `insert_fd` and `insert_fd_with_min_fd` should take a `FileDescriptionRef`.

oli-obk: The API didn't need to be as powerful as it was, so this seems good (restricting to the use case we need). It's also the only sane design I see for having each `FileDescriptionrRef` carry a unique id that we can use instead of its address.

oli-obk: We considered do it in `FileDescriptionRef::new`, but then that needs access to the InterpCx or at least the `FdTable`.

RalfJung: It is pure coincidence that we have no operation which creates a 2nd file descriptor for an existing file description, and calls this function. In fact arguably [this here](https://github.com/rust-lang/miri/blob/79285bb87fb959d72d538a9c5eeaf5aabcea1829/src/shims/unix/fd.rs#L307) *should* just call `insert_fd`.

And having two `insert_fd` functions that don't even take the same kind of argument seems like an obviously bad API to me, *way* worse than an API that is slightly more general than what we happen to need right now.

Creating a `FileDescriptionRef` from a `FileDescription` should be a completely independent operation from inserting a `FileDescriptionRef` into the FD table.

RalfJung: > It's also the only sane design I see for having each FileDescriptionrRef carry a unique id that we can use instead of its address.

Just have `FileDescriptionRef::new` take a reference like `ecx: &mut MiriInterpCx<'tcx>` and then it can do the ID assignment.

RalfJung: I guess we could call the functions `insert_new_fd` and `insert_existing_fd` and then we don't even have to change their implementation. (Though i think factoring `FileDescriptionRef::new` into a separate operation, even if it remains private, would still help to make the conceptual structure more clear.)

The PR description should have mentioned the point about `FileDescriptionRef::new` needing access to global state in https://github.com/rust-lang/miri/pull/3712, that would have avoided some confusion. :)

Issue #3764, opened 2024-07-26
Tree-borrows false positive with `iter::Extend`?


ohsayan:
## Violating code

While this code might look unnecessarily funky, it is so because I stripped it down to the (hopefully) bare minimum for a repro. Here's the code:

```rust
use std::{mem::ManuallyDrop, slice};

#[test]
fn extend_fail() {
    let data = Data::allocate(b"hello");
    assert_eq!(data.read(), b"hello");
    let mut buffer = Vec::<u8>::with_capacity(1024);
    buffer.extend(data.read());
    assert_eq!(buffer.len(), 5);
}

#[test]
fn extend_pass() {
    let data = Data::allocate(b"hello");
    assert_eq!(data.read(), b"hello");
    let mut buffer = Vec::<u8>::with_capacity(1024);
    for byte in data.read() {
        buffer.push(*byte);
    }
    assert_eq!(buffer.len(), 5);
}

pub struct Data {
    data: DataUnion,
}

impl Data {
    pub fn allocate(data: &[u8]) -> Self {
        let mut data = ManuallyDrop::new(data.to_owned().into_boxed_slice());
        Self {
            data: DataUnion {
                x: ManuallyDrop::new(FatPointer::new(
                    data.as_mut_ptr() as usize,
                    data.len() as u64,
                )),
            },
        }
    }
    pub fn read(&self) -> &[u8] {
        unsafe {
            // SAFETY: fine since we r/w it directly!
            slice::from_raw_parts(self.data.x.ptr as *mut u8, self.data.x.len as usize)
        }
    }
}

impl Drop for Data {
    fn drop(&mut self) {
        unsafe {
            Vec::from_raw_parts(
                self.data.x.ptr as *mut u8,
                self.data.x.len as usize,
                self.data.x.len as usize,
            );
        }
    }
}

struct FatPointer {
    len: u64,
    ptr: usize,
}

impl FatPointer {
    fn new(ptr: usize, len: u64) -> Self {
        Self { len, ptr }
    }
}

union DataUnion {
    x: ManuallyDrop<FatPointer>,
}
```

What seemed interesting to me is that neither stacked borrows nor tree borrows reports an error with the `extend_pass` test when run with either of:
```sh
MIRIFLAGS="-Zmiri-permissive-provenance -Zmiri-tree-borrows" cargo miri test extend_pass
MIRIFLAGS="-Zmiri-permissive-provenance" cargo miri test extend_pass
```

[...]

PR #3766, opened 2024-07-26
better diagnostics for Tree Borrows + int2ptr casts

RalfJung:
- Entirely reject `-Zmiri-permissive-provenance -Zmiri-tree-borrows` since that combination just doesn't work
- In the int2ptr cast warning, when Tree Borrows is enabled, do not recommend `-Zmiri-permissive-provenance`, instead note that Tree Borrows does not support int2ptr casts

Fixes https://github.com/rust-lang/miri/issues/3764


ohsayan:
Also, is there a way to suppress the `int2ptr` casts warning when using tree borrows?

RalfJung:
Not really -- since Tree Borrows doesn't support int2ptr casts, not showing the warning seems hazardous. Now that I understand that your original issue was unrelated, we could possibly partially revert this PR -- but it is still the case that Tree Borrows + int2ptr will miss a lot of UB (even more so than usual for int2ptr), and I am not sure how to best communicate that to users.

The warning should be shown only once. Is it so bad that it needs to be hidden?

ohsayan:
@RalfJung I see, and it makes complete sense. Actually, I've been hunting a bunch of bugs down and the int-to-ptr cast warnings filled the logs...because we use them excessively for storing extra pointer metadata using alignment hacks; that's why I was wondering if there was a way around it.

Also, since we use `crossbeam`, as it happens, tests fail when using stacked borrows but pass using tree borrows (and it's a known issue; see crossbeam-rs/crossbeam#545). So yeah, in a bit of a situation. But I can totally see why suppressing `int2ptr` warnings could be dangerous.

RalfJung:
> storing extra pointer metadata using alignment hacks

FWIW this can be done without int2ptr casts using strict provenance, but that currently requires either nightly or using the [sptr crate](https://crates.io/crates/sptr).

Issue #3498, opened 2024-04-22
```no_std``` issues with implementing ```panic_handler```

### ISSUE THREAD

**Issue by Necromaticon** [rust-lang/miri#3498]
I'm trying to interpet a very simple ```no_std``` example before going for any bigger problems.
 Problem appears to be that none of the ```panic_handler``` crates are working with Miri.
 
 The command I'm using to run the code is:
```MIRI_NO_STD=1 cargo miri run```
 
 Code:
 ```
#![no_main]
#![no_std]

//insert individual panic methods here

use cortex_m_rt;

#[cortex_m_rt::entry]
fn main() -> ! {
    loop {}
}
```
 
 First off, both ```use panic_halt as _;``` and ```use panic_abort as _;``` create the following issue:
 ```
 error: unwinding panics are not supported without std
  |
  = help: using nightly cargo, use -Zbuild-std with panic="abort" to avoid unwinding
  = note: since the core library is usually precompiled with panic="unwind", rebuilding your crate with panic="abort" may not be enough to fix the problem
```

  That's something that seems to be an issue with ```cargo build``` in general and not just Miri specific. I don't quite understand why ```abort``` and ```halt``` wouldn't work for ```no_std``` environments though considering these panic handlers were made for this express purpose.
  
  So I tried going for ```use panic_semihosting as _;``` and encountered an even weirder problem where the crates aren't recognized on build but only when I run ```cargo miri```. If I run ```cargo run``` everything works without a hitch.
Error:
```
error[E0463]: can't find crate for `panic_semihosting`
 --> embeddedMiri/src/main.rs:4:5
  |
4 | use panic_semihosting as _;
  |     ^^^^^^^^^^^^^^^^^ can't find crate
  |
  = note: duplicate diagnostic emitted due to `-Z deduplicate-diagnostics=no`

error: `#[panic_handler]` function required, but not found

error: unwinding panics are not supported without std
  |
  = help: using nightly cargo, use -Zbuild-std with panic="abort" to avoid unwinding
  = note: since the core library is usually precompiled with panic="unwind", rebuilding your crate with panic="abort" may not be enough to fix the problem
  ```
  It seems that Miri fails to somehow link the cargo crates when building. I've also tried building with ```extern crate``` but then the problem of the panic handler being missing consists and all examples of panic handlers using the ```#[panic_handler]``` flag I found invoke the unwinding panics error.
  
 Thank you in advance for the help!

RalfJung:
[...]

PR #3769, opened 2024-07-28
Add `miri_start` support

### LINKED ISSUE #3758: Support miri-specific entry point / start function for no-std binaries
For no-std binaries, Miri currently relies on the `start` feature, but that feature [doesn't really make a lot of sense](https://github.com/rust-lang/rust/issues/29633#issuecomment-2088596042) and should likely be removed. It also causes [confusion](https://github.com/rust-lang/miri/issues/3498) for Miri users.

So we should instead say that in a no-std binary, we Miri will start the program by jumping to a function with link_name `miri_start`, and then if someone has a no-std binary they want to run in Miri, they can add something like
```rust
#[cfg(miri)]
#[no_mangle]
fn miri_start() {
  // Call the actual start function that your project implements, based on your target's conventions.
}
```

primoly:
This PR uses a function with the exported symbol `miri_start` as a drop-in alternative to `#[start]`. So the signature stays the same as suggested in [this comment](https://github.com/rust-lang/miri/issues/3498#issuecomment-2088560526). <del>I’ve also removed Miri’s restriction  to only work on bin crates as I don’t think this is necessary anymore.</del>

Closes #3758

### DIFF
--- a/src/bin/miri.rs
+++ b/src/bin/miri.rs
@@ -358,7 +358,7 @@ fn entry_fn(tcx: TyCtxt<'_>) -> (DefId, EntryFnType) {
     if let Some(entry_def) = tcx.entry_fn(()) {
         return entry_def;
     }
-    // Look for a symbol in the local crate named `miri_start`, and threat that as the entry point.
+    // Look for a symbol in the local crate named `miri_start`, and treat that as the entry point.
     let sym = tcx.exported_symbols(LOCAL_CRATE).iter().find_map(|(sym, _)| {
         if sym.symbol_name_for_local_instance(tcx).name == "miri_start" { Some(sym) } else { None }
     });
@@ -391,14 +391,16 @@ fn entry_fn(tcx: TyCtxt<'_>) -> (DefId, EntryFnType) {
             );
         }
     } else {
-        tcx.dcx().fatal("Miri can only run programs that have a main function.\n\
-                    Alternatively, you can export a `miri_start` function:\n\
-                    \n\
-                    #[cfg(miri)]\n\
-                    #[no_mangle]\n\
-                    fn miri_start(argc: isize, argv: *const *const u8) -> isize {\
-                    \n    // Call the actual start function that your project implements, based on your target's conventions.\n\
-                    }");
+        tcx.dcx().fatal(
+            "Miri can only run programs that have a main function.\n\
+            Alternatively, you can export a `miri_start` function:\n\
+            \n\
+            #[cfg(miri)]\n\
+            #[no_mangle]\n\
+            fn miri_start(argc: isize, argv: *const *const u8) -> isize {\
+            \n    // Call the actual start function that your project implements, based on your target's conventions.\n\
+            }"
+        );
     }
 }
 
--- a/src/bin/miri.rs
+++ b/src/bin/miri.rs
@@ -14,11 +14,14 @@ extern crate tracing;
 extern crate rustc_data_structures;
 extern crate rustc_driver;
 extern crate rustc_hir;
+extern crate rustc_hir_analysis;
 extern crate rustc_interface;
 extern crate rustc_log;
 extern crate rustc_metadata;
 extern crate rustc_middle;
 extern crate rustc_session;
+extern crate rustc_span;
+extern crate rustc_target;
 
 use std::env::{self, VarError};
 use std::num::NonZero;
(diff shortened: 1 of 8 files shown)


src/bin/miri.rs
```diff
@@ -78,14 +78,22 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
 
             let early_dcx = EarlyDiagCtxt::new(tcx.sess.opts.error_format);
             init_late_loggers(&early_dcx, tcx);
-            if !tcx.crate_types().contains(&CrateType::Executable) {
```
Noratrieb, line 81: I think this restriction should stay, it doesn't make a lot of sense to execute a library.

src/bin/miri.rs
```diff
@@ -78,14 +78,22 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
...
+                    if sym.symbol_name_for_local_instance(tcx).name == "miri_start" {
+                        Some(sym)
+                    } else {
+                        None
+                    }
+                });
+                if let Some(ExportedSymbol::NonGeneric(id)) = sym {
+                    (*id, EntryFnType::Start)
+                } else {
+                    tcx.dcx().fatal("miri can only run programs that have a main function or one exported with symbol miri_start");
```
Noratrieb, line 95: the error message should probably differentiate between no symbol found and a generic symbol found

Noratrieb: I'd then reword the not found one to
> Miri can only run programs that have a main function or export the symbol `miri_start` (for example, with `#[no_mangle] fn miri_start()`)"

src/bin/miri.rs
```diff
@@ -85,7 +90,55 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
             let (entry_def_id, entry_type) = if let Some(entry_def) = tcx.entry_fn(()) {
                 entry_def
             } else {
-                tcx.dcx().fatal("miri can only run programs that have a main function");
+                let sym = tcx.exported_symbols(LOCAL_CRATE).iter().find_map(|(sym, _)| {
```
RalfJung, line 93: Please factor this new entry point logic into a separate function.

tests/fail/no_main.rs
```diff
@@ -1,2 +1,9 @@
-//@error-in-other-file: miri can only run programs that have a main function
+//@error-in-other-file: Miri can only run programs that have a main function.
+//Alternatively, you can export a `miri_start` function:
+//
+//#[cfg(miri)]
+//#[no_mangle]
+//fn miri_start(argc: isize, argv: *const *const u8) -> isize {
+//    // Call the actual start function that your project implements, based on your target's conventions.
+//}
```
RalfJung, line 8: This should not be needed, `error-in-other-file` is not multi-line.

RalfJung (commented):
This is very nice, thanks. :) I just have some minor comments.

src/bin/miri.rs
```diff
@@ -404,6 +354,61 @@ fn jemalloc_magic() {
     }
 }
 
+fn entry_fn(tcx: TyCtxt<'_>) -> (DefId, EntryFnType) {
+    if let Some(entry_def) = tcx.entry_fn(()) {
+        entry_def
+    } else {
+        let sym = tcx.exported_symbols(LOCAL_CRATE).iter().find_map(|(sym, _)| {
```
RalfJung, line 361:
```suggestion
        // Look for a symbol in the local crate named `miri_start`, and threat that as the entry point.
        let sym = tcx.exported_symbols(LOCAL_CRATE).iter().find_map(|(sym, _)| {
```

src/bin/miri.rs
```diff
@@ -404,6 +354,61 @@ fn jemalloc_magic() {
...
+                    (*id, EntryFnType::Start)
+                } else {
+                    tcx.dcx().fatal(
+                        "`miri_start` must have the following signature:\
+                        \nfn miri_start(argc: isize, argv: *const *const u8) -> isize",
+                    );
+                }
+            }
+            _ => {
+                tcx.dcx().fatal("Miri can only run programs that have a main function.\
```
RalfJung, line 399: Please put the `\n` at the ends of these lines, rather than the beginning of the next lines.

RalfJung:
Please also squash this into a single commit. :)

RalfJung:
Thanks. :)

@bors r+

Issue #3772, opened 2024-07-29
Project idea: strongly typed file descriptor integers and deduplication of various checks around them


oli-obk:
This is less of a fundamental/large-in-behavior change and more a very bitrotty change with possibly not-really-deduplicating-that-much changes that hopefully still make the code more manageable.

### Problem statement

Currently we represent file descriptors just as `i32` within miri. We mix them with various negative integer values for signalling errors, just like libc does. We also duplicate a lot of error handling logic like

```rust
let Some(fd) = this.machine.fds.some_getter(fd) else {
   return this.fd_not_found();
};
```

and

```rust
Ok(Scalar::from_i32(this.try_unwrap_io_result(result)?))
```

### Solution parts

0. consistently use `Scalar` instead of `i32` as the return type of foreign function handlers (groundwork for the following, as we inconsistently use `i32` and `Scalar`. Even using `Scalar` when `i32` would definitely suffice, not just when it could be a pointer).
1. wrap all `i32` known-to-be-file-descriptors in a newtype struct together with some convenience functions for reading them and getting the integer back out (into a scalar)
2. return `io::Result<FileDescriptionRef>` (or the reference versions) from the various getters, instead of `Option<FileDescriptionRef>`. Forward that where useful, avoiding explicit `-1` returns paired with `set_last_error`, instead handling it with `try_unwrap_io_result`
3. add `get_mut`, `get`, and `dup` methods to `FdTable` that return `io::Result` instead of `Option`, even though the only io error they ever return is `NotFound`, allowing us to `?` away their result

RalfJung:
It would also be good to be able to write something like `return this.set_last_error_and_return_neg1(...);` which would set the last error and then return `-1`, as that's such a common pattern.

oli-obk:
These constants (almost?) always have a corresponding ErrKind, so I think we can transform them to `io::Result` usage. I mean, that is how we convert `io::Error` to an error value to set, we eval the corresponding constant

RalfJung:
It seems silly to construct a Result if we hard-code the Err variant - at least we want a method that takes an io::Errror.

But I am not sure we can always construct an ErrorKind.

oli-obk:
Separate idea: instead of hardcoding the io::Result to integer conversion logic, we add a new enum around `io::Error` that has variants for the extra cases  we need. Then we use that enum as the error type everywhere

noahmbright:
[...]

PR #3776, opened 2024-07-30
Use Scalar consistently in foreign item emulation

oli-obk:
Step 0 of #3772

This just makes the code consistent. While we could also go for consistency the other way, that would only allow us to use integers in some cases where we use `Scalar` right now, because `Scalar` can also hold pointers where applicable.

There's also no danger in messing up (even though we do lose some compile-time checks), as `Scalar` knows the size of the integer stored within, so it will check that against the destination when it is written. In fact, this makes the checks much stronger compared with `write_int`, which just checks that the integer fits into the destination size.

### DIFF
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -62,13 +62,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             "unsetenv" => {
                 let [name] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
                 let result = this.unsetenv(name)?;
-                this.write_scalar(Scalar::from_i32(result), dest)?;
+                this.write_scalar(result, dest)?;
             }
             "setenv" => {
                 let [name, value, overwrite] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
                 this.read_scalar(overwrite)?.to_i32()?;
                 let result = this.setenv(name, value)?;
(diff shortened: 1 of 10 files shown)


src/shims/unix/foreign_items.rs
```diff
@@ -552,28 +552,28 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                this.write_scalar(Scalar::from_i32(result), dest)?;
+                this.pthread_cond_signal(cond)?;
+                this.write_null(dest)?;
             }
             "pthread_cond_broadcast" => {
                 let [cond] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
-                let result = this.pthread_cond_broadcast(cond)?;
-                this.write_scalar(Scalar::from_i32(result), dest)?;
+                this.pthread_cond_broadcast(cond)?;
+                this.write_null(dest)?;
```
RalfJung, line 576: Isn't it nicer to have the knowledge that these always return null inside `pthread_cond_broadcast` rather than here?

I feel like ideally, there's as little function-specific knowledge in this file as possible. (In fact ideally we'd have a macro generate the `match` in perfect uniformity... if someone is motivated enough to write such a macro, one day.)

oli-obk: I guess this would be resolved by always passing in the `dest` and handling it inside entirely

RalfJung: That's also an option, yes.

PR #3778, opened 2024-07-30
FD: refactor API names a little

RalfJung:
I feel like these two functions are different enough that their names should not indicate complete symmetry. I am not sure what the best names would be, though... @oli-obk any opinions?

### DIFF
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -294,7 +294,7 @@ impl FdTable {
         Some(fd.borrow_mut())
     }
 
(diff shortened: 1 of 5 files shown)


oli-obk:
No real opinion. The `new` as part of the name is useful.

What actually bugs me is the `dup` function xD should probably be `get_cloned_ref` or sth

RalfJung:
True, `dup` is an odd one. I called it `get_ref` for now. I wonder if we should just call it `get` and remove the other two getters -- it is slightly less efficient to always clone, but that's now what is going to make Miri slow...

PR #3779, opened 2024-07-31
WIP: handle writing to dest within the foreign item handler

oli-obk:
alternative to https://github.com/rust-lang/miri/pull/3777

### DIFF
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -775,6 +775,16 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         this.write_scalar(scalar, &errno_place)
     }
 
+    fn set_libc_err_and_return_neg1(
+        &mut self,
+        err: &str,
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx> {
+        let err = self.eval_libc(err);
+        self.set_last_error(err)?;
+        self.eval_context_mut().write_int(-1, dest)
+    }
+
     /// Gets the last error variable.
     fn get_last_error(&mut self) -> InterpResult<'tcx, Scalar> {
         let this = self.eval_context_mut();
@@ -844,6 +854,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     /// Sets the last OS error using a `std::io::ErrorKind`.
     fn set_last_error_from_io_error(&mut self, err: std::io::Error) -> InterpResult<'tcx> {
         self.set_last_error(self.io_error_to_errnum(err)?)
+    }
+
+    /// Helper function that consumes an `std::io::Result<i32>` and writes the `Ok` integer
+    /// to the destination. In case the result is an error, this function writes
+    /// `-1` and sets the last OS error accordingly.
+    fn write_io_result(
+        &mut self,
+        result: std::io::Result<i32>,
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx> {
+        let res = self.try_unwrap_io_result(result)?;
+        self.eval_context_mut().write_int(res, dest)
     }
 
     /// Helper function that consumes an `std::io::Result<T>` and returns an
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -781,6 +781,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         dest: &MPlaceTy<'tcx>,
     ) -> InterpResult<'tcx> {
         let err = self.eval_libc(err);
+        self.set_last_err_and_return_neg1(err, dest)
+    }
+
+    fn set_last_err_and_return_neg1(
+        &mut self,
+        err: Scalar,
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx> {
         self.set_last_error(err)?;
         self.eval_context_mut().write_int(-1, dest)
     }
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -1038,7 +1038,11 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         }
     }
 
-    fn linux_readdir64(&mut self, dirp_op: &OpTy<'tcx>) -> InterpResult<'tcx, Scalar> {
+    fn linux_readdir64(
+        &mut self,
+        dirp_op: &OpTy<'tcx>,
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx> {
         let this = self.eval_context_mut();
 
         this.assert_target_os("linux", "readdir64");
@@ -1050,7 +1054,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             this.reject_in_isolation("`readdir`", reject_with)?;
             let eacc = this.eval_libc("EBADF");
             this.set_last_error(eacc)?;
-            return Ok(Scalar::null_ptr(this));
+            return this.write_scalar(eacc, dest);
         }
 
         let open_dir = this.machine.dirs.streams.get_mut(&dirp).ok_or_else(|| {
@@ -1128,7 +1132,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             this.deallocate_ptr(old_entry, None, MiriMemoryKind::Runtime.into())?;
         }
 
-        Ok(Scalar::from_maybe_pointer(entry.unwrap_or_else(Pointer::null), this))
+        this.write_pointer(entry.unwrap_or_else(Pointer::null), dest)
     }
 
     fn macos_fbsd_readdir_r(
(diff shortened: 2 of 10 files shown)


src/shims/unix/fs.rs
```diff
@@ -652,26 +653,26 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-            return Ok(Scalar::from_i32(-1));
+            return this.set_libc_err_and_return_neg1("EACCES", dest);
         }
 
         // `stat` always follows symlinks.
-        let metadata = match FileMetadata::from_path(this, &path, true)? {
+        let metadata = match FileMetadata::from_path(this, &path, true, dest)? {
             Some(metadata) => metadata,
-            None => return Ok(Scalar::from_i32(-1)), // `FileMetadata` has set errno
+            None => return Ok(EmulateItemResult::NeedsReturn),
```
RalfJung, line 662:
```suggestion
            None => return Ok(EmulateItemResult::NeedsReturn), // `FileMetadata::from_path` has set the return value
```
But this is a bit odd, seems cleaner to instead have `from_path` return a `Result` and do the errno+return handling here?

src/shims/unix/fs.rs
```diff
@@ -1130,14 +1142,17 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // Reject if isolation is enabled.
         if let IsolatedOp::Reject(reject_with) = this.machine.isolated_op {
             this.reject_in_isolation("`readdir_r`", reject_with)?;
-            // Set error code as "EBADF" (bad fd)
-            return Ok(Scalar::from_i32(this.fd_not_found()?));
+            // TODO: why does this set the global error code and returns -1?
+            // at the end of this function we don't set the code, we just return it.
```
RalfJung, line 1146: Probably a copy-paste mistake. Good catch!

src/helpers.rs
```diff
@@ -780,6 +780,41 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    ) -> InterpResult<'tcx, EmulateItemResult> {
+        self.set_libc_err_and_return_neg1("EINVAL", dest)
+    }
+
+    fn set_fd_not_found_and_return_neg1(
+        &mut self,
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx, EmulateItemResult> {
+        self.set_libc_err_and_return_neg1("EBADF", dest)
+    }
```
RalfJung, line 795: I'm not convinced that it's worth having these functions.

src/shims/unix/fs.rs
```diff
@@ -639,7 +639,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         &mut self,
         path_op: &OpTy<'tcx>,
         buf_op: &OpTy<'tcx>,
-    ) -> InterpResult<'tcx, Scalar> {
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx, EmulateItemResult> {
```
RalfJung, line 643: Is it really worth having the `EmulateItemResult` everywhere? What's the problem with the previous approach here?

oli-obk: consistency. If we want all FFI shims to work the same way and generate all of them with a macro, we either need to do this or have some trait shenanigans that allow `()` or `EmulateItemResult` and treating `()` as the usual case

RalfJung: We can have the macro support "simple" shims and "destination passing style" shims, to avoid overcomplicating the extremely common case.

src/shims/unix/fs.rs
oli-obk: A bit annoying as I have to return a `Result<T, Scalar>` to be able to handle both EBADF and `io::Error`, but generally nicer indeed

oli-obk:
if this is to your liking, I'll port everything

@rustbot ready

src/helpers.rs
```diff
@@ -775,6 +775,24 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+    fn set_last_err_and_return_neg1(
+        &mut self,
+        err: Scalar,
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx> {
+        self.set_last_error(err)?;
+        self.eval_context_mut().write_int(-1, dest)
+    }
```
RalfJung, line 794: Both of these new functions are superseded by the helper that @tiif added and that I extended. You should be able to use `set_last_error_and_return` now (after a rebase).

src/helpers.rs
```diff
@@ -846,6 +864,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         self.set_last_error(self.io_error_to_errnum(err)?)
     }
 
+    /// Helper function that consumes an `std::io::Result<i32>` and writes the `Ok` integer
+    /// to the destination. In case the result is an error, this function writes
+    /// `-1` and sets the last OS error accordingly.
+    fn write_io_result(
```
RalfJung, line 870: Please move this to the new io_error.rs file, and put it right below `try_unwrap_io_result` (since it extends that helper).

src/shims/unix/fs.rs
```diff
@@ -670,26 +671,25 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            return this.set_libc_err_and_return_neg1("EACCES", dest);
         }
 
         // `stat` always follows symlinks.
         let metadata = match FileMetadata::from_path(this, &path, true)? {
-            Some(metadata) => metadata,
-            None => return Ok(Scalar::from_i32(-1)), // `FileMetadata` has set errno
+            Ok(metadata) => metadata,
+            Err(e) => return this.set_last_err_and_return_neg1(e, dest),
         };
```
RalfJung, line 681: I get the feeling we will have this pattern a lot... given an `io::Result<T>`, call `return this.set_last_error_and_return` on the error path and unwrap the `T` otherwise. Do you think that'd be worth a `try!`-like macro?

src/shims/unix/fs.rs
```diff
@@ -1050,7 +1051,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             this.reject_in_isolation("`readdir`", reject_with)?;
             let eacc = this.eval_libc("EBADF");
             this.set_last_error(eacc)?;
-            return Ok(Scalar::null_ptr(this));
+            return this.write_scalar(eacc, dest);
```
RalfJung, line 1054: This changes behavior, doesn't it? We should be returning a null pointer but you made it return `eacc` now.

src/shims/unix/fs.rs
```diff
@@ -1710,9 +1709,9 @@ impl FileMetadata {
     fn from_fd_num<'tcx>(
         ecx: &mut MiriInterpCx<'tcx>,
         fd_num: i32,
-    ) -> InterpResult<'tcx, Option<FileMetadata>> {
+    ) -> InterpResult<'tcx, Result<FileMetadata, Scalar>> {
```
RalfJung, line 1712: Please add a doc comment explaining the meaning of this return type. (Same for the next function.)

src/shims/unix/fs.rs
```diff
@@ -1148,14 +1150,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // Reject if isolation is enabled.
         if let IsolatedOp::Reject(reject_with) = this.machine.isolated_op {
             this.reject_in_isolation("`readdir_r`", reject_with)?;
-            // Set error code as "EBADF" (bad fd)
-            return Ok(Scalar::from_i32(this.fd_not_found()?));
+            // return positive error number on error
```
RalfJung, line 1153:
```suggestion
            // Return positive error number on error (*without* setting errno!)
```

src/shims/unix/mem.rs
```diff
@@ -27,7 +27,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         flags: &OpTy<'tcx>,
         fd: &OpTy<'tcx>,
         offset: i128,
-    ) -> InterpResult<'tcx, Scalar> {
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx, EmulateItemResult> {
```
RalfJung, line 31: So these here now return `EmulateItemResult` but the ones in `fs.rs` do not. As I said [here](https://github.com/rust-lang/miri/pull/3779#discussion_r1709302063) I'd prefer to avoid having the `EmulateItemResult` for functions that don't need it. It is extremely rare to need this.

RalfJung (commented):
I generally like it, modulo the comments here. Note that this has more overlap with https://github.com/rust-lang/miri/issues/3930 than I thought, so we should be careful to avoid conflicts.

src/shims/unix/fs.rs
oli-obk: I should fix this one in a separate PR with a test that shows the behaviour change. The previous behaviour is buggy

src/shims/unix/fs.rs
oli-obk: Not necessary, I can use your new error enum instead

src/shims/unix/mem.rs
oli-obk: Probably just an oversight, I wanted to change all of them back.

src/shims/unix/fs.rs
RalfJung: Ah, even better :)

RalfJung: https://github.com/rust-lang/miri/pull/3990 did this part :)

RalfJung:
@oli-obk do you still plan to pursue this? It seems to have been superseded by a bunch of the other work in that area recently.

PR #3784, opened 2024-08-02
Detect when pthread_mutex_t is moved

### LINKED ISSUE #3749: pthreads synchronization primitives: detect mutex/rwlock/... being moved to a different location
pthread's synchronization primitives must not be moved to a different location once they were initialized. However, Miri does currently not catch that kind of UB. We probably want a fundamentally different implementation of these primitives to detect such bugs -- instead of storing data in the memory that is occupied by `pthread_mutex_t` et al, we use the *address* of the `pthread_mutex_t` as the key into some table for managing the lock.

This would also avoid having to figure out which "offsets" inside `pthread_mutex_t` to use on which OS to do what... we'd entirely stop caring about the layout of `pthread_mutex_t`. So this is a better way to achieve FreeBSD support for the pthread primitives as well.

- [x] mutex
- [x] rwlock
- [ ] condvar

Mandragorian:
What I am not sure about this PR is how to support storing the additional mutex data like its address and kind. If I understand correctly the `concurrency::sync::Mutex` struct is to be used by any mutex implementation. This possibly means that different implementation might want to store different data in the mutex. So any additional data should be implementation defined somehow. Solutions that come to mind:

- Store the additional data as `Box<dyn Any>` and the implementations can downcast their data when they fetch them.
- Have each shim implementation define a `static mut` map between `MutexID`s and the additional data. 

Let me know

Fixes #3749

### DIFF
--- a/src/concurrency/sync.rs
+++ b/src/concurrency/sync.rs
@@ -66,6 +66,27 @@ pub(super) use declare_id;
 
 declare_id!(MutexId);
 
+/// The mutex kind.
+#[derive(Debug, Clone, Copy)]
+#[non_exhaustive]
+pub enum MutexKind {
+    Invalid,
+    Normal,
+    Default,
+    Recursive,
+    ErrorCheck,
+}
+
+#[derive(Debug)]
+/// Additional data that may be used by shim implementations.
+pub struct AdditionalMutexData {
+    /// The mutex kind, used by some mutex implementations like pthreads mutexes.
+    pub kind: MutexKind,
+
+    /// The address of the mutex.
+    pub address: u64,
+}
+
 /// The mutex state.
 #[derive(Default, Debug)]
 struct Mutex {
@@ -77,6 +98,9 @@ struct Mutex {
     queue: VecDeque<ThreadId>,
     /// Mutex clock. This tracks the moment of the last unlock.
     clock: VClock,
+
+    /// Additional data that can be set by shim implementations.
+    data: Option<AdditionalMutexData>,
 }
 
 declare_id!(RwLockId);
@@ -168,13 +192,15 @@ pub(super) trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
     /// Returns `None` if memory stores a non-zero invalid ID.
     ///
     /// `get_objs` must return the `IndexVec` that stores all the objects of this type.
+    /// `create_obj` must create the new object if initialization is needed.
     #[inline]
-    fn get_or_create_id<Id: SyncId + Idx, T: Default>(
+    fn get_or_create_id<Id: SyncId + Idx, T>(
         &mut self,
         lock_op: &OpTy<'tcx>,
         lock_layout: TyAndLayout<'tcx>,
         offset: u64,
         get_objs: impl for<'a> Fn(&'a mut MiriInterpCx<'tcx>) -> &'a mut IndexVec<Id, T>,
+        create_obj: impl for<'a> FnOnce(&'a mut MiriInterpCx<'tcx>) -> InterpResult<'tcx, T>,
     ) -> InterpResult<'tcx, Option<Id>> {
         let this = self.eval_context_mut();
         let value_place =
@@ -196,7 +222,8 @@ pub(super) trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
         Ok(if success.to_bool().expect("compare_exchange's second return value is a bool") {
             // We set the in-memory ID to `next_index`, now also create this object in the machine
             // state.
-            let new_index = get_objs(this).push(T::default());
+            let obj = create_obj(this)?;
+            let new_index = get_objs(this).push(obj);
             assert_eq!(next_index, new_index);
             Some(new_index)
         } else {
@@ -208,6 +235,32 @@ pub(super) trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 Some(id)
             }
         })
+    }
+
+    /// Eagerly creates a Miri sync structure.
+    ///
+    /// `create_id` will store the index of the sync_structure in the memory pointed to by
+    /// `lock_op`, so that future calls to `get_or_create_id` will see it as initialized.
+    /// - `lock_op` must hold a pointer to the sync structure.
+    /// - `lock_layout` must be the memory layout of the sync structure.
+    /// - `offset` must be the offset inside the sync structure where its miri id will be stored.
+    /// - `get_objs` is described in `get_or_create_id`.
+    /// - `obj` must be the new sync object.
+    fn create_id<Id: SyncId + Idx, T>(
+        &mut self,
+        lock_op: &OpTy<'tcx>,
+        lock_layout: TyAndLayout<'tcx>,
+        offset: u64,
+        get_objs: impl for<'a> Fn(&'a mut MiriInterpCx<'tcx>) -> &'a mut IndexVec<Id, T>,
+        obj: T,
+    ) -> InterpResult<'tcx, Id> {
+        let this = self.eval_context_mut();
+        let value_place =
+            this.deref_pointer_and_offset(lock_op, offset, lock_layout, this.machine.layouts.u32)?;
+
+        let new_index = get_objs(this).push(obj);
+        this.write_scalar(Scalar::from_u32(new_index.to_u32()), &value_place)?;
+        Ok(new_index)
     }
 
     fn condvar_reacquire_mutex(
@@ -236,15 +289,53 @@ pub(super) trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
 // situations.
 impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    /// Eagerly create and initialize a new mutex.
+    fn mutex_create(
+        &mut self,
+        lock_op: &OpTy<'tcx>,
+        lock_layout: TyAndLayout<'tcx>,
+        offset: u64,
+        data: Option<AdditionalMutexData>,
+    ) -> InterpResult<'tcx, MutexId> {
+        let this = self.eval_context_mut();
+        this.create_id(
+            lock_op,
+            lock_layout,
+            offset,
+            |ecx| &mut ecx.machine.sync.mutexes,
+            Mutex { data, ..Default::default() },
+        )
+    }
+
+    /// Lazily create a new mutex.
+    /// `initialize_data` must return any additional data that a user wants to associate with the mutex.
     fn mutex_get_or_create_id(
         &mut self,
         lock_op: &OpTy<'tcx>,
         lock_layout: TyAndLayout<'tcx>,
         offset: u64,
+        initialize_data: impl for<'a> FnOnce(
+            &'a mut MiriInterpCx<'tcx>,
+        ) -> InterpResult<'tcx, Option<AdditionalMutexData>>,
     ) -> InterpResult<'tcx, MutexId> {
         let this = self.eval_context_mut();
-        this.get_or_create_id(lock_op, lock_layout, offset, |ecx| &mut ecx.machine.sync.mutexes)?
-            .ok_or_else(|| err_ub_format!("mutex has invalid ID").into())
+        this.get_or_create_id(
+            lock_op,
+            lock_layout,
+            offset,
+            |ecx| &mut ecx.machine.sync.mutexes,
+            |ecx| initialize_data(ecx).map(|data| Mutex { data, ..Default::default() }),
+        )?
+        .ok_or_else(|| err_ub_format!("mutex has invalid ID").into())
+    }
+
+    /// Retrieve the additional data stored for a mutex.
+    fn mutex_get_data<'a>(&'a mut self, id: MutexId) -> Option<&'a AdditionalMutexData>
+    where
+        'tcx: 'a,
+    {
+        let this = self.eval_context_ref();
+        this.machine.sync.mutexes[id].data.as_ref()
     }
 
     fn rwlock_get_or_create_id(
@@ -254,8 +345,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         offset: u64,
     ) -> InterpResult<'tcx, RwLockId> {
         let this = self.eval_context_mut();
-        this.get_or_create_id(lock_op, lock_layout, offset, |ecx| &mut ecx.machine.sync.rwlocks)?
-            .ok_or_else(|| err_ub_format!("rwlock has invalid ID").into())
+        this.get_or_create_id(
+            lock_op,
+            lock_layout,
+            offset,
+            |ecx| &mut ecx.machine.sync.rwlocks,
+            |_| Ok(Default::default()),
+        )?
+        .ok_or_else(|| err_ub_format!("rwlock has invalid ID").into())
     }
 
     fn condvar_get_or_create_id(
@@ -265,8 +362,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         offset: u64,
     ) -> InterpResult<'tcx, CondvarId> {
         let this = self.eval_context_mut();
-        this.get_or_create_id(lock_op, lock_layout, offset, |ecx| &mut ecx.machine.sync.condvars)?
-            .ok_or_else(|| err_ub_format!("condvar has invalid ID").into())
+        this.get_or_create_id(
+            lock_op,
+            lock_layout,
+            offset,
+            |ecx| &mut ecx.machine.sync.condvars,
+            |_| Ok(Default::default()),
+        )?
+        .ok_or_else(|| err_ub_format!("condvar has invalid ID").into())
     }
 
     #[inline]
--- a/src/shims/unix/sync.rs
+++ b/src/shims/unix/sync.rs
@@ -74,6 +74,8 @@ fn mutex_id_offset<'tcx>(ecx: &MiriInterpCx<'tcx>) -> InterpResult<'tcx, u64> {
 
     // Sanity-check this against PTHREAD_MUTEX_INITIALIZER (but only once):
     // the id must start out as 0.
+    // FIXME on some platforms (e.g linux) there are more static initializers for
+    // recursive or error checking mutexes. We should also add thme in this sanity check.
     static SANITY: AtomicBool = AtomicBool::new(false);
     if !SANITY.swap(true, Ordering::Relaxed) {
         let static_initializer = ecx.eval_path(&["libc", "PTHREAD_MUTEX_INITIALIZER"]);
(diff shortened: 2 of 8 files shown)


Mandragorian:
Hi! 

Just wanted to quickly check if this fell through the cracks. (But no pressure if you are busy!). Thanks!

RalfJung:
It's in my list. :) Sorry for the slow reply. Too many things that need to be done...

src/shims/unix/sync.rs
RalfJung, line 1: One goal of the "detect moves" scheme is to no longer store the mutex kind inside `pthread_mutex_t`. So this isnt quite there yet.

So here's what I would suggest:
- We keep storing the mutex ID inside the `pthread_mutex_t`. So `mutex_id_offset` remains.
- `mutex_kind_offset` is removed, we handle this differently.
- `pthread_mutex_init` works by storing `libc::PTHREAD_MUTEX_INITIALIZER` to `mutex_op`, then calls `mutex_get_id`. Everything else just calls `mutex_get_id`.
- I like giving `mutex_get_or_create_id` a closure to initialize the "extra" data for a new mutex; keep that. 
- `mutex_get_id` should then inside that closure determine the initial mutex kind. (This requires `ecx` to be passed to the closure.)  And that should work basically as follows: check whether the mutex is byte-for-byte equal to `PTHREAD_MUTEX_INITIALIZER`, and if so then set the kind to `PTHREAD_MUTEX_DEFAULT`. Else, if we are on Linux, if the value is byte-for-byte equal to `PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP`, set the kind to `PTHREAD_MUTEX_RECURSIVE`. Else, error. (There may be more initializers we have to add, the test suite should tell.)
- `mutex_get_id` should also check that the mutex ID it got has the right address -- `mutex_check_moved` should not be a separate function. That avoids the risk of forgetting to call `mutex_check_moved`.

src/concurrency/sync.rs
```diff
@@ -66,6 +66,30 @@ pub(super) use declare_id;
...
+}
+
+#[derive(Debug)]
+/// Additional data that may be used by shim implementations.
+pub struct AdditionalMutexData {
+    /// The mutex kind, used by some mutex implementations like pthreads mutexes.
+    pub kind: MutexKind,
+
+    /// The address of the mutex.
+    pub address: Pointer,
```
RalfJung, line 90: Since we really only care about the address, this should be a `u64`. Currently you are also storing provenance but we don't actually care about that.

src/shims/unix/sync.rs
```diff
@@ -118,13 +118,29 @@ fn mutex_get_id<'tcx>(
     ecx: &mut MiriInterpCx<'tcx>,
     mutex_op: &OpTy<'tcx>,
 ) -> InterpResult<'tcx, MutexId> {
+    let address = ecx.deref_pointer_as(mutex_op, ecx.libc_ty_layout("pthread_mutex_t"))?.ptr();
```
RalfJung, line 121:
```suggestion
    // FIXME: might be worth changing mutex_get_or_create_id to take the mplace
    // rather than the `OpTy`.
    let address = ecx.deref_pointer_as(mutex_op, ecx.libc_ty_layout("pthread_mutex_t"))?.ptr();
```

RalfJung (commented):
Thanks for giving this a shot! I like the `AdditionalMutexData` part, but the way it gets used in `unix/sync` should be changed a bit -- see the comments in the review.

src/shims/unix/sync.rs
Mandragorian: Thanks for the comment! I have one thing that confuses me. What should happen with the attribute argument in `pthread_mutex_init`? If it just stores `PTHREAD_MUTEX_INITIALIZER` then the `attr` argument is ignored. 

If i understand correctly it must be valid for a recursive mutex to be created with `pthread_mutex_init`, and not just the static initializer? Maybe storing the corresponding static initializer based on the `attr` argument will take care of this? (maybe this is what you meant now that I reread the comment).

RalfJung: Ah, good point. I forgot about that part. `get_or_create_id` should probably have a way of calling it where it will always create a new ID, and not look at the old memory at all. Not sure what the best way is to set that, maybe just a `force_create: bool`?

`pthread_mutex_init` should read the mutex kind from the attribute argument, and pass it to `mutex_get_or_create_id`. So `mutex_get_id` should take an `init_kind: Option<i32>` argument for the mutex kind. If that is `Some`, then we set `force_create` to true and inside the closure that determines the initial mutex kind, we use the one in init_kind.

This also means `pthread_mutex_init` can overwrite the contents of its `pthread_mutex_t` argument with "uninit" rather than putting `PTHREAD_MUTEX_DEFAULT` there.

RalfJung:
(please write @`rustbot ready` when this is ready for review again)

Mandragorian:
Yeah, I wanted to post a comment first with a few things I am not sure are correct. I realized it was late after I pushed the changes so I decided to do that today. sorry if it caused confusion. Here is what I wanted to say:

To check for the static initializers I have to ignore a "header" in the `pthread_mutex_t` where miri stores the `MutexId`. The way the code works now, initialization of the additional data happens lazily, after the compare exchange operation. So at that point the mutex id is stored, and doing a bitwise comparison with the static initializers will fail. I could call the closure before checking for initialization in `get_or_create_id`, but then this would happen every time a mutex is locked/unlocked which could be slow. So i decided to ignore `4 + mutex_id_offset` bytes from the contents of a mutex when checking if it is equal to a static initializer.

This also solves another issue. When i was reading the full contents of the mutex, miri thought there was a race condition. I did not spend too much time debugging this. But I believe it is because when trying to lock, for example, a mutex that was initialized with a static initializer, in thread A, all the bytes of `pthread_mutex_t` (including the id) would be read. Then thread B would try to also lock the mutex, and call `get_or_create_id`. The previous unsynchronized read then clashed with the atomic read in `compare_exchange`. I couldn't find the proper way to fix this. I think I might need to somehow use `acquire_clock` but I couldn't exactly understand how it works. Since I am now ignoring the mutex id part, I think this sidesteps the issue.

Finally, instead of adding a `forece_create` boolean argument, I opted for defining a new function that creates the mutex eagerly. I thought that adding more logic in `get_or_create_id` might make it too complex. But I a have no strong feelings about this. The current way also removes the need for `mutex_reset_id`. Otherwise, I think we need to call it to initialize the mutex id memory, otherwise miri complains that there was a read of uninit data, I am assuming at `compare_exchange`.

@rustbot ready

RalfJung:
> To check for the static initializers I have to ignore a "header" in the pthread_mutex_t where miri stores the MutexId. The way the code works now, initialization of the additional data happens lazily, after the compare exchange operation. So at that point the mutex id is stored, and doing a bitwise comparison with the static initializers will fail. I could call the closure before checking for initialization in get_or_create_id, but then this would happen every time a mutex is locked/unlocked which could be slow. So i decided to ignore 4 + mutex_id_offset bytes from the contents of a mutex when checking if it is equal to a static initializer.

Ah yes, this has to be done quite carefully to avoid races.

I'd say for now just scratch this sanity check. Checking just some of the bytes is a bit too ugly IMO, after all we don't *have* to catch these kinds of bugs anyway. So let's just use the "zero ID" as indicator that this must be initialized and not do a separate check for now; we can explore how to add such a check separately in the future.

src/concurrency/sync.rs
```diff
@@ -66,6 +66,30 @@ pub(super) use declare_id;
 
 declare_id!(MutexId);
 
+/// The mutex kind.
```
RalfJung, line 69:
```suggestion
/// The mutex kind.
/// The meaning of this type is defined by concrete mutex implementations.
```

src/shims/unix/sync.rs
```diff
@@ -90,79 +90,120 @@ fn mutex_id_offset<'tcx>(ecx: &MiriInterpCx<'tcx>) -> InterpResult<'tcx, u64> {
...
-    }
-
-    offset
+/// Eagerly create and initialize a new mutex.
+fn mutex_create<'tcx>(
+    ecx: &mut MiriInterpCx<'tcx>,
+    mutex_op: &OpTy<'tcx>,
+    kind: i32,
+) -> InterpResult<'tcx> {
+    let address = ecx.read_pointer(mutex_op)?.addr().bytes();
```
RalfJung, line 99:
```suggestion
    // FIXME: might be worth changing mutex_create to take the mplace
    // rather than the `OpTy`.
    let address = ecx.read_pointer(mutex_op)?.addr().bytes();
```

src/shims/unix/sync.rs
```diff
@@ -90,79 +90,120 @@ fn mutex_id_offset<'tcx>(ecx: &MiriInterpCx<'tcx>) -> InterpResult<'tcx, u64> {
...
+    )?;
+    ecx.read_bytes_ptr_strip_provenance(
+        ptr,
+        ecx.libc_ty_layout("pthread_mutex_t").layout.size() - header_size,
+    )
 }
 
-fn mutex_set_kind<'tcx>(
+/// Checks if a mutex has been moved since its first use and returns an error if it has.
+fn mutex_check_moved<'tcx>(
```
RalfJung, line 195: This is only called once, please inline it at the call site.

src/shims/unix/sync.rs
```diff
@@ -90,79 +90,120 @@ fn mutex_id_offset<'tcx>(ecx: &MiriInterpCx<'tcx>) -> InterpResult<'tcx, u64> {
...
-    ecx.mutex_get_or_create_id(
+    // FIXME: might be worth changing mutex_get_or_create_id to take the mplace
+    // rather than the `OpTy`.
+    let id = ecx.mutex_get_or_create_id(
         mutex_op,
         ecx.libc_ty_layout("pthread_mutex_t"),
         mutex_id_offset(ecx)?,
-    )
+        |ecx| {
+            let address = ecx.read_pointer(mutex_op)?.addr().bytes();
```
RalfJung, line 121:
```suggestion
            // This is called if a static initializer was used and the lock has not been assigned
            // an ID yet. We have to determine the mutex kind from the static initializer.
            let address = ecx.read_pointer(mutex_op)?.addr().bytes();
```

src/shims/unix/sync.rs
```diff
@@ -90,79 +90,120 @@ fn mutex_id_offset<'tcx>(ecx: &MiriInterpCx<'tcx>) -> InterpResult<'tcx, u64> {
...
-    ecx.deref_pointer_and_write(
-        mutex_op,
-        mutex_id_offset(ecx)?,
-        Scalar::from_u32(0),
-        ecx.libc_ty_layout("pthread_mutex_t"),
-        ecx.machine.layouts.u32,
-    )
+/// Returns the kind of a mutex, based on its contents.
+/// The contents must not contain the initial "header" where the mutex id is stored.
+fn kind_from_contents<'tcx>(
```
RalfJung, line 136: `kind_from_static_initializer` or so would be better IMO, making it clear that we're parsing target-specific constants here but never our own data.

src/shims/unix/sync.rs
```diff
@@ -90,79 +90,120 @@ fn mutex_id_offset<'tcx>(ecx: &MiriInterpCx<'tcx>) -> InterpResult<'tcx, u64> {
     Ok(offset)
```
RalfJung, line 90: Can you add a FIXME to the sanity check logic above that we should also check the other static initializers for platforms where they exist?

src/shims/unix/sync.rs
```diff
@@ -90,79 +90,120 @@ fn mutex_id_offset<'tcx>(ecx: &MiriInterpCx<'tcx>) -> InterpResult<'tcx, u64> {
...
+    ecx: &MiriInterpCx<'tcx>,
+    contents: &[u8],
+) -> InterpResult<'tcx, MutexKind> {
+    let on_linux = &*ecx.tcx.sess.target.os == "linux";
+
+    let default_initializer =
+        ecx.eval_path(&["libc", "PTHREAD_MUTEX_INITIALIZER"]).assert_mem_place().ptr();
+    if contents == get_contents(ecx, default_initializer)? {
+        return Ok(MutexKind::new(ecx.eval_libc_i32("PTHREAD_MUTEX_DEFAULT")));
+    }
```
RalfJung, line 146: This still makes the assumption that the platform-specific field that encodes the mutex kind in the static initializer is after the "header". So it's not enough to be sure we are matching platform behavior, and it's not clean enough IMO to keep it as a half-measure.

I also just realized that this function is also called *after* the new ID has been written, so a full comparison with `pthread_mutex_t` would anyway not work.

So I think the best option is to have a `match &*ecx.tcx.sess.target.os` inside this function and then determine for each platform what we have to read to find the initial mutex kind -- basically exactly what `mutex_get_kind` did before this PR, but only doing it here when we "parse" a static initializer, rather than doing it each time the mutex is used. On Linux this would read 4 bytes at offset `if ecx.pointer_size().bytes() == 8 { 16 } else { 12 }`. For the other supported targets we'll have to check whether they have any static initializers other than PTHREAD_MUTEX_DEFAULT. And if any platform adds such initializers in the future, Miri will just ignore them until we realize that... maybe one day we can do the comparison with `PTHREAD_MUTEX_DEFAULT` but it will require reworking `get_or_create_id` a bit more so that `create_obj` is called before the new ID is written.

src/shims/unix/sync.rs
```diff
@@ -573,8 +623,6 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             throw_ub_format!("destroyed a locked mutex");
         }
 
-        // Destroying an uninit pthread_mutex is UB, so check to make sure it's not uninit.
```
RalfJung, line 576: Please preserve this comment, it is still relevant.

RalfJung (commented):
Having a separate `create_id` leads to some duplication in the API surface -- we'll need this for all types that have static and explicit initializers. But on the other hand, having the "explicit initialization" separate from "initialize from static initializer" is also nice, so yeah let's go with that.

src/shims/unix/sync.rs
RalfJung: I checked and none of the other targets has any other static initializers.

RalfJung (commented):
@rustbot author

src/shims/unix/sync.rs
Mandragorian: Is this still needed after you checked that there are no other other initializers?

src/shims/unix/sync.rs
Mandragorian: done. I am not sure I interpreted all of your comment correctly, so let me know.

src/shims/unix/sync.rs
RalfJung: Yeah it needs to be checked for Linux.

src/shims/unix/sync.rs
```diff
@@ -124,79 +124,40 @@ fn mutex_get_id<'tcx>(
...
-    contents: &[u8],
+    mutex_op: &OpTy<'tcx>,
 ) -> InterpResult<'tcx, MutexKind> {
-    let on_linux = &*ecx.tcx.sess.target.os == "linux";
-
-    let default_initializer =
-        ecx.eval_path(&["libc", "PTHREAD_MUTEX_INITIALIZER"]).assert_mem_place().ptr();
-    if contents == get_contents(ecx, default_initializer)? {
+    // Only linux has static initializers other than PTHREAD_MUTEX_DEFAULT.
+    if &*ecx.tcx.sess.target.os != "linux" {
```
RalfJung, line 148: Please make this an exhaustive match, like the old `mutex_get_kind`. When we add a new target we need to check that this is still true.

src/shims/unix/sync.rs
```diff
@@ -124,79 +124,40 @@ fn mutex_get_id<'tcx>(
...
-    )
+    let offset = if ecx.pointer_size().bytes() == 8 { 16 } else { 12 };
+    let kind = ecx.deref_pointer_and_read(
+        mutex_op,
+        offset,
+        ecx.libc_ty_layout("pthread_mutex_t"),
+        ecx.machine.layouts.i32,
+    )?
+    .to_i32()?;
+    return Ok(MutexKind::new(kind));
```
RalfJung, line 161: Please check that `kind` is a valid mutex kind.

In fact, probably it would make sense to make this an enum? Then all the other mutex functions won't have to compare this with libc constants any more...

Mandragorian: re making this an enum:

I don't disagree, but I am a bit unsure on how to do this. In my mind ideally the `MutexKind` would be a type known only to the pthreads implementation, but currently it lives with the generic mutex implementation. At least by making this an `i32` it would allow other mutex APIs other than pthreads to possibly reuse it. However if we make it an enum it might be too coupled with pthreads terminology.

As the code is now, we can't make the `AdditionalMutexData` generic so that each mutex implementation can define its own type to hold whatever data they want.

On the other hand this might be abstracting way to soon since there is no other implementation that needs additional data. What do you think?

RalfJung: > As the code is now, we can't make the AdditionalMutexData generic so that each mutex implementation can define its own type to hold whatever data they want.

We could use `dyn Any` trickery.

But IMO it's not worth it. We can just say the enum lists all  mutex kinds of all implementations. :shrug: You can mark the enum `#[non_exhaustive]` to drive home that point, and force all matches to end with `_ => panic!("pthread mutex with invalid mutex kind")` or so.

src/concurrency/sync.rs
```diff
@@ -69,16 +69,13 @@ declare_id!(MutexId);
 /// The mutex kind.
 /// The meaning of this type is defined by concrete mutex implementations.
```
RalfJung, line 70: That line of the comment seems outdated now.

tests/fail-dep/concurrency/libc_pthread_mutex_move.rs
```diff
@@ -0,0 +1,72 @@
+//@ignore-target-windows: No pthreads on Windows
+//@[unlock_register]only-target-linux: recursive initializers are non-standard.
+//@revisions: lock trylock unlock_register unlock_detect init
```
RalfJung, line 3: Please declare the revisions before using them.

Also, I am confused by the names. `lock` really is `static_initializer`, right? `init` is `initializer_function`. And why is `unlock_register` only for Linux? Recursive mutexes exist on all platforms, they just don't have a static initializer on non-Linux.

Maybe it just needs comments explaining what these variants are testing, but currently I am quite confused. I expected only 2 tests -- one that uses the static initializer and one that uses `pthread_mutex_init`.

Mandragorian: Regarding the tests:

In my mind what needs to be tested is the ability of the `pthread_mutex_*` functions to record and then later detect a moved `pthread_mutex_t`. So for each such function I call it two times. If the move is caught it means that the function both recorded the original address and detected the move. At least that's the intention.

For the `unlock` function however I needed a slight different approach. Static initializers are only supported on linux, and `unlock` can only be called on recursive mutexes if they are unlocked. So I used two different revisions, one to test that `unlock` correctly registers the address (with a static initializer), and one to test that it detects a move, when called after init.

RalfJung: I see, thanks.

However, given our implementation, we know this is all the same codepath, so it doesn't seem worth testing *all* combinations.

Mandragorian: ack, removed the extra tests.

tests/fail-dep/concurrency/libc_pthread_mutex_move.rs
```diff
@@ -17,54 +16,13 @@ fn check() {
...
 }
 
-#[cfg(lock)]
+#[cfg(static_initializer)]
 fn check() {
     unsafe {
         let mut m: libc::pthread_mutex_t = libc::PTHREAD_MUTEX_INITIALIZER;
         libc::pthread_mutex_lock(&mut m as *mut _);
 
         let mut m2 = m;
```
RalfJung, line 25:
```suggestion
        let mut m2 = m; // move the mutex
```

tests/fail-dep/concurrency/libc_pthread_mutex_move.rs
```diff
@@ -0,0 +1,28 @@
...
+    check();
+}
+
+#[cfg(init)]
+fn check() {
+    unsafe {
+        let mut m: libc::pthread_mutex_t = std::mem::zeroed();
+        assert_eq!(libc::pthread_mutex_init(&mut m as *mut _, std::ptr::null()), 0);
+
+        let mut m2 = m;
```
RalfJung, line 14:
```suggestion
        let mut m2 = m; // move the mutex
```

RalfJung (commented):
Thanks! LGTM, just two comment nits. Then please squash the commits.

RalfJung:
Thanks. :-)

@bors r+

RalfJung:
I only just noticed that you even suggested this initially. :D

> Store the additional data as Box\<dyn Any> and the implementations can downcast their data when they fetch them.

If you want to do this in a follow-up PR, let me know. :)

PR #3787, opened 2024-08-05
Add hook for updating the canonical address

dheaton-arm:
Where a method modifies the address that should be considered canonical - such as via TBI or when an MTE tag has been set - Miri will need to be notified.

This adds a hook to inform Miri of the new address.

---

The implementation used here permits a single (mutable) alias, with the intent that user code will only use the alias (once set) rather than the original `base_addr`. I'm open to feedback on this approach, but ran into issues when trying to update the actual `base_addr` when popping the function off the stack (and thus deallocating anything created in the function), hence this implementation which allows that to work correctly.

In particular, this would allow https://github.com/rust-lang/stdarch/pull/1622 to work under Miri, as it could use this hook to ensure that Miri is aware that the change to the pointer address is actually fine in context.

### DIFF
--- a/src/alloc_addresses/mod.rs
+++ b/src/alloc_addresses/mod.rs
@@ -460,6 +460,21 @@ impl<'tcx> MiriMachine<'tcx> {
             }
         })
     }
+
+    /// Updates the current canonical address for the allocation. Note that any access to this allocation *must* use this address from this point on.
+    pub fn set_alloc_address(&mut self, id: AllocId, new: u64) {
+        let global_state = self.alloc_addresses.get_mut();
+        if let Some(addr) = global_state.base_addr.insert(id, new) {
+            // Remove the old address' int->ptr mapping.
+            let pos =
+                global_state.int_to_ptr_map.binary_search_by_key(&addr, |(addr, _)| *addr).unwrap();
+            let removed = global_state.int_to_ptr_map.remove(pos);
+            assert_eq!(removed, (addr, id));
+        }
+        let new_pos =
+            global_state.int_to_ptr_map.binary_search_by_key(&new, |(addr, _)| *addr).unwrap_err();
+        global_state.int_to_ptr_map.insert(new_pos, (new, id));
+    }
 }
 
 #[cfg(test)]
--- a/src/shims/foreign_items.rs
+++ b/src/shims/foreign_items.rs
@@ -423,6 +423,13 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     }
                 }
             }
+            "miri_set_canonical_address" => {
+                let [old_ptr, new_ptr] = this.check_shim(abi, Abi::Rust, link_name, args)?;
+                let old_ptr = this.read_pointer(old_ptr)?;
+                let new_ptr = this.read_pointer(new_ptr)?;
+                let (alloc_id, _, _) = this.ptr_get_alloc_id(old_ptr, 0)?;
+                this.machine.set_alloc_address(alloc_id, new_ptr.addr().bytes());
+            }
 
             // Aborting the process.
             "exit" => {
--- a/tests/utils/miri_extern.rs
+++ b/tests/utils/miri_extern.rs
@@ -147,4 +147,9 @@ extern "Rust" {
     /// "symbolic" alignment checks. Will fail if the pointer is not actually aligned or `align` is
     /// not a power of two. Has no effect when alignment checks are concrete (which is the default).
     pub fn miri_promise_symbolic_alignment(ptr: *const (), align: usize);
+
+    /// Miri-provided extern function to specify that a new address is to be considered the
+    /// canonical address, where `new` is a valid alias to the `old` allocation,
+    /// usually due to them having different values in bits that are ignored by hardware.
+    pub fn miri_set_canonical_address(old: *const (), new: *const ());
 }


RalfJung:
I'm sorry I am a bit confused here, could you explain the problem this is trying to solve?

In Rust it is generally *not correct* to have an allocation be accessed via two different addresses. The existing t-opsem discussions for pointer tagging schemes concluded that these schemes are fine if we can just pretend that *the full pointer with the tag* is the actual address that the allocation is located at. However, accessing the same allocation with pointers that use different tags is equivalent to accessing physical memory that is mapped into the virtual address space multiple times, and that's something Rust doesn't really support currently (and AFAIK neither doses LLVM).

RalfJung:
Maybe `miri_set_canonical_address` should be considered a `realloc`-like operation, in which case it could move the memory to a new place. However, in that case it's very suspicious that this would only be done inside Miri -- such a realloc would have to occur in real executions as well, or else there is UB.

This seems to me like something that needs opsem discussions, so maybe bring this up [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/136281-t-opsem) or at https://github.com/rust-lang/unsafe-code-guidelines/ ?

dheaton-arm:
> I'm sorry I am a bit confused here, could you explain the problem this is trying to solve?

It's currently not possible to run code which modifies the top byte of a pointer (whether via TBI or MTE) under Miri, as Miri will consider the pointer to be dangling on dereference. Given that rustc itself does not currently have any issues (in practice) with this, and the only issues arise when running under Miri, the intent is to allow for updating the address Miri considers an allocation to live at.

> The existing t-opsem discussions for pointer tagging schemes concluded that these schemes are fine if we can just pretend that _the full pointer with the tag_ is the actual address that the allocation is located at.

I [asked on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/242906-t-compiler.2Farm/topic/mte.2Ftbi.20accessors) for advice regarding adding TBI accessors and MTE intrinsics, and was referred to discussions which - to my reading - concluded what you describe.
That is fine, but I'm uncertain as to how to fully change the canonical address (ie, the actual address that Miri expects from that point on) - changing its entry in the pointer maps (`base_addr`, etc) works but the 'old' pointer seems to be held in a stack frame (which causes issues when said stack frame is popped off and all allocations are deallocated, in this case using the wrong/old pointer), which I haven't been able to work out, hence this implementation.
I'm happy to discuss this more if necessary, also.

> Maybe miri_set_canonical_address should be considered a realloc-like operation

Yes, that would be the intent, I'm just not sure how best to handle this as `realloc`, for example, generates an address without the capability to set the top byte.

> Rust doesn't really support currently (and AFAIK neither doses LLVM).

Could you elaborate on what you mean by this, sorry? The code that this would enable to run under Miri currently already runs under Rust, and MTE has, at least, been generally supported by LLVM for some time now; in the case of TBI, the only issues I'm aware of are that some operations (such as `memcpy`) are unsafe between aliasing pointers as the aliasing analysis fails, but otherwise it works.

But, in short, I would be happy to replace the current approach which uses a single, mutable, alias with actually changing the address that the allocation is considered to live at, as you describe - but advice on where such a change needs to take place would be appreciated, as it's not clear to me.

dheaton-arm:
So, in general, anything that looks like the following is UB:
```rust
// let pointer be some pointer to an allocation
let addr = pointer.addr() & 0x00ffffffffffffff;
let p = addr | ((0x70 as usize) << 56);
let pointer = pointer.with_addr(p);
// dereferencing `pointer` will now always trip Miri.
```
Of course, generally, we're happy to accept that this should be considered UB, but we were therefore hoping that by providing a Rust-controlled mechanism to do this instead (https://github.com/rust-lang/stdarch/pull/1622), the UB could be mitigated as Rust/Miri can track the address change. That's the intent of this work, as Miri raises an error when dereferencing a pointer from `with_top_byte` from that PR, for example.

Put another way, the `with_top_byte` helper in that PR should be considered itself to be the 'kind of `realloc`' that this PR introduces, and indeed the plan would be to have that method call this PR's hook. I did investigate to see whether I needed to adjust rustc for that PR, but couldn't find any instance of rustc tracking such allocations at this time, and hence the focus on making it work within Miri. If rustc started tracking such allocations later down the line, then that be the function that would need to be responsible for updating that, too. Do you disagree with that approach?

RalfJung:
> So, in general, anything that looks like the following is UB:

Can you provide a small program with a `main` function showing how this is used? I am still very confused about the entire context here.

RalfJung:
> I did investigate to see whether I needed to adjust rustc for that PR, but couldn't find any instance of rustc tracking such allocations at this time, and hence the focus on making it work within Miri. If rustc started tracking such allocations later down the line, then that would be the function that would need to be responsible for updating that, too. Do you disagree with that approach?

LLVM tracks allocations, so this definitely needs changes in rustc and maybe LLVM to make them aware of the `realloc`. (The same problem exists when doing this in C, btw. If you are just porting C code here it seems likely that the original code was written without much regard for whether that is UB or not.)

RalfJung:
To give one example, if you do something like this
```rust
fn tag_ptr(ptr: *const i32, tag: u8) -> *const i32 {
  ptr.map_addr(|a| {
    (a & 0x00ffffffffffffff) | ((tag as usize) << 56)
  })
}

let my_local_var = 42;
tag_ptr(&my_local_var, 0x70).read();
tag_ptr(&my_local_var, 0x80).read();
```
Then LLVM might be able to see that this pointer points to 4 bytes of memory and it has been offset in two different ways that are way more than 4 bytes apart, so one of these reads *must* be out-of-bounds, and therefore this program has UB.

I don't know if LLVM has any optimizations today that may actually detect this (@nikic and @comex might know), but it would totally be in its right to start doing analyses like that any time, so we can't just say "yeah let's accept such code and hope nothing goes wrong".

RalfJung:
Here's an example where LLVM seems more likely to notice and cause trouble:
```rust
fn tag_ptr(ptr: *mut i32, tag: u8) -> *mut i32 {
  ptr.map_addr(|a| {
    (a & 0x00ffffffffffffff) | ((tag as usize) << 56)
  })
}

let mut my_local_var = 0;
tag_ptr(&my_local_var, 0x70).write(1);
tag_ptr(&my_local_var, 0x80).write(2);
assert_eq!(
  tag_ptr(&my_local_var, 0x70).read(),
  2
);
```
From LLVMs perspective it should not be too hard to notice that the two writes cannot alias, and therefore the final read must return what was written by the first write.

mrkajetanp:
Which way around should this be done then? Since the TBIBox will require this hook before it gets merged to avoid breaking miri I was thinking we should focus on merging this first. Should we then finalise [stdarch#1622](https://github.com/rust-lang/stdarch/pull/1622) first, then finalise & merge this and then add the hooks to TBIBox and merge that?

RalfJung:
Which existing code breaks if that stdarch PR lands? It's new code, it's okay if Miri does not support it from the start.

First a design needs to be proposed for how TbiBox's API looks and how it is implemented. Once t-libs-api and t-libs are happy with that, we can look at how it should be supported in Miri. (t-opsem should likely also be involved there, due to the nature of this type.)

mrkajetanp:
That approach makes sense to me as well, fair enough, thanks!

RalfJung:
I'll close this for now as work on this will have to start elsewhere in the project, with an RFC: https://github.com/rust-lang/rfcs/pull/3700. It's probably worth opening a new PR when the RFC has been accepted as most of the discussion above is about the high-level conceptual points covered by the RFC.

Thanks for your work on this!

PR #3795, opened 2024-08-07
throw_unsup_format for alignment greater than 2^29

### LINKED ISSUE #3687: Attempting to allocate with alignment greater than 2^29 ICEs Miri
This program (reduced from https://crates.io/crates/ialloc)
```rust
fn main() {
    let layout = std::alloc::Layout::from_size_align(1, 1 << 30).unwrap();
    unsafe {
        std::alloc::alloc(layout);
    }
}
```
Will ICE with
```
thread 'rustc' panicked at src/tools/miri/src/shims/foreign_items.rs:475:50:
called `Result::unwrap()` on an `Err` value: `1073741824` is too large
```
This limit is deliberate:
https://github.com/rust-lang/rust/blob/5c8459f1eceba84dff8d622768dae556ea7c2495/compiler/rustc_abi/src/lib.rs#L699-L700
```rust
    // LLVM has a maximal supported alignment of 2^29, we inherit that.
    pub const MAX: Align = Align { pow2: 29 };
```
We are not LLVM, so I think this limit is simply wrong for Miri. Should we change the struct in the compiler?

tiif:
Fixes #3687

### DIFF
--- a/src/shims/foreign_items.rs
+++ b/src/shims/foreign_items.rs
@@ -8,7 +8,7 @@ use rustc_middle::mir;
 use rustc_middle::ty;
 use rustc_span::Symbol;
 use rustc_target::{
-    abi::{Align, Size},
+    abi::{Align, AlignFromBytesError, Size},
     spec::abi::Abi,
 };
 
@@ -199,9 +199,20 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
         if i128::from(size) > this.tcx.data_layout.pointer_size.signed_int_max() {
             throw_ub_format!("creating an allocation larger than half the address space");
         }
-        if !align.is_power_of_two() {
-            throw_ub_format!("creating allocation with non-power-of-two alignment {}", align);
+        if let Err(e) = Align::from_bytes(align) {
+            match e {
+                AlignFromBytesError::TooLarge(_) => {
+                    throw_unsup_format!(
+                        "creating allocation with alignment {align} exceeding rustc's maximum \
+                         supported value"
+                    );
+                }
+                AlignFromBytesError::NotPowerOfTwo(_) => {
+                    throw_ub_format!("creating allocation with non-power-of-two alignment {align}");
+                }
+            }
         }
+
         Ok(())
     }
 
--- a/tests/fail/alloc/unsupported_big_alignment.rs
+++ b/tests/fail/alloc/unsupported_big_alignment.rs
@@ -1,0 +1,14 @@
+// Previously, attempting to allocate with an alignment greater than 2^29 would cause miri to ICE
+// because rustc does not support alignments that large.
+// https://github.com/rust-lang/miri/issues/3687
+
+extern "Rust" {
+    fn __rust_alloc(size: usize, align: usize) -> *mut u8;
+}
+
+fn main() {
+    unsafe {
+        __rust_alloc(1, 1 << 30);
+        //~^ERROR: exceeding rustc's maximum supported value
+    }
+}
--- a/tests/fail/alloc/unsupported_big_alignment.stderr
+++ b/tests/fail/alloc/unsupported_big_alignment.stderr
@@ -1,0 +1,14 @@
+error: unsupported operation: creating allocation with alignment ALIGN exceeding rustc's maximum supported value
+  --> $DIR/unsupported_big_alignment.rs:LL:CC
+   |
+LL |         __rust_alloc(1, 1 << 30);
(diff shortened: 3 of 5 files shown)


src/shims/foreign_items.rs
```diff
@@ -463,6 +463,11 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     let size = this.read_target_usize(size)?;
                     let align = this.read_target_usize(align)?;
 
+                    // In rustc, the maximal supported alignment is 2^29.
+                    if align > 2u64.checked_pow(29).unwrap() {
+                        throw_unsup_format!("Alignment exceeded 2^29 is unsupported");
+                    }
+
                     this.check_rustc_alloc_request(size, align)?;
```
RalfJung, line 471: Please add the check in check_rustc_alloc_request instead so that it is run on all entry points (you forgot reallocate).

Also, please add a test.

RalfJung (commented):
Thanks for the PR!

src/shims/foreign_items.rs
RalfJung: Also is there a way to not hard-code the max value but use the same constant as rustc?

tiif: > Also is there a way to not hard-code the max value but use the same constant as rustc?

Yup, we can check from the error type returned by ``Align::from_bytes`` https://github.com/rust-lang/rust/blob/8d0066922b14cee3175e8c141e5e909fa6d9a6eb/compiler/rustc_abi/src/lib.rs#L731

tiif:
It is a bit surprising that I couldn't use this test and ``./miri test --bless`` it: 
```rust
fn main() {
    let layout = std::alloc::Layout::from_size_align(1, 1 << 30).unwrap();
    unsafe {
        let _a = std::alloc::alloc(layout);
    }
}
```
I guess the reason is the ``throw_unsup_format`` is not directly annotated on the test file (ie ``./tests/fail/alloc/unsupported_alignment.rs:7:18``)

<details>

<summary>This is how the original error looks like</summary>

```bash 
error: unsupported operation: creating allocation with alignment exceeding rustc's maximum supported value
  --> /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/alloc/src/alloc.rs:98:9
   |
98 |         __rust_alloc(layout.size(), layout.align())
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ creating allocation with alignment exceeding rustc's maximum supported value
   |
   = note: BACKTRACE:
   = note: inside `std::alloc::alloc` at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/alloc/src/alloc.rs:98:9: 98:52
note: inside `main`
  --> ./tests/fail/alloc/unsupported_alignment.rs:7:18
   |
7  |         let _a = std::alloc::alloc(layout);
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^

Error: command exited with non-zero code `cargo +miri --quiet run --manifest-path /home/byt/Documents/miri/Cargo.toml -- --edition 2021 --sysroot /home/byt/.cache/miri ./tests/fail/alloc/unsupported_alignment.rs`: 1
```
</details>

So I just follow what is done in ``./tests/fail/alloc/too_large`` instead, using ``__rust_alloc`` directly.

src/shims/foreign_items.rs
```diff
@@ -202,6 +202,13 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
         if !align.is_power_of_two() {
             throw_ub_format!("creating allocation with non-power-of-two alignment {}", align);
         }
+        if let Err(AlignFromBytesError::TooLarge(_)) = Align::from_bytes(align) {
+            throw_unsup_format!(
+                "creating allocation with alignment exceeding rustc's maximum \
+                 supported value"
+            );
```
tiif, line 209: I deliberately didn't use the exact value ``2^29``here because it would be incorrect if rustc were to change this limit value one day.

src/shims/foreign_items.rs
```diff
@@ -202,6 +202,13 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
         if !align.is_power_of_two() {
             throw_ub_format!("creating allocation with non-power-of-two alignment {}", align);
         }
+        if let Err(AlignFromBytesError::TooLarge(_)) = Align::from_bytes(align) {
```
RalfJung, line 205: Good idea to call `from_bytes`. Now you can even do the power-of-two checking that way, removing the manual check above. :)

RalfJung:
Thanks, looks good!

Please squash the commits.

PR #3798, opened 2024-08-09
miri-script: use --remap-path-prefix to print errors relative to the right root

RalfJung:
Inspired by https://github.com/rust-lang/rust-clippy/pull/13232, this makes it so that when cargo-miri fails to build, `./miri check` will print errors with paths like `cargo-miri/src/setup.rs`. That means we can get rid of the miri-symlink-hacks and instead tell RA to just always invoke the `./miri clippy` script just once, in the root.

This means that we can no longer share a target dir between cargo-miri and miri as the RUSTFLAGS are different to crates that are shared in the dependency tree need to be built twice with two different flags. `miri-script` hence now has to set the MIRI environment variable to tell the `cargo miri setup` invocation where to find Miri.

I also made it so that errors in miri-script itself are properly shown in RA, for which the `./miri` shell wrapper needs to set the right flags.

### DIFF
--- a/.github/workflows/setup/action.yml
+++ b/.github/workflows/setup/action.yml
@@ -35,6 +35,10 @@ runs:
         run: cargo install -f rustup-toolchain-install-master hyperfine
         shell: bash
 
+      - name: Install nightly toolchain
+        run: rustup toolchain install nightly --profile minimal
+        shell: bash
+
       - name: Install "master" toolchain
         run: |
           if [[ ${{ github.event_name }} == 'schedule' ]]; then
--- a/miri
+++ b/miri
@@ -1,8 +1,15 @@
 #!/usr/bin/env bash
 set -e
+# We want to call the binary directly, so we need to know where it ends up.
+MIRI_SCRIPT_TARGET_DIR="$(dirname "$0")"/miri-script/target
+# If stdout is not a terminal and we are not on CI, assume that we are being invoked by RA, and use JSON output.
+if ! [ -t 1 ] && [ -z "$CI" ]; then
+  MESSAGE_FORMAT="--message-format=json"
+fi
+# We need a nightly toolchain, for the `profile-rustflags` cargo feature.
+cargo +nightly build $CARGO_EXTRA_FLAGS --manifest-path "$(dirname "$0")"/miri-script/Cargo.toml \
+  -q --target-dir "$MIRI_SCRIPT_TARGET_DIR" $MESSAGE_FORMAT || \
+  ( echo "Failed to build miri-script. Is the 'nightly' toolchain installed?"; exit 1 )
 # Instead of doing just `cargo run --manifest-path .. $@`, we invoke miri-script binary directly. Invoking `cargo run` goes through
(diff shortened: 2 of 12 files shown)


RalfJung:
@bjorn3 [wrote](https://github.com/rust-lang/rust/pull/128904#issuecomment-2278774026)

> I think it fundamentally needs the wrapper to ensure that the remapping only applies to workspace members and not deps from crates.io, which have a different cwd. 

Oh... should we be concerned about that here for Miri? I don't want to add a RUSTC_WRAPPER, that would be way to heavy-weight IMO.

miri
```diff
@@ -1,8 +1,16 @@
...
 set -e
-# Instead of doing just `cargo run --manifest-path .. $@`, we invoke miri-script binary directly. Invoking `cargo run` goes through
-# rustup (that sets it's own environmental variables), which is undesirable.
+# We want to call the binary directly, so we need to know where it ends up.
 MIRI_SCRIPT_TARGET_DIR="$(dirname "$0")"/miri-script/target
-cargo +stable build $CARGO_EXTRA_FLAGS -q --target-dir "$MIRI_SCRIPT_TARGET_DIR" --manifest-path "$(dirname "$0")"/miri-script/Cargo.toml || \
+# If stdout is not a terminal, assume that we are being invoked by RA, and use JSON output.
+if ! [ -t 1 ]; then
+  MESSAGE_FORMAT="--message-format=json"
+fi
```
RalfJung, line 8: Ah, this heuristic isn't great as it will now also show things in JSON on CI...

RalfJung:
All right, this seems to be working now.

Also note that when this lands, you'll have to update your RA config.

RalfJung:
Oh, looking at the Miri thing again we could also use a totally different approach... namely, to set this as a per-package variable in the profile, which will also avoid applying it to dependencies.

.github/workflows/setup/action.yml
```diff
@@ -35,6 +35,10 @@ runs:
         run: cargo install -f rustup-toolchain-install-master hyperfine
         shell: bash
 
+      - name: Install nightly toolchain
+        run: rustup toolchain install nightly --profile minimal
+        shell: bash
```
saethlin, line 40: Why do we now need this?

saethlin: I suppose more than anything I'm confused that we didn't need this before.

RalfJung: We now need this because miri-script needs nightly because it picks up the `.config/cargo.toml` which uses nightly features.

Previously we didn't need this because miri-script was built with stable, and then it used rustup-toolchain-install-master to install the toolchain used for Miri.

saethlin: Ah.

saethlin:
> @rust-lang/miri what do you think?

No objections. I trip over the path prefixes more in the rust-lang/rust tree than this repo, but still a nice change.

RalfJung:
Okay let's land this then. :)
@bors r=saethlin

Issue #3665, opened 2024-06-10
Implement blocking support for eventfd and socketpair


RalfJung:
Our `socketpair` and `eventfd` implementations currently do not support blocking.

To fix that will require a bit of re-structuring to enable `FileDescription::{read,write}` to block. In that case we can't use the convenient `bytes: &mut [u8]`/`bytes: &[u8]` arguments, we need the underlying Miri place, so that we can store it in a callback to be used on unblocking. (Incidentally that would also simplify eventfd read/write a bit since we could use the normal memory APIs to reading/writing u64 in an endianess-aware way, rather than having to implement that on a byte slice.) The functions also need access to the return place so they can write the return value later, when they are unblocked.

Basically, a bunch of the convenience that is currently implemented [here](https://github.com/rust-lang/miri/blob/524de9f6d0841aec9212a610ea8f444b3822870a/src/shims/unix/fd.rs#L385-L404) can't be used when blocking is involved. We have to make read/write lower-level functions that look more like the usual Miri hooks, and possibly add some helper functions to make it still easy to implement the current behavior. (Those helper functions are then likely useful with other file system functions as well.)

oli-obk:
Alternatively we could implement it similar to `sleep`, but instead of waking on the next instruction, just reexecute the current one.

That does mean we cannot mutate any state before blocking or we'd get incorrect behavior

RalfJung:
`sleep` uses the same blocking machinery as everything else. We don't currently use the "re-execute instruction" approach for anything and TBH I think that is a bit too hacky for my taste. For instance if an intrinsic argument is `*ptr` then we don't want to re-load it from memory each time... though I guess if the value would change that would be a data race so it doesn't actually make a difference?

adwinwhite:
May I work on this issue? Or it belongs to a larger project?

tiif:
I think my work on #3448 might potentially block this. The current specification for epoll is only for non-blocking eventfd and socketpair. I wish to let that land then incorporate blocking mechanism to it.

adwinwhite:
I see. Thanks for letting me know. I'll look for something else then.

RalfJung:
FWIW we could have blocking socketpair without that being supported by epoll.

But also we currently don't have enough review capacity for another large projects like this. I can't even keep up with the PRs we already have. So picking something smaller is probably better.

tiif:
I'd like to work on this. 

@rustbot claim

RalfJung:
[...]

PR #3804, opened 2024-08-14
Support blocking for epoll

tiif:
This PR enabled epoll to have blocking operation. 

The changes introduced by this PR are:
- Refactored part of the logic in ``epoll_wait`` to ``blocking_epoll_callback`` 
- Added a new field ``thread_ids`` in ``Epoll`` for blocked thread ids
- Added a new ``BlockReason::Epoll``


RalfJung:
> place: &'tcx MPlaceTy<'tcx> is required to be annotated with 'tcx. 

This should be `place: MPlaceTy<'tcx>`. (Use `clone` if needed.) I have no idea how you even got a reference with `'tcx` lifetime...

That won't solve the issue of propagating the `'tcx` lifetime, but maybe it helps simplify things a bit.

src/shims/unix/linux/epoll.rs
```diff
@@ -166,25 +376,160 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                        }
+                        @unblock = |this| {
+                                    this.blocking_epoll_callback(epfd_value, place, &event);
+                                    Ok(())
+                                }
+                        @timeout = |this| {
+                        this.write_scalar(Scalar::from_i32(0), place)?;
+                                    Ok(())
+                                }
+                    ),
```
RalfJung, line 438: I know this is a WIP, but please do spend the effort of using proper indentation inside this macro. rustfmt can't format macros so this has to be done by hand; see the other `callback!` invocations for our usual style.

src/shims/unix/linux/epoll.rs
```diff
@@ -166,25 +376,160 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     fn epoll_wait(
         &mut self,
         epfd: &OpTy<'tcx>,
-        events: &OpTy<'tcx>,
+        events_op: &OpTy<'tcx>,
         maxevents: &OpTy<'tcx>,
         timeout: &OpTy<'tcx>,
-    ) -> InterpResult<'tcx, Scalar> {
+        place: &'tcx MPlaceTy<'tcx>,
```
RalfJung, line 382: If this is the `dest`  place, please call it `dest` so that the reader has some chance of understanding the purpose of the place.

src/shims/unix/linux/epoll.rs
```diff
@@ -477,4 +485,51 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         }
         Ok(())
     }
+
+    /// Callback function after epoll_wait unblocks
+    fn blocking_epoll_callback(
```
oli-obk, line 490: Please create a commit where you add this function and use it and then add the blocking logic in a follow-up commit. If you add it right below `epoll_wait`, the diff will be very simple

src/shims/unix/linux/epoll.rs
```diff
@@ -378,30 +379,25 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             this.write_scalar(Scalar::from_i32(-1), place)?;
             return Ok(());
         }
-        // FIXME: Implement blocking support
-        if timeout != 0 {
-            throw_unsup_format!("epoll_wait: timeout value can only be 0");
-        }
-
-        let Some(epfd) = this.machine.fds.get_ref(epfd) else {
+        let Some(epfd) = this.machine.fds.get_ref(epfd_value) else {
```
oli-obk, line 400: use `dup` here and move the `epfd` (after making it a weak) into the callback. This way you get the check for whether it was closed before waking up for free

src/shims/unix/linux/epoll.rs
oli-obk: you can turn the last commit back into a staging situation with `git reset --soft HEAD~1`

src/shims/unix/linux/epoll.rs
tiif: Oh I didn't know that rustfmt can't format macros, nice to know that. I will pay attention to this.

tiif:
> Just using an owned MPlaceTy should fix that.

This works, thanks!

src/shims/unix/linux/epoll.rs
```diff
@@ -378,29 +379,87 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         events_op: &OpTy<'tcx>,
         maxevents: &OpTy<'tcx>,
         timeout: &OpTy<'tcx>,
-    ) -> InterpResult<'tcx, Scalar> {
+        des: MPlaceTy<'tcx>,
```
oli-obk, line 382:
```suggestion
        dest: MPlaceTy<'tcx>,
```

src/shims/unix/linux/epoll.rs
```diff
@@ -378,29 +379,87 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let ready_list = binding.borrow_mut();
+
+        if ready_list.is_empty() {
+            if timeout == 0 {
+                // Non-blocking with no notification returned.
+                this.write_scalar(Scalar::from_i32(0), &des)?;
+                return Ok(());
+            } else {
+                // Blocking
+                let duration = Duration::from_secs(timeout.into());
```
oli-obk, line 420: Isn't the timeout in milliseconds?

tiif: Good catch! I totally forgot that.

src/shims/unix/fd.rs
```diff
@@ -296,6 +296,12 @@ impl WeakFileDescriptionRef {
     }
 }
 
+impl VisitProvenance for WeakFileDescriptionRef {
+    fn visit_provenance(&self, _visit: &mut VisitWith<'_>) {
+        // All our WeakFileDescriptionRef do not have any tags.
+    }
+}
+
```
tiif, line 304: I am not sure what is the purpose of ``VisitProvenance``. The compiler requires me to implement this in order to pass ``WeakFileDescriptionRef`` to ``callback!``.

src/shims/unix/linux/epoll.rs
```diff
@@ -378,39 +379,105 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let this = self.eval_context_mut();
+
+        let Some(epfd) = weak_epfd.upgrade() else {
+            throw_unsup_format!("epoll FD {epfd_value} is closed while blocking.")
+        };
+
+        //TODO: What if, there is another thread that make changes to the global fd table here?
+        // For example close the current epfd, then insert a new one inside?
+
+        let binding = epfd.borrow_mut().downcast_mut::<Epoll>().unwrap().get_ready_list();
```
tiif, line 477: I am currently trying to resolve borrow_mut error that happens to this line. I tried dropping ``epfd``or ``binding`` before calling ``blocking_epoll_callback`` but it couldn't be done because they are all borrowed. So I proceed to try dropping ``ready_list``, the ``borrow_mut`` error still persists.

src/shims/unix/fd.rs
RalfJung: The purpose is to call the `visit` callbacks on all pointers and places stored anywhere in the machine state.

However, a *weak* reference can never be the only reference to some pointer or place, so it is okay to make this a NOP for weak references. On the full `FileDescriptionRef` type, this would have to visit the file description itself, which would require making `VisitProvenance` a supertrait of `FileDescription`... lucky enough it seems like we do not have to do that.

Please fix the comment according to what I wrote here.

src/shims/unix/linux/epoll.rs
```diff
@@ -413,34 +417,40 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                     ),
                 );
             }
         } else {
-            // Non-blocking with notification returned.
-            this.blocking_epoll_callback(epfd_value, &des, &event)?;
+            // Immediately return epoll notification without blocking.
+
+            // Attempt to prevent epfd.borrow_mut in blocking_epoll_callback causing ICE.
+            drop(ready_list);
```
oli-obk, line 452: You need to drop `epfd`, which may be hard because of other things that need to be dropped first which are shadowed (e.g. `binding`). Maybe just move everything from the first `borrow_mut` into a nested block and only declare `weak_epfd` before the block

RalfJung:
I wonder if it would make sense to implement https://github.com/rust-lang/miri/issues/3665 first, as that's a much easier case of blocking? That might be a good way to learn how Miri deals with blocking.

oli-obk:
That's what I said, too, in our last meeting, and we realized it is not easier, as you need to refactor all read/write methods instead of just focussing on this one

oli-obk:
(If those are done first, I'd like to have a PR that passes `dest` to r/w and writes directly from there instead of returning the io result.)

RalfJung:
> I'd like to have a PR that passes dest to r/w and writes directly from there instead of returning the io result

Yeah that will anyway be needed for blocking.

That could also be done in a separate PR, to land the more conflict-prone change first.

src/shims/unix/linux/epoll.rs
tiif: Good idea, it works!

src/shims/unix/fd.rs
```diff
@@ -278,6 +278,12 @@ impl WeakFileDescriptionRef {
     }
 }
 
+impl VisitProvenance for WeakFileDescriptionRef {
+    fn visit_provenance(&self, _visit: &mut VisitWith<'_>) {
+        // All our WeakFileDescriptionRef do not have any tags.
```
oli-obk, line 283: Weak file descriptors are either tracked by a strong ref somewhere, or already dangling and thus not containing anything. So there's no need to visit them here

src/shims/unix/linux/epoll.rs
```diff
@@ -415,43 +416,113 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        dest: MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx> {
         let this = self.eval_context_mut();
 
-        let epfd = this.read_scalar(epfd)?.to_i32()?;
+        let epfd_value = this.read_scalar(epfd)?.to_i32()?;
         let events = this.read_immediate(events_op)?;
         let maxevents = this.read_scalar(maxevents)?.to_i32()?;
-        let timeout = this.read_scalar(timeout)?.to_i32()?;
+        let timeout = this.read_scalar(timeout)?.to_u32()?;
```
oli-obk, line 426: According to the docs the timeout is signed. What happens on a negative timeout?

src/shims/unix/linux/epoll.rs
```diff
@@ -415,43 +416,113 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        event: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx> {
+        let this = self.eval_context_mut();
+
+        let Some(epfd) = weak_epfd.upgrade() else {
+            throw_unsup_format!("epoll FD {epfd_value} is closed while blocking.")
         };
+
+        //TODO: What if, there is another thread that make changes to the global fd table here?
+        // For example close the current epfd, then insert a new one inside?
```
oli-obk, line 516: Since we're using a file description ref, that should not cause any issues now, right?

oli-obk (changes requested):
Please bless and fix the tokio sleep test

src/shims/unix/linux/epoll.rs
tiif: Yes, this shouldn't be a problem. It will only be a problem if we use the file descriptor value to retrieve the file description ref again.

src/shims/unix/linux/epoll.rs
tiif: Oh right, I haven't handled this. The documentation said if the timeout is -1, it will block indefinitely until it received a notification. I haven't found documentation on behaviour of other negative numbers, but it still seems to block indefinitely for every negative values on my host system (``x86_64-unknown-linux-gnu``).

src/shims/unix/linux/epoll.rs
```diff
@@ -500,9 +578,21 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                        // Collect all thread id of blocked threads that received notification.
+                        let thread_id = epoll_event_interest.thread_id;
+                        if this.has_blocked_on_epoll(thread_id) {
+                            waiters.push(thread_id);
+                        }
                     }
                 }
             }
+            // Dedup the thread id so we won't wake up a thread twice,
+            // TODO: is it necessary?
```
oli-obk, line 590: yes, miri will panic if you wake a thread twice

src/shims/unix/linux/epoll.rs
```diff
@@ -500,9 +578,21 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                         let event_instance =
                             EpollEventInstance::new(flags, epoll_event_interest.data);
                         ready_list.insert(epoll_key, event_instance);
+                        // Collect all thread id of blocked threads that received notification.
+                        let thread_id = epoll_event_interest.thread_id;
+                        if this.has_blocked_on_epoll(thread_id) {
+                            waiters.push(thread_id);
```
oli-obk, line 584: could also just wake here and record in a hashset whether the thread id was already woken

src/shims/unix/linux/epoll.rs
```diff
@@ -415,43 +420,110 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        // We just need to know if the ready list is empty. The whole logic is wrapped inside a block
+        // so we don't need to manually drop epfd later.
+        let ready_list_empty;
+        {
+            let epoll_file_description = epfd
+                .downcast::<Epoll>()
+                .ok_or_else(|| err_unsup_format!("non-epoll FD passed to `epoll_wait`"))?;
+            let binding = epoll_file_description.get_ready_list();
+            ready_list_empty = binding.borrow_mut().is_empty();
+        }
```
oli-obk, line 464:
```suggestion
        let ready_list_empty = epfd
                .downcast::<Epoll>()
                .ok_or_else(|| err_unsup_format!("non-epoll FD passed to `epoll_wait`"))?
                .get_ready_list()
                .borrow_mut()
                .is_empty();
```

src/shims/unix/linux/epoll.rs
```diff
@@ -415,43 +420,110 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                            this.write_scalar(Scalar::from_i32(0), &dest)?;
+                            Ok(())
+                        }
+                    ),
+                );
+            }
+        } else {
+            // Immediately return epoll notification without blocking.
+            this.blocking_epoll_callback(epfd_value, weak_epfd, &dest, &event)?;
+            return Ok(());
```
oli-obk, line 499:
```suggestion
```

src/shims/unix/linux/epoll.rs
```diff
@@ -415,43 +420,110 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                .ok_or_else(|| err_unsup_format!("non-epoll FD passed to `epoll_wait`"))?;
+            let binding = epoll_file_description.get_ready_list();
+            ready_list_empty = binding.borrow_mut().is_empty();
+        }
+
+        if ready_list_empty {
+            if timeout == 0 {
+                // Non-blocking with no notification returned.
+                this.write_scalar(Scalar::from_i32(0), &dest)?;
+                return Ok(());
```
oli-obk, line 470:
```suggestion
```

src/shims/unix/linux/epoll.rs
tiif: For the new change in https://github.com/rust-lang/miri/pull/3804/commits/9cb087dc8b529266bebcc9637d9ff8993c88c811, I was trying to avoid using ``this.unblock_thread`` inside 
```rust
if let Some(epoll_interests) = this.machine.epoll_interests.get_epoll_interest(id) { }
```
because it will trigger this error: 
```bash
error[E0502]: cannot borrow `*this` as mutable because it is also borrowed as immutable
   --> src/shims/unix/linux/epoll.rs:594:29
    |
568 |         if let Some(epoll_interests) = this.machine.epoll_interests.get_epoll_interest(id) {
    |                                        ---------------------------- immutable borrow occurs here
...
573 |             for weak_epoll_interest in epoll_interests {
    |                                        --------------- immutable borrow later used here
...
594 |                             this.unblock_thread(thread_id, BlockReason::Epoll)?;
    |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ mutable borrow occurs here
```
So instead I just store the ThreadIds that I want to unblock into a Vec then unblock them altogether outside of the block.

src/shims/unix/linux/epoll.rs
```diff
@@ -500,10 +583,23 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                        if let Some(thread_id) =
+                            epfd.downcast::<Epoll>().unwrap().thread_id.borrow_mut().pop()
+                        {
+                            waiter.push(thread_id);
+                        };
                     }
                 }
             }
         }
+        waiter.dedup();
```
oli-obk, line 599: You need to sort first. Dedup only dedups consecutive occurrences of the same value

oli-obk (changes requested):

tiif:
I will slightly explain how it currently works now: 
Edge-triggered notification, which is what we currently supporting, only unblock one thread when a notification is received, even when multiple thread is blocked on the same epoll instance. So how it currently works is before we call ``this.block_thread``, we record the ``ThreadId`` in ``Epoll`` file description using the ``thread_id`` field:
```rust
struct Epoll {
    /// A map of EpollEventInterests registered under this epoll instance.
    /// Each entry is differentiated using FdId and file descriptor value.
    interest_list: RefCell<BTreeMap<(FdId, i32), Rc<RefCell<EpollEventInterest>>>>,
    /// A map of EpollEventInstance that will be returned when `epoll_wait` is called.
    /// Similar to interest_list, the entry is also differentiated using FdId
    /// and file descriptor value.
    // This is an Rc because EpollInterest need to hold a reference to update
    // it.
    ready_list: Rc<RefCell<BTreeMap<(FdId, i32), EpollEventInstance>>>,
    /// A list of thread ids blocked on this epoll instance.
    thread_id: RefCell<Vec<ThreadId>>,
}
```
Later in ``check_and_update_readiness``, when we return a notification, we retrieve the list of blocked thread in that epoll instance, then unblock one of them.

src/shims/unix/linux/epoll.rs
```diff
@@ -415,43 +423,112 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // If the ready list is not empty, or the timeout is 0, we can return immediately.
+            this.blocking_epoll_callback(epfd_value, weak_epfd, &dest, &event)?;
+        } else {
+            // Blocking
+            let timeout = match timeout {
+                0.. => {
+                    let duration = Duration::from_millis(timeout.try_into().unwrap());
+                    Some((TimeoutClock::Monotonic, TimeoutAnchor::Relative, duration))
+                }
+                // If timeout value less than 0, block indefinitely until a notification
```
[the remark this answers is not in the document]
tiif: That's what I assumed at first, but I saw [this comment](https://github.com/torvalds/linux/blob/3d5f968a177d468cd13568ef901c5be84d83d32b/fs/eventpoll.c#L1957-L1961). It seems like in Linux it should block indefinitely for other negative value?

src/shims/unix/linux/epoll.rs
tiif: Ah I see. I misread the documentation.

src/shims/unix/linux/epoll.rs
tiif: This might not be applicable anymore because ``thread_ids = epoll_file_description.thread_id.borrow_mut();`` is added.

src/shims/unix/linux/epoll.rs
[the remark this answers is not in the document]
oli-obk: That seems to be for the timespec APIs. https://www.man7.org/linux/man-pages/man2/epoll_wait.2.html just talks about `-1`. Let's unsup all other negative values until we have a use case

src/shims/unix/linux/epoll.rs
```diff
@@ -500,10 +584,24 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                         let event_instance =
                             EpollEventInstance::new(flags, epoll_event_interest.data);
                         ready_list.insert(epoll_key, event_instance);
+                        // Edge-triggered notification only notify one thread even if there are
+                        // multiple threads block on the same epfd.
+                        let epfd = this.machine.fds.get(epoll_event_interest.epfd).unwrap();
+                        // FIXME: We can randomly pick a thread to unblock.
+                        if let Some(thread_id) =
+                            epfd.downcast::<Epoll>().unwrap().thread_id.borrow_mut().pop()
```
oli-obk, line 592: This unwrap can never fail or pick the wrong epfd, because if the epoll had been closed and the FD reused, the upgrade above would have failed

oli-obk: Add sth like this as a comment

oli-obk (approved):

src/shims/unix/linux/epoll.rs
```diff
@@ -422,43 +431,115 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // Blocking
+            let timeout = match timeout {
+                0.. => {
+                    let duration = Duration::from_millis(timeout.try_into().unwrap());
+                    Some((TimeoutClock::Monotonic, TimeoutAnchor::Relative, duration))
+                }
+                -1 => None,
+                ..-1 => {
+                    throw_unsup_format!(
+                        "epoll_wait: Only timeout values greater than -1 are supported."
```
RalfJung, line 491: equal to -1 also seems supported?

src/shims/unix/linux/epoll.rs
```diff
@@ -422,43 +431,115 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         events_op: &OpTy<'tcx>,
         maxevents: &OpTy<'tcx>,
         timeout: &OpTy<'tcx>,
-    ) -> InterpResult<'tcx, Scalar> {
+        dest: MPlaceTy<'tcx>,
```
RalfJung, line 434: Might be a bit nicer in terms of the interface to pass a reference here, and only clone inside this function when needed

src/shims/unix/linux/epoll.rs
```diff
@@ -422,43 +431,115 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        &mut self,
+        epfd_value: i32,
+        weak_epfd: WeakFileDescriptionRef,
+        dest: &MPlaceTy<'tcx>,
+        event: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx> {
+        let this = self.eval_context_mut();
+
+        let Some(epfd) = weak_epfd.upgrade() else {
+            throw_unsup_format!("epoll FD {epfd_value} is closed while blocking.")
```
RalfJung, line 532:
```suggestion
            throw_unsup_format!("epoll FD {epfd_value} got closed while blocking.")
```

src/shims/unix/linux/epoll.rs
```diff
@@ -422,43 +431,115 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
 
-        let Some(epfd) = this.machine.fds.get(epfd) else {
-            return Ok(Scalar::from_i32(this.fd_not_found()?));
+    /// Callback function after epoll_wait unblocks
+    fn blocking_epoll_callback(
+        &mut self,
+        epfd_value: i32,
+        weak_epfd: WeakFileDescriptionRef,
+        dest: &MPlaceTy<'tcx>,
+        event: &MPlaceTy<'tcx>,
```
RalfJung, line 527: This is an array of events, not just a single event, right?

tests/pass-dep/tokio/sleep.rs
RalfJung, line 1: Awesome. :-)

It is probably best to increase the allowed range for the "time elapsed" below: some CI runners sometimes sleep a lot longer than they should. Also, sleeping a full second makes the test quite slow.

I'd suggest to sleep 100ms, and then check that the sleep time is between 100ms and 1000ms.

src/shims/unix/linux/epoll.rs
tiif: >  Specifying a timeout of -1 causes epoll_wait() to block indefinitely, while specifying a timeout equal to zero causes epoll_wait() to return immediately, even if no events are available.

So equal to -1 should block indefinitely

src/shims/unix/linux/epoll.rs
tiif: Ah I think at first I tried passing the dest place by deference, but I ran into problem of needing to propagate the ``tcx`` runtime. Unfortunately the commit is squashed, but the initial problem description is still in the history of PR description. 

----------------------------------------------------------------------------------------------------------------------------------------------------------
This is the summary, I ommited the compiler error: 

This is a work in progress PR opened for discussion purpose.

In https://github.com/rust-lang/miri/commit/f139c7225ae0ce1adf3c42aab905092942203ad8, the reason that 'tcx is annotated in so many place is because I have used this:
```rust
                this.block_thread(
                    BlockReason::Epoll,
                    Some((TimeoutClock::Monotonic, TimeoutAnchor::Relative, duration)),
                    callback!(
                        @capture<'tcx> {
                        epfd_value: i32,
                        place: &'tcx MPlaceTy<'tcx>, // the culprit is here!!
                        event: MPlaceTy<'tcx>,
                        }
                        @unblock = |this| {
                                    this.blocking_epoll_callback(epfd_value, place, &event);
                                    Ok(())
                                }
                        @timeout = |this| {
                        this.write_scalar(Scalar::from_i32(0), place)?;
                                    Ok(())
                                }
                    ),
                );
```

place: &'tcx MPlaceTy<'tcx> is required to be annotated with 'tcx. Since it is passed from shims/foreign_items.rs, every of them need to be annotated with 'tcx too. I managed to follow the compiler's hint and add them one by one until I hit this error:

It seems like if I want to change the parameters in call_extra_fn, I need to go all the way to rustc miri driver to change [this function](https://github.com/rust-lang/rust/blob/9859bf27fd9892f48725c59b56aeee2be1d2fbad/compiler/rustc_const_eval/src/interpret/machine.rs#L217). Is there any way to get around this?

src/shims/unix/linux/epoll.rs
RalfJung: Yes, but the text is wrong. It says "only greater than -1 is supported". But equal to -1 is also supported, and -1 is not greater than -1.

It should say "greater than or equal to".

src/shims/unix/linux/epoll.rs
```diff
@@ -422,43 +431,115 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                }
+                -1 => None,
+                ..-1 => {
+                    throw_unsup_format!(
+                        "epoll_wait: Only timeout values greater than -1 are supported."
+                    );
+                }
+            };
+            thread_ids.push(this.active_thread());
+            this.block_thread(
```
RalfJung, line 496: You simply need `let dets = dest.clone();` before `block_thread` here.

src/shims/unix/linux/epoll.rs
tiif: Alright, I deleted a reply because I misunderstood what you meant just now.

src/shims/unix/linux/epoll.rs
tiif: Yup, I renamed it to ``events`` now so it'd be clearer.

src/shims/unix/linux/epoll.rs
tiif: Ahh thanks, I completely missed that.

tiif:
Weird, why the new sleep.rs test ICE while the previous one passes

full ICE error:
```bash
thread 'rustc' panicked at src/concurrency/thread.rs:1069:18:
unblock_thread: thread was not blocked
stack backtrace:
   0:     0x7ad94a5dd1ed - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h2378b8792e433be4
   1:     0x7ad94ae05297 - core::fmt::write::h6defb7cb622411a0
   2:     0x7ad94bdb3951 - std::io::Write::write_fmt::h4fdbd5ff16cd4c77
   3:     0x7ad94a5df8cb - std::panicking::default_hook::{{closure}}::h877422fca1bd59b0
   4:     0x7ad94a5df53e - std::panicking::default_hook::h8addcd7117467f88
   5:     0x7ad94975b4b9 - std[7fd3e6b47a20060b]::panicking::update_hook::<alloc[30656b0279f8cbd9]::boxed::Box<rustc_driver_impl[541434d8921df733]::install_ice_hook::{closure#0}>>::{closure#0}
   6:     0x7ad94a5e01e7 - std::panicking::rust_panic_with_hook::h42eb79b8d60a1a34
   7:     0x7ad94a5dfe73 - std::panicking::begin_panic_handler::{{closure}}::h40846de059469689
   8:     0x7ad94a5dd6a9 - std::sys::backtrace::__rust_end_short_backtrace::h1a4e038443a2e97c
   9:     0x7ad94a5dfb74 - rust_begin_unwind
  10:     0x7ad9474b9e93 - core::panicking::panic_fmt::h4189083a9204f8ed
  11:     0x63410a32ce70 - miri::concurrency::thread::EvalContextExt::unblock_thread::hbf8b0c993420aca5
                               at /home/byt/Documents/miri/src/concurrency/thread.rs:1069:18
  12:     0x63410a38cccd - miri::shims::unix::linux::epoll::EvalContextExt::check_and_update_readiness::heeed6bf043256241
                               at /home/byt/Documents/miri/src/shims/unix/linux/epoll.rs:618:13
  13:     0x63410a46dc6b - <miri::shims::unix::linux::eventfd::Event as miri::shims::unix::fd::FileDescription>::write::h194bb96fbea414cf
                               at /home/byt/Documents/miri/src/shims/unix/linux/eventfd.rs:151:9
  14:     0x63410a376f27 - miri::shims::unix::fd::EvalContextExt::write::hc62fb912d0b7f058
                               at /home/byt/Documents/miri/src/shims/unix/fd.rs:626:21
  15:     0x63410a36dd74 - miri::shims::unix::foreign_items::EvalContextExt::emulate_foreign_item_inner::hbbe3a3e25f95fb93
                               at /home/byt/Documents/miri/src/shims/unix/foreign_items.rs:104:30
  16:     0x63410a3cb210 - miri::shims::foreign_items::EvalContextExtPriv::emulate_foreign_item_inner::h214a6110838250a3
                               at /home/byt/Documents/miri/src/shims/foreign_items.rs:967:25
  17:     0x63410a3c792a - miri::shims::foreign_items::EvalContextExt::emulate_foreign_item::h045bc32060e067bb
                               at /home/byt/Documents/miri/src/shims/foreign_items.rs:70:15
  18:     0x63410a2eeb36 - <miri::machine::MiriMachine as rustc_const_eval::interpret::machine::Machine>::find_mir_or_eval_fn::hf172726fe48df113
                               at /home/byt/Documents/miri/src/machine.rs:979:20
  19:     0x63410a2eeb36 - rustc_const_eval::interpret::call::<impl rustc_const_eval::interpret::eval_context::InterpCx<M>>::init_fn_call::h9b948389f5d7c634
                               at /rustc/fdf61d499c8a8421ecf98e7924bb87caf43a9938/compiler/rustc_const_eval/src/interpret/call.rs:588:46
  20:     0x63410a32de75 - rustc_const_eval::interpret::step::<impl rustc_const_eval::interpret::eval_context::InterpCx<M>>::eval_terminator::ha7d1d92c3c4a1c9e
                               at /rustc/fdf61d499c8a8421ecf98e7924bb87caf43a9938/compiler/rustc_const_eval/src/interpret/step.rs:499:17
  21:     0x63410a32de75 - rustc_const_eval::interpret::step::<impl rustc_const_eval::interpret::eval_context::InterpCx<M>>::step::hc1688bd5177f2073
                               at /rustc/fdf61d499c8a8421ecf98e7924bb87caf43a9938/compiler/rustc_const_eval/src/interpret/step.rs:64:9
  22:     0x63410a32de75 - miri::concurrency::thread::EvalContextExt::run_threads::he2750f0bbd8226dc
                               at /home/byt/Documents/miri/src/concurrency/thread.rs:1195:25
  23:     0x63410a4687de - miri::eval::eval_entry::{{closure}}::hc648965f40e2cdc9
                               at /home/byt/Documents/miri/src/eval.rs:455:49
  24:     0x63410a4687de - core::ops::function::FnOnce::call_once::h601fcf33322d45d6
                               at /rustc/fdf61d499c8a8421ecf98e7924bb87caf43a9938/library/core/src/ops/function.rs:250:5
  25:     0x63410a4687de - <core::panic::unwind_safe::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once::hd0650cdc6c3ba54a
                               at /rustc/fdf61d499c8a8421ecf98e7924bb87caf43a9938/library/core/src/panic/unwind_safe.rs:272:9
  26:     0x63410a4687de - std::panicking::try::do_call::h48bd7d5fd29ae83e
                               at /rustc/fdf61d499c8a8421ecf98e7924bb87caf43a9938/library/std/src/panicking.rs:554:40
  27:     0x63410a4687de - std::panicking::try::hf31f7b75e83b93ba
                               at /rustc/fdf61d499c8a8421ecf98e7924bb87caf43a9938/library/std/src/panicking.rs:518:19
  28:     0x63410a4687de - std::panic::catch_unwind::h3d0e6df908ad608f
                               at /rustc/fdf61d499c8a8421ecf98e7924bb87caf43a9938/library/std/src/panic.rs:345:14
  29:     0x63410a4687de - miri::eval::eval_entry::hb393bc939b2239c3
                               at /home/byt/Documents/miri/src/eval.rs:455:9
  30:     0x63410a1b230b - <miri::MiriCompilerCalls as rustc_driver_impl::Callbacks>::after_analysis::{{closure}}::hce6aa30eb215eea3
                               at /home/byt/Documents/miri/src/bin/miri.rs:118:40
  31:     0x63410a1b230b - rustc_middle::ty::context::GlobalCtxt::enter::{{closure}}::h744506cae6e1e4b3
                               at /rustc/fdf61d499c8a8421ecf98e7924bb87caf43a9938/compiler/rustc_middle/src/ty/context.rs:1322:37
  32:     0x63410a1b230b - rustc_middle::ty::context::tls::enter_context::{{closure}}::h787c39c426869962
                               at /rustc/fdf61d499c8a8421ecf98e7924bb87caf43a9938/compiler/rustc_middle/src/ty/context/tls.rs:82:9
  33:     0x63410a1b230b - std::thread::local::LocalKey<T>::try_with::h9b34ac5dbe714836
                               at /rustc/fdf61d499c8a8421ecf98e7924bb87caf43a9938/library/std/src/thread/local.rs:283:12
  34:     0x63410a1b230b - std::thread::local::LocalKey<T>::with::h87102dc8a4017475
                               at /rustc/fdf61d499c8a8421ecf98e7924bb87caf43a9938/library/std/src/thread/local.rs:260:9
  35:     0x63410a1b230b - rustc_middle::ty::context::tls::enter_context::hcfae43a1a616f7d0
                               at /rustc/fdf61d499c8a8421ecf98e7924bb87caf43a9938/compiler/rustc_middle/src/ty/context/tls.rs:79:9
  36:     0x63410a1b230b - rustc_middle::ty::context::GlobalCtxt::enter::he124cf1b4f82396e
                               at /rustc/fdf61d499c8a8421ecf98e7924bb87caf43a9938/compiler/rustc_middle/src/ty/context.rs:1322:9
  37:     0x63410a1b8329 - rustc_interface::queries::QueryResult<&rustc_middle::ty::context::GlobalCtxt>::enter::hedc0213c26249cff
                               at /rustc/fdf61d499c8a8421ecf98e7924bb87caf43a9938/compiler/rustc_interface/src/queries.rs:65:9
  38:     0x63410a1b8329 - <miri::MiriCompilerCalls as rustc_driver_impl::Callbacks>::after_analysis::h4a21ad44755ab7b4
                               at /home/byt/Documents/miri/src/bin/miri.rs:82:9
  39:     0x7ad94bbffceb - rustc_interface[11143f1d128096df]::interface::run_compiler::<core[f39a9c8bf043078e]::result::Result<(), rustc_span[373c84a13cb08004]::ErrorGuaranteed>, rustc_driver_impl[541434d8921df733]::run_compiler::{closure#0}>::{closure#1}
  40:     0x7ad94bbcd850 - std[7fd3e6b47a20060b]::sys::backtrace::__rust_begin_short_backtrace::<rustc_interface[11143f1d128096df]::util::run_in_thread_with_globals<rustc_interface[11143f1d128096df]::util::run_in_thread_pool_with_globals<rustc_interface[11143f1d128096df]::interface::run_compiler<core[f39a9c8bf043078e]::result::Result<(), rustc_span[373c84a13cb08004]::ErrorGuaranteed>, rustc_driver_impl[541434d8921df733]::run_compiler::{closure#0}>::{closure#1}, core[f39a9c8bf043078e]::result::Result<(), rustc_span[373c84a13cb08004]::ErrorGuaranteed>>::{closure#0}, core[f39a9c8bf043078e]::result::Result<(), rustc_span[373c84a13cb08004]::ErrorGuaranteed>>::{closure#0}::{closure#0}, core[f39a9c8bf043078e]::result::Result<(), rustc_span[373c84a13cb08004]::ErrorGuaranteed>>
  41:     0x7ad94bbcdeba - <<std[7fd3e6b47a20060b]::thread::Builder>::spawn_unchecked_<rustc_interface[11143f1d128096df]::util::run_in_thread_with_globals<rustc_interface[11143f1d128096df]::util::run_in_thread_pool_with_globals<rustc_interface[11143f1d128096df]::interface::run_compiler<core[f39a9c8bf043078e]::result::Result<(), rustc_span[373c84a13cb08004]::ErrorGuaranteed>, rustc_driver_impl[541434d8921df733]::run_compiler::{closure#0}>::{closure#1}, core[f39a9c8bf043078e]::result::Result<(), rustc_span[373c84a13cb08004]::ErrorGuaranteed>>::{closure#0}, core[f39a9c8bf043078e]::result::Result<(), rustc_span[373c84a13cb08004]::ErrorGuaranteed>>::{closure#0}::{closure#0}, core[f39a9c8bf043078e]::result::Result<(), rustc_span[373c84a13cb08004]::ErrorGuaranteed>>::{closure#1} as core[f39a9c8bf043078e]::ops::function::FnOnce<()>>::call_once::{shim:vtable#0}
  42:     0x7ad94bbce22b - std::sys::pal::unix::thread::Thread::new::thread_start::h4d6c0379e10f6e16
  43:     0x7ad945e94ac3 - start_thread
                               at ./nptl/pthread_create.c:442:8
  44:     0x7ad945f26850 - __GI___clone3
                               at ./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
  45:                0x0 - <unknown>

error: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/miri/issues/new

note: please make sure that you have updated to the latest nightly

note: please attach the file at `/home/byt/Documents/miri/rustc-ice-2024-08-24T15_47_02-721586.txt` to your bug report

query stack during panic:
end of query stack

Miri caused an ICE during evaluation. Here's the interpreter backtrace at the time of the panic:
note: the place in the program where the ICE was triggered
   --> /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/sys/pal/unix/fd.rs:291:13
    |
291 | /             libc::write(
292 | |                 self.as_raw_fd(),
293 | |                 buf.as_ptr() as *const libc::c_void,
294 | |                 cmp::min(buf.len(), READ_LIMIT),
295 | |             )
    | |_____________^
    |
    = note: BACKTRACE:
    = note: inside `std::sys::pal::unix::fd::FileDesc::write` at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/sys/pal/unix/fd.rs:291:13: 295:14
    = note: inside `std::sys::pal::unix::fs::File::write` at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/sys/pal/unix/fs.rs:1265:9: 1265:26
    = note: inside `<&std::fs::File as std::io::Write>::write` at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/fs.rs:850:9: 850:30
    = note: inside `mio::sys::unix::waker::Waker::wake` at /home/byt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/mio-1.0.1/src/sys/unix/waker/eventfd.rs:46:15: 46:37
    = note: inside `mio::waker::Waker::wake` at /home/byt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/mio-1.0.1/src/waker.rs:94:9: 94:26
    = note: inside `tokio::runtime::io::driver::Handle::unpark` at /home/byt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.39.2/src/runtime/io/driver.rs:209:9: 209:26
    = note: inside `tokio::runtime::driver::IoHandle::unpark` at /home/byt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.39.2/src/runtime/driver.rs:198:46: 198:61
    = note: inside `tokio::runtime::driver::Handle::unpark` at /home/byt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.39.2/src/runtime/driver.rs:90:9: 90:25
    = note: inside `tokio::runtime::scheduler::multi_thread::park::Inner::unpark` at /home/byt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.39.2/src/runtime/scheduler/multi_thread/park.rs:213:30: 213:45
    = note: inside `tokio::runtime::scheduler::multi_thread::park::Unparker::unpark` at /home/byt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.39.2/src/runtime/scheduler/multi_thread/park.rs:109:9: 109:34
    = note: inside `tokio::runtime::scheduler::multi_thread::worker::<impl tokio::runtime::scheduler::multi_thread::handle::Handle>::notify_all` at /home/byt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.39.2/src/runtime/scheduler/multi_thread/worker.rs:1152:13: 1152:47
    = note: inside `tokio::runtime::scheduler::multi_thread::worker::<impl tokio::runtime::scheduler::multi_thread::handle::Handle>::close` at /home/byt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.39.2/src/runtime/scheduler/multi_thread/worker.rs:1131:13: 1131:30
    = note: inside `tokio::runtime::scheduler::multi_thread::handle::Handle::shutdown` at /home/byt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.39.2/src/runtime/scheduler/multi_thread/handle.rs:44:9: 44:21
    = note: inside `tokio::runtime::scheduler::multi_thread::MultiThread::shutdown` at /home/byt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.39.2/src/runtime/scheduler/multi_thread/mod.rs:93:55: 93:72
    = note: inside `<tokio::runtime::Runtime as std::ops::Drop>::drop` at /home/byt/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.39.2/src/runtime/runtime.rs:494:17: 494:58
    = note: inside `std::ptr::drop_in_place::<tokio::runtime::Runtime> - shim(Some(tokio::runtime::Runtime))` at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:536:1: 536:56
note: inside `main`
   --> ./tests/pass-dep/tokio/sleep.rs:11:68
    |
11  |     assert!((100..1000).contains(time_elapsed), "{}", time_elapsed);
    |                                                                    ^
    = note: inside `<fn() as std::ops::FnOnce<()>>::call_once - shim(fn())` at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5: 250:71
    = note: inside `std::sys::backtrace::__rust_begin_short_backtrace::<fn(), ()>` at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/sys/backtrace.rs:154:18: 154:21
    = note: inside closure at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/rt.rs:164:18: 164:75
    = note: inside `std::ops::function::impls::<impl std::ops::FnOnce<()> for &dyn std::ops::Fn() -> i32 + std::marker::Sync + std::panic::RefUnwindSafe>::call_once` at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/core/src/ops/function.rs:284:13: 284:31
    = note: inside `std::panicking::r#try::do_call::<&dyn std::ops::Fn() -> i32 + std::marker::Sync + std::panic::RefUnwindSafe, i32>` at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/panicking.rs:554:40: 554:43
    = note: inside `std::panicking::r#try::<i32, &dyn std::ops::Fn() -> i32 + std::marker::Sync + std::panic::RefUnwindSafe>` at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/panicking.rs:518:19: 518:88
    = note: inside `std::panic::catch_unwind::<&dyn std::ops::Fn() -> i32 + std::marker::Sync + std::panic::RefUnwindSafe, i32>` at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/panic.rs:345:14: 345:33
    = note: inside closure at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/rt.rs:143:48: 143:73
    = note: inside `std::panicking::r#try::do_call::<{closure@std::rt::lang_start_internal::{closure#2}}, isize>` at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/panicking.rs:554:40: 554:43
    = note: inside `std::panicking::r#try::<isize, {closure@std::rt::lang_start_internal::{closure#2}}>` at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/panicking.rs:518:19: 518:88
    = note: inside `std::panic::catch_unwind::<{closure@std::rt::lang_start_internal::{closure#2}}, isize>` at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/panic.rs:345:14: 345:33
    = note: inside `std::rt::lang_start_internal` at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/rt.rs:143:20: 143:98
    = note: inside `std::rt::lang_start::<()>` at /home/byt/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/rt.rs:163:17: 168:6

Caused by:
  process didn't exit successfully: `/home/byt/Documents/miri/target/debug/deps/ui-660f8988aa44457c --miri-run-dep-mode --edition 2021 --sysroot /home/byt/.cache/miri ./tests/pass-dep/tokio/sleep.rs` (exit status: 1)
Error: command exited with non-zero code `cargo +miri test --manifest-path /home/byt/Documents/miri/./Cargo.toml --test ui --quiet -- --miri-run-dep-mode --edition 2021 --sysroot /home/byt/.cache/miri ./tests/pass-dep/tokio/sleep.rs`: 1
```

tiif:
For the current sleep.rs test, before I called ``this.block_thread``, the result of ``this.active_thread()`` is ``ThreadId(1)``, but it tried to unblock ``ThreadId(1)``, it panics? I logged ``this.unblock_thread`` and there is nothing that unblocks ``ThreadId(1)`` before it panics. 

@rustbot author

src/shims/unix/linux/epoll.rs
```diff
@@ -507,10 +594,31 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                     }
                 }
             }
         }
+        waiter.sort();
+        waiter.dedup();
+        for thread_id in waiter {
+            if this.has_blocked_on_epoll(thread_id) {
+                this.unblock_thread(thread_id, BlockReason::Epoll)?;
+            }
```
tiif, line 620: It turned out that it might be possible for a thread to wake up before we call unblock, so we'd need an additional check here.

RalfJung: That sounds like the blocked thread bookkeeping is going wrong somewhere. Who else is waking up this thread?

src/shims/unix/linux/epoll.rs
```diff
@@ -422,43 +431,116 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                ),
+            );
         }
+        Ok(())
+    }
 
-        let Some(epfd) = this.machine.fds.get(epfd) else {
-            return Ok(Scalar::from_i32(this.fd_not_found()?));
+    /// Callback function after epoll_wait unblocks
+    fn blocking_epoll_callback(
```
RalfJung, line 523: This is a private helper function, right?

Currently this PR makes it publicly callable. That doesn't seem good.

src/shims/unix/linux/epoll.rs
tiif: Ah right, there is a problem with the thread bookkeeping. The thread is woken up because of timeout, I should clean it up when timeout has reached.

oli-obk:
The PR's main message is a bit outdated

tests/pass-dep/libc/libc-epoll-blocking.rs
```diff
@@ -0,0 +1,98 @@
+//@only-target-linux
```
RalfJung, line 1: Maybe make this set the preemption rate to 0, to ensure the thread interleaving is deterministic?

tiif: Yup, already did, that's what I am thinking too :D

src/shims/unix/linux/epoll.rs
```diff
@@ -421,59 +430,100 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    @unblock = |this| {
+                        blocking_epoll_callback(epfd_value, weak_epfd, &dest, &event, this)?;
+                        Ok(())
+                    }
+                    @timeout = |this| {
+                        // No notification after blocking timeout.
+                        let Some(epfd) = weak_epfd.upgrade() else {
+                            throw_unsup_format!("epoll FD {epfd_value} got closed while blocking.")
+                        };
+                        // Remove the current active thread_id from the blocked thread_id list.
```
RalfJung, line 515: Ah yes indeed, it is important that *both* the `unblock` and `timeout` callbacks remove the thread from the wait list.

tiif: For ``unblock``, the thread id should be removed in ``check_and_update_readiness`` through 
```rust
if let Some(thread_id) = epfd.downcast::<Epoll>().unwrap().thread_id.borrow_mut().pop()
```

For now, I think thread that blocks on epoll can only be unblocked by ``this.unblock_thread`` in ``check_and_update_readiness`` or the timeout. Is there any more operation that can unblock the thread?

RalfJung: That sounds right. No other module should ever unblock epoll-blocked threads, *except* for timeouts.

tiif:
Do we want another test for triggering a notification after the ``epoll_wait`` times out or is the sleep.rs test good enough?

RalfJung:
There should be such a test in your libc-epoll-blocking test file.

tiif:
> But doesn't `test_epoll_block_without_notification` do that?

No, that test only epoll_wait, then timeout. A notification need to be sent after timeout to trigger the bug. Also there is an error on the comment for that test, sorry for the confusion.

RalfJung:
Okay, just make sure the test has a comment explaining the exact scenario. :)

Issue #3443, opened 2024-04-03
Meta: Adopt a "project" process


RalfJung:
Close to two years ago, we landed stubs for epoll and some related functions to get an empty tokio program to work. Those stubs were buggy and lacked basic sanity checks, and two years later they are still just stubs. IMO we can conclude that this approach of just landing something and then figuring things out as we go was a failure.

When embarking on larger projects, like epoll support, we need a plan -- t-lang has design meetings, t-compiler has MCPs, t-libs-api has ACPs. The pattern is always the same: we should have an idea and consensus on where we want to go and how we want to get there before we start even reviewing PRs that move us in that direction.

So I think for Miri we want a "project" process. It doesn't have to be heavyweight, but I am imagining something roughly like this:
1. A project proposal defines clearly what the **goal** is of this project. This defines the scope of the project, i.e. which part of which APIs should be supported. If this involves functions that expose a big API surface with lots of flags, the project may want to support only a tiny subset of flags; that should be documented. There must be at least one testcase given that Miri currently doesn't support and that should be supported when the project is done.  This testcase doesn't need to cover *everything* but it should call each relevant API at least once with a representative set of arguments and flags. IOW, the project is not defined by "we are done when the testcase passes", it just documents more concretely the scope of the project.
    The testcase should be expressed by directly calling the relevant C FFI functions, not a high-level Rust crate. It's okay to start with a high-level test, but it's hard to judge the scope of the project and to evaluate the plan in step 2 without knowing the actual API surface Miri will be responsible for. So in that case developing the low-level test case is a blocker for moving to the next step of the project.
2. Once the team accepts that this is a goal we want to pursue, we need to make a **plan**. What exactly the plan looks like will depend heavily on the concrete project. Any extension to the machine state (adding anything new to `MiriMachine` or any of its fields) definitely needs to be part of the plan, specifying the type of the new state component and where it goes.  Similar if a new file descriptor type is involved, the data stored in the FD table for this file descriptor type needs to be pat of the plan. There doesn't have to be any code, just a rough description of how the new operations Miri needs to implement for this project interact with the new state -- and, if applicable, how existing operations interact with the new state.
3. Once the team accepts the plan, **implementation** can begin. Now that we have a plan, implementation doesn't need to be one big PR, but we can still make sense of each PR as taking small steps towards the goal, following the plan.

[...]

PR #3807, opened 2024-08-15
add 'project' process guidlines for larger contributions

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/3443

I am honestly not entirely sure what the consensus from what issue was. I feel like the epoll PR worked reasonably well, and not having been closely involved I am not sure which process @oli-obk followed there. Compared to the first draft in #3443 I tried to make this less formal and framed more as guidelines than hard rules.


saethlin:
The only disagreement I had with you on the issue was whether this guidance would have prevented a past problem.

I'm quite happy with the wording here.

PR #3809, opened 2024-08-16
FD: remove big surrounding RefCell, simplify socketpair

RalfJung:
A while ago, I added the big implicit RefCell for all file descriptions since it avoided interior mutability in `eventfd`. However, this requires us to hold the RefCell "lock" around the entire invocation of the `read`/`write` methods on an FD, which is not great. For instance, if an FD wants to update epoll notifications from inside its `read`/`write`, it is very crucial that the notification check does not end up accessing the FD itself. Such cycles, however, occur naturally:
- eventfd wants to update notifications for itself
- socketfd wants to update notifications on its "peer", which will in turn check *its* peer to see whether that buffer is empty -- and my peer's peer is myself.

This then also lets us simplify socketpair, which currently holds a weak reference to its peer *and* a weak reference to the peer's buffer -- that was previously needed precisely to avoid this issue.

### DIFF
--- a/src/shims/unix/linux/epoll.rs
+++ b/src/shims/unix/linux/epoll.rs
@@ -35,6 +35,7 @@ impl EpollEventInstance {
         EpollEventInstance { events, data }
     }
 }
+
 /// EpollEventInterest registers the file description information to an epoll
 /// instance during a successful `epoll_ctl` call. It also stores additional
 /// information needed to check and update readiness state for `epoll_wait`.
@@ -434,7 +435,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
     /// For a specific file description, get its ready events and update
     /// the corresponding ready list. This function is called whenever a file description
-    /// is registered with epoll, or when its readiness *might* have changed.
+    /// is registered with epoll, or the buffer it reads from / writes to changed.
+    /// This *will* report an event if anyone is subscribed to it, without any further
+    /// filtering, so do not call this function when an FD didn't have anything happen to it!
     fn check_and_update_readiness(&self, fd_ref: &FileDescriptionRef) -> InterpResult<'tcx, ()> {
         let this = self.eval_context_ref();
         let id = fd_ref.get_id();
--- a/src/shims/unix/linux/epoll.rs
+++ b/src/shims/unix/linux/epoll.rs
@@ -433,11 +433,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         Ok(Scalar::from_i32(num_of_events))
     }
 
-    /// For a specific file description, get its ready events and update
-    /// the corresponding ready list. This function is called whenever a file description
-    /// is registered with epoll, or the buffer it reads from / writes to changed.
-    /// This *will* report an event if anyone is subscribed to it, without any further
-    /// filtering, so do not call this function when an FD didn't have anything happen to it!
+    /// For a specific file description, get its ready events and update the corresponding ready
+    /// list. This function should be called whenever an event causes more bytes or an EOF to become
+    /// newly readable from an FD, and whenever more bytes can be written to an FD or no more future
+    /// writes are possible.
+    ///
+    /// This *will* report an event if anyone is subscribed to it, without any further filtering, so
+    /// do not call this function when an FD didn't have anything happen to it!
     fn check_and_update_readiness(&self, fd_ref: &FileDescriptionRef) -> InterpResult<'tcx, ()> {
         let this = self.eval_context_ref();
         let id = fd_ref.get_id();
--- a/src/shims/unix/linux/epoll.rs
+++ b/src/shims/unix/linux/epoll.rs
@@ -12,7 +12,7 @@ use crate::*;
 struct Epoll {
     /// A map of EpollEventInterests registered under this epoll instance.
     /// Each entry is differentiated using FdId and file descriptor value.
-    interest_list: BTreeMap<(FdId, i32), Rc<RefCell<EpollEventInterest>>>,
+    interest_list: RefCell<BTreeMap<(FdId, i32), Rc<RefCell<EpollEventInterest>>>>,
     /// A map of EpollEventInstance that will be returned when `epoll_wait` is called.
     /// Similar to interest_list, the entry is also differentiated using FdId
     /// and file descriptor value.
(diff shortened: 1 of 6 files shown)


tests/pass-dep/libc/libc-epoll.rs
```diff
@@ -84,7 +85,7 @@ fn test_epoll_socketpair() {
     assert_eq!(res, 0);
 
     // Check result from epoll_wait.
-    let expected_event = u32::try_from(libc::EPOLLRDHUP | libc::EPOLLIN | libc::EPOLLOUT).unwrap();
+    let expected_event = u32::try_from(libc::EPOLLRDHUP | libc::EPOLLIN).unwrap();
```
RalfJung, line 88: For some reason I do not understand, this changes behavior in this test. @tiif any idea what is happening here? Sadly the test does not explain *why* EPOLLOUT should be triggered here. If I look at SocketPair::get_epoll_ready_events I am not sure why EPOLLOUT would be triggered -- no new space became available in the write buffer, after all.

src/shims/unix/socket.rs
```diff
@@ -49,25 +57,23 @@ impl FileDescription for SocketPair {
...
+            let writebuf = &peer_fd.downcast::<SocketPair>().unwrap().readbuf.borrow();
             let data_size = writebuf.buf.len();
             let available_space = MAX_SOCKETPAIR_BUFFER_CAPACITY.strict_sub(data_size);
             if available_space != 0 {
                 epoll_ready_events.epollout = true;
             }
-        }
-
-        // Check if the peer_fd closed
-        if self.peer_fd.upgrade().is_none() {
```
RalfJung, line 70: Okay looking at the old code here I think I have a theory for why behavior changed. But I can't imagine that the old behavior was intentional.

In think previously we had a situation where writebuf still existed but the peer was already destroyed. That's because a file descriptor reference gets invalidated *before* the call to `FileDescriptor::close`, but the `Rc` inside of it (in this case, for the buffer) survive until they get dropped, which happens at the *end of* `FileDescriptor::close`.

src/shims/unix/linux/epoll.rs
```diff
@@ -153,6 +153,60 @@ impl EpollInterestTable {
...
+impl<'tcx> EvalContextExtPriv<'tcx> for crate::MiriInterpCx<'tcx> {}
+trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    /// For a specific unique file descriptor id, get its ready events and update
+    /// the corresponding ready list. This function is called whenever a file description
+    /// is registered with epoll, or when read, write, or close operations are performed,
+    /// regardless of any changes in readiness.
+    ///
+    /// This is an internal helper function and is typically not meant to be used directly.
+    /// `FileDescriptionRef::check_and_update_readiness` should be preferred.
+    fn check_and_update_readiness(
```
oli-obk, line 165: Can't we inline this into FileDescriptionRef::check_and_update_readiness now?

RalfJung: Yeah we can... I guess there's no good reason to keep the previous code structure.

tests/pass-dep/libc/libc-epoll.rs
RalfJung: If I run the tests against real libc, I get yet another result here: `libc::EPOLLRDHUP | libc::EPOLLHUP | libc::EPOLLIN | libc::EPOLLOUT`. I also get other differences between the test and real libc, e.g. on line 172.

tiif: This is the best explanation I could find for this behaviour. This is generally true for all the test that I have written for libc epoll in real system. 

> To be perhaps clearer, epoll_wait() won't return an fd unless something changed on that socket, but if something did change, it returns all the flags representing the current state.

Source: https://stackoverflow.com/questions/59288735/epoll-wait-return-epollout-even-with-epollet-flag

So since this file description is available to be written at the point when ``close`` is triggered, epoll will also mark it with ``EPOLLOUT`` to represent the current state of the file description. 

This is also documented in https://hackmd.io/@tiif/SJatftrH0#Epoll-semantics. Most of the design in this docs is pretty outdated, but I can go through it again and update it if it helps to serve as an alternative documentation.

RalfJung: > So since this file description is available to be written at the point when close is triggered, epoll will also mark it with EPOLLOUT to represent the current state of the file description.

That doesn't answer the question though.
When the peer FD is closed, how can the FD be considered writeable?

There's already this very odd special case:
```
// This is an edge case. Whenever epollrdhup is triggered, epollin will be added
// even though there is no data in the buffer.
```
This sounds like we should also set `epollout = true`, tog et the old behavior back, where it worked just by accident.

But that doesn't explain why we should set it to true.

tiif: > When the peer FD is closed, how can the FD be considered writeable?

Now I think about it, I agree with this. it shouldn't be considered writable although this is what libc returns. This should be considered part of the edge case behaviour instead of "it is writable". 

This indeed works by accident, because we notify the peer of closed fd to ``check_and_update_readiness`` before the fd actually closed. 

The logic is in ``socketpair::close``, in particular this chunk of code: 
```rust
        // Notify peer fd that closed has happened.
        if let Some(peer_fd) = self.peer_fd.upgrade() {
            // When any of the event happened, we check and update the status of all supported events
            // types of peer fd.
            peer_fd.check_and_update_readiness(ecx)?;
        }
```

oli-obk: My assumption was for it to unblock all threads waiting on writing or reading to/from this, and letting them retry and subsequently fail because the FD was already closed.

RalfJung: So the idea is -- if your peer gets closed, then a read can now return EOF, and a write can return EPIPE, and that's the sense in which sense there was a change in read status and write status?

Yeah I can see that. I'll update the comments.

oli-obk:
bors r plus

RalfJung:
Sorry, I tweaked the comment a bit more.^^

@bors r=oli-obk

tests/pass-dep/libc/libc-epoll.rs
```diff
@@ -96,35 +109,28 @@ fn test_epoll_ctl_mod() {
...
     let mut fds = [-1, -1];
-    let mut res =
-        unsafe { libc::socketpair(libc::AF_UNIX, libc::SOCK_STREAM, 0, fds.as_mut_ptr()) };
+    let res = unsafe { libc::socketpair(libc::AF_UNIX, libc::SOCK_STREAM, 0, fds.as_mut_ptr()) };
     assert_eq!(res, 0);
 
     // Write to fd[0].
     let data = "abcde".as_bytes().as_ptr();
-    res = unsafe { libc::write(fds[0], data as *const libc::c_void, 5).try_into().unwrap() };
+    let res = unsafe { libc::write(fds[0], data as *const libc::c_void, 5) };
```
RalfJung, line 117: @tiif btw, the test becomes a lot more readable if you avoid `let mut res` and all the `try_into().unwrap()`. Just use a new fresh `res` variable each time. :)

PR #3813, opened 2024-08-16
Testcase fix for epoll

### LINKED ISSUE #3811: pass-dep/libc/libc-epoll.rs test does not pass when run against real system
Something seems wrong with that test. To make it pass against a real system, I need to apply the following diff:

```diff
diff --git a/tests/pass-dep/libc/libc-epoll.rs b/tests/pass-dep/libc/libc-epoll.rs
index 11a0257dc..57bec1c52 100644
--- a/tests/pass-dep/libc/libc-epoll.rs
+++ b/tests/pass-dep/libc/libc-epoll.rs
@@ -19,6 +19,7 @@ fn main() {
     test_socketpair_read();
 }
 
+#[track_caller]
 fn check_epoll_wait<const N: usize>(
     epfd: i32,
     mut expected_notifications: Vec<(u32, u64)>,
@@ -28,7 +29,10 @@ fn check_epoll_wait<const N: usize>(
     let maxsize = N;
     let array_ptr = array.as_mut_ptr();
     let res = unsafe { libc::epoll_wait(epfd, array_ptr, maxsize.try_into().unwrap(), 0) };
-    assert_eq!(res, expected_notifications.len().try_into().unwrap());
+    if res < 0 {
+        panic!("epoll_wait failed: {}", std::io::Error::last_os_error());
+    }
+    assert_eq!(res, expected_notifications.len().try_into().unwrap(), "didn't get the expected number of notifications");
     let slice = unsafe { std::slice::from_raw_parts(array_ptr, res.try_into().unwrap()) };
     let mut return_events = slice.iter();
     while let Some(return_event) = return_events.next() {
@@ -84,7 +88,7 @@ fn test_epoll_socketpair() {
     assert_eq!(res, 0);
 
     // Check result from epoll_wait.
-    let expected_event = u32::try_from(libc::EPOLLRDHUP | libc::EPOLLIN | libc::EPOLLOUT).unwrap();
+    let expected_event = u32::try_from(libc::EPOLLRDHUP | libc::EPOLLHUP | libc::EPOLLIN | libc::EPOLLOUT).unwrap();
     let expected_value = u64::try_from(fds[1]).unwrap();
     assert!(check_epoll_wait::<8>(epfd, vec![(expected_event, expected_value)]));
 }
@@ -135,7 +139,7 @@ fn test_epoll_ctl_mod() {
     assert_eq!(res, 0);
 
     // Check result from epoll_wait.
-    let expected_event = u32::try_from(libc::EPOLLRDHUP | libc::EPOLLIN | libc::EPOLLOUT).unwrap();
+    let expected_event = u32::try_from(libc::EPOLLRDHUP | libc::EPOLLHUP | libc::EPOLLIN | libc::EPOLLOUT).unwrap();
     let expected_value = u64::try_from(fds[1]).unwrap();
     assert!(check_epoll_wait::<8>(epfd, vec![(expected_event, expected_value)]));
 }
@@ -168,7 +172,8 @@ fn test_epoll_ctl_del() {
     assert_ne!(res, -1);
 
     // Test EPOLL_CTL_DEL.
-    assert!(check_epoll_wait::<0>(epfd, vec![]));
+    //assert!(check_epoll_wait::<0>(epfd, vec![]));
+    // epoll_wait failed: Invalid argument (os error 22)
 }
 
 // This test is for one fd registered under two different epoll instance.
@@ -534,7 +539,7 @@ fn test_socketpair_read() {
     // But in real system, no notification will be provided here.
     let expected_event = u32::try_from(libc::EPOLLOUT).unwrap();
     let expected_value = fds[1] as u64;
-    assert!(check_epoll_wait::<8>(epfd, vec![(expected_event, expected_value)]));
+    assert!(check_epoll_wait::<8>(epfd, vec![]));
 
     // Read until the buffer is empty.
     let mut buf: [u8; 2] = [0; 2];
```
The last one is expected, it seems, given the comment -- that's the case we discussed. But the others are surprising to me, they should at least have comments explaining why we are diverging from the behavior of the real system.

There's also something else odd with the test: the `test_epoll_ctl_del` test doesn't ever use `EPOLL_CTL_DEL`!
Also, the `test_epoll_ctl_mod` test seems to be testing EPOLLRDHUP as well. I don't understand the organization of this test at all.

`check_epoll_wait` doesn't give very nice errors when something goes wrong, so this is also not easy to debug.

Cc @tiif

tiif:
Fixes #3811 

This PR: 
- Fixed the error in ``epoll_ctl_del`` test
- Simplified the logic in ``epoll_ctl_mod`` test
- Added a new test to check if flag that we don't register won't trigger notification (double negation sounds a bit confusing here, feel free to suggest a better one ;) )
- Use assert_eq(0) for epoll_ctl test

### DIFF
--- a/tests/pass-dep/libc/libc-epoll.rs
+++ b/tests/pass-dep/libc/libc-epoll.rs
@@ -14,6 +14,7 @@ fn main() {
     test_not_fully_closed_fd();
     test_closed_fd();
     test_two_epoll_instance();
+    test_no_notification_for_unregister_flag();
     test_epoll_ctl_mod();
     test_epoll_ctl_del();
     test_pointer();
@@ -68,7 +69,7 @@ fn test_epoll_socketpair() {
         u64: u64::try_from(fds[1]).unwrap(),
     };
     let res = unsafe { libc::epoll_ctl(epfd, libc::EPOLL_CTL_ADD, fds[1], &mut ev) };
-    assert_ne!(res, -1);
+    assert_eq!(res, 0);
 
     // Check result from epoll_wait.
     let expected_event = u32::try_from(libc::EPOLLIN | libc::EPOLLOUT).unwrap();
@@ -99,6 +100,8 @@ fn test_epoll_socketpair() {
     check_epoll_wait::<8>(epfd, &[(expected_event, expected_value)]);
 }
 
+// This test first registers a file description with a flag that does not lead to notification,
+// then EPOLL_CTL_MOD to add another flag that will lead to notification.
 fn test_epoll_ctl_mod() {
     // Create an epoll instance.
     let epfd = unsafe { libc::epoll_create1(0) };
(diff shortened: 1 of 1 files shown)


tiif:
https://github.com/rust-lang/miri/pull/3813/commits/7db8243d25867839cf7489f9e68520e1e25dcaa7 is a set of test that will pass real system. The only thing that we can do here is support ``EPOLLHUP`` flag. The other test is something that we discussed before.

tests/pass-dep/libc/libc-epoll.rs
```diff
@@ -18,6 +18,7 @@ fn main() {
     test_epoll_ctl_del();
     test_pointer();
     test_two_same_fd_in_same_epoll_instance();
+    test_no_notification_for_unregister_flag();
```
RalfJung, line 21: This test is "simpler" than test_epoll_ctl_mod, in the sense that test_epoll_ctl_mod also tests that not-registered events do not trigger notifications.
So better test this before test_epoll_ctl_mod, or else it won't ever fail...

tests/pass-dep/libc/libc-epoll.rs
```diff
@@ -167,7 +161,9 @@ fn test_epoll_ctl_del() {
     assert_ne!(res, -1);
 
     // Test EPOLL_CTL_DEL.
-    assert!(check_epoll_wait::<0>(epfd, &[]));
+    let res = unsafe { libc::epoll_ctl(epfd, libc::EPOLL_CTL_DEL, fds[1], &mut ev) };
+    assert_ne!(res, -1);
```
RalfJung, line 165: Is there a reason you're not doing `assert_eq!(res, 0)` here? `epoll_ctl` is documented to return 0 on success, so it seems odd to make the test succeed for other values like 2 or -2.

tiif: Yup this makes sense.

RalfJung:
Looks good, thanks!
@bors r+

PR #3814, opened 2024-08-16
Add epoll EPOLLHUP flag support

tiif:
Related discussion in https://github.com/rust-lang/miri/issues/3811#issuecomment-2293854742. 

This PR added support for ``EPOLLHUP`` flag.

### DIFF
--- a/src/shims/unix/linux/epoll.rs
+++ b/src/shims/unix/linux/epoll.rs
@@ -69,17 +69,21 @@ pub struct EpollReadyEvents {
     /// Stream socket peer closed connection, or shut down writing
     /// half of connection.
     pub epollrdhup: bool,
+    /// For stream socket, this event merely indicates that the peer
+    /// closed its end of the channel.
+    pub epollhup: bool,
 }
 
 impl EpollReadyEvents {
     pub fn new() -> Self {
-        EpollReadyEvents { epollin: false, epollout: false, epollrdhup: false }
+        EpollReadyEvents { epollin: false, epollout: false, epollrdhup: false, epollhup: false }
     }
 
     pub fn get_event_bitmask<'tcx>(&self, ecx: &MiriInterpCx<'tcx>) -> u32 {
         let epollin = ecx.eval_libc_u32("EPOLLIN");
         let epollout = ecx.eval_libc_u32("EPOLLOUT");
         let epollrdhup = ecx.eval_libc_u32("EPOLLRDHUP");
+        let epollhup = ecx.eval_libc_u32("EPOLLHUP");
 
         let mut bitmask = 0;
         if self.epollin {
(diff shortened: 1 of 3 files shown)


src/shims/unix/socket.rs
```diff
@@ -76,6 +76,7 @@ impl FileDescription for SocketPair {
             // for read and write.
             epoll_ready_events.epollin = true;
             epoll_ready_events.epollout = true;
+            epoll_ready_events.epollhup = true;
```
RalfJung, line 79: Please set this a few lines up, together with rdhup, to make the order of operations here make some kind of sense.

src/shims/unix/linux/epoll.rs
```diff
@@ -69,17 +69,21 @@ pub struct EpollReadyEvents {
     /// Stream socket peer closed connection, or shut down writing
     /// half of connection.
     pub epollrdhup: bool,
+    /// For stream socket, this event merely indicates that the peer
+    /// closed its end of the channel.
+    pub epollhup: bool,
```
RalfJung, line 74: So, there's some subtle difference between RDHUP and HUP. But I don't really understand it... and it seems we are ignoring that difference and setting the two flags together. Could that become a problem?

tiif: From my understanding, EPOLLRDHUP is more specific than EPOLLHUP. It will be returned even when the peer only closing its writing half. Closing wriitng half will require ``shutdown``. 

For example, 
If we use ``shutdown(sv[1], SHUT_WR);`` only ``EPOLLRDHUP`` will be triggered. If we use ``shutdown(sv[1], SHUT_RDWR);`` or ``close(sv[1])``, both ``EPOLLHUP`` and ``EPOLLRDHUP`` will be triggered. 

Since we don't have ``shutdown`` for now, I think they will have the same behaviour.

src/shims/unix/linux/epoll.rs
```diff
@@ -69,17 +69,21 @@ pub struct EpollReadyEvents {
     /// Stream socket peer closed connection, or shut down writing
     /// half of connection.
     pub epollrdhup: bool,
+    /// For stream socket, this event merely indicates that the peer
+    /// closed its end of the channel.
```
RalfJung, line 73:
```suggestion
    /// closed its end of the channel.
    /// Unlike epollrdhup, this should only be set when the stream is fully closed.
    /// epollrdhup also gets set when only the write half is closed, which is possible
    /// via `shutdown(_, SHUT_WR)`.
```

src/shims/unix/socket.rs
```diff
@@ -71,6 +71,7 @@ impl FileDescription for SocketPair {
         } else {
             // Peer FD has been closed.
```
RalfJung, line 72:
```suggestion
            // Peer FD has been closed. This always sets both the RDHUP and HUP flags
            // as we do not support `shutdown` that could be used to partially close the stream.
```

Issue #3746, opened 2024-07-13
Implement pipe/pipe2


tiif:
As mentioned in https://github.com/rust-lang/miri/issues/3442#issuecomment-2097703842, it is fairly easy to implement ``pipe/pipe2`` after https://github.com/rust-lang/miri/pull/3609. The main difference between ``socketpair`` and ``pipe`` is ``socketpair`` is bidirectional where ``pipe`` is unidirectional, other logics should be similar. The usage is also quite high according to [the shim wishlist](https://github.com/rust-lang/miri/issues/2057).

RalfJung:
> The usage is also quite high according to https://github.com/rust-lang/miri/issues/2057.

I suspect most of that is due to uses of `Command`, which involves constructing a pipe... so adding support for `pipe2` will just lead to a slightly nicer error, saying that `fork` is not supported. But that's still a nice improvement, `fork` is the actual culprit here and I don't think we'll ever support it.

PR #3815, opened 2024-08-16
implement pipe and pipe2

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/3746

### DIFF
--- a/src/shims/unix/unnamed_socket.rs
+++ b/src/shims/unix/unnamed_socket.rs
@@ -1,3 +1,7 @@
+//! This implements "anonymous" sockets, that do not correspond to anything on the host system and
+//! are entirely implemented inside Miri.
+//! We also use the same infrastructure to implement unnamed pipes.
+
 use std::cell::{OnceCell, RefCell};
 use std::collections::VecDeque;
 use std::io;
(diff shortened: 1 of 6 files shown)


src/shims/unix/unnamed_socket.rs
```diff
@@ -55,17 +60,25 @@ impl FileDescription for SocketPair {
...
-            let available_space = MAX_SOCKETPAIR_BUFFER_CAPACITY.strict_sub(data_size);
-            if available_space != 0 {
+            if let Some(writebuf) = &peer_fd.downcast::<SocketPair>().unwrap().readbuf {
+                let data_size = writebuf.borrow().buf.len();
+                let available_space = MAX_SOCKETPAIR_BUFFER_CAPACITY.strict_sub(data_size);
+                if available_space != 0 {
+                    epoll_ready_events.epollout = true;
+                }
+            } else {
+                // Without a write buffer, writing never blocks.
```
oli-obk, line 81: How can we have a peer without a read buffer?

RalfJung: For the read end of a pipe, the peer is the write end of the pipe, and the write end doesn't have a read buffer.

oli-obk: oh... right, that will just error

PR #3818, opened 2024-08-17
epoll: iterate through output buffer then fetch an event from ready list

### LINKED ISSUE #3812: epoll_wait can lose events when they do not all fit into the output buffer
See [here](https://github.com/rust-lang/miri/pull/3712#pullrequestreview-2243021331) for the relevant code and what is wrong with it: we first fetch an event from the queue, and then try to find a spot for it in the buffer. That's the wrong way around; we should iterate over the output buffer and only if we are sure we can store an event, then fetch one from the queue.

Cc @tiif

tiif:
Fixes #3812

### DIFF
--- a/src/shims/unix/linux/epoll.rs
+++ b/src/shims/unix/linux/epoll.rs
@@ -436,23 +436,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let mut num_of_events: i32 = 0;
         let mut array_iter = this.project_array_fields(&events)?;
 
-        while let Some((epoll_key, epoll_return)) = ready_list.pop_first() {
-            // If the file description is fully close, the entry for corresponding FdID in the
-            // global epoll event interest table would be empty.
-            if this.machine.epoll_interests.get_epoll_interest(epoll_key.0).is_some() {
-                // Return notification to the caller if the file description is not fully closed.
-                if let Some(des) = array_iter.next(this)? {
-                    this.write_int_fields_named(
-                        &[
-                            ("events", epoll_return.events.into()),
-                            ("u64", epoll_return.data.into()),
-                        ],
-                        &des.1,
-                    )?;
-                    num_of_events = num_of_events.checked_add(1).unwrap();
-                } else {
-                    break;
-                }
+        while let Some(des) = array_iter.next(this)? {
+            if let Some(epoll_event_instance) = ready_list_next(this, &mut ready_list) {
+                this.write_int_fields_named(
+                    &[
+                        ("events", epoll_event_instance.events.into()),
+                        ("u64", epoll_event_instance.data.into()),
+                    ],
+                    &des.1,
+                )?;
+                num_of_events = num_of_events.strict_add(1);
+            } else {
+                break;
             }
         }
         Ok(Scalar::from_i32(num_of_events))
@@ -494,4 +489,22 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         }
         Ok(())
     }
+}
+
+/// This function takes in ready list and returns EpollEventInstance with file description
+/// that is not closed.
+fn ready_list_next(
+    ecx: &MiriInterpCx<'_>,
+    ready_list: &mut BTreeMap<(FdId, i32), EpollEventInstance>,
+) -> Option<EpollEventInstance> {
+    while let Some((epoll_key, epoll_event_instance)) = ready_list.pop_first() {
+        // This ensures that we only return events that we are interested. The FD might have been closed since
+        // the event was generated, in which case we are not interested anymore.
+        // When a file description is fully closed, it gets removed from `machine.epoll_interests`,
+        // so we skip events whose FD is not in that map anymore.
+        if ecx.machine.epoll_interests.get_epoll_interest(epoll_key.0).is_some() {
+            return Some(epoll_event_instance);
+        }
+    }
+    return None;
 }
--- a/tests/pass-dep/libc/libc-epoll.rs
+++ b/tests/pass-dep/libc/libc-epoll.rs
@@ -20,6 +20,8 @@ fn main() {
     test_pointer();
     test_two_same_fd_in_same_epoll_instance();
     test_epoll_wait_maxevent_zero();
+    test_epoll_lost_events();
+    test_ready_list_fetching_logic();
 }
 
 // Using `as` cast since `EPOLLET` wraps around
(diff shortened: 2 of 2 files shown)


src/shims/unix/linux/epoll.rs
```diff
@@ -411,12 +411,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
 
-        while let Some((epoll_key, epoll_return)) = ready_list.pop_first() {
-            // If the file description is fully close, the entry for corresponding FdID in the
-            // global epoll event interest table would be empty.
-            if this.machine.epoll_interests.get_epoll_interest(epoll_key.0).is_some() {
-                // Return notification to the caller if the file description is not fully closed.
-                if let Some(des) = array_iter.next(this)? {
+        while let Some(des) = array_iter.next(this)? {
+            // Fetch an event from the ready list.
+            if let Some((epoll_key, epoll_return)) = ready_list.pop_first() {
```
oli-obk, line 416: This will cause empty entries whenever a FD was already closed. You'll need to use a for loop here and exit it only if you actually write an entry.

This will cause us to finish iterating over the array_iter and never doing anything once the ready list is empty. That's fine, but we could also ensure we `continue` to the outer loop (by giving it a label) and unconditionally breaking after the ready list loop

oli-obk: (fwiw: this is why we abandoned this approach before, but forgot that in abandoning it we'd not fix this issue)

RalfJung: Indeed, this needs to be a *loop* getting the next entry from the ready list that refers to an FD that is still open.

(I assume there's a test covering this case? That test would generate an interesting event on an FD, and then close the FD, and only then call epoll_wait and expect to *not* see that event.)

tiif: Right, the test that I added initially failed because there is empty entry where it shouldn't. I added a condition to check if the entry is actually written, also terminate earlier if the ready list is empty (although this won't affect correctness).

src/shims/unix/linux/epoll.rs
```diff
@@ -411,20 +411,27 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                            ("events", epoll_return.events.into()),
-                            ("u64", epoll_return.data.into()),
-                        ],
-                        &des.1,
-                    )?;
-                    num_of_events = num_of_events.checked_add(1).unwrap();
+        while let Some(des) = array_iter.next(this)?
+            && !ready_list.is_empty()
+        {
+            let mut entry_written = false;
```
RalfJung, line 417: I don't think using mutable state is necessary here. That makes reasoning about the code harder. The `!is_empty` check above also adds extra complexity that should not be needed.

Instead, I suggest you add a function like `ready_list_next` that returns the next item from the ready list that someone is still interested in:
```rust
while let Some((epoll_key, epoll_return)) = ready_list.pop_first() {
  // Ensure we are still interested in this event. The FD might have been closed since
  // the event was generated, in which case we are not interested any more.
  if this.machine.epoll_interests.get_epoll_interest(epoll_key.0).is_some() {
    return Some((epoll_key, epoll_return));
  }
}
return None;
```
Then you just use that method here, the loop here should become fairly simple.

src/shims/unix/linux/epoll.rs
```diff
@@ -411,20 +411,27 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                        // Return notification to the caller if the file description is not fully closed.
+                        this.write_int_fields_named(
+                            &[
+                                ("events", epoll_return.events.into()),
+                                ("u64", epoll_return.data.into()),
+                            ],
+                            &des.1,
+                        )?;
+                        num_of_events = num_of_events.checked_add(1).unwrap();
+                        entry_written = true;
```
oli-obk, line 433:
```suggestion
                        break;
```

src/shims/unix/linux/epoll.rs
tiif: Updated. By the way, if the ``for`` loop you meant is for ``while let Some(des) = array_iter.next(this)? ``, it might not be possible because [ArrayIterator](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_const_eval/interpret/eval_context/struct.InterpCx.html#method.project_array_fields), which is the type of ``array_iter``, doesn't implement ``IntoIterator``.

RalfJung: Ah right this has to remain `while let`, good point.

src/shims/unix/linux/epoll.rs
```diff
@@ -431,28 +431,42 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            } else {
+                break;
             }
         }
         Ok(Scalar::from_i32(num_of_events))
     }
 
+    /// This function takes in ready list and returns EpollEventInstance with file description
+    /// that is not closed.
+    fn ready_list_next(
```
RalfJung, line 453: You are adding this as a public function that can be called everywhere in Miri -- that's not the intention, as this is just a private helper. Please make the function private.

src/shims/unix/linux/epoll.rs
```diff
@@ -431,28 +431,42 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        &self,
+        ready_list: &mut BTreeMap<(FdId, i32), EpollEventInstance>,
+    ) -> Option<EpollEventInstance> {
+        let this = self.eval_context_ref();
+        while let Some((epoll_key, epoll_event_instance)) = ready_list.pop_first() {
+            // This ensures that we only return events that we are interested. The FD might have been closed since
+            // the event was generated, in which case we are not interested anymore.
```
RalfJung, line 460:
```suggestion
            // the event was generated, in which case we are not interested anymore.
            // When a file description is fully closed, it gets removed from `machine.epoll_interests`,
            // so we skip events whose FD is not in that map any more.
```

src/shims/unix/linux/epoll.rs
```diff
@@ -431,28 +431,42 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            if this.machine.epoll_interests.get_epoll_interest(epoll_key.0).is_some() {
+                // If the file description is fully close, the entry for corresponding FdID in the
+                // global epoll event interest table would be empty.
```
RalfJung, line 463:
```suggestion
```

src/shims/unix/linux/epoll.rs
```diff
@@ -431,28 +431,42 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        while let Some(des) = array_iter.next(this)? {
+            if let Some(epoll_event_instance) = this.ready_list_next(&mut ready_list) {
+                this.write_int_fields_named(
+                    &[
+                        ("events", epoll_event_instance.events.into()),
+                        ("u64", epoll_event_instance.data.into()),
+                    ],
+                    &des.1,
+                )?;
+                num_of_events = num_of_events.checked_add(1).unwrap();
```
RalfJung, line 443:
```suggestion
                num_of_events = num_of_events.strict_add(1);
```

tests/pass-dep/libc/libc-epoll.rs
```diff
@@ -501,3 +503,65 @@ fn test_socketpair_read() {
...
+
+    // Create a socketpair instance.
+    let mut fds = [-1, -1];
+    let res = unsafe { libc::socketpair(libc::AF_UNIX, libc::SOCK_STREAM, 0, fds.as_mut_ptr()) };
+    assert_eq!(res, 0);
+
+    // Register both fd to the same epoll instance.
+    let mut ev = libc::epoll_event { events: EPOLL_IN_OUT_ET, u64: fds[0] as u64 };
+    let res = unsafe { libc::epoll_ctl(epfd, libc::EPOLL_CTL_ADD, fds[0], &mut ev) };
+    assert_ne!(res, -1);
```
RalfJung, line 522:
```suggestion
    assert_eq!(res, 0);
```
also in the rest of the test

src/shims/unix/linux/epoll.rs
tiif: I moved this out and turned this a free function. Alternatively, we can create a wrapper type for ``ready_list`` and make this a method of that type.

RalfJung: A free function sounds good for this!

RalfJung:
Looks good, thanks. :) Please squash the commits.

tiif:
@RalfJung The CI is green here, would you like to land this first? I will go back https://github.com/rust-lang/miri/pull/3820 to resolve potential merge conflict after this landed.

RalfJung:
Okay :)
@bors r+

PR #3819, opened 2024-08-17
epoll test: further clean up check_epoll_wait

RalfJung:
Given that `check_epoll_wait` compared the length of the two slices, I don't think it was possible for it to ever return `false`. It's also strange to have some requirements checked inside the function and some checked by the caller, so let's just move it all inside the function.

Cc @tiif -- did I miss anything?


tiif:
Correct, this looks good!

RalfJung:
Great, thanks for checking. :)

@bors r+

PR #3820, opened 2024-08-17
Add epollerr support

### LINKED ISSUE #3816: epoll does not report EPOLLERR events
According to the [man page](https://man7.org/linux/man-pages/man2/epoll_ctl.2.html), epoll will always report EPOLLERR events. However, our implementation never reports them. This is a problem as Miri will silently accept the program but run it in an incorrect way. There's no good way to throw a "not supported" error here, so we should implement this event notification.

Cc @tiif

tiif:
Fixes #3816 

For socketpair, if the peer fd is closed while having data in its read buffer, ``EPOLLER`` will be thrown. 

This is still WIP because I am currently finding a way to check if peer fd still has something in its ``readbuf`` when it is closed and add the ``EPOLLER`` flag In ``get_epoll_ready_events``.

### DIFF
--- a/src/shims/unix/linux/epoll.rs
+++ b/src/shims/unix/linux/epoll.rs
@@ -76,11 +76,19 @@ pub struct EpollReadyEvents {
     /// epollrdhup also gets set when only the write half is closed, which is possible
     /// via `shutdown(_, SHUT_WR)`.
     pub epollhup: bool,
+    /// Error condition happened on the associated file descriptor.
+    pub epollerr: bool,
 }
 
 impl EpollReadyEvents {
     pub fn new() -> Self {
-        EpollReadyEvents { epollin: false, epollout: false, epollrdhup: false, epollhup: false }
+        EpollReadyEvents {
+            epollin: false,
+            epollout: false,
+            epollrdhup: false,
+            epollhup: false,
+            epollerr: false,
+        }
     }
 
     pub fn get_event_bitmask<'tcx>(&self, ecx: &MiriInterpCx<'tcx>) -> u32 {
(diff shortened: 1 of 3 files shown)


RalfJung:
> For socketpair, if the peer fd is closed while having data in its read buffer, EPOLLER will be thrown.

Ah, very interesting. If the peer FD is closed without any data in its read buffer, all is fine, but if there is data being discarded, that's not fine? I guess I can see how it makes sense, but it'd be better if we'd find documentation explaining this...

First sketch of an implementation:
- add a new field to `SocketPair`: `peer_lost_data: Cell<bool>`, initially `false`
- in close, if our own read buffer is non-empty and we still have a peer, set the peer's `peer_lost_data` flag to `true`
- in `get_epoll_ready_events`, if `peer_lost_data` is true, set EPOLLERR

tiif:
> If the peer FD is closed without any data in its read buffer, all is fine, but if there is data being discarded, that's not fine?

Yes. If the peer FD read all its data before it get closed, ``EPOLLER`` won't be triggered. 

> Does EPOLERR ever get unset again after this?

In the context of the current test case we have in this PR, I think ``EPOLLER`` can only be triggered once because we only can ``close`` the peer_fd for once. 

@rustbot author

RalfJung:
Looks good, thanks :)
@bors r+

tiif:
Maybe it will conflict again after https://github.com/rust-lang/miri/pull/3818 landed, I will come back later to check.

RalfJung:
We'll see. :)
@bors r+

RalfJung:
Ah yes they both add tests to the end of the same file so they will definitely conflict there.

tiif:
@RalfJung Alright, last one. Thanks for being fast in reviewing this stream of PRs. :)

RalfJung:
Thanks for being fast in reacting to feedback and conflicts. :)
@bors r+

PR #3824, opened 2024-08-17
Set EINVAL for epoll_wait maxevent value 0

### LINKED ISSUE #3821: epoll_wait with maxevents set to 0 should return EINVAL
See [here](https://github.com/rust-lang/miri/issues/3811#issuecomment-2293748910): when epoll_wait is given no space for events, we currently do nothing, but the kernel returns EINVAL. We should match kernel behavior.

The documentation says that error is returned when

> epfd is not an epoll file descriptor, or maxevents is less than or equal to zero.

Cc @tiif

tiif:
Fixes #3821

### DIFF
--- a/src/shims/unix/linux/epoll.rs
+++ b/src/shims/unix/linux/epoll.rs
@@ -403,17 +403,19 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
         let epfd = this.read_scalar(epfd)?.to_i32()?;
         let maxevents = this.read_scalar(maxevents)?.to_i32()?;
-        let event = this.deref_pointer_as(
-            events_op,
-            this.libc_array_ty_layout("epoll_event", maxevents.try_into().unwrap()),
-        )?;
         let timeout = this.read_scalar(timeout)?.to_i32()?;
-
         if epfd <= 0 || maxevents <= 0 {
             let einval = this.eval_libc("EINVAL");
             this.set_last_error(einval)?;
             return Ok(Scalar::from_i32(-1));
         }
+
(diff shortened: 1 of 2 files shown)


tests/pass-dep/libc/libc-epoll.rs
```diff
@@ -165,7 +166,9 @@ fn test_epoll_ctl_del() {
     assert_ne!(res, -1);
 
     // Test EPOLL_CTL_DEL.
-    check_epoll_wait::<0>(epfd, &[]);
+    let res = unsafe { libc::epoll_ctl(epfd, libc::EPOLL_CTL_DEL, fds[1], &mut ev) };
+    assert_eq!(res, 0);
+    check_epoll_wait::<8>(epfd, &[]);
```
tiif, line 171: This is a duplicate of #3813 to let all the testcases pass.

RalfJung: Okay, please rebase once https://github.com/rust-lang/miri/pull/3813 landed then.

src/shims/unix/linux/epoll.rs
```diff
@@ -409,11 +409,12 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         )?;
         let timeout = this.read_scalar(timeout)?.to_i32()?;
 
-        if epfd <= 0 {
+        if epfd <= 0 || maxevents <= 0 {
```
RalfJung, line 412: The `maxevents.try_into().unwrap()` above needs to be moved down after this check, it would ICE on a negative `maxevents`.

RalfJung:
Thanks :)

@bors r+

PR #3829, opened 2024-08-20
epoll: handle edge case for epoll_ctl

tiif:
There is a test case that revealed that our implementation differs from the real system:
 - Set up an epoll watching the FD
- Call epoll_wait
- Set up another epoll watching the same FD
- Call epoll_wait on the first epoll. Nothing should be reported!

This happened because, in ``epoll_ctl``, we used ``check_and_update_readiness``, which is a function that would return notification for all epoll file description that registered a particular file description. But we shouldn't do that because no notification should be returned if there is no I/O activity between two ``epoll_wait`` (every first ``epoll_wait`` that happens after ``epoll_ctl`` is an exception, we should return notification that reflects the readiness of file description). 

r? @oli-obk

### DIFF
--- a/src/shims/unix/linux/epoll.rs
+++ b/src/shims/unix/linux/epoll.rs
@@ -269,10 +269,10 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let mut interest_list = epoll_file_description.interest_list.borrow_mut();
         let ready_list = &epoll_file_description.ready_list;
 
-        let Some(file_descriptor) = this.machine.fds.get(fd) else {
+        let Some(fd_ref) = this.machine.fds.get(fd) else {
             return Ok(Scalar::from_i32(this.fd_not_found()?));
         };
-        let id = file_descriptor.get_id();
+        let id = fd_ref.get_id();
 
         if op == epoll_ctl_add || op == epoll_ctl_mod {
             // Read event bitmask and data from epoll_event passed by caller.
@@ -332,7 +332,6 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 }
             }
 
-            let id = file_descriptor.get_id();
             // Create an epoll_interest.
             let interest = Rc::new(RefCell::new(EpollEventInterest {
                 file_descriptor: fd,
@@ -344,7 +343,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             if op == epoll_ctl_add {
                 // Insert an epoll_interest to global epoll_interest list.
                 this.machine.epoll_interests.insert_epoll_interest(id, Rc::downgrade(&interest));
-                interest_list.insert(epoll_key, interest);
+                interest_list.insert(epoll_key, Rc::clone(&interest));
             } else {
                 // Directly modify the epoll_interest so the global epoll_event_interest table
                 // will be updated too.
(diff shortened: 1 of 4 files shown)


src/shims/unix/linux/epoll.rs
```diff
@@ -525,3 +508,28 @@ fn ready_list_next(
...
+/// epoll_interest and, if necessary, triggers the notification. Unlike check_and_update_readiness,
+/// this function sends a notification to only one epoll instance.
+fn check_fd_ready_event(
+    fd_ref: &FileDescriptionRef,
+    interest: Rc<RefCell<EpollEventInterest>>,
+    id: FdId,
+    ecx: &MiriInterpCx<'_>,
+) {
+    // Get the bitmask of ready events for a file description.
+    let ready_events_bitmask = fd_ref.get_epoll_ready_events().unwrap().get_event_bitmask(ecx);
```
tiif, line 522: ``get_epoll_ready_events`` will either throw_unsup_format or return an Ok(), Since we use ``unwrap()`` here this function only can be used to check file description that support epoll. Alternatively, we can let this function return an InterpResult and use ? instead of unwrap(). I am not sure which approach is better.

src/shims/unix/linux/epoll.rs
```diff
@@ -525,3 +508,28 @@ fn ready_list_next(
     }
     return None;
 }
+
+/// This helper function checks whether an epoll notification should be triggered for a specific
+/// epoll_interest and, if necessary, triggers the notification. Unlike check_and_update_readiness,
+/// this function sends a notification to only one epoll instance.
+fn check_fd_ready_event(
```
tiif, line 515: I have a hard time coming up with a good naming to differentiate this function from ``check_and_update_readiness``. They essentially do the same thing, but this function only checks one ``epoll_event_interest``, where ``check_and_update_readiness`` checks all ``epoll_event_interest``. As a result, this function can only return notification to one epoll instance where ``check_and_update_readiness`` can return notification to multiple epoll instances.

RalfJung:
> Previously, in epoll_ctl, we used check_and_update_readiness, which is a function that would return notification for all epoll file description. But we shouldn't do that because no notification should be returned if there is no I/O activity between two epoll_wait. 

I am confused by this description. There's only one epoll_wait in the first sentence so the second sentence seems to talk about a different case?

Looking at the testcase I think I understand now, but the PR description doesn't explain the problem. The problem is
- Set up an epoll watching the FD
- Call epoll_wait
- Set up another epoll watching the same FD
- Call epoll_wait on the first epoll. Nothing should be reported!

tests/pass-dep/libc/libc-epoll.rs
```diff
@@ -650,3 +655,46 @@ fn test_ready_list_fetching_logic() {
     let expected_value1 = fd1 as u64;
     check_epoll_wait::<1>(epfd, &[(expected_event1, expected_value1)]);
 }
+
+// epoll has different behaviour in multithreading. We previously check_and_update_readiness the
+// moment a file description is registered in an epoll instance. But this has an unfortunate side
+// effect of returning notification to another epfd that shouldn't receive notification.
+fn test_epoll_ctl_multithread() {
```
RalfJung, line 662: The problem doesn't seem related to concurrency at all, so why is this test multithreaded?

tiif: Right, I discovered this when I am testing it on multi-threaded testcase, but I found a way to write a single-threaded testcase for this.

src/shims/unix/linux/epoll.rs
oli-obk: Oh right, we don't have a test for inserting an FD (that doesn't support epoll) into an epoll FD. Please add such a test. Pretty sure it should hit that unwrap

tiif: Yup, it hits the unwrap. ``InterpResult`` is better.

oli-obk (approved):

tiif:
Rebased and CI is green. 

r? @oli-obk

PR #3832, opened 2024-08-21
epoll test: avoid some subtly dangling pointers

RalfJung:
Turns out `let data = MaybeUninit::<u64>::uninit().as_ptr();` is a dangling pointer, the memory gets freed at the end of that line. For these cases we don't care as we don't actually access the pointer, but let's not do such subtle things.

### DIFF
--- a/tests/pass-dep/libc/libc-epoll.rs
+++ b/tests/pass-dep/libc/libc-epoll.rs
@@ -1,8 +1,7 @@
 //@only-target-linux
 
-#![feature(exposed_provenance)] // Needed for fn test_pointer()
+#![feature(strict_provenance)]
 use std::convert::TryInto;
-use std::mem::MaybeUninit;
 
 fn main() {
     test_epoll_socketpair();
@@ -17,7 +16,6 @@ fn main() {
     test_no_notification_for_unregister_flag();
     test_epoll_ctl_mod();
     test_epoll_ctl_del();
-    test_pointer();
     test_two_same_fd_in_same_epoll_instance();
     test_epoll_wait_maxevent_zero();
     test_socketpair_epollerr();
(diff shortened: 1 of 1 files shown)


tests/pass-dep/libc/libc-epoll.rs
```diff
@@ -261,24 +259,6 @@ fn test_epoll_eventfd() {
...
-    let res = unsafe { libc::socketpair(libc::AF_UNIX, libc::SOCK_STREAM, 0, fds.as_mut_ptr()) };
-    assert_eq!(res, 0);
-
-    // Register fd[1] with EPOLLIN|EPOLLOUT|EPOLLET
-    let data = MaybeUninit::<u64>::uninit().as_ptr();
-    let mut ev =
-        libc::epoll_event { events: EPOLL_IN_OUT_ET, u64: data.expose_provenance() as u64 };
-    let res = unsafe { libc::epoll_ctl(epfd, libc::EPOLL_CTL_ADD, fds[1], &mut ev) };
-    assert_eq!(res, 0);
-}
```
RalfJung, line 280: @tiif what is the point of this test? `data.expose_provenance()` is a completely normal integer, so this isn't testing anything that the other tests wouldn't also cover.

RalfJung: We can already get it back out now though, without blocking epoll_wait.

And even then I don't get the point of the test. It's just testing that we're passing through the integer? We're already testing that in the other tests...

oli-obk: I assumed we'd want to test pointers in a way that people would use this API, even though it is broken

RalfJung: :shrug: seems rather redundant to me since we already test int2ptr casts and the u64 data behavior separately and there's no reason to think there could be any interesting interactions here. So having to maintain more tests is a waste of time IMO.

RalfJung: The one case that I think could be interesting is passing epoll_ctl a buffer that actually stores a pointer there. This requires some shenanigans, like defining an `epoll_event_ptr` type and then casting the `*mut epoll_event_ptr` to `*mut epoll_event` and passing that to `epoll_ctl`. There would be no `expose_provenance`. I don't know if the current implementation would preserve the provenance or not in that case.

oli-obk: Yea, that seems better. It's just a libc crate limitation, not something of the libc API/kernel API

oli-obk:
bors r plus

PR #3833, opened 2024-08-21
Make Tree Borrows Provenance GC no longer produce stack overflows

JoJoDeveloping:
Most functions operating on Tree Borrows' trees are carefully written to not cause stack overflows due to too much recursion. The one exception is [`Tree::keep_only_needed`](https://github.com/rust-lang/miri/blob/94f5588fafcc7d59fce60ca8f7af0208e6f618d4/src/borrow_tracker/tree_borrows/tree.rs#L724), which just uses regular recursion.
This function is part of the provenance GC, so it is called regularly for every allocation in the program.

Tests show that this is a problem in practice. For example, the test `fill::horizontal_line` in crate `tiny-skia` (version 0.11.4) is such a test.

This PR changes this, this test no now longer crashes. Instead, it succeeds (after a _long_ time).

### DIFF
--- a/src/borrow_tracker/tree_borrows/tree.rs
+++ b/src/borrow_tracker/tree_borrows/tree.rs
@@ -10,7 +10,7 @@
 //!   and the relative position of the access;
 //! - idempotency properties asserted in `perms.rs` (for optimizations)
 
-use std::fmt;
+use std::{fmt, mem};
 
 use smallvec::SmallVec;
 
@@ -699,8 +699,7 @@ impl<'tcx> Tree {
 /// Integration with the BorTag garbage collector
 impl Tree {
     pub fn remove_unreachable_tags(&mut self, live_tags: &FxHashSet<BorTag>) {
-        let root_is_needed = self.keep_only_needed(self.root, live_tags); // root can't be removed
-        assert!(root_is_needed);
+        self.remove_useless_children(self.root, live_tags);
         // Right after the GC runs is a good moment to check if we can
         // merge some adjacent ranges that were made equal by the removal of some
         // tags (this does not necessarily mean that they have identical internal representations,
@@ -708,9 +707,16 @@ impl Tree {
         self.rperms.merge_adjacent_thorough();
     }
 
+    /// Checks if a node is useless and should be GC'ed.
+    /// A node is useless if it has no children and also the tag is no longer live.
+    fn is_useless(&self, idx: UniIndex, live: &FxHashSet<BorTag>) -> bool {
+        let node = self.nodes.get(idx).unwrap();
+        node.children.is_empty() && !live.contains(&node.tag)
+    }
+
     /// Traverses the entire tree looking for useless tags.
-    /// Returns true iff the tag it was called on is still live or has live children,
-    /// and removes from the tree all tags that have no live children.
+    /// Removes from the tree all useless child nodes of root.
+    /// It will not delete the root itself.
     ///
     /// NOTE: This leaves in the middle of the tree tags that are unreachable but have
     /// reachable children. There is a potential for compacting the tree by reassigning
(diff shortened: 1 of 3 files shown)


JoJoDeveloping:
Note: I'm not too sure what the point of the assert here was, but with the new implementation, it is already ensured otherwise that the root is never deleted.

https://github.com/rust-lang/miri/blob/a84aba2f5b8a64180e141fff68ebf9eca0bf2e4d/src/borrow_tracker/tree_borrows/tree.rs#L703

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -721,43 +729,50 @@ impl Tree {
...
-            // `UniValMap::remove` for the key of this tag on *all* maps that used it
-            // (which are `self.nodes` and every range of `self.rperms`)
-            // before we can safely apply `UniValMap::forget` to truly remove
-            // the tag from the mapping.
-            let tag = node.tag;
-            self.nodes.remove(idx);
-            for (_perms_range, perms) in self.rperms.iter_mut_all() {
-                perms.remove(idx);
+    fn keep_only_needed(&mut self, root: UniIndex, live: &FxHashSet<BorTag>) -> bool {
+        let mut stack = vec![(root, 0)];
```
RalfJung, line 733: Please add a comment explaining why we don't use the `TreeVisitorStack`

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -721,43 +729,50 @@ impl Tree {
...
+                // Next, ensure all useless children are properly unregistered everywhere.
+                for idx in useless {
+                    // Note: In the rest of this comment, "this node" means that at idx.
+                    // All of the children and this node are unreachable, delete this tag
+                    // from the tree (the children have already been deleted by recursive
+                    // calls).
+                    // Due to the API of UniMap we must absolutely call
+                    // `UniValMap::remove` for the key of this tag on *all* maps that used it
+                    // (which are `self.nodes` and every range of `self.rperms`)
+                    // before we can safely apply `UniValMap::forget` to truly remove
```
Vanille-N, line 755: Apparently this comment has gotten out of sync when we renamed `UniValMap::forget`
```suggestion
                    // before we can safely apply `UniValMap::remove` to truly remove
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -721,43 +729,50 @@ impl Tree {
...
+    fn keep_only_needed(&mut self, root: UniIndex, live: &FxHashSet<BorTag>) -> bool {
+        let mut stack = vec![(root, 0)];
+        // To avoid stack overflows, we roll our own stack.
+        // Each frame consists of the current tag, and the number of the next child.
+        while let Some((tag, nth_child)) = stack.last_mut() {
+            let node = self.nodes.get(*tag).unwrap();
+            if *nth_child == node.children.len() {
+                // We have visited all children. Let's remove those that are useless.
+                // Split the children into useful and useless.
+                let (useless, new_children) =
```
RalfJung, line 741:
```suggestion
                // FIXME: this copies new_children, even if nothing gets removed.
                let (useless, new_children) =
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -721,43 +729,50 @@ impl Tree {
...
+                    node.children.iter().copied().partition(|x| self.is_useless(*x, live));
+                // Assign the new children.
+                // This might make this node useless as well, but removal is the parent's job.
+                self.nodes.get_mut(*tag).unwrap().children = new_children;
+                // Next, ensure all useless children are properly unregistered everywhere.
+                for idx in useless {
+                    // Note: In the rest of this comment, "this node" means that at idx.
```
RalfJung, line 748:
```suggestion
                    // Note: In the rest of this comment, "this node" refers to `idx`.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -721,42 +727,56 @@ impl Tree {
...
-            // (which are `self.nodes` and every range of `self.rperms`)
-            // before we can safely apply `UniValMap::forget` to truly remove
-            // the tag from the mapping.
-            let tag = node.tag;
-            self.nodes.remove(idx);
-            for (_perms_range, perms) in self.rperms.iter_mut_all() {
-                perms.remove(idx);
+    fn remove_useless_children(&mut self, root: UniIndex, live: &FxHashSet<BorTag>) {
+        // To avoid stack overflows, we roll our own stack.
+        // Each element in the stack consists of the current tag, and the number of the next child.
```
RalfJung, line 732:
```suggestion
        // Each element in the stack consists of the current tag, and the number
        // of the next child that should be processed.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -721,42 +727,56 @@ impl Tree {
...
+
+        // The other functions are written using the `TreeVisitorStack`, which does not work
```
RalfJung, line 734:
```suggestion
        // The other functions are written using the `TreeVisitorStack`, but that does not work here
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -721,42 +727,56 @@ impl Tree {
...
+                }
+                // We are done, the parent can continue.
+                stack.pop();
+                continue;
+            } else {
+                // We are starting to visit that subtree.
+                // For that, we push a new frame with the initial conditions, and remember that we continue with the next once that subtree is done.
+                let next_child = node.children[*nth_child];
+                *nth_child += 1;
+                stack.push((next_child, 0));
```
RalfJung, line 777:
```suggestion
                // Visit the child by pushing it to the stack.
                // Also increase `nth_child` so that when we come back to the `tag` node, we
                // look at the next child.
                let next_child = node.children[*nth_child];
                *nth_child += 1;
                stack.push((next_child, 0));
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -721,42 +727,56 @@ impl Tree {
...
+        // Each element in the stack consists of the current tag, and the number of the next child.
+
+        // The other functions are written using the `TreeVisitorStack`, which does not work
+        // since we need to 1) do a post-traversal and 2) remove nodes from the tree.
+        // Since we do a post-traversal (by deleting nodes only after handling all children),
+        // we also need to be a bit smarter than "pop node, push all children."
+        let mut stack = vec![(root, 0)];
+        while let Some((tag, nth_child)) = stack.last_mut() {
+            let node = self.nodes.get(*tag).unwrap();
+            if *nth_child == node.children.len() {
```
RalfJung, line 741: I think it'd be better if swap the `if` and `else` branches here. Then reading the code top-to-bottom corresponds to the order it is executed in.

src/borrow_tracker/tree_borrows/unimap.rs
```diff
@@ -195,6 +195,18 @@ impl<V> UniValMap<V> {
         }
         self.data[idx.idx as usize] = None;
     }
+
+    /// Delete any value associated with this index.
+    /// Returns None if the value was not present, otherwise
+    /// returns the previously stored value.
+    pub fn remove_owned(&mut self, idx: UniIndex) -> Option<V> {
```
RalfJung, line 202: Please just call this `remove`.

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: This would presumably also need `Vec::drain_filter` which is currently unstable.

RalfJung: Yeah, though we're using other unstable features in Miri as well so that wouldn't be a blocker.

JoJoDeveloping: I've played around with the filter method. Sadly it requires two `mem::swap`s to make it work. See specifically [this commit](https://github.com/rust-lang/miri/pull/3833/commits/c9c21b5a7145616bdf980e6a7d7b72d56b46ec50).

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -729,38 +729,42 @@ impl Tree {
...
+                // Visit the child by pushing it to the stack.
+                // Also increase `nth_child` so that when we come back to the `tag` node, we
+                // look at the next child.
+                let next_child = node.children[*nth_child];
+                *nth_child += 1;
+                stack.push((next_child, 0));
+                continue;
+            } else {
+                // First, get the current children of `node`.
+                mem::swap(&mut self.nodes.get_mut(*tag).unwrap().children, &mut children_of_node);
```
RalfJung, line 753: This is probably more elegant as `let mut children_of_node = mem::take(&mut ...);`

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -729,38 +729,42 @@ impl Tree {
...
+                let to_remove: Vec<_> =
+                    children_of_node.drain_filter(|x| self.is_useless(*x, live)).collect();
+                mem::swap(&mut self.nodes.get_mut(*tag).unwrap().children, &mut children_of_node);
```
RalfJung, line 756: And then this can just be `...children = children_of_node`, right?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -729,38 +729,42 @@ impl Tree {
...
+            if *nth_child < node.children.len() {
+                // Visit the child by pushing it to the stack.
+                // Also increase `nth_child` so that when we come back to the `tag` node, we
+                // look at the next child.
+                let next_child = node.children[*nth_child];
+                *nth_child += 1;
+                stack.push((next_child, 0));
+                continue;
+            } else {
+                // First, get the current children of `node`.
```
RalfJung, line 752:
```suggestion
                // We have processes all children of `node`, so now it is time to process `node` itself.
                // First, get the current children of `node`. To appease the borrow checker,
                // we have to temporarily move the list out of the node, and then put the
                // list of remaining children back in.
```

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: Maybe. Since it's a `SmallVec` you can probably create (and discard) the empty ones relatively cheaply. But also, since it's a `SmallVec`, the two copies/swaps are not going to be completely cheap.

RalfJung: A global `let mut` outside the loop that survives all loop iterations is not a good idea IMO. So `children_of_node` should be per-iteration. That means it is dead after this `swap` call, so it makes little sense to use `swap` in the first place.

JoJoDeveloping: Is it better now?

RalfJung: yes. :)

There's just one thing I am wondering:
can we do the `self.nodes.get_mut(*tag).unwrap().children = children_of_node;` *after* the `for idx in to_remove {`? If yes, then we could entirely avoid the `collect`, which seems quite nice.

JoJoDeveloping: We could, but it does not avoid the `collect`. That is needed because the closure passed to `drain_filter` needs a shared reference to `self`, while the loop body needs a mutable one.

JoJoDeveloping: Which makes me think, perhaps there's a better API than `drain_filter` :thinking:

RalfJung: Ah, makes sense, thanks.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -721,42 +727,58 @@ impl Tree {
...
+                stack.push((next_child, 0));
+                continue;
+            } else {
+                // We have processed all children of `node`, so now it is time to process `node` itself.
+                // First, get the current children of `node`. To appease the borrow checker,
+                // we have to temporarily move the list out of the node, and then put the
+                // list of remaining children back in.
+                let mut children_of_node =
+                    mem::take(&mut self.nodes.get_mut(*tag).unwrap().children);
+                // Remove all useless children, and save them for later (again to appease the borrow checker)
```
RalfJung, line 757:
```suggestion
                // Remove all useless children, and save them for later.
                // The closure needs `&self` and the loop below needs `&mut self`, so we need to `collect`
                // in to a temporary list.
```

RalfJung:
Just one last comment nit, then please squash the commits.

PR #3834, opened 2024-08-22
Fix a misleading comment in `tests/pass/tree_borrows/tree-borrows.rs`

sun-jacobi:
The original comment is somewhat misleading.

Since we don't add a protector for `x` here, `f` should be allowed to deallocate `x`.


RalfJung:
Good catch, looks like a copy-paste error. Thanks!

@bors r+

PR #3835, opened 2024-08-22
Avoid extra copy by using `retain_mut` and moving the deletion into the closure

JoJoDeveloping:
Fixes the FIXME introduced in #3833. Thanks to @dmitrii-ubskii for the idea :slightly_smiling_face:


RalfJung:
Very nice. :) Thanks!

@bors r+

PR #3836, opened 2024-08-23
epoll: Add a EINVAL case

tiif:
In ``epoll_ctl`` documentation, it is mentioned that: 
> EINVAL epfd is not an epoll file descriptor, or fd is the same as epfd, or the requested operation op is not supported by this interface.

So I added this EINVAL case for ``epfd == fd`` in ``epoll_ctl``


oli-obk (changes requested):
Add a test for this code path

PR #3837, opened 2024-08-23
Make Tree Borrows Provenance GC compact the tree

JoJoDeveloping:
Follow-up on #3833 and #3835. In these PRs, the TB GC was fixed to no longer cause a stack overflow. One test that motivated it was the test `fill::horizontal_line` in [`tiny-skia`](https://github.com/RazrFalcon/tiny-skia). But not causing stack overflows was not a large improvents, since it did not fix the fundamental issue: The tree was too large. The test now ran, but it required gigabytes of memory and hours of time (only for it to be OOM-killed 🤬), whereas it finishes within 24 seconds in Stacked Borrows. With this merged, it finishes in about 40 seconds under TB.

The problem in that test was that it used [`slice::chunked`](https://doc.rust-lang.org/std/primitive.slice.html#method.chunks) to iterate a slice in chunks. That iterator is written to reborrow at each call to `next`, which creates a linear tree with a bunch of intermediary nodes, which also fragments the `RangeMap` for that allocation.

The solution is to now compact the tree, so that these interior nodes are removed. Care is taken to not remove nodes that are protected, or that otherwise restrict their children.

I am currently only 99% sure that this is sound, and I do also think that this could compact even more. So @Vanille-N please also have a look at whether I got the compacting logic right.

For a more visual comparison, [here is a gist](https://gist.github.com/JoJoDeveloping/ae4a7f7c29335a4c233ef42d2f267b01) of what the tree looks like at one point during that test, with and without compacting.

This new GC requires a different iteration order during accesses (since the current one can make the error messages non-deterministic), so it is rebased on top of #3843 and requires that PR to be merged first.


src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -267,6 +267,19 @@ impl Permission {
...
+    /// But we need to be careful: If the parent is Frozen, and the child is Reserved,
+    /// we can not do such a merge. In general, such a merge is possible if the parent
+    /// is not protected, and if the child is reachable from the parent. This
+    /// corresponds to the `partial_cmp` order on permissions.
+    pub fn can_be_replaced_by_child(self, child: Self) -> bool {
+        match self.inner.partial_cmp(&child.inner) {
+            Some(Ordering::Less) | Some(Ordering::Equal) => true,
+            Some(Ordering::Greater) | None => false,
+        }
+    }
```
JoJoDeveloping, line 282: This is the magic function. I think that this is sound, i.e. if it returns `true`, then the child can indeed replace the parent.

Here is an example: The parent is `Active`, and the child is `Frozen`. On a foreign write, both go to `Disabled`. On a foreign read, both go to `Frozen`. On a child read, both remain, and the `Frozen` tag blocks child writes. So the `Active` has no effect.

JoJoDeveloping, line 282: I also don't think this function is complete. For once, if the parent is `ReservedIM`, and the child is `Reserved`, we can still replace the parent by the child, whereas `partial_cmp` says that they are incomparable.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -714,6 +714,60 @@ impl Tree {
...
+            || global.borrow().protected_tags.contains_key(&node.tag)
+        {
+            return None;
+        }
+        let child_idx = node.children[0];
+        let child = self.nodes.get(child_idx).unwrap();
+        if global.borrow().protected_tags.contains_key(&child.tag) {
+            // todo think really hard whether we can't just handle this as we would regularly
+            return None;
+        }
```
JoJoDeveloping, line 736: I am not sure if this check is necessary. I reckon that it is not, i.e. we can merge protected tags up. Part of the intuition is:
* If the child is initialized at an offset, so was the parent at that offset.
* Almost all the protector does is make more transitions UB. Since we do not remove the protector, these same transitions remain UB
* The one exception is that `Reserved` becomes conflicted on foreign accesses -- but this should not cause any issues, since the foreign read still happens similarly on the child.

src/borrow_tracker/tree_borrows/perms.rs
JoJoDeveloping: Note that the handling of `Reserved {conflicted: true}` is *very* subtle:
* Usually, it's not OK to replace a conflicted `Reserved` by an `Active`. But a `Reserved` never has `Active` children anyways (as all `Active` have `Active` parents), and further, since we know the parent is not protected, we know that the conflictedness does not matter anyway
* This exposes another way the method is incomplete: a conflicted `Reserved` parent can be replaced by a non-conflicted `Reserved` child, since the parent's conflictedness no longer matters, as the parent is not protected. But this is not how `partial_cmp` works.

(Note that I am highlighting all these cases to show that the reasoning here is rather subtle in many cases)

JoJoDeveloping:
We should probably have a definition of `replaceable(parent, child)` and then exhaustively test that:

Let `p_parent` and `p_child` be two permissions such that `replaceable(p_parent, p_child), and where `p_parent` is not protected. Then all accesses to them either
* move them further along such that afterwards `replaceable(p_parent', p_child')
* cause UB. In that case, the UB must also be caused if only `p_child` is accessed in that manner.

@Vanille-N can you add such a test. Note that `replaceable` is [this method](https://github.com/JoJoDeveloping/miri/blob/14e81cf2e11b204b3c23ceeecf8724342fb23f2a/src/borrow_tracker/tree_borrows/perms.rs#L277).

JoJoDeveloping:
The test failures are caused by the GC changes, in combination with setting `-Zmiri-gc-interval=1` to "stress-test" the GC. Since the compacting operation takes some time, this causes timing-assertive tests to fail. Additionally, it causes UI tests to fail since some intermediary nodes disappear, and it no longer prints information about those.

In general, how this new GC feature interacts with diagnostics is not yet decided.

JoJoDeveloping:
The tests are failing for several reasons, all related to the fact that it's a "GC-stress" mode test where the GC runs _every_ tick.

First, `linux-futex.rs` fails because (presumably) the GC now does a bunch more work, so things get delayed, and the time taken for an operation is outside of the acceptible range.

The other tests because the GC messes with the tree structure, combined with a suboptimal tree visiting order when checking accesses. To check an access, one currently starts at the root, and first checks that all nodes from there downwards support a child access. If the violating node is a parent, the error message will point this out (and print information about both tags). But quite often, this parent is simply the immediate parent of the node that was accessed, and not used otherwise, to the GC has merged it with the child. This then means that there is no violating parent, and thus the error message changes.

My solution would be to change the iteration order to go upwards but not downwards, which would be resilient to GC tree compactings. The comments currently in code seem to imply that the current iteration order has a deeper meaning, but sadly without going any deeper. @Vanille-N do you know more?

src/borrow_tracker/tree_borrows/perms.rs
Vanille-N: `partial_cmp` on permissions denotes reachability, and is not by definition tied to the amount of UB that the permission can produce. I like the idea of `can_be_replaced_by_child` but I think the above implementation suggests that the replaceability derives from `partial_cmp` when it's actually very dependent on invariants of the tree. I think it should be decoupled from `partial_cmp` and hand-written to explicitly show all compactable cases with their justification.

Vanille-N: For example the fact that you can compact an `Active` parent with a `Frozen` child is entirely dependent on whether the `Active` is protected, so there are at minimum many unwritten assumptions in `can_be_replaced_by_child` about the fact that protected tags can't be GC'd and the valid parent-child combinations.

Vanille-N:
> My solution would be to change the iteration order to go upwards but not downwards, which would be resilient to GC tree compactings. The comments currently in code seem to imply that the current iteration order has a deeper meaning, but sadly without going any deeper. @Vanille-N do you know more?

Sure. The current iteration order is of course a heuristic (TB doesn't specifically require an order to these) in an attempt to quickly identify the "root cause" of the issue, with a priority for more straightforward errors.

Insufficient permissions are more straightforward than protector violations, so we check parents before children.
The hidden assumption here is that parents can only trigger UB of the kind "insufficient permissions", and children can only trigger UB of the kind "protector violation".

As for the iteration order (top-down), the reasoning is that if we have `x` parent of `y` parent of `z` all `Disabled` and we do a child access through `z`, we are going to get the same kind of UB whichever we choose to report, but if we report `z` and the user fixes just `z` then we still get UB on `y` on the next run. At the time of writing this code I guessed that `z` was more likely to be the root cause of the bug and get the user to fix the UB in fewer steps. I don't have concrete evidence this is mostly guesswork.

I recognize that the combination of this iteration order and this tree compacting strategy can lead to nondeterministic error messages. Fixing that should take absolute priority over whatever diagnostics heuristics is implemented, and any iteration order will probably still give good enough messages.

Vanille-N:
Finally, from what I understand (I haven't looked at the detail of your compaction logic yet), you seem to be keeping the child and deleting the parent. This is not the only compacting logic possible.

The one I had in mind originally was to do the opposite : look at the parent to determine if we can delete the child. This removes the constraint of "if the parent is a singleton" because it preserves the ancestry relation even in branching trees.
For example it seems to me that your approach is not capable of compacting
```
0: Frozen
   |-- 1: Frozen (GC)
   |      '--  2: Frozen
   '-- 3: Frozen (GC)
          '-- 4: Frozen
```
into
```
0: Frozen
   |-- 2: Frozen
   '-- 4: Frozen
```

I don't doubt that it can merge some configurations that a top-down approach cannot, but the point is that the iteration order that I chose is compatible with the compacting logic that I initially had in mind.

Vanille-N:
Also since wide trees cannot get very deep, it's probably fine to have a compacting logic that can't handle branching if we observe that issues come mostly from concrete examples where the tree is deep.

JoJoDeveloping:
> The one I had in mind originally was to do the opposite : look at the parent to determine if we can delete the child. This removes the constraint of "if the parent is a singleton" because it preserves the ancestry relation even in branching trees.

Compared with that approach, doing it like I do here has several advantages:
* It is in-place: Since we only ever move a child up, we don't need to extend the array of children
* It is faster: Since we only check single-child nodes, there will never be a case where we check a long array only to realize we can't compact at the last node.
* It preserves the branching structure: In other words, not flattening nodes with multiple children can be seen as a feature.

Also note that the number of additional nodes that your approach would remove, but mine would not, is bounded: My approach produces trees at most twice as large as yours. The reason is that in a binary tree, the number of interior nodes is at most the number of leaves. So the efficiency gains offered by an even more collapsing GC are constant.

src/borrow_tracker/tree_borrows/perms.rs
JoJoDeveloping: Yes, we do not remove protected parents. Such parents should never be GC'd anyways since they still have a protector end write coming.

But I do think there's a deeper connection between reachability and replaceability: Permissions that come "later" always have "less" permissions than those that come earlier. It should always be sound to replace a node by one that's further ahead in the state machine (protectors notwithstanding). But I agree that it should probably be decoupled in code.

JoJoDeveloping:
> Also since wide trees cannot get very deep, it's probably fine to have a compacting logic that can't handle branching if we observe that issues come mostly from concrete examples where the tree is deep.

Note that further optimizations are possible elsewhere to handle wide trees. For instance, those implemented [in this commit](https://github.com/JoJoDeveloping/miri/commit/52e298cf096a91edfc80bdd9c02f7420b881f43d), which makes reads and writes skip more subtrees. But I did not see them cause a huge speedup, and they also affect diagnostics in rare cases, so I did not PR them yet.

RalfJung:
> We should probably have a definition of `replaceable(parent, child)` and then exhaustively test that:
> 
> Let `p_parent` and `p_child` be two permissions such that `replaceable(p_parent, p_child)`, and where `p_parent` is not protected. Then all accesses to them either
> 
>     * move them further along such that afterwards `replaceable(p_parent', p_child')` (they remain in simulation)
> 
>     * cause UB. In that case, the UB must also be caused if only `p_child` is accessed in that manner.
> 
> 
> @Vanille-N can you add such a test? Note that `replaceable` is [this method](https://github.com/JoJoDeveloping/miri/blob/14e81cf2e11b204b3c23ceeecf8724342fb23f2a/src/borrow_tracker/tree_borrows/perms.rs#L277).

Is this the right check? "Remain in simulation" is very weak. For instance, juts making `replaceable` always return `true` seems to satisfy this check, but is obviously not a valid implementation for `replaceable`.

JoJoDeveloping:
> Is this the right check? "Remain in simulation" is very weak. For instance, juts making `replaceable` always return `true` seems to satisfy this check, but is obviously not a valid implementation for `replaceable`.

Such an implementation would fail the second condition (e.g. if the parent is `Frozen` and the child is `Active`, child writes will be UB only at the parent). And I think that the relation we are looking for here is precisely "the largest simulation where UB at the child implies UB at the parent."

RalfJung:
> Such an implementation would fail the second condition (e.g. if the parent is Frozen and to be replaced by a child being Active, child writes will be UB only at the parent). 

Your comment says they have to satisfy one of the conditions, and the first one is trivially satisfied. So it's okay to violate the second condition, the entire soundness criterion you are suggesting is still satisfied as far as I can tell.

JoJoDeveloping:
> Your comment says they have to satisfy one of the conditions, and the first one is trivially satisfied. So it's okay to violate the second condition, the entire soundness criterion you are suggesting is still satisfied as far as I can tell.

I mean you either move along or cause UB, you can't have both at the same time, so that's why it's "either."

RalfJung:
Ah, "move along" implies "without UB". :)

JoJoDeveloping:
I squashed all the comments here. The reason is that I made some changes to the garbage collector in the first commit here, that I undid in one of the later commits, but when rebasing to the master branch where the GC also changed, git got confused, and I could not be bothered to correctly replay my changes if I know that a later commit would undo them.

bench-cargo-miri/slice-chunked/src/main.rs
```diff
@@ -0,0 +1,26 @@
+//! This is a small example using slice::chunks, which creates a very large Tree Borrows tree.
+//! Thanks to ##3837, the GC now compacts the tree, so this test can be run in a reasonable time again.
+//! The actual code is adapted from tiny_skia, see https://github.com/RazrFalcon/tiny-skia/blob/master/src/pixmap.rs#L121
+//! To make this benchmark demonstrate the effectiveness, run with MIRIFLAGS="-Zmiri-tree-borrows -Zmiri-provenance-gc=100"
```
RalfJung, line 4: How long does this benchmark take to run with default flags (i.e., Stacked Borrows)?

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -267,6 +267,43 @@ impl Permission {
...
+            // Reserved (as parent, where conflictedness does not matter)
+            // can be replaced by all but ReservedIM,
+            // since ReservedIM alone would survive foreign writes
+            (ReservedFrz { .. }, ReservedIM) => false,
+            (ReservedFrz { .. }, _) => true,
+            // Active can not be replaced by something surviving
+            // foreign reads and then remaining writable
+            (Active, ReservedIM) => false,
+            (Active, ReservedFrz { .. }) => false,
+            (Active, Active) => true,
```
RalfJung, line 294:
```suggestion
            // Replacing a state by itself is always okay.
            (Active, Active) => true,
```

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -267,6 +267,43 @@ impl Permission {
...
+            // Active can be replaced by Frozen, since it is not protected
+            (Active, Frozen) => true,
+            (Active, Disabled) => true,
+            // Frozen can only be replaced by Disabled
```
RalfJung, line 298:
```suggestion
            // Frozen can only be replaced by Disabled (and itself).
```

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -267,6 +267,43 @@ impl Permission {
...
+            // can be replaced by all but ReservedIM,
+            // since ReservedIM alone would survive foreign writes
+            (ReservedFrz { .. }, ReservedIM) => false,
+            (ReservedFrz { .. }, _) => true,
+            // Active can not be replaced by something surviving
+            // foreign reads and then remaining writable
+            (Active, ReservedIM) => false,
+            (Active, ReservedFrz { .. }) => false,
+            (Active, Active) => true,
+            // Active can be replaced by Frozen, since it is not protected
```
RalfJung, line 295:
```suggestion
            // Active can be replaced by Frozen, since it is not protected.
```

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -267,6 +267,43 @@ impl Permission {
...
+            (Active, Frozen) => true,
+            (Active, Disabled) => true,
+            // Frozen can only be replaced by Disabled
+            (Frozen, Frozen) => true,
+            (Frozen, Disabled) => true,
+            (Frozen, _) => false,
+            // Disabled can not be replaced by anything else
```
RalfJung, line 302:
```suggestion
            // Disabled can not be replaced by anything else.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -128,6 +128,22 @@ impl LocationState {
         Ok(transition)
     }
 
+    /// Like `perform_access`, but ignores the diagnostics, and also is pure.
```
RalfJung, line 131:
```suggestion
    /// Like `perform_access`, but ignores the concrete error cause and also uses state-passing
    /// rather than a mutable reference.
```

bench-cargo-miri/slice-chunked/src/main.rs
JoJoDeveloping: About 450 millis on my machine. For TB, if you tune the GC properly (i.e. set it to run every 100 BBs), you can get to about 530 millis. With the normal GC setting, it takes about 3 seconds, without the GC about 50.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -840,6 +856,57 @@ impl Tree {
...
+    /// Checks whether a node can be replaced by its only child.
+    /// If so, returns the index of said only child.
+    /// If not, returns none.
+    fn can_be_replaced_by_single_child(
+        &self,
+        idx: UniIndex,
+        live: &FxHashSet<BorTag>,
+    ) -> Option<UniIndex> {
+        let node = self.nodes.get(idx).unwrap();
+        if node.children.len() != 1 || live.contains(&node.tag) || node.parent.is_none() {
```
RalfJung, line 868:
```suggestion
        // We never want to replace the root node, as it is also kept in `root_ptr_tags`.
        if node.children.len() != 1 || live.contains(&node.tag) || node.parent.is_none() {
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -840,6 +856,57 @@ impl Tree {
...
+            return None;
+        }
+        // Since protected nodes are never GC'd (see `borrow_tracker::GlobalStateInner::visit_provenance`),
+        // we know that `node` is not protected because otherwise `live` would
+        // have contained `node.tag`.
```
RalfJung, line 873: Can you add a debug assertion checking that the node is not protected? Or do we not have access to that state?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -840,6 +856,57 @@ impl Tree {
...
+        let child_idx = node.children[0];
```
RalfJung, line 874:
```suggestion
        // Check that for that one child, `can_be_replaced_by_child` holds for the permission
        // on all locations.
        let child_idx = node.children[0];
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -840,6 +856,57 @@ impl Tree {
...
+
+        Some(child_idx)
+    }
+
+    /// Properly removes a node.
+    /// The node to be removed should not otherwise be usable. It also
+    /// should have no children, but this is not checked, so that nodes
+    /// whose children were rotated somewhere else can be deleted without
+    /// having to first modify them to clear that array.
+    /// otherwise (i.e. the GC should have marked it as removable).
```
RalfJung, line 896: The last line doesn't sound like a sentence...?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -883,23 +950,20 @@ impl Tree {
...
-                        // (which are `self.nodes` and every range of `self.rperms`)
-                        // before we can safely apply `UniKeyMap::remove` to truly remove
-                        // this tag from the `tag_mapping`.
-                        let node = self.nodes.remove(*idx).unwrap();
-                        for (_perms_range, perms) in self.rperms.iter_mut_all() {
-                            perms.remove(*idx);
-                        }
-                        self.tag_mapping.remove(&node.tag);
-                        // now delete it
+                        // delete it everywhere else
```
RalfJung, line 953:
```suggestion
                        // Delete `idx` node everywhere else.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -883,23 +950,20 @@ impl Tree {
...
+                        self.remove_useless_node(*idx);
+                        // and delete it from children_of_node
                         false
                     } else {
-                        // do nothing, but retain
+                        if let Some(nextchild) = self.can_be_replaced_by_single_child(*idx, live) {
+                            // delete the in-between child
```
RalfJung, line 959:
```suggestion
                            // `nextchild` is our grandchild, and will become our direct child.
                            // Delete the in-between node, `idx`.
```

RalfJung (commented):
I still have to check the tests

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: That requires passing extra state around, unfortunately.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -128,6 +128,22 @@ impl LocationState {
         Ok(transition)
     }
 
+    /// Like `perform_access`, but ignores the diagnostics, and also is pure.
+    /// As such, it returns `Some(x)` if the transition succeeded, or `None`
+    /// if there was an error.
+    #[allow(unused)]
```
RalfJung, line 134:
```suggestion
    #[cfg(test)]
```

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: Okay, in that case please add a comment in `FrameExtra::visit_provenance` saying that visiting the protected nodes here is load-bearing for the TB GC. Also this comment here should reference `FrameExtra::visit_provenance`, not `GlobalStateInner`.

RalfJung:
Looks good, thanks. :)

@bors r+

PR #3843, opened 2024-08-25
Make TB tree traversal bottom-up

JoJoDeveloping:
In preparation for #3837, the tree traversal needs to be made bottom-up, because the current top-down tree traversal, coupled with that PR's changes to the garbage collector, can introduce non-deterministic error messages if the GC removes a parent tag of the accessed tag that would have triggered the error first.

This is a breaking change for the diagnostics emitted by TB. The implemented semantics stay the same.

### DIFF
--- a/src/borrow_tracker/tree_borrows/tree.rs
+++ b/src/borrow_tracker/tree_borrows/tree.rs
@@ -302,7 +302,20 @@ enum ContinueTraversal {
     SkipSelfAndChildren,
 }
 
+#[derive(Clone, Copy)]
+pub enum ChildrenVisitMode {
+    VisitChildrenOfAccessed,
+    SkipChildrenOfAccessed,
+}
+
+enum RecursionState {
+    BeforeChildren,
+    AfterChildren,
+}
+
 /// Stack of nodes left to explore in a tree traversal.
+/// See the docs of `traverse_this_parents_children_other` for details on the
+/// traversal order.
 struct TreeVisitorStack<NodeContinue, NodeApp, ErrHandler> {
     /// Identifier of the original access.
     initial: UniIndex,
@@ -316,10 +329,12 @@ struct TreeVisitorStack<NodeContinue, NodeApp, ErrHandler> {
     /// Mutable state of the visit: the tags left to handle.
     /// Every tag pushed should eventually be handled,
     /// and the precise order is relevant for diagnostics.
-    /// Since the traversal is bottom-up, we need to remember
-    /// whether we're here initially (false) or after visiting all
-    /// children (true). The bool indicates this.
-    stack: Vec<(UniIndex, AccessRelatedness, bool)>,
+    /// Since the traversal is piecewise bottom-up, we need to
+    /// remember whether we're here initially, or after visiting all children.
+    /// The last element indicates this.
+    /// This is just an artifact of how you hand-roll recursion,
+    /// it does not have a deeper meaning otherwise.
+    stack: Vec<(UniIndex, AccessRelatedness, RecursionState)>,
 }
 
 impl<NodeContinue, NodeApp, InnErr, OutErr, ErrHandler>
@@ -362,64 +377,85 @@ where
         &mut self,
         this: &mut TreeVisitor<'_>,
         accessed_node: UniIndex,
-        push_children_of_accessed: bool,
+        visit_children: ChildrenVisitMode,
     ) -> Result<(), OutErr> {
+        // We want to visit the accessed node's children first.
+        // However, we will below walk up our parents and push their children (our cousins)
+        // onto the stack. To ensure correct iteration order, this method thus finishes
+        // by reversing the stack. This only works if the stack is empty initially.
         assert!(self.stack.is_empty());
         // First, handle accessed node. A bunch of things need to
         // be handled differently here compared to the further parents
         // of `accesssed_node`.
         {
             self.propagate_at(this, accessed_node, AccessRelatedness::This)?;
-            if push_children_of_accessed {
+            if matches!(visit_children, ChildrenVisitMode::VisitChildrenOfAccessed) {
                 let accessed_node = this.nodes.get(accessed_node).unwrap();
+                // We `rev()` here because we reverse the entire stack later.
                 for &child in accessed_node.children.iter().rev() {
-                    self.stack.push((child, AccessRelatedness::AncestorAccess, false));
+                    self.stack.push((
+                        child,
+                        AccessRelatedness::AncestorAccess,
+                        RecursionState::BeforeChildren,
+                    ));
                 }
             }
         }
-        // Then, handle the accessed node's parent. Here, we need to
+        // Then, handle the accessed node's parents. Here, we need to
         // make sure we only mark the "cousin" subtrees for later visitation,
         // not the subtree that contains the accessed node.
         let mut last_node = accessed_node;
         while let Some(current) = this.nodes.get(last_node).unwrap().parent {
             self.propagate_at(this, current, AccessRelatedness::StrictChildAccess)?;
             let node = this.nodes.get(current).unwrap();
+            // We `rev()` here because we reverse the entire stack later.
             for &child in node.children.iter().rev() {
                 if last_node == child {
                     continue;
                 }
-                self.stack.push((child, AccessRelatedness::DistantAccess, false));
+                self.stack.push((
+                    child,
+                    AccessRelatedness::DistantAccess,
+                    RecursionState::BeforeChildren,
+                ));
             }
             last_node = current;
         }
+        // Reverse the stack, as discussed above.
         self.stack.reverse();
         Ok(())
     }
 
     fn finish_foreign_accesses(&mut self, this: &mut TreeVisitor<'_>) -> Result<(), OutErr> {
-        while let Some((idx, rel_pos, is_final)) = self.stack.last_mut() {
+        while let Some((idx, rel_pos, step)) = self.stack.last_mut() {
             let idx = *idx;
             let rel_pos = *rel_pos;
-            if *is_final {
-                self.stack.pop();
-                self.propagate_at(this, idx, rel_pos)?;
-            } else {
-                *is_final = true;
-                let handle_children = self.should_continue_at(this, idx, rel_pos);
-                match handle_children {
-                    ContinueTraversal::Recurse => {
-                        // add all children, and also leave the node itself
-                        // on the stack so that it can be visited later.
-                        let node = this.nodes.get(idx).unwrap();
-                        for &child in node.children.iter() {
-                            self.stack.push((child, rel_pos, false));
+            match *step {
+                // How to do bottom-up traversal, 101: Before you handle a node, you handle all children.
+                // For this, you must first find the children, which is what this code here does.
+                RecursionState::BeforeChildren => {
+                    // Next time we come back will be when all the children are handled.
+                    *step = RecursionState::AfterChildren;
+                    // Now push the children, except if we are told to skip this subtree.
+                    let handle_children = self.should_continue_at(this, idx, rel_pos);
+                    match handle_children {
+                        ContinueTraversal::Recurse => {
+                            let node = this.nodes.get(idx).unwrap();
+                            for &child in node.children.iter() {
+                                self.stack.push((child, rel_pos, RecursionState::BeforeChildren));
+                            }
+                        }
+                        ContinueTraversal::SkipSelfAndChildren => {
+                            // skip self
+                            self.stack.pop();
+                            continue;
                         }
                     }
-                    ContinueTraversal::SkipSelfAndChildren => {
-                        // skip self
-                        self.stack.pop();
-                        continue;
-                    }
+                }
+                // All the children are handled, let's actually visit this node
+                RecursionState::AfterChildren => {
+                    self.stack.pop();
+                    self.propagate_at(this, idx, rel_pos)?;
                 }
             }
         }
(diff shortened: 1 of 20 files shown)


src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -282,166 +292,195 @@ struct TreeVisitor<'tree> {
...
     f_propagate: NodeApp,
     /// Handler to add the required context to diagnostics.
     err_builder: ErrHandler,
     /// Mutable state of the visit: the tags left to handle.
     /// Every tag pushed should eventually be handled,
     /// and the precise order is relevant for diagnostics.
-    stack: Vec<(UniIndex, AccessRelatedness)>,
+    /// Since the traversal is bottom-up, we need to remember
+    /// whether we're here initially (false) or after visiting all
+    /// children (true). The bool indicates this.
```
RalfJung, line 314: If it is bottom-up, the access would always be after visiting all children, right?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -282,166 +292,195 @@ struct TreeVisitor<'tree> {
...
+        assert!(self.stack.is_empty());
+        // First, handle accessed node. A bunch of things need to
+        // be handled differently here compared to the further parents
+        // of `accesssed_node`.
+        {
+            self.propagate_at(this, accessed_node, AccessRelatedness::This)?;
+            if push_children_of_accessed {
+                let accessed_node = this.nodes.get(accessed_node).unwrap();
+                for &child in accessed_node.children.iter().rev() {
+                    self.stack.push((child, AccessRelatedness::AncestorAccess, false));
```
RalfJung, line 369: Please either use an `enum` instead of just `false` here, or make this something like `/* after_children */ false` or so.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -282,166 +292,195 @@ struct TreeVisitor<'tree> {
...
+        Self { initial, f_continue, f_propagate, err_builder, stack: Vec::new() }
     }
 }
 
 impl<'tree> TreeVisitor<'tree> {
-    // Applies `f_propagate` to every vertex of the tree top-down in the following order: first
-    // all ancestors of `start`, then `start` itself, then children of `start`, then the rest.
+    // Applies `f_propagate` to every vertex of the tree bottom-up in the following order: first
+    // all ancestors of `start` (starting with `start` itself), then children of `start`, then the rest,
+    // always going bottom-up.
```
RalfJung, line 435: I am confused. It may be good to have a little image here or so explaining the order. It seems like we're jumping up and down the tree pretty wild here? First move up from `start`, then jump to the bottom and move towards `start`, and jump to other leaves and move towards ancestors of `start`?

Maybe call this "piecewise bottom-up" to make it clear that this is not a normal bottom-up traversal.

That's a very strange iteration order. I guess the key point is that for TB correctness, the order does not matter, so this only affects error messages? If so, then please say that explicitly.

tests/fail/both_borrows/alias_through_mutation.tree.stderr
```diff
@@ -5,24 +5,18 @@ LL |     let _val = *target_alias;
    |                ^^^^^^^^^^^^^ read access through <TAG> at ALLOC[0x0] is forbidden
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Tree Borrows rules it violated are still experimental
-   = help: the accessed tag <TAG> is a child of the conflicting tag <TAG>
-   = help: the conflicting tag <TAG> has state Disabled which forbids this child read access
-help: the accessed tag <TAG> was created here
+   = help: the accessed tag <TAG> has state Disabled which forbids this child read access
+help: the accessed tag <TAG> was created here, in the initial state Frozen
```
RalfJung, line 9: I like these changes! If the access is UB by looking juts at the accessed tag, we shouldn't be pointing at some parent instead.

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: Well, yes, but actually no. There is infrastructure for having the visitation skip entire subtrees if it would not change anything. This is done by remembering the last visit, and if the last visit was a foreign read, and the next visit is a foreign read, then it skips that.

(Note: I just realized that this may be unsound to do in combination with `lazy`, let me check.. But that's the motivation for visiting everything twice).

JoJoDeveloping: Indeed, skipping subtree updates the way we do currently is, eh, wrong and broken :grimacing:  See https://github.com/rust-lang/miri/issues/3846. We should find a solution to that before we continue here.

JoJoDeveloping: And independently of that infrastructure, you must first _get to_ the child to visit. For this you need to walk down.

RalfJung: This comment makes a lot more sense after reading the other comments further down that explain the actual iteration order. :)  I wouldn't quite call this "bottom-up".

RalfJung:
That's fancy. :-)

The best way to put this in ASCII is probably to label each node with a number, indicating its position in the traversal. Also we can use a smaller tree, it doesn't have to be quite so deep and wide.

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: I added an ASCII art iamge.

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: This is not resolved yet -- there's no enum and no comment.

I think an enum would be better.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -285,166 +299,224 @@ struct TreeVisitor<'tree> {
...
+        {
+            self.propagate_at(this, accessed_node, AccessRelatedness::This)?;
+            if matches!(visit_children, ChildrenVisitMode::VisitChildrenOfAccessed) {
+                let accessed_node = this.nodes.get(accessed_node).unwrap();
+                for &child in accessed_node.children.iter().rev() {
+                    self.stack.push((child, AccessRelatedness::AncestorAccess, false));
                 }
             }
         }
+        // Then, handle the accessed node's parent. Here, we need to
```
RalfJung, line 386:
```suggestion
        // Then, handle the accessed node's parents. Here, we need to
```

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: Ah whoops, I misidentified this line. I thought you meant the argument to that method, not the transient internal element that is pushed onto the encapsulated stack.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -285,166 +299,224 @@ struct TreeVisitor<'tree> {
...
-        while let Some((idx, rel_pos)) = self.stack.pop() {
-            self.exec_and_visit(visitor, idx, /* no children to exclude */ None, rel_pos)?;
+    fn finish_foreign_accesses(&mut self, this: &mut TreeVisitor<'_>) -> Result<(), OutErr> {
+        while let Some((idx, rel_pos, is_final)) = self.stack.last_mut() {
+            let idx = *idx;
+            let rel_pos = *rel_pos;
+            if *is_final {
+                self.stack.pop();
+                self.propagate_at(this, idx, rel_pos)?;
+            } else {
```
RalfJung, line 412: I think this is easier to follow if the order of the code matches the execution order. Also, please add comments.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -285,166 +299,224 @@ struct TreeVisitor<'tree> {
...
+            let node = this.nodes.get(current).unwrap();
+            for &child in node.children.iter().rev() {
+                if last_node == child {
+                    continue;
+                }
+                self.stack.push((child, AccessRelatedness::DistantAccess, false));
+            }
+            last_node = current;
+        }
+        self.stack.reverse();
```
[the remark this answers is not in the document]
JoJoDeveloping: yes, we're flipping it so that later, the children of the first node are popped first. This is also why we iterate the children `rev()`'ed here.

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: I really don't know what to call them, except for `Phase1` and `Phase2`, or maybe `BeforeTheRecursiveCallsIfThisWasWrittenAsARecursiveFunction` and `AfterTheRecursiveCallsIfThisWasWrittenAsARecursiveFunction`.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -323,9 +328,11 @@ struct TreeVisitorStack<NodeContinue, NodeApp, ErrHandler> {
     /// Every tag pushed should eventually be handled,
     /// and the precise order is relevant for diagnostics.
     /// Since the traversal is piecewise bottom-up, we need to
-    /// remember whether we're here initially (false) or after
-    /// visiting all children (true). The bool indicates this.
-    stack: Vec<(UniIndex, AccessRelatedness, bool)>,
+    /// remember whether we're here initially  or after visiting all children.
```
RalfJung, line 331:
```suggestion
    /// remember whether we're here initially or after visiting all children.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -370,6 +377,11 @@ where
         accessed_node: UniIndex,
         visit_children: ChildrenVisitMode,
     ) -> Result<(), OutErr> {
+        // Since we go bottom-up, we push the accessed node's children first,
+        // so that they are at the bottom of the stack. But later, in `finish_foreign_accesses`
+        // we want to start visiting these children of the accessed node first.
+        // So this method finishes by reversing the stack. This only works if
+        // the stack is empty initially.
```
RalfJung, line 384: Now I am even more confused... we add them first since we want to visit them first, but later we reverse so they aren't actually first then...?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -471,6 +500,10 @@ impl<'tree> TreeVisitor<'tree> {
     // visiting the subtree at that node. If it e.g. returns `SkipSelfAndChildren` on node 6
     // above, then nodes 5 _and_ 6 would not be visited by `f_propagate`. It is not used for
     // notes having a child access (nodes 1, 2, 3).
+    //
+    // Finally, remember that the iteration order is not relevant for UB, it only affects
+    // diagnostics. It also affects tree traversal optimizations build on top of this, so
```
RalfJung, line 505:
```suggestion
    // diagnostics. It also affects tree traversal optimizations built on top of this, so
```

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: It's a stack. Usually, the things you add first are visited last. Since we visit the accessed node first, it's children would usually go last. This is why we `reverse()`.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -285,166 +299,257 @@ struct TreeVisitor<'tree> {
...
+        &mut self,
+        this: &mut TreeVisitor<'_>,
+        accessed_node: UniIndex,
+        visit_children: ChildrenVisitMode,
+    ) -> Result<(), OutErr> {
+        // Since we go bottom-up, we push the accessed node's children first,
+        // so that they are at the bottom of the stack. But later, in `finish_foreign_accesses`
+        // we want to start visiting these children of the accessed node first.
+        // So this method finishes by reversing the stack. This only works if
+        // the stack is empty initially.
```
RalfJung, line 384:
```suggestion
        // We want to visit the accessed node's children first.
        // However, we will below walk up our parents and push their children (our cousins)
        // onto the stack. To ensure correct iteration order, this method thus finishes
        // by reversing the stack. This only works if the stack is empty initially.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -285,166 +299,257 @@ struct TreeVisitor<'tree> {
...
+        err_builder: ErrHandler,
+    ) -> Self {
+        Self { initial, f_continue, f_propagate, err_builder, stack: Vec::new() }
     }
 }
 
 impl<'tree> TreeVisitor<'tree> {
-    // Applies `f_propagate` to every vertex of the tree top-down in the following order: first
-    // all ancestors of `start`, then `start` itself, then children of `start`, then the rest.
+    // Applies `f_propagate` to every vertex of the tree in a piecewise bottom-up way: First, visit
```
RalfJung, line 475: Can you make this a doc comment, i.e. make all the `//` into `///`?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -285,166 +299,257 @@ struct TreeVisitor<'tree> {
...
+    VisitChildrenOfAccessed,
+    SkipChildrenOfAccessed,
+}
+
+enum RecursionState {
+    BeforeChildren,
+    AfterChildren,
 }
 
 /// Stack of nodes left to explore in a tree traversal.
```
RalfJung, line 316:
```suggestion
/// Stack of nodes left to explore in a tree traversal.
/// See the docs of `traverse_this_parents_children_other` for details on the
/// traversal order.
```

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: Yes, but I would like to point out it wasn't one before.

RalfJung: I saw that. :)

JoJoDeveloping:
Help how do I get rustdoc to like my ASCII art :laughing:

RalfJung:
Looks good! But please squash the history a little. :)

JoJoDeveloping:
> But please squash the history a little. :)

Squashed! Let's hope that CI is happy this time

PR #3847, opened 2024-08-26
Disable tree traversal optimization that is wrong due to lazy nodes.

### LINKED ISSUE #3846: TB: Missed UB in the "optimization" that allows skipping subtrees
Tree Borrows has an [optimization](https://github.com/rust-lang/miri/blob/master/src/borrow_tracker/tree_borrows/tree.rs#L153) where we skip subtree updates if we know that the subtree has not changed -- at least that was the idea. More specifically, every node records if its last access was a child access, a foreign read, or a foreign read/write (mixed, since a write followed by a read is idempotent). Then, if e.g. the last access was a foreign read, and the next access is also a foreign read, we should be able to skip the entire subtree, since it should already be idempotent.

But turns out, we can not always skip the tree. The reason is that lazy nodes exist, which are "silently" created across the entire allocation whenever you do a reborrow, while the reborrow only emits an event for the actually accessed range. Combined with protected `Reserved` being supposed to become conflicted on a foreign read (even when lazy), this leads to missed UB. The optimization for foreign writes can also be broken by a `ReservedIM` node, which would shield its children from foreign write accesses, which is even worse.

Here is an example:

```Rust
use std::ptr::addr_of_mut;

fn do_something(_: u8) {}

unsafe fn access_after_sub_1(x: &mut u8, orig_ptr: *mut u8) {
    // causes a second access, which should make the lazy part of `x` be `Reserved {conflicted: true}`
    do_something(*orig_ptr);
    // read from the conflicted pointer
    *(x as *mut u8).byte_sub(1) = 42;
}

pub fn main() {
    unsafe {
        let mut alloc = [0u8, 0u8];
        let orig_ptr = addr_of_mut!(alloc) as *mut u8;
        let foo = &mut *orig_ptr;
        // cause a foreign read access to foo
        do_something(alloc[0]);
        access_after_sub_1(&mut *(foo as *mut u8).byte_add(1), orig_ptr);
    }
}
```

Because the `foo` reference is internally marked as "already experienced a foreign read,"  the second foreign read in `do_something` skips the entire subtree, and thus does not mark that the lazy part of `x` as conflicted. Thus, the later read is possible, whereas it should not be. 

CC @Vanille-N 

I was not able to turn this into a LLVM misoptimization, but the generated LLVM code has UB, I believe (accesses a `noalias` ptr and another pointer pointing to the same memory location). In general, turning this into a miscompilation is hard since LLVM does not seem to be willing to reorder reads and writes that are the first access.

JoJoDeveloping:
See #3846 for more information.

For now, the optimization is disabled in a very "hotfix" way, while we think about potential fixes. Nonetheless, this fixes #3846

### DIFF
--- a/src/borrow_tracker/tree_borrows/tree.rs
+++ b/src/borrow_tracker/tree_borrows/tree.rs
@@ -150,7 +150,10 @@ impl LocationState {
     // the propagation can be skipped next time.
     // It is a performance loss not to call this function when a foreign access occurs.
     // It is unsound not to call this function when a child access occurs.
-    fn skip_if_known_noop(
+    // FIXME: This optimization is wrong, and is currently disabled (by ignoring the
+    // result returned here). Since we presumably want an optimization like this,
+    // we should add it back. See #3864 for more information.
+    fn update_last_foreign_access(
         &mut self,
         access_kind: AccessKind,
         rel_pos: AccessRelatedness,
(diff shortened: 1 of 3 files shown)


RalfJung:
Thanks! Please also add a test case.

This does fix the "misses-ub" bug; getting back optimizations in a sound way is a separate issue IMO.

RalfJung:
Great, thanks. :)

@bors r+

PR #3848, opened 2024-08-27
Add tokio io test

tiif:
After #3804 landed, these tests passed.

### DIFF
--- a/test_dependencies/Cargo.toml
+++ b/test_dependencies/Cargo.toml
@@ -20,7 +20,7 @@ tempfile = "3"
 page_size = "0.6"
 # Avoid pulling in all of tokio's dependencies.
 # However, without `net` and `signal`, tokio uses fewer relevant system APIs.
-tokio = { version = "1.24", features = ["macros", "rt-multi-thread", "time", "net", "fs", "sync", "signal"] }
+tokio = { version = "1.24", features = ["macros", "rt-multi-thread", "time", "net", "fs", "sync", "signal", "io-util"] }
 
 [target.'cfg(windows)'.dependencies]
 windows-sys = { version = "0.52", features = [ "Win32_Foundation", "Win32_System_Threading" ] }
--- a/tests/pass-dep/tokio/sleep.rs
+++ b/tests/pass-dep/tokio/sleep.rs
@@ -1,5 +1,4 @@
-//@compile-flags: -Zmiri-permissive-provenance -Zmiri-backtrace=full
-//@only-target-x86_64-unknown-linux: support for tokio only on linux and x86
+//@only-target-linux: We only support tokio on Linux
 
 use tokio::time::{sleep, Duration, Instant};
 
--- a/tests/pass-dep/tokio/tokio_mvp.rs
+++ b/tests/pass-dep/tokio/tokio_mvp.rs
@@ -1,6 +1,0 @@
-// Need to disable preemption to stay on the supported MVP codepath in mio.
-//@compile-flags: -Zmiri-permissive-provenance -Zmiri-preemption-rate=0
-//@only-target-x86_64-unknown-linux: support for tokio exists only on linux and x86
-
-#[tokio::main]
-async fn main() {}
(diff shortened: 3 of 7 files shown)


tests/pass-dep/tokio/mpsc.stdout
```diff
@@ -0,0 +1,2 @@
+GOT = sending from first handle
+GOT = sending from second handle
```
RalfJung, line 2: This is not deterministic, is it? A different interleaving of threads can result in a different output and thus make this test fail.

test_dependencies/Cargo.toml
```diff
@@ -20,7 +20,7 @@ tempfile = "3"
 page_size = "0.6"
 # Avoid pulling in all of tokio's dependencies.
 # However, without `net` and `signal`, tokio uses fewer relevant system APIs.
-tokio = { version = "1.24", features = ["macros", "rt-multi-thread", "time", "net", "fs", "sync", "signal"] }
+tokio = { version = "1.24", features = ["macros", "rt-multi-thread", "time", "net", "fs", "sync", "signal", "io-util"] }
```
RalfJung, line 23: Which part of the test needs `io-util`?

tiif: The tests in ``file-io.rs`` that use ``file.read`` or ``file.write`` need that.

RalfJung: The [docs](https://docs.rs/tokio/1.39.3/tokio/fs/struct.File.html) say that needs the `fs` feature but don't mention `io-utils`. Strange.

RalfJung: Ah, it's from the [`*Ext` traits](https://docs.rs/tokio/1.39.3/tokio/io/trait.AsyncReadExt.html).

tests/pass-dep/tokio/mpsc.stdout
tiif: Right, I just change the output instead, so even if it has a different interleaving, the test still won't fail.

RalfJung:
Looks good :) please squash it into a single commit.

tests/pass-dep/tokio/mpsc.rs
```diff
@@ -0,0 +1,20 @@
+//@only-target-x86_64-unknown-linux: support for tokio only on linux and x86
```
RalfJung, line 1: This won't run the test on 32bit x86. Is that deliberate?

tiif: Good point, it should be allowed on all linux target.

tiif:
i forgot to put the ``//@only-target-linux`` flag for mpsc test, I wonder how it even pass the CI? It fails locally when I ran ``./miri run --dep tests/pass-dep/tokio/mpsc.rs --target i686-pc-windows-msvc``.

tiif:
Please don't r+ this first, I am not sure what's wrong with github, my local mpsc.rs has this, but github won't update it. 
```rust
//@only-target-linux: support for tokio only on linux
use std::sync::mpsc::channel;
```

tests/pass/mpsc.rs
```diff
@@ -1,3 +1,4 @@
+//@only-target-linux: support for tokio only on linux
```
RalfJung, line 1: This isn't a tokio test, please don't change it. :)

tiif: Yea I just realised I edited the wrong mpsc.rs, I renamed the original one so it is less confusing :')

tests/pass-dep/tokio/sleep.rs
```diff
@@ -1,5 +1,4 @@
-//@compile-flags: -Zmiri-permissive-provenance -Zmiri-backtrace=full
-//@only-target-x86_64-unknown-linux: support for tokio only on linux and x86
+//@only-target-linux: support for tokio only on linux
```
RalfJung, line 1:
```suggestion
//@only-target-linux: We only support tokio on Linux
```
same for the other files

tests/pass-dep/tokio/tokio_mvp.rs
tiif, line 1: Actually, do we still need this?

RalfJung:
Awesome. :)

@bors r+

PR #3852, opened 2024-08-29
Refactor fd read/write

tiif:
This PR passed the responsibility of reading to user supplied buffer and dest place to each implementation of ``FileDescription::read/write/pread/pwrite``. 

This is part of #3665.

### DIFF
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -25,10 +25,9 @@ pub(crate) enum FlockOp {
 pub trait FileDescription: std::fmt::Debug + Any {
     fn name(&self) -> &'static str;
 
-    /// Reads as much as possible into the given buffer, and returns the number of bytes read.
-    /// `ptr` is the pointer to the user supplied read buffer.
-    /// `len` indicates how many bytes the user requested.
-    /// `dest` is where the return value should be stored.
+    /// Reads as much as possible into the given buffer `ptr`.
+    /// `len` indicates how many bytes we should try to read.
+    /// `dest` is where the return value should be stored: number of bytes read, or `-1` in case of error.
     fn read<'tcx>(
         &self,
         _self_ref: &FileDescriptionRef,
@@ -41,10 +40,9 @@ pub trait FileDescription: std::fmt::Debug + Any {
         throw_unsup_format!("cannot read from {}", self.name());
     }
 
-    /// Writes as much as possible from the given buffer, and returns the number of bytes written.
-    /// `ptr` is the pointer to the user supplied read buffer.
-    /// `len` indicates how many bytes the user requested.
-    /// `dest` is where the return value should be stored.
+    /// Writes as much as possible from the given buffer `ptr`.
+    /// `len` indicates how many bytes we should try to write.
+    /// `dest` is where the return value should be stored: number of bytes written, or `-1` in case of error.
     fn write<'tcx>(
         &self,
         _self_ref: &FileDescriptionRef,
@@ -57,11 +55,9 @@ pub trait FileDescription: std::fmt::Debug + Any {
         throw_unsup_format!("cannot write to {}", self.name());
     }
 
-    /// Reads as much as possible into the given buffer from a given offset,
-    /// and returns the number of bytes read.
-    /// `ptr` is the pointer to the user supplied read buffer.
-    /// `len` indicates how many bytes the user requested.
-    /// `dest` is where the return value should be stored.
+    /// Reads as much as possible into the given buffer `ptr` from a given offset.
+    /// `len` indicates how many bytes we should try to read.
+    /// `dest` is where the return value should be stored: number of bytes read, or `-1` in case of error.
     fn pread<'tcx>(
         &self,
         _communicate_allowed: bool,
@@ -74,11 +70,10 @@ pub trait FileDescription: std::fmt::Debug + Any {
         throw_unsup_format!("cannot pread from {}", self.name());
     }
 
-    /// Writes as much as possible from the given buffer starting at a given offset,
-    /// and returns the number of bytes written.
+    /// Writes as much as possible from the given buffer `ptr` starting at a given offset.
     /// `ptr` is the pointer to the user supplied read buffer.
-    /// `len` indicates how many bytes the user requested.
-    /// `dest` is where the return value should be stored.
+    /// `len` indicates how many bytes we should try to write.
+    /// `dest` is where the return value should be stored: number of bytes written, or `-1` in case of error.
     fn pwrite<'tcx>(
         &self,
         _communicate_allowed: bool,
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -26,13 +26,16 @@ pub trait FileDescription: std::fmt::Debug + Any {
     fn name(&self) -> &'static str;
 
     /// Reads as much as possible into the given buffer, and returns the number of bytes read.
+    /// `ptr` is the pointer to user supplied read buffer.
     fn read<'tcx>(
         &self,
         _self_ref: &FileDescriptionRef,
         _communicate_allowed: bool,
-        _bytes: &mut [u8],
+        _ptr: Pointer,
+        _len: u64,
+        _dest: &MPlaceTy<'tcx>,
         _ecx: &mut MiriInterpCx<'tcx>,
-    ) -> InterpResult<'tcx, io::Result<usize>> {
+    ) -> InterpResult<'tcx> {
         throw_unsup_format!("cannot read from {}", self.name());
     }
 
@@ -42,8 +45,9 @@ pub trait FileDescription: std::fmt::Debug + Any {
         _self_ref: &FileDescriptionRef,
         _communicate_allowed: bool,
         _bytes: &[u8],
+        _dest: &MPlaceTy<'tcx>,
         _ecx: &mut MiriInterpCx<'tcx>,
-    ) -> InterpResult<'tcx, io::Result<usize>> {
+    ) -> InterpResult<'tcx> {
         throw_unsup_format!("cannot write to {}", self.name());
     }
 
@@ -52,10 +56,12 @@ pub trait FileDescription: std::fmt::Debug + Any {
     fn pread<'tcx>(
         &self,
         _communicate_allowed: bool,
-        _bytes: &mut [u8],
         _offset: u64,
+        _ptr: Pointer,
+        _len: u64,
+        _dest: &MPlaceTy<'tcx>,
         _ecx: &mut MiriInterpCx<'tcx>,
-    ) -> InterpResult<'tcx, io::Result<usize>> {
+    ) -> InterpResult<'tcx> {
         throw_unsup_format!("cannot pread from {}", self.name());
     }
 
@@ -66,8 +72,9 @@ pub trait FileDescription: std::fmt::Debug + Any {
         _communicate_allowed: bool,
         _bytes: &[u8],
         _offset: u64,
+        _dest: &MPlaceTy<'tcx>,
         _ecx: &mut MiriInterpCx<'tcx>,
-    ) -> InterpResult<'tcx, io::Result<usize>> {
+    ) -> InterpResult<'tcx> {
         throw_unsup_format!("cannot pwrite to {}", self.name());
     }
 
@@ -125,14 +132,18 @@ impl FileDescription for io::Stdin {
         &self,
         _self_ref: &FileDescriptionRef,
         communicate_allowed: bool,
-        bytes: &mut [u8],
-        _ecx: &mut MiriInterpCx<'tcx>,
-    ) -> InterpResult<'tcx, io::Result<usize>> {
+        ptr: Pointer,
+        len: u64,
+        dest: &MPlaceTy<'tcx>,
+        ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx> {
+        let mut bytes = vec![0; usize::try_from(len).unwrap()];
         if !communicate_allowed {
             // We want isolation mode to be deterministic, so we have to disallow all reads, even stdin.
             helpers::isolation_abort_error("`read` from stdin")?;
         }
-        Ok(Read::read(&mut { self }, bytes))
+        let result = Read::read(&mut { self }, &mut bytes);
+        ecx.return_read_bytes_and_count(ptr, bytes, result, dest)
     }
 
     fn is_tty(&self, communicate_allowed: bool) -> bool {
@@ -150,8 +161,9 @@ impl FileDescription for io::Stdout {
         _self_ref: &FileDescriptionRef,
         _communicate_allowed: bool,
         bytes: &[u8],
-        _ecx: &mut MiriInterpCx<'tcx>,
-    ) -> InterpResult<'tcx, io::Result<usize>> {
+        dest: &MPlaceTy<'tcx>,
+        ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx> {
         // We allow writing to stderr even with isolation enabled.
         let result = Write::write(&mut { self }, bytes);
         // Stdout is buffered, flush to make sure it appears on the
@@ -160,8 +172,7 @@ impl FileDescription for io::Stdout {
         // the host -- there is no good in adding extra buffering
         // here.
         io::stdout().flush().unwrap();
-
-        Ok(result)
+        ecx.return_written_byte_count_or_error(result, dest)
     }
 
     fn is_tty(&self, communicate_allowed: bool) -> bool {
@@ -179,11 +190,13 @@ impl FileDescription for io::Stderr {
         _self_ref: &FileDescriptionRef,
         _communicate_allowed: bool,
         bytes: &[u8],
-        _ecx: &mut MiriInterpCx<'tcx>,
-    ) -> InterpResult<'tcx, io::Result<usize>> {
+        dest: &MPlaceTy<'tcx>,
+        ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx> {
         // We allow writing to stderr even with isolation enabled.
         // No need to flush, stderr is not buffered.
-        Ok(Write::write(&mut { self }, bytes))
+        let result = Write::write(&mut { self }, bytes);
+        ecx.return_written_byte_count_or_error(result, dest)
     }
 
     fn is_tty(&self, communicate_allowed: bool) -> bool {
@@ -205,10 +218,12 @@ impl FileDescription for NullOutput {
         _self_ref: &FileDescriptionRef,
         _communicate_allowed: bool,
         bytes: &[u8],
-        _ecx: &mut MiriInterpCx<'tcx>,
-    ) -> InterpResult<'tcx, io::Result<usize>> {
+        dest: &MPlaceTy<'tcx>,
+        ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx> {
         // We just don't write anything, but report to the user that we did.
-        Ok(Ok(bytes.len()))
+        let result = Ok(bytes.len());
+        ecx.return_written_byte_count_or_error(result, dest)
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/shims/unix/fd.rs
```diff
@@ -647,3 +631,32 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         Ok(Scalar::from_target_isize(this.try_unwrap_io_result(result)?, this))
     }
 }
+// TODO: move this as a free function later
+fn write_byte_helper<'tcx>(
+    buf: Pointer,
+    bytes: Vec<u8>,
+    result: InterpResult<'tcx, io::Result<usize>>,
+    ecx: &mut MiriInterpCx<'tcx>,
+) -> InterpResult<'tcx, Scalar> {
```
oli-obk, line 630: Return i64 here

oli-obk (changes requested):

src/shims/unix/fd.rs
```diff
@@ -31,6 +31,7 @@ pub trait FileDescription: std::fmt::Debug + Any {
         _self_ref: &FileDescriptionRef,
         _communicate_allowed: bool,
         _bytes: &mut [u8],
+        _ptr: Pointer,
```
RalfJung, line 34: Please add a doc comment explaining what the pointer is for.

Why is this not `dest: &MPlaceTy<'tcx>`?

src/shims/unix/fd.rs
```diff
@@ -619,22 +613,63 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
         };
-
-        let result = result?.map(|c| i64::try_from(c).unwrap());
-        Ok(Scalar::from_target_isize(this.try_unwrap_io_result(result)?, this))
+        let res = write_byte_helper(offset, fd, bytes, communicate, this)?;
+        this.write_scalar(res, dest)?;
+        Ok(())
     }
 }
+fn read_byte_helper<'tcx>(
```
RalfJung, line 625: Please add a doc comment explaining what this does.

Also, would be good to find a better name -- "helper" says basically nothing.

src/shims/unix/fd.rs
```diff
@@ -556,43 +559,33 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                    bytes[..usize::try_from(read_bytes).unwrap()].iter().copied(),
-                )?;
-                Ok(Scalar::from_target_isize(read_bytes, this))
-            }
-            Err(e) => {
-                this.set_last_error_from_io_error(e)?;
-                Ok(Scalar::from_target_isize(-1, this))
-            }
-        }
+        let res = read_byte_helper(buf, bytes, result, this)?;
```
RalfJung, line 586: So, you are passing the `buf` pointer to `read`, but then you are *also* still writing to it afterwards here? That doesn't seem to make a lot of sense -- either it is the responsibility of `read` to write to the buffer, or not.

tiif:
Thanks for the feedback, I will slowly address them. So currently ``read_byte_helper`` is writing bytes to the user supplied buffer pointer, and write the total number of bytes written to the dest place. Right now, each ``FileDescription::read`` is responsible to writing to the buffer by calling  ``read_byte_helper``. A lot of error handling doesn't make sense yet, but feel free to suggest improvement while I am working on it. 

Currently, I still need to: 
1. let ``FileDescription::write`` to do the actual writing 
2. get ``FileDescription::pread`` to do the actual reading by calling ``read_byte_helper``
3. Simplify error handling, there are a few InterpResult that I want to get rid of

src/shims/unix/fd.rs
```diff
@@ -619,22 +627,68 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
         let Some(fd) = this.machine.fds.get(fd) else {
-            return Ok(Scalar::from_target_isize(this.fd_not_found()?, this));
+            let res = this.fd_not_found()?;
+            this.write_scalar(Scalar::from_target_isize(res, this), dest)?;
+            return Ok(());
         };
 
-        let result = match offset {
-            None => fd.write(&fd, communicate, &bytes, this),
+        let _result = match offset {
```
oli-obk, line 635: The reason you're getting a bogus message about undef bytes is that you're throwing away the unsupported error here. Thus the return integer is never written, so it's undef and immediately fails with an undef error when read for validation

tiif: Oh right, this makes sense, thanks!

src/shims/unix/fd.rs
```diff
@@ -633,7 +633,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         };
 
         let _result = match offset {
```
oli-obk:
```suggestion
        match offset {
```

src/shims/unix/fd.rs
tiif: The pointer is the buf pointer provided by the caller, it is used in ``read_byte_helper`` to write to user supplied buffer.

tiif: This might be able optimised further, but how ``read`` currently works is:
1. In ``FileDescription::read``, the bytes is written into ``&mut [u8]``
2. ``&mut [u8]`` is passed into ``read_byte_helper``
3. In ``read_byte_helper``, the data is copied from ``&mut [u8]`` to the user supplied buffer. 

I am not sure if there is a convenient way to move step 1 into ``read_byte_helper``, because read for stdin uses ``Read::read(&mut { self }, bytes)`` where eventfd uses 
```rust            
*bytes = match ecx.tcx.sess.target.endian {
                Endian::Little => counter.to_le_bytes(),
                Endian::Big => counter.to_be_bytes(),
};
```

RalfJung:
You changed evaluation order:
```rust
// Runs `bytes.to_vec()` after `Read::read.
let result = Read::read(&mut { self }, bytes);
ecx.read_byte_helper(ptr, bytes.to_vec(), result, dest)?;
```
```rust
// Runs `bytes.to_vec()` before `Read::read.
ecx.read_byte_helper(ptr, bytes.to_vec(), Read::read(&mut { self }, bytes), dest)?;
```
So maybe that explains the difference.

src/shims/unix/fd.rs
RalfJung: We should definitely not have both the `bytes` buffer and the `ptr` here, that's a too confusing API.

RalfJung: So replace `bytes` by `len: usize`, and in `eventd` you should deref the `buf: Pointer` as type `u32` -- that's a lot nicer than what we do right now, in fact.

tiif:
Is there any nice way to deref Pointer type to a place? So for eventfd, I perhaps can directly write to the user supplied buffer using ``write_int``: 

```rust
    fn read_byte_helper_ev(
        &mut self,
        buf_place: &MPlaceTy<'tcx>,
        read_val: u64,
        result: io::Result<usize>,
        dest: &MPlaceTy<'tcx>,
    ) -> InterpResult<'tcx> {
        // `File::read` never returns a value larger than `count`, so this cannot fail.
        match result.map(|c| i64::try_from(c).unwrap()) {
            // try to pass this the write_ptr inside write
            // Pass the pointer inside the write function.
            Ok(read_bytes) => {
                // Write to the user supplied buffer.
                this.write_int(read_val, buf_place)?;  // <-- Here!!
                // Write to the function return value place.
                this.write_int(read_bytes, dest)?;
                return Ok(());
            }
            Err(e) => {
                this.set_last_error_from_io_error(e)?;
                this.write_int(-1, dest)?;
                return Ok(());
            }
        }
    }

```

Unfortunately I can't just use ``deref_pointer_as`` for the ``Pointer`` type because it doesn't implement ``Readable``: 

```rust
    |
71  |         let buf_place = ecx.deref_pointer_as(&ptr, ecx.machine.layouts.u64);
    |                             ---------------- ^^^^ the trait `rustc_const_eval::interpret::Readable<'_, machine::Provenance>` is not implemented for `rustc_const_eval::interpret::Pointer<std::option::Option<machine::Provenance>>`
    |                             |
    |                             required by a bound introduced by this call
    |
    = help: the following other types implement trait `rustc_const_eval::interpret::Readable<'tcx, Prov>`:
              rustc_const_eval::interpret::ImmTy<'tcx, Prov>
              rustc_const_eval::interpret::MPlaceTy<'tcx, Prov>
              rustc_const_eval::interpret::OpTy<'tcx, Prov>
```

We might also need to keep it as a Pointer type (instead of using ``OpTy``) because we need to check pointer access in ``EvalContextExt::read``

```rust
        // Check that the *entire* buffer is actually valid memory.
        this.check_ptr_access(buf, Size::from_bytes(count), CheckInAllocMsg::MemoryAccessTest)?;
```

RalfJung:
Yes, `ptr_to_mplace` takes a `Pointer` and a `Layout` and makes it an `MPlace`.

Instead of passing it as a pointer, you could also try passing it as an `ImmTy`. That implements `Readable`. But I think passing as `Pointer` is probably the better choice.

src/shims/unix/linux/eventfd.rs
```diff
@@ -62,23 +62,27 @@ impl FileDescription for Event {
...
         _communicate_allowed: bool,
-        bytes: &mut [u8],
+        _bytes: &mut [u8],
+        len: usize,
         ptr: Pointer,
         dest: &MPlaceTy<'tcx>,
         ecx: &mut MiriInterpCx<'tcx>,
     ) -> InterpResult<'tcx> {
+        // eventfd read at the size of u64.
+        let buf_place = ecx.ptr_to_mplace(ptr, ecx.machine.layouts.u64);
```
tiif, line 72: This is a fail attempt, it compiles but the test threw UB error: 
```rust
error: Undefined Behavior: accessing memory based on pointer with alignment 2, but alignment 8 is required
  --> tests/pass-dep/libc/libc-eventfd.rs:13:29
   |
LL |     let res: i32 = unsafe { libc::read(fd, buf.as_mut_ptr().cast(), N).try_into().unwrap() };
   |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ accessing memory based on pointer with alignment 2, but alignment 8 is required
   |
   = note: BACKTRACE:
   = note: inside `read_bytes::<8>` at tests/pass-dep/libc/libc-eventfd.rs:13:29: 13:71
note: inside `test_read_write`
  --> tests/pass-dep/libc/libc-eventfd.rs:33:15
   |
LL |     let res = read_bytes(fd, &mut buf);
   |               ^^^^^^^^^^^^^^^^^^^^^^^^
note: inside `main`
  --> tests/pass-dep/libc/libc-eventfd.rs:8:5
   |
LL |     test_read_write();
   |     ^^^^^^^^^^^^^^^^^
```
Writing to a slice that at least has size 8 should be different from writing to u64.

tiif:
Another idea here:

These are all the file description that implements ``FileDescription::read``, and what they did is
1. fs.rs:
    - ``File::read`` takes in ``&mut [u8]`` and writes the bytes to it
2. fd.rs:
    - ``Read::read`` takes in ``&mut [u8]`` and writes the bytes to it
3. unnamed_socket.rs: 
    - ``VecDeque::read`` take in ``&mut [u8]`` writes to it
    
So perhaps I can turn the user supplied buffer into ``&mut [u8]`` and let all these different ``read`` apis directly write to it, but I am still not sure how to handle ``eventfd``. It is slightly different because currently what it is doing is turning ``u64`` to ``&mut [u8]`` to match the implementation of all the other file description above.

src/shims/unix/linux/eventfd.rs
oli-obk: Yea, alignment was never required here. Places can also be unaligned I think. Or we read unaligned from the place. I don't remember which it was but you'll need to somehow do an unaligned read

RalfJung: Ah, yes. We'll either need a `TyAndLayout` for a packed struct containing a `u64` field, or we need to add a `ptr_to_mplace_unaligned` on the rustc side.

RalfJung:
I think they should all just take a Pointer. We already have a helper function (I think) to write a bunch of bytes to a pointer, so it should not be hard for fs and socket to use that.

tiif:
This is partially blocked by https://github.com/rust-lang/rust/pull/130215 as I'd need the ``ptr_to_mplace_unaligned``to pass the tests.

src/shims/unix/fd.rs
```diff
@@ -618,22 +624,67 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    fn read_byte_helper(
+        &mut self,
+        buf: Pointer,
+        bytes: Vec<u8>,
+        result: io::Result<usize>,
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx> {
+        let this = self.eval_context_mut();
+        // `File::read` never returns a value larger than `count`, so this cannot fail.
+        match result.map(|c| i64::try_from(c).unwrap()) {
```
oli-obk, line 657:
```suggestion
        match result {
```

This conversion isn't necessary I think?

src/shims/unix/fd.rs
```diff
@@ -618,22 +624,67 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    return Ok(());
                 };
-                fd.pwrite(communicate, &bytes, offset, this)
+                fd.pwrite(communicate, &bytes, offset, dest, this)?
             }
         };
+        Ok(())
+    }
+
+    /// This function either writes to the user supplied buffer and to dest place, or return error.
```
oli-obk, line 647:
```suggestion
    /// This function either writes to the user supplied buffer and to dest place, or sets the last libc error and writes -1 to thr dest.
```

Also name the function something more descriptive than `helper`

src/shims/unix/fs.rs
```diff
@@ -97,7 +111,9 @@ impl FileDescription for FileHandle {
                 .expect("failed to restore file position, this shouldn't be possible");
             res
         };
-        Ok(f())
+        let result = f();
+        ecx.write_byte_helper(result, dest)?;
+        Ok(())
```
oli-obk, line 116:
```suggestion
```

Just remove the `?;` from the previous line

Applies in lots of other places in this PR, too

src/shims/unix/fd.rs
tiif: Yup, we can use ``usize`` directly.

RalfJung:
@bors r-
Since since affects Miri globally, I'm going to have some comments as well, sorry.

src/shims/unix/linux/eventfd.rs
```diff
@@ -150,7 +157,8 @@ impl FileDescription for Event {
         // types for current file description.
         ecx.check_and_update_readiness(self_ref)?;
 
-        Ok(Ok(U64_ARRAY_SIZE))
+        let result = Ok(U64_ARRAY_SIZE);
+        ecx.return_written_byte_count_or_error(result, dest)
```
RalfJung, line 161: Why is `return_written_byte_count_or_error` a helper function? Every single caller fixes either `Err` or `Ok` so this is funneling through a `Result` without a good reason, it just makes the code harder to follow.

If there is any helper we want here, it is a new very generic helper that sets the last errors and also writes `-1` to `dest`. But that's probably better done in a future PR, for now you can just copy those 2 lines here I would say, like we do everywhere else.

src/shims/unix/linux/eventfd.rs
```diff
@@ -62,36 +62,39 @@ impl FileDescription for Event {
...
-                Endian::Big => counter.to_be_bytes(),
-            };
+            let result = Ok(U64_ARRAY_SIZE);
             self.counter.set(0);
             // When any of the event happened, we check and update the status of all supported event
             // types for current file description.
             ecx.check_and_update_readiness(self_ref)?;
 
-            return Ok(Ok(U64_ARRAY_SIZE));
+            return_read_bytes_and_count_ev(&buf_place, Some(counter), result, dest, ecx)
```
RalfJung, line 97: Why is `return_read_bytes_and_count_ev` a helper function? Every single caller fixes either `Err` or `Ok` so this is funneling through a `Result` without a good reason, it just makes the code harder to follow.

If there is any helper we want here, it is a new very generic helper that sets the last errors and also writes `-1` to `dest`. But that's probably better done in a future PR, for now you can just copy those 2 lines here I would say, like we do everywhere else.

src/shims/unix/linux/eventfd.rs
```diff
@@ -112,11 +115,13 @@ impl FileDescription for Event {
...
-    ) -> InterpResult<'tcx, io::Result<usize>> {
+    ) -> InterpResult<'tcx> {
         // Check the size of slice, and return error only if the size of the slice < 8.
         let Some(bytes) = bytes.first_chunk::<U64_ARRAY_SIZE>() else {
-            return Ok(Err(Error::from(ErrorKind::InvalidInput)));
+            let result = Err(Error::from(ErrorKind::InvalidInput));
+            return ecx.return_written_byte_count_or_error(result, dest);
         };
         // Convert from bytes to int according to host endianness.
         let num = match ecx.tcx.sess.target.endian {
```
RalfJung, line 127: This is very strange... why do we now have completely different code paths for `read` and `write` here? That indicates an oddly asymmetric API: `read` gets a `Pointer`, but `write` not. Why that? Seems better to do this consistently. Here we'd really like to also just do `ptr_to_mplace_unaligned` and then read from that place.

src/shims/unix/linux/eventfd.rs
```diff
@@ -62,36 +62,39 @@ impl FileDescription for Event {
...
         _communicate_allowed: bool,
-        bytes: &mut [u8],
+        ptr: Pointer,
+        len: u64,
+        dest: &MPlaceTy<'tcx>,
         ecx: &mut MiriInterpCx<'tcx>,
-    ) -> InterpResult<'tcx, io::Result<usize>> {
+    ) -> InterpResult<'tcx> {
+        // eventfd read at the size of u64.
+        let buf_place = ecx.ptr_to_mplace_unaligned(ptr, ecx.machine.layouts.u64);
```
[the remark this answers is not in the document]
tiif: Unfortunately inside the size check, ``return return_read_bytes_and_count_ev(&buf_place, None, result, dest, ecx);`` is called, so we need ``buf_place``. I probably should separate these two code path in ``return_read_bytes_and_count`` into different function. 

```rust
        match result {
            Ok(read_bytes) => {
                // If reading to `bytes` did not fail, we write those bytes to the buffer.
                // Crucially, if fewer than `bytes.len()` bytes were read, only write
                // that much into the output buffer!
                this.write_bytes_ptr(buf, bytes[..read_bytes].iter().copied())?;
                // The actual read size is always less than what got originally requested so this cannot fail.
                this.write_int(u64::try_from(read_bytes).unwrap(), dest)?;
                return Ok(());
            }
            Err(e) => {
                this.set_last_error_from_io_error(e)?;
                this.write_int(-1, dest)?;
                return Ok(());
            }
        }

```

This is the same issue as https://github.com/rust-lang/miri/pull/3852#discussion_r1761521341.

tiif:
Instead of separating the helper with ``read`` and ``write``, it would be better to separate it according to this two code path:
```rust
        match result {
            Ok(read_bytes) => {
                // If reading to `bytes` did not fail, we write those bytes to the buffer.
                // Crucially, if fewer than `bytes.len()` bytes were read, only write
                // that much into the output buffer!
                this.write_bytes_ptr(buf, bytes[..read_bytes].iter().copied())?;
                // The actual read size is always less than what got originally requested so this cannot fail.
                this.write_int(u64::try_from(read_bytes).unwrap(), dest)?;
                return Ok(());
            }
            Err(e) => {
                this.set_last_error_from_io_error(e)?;
                this.write_int(-1, dest)?;
                return Ok(());
            }
        }
```
or use ``try_unwrap_io_result``. 

As this could be done in a new PR, 

@rustbot ready

RalfJung:
Thanks! I've done some further minor tweaks, mostly to remove a bunch of `to_owned()` calls that were completely unnecessary. Please think a bit about the involved lifetimes to understand whether getting full ownership is needed or not -- or at least try it without to_owned before adding a call to that function. :)

Also I re-did eventfd a bit, see the last commit. I didn't think that helper function was helpful so I removed it.

Further improvements are possible I think (specifically some sort of helper for the quite common case of "set errno and return -1"), but that can be a separate PR.

@bors r+

PR #3856, opened 2024-09-01
Enable native libraries on macOS

### LINKED ISSUE #3595: macOS: support calling native functions
We have some preliminary support for calling native functions, but it's Linux only. We probably need someone with macOS knowledge to take a look. Currently what happens when one just [enables the tests on macOS](https://github.com/rust-lang/miri/pull/3594/commits/11f12568394d1d7f0d6fa31d993c0773451ffd0b) is that it fails to build the test library we want to make native calls from as `--version-script` is not supported. We need that test library to only export a select few symbols, so we need to find some suitable replacement. `exported_symbols_list` seems to be a thing for that.

There's likely more issues lurking behind this one, or maybe not. It's an adventure. :)

jder:
Fixes #3595 by using `-fvisibility=hidden` and the visibility attribute supported by both gcc and clang rather than the previous gcc-only mechanism for symbol hiding. Also brings over cfg changes from #3594 which enable native-lib functionality on all unixes.

Thanks for taking a look, feedback very welcome!

cc @RalfJung


tests/native-lib/pass/ptr_read_access.rs
```diff
@@ -1,4 +1,5 @@
-//@only-target-linux
+//@ignore-target-windows
+//@ignore-target-wasm
```
jder, line 2: I really want something like `@only-target-unix` but that seems to be unsupported. I looked at the other tests and didn't see an obvious replacement, does this seem reasonable?

tests/native-lib/scalar_arguments.c
```diff
@@ -1,27 +1,35 @@
...
-int add_one_int(int x) {
+// See comments in build_native_lib()
+#define EXPORT __attribute__((visibility("default")))
+
+EXPORT int add_one_int(int x) {
   return 2 + x;
 }
 
-void printer() {
+EXPORT void printer(void) {
```
jder, line 10: `void printer()` was causing a warning in clang, so added `(void)` here and below.

tests/native-lib/pass/ptr_read_access.rs
oli-obk: To do this nicely we need https://github.com/oli-obk/ui_test/issues/162

tests/native-lib/scalar_arguments.c
RalfJung: Lol I entirely forgot that that's how C works...

tests/native-lib/pass/ptr_read_access.rs
RalfJung: I think even that would still be ugly. We'd need ui-test to know what a "unix" target is and recognize that as `only-unix` or so.

For now, "ignore-wasm" + "ignore-windows" is the way to go.

tests/native-lib/pass/scalar_arguments.rs
```diff
@@ -1,4 +1,5 @@
-//@only-target-linux
+//@ignore-target-windows
```
RalfJung, line 1:
```suggestion
// Only works on Unix targets
//@ignore-target-windows
```
same for the other file

src/machine.rs
```diff
@@ -700,9 +700,9 @@ impl<'tcx> MiriMachine<'tcx> {
                     lib_file_path.clone(),
                 )
             }),
-            #[cfg(not(target_os = "linux"))]
+            #[cfg(not(unix))]
             native_lib: config.native_lib.as_ref().map(|_| {
-                panic!("loading external .so files is only supported on Linux")
+                panic!("loading external .so/.dylib files is only supported on Unix")
```
RalfJung, line 705:
```suggestion
                panic!("calling functions from native libraries via FFI is only supported on Unix")
```

tests/native-lib/scalar_arguments.c
```diff
@@ -1,27 +1,35 @@
...
+EXPORT short add_int16(short x) {
   return x + 3;
 }
 
-long add_short_to_long(short x, long y) {
+EXPORT long add_short_to_long(short x, long y) {
   return x + y;
 }
+
+// To test that functions not marked with EXPORT are not called by miri
```
RalfJung, line 32:
```suggestion
// To test that functions not marked with EXPORT cannot be called by Miri.
```

tests/ui.rs
```diff
@@ -36,20 +36,21 @@ fn build_native_lib() -> PathBuf {
...
         .expect("Failed to create directory for shared object file");
-    let so_file_path = so_target_dir.join("native-lib.so");
+    // We use a platform-neutral file extension to avoid having to hard-code alternatives.
+    let native_lib_path = so_target_dir.join("native-lib.module");
     let cc_output = Command::new(cc)
         .args([
             "-shared",
+            // We hide all symbols by default and export just the ones we need
+            // This is to future-proof against a more complex shared object which eg defines its own malloc
+            // (but we wouldn't want miri to call that, as it would if it was exported)
```
RalfJung, line 46:
```suggestion
            // This is to future-proof against a more complex shared object which eg defines its own malloc
            // (but we wouldn't want miri to call that, as it would if it was exported).
```

tests/ui.rs
```diff
@@ -36,20 +36,21 @@ fn build_native_lib() -> PathBuf {
...
             "-o",
-            so_file_path.to_str().unwrap(),
+            native_lib_path.to_str().unwrap(),
             // FIXME: Automate gathering of all relevant C source files in the directory.
             "tests/native-lib/scalar_arguments.c",
             "tests/native-lib/ptr_read_access.c",
-            // Only add the functions specified in libcode.version to the shared object file.
-            // This is to avoid automatically adding `malloc`, etc.
-            // Source: https://anadoxin.org/blog/control-over-symbol-exports-in-gcc.html/
             "-fPIC",
```
RalfJung, line 53: Can you move this argument up, above the file list? Next to `-shared` probably makes sense.

RalfJung (commented):
This looks great, thanks. :-) I just have some nits.

@rustbot author

jder:
Thanks for the quick review, I made these changes.

jder:
Hrm, I expected that to fix the labels. Maybe:

@rustbot ready

tests/native-lib/fail/private_function.rs
```diff
@@ -0,0 +1,14 @@
+//@ignore-target-windows
```
RalfJung, line 1:
```suggestion
// Only works on Unix targets
//@ignore-target-windows
```
You missed that here, and in `tests/native-lib/fail/function_not_in_so.rs` as well.

RalfJung (commented):
Just some comments that were missed. Then please squash the commits :) (with `--keep-base` to keep the force push diff small)

RalfJung:
Great, thanks :-)

@bors r+

PR #3865, opened 2024-09-06
Fix comment in mutex_id_offset

Mandragorian:
We no longer store the kind inside the pthread_mutex_t, so this comment is outdated.
Sorry I didn't catch this in the original PR.


RalfJung:
Thanks!

@bors r+

PR #3866, opened 2024-09-07
Bump ui test to 0.26

oli-obk:
lots of issues fixed and various improvements

🤞 that it works this time around


tests/pass/shims/x86/intrinsics-x86-adx.rs
```diff
@@ -1,12 +1,7 @@
 // Ignore everything except x86 and x86_64
 // Any new targets that are added to CI should be ignored here.
 // (We cannot use `cfg`-based tricks here since the `target-feature` flags below only work on x86.)
-//@ignore-target-aarch64
-//@ignore-target-arm
-//@ignore-target-avr
-//@ignore-target-s390x
-//@ignore-target-thumbv7em
-//@ignore-target-wasm
+//@ignore-target: aarch64 arm avr s390x thumbv7em wasm
```
oli-obk, line 4: neat side effect of disjunctive `only` filters: disjunctive `ignore` filters

tests/ui.rs
```diff
@@ -287,49 +320,25 @@ fn main() -> Result<()> {
...
-            WithoutDependencies,
-            tmpdir.path(),
-        )?;
+        ui(Mode::Fail, "tests/native-lib/fail", &target, WithoutDependencies, tmpdir.path())?;
     }
 
     Ok(())
 }
 
 fn run_dep_mode(target: String, mut args: impl Iterator<Item = OsString>) -> Result<()> {
```
oli-obk, line 333: do we have tests for run_dep_mode?

tests/pass/shims/x86/intrinsics-x86-adx.rs
oli-obk: ah we should probably switch to `only` here: https://github.com/rust-lang/miri/pull/3492#discussion_r1575960234

RalfJung:
No AFAIK we have no CI test for "run --dep"... We should, though.

tests/ui.rs
```diff
@@ -315,8 +315,7 @@ fn main() -> Result<()> {
     Ok(())
 }
 
-fn run_dep_mode(target: String, mut args: impl Iterator<Item = OsString>) -> Result<()> {
-    let path = args.next().expect("./miri run-dep must be followed by a file name");
+fn run_dep_mode(target: String, args: impl Iterator<Item = OsString>) -> Result<()> {
```
oli-obk: `path` was `--edition` 🙃 we have started adding more flags.

This was also just a refactoring leftover, as we always added `path` right before `args` and never used `path` elsewhere

oli-obk:
uh... can someone bless the native lib tests? I don't have a linux machine

eduardosm:
There you go: [miri-bless.diff.zip](https://github.com/user-attachments/files/16922393/miri-bless.diff.zip)

RalfJung:
macOS should also work for the native lib tests.

oli-obk:
> There you go: [miri-bless.diff.zip](https://github.com/user-attachments/files/16922393/miri-bless.diff.zip)

Thanks!

> macOS should also work for the native lib tests.

I don't have a *unix* machine and don't want to figure out wsl 😆

RalfJung:
Ah, so I can ask you to test the bat file and other Windows things then when that comes up? ;)

oli-obk:
Yea, they work great modulo bat files having a prompt when cancelled with Ctrl + C

oli-obk:
Oh fun. Line ending failures I think

tests/pass/shims/x86/intrinsics-x86-adx.rs
RalfJung: For `ignore`, this is equivalent to the old form, right?

I wonder if it would be worth to make `ignore-target: x86 linux` mean "ignore targets that contain both of these substrings"... but it's probably too confusing and that seems rarely needed.

ci/ci.sh
```diff
@@ -76,6 +76,9 @@ function run_tests {
     time HYPERFINE="hyperfine -w0 -r1" ./miri bench $TARGET_FLAG
   fi
 
+  ## Test that `miri run` with dependencies works
+  ./miri run --dep tests/pass-dep/getrandom.rs
```
RalfJung, line 80:
```suggestion
  ./miri run $TARGET_FLAG --dep tests/pass-dep/getrandom.rs
```

tests/fail-dep/libc/env-set_var-data-race.rs
```diff
@@ -1,5 +1,6 @@
 //@compile-flags: -Zmiri-disable-isolation -Zmiri-preemption-rate=0
-//@ignore-target-windows: No libc env support on Windows
+// No libc env support on Windows
+//@ignore-target: windows
```
RalfJung, line 3: When grepping the repo it is quite nice to have the explanation for an `ignore` on the same line as the `ignore`... I wonder if there is any way that can be preserved? Like
```
//@ignore-target: windows # No libs env support on Windows
```
Not sure what looks better: `#` or `//` as separator here.

Cargo.toml
```diff
@@ -27,6 +27,7 @@ ctrlc = "3.2.5"
 chrono = { version = "0.4.38", default-features = false }
 chrono-tz = "0.9"
 directories = "5"
+crossbeam = "0.8.4"
```
RalfJung, line 30: Wait why do we depend directly on crossbeam now...?

tests/pass/shims/x86/intrinsics-x86-adx.rs
RalfJung: And yes this should probably be `only-target: x86 i686` now.

RalfJung:
> Oh fun. Line ending failures I think

This fails locally even when I pass `--bless`. In fact something seems quite wrong, since I am getting
```
```
when `--bless` is set.

tests/ui.rs
```diff
@@ -155,10 +189,9 @@ fn run_tests(
     // Handle command-line arguments.
     let args = ui_test::Args::test()?;
     let default_bless = env::var_os("RUSTC_BLESS").is_some_and(|v| v != "0");
```
RalfJung, line 191: This variable is now only used in an assertion. That can't be right?

Cargo.toml
oli-obk: Oh I think that was an intermediate thing before I fixed ui_test

RalfJung:
I can run the native tests with `--bless` and push that to this branch -- once `--bless` works again.

oli-obk:
I'll fix that, tho you can use `-- -- --bless` until then

oli-obk:
Yea, we need to escape `cargo test` and then escape the normal test binaries, and finally the ui_test harness eats that flag

RalfJung:
A single `--` should work then since `./miri test` already escapes `cargo test`.

RalfJung:
Oh wait that just needs the comments put in separate lines.

tests/fail/tree_borrows/alternate-read-write.rs
```diff
@@ -13,7 +13,8 @@ pub fn main() {
     // This time y gets Frozen...
     let _val = *x;
     // ... and the next Write attempt fails.
-    *y += 1; // Failure //~ ERROR: /write access through .* is forbidden/
+    //~v ERROR: /write access through .* is forbidden/
+    *y += 1; // Failure
```
RalfJung, line 17: Why did this have to change? What happens when one does `// Failure //~ERROR ...`?

tests/fail-dep/libc/env-set_var-data-race.rs
RalfJung: I made that a feature request, it doesn't have to block this PR: https://github.com/oli-obk/ui_test/issues/277. (Though undoing all the multi-lining here if/when that feature ever gets implemented will be quite painful...)

RalfJung:
I got the Linux-only tests to work, now there are new failures due to new tests having landed in the mean time. I'll stop pushing to your branch since those should work on a Windows host with `--target`. :)

tests/fail/tree_borrows/alternate-read-write.rs
oli-obk: We don't allow annotations in comments anymore and will error if they occur there.

It's annoying on the parser to allow this, as I can't just look for the first `//` and parse from there

tests/ui.rs
```diff
@@ -5,12 +5,23 @@ use std::sync::OnceLock;
...
+use ui_test::dependencies::DependencyBuilder;
+use ui_test::per_test_config::TestConfig;
+use ui_test::spanned::Spanned;
+use ui_test::{status_emitter, CommandBuilder, Config, Format, Match, OutputConflictHandling};
+
+#[derive(Copy, Clone, Debug)]
+enum Mode {
+    Pass,
+    Fail,
+    Yolo,
```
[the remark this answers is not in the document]
RalfJung: What I meant was to rename the `Yolo` variant, which is very non-descriptive.

For the type, `RunDepMode` doesn't make a lot of sense, does it? `TestMode` or so could work.

tests/fail-dep/libc/env-set_var-data-race.rs
oli-obk: I gave this some thought, and doing it in general for all ´//@` commands adds quite a bit of complexity to the parser. I guess I could special case it just for the only/ignore parser

tests/ui.rs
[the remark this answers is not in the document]
oli-obk: 😆 I was wondering why you wanted that name

miri-script/src/commands.rs
```diff
@@ -499,9 +499,12 @@ impl Command {
...
-        if let Some(target) = &target {
-            early_flags.push("--target".into());
-            early_flags.push(target.into());
+        // For `--dep` we need to set the env var instead of passing the flag.
+        if !dep {
+            if let Some(target) = &target {
+                early_flags.push("--target".into());
+                early_flags.push(target.into());
+            }
         }
```
oli-obk, line 508: hmm now it's not setting the target anymore, before it was setting it twice. not sure when I'll get around to debugging this, probably not this week

ci/ci.sh
```diff
@@ -76,6 +76,9 @@ function run_tests {
     time HYPERFINE="hyperfine -w0 -r1" ./miri bench $TARGET_FLAG
   fi
 
+  ## Test that `miri run` with dependencies works
+  ./miri run $TARGET_FLAG --dep tests/pass-dep/getrandom.rs
```
RalfJung, line 80: I'm going to move this in its own PR, it shouldn't be blocked on the ui_test update.

miri-script/src/commands.rs
```diff
@@ -503,9 +503,13 @@ impl Command {
         // More flags that we will pass before `flags`
         // (because `flags` may contain `--`).
         let mut early_flags = Vec::<OsString>::new();
-        if let Some(target) = &target {
-            early_flags.push("--target".into());
-            early_flags.push(target.into());
+
+        // The target is already passed via `MIRI_TEST_TARGET`
```
RalfJung, line 507:
```suggestion
        // In `dep` mode, the target is already passed via `MIRI_TEST_TARGET`.
```

tests/ui.rs
```diff
@@ -77,32 +89,52 @@ fn miri_config(target: &str, path: &str, mode: Mode, with_dependencies: bool) ->
...
         out_dir: PathBuf::from(std::env::var_os("CARGO_TARGET_DIR").unwrap()).join("miri_ui"),
-        edition: Some("2021".into()), // keep in sync with `./miri run`
         threads: std::env::var("MIRI_TEST_THREADS")
             .ok()
             .map(|threads| NonZero::new(threads.parse().unwrap()).unwrap()),
         ..Config::rustc(path)
     };
 
+    config.comment_defaults.base().exit_status = match mode {
+        Mode::Pass => Some(0_i32),
```
RalfJung, line 101:
```suggestion
        Mode::Pass => Some(0),
```
Or is this somehow required?

tests/ui.rs
```diff
@@ -77,32 +89,52 @@ fn miri_config(target: &str, path: &str, mode: Mode, with_dependencies: bool) ->
...
         let builder_args = ["miri", "run"]; // There is no `cargo miri build` so we just use `cargo miri run`.
-        config.dependency_builder.args = builder_args.into_iter().map(Into::into).collect();
-        config.dependencies_crate_manifest_path =
-            Some(Path::new("test_dependencies").join("Cargo.toml"));
+        program.args = builder_args.into_iter().map(Into::into).collect();
+        let crate_manifest_path = Path::new("test_dependencies").join("Cargo.toml");
         // Reset `RUSTFLAGS` to work around <https://github.com/rust-lang/rust/pull/119574#issuecomment-1876878344>.
-        config.dependency_builder.envs.push(("RUSTFLAGS".into(), None));
+        program.envs.push(("RUSTFLAGS".into(), None));
+        config.comment_defaults.base().set_custom(
```
RalfJung, line 134: Please move this up, directly below `let crate_manifest_path` -- currently the RUSTFLAGS thing just sits in the middle of the dependency building setup and that doesn't seem to make any sense?

tests/ui.rs
```diff
@@ -153,14 +183,13 @@ fn run_tests(
...
 
     // Handle command-line arguments.
-    let args = ui_test::Args::test()?;
-    let default_bless = env::var_os("RUSTC_BLESS").is_some_and(|v| v != "0");
-    config.with_args(&args, default_bless);
-    if let OutputConflictHandling::Error(msg) = &mut config.output_conflict_handling {
-        *msg = "./miri test --bless".into();
-    }
+    let mut args = ui_test::Args::test()?;
+    args.bless = env::var_os("RUSTC_BLESS").is_some_and(|v| v != "0");
```
RalfJung, line 187: Should we use `|=` so that if the ui_test parser sees a `--bless` we don't just ignore that?

tests/ui.rs
```diff
@@ -287,50 +316,29 @@ fn main() -> Result<()> {
...
-        &target,
-        "",
-        Mode::Yolo { rustfix: RustfixMode::Disabled },
-        /* with dependencies */ true,
-    );
-    config.program.args.clear(); // remove the `--error-format` that ui_test adds by default
-    let dep_args = config.build_dependencies()?;
+fn run_dep_mode(target: String, args: impl Iterator<Item = OsString>) -> Result<()> {
+    let mut config = miri_config(&target, "", Mode::RunDep, /* with dependencies */ true);
+    config.comment_defaults.base().custom.remove("edition");
```
RalfJung, line 331:
```suggestion
    config.comment_defaults.base().custom.remove("edition"); // `./miri` adds an `--edition` in `args`, so don't set it twice
```

tests/pass/shims/x86/intrinsics-sha.rs
```diff
@@ -1,12 +1,10 @@
...
-//@ignore-target-s390x
-//@ignore-target-thumbv7em
-//@ignore-target-wasm
+// We're testing x86 target specific features
+//@ignore-target: aarch64
+//@ignore-target: arm
+//@ignore-target: avr
+//@ignore-target: s390x
+//@ignore-target: thumbv7em
+//@ignore-target: wasm
```
RalfJung, line 7:
```suggestion
//@only-target: x86_64 i686
```
You updated the comment but then not the actual ui test declaration?

RalfJung (commented):
LGTM apart from some nits.

tests/pass/shims/x86/intrinsics-sha.rs
oli-obk: Must have lost this change somewhere. I know I did it

tests/ui.rs
oli-obk: I just tried to keep the diff minimal. I'll add a separate commit that refactors this

RalfJung: Ah that's much nicer, thanks :)

RalfJung:
Looks great, thanks! r=me when CI is happy. :)

PR #3867, opened 2024-09-07
Renamed variable and fixed comments referring to renamed FileDescriptor

### LINKED ISSUE #3808: Rename ``file_descriptor`` to ``file_description_ref``
After #3712  renamed previous ``FileDescriptor`` to ``FileDescriptionRef``, several variables still use the old name ``file_descriptor`` and should be updated to ``file_description_ref`` to avoid confusion. This is a simple cleanup task, but I am currently focused on other work, so I am just opening this issue for anyone who might be interested in helping.

DeSevilla:
Fixes #3808, does not change any functionality

### DIFF
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -21,7 +21,7 @@ pub(crate) enum FlockOp {
     Unlock,
 }
 
-/// Represents an open file descriptor.
+/// Represents an open file description.
 pub trait FileDescription: std::fmt::Debug + Any {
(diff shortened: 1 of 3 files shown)


RalfJung:
Thanks! I did a bunch more renaming to hopefully make things more consistent wrt file description vs file descriptor: the latter (of type `i32`) are now called `fd_num`. We should only use them in the outer layers of the shims anyway, internally in Miri everything should be file descriptions.

@bors r+

PR #3871, opened 2024-09-09
detect when pthread_rwlock_t is moved

Mandragorian:
For some implementations of pthreads, the address of pthread_rwlock_t (or its fields) is used to identify the lock. That means that if the contents of a pthread_rwlock_t are moved in memory, effectively a new lock object is created, which is completely independted from the original. Thus we want to detect when when such objects are moved and show an error.

see also #3749 for more context


RalfJung:
Looks good, thanks. :)
@bors r+

PR #2451, opened 2022-07-29
Fetch miri flags from cargo config

avrong:
Closes #2347

This PR introduces cargo config parsing, and in a current implementation uses `miri.flags` property if there is no `MIRIFLAGS` env variable present. Maybe the best way to do that is to have some kind of merge of the parameters in both configs, sort of like cargo configs do it by themselves depending on location.

I did not make other tests, nor changed the docs, would be happy to do that if it's better to have them :)
Didn't figure out yet how to make a independent test to check the handling of `.cargo/config.toml` flags.

PR #3875, opened 2024-09-10
Add a way to extract miri flags from --config, env and toml

### LINKED ISSUE #2347: Set MIRIFLAGS from .cargo/config.toml
You can pass flags to rustc with [`build.rustflags`](https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags), and to rustdoc with [`build.rustdocflags`](https://doc.rust-lang.org/cargo/reference/config.html#buildrustdocflags).

```toml
[build]
rustflags = ["-C", "target-cpu=native"]
```

Supporting MIRIFLAGS the same way would be nice. Most of the time I want to run _every_ miri invocation for a project with a particular flag like `-Zmiri-disable-isolation`, so just setting it globally for the workspace would be ideal.

```toml
[miri]
miriflags = ["-Zmiri-disable-isolation"]
```

Related issue for MIRIFLAGS as command line arguments: https://github.com/rust-lang/miri/issues/2051

badumbatish:
Fixes #2347. 

The main driver is get_miri_flags in `cargo-miri/src/util.rs`, which follow this stratergy:
```
    // Strategy: (1) check pseudo var CARGO_ENCODED_MIRIFLAGS first (this is only set after we check for --config
    // in the cargo_dash_dash in the if else)
    //
    // if CARGO_ENCODED_MIRIFLAGS doesn't exist, we check in --config (2)
    // if --config doesn't exist, we check offical env var MIRIFLAGS (3)
    //
    // if MIRIFLAGS is non-existent, we then check for toml (4)
```

This is called once in phase_cargo_miri, then again in phase_runner.

For --config, I only retains all String that contains "miri", and then join them via " " for env.  This is because flagsplit splits based on the space character. Please let me know if this is not the procedure.

In later stage this env is resolved into Vec<String> via flagsplit().

I'm not sure how to write the test to describe this precedence of (1) over (2). Advice would be helpful :)

### DIFF
--- a/cargo-miri/src/phases.rs
+++ b/cargo-miri/src/phases.rs
@@ -181,6 +181,7 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
     let target_dir = get_target_dir(&metadata);
     cmd.arg("--target-dir").arg(target_dir);
 
+    eprintln!("Getting miri flags in phase_cargo_miri");
     cmd.args(get_miriflags());
     // Store many-seeds argument.
     let mut many_seeds = None;
@@ -642,6 +643,7 @@ pub fn phase_runner(mut binary_args: impl Iterator<Item = String>, phase: Runner
             }
         }
         // Respect miriflags.
+        eprintln!("Get miri flags in phase_runner");
         cmd.args(get_miriflags());
         // Set the current seed.
         if let Some(seed) = seed {
--- a/cargo-miri/src/phases.rs
+++ b/cargo-miri/src/phases.rs
@@ -641,9 +641,7 @@ pub fn phase_runner(mut binary_args: impl Iterator<Item = String>, phase: Runner
             }
         }
         // Respect miriflags.
-        if let Some(flags) = get_miriflags() {
-            cmd.args(flags);
-        }
+        cmd.args(get_miriflags());
         // Set the current seed.
         if let Some(seed) = seed {
             eprintln!("Trying seed: {seed}");
--- a/cargo-miri/src/phases.rs
+++ b/cargo-miri/src/phases.rs
@@ -181,7 +181,7 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
     let target_dir = get_target_dir(&metadata);
     cmd.arg("--target-dir").arg(target_dir);
 
-    eprintln!("Getting miri flags in phase_cargo_miri");
+    // eprintln!("Getting miri flags in phase_cargo_miri");
     cmd.args(get_miriflags_cargo_mini());
     // Store many-seeds argument.
     let mut many_seeds = None;
@@ -643,7 +643,7 @@ pub fn phase_runner(mut binary_args: impl Iterator<Item = String>, phase: Runner
             }
         }
         // Respect miriflags.
-        eprintln!("Get miri flags in phase_runner");
+        // eprintln!("Get miri flags in phase_runner");
         cmd.args(get_miriflags_runner());
         // Set the current seed.
         if let Some(seed) = seed {
--- a/cargo-miri/src/phases.rs
+++ b/cargo-miri/src/phases.rs
@@ -182,7 +182,7 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
     cmd.arg("--target-dir").arg(target_dir);
 
     eprintln!("Getting miri flags in phase_cargo_miri");
-    cmd.args(get_miriflags());
+    cmd.args(get_miriflags_cargo_mini());
     // Store many-seeds argument.
     let mut many_seeds = None;
     // *After* we set all the flags that need setting, forward everything else. Make sure to skip
@@ -644,7 +644,7 @@ pub fn phase_runner(mut binary_args: impl Iterator<Item = String>, phase: Runner
         }
         // Respect miriflags.
         eprintln!("Get miri flags in phase_runner");
-        cmd.args(get_miriflags());
+        cmd.args(get_miriflags_runner());
         // Set the current seed.
         if let Some(seed) = seed {
             eprintln!("Trying seed: {seed}");
--- a/cargo-miri/src/phases.rs
+++ b/cargo-miri/src/phases.rs
@@ -640,25 +640,15 @@ pub fn phase_runner(mut binary_args: impl Iterator<Item = String>, phase: Runner
                 cmd.arg(arg);
             }
         }
-        // Respect `MIRIFLAGS`.
-        if let Ok(a) = env::var("MIRIFLAGS") {
-            let args = flagsplit(&a);
-            cmd.args(args);
+        // Respect miriflags.
+        if let Some(flags) = get_miriflags() {
+            cmd.args(flags);
         }
         // Set the current seed.
         if let Some(seed) = seed {
             eprintln!("Trying seed: {seed}");
             cmd.arg(format!("-Zmiri-seed={seed}"));
         }
-<<<<<<< HEAD
-=======
-    }
-
-    // Respect flags.
-    if let Some(flags) = get_miriflags() {
-        cmd.args(flags);
-    }
->>>>>>> 7e32c0f90 (Fetch miri flags from cargo config)
 
         // Then pass binary arguments.
         cmd.arg("--");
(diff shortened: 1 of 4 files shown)


badumbatish:
hmmm hahha ok maybe i should have retained "-Zmiri" instead of just miri. The latter makes file name and other things gets tangled in

badumbatish:
after debugging i found out that in phase_cargo_miri if i try to extract MIRIFLAGS (a.k.a choice (3)) from env then some of the program will fail like this
```
FAILED TEST: tests/pass-dep/num_cpus.rs
command: "Dependencies"

full stderr:

   0: [91mfailed to compile dependencies:
      command: env -u RUSTFLAGS "/home/jjasmine/Developer/github/miri/target/debug/cargo-miri" "miri" "run" "--target-dir" "/home/jjasmine/Developer/github/miri/target/miri_ui" "--manifest-path" "test_dependencies/Cargo.toml" "--target=x86_64-unknown-linux-gnu" "--locked" "--message-format=json"
      stderr:
      Getting miri flags in phase_cargo_miri
      Choice 3
       -Zmiri-provenance-gc=1
      error: unknown `-Z` flag specified: miri-provenance-gc

      For available unstable features, see https://doc.rust-lang.org/nightly/cargo/reference/unstable.html
      If you intended to use an unstable rustc feature, try setting `RUSTFLAGS="-Zmiri-provenance-gc"`


      stdout:[0m
```
 but if i seperate them into so that in phase_cargo_miri I only perform choice (1) and (2) and then in phase_runner the rest then the ui test passes fine.  I think I'm missing sth important?

badumbatish:
heheh now i know how to ask for review

RalfJung:
Yeah I've seen the PR. :) Thanks for creating it!

It's quite busy in my life currently though, so I can't promise when I will get around to take a look at this.

cargo-miri/src/phases.rs
```diff
@@ -181,6 +181,8 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
     let target_dir = get_target_dir(&metadata);
     cmd.arg("--target-dir").arg(target_dir);
 
+    // eprintln!("Getting miri flags in phase_cargo_miri");
```
tiif, line 184: This commented ``eprintln`` might need to be removed. 

But I am not the maintainer and this PR might be on a draft state, so this is just a reminder. :)

badumbatish: haha yes I'll remove it on the next iteration :) ty for the review

cargo-miri/src/util.rs
```diff
@@ -88,6 +88,77 @@ pub fn escape_for_toml(s: &str) -> String {
     format!("\"{s}\"")
 }
 
+pub fn flagsplit(flags: &str) -> Vec<String> {
+    // This code is taken from `RUSTFLAGS` handling in cargo.
+    // Taken from miri-script util.rs
```
RalfJung, line 93: That last part is not true, this function already existed here? You just moved it up. Why did you move it?

cargo-miri/src/util.rs
```diff
@@ -88,6 +88,77 @@ pub fn escape_for_toml(s: &str) -> String {
...
+    flags.split(' ').map(str::trim).filter(|s| !s.is_empty()).map(str::to_string).collect()
+}
+
+pub fn get_miriflags_cargo_mini() -> Vec<String> {
```
RalfJung, line 97: Please add a doc comment explaining what this does. (Seems like you have a bunch of that inside the function, but that's now how doc comments work in Rust.)

Why "mini"?

cargo-miri/src/util.rs
```diff
@@ -88,6 +88,77 @@ pub fn escape_for_toml(s: &str) -> String {
...
+    // Respect `MIRIFLAGS` and `miri.flags` setting in cargo config.
+    // If MIRIFLAGS is present, flags from cargo config are ignored.
+    // This matches cargo behavior for RUSTFLAGS.
```
RalfJung, line 100: Please add a link to where this behavior is documented for cargo.

cargo-miri/src/util.rs
```diff
@@ -88,6 +88,77 @@ pub fn escape_for_toml(s: &str) -> String {
...
+    //
+    // Strategy: (1) check pseudo var CARGO_ENCODED_MIRIFLAGS first (this is only set after we check for --config
+    // in the cargo_dash_dash in the if else)
```
RalfJung, line 103: Now I am confused, this doesn't match the previous paragraph.

cargo-miri/src/util.rs
```diff
@@ -88,6 +88,77 @@ pub fn escape_for_toml(s: &str) -> String {
...
+    //
+    // if CARGO_ENCODED_MIRIFLAGS doesn't exist, we check in --config (2)
+    // if --config doesn't exist, we check offical env var MIRIFLAGS (3)
+    //
+    // if MIRIFLAGS is non-existent, we then check for toml (4)
+    if let Ok(cargo_encoded_miri_flags) = env::var("CARGO_ENCODED_MIRIFLAGS") {
+        // (1)
+        // eprintln!("Choice 1");
+        flagsplit(cargo_encoded_miri_flags.as_str())
```
RalfJung, line 112: This doesn't make sense, "ENCODED" env vars shouldn't get `flagsplit`. They use a different format that doesn't rely on whitespace.

cargo-miri/src/util.rs
```diff
@@ -88,6 +88,77 @@ pub fn escape_for_toml(s: &str) -> String {
...
+    } else if cargo_extra_flags().iter().any(|s| s.contains(&"-Zmiri".to_string())) {
```
RalfJung, line 113: What are you trying to do here? `cargo_extra_flags` will never contain any `-Zmiri`. Please read the doc comment of the functions you call.

cargo-miri/src/util.rs
```diff
@@ -88,6 +88,77 @@ pub fn escape_for_toml(s: &str) -> String {
...
+            .collect::<Vec<String>>();
+        let miri_flags_string = miri_flags_vec.join(" ");
+        env::set_var("CARGO_ENCODED_MIRIFLAGS", miri_flags_string);
+        miri_flags_vec
+    } else {
+        Vec::default()
+    }
+}
+pub fn get_miriflags_runner() -> Vec<String> {
+    // TODO: I quite not understand what Carl Jung means by Oh and please add a link to https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags.
```
RalfJung, line 129: Who is Carl Jung?

cargo-miri/src/phases.rs
```diff
@@ -181,6 +181,8 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
...
+    cmd.args(get_miriflags_cargo_mini());
```
RalfJung, line 185: So you are taking the miri flags and passing them to cargo...? How does that make sense?

RalfJung (commented):
Thanks for the PR! However, I'm afraid I mostly don't understand what you are trying to do here...

We want to either compute the flags in `phase_runner`, or compute them once upfront in `phase_cargo_miri`. You currently have complicated logic in both places which definitely is not the right approach. Probably doing it once upfront is better, since computing the flags will be more expensive now.

So, in `phase_cargo_miri` we compute the flags (details TBD, should work exactly like RUSTFLAGS). We take the result and store it in CARGO_ENCODED_MIRIFLAGS, using the usual ENCODED format cargo uses (documented e.g. [here](https://doc.rust-lang.org/cargo/reference/environment-variables.html)). Then phase_runner simply has to read that env var and set the arguments.

cargo-miri/src/util.rs
badumbatish: I am very sorry, I meant to say Ralf Jung. Again, deeply sorry about this...

RalfJung: Ah, no worries :)

cargo-miri/src/util.rs
badumbatish: An overside on my part, I'll put it back before definition of exec(), must have pressed d or x when in visual mode

cargo-miri/src/util.rs
badumbatish: Hi Ralf, this is from PR #2451 https://github.com/rust-lang/miri/pull/2451/files#diff-922365a8fd004632c70c2426b2851c89ed2e13229b7e7693550b87f00b100badR552 for the function get_miriflags().

I'll remove this block of comment from this function.

The comment can be made clearer by specifying `flags from cargo config` to `flags from .cargo/config.toml` (configuration file instead of --config option) since we run this in `get_miriflags_runner`
```rust
    let mut cmd = cargo();
    cmd.args(["-Zunstable-options", "config", "get", "miri.flags", "--format=json-value"]);
```

The behavior is from https://doc.rust-lang.org/cargo/reference/config.html#command-line-overrides

> Cargo also accepts arbitrary configuration overrides through the --config command-line option. The argument should be in TOML syntax of KEY=VALUE:
`cargo --config net.git-fetch-with-cli=true fetch`
The --config option may be specified multiple times, in which case the values are merged in left-to-right order, using the same merging logic that is used when multiple configuration files apply. 
Configuration values specified this way take precedence over environment variables, which take precedence over configuration files.

We check the arguments from --config first, then environment variable, then check the configuration files via `cargo config get`.

I couldn't find the evidence for `This matches cargo behavior for RUSTFLAGS.` Should I remove that line?

cargo-miri/src/util.rs
badumbatish: A misplaced of comment from previous PR that I forgot to remove. Sorry about that

cargo-miri/src/util.rs
badumbatish: From https://doc.rust-lang.org/cargo/reference/environment-variables.html

> CARGO_ENCODED_RUSTFLAGS — A list of custom flags separated by 0x1f (ASCII Unit Separator) to pass to all compiler invocations that Cargo performs.

I'll change this soon. Keeping this unresolved. Thank you

badumbatish:
Apologies for the delay. I'm jampacked with classes, exams and projects right now. I'll get back to it soon. Really sorry for this

RalfJung:
Thanks for the update, and no worries. :)

RalfJung:
@badumbatish I'll close this for now due to inactivity. Feel free to reopen the PR if you want to get back to this. :)

Note: do *not* push to the branch while the PR is closed. Github will only let you reopen the PR if the branch is exactly in the current state (commit 914bda3d31fa1cc39ef809dff639e46fe1001ed8).

badumbatish:
Sounds good! Sorry for your inconvience ~~

PR #3881, opened 2024-09-13
./miri run: directly run binary instead of using 'cargo run'

RalfJung:
This avoids re-building miri without dev dependencies, so it makes `./miri run` a lot faster if `./miri test` was already executed before.


RalfJung:
Hm, it works everywhere except on Windows... but the file path seems right. The error code translates to `0xC0000135`. Apparently this can mean STATUS_DLL_NOT_FOUND? Damn... we set `-C link-args=-Wl,-rpath,` which successfully prevents similar errors on Unix. What can we do on Windows if we don't want to use `cargo run` to run this binary?

Cc @ChrisDenton

RalfJung:
Maybe using `rustup run` suffices...
@bors r+

RalfJung:
Nope, unfortunately this does not help. :(

RalfJung:
The only other idea I have is to keep using `cargo run` on Windows, and if a Windows dev is sufficiently annoyed by the waiting time this incurs then they can figure out how to fix this :shrug:

ChrisDenton:
A DLL's directory has to be in `PATH` for it to be found or else be in the same directory as the exe.

(Well there are also [other options](https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#standard-search-order-for-unpackaged-apps) but they're more complicated)

RalfJung:
So probably we have to add `toolchains/miri/lib/rustlib/x86_64-unknown-linux-gnu/lib` to the PATH so that it can find librustc_driver and libLLVM.

RalfJung:
Let's see...
@bors r+

PR #3884, opened 2024-09-13
detect when pthread_cond_t is moved

### LINKED ISSUE #3749: pthreads synchronization primitives: detect mutex/rwlock/... being moved to a different location

Mandragorian:
Closes #3749

### DIFF
--- a/src/shims/unix/sync.rs
+++ b/src/shims/unix/sync.rs
@@ -206,7 +206,7 @@ fn translate_kind<'tcx>(ecx: &MiriInterpCx<'tcx>, kind: i32) -> InterpResult<'tc
 // - id: u32
 
 #[derive(Debug)]
-/// Additional data that may be used by shim implementations.
+/// Additional data that we attach with each rwlock instance.
 pub struct AdditionalRwLockData {
     /// The address of the rwlock.
     pub address: u64,
@@ -286,6 +286,19 @@ fn condattr_get_clock_id<'tcx>(
     .to_i32()
 }
 
+fn translate_clock_id<'tcx>(ecx: &MiriInterpCx<'tcx>, raw_id: i32) -> InterpResult<'tcx, ClockId> {
+    // To ensure compatibility with PTHREAD_COND_INITIALIZER on all platforms,
+    // we can't just compare with CLOCK_REALTIME: on Solarish, PTHREAD_COND_INITIALIZER
+    // makes the clock 0 but CLOCK_REALTIME is 3.
+    Ok(if raw_id == ecx.eval_libc_i32("CLOCK_REALTIME") || raw_id == 0 {
+        ClockId::Realtime
+    } else if raw_id == ecx.eval_libc_i32("CLOCK_MONOTONIC") {
+        ClockId::Monotonic
+    } else {
+        throw_unsup_format!("unsupported clock id: {raw_id}");
+    })
+}
+
 fn condattr_set_clock_id<'tcx>(
     ecx: &mut MiriInterpCx<'tcx>,
     attr_ptr: &OpTy<'tcx>,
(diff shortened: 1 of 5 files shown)


src/shims/unix/sync.rs
```diff
@@ -206,7 +206,7 @@ fn translate_kind<'tcx>(ecx: &MiriInterpCx<'tcx>, kind: i32) -> InterpResult<'tc
 // - id: u32
 
 #[derive(Debug)]
-/// Additional data that may be used by shim implementations.
+/// Additional data that we attach with each rwlock instance.
```
Mandragorian, line 209: I also changed this documentation comment since `AdditionalRwLockData` is now something internal to the shim implementation, so the previous wording felt a bit off.

src/shims/unix/sync.rs
```diff
@@ -364,25 +364,32 @@ fn cond_clock_offset<'tcx>(ecx: &MiriInterpCx<'tcx>) -> u64 {
     offset
 }
 
+#[derive(Debug)]
+/// Additional data that we attach with each cond instance.
+struct AdditionalCondData {
+    /// The address of the cond.
+    address: u64,
```
RalfJung, line 371: Please also store the clock kind in here instead of in-memory, similar to what you did with the mutex kind.

src/shims/unix/sync.rs
```diff
@@ -355,20 +361,30 @@ fn cond_clock_offset<'tcx>(ecx: &MiriInterpCx<'tcx>) -> u64 {
...
+            matches!(id, ClockType::Realtime),
             "PTHREAD_COND_INITIALIZER is incompatible with our pthread_cond layout: clock is not CLOCK_REALTIME"
         );
     }
 
     offset
 }
 
+#[derive(Debug, Clone, Copy)]
+enum ClockType {
```
RalfJung, line 375: We call it the clock "ID" in most places, and the C type name for this is `clockid_t`. So `ClockId` seems like a better name for this type?

src/shims/unix/sync.rs
```diff
@@ -902,13 +912,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 return Ok(());
             }
         };
-        let timeout_clock = if is_cond_clock_realtime(this, clock_id) {
+        let timeout_clock = if matches!(clock_type, ClockType::Realtime) {
```
RalfJung, line 915: Please make this a `match clock_id { ... }`

src/shims/unix/sync.rs
```diff
@@ -933,7 +943,6 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
         // Destroying an uninit pthread_cond is UB, so check to make sure it's not uninit.
         cond_get_id(this, cond_op)?;
```
RalfJung, line 945: I realize that's pre-existing, but -- this function is already called just above, so this call can be removed. Please preserve the comment though.

RalfJung:
Looks good, thanks. :)

Can you squash this into a single commit?

Mandragorian:
Thanks for the blazingly fast reviews!

PR #3893, opened 2024-09-16
Automatically add/remove labels when github review (requests) are used

oli-obk:


triagebot.toml
```diff
@@ -17,3 +17,15 @@ contributing_url = "https://github.com/rust-lang/miri/blob/master/CONTRIBUTING.m
 
 [no-merges]
 exclude_titles = ["Rustup"]
+
+[review-submitted]
+# This label is added when a "request changes" review is submitted.
+reviewed_label = "S-waiting-on-author"
+# These labels are removed when a "request changes" review is submitted.
+review_labels = ["S-waiting-on-review"]
```
RalfJung, line 25: Does it matter who writes this review? Seems a bit odd to change the label if some random bystander leaves a review.

RalfJung (commented):
Ah wow, I had no idea triagebot can do this. :D

triagebot.toml
oli-obk: They cannot. Github only allows the request changes button if you have enough permissions on a repo

RalfJung: Ah, okay :)

PR #3895, opened 2024-09-17
Implement LLVM x86 gfni intrinsics

TDecking:


src/shims/x86/gfni.rs
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        match unprefixed_name {
+            // Used to implement the `_mm{, 256, 512}_gf2p8affine_epi64_epi8` functions.
+            // See `affine_transform` for details.
+            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=gf2p8affine_
+            "vgf2p8affineqb.128" | "vgf2p8affineqb.256" | "vgf2p8affineqb.512" => {
+                let [left, right, imm8] =
+                    this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                affine_transform(this, left, right, imm8, dest, false)?;
```
RalfJung, line 34: Please no unnamed `bool`s. Either make this `/* inverse */ false`, or add an enum.

src/shims/x86/gfni.rs
```diff
@@ -0,0 +1,215 @@
...
+
+                for i in 0..dest_len {
+                    let left =
+                        u32::from(this.read_scalar(&this.project_index(&left, i)?)?.to_u8()?);
+                    let right =
+                        u32::from(this.read_scalar(&this.project_index(&right, i)?)?.to_u8()?);
+                    let dest = this.project_index(&dest, i)?;
+
+                    // Step 1: Carryless multiplication of `left` and `right`.
+                    // Carryless multiplication is explained in the `pclmulqdq`function.
```
RalfJung, line 68:
```suggestion
                    // Carryless multiplication is explained in the `x86::pclmulqdq` function.
```

src/shims/x86/gfni.rs
```diff
@@ -0,0 +1,215 @@
...
+                    let mut result = 0u32;
+                    for i in 0..8 {
+                        if left & (1 << i) != 0 {
+                            result ^= right << i;
+                        }
+                    }
+
+                    // Step 2: Polynomial division modulo 2.
+                    // Polynomial division modulo 2 is defined inside the `sse42` module.
```
RalfJung, line 77: It is? Where?

src/shims/x86/gfni.rs
```diff
@@ -0,0 +1,215 @@
...
+///             print!("0x{:02x}, ", get_value(i));
+///         }
+///         if i & 15 == 15 {
+///             println!();
+///         }
+///     }
+/// }
+/// ```
+#[rustfmt::skip]
+static TABLE: [u8; 256] = [
```
RalfJung, line 198: I am not happy with just carrying a random bunch of blobs here. Does the official Intel spec for these instruction refer to this table somehow? Or does it describe these instructions algorithmically? If it is the latter, can we use the algorithmic version in Miri?

tests/pass/shims/x86/intrinsics-x86-gfni.rs
```diff
@@ -0,0 +1,508 @@
...
+
+    for i in 0..NUM_BYTES {
+        input[i] = (i % 256) as u8;
+        result[i] = if i == 0 { 0 } else { 1 };
+    }
+
+    (input, result)
+}
+
+const AES_S_BOX: [u8; NUM_BYTES] = [
```
RalfJung, line 476: Did you copy this from somewhere? If yes, where?

tests/pass/shims/x86/intrinsics-x86-gfni.rs
RalfJung, line 1: Did you write all these tests from scratch or are they based on something?

RalfJung (commented):
Thanks for the PR! Are these the (in)famous NSA instructions in the x86 instruction set? They surely look sufficiently obscure. ;)

Unfortunately, what these instructions do is complete gibberish to me. Like, I know the basics of finite ring/field arithmetic, but I stare at the test file and nothing makes any sense.^^ This requires already knowing what all these instructions do to make any sense of -- which I don't, and I won't have the time to learn these either. Is there any way to make this test file a bit more approachable to someone who knows Rust and rings "modulo N" but nothing else? By adding comments that map all the Intel idiosyncrasies to something that makes mathematical sense, or so?

tests/pass/shims/x86/intrinsics-x86-gfni.rs
TDecking: I copied these tests from the stdarch test suite. The only change is test data caching for a reduced runtime.

src/shims/x86/gfni.rs
TDecking: CRC definition.

src/shims/x86/gfni.rs
TDecking: Done

src/shims/x86/gfni.rs
TDecking: Done.

src/shims/x86/gfni.rs
TDecking: The intrinsics guide only specifies that the byte is inverted, even inside the pseudocode.
I did find a source which includes a definition of the inverse byte, and created the table based on that.

RalfJung: Please give the source.

but if possible I'd like to not have the table at all.

src/shims/x86/gfni.rs
RalfJung: If I Ctrl-F "CRC definition" in the `sse42` module, there is no result. So I don't know what you mean by that.

Please fix this comment to be a reference that unambiguously references the correct part of `sse42` in a stable way.

src/shims/x86/gfni.rs
TDecking: The link is mentioned in `gf2p8_mul`. Should I add a reference here aswell?

RalfJung: I see you made it compute the table in the static initializer now, great. :)

Yes the link should be next to the code that was written based on that source.

TDecking: Done.

tests/pass/shims/x86/intrinsics-x86-gfni.rs
TDecking: This was already provided by the test suite. This is the Rijndale S-Box used inside the AES encryption standard.

src/shims/x86/gfni.rs
TDecking: The intrinsic names are now mentioned explicitly.

src/shims/x86/gfni.rs
```diff
@@ -0,0 +1,198 @@
...
+    while i < 8 {
+        if left & (1 << i) != 0 {
+            result ^= right << i;
+        }
+        i = i.wrapping_add(1);
+    }
+
+    // Step 2: Polynomial division modulo 2.
+    // Polynomial division modulo 2 is defined inside the `x86::sse42` module
+    // inside the "_mm_crc32_u{8, 16, 32, 64}" intrinsics implementation.
```
RalfJung, line 188: This looks quite a bit different from that implementation though. I can only pattern-match here since I don't understand the code, but other than "it's a loop with XOR and shift", it seems unrelated to what `_mm_crc32_u*` does.

If this is the same procedure, can it be extracted into a helper function?

TDecking: Algorithmically, both the crc32 intrinsics and the finite field multiplication use polynomial division modulo 2 in their definition, so the loops do the same thing. Maybe "described" is a better word than "defined".

The difference between the implementations stems from the fact that the intrinsics guide presents this step differently between the crc32 and gfni intrinsics. In case of the the crc32 intrinsics, the step is reduced to an unspecified `MOD2` function call, while the gfni intrinsics provides a full loop.

A nice property of this implementation is the one-to-one correspondence between the implementation and the definition of the intrinsics guide: Comparing the intrinsics guide definition with this one will make the correctness aspect very obvious. While this step can be extracted into a separate function, that correspondence would be lost.

If it should be extracted, where should the function be placed?

RalfJung: > A nice property of this implementation is the one-to-one correspondence between the implementation and the definition of the intrinsics guide

Where can I find that definition?

TDecking: One of the entries in https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=gf2p8mul.
I can add the link to the implementation.

RalfJung: Yes, that would probably be a better link than `_mm_crc32_u*` then.

src/shims/x86/gfni.rs
```diff
@@ -0,0 +1,198 @@
...
+                for i in 0..dest_len {
+                    let left = this.read_scalar(&this.project_index(&left, i)?)?.to_u8()?;
+                    let right = this.read_scalar(&this.project_index(&right, i)?)?.to_u8()?;
+                    let dest = this.project_index(&dest, i)?;
+                    this.write_scalar(Scalar::from_u8(gf2p8_mul(left, right)), &dest)?;
+                }
+            }
+            _ => return Ok(EmulateItemResult::NotSupported),
+        }
+        Ok(EmulateItemResult::NeedsReturn)
```
RalfJung, line 69: This will need a rebase due to some recent changes in how `InterpResult` works. These `Ok` now need to be written as `interp_ok`.

src/shims/x86/gfni.rs
TDecking: Done.

RalfJung:
Looks good, as far as I can tell... I still don't understand most of this code, but worst-case we get some complains that the intrinsics are doing the wrong thing. ;)

Thanks for the PR!
@bors r+

PR #3897, opened 2024-09-17
Use `@only-target` in SSE and SSE2 tests too

eduardosm:
It looks cleaner and makes it consistent with other X86 tests.

The huge diffs are mostly indentation changes.


RalfJung:
This is great, thanks!

@bors r+

Issue #3618, opened 2024-05-19
Android: make std thread APIs work


RalfJung:
Basic thread spawning works on Android, but some things are still missing:
- [x] setting/getting the thread name. This needs `prctl` commands `PR_SET_NAME` and `PR_GET_NAME`.
- [ ] the std synchronization primitives, which are built on futexes. Since Android uses the Linux kernel, futex support can be shared with Linux; this just requires a bit of refactoring to avoid copy-pasting all [this](https://github.com/rust-lang/miri/blob/1ae97809bd30d347f9b976254f49e6000330d69c/src/shims/unix/linux/foreign_items.rs#L98-L150).

(Also, we're looking for a target maintainer for Android, so please let us know if you're up to that -- basically someone we'd ping when there are Android questions, and in case std starts using a new Android API we'd hope they would be able to provide a Miri implementation.)

YohDeadfall:
How should the code be refactored to share the `sync` module between `linux` and `android`? The last one is based on Linux, but has different user space which is compatible in some points. So, at least for me it means that for these places the shared part should be under `linux` module. Even some parts made for `android` can be interpreted for `linux` also like `prctl`.

How do you see the refactoring?

RalfJung:
The easiest thing to do seems to be to move the Linux `syscall` shim into a separate function (still in the `linux` folder), and call that from Android.

> Even some parts made for android can be interpreted for linux also like prctl.

That could do the same, have a function in the `linux` folder that is called from `linux/foreign_items.rs` and `android/foreign_items.rs`. The most tricky part here is how to test this without too much code duplication...

YohDeadfall:
Got you, thanks!

There's one thing which is unclear to me. How do you decide what Miri should cover and in which cases? Since the issue is about `prctl` let's take it as an example. That system call is also available for `linux`, but its `libc` has `prctl` in it, but according to the implementation it won't be supported. That makes it's impossible to use `prctl` with Miri if someone decided to use it directly for whatever reason.

RalfJung:
Generally we want to support what std needs. On top of that if things are widely used out there we'll consider adding support for them (if the API can be implemented and maintained with reasonable effort).

For `prctl`, given that we anyway have to carry an implementation for Android, we might as well expose it on Linux, too.

PR #3899, opened 2024-09-20
Android: Added support for prctl handling thread names

YohDeadfall:
Addresses the first part of #3618.


RalfJung:
Thanks for the PR!

There's currently a bit of a queue and I have fairly little time, so it may take a bit until I can review this. (Or maybe someone else will pick up the review.)

YohDeadfall:
Sure, and meanwhile I will fix issues reported by the CI.

A side question, I noticed that the repository doesn't use git hooks like the `rust` repo. Is it intentional?

saethlin:
> the repository doesn't use git hooks

Repos can't have git hooks. What exactly are you referring to?

YohDeadfall:
The main repo installs them on bootstrap: https://github.com/rust-lang/rust/blob/1a5a2240bc1b8cf0bcce7acb946c78d6493a4fd3/src/bootstrap/src/core/build_steps/setup.rs#L462.

saethlin:
The equivalent for this repo would be a `miri-script` feature. I've always found git hooks more annoying than time-saving, so it's not something I would spend my own time on. But if you really like them, you're welcome to propose something.

YohDeadfall:
A script feature would be also fine for me. The key point is to have something which can be used to run the same stuff as in the CI but locally. That thing was implemented in `bevy` and `pgrx` is considering that too.

RalfJung:
To have CI check that your PR works, please change this line

https://github.com/rust-lang/miri/blob/18a8ab77407f0d67911e115619d0669b1d3e39b9/ci/ci.sh#L157

to something like
```
...  run_tests_minimal $BASIC $UNIX threadname pthread --skip pthread_cond_timedwait
```

tests/pass-dep/libc/prctl.rs
```diff
@@ -0,0 +1,56 @@
+//@ignore-target: windows # No prctl on Windows
+//@ignore-target: apple # No prctl on macOS
```
RalfJung, line 2: `@only-target: linux android # prctl is Linux-specific` would probably be better

src/shims/unix/linux/foreign_items.rs
```diff
@@ -72,6 +72,60 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             }
 
             // Threading
+            "prctl" => {
```
RalfJung, line 75: Note that this here will *not* be used on Android -- Linux is just the "normal" Linux stack, from which Android has diverged too far.

Not sure if it's worth supporting this on regular Linux targets. For now, I'd say just move this to `android/foreign_items.rs`. We can always move it to some shared place later.

src/shims/unix/linux/foreign_items.rs
```diff
@@ -72,6 +72,60 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                // We do not use `check_shim` here because `syscall` is variadic. The argument
```
RalfJung, line 76:
```suggestion
                // We do not use `check_shim` here because `prctl` is variadic. The argument
```

src/shims/unix/linux/foreign_items.rs
```diff
@@ -72,6 +72,60 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                this.check_abi_and_shim_symbol_clash(abi, Abi::C { unwind: false }, link_name)?;
+
+                if args.is_empty() {
+                    throw_ub_format!(
+                        "incorrect number of arguments for prctl: got 0, expected at least 1"
+                    );
+                }
+
+                // Since only two prctl calls are currently handled, and both expext just two arguments,
+                // it makes sense to handle validation here, outside of the match block below.
```
RalfJung, line 87: No please don't do this -- you are now showing an error that says PR_SET_NAME befre even checking whether that's the op we are executing.

src/shims/unix/linux/foreign_items.rs
```diff
@@ -72,6 +72,60 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                        "incorrect number of arguments for `PR_SET_NAME` prctl: got {}, expected at least 2",
+                        args.len()
+                    );
+                }
+
+                for (i, arg) in args.iter().enumerate() {
+                    let val = this.read_scalar(arg)?.to_i32()?;
+                    if val != 0 {
+                        throw_ub_format!(
+                            "incorrect value of {}th argument for 'PR_SET_NAME' prctl: got {}, expected 0L",
```
RalfJung, line 99: Not sure what `0L` is. We're writing Rust here, not C. :)
Why should all arguments have to be 0? This check is quite strange and seems to reject most callers...?

src/shims/unix/linux/foreign_items.rs
```diff
@@ -72,6 +72,60 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                        this.pthread_setname_np(tid, name, name_len)?
+                    }
+                    op if op == pr_get_name => {
+                        let tid = this.linux_gettid()?;
+                        let name = this.read_scalar(&args[1])?;
+                        let name_len = Scalar::from_target_usize(16, this);
+
+                        this.pthread_getname_np(tid, name, name_len)?
+                    }
+                    _ => return Ok(EmulateItemResult::NotSupported),
```
RalfJung, line 125: Usually we do `throw_unsup_format!` here. It is often quite confusing to silently return "not supported" for things that all real systems support.

src/shims/unix/linux/foreign_items.rs
YohDeadfall: My mistake, forgot that I included the name into the message.

src/shims/unix/linux/foreign_items.rs
YohDeadfall: I saw a lot of `prctl` calls from the C code where additional arguments were provided while they were unused. And the function documents state that it's okay.

RalfJung: But you are checking that *all* arguments are 0, not just all "additional" arguments?

Anyway, please follow what `fnctl` does. It just ignores extra arguments entirely.

src/shims/unix/linux/foreign_items.rs
```diff
@@ -72,6 +72,60 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    if val != 0 {
+                        throw_ub_format!(
+                            "incorrect value of {}th argument for 'PR_SET_NAME' prctl: got {}, expected 0L",
+                            i,
+                            val
+                        );
+                    }
+                }
+
+                let op = this.read_scalar(&args[0])?.to_i32()?;
```
RalfJung, line 106: Please move this logic to a separate function. It can be in the same file for now.

tests/pass-dep/libc/prctl.rs
```diff
@@ -0,0 +1,56 @@
...
+                unsafe {
+                    libc::pthread_getname_np(libc::pthread_self(), name.as_mut_ptr().cast(), name.len())
+                }
+            } else {
+                compile_error!("get_thread_name not supported for this OS")
+            }
+        }
+    }
+
+    let result = thread::Builder::new().name(long_name.clone()).spawn(move || {
```
RalfJung, line 39: This is testing setting thread names via the builder. We already have a test for that, you just need to enable it for Android.

We also should have a test directly invoking `prctl`. But that shouldn't use the Builder at all. It can just set/get the main thread name.

src/shims/unix/linux/foreign_items.rs
YohDeadfall: Right, another thing caused by late time coding (: I will fix it when I come to the keyboard.

tests/pass-dep/libc/prctl.rs
YohDeadfall: Since I'm moving the code to in `android/foreign_items.rs` the description should be like `Supported for Android only" or something like that.

src/shims/unix/linux/foreign_items.rs
YohDeadfall: By the way, it can be also used on FreeBSD which isn't Linux, but not on macOS which FreeBSD based.

YohDeadfall:
@RalfJung, I made changes almost like you requested, but I didn't move `prctl` handling into a separate function since it takes not so much space now and there's not so much code around. Should it still be moved to a separate function? I checked the code around, so I guess it can be done like `futex` and move to the same file as where other thread handling functions reside.

Another thing I still cannot finish is the test. I tried to runt it locally, but it ends up with the following message:

```
full stderr:
error[E0425]: cannot find value `PR_SET_NAME` in crate `libc`
  --> tests/pass-dep/libc/prctl.rs:28:40
   |
LL |             unsafe { libc::prctl(libc::PR_SET_NAME, name.as_ptr().cast::<*const c_char>()) }
   |                                        ^^^^^^^^^^^ help: a constant with a similar name exists: `PR_SET_VMA`
   |
  ::: /home/yoh/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libc-0.2.158/src/unix/linux_like/android/mod.rs:3138:1
   |
LL | pub const PR_SET_VMA: ::c_int = 0x53564d41;
   | ----------------------------- similarly named constant `PR_SET_VMA` defined here

error[E0425]: cannot find value `PR_GET_NAME` in crate `libc`
  --> tests/pass-dep/libc/prctl.rs:41:35
   |
LL |                 libc::prctl(libc::PR_GET_NAME, name.as_mut_ptr().cast::<*mut c_char>())
   |                                   ^^^^^^^^^^^ not found in `libc`

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0425`.
```

Meanwhile these constants exist and I definitely can compile the code.

tiif:
From what I understand from the CI test error,

```bash
  Error: the compiler panicked
  Error: you likely need to bless the tests with `./miri test --bless`
  Error: 
     0: ui tests in tests/pass for aarch64-linux-android failed
     1: tests failed
  
  Location:
     /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ui_test-0.26.5/src/lib.rs:357
  
  Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
  Run with RUST_BACKTRACE=full to include source snippets.
  
  Caused by:
    process didn't exit successfully: `/Users/runner/work/miri/miri/target/debug/deps/ui-0ba442cf755d06fa empty_main integer vec string btreemap hello hashmap heap_alloc align panic/panic panic/unwind concurrency/simple atomic libc-mem libc-misc libc-random env num_cpus pthread threadname --skip pthread_cond_timedwait` (exit status: 1)
  Error: command exited with non-zero code `cargo +miri test --locked --all-features --manifest-path /Users/runner/work/miri/miri/./Cargo.toml -- empty_main integer vec string btreemap hello hashmap heap_alloc align panic/panic panic/unwind concurrency/simple atomic libc-mem libc-misc libc-random env num_cpus pthread threadname --skip pthread_cond_timedwait`: 1
  error: test got exit status: 101, but expected 0
   = note: the compiler panicked
  
  Execute `./miri test --bless` to update `tests/pass/concurrency/threadname.stderr` to the actual output
  --- tests/pass/concurrency/threadname.stderr
  +++ <stderr output>
  +thread 'rustc' panicked at src/helpers.rs:LL:CC:
  +failed to find required Rust item: ["libc", "PR_SET_NAME"]
```

I guess it is because ``aarch64-linux-android`` platform in libc indeed doesn't have ``PR_SET_NAME``. There is no result when I try to search it in libc https://docs.rs/libc/latest/aarch64-linux-android/libc/?search=PR_SET_NAME.

YohDeadfall:
So, `libc` hasn't these constants for Android, but as soon as rust-lang/libc#3941 merged the CI will be unblocked. Surely it requires a new version of `libc` released.

src/shims/unix/android/foreign_items.rs
```diff
@@ -25,8 +26,63 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            "prctl" => {
+                // We do not use `check_shim` here because `prctl` is variadic. The argument
+                // count is checked bellow.
+                this.check_abi_and_shim_symbol_clash(abi, Abi::C { unwind: false }, link_name)?;
+
+                check_args_len("prctl", args, 1)?;
+
+                let id = this.read_scalar(&args[0])?.to_i32()?;
+                let pr_set_name = this.eval_libc_i32("PR_SET_NAME");
+                let pr_get_name = this.eval_libc_i32("PR_GET_NAME");
```
tiif, line 39: Are there any more android target that doesn't support these two flags?

YohDeadfall: According to the git history these constants were added in 2011, so I guess it's supported everywhere now.

RalfJung:
> So, `libc` hasn't these constants for Android, but as soon as [rust-lang/libc#3941](https://github.com/rust-lang/libc/pull/3941) merged the CI will be unblocked. Surely it requires a new version of `libc` released.

That could take a while, so we shouldn't block on it. You can hard-code the numbers now and leave a FIXME pointing to the libc PR.

src/shims/unix/linux/foreign_items.rs
YohDeadfall: Found that piece of code even in the `std` library:

```rust
#[cfg(target_os = "android")]
pub fn set_name(name: &CStr) {
    const PR_SET_NAME: libc::c_int = 15;
    unsafe {
        libc::prctl(
            PR_SET_NAME,
            name.as_ptr(),
            0 as libc::c_ulong,
            0 as libc::c_ulong,
            0 as libc::c_ulong,
        );
    }
}
```

Not required, but provided for an unknown reason.

YohDeadfall:
@RalfJung, I need a bit of help to finish this pull request. I run the `threadname` test and got this error:

```
error: Undefined Behavior: scalar size mismatch: expected 8 bytes but got 4 bytes instead
  --> /home/yoh/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/sys/pal/unix/thread.rs:120:13
   |
LL | /             libc::prctl(
LL | |                 PR_SET_NAME,
LL | |                 name.as_ptr(),
LL | |                 0 as libc::c_ulong,
LL | |                 0 as libc::c_ulong,
LL | |                 0 as libc::c_ulong,
LL | |             );
   | |_____________^ scalar size mismatch: expected 8 bytes but got 4 bytes instead
   |
   = note: BACKTRACE on thread `unnamed-2`:
   = note: inside `std::sys::pal::unix::thread::Thread::set_name` at /home/yoh/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/sys/pal/unix/thread.rs:120:13: 126:14
   = note: inside closure at /home/yoh/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/thread/mod.rs:514:17: 514:44
   = note: inside `<{closure@std::thread::Builder::spawn_unchecked_<'_, {closure@tests/pass/concurrency/threadname.rs:12:16: 12:23}, ()>::{closure#1}} as std::ops::FnOnce<()>>::call_once - shim(vtable)` at /home/yoh/.rustup/toolchains/miri/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5: 250:71
   = note: inside `<std::boxed::Box<dyn std::ops::FnOnce()> as std::ops::FnOnce<()>>::call_once` at /home/yoh/.rustup/toolchains/miri/lib/rustlib/src/rust/library/alloc/src/boxed.rs:2453:9: 2453:52
   = note: inside `<std::boxed::Box<std::boxed::Box<dyn std::ops::FnOnce()>> as std::ops::FnOnce<()>>::call_once` at /home/yoh/.rustup/toolchains/miri/lib/rustlib/src/rust/library/alloc/src/boxed.rs:2453:9: 2453:52
   = note: inside `std::sys::pal::unix::thread::Thread::new::thread_start` at /home/yoh/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/sys/pal/unix/thread.rs:105:17: 105:64
```

The return types both for `pthread_setname_np` returns `c_int`, and `prctl` does that as well. The first argument of `prctl` is `c_int` and it's passed as `c_int` in `std`:

```rust
#[cfg(target_os = "android")]
pub fn set_name(name: &CStr) {
    const PR_SET_NAME: libc::c_int = 15;
    unsafe {
        libc::prctl(
            PR_SET_NAME,
            name.as_ptr(),
            0 as libc::c_ulong,
            0 as libc::c_ulong,
            0 as libc::c_ulong,
        );
    }
}
```

So, I'm a bit lost. What the compiler expects to be 8-bytes long?

src/shims/unix/android/foreign_items.rs
```diff
@@ -25,8 +26,65 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                // FIXME: Use PR_SET_NAME and PR_GET_NAME constants when
+                // https://github.com/rust-lang/libc/pull/3941 lands.
+                let pr_set_name = 15_i32;
+                let pr_get_name = 16_i32;
+
+                let res = match id {
+                    id if id == pr_set_name => {
+                        check_args_len("'PR_SET_NAME' prctl", args, 2)?;
+
+                        let tid = this.linux_gettid()?;
```
tiif, line 47: The problem seems to be in this line, it works with ``let tid = this.pthread_self()?;``. 

The test report UB because of the size mismatch in this line:
https://github.com/rust-lang/miri/blob/bf0801d3d90e3077e2fd7bf4eaf913a4a499b5dc/src/shims/unix/thread.rs#L75

src/shims/unix/android/foreign_items.rs
```diff
@@ -25,8 +26,65 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+                        this.pthread_getname_np(tid, name, name_len)?
+                    }
+                    id => {
+                        this.handle_unsupported_foreign_item(format!(
+                            "can't execute prctl with ID {id}"
+                        ))?;
+                        return Ok(EmulateItemResult::AlreadyJumped);
+                    }
+                };
```
tiif, line 68:
```suggestion
                const PR_SET_NAME: i32 = 15;
                const PR_GET_NAME: i32 = 16;

                let res = match id {
                    PR_SET_NAME => {
                        check_args_len("'PR_SET_NAME' prctl", args, 2)?;

                        let tid = this.linux_gettid()?;
                        let name = this.read_scalar(&args[1])?;
                        let name_len = 16;

                        this.pthread_setname_np(tid, name, name_len)?
                    }
                    PR_GET_NAME => {
                        check_args_len("'PR_GET_NAME' prctl", args, 2)?;

                        let tid = this.linux_gettid()?;
                        let name = this.read_scalar(&args[1])?;
                        let name_len = Scalar::from_target_usize(16, this);

                        this.pthread_getname_np(tid, name, name_len)?
                    }
                    _ => {
                        this.handle_unsupported_foreign_item(format!(
                            "can't execute prctl with ID {id}"
                        ))?;
                        return Ok(EmulateItemResult::AlreadyJumped);
                    }
                };
```

src/shims/unix/android/foreign_items.rs
```diff
@@ -25,8 +26,65 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+fn check_args_len<'tcx>(
+    link_name: &str,
+    args: &[OpTy<'tcx>],
+    args_expected: usize,
+) -> InterpResult<'tcx, ()> {
+    let args_actual = args.len();
+    if args_actual < args_expected {
+        throw_ub_format!(
+            "incorrect number of arguments for {link_name}: got {args_actual}, expected at least {args_expected}"
+        );
```
tiif, line 87: I know that there are a lot of place using ``throw_ub_format!`` when the number of arguments is incorrect, so I am wondering if there is a consensus on when to use ``throw_unsup_format`` vs ``throw_ub_format``,because an incorrect number of arguments is not the kind of [Undefined Behaviour](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) I am thinking when it comes to miri.

tests/pass-dep/libc/prctl.rs
```diff
@@ -0,0 +1,47 @@
...
+    assert_eq!(get_thread_name(&mut buf), 0);
+}
+
+fn set_thread_name(name: &CStr) -> i32 {
+    cfg_if::cfg_if! {
+        if #[cfg(any(target_os = "android"))] {
+            unsafe { libc::prctl(libc::PR_SET_NAME, name.as_ptr().cast::<*const c_char>()) }
+        } else {
+            compile_error!("set_thread_name not supported for this OS")
+        }
```
tiif, line 31: I expect with ``//@only-target: android``, this test will only be run in ``android``. So it might not be necessary to check for the target os again?

tests/pass-dep/libc/prctl.rs
```diff
@@ -0,0 +1,47 @@
...
+fn main() {
+    let long_name = CString::new(
+        std::iter::once("test_named_thread_truncation")
+            .chain(std::iter::repeat(" yada").take(100))
+            .collect::<String>(),
+    )
+    .unwrap();
+
+    // prctl supports thread names up to 16 characters includding nul.
+    // And since the input string is longer it should fail.
```
tiif, line 15: Typo and suggestion: 
```suggestion
    // prctl supports thread names up to 16 characters including null.
    //  Since the input string is longer, it should fail.
```

YohDeadfall:
Thanks, @tiif! I adjusted the code as you suggested.

RalfJung:
Before merging we'll ask you to squash, yes. Once there was a review, please don't squash as squashing makes it very hard to see what changed since the last review -- always add a new commit.

src/shims/unix/android/foreign_items.rs
RalfJung: It is definitely UB to call a function with fewer arguments than it expects, and Miri should report this as such.

YohDeadfall: Switched back to `throw_ub_format`.

tests/pass-dep/libc/prctl.rs
RalfJung, line 1: Would it make sense to make this part of `tests/pass-dep/libc/pthread-threadname.rs`? Rename the file to just `threadname`, and then implement `set_thread_name` and `get_thread_name` using `prctl` on Android.

tests/pass-dep/libc/prctl.rs
```diff
@@ -0,0 +1,33 @@
...
+
+    let cstr = CString::new(&long_name.as_bytes()[..15]).unwrap();
+    assert_eq!(set_thread_name(&cstr), 0);
+
+    let mut buf = vec![0u8; 16];
+    assert_eq!(get_thread_name(&mut buf), 0);
+}
+
+fn set_thread_name(name: &CStr) -> i32 {
+    const PR_SET_NAME: i32 = 15;
```
RalfJung, line 26:
```suggestion
    // FIXME: use libc constants once https://github.com/rust-lang/libc/pull/3941 is released
    const PR_SET_NAME: i32 = 15;
```

src/shims/unix/android/foreign_items.rs
```diff
@@ -25,8 +26,65 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                this.write_scalar(res, dest)?;
+            }
+
             _ => return interp_ok(EmulateItemResult::NotSupported),
         }
         interp_ok(EmulateItemResult::NeedsReturn)
     }
 }
+
+fn check_args_len<'tcx>(
```
RalfJung, line 78: We already have `check_arg_count` in helpers.rs. So maybe name this `check_min_arg_count` and also put it in that file?

Seems like it would also make sense to make this return an array of the right size, like `check_arg_count`.

src/shims/unix/android/foreign_items.rs
```diff
@@ -25,8 +26,65 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.write_scalar(errno_place.to_ref(this).to_scalar(), dest)?;
             }
 
+            // Threading
+            "prctl" => {
```
RalfJung, line 30: This is still long enough to justify a separate function, I would say.

src/shims/unix/android/foreign_items.rs
```diff
@@ -25,8 +26,65 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                // We do not use `check_shim` here because `prctl` is variadic. The argument
+                // count is checked bellow.
+                this.check_abi_and_shim_symbol_clash(abi, Abi::C { unwind: false }, link_name)?;
+
+                check_args_len("prctl", args, 1)?;
+
+                let id = this.read_scalar(&args[0])?.to_i32()?;
+                // FIXME: Use PR_SET_NAME and PR_GET_NAME constants when
```
RalfJung, line 38:
```suggestion

                // FIXME: Use PR_SET_NAME and PR_GET_NAME constants when
```

src/shims/unix/android/foreign_items.rs
```diff
@@ -25,8 +26,65 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                // https://github.com/rust-lang/libc/pull/3941 lands.
+                const PR_SET_NAME: i32 = 15;
+                const PR_GET_NAME: i32 = 16;
+
+                let res = match id {
+                    PR_SET_NAME => {
+                        check_args_len("'PR_SET_NAME' prctl", args, 2)?;
```
RalfJung, line 45:
```suggestion
                        check_args_len("prctl(PR_SET_NAME, ...)", args, 2)?;
```

src/shims/unix/android/foreign_items.rs
```diff
@@ -25,8 +26,65 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+                        let tid = this.pthread_self()?;
+                        let name = this.read_scalar(&args[1])?;
+                        let name_len = 16;
+
+                        this.pthread_setname_np(tid, name, name_len)?
+                    }
+                    PR_GET_NAME => {
+                        check_args_len("'PR_GET_NAME' prctl", args, 2)?;
```
RalfJung, line 54:
```suggestion
                        check_args_len("prctl(PR_GET_NAME, ...)", args, 2)?;
```

src/shims/unix/android/foreign_items.rs
YohDeadfall: It contradicts with another suggestion if `check_args_len` made even with `check_arg_count` as you asked. Should then `check_min_arg_count` accept a function name as the first argument?

RalfJung: Yeah it seems to me it would make sense to pass the function name to `check_min_arg_count`. (`check_arg_count` doesn't take a function name and we should probably fix that.)

src/helpers.rs
```diff
@@ -1216,6 +1216,19 @@ where
...
 
+/// Check that the number of args is at least the minumim what we expect
+pub fn check_min_arg_count<'a, 'tcx, const N: usize>(
+    name: &'a str,
+    args: &[OpTy<'tcx>],
+) -> InterpResult<'tcx, &'a [OpTy<'tcx>; N]> {
+    if args.len() >= N {
+        let ptr = args.as_ptr() as *const [OpTy<'tcx>; N];
+        // SAFETY: ok because we just checked that the length fits
+        return interp_ok(unsafe { &*ptr });
```
RalfJung, line 1227: Please do this without unsafe code:
```rust
return interp_ok(args[..N].try_into().unwrap());
```

src/helpers.rs
```diff
@@ -1216,6 +1216,19 @@ where
...
+    }
+    throw_ub_format!("incorrect number of arguments for {name}: got {}, expected {}", args.len(), N)
```
RalfJung, line 1229:
```suggestion
    throw_ub_format!("incorrect number of arguments for {name}: got {}, expected at least {}", args.len(), N)
```

src/shims/unix/linux/thread.rs
RalfJung, line 1: Okay it didn't have to be in a separate file, but we can do this too. :)

But this file should be in the `android` folder.

tests/pass-dep/libc/threadname.rs
```diff
@@ -19,6 +19,10 @@ fn main() {
                 0
             } else if #[cfg(target_os = "macos")] {
                 unsafe { libc::pthread_setname_np(name.as_ptr().cast()) }
+            } else if #[cfg(target_os = "android")] {
+                // FIXME: Use PR_SET_NAME constant when https://github.com/rust-lang/libc/pull/3941 lands.
+                const PR_SET_NAME: i32 = 15;
+                unsafe { libc::prctl(PR_SET_NAME, name.as_ptr().cast::<*const c_char>()) }
```
RalfJung, line 25:
```suggestion
                unsafe { libc::prctl(PR_SET_NAME, name.as_ptr().cast::<libc::c_char>()) }
```
You made this cast to a pointer-to-pointer.

tests/pass-dep/libc/threadname.rs
```diff
@@ -42,6 +46,10 @@ fn main() {
                     libc::pthread_get_name_np(libc::pthread_self(), name.as_mut_ptr().cast(), name.len())
                 };
                 0
+            } else if #[cfg(target_os = "android")] {
+                // FIXME: Use PR_GET_NAME constant when https://github.com/rust-lang/libc/pull/3941 lands.
+                const PR_GET_NAME: i32 = 16;
+                unsafe { libc::prctl(PR_GET_NAME, name.as_mut_ptr().cast::<*mut c_char>()) }
```
RalfJung, line 52:
```suggestion
                unsafe { libc::prctl(PR_GET_NAME, name.as_mut_ptr().cast::<libc::c_char>()) }
```

src/helpers.rs
YohDeadfall: There is a better option, `split_first_chunk`.

RalfJung: Ah, that's very nice. :)

RalfJung:
This looks great. :) If CI is happy, please squash the commits (with `rebase --keep-base` to avoid diff bloat).

YohDeadfall:
Okay, I dived into `std` and found that for `prctl` there's no truncation done, and that it's completely delegated to the kernel space. There was even no assertion that it doesn't fail.

RalfJung:
Ah, good catch!

Maybe make it so that `pthread_setname_np` takes a flag saying whether it should truncate or error on a too-long string? Then we can share more of the implementation.

YohDeadfall:
Not sure about it, really, as it would complicate the logic there and it's not a responsibility of `pthread_setname_np` at all. But I'm thinking of terms of the real world implementation, and it might not be applicable here.

I would better dive first into `bionic` and help you supporting Android with which I've never worked before, and come back with my opinion. So, for now just tell what do you wish (:

RalfJung:
I mostly wish things to be coherent. :) So either we consistently use the "nice" pthread_{set,get}name APIs from pctll, or we directly call the underlying thread manager functions.

So then the alternative I would propose is to directly call `get_thread_name` and `write_c_str` in PR_GET_NAME.

src/shims/unix/env.rs
```diff
@@ -276,7 +276,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
     fn linux_gettid(&mut self) -> InterpResult<'tcx, Scalar> {
         let this = self.eval_context_ref();
-        this.assert_target_os("linux", "gettid");
+        this.assert_target_os_is_one_of(&["linux", "android"], "gettid");
```
RalfJung, line 279: This diff seems unnecessary? `linux_gettid` is not called from any new places in this PR, is it?

YohDeadfall: Yes, not anymore. Will remove.

YohDeadfall:
> So then the alternative I would propose is to directly call `get_thread_name` and `write_c_str` in PR_GET_NAME. Maybe factor `b"<unnamed>"` into a global constant `DEFAULT_THREAD_NAME`.

That sounds definitely better.

src/shims/unix/android/thread.rs
```diff
@@ -0,0 +1,72 @@
...
+            let name_out = name_out.to_pointer(this)?;
+
+            let thread = this.pthread_self()?.to_int(this.libc_ty_layout("pthread_t").size)?;
+            let thread = ThreadId::try_from(thread).unwrap();
+
+            // FIXME: we should use the program name if the thread name is not set
+            let name = this.get_thread_name(thread).unwrap_or(DEFAULT_THREAD_NAME).to_owned();
+            let name_len = name.len().max(TASK_COMM_LEN - 1);
+
+            this.eval_context_mut().write_bytes_ptr(
```
YohDeadfall, line 55: The reason why `write_c_str` isn't used here is to follow up the `prctl` behavior, and let `write_bytes_ptr` to verify the destination length to see if the buffer doesn't match `TASK_COMM_LEN`.

RalfJung: `write_c_str` can do that, though?

RalfJung: You should be able to do something like `write_c_str` on `&name[..name_len]`.

YohDeadfall: Yes, it can do that, but it has a `size` check inside it which has no meaning in the current situation (the `name` is truncated already). Anyway, I will follow up your suggestion.

RalfJung: Still better than duplicating the null termination handling.

YohDeadfall: Actually, it also requires changes to the input string because `write_c_str` requires the source and destination buffers sizes to have the same length. Then it should collected into a `Vec` with appended `0u8` to it, and only then to be passed `write_c_str` which internally will do the check, append one more `0u8` and only then call `write_bytes_ptr`.

YohDeadfall: Rewrote it using `write_c_str`. There's no null terminator handling, but additional nulls appended at the end.

RalfJung: Why are you appending additional nulls at the end?

YohDeadfall: That's due to these lines ([source](https://github.com/rust-lang/rust/blob/baaf3e65abab5a281715230bcb1785610629ae7d/compiler/rustc_const_eval/src/interpret/memory.rs#L1281-L1286)):

```rust
// `zip` would stop when the first iterator ends; we want to definitely
// cover all of `bytes`.
for dest in bytes {
    *dest = src.next().expect("iterator was shorter than it said it would be");
}
assert_matches!(src.next(), None, "iterator was longer than it said it would be");
```

The destination and source buffers must have exactly the same size, and `PR_GET_NAME` requires the buffer to be exactly 16 bytes long ([docs](https://www.man7.org/linux/man-pages/man2/PR_GET_NAME.2const.html)):

```c
int prctl(PR_GET_NAME, const char name[16]);
```

In contrast, `pthread_getname_np` accepts the buffer length ([source](https://www.man7.org/linux/man-pages/man3/pthread_getname_np.3.html)):

```c
int pthread_getname_np(pthread_t thread, char name[.size], size_t size);
```

Though, it doesn't mean that it will return a name back because the `size` must be at least like `TASK_COMM_LEN` or even longer.

```
The pthread_getname_np() function can fail with the following
error:

ERANGE The buffer specified by name and size is too small to hold
       the thread name.
```

The code from `bionic` (I don't have another `libc` implementation openned right now, so let's use it as an example) demonstrates that check ([source](https://android.googlesource.com/platform/bionic/+/8a18af52d9b9344497758ed04907a314a083b204/libc/bionic/pthread_setname_np.cpp)):

```c
__BIONIC_WEAK_FOR_NATIVE_BRIDGE
int pthread_getname_np(pthread_t t, char* buf, size_t buf_size) {
  ErrnoRestorer errno_restorer;
  if (buf_size < MAX_TASK_COMM_LEN) return ERANGE;
```

I also assume that you would ask why `pthread_getname_np` cannot be used in `std` instead of `prctl`, so the question is it requires `bionic` being compiled with `_GNU_SOURCE` defined. Presence of `pthread_getname_np` in `bionic` without `_GNU_SOURCE` is just a bug. So, no it cannot be use (:

I feel that I'm close to mind frapping if it hasn't started yet, so fixing the tests is a task for tomorrow.

YohDeadfall: Forgot to mention, according to the explanation above, you have `pthread_getname_np` implemented wrongly. It must accept only `TASK_COMM_LEN` sized buffers and reject others with `ERANGE`. I will fix that too.

RalfJung: I would say for a thread name of length N < 15, `prctl` only writes the first N+1 bytes and leaves the rest untouched. There is *a* null terminator, but why do you want to fill *the entire rest of the buffer* with NULL? That's not usually how we implement APIs like this.

src/shims/unix/android/thread.rs
```diff
@@ -0,0 +1,69 @@
...
+) -> InterpResult<'tcx> {
+    // We do not use `check_shim` here because `prctl` is variadic. The argument
+    // count is checked bellow.
+    this.check_abi_and_shim_symbol_clash(abi, Abi::C { unwind: false }, link_name)?;
+
+    // FIXME: Use PR_SET_NAME and PR_GET_NAME constants when
+    // https://github.com/rust-lang/libc/pull/3941 lands.
+    const PR_SET_NAME: i32 = 15;
+    const PR_GET_NAME: i32 = 16;
+    const TASK_COMM_LEN: usize = 16;
```
RalfJung, line 23: Seems like the logic would be nicer if you make this `16 - 1`, so that this is really the length of the name.
```suggestion
    const MAX_NAME_LEN: usize = 16-1; // 16 *including* the null terminator
```

src/shims/unix/android/thread.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    let [op] = check_min_arg_count("prctl", args)?;
+    let res = match this.read_scalar(op)?.to_i32()? {
+        PR_SET_NAME => {
+            let [_, name] = check_min_arg_count("prctl(PR_SET_NAME, ...)", args)?;
+
+            let name = this.read_scalar(name)?.to_pointer(this)?;
+            let name =
+                this.read_c_str(name)?.iter().take(TASK_COMM_LEN - 1).copied().collect::<Vec<_>>();
```
RalfJung, line 32: That seems like a complicated way to express truncation. Instead you can do:
```rust
let name = this.read_c_str(name)?:
let name = name[..=MAX_NAME_LEN.max(name.len()].to_owned();
```

src/shims/unix/android/thread.rs
```diff
@@ -0,0 +1,69 @@
...
+            let thread = ThreadId::try_from(thread).unwrap();
+
+            this.set_thread_name(thread, name);
+            Scalar::from_u32(0)
+        }
+        PR_GET_NAME => {
+            let [_, name] = check_min_arg_count("prctl(PR_GET_NAME, ...)", args)?;
+
+            let name_out = this.read_scalar(name)?;
+            let name_out = name_out.to_pointer(this)?;
```
RalfJung, line 44:
```suggestion
            let name_out = this.read_pointer(name)?;
```

src/shims/unix/android/thread.rs
```diff
@@ -0,0 +1,69 @@
...
+    const PR_SET_NAME: i32 = 15;
+    const PR_GET_NAME: i32 = 16;
+    const TASK_COMM_LEN: usize = 16;
+
+    let [op] = check_min_arg_count("prctl", args)?;
+    let res = match this.read_scalar(op)?.to_i32()? {
+        PR_SET_NAME => {
+            let [_, name] = check_min_arg_count("prctl(PR_SET_NAME, ...)", args)?;
+
+            let name = this.read_scalar(name)?.to_pointer(this)?;
```
RalfJung, line 30:
```suggestion
            let name = this.read_pointer(name)?;
```

src/shims/unix/android/thread.rs
```diff
@@ -0,0 +1,69 @@
...
+                name.drain(name_len..);
+            } else {
+                name.resize(name_len, 0);
+            }
+
+            this.write_c_str(&name, name_out, TASK_COMM_LEN as u64)?;
+            Scalar::from_u32(0)
+        }
+        op => {
+            this.handle_unsupported_foreign_item(format!("can't execute prctl with OP {op}"))?;
```
RalfJung, line 63: This doesn't work here, it would need a different `EmulateItemResult` and that's not wired up this far down.

Just call `throw_unsup_format!`.

src/shims/unix/android/thread.rs
YohDeadfall: > There is a null terminator, but why do you want to fill the entire rest of the buffer with NULL? That's not usually how we implement APIs like this.

It was my tired mind doing shit. I read the `miri`'s code incorrectly, so my picture was that the `write_bytes_ptr` requires filling the whole allocated buffer. That's correct, but an earlier call to `get_ptr_alloc_mut` request the exact amount of memory to be borrowed, and it throws if it goes out of the bounds.

src/helpers.rs
```diff
@@ -1196,6 +1196,21 @@ where
     throw_ub_format!("incorrect number of arguments: got {}, expected {}", args.len(), N)
 }
 
+/// Check that the number of args is at least the minumim what we expect
+pub fn check_min_arg_count<'a, 'tcx, const N: usize>(
+    name: &'a str,
+    args: &'a [OpTy<'tcx>],
+) -> InterpResult<'tcx, &'a [OpTy<'tcx>; N]> {
+    if let Some((ops, _)) = args.split_first_chunk() {
```
RalfJung, line 1204: Would be really nice to get this helper in and start using it for all our variadic shims. :)

RalfJung: Maybe it'd make sense to split that into a separate PR, to avoid it being held up by getting the Android thing right?

OTOH I think this PR is fairly close to done.

YohDeadfall: Yeah, that's a good idea.

RalfJung:
All right I think all the other cleanup is done. :D  So after a rebase + squash this PR should be only adding the new shims, and extend the threadname test accordingly.

src/shims/unix/android/thread.rs
```diff
@@ -0,0 +1,53 @@
...
+    abi: Abi,
+    args: &[OpTy<'tcx>],
+    dest: &MPlaceTy<'tcx>,
+) -> InterpResult<'tcx> {
+    // We do not use `check_shim` here because `prctl` is variadic. The argument
+    // count is checked bellow.
+    this.check_abi_and_shim_symbol_clash(abi, Abi::C { unwind: false }, link_name)?;
+
+    let pr_set_name = this.eval_libc("PR_SET_NAME");
+    let pr_get_name = this.eval_libc("PR_GET_NAME");
```
RalfJung, line 22: Please just hard-code the values for now, with a FIXME saying we can use the values from libc once https://github.com/rust-lang/libc/pull/3941 gets backported to the 0.2 branch.

Cargo.lock
RalfJung, line 1: The lockfile in test_dependencies is the only one that actually matters here; please leave all the others unchanged.

src/shims/unix/thread.rs
```diff
@@ -64,23 +64,29 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
     /// Set the name of the specified thread. If the name including the null terminator
-    /// is longer than `name_max_len`, then `false` is returned.
+    /// is longer or equals to `name_max_len`, then if `truncate` is set the truncated name
+    /// is used as the thread name, otherwise `false` is returned.
     fn pthread_setname_np(
         &mut self,
         thread: Scalar,
         name: Scalar,
         name_max_len: usize,
+        truncate: bool,
```
RalfJung, line 74: This seems to be `false` for every single caller... so why add it at all?

tests/pass-dep/libc/threadname.rs
RalfJung, line 1: Seeing this now, I am not convinced any more that it is a good idea to mix the prctl and pthread threadname tests into one. I didn't realize how different the Android function behaves. So your original plan of making this a separate test, let's call it `prctl-threadname`, is probably better.

Sorry for that! Sometimes one has to see something to realize it is a bad idea.

src/shims/unix/thread.rs
YohDeadfall: It's `true` for `prctl` only.

tests/pass-dep/libc/threadname.rs
YohDeadfall: It's not Android, but the Linux kernel (: Ad one thing I forgot is links to the source code and a short explanation why.

src/shims/unix/thread.rs
```diff
@@ -64,23 +64,29 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
         let thread = ThreadId::try_from(thread).unwrap();
         let name = name.to_pointer(this)?;
-        let name = this.read_c_str(name)?.to_owned();
+        let mut name = this.read_c_str(name)?.to_owned();
 
         // Comparing with `>=` to account for null terminator.
         if name.len() >= name_max_len {
-            return interp_ok(false);
+            if truncate {
+                name.drain(name_max_len.saturating_sub(1)..);
```
RalfJung, line 86: Please use `name.truncate` instead. Also, `name_max_len` seems like a bug to me, so please use `strict_sub`.

tests/pass-dep/libc/threadname.rs
RalfJung: > It's not Android, but the Linux kernel (:

Sure, I meant prctl vs pthread_{set,get}_name. ;)

> one thing I forgot is links to the source code and a short explanation why.

No need to link the source code, the linux kernel has nice docs on the man page that we can link / copy instead.

src/shims/unix/android/thread.rs
```diff
@@ -0,0 +1,53 @@
...
+                this.pthread_setname_np(thread, name, TASK_COMM_LEN, /* truncate */ true)?;
+            assert!(res);
+            Scalar::from_u32(0)
+        }
+        op if op == pr_get_name => {
+            let [_, name] = check_min_arg_count("prctl(PR_GET_NAME, ...)", args)?;
+            let name = this.read_scalar(name)?;
+            let thread = this.pthread_self()?;
+            let len = Scalar::from_target_usize(TASK_COMM_LEN as u64, this);
+            if this.pthread_getname_np(thread, name, len, /* truncate*/ false)? {
```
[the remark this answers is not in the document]
RalfJung: Ah right I entirely forgot, please add a `fail-dep` test checking that we mark UB when the buffer is only 15 bytes large, *even if* the thread name is even shorter.

You can call `check_ptr_access` to check that here. IMO that shows the error message we want.

miri-script/Cargo.lock
RalfJung, line 1: There's still an unnecessary lockfile change here.

src/shims/unix/android/thread.rs
```diff
@@ -0,0 +1,66 @@
...
+                    CheckInAllocMsg::MemoryAccessTest,
+                )
+                .report_err()
+                .is_err()
+            {
+                throw_ub_format!(
+                    "`prctl(PR_GET_NAME, name)` requires the `name` argument to be at least {} bytes long",
+                    TASK_COMM_LEN
+                );
+            };
```
RalfJung, line 57:
```suggestion
            this
                .check_ptr_access(
                    name.to_pointer(this)?,
                    Size::from_bytes(TASK_COMM_LEN),
                    CheckInAllocMsg::MemoryAccessTest,
                )?;
```
Never "catch" interpreter errors, always propagate them via `?`.

src/shims/unix/android/thread.rs
```diff
@@ -0,0 +1,66 @@
...
+            if this
+                .check_ptr_access(
+                    name.to_pointer(this)?,
+                    Size::from_bytes(TASK_COMM_LEN),
+                    CheckInAllocMsg::MemoryAccessTest,
+                )
+                .report_err()
+                .is_err()
+            {
+                throw_ub_format!(
```
RalfJung, line 53: This UB needs a test.

tests/pass-dep/libc/pthread-threadname.rs
```diff
@@ -1,4 +1,5 @@
 //@ignore-target: windows # No pthreads on Windows
+//@ignore-target: android # No pthreads to get/set name on Android
```
RalfJung, line 2:
```suggestion
//@ignore-target: android # No pthread_{get,set}_name on Android
```

tests/pass-dep/libc/prctl-threadname.rs
```diff
@@ -0,0 +1,73 @@
...
+    let long_name = std::iter::once("test_named_thread_truncation")
+        .chain(std::iter::repeat(" yada").take(100))
+        .collect::<String>();
+
+    fn set_thread_name(name: &CStr) -> i32 {
+        unsafe { libc::prctl(libc::PR_SET_NAME, name.as_ptr().cast::<libc::c_char>()) }
+    }
+
+    fn get_thread_name(name: &mut [u8]) -> i32 {
+        unsafe { libc::prctl(libc::PR_GET_NAME, name.as_mut_ptr().cast::<libc::c_char>()) }
```
RalfJung, line 21: Please add an assert that the buffer is at least 16 bytes large. Otherwise this is not even a sound function.

tests/pass-dep/libc/prctl-threadname.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    // Set name via Rust API, get it via prctl.
+    let long_name2 = long_name.clone();
+    thread::Builder::new()
+        .name(long_name.clone())
+        .spawn(move || {
+            let mut buf = vec![0u8; long_name2.len() + 1];
```
RalfJung, line 29:
```suggestion
            let mut buf = vec![0u8; MAX_THREAD_NAME_LEN];
```
Given that this is a hard property of this API, let's just hard-code the length here as well.

tests/pass-dep/libc/prctl-threadname.rs
```diff
@@ -0,0 +1,73 @@
...
+            assert_eq!(get_thread_name(&mut buf), 0);
+            let cstr = CStr::from_bytes_until_nul(&buf).unwrap();
+            let truncated_name = &long_name2[..long_name2.len().min(MAX_THREAD_NAME_LEN - 1)];
+            assert_eq!(cstr.to_bytes(), truncated_name.as_bytes());
+        })
+        .unwrap()
+        .join()
+        .unwrap();
+
+    // Set name via pthread and get it again (short name).
```
RalfJung, line 39:
```suggestion
    // Set name via prctl and get it again (short name).
```

tests/pass-dep/libc/prctl-threadname.rs
```diff
@@ -0,0 +1,73 @@
...
+            let mut buf = vec![0u8; MAX_THREAD_NAME_LEN];
+            assert_eq!(get_thread_name(&mut buf), 0);
+            let cstr = CStr::from_bytes_until_nul(&buf).unwrap();
+            assert_eq!(cstr.to_bytes(), short_name.as_bytes());
+        })
+        .unwrap()
+        .join()
+        .unwrap();
+
+    // Set name via pthread and get it again (long name).
```
RalfJung, line 56:
```suggestion
    // Set name via prctl and get it again (long name).
```

RalfJung (commented):
@rustbot author

src/shims/unix/android/thread.rs
YohDeadfall: It was made, but I forgot to add its untracked files.

src/shims/unix/android/thread.rs
YohDeadfall: The idea isn't to catch an error but to translate it into something more meaningful. Otherwise, there's no reason in `check_ptr_access` since `check_and_deref_ptr` will be called anyway.

RalfJung: > The idea isn't to catch an error but to translate it into something more meaningful.

That's a nice idea but not currently something we support properly. The right way to do that would be to use `CheckInAllocMsg`, since that is its purpose. Please stick to our usual patterns.

> Otherwise, there's no reason in check_ptr_access since check_and_deref_ptr will be called anyway.

If that's the case then your test isn't sufficient. Consider the case where the thread name has length 4, and we provide a buffer of size 10.  That should still be UB, the syscall requries 16 bytes.

YohDeadfall: > If that's the case then your test isn't sufficient. Consider the case where the thread name has length 4, and we provide a buffer of size 10. That should still be UB, the syscall requries 16 bytes.

Forgot that the internal check does it a bit differently, so you're right. The additional `check_ptr_access` is needed before getting the thread name.

YohDeadfall:
Just a few hundred comments more and that pull request can be a new "legend" in the `rustc` contribution docs. Unfortunately not because there's something interesting going on, but because I constantly forget something. So, thank you for your patience (:

tests/fail-dep/libc/prctl-threadname.rs
```diff
@@ -0,0 +1,8 @@
+//@only-target: android  # Miri supports prctl for Android only
```
RalfJung, line 1:
```suggestion
//! Ensure we report UB when the buffer is smaller than 16 bytes (even if the thread
//! name would fit in the smaller buffer).
//@only-target: android  # Miri supports prctl for Android only
```

tests/fail-dep/libc/prctl-threadname.rs
RalfJung, line 1: Fail tests should roughly indicate what they are about in the filename...  so I suggest `prctl-get-name-buffer-too-small`.

RalfJung (changes requested):
I think this is the last round. :)
Please squash (with `get rebase --keep-base`) after resolving the last two comments.

RalfJung:
Awesome, now please just squash (with `git rebase --keep-base`) the commits :)

RalfJung:
Thanks!

@bors r+

PR #3909, opened 2024-09-24
Scope CI permissions to the job that needs it

Kobzol:
I scoped the permissions only to the job that requires them.


RalfJung:
Thanks! I did not know you can scope permissions. Now <https://github.com/rust-lang/miri/pull/3908> already landed, which conflicts. Could you rebase, and also keep the comments that my PR added?

RalfJung:
This seems strictly better than what we have in-tree currently, so I think we should just land it and see what happens tomorrow morning.

@bors r+

PR #3910, opened 2024-09-24
Android: Fixed tests for libc time API

### LINKED ISSUE #3888: Android: implement support for libc time APIs
We currently don't support getting the current time on Android.
The test command for this is `./miri --target aarch64-linux-android libc-time`.

YohDeadfall:
Closes #3888.

There's nothing to do, really. All of the API is already supported by `miri`, but I just tweaked tests a bit to match the data structures and functions from `bionic`.

### DIFF
--- a/src/shims/time.rs
+++ b/src/shims/time.rs
@@ -36,8 +36,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let mut relative_clocks;
 
         match this.tcx.sess.target.os.as_ref() {
-            "linux" | "freebsd" => {
-                // Linux and FreeBSD have two main kinds of clocks. REALTIME clocks return the actual time since the
+            "linux" | "freebsd" | "android" => {
+                // Linux, Android, and FreeBSD have two main kinds of clocks. REALTIME clocks return the actual time since the
                 // Unix epoch, including effects which may cause time to move backwards such as NTP.
                 // Linux further distinguishes regular and "coarse" clocks, but the "coarse" version
                 // is just specified to be "faster and less precise", so we implement both the same way.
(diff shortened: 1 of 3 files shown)


src/shims/time.rs
```diff
@@ -36,7 +36,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let mut relative_clocks;
 
         match this.tcx.sess.target.os.as_ref() {
-            "linux" | "freebsd" => {
+            "linux" | "freebsd" | "android" => {
                 // Linux and FreeBSD have two main kinds of clocks. REALTIME clocks return the actual time since the
```
RalfJung, line 40: Does the comment here apply to Android as well? If yes, please make it say "Linux, Android, and FreeBSD"

RalfJung:
Ah, that's great, thanks. :)

Unfortunately there are conflicts, can you please rebase? Also, I have a comment nit, see above.

src/shims/time.rs
YohDeadfall: Yes, it applies to Android as well. The thing with Android actually is it's based on LTS version of Linux, so technically it's Linux ([source](https://source.android.com/docs/core/architecture/kernel)):

> The Android kernel is based on an upstream [Linux Long Term Supported (LTS) kernel](https://www.kernel.org/). At Google, LTS kernels are combined with Android-specific patches to form what are known as Android Common Kernels (ACKs).

So, shouldn't the code be more unified? If you're okay, I can check all `Linux specific foreign items to be supported by Android looking into its `libc`, and then refactor under your guidance.

RalfJung: I know Android shares the kernel with Linux, but the userland is quite different. I don't know Android well enough to judge whether we'd better treat them as separate Unix flavors in Miri, or as a single flavor with some occasional special-casing.

One concern with treating them as one Unix flavor is that we don't have an Android expert in the team, so we'd want to avoid someone adding a new shim for Linux and accidentally enabling it on Android without testing that. If we treated shims more declaratively, that would be fairly easy to enforce... without that, the best we can do is probably a comment in the Linux file.

So yeah this seems worth exploring, if you think it has merit. In a future PR. :)

PR #3918, opened 2024-09-26
implements arc4random_buf shim for freebsd/solarish platforms.

### LINKED ISSUE #3914: Implement arc4random_buf for FreeBSD, Solarish
Ever since https://github.com/rust-lang/rust/pull/129201, HashMap no longer works on FreeBSD and Solarish targets since we don't have a shim for `arc4random_buf`. So to get back basic std support, that shim should be implemented.

Cc @devnexen (listed target maintainer for Solarish)

devnexen:
close #3914

### DIFF
--- a/tests/pass-dep/libc/libc-random.rs
+++ b/tests/pass-dep/libc/libc-random.rs
@@ -6,6 +6,8 @@ fn main() {
     test_getentropy();
     #[cfg(not(target_os = "macos"))]
     test_getrandom();
+    #[cfg(any(target_os = "freebsd", target_os = "illumos", target_os = "solaris"))]
+    test_arc4random_buf();
 }
 
 fn test_getentropy() {
@@ -60,4 +62,15 @@ fn test_getrandom() {
             5,
         );
     }
+}
+
+#[cfg(any(target_os = "freebsd", target_os = "illumos", target_os = "solaris"))]
+fn test_arc4random_buf() {
+    // FIXME: Use declaration from libc once <https://github.com/rust-lang/libc/pull/3944> lands.
+    extern "C" {
+        fn arc4random_buf(buf: *mut libc::c_void, size: libc::size_t);
+    }
+    let mut buf = [0u8; 5];
+    unsafe { arc4random_buf(buf.as_mut_ptr() as _, buf.len()) };
+    assert!(buf != [0u8; 5]);
 }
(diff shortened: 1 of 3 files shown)


tests/pass-dep/libc/libc-random.rs
```diff
@@ -61,3 +63,12 @@ fn test_getrandom() {
         );
     }
 }
+
+#[cfg(any(target_os = "solaris", target_os = "illumos", target_os = "solaris"))]
+fn test_arc4random_buf() {
+    let mut buf = [0u8; 5];
+    unsafe {
+        libc::arc4random_buf(buf.as_mut_ptr(), buf.len());
```
RalfJung, line 71: `libc` doesn't seem to have this on Solarish targets. For now you can just [copy what `std` does](https://github.com/rust-lang/rust/blob/b9d47cfa9b8f0805bfab2c254a02e598a906f102/library/std/src/sys/random/arc4random.rs#L12-L30).

tests/pass-dep/libc/libc-random.rs
```diff
@@ -61,3 +63,12 @@ fn test_getrandom() {
...
+        assert!(buf[0] != 0u8);
```
RalfJung, line 72: This can fail, if we happen to get all-zero... 1/256 is not *so* unlikely, that means it'll fail on rustc CI around once a month. So let's do:
```suggestion
        assert!(buf != [0u8; 5]);
```

RalfJung:
Thanks a lot for the quick PR :)

tests/pass-dep/libc/libc-random.rs
devnexen: true

tests/pass-dep/libc/libc-random.rs
```diff
@@ -61,3 +63,15 @@ fn test_getrandom() {
         );
     }
 }
+
+#[cfg(any(target_os = "freebsd", target_os = "illumos", target_os = "solaris"))]
+fn test_arc4random_buf() {
+    extern "C" {
```
RalfJung, line 69:
```suggestion
    // FIXME: Use declaration from libc once <https://github.com/rust-lang/libc/pull/3944> lands.
    extern "C" {
```

tests/pass-dep/libc/libc-random.rs
```diff
@@ -61,3 +63,15 @@ fn test_getrandom() {
...
+        fn arc4random_buf(buf: *mut libc::c_void, size: libc::size_t);
+    }
+    let mut buf = [0u8; 5];
+    unsafe {
+        arc4random_buf(buf.as_mut_ptr() as _, buf.len());
+        assert!(buf != [0u8; 5]);
+    }
```
RalfJung, line 76:
```suggestion
        arc4random_buf(buf.as_mut_ptr() as _, buf.len())
    };
    assert!(buf != [0u8; 5]);
```

RalfJung:
I will take the liberty yo make this as "ready for review".

RalfJung:
Looks good, thanks. :)
@bors r+

PR #3920, opened 2024-09-27
Update cc to 1.1.22

ChrisDenton:
This version of `cc` contains a fix to prevent spurious rebuilds. Hopefully this should help avoid the CI issues rustc has been having.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -137,9 +137,12 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.1.7"
+version = "1.1.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "26a5c3fd7bfa1ce3897a3a3501d362b2d87b7f2583ebcb4a949ec25911025cbc"
+checksum = "9540e661f81799159abee814118cc139a2004b3a3aa3ea37724a1b66530b90e0"
+dependencies = [
+ "shlex",
+]
 
 [[package]]
 name = "cfg-if"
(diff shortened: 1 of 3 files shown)


RalfJung:
Note that the Miri and rustc repos have separate lockfiles, so this PR won't make any difference on the rustc side even if we do a sync.

ChrisDenton:
> Note that the Miri and rustc repos have separate lockfiles, so this PR won't make any difference on the rustc side even if we do a sync.

Ah, that'll be taken care of by the rustc update. I knew there were now separate lock files but it looks like that's just a compiler/libs split.

tgross35:
Do you mean that, in the main repo, files like `src/tools/miri/Cargo.lock` `src/tools/miri/bench-cargo-miri/backtraces/Cargo.lock`, and `src/tools/miri/test_dependencies/Cargo.lock` are not used at all and the root `Cargo.lock` is used instead? Or that those files are not identical to the ones in this repo and need to get `cargo update`d separately?

RalfJung:
> Do you mean that, in the main repo, files like `src/tools/miri/Cargo.lock` `src/tools/miri/bench-cargo-miri/backtraces/Cargo.lock`, and `src/tools/miri/test_dependencies/Cargo.lock` are not used at all and the root `Cargo.lock` is used instead? 

yes, that's what I mean -- but only for some of these lockfiles, namely the ones that are in the rustc workspace. AFAIK that's `Cargo.lock` and `cargo-miri/Cargo.lock`, but I am fairly sure it does *not* include `test_dependencies/Cargo.lock`.

PR #3923, opened 2024-09-28
Refactor fd.rs

### LINKED ISSUE #3904: Refactor fd.rs ``return_read_bytes_and_count`` and ``return_written_byte_count_or_error``
It'd be clearer if ``return_read_bytes_and_count`` and ``return_written_byte_count_or_error`` can be combined and separate according to the ``Ok`` and ``Err`` code path. This is mentioned in https://github.com/rust-lang/miri/pull/3852#issuecomment-2355169819. 

(This probably don't need to be an issue, but I am creating this to keep track of what I want to do.)

tiif:
Fixes #3904

This PR combined the ``return_read_bytes_and_count`` and ``return_written_byte_count_or_error `` into ``return_read_write_success`` and added a helper function ``set_io_error_and_return_neg_one`` to set the ``io::Error`` and write ``-1`` to dest place.


src/shims/unix/fd.rs
```diff
@@ -182,7 +186,10 @@ impl FileDescription for io::Stdout {
         // the host -- there is no good in adding extra buffering
         // here.
         io::stdout().flush().unwrap();
-        ecx.return_written_byte_count_or_error(result, dest)
+        match result {
+            Ok(rw_bytes) => ecx.return_read_write_success(None, None, false, rw_bytes, dest),
```
RalfJung, line 190: These calls are very hard to read. What does `None, None, false` mean? I shouldn't have to know the argument names to make sense of a function call.

src/helpers.rs
```diff
@@ -846,6 +846,17 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         self.set_last_error(self.io_error_to_errnum(err)?)
     }
 
+    /// Sets the last OS error using a `std::io::ErrorKind` and writes -1 to dest place.
+    fn set_io_error_and_return_neg_one(
```
RalfJung, line 850: Should be `set_last_error_and...`. It's not setting an io error, after all.

Maybe just "and return" is enough, leaving `-1` as the usual convention?

So what about `set_last_error_and_return`.

src/shims/unix/fd.rs
```diff
@@ -655,47 +664,30 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
 
-    /// Helper to implement `FileDescription::read`:
-    /// `result` should be the return value of some underlying `read` call that used `bytes` as its output buffer.
+    /// Helper to implement `FileDescription::read/write`:
     /// The length of `bytes` must not exceed either the host's or the target's `isize`.
-    /// If `Result` indicates success, `bytes` is written to `buf` and the size is written to `dest`.
-    /// Otherwise, `-1` is written to `dest` and the last libc error is set appropriately.
-    fn return_read_bytes_and_count(
+    /// `is_read` is used to indicate if this helper is used for `read`.
+    /// If it is `read`, `bytes` is written to `buf`.
```
RalfJung, line 670: No please don't merge the read and write helpers, not for the success case. That's too confusing.

src/helpers.rs
```diff
@@ -846,6 +846,17 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        &mut self,
+        err: io::Error,
```
RalfJung, line 852: If you use `err: impl Into<io::Error>`, some of the call sites should become a bit simpler.

src/shims/unix/linux/eventfd.rs
RalfJung, line 1: There is code in `read` that could also use your new helper.

RalfJung (commented):
Thanks for the PR!

src/shims/unix/fd.rs
tiif: Yup, after taking a second look, I realise it is indeed confusing.

src/helpers.rs
tiif: Wow, yea this is better, now I can do ``ecx.set_last_error_and_return(ErrorKind::BrokenPipe, dest)``

RalfJung: Exactly. :)

RalfJung:
This looks great, thanks. :)
Please squash the commits.

PR #3927, opened 2024-09-29
let rustfmt format imports

RalfJung:
This matches the recent change in rustc.


oli-obk (approved):

saethlin:
Looks good to me. I much prefer things like formatting to be automated.

RalfJung:
All right, let's go then.

@bors r=oli-obk,saethlin

PR #3936, opened 2024-10-03
Added rust-analyzer instructions for Helix

YohDeadfall:
That pull request adds information on how to configure Helix to use `rust-analyzer`, and moves the existing configuration to the `src/etc` directory as it's in the `rust` repository. Not adding instructions for other IDE because there's a link leading to the how-to for `rustc`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -195,48 +195,42 @@ installed (`cargo install hyperfine`).
 
 ## Configuring `rust-analyzer`
 
-To configure `rust-analyzer` and VS Code for working on Miri, save the following
-to `.vscode/settings.json` in your local Miri clone:
+To configure `rust-analyzer` and the IDE for working on Miri, use one of the provided
+configuration files according to the instructions below.
 
-```json
-{
-    "rust-analyzer.rustc.source": "discover",
-    "rust-analyzer.linkedProjects": [
-        "Cargo.toml",
-        "cargo-miri/Cargo.toml",
-        "miri-script/Cargo.toml",
-    ],
-    "rust-analyzer.check.invocationLocation": "root",
-    "rust-analyzer.check.invocationStrategy": "once",
-    "rust-analyzer.check.overrideCommand": [
-        "env",
-        "MIRI_AUTO_OPS=no",
-        "./miri",
-        "clippy", // make this `check` when working with a locally built rustc
-        "--message-format=json",
-    ],
-    // Contrary to what the name suggests, this also affects proc macros.
-    "rust-analyzer.cargo.buildScripts.invocationLocation": "root",
-    "rust-analyzer.cargo.buildScripts.invocationStrategy": "once",
-    "rust-analyzer.cargo.buildScripts.overrideCommand": [
-        "env",
-        "MIRI_AUTO_OPS=no",
-        "./miri",
-        "check",
-        "--message-format=json",
-    ],
-}
-```
 
-> #### Note
+### Visual Studio Code
+
+Copy [`etc/rust_analyzer_vscode.json`] to `.vscode/settings.json` in the project root directory.
+
+> #### Hint
 >
-> If you are [building Miri with a locally built rustc][], set
-> `rust-analyzer.rustcSource` to the relative path from your Miri clone to the
-> root `Cargo.toml` of the locally built rustc. For example, the path might look
-> like `../rust/Cargo.toml`.
+> To keep the `rust-analyzer` configuration up-to-date, make a symbolic link to one 
+> of the provided files depending on the IDE you use.
+
+[`etc/rust_analyzer_vscode.json`]: https://github.com/rust-lang/miri/blob/master/etc/rust_analyzer_vscode.json
+
+### Helix
+
+Copy [`etc/rust_analyzer_helix.toml`] to `.helix/languages.toml` in the project root directory.
+
+Since working on Miri requires a custom toolchain, and Helix requires the language server
+to be installed with the toolchain, you have to run `./miri toolchain -c rust-analyzer`
+when installing the Miri toolchain. Alternatively, set the `RUSTUP_TOOLCHAIN` environment variable according to
+[the documentation](https://rust-analyzer.github.io/manual.html#toolchain).
+
+[`etc/rust_analyzer_helix.toml`]: https://github.com/rust-lang/miri/blob/master/etc/rust_analyzer_helix.toml
+
+### Advanced configuration
+
+If you are building Miri with a locally built rustc, set
+`rust-analyzer.rustcSource` to the relative path from your Miri clone to the
+root `Cargo.toml` of the locally built rustc. For example, the path might look
+like `../rust/Cargo.toml`. In addition to that, replace `clippy` by `check`
+in the `rust-analyzer.check.overrideCommand` setting.
 
 See the rustc-dev-guide's docs on ["Configuring `rust-analyzer` for `rustc`"][rdg-r-a]
-for more information about configuring VS Code and `rust-analyzer`.
+for more information about configuring the IDE and `rust-analyzer`.
 
 [rdg-r-a]: https://rustc-dev-guide.rust-lang.org/building/suggested.html#configuring-rust-analyzer-for-rustc
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -195,18 +195,13 @@ installed (`cargo install hyperfine`).
 
 ## Configuring `rust-analyzer`
 
-To configure `rust-analyzer` and the IDE for working on Miri, use one of the provided
-configuration files according to the instructions below.
-
+To configure `rust-analyzer` and the IDE for working on Miri, copy one of the provided
+configuration files according to the instructions below. You can also set up a symbolic
+link to keep the configuration in sync with our recommendations.
 
 ### Visual Studio Code
 
 Copy [`etc/rust_analyzer_vscode.json`] to `.vscode/settings.json` in the project root directory.
-
-> #### Hint
->
-> To keep the `rust-analyzer` configuration up-to-date, make a symbolic link to one 
-> of the provided files depending on the IDE you use.
 
 [`etc/rust_analyzer_vscode.json`]: https://github.com/rust-lang/miri/blob/master/etc/rust_analyzer_vscode.json
 
--- a/etc/rust_analyzer_helix.toml
+++ b/etc/rust_analyzer_helix.toml
@@ -1,0 +1,32 @@
+[language-server.rust-analyzer.config.rustc]
+source = "discover"
+
+[language-server.rust-analyzer.config]
+linkedProjects = [
+    "Cargo.toml",
+    "cargo-miri/Cargo.toml",
+    "miri-script/Cargo.toml",
+]
+
+[language-server.rust-analyzer.config.check]
+invocationLocation = "root"
+invocationStrategy = "once"
+overrideCommand = [
+    "env",
+    "MIRI_AUTO_OPS=no",
+    "./miri",
+    "clippy", # make this `check` when working with a locally built rustc
+    "--message-format=json",
+]
+
+# Contrary to what the name suggests, this also affects proc macros.
+[language-server.rust-analyzer.config.buildScripts]
+invocationLocation = "root"
+invocationStrategy = "once"
+overrideCommand = [
+    "env",
+    "MIRI_AUTO_OPS=no",
+    "./miri",
+    "check",
+    "--message-format=json",
+]
--- a/etc/rust_analyzer_vscode.json
+++ b/etc/rust_analyzer_vscode.json
@@ -1,0 +1,27 @@
+{
+    "rust-analyzer.rustc.source": "discover",
+    "rust-analyzer.linkedProjects": [
+        "Cargo.toml",
+        "cargo-miri/Cargo.toml",
+        "miri-script/Cargo.toml",
+    ],
+    "rust-analyzer.check.invocationLocation": "root",
+    "rust-analyzer.check.invocationStrategy": "once",
+    "rust-analyzer.check.overrideCommand": [
+        "env",
+        "MIRI_AUTO_OPS=no",
+        "./miri",
+        "clippy", // make this `check` when working with a locally built rustc
+        "--message-format=json",
+    ],
+    // Contrary to what the name suggests, this also affects proc macros.
+    "rust-analyzer.cargo.buildScripts.invocationLocation": "root",
+    "rust-analyzer.cargo.buildScripts.invocationStrategy": "once",
+    "rust-analyzer.cargo.buildScripts.overrideCommand": [
+        "env",
+        "MIRI_AUTO_OPS=no",
+        "./miri",
+        "check",
+        "--message-format=json",
+    ],
+}


src/etc/rust_analyzer_helix.toml
RalfJung, line 1: I don't think we want this inside `src`.  Maybe just `etc`?

YohDeadfall:
One more thing before finishing the pull request. `rust-analyzer` won't work just out of the box due to a custom toolchain created by the `./miri` script. Should then it be mentioned that `./miri toolchain` requires `-c rust-analyzer`? An alternative solution is to override the toolchain in the configuration files, or just add `-c rust-analyzer` directly to the script.

RalfJung:
I'm using RA and I am not setting `-c rust-analyzer` anwhere. So not sure what you mean?

YohDeadfall:
It depends how it was set up. For example, if I have no overrides in the config and no `rust-analyzer` for the `miri` toolchain then I can see that in the logs:

```
2024-10-03T23:29:00.949 helix_lsp::transport [ERROR] rust-analyzer err <- "error: 'rust-analyzer' is not installed for the custom toolchain 'miri'.\n"
2024-10-03T23:29:00.949 helix_lsp::transport [ERROR] rust-analyzer err <- "note: this is a custom toolchain, which cannot use `rustup component add`\n"
2024-10-03T23:29:00.949 helix_lsp::transport [ERROR] rust-analyzer err <- "help: if you built this toolchain from source, and used `rustup toolchain link`, then you may be able to build the component with `x.py`\n"
```

You don't have any issue, so I assume that you made a custom toolchain yourself with the analyzer and linked it, right?

RalfJung:
> You don't have any issue, so I assume that you made a custom toolchain yourself with the analyzer and linked it, right?

No, I am just using `./miri toolchain` unchanged. With the vscode plugin, this just works. :shrug:

YohDeadfall:
Interesting, maybe it's the plugin's magic, don't know... Then it can merged as is for now, or wait when I see what happens with VS Code.

Squash changes?

YohDeadfall:
I checked why you don't get that error, and the reason is that the VS Code extension isn't a simple bridge between the IDE and `rust-analyzer`. It does a bit of smart job while bootstrapping the server ([soucre](https://github.com/rust-lang/rust-analyzer/blob/5982d9c420d0dc90739171829f0d2e9c80d98979/editors/code/src/bootstrap.ts#L58-L77)):

```ts
for (const workspaceFolder of vscode.workspace.workspaceFolders) {
    // otherwise check if there is a toolchain override for the current vscode workspace
    // and if the toolchain of this override has a rust-analyzer component
    // if so, use the rust-analyzer component
    const toolchainUri = vscode.Uri.joinPath(workspaceFolder.uri, "rust-toolchain.toml");
    if (await hasToolchainFileWithRaDeclared(toolchainUri)) {
        const res = spawnSync("rustup", ["which", "rust-analyzer"], {
            encoding: "utf8",
            env: { ...process.env },
            cwd: workspaceFolder.uri.fsPath,
        });
        if (!res.error && res.status === 0) {
            toolchainServerPath = earliestToolchainPath(
                toolchainServerPath,
                res.stdout.trim(),
                raVersionResolver,
            );
        }
    }
}
```

It means that `-c rust-analyzer` should be used for Helix users at least.

CONTRIBUTING.md
```diff
@@ -195,48 +195,35 @@ installed (`cargo install hyperfine`).
...
+
+Copy [`src/etc/rust_analyzer_settings.json`] to `.vscode/settings.json` in the project root directory.
+
+### Helix
+
+Copy [`src/etc/rust_analyzer_helix.toml`] to `.helix/languages.toml` in the project root directory.
+
+Since working on Miri requires a custom toolchain consider running `./miri toolchain` with the
+`-c rust-analyzer` flag, or set the `RUSTUP_TOOLCHAIN` environment variable according to
+[the documentation](https://rust-analyzer.github.io/manual.html#toolchain).
```
YohDeadfall, line 211: Added a note how to properly configure `rust-analyzer` for working on Miri. Another option would be following up the referenced documentation and setting the environment variable in the provided config, but I'm not sure about that. It seems too implicit, and in theory `rust-analyzer` might not be able to support everything.

CONTRIBUTING.md
```diff
@@ -195,48 +195,35 @@ installed (`cargo install hyperfine`).
...
-        "--message-format=json",
-    ],
-}
-```
+To configure `rust-analyzer` and the IDE for working on Miri, use one of the provided
+configuration files according to the instructions below.
+
+### Visual Studio Code
+
+Copy [`src/etc/rust_analyzer_settings.json`] to `.vscode/settings.json` in the project root directory.
```
RalfJung, line 203: The paths are still pointing to `src/`

RalfJung:
> I checked why you don't get that error, and the reason is that the VS Code extension isn't a simple bridge between the IDE and rust-analyzer. It does a bit of smart job while bootstrapping the server ([soucre](https://github.com/rust-lang/rust-analyzer/blob/5982d9c420d0dc90739171829f0d2e9c80d98979/editors/code/src/bootstrap.ts#L58-L77)):

That logic will do nothing for Miri since we don't have  a `rust-toolchain.toml` file.

The actually relevant part is below:
```
// finally, use the bundled one
```
They simply bundle the server with the extension so it works without being installed via rustup.

CONTRIBUTING.md
```diff
@@ -195,48 +195,35 @@ installed (`cargo install hyperfine`).
...
+
+### Helix
+
+Copy [`src/etc/rust_analyzer_helix.toml`] to `.helix/languages.toml` in the project root directory.
+
+Since working on Miri requires a custom toolchain consider running `./miri toolchain` with the
+`-c rust-analyzer` flag, or set the `RUSTUP_TOOLCHAIN` environment variable according to
```
RalfJung, line 210:
```suggestion
Since working on Miri requires a custom toolchain, and Helix requires the language server
to be installed with the toolchain, you have to run `./miri toolchain -c rust-analyzer`
when installing the Miri toolchain. Alternatively, set the `RUSTUP_TOOLCHAIN` environment variable according to
```

CONTRIBUTING.md
```diff
@@ -195,48 +195,35 @@ installed (`cargo install hyperfine`).
...
+>
+> To keep the `rust-analyzer` configuration up-to-date, make a symbolic link to the provided file.
 
 > #### Note
 >
 > If you are [building Miri with a locally built rustc][], set
 > `rust-analyzer.rustcSource` to the relative path from your Miri clone to the
 > root `Cargo.toml` of the locally built rustc. For example, the path might look
-> like `../rust/Cargo.toml`.
+> like `../rust/Cargo.toml`. In addition to that replace `clippy` by `check` command
```
RalfJung, line 222:
```suggestion
> like `../rust/Cargo.toml`. In addition to that, replace `clippy` by `check`
```

CONTRIBUTING.md
```diff
@@ -195,48 +195,35 @@ installed (`cargo install hyperfine`).
...
+
+Since working on Miri requires a custom toolchain consider running `./miri toolchain` with the
+`-c rust-analyzer` flag, or set the `RUSTUP_TOOLCHAIN` environment variable according to
+[the documentation](https://rust-analyzer.github.io/manual.html#toolchain).
+
+> #### Hint
+>
+> To keep the `rust-analyzer` configuration up-to-date, make a symbolic link to the provided file.
 
 > #### Note
```
RalfJung, line 217: This now looks like the note is specific for Helix, which is not the case.

Any suggestion for how to avoid that confusion?

etc/rust_analyzer_settings.json
```diff
@@ -0,0 +1,26 @@
...
+    "rust-analyzer.check.invocationLocation": "root",
+    "rust-analyzer.check.invocationStrategy": "once",
+    "rust-analyzer.check.overrideCommand": [
+        "env",
+        "MIRI_AUTO_OPS=no",
+        "./miri",
+        "clippy",
+        "--message-format=json",
+    ],
+    "rust-analyzer.cargo.buildScripts.invocationLocation": "root",
```
RalfJung, line 17:
```suggestion
    // Contrary to what the name suggests, this also affects proc macros.
    "rust-analyzer.cargo.buildScripts.invocationLocation": "root",
```

etc/rust_analyzer_helix.toml
```diff
@@ -0,0 +1,33 @@
...
+]
+
+[language-server.rust-analyzer.config.check]
+invocationLocation = "root"
+invocationStrategy = "once"
+overrideCommand = [
+    "env",
+    "MIRI_AUTO_OPS=no",
+    "./miri",
+    "clippy", # make this `check` when working with a locally built rustc
```
RalfJung, line 18: Why did you keep this comment here but remove it in the vscode file?

etc/rust_analyzer_settings.json
RalfJung, line 1: Shouldn't this be `rust_analyzer_vscode.json` to match the naming scheme?

etc/rust_analyzer_helix.toml
```diff
@@ -0,0 +1,33 @@
...
+    "--message-format=json",
+]
+
+# Contrary to what the name suggests, this also affects proc macros.
+[language-server.rust-analyzer.config.buildScripts]
+enable = true
```
RalfJung, line 24: Why is this needed here? The vscode file doesn't need it.

CONTRIBUTING.md
YohDeadfall: I haven't decided how it is better yet. Moving it before the IDE paragraphs makes it too bold and the follow up seems disconnected. I will think about it.

etc/rust_analyzer_settings.json
YohDeadfall: Comments aren't supported by JSON officially, so even when it was in a code block, these lines had a red background suggesting that it's invalid.

etc/rust_analyzer_settings.json
YohDeadfall: Yeah, makes sense. Originally I followed the naming from the main repo, but it definitely can be improved.

etc/rust_analyzer_settings.json
RalfJung: It works in vscode though, so IMO that's a bug in github's JSON rendered and we should ignore it.

CONTRIBUTING.md
```diff
@@ -195,48 +195,35 @@ installed (`cargo install hyperfine`).
...
 >
 > If you are [building Miri with a locally built rustc][], set
```
RalfJung, line 219: This link does not work, it seems. Probably pre-existing, but since you are touching this anyway...
```suggestion
> If you are building Miri with a locally built rustc, set
```

etc/rust_analyzer_helix.toml
YohDeadfall: Sorry, a leak from my config I made according to the documentation from `rustc`. The idea behind this setting is running build scripts for more precise code analysis. That's enabled for the `rust` repo.

I can mirror it for VS Code or remove it from here. What do you think?

RalfJung: Aren't these enabled by default? AFAIK at least in vscode, they are.

etc/rust_analyzer_settings.json
YohDeadfall: That's not a bug, but a feature (: Some parsers go against the specs and allow comments in JSON.

etc/rust_analyzer_helix.toml
YohDeadfall: According to the docs, yes, but in the settings from `rustc` it is enabled explicitly.

etc/rust_analyzer_settings.json
RalfJung: What's relevant here is the vscode parser, and it (reasonably) supports comments. So we should use them -- it is quite useful to be able to explain configuration.

I'm not going to merge this with the comments removed.

etc/rust_analyzer_helix.toml
RalfJung: Not sure why they are doing that, but no need to copy it then -- let's just rely on the default.

etc/rust_analyzer_settings.json
YohDeadfall: I already added them.

CONTRIBUTING.md
YohDeadfall: Seems better now. Are you OK with it?

CONTRIBUTING.md
```diff
@@ -195,48 +195,38 @@ installed (`cargo install hyperfine`).
...
+> If you are building Miri with a locally built rustc, set
 > `rust-analyzer.rustcSource` to the relative path from your Miri clone to the
 > root `Cargo.toml` of the locally built rustc. For example, the path might look
-> like `../rust/Cargo.toml`.
+> like `../rust/Cargo.toml`. In addition to that, replace `clippy` by `check`
+> in the `rust-analyzer.check.overrideCommand` setting.
+
+### Visual Studio Code
+
+Copy [`etc/rust_analyzer_settings.json`] to `.vscode/settings.json` in the project root directory.
```
RalfJung, line 211: What is the point of the brackets around the filename?

These do not become links like this.

CONTRIBUTING.md
RalfJung: I would suggest to move the hint (about using a symlink) to the first section, the one about VSCode. People reading to-down will hopefully notice it then even if they are using a different IDE.

The note about "If you are building Miri with a locally built rustc" could move into the "Advanced configuration" section.

CONTRIBUTING.md
YohDeadfall: I thought that it was a style in [rustc-dev-guide](https://rustc-dev-guide.rust-lang.org/building/suggested.html#helix), but it seems that there are just broken links.

CONTRIBUTING.md
YohDeadfall: > People reading to-down will hopefully notice it then even if they are using a different IDE.

Maybe above the `### Visual Studio Code`? Then it wouldn't look like it applies only to VS Code, but also for Helix.

RalfJung: Then it is confusing because there haven't been any "files" yet that this comment refers to.

YohDeadfall: Make a symbolic link to the provided files below? Still the same place, but it point that the files are inside the IDE and Helix sections.

If you're not satisfied with that idea, I will just follow your suggestion to get it merged.

CONTRIBUTING.md
```diff
@@ -195,48 +195,43 @@ installed (`cargo install hyperfine`).
...
 
 > #### Note
 >
-> If you are [building Miri with a locally built rustc][], set
+> If you are building Miri with a locally built rustc, set
 > `rust-analyzer.rustcSource` to the relative path from your Miri clone to the
 > root `Cargo.toml` of the locally built rustc. For example, the path might look
-> like `../rust/Cargo.toml`.
+> like `../rust/Cargo.toml`. In addition to that, replace `clippy` by `check`
+> in the `rust-analyzer.check.overrideCommand` setting.
```
RalfJung, line 231: It's odd to make this a "note" but the next thing here not. Please just make them two paragraphs.

CONTRIBUTING.md
RalfJung: I don't understand the suggestion. But a forward reference will just be confusing I think, this hint can't come before the first concrete IDE configuration section.

CONTRIBUTING.md
YohDeadfall: It was originally a note, I haven't touched that at all, just moved as it was asked (:

RalfJung: Yeah fair. :) But previously there was only a single IDE so we didn't need any of this structure. If you want I can push push some edits to your PR, that's probably a more effective way to do the final fine-tuning.

RalfJung:
Okay, please squash the commits. :)

RalfJung:
Thanks for the patience! I did some minor final tweaks.

@bors r+

PR #3937, opened 2024-10-03
syscall eventfd2

FrankReh:
Add plumbing so syscall of SYS_eventfd2 can take advantage of the eventfd support already built into Miri.


FrankReh:
Marked as draft because of the TODO I left in the source. Hoping someone knowledge could point me to the conversion necessary.

FrankReh:
I should also point out that the SYS_eventfd could also be supported but I don't know if Rust crates are using it, as SYS_eventfd2 is a more modern libc call.

Also, the clippy fatal warning found by the CI/style check will either go away with a better conversion where I have the TODO comment or if that conversion is deemed necessary, there is probably some other version that doesn't suffer the clippy warning.

src/shims/unix/linux/foreign_items.rs
```diff
@@ -143,6 +144,21 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    id if id == sys_eventfd2 => {
+                        if args.len() < 3 {
+                            throw_ub_format!(
+                                "incorrect number of arguments for `eventfd2` syscall: got {}, expected at least 3",
+                                args.len()
+                            );
+                        }
+
+                        let result = this.eventfd(&args[1], &args[2])?;
+                        // TODO what's a cleaner way to go from a Scalar i32 to Scalar integer?
```
saethlin, line 156: I think `to_int` is what you're looking for? https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_middle/mir/interpret/value.rs.html#403-406

FrankReh: > I think `to_int` is what you're looking for? https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_middle/mir/interpret/value.rs.html#403-406

Thank you. Do you know what value for the Size type to pass in?

RalfJung: IMO going via a host `i32` is completely fine in this case.

The alternative would be to call `int_to_int_or_float`.

FrankReh: > IMO going via a host `i32` is completely fine in this case.

Sorry. I'm not understanding. When I first plumbed this together, I wasn't trying to change the Scalar::from_i32(fd_value) that the eventfd function returns and the Miri runtime was reporting  a write with uninitialized data. It took a lot of trial and error before hitting on the idea that the four bytes being created by the eventfd function weren't enough to write the result to an eight byte result that syscall is designed to return.

But is there something else you mean by host `i32`?

Also, I don't see how to convert the type I have to one that `this.int_to_int_or_float` would accept. I have a `Scalar<Provenance>` and the method takes an `ImmTy<'_, machine::Provenance>`.

tests/pass-dep/libc/libc_syscall_eventfd2.rs
```diff
@@ -0,0 +1,11 @@
+//@only-target: linux
+
+// This is a test for calling eventfd2 through a syscall.
+fn main() {
+    let initval = 0 as libc::c_uint;
+    let flags = (libc::EFD_CLOEXEC | libc::EFD_NONBLOCK) as libc::c_int;
+
+    let result = unsafe { libc::syscall(libc::SYS_eventfd2, initval, flags) };
+
+    assert_eq!(result, 3); // The first FD provided would be 3.
```
tiif, line 10: Just a nit, will it be better if the assert here is``assert_ne!(result, -1)``? Although the current implementation will always give 3 as first fd value, but imo this can change any time. We just need to make sure that it won't be an error, so any value other than -1 should pass the assert.

src/shims/unix/linux/foreign_items.rs
tiif: I don't have the answer to the question, but I wonder if it helps to do this:

```rust
let result = this.eventfd(&args[1], &args[2])?;
this.write_int(result.to_i32()?, dest)?;
```

RalfJung: Going via host `i32` is what you are doing by calling `to_i32()`.

And yes, the proposal @tiif makes looks great, and it's more compact than what the PR does, so let's do that. :)

FrankReh: But that doesn't compile.
```
error[E0277]: the trait bound `rustc_const_eval::interpret::Scalar<machine::Provenance>: std::convert::From<i32>` is not satisfied
   --> src/shims/unix/linux/foreign_items.rs:161:43
    |
161 |                         this.write_scalar(result.to_i32()?, dest)?;
    |                              ------------ ^^^^^^^^^^^^^^^^ the trait `std::convert::From<i32>` is not implemented for `rustc_const_eval::interpret::Scalar<machine::Provenance>`, which is required by `i32: std::convert::Into<rustc_const_eval::interpret::Scalar<machine::Provenance>>`
    |                              |
    |                              required by a bound introduced by this call
    |
    = help: the following other types implement trait `std::convert::From<T>`:
              `rustc_const_eval::interpret::Scalar<Prov>` implements `std::convert::From<rustc_apfloat::ieee::IeeeFloat<rustc_apfloat::ieee::DoubleS>>`
              `rustc_const_eval::interpret::Scalar<Prov>` implements `std::convert::From<rustc_apfloat::ieee::IeeeFloat<rustc_apfloat::ieee::HalfS>>`
              `rustc_const_eval::interpret::Scalar<Prov>` implements `std::convert::From<rustc_apfloat::ieee::IeeeFloat<rustc_apfloat::ieee::QuadS>>`
              `rustc_const_eval::interpret::Scalar<Prov>` implements `std::convert::From<rustc_apfloat::ieee::IeeeFloat<rustc_apfloat::ieee::SingleS>>`
              `rustc_const_eval::interpret::Scalar<Prov>` implements `std::convert::From<rustc_middle::ty::ScalarInt>`
    = note: required for `i32` to implement `std::convert::Into<rustc_const_eval::interpret::Scalar<machine::Provenance>>`
note: required by a bound in `rustc_const_eval::interpret::place::<impl rustc_const_eval::interpret::InterpCx<'tcx, M>>::write_scalar`
```

FrankReh: I'm removing the TODO comment.

tests/pass-dep/libc/libc_syscall_eventfd2.rs
FrankReh: Okay.

FrankReh:
Locally, I updated master from upstream but now find `./miri test` fails, even before my commits for this PR are added.
```
FAILED TEST: tests/pass/backtrace/backtrace-std.rs
command: ..elided..

Execute `./miri test --bless` to update `tests/pass/backtrace/backtrace-std.stderr` to the actual output
--- tests/pass/backtrace/backtrace-std.stderr
+++ <stderr output>
    0: func_d
  at tests/pass/backtrace/backtrace-std.rs:LL:CC
... 20 lines skipped ...
   11: std::panic::catch_unwind
  at RUSTLIB/std/src/panic.rs:LL:CC
-  12: std::rt::lang_start_internal::{closure#1}
+  12: std::rt::lang_start_internal::{closure#2}
  at RUSTLIB/std/src/rt.rs:LL:CC
   13: std::panicking::r#try::do_call
... 7 lines skipped ...
   17: std::rt::lang_start
  at RUSTLIB/std/src/rt.rs:LL:CC
```

src/shims/unix/linux/foreign_items.rs
oli-obk: write_int works, which was suggested above

RalfJung:
> Locally, I updated master from upstream but now find ./miri test fails, even before my commits for this PR are added.

This is likely an outdated toolchain. After each pull/rebase, run `./miri toolchain` to make sure you are still in sync.

src/shims/unix/linux/foreign_items.rs
FrankReh: Thanks for the help!

tests/pass-dep/libc/libc_syscall_eventfd2.rs
RalfJung, line 1: Please just add this to the existing `libc-eventfd` test.

RalfJung:
I am curious, is there some actual user for this or is it just "because we can"?

FrankReh:
> I am curious, is there some actual user for this or is it just "because we can"?

When I was going to try to test `futures-concurrency` with miri, I found it had two test cases disabled for cfg(miri). Turned out it was because the tests use async_io::Timer, and that uses polling::Poller which on Linux uses polling::epoll::Poller which seems like a misnomer, because that actually uses rustix::event::eventfd and that makes a syscall with SYS_eventfd2.

But that wouldn't be enough. If that syscall is ever supported, then there is still the whole timerfd that this stack also uses and that's not in Miri yet either. I thought I could offer timerfd once that was the sticking point. I'm not a maintainer of rustix and I'm not committed to using it.

So no, there is no actual user. It's all just because we could. We can close this.

tests/pass-dep/libc/libc_syscall_eventfd2.rs
FrankReh: Done.

RalfJung:
> So no, there is no actual user.

Well it seems like there's relevant real-world code that uses SYS_eventfd2, which I would count as an actual user. :)

RalfJung:
Looks good, thanks!

@bors r+

PR #3939, opened 2024-10-04
Implement blocking eventfd

tiif:
This PR 
- Implemented blocking for both read and write of eventfd
- Added test for eventfd blocking read and write
- Removed eventfd blocking tests from ``fail-dep``
- Added a new ``BlockReason::Eventfd``

cc #3665


tiif:
Hmm...why is rustfmt complaining? Let me try rebase it instead.

src/shims/unix/linux/eventfd.rs
```diff
@@ -217,8 +238,86 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            blocked_read_tid: RefCell::new(None),
+            blocked_write_tid: RefCell::new(None),
         });
 
         interp_ok(Scalar::from_i32(fd_value))
     }
 }
+
+/// Callback function after eventfd read unblocks.
+fn blocking_eventfd_read_callback<'tcx>(
```
oli-obk, line 250: For future PRs: it is easier to review if you split things out into functions with no changes in a commit first, and then adjust the function to suit your additional use case

oli-obk (changes requested):

src/shims/unix/linux/eventfd.rs
tiif: Good suggestion. I actually ended up opening the original copy and compare it side by side to make sure it is correct because I did quite a lot of change in one go. So this minor refactoring might need a bit more attention.

src/shims/unix/linux/eventfd.rs
```diff
@@ -74,28 +78,33 @@ impl FileDescription for Event {
...
         // Block when counter == 0.
         let counter = self.counter.get();
+        let weak_eventfd = self_ref.downgrade();
         if counter == 0 {
             if self.is_nonblock {
                 return ecx.set_last_error_and_return(ErrorKind::WouldBlock, dest);
             }
+            let dest = dest.clone();
+            let mut blocked_read_tid = self.blocked_read_tid.borrow_mut();
+            *blocked_read_tid = Some(ecx.active_thread());
```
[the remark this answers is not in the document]
tiif: Yea I missed that and that's indeed a possible case. Multithreaded case is a bit more complicated, I need more time to think about this, but this is the case I am considering: 

If there are main thread + 2 additional threads: 
let the counter be u64::max, so any ``write`` happened after will block. 
2. thread1 try to write, and get blocked.
3. thread2 try to write, also get blocked.
4. Main thread read. Now, will thread1 and thread2 both get unblocked? 

If both are unblocked, if thread1 write a u64::max again, thread2 should be blocked again. But the current implementation won't allow this to happen because thread2 is currently calling the unblock callback.

It also doesn't make sense to only unblock one thread, because the moment ``read`` happens, both thread1 and thread2 indeed can write.

I will check what actually happened in real system.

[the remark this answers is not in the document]
tiif: I think eventfd will unblock every thread at once. In this case, if this happens 

If there are main thread + 2 additional threads:
1. let the counter be (u64::max - 1), so any write happened after will block.
2. thread1 try to write, and get blocked.
3. thread2 try to write, also get blocked.
4. main thread 1, all thread gets unblocked
5. thread1 unblocked and writes (u64::max - 1)
6. thread 2 unblocked and try to continue writing 1, it gets block again

So essentially, we need to check the value and continue to block the thread again in the blocking callback functions. I am not sure how we can approach this without overcomplicated it because if we need to block a thread multiple times, we might run into something like

1. block_thread
2. in the blocking callback, block_thread
3. In the second block_thread callback, block_thread again
... and it continues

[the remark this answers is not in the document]
RalfJung: You can have a `fn eventfd_write` that blocks the thread if necessary, and in the "unblock" callback calls `eventfd_write` again. Recursion is a thing. :)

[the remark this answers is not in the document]
tiif: Yea, I think this will work. I feel this has an unfortunate consequence of making it more complex, but I couldn't see another solution for this.

tests/fail-dep/libc/eventfd_block_twice.stderr
tiif, line 1: This is expected as I haven't fixed https://github.com/rust-lang/miri/pull/3939#discussion_r1794559765. 

The desired behaviour is deadlock on thread2 instead of ICE because integer overflow.

tiif:
@oli-obk When running ``/miri test eventfd_block_twice  --bless``, the error below is shown. Is there any way to annotate the error when the diagnostics accured outside the test file? 

```rust
error: there were 1 unmatched diagnostics that occurred outside the testfile and had no pattern
    Error: deadlock: the evaluated program deadlocked

error: there were 1 unmatched diagnostics
 --> tests/fail-dep/libc/eventfd_block_twice.rs:1:1
  |
1 | //@only-target: linux
  | ^ Error: deadlock: the evaluated program deadlocked
  |

```

src/shims/unix/linux/eventfd.rs
```diff
@@ -217,8 +179,195 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            if eventfd.is_nonblock {
+                return ecx.set_last_error_and_return(ErrorKind::WouldBlock, dest);
+            }
+
+            let dest = dest.clone();
+
+            // We don't want to record the same thread if the current thread is still blocked.
+            if !reblock {
+                eventfd.blocked_write_tid.borrow_mut().push(ecx.active_thread());
+            }
```
tiif, line 300: If this function is recursively called, it will run into ``already borrowed: BorrowMutError`` panic and dropping ``blocked_write_tid`` like below won't work. 

```rust
let mut blocked_write_tid = eventfd.blocked_write_tid.borrow_mut();
blocked_write_tid.push(ecx.active_thread());
drop(blocked_write_tid)
```

So I use this to prevent reborrow, and prevent a thread id being pushed to ``blocked_write_tid`` again if it is already there.

RalfJung: That sounds like somewhere, a `RefMut` should be dropped earlier.

tiif: Yea, I will check again and see how can I prevent the panic. I don't really like the idea of adding ``reblock``.

tests/fail-dep/libc/eventfd_block_twice.rs
```diff
@@ -0,0 +1,72 @@
+//@only-target: linux
+//~^ERROR: deadlocked
+//~^^ERROR: deadlocked
```
tiif, line 3: I need these error annotations to make it pass, and it is kind of weird  :D

src/shims/unix/linux/eventfd.rs
tiif: Yea, the issue is indeed somewhere else.

src/shims/unix/linux/eventfd.rs
```diff
@@ -217,8 +179,205 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    while let Some(tid) = blocked_write_tid.pop() {
+        waiter.push(tid);
+    }
+    drop(blocked_write_tid);
+
+    waiter.sort();
+    waiter.dedup();
+    for thread_id in waiter {
+        ecx.unblock_thread(thread_id, BlockReason::Eventfd)?;
+    }
```
tiif, line 231: I couldn't think of a testcase that can actually produce thread id duplication in blocked_write/read_id, but I will just do keep the ``dedup`` here first.

src/shims/unix/linux/eventfd.rs
```diff
@@ -217,8 +179,205 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+    // Unblock *all* threads previously blocked on `write`.
+    // We need to store the blocked thread ids and unblock them together to prevent BorrowMutError
+    // panic because when unblock_thread is called, blocked_write_tid will be mutably borrowed again.
+    let mut waiter = Vec::new();
+    let mut blocked_write_tid = eventfd.blocked_write_tid.borrow_mut();
+    while let Some(tid) = blocked_write_tid.pop() {
+        waiter.push(tid);
+    }
+    drop(blocked_write_tid);
```
oli-obk, line 227:
```suggestion
let waiting_threads = std::mem::take(&mut *eventfd.blocked_write_tid.borrow_mut());
```

takes out all the waiting threads in one go

src/shims/unix/linux/eventfd.rs
oli-obk: we'll get a panic if that happens, so maybe just don't sort and dedup and wait until we get a test case. I don't think it can happen though, as a thread would need to get blocked again without having been removed and unblocked.

src/shims/unix/linux/eventfd.rs
```diff
@@ -217,8 +179,205 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    ecx.write_int(counter, buf_place)?;
+    eventfd.counter.set(0);
+
+    // When any of the event happened, we check and update the status of all supported event
+    // types for current file description.
+    ecx.check_and_update_readiness(&eventfd_ref)?;
+
+    // Unblock *all* threads previously blocked on `write`.
+    // We need to store the blocked thread ids and unblock them together to prevent BorrowMutError
+    // panic because when unblock_thread is called, blocked_write_tid will be mutably borrowed again.
```
oli-obk, line 221:
```suggestion
    // We need to take out the blocked thread ids and unblock them together,
    // because `unblock_threads` may block them again and end up re-adding the
    // thread to the blocked list.
```

src/shims/unix/linux/eventfd.rs
oli-obk: Also: the unblocking order is user-visible, and tho I assume that eventfd does not specify which thread gets woken first, we'd probably want a random (deterministic, but depending on the seed) order of unblocking to happen

cc @RalfJung I think we need a new system for unblocking multiple threads at the same time. So far we only ever unblocked one thread, and "unblock N threads, then randomly run them with the normal rules for picking the next thread" seems like a recurring thing.

Also I wonder if the normal thread unblocking is subtly wrong, too. Or just wrong in this PR:

When a thread gets unblocked and immediately performs some operation that is visible from other threads, then that means when the current thread continues after unblocking the other thread, it may behave as if the unblocked thread already had a few CPU cycles to do something. While that may be expected behaviour depending on preemption, the behaviour here does not depend on preemption. So, to avoid such a footgun, thread unblocking should not immediately execute the unblocking operation, but just unblock the thread and make the thread execute the unblocking operation when it gets scheduled next.

oli-obk (changes requested):
not finished with my review, but we need to figure out a few thread unblocking things first

src/shims/unix/linux/eventfd.rs
```diff
@@ -217,8 +179,205 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+    // Unblock *all* threads previously blocked on `read`.
+    // We need to store the blocked thread ids and unblock them together to prevent BorrowMutError
+    // panic because when unblock_thread is called, blocked_read_tid will be mutably borrowed again.
+    let mut waiter = Vec::new();
+    let mut blocked_read_tid = eventfd.blocked_read_tid.borrow_mut();
+    while let Some(tid) = blocked_read_tid.pop() {
+        waiter.push(tid);
+    }
+    drop(blocked_read_tid);
```
RalfJung, line 276: You can do `eventfd.blocked_read_tid.borrow_mut().take()`, then you don't need to push everything to a new vector.

src/shims/unix/linux/eventfd.rs
```diff
@@ -217,8 +179,205 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+    waiter.sort();
+    waiter.dedup();
```
RalfJung, line 279: This is sus. There should never be duplicates.

src/shims/unix/linux/eventfd.rs
```diff
@@ -217,8 +179,205 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        throw_unsup_format!("eventfd FD got closed while blocking.")
+    };
+
+    // Since we pass the weak file description ref, it is guaranteed to be
+    // an eventfd file description.
+    let eventfd = eventfd_ref.downcast::<Event>().unwrap();
+
+    match eventfd.counter.get().checked_add(num) {
+        Some(_new_count @ 0..=MAX_COUNTER) => {
+            return eventfd_write(num, &buf_place, dest, weak_eventfd, ecx);
```
RalfJung, line 307: Why is this a separate function? And why the name? `check_write_value_and_block_thread` is actually implementing the full up, `eventfd_write` just a part of it.

I'd suggest to inline `eventfd_write` here, and rename `check_write_value_and_block_thread` to `eventfd_write`.

oli-obk:
@RalfJung we raced on the review, do you have any thoughts on https://github.com/rust-lang/miri/pull/3939#discussion_r1817806077

RalfJung:
I think for now making wakeup deterministic with a queue (via VecDeque) is fine. We can leave an issue to randomize this.

Regarding the unblocking callback, it is required for correctness right now to run immediately. As long as this does not perform atomic ops (that can be observed from other threads), I think that is fine? What exact issue are you concerned about?

oli-obk:
> What exact issue are you concerned about?

well, in this case, that we perform the reads and writes of the unblocked thread immediately, but then continue on the current thread. So unblocking a thread affects the current thread's next operations (there can now be more space to write to, even tho no thread switch happened). This makes no-preemption tests harder to write, but is not a fundamental issue I guess, just weird and not behavior a real system would have

tiif:
>  that we perform the reads and writes of the unblocked thread immediately, but then continue on the current thread. 

Oh that's surprising. I always assumed the current thread should finish first before executing other newly unblocked threads. 

But I think I observed what is mentioned here before, and I was pretty confused by the execution sequence.

src/shims/unix/linux/eventfd.rs
```diff
@@ -217,8 +179,154 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+    match eventfd.counter.get().checked_add(num) {
+        Some(_new_count @ 0..=MAX_COUNTER) => {
+            // Future `read` calls will synchronize with this write, so update the FD clock.
+            ecx.release_clock(|clock| {
+                eventfd.clock.borrow_mut().join(clock);
+            });
+
+            // When this function is called, the addition is guaranteed to not exceed u64::MAX - 1.
+            let new_count = eventfd.counter.get().checked_add(num).unwrap();
```
oli-obk, line 215:
```suggestion
```

you already get the `new_count` in the match pattern above

src/shims/unix/linux/eventfd.rs
```diff
@@ -217,8 +179,154 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            ),
+        );
+    } else {
+        // Synchronize with all prior `write` calls to this FD.
+        ecx.acquire_clock(&eventfd.clock.borrow());
+
+        // Give old counter value to userspace, and set counter value to 0.
+        let counter = eventfd.counter.get();
+        // When the counter value is 0, the read either blocks, or fails with error.
+        assert_ne!(counter, 0);
```
oli-obk, line 311:
```suggestion
```

This was already handled by reading and checking in the `if` above

src/shims/unix/linux/eventfd.rs
```diff
@@ -217,8 +179,154 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    let Some(eventfd_ref) = weak_eventfd.upgrade() else {
+        throw_unsup_format!("eventfd FD got closed while blocking.")
+    };
+
+    // Since we pass the weak file description ref to the callback function, it is guaranteed to be
+    // an eventfd file description.
+    let eventfd = eventfd_ref.downcast::<Event>().unwrap();
+
+    // Block when counter == 0.
+    let counter = eventfd.counter.get();
```
oli-obk, line 280:
```suggestion
    let counter = eventfd.counter.replace(0);
```

to avoid needing to set to `0` in the `else` arm

src/shims/unix/linux/eventfd.rs
```diff
@@ -217,8 +179,154 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        ecx.acquire_clock(&eventfd.clock.borrow());
+
+        // Give old counter value to userspace, and set counter value to 0.
+        let counter = eventfd.counter.get();
+        // When the counter value is 0, the read either blocks, or fails with error.
+        assert_ne!(counter, 0);
+        ecx.write_int(counter, &buf_place)?;
+        eventfd.counter.set(0);
+
+        // When any of the event happened, we check and update the status of all supported event
```
oli-obk, line 315:
```suggestion
        // When any of the events happened, we check and update the status of all supported event
```

tests/fail-dep/libc/eventfd_block_write_twice.rs
```diff
@@ -0,0 +1,71 @@
...
+use std::thread;
+
+// Test the behaviour of a thread being blocked on an eventfd `write`, get unblocked, and then
+// get blocked again.
+
+// The expected execution is
+// 1. Thread 1 blocks.
+// 2. Thread 2 blocks.
+// 3. Thread 3 unblocks both thread 1 and thread 2.
+// 4. Either thread 1 or thread 2 writes u64::MAX.
```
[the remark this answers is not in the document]
tiif: Yes it is deterministic, it consistently blocked on thread2 when running with ``many-seeds`` . The unblock sequence is related to the unblocking order below:
```rust
let waiting_threads = std::mem::take(&mut *eventfd.blocked_write_tid.borrow_mut());
for thread_id in waiting_threads {
    ecx.unblock_thread(thread_id, BlockReason::Eventfd)?;
}
```

In the test, thread1 get blocked before thread2, so during the unblock, thread1 gets unblocked before thread2. So thread1 gets to return from ``eventfd_write`` first, while thread2 hits the blocking condition again.

tiif:
I left  FIXME for randomising unblock sequence, and slightly changed the comments for expected execution in ``eventfd_blocked_read_twice`` and ``eventfd_blocked_write_twice`` since it is deterministic.

src/shims/unix/linux/eventfd.rs
```diff
@@ -217,8 +179,151 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            ecx.release_clock(|clock| {
+                eventfd.clock.borrow_mut().join(clock);
+            });
+
+            // When this function is called, the addition is guaranteed to not exceed u64::MAX - 1.
+            eventfd.counter.set(new_count);
+
+            // When any of the event happened, we check and update the status of all supported event
+            // types for current file description.
+            ecx.check_and_update_readiness(&eventfd_ref)?;
```
RalfJung, line 219: This used to be done in both match arms. The comment even says that. Now it was moved to only run in one match arm. The comment is outdated now, and also -- why was it moved?

RalfJung: I guess it is because in the other arm, nothing actually changes, we just block?

tests/pass-dep/libc/libc-eventfd.rs
```diff
@@ -118,3 +121,117 @@ fn test_syscall() {
...
+        let res = read_bytes(fd, &mut buf);
+        // read returns number of bytes has been read, which is always 8.
+        assert_eq!(res, 8);
+        let counter = u64::from_ne_bytes(buf);
+        assert_eq!(counter, (u64::MAX - 1));
+    });
+
+    thread1.thread().unpark();
+    thread2.thread().unpark();
+    thread3.thread().unpark();
```
RalfJung, line 232: What is the reason for this parking business? That should be explained in comments.

src/shims/unix/linux/eventfd.rs
tiif: > I guess it is because in the other arm, nothing actually changes, we just block?

Yes exactly. There are four scenarios that could happen here:
1. It succeed without blocking, we do ``check_and_update_readiness``.
1. It blocked and eventually unblock because some event happened, it will hit the succeed path that contains ``check_and_update_readiness``.
2. It errors out with ``ErrorKind::WouldBlock``.
3. It blocked and never get unblocked, in that case, that's a deadlock.

tests/pass-dep/libc/libc-eventfd.rs
tiif: I used it to get the execution sequence I want under ``-Zmiri-preemption-rate=0``. It's not ideal as it is pretty hard to reason from reading the test alone (right now I verify them manually using ``Zmiri-report-progress``). I am trying to figure out a better way to write clearer blocking test case. 

In particular, I want to find a way to execute another thread *only after* a thread is blocked. But when a thread is blocked, I couldn't find a way for it to signal "hey I am blocked now, so you should start doing something".

RalfJung: We have plenty of tests that need a particular execution sequence. None of them use thread parking. So either your test is quite special, or you can achieve the same thing in a much less complicated way. I think it's the latter.

The scheduler with `-Zmiri-preemption-rate=0` is quite simple: the current thread keeps going until it blocks. Then the next thread goes -- in the order that threads were spawned in. So please rewrite these tests to not use thread parking any more.

tiif: >  Then the next thread goes -- in the order that threads were spawned in. 

This information is useful. I will take a look at the test again and open a pr to simplify the test here and https://github.com/rust-lang/miri/pull/4033.

PR #3940, opened 2024-10-04
Prefer refutable slice patterns over len check + index op

oli-obk:
Just something I noticed while reviewing other PRs

We do it for shim arguments almost everywhere, but when the size is not statically known, we didn't use the helpers as they rely on array ops. But we can avoid a len check followed by several index ops by just using a refutable slice pattern with `let else`.

The pattern is so common, it seems almost worth a dedicated macro


RalfJung:
Nice. :)
@bors r+

FrankReh:
Wow. This is a very nice pattern.

Issue #3930, opened 2024-10-01
Consistently use io error handlers in Unix shims


RalfJung:
Unix shims return `-1` and set the thread-local "last error" variable to indicate an error occurred.  With https://github.com/rust-lang/miri/pull/3923 and https://github.com/rust-lang/miri/pull/3929, Miri has some nice helpers to help with that, but they are not being used everywhere yet. So this is about refactoring the existing code to use the new helpers.

To help, start by grepping for `-1` and for `this.eval_libc("E` in the `src/shims/unix` folder. You will find a few patterns:

```rust
                    let efault = this.eval_libc("EFAULT");
                    this.set_last_error(efault)?;
                    this.write_int(-1, dest)?;
```
should become `this.set_last_error_and_return(LibcError("EFAULT"), dest)?;`.

```rust
            this.set_last_error(ErrorKind::PermissionDenied)?;
            return Ok(Scalar::from_i32(-1));
```
should become `return this.set_last_error_and_return_i32(ErrorKind::PermissionDenied);`.

Feel free to ask if you encounter other patterns you are unsure about. It's also okay for the first PR to only cover parts of this and leave the harder cases for a second round.

noahmbright:
@rustbot claim

I'll make a PR after a first pass. I can add a second commit or just fixup depending on how that goes. Thanks for the instructions!

PR #3941, opened 2024-10-04
Replace set_last_error with set_last_error_and_return

noahmbright:
Took care of the simple patterns. Other patterns involved setting an error and then using `write_int` or setting metadata and returning -1. Unsure if those are in the scope of this change

Looks like this has conflicts with #3779, so I can update when how to handle that is decided. 

Part of https://github.com/rust-lang/miri/issues/3930.

### DIFF
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -561,8 +561,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     /// types (like `read`, that returns an `i64`).
     fn fd_not_found<T: From<i32>>(&mut self) -> InterpResult<'tcx, T> {
         let this = self.eval_context_mut();
-        let ebadf = this.eval_libc("EBADF");
-        this.set_last_error(ebadf)?;
+        this.set_last_error(LibcError("EBADF"))?;
         interp_ok((-1).into())
     }
 
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -528,8 +528,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             let o_tmpfile = this.eval_libc_i32("O_TMPFILE");
             if flag & o_tmpfile == o_tmpfile {
                 // if the flag contains `O_TMPFILE` then we return a graceful error
-                this.set_last_error(LibcError("EOPNOTSUPP"))?;
-                return interp_ok(Scalar::from_i32(-1));
+                return this.set_last_error_and_return_i32(LibcError("EOPNOTSUPP"));
             }
         }
 
@@ -548,9 +547,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // O_NOFOLLOW only fails when the trailing component is a symlink;
                 // the entire rest of the path can still contain symlinks.
                 if path.is_symlink() {
-                    let eloop = this.eval_libc("ELOOP");
-                    this.set_last_error(eloop)?;
-                    return interp_ok(Scalar::from_i32(-1));
+                    return this.set_last_error_and_return_i32(LibcError("ELOOP"));
                 }
             }
             mirror |= o_nofollow;
@@ -565,8 +562,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // Reject if isolation is enabled.
         if let IsolatedOp::Reject(reject_with) = this.machine.isolated_op {
             this.reject_in_isolation("`open`", reject_with)?;
-            this.set_last_error(ErrorKind::PermissionDenied)?;
-            return interp_ok(Scalar::from_i32(-1));
+            return this.set_last_error_and_return_i32(ErrorKind::PermissionDenied);
         }
 
         let fd = options
@@ -584,8 +580,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let seek_from = if whence == this.eval_libc_i32("SEEK_SET") {
             if offset < 0 {
                 // Negative offsets return `EINVAL`.
-                this.set_last_error(LibcError("EINVAL"))?;
-                return interp_ok(Scalar::from_i64(-1));
+                return this.set_last_error_and_return_i64(LibcError("EINVAL"));
             } else {
                 SeekFrom::Start(u64::try_from(offset).unwrap())
             }
@@ -594,8 +589,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         } else if whence == this.eval_libc_i32("SEEK_END") {
             SeekFrom::End(i64::try_from(offset).unwrap())
         } else {
-            this.set_last_error(LibcError("EINVAL"))?;
-            return interp_ok(Scalar::from_i64(-1));
+            return this.set_last_error_and_return_i64(LibcError("EINVAL"));
         };
 
         let communicate = this.machine.communicate();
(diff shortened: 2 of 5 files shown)


src/shims/unix/fs.rs
oli-obk, line 1: Where did the stray formatting changes come from?

src/shims/unix/fd.rs
```diff
@@ -554,8 +554,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     fn fd_not_found<T: From<i32>>(&mut self) -> InterpResult<'tcx, T> {
         let this = self.eval_context_mut();
         let ebadf = this.eval_libc("EBADF");
-        this.set_last_error(ebadf)?;
-        Ok((-1).into())
+        return this.set_last_error_and_return_i32(ebadf);
```
oli-obk, line 557: Use `LibcError("EBADF")`

oli-obk: Also in other places

oli-obk (changes requested):
Please revert the formatting changes unless they are from `./miri fmt` (you may need to configure your vscode correctly with the settings from `CONTRIBUTING.md`)

Also use `LibcError("foobar")` everywhere while you're doing this change.

oli-obk:
Ignore my PR, i'll rebase at some point

src/shims/unix/fs.rs
```diff
@@ -600,8 +594,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         } else if whence == this.eval_libc_i32("SEEK_END") {
             SeekFrom::End(i64::try_from(offset).unwrap())
         } else {
-            this.set_last_error(LibcError("EINVAL"))?;
-            return Ok(Scalar::from_i64(-1));
```
RalfJung, line 604: This returns an `i64`, not an `i32`, so your patch introduces an ICE here! So please add a new `set_last_error_and_return_i64` instead and use that here, or leave this to a future PR.

Please make sure to always maintain the original type, it's quite important to not screw this up and because this is "dynamic" (it's types of the interpreted program, not of the interpreter), even `./miri check` will not catch this.

src/shims/unix/fs.rs
```diff
@@ -590,8 +585,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let seek_from = if whence == this.eval_libc_i32("SEEK_SET") {
             if offset < 0 {
                 // Negative offsets return `EINVAL`.
-                this.set_last_error(LibcError("EINVAL"))?;
-                return Ok(Scalar::from_i64(-1));
```
RalfJung, line 594: More incorrect `i64` -> `i32` changes

src/shims/unix/fs.rs
```diff
@@ -1026,8 +1013,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         if let IsolatedOp::Reject(reject_with) = this.machine.isolated_op {
             this.reject_in_isolation("`opendir`", reject_with)?;
             let eacc = this.eval_libc("EACCES");
-            this.set_last_error(eacc)?;
-            return Ok(Scalar::null_ptr(this));
```
RalfJung, line 1030: This returned a null pointer before and now you made it return -1. That's definitely wrong, please keep the old code here.

src/shims/unix/fs.rs
```diff
@@ -1441,8 +1419,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         if let IsolatedOp::Reject(reject_with) = this.machine.isolated_op {
             this.reject_in_isolation("`readlink`", reject_with)?;
             let eacc = this.eval_libc("EACCES");
-            this.set_last_error(eacc)?;
-            return Ok(-1);
+            return Ok(this.set_last_error_and_return_i32(eacc));
```
tiif, line 1422: It seems like previously it returns ``Ok(-1)`` but now it returns ``Ok(Ok(-1))``?

RalfJung:
Not sure why the comments above have all been marked as resolved -- in the PR they are definitely still unresolved.

@rustbot author

noahmbright:
> Not sure why the comments above have all been marked as resolved -- in the PR they are definitely still unresolved.

Yes, sorry for the wait. How do you want to handle the `InterpResult<'tcx, i64>` vs . `InterpResult<'tcx, Scalar>` issue in my comment above? Once I fix that I'll push the update.

RalfJung:
Which comment?

src/shims/unix/fs.rs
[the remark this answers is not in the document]
RalfJung: For now just change this to `this.set_last_error(LibcError("EACCES"))?;` but leave the `return` alone.

noahmbright:
Again sorry about the wait and the careless mistakes, but I added two commits, one a small formatting fix, and in the other I added `set_last_error_and_return_i64` and fixed the type issues I introduced. Diff is smaller now too so hopefully this will be easier to re-review.

RalfJung:
There are unfortunately conflicts with the master branch, could you take care of them?
Then write @`rustbot ready`  to initiate the next round of review.

noahmbright:
> There are unfortunately conflicts with the master branch, could you take care of them with a rebase? Then write @`rustbot ready` to initiate the next round of review.

Of course. All the conflicts are something like this:
```
<<<<<<< unix_shims
            return this.set_last_error_and_return_i32(ErrorKind::PermissionDenied);
=======
            this.set_last_error(ErrorKind::PermissionDenied)?;
            return interp_ok(Scalar::from_i32(-1));
>>>>>>> master
```

I can wrap `interp_ok(this.set_last_error_and_return_*(e))` where these all occur, and we'll have an `interp_ok(InterpResult<'tcx, Scalar>)`. Does that sound alright for the resolution?

RalfJung:
`return this.set_last_error_and_return_i32(ErrorKind::PermissionDenied);` looks like the right resolution for that example.

noahmbright:
> `return this.set_last_error_and_return_i32(ErrorKind::PermissionDenied);` looks like the right resolution for that example.

I agree, after seeing the new changes while rebasing. @rustbot ready

src/shims/unix/linux/epoll.rs
```diff
@@ -375,8 +370,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             // Remove epoll_event_interest from interest_list.
             let Some(epoll_interest) = interest_list.remove(&epoll_key) else {
                 let enoent = this.eval_libc("ENOENT");
```
RalfJung, line 372: This could use `LibcError`

RalfJung (commented):
Thanks, this all looks correct but there is one spot where we can make it a bit more compact. :)

Please also squash the commits (with `git rebase --keep-base`).

noahmbright:
> Thanks, this all looks correct but there is one spot where we can make it a bit more compact. :)

Good catch. Thank you for the close reviews, this should have been much simpler. If I'm still welcome, I'll make sure my next PR goes more smoothly

RalfJung:
Looks good, thanks for the PR!
@bors r+

> Thank you for the close reviews, this should have been much simpler. If I'm still welcome, I'll make sure my next PR goes more smoothly

Sure thing, hickups happen. :) Just be careful and ask us rather than making guesses if you're not sure about something. :D When submitting a PR, it is a very good idea to list which things in the PR you are least sure about -- that greatly helps focus the review effort.

RalfJung:
That's odd... @rust-lang/infra bors has a problem if we leave a "changes required" review, and then the changes are done? CI is green but the PR has not been pushed to master.

RalfJung (approved):
Let's see if this helps github.

Kobzol:
Wtf, I never saw this error before.. Anyway, what do you think about switching miri from bors to merge queues? xD

RalfJung:
> Wtf, I never saw this error before.. Anyway, what do you think about switching miri from bors to merge queues? xD

Yeah someone just needs to put in the work for that and I don't currently have the time.^^

RalfJung:
@oli-obk I think we need to you to send an "approving" review or so, otherwise this PR can't land...

Is that really how github reviews work? Nobody else can approve a PR after a team member left a "changes required" review? That's a complete disaster.

oli-obk:
you should be able to dismiss my review in the github ui (i approved it now, but for the future)

@bors r+

RalfJung:
Where is the "dismiss review" button?

RalfJung:
Ah I think I found it. That's quite well-hidden...

PR #3946, opened 2024-10-06
Fix over synchronization of epoll

### LINKED ISSUE #3944: epoll-wait induces too much synchronization
Let's consider the following testcase:
- We have two pipes, A and B, and some global `static mut S`
- The main thead sets up an epoll for both of these pipes' read ends
- We spawn a sub-thread that writes to pipe A, then mutates `S`, then writes to pipe B
- The main thread yields, and we have `miri-preemption-rate=0`, so this will fully run the sub-thread
- The main thread waits on both pipes with epoll, but only wants to receive a single event
- The main thread asserts that it was informed about pipe A, not B
- Then the main thread reads from `S`

This should report UB due to a data race.

However, I think what will happen here is that  the main thread will receive the notification for pipe A, but because we call `acquire_clock` on the `ready_list.clock`, we acquire the clock for *all* events. This means we miss the UB as the main thread read of `S` seems to happen-after the sub-thread write of `S`.

FrankReh:
Fixes #3944.

The clock used by epoll is now per event generated, rather than by the `epoll's` ready_list.

The same epoll tests that existed before are unchanged and still pass. Also the `tokio` test case we had worked on last week still passes with this change.

This change does beg the question of how the epoll event states should change. Perhaps rather than expose public crate bool fields, so setters should be provided that include a clock parameter or an optional clock parameter. Also should all the epoll event possibilities have their clock sync tested the way these commit lay out testing. In this first go around, only the pipe's EPOLLIN is tested. The EPOLLOUT might deserve testing too, as would the eventfd. Any future source of epoll events would also fit into that category.

### DIFF
--- a/src/shims/unix/linux/epoll.rs
+++ b/src/shims/unix/linux/epoll.rs
@@ -32,11 +32,13 @@ pub struct EpollEventInstance {
     events: u32,
     /// Original data retrieved from `epoll_event` during `epoll_ctl`.
     data: u64,
+    /// The release clock associated with this event.
+    clock: VClock,
 }
 
 impl EpollEventInstance {
     pub fn new(events: u32, data: u64) -> EpollEventInstance {
-        EpollEventInstance { events, data }
+        EpollEventInstance { events, data, clock: Default::default() }
     }
 }
 
@@ -92,7 +94,6 @@ pub struct EpollReadyEvents {
 #[derive(Debug, Default)]
 struct ReadyList {
     mapping: RefCell<BTreeMap<(FdId, i32), EpollEventInstance>>,
-    clock: RefCell<VClock>,
 }
 
 impl EpollReadyEvents {
@@ -567,11 +568,6 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
                         let epoll = epfd.downcast::<Epoll>().unwrap();
 
-                        // Synchronize running thread to the epoll ready list.
-                        this.release_clock(|clock| {
-                            epoll.ready_list.clock.borrow_mut().join(clock);
-                        });
-
                         if let Some(thread_id) = epoll.thread_id.borrow_mut().pop() {
                             waiter.push(thread_id);
                         };
@@ -627,7 +623,11 @@ fn check_and_update_one_event_interest<'tcx>(
     if flags != 0 {
         let epoll_key = (id, epoll_event_interest.fd_num);
         let ready_list = &mut epoll_event_interest.ready_list.mapping.borrow_mut();
-        let event_instance = EpollEventInstance::new(flags, epoll_event_interest.data);
+        let mut event_instance = EpollEventInstance::new(flags, epoll_event_interest.data);
+        // If we are tracking data races, remember the current clock so we can sync with it later.
+        ecx.release_clock(|clock| {
+            event_instance.clock.join(clock);
+        });
         // Triggers the notification by inserting it to the ready list.
         ready_list.insert(epoll_key, event_instance);
         interp_ok(true)
@@ -654,9 +654,6 @@ fn blocking_epoll_callback<'tcx>(
 
     let ready_list = epoll_file_description.get_ready_list();
 
-    // Synchronize waking thread from the epoll ready list.
-    ecx.acquire_clock(&ready_list.clock.borrow());
-
     let mut ready_list = ready_list.mapping.borrow_mut();
     let mut num_of_events: i32 = 0;
     let mut array_iter = ecx.project_array_fields(events)?;
@@ -670,6 +667,9 @@ fn blocking_epoll_callback<'tcx>(
                 ],
                 &des.1,
             )?;
+            // Synchronize waking thread with the event of interest.
+            ecx.acquire_clock(&epoll_event_instance.clock);
+
             num_of_events = num_of_events.strict_add(1);
         } else {
             break;
--- a/tests/fail-dep/libc/libc-epoll-data-race.rs
+++ b/tests/fail-dep/libc/libc-epoll-data-race.rs
@@ -1,5 +1,9 @@
+//! This ensures that when an epoll_wait wakes up and there are multiple events,
+//! and we only read one of them, we do not synchronize with the other events
+//! and therefore still report a data race for things that need to see the second event
+//! to be considered synchronized.
 //@only-target: linux
-// ensure single way to order the thread tests
+// ensure deterministic schedule
 //@compile-flags: -Zmiri-preemption-rate=0
 
 use std::convert::TryInto;
@@ -30,7 +34,7 @@ fn check_epoll_wait<const N: usize>(epfd: i32, expected_notifications: &[(u32, u
     }
 }
 
-fn common_setup() -> (i32, [i32; 2], [i32; 2]) {
+fn main() {
     // Create an epoll instance.
     let epfd = unsafe { libc::epoll_create1(0) };
     assert_ne!(epfd, -1);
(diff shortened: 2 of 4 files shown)


FrankReh:
This left as a draft for at least two reasons. There is an open question why the fix doesn't work for the test case as it was outlined in the issue. That open question is explained in the test cases.

Then there are some field names that can be simplified, to conform to existing names, but while working on this file, I prefer having field names that are unique so it is easy to move around the file based on where those fields are used.

I guess a third reason. Before final review, the open test question will have been resolved and the tests will likely be reduced.

src/shims/unix/unnamed_socket.rs
```diff
@@ -68,8 +68,10 @@ impl FileDescription for AnonSocket {
 
         // Check if it is readable.
         if let Some(readbuf) = &self.readbuf {
-            if !readbuf.borrow().buf.is_empty() {
+            let buffer = readbuf.borrow();
+            if !buffer.buf.is_empty() {
                 epoll_ready_events.epollin = true;
+                epoll_ready_events.ready_events_clock.join(&buffer.clock)
```
tiif, line 74: Just curious, why ``join`` is used here instead of ``release_clock``?

FrankReh: I believe because the clock is being merged. There is no happens-before at this stage. Another way to view it is neither of the two objects is a thread. Just the glue to get the clock from the source thread to the destination thread.

FrankReh:
I wonder if the problem I’m seeing is cause by my assumption that the source thread should have its clock incremented just once. The one time is to be part of the happens-before relationship between this thread and any thread receiving the event. But to create a happens after clock, the clock could simply be incremented one more time as part of the cleanup to creating one or more epoll events. Then my tests would pass as expected and the question I had is solved.

FrankReh:
I have found a way to fix the problem I was seeing. It involves incrementing the clock twice when the pipe's buffer is being written to. Once to prepare the clock for the happens-before relationship. And once again to represent a happens-after relationship for writes by the same thread that happen after this write. If the write to the global static also updated the clock on its own, this probably wouldn't be necessary.

This begs new questions.

Should the pipe's read side also be epoll-able? Right now there is no such event created.

And does the eventfd, which also supports epoll, suffer from this problem of syncing clocks through epoll but then not advancing the clock, thereby keeping UB with static global writes afterwards from being detected?

RalfJung:
> Should the pipe's read side also be epoll-able? Right now there is no such event created.

Both ends are already epoll-able, I think.

src/shims/unix/unnamed_socket.rs
```diff
@@ -237,6 +241,7 @@ impl FileDescription for AnonSocket {
         if let Some(clock) = &ecx.release_clock() {
             writebuf.clock.join(clock);
         }
+        let _ = ecx.release_clock().is_some(); // Increment clock again, to establish a happens-after timeframe
```
RalfJung, line 244: Something seems quite wrong if you create a clock just to throw it away then.

And I still haven't understood what problem you are even trying to fix. Do you have a testcase demonstrating the issue?

src/shims/unix/unnamed_socket.rs
RalfJung, line 1: I wouldn't expect *any* code change in this file at all, TBH. I would expect the logic in epoll.rs to do a `release_clock()` when adding an event to the ready list, and that's it. Why is the diff here so big?

src/shims/unix/unnamed_socket.rs
FrankReh: That's a good question to resolve first. What's the other problem I'm trying to fix? I've created #3947 for that.

FrankReh: > Do you have a testcase demonstrating the issue?

Yes, `test_issue_3944_with_just_one_event`. But what it demonstrated before and then after the fix is now secondary to the discussion I hope we can have in #3947.

RalfJung:
Let's fix https://github.com/rust-lang/miri/issues/3947 (in a different PR) before attempting this.

FrankReh:
The clock field could also be an Option<VClock>. Maybe less work when clocks aren't enabled.

RalfJung:
Default clocks are trivial, so no that shouldn't be necessary, and I don't think it helps the code either.

RalfJung:
Not sure what the status here is, so
@rustbot author

FrankReh:
> Not sure what the status here is, so
> @rustbot author

I'm not familiar with this step. Am I the author being asked for the status or is this for a maintainer?
It's marked ready for review. It has a test for before and a test for after showing the improvement. Is there something else for me to do?

RalfJung:
Sorry, I should have explained - If it is ready for review, you are expected to do
@rustbot ready

src/shims/unix/linux/epoll.rs
```diff
@@ -32,11 +32,13 @@ pub struct EpollEventInstance {
     events: u32,
     /// Original data retrieved from `epoll_event` during `epoll_ctl`.
     data: u64,
+
```
RalfJung, line 35:
```suggestion
    /// The release clock associated with this event.
```

src/shims/unix/linux/epoll.rs
```diff
@@ -627,7 +623,10 @@ fn check_and_update_one_event_interest<'tcx>(
     if flags != 0 {
         let epoll_key = (id, epoll_event_interest.fd_num);
         let ready_list = &mut epoll_event_interest.ready_list.mapping.borrow_mut();
-        let event_instance = EpollEventInstance::new(flags, epoll_event_interest.data);
+        let mut event_instance = EpollEventInstance::new(flags, epoll_event_interest.data);
+        ecx.release_clock(|clock| {
```
RalfJung, line 627:
```suggestion
        // If we are tracking data races, remember the current clock so we can sync with it later.
        ecx.release_clock(|clock| {
```

tests/fail-dep/libc/libc-epoll-clock-sync.rs
```diff
@@ -0,0 +1,102 @@
...
+    check_epoll_wait::<1>(epfd, &[(expected_event, expected_value)]);
+
+    #[allow(static_mut_refs)]
+    unsafe {
+        assert_eq!(VAL_ONE, 41) // This one is not UB
+    };
+    #[allow(static_mut_refs)]
+    unsafe {
+        assert_eq!(VAL_TWO, 51) //~ERROR: Data race detected
+    };
```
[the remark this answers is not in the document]
FrankReh: I like this new comment a lot.

tests/fail-dep/libc/libc-epoll-clock-sync.rs
```diff
@@ -0,0 +1,102 @@
...
+    let slice = unsafe { std::slice::from_raw_parts(array_ptr, res.try_into().unwrap()) };
+    for (return_event, expected_event) in slice.iter().zip(expected_notifications.iter()) {
+        let event = return_event.events;
+        let data = return_event.u64;
+        assert_eq!(event, expected_event.0, "got wrong events");
+        assert_eq!(data, expected_event.1, "got wrong data");
+    }
+}
+
+fn common_setup() -> (i32, [i32; 2], [i32; 2]) {
```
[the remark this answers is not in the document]
FrankReh: I've removed the function. It's just like the original it was copied from now.

tests/fail-dep/libc/libc-epoll-clock-sync.rs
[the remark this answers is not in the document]
FrankReh: Done.

tests/fail-dep/libc/libc-epoll-data-race.rs
```diff
@@ -0,0 +1,100 @@
...
+        let event = return_event.events;
+        let data = return_event.u64;
+        assert_eq!(event, expected_event.0, "got wrong events");
+        assert_eq!(data, expected_event.1, "got wrong data");
+    }
+}
+
+// Test that the clock sync that happens through an epoll_wait only synchronizes with the clock(s)
+// that were reported. It is possible more events had become ready but the epoll_wait didn't
+// provide room for them all.
```
RalfJung, line 39:
```suggestion
```
This is already explained at the top of the file, isn't it? No need to say the same thing twice.

src/shims/unix/linux/epoll.rs
```diff
@@ -670,6 +667,9 @@ fn blocking_epoll_callback<'tcx>(
                 ],
                 &des.1,
             )?;
+            // Synchronize waking thread from the event of interest.
```
RalfJung, line 670:
```suggestion
            // Synchronize waking thread with the event of interest.
```

RalfJung:
Looks great, thanks!

@bors r+

PR #3948, opened 2024-10-07
Run all pass-dep tests natively, too

### LINKED ISSUE #3810: Run pass tests against real rustc as well
Some of our pass-dep tests currently do not pass when they are run against real rustc, e.g. `tests/pass-dep/libc/libc-epoll.rs`. That's quite strange, but I guess it's not surprising given that we are not even running these tests against real rustc in CI.

oli-obk:
fixes #3810


tests/pass/intptrcast.rs
```diff
@@ -1,4 +1,5 @@
 //@compile-flags: -Zmiri-permissive-provenance
+//@only-miri: FIXME: CI found a miscompile: `assert_eq!(2, 2)` failed in `should_be_ub`
```
oli-obk, line 2: 😆 apparently LLVM thinks it's UB indeed, just as the comment on the test function suggests

oli-obk:
> unknown unstable option: `miri-provenance-gc`

lol oops, dependency building on linux does different stuff I guess

RalfJung:
Yeah we set that flag here

https://github.com/rust-lang/miri/blob/6e83c993e36e916ab3a58712d45313a7a2d4cfe2/ci/ci.sh#L53

GC_STRESS is disabled on Windows since the runner is so slow.^^

tests/pass/intptrcast.rs
RalfJung: Hehe, nice :D

tests/ui.rs
```diff
@@ -98,8 +114,89 @@ fn miri_config(target: &str, path: &str, mode: Mode, with_dependencies: bool) ->
         ..Config::rustc(path)
     };
 
+    if let Mode::Pass { rustc: true } = mode {
+        config.comment_defaults.base().add_custom("run", Run {
+            exit_code: 0,
+            output_conflict_handling: Some(|path, actual, errors, config| {
+                let path = path.with_file_name(
+                    path.file_name().unwrap().to_str().unwrap().replace(".run.", "."),
```
RalfJung, line 122: Where does this `.run.` come from?

tests/ui.rs
```diff
@@ -98,8 +114,89 @@ fn miri_config(target: &str, path: &str, mode: Mode, with_dependencies: bool) ->
...
+            }
+
+            fn test_condition(
+                &self,
+                _config: &Config,
+                comments: &Comments,
+                revision: &str,
+            ) -> bool {
+                for rev in comments.for_revision(revision) {
+                    for arg in &rev.compile_flags {
```
RalfJung, line 172: Shouldn't we just strip everything that starts with `-Zmiri`?

tests/ui.rs
```diff
@@ -98,8 +114,89 @@ fn miri_config(target: &str, path: &str, mode: Mode, with_dependencies: bool) ->
...
+                let path = path.with_file_name(
+                    path.file_name().unwrap().to_str().unwrap().replace(".run.", "."),
+                );
+                // Blessing is only allowed in miri mode, rustc mode must match
+                ui_test::error_on_output_conflict(&path, actual, errors, config);
+            }),
+        });
+        config.program.envs.push(("MIRI_BE_RUSTC".into(), Some("host".into())));
+
+        #[derive(Debug)]
```
RalfJung, line 130:
```suggestion
        /// A ui_test filter that strips `-Zmiri` flags.
        #[derive(Debug)]
```

tests/ui.rs
```diff
@@ -145,23 +248,56 @@ fn run_tests(
...
+
     let mut config = miri_config(target, path, mode, with_dependencies);
+    config.with_args(&args);
+    if let Mode::Pass { rustc: true } = mode {
+        config.fill_host_and_target()?;
+        // Rustc mode only works on the host
+        if !config.host_matches_target() {
+            return Ok(());
+        }
+        config.output_conflict_handling = ui_test::ignore_output_conflict;
```
RalfJung, line 265: Why do we want to ignore output conflicts when building with rustc?

tests/pass-dep/libc/libc-mem.rs
```diff
@@ -1,3 +1,4 @@
+//@only-miri: testing specific alignment shenanigans
```
RalfJung, line 1: Seems unfortunate to skip the *entire* test on the host. Can you instead of `cfg(miri)` to disable just the offending test?

tests/pass/address-reuse.rs
```diff
@@ -1,5 +1,6 @@
 //! Check that we do sometimes reuse addresses.
 use std::collections::HashSet;
+//@only-miri: global allocators behave differently
```
RalfJung, line 3:
```suggestion
//@only-miri: reuse is rare on the real allocator
```

tests/pass/adjacent-allocs.rs
```diff
@@ -1,4 +1,5 @@
 //@compile-flags: -Zmiri-permissive-provenance
+//@only-miri: CI found nonadjacent stack variables
```
RalfJung, line 2:
```suggestion
//@only-miri: stack variables might never be adjacent in real executions
```

tests/pass/const-addrs.rs
```diff
@@ -8,6 +8,7 @@
 // stack frame, breaking this test.
 //@compile-flags: -Zinline-mir=no
 #![feature(strict_provenance)]
+//@only-miri: expecting that pointer addresses have specific behaviour
```
RalfJung, line 11:
```suggestion
//@only-miri: expecting that pointer addresses have specific behavior
#![feature(strict_provenance)]
```
Please generally keep the `//@` flags together

tests/pass/extern_types.rs
```diff
@@ -1,5 +1,6 @@
 //@revisions: stack tree
 //@[tree]compile-flags: -Zmiri-tree-borrows
+//@[stack]only-miri: miri specific warning
```
RalfJung, line 3:
```suggestion
//@only-miri: miri specific warning
```

tests/pass/float_nan.rs
```diff
@@ -1,3 +1,5 @@
+//@only-miri: FIXME: did not get value that should be possible: 0x7fc00000 (NaN: Pos, Quiet, payload = 0x0)
```
RalfJung, line 1:
```suggestion
//@only-miri: real executions don't actively randomize NaNs
```

tests/pass/miri_start.rs
```diff
@@ -1,4 +1,5 @@
 //@compile-flags: -Cpanic=abort
+//@only-miri: using miri internal apis
```
RalfJung, line 2:
```suggestion
//@only-miri: using internal APIs
```
Happens a few more times -- please use a consistent string here. Maybe `using Miri APIs` would be better, actually?

tests/pass/ptr_int_casts.rs
```diff
@@ -2,6 +2,7 @@
 // Tree Borrows doesn't support int2ptr casts, but let's make sure we don't immediately crash either.
 //@[tree]compile-flags: -Zmiri-tree-borrows
 //@[stack]compile-flags: -Zmiri-permissive-provenance
+//@[tree]only-miri: miri specific warnings
```
RalfJung, line 5:
```suggestion
//@only-miri: miri specific warnings
```

tests/pass/ptr_int_from_exposed.rs
```diff
@@ -2,6 +2,7 @@
 // Tree Borrows doesn't support int2ptr casts, but let's make sure we don't immediately crash either.
 //@[tree]compile-flags: -Zmiri-tree-borrows
 //@[stack]compile-flags: -Zmiri-permissive-provenance
+//@[tree]only-miri: miri specific warnings
```
RalfJung, line 5:
```suggestion
//@only-miri: miri specific warnings
```
Let's just disable these entirely, the other aliasing model might start warning any time.

tests/pass/tls/win_tls_callback.rs
```diff
@@ -5,6 +5,8 @@
 // that's not what this test is about.
 //@compile-flags: -Zmiri-ignore-leaks
 
+//@only-miri: FIXME: fatal runtime error: thread::set_current should only be called once per thread
```
RalfJung, line 8: Oh lol.^^

We probably can't use `eprintln!` so early during the setup. So maybe directly use the underlying system functions instead?

tests/pass/zero-sized-accesses-and-offsets.rs
```diff
@@ -1,6 +1,8 @@
 //! Tests specific for <https://github.com/rust-lang/rust/issues/117945>: zero-sized operations.
 #![feature(strict_provenance)]
 
+//@only-miri: FIXME: copy_nonoverlapping assertion triggered
```
RalfJung, line 4:
```suggestion
//@only-miri: FIXME: copy_nonoverlapping assertion triggered (https://github.com/rust-lang/rust/issues/131371)
```

RalfJung (commented):
This looks great!

I wonder if it'd be worth to first only land this for `pass-dep`, also to keep the diff smaller. Then we can iron out the kinks for the large number of `pass` tests in a next step.

tests/ui.rs
oli-obk: This is so you get `foo.run.stderr` that does not conflict with `foo.stderr`, which is for stderr during compilation

tests/ui.rs
oli-obk: This is filtering out tests. Otherwise I need to add `only-miri` a lot

tests/ui.rs
```diff
@@ -98,8 +114,89 @@ fn miri_config(target: &str, path: &str, mode: Mode, with_dependencies: bool) ->
...
+                    match v {
+                        Some(v) => c.env(k, v),
+                        None => c.env_remove(k),
+                    };
+                }
+                if let Some(dir) = cmd.get_current_dir() {
+                    c.current_dir(dir);
+                }
+                c.args(
+                    cmd.get_args().filter(|arg| !arg.as_encoded_bytes().starts_with(b"-Zmiri-")),
```
oli-obk, line 159: This is the arg filter

tests/pass/extern_types.rs
oli-obk: But the test works, it just differs with the diagnostics

tests/pass/intptrcast.rs
oli-obk: Ah should use cfg(miri) here, too

tests/ui.rs
oli-obk: This is compilation output. We never have compilation output from miri, because that normally means the test can't run at all. I have not seen a single example in the run test suites that show warnings except for miri specific warnings

oli-obk: I could force the output to be empty

tests/ui.rs
RalfJung: But this *removes* `.run.`

Is this to make sure that the output of the "run" is compared with the normal reference file for Miri?

tests/ui.rs
RalfJung: I guess the comment is wrong then. Please replace it by something that is correct :)

tests/ui.rs
RalfJung: More comments would help :)

tests/ui.rs
```diff
@@ -98,8 +114,89 @@ fn miri_config(target: &str, path: &str, mode: Mode, with_dependencies: bool) ->
         ..Config::rustc(path)
     };
 
+    if let Mode::Pass { rustc: true } = mode {
+        config.comment_defaults.base().add_custom("run", Run {
+            exit_code: 0,
+            output_conflict_handling: Some(|path, actual, errors, config| {
+                let path = path.with_file_name(
```
RalfJung, line 121:
```suggestion
                // ui_test would compare the output with `test.run.stdout`, but we want to
                // compare with `test.stdout`, so adjust the filename.
                let path = path.with_file_name(
```
At least I think that's what happens?

tests/ui.rs
RalfJung, line 265:
```suggestion
        // This is about the output during compilation. We want to ignore that.
        // Above in `add_custom("run", ...)`, we set things up so that the output
        // from *running* the program is compared as it should.
        config.output_conflict_handling = ui_test::ignore_output_conflict;
```

tests/ui.rs
oli-obk: Yes

tests/ui.rs
oli-obk: Not sure what that changes. I would still have to do all the same changes in ui.rs

I will do that to reduce the scope of this PR and future unrelated PRs that may be affected, but the impl is just as complex

RalfJung: Well the logic for implicitly making tests `only-miri` when they use these particular flag could be removed, I think?

tests/ui.rs
```diff
@@ -312,21 +440,30 @@ fn main() -> Result<()> {
         }
     }
 
-    ui(Mode::Pass, "tests/pass", &target, WithoutDependencies, tmpdir.path())?;
-    ui(Mode::Pass, "tests/pass-dep", &target, WithDependencies, tmpdir.path())?;
+    ui(Mode::Pass { rustc: false }, "tests/pass", &target, WithoutDependencies, tmpdir.path())?;
```
RalfJung, line 443: Can you also adjust the `eprintln` in `fn ui` to add something like "on rustc" or so, so that one can see which way the tests are being run?

RalfJung:

PR #3950, opened 2024-10-08
remove -Zmiri-panic-on-unsupported flag

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/3952, see that issue for discussion.


saethlin:
> nextest is the better option for "continue running other tests if one test failed"

Nextest does not and cannot support doctests, so this would mean we don't have a way to continue running if one doctest fails.

RalfJung:
Cargo/rustdoc already supports continuing if one doctest fails, I think?

It also runs all doctests concurrently. That's why in miri-test-libstd, doctests are so much faster than unit/integration tests. ;)

saethlin:
Ah of course it does, this is probably the same reason to get a single-threaded doctest run you need `-j1` plus `--test-threads=1`.

saethlin:
So if all use of this flag can be replaced by nextest, why should we keep it at all? I don't *think* I've seen this actually happen, but injecting new unwinding sites into a bunch of unsafe code may cause UB that's only there because of unwinding that can only happen with Miri.

RalfJung:
All right, I now made this just entirely remove the flag. @rust-lang/miri any objections?

saethlin:
No objections.

PR #3953, opened 2024-10-08
Fixed pthread_getname_np impl for glibc

YohDeadfall:
The behavior of `glibc` differs a bit different for `pthread_getname_np` from other implementations. It requires the buffer to be at least 16 bytes wide without exception.

[Docs](https://www.man7.org/linux/man-pages/man3/pthread_setname_np.3.html):

```
The pthread_getname_np() function can be used to retrieve the
name of the thread.  The thread argument specifies the thread
whose name is to be retrieved.  The buffer name is used to return
the thread name; size specifies the number of bytes available in
name.  The buffer specified by name should be at least 16
characters in length.  The returned thread name in the output
buffer will be null terminated.
```

[Source](https://sourceware.org/git/?p=glibc.git;a=blob;f=nptl/pthread_getname.c;hb=dff8da6b3e89b986bb7f6b1ec18cf65d5972e307#l37):

```c
int
__pthread_getname_np (pthread_t th, char *buf, size_t len)
{
  const struct pthread *pd = (const struct pthread *) th;

  /* Unfortunately the kernel headers do not export the TASK_COMM_LEN
     macro.  So we have to define it here.  */
#define TASK_COMM_LEN 16
  if (len < TASK_COMM_LEN)
    return ERANGE;
```

### DIFF
--- a/src/shims/unix/linux/foreign_items.rs
+++ b/src/shims/unix/linux/foreign_items.rs
@@ -9,6 +9,11 @@ use crate::machine::{SIGRTMAX, SIGRTMIN};
 use crate::shims::unix::*;
 use crate::*;
 
+// The documentation of glibc complains that the kernel never exposes
+// TASK_COMM_LEN through the headers, so it's assumed to always be 16 bytes
+// long including a null terminator.
+const TASK_COMM_LEN: usize = 16;
+
 pub fn is_dyn_sym(name: &str) -> bool {
     matches!(name, "statx")
 }
@@ -74,22 +79,29 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             "pthread_setname_np" => {
                 let [thread, name] =
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
-                let max_len = 16;
                 let res = this.pthread_setname_np(
                     this.read_scalar(thread)?,
                     this.read_scalar(name)?,
-                    max_len,
+                    TASK_COMM_LEN,
                 )?;
                 this.write_scalar(res, dest)?;
             }
             "pthread_getname_np" => {
                 let [thread, name, len] =
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
-                let res = this.pthread_getname_np(
-                    this.read_scalar(thread)?,
-                    this.read_scalar(name)?,
-                    this.read_scalar(len)?,
-                )?;
+                // The function's behavior isn't portable between platforms.
+                // In case of glibc, the length of the output buffer must
+                // be not shorter than TASK_COMM_LEN.
+                let len = this.read_scalar(len)?;
+                let res = if len.to_target_usize(this)? < TASK_COMM_LEN as u64 {
+                    this.eval_libc("ERANGE")
+                } else {
+                    this.pthread_getname_np(
+                        this.read_scalar(thread)?,
+                        this.read_scalar(name)?,
+                        len,
+                    )?
+                };
                 this.write_scalar(res, dest)?;
             }
             "gettid" => {
--- a/tests/pass-dep/libc/pthread-threadname.rs
+++ b/tests/pass-dep/libc/pthread-threadname.rs
@@ -5,6 +5,9 @@ use std::ffi::CString;
 use std::thread;
 
 fn main() {
+    // The short name should be shorter than 16 bytes which POSIX promises
+    // for thread names. The length includes a null terminator.
+    let short_name = "test_named".to_owned();
     let long_name = std::iter::once("test_named_thread_truncation")
         .chain(std::iter::repeat(" yada").take(100))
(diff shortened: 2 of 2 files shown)


src/shims/unix/linux/foreign_items.rs
```diff
@@ -9,6 +9,11 @@ use crate::machine::{SIGRTMAX, SIGRTMIN};
 use crate::shims::unix::*;
 use crate::*;
 
+// The documentation of glibc complaints that the kernel never exposes
+// TASK_COMM_LEN throut the headers, so it's assumed to always be 16 bytes
+// long including a null terminator.
```
RalfJung, line 14:
```suggestion
// The documentation of glibc complains that the kernel never exposes
// TASK_COMM_LEN through the headers, so it's assumed to always be 16 bytes
// long including a null terminator.
```

src/shims/unix/linux/foreign_items.rs
```diff
@@ -74,22 +79,30 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
-                let res = this.pthread_getname_np(
-                    this.read_scalar(thread)?,
-                    this.read_scalar(name)?,
-                    this.read_scalar(len)?,
-                )?;
+                // The function's behavior isn't portable between platforms.
+                // In case of glibc, the length of the output buffer must
+                // be not shorter than TASK_COMM_LEN. That requirements
+                // comes from the fact that there's no intermediate buffer.
```
RalfJung, line 95:
```suggestion
                // be not shorter than TASK_COMM_LEN.
```

tests/pass-dep/libc/pthread-threadname.rs
```diff
@@ -55,8 +55,15 @@ fn main() {
         // But the system is limited -- make sure we successfully set a truncation.
         let mut buf = vec![0u8; long_name.len() + 1];
         assert_eq!(get_thread_name(&mut buf), 0);
+
+        // A specific case of glibc which requires the buffer.to be at least 15 bytes long
+        // excluding a null terminator;
+        #[cfg(target_os = "linux")]
+        assert_ne!(get_thread_name(&mut buf[..15]), 0);
```
RalfJung, line 62:
```suggestion
        assert_eq!(get_thread_name(&mut buf[..15]), libc::ERANGE);
```

tests/pass-dep/libc/pthread-threadname.rs
```diff
@@ -55,8 +55,15 @@ fn main() {
         // But the system is limited -- make sure we successfully set a truncation.
         let mut buf = vec![0u8; long_name.len() + 1];
         assert_eq!(get_thread_name(&mut buf), 0);
+
+        // A specific case of glibc which requires the buffer.to be at least 15 bytes long
+        // excluding a null terminator;
```
RalfJung, line 60:
```suggestion
        // glibc requires the buffer to be at least 15 bytes long
        // excluding the null terminator.
```

tests/pass-dep/libc/pthread-threadname.rs
RalfJung, line 62: This here should return ERANGE also on Solarish -- and it passes already without your change. We already return ERANGE when the thread name is longer than the buffer.

YohDeadfall: Not a commit I wanted to open a PR for. Forgot to push it.

YohDeadfall: Initially I made what you have seen and pushed it. Then checked the diff and improved the test, but forgot to force push it, so made a soft reset on top of the current changes to avoid breaking the history.

There was no test for a shorter name which is suitable to verify the `glibc` behavior.

tests/pass-dep/libc/pthread-threadname.rs
```diff
@@ -48,23 +51,63 @@ fn main() {
...
-    result.unwrap().join().unwrap();
+            let cstr = CStr::from_bytes_until_nul(&buf).unwrap();
+            // POSIX seems to promise at least 15 chars excluding a null terminator.
+            assert_eq!(short_name.as_bytes(), cstr.to_bytes());
+
+            // Also test directly calling pthread_setname to check its return value.
+            assert_eq!(set_thread_name(&cstr), 0);
+
+            // For glibc used by linux-gnu there should be a failue,
+            // if a shorter than 16 bytes buffer is provided..
```
RalfJung, line 73:
```suggestion
            // if a shorter than 16 bytes buffer is provided, even if that would be
            // large enough for the thread name.
```

RalfJung (commented):
Looks good, just a comment nit. Then please squash the commits (with `--keep-base` to keep the diff reviewable)

RalfJung:
Thanks. :-)

@bors r+

Issue #240, opened 2017-07-12
Coverage


oli-obk:
Figure out a way to run gcov on miri, so we know if we aren't testing something or if some hack becomes obsolete

cgettys-microsoft:
I think this is relatively doable nowadays, as a far bit of infrastructure in this area has stabilized:

[Instrumentation-based Code Coverage](https://doc.rust-lang.org/rustc/instrument-coverage.html#instrumentation-based-code-coverage)

The LLVM coverage data in particular is pretty nice.

Might be a good first issue now?

RalfJung:
Maybe not a good "first" issue, since I assume it will involve some fighting with the build system. But yeah seems worth promoting this a bit so maybe someone can look into it.

What is not clear at all to me is how to best get this into CI. Ideally we'd somehow know when a line that used to be covered is no longer covered, but that requires keeping state across CI runs... not sure how that would work.

Mandragorian:
I would be interested in trying this. I would need to do some research first though. I haven't done anything similar before.

Mandragorian:
I finally found some time to work on this. Just by adding `RUSTFLAGS="-C instrument-coverage"` in the `./miri test` command:

```
[mandragore@archibald miri]$ git diff
diff --git a/miri-script/src/util.rs b/miri-script/src/util.rs
index f5a6a8188..45f587dfa 100644
--- a/miri-script/src/util.rs
+++ b/miri-script/src/util.rs
@@ -188,7 +188,7 @@ impl MiriEnv {
     }
 
     pub fn test(&self, crate_dir: impl AsRef<OsStr>, args: &[String]) -> Result<()> {
-        self.cargo_cmd(crate_dir, "test").args(args).run()?;
+        self.cargo_cmd(crate_dir, "test").args(args).env("RUSTFLAGS", "-C instrument-coverage").run()?;
         Ok(())
     }
```

I was able to generate the `profraw` files. Then after install `cargo-binutils` and running:

```
[root@e4aaf2fc8c9d miri]# cargo-profdata -- merge -sparse default_*.profraw  -o cooked.profdata
[root@e4aaf2fc8c9d miri]# cargo-cov -- report --instr-profile=cooked.profdata --object target/debug/miri  -ignore-filename-regex="root/.cargo/registry/*|root/.rustup/*|/rust/deps/*|target/*"
```

(the `root` in the regex above is because I use a container to run tests)

I got this report:

```
[...]

PR #3954, opened 2024-10-09
Add option for generating coverage reports

Mandragorian:
Add a `--coverage` option in the `test` subcommand of the miri script. This option, when set, will generate a coverage report after running the tests.

`cargo-binutils` is needed as a dependency to generate the reports.


RalfJung:
Would be good to also have a smoke test in CI to at least ensure this works. That should likely go somewhere in `ci/ci.sh`.

oli-obk:
If you additionally pipe the output to `$GITHUB_STEP_SUMMARY` we can see it in CI of every run

Little effort for a minimal report shown on every run

miri-script/src/commands.rs
```diff
@@ -479,6 +482,41 @@ impl Command {
...
+        // Merge the profraw files
+        let profraw_files_cloned= profraw_files.iter();
+        cmd!(
+            e.sh,
+            "cargo-profdata -- merge -sparse {profraw_files_cloned...} -o merged.profdata"
+        ).quiet().run()?;
+
+        // Create the coverage report.
+        let home = std::env::var("HOME")?;
+        let ignored = format!("{home}/*|miri/target/*|target/debug/*|rust/deps/*");
```
saethlin, line 509: What exactly is this supposed to ignore? Or, what undesirable thing happens if we don't ignore anything?

Mandragorian: If I don't add these ignore regular expressions, the report also show files like:

- `miri/target/debug/build/chrono-tz-65fd732542149546/out/timezones.rs`
- `root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/aes-0.8.4/src/autodetect.rs `
- `rust/deps/tracing-0.1.37/src/macros.rs`

(the root path is because I run it in a container, so it is the home directory)

Additionally the coverage percentages shown in the report are smaller than the real ones, since non-miri files are included.

Ignoring these files is the simplest way I could think to fix this.

Mandragorian:
Regarding smoke tests/showing the output in the CI:

I am only worried about the run time of the tests growing. I could maybe add a filter in the `./miri test` invocation so that it only runs one test, so the smoke test doesn't take too long. However if we do this then we won't get a meaningful report in the CI. 

I am not sure how github actions work. Would creating a new action so that it can run in parallel with the other tests work?

This also makes me think that the way this is implemented, will only produce coverage reports for a single target. It wont aggregate coverage information across different targets, so we won't have a complete picture of the coverage in a single report. A possible solution to this might be to support running multiple targets with a single `./miri test` invocation, as we did with `cargo-miri`. This way the `profraw` files from all targets would me merged together. But then this might give wrong results when we see how many times a line was executed during the tests?

@RalfJung @oli-obk

miri-script/src/commands.rs
saethlin: Interesting. I suppose the report isn't supposed to look like this?
```
$ cargo-cov -- report --instr-profile=merged.profdata --object target/debug/miri -ignore-filename-regex=/home/ben/*|miri/target/*|target/debug/*|rust/deps/*
Filename                      Regions    Missed Regions     Cover   Functions  Missed Functions  Executed       Lines      Missed Lines     Cover    Branches   Missed Branches     Cover
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
TOTAL                               0                 0         -           0                 0         -           0                 0         -           0                 0         -
```
I do all development work from directories under `$HOME` which I think is pretty typical. So I think this regex is excluding all the profile data.

Is it possible for us to only include certain directories, instead of excluding them? We just want to include `src/` I think, which seems like a much more robust filter, if it's possible to implement.

RalfJung:
Regarding the multi-target situation, let's move this discussion to the issue as it is not needed for an MVP.

What is the runtime overhead of this? We can add a new dedicated coverage target if needed, though it will duplicate all the work of downloading the right rustc and building Miri. Or does it need its own separate build anyway?

RalfJung:
Yeah excluding HOME is definitely not going to work. Does excluding HOME/.cargo work?

RalfJung:
Also, why does it exclude "target/debug" on top of the others? Hard-coding "debug" is not great.

Please have a comment explaining for every exclusion why it is needed.

(Sorry I can't reply in the thread right now, only via email.)

miri-script/src/commands.rs
```diff
@@ -468,6 +468,9 @@ impl Command {
         if bless {
             e.sh.set_var("RUSTC_BLESS", "Gesundheit");
         }
+        if coverage {
+            e.sh.set_var("RUSTFLAGS", "-C instrument-coverage");
```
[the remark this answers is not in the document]
RalfJung: Also, Miri will be built twice this way: once without the flag in `e.build_miri_sysroot`, and then again with the flag in `e.test`.

So the flag should probably be set before `build_miri_sysroot`.

miri-script/src/commands.rs
Mandragorian: yeah that makes sense. Working in the container, I did my dev work under `/miri`, which was not under `$HOME` (`/root`). So I didn't think of that. Unfortunately I can't find an option to include files/directories, instead of excluding them.

We could use the [llvm_profparser](https://crates.io/crates/llvm_profparser) crate, used in tarpaulin, as i suggested in #240. It seems to provide a mapping which [can produce reports](https://docs.rs/llvm_profparser/0.7.0/llvm_profparser/coverage/coverage_mapping/struct.CoverageMapping.html#method.generate_subreport) only for the paths that we are interested in.

I went with using cargo-binutils for this PR, since it seemed like a more common crate, than tarpauling/llvm_profparser. Maybe adding llvm_profparser as a dependency for coverage reports is acceptable.

RalfJung: Might be worth a feature request for cargo-binutils?

I'd rather avoid that dependency, at least for now.

saethlin: I think `--sources src/` selects the files we want. At least the coverage report it emits looks reasonable to me?

saethlin: I don't think there's any need to use `cargo-binutils`, in this case `cargo-cov` is a trivial wrapper around `llvm-cov`.

Mandragorian: Yes, `--sources src/` actually works. I tried using it like `--sources=src/` and it didn't work for me.

I use cargo-binutils because I saw this in the instrument coverage documentation:

```
note that the coverage mapping data generated by the Rust compiler requires LLVM version 12 or higher, and processing the raw data may require exactly the LLVM version used by the compiler. (llvm-cov --version typically shows the tool's LLVM version number, and rustc --verbose --version shows the version of LLVM used by the Rust compiler.)

    You can install compatible versions of these tools via the rustup component llvm-tools-preview. This component is the recommended path, though the specific tools available and their interface is not currently subject to Rust's usual stability guarantees. In this case, you may also find cargo-binutils useful as a wrapper around these tools.
    ```

I am not really familiar with llvm ecosystem so I don't know how easy it is to get compatible versions of the llvm tools. I used `cargo-binutils` to be more portable.

Comment (saethlin) [reply miri-script/src/commands.rs:509]: As the cargo-binutils docs say, all that package does is exec the llvm tools from the sysroot, relying on the llvm-tools-preview component being installed. We're already doing toolchain management and finding stuff in the sysroot as part of miri-script. Can you reuse that code? Or make it reusable?

File: miri-script/src/commands.rs (hunk 2 above)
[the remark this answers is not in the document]
Comment (Mandragorian) [reply miri-script/src/commands.rs]: i am now appending the flag, and have moved the code before building the sysroot.

File: miri-script/src/commands.rs (hunk 1 above)
Comment (Mandragorian) [reply miri-script/src/commands.rs]: During the `toolchain` subcommand we install the `llvm-tools` component (not the `-preview` version). I copy-pasted and modified some code from cargo-binutils and it seems to work. (do we need to do anything license-wise? I don't think so but I am not sure. Both miri and cargo-binutils have an MIT license).

After doing some more search it seems that `llvm-tools` should never have existed and the "proper" name for it should have been `llvm-tools-preview`. https://github.com/rust-lang/rust/issues/119164. However after inadvertently using the name `llvm-tools` and people started depending on it it was decided not to change the name again and break people. Indeed when I add the `llvm-tools-preview` component I see:

```
$ rustup component add llvm-tools-preview
info: downloading component 'llvm-tools'
info: installing component 'llvm-tools'
```

**Comment by RalfJung** [comment]
> For the smoke test, I invoke ./miri test --coverage do_not_match_a_test at the very end of ci.sh. This way no tests are actually (re-)run, but a coverage report is still produced, and should fail the test if something fails.

This will still do a full rebuild, right, because it needs an instrumented Miri? And then it will run zero tests, so it doesn't actually check that we can invoke Miri with instrumentation and gather the results.

Note that `./miri test` runs Miri many times, so what I would expect to happen is that the profile gets accumulated from all invocations -- and `./miri test --coverage do_not_match_a_test` would show a basically empty profile since nothing runs.

I think I prefer the option where we add a new job that does the full `./miri test --coverage`.

> What I am not sure about is how I find the miri binary 

There's a `let miri_bin` in `fn run`, is that what you need?

File: miri-script/src/commands.rs [hunk 3]
```diff
@@ -479,9 +492,61 @@ impl Command {
...
+        Ok(())
+    }
+
+    fn rustlib(sysroot: String) -> Result<PathBuf> {
+        let mut pathbuf = PathBuf::from(sysroot);
+        pathbuf.push("lib");
+        pathbuf.push("rustlib");
+        pathbuf.push(rustc_version::version_meta()?.host);
+        pathbuf.push("bin");
+        Ok(pathbuf)
```
RalfJung, line 538: This seems to duplicate logic in `MiriEnv::new`. Instead, please just store the `libdir` that is computed there in a field.

miri-script/src/commands.rs
```diff
@@ -479,9 +492,61 @@ impl Command {
...
         e.test(".", &flags)?;
+
+        if coverage {
+            Self::show_coverage_report(&e)?;
+        }
         Ok(())
     }
 
+    fn show_coverage_report(e: &MiriEnv) -> Result<()> {
+        let profraw_files: Vec<_> = Self::profraw_files(".")?;
```
RalfJung, line 503: This seems a bit messy... it will just go over all files that happened to be called `.profraw`, no matter whether they are from this run or not? Isn't there a cleaner way, like tell the instrumented binary to put the output into a particular folder?

We should not be creating files just randomly in the Miri source tree, and we certainly should not *delete* files randomly from the Miri source tree. Temporary files should go in a tempdir, and then it also shouldn't be needed to change `.gitignore`.

miri-script/src/commands.rs
Mandragorian: ah I missed that! thanks!

miri-script/src/commands.rs
Mandragorian: I am using the `tempfile` crate now. I saw that it is used in `dev-dependencies` for some crates so maybe it is ok depending on it for the script? Otherwise I could use `std::env::temp_dir` and manually create a subdir there.

RalfJung: Yeah tempfile is good. It pulls in a few more dependencies, I hope those don't have significant build time.

miri-script/src/coverage.rs
```diff
@@ -0,0 +1,84 @@
...
+        Ok(Self { path: tempfile::TempDir::new()? })
+    }
+
+    /// add_env_vars will add the required environment variables to MiriEnv `e`.
+    pub fn add_env_vars(&self, e: &mut MiriEnv) -> Result<()> {
+        let rustflags = e.sh.var("RUSTFLAGS")?;
+        let rustflags = format!("{rustflags} -C instrument-coverage");
+        e.sh.set_var("RUSTFLAGS", rustflags);
+
+        let file_template = self.path.path().join("default_%m_%p.profraw");
```
RalfJung, line 32: Please explain the magic formatting symbols here. Also, why "default"?

miri-script/src/coverage.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// 
+    /// An error will be returned if a temporary directory could not be created.
+    pub fn new() -> Result<Self> {
+        Ok(Self { path: tempfile::TempDir::new()? })
+    }
+
+    /// add_env_vars will add the required environment variables to MiriEnv `e`.
+    pub fn add_env_vars(&self, e: &mut MiriEnv) -> Result<()> {
+        let rustflags = e.sh.var("RUSTFLAGS")?;
+        let rustflags = format!("{rustflags} -C instrument-coverage");
```
RalfJung, line 29:
```suggestion
        let mut rustflags = e.sh.var("RUSTFLAGS")?;
        rustflags.push_str(" -C instrument-coverage");
```

miri-script/src/coverage.rs
```diff
@@ -0,0 +1,84 @@
...
+        let profraw_files_cloned = profraw_files.iter();
+        cmd!(e.sh, "{profdata_bin} merge -sparse {profraw_files_cloned...} -o {merged_file}")
+            .quiet()
+            .run()?;
+
+        // Create the coverage report.
+        let mut cov_bin = e.libdir.clone();
+        cov_bin.push("..");
+        cov_bin.push("bin");
+        cov_bin.push("llvm-cov");
```
RalfJung, line 59: Can this be written nicer with the `path!` macro?

Mandragorian: yes, thanks!

miri-script/src/coverage.rs
Mandragorian: done. as for "default" i just copy pasted the default value for the env var.

RalfJung: Maybe we should just make it `miri_` then to make it a bit more self-explaining :shrug:

Mandragorian: done.

saethlin:
As soon as I get a coverage report, the next thing I want to know is what lines are lacking coverage. As far as I can tell, `./miri test --coverage` shuffles everything into a `TempDir`, which is recursively removed when the command exits. That's a bummer. Can we keep the merged profdata around? Perhaps just the most recent one?

`cargo-cov` stashes everything in `target/cov` which seems like a nice trick, but I don't want to scope-creep this PR. Just making it *possible* to dig deeper into the coverage data would make me happy.

Mandragorian:
For persisting the coverage data, maybe we can check if `LLVM_PROFILE_FILE` is already set, and if it is we extract the parent to use as our base dir?

RalfJung:
> cargo-cov stashes everything in target/cov which seems like a nice trick, but I don't want to scope-creep this PR. Just making it possible to dig deeper into the coverage data would make me happy.

I didn't know there is such a convention.
Does it clear previous reports?

We could make a new subdir under `target/cov` to ensure different runs don't get mixed, but that would accumulate a lot of stuff over time. Also maybe you want to mix different runs?

saethlin:
> I didn't know there is such a convention.

I don't know if I'd call it that, but at least a few tools stash artifacts in magic directories under `target/`, as far as I can tell for this reason.

I don't think the complexity of retaining previous runs is worth adding to this PR. If we're going to retain multiple runs someone needs to decide if we're going to retain coverage info indefinitely, if not how many runs to keep, and how to organize those. Just being able to call up the report of which lines weren't covered is enough to make this useable, and likely all that anyone will need. After all, `cargo build` doesn't keep the last N executables.

RalfJung:
Didn't you suggest earlier to keep the info from the runs? I am confused now what you are suggesting.

saethlin:
The implementation in this PR stores the `merged.profdata` file in a temporary directory, runs an LLVM tool to emit the per-file coverage summary, then deletes the temporary directory before the command exits. It is impossible by design to generate any output other than what is printed by `./miri test --coverage`.

I don't want the profile data from the latest runs. Just the most recent one. And I'm not really talking about custom analysis, I just want to be able to run `llvm-cov` with various other arguments.

RalfJung:
So you want the per-test files to be cleaned, but the merged file stored in `target` somewhere under a fixed name, so each run overwrites it? Sure that sounds reasonable.

saethlin:
That's one possible implementation. I'm not picky, and I think if I have a more specific opinion it will only be after using this.

Mandragorian:
Would this work?

saethlin:
Works for me!

Mandragorian:
Just a friendly ping for this. Is there anything else missing?

RalfJung:
It's in my queue, but $DAYJOB is keeping me quite busy these days.

miri-script/src/commands.rs
```diff
@@ -172,7 +172,13 @@ impl Command {
             Command::Install { flags } => Self::install(flags),
             Command::Build { flags } => Self::build(flags),
             Command::Check { flags } => Self::check(flags),
-            Command::Test { bless, flags, target } => Self::test(bless, flags, target),
+            Command::Test { bless, flags, target, coverage } =>
+                Self::test(
+                    bless,
+                    flags,
+                    target,
+                    coverage.then_some(crate::coverage::CoverageReport::new()?),
```
RalfJung, line 180: Please move this logic into the `test` function; the arguments should corresponds to the enum variant.

miri-script/src/coverage.rs
```diff
@@ -0,0 +1,90 @@
...
+use anyhow::{Context, Result};
+use path_macro::path;
+use tempfile::TempDir;
+use xshell::cmd;
+
+use crate::util::MiriEnv;
+
+/// CoverageReport can generate code coverage reports for miri.
+pub struct CoverageReport {
+    /// path is a temporary directory where coverage artifacts will be stored.
```
RalfJung, line 12:
```suggestion
    /// path is a temporary directory where intermediate coverage artifacts will be stored.
    /// (The final output will be stored in a permanent location.)
```

miri-script/src/coverage.rs
```diff
@@ -0,0 +1,90 @@
...
+    path: TempDir,
+}
+
+impl CoverageReport {
+    /// new creates a new CoverageReport
```
RalfJung, line 17:
```suggestion
    /// Creates a new CoverageReport.
```

miri-script/src/coverage.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// profraw_files returns the profraw files in `self.path`.
+    ///
+    /// # Errors
+    ///
+    /// An error will be returned if `self.path` can't be read.
+    fn profraw_files(&self) -> Result<Vec<OsString>> {
+        Ok(std::fs::read_dir(&self.path)?
+            .filter_map(|r| r.ok())
+            .map(|e| e.path())
+            .filter(|p| p.extension().map(|e| e == "profraw").unwrap_or(false))
```
RalfJung, line 86:
```suggestion
            .filter(|p| p.extension().is_some_and(|e| e == "profraw"))
```

miri-script/src/coverage.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    /// profraw_files returns the profraw files in `self.path`.
+    ///
+    /// # Errors
+    ///
+    /// An error will be returned if `self.path` can't be read.
+    fn profraw_files(&self) -> Result<Vec<OsString>> {
+        Ok(std::fs::read_dir(&self.path)?
+            .filter_map(|r| r.ok())
+            .map(|e| e.path())
```
RalfJung, line 85: Shouldn't we exclude non-files somewhere?

miri-script/src/coverage.rs
```diff
@@ -0,0 +1,90 @@
...
+
+        Ok(())
+    }
+
+    /// profraw_files returns the profraw files in `self.path`.
+    ///
+    /// # Errors
+    ///
+    /// An error will be returned if `self.path` can't be read.
+    fn profraw_files(&self) -> Result<Vec<OsString>> {
```
RalfJung, line 82: Since these are paths, why not use `PathBuf`?

miri-script/src/coverage.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// files in `self.path`.
+    pub fn show_coverage_report(&self, e: &MiriEnv) -> Result<()> {
+        let profraw_files: Vec<_> = self.profraw_files()?;
+
+        let profdata_bin = path!(e.libdir / ".." / "bin" / "llvm-profdata");
+
+        let merged_file = path!(e.miri_dir / "target" / "coverage.profdata");
+
+        // Merge the profraw files
+        let profraw_files_cloned = profraw_files.iter();
```
RalfJung, line 59: This doesn't even call `clone`, so why `_cloned`? Also `profraw_files` isn't used again afterwards so this seems unnecessary?

miri-script/src/coverage.rs
```diff
@@ -0,0 +1,90 @@
...
+        // Create the coverage report.
+        let cov_bin = path!(e.libdir / ".." / "bin" / "llvm-cov");
+        let miri_bin =
+            e.build_get_binary(".").context("failed to get filename of miri executable")?;
+        cmd!(
+            e.sh,
+            "{cov_bin} report --instr-profile={merged_file} --object {miri_bin} --sources src/"
+        )
+        .run()?;
+
```
RalfJung, line 73: Should we emit something like "the full report file can be found in <path>", so people know how to invoke further commands?

RalfJung:
Please don't squash commits while review is in progress. Thanks to github being kind of bad at dealing with force pushes, that makes it totally impossible to tell what changed since my last review.

miri-script/src/coverage.rs
```diff
@@ -0,0 +1,91 @@
...
+        // So we just use the same template, replacing "default" with "miri".
+        let file_template = self.path.path().join("miri_%m_%p.profraw");
+        e.sh.set_var("LLVM_PROFILE_FILE", file_template);
+        Ok(())
+    }
+
+    /// show_coverage_report will print coverage information using the artifact
+    /// files in `self.path`.
+    pub fn show_coverage_report(&self, e: &MiriEnv) -> Result<()> {
+        let profraw_files: Vec<_> = self.profraw_files()?;
```
RalfJung, line 53:
```suggestion
        let profraw_files = self.profraw_files()?;
```
This doesn't seem needed.

.github/workflows/ci.yml
```diff
@@ -58,11 +58,20 @@ jobs:
       - name: rustdoc
         run: RUSTDOCFLAGS="-Dwarnings" ./miri doc --document-private-items
 
+  coverage:
+    name: Coverage report
```
RalfJung, line 62:
```suggestion
    name: coverage report
```
Just to match the style of the other jobs

RalfJung (commented):
Just some nits left :)

Mandragorian:
Ah, sorry about the squashing! Didn't know that the diff would get messed up :/

@rustbot ready

RalfJung:
No worries. :)

All looking good, so *now* is the time to squash. :) (with `git rebase --keep-base`)

Mandragorian:
Hoping this worked as expected (out of curiosity, do you use the "changes from last review" in the "files changed" tab? just asking so I can understand the workflow better)

RalfJung:
Looks good, thanks!

> (out of curiosity, do you use the "changes from last review" in the "files changed" tab? just asking so I can understand the workflow better)

No, I have not found that to work reliably. After a force-push, it always says "We went looking everywhere, but couldn’t find those commits". I am looking at the commits you push and the "force-push" diffs.

PR #3956, opened 2024-10-09
epoll: rename blocking_epoll_callback since it is not just called after unblocking

RalfJung:
@tiif does `return_ready_list` seem like a reasonable name?


tiif:
Yup, this name is great.

PR #3957, opened 2024-10-09
Fixed get/set thread name implementations for macOS and FreeBSD

YohDeadfall:
So, the story of fixing `pthread_getname_np` and `pthread_setname_np` continues, but this time I fixed the macOS implementation.

### [`pthread_getname_np`](https://github.com/apple-oss-distributions/libpthread/blob/c032e0b076700a0a47db75528a282b8d3a06531a/src/pthread.c#L1160-L1175)

The function never fails except for an invalid thread. Miri never verifies thread identifiers and uses them as indices when accessing a vector of threads. Therefore, the only possible result is `0` and a possibly trimmed output.

```c
int
pthread_getname_np(pthread_t thread, char *threadname, size_t len)
{
	if (thread == pthread_self()) {
		strlcpy(threadname, thread->pthread_name, len);
		return 0;
	}


	if (!_pthread_validate_thread_and_list_lock(thread)) {
		return ESRCH;
	}


	strlcpy(threadname, thread->pthread_name, len);
	_pthread_lock_unlock(&_pthread_list_lock);
	return 0;
}
```

#### [`strcpy`](https://www.man7.org/linux/man-pages/man7/strlcpy.7.html)

```
strlcpy(3bsd)
strlcat(3bsd)
      Copy and catenate the input string into a destination
      string.  If the destination buffer, limited by its size,
      isn't large enough to hold the copy, the resulting string
      is truncated (but it is guaranteed to be null-terminated).
      They return the length of the total string they tried to
      create.
```

### [`pthread_setname_np`](https://github.com/apple-oss-distributions/libpthread/blob/c032e0b076700a0a47db75528a282b8d3a06531a/src/pthread.c#L1178-L1200)

```c
pthread_setname_np(const char *name)
{
	int res;
	pthread_t self = pthread_self();


	size_t len = 0;
	if (name != NULL) {
		len = strlen(name);
	}


	_pthread_validate_signature(self);


	res = __proc_info(5, getpid(), 2, (uint64_t)0, (void*)name, (int)len);
	if (res == 0) {
		if (len > 0) {
			strlcpy(self->pthread_name, name, MAXTHREADNAMESIZE);
		} else {
			bzero(self->pthread_name, MAXTHREADNAMESIZE);
		}
	}
	return res;


}
```

Where `5` is [`PROC_INFO_CALL_SETCONTROL`](https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/bsd/sys/proc_info_private.h#L274), and `2` is [`PROC_INFO_CALL_SETCONTROL`](https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/bsd/sys/proc_info.h#L821). And `__proc_info` is a syscall handled by the XNU kernel by [`proc_info_internal`](https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/bsd/kern/proc_info.c#L300-L314):

```c
int
proc_info_internal(int callnum, int pid, uint32_t flags, uint64_t ext_id, int flavor, uint64_t arg, user_addr_t buffer, uint32_t  buffersize, int32_t * retval)
{
	switch (callnum) {
	// ...
	case PROC_INFO_CALL_SETCONTROL:
		return proc_setcontrol(pid, flavor, arg, buffer, buffersize, retval);
```

And the actual logic from [`proc_setcontrol`](https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/bsd/kern/proc_info.c#L3218-L3227):
```c
	case PROC_SELFSET_THREADNAME: {
		/*
		 * This is a bit ugly, as it copies the name into the kernel, and then
		 * invokes bsd_setthreadname again to copy it into the uthread name
		 * buffer.  Hopefully this isn't such a hot codepath that an additional
		 * MAXTHREADNAMESIZE copy is a big issue.
		 */
		if (buffersize > (MAXTHREADNAMESIZE - 1)) {
			return ENAMETOOLONG;
		}
```

Unrelated to the current pull request, but perhaps, there's a very ugly thing in the kernel/libc because the last thing happening in `PROC_SELFSET_THREADNAME` is `bsd_setthreadname` which sets the name in the user space. But we just saw that `pthread_setname_np` sets the name in the user space too. Guess, I need to open a ticket in one of Apple's repositories at least to clarify that :D

### DIFF
--- a/src/shims/unix/freebsd/foreign_items.rs
+++ b/src/shims/unix/freebsd/foreign_items.rs
@@ -34,12 +34,15 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             "pthread_get_name_np" => {
                 let [thread, name, len] =
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
-                // FreeBSD's pthread_get_name_np does not return anything.
+                // FreeBSD's pthread_get_name_np does not return anything
+                // and uses strlcpy, which truncates the resulting value,
+                // but always adds a null terminator (except for zero-sized buffers).
+                // https://github.com/freebsd/freebsd-src/blob/c2d93a803acef634bd0eede6673aeea59e90c277/lib/libthr/thread/thr_info.c#L119-L144
                 this.pthread_getname_np(
                     this.read_scalar(thread)?,
                     this.read_scalar(name)?,
                     this.read_scalar(len)?,
-                    false,
+                    /* truncate */ true,
                 )?;
             }
 
--- a/src/shims/unix/freebsd/foreign_items.rs
+++ b/src/shims/unix/freebsd/foreign_items.rs
@@ -39,6 +39,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     this.read_scalar(thread)?,
                     this.read_scalar(name)?,
                     this.read_scalar(len)?,
+                    false,
                 )?;
             }
 
--- a/src/shims/unix/macos/foreign_items.rs
+++ b/src/shims/unix/macos/foreign_items.rs
@@ -164,13 +164,28 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             // Threading
             "pthread_setname_np" => {
                 let [name] = this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+
+                // The real implementation has logic in two places:
+                // * in userland at https://github.com/apple-oss-distributions/libpthread/blob/c032e0b076700a0a47db75528a282b8d3a06531a/src/pthread.c#L1178-L1200,
+                // * in kernel at https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/bsd/kern/proc_info.c#L3218-L3227.
+                //
+                // The function in libc calls the kernel to validate
+                // the security policies and the input. If all of the requirements
+                // are met, then the name is set and 0 is returned. Otherwise, if
+                // the specified name is lomnger than MAXTHREADNAMESIZE, then
+                // ENAMETOOLONG is returned.
+                //
+                // FIXME: the real implementation maybe returns ESRCH if the thread ID is invalid.
                 let thread = this.pthread_self()?;
-                let max_len = this.eval_libc("MAXTHREADNAMESIZE").to_target_usize(this)?;
-                let res = this.pthread_setname_np(
+                let res = if this.pthread_setname_np(
                     thread,
                     this.read_scalar(name)?,
-                    max_len.try_into().unwrap(),
-                )?;
+                    this.eval_libc("MAXTHREADNAMESIZE").to_target_usize(this)?.try_into().unwrap(),
+                )? {
+                    Scalar::from_u32(0)
+                } else {
+                    this.eval_libc("ENAMETOOLONG")
+                };
                 // Contrary to the manpage, `pthread_setname_np` on macOS still
                 // returns an integer indicating success.
                 this.write_scalar(res, dest)?;
@@ -178,10 +193,23 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             "pthread_getname_np" => {
                 let [thread, name, len] =
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
-                let res = this.pthread_getname_np(
+
+                // The function's behavior isn't portable between platforms.
+                // In case of macOS, a truncated name (due to a too small buffer)
+                // does not lead to an error.
+                //
+                // For details, see the implementation at
+                // https://github.com/apple-oss-distributions/libpthread/blob/c032e0b076700a0a47db75528a282b8d3a06531a/src/pthread.c#L1160-L1175.
+                // The key part is the strlcpy, which truncates the resulting value,
+                // but always null terminates (except for zero sized buffers).
+                //
+                // FIXME: the real implementation returns ESRCH if the thread ID is invalid.
+                let res = Scalar::from_u32(0);
+                this.pthread_getname_np(
                     this.read_scalar(thread)?,
                     this.read_scalar(name)?,
                     this.read_scalar(len)?,
+                    /* truncate */ true,
                 )?;
                 this.write_scalar(res, dest)?;
             }
--- a/src/shims/unix/solarish/foreign_items.rs
+++ b/src/shims/unix/solarish/foreign_items.rs
@@ -31,16 +31,20 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     this.read_scalar(name)?,
                     max_len,
                 )?;
+                let res = if res { Scalar::from_u32(0) } else { this.eval_libc("ERANGE") };
                 this.write_scalar(res, dest)?;
             }
             "pthread_getname_np" => {
                 let [thread, name, len] =
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
+                // https://github.com/illumos/illumos-gate/blob/c56822be04b6c157c8b6f2281e47214c3b86f657/usr/src/lib/libc/port/threads/thr.c#L2449-L2480
                 let res = this.pthread_getname_np(
                     this.read_scalar(thread)?,
                     this.read_scalar(name)?,
                     this.read_scalar(len)?,
+                    /* truncate */ false,
                 )?;
+                let res = if res { Scalar::from_u32(0) } else { this.eval_libc("ERANGE") };
                 this.write_scalar(res, dest)?;
             }
 
--- a/src/shims/unix/thread.rs
+++ b/src/shims/unix/thread.rs
@@ -63,38 +63,41 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         interp_ok(Scalar::from_uint(thread_id.to_u32(), this.libc_ty_layout("pthread_t").size))
     }
 
-    /// Set the name of the current thread. `max_name_len` is the maximal length of the name
-    /// including the null terminator.
+    /// Set the name of the specified thread. If the name including the null terminator
+    /// is longer than `name_max_len`, then `false` is returned.
     fn pthread_setname_np(
         &mut self,
         thread: Scalar,
         name: Scalar,
-        max_name_len: usize,
-    ) -> InterpResult<'tcx, Scalar> {
+        name_max_len: usize,
+    ) -> InterpResult<'tcx, bool> {
         let this = self.eval_context_mut();
 
         let thread = thread.to_int(this.libc_ty_layout("pthread_t").size)?;
         let thread = ThreadId::try_from(thread).unwrap();
         let name = name.to_pointer(this)?;
-
         let name = this.read_c_str(name)?.to_owned();
 
         // Comparing with `>=` to account for null terminator.
-        if name.len() >= max_name_len {
-            return interp_ok(this.eval_libc("ERANGE"));
+        if name.len() >= name_max_len {
+            return interp_ok(false);
         }
 
         this.set_thread_name(thread, name);
 
-        interp_ok(Scalar::from_u32(0))
+        interp_ok(true)
     }
 
+    /// Get the name of the specified thread. If the thread name doesn't fit
+    /// the buffer, then if `truncate` is set the truncated name is written out,
+    /// otherwise `false` is returned.
     fn pthread_getname_np(
         &mut self,
         thread: Scalar,
         name_out: Scalar,
         len: Scalar,
-    ) -> InterpResult<'tcx, Scalar> {
+        truncate: bool,
+    ) -> InterpResult<'tcx, bool> {
         let this = self.eval_context_mut();
 
         let thread = thread.to_int(this.libc_ty_layout("pthread_t").size)?;
@@ -104,9 +107,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
         // FIXME: we should use the program name if the thread name is not set
         let name = this.get_thread_name(thread).unwrap_or(b"<unnamed>").to_owned();
-        let (success, _written) = this.write_c_str(&name, name_out, len)?;
+        let name = match truncate {
+            true => &name[..name.len().min(len.try_into().unwrap_or(usize::MAX).saturating_sub(1))],
+            false => &name,
+        };
 
-        interp_ok(if success { Scalar::from_u32(0) } else { this.eval_libc("ERANGE") })
+        let (success, _written) = this.write_c_str(name, name_out, len)?;
+
+        interp_ok(success)
     }
 
     fn sched_yield(&mut self) -> InterpResult<'tcx, ()> {
--- a/tests/pass-dep/libc/pthread-threadname.rs
+++ b/tests/pass-dep/libc/pthread-threadname.rs
@@ -79,10 +79,9 @@ fn main() {
             // if the buffer is shorter than the thread name.
             #[cfg(any(target_os = "illumos", target_os = "solaris"))]
             assert_eq!(get_thread_name(&mut buf[..4]), libc::ERANGE);
-
-            // For libc implementation for macOS it's not an error
-            // for a buffer being too short for the thread name.
-            #[cfg(target_os = "macos")]
+            // On macOS and FreeBSD it's not an error for the buffer to be
+            // too short for the thread name -- they truncate instead.
+            #[cfg(any(target_os = "freebsd", target_os = "macos"))]
             {
                 // Ensure that a zero sized buffer returns no error.
                 assert_eq!(get_thread_name(&mut buf[..0]), 0);
@@ -123,8 +122,8 @@ fn main() {
             // Also test directly calling pthread_setname to check its return value.
             assert_eq!(set_thread_name(&cstr), 0);
 
-            // But with a too long name it should fail (except on FreeBSD where the
-            // function has no return, hence cannot indicate failure).
+            // But with a too long name it should fail (except on FreeBSD where
+            // names of arbitrary size seem to be supported).
             // On macOS, the error code is different.
             #[cfg(not(any(target_os = "freebsd", target_os = "macos")))]
             assert_eq!(set_thread_name(&CString::new(long_name).unwrap()), libc::ERANGE);
--- a/tests/pass-dep/libc/pthread-threadname.rs
+++ b/tests/pass-dep/libc/pthread-threadname.rs
@@ -1,8 +1,22 @@
 //@ignore-target: windows # No pthreads on Windows
-use std::ffi::CStr;
-#[cfg(not(target_os = "freebsd"))]
-use std::ffi::CString;
+use std::ffi::{CStr, CString};
 use std::thread;
+
+const MAX_THREAD_NAME_LEN: usize = {
+    cfg_if::cfg_if! {
+        if #[cfg(any(target_os = "linux"))] {
+            16
+        } else if #[cfg(any(target_os = "illumos", target_os = "solaris"))] {
+            32
+        } else if #[cfg(target_os = "macos")] {
+            libc::MAXTHREADNAMESIZE // 64, at the time of writing
+        } else if #[cfg(target_os = "freebsd")] {
+            usize::MAX // as far as I can tell
+        } else {
+            panic!()
+        }
+    }
+};
 
 fn main() {
     // The short name should be shorter than 16 bytes which POSIX promises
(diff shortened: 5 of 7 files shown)


RalfJung:
Thanks!

Given how non-consistent the behavior is, I am not sure it is worth having the ERANGE in the shared code. Let's just return a bool, and every call site handles it in the way they see fit.

src/shims/unix/thread.rs
```diff
@@ -63,38 +63,66 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    }
+
+    // Get the name of the specified thread. If the thread name doesn't fit
+    // the buffer and trimming isn't allowed, `false` is returned.
+    fn pthread_getname_np(
+        &mut self,
+        thread: Scalar,
+        name_out: Scalar,
+        len: Scalar,
+        trim: bool,
```
RalfJung, line 124: `truncate` would be a better name, IMO

src/shims/unix/thread.rs
```diff
@@ -63,38 +63,66 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
         name_out: Scalar,
         len: Scalar,
     ) -> InterpResult<'tcx, Scalar> {
+        self.pthread_getname_np(thread, name_out, len, false).map(|res| {
+            if res { Scalar::from_u32(0) } else { self.eval_context_ref().eval_libc("ERANGE") }
+        })
+    }
+
+    // Get the name of the specified thread. If the thread name doesn't fit
+    // the buffer and trimming isn't allowed, `false` is returned.
```
RalfJung, line 118:
```suggestion
    // the buffer, then if `truncate` is set the truncated name is written out,
    // otherwise `false` is returned.
```

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -39,6 +39,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     this.read_scalar(thread)?,
                     this.read_scalar(name)?,
                     this.read_scalar(len)?,
+                    false,
```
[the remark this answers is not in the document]
RalfJung: Is `trim: false` even correct? That means the function does nothing at all and returns no error if the buffer is too small? How would I even know whether anything was written to the buffer?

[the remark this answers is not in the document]
YohDeadfall: I will check that later, outside of the current pull request, but for now that is just to keep the current behavior. Anyway, FreeBSD has no return at all.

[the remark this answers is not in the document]
RalfJung: It probably behaves like macOS then...

But you are right that this is pre-existing, so you can also add a FIXME instead of changing behavior here.

YohDeadfall:
> Can you also add a link to some source documenting that a truncated thread name does not lead to an error being returned? That seems like odd behavior...

In my first message you can see that `pthread_getname_np` uses `strlcpy`, and there's documentation for it too. You can see that the only possible error is `ESRCH`, but that requires thread id validation. I can do that, but the current behavior is also fine because:
* Miri makes an out of bounds access to the threads vector and panics,
* it requires changes for other implementations then too, but it won't be correct anyway (you use a vector and indices, but for FreeBSH and macOS a thread id is a pointer to data).

Therefore, the only possible outcome for `pthread_getname_np` is `0`.

src/shims/unix/macos/foreign_items.rs
```diff
@@ -166,22 +166,37 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                let res = this.pthread_getname_np(
+
+                // The function's behavior isn't portable between platforms.
+                // In case of macOS, the result doesn't depend on the output
+                // bufer length. It inherits implementation from FreeBSD, but
+                // on top of it adds the thread pointer validation which can
+                // result in `ESRCH` error code.
+                //
+                // Since all other function implementations don't validate
+                // the provided thread at all, it's not happening here too.
```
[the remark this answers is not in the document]
YohDeadfall: Unfortunately, it means reworking the whole thread manager and making it generation based because for FreeBSD and macOS pointers to the thread data are used, but Miri happily lives in a world of indices. Do you still want a `FIXME` instead of an issue?

src/shims/unix/macos/foreign_items.rs
```diff
@@ -166,22 +166,37 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             "pthread_getname_np" => {
                 let [thread, name, len] =
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
-                let res = this.pthread_getname_np(
+
+                // The function's behavior isn't portable between platforms.
+                // In case of macOS, the result doesn't depend on the output
+                // bufer length. It inherits implementation from FreeBSD, but
+                // on top of it adds the thread pointer validation which can
+                // result in `ESRCH` error code.
```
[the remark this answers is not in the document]
YohDeadfall: Before doing any commit to my branch, do you want it to e squashed into a single commit or keep two as I did?

YohDeadfall:
> Given how non-consistent the behavior is, I am not sure it is worth having the ERANGE in the shared code. Let's just return a bool, and every call site handles it in the way they see fit.

Mhm, let's keep `erange` suffixed functions for a while. I guess they can be used for my Android pull request as well, and maybe for other fixes too. If it wouldn't be so widely used, I will remove them.

src/shims/unix/freebsd/foreign_items.rs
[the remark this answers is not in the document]
YohDeadfall: We can go without a `FIXME` since I plan do that anyway tonight or tomorrow. Solaris and Illumos will be verified too.

RalfJung:
> In my first message you can see that pthread_getname_np uses strlcpy, and there's documentation for it too. 

I asked for this to be included in the source, so that it's not lost to history. :)
Also, including the entire code isn't that useful. "Read the code" isn't good documentation.

> Mhm, let's keep erange suffixed functions for a while.

No, let's not. I think it's nicer to require every caller to handle this explicitly.

src/shims/unix/macos/foreign_items.rs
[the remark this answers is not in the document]
RalfJung: We can have both. :)

> it means reworking the whole thread manager and making it generation based because for FreeBSD and macOS pointers to the thread data are used

I don't understand what you mean by this.

src/shims/unix/macos/foreign_items.rs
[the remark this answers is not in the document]
RalfJung: Please add new commits for changes you are making, so that I can easily review "what changed since the last review". We'll squash at the end.

YohDeadfall:
> I asked for this to be included in the source, so that it's not lost to history. :)
> Also, including the entire code isn't that useful. "Read the code" isn't good documentation.

Should it be a ling then to the source?

src/shims/unix/macos/foreign_items.rs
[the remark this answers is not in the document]
YohDeadfall: I meant when squashing later, should it be a single commit or do you want it be two separate commits as I originally did?

RalfJung:
A link to the source, and a summary of what one sees there.

src/shims/unix/macos/foreign_items.rs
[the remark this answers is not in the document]
RalfJung: Yeah later it can become a single commit (or multiple logically sensible commits if you want, but it's not required for small PRs).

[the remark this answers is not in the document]
YohDeadfall: I thought that two separate commits make more sense, but I made some changes you requested in such a way that they affect both methods at once. So, it requires some time to split changes on squashing into two commits. Let it be a singe commit this time too.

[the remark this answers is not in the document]
RalfJung: I think @oli-obk and me might have different opinions on this "two separate commits" thing. ;)

src/shims/unix/macos/foreign_items.rs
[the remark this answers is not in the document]
YohDeadfall: Never mind, I just checked the source of the `ThreadManager` implementation and found that in never removes threads from the vector, so a situation I had in my mind isn't possible. It was about reusing a preserved identifier of a terminated thread.

tests/pass-dep/libc/pthread-threadname.rs
```diff
@@ -105,8 +118,11 @@ fn main() {
 
             // But with a too long name it should fail (except on FreeBSD where the
             // function has no return, hence cannot indicate failure).
-            #[cfg(not(target_os = "freebsd"))]
-            assert_ne!(set_thread_name(&CString::new(long_name).unwrap()), 0);
+            #[cfg(any(target_os = "linux", target_os = "illumos", target_os = "solaris"))]
```
RalfJung, line 121:
```suggestion
            #[cfg(not(any(target_os = "freebsd", target_os = "macos")))]
```

tests/pass-dep/libc/pthread-threadname.rs
```diff
@@ -105,8 +118,11 @@ fn main() {
 
             // But with a too long name it should fail (except on FreeBSD where the
             // function has no return, hence cannot indicate failure).
```
RalfJung, line 120:
```suggestion
            // But with a too long name it should fail (except on FreeBSD where the
            // function has no return, hence cannot indicate failure).
            // On macOS, the error code is different.
```

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -26,21 +26,22 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                 // https://github.com/illumos/illumos-gate/blob/7671517e13b8123748eda4ef1ee165c6d9dba7fe/usr/src/uts/common/sys/thread.h#L613
                 let max_len = 32;
-                let res = this.pthread_setname_np(
-                    this.read_scalar(thread)?,
-                    this.read_scalar(name)?,
-                    max_len,
-                )?;
+                let res = this
+                    .pthread_setname_np(this.read_scalar(thread)?, this.read_scalar(name)?, max_len)
+                    .map(|r| if r { Scalar::from_u32(0) } else { this.eval_libc("ERANGE") })?;
```
RalfJung, line 31: Why do you do this with `map`? Just do something like
```rust
let success = pthread_setname_np()?;
let res = if success { ... } else { ... };
```

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -26,21 +26,22 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                    this.read_scalar(len)?,
-                )?;
+                let res = this
+                    .pthread_getname_np(
+                        this.read_scalar(thread)?,
+                        this.read_scalar(name)?,
+                        this.read_scalar(len)?,
+                        /* truncate */ false,
+                    )
+                    .map(|r| if r { Scalar::from_u32(0) } else { this.eval_libc("ERANGE") })?;
```
RalfJung, line 44: We seem to have no test for the ERANGE in getname, do we? Can you add that in the "long" part of the test?

src/shims/unix/macos/foreign_items.rs
```diff
@@ -164,24 +164,58 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             "pthread_getname_np" => {
                 let [thread, name, len] =
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
-                let res = this.pthread_getname_np(
+
+                // The function's behavior isn't portable between platforms.
+                // In case of macOS, a truncated name (due to a too small buffer)
+                // does not lead to an error.
+                //
+                // Implemented purely in the userland at
```
RalfJung, line 203:
```suggestion
                // For details, see the implementation at
```

src/shims/unix/macos/foreign_items.rs
```diff
@@ -164,24 +164,58 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                // https://github.com/apple-oss-distributions/libpthread/blob/c032e0b076700a0a47db75528a282b8d3a06531a/src/pthread.c#L1160-L1175.
+                //
+                // The function verifies the thread ID, and if it's valid,
+                // it fills the buffer using strlcpy which truncates the resulting value,
+                // but always null terminates (except zero sized buffers). Therefore,
+                // the result is always 0 for any valid thread ID. If the thread doesn't
+                // exist, then the fucntion returns ESRCH.
+                //
```
RalfJung, line 211:
```suggestion
                // The key part is the strlcpy, which truncates the resulting value,
                // but always null terminates (except for zero sized buffers).
                //
```

src/shims/unix/macos/foreign_items.rs
```diff
@@ -164,24 +164,58 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
-                let res = this.pthread_getname_np(
+
+                // The function's behavior isn't portable between platforms.
+                // In case of macOS, a truncated name (due to a too small buffer)
+                // does not lead to an error.
+                //
+                // Implemented purely in the userland at
+                // https://github.com/apple-oss-distributions/libpthread/blob/c032e0b076700a0a47db75528a282b8d3a06531a/src/pthread.c#L1160-L1175.
+                //
```
RalfJung, line 205:
```suggestion
```

src/shims/unix/macos/foreign_items.rs
```diff
@@ -164,24 +164,58 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                // The function in libc calls the kernel to validate
+                // the security policies and the input. If all of the requirements
+                // are met, then the name is set and 0 is returned. Otherwise, if
+                // the specified name is lomnger than MAXTHREADNAMESIZE, then
+                // ENAMETOOLONG is returned.
+                //
+                // FIXME: the real implementation maybe returns ESRCH if the thread ID is invalid.
+                // Perhaps due to the fact, that there's no man page for that function by Apple,
+                // and not all the kernel code is open sourced.
```
RalfJung, line 180: I don't understand this comment. What is "perhaps due to the fact"?

YohDeadfall: I meant that that it perhaps returns `ESRCH`, but that's not 100% true. Unfortunately, Apples OSS means here's a mirror of our repository and nothing more.

src/shims/unix/solarish/foreign_items.rs
YohDeadfall: We have.

src/shims/unix/solarish/foreign_items.rs
RalfJung: Higher-order functions increase complexity and should only be used when they strictly improve code clarity, which I don't think is the case here. It's also a rather unusual pattern to `map` before `?`, making the reader go "hu, why are they doing this, there must be something non-trivial happening".

src/shims/unix/macos/foreign_items.rs
RalfJung: Then just remove the comment entirely, I don't see why we'd include such speculation here.

src/shims/unix/solarish/foreign_items.rs
RalfJung: Where?

YohDeadfall: [Here](https://github.com/rust-lang/miri/pull/3957/files#diff-d3cdc306c679f0a23f89eaa0872c07b57a7ce0b1f0ecb0840a550114bdbdbb8dR123). You suggested a change to the line above.

src/shims/unix/solarish/foreign_items.rs
YohDeadfall: But isn't `map` on `Result` designed exactly for that?

src/shims/unix/macos/foreign_items.rs
YohDeadfall: I will leave the FIXME still. Otherwise, there's an ICE.

src/shims/unix/freebsd/foreign_items.rs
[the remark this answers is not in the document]
RalfJung: Now we have neither a FIXME nor an issue -- what happened to this thread?

src/shims/unix/thread.rs
```diff
@@ -63,38 +63,41 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
 
         this.set_thread_name(thread, name);
 
-        interp_ok(Scalar::from_u32(0))
+        interp_ok(true)
     }
 
+    // Get the name of the specified thread. If the thread name doesn't fit
+    // the buffer, then if `truncate` is set the truncated name is written out,
+    // otherwise `false` is returned.
```
RalfJung, line 93: Please make this a doc comment

src/shims/unix/thread.rs
```diff
@@ -63,38 +63,41 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         interp_ok(Scalar::from_uint(thread_id.to_u32(), this.libc_ty_layout("pthread_t").size))
     }
 
-    /// Set the name of the current thread. `max_name_len` is the maximal length of the name
-    /// including the null terminator.
+    /// Set the name of the specified thread. If the name including the null terminator
+    // is longer than `name_max_len`, then `false` is returned.
```
RalfJung, line 67:
```suggestion
    /// is longer than `name_max_len`, then `false` is returned.
```

src/shims/unix/solarish/foreign_items.rs
RalfJung: That's set_thread_name. I was talking about get_thread_name.

tests/pass-dep/libc/pthread-threadname.rs
```diff
@@ -74,6 +74,19 @@ fn main() {
...
+            // For libc implementation for macOS it's not an error
+            // for a buffer being too short for the thread name.
+            #[cfg(target_os = "macos")]
+            {
+                assert_eq!(get_thread_name(&mut buf[..0]), 0);
+                assert_eq!(get_thread_name(&mut buf[..4]), 0);
+
+                let cstr = CStr::from_bytes_until_nul(&buf).unwrap();
+
+                assert_eq!(cstr.to_bytes_with_nul().len(), 4);
```
RalfJung, line 87:
```suggestion
                // Ensure we got a prefix of the thread name.
                let cstr = CStr::from_bytes_until_nul(&buf).unwrap();
                assert_eq!(cstr.to_bytes_with_nul().len(), 4);
```

src/shims/unix/freebsd/foreign_items.rs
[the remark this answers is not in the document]
YohDeadfall: A fix is in this pull request now.

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -34,11 +34,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             "pthread_get_name_np" => {
                 let [thread, name, len] =
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
-                // FreeBSD's pthread_get_name_np does not return anything.
+                // FreeBSD's pthread_get_name_np does not return anything
+                // and uses  strlcpy, which truncates the resulting value,
+                // but always null terminates (except for zero sized buffers).
```
RalfJung, line 39:
```suggestion
                // and uses strlcpy, which truncates the resulting value,
                // but always adds a null terminator (except for zero-sized buffers).
```
Also, do you have source for this? Ideally some documentation, since implementation details can change.

src/shims/unix/solarish/foreign_items.rs
RalfJung: In my understanding it is designed for cases where you need to keep the `Result`. But here we are taking the `Result` apart with `?`.

RalfJung:
Thanks for gathering all the intel for all the targets. :)  I guess we should have been more careful when adding these, the "np" should have been a warning that the behavior is really quite non-portable...

@rustbot author

tests/pass-dep/libc/pthread-threadname.rs
```diff
@@ -74,6 +74,21 @@ fn main() {
             // large enough for the thread name.
             #[cfg(target_os = "linux")]
             assert_eq!(get_thread_name(&mut buf[..15]), libc::ERANGE);
+
+            // For libc implementation for macOS it's not an error
+            // for a buffer being too short for the thread name.
```
RalfJung, line 79:
```suggestion
            // On macOS and FreeBSD it's not an error for the buffer to be
            // too short for the thread name -- they truncate instead.
```

tests/pass-dep/libc/pthread-threadname.rs
```diff
@@ -74,6 +74,21 @@ fn main() {
...
+            #[cfg(any(target_os = "freebsd", target_os = "macos"))]
```
RalfJung, line 80: Basically what I think we should have is a test for `#[cfg(not(any(target_os = "freebsd", target_os = "macos")))]` where we call `get_thread_name(&mut buf[..4])` and assert that we get back ERANGE.

tests/pass-dep/libc/pthread-threadname.rs
```diff
@@ -105,8 +120,12 @@ fn main() {
 
             // But with a too long name it should fail (except on FreeBSD where the
             // function has no return, hence cannot indicate failure).
```
RalfJung, line 122:
```suggestion
            // But with a too long name it should fail (except on FreeBSD where
            // names of arbitrary size seem to be supported).
```
That's what the comment in the shim indicates, at least?

tests/pass-dep/libc/pthread-threadname.rs
YohDeadfall: I wouldn't add a test right now, because I haven't verified other platforms still. So, it would be misleading even if it looks correct according to the current interpreter logic. I also have some questions regarding FreeBSD, because it supports GNU extension method to get and set thread names, but they are not used by `std` at the moment while being present in `libc`, and they might have slightly different return codes.

YohDeadfall: Okay, had a quick look at Illumos which should be compatible with Solaris, and there's `ERANGE` too. Will do a test and write a comment.

YohDeadfall: Added just for Solaris like OSes because Linux has it's own already.

YohDeadfall:
@rustbot review

@RalfJung, is it fine now for squashing and merging afterwards?

RalfJung:
Yes please squash the commits. :)

YohDeadfall:
> Yes please squash the commits. :)

Squashed into two logical commits, so there's one for macOS and another for FreeBSD.

@rustbot review

RalfJung:
Thanks! I have tweaked the test a little to have a more consistent shape.

@bors r+

PR #3959, opened 2024-10-10
Fix spelling in README

JakeRoggenbuck:
Great project! I was reading some docs and found these that I think are spelling errors.


saethlin:
Yup! Thanks!
@bors r+

PR #3960, opened 2024-10-10
Pipe minor changes: diagnostics, flag support and comments

### LINKED ISSUE #3912: Improve unnamed_socket diagnostics
Here are some improvements that could be done: 
1. Since unnamed_socket ``read/write`` is shared between ``pipe2`` and ``socketpair``, the diagnostic for ``throw_unsup_format`` shouldn't only mention ``socketpair``:
  - https://github.com/rust-lang/miri/blob/00bafbe76c11dc3ed6a36e4eafafa0e2f4abf388/src/shims/unix/unnamed_socket.rs#L168
  - https://github.com/rust-lang/miri/blob/00bafbe76c11dc3ed6a36e4eafafa0e2f4abf388/src/shims/unix/unnamed_socket.rs#L239
2. Currently, if ``O_NONBLOCK`` flag is used on ``pipe2``, ``unsupported_error`` will be thrown. Since we support non-blocking ``pipe2``, we might want to allow this flag.

tiif:
This PR:
- Add the exact syscall names to the blocking not supported diagnostic
- Added support for ``pipe2`` ``O_NONBLOCK`` 
- Fix minor comment error in ``tests/pass-dep/libc/libc-epoll-blocking.rs``

Fixes #3912

### DIFF
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -292,6 +292,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.write_scalar(result, dest)?;
             }
             "pipe2" => {
+                // Currently this function does not exist on all Unixes, e.g. on macOS.
+                if !matches!(&*this.tcx.sess.target.os, "linux" | "freebsd" | "solaris" | "illumos") {
+                    throw_unsup_format!(
+                        "`pipe2` is not supported on {}",
+                        this.tcx.sess.target.os
+                    );
+                }
                 let [pipefd, flags] =
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
                 let result = this.pipe2(pipefd, Some(flags))?;
--- a/src/shims/unix/unnamed_socket.rs
+++ b/src/shims/unix/unnamed_socket.rs
@@ -163,7 +163,7 @@ impl FileDescription for AnonSocket {
                 } else {
                     // Blocking socketpair with writer and empty buffer.
                     // FIXME: blocking is currently not supported
-                    throw_unsup_format!("socketpair read: blocking isn't supported yet");
+                    throw_unsup_format!("socketpair/pipe/pipe2 read: blocking isn't supported yet");
                 }
             }
         }
@@ -230,7 +230,7 @@ impl FileDescription for AnonSocket {
                 return ecx.set_last_error_and_return(ErrorKind::WouldBlock, dest);
             } else {
                 // Blocking socketpair with a full buffer.
-                throw_unsup_format!("socketpair write: blocking isn't supported yet");
+                throw_unsup_format!("socketpair/pipe/pipe2 write: blocking isn't supported yet");
             }
         }
         // Remember this clock so `read` can synchronize with us.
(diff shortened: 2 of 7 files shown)


src/shims/unix/unnamed_socket.rs
```diff
@@ -347,13 +347,22 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             Some(flags) => this.read_scalar(flags)?.to_i32()?,
             None => 0,
         };
 
         // As usual we ignore CLOEXEC.
         let cloexec = this.eval_libc_i32("O_CLOEXEC");
+
+        let mut is_nonblock = false;
+        // We need this check because `pipe` doesn't have any flags.
+        if flags != 0 {
```
RalfJung, line 360: If `flags == 0` then `flags & this.eval_libc_i32("O_NONBLOCK") == this.eval_libc_i32("O_NONBLOCK")` will always be `false`, so -- why do we need this check?

src/shims/unix/unnamed_socket.rs
```diff
@@ -347,13 +347,22 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-        let flags = match flags {
+        let mut flags = match flags {
             Some(flags) => this.read_scalar(flags)?.to_i32()?,
             None => 0,
         };
 
         // As usual we ignore CLOEXEC.
         let cloexec = this.eval_libc_i32("O_CLOEXEC");
+
+        let mut is_nonblock = false;
```
RalfJung, line 358:
```suggestion
        // Interpret the flag. Every flag we recognize is "subtracted" from `flags`, so
        // if there is anything left at the end, that's an unsupported flag.
        let mut is_nonblock = false;
```

tests/pass-dep/libc/libc-pipe.rs
```diff
@@ -110,3 +112,10 @@ fn test_pipe_array() {
     let mut fds: [i32; 2] = [0; 2];
     assert_eq!(unsafe { pipe(&mut fds) }, 0);
 }
+
+// Test if O_NONBLOCK flag is supported.
+fn test_flag() {
```
RalfJung, line 117: Maybe call this `test_pipe2`? It seems so far we don't have any test for that *oops*.

The `cfg` needs to be repeated here.

tests/pass-dep/libc/libc-pipe.rs
```diff
@@ -7,6 +7,8 @@ fn main() {
     test_pipe_threaded();
     test_race();
     test_pipe_array();
+    #[cfg(target_os = "linux")] // No `pipe2` in macos.
```
RalfJung, line 10: Oh, I didn't realize macOS doesn't have this... then it also needs a check in the shim, similar to [this](https://github.com/rust-lang/miri/blob/124c7adac068f5572a8b4b8f924e764d4a8d0654/src/shims/unix/foreign_items.rs#L671-L676).

RalfJung (commented):
Thanks for the cleanup!

src/shims/unix/unnamed_socket.rs
tiif: Oh yea, right :D

src/shims/unix/foreign_items.rs
```diff
@@ -292,6 +292,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.write_scalar(result, dest)?;
             }
             "pipe2" => {
+                // Currently this function does not exist on all Unixes, e.g. on macOS.
+                if !matches!(&*this.tcx.sess.target.os, "linux") {
```
RalfJung, line 296: Looking at the libc crate, it seems to exist on freebsd, solaris, illumos as well. Can you add those here and in the test?

tests/pass-dep/libc/libc-pipe.rs
```diff
@@ -110,3 +112,11 @@ fn test_pipe_array() {
     let mut fds: [i32; 2] = [0; 2];
     assert_eq!(unsafe { pipe(&mut fds) }, 0);
 }
+
+// Test if O_NONBLOCK flag is supported.
```
RalfJung, line 116:
```suggestion
// Test if pipe2 (including the O_NONBLOCK flag) is supported.
```

src/shims/unix/unnamed_socket.rs
```diff
@@ -347,13 +347,21 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
 
         // As usual we ignore CLOEXEC.
         let cloexec = this.eval_libc_i32("O_CLOEXEC");
+
+        // Interpret the flag. Every flag we recognize is "subtracted" from `flags`, so
+        // if there is anything left at the end, that's an unsupported flag.
+        let mut is_nonblock = false;
+        if flags & this.eval_libc_i32("O_NONBLOCK") == this.eval_libc_i32("O_NONBLOCK") {
+            is_nonblock = true;
+            flags &= !(this.eval_libc_i32("SOCK_NONBLOCK"));
```
RalfJung, line 363: Why is this SOCK_NONBLOCK when the others are O_NONBLOCK?

Please call `eval_libc` only once here. Add a `let nonblock = this.eval_libc_i32("O_NONBLOCK")` just after `let cloexec`.

tests/pass-dep/libc/libc-pipe.rs
tiif: Is it necessary to use doc comment in the test? But sure I will commit it first.

src/shims/unix/unnamed_socket.rs
```diff
@@ -267,21 +267,25 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let mut flags = this.read_scalar(type_)?.to_i32()?;
         let protocol = this.read_scalar(protocol)?.to_i32()?;
         let sv = this.deref_pointer(sv)?;
 
         let mut is_sock_nonblock = false;
 
-        // Parse and remove the type flags that we support.
+        let sock_nonblock = this.eval_libc_i32("SOCK_NONBLOCK");
+        let sock_cloexec = this.eval_libc_i32("SOCK_CLOEXEC");
+
```
tiif, line 278: Changed ``socketpair`` too for consistency :)

tests/pass-dep/libc/libc-pipe.rs
RalfJung: Not *necessary* but it's generally a good idea to use a doc comment when documenting an item. It expresses intent. And since I was changing the comment anyway I made it a doc comment while I am at it.

tiif: Alright, thanks for the explanation!

RalfJung:
Thanks! Looks great. :) I added one more commit to actually test the thing on CI for FreeBSD and Solarish.

@bors r+

Issue #1649, opened 2020-12-13
Posix mutex, rwlock & condvar store identifier index for miri in normal memory.


JCTyblaidd:
The posix mutex, rwlock and condvar store the id miri uses to correspond with the internal implementation and does not prevent any interpreted code from mutating the value.

This can cause unusual behaviour and cause miri to panic.

Although I think this is unlikely to occur in practice it would be good to detect this case, it might be better to store the memory location & associated size for the posix mutex, rwlock and condvar and detect and report any reading and writing of the associated memory as well. I however have not looked at the docs for long enough to be sure of the best solution.

```
// ignore-windows: No libc on Windows

#![feature(rustc_private)]

extern crate libc;

fn main() {
    unsafe {
        let mut lock = libc::PTHREAD_MUTEX_INITIALIZER;
        assert_eq!(libc::pthread_mutex_lock(&mut lock as *mut _), 0);
        
        let miri_ptr: *mut u32 = &mut lock as *mut _ as *mut u32;
        
        // If false miss undefined behaviour or unsupported operation
        // If true then cause miri to panic.
        const ICE: bool = true;
        
        // This operation edits the index that miri uses to map the mutex to the internal
        // mutex object.
        // If set to 0 makes any pthread_mutex operation assign a new internal mutex.
        // Otherwise can be set to an invalid value and cause miri to panic
        // via loading an invalid mutex index.
        // And should probably be reported as UB or unsupported since
        // it modifies internal mutex state.
        *miri_ptr.add(1) = if ICE { 5 } else { 0 };
        
        
        // If ICE = false then will assign a new lock & not detect any undefined behaviour.
        // If ICE = true then will cause miri to panic.
        assert_eq!(libc::pthread_mutex_lock(&mut lock as *mut _), 0);
    }
}
```

RalfJung:
Yeah, there are ways to "mess" with Miri's internals here in ways that can cause ICEs or might not even be detected as errors. That was a deliberate choice to make the implementation simpler -- detecting all possible misbehavior even for programs that "maliciously" try to mess with Miri was declared a non-goal for now.

Improvements for this are welcome. :) But I do not consider the current behavior a bug.

PR #3966, opened 2024-10-12
Do not store synchronization primitive IDs in adressable memory

RalfJung:
We shouldn't store this in a place where the program can mess with it.

Fixes https://github.com/rust-lang/miri/issues/1649

Blocked by https://github.com/rust-lang/rust/pull/131593

### DIFF
--- a/src/concurrency/init_once.rs
+++ b/src/concurrency/init_once.rs
@@ -2,7 +2,6 @@ use std::collections::VecDeque;
 
 use rustc_index::Idx;
 
-use super::sync::EvalContextExtPriv as _;
 use super::vector_clock::VClock;
 use crate::*;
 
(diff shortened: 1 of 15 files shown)


RalfJung:
@bors r-

Blocked by https://github.com/rust-lang/rust/pull/131593

src/shims/unix/linux/sync.rs
```diff
@@ -29,9 +34,15 @@ pub fn futex<'tcx>(
     let op = this.read_scalar(op)?.to_i32()?;
     let val = this.read_scalar(val)?.to_i32()?;
 
+    // Storing this inside the allocation means that if an address gets reused by a new allocation,
+    // we'll use an independent futex queue for this... that seems acceptable.
```
RalfJung, line 38: @rust-lang/miri this is a change in behavior. We used to index futexes entirely on their absolute address, so they stuck around even if the allocation backing that address disappears -- in fact one signal FUTEX_WAKE on unallocated memory. That works fine on Linux but is quite odd from a Rust AM perspective, and gives us no chance to ever clean up old futex state.

So this PR attaches the futex to the "virtual" address inside an allocation, meaning it disappears when the allocation is freed. Strictly speaking this means we will miss wakeups if the address gets reused, and then someone does FUTEX_WAKE and expects to wake up a thread that started blocking on the old address before it got deallocated... but that seems really contrived?

RalfJung: I will factor this commit into a separate PR.

PR #3968, opened 2024-10-13
Added a variadic argument helper

YohDeadfall:
@RalfJung, as you wished (:

### DIFF
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -1180,6 +1180,21 @@ where
     throw_ub_format!("incorrect number of arguments: got {}, expected {}", args.len(), N)
 }
 
+/// Check that the number of args is at least the minumim what we expect.
+pub fn check_min_arg_count<'a, 'tcx, const N: usize>(
+    name: &'a str,
+    args: &'a [OpTy<'tcx>],
+) -> InterpResult<'tcx, &'a [OpTy<'tcx>; N]> {
+    if let Some((ops, _)) = args.split_first_chunk() {
+        return interp_ok(ops);
+    }
+    throw_ub_format!(
+        "incorrect number of arguments for `{name}`: got {}, expected at least {}",
+        args.len(),
+        N
+    )
+}
+
 pub fn isolation_abort_error<'tcx>(name: &str) -> InterpResult<'tcx> {
     throw_machine_stop!(TerminationInfo::UnsupportedInIsolation(format!(
         "{name} not available when isolation is enabled",
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -9,6 +9,7 @@ use std::rc::{Rc, Weak};
 
 use rustc_target::abi::Size;
 
+use crate::helpers::check_min_arg_count;
 use crate::shims::unix::linux::epoll::EpollReadyEvents;
 use crate::shims::unix::*;
 use crate::*;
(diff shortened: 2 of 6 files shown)


src/shims/unix/linux/foreign_items.rs
```diff
@@ -156,22 +149,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                        let [_, initval, flags] =
+                            check_min_arg_count("syscall(SYS_evetfd2, ...)", args)?;
 
                         let result = this.eventfd(initval, flags)?;
                         this.write_int(result.to_i32()?, dest)?;
                     }
-                    id => {
-                        throw_unsup_format!("can't execute syscall with ID {id}");
+                    op => {
+                        throw_unsup_format!("can't execute syscall with OP {op}");
```
YohDeadfall, line 163: I switched to `op` to unify it with other places I have seen across the repo. Moreover, some `libc` variadic functions use `op` too.

src/shims/unix/fd.rs
```diff
@@ -479,58 +480,64 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    return this.set_last_error_and_return_i32(ErrorKind::PermissionDenied);
+                }
 
-            this.ffullsync_fd(fd_num)
-        } else {
-            throw_unsup_format!("the {:#x} command is not supported for `fcntl`)", cmd);
+                this.ffullsync_fd(fd_num)
+            }
+            cmd => {
+                throw_unsup_format!("the {:#x} command is not supported for `fcntl`)", cmd);
```
YohDeadfall, line 539: I feel that the closing parenthesis here was added by mistake, right?

src/shims/unix/fd.rs
```diff
@@ -479,58 +480,64 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    "fnctl(fd, F_DUPFD_CLOEXEC, ...)"
+                };
+
+                let [_, _, start] = check_min_arg_count(cmd_name, args)?;
+                let start = this.read_scalar(start)?.to_i32()?;
+
+                if let Some(fd) = this.machine.fds.get(fd_num) {
+                    interp_ok(Scalar::from_i32(this.machine.fds.insert_with_min_num(fd, start)))
+                } else {
+                    interp_ok(Scalar::from_i32(this.fd_not_found()?))
```
YohDeadfall, line 524: Switched from `if-else` to have the two block exactly under each other. It should be more pleasant to read than before.

src/shims/unix/linux/foreign_items.rs
RalfJung: uppercase "OP" makes little sense though? ID is often used upper-case, but not op...

`man syscall` calls this "number", so maybe the best here is "can't execute syscall number {number}".

src/shims/unix/fd.rs
RalfJung: yeah probably :D

src/helpers.rs
```diff
@@ -1180,6 +1180,21 @@ where
     throw_ub_format!("incorrect number of arguments: got {}, expected {}", args.len(), N)
 }
 
+/// Check that the number of args is at least the minumim what we expect
```
RalfJung, line 1183:
```suggestion
/// Check that the number of args is at least the minumim what we expect.
```

src/shims/unix/fd.rs
```diff
@@ -479,58 +480,64 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     }
 
     fn fcntl(&mut self, args: &[OpTy<'tcx>]) -> InterpResult<'tcx, Scalar> {
+        let [fd_num, cmd] = check_min_arg_count("fnctl", args)?;
+
         let this = self.eval_context_mut();
```
RalfJung, line 486:
```suggestion
        let this = self.eval_context_mut();

        let [fd_num, cmd] = check_min_arg_count("fnctl", args)?;
```
`let this` always comes first, by convention

src/shims/unix/fd.rs
```diff
@@ -479,58 +480,64 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                return this.set_last_error_and_return_i32(ErrorKind::PermissionDenied);
+            cmd if cmd == f_dupfd || cmd == f_dupfd_cloexec => {
+                // Note that we always assume the FD_CLOEXEC flag is set for every open file, in part
+                // because exec() isn't supported. The F_DUPFD and F_DUPFD_CLOEXEC commands only
+                // differ in whether the FD_CLOEXEC flag is pre-set on the new file descriptor,
+                // thus they can share the same implementation here.
+                let cmd_name = if cmd == f_dupfd {
+                    "fnctl(fd, F_DUPFD, ...)"
+                } else {
+                    "fnctl(fd, F_DUPFD_CLOEXEC, ...)"
```
RalfJung, line 515:
```suggestion
                    "fcntl(fd, F_DUPFD, ...)"
                } else {
                    "fcntl(fd, F_DUPFD_CLOEXEC, ...)"
```

src/shims/unix/fd.rs
RalfJung, line 539:
```suggestion
                throw_unsup_format!("fcntl: unsupported command {cmd:#x}");
```

src/shims/unix/linux/foreign_items.rs
YohDeadfall: If number suits you too, then let's change it. I'm just not sure what's intentional here and what naming do you prefer, so trying to be a bit more conservative (:

RalfJung: Well the conservative thing would be to not change it. ;)

But I don't think "syscall id" or "syscall op" are common. "syscall number" however is common. So if we change it, it should be to "number"/`num`.

src/shims/unix/fd.rs
YohDeadfall: Should then I change other places where I find it going after?

RalfJung:
> Should then I change other places where I find it going after?

Yes please do.

tests/fail-dep/libc/fs/unix_open_missing_required_mode.rs
```diff
@@ -8,5 +8,5 @@ fn main() {
 fn test_file_open_missing_needed_mode() {
     let name = b"missing_arg.txt\0";
     let name_ptr = name.as_ptr().cast::<libc::c_char>();
-    let _fd = unsafe { libc::open(name_ptr, libc::O_CREAT) }; //~ ERROR: Undefined Behavior: incorrect number of arguments for `open` with `O_CREAT`: got 2, expected at least 3
+    let _fd = unsafe { libc::open(name_ptr, libc::O_CREAT) }; //~ ERROR: Undefined Behavior: incorrect number of arguments for `open(pathname, O_CRETE, ...)`: got 2, expected at least 3
```
RalfJung, line 11: `O_CRETE` looks wrong?

src/shims/unix/linux/foreign_items.rs
```diff
@@ -127,23 +128,15 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                        "incorrect number of arguments for syscall: got 0, expected at least 1"
-                    );
-                }
-                match this.read_target_usize(&args[0])? {
+                let [op] = check_min_arg_count("syscall", args)?;
+                match this.read_target_usize(op)? {
                     // `libc::syscall(NR_GETRANDOM, buf.as_mut_ptr(), buf.len(), GRND_NONBLOCK)`
                     // is called if a `HashMap` is created the regular way (e.g. HashMap<K, V>).
-                    id if id == sys_getrandom => {
+                    op if op == sys_getrandom => {
```
RalfJung, line 135: So the PR now renames `id` to `op` but then for the message says "number"... that doesn't make a lot of sense. ;) Please use `num` as the variable name.

src/shims/unix/linux/foreign_items.rs
```diff
@@ -156,22 +149,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                        let [_, initval, flags] =
+                            check_min_arg_count("syscall(SYS_evetfd2, ...)", args)?;
 
                         let result = this.eventfd(initval, flags)?;
                         this.write_int(result.to_i32()?, dest)?;
                     }
-                    id => {
-                        throw_unsup_format!("can't execute syscall with ID {id}");
+                    op => {
+                        throw_unsup_format!("can't execute syscall with number {op}");
```
RalfJung, line 163:
```suggestion
                        throw_unsup_format!("syscall: unsupported syscall number {num}");
```

tests/fail-dep/libc/fs/unix_open_missing_required_mode.rs
YohDeadfall: A typo multiplied by copy-paste.

Squash after addressing all the three comments?

RalfJung: > Squash after addressing all the three comments?

Yes

Issue #3898, opened 2024-09-18
Soft-float math library


eduardosm:
I recently published a Rust [floating point math library](https://crates.io/crates/fpmath) which implements math operations (`sqrt`, `sin`, `exp`, `log`...) for both native floating point and soft-floats. When using soft-floats, all operations should produce consistent results regardless of target-specific hard-float subtleties.

Miri could use it to avoid some "host floats" FIXMEs. It that is ok, I'll be glad to create a PR.

RalfJung:
That sounds like a cool library!

However, I am always hesitant to add new dependencies to the rustc workspace, so I think I'd prefer to wait a bit and see how it develops. Also, note that we already have a softfloat library, apfloat -- can your library work with the apfloat types? I see apfloat as a dependency so maybe the answer is yes?

eduardosm:
My library exposes two soft-float types (single and double precision). They internally use rustc_apfloat, but it is an implementation detail, not exposed in the public API. Lossless interoperability is possible through `to_bits` / `from_bits`.

RalfJung:
For Miri it would be a lot nicer if it worked generically on any apfloat type. That would allow reducing code duplication between the f32 and f64 code paths, and will become crucial as we add more support for f16 and f128.

RalfJung:
Looking at the library API docs, I am wondering about two more things:
- It says "with an error of less than 1 ULP". However, I think this is *not* equivalent to saying "rounded to the nearest representable number (using some specific round mode)". There are generally *two* results that are less than 1 ULP away from the infinite-precision result (one "rounding up" and one "rounding down"). So there is still some amount of underspecification here, right?
- Rust doesn't make precision guarantees for these so we'd probably want to disturb the result a bit non-deterministically. ;)

eduardosm:
> For Miri it would be a lot nicer if it worked generically on any apfloat type.

You mean any type that implements the [`rustc_apfloat::Float`](https://docs.rs/rustc_apfloat/0.2.1+llvm-462a31f5a5ab/rustc_apfloat/index.html) trait? That would require arbitrary precision algorithms. Although my implementation is mostly generic, certain parts of the algorithms (such as polynomial approximation or certain constants) are specialized for single and double precision.

Note that direct interoperability with apfloat would generate friction when a new semver-incompatible version of apfloat is released (Miri wouldn't be able to use the new version until my library does).

> * It says "with an error of less than 1 ULP".

[...]

PR #3969, opened 2024-10-14
miri: implement most floating point math operations using soft floats

eduardosm:
This is a preview of how it would look to implement math intrinsics using soft floats.

Deterministic non-determinism (`this.machine.rng`) is used to apply some random error to the results to model the non-guaranteed precision (except for sqrt).

cc https://github.com/rust-lang/miri/issues/3898


RalfJung:
As already mentioned, my main concern here is about pulling in a not-widely-used (and not-widely-vetted) third-party dependency to fix something that's not fundamentally wrong. The only part of this that is a bugfix is using a softfloat implementation for `sqrt`.

How large is an `sqrt` implementation built directly on rustc_apfloat, and could we reasonably carry it in-tree?

PR #3971, opened 2024-10-14
store futexes in per-allocation data rather than globally

RalfJung:
This is a change in behavior. We used to index futexes entirely on their absolute address, so they stuck around even if the allocation backing that address disappears -- in fact one could signal FUTEX_WAKE on unallocated memory. That works fine on Linux but is quite odd from a Rust AM perspective. It also means Miri has a leak, as some futex state never disappears.

So this PR attaches the futex to the "virtual" address inside an allocation, meaning it disappears when the allocation is freed. Strictly speaking this means we will miss wakeups if the address gets reused, and then someone does FUTEX_WAKE and expects to wake up a thread that started blocking on the old address before it got deallocated... but that seems really contrived?

### DIFF
--- a/tests/pass-dep/concurrency/linux-futex.rs
+++ b/tests/pass-dep/concurrency/linux-futex.rs
@@ -41,9 +41,12 @@ fn wake_dangling() {
     let ptr: *const i32 = &*futex;
     drop(futex);
 
-    // Wake 1 waiter. Expect zero waiters woken up, as nobody is waiting.
+    // Expect error since this is now "unmapped" memory.
+    // parking_lot relies on this:
+    // <https://github.com/Amanieu/parking_lot/blob/ca920b31312839013b4455aba1d53a4aede21b2f/core/src/thread_parker/linux.rs#L138-L145>
     unsafe {
-        assert_eq!(libc::syscall(libc::SYS_futex, ptr, libc::FUTEX_WAKE, 1), 0);
+        assert_eq!(libc::syscall(libc::SYS_futex, ptr, libc::FUTEX_WAKE, 1), -1);
+        assert_eq!(io::Error::last_os_error().raw_os_error().unwrap(), libc::EFAULT);
     }
 }
 
--- a/src/concurrency/sync.rs
+++ b/src/concurrency/sync.rs
@@ -1,6 +1,8 @@
+use std::cell::RefCell;
 use std::collections::VecDeque;
 use std::collections::hash_map::Entry;
 use std::ops::Not;
+use std::rc::Rc;
 use std::time::Duration;
 
 use rustc_abi::Size;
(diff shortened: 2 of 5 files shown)


saethlin:
> in fact one could signal FUTEX_WAKE on unallocated memory. That works fine on Linux but is quite odd from a Rust AM perspective

I don't mind declaring such cases Too Weird For Rust. Same goes for reusing addresses; if someone comes up with an actual use case for these we can consider figuring it out then. But that seems unlikely.

RalfJung:
Cc @m-ou-se who did the original implementation here and made it work on al (even dangling) addresses.

m-ou-se:
The only thing that's important is that it doesn't segfault/crash/panic/do anything bad when calling futex_wake on a futex that no longer exists.

So, something like:

```rust
let b = Box::new(AtomicU32::new(0));
let p = b.as_ptr();
// ...
drop(b);
// ...
futex_wake(p, 1); // "use after free" is perfectly fine for futex_wake!
```

This should just do nothing (or spuriously wake another futex, whatever).

tests/pass-dep/concurrency/linux-futex.rs
```diff
@@ -35,17 +35,6 @@ fn wake_nobody() {
     }
 }
 
-fn wake_dangling() {
```
m-ou-se, line 38: Would be nice if this were replaced with a test to check it still accepts "wake after free".

m-ou-se, line 38: This should still work. At least parking_lot depends on "wake after free" to be acceptable.

m-ou-se:
From [parking_lot_core](https://github.com/Amanieu/parking_lot/blob/ca920b31312839013b4455aba1d53a4aede21b2f/core/src/thread_parker/linux.rs#L138-L145):

```rust
        // The thread data may have been freed at this point, but it doesn't
        // matter since the syscall will just return EFAULT in that case.
        let r = libc::syscall(
            libc::SYS_futex,
            self.futex,
            libc::FUTEX_WAKE | libc::FUTEX_PRIVATE_FLAG,
            1,
        );
```

Note that the linux kernel returns EFAULT when the memory is unmapped. In most cases, memory freed by the global allocator is still mapped (so the kernel doesn't know it's deallocated), but it seems perfectly fine for miri to just always return EFAULT if the allocation is gone.

RalfJung:
Returning EFAULT would for the first time let a program "test" whether a given address is allocated. But maybe we should still do it.

m-ou-se:
> Returning EFAULT would for the first time let a program "test" whether a given address is allocated. But maybe we should still do it.

Yeah. It's more accurate to just return success if the memory is deallocated, without waking up anything. I don't think it matters much.

RalfJung:
>  It's more accurate to just return success if the memory is deallocated, without waking up anything.

Yeah I think I prefer that.

RalfJung:
I've also changed the ordering of the read in `futex_wait` to be "acquire" -- it only seems reasonable that if we observe a value this way, we actually establish proper ordering with it. (In practice this is a syscall so should definitely have this "acquire" effect.)

PR #3977, opened 2024-10-17
miri: improve support for `f16` and `f128`

eduardosm:
Rounding intrinsics are now implemented for `f16` and `f128` and tests for `is_infinite`, NaN, `abs`, `copysign`, `min`, `max`, rounding, `*_fast` and `*_algebraic` have been added.


RalfJung:
That looks great, thanks!

@bors r+

PR #3981, opened 2024-10-21
Switch CI to merge queues

Kobzol:
This should be merged in cooperation with an infra admin.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,11 +1,7 @@
 name: CI
 
 on:
-  push:
-    # Run in PRs and for bors, but not on master.
-    branches:
-      - 'auto'
-      - 'try'
+  merge_group:
   pull_request:
     branches:
       - 'master'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -68,27 +64,18 @@ jobs:
...
+  conclusion:
     needs: [build, style]
-    if: github.event.pusher.name == 'bors' && success()
-    steps:
-      - name: mark the job as a success
-        run: exit 0
-  end-failure:
-    name: bors build finished
+    # !cancelled() executes the job regardless of whether the previous jobs passed, failed or get skipped.
+    if: ${{ !cancelled() }}
```
RalfJung, line 70: In https://github.com/rust-lang/miri-test-libstd/blob/master/.github/workflows/ci.yml we use `if: always()`, that seems a bit less odd?

.github/workflows/ci.yml
```diff
@@ -68,27 +64,18 @@ jobs:
...
     runs-on: ubuntu-latest
-    needs: [build, style]
-    if: github.event.pusher.name == 'bors' && (failure() || cancelled())
     steps:
-      - name: mark the job as a failure
-        run: exit 1
+      - name: Conclusion
```
RalfJung, line 73:
```suggestion
      # Manually check the status of all dependencies. `if: failure()` does not work.
      - name: Conclusion
```

.github/workflows/ci.yml
Kobzol: I now tested that it's actually not the same. `always()` will run the conclusion job even if the whole workflow is cancelled (which we rarely do manually). But if we cancel the workflow, we don't want the conclusion job to run. So `!cancelled()` is the better option, I think (yes, GHA statuses are confusing...).

.github/workflows/ci.yml
Kobzol: Added the comment.

.github/workflows/ci.yml
```diff
@@ -68,27 +64,19 @@ jobs:
...
-    runs-on: ubuntu-latest
+  conclusion:
     needs: [build, style]
-    if: github.event.pusher.name == 'bors' && success()
-    steps:
-      - name: mark the job as a success
-        run: exit 0
-  end-failure:
-    name: bors build finished
+    # !cancelled() executes the job regardless of whether the previous jobs passed, failed or get skipped.
```
RalfJung, line 69:
```suggestion
    # We need to ensure this job does *not* get skipped if its dependencies fail,
    # because a skipped job is considered a success by Github. So we have to
    # overwrite `if:`. We use `!cancelled()` to ensure the job does still not get run
    # when the workflow is canceled manually.
```

RalfJung (commented):
Okay, I have one more comment nit that helps explain why we need the `if:` at all.

With that nit applied, feel free to land this and do the infra-side changes whenever it is convenient for you.

.github/workflows/ci.yml
```diff
@@ -68,27 +64,19 @@ jobs:
       - name: rustdoc
         run: RUSTDOCFLAGS="-Dwarnings" ./miri doc --document-private-items
 
-  # These jobs doesn't actually test anything, but they're only used to tell
-  # bors the build completed, as there is no practical way to detect when a
-  # workflow is successful listening to webhooks only.
-  #
-  # ALL THE PREVIOUS JOBS NEED TO BE ADDED TO THE `needs` SECTION OF THIS JOB!
```
RalfJung, line 75: Hm, at least the all-caps part here also seems worth preserving I think?

Kobzol:
Added the all caps part and your suggestion for the comment. Thanks for the review!

MarcoIeni (approved):
great description. We could copy it to other rust repos as well 👍

PR #3982, opened 2024-10-21
epoll_ctl: throw unsupported error on unsupported opcode

RalfJung:
@tiif this is a somewhat suspicious "return -1" without setting the `errno` -- what is the reasoning behind that?

Throwing a clear error seems better to me.


tiif:
This line will only be executed when op is not ``epoll_ctl_del``, ``epoll_ctl_add``, or ``epoll_ctl_mod``. But this will always error out earlier in this check: 
https://github.com/rust-lang/miri/blob/0ebdb0c71cbc023dc34e3a0d41673541b9b58265/src/shims/unix/linux/epoll.rs#L260-L265

But yea the change is reasonable.

PR #3984, opened 2024-10-21
Clear `eval_libc` errors from unix shims

noahmbright:
Each instance of `eval_libc("E` is scrubbed out. Currently, this has the idealistic function calls, but this won't compile yet. Marking as a draft and making comments with the representative compilation errors. Whether we want to hold off on making these erring changes or address the root problem now is up to the reviewers. 

See comments for the issues concretely, but it seems like everything stems from differences between the return types of `this.eval_libc` and `LibcError`.

### DIFF
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -362,8 +362,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // FreeBSD: https://man.freebsd.org/cgi/man.cgi?query=reallocarray
                 match this.compute_size_in_bytes(Size::from_bytes(size), nmemb) {
                     None => {
-                        let enmem = this.eval_libc("ENOMEM");
-                        this.set_last_error(enmem)?;
+                        this.set_last_error(LibcError("ENOMEM"))?;
                         this.write_null(dest)?;
                     }
                     Some(len) => {
@@ -653,13 +652,10 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let chunk_size = CpuAffinityMask::chunk_size(this);
 
                 if this.ptr_is_null(mask)? {
-                    let efault = this.eval_libc("EFAULT");
-                    this.set_last_error(efault)?;
-                    this.write_int(-1, dest)?;
+                    this.set_last_error_and_return(LibcError("EFAULT"), dest)?;
                 } else if cpusetsize == 0 || cpusetsize.checked_rem(chunk_size).unwrap() != 0 {
                     // we only copy whole chunks of size_of::<c_ulong>()
-                    this.set_last_error(LibcError("EINVAL"))?;
-                    this.write_int(-1, dest)?;
+                    this.set_last_error_and_return(LibcError("EINVAL"), dest)?;
                 } else if let Some(cpuset) = this.machine.thread_cpu_affinity.get(&thread_id) {
                     let cpuset = cpuset.clone();
                     // we only copy whole chunks of size_of::<c_ulong>()
@@ -668,9 +664,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     this.write_null(dest)?;
                 } else {
                     // The thread whose ID is pid could not be found
-                    let esrch = this.eval_libc("ESRCH");
-                    this.set_last_error(esrch)?;
-                    this.write_int(-1, dest)?;
+                    this.set_last_error_and_return(LibcError("ESRCH"), dest)?;
                 }
             }
             "sched_setaffinity" => {
@@ -695,9 +689,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 };
 
                 if this.ptr_is_null(mask)? {
-                    let efault = this.eval_libc("EFAULT");
-                    this.set_last_error(efault)?;
-                    this.write_int(-1, dest)?;
+                    this.set_last_error_and_return(LibcError("EFAULT"), dest)?;
                 } else {
                     // NOTE: cpusetsize might be smaller than `CpuAffinityMask::CPU_MASK_BYTES`.
                     // Any unspecified bytes are treated as zero here (none of the CPUs are configured).
@@ -713,8 +705,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                         }
                         None => {
                             // The intersection between the mask and the available CPUs was empty.
-                            this.set_last_error(LibcError("EINVAL"))?;
-                            this.write_int(-1, dest)?;
+                            this.set_last_error_and_return(LibcError("EINVAL"), dest)?;
                         }
                     }
                 }
@@ -770,9 +761,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // macOS: https://keith.github.io/xcode-man-pages/getentropy.2.html
                 // Solaris/Illumos: https://illumos.org/man/3C/getentropy
                 if bufsize > 256 {
-                    let err = this.eval_libc("EIO");
-                    this.set_last_error(err)?;
-                    this.write_int(-1, dest)?;
+                    this.set_last_error_and_return(LibcError("EIO"), dest)?;
                 } else {
                     this.gen_random(buf, bufsize)?;
                     this.write_null(dest)?;
--- a/src/shims/unix/linux/sync.rs
+++ b/src/shims/unix/linux/sync.rs
@@ -158,9 +158,7 @@ pub fn futex<'tcx>(
             } else {
                 // The futex value doesn't match the expected value, so we return failure
                 // right away without sleeping: -1 and errno set to EAGAIN.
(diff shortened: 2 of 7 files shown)


src/shims/alloc.rs
```diff
@@ -109,7 +109,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // Align must be power of 2, and also at least ptr-sized (POSIX rules).
         // But failure to adhere to this is not UB, it's an error condition.
         if !align.is_power_of_two() || align < this.pointer_size().bytes() {
-            interp_ok(this.eval_libc("EINVAL"))
+            interp_ok(LibcError("EINVAL"))
```
noahmbright, line 112: Passing a `LibcError("Efoo")` to `interp_ok` results in 
```expected Scalar<Provenance>, found IoError```

File: src/shims/unix/foreign_items.rs [hunk 2]
```diff
@@ -653,13 +652,10 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let chunk_size = CpuAffinityMask::chunk_size(this);
 
                 if this.ptr_is_null(mask)? {
-                    let efault = this.eval_libc("EFAULT");
-                    this.set_last_error(efault)?;
-                    this.write_int(-1, dest)?;
+                    this.set_last_error_and_return(LibcError("EFAULT"), dest);
```
noahmbright, line 655: I think semantically `set_last_error_and_return` is the reduction of this pattern, but please double check.

src/shims/unix/foreign_items.rs
```diff
@@ -936,7 +925,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     this.write_null(dest)?;
                 } else {
                     this.write_null(&result)?;
-                    this.write_scalar(this.eval_libc("ERANGE"), dest)?;
+                    this.write_scalar(LibcError("ERANGE"), dest)?;
```
noahmbright, line 928: The issue here is that `From<IoError>` is not implemented for `rustc_const_eval::interpret::Scalar<machine::Provenance>`. This can probably be handled in another commit.

src/shims/unix/linux/foreign_items.rs
```diff
@@ -85,7 +85,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     this.read_scalar(name)?,
                     TASK_COMM_LEN,
                 )?;
-                let res = if res { Scalar::from_u32(0) } else { this.eval_libc("ERANGE") };
+                let res = if res { Scalar::from_u32(0) } else { LibcError("ERANGE") };
```
noahmbright, line 88: Here, the `if` and `else` have incompatible types. There might be a simple conversion we can wrap the error in.

src/shims/unix/macos/foreign_items.rs
```diff
@@ -180,11 +180,11 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let res = if this.pthread_setname_np(
                     thread,
                     this.read_scalar(name)?,
-                    this.eval_libc("MAXTHREADNAMESIZE").to_target_usize(this)?.try_into().unwrap(),
+                    LibcError("MAXTHREADNAMESIZE").to_target_usize(this)?.try_into().unwrap(),
```
noahmbright, line 183: `to_target_usize(this)` is not implemented for `IoError`

src/shims/alloc.rs
RalfJung: Yeah leave this unchanged, it can't use the helper. It doesn't set errno anyway.

src/shims/unix/foreign_items.rs
RalfJung: Yes this looks good, except the trailing `?` is missing.

src/shims/unix/foreign_items.rs
RalfJung: Leave this unchanged, it doesn't set errno.

src/shims/unix/linux/foreign_items.rs
RalfJung: Yeah this doesn't set errno so don't change it.

src/shims/unix/linux/sync.rs
```diff
@@ -153,13 +153,12 @@ pub fn futex<'tcx>(
...
-                    this.eval_libc("ETIMEDOUT"),
+                    LibcError("ETIMEDOUT"),
                 );
             } else {
                 // The futex value doesn't match the expected value, so we return failure
                 // right away without sleeping: -1 and errno set to EAGAIN.
-                let eagain = this.eval_libc("EAGAIN");
-                this.set_last_error(eagain)?;
+                this.set_last_error(LibcError("EAGAIN"))?;
                 this.write_scalar(Scalar::from_target_isize(-1, this), dest)?;
```
RalfJung, line 162: This can use `set_last_error_and_return`.

src/shims/unix/linux/sync.rs
```diff
@@ -153,13 +153,12 @@ pub fn futex<'tcx>(
                     Scalar::from_target_isize(0, this), // retval_succ
                     Scalar::from_target_isize(-1, this), // retval_timeout
                     dest.clone(),
-                    this.eval_libc("ETIMEDOUT"),
+                    LibcError("ETIMEDOUT"),
```
RalfJung, line 156: This will require `futex_wait` to be changed: the type of `errno_timeout` should become `IoError`.

src/shims/unix/macos/foreign_items.rs
RalfJung: Yeah leave these unchanged, they don't set errno.

src/shims/unix/solarish/foreign_items.rs
RalfJung, line 1: Also leave these in this file unchanged, again they are not setting errno.

src/shims/unix/sync.rs
RalfJung, line 1: More things to not change as they are not setting errno.

src/shims/alloc.rs
noahmbright: I can revert this, but I'm a little confused. As far as I can see `this.eval_libc` doesn't set the errno, and `LibcError` is just an enum constant. This change would just be for API consistency, right? Just asking in case this is a change you'd want to make eventually.

RalfJung: `LibcError` can only be used with `set_last_error`, so only places that set the errno can be converted.

src/concurrency/sync.rs
```diff
@@ -731,9 +731,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     let futex = this.machine.sync.futexes.get_mut(&addr).unwrap();
                     futex.waiters.retain(|waiter| waiter.thread != thread);
                     // Set errno and write return value.
-                    this.set_last_error(errno_timeout)?;
-                    this.write_scalar(retval_timeout, &dest)?;
-                    interp_ok(())
+                    this.set_last_error_and_return(errno_timeout, &dest)
```
noahmbright, line 734: Checked the types in futex_wait and made this change - changing the type of errno_timeout broke some other functions downstream: a call to `eval_windows` in the Windows shims passed to futex_wait broke, and in concurrency/thread.rs, a `self.$name.visit_provenance(visit)` isn't implemented for IoErrors. 

I could probably add a `WindowsError` variant to the IoError enum to fix the shims error. I could also fill out more of the Windows error table while I'm at it. Less sure about the architecture of the provenance types, to be honest.

RalfJung: Okay let's defer that to a future PR then.

(The right thing would indeed be to add a `WindowsError` variant, and to implement the `VisitProvenance` trait for `IoError`.)

RalfJung:

Note that this does not trigger a notification, so it is easy to be missed. Please do this instead:
@rustbot ready

src/concurrency/sync.rs
```diff
@@ -731,9 +731,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     let futex = this.machine.sync.futexes.get_mut(&addr).unwrap();
                     futex.waiters.retain(|waiter| waiter.thread != thread);
                     // Set errno and write return value.
-                    this.set_last_error(errno_timeout)?;
-                    this.write_scalar(retval_timeout, &dest)?;
```
RalfJung, line 735: This didn't write `-1`, it wrote `retval_timeout`... so your patch changes behavior, and now `retval_timeout` is entirely ignored?

Please revert this part. And generally be more careful about this -- don't touch code you don't understand.

noahmbright: I see, I understand why this is wrong. I'll be more careful. Thank you for the close reviews again.

noahmbright:
Reverted that change in sync.rs and rebased a merge conflict, hence the force push. 
@rustbot ready

RalfJung:
Thanks! Please squash this into a single commit.
@rustbot author

noahmbright:
Done - again thanks for your patience and willingness to help. I was definitely too excited about mindlessly changing every `eval_libc` I could find. This was a good lesson for me. @rustbot ready

RalfJung (approved):

PR #3987, opened 2024-10-22
Implement LLVM x86 vpclmulqdq intrinsics

TDecking:
The vpclmulqdq extension provides carryless multiplication operations for 256-bit and 512-bit length simd operands and expands the similar pclmulqdq extension (notice the missing v). That extension is already supported, which means that most of the work was already done, with no new algorithms getting added.

### DIFF
--- a/src/shims/x86/mod.rs
+++ b/src/shims/x86/mod.rs
@@ -95,11 +95,22 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 }
             }
 
-            "pclmulqdq" => {
+            "pclmulqdq" | "pclmulqdq.256" | "pclmulqdq.512" => {
+                let mut len = 2; // in units of 64bits
+                this.expect_target_feature_for_intrinsic(link_name, "pclmulqdq")?;
+                if unprefixed_name.ends_with(".256") {
+                    this.expect_target_feature_for_intrinsic(link_name, "vpclmulqdq")?;
+                    len = 4;
+                } else if unprefixed_name.ends_with(".512") {
+                    this.expect_target_feature_for_intrinsic(link_name, "vpclmulqdq")?;
+                    this.expect_target_feature_for_intrinsic(link_name, "avx512f")?;
+                    len = 8;
+                }
+
                 let [left, right, imm] =
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
 
-                pclmulqdq(this, left, right, imm, dest)?;
+                pclmulqdq(this, left, right, imm, dest, len)?;
             }
 
             name if name.starts_with("bmi.") => {
@@ -1134,8 +1145,11 @@ fn pmulhrsw<'tcx>(
 /// Perform a carry-less multiplication of two 64-bit integers, selected from `left` and `right` according to `imm8`,
 /// and store the results in `dst`.
 ///
-/// `left` and `right` are both vectors of type 2 x i64. Only bits 0 and 4 of `imm8` matter;
+/// `left` and `right` are both vectors of type `len` x i64. Only bits 0 and 4 of `imm8` matter;
 /// they select the element of `left` and `right`, respectively.
+///
+/// `len` is the SIMD vector length (in counts of `i64` values). It is expected to be one of
+/// `2`, `4`, or `8`.
 ///
 /// <https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_clmulepi64_si128>
 fn pclmulqdq<'tcx>(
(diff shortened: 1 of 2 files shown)


src/shims/x86/mod.rs
```diff
@@ -1144,51 +1155,55 @@ fn pclmulqdq<'tcx>(
...
-
-    for i in 0..64 {
-        // if the i-th bit in right is set
-        if (right & (1 << i)) != 0 {
-            // xor result with `left` shifted to the left by i positions
-            result ^= u128::from(left) << i;
+    for i in 0..(width / 2) {
+        // select the 64-bit integer from left that the user specified (low or high)
+        let index =
+            if (imm8 & 0x01) == 0 { i.wrapping_mul(2) } else { i.wrapping_mul(2).wrapping_add(1) };
```
saethlin, line 1180: I don't think we want wrapping behavior here, if this wraps something has gone quite wrong, right? I think we want `strict_mul` and `strict_add`.

TDecking: Done.

I think it is important to point out that this operation absolutely _cannot_ wrap. It is iterating up to a division by 2, so there is always room for a multiplication by 2 and consequentially an addition by 1.

src/shims/x86/mod.rs
```diff
@@ -1144,51 +1155,55 @@ fn pclmulqdq<'tcx>(
     right: &OpTy<'tcx>,
     imm8: &OpTy<'tcx>,
     dest: &MPlaceTy<'tcx>,
+    width: u64,
```
RalfJung, line 1158: The doc comment should explain the meaning of this argument

src/shims/x86/mod.rs
```diff
@@ -1144,51 +1155,55 @@ fn pclmulqdq<'tcx>(
...
 ) -> InterpResult<'tcx, ()> {
     assert_eq!(left.layout, right.layout);
     assert_eq!(left.layout.size, dest.layout.size);
+    assert!([2u64, 4, 8].contains(&width));
 
-    // Transmute to `[u64; 2]`
+    // Transmute the input into arrays of `u64`.
```
RalfJung, line 1164:
```suggestion
    // Transmute the input into `[u64; width]`.
```

src/shims/x86/mod.rs
```diff
@@ -1144,51 +1155,55 @@ fn pclmulqdq<'tcx>(
...
+    // Transmute the output into an array of `u128`.
```
RalfJung, line 1165:
```suggestion
    // Transmute the output into `[u128; width/2]`.
```

src/shims/x86/mod.rs
```diff
@@ -95,11 +95,22 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 }
             }
 
-            "pclmulqdq" => {
+            "pclmulqdq" | "pclmulqdq.256" | "pclmulqdq.512" => {
+                let mut width = 2;
```
RalfJung, line 99:
```suggestion
                let mut width = 2; // in units of 64bits
```

src/shims/x86/mod.rs
```diff
@@ -1144,51 +1155,55 @@ fn pclmulqdq<'tcx>(
...
-
-    for i in 0..64 {
-        // if the i-th bit in right is set
-        if (right & (1 << i)) != 0 {
-            // xor result with `left` shifted to the left by i positions
-            result ^= u128::from(left) << i;
+    for i in 0..(width / 2) {
+        // select the 64-bit integer from left that the user specified (low or high)
+        let index =
+            if (imm8 & 0x01) == 0 { i.strict_mul(2) } else { i.strict_mul(2).strict_add(1) };
```
RalfJung, line 1180: Seems like it may make sense to `let`-bind `i*2` and `i*2+1`, since both arms use it.

tests/pass/shims/x86/intrinsics-x86-vpclmulqdq.rs
```diff
@@ -0,0 +1,223 @@
...
+        0xA3927047B3FB61D8,
+        0x978093862CDE7187,
+        0xB1927AB22F31D0EC,
+        0xA9A5DA619BE4D7AF,
+        0xCA2590F56884FDC6,
+        0x19BE9F660038BDB5,
+    );
+
+    let mut a_decomp = [_mm_setzero_si128(); 2];
+    unroll! {a_decomp[2] = _mm512_extracti32x4_epi32::<2>(a);}
```
RalfJung, line 151: This function needs `avx512f`, so shouldn't we also enable that in the caller?

src/shims/x86/mod.rs
```diff
@@ -95,11 +95,22 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 }
             }
 
-            "pclmulqdq" => {
+            "pclmulqdq" | "pclmulqdq.256" | "pclmulqdq.512" => {
+                let mut width = 2;
+                this.expect_target_feature_for_intrinsic(link_name, "pclmulqdq")?;
+                if unprefixed_name.ends_with(".256") {
+                    this.expect_target_feature_for_intrinsic(link_name, "vpclmulqdq")?;
```
RalfJung, line 102: Is there a way to test this without requiring avx512f? The current test uses __m512 types even to test the 256bit version of the intrinsic.

TDecking: At minimum it is possible to split the test into two, one for testing with fvx512f, one without. My personal preference however would be to run the same test twice with different compiler flags. Is there a way to do this?

tests/pass/shims/x86/intrinsics-x86-vpclmulqdq.rs
TDecking: I removed avx512 related code for the non-avx512 function.

src/shims/x86/mod.rs
TDecking: Done.

src/shims/x86/mod.rs
TDecking: Fixed.

src/shims/x86/mod.rs
TDecking: Fixed.

src/shims/x86/mod.rs
TDecking: Fixed.

src/shims/x86/mod.rs
TDecking: Fixed. I also renamed it to `len` since we're usually talking about vecor _length_.

src/shims/x86/mod.rs
RalfJung: Yes, that can be done with revisions. Grep the test folder for `@revisions` to see some examples.

TDecking: Done.

RalfJung:
Looks great, thanks. :)

PR #3992, opened 2024-10-25
Android: Added syscall support

### LINKED ISSUE #3618: Android: make std thread APIs work

YohDeadfall:
According to the issue's description I moved the whole `syscall` implementation into a separate file, so it's a little bit more than just "thread APIs". Anyway, all of these system calls are available on Android.

Closes #3618.


tests/pass-dep/concurrency/linux-futex.rs
```diff
@@ -1,4 +1,5 @@
 //@only-target: linux
+//@only-target: android
```
YohDeadfall, line 2: I haven't renamed it since I don't understand the naming rule, but I guess it should be prefixed by `libc` now, right?

RalfJung: No, this is not a libc thing, it's a Linux kernel thing... we don't have clear names to distinguish them (probably should be using Linux vs GNU/Linux). So the existing name is fine for now.

tests/pass-dep/libc/libc-eventfd.rs
```diff
@@ -1,4 +1,5 @@
 //@only-target: linux
+//@only-target: android
```
RalfJung, line 2: You didn't enable the test on Android so this change does nothing.

src/shims/unix/linux/syscall.rs
RalfJung, line 1: Do you want to also put the Android prctl here, and then expose it on Linux targets? It's basically a thin syscall wrapper from what I can see so seems reasonable to have it in that file.

ci/ci.sh
```diff
@@ -154,7 +154,7 @@ case $HOST_TARGET in
     TEST_TARGET=i686-unknown-freebsd   run_tests_minimal $BASIC $UNIX time hashmap random threadname pthread fs libc-pipe
     TEST_TARGET=x86_64-unknown-illumos run_tests_minimal $BASIC $UNIX time hashmap random thread sync available-parallelism tls libc-pipe
     TEST_TARGET=x86_64-pc-solaris      run_tests_minimal $BASIC $UNIX time hashmap random thread sync available-parallelism tls libc-pipe
-    TEST_TARGET=aarch64-linux-android  run_tests_minimal $BASIC $UNIX time hashmap threadname pthread
+    TEST_TARGET=aarch64-linux-android  run_tests_minimal $BASIC $UNIX time hashmap random sync threadname pthread
```
RalfJung, line 157: Do all `thread` tests work with this PR?
```suggestion
    TEST_TARGET=aarch64-linux-android  run_tests_minimal $BASIC $UNIX time hashmap random thread sync
```

src/shims/unix/linux/syscall.rs
YohDeadfall: It's controversial, I would says since almost any call to a `libc` function ends up in `syscall`. Moreover, it's also possible to rename a thread by writing to the corresponding `/proc/self/task/%u/comm` file. Following all the possible ways would end up in Miri turning into a real virtual machine but for Rust only (:

I'm not against that, but I really has no clue if you need that.

tests/pass-dep/concurrency/linux-futex.rs
YohDeadfall: Yeah, it's a Linux thing, but the naming is a bit confusing from the code perspective. Perhaps, it can be named as `linux-like` as in `std`.

tests/pass-dep/libc/libc-eventfd.rs
YohDeadfall: Added an issue: https://github.com/rust-lang/miri/issues/3996.

ci/ci.sh
YohDeadfall: Yup, all the `thread` tests pass.

tests/pass-dep/concurrency/linux-futex.rs
RalfJung: I'm fine with keeping the name here for now.

RalfJung:
Please don't rebase when there isn't a conflict. Github is terrible at dealing with force pushes so rebasing makes re-reviewing a lot harder.

RalfJung:
Looks good, thanks!

Please squash the commits.

YohDeadfall:
> Please don't rebase when there isn't a conflict. Github is terrible at dealing with force pushes so rebasing makes re-reviewing a lot harder.

There were no conflicts at all, but for whatever reason the repository required to update the branch (can be set in the settings). This was blocking the merge ability.

Right now I see the same button, but all the checks are green and the update isn't mandatory.

YohDeadfall:
> This was blocking the merge ability.

Seeing that once again, it seems that there's just an UI issue. While some checks are running, the message "This branch is out-of-date with the base branch" has a gray icon like it's mandatory while it isn't.

RalfJung:
> There were no conflicts at all, but for whatever reason the repository required to update the branch (can be set in the settings). This was blocking the merge ability.

No it doesn't block the merge, you can just ignore that message.

RalfJung:
That force-push has a diff, so when you squashed you didn't use `--keep-base`... I was asking you to use that flag for a reason, it makes reviewing a lot easier. Please follow the instructions we give, they are there for a reason.

RalfJung:
Actually that force push diff is really strange, it goes in the wrong direction?!? What on Earth is github doing here...

YohDeadfall:
> Or did you un-do your earlier rebase? That's not helpful, it just leads to even more diffs github can't deal with...

Shouldn't be since I used `--keep-base`.

YohDeadfall:
Buy the way, I see that this time you didn't use `bors` but instead the built-in  merge queue. Is the Rust project switching them, or is it just that repo?

RalfJung:
> Buy the way, I see that this time you didn't use `bors` but instead the built-in merge queue. Is the Rust project switching them, or is it just that repo?

The main Rust repo won't switch, but most of the rest is slowly switching.

> Shouldn't be since I used --keep-base.

Very strange... but I think somehow your branch did get un-rebased. Maybe you did the rebase with the button in the github UI but didn't pull that change so when you did the squash locally and force-pushed that it undid what you did in the UI?

> Yeah, but the UI is the same as if it's mandatory. There's no way to distinguish these two cases while checks are running. That led me to thing that I had to update the branch first. It's only possible when it's finished, so the check mark is green even if there's the message.

When there is an actual conflict github shows a list of files that have conflicts. It doesn't even start CI in that case. So that should be enough to distinguish the two cases.

YohDeadfall:
> Very strange... but I think somehow your branch did get un-rebased. Maybe you did the rebase with the button in the github UI but didn't pull that change so when you did the squash locally and force-pushed that it undid what you did in the UI?

Indeed, just rechecked the bash history and found that I did:

```bash
git fetch origin
git reset --hard android-syscall
```

Note that no remote was provided, so technically it's noop. The local branch wasn't updated. Sorry for inconvenience (:

> When there is an actual conflict github shows a list of files that have conflicts. It doesn't even start CI in that case. So that should be enough to distinguish the two cases.

Unfortunately not so helpful, I wrote https://github.com/rust-lang/miri/pull/3998#discussion_r1818788011 where explained the situation. Some repositories require updates even if there are no conflicts. But if the CI isn't triggered on mandatory updates, then it might be fine. Still it requires awareness about all GH behaviors which are easy to forget or simply don't know (:

PR #3995, opened 2024-10-26
fix error returned from readdir_r when isolation is enabled, and uses of raw_os_error

RalfJung:
In https://github.com/rust-lang/miri/pull/3779 @oli-obk noticed that our `readdir_r` is buggy -- it shouldn't ever set errno, but the "isolation enabled" path used the shared helper that sets errno.

So this PR fixes that. I think that is the only part of https://github.com/rust-lang/miri/pull/3779 that intentionally changes behavior.

I also noticed some uses of raw_os_error, which is buggy since it returns host error codes that we can't just use for the target.

### DIFF
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -339,7 +339,10 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     _ => interp_ok(this.eval_libc("DT_UNKNOWN").to_u8()?.into()),
                 }
             }
-            Err(e) => this.io_error_to_errnum(e)?.to_i32(),
+            Err(_) => {
+                // Fallback on error
+                interp_ok(this.eval_libc("DT_UNKNOWN").to_u8()?.into())
+            }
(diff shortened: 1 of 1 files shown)


src/shims/unix/fs.rs
```diff
@@ -337,15 +339,7 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-            Err(e) =>
-                match e.raw_os_error() {
-                    Some(error) => interp_ok(error),
-                    None =>
-                        throw_unsup_format!(
-                            "the error {} couldn't be converted to a return value",
-                            e
-                        ),
-                },
+            Err(e) => this.io_error_to_errnum(e)?.to_i32(),
```
RalfJung, line 342: Oh wait this makes no sense at all, we are returning an error code as a file type?!??

saethlin (approved):

RalfJung:
Thanks. :) Could you also press the "merge" button? I am on mobile and do not have access right now.

PR #3997, opened 2024-10-27
Add GC heuristic

JoJoDeveloping:
Miri's GC runs every 10000 basic block, by default. It turns out that sometimes, it is better if the GC runs more often. This is especially true for some Tree Borrows programs, since in Tree Borrow, memory accesses take time linear in the size of a certain tree, and this tree can grow quite large, but is usually compacted significantly at every GC run. For some programs, it can be advantageous to run the GC every 200 blocks, or even more often.

Unfortunately, such a GC frequency would slow down other programs. To achieve a balance, this PR proposes a "heuristics" that allows running the GC more often based on how large the Tree Borrows tree grows. It is quite likely that other parts of Miri want to plug into this mechanisms as well, eventually. However, such concrete other cases are further work, as long as the system is designed to be general enough. This PR only contributes integration with Tree Borrows. It is also possible that the actual heuristics need more tweaks, but for now, the performance improvements look promising.

It is further likely that the overall design needs some change, since it currently relies on `Cell`s. Also, I was not too sure where to put the new `enum`s. So this PR is WIP for now, although it is functionally complete.

### DIFF
--- a/src/borrow_tracker/tree_borrows/tree.rs
+++ b/src/borrow_tracker/tree_borrows/tree.rs
@@ -249,6 +249,8 @@ pub struct Tree {
     pub(super) rperms: RangeMap<UniValMap<LocationState>>,
     /// The index of the root node.
     pub(super) root: UniIndex,
+    /// The number of nodes when we last ran the Garbage Collector
+    nodes_at_last_gc: usize,
 }
 
 /// A node in the borrow tree. Each node is uniquely identified by a tag via
@@ -607,7 +609,7 @@ impl Tree {
             );
             RangeMap::new(size, perms)
         };
-        Self { root: root_idx, nodes, rperms, tag_mapping }
+        Self { root: root_idx, nodes, rperms, tag_mapping, nodes_at_last_gc: 1 }
     }
 }
 
@@ -886,6 +888,22 @@ impl<'tcx> Tree {
 
 /// Integration with the BorTag garbage collector
 impl Tree {
+    /// The number of nodes in this tree
+    pub fn nodes_count(&self) -> usize {
+        self.tag_mapping.len()
+    }
+
+    /// Whether the tree grew significantly since the last provenance GC run
+    pub fn tree_grew_significantly_since_last_gc(&self) -> bool {
+        let current = self.nodes_count();
+        // do not trigger the GC for small nodes
+        let last = self.nodes_at_last_gc.max(50);
+        // trigger the GC if the tree doubled since the last run,
+        // or otherwise got "significantly" larger.
+        // Note that for trees < 100 nodes, nothing happens.
+        current > 2 * last || current > last + 1500
+    }
+
     pub fn remove_unreachable_tags(&mut self, live_tags: &FxHashSet<BorTag>) {
         self.remove_useless_children(self.root, live_tags);
         // Right after the GC runs is a good moment to check if we can
@@ -893,6 +911,7 @@ impl Tree {
         // tags (this does not necessarily mean that they have identical internal representations,
         // see the `PartialEq` impl for `UniValMap`)
         self.rperms.merge_adjacent_thorough();
+        self.nodes_at_last_gc = self.nodes_count();
     }
 
     /// Checks if a node is useless and should be GC'ed.
--- a/src/borrow_tracker/tree_borrows/tree.rs
+++ b/src/borrow_tracker/tree_borrows/tree.rs
@@ -897,11 +897,11 @@ impl Tree {
     pub fn tree_grew_significantly_since_last_gc(&self) -> bool {
         let current = self.nodes_count();
         // do not trigger the GC for small nodes
-        let last = self.nodes_at_last_gc.max(50);
+        let last = self.nodes_at_last_gc.max(100);
         // trigger the GC if the tree doubled since the last run,
         // or otherwise got "significantly" larger.
         // Note that for trees < 100 nodes, nothing happens.
-        current > 2 * last || current > last + 1500
+        current > 2 * last || current > last + 2500
     }
 
     pub fn remove_unreachable_tags(&mut self, live_tags: &FxHashSet<BorTag>) {
--- a/src/provenance_gc.rs
+++ b/src/provenance_gc.rs
@@ -220,8 +220,23 @@ fn remove_unreachable_allocs<'tcx>(ecx: &mut MiriInterpCx<'tcx>, allocs: FxHashS
 
 impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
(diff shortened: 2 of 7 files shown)


saethlin:
Hm, I'm wary of this heuristic because I tried something like it to solve the inverse problem and didn't get good results. Do you have any demonstrations of workloads where this implementation makes a huge difference?

RalfJung:
My main immediate concern is that this can lead to significant slowdown if there are large trees that do *not* get compacted. So I feel like if we have such a heuristic, it should also take into account "did the big tree actually shrink on the last GC run" or something like that, and increase the GC period if the answer is "no".

RalfJung:
What does this PR do on `MIRIFLAGS=-Zmiri-tree-borrows ./miri bench`? Please show numbers from before and after this change.

(Unfortunately we don't have a way to automatically compare two bench reports and print "things got X% slower/faster"... see https://github.com/rust-lang/miri/issues/3999 )

JoJoDeveloping:
> My main immediate concern is that this can lead to significant slowdown if there are large trees that do _not_ get compacted. So I feel like if we have such a heuristic, it should also take into account "did the big tree actually shrink on the last GC run" or something like that, and increase the GC period if the answer is "no".

The way this works is by initially having some exponential back off with how often the GC runs (whenever the size doubles since the last GC run). Eventually that goes to linear (whenever a tree grew by X nodes since the last GC run). The reason is that I recon that most programs do not have many different provenances around. For the tree to not get compacted by the GC requires that programs execute lots of retags and then keep all these pointers alive. On such workloads, the GC already does nothing. Even even such programs would have to create and store a new reference every 7 basic blocks (with the current settings) for it to be slowed down by this patch.

In the end it is all trade-offs. This one seems to be worth it. I am not sure the current performance testsuite consists of programs where this shows (one way or another) but I can add some where it does.

RalfJung:
> The way this works is by initially having some exponential back off with how often the GC runs (whenever the size doubles since the last GC run). 

Right, that makes sense, it just wasn't clear from the PR description. :)

> Eventually that goes to linear (whenever a tree grew by X nodes since the last GC run). 

That's a bit more surprising. What is the rationale for that? Did you compare the PR with and without this part?

> I am not sure the current performance testsuite consists of programs where this shows (one way or another) but I can add some where it does.

Yes that would be good -- preferably one benchmark demonstrating a clear benefit. (We don't want tog row our number of benchmarks too much, either...)

src/provenance_gc.rs
```diff
@@ -208,8 +208,23 @@ fn remove_unreachable_allocs<'tcx>(this: &mut MiriInterpCx<'tcx>, allocs: FxHash
...
 impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
+    fn should_run_provenance_gc(&self) -> bool {
+        let this = self.eval_context_ref();
+        match this.machine.gc_settings {
+            ProvenanceGcSettings::Disabled => false,
+            ProvenanceGcSettings::Regularly { interval } => this.machine.since_gc >= interval,
+            ProvenanceGcSettings::Heuristically =>
+            // no GC run if the last one was recently
+                this.machine.since_gc >= 75
```
RalfJung, line 218: 75 is an oddly specific number... did you try various values? On which benchmark?

JoJoDeveloping: It is mostly based on vibes/previous experience with running TB on a bunch of crates. I found that usually, you find some local minimum (of the GC frequency -> performance) function around 100-500, so 75 seemed like a reasonable lower bound.

Unfortunately I don't think I can pack these all up into a nice test case. But in general, you want to put the cutoff somewhere. If you feel like a different number should be chosen, suggest a better one.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -833,13 +835,30 @@ impl<'tcx> Tree {
...
+    }
+
+    /// Whether the tree grew significantly since the last provenance GC run
+    pub fn tree_grew_significantly_since_last_gc(&self) -> bool {
+        let current = self.nodes_count();
+        // do not trigger the GC for small nodes
+        let last = self.nodes_at_last_gc.max(50);
+        // trigger the GC if the tree doubled since the last run,
+        // or otherwise got "significantly" larger.
+        // Note that for trees < 100 nodes, nothing happens.
```
RalfJung, line 850:
```suggestion
        // Note that for trees <= 100 nodes, nothing happens.
```

src/provenance_gc.rs
RalfJung: > But in general, you want to put the cutoff somewhere. 

This is somewhat surprising, given that the heuristic only kicks in when a tree doubled in size. Did you try different values with the heuristic already in place? Measurements done before you added the heuristic wouldn't really be valid any more.

Also note that this code here is not TB-specific.

This looks more like "guarding against an overeager heuristic" than anything else? Given it is entirely based on vibes, I'd go for 100 and add a comment for why the number was picked. But I am not sure we want a lower bound here at all; each heuristic is really responsible for ensuring a lower bound itself since it has more data.

JoJoDeveloping: > This looks more like "guarding against an overeager heuristic" than anything else?

Indeed, that is also one of the points. It ensures that performance does not tank when a heuristic goes wrong. With the current setup, all the heuristics can do is request the GC happens earlier, so some measures should be taken that it does not happen too early.

JoJoDeveloping:
> We don't want to grow our number of benchmarks too much, either...

Do we not? Changes like the one in this PR would be much easier to find/develop/test if the suite of benchmarks was more diverse and larger in general. It its current form, it is useful only to ensure there are no _really major_ performance regressions. But that's a separate issue.

RalfJung:
More benchmarks means they take forever to run, and at least until we figure out https://github.com/rust-lang/miri/issues/3999 also forever to analyze.

src/provenance_gc.rs
RalfJung: A buggy heuristic making the GC fire all the time is bad even if we "cap" it at some number here this way. So this would paper over the issue and make it harder to notice, not fix it.

RalfJung:
>  I found that there are some more cases where changing the numbers here so that the GC runs less often.

Sorry I can't parse this, what do you mean?

JoJoDeveloping:
Huh, I seem to have swallowed a sentence there. What I mean is that I found some programs where this change is bad because it makes the GC run too often. Instead, these programs achieve optimal performance when the "magic numbers" in this PR are moved up a bit, so that the trees need to grow more before the GC is triggered. So finding what is a good number there is nontrivial/impossible, which makes me more unhappy with the current approach.

RalfJung:
Well that's the usual issue with heuristics...

JoJoDeveloping:
I did not intent to close it.

RalfJung:

Issue #3967, opened 2024-10-12
Get rid of IDs for synchronization primitives, make pthread shims not leaky


RalfJung:
We currently store the state of our synchronization primitives with a layer of indirection: there are lists in `SynchronizationObjects` and then we use IDs to index into these lists. These lists keep growing while the program runs and never get cleaned up, which isn't very clean and also makes @saethlin unhappy. ;)

As of https://github.com/rust-lang/miri/pull/3966, we no longer store these IDs in machine memory, so there isn't actually a good reason to still use IDs. Instead, the `PthreadMutex` used by `pthread_mutex_t` can just directly store a `sync::Mutex`. That way, when the allocation that contains the mutex is freed, we also free all the extra data associated with the mutex, thus avoiding memory leaks.

The main trouble here is functions like `mutex_unlock(&mut self, id: MutexId)`: these would want to take a reference to the `Mutex` instead, but since they also take a mutable reference to the entire machine state, that can't work! We could carefully arrange things to avoid overlapping references here... but probably the easier approach is to use reference counting. So we'd make `PthreadMutex` store a `MutexRef` (which is a newtype for `Rc<RefCell<Mutex>>`) and the function signatures would look like
```rust
fn mutex_unlock(&mut self, id: &MutexRef) -> ...
```
https://github.com/rust-lang/miri/pull/3971 does this for `Futex` so that can serve as a model.

Currently we are also using these IDs in `BlockReason`, but that's just a sanity check, so it's okay to remove them there.

Also as part of this, we could finally fix the leaks in `pthread_mutex_destroy` and friends.

tiif:
I still have a few things to clear in my todo list, but I'd like to work on a different part of the codebase, so, 

@rustbot claim

RalfJung:
Note that this is blocked on https://github.com/rust-lang/miri/pull/3966

tiif:
@RalfJung Is this also blocked on #3971? If not I could start working on this soon :)

RalfJung:
> Is this also blocked on https://github.com/rust-lang/miri/pull/3971?

No. That PR can just be helpful as a guide for what this could look like -- there we have `Futex` and `FutexRef`, so here we'd have e.g. `Mutex` and `MutexRef`.

tiif:
@rustbot release-assignment

I have quite a few things on my queue right now. Current ``SynchronizationObjects`` is left with three lists, which I expect all of them could be removed. 

```rust
pub struct SynchronizationObjects {
[...]

PR #4002, opened 2024-10-30
Remove MutexID list

tiif:
Remove MutexID list in ``SynchronizationObjects`` to avoid memory leaks. 

cc #3967

### DIFF
--- a/src/concurrency/sync.rs
+++ b/src/concurrency/sync.rs
@@ -1,6 +1,9 @@
+use std::cell::RefCell;
 use std::collections::VecDeque;
 use std::collections::hash_map::Entry;
+use std::default::Default;
 use std::ops::Not;
+use std::rc::Rc;
 use std::time::Duration;
 
 use rustc_data_structures::fx::FxHashMap;
@@ -44,8 +47,6 @@ macro_rules! declare_id {
 }
 pub(super) use declare_id;
 
-declare_id!(MutexId);
-
 /// The mutex state.
 #[derive(Default, Debug)]
 struct Mutex {
@@ -57,6 +58,21 @@ struct Mutex {
     queue: VecDeque<ThreadId>,
     /// Mutex clock. This tracks the moment of the last unlock.
     clock: VClock,
+}
+
+#[derive(Default, Clone, Debug)]
+pub struct MutexRef(Rc<RefCell<Mutex>>);
+
+impl MutexRef {
+    fn new() -> Self {
+        MutexRef(Rc::new(RefCell::new(Mutex::default())))
+    }
+}
+
+impl VisitProvenance for MutexRef {
+    fn visit_provenance(&self, _visit: &mut VisitWith<'_>) {
+        // Mutex contains no provenance.
+    }
 }
 
 declare_id!(RwLockId);
@@ -133,7 +149,6 @@ struct FutexWaiter {
 /// The state of all synchronization objects.
 #[derive(Default, Debug)]
 pub struct SynchronizationObjects {
-    mutexes: IndexVec<MutexId, Mutex>,
     rwlocks: IndexVec<RwLockId, RwLock>,
     condvars: IndexVec<CondvarId, Condvar>,
     pub(super) init_onces: IndexVec<InitOnceId, InitOnce>,
@@ -147,17 +162,17 @@ impl<'tcx> EvalContextExtPriv<'tcx> for crate::MiriInterpCx<'tcx> {}
 pub(super) trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
     fn condvar_reacquire_mutex(
         &mut self,
-        mutex: MutexId,
+        mutex_ref: &MutexRef,
         retval: Scalar,
         dest: MPlaceTy<'tcx>,
     ) -> InterpResult<'tcx> {
         let this = self.eval_context_mut();
-        if this.mutex_is_locked(mutex) {
-            assert_ne!(this.mutex_get_owner(mutex), this.active_thread());
-            this.mutex_enqueue_and_block(mutex, Some((retval, dest)));
+        if this.mutex_is_locked(mutex_ref) {
+            assert_ne!(this.mutex_get_owner(mutex_ref), this.active_thread());
+            this.mutex_enqueue_and_block(mutex_ref, Some((retval, dest)));
         } else {
             // We can have it right now!
-            this.mutex_lock(mutex);
+            this.mutex_lock(mutex_ref);
             // Don't forget to write the return value.
             this.write_scalar(retval, &dest)?;
         }
@@ -166,10 +181,9 @@ pub(super) trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
 }
 
 impl SynchronizationObjects {
-    pub fn mutex_create(&mut self) -> MutexId {
-        self.mutexes.push(Default::default())
+    pub fn mutex_create(&mut self) -> MutexRef {
+        MutexRef::new()
     }
-
     pub fn rwlock_create(&mut self) -> RwLockId {
         self.rwlocks.push(Default::default())
     }
@@ -201,7 +215,7 @@ impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     /// Helper for lazily initialized `alloc_extra.sync` data:
     /// this forces an immediate init.
-    fn lazy_sync_init<T: 'static + Copy>(
+    fn lazy_sync_init<T: 'static>(
         &mut self,
         primitive: &MPlaceTy<'tcx>,
         init_offset: Size,
@@ -227,7 +241,10 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     /// - If yes, fetches the data from `alloc_extra.sync`, or calls `missing_data` if that fails
     ///   and stores that in `alloc_extra.sync`.
     /// - Otherwise, calls `new_data` to initialize the primitive.
-    fn lazy_sync_get_data<T: 'static + Copy>(
+    ///
+    /// The return value is a *clone* of the stored data, so if you intend to mutate it
+    /// better wrap everything into an `Rc`.
+    fn lazy_sync_get_data<T: 'static + Clone>(
         &mut self,
         primitive: &MPlaceTy<'tcx>,
         init_offset: Size,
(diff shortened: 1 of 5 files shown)


tests/pass-dep/libc/pthread-sync.rs
```diff
@@ -215,6 +214,8 @@ fn check_rwlock_read_no_deadlock() {
     }
 }
 
+// TODO: There is a data race error when there is no -Zmiri-preemption-rate=0, test this again after
+// the ICE is resolved.
 fn check_cond() {
```
tiif, line 219: Surprisingly miri reported data race on this test when there is no ``-Zmiri-preemption-rate=0``. I haven't looked into this yet, but I will revert this change for now as ``-Zmiri-preemption-rate=0`` is needed to trigger the ICE.

tiif:
I feel I have been constantly running into ``already mutably borrowed: BorrowError`` when it comes to ``unblock_thread``. It always takes me a while to realise that it has something to do with ``unblock_thread`` again.

RalfJung:
Generally you have to take great care of your `RefCell` handles. Never keep them any longer than you have to.

If we didn't use `RefCell`, you'd get a borrow checking error due to changing a vector while iterating it. That would be nicer but unfortunately the reference patterns are too dynamic to make that work.

tiif:
> For every function you call while holding the handle, ensure it doesn't reentrantly access the same data again.

This is a pretty nice way to see it, thanks.

tests/pass-dep/libc/pthread-sync.rs
tiif: Before the data race is reported, there is another failed assertion reported, so the data race might be an side effect of the failed assertion? If no one's alarm went off about this, it's probably fine. 
<details>
<summary> error trace </summary>

```rust
thread 'main' panicked at ./tests/pass-dep/libc/pthread-sync.rs:247:9:
assertion failed: data.ptr.read() == 1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
note: in Miri, you may have to set `MIRIFLAGS=-Zmiri-env-forward=RUST_BACKTRACE` for the environment variable to have an effect
error: Undefined Behavior: Data race detected between (1) atomic store on thread `unnamed-11` and (2) deallocation on thread `main` at alloc23901. (2) just happened here
   --> ./tests/pass-dep/libc/pthread-sync.rs:218:1
    |
218 | / fn check_cond() {
219 | |     unsafe {
220 | |         let mut cond: MaybeUninit<libc::pthread_cond_t> = MaybeUninit::uninit();
221 | |         assert_eq!(libc::pthread_cond_init(cond.as_mut_ptr(), ptr::null()), 0);
...   |
263 | |     }
264 | | }
    | |_^ Data race detected between (1) atomic store on thread `unnamed-11` and (2) deallocation on thread `main` at alloc23901. (2) just happened here
    |
   --> ./tests/pass-dep/libc/pthread-sync.rs:234:24
    |
234 |             assert_eq!(libc::pthread_mutex_lock(mutex.ptr), 0); // This is one
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: inside `check_cond` at ./tests/pass-dep/libc/pthread-sync.rs:218:1: 264:2
note: inside `main`
   --> ./tests/pass-dep/libc/pthread-sync.rs:21:5
    |
21  |     check_cond();
    |     ^^^^^^^^^^^^

```
</details>

But I couldn't understand the diagnostic. ``(2)`` is pointed at the end of the function, which is not really helpful for me to understand where exactly it is. 

```rust
218 | / fn check_cond() {
219 | |     unsafe {
220 | |         let mut cond: MaybeUninit<libc::pthread_cond_t> = MaybeUninit::uninit();
221 | |         assert_eq!(libc::pthread_cond_init(cond.as_mut_ptr(), ptr::null()), 0);
...   |
263 | |     }
264 | | }
    | |_^ Data race detected between (1) atomic store on thread `unnamed-11` and (2) deallocation on thread `main` at alloc23901. (2) just happened here
```

src/concurrency/sync.rs
```diff
@@ -44,7 +47,20 @@ macro_rules! declare_id {
...
+    fn new() -> Self {
+        MutexRef(Rc::new(RefCell::new(Mutex::default())))
+    }
+}
+
+impl VisitProvenance for MutexRef {
+    fn visit_provenance(&self, _visit: &mut VisitWith<'_>) {
+        // No provenance
+    }
+}
```
tiif, line 63: Is there anything need to be implemented here?

src/shims/unix/sync.rs
```diff
@@ -176,10 +176,11 @@ fn mutex_create<'tcx>(
...
     let data = PthreadMutex { id, kind };
-    ecx.lazy_sync_init(&mutex, mutex_init_offset(ecx)?, data)?;
+    ecx.lazy_sync_init(&mutex, mutex_init_offset(ecx)?, data.clone())?;
     interp_ok(data)
 }
 
+// TODO: check if the comment is correct?
 /// Returns the `MutexId` of the mutex stored at `mutex_op`.
 ///
 /// `mutex_get_id` will also check if the mutex has been moved since its first use and
```
tiif, line 186: This comment looks a bit weird. ``mutex_get_data`` returns a ``PthreadMutex``, and why is there ``mutex_op`` and ``mutex_get_id``?

tiif (commented):
Aside from the questions in the comments, this is ready

@rustbot ready

tests/pass-dep/libc/pthread-sync.rs
RalfJung: What is the status of this? The diff seems to be done, there is no ICE.

> (2) is pointed at the end of the function

Yes, that's where the local variables get deallocated.

RalfJung: The reason for the race is that we are panicking, and then we are not executing the `t.join().unwrap()`, and then we get a race.

RalfJung: Resolved by https://github.com/rust-lang/miri/pull/4020

src/concurrency/sync.rs
```diff
@@ -44,7 +47,20 @@ macro_rules! declare_id {
 }
 pub(super) use declare_id;
 
-declare_id!(MutexId);
+#[derive(Default, Clone, Debug)]
+pub struct MutexRef(Rc<RefCell<Mutex>>);
```
RalfJung, line 51: Please define this right after `struct Mutex` so that related things are close together.

src/concurrency/sync.rs
RalfJung: No, please add a comment saying that `Mutex` contains no provenance.

src/concurrency/sync.rs
```diff
@@ -147,17 +162,17 @@ impl<'tcx> EvalContextExtPriv<'tcx> for crate::MiriInterpCx<'tcx> {}
 pub(super) trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
     fn condvar_reacquire_mutex(
         &mut self,
-        mutex: MutexId,
+        mutex_ref: MutexRef,
```
RalfJung, line 165: This should take `&MutexRef`, no need to make the caller give up ownership here

src/concurrency/sync.rs
```diff
@@ -354,15 +366,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                         mutex.clock.clone_from(clock)
                     });
                 }
-                if let Some(thread) = this.machine.sync.mutexes[id].queue.pop_front() {
-                    this.unblock_thread(thread, BlockReason::Mutex(id))?;
+                let thread_id = mutex.queue.pop_front();
+                if thread_id.is_some() {
+                    // We need to drop mutex before unblock_thread because mutex will be borrowed again
+                    // in the unblock callback.
```
RalfJung, line 372:
```suggestion
                    // We need to drop mutex before unblock_thread because mutex will be borrowed again
                    // in the unblock callback.
```
```suggestion
                    // We need to drop our mutex borrow before unblock_thread
                    //  because it will be borrowed again in the unblock callback.
```

src/concurrency/sync.rs
```diff
@@ -354,15 +366,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    drop(mutex);
```
RalfJung, line 373: Can't you do the drop before the if? That seems cleaner. Always drop as early as possible.

src/concurrency/sync.rs
```diff
@@ -354,15 +366,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    this.unblock_thread(thread_id.unwrap(), BlockReason::Mutex)?;
                 }
             }
-            Some(old_lock_count)
-        } else {
-            // Mutex is not locked.
-            None
-        })
+            return interp_ok(Some(old_lock_count));
```
RalfJung, line 377: Why did you change this to use early-return?

src/concurrency/sync.rs
```diff
@@ -227,7 +241,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     /// - If yes, fetches the data from `alloc_extra.sync`, or calls `missing_data` if that fails
     ///   and stores that in `alloc_extra.sync`.
     /// - Otherwise, calls `new_data` to initialize the primitive.
```
RalfJung, line 243:
```suggestion
    /// - Otherwise, calls `new_data` to initialize the primitive.
    ///
    /// The return value is a *clone* of the stored data, so if you intend to mutate it
    /// better wrap everything into an `Rc`.
```

src/shims/unix/sync.rs
RalfJung: Yeah that comment is just outdated. Make it like this:
```
/// Returns the mutex data stored at the address that `mutex_ptr` points to.
/// Will raise an error if the mutex has been move since its first use.
```

RalfJung (changes requested):
Thanks, looks great overall! But there's a bunch of minor things.

src/concurrency/sync.rs
tiif: That's a spurious change introduced when I try to debug fix the reborrow ICE. It is now fixed.

src/concurrency/sync.rs
```diff
@@ -610,14 +630,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     fn condvar_wait(
         &mut self,
         condvar: CondvarId,
-        mutex: MutexId,
+        mutex_ref: &MutexRef,
```
RalfJung, line 633: This one needs to clone anyway so it makes sense that it would take an owned ref (so we can avoid double-cloning).

src/shims/unix/sync.rs
```diff
@@ -175,15 +175,13 @@ fn mutex_create<'tcx>(
...
+    ecx.lazy_sync_init(&mutex, mutex_init_offset(ecx)?, data.clone())?;
     interp_ok(data)
 }
 
-/// Returns the `MutexId` of the mutex stored at `mutex_op`.
-///
-/// `mutex_get_id` will also check if the mutex has been moved since its first use and
-/// return an error if it has.
+/// Returns the mutex data stored at the address that `mutex_ptr` points to.
+/// Will raise an error if the mutex has been move since its first use.
```
RalfJung, line 184:
```suggestion
/// Will raise an error if the mutex has been moved since its first use.
```

src/concurrency/sync.rs
```diff
@@ -354,8 +369,12 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                         mutex.clock.clone_from(clock)
                     });
                 }
-                if let Some(thread) = this.machine.sync.mutexes[id].queue.pop_front() {
-                    this.unblock_thread(thread, BlockReason::Mutex(id))?;
+                let thread_id = mutex.queue.pop_front();
+                // We need to drop our mutex borrow before unblock_thread
+                //  because it will be borrowed again in the unblock callback.
```
RalfJung, line 374:
```suggestion
                // because it will be borrowed again in the unblock callback.
```

RalfJung:
Just some final nits. :)
You can squash after taking care of those (with `--keep-base` as usual).

@rustbot author

RalfJung (approved):

RalfJung:
Oh, wtf... as long as there is a "requested changes" review, it won't merge, even if the merge button is pressed by the person that requested the changes...

PR #4004, opened 2024-10-30
Get/set thread name shims return errors for invalid handles

### LINKED ISSUE #3965: ICE on invalid thread ID in Linux pthread_getname_np
This program ICEs Miri:
```rust
fn get_thread_name(name: &mut [u8]) -> i32 { unsafe {
    libc::pthread_getname_np(999, name.as_mut_ptr().cast(), name.len())
} }

fn main() {
     get_thread_name(&mut []);
}
```

Thanks to @YohDeadfall for mentioning that. This is definitely a bug, we should never ICE.

YohDeadfall:
Fixes #3965.

### DIFF
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -1,7 +1,6 @@
 //! Implements threads.
 
 use std::mem;
-use std::num::TryFromIntError;
 use std::sync::atomic::Ordering::Relaxed;
 use std::task::Poll;
 use std::time::{Duration, SystemTime};
@@ -124,26 +123,6 @@ impl Idx for ThreadId {
 
     fn index(self) -> usize {
         usize::try_from(self.0).unwrap()
-    }
-}
-
-impl TryFrom<u64> for ThreadId {
-    type Error = TryFromIntError;
-    fn try_from(id: u64) -> Result<Self, Self::Error> {
-        u32::try_from(id).map(Self)
-    }
-}
-
-impl TryFrom<i128> for ThreadId {
-    type Error = TryFromIntError;
-    fn try_from(id: i128) -> Result<Self, Self::Error> {
-        u32::try_from(id).map(Self)
-    }
-}
-
-impl From<u32> for ThreadId {
-    fn from(id: u32) -> Self {
-        Self(id)
     }
 }
 
@@ -448,6 +427,10 @@ pub enum TimeoutAnchor {
     Absolute,
 }
 
+/// An error signaling that the requested thread doesn't exist.
+#[derive(Debug, Copy, Clone)]
+pub struct ThreadNotFound;
+
 /// A set of threads.
 #[derive(Debug)]
 pub struct ThreadManager<'tcx> {
@@ -509,6 +492,16 @@ impl<'tcx> ThreadManager<'tcx> {
         }
     }
 
+    pub fn thread_id_try_from(&self, id: impl TryInto<u32>) -> Result<ThreadId, ThreadNotFound> {
+        if let Ok(id) = id.try_into()
+            && usize::try_from(id).is_ok_and(|id| id < self.threads.len())
+        {
+            Ok(ThreadId(id))
+        } else {
+            Err(ThreadNotFound)
+        }
+    }
+
     /// Check if we have an allocation for the given thread local static for the
     /// active thread.
     fn get_thread_local_alloc_id(&self, def_id: DefId) -> Option<StrictPointer> {
@@ -534,6 +527,7 @@ impl<'tcx> ThreadManager<'tcx> {
     ) -> &mut Vec<Frame<'tcx, Provenance, FrameExtra<'tcx>>> {
         &mut self.threads[self.active_thread].stack
     }
+
     pub fn all_stacks(
         &self,
     ) -> impl Iterator<Item = (ThreadId, &[Frame<'tcx, Provenance, FrameExtra<'tcx>>])> {
@@ -868,6 +862,11 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
 // Public interface to thread management.
 impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    #[inline]
+    fn thread_id_try_from(&self, id: impl TryInto<u32>) -> Result<ThreadId, ThreadNotFound> {
+        self.eval_context_ref().machine.threads.thread_id_try_from(id)
+    }
+
     /// Get a thread-specific allocation id for the given thread-local static.
     /// If needed, allocate a new one.
     fn get_or_create_thread_local_alloc(
@@ -1160,8 +1159,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     /// Set the name of the current thread. The buffer must not include the null terminator.
     #[inline]
     fn set_thread_name(&mut self, thread: ThreadId, new_thread_name: Vec<u8>) {
-        let this = self.eval_context_mut();
-        this.machine.threads.set_thread_name(thread, new_thread_name);
+        self.eval_context_mut().machine.threads.set_thread_name(thread, new_thread_name);
     }
 
     #[inline]
--- a/src/shims/unix/linux/foreign_items.rs
+++ b/src/shims/unix/linux/foreign_items.rs
@@ -81,13 +81,17 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             "pthread_setname_np" => {
                 let [thread, name] =
                     this.check_shim(abi, ExternAbi::C { unwind: false }, link_name, args)?;
-                let res = this.pthread_setname_np(
+                let res = match this.pthread_setname_np(
                     this.read_scalar(thread)?,
                     this.read_scalar(name)?,
                     TASK_COMM_LEN,
                     /* truncate */ false,
-                )?;
-                let res = if res { Scalar::from_u32(0) } else { this.eval_libc("ERANGE") };
+                )? {
+                    ThreadNameResult::Ok => Scalar::from_u32(0),
+                    ThreadNameResult::NameTooLong => this.eval_libc("ERANGE"),
+                    // Act like we faild to open `/proc/self/task/$tid/comm`.
+                    ThreadNameResult::ThreadNotFound => this.eval_libc("ENOENT"),
+                };
                 this.write_scalar(res, dest)?;
             }
             "pthread_getname_np" => {
@@ -97,14 +101,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // In case of glibc, the length of the output buffer must
                 // be not shorter than TASK_COMM_LEN.
                 let len = this.read_scalar(len)?;
-                let res = if len.to_target_usize(this)? >= TASK_COMM_LEN as u64
-                    && this.pthread_getname_np(
+                let res = if len.to_target_usize(this)? >= TASK_COMM_LEN as u64 {
+                    match this.pthread_getname_np(
                         this.read_scalar(thread)?,
                         this.read_scalar(name)?,
                         len,
                         /* truncate*/ false,
                     )? {
-                    Scalar::from_u32(0)
+                        ThreadNameResult::Ok => Scalar::from_u32(0),
+                        ThreadNameResult::NameTooLong => unreachable!(),
+                        // Act like we faild to open `/proc/self/task/$tid/comm`.
+                        ThreadNameResult::ThreadNotFound => this.eval_libc("ENOENT"),
+                    }
                 } else {
                     this.eval_libc("ERANGE")
                 };
--- a/src/shims/unix/macos/foreign_items.rs
+++ b/src/shims/unix/macos/foreign_items.rs
@@ -181,18 +181,16 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // are met, then the name is set and 0 is returned. Otherwise, if
                 // the specified name is lomnger than MAXTHREADNAMESIZE, then
                 // ENAMETOOLONG is returned.
-                //
-                // FIXME: the real implementation maybe returns ESRCH if the thread ID is invalid.
                 let thread = this.pthread_self()?;
-                let res = if this.pthread_setname_np(
+                let res = match this.pthread_setname_np(
                     thread,
                     this.read_scalar(name)?,
                     this.eval_libc("MAXTHREADNAMESIZE").to_target_usize(this)?.try_into().unwrap(),
                     /* truncate */ false,
                 )? {
-                    Scalar::from_u32(0)
-                } else {
-                    this.eval_libc("ENAMETOOLONG")
+                    ThreadNameResult::Ok => Scalar::from_u32(0),
+                    ThreadNameResult::NameTooLong => this.eval_libc("ENAMETOOLONG"),
+                    ThreadNameResult::ThreadNotFound => unreachable!(),
                 };
                 // Contrary to the manpage, `pthread_setname_np` on macOS still
                 // returns an integer indicating success.
@@ -210,15 +208,17 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // https://github.com/apple-oss-distributions/libpthread/blob/c032e0b076700a0a47db75528a282b8d3a06531a/src/pthread.c#L1160-L1175.
                 // The key part is the strlcpy, which truncates the resulting value,
                 // but always null terminates (except for zero sized buffers).
-                //
-                // FIXME: the real implementation returns ESRCH if the thread ID is invalid.
-                let res = Scalar::from_u32(0);
-                this.pthread_getname_np(
+                let res = match this.pthread_getname_np(
                     this.read_scalar(thread)?,
                     this.read_scalar(name)?,
                     this.read_scalar(len)?,
                     /* truncate */ true,
-                )?;
+                )? {
+                    ThreadNameResult::Ok => Scalar::from_u32(0),
+                    // `NameTooLong` is possible when the buffer is zero sized,
+                    ThreadNameResult::NameTooLong => Scalar::from_u32(0),
+                    ThreadNameResult::ThreadNotFound => this.eval_libc("ESRCH"),
+                };
                 this.write_scalar(res, dest)?;
             }
 
--- a/src/shims/unix/solarish/foreign_items.rs
+++ b/src/shims/unix/solarish/foreign_items.rs
@@ -26,26 +26,33 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // THREAD_NAME_MAX allows a thread name of 31+1 length
                 // https://github.com/illumos/illumos-gate/blob/7671517e13b8123748eda4ef1ee165c6d9dba7fe/usr/src/uts/common/sys/thread.h#L613
                 let max_len = 32;
-                let res = this.pthread_setname_np(
+                // See https://illumos.org/man/3C/pthread_setname_np for the error codes.
+                let res = match this.pthread_setname_np(
                     this.read_scalar(thread)?,
                     this.read_scalar(name)?,
                     max_len,
                     /* truncate */ false,
-                )?;
-                let res = if res { Scalar::from_u32(0) } else { this.eval_libc("ERANGE") };
+                )? {
+                    ThreadNameResult::Ok => Scalar::from_u32(0),
+                    ThreadNameResult::NameTooLong => this.eval_libc("ERANGE"),
+                    ThreadNameResult::ThreadNotFound => this.eval_libc("ESRCH"),
+                };
                 this.write_scalar(res, dest)?;
             }
             "pthread_getname_np" => {
                 let [thread, name, len] =
                     this.check_shim(abi, ExternAbi::C { unwind: false }, link_name, args)?;
-                // https://github.com/illumos/illumos-gate/blob/c56822be04b6c157c8b6f2281e47214c3b86f657/usr/src/lib/libc/port/threads/thr.c#L2449-L2480
-                let res = this.pthread_getname_np(
+                // See https://illumos.org/man/3C/pthread_getname_np for the error codes.
+                let res = match this.pthread_getname_np(
                     this.read_scalar(thread)?,
                     this.read_scalar(name)?,
                     this.read_scalar(len)?,
                     /* truncate */ false,
-                )?;
-                let res = if res { Scalar::from_u32(0) } else { this.eval_libc("ERANGE") };
+                )? {
+                    ThreadNameResult::Ok => Scalar::from_u32(0),
+                    ThreadNameResult::NameTooLong => this.eval_libc("ERANGE"),
+                    ThreadNameResult::ThreadNotFound => this.eval_libc("ESRCH"),
+                };
                 this.write_scalar(res, dest)?;
             }
 
--- a/src/shims/unix/thread.rs
+++ b/src/shims/unix/thread.rs
@@ -2,6 +2,13 @@ use rustc_abi::ExternAbi;
 
 use crate::*;
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ThreadNameResult {
+    Ok,
+    NameTooLong,
+    ThreadNotFound,
+}
+
 impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     fn pthread_create(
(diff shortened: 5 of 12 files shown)


YohDeadfall:
So, I feel satisfied now. There was an issue with the thread identifier conversion causing ICEs, but that happens no more.

@rustbot review

src/shims/unix/linux/foreign_items.rs
```diff
@@ -79,13 +79,16 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                     this.read_scalar(thread)?,
                     this.read_scalar(name)?,
                     TASK_COMM_LEN,
                     /* truncate */ false,
-                )?;
-                let res = if res { Scalar::from_u32(0) } else { this.eval_libc("ERANGE") };
+                )? {
+                    ThreadNameResult::Ok => Scalar::from_u32(0),
+                    ThreadNameResult::NameTooLong => this.eval_libc("ERANGE"),
+                    ThreadNameResult::ThreadNotFound => this.eval_libc("ENOENT"),
```
RalfJung, line 90: I don't see ENOENT mentioned in the man page. Where did you get that error code from?

src/shims/unix/macos/foreign_items.rs
```diff
@@ -174,18 +174,16 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                     this.read_scalar(name)?,
                     this.eval_libc("MAXTHREADNAMESIZE").to_target_usize(this)?.try_into().unwrap(),
                     /* truncate */ false,
                 )? {
-                    Scalar::from_u32(0)
-                } else {
-                    this.eval_libc("ENAMETOOLONG")
+                    ThreadNameResult::Ok => Scalar::from_u32(0),
+                    ThreadNameResult::NameTooLong => this.eval_libc("ENAMETOOLONG"),
+                    ThreadNameResult::ThreadNotFound => this.eval_libc("ESRCH"),
```
RalfJung, line 186: Where did you get the ESRCH from? I don't see it in the code links above.

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -26,26 +26,32 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // THREAD_NAME_MAX allows a thread name of 31+1 length
                 // https://github.com/illumos/illumos-gate/blob/7671517e13b8123748eda4ef1ee165c6d9dba7fe/usr/src/uts/common/sys/thread.h#L613
                 let max_len = 32;
-                let res = this.pthread_setname_np(
+                let res = match this.pthread_setname_np(
```
RalfJung, line 29:
```suggestion
                // See https://illumos.org/man/3C/pthread_setname_np for the error codes.
                let res = match this.pthread_setname_np(
```

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -26,26 +26,32 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    ThreadNameResult::Ok => Scalar::from_u32(0),
+                    ThreadNameResult::NameTooLong => this.eval_libc("ERANGE"),
+                    ThreadNameResult::ThreadNotFound => this.eval_libc("ESRCH"),
+                };
                 this.write_scalar(res, dest)?;
             }
             "pthread_getname_np" => {
                 let [thread, name, len] =
                     this.check_shim(abi, Abi::C { unwind: false }, link_name, args)?;
                 // https://github.com/illumos/illumos-gate/blob/c56822be04b6c157c8b6f2281e47214c3b86f657/usr/src/lib/libc/port/threads/thr.c#L2449-L2480
```
RalfJung, line 44:
```suggestion
                // See https://illumos.org/man/3C/pthread_getname_np for the error codes.
```

src/shims/windows/foreign_items.rs
```diff
@@ -509,9 +509,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 };
 
                 // FIXME: use non-lossy conversion
-                this.set_thread_name(thread, String::from_utf16_lossy(&name).into_bytes());
+                let res = match this
+                    .set_thread_name(thread, String::from_utf16_lossy(&name).into_bytes())
+                {
+                    Ok(_) => Scalar::from_u32(0),
+                    Err(_) => Scalar::from_u32(3489660936),
```
RalfJung, line 516: What is this random number?^^

Please use a named constant instead.

tests/pass-dep/libc/pthread-threadname.rs
```diff
@@ -199,4 +199,24 @@ fn main() {
         .unwrap()
         .join()
         .unwrap();
+
+    // Now set the name for a non-existing thread and verify error codes.
+    cfg_if::cfg_if! {
+        if #[cfg(any(target_os = "linux", target_os = "illumos", target_os = "solaris"))] {
```
RalfJung, line 205: What about macOS?

src/shims/windows/foreign_items.rs
YohDeadfall: A magic undocumented number I wrote you in the issue about. If you write in a binary form then you will find that two higher bits are set to `0b11` which indicates an error. The next bytes are for the reversed and customer bits (don't remember the order, but it doesn't matter) are zeroes. The facility is zero too, and the error code is `0x1000`. That's all I can say to you about that magic number.

It's not documented at all, neither in the function returns, nor as a well known error code.

And just for your info, in Windows there are two very similar date types, ` NTSTATUS` in the kernel space and `HRESULT` in the user space. Both are bit fields rather than just numbers, but both follow the same rule as everywhere, a negative number means an error. Positive numbers can mean success or a warning depending on the 30 bit.

src/shims/unix/linux/foreign_items.rs
YohDeadfall: That's again the same problem I told you about in the corresponding issue. Not every documentation is perfect (:

tests/pass-dep/libc/pthread-threadname.rs
YohDeadfall: Forgot about it since it allows changing the thread name only from it  will do that, but only for the get method.

src/shims/unix/macos/foreign_items.rs
YohDeadfall: From the source code (:

See https://github.com/rust-lang/miri/issues/3965#issuecomment-2449413110.

src/concurrency/thread.rs
```diff
@@ -697,13 +701,22 @@ impl<'tcx> ThreadManager<'tcx> {
     }
 
     /// Set the name of the given thread.
-    pub fn set_thread_name(&mut self, thread: ThreadId, new_thread_name: Vec<u8>) {
-        self.threads[thread].thread_name = Some(new_thread_name);
+    pub fn set_thread_name(
+        &mut self,
+        thread: ThreadId,
+        new_thread_name: Vec<u8>,
+    ) -> Result<(), ThreadNotFound> {
```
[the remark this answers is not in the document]
YohDeadfall: Make sense, will move the validation logic and remove the conversions from `ThreadId`.

src/shims/windows/foreign_items.rs
RalfJung: Where did you get that number from?

tests/pass-dep/libc/pthread-threadname.rs
RalfJung: Sorry I can't parse that sentence.

src/shims/unix/macos/foreign_items.rs
RalfJung: Miri's source code needs to be self-explaining. So please include sufficient references to explain the errors codes.

src/shims/unix/linux/foreign_items.rs
RalfJung: Well you got the code from somewhere. So please add comments explaining where you got it from.

src/shims/windows/foreign_items.rs
ChrisDenton: It's the NTSTATUS [`STATUS_INVALID_HANDLE`](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55) (`0xC0000008`) encoded as a [HRESULT](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/0642cb2f-2075-4469-918c-4441e69c548a) by setting the `N` bit. Which gives the value `0xD0000008`.

YohDeadfall: > Where did you get that number from?

I got it by writing an app that used the thread description functions and feeding them invalid thread handles.

src/shims/unix/linux/foreign_items.rs
YohDeadfall: Ah, wrong place, wasn't able to see the file you left the comment for, so thought that you are about FREEBASE or something else.

For Linux it's quite clear, the man pages state that all error codes unrelated to the buffer length are from `open` system call. For others there are links to the source already.

src/shims/unix/macos/foreign_items.rs
YohDeadfall: Should I write it as a comment? The link to the original source is here already.

src/shims/windows/foreign_items.rs
RalfJung: Please use a `const STATUS_INVALID_HANDLE_HRESULT = 0xD0000008` to make this more readable.

src/shims/unix/linux/foreign_items.rs
RalfJung: > For Linux it's quite clear, the man pages state that all error codes unrelated to the buffer length are from open system call. 

Ah good point, I had missed that.

RalfJung, line 90:
```suggestion
                    // Act like we faild to open `/proc/self/task/$tid/comm`.
                    ThreadNameResult::ThreadNotFound => this.eval_libc("ENOENT"),
```

src/shims/unix/macos/foreign_items.rs
RalfJung: I didn't see the ESRCH in the original source though, did I look in the wrong place?

tests/pass-dep/libc/pthread-threadname.rs
YohDeadfall: `pthread_setname_np` doesn't accept a thread handle as the first argument like it is for the other platforms, so the function sets the name for the current thread only.  That's why I forgot about a test for `pthread_getname_np`.

src/shims/unix/macos/foreign_items.rs
YohDeadfall: Ah, sorry, that's about setting the name. Should be unreachable then, there's no way to set the name for a thread other than the current.

tests/pass-dep/libc/pthread-threadname.rs
RalfJung: Then there should be an `unreachable!()` for this error case in the macOS shim.

RalfJung: Also seems like `pthread_getname_np` still allows querying other threads on macOS, so please add a test for that.

src/shims/unix/macos/foreign_items.rs
RalfJung: Yes for getname I can see the ESRCH :)

src/concurrency/thread.rs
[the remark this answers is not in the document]
YohDeadfall: @RalfJung, I moved validation into the thread manager and now, from my point of view, it's worse due to the Windows' handles. The `Handle` type is an enumeration one of which variants stores a strongly typed thread identifier. That implies that it's valid, but in the reality each handle should be validated. Please advise.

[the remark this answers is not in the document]
RalfJung: Hm... we have two options:
- change `Handle::Thread` to hold a `u32`, meaning that the ID hasn't been validated yet
- wrap `Handle::from_scalar` in a helper with access to `InterpCx`, like `handle_try_from`, and have it to `ThreadId` validation

The first option is probably easier to realize, and should be good enough for now.

src/concurrency/thread.rs
```diff
@@ -513,6 +498,15 @@ impl<'tcx> ThreadManager<'tcx> {
         }
     }
 
+    pub fn thread_id_try_from(&self, id: impl TryInto<u32>) -> Result<ThreadId, ThreadNotFound> {
+        if let Ok(id) = id.try_into()
+            && id.try_into().unwrap_or(usize::MAX) < self.threads.len()
```
RalfJung, line 503: The second `try_into` is converting `u32` to `usize`, right?
```suggestion
            && usize::try_from(id).is_some_and(|id| id < self.threads.len())
```

src/concurrency/thread.rs
```diff
@@ -513,6 +498,15 @@ impl<'tcx> ThreadManager<'tcx> {
...
+        {
+            return Ok(ThreadId(id));
+        }
+        Err(ThreadNotFound)
```
RalfJung, line 507:
```suggestion
            Ok(ThreadId(id))
        } else {
            Err(ThreadNotFound)
        }
```

src/shims/windows/foreign_items.rs
```diff
@@ -11,6 +11,8 @@ use crate::shims::os_str::bytes_to_os_str;
 use crate::shims::windows::*;
 use crate::*;
 
+const STATUS_INVALID_HANDLE: u32 = 0xD0000008;
```
RalfJung, line 14:
```suggestion
// The NTSTATUS STATUS_INVALID_HANDLE (0xC0000008) encoded as a HRESULT by setting the N bit.
// (https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/0642cb2f-2075-4469-918c-4441e69c548a)
const STATUS_INVALID_HANDLE: u32 = 0xD0000008;
```

src/concurrency/thread.rs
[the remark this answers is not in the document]
YohDeadfall: Now it's even worse, but I made an issue #4014 to fix that (:

src/shims/windows/foreign_items.rs
```diff
@@ -513,10 +513,11 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
                 let thread = match Handle::from_scalar(handle, this)? {
                     Some(Handle::Thread(thread)) => thread,
```
RalfJung, line 515: Please to the `thread_id_try_from` in here; no need to convert `active_thread()` to a `u32` and back.

src/shims/windows/thread.rs
```diff
@@ -66,7 +66,12 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let timeout = this.read_scalar(timeout_op)?.to_u32()?;
 
         let thread = match Handle::from_scalar(handle, this)? {
-            Some(Handle::Thread(thread)) => thread,
+            Some(Handle::Thread(thread)) => {
+                match this.thread_id_try_from(thread) {
+                    Ok(thread) => thread,
+                    Err(_) => return interp_ok(0xFFFFFFFF), // WAIT_FAILED
```
RalfJung, line 72: The docs seem to say that this also sets the last-error variable.

tests/pass-dep/libc/pthread-threadname.rs
```diff
@@ -213,9 +213,15 @@ fn main() {
...
-            let res = unsafe { libc::pthread_setname_np(invalid_thread, [0].as_ptr()) };
-            assert_eq!(res, error);
-            let res = unsafe { libc::pthread_getname_np(invalid_thread, [0; MAX_THREAD_NAME_LEN].as_mut_ptr(), MAX_THREAD_NAME_LEN) };
+            cfg_if::cfg_if!{
+                if #[cfg(not(target_os = "macos"))] {
+                    // macOS has no function accepting a thread id as the first argument.
+                    let res = unsafe { libc::pthread_setname_np(invalid_thread, [0].as_ptr()) };
+                    assert_eq!(res, error);
+                }
+            }
```
RalfJung, line 222: No need to use `cfg_if` here, just use normal `#[cfg]` on a block. `cfg_if` only makes sense when we want an `else`.

src/shims/windows/thread.rs
YohDeadfall: That's not documented too, so I will make a test app tomorrow. Or maybe will find something in [the definitive list of the error codes](https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes).

RalfJung: Maybe @ChrisDenton knows

ChrisDenton: That would be `ERROR_INVALID_HANDLE`.

src/concurrency/thread.rs
```diff
@@ -127,23 +126,9 @@ impl Idx for ThreadId {
...
-    fn try_from(id: i128) -> Result<Self, Self::Error> {
-        u32::try_from(id).map(Self)
-    }
-}
-
 impl From<u32> for ThreadId {
-    fn from(id: u32) -> Self {
-        Self(id)
+    fn from(value: u32) -> Self {
+        Self(value)
```
RalfJung, line 131: Is this `impl` still needed?

src/shims/unix/linux/foreign_items.rs
RalfJung: @YohDeadfall you marked this as resolved, but the comment isn't there. Please don't do that, it makes a lot of extra work to re-review the same thing again and again.

src/shims/unix/linux/foreign_items.rs
```diff
@@ -97,14 +100,17 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    match this.pthread_getname_np(
                         this.read_scalar(thread)?,
                         this.read_scalar(name)?,
                         len,
                         /* truncate*/ false,
                     )? {
-                    Scalar::from_u32(0)
+                        ThreadNameResult::Ok => Scalar::from_u32(0),
+                        ThreadNameResult::NameTooLong => unreachable!(),
+                        ThreadNameResult::ThreadNotFound => this.eval_libc("ENOENT"),
```
RalfJung, line 112:
```suggestion
                        // Act like we faild to open `/proc/self/task/$tid/comm`.
                        ThreadNameResult::ThreadNotFound => this.eval_libc("ENOENT"),
```

src/shims/unix/solarish/foreign_items.rs
RalfJung: Yet another comment that was marked as resolved without actually being resolved.

src/shims/unix/solarish/foreign_items.rs
RalfJung: And another one

RalfJung:
I will stop here, please re-check all the comments I left that you marked as resolved to ensure they are truly resolved.

@rustbot author

src/shims/unix/solarish/foreign_items.rs
YohDeadfall: It was automatically resolved when I applied `5dbf95bfd27066bf64711bdd5c53e592f8c6f727` with your suggestions, but then I rebased the pull request and lost it. I know that I'm quite unfocused lately and try to improve that, but that's not easy to do. Sorry!

RalfJung: I recommend never using `git push --force`, and always `--force-with-lease`. That can help prevent mistakes like that.

You can set up an alias to make this easier, as `git pushf`:
```
[alias]
        pushf = push --force-with-lease
```

YohDeadfall:
Seems to be fine now, @RalfJung. The previously lost suggestions are restored via https://github.com/rust-lang/miri/pull/4004/commits/7d8fe1287e389a1d7d4bdea92f4228943a62a7c5, and the latest were addressed by https://github.com/rust-lang/miri/pull/4004/commits/853593e202cbc35a4297483887f9e023c41c9747.

@rustbot review

src/shims/unix/thread.rs
```diff
@@ -30,30 +37,43 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
 
         if !this.ptr_is_null(this.read_pointer(retval)?)? {
             // FIXME: implement reading the thread function's return place.
             throw_unsup_format!("Miri supports pthread_join only with retval==NULL");
         }
 
-        let thread_id = this.read_scalar(thread)?.to_int(this.libc_ty_layout("pthread_t").size)?;
-        this.join_thread_exclusive(thread_id.try_into().expect("thread ID should fit in u32"))?;
+        let thread = this.read_scalar(thread)?.to_int(this.libc_ty_layout("pthread_t").size)?;
+        let res = match this.thread_id_try_from(thread) {
```
RalfJung, line 53: Can you use `let Ok(thread) = ... else { ... }` here like below?

I think the code becomes a lot clearer if errors are handled via early return Currently the actual core functionality of this function is hidden inside a `match`.

src/shims/unix/thread.rs
```diff
@@ -30,30 +37,43 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    fn pthread_detach(&mut self, thread: &OpTy<'tcx>) -> InterpResult<'tcx, Scalar> {
         let this = self.eval_context_mut();
 
-        let thread_id = this.read_scalar(thread)?.to_int(this.libc_ty_layout("pthread_t").size)?;
-        this.detach_thread(
-            thread_id.try_into().expect("thread ID should fit in u32"),
-            /*allow_terminated_joined*/ false,
-        )?;
+        let thread = this.read_scalar(thread)?.to_int(this.libc_ty_layout("pthread_t").size)?;
+        let res = match this.thread_id_try_from(thread) {
```
RalfJung, line 68: Same here for `let ... else`

tests/pass-dep/libc/pthread-threadname.rs
```diff
@@ -199,4 +199,28 @@ fn main() {
         .unwrap()
         .join()
         .unwrap();
+
+    // Now set the name for a non-existing thread and verify error codes.
+    cfg_if::cfg_if! {
+        if #[cfg(any(target_os = "linux", target_os = "illumos", target_os = "solaris", target_os = "macos"))] {
```
RalfJung, line 205: This has no good reason to be `cfg_if` either, you can make it a block.

Also, can you make this `cfg(not(any(...)))` to ensure that when we add new OSes, we consider the question whether they need to be checked here?

RalfJung (commented):
Thanks!
I think we are getting close. :)

src/shims/windows/handle.rs
```diff
@@ -154,17 +154,22 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-        match Handle::from_scalar(handle, this)? {
-            Some(Handle::Thread(thread)) =>
-                this.detach_thread(thread, /*allow_terminated_joined*/ true)?,
+        let ret = match Handle::from_scalar(handle, this)? {
+            Some(Handle::Thread(thread)) => {
+                if let Ok(thread) = this.thread_id_try_from(thread) {
+                    this.detach_thread(thread, /*allow_terminated_joined*/ true)?;
+                    this.eval_windows("c", "TRUE")
+                } else {
+                    this.invalid_handle("CloseHandle")?
```
YohDeadfall, line 167: According to https://github.com/rust-lang/miri/issues/4015#issuecomment-2460207472.

src/shims/windows/thread.rs
```diff
@@ -59,14 +59,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
 
         let handle = this.read_scalar(handle_op)?;
         let timeout = this.read_scalar(timeout_op)?.to_u32()?;
 
         let thread = match Handle::from_scalar(handle, this)? {
-            Some(Handle::Thread(thread)) => thread,
+            Some(Handle::Thread(thread)) =>
+                match this.thread_id_try_from(thread) {
+                    Ok(thread) => thread,
+                    Err(_) => this.invalid_handle("WaitForSingleObject")?,
```
YohDeadfall, line 72: According to https://github.com/rust-lang/miri/issues/4015#issuecomment-2460207472.

tests/pass-dep/libc/pthread-threadname.rs
```diff
@@ -199,4 +199,28 @@ fn main() {
         .unwrap()
         .join()
         .unwrap();
+
+    // Now set the name for a non-existing thread and verify error codes.
```
RalfJung, line 203:
```suggestion
    // Now set the name for a non-existing thread and verify error codes.
    // (FreeBSD does not have a thread ID argument.)
```

YohDeadfall: Nope, it's not because FreeBSD having no thread ID arguments. It has it, but has no result.

RalfJung: Ah okay. Whatever the reason is, please add a comment explaining it. :)

YohDeadfall: Done.

RalfJung:
Awesome, thanks. :)  Please squash. :D

PR #4006, opened 2024-11-01
Fix #3846 properly, so that subtrees can be skipped again

### LINKED ISSUE #3846: TB: Missed UB in the "optimization" that allows skipping subtrees

JoJoDeveloping:
Issue #3846 was "fixed" by PR #3847, but this fix was a hotfix. After careful thinking about the invariants at play there, the more proper solution is to instead do some extra magic during retags. This PR re-adds the disabled functionality, and ensures that default-lazy new nodes introduced by a retag outside that retag's range do not break anything, by having such retags reset the state that tracks when subtrees can be skipped.

The details can be seen in the code comments.

Here's the difference in running times for the benchmark suite. Note that it's a log graph, since the differences are so large it would otherwise be unreadable: 

origin_master and johannes_tb-fix-3846-retag


src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -640,6 +701,56 @@ impl<'tcx> Tree {
...
+                    // if there is no permission yet, it is still "default lazy".
+                    // in that case, we would need to compare with that node's original default permission.
+                    // But that depended on whether that node was protected when it was created, which is no
+                    // longer the case now.
+                    // Anyways, since the node is "default", its latest recorded foreign access will default to `None`,
+                    // but we might need to continue weakening the parent.
+                    // So we return `true` here, which is definitely correct, but maybe not fully optimal.
+                    // But in cases where we are imprecise, the iteration stops at the next initialized parent anyway,
+                    // so it is likely not a huge loss in practice.
+                    None => true,
```
JoJoDeveloping, line 743: I'm still somewhat unhappy with this. In cases where a node is mostly default-initialized, this code here can become significant. A better solution might be to track the "default last foreign access" for not-yet-materialized nodes, similar to how we track the default permission for these. And then update that in here.

RalfJung:
Well those are some solid wins. :)

(How did you make those graphs?)

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -87,6 +87,28 @@ impl PermissionPriv {
...
+    /// Returns the strongest foreign action this node survives (without change),
+    /// where `prot` indicates if it is protected.
+    /// They are ordered as `None` < `Some(Read)` < `Some(Write)`, in order of power.
+    pub fn strongest_survivable_foreign_action(&self, prot: bool) -> Option<AccessKind> {
+        match self {
+            // The read will make it conflicted, so it is not invariant under it.
+            ReservedFrz { conflicted } if prot && !conflicted => None,
+            // Otherwise, foreign reads do not affect it
+            ReservedFrz { .. } => Some(AccessKind::Read),
+            // Famously, ReservedIM survives foreign writes. It is never protected.
```
RalfJung, line 100: Might we worth a (debug) assertion of `!prot`.

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -87,6 +87,28 @@ impl PermissionPriv {
     fn compatible_with_protector(&self) -> bool {
         !matches!(self, ReservedIM)
     }
+
+    /// Returns the strongest foreign action this node survives (without change),
+    /// where `prot` indicates if it is protected.
+    /// They are ordered as `None` < `Some(Read)` < `Some(Write)`, in order of power.
+    pub fn strongest_survivable_foreign_action(&self, prot: bool) -> Option<AccessKind> {
```
RalfJung, line 94: This function can't actually be invoked from the outside since `PermissionPriv` is private, I think? So please remove the `pub`.

RalfJung:
> Not really--it's more that https://github.com/rust-lang/miri/pull/3847 was a solid loss 🙃 this just brings the performance back to where it used to be.

The previous version of this cheated by being wrong, so it still counts. ;)

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -207,21 +207,82 @@ impl LocationState {
...
             self.latest_foreign_access = None;
         }
     }
+
+    /// Check if the latest recorded foreign update needs to be updated after a retag.
+    /// Retags can violate the invariant on `LocationState::latest_foreign_access`,
+    /// which states that a node's children survive accesses at least as strong as that node itself.
+    /// While retags act as a read for the offsets specified in the retag, it is also present at the other
+    /// offsets, but marked as "lazy." Such lazy nodes can, however, still be affected by foreign accesses.
+    /// This would violate the invariants that children always survive accesses at least as strong as the last
```
RalfJung, line 233:
```suggestion
    /// This would violate the invariant that children always survive accesses at least as strong as the last
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -207,21 +207,82 @@ impl LocationState {
...
+    /// recorded foreign access at the parent.
+    /// To restore the invariant, we need to weaken the information stored in the parent, until it is weak enough
+    /// to encompass the newly added lazy node.
+    /// This function compares two accesses, and indicates if the parents needs to be updated.
```
RalfJung, line 237: should be either "parents need" or "parent needs"

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -207,21 +207,82 @@ impl LocationState {
...
+    /// `None` is weakest and `Some(Write)` is strongest.
+    pub fn foreign_access_requires_update(
+        last_recorded: Option<AccessKind>,
+        happening_now: Option<AccessKind>,
+    ) -> bool {
+        match (last_recorded, happening_now) {
+            // if the last access here was a child access, we need to do nothing
+            (None, _) => false,
+            // if the last access here was a foreign access, but our new child does not
+            // survive any foreign access, we must let the next foreign access through again
```
RalfJung, line 248:
```suggestion
            // If the last access here was a foreign access (`last_recorded` is `Some`),
            // but our new child does not survive any foreign access (`happening_now`,
            // is `None`), we must let the next foreign access through again.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -207,21 +207,82 @@ impl LocationState {
...
+    /// This would violate the invariants that children always survive accesses at least as strong as the last
+    /// recorded foreign access at the parent.
+    /// To restore the invariant, we need to weaken the information stored in the parent, until it is weak enough
+    /// to encompass the newly added lazy node.
+    /// This function compares two accesses, and indicates if the parents needs to be updated.
+    /// It returns `true` iff `last_recorded` is stronger than `happening_now`, where
+    /// `None` is weakest and `Some(Write)` is strongest.
+    pub fn foreign_access_requires_update(
+        last_recorded: Option<AccessKind>,
+        happening_now: Option<AccessKind>,
```
RalfJung, line 242: `happening_now` is the strongest access that the new node would survive, isn't it? I am confused by the name.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -207,21 +207,82 @@ impl LocationState {
...
+            // last access was a write, the child does not survive writes -> reset it to read
+            (Some(AccessKind::Write), Some(AccessKind::Read)) => true,
+        }
+    }
+
+    /// Restores the "children are stronger" invariant of `latest_foreign_access` after a retag.
+    /// See also `foreign_access_requires_update`.
+    /// Retags can violate the invariant on `latest_foreign_access`,
+    /// which states that a node's children survive accesses at least as strong as that node itself.
+    /// While retags act as a read for the offsets specified in the retag, it is also present at the other
```
RalfJung, line 265:
```suggestion
    /// While retags act as a read for the offsets specified in the retag, the new tag is also present at the other
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -207,21 +207,82 @@ impl LocationState {
...
             self.latest_foreign_access = Some(access_kind);
         } else {
             self.latest_foreign_access = None;
         }
     }
+
+    /// Check if the latest recorded foreign update needs to be updated after a retag.
+    /// Retags can violate the invariant on `LocationState::latest_foreign_access`,
+    /// which states that a node's children survive accesses at least as strong as that node itself.
+    /// While retags act as a read for the offsets specified in the retag, it is also present at the other
```
RalfJung, line 231:
```suggestion
    /// While retags act as a read for the offsets specified in the retag, the new tag is also present at the other
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -207,21 +207,82 @@ impl LocationState {
...
+        }
+    }
+
+    /// Restores the "children are stronger" invariant of `latest_foreign_access` after a retag.
+    /// See also `foreign_access_requires_update`.
+    /// Retags can violate the invariant on `latest_foreign_access`,
+    /// which states that a node's children survive accesses at least as strong as that node itself.
+    /// While retags act as a read for the offsets specified in the retag, it is also present at the other
+    /// offsets, but marked as "lazy." Such lazy nodes can, however, still be affected by foreign accesses.
+    /// This would violate the invariants that children always survive accesses at least as strong as the last
```
RalfJung, line 267:
```suggestion
    /// This would violate the invariant that children always survive accesses at least as strong as the last
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -207,21 +207,82 @@ impl LocationState {
...
+    /// also needs to be updated. If it returns `false`, then the invariant is restored globally, since all of `self`'s
+    /// parents are already weaker than `self` itself.
+    fn reset_last_foreign_access_after_retag(
+        &mut self,
+        strongest_survivable: Option<AccessKind>,
+    ) -> bool {
+        let needs_update =
+            Self::foreign_access_requires_update(self.latest_foreign_access, strongest_survivable);
+        if needs_update {
+            self.latest_foreign_access = strongest_survivable;
```
RalfJung, line 282: Might be worth a debug assertion that `strongest_survivable` is "less than" `latest_foreign_access`.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -640,6 +701,56 @@ impl<'tcx> Tree {
         interp_ok(())
     }
 
+    /// Restores the "children are stronger" invariant of `latest_foreign_access` after a retag.
+    /// See also `reset_last_foreign_access_after_reborrow`.
+    /// Retags can violate the invariant on `latest_foreign_access`,
+    /// which states that a node's children survive accesses at least as strong as that node itself.
+    /// While retags act as a read for the offsets specified in the retag, it is also present at the other
```
RalfJung, line 708: This same comment is copied 3 times it seems... seems better to instead reference a common pace where this is explained once.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -640,6 +701,56 @@ impl<'tcx> Tree {
...
+            current = next;
+            // record whether we did any change (if not, the invariant is restored).
+            let any_change = self.rperms.iter_mut_all().any(|(_, map)| {
+                match map.get_mut(current) {
+                    // if there is a permission, ensure that it's latest recorded foreign access is not stronger
+                    // than the `strongest_survivable` by the newly inserted child.
+                    // Returns `true` if it changed something, and in that case, we need to continue with the parent.
+                    Some(perm) => perm.reset_last_foreign_access_after_retag(strongest_survivable),
+                    // if there is no permission yet, it is still "default lazy".
+                    // in that case, we would need to compare with that node's original default permission.
```
RalfJung, line 735:
```suggestion
                    // In that case, we would need to compare with that node's original default permission.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -640,6 +701,56 @@ impl<'tcx> Tree {
...
+        while let Some(next) = self.nodes.get(current).unwrap().parent {
+            current = next;
+            // record whether we did any change (if not, the invariant is restored).
+            let any_change = self.rperms.iter_mut_all().any(|(_, map)| {
+                match map.get_mut(current) {
+                    // if there is a permission, ensure that it's latest recorded foreign access is not stronger
+                    // than the `strongest_survivable` by the newly inserted child.
+                    // Returns `true` if it changed something, and in that case, we need to continue with the parent.
+                    Some(perm) => perm.reset_last_foreign_access_after_retag(strongest_survivable),
+                    // if there is no permission yet, it is still "default lazy".
```
RalfJung, line 734:
```suggestion
                    // If there is no permission yet, it is still "default lazy".
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -640,6 +701,56 @@ impl<'tcx> Tree {
...
+    ) {
+        let mut current = self.tag_mapping.get(&child_tag).unwrap();
+        let strongest_survivable = new_perm.strongest_survivable_foreign_action(prot);
+        // We walk the tree upwards, until the invariant is restored
+        while let Some(next) = self.nodes.get(current).unwrap().parent {
+            current = next;
+            // record whether we did any change (if not, the invariant is restored).
+            let any_change = self.rperms.iter_mut_all().any(|(_, map)| {
+                match map.get_mut(current) {
+                    // if there is a permission, ensure that it's latest recorded foreign access is not stronger
```
RalfJung, line 730:
```suggestion
                    // If there is a permission, ensure that its latest recorded foreign access is not stronger
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -640,6 +701,56 @@ impl<'tcx> Tree {
...
+                    // than the `strongest_survivable` by the newly inserted child.
+                    // Returns `true` if it changed something, and in that case, we need to continue with the parent.
+                    Some(perm) => perm.reset_last_foreign_access_after_retag(strongest_survivable),
+                    // if there is no permission yet, it is still "default lazy".
+                    // in that case, we would need to compare with that node's original default permission.
+                    // But that depended on whether that node was protected when it was created, which is no
+                    // longer the case now.
```
RalfJung, line 737: I don't understand this sentence. What is no longer the case now?

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: `happening_now` is what we are weakening everything to. I agree the name is confusing. How about `at_most` instead?

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: I don't understand this suggestion. What do you want to assert, and where? Note that the whole point of this `if` is to figure out whether `strongest_survivable` is "less than" `latest_foreign_access`. So if you `debug_assert` it in the wrong place, you get an assertion failure, and if you `debug_assert` it in the right place, you're just writing code of shape

```
if foo {
  debug_assert!(foo);
```

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: Indeed. But unfortunately `Option<AccessKind>` has no obvious place where one would add such an annotation.
Note that `Option<AccessKind>` already has the here-suggested order "natively" since it implements `Ord`, and it turns out that the auto-derived order is the correct one here (this is not entirely accidental). But relying on that seems wrong, and the code does *not* do that currently. Perhaps one could introduce a newtype?

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: The protector might have in the meantime gone away.

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: > How about at_most instead?

Is this entire logic just doing `self.latest_foreign_access.min(strongest_survivable)`? Maybe we should use something more like that.

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: I think I am asking for a `Ord` implementation on `Option<AccessKind>` and then this entire thing can be written much simpler as `self.latest_foreign_access = self.latest_foreign_access.min(strongest_survivable)`.

RalfJung: The `Ord` on `Option` seems to be such that `None < Some(_)`, so if you implement `Ord` on `AccessKind` correctly this should just work.

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: > But relying on that seems wrong, 

Why? Your comments treat the type as having an order. And it does even actually have that order. We should do that.

If you feel uneasy relying on it, add a little `#[test]` function asserting that the order is the way we want it to be.

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: So the point is if the protector disappeared the "strongest idempotent access" might become stronger but currently the code doesn't exploit that?

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: I actually used that at one point in time, but removed it because I figured you'd not like it :laughing:

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: Thinking about it a bit more, I think that without the protector the strongest idempotent foreign access (SIPA) only becomes stronger. Which means that one could (I think) always compute that SIPA at this point with `protector := false`, and it's still correct, because it will over-approximate.

RalfJung: > without the protector the strongest idempotent foreign access (SIPA) only becomes stronger.

That's easy to test exhaustively.

However, isn't the *smaller* ("weaker") access the conservative choice? I am confused.

JoJoDeveloping: Yes, potentially. I'm often confused by the order.

JoJoDeveloping: I think a better solution than over-approximating that it's still protected is to actually store the initial SIPA similar to how we store the initial permission, and then modify that here.

RalfJung: Shouldn't it be "SIFA"? ;)

I'm afraid I don't understand why the initial SIFA is what we need here.

JoJoDeveloping: because it's the one that was used to modify that field in the parents, i.e. the one that the invariant relates to. If you now have a new child, you want to make an inference about the parents, whether they're already weak enough for you to stop iterating. So you need to know a lower (or is it upper) bound on the parents, which (in the case of still-to-be-default-initialized fields) is the the thing you used when you inserted it. Which is the SIFA for the permission it had when you inserted it.

RalfJung: Oh this is just about stopping the traversal upwards, not about changing anything in what we store in the current node?

JoJoDeveloping: Yes. Currently, the traversal continues. But I'd have hoped one could be smarter here, since this will be hit quite often (basically whenever you only have a reference to a sub-range).

JoJoDeveloping:
I pushed some quite large changes, which restructure a lot and rewrite most of the comments. Please have a look at it again, starting in the new file `foreign_access_skipping.rs` :slightly_smiling_face:

src/borrow_tracker/tree_borrows/foreign_access_skipping.rs
```diff
@@ -0,0 +1,94 @@
...
+            *self = strongest_allowed;
+            true
+        } else {
+            false
+        }
+    }
+}
+
+mod tests {
+    use super::LastForeignAccess;
```
JoJoDeveloping, line 87: I get a warning during build that this `use` is not necessary. But if I remove it, I get an error:

```
error[E0433]: failed to resolve: use of undeclared type `LastForeignAccess`
  --> src/borrow_tracker/tree_borrows/foreign_access_skipping.rs:89:17
   |
89 |         assert!(LastForeignAccess::LocalAccess < LastForeignAccess::ForeignRead);
   |                 ^^^^^^^^^^^^^^^^^ use of undeclared type `LastForeignAccess`
   |
help: consider importing this enum
   |
87 +     use crate::borrow_tracker::tree_borrows::foreign_access_skipping::LastForeignAccess;
   |
```

(where again, it then tells me the inserted `use` is unused)

src/borrow_tracker/tree_borrows/perms.rs
JoJoDeveloping: Done

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: Indeed, this seems to be the issue in `zip-equal` where a better handling of this makes things go 28% more brrr.

src/borrow_tracker/tree_borrows/foreign_access_skipping.rs
```diff
@@ -0,0 +1,95 @@
...
+/// This enum represents the kinds of values we store:
+/// `LocalAccess` means that the last access was local, and thus the next foreign access needs to happen, it can not be skipped.
+/// `ForeignRead` means that we are idempotent under foreign reads, but not (yet) under foreign writes.
+/// `ForeignWrite` means that we are idempotent under foreign writes. In other words, this node can be skipped for all foreign accesses.
+/// Due to the subtree property of traversals, this also means that all our children can be skipped, which makes this optimization
+/// worthwhile. In order for this to work, the invariant for multiple nodes in a tree is that at each location, the SIFA at each
+/// child must be stronger than that at the parent. However, note that this is quite invariant, due to retags inserting nodes across
+/// the entire range, which can violate this invariant without explicitly causing a local access which would reset things.
+/// So this needs to be done manually, thus `ensure_no_stronger_than`, and also the test `permission_sifa_is_correct` in `perms.rs`
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Default)]
```
RalfJung, line 17: Seems worth mentioning that since we derive `PartialOrd`, the order in which we write the elements here is crucial! They must be order from weakest to strongest.

src/borrow_tracker/tree_borrows/foreign_access_skipping.rs
```diff
@@ -0,0 +1,95 @@
...
+        matches!(self, LastForeignAccess::LocalAccess)
+    }
+
+    /// Returns true if this access is foreign, i.e. not local.
+    pub fn is_foreign(self) -> bool {
+        !self.is_local()
+    }
+
+    /// Constructs a foreign access from an `AccessKind`
+    pub fn foreign(acc: AccessKind) -> LastForeignAccess {
```
RalfJung, line 59:
```suggestion
    pub fn from_foreign(acc: AccessKind) -> LastForeignAccess {
```

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -87,6 +88,27 @@ impl PermissionPriv {
     fn compatible_with_protector(&self) -> bool {
         !matches!(self, ReservedIM)
     }
+
+    /// See `foreign_access_skipping.rs`. Compotes the SIFA of a permission.
```
RalfJung, line 92:
```suggestion
    /// See `foreign_access_skipping.rs`. Computes the SIFA of a permission.
```

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -87,6 +88,27 @@ impl PermissionPriv {
...
+    fn strongest_idempotent_foreign_access(&self, prot: bool) -> LastForeignAccess {
+        match self {
+            // The read will make it conflicted, so it is not invariant under it.
```
RalfJung, line 95:
```suggestion
            // A protected non-conflicted reserve will become conflicted under a foreign read,
            // and is hence not idempotent under it.
```

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -87,6 +88,27 @@ impl PermissionPriv {
...
+            ReservedFrz { conflicted } if prot && !conflicted => LastForeignAccess::LocalAccess,
+            // Otherwise, foreign reads do not affect it
```
RalfJung, line 97:
```suggestion
            // Otherwise, foreign reads do not affect reserved.
```

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -634,6 +663,33 @@ mod propagation_optimization_checks {
         }
     }
 
+    #[test]
+    #[rustfmt::skip]
+    fn permission_sifa_is_correct() {
+        // Tests that `strongest_idempotent_foreign_access` is correct. See `foreign_access_skipping.rs`.
+        for perm in PermissionPriv::exhaustive() {
+            // Assert that adding a protector makes is less idempotent.
```
RalfJung, line 671:
```suggestion
            // Assert that adding a protector makes it less idempotent.
```

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -634,6 +663,33 @@ mod propagation_optimization_checks {
...
+                assert!(perm.strongest_idempotent_foreign_access(true) <= perm.strongest_idempotent_foreign_access(false));
+            }
+            for prot in bool::exhaustive() {
+                if prot {
+                    precondition!(perm.compatible_with_protector());
+                }
+                let access = perm.strongest_idempotent_foreign_access(prot);
+                // We now assert it is idempotent, and never causes UB.
+                // First, if the SIFA includes foreign reads, assert it is idempotent under foreign reads.
+                if LastForeignAccess::ForeignRead <= access {
```
RalfJung, line 682:
```suggestion
                if access >= LastForeignAccess::ForeignRead {
```
same below

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -634,6 +663,33 @@ mod propagation_optimization_checks {
...
+                    assert_eq!(perm, transition::perform_access(AccessKind::Read, AccessRelatedness::DistantAccess, perm, prot).unwrap());
```
RalfJung, line 683: What's a "DistantAccess"? Is that the same as "foreign"? If yes, we should rename. (Can be a separate PR.)

src/borrow_tracker/tree_borrows/tree/tests.rs
```diff
@@ -598,13 +602,15 @@ mod spurious_read {
         let source = LocStateProtPair {
             xy_rel: RelPosXY::MutuallyForeign,
             x: LocStateProt {
-                state: LocationState::new_init(Permission::new_frozen()),
+                // For the tests, the strongest idempotent foreign access does not matter, so we use `Default::default`
+                state: LocationState::new_init(Permission::new_frozen(), Default::default()),
```
RalfJung, line 606:
```suggestion
                state: LocationState::new_init(Permission::new_frozen(), LastForeignAccess::default()),
```
To make it more clear which type things have. (Same below.)

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -45,28 +46,27 @@ pub(super) struct LocationState {
...
-    /// Strongest foreign access whose effects have already been applied to
-    /// this node and all its children since the last child access.
-    /// This is `None` if the most recent access is a child access,
-    /// `Some(Write)` if at least one foreign write access has been applied
-    /// since the previous child access, and `Some(Read)` if at least one
-    /// foreign read and no foreign write have occurred since the last child access.
-    latest_foreign_access: Option<AccessKind>,
+    /// See `foreign_access_skipping.rs`
+    /// Stores the strongest idempotent foreign access for this location.
+    /// Invariant is that at all children, the strongest idempotent foreign access is stronger (greater).
```
RalfJung, line 51: This is an approximation, right? It's okay for this to always be `LocalAccess` even if it would be idempotent under foreign reads? That should be documented.

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: This still applies.

src/borrow_tracker/tree_borrows/foreign_access_skipping.rs
```diff
@@ -0,0 +1,95 @@
...
+/// `LocalAccess` means that the last access was local, and thus the next foreign access needs to happen, it can not be skipped.
+/// `ForeignRead` means that we are idempotent under foreign reads, but not (yet) under foreign writes.
+/// `ForeignWrite` means that we are idempotent under foreign writes. In other words, this node can be skipped for all foreign accesses.
+/// Due to the subtree property of traversals, this also means that all our children can be skipped, which makes this optimization
+/// worthwhile. In order for this to work, the invariant for multiple nodes in a tree is that at each location, the SIFA at each
+/// child must be stronger than that at the parent. However, note that this is quite invariant, due to retags inserting nodes across
+/// the entire range, which can violate this invariant without explicitly causing a local access which would reset things.
+/// So this needs to be done manually, thus `ensure_no_stronger_than`, and also the test `permission_sifa_is_correct` in `perms.rs`
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Default)]
+pub enum LastForeignAccess {
```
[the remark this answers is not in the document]
JoJoDeveloping: I felt uncomfortable with using `strongest_idempotent_foreign_access` in field names, can I shorten it to `sifa`? Apart from that, the field used to be called `latest_foreign_access` but has not changed its semantics. So you're saying it should have had a different name already a year ago?

src/borrow_tracker/tree_borrows/foreign_access_skipping.rs
```diff
@@ -0,0 +1,95 @@
...
+
+/// To speed up tree traversals, we want to skip traversing subtrees when we know the traversal will have no effect.
+/// This is often the case for foreign accesses, since usually foreign accesses happen several times in a row, but also
+/// foreign accesses are idempotent. In particular, see tests `foreign_read_is_noop_after_foreign_write` and `all_transitions_idempotent`.
+/// Thus, for each node we keep track of the "strongest idempotent foreign access," (SIFA) i.e. which foreign access can be skipped.
+/// This enum represents the kinds of values we store:
+/// `LocalAccess` means that the last access was local, and thus the next foreign access needs to happen, it can not be skipped.
+/// `ForeignRead` means that we are idempotent under foreign reads, but not (yet) under foreign writes.
+/// `ForeignWrite` means that we are idempotent under foreign writes. In other words, this node can be skipped for all foreign accesses.
+/// Due to the subtree property of traversals, this also means that all our children can be skipped, which makes this optimization
```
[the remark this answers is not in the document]
JoJoDeveloping: If you get visited, your children also get visited.

src/borrow_tracker/tree_borrows/foreign_access_skipping.rs
```diff
@@ -0,0 +1,95 @@
...
+/// worthwhile. In order for this to work, the invariant for multiple nodes in a tree is that at each location, the SIFA at each
+/// child must be stronger than that at the parent. However, note that this is quite invariant, due to retags inserting nodes across
+/// the entire range, which can violate this invariant without explicitly causing a local access which would reset things.
+/// So this needs to be done manually, thus `ensure_no_stronger_than`, and also the test `permission_sifa_is_correct` in `perms.rs`
```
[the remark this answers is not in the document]
JoJoDeveloping: Yes, seems like I dropped a `not` in the first sentence.

I don't get what is so difficult about the last sentence. We reset the invariant "manually" by calling the `ensure_no_stronger_than` function (and related ones). You could call that one `manually_reset_things_so_that_the_invariant_holds_again_after_retags_inserted_nodes_silently` perhaps?

src/borrow_tracker/tree_borrows/foreign_access_skipping.rs
JoJoDeveloping: there's a test for that.

src/borrow_tracker/tree_borrows/perms.rs
JoJoDeveloping: no, it's a cousin access. The `AccessRelatedness` enum has 4 constructors (child, self, cousin, parent), but for almost all cases, two of them (child+self, cousin+parent) are treated the same. The only exception is in protector end writes.

But here we need to choose a concrete inhabitant. We could also pick `Parent`, i.e. a different member of the same equivalence class.

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: It's always OK for this field to be `LocalAccess`. Also for the one in `Node`. That would be as if this whole optimization did not exist.

src/borrow_tracker/tree_borrows/foreign_access_skipping.rs
RalfJung: Tests don't replace comments.

src/borrow_tracker/tree_borrows/perms.rs
RalfJung: Ah, we should rename to `CousinAccess` then.

And please add a comment here explaining what you just said.

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: Okay, then please explain that. Currently the definition of SIFA makes it sound like an exact bound.

src/borrow_tracker/tree_borrows/foreign_access_skipping.rs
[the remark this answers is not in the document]
RalfJung: I'm not interested in discussing which name it should have had a year ago. ;)

The "strongest" part isn't actually relevant for correctness, is it? So IMO "idempotent_foreign_access" should suffice. "local access" doesn't make sense as a variant name, so I would propose
```rust
enum IdempotentForeignAccess { None, Read, Write }
```

src/borrow_tracker/tree_borrows/foreign_access_skipping.rs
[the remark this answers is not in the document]
JoJoDeveloping: > what role does that test play?

which test?

JoJoDeveloping:
better now^^`

src/borrow_tracker/tree_borrows/foreign_access_skipping.rs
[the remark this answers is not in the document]
JoJoDeveloping: It tests that the permission is idempotent. It does not test that this is strongest, but this is easily seen by trying to make it stronger and then observing that the test fails.

src/borrow_tracker/tree_borrows/foreign_access_skipping.rs
```diff
@@ -0,0 +1,106 @@
...
+/// *the entire subtree* rooted at that node can be skipped. In order for this to work, we maintain the global invariant that at
+/// each location, the SIFA at each child must be stronger than that at the parent. For normal reads and writes, this is easily accomplished by
+/// tracking each foreign access as it occurs, so that then the next access can be skipped. This also obviously maintains the invariant, because
+/// if a node undergoes a foreign access, then all its children also see this as a foreign access. However, the invariant is broken during retags,
+/// because retags act across the entire allocation, but only emit a read event across a specific range. This means that for all nodes outside that
+/// range, the invariant is potentially broken, since a new child with a weaker SIFA is inserted. Thus, during retags, special care is taken to
+/// "manually" reset the parent's SIFA to be at least as strong as the new child's. This is accomplished with the `ensure_no_stronger_than` method.
+///
+/// Note that we derive Ord and PartialOrd such that None < Read < Write. Do not change that order. See the `test_order` test.
```
RalfJung, line 25:
```suggestion
/// Note that we derive Ord and PartialOrd, so the order in which variants are listed below matters:
/// None < Read < Write. Do not change that order. See the `test_order` test.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -45,28 +46,32 @@ pub(super) struct LocationState {
...
-    /// foreign read and no foreign write have occurred since the last child access.
-    latest_foreign_access: Option<AccessKind>,
+    /// See `foreign_access_skipping.rs`
+    /// Stores an idempotent foreign access for this location. This should be the strongest such access,
+    /// except that of course this is just an optimization and so it is expected that this is not always
+    /// necessarily the strongest one, but just an approximation. All that is important for correctness
+    /// is that it is never to strong. Nonetheless, we try to make it as strong as possible, so that the
+    /// optimization actually makes things faster.
+    /// A correctness invariant is that at all children of this node, their recorded
+    /// idempotent foreign access is stronger (greater).
```
RalfJung, line 56:
```suggestion
    /// See `foreign_access_skipping.rs`.
    /// Stores an idempotent foreign access for this location and its children.
    /// For correctness, this must not be too strong, and the recorded idempotent foreign access
    /// of all children must be at least as strong as this. For performance, it should be as strong as possible.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -617,29 +607,77 @@ impl<'tcx> Tree {
...
+        loop {
+            let current_node = self.nodes.get_mut(current).unwrap();
+            // record whether we did any change (if not, the invariant is restored).
+            let any_change = self.rperms.iter_mut_all().any(|(_, map)| {
+                match map.get_mut(current) {
+                    // if there is a permission, ensure that its latest recorded foreign access is not stronger
+                    // than the `strongest_survivable` by the newly inserted child.
+                    // Returns `true` if it changed something, and in that case, we need to continue with the parent.
+                    Some(perm) =>
+                        perm.reset_idempotent_foreign_access_after_retag(strongest_allowed),
```
RalfJung, line 660: I think it'd actually be better to inline that function here. That makes things more symmetric, since you also directly call `ensure_no_stronger_than` a few lines below.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -617,29 +607,77 @@ impl<'tcx> Tree {
...
+    /// See `foreign_access_skipping` and `new_child`.
+    fn update_last_accessed_after_retag(
+        &mut self,
+        mut current: UniIndex,
+        strongest_allowed: IdempotentForeignAccess,
+    ) {
+        // We walk the tree upwards, until the invariant is restored
+        loop {
+            let current_node = self.nodes.get_mut(current).unwrap();
+            // record whether we did any change (if not, the invariant is restored).
```
RalfJung, line 653:
```suggestion
            // Call `ensure_no_stronger_than` on all SIFAs for this node: the per-location SIFA, as well
            // as the default SIFA for not-yet-initialized locations.
            // Record whether we did any change; if not, the invariant is restored and we can stop the traversal.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -617,29 +607,77 @@ impl<'tcx> Tree {
...
+            let any_change = self.rperms.iter_mut_all().any(|(_, map)| {
```
RalfJung, line 654: Is the use of `any` and `||` here really correct? The moment the first `true` is returned, this will abort the further traversal! Why is short-circuiting correct here?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -617,29 +607,77 @@ impl<'tcx> Tree {
...
+                match map.get_mut(current) {
+                    // if there is a permission, ensure that its latest recorded foreign access is not stronger
+                    // than the `strongest_survivable` by the newly inserted child.
+                    // Returns `true` if it changed something, and in that case, we need to continue with the parent.
+                    Some(perm) =>
+                        perm.reset_idempotent_foreign_access_after_retag(strongest_allowed),
+                    // All which are not yet lazily initialized use the Node::default_initial_foreign_access, which is handled below
+                    // So we return false, which is the identity in `any()`
```
RalfJung, line 662:
```suggestion
                    // All which are not yet lazily initialized use the Node::default_initial_foreign_access, which is handled below.
                    // So we return false, which is the identity in `any()`.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -617,29 +607,77 @@ impl<'tcx> Tree {
...
+        mut current: UniIndex,
+        strongest_allowed: IdempotentForeignAccess,
+    ) {
+        // We walk the tree upwards, until the invariant is restored
+        loop {
+            let current_node = self.nodes.get_mut(current).unwrap();
+            // record whether we did any change (if not, the invariant is restored).
+            let any_change = self.rperms.iter_mut_all().any(|(_, map)| {
+                match map.get_mut(current) {
+                    // if there is a permission, ensure that its latest recorded foreign access is not stronger
```
RalfJung, line 656:
```suggestion
                    // If there is a permission, ensure that its latest recorded foreign access is not stronger
```

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: Oh, good catch. Indeed, short-circuiting is not correct here.

JoJoDeveloping: I rewrote large parts of that function, please have a look at it again.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -650,24 +637,25 @@ impl<'tcx> Tree {
...
-                        perm.reset_idempotent_foreign_access_after_retag(strongest_allowed),
-                    // All which are not yet lazily initialized use the Node::default_initial_foreign_access, which is handled below
-                    // So we return false, which is the identity in `any()`
-                    None => false,
+            // Call `ensure_no_stronger_than` on all SIFAs for this node: the per-location SIFA, as well
+            // as the default SIFA for not-yet-initialized locations.
+            // Record whether we did any change; if not, the invariant is restored and we can stop the traversal.
+            let mut any_change = false;
+            for (_, map) in self.rperms.iter_mut_all() {
+                if let Some(perm) = map.get_mut(current) {
```
RalfJung, line 645:
```suggestion
                // Check if this node has a state for this location (or range of locations).
                if let Some(perm) = map.get_mut(current) {
```

RalfJung (commented):
Just one more comment nit. :)  Then please squash the commits together (with `--keep-rebase` to keep the diff nice)

JoJoDeveloping:
I have squashed everything. It might be prudent to look over everything again once more.

RalfJung:
Why? I looked over everything last time, and I can see there's no diff.

Thanks a lot, looks great!

PR #4008, opened 2024-11-01
[TB Optimization] Skip subtrees based on the subtree's root node's permissions

JoJoDeveloping:
In #4006, we re-added the functionality for skipping subtrees. It turns out that just skipping subtrees based on their last recorded access is imprecise. In certain cases, we know we can skip subtrees purely based on the root's current permission, without having to track the last access. Specifically:
* Disabled nodes can always be skipped, since the whole subtree must necessarily be invariant under all foreign accesses
* Frozen nodes an be skipped for foreign reads, since the whole subtree must necessarily be invariant under all foreign reads.

Note that this PR loosens the notion of "invariant" a bit. For example, it is possible that there is a `Reserved` protected node that is a child of a `Frozen` node. When that undergoes a foreign read, it becomes conflicted. If we skip accessing that subtree, it no longer does become conflicted.
The reason this is still OK is that the only effect of this conflictedness is blocking child write accesses. But such accesses are already blocked by the `Frozen` node further up the tree. So no UB is missed, all that happens is that diagnostics are triggered at a different node. 

For more detailed analysis of why this is correct, see the in-code comments.

Here is a performance analysis, comparing this PR's improvements with that of #4006:

performance comparison

As in #4006, this is a log graph. The blue line shows performance without #4006, red is with the re-added optimization of #4006, yellow is this PR (which is stacked on top of #4006), and green is just the changes proposed here, but with the "latest foreign access tracking" machinery of #4006 removed. As can be seen, having both combined gives the greatest performance.

### DIFF
--- a/src/borrow_tracker/tree_borrows/tree.rs
+++ b/src/borrow_tracker/tree_borrows/tree.rs
@@ -153,8 +153,31 @@ impl LocationState {
     ) -> ContinueTraversal {
         if rel_pos.is_foreign() {
             let happening_now = IdempotentForeignAccess::from_foreign(access_kind);
-            let new_access_noop =
+            let mut new_access_noop =
                 self.idempotent_foreign_access.can_skip_foreign_access(happening_now);
+            if self.permission.is_disabled() {
+                // A foreign access to a `Disabled` tag will have almost no observable effect.
+                // It's a theorem that `Disabled` node have no protected initialized children,
+                // and so this foreign access will never trigger any protector.
+                // (Intuition: You're either protected initialized, and thus can't become Disabled
+                // or you're already Disabled protected, but not initialized, and then can't
+                // become initialized since that requires a child access, which Disabled blocks.)
+                // Further, the children will never be able to read or write again, since they
+                // have a `Disabled` parent. So this only affects diagnostics, such that the
+                // blocking write will still be identified directly, just at a different tag.
+                new_access_noop = true;
+            }
+            if self.permission.is_frozen() && access_kind == AccessKind::Read {
+                // A foreign read to a `Frozen` tag will have almost no observable effect.
+                // It's a theorem that `Frozen` nodes have no active children, so all children
+                // already survive foreign reads. Foreign reads in general have almost no
+                // effect, the only further thing they could do is make protected `Reserved`
+                // nodes become conflicted, i.e. make them reject child writes for the further
+                // duration of their protector. But such a child write is already rejected
+                // because this node is frozen. So this only affects diagnostics, but the
+                // blocking read will still be identified directly, just at a different tag.
+                new_access_noop = true;
+            }
             if new_access_noop {
                 // Abort traversal if the new access is indeed guaranteed
                 // to be noop.
--- a/src/borrow_tracker/tree_borrows/perms.rs
+++ b/src/borrow_tracker/tree_borrows/perms.rs
@@ -237,6 +237,10 @@ impl Permission {
     pub fn is_active(&self) -> bool {
         self.inner == Active
     }
+    /// Check if `self` is the never-allow-writes-again state of a pointer (is `Frozen`).
+    pub fn is_frozen(&self) -> bool {
+        self.inner == Frozen
+    }
 
     /// Default initial permission of the root of a new tree at inbounds positions.
     /// Must *only* be used for the root, this is not in general an "initial" permission!
--- a/tests/fail/tree_borrows/subtree_traversal_skipping_diagnostics.rs
+++ b/tests/fail/tree_borrows/subtree_traversal_skipping_diagnostics.rs
@@ -1,0 +1,29 @@
+//@compile-flags: -Zmiri-tree-borrows -Zmiri-provenance-gc=0
+
+// Shows the effect of the optimization of #4008.
+// The diagnostics change, but not the error itself.
+
+// When this method is called, the tree will be a single line and look like this,
+// with other_ptr being the root at the top
+// other_ptr = root : Active
+// intermediary     : Frozen // an intermediary node
+// m                : Reserved
+fn write_to_mut(m: &mut u8, other_ptr: *const u8) {
+    unsafe {
+        std::hint::black_box(*other_ptr);
+    }
+    // In line 17 above, m should have become Reserved (protected) so that this write is impossible.
+    // However, that does not happen because the read above is not forwarded to the subtree below
+    // the Frozen intermediary node. This does not affect UB, however, because the Frozen that blocked
+    // the read already prevents any child writes.
+    *m = 42; //~ERROR: /write access through .* is forbidden/
+}
+
+fn main() {
+    let root = 42u8;
+    unsafe {
+        let intermediary = &root;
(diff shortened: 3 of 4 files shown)


RalfJung:
> all that happens is that diagnostics are triggered at a different node.

That is at least potentially confusing. :/ But maybe the fix here should be on the diagnostic side, not the core algorithm.

Does this depend on when the GC runs, or is it deterministic?

> As in https://github.com/rust-lang/miri/pull/4006, this is a log graph. The blue line shows performance without https://github.com/rust-lang/miri/pull/4006, red is with the re-added optimization of https://github.com/rust-lang/miri/pull/4006, yellow is this PR (which is stacked on top of https://github.com/rust-lang/miri/pull/4006), and green is just the changes proposed here, but with the "latest foreign access tracking" machinery of https://github.com/rust-lang/miri/pull/4006 removed. As can be seen, having both combined gives the greatest performance.

It seems like most benchmarks are unchanged by this PR (compared to just #4006), only a few of them benefit. big-allocs gets slightly worse.

Do you have evidence that this is beneficial on (a non-trivial fraction of) real-world code?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -185,6 +185,30 @@ impl LocationState {
...
+            if self.permission.is_disabled() {
+                // A foreign access to a `Disabled` tag will have almost no observable effect.
+                // It's a theorem that `Disabled` node have no protected initialized children,
+                // and so this foreign access will never trigger any protector.
+                // Further, the children will never be able to read or write again, since they
+                // have a `Disabled` parents. Even further, all children of `Disabled` are one
+                // of `ReservedIM`, `Disabled`, or a not-yet-accessed "lazy" permission thing.
+                // The two former are already invariant under all foreign accesses, and for
+                // the latter it does not really matter, since they can not be used/initialized
+                // due to having a protected parent. So this only affects diagnostics, but the
```
RalfJung, line 197: should be "disabled parent", right?

JoJoDeveloping:
> only a few of them benefit

Indeed. Intuitively, if you use lots of shared references, you benefit.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -185,6 +185,30 @@ impl LocationState {
                 // need to be applied to this subtree.
                 _ => false,
             };
+            if self.permission.is_disabled() {
+                // A foreign access to a `Disabled` tag will have almost no observable effect.
+                // It's a theorem that `Disabled` node have no protected initialized children,
```
RalfJung, line 190: That's not an obvious theorem -- can you give a brief argument?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -185,6 +185,30 @@ impl LocationState {
...
+                // and so this foreign access will never trigger any protector.
+                // Further, the children will never be able to read or write again, since they
+                // have a `Disabled` parents. Even further, all children of `Disabled` are one
```
RalfJung, line 193: The argument could end here, right? The permissions below don't matter since anyway no access is possible.

JoJoDeveloping:
> That is at least potentially confusing. :/ But maybe the fix here should be on the diagnostic side, not the core algorithm.

It depends. Arguably, the fact that it's because there's a frozen parent could be more clear than the fact that it's because you are reserved conflicted protected. But note that there's nothing the diagnostics can do to do things differently here, because the child node would not never become conflicted with this.



> Does this depend on when the GC runs, or is it deterministic?

It is deterministic.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -185,6 +185,30 @@ impl LocationState {
...
+                new_access_noop = true;
+            }
+            if self.permission.is_frozen() && access_kind == AccessKind::Read {
+                // A foreign read to a `Frozen` tag will have almost no observable effect.
+                // It's a theorem that `Frozen` nodes have no active children, so all children
+                // already survive foreign reads. Foreign reads in general have almost no
+                // effect, the only further thing they could do is make protected `Reserved`
+                // nodes become conflicted, i.e. make them reject child writes for the further
+                // duration of their protector. But such a child write is already rejected
+                // because this node is frozen. So this only affects diagnostics, but the
```
RalfJung, line 208: Is it possible to add a testcase that demonstrates the effect on diagnostics?

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: It's proven in Coq :stuck_out_tongue:.

The reason it holds is that to become disabled, you need to have a foreign write access happen. But that would have triggered any protected initialized nodes that are children of the node being disabled. And you can't have a new child of Disabled become initialized, because that would mean the to-be-initialized node has a child access, which is however blocked by the Disabled parent.

RalfJung:
Please rebase over master again so that github updates the diff for this PR.

JoJoDeveloping:
On it.

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: there is one now

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -153,8 +153,31 @@ impl LocationState {
...
         if rel_pos.is_foreign() {
             let happening_now = IdempotentForeignAccess::from_foreign(access_kind);
-            let new_access_noop =
+            let mut new_access_noop =
                 self.idempotent_foreign_access.can_skip_foreign_access(happening_now);
+            if self.permission.is_disabled() {
+                // A foreign access to a `Disabled` tag will have almost no observable effect.
+                // It's a theorem that `Disabled` node have no protected initialized children,
+                // and so this foreign access will never trigger any protector.
+                // (Intuition: You're either protected initialized, and thus can't become
```
RalfJung, line 162: "can't become" what? Missing a word.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -153,8 +153,31 @@ impl LocationState {
...
+                // or you're already Disabled protected, but not initialized, and then can't
+                // become initialized since that requires a child access, which Disabled blocks.)
+                // Further, the children will never be able to read or write again, since they
+                // have a `Disabled` parents. So this only affects diagnostics, but the
```
RalfJung, line 166:
```suggestion
                // have a `Disabled` parent. So this only affects diagnostics, but the
```

RalfJung (commented):
I'll have to carefully look at the test diff still to make sure I understand what happens here.

RalfJung:
Yeah okay makes sense, thanks for the detailed comments!

After fixing the typos, please squash this to 2 commits (leave the first that adds the test separate).

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: can't become Disabled. The word is there now.

PR #4016, opened 2024-11-06
Added epoll and eventfd for Android

### LINKED ISSUE #3996: Android: Add support for epoll, eventfd
Android is based on the Linux kernel, and its `libc` implementation has I/O notification APIs like [`epoll`](https://android.googlesource.com/platform/bionic/+/bbd39aa8e66dfe325732934d1aeb0da52b4750d5/libc/include/sys/epoll.h) or [`eventfd`](https://android.googlesource.com/platform/bionic/+/bbd39aa8e66dfe325732934d1aeb0da52b4750d5/libc/include/sys/eventfd.h). An implementation requires a bit of refactoring to avoid copy-pasting of [this](https://github.com/rust-lang/miri/blob/5bdbd272ed60f6cf8df4e903dfb2307c5f6e35e4/src/shims/unix/linux/foreign_items.rs#L56-L77).

YohDeadfall:
Fixes #3996.

I decided not to move `check_shim` into the implementations because all these functions are no more than just shims over `syscall`. Moreover, `eventfd` is even used by the `syscall` implementation.


src/shims/unix/linux/eventfd.rs
```diff
@@ -182,9 +182,6 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     fn eventfd(&mut self, val: &OpTy<'tcx>, flags: &OpTy<'tcx>) -> InterpResult<'tcx, Scalar> {
         let this = self.eval_context_mut();
 
-        // eventfd is Linux specific.
-        this.assert_target_os("linux", "eventfd");
-
```
tiif, line 187: If the intend is to only eventfd on linux and android only, instead of removing the assert, maybe we could use

```rust
if !matches!(&*this.tcx.sess.target.os, "linux" | "android") {
     panic!("`eventfd` should not be called on {}", this.tcx.sess.target.os);
}
```

YohDeadfall: The thing here is that the `eventfd` implementation resides in the `linux` module, so I see no reason in that assertion. If it would be for any `unix` with some exceptions, then yes, there's a good reason to keep the assertion.

YohDeadfall: @RalfJung, what's your opinion?

RalfJung: Seems like maybe it's a good time to introduce a `unix::linux_like` module, and then put some things there that both `unix::android` and `unix::linux` can import and use as helpers.

I agree that in that module, if things work on all linux-like systems, they don't need a further assertion.

tests/fail-dep/libc/libc-epoll-data-race.rs
```diff
@@ -3,6 +3,7 @@
 //! and therefore still report a data race for things that need to see the second event
 //! to be considered synchronized.
 //@only-target: linux
+//@only-target: android
```
RalfJung, line 6: This means it will only run on targets that are both linux and android. Which is no target.

You need to use this instead:
```suggestion
//@only-target: linux android
```

RalfJung:
Thanks!

PR #4018, opened 2024-11-07
Standardized variable names for InterpCx

### LINKED ISSUE #3958: "Standardize" name for variables of type `InterpCx`
When I made my first contribution to the main miri code base, I was for a brief moment (slightly) confused about the variables `this` and `ecx`, mostly used to store the result of `eval_context_{ref,mut}`. It would be nice if the name was consistent.

I see that `this` is most commonly used:

```
$ grep -r "this: " | wc -l
62
$ grep -r "let this = " | wc -l
341
```

```
$ grep -r "ecx: " | wc -l
107
$ grep -r "let ecx = " | wc -l
8
```

However, sometimes this is used to refer to other values, for example: 

```
borrow_tracker/tree_borrows/tree.rs:        this: &mut TreeVisitor<'_>,
```

So personally I would propose using `ecx` to store the interpreter context.

Obviously this is not really important, and adds no value to the users, but if you agree that this might improve the code base, I am glad to send a PR.

YohDeadfall:
Partly addresses #3958.

### DIFF
--- a/src/alloc_addresses/mod.rs
+++ b/src/alloc_addresses/mod.rs
@@ -111,8 +111,8 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
     // Returns the exposed `AllocId` that corresponds to the specified addr,
     // or `None` if the addr is out of bounds
     fn alloc_id_from_addr(&self, addr: u64, size: i64) -> Option<AllocId> {
-        let ecx = self.eval_context_ref();
-        let global_state = ecx.machine.alloc_addresses.borrow();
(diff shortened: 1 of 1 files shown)


RalfJung (approved):
These all look fine. :)

Why is this marked as "draft"?

YohDeadfall:
It doesn't touch `ecx:`/`this:` but the issue asks to address that too. So, as soon as there's a decision on argument names I will make another commit.

RalfJung:
You can also just make another PR then. Smaller PRs are generally better. :)

PR #4020, opened 2024-11-09
pthread-sync test: avoid confusing error when running with preemption

RalfJung:
Currently, panics can lead to data races, which is [confusing](https://github.com/rust-lang/miri/pull/4002#discussion_r1823148951).


RalfJung:
Ah damn, this needs thread parking which needs futexes...

tiif:
Thanks for fixing the confusing error! :)

PR #4026, opened 2024-11-10
miri: implement square root without relying on host floats

eduardosm:
https://github.com/rust-lang/miri/pull/3969#issuecomment-2440754959

### DIFF
--- a/src/math.rs
+++ b/src/math.rs
@@ -1,0 +1,164 @@
+use rand::Rng as _;
+use rand::distributions::Distribution as _;
+use rustc_apfloat::Float as _;
+use rustc_apfloat::ieee::IeeeFloat;
+
+/// Disturbes a floating-point result by a relative error on the order of (-2^scale, 2^scale).
+pub(crate) fn apply_random_float_error<F: rustc_apfloat::Float>(
+    ecx: &mut crate::MiriInterpCx<'_>,
+    val: F,
+    err_scale: i32,
+) -> F {
+    let rng = ecx.machine.rng.get_mut();
+    // Generate a random integer in the range [0, 2^PREC).
+    let dist = rand::distributions::Uniform::new(0, 1 << F::PRECISION);
+    let err = F::from_u128(dist.sample(rng))
+        .value
+        .scalbn(err_scale.strict_sub(F::PRECISION.try_into().unwrap()));
+    // give it a random sign
+    let err = if rng.gen::<bool>() { -err } else { err };
+    // multiple the value with (1+err)
+    (val * (F::from_u128(1).value + err).value).value
+}
+
+pub(crate) fn sqrt<S: rustc_apfloat::ieee::Semantics>(x: IeeeFloat<S>) -> IeeeFloat<S> {
+    match x.category() {
+        // preserve zero sign
+        rustc_apfloat::Category::Zero => x,
+        // propagate NaN
+        rustc_apfloat::Category::NaN => x,
+        // sqrt of negative number is NaN
+        _ if x.is_negative() => IeeeFloat::NAN,
+        // sqrt(∞) = ∞
+        rustc_apfloat::Category::Infinity => IeeeFloat::INFINITY,
+        rustc_apfloat::Category::Normal => {
+            // Floating point precision, excluding the integer bit
+            let prec = i32::try_from(S::PRECISION).unwrap() - 1;
+
+            // x = 2^(exp - prec) * mant
+            // where mant is an integer with prec+1 bits
+            // mant is a u128, which should be large enough for the largest prec (112 for f128)
+            let mut exp = x.ilogb();
+            let mut mant = x.scalbn(prec - exp).to_u128(128).value;
+
+            if exp % 2 != 0 {
+                // Make exponent even, so it can be divided by 2
+                exp -= 1;
+                mant <<= 1;
+            }
+
+            // Bit-by-bit (base-2 digit-by-digit) sqrt of mant.
+            // mant is treated here as a fixed point number with prec fractional bits.
+            // mant will be shifted left by one bit to have an extra fractional bit, which
+            // will be used to determine the rounding direction.
+
+            // res is the truncated sqrt of mant, where one bit is added at each iteration.
+            let mut res = 0u128;
+            // rem is the remainder with the current res
+            // rem_i = 2^i * ((mant<<1) - res_i^2)
+            // starting with res = 0, rem = mant<<1
+            let mut rem = mant << 1;
+            // s_i = 2*res_i
+            let mut s = 0u128;
+            // d is used to iterate over bits, from high to low (d_i = 2^(-i))
+            let mut d = 1u128 << (prec + 1);
+
+            // For iteration j=i+1, we need to find largest b_j = 0 or 1 such that
+            //  (res_i + b_j * 2^(-j))^2 <= mant<<1
+            // Expanding (a + b)^2 = a^2 + b^2 + 2*a*b:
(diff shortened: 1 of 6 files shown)


src/math.rs
```diff
@@ -0,0 +1,132 @@
...
+pub(crate) fn sqrt<S: rustc_apfloat::ieee::Semantics>(x: IeeeFloat<S>) -> IeeeFloat<S> {
+    match x.category() {
+        // preserve zero sign
+        rustc_apfloat::Category::Zero => x,
+        // propagate NaN
+        rustc_apfloat::Category::NaN => x,
+        // sqrt of negative number is NaN
+        _ if x.is_negative() => IeeeFloat::NAN,
+        // sqrt(∞) = ∞
+        rustc_apfloat::Category::Infinity => IeeeFloat::INFINITY,
```
RalfJung, line 33: Do we have tests for all of these (in `tests/pass/float.rs` or a similar file)?

src/math.rs
```diff
@@ -0,0 +1,132 @@
...
+            let prec = i32::try_from(S::PRECISION).unwrap() - 1;
+
+            // x = 2^(exp - prec) * mant
+            let mut exp = x.ilogb();
+            let mut mant = x.scalbn(prec - exp).to_u128(128).value;
+
+            if exp % 2 != 0 {
+                // Make exponent even, so it can be divided by 2
+                exp -= 1;
+                mant <<= 1;
```
RalfJung, line 45: This can't lead to overflow in `mant` because...?

src/math.rs
```diff
@@ -0,0 +1,132 @@
...
+
+            while d != 0 {
+                let t = s + d;
+                if rem >= t {
+                    s += d + d;
+                    rem -= t;
+                    res += d;
+                }
+                rem <<= 1;
+                d >>= 1;
```
RalfJung, line 62: Could you add some comments or reference explaining what happens here?

src/math.rs
```diff
@@ -0,0 +1,132 @@
...
+            let mut exp = x.ilogb();
+            let mut mant = x.scalbn(prec - exp).to_u128(128).value;
+
+            if exp % 2 != 0 {
+                // Make exponent even, so it can be divided by 2
+                exp -= 1;
+                mant <<= 1;
+            }
+
+            // Bit-by-bit sqrt with one extra bit
```
RalfJung, line 48: Why an extra bit?

src/math.rs
```diff
@@ -0,0 +1,132 @@
...
+                    s += d + d;
+                    rem -= t;
+                    res += d;
+                }
+                rem <<= 1;
+                d >>= 1;
+            }
+
+            // Round to nearest and remove extra bit
+            res = (res + (res & 1)) >> 1;
```
RalfJung, line 66: Round to nearest what? Is this the same as `res.next_multiple_of(2) >> 1`?

RalfJung (commented):
This is great, thanks! However, there are some parts I don't quite understand.

eduardosm:
I added some comments, I hope it is more clear now.

@rustbot ready

src/math.rs
RalfJung: So, is this the same as `res.next_multiple_of(2) >> 1`?

When a number is odd, it is equally near to both it even neighbor, so I don't get the "rounding to nearest" part.

src/math.rs
RalfJung: Thanks for adding the comments! TBH I don't really understand the math here, I'd have to spend a few hours working through this. I'll just trust that given the results are right, it's probably correct, and if someone needs to dig into this in the future the comments give them a starting point.

src/math.rs
eduardosm: It is the same as `next_multiple_of`, but I think the inline expression better expresses the intent of "isolate the lowest bit and add it to the value".

When the lowest bit is 1, the result is irrational (otherwise, the lowest ~prec/2 bits would be zero) and there is an infinite amount of 1 bits that have been truncated, so it is actually closer to the upper neighbor.

RalfJung: > but I think the inline expression better expresses the intent of "isolate the lowest bit and add it to the value".

But it doesn't explain what "isolate the lowest bit and add it to the value" has to do with "round to nearest".

> When the lowest bit is 1, the result is irrational (otherwise, the lowest ~prec/2 bits would be zero) and there is an infinite amount of 1 bits that have been truncated, so it (the full irrational result) is actually closer to the upper neighbor.

Okay I am completely lost now.^^ Can you expand this explanation (why are there infinitely many 1) and add it as a comment in the code?

eduardosm: I don't know if the comment I added makes it clearer.

RalfJung: Hm, not quite.

Let me try to re-phrase the comment in my own words. Is this correct?
```
            // Remove extra fractional bit from result, rounding to nearest.
            // If the last bit is 0, then the nearest neighbor is definitely the lower one.
            // If the last bit is 1, it sounds like this may either be a tie (if there's infinitely many
            // 0s after this 1), or the nearest neighbor is the upper one. However,
            // since square roots are either exact or irrational, and an exact root would
            // lead to the last "extra" bit being 0, we can exclude a tie in this case.
            // We therefore always round up if the last bit is 1.
```

RalfJung: Okay, thanks for explaining and checking! Please put this comment in the code. :)

eduardosm: Done :)

src/math.rs
```diff
@@ -0,0 +1,163 @@
...
+            }
+
+            // Remove extra fractional bit from result, rounding to nearest.
+            // If the last bit is 0, then the nearest neighbor is definitely the lower one.
+            // If the last bit is 1, it sounds like this may either be a tie (if there's
+            // infinitely many 0s after this 1), or the nearest neighbor is the upper one.
+            // However, since square roots are either exact or irrational, and an exact root
+            // would lead to the last "extra" bit being 0, we can exclude a tie in this case.
+            // We therefore always round up if the last bit is 1.
+            res = (res + (res & 1)) >> 1;
```
RalfJung, line 97: I have adjusted this line to `(res + 1) >> 1` (which is clearly wrong), and all tests still pass. Could you add a testcase (below in this module) that covers this case?

(`res >> 1` is successfully caught by the existing tests.)

eduardosm: Now that I think it, `(res + 1) >> 1` is actually correct. If the lowest bit is 0, it will become 1, just to be discarded by the shift.

RalfJung: Ah... that makes sense. :facepalm:

RalfJung:
Thanks a lot, great to finally have a proper implementation of the last missing standard IEEE float operation. :)

src/math.rs
```diff
@@ -0,0 +1,164 @@
...
+    let err = F::from_u128(dist.sample(rng))
+        .value
+        .scalbn(err_scale.strict_sub(F::PRECISION.try_into().unwrap()));
+    // give it a random sign
+    let err = if rng.gen::<bool>() { -err } else { err };
+    // multiple the value with (1+err)
+    (val * (F::from_u128(1).value + err).value).value
+}
+
+pub(crate) fn sqrt<S: rustc_apfloat::ieee::Semantics>(x: IeeeFloat<S>) -> IeeeFloat<S> {
```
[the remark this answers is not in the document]
RalfJung: Yeah, we have an open issue for that: https://github.com/rust-lang/rustc_apfloat/issues/14.

However, last time we wanted to change rustc_apfloat the answer was that things should land in LLVM first and then rustc_apfloat will pick them up from there. That's a quite tedious process. Not sure if that is still the current policy.

PR #4028, opened 2024-11-11
stacked borrows tests: add those that fail under TB

RalfJung:


RalfJung:
Oh that is exciting, one of these has UB when compiled with optimizations...

PR #4030, opened 2024-11-12
Renamed `this` to `ecx` in `extern_static`

YohDeadfall:
A small addition to #4029. There's no conflict with `self` and it's not an extension, so should be `ecx` too, I guess.


RalfJung:
Thanks!

PR #4031, opened 2024-11-12
filesystem support for solarish: stat

### LINKED ISSUE #3890: Solarish: Implement file system support
Our Solarish support is overall pretty good; the main thing that is missing is filesystem support. Solarish is probably very close to the other Unixes we support here, so this shouldn't be very hard.

The test command for this is: `./miri test --target x86_64-unknown-illumos fs`.

devnexen:
part of #3890

### DIFF
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -265,46 +265,58 @@ impl FileDescription for FileHandle {
 
 impl<'tcx> EvalContextExtPrivate<'tcx> for crate::MiriInterpCx<'tcx> {}
 trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
-    fn macos_stat_write_buf(
+    fn macos_fbsd_solaris_write_buf(
         &mut self,
         metadata: FileMetadata,
         buf_op: &OpTy<'tcx>,
     ) -> InterpResult<'tcx, i32> {
         let this = self.eval_context_mut();
 
-        let mode: u16 = metadata.mode.to_u16()?;
-
         let (access_sec, access_nsec) = metadata.accessed.unwrap_or((0, 0));
         let (created_sec, created_nsec) = metadata.created.unwrap_or((0, 0));
         let (modified_sec, modified_nsec) = metadata.modified.unwrap_or((0, 0));
+        let mode = metadata.mode.to_uint(this.libc_ty_layout("mode_t").size)?;
 
         let buf = this.deref_pointer_as(buf_op, this.libc_ty_layout("stat"))?;
-
         this.write_int_fields_named(
             &[
                 ("st_dev", 0),
-                ("st_mode", mode.into()),
+                ("st_mode", mode.try_into().unwrap()),
                 ("st_nlink", 0),
                 ("st_ino", 0),
                 ("st_uid", 0),
                 ("st_gid", 0),
                 ("st_rdev", 0),
                 ("st_atime", access_sec.into()),
-                ("st_atime_nsec", access_nsec.into()),
                 ("st_mtime", modified_sec.into()),
-                ("st_mtime_nsec", modified_nsec.into()),
                 ("st_ctime", 0),
-                ("st_ctime_nsec", 0),
-                ("st_birthtime", created_sec.into()),
-                ("st_birthtime_nsec", created_nsec.into()),
                 ("st_size", metadata.size.into()),
                 ("st_blocks", 0),
                 ("st_blksize", 0),
-                ("st_flags", 0),
-                ("st_gen", 0),
             ],
             &buf,
         )?;
+
+        if matches!(&*this.tcx.sess.target.os, "macos" | "freebsd") {
+            this.write_int_fields_named(
+                &[
+                    ("st_atime_nsec", access_nsec.into()),
+                    ("st_mtime_nsec", modified_nsec.into()),
+                    ("st_ctime_nsec", 0),
+                    ("st_birthtime", created_sec.into()),
+                    ("st_birthtime_nsec", created_nsec.into()),
+                    ("st_flags", 0),
+                    ("st_gen", 0),
+                ],
+                &buf,
+            )?;
+        }
+
+        if matches!(&*this.tcx.sess.target.os, "solaris" | "illumos") {
+            // FIXME: write st_fstype field once libc is updated.
+            // https://github.com/rust-lang/libc/pull/4145
+            //this.write_int_fields_named(&[("st_fstype", 0)], &buf)?;
+        }
 
         interp_ok(0)
     }
@@ -648,15 +660,15 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         interp_ok(Scalar::from_i32(this.try_unwrap_io_result(result)?))
     }
 
-    fn macos_fbsd_stat(
+    fn macos_fbsd_solaris_stat(
         &mut self,
         path_op: &OpTy<'tcx>,
         buf_op: &OpTy<'tcx>,
     ) -> InterpResult<'tcx, Scalar> {
         let this = self.eval_context_mut();
 
-        if !matches!(&*this.tcx.sess.target.os, "macos" | "freebsd") {
-            panic!("`macos_fbsd_stat` should not be called on {}", this.tcx.sess.target.os);
+        if !matches!(&*this.tcx.sess.target.os, "macos" | "freebsd" | "solaris" | "illumos") {
+            panic!("`macos_fbsd_solaris_stat` should not be called on {}", this.tcx.sess.target.os);
         }
 
         let path_scalar = this.read_pointer(path_op)?;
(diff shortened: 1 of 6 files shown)


src/shims/unix/fs.rs
```diff
@@ -290,22 +289,30 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                ("st_birthtime", created_sec.into()),
-                ("st_birthtime_nsec", created_nsec.into()),
                 ("st_size", metadata.size.into()),
                 ("st_blocks", 0),
                 ("st_blksize", 0),
-                ("st_flags", 0),
-                ("st_gen", 0),
             ],
             &buf,
         )?;
```
RalfJung, line 299: Seems like at least Illumos has one more field: `st_fstype`. So that should be initialized when running for that OS.

devnexen: unfortunately, libc crate [does not contain it](https://github.com/rust-lang/libc/blob/951cab59826837ccf53e47c75efef53cd3930693/src/unix/solarish/mod.rs#L324). solaris 11.4 and illumos has it tough, but the time it get fixed and release there, it will take a little while.

RalfJung:
> unfortunately, libc crate does not contain it.

In that case please add a comment so that we can remember to add it in the future.

src/shims/unix/fs.rs
```diff
@@ -272,8 +272,6 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
     ) -> InterpResult<'tcx, i32> {
         let this = self.eval_context_mut();
 
-        let mode: u16 = metadata.mode.to_u16()?;
```
RalfJung, line 275: This should be `metadata.mode.to_uint(this.libc_ty_layout("mode_t").size)`. Then the `mode.into()` below should become `mode.try_into().unwrap()`.

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -82,6 +82,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let result = this.macos_fbsd_solaris_readdir_r(dirp, entry, result)?;
                 this.write_scalar(result, dest)?;
             }
+            // TODO readdir ?
```
RalfJung, line 85: Please leave readdir for a future PR.

devnexen: I have to because test fs do not pass, it is just a reminder.

RalfJung: You can just disable parts of the test on Solarish for now. Please keep the PR small.

tests/pass/shims/fs.rs
```diff
@@ -27,7 +28,10 @@ fn main() {
     test_file_sync();
     test_errors();
     test_rename();
+    // solarish needs to intercept readdir/readdir64 for these tests.
+    #[cfg(not(any(target_os = "solaris", target_os = "illumos")))]
     test_directory();
+    #[cfg(not(any(target_os = "solaris", target_os = "illumos")))]
     test_canonicalize();
```
RalfJung, line 35: Please use `if cfg!(...)` instead of `#[cfg]`, then you can avoid making the import so messy.

tests/pass/shims/fs.rs
```diff
@@ -27,7 +28,10 @@ fn main() {
     test_file_sync();
     test_errors();
     test_rename();
+    // solarish needs to intercept readdir/readdir64 for these tests.
```
RalfJung, line 31: I'd say "support" rather than "intercept"

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -82,7 +82,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let result = this.macos_fbsd_solaris_readdir_r(dirp, entry, result)?;
                 this.write_scalar(result, dest)?;
             }
-            // TODO readdir ?
+            // TODO readdir/readdir64: for tests fs::test_directory/canonicalize
```
RalfJung, line 85: Please remove this, open an issue instead.

src/shims/unix/fs.rs
```diff
@@ -265,47 +265,65 @@ impl FileDescription for FileHandle {
...
+            let mode: u32 = metadata.mode.to_u32()?;
+
+            // FIXME: libc crate as `st_fstype` will be uninitialised
+            //this.write_int_fields_named(&[("st_mode", mode.into()), ("st_fstype", 0)], &buf)?;
+            this.write_int_fields_named(&[("st_mode", mode.into())], &buf)?;
+        } else {
+            let mode: u16 = metadata.mode.to_u16()?;
+
+            this.write_int_fields_named(&[("st_mode", mode.into())], &buf)?;
+        }
```
RalfJung, line 325: Why is this code still there?

RalfJung:
> unfortunately, libc crate does not contain it.
> In that case please add a comment so that we can remember to add it in the future.

This comment is still missing

src/shims/unix/fs.rs
RalfJung: The comment explaining this is still missing.

tests/pass/shims/fs.rs
devnexen: but what would be the condition tough ?

devnexen:
broader question would be should we wait the libc crate update or do we keep the missing field as TODO ?

RalfJung:
TODO is fine for now.

tests/pass/shims/fs.rs
RalfJung: `if cfg!(not(any(target_os = "solaris", target_os = "illumos"))) { ... }`

devnexen: but that s the thing, I tried this and the tests are called still on my illumos instance.

devnexen: ok will give it a go in few hours.

devnexen: so, yes that was an error on my part indeed :) let s see how it goes ..

src/shims/unix/fs.rs
```diff
@@ -265,47 +265,59 @@ impl FileDescription for FileHandle {
...
+                    ("st_flags", 0),
+                    ("st_gen", 0),
+                ],
+                &buf,
+            )?;
+        }
+
+        if matches!(&*this.tcx.sess.target.os, "solaris" | "illumos") {
+            // FIXME: libc crate, update stat struct, as `st_fstype` will be uninitialised
+            // https://github.com/rust-lang/libc/pull/4145
```
RalfJung, line 317:
```suggestion
            // FIXME: write st_fstype field once libc is updated
            // https://github.com/rust-lang/libc/pull/4145
```

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -57,6 +57,32 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    this.check_shim(abi, ExternAbi::C { unwind: false }, link_name, args)?;
+                let result = this.macos_fbsd_solaris_fstat(fd, buf)?;
+                this.write_scalar(result, dest)?;
+            }
+            "readdir_r" => {
+                let [dirp, entry, result] =
+                    this.check_shim(abi, ExternAbi::C { unwind: false }, link_name, args)?;
+                let result = this.macos_fbsd_solaris_readdir_r(dirp, entry, result)?;
+                this.write_scalar(result, dest)?;
+            }
```
RalfJung, line 84: I am confused -- why is readdir implemented now?

Please don't add untested code. If it "just works" without any other changes, enable the test, otherwise remove the implementation.

RalfJung (commented):
@rustbot author

src/shims/unix/solarish/foreign_items.rs
devnexen: I did not add readdir**_r** it had been there since [in the early phase](https://github.com/rust-lang/miri/commit/13b741d1c858405b85c94510dec678b4571f52a0). Anyhow, I do not need it at all for now.

src/shims/unix/macos/foreign_items.rs
```diff
@@ -64,7 +64,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             "readdir_r" | "readdir_r$INODE64" => {
                 let [dirp, entry, result] =
                     this.check_shim(abi, ExternAbi::C { unwind: false }, link_name, args)?;
-                let result = this.macos_fbsd_readdir_r(dirp, entry, result)?;
+                let result = this.macos_fbsd_solaris_readdir_r(dirp, entry, result)?;
```
RalfJung, line 67: Please also un-do this rename then.

src/shims/unix/fs.rs
```diff
@@ -1112,16 +1130,19 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
     ) -> InterpResult<'tcx, Scalar> {
         let this = self.eval_context_mut();
 
-        if !matches!(&*this.tcx.sess.target.os, "macos" | "freebsd") {
-            panic!("`macos_fbsd_readdir_r` should not be called on {}", this.tcx.sess.target.os);
+        if !matches!(&*this.tcx.sess.target.os, "macos" | "freebsd" | "solaris" | "illumos") {
+            panic!(
+                "`macos_fbsd_solaris_readdir_r` should not be called on {}",
+                this.tcx.sess.target.os
+            );
```
RalfJung, line 1145: and undo this as well

RalfJung:
Thanks :)

PR #4033, opened 2024-11-13
Add test for epoll

tiif:
I forgot how the implementation works and was wondering how the assumption here is being upheld. So I wrote a test. 

https://github.com/rust-lang/miri/blob/9d9da34f49562bbe78e09213ee6e8aa9b9db1cce/src/shims/unix/linux/epoll.rs#L541-L542

This is mainly written for me to verify if it actually works, so it is perfectly ok to not merge this.

### DIFF
--- a/tests/fail-dep/libc/libc_epoll_block_two_thread.rs
+++ b/tests/fail-dep/libc/libc_epoll_block_two_thread.rs
@@ -1,0 +1,96 @@
+//@compile-flags: -Zmiri-preemption-rate=0
+//~^ERROR: deadlocked
+//~^^ERROR: deadlocked
+//@only-target: linux
+//@error-in-other-file: deadlock
+
+use std::convert::TryInto;
+use std::thread;
+use std::thread::spawn;
+
+// Using `as` cast since `EPOLLET` wraps around
+const EPOLL_IN_OUT_ET: u32 = (libc::EPOLLIN | libc::EPOLLOUT | libc::EPOLLET) as _;
+
+#[track_caller]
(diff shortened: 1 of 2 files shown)


tests/fail-dep/libc/libc_epoll_block_two_thread.rs
```diff
@@ -0,0 +1,96 @@
...
+    let thread3 = spawn(move || {
+        thread::park();
+        check_epoll_wait::<1>(epfd, &[(expected_event, expected_value)], -1);
+    });
+
+    thread2.thread().unpark();
+    thread::yield_now();
+    thread3.thread().unpark();
+    thread::yield_now();
+    thread1.thread().unpark();
```
tiif, line 91: This reminds me that a thread won't immediately get executed after being unparked, I will check the tests in #3939 again.

RalfJung:
Ah I guess this is a different aspect of "edge semantics".

tiif:
> Ah I guess this is a different aspect of "edge semantics".

Yup it's different. [The non-blocking test](https://github.com/rust-lang/miri/blob/9e0205daee66307a2ac7793d449e934d7c6de4e9/tests/pass-dep/libc/libc-epoll-no-blocking.rs#L83-L84) checks "if no event occurs between two epoll_wait calls, we should not receive any notification during the second epoll_wait". The test in this PR checks "if multiple threads are blocking on the same epoll fd, a single notification on that epoll fd should wake up only one thread (the alternative would be waking up all threads, i.e. [thundering herd](https://en.wikipedia.org/wiki/Thundering_herd_problem)  behaviour )"

PR #4035, opened 2024-11-15
refactor: refine thread variant for windows

### LINKED ISSUE #4014: Refine the `Thread` variant of Windows `Handle` enum
Pull request #4004 followed the first proposal from https://github.com/rust-lang/miri/pull/4004#discussion_r1828351341 which led to some ugliness in Windows shims. That should be fixed by wrapping the logic of `Handle::from_scalar` in a function with access to `InterpCx`. See `thread_id_try_from` for an example.

discord9:
close #4014, check thread id before making it into `handle` on Windows.

### DIFF
--- a/src/shims/windows/handle.rs
+++ b/src/shims/windows/handle.rs
@@ -2,6 +2,7 @@ use std::mem::variant_count;
 
 use rustc_abi::HasDataLayout;
 
+use crate::concurrency::thread::ThreadNotFound;
 use crate::*;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
@@ -14,7 +15,7 @@ pub enum PseudoHandle {
 pub enum Handle {
     Null,
     Pseudo(PseudoHandle),
-    Thread(u32),
+    Thread(ThreadId),
 }
 
 impl PseudoHandle {
@@ -34,6 +35,14 @@ impl PseudoHandle {
     }
 }
 
+/// Errors that can occur when constructing a [`Handle`] from a Scalar.
+pub enum HandleError {
+    /// There is no thread with the given ID.
+    ThreadNotFound(ThreadNotFound),
+    /// Can't convert scalar to handle because it is structurally invalid.
+    InvalidHandle,
+}
+
 impl Handle {
     const NULL_DISCRIMINANT: u32 = 0;
     const PSEUDO_DISCRIMINANT: u32 = 1;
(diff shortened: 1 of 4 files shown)


src/shims/windows/handle.rs
```diff
@@ -126,21 +127,32 @@ impl Handle {
         Scalar::from_target_isize(signed_handle.into(), cx)
     }
 
-    pub fn from_scalar<'tcx>(
+    /// fail if thread id is invalid
```
RalfJung, line 130:
```suggestion
    /// Convert a scalar into a structured `Handle`.
    /// Structurally invalid handles return `Ok(None)`.
    /// If the handle is structurally valid but semantically invalid, e.g. a for non-existent thread
    /// ID, returns `Err`.
```
This makes me wonder if that's really the semantics we want. Wouldn't `Result<Self, ...>` make more sense?

RalfJung (commented):
Thanks for the PR!

src/shims/windows/handle.rs
discord9: > This makes me wonder if that's really the semantics we want. Wouldn't `Result<Self, ...>` make more sense?

Make sense, this seems a bit repeating, how about `Result<Self, HandleError>` with:
```rust
enum HandleError{
  ThreadNotFound(ThreadNotFound),
  InvalidHandle
}
```

src/shims/windows/handle.rs
```diff
@@ -35,6 +35,14 @@ impl PseudoHandle {
     }
 }
 
+/// Handle errors
```
RalfJung, line 38:
```suggestion
/// Errors that can occur when constructing a [`Handle`] from a Scalar.

File: src/shims/windows/handle.rs [hunk 3]
```diff
@@ -35,6 +35,14 @@ impl PseudoHandle {
...
+pub enum HandleError {
+    /// There is no thread with the given ID
+    ThreadNotFound(ThreadNotFound),
+    /// Can't convert scalar to handle because it is structurally invalid
```
RalfJung, line 42:
```suggestion
    /// There is no thread with the given ID.
    ThreadNotFound(ThreadNotFound),
    /// Can't convert scalar to handle because it is structurally invalid.
```

RalfJung (commented):
Looks good! Just some comment nits.

RalfJung:
Great, thanks! Could you squash this into a single commit? (Using `git rebase --interactive --keep-base` to keep the diff nicer.)

Issue #3592, opened 2024-05-09
miri-script: use a proper CLI arg parsing tool


RalfJung:
We are currently using ad-hoc hand-written parsers for this, and it's kind of annoying. However, we tried to use clap, and it currently doesn't seem possible: what we'd like to do is that once there is a positional argument, that argument and everything that follows gets collected in `Vec` -- our flags, like `--dep`, need to come before all of them.

This is non-standard, as it makes `./miri run foo.rs --dep` different from `./miri run --dep foo.rs`. But it's kind of necessary as `./miri run` (and a bunch of other commands) are wrappers around other, complicated tools for which we want to just forward all flags (including `--`) without parsing them ourselves.

Using clap here is blocked on https://github.com/clap-rs/clap/issues/5055. But maybe there's another crate we can use?

RalfJung:
With https://github.com/rust-lang/miri/pull/3621, our parser's behavior changed: `./miri run foo.rs --dep` and `./miri run --dep foo.rs` are now equivalent. However, we'd still like to preserve the `--` and forward it to miri (e.g. in `./miri run foo.rs -- arg1 arg2`), so we still can't use clap.

Mandragorian:
Would it make any sense to split the argv in the before and after `--` part, and use [`clap::Parser::parse_from`](https://docs.rs/clap/latest/clap/trait.Parser.html#method.parse_from) for the first part, and forward the second part?

RalfJung:
Maybe? Not sure how ugly that would be.

Also, thinking about this now I realize adding clap would likely significantly increase the build times for miri-script.  So maybe the current solution isn't such a bad tradeoff after all.

RalfJung:
This got fixed by https://github.com/rust-lang/miri/pull/4036.

PR #4036, opened 2024-11-17
Use clap in miri-script

Mandragorian:
Use clap in miri-script, following discussion in #3592. 

Simplify the code by using the derive API from clap. This way we can just declare how the cli parser should behave, instead of implementing it ourselves.

This come with a slowdown to the compilation time of miri-script, but it is acceptable compared to the benefits.


miri-script/src/main.rs
```diff
@@ -87,7 +120,33 @@ pub enum Command {
...
+            | Self::Doc { flags }
+            | Self::Fmt { flags }
+            | Self::Clippy { flags }
+            | Self::Run { flags, .. }
+            | Self::Test { flags, .. } =>
+                if !remainder.is_empty() {
+                    flags.push("--".into());
+                    flags.extend(remainder);
+                },
+            _ => (),
```
RalfJung, line 138: Shouldn't this produce some sort of error?

miri-script/src/main.rs
```diff
@@ -87,7 +120,33 @@ pub enum Command {
...
+    }
+}
+
+#[derive(Parser)]
+pub struct Cli {
+    #[command(subcommand)]
+    pub command: Command,
+}
+
+const _HELP: &str = r#"  COMMANDS
```
RalfJung, line 149: clap can generate this for us, can't it? (It's okay to only do this once we decided, but I want to know whether further simplification is possible here.)

miri-script/src/main.rs
```diff
@@ -150,108 +209,15 @@ CARGO_EXTRA_FLAGS:
...
-            }
-            Command::RustcPush { github_user, branch }
-        }
-        _ => {
-            eprintln!("Unknown or missing command. Usage:\n\n{HELP}");
-            std::process::exit(1);
-        }
-    };
+    let miri_args: Vec<_> =
+        std::env::args().peekable().take_while(|x| *x != "--").filter(|x| *x != "--").collect();
```
RalfJung, line 213: Why `peekable`? And why the extra `filter`?

miri-script/src/main.rs
```diff
@@ -150,108 +209,15 @@ CARGO_EXTRA_FLAGS:
...
+    let remainder: Vec<_> =
+        std::env::args().peekable().skip_while(|x| *x != "--").skip(1).collect();
```
RalfJung, line 215: I'd remove the `skip` here and make the `--` part of the remainder, that also simplifies `add_remainder`.

Cargo.lock
RalfJung, line 1: There should be no reason for this file to change.

RalfJung (commented):
Thanks for trying this! I agree it is quite elegant.

However, it also doubles our build times for the script. The script hopefully doesn't get built too often, so this is mostly a thing for first-time users -- they get no feedback while `miri-script` itself is built. Ideally we'd build it on stable Rust so it wouldn't be rebuilt a lot, but we need some nightly cargo features for the diagnostic integration...

Overall I am leaning slightly in favor of this, just to make maintenance of the script easier. I'd still be curious how this would look like without the derive; is there a good example somewhere of what using the clap builder directly looks like?

RalfJung:
Also @rust-lang/miri what do you all think?

miri-script/src/main.rs
Mandragorian: I actually just forgot to delete this in this pr :sweat_smile:. Since I wasn't sure this would be useful, I kept the PoC low  effort. Clap generates the usage string automatically.

miri-script/src/main.rs
Mandragorian: Yeap, `peekable` is not needed. I just quickly copy pasted the code from the `args` module without thinking too much for the prototype. I used the extra  filter because for some reason i was under the impression that the `--` was included in the result. This is wrong and the filter is not needed.

miri-script/src/main.rs
Mandragorian: I agree this is a much better way of doing it.

saethlin:
> a much longer compilation time from 2.5s to 5s after the changes.

I see a regression from 2.5s to 3s, which would be fine with me. I like clap, and think it's worth a bit of compile time overhead, but doubling the compile time of miri-script seems a bit steep. I'd quite like to know what the compile time is without using the proc macro (maybe you can just `cargo expand` miri-script and clean up what comes out?).

RalfJung:
I am measuring a slowdown from 4s to 5.2s.

Mandragorian:
what command do you use to build? I can write a small script to measure the build times a bit more systematically. maybe I got an outlier.

saethlin:
`rm -r miri-script/target/ && time ./miri --help`

RalfJung:
Absolute times are obviously going to differ. What's the relative slowdown? Seems to be 40% for you, 30% for me, 20% for saethlin.

Mandragorian:
Yes, these are the relative slowdowns. I compared with saethlin's numbers since we had similar baselines.

What would be the threshold for an acceptable slowdown?

RalfJung:
IMO that is the wrong question. The right question is, how low can we get the slowdown while still getting most of the benefits of using clap (or some other similar library)?

RalfJung:
> But then, as long as clap's version doesn't change it won't need to be rebuilt (?).

It gets rebuild whenever the compiler changes, which is `nightly`, so it can change quite frequently.

RalfJung:
Now that we decided to go with the `derive` version, the next step would be to clean up and finish this PR. :)

@rustbot author

Mandragorian:
I see that this now has conflicts. Not sure if you prefer to rebase it now or go through the review first beforehand. I also see that the checks do not run, and I am not sure why.

@rustbot ready

RalfJung:
Please rebase first. Github doesn't run the checks if there are conflicts (since the checks are run against your PR merged with master).

miri-script/src/main.rs
```diff
@@ -87,171 +122,44 @@ pub enum Command {
...
+            | Self::Fmt { flags }
+            | Self::Toolchain { flags }
+            | Self::Clippy { flags }
+            | Self::Run { flags, .. }
+            | Self::Test { flags, .. } => {
+                flags.extend(remainder);
+                Ok(())
+            }
+            Self::Bench { .. } | Self::RustcPull { .. } | Self::RustcPush { .. } =>
+                Err(anyhow::Error::msg("unexpected \"--\" found in arguments")),
```
RalfJung, line 145: Please use the `bail!` macro fro  anyhow.

miri-script/src/main.rs
```diff
@@ -87,171 +122,44 @@ pub enum Command {
...
-                bail!("Too many arguments for `./miri rustc-push GITHUB_USER BRANCH`");
-            }
-            Command::RustcPush { github_user, branch }
-        }
-        _ => {
-            eprintln!("Unknown or missing command. Usage:\n\n{HELP}");
-            std::process::exit(1);
-        }
-    };
+    let miri_args: Vec<_> = std::env::args().take_while(|x| *x != "--").collect();
```
RalfJung, line 157:
```suggestion
    /// Split the arguments into the part before the `--` and the part after.
    /// The `--` itself ends up in the second part.
    let miri_args: Vec<_> = std::env::args().take_while(|x| *x != "--").collect();
```

RalfJung (commented):
Could you please post what the `--help` output looks like? We might want to tweak the doc comments now that they are more user-visible.

RalfJung:
Also please update the PR title and commit message to make it clear that this is not an experiment any more.

@rustbot author

miri-script/src/main.rs
Mandragorian: Were the doc comments intended? I don't think they will show up somewhere, and indeed I get warnings about them when I compile.

RalfJung: *oops* no

Mandragorian:
This is what the help message looks like:

```
[root@1dfa6c4d75e7 miri]# ./miri help
Usage: miri-script <COMMAND>

Commands:
  install     Installs the miri driver and cargo-miri. Sets up the rpath such that the installed binary should work in any working directory. Note that the binaries are placed in the `miri` toolchain sysroot, to prevent conflicts with other toolchains
  build       Just build miri
  check       Just check miri
  test        Build miri, set up a sysroot and then run the test suite
  run         Build miri, set up a sysroot and then run the driver with the given <flags>. (Also respects MIRIFLAGS environment variable.)
  doc         Build documentation
  fmt         Format all sources and tests
  clippy      Runs clippy on all sources
  bench       Runs the benchmarks from bench-cargo-miri in hyperfine. hyperfine needs to be installed
  toolchain   Update and activate the rustup toolchain 'miri' to the commit given in the `rust-version` file. `rustup-toolchain-install-master` must be installed for this to work. Any extra flags are passed to `rustup-toolchain-install-master`
  rustc-pull  Pull and merge Miri changes from the rustc repo. Defaults to fetching the latest rustc commit. The fetched commit is stored in the `rust-version` file, so the next `./miri toolchain` will install the rustc that just got pulled
  rustc-push  Push Miri changes back to the rustc repo. This will pull a copy of the rustc history into the Miri repo, unless you set the RUSTC_GIT env var to an existing clone of the rustc repo
  help        Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help
  ```

RalfJung:
I have resolved the conflicts and done a pass over all the help texts. I noticed some things that were missing (`rustc-push` lost its default branch name, and the environment variables part of the help was removed without replacement).

Should be good to go now :)

RalfJung:
FWIW this did change the behavior for some flags. Something like `./miri test foo bar --bless baz` used to work, but it doesn't work any more. @Mandragorian do you think there's any way to keep that working?

PR #4037, opened 2024-11-17
Refactor AnonSocket::read/write for blocking socketpair

tiif:
This PR moves the actual read/write operations of ``AnonSocket::read/write`` into separate functions, in preparation for #3665. The new functions will be used as blocking callback.

### DIFF
--- a/src/shims/unix/unnamed_socket.rs
+++ b/src/shims/unix/unnamed_socket.rs
@@ -146,8 +146,7 @@ impl FileDescription for AnonSocket {
             // corresponding ErrorKind variant.
             throw_unsup_format!("reading from the write end of a pipe");
         };
-        let mut readbuf = readbuf.borrow_mut();
-        if readbuf.buf.is_empty() {
+        if readbuf.borrow().buf.is_empty() {
             if self.peer_fd().upgrade().is_none() {
                 // Socketpair with no peer and empty buffer.
                 // 0 bytes successfully read indicates end-of-file.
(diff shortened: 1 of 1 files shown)


src/shims/unix/unnamed_socket.rs
```diff
@@ -167,31 +167,9 @@ impl FileDescription for AnonSocket {
...
-        // don't know what that *certain number* is, we will provide a notification every time
-        // a read is successful. This might result in our epoll emulation providing more
-        // notifications than the real system.
-        if let Some(peer_fd) = self.peer_fd().upgrade() {
-            ecx.check_and_update_readiness(&peer_fd)?;
-        }
-
-        ecx.return_read_success(ptr, &bytes, actual_read_size, dest)
+        // TODO: We might need to decide what to do if peer_fd is closed when read is blocked.
+        anonsocket_read(self, self.peer_fd().upgrade(), &mut bytes, ptr, dest, ecx)
```
tiif, line 172: I am not entirely sure if closing ``peer_fd`` when it is blocking should be allowed. I currently prefer to leave the TODO here and revisit it again when implementing blocking (and write a testcase for this).

src/shims/unix/unnamed_socket.rs
```diff
@@ -146,7 +146,7 @@ impl FileDescription for AnonSocket {
             // corresponding ErrorKind variant.
             throw_unsup_format!("reading from the write end of a pipe");
         };
-        let mut readbuf = readbuf.borrow_mut();
+        let readbuf = readbuf.borrow_mut();
         if readbuf.buf.is_empty() {
```
oli-obk, line 150:
```suggestion
        if readbuf.borrow_mut().buf.is_empty() {
```

Then you don't need the explicit drop

src/shims/unix/unnamed_socket.rs
```diff
@@ -233,24 +211,86 @@ impl FileDescription for AnonSocket {
...
-        // Need to stop accessing peer_fd so that it can be notified.
         drop(writebuf);
+        anonsocket_write(available_space, &peer_fd, ptr, len, dest, ecx)
+    }
+}
 
-        // Notification should be provided for peer fd as it became readable.
-        // The kernel does this even if the fd was already readable before, so we follow suit.
+/// Read from AnonSocket and return the number of bytes read.
+fn anonsocket_read<'tcx>(
```
oli-obk, line 220: Move this function below `anonsocket_write` to avoid intermingling the diffs

src/shims/unix/unnamed_socket.rs
tiif: Nice! I wonder if it would be correct to assume that ``Ref`` or ``RefMut`` will be automatically dropped at the end of the statement if it is not used elsewhere?

oli-obk: Yes. Temporary values are dropped at the end of their statement

oli-obk:
Lgtm. Please squash

src/shims/unix/unnamed_socket.rs
RalfJung: I would expect that closing one side wakes up the other side, and makes it return an EOF read.

I'm a bit surprised that all that `if readbuf.borrow().buf.is_empty() {` stuff is not insider the helper, but :shrug: I guess it can be moved later if it turns out to be useful.

Issue #3482, opened 2024-04-18
Windows file system shims are missing


RalfJung:
Miri supports accessing the file system (opening, reading, writing files; listing directories) on Unix systems, but not on Windows. This is probably the biggest remaining gap in our Windows support.

If you run into this, try `cargo miri test --target x86_64-unknown-linux-gnu`. This works even on a Windows host (without installing anything extra) and interprets programs on a well-supported Linux target -- file system access should generally work then. If it doesn't please file an issue (do not comment on this issue, instead file a new one).

Nobody on the Miri team is very knowledgeable in Windows APIs, so this is unlikely to be on our agenda any time soon. Contributions are always welcome though. :)

mbyt:
@RalfJung thanks for the fast response in #1537 and creating this ticket (I'll subscribe to it). Also thanks for pointing out that Miri can run with target linux on windows. I was not aware of that.

Unfortunately, this workaround is not possible for me, as I am depending on a large external C dependency which is wrapped via bindgen. Unfortunately, this external C dependency has no linux support, thus I cannot cross-compile for linux.

RalfJung:
Miri does not work with external C dependencies anyway, so that means it won't work for you on Windows either.

mbyt:
Thanks again for your fast response.

What I currently do, is that I run the complete unittest suite on windows and skip the tests which (i) access the filesystem or (ii) a foreign function. This works actually quite well and thanks to miri I already found a few places of undefined behavior. That's marvelous!

With this ticket I could also run the tests which access the filesystem and would only need to skip the tests with the foreign function.

Regarding external C dependencies: Is this the topic of #2365?

RalfJung:
Okay, so the problem is getting the build scripts to shut up and do nothing so that you can use `--target x86_64-unknown-linux-gnu`. On Windows the external dependencies will be built and linked even in Miri, which is useless but harmless. With a cross-target then it is nor harmless any more, it causes the build to fail. I can see how disabling those build scripts could become tricky. Might still be simpler than implementing the Windows FS shims though. ;)

> Regarding external C dependencies: Is this the topic of https://github.com/rust-lang/miri/issues/2365?

Yes.

> This works actually quite well and thanks to miri I already found a few places of undefined behavior. That's marvelous!

Awesome. :)

If these bugs have a public track record / issue / PR somewhere, we always appreciate PRs that add more bugs to the "bugs found with Miri" list in the readme. :D

mbyt:
[...]

PR #4043, opened 2024-11-21
Implement Windows file read, write, and delete APIs.

CraftSpider:
This also improves how HANDLEs are, well, handled a bit. STDIN/ERR/OUT are now proper pseudo-handles. The invalid handle is added and encode/decodes to exactly 0xFFFFFFFF correctly.

Unix shims require changes to hoist FdTable up, so that Windows can share most of its implementation.

### DIFF
--- a/src/shims/files.rs
+++ b/src/shims/files.rs
@@ -1,6 +1,6 @@
 use std::any::Any;
 use std::collections::BTreeMap;
-use std::io::{IsTerminal, Read, SeekFrom, Write};
+use std::io::{ErrorKind, IsTerminal, Read, SeekFrom, Write};
 use std::ops::Deref;
 use std::rc::{Rc, Weak};
 use std::{fs, io};
@@ -25,7 +25,7 @@ pub trait FileDescription: std::fmt::Debug + Any {
         _len: usize,
         _dest: &MPlaceTy<'tcx>,
         _ecx: &mut MiriInterpCx<'tcx>,
-    ) -> InterpResult<'tcx> {
+    ) -> InterpResult<'tcx, Result<(), io::ErrorKind>> {
         throw_unsup_format!("cannot read from {}", self.name());
     }
 
@@ -40,7 +40,7 @@ pub trait FileDescription: std::fmt::Debug + Any {
         _len: usize,
         _dest: &MPlaceTy<'tcx>,
         _ecx: &mut MiriInterpCx<'tcx>,
-    ) -> InterpResult<'tcx> {
+    ) -> InterpResult<'tcx, Result<(), io::ErrorKind>> {
         throw_unsup_format!("cannot write to {}", self.name());
     }
 
@@ -97,7 +97,7 @@ impl FileDescription for io::Stdin {
         len: usize,
         dest: &MPlaceTy<'tcx>,
         ecx: &mut MiriInterpCx<'tcx>,
-    ) -> InterpResult<'tcx> {
+    ) -> InterpResult<'tcx, Result<(), io::ErrorKind>> {
         let mut bytes = vec![0; len];
         if !communicate_allowed {
             // We want isolation mode to be deterministic, so we have to disallow all reads, even stdin.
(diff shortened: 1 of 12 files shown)


RalfJung:
Thanks for the PR!

This is a *huge* PR. It will take a while until I have enough time to take a look at such a big change. If there is any way to stage this as a series of smaller changes, that would be much appreciated.

CraftSpider:
I could possibly further split this down by first PRing the `FdMap` move, and the handle changes, as their own PRs. In that format, both of those would basically be refactors with no/minimal functional changes, then this PR would just add the shims. It would still be a good bit of stuff, just because even a minimal test needs 4 or 5 shims, but it would be notably smaller hopefully.

RalfJung:
That sounds like a good plan. You can keep this PR open (maybe mark it as draft); it can serve as good context for why you want to refactor things the way you suggest.

RalfJung:
Also, I know very little about the Windows API... @ChrisDenton @beepster4096  if you have any chance at taking a look at the windows-specific parts here and check whether they look sensible, that would be great. :)

cgettys-microsoft:

src/shims/windows/handle.rs
```diff
@@ -8,6 +9,10 @@ use crate::*;
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub enum PseudoHandle {
     CurrentThread,
+    CurrentProcess,
+    Stdin,
+    Stdout,
+    Stderr,
```
ChrisDenton, line 15: Note that the handles returned by `GetStdHandle` are proper file handles and not pseudo-handles. This differs from posix where the fds have fixed values. I realise that miri does want to track the difference but I worry about confusing the terminology.

CraftSpider: Hm, some ideas on paths forwards:
- Leave it as-is, but add documentation about the terminology here
- Try to merge stdin/out into `Handle::File` as normal `FileDescription` (basically what unix does)
  - This will require some more changes to the `FileDescription` trait. Its `read` and `write` methods are a bit tailored towards unix flavor currently, you may notice I have some hacky code in here to work around that. I'd like to fix it eventually anyways, but it adds more friction to MVP
- Make `Handle::File` hold an enum like `enum FileHandle { Stdin, Stdout, File(i32) }`

beepster4096: I don't think miri actually needs to track the difference. It should be fine to use FileDescriptions for them like it does on Unix

RalfJung: > Its read and write methods are a bit tailored towards unix flavor currently, you may notice I have some hacky code in here to work around that. I'd like to fix it eventually anyways, but it adds more friction to MVP

Is it even worth using the same trait for Unix and Windows? The trait is indeed very tailored towards Unix, deliberately so. Knowing nothing about the Windows APIs, I expected they would need their own trait anyway.

I haven't looked yet at those hacks. Can you write a high-level summary of the pros and cons of using the same vs a different trait?

CraftSpider:
Rebased on top of #4045

src/shims/files.rs
```diff
@@ -402,10 +411,21 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         &mut self,
         actual_write_size: usize,
         dest: &MPlaceTy<'tcx>,
-    ) -> InterpResult<'tcx> {
+    ) -> InterpResult<'tcx, Result<(), io::ErrorKind>> {
```
RalfJung, line 414: If you have to completely change how we do error handling for IO functions on Unix, please make that a separate PR.

CraftSpider: Will do. I wanted to make the change to this branch first to ensure it would work well and pass local tests with both the existing unix code, and my new Windows shims. I already intended to split this change off and merge it separately.

CraftSpider:
All parts of this PR are either merged, in PR, or ready to PR soon. I'm going to close it.

PR #4045, opened 2024-11-21
Move FdTable to a common location and split off Unix behavior 

CraftSpider:
I'm open to instead keeping all the functions on the base trait, but this seemed like an easy way to keep them clearly only intended to be used in the unix shims.

Preparation for Windows FS support, see #4043


cgettys-microsoft:

src/shims/files.rs
RalfJung, line 1: Is anything in this file *different* from what used to be in `fd.rs`? If yes, please split this up in multiple commits -- one that just moves code without changing it, and one that changes it. As-is, it is nearly impossible to review this since I'd have to manually compare your new code with the current code.

RalfJung:
> I'm open to instead keeping all the functions on the base trait, but this seemed like an easy way to keep them clearly only intended to be used in the unix shims.

What are you referring to here?

src/shims/unix/fd.rs
```diff
@@ -21,40 +18,8 @@ pub(crate) enum FlockOp {
...
-        _ptr: Pointer,
-        _len: usize,
-        _dest: &MPlaceTy<'tcx>,
-        _ecx: &mut MiriInterpCx<'tcx>,
-    ) -> InterpResult<'tcx> {
-        throw_unsup_format!("cannot write to {}", self.name());
-    }
-
+/// Represents unix-specific file descriptions.
+pub trait UnixFileDescription: FileDescription {
```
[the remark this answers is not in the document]
CraftSpider: My comment on the other PR noted that I may have been wrong about the signature not being good for Windows. I'm going to try something and confirm.

Separately, your first sentence of this comment is what I was referring to by 'keeping all the functions on the base trait' - I'm not sure if it's worth it to split the trait, or to keep it all as one. I just did it this way because it was relatively simple, but I have no strong opinions.

src/shims/files.rs
CraftSpider: I'll double check and split it if there's any difference.

CraftSpider:
This is now 3 commits - the first only moves the FdMap, the seconds splits out Unix functionality, the third makes the `fn metadata` change used by the Windows PR

RalfJung:
@CraftSpider Thanks!
This seems to have conflicts with master now, so CI cannot run. Could you resolve those conflicts?

@rustbot author

(write `@rustbot ready` when this is ready for review)

src/shims/files.rs
```diff
@@ -0,0 +1,411 @@
...
+        // A weak reference can never be the only reference to some pointer or place.
+        // Since the actual file description is tracked by strong ref somewhere,
+        // it is ok to make this a NOP operation.
+    }
+}
+
+/// A unique id for file descriptions. While we could use the address, considering that
+/// is definitely unique, the address would expose interpreter internal state when used
+/// for sorting things. So instead we generate a unique id per file description that stays
+/// the same even if a file descriptor is duplicated and gets a new integer file descriptor.
```
RalfJung, line 281:
```suggestion
/// for sorting things. So instead we generate a unique id per file description is the name
/// for all `dup`licates and is never reused.
```
(Yes this is pre-existing, but we might as well improve this while we are at it)

src/shims/files.rs
```diff
@@ -99,25 +62,14 @@ pub trait FileDescription: std::fmt::Debug + Any {
...
         false
     }
 
-    /// Check the readiness of file description.
-    fn get_epoll_ready_events<'tcx>(
-        &self,
-    ) -> InterpResult<'tcx, crate::shims::unix::linux::epoll::EpollReadyEvents> {
-        throw_unsup_format!("{}: epoll does not support this file description", self.name());
+    fn as_unix<'tcx>(&self) -> InterpResult<'tcx, &dyn UnixFileDescription> {
+        throw_unsup_format!("Not a unix file descriptor: {}", self.name());
```
RalfJung, line 76: Shouldn't this be a panic? It would be a Miri bug to have a non-Unix FD oif the current target is a Unix target.

src/shims/files.rs
```diff
@@ -0,0 +1,411 @@
...
+impl Deref for FileDescriptionRef {
+    type Target = dyn FileDescription;
+
+    fn deref(&self) -> &Self::Target {
+        &*self.0.file_description
+    }
+}
+
+impl FileDescriptionRef {
+    pub fn new(fd: impl FileDescription, id: FdId) -> Self {
```
RalfJung, line 223: Why did you change this to `pub`?

src/shims/files.rs
CraftSpider: Yes, probably. There wasn't a reason behind the choice of `unsup`, I just didn't know what kind of error this should be.

RalfJung: Okay, then please make the return type `&dyn UnixFileDescription`, and "wrong" FD types should panic.

src/shims/files.rs
CraftSpider: Holdover from previous implementation details - removed

RalfJung:
Looks good, thanks! Please squash the last 2 commits in the previous ones (but definitely leave the first commit separate).

PR #4046, opened 2024-11-21
Fill out windows io error mapping table

CraftSpider:
Adds a bunch more of the Windows error codes to the error mapping list. This is all the error codes currently recognized by std IO decoding (https://github.com/rust-lang/rust/blob/75703c1a78c3cd99fa1347b237f3966fb8860e98/library/std/src/sys/pal/windows/mod.rs#L74). When multiple errors map to the same io value, I chose to prefer-non WSA and the most generic sounding variant for the backwards mapping.


RalfJung:
Thanks!

>  When multiple errors map to the same io value, I chose to prefer-non WSA and the most generic sounding variant for the backwards mapping.

Please add a comment in the code documenting this choice.

cgettys-microsoft:

PR #4053, opened 2024-11-23
sysconf adding few more constants.

devnexen:
e.g. _SC_NGROUPS_MAX can be useful for socket ancillary data.


src/shims/unix/foreign_items.rs
```diff
@@ -58,6 +58,16 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             // 512 seems to be a reasonable default. The value is not critical, in
             // the sense that getpwuid_r takes and checks the buffer length.
             ("_SC_GETPW_R_SIZE_MAX", |this| Scalar::from_int(512, this.pointer_size())),
+            // the machine's fd table does not start with a pre-allocated size but just the
+            // standard streams so we can't use it here and the spec imposes
+            // a minimum of `_POSIX_OPEN_MAX` (20).
+            ("_SC_OPEN_MAX", |this| Scalar::from_int(20, this.pointer_size())),
```
RalfJung, line 64: 20 seems really really low here. Why would we report such a low limit?

src/shims/unix/foreign_items.rs
```diff
@@ -58,6 +58,16 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // The spec imposes a minimum of `_POSIX_ARG_MAX` (4096).
+            ("_SC_ARG_MAX", |this| Scalar::from_int(4096, this.pointer_size())),
```
RalfJung, line 66: We don't support exec anyway so I don't think there's any point in supporting this constant.

src/shims/unix/foreign_items.rs
```diff
@@ -58,6 +58,16 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // The spec imposes a minimum of `_POSIX_CHILD_MAX` (25).
+            ("_SC_CHILD_MAX", |this| Scalar::from_int(25, this.pointer_size())),
```
RalfJung, line 68: We don't support process spawning so it seems pointless to support this.

src/shims/unix/foreign_items.rs
```diff
@@ -58,6 +58,16 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // 16 here is the least common denominator (i.e. macos, solarish).
+            ("_SC_NGROUPS_MAX", |this| Scalar::from_int(16, this.pointer_size())),
```
RalfJung, line 70: Not sure how this is going to be helpful for any program Miri supports?

RalfJung (commented):
Thanks for the PR! However, I think except for _SC_OPEN_MAX it makes little sense for Miri to support them. Unless you have real-world evidence that this is useful, I'd rather not add those.

src/shims/unix/foreign_items.rs
devnexen: I can set to 1024 (what linux returns) some other platforms return higher values. wdyt ?

src/shims/unix/foreign_items.rs
devnexen: that was just "anticipation" I planned try to use the value for [this line instead of the hardcoded 1](https://github.com/rust-lang/rust/blob/386a7c7ae22a3727c6a06633212d5a95edabdd74/library/std/src/os/unix/net/ancillary.rs#L99) to see if this help ; not urgent tough.

src/shims/unix/foreign_items.rs
RalfJung: Isn't it dynamic on Linux, determined by `ulimit`?

src/shims/unix/foreign_items.rs
RalfJung: `net` is anyway not supported by Miri, so seems a bit premature to add that constant already.

src/shims/unix/foreign_items.rs
devnexen: 1024 being the max here indeed. but then 20 is to low for you so what do you propose ?

RalfJung: The `ulimit` can easily be much higher, like 64k.

devnexen: on illumos it s 65536 on my debian it is only 1024 by default with unlimited but as root I can push it higher indeed.

RalfJung: But I guess 1024 is not an unreasonable value. The comment should make it clear that this is completely arbitrary and we can adjust if it that turns out to be helpful.

src/shims/unix/foreign_items.rs
```diff
@@ -58,6 +58,10 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             // 512 seems to be a reasonable default. The value is not critical, in
             // the sense that getpwuid_r takes and checks the buffer length.
             ("_SC_GETPW_R_SIZE_MAX", |this| Scalar::from_int(512, this.pointer_size())),
+            // the machine's fd table does not start with a pre-allocated size but just the
+            // standard streams so we can't use it here and the spec imposes
+            // a minimum of `_POSIX_OPEN_MAX` (20).
+            ("_SC_OPEN_MAX", |this| Scalar::from_int(65536, this.pointer_size())),
```
RalfJung, line 64:
```suggestion
            // Miri doesn't have a fixed limit on FDs, but we may be limited in terms of how
            // many *host* FDs we can open. Just use some arbitrary, pretty big value;
            // this can be adjusted if it causes problems.
            // The spec imposes a minimum of `_POSIX_OPEN_MAX` (20).
            ("_SC_OPEN_MAX", |this| Scalar::from_int(2.pow(16), this.pointer_size())),
```

PR #4052, opened 2024-11-22
sysconf interception fix for solarish systems.

devnexen:
also adding the `_SC_PAGESIZE` alias `_SC_PAGE_SIZE` supported by Linux, macOS and FreeBSD.

close #4050

PR #4054, opened 2024-11-23
follow-up on #4052, making a miri evaluation context fn for strerror_r.

devnexen:


src/shims/unix/foreign_items.rs
```diff
@@ -724,21 +748,6 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // We do not support forking, so there is nothing to do here.
                 this.write_null(dest)?;
             }
-            "strerror_r" | "__xpg_strerror_r" => {
```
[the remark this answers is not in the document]
devnexen: yes makes sense.

src/shims/unix/linux/foreign_items.rs
```diff
@@ -125,11 +125,11 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.write_scalar(result, dest)?;
             }
 
-            // Querying system information
-            "sysconf" => {
-                let [val] =
+            // String handling
+            "__xpg_strerror_r" => {
```
RalfJung, line 129: This should stay in "miscellaneous", where it also is in the "unix" file

tests/pass-dep/libc/libc-strerror_r.rs
```diff
@@ -0,0 +1,13 @@
+//@ignore-target: windows # Supported only on unixes
+
+fn main() {
+    unsafe {
+        let mut buf = vec![0u8; 32];
+        assert_eq!(libc::strerror_r(libc::EPERM, buf.as_mut_ptr().cast(), buf.len()), 0);
+        let mut buf2 = vec![0u8; 64];
+        assert_eq!(libc::strerror_r(-1i32, buf2.as_mut_ptr().cast(), buf2.len()), 0);
+        let mut buf3 = vec![0u8; 32];
+        // we purposely fill not enough length of the buffer to trigger ERANGE.
```
RalfJung, line 10:
```suggestion
        // This buffer is deliberately too small so this triggers ERANGE.
        let mut buf3 = vec![0u8; 32];
```

RalfJung (changes requested):

tests/pass-dep/libc/libc-strerror_r.rs
RalfJung: You didn't apply my suggestion. The comment should be above the `let mut`, not below -- it's odd for a comment to refer to things above it.

devnexen: well because it refers to the fact I use.2 bytes off 32, the buffer itself is large enough it is just "misused".

devnexen: it was just an example but whatever..

RalfJung (approved):

PR #4059, opened 2024-11-25
Simplify thread blocking tests

tiif:
Follow up from https://github.com/rust-lang/miri/pull/3939#discussion_r1855367789 to simplify the thread blocking test.

### DIFF
--- a/tests/pass-dep/libc/libc-eventfd.rs
+++ b/tests/pass-dep/libc/libc-eventfd.rs
@@ -197,7 +197,6 @@ fn test_two_threads_blocked_on_eventfd() {
     assert_eq!(res, 8);
 
     let thread1 = thread::spawn(move || {
-        thread::park();
         let sized_8_data = 1_u64.to_ne_bytes();
         let res: i64 = unsafe {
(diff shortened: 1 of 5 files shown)


tests/pass-dep/libc/libc-eventfd.rs
```diff
@@ -227,9 +224,7 @@ fn test_two_threads_blocked_on_eventfd() {
         assert_eq!(counter, (u64::MAX - 1));
     });
 
-    thread1.thread().unpark();
-    thread2.thread().unpark();
-    thread3.thread().unpark();
+    thread::yield_now();
```
RalfJung, line 227: This wouldn't even be needed. Calling `thread1.join()` will block the main thread, and thus move control to thread 1.

RalfJung:
What about the other tests doing similar things?
- tests/fail-dep/libc/eventfd_block_read_twice.rs
- tests/fail-dep/libc/eventfd_block_write_twice.rs

RalfJung:
Looks good, thanks. :) Please squash.

@rustbot author

PR #4061, opened 2024-11-26
Use `PathBuf` APIs to correctly do some path manipulation cross-platform

asquared31415:
Under certain weird configurations (for example running default powershell through the Zed editor built from source on Windows), the path manipulation could create a file path that looked like `\\?\C:\Path\To\Project/file.d`, and Windows does not accept a `/` for paths that use the `\\?\` syntax.

This would cause an error like `failed to create fake .d file: Os { code: 123, kind: InvalidFilename, message: "The filename, directory name, or volume label syntax is incorrect." }`


RalfJung:
Makes sense, thanks!

PR #4067, opened 2024-12-02
Support getting file metadata on Windows

CraftSpider:
This is two commits - one refactors handles a bit to make their usage more consistent and adds the file variant, the other actually adds the metadata shims.

### DIFF
--- a/src/shims/windows/foreign_items.rs
+++ b/src/shims/windows/foreign_items.rs
@@ -241,6 +241,33 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 };
                 this.write_scalar(result, dest)?;
             }
+            "CreateFileW" => {
+                let [
+                    file_name,
+                    desired_access,
+                    share_mode,
+                    security_attributes,
+                    creation_disposition,
+                    flags_and_attributes,
+                    template_file,
+                ] = this.check_shim(abi, ExternAbi::System { unwind: false }, link_name, args)?;
+                let handle = this.CreateFileW(
+                    file_name,
+                    desired_access,
+                    share_mode,
+                    security_attributes,
+                    creation_disposition,
+                    flags_and_attributes,
+                    template_file,
+                )?;
+                this.write_scalar(handle.to_scalar(this), dest)?;
+            }
+            "GetFileInformationByHandle" => {
+                let [handle, info] =
+                    this.check_shim(abi, ExternAbi::System { unwind: false }, link_name, args)?;
+                let res = this.GetFileInformationByHandle(handle, info)?;
+                this.write_scalar(res, dest)?;
+            }
 
             // Allocation
             "HeapAlloc" => {
--- a/src/shims/windows/foreign_items.rs
+++ b/src/shims/windows/foreign_items.rs
@@ -7,13 +7,8 @@ use rustc_span::Symbol;
 
 use self::shims::windows::handle::{Handle, PseudoHandle};
 use crate::shims::os_str::bytes_to_os_str;
-use crate::shims::windows::handle::HandleError;
 use crate::shims::windows::*;
 use crate::*;
-
-// The NTSTATUS STATUS_INVALID_HANDLE (0xC0000008) encoded as a HRESULT by setting the N bit.
-// (https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/0642cb2f-2075-4469-918c-4441e69c548a)
-const STATUS_INVALID_HANDLE: u32 = 0xD0000008;
 
 pub fn is_dyn_sym(name: &str) -> bool {
     // std does dynamic detection for these symbols
@@ -520,53 +515,38 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let [handle, name] =
                     this.check_shim(abi, ExternAbi::System { unwind: false }, link_name, args)?;
 
-                let handle = this.read_scalar(handle)?;
+                let handle = this.read_handle(handle)?;
                 let name = this.read_wide_str(this.read_pointer(name)?)?;
 
-                let thread = match Handle::try_from_scalar(handle, this)? {
-                    Ok(Handle::Thread(thread)) => Ok(thread),
-                    Ok(Handle::Pseudo(PseudoHandle::CurrentThread)) => Ok(this.active_thread()),
-                    Ok(_) | Err(HandleError::InvalidHandle) =>
-                        this.invalid_handle("SetThreadDescription")?,
-                    Err(HandleError::ThreadNotFound(e)) => Err(e),
+                let thread = match handle {
+                    Handle::Thread(thread) => thread,
+                    Handle::Pseudo(PseudoHandle::CurrentThread) => this.active_thread(),
+                    _ => this.invalid_handle("SetThreadDescription")?,
                 };
-                let res = match thread {
-                    Ok(thread) => {
-                        // FIXME: use non-lossy conversion
-                        this.set_thread_name(thread, String::from_utf16_lossy(&name).into_bytes());
-                        Scalar::from_u32(0)
-                    }
-                    Err(_) => Scalar::from_u32(STATUS_INVALID_HANDLE),
-                };
-
-                this.write_scalar(res, dest)?;
+                // FIXME: use non-lossy conversion
+                this.set_thread_name(thread, String::from_utf16_lossy(&name).into_bytes());
+                this.write_scalar(Scalar::from_u32(0), dest)?;
             }
             "GetThreadDescription" => {
                 let [handle, name_ptr] =
                     this.check_shim(abi, ExternAbi::System { unwind: false }, link_name, args)?;
 
-                let handle = this.read_scalar(handle)?;
+                let handle = this.read_handle(handle)?;
                 let name_ptr = this.deref_pointer(name_ptr)?; // the pointer where we should store the ptr to the name
 
-                let thread = match Handle::try_from_scalar(handle, this)? {
-                    Ok(Handle::Thread(thread)) => Ok(thread),
-                    Ok(Handle::Pseudo(PseudoHandle::CurrentThread)) => Ok(this.active_thread()),
-                    Ok(_) | Err(HandleError::InvalidHandle) =>
-                        this.invalid_handle("GetThreadDescription")?,
-                    Err(HandleError::ThreadNotFound(e)) => Err(e),
+                let thread = match handle {
+                    Handle::Thread(thread) => thread,
+                    Handle::Pseudo(PseudoHandle::CurrentThread) => this.active_thread(),
+                    _ => this.invalid_handle("GetThreadDescription")?,
                 };
-                let (name, res) = match thread {
-                    Ok(thread) => {
-                        // Looks like the default thread name is empty.
-                        let name = this.get_thread_name(thread).unwrap_or(b"").to_owned();
-                        let name = this.alloc_os_str_as_wide_str(
-                            bytes_to_os_str(&name)?,
-                            MiriMemoryKind::WinLocal.into(),
-                        )?;
-                        (Scalar::from_maybe_pointer(name, this), Scalar::from_u32(0))
-                    }
-                    Err(_) => (Scalar::null_ptr(this), Scalar::from_u32(STATUS_INVALID_HANDLE)),
-                };
+                // Looks like the default thread name is empty.
+                let name = this.get_thread_name(thread).unwrap_or(b"").to_owned();
+                let name = this.alloc_os_str_as_wide_str(
+                    bytes_to_os_str(&name)?,
+                    MiriMemoryKind::WinLocal.into(),
+                )?;
+                let name = Scalar::from_maybe_pointer(name, this);
+                let res = Scalar::from_u32(0);
 
                 this.write_scalar(name, &name_ptr)?;
                 this.write_scalar(res, dest)?;
@@ -667,11 +647,11 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     this.check_shim(abi, ExternAbi::System { unwind: false }, link_name, args)?;
                 this.check_no_isolation("`GetModuleFileNameW`")?;
 
-                let handle = this.read_target_usize(handle)?;
+                let handle = this.read_handle(handle)?;
                 let filename = this.read_pointer(filename)?;
                 let size = this.read_scalar(size)?.to_u32()?;
 
-                if handle != 0 {
+                if handle != Handle::Null {
                     throw_unsup_format!("`GetModuleFileNameW` only supports the NULL handle");
                 }
 
--- a/src/shims/windows/fs.rs
+++ b/src/shims/windows/fs.rs
@@ -1,0 +1,392 @@
+use std::fs::{File, Metadata, OpenOptions};
+use std::io;
+use std::io::{IsTerminal, Read, Seek, SeekFrom, Write};
+use std::path::{Path, PathBuf};
+use std::time::SystemTime;
+
+use rustc_abi::Size;
+
+use crate::shims::files::{EvalContextExt as _, FileDescription, FileDescriptionRef};
+use crate::shims::time::system_time_to_duration;
+use crate::shims::windows::handle::{EvalContextExt as _, Handle};
+use crate::*;
+
+#[derive(Debug)]
+pub struct FileHandle {
+    pub(crate) file: File,
+    pub(crate) writable: bool,
+}
+
+impl FileDescription for FileHandle {
+    fn name(&self) -> &'static str {
+        "file"
+    }
+
+    fn read<'tcx>(
+        &self,
+        _self_ref: &FileDescriptionRef,
+        communicate_allowed: bool,
+        ptr: Pointer,
+        len: usize,
+        dest: &MPlaceTy<'tcx>,
+        ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx> {
+        assert!(communicate_allowed, "isolation should have prevented even opening a file");
+        let mut bytes = vec![0; len];
+        let result = (&mut &self.file).read(&mut bytes);
+        match result {
+            Ok(read_size) => ecx.return_read_success(ptr, &bytes, read_size, dest),
+            Err(e) => ecx.set_last_error_and_return(e, dest),
+        }
+    }
+
+    fn write<'tcx>(
+        &self,
+        _self_ref: &FileDescriptionRef,
+        communicate_allowed: bool,
+        ptr: Pointer,
+        len: usize,
+        dest: &MPlaceTy<'tcx>,
+        ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx> {
+        assert!(communicate_allowed, "isolation should have prevented even opening a file");
+        let bytes = ecx.read_bytes_ptr_strip_provenance(ptr, Size::from_bytes(len))?;
+        let result = (&mut &self.file).write(bytes);
+        match result {
+            Ok(write_size) => ecx.return_write_success(write_size, dest),
+            Err(e) => ecx.set_last_error_and_return(e, dest),
+        }
+    }
+
+    fn seek<'tcx>(
+        &self,
+        communicate_allowed: bool,
+        offset: SeekFrom,
+    ) -> InterpResult<'tcx, io::Result<u64>> {
+        assert!(communicate_allowed, "isolation should have prevented even opening a file");
+        interp_ok((&mut &self.file).seek(offset))
+    }
+
+    fn close<'tcx>(
+        self: Box<Self>,
+        communicate_allowed: bool,
+        _ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, io::Result<()>> {
+        assert!(communicate_allowed, "isolation should have prevented even opening a file");
+        // We sync the file if it was opened in a mode different than read-only.
+        if self.writable {
+            // `File::sync_all` does the checks that are done when closing a file. We do this to
+            // to handle possible errors correctly.
+            let result = self.file.sync_all();
+            // Now we actually close the file and return the result.
+            drop(*self);
+            interp_ok(result)
+        } else {
+            // We drop the file, this closes it but ignores any errors
+            // produced when closing it. This is done because
+            // `File::sync_all` cannot be done over files like
+            // `/dev/urandom` which are read-only. Check
+            // https://github.com/rust-lang/miri/issues/999#issuecomment-568920439
+            // for a deeper discussion.
+            drop(*self);
+            interp_ok(Ok(()))
+        }
+    }
+
+    fn metadata<'tcx>(&self) -> InterpResult<'tcx, io::Result<Metadata>> {
+        interp_ok(self.file.metadata())
+    }
+
+    fn is_tty(&self, communicate_allowed: bool) -> bool {
+        communicate_allowed && self.file.is_terminal()
+    }
+}
+
+#[derive(Debug)]
+pub struct DirHandle {
+    pub(crate) path: PathBuf,
+}
+
+impl FileDescription for DirHandle {
+    fn name(&self) -> &'static str {
+        "directory"
+    }
+
+    fn metadata<'tcx>(&self) -> InterpResult<'tcx, io::Result<Metadata>> {
+        interp_ok(self.path.metadata())
+    }
+}
+
+#[derive(Debug)]
+pub struct MetadataHandle {
+    pub(crate) path: PathBuf,
+}
+
+impl FileDescription for MetadataHandle {
+    fn name(&self) -> &'static str {
+        "metadata-only"
+    }
+
+    fn metadata<'tcx>(&self) -> InterpResult<'tcx, io::Result<Metadata>> {
+        interp_ok(self.path.metadata())
+    }
+}
+
+impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
+#[allow(non_snake_case)]
+pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    fn CreateFileW(
+        &mut self,
+        file_name: &OpTy<'tcx>,            // LPCWSTR
+        desired_access: &OpTy<'tcx>,       // DWORD
+        share_mode: &OpTy<'tcx>,           // DWORD
+        security_attributes: &OpTy<'tcx>,  // LPSECURITY_ATTRIBUTES
+        creation_disposition: &OpTy<'tcx>, // DWORD
+        flags_and_attributes: &OpTy<'tcx>, // DWORD
+        template_file: &OpTy<'tcx>,        // HANDLE
+    ) -> InterpResult<'tcx, Handle> {
+        // ^ Returns HANDLE
+        let this = self.eval_context_mut();
+        this.assert_target_os("windows", "CreateFileW");
+        this.check_no_isolation("`CreateFileW`")?;
+
+        let file_name =
+            String::from_utf16_lossy(&this.read_wide_str(this.read_pointer(file_name)?)?);
+        let file_name = Path::new(&file_name);
+        let desired_access = this.read_scalar(desired_access)?.to_u32()?;
+        let share_mode = this.read_scalar(share_mode)?.to_u32()?;
+        let security_attributes = this.read_pointer(security_attributes)?;
+        let creation_disposition = this.read_scalar(creation_disposition)?.to_u32()?;
+        let flags_and_attributes = this.read_scalar(flags_and_attributes)?.to_u32()?;
+        let template_file = this.read_target_usize(template_file)?;
+
+        let generic_read = this.eval_windows_u32("c", "GENERIC_READ");
+        let generic_write = this.eval_windows_u32("c", "GENERIC_WRITE");
+
+        if desired_access & !(generic_read | generic_write) != 0 {
+            throw_unsup_format!("CreateFileW: Unsupported access mode: {desired_access}");
+        }
+
+        let file_share_delete = this.eval_windows_u32("c", "FILE_SHARE_DELETE");
+        let file_share_read = this.eval_windows_u32("c", "FILE_SHARE_READ");
+        let file_share_write = this.eval_windows_u32("c", "FILE_SHARE_WRITE");
+
+        if share_mode & !(file_share_delete | file_share_read | file_share_write) != 0
+            || share_mode == 0
+        {
+            throw_unsup_format!("CreateFileW: Unsupported share mode: {share_mode}");
+        }
+        if !this.ptr_is_null(security_attributes)? {
+            throw_unsup_format!("CreateFileW: Security attributes are not supported");
+        }
+
+        let create_always = this.eval_windows_u32("c", "CREATE_ALWAYS");
+        let create_new = this.eval_windows_u32("c", "CREATE_NEW");
+        let open_always = this.eval_windows_u32("c", "OPEN_ALWAYS");
+        let open_existing = this.eval_windows_u32("c", "OPEN_EXISTING");
+        let truncate_existing = this.eval_windows_u32("c", "TRUNCATE_EXISTING");
+
+        if ![create_always, create_new, open_always, open_existing, truncate_existing]
+            .contains(&creation_disposition)
+        {
+            throw_unsup_format!(
+                "CreateFileW: Unsupported creation disposition: {creation_disposition}"
+            );
+        }
+
+        let file_attribute_normal = this.eval_windows_u32("c", "FILE_ATTRIBUTE_NORMAL");
+        // This must be passed to allow getting directory handles. If not passed, we error on trying
+        // to open directories below
+        let file_flag_backup_semantics = this.eval_windows_u32("c", "FILE_FLAG_BACKUP_SEMANTICS");
+        let file_flag_open_reparse_point =
+            this.eval_windows_u32("c", "FILE_FLAG_OPEN_REPARSE_POINT");
+
+        let flags_and_attributes = match flags_and_attributes {
+            0 => file_attribute_normal,
+            _ => flags_and_attributes,
+        };
+        if !(file_attribute_normal | file_flag_backup_semantics | file_flag_open_reparse_point)
+            & flags_and_attributes
+            != 0
+        {
+            throw_unsup_format!(
+                "CreateFileW: Unsupported flags_and_attributes: {flags_and_attributes}"
+            );
+        }
+
+        if flags_and_attributes & file_flag_open_reparse_point != 0
+            && creation_disposition == create_always
+        {
+            throw_machine_stop!(TerminationInfo::Abort("Invalid CreateFileW argument combination: FILE_FLAG_OPEN_REPARSE_POINT with CREATE_ALWAYS".to_string()));
+        }
+
+        if template_file != 0 {
+            throw_unsup_format!("CreateFileW: Template files are not supported");
+        }
+
+        let desired_read = desired_access & generic_read != 0;
+        let desired_write = desired_access & generic_write != 0;
+        let exists = file_name.exists();
+        let is_dir = file_name.is_dir();
+
+        if flags_and_attributes == file_attribute_normal && is_dir {
+            this.set_last_error(IoError::WindowsError("ERROR_ACCESS_DENIED"))?;
+            return interp_ok(Handle::Invalid);
+        }
+
+        let mut options = OpenOptions::new();
+        if desired_read {
+            options.read(true);
+        }
+        if desired_write {
+            options.write(true);
+        }
+
+        if creation_disposition == create_always {
+            if file_name.exists() {
+                this.set_last_error(IoError::WindowsError("ERROR_ALREADY_EXISTS"))?;
+            }
+            options.create(true);
+            options.truncate(true);
+        } else if creation_disposition == create_new {
+            options.create_new(true);
+            if !desired_write {
+                options.append(true);
+            }
+        } else if creation_disposition == open_always {
+            if file_name.exists() {
+                this.set_last_error(IoError::WindowsError("ERROR_ALREADY_EXISTS"))?;
+            }
+            options.create(true);
+        } else if creation_disposition == open_existing {
+            // Nothing
+        } else if creation_disposition == truncate_existing {
+            options.truncate(true);
+        }
+
+        let handle = if is_dir && exists {
+            let fh = &mut this.machine.fds;
+            let fd = fh.insert_new(DirHandle { path: file_name.into() });
+            Ok(Handle::File(fd))
+        } else if creation_disposition == open_existing && desired_access == 0 {
+            // Windows supports handles with no permissions. These allow things such as reading
+            // metadata, but not file content.
+            let fh = &mut this.machine.fds;
+            let fd = fh.insert_new(MetadataHandle { path: file_name.into() });
+            Ok(Handle::File(fd))
+        } else {
+            options.open(file_name).map(|file| {
+                let fh = &mut this.machine.fds;
+                let fd = fh.insert_new(FileHandle { file, writable: desired_write });
+                Handle::File(fd)
+            })
+        };
+
+        match handle {
+            Ok(handle) => interp_ok(handle),
+            Err(e) => {
+                this.set_last_error(e)?;
+                interp_ok(Handle::Invalid)
+            }
+        }
+    }
+
+    fn GetFileInformationByHandle(
+        &mut self,
+        file: &OpTy<'tcx>,             // HANDLE
+        file_information: &OpTy<'tcx>, // LPBY_HANDLE_FILE_INFORMATION
+    ) -> InterpResult<'tcx, Scalar> {
+        // ^ Returns BOOL (i32 on Windows)
+        let this = self.eval_context_mut();
+        this.assert_target_os("windows", "GetFileInformationByHandle");
+        this.check_no_isolation("`GetFileInformationByHandle`")?;
+
+        let file = this.read_handle(file)?;
+        let file_information = this.deref_pointer_as(
+            file_information,
+            this.windows_ty_layout("BY_HANDLE_FILE_INFORMATION"),
+        )?;
+
+        let fd = if let Handle::File(fd) = file {
+            fd
+        } else {
+            this.invalid_handle("GetFileInformationByHandle")?
+        };
+
+        let Some(desc) = this.machine.fds.get(fd) else {
+            this.invalid_handle("GetFileInformationByHandle")?
+        };
+
+        let metadata = match desc.metadata()? {
+            Ok(meta) => meta,
+            Err(e) => {
+                this.set_last_error(e)?;
+                return interp_ok(this.eval_windows("c", "FALSE"));
+            }
+        };
+
+        let size = metadata.len();
+
+        let file_type = metadata.file_type();
+        let attributes = if file_type.is_dir() {
+            this.eval_windows_u32("c", "FILE_ATTRIBUTE_DIRECTORY")
+        } else if file_type.is_file() {
+            this.eval_windows_u32("c", "FILE_ATTRIBUTE_NORMAL")
+        } else {
+            this.eval_windows_u32("c", "FILE_ATTRIBUTE_DEVICE")
+        };
+
+        let created = extract_windows_epoch(metadata.created())?.unwrap_or((0, 0));
+        let accessed = extract_windows_epoch(metadata.accessed())?.unwrap_or((0, 0));
+        let written = extract_windows_epoch(metadata.modified())?.unwrap_or((0, 0));
+
+        this.write_int_fields_named(&[("dwFileAttributes", attributes.into())], &file_information)?;
+        write_filetime_field(this, &file_information, "ftCreationTime", created)?;
+        write_filetime_field(this, &file_information, "ftLastAccessTime", accessed)?;
+        write_filetime_field(this, &file_information, "ftLastWriteTime", written)?;
+        this.write_int_fields_named(
+            &[
+                ("dwVolumeSerialNumber", 0),
+                ("nFileSizeHigh", (size >> 32).into()),
+                ("nFileSizeLow", (size & 0xFFFFFFFF).into()),
+                ("nNumberOfLinks", 1),
+                ("nFileIndexHigh", 0),
+                ("nFileIndexLow", 0),
+            ],
+            &file_information,
+        )?;
+
+        interp_ok(this.eval_windows("c", "TRUE"))
+    }
+}
+
+/// Windows FILETIME is measured in 100-nanosecs since 1601
+fn extract_windows_epoch<'tcx>(
+    time: io::Result<SystemTime>,
+) -> InterpResult<'tcx, Option<(u32, u32)>> {
+    // (seconds in a year) * (369 years between 1970 and 1601) * 10 million (nanoseconds/second / 100)
+    const TIME_TO_EPOCH: u64 = 31_556_926 * 369 * 10_000_000;
+    match time.ok() {
+        Some(time) => {
+            let duration = system_time_to_duration(&time)?;
+            let secs = duration.as_secs().saturating_mul(10_000_000);
+            let nanos_hundred: u64 = (duration.subsec_nanos() / 100).into();
+            let total = secs.saturating_add(nanos_hundred).saturating_add(TIME_TO_EPOCH);
+            #[allow(clippy::cast_possible_truncation)]
+            interp_ok(Some((total as u32, (total >> 32) as u32)))
+        }
+        None => interp_ok(None),
+    }
+}
+
+fn write_filetime_field<'tcx>(
+    cx: &mut MiriInterpCx<'tcx>,
+    val: &MPlaceTy<'tcx>,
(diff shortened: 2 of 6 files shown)


tests/pass/shims/fs.rs
RalfJung, line 1: Could you also add a test that directly calls the relevant Windows API functions? std sometimes changes what it uses and then we lose test coverge, which is not great.

src/shims/windows/handle.rs
```diff
@@ -16,6 +17,8 @@ pub enum Handle {
     Null,
     Pseudo(PseudoHandle),
     Thread(ThreadId),
+    File(i32),
```
RalfJung, line 20: Are file IDs truly *signed*?

src/shims/windows/handle.rs
```diff
@@ -61,11 +68,16 @@ impl Handle {
             Self::Null => 0,
             Self::Pseudo(pseudo_handle) => pseudo_handle.value(),
             Self::Thread(thread) => thread.to_u32(),
+            #[expect(clippy::cast_sign_loss)]
+            Self::File(fd) => fd as u32,
+            Self::Invalid => 0x1FFFFFFF,
```
RalfJung, line 73: Where does this constant come from?

src/shims/windows/handle.rs
```diff
@@ -61,11 +68,16 @@ impl Handle {
...
         }
     }
 
     fn packed_disc_size() -> u32 {
         // ceil(log2(x)) is how many bits it takes to store x numbers
+        // We ensure that INVALID_HANDLE_VALUE (0xFFFFFFFF) decodes to Handle::Invalid
+        // see https://devblogs.microsoft.com/oldnewthing/20230914-00/?p=108766
```
RalfJung, line 80: When a comment is more than one sentence, please use proper punctuation. (Ideally this is true even for one-sentence comments, but we don't always do this in existing code.)

src/shims/windows/handle.rs
```diff
@@ -93,7 +105,7 @@ impl Handle {
         assert!(discriminant < 2u32.pow(disc_size));
 
         // make sure the data fits into `data_size` bits
-        assert!(data < 2u32.pow(data_size));
+        assert!(data <= 2u32.pow(data_size));
```
RalfJung, line 108: This change seems wrong? `2^data_size` does not fit in `data_size` bits.

src/shims/windows/handle.rs
```diff
@@ -47,12 +50,16 @@ impl Handle {
     const NULL_DISCRIMINANT: u32 = 0;
     const PSEUDO_DISCRIMINANT: u32 = 1;
     const THREAD_DISCRIMINANT: u32 = 2;
+    const FILE_DISCRIMINANT: u32 = 3;
+    const INVALID_DISCRIMINANT: u32 = 7;
```
RalfJung, line 54: Why jump from 3 to 7?

src/shims/windows/foreign_items.rs
```diff
@@ -520,53 +515,38 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                        this.invalid_handle("SetThreadDescription")?,
-                    Err(HandleError::ThreadNotFound(e)) => Err(e),
-                };
-                let res = match thread {
-                    Ok(thread) => {
-                        // FIXME: use non-lossy conversion
-                        this.set_thread_name(thread, String::from_utf16_lossy(&name).into_bytes());
-                        Scalar::from_u32(0)
-                    }
-                    Err(_) => Scalar::from_u32(STATUS_INVALID_HANDLE),
```
RalfJung, line 539: This isn't just refactoring, you are changing behavior here for the case where the thread ID is for a non-existent thread. Why?

src/shims/windows/handle.rs
```diff
@@ -171,6 +186,26 @@ impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
...
+    fn read_handle(&self, handle: &OpTy<'tcx>) -> InterpResult<'tcx, Handle> {
+        let this = self.eval_context_ref();
+        let handle = this.read_scalar(handle)?;
+        match Handle::try_from_scalar(handle, this)? {
+            Ok(handle) => interp_ok(handle),
+            Err(HandleError::InvalidHandle) =>
+                throw_machine_stop!(TerminationInfo::Abort(format!(
+                    "invalid handle {} at {}",
+                    handle.to_target_isize(this)?,
+                    Location::caller(),
```
RalfJung, line 199: This would print some Miri file names. That's implementation details we should not leak.

The typical thing we do here is have the function take an `&str` so that we can have errors like "invalid handle in `GetThreadDescription`".

src/shims/windows/handle.rs
```diff
@@ -180,12 +215,24 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-        let ret = match Handle::try_from_scalar(handle, this)? {
-            Ok(Handle::Thread(thread)) => {
+        let handle = this.read_handle(handle_op)?;
+        let ret = match handle {
+            Handle::Thread(thread) => {
                 this.detach_thread(thread, /*allow_terminated_joined*/ true)?;
                 this.eval_windows("c", "TRUE")
             }
+            Handle::File(fd) =>
+                if let Some(file) = this.machine.fds.get(fd) {
```
RalfJung, line 225: Shouldn't this be `fds.remove`?

Currently it will duplicate the FD and then close the duplicate.

src/shims/windows/handle.rs
```diff
@@ -180,12 +215,24 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-        let handle = this.read_scalar(handle_op)?;
-        let ret = match Handle::try_from_scalar(handle, this)? {
-            Ok(Handle::Thread(thread)) => {
+        let handle = this.read_handle(handle_op)?;
+        let ret = match handle {
+            Handle::Thread(thread) => {
                 this.detach_thread(thread, /*allow_terminated_joined*/ true)?;
                 this.eval_windows("c", "TRUE")
             }
+            Handle::File(fd) =>
```
RalfJung, line 224: In the Unix part of Miri, we usually use `fd` for "file description", and `fd_num` for "file descriptor". So please don't use `fd` for the numbers here, that could get confusing.

tests/pass/shims/fs.rs
```diff
@@ -18,23 +17,25 @@ mod utils;
...
-    // solarish needs to support readdir/readdir64 for these tests.
-    if cfg!(not(any(target_os = "solaris", target_os = "illumos"))) {
-        test_directory();
-        test_canonicalize();
+    if cfg!(not(windows)) {
+        test_file();
+        test_file_create_new();
+        test_seek();
+        test_file_clone();
+        test_metadata();
```
RalfJung, line 25: Shouldn't `test_metadata` be executed since you now support getting metadata?

`test_path_conversion` doesn't test anything interesting, I think it doesn't even call any shims.

src/shims/windows/fs.rs
```diff
@@ -152,7 +153,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
         let file_name =
             String::from_utf16_lossy(&this.read_wide_str(this.read_pointer(file_name)?)?);
-        let file_name = Path::new(&file_name);
+        let file_name = local_path(&file_name);
```
RalfJung, line 156: This does not look right, you should not have to write any new logic here.

Did you try `this.read_path_from_wide_str`? That is the intended function for cases like this.

Also, why are you going through `String`?  `PathBuf` should generally be used for host file names, and the Rust file APIs typically support that.

RalfJung (commented):
Thanks! Here are some comments for the first and last commit, I skipped the second one for now.

src/shims/windows/handle.rs
CraftSpider: Not really, it just made the types line up better.

src/shims/windows/handle.rs
CraftSpider: @ChrisDenton in Discord, same with the seven tag - I may be wrong, but as I understood it invalid tag is all ones so these values make that still true under miri

src/shims/windows/foreign_items.rs
CraftSpider: Consistency, before the PR that added this we always threw an error on invalid handle. And based on what I was told, it's also standard for debugging to make invalid handle a break. So I figured that was reasonable behavior to standardize on.

tests/pass/shims/fs.rs
CraftSpider: It does, actually! It's where I started on shims, since it failed, and apparently under the hood it just calls platform `metadata`. But I'll hoist this function too

src/shims/windows/fs.rs
CraftSpider: I added this to fix test failures on Linux and Mac.

Windows file names are (at least sometimes) prefixed with `\\?\` to make them able to go over the old max path size, or may include the drive like `C:\`. This prefix must be stripped before the path is passed to the actual local APIs, or else it will cause errors on unix (since PathBuf assumes paths are local, AFAIK). The other way around (emulating unix on Windows) doesn't need this because Windows gracefully handles `/` as a secondary separator, and a starting one as root of the current drive.

src/shims/windows/foreign_items.rs
RalfJung: That makes sense. In the future please mention all intended behavior changes in the PR description, that makes review a lot easier because I don't have to reverse engineer the diff quite as much. ;)

src/shims/windows/handle.rs
RalfJung: Miri generally has its own handle encoding, but if "invalid handle is all ones" is a guarantee that we (have to) provide, then please
- add a comment saying that on every place that matters (the choice of 7, and right nere)
- add a test checking this

src/shims/windows/handle.rs
RalfJung: I see several casts between `i32` and `u32` in this PR. In which sense do the types line up better with `i32`?

src/shims/windows/fs.rs
RalfJung: `read_path_from_wide_str` handled all that already. At least, it should.

src/shims/windows/handle.rs
CraftSpider: There were a lot more without it - most uses pass the contained number to FdMap, which expects `i32`, while this just needs it at the Handle/Scalar boundary. I can mock up the diff for that change when I get a chance, if you'd like to see the difference (none also that with that model, all new shims that deal with files will also be adding casts, but this way they mostly don't)

src/shims/windows/fs.rs
CraftSpider: Ah, cool. Sorry I missed that, I'll swap to that function

src/shims/windows/handle.rs
RalfJung: Ah, they are, true.

We should probably have at least a type alias to make the intent more clear, like
```rust
type FdNum = i32;
```
(A newtype might be better but is a lot more work so let's hold off on that.)

src/shims/windows/handle.rs
```diff
@@ -105,6 +117,9 @@ impl Handle {
             Self::NULL_DISCRIMINANT if data == 0 => Some(Self::Null),
             Self::PSEUDO_DISCRIMINANT => Some(Self::Pseudo(PseudoHandle::from_value(data)?)),
             Self::THREAD_DISCRIMINANT => Some(Self::Thread(ThreadId::new_unchecked(data))),
+            #[expect(clippy::cast_possible_wrap)]
+            Self::FILE_DISCRIMINANT => Some(Self::File(data as i32)),
```
RalfJung, line 121: I wish we had some nice way to assert that this cast is not truncating. In lieu of that, maybe let's do this?
```suggestion
            Self::FILE_DISCRIMINANT => {
                // This cast preserves all bits.
                assert_eq!(size_of_val(&data), size_of::<i32>());
                Some(Self::File(data as i32))
            }
```

src/shims/windows/handle.rs
CraftSpider: Windows documentation doesn't have a page for INVALID_HANDLE_VALUE explicitly pinning it to -1, but several other pages do make the claim that it will be that value:

https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/get-osfhandle?view=msvc-170
https://learn.microsoft.com/en-us/dotnet/api/microsoft.win32.safehandles.safefilehandle?view=net-9.0

I'm not enough of an expert to say for sure, but it seems like the kind of thing that code might rely on even if it's strictly considered an implementation detail. I'm adding docs currently, just explaining my reasoning for this choice.

tests/pass/shims/fs.rs
CraftSpider: After trying the move, it's not really possible while keeping this PR what it is - test_metadata expects the ability to delete and write to files, as it uses the standard `prepare_with_content` utility to set up a test file.

CraftSpider:
Before we get too much further on this PR, I'd like to point out https://github.com/rust-lang/miri/pull/4043/commits/296d806930304c03f2cba91cc8a2fed4a2f38d16 on the big PR. It's the changes needed to FileDescription methods to make them platform-agnostic enough for Windows use-cases. If they look fine, we can continue, but if you think it's a bad design then the FileDescription trait will need entirely split in two, which will change code even in this PR.

RalfJung:
Reducing the places where we have to use global state like errno seems like an improvement. But please use `Result<_, IoError>` (with Miri's `IoError` type).

tests/pass/shims/fs.rs
RalfJung: > It does, actually! It's where I started on shims, since it failed, and apparently under the hood it just calls platform metadata. But I'll hoist this function too

Ah, it calls `is_dir`. Seems good then, thanks.

tests/pass/shims/fs.rs
CraftSpider: Added a windows-fs test for the shims directly

tests/pass/shims/windows-fs.rs
```diff
@@ -0,0 +1,131 @@
...
+        dwFlagsAndAttributes: DWORD,
+        hTemplateFile: HANDLE,
+    ) -> HANDLE;
+    fn GetFileInformationByHandle(
+        handle: HANDLE,
+        file_info: *mut BY_HANDLE_FILE_INFORMATION,
+    ) -> BOOL;
+    fn CloseHandle(handle: HANDLE) -> BOOL;
+    fn GetLastError() -> DWORD;
+}
```
RalfJung, line 72: Is there some crate we can use to avoid having to carry all these definitions ourselves?

`pass-dep` tests can use the dependencies declared in `test_dependencies/Cargo.toml`.

src/shims/windows/handle.rs
RalfJung: Okay, so please add comments and a test then. I'd suggest a `#[test]` function in `handle.rs` that encodes the invalid handle and asserts it to be `u32::MAX`.

src/shims/windows/handle.rs
RalfJung: You resolved this but didn't actually change the code?

RalfJung: Ah you did, I just missed the commit. :)

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,392 @@
...
+        )?;
+
+        interp_ok(this.eval_windows("c", "TRUE"))
+    }
+}
+
+/// Windows FILETIME is measured in 100-nanosecs since 1601
+fn extract_windows_epoch<'tcx>(
+    time: io::Result<SystemTime>,
+) -> InterpResult<'tcx, Option<(u32, u32)>> {
```
RalfJung, line 380: We already have some logic that looks pretty similar in `GetSystemTimeAsFileTime`. Could that be reused?

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,392 @@
...
+    // (seconds in a year) * (369 years between 1970 and 1601) * 10 million (nanoseconds/second / 100)
+    const TIME_TO_EPOCH: u64 = 31_556_926 * 369 * 10_000_000;
+    match time.ok() {
+        Some(time) => {
+            let duration = system_time_to_duration(&time)?;
```
RalfJung, line 385: We should probably rename this to `system_time_since_unix_epoch`. (Can be a future PR though.)

RalfJung:
Okay I finally had a look at the core new file in this PR. :)
@rustbot author

tests/pass/shims/windows-fs.rs
```diff
@@ -0,0 +1,131 @@
...
+        test_create_normal_file();
+    }
+}
+
+unsafe fn test_create_dir_file() {
+    let temp = utils::tmp();
+    let mut raw_path = temp.as_os_str().encode_wide().collect::<Vec<_>>();
+    // encode_wide doesn't add a null-terminator
+    raw_path.push(0);
+    raw_path.push(0);
```
RalfJung, line 86: This seems like we want a "to WideCStr" helper or something like that.

tests/pass/shims/windows-fs.rs
CraftSpider: Ah, cool. I just followed the other *-windows shim tests, many of these are almost literally copied from `windows_sys`.

(I can add a follow up task to my to-do list for other windows shim tests defining their own API if we want)

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,392 @@
...
+use crate::shims::files::{EvalContextExt as _, FileDescription, FileDescriptionRef};
+use crate::shims::time::system_time_to_duration;
+use crate::shims::windows::handle::{EvalContextExt as _, Handle};
+use crate::*;
+
+#[derive(Debug)]
+pub struct FileHandle {
+    pub(crate) file: File,
+    pub(crate) writable: bool,
+}
```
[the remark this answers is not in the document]
CraftSpider: Actually, they started out needing to be different eventually, but with other changes... I think this might be a good candidate for shared implementation

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,392 @@
...
+
+    fn metadata<'tcx>(&self) -> InterpResult<'tcx, io::Result<Metadata>> {
+        interp_ok(self.path.metadata())
+    }
+}
+
+#[derive(Debug)]
+pub struct MetadataHandle {
+    pub(crate) path: PathBuf,
+}
```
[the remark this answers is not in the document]
CraftSpider: Windows supports handles that don't have any permissions to read or write the file. I wasn't sure how to best represent these (or normal directory handles) in a stable way. I can try making a File without read or write, I didn't really try that.

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,392 @@
...
+        flags_and_attributes: &OpTy<'tcx>, // DWORD
+        template_file: &OpTy<'tcx>,        // HANDLE
+    ) -> InterpResult<'tcx, Handle> {
+        // ^ Returns HANDLE
+        let this = self.eval_context_mut();
+        this.assert_target_os("windows", "CreateFileW");
+        this.check_no_isolation("`CreateFileW`")?;
+
+        let file_name =
+            String::from_utf16_lossy(&this.read_wide_str(this.read_pointer(file_name)?)?);
```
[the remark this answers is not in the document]
CraftSpider: Already fixed in a later commit

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,392 @@
...
+        interp_ok(this.eval_windows("c", "TRUE"))
+    }
+}
+
+/// Windows FILETIME is measured in 100-nanosecs since 1601
+fn extract_windows_epoch<'tcx>(
+    time: io::Result<SystemTime>,
+) -> InterpResult<'tcx, Option<(u32, u32)>> {
+    // (seconds in a year) * (369 years between 1970 and 1601) * 10 million (nanoseconds/second / 100)
+    const TIME_TO_EPOCH: u64 = 31_556_926 * 369 * 10_000_000;
```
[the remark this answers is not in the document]
CraftSpider: The seconds/year number currently isn't that multiplication - it's based on seconds/year numbers from places like this:
https://www.epochconverter.com/

It's basically just averaged/actual time per year, 365.24 days. It gave day-correct times when I tests, but it appears to actually be a little off the numbers given by the variant where you do exact seconds per year plus leap days in that period. The alternative way to do the calculation would be... `3600 * 24 * 365 * 369 + 3600 * 24 * 89` , which matches with what Python uses. I'll switch to using that calculation, as it matches with how at least one actual language does the calculation.

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,392 @@
...
+            if file_name.exists() {
+                this.set_last_error(IoError::WindowsError("ERROR_ALREADY_EXISTS"))?;
+            }
+            options.create(true);
+            options.truncate(true);
+        } else if creation_disposition == create_new {
+            options.create_new(true);
+            if !desired_write {
+                options.append(true);
+            }
```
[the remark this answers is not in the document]
CraftSpider: When using 'create_new', you must specify either 'write' or 'append'. See the last line of the function documentation:
https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.create_new

So when not writing, we set append.

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,392 @@
...
+            options.create(true);
+        } else if creation_disposition == open_existing {
+            // Nothing
+        } else if creation_disposition == truncate_existing {
+            options.truncate(true);
+        }
+
+        let handle = if is_dir && exists {
+            let fh = &mut this.machine.fds;
+            let fd = fh.insert_new(DirHandle { path: file_name.into() });
```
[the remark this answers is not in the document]
CraftSpider: Based on other tests, the temp directory is assumed to exist, so that should cover this branch. It's the metadata branch that isn't properly covered yet, I'll add another test for it.

src/shims/windows/fs.rs
CraftSpider: That is implementing the same logic it looks like, only using windows constants. I'd be down to extract that into a helper used both there and here instead of `extra_windows_epoch`, but I'm not sure if that should be down now or in a follow-up. For the moment I've left it alone.

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,392 @@
...
+        let creation_disposition = this.read_scalar(creation_disposition)?.to_u32()?;
+        let flags_and_attributes = this.read_scalar(flags_and_attributes)?.to_u32()?;
+        let template_file = this.read_target_usize(template_file)?;
+
+        let generic_read = this.eval_windows_u32("c", "GENERIC_READ");
+        let generic_write = this.eval_windows_u32("c", "GENERIC_WRITE");
+
+        if desired_access & !(generic_read | generic_write) != 0 {
+            throw_unsup_format!("CreateFileW: Unsupported access mode: {desired_access}");
+        }
```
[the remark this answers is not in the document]
CraftSpider: Updated what I could to that pattern, not sure if I can nicely do some of the other variables. Cleaner implementation ideas welcome, but if not it at least is better than it was.

tests/pass/shims/windows-fs.rs
RalfJung: > Ah, cool. I just followed the other *-windows shim tests, many of these are almost literally copied from windows_sys.

Yeah, fair. But if we're now getting more of these tests IMO we should start doing this the proper way. We also use the libc crate for Unix targets, so using the windows crate here is more consistent as well.

src/shims/windows/fs.rs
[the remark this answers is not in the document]
RalfJung: I assume these are still "stable" handles though, in that if the file gets renamed later the handle refers to the old file still?

If all you can do with these handles is get their metadata, maybe the best representation for them is that metadata.

src/shims/windows/fs.rs
RalfJung: Given our [other discussion](https://github.com/rust-lang/miri/pull/4067#discussion_r1874356908) around the constants, which would be resolved by using Windows constants, I think that should be part of this PR.

src/shims/windows/fs.rs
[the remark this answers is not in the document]
RalfJung: Please add a comment explaining that.

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,363 @@
...
+            // Nothing
+        } else if creation_disposition == truncate_existing {
+            options.truncate(true);
+        } else {
+            throw_unsup_format!(
+                "CreateFileW: Unsupported creation disposition: {creation_disposition}"
+            );
+        }
+
+        let handle = if is_dir && exists {
```
RalfJung, line 233: Isn't it the case that is_dir implies exists?

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,392 @@
...
+        let mut options = OpenOptions::new();
+        if desired_read {
+            options.read(true);
+        }
+        if desired_write {
+            options.write(true);
+        }
+
+        if creation_disposition == create_always {
+            if file_name.exists() {
```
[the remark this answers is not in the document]
CraftSpider: Fixed the second two locally. As for the first - we could always use `create_new`, then if it's an always variant, try again on `ErrorKind::AlreadyExists`. That doesn't really make it not racy though, just changes it from erroring even if the file is deleted after the check, to not erroring if the file is created after the check. I don't see a way to implement this without racing one way or the other using std APIs.

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,334 @@
...
+            throw_unsup_format!("CreateFileW: Security attributes are not supported");
+        }
+
+        let flags_and_attributes = match flags_and_attributes {
+            0 => file_attribute_normal,
+            _ => flags_and_attributes,
+        };
+        if !(file_attribute_normal | file_flag_backup_semantics | file_flag_open_reparse_point)
+            & flags_and_attributes
+            != 0
```
RalfJung, line 115: This is another one of these bitmasks where validation is separate from interpretation. I don't understand all these flags so I am not sure, can this be done in a better way? There should be one `if` for each of these 3 cases, performing whatever action the respective flag should perform.

Maybe Windows file flag parsing is so complicated that we should have a separate function that takes in the raw flaws as `u32`s and returns some nice typed representation?

`file_flag_backup_semantics` is sus because it is considered supported here but entirely ignored later.

src/shims/windows/fs.rs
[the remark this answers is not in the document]
RalfJung: Thanks! Unfortunately it is still quite hard to follow. Reminds me why I never wanted to deal with the windows API. ;)

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,334 @@
...
+            } else {
+                this.set_last_error(IoError::Raw(Scalar::from_u32(0)))?;
+            }
+            options.create(true);
+            options.truncate(true);
+        } else if creation_disposition == create_new {
+            options.create_new(true);
+            // Per `create_new` documentation:
+            // If the specified file does not exist and is a valid path to a writable location, the
+            // function creates a file and the last-error code is set to zero.
```
RalfJung, line 182: The setting to zero does not seem to happen here?

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,334 @@
...
+            // https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew
+            if exists {
+                this.set_last_error(IoError::WindowsError("ERROR_ALREADY_EXISTS"))?;
+            } else {
+                this.set_last_error(IoError::Raw(Scalar::from_u32(0)))?;
+            }
+            options.create(true);
+        } else if creation_disposition == open_existing {
+            // Nothing
```
RalfJung, line 201: Are you sure this does not reset the error code to 0?

Maybe we should just set it to 0 at the top, to make sure we don't forget?

src/shims/files.rs
```diff
@@ -275,6 +275,9 @@ impl VisitProvenance for WeakFileDescriptionRef {
     }
 }
 
+/// Internal type of a file-descriptor - this is what [`FdTable`] expects
+pub type FdNum = i32;
```
RalfJung, line 379: Please also use this type in the definition of `FdTable`

src/shims/windows/fs.rs
```diff
@@ -127,7 +80,7 @@ impl FileDescription for DirHandle {
 
 #[derive(Debug)]
 pub struct MetadataHandle {
```
RalfJung, line 34: Please add a doc comment explaining this handle type.

src/shims/windows/fs.rs
```diff
@@ -204,20 +202,46 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             );
         }
 
+        // This is racy, but there doesn't appear to be an std API that both succeeds if a file
+        // exists but tells us it isn't new. Either we accept racing one way or another, or we
+        // use an iffy heuristic like file creation time. This implementation prefers to fail in the
+        // direction of erroring more often.
+        let exists = file_name.exists();
```
RalfJung, line 162: This comment is too early, since one doesn't know yet that this is used for. IMO it'd be better placed at the `if exists` below.

(Also, no need to `let`-bind, I don't feel like this makes things much clearer.)

src/shims/windows/handle.rs
```diff
@@ -249,3 +249,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_invalid_encoding() {
+        // Ensure the invalid handle encodes to `u32::MAX`/`INVALID_HANDLE_VALUE`.
+        assert_eq!(Handle::Invalid.to_packed(), 0xFFFFFFFF)
```
RalfJung, line 260:
```suggestion
        assert_eq!(Handle::Invalid.to_packed(), u32::MAX)
```

src/shims/windows/fs.rs
CraftSpider: This is the wrong copy-paste - I meant to put the std docs for create_new here. Microsoft doesn't say anything about zero error here

RalfJung: If you have access to a Windows host, maybe you can try this:
- make sure the last error is non-zero
- then call CreateFile in one of the cases where the docs say nothing about the last error, and see what the last error looks like when the function returns successfully

src/shims/windows/fs.rs
```diff
@@ -53,6 +59,88 @@ impl FileDescription for MetadataHandle {
...
+
+bitflags! {
+    #[derive(PartialEq)]
+    struct FileAttributes: u32 {
+        const ZERO = 0;
+        const NORMAL = 1 << 0;
+        /// This must be passed to allow getting directory handles. If not passed, we error on trying
+        /// to open directories
+        const BACKUP_SEMANTICS = 1 << 1;
+        const OPEN_REPARSE = 1 << 2;
```
RalfJung, line 107: Can you doc-comment this one as well? "reparse" is not standard terminology outside the Windows world.

src/shims/windows/fs.rs
```diff
@@ -53,6 +59,88 @@ impl FileDescription for MetadataHandle {
...
+            out |= FileAttributes::NORMAL;
+        }
+
+        if value != 0 {
+            throw_unsup_format!("CreateFileW: Unsupported flags_and_attributes: {value}");
+        }
+
+        if out == FileAttributes::ZERO {
+            out = FileAttributes::NORMAL;
+        }
```
RalfJung, line 139: Why is this done?

src/shims/windows/fs.rs
```diff
@@ -155,63 +220,48 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-            // If the specified file exists, the function succeeds and the last-error code is set
-            // to ERROR_ALREADY_EXISTS.
-            // If the specified file does not exist and is a valid path to a writable location, the
-            // function creates a file and the last-error code is set to zero.
-            // https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew
-            if exists {
-                this.set_last_error(IoError::WindowsError("ERROR_ALREADY_EXISTS"))?;
-            } else {
-                this.set_last_error(IoError::Raw(Scalar::from_u32(0)))?;
+            OpenExisting => (), // Nothing
```
RalfJung, line 254:
```suggestion
            OpenExisting => {} // Default options
```

CraftSpider:
Amazing timing, me. I'm on vacation for the next week, but will rebase again after

RalfJung:
Oh, was this waiting for review? Sorry, that wasn't clear to me.  Please write `@rustbot review` once the PR is ready again.

CraftSpider:
No problem, I forgot to do that a while back, then didn't want to do it until I rebased, and somewhere along the way a test started failing... So I'll fix everything that fell out from a month of going stale, then mark it ready.

src/shims/windows/fs.rs
CraftSpider: It looks like CreateFileW is setting last error to 0 in non-error cases where the docs don't specify, locally. I'll update to set it to 0 at the beginning of the function.

src/shims/windows/fs.rs
CraftSpider: For some reason I thought I responded to this. Anyways, this is done because, per the docs, NORMAL means "The file does not have other attributes set". This also matches with my testing. Since 0 and NORMAL seem to mean the same thing, I thought it would simplify implementation to only ever have to check for one of them.

CraftSpider:
@RalfJung any update on this? It's fine if you're busy or such, just don't want it to get too stale again hopefully

RalfJung:
Sorry for the delay. It's in my queue, I have not lost track. :)

CraftSpider:
Thanks, sorry for the unnecessary ping then.

src/shims/windows/fs.rs
RalfJung: Okay -- please add a comment explaining this.

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,389 @@
...
+        let file_attribute_normal = ecx.eval_windows_u32("c", "FILE_ATTRIBUTE_NORMAL");
+        let file_flag_backup_semantics = ecx.eval_windows_u32("c", "FILE_FLAG_BACKUP_SEMANTICS");
+        let file_flag_open_reparse_point =
+            ecx.eval_windows_u32("c", "FILE_FLAG_OPEN_REPARSE_POINT");
+
+        let mut out = FileAttributes::ZERO;
+        if value & file_flag_backup_semantics != 0 {
+            value &= !file_flag_backup_semantics;
+            out |= FileAttributes::BACKUP_SEMANTICS;
+        } else if value & file_flag_open_reparse_point != 0 {
```
RalfJung, line 127: Why `else`?

if `FileAttributes` is always *one of* these, shouldn;t it be an `enum` rather than a bitflag?

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,389 @@
...
+        if attributes.contains(FileAttributes::OPEN_REPARSE) && creation_disposition == CreateAlways
+        {
+            throw_machine_stop!(TerminationInfo::Abort("Invalid CreateFileW argument combination: FILE_FLAG_OPEN_REPARSE_POINT with CREATE_ALWAYS".to_string()));
+        }
+
+        if template_file != 0 {
+            throw_unsup_format!("CreateFileW: Template files are not supported");
+        }
+
+        let is_dir = file_name.is_dir();
```
RalfJung, line 201: This also seems to be racy?

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,389 @@
...
+                // This is racy, but there doesn't appear to be an std API that both succeeds if a
+                // file exists but tells us it isn't new. Either we accept racing one way or another,
+                // or we use an iffy heuristic like file creation time. This implementation prefers
+                // to fail in the direction of erroring more often.
+                // Per the documentation:
+                // If the specified file exists and is writable, the function truncates the file,
+                // the function succeeds, and last-error code is set to ERROR_ALREADY_EXISTS.
+                // If the specified file does not exist and is a valid path, a new file is created,
+                // the function succeeds, and the last-error code is set to zero.
+                // https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew
```
RalfJung, line 238: Please put this first -- it is required background to understand the first part of this comment.

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,389 @@
...
+        flags_and_attributes: &OpTy<'tcx>, // DWORD
+        template_file: &OpTy<'tcx>,        // HANDLE
+    ) -> InterpResult<'tcx, Handle> {
+        // ^ Returns HANDLE
+        use CreationDisposition::*;
+
+        let this = self.eval_context_mut();
+        this.assert_target_os("windows", "CreateFileW");
+        this.check_no_isolation("`CreateFileW`")?;
+        this.set_last_error(IoError::Raw(Scalar::from_i32(0)))?;
```
RalfJung, line 165: Please add a comment explaining why we reset the error.

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,389 @@
...
+        let file_type = metadata.file_type();
+        let attributes = if file_type.is_dir() {
+            this.eval_windows_u32("c", "FILE_ATTRIBUTE_DIRECTORY")
+        } else if file_type.is_file() {
+            this.eval_windows_u32("c", "FILE_ATTRIBUTE_NORMAL")
+        } else {
+            this.eval_windows_u32("c", "FILE_ATTRIBUTE_DEVICE")
+        };
+
+        let created = extract_windows_epoch(this, metadata.created())?.unwrap_or((0, 0));
```
RalfJung, line 339: Why the fallback to `(0, 0)`?

tests/pass/shims/fs.rs
```diff
@@ -144,10 +145,10 @@ fn test_metadata() {
     let path = utils::prepare_with_content("miri_test_fs_metadata.txt", bytes);
 
     // Test that metadata of an absolute path is correct.
-    check_metadata(bytes, &path).unwrap();
+    check_metadata(bytes, &path).expect("Absolute path metadata");
```
RalfJung, line 148:
```suggestion
    check_metadata(bytes, &path).expect("absolute path metadata");
```

tests/pass/shims/fs.rs
```diff
@@ -144,10 +145,10 @@ fn test_metadata() {
...
     // Test that metadata of a relative path is correct.
     std::env::set_current_dir(path.parent().unwrap()).unwrap();
-    check_metadata(bytes, Path::new(path.file_name().unwrap())).unwrap();
+    check_metadata(bytes, Path::new(path.file_name().unwrap())).expect("Relative path metadata");
```
RalfJung, line 151:
```suggestion
    check_metadata(bytes, Path::new(path.file_name().unwrap())).expect("relative path metadata");
```

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,389 @@
...
+                // or we use an iffy heuristic like file creation time. This implementation prefers
+                // to fail in the direction of erroring more often.
+                // Per the documentation:
+                // If the specified file exists and is writable, the function truncates the file,
+                // the function succeeds, and last-error code is set to ERROR_ALREADY_EXISTS.
+                // If the specified file does not exist and is a valid path, a new file is created,
+                // the function succeeds, and the last-error code is set to zero.
+                // https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew
+                if file_name.exists() {
+                    this.set_last_error(IoError::WindowsError("ERROR_ALREADY_EXISTS"))?;
```
RalfJung, line 240: So we want this error to be set even if we ignore `options` in the code below?

tests/pass-dep/shims/windows-fs.rs
```diff
@@ -0,0 +1,170 @@
...
+        test_create_normal_file();
+        test_create_always_twice();
+        test_open_always_twice();
+    }
+}
+
+unsafe fn test_create_dir_file() {
+    let temp = utils::tmp();
+    let raw_path = to_wide_cstr(&temp);
+    let handle = CreateFileW(
```
RalfJung, line 33:
```suggestion
    // Open the `temp` directory.
    let handle = CreateFileW(
```

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,389 @@
...
+            throw_machine_stop!(TerminationInfo::Abort("Invalid CreateFileW argument combination: FILE_FLAG_OPEN_REPARSE_POINT with CREATE_ALWAYS".to_string()));
+        }
+
+        if template_file != 0 {
+            throw_unsup_format!("CreateFileW: Template files are not supported");
+        }
+
+        let is_dir = file_name.is_dir();
+
+        if !attributes.contains(FileAttributes::BACKUP_SEMANTICS) && is_dir {
```
RalfJung, line 203:
```suggestion
        // BACKUP_SEMANTICS is how Windows calls the act of opening a directory handle.
        if !attributes.contains(FileAttributes::BACKUP_SEMANTICS) && is_dir {
```
I still wonder what on Earth was in the water when they picked that name...

RalfJung (commented):
Thanks for the patience!

src/shims/unix/fd.rs
```diff
@@ -121,7 +121,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             throw_unsup_format!("unsupported flags {:#x}", op);
         };
 
-        let result = fd.as_unix().flock(this.machine.communicate(), parsed_op)?;
+        let result = fd.as_unix(this).flock(this.machine.communicate(), parsed_op)?;
+        drop(fd);
```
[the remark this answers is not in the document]
CraftSpider: I think this is a mistake of rebasing - I believe there used to be a drop here, or I needed to add one at some point, but it's no longer needed. I'll remove it.

src/shims/windows/handle.rs
```diff
@@ -171,6 +197,25 @@ impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 
 #[allow(non_snake_case)]
 pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    #[track_caller]
+    fn read_handle(&self, handle: &OpTy<'tcx>, function_name: &str) -> InterpResult<'tcx, Handle> {
+        let this = self.eval_context_ref();
+        let handle = this.read_scalar(handle)?;
+        match Handle::try_from_scalar(handle, this)? {
```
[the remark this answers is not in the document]
CraftSpider: I'm pretty sure everyone should just use `read_handle` currently. It ensures consistency, which is nice since previously the Windows shims didn't have that. I think them being different methods makes the code cleaner, but I'll make `try_from_scalar` non-public. It's also possible it will be useful in the future, if a shim ever needs to handle invalidity with something other than an abort.

src/shims/windows/fs.rs
CraftSpider: Mistake that happened to work. It can in fact be multiple, I just messed up.

src/shims/windows/fs.rs
CraftSpider: Yes, opening a directory with these flags should still set the error.

src/shims/windows/fs.rs
CraftSpider: I'm considering if there's a better way to handle this. Since we need to know it's a dir to open a dir handle anyways.

src/shims/windows/fs.rs
CraftSpider: From the docs for all of these fields:
"If the underlying file system does not support the last access time, this member is zero (0)."

src/shims/windows/handle.rs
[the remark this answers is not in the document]
RalfJung: Hm, I'm not convinced a separate `try_from_scalar` still carries its weight -- but I am fine keeping it for now and will (try to remember to) take a look at the code once this lands.

src/shims/windows/fs.rs
RalfJung: For now I'm also fine with a comment explaining this. If you have some ideas for a better way, you can also open an issue.

src/shims/windows/fs.rs
RalfJung: Please make sure to add a test that covers that case, as apparently it is not covered by the current tests.

src/shims/windows/handle.rs
```diff
@@ -171,6 +197,27 @@ impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 
 #[allow(non_snake_case)]
 pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    /// Convert a scalar into a structured `Handle`.
+    /// If the handle is invalid, or references a non-existent item, this returns a machine abort.
```
RalfJung, line 201:
```suggestion
    /// If the handle is invalid, or references a non-existent item, execution is aborted.
```

src/shims/windows/fs.rs
RalfJung: I find the code flow quite confusing here. There's a big `if` that has a bunch of side-effects mixed up; mostly it sets `option` but it also does some other stuff. `option` is then ignored in 2 out of 3 cases later but the other stuff is not.

Do you have ideas for how to better make the structure of the code reflect what happens in the various cases?

src/shims/windows/fs.rs
RalfJung: Okay, please add that quite in a comment to the code.

src/shims/windows/fs.rs
CraftSpider: It's not perfect, but I've tried moving some stuff around. Let me know if you think it's better or worse this way.

Cargo.toml
```diff
@@ -42,6 +43,7 @@ libloading = "0.8"
 [target.'cfg(target_family = "windows")'.dependencies]
 windows-sys = { version = "0.59", features = [
     "Win32_Foundation",
+    "Win32_Security",
```
RalfJung, line 46: Are you sure we need this? This is for Miri itself calling Windows API functions, which we only do in one place:

https://github.com/rust-lang/miri/blob/b63d8fcbda5a2617f2cb351f54668a55a14fa344/src/shims/unix/fs.rs#L211-L216

I don't know why "System_IO" is even in this list, we don't seem to use that.

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,402 @@
...
+        // to fail in the direction of erroring more often.
+        if let CreateAlways | OpenAlways = creation_disposition
+            && file_name.exists()
+        {
+            this.set_last_error(IoError::WindowsError("ERROR_ALREADY_EXISTS"))?;
+        }
+
+        let handle = if is_dir {
+            let fh = &mut this.machine.fds;
+            let fd_num = fh.insert_new(DirHandle { path: file_name });
```
RalfJung, line 255:
```suggestion
            // Open this as a directory.
            let fd_num = this.machine.fds.insert_new(DirHandle { path: file_name });
```

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,402 @@
...
+            Ok(Handle::File(fd_num))
+        } else if creation_disposition == OpenExisting && !(desired_read || desired_write) {
+            // Windows supports handles with no permissions. These allow things such as reading
+            // metadata, but not file content.
+            let fh = &mut this.machine.fds;
```
RalfJung, line 260: Please inline this `let`

src/shims/windows/fs.rs
```diff
@@ -0,0 +1,402 @@
...
+            file_name.metadata().map(|meta| {
+                let fd_num = fh.insert_new(MetadataHandle { meta });
+                Handle::File(fd_num)
+            })
+        } else {
+            match creation_disposition {
```
RalfJung, line 266:
```suggestion
            // Open this as a standard file.
            match creation_disposition {
```

tests/pass/shims/fs.rs
```diff
@@ -18,20 +17,22 @@ mod utils;
...
-    test_file_set_len();
-    test_file_sync();
-    test_errors();
-    test_rename();
-    test_directory();
-    test_canonicalize();
-    test_from_raw_os_error();
-    #[cfg(unix)]
-    test_pread_pwrite();
+    if cfg!(not(windows)) {
```
RalfJung, line 20:
```suggestion
    // Windows file handling is very incomplete.
    if cfg!(not(windows)) {
```

RalfJung (commented):
Thanks a lot! I'm happy with the general approach and structure now, we're just finalizing comments and style. :)

@rustbot author

src/shims/windows/fs.rs
RalfJung: Much better, thanks :)

CraftSpider:
Thanks for the patience and help with getting this over the line! After this merges, I'll work to flesh out the shims more, make it so you can actually do useful things with files targeting Windows.

@rustbot ready

RalfJung:
Thank you for your patience with the sometimes slow round-trip times. :)

@rustbot author

PR #4068, opened 2024-12-02
solarish stat following-up, supports for readdir.

devnexen:


tests/pass/shims/fs.rs
```diff
@@ -279,6 +276,12 @@ fn test_directory() {
             .collect::<BTreeMap<_, _>>()
     );
     // Deleting the directory should fail, since it is not empty.
+    #[cfg(any(target_os = "solaris", target_os = "illumos"))]
+    // Solaris/Illumos `rmdir` call set errno to EEXIST if directory contains
+    // other entries than `.` and `..`.
+    // https://docs.oracle.com/cd/E86824_01/html/E54765/rmdir-2.html
+    assert_eq!(ErrorKind::AlreadyExists, remove_dir(&dir_path).unwrap_err().kind());
+    #[cfg(not(any(target_os = "solaris", target_os = "illumos")))]
     assert_eq!(ErrorKind::DirectoryNotEmpty, remove_dir(&dir_path).unwrap_err().kind());
```
RalfJung, line 285: Please don't duplicate the entire `assert`. Instead, you can do something like
```rust
let expected_err = if cfg!(any(target_os = "solaris", target_os = "illumos")) { ErrorKind::AlreadyExists } else { ErrorKind::DirectoryNotEmpty };
assert_eq!(expected_err, remove_dir(&dir_path).unwrap_err().kind());
```

tests/pass/shims/fs.rs
```diff
@@ -279,6 +276,12 @@ fn test_directory() {
             .collect::<BTreeMap<_, _>>()
     );
     // Deleting the directory should fail, since it is not empty.
+    #[cfg(any(target_os = "solaris", target_os = "illumos"))]
+    // Solaris/Illumos `rmdir` call set errno to EEXIST if directory contains
+    // other entries than `.` and `..`.
+    // https://docs.oracle.com/cd/E86824_01/html/E54765/rmdir-2.html
+    assert_eq!(ErrorKind::AlreadyExists, remove_dir(&dir_path).unwrap_err().kind());
```
RalfJung, line 283: This assertion seems to fail on illumos now.

devnexen: Locally it passes tough, dunno if this is due to native vs non native run.

RalfJung: "Locally" meaning on an Illumos system? If you are passing through host errors, and host errors differ between OSes, then Miri will have to adjust the errors.

This likely also means that on your system, the tests will fail if you run them with `--target x86_64-unknown-linux-gnu`?

saethlin: > Does anyone really rely on the _specific_ errors being emitted?

The standard library does/has relied on specific error codes being returned by standard library functions, and has panicked or hit a debug assertion when an unexpected error code is produced.

I thought "pass through to the host" was something we explicitly avoided?

RalfJung: > I thought "pass through to the host" was something we explicitly avoided?

For IO functions it is unclear what else we could practically do. We already pass through the actual file system of the host (i.e., this anyway requires disabling isolation), also passing through the errors codes doesn't make a big difference any more at that point.

> The standard library does/has relied on specific error codes being returned by standard library functions, and has panicked or hit a debug assertion when an unexpected error code is produced.

Yes, but AFAIK this usually only checks that the error is some code allowed by POSIX -- whereas here we are apparently talking about POSIX allowing different codes, and different Unixes indeed using different codes.

saethlin: Oh dear, I see what's up here. Solarish is returning EEXIST for an attempt to remove a non-empty directory, because the POSIX description is "File Exists". But then std translates this error code to `ErrorKind::AlreadyExists`, adding the "already" which at least in this specific case, is definitely wrong. But when we display an io::Error, we punt to `strerror` which should only say "File exists" so we loop back to reasonable behavior.

So I'm okay with some non-portability here. Is not the first time the ErrorKind translation has been rough.

src/shims/unix/fs.rs
```diff
@@ -1086,8 +1088,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let name_bytes = name.as_encoded_bytes();
                 let name_len = u64::try_from(name_bytes.len()).unwrap();
 
-                let dirent64_layout = this.libc_ty_layout("dirent64");
-                let d_name_offset = dirent64_layout.fields.offset(4 /* d_name */).bytes();
+                let is_linux = matches!(&*this.tcx.sess.target.os, "linux");
+
+                let dirent64_layout = if is_linux {
+                    this.libc_ty_layout("dirent64")
```
RalfJung, line 1094: So Solaris has a readdir64 function that takes an argument of type dirent (not dirent64)? How does this make sense...?

In the libc crate I cannot find a `readdir64` for Solarish at all.

src/shims/unix/fs.rs
```diff
@@ -1086,8 +1088,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                } else {
+                    this.libc_ty_layout("dirent")
+                };
+                let d_name_offset = if is_linux {
+                    dirent64_layout.fields.offset(4 /* d_name */).bytes()
+                } else {
+                    dirent64_layout.fields.offset(3 /* d_name */).bytes()
+                };
```
RalfJung, line 1102: This has to always be the last field, right? Otherwise our allocation logic makes no sense. So please make it pick the last field here rather than hard-coding an index.

src/shims/unix/fs.rs
```diff
@@ -1086,8 +1088,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let name_bytes = name.as_encoded_bytes();
                 let name_len = u64::try_from(name_bytes.len()).unwrap();
```
RalfJung, line 1089: There's a comment above here, documenting the struct layout, that is now outdated.

src/shims/unix/fs.rs
```diff
@@ -1105,17 +1117,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                        ("d_ino", ino.into()),
-                        ("d_off", 0),
-                        ("d_reclen", size.into()),
-                        ("d_type", file_type.into()),
-                    ],
+                    &[("d_ino", ino.into()), ("d_off", 0), ("d_reclen", size.into())],
                     &this.ptr_to_mplace(entry, dirent64_layout),
                 )?;
 
+                if is_linux {
```
RalfJung, line 1125: I think ideally we'd say "if the d_type field exists, then write to it".

So I suggest you go to `helpers.rs` and find `fn project_field_named`. Then you add a new `fn try_project_field_named` that returns `InterpResult<'tcx, Option<P>>` so that when the field does not exist, it is `None`. And then you use that here.

tests/pass/shims/fs.rs
RalfJung: Okay, so let's p;ass through the host error, and let's make the test accept both `AlreadyExists` and `NotEmpty`.

src/shims/unix/fs.rs
devnexen: no there is no readdir64 on solaris but because of [this alias ](https://github.com/rust-lang/rust/blob/c94848c046d29f9a80c09aae758e27e418a289f2/library/std/src/sys/pal/unix/fs.rs#L33) I got readdir64 in the unsupported call error before.

RalfJung: > because of [this alias ](https://github.com/rust-lang/rust/blob/c94848c046d29f9a80c09aae758e27e418a289f2/library/std/src/sys/pal/unix/fs.rs#L33)I got readdir64 as symbol name in the unsupported call error before.

That can't be right, such an alias does not change the symbol name.

Please remove the readdir64 on Solarish then if it does not exist there.

src/helpers.rs
```diff
@@ -324,6 +324,21 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         bug!("No field named {} in type {}", name, base.layout().ty);
     }
 
+    fn try_project_field_named<P: Projectable<'tcx, Provenance>>(
```
RalfJung, line 327: Please don't duplicate this code. Instead, use this function to implement `project_field_named` and `projectable_has_field`.

src/shims/unix/fs.rs
```diff
@@ -1070,7 +1072,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             Some(Ok(dir_entry)) => {
                 // Write the directory entry into a newly allocated buffer.
                 // The name is written with write_bytes, while the rest of the
-                // dirent64 struct is written using write_int_fields.
+                // dirent64 (or dirent) struct is written using write_int_fields.
 
                 // For reference:
```
RalfJung, line 1077:
```suggestion
                // On Linux:
```

src/shims/unix/fs.rs
```diff
@@ -1080,14 +1082,29 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 //     pub d_type: c_uchar,
                 //     pub d_name: [c_char; 256],
                 // }
+                //
+                // pub struct dirent {
```
RalfJung, line 1086:
```suggestion
                // On Solarish:
                // pub struct dirent {
```

tests/pass/shims/fs.rs
```diff
@@ -279,7 +276,12 @@ fn test_directory() {
             .collect::<BTreeMap<_, _>>()
     );
     // Deleting the directory should fail, since it is not empty.
-    assert_eq!(ErrorKind::DirectoryNotEmpty, remove_dir(&dir_path).unwrap_err().kind());
+
+    // Solaris/Illumos `rmdir` call set errno to EEXIST if directory contains
+    // other entries than `.` and `..`.
+    // https://docs.oracle.com/cd/E86824_01/html/E54765/rmdir-2.html
+    let err = remove_dir(&dir_path).unwrap_err().kind();
+    assert!(err == ErrorKind::AlreadyExists || err == ErrorKind::DirectoryNotEmpty);
```
RalfJung, line 284:
```suggestion
    assert!(matches!(err, ErrorKind::AlreadyExists | ErrorKind::DirectoryNotEmpty));
```

src/shims/unix/fs.rs
```diff
@@ -1080,14 +1082,29 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let is_linux = matches!(&*this.tcx.sess.target.os, "linux");
+
+                let dirent64_layout = if is_linux {
+                    this.libc_ty_layout("dirent64")
+                } else {
+                    this.libc_ty_layout("dirent")
+                };
+                let fields = &dirent64_layout.fields;
+                let fcount = (*fields).count().checked_sub(1).expect("invalid count");
+                let d_name_offset = (*fields).offset(fcount).bytes();
```
RalfJung, line 1107:
```suggestion
                let last_field = dirent64_layout.fields.count().strict_sub(1);
                let d_name_offset = (*fields).offset(last_field).bytes();
```

devnexen:
I think I know why there is weird commit hours from me, if I let the VM/host hibernate for so many hours, the time does not catch up once I wake it up.

src/shims/unix/fs.rs
devnexen: Seems locally I no longer need to add readdir64. Glad that's the case.. Let's see CI then I ll rephrase a bit the original commit message.

src/shims/unix/fs.rs
```diff
@@ -1070,24 +1072,41 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                 let mut name = dir_entry.file_name(); // not a Path as there are no separators!
                 name.push("\0"); // Add a NUL terminator
                 let name_bytes = name.as_encoded_bytes();
                 let name_len = u64::try_from(name_bytes.len()).unwrap();
 
-                let dirent64_layout = this.libc_ty_layout("dirent64");
-                let d_name_offset = dirent64_layout.fields.offset(4 /* d_name */).bytes();
+                let is_linux = matches!(&*this.tcx.sess.target.os, "linux");
+
+                let dirent64_layout = if is_linux {
```
RalfJung, line 1102: My suggestion was to pass the type name in as a string. IMO that makes more sense since both dirent64 and dirent exist on Linux.

Why did you deviate from my suggestion? Please don't just silently do something different and ask for the next review. That is tiring as I have to painstakingly check for things like that. Instead, write a comment explaining why you deviated from the proposed plan.

src/shims/unix/fs.rs
```diff
@@ -1070,24 +1072,41 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    this.libc_ty_layout("dirent64")
+                } else {
+                    this.libc_ty_layout("dirent")
+                };
+                let fields = &dirent64_layout.fields;
+                let last_field = (*fields).count().strict_sub(1);
+                let d_name_offset = (*fields).offset(last_field).bytes();
```
RalfJung, line 1109:
```suggestion
                let last_field = fields.count().strict_sub(1);
                let d_name_offset = fields.offset(last_field).bytes();
```

src/shims/unix/fs.rs
devnexen: No I just did not get you (or forgot to address it). I ll look at it shortly.

devnexen:
@RalfJung before I squash do you like it better now ?

RalfJung:
My proposal was to add a `type_name: &str` argument to the `linux_solarish_readdir64` function.

RalfJung:
*oops* seems like I never actually posted that proposal here. Sorry for that! I was sure I submitted this, but apparently I replaced it by the question about readdir64 on Solarish.

devnexen:
> _oops_ seems like I never actually posted that proposal here. Sorry for that! I was sure I submitted this, but apparently I replaced it by the question about readdir64 on Solarish.

no worries, updating :)

RalfJung:
Btw, please avoid rebasing over the latest master branch when there are no conflicts. Github is terrible at dealing with force pushes during review, so the more you can avoid force pushes outside of squashing, the better. :)

RalfJung:
Apart from one final nit, this looks great, thanks! Please squash the commits, then we can land this. Please use the `--keep-base` flag when squashing so that the force-push diff is easier to review. Then write `@rustbot ready` after you pushed the squashed PR.

@rustbot author

Issue #4025, opened 2024-11-10
String de-duplication for ``localtime_r``


tiif:
Since this is already brought up in https://github.com/rust-lang/miri/issues/3470#issuecomment-2460600407, I will just open a new issue for this. 

To resolve the FIXME here, https://github.com/rust-lang/miri/blob/052bdcb72dda94f0ac0284304503d25de28acc6b/src/shims/time.rs#L198-L199

we could do: 

> We should have allocate_bytes which allocates a byte slice, and that should use the cache. allocate_str can call that, and transmute the result to &str.

shamb0:
Hi @tiif, @RalfJung, could you please assign this issue to me? I’d like to work on it.

saethlin:
Done.

RalfJung:
Thanks for the PR! I am replying here so that the issue has at least a rough sketch of some mentoring instructions.

However, adding a new string cache is the wrong approach here. We already have a deduplication machinery in Miri: `this.allocate_str`. We don't want a second system doing the same thing. We definitely don't want `static` storing `Pointer`, that will go completely wrong if someone runs more than one Miri interpreter instance in the same process.

So this change will have to be done in stages:
- first a rustc PR that adds a new `allocate_bytes`, very similar to `allocate_str` that exist [here](https://github.com/rust-lang/rust/blob/7a90e84f4d760cb49035f1446d4c6cfb3993aad2/compiler/rustc_const_eval/src/interpret/place.rs#L1023). `allocate_str` should then use `allocate_bytes`.
- Then a Miri PR that changes the timezone logic to use `allocate_bytes`. (We don't really need the `alloc_os_str_as_c_str` stuff there since we start with a `String` for the timezone, not an `OsString`.)

RalfJung:
No this has not been fixed yet...

RalfJung:
It's because you pushed a commit that contains "closes: ..." to a master branch, and you have the permission to close issues here.

saethlin:
This is a long-standing GifHub design flaw. The "fixes" magic commit messages are applied when you push to the default branch of a fork.

Most contributors "fix" this by not keeping their master branch up to date which is gross but better than randomly closing issues. I recently ran into the same problem and had to tweak my local workflow as a result. It's really unfortunate.

PR #4069, opened 2024-12-03
localtime_r: deduplicate timezone name allocation

shamb0:
Fixes https://github.com/rust-lang/miri/issues/4025

This PR improves timezone string handling in `localtime_r()` by utilizing the newly introduced `allocate_bytes()` function from rust-lang/rust#133861. This change provides a more efficient approach to string deduplication and memory management.

Changes implemented:
1. Integration of `allocate_bytes()` in `localtime_r()`:
   - Replaces existing string buffer allocation logic
   - Provides proper handling for null-terminated C strings
   - Ensures efficient memory deduplication for timezone strings

This is part 2 of the planned improvements for timezone handling:
- Part 1 (rust-lang/rust#133861): Added `allocate_bytes()` to InterpCx
- Part 2 (this PR): Implementing timezone string handling using `allocate_bytes()`

Related: https://github.com/rust-lang/rust/pull/133861

### DIFF
--- a/tests/pass-dep/libc/libc-time.rs
+++ b/tests/pass-dep/libc/libc-time.rs
@@ -5,7 +5,21 @@ use std::{env, mem, ptr};
 fn main() {
     test_clocks();
     test_posix_gettimeofday();
-    test_localtime_r();
+    test_localtime_r_gmt();
+    test_localtime_r_pst();
+    test_localtime_r_epoch();
+    #[cfg(any(
+        target_os = "linux",
+        target_os = "macos",
+        target_os = "freebsd",
+        target_os = "android"
+    ))]
+    test_localtime_r_multiple_calls_deduplication();
+    // Architecture-specific tests.
+    #[cfg(target_pointer_width = "32")]
+    test_localtime_r_future_32b();
+    #[cfg(target_pointer_width = "64")]
+    test_localtime_r_future_64b();
 }
 
 /// Tests whether clock support exists at all
@@ -46,14 +60,9 @@ fn test_posix_gettimeofday() {
     assert_eq!(is_error, -1);
 }
 
-fn test_localtime_r() {
-    // Set timezone to GMT.
-    let key = "TZ";
-    env::set_var(key, "GMT");
-
-    const TIME_SINCE_EPOCH: libc::time_t = 1712475836;
-    let custom_time_ptr = &TIME_SINCE_EPOCH;
-    let mut tm = libc::tm {
+/// Helper function to create an empty tm struct.
+fn create_empty_tm() -> libc::tm {
+    libc::tm {
         tm_sec: 0,
         tm_min: 0,
         tm_hour: 0,
@@ -77,7 +86,17 @@ fn test_localtime_r() {
             target_os = "android"
         ))]
         tm_zone: std::ptr::null_mut::<libc::c_char>(),
-    };
+    }
+}
+
+/// Original GMT test
+fn test_localtime_r_gmt() {
+    // Set timezone to GMT.
+    let key = "TZ";
+    env::set_var(key, "GMT");
+    const TIME_SINCE_EPOCH: libc::time_t = 1712475836; // 2024-04-07 07:43:56 GMT
+    let custom_time_ptr = &TIME_SINCE_EPOCH;
+    let mut tm = create_empty_tm();
     let res = unsafe { libc::localtime_r(custom_time_ptr, &mut tm) };
 
     assert_eq!(tm.tm_sec, 56);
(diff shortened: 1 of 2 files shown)


RalfJung:
Thanks for the PR! However, I am afraid this is the wrong approach. I have added a comment in https://github.com/rust-lang/miri/issues/4025 laying out the right way to do deduplication here.

> Resolves: https://github.com/rust-lang/miri/issues/4025

Please write comments like this in a way that github recognized so that issues get automatically closed. For instance: "Fixes https://github.com/rust-lang/miri/issues/4025".

shamb0:
> Thanks for the PR! However, I am afraid this is the wrong approach. I have added a comment in #4025 laying out the right way to do deduplication here.
> 
> > Resolves: #4025
> 
> Please write comments like this in a way that github recognizes so that issues get automatically closed. For instance: "Fixes #4025".

Thank you, @Ralfjung, for providing detailed implementation guidelines. 

I referred to your comment here: https://github.com/rust-lang/miri/issues/4025#issuecomment-2513760726.

In the current implementation, I assumed that all `shims` system calls are emulated in a single-thread context. Your observation is indeed correct—the implemented function is not thread-safe when multiple Miri interpreter instances are used in the same process.

I’ll begin working on the suggested guidelines, as they provide a solid starting point for refactoring the code.

RalfJung:
> In the current implementation, I assumed that all shims system calls are emulated in a single-thread context. Your observation is indeed correct—the implemented function is not thread-safe when multiple Miri interpreter instances are used in the same process.

Concurrency is not even needed to cause problems. One could run the interpreter to completion once, and then afterwards start a second interpreter and run that, and your cache would try to use old pointers from the first interpreter.

Generally, `static` should be avoided whenever possible. :)

shamb0:
Hi @RalfJung,

Completed implementing the changes requested in [#4025](https://github.com/rust-lang/miri/issues/4025#issuecomment-2513760726).

Current Status:

1. RUSTC PR (rust-lang/rust#133861):
   - Implemented `allocate_bytes()`
   - Refactored `allocate_str()` to use the new function

2. MIRI PR (this PR):
   - Implemented timezone logic using `allocate_bytes()`
   - Verified changes locally in `rust-lang/rust/src/tools/miri`

Question: Since this PR depends on the RUSTC changes and isn't currently buildable independently, would it make sense to:
1. Close this PR
2. Move the MIRI-specific changes to the RUSTC PR

Looking forward to your guidance on how to proceed.

RalfJung:
> Question: Since this PR depends on the RUSTC changes and isn't currently buildable independently, would it make sense to:

I'd say we land this in two stages -- rustc PR first, then this one. So don't worry about CI here for now. Just make sure that the rustc PR is ready and passes CI. :)

shamb0:
> > Question: Since this PR depends on the RUSTC changes and isn't currently buildable independently, would it make sense to:
> 
> I'd say we land this in two stages -- rustc PR first, then this one. So don't worry about CI here for now. Just make sure that the rustc PR is ready and passes CI. :)

Hi @RalfJung,

Thank you for the guidance on the two-stage implementation approach. I've prepared the RUSTC PR with the following implementation details:

Key Implementation Choices:
1. For `allocate_str()`:
   - Maintained existing metadata and layout configuration for consistency
   - Preserved current behavior while using the new `allocate_bytes()` function

2. For `allocate_bytes()`:
   - Using `self.tcx.types.u8` for layout
   - Set `MemPlaceMeta::None` as the type is sized

Please let me know if any of these choices need adjustment to better align with the requirements. I'm happy to make any necessary refinements.

Looking forward to your review of the RUSTC PR before proceeding with the MIRI changes.

RalfJung:
Let's discuss that in the rustc PR. :) I added it to my review queue.

(And btw, it's rustc, not RUSTC. And we usually write Miri, not MIRI. Not a big deal, just FYI :)

shamb0:
> Let's discuss that in the rustc PR. :) I added it to my review queue.
> 
> (And btw, it's rustc, not RUSTC. And we usually write Miri, not MIRI. Not a big deal, just FYI :)

Thankyou :)

shamb0:
> @shamb0 I have synced Miri with the latest rustc. So please rebase this PR over the latest Miri master branch, run `./miri toolchain` to sync your local toolchain, and then you should be able to use the new method you added. :)

Hi @RalfJung,  

Thank you for the guidance! The local tests look good, and I believe the CI build should pass now. Once confirmed, we can proceed with the review and intake.  

Let me know if any further enhancements are needed. :)

tests/pass-dep/libc/libc-time.rs
```diff
@@ -95,20 +107,172 @@ fn test_localtime_r() {
...
+
+    assert_eq!(tm.tm_sec, 56);
+    assert_eq!(tm.tm_min, 43);
+    assert_eq!(tm.tm_hour, 0); // 7 - 7 = 0 (PDT offset)
+    assert_eq!(tm.tm_mday, 7);
+    assert_eq!(tm.tm_mon, 3);
+    assert_eq!(tm.tm_year, 124);
+    assert_eq!(tm.tm_wday, 0);
+    assert_eq!(tm.tm_yday, 97);
+    assert_eq!(tm.tm_isdst, -1); // DST information unavailable
```
saethlin, line 142: If you want, `tm` has `PartialEq`, so you could compare all the fields with one assertion. Sometimes that produces hard-to-read output when the test fails though.

shamb0: Hi @saethlin,

Thank you for your comment!

The decision to use field-wise assertions was mainly driven by two factors:

* Minimizing deviation from the existing implementation.
* As you pointed out, field-wise assertions make it easier to identify the specific cause of a test failure, whereas a single assertion comparing all fields can sometimes result in hard-to-read error messages.

It's a tradeoff, but I’m open to revisiting it. Would you prefer switching to a single assertion to compare all fields using PartialEq, or do you think the current approach works better for readability and debugging?

Looking forward to your thoughts!

saethlin: Whichever you prefer is fine with me. Just making sure you'd thought about this, and clearly you have :)

tests/pass-dep/libc/libc-time.rs
```diff
@@ -95,20 +107,172 @@ fn test_localtime_r() {
...
+
+    // Remove timezone setting.
+    env::remove_var(key);
+}
+
+// PST timezone test (testing different timezone handling).
+fn test_localtime_r_pst() {
+    let key = "TZ";
+    env::set_var(key, "PST8PDT");
+    const TIME_SINCE_EPOCH: libc::time_t = 1712475836;
```
RalfJung, line 128:
```suggestion
    const TIME_SINCE_EPOCH: libc::time_t = 1712475836;// 2024-04-07 07:43:56 GMT
```

tests/pass-dep/libc/libc-time.rs
```diff
@@ -95,20 +107,172 @@ fn test_localtime_r() {
...
         target_os = "linux",
         target_os = "macos",
         target_os = "freebsd",
         target_os = "android"
     ))]
-    unsafe {
-        assert_eq!(std::ffi::CStr::from_ptr(tm.tm_zone).to_str().unwrap(), "+00")
-    };
+    {
+        assert_eq!(tm.tm_gmtoff, -25200); // -7 hours in seconds
```
RalfJung, line 151:
```suggestion
        assert_eq!(tm.tm_gmtoff, -7 * 3600); // -7 hours in seconds
```

RalfJung (commented):
Thanks! Some minor comments for the tests.

The new tests have nothing to do with the deduplication change, right? That confused me for a bit.^^

Would it be worth adding a test that checks the deduplication? That should be fairly easy, just call `localtime_r` again and check that the tm_zone pointer stays the same.

RalfJung:
> From the trace at fn allocate_bytes_dedup(), it seems like the salt value is varying:

Ah, right... so what the test would have to do is call `localtime_r` 50 times and then check that the number of different addresses we saw is at most 49 (and at least 2).

shamb0:
Hi @RalfJung,

I’ve implemented the test scenario `fn test_localtime_r_multiple_calls_deduplication()` based on your suggestion. Below are the captured traces and my observations:

    * From the traces, I noticed approximately 50% repetition in the salt values.
    * Salt values like 0, 12, 13, 22, and others appear multiple times.
    * The unique salt count indicates 24 unique values across a larger number of allocations.

```shell
--- tests/pass-dep/libc/libc-time.stdout
+++ <stdout output>
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 24, allocation id: a2863
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 31, allocation id: a3683
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 21, allocation id: a4120
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 10, allocation id: a4557
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 22, allocation id: a6400
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 19, allocation id: a6837
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 28, allocation id: a7274
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 6, allocation id: a7711
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 7, allocation id: a10898
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 1, allocation id: a11334
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 28, allocation id: a7274
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 6, allocation id: a7711
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 9, allocation id: a12517
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 31, allocation id: a3683
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 26, allocation id: a13328
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 2, allocation id: a13765
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 22, allocation id: a6400
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 13, allocation id: a14691
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 20, allocation id: a15127
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 22, allocation id: a6400
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 20, allocation id: a15127
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 9, allocation id: a12517
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 17, allocation id: a21807
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 21, allocation id: a4120
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 22, allocation id: a6400
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 3, allocation id: a22991
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 5, allocation id: a23428
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 13, allocation id: a14691
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 12, allocation id: a24238
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 31, allocation id: a3683
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 17, allocation id: a21807
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 18, allocation id: a25420
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 26, allocation id: a13328
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 15, allocation id: a26231
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 15, allocation id: a26231
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 0, allocation id: a27272
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 0, allocation id: a27272
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 24, allocation id: a2863
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 12, allocation id: a24238
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 3, allocation id: a22991
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 12, allocation id: a24238
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 5, allocation id: a23428
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 14, allocation id: a30178
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 22, allocation id: a6400
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 0, allocation id: a27272
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 23, allocation id: a31476
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 13, allocation id: a14691
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 13, allocation id: a14691
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 0, allocation id: a27272
+Allocating deduplicated bytes: [45, 48, 55, 0], salt: 26, allocation id: a13328
+Number of unique tm_zone pointers: 24
```

RalfJung (commented):
I don't understand the point of test_localtime_r_verify_string_deduplication... that's a very complicated test but what does it test?

test_localtime_r_multiple_calls_deduplication looks great! However, please make the entire test `cfg(any(...))`; there's no point in even running this test on targets that don't have the tm_zone field.

shamb0:
Hi @RalfJung,  

Thank you for the feedback!  

The function `test_localtime_r_verify_string_deduplication()` was an intermediate minimal test scenario meant to demonstrate unexpected behavior related to string deduplication. However, it has since been removed to streamline the test suite.  

Currently, all tests in `libc-time.rs` are verified using the `fn allocate_bytes()` function. If needed, I can update the tests to validate against the new, cleaned-up API `pub fn allocate_bytes_dedup()`. Please let me know if you'd like me to take this up!  

Thanks again for your suggestions.

tests/pass-dep/libc/libc-time.rs
```diff
@@ -95,20 +114,206 @@ fn test_localtime_r() {
...
+            let tm_ptr = libc::localtime_r(&timestamp, &mut tm);
+
+            assert!(!tm_ptr.is_null(), "localtime_r failed for timestamp {timestamp}");
+
+            let tm_zone_ptr = tm.tm_zone;
+            unique_pointers.insert(tm_zone_ptr);
+        }
+
+        let unique_count = unique_pointers.len();
+        println!("Number of unique tm_zone pointers: {}", unique_count);
```
RalfJung, line 310: Please remove this `println`, it will keep changing and CI will fail if the output does not match the reference file.

tests/pass-dep/libc/libc-time.rs
```diff
@@ -95,20 +114,206 @@ fn test_localtime_r() {
...
+    unsafe {
+        for i in 0..NUM_CALLS {
+            let timestamp = TIME_SINCE_EPOCH_BASE + (i as libc::time_t * 3600); // Increment by 1 hour for each call
+            let mut tm: libc::tm = create_empty_tm();
+            let tm_ptr = libc::localtime_r(&timestamp, &mut tm);
+
+            assert!(!tm_ptr.is_null(), "localtime_r failed for timestamp {timestamp}");
+
+            let tm_zone_ptr = tm.tm_zone;
+            unique_pointers.insert(tm_zone_ptr);
```
RalfJung, line 306:
```suggestion
            unique_pointers.insert(tm.tm_zone);
```
The `let` here doesn't make anything more clear IMO

RalfJung:
> Currently, all tests in libc-time.rs are verified using the fn allocate_bytes() function. If needed, I can update the tests to validate against the new, cleaned-up API pub fn allocate_bytes_dedup(). Please let me know if you'd like me to take this up!

`allocate_bytes_dedup` doesn't exist yet in the version of rustc that Miri uses. So just go ahead with `allocate_bytes` for now. If a rustup PR lands before this PR lands, you'll have to update, otherwise I will do the update as part of the rustup.

shamb0:
> > Currently, all tests in libc-time.rs are verified using the fn allocate_bytes() function. If needed, I can update the tests to validate against the new, cleaned-up API pub fn allocate_bytes_dedup(). Please let me know if you'd like me to take this up!
> 
> `allocate_bytes_dedup` doesn't exist yet in the version of rustc that Miri uses. So just go ahead with `allocate_bytes` for now. If a rustup PR lands before this PR lands, you'll have to update, otherwise I will do the update as part of the rustup.

Thankyou :), I hope PR is ready for intake review 👍🏾

RalfJung:
Remember to write this when the PR is ready :)
@rustbot ready

tests/pass-dep/libc/libc-time.rs
```diff
@@ -46,14 +60,9 @@ fn test_posix_gettimeofday() {
...
 
-fn test_localtime_r() {
-    // Set timezone to GMT.
-    let key = "TZ";
-    env::set_var(key, "GMT");
-
-    const TIME_SINCE_EPOCH: libc::time_t = 1712475836;
-    let custom_time_ptr = &TIME_SINCE_EPOCH;
-    let mut tm = libc::tm {
+// Helper function to create an empty tm struct.
```
RalfJung, line 63:
```suggestion
/// Helper function to create an empty tm struct.
```
`///` indicates a doc comment that documents the following function.

Please also do that with the other similar comments in this file.

tests/pass-dep/libc/libc-time.rs
```diff
@@ -95,20 +114,204 @@ fn test_localtime_r() {
...
+fn test_localtime_r_multiple_calls_deduplication() {
+    let key = "TZ";
+    env::set_var(key, "PST8PDT");
+
+    const TIME_SINCE_EPOCH_BASE: libc::time_t = 1712475836; // Base timestamp: 2024-04-07 07:43:56 GMT
+    const NUM_CALLS: usize = 50;
+
+    let mut unique_pointers = std::collections::HashSet::new();
+
+    unsafe {
```
RalfJung, line 297: Please make this unsafe block saller. Only the call to `localtime_r` needs to be inside of it, I think?

shamb0: Good One, Thankyou :)

RalfJung:
@rustbot author

RalfJung:
No this is not ready yet, some of my comments from the previous round have not been resolved.
@rustbot author

tests/pass-dep/libc/libc-time.rs
RalfJung: Note that I wrote this:

**Please also do that with the other similar comments in this file.**

tests/pass-dep/libc/libc-time.rs
RalfJung: This one is not resolved yet, either.

shamb0: - I modified the code, but I'm wondering why it hasn't been committed :)
```
    for i in 0..NUM_CALLS {
        let timestamp = TIME_SINCE_EPOCH_BASE + (i as libc::time_t * 3600); // Increment by 1 hour for each call
        let mut tm: libc::tm = create_empty_tm();
        let tm_ptr = unsafe { libc::localtime_r(&timestamp, &mut tm) };

        assert!(!tm_ptr.is_null(), "localtime_r failed for timestamp {timestamp}");

        unique_pointers.insert(tm.tm_zone);
    }
```

RalfJung: I'm afraid I can't help you with that -- you have to commit and push your changes as usual.

RalfJung:
Looks great, thanks. :)

PR #4072, opened 2024-12-04
Implement blocking unnamed_socket

### LINKED ISSUE #3665: Implement blocking support for eventfd and socketpair

tiif:
Support blocking for socketpair and pipe. 

closes #3665


tiif:
Planning to open a minor clean-up PR to remove some unused variable here before getting back to this. 

@rustbot label S-blocked

src/shims/unix/unnamed_socket.rs
```diff
@@ -114,14 +114,32 @@ impl FileDescription for AnonSocket {
...
+                                // TODO: We might need to decide what to do if peer_fd is closed when read is blocked.
+                                anonsocket_read(weak_self_ref, peer_fd, len, ptr, &dest, this)
+                            }
+                        ),
+                    );
                 }
             }
         }
-        // TODO: We might need to decide what to do if peer_fd is closed when read is blocked.
-        anonsocket_read(self, self.peer_fd().upgrade(), len, ptr, dest, ecx)
```
oli-obk, line 124: You need to keep doing this here for in the case where we don't need to block

src/shims/unix/unnamed_socket.rs
```diff
@@ -207,15 +225,16 @@ fn anonsocket_write<'tcx>(
 
 /// Read from AnonSocket and return the number of bytes read.
 fn anonsocket_read<'tcx>(
-    anonsocket: &AnonSocket,
-    peer_fd: Option<FileDescriptionRef>,
+    self_ref: WeakFileDescriptionRef,
+    peer_fd: WeakFileDescriptionRef,
```
oli-obk, line 253: `peer_fd` doesn't need to be passed, you can always get it from the AnonSocket (preexisting issue)

src/shims/unix/unnamed_socket.rs
tiif: Yea this is caught by the testcases too, it is now fixed :D

src/shims/unix/unnamed_socket.rs
```diff
@@ -171,82 +167,143 @@ impl FileDescription for AnonSocket {
 }
 
 /// Write to AnonSocket based on the space available and return the written byte size.
+/// This is only for socketpair/pipe without O_NONBLOCK flag.
```
tiif, line 170: Just something on top of my mind while working on this, if we were to support ``O_NONBLOCK`` flag in ``fcntl``, which has the ability of setting a file descriptor with blocking ability to non-blocking, we need to be careful of it's interaction with all the blocking operation here.

src/shims/unix/unnamed_socket.rs
```diff
@@ -100,33 +104,31 @@ impl FileDescription for AnonSocket {
             // corresponding ErrorKind variant.
             throw_unsup_format!("reading from the write end of a pipe");
         };
+        let dest = dest.clone();
+        let weak_self_ref = self_ref.downgrade();
+
+        // We handle return earlier if we hit these two case, and only do the actual read and
+        // blocking in anonsocket_read.
```
oli-obk, line 111: I don't understand this comment. What do you mean by "handle return"?

tiif: Ah it should be ``"We return early..."``

But please ignore the comment for now, I plan to remove that because one of the cases should be moved to the helper function.

src/shims/unix/unnamed_socket.rs
```diff
@@ -100,33 +104,23 @@ impl FileDescription for AnonSocket {
...
+
+        if readbuf.borrow().buf.is_empty() && self.is_nonblock {
+            // Non-blocking socketpair with writer and empty buffer.
+            // https://linux.die.net/man/2/read
+            // EAGAIN or EWOULDBLOCK can be returned for socket,
+            // POSIX.1-2001 allows either error to be returned for this case.
+            // Since there is no ErrorKind for EAGAIN, WouldBlock is used.
+            return ecx.set_last_error_and_return(ErrorKind::WouldBlock, &dest);
         }
-        // TODO: We might need to decide what to do if peer_fd is closed when read is blocked.
```
tiif, line 123: It turns out that we can just follow what we did previously:

for ``read``, just keep reading if there are data inside ``readbuf``. 

for ``write``, return ``ecx.set_last_error_and_return(ErrorKind::BrokenPipe, &dest);``.

src/shims/unix/unnamed_socket.rs
```diff
@@ -36,6 +36,10 @@ struct AnonSocket {
     /// This flag is set to `true` if the peer's `readbuf` is non-empty at the time
     /// of closure.
     peer_lost_data: Cell<bool>,
+    /// A list of thread ids blocked on read.
+    blocked_read_tid: RefCell<Vec<ThreadId>>,
+    /// A list of thread ids blocked on write.
```
oli-obk, line 41:
```suggestion
    /// A list of thread ids blocked because the buffer was full.
    /// Once another thread reads some bytes, these threads will be unblocked.
```

src/shims/unix/unnamed_socket.rs
```diff
@@ -36,6 +36,10 @@ struct AnonSocket {
     /// This flag is set to `true` if the peer's `readbuf` is non-empty at the time
     /// of closure.
     peer_lost_data: Cell<bool>,
+    /// A list of thread ids blocked on read.
```
oli-obk, line 39:
```suggestion
    /// A list of thread ids blocked because the buffer was empty.
    /// Once another thread writes some bytes, these threads will be unblocked.
```

src/shims/unix/unnamed_socket.rs
```diff
@@ -153,16 +147,13 @@ impl FileDescription for AnonSocket {
...
-        if available_space == 0 {
-            if self.is_nonblock {
-                // Non-blocking socketpair with a full buffer.
-                return ecx.set_last_error_and_return(ErrorKind::WouldBlock, dest);
-            } else {
-                // Blocking socketpair with a full buffer.
-                throw_unsup_format!("socketpair/pipe/pipe2 write: blocking isn't supported yet");
-            }
+        let weak_self_ref = self_ref.downgrade();
+        let dest = dest.clone();
```
oli-obk, line 151: these can be done after the if condition.

src/shims/unix/unnamed_socket.rs
```diff
@@ -100,33 +104,23 @@ impl FileDescription for AnonSocket {
...
-                    // Since there is no ErrorKind for EAGAIN, WouldBlock is used.
-                    return ecx.set_last_error_and_return(ErrorKind::WouldBlock, dest);
-                } else {
-                    // Blocking socketpair with writer and empty buffer.
-                    // FIXME: blocking is currently not supported
-                    throw_unsup_format!("socketpair/pipe/pipe2 read: blocking isn't supported yet");
-                }
-            }
+        let dest = dest.clone();
+        let weak_self_ref = self_ref.downgrade();
```
oli-obk, line 108: these can be done after the if condition

src/shims/unix/unnamed_socket.rs
```diff
@@ -100,33 +104,23 @@ impl FileDescription for AnonSocket {
...
+
+        if readbuf.borrow().buf.is_empty() && self.is_nonblock {
+            // Non-blocking socketpair with writer and empty buffer.
+            // https://linux.die.net/man/2/read
+            // EAGAIN or EWOULDBLOCK can be returned for socket,
+            // POSIX.1-2001 allows either error to be returned for this case.
+            // Since there is no ErrorKind for EAGAIN, WouldBlock is used.
+            return ecx.set_last_error_and_return(ErrorKind::WouldBlock, &dest);
```
oli-obk, line 116: And subsequently change this back

```suggestion
            return ecx.set_last_error_and_return(ErrorKind::WouldBlock, dest);
```

src/shims/unix/unnamed_socket.rs
```diff
@@ -153,16 +147,13 @@ impl FileDescription for AnonSocket {
...
-                return ecx.set_last_error_and_return(ErrorKind::WouldBlock, dest);
-            } else {
-                // Blocking socketpair with a full buffer.
-                throw_unsup_format!("socketpair/pipe/pipe2 write: blocking isn't supported yet");
-            }
+        let weak_self_ref = self_ref.downgrade();
+        let dest = dest.clone();
+        if available_space == 0 && self.is_nonblock {
+            // Non-blocking socketpair with a full buffer.
+            return ecx.set_last_error_and_return(ErrorKind::WouldBlock, &dest);
```
oli-obk, line 154: And subsequently switch this back

```suggestion
            return ecx.set_last_error_and_return(ErrorKind::WouldBlock, dest);
```

oli-obk (changes requested):
just some nits, then this is ready, please apply and squash

oli-obk (approved):

src/shims/unix/unnamed_socket.rs
```diff
@@ -172,81 +161,161 @@ impl FileDescription for AnonSocket {
...
-    }
 
-    ecx.return_read_success(ptr, &bytes, actual_read_size, dest)
+    if readbuf.borrow_mut().buf.is_empty() {
+        if self_anonsocket.peer_fd().upgrade().is_none() {
+            // Socketpair with no peer and empty buffer.
+            // 0 bytes successfully read indicates end-of-file.
+            return ecx.return_read_success(ptr, &[], 0, &dest);
+        } else {
+            // Blocking socketpair with writer and empty buffer.
```
RalfJung, line 263: How do we know it is blocking?

If that's a precondition, please add an `assert!` further up this function.

src/shims/unix/unnamed_socket.rs
```diff
@@ -172,81 +161,161 @@ impl FileDescription for AnonSocket {
...
+            let weak_self_ref = weak_self_ref.clone();
```
RalfJung, line 264: Why do you have to clone this reference? Seems like you should move it.

src/shims/unix/unnamed_socket.rs
```diff
@@ -100,33 +106,21 @@ impl FileDescription for AnonSocket {
...
-                    // Since there is no ErrorKind for EAGAIN, WouldBlock is used.
-                    return ecx.set_last_error_and_return(ErrorKind::WouldBlock, dest);
-                } else {
-                    // Blocking socketpair with writer and empty buffer.
-                    // FIXME: blocking is currently not supported
-                    throw_unsup_format!("socketpair/pipe/pipe2 read: blocking isn't supported yet");
-                }
-            }
+
+        if readbuf.borrow().buf.is_empty() && self.is_nonblock {
```
RalfJung, line 110: The separation of concerns between read and anonsocket_read is a bit unclear. Why is this one case handled here and everything else handled below?

I think the code would be easier to follow if you moved this logic into anonsocket_read as well.

src/shims/unix/unnamed_socket.rs
```diff
@@ -172,81 +161,161 @@ impl FileDescription for AnonSocket {
...
 ) -> InterpResult<'tcx> {
+    let Some(self_ref) = weak_self_ref.upgrade() else {
+        // FIXME:  We should raise a deadlock error if the self_ref upgrade failed.
+        throw_unsup_format!("This will be a deadlock error in future")
+    };
+    let self_anonsocket = self_ref.downcast::<AnonSocket>().unwrap();
+    let Some(peer_fd) = self_anonsocket.peer_fd().upgrade() else {
+        // If the upgrade from Weak to Rc fails, it indicates that all read ends have been
+        // closed.
+        return ecx.set_last_error_and_return(ErrorKind::BrokenPipe, &dest);
```
RalfJung, line 178: Is it really necessary to duplicate this check (it is both here and in `write` above)? I think that should be avoidable, so please reduce the code duplication.

src/shims/unix/unnamed_socket.rs
```diff
@@ -153,16 +147,11 @@ impl FileDescription for AnonSocket {
...
             MAX_SOCKETPAIR_BUFFER_CAPACITY.strict_sub(writebuf.borrow().buf.len());
-        if available_space == 0 {
-            if self.is_nonblock {
-                // Non-blocking socketpair with a full buffer.
-                return ecx.set_last_error_and_return(ErrorKind::WouldBlock, dest);
-            } else {
-                // Blocking socketpair with a full buffer.
-                throw_unsup_format!("socketpair/pipe/pipe2 write: blocking isn't supported yet");
-            }
+        if available_space == 0 && self.is_nonblock {
```
RalfJung, line 150: This should also be moved into anonsocket_write I think, again to keep all the logic in the same place.

src/shims/unix/unnamed_socket.rs
```diff
@@ -153,16 +147,11 @@ impl FileDescription for AnonSocket {
         };
         let available_space =
             MAX_SOCKETPAIR_BUFFER_CAPACITY.strict_sub(writebuf.borrow().buf.len());
```
RalfJung, line 149: Please do not duplicate this logic.

src/shims/unix/unnamed_socket.rs
```diff
@@ -172,81 +161,161 @@ impl FileDescription for AnonSocket {
...
+        drop(writebuf);
 
-    // Notification should be provided for peer fd as it became readable.
-    // The kernel does this even if the fd was already readable before, so we follow suit.
-    ecx.check_and_update_readiness(peer_fd)?;
+        // Notification should be provided for peer fd as it became readable.
+        // The kernel does this even if the fd was already readable before, so we follow suit.
+        ecx.check_and_update_readiness(&peer_fd)?;
+        let peer_anonsocket = peer_fd.downcast::<AnonSocket>().unwrap();
+        // Unblock all threads that are currently blocked on peer_fd's read.
```
RalfJung, line 225: The `let peer_anonsocket` is already part of the "unblock all threads", right? So the comment should be moved up by 1 line.

src/shims/unix/unnamed_socket.rs
```diff
@@ -172,81 +161,161 @@ impl FileDescription for AnonSocket {
...
+        // only write 1 byte. This implementation is not compliant with the actual Linux kernel
+        // implementation. For optimization reasons, the kernel will only mark the file description
+        // as "writable" when it can write more than a certain number of bytes. Since we
+        // don't know what that *certain number* is, we will provide a notification every time
+        // a read is successful. This might result in our epoll emulation providing more
+        // notifications than the real system.
+        if let Some(peer_fd) = self_anonsocket.peer_fd().upgrade() {
+            ecx.check_and_update_readiness(&peer_fd)?;
+            let peer_anonsocket = peer_fd.downcast::<AnonSocket>().unwrap();
+            // Unblock all threads that are currently blocked on peer_fd's write.
```
RalfJung, line 307: The `let peer_anonsocket` is already part of the "unblock all threads", right? So the comment should be moved up by 1 line.

src/shims/unix/unnamed_socket.rs
tiif: My idea is keeping the ``non-blocking && error`` logic in ``read`` and ``write``. In ``anonsocket_read`` and ``anonsocket_write``, we either go through the ``block_thread`` path or the actual ``read/write`` path. 

For now, ``if readbuf.borrow().buf.is_empty() && self.is_nonblock`` check will never be hit by blocking fd (the one without ``SOCK_NONBLOCK flag``) after unblocked, so I just kept them in ``read/write``. 

It is fine totally fine to move it in ``anonsocket_read`` too if it is clearer.

src/shims/unix/unnamed_socket.rs
tiif: Yea we can always do the check in ``anonsocket_write``, and remove the one in ``write``. This is still in ``write`` because I wanted to throw error as early as possible.

src/shims/unix/unnamed_socket.rs
tiif: Oh yea, I should add a check here. The general idea is in https://github.com/rust-lang/miri/pull/4072#discussion_r1894133420, since ``readbuf.borrow().buf.is_empty() && self.is_nonblock `` is already handled in ``read``, this is always blocking. It is also impossible to change an fd from blocking type to non-blocking type for now.  But it is likely that I will move the checks around, I will add it if it is still necessary.

src/shims/unix/unnamed_socket.rs
RalfJung: Splitting the logic across two functions leads to a very odd precondition for `anonsocket_read`. It also means I have to scroll up and down to look at the entire logic. The logic really isn't that complicated so IMO it'd be better to have it in a single function where one can easily see the entire thing at a single glance.

tiif: This makes sense, I can combine them together.

RalfJung: Actually the new logic here is just wrong. When the buffer is empty, we have to *first* check `self_anonsocket.peer_fd().upgrade()` and then check ``is_nonblock``.

In the future please just keep the logic the way it is rather than trying to break it apart into two functions. :)

PR #4080, opened 2024-12-07
add weak memory consistency test for mixing SC accesses and fences

RalfJung:
@cbeuw I could not find any existing test covering this. Did I miss something?


cbeuw (approved):
I only added test cases that I found in papers and didn't come up with my own ones. So yeah there weren't any tests with both SC fences and accesses

PR #4093, opened 2024-12-13
Allow backtraces to cross thread boundaries

### LINKED ISSUE #4066: Backtraces stop at thread boundaries
When an action is performed that miri cannot handle (or UB is hit), the backtrace shown ends at the start of the thread in which the failure happened. As an example

```rust
extern {
    fn unknown();
}

fn bad() { unsafe { unknown() } }

fn main() {
    std::thread::spawn(bad);
}
```

fails with

```
error: unsupported operation: can't call foreign function `unknown` on OS `linux`
 --> src/main.rs:5:21
  |
5 | fn bad() { unsafe { unknown() } }
  |                     ^^^^^^^^^ can't call foreign function `unknown` on OS `linux`
  |
  = help: if this is a basic API commonly used on this target, please report an issue with Miri
  = help: however, note that Miri does not aim to support every FFI function out there; for instance, we will not support APIs for things such as GUIs, scripting languages, or databases
  = note: BACKTRACE on thread `unnamed-1`:
  = note: inside `bad` at src/main.rs:5:21: 5:30
```

pointing solely to the function `bad`.

In this simple example that is not an issue, but if a crate spawns a thread somewhere in its internals, it may not be clear why the thread is even running. I don't think it would be expensive to collect a backtrace at every thread spawn and store this information in the thread data structure itself (as we're not going through the backtrace crate, but just iterating over miri's stack).

actuallyatoaster:
Fixes #4066

When a thread is spawned we store the ID of the spawning thread as well as a backtrace to the spawn location. When there's an error we can traverse upwards, displaying a backtrace up to the main thread.

For instance, with the following program:
```
extern "C" {
    fn unknown();
}

fn bad() {
    unsafe { unknown() }
}

fn main() {
    std::thread::spawn(bad);
}
```

Whereas previously we would see:
```
error: unsupported operation: can't call foreign function `unknown` on OS `linux`
 --> tests/temp.rs:6:14
  |
6 |     unsafe { unknown() }
  |              ^^^^^^^^^ can't call foreign function `unknown` on OS `linux`
  |
  = help: if this is a basic API commonly used on this target, please report an issue with Miri
  = help: however, note that Miri does not aim to support every FFI function out there; for instance, we will not support APIs for things such as GUIs, scripting languages, or databases
  = note: BACKTRACE on thread `unnamed-1`:
  = note: inside `bad` at tests/temp.rs:6:14: 6:23

```

We now have:
```
error: unsupported operation: can't call foreign function `unknown` on OS `linux`
  --> tests/temp.rs:6:14
   |
6  |     unsafe { unknown() }
   |              ^^^^^^^^^ can't call foreign function `unknown` on OS `linux`
   |
   = help: if this is a basic API commonly used on this target, please report an issue with Miri
   = help: however, note that Miri does not aim to support every FFI function out there; for instance, we will not support APIs for things such as GUIs, scripting languages, or databases
   = note: BACKTRACE on thread `unnamed-1`:
   = note: inside `bad` at tests/temp.rs:6:14: 6:23
   = note: thread `unnamed-1` was spawned by thread `main`
   = note: inside `std::sys::pal::unix::thread::Thread::new` at /home/reimu/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/sys/pal/unix/thread.rs:84:19: 84:86
   = note: inside `std::thread::Builder::spawn_unchecked_::<'_, fn() {bad}, ()>` at /home/reimu/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/thread/mod.rs:600:30: 600:64
   = note: inside `std::thread::Builder::spawn_unchecked::<fn() {bad}, ()>` at /home/reimu/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/thread/mod.rs:467:32: 467:62
   = note: inside `std::thread::Builder::spawn::<fn() {bad}, ()>` at /home/reimu/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/thread/mod.rs:400:18: 400:41
   = note: inside `std::thread::spawn::<fn() {bad}, ()>` at /home/reimu/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/thread/mod.rs:730:5: 730:28
note: inside `main`
  --> tests/temp.rs:10:5
   |
10 |     std::thread::spawn(bad);
   |     ^^^^^^^^^^^^^^^^^^^^^^^

```

### DIFF
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -235,6 +235,9 @@ pub struct Thread<'tcx> {
     /// The join status.
     join_status: ThreadJoinStatus,
 
+    // ThreadId that spawned this thread and backtrace to where this thread was spawned
+    thread_spawn_context: Option<(ThreadId, Vec<FrameInfo<'tcx>>)>,
+
     /// Stack of active panic payloads for the current thread. Used for storing
     /// the argument of the call to `miri_start_unwind` (the panic payload) when unwinding.
     /// This is pointer-sized, and matches the `Payload` type in `src/libpanic_unwind/miri.rs`.
@@ -321,13 +324,18 @@ impl<'tcx> std::fmt::Debug for Thread<'tcx> {
 }
 
 impl<'tcx> Thread<'tcx> {
-    fn new(name: Option<&str>, on_stack_empty: Option<StackEmptyCallback<'tcx>>) -> Self {
+    fn new(
+        name: Option<&str>,
+        on_stack_empty: Option<StackEmptyCallback<'tcx>>,
+        thread_spawn_context: Option<(ThreadId, Vec<FrameInfo<'tcx>>)>,
+    ) -> Self {
         Self {
             state: ThreadState::Enabled,
             thread_name: name.map(|name| Vec::from(name.as_bytes())),
             stack: Vec::new(),
             top_user_relevant_frame: None,
             join_status: ThreadJoinStatus::Joinable,
+            thread_spawn_context,
             panic_payloads: Vec::new(),
             last_error: None,
             on_stack_empty,
@@ -345,6 +353,7 @@ impl VisitProvenance for Thread<'_> {
             state: _,
             thread_name: _,
             join_status: _,
+            thread_spawn_context: _,
             on_stack_empty: _, // we assume the closure captures no GC-relevant state
         } = self;
 
@@ -475,7 +484,7 @@ impl<'tcx> Default for ThreadManager<'tcx> {
     fn default() -> Self {
         let mut threads = IndexVec::new();
         // Create the main thread and add it to the list of threads.
-        threads.push(Thread::new(Some("main"), None));
+        threads.push(Thread::new(Some("main"), None, None));
         Self {
             active_thread: ThreadId::MAIN_THREAD,
             threads,
@@ -542,9 +551,13 @@ impl<'tcx> ThreadManager<'tcx> {
     }
 
     /// Create a new thread and returns its id.
-    fn create_thread(&mut self, on_stack_empty: StackEmptyCallback<'tcx>) -> ThreadId {
+    fn create_thread(
+        &mut self,
+        on_stack_empty: StackEmptyCallback<'tcx>,
+        thread_spawn_context: Option<(ThreadId, Vec<FrameInfo<'tcx>>)>,
+    ) -> ThreadId {
         let new_thread_id = ThreadId::new(self.threads.len());
-        self.threads.push(Thread::new(None, Some(on_stack_empty)));
+        self.threads.push(Thread::new(None, Some(on_stack_empty), thread_spawn_context));
         new_thread_id
     }
 
@@ -709,6 +722,13 @@ impl<'tcx> ThreadManager<'tcx> {
 
     pub fn get_thread_display_name(&self, thread: ThreadId) -> String {
         self.threads[thread].thread_display_name(thread)
+    }
+
+    pub fn get_thread_spawn_context(
+        &self,
+        thread: ThreadId,
+    ) -> Option<&(ThreadId, Vec<FrameInfo<'tcx>>)> {
+        self.threads[thread].thread_spawn_context.as_ref()
     }
 
     /// Put the thread into the blocked state.
@@ -924,10 +944,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let this = self.eval_context_mut();
 
         // Create the new thread
-        let new_thread_id = this.machine.threads.create_thread({
-            let mut state = tls::TlsDtorsState::default();
-            Box::new(move |m| state.on_stack_empty(m))
-        });
+        let new_thread_id = this.machine.threads.create_thread(
+            {
+                let mut state = tls::TlsDtorsState::default();
+                Box::new(move |m| state.on_stack_empty(m))
+            },
+            Some((this.machine.threads.active_thread(), this.generate_stacktrace())),
+        );
         let current_span = this.machine.current_span();
         if let Some(data_race) = &mut this.machine.data_race {
             data_race.thread_created(&this.machine.threads, new_thread_id, current_span);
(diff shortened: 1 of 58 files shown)


src/diagnostics.rs
```diff
@@ -588,6 +588,36 @@ pub fn report_msg<'tcx>(
...
 
+    // Traverse thread spawns to display thread backtraces up to main thread
+    if let Some(mut thread) = thread {
+        let mut thread_spawn_context = machine.threads.get_thread_spawn_context(thread);
+
+        while let Some((spawning_thread, thread_spawn_backtrace)) = thread_spawn_context {
+            err.note(format!(
+                "thread `{}` was spawned by thread `{}`",
+                machine.threads.get_thread_display_name(thread),
+                machine.threads.get_thread_display_name(*spawning_thread)
```
oli-obk, line 599: Does this work if the thread has already finished?

oli-obk:
You'll need to bless the windows tests, too. It should suffice to pass `--target` with a windows target, no windows host needed

src/diagnostics.rs
actuallyatoaster: From the logic in join_thread and terminate_active_thread
https://github.com/rust-lang/miri/blob/21d531b37b51d80a997ef03be1fc09dfe40556b2/src/concurrency/thread.rs#L642
https://github.com/rust-lang/miri/blob/21d531b37b51d80a997ef03be1fc09dfe40556b2/src/concurrency/thread.rs#L985

I was under the impression that the thread struct itself never actually gets deallocated so that future joins can be checked to ensure they aren't repeat joins on the given thread, but I could be mistaken. Would it be better to be safe and save the whole backtrace with thread names on thread creation?

src/concurrency/thread.rs
```diff
@@ -235,6 +235,9 @@ pub struct Thread<'tcx> {
     /// The join status.
     join_status: ThreadJoinStatus,
 
+    // ThreadId that spawned this thread and backtrace to where this thread was spawned
```
RalfJung, line 238:
```suggestion
    /// ThreadId that spawned this thread and backtrace to where this thread was spawned.
```

RalfJung:
I have some nits.

src/diagnostics.rs
```diff
@@ -588,6 +588,36 @@ pub fn report_msg<'tcx>(
...
+            ));
+            let (pruned_stacktrace, _was_pruned) =
+                prune_stacktrace(thread_spawn_backtrace.to_vec(), machine);
+
+            for (idx, frame_info) in pruned_stacktrace.iter().enumerate() {
+                let is_local = machine.is_local(frame_info);
+                // No span for non-local frames except the first frame (which is the error site).
+                if is_local && idx > 0 {
+                    err.subdiagnostic(frame_info.as_note(machine.tcx));
```
RalfJung, line 608: This comment is not right, the first frame is not the error site.

src/diagnostics.rs
```diff
@@ -588,6 +588,36 @@ pub fn report_msg<'tcx>(
...
+                } else {
+                    let sm = sess.source_map();
+                    let span = sm.span_to_embeddable_string(frame_info.span);
+                    err.note(format!("{frame_info} at {span}"));
+                }
+            }
```
RalfJung, line 614: This is duplicating existing logic. Can you move this into a shared helper fn?

tests/fail-dep/concurrency/libc_pthread_mutex_deadlock.stderr
```diff
@@ -6,6 +6,19 @@ LL |             assert_eq!(libc::pthread_mutex_lock(lock_copy.0.get() as *mut _
    |
    = note: BACKTRACE on thread `unnamed-ID`:
    = note: inside closure at tests/fail-dep/concurrency/libc_pthread_mutex_deadlock.rs:LL:CC
+   = note: thread `unnamed-ID` was spawned by thread `main`
+   = note: inside `std::sys::pal::PLATFORM::thread::Thread::new` at RUSTLIB/std/src/sys/pal/PLATFORM/thread.rs:LL:CC
+   = note: inside `std::thread::Builder::spawn_unchecked_::<'_, {closure@tests/fail-dep/concurrency/libc_pthread_mutex_deadlock.rs:LL:CC}, ()>` at RUSTLIB/std/src/thread/mod.rs:LL:CC
+   = note: inside `std::thread::Builder::spawn_unchecked::<{closure@tests/fail-dep/concurrency/libc_pthread_mutex_deadlock.rs:LL:CC}, ()>` at RUSTLIB/std/src/thread/mod.rs:LL:CC
+   = note: inside `std::thread::Builder::spawn::<{closure@tests/fail-dep/concurrency/libc_pthread_mutex_deadlock.rs:LL:CC}, ()>` at RUSTLIB/std/src/thread/mod.rs:LL:CC
+   = note: inside `std::thread::spawn::<{closure@tests/fail-dep/concurrency/libc_pthread_mutex_deadlock.rs:LL:CC}, ()>` at RUSTLIB/std/src/thread/mod.rs:LL:CC
```
RalfJung, line 14: It's not pretty to show this for every backtrace. We spent a bunch of effort making backtraces nice, i.e. they usually stop at `main` and don't go all the way to the start lang item. We should do something similar here to avoid drowning the user in noise.

I could imagine two options:
- we mark these frames as `#[caller_location]` in std (we have done this for a bunch of other functions to make backtraces prettier)
- when pruning the stack for another thread, we also remove all non-local frames at the *top*

I am leaning towards the first option since the second option might prune too much when the thread is spawned deep inside some other library.

RalfJung:
@rustbot author

tests/fail-dep/concurrency/libc_pthread_mutex_deadlock.stderr
RalfJung: I have posted https://github.com/rust-lang/rust/pull/134560 to add the `#[caller_location]` for this.

RalfJung:

PR #4099, opened 2024-12-20
Follow-up changes for blocking unamed_socket

tiif:
Since I left some comments on why the initial ``throw_unsupported`` is unreachable, I will just push the follow-up changes from #4072 to here. :)

@rustbot author


tests/fail-dep/libc/socketpair_closed_while_blocking.stderr
tiif, line 1: Currently the diagnostic already shows deadlock on ``write`` because we wake up the blocked ``write`` thread only if the fd blocked on ``write`` is not closed:

Relevant code snippet: 
https://github.com/rust-lang/miri/blob/3eea5b886fc60c0a5308b2314d9c8769d52b0a33/src/shims/unix/unnamed_socket.rs#L304-L314

This is acceptable, but I wanted to try if it is possible to let the diagnostic point to ``close`` and tell the user that "this thread never wakes up because of the ``close`` here".

RalfJung: > Currently the diagnostic already shows deadlock on write because we wake up the blocked write thread only if the fd blocked on write is not closed:

There's still an odd `throw_unsupported` in that logic that we should get rid of. Either it is unreachable and should become an assert, or else you can adjust the test to reach it. (I think it is reachable.)

tiif: After thinking a little bit more, I am inclined to think that the unsupported for both might ``unnamed_socket read/write`` be both unreachable. But I will add test(s) to reconfirm it.

tests/fail-dep/libc/socketpair_closed_while_read_blocking.rs
tiif, line 1: The unsupported in ``anonsocket_read`` should be unreachable too because in order to hit that path, it needs to fulfil these conditions:
1. ``read`` must be unblocked (after previously blocked)
2. The fd blocked (in this test, it is ``fds[1]``) on ``read`` is closed. 

But after the fd blocked on ``read`` (``fds[1]``) is closed, it is impossible to have a success ``write`` on ``fds[0]``, so ``read`` is never unblocked, condition ``1.`` couldn't be fulfilled.

tiif:
I initially separated it into ``read/write`` and ``anonsocket_read/write`` because I wanted to upgrade from weak to check if it is closed while blocking. 

https://github.com/rust-lang/miri/blob/1c56a5cf8f3d543021def2150a1ff8ca11faaa40/src/shims/unix/unnamed_socket.rs#L245-L248

But since it is both unreachable, it might be a good idea to combine them as suggested [here](https://github.com/rust-lang/miri/pull/4072#discussion_r1894571092), and it is easier to read too.

tests/fail-dep/libc/socketpair_closed_while_read_blocking.rs
RalfJung: Oh, right... we only wake up a thread if we successfully upgraded the weak ref stored in the peer, and then ofc the weak ref stored in the callback also must still be valid.

tiif:
These are duplicates of #4112, so I will just close this.

PR #4113, opened 2024-12-28
fix toolchain flag parsing

rmehri01:
I think this one was missed in #4036, I was getting the following when trying `./miri toolchain -c rust-analyzer`:

```console
error: unexpected argument '-c' found

  tip: to pass '-c' as a value, use '-- -c'

Usage: miri-script toolchain [FLAGS]...

For more information, try '--help'.
```


RalfJung:
Good catch, thanks!

Issue #4065, opened 2024-11-28
Handle epoll fd closed while blocking


tiif:
Instead of throwing an unsupported error, we should allow future wakeup and properly destruct the epoll fd after finish using it. 

Discussion happened in #4051

PR #4114, opened 2024-12-28
FD handling: avoid unnecessary dynamic downcasts

RalfJung:
The eventfd and socket implementations had to work with a `FileDescriptionRef` and then dynamically downcast it to the right type. That should be unnecessary, we should be able to just track what the actual type of these references is.

So that's what this PR implements: we now have `FileDescriptionRef<T>` where `T` indicates the type of the FD. We also have `DynFileDescriptionRef` for when the underlying type is not statically known. What makes this tricky is that we want to also have an `FdId` available via the reference. We can use `CoercePointee` to make `Rc<(FdId, T)>` a dyn-compatible pointer type that can be the receiver of `read`/`write`/`close`. However, that's still not quite enough since we also want dynamic downcasting, and that is only possible with `Rc<dyn Any>`, so the usual "make `Any` a supertrait" does not work... we need `FdIdWith<Self>: Any`, which doesn't work with supertrait upcasting so it needs a custom little helper trait.

This also fixes the annoying wart that `FileDescription::{read,write}` received two `self` pointers: a `&self` and a `FileDescriptionRef`.

The 2nd commit switches `epoll_wait` over to a strong ref, which is now a lot easier. That fixes https://github.com/rust-lang/miri/issues/4065.

### DIFF
--- a/src/shims/files.rs
+++ b/src/shims/files.rs
@@ -1,6 +1,7 @@
 use std::any::Any;
 use std::collections::BTreeMap;
 use std::io::{IsTerminal, Read, SeekFrom, Write};
+use std::marker::CoercePointee;
 use std::ops::Deref;
 use std::rc::{Rc, Weak};
 use std::{fs, io};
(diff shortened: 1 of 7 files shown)


src/shims/files.rs
```diff
@@ -54,8 +161,9 @@ pub trait FileDescription: std::fmt::Debug + Any {
         throw_unsup_format!("cannot seek on {}", self.name());
     }
 
-    fn close<'tcx>(
-        self: Box<Self>,
+    /// Close the file descriptor. Will only be called when `strong_count` is 1.
+    fn close_unique<'tcx>(
+        self: FileDescriptionRef<Self>,
```
RalfJung, line 166: This is the only ugly thing I wasn't able to fix... we can't convert a `Rc<dyn ...>` to a `Box<dyn ...>`, nor can we move out the inner value, so there's no better argument type we can use here.

RalfJung: I found a way to also avoid this. :)  So overall the PR doesn't add any new `unwrap`, and removes a bunch of previously existing ones. :D

Issue #4013, opened 2024-11-05
Shims for vararg functions: check that we get the right number of "fixed" arguments


RalfJung:
Most Miri shims use `check_shim` to ensure they are called with the right ABI and right number of arguments. However, some shims emulate vararg functions. There, we currently separately call `check_abi_and_shim_symbol_clash` and then `check_min_arg_count`,however, that misses potential UB: when a function, like `open`, is declared with 2 fixed args followed by varargs, then it is crucial that the caller uses a signature that actually involves 2 fixed args followed by varargs. If someone were to, say, declare this function as
```rust
pub fn open(path: *const c_char, ...) -> ::c_int;
```
and then call it as `open(path, flags)`, that is Undefined Behavior!

Similarly, non-vararg shims can actually currently be invoked with a vararg import, which should also be detected as UB.

Unfortunately, `emulate_foreign_item` is not even given enough information to detect this -- we are given a slice of `args`, but we don't learn how many of those were passed as fixed args vs varargs. So this requires changing the rustc side of this to pass more information to `find_mir_or_eval_fn` -- basically, we should pass down the full  `FnAbi`.

tiif:
This sounds interesting :) 

@rustbot claim

RalfJung:
> basically, we should pass down the full FnAbi.

In fact we probably want to pass that down *instead of* the `ExternAbi` we pass currently. That will require changing our shim ABI compat checks a bit, but that's fine.

> This sounds interesting :)

Great. :D
The first part of this, about passing more things down to `find_mir_or_eval_fn`, will have to be a rustc PR.

tiif:
> In fact we probably want to pass that down instead of the ExternAbi we pass currently. That will require changing our shim ABI compat checks a bit, but that's fine.

Do we really want to remove passing ``ExternAbi`` in ``find_mir_or_eval_fn``? It seems to be needed for ``check_abi_and_shim_symbol_clash``.  I can add a new function parameter passing down passing down ``FnAbi`` first while keeping ``ExternAbi`` .

RalfJung:
> It seems to be needed for check_abi_and_shim_symbol_clash

That should be changed to check the calling convention stored in `FnAbi`, instead of checking `ExternAbi`.

tiif:
This is currently the next thing on my queue, but I am on a break for a few days, so this probably won't be resumed until 2025 ^^.

tiif:
Progress on this issue (mostly a reminder for myself): 

[...]

PR #4122, opened 2025-01-04
Check fixed args number for variadic function 

tiif:
cc #4013 

@rustbot ready


oli-obk (approved):
Please squash

src/shims/unix/foreign_items.rs
```diff
@@ -236,6 +237,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             "open" | "open64" => {
                 // `open` is variadic, the third argument is only present when the second argument has O_CREAT (or on linux O_TMPFILE, but miri doesn't support that) set
                 this.check_abi_and_shim_symbol_clash(abi, Conv::C , link_name)?;
+                this.check_fixed_args_count("open/open64", abi, 2)?;
```
RalfJung, line 240: Now it's still fairly easy to forget to call `check_fixed_args_count`.

So instead, please add a new `this.check_shim_variadic` that is a lot like `check_shim` but returns something like `(&'a [OpTy<'tcx>; N], &'a [OpTy<'tcx>])` where the "fixed" arguments are mapped to the array and the variadic ones are returned in the slice.

src/helpers.rs
```diff
@@ -920,6 +920,22 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    /// Check the number of fixed args.
+    fn check_fixed_args_count<'a>(
+        &self,
+        name: &'a str,
+        abi: &FnAbi<'tcx, Ty<'tcx>>,
+        expected_count: u32,
+    ) -> InterpResult<'tcx> {
+        if abi.fixed_count != expected_count {
+            throw_ub_format!(
+                "incorrect number of fixed args for `{name}`: got {}, expected {expected_count}",
```
RalfJung, line 932:
```suggestion
                "incorrect number of fixed arguments for variadic function `{name}`: got {}, expected {expected_count}",
```

tiif:
So now we have this weird function definition
```rust
    fn open(
        &mut self,
        args: &[OpTy<'tcx>],
        fixed: &[OpTy<'tcx>; 2],
        _var: &[OpTy<'tcx>],

```

where ``args`` is the full function argument, ``fixed`` is the fixed args, and ``_var`` is var args. 

Since we already split fixed and var args through ``check_shim_variadic`` before we call ``this.open``, I have tried to not pass ``args`` into ``open``, but the problem is if I avoid using ``args`` and do this instead

```
            let [mode] = check_min_arg_count("open(pathname, O_CREAT, ...)", var)?
```
It will lead to diagnostic like this:
```rust
11 |     let _fd = unsafe { libc::open(name_ptr, libc::O_CREAT) }; 
   |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ incorrect number of arguments for `open(pathname, O_CREAT, ...)`: got 0, expected at least 1
```

Or maybe we can just ignore the fixed args and var args returned by ``check_shim_variadic``?

RalfJung:
> It will lead to diagnostic like this:

I would suggest to rename `check_min_arg_count` to `check_min_vararg_count` and make the error say something like "not enough variadic arguments for ...".

src/shims/unix/linux_like/syscall.rs
```diff
@@ -15,8 +15,8 @@ pub fn syscall<'tcx>(
     dest: &MPlaceTy<'tcx>,
 ) -> InterpResult<'tcx> {
     // We do not use `check_shim` here because `syscall` is variadic. The argument
-    // count is checked bellow.
-    ecx.check_abi_and_shim_symbol_clash(abi, Conv::C, link_name)?;
+    // count is checked below.
+    let ([op], _var) = ecx.check_shim_variadic::<1>(abi, Conv::C, link_name, "syscall", args)?;
```
oli-obk, line 19:
```suggestion
    let ([op], _var) = ecx.check_shim_variadic(abi, Conv::C, link_name, "syscall", args)?;
```

is that explicit length necessary? doesn't it get inferred from the return type's array length?

tiif: Oh yea, it does

src/shims/unix/foreign_items.rs
```diff
@@ -207,8 +207,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             "fcntl" => {
                 // `fcntl` is variadic. The argument count is checked based on the first argument
```
tiif, line 208: This comment looks weird, the argument count should be checked based the second argument ``cmd`` instead the first? ([source](https://docs.rs/libc/latest/libc/fn.fcntl.html))

RalfJung: Probably a copy-paste mistake.

RalfJung:
> I think as a miri specific extern function, the argument here shouldn't be classified as a vararg?

Oh dang.

We have a comment in the docs for this function saying "The `flags` argument must be `1`."
So could you try making a PR that forces the function to have 2 arguments and `flags == 1`? We recently did the same with `miri_resolve_frame`, removing support for the old "version 0 protocol"; if we do the same here then we can get rid of this use of `check_min_arg_count` hopefully. :)

src/helpers.rs
```diff
@@ -928,6 +928,30 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         interp_ok(())
     }
 
+    /// Check the number of fixed args.
```
RalfJung, line 931:
```suggestion
    /// Check the number of fixed args of a vararg function.
```
This should probably error if the function is *not* a vararg fn?

Please also put "vararg" in the name then, e.g. `check_vargarg_fixed_arg_count`.

(We used `arg_count` before so please don't use `args_count` here now.)

src/helpers.rs
```diff
@@ -990,6 +1014,23 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        exp_abi: Conv,
+        link_name: Symbol,
+        shim_name: &'a str,
+        args: &'a [OpTy<'tcx>],
+    ) -> InterpResult<'tcx, (&'a [OpTy<'tcx>; N], &'a [OpTy<'tcx>])>
+    where
+        &'a [OpTy<'tcx>; N]: TryFrom<&'a [OpTy<'tcx>]>,
+    {
+        self.check_abi_and_shim_symbol_clash(abi, exp_abi, link_name)?;
+        self.check_fixed_args_count(shim_name, abi, args)
```
RalfJung, line 1031: Why does this need `self` when `check_arg_count` does not?

src/helpers.rs
```diff
@@ -1183,7 +1224,7 @@ where
     throw_ub_format!("incorrect number of arguments: got {}, expected {}", args.len(), N)
 }
 
-/// Check that the number of args is at least the minumim what we expect.
+/// Check that the number of args is at least the minimum what we expect.
```
RalfJung, line 1227:
```suggestion
/// Check that the number of args is at least the minimum what we expect.
/// FIXME: Remove this function, use varargs and `check_min_vararg_count` instead.
```

src/shims/unix/android/thread.rs
```diff
@@ -16,18 +16,18 @@ pub fn prctl<'tcx>(
     args: &[OpTy<'tcx>],
     dest: &MPlaceTy<'tcx>,
 ) -> InterpResult<'tcx> {
-    // We do not use `check_shim` here because `prctl` is variadic. The argument
-    // count is checked bellow.
-    ecx.check_abi_and_shim_symbol_clash(abi, Conv::C, link_name)?;
+    // `prctl` is variadic. The argument count is checked below.
```
RalfJung, line 19:
```suggestion
```
Not sure which value this comment still adds.

src/shims/unix/android/thread.rs
```diff
@@ -16,18 +16,18 @@ pub fn prctl<'tcx>(
...
+    let (fixed_args, varargs) =
+        ecx.check_shim_variadic::<1>(abi, Conv::C, link_name, "prctl", args)?;
 
     // FIXME: Use constants once https://github.com/rust-lang/libc/pull/3941 backported to the 0.2 branch.
     let pr_set_name = 15;
     let pr_get_name = 16;
 
-    let [op] = check_min_arg_count("prctl", args)?;
+    let [op] = fixed_args;
```
RalfJung, line 27:
```suggestion
```

src/shims/unix/android/thread.rs
```diff
@@ -16,18 +16,18 @@ pub fn prctl<'tcx>(
     args: &[OpTy<'tcx>],
     dest: &MPlaceTy<'tcx>,
 ) -> InterpResult<'tcx> {
-    // We do not use `check_shim` here because `prctl` is variadic. The argument
-    // count is checked bellow.
-    ecx.check_abi_and_shim_symbol_clash(abi, Conv::C, link_name)?;
+    // `prctl` is variadic. The argument count is checked below.
+    let (fixed_args, varargs) =
+        ecx.check_shim_variadic::<1>(abi, Conv::C, link_name, "prctl", args)?;
```
RalfJung, line 21:
```suggestion
    let ([op], varargs) =
        ecx.check_shim_variadic(abi, Conv::C, link_name, "prctl", args)?;
```
Same everywhere else -- you should never use turbofish here.

src/shims/unix/fd.rs
```diff
@@ -127,10 +127,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         interp_ok(Scalar::from_i32(this.try_unwrap_io_result(result)?))
     }
 
-    fn fcntl(&mut self, args: &[OpTy<'tcx>]) -> InterpResult<'tcx, Scalar> {
+    fn fcntl(
+        &mut self,
+        fixed_args: &[OpTy<'tcx>; 2],
```
RalfJung, line 132: Instead of passing a 2-element array, just pass two arguments.

RalfJung (commented):
I left some comments on the first place where you used the new infra; they apply everywhere in the PR.

src/helpers.rs
tiif: Oh ``check_arg_count`` is free function, I will move ``check_vargarg_fixed_arg_count`` to free function too.

tiif:
> So could you try making a PR that forces the function to have 2 arguments and flags == 1? We recently did the same with miri_resolve_frame, removing support for the old "version 0 protocol"; if we do the same here then we can get rid of this use of check_min_arg_count hopefully. :)

Alright, I can send a follow-up PR after this landed.

src/helpers.rs
```diff
@@ -1198,6 +1216,49 @@ pub fn check_min_arg_count<'a, 'tcx, const N: usize>(
...
+/// Check the number of fixed args of a vararg function.
+/// Returns a tuple that consisting of an array of fixed args, and a slice of varargs.
+fn check_vargarg_fixed_arg_count<'a, 'tcx, const N: usize>(
+    name: &'a str,
+    abi: &FnAbi<'tcx, Ty<'tcx>>,
+    args: &'a [OpTy<'tcx>],
+) -> InterpResult<'tcx, (&'a [OpTy<'tcx>; N], &'a [OpTy<'tcx>])> {
+    if !abi.c_variadic {
+        panic!("This should not be used on non-vararg function");
+    }
```
tiif, line 1245: I am going for a panic here, but we can use other mechanism to prevent this being used at the wrong place.

RalfJung: Which other mechanism did you have in mind?

RalfJung: Actually yes this should not be a panic, this should be `throw_ub_format!`. It is UB to call a variadic function with a non-variadic caller-side signature.

src/shims/unix/foreign_items.rs
```diff
@@ -236,8 +235,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             "open" | "open64" => {
                 // `open` is variadic, the third argument is only present when the second argument
                 // has O_CREAT (or on linux O_TMPFILE, but miri doesn't support that) set
-                this.check_abi_and_shim_symbol_clash(abi, Conv::C, link_name)?;
-                let result = this.open(args)?;
+                let ([path_raw, flag], varargs) =
+                    this.check_shim_variadic(abi, Conv::C, link_name, "open/open64", args)?;
```
RalfJung, line 239: Just keep using `link_name` here, that will always match the name the user used which is better. This also applies everywhere else you changed this.

src/helpers.rs
tiif: > t is UB to call a variadic function with a non-variadic caller-side signature.

I wonder if there is any way to test for this?

RalfJung: Yes, just write an `extern` block declaring `open` as a non-variadic function and call that.

tests/fail-dep/libc/wrong_fixed_arg_count.rs
RalfJung, line 1: This should not be a pass-dep test, it's not actually about libc after all.  `tests/fail/shims` seems like a better place.

src/helpers.rs
```diff
@@ -1198,6 +1215,53 @@ pub fn check_min_arg_count<'a, 'tcx, const N: usize>(
...
+/// Check the number of fixed args of a vararg function.
+/// Returns a tuple that consisting of an array of fixed args, and a slice of varargs.
+fn check_vargarg_fixed_arg_count<'a, 'tcx, const N: usize>(
+    link_name: Symbol,
+    abi: &FnAbi<'tcx, Ty<'tcx>>,
+    args: &'a [OpTy<'tcx>],
+) -> InterpResult<'tcx, (&'a [OpTy<'tcx>; N], &'a [OpTy<'tcx>])> {
+    if !abi.c_variadic {
+        throw_ub_format!(
+            "it is incorrect to call variadic function with a non-variadic caller-side signature"
```
RalfJung, line 1244:
```suggestion
            "calling a variadic function with a non-variadic caller-side signature"
```

tests/fail-dep/libc/wrong_fixed_arg_count.rs
tiif: After I moved it to ``./tests/fail/shims``, when I run ``./miri test``, the error ``undeclared crate or module `libc` `` appears. 

```rust
error[E0433]: failed to resolve: use of undeclared crate or module `libc`
  --> tests/fail/shims/wrong_fixed_arg_count.rs:12:46
   |
LL |     let _fd = unsafe { open(c_path.as_ptr(), libc::O_RDWR) };
   |                                              ^^^^ use of undeclared crate or module `libc`

error[E0433]: failed to resolve: use of undeclared crate or module `libc`
  --> tests/fail/shims/wrong_fixed_arg_count.rs:7:26
   |
LL |     fn open(path: *const libc::c_char, ...) -> libc::c_int;
   |                          ^^^^ use of undeclared crate or module `libc`

error[E0433]: failed to resolve: use of undeclared crate or module `libc`
  --> tests/fail/shims/wrong_fixed_arg_count.rs:7:48
   |
LL |     fn open(path: *const libc::c_char, ...) -> libc::c_int;
   |                                                ^^^^ use of undeclared crate or module `libc`

```

I guess if I need to use something related to ``libc``, I need to put it in ``fail-dep``.

tiif: Hmm should this even happen? The comment inside test_dependencies said ``libc`` should be able to be used in all files under ``tests/``

https://github.com/rust-lang/miri/blob/ce9c9ff2de0aaf410fd461d4f8349b8765894820/test_dependencies/Cargo.toml#L11-L12

RalfJung: Replace the use of `libc` by [`std:ffi`](https://doc.rust-lang.org/nightly/std/ffi/index.html).

RalfJung: > The comment inside test_dependencies said libc should be able to be used in all files under tests/

The comment is wrong; it is only for the tests in `*-deps`.

RalfJung:
Please also squash the PR into a single commit.

tiif:
Ooh I accidentally push to github then rebase then push, so the change is not visible from ``compare change``. 

What has changed:
- I use ``/* value does not matter */ 0`` for ``wrong_fixed_arg_count ``test too since it also has magic number
- bless the test 

@rustbot ready

RalfJung:
Looks good :)

PR #4129, opened 2025-01-09
Supported fioclex for ioctl on macos

### LINKED ISSUE #4126: Support ``FIOCLEX`` for ``ioctl`` in macos
Currently ``std::pipe`` on macOS is blocked on ``ioctl`` with ``FIOCLEX`` flag. 

```rust
error: unsupported operation: can't call foreign function `ioctl` on OS `macos`
   --> /home/tiif/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/sys/pal/unix/fd.rs:510:17
    |
510 |             cvt(libc::ioctl(self.as_raw_fd(), libc::FIOCLEX))?;
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't call foreign function `ioctl` on OS `macos`
```

It should be fairly easy to support because ``FIOCLEX`` is a ``close-on-exec`` flag. Since miri doesn't support ``exec``, nothing needs to be done there except for some checks. It might be similar to how ``fcntl`` handles ``FD_CLOXEC``. 

https://github.com/rust-lang/miri/blob/d9894a37a517ca7d367c681b39beb2064ca1ed46/src/shims/unix/fd.rs#L155-L173

We just need to add a shim for ``ioctl``, and support ``FIOCLEX`` flag on it. It'd be nice to test ``std::pipe`` after adding this and see what else do we need to support in order to get it ``std::pipe`` test work on macOS. 

@rustbot label +A-shims +C-enhancement +E-good-first-issue +A-mac

geetanshjuneja:
Closes #4126


RalfJung:
Thanks for the PR!

Please also add a test calling the function, as noted by @tiif. If you cannot make the test pass, that's fine, please still push it and then hopefully someone has the time to look at it and help. :)

src/shims/unix/fd.rs
```diff
@@ -189,6 +189,26 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         }
     }
 
+    fn ioctl(&mut self, args: &[OpTy<'tcx>]) -> InterpResult<'tcx, Scalar> {
+        let this = self.eval_context_mut();
+
+        let [fd_num, cmd] = check_min_arg_count("ioctl", args)?;
+        let fioclex = this.eval_libc_i32("FIOCLEX");
```
tiif, line 196: ``FIOCLEX`` is ``u64`` in macos ([you can find it here](https://docs.rs/libc/latest/aarch64-apple-darwin/libc/constant.FIOCLEX.html)), so consider adding a ``eval_libc_u64`` . It is similar to https://github.com/rust-lang/miri/blob/3736728c97d3c8484b63136e4b21a03a2ae5b91d/src/helpers.rs#L258-L263, but everything ``u32`` there will change to u64. Then use that helper function here.

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -77,6 +79,21 @@ fn test_dup_stdout_stderr() {
...
 
+#[cfg(target_os = "macos")]
+fn test_ioctl() {
+    let path = utils::prepare_with_content("miri_test_libc_ioctl.txt", &[]);
+
+    let mut name = path.into_os_string();
+    name.push("\0");
+    let name_ptr = name.as_bytes().as_ptr().cast::<libc::c_char>();
+    unsafe {
+        assert_eq!(libc::ioctl(1, libc::FIOCLEX), -1);
```
tiif, line 90: If the intention of this is to test the ``EBADF`` path, this probably won't work because fd value 1 and 2 exist by default in miri. So please assign a higher fd value (larger than 2).

src/shims/unix/fd.rs
geetanshjuneja: I think we should use `i64` instead of `u64` as cmd is of type `i32`

tiif: It might be possible to do 

```
let cmd = this.read_scalar(cmd)?.to_u64()?;
```

src/helpers.rs
```diff
@@ -254,6 +254,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         })
     }
 
+    /// Helper function to get a `libc` constant as an `u64`.
+    fn eval_libc_u64(&self, name: &str) -> u64 {
```
RalfJung, line 258: Please pick some sensible order for these functions. i32, u64, u32 is definitely not sensible. ;)

src/shims/unix/fd.rs
```diff
@@ -189,6 +189,26 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         }
     }
 
+    fn ioctl(&mut self, args: &[OpTy<'tcx>]) -> InterpResult<'tcx, Scalar> {
```
RalfJung, line 192: If this is only called from macOS, it shouldn't live in this file. `macos/foreign_items.rs` seems like a fine location.

src/shims/unix/fd.rs
```diff
@@ -189,6 +189,26 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let this = self.eval_context_mut();
+
+        let [fd_num, cmd] = check_min_arg_count("ioctl", args)?;
+        let fioclex = this.eval_libc_u64("FIOCLEX");
+
+        let fd_num = this.read_scalar(fd_num)?.to_i32()?;
+        let cmd = this.read_scalar(cmd)?.to_u64()?;
```
RalfJung, line 199:
```suggestion
        let fioclex = this.eval_libc_u64("FIOCLEX");

        let [fd_num, cmd] = check_min_arg_count("ioctl", args)?;
        let fd_num = this.read_scalar(fd_num)?.to_i32()?;
        let cmd = this.read_scalar(cmd)?.to_u64()?;
```
Let's separate the steps of parsing our arguments, and evaluating some global constants.

src/shims/unix/fd.rs
```diff
@@ -189,6 +189,26 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        if cmd == fioclex {
+            if this.machine.fds.is_fd_num(fd_num) {
```
RalfJung, line 202:
```suggestion
            // Since we don't support `exec`, this is a NOP. However, we want to
            // return EBADF if the FD is invalid.
            if this.machine.fds.is_fd_num(fd_num) {
```

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -17,6 +17,8 @@ mod utils;
 fn main() {
     test_dup();
     test_dup_stdout_stderr();
+    #[cfg(target_os = "macos")]
+    test_ioctl();
```
RalfJung, line 21: Please add new things at the end unless there's a good reason to put them elsewhere.

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -77,6 +79,21 @@ fn test_dup_stdout_stderr() {
...
 
+#[cfg(target_os = "macos")]
+fn test_ioctl() {
+    let path = utils::prepare_with_content("miri_test_libc_ioctl.txt", &[]);
+
+    let mut name = path.into_os_string();
+    name.push("\0");
+    let name_ptr = name.as_bytes().as_ptr().cast::<libc::c_char>();
+    unsafe {
+        assert_eq!(libc::ioctl(10, libc::FIOCLEX), -1);
```
RalfJung, line 90:
```suggestion
        // 10 surely is an invalid FD.
        assert_eq!(libc::ioctl(10, libc::FIOCLEX), -1);
```

RalfJung, line 90: Please also check the errno here.

RalfJung (changes requested):
This looks great, thanks! I only have minor comments.

RalfJung:
@rustbot author

RalfJung:
That should do it, thanks!

Please squash the commits. (Keep the new pipe test a separate commit if you can, but it's not a big deal to have that squashed as well.)

Issue #4062, opened 2024-11-26
Add TID support to sched_getaffinity/sched_setaffinity


RalfJung:
Our sched_getaffinity/sched_setaffinity shims currently only support a `pid` of 0, indicating the current thread. However, now that we have `gettid`, we should support using a TID in sched_getaffinity/sched_setaffinity as well. We probably want some helper functions to convert between a `LinuxTid` and a `ThreadId`, to ensure this is consistent.

Note that `gettid` is Linux-specific, while we support sched_getaffinity/sched_setaffinity also on FreeBSD. This new support for non-0 `pid` should remain Linux-specific for now; `gettid` does not seem to be exposed by the libc on FreeBSD.

RalfJung:
FWIW apparently `gettid` is also a [thing on FreeBSD](https://man.freebsd.org/cgi/man.cgi?query=gettid&sektion=2&apropos=0&manpath=CentOS+7.1) so we could do the same there.

EDIT: Turns out it's not actually a thing, see https://github.com/rust-lang/libc/issues/5095.

PR #4130, opened 2025-01-09
Add support for tid argument in sched_(get/set)affinity

sleiderr:
Linked to #4062 

On Linux, values returned by `gettid` can be passed as the `pid` argument when calling `sched_getaffinity` or `sched_setaffinity`.

### DIFF
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -677,9 +677,17 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let cpusetsize = this.read_target_usize(cpusetsize)?;
                 let mask = this.read_pointer(mask)?;
 
-                // TODO: when https://github.com/rust-lang/miri/issues/3730 is fixed this should use its notion of tid/pid
-                let thread_id = match pid {
-                    0 => this.active_thread(),
+                let thread_id = match (pid, &*this.tcx.sess.target.os) {
+                    (0, _) => this.active_thread(),
+                    (tid, "linux") => {
+                        if let Ok(tid) = this.machine.threads.thread_id_try_from(tid) {
+                            tid
+                        } else {
+                            this.set_last_error_and_return(LibcError("ESRCH"), dest)?;
+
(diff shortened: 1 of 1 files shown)


src/shims/unix/foreign_items.rs
```diff
@@ -677,9 +677,17 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let cpusetsize = this.read_target_usize(cpusetsize)?;
                 let mask = this.read_pointer(mask)?;
 
-                // TODO: when https://github.com/rust-lang/miri/issues/3730 is fixed this should use its notion of tid/pid
-                let thread_id = match pid {
-                    0 => this.active_thread(),
+                let thread_id = match (pid, &*this.tcx.sess.target.os) {
```
RalfJung, line 680: Seems a bit odd to match on the pair here. I think this is more clear as
```rust
if pid == 0 { ... }
else if this.tcx.sess.target.os == "linux" {
  // On Linux we support `gettid`, which is guaranteed to support IDs matching `sched_getaffinity`.
  ...
} else { ... }
```

src/shims/unix/foreign_items.rs
```diff
@@ -677,9 +677,17 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    (0, _) => this.active_thread(),
+                    (tid, "linux") => {
+                        if let Ok(tid) = this.machine.threads.thread_id_try_from(tid) {
```
RalfJung, line 683: This will not return the right ID. See the gettid implementation to see how ThreadId is converted to the user-visible TID; this here will have to do the opposite transformation. Please put both directions in some shared place to make sure they remain consistent in the future.

RalfJung (changes requested):
Thanks for the PR! However, as-is this doesn't quite have the right behavior.

Please also add a test ensuring we can set an affinity via the TID and then get it via "0" and vice versa.

RalfJung:
I'm going to close this PR due to inactivity. Thanks for trying to contribute, and feel free to reopen or make a new PR if you want to get back it it again!

PR #4132, opened 2025-01-10
Switched FreeBSD to pthread_setname_np

YohDeadfall:
Updated FreeBSD name handling and tests to reflect the changes made in https://github.com/rust-lang/rust/pull/132607.

### DIFF
--- a/src/shims/unix/freebsd/foreign_items.rs
+++ b/src/shims/unix/freebsd/foreign_items.rs
@@ -21,29 +21,38 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let this = self.eval_context_mut();
         match link_name.as_str() {
             // Threading
-            "pthread_set_name_np" => {
+            "pthread_setname_np" => {
                 let [thread, name] = this.check_shim(abi, Conv::C, link_name, args)?;
                 let max_len = usize::MAX; // FreeBSD does not seem to have a limit.
-                // FreeBSD's pthread_set_name_np does not return anything.
-                this.pthread_setname_np(
+                let res = match this.pthread_setname_np(
                     this.read_scalar(thread)?,
                     this.read_scalar(name)?,
                     max_len,
                     /* truncate */ false,
-                )?;
+                )? {
+                    ThreadNameResult::Ok => Scalar::from_u32(0),
+                    ThreadNameResult::NameTooLong => unreachable!(),
+                    ThreadNameResult::ThreadNotFound => this.eval_libc("ESRCH"),
+                };
+                this.write_scalar(res, dest)?;
             }
-            "pthread_get_name_np" => {
+            "pthread_getname_np" => {
                 let [thread, name, len] = this.check_shim(abi, Conv::C, link_name, args)?;
-                // FreeBSD's pthread_get_name_np does not return anything
-                // and uses strlcpy, which truncates the resulting value,
+                // FreeBSD's pthread_getname_np uses strlcpy, which truncates the resulting value,
                 // but always adds a null terminator (except for zero-sized buffers).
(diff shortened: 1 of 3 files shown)


src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -21,29 +21,38 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                     this.read_scalar(thread)?,
                     this.read_scalar(name)?,
                     max_len,
                     /* truncate */ false,
-                )?;
+                )? {
+                    ThreadNameResult::Ok => Scalar::from_u32(0),
+                    ThreadNameResult::NameTooLong => unreachable!(),
+                    ThreadNameResult::ThreadNotFound => this.eval_libc("ESRCH"),
+                };
```
oli-obk, line 36: Can any of this be deduplicated (moved into the function) with the other pthread_setname_np callers or is it still different?

YohDeadfall: The `np` suffix stands for `non-portable, and it's not without a reason (: Still, in the current case it's almost identical to what macOS has except the comments and a small difference in the signatures (macOS has no `thread` argument for the setter).

Wrapping it up, we can have some `bsd_like` module with two functions:

```rust
fn pthread_setname_np<'tcx>(
    ecx: &mut MiriInterpCx<'tcx>,
    thread: OpTy<'tcx>,
    name: OpTy<'tcx>,
    max_len: usize,
    match this.pthread_setname_np(
        this.read_scalar(thread)?,
        this.read_scalar(name)?,
        max_len,
        /* truncate */ false,
    )? {
        ThreadNameResult::Ok => Scalar::from_u32(0),
        ThreadNameResult::NameTooLong => unreachable!(),
        ThreadNameResult::ThreadNotFound => this.eval_libc("ESRCH"),
    }
}


fn pthread_getname_np<'tcx>(
    ecx: &mut MiriInterpCx<'tcx>,
    thread: OpTy<'tcx>,
    name: OpTy<'tcx>,
    len: OpTy<'tcx>,
    match this.pthread_getname_np(
        this.read_scalar(thread)?,
        this.read_scalar(name)?,
        this.read_scalar(len)?,
        /* truncate */ true,
    )? {
        ThreadNameResult::Ok => Scalar::from_u32(0),
        // `NameTooLong` is possible when the buffer is zero sized,
        ThreadNameResult::NameTooLong => Scalar::from_u32(0),
        ThreadNameResult::ThreadNotFound => this.eval_libc("ESRCH"),
    }
}
```

oli-obk: Oh heh. Thanks, that explanation makes sense and the "duplication" as well

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -21,29 +21,38 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let res = match this.pthread_getname_np(
                     this.read_scalar(thread)?,
                     this.read_scalar(name)?,
                     this.read_scalar(len)?,
                     /* truncate */ true,
-                )?;
+                )? {
+                    ThreadNameResult::Ok => Scalar::from_u32(0),
+                    // `NameTooLong` is possible when the buffer is zero sized,
+                    ThreadNameResult::NameTooLong => Scalar::from_u32(0),
```
RalfJung, line 52: Do we have a test for the zero-sized buffer case?

YohDeadfall: Yes, it was originally done for macOS. Line 130 in the tests.

RalfJung: Great, thanks :)

PR #4133, opened 2025-01-11
Supported fioclex for ioctl on macos

### LINKED ISSUE #4126: Support ``FIOCLEX`` for ``ioctl`` in macos

geetanshjuneja:
Closes #4126

### DIFF
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -262,6 +262,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         })
     }
 
+    /// Helper function to get a `libc` constant as an `u64`.
+    fn eval_libc_u64(&self, name: &str) -> u64 {
+        // TODO: Cache the result.
+        self.eval_libc(name).to_u64().unwrap_or_else(|_err| {
+            panic!("required libc item has unexpected type (not `u64`): {name}")
+        })
+    }
+
     /// Helper function to get a `windows` constant as a `Scalar`.
     fn eval_windows(&self, module: &str, name: &str) -> Scalar {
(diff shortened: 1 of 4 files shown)


geetanshjuneja:
@RalfJung sorry messed up the last PR, looks like I need to practise git.

RalfJung:
Yeah git can be hard to use but rebasing is an important skill. This is a small PR so I can quickly re-review this, no big deal. :)

PR #4134, opened 2025-01-11
adjust the way we build miri-script in RA, to fix proc-macros

RalfJung:
RA doesn't seem to like it when different parts of the workspace are built with different compilers, and we can't even mix those in the same `target` dir it seems.

Also use a less repetitive way to set env vars for the cargo invocations.


oli-obk:
Oof

PR #4136, opened 2025-01-12
Illumos: Added epoll and eventfd

### LINKED ISSUE #4017: Illumos: Add support for `epoll` and `eventfd`
According to Illumos man pages, it supports [`epoll`](https://www.illumos.org/man/7/epoll) and [`eventfd`](https://www.illumos.org/man/3C/eventfd). From `libc` crate it's fine too.

@rustbot label +A-solarish +C-enhancement
@rustbot claim

YohDeadfall:
Fixes #4017.

### DIFF
--- a/src/shims/unix/solarish/foreign_items.rs
+++ b/src/shims/unix/solarish/foreign_items.rs
@@ -3,6 +3,8 @@ use rustc_span::Symbol;
 use rustc_target::callconv::{Conv, FnAbi};
 
 use crate::shims::unix::foreign_items::EvalContextExt as _;
+use crate::shims::unix::linux_like::epoll::EvalContextExt as _;
+use crate::shims::unix::linux_like::eventfd::EvalContextExt as _;
 use crate::shims::unix::*;
 use crate::*;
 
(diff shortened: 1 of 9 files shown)


tests/pass-dep/libc/libc-eventfd.rs
```diff
@@ -116,10 +116,16 @@ fn test_race() {
 
 // This is a test for calling eventfd2 through a syscall.
 fn test_syscall() {
-    let initval = 0 as libc::c_uint;
-    let flags = (libc::EFD_CLOEXEC | libc::EFD_NONBLOCK) as libc::c_int;
-    let fd = unsafe { libc::syscall(libc::SYS_eventfd2, initval, flags) };
-    assert_ne!(fd, -1);
+    cfg_if::cfg_if! {
+        if #[cfg(target_os = "illumos")] {
```
RalfJung, line 120: Please add the `#[cfg]` to the function instead.

RalfJung:
Good catch!

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -21,6 +23,32 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     ) -> InterpResult<'tcx, EmulateItemResult> {
         let this = self.eval_context_mut();
         match link_name.as_str() {
+            // epoll, eventfd
+            "epoll_create1" => {
+                this.assert_target_os("illumos", "gettid");
```
RalfJung, line 28: The point of this is to prevent these from being exposed on Solarish, I assume?

If yes, fine to keep them, but please add a comment -- and please fix the function name. :)

tests/fail-dep/libc/libc_epoll_block_two_thread.rs
```diff
@@ -1,7 +1,7 @@
 //@compile-flags: -Zmiri-preemption-rate=0
 //~^ERROR: deadlocked
 //~^^ERROR: deadlocked
-//@only-target: linux
+//@only-target: linux illumos
```
RalfJung, line 4: Shouldn't this include `android` as well?

YohDeadfall: Yeah, it should.

src/shims/unix/solarish/foreign_items.rs
YohDeadfall: Yes, there's a difference between Solaris and Illumos. The last one implements `epoll` as Linux does, but doesn't include an ability to use it through `syscall`.

YohDeadfall: What if `assert_target_os` would take a link name itself? It should be less error prone then.

RalfJung: We call it in many places where that would be awkward.

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -21,6 +23,32 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     ) -> InterpResult<'tcx, EmulateItemResult> {
         let this = self.eval_context_mut();
         match link_name.as_str() {
+            // epoll, eventfd
```
RalfJung, line 26:
```suggestion
            // epoll, eventfd (NOT available on Solaris!)
```

src/shims/unix/solarish/foreign_items.rs
RalfJung: It is rather unusual to have `assert_target_os` in `foreign_items`. In fact we never did that before your PR. So I don't think this is worth it.

RalfJung:
When updating a PR, please *either* do a force push *or* add a new commit. If you first add a new commit and the force-push that away, Github makes it entirely impossible for me to see the diff since my last review. That's clearly a sign of Github being a terrible tool for PR review, but we have to live with the tools we have and work around their limitations...

PR #4137, opened 2025-01-12
Added Android to `epoll` and `eventfd` test targets

YohDeadfall:
It seems like `android` was forgotten to be targeted in all `epoll` and `eventfd` tests while it's a Linux-based OS and supports these operations.


saethlin:
Good catch!

Issue #3841, opened 2024-08-24
Shims cannot trust user-defined types, need to always use `deref_pointer_as`


RalfJung:
Calling `deref_pointer` means we create a place at whatever type that argument was declared at by the user. But we can't trust that type, we should ideally entirely ignore it and use `deref_pointer_as` everywhere. Failure to do so can lead to ICEs like https://github.com/rust-lang/miri/issues/3839.

Once this is consistently done everywhere, we can also remove the [`ScalarSizeMismatch`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/interpret/struct.ScalarSizeMismatch.html) error from the interpreter. This was added because otherwise ICEs can be triggered by using an incorrect scalar type in the signature of a function that has a Miri shim, but as we have seen, ICEs can still be triggered -- and once we use `deref_pointer_as` everywhere, `ScalarSizeMismatch` can no longer be triggered at all.

To help with this, do the following:
- `rg -w deref_pointer src/shims`
- Pick one of the remaining places where we call `deref_pointer`
- Figure out which type that argument is supposed to have
- Replace the call with `deref_pointer_as` and the correct type

Getting the correct type will be the hardest part of this, and it depends on what the correct type is. Many basic types are available in `this.machine.layouts`, but `struct` types need to be looked up in `libc`/`std` or maybe we even have to construct one ourselves (or entirely change the code). If you are unsure about a concrete case, just ask here. :)

oli-obk:
Once all of them are converted we should probably add a `./miri tidy` (or make it part of `./miri check` and similar commands) that automatically greps and asserts that there are no new ones introduced

RalfJung:
Ah yes that's a good low-tech solution.

badumbatish:
can i have a try at this?

RalfJung:
Sure. :) I'd suggest to start with just one file and fix the `deref_pointer` there, to make sure you get the hang of this.

Unfortunately I won't have the time to provide a lot of mentoring. You can always try Zulip, there are other people there that can help.

geetanshjuneja:
Hi, If no one working on this issue. Can I work on this?

RalfJung:
@geetanshjuneja yes feel free to. :) [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/269128-miri) is a great place to ask questions. For your first PR, please only change a few places so that we can make sure things are working as intended.

PR #4138, opened 2025-01-13
Use deref_poiner_as instead of deref_pointer

geetanshjuneja:
Links to #3841.


RalfJung:
Yes this looks correct, thanks :)

PR #4140, opened 2025-01-17
Use deref_pointer_as instead of deref_pointer

geetanshjuneja:
Links to #3841.


src/shims/time.rs
```diff
@@ -132,7 +132,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         this.assert_target_os_is_unix("localtime_r");
         this.check_no_isolation("`localtime_r`")?;
 
-        let timep = this.deref_pointer(timep)?;
+        let timep = this.deref_pointer_as(timep, this.libc_ty_layout("time_t"))?;
```
RalfJung, line 135: `this.libc_ty_layout("time_t")` is now called twice in this function, can you move this into a `let`-bound variable so we call it only once?

src/shims/unix/foreign_items.rs
```diff
@@ -894,7 +894,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let _attr_place =
                     this.deref_pointer_as(attr_place, this.libc_ty_layout("pthread_attr_t"))?;
                 let addr_place = this.deref_pointer(addr_place)?;
-                let size_place = this.deref_pointer(size_place)?;
+                let size_place =
+                    this.deref_pointer_as(size_place, this.libc_ty_layout("size_t"))?;
```
RalfJung, line 898: Below we pretty much hard-code this to be ptr-sized, so you might as well use `this.machine.layouts.usize`.

And then please adjust `addr_place`  just above as well so that we are done with this particular shim. :)

src/shims/unix/fs.rs
```diff
@@ -1253,15 +1253,17 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    this.deref_pointer_as(result_op, this.machine.layouts.mut_raw_ptr)?;
                 this.write_scalar(this.read_scalar(entry_op)?, &result_place)?;
 
                 Scalar::from_i32(0)
             }
             None => {
                 // end of stream: return 0, assign *result=NULL
-                this.write_null(&this.deref_pointer(result_op)?)?;
+                let result_place =
+                    this.deref_pointer_as(result_op, this.machine.layouts.mut_raw_ptr)?;
```
RalfJung, line 1265: Please move this to the top of the function (just after `let dirp`) to avoid duplicating it.

src/shims/unix/unnamed_socket.rs
```diff
@@ -362,7 +362,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let domain = this.read_scalar(domain)?.to_i32()?;
         let mut flags = this.read_scalar(type_)?.to_i32()?;
         let protocol = this.read_scalar(protocol)?.to_i32()?;
-        let sv = this.deref_pointer(sv)?;
+        let sv = this.deref_pointer_as(sv, this.machine.layouts.i32)?;
```
RalfJung, line 365:
```suggestion
        // This is really a pointer to `[i32; 2]` but we use a ptr-to-first-element representation.
        let sv = this.deref_pointer_as(sv, this.machine.layouts.i32)?;
```

geetanshjuneja:
@RalfJung since libc is reliably available on Windows targets, I won't be able to get windows specific`layouts` like `this.libc_ty_layout("CONSOLE_SCREEN_BUFFER_INFO")`. What should I do in this case?

RalfJung:
Yeah `CONSOLE_SCREEN_BUFFER_INFO` is not in libc. It's a Windows API type. I'd say remove that one from the PR for now and open a Zulip thread; either we can find this type declaration somewhere else or we have to write some helpers to declare such struct types entirely ourselves.

RalfJung:
Wait, did you add more `deref_pointer` → `deref_pointer_as` after my first review? Please don't do that, and *especially* please don't do that without explicitly posting a comment saying that you did! The general expectation is that one a review was done, the PR will not be extended in scope any more; all remaining changes are for resolving review concerns.

I will have to do a complete re-review, and since I am traveling this will take at least a week, maybe two.

geetanshjuneja:
> Wait, did you add more `deref_pointer` → `deref_pointer_as` after my first review? Please don't do that, and _especially_ please don't do that without explicitly posting a comment saying that you did!

Sorry for not mentioning earlier that the PR is incomplete and making you review it unnecessarily. I have now made all the changes except for the windows part.

> I will have to do a complete re-review, and since I am traveling this will take at least a week, maybe two.
I am not in a hurry, do re-review whenever you are free.

geetanshjuneja:
@RalfJung plz review this.

src/shims/time.rs
```diff
@@ -132,16 +132,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         this.assert_target_os_is_unix("localtime_r");
         this.check_no_isolation("`localtime_r`")?;
 
-        let timep = this.deref_pointer(timep)?;
+        let timep_layout = this.libc_ty_layout("time_t");
```
RalfJung, line 135:
```suggestion
        let time_layout = this.libc_ty_layout("time_t");
```
This is not the layout of a pointer, after all. `timep` is just a pointer-to-time.

src/shims/windows/sync.rs
```diff
@@ -85,7 +86,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
         let id = this.init_once_get_data(init_once_op)?.id;
         let flags = this.read_scalar(flags_op)?.to_u32()?;
-        let pending_place = this.deref_pointer(pending_op)?;
+        // PBOOL is a typedef of int
```
RalfJung, line 89:
```suggestion
        // PBOOL is int*
```

RalfJung (changes requested):
Almost there :)

RalfJung:
@rustbot author

RalfJung:

@rustbot author

geetanshjuneja:
> Please use the `--keep-base` flag when squashing so that the force-push diff is easier to review. 
Sorry I just rewrote the history from the next time I will make sure that the new commit message comes with a concatenation of the existing commit messages.

geetanshjuneja:
I didn't made any more changes so you don't need to review again.

RalfJung:
Looks good, thanks. :)

RalfJung (approved):

PR #4142, opened 2025-01-19
shim Apple's futex primitives

joboet:
This is necessary to unblock rust-lang/rust#122408. The documentation for these is available [here](https://developer.apple.com/documentation/os/os_sync_wait_on_address?language=objc).

Because the futex wait operations (`os_sync_wait_on_address` et al.) return the number of remaining waiters after returning, this required some changes to the common futex infrastructure, which I've changed to take a callback instead of precalculating the return values.

### DIFF
--- a/src/shims/unix/macos/sync.rs
+++ b/src/shims/unix/macos/sync.rs
@@ -10,14 +10,38 @@
 //! and we do not detect copying of the lock, but macOS doesn't guarantee anything
 //! in that case either.
 
+use std::cell::Cell;
+use std::time::Duration;
+
 use rustc_abi::Size;
 
+use crate::concurrency::sync::FutexRef;
 use crate::*;
 
 #[derive(Clone)]
 enum MacOsUnfairLock {
     Poisoned,
     Active { mutex_ref: MutexRef },
+}
+
+pub enum MacOsFutexTimeout<'a, 'tcx> {
+    None,
+    Relative { clock_op: &'a OpTy<'tcx>, timeout_op: &'a OpTy<'tcx> },
+    Absolute { clock_op: &'a OpTy<'tcx>, timeout_op: &'a OpTy<'tcx> },
+}
+
+/// Metadata for a macOS futex.
+///
+/// Since macOS 11.0, Apple has exposed the previously private futex API consisting
+/// of `os_sync_wait_on_address` (and friends) and `os_sync_wake_by_address_{any, all}`.
+/// These work with different value sizes and flags, which are validated to be consistent.
+/// This structure keeps track of both the futex queue and these values.
+struct MacOsFutex {
+    futex: FutexRef,
+    /// The size in bytes of the atomic primitive underlying this futex.
+    size: Cell<u64>,
+    /// Whether the futex is shared across process boundaries.
+    shared: Cell<bool>,
 }
 
 impl<'tcx> EvalContextExtPriv<'tcx> for crate::MiriInterpCx<'tcx> {}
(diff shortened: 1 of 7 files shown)


RalfJung:
Thanks for the PR! I just left for a 1-week business trip and will have to catch up with stuff when I am back so it could take a bit before I can take a closer look.

src/shims/unix/macos/sync.rs
```diff
@@ -20,6 +24,24 @@ enum MacOsUnfairLock {
...
+/// Metadata for a macOS futex.
+///
+/// Since macOS 11.0, Apple has exposed the previously private futex API consisting
+/// of `os_sync_wait_on_address` (and friends) and `os_sync_wake_by_address_{any, all}`.
+/// These work with different value sizes and flags, which are validated to be consistent.
+/// This structure keeps track of both the futex queue and these values.
+struct MacOsFutex {
+    futex: FutexRef,
+    size: Cell<u64>,
+    shared: Cell<bool>,
```
RalfJung, line 42: Please add doc comments explaining these fields

src/shims/unix/macos/sync.rs
```diff
@@ -54,6 +76,191 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                Some((clock, TimeoutAnchor::Relative, timeout))
+            }
+            MacOsFutexTimeout::Absolute { clock_op, timeout_op } => {
+                let clock = this.read_scalar(clock_op)?.to_u32()?;
+                let timeout = this.read_scalar(timeout_op)?.to_u64()?;
+                Some((clock, TimeoutAnchor::Absolute, timeout))
+            }
+        };
+
+        // Perform validation of the arguments.
```
RalfJung, line 118: Are all of these explicitly documented?

src/shims/unix/macos/sync.rs
```diff
@@ -54,6 +76,191 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                size: Cell::new(size),
+                shared: Cell::new(is_shared),
+            }
+        }) else {
+            this.set_last_error_and_return(LibcError("ENOMEM"), dest)?;
+            return interp_ok(());
+        };
+
+        // Detect mismatches between the flags and sizes used on this address
+        // by storing the parameters of the first waiter in a batch of waiters.
```
RalfJung, line 156:
```suggestion
        // by comparing with the parameters stored with the first waiter.
```

src/shims/unix/macos/sync.rs
```diff
@@ -54,6 +76,191 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        if futex.futex.waiters() == 0 {
+            futex.size.set(size);
+            futex.shared.set(is_shared);
+        } else if futex.size.get() != size || futex.shared.get() != is_shared {
+            this.set_last_error_and_return(LibcError("EINVAL"), dest)?;
```
RalfJung, line 161: So if I used a location with a given size, I can never ever again use it with another size? That seems odd?

src/shims/unix/macos/sync.rs
```diff
@@ -54,6 +76,191 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            return interp_ok(());
+        }
+
+        if futex_val == value as u128 {
```
RalfJung, line 165:
```suggestion
        if futex_val == value.into() {
```

src/shims/unix/macos/sync.rs
```diff
@@ -54,6 +76,191 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                callback!(
+                    @capture<'tcx> {
+                        dest: MPlaceTy<'tcx>,
+                        futex_ref: FutexRef,
+                    }
+                    |this, unblock: UnblockKind| {
+                        match unblock {
+                            UnblockKind::Ready => {
+                                let remaining = futex_ref.waiters();
+                                this.write_scalar(Scalar::from_i32(remaining as i32), &dest)
```
RalfJung, line 182: Please avoid `as` casts.

src/shims/unix/macos/sync.rs
```diff
@@ -54,6 +76,191 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let is_shared = flags == shared;
+
+        let Some(futex) = this.get_sync_or_init(ptr, |_| {
+            MacOsFutex {
+                futex: Default::default(),
+                size: Cell::new(size),
+                shared: Cell::new(is_shared),
+            }
+        }) else {
+            this.set_last_error_and_return(LibcError("ENOENT"), dest)?;
```
RalfJung, line 236:
```suggestion
            // No AllocId, or no live allocation at that AllocId.
            // Return an error code. (That seems nicer than silently doing something non-intuitive.)
            // This means that if an address gets reused by a new allocation,
            // we'll use an independent futex queue for this... that seems acceptable.
            this.set_last_error_and_return(LibcError("ENOENT"), dest)?;
```

src/shims/unix/macos/sync.rs
```diff
@@ -54,6 +76,191 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            return interp_ok(());
+        };
+
+        if futex.futex.waiters() == 0 {
+            this.set_last_error_and_return(LibcError("ENOENT"), dest)?;
+            return interp_ok(());
+        // Detect mismatches between the flags and sizes used on this address.
+        } else if futex.size.get() != size || futex.shared.get() != is_shared {
```
RalfJung, line 244: Why is that in an `else` *after* checking whether there are any waiters?

tests/pass-dep/concurrency/apple-futex.rs
```diff
@@ -0,0 +1,211 @@
+//@only-target: darwin
+
+use std::time::{Duration, Instant};
+use std::{io, ptr, thread};
+
+fn wake_nobody() {
+    let futex = 0;
+
+    // Wake 1 waiter. Expect zero waiters woken up, as nobody is waiting.
```
RalfJung, line 9:
```suggestion
    // Wake 1 waiter. Expect ENOENT, as nobody is waiting.
```

tests/pass-dep/concurrency/apple-futex.rs
```diff
@@ -0,0 +1,211 @@
...
+    unsafe {
+        assert_eq!(
+            libc::os_sync_wake_by_address_any(
+                ptr,
+                size_of::<i32>(),
+                libc::OS_SYNC_WAKE_BY_ADDRESS_NONE
+            ),
+            -1
+        );
+        assert_eq!(io::Error::last_os_error().raw_os_error().unwrap(), libc::ENOENT);
```
RalfJung, line 38: So this is the same error as "nobody is waiting"? Interesting.

tests/pass-dep/concurrency/apple-futex.rs
```diff
@@ -0,0 +1,211 @@
...
+fn wait_absolute_timeout() {
+    let start = Instant::now();
+
+    // Get the current monotonic timestamp.
+    #[allow(deprecated)]
+    let mut deadline = unsafe { libc::mach_absolute_time() };
+
+    // Add 200ms (note: this only works under miri because it uses nanosecond units for
+    // mach_absolute_time).
+    deadline += 200_000_000;
```
RalfJung, line 92: This sounds like we are doing something wrong, then?

tests/pass-dep/concurrency/apple-futex.rs
```diff
@@ -0,0 +1,211 @@
...
+        );
+    }
+
+    // When running this in stress-gc mode, things can take quite long.
+    // So the timeout is 3000 ms.
+    assert!((200..3000).contains(&start.elapsed().as_millis()));
+    t.join().unwrap();
+}
+
+fn param_mismatch() {
```
RalfJung, line 152: Please add comments pointing out what the mismatch is in each case.

tests/pass-dep/concurrency/apple-futex.rs
```diff
@@ -0,0 +1,211 @@
...
+
+    unsafe {
+        assert_eq!(
+            libc::os_sync_wait_on_address(
+                (&raw const FUTEX).cast_mut().cast(),
+                0,
+                size_of::<i32>(),
+                libc::OS_SYNC_WAIT_ON_ADDRESS_NONE,
+            ),
+            0,
```
RalfJung, line 142: Please add a test where wait_on_address returns something non-zero (since there are more waiters).

joboet: While doing this, I noticed that the reported count was incorrect if `os_sync_wake_by_address_all` was used as the threads were woken one-by-one. dfa8d2fa13d2c9972208ab7dbc62e197a44f1920 fixes this (and also gets rid of the quadratic runtime of wakeup 😄).

src/shims/unix/macos/sync.rs
```diff
@@ -54,6 +76,191 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            .to_bits(Size::from_bytes(size))?;
+
+        let Some(futex) = this.get_sync_or_init(ptr, |_| {
+            MacOsFutex {
+                futex: Default::default(),
+                size: Cell::new(size),
+                shared: Cell::new(is_shared),
+            }
+        }) else {
+            this.set_last_error_and_return(LibcError("ENOMEM"), dest)?;
```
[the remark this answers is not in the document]
RalfJung: That affects all the other futex implementations as well, though. I'd prefer if they were consistent as there's nothing macOS-specific about this behavior, as far as I can tell.

src/shims/unix/macos/sync.rs
RalfJung: Ah, the `futex.futex.waiters() == 0` is quite important here. Please update the comment, it doesn't make this clear enough currently. It only really talks about the `futex.size.get() != size || futex.shared.get() != is_shared` but not the rest of this block.

src/shims/unix/macos/sync.rs
RalfJung: That's non-obvious, the comment should talk about that. (Same as in `os_sync_wait_on_address` where I was confused by the same thing.)

src/shims/unix/macos/sync.rs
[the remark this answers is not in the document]
joboet: Fair enough, I've changed it to `.unwrap()`

tests/pass-dep/concurrency/apple-futex.rs
RalfJung: Yeah but none of the other systems return an error when nobody is waiting, AFAIK.

tests/pass-dep/concurrency/apple-futex.rs
RalfJung: We already shim `mach_absolute_time` and `mach_timebase_info`.

RalfJung: But mostly I just found the comment insufficiently clear. Maybe something like:
```rust
// What we should be doing here is call `mach_timebase_info` to determine the
// unit used for `deadline`, but we know what Miri returns for that function:
// the unit is nanoseconds.
```

tests/pass-dep/concurrency/apple-futex.rs
joboet: True, but Apple explicitly documents `ENOENT` to be the error code returned in this case...

src/shims/unix/macos/sync.rs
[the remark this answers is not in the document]
RalfJung: Note that if this is truly immutable memory, the acquire read will already fail. So I am not even sure that case can happen.

tests/pass-dep/concurrency/apple-futex.rs
RalfJung: Yeah, it's fine. I was just somewhat surprised here that "no waiter" and "no memory" are the same error code, but I guess if there's no memory then there's definitely nobody waiting so this makes some sense.

src/shims/unix/macos/sync.rs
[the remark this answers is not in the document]
joboet: Ah, I see... yeah, unwrap is really fine then.

tests/pass-dep/concurrency/apple-futex.rs
```diff
@@ -6,7 +6,7 @@ use std::{io, ptr, thread};
 fn wake_nobody() {
     let futex = 0;
 
-    // Wake 1 waiter. Expect zero waiters woken up, as nobody is waiting.
+    // Wake 1 waiter. Expect ENOENT woken up, as nobody is waiting.
```
RalfJung, line 9: "ENOENT woken up" does not sound grammatically correct?

tests/pass-dep/concurrency/apple-futex.rs
```diff
@@ -149,6 +149,64 @@ fn wait_wake() {
     t.join().unwrap();
 }
 
+fn wait_wake_multiple() {
```
RalfJung, line 154: This looks like it could depend quite a bit on schedule non-determinism?

Might be good to make the entire test use `-Zmiri-preemption-rate=0` like we do for many concurrency tests; then `yield` can be used to explicitly control which thread gets to run when. That avoids having to sleep so much here as well.

src/shims/unix/macos/sync.rs
```diff
@@ -54,6 +76,191 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            || clock_timeout
+                .is_some_and(|(clock, _, timeout)| clock != absolute_clock || timeout == 0)
+        {
+            this.set_last_error_and_return(LibcError("EINVAL"), dest)?;
+            return interp_ok(());
+        }
+
+        let is_shared = flags == shared;
+        let timeout = clock_timeout.map(|(_, anchor, timeout)| {
+            (TimeoutClock::Monotonic, anchor, Duration::from_nanos(timeout))
```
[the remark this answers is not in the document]
RalfJung: Please add a comment making that connection.

tests/pass-dep/concurrency/apple-futex.rs
RalfJung: Looks good, thanks!

RalfJung:

@rustbot author

PR #4156, opened 2025-01-27
Apply random float error

### LINKED ISSUE #3555: Use non-deterministic precision for float operations that do not have guaranteed precision

LorrensP-2158466:
This partially tries to fix #3555  (does not include the `pow` operations)

this just does `apply_random_float_error` on the results of the intrinsics. 
Things changed from original:
- seperated the `sqrtf*` intrinsics to make the implementation cleaner, since applying the error was not required
- changes the way we do approximation checks, instead of using the epsilon technique, we use ULP. 
- create the relative errors of 16ULP using `ulp_err_scale`, which does `-(F::PRECISION - 1 - n` for a `2^n` ULP error.
- adds a test `test_non_determinism` which tests a lot of non-deterministic operations with the same input to check if they are almost but not entirely equal.

### DIFF
--- a/src/intrinsics/mod.rs
+++ b/src/intrinsics/mod.rs
@@ -254,7 +254,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let res = apply_random_float_error_ulp(
                     this,
                     res,
-                    4
+                    4, // log2(16)
                 );
                 let res = this.adjust_nan(res, &[f]);
                 this.write_scalar(res, dest)?;
@@ -289,7 +289,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let res = apply_random_float_error_ulp(
                     this,
                     res,
-                    4
+                    4, // log2(16)
                 );
                 let res = this.adjust_nan(res, &[f]);
                 this.write_scalar(res, dest)?;
@@ -411,7 +411,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // `binary_op` already called `generate_nan` if needed.
                 // Apply a relative error of 16ULP to simulate non-deterministic precision loss
                 // due to optimizations.
-                let res = apply_random_float_error_to_imm(this, res)?;
+                let res = apply_random_float_error_to_imm(this, res, 4 /* log2(16) */)?;
                 this.write_immediate(*res, dest)?;
             }
 
@@ -464,7 +464,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 }
                 // Apply a relative error of 16ULP to simulate non-deterministic precision loss
                 // due to optimizations.
-                let res = apply_random_float_error_to_imm(this, res)?;
+                let res = apply_random_float_error_to_imm(this, res, 4 /* log2(16) */)?;
                 this.write_immediate(*res, dest)?;
             }
 
@@ -503,13 +503,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 fn apply_random_float_error_to_imm<'tcx>(
     ecx: &mut MiriInterpCx<'tcx>,
     val: ImmTy<'tcx>,
+    ulp_exponent: u32,
 ) -> InterpResult<'tcx, ImmTy<'tcx>> {
     let scalar = val.to_scalar_int()?;
     let res: ScalarInt = match val.layout.ty.kind() {
-        ty::Float(FloatTy::F16) => apply_random_float_error_ulp(ecx, scalar.to_f16(), 4).into(),
-        ty::Float(FloatTy::F32) => apply_random_float_error_ulp(ecx, scalar.to_f32(), 4).into(),
-        ty::Float(FloatTy::F64) => apply_random_float_error_ulp(ecx, scalar.to_f64(), 4).into(),
-        ty::Float(FloatTy::F128) => apply_random_float_error_ulp(ecx, scalar.to_f128(), 4).into(),
+        ty::Float(FloatTy::F16) =>
+            apply_random_float_error_ulp(ecx, scalar.to_f16(), ulp_exponent).into(),
+        ty::Float(FloatTy::F32) =>
+            apply_random_float_error_ulp(ecx, scalar.to_f32(), ulp_exponent).into(),
+        ty::Float(FloatTy::F64) =>
+            apply_random_float_error_ulp(ecx, scalar.to_f64(), ulp_exponent).into(),
+        ty::Float(FloatTy::F128) =>
+            apply_random_float_error_ulp(ecx, scalar.to_f128(), ulp_exponent).into(),
         _ => bug!("intrinsic called with non-float input type"),
     };
 
--- a/src/intrinsics/mod.rs
+++ b/src/intrinsics/mod.rs
@@ -7,12 +7,13 @@ use rand::Rng;
 use rustc_abi::Size;
 use rustc_apfloat::{Float, Round};
 use rustc_middle::mir;
-use rustc_middle::ty::{self, FloatTy};
+use rustc_middle::ty::{self, FloatTy, ScalarInt};
 use rustc_span::{Symbol, sym};
 
 use self::atomic::EvalContextExt as _;
 use self::helpers::{ToHost, ToSoft, check_intrinsic_arg_count};
 use self::simd::EvalContextExt as _;
+use crate::math::apply_random_float_error_ulp;
 use crate::*;
 
 impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
@@ -206,10 +207,26 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.write_scalar(res, dest)?;
             }
 
+            "sqrtf32" => {
+                let [f] = check_intrinsic_arg_count(args)?;
+                let f = this.read_scalar(f)?.to_f32()?;
+                // Sqrt is specified to be fully precise.
+                let res = math::sqrt(f);
+                let res = this.adjust_nan(res, &[f]);
+                this.write_scalar(res, dest)?;
+            }
+            "sqrtf64" => {
+                let [f] = check_intrinsic_arg_count(args)?;
+                let f = this.read_scalar(f)?.to_f64()?;
+                // Sqrt is specified to be fully precise.
+                let res = math::sqrt(f);
+                let res = this.adjust_nan(res, &[f]);
+                this.write_scalar(res, dest)?;
+            }
+
             #[rustfmt::skip]
             | "sinf32"
             | "cosf32"
-            | "sqrtf32"
             | "expf32"
             | "exp2f32"
             | "logf32"
@@ -218,26 +235,33 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             => {
                 let [f] = check_intrinsic_arg_count(args)?;
                 let f = this.read_scalar(f)?.to_f32()?;
-                // Using host floats except for sqrt (but it's fine, these operations do not have
+                // Using host floats (but it's fine, these operations do not have
                 // guaranteed precision).
+                let host = f.to_host();
                 let res = match intrinsic_name {
-                    "sinf32" => f.to_host().sin().to_soft(),
-                    "cosf32" => f.to_host().cos().to_soft(),
-                    "sqrtf32" => math::sqrt(f),
-                    "expf32" => f.to_host().exp().to_soft(),
-                    "exp2f32" => f.to_host().exp2().to_soft(),
-                    "logf32" => f.to_host().ln().to_soft(),
-                    "log10f32" => f.to_host().log10().to_soft(),
-                    "log2f32" => f.to_host().log2().to_soft(),
+                    "sinf32" => host.sin(),
+                    "cosf32" => host.cos(),
+                    "expf32" => host.exp(),
+                    "exp2f32" => host.exp2(),
+                    "logf32" => host.ln(),
+                    "log10f32" => host.log10(),
+                    "log2f32" => host.log2(),
                     _ => bug!(),
                 };
+                let res = res.to_soft();
+                // Apply a relative error of 16ULP to introduce some non-determinism
+                // simulating imprecise implementations and optimizations.
+                let res = apply_random_float_error_ulp(
+                    this,
+                    res,
+                    4
+                );
                 let res = this.adjust_nan(res, &[f]);
                 this.write_scalar(res, dest)?;
             }
             #[rustfmt::skip]
             | "sinf64"
             | "cosf64"
-            | "sqrtf64"
             | "expf64"
             | "exp2f64"
             | "logf64"
@@ -246,19 +270,27 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             => {
                 let [f] = check_intrinsic_arg_count(args)?;
                 let f = this.read_scalar(f)?.to_f64()?;
-                // Using host floats except for sqrt (but it's fine, these operations do not have
+                // Using host floats (but it's fine, these operations do not have
                 // guaranteed precision).
+                let host = f.to_host();
                 let res = match intrinsic_name {
-                    "sinf64" => f.to_host().sin().to_soft(),
-                    "cosf64" => f.to_host().cos().to_soft(),
-                    "sqrtf64" => math::sqrt(f),
-                    "expf64" => f.to_host().exp().to_soft(),
-                    "exp2f64" => f.to_host().exp2().to_soft(),
-                    "logf64" => f.to_host().ln().to_soft(),
-                    "log10f64" => f.to_host().log10().to_soft(),
-                    "log2f64" => f.to_host().log2().to_soft(),
+                    "sinf64" => host.sin(),
+                    "cosf64" => host.cos(),
+                    "expf64" => host.exp(),
+                    "exp2f64" => host.exp2(),
+                    "logf64" => host.ln(),
+                    "log10f64" => host.log10(),
+                    "log2f64" => host.log2(),
                     _ => bug!(),
                 };
+                let res = res.to_soft();
+                // Apply a relative error of 16ULP to introduce some non-determinism
+                // simulating imprecise implementations and optimizations.
+                let res = apply_random_float_error_ulp(
+                    this,
+                    res,
+                    4
+                );
                 let res = this.adjust_nan(res, &[f]);
                 this.write_scalar(res, dest)?;
             }
@@ -316,6 +348,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             }
 
             "powf32" => {
+                // FIXME: apply random relative error but without altering behaviour of powf
                 let [f1, f2] = check_intrinsic_arg_count(args)?;
                 let f1 = this.read_scalar(f1)?.to_f32()?;
                 let f2 = this.read_scalar(f2)?.to_f32()?;
@@ -325,6 +358,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.write_scalar(res, dest)?;
             }
             "powf64" => {
+                // FIXME: apply random relative error but without altering behaviour of powf
                 let [f1, f2] = check_intrinsic_arg_count(args)?;
                 let f1 = this.read_scalar(f1)?.to_f64()?;
                 let f2 = this.read_scalar(f2)?.to_f64()?;
@@ -335,6 +369,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             }
 
             "powif32" => {
+                // FIXME: apply random relative error but without altering behaviour of powi
                 let [f, i] = check_intrinsic_arg_count(args)?;
                 let f = this.read_scalar(f)?.to_f32()?;
                 let i = this.read_scalar(i)?.to_i32()?;
@@ -344,6 +379,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.write_scalar(res, dest)?;
             }
             "powif64" => {
+                // FIXME: apply random relative error but without altering behaviour of powi
                 let [f, i] = check_intrinsic_arg_count(args)?;
                 let f = this.read_scalar(f)?.to_f64()?;
                 let i = this.read_scalar(i)?.to_i32()?;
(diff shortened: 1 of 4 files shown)


RalfJung:
Please also add a test that checks for non-determinism by running the same operation multiple times and ensuring we get different (but similar) results.

src/intrinsics/mod.rs
```diff
@@ -202,14 +212,33 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     _ => bug!(),
                 };
                 let res = f.round_to_integral(mode).value;
+                // Apply a relative error with a magnitude on the order of 2^-12 to simulate
+                // non-deterministic behaviour of floats
```
RalfJung, line 216: Rounding is fully precise, it should not get any error.

src/intrinsics/mod.rs
```diff
@@ -202,14 +212,33 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let res = apply_random_float_error(this, res, -12);
+                let res = this.adjust_nan(res, &[f]);
+                this.write_scalar(res, dest)?;
+            }
+
+            "sqrtf32" => {
+                let [f] = check_arg_count(args)?;
+                let f = this.read_scalar(f)?.to_f32()?;
+                // no host floats for sqrt
```
RalfJung, line 225:
```suggestion
                // Sqrt is specified to be fully precise.
```
(same below for sqrtf64)

src/intrinsics/mod.rs
LorrensP-2158466: Yes, I caught that and fixed it locally

LorrensP-2158466:
Almost done I think.

I have 1 TODO left, inside of the non-determinism test, it is about this:
> Please also add a test (as a new function in tests/pass/float.rs) that checks for non-determinism by running the same operation multiple times and ensuring we get different (but similar) results.

Should we do an `assert_ne!` test on the results, because it is possible that the same error is applied to 2 separate operations that we then check. Maybe it is possible to count these exactly equal results and if it is the same as the amount of checks we can safely say it is deterministic and should fail the test?.

RalfJung:
> Should we do an assert_ne! test on the results, because it is possible that the same error is applied to 2 separate operations that we then check. Maybe it is possible to count these exactly equal results and if it is the same as the amount of checks we can safely say it is deterministic and should fail the test?.

I think the easiest way is something like [this](https://github.com/rust-lang/miri/blob/fa05151e0567883f4d72abeaf8165850f6a4ffaf/tests/pass/float.rs#L1220-L1231); you can do that with a closure returning any `PartialEq` type instead of `bool`.

src/math.rs
```diff
@@ -2,6 +2,16 @@ use rand::Rng as _;
 use rustc_apfloat::Float as _;
 use rustc_apfloat::ieee::IeeeFloat;
 
+/// Creates an error scale to pass to `apply_random_float_error` when applying a 2^n ULP error.
+/// For example, if you want to apply a 16 ULP error, you should pass 4, because 2^4 = 16.
+pub(crate) fn ulp_err_scale<F: rustc_apfloat::Float>(n: u32) -> i32 {
```
RalfJung, line 7: Please move this below `apply_random_float_error`.

Also, `ulp_err_scale(4)` looks like a scale of 4 ULP, but it's really 2^4. Any ideas for a better name that makes this more clear?

src/intrinsics/mod.rs
```diff
@@ -218,26 +235,31 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    "logf32" => host.ln(),
+                    "log10f32" => host.log10(),
+                    "log2f32" => host.log2(),
                     _ => bug!(),
                 };
+                // Apply a relative error of 16ULP to simulate non-determinism
+                let res = apply_random_float_error(
+                    this,
+                    res.to_soft(),
+                    ulp_err_scale::<rustc_apfloat::ieee::Single>(4)
```
RalfJung, line 255: Having to give the type again here seems error prone.

What about having a `apply_random_float_error_ulp` function that can use the type of `val` to ensure that it uses the right scale?

src/intrinsics/mod.rs
```diff
@@ -218,26 +235,31 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                    "log2f32" => f.to_host().log2().to_soft(),
+                    "sinf32" => host.sin(),
+                    "cosf32" => host.cos(),
+                    "expf32" => host.exp(),
+                    "exp2f32" => host.exp2(),
+                    "logf32" => host.ln(),
+                    "log10f32" => host.log10(),
+                    "log2f32" => host.log2(),
                     _ => bug!(),
                 };
```
RalfJung, line 250:
```suggestion
                };
                let res = res.to_soft();
```
I'd like to have the soft-hard-soft roundtrip separated from the post-processing.

src/intrinsics/mod.rs
```diff
@@ -218,26 +235,31 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                // Apply a relative error of 16ULP to simulate non-determinism
```
RalfJung, line 251:
```suggestion
                // Apply a relative error of 16ULP to introduce some non-determinism
                // simulating imprecise implementations and optimizations.
```
This doesn't "simulate non-determinism", it *is* non-determinism. So let's be a bit more clear. Also please end sentences in `.`.

The comment occurs a bunch of times, please replace this everywhere (except where I made more specific suggestions).

src/intrinsics/mod.rs
```diff
@@ -372,8 +399,28 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     _ => bug!(),
                 };
                 let res = this.binary_op(op, &a, &b)?;
-                // `binary_op` already called `generate_nan` if necessary.
-                this.write_immediate(*res, dest)?;
+                // `binary_op` already called `generate_nan` if needed
+                // Apply a relative error of 16ULP to simulate non-determinism
```
RalfJung, line 403:
```suggestion
                // `binary_op` already called `generate_nan` if needed.
                // Apply a relative error of 16ULP to simulate non-deterministic precision loss
                // due to optimizations.
```

src/intrinsics/mod.rs
```diff
@@ -372,8 +399,28 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                fn apply_error_and_write<'tcx, F: Float + Into<Scalar> >(
+                    ecx: &mut MiriInterpCx<'tcx>,
+                    res: F,
+                    dest: &MPlaceTy<'tcx>
+                ) -> InterpResult<'tcx> {
```
RalfJung, line 408: Please add a helper function that does `apply_random_float_error_ulp` on an `ImmTy`, so that the `match` can be shared between here and the `*_fast` intrinsics below. Keep the `write_scalar` separate.

src/intrinsics/mod.rs
```diff
@@ -421,9 +468,29 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 if !float_finite(&res)? {
                     throw_ub_format!("`{intrinsic_name}` intrinsic produced non-finite value as result");
                 }
+                // Apply a relative error of 16ULP to simulate non-determinism
```
RalfJung, line 471:
```suggestion
                // Apply a relative error of 16ULP to simulate non-deterministic precision loss
                // due to optimizations.
```

tests/pass/float.rs
```diff
@@ -12,11 +12,37 @@ use std::fmt::{Debug, Display, LowerHex};
...
 macro_rules! assert_approx_eq {
-    ($a:expr, $b:expr) => {{
+    ($a:expr, $b:expr, $ulp:expr) => {{
         let (a, b) = (&$a, &$b);
-        assert!((*a - *b).abs() < 1.0e-6, "{} is not approximately equal to {}", *a, *b);
+        let a_signum = a.signum();
+        let b_signum = b.signum();
+
+        if a_signum != b_signum {
+            // checking -0.0 == +0.0
```
RalfJung, line 25:
```suggestion
            // Floats with different signs must both be 0.
```

tests/pass/float.rs
```diff
@@ -12,11 +12,37 @@ use std::fmt::{Debug, Display, LowerHex};
...
+            assert!(a == b, "{:?} is not approximately equal to {:?}: signs differ", *a, *b);
+        }
+        let a_bits = a.to_bits() as i128;
+        let b_bits = b.to_bits() as i128;
+        let allowed_ulp = $ulp;
+        let ulp_difference = (a_bits - b_bits).abs();
```
RalfJung, line 31: Please use abs_diff

tests/pass/float.rs
```diff
@@ -12,11 +12,37 @@ use std::fmt::{Debug, Display, LowerHex};
...
         let (a, b) = (&$a, &$b);
-        assert!((*a - *b).abs() < 1.0e-6, "{} is not approximately equal to {}", *a, *b);
+        let a_signum = a.signum();
+        let b_signum = b.signum();
+
+        if a_signum != b_signum {
+            // checking -0.0 == +0.0
+            assert!(a == b, "{:?} is not approximately equal to {:?}: signs differ", *a, *b);
+        }
+        let a_bits = a.to_bits() as i128;
```
RalfJung, line 28: Shouldn't this be in an `else` block?

tests/pass/float.rs
```diff
@@ -12,11 +12,37 @@ use std::fmt::{Debug, Display, LowerHex};
...
+        let b_bits = b.to_bits() as i128;
```
RalfJung, line 29: Please add comments explaining what happens here to people not familiar with the binary representation of floats.

tests/pass/float.rs
```diff
@@ -12,11 +12,37 @@ use std::fmt::{Debug, Display, LowerHex};
...
+        let allowed_ulp = $ulp;
+        let ulp_difference = (a_bits - b_bits).abs();
+        assert!(
+            ulp_difference <= allowed_ulp,
+            "
+            {:?} is not approximately equal to {:?}
+            ulp diff: {ulp_difference} > {allowed_ulp}
+            ",
+            *a,
+            *b
```
RalfJung, line 39: This will print complete nonsense when the exponent is different, won't it?

I'm not convinced going via the hex representation is the best option here.

tests/pass/float.rs
```diff
@@ -1004,8 +1031,8 @@ pub fn libm() {
 
     #[allow(deprecated)]
     {
-        assert_approx_eq!(5.0f32.abs_sub(3.0), 2.0);
-        assert_approx_eq!(3.0f64.abs_sub(5.0), 0.0);
+        assert_approx_eq!(5.0f32.abs_sub(3.0), 2.0f32);
+        assert_approx_eq!(3.0f64.abs_sub(5.0), 0.0f64);
```
RalfJung, line 1035: The macro shouldn't accept arguments of different type.

tests/pass/float.rs
```diff
@@ -1239,3 +1266,109 @@ fn test_min_max_nondet() {
...
     ensure_both(|| f128::max(0.0, -0.0).is_sign_positive());
 }
+
+fn test_non_determinism() {
+    use std::intrinsics::{
+        fadd_algebraic, fadd_fast, fdiv_algebraic, fdiv_fast, fmul_algebraic, fmul_fast,
+        frem_algebraic, frem_fast, fsub_algebraic, fsub_fast,
+    };
+    use std::{f32, f64};
+    // FIXME(powi_powf): add when supported
```
RalfJung, line 1276: what is that about?

tests/pass/float.rs
```diff
@@ -1239,3 +1266,109 @@ fn test_min_max_nondet() {
...
+
+    /// Ensure that the operation is non-deterministic
+    #[track_caller]
+    fn ensure<T: PartialEq + std::fmt::Debug>(f: impl Fn() -> T) {
```
RalfJung, line 1280:
```suggestion
    fn ensure_nondet<T: PartialEq + std::fmt::Debug>(f: impl Fn() -> T) {
```

tests/pass/float.rs
```diff
@@ -1239,3 +1266,109 @@ fn test_min_max_nondet() {
...
+        };
+    }
+
+    pub fn test_operations_f16(a: f16, b: f16) {
+        test_operations_f!(a, b);
+    }
+    pub fn test_operations_f32(a: f32, b: f32) {
+        test_operations_f!(a, b);
+        ensure(|| a.log(b));
+        ensure(|| a.exp());
```
RalfJung, line 1317: Why are some operations here and some in "extras"?

RalfJung (commented):
Thanks a lot, this looks great! All comments are minor.

@rustbot author

tests/pass/float.rs
LorrensP-2158466: you mentioned in the zulip chat to not include the `pow` intrinsics in this pr. I should explain this better inside the comment.

tests/pass/float.rs
LorrensP-2158466: You're right, I'll add them to f32 and f64

tests/pass/float.rs
LorrensP-2158466: `f16` and `f128` currently don't implement `Display`, so using debug was the only thing that worked (that i know of).

src/intrinsics/mod.rs
LorrensP-2158466: yeah i thought about it, should have asked, I'll do it!

src/math.rs
LorrensP-2158466: I'm thinking of using `ulp_exponent_err_scale` or maybe accepting the ULP as is and then doing `log2`?

tests/pass/float.rs
```diff
@@ -12,11 +12,68 @@ use std::fmt::{Debug, Display, LowerHex};
...
+/// Same with exponents but no zero checking.
+///
+/// So when Sign and Exponent are the same, we can get a reasonable ULP value
+/// by doing the operation explained above. And if this is less than or equal to our chosen upper bound
+/// we can say that `a` and `b` are approximately equal.
+///
+/// Basically ULP can be seen as a distance metric of floating point numbers, but having
+/// the same amount of "spacing" between consecutive numbers. So eventhough 2 very large floating point numbers
+/// have a large value difference, their ULP can still be 1, so they are still "approximatly equal",
+/// but the EPSILON check would have failed.
```
LorrensP-2158466, line 42: I only learned about this concept last week, but I tried my best. Let me know if i missed something or if something is explained badly.

tests/pass/float.rs
LorrensP-2158466: It doesn't. I don't understand what you meant here

LorrensP-2158466:
Fixed these CI checks in the last PR but CI never ran on that one...
@rustbot ready

LorrensP-2158466:
@rustbot blocked

I don't know whats going on

saethlin:
CI tests your branch merged with the target branch. The errors in CI look like you have a semantic merge conflict (GitHub can only automatically detect textual merge conflicts).

You should rebase onto the upstream target branch. The most foolproof way to do this is a rebase pull: `git pull --rebase git@github.com:rust-lang/miri`.

tests/pass/float.rs
RalfJung: Ah I see. The better place for the FIXME would be in the shim implementation.

tests/pass/float.rs
RalfJung: Oh wow that's a lot more detailed than I expected, thanks!

src/math.rs
RalfJung: The function will disappear anyway due to `apply_random_float_error_ulp`, so this becomes part of that bikeshed. Maybe `apply_random_float_error_ulp_exp`? OTOH `apply_random_float_error` is also exponential so then we should rename both. So maybe just `apply_random_float_error_ulp` is fine.

tests/pass/float.rs
RalfJung: I don't understand your reply.

If the exponent is different, subtracting the two `i128` representations will not be the difference in ULP, it will be complete junk.

RalfJung: So here's an alternative way to approximate the distance in ULP, but it may fail spectacularly due to rounding:
```rust
let ulp = (expected.next_up() - expected.next_down()) / 2;
let ulp_diff = ((real - expected) / ulp).round();
```

src/intrinsics/mod.rs
```diff
@@ -206,10 +207,26 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.write_scalar(res, dest)?;
             }
 
+            "sqrtf32" => {
+                let [f] = check_arg_count(args)?;
```
RalfJung, line 211: This function does not exist any more, it is now called `check_intrinsic_arg_count`

tests/pass/float.rs
RalfJung: Sorry, I was not very clear: I am surprised that you need to add type annotations for the 2nd argument here. This should be forced by type inference.

saethlin:
If you still have a clean local build, run `./miri toolchain`, which will probably update the toolchain override.

tests/pass/float.rs
LorrensP-2158466: It then shows this error:
```
error[E0599]: no method named `signum` found for reference `&{float}` in the current scope
  --> tests/pass/float.rs:48:26
   |
LL |         let b_signum = b.signum();
   |                          ^^^^^^ method not found in `&{float}`
...
LL |     assert_approx_eq!(25f32.powi(-2), 0.0016);
   |     ----------------------------------------- in this macro invocation
   |
   = note: this error originates in the macro `assert_approx_eq` (in Nightly builds, run with -Z macro-backtrace for more info)

```

RalfJung: Yeah that's what I said: the macro does not constrain the types of the two arguments to be the same, and therefore you can this error as it cannot figure out the type of the literal on the RHS.

LorrensP-2158466: I'll try something to make this more intuitive to work with.

tests/pass/float.rs
LorrensP-2158466: I have to be honest, I don't see why we would use this way of calculating ULP. Isn't it unlikely to get approximately equal values with different exponents in our tests?

RalfJung: It's not a question of probability -- if we happen to write a test where the result is near the boundary between values represented with different exponents, we'll suddenly get completely nonsensical results. That's fragile and hard to debug. (Maybe that explains what is going on with `pow`.)  We shouldn't have surprising footguns in our code that can go off any day in the future that someone adds a new test.

tests/pass/float.rs
RalfJung: I think this will be resolved by changing the macro back to using FP arithmetic rather than going via the raw binary representation.

src/intrinsics/mod.rs
```diff
@@ -464,33 +452,16 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     ),
                     _ => {}
                 }
+                // This cannot be a NaN so we also don't have to apply any non-determinism.
+                // (Also, `binary_op` already called `generate_nan` if needed.)
```
RalfJung, line 460: This comment also got moved up incorrectly.

Please always make sure the comment logically talks about the point in the code where it sits. If the comment refers to something in past tense that will happen further down the code, that makes no sense.

src/math.rs
```diff
@@ -37,6 +27,27 @@ pub(crate) fn apply_random_float_error<F: rustc_apfloat::Float>(
...
+    ecx: &mut crate::MiriInterpCx<'_>,
+    val: F,
+    ulp_exponent: u32,
+) -> F {
+    apply_random_float_error(ecx, val, ulp_exponent_err_scale::<F>(ulp_exponent))
+}
+
+/// Creates an error scale to pass to `apply_random_float_error` when applying a 2^n ULP error.
+/// For example, if you want to apply a 16 ULP error, you should pass 4, because 2^4 = 16.
+pub(crate) fn ulp_exponent_err_scale<F: rustc_apfloat::Float>(n: u32) -> i32 {
```
RalfJung, line 43: This function is now used only once, please inline it at the call site.

src/intrinsics/mod.rs
```diff
@@ -522,3 +493,22 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
 }
+
+/// Applies a random 16ULP floating point error to `val` and returns the new value.
+/// Will fail if `val` is not a floating point number.
+fn apply_random_float_error_to_imm<'tcx>(
+    ecx: &mut MiriInterpCx<'tcx>,
+    val: ImmTy<'tcx>,
+) -> InterpResult<'tcx, ImmTy<'tcx>> {
+    let scalar = val.to_scalar_int()?;
+    let layout = val.layout;
```
RalfJung, line 508: Please remove this let-binding, it's not worth it. (And you are not even using it in the very next line...)

LorrensP-2158466:
Sorry for the delay, Uni started this week.

I extended the explanation of the Approx Check, let me know if this is correct and not too long?

@rustbot ready

tests/pass/float.rs
```diff
@@ -12,11 +12,104 @@ use std::fmt::{Debug, Display, LowerHex};
...
+    } else {
+        let ulp = (expected.next_up() - expected.next_down()).halve();
+        let ulp_diff = ((actual - expected) / ulp).round().as_int();
+
+        if ulp_diff > allowed_ulp {
+            return Err(NotApproxEq::UlpFail(ulp_diff));
+        }
+    }
+    Ok(())
+}
```
RalfJung, line 86: Now that this is just regular float arithmetic, why do we still need this trait and all the rest?

Also, most of the explanation is not relevant any more since we are not working on the bit representation any more.

LorrensP-2158466: Trait is still needed for other helper functions I didn't write, they were also used in macros and such so i followed that approach. It's still a lot of code and the macro will thus generate a lot of code (not sure if this matters), and there is no way to make a float generic function other than using the trait.

I'll update the explanation, let me know what you think about the rest.

LorrensP-2158466: Extra: Also the weird extra functions (halve, as_int, round) i had to add to the Trait were necessary to make the generic function work. Just putting in `as F::int` or ` / 2.0` did not work in any clean way other than this. It's no problem if i have to remove it if needed.

RalfJung:
I found it easier to just do the remaining changes than to try to explain them, so I have updated the PR accordingly. I also rebased over the latest rustup which adds one more function that had to be tested for non-determinism, and squashed the commits. (If you want to see what I changed compared to your version, look at the 2nd commit.)

Turns out a macro-based solution still works just fine. Maybe a method is cleaner, but switching to that should not be done in the same PR as adding new functionality.

Thanks for your work!

LorrensP-2158466:
Alright that's good. Thanks, learned a lot!

src/intrinsics/mod.rs
```diff
@@ -372,7 +408,10 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     _ => bug!(),
                 };
                 let res = this.binary_op(op, &a, &b)?;
-                // `binary_op` already called `generate_nan` if necessary.
+                // `binary_op` already called `generate_nan` if needed.
+                // Apply a relative error of 16ULP to simulate non-deterministic precision loss
+                // due to optimizations.
+                let res = apply_random_float_error_to_imm(this, res, 4 /* log2(16) */)?;
```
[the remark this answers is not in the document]
RalfJung: Note that the docs even call this out:

> Because of the unpredictable nature of compiler optimizations, the same inputs may produce different results even within a single program run. Unsafe code must not rely on any property of the return value for soundness.

PR #4159, opened 2025-01-28
Set st_fstype of stat on Solaris and Illumos OSes

YohDeadfall:
Since https://github.com/rust-lang/libc/pull/4145 is merged that FIXME is unblocked. I also decided to rename the method as it applies to Illumos too, so `solarish` seems more natural there.

### DIFF
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -273,7 +273,7 @@ impl UnixFileDescription for FileHandle {
 
 impl<'tcx> EvalContextExtPrivate<'tcx> for crate::MiriInterpCx<'tcx> {}
 trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
-    fn macos_fbsd_solaris_write_buf(
+    fn macos_fbsd_solarish_write_stat_buf(
         &mut self,
         metadata: FileMetadata,
         buf_op: &OpTy<'tcx>,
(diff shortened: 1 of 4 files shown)


src/shims/unix/fs.rs
```diff
@@ -273,7 +273,7 @@ impl UnixFileDescription for FileHandle {
 
 impl<'tcx> EvalContextExtPrivate<'tcx> for crate::MiriInterpCx<'tcx> {}
 trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
-    fn macos_fbsd_solaris_write_buf(
+    fn macos_fbsd_solarish_write_buf(
```
RalfJung, line 276: If we're renaming anyway, what about `write_stat` or so? "buf" is very non-descriptive.

RalfJung:
This will require the libc version in rust'c `library/Cargo.lock` to be updated (and probably also in Miri's `test_dependencies/Cargo.lock`).

RalfJung:
Actually never mind, the error is different than what I expected:
```
write_int on non-scalar type [i8; 16]
```

And indeed, looking at libc this is an array.

src/shims/unix/fs.rs
```diff
@@ -321,7 +321,8 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
         }
 
         if matches!(&*this.tcx.sess.target.os, "solaris" | "illumos") {
-            this.write_int_fields_named(&[("st_fstype", 0)], &buf)?;
+            let st_fstype = this.project_field_named(&buf, "st_fstype")?;
```
RalfJung, line 324:
```suggestion
            let st_fstype = this.project_field_named(&buf, "st_fstype")?;
            // This is an array; write 0 into first element so that it encodes the empty string.
```

RalfJung:

@rustbot author

PR #4167, opened 2025-01-31
clarify apply_random_float_error logic

RalfJung:
@eduardosm does this make sense?

### DIFF
--- a/src/math.rs
+++ b/src/math.rs
@@ -3,7 +3,11 @@ use rand::distributions::Distribution as _;
 use rustc_apfloat::Float as _;
 use rustc_apfloat::ieee::IeeeFloat;
 
-/// Disturbes a floating-point result by a relative error on the order of (-2^scale, 2^scale).
+/// Disturbes a floating-point result by a relative error in the range (-2^scale, 2^scale).
+///
+/// For a 2^N ULP error, you can use an `err_scale` of `-(F::PRECISION - 1 - N)`.
+/// In other words, a 1 ULP (absolute) error is the same as a `2^-(F::PRECISION-1)` relative error.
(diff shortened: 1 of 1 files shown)


src/math.rs
```diff
@@ -4,17 +4,23 @@ use rustc_apfloat::Float as _;
...
 pub(crate) fn apply_random_float_error<F: rustc_apfloat::Float>(
     ecx: &mut crate::MiriInterpCx<'_>,
     val: F,
     err_scale: i32,
 ) -> F {
     let rng = ecx.machine.rng.get_mut();
     // Generate a random integer in the range [0, 2^PREC).
+    // (This is one more bit of data than the mantissa can actually hold as `PRECISION` includes
+    // the implicit leading 1, but there is no harm in that.)
     let dist = rand::distributions::Uniform::new(0, 1 << F::PRECISION);
```
eduardosm, line 19: It's true that `PRECISION` includes the implicit bit, but it sounds like one bit will be rounded away, which is not the case. The highest `1` bit will become the implicit bit, and the rest (at most `PRECISION - 1` bits) will be part of the explicit mantissa.

RalfJung: So you are saying we actually need 24bits of entropy to generate a uniform random number in `(0, 1]`? Hm... :thinking:

RalfJung:
Thanks for taking a look. :)

src/math.rs
RalfJung: Okay, thanks. I assume we do lose some parts of that bit since things get rounded, so the true entropy is somewhere strictly below 24, but it is strictly bigger than 23 so we need to ask `rand` for 23 bits.

Issue #4149, opened 2025-01-24
Remove support for miri_get_backtrace with flags == 0


RalfJung:
The main user here is the backtrace crate and it has used the v1 protocol for a while now, so we can remove support for the old protocol.

tiif:
I'd probably get to this soon, but in case someone wants to pick it up, there are more instructions here. https://github.com/rust-lang/miri/pull/4122#issuecomment-2600950358

@rustbot label +C-cleanup +C-enhancement

RalfJung:
https://github.com/rust-lang/miri/pull/4172 takes care of this

PR #4172, opened 2025-02-02
miri_get_backtrace: stop supporting the v0 protocol

RalfJung:
We did the same for miri_resolve_frame in https://github.com/rust-lang/rust/pull/134055. This lets us get rid of the `check_min_arg_count` helper.

Fixes https://github.com/rust-lang/miri/issues/4149


RalfJung:
Taking @saethlin's :+1: as approval

PR #4177, opened 2025-02-04
Use fcntl locking on Solaris instead of flock which is missing there.

psumbera:


RalfJung:
Thanks for the PR!
Could you explain a bit more the motivation for this change and what it does? The code will also definitely need some comments.

This seems to be about what happens when building Miri on a Solaris *host*. That's an entirely untested configuration and not something we can officially support, or even test on CI. I don't want to have Solaris-specific code all over the codebase either, lest it gets in the way of the maintaining the code for our targets that we can provide support for. So another option here would be to just say that `flock` is not supported on a Solaris host -- if the issue is "Miri does not build", that would IMO be preferred.

psumbera:
I have run into this problem (missing `flock()` and `LOCK_*`) while I was cross compiling Rust on Linux for Solaris using Rust Docker infrastructure. My ultimate goal is to allow regular builds for Solaris as it's done for example for Illumos (Solaris derivative). With this change the full Rust build passed (including with `miri-nightly-x86_64-pc-solaris.tar.xz`).

It's true that I also tried to build Miri natively and there were other problems too. E.g.

```
   Compiling miri v0.1.0 (/export/home/test/miri)
error[E0463]: can't find crate for `either`
  --> src/lib.rs:59:1
   |
59 | extern crate either;
   | ^^^^^^^^^^^^^^^^^^^^ can't find crate
```
Not sure whether it was the same version though. I used latest `git` sources.

There were no differences for `flock` fuction so I think that my testing from Docker build remains valid.

I can eventually add CI for Solaris using vm-actions as I did already for several projects (libc quinn, nix). But I would need to resovle build issues first.

Note sure what can be done better with commenting. It seems to be similar to what is there for Unix flock locking. `fcntl` locking is another way of file locking. And it's portable. In fact it should be available on more platforms than `flock()` is.

RalfJung:
Yeah we're not well-equipped to maintain the original `flock`-based code either, I don't really understand it.^^

If the primary goal for now is just to make the build work, I would suggest to
- change the code to use `cfg_match` so that we can more easily have an if-then-else chain
- fix the logic for the `flock` case -- judging from the libc crate, solaris is the sole exception here, so `#[cfg(all(target_family = "unix", not(target_os = "solaris")))]` sounds right
- change the fallback branch from a `compile_error!` to a `throw_unsup_format!`

What currently confuses me is that the `tests/pass-dep/libc/libc-fs-flock.rs` test apparently passes on Solaris which indicates that the `flock` symbol is available there...

RalfJung:
> What currently confuses me is that the tests/pass-dep/libc/libc-fs-flock.rs test apparently passes on Solaris which indicates that the flock symbol is available there...

Ah, I guess https://github.com/rust-lang/libc/commit/251e8e839a22e3d3140091f780e5123d9e8aac15 hasn't made it into the libc we use yet. So that test will soon start failing.

So then, the `flock` shim [here](https://github.com/rust-lang/miri/blob/9611e78d8a9ba3eae6b3741edccf2c2d63e4eb42/src/shims/unix/foreign_items.rs#L226-L232) should also fail on Solaris, and the `tests/pass-dep/libc/libc-fs-flock.rs` test should gain an `ignore-solaris` directive.

psumbera:
I think I'm confused here.

>     * change the code to use `cfg_match` so that we can more easily have an if-then-else chain

Shall I do something like?
```
    fn flock<'tcx>(
        &self,
        communicate_allowed: bool,
        op: FlockOp,
        assert!(communicate_allowed, "isolation should have prevented even opening a file");
        cfg_if! {
        if #[cfg(all(target_family = "unix", not(target_os = "solaris")))] {
 ```
Do I need to move the code for one more indentation then?
  
>     * fix the logic for the `flock` case -- judging from the libc crate, solaris is the sole exception here, so `#[cfg(all(target_family = "unix", not(target_os = "solaris")))]` sounds right

So, is it right or should I fix something?

>     * change the fallback branch from a `compile_error!` to a `throw_unsup_format!`

It would be:
```
        } else {
            let _ = op;
            throw_unsup_format!("flock is supported only on UNIX and Windows hosts");
        }
        }
```
And Solaris fcntl() code should remain there, right?

psumbera:
Now I see used `cfg_if` instead of `cfg_match`. Should I change it?

RalfJung:
I'd prefer not adding a new dependency for something that exists (albeit unstably) in the sysroot. So yeah please change it.

RalfJung:
I guess this also works without cfg_match. ;)  Please fix the rustfmt issue; I might try my hand at using cfg_match or just land the PR as-is.

psumbera:
Can we use this minimal version?

`cfg_match` seems to be too much complicated. The issue here is that `cfg_match` expects items (not statement). So another indent would be required?!

RalfJung:
I got it to work with expressions.

PR #4179, opened 2025-02-06
allow code to call geteuid()

rayslava:
Just return the effective UID like it's done for the real UID.

This fixes the `whoami` crate usage.


RalfJung:
That makes sense, thanks!

Issue #3842, opened 2024-08-24
Shims ICE when a scalar argument isn't actually a scalar


RalfJung:
For instance, consider:
```rust
extern "C" {
    fn pipe(pipefd: [i32; 2]) -> i32;
}

fn main() {
    let mut fds: [i32; 2] = [0; 2]; 
    assert_eq!(unsafe { pipe(fds) }, 0); 
}
```
I think to fix this properly we need to completely re-do the way we handle shim arguments: every `OpTy` needs to be transmuted to the right type before we do anything else with it. (And we can't even just call `OpTy::transmute` as that will ICE if the transmute would change the size. Though maybe we should change that.)

We probably want a macro that helps with that.

RalfJung:
I think we want a new function like `get_arg` or so that preprocesses the `OpTy`. Given an `op: OpTy` and `ty: Ty`, we do a check like
- `let layout = layout_of(ty);`
- Check `layout.abi.eq_up_to_validity(&op.layout.abi)`. If that is `false`, the argument type is invalid -> raise UB.
- Return `op.transmute(layout)`.

RalfJung:
Similar ICEs can occur when the return type of the function is wrong -- we end up calling `write_scalar` with a scalar of the wrong size. This either fails to initialize a part of the result (if the scalar is smaller than the return place), or ICEs since we are exceeding the bounds of our `AllocRefMut` (if the scalar is bigger than the return place).

geetanshjuneja:
Can I look into this task?

RalfJung:
Definitely! However, this will require some design work. As mentioned in the issue description, we don't have a clear plan yet for how to best fix this issue.

geetanshjuneja:
what does this fn `eq_up_to_validity` do?  I couldn't find its definition anywhere?

tiif:
Quick answer: It is part of the rustc https://github.com/rust-lang/rust/blob/64e06c0f5578829373743884b708d494136c3e8f/compiler/rustc_abi/src/lib.rs#L1517.

RalfJung:
Please use [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/269128-miri) for basic questions about how Miri works; the issue should be for discussion specific to the task described above. :)

RalfJung:
[...]

PR #4185, opened 2025-02-09
FnAbi Compatability check 

geetanshjuneja:
Links to #3842

### DIFF
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -14,7 +14,7 @@ use rustc_middle::middle::codegen_fn_attrs::CodegenFnAttrFlags;
 use rustc_middle::middle::dependency_format::Linkage;
 use rustc_middle::middle::exported_symbols::ExportedSymbol;
 use rustc_middle::ty::layout::{FnAbiOf, LayoutOf, MaybeResult, TyAndLayout};
-use rustc_middle::ty::{self, FloatTy, IntTy, Ty, TyCtxt, UintTy};
+use rustc_middle::ty::{self, Binder, FloatTy, FnSig, IntTy, Ty, TyCtxt, UintTy};
 use rustc_session::config::CrateType;
 use rustc_span::{Span, Symbol};
 use rustc_target::callconv::{Conv, FnAbi};
@@ -994,10 +994,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         exp_abi: Conv,
         link_name: Symbol,
         args: &'a [OpTy<'tcx>],
-    ) -> InterpResult<'tcx, &'a [OpTy<'tcx>; N]>
-    where
-        &'a [OpTy<'tcx>; N]: TryFrom<&'a [OpTy<'tcx>]>,
-    {
+    ) -> InterpResult<'tcx, &'a [OpTy<'tcx>; N]> {
         self.check_abi_and_shim_symbol_clash(abi, exp_abi, link_name)?;
 
         if abi.c_variadic {
@@ -1013,6 +1010,80 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             args.len(),
             N
         )
+    }
+
+    /// Check that the given `caller_fn_abi` matches the expected ABI described by
+    /// `callee_abi`, `callee_input_tys`, `callee_output_ty`, and the return the list of
+    /// arguments.
+    fn check_shim_abi<'a, const N: usize>(
+        &mut self,
+        link_name: Symbol,
+        caller_fn_abi: &FnAbi<'tcx, Ty<'tcx>>,
+        callee_abi: ExternAbi,
+        callee_input_tys: [Ty<'tcx>; N],
+        callee_output_ty: Ty<'tcx>,
+        caller_args: &'a [OpTy<'tcx>],
+    ) -> InterpResult<'tcx, &'a [OpTy<'tcx>; N]> {
+        let this = self.eval_context_mut();
+        let mut inputs_and_output = callee_input_tys.to_vec();
+        inputs_and_output.push(callee_output_ty);
+        let fn_sig_binder = Binder::dummy(FnSig {
+            inputs_and_output: this.machine.tcx.mk_type_list(&inputs_and_output),
+            c_variadic: false,
+            // This does not matter for the ABI.
+            safety: Safety::Safe,
+            abi: callee_abi,
+        });
+        let callee_fn_abi = this.fn_abi_of_fn_ptr(fn_sig_binder, Default::default())?;
+
+        this.check_abi_and_shim_symbol_clash(caller_fn_abi, callee_fn_abi.conv, link_name)?;
+
+        if caller_fn_abi.c_variadic {
+            throw_ub_format!(
+                "ABI mismatch: calling a non-variadic function with a variadic caller-side signature"
+            );
+        }
+
+        if callee_fn_abi.fixed_count != caller_fn_abi.fixed_count {
+            throw_ub_format!(
+                "ABI mismatch: expected {} arguments, found {} arguments ",
+                callee_fn_abi.fixed_count,
+                caller_fn_abi.fixed_count
+            );
+        }
+
+        if callee_fn_abi.can_unwind && !caller_fn_abi.can_unwind {
+            throw_ub_format!(
+                "ABI mismatch: callee may unwind, but caller-side signature prohibits unwinding",
+            );
+        }
+
+        if !this.check_argument_compat(&caller_fn_abi.ret, &callee_fn_abi.ret)? {
+            throw_ub!(AbiMismatchReturn {
+                caller_ty: caller_fn_abi.ret.layout.ty,
+                callee_ty: callee_fn_abi.ret.layout.ty
+            });
+        }
+
+        if let Some(index) = caller_fn_abi
+            .args
+            .iter()
+            .zip(callee_fn_abi.args.iter())
+            .map(|(caller_arg, callee_arg)| this.check_argument_compat(caller_arg, callee_arg))
+            .collect::<InterpResult<'tcx, Vec<bool>>>()?
+            .into_iter()
+            .position(|b| !b)
+        {
+            throw_ub!(AbiMismatchArgument {
+                caller_ty: caller_fn_abi.args[index].layout.ty,
+                callee_ty: callee_fn_abi.args[index].layout.ty
+            });
+        }
+
+        if let Ok(ops) = caller_args.try_into() {
+            return interp_ok(ops);
+        }
+        unreachable!()
     }
 
     /// Check shim for variadic function.
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -1,7 +1,7 @@
 use std::ffi::OsStr;
 use std::str;
 
-use rustc_abi::Size;
+use rustc_abi::{ExternAbi, Size};
 use rustc_middle::ty::Ty;
 use rustc_middle::ty::layout::LayoutOf;
 use rustc_span::Symbol;
@@ -200,7 +200,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.write(fd, buf, count, Some(offset), dest)?;
             }
             "close" => {
-                let [fd] = this.check_shim(abi, Conv::C, link_name, args)?;
+                let [fd] = this.check_shim_abi(
+                    link_name,
+                    abi,
+                    ExternAbi::C { unwind: false },
+                    [this.tcx.types.i32],
+                    this.tcx.types.i32,
+                    args,
+                )?;
                 let result = this.close(fd)?;
                 this.write_scalar(result, dest)?;
             }
--- a/tests/fail/shims/input_arg_mismatch.rs
+++ b/tests/fail/shims/input_arg_mismatch.rs
@@ -1,0 +1,21 @@
+//@ignore-target: windows # File handling is not implemented yet
+//@compile-flags: -Zmiri-disable-isolation
+use std::ffi::{CString, OsStr, c_char, c_int};
+use std::os::unix::ffi::OsStrExt;
+
+extern "C" {
+    fn open(path: *const c_char, oflag: c_int, ...) -> c_int;
+    // correct fd type is i32
+    fn close(fd: u32) -> c_int;
+}
+
+fn main() {
+    let c_path = CString::new(OsStr::new("./text").as_bytes()).expect("CString::new failed");
+    let fd = unsafe {
+        open(c_path.as_ptr(), /* value does not matter */ 0)
+    } as u32;
+    let _ = unsafe {
+        close(fd);
+        //~^ ERROR: calling a function with argument of type i32 passing data of type u32
+    };
+}
--- a/tests/fail/shims/input_arg_mismatch.stderr
+++ b/tests/fail/shims/input_arg_mismatch.stderr
@@ -1,0 +1,17 @@
+error: Undefined Behavior: calling a function with argument of type i32 passing data of type u32
+  --> tests/fail/shims/input_arg_mismatch.rs:LL:CC
+   |
+LL |         close(fd);
+   |         ^^^^^^^^^ calling a function with argument of type i32 passing data of type u32
+   |
+   = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
+   = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
+   = help: this means these two types are not *guaranteed* to be ABI-compatible across all targets
+   = help: if you think this code should be accepted anyway, please report an issue with Miri
+   = note: BACKTRACE:
+   = note: inside `main` at tests/fail/shims/input_arg_mismatch.rs:LL:CC
+
+note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace
+
+error: aborting due to 1 previous error
+
--- a/tests/fail/shims/return_type_mismatch.rs
+++ b/tests/fail/shims/return_type_mismatch.rs
@@ -1,0 +1,21 @@
+//@ignore-target: windows # File handling is not implemented yet
+//@compile-flags: -Zmiri-disable-isolation
+use std::ffi::{CString, OsStr, c_char, c_int, c_short};
+use std::os::unix::ffi::OsStrExt;
+
+extern "C" {
+    fn open(path: *const c_char, oflag: c_int, ...) -> c_int;
+    // correct return type is i32
+    fn close(fd: c_int) -> c_short;
+}
+
+fn main() {
+    let c_path = CString::new(OsStr::new("./text").as_bytes()).expect("CString::new failed");
+    let fd = unsafe {
+        open(c_path.as_ptr(), /* value does not matter */ 0)
+    };
+    let _ = unsafe {
+        close(fd);
+        //~^ ERROR: calling a function with return type i32 passing return place of type i16
+    };
+}
--- a/tests/fail/shims/return_type_mismatch.stderr
+++ b/tests/fail/shims/return_type_mismatch.stderr
@@ -1,0 +1,17 @@
+error: Undefined Behavior: calling a function with return type i32 passing return place of type i16
+  --> tests/fail/shims/return_type_mismatch.rs:LL:CC
+   |
+LL |         close(fd);
+   |         ^^^^^^^^^ calling a function with return type i32 passing return place of type i16
+   |
+   = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
+   = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
+   = help: this means these two types are not *guaranteed* to be ABI-compatible across all targets
+   = help: if you think this code should be accepted anyway, please report an issue with Miri
+   = note: BACKTRACE:
+   = note: inside `main` at tests/fail/shims/return_type_mismatch.rs:LL:CC
+
+note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace
+
+error: aborting due to 1 previous error
+


geetanshjuneja:
I have made changes only in a single place to make sure that this what we want.

src/helpers.rs
```diff
@@ -1312,3 +1312,15 @@ pub(crate) fn windows_check_buffer_size((success, len): (bool, u64)) -> u32 {
         u32::try_from(len).unwrap()
     }
 }
+
+pub(crate) fn get_arg<'tcx>(
+    ecx: &mut MiriInterpCx<'tcx>,
```
RalfJung, line 1317: Please make this a method on the extension trait instead.

src/helpers.rs
```diff
@@ -1312,3 +1312,15 @@ pub(crate) fn windows_check_buffer_size((success, len): (bool, u64)) -> u32 {
...
+    op: &OpTy<'tcx>,
+    ty: Ty<'tcx>,
+) -> InterpResult<'tcx, OpTy<'tcx>> {
+    let ty_layout = ecx.layout_of(ty)?;
+    if !op.layout.backend_repr.eq_up_to_validity(&ty_layout.backend_repr) {
+        throw_ub_format!("Invalid argument type: ABI mismatch");
+    }
```
RalfJung, line 1324: Hm, this is not quite the same logic as what we use to test ABI compatibility. That check cannot really be done on a per-argument / per-type basis; it works on `ArgAbi` which is part of `FnAbi`.

So, while this gives some extra coverage, it doesn't really go far enough, and we'd have to change everything completely again if we wanted to do the "proper" thing. I am not convinced it is worth investing a lot of effort into a half-way solution.

RalfJung (commented):
Thanks for the PR!

Judging from the CI failures, this does not seem to be quite right. And (as noted inline) I think this will never be able to do the "proper" check. So I don't think this is the right approach.

The right approach has to be somewhat similar to what happens for regular function calls in Miri. Basically, we need a helper function that takes a list of types (the expected argument types) and a return type, and then somehow checks that against the `FnAbi` that is passed around in the Miri shim handling. This likely involved constructing our own `FnAbi` to compare, and I have no idea how to do that...

Also, I would suggest to *not* touch intrinsics for now. They are somewhat special ABI-wise. Let's start with normal shims, i.e. all the things on `foreign_items`.

geetanshjuneja:
> The right approach has to be somewhat similar to what happens for regular function calls in Miri. Basically, we need a helper function that takes a list of types (the expected argument types) and a return type, and then somehow checks that against the `FnAbi` that is passed around in the Miri shim handling. This likely involved constructing our own `FnAbi` to compare, and I have no idea how to do that...

For non variadic fn's can't we just simply create FnAbi struct using the types available in docs. The issue only comes for variadic fn's as we dont know the types of the extra args.

RalfJung:
Yes we "just" have to do that but it is not necessarily so easy. :)
But we can probably construct a fn ptr type corresponding to the desired signature, and then use `fn_abi_of_fn_ptr`.

geetanshjuneja:
To resolve I was thinking of implementing the following things:

Create a fn `check_abi` which would take a list of input types , an output type and `Fnabi` that is passed in the miri shim handling. This fn would create a [`FnSig`](https://github.com/rust-lang/rust/blob/54cdc751df770517e70db0588573e32e6a7b9821/compiler/rustc_type_ir/src/ty_kind.rs#L872) using the args. Now we will create expected `FnAbi` using this fn sig and compare it with the input Fnsig. If it doesn't matches throw ub.
Now Transmute can be called  safely inside shim.

RalfJung:
That sounds like a good plan. There are methods defined in https://github.com/rust-lang/rust/blob/master/compiler/rustc_const_eval/src/interpret/call.rs that are probably useful but have to be made `pub` before they can be called from Miri.

geetanshjuneja:
> There are methods defined in https://github.com/rust-lang/rust/blob/master/compiler/rustc_const_eval/src/interpret/call.rs that are probably useful but have to be made `pub` before they can be called from Miri.

Are you talking about `[check_argument_compat](https://github.com/rust-lang/rust/blob/a567209daab72b7ea59eac533278064396bb0534/compiler/rustc_const_eval/src/interpret/call.rs#L244)`  method here? If yes now then I think we can compare the two `FnAbi` by iterating over the args of the two abis by calling `check_argument_compat` method.

Also `inputs_and_output` field of `FnSig` requires list as `RawList<(), T>` so I need to convert given input and output arg slice into `RawList<(), T>` using `[from_arena](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/list/struct.RawList.html#method.from_arena)` method?

RalfJung:
> Are you talking about [check_argument_compat](https://github.com/rust-lang/rust/blob/a567209daab72b7ea59eac533278064396bb0534/compiler/rustc_const_eval/src/interpret/call.rs#L244)method here?

Yes.

> Also inputs_and_output field of FnSig requires list as RawList<(), T> so I need to convert given input and output arg slice into[ RawList<(), T>](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/list/type.List.html) using [from_arena](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/list/struct.RawList.html#method.from_arena) method?

You can use these methods: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/type.FnSig.html?search=mk_type_list.

src/helpers.rs
geetanshjuneja: In this `MiriInterpCxExt` trait?

RalfJung: It's the big block that beings with `pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {`.

geetanshjuneja:
I am mostly done with the code now I just need `check_argument_compat` to be public.

RalfJung:
Okay, please make a PR against the rustc repo to make that function public then. :)

RalfJung:
Please don't do rustup in the middle of an unrelated PR. I did a regular rustup, so if you rebase your PR should pick that up. :)

src/helpers.rs
```diff
@@ -1187,6 +1187,58 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
         interp_ok(array)
     }
+
+    fn check_fn_abi(
```
RalfJung, line 1191: Please add a doc comment.

src/helpers.rs
```diff
@@ -1187,6 +1187,58 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        &mut self,
+        input_args: &[Ty<'tcx>],
+        output_ty: Ty<'tcx>,
+        safety: Safety,
```
RalfJung, line 1195: safety is not relevant for the ABI, it should not show up here.

src/helpers.rs
```diff
@@ -1187,6 +1187,58 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        abi: ExternAbi,
```
RalfJung, line 1196: This should probably go before `input_args`, also to match the order in which these things show up in the Rust source code (`extern "ABI" fn(input) -> output`).

src/helpers.rs
```diff
@@ -1187,6 +1187,58 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        });
+        let exp_fn_abi = this.fn_abi_of_fn_ptr(fn_sig_binder, Default::default())?;
+        if !self.check_abi_compat(fn_abi, exp_fn_abi)? {
+            throw_ub_format!("Invalid argument type: ABI mismatch");
+        }
+        interp_ok(())
+    }
+
+    // Check where two FnAbi's are compatible
+    fn check_abi_compat(
```
RalfJung, line 1216: Why is this a separate function?

src/helpers.rs
```diff
@@ -1187,6 +1187,58 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let caller_args_abi = &caller_fn_abi.args;
+        let callee_args_abi = &callee_fn_abi.args;
+
+        if caller_fn_abi.c_variadic != callee_fn_abi.c_variadic
+            || callee_fn_abi.fixed_count != caller_fn_abi.fixed_count
+            || callee_fn_abi.can_unwind != caller_fn_abi.can_unwind
+            || !this.check_argument_compat(&caller_fn_abi.ret, &callee_fn_abi.ret)?
+            || !caller_args_abi
+                .iter()
+                .zip(callee_args_abi.iter())
```
RalfJung, line 1231: You have to also check that the two lists have the same length.

src/shims/unix/foreign_items.rs
```diff
@@ -200,6 +201,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             }
             "close" => {
+                this.check_fn_abi(
+                    &[this.machine.layouts.i32.ty],
+                    this.machine.layouts.i32.ty,
+                    Safety::Unsafe,
+                    ExternAbi::C { unwind: false },
+                    abi,
+                )?;
                 let [fd] = this.check_shim(abi, Conv::C, link_name, args)?;
```
RalfJung, line 211: I think we want to look for a better API here -- in the end, `check_shim` and `check_fn_abi` should be just one call. Let's call that function `check_shim_abi`. It could look something like:
```rust
let [fd] = this.check_shim_abi(link_name, abi, ExternAbi::C { unwind: false }, [this.tcx.types.i32], this.tcx.types.i32, args);
```
I feel like we might want to explore ways to make typing the signature shorter (e.g. with a macro), but let's not do that in this PR.

RalfJung:
@rustbot author

src/helpers.rs
geetanshjuneja: isn't that already checked by `callee_fn_abi.fixed_count != caller_fn_abi.fixed_count`

src/helpers.rs
geetanshjuneja: I thought it would be better to create a function which check whether two fnabis are compatible similar to `check_argument_compat`.

src/helpers.rs
geetanshjuneja: So I don't have to perform safety check on the two abi's?

src/helpers.rs
RalfJung: That's the fixed arguments... so for variadics this seems still wrong. Just don't even pretend to support variadics for now.

Also if we want to rely on such a thing, there should still be an assertion.

src/shims/unix/foreign_items.rs
geetanshjuneja: But this change would require a lot of changes in the codebase as `check_shim` is used everywhere. So I was thinking changing `check_shim` to `check_shim_abi` can be done in the other PR.

src/helpers.rs
RalfJung: You're making the problem harder that way, e.g. now you have to think about variadic functions, which we don't want to think about in this PR. So I don't think this function carries its weight.

src/helpers.rs
RalfJung: Safety is not relevant for ABI. I don't know why it even is part of `FnAbi` but you should ignore it here.

src/shims/unix/foreign_items.rs
RalfJung: I didn't say to replace *all* `check_shim`. I said to add a new `check_shim_abi` and use it here and nowhere else.

src/helpers.rs
geetanshjuneja: Ok I will remove it and merge it's functionality with `check_shim_abi`.

src/helpers.rs
```diff
@@ -1015,6 +1015,67 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         )
     }
 
+    // Creates Expected FnAbi of the called fn from the args and matches it's compatability with the
+    // FnAbi passed along miri the shim handling and also checks for mismatch between shim symbol and abi.
```
RalfJung, line 1019:
```suggestion
    /// Check that the given `caller_fn_abi` matches the expected ABI described by
    /// `callee_abi`, `callee_input_tys`, `callee_output_ty`, and the return the list of
    /// arguments.
```

src/helpers.rs
```diff
@@ -1015,6 +1015,67 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    fn check_shim_abi<'a, const N: usize>(
+        &mut self,
+        link_name: Symbol,
+        caller_fn_abi: &FnAbi<'tcx, Ty<'tcx>>,
+        abi: ExternAbi,
+        callee_input_args_ty: &[Ty<'tcx>],
+        callee_output_ty: Ty<'tcx>,
+        caller_args: &'a [OpTy<'tcx>],
+    ) -> InterpResult<'tcx, &'a [OpTy<'tcx>; N]>
```
RalfJung, line 1028:
```suggestion
        callee_abi: ExternAbi,
        callee_input_tys: [Ty<'tcx>; N],
        callee_output_ty: Ty<'tcx>,
        caller_args: &'a [OpTy<'tcx>],
    ) -> InterpResult<'tcx, &'a [OpTy<'tcx>; N]>
```

src/helpers.rs
```diff
@@ -1015,6 +1015,67 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    where
+        &'a [OpTy<'tcx>; N]: TryFrom<&'a [OpTy<'tcx>]>,
+    {
+        let this = self.eval_context_mut();
+        let mut inputs_and_output = callee_input_args_ty.to_vec();
+        inputs_and_output.push(callee_output_ty);
+        let fn_sig_binder = Binder::dummy(FnSig {
+            inputs_and_output: this.machine.tcx.mk_type_list(&inputs_and_output),
+            c_variadic: false,
+            safety: Safety::Safe,
```
RalfJung, line 1038:
```suggestion
            // This does not matter for the ABI.
            safety: Safety::Safe,
```

src/helpers.rs
```diff
@@ -1015,6 +1015,67 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            abi,
+        });
+        let callee_fn_abi = this.fn_abi_of_fn_ptr(fn_sig_binder, Default::default())?;
+
+        let caller_args_abi = &caller_fn_abi.args;
+        let callee_args_abi = &callee_fn_abi.args;
```
RalfJung, line 1044: Please remove these let-bindings, they are not worth it.

src/helpers.rs
```diff
@@ -1015,6 +1015,67 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            || !caller_args_abi
+                .iter()
+                .zip(callee_args_abi.iter())
+                .map(|(caller_arg, callee_arg)| this.check_argument_compat(caller_arg, callee_arg))
+                .collect::<InterpResult<'tcx, Vec<bool>>>()?
+                .into_iter()
+                .all(|b| b)
+        {
+            throw_ub_format!("Invalid argument type: ABI mismatch");
+        }
```
RalfJung, line 1060: Instead of one big `if`, please do separate checks so that we can give an error that says a bit more precisely what is going wrong.

src/helpers.rs
```diff
@@ -1015,6 +1015,67 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        self.check_abi_and_shim_symbol_clash(caller_fn_abi, callee_fn_abi.conv, link_name)?;
+
+        if caller_fn_abi.c_variadic {
```
RalfJung, line 1064: This is redundant with the check above.

RalfJung (commented):
@rustbot author

src/helpers.rs
RalfJung: Indeed that does not seem redundant.

Also to be clear, please keep the `if caller_fn_abi.c_variadic {`. Probably the order should be:
- `check_abi_and_shim_symbol_clash`
- `if caller_fn_abi.c_variadic {`
- arguments and return types

geetanshjuneja: I have to do `let this = self.eval_context_mut();` two times because of the above required order as check_abi_and_shim_symbol_clash require mutable ref of `self` and also `conv` of the constructed fn abi.
Is this fine?

RalfJung: No that should not be needed. You can call `this.check_abi_and_shim_symbol_clash()`.

geetanshjuneja:
@RalfJung plz review I have made the required changes.

RalfJung:
Hey, have some patience. I do this mostly in my free time. You're in the queue, and I will get to your PR eventually.

src/helpers.rs
```diff
@@ -1015,6 +1015,81 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        &mut self,
+        link_name: Symbol,
+        caller_fn_abi: &FnAbi<'tcx, Ty<'tcx>>,
+        callee_abi: ExternAbi,
+        callee_input_tys: [Ty<'tcx>; N],
+        callee_output_ty: Ty<'tcx>,
+        caller_args: &'a [OpTy<'tcx>],
+    ) -> InterpResult<'tcx, &'a [OpTy<'tcx>; N]>
+    where
+        &'a [OpTy<'tcx>; N]: TryFrom<&'a [OpTy<'tcx>]>,
```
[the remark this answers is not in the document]
oli-obk: Ah, probably from the olden days when we only had impls for arrays of length 32

src/helpers.rs
```diff
@@ -1015,57 +1015,71 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                caller_fn_abi.fixed_count
+            );
+        }
+        if callee_fn_abi.can_unwind != caller_fn_abi.can_unwind {
+            throw_ub_format!(
+                "ABI mismatch: unwinding behavior differs called function {} unwind, but expected function {}",
+                if caller_fn_abi.can_unwind { "can" } else { "cannot" },
+                if callee_fn_abi.can_unwind { "can" } else { "cannot" }
+            );
+        }
```
RalfJung, line 1065: This is not UB, please remove the check.

What is relevant is that a function called with a non-unwinding ABI must never unwind. We need to be mindful of this whenever a shim implementation unwinds.

src/helpers.rs
```diff
@@ -1015,57 +1015,71 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        if !this.check_argument_compat(&caller_fn_abi.ret, &callee_fn_abi.ret)? {
+            throw_ub_format!("ABI mismatch: return type mismatches expected return type");
```
RalfJung, line 1067: Please use the same error as for regular calls:
```rust
                throw_ub!(AbiMismatchReturn {
                    caller_ty: caller_fn_abi.ret.layout.ty,
                    callee_ty: callee_fn_abi.ret.layout.ty
                });
```

src/helpers.rs
```diff
@@ -1015,57 +1015,71 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            .iter()
+            .zip(callee_fn_abi.args.iter())
+            .map(|(caller_arg, callee_arg)| this.check_argument_compat(caller_arg, callee_arg))
+            .collect::<InterpResult<'tcx, Vec<bool>>>()?
+            .into_iter()
+            .all(|b| b)
+        {
+            throw_ub_format!(
+                "ABI mismatch: function argument types mismatches expected argument types"
+            );
```
RalfJung, line 1080: Please use the same error as for regular calls:
```rust
            throw_ub!(AbiMismatchArgument {
                caller_ty: caller_abi.layout.ty,
                callee_ty: callee_abi.layout.ty
            });
```

RalfJung (commented):
Sorry for the long wait!

src/helpers.rs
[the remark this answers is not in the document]
RalfJung: If you can just remove the bound here, then please do.

src/helpers.rs
geetanshjuneja: So I need to add a ub for the case where the callee can unwind but caller abi cannot unwind?

geetanshjuneja:
Also do I need to add a test for return type mismatch?

src/helpers.rs
RalfJung: Yes that sounds like a good idea.

src/helpers.rs
```diff
@@ -1015,6 +1015,81 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        }
+
+        if let Ok(ops) = caller_args.try_into() {
+            return interp_ok(ops);
+        }
+        throw_ub_format!(
+            "incorrect number of arguments for `{link_name}`: got {}, expected {}",
+            caller_args.len(),
+            N
+        )
```
RalfJung, line 1090: You already checked this above, didn't you?
```suggestion
        unreachable!()
```

src/helpers.rs
RalfJung: (This does not seem resolved yet)

src/helpers.rs
```diff
@@ -1015,6 +1012,80 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            throw_ub_format!(
+                "ABI mismatch: expected {} arguments, found {} arguments ",
+                callee_fn_abi.fixed_count,
+                caller_fn_abi.fixed_count
+            );
+        }
+
+        if callee_fn_abi.can_unwind && !caller_fn_abi.can_unwind {
+            throw_ub_format!(
+                "ABI mismatch: unwinding behavior differs called function cannot unwind, but expected function can",
```
RalfJung, line 1057:
```suggestion
                "ABI mismatch: callee may unwind, but caller-side signature prohibits unwinding",
```

src/helpers.rs
RalfJung: This one is still open, isn't it?

geetanshjuneja: I have removed this check and added this check, as you asked.
```rust
if callee_fn_abi.can_unwind && !caller_fn_abi.can_unwind {
            throw_ub_format!(
                "ABI mismatch: callee may unwind, but caller-side signature prohibits unwinding",
            );
        }

RalfJung: AH sorry I posted this in the wrong comment.

RalfJung:
> Also do I need to add a test for return type mismatch?

Yes please do.

RalfJung:

@rustbot author

RalfJung:
Just use `c_short` instead of `c_char`.

RalfJung:
Thanks for the PR and the patience. :)

geetanshjuneja:
So the next step would be to use `check_shim_abi` for non variadic shims? If yes I can work on that as well

RalfJung:
Yeah the next step is to use this for all the existing shims. Please don't do them all in one PR. :) Pick one file, e.g. `src/shims/foreign_items.rs`, and make a PR just for that. In the first pass, skip any that cause some extra complications.

PR #4192, opened 2025-02-14
Remove the build script for miri

bjorn3:
Setting the `TARGET` env var can be replaced with using `rustc_session::config::host_tuple` at runtime. And the check-cfg can be replaced with using the lints section in Cargo.toml.


RalfJung:
LGTM, but please replace the import by an absolute path at the call site to avoid unused imports on Windows hosts.

Issue #4191, opened 2025-02-13
Support memchr on arm64 too


bjorn3:
Miri has implemented all vendor intrinsics needed by the memchr crate on x86. It still misses them for arm64 however.

RalfJung:
Yeah we had a heroic contributor implement a lot of x86 intrinsics.

Do you have a list of intrinsics we'd need for this? Is there some place the semantics of these intrinsics are documented?

bjorn3:
Only `vpmaxq_u8` needed a miri implementation. There rest seems to use portable simd intrinsics internally.

RalfJung:
> the core::arch doc pages link to the arm documentation 

I can't find such a link at https://doc.rust-lang.org/nightly/core/arch/aarch64/fn.vpmaxq_u8.html ?

bjorn3:
I guess some are missing the link. In any case I've implemented it locally, but I'm getting some weird test failures.

PR #4193, opened 2025-02-14
Implement vpmaxq_u8 on aarch64

bjorn3:
This is documented in https://developer.arm.com/architectures/instruction-sets/intrinsics/#q=vpmaxq_u8

This makes all memchr tests suite pass when I remove `#[cfg(not(miri))]` from tests.

Fixes https://github.com/rust-lang/miri/issues/4191

### DIFF
--- a/src/shims/aarch64.rs
+++ b/src/shims/aarch64.rs
@@ -46,15 +46,15 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
                 for lane_idx in 0..lane_count {
                     let src = if lane_idx < (lane_count / 2) { &left } else { &right };
-                    #[allow(clippy::arithmetic_side_effects)]
-                    let src_idx = lane_idx % (lane_count / 2);
+                    let src_idx = lane_idx.strict_rem(lane_count / 2);
 
-                    #[allow(clippy::arithmetic_side_effects)]
-                    let lhs_lane = this.read_immediate(&this.project_index(src, src_idx * 2)?)?;
-                    #[allow(clippy::arithmetic_side_effects)]
-                    let rhs_lane =
-                        this.read_immediate(&this.project_index(src, src_idx * 2 + 1)?)?;
+                    let lhs_lane =
+                        this.read_immediate(&this.project_index(src, src_idx.strict_mul(2))?)?;
+                    let rhs_lane = this.read_immediate(
+                        &this.project_index(src, src_idx.strict_mul(2).strict_add(1))?,
+                    )?;
 
+                    // Compute `if lhs > rhs { lhs } else { rhs }`, i.e., `max`.
                     let res_lane = if this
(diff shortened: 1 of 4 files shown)


src/shims/aarch64.rs
```diff
@@ -0,0 +1,69 @@
...
+                    15 => {
+                        this.yield_active_thread();
+                    }
+                    _ => {
+                        throw_unsup_format!("unsupported llvm.aarch64.isb argument {}", arg);
+                    }
+                }
+            }
+
+            "neon.umaxp.v16i8" => {
```
RalfJung, line 37: Please follow the intel intrinsics: there should always be a comment saying which actual intrinsic this corresponds to, and a very brief description of what the intrinsic does.

src/shims/aarch64.rs
```diff
@@ -0,0 +1,69 @@
...
+                    let src = if lane_idx < (lane_count / 2) { &left } else { &right };
+                    let src_idx = lane_idx % (lane_count / 2);
+
+                    let lhs_lane = this.read_scalar(&this.project_index(src, src_idx * 2)?)?;
+                    let rhs_lane = this.read_scalar(&this.project_index(src, src_idx * 2 + 1)?)?;
+
+                    let res_lane = match unprefixed_name {
+                        "neon.umaxp.v16i8" => {
+                            Scalar::from_i8(cmp::max(lhs_lane.to_i8()?, rhs_lane.to_i8()?))
+                        }
```
RalfJung, line 56: If this ends up existing for a bunch of types, it's probably better to use a generic implementation directly on `ImmTy`, using the interpreter methods for arithmetic on immediates.

bjorn3: Which function should I use for that?

RalfJung: `this.binary_op`

bjorn3:
Turns out the memchr test failures were because I was accidentally doing signed rather than unsigned comparisons. All tests are passing now.

RalfJung:
Please add a test; you can use the x86 intrinsic tests as template. In particular there should be a Miri test directly checking the signed vs unsigned issue.

RalfJung (commented):
@rustbot author

bjorn3:
Added tests.

@rustbot ready

src/shims/aarch64.rs
```diff
@@ -0,0 +1,77 @@
...
+                        this.yield_active_thread();
+                    }
+                    _ => {
+                        throw_unsup_format!("unsupported llvm.aarch64.isb argument {}", arg);
+                    }
+                }
+            }
+
+            // Used to implement the vpmaxq_u8 function.
+            // Folding maximum of adjacent pairs.
```
RalfJung, line 36:
```suggestion
            // Computes the maximum of adjacent pairs; the first half of the output is produced from the first
            // input, the second half of the output from the second input.
```

RalfJung (commented):
LGTM, just a comment nit.

PR #4209, opened 2025-02-25
Freebsd futex

### LINKED ISSUE #3553: Implement futex support on FreeBSD (_umtx_op)
When running our test suite on FreeBSD, almost all test failures are due to missing an implementation of `_umtx_op`, which is the FreeBSD version of futexes. Given that we have shared futex infrastructure that backs both the Linux and Windows futex support, it shouldn't be too hard to also add support for the FreeBSD variant of this.

There was a PR at https://github.com/rust-lang/miri/pull/2321 that had some issues but could be used as a starting point.

LorrensP-2158466:
Fixes #3553.

Currently This implements the WAIT and WAKE operations of the `_umtx_op` syscall.

I have not yet changed `ci/ci.sh` because I don't exactly know which filters to pass. Currently if I test it locally only the the tests that use `libc::cpuset_getaffinity` fail.

### DIFF
--- a/src/shims/unix/freebsd/foreign_items.rs
+++ b/src/shims/unix/freebsd/foreign_items.rs
@@ -2,6 +2,7 @@ use rustc_middle::ty::Ty;
 use rustc_span::Symbol;
 use rustc_target::callconv::{Conv, FnAbi};
 
+use super::sync::EvalContextExt as _;
 use crate::shims::unix::*;
 use crate::*;
 
@@ -53,6 +54,12 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     ThreadNameResult::ThreadNotFound => this.eval_libc("ESRCH"),
                 };
                 this.write_scalar(res, dest)?;
+            }
+
+            "_umtx_op" => {
+                let [obj, op, val, uaddr, uaddr2] =
+                    this.check_shim(abi, Conv::C, link_name, args)?;
+                this._umtx_op(obj, op, val, uaddr, uaddr2, dest)?;
             }
 
             // File related shims
--- a/src/shims/unix/freebsd/sync.rs
+++ b/src/shims/unix/freebsd/sync.rs
@@ -137,8 +137,12 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                             }
                         ),
                     );
-                };
-                interp_ok(())
+                    interp_ok(())
+                } else {
+                    // The manual doesn't document what should happen if `val` is invalid, so we error out.
+                    // TODO: do something else?
+                    this.set_last_error_and_return(LibcError("EINVAL"), dest)
+                }
             }
             // UMTX_OP_WAKE has a private variant that enables an optimization that stops it from working across processes.
             // Miri doesn't support that anyway, so we ignore that variant and use the same implementation for all wait ops.
--- a/src/shims/unix/freebsd/sync.rs
+++ b/src/shims/unix/freebsd/sync.rs
@@ -1,0 +1,189 @@
+//! Contains FreeBSD-specific synchronization functions
+
+use core::time::Duration;
+
+use crate::concurrency::sync::FutexRef;
+use crate::*;
+
+pub struct FreeBSDFutex {
+    futex: FutexRef,
+}
+
+pub struct UmtxTime {
+    timeout: Duration,
+    flags: u32,
+    _clock_id: u32, // TODO: I'm not understanding why this is needed atm
+}
+
+impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
+pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    /// Implementation of the FreeBSD [`_umtx_op`](https://man.freebsd.org/cgi/man.cgi?query=_umtx_op&sektion=2&manpath=FreeBSD+14.2-RELEASE+and+Ports) syscall. :
+    /// This is used for futex operations.
+    ///
+    /// `obj`: a pointer to the futex object (can be a lot of things, mostly *AtomicU32)
+    /// `op`: the futex operation to run
+    /// `val`: the current value of the object as a `c_long` (for wait/wake)
+    /// `uaddr`: pointer to optional parameter (mostly timeouts)
+    /// `uaddr2`: pointer to optional parameter (mostly timeouts)
+    /// `dest`: the place this syscall returns to, 0 for success, -1 for failure
+    fn _umtx_op(
+        &mut self,
+        obj: &OpTy<'tcx>,
+        op: &OpTy<'tcx>,
+        val: &OpTy<'tcx>,
+        uaddr: &OpTy<'tcx>,
+        uaddr2: &OpTy<'tcx>,
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx> {
+        let this = self.eval_context_mut();
+
+        let obj = this.read_pointer(obj)?;
+        let op = this.read_scalar(op)?.to_i32()?;
+        let val = this.read_target_usize(val)?;
+        let uaddr2 = this.read_pointer(uaddr2)?;
+
+        let wait = this.eval_libc_i32("UMTX_OP_WAIT");
+        let wait_uint = this.eval_libc_i32("UMTX_OP_WAIT_UINT");
+        let wait_uint_private = this.eval_libc_i32("UMTX_OP_WAIT_UINT_PRIVATE");
+
+        let wake = this.eval_libc_i32("UMTX_OP_WAKE");
+        let wake_private = this.eval_libc_i32("UMTX_OP_WAKE_PRIVATE");
+
+        let absolute_time_flag = this.eval_libc_u32("UMTX_ABSTIME");
+
+        match op {
+            // UMTX_OP_WAIT_UINT has a private variant that enables an optimization that stops it from working across processes.
+            // Miri doesn't support that anyway, so we ignore that variant and use the same implementation for all wait ops.
+            op if op == wait || op == wait_uint || op == wait_uint_private => {
+                // TODO: the normal UMTX_OP_WAIT accepts a `long` instead of u64,
+                // does this really matter?? (probably)
+
+                let obj = this.ptr_to_mplace(obj, this.machine.layouts.i32);
+
+                // TODO: the Linux futex syscall sets a fence here, but what I understand from the manual is that
+                // these semantics need to be done by the user. Should this still be added?
+                if u64::from(this.read_scalar_atomic(&obj, AtomicReadOrd::Acquire)?.to_u32()?)
+                    == val
+                {
+                    let futex_ref = this
+                        .get_sync_or_init(obj.ptr(), |_| FreeBSDFutex { futex: Default::default() })
+                        .unwrap()
+                        .futex
+                        .clone();
+
+                    // TODO: This can be cleaned up no? :)
+                    // From the man page:
+                    // If `uaddr2` is null than `uaddr` can point to an optional timespec parameter
+                    // otherwise `uaddr2` must point to a `_umtx_time` parameter and the value of `uaddr`
+                    // must be equal to the size of that struct.
+                    let timeout = if this.ptr_is_null(uaddr2)? {
+                        if this.ptr_is_null(this.read_pointer(uaddr)?)? {
+                            None
+                        } else {
+                            let timespec =
+                                this.deref_pointer_as(uaddr, this.libc_ty_layout("timespec"))?;
+                            let duration = match this.read_timespec(&timespec)? {
+                                Some(duration) => duration,
+                                None => {
+                                    return this
+                                        .set_last_error_and_return(LibcError("EINVAL"), dest);
+                                }
+                            };
+
+                            Some((TimeoutClock::Monotonic, TimeoutAnchor::Relative, duration))
+                        }
+                    } else {
+                        let umtx_time_place =
+                            this.ptr_to_mplace(uaddr2, this.libc_ty_layout("_umtx_time"));
+                        let uaddr_as_size = this.read_target_usize(uaddr)?;
+
+                        if umtx_time_place.layout().size.bytes() != uaddr_as_size {
+                            return this.set_last_error_and_return(LibcError("EINVAL"), dest);
+                        }
+
+                        // Inner `timespec` must still be valid.
+                        let umtx_time = match this.read_umtx_time(&umtx_time_place)? {
+                            Some(duration) => duration,
+                            None => {
+                                return this.set_last_error_and_return(LibcError("EINVAL"), dest);
+                            }
+                        };
+
+                        let anchor = if umtx_time.flags == absolute_time_flag {
+                            TimeoutAnchor::Absolute
+                        } else {
+                            TimeoutAnchor::Relative
+                        };
+
+                        Some((TimeoutClock::Monotonic, anchor, umtx_time.timeout))
+                    };
+
+                    let dest = dest.clone();
+                    this.futex_wait(
+                        futex_ref,
+                        u32::MAX, // we set the bitset to include all bits
+                        timeout,
+                        callback!(
+                            @capture<'tcx> {
+                                dest: MPlaceTy<'tcx>,
+                            }
+                            |ecx, unblock: UnblockKind| match unblock {
+                                UnblockKind::Ready => {
+                                    ecx.write_int(0, &dest)
+                                }
+                                UnblockKind::TimedOut => {
+                                    ecx.set_last_error_and_return(LibcError("ETIMEDOUT"), &dest)
+                                }
+                            }
+                        ),
+                    );
+                };
+                interp_ok(())
+            }
+            // UMTX_OP_WAKE has a private variant that enables an optimization that stops it from working across processes.
+            // Miri doesn't support that anyway, so we ignore that variant and use the same implementation for all wait ops.
+            op if op == wake || op == wake_private => {
+                let Some(futex_ref) =
+                    this.get_sync_or_init(obj, |_| FreeBSDFutex { futex: Default::default() })
+                else {
+                    // From Linux implemenation:
+                    // No AllocId, or no live allocation at that AllocId.
+                    // Return an error code. (That seems nicer than silently doing something non-intuitive.)
+                    // This means that if an address gets reused by a new allocation,
+                    // we'll use an independent futex queue for this... that seems acceptable.
+                    return this.set_last_error_and_return(LibcError("EFAULT"), dest);
+                };
+                let futex_ref = futex_ref.futex.clone();
+
+                // `_umtx_op` doesn't return the amount of woken threads.
+                let _woken = this.futex_wake(
+                &futex_ref,
+                u32::MAX,     // we set the bitset to include all bits
+                val as usize, /* TODO: I can't read a usize from val, I haven't found a way as of now */
+            )?;
+                this.write_int(0, dest)?;
+                interp_ok(())
+            }
+            op => {
+                throw_unsup_format!("Miri does not support `_umtx_op` syscall with op={}", op)
+            }
+        }
+    }
+
+    /// Parses a `_umtx_time` struct.
+    /// Returns `None` if the underlying `timespec` struct is invalid.
+    fn read_umtx_time(&mut self, ut: &MPlaceTy<'tcx>) -> InterpResult<'tcx, Option<UmtxTime>> {
+        let this = self.eval_context_mut();
+        let timespec_place = this.project_field(ut, 0)?;
+        // Inner `timespec` must still be valid.
+        let duration = match this.read_timespec(&timespec_place)? {
+            Some(dur) => dur,
+            None => return interp_ok(None),
+        };
+        let flags_place = this.project_field(ut, 1)?;
+        let flags = this.read_scalar(&flags_place)?.to_u32()?;
+        let clock_id_place = this.project_field(ut, 2)?;
+        let clock_id = this.read_scalar(&clock_id_place)?.to_u32()?;
+        interp_ok(Some(UmtxTime { timeout: duration, flags, _clock_id: clock_id }))
+    }
+}
--- a/src/shims/unix/freebsd/mod.rs
+++ b/src/shims/unix/freebsd/mod.rs
@@ -1,1 +1,2 @@
 pub mod foreign_items;
+pub mod sync;


RalfJung:
Thanks for the PR! We have a significant backlog currently so unfortunately it could take a while until you get a proper review, sorry.

> I have not yet changed ci/ci.sh because I don't exactly know which filters to pass. Currently if I test it locally only the the tests that use libc::cpuset_getaffinity fail.

You should add `sync` to have it test the libstd synchronization primitives.

Also, please add a test that directly calls the FreeBSD functions.

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,219 @@
+//! Contains FreeBSD-specific synchronization functions
+
+use core::time::Duration;
+use std::u64;
+
+use rustc_span::sym::TryFrom;
```
saethlin, line 6:
```suggestion
```

Are these IDE-generated suggestions? You shouldn't need either of them, and `TryFrom` (if it's even accessible from this path) is an accidental re-export.

LorrensP-2158466: Oops you are right :) ! I noticed them but still forgot to remove them.

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,216 @@
+//! Contains FreeBSD-specific synchronization functions
+
+use core::time::Duration;
+
+use crate::concurrency::sync::FutexRef;
+use crate::*;
+
+pub struct FreeBSDFutex {
```
RalfJung, line 8: Please follow standard Rust naming conventions and call this `FreeBsdFutex`.

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,216 @@
...
+        uaddr: &OpTy<'tcx>,
+        uaddr2: &OpTy<'tcx>,
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx> {
+        let this = self.eval_context_mut();
+
+        let obj = this.read_pointer(obj)?;
+        let op = this.read_scalar(op)?.to_i32()?;
+        let val = this.read_target_usize(val)?;
+        let uaddr2 = this.read_pointer(uaddr2)?;
```
RalfJung, line 43: Why is `uaddr` being skipped here?

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,216 @@
...
+        let wait_uint_private = this.eval_libc_i32("UMTX_OP_WAIT_UINT_PRIVATE");
+
+        let wake = this.eval_libc_i32("UMTX_OP_WAKE");
+        let wake_private = this.eval_libc_i32("UMTX_OP_WAKE_PRIVATE");
+
+        let absolute_time_flag = this.eval_libc_u32("UMTX_ABSTIME");
+
+        match op {
+            // UMTX_OP_WAIT_UINT has a private variant that enables an optimization that stops it from working across processes.
+            // Miri doesn't support that anyway, so we ignore that variant and use the same implementation for all wait ops.
```
RalfJung, line 56:
```suggestion
            // UMTX_OP_WAIT_UINT and UMTX_OP_WAIT_UINT_PRIVATE only differ in whether they work across
            // processes or not. For Miri, we can treat them the same.
```
I was very confused at first because "private variant" sounded like this was not part of the public API. ;)

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,216 @@
...
+            op if op == wait || op == wait_uint || op == wait_uint_private => {
+                // TODO: A better way to do this? Because I don't want to duplicate the actual logic.
```
RalfJung, line 58: You can do something like
```rust
let obj_layout = if op == wait { this.machine.layouts.isize } else {this.machine.layouts.u32};
```
and then write the rest of the code uniformly.

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,216 @@
...
+                    let obj = this.ptr_to_mplace(obj, this.machine.layouts.isize);
+
+                    // Read the Linux futex implementation in Miri to understand why this fence is needed.
+                    this.atomic_fence(AtomicFenceOrd::SeqCst)?;
+                    let obj_val = this
+                        .read_scalar_atomic(&obj, AtomicReadOrd::Acquire)?
+                        .to_target_isize(this)?;
+
+                    // If obj points to something negative this can never hit, otherwise we can test.
+                    if obj_val < 0 { false } else { val == u64::try_from(obj_val).unwrap() }
```
RalfJung, line 70: I don't understand why negative values are different here. Is this mentioned in the docs somewhere?

I assume when the docs say "The  current  value of the variable pointed to by the obj argument is compared	with the val", they mean that the bits are compared, not the mathematical values. So probably we should just treat the `obj` as `usize` here, not `isize`.

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,216 @@
...
+                                this.deref_pointer_as(uaddr, this.libc_ty_layout("timespec"))?;
+                            let duration = match this.read_timespec(&timespec)? {
+                                Some(duration) => duration,
+                                None => {
+                                    return this
+                                        .set_last_error_and_return(LibcError("EINVAL"), dest);
+                                }
+                            };
+
+                            Some((TimeoutClock::Monotonic, TimeoutAnchor::Relative, duration))
```
RalfJung, line 109: Please quote the relevant part of the docs for which clock and anchor you are using here.

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,216 @@
...
+                    } else {
+                        let umtx_time_place =
+                            this.ptr_to_mplace(uaddr2, this.libc_ty_layout("_umtx_time"));
+                        let uaddr_as_size = this.read_target_usize(uaddr)?;
+
+                        if umtx_time_place.layout().size.bytes() != uaddr_as_size {
+                            return this.set_last_error_and_return(LibcError("EINVAL"), dest);
+                        }
+
+                        // Inner `timespec` must still be valid.
```
RalfJung, line 120: I don't understand this comment.

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,216 @@
...
+                            }
+                        };
+
+                        let anchor = if umtx_time.flags == absolute_time_flag {
+                            TimeoutAnchor::Absolute
+                        } else {
+                            TimeoutAnchor::Relative
+                        };
+
+                        Some((TimeoutClock::Monotonic, anchor, umtx_time.timeout))
```
RalfJung, line 134: This should respect the clock ID set in the `umtx_time`.

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,216 @@
...
+                };
+
+                if equal {
+                    let futex_ref = this
+                        .get_sync_or_init(obj, |_| FreeBSDFutex { futex: Default::default() })
+                        .unwrap()
+                        .futex
+                        .clone();
+
+                    // TODO: This can be cleaned up no? :)
```
RalfJung, line 89: Can it? Given the somewhat crazy semantics, this doesn't seem unreasonable.^^

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,216 @@
...
+                        futex_ref,
+                        u32::MAX, // we set the bitset to include all bits
+                        timeout,
+                        callback!(
+                            @capture<'tcx> {
+                                dest: MPlaceTy<'tcx>,
+                            }
+                            |ecx, unblock: UnblockKind| match unblock {
+                                UnblockKind::Ready => {
+                                    ecx.write_int(0, &dest)
```
RalfJung, line 148: Please quote the docs that say that we should return 0 on success.

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,216 @@
...
+                                UnblockKind::TimedOut => {
+                                    ecx.set_last_error_and_return(LibcError("ETIMEDOUT"), &dest)
+                                }
+                            }
+                        ),
+                    );
+                    interp_ok(())
+                } else {
+                    // The manual doesn't document what should happen if `val` is invalid, so we error out.
+                    this.set_last_error_and_return(LibcError("EINVAL"), dest)
```
RalfJung, line 159: This is the "inequal" case, right? that's definitely not an error, it just means another thread was faster. "EINVAL" sounds odd.
Please confirm the actual behavior with a test against the real implementation.

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,216 @@
...
+                }
+            }
+            // UMTX_OP_WAKE has a private variant that enables an optimization that stops it from working across processes.
+            // Miri doesn't support that anyway, so we ignore that variant and use the same implementation for all wake ops.
```
RalfJung, line 163: Please adjust this comment the same way as the one above.

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,216 @@
...
+                        .clone();
+
+                    // TODO: This can be cleaned up no? :)
+                    // From the man page:
+                    // If `uaddr2` is null than `uaddr` can point to an optional timespec parameter
+                    // otherwise `uaddr2` must point to a `_umtx_time` parameter and the value of `uaddr`
+                    // must be equal to the size of that struct.
+                    let timeout = if this.ptr_is_null(uaddr2)? {
+                        if this.ptr_is_null(this.read_pointer(uaddr)?)? {
+                            // ptr is null -> no timespec
```
RalfJung, line 96:
```suggestion
                            // Both ptrs are null -> no timespec
```

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,216 @@
...
+                    // From Linux implemenation:
+                    // No AllocId, or no live allocation at that AllocId.
+                    // Return an error code. (That seems nicer than silently doing something non-intuitive.)
+                    // This means that if an address gets reused by a new allocation,
+                    // we'll use an independent futex queue for this... that seems acceptable.
+                    return this.set_last_error_and_return(LibcError("EFAULT"), dest);
+                };
+                let futex_ref = futex_ref.futex.clone();
+
+                let count = match val{
```
RalfJung, line 177:
```suggestion
                let count = match val {
```
Seems like rustfmt gave up on your code.^^

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,216 @@
...
+                    u64::MAX => usize::MAX, // Preserve MAX because it specifies to wake everyone.
+                    // This fits because we did `read_target_usize`.
```
RalfJung, line 179: This is incorrect, `usize` might be 32bit but target-usize can still be 64bit.

Also, special-casing `u64::MAX` is odd. Let's not do that.
I would suggest you do a saturating cast here, i.e. something like `usize::try_from(val).unwrap_or(usize::MAX)`.

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,216 @@
...
+        // Inner `timespec` must still be valid.
+        let duration = match this.read_timespec(&timespec_place)? {
+            Some(dur) => dur,
+            None => return interp_ok(None),
+        };
+        let flags_place = this.project_field(ut, 1)?;
+        let flags = this.read_scalar(&flags_place)?.to_u32()?;
+        let clock_id_place = this.project_field(ut, 2)?;
+        let clock_id = this.read_scalar(&clock_id_place)?.to_u32()?;
+        interp_ok(Some(UmtxTime { timeout: duration, flags, _clock_id: clock_id }))
```
RalfJung, line 214: Please fully interpret the struct data here:
- Instead of `flags`, have a field `abstime: bool` in the struct. `flags` must be either 0 or UMTX_ABSTIME, all other values are unsupported.
- Interpret the clock ID. This will encode whether to use the system clock or monotonic clock. The logic should probably be the same as [here](https://github.com/rust-lang/rust/blob/562b282fcc8b6ffa3770d30cd40df847ebb56182/src/tools/miri/src/shims/unix/sync.rs#L327-L339).

tests/pass-dep/concurrency/freebsd-futex.rs
```diff
@@ -0,0 +1,170 @@
...
+
+use std::mem::MaybeUninit;
+use std::ptr::{self, addr_of};
+use std::sync::atomic::AtomicU32;
+use std::time::{Duration, Instant};
+use std::{io, thread};
+
+fn wake_nobody() {
+    // TODO: _umtx_op does not return how many threads were woken up
+    // How do i test this?
```
RalfJung, line 12: What do you want to test?

tests/pass-dep/concurrency/freebsd-futex.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+fn wake_dangling() {
```
RalfJung, line 14:
```suggestion

fn wake_dangling() {
```

tests/pass-dep/concurrency/freebsd-futex.rs
```diff
@@ -0,0 +1,170 @@
...
+                ptr::from_ref(&futex).cast_mut().cast(),
+                libc::UMTX_OP_WAIT_UINT_PRIVATE,
+                123,
+                &mut libc::timespec { tv_sec: 0, tv_nsec: 200_000_000 } as *mut _ as *mut _,
+                ptr::null_mut::<libc::c_void>(),
+            ),
+            -1
+        );
+        // man page doesn't document but we set EINVAL for consistency
+        assert_eq!(io::Error::last_os_error().raw_os_error().unwrap(), libc::ETIMEDOUT);
```
RalfJung, line 72: The comment doesn't match the assert. Which one is right?

tests/pass-dep/concurrency/freebsd-futex.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    assert!((200..1000).contains(&start.elapsed().as_millis()));
+}
+fn wait_absolute_timeout() {
```
RalfJung, line 77:
```suggestion

fn wait_absolute_timeout() {
```
In general please take a bit more care of formatting and don't just blindly hope that rustfmt will fix everything. :)

tests/pass-dep/concurrency/freebsd-futex.rs
```diff
@@ -0,0 +1,170 @@
...
+        assert_eq!(
+            libc::_umtx_op(
+                ptr::from_ref(&futex).cast_mut().cast(),
+                libc::UMTX_OP_WAIT_UINT_PRIVATE,
+                123,
+                &mut libc::timespec { tv_sec: 0, tv_nsec: 200_000_000 } as *mut _ as *mut _,
+                ptr::null_mut::<libc::c_void>(),
+            ),
+            -1
+        );
```
RalfJung, line 70: If this tests specifically the relative timeout, please call the test `wait_relative_timeout`.

Also, the test should cover both ways of doing a relative timeout: with and without `_umtx_time`.

tests/pass-dep/concurrency/freebsd-futex.rs
```diff
@@ -0,0 +1,170 @@
...
+    static mut FUTEX: u32 = 0;
+
+    let t = thread::spawn(move || {
+        thread::sleep(Duration::from_millis(200));
+        unsafe {
+            assert_eq!(
+                libc::_umtx_op(
+                    addr_of!(FUTEX) as *mut _,
+                    libc::UMTX_OP_WAKE_PRIVATE,
+                    10, // Wake up 10 threads, but we can't check that we woken up 1.
```
RalfJung, line 135: You could write a test that involves two threads blocked on the futex, and then wake up one of them and check that only one gets woken.

RalfJung (commented):
Sorry for the long wait! These PRs are kinda tricky to review. Overall it looks good but I have a lot of questions. :)

When resolving these comments, please do *not* force-push or amend commits. Instead, make new commits for your new changes. We'll worry about cleaning up the git history later.

src/shims/unix/freebsd/sync.rs
LorrensP-2158466: Because it can be treated as a pointer to `timespec` or as a value equalling to the size of the `umtx_time` struct. I don't know how I should read it at the beginning without converting it back to either one.

src/shims/unix/freebsd/sync.rs
LorrensP-2158466: The function `read_umtx_time` returns `None` if the "inner" `timespec` struct, which is parsed via `read_timespec`, returns `None`.

I will make the comment cleared on what is happening

src/shims/unix/freebsd/sync.rs
LorrensP-2158466: So, looking at the actual implementation, nothing really happens.

It checks if `val` equals the value read from the obj pointer. If this is equal, the thread sleeps; otherwise, the function will exit with zero.

Do you think I should still test this on a real FreeBSD system?

src/shims/unix/freebsd/sync.rs
LorrensP-2158466: Ah, that's the clock ID. I got it!

FreeBSD has a lot of clocks:
`CLOCK_REALTIME` but also `CLOCK_REALTIME_FAST`,...
`CLOCK_MONOTONIC` but also `CLOCK_MONOTONIC_FAST`, ...
And a lot more, so I should only support the above 2, but what about its variants? Just interpret them as the default, or don't support them?

src/shims/unix/freebsd/sync.rs
RalfJung: You can do `read_scalar` here and then later convert the scalar as needed.

src/shims/unix/freebsd/sync.rs
RalfJung: That's an implementation detail of `read_umtx_time`, I don't think it even needs to be mentioned here.

src/shims/unix/freebsd/sync.rs
RalfJung: So, it will return 0 both when the thread was blocked (values are equal) and when it returned immediately (values are inequal)? That is odd, usually a different return value is used for these cases. So yeah would be good to confirm on a real system.

src/shims/unix/freebsd/sync.rs
RalfJung: The proper thing to do is to share the code with [here](https://github.com/rust-lang/miri/blob/9611e78d8a9ba3eae6b3741edccf2c2d63e4eb42/src/shims/time.rs#L39-L76) and have a type like
```rust
enum ClockKind { System, Monotonic }
```
and a function that converts a clock_id to that enum, and use it everywhere. Then we can centrally decide which of the variants it makes sense to support.

But I'm fine with postponing this, and just supporting CLOCK_REALTIME and CLOCK_MONOTONIC for now.

Any other, unknown clock value should throw an "unsupported" error.

src/shims/unix/freebsd/sync.rs
LorrensP-2158466: because i have to handle a signed or an unsigned integer, but you're right, I can just compare at bit level, thanks!

tests/pass-dep/concurrency/freebsd-futex.rs
LorrensP-2158466: comment was a copy paste error, oops... :)

tests/pass-dep/concurrency/freebsd-futex.rs
LorrensP-2158466: the Linux and MacOS tests have something like this:
```rust
// tests/pass-dep/concurrency/linux-futex.rs
fn wake_nobody() {
    let futex = 0;

    // Wake 1 waiter. Expect zero waiters woken up, as nobody is waiting.
    unsafe {
        assert_eq!(libc::syscall(libc::SYS_futex, addr_of!(futex), libc::FUTEX_WAKE, 1), 0);
    }

    // Same, but without omitting the unused arguments.
    unsafe {
        assert_eq!(
            libc::syscall(
                libc::SYS_futex,
                addr_of!(futex),
                libc::FUTEX_WAKE,
                1,
                ptr::null::<libc::timespec>(),
                0usize,
                0,
            ),
            0,
        );
    }
}
```

But I don't know how I can really test this, given _umtx_op doesn't return a value saying how many threads were woken up.

LorrensP-2158466: Update, I used the same approach as you said in your last review comment:
> You could write a test that involves two threads blocked on the futex, and then wake up one of them and check that only one gets woken.

So I have a way I think

src/shims/unix/freebsd/sync.rs
LorrensP-2158466: I tested it out on a freebsd VM with QEMU using this [guide](https://gist.github.com/ctsrc/a1f57933a2cde9abc0f07be12889f97f) and [this](https://ftp2.uk.freebsd.org/pub/FreeBSD/releases/VM-IMAGES/14.2-RELEASE/aarch64/Latest/FreeBSD-14.2-RELEASE-arm64-aarch64.raw.xz) freebsd version.

This is the C code I used:
```c
#define _WANT_UMUTEX
#include <sys/types.h>
#include <sys/umtx.h>
#include <sys/umtx.h>
#include <errno.h>
#include <string.h>
#include <unistd.h>

int main(){
    int futex = 123;
    int expected = 456;

    int ret = _umtx_op(&futex, UMTX_OP_WAIT, expected, NULL, NULL);
    printf("return value is: %d\n", ret);
    return 0;
}
```

This prints:
```
return value is: 0
```

This is weird, but confirms it no?

src/shims/unix/freebsd/sync.rs
LorrensP-2158466: This will probably also work:
```rust
let count = val.min(usize::MAX as u64) as usize;
```

But then clippy yells at me:
```
warning: casting `u64` to `usize` may truncate the value on targets with 32-bit wide pointers
   --> src/shims/unix/freebsd/sync.rs:164:29
    |
164 |                 let count = val.min(usize::MAX as u64) as usize;
    |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: if this is intentional allow the lint with `#[allow(clippy::cast_possible_truncation)]` ...
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#cast_possible_truncation
```

but this can never happen here... I'll just stick to your solution :)

src/shims/unix/freebsd/sync.rs
LorrensP-2158466: I still haven't found it, even after looking at the source code. But what I found really weird is the fact that the syscall doesn't take a normal `timespec` argument into account:
```c
static int
__umtx_op_wait(struct thread *td, struct _umtx_op_args *uap,
    const struct umtx_copyops *ops)
{
	struct _umtx_time timeout, *tm_p;
	int error;

	if (uap->uaddr2 == NULL)
		tm_p = NULL;
	else {
		error = ops->copyin_umtx_time(
		    uap->uaddr2, (size_t)uap->uaddr1, &timeout);
		if (error != 0)
			return (error);
		tm_p = &timeout;
	}
	return (do_wait(td, uap->obj, uap->val, tm_p, ops->compat32, 0));
}
```
`do_wait` gets `tm_p` as the only argument which specifies a timeout. But if `uaddr2` is NULL, which is the case when you only pass in `timespec` as a timeout (in the manual), then this `tm_p` is also NULL. Which has no timeout then?

Testing this on the FReeBSD VM there is indeed no timeout?
```c
#include <sys/types.h>
#include <sys/umtx.h>
#include <errno.h>
#include <string.h>
#include <unistd.h>

int main(){
        int futex = 123;
        int expected = 123;

        struct timespec ts;

        ts.tv_sec = 0;
        ts.tv_nsec = 200000;

        int ret = _umtx_op(&futex, UMTX_OP_WAIT, expected, &ts, NULL);

        printf("return value is: %d\n", ret);
        printf("errno: %s\n", strerror(errno));
        return 0;
}
```
This hangs forever.

What should we do here? We can't really follow the manual because it doesn't specify the Clock to use, and if we use the source code, it doesn't follow the manual.

src/shims/unix/freebsd/sync.rs
RalfJung: Yeah looks like it. Odd.

Cc @m-ou-se to get another set of eyes on this (you started using these functions in https://github.com/rust-lang/rust/pull/96510). The question is, does _umtx_op really not distinguish between "comparison successful, we want to sleep and were woken up again" and "comparison failed"? It seems to always return 0 in both cases.

src/shims/unix/freebsd/sync.rs
RalfJung: > DISREGARD anything said in this comment, I just now noticed that I completely misread the docs a bunch of times, I finally get it...

So what is the answer then?

RalfJung:
@rustbot author

src/shims/unix/freebsd/sync.rs
RalfJung: Well, put the source for that claim in a comment please. :)

LorrensP-2158466:
Those are not my commits, I hope I didn't do something wrong.

As for the changes, I:
- changed the handling of `uaddr` and `uaddr2` to now being correct
- fixed/updated/added tests to correctly test behaviour of the syscall

RalfJung:
Yeah something is definitely wrong with the PR, I can't review it in this state. Please make sure only your commits are on this branch.

@rustbot author

LorrensP-2158466:
I think i fixed it, only my commits are present on this branch

@rustbot ready

RalfJung:
In the future, please avoid rebasing unless there are conflicts. Github is a pretty bad tool for code review, and rebasing throws it off entirely, causing a bunch of extra work on my side.

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -55,6 +56,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.write_scalar(res, dest)?;
             }
 
+            // Futex
```
RalfJung, line 59:
```suggestion
            // Synchronization primitives
```

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,238 @@
...
+
+impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
+pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    /// Implementation of the FreeBSD [`_umtx_op`](https://man.freebsd.org/cgi/man.cgi?query=_umtx_op&sektion=2&manpath=FreeBSD+14.2-RELEASE+and+Ports) syscall.
+    /// This is used for futex operations on FreeBSD.
+    ///
+    /// `obj`: a pointer to the futex object (can be a lot of things, mostly *AtomicU32)
+    /// `op`: the futex operation to run
+    /// `val`: the current value of the object as a `c_long` (for wait/wake)
+    /// `uaddr`: pointer to optional parameter (mostly timeouts)
```
RalfJung, line 27: Given that this is now read as `target_usize`, that does not seem right?

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,238 @@
...
+                    let futex_ref = this
+                        .get_sync_or_init(obj.ptr(), |_| FreeBsdFutex { futex: Default::default() })
+                        .unwrap()
+                        .futex
+                        .clone();
+
+                    // From the manual:
+                    // The timeout is specified by passing either the address of `struct timespec`, or its
+                    // extended  variant, `struct _umtx_time`, as the `uaddr2` argument of _umtx_op().
+                    // They are distinguished by the `uaddr` value, which  must be equal
```
RalfJung, line 81:
```suggestion
                    // They are distinguished by the `uaddr` value, which must be equal
```

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,238 @@
...
+                    // to the size of the structure pointed to by `uaddr2`, casted to uintptr_t.
+                    let timeout = if this.ptr_is_null(uaddr2)? {
+                        // no timeout parameter
+                        None
+                    } else {
+                        if uaddr == umtx_time_layout.size.bytes() {
```
RalfJung, line 87: Please add an assert that the two structs have different sizes. Otherwise this won't work...

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,238 @@
...
+                            let duration = match this.read_timespec(&timespec)? {
+                                Some(duration) => duration,
+                                None => {
+                                    return this
+                                        .set_last_error_and_return(LibcError("EINVAL"), dest);
+                                }
+                            };
+
+                            // From the FreeBSD source code:
+                            // In the function `umtx_copyin_umtx_time` in kern_umtx.c, _clock_id is set to CLOCK_REALTIME when using `timespec`
```
RalfJung, line 121: So the docs don't say anywhere which clock is used?

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,238 @@
...
+                            } else {
+                                TimeoutAnchor::Relative
+                            };
+
+                            Some((umtx_time.timeout_clock, anchor, umtx_time.timeout))
+                        } else if uaddr == timespec_layout.size.bytes() {
+                            // RealTime clock can't be used in isolation mode.
+                            this.check_no_isolation("`_umtx_op` with `TimeoutClock::RealTime`")?;
+
+                            // `uaddr2` points to a `struct timespec`
```
RalfJung, line 110:
```suggestion
                            // `uaddr2` points to a `struct timespec`.
```

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,238 @@
...
+                    // The timeout is specified by passing either the address of `struct timespec`, or its
+                    // extended  variant, `struct _umtx_time`, as the `uaddr2` argument of _umtx_op().
+                    // They are distinguished by the `uaddr` value, which  must be equal
+                    // to the size of the structure pointed to by `uaddr2`, casted to uintptr_t.
+                    let timeout = if this.ptr_is_null(uaddr2)? {
+                        // no timeout parameter
+                        None
+                    } else {
+                        if uaddr == umtx_time_layout.size.bytes() {
+                            // `uaddr2` points to a `struct _umtx_time`
```
RalfJung, line 88:
```suggestion
                            // `uaddr2` points to a `struct _umtx_time`.
```

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,238 @@
...
+                        timeout,
+                        callback!(
+                            @capture<'tcx> {
+                                dest: MPlaceTy<'tcx>,
+                            }
+                            |ecx, unblock: UnblockKind| match unblock {
+                                UnblockKind::Ready => {
+                                    // From the manual:
+                                    // If successful, all requests, except UMTX_SHM_CREAT and  UMTX_SHM_LOOKUP
+                                    // sub-requests  of	 the  UMTX_OP_SHM  request,  will  return  zero.
```
RalfJung, line 141:
```suggestion
                                    // sub-requests of the  UMTX_OP_SHM  request, will return zero.
```

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,238 @@
...
+                // `_umtx_op` doesn't return the amount of woken threads.
+                let _woken = this.futex_wake(
+                    &futex_ref,
+                    u32::MAX, // we set the bitset to include all bits
+                    count,
+                )?;
+
+                // From the manual:
+                // If successful, all requests, except UMTX_SHM_CREAT and  UMTX_SHM_LOOKUP
+                // sub-requests  of	 the  UMTX_OP_SHM  request,  will  return  zero.
```
RalfJung, line 188:
```suggestion
                // sub-requests of the UMTX_OP_SHM  request, will return zero.
```

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,238 @@
...
+                            let anchor = if umtx_time.abs_time {
+                                TimeoutAnchor::Absolute
+                            } else {
+                                TimeoutAnchor::Relative
+                            };
+
+                            Some((umtx_time.timeout_clock, anchor, umtx_time.timeout))
+                        } else if uaddr == timespec_layout.size.bytes() {
+                            // RealTime clock can't be used in isolation mode.
+                            this.check_no_isolation("`_umtx_op` with `TimeoutClock::RealTime`")?;
```
RalfJung, line 108:
```suggestion
                            this.check_no_isolation("`_umtx_op` with `timespec` timeout")?;
```

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,238 @@
...
+        interp_ok(Some(UmtxTime { timeout: duration, abs_time: abs_time_flag, timeout_clock }))
+    }
+
+    /// Translate raw FreeBSD clockid to a Miri TimeoutClock
+    fn translate_umtx_time_clock_id(&mut self, raw_id: i32) -> InterpResult<'tcx, TimeoutClock> {
+        let this = self.eval_context_mut();
+
+        let timeout = if raw_id == this.eval_libc_i32("CLOCK_REALTIME") {
+            // RealTime clock can't be used in isolation mode.
+            this.check_no_isolation("`_umtx_op` with `TimeoutClock::RealTime`")?;
```
RalfJung, line 229:
```suggestion
            this.check_no_isolation("`_umtx_op` with `CLOCK_REALTIME` timeout")?;
```

src/shims/unix/freebsd/sync.rs
```diff
@@ -0,0 +1,238 @@
...
+        let abs_time_flag = flags == abs_time;
+
+        let clock_id_place = this.project_field(ut, 2)?;
+        let clock_id = this.read_scalar(&clock_id_place)?.to_i32()?;
+        let timeout_clock = this.translate_umtx_time_clock_id(clock_id)?;
+
+        interp_ok(Some(UmtxTime { timeout: duration, abs_time: abs_time_flag, timeout_clock }))
+    }
+
+    /// Translate raw FreeBSD clockid to a Miri TimeoutClock
```
RalfJung, line 223:
```suggestion
    /// Translate raw FreeBSD clockid to a Miri TimeoutClock.
    /// FIXME: share this code with the pthread and clock_gettime shims.
```

tests/pass-dep/concurrency/freebsd-futex.rs
```diff
@@ -0,0 +1,250 @@
...
+    });
+    let mut timeout = libc::timespec { tv_sec: 0, tv_nsec: 400_000_000 };
+    let timeout_size_arg =
+        ptr::without_provenance_mut::<libc::c_void>(mem::size_of::<libc::timespec>());
+    unsafe {
+        assert_eq!(
+            libc::_umtx_op(
+                addr_of!(FUTEX) as *mut _,
+                libc::UMTX_OP_WAIT_UINT_PRIVATE,
+                0,
```
RalfJung, line 40: Is this the number of threads to wake?
What's the point of the sleep...? You disabled preemption, so this can be tested without `thread::sleep`.

LorrensP-2158466:
Sorry ):, you said that a rebase would probably fix it, and it sort of did. I had never seen something like that before and could not find anything that resembled that. Hopefully it doesn't cause to much problems.

tests/pass-dep/concurrency/freebsd-futex.rs
```diff
@@ -0,0 +1,250 @@
...
+            );
+            io::Error::last_os_error().raw_os_error().unwrap() == libc::ETIMEDOUT
+        }
+    }
+
+    let t1 = thread::spawn(waiter);
+    let t2 = thread::spawn(waiter);
+
+    // Wake up 1 thread and make sure the other is still waiting
+    thread::sleep(Duration::from_millis(200));
```
RalfJung, line 223: Again no `sleep` needed. You should `yield_now` here once, then the two threads will be scheduled and block on the futex. Then you wake up one. There shouldn't be any `sleep` in this test.

tests/pass-dep/concurrency/freebsd-futex.rs
```diff
@@ -0,0 +1,250 @@
...
+                umtx_timeout_ptr as *mut _,
+            ),
+            -1
+        );
+        assert_eq!(io::Error::last_os_error().raw_os_error().unwrap(), libc::ETIMEDOUT);
+    }
+    assert!((200..1000).contains(&start.elapsed().as_millis()));
+}
+
+fn wait_wake() {
```
RalfJung, line 197:
```suggestion
fn wait_wake_one() {
```

tests/pass-dep/concurrency/freebsd-futex.rs
```diff
@@ -0,0 +1,250 @@
...
+    assert!(t1_woke_up ^ t2_woke_up, "Expected 1 thread to wake up");
+}
+
+fn main() {
+    wake_nobody();
+    wake_dangling();
+    wait_wrong_val();
+    wait_relative_timeout();
+    wait_absolute_timeout();
+    wait_wake();
```
RalfJung, line 249: `wait_wake`  and `wait_nobody` are both testing the "wake count", right? It might make sense to test 0 as a special case, but conceptually similar tests should still go together, not be at opposite ends of the test list (and the file).

RalfJung (commented):
@rustbot author

RalfJung:
> Sorry ):, you said that a rebase would probably fix it, and it sort of did. I had never seen something like that before and could not find anything that resembled that. Hopefully it doesn't cause to much problems.

Yeah you must have done a  rebase or a merge first to even cause the problem that the rebase then fixed. That first step is what I referred to. :)

src/shims/unix/freebsd/sync.rs
LorrensP-2158466: I wanted to keep the doc as "global" as possible, even this doc is wrong given the entire manual. `uaddr` and `uaddr2` can be many things, pointers to structs, values, keys, .... 

Is it better to specify in the `_umtx_op` documentation that only `WAIT` and `WAKE` are implemented so the comments for the `uaddr` values can be more specific?

src/shims/unix/freebsd/sync.rs
LorrensP-2158466: I haven't found anything in the docs that said to use a specific Clock ID. Only thing I did found that talked about the Real Time clock whas the operation ``UMTX_OP_CV_WAIT:
From the docs:

>Optionally, a timeout for the request may be specified.	Unlike
>other requests, the timeout value is specified  directly	 by  a
>struct  timespec,  pointed  to  by the uaddr2 argument.	If the
>CVWAIT_CLOCKID flag is provided,	the  timeout  uses  the	 clock
>from  the  c_clockid  member of the struct ucond, pointed to by
>obj argument.  Otherwise, CLOCK_REALTIME	is used, regardless of
>the  clock  identifier  possibly	 specified   in	  the	struct
>_umtx_time.   If	the CVWAIT_ABSTIME flag	is supplied, the time-
>out specifies absolute time value, otherwise it denotes a rela-
>tive time interval.

But this is not what we are doing...

tests/pass-dep/concurrency/freebsd-futex.rs
LorrensP-2158466: That indeed makes more sense. I just followed the same pattern as the tests of MacOS and Linux.

Currently, I have a `wait_wake` test function, which has 2 tests: `wake_nobody` and `wake_one`. Should I also add a `wake_multiple` test?

src/shims/unix/freebsd/sync.rs
RalfJung: Well at least the docs shouldn't say anything wrong, such as saying something is a pointer which is clearly used as an integer. :)  So maybe
```
`uaddr`: `op`-specific optional parameter, pointer-sized integer or pointer
```
and then probably the same for `uaddr2`.

src/shims/unix/freebsd/sync.rs
RalfJung: It really seems entirely undocumented. And according to https://github.com/golang/go/issues/17168#issuecomment-250235271 it actually changed, it used to use the monotonic clock but then switch to REALTIME. Wild.

Please also add a link to that github discussion, and say something like:
> FreeBSD does not seem to document which clock is used when the timeout is passed as a `timespec*`. Based on discussions online and the source code, it seems to default to CLOCK_REALTIME, so that's what we also do.

tests/pass-dep/concurrency/freebsd-futex.rs
RalfJung: > I just followed the same pattern as the tests of MacOS and Linux.

Fair, sorry for having messy existing tests.^^

> Currently, I have a wait_wake test function, which has 2 tests: wake_nobody and wake_one. Should I also add a wake_multiple test?

I'd say instead of `wake_one`, have "wake 2 out of 3". That should cover most of the interesting corner cases.

src/shims/unix/freebsd/sync.rs
LorrensP-2158466: Yeah that's fair, will do!

tests/pass-dep/concurrency/freebsd-futex.rs
```diff
@@ -0,0 +1,257 @@
...
+                        libc::UMTX_OP_WAKE_PRIVATE,
+                        0, // wake up 0 waiters
+                        ptr::null_mut::<libc::c_void>(),
+                        ptr::null_mut::<libc::c_void>(),
+                    ),
+                    0
+                );
+            }
+        });
+        thread::yield_now();
```
RalfJung, line 31: This means the "wake 0 threads" occurs before the main thread even goes to sleep. Doesn't that defeat the purpose of the test? The test should ensure that nobody is woken up even if there's someone waiting.

tests/pass-dep/concurrency/freebsd-futex.rs
```diff
@@ -0,0 +1,257 @@
...
+        let mut timeout = libc::timespec { tv_sec: 0, tv_nsec: 400_000_000 };
```
RalfJung, line 32:
```suggestion
        // 10ms should be enough.
        let mut timeout = libc::timespec { tv_sec: 0, tv_nsec: 10_000_000 };
```

tests/pass-dep/concurrency/freebsd-futex.rs
```diff
@@ -0,0 +1,257 @@
...
+            );
+            // Main thread did not get woken up, so it timed out.
+            assert_eq!(io::Error::last_os_error().raw_os_error().unwrap(), libc::ETIMEDOUT);
+        }
+
+        waker.join().unwrap();
+    }
+
+    fn wake_two_of_three() {
+        // We create 2 threads that wait on a futex with a 500ms timeout.
```
RalfJung, line 54: A 100ms timeout should be enough hopefully. I don't want this to become the slowest of all tests just due to all the waiting. :)

tests/pass-dep/concurrency/freebsd-futex.rs
```diff
@@ -0,0 +1,257 @@
...
+                // Return true if this thread woke up.
+                io::Error::last_os_error().raw_os_error().unwrap() != libc::ETIMEDOUT
+            }
+        }
+
+        let t1 = thread::spawn(waiter);
+        let t2 = thread::spawn(waiter);
+        let t3 = thread::spawn(waiter);
+
+        thread::yield_now();
```
RalfJung, line 80:
```suggestion
        // Run all the waiters, so they all go to sleep.
        thread::yield_now();
```

RalfJung (commented):
Just some minor nits for the test. :)

tests/pass-dep/concurrency/freebsd-futex.rs
LorrensP-2158466: Yeah, fair. Half a second might be a bit extreme.

tests/pass-dep/concurrency/freebsd-futex.rs
LorrensP-2158466: Yes you're right, I'll fix it.

LorrensP-2158466:
The tests should be ok, at least on my machine.

@rustbot ready

RalfJung:

@rustbot author

LorrensP-2158466:
I learned something new about git today, I hope this is what you meant :).

@rustbot ready

RalfJung:
Looks good, thanks :)

PR #4212, opened 2025-02-26
Support F_GETFL and F_SETFL for fcntl

tiif:
This PR supports ``F_SETFL`` and ``F_GETFL`` flags for ``fcntl``. In this implementation, ``F_SETFL`` can only set ``O_NONBLOCK`` flag (and file access mode flag like ``O_RDONLY``, but that will be ignored).

>  If setfl is called while a fd is blocking, that fd will not be affected by the new flag value and will continue blocking like normal until a read/write wakes it up (it acts like it never knew the setfl has happened). But any operation after setfl will see the new flag value.

The interaction between these ``fcntl`` operations and blocking fd is summarised [here](https://github.com/rust-lang/miri/issues/4119#issuecomment-2659551156). 

Fixes rust-lang/miri#4119

### DIFF
--- a/src/shims/files.rs
+++ b/src/shims/files.rs
@@ -196,6 +196,20 @@ pub trait FileDescription: std::fmt::Debug + FileDescriptionExt {
     fn as_unix<'tcx>(&self, _ecx: &MiriInterpCx<'tcx>) -> &dyn UnixFileDescription {
         panic!("Not a unix file descriptor: {}", self.name());
     }
+
+    /// Implementation of fcntl(F_GETFL) for this FD.
+    fn get_flags<'tcx>(&self, _ecx: &mut MiriInterpCx<'tcx>) -> InterpResult<'tcx, Scalar> {
+        throw_unsup_format!("fcntl: {} is not supported for F_GETFL", self.name());
+    }
+
+    /// Implementation of fcntl(F_SETFL) for this FD.
+    fn set_flags<'tcx>(
+        &self,
+        _flag: i32,
+        _ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, Scalar> {
+        throw_unsup_format!("fcntl: {} is not supported for F_SETFL", self.name());
+    }
 }
 
 impl FileDescription for io::Stdin {
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -141,6 +141,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let f_getfd = this.eval_libc_i32("F_GETFD");
         let f_dupfd = this.eval_libc_i32("F_DUPFD");
         let f_dupfd_cloexec = this.eval_libc_i32("F_DUPFD_CLOEXEC");
+        let f_getfl = this.eval_libc_i32("F_GETFL");
+        let f_setfl = this.eval_libc_i32("F_SETFL");
 
         // We only support getting the flags for a descriptor.
         match cmd {
@@ -174,6 +176,25 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 } else {
                     this.set_last_error_and_return_i32(LibcError("EBADF"))
                 }
+            }
+            cmd if cmd == f_getfl => {
+                // Check if this is a valid open file descriptor.
+                let Some(fd) = this.machine.fds.get(fd_num) else {
+                    return this.set_last_error_and_return_i32(LibcError("EBADF"));
+                };
+
+                fd.get_flags(this)
+            }
+            cmd if cmd == f_setfl => {
+                // Check if this is a valid open file descriptor.
+                let Some(fd) = this.machine.fds.get(fd_num) else {
+                    return this.set_last_error_and_return_i32(LibcError("EBADF"));
+                };
+
+                let [flag] = check_min_vararg_count("fcntl(fd, F_SETFL, ...)", varargs)?;
+                let flag = this.read_scalar(flag)?.to_i32()?;
+
+                fd.set_flags(flag, this)
             }
             cmd if this.tcx.sess.target.os == "macos"
                 && cmd == this.eval_libc_i32("F_FULLFSYNC") =>
--- a/src/shims/unix/unnamed_socket.rs
+++ b/src/shims/unix/unnamed_socket.rs
@@ -20,6 +20,16 @@ use crate::*;
 /// be configured in the real system.
 const MAX_SOCKETPAIR_BUFFER_CAPACITY: usize = 212992;
 
+#[derive(Debug, PartialEq)]
+enum AnonSocketType {
+    // Either end of the socketpair fd.
+    Socketpair,
+    // Read end of the pipe.
+    PipeRead,
+    // Write end of the pipe.
+    PipeWrite,
+}
+
 /// One end of a pair of connected unnamed sockets.
 #[derive(Debug)]
 struct AnonSocket {
@@ -40,7 +50,10 @@ struct AnonSocket {
     /// A list of thread ids blocked because the buffer was full.
     /// Once another thread reads some bytes, these threads will be unblocked.
     blocked_write_tid: RefCell<Vec<ThreadId>>,
-    is_nonblock: bool,
+    /// Whether this fd is non-blocking or not.
+    is_nonblock: Cell<bool>,
+    // Differentiate between different AnonSocket fd types.
+    fd_type: AnonSocketType,
 }
 
 #[derive(Debug)]
@@ -63,7 +76,10 @@ impl AnonSocket {
 
 impl FileDescription for AnonSocket {
     fn name(&self) -> &'static str {
-        "socketpair"
+        match self.fd_type {
+            AnonSocketType::Socketpair => "socketpair",
+            AnonSocketType::PipeRead | AnonSocketType::PipeWrite => "pipe",
+        }
     }
 
     fn close<'tcx>(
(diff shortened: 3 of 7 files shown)


src/shims/unix/unnamed_socket.rs
```diff
@@ -22,7 +22,7 @@ const MAX_SOCKETPAIR_BUFFER_CAPACITY: usize = 212992;
 
 /// One end of a pair of connected unnamed sockets.
 #[derive(Debug)]
-struct AnonSocket {
+pub struct AnonSocket {
```
tiif, line 25: This is needed for 
```rust
// We only support F_GETFL for socketpair and pipe.
let anonsocket_fd = fd.downcast::<AnonSocket>().ok_or_else(|| {
    err_unsup_format!("fcntl: only socketpair / pipe are supported for F_SETFL")
})?;
```

src/shims/unix/fd.rs
```diff
@@ -175,6 +178,47 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                // We only support F_SETFL for socketpair and pipe.
+                let anonsocket_fd = fd.downcast::<AnonSocket>().ok_or_else(|| {
+                    err_unsup_format!("fcntl: only socketpair / pipe are supported for F_GETFL")
+                })?;
+
+                let cmd_name = "fcntl(fd, F_SETFL, ...)";
+                let [flag] = check_min_vararg_count(cmd_name, varargs)?;
+                let flag = this.read_scalar(flag)?.to_i32()?;
+
+                // FIXME: File access mode and file creation flags should be ignored.
```
tiif, line 214: From [the man page](https://man7.org/linux/man-pages/man2/fcntl.2.html):
```
       F_SETFL (int)
              Set the file status flags to the value specified by arg.
              File access mode (O_RDONLY, O_WRONLY, O_RDWR) and file
              creation flags (i.e., O_CREAT, O_EXCL, O_NOCTTY, O_TRUNC)
              in arg are ignored.
```
There are quite a few file access mode and file creation flags, would prefer to open an issue for this and do it later.

tiif (commented):
@rustbot ready

src/shims/unix/fd.rs
```diff
@@ -175,6 +178,47 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             }
+            cmd if cmd == f_getfl => {
+                // Check if this is a valid open file descriptor.
+                let Some(fd) = this.machine.fds.get(fd_num) else {
+                    return this.set_last_error_and_return_i32(LibcError("EBADF"));
+                };
+
+                // We only support F_GETFL for socketpair and pipe.
+                let anonsocket_fd = fd.downcast::<AnonSocket>().ok_or_else(|| {
+                    err_unsup_format!("fcntl: only socketpair / pipe are supported for F_SETFL")
```
RalfJung, line 189:
```suggestion
                    err_unsup_format!("fcntl: only socketpair / pipe are supported for F_GETFL")
```

src/shims/unix/fd.rs
```diff
@@ -175,6 +178,47 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                })?;
+
+                if anonsocket_fd.is_nonblock() {
+                    // socketpair's SOCK_NONBLOCK and pipe's O_NONBLOCK have the same value.
```
RalfJung, line 193: The fcntl docs seem to anyway indicate that we always return the `O_*` flags here. So this fact should not matter.

If it actually did matter, we should have an `assert_eq!` here.

src/shims/unix/fd.rs
```diff
@@ -175,6 +178,47 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let Some(fd) = this.machine.fds.get(fd_num) else {
+                    return this.set_last_error_and_return_i32(LibcError("EBADF"));
+                };
+
+                // We only support F_SETFL for socketpair and pipe.
+                let anonsocket_fd = fd.downcast::<AnonSocket>().ok_or_else(|| {
+                    err_unsup_format!("fcntl: only socketpair / pipe are supported for F_GETFL")
+                })?;
+
+                let cmd_name = "fcntl(fd, F_SETFL, ...)";
```
RalfJung, line 210: Please inline this `let`.

src/shims/unix/fd.rs
```diff
@@ -175,6 +178,47 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let [flag] = check_min_vararg_count(cmd_name, varargs)?;
+                let flag = this.read_scalar(flag)?.to_i32()?;
+
+                // FIXME: File access mode and file creation flags should be ignored.
+                if flag == this.eval_libc_i32("O_NONBLOCK") {
+                    anonsocket_fd.set_nonblock();
+                } else {
+                    throw_unsup_format!("fcntl: only O_NONBLOCK are supported for F_GETFL")
+                }
```
RalfJung, line 219: This should also support *un*-setting the `O_NONBLOCK` flag, right? I assume that's what happens on `flag == 0`.

src/shims/unix/unnamed_socket.rs
```diff
@@ -40,7 +40,7 @@ struct AnonSocket {
     /// A list of thread ids blocked because the buffer was full.
     /// Once another thread reads some bytes, these threads will be unblocked.
     blocked_write_tid: RefCell<Vec<ThreadId>>,
-    is_nonblock: bool,
+    is_nonblock: Cell<bool>,
```
RalfJung, line 43:
```suggestion
    /// Whether this is a non-blocking socket or not.
    is_nonblock: Cell<bool>,
```

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -451,3 +454,54 @@ fn test_ioctl() {
...
+
+    let res = unsafe { libc::fcntl(fds[0], libc::F_GETFL) };
+    assert_eq!(res, libc::O_NONBLOCK);
+}
+
+// Test the behaviour of setfl/getfl when a fd is blocking.
+// The expected execution is:
+// 1. Main thread blocks on fds[0] `read`.
+// 2. Thread 1 set O_NONBLOCK  flag on fds[0], then check the value of F_GETFL.
+// 3. Thread 2 writes to fds[1] to unblock the `read` in main thread.
```
RalfJung, line 481: Why is this done by 2 different threads?

RalfJung (commented):
Here's the first round of comments. Sorry for the long wait!

@rustbot author

src/shims/unix/fd.rs
tiif: Good call!

I just realised this is indeed how tokio set their fd to non-blocking. https://github.com/tokio-rs/tokio/blob/817fa605ee6a2549fe8e6057ec23a8309d42d2e9/tokio/src/net/unix/pipe.rs#L1432

src/shims/unix/fd.rs
tiif: True, ``F_GETFL`` returns file status flag, and only ``O_*`` is considered file status flag.

tests/pass-dep/libc/libc-fs.rs
tiif: Yeah this can be done in one thread. I modified this test to use one thread.

tiif:
On the latest commit [b22019d](https://github.com/rust-lang/miri/pull/4212/commits/b22019d85a251d4301705b95d8df93db6431a890), I made all the tests in this PR to use isolation mode.

src/shims/unix/fd.rs
```diff
@@ -175,6 +178,45 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let Some(fd) = this.machine.fds.get(fd_num) else {
+                    return this.set_last_error_and_return_i32(LibcError("EBADF"));
+                };
+
+                // We only support F_GETFL for socketpair and pipe.
+                let anonsocket_fd = fd.downcast::<AnonSocket>().ok_or_else(|| {
+                    err_unsup_format!("fcntl: only socketpair / pipe are supported for F_GETFL")
+                })?;
+
+                if anonsocket_fd.is_nonblock() {
```
RalfJung, line 192: Does this match real implementations -- only returning O_NONBLOCK and no other flag?

tiif: I overlooked something, we need to return ``O_RDWR`` for ``socketpair``, ``O_RDONLY`` and ``O_WRONLY`` for ``pipe``. So we need to add something to differentiate between ``socketpair`` and ``pipe``'s reading / writing end here.

RalfJung: Probably we should have `get_flags` and `set_flags` methods on the FD trait so this logic can live with the FD implementation.

When a socketpair's peer FD has been closed, does it become RDONLY?

tiif: > When a socketpair's peer FD has been closed, does it become RDONLY?

``fcntl(F_GETFL)`` will still return ``O_RDWR`` even though the subsequent write will error with ``EPIPE``.

RalfJung: Okay, then please implement that and add a comment explaining why.

tiif:
Just a heads up, this PR will stall for a while due to exam prep, but I will be back on it after rust week.

tiif:
All feedbacks have been addressed :)

@rustbot ready

src/shims/unix/unnamed_socket.rs
```diff
@@ -110,6 +126,71 @@ impl FileDescription for AnonSocket {
     fn as_unix<'tcx>(&self, _ecx: &MiriInterpCx<'tcx>) -> &dyn UnixFileDescription {
         self
     }
+
+    fn get_flags<'tcx>(&self, ecx: &mut MiriInterpCx<'tcx>) -> InterpResult<'tcx, Scalar> {
+        let mut flags = 0;
+
+        // Get flag for file access mode.
+        // The flag for both socketpair and pipe will remain the same even when the peer
+        // fd is closed.
```
RalfJung, line 135:
```suggestion
        // fd is closed, so we need to look at the original type of this socket, not at whether
        // the peer socket still exists.
```

src/shims/files.rs
```diff
@@ -196,6 +196,20 @@ pub trait FileDescription: std::fmt::Debug + FileDescriptionExt {
     fn as_unix<'tcx>(&self, _ecx: &MiriInterpCx<'tcx>) -> &dyn UnixFileDescription {
         panic!("Not a unix file descriptor: {}", self.name());
     }
+
+    /// Helper function for fcntl(F_GETFL) to get fd's flag value.
+    fn get_flags<'tcx>(&self, _ecx: &mut MiriInterpCx<'tcx>) -> InterpResult<'tcx, Scalar> {
+        throw_unsup_format!("cannot get flag value for {}", self.name());
+    }
```
RalfJung, line 203: Why does this function exist if you anyway downcast the FD type in the shim?

Seems like you already have everything you need to make this work on all FD types, just most of them default to throwing an error?

src/shims/unix/unnamed_socket.rs
```diff
@@ -110,6 +126,71 @@ impl FileDescription for AnonSocket {
...
+        if flag & o_wronly == o_wronly {
+            flag &= !o_wronly;
+        }
+        if flag & o_rdwr == o_rdwr {
+            flag &= !o_rdwr;
+        }
+
+        // Throw error if there is any unsupported flag.
+        if flag != 0 {
+            throw_unsup_format!("fcntl: only O_NONBLOCK is supported for F_SETFL")
```
RalfJung, line 189:
```suggestion
            throw_unsup_format!("fcntl: only O_NONBLOCK is supported for F_SETFL on anonsockets and pipes")
```

RalfJung (commented):
Thanks. :)
@rustbot author

src/shims/files.rs
tiif: I think the downcast logic inside the shim is no longer needed. It should be possible to directly call ``get_flags`` on all fd types, and most of them default to error, yes.

src/shims/unix/unnamed_socket.rs
```diff
@@ -110,6 +126,71 @@ impl FileDescription for AnonSocket {
...
+            self.is_nonblock.set(true);
+            flag &= !o_nonblock;
+        } else {
+            self.is_nonblock.set(false);
+        }
+
+        // Ignore all file access mode flags.
+        if flag & o_rdonly == o_rdonly {
+            flag &= !o_rdonly;
+        }
```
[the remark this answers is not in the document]
tiif: With the set flags we currently support here, this is fine. A hypothetical scenario that this won't work is if we happen to have flag values ``0b1011`` and ``0b0111``, then unsetting the flag without checking if it actually exists might lead to buggy behaviour.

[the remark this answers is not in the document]
RalfJung: Flags typically only set a single bit, otherwise they aren't flags in the usual sense anyway. Like, in your example, even if we did check first the result would have odd cross-flag interactions.

src/shims/unix/unnamed_socket.rs
RalfJung: Ah, right :)

RalfJung (commented):
We're almost there, just a few nits. :)
@rustbot author

RalfJung:
You know you can land multiple GH suggestions as a single commit? :) Just go to the diff view, there's a button for that.

tiif:
Ohh I didn't know that! Clicking one by one is indeed a bit annoying because the webpage will refresh for every commit.

RalfJung:

@rustbot author

tiif:
Done, thanks for the review :)

@rustbot ready

PR #4219, opened 2025-03-05
Use `size_of` from the prelude instead of imported

thaliaarchi:
It was added to all preludes in Rust 1.80.


RalfJung:
This only changes a benchmark file. We don't usually keep our benchmarks updated to the latest coding style because why would we.

But now you already did the change and I already spent time looking at it so :shrug:

PR #4223, opened 2025-03-07
Show interpreter backtrace error on Ctrl+C

### LINKED ISSUE #3116: Show interpreter backtrace on Ctrl-C
It would be nice if we could show where in the interpreted program we were running upon Ctrl-C. We have a `ctrlc` handler so this shouldn't even be that hard to do...

Stypox:
Closes #3116 by showing the following error message and interpreter backtrace on Ctrl+C:
```
error: interpretation was interrupted                                                                                                                                                            
 --> a.rs:2:5
  |
2 |     loop {}
  |     ^^^^^^^ interpretation was interrupted
  |
  = note: BACKTRACE:
  = note: inside `main` at a.rs:2:5: 2:12

```

The contents of `a.rs` were:
```rust
fn main() {
    loop {}
}
```

No new tests were added because it seems unfeasible, see the discussion in https://github.com/rust-lang/miri/issues/3116#issuecomment-2705134484

### DIFF
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -1150,7 +1150,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         loop {
             if CTRL_C_RECEIVED.load(Relaxed) {
                 this.machine.handle_abnormal_termination();
-                std::process::exit(1);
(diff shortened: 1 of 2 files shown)


oli-obk (approved):

RalfJung:
Thanks :)

Issue #3266, opened 2024-01-13
Detecting races when releasing Tree Borrows protectors?


Imberflur:
IIUC in tree borrows protectors generate implicit reads when they are released. However, the data race detector doesn't seem to be aware of them. Does it make sense for these to be accesses for the data race model?

This looks like a similar issue https://github.com/rust-lang/miri/issues/2648

Here is an example I tested with `MIRIFLAGS="-Zmiri-tree-borrows" cargo miri run`.
```rs
use core::sync::atomic::AtomicBool;
use core::sync::atomic::Ordering;
use std::alloc;

static FLAG: AtomicBool = AtomicBool::new(false);

fn protector_data_race(v: &u8) {
    println!("{v}");
    FLAG.store(true, Ordering::Release);
    // Implicit read from protector not detected as a data race(?), but uncommenting this will be
    // detected:
    //println!("{v}");
}

fn main() {
    let layout = alloc::Layout::new::<u8>();
    let ptr = unsafe { alloc::alloc(layout) };
    assert!(!ptr.is_null());
    unsafe { *ptr = 8 };

    struct SendPtr(*mut u8);
    unsafe impl Send for SendPtr {}
    let send_ptr = SendPtr(ptr);

    std::thread::spawn(|| {
        let send_ptr = send_ptr;
        let r = unsafe { &*send_ptr.0 };
        protector_data_race(r);
    });

    loop {
        if FLAG.load(Ordering::Acquire) {
            // There is an actual race since we need to delay here to prevent the dealloc from
            // running while the protector still exists (generating a different kind of error).
            //
            // But we want to preserve the race to test the race detector... so this just loops,
            // perhaps there is a better and less fragile way to do this.
            for _ in 0..100 {}
            unsafe { alloc::dealloc(ptr.cast(), layout) };
            break;
        }
    }
}
```

The relevant code seems to be:
https://github.com/rust-lang/miri/blob/13c915bb42aab8ee4075632e2aff42e95d570b56/src/borrow_tracker/tree_borrows/mod.rs#L109
https://github.com/rust-lang/miri/blob/13c915bb42aab8ee4075632e2aff42e95d570b56/src/borrow_tracker/tree_borrows/tree.rs#L660

RalfJung:
Cc @Vanille-N

Vanille-N:
Oh no is this a case of "the implementation didn't support this out of the box so I didn't implement it at all" ?
[...]

PR #4254, opened 2025-04-07
Tree Borrows: Make protector end semantics trigger data races, fixing #3266

JoJoDeveloping:
As remarked in #3266, implicit accesses due to protector end semantics ought to trigger data races. While the general story for data races and their interaction with the aliasing model is not completely clear, this is pretty much certain, since otherwise the optimizations relying on protector end semantics are unlikely to hold in a concurrent setting.

The fix messes with the API of the vector clock (making it no longer require `&mut` when doing a write in the data race model), because doing it otherwise would require code duplication due to the Borrow Checker being annoyed by too many mutable references to parts of `MiriMachine`.

Since the aliasing model can now cause data races not related to retags, I tweaked the help text that is printed by data races based on implicit reads inserted by the aliasing model. We could also make this more complicated, but this would require more bits in the [clever bitwise encoding](https://github.com/rust-lang/miri/blob/master/src/concurrency/vector_clock.rs#L63) of read types into the vector clock.


JoJoDeveloping:
Hehe, the pipeline fails because of the newly introduced check. However, it only changes the kind of UB that is produced. What is more funny is that this new behavior is only observed on apple/darwin. Presumably the implementation of that platform's ordering primitives makes things different?

RalfJung:
The basic atomics are implemented the same everywhere, and same for `Arc`... but yeah the `Barrier` implementation could be platform-specific.

RalfJung:
Still, very strange that this would result in different behavior outside of the `sync::Barrier` implementation.  Any idea why macOS sees a data race instead of the TB violation? The tree and the synchronization should be the same, no?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -306,6 +307,20 @@ struct TreeVisitor<'tree> {
     perms: &'tree mut UniValMap<LocationState>,
 }
 
+pub enum AccessRangeAndKind<'a, 'tcx> {
+    Regular { range: AllocRange, kind: AccessKind, cause: diagnostics::AccessCause },
+    MagicProtectorEnd(Option<&'a VClockAlloc>, &'a MiriMachine<'tcx>),
+}
```
RalfJung, line 313: Could you factor this new enum into a separate PR? That seems worth having independently of this -- and it's always easier if commits without functional changes are split from those with functional changes.

JoJoDeveloping:
The semantics implemented here are a non-candidate, since they break safe code.
Here is the counterexample:

```
use std::sync::*;
use std::thread;

fn main() {
    // create a Mutex
    let data = Box::leak(Box::new(Mutex::new(42)));
    fn foo(x: &'static mut Mutex<i32>) {
        // we get the Mutex as a mutable reference. This means we
        // 1) get it with a protector
        // 2) can use non-atomic accesses
        *Mutex::get_mut(x).unwrap() = 21;
        // we have now written to the Mutex, non-atomically.
        // the protector will now protect us from _all_ foreign accesses,
        // and since we would really like to add/insert/reorder writes to mutable references,
        // there will be a protector end semantics write later.

        // create a shared reference to the mutex, which does not reborrow (due to interior mutability)
        let x = &*x;
        thread::spawn(move || {
            // we move the shared reference to another thread
            let x = x;
            // yield, to ensure the protector ends "first"
            std::thread::yield_now();
            // now lock the mutex and use it for something
            let mut xl = x.lock().unwrap();
            *xl = *xl + 1;
        });
        // here, the protector end access happens
        // if we count this as a data race, it will race with the mutex above.
        // indeed, the code in this PR detects the data race. In entirely safe code (!)
    }
    thread::spawn(move || {
        foo(data);
    })
    .join()
    .unwrap();
}
```

PR #4259, opened 2025-04-09
triagebot: enable `[canonicalize-issue-links]` and `[no-mentions]`

Urgau:
This PR enables two new triagebot feature `[canonicalize-issue-links]` and `[no-mentions]`.

Documentation pending at https://github.com/rust-lang/rust-forge/pull/825 and https://github.com/rust-lang/rust-forge/pull/827.

### DIFF
--- a/triagebot.toml
+++ b/triagebot.toml
@@ -40,3 +40,9 @@ unless = ["S-blocked", "S-waiting-on-team", "S-waiting-on-review"]
 
 # Automatically close and reopen PRs made by bots to run CI on them
 [bot-pull-requests]
+
+# Canonicalize issue numbers to avoid closing the wrong issue when upstreaming this subtree
+[canonicalize-issue-links]
+
+# Prevents mentions in commits to avoid users being spammed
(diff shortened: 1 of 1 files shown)


oli-obk:
Cool

RalfJung:
This does not work the way I expected: it posted a long comment in https://github.com/rust-lang/miri/pull/4277 saying we should change a bunch of links. That's not very useful, that PR is auto-generated based on rustc history and we can't change any part of it.

Urgau:
I believe it's flagging every "Auto merge" and "Rollup merge" commits, we should either exclude those, or your "Automatic Rustup" PRs, up to you as what you prefer.

As for flagging on external repo (rust-lang/rust here) it's currently a limitation since not all repos have the feature enable yet, I'm waiting on some feedback, like this one, to iron out some details before enabling it there.

RalfJung:
Ignoring the sync PRs makes most sense IMO, we already ignore them for the merge commit check. Only regular user PRs should be subject to these checks.

PR #4260, opened 2025-04-10
Implement file deletion on Windows

CraftSpider:
This adds a shim for DeleteFileW, and both std and Windows API tests for it


ChrisDenton:
My one concern here is in the churn around this function. We may or may not switch entirely to using `SetFileInformationByHandle` to set the delete disposition.

CraftSpider:
IMO, it's fine if std stops using this in the future - I'm not barring these shims to std only, like most libc shims, any library can use them. The direct tests of the function are at Ralf's suggestion, specifically so that we don't rely on std's specific implementation to ensure each shim works as expected. `SetFileInformationByHandle` is used in renaming already, so I'll try to simply implement deletion as well for future-proofing when I reach it.

PR #4261, opened 2025-04-10
Implement RtlNtStatusToDosError and shim test for it

CraftSpider:
This is required for error handling of read/write, and fairly simple to PR on its own beforehand.


CraftSpider:
If you happen to have a list on hand, that could be useful. For now the list is roughly just the statuses my WIP implementation has the ability to emit so far.

ChrisDenton:
For entertainment purposes only then, here's [a big csv](https://github.com/user-attachments/files/19688252/ntdos.csv) I have laying around (note it may not be up to date with new codes).

PR #4262, opened 2025-04-10
path: add more Windows tests

RalfJung:
Check that these tests pass on a Windows host. They will still fail on Unix hosts.

### DIFF
--- a/src/shims/windows/foreign_items.rs
+++ b/src/shims/windows/foreign_items.rs
@@ -26,57 +26,107 @@ pub fn is_dyn_sym(name: &str) -> bool {
 }
 
 #[cfg(windows)]
-fn win_absolute<'tcx>(path: &Path) -> InterpResult<'tcx, io::Result<PathBuf>> {
+fn win_get_full_path_name<'tcx>(path: &Path) -> InterpResult<'tcx, io::Result<PathBuf>> {
     // We are on Windows so we can simply let the host do this.
     interp_ok(path::absolute(path))
 }
 
 #[cfg(unix)]
 #[expect(clippy::get_first, clippy::arithmetic_side_effects)]
-fn win_absolute<'tcx>(path: &Path) -> InterpResult<'tcx, io::Result<PathBuf>> {
-    // We are on Unix, so we need to implement parts of the logic ourselves.
+fn win_get_full_path_name<'tcx>(path: &Path) -> InterpResult<'tcx, io::Result<PathBuf>> {
+    use std::sync::LazyLock;
+
+    use rustc_data_structures::fx::FxHashSet;
+
+    // We are on Unix, so we need to implement parts of the logic ourselves. `path` will use `/`
+    // separators, and the result should also use `/`.
+    // See <https://chrisdenton.github.io/omnipath/Overview.html#absolute-win32-paths> for more
+    // information about Windows paths.
+    // This does not handle all corner cases correctly, see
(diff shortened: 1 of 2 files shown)


ChrisDenton:
Windows 11 changed the handling of "magic filenames" so that now `C:\path\to\COM1` is not changed, only `COM1` on its own is.

`GetFullPathNameW` treats `\\?\` as equivalent to `\\.\`. However, this is not how Windows otherwise treats such paths so the standard library will always avoid passing `\\?\` paths to it.

If you're desperate for more edge cases then I can think of a few

|Original|Full path|
|--|--|
`C:\path\to\COM1`|`C:\path\to\COM1`
`COM1`|`\\.\COM1`|
`C:\path\to\NUL`|`C:\path\to\NUL`
`C:\path\to\nul`|`\\.\nul`
`C:\path\to\nUl`|`\\.\nUl`
`C:\path\to\NUL..`|`\\.\NUL`
`C:\path\to\NUL:..`|`\\.\NUL`
`C:\path\to\NUL.:.`|`C:\path\to\NUL.:`
`C:\path\to\NUL. .`|`\\.\NUL`
`\\server\share\NUL`|`\\server\share\NUL`
`\\server\share\NUL`|`\\server\share\NUL`
`\\.\PIPE\NUL`|`\\.\PIPE\NUL`
`C:\path.\to..\file`|`C:\path\to..\file`|
`\\server\share.`|`\\server\share`|
`\\server\share.\path.\to..\file`|`\\server\share.\path\to..\file`|
`C:\path\to\file.. ..`|`C:\path\to\file`|
`C:\..`|`C:\`
`C:\path\..`|`C:\`
`\\server\share\..`|`\\server\share`
`\\server\share\path\..`|`\\server\share\path\..`
`\\.\PIPE\..`|`\\.\`

ChrisDenton:
>  Is `PIPE` somehow special? It's not on the list of magic filenames, it seems.

No, it's just the most common use of `\\.\` paths. But it could be the name of any device. Even `\\.\C:`.

RalfJung:
Ah so magic filenames are case-insensitive? Damn...
And the trailing `.` removal also removes trailing ` ` it seems?
I'm extremely confused by `C:\path\to\file.. ..`, why does that remove the trailing stuff but `file..` has it preserved? Honestly I'll just wait until someone files a bug, I don't want to spend a day replicating completely idiotic Windows semantics.^^ But thanks for letting me know that this is even more silly than I thought it was.

Btw `\\server\share\NUL` appears twice in your list.

ChrisDenton:
All sequences of `.` and ` ` are trimmed from the end of the string (except a single `.` or a double `..`). Think of it like a DOS shell which trims whitespace from the commands and arguments that are entered in.

I'm not sure why but a single dot at the end of a path component, e.g. `C:\path.\`, is stripped but if there are two or more then it is not. This might just be a consequence of old code that had a bug in handling `.` components (maybe matching on `[^.].\` or something silly) and now the faulty behaviour is preserved forever. Or well, unless someone decides they can fix it without breaking programs.

But yeah, I think that it might be best not to over test for the current behaviour in any case. It can potentially change, as with the Windows 11 changes.

> Btw `\\server\share\NUL` appears twice in your list.

Ah yeah, I updated to fix some copy/paste errors.

RalfJung:
I fixed the `..` handling for `C:\..` and `//server/share/..`. I'm willing to ignore the other odd corner cases I think...

PR #4263, opened 2025-04-10
Replace check_shim with check_shim_abi in unix/foreign_items shims

geetanshjuneja:
Links to #3842.

### DIFF
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -112,51 +112,122 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         match link_name.as_str() {
             // Environment related shims
             "getenv" => {
-                let [name] = this.check_shim(abi, Conv::C, link_name, args)?;
+                let [name] = this.check_shim_abi(
+                    link_name,
+                    abi,
+                    ExternAbi::C { unwind: false },
+                    [this.machine.layouts.const_raw_ptr.ty],
+                    this.machine.layouts.mut_raw_ptr.ty,
+                    args,
+                )?;
                 let result = this.getenv(name)?;
                 this.write_pointer(result, dest)?;
             }
             "unsetenv" => {
-                let [name] = this.check_shim(abi, Conv::C, link_name, args)?;
+                let [name] = this.check_shim_abi(
+                    link_name,
+                    abi,
+                    ExternAbi::C { unwind: false },
+                    [this.machine.layouts.const_raw_ptr.ty],
+                    this.tcx.types.i32,
+                    args,
+                )?;
                 let result = this.unsetenv(name)?;
                 this.write_scalar(result, dest)?;
             }
             "setenv" => {
-                let [name, value, overwrite] = this.check_shim(abi, Conv::C, link_name, args)?;
+                let [name, value, overwrite] = this.check_shim_abi(
+                    link_name,
+                    abi,
+                    ExternAbi::C { unwind: false },
+                    [
+                        this.machine.layouts.const_raw_ptr.ty,
+                        this.machine.layouts.const_raw_ptr.ty,
+                        this.tcx.types.i32,
+                    ],
+                    this.tcx.types.i32,
+                    args,
+                )?;
                 this.read_scalar(overwrite)?.to_i32()?;
                 let result = this.setenv(name, value)?;
                 this.write_scalar(result, dest)?;
             }
             "getcwd" => {
-                let [buf, size] = this.check_shim(abi, Conv::C, link_name, args)?;
+                let [buf, size] = this.check_shim_abi(
+                    link_name,
+                    abi,
+                    ExternAbi::C { unwind: false },
+                    [this.machine.layouts.mut_raw_ptr.ty, this.tcx.types.usize],
+                    this.machine.layouts.mut_raw_ptr.ty,
+                    args,
+                )?;
                 let result = this.getcwd(buf, size)?;
                 this.write_pointer(result, dest)?;
             }
             "chdir" => {
-                let [path] = this.check_shim(abi, Conv::C, link_name, args)?;
+                let [path] = this.check_shim_abi(
+                    link_name,
+                    abi,
+                    ExternAbi::C { unwind: false },
+                    [this.machine.layouts.const_raw_ptr.ty],
+                    this.tcx.types.i32,
+                    args,
+                )?;
                 let result = this.chdir(path)?;
                 this.write_scalar(result, dest)?;
             }
             "getpid" => {
-                let [] = this.check_shim(abi, Conv::C, link_name, args)?;
+                let [] = this.check_shim_abi(
+                    link_name,
+                    abi,
+                    ExternAbi::C { unwind: false },
+                    [],
+                    this.libc_ty_layout("pid_t").ty,
+                    args,
+                )?;
                 let result = this.getpid()?;
                 this.write_scalar(result, dest)?;
             }
             "sysconf" => {
-                let [val] = this.check_shim(abi, Conv::C, link_name, args)?;
+                let [val] = this.check_shim_abi(
+                    link_name,
+                    abi,
+                    ExternAbi::C { unwind: false },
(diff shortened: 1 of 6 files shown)


RalfJung:
No please don't make this PR any bigger. :)

src/shims/unix/foreign_items.rs
```diff
@@ -112,67 +112,162 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                let [fd, buf, count, offset] = this.check_shim(abi, Conv::C, link_name, args)?;
+                let [fd, buf, count, offset] = this.check_shim_abi(
+                    link_name,
+                    abi,
+                    ExternAbi::C { unwind: false },
+                    [
+                        this.tcx.types.i32,
+                        this.machine.layouts.mut_raw_ptr.ty,
+                        this.tcx.types.usize,
+                        this.libc_ty_layout("off_t").ty,
```
RalfJung, line 246: Please let-bind this; it us used twice in this match arm. (Happens a few more times below.)

src/shims/unix/foreign_items.rs
```diff
@@ -250,87 +390,201 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                 let result = this.closedir(dirp)?;
                 this.write_scalar(result, dest)?;
             }
             "lseek64" => {
-                let [fd, offset, whence] = this.check_shim(abi, Conv::C, link_name, args)?;
+                let [fd, offset, whence] = this.check_shim_abi(
+                    link_name,
+                    abi,
+                    ExternAbi::C { unwind: false },
+                    [this.tcx.types.i32, this.libc_ty_layout("off64_t").ty, this.tcx.types.i32],
```
RalfJung, line 481: Below we assume this is `i64`...

src/shims/unix/foreign_items.rs
```diff
@@ -250,87 +390,201 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                 let result = this.lseek64(fd, offset, whence)?;
                 this.write_scalar(result, dest)?;
             }
             "ftruncate64" => {
-                let [fd, length] = this.check_shim(abi, Conv::C, link_name, args)?;
+                let [fd, length] = this.check_shim_abi(
+                    link_name,
+                    abi,
+                    ExternAbi::C { unwind: false },
+                    [this.tcx.types.i32, this.libc_ty_layout("off64_t").ty],
```
RalfJung, line 511: Below we assume `i64`.

src/shims/unix/foreign_items.rs
```diff
@@ -250,87 +390,201 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let [pathname, buf, bufsize] = this.check_shim_abi(
+                    link_name,
+                    abi,
+                    ExternAbi::C { unwind: false },
+                    [
+                        this.machine.layouts.const_raw_ptr.ty,
+                        this.machine.layouts.mut_raw_ptr.ty,
+                        this.tcx.types.usize,
+                    ],
+                    this.libc_ty_layout("ssize_t").ty,
```
RalfJung, line 568: Below we use `isize`.

src/shims/unix/foreign_items.rs
```diff
@@ -250,87 +390,201 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             "posix_fadvise" => {
-                let [fd, offset, len, advice] = this.check_shim(abi, Conv::C, link_name, args)?;
+                let [fd, offset, len, advice] = this.check_shim_abi(
+                    link_name,
+                    abi,
+                    ExternAbi::C { unwind: false },
+                    [
+                        this.tcx.types.i32,
+                        this.libc_ty_layout("off_t").ty,
+                        this.libc_ty_layout("off_t").ty,
```
RalfJung, line 582: Below we assume `isize`.

src/shims/unix/foreign_items.rs
```diff
@@ -339,49 +593,109 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                 let result = this.localtime_r(timep, result_op)?;
                 this.write_pointer(result, dest)?;
             }
             "clock_gettime" => {
-                let [clk_id, tp] = this.check_shim(abi, Conv::C, link_name, args)?;
+                let [clk_id, tp] = this.check_shim_abi(
+                    link_name,
+                    abi,
+                    ExternAbi::C { unwind: false },
+                    [this.libc_ty_layout("clockid_t").ty, this.machine.layouts.mut_raw_ptr.ty],
```
RalfJung, line 695: The code inside `clock_gettime` assumes `i32`.

RalfJung (commented):
Thanks! You uncovered a few cases where we are not properly consistent with our types.

We generally assume that `long` is `isize` and `unsigned long` is `usize`, so please also keep doing that here. Same for `size_t` and `ssize_t`, those are `usize` and `isize`, respectively.

For types like `off_t`, `off64_t`, `clockid_t`, we *should* be using their proper types, but some functions do not. So please do one of the following:
- change the logic inside these functions to use the proper type. Typically this means using e.g. `to_int(this.libc_ty_layout("off_t").size)` instead of `to_i32()`.
- or keep the logic the same, put `i32` into the `check_shim_abi` invocation, and add a comment saying "FIXME: should be off_t".

@rustbot author

src/shims/unix/foreign_items.rs
```diff
@@ -112,67 +112,162 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                 this.write_scalar(result, dest)?;
             }
             "sysconf" => {
-                let [val] = this.check_shim(abi, Conv::C, link_name, args)?;
+                let [val] = this.check_shim_abi(
+                    link_name,
+                    abi,
+                    ExternAbi::C { unwind: false },
+                    [this.tcx.types.i32],
+                    this.libc_ty_layout("c_long").ty,
```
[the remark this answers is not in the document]
geetanshjuneja:
```
unsafe { libc::sysconf(libc::_SC_PAGESIZE) as usize }
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ calling a function with return type usize passing return place of type i64
```
Converting it into usize is causing error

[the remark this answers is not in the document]
geetanshjuneja: https://docs.rs/libc/latest/libc/type.c_long.html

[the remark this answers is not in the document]
RalfJung: I think I meant `isize`, sorry.

This is where the return value is computed:
https://github.com/rust-lang/miri/blob/74f4d67ae9e64b04512a6d52ec9775429c379bd9/src/shims/unix/foreign_items.rs#L48-L74

But you are right "long" should mean `i64`. S why does this code even work then?^^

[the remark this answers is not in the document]
RalfJung: Ah no of course, the docs you are linking do are the 64bit docs. `long` is `isize`; it's `i32` on 32bit systems.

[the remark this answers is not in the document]
geetanshjuneja: I also don't know why it is working, it should fail on 32 bit architecture but tests are passing on `i686-unknown-linux-gnu`after making it `i64`.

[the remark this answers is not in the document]
RalfJung: See https://github.com/rust-lang/miri/pull/4263#discussion_r2039063586 for why it works.

You should use `isize` here.

src/shims/unix/foreign_items.rs
RalfJung: This has not been resolved. There's still `this.read_scalar(clk_id_op)?.to_i32()?` in there.

You changed the return value handling, but that is unnecessary.

RalfJung:

@rustbot author

PR #4269, opened 2025-04-11
Expose Pointer to/from AllocId conversion functions, make some arguments optional

Patrick-6:
We need to have access to `alloc_id_from_addr` and `addr_from_alloc_id` for some ongoing work in Miri, so this change makes them available.

The `MemoryKind` argument to `addr_from_alloc_id` is only required if the result is not cached, and we don't have easy access to the `MemoryKind` in the location where we want to call the function, so it is now optional.

We need to be able to get `AllocId`s for allocations that were not exposed, so there is now a parameter to `alloc_id_from_addr` to get results for all allocations vs only exposed ones.

r? @RalfJung

### DIFF
--- a/src/alloc_addresses/mod.rs
+++ b/src/alloc_addresses/mod.rs
@@ -107,47 +107,6 @@ fn align_addr(addr: u64, align: u64) -> u64 {
 
 impl<'tcx> EvalContextExtPriv<'tcx> for crate::MiriInterpCx<'tcx> {}
 trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
-    // Returns the exposed `AllocId` that corresponds to the specified addr,
-    // or `None` if the addr is out of bounds
-    fn alloc_id_from_addr(&self, addr: u64, size: i64) -> Option<AllocId> {
-        let this = self.eval_context_ref();
-        let global_state = this.machine.alloc_addresses.borrow();
-        assert!(global_state.provenance_mode != ProvenanceMode::Strict);
-
-        // We always search the allocation to the right of this address. So if the size is structly
-        // negative, we have to search for `addr-1` instead.
-        let addr = if size >= 0 { addr } else { addr.saturating_sub(1) };
-        let pos = global_state.int_to_ptr_map.binary_search_by_key(&addr, |(addr, _)| *addr);
-
-        // Determine the in-bounds provenance for this pointer.
-        let alloc_id = match pos {
-            Ok(pos) => Some(global_state.int_to_ptr_map[pos].1),
-            Err(0) => None,
-            Err(pos) => {
-                // This is the largest of the addresses smaller than `int`,
-                // i.e. the greatest lower bound (glb)
-                let (glb, alloc_id) = global_state.int_to_ptr_map[pos - 1];
-                // This never overflows because `addr >= glb`
-                let offset = addr - glb;
-                // We require this to be strict in-bounds of the allocation. This arm is only
-                // entered for addresses that are not the base address, so even zero-sized
-                // allocations will get recognized at their base address -- but all other
-                // allocations will *not* be recognized at their "end" address.
-                let size = this.get_alloc_info(alloc_id).size;
-                if offset < size.bytes() { Some(alloc_id) } else { None }
-            }
-        }?;
-
-        // We only use this provenance if it has been exposed.
-        if global_state.exposed.contains(&alloc_id) {
-            // This must still be live, since we remove allocations from `int_to_ptr_map` when they get freed.
-            debug_assert!(this.is_alloc_live(alloc_id));
-            Some(alloc_id)
-        } else {
-            None
-        }
-    }
-
     fn addr_from_alloc_id_uncached(
         &self,
         global_state: &mut GlobalStateInner,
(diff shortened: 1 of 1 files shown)


src/alloc_addresses/mod.rs
```diff
@@ -105,16 +105,22 @@ fn align_addr(addr: u64, align: u64) -> u64 {
     }
 }
 
-impl<'tcx> EvalContextExtPriv<'tcx> for crate::MiriInterpCx<'tcx> {}
-trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
-    // Returns the exposed `AllocId` that corresponds to the specified addr,
-    // or `None` if the addr is out of bounds
-    fn alloc_id_from_addr(&self, addr: u64, size: i64) -> Option<AllocId> {
+impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
+pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
```
RalfJung, line 109: This also makes `addr_from_alloc_id_uncached` public, which should remain private.

src/alloc_addresses/mod.rs
```diff
@@ -423,7 +432,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             alloc_id
         } else {
             // A wildcard pointer.
-            this.alloc_id_from_addr(addr.bytes(), size)?
+            this.alloc_id_from_addr(addr.bytes(), size, true)?
```
RalfJung, line 435:
```suggestion
            this.alloc_id_from_addr(addr.bytes(), size, /* only_exposed_allocations */ true)?
```

RalfJung:
Thanks! I have some minor comments.
@rustbot author

src/alloc_addresses/mod.rs
Patrick-6: It's now private again, but I had to add `#[allow(private_bounds)]` to `trait EvalContextExt` to silence the warning about the private super-trait.

src/alloc_addresses/mod.rs
Patrick-6: I've used an extra variable here instead of your suggestion, because otherwise with rust-analyzer the parameter name is shown twice.

src/alloc_addresses/mod.rs
RalfJung: None of the other files do that. So this is not necessary.

src/alloc_addresses/mod.rs
RalfJung: I'm not happy about your editor settings making you write code in a way that it becomes unnecessarily verbose when read in other editors that avoid those distracting inline hints (and ofc they are optional in RA as well).

src/alloc_addresses/mod.rs
```diff
@@ -248,6 +201,57 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // This must still be live, since we remove allocations from `int_to_ptr_map` when they get freed.
+            debug_assert!(this.is_alloc_live(alloc_id));
+            Some(alloc_id)
+        } else {
+            None
+        }
+    }
 
     /// Returns the base address of an allocation, or an error if no base address could be found
     ///
```
RalfJung, line 257: There's a `self.addr_from_alloc_id_uncached` below that should be `this.self.addr_from_alloc_id_uncached` to fix the bound issue.

(If only github would let me write this comment on the right line...)

src/alloc_addresses/mod.rs
RalfJung: This is is pretty harmless so :shrug: we can leave it. But please keep in mind that not everyone uses RA, so first and foremost code must be readable in plain ASCII rendering. If RA's settings get in your way, don't default to externalizing that cost -- consider fixing your setup instead.

src/alloc_addresses/mod.rs
Patrick-6: Thx, that worked (but with `this` instead of `this.self`)

RalfJung: *oops* yaeh that was a typo, sorry.

src/alloc_addresses/mod.rs
```diff
@@ -173,9 +132,9 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     if memory_kind == MiriMemoryKind::Global.into() {
                         // For new global allocations, we always pre-allocate the memory to be able use the machine address directly.
                         let prepared_bytes = MiriAllocBytes::zeroed(info.size, info.align)
-                            .unwrap_or_else(|| {
-                                panic!("Miri ran out of memory: cannot create allocation of {size:?} bytes", size = info.size)
-                            });
+                        .unwrap_or_else(|| {
+                            panic!("Miri ran out of memory: cannot create allocation of {size:?} bytes", size = info.size)
+                        });
```
RalfJung, line 137: Why did you un-indent this? Or did rustfmt do this automatically for some reason?

RalfJung: Yes please undo the change. Good job tracking down the source of why rustfmt does not format this!

RalfJung:
Also, that's the last nit I have here, so please [squash the commits](https://rustc-dev-guide.rust-lang.org/git.html#squash-your-commits). Please use the `--keep-base` flag when squashing so that the force-push diff is easier to review.

Patrick-6:
Done, I've rebased on the current master branch and squashed the commits.

RalfJung:
> I've rebased on the current master branch 

Why? I explicitly asked you not do that... I said to use `--keep-base`.

RalfJung:
Otherwise, looks good, thanks!
@bors r+

Patrick-6:
> Why? I explicitly asked you not do that... I said to use `--keep-base`.

I did, but then I noticed that there were new commits that happened between when I started working on my branch and now, I wanted to make sure everything was up to date.

PR #4272, opened 2025-04-13
Make thread scheduling fully random

### LINKED ISSUE #4230: Support non-round-robin scheduling
Miri now randomly yields the current thread every now and then, but currently it always yields to the next thread, i.e. it's a round-robin scheduler. It would probably be good to randomize this more and pick some arbitrary unblocked thread whenever we have yielded.

To still provide a fixed schedule for testcases, we could have a flag `-Zmiri-fixed-schedule` that disables both automatic yielding and that forces things to be round-robin again. All existing tests that use `-Zmiri-preemption-rate=0` should then use this new flag instead.

geetanshjuneja:
Closes rust-lang/miri#4230.

### DIFF
--- a/src/bin/miri.rs
+++ b/src/bin/miri.rs
@@ -573,6 +573,8 @@ fn main() {
             miri_config.mute_stdout_stderr = true;
         } else if arg == "-Zmiri-retag-fields" {
             miri_config.retag_fields = RetagFields::Yes;
+        } else if arg == "-Zmiri-fixed-schedule" {
+            miri_config.fixed_scheduling = true;
         } else if let Some(retag_fields) = arg.strip_prefix("-Zmiri-retag-fields=") {
             miri_config.retag_fields = match retag_fields {
                 "all" => RetagFields::Yes,
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -6,6 +6,8 @@ use std::task::Poll;
 use std::time::{Duration, SystemTime};
 
 use either::Either;
+use rand::rngs::StdRng;
+use rand::seq::IteratorRandom;
 use rustc_abi::ExternAbi;
 use rustc_const_eval::CTRL_C_RECEIVED;
 use rustc_data_structures::fx::FxHashMap;
@@ -401,6 +403,8 @@ pub struct ThreadManager<'tcx> {
     thread_local_allocs: FxHashMap<(DefId, ThreadId), StrictPointer>,
     /// A flag that indicates that we should change the active thread.
     yield_active_thread: bool,
+    /// A flag that indicates that we should do round robin scheduling of threads else randomized scheduling is used.
+    fixed_scheduling: bool,
 }
 
 impl VisitProvenance for ThreadManager<'_> {
@@ -410,6 +414,7 @@ impl VisitProvenance for ThreadManager<'_> {
             thread_local_allocs,
             active_thread: _,
             yield_active_thread: _,
+            fixed_scheduling: _,
         } = self;
 
         for thread in threads {
@@ -421,8 +426,8 @@ impl VisitProvenance for ThreadManager<'_> {
     }
 }
 
-impl<'tcx> Default for ThreadManager<'tcx> {
-    fn default() -> Self {
+impl<'tcx> ThreadManager<'tcx> {
+    pub(crate) fn new(config: &MiriConfig) -> Self {
         let mut threads = IndexVec::new();
         // Create the main thread and add it to the list of threads.
         threads.push(Thread::new(Some("main"), None));
@@ -431,11 +436,10 @@ impl<'tcx> Default for ThreadManager<'tcx> {
             threads,
             thread_local_allocs: Default::default(),
             yield_active_thread: false,
+            fixed_scheduling: config.fixed_scheduling,
         }
     }
-}
 
-impl<'tcx> ThreadManager<'tcx> {
     pub(crate) fn init(
         ecx: &mut MiriInterpCx<'tcx>,
         on_main_stack_empty: StackEmptyCallback<'tcx>,
@@ -702,7 +706,11 @@ impl<'tcx> ThreadManager<'tcx> {
     /// used in stateless model checkers such as Loom: run the active thread as
     /// long as we can and switch only when we have to (the active thread was
     /// blocked, terminated, or has explicitly asked to be preempted).
-    fn schedule(&mut self, clock: &MonotonicClock) -> InterpResult<'tcx, SchedulingAction> {
+    fn schedule(
+        &mut self,
+        clock: &MonotonicClock,
+        rng: &mut StdRng,
+    ) -> InterpResult<'tcx, SchedulingAction> {
         // This thread and the program can keep going.
         if self.threads[self.active_thread].state.is_enabled() && !self.yield_active_thread {
             // The currently active thread is still enabled, just continue with it.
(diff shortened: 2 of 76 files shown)


src/concurrency/thread.rs
```diff
@@ -730,19 +744,24 @@ impl<'tcx> ThreadManager<'tcx> {
...
-            if thread.state.is_enabled() {
-                info!(
-                    "---------- Now executing on thread `{}` (previous: `{}`) ----------------------------------------",
-                    self.get_thread_display_name(id),
-                    self.get_thread_display_name(self.active_thread)
-                );
-                self.active_thread = id;
-                break;
-            }
+            .chain(self.threads.iter_enumerated().take(self.active_thread.index() + 1))
```
oli-obk, line 747: What's with the +1 here?

geetanshjuneja: this is to include the current thread as well in the iterator.

oli-obk: Doesn't round robin scheduling get stuck on the same thread then possibly if the current thread is also the first thread? Or does yielding mark it as not ready so this was always just an optimization on master?

Either way, just remove the chain and take and thus just iterate all threads always?

geetanshjuneja: > Doesn't round robin scheduling get stuck on the same thread then possibly if the current thread is also the first thread?
 
if the current thread is the first it means that all the other threads are terminated or blocked. So it is fine to schedule it again ig.
  
> Or does yielding mark it as not ready so this was always just an optimization on master?

No I dont think yielding marks it as not ready.

src/bin/miri.rs
```diff
@@ -573,6 +573,9 @@ fn main() {
             miri_config.mute_stdout_stderr = true;
         } else if arg == "-Zmiri-retag-fields" {
             miri_config.retag_fields = RetagFields::Yes;
+        } else if arg == "-Zmiri-fixed-schedule" {
+            miri_config.fixed_scheduling = true;
+            miri_config.preemption_rate = 0.0;
```
RalfJung, line 578: We should just not auto-preempt when `fixed_scheduling` is set, so we don't have to set two fields here.

src/concurrency/thread.rs
```diff
@@ -730,19 +744,24 @@ impl<'tcx> ThreadManager<'tcx> {
...
+            .filter(|(_id, thread)| thread.state.is_enabled())
+            .collect::<Vec<_>>();
+
+        let mut new_thread = threads.first();
```
RalfJung, line 751: Also here, please don't use an initialize-then-mutate pattern. It's almost always better to initialize things with the right value immediately and eschew mutation. So here that would be `let new_thread = if self.round_robin_scheduling { ... } else { ... }`.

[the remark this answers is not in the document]
geetanshjuneja: but the current thread can be in terminated stated or blocked state. So this can be `None`.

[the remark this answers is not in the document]
RalfJung: Oh, right.

src/concurrency/thread.rs
```diff
@@ -436,12 +436,25 @@ impl<'tcx> Default for ThreadManager<'tcx> {
...
             thread_local_allocs: Default::default(),
             yield_active_thread: false,
-            round_robin_scheduling: false,
+            fixed_scheduling: false,
         }
     }
 }
 
 impl<'tcx> ThreadManager<'tcx> {
+    pub(crate) fn new(config: &MiriConfig) -> Self {
```
RalfJung, line 445: Please remove the `Default for ThreadManager`, we should not use it any more.

src/concurrency/thread.rs
```diff
@@ -730,20 +753,24 @@ impl<'tcx> ThreadManager<'tcx> {
             .threads
```
RalfJung, line 753: The comment above is still outdated, as noted before.

src/concurrency/thread.rs
```diff
@@ -730,20 +753,24 @@ impl<'tcx> ThreadManager<'tcx> {
...
                     "---------- Now executing on thread `{}` (previous: `{}`) ----------------------------------------",
-                    self.get_thread_display_name(id),
+                    self.get_thread_display_name(*id),
                     self.get_thread_display_name(self.active_thread)
                 );
-                self.active_thread = id;
-                break;
+                self.active_thread = *id;
             }
+            self.yield_active_thread = false;
```
RalfJung, line 771: This should be outside the `if`.

RalfJung (commented):
@rustbot author

src/concurrency/thread.rs
```diff
@@ -730,20 +753,24 @@ impl<'tcx> ThreadManager<'tcx> {
             .threads
             .iter_enumerated()
             .skip(self.active_thread.index() + 1)
-            .chain(self.threads.iter_enumerated().take(self.active_thread.index()));
-        for (id, thread) in threads {
-            debug_assert_ne!(self.active_thread, id);
-            if thread.state.is_enabled() {
+            .chain(self.threads.iter_enumerated().take(self.active_thread.index() + 1))
+            .filter(|(_id, thread)| thread.state.is_enabled())
+            .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
geetanshjuneja: I did this to use `choose` method available on `Vec`.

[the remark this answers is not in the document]
geetanshjuneja: this can be we done using `for` but we first need to find the number of available threads for scheduling to calculate probability.

src/concurrency/thread.rs
geetanshjuneja:
```
// No callbacks immediately scheduled, pick a regular thread to execute.
// The active thread blocked or yielded. So we go search for another enabled thread.
// Crucially, we start searching at the current active thread ID, rather than at 0, since we
// want to avoid always scheduling threads 0 and 1 without ever making progress in thread 2.
//
// `skip(N)` means we start iterating at thread N, so we skip 1 more to start just *after*
// the active thread. Then after that we look at `take(N)`, i.e., the threads *before* the
// active thread.
```

this comment?

src/concurrency/thread.rs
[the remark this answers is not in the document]
RalfJung: `choose` works on arbitrary iterators, you don't need a `Vec`: https://docs.rs/rand/latest/rand/seq/trait.IteratorRandom.html.

src/concurrency/thread.rs
RalfJung: Yes. It talks about `take(N)` which you changed to `take(N+1)`.

src/concurrency/thread.rs
```diff
@@ -720,30 +728,34 @@ impl<'tcx> ThreadManager<'tcx> {
         }
         // No callbacks immediately scheduled, pick a regular thread to execute.
         // The active thread blocked or yielded. So we go search for another enabled thread.
-        // Crucially, we start searching at the current active thread ID, rather than at 0, since we
-        // want to avoid always scheduling threads 0 and 1 without ever making progress in thread 2.
-        //
+        // If fixed scheduling is enabled we start searching after the current active thread ID, rather than at 0,
+        // since we want to avoid always scheduling threads 0 and 1 without ever making progress in thread 2,
+        // otherwise we pick a random thread from the list of schedulable threads.
```
RalfJung, line 733:
```suggestion
        // We build the list of threads by starting with the thread after the current one, followed by
        // the threads before the current one and then the current thread itself (i.e., this iterator acts
        // like `threads.rotate_left(self.active_thread.index() + 1)`. This ensures that if we pick the first
        // eligible thread, we do regular round-robin scheduling, and all threads get a chance to take a step.
```

src/concurrency/thread.rs
```diff
@@ -720,30 +728,34 @@ impl<'tcx> ThreadManager<'tcx> {
...
-        for (id, thread) in threads {
-            debug_assert_ne!(self.active_thread, id);
-            if thread.state.is_enabled() {
+            .chain(self.threads.iter_enumerated().take(self.active_thread.index() + 1))
+            .filter(|(_id, thread)| thread.state.is_enabled());
+
+        let new_thread =
+            if self.fixed_scheduling { threads_iter.next() } else { threads_iter.choose(rng) };
+
+        if let Some((id, _thread)) = new_thread {
```
RalfJung, line 747:
```suggestion
        // Pick a new thread, and switch to it.
        let new_thread =
            if self.fixed_scheduling { threads_iter.next() } else { threads_iter.choose(rng) };
        if let Some((id, _thread)) = new_thread {
```

src/concurrency/thread.rs
```diff
@@ -720,30 +728,34 @@ impl<'tcx> ThreadManager<'tcx> {
...
                 info!(
                     "---------- Now executing on thread `{}` (previous: `{}`) ----------------------------------------",
                     self.get_thread_display_name(id),
                     self.get_thread_display_name(self.active_thread)
                 );
                 self.active_thread = id;
-                break;
             }
         }
         self.yield_active_thread = false;
```
RalfJung, line 757:
```suggestion
        //This completes the `yield`, if any was requested.
        self.yield_active_thread = false;
```

RalfJung (commented):
Just a few final comment nits.

RalfJung:

@rustbot author

src/concurrency/thread.rs
```diff
@@ -720,30 +728,33 @@ impl<'tcx> ThreadManager<'tcx> {
...
                 info!(
                     "---------- Now executing on thread `{}` (previous: `{}`) ----------------------------------------",
                     self.get_thread_display_name(id),
                     self.get_thread_display_name(self.active_thread)
                 );
                 self.active_thread = id;
-                break;
             }
         }
+        //This completes the `yield`, if any was requested.
```
RalfJung, line 755:
```suggestion
        // This completes the `yield`, if any was requested.
```

src/concurrency/thread.rs
```diff
@@ -720,30 +728,33 @@ impl<'tcx> ThreadManager<'tcx> {
...
         // No callbacks immediately scheduled, pick a regular thread to execute.
         // The active thread blocked or yielded. So we go search for another enabled thread.
-        // Crucially, we start searching at the current active thread ID, rather than at 0, since we
-        // want to avoid always scheduling threads 0 and 1 without ever making progress in thread 2.
-        //
-        // `skip(N)` means we start iterating at thread N, so we skip 1 more to start just *after*
-        // the active thread. Then after that we look at `take(N)`, i.e., the threads *before* the
-        // active thread.
-        let threads = self
+        // We build the list of threads by starting with the thread after the current one, followed by
```
RalfJung, line 731:
```suggestion
        // We build the list of threads by starting with the threads after the current one, followed by
```

PR #4273, opened 2025-04-14
TB: add `Cell` state to support more fine-grained tracking of interior mutable data

yoctocell:
This adds a new `Cell` state to TB to represent shared references to interior mutable data. This is still a WIP for making the tracking of interior mutable data per-byte, like in Stacked Borrows.

cc @RalfJung @JoJoDeveloping

### DIFF
--- a/src/borrow_tracker/tree_borrows/mod.rs
+++ b/src/borrow_tracker/tree_borrows/mod.rs
@@ -123,6 +123,10 @@ struct NewPermission {
     /// Whether this pointer is part of the arguments of a function call.
     /// `protector` is `Some(_)` for all pointers marked `noalias`.
     protector: Option<ProtectorKind>,
+    /// Whether a read should be performed on a retag.  This should be `false`
+    /// for `Cell` because this could cause data races when using thread-safe
+    /// data types like `Mutex<T>`.
+    initial_read: bool,
 }
 
 impl<'tcx> NewPermission {
@@ -141,18 +145,19 @@ impl<'tcx> NewPermission {
         // To eliminate the case of Protected Reserved IM we override interior mutability
         // in the case of a protected reference: protected references are always considered
         // "freeze" in their reservation phase.
-        let initial_state = match mutability {
-            Mutability::Mut if ty_is_unpin => Permission::new_reserved(ty_is_freeze, is_protected),
-            Mutability::Not if ty_is_freeze => Permission::new_frozen(),
+        let (initial_state, initial_read) = match mutability {
+            Mutability::Mut if ty_is_unpin =>
+                (Permission::new_reserved(ty_is_freeze, is_protected), true),
+            Mutability::Not if ty_is_freeze => (Permission::new_frozen(), true),
+            Mutability::Not if !ty_is_freeze => (Permission::new_cell(), false),
             // Raw pointers never enter this function so they are not handled.
             // However raw pointers are not the only pointers that take the parent
-            // tag, this also happens for `!Unpin` `&mut`s and interior mutable
-            // `&`s, which are excluded above.
+            // tag, this also happens for `!Unpin` `&mut`s, which are excluded above.
             _ => return None,
         };
 
         let protector = is_protected.then_some(ProtectorKind::StrongProtector);
-        Some(Self { zero_size: false, initial_state, protector })
+        Some(Self { zero_size: false, initial_state, protector, initial_read })
     }
 
     /// Compute permission for `Box`-like type (`Box` always, and also `Unique` if enabled).
@@ -175,6 +180,7 @@ impl<'tcx> NewPermission {
                 zero_size,
                 initial_state,
                 protector: protected.then_some(ProtectorKind::WeakProtector),
+                initial_read: true,
             }
         })
     }
@@ -289,13 +295,15 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let mut tree_borrows = alloc_extra.borrow_tracker_tb().borrow_mut();
 
         // All reborrows incur a (possibly zero-sized) read access to the parent
-        tree_borrows.perform_access(
-            orig_tag,
-            Some((range, AccessKind::Read, diagnostics::AccessCause::Reborrow)),
-            this.machine.borrow_tracker.as_ref().unwrap(),
-            alloc_id,
-            this.machine.current_span(),
-        )?;
+        if new_perm.initial_read {
+            tree_borrows.perform_access(
+                orig_tag,
+                Some((range, AccessKind::Read, diagnostics::AccessCause::Reborrow)),
+                this.machine.borrow_tracker.as_ref().unwrap(),
+                alloc_id,
+                this.machine.current_span(),
+            )?;
+        }
         // Record the parent-child pair in the tree.
         tree_borrows.new_child(
             orig_tag,
@@ -308,7 +316,7 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         drop(tree_borrows);
 
         // Also inform the data race model (but only if any bytes are actually affected).
-        if range.size.bytes() > 0 {
+        if range.size.bytes() > 0 && new_perm.initial_read {
             if let Some(data_race) = alloc_extra.data_race.as_ref() {
                 data_race.read(
                     alloc_id,
@@ -535,6 +543,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             initial_state: Permission::new_reserved(ty_is_freeze, /* protected */ true),
             zero_size: false,
             protector: Some(ProtectorKind::StrongProtector),
+            initial_read: true,
         };
         this.tb_retag_place(place, new_perm)
     }
--- a/src/borrow_tracker/tree_borrows/perms.rs
+++ b/src/borrow_tracker/tree_borrows/perms.rs
@@ -8,6 +8,10 @@ use crate::borrow_tracker::tree_borrows::tree::AccessRelatedness;
 /// The activation states of a pointer.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 enum PermissionPriv {
+    /// represents: a shared reference to interior mutable data.
+    /// allows: all foreign and child accesses;
+    /// rejects: nothing
+    Cell,
     /// represents: a local mutable reference that has not yet been written to;
     /// allows: child reads, foreign reads;
     /// affected by: child writes (becomes Active),
@@ -60,6 +64,14 @@ impl PartialOrd for PermissionPriv {
         use Ordering::*;
         Some(match (self, other) {
             (a, b) if a == b => Equal,
+            // Versions of `Reserved` with different interior mutability are incomparable with each
+            // other.
+            (ReservedIM, ReservedFrz { .. })
+            | (ReservedFrz { .. }, ReservedIM)
+            // `Cell` is not comparable with any other permission
+            // since it never transitions to any other state and we
+            // can never get to `Cell` from another state.
+            | (Cell, _) | (_, Cell) => return None,
             (Disabled, _) => Greater,
             (_, Disabled) => Less,
             (Frozen, _) => Greater,
@@ -71,9 +83,6 @@ impl PartialOrd for PermissionPriv {
                 // `bool` is ordered such that `false <= true`, so this works as intended.
                 c1.cmp(c2)
             }
-            // Versions of `Reserved` with different interior mutability are incomparable with each
-            // other.
-            (ReservedIM, ReservedFrz { .. }) | (ReservedFrz { .. }, ReservedIM) => return None,
         })
     }
 }
@@ -81,17 +90,22 @@ impl PartialOrd for PermissionPriv {
 impl PermissionPriv {
     /// Check if `self` can be the initial state of a pointer.
     fn is_initial(&self) -> bool {
-        matches!(self, ReservedFrz { conflicted: false } | Frozen | ReservedIM)
+        matches!(self, ReservedFrz { conflicted: false } | Frozen | ReservedIM | Cell)
     }
 
     /// Reject `ReservedIM` that cannot exist in the presence of a protector.
     fn compatible_with_protector(&self) -> bool {
-        !matches!(self, ReservedIM)
+        // FIXME(TB-Cell): It is unclear what to do here.
+        // `Cell` will occur with a protector but won't provide the guarantees
+        // of noalias (it will fail the `protected_enforces_noalias` test).
+        !matches!(self, ReservedIM | Cell)
     }
 
     /// See `foreign_access_skipping.rs`. Computes the SIFA of a permission.
     fn strongest_idempotent_foreign_access(&self, prot: bool) -> IdempotentForeignAccess {
         match self {
+            // Cell survives any foreign access
+            Cell => IdempotentForeignAccess::Write,
             // A protected non-conflicted Reserved will become conflicted under a foreign read,
             // and is hence not idempotent under it.
             ReservedFrz { conflicted } if prot && !conflicted => IdempotentForeignAccess::None,
@@ -124,7 +138,7 @@ mod transition {
             Disabled => return None,
             // The inner data `ty_is_freeze` of `Reserved` is always irrelevant for Read
             // accesses, since the data is not being mutated. Hence the `{ .. }`.
-            readable @ (ReservedFrz { .. } | ReservedIM | Active | Frozen) => readable,
+            readable @ (Cell | ReservedFrz { .. } | ReservedIM | Active | Frozen) => readable,
         })
     }
 
@@ -132,6 +146,8 @@ mod transition {
     /// is protected; invalidate `Active`.
     fn foreign_read(state: PermissionPriv, protected: bool) -> Option<PermissionPriv> {
         Some(match state {
+            // Cell ignores foreign reads.
+            Cell => Cell,
             // Non-writeable states just ignore foreign reads.
             non_writeable @ (Frozen | Disabled) => non_writeable,
             // Writeable states are more tricky, and depend on whether things are protected.
@@ -167,6 +183,8 @@ mod transition {
     /// write permissions, `Frozen` and `Disabled` cannot obtain such permissions and produce UB.
     fn child_write(state: PermissionPriv, protected: bool) -> Option<PermissionPriv> {
         Some(match state {
+            // Cell ignores child writes.
+            Cell => Cell,
             // If the `conflicted` flag is set, then there was a foreign read during
             // the function call that is still ongoing (still `protected`),
             // this is UB (`noalias` violation).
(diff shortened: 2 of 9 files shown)


RalfJung:
Testing if this works:
r? @RalfJung

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -123,6 +123,10 @@ struct NewPermission {
     /// Whether this pointer is part of the arguments of a function call.
     /// `protector` is `Some(_)` for all pointers marked `noalias`.
     protector: Option<ProtectorKind>,
+    /// Whether a read should be performed on a retag.  This should be `false`
+    /// for `Cell` because this could cause data races when using thread-safe
+    /// data types like `Mutex<T>`.
+    should_do_initial_read: bool,
```
RalfJung, line 129: Just `initial_read` should be enough as a field name here.

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -278,11 +302,27 @@ impl Permission {
...
 
+    /// What kind of access to perform before releasing the protector.
+    pub fn protector_end_access(&self) -> Option<AccessKind> {
+        Some(match self.inner {
+            // Do not do perform access if it is a `Cell`, as this
+            // can cause data races when using thread-safe data types.
+            Cell => return None,
+            Active => AccessKind::Write,
+            _ => AccessKind::Read,
+        })
```
RalfJung, line 323:
```suggestion
        match self.inner {
            // Do not do perform access if it is a `Cell`, as this
            // can cause data races when using thread-safe data types.
            Cell => None,
            Active => Some(AccessKind::Write),
            _ => Some(AccessKind::Read),
        }
```

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -306,18 +346,20 @@ impl Permission {
     /// remove protected parents.
     pub fn can_be_replaced_by_child(self, child: Self) -> bool {
         match (self.inner, child.inner) {
-            // ReservedIM can be replaced by anything, as it allows all
+            // Cell allows all transitions.
+            (Cell, _) => true,
+            // ReservedIM can be replaced by anything besides Cell, as it allows all
```
RalfJung, line 351: "it" here is ambiguous, does it refer to Cell or ReservedIM?

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -306,18 +346,20 @@ impl Permission {
...
             // transitions.
+            (ReservedIM, Cell) => false,
             (ReservedIM, _) => true,
             // Reserved (as parent, where conflictedness does not matter)
-            // can be replaced by all but ReservedIM,
-            // since ReservedIM alone would survive foreign writes
-            (ReservedFrz { .. }, ReservedIM) => false,
+            // can be replaced by all but ReservedIM and Cell,
+            // since ReservedIM and Cell alone would survive foreign writes
+            (ReservedFrz { .. }, ReservedIM) | (ReservedFrz { .. }, Cell) => false,
```
RalfJung, line 358: This seems to accumulate the "states handled above". Instead of doing that, would it make sense to do something like this?
```rust
// Nothing can be replaced by Cell.
(_, Cell) => false,
// ReservedIM can be replaced by anything (besides Cell, handled above).
(ReservedIM, _) => true,
```
Then we don't have to repeat `Cell ` a few more times below, right?

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -459,7 +503,7 @@ pub mod diagnostics {
         ///   (Reserved < Active < Frozen < Disabled);
         /// - between `self` and `err` the permission should also be increasing,
         ///   so all permissions inside `err` should be greater than `self.1`;
-        /// - `Active` and `Reserved(conflicted=false)` cannot cause an error
+        /// - `Active`, `Reserved(conflicted=false)` and `Cell` cannot cause an error
```
RalfJung, line 506:
```suggestion
        /// - `Active`, `Reserved(conflicted=false)`, and `Cell` cannot cause an error
```

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -492,13 +536,14 @@ pub mod diagnostics {
                         (ReservedFrz { conflicted: true } | Active | Frozen, Disabled) => false,
                         (ReservedFrz { conflicted: true }, Frozen) => false,
 
-                        // `Active` and `Reserved` have all permissions, so a
+                        // `Active`, `Reserved` and `Cell` have all permissions, so a
```
RalfJung, line 539:
```suggestion
                        // `Active`, `Reserved`, and `Cell` have all permissions, so a
```

tests/pass/tree_borrows/cell-local-after-foreign.rs
```diff
@@ -0,0 +1,60 @@
...
+        self.set(100);
+    }
+
+    // local read after foreign write
+    fn lr_fw(&mut self, _val: ()) {
+        self.get();
+    }
+
+    // local read after foreign read
+    #[allow(path_statements)]
```
RalfJung, line 32: Please don't allow the lint.

Just rename `val` to `_val` if you don't intend to use it.

tests/pass/tree_borrows/cell-local-after-foreign.rs
```diff
@@ -0,0 +1,60 @@
+//@compile-flags: -Zmiri-tree-borrows
+
+// Perform foreign access to a Cell and then local access.
+// Takes advantage of two-phase borrows.
```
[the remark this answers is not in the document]
yoctocell: Now I see that there is also `2phase-interiormut.rs` which does something
similar to this test.  I guess I should add tests that perform foreign
accesses on the `Cell` state.  The current interior mutability tests
almost always do `&mut UnsafeCell` which is in the `ResIM` state.

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -71,27 +83,28 @@ impl PartialOrd for PermissionPriv {
...
     fn is_initial(&self) -> bool {
-        matches!(self, ReservedFrz { conflicted: false } | Frozen | ReservedIM)
+        matches!(self, ReservedFrz { conflicted: false } | Frozen | ReservedIM | Cell)
     }
 
     /// Reject `ReservedIM` that cannot exist in the presence of a protector.
     fn compatible_with_protector(&self) -> bool {
-        !matches!(self, ReservedIM)
+        // TODO `Cell` will occur with a protector but won't provide the guarantees
+        // of noalias (it will fail the `protected_enforces_noalias` test).
```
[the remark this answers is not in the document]
yoctocell: I added this as a workaround/hack, since without it, the
`borrow_tracker::tree_borrows::tree::tests::protected_enforces_noalias`
test would fail because it checks that "if 2 accesses happen, one
foreign and one a child, and we are protected, that we get UB unless
they are both reads", but this should not be true for the `Cell` state.
I think an alternative to this could be to change the
`protected_enforces_noalias` test to not check for `Cell`, but I don't remember if there were
also other tests that also failed because of the same reason.

tests/pass/tree_borrows/cell-local-after-foreign.rs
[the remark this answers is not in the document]
RalfJung: That file is for 2-phase borrows, so no that's not the right file.

Turns out we have `tests/pass/stacked-borrows/interior_mutability.rs` but that never got copied to the TB test suite for some reason... oops. I will add that file for TB as well. Please take a look at the file and see if it covers all the cases you wanted to cover here.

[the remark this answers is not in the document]
RalfJung: Please rebase over latest master, then you should have a good first set of tests -- feel free to extend them, of course. :)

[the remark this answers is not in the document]
yoctocell: I think the tests in `interior_mutability.rs` cover enough things, so I have removed the `cell-local-after-foreign.rs` file.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -721,9 +721,12 @@ impl<'tcx> Tree {
                     // visit all children, skipping none
                     |_| ContinueTraversal::Recurse,
                     |args: NodeAppArgs<'_>| -> Result<(), TransitionError> {
-                        let NodeAppArgs { node, .. } = args;
+                        let NodeAppArgs { node, perm, .. } = args;
+                        let perm = perm.get().copied().unwrap_or_else(|| node.default_location_state());
                         if global.borrow().protected_tags.get(&node.tag)
                             == Some(&ProtectorKind::StrongProtector)
+                            // TODO Think about what the right semantics should be.  What if only part of the range is Cell.
+                            && !perm.permission().is_cell()
```
JoJoDeveloping, line 730: @RalfJung do you have opinions on what the right semantics should be here? Can you deallocated protected `Cell`s?

RalfJung: I would not introduce a special case unless we have a good reason for this. So for now I'd say this should be UB.

JoJoDeveloping: This would break [this test](https://github.com/rust-lang/miri/blob/master/tests/pass/tree_borrows/interior_mutability.rs#L82).

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -306,18 +346,20 @@ impl Permission {
...
-            // since ReservedIM alone would survive foreign writes
+            // can be replaced by all but ReservedIM and Cell,
+            // since ReservedIM and Cell alone would survive foreign writes
             (ReservedFrz { .. }, ReservedIM) => false,
             (ReservedFrz { .. }, _) => true,
             // Active can not be replaced by something surviving
             // foreign reads and then remaining writable.
-            (Active, ReservedIM) => false,
-            (Active, ReservedFrz { .. }) => false,
+            (Active, ReservedIM) | (Active, ReservedFrz { .. }) => false,
```
RalfJung, line 362:
```suggestion
            (Active, ReservedIM | ReservedFrz { .. }) => false,
```

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -306,18 +346,20 @@ impl Permission {
...
     pub fn can_be_replaced_by_child(self, child: Self) -> bool {
         match (self.inner, child.inner) {
-            // ReservedIM can be replaced by anything, as it allows all
-            // transitions.
+            // Cell allows all transitions.
+            (Cell, _) => true,
+            (_, Cell) => false,
+            // ReservedIM can be replaced by anything besides Cell, as
+            // ReservedIM allows all transitions.
             (ReservedIM, _) => true,
```
RalfJung, line 354: Please consistently follow the structure that you use with `Cell`: there should be `(_, ReservedIM) => false` here.

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -71,27 +83,28 @@ impl PartialOrd for PermissionPriv {
...
     fn is_initial(&self) -> bool {
-        matches!(self, ReservedFrz { conflicted: false } | Frozen | ReservedIM)
+        matches!(self, ReservedFrz { conflicted: false } | Frozen | ReservedIM | Cell)
     }
 
     /// Reject `ReservedIM` that cannot exist in the presence of a protector.
     fn compatible_with_protector(&self) -> bool {
-        !matches!(self, ReservedIM)
+        // FIXME(TB-Cell) `Cell` will occur with a protector but won't provide the guarantees
+        // of noalias (it will fail the `protected_enforces_noalias` test).
```
RalfJung, line 99:
```suggestion
        // FIXME(TB-Cell): It is unclear what to do here.
        // `Cell` will occur with a protector but won't provide the guarantees
        // of noalias (it will fail the `protected_enforces_noalias` test).
```

tests/pass/tree_borrows/interior_mutability.rs
RalfJung, line 1: Good catch, thanks for the update!

RalfJung:
@rustbot author

Also, why is this marked as "draft"? Since you were asking for a review, I assume this is ready to land, so please update the PR status accordingly. Usually "draft" means "not yet ready for review".

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -306,18 +346,20 @@ impl Permission {
     /// remove protected parents.
     pub fn can_be_replaced_by_child(self, child: Self) -> bool {
         match (self.inner, child.inner) {
-            // ReservedIM can be replaced by anything, as it allows all
-            // transitions.
+            // Cell allows all transitions.
+            (Cell, _) => true,
+            (_, Cell) => false,
+            // ReservedIM can be replaced by anything besides Cell, as
+            // ReservedIM allows all transitions.
```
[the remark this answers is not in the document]
yoctocell: `ReservedIM` cannot be replaced by `Cell` because when doing a local
write, `ReservedIM` transitions to `Active`, while `Cell` doesn't
change.  This would break the `tree_compacting_is_sound` test.

I will add this to the comment.

src/borrow_tracker/tree_borrows/perms.rs
yoctocell: Should I also do this with the other states (`ReservedFrz`, `Active`,
`Frozen`)?  This style would make things a little less explicit

The following

```rust
// Active can not be replaced by something surviving
// foreign reads and then remaining writable.
(Active, ReservedFrz { .. }) => false,
// Replacing a state by itself is always okay, even if the child state is protected.
(Active, Active) => true,
// Active can be replaced by Frozen, since it is not protected.
(Active, Frozen) => true,
(Active, Disabled) => true,
// Frozen can only be replaced by Disabled (and itself).
(Frozen, Frozen) => true,
(Frozen, Disabled) => true,
(Frozen, _) => false,
// Disabled can not be replaced by anything else.
(Disabled, Disabled) => true,
(Disabled, _) => false,
```

would become

```rust
// Active can not be replaced by something surviving
// foreign reads and then remaining writable.
(_, ReservedFrz { .. }) => false,
// Replacing a state by itself is always okay, even if the child state is protected.
// Active can be replaced by Frozen, since it is not protected.
(Active, _) => true,
(_, Active) => false,
// Frozen can only be replaced by Disabled (and itself).
(Frozen, _) => true,
(_, Frozen) => false,
// Disabled can not be replaced by anything else.
(Disabled, _) => true,
```

Or just do this with `ReservedIM` and keep the rest as-is?

RalfJung: Yes I think you should do it for all states.

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -348,30 +349,30 @@ impl Permission {
...
+            (_, ReservedIM) => false,
             // Reserved (as parent, where conflictedness does not matter)
             // can be replaced by all but ReservedIM and Cell,
             // since ReservedIM and Cell alone would survive foreign writes
-            (ReservedFrz { .. }, ReservedIM) => false,
             (ReservedFrz { .. }, _) => true,
             // Active can not be replaced by something surviving
             // foreign reads and then remaining writable.
-            (Active, ReservedIM) | (Active, ReservedFrz { .. }) => false,
+            (_, ReservedFrz { .. }) => false,
```
RalfJung, line 366: The comment seems to be misplaced now?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -726,7 +726,8 @@ impl<'tcx> Tree {
                             perm.get().copied().unwrap_or_else(|| node.default_location_state());
                         if global.borrow().protected_tags.get(&node.tag)
                             == Some(&ProtectorKind::StrongProtector)
-                            // TODO Think about what the right semantics should be.  What if only part of the range is Cell.
+                            // Don't check for protector if it is a Cell (see `unsafe_cell_deallocate` in `interior_mutability.rs`)
```
RalfJung, line 729:
```suggestion
                            // Don't check for protector if it is a Cell (see `unsafe_cell_deallocate` in `interior_mutability.rs`).
```

src/borrow_tracker/tree_borrows/perms.rs
yoctocell: Hm, now the `_` matches on `Active, `Frozen`, and `Disabled`, so it looks a bit weird. I think we can either remove the comment, or add a line `(Active, ReservedFrz { .. }) => false,` above line 366 to make it more explicit.

RalfJung: Isn't the comment just one line too far up?
The typical pattern is to have the `(_, ...) => false` line just after the `=> true` line -- that basically fully handles all transitions from and to `...`.

yoctocell: If we move the comment below ` (_, ReservedFrz { .. }) => false`, then we would have

```rust
            (ReservedFrz { .. }, _) => true,
            (_, ReservedFrz { .. }) => false, // same as (Active | Frozen | Disabled, ReservedFrz { .. })
            // Active can not be replaced by something surviving
            // foreign reads and then remaining writable.
            // Replacing a state by itself is always okay, even if the child state is protected.
            // Active can be replaced by Frozen, since it is not protected.
            (Active, _) => true,
            (_, Active) => false,
```

But the comment refers to the "Active" on the line above it (in the comment that I added), not to the "Active" on the lines below.  Does this make sense?

Maybe we can move up the comment (reworded slightly) like this?

```rust
            // Active can not be replaced by Reserved, because Reserved remains
            // writable after foreign reads, while Active does not.
            (ReservedFrz { .. }, _) => true,
            (_, ReservedFrz { .. }) => false,
```

RalfJung: > But the comment refers to the "Active" on the line above it (in the comment that I added), not to the "Active" on the lines below. 

What about
```rust
            // Active can not be replaced by something surviving
            // foreign reads and then remaining writable (i.e., Reserved*).
            // Replacing a state by itself is always okay, even if the child state is protected.
            // Active can be replaced by Frozen, since it is not protected.
            (Active, Active | Frozen | Disabled) => true,
```

RalfJung:

@rustbot author

RalfJung:
It seems like something went wrong with the rebase. Make sure you have the latest upstream commit fetched into your local clone.

RalfJung:
It depends on your setup. For me, it's `git fetch origin`, but you have to replace "origin" with whatever the name of the Miri upstream remote (https://github.com/rust-lang/miri, not your clone) is.

yoctocell:
Ok, now I have fetched the latest commit from the upstream remote. What would be the next step?

RalfJung:
I have linked the tutorial for rebasing above: https://rustc-dev-guide.rust-lang.org/git.html#squash-your-commits

yoctocell:
Running `git rebase origin/master --keep-base` outputs `Current branch new-cell-state is up to date.`.  Is this expected?

RalfJung:
Hm that doesn't check out, your branch is not based on `origin/master`. Maybe try without `--keep-base`?

Or you can check `git reflog` for the commit ID *before* the broken rebase, to reset to a good state, and then try again.

RalfJung:
The commit ID before the broken rebase seems to be 64b11de958c114f6de837a25daec9ba8eda6d971, so the best fix is probably `git reset --hard 64b11de958c114f6de837a25daec9ba8eda6d971`, and then trying again.

yoctocell:
Doing `git reset` seemed to fix things :)

RalfJung:
Great. :)

PR #4275, opened 2025-04-17
Implement file read/write/seek in Windows

CraftSpider:
I can split seek off if desired, though that means I can't enable the whole file read/write shim test in `fs.rs` in one go

### DIFF
--- a/src/shims/windows/fs.rs
+++ b/src/shims/windows/fs.rs
@@ -1,5 +1,6 @@
 use std::fs::{Metadata, OpenOptions};
 use std::io;
+use std::io::SeekFrom;
 use std::path::PathBuf;
 use std::time::SystemTime;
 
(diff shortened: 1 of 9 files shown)


src/shims/windows/handle.rs
```diff
@@ -224,6 +225,35 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         )))
     }
 
+    fn GetStdHandle(&mut self, which: &OpTy<'tcx>) -> InterpResult<'tcx, Scalar> {
```
RalfJung, line 228: This function looks to not just get a handle, it creates new handles (that must be closed) -- quit surprising given the function name. I guess the name is what the Windows API uses, but please at least add a doc comment.

src/shims/windows/fs.rs
```diff
@@ -390,6 +391,253 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        key: &OpTy<'tcx>,             // PULONG
+    ) -> InterpResult<'tcx, Scalar> {
+        // ^ Returns NTSTATUS (u32 on Windows)
+        let this = self.eval_context_mut();
+        let handle = this.read_handle(handle, "NtWriteFile")?;
+        let event = this.read_handle(event, "NtWriteFile")?;
+        let apc_routine = this.read_pointer(apc_routine)?;
+        let apc_ctx = this.read_pointer(apc_ctx)?;
+        let buf = this.read_pointer(buf)?;
+        let count = this.read_scalar(n)?.to_u32()? as usize;
```
RalfJung, line 414: Please avoid `as` casts. Use `try_into().unwrap()` instead -- and do it below where we call `dest.write`.

src/shims/windows/fs.rs
```diff
@@ -390,6 +391,253 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        }
+
+        let fd = match handle {
+            Handle::File(fd) => fd,
+            _ => this.invalid_handle("NtWriteFile")?,
+        };
+
+        let Some(desc) = this.machine.fds.get(fd) else { this.invalid_handle("NtWriteFile")? };
+
+        let io_status = {
```
RalfJung, line 455: Please add some comments so people that have never seen the windows API (likely the majority of Miri maintainers ;) have some idea of what these things do. I can infer it from the code but that's not the best kind of documentation.

src/shims/windows/fs.rs
```diff
@@ -390,6 +391,253 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                            this.write_int(error_kind_to_ntstatus(e), &io_status)
+                        }
+                }}
+            )
+        };
+
+        desc.write(this.machine.communicate(), buf, count, this, finish)?;
+
+        // Return whether this was a success. 0 is success.
+        interp_ok(this.read_scalar(&io_status)?)
```
RalfJung, line 489: The comment doesn't explain why we just return the contents of that place?

src/shims/windows/fs.rs
```diff
@@ -390,6 +391,253 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        let fd = match handle {
+            Handle::File(fd) => fd,
+            _ => this.invalid_handle("NtWriteFile")?,
+        };
+
+        let Some(desc) = this.machine.fds.get(fd) else { this.invalid_handle("NtReadFile")? };
+
+        // TODO: Windows expects this function to return its error, not set the global one
+        desc.read(this.machine.communicate(), buf, count, this, finish)?;
```
RalfJung, line 582: Does this comment only apply to `read`, not `write`?

Also, what to do with the error is entirely up to you -- the callback receives the error info. Your callback doesn't seem to set the global error so I'm not sure what the comment refers to?

Seems worth having a test which checks that this does indeed not clobber the global error.

src/shims/windows/fs.rs
```diff
@@ -419,3 +667,24 @@ fn write_filetime_field<'tcx>(
         &cx.project_field_named(val, name)?,
     )
 }
+
+fn error_kind_to_ntstatus(e: IoError) -> i32 {
```
RalfJung, line 671: This looks like it should go into `src/shims/io_error.rs`?

RalfJung (commented):
Looks great overall, thanks! Just some minor comments. Nice to see that the blocking read/write API works for the Windows functions as well.

@rustbot author

src/shims/windows/fs.rs
CraftSpider: Accidentally copied over from an older version of the code. I'll add a test for that behavior.

src/shims/windows/handle.rs
CraftSpider: So, double-checking the docs, this is... still a somewhat wrong implementation. GetStdHandle returns a 'real' handle, but it should actually be the same one each time. You can close it, which causes _all_ handles across the whole process to become invalid. This is intentional, because in a real system you can replace the value of these handles with other ones using `SetStdHandle`.

RalfJung: Yeah, that's closer to what I expected.

RalfJung:
@CraftSpider what's the status of this? A standard library test started using `NtReadFile` directly, and I think this PR would let us run that test (rather than having to disable it in Miri which I will do now). :)

RalfJung:
There's a FIXME added in https://github.com/rust-lang/rust/pull/141210 that we can hopefully remove when this lands.

CraftSpider:
(modulo tests passing)
@rustbot ready

src/shims/windows/fs.rs
```diff
@@ -463,29 +467,31 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             )
         };
 
-        desc.write(this.machine.communicate(), buf, count, this, finish)?;
+        desc.write(this.machine.communicate(), buf, count.try_into().unwrap(), this, finish)?;
 
-        // Return whether this was a success. 0 is success.
+        // Return whether this was a success. 0 is a success.
+        // We wrote the return status into the IO_STATUS_BLOCK in the above callback,
+        // just pull it from there.
```
RalfJung, line 494: This doesn't work -- the entire reason that that's a callback is that `desc.write` may block the current thread, return immediately, and run the callback later!

This function needs to take a `dest` argument to capture the return place, pass that into the callback, and then write the return value that way.

src/shims/windows/fs.rs
```diff
@@ -578,10 +585,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
         let Some(desc) = this.machine.fds.get(fd) else { this.invalid_handle("NtReadFile")? };
 
-        // TODO: Windows expects this function to return its error, not set the global one
-        desc.read(this.machine.communicate(), buf, count, this, finish)?;
+        desc.read(this.machine.communicate(), buf, count.try_into().unwrap(), this, finish)?;
 
-        // Return whether this was a success. 0 is success.
+        // See NtWriteFile for commentary on this
         interp_ok(this.read_scalar(&io_status)?)
```
RalfJung: Same here regarding the callback being potentially invoked later.

src/shims/windows/handle.rs
```diff
@@ -230,19 +230,11 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let stderr = this.eval_windows("c", "STD_ERROR_HANDLE").to_i32()?;
 
         let fd_num = if which == stdin {
```
RalfJung, line 231: Please add a comment saying where those numbers come from. AFAIK those are the typical Unix numbers, right? Miri just happens to set them up always, even on Windows.

RalfJung (commented):
Actually it seems there's one pretty big issue, see the comments regarding when the callback is called...
@rustbot author

src/shims/windows/handle.rs
CraftSpider: Will do. And yeah, this is just taking advantage of the existing setup. The user won't even see these values directly, due to handle encoding.

src/shims/windows/fs.rs
CraftSpider: This took me a moment to get - for future reference when I'm fixing this: if the callback doesn't happen immediately, we will return the (possibly uninit) value that is in the IO_STATUS_BLOCK we were passed, then before the user code resumes, the callback will happen, write the value into IO_STATUS_BLOCK, then resume execution. We have to pass the return place into the callback to avoid that problem.

This is fairly obvious in retrospect, but it wasn't clear to me until I reread your comment a couple times.

src/shims/windows/handle.rs
CraftSpider: If we added support for `SetStdHandle`, then we'd need to add state to track the current values of these handles. But since we don't support that, this will just only ever return these values. Adding such support is completely possible, but not really important for `std` as far as I can tell.

src/shims/windows/fs.rs
RalfJung: Yeah, "blocking" calls in Miri are somewhat funny in how they behave, though given that it is a single-threaded interpreter simulating concurrency, that's the "obvious" architecture.

RalfJung:
I assume this is
@rustbot ready

src/shims/windows/fs.rs
```diff
@@ -390,6 +391,269 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        apc_routine: &OpTy<'tcx>,     // PIO_APC_ROUTINE
+        apc_ctx: &OpTy<'tcx>,         // PVOID
+        io_status_block: &OpTy<'tcx>, // PIO_STATUS_BLOCK
+        buf: &OpTy<'tcx>,             // PVOID
+        n: &OpTy<'tcx>,               // ULONG
+        byte_offset: &OpTy<'tcx>,     // PLARGE_INTEGER
+        key: &OpTy<'tcx>,             // PULONG
+        dest: &MPlaceTy<'tcx>,        // NTSTATUS
+    ) -> InterpResult<'tcx, ()> {
+        // ^ Returns NTSTATUS (u32 on Windows)
```
RalfJung, line 408:
```suggestion
        dest: &MPlaceTy<'tcx>,        // return type: NTSTATUS
    ) -> InterpResult<'tcx, ()> {
```
Same for the other functions with a `dest` parameter

src/shims/windows/fs.rs
```diff
@@ -390,6 +391,269 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let this = self.eval_context_mut();
+        let handle = this.read_handle(handle, "NtWriteFile")?;
+        let event = this.read_handle(event, "NtWriteFile")?;
+        let apc_routine = this.read_pointer(apc_routine)?;
+        let apc_ctx = this.read_pointer(apc_ctx)?;
+        let buf = this.read_pointer(buf)?;
+        let count = this.read_scalar(n)?.to_u32()?;
+        let byte_offset = this.read_target_usize(byte_offset)?; // is actually a pointer
```
RalfJung, line 416:
```suggestion
        let byte_offset = this.read_target_usize(byte_offset)?; // is actually a pointer, but we only support null
```

src/shims/windows/fs.rs
```diff
@@ -390,6 +391,269 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        if event != Handle::Null {
+            throw_unsup_format!(
+                "`NtWriteFile` `Event` parameter is non-null, which is unsupported"
+            );
+        }
+
+        if !this.ptr_is_null(apc_routine)? {
+            throw_unsup_format!(
+                "`NtWriteFile` `ApcRoutine` parameter is not null, which is unsupported"
```
RalfJung, line 429:
```suggestion
                "`NtWriteFile` `ApcRoutine` parameter is non-null, which is unsupported"
```
Some messages use "not null", some "non-null", let's be consistent (applies a few more times below)

RalfJung:
This looks great, thanks! After fixing the last nits, please squash the commits using `./miri squash`. Then write `@rustbot ready` after you pushed the squashed PR.

@rustbot author

CraftSpider:
`miri squash` is failing locally on my Windows machine, I'll try to fix it tonight...

RalfJung:
Ah, bummer... you can do a manual squash instead, this was just meant to make things easier.

RalfJung:
I assume you meant to
@rustbot ready

PR #4287, opened 2025-04-22
implement cpuset_getaffinity for freebsd

### LINKED ISSUE #4265: Support num_cpus on FreeBSD
Almost all tests pass on FreeBSD now, with the exception of:
```
FAILURES:
    tests/pass/shims/available-parallelism-miri-num-cpus.rs
    tests/pass/shims/available-parallelism.rs (revision stack)
    tests/pass/shims/available-parallelism.rs (revision tree)
```
This is caused by a lack of `cpuset_getaffinity`. We already support a similar function `sched_getaffinity`  -- I don't know how similar the APIs are, though.

LorrensP-2158466:
Fixes rust-lang/miri#4265.

This implements the FreeBSD syscall `cpuset_getaffinity`. In practice, this syscall behaves like a more fine-grained api than `sched_getaffinity`.

The tests only require one specific combination of parameters to pass, so this is all that is implemented. This can be done in future work (or now if needed).

However, I had to change 1 test file `available-parallelism-miri-num-cpus`. This test sets the number of CPUs to 1024, but the libc API is structured so that only 256 can be accepted. (I'll explain in a comment.)

I have not changed the ci script since I don't know what I should put there.

### DIFF
--- a/src/shims/unix/freebsd/foreign_items.rs
+++ b/src/shims/unix/freebsd/foreign_items.rs
@@ -56,6 +56,70 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.write_scalar(res, dest)?;
             }
 
+            "cpuset_getaffinity" => {
+                // The "same" kind of api as `sched_getaffinity` but more fine grained control for FreeBSD specifically.
+                let [level, which, id, set_size, mask] =
+                    this.check_shim(abi, Conv::C, link_name, args)?;
+
+                let level = this.read_scalar(level)?.to_i32()?;
+                let which = this.read_scalar(which)?.to_i32()?;
+                let id = this.read_scalar(id)?.to_i64()?;
+                let set_size = this.read_target_usize(set_size)?; // measured in bytes
+                let mask = this.read_pointer(mask)?;
+
+                let _level_root = this.eval_libc_i32("CPU_LEVEL_ROOT");
+                let _level_cpuset = this.eval_libc_i32("CPU_LEVEL_CPUSET");
+                let level_which = this.eval_libc_i32("CPU_LEVEL_WHICH");
+
+                let _which_tid = this.eval_libc_i32("CPU_WHICH_TID");
+                let which_pid = this.eval_libc_i32("CPU_WHICH_PID");
+                let _which_jail = this.eval_libc_i32("CPU_WHICH_JAIL");
+                let _which_cpuset = this.eval_libc_i32("CPU_WHICH_CPUSET");
+                let _which_irq = this.eval_libc_i32("CPU_WHICH_IRQ");
+
+                // For sched_getaffinity, the current process is identified by -1.
+                // TODO: Use gettid? I'm (LorrensP-2158466) not that familiar with this api .
+                let id = match id {
+                    -1 => this.active_thread(),
+                    _ =>
+                        throw_unsup_format!(
+                            "`cpuset_getaffinity` is only supported with a pid of -1 (indicating the current thread)"
+                        ),
+                };
+
+                if this.ptr_is_null(mask)? {
+                    this.set_last_error_and_return(LibcError("EFAULT"), dest)?;
+                }
+                // We only support CPU_LEVEL_WHICH and CPU_WHICH_PID for now.
+                // This is the bare minimum to make the tests pass.
+                else if level != level_which || which != which_pid {
+                    throw_unsup_format!(
+                        "`cpuset_getaffinity` is only supported with `level` set to CPU_LEVEL_WHICH and `which` set to CPU_WHICH_PID."
+                    );
+                } else if let Some(cpuset) = this.machine.thread_cpu_affinity.get(&id) {
+                    // `cpusetsize` must be large enough to contain the entire CPU mask.
+                    // FreeBSD only uses `cpusetsize` to verify that it's sufficient for the kernel's CPU mask.
+                    // If it's too small, the syscall returns ERANGE.
+                    // If it's large enough, copying the kernel mask to user space is safe, regardless of the actual size.
+                    // See https://github.com/freebsd/freebsd-src/blob/909aa6781340f8c0b4ae01c6366bf1556ee2d1be/sys/kern/kern_cpuset.c#L1985
+                    if set_size < u64::from(this.machine.num_cpus).div_ceil(8) {
+                        this.set_last_error_and_return(LibcError("ERANGE"), dest)?;
+                    } else {
+                        let cpuset = cpuset.clone();
+                        let byte_count =
+                            Ord::min(cpuset.as_slice().len(), set_size.try_into().unwrap());
+                        this.write_bytes_ptr(
+                            mask,
+                            cpuset.as_slice()[..byte_count].iter().copied(),
+                        )?;
+                        this.write_null(dest)?;
+                    }
+                } else {
+                    // `id` is always that of the active thread, so this is currently unreachable.
+                    unreachable!();
+                }
+            }
+
             // Synchronization primitives
             "_umtx_op" => {
                 let [obj, op, val, uaddr, uaddr2] =
--- a/tests/pass-dep/shims/freebsd-cpuset-affinity.rs
+++ b/tests/pass-dep/shims/freebsd-cpuset-affinity.rs
@@ -1,0 +1,51 @@
+//@only-target: freebsd
+//@compile-flags: -Zmiri-num-cpus=256
+
+use std::mem;
+
+fn getaffinity() {
+    let mut set: libc::cpuset_t = unsafe { mem::zeroed() };
+    unsafe {
+        if libc::cpuset_getaffinity(
+            libc::CPU_LEVEL_WHICH,
+            libc::CPU_WHICH_PID,
+            -1,
+            size_of::<libc::cpuset_t>(),
+            &mut set,
+        ) == 0
+        {
+            assert!(libc::CPU_COUNT(&set) == 256);
+        }
+    }
+}
+
+fn get_small_cpu_mask() {
+    let mut set: libc::cpuset_t = unsafe { core::mem::MaybeUninit::zeroed().assume_init() };
+
+    // 256 CPUs so we need 32 bytes to represent this mask.
+    // According to Freebsd only when `cpusetsize` is smaller than this value, does it return with ERANGE
+
+    let err = unsafe {
+        libc::cpuset_getaffinity(libc::CPU_LEVEL_WHICH, libc::CPU_WHICH_PID, -1, 32, &mut set)
+    };
+    assert_eq!(err, 0, "Success Expected");
+
+    // 31 is not enough, so it should fail.
+    let err = unsafe {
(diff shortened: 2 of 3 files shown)


LorrensP-2158466:
So the tests use the libc function `libc::CPU_COUNT`, which counts how many bits are flipped to `1` in the cpuset. The source code that calls the syscall uses the struct `libc::cpuset_t`, and the size of this set struct is used to write the specified bytes to the mask pointer.

The source of this struct is this:
```rust
pub struct cpuset_t {
    #[cfg(all(any(freebsd15, freebsd14), target_pointer_width = "64"))]
    __bits: [c_long; 16],
    #[cfg(all(any(freebsd15, freebsd14), target_pointer_width = "32"))]
    __bits: [c_long; 32],
    #[cfg(all(not(any(freebsd15, freebsd14)), target_pointer_width = "64"))]
    __bits: [c_long; 4],
    #[cfg(all(not(any(freebsd15, freebsd14)), target_pointer_width = "32"))]
    __bits: [c_long; 8],
}
```

So on newer FreeBSD version, this struct is 128 bytes large, otherwise 32 bytes. During testing I have found out that the current compilation results in the 32 bytes variant. This means we can only "use" 256 CPUs since the size of the mask is: 32 * 8 = 256. Miri's maximum number is 1024 and the test also sets and expects this value.

I have no idea how to alter compilation to use the 128 size variant, which would result in 1024 bits being used. But the test passes when using `-Zmiri-num-cpus=256`.

tests/pass/shims/available-parallelism-miri-num-cpus.rs
RalfJung, line 1: We should still test that 1024 works on platforms that support larger numbers of CPUs.

RalfJung:
> I have no idea how to alter compilation to use the 128 size variant, which would result in 1024 bits being used. But the test passes when using -Zmiri-num-cpus=256.

The standard library is built for compatibility with FreeBSD 12, so we'd have to wait until that gets bumped to FreeBSD 14.

What happens on FreeBSD when setting num-cpus too high? We should at least fail somewhat gracefully.

RalfJung:
So `set_size` would be 32 bytes and we'd just not write the rest, truncating as usual? But no UB?

I'd say make that test have two revisions, one for 256 cores and one for 1024 cores. The 256 revision should always give the right reply, and the 1024 revision will then presumably just return 256 on FreeBSD.

LorrensP-2158466:
> So set_size would be 32 bytes and we'd not write the rest, truncating as usual? But no UB?

Nothing is truncated per se, but the full mask that Miri holds can't be written to the user-provided mask, writing more causes UB:
```
   |
LL |                       if libc::cpuset_getaffinity(
   |  ________________________^
LL | |                         libc::CPU_LEVEL_WHICH,
LL | |                         libc::CPU_WHICH_PID,
LL | |                         -1,
LL | |                         size_of::<libc::cpuset_t>(),
LL | |                         &mut set,
LL | |                     ) == 0 {
   | |_____________________^ memory access failed: expected a pointer to 128 bytes of memory, but got alloc994 which is only 32 bytes from the end of the allocation
```

> I'd say make that test have two revisions

Revision as in 2 test files?

RalfJung:
> Revision as in 2 test files?

No, as in one test file with 2 revisions. Grep the test suite for `revisions:` if you have not seen this yet.

> Nothing is truncated per se, but the full mask that Miri holds can't be written to the user-provided mask, writing more causes UB:

Wait, so calling this safe standard library method causes UB? That's bad. It's either a bug in Miri or a bug in the standard library.

LorrensP-2158466:
Alright thanks!

> Wait, so calling this safe standard library method causes UB? That's bad. It's either a bug in Miri or a bug in the standard library.

No, hahaha, I'm explaining this very badly. The implementation handles this correctly, either writing `set_size` bytes into the user mask, or writing the entire Mask held in Miri. It's just when I was implementing this I did cause a bug in Miri by writing the full Mask of Miri into the mask of the user. I did this because all those bits are needed for the test.

The correct thing to do here is this (in pseudo-code):
```rust
let byte_count = Ord::min(set_size, miri_cpu_set_of_thread.byte_len());
// copy byte_count amount of bytes from miri_cpu_set_of_thread into user_mask
```

Writing the full mask held in Miri *would* cause UB, because the user mask in older FreeBSD versions is much smaller.

Sorry for the bad explanation :(.

RalfJung:
Yeah, Miri needs to clamp somewhere, agreed. That's exactly why I want to see a test of "mask too small". :)

`tests/pass-dep/libc/libc-affinity.rs` has such a test for `sched_getaffinity`; we'll want similar tests directly calling `cpuset_getaffinity`.

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -101,6 +101,57 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    -1 => this.active_thread(),
+                    _ =>
+                        throw_unsup_format!(
+                            "`cpuset_getaffinity` is only supported with a pid of -1 (indicating the current process)"
+                        ),
+                };
+
+                // We only support CPU_LEVEL_WHICH and CPU_WHICH_PID for now.
+                // This is the bare minimum to make the tests pass.
+                // TODO: Support more.
```
RalfJung, line 137:
```suggestion
                // This is the bare minimum to make the tests pass.
```
There's no point in having such a TODO in the code IMO. If we want to track this, we can have an issue.

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -101,6 +101,57 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                        "`cpuset_getaffinity` is only supported with `level` set to CPU_LEVEL_WHICH and `which` set to CPU_WHICH_PID."
+                    );
+                } else if let Some(cpuset) = this.machine.thread_cpu_affinity.get(&id) {
+                    let cpuset = cpuset.clone();
+                    let byte_count = set_size.try_into().unwrap();
+                    this.write_bytes_ptr(mask, cpuset.as_slice()[..byte_count].iter().copied())?;
+                    this.write_null(dest)?;
+                } else {
+                    // The thread whose ID is pid could not be found.
+                    this.set_last_error_and_return(LibcError("ESRCH"), dest)?;
```
RalfJung, line 151: This should be `unreachable!()`, right? The ID is always that of the active thread, after all.

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -101,6 +101,57 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                if this.ptr_is_null(mask)? {
+                    this.set_last_error_and_return(LibcError("EFAULT"), dest)?;
+                } else if level != level_which || which != which_pid {
+                    throw_unsup_format!(
+                        "`cpuset_getaffinity` is only supported with `level` set to CPU_LEVEL_WHICH and `which` set to CPU_WHICH_PID."
+                    );
+                } else if let Some(cpuset) = this.machine.thread_cpu_affinity.get(&id) {
+                    let cpuset = cpuset.clone();
+                    let byte_count = set_size.try_into().unwrap();
+                    this.write_bytes_ptr(mask, cpuset.as_slice()[..byte_count].iter().copied())?;
```
RalfJung, line 147: Won't this ICE Miri if `byte_count` is bigger than the mask?

tests/pass/shims/available-parallelism-miri-num-cpus.rs
RalfJung, line 1: Please also add a test directly calling `cpuset_getaffinity`.

RalfJung (commented):
@rustbot author

src/shims/unix/freebsd/foreign_items.rs
RalfJung: The place where you copied this from likely supports other IDs, so it makes sense there, but it doesn't make sense to follow the same convention here.

LorrensP-2158466:
> Yeah, Miri needs to clamp somewhere, agreed. That's exactly why I want to see a test of "mask too small". :)
>tests/pass-dep/libc/libc-affinity.rs has such a test for sched_getaffinity; we'll want similar tests directly calling cpuset_getaffinity.

So I finally got my FreeBSD VM working again when I had some extra time, and tried testing what they do.

It only fails when the `cpusetsize` is less than the amount needed to represent the mask. So when using 4 CPUs on my VM, Anything less than 1 fails. (also tested with 64).

The code responsible for this is in `kern_cpuset_getaffinity`:
```c
if (error == 0) {
    if (cpusetsize < howmany(CPU_FLS(mask), NBBY))
	return (ERANGE);
```

This checks how many bytes are needed to represent a number of bits (number of CPUS). And `cpussetsize` is less than this number, `ERANGE` is returned, which is quite different than the Linux `sched_getaffinity`. I ported the Linux test of the small mask to FreeBSD, and it behaved differently. It only fails when `cpusetsize` is smaller than actually needed, not when smaller than the kernel mask.

So I did the same.

Please let me know if something is unclear.

LorrensP-2158466:
I think the implementation is done, the only thing that needs to change is the CI script.

tests/pass-dep/freebsd-cpuset-affinity.rs
RalfJung, line 1: Please move this into `pass-dep/shims`.

RalfJung (commented):
Looks good, mostly minor comments. :)

@rustbot author

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -101,6 +101,68 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                // This is the bare minimum to make the tests pass.
+                if this.ptr_is_null(mask)? {
+                    this.set_last_error_and_return(LibcError("EFAULT"), dest)?;
+                } else if level != level_which || which != which_pid {
+                    throw_unsup_format!(
+                        "`cpuset_getaffinity` is only supported with `level` set to CPU_LEVEL_WHICH and `which` set to CPU_WHICH_PID."
+                    );
+                } else if let Some(cpuset) = this.machine.thread_cpu_affinity.get(&id) {
+                    // `cpusetsize` should be big enough so the whole mask can be obtained.
+                    // FreeBSD doesn't actually handle cases where `cpusetsize` is bigger than the kernal mask
```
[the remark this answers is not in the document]
LorrensP-2158466: Good point, i'll make it clearer

tests/pass-dep/shims/freebsd-cpuset-affinity.rs
RalfJung, line 1: I think the test isn't actually being executed right now -- it has to be added to `ci.sh` as well. Just add `affinity` to the freebsd lines there.

RalfJung:
Nice. :)

This looks great, thanks! Once you added it to CI, please squash the commits using `./miri squash`. Then write `@rustbot ready` after you force-pushed the squashed PR.

@rustbot author

LorrensP-2158466:
That `./miri squash` command is really cool and handy!

@rustbot ready

RalfJung:
Ah fair, I forgot how old this PR is.^^ It's fine :)

PR #4291, opened 2025-04-24
GenMC implementation initial changes

Patrick-6:
In preparation for adding GenMC support, this PR contains a subset of changes that will be required.

The changes include:
- Adding an off-by-default `genmc` cargo feature
- Adding types `GenmcConfig` and `GenmcCtx` and methods that will be interfacing with GenMC
- Adding `GenmcCtx` to `MiriMachine`, changing the field `data_race` from an `Option` to a 3-variant enum
- Adding a method that will be parsing GenMC command line arguments

r? @RalfJung

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@ tex/*/out
 *.mm_profdata
 perf.data
 perf.data.old
-flamegraph.svg
+flamegraph*.svg
+rustc-ice*.txt
 tests/native-lib/libtestlib.so
 .auto-*
+
+/genmc/
--- a/src/bin/miri.rs
+++ b/src/bin/miri.rs
@@ -28,13 +28,14 @@ use std::env::{self, VarError};
 use std::num::NonZero;
 use std::ops::Range;
 use std::path::PathBuf;
+use std::rc::Rc;
 use std::str::FromStr;
 use std::sync::atomic::{AtomicI32, AtomicU32, Ordering};
 use std::sync::{Arc, Once};
 
 use miri::{
-    BacktraceStyle, BorrowTrackerMethod, MiriConfig, MiriEntryFnType, ProvenanceMode, RetagFields,
-    ValidationMode,
+    BacktraceStyle, BorrowTrackerMethod, GenmcConfig, GenmcCtx, MiriConfig, MiriEntryFnType,
+    ProvenanceMode, RetagFields, ValidationMode,
 };
 use rustc_abi::ExternAbi;
 use rustc_data_structures::sync;
@@ -60,6 +61,8 @@ use tracing::debug;
 struct MiriCompilerCalls {
     miri_config: Option<MiriConfig>,
     many_seeds: Option<ManySeedsConfig>,
+    /// Settings for using GenMC with Miri.
+    genmc_config: Option<GenmcConfig>,
 }
 
 struct ManySeedsConfig {
@@ -68,8 +71,12 @@ struct ManySeedsConfig {
 }
 
 impl MiriCompilerCalls {
-    fn new(miri_config: MiriConfig, many_seeds: Option<ManySeedsConfig>) -> Self {
-        Self { miri_config: Some(miri_config), many_seeds }
+    fn new(
+        miri_config: MiriConfig,
+        many_seeds: Option<ManySeedsConfig>,
+        genmc_config: Option<GenmcConfig>,
+    ) -> Self {
+        Self { miri_config: Some(miri_config), many_seeds, genmc_config }
     }
 }
 
@@ -179,6 +186,12 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
                     optimizations is usually marginal at best.");
         }
 
+        if let Some(genmc_config) = &self.genmc_config {
+            let _genmc_ctx = Rc::new(GenmcCtx::new(&config, genmc_config));
+
+            todo!("GenMC mode not yet implemented");
+        };
+
         if let Some(many_seeds) = self.many_seeds.take() {
             assert!(config.seed.is_none());
             let exit_code = sync::IntoDynSyncSend(AtomicI32::new(rustc_driver::EXIT_SUCCESS));
@@ -187,8 +200,14 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
                 let mut config = config.clone();
                 config.seed = Some((*seed).into());
                 eprintln!("Trying seed: {seed}");
-                let return_code = miri::eval_entry(tcx, entry_def_id, entry_type, config)
-                    .unwrap_or(rustc_driver::EXIT_FAILURE);
+                let return_code = miri::eval_entry(
+                    tcx,
+                    entry_def_id,
+                    entry_type,
+                    &config,
+                    /* genmc_ctx */ None,
+                )
+                .unwrap_or(rustc_driver::EXIT_FAILURE);
                 if return_code != rustc_driver::EXIT_SUCCESS {
                     eprintln!("FAILING SEED: {seed}");
                     if !many_seeds.keep_going {
@@ -206,11 +225,12 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
             }
             std::process::exit(exit_code.0.into_inner());
         } else {
-            let return_code = miri::eval_entry(tcx, entry_def_id, entry_type, config)
+            let return_code = miri::eval_entry(tcx, entry_def_id, entry_type, &config, None)
                 .unwrap_or_else(|| {
                     tcx.dcx().abort_if_errors();
                     rustc_driver::EXIT_FAILURE
                 });
+
             std::process::exit(return_code);
         }
 
@@ -506,6 +526,7 @@ fn main() {
     let mut many_seeds_keep_going = false;
     let mut miri_config = MiriConfig::default();
     miri_config.env = env_snapshot;
+    let mut genmc_config = None;
 
     let mut rustc_args = vec![];
     let mut after_dashdash = false;
@@ -603,6 +624,10 @@ fn main() {
             many_seeds = Some(0..64);
         } else if arg == "-Zmiri-many-seeds-keep-going" {
             many_seeds_keep_going = true;
+        } else if let Some(trimmed_arg) = arg.strip_prefix("-Zmiri-genmc") {
+            // FIXME(GenMC): Currently, GenMC mode is incompatible with aliasing model checking.
+            miri_config.borrow_tracker = None;
+            GenmcConfig::parse_arg(&mut genmc_config, trimmed_arg);
         } else if let Some(param) = arg.strip_prefix("-Zmiri-env-forward=") {
             miri_config.forwarded_env_vars.push(param.to_owned());
         } else if let Some(param) = arg.strip_prefix("-Zmiri-env-set=") {
@@ -697,7 +722,7 @@ fn main() {
             rustc_args.push(arg);
         }
     }
-    // `-Zmiri-unique-is-unique` should only be used with `-Zmiri-tree-borrows`
+    // `-Zmiri-unique-is-unique` should only be used with `-Zmiri-tree-borrows`.
     if miri_config.unique_is_unique
         && !matches!(miri_config.borrow_tracker, Some(BorrowTrackerMethod::TreeBorrows))
     {
@@ -734,7 +759,24 @@ fn main() {
     let many_seeds =
         many_seeds.map(|seeds| ManySeedsConfig { seeds, keep_going: many_seeds_keep_going });
 
+    // Validate settings for data race detection and GenMC mode.
+    assert_eq!(genmc_config.is_some(), miri_config.genmc_mode);
+    if genmc_config.is_some() {
+        if !miri_config.data_race_detector {
+            show_error!("Cannot disable data race detection in GenMC mode (currently)");
+        } else if !miri_config.weak_memory_emulation {
+            show_error!("Cannot disable weak memory emulation in GenMC mode");
+        }
+    } else if miri_config.weak_memory_emulation && !miri_config.data_race_detector {
+        show_error!(
+            "Weak memory emulation cannot be enabled when the data race detector is disabled"
+        );
+    };
+
     debug!("rustc arguments: {:?}", rustc_args);
     debug!("crate arguments: {:?}", miri_config.args);
-    run_compiler_and_exit(&rustc_args, &mut MiriCompilerCalls::new(miri_config, many_seeds))
+    run_compiler_and_exit(
+        &rustc_args,
+        &mut MiriCompilerCalls::new(miri_config, many_seeds, genmc_config),
+    )
 }
--- a/src/concurrency/data_race.rs
+++ b/src/concurrency/data_race.rs
@@ -1,4 +1,4 @@
-//! Implementation of a data-race detector using Lamport Timestamps / Vector-clocks
+//! Implementation of a data-race detector using Lamport Timestamps / Vector clocks
 //! based on the Dynamic Race Detection for C++:
 //! <https://www.doc.ic.ac.uk/~afd/homepages/papers/pdfs/2017/POPL.pdf>
 //! which does not report false-positives when fences are used, and gives better
@@ -54,6 +54,7 @@ use rustc_span::Span;
 
 use super::vector_clock::{VClock, VTimestamp, VectorIdx};
 use super::weak_memory::EvalContextExt as _;
+use crate::concurrency::GlobalDataRaceHandler;
 use crate::diagnostics::RacingOp;
 use crate::*;
 
@@ -259,7 +260,7 @@ enum AccessType {
 /// Per-byte vector clock metadata for data-race detection.
 #[derive(Clone, PartialEq, Eq, Debug)]
 struct MemoryCellClocks {
-    /// The vector-clock timestamp and the thread that did the last non-atomic write. We don't need
+    /// The vector clock timestamp and the thread that did the last non-atomic write. We don't need
     /// a full `VClock` here, it's always a single thread and nothing synchronizes, so the effective
     /// clock is all-0 except for the thread that did the write.
     write: (VectorIdx, VTimestamp),
@@ -269,7 +270,7 @@ struct MemoryCellClocks {
     /// a deallocation of memory.
     write_type: NaWriteType,
 
-    /// The vector-clock of all non-atomic reads that happened since the last non-atomic write
+    /// The vector clock of all non-atomic reads that happened since the last non-atomic write
     /// (i.e., we join together the "singleton" clocks corresponding to each read). It is reset to
     /// zero on each write operation.
     read: VClock,
@@ -298,7 +299,7 @@ struct ThreadExtraState {
 }
 
 /// Global data-race detection state, contains the currently
-/// executing thread as well as the vector-clocks associated
+/// executing thread as well as the vector clocks associated
 /// with each of the threads.
 // FIXME: it is probably better to have one large RefCell, than to have so many small ones.
 #[derive(Debug, Clone)]
@@ -335,7 +336,7 @@ pub struct GlobalState {
     /// for use as the index for a new thread.
     /// Elements in this set may still require the vector index to
     /// report data-races, and can only be re-used after all
-    /// active vector-clocks catch up with the threads timestamp.
+    /// active vector clocks catch up with the threads timestamp.
     reuse_candidates: RefCell<FxHashSet<VectorIdx>>,
 
     /// We make SC fences act like RMWs on a global location.
@@ -348,6 +349,9 @@ pub struct GlobalState {
 
     /// Track when an outdated (weak memory) load happens.
     pub track_outdated_loads: bool,
+
+    /// Whether weak memory emulation is enabled
+    pub weak_memory: bool,
 }
 
 impl VisitProvenance for GlobalState {
@@ -680,6 +684,23 @@ impl MemoryCellClocks {
     }
 }
 
+impl GlobalDataRaceHandler {
+    /// Select whether data race checking is disabled. This is solely an
+    /// implementation detail of `allow_data_races_*` and must not be used anywhere else!
+    fn set_ongoing_action_data_race_free(&self, enable: bool) {
+        match self {
+            GlobalDataRaceHandler::None => {}
+            GlobalDataRaceHandler::Vclocks(data_race) => {
+                let old = data_race.ongoing_action_data_race_free.replace(enable);
+                assert_ne!(old, enable, "cannot nest allow_data_races");
+            }
+            GlobalDataRaceHandler::Genmc(genmc_ctx) => {
+                genmc_ctx.set_ongoing_action_data_race_free(enable);
+            }
+        }
+    }
+}
+
 /// Evaluation context extensions.
 impl<'tcx> EvalContextExt<'tcx> for MiriInterpCx<'tcx> {}
 pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
@@ -696,6 +717,19 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
         // This is fine with StackedBorrow and race checks because they don't concern metadata on
         // the *value* (including the associated provenance if this is an AtomicPtr) at this location.
         // Only metadata on the location itself is used.
+
+        if let Some(genmc_ctx) = this.machine.data_race.as_genmc_ref() {
+            // FIXME(GenMC): Inform GenMC what a non-atomic read here would return, to support mixed atomics/non-atomics
+            let old_val = None;
+            return genmc_ctx.atomic_load(
+                this,
+                place.ptr().addr(),
+                place.layout.size,
+                atomic,
+                old_val,
+            );
+        }
+
         let scalar = this.allow_data_races_ref(move |this| this.read_scalar(place))?;
         let buffered_scalar = this.buffered_atomic_read(place, atomic, scalar, || {
             this.validate_atomic_load(place, atomic)
@@ -718,6 +752,12 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
         // This is also a very special exception where we just ignore an error -- if this read
         // was UB e.g. because the memory is uninitialized, we don't want to know!
         let old_val = this.run_for_validation_mut(|this| this.read_scalar(dest)).discard_err();
+        // Inform GenMC about the atomic store.
+        if let Some(genmc_ctx) = this.machine.data_race.as_genmc_ref() {
+            // FIXME(GenMC): Inform GenMC what a non-atomic read here would return, to support mixed atomics/non-atomics
+            genmc_ctx.atomic_store(this, dest.ptr().addr(), dest.layout.size, val, atomic)?;
+            return interp_ok(());
+        }
         this.allow_data_races_mut(move |this| this.write_scalar(val, dest))?;
         this.validate_atomic_store(dest, atomic)?;
         this.buffered_atomic_write(val, dest, atomic, old_val)
@@ -736,6 +776,21 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
         this.atomic_access_check(place, AtomicAccessType::Rmw)?;
 
         let old = this.allow_data_races_mut(|this| this.read_immediate(place))?;
+
+        // Inform GenMC about the atomic rmw operation.
+        if let Some(genmc_ctx) = this.machine.data_race.as_genmc_ref() {
+            // FIXME(GenMC): Inform GenMC what a non-atomic read here would return, to support mixed atomics/non-atomics
+            let (old_val, new_val) = genmc_ctx.atomic_rmw_op(
+                this,
+                place.ptr().addr(),
+                place.layout.size,
+                atomic,
+                (op, not),
+                rhs.to_scalar(),
+            )?;
+            this.allow_data_races_mut(|this| this.write_scalar(new_val, place))?;
+            return interp_ok(ImmTy::from_scalar(old_val, old.layout));
+        }
 
         let val = this.binary_op(op, &old, rhs)?;
         let val = if not { this.unary_op(mir::UnOp::Not, &val)? } else { val };
@@ -761,6 +816,19 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
         let old = this.allow_data_races_mut(|this| this.read_scalar(place))?;
         this.allow_data_races_mut(|this| this.write_scalar(new, place))?;
 
+        // Inform GenMC about the atomic atomic exchange.
+        if let Some(genmc_ctx) = this.machine.data_race.as_genmc_ref() {
+            // FIXME(GenMC): Inform GenMC what a non-atomic read here would return, to support mixed atomics/non-atomics
+            let (old_val, _is_success) = genmc_ctx.atomic_exchange(
+                this,
+                place.ptr().addr(),
+                place.layout.size,
+                new,
+                atomic,
+            )?;
+            return interp_ok(old_val);
+        }
+
         this.validate_atomic_rmw(place, atomic)?;
 
         this.buffered_atomic_rmw(new, place, atomic, old)?;
@@ -780,6 +848,23 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
         this.atomic_access_check(place, AtomicAccessType::Rmw)?;
 
         let old = this.allow_data_races_mut(|this| this.read_immediate(place))?;
+
+        // Inform GenMC about the atomic min/max operation.
+        if let Some(genmc_ctx) = this.machine.data_race.as_genmc_ref() {
+            // FIXME(GenMC): Inform GenMC what a non-atomic read here would return, to support mixed atomics/non-atomics
+            let (old_val, new_val) = genmc_ctx.atomic_min_max_op(
+                this,
+                place.ptr().addr(),
+                place.layout.size,
+                atomic,
+                min,
+                old.layout.backend_repr.is_signed(),
+                rhs.to_scalar(),
+            )?;
+            this.allow_data_races_mut(|this| this.write_scalar(new_val, place))?;
+            return interp_ok(ImmTy::from_scalar(old_val, old.layout));
+        }
+
         let lt = this.binary_op(mir::BinOp::Lt, &old, &rhs)?.to_scalar().to_bool()?;
 
         #[rustfmt::skip] // rustfmt makes this unreadable
@@ -823,6 +908,25 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
         // read ordering and write in the success case.
         // Read as immediate for the sake of `binary_op()`
         let old = this.allow_data_races_mut(|this| this.read_immediate(place))?;
+
+        // Inform GenMC about the atomic atomic compare exchange.
+        if let Some(genmc_ctx) = this.machine.data_race.as_genmc_ref() {
+            let (old, cmpxchg_success) = genmc_ctx.atomic_compare_exchange(
+                this,
+                place.ptr().addr(),
+                place.layout.size,
+                this.read_scalar(expect_old)?,
+                new,
+                success,
+                fail,
+                can_fail_spuriously,
+            )?;
+            if cmpxchg_success {
+                this.allow_data_races_mut(|this| this.write_scalar(new, place))?;
+            }
+            return interp_ok(Immediate::ScalarPair(old, Scalar::from_bool(cmpxchg_success)));
+        }
+
         // `binary_op` will bail if either of them is not a scalar.
         let eq = this.binary_op(mir::BinOp::Eq, &old, expect_old)?;
         // If the operation would succeed, but is "weak", fail some portion
@@ -859,49 +963,11 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
     /// Update the data-race detector for an atomic fence on the current thread.
     fn atomic_fence(&mut self, atomic: AtomicFenceOrd) -> InterpResult<'tcx> {
         let this = self.eval_context_mut();
-        let current_span = this.machine.current_span();
-        if let Some(data_race) = &mut this.machine.data_race {
-            data_race.maybe_perform_sync_operation(
-                &this.machine.threads,
-                current_span,
-                |index, mut clocks| {
-                    trace!("Atomic fence on {:?} with ordering {:?}", index, atomic);
-
-                    // Apply data-race detection for the current fences
-                    // this treats AcqRel and SeqCst as the same as an acquire
-                    // and release fence applied in the same timestamp.
-                    if atomic != AtomicFenceOrd::Release {
-                        // Either Acquire | AcqRel | SeqCst
-                        clocks.apply_acquire_fence();
-                    }
-                    if atomic == AtomicFenceOrd::SeqCst {
-                        // Behave like an RMW on the global fence location. This takes full care of
-                        // all the SC fence requirements, including C++17 §32.4 [atomics.order]
-                        // paragraph 6 (which would limit what future reads can see). It also rules
-                        // out many legal behaviors, but we don't currently have a model that would
-                        // be more precise.
-                        // Also see the second bullet on page 10 of
-                        // <https://www.cs.tau.ac.il/~orilahav/papers/popl21_robustness.pdf>.
-                        let mut sc_fence_clock = data_race.last_sc_fence.borrow_mut();
-                        sc_fence_clock.join(&clocks.clock);
-                        clocks.clock.join(&sc_fence_clock);
-                        // Also establish some sort of order with the last SC write that happened, globally
-                        // (but this is only respected by future reads).
-                        clocks.write_seqcst.join(&data_race.last_sc_write_per_thread.borrow());
-                    }
-                    // The release fence is last, since both of the above could alter our clock,
-                    // which should be part of what is being released.
-                    if atomic != AtomicFenceOrd::Acquire {
-                        // Either Release | AcqRel | SeqCst
-                        clocks.apply_release_fence();
-                    }
-
-                    // Increment timestamp in case of release semantics.
-                    interp_ok(atomic != AtomicFenceOrd::Acquire)
-                },
-            )
-        } else {
-            interp_ok(())
+        let machine = &this.machine;
+        match &this.machine.data_race {
+            GlobalDataRaceHandler::None => interp_ok(()),
+            GlobalDataRaceHandler::Vclocks(data_race) => data_race.atomic_fence(machine, atomic),
+            GlobalDataRaceHandler::Genmc(genmc_ctx) => genmc_ctx.atomic_fence(machine, atomic),
         }
     }
 
@@ -910,10 +976,7 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
     fn allow_data_races_all_threads_done(&mut self) {
         let this = self.eval_context_ref();
         assert!(this.have_all_terminated());
-        if let Some(data_race) = &this.machine.data_race {
-            let old = data_race.ongoing_action_data_race_free.replace(true);
-            assert!(!old, "cannot nest allow_data_races");
-        }
+        this.machine.data_race.set_ongoing_action_data_race_free(true);
     }
 
     /// Calls the callback with the "release" clock of the current thread.
@@ -923,14 +986,16 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
     /// The closure will only be invoked if data race handling is on.
     fn release_clock<R>(&self, callback: impl FnOnce(&VClock) -> R) -> Option<R> {
         let this = self.eval_context_ref();
-        Some(this.machine.data_race.as_ref()?.release_clock(&this.machine.threads, callback))
+        Some(
+            this.machine.data_race.as_vclocks_ref()?.release_clock(&this.machine.threads, callback),
+        )
     }
 
     /// Acquire the given clock into the current thread, establishing synchronization with
     /// the moment when that clock snapshot was taken via `release_clock`.
     fn acquire_clock(&self, clock: &VClock) {
         let this = self.eval_context_ref();
-        if let Some(data_race) = &this.machine.data_race {
+        if let Some(data_race) = this.machine.data_race.as_vclocks_ref() {
             data_race.acquire_clock(clock, &this.machine.threads);
         }
     }
@@ -1132,7 +1197,7 @@ impl VClockAlloc {
         machine: &MiriMachine<'_>,
     ) -> InterpResult<'tcx> {
         let current_span = machine.current_span();
-        let global = machine.data_race.as_ref().unwrap();
+        let global = machine.data_race.as_vclocks_ref().unwrap();
         if !global.race_detecting() {
             return interp_ok(());
         }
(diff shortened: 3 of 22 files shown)


src/eval.rs
```diff
@@ -230,6 +235,14 @@ impl<'tcx> MainThreadState<'tcx> {
                 match state.on_stack_empty(this)? {
                     Poll::Pending => {} // just keep going
                     Poll::Ready(()) => {
+                        if let Some(genmc_ctx) = this.machine.concurrency_handler.as_genmc_ref() {
+                            // In GenMC mode, we don't yield, instead we inform the GenmcCtx that the main thread is done, which will then handle it.
+                            genmc_ctx.handle_thread_stack_empty(this.active_thread());
```
RalfJung, line 240: We seem to only call this when the *main* thread stack is empty? If that's correct, the function name should reflect it.

src/eval.rs
```diff
@@ -477,6 +510,14 @@ pub fn eval_entry<'tcx>(
         EnvVars::cleanup(&mut ecx).expect("error during env var cleanup");
     }
 
+    // FIXME: is this the correct place to put this? Should this go into `on_main_stack_empty`?
```
RalfJung, line 513: What kinds of errors are reported this way?

src/machine.rs
```diff
@@ -1371,22 +1387,29 @@ impl<'tcx> Machine<'tcx> for MiriMachine<'tcx> {
...
         (alloc_id, prov_extra): (AllocId, Self::ProvenanceExtra),
         range: AllocRange,
     ) -> InterpResult<'tcx> {
         if machine.track_alloc_accesses && machine.tracked_alloc_ids.contains(&alloc_id) {
             machine
                 .emit_diagnostic(NonHaltingDiagnostic::AccessedAlloc(alloc_id, AccessKind::Read));
         }
-        if let Some(data_race) = &alloc_extra.data_race {
-            data_race.read(alloc_id, range, NaReadType::Read, None, machine)?;
-        }
```
RalfJung, line 1384: I deliberately placed the data race check before the borrow tracker check, so that if an access violates both, the data race is reported. So please restore the old ordering.

src/machine.rs
```diff
@@ -1371,22 +1387,29 @@ impl<'tcx> Machine<'tcx> for MiriMachine<'tcx> {
...
         if let Some(borrow_tracker) = &alloc_extra.borrow_tracker {
             borrow_tracker.before_memory_read(alloc_id, prov_extra, range, machine)?;
         }
+        if let Some(genmc_ctx) = machine.concurrency_handler.as_genmc_ref() {
+            // We throw away the value we get from GenMC for non-atomic accesses, these are still handled by Miri
+            let _value = genmc_ctx.memory_load(machine, ptr.addr(), range.size)?;
+            return interp_ok(());
+        }
+
```
RalfJung, line 1406: Why is there an empty line here? Empty lines should indicate some sort of semantic grouping within the function.

src/machine.rs
```diff
@@ -1430,7 +1459,10 @@ impl<'tcx> Machine<'tcx> for MiriMachine<'tcx> {
         if machine.tracked_alloc_ids.contains(&alloc_id) {
             machine.emit_diagnostic(NonHaltingDiagnostic::FreedAlloc(alloc_id));
         }
-        if let Some(data_race) = &mut alloc_extra.data_race {
+
+        if let Some(genmc_ctx) = machine.concurrency_handler.as_genmc_ref() {
+            genmc_ctx.handle_dealloc(machine, ptr.addr(), size, align, kind)?;
+        } else if let Some(data_race) = &mut alloc_extra.data_race {
```
RalfJung, line 1465: Please make all three `before_memory_*` functions consistent in how they access these fields. Right now, some use `else` and some do not.

It's a shame that we can't just do a nice enum match here... or maybe we can? We could have an enum for the per-allocation state that is like
```rust
enum AllocDataRaceHandler {
  None,
  GenMc,
  DataRace(data_race::AllocState, Option<weak_memory::AllocState>),
}
```
with the invariant that it must match the enum variant in `MiriMachine`.

src/concurrency/concurrency_handler.rs
```diff
@@ -0,0 +1,51 @@
+use std::rc::Rc;
+
+use super::data_race;
+use crate::concurrency::GenmcCtx;
+use crate::{VisitProvenance, VisitWith};
+
+pub enum ConcurrencyHandler {
```
RalfJung, line 7: This doesn't handle concurrency, it just handles data race detection and scheduling. So I think `DataRaceHandler` or so might be a better name?

src/concurrency/concurrency_handler.rs
```diff
@@ -0,0 +1,51 @@
...
+    None,
+    DataRace(Box<data_race::GlobalState>),
+    GenMC(Rc<GenmcCtx>),
```
RalfJung, line 10: Following the usual Rust naming convention, this should be `GenMc`. Or maybe `Genmc`, that's what you picked for `GenmcCtx`. Either way, the variant name and the `*Ctx` type name should be consistent.

src/diagnostics.rs
```diff
@@ -31,6 +31,8 @@ pub enum TerminationInfo {
     },
     Int2PtrWithStrictProvenance,
     Deadlock,
+    /// In GenMC mode, an execution can get stuck in certain cases. This is not always an error.
```
RalfJung, line 34: "not always"? The code currently works as if it is never an error.

src/concurrency/genmc/dummy.rs
```diff
@@ -0,0 +1,233 @@
...
+
+#[derive(Debug)]
+pub struct GenmcCtx {}
+
+#[derive(Debug, Default, Clone)]
+pub struct GenmcConfig {}
+
+impl GenmcCtx {
+    pub fn try_new(_miri_config: &MiriConfig) -> Option<Self> {
+        unimplemented!("GenMC feature in Miri is currently disabled.");
```
RalfJung, line 20: Shouldn't this be `unreachable!` as well?

src/concurrency/genmc/config.rs
```diff
@@ -0,0 +1,29 @@
...
+    pub fn parse_arg(miri_config: &mut MiriConfig, trimmed_arg: &str) {
+        if miri_config.genmc_config.is_none() {
+            miri_config.genmc_config = Some(Default::default());
+
+            // FIXME: make sure none of the following settings are reactivated by other passed arguments:
+
+            // GenMC handles data race detection and weak memory emulation, so we disable the Miri equivalents:
+            miri_config.data_race_detector = false;
+            miri_config.weak_memory_emulation = false;
+            // FIXME: Currently GenMC mode incompatible with aliasing model checking
```
RalfJung, line 22:
```suggestion
            // FIXME: Currently GenMC mode is incompatible with aliasing model checking
```

src/concurrency/genmc/config.rs
```diff
@@ -0,0 +1,29 @@
...
+impl GenmcConfig {
+    /// Function for parsing command line options for GenMC mode.
+    /// All GenMC arguments start with the string "-Zmiri-genmc".
+    ///
+    /// `trimmed_arg` should be the argument to be parsed, with the suffix "-Zmiri-genmc" removed
+    pub fn parse_arg(miri_config: &mut MiriConfig, trimmed_arg: &str) {
+        if miri_config.genmc_config.is_none() {
+            miri_config.genmc_config = Some(Default::default());
+
+            // FIXME: make sure none of the following settings are reactivated by other passed arguments:
```
RalfJung, line 17: You could add a `GenmcConfig::finalize` method to deal with that.

RalfJung (commented):
I haven't looked at everything yet, here's a first round of feedback.

src/eval.rs
Patrick-6: GenMC can return a "Linearizability" or a "Liveness" error from this function.
It can also cause an "UnfreedMemory" warning.

src/machine.rs
```diff
@@ -636,7 +644,12 @@ impl<'tcx> MiriMachine<'tcx> {
         });
         let rng = StdRng::seed_from_u64(config.seed.unwrap_or(0));
         let borrow_tracker = config.borrow_tracker.map(|bt| bt.instantiate_global_state(config));
-        let data_race = config.data_race_detector.then(|| data_race::GlobalState::new(config));
+        let concurrency_handler = match (config.data_race_detector, genmc_ctx) {
+            (false, None) => ConcurrencyHandler::None,
+            (true, None) =>
+                ConcurrencyHandler::DataRace(Box::new(data_race::GlobalState::new(config))),
+            (_, Some(genmc_ctx)) => ConcurrencyHandler::GenMC(genmc_ctx),
```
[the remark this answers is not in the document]
Patrick-6: Currently yes, but GenMC has an option to disable race detection. We could allow the `-Zmiri-disable-data-race-detector` flag to also disable race detection in GenMC mode, in which case `(false, Some(_))` would be possible.

src/concurrency/genmc/config.rs
Patrick-6: Currently I do these checks in `GenmcCtx::try_new(miri_config: &MiriConfig)`, I think it makes sense to check the settings when creating the `GenmcCtx`.

src/concurrency/genmc/dummy.rs
Patrick-6: This is the one function that could still be called, so I added a nicer error message, but I don't mind either way.

src/machine.rs
Patrick-6: So this means it is invalid to have `weak_memory_emulation == true` and `data_race_detector == false`?

Should I also combine these 2 settings and the `genmc_config` in `MiriConfig` to be an enum instead, so we can prevent even setting an invalid configuration?
```rust
enum DataRaceConfig {
    None,
    DataRace { weak_memory_emulation: bool },
    Genmc(GenmcConfig),
}
```

src/concurrency/concurrency_handler.rs
```diff
@@ -0,0 +1,51 @@
...
+
+use super::data_race;
+use crate::concurrency::GenmcCtx;
+use crate::{VisitProvenance, VisitWith};
+
+pub enum DataRaceHandler {
+    None,
+    DataRace(Box<data_race::GlobalState>),
+    Genmc(Rc<GenmcCtx>),
+}
```
Patrick-6, line 11: The `data_race::GlobalState` is now boxed, an it wasn't before. The new enum triggered the [large_enum_variant](https://rust-lang.github.io/rust-clippy/master/index.html?versions=lte%3A80#large_enum_variant) clippy lint since the `GlobalState` is quite large (288 bytes). With the box the enum is just 16 bytes large (vs 288 bytes).

The `GenmcCtx` is behind an `Rc` since that made it easier to persist across executions.

Is this ok?

src/eval.rs
Patrick-6: It is also called [here](https://github.com/rust-lang/miri/blob/8f2362180d99c0b87e1598ac1a153bded4b7b3a7/src/concurrency/thread.rs#L846-L850), but this may not be needed.

RalfJung: If you also call it there then I think it will be called twice for the main thread. So just remove the call in eval.rs then.

src/machine.rs
RalfJung: > So this means it is invalid to have weak_memory_emulation == true and data_race_detector == false?

Yes.

> Should I also combine these 2 settings and the genmc_config in MiriConfig to be an enum instead, so we can prevent even setting an invalid configuration?

`MiriConfig` reflects the user-provided `-Z` flags directly, so this might be annoying and introduce some order-dependence to the flags... but you can give it a try.

src/concurrency/thread.rs
```diff
@@ -580,13 +580,28 @@ impl<'tcx> ThreadManager<'tcx> {
...
         joined_thread_id: ThreadId,
-        data_race: Option<&mut data_race::GlobalState>,
+        data_race_handler: &mut DataRaceHandler,
     ) -> InterpResult<'tcx> {
         if self.threads[joined_thread_id].join_status == ThreadJoinStatus::Detached {
             // On Windows this corresponds to joining on a closed handle.
             throw_ub_format!("trying to join a detached thread");
         }
 
+        fn handle_join<'tcx>(
```
RalfJung, line 590: Maybe call this `after_join` to make its purpose more clear?

src/concurrency/thread.rs
```diff
@@ -702,7 +712,32 @@ impl<'tcx> ThreadManager<'tcx> {
...
+            thread_manager.yield_active_thread = false;
+
+            assert!(thread_manager.threads[thread_manager.active_thread].state.is_enabled());
+            return interp_ok(SchedulingAction::ExecuteStep);
+        }
+
+        ecx.machine.threads.schedule_impl(clock)
+    }
+
+    fn schedule_impl(&mut self, clock: &MonotonicClock) -> InterpResult<'tcx, SchedulingAction> {
```
RalfJung, line 740: This function seems to be called only here -- so please don't make it a separate function.

src/concurrency/thread.rs
```diff
@@ -726,6 +761,7 @@ impl<'tcx> ThreadManager<'tcx> {
         // `skip(N)` means we start iterating at thread N, so we skip 1 more to start just *after*
         // the active thread. Then after that we look at `take(N)`, i.e., the threads *before* the
         // active thread.
+
```
RalfJung, line 764: This separates the comment from the code it refers to. Why?

src/concurrency/thread.rs
```diff
@@ -702,7 +712,32 @@ impl<'tcx> ThreadManager<'tcx> {
     /// used in stateless model checkers such as Loom: run the active thread as
     /// long as we can and switch only when we have to (the active thread was
     /// blocked, terminated, or has explicitly asked to be preempted).
-    fn schedule(&mut self, clock: &MonotonicClock) -> InterpResult<'tcx, SchedulingAction> {
+    fn schedule(
+        ecx: &mut InterpCx<'tcx, MiriMachine<'tcx>>,
```
RalfJung, line 716: Since this now becomes a method on `InterpCx`, please move it to the extension trait (`pub trait EvalContextExt` below).

src/concurrency/weak_memory.rs
```diff
@@ -459,8 +461,11 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let this = self.eval_context_mut();
         let (alloc_id, base_offset, ..) = this.ptr_get_alloc_id(place.ptr(), 0)?;
         if let (
-            crate::AllocExtra { weak_memory: Some(alloc_buffers), .. },
-            crate::MiriMachine { data_race: Some(global), threads, .. },
+            crate::AllocExtra {
+                data_race: AllocDataRaceHandler::DataRace(_data_race, Some(alloc_buffers)),
```
RalfJung, line 465:
```suggestion
                data_race: AllocDataRaceHandler::DataRace(_, Some(alloc_buffers)),
```

src/concurrency/weak_memory.rs
```diff
@@ -534,8 +541,11 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let this = self.eval_context_mut();
         let (alloc_id, base_offset, ..) = this.ptr_get_alloc_id(dest.ptr(), 0)?;
         if let (
-            crate::AllocExtra { weak_memory: Some(alloc_buffers), .. },
-            crate::MiriMachine { data_race: Some(global), threads, .. },
+            crate::AllocExtra {
+                data_race: AllocDataRaceHandler::DataRace(_data_race, Some(alloc_buffers)),
```
RalfJung, line 545:
```suggestion
                data_race: AllocDataRaceHandler::DataRace(_, Some(alloc_buffers)),
```

src/intrinsics/atomic.rs
```diff
@@ -199,21 +202,32 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
             span_bug!(this.cur_span(), "atomic arithmetic operation type mismatch");
         }
 
+        // Inform GenMC about the atomic rmw operation.
+        if let Some(genmc_ctx) = this.machine.data_race.as_genmc_ref() {
+            let is_unsigned = matches!(rhs.layout.ty.kind(), ty::Uint(_));
+            let (old, _is_success) = genmc_ctx.atomic_rmw_op(
+                this,
```
RalfJung, line 209: Why is this hooked here and not inside `atomic_min_max_scalar`/`atomic_rmw_op_immediate`?

src/intrinsics/atomic.rs
```diff
@@ -230,9 +244,20 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
         let place = this.deref_pointer(place)?;
         let new = this.read_scalar(new)?;
 
+        // Inform GenMC about the atomic atomic exchange.
+        if let Some(genmc_ctx) = this.machine.data_race.as_genmc_ref() {
+            let (old, _is_success) = genmc_ctx.atomic_exchange(
```
RalfJung, line 249: Why is this hooked here and not inside `atomic_exchange_scalar`?

src/intrinsics/atomic.rs
```diff
@@ -247,9 +272,29 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
 
         let [place, expect_old, new] = check_intrinsic_arg_count(args)?;
         let place = this.deref_pointer(place)?;
-        let expect_old = this.read_immediate(expect_old)?; // read as immediate for the sake of `binary_op()`
         let new = this.read_scalar(new)?;
 
+        // Inform GenMC about the atomic atomic compare exchange.
+        if let Some(genmc_ctx) = this.machine.data_race.as_genmc_ref() {
+            let expect_old_scalar = this.read_scalar(expect_old)?;
+            let (old, cmpxchg_success) = genmc_ctx.atomic_compare_exchange(
```
RalfJung, line 280: Same question, seems like this should be hooked inside `atomic_compare_exchange_scalar`.

Other parts of Miri might call `atomic_compare_exchange_scalar` directly (e.g. [here](https://github.com/rust-lang/miri/blob/3954d5c096758fd2699c6941e07c41a5f31836bf/src/concurrency/sync.rs#L280)), and I assume you want to hook *all* atomic operations.

src/intrinsics/mod.rs
```diff
@@ -10,6 +10,7 @@ use rustc_middle::mir;
 use rustc_middle::ty::{self, FloatTy, ScalarInt};
 use rustc_span::{Symbol, sym};
 
+pub use self::atomic::AtomicOp;
```
RalfJung, line 13: Please don't put a public re-export in the middle of a whole bunch of private imports. Those are very different uses of `use`!

Why is this re-exported now anyway? And why does that happen here and not in `lib.rs`?

src/concurrency/data_race_handler.rs
```diff
@@ -0,0 +1,100 @@
+use std::rc::Rc;
+
+use super::{GenmcConfig, data_race, weak_memory};
+use crate::concurrency::GenmcCtx;
+use crate::{VisitProvenance, VisitWith};
+
+// FIXME: naming
+
+pub enum DataRaceHandler {
```
RalfJung, line 9: This should probably be `GlobalDataRaceHandler`.

src/machine.rs
```diff
@@ -1371,23 +1383,30 @@ impl<'tcx> Machine<'tcx> for MiriMachine<'tcx> {
...
+        ptr: Pointer,
         (alloc_id, prov_extra): (AllocId, Self::ProvenanceExtra),
         range: AllocRange,
     ) -> InterpResult<'tcx> {
         if machine.track_alloc_accesses && machine.tracked_alloc_ids.contains(&alloc_id) {
             machine
                 .emit_diagnostic(NonHaltingDiagnostic::AccessedAlloc(alloc_id, AccessKind::Read));
         }
-        if let Some(data_race) = &alloc_extra.data_race {
+        // The order of checks is deliberate, to prefer reporting a data race over of a borrow tracker error
```
RalfJung, line 1394:
```suggestion
        // The order of checks is deliberate, to prefer reporting a data race over a borrow tracker error.
```

src/machine.rs
```diff
@@ -1396,23 +1415,37 @@ impl<'tcx> Machine<'tcx> for MiriMachine<'tcx> {
...
-        if let Some(data_race) = &mut alloc_extra.data_race {
-            data_race.write(alloc_id, range, NaWriteType::Write, None, machine)?;
-        }
         if let Some(borrow_tracker) = &mut alloc_extra.borrow_tracker {
             borrow_tracker.before_memory_write(alloc_id, prov_extra, range, machine)?;
         }
-        if let Some(weak_memory) = &alloc_extra.weak_memory {
-            weak_memory.memory_accessed(range, machine.data_race.as_ref().unwrap());
+
+        match &machine.data_race {
```
RalfJung, line 1430: Here you are still going the borrow tracker check first.

RalfJung (commented):
Okay now I looked at everything :)

src/bin/miri.rs
```diff
@@ -536,8 +547,11 @@ fn main() {
         } else if arg == "-Zmiri-unique-is-unique" {
             miri_config.unique_is_unique = true;
         } else if arg == "-Zmiri-disable-data-race-detector" {
-            miri_config.data_race_detector = false;
-            miri_config.weak_memory_emulation = false;
+            if let DataRaceConfig::Genmc(_) = &miri_config.data_race_detection {
+                // FIXME?: Should we allow this?
+                show_error!("Cannot turn off data race detection in GenMC mode");
+            }
+            miri_config.data_race_detection = DataRaceConfig::None;
```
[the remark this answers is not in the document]
Patrick-6: I added the opposite check in `GenmcConfig::parse_arg`, but yes, it's probably better to keep to old version with separate settings and check for consistency at the end.

src/concurrency/genmc/config.rs
```diff
@@ -11,19 +11,26 @@ impl GenmcConfig {
...
 
-            // GenMC handles data race detection and weak memory emulation, so we disable the Miri equivalents:
-            miri_config.data_race_detector = false;
-            miri_config.weak_memory_emulation = false;
-            // FIXME: Currently GenMC mode is incompatible with aliasing model checking
-            miri_config.borrow_tracker = None;
-
-            todo!("GenMC mode not yet supported")
+                miri_config.data_race_detection = DataRaceConfig::Genmc(Default::default());
+                todo!("GenMC mode not yet supported")
```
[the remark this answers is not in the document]
Patrick-6: Currently I have 3 flags, but there are ~10 - 30 GenMC options we may want to expose to the user (Most to control the exploration and which checks are done, some for code transformations like replacing spin loops with `assume`, some for debugging). For some of these we can probably reuse existing Miri flags (like for `disableRaceDetection`), but I don't know yet how many new ones we will need. That's also why I added `parse_arg`, this keeps all the GenMC option parsing together and out of `miri.rs`.

src/concurrency/thread.rs
Patrick-6: Should it be public or in a new `trait EvalContextExtPriv`? I don't think this is used anywhere else.
(If we make it public we should also make the returned `enum SchedulingAction` public.)

src/concurrency/thread.rs
Patrick-6: I left it a method on `ThreadManager` to reduce the diff, but sure, I can remove it.

src/concurrency/genmc/config.rs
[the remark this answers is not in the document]
RalfJung: Okay, I see.

Still this function here is quite confusing. Since this PR is mostly about adding the dummy, I'd suggest to keep the other file entirely a stub.

src/concurrency/thread.rs
RalfJung: Ah, there's no `Priv` trait in this file yet?

You can also keep it here then, but please inline `schedule_impl` at its own call site.

src/concurrency/data_race_handler.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    pub fn as_data_race_mut(&mut self) -> Option<&mut data_race::GlobalState> {
+        if let Self::DataRace(data_race) = self { Some(data_race) } else { None }
+    }
+
+    pub fn as_genmc_ref(&self) -> Option<&GenmcCtx> {
+        if let Self::Genmc(genmc_ctx) = self { Some(genmc_ctx) } else { None }
+    }
+
+    pub fn set_ongoing_action_data_race_free(&self, enable: bool) {
```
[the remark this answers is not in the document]
Patrick-6: It is not replacing that `allow_data_races_all_threads_done`, but that function calls `set_ongoing_action_data_race_free`. I've added a doc comment to it.

src/diagnostics.rs
```diff
@@ -31,7 +31,8 @@ pub enum TerminationInfo {
     },
     Int2PtrWithStrictProvenance,
     Deadlock,
-    /// In GenMC mode, an execution can get stuck in certain cases. This is not always an error.
+    /// In GenMC mode, an execution can get stuck in certain cases.
+    /// There may be other executions to explore, so this should *not* be reported as an error.
```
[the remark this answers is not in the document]
Patrick-6: Which "then"?

src/intrinsics/atomic.rs
Patrick-6: Yeah, it should be there, I moved it.

src/intrinsics/mod.rs
Patrick-6: Removed it, it's no longer needed

src/diagnostics.rs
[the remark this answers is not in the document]
RalfJung: Sorry, I meant the "so".

src/concurrency/data_race_handler.rs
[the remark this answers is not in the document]
RalfJung: We don't want this functionality publicly exposed like this though.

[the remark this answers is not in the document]
RalfJung: I think it shouldn't exist at all here.

src/intrinsics/atomic.rs
Patrick-6: Moved it

src/concurrency/data_race.rs
```diff
@@ -1486,15 +1469,20 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
     ) -> InterpResult<'tcx> {
         let this = self.eval_context_ref();
         assert!(access.is_atomic());
-        let Some(data_race) = &this.machine.data_race else { return interp_ok(()) };
+        let data_race = match &this.machine.data_race {
+            DataRaceHandler::None => return interp_ok(()),
+            DataRaceHandler::DataRace(data_race) => data_race,
+            DataRaceHandler::Genmc(_genmc_ctx) => return interp_ok(()),
+        };
```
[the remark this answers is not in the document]
Patrick-6: Good catch

src/concurrency/data_race.rs
```diff
@@ -717,9 +734,16 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
...
+
         let old_val = this.run_for_validation_mut(|this| this.read_scalar(dest)).discard_err();
         this.allow_data_races_mut(move |this| this.write_scalar(val, dest))?;
         this.validate_atomic_store(dest, atomic)?;
+
+        // Inform GenMC about the atomic store.
+        if let Some(genmc_ctx) = this.machine.data_race.as_genmc_ref() {
+            genmc_ctx.atomic_store(this, dest.ptr().addr(), dest.layout.size, val, atomic)?;
+        }
+
```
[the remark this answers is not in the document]
Patrick-6: It doesn't need to, I'll change it

src/concurrency/data_race.rs
```diff
@@ -1579,6 +1568,15 @@ impl GlobalState {
         global_state
     }
 
+    /// Select whether data race free actions should be allowed.
+    ///
+    /// # PANICS
+    /// If data race free is attempted to be set more than once (no nesting allowed)
+    pub fn set_ongoing_action_data_race_free(&self, enable: bool) {
```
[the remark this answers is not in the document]
Patrick-6: I've changed it to `pub(super)`. It is called in [AllocDataRaceHandler::set_ongoing_action_data_race_free](https://github.com/Patrick-6/miri/blob/9b9fe0430f1907176a268d3a5253084365f9b7ff/src/concurrency/data_race_handler.rs#L59-L69), which is there so the 5 places that call it don't all have to match on the `GlobalDataRaceHandler `.

src/eval.rs
Patrick-6: There are now two places where `handle_execution_end` is called. The function needs also be called in the `GenmcStuckExecution` case. We need to handle this case before the `report_error`, but in the non-error case we probably shouldn't call `handle_execution_end` before the `EnvVars::cleanup` is done, so we can't combine them.

We could combine the 2 calls if we can guarantee that nothing that happens in the `EnvVars::cleanup` is relevant for GenMC. For example, can there be any atomic operations during cleanup? This would currently trigger a panic due to the `allow_data_races_all_threads_done` (e.g., [here](https://github.com/Patrick-6/miri/blob/c6841ec29b929e72c39fa5f56a0d57f2f2d54a91/src/concurrency/genmc/mod.rs#L86)).

src/concurrency/thread.rs
Patrick-6: I've already created the `Priv` trait, and I've inlined the `schedule_impl`.

src/eval.rs
RalfJung: No, there cannot be atomic operations during `EnvVars::cleanup`.

I feel like `EnvVars::cleanup` should probably be moved... I'll experiment with that and likely make a PR myself.

RalfJung: I'm moving it in https://github.com/rust-lang/miri/pull/4302.

src/bin/miri.rs
```diff
@@ -749,6 +744,29 @@ fn main() {
     let many_seeds =
         many_seeds.map(|seeds| ManySeedsConfig { seeds, keep_going: many_seeds_keep_going });
 
+    // Validate settings for data race detection and GenMC mode.
+    match (
```
RalfJung, line 748: I think this would become easier to read if you did a top-level `if miri_config.genmc_config.is_some()` rahter than cramming it all into a single `match`.

I assume you cannot disable weak_memory_emulation in GenMC mode either.

src/machine.rs
```diff
@@ -639,15 +637,20 @@ impl<'tcx> MiriMachine<'tcx> {
...
-            DataRaceConfig::DataRace { weak_memory_emulation } =>
-                DataRaceHandler::DataRace(Box::new(data_race::GlobalState::new(
-                    config,
-                    *weak_memory_emulation,
-                ))),
-            DataRaceConfig::Genmc(_genmc_config) => DataRaceHandler::Genmc(genmc_ctx.unwrap()),
-        };
+        let data_race =
+            match (config.data_race_detector, config.weak_memory_emulation, &config.genmc_config) {
+                (true, weak_memory_emulation, None) =>
```
RalfJung: This is super hard to read. An `if let Some(genmc_cfg) = config.genmc_config` would be a lot easier to read I think.

src/machine.rs
```diff
@@ -639,15 +637,20 @@ impl<'tcx> MiriMachine<'tcx> {
...
+                    GlobalDataRaceHandler::Vclocks(Box::new(data_race::GlobalState::new(
+                        config,
+                        weak_memory_emulation,
+                    ))),
+                (false, false, None) => GlobalDataRaceHandler::None,
+                (false, false, Some(_)) => {
+                    // `genmc_ctx` persists across executions, so we don't create a new one here:
+                    GlobalDataRaceHandler::Genmc(genmc_ctx.unwrap())
```
RalfJung, line 647: Why is the genmc configuration ignored here?

src/concurrency/data_race.rs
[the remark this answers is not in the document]
RalfJung: > which is there so the 5 places that call it don't all have to match on the GlobalDataRaceHandler .

That shouldn't happen -- those places should all only be executed when we are using the vclocks data race checker. Where do we need this function for the other data race handlers?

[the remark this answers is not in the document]
Patrick-6: We also need to inform the `GenmcCtx` when data races are allowed, so it can filter the data race free accesses from GenMC.

The calls are [here](https://github.com/Patrick-6/miri/blob/c6841ec29b929e72c39fa5f56a0d57f2f2d54a91/src/concurrency/data_race.rs#L943) and [here](https://github.com/Patrick-6/miri/blob/c6841ec29b929e72c39fa5f56a0d57f2f2d54a91/src/concurrency/data_race.rs#L1349-L1363); adding the `match` in all 5 places is very verbose.

I could add an `impl GlobalDataRaceHandler` in `data_race.rs` and move the `set_ongoing_action_data_race_free` method there, then everything stays private to that file (except for `GenmcCtx::set_ongoing_action_data_race_free`, that needs to be at least `pub(super)`).

src/machine.rs
Patrick-6: Because of the reason mentioned in the comment, we don't want to create a new context for every execution, but to reuse the existing one.

src/concurrency/data_race.rs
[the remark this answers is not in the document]
RalfJung: > adding the match in all 5 places is very verbose.

As I said, most of these should be dead code when running with genmc. The only point where data races should possibly be disabled in genmc is for the env var cleanup. That's a single place, `allow_data_races_all_threads_done`, and having a `match` there seems fine.

src/machine.rs
RalfJung: Ah it wasn't clear that this refers to the fact that we ignore the configuration.

Why do we even pass the genmc configuration to the interpreter then? Seems like it should not be in `MiriConfig` at all.

src/concurrency/data_race.rs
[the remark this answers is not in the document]
Patrick-6: What about a place like [this](https://github.com/Patrick-6/miri/blob/c6841ec29b929e72c39fa5f56a0d57f2f2d54a91/src/concurrency/data_race.rs#L761)? `old` is currently unused for GenMC, but we may need it once we implement supporting mixed atomic/non-atomic access, by giving GenMC the value that we would read at this point, if we were to do a non-atomic read (Not sure if that's how we would get that value or if there is a better way).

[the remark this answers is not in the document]
RalfJung: That sounds like you are talking about [this](https://github.com/Patrick-6/miri/blob/c6841ec29b929e72c39fa5f56a0d57f2f2d54a91/src/concurrency/data_race.rs#L737), and it already uses a different (even bigger^^) hammer than `allow_data_races_mut`.

src/machine.rs
Patrick-6: Would it make more sense to have `Option<GenmcConfig>` as a field on [`struct MiriCompilerCalls`](https://github.com/Patrick-6/miri/blob/c6841ec29b929e72c39fa5f56a0d57f2f2d54a91/src/bin/miri.rs#L61) (with only a `bool` in `MiriConfig`)?

RalfJung: Yeah that makes sense. :)

src/concurrency/thread.rs
Patrick-6: Ah, there was already a `Priv` trait, but with a different name. Did you know that Miri is using `EvalContextExtPriv` in some files, and `EvalContextPrivExt` in others?

RalfJung: Oops, that's an accident.^^ Good catch.

src/concurrency/data_race_handler.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// Weak memory emulation via the use of store buffers (if enabled).
+    Vclocks(data_race::AllocState, Option<weak_memory::AllocState>),
+}
+
+impl GlobalDataRaceHandler {
+    pub fn is_none(&self) -> bool {
+        matches!(self, GlobalDataRaceHandler::None)
+    }
+
+    pub fn as_data_race_ref(&self) -> Option<&data_race::GlobalState> {
```
RalfJung, line 41: This should be called `as_vclocks_ref` now (and same below for the `_mut` variant)

src/concurrency/data_race.rs
```diff
@@ -1302,6 +1334,27 @@ impl FrameState {
...
 }
 
+impl GlobalDataRaceHandler {
+    /// Select whether data race free actions should be allowed. This function should be used carefully!
+    ///
+    /// (This has no effect if no data race checker is enabled.)
+    ///
+    /// # PANICS
+    /// If data race free mode is attempted to be nested (i.e., cannot call this function two times in a row with `true`, or two times in a row with `false`).
+    fn set_ongoing_action_data_race_free(&self, enable: bool) {
```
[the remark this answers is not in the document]
RalfJung: I checked out the code to take a closer look, and I think I understand now what the point of this is. Even in genmc mode, we need to do some reads/writes that are actually conceptually reads/writes of the memory (i.e., they should not use `run_for_validation_mut`), but they should be treated as atomic and therefore not run the genmc hooks.

In that case we should have this function be as private as possible, so yeah having it here sounds good. I'd rather have it further up in the file though instead of hiding it somewhere in the middle. Just before the `/// Evaluation context extensions.` sounds good.

src/eval.rs
```diff
@@ -476,11 +498,34 @@ pub fn eval_entry<'tcx>(
...
 
     // If we get here there was no fatal error.
 
+    if let Some(genmc_ctx) = ecx.machine.data_race.as_genmc_ref()
+        && let Err(error) = genmc_ctx.handle_execution_end(&ecx)
+    {
+        // FIXME(GenMC): Improve error reporting
+        tcx.dcx().err(format!("GenMC returned an error: \"{error}\""));
+        return None;
+    }
```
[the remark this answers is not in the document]
Patrick-6: I can move it up, but we need to ensure that we report errors correctly: If we get an error from the execution, and then `handle_execution_end` also gives an error, how should we report this?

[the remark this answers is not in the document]
RalfJung: Well, you tell me, I don't know which error should take precedent.^^

You can also move `report_error` up instead, I think.

src/concurrency/data_race_handler.rs
Patrick-6: Should I also rename the functions for the `AllocDataRaceHandler`?

Patrick-6: What should I do for the variable names in the places where `as_vclocks_ref` is called (or the enum variant is matched)? Currently it is different in different places, e.g., `let global = machine.data_race.as_vclocks_ref();` or `let data_race = ...`. Should all be called `data_race`? `vclocks`?

RalfJung: > Should I also rename the functions for the AllocDataRaceHandler?

Yes.

> What should I do for the variable names in the places where as_vclocks_ref is called (or the enum variant is matched)? 

Eventually we should probably clean that up, but it's not very important.

src/eval.rs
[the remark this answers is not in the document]
Patrick-6: Yes that worked, I managed to move `report_error` and remove the second `handle_execution_end`

src/machine.rs
```diff
@@ -1395,23 +1409,35 @@ impl<'tcx> Machine<'tcx> for MiriMachine<'tcx> {
...
         (alloc_id, prov_extra): (AllocId, Self::ProvenanceExtra),
         range: AllocRange,
     ) -> InterpResult<'tcx> {
         if machine.track_alloc_accesses && machine.tracked_alloc_ids.contains(&alloc_id) {
             machine
                 .emit_diagnostic(NonHaltingDiagnostic::AccessedAlloc(alloc_id, AccessKind::Write));
         }
-        if let Some(data_race) = &mut alloc_extra.data_race {
-            data_race.write(alloc_id, range, NaWriteType::Write, None, machine)?;
+        match &machine.data_race {
```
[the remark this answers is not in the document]
Patrick-6: Should be consistent now

src/bin/miri.rs
```diff
@@ -729,7 +755,24 @@ fn main() {
     let many_seeds =
         many_seeds.map(|seeds| ManySeedsConfig { seeds, keep_going: many_seeds_keep_going });
 
+    // Validate settings for data race detection and GenMC mode.
+    assert_eq!(genmc_config.is_some(), miri_config.genmc_mode);
+    if genmc_config.is_some() {
+        if miri_config.data_race_detector {
+            show_error!("GenMC mode cannot be used together with Miri's data race detector")
```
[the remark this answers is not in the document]
Patrick-6: Yes, changed it to that

src/intrinsics/atomic.rs
```diff
@@ -199,23 +199,16 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
...
-                this.write_immediate(*old, dest)?; // old value is returned
-                interp_ok(())
-            }
-            AtomicOp::MirOp(op, not) => {
-                let old = this.atomic_rmw_op_immediate(&place, &rhs, op, not, atomic)?;
-                this.write_immediate(*old, dest)?; // old value is returned
-                interp_ok(())
-            }
-        }
+        let old = match atomic_op {
```
Patrick-6, line 202: Is this ok? Looks a bit cleaner that way

src/concurrency/data_race_handler.rs
```diff
@@ -0,0 +1,99 @@
...
+    ///   an atomic operation to see at any specific point of the program.
+    Genmc(Rc<GenmcCtx>),
+    /// The default data race detector for Miri using vector-clocks.
+    Vclocks(Box<data_race::GlobalState>),
+}
+
+/// Per allocation data structures for handling data races.
+///
+/// ## Invariant
+/// The variant of the `enum GlobalDataRaceHandler` in `MiriMachine` must match the variant of the `enum AllocDataRaceHandler` stored in `AllocExtra`.
```
Patrick-6, line 32: I've added the invariants in 4 places: here, and in the `AllocExtra` and `MiriMachine` fields they are used in. Is this ok or should it only be on the fields?

.gitignore
```diff
@@ -14,4 +14,4 @@ rustc-ice*.txt
 tests/native-lib/libtestlib.so
 .auto-*
 
-./genmc/
+genmc/
```
RalfJung: This will ignore all folders called `genmc` everywhere, right? That doesn't sound right?

src/concurrency/data_race.rs
```diff
@@ -680,6 +684,27 @@ impl MemoryCellClocks {
     }
 }
 
+impl GlobalDataRaceHandler {
+    /// Select whether data race free actions should be allowed. This function should be used carefully!
+    ///
+    /// (This has no effect if no data race checker is enabled.)
+    ///
+    /// # PANICS
+    /// If data race free mode is attempted to be nested (i.e., cannot call this function two times in a row with `true`, or two times in a row with `false`).
```
RalfJung, line 693:
```suggestion
    /// Select whether data race checking is disabled. This is solely an
    /// implementation detail of `allow_data_races_*` and must not be used anywhere else!
```

src/machine.rs
```diff
@@ -1396,14 +1396,22 @@ impl<'tcx> Machine<'tcx> for MiriMachine<'tcx> {
...
-        } else if let AllocDataRaceHandler::Vclocks(data_race, weak_memory) = &alloc_extra.data_race
-        {
-            data_race.read(alloc_id, range, NaReadType::Read, None, machine)?;
-            if let Some(weak_memory) = weak_memory {
-                weak_memory.memory_accessed(range, machine.data_race.as_data_race_ref().unwrap());
+        match &machine.data_race {
+            GlobalDataRaceHandler::None => {}
+            GlobalDataRaceHandler::Genmc(genmc_ctx) => {
+                // We throw away the value we get from GenMC for non-atomic accesses, these are still handled by Miri
+                let _value = genmc_ctx.memory_load(machine, ptr.addr(), range.size)?;
```
RalfJung, line 1406: There's no reason the genmc wrapper should even return the value -- please discard it inside the wrapper.

src/machine.rs
```diff
@@ -1396,14 +1396,22 @@ impl<'tcx> Machine<'tcx> for MiriMachine<'tcx> {
...
+            }
+            GlobalDataRaceHandler::Vclocks(_data_race) => {
+                let AllocDataRaceHandler::Vclocks(data_race, weak_memory) = &alloc_extra.data_race
+                else {
+                    unreachable!();
+                };
```
RalfJung, line 1412: Can't this be `.as_vclocks_ref().unwrap()` or so?

src/concurrency/data_race_handler.rs
RalfJung: I think just on the fields is enough.

src/intrinsics/atomic.rs
```diff
@@ -199,23 +199,16 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
...
+            AtomicOp::Min =>
+                this.atomic_min_max_scalar(&place, rhs, /* min */ true, atomic)?,
+            AtomicOp::Max =>
+                this.atomic_min_max_scalar(&place, rhs, /* min */ false, atomic)?,
+            AtomicOp::MirOp(op, not) =>
+                this.atomic_rmw_op_immediate(&place, &rhs, op, not, atomic)?,
+        };
+        this.write_immediate(*old, dest)?; // old value is returned
+        interp_ok(())
```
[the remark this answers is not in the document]
RalfJung: Yes, that's very nice. :)

RalfJung:
We're almost there. :)

@rustbot author

src/concurrency/data_race_handler.rs
```diff
@@ -20,7 +18,7 @@ pub enum GlobalDataRaceHandler {
     /// - Atomic operations:    GenMC will ensure that we explore all possible values that the memory model allows
     ///   an atomic operation to see at any specific point of the program.
     Genmc(Rc<GenmcCtx>),
-    /// The default data race detector for Miri.
+    /// The default data race detector for Miri using vector-clocks.
```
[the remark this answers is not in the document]
Patrick-6: I used this without the "-", but then I saw it written this way [here](https://github.com/rust-lang/miri/blob/16ebd1241be53a5be29a49b1269d7dbc45d0004c/src/machine.rs#L335) (and in other places) and changed it ;)

Should I change the others too?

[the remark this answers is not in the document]
RalfJung: Ah yeah I guess I didn't carefully spell-check the big PR that added the checker.

Sure, you can change those as well.

src/machine.rs
Patrick-6: It can, but then I have to add either another accessor method that returns both `data_race` and `weak_memory`, or change the signature to of `as_vclocks_ref` to return both, which would mean matching away the `weak_memory` in a lot of places.

RalfJung: Ah, I didn't realize this is about also getting the weak_memory. Makes sense then, thanks!

src/concurrency/vector_clock.rs
```diff
@@ -40,7 +40,7 @@ impl From<u32> for VectorIdx {
     }
 }
 
-/// The size of the vector-clock to store inline
+/// The size of the vector clock to store inline
 /// clock vectors larger than this will be stored on the heap
```
RalfJung, line 44:
```suggestion
/// The size of the vector clock to store inline.
/// Clock vectors larger than this will be stored on the heap.
```
I know you didn't write this comment, but if we're touching it anyway we should at least make sure comments that consist of more than one sentence use `.` properly.

RalfJung:

@rustbot author

RalfJung:
Something went very wrong in that rebase. Your PR now includes unrelated recent changes. Please retry that.^^

src/concurrency/vector_clock.rs
RalfJung: This change got lost in the rebase.

PR #4293, opened 2025-04-25
Mark unsafe attributes as such

primoly:
Unsafe attributes are supported and recommended since Rust 1.82[^1] and mandated by edition 2024[^2].

[^1]: [Announcing Rust 1.82.0](https://blog.rust-lang.org/2024/10/17/Rust-1.82.0/#unsafe-attributes)
[^2]: [Unsafe attributes](https://doc.rust-lang.org/stable/edition-guide/rust-2024/unsafe-attributes.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -489,7 +489,7 @@ Miri knows where it is supposed to start execution:
 
 ```rust
 #[cfg(miri)]
-#[no_mangle]
+#[unsafe(no_mangle)]
 fn miri_start(argc: isize, argv: *const *const u8) -> isize {
     // Call the actual start function that your project implements, based on your target's conventions.
 }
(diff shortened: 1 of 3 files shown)


saethlin:
I'm happy to approve the README and error change. I'd rather migrate the whole project to the new edition instead of just doing one change that just affects the test suite. The main Rust source tree is on edition 2024 so there's no particular reason we can't, but I'd rather do it entirely or not at all.

RalfJung:
Just like the rustc test suite, our test suite doesn't use the latest edition by default -- and Miri itself doesn't use any unsafe attributes. So I think this is entirely independent of migrating Miri to the new edition.

Anyway, the minimal version of this PR (as-is) also works for me. Please [squash the commits](https://rustc-dev-guide.rust-lang.org/git.html#squash-your-commits), then we can land this. Please use the `--keep-base` flag when squashing so that the force-push diff is easier to review; do *not* rebase on the latest master branch. Then write `@rustbot ready` after you pushed the squashed PR.

@rustbot author

PR #4305, opened 2025-05-01
add ./miri squash

RalfJung:
This adds `./miri squash` which automatically does most of the work of squashing a feature branch into a single commit. The only thing the user still has to do is decide on the final commit message. Hopefully this reduces the number of rebase accidents. :)

`git rebase` is sadly not very scriptable so I had to make it invoke the miri-script as the "editor" for the rebase sequence

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,12 +19,10 @@ When you get a review, please take care of the requested changes in new commits.
 existing commits. Generally avoid force-pushing. The only time you should force push is when there
 is a conflict with the master branch (in that case you should rebase across master, not merge), and
 all the way at the end of the review process when the reviewer tells you that the PR is done and you
-should squash the commits. For the latter case, use `git rebase --keep-base ...` to squash without
-changing the base commit your PR branches off of. Use your own judgment and the reviewer's guidance
-to decide whether the PR should be squashed into a single commit or multiple logically separate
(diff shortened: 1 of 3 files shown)


saethlin:
I think this is very cool, but on this kind of code the only kind of review I think I could offer is trying it out. Which I have done, and it seems to work. I see the care you've taken to make this resilient, but also I fully expect someone to show up with a wacky git setup that breaks this.

But this feature should definitely reduce the number of rebase accidents. This is cool enough that someone might want to make this reusable in rust-lang/rust.

oli-obk:
At the point of providing our own tooling we may also suggest alternatively using jj as that has none of the problems and in case of visual jj is usable without knowing anything about jj itself

RalfJung:
@saethlin yeah I figured if this works well for us we could suggest it for inclusion in `./x` or so.

> I think this is very cool, but on this kind of code the only kind of review I think I could offer is trying it out. Which I have done, and it seems to work. I see the care you've taken to make this resilient, but also I fully expect someone to show up with a wacky git setup that breaks this.

Yeah, this will need some battle-testing in the real word. Thanks for trying it, I'll press the merge button based on this. :)

@oli-obk I'm not familiar with `jj` so I can't help/direct anyone using it.

oli-obk:
> I'm not familiar with `jj` so I can't help/direct anyone using it.

Ah then I know what I can spend lots of time at the all hands evangelizing.

PR #4306, opened 2025-05-02
Tree Borrows: Correctly handle interior mutable data in `Box`

yoctocell:
Previously, the pointee type would be behind a `*const` pointer, so `ty_is_freeze` would always be `true`, even if there was an `UnsafeCell` in `Box`.  The output of the `cell-inside-box.rs` test would be 

```
| Act |    └─┬──<TAG=root of the allocation>
| Res |      └────<TAG=a>
``` 
But the `Res` should be `ReIM` because we have an `UnsafeCell` in `Box`.


RalfJung:
Great find!

Is it possible to have a test that is incorrectly declared UB by current Miri but that should be allowed?

RalfJung:
Ah of course, in SB we don't treat `Box` as 2phase and then this clearly becomes UB.

Makes sense then, thanks!

PR #4307, opened 2025-05-02
Remove -Zunique-is-unique

JoJoDeveloping:
I propose we declare this experiment a failure and clean up the code a bit. I don't know who is using it, I fear no one is.


tests/fail/tree_borrows/children-can-alias.rs
RalfJung, line 1: This test doesn't make much sense any more in the form that you left it in, with all the comments stripped. I think the test can just be removed now (maybe double-check that we have an existing "pass" test equivalent to `raw_children_of_refmut_can_alias`).

tests/pass/tree_borrows/interior_mutability.rs
```diff
@@ -1,6 +1,5 @@
-//@revisions: default uniq
+//@revisions: default
```
RalfJung, line 1: This doesn't need to use revisions any more.

tests/pass/tree_borrows/tree-borrows.rs
```diff
@@ -1,6 +1,5 @@
-//@revisions: default uniq
+//@revisions: default
```
RalfJung, line 1: This doesn't need to use revisions any more.

tests/pass/tree_borrows/unique.rs
```diff
@@ -1,7 +1,6 @@
-//@revisions: default uniq
+//@revisions: default
```
RalfJung, line 1: This doesn't need to use revisions any more.

RalfJung (commented):
Yeah, I tend to agree -- it is clear that a viable model for `Unique` requires a permission that completely degenerates when the protector is over, i.e., a very different kind of uniquenss than what we have for `&mut`. We have https://github.com/rust-lang/unsafe-code-guidelines/issues/384 to track whether we want to do anything here in the future, and to record the outcome of this experiment.

There is an interesting discussion to be had whether we want this for all references -- this is tracked at https://github.com/rust-lang/unsafe-code-guidelines/issues/450.

tests/pass/tree_borrows/unique.rs
JoJoDeveloping: Presumably this test should also be removed? There is no special handling for `Unique` any more so nothing needs to be tested.

RalfJung: Yeah makes sense.

tests/fail/tree_borrows/children-can-alias.rs
JoJoDeveloping: I removed the file and moved the test to the general `tree_borrows.rs` test file

RalfJung: Isn't this an almost exact duplicate of `two_raw`?

JoJoDeveloping: Oh, yes. But we could also make the test here stronger and also do reads directly to/from the reference, which `two_raw` does not do since it was originally taken from SB where this is not allowed.

RalfJung: That's already pretty much covered by `local_addr_of_mut`.

JoJoDeveloping: ok, then we can simply remove the test

RalfJung:

@rustbot author

JoJoDeveloping:
@rustbot ready

PS: `./miri squash` worked very nicely

RalfJung:
> PS: ./miri squash worked very nicely

Thanks for testing it. :) I'll add it to my standard message then.

PR #4311, opened 2025-05-06
Update to edition 2024

est31:
Updates everything but the tests to edition 2024, of which a bunch are failing on older editions.

The edition 2024 is a prerequisite for removal of the `#![feature(let_chains)]`, and to forbidding let chains on edition 2021. I will do this in a later PR.

https://github.com/rust-lang/rust/issues/140722

### DIFF
--- a/tests/pass/coroutine.rs
+++ b/tests/pass/coroutine.rs
@@ -183,18 +183,18 @@ fn basic() {
 
 fn smoke_resume_arg() {
     fn drain<G: Coroutine<R, Yield = Y> + Unpin, R, Y>(
-        gen: &mut G,
+        gen_: &mut G,
         inout: Vec<(R, CoroutineState<Y, G::Return>)>,
     ) where
         Y: Debug + PartialEq,
         G::Return: Debug + PartialEq,
     {
-        let mut gen = Pin::new(gen);
+        let mut gen_ = Pin::new(gen_);
 
         for (input, out) in inout {
-            assert_eq!(gen.as_mut().resume(input), out);
+            assert_eq!(gen_.as_mut().resume(input), out);
             // Test if the coroutine is valid (according to type invariants).
-            let _ = unsafe { ManuallyDrop::new(ptr::read(gen.as_mut().get_unchecked_mut())) };
+            let _ = unsafe { ManuallyDrop::new(ptr::read(gen_.as_mut().get_unchecked_mut())) };
         }
     }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ name = "miri"
 repository = "https://github.com/rust-lang/miri"
 version = "0.1.0"
 default-run = "miri"
-edition = "2021"
+edition = "2024"
 
 [lib]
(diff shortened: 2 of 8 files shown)


bench-cargo-miri/backtraces/Cargo.toml
RalfJung, line 1: This might be easier if you only update Miri itself, not all the test / benchmark crates.

test-cargo-miri/no-std-smoke/Cargo.toml
RalfJung, line 1: How did you decide which of the test-cargo-miri creates to update?

est31: the rule I used was to look for usages of edition 2021. I see there is a bunch of crates on 2018 as well. Do you want me to update them as well, not update any of the bench crates, or keep this PR as is?

test-cargo-miri/cdylib/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 use byteorder::{BigEndian, ByteOrder};
 
-#[no_mangle]
-extern "C" fn use_the_dependency() {
+#[unsafe(no_mangle)]
+unsafe extern "C" fn use_the_dependency() {
```
RalfJung, line 4: Why did you make this function `unsafe`?

tests/pass/coroutine.rs
```diff
@@ -183,18 +183,18 @@ fn basic() {
 
 fn smoke_resume_arg() {
     fn drain<G: Coroutine<R, Yield = Y> + Unpin, R, Y>(
-        gen: &mut G,
+        gen_: &mut G,
```
RalfJung, line 186: Why is this change needed? You didn't change the edition used for tests. Is it due to the rustfmt change?

test-cargo-miri/no-std-smoke/Cargo.toml
RalfJung: > the rule I used was to look for usages of edition 2021

Ah, I see. I don't think we ever systematically decided which crates stay on 2018 vs move to 2021, but :shrug: sure why not.

tests/pass/coroutine.rs
est31: Yeah rustfmt needs the files to parse, and if there is a keyword (which it is on 2024), it doesn't like that.

test-cargo-miri/cdylib/src/lib.rs
est31: Fair point, it's not required indeed.

est31:
@RalfJung I have removed the changes to `cargo-miri` benches and tests and only updated miri itself, as well as `miri-script` and `cargo-miri`. This should be smaller overall. I have the changes to the benches and tests in a separate branch [here](https://github.com/rust-lang/miri/compare/master...est31:miri:edition_2024_cargo_miri_tests_benches).

RalfJung:
This looks great, thanks! Please squash the commits using `./miri squash`. Then write `@rustbot ready` after you pushed the squashed PR.

@rustbot author

PR #4312, opened 2025-05-07
miri-script: also overwrite CARGO_BUILD_BUILD_DIR

RalfJung:
Setting just `CARGO_TARGET_DIR` is no longer enough to control where the build artifacts will go, there's now also `CARGO_BUILD_BUILD_DIR`. See https://github.com/RalfJung/rustc-build-sysroot/issues/25, https://github.com/rust-lang/cargo/issues/14125.

### DIFF
--- a/miri-script/src/util.rs
+++ b/miri-script/src/util.rs
@@ -62,7 +62,10 @@ impl MiriEnv {
         }
 
         // Hard-code the target dir, since we rely on all binaries ending up in the same spot.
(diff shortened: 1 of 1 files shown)


RalfJung:
Thinking about it, we probably don't need this -- we care about the location of the final binaries, but `build.build-dir` doesn't affect that, it affects temporary artifacts only (IIUC).

PR #4314, opened 2025-05-08
TB: Track permissions on the byte-level

yoctocell:
This makes the tracking of permissions on the byte-level, like in Stacked Borrows, which makes the tracking of interior mutable data more fine-grained.

cc @RalfJung @JoJoDeveloping

### DIFF
--- a/src/borrow_tracker/tree_borrows/mod.rs
+++ b/src/borrow_tracker/tree_borrows/mod.rs
@@ -3,6 +3,8 @@ use rustc_middle::mir::{Mutability, RetagKind};
 use rustc_middle::ty::layout::HasTypingEnv;
 use rustc_middle::ty::{self, Ty};
 
+use self::foreign_access_skipping::IdempotentForeignAccess;
+use self::tree::LocationState;
 use crate::borrow_tracker::{GlobalState, GlobalStateInner, ProtectorKind};
 use crate::concurrency::data_race::NaReadType;
 use crate::*;
@@ -113,16 +115,19 @@ impl<'tcx> Tree {
 
 /// Policy for a new borrow.
 #[derive(Debug, Clone, Copy)]
-struct NewPermission {
-    /// Which permission should the pointer start with.
-    initial_state: Permission,
+pub struct NewPermission {
+    /// Permission for the frozen part of the range.
+    freeze_perm: Permission,
+    /// Whether a read access should be performed on the frozen part on a retag.
+    freeze_access: bool,
+    /// Permission for the non-frozen part of the range.
+    nonfreeze_perm: Permission,
+    /// Whether a read access should be performed on the non-frozen
+    /// part on a retag.
+    nonfreeze_access: bool,
     /// Whether this pointer is part of the arguments of a function call.
     /// `protector` is `Some(_)` for all pointers marked `noalias`.
     protector: Option<ProtectorKind>,
-    /// Whether a read should be performed on a retag.  This should be `false`
-    /// for `Cell` because this could cause data races when using thread-safe
-    /// data types like `Mutex<T>`.
-    initial_read: bool,
 }
 
 impl<'tcx> NewPermission {
@@ -133,27 +138,42 @@ impl<'tcx> NewPermission {
         kind: RetagKind,
         cx: &crate::MiriInterpCx<'tcx>,
     ) -> Option<Self> {
-        let ty_is_freeze = pointee.is_freeze(*cx.tcx, cx.typing_env());
         let ty_is_unpin = pointee.is_unpin(*cx.tcx, cx.typing_env());
         let is_protected = kind == RetagKind::FnEntry;
-        // As demonstrated by `tests/fail/tree_borrows/reservedim_spurious_write.rs`,
-        // interior mutability and protectors interact poorly.
-        // To eliminate the case of Protected Reserved IM we override interior mutability
-        // in the case of a protected reference: protected references are always considered
-        // "freeze" in their reservation phase.
-        let (initial_state, initial_read) = match mutability {
-            Mutability::Mut if ty_is_unpin =>
-                (Permission::new_reserved(ty_is_freeze, is_protected), true),
-            Mutability::Not if ty_is_freeze => (Permission::new_frozen(), true),
-            Mutability::Not if !ty_is_freeze => (Permission::new_cell(), false),
-            // Raw pointers never enter this function so they are not handled.
-            // However raw pointers are not the only pointers that take the parent
-            // tag, this also happens for `!Unpin` `&mut`s, which are excluded above.
-            _ => return None,
-        };
-
         let protector = is_protected.then_some(ProtectorKind::StrongProtector);
-        Some(Self { initial_state, protector, initial_read })
+
+        Some(match mutability {
+            Mutability::Mut if ty_is_unpin =>
+                NewPermission {
+                    freeze_perm: Permission::new_reserved(
+                        /* ty_is_freeze */ true,
+                        is_protected,
+                    ),
+                    freeze_access: true,
+                    nonfreeze_perm: Permission::new_reserved(
+                        /* ty_is_freeze */ false,
+                        is_protected,
+                    ),
+                    // If we have a mutable reference, then the non-frozen part will
+                    // have state `ReservedIM` or `Reserved`, which can have an initial read access
+                    // performed on it because you cannot have multiple mutable borrows.
+                    nonfreeze_access: true,
+                    protector,
+                },
+            Mutability::Not =>
+                NewPermission {
+                    freeze_perm: Permission::new_frozen(),
+                    freeze_access: true,
+                    nonfreeze_perm: Permission::new_cell(),
+                    // If it is a shared reference, then the non-frozen
+                    // part will have state `Cell`, which should not have an initial access,
+                    // as this can cause data races when using thread-safe data types like
+                    // `Mutex<T>`.
+                    nonfreeze_access: false,
+                    protector,
+                },
+            _ => return None,
+        })
     }
 
     /// Compute permission for `Box`-like type (`Box` always, and also `Unique` if enabled).
@@ -168,13 +188,17 @@ impl<'tcx> NewPermission {
         pointee.is_unpin(*cx.tcx, cx.typing_env()).then_some(()).map(|()| {
             // Regular `Unpin` box, give it `noalias` but only a weak protector
             // because it is valid to deallocate it within the function.
-            let ty_is_freeze = pointee.is_freeze(*cx.tcx, cx.typing_env());
-            let protected = kind == RetagKind::FnEntry;
-            let initial_state = Permission::new_reserved(ty_is_freeze, protected);
-            Self {
-                initial_state,
-                protector: protected.then_some(ProtectorKind::WeakProtector),
-                initial_read: true,
+            let is_protected = kind == RetagKind::FnEntry;
+            let protector = is_protected.then_some(ProtectorKind::WeakProtector);
+            NewPermission {
+                freeze_perm: Permission::new_reserved(/* ty_is_freeze */ true, is_protected),
+                freeze_access: true,
+                nonfreeze_perm: Permission::new_reserved(
+                    /* ty_is_freeze */ false,
+                    is_protected,
+                ),
+                nonfreeze_access: true,
+                protector,
             }
         })
     }
@@ -194,8 +218,6 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         new_tag: BorTag,
     ) -> InterpResult<'tcx, Option<Provenance>> {
         let this = self.eval_context_mut();
-        // Make sure the new permission makes sense as the initial permission of a fresh tag.
-        assert!(new_perm.initial_state.is_initial());
         // Ensure we bail out if the pointer goes out-of-bounds (see miri#1050).
         this.check_ptr_access(place.ptr(), ptr_size, CheckInAllocMsg::Dereferenceable)?;
 
@@ -206,7 +228,13 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             let global = this.machine.borrow_tracker.as_ref().unwrap().borrow();
             let ty = place.layout.ty;
             if global.tracked_pointer_tags.contains(&new_tag) {
-                let kind_str = format!("initial state {} (pointee type {ty})", new_perm.initial_state);
+                 let ty_is_freeze = ty.is_freeze(*this.tcx, this.typing_env());
+                 let kind_str =
+                     if ty_is_freeze {
+                         format!("initial state {} (pointee type {ty})", new_perm.freeze_perm)
+                     } else {
+                         format!("initial state {}/{} outside/inside UnsafeCell (pointee type {ty})", new_perm.freeze_perm, new_perm.nonfreeze_perm)
+                     };
                 this.emit_diagnostic(NonHaltingDiagnostic::CreatedPointerTag(
                     new_tag.inner(),
                     Some(kind_str),
@@ -285,42 +313,102 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
         let span = this.machine.current_span();
         let alloc_extra = this.get_alloc_extra(alloc_id)?;
-        let range = alloc_range(base_offset, ptr_size);
         let mut tree_borrows = alloc_extra.borrow_tracker_tb().borrow_mut();
 
-        // All reborrows incur a (possibly zero-sized) read access to the parent
-        if new_perm.initial_read {
-            tree_borrows.perform_access(
-                orig_tag,
-                Some((range, AccessKind::Read, diagnostics::AccessCause::Reborrow)),
-                this.machine.borrow_tracker.as_ref().unwrap(),
-                alloc_id,
-                this.machine.current_span(),
-            )?;
-        }
+        // Store initial permissions and their corresponding range.
+        let mut perms_map: RangeMap<LocationState> = RangeMap::new(
+            ptr_size,
+            LocationState::new_init(Permission::new_disabled(), IdempotentForeignAccess::None), // this will be overwritten
+        );
+        // Keep track of whether the node has any part that allows for interior mutability.
+        // FIXME: This misses `PhantomData<UnsafeCell<T>>` which could be considered a marker
+        // for requesting interior mutability.
+        let mut has_unsafe_cell = false;
+
+        // When adding a new node, the SIFA of its parents needs to be updated, potentially across
+        // the entire memory range. For the parts that are being accessed below, the access itself
+        // trivially takes care of that. However, we have to do some more work to also deal with
+        // the parts that are not being accessed. Specifically what we do is that we
+        // call `update_last_accessed_after_retag` on the SIFA of the permission set for the part of
+        // memory outside `perm_map` -- so that part is definitely taken care of. The remaining concern
+        // is the part of memory that is in the range of `perms_map`, but not accessed below.
+        // There we have two cases:
+        // * If we do have an `UnsafeCell` (`has_unsafe_cell` becomes true), then the non-accessed part
+        //   uses `nonfreeze_perm`, so the `nonfreeze_perm` initialized parts are also fine. We enforce
+        //   the `freeze_perm` parts to be accessed, and thus everything is taken care of.
+        // * If there is no `UnsafeCell`, then `freeze_perm` is used everywhere (both inside and outside the initial range),
+        //   and we update everything to have the `freeze_perm`'s SIFA, so there are no issues. (And this assert below is not
+        //   actually needed in this case).
+        assert!(new_perm.freeze_access);
+
+        let protected = new_perm.protector.is_some();
+        this.visit_freeze_sensitive(place, ptr_size, |range, frozen| {
+            has_unsafe_cell = has_unsafe_cell || !frozen;
+
+            // We are only ever `Frozen` inside the frozen bits.
+            let (perm, access) = if frozen {
+                (new_perm.freeze_perm, new_perm.freeze_access)
+            } else {
+                (new_perm.nonfreeze_perm, new_perm.nonfreeze_access)
+            };
+
+            // Store initial permissions.
+            for (_loc_range, loc) in perms_map.iter_mut(range.start, range.size) {
+                let sifa = perm.strongest_idempotent_foreign_access(protected);
+                // NOTE: Currently, `access` is false if and only if `perm` is Cell, so this `if`
+                // doesn't not change whether any code is UB or not. We could just always use
+                // `new_init` and everything would stay the same. But that seems conceptually
+                // odd, so we keep the initial "accessed" bit of the `LocationState` in sync with whether
+                // a read access is performed below.
+                if access {
+                    *loc = LocationState::new_init(perm, sifa);
+                } else {
+                    *loc = LocationState::new_uninit(perm, sifa);
+                }
+            }
+
+            // Some reborrows incur a read access to the parent.
+            if access {
+                // Adjust range to be relative to allocation start (rather than to `place`).
+                let mut range_in_alloc = range;
+                range_in_alloc.start += base_offset;
+
+                tree_borrows.perform_access(
+                    orig_tag,
+                    Some((range_in_alloc, AccessKind::Read, diagnostics::AccessCause::Reborrow)),
+                    this.machine.borrow_tracker.as_ref().unwrap(),
+                    alloc_id,
+                    this.machine.current_span(),
+                )?;
+
+                // Also inform the data race model (but only if any bytes are actually affected).
+                if range.size.bytes() > 0 {
+                    if let Some(data_race) = alloc_extra.data_race.as_vclocks_ref() {
+                        data_race.read(
+                            alloc_id,
+                            range_in_alloc,
+                            NaReadType::Retag,
+                            Some(place.layout.ty),
+                            &this.machine,
+                        )?
+                    }
+                }
+            }
+            interp_ok(())
+        })?;
+
         // Record the parent-child pair in the tree.
         tree_borrows.new_child(
+            base_offset,
             orig_tag,
             new_tag,
-            new_perm.initial_state,
-            range,
+            perms_map,
+            // Allow lazily writing to surrounding data if we found an `UnsafeCell`.
+            if has_unsafe_cell { new_perm.nonfreeze_perm } else { new_perm.freeze_perm },
+            protected,
             span,
-            new_perm.protector.is_some(),
         )?;
         drop(tree_borrows);
-
-        // Also inform the data race model (but only if any bytes are actually affected).
-        if range.size.bytes() > 0 && new_perm.initial_read {
-            if let Some(data_race) = alloc_extra.data_race.as_vclocks_ref() {
-                data_race.read(
-                    alloc_id,
-                    range,
-                    NaReadType::Retag,
-                    Some(place.layout.ty),
-                    &this.machine,
-                )?;
-            }
-        }
 
         interp_ok(Some(Provenance::Concrete { alloc_id, tag: new_tag }))
     }
@@ -508,15 +596,21 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     fn tb_protect_place(&mut self, place: &MPlaceTy<'tcx>) -> InterpResult<'tcx, MPlaceTy<'tcx>> {
         let this = self.eval_context_mut();
 
-        // Note: if we were to inline `new_reserved` below we would find out that
-        // `ty_is_freeze` is eventually unused because it appears in a `ty_is_freeze || true`.
-        // We are nevertheless including it here for clarity.
-        let ty_is_freeze = place.layout.ty.is_freeze(*this.tcx, this.typing_env());
         // Retag it. With protection! That is the entire point.
         let new_perm = NewPermission {
-            initial_state: Permission::new_reserved(ty_is_freeze, /* protected */ true),
+            // Note: If we are creating a protected Reserved, which can
+            // never be ReservedIM, the value of the `ty_is_freeze`
+            // argument doesn't matter
+            // (`ty_is_freeze || true` in `new_reserved` will always be `true`).
+            freeze_perm: Permission::new_reserved(
+                /* ty_is_freeze */ true, /* protected */ true,
+            ),
+            freeze_access: true,
+            nonfreeze_perm: Permission::new_reserved(
+                /* ty_is_freeze */ false, /* protected */ true,
+            ),
+            nonfreeze_access: true,
             protector: Some(ProtectorKind::StrongProtector),
-            initial_read: true,
         };
         this.tb_retag_place(place, new_perm)
     }
--- a/src/borrow_tracker/tree_borrows/mod.rs
+++ b/src/borrow_tracker/tree_borrows/mod.rs
@@ -97,7 +97,7 @@ impl<'tcx> Tree {
     /// A tag just lost its protector.
     ///
     /// This emits a special kind of access that is only applied
-    /// to initialized locations, as a protection against other
+    /// to accessed locations, as a protection against other
     /// tags not having been made aware of the existence of this
     /// protector.
     pub fn release_protector(
@@ -318,7 +318,7 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // Store initial permissions and their corresponding range.
         let mut perms_map: RangeMap<LocationState> = RangeMap::new(
             ptr_size,
-            LocationState::new_init(Permission::new_disabled(), IdempotentForeignAccess::None), // this will be overwritten
+            LocationState::new_accessed(Permission::new_disabled(), IdempotentForeignAccess::None), // this will be overwritten
         );
         // Keep track of whether the node has any part that allows for interior mutability.
         // FIXME: This misses `PhantomData<UnsafeCell<T>>` which could be considered a marker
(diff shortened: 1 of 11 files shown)


src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -205,8 +227,14 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
          -> InterpResult<'tcx> {
             let global = this.machine.borrow_tracker.as_ref().unwrap().borrow();
             let ty = place.layout.ty;
+            let ty_is_freeze = ty.is_freeze(*this.tcx, this.typing_env());
```
RalfJung, line 230: `is_freeze` can be expensive, so only execute it when needed. Here, you are only using the result inside the `if`.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -205,8 +227,14 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             if global.tracked_pointer_tags.contains(&new_tag) {
-                let kind_str = format!("initial state {} (pointee type {ty})", new_perm.initial_state);
+                 let kind_str =
+                     if ty_is_freeze {
+                         format!("initial state {} (pointee type {ty})", new_perm.freeze_perm)
+                     } else {
+                         format!("initial frozen/non-frozen states {}/{} (pointee type {ty})", new_perm.freeze_perm, new_perm.nonfreeze_perm)
```
RalfJung, line 236:
```suggestion
                         format!("initial state {}/{} outside/inside UnsafeCell (pointee type {ty})", new_perm.freeze_perm, new_perm.nonfreeze_perm)
```

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -285,42 +313,65 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                 Some((range, AccessKind::Read, diagnostics::AccessCause::Reborrow)),
                 this.machine.borrow_tracker.as_ref().unwrap(),
                 alloc_id,
                 this.machine.current_span(),
-            )?;
-        }
+            )
+        };
+
+        // Some reborrows incur a read access to the parent
```
RalfJung, line 328: This doesn't just perform the read access, this does the full initialization, right?

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -285,42 +313,65 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                )?;
+                )?
             }
-        }
+            interp_ok(())
+        };
+
+        // Also inform the data race model (but only if any bytes are actually affected).
+        let NewPermission { freeze_access, nonfreeze_access, .. } = new_perm;
+        this.visit_freeze_sensitive(place, ptr_size, |mut range, frozen| {
```
RalfJung, line 367: What is going on here? We should definitely not call `visit_freeze_sensitive` twice.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -640,12 +651,23 @@ impl<'tcx> Tree {
...
         self.nodes.get_mut(parent_idx).unwrap().children.push(idx);
-        // Initialize perms
-        let perm = LocationState::new_init(default_initial_perm, strongest_idempotent);
-        for (_perms_range, perms) in self.rperms.iter_mut(reborrow_range.start, reborrow_range.size)
-        {
-            perms.insert(idx, perm);
-        }
+
+        let NewPermission { freeze_perm, nonfreeze_perm, .. } = perm;
+        this.visit_freeze_sensitive(place, size, |mut range, frozen| {
```
RalfJung, line 656: And here we have a third call to `visit_freeze_sensitive`? I don't understand the structure of this, but it's definitely not the structure I was hoping for. Check out Stacked Borrows for a template.

tests/pass/tree_borrows/cell-inside-struct.rs
RalfJung, line 1: This test could be entirely written in safe code, right? 
```rust
let x: &mut Foo = ...;
x.field1 = 88;
```
A whole separate test for this seems like overkill. You can add a new function to `tests/pass/both_borrows/basic_aliasing_model.rs` if you want (using safe code only and not using the TB debug macros).

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -113,47 +113,77 @@ impl<'tcx> Tree {
...
+                    // To eliminate the case of Protected Reserved IM we override interior mutability
+                    // in the case of a protected reference: protected references are always considered
+                    // "freeze" in their reservation phase.
+                    //
+                    // If we are creating protected Reserved, which can
+                    // never be ReservedIM, the value of the `ty_is_freeze`
+                    // argument doesn't matter (`ty_is_freeze || true` will always be `true`).
+                    Permission::new_reserved(true, is_protected),
+                    Permission::new_reserved(false, is_protected),
+                    true,
```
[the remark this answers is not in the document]
JoJoDeveloping: It's what we used to do and no one complained

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -113,47 +113,77 @@ impl<'tcx> Tree {
...
+                (
+                    // As demonstrated by `tests/fail/tree_borrows/reservedim_spurious_write.rs`,
+                    // interior mutability and protectors interact poorly.
+                    // To eliminate the case of Protected Reserved IM we override interior mutability
+                    // in the case of a protected reference: protected references are always considered
+                    // "freeze" in their reservation phase.
+                    //
+                    // If we are creating protected Reserved, which can
+                    // never be ReservedIM, the value of the `ty_is_freeze`
+                    // argument doesn't matter (`ty_is_freeze || true` will always be `true`).
```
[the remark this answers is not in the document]
yoctocell: The `true` is the `protected` argument in `new_reserved` when
we enable protectors

```rust
        if ty_is_freeze || protected { Self::new_reserved_frz() } else { Self::new_reserved_im() }
```

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -113,47 +113,77 @@ impl<'tcx> Tree {
...
+    /// Whether a read access should be performed on the non-frozen
+    /// part on a retag.  This should be `false` for `Cell` because
+    /// this could cause data races when using thread-safe data types
+    /// like `Mutex<T>`.
+    nonfreeze_access: bool,
 }
 
 impl<'tcx> NewPermission {
+    /// Helper method to create a NewPermission for a reference.
+    fn new(mutability: Mutability, is_protected: bool, protector: ProtectorKind) -> Self {
```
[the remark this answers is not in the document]
yoctocell: I created this when `NewPermission` was defined as a enum (similar to
in SB), and the `match` had more cases, but maybe the function doesn't
provide much value now.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -194,8 +218,6 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         new_tag: BorTag,
     ) -> InterpResult<'tcx, Option<Provenance>> {
         let this = self.eval_context_mut();
-        // Make sure the new permission makes sense as the initial permission of a fresh tag.
-        assert!(new_perm.initial_state.is_initial());
```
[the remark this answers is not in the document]
yoctocell: I think this was also when `NewPermission` was defined as an enum, and
`initial_state` was a method instead of a field.  But `initial_state()` was
basically only used here, and the assert was never triggered
because non-initial states would only be created (using `new_disabled`
or `new_active`) when creating the root.  So the assert didn't seem so
necessary.

src/borrow_tracker/tree_borrows/mod.rs
yoctocell: What do you mean by "full initialization"?  At this point, the
node hasn't been inserted into the tree yet; this happens in the
`new_child` method, called later.

src/borrow_tracker/tree_borrows/mod.rs
yoctocell: I tried to put the `perform_parent_access` part into the
`visit_freeze_sensitive` call in `new_child`, so we can remove this
`visit_freeze_sensitive` call, but this causes some tests to fail.

src/borrow_tracker/tree_borrows/tree.rs
yoctocell: Putting this part into the `visit_freeze_sensitive` call in
`new_child` seems to work (all tests pass), but I don't know if this
will interfere with the `update_last_accessed_after_retag` call.  I
will look more into this.

tests/pass/tree_borrows/cell-inside-struct.rs
yoctocell: Yes, apart from printing the tree, everyting else is safe.  But I
think the main point is the tree, because previously, the tracking of
`UnsafeCell`s wasn't as fine grained, so the tree used to look
like this

```
| Act |    └─┬──<TAG=root of the allocation>
| ReIM|      └────<TAG=a>
```

JoJoDeveloping:
So, how does adding a new node work (and why do we need to call `visit_freeze_sensitive` three times:

* First, you do a read access at the parent node
* Then, you add the new node. Note that adding a new node is a per-block operation, not per-access
* Then you set the correct permissions for the new node
* Then you reset the SIFA thingy (also per-block)
* Then you call into the data race model

For Stacked Borrows, there are no per-block operations: each stack at each offset its own stack and completely independent from the other ones at different offsets in the same allocation. So there, more (basically everything) can be moved into the per-offset function.

So what do we do for Tree Borrows? One answer is to combine parent reading, permission setting and data race notifying together. This would be somewhat ugly, because you're constructing a node, while also accessing the tree, and this would only be fine because you're careful to not actually add the node into the `children` array of its parent so that it's not visited when walking the tree.

Should we do this?

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -168,14 +198,8 @@ impl<'tcx> NewPermission {
...
             // because it is valid to deallocate it within the function.
-            let ty_is_freeze = pointee.is_freeze(*cx.tcx, cx.typing_env());
             let protected = kind == RetagKind::FnEntry;
-            let initial_state = Permission::new_reserved(ty_is_freeze, protected);
-            Self {
-                initial_state,
-                protector: protected.then_some(ProtectorKind::WeakProtector),
-                initial_read: true,
-            }
+            NewPermission::new(Mutability::Mut, protected, ProtectorKind::WeakProtector)
```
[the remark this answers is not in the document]
JoJoDeveloping: I think the code without a function like `new` (or maybe `new_mut`) would be less readable.

src/borrow_tracker/tree_borrows/mod.rs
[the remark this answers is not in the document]
RalfJung: That's not really an answer ;)

My guess is that: we do decorate `&mut Cell` arguments with `noalias dereferenceable`, so we have to do something on these retags. We should ensure we have a test that catches this.

src/borrow_tracker/tree_borrows/mod.rs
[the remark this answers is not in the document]
RalfJung: The current code is hard to read, the logic for determining the new permissions of a pointer is spread across too much code. The entire point of having a `NewPermission` type is to have that logic visible in a central place, fitting on a single screen.

So this should all be in a single function, except when there is some obvious, easy-to-explain concept that can be factored out into a shared reusable function with a self-explaining name. `new` explains nothing, so the current approach does not qualify for that exception.

src/borrow_tracker/tree_borrows/mod.rs
RalfJung: So what is the reasons for those tests to fail?

tests/pass/tree_borrows/cell-inside-struct.rs
RalfJung: The test should be written in a way that it is UB or not. The tree are useful for debugging but they should not be load-bearing for a test. Consider that `--bless` will happily overwrite the tree! All tests mut be written in a way that even with `--bless`, the key point of the test always remains in place.

RalfJung:
Ignoring the SIFA thingy since I already forgot how it works, what I think would make sense is to do a single `visit_freeze_sensitive` where you
- do the parent access
- do the data race access
- prepare a `RangeMap<LocationState>` of per-location initial states that will later be passed to `Tree::new_child`

Does that makes sense or is it too ugly as well?

src/borrow_tracker/tree_borrows/mod.rs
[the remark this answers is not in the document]
RalfJung: Okay.... so what is the point of this comment then? I am still confused.

We do set `ty_is_freeze` properly here, right? So why is it relevant that the value does not matter?

src/borrow_tracker/tree_borrows/mod.rs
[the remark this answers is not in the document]
JoJoDeveloping: I suggested removing this assertion. I never considered the split into "initial"/"non-initial" permissions useful. For example `Active` is an initial permission for the root, as is `Disabled` for the root outside of the available range.

Currently, all permissions that ever arrive there are set in `NewPermission::new`, and it is very easy to see there that these are the "correct" permissions. If this assertion was to exist anywhere, it should exist in the constructor of `NewPermission`, not at a random place consuming `NewPermission`.

JoJoDeveloping:
That presumably works. It has to construct an extra `RangeMap` but presumably this is cheaper than calling `visit_freeze_sensitive` thrice.

src/borrow_tracker/tree_borrows/mod.rs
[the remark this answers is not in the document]
RalfJung: The assertion serves as a sanity check, a 2nd line of defense. `Active` works for the root but not for anything else, so that only confirms that it's important to have the assertion as that seems easy to mix up.

RalfJung:
> presumably this is cheaper than calling visit_freeze_sensitive thrice.

I'm not sure, but that's at least plausible. I also prefer it conceptually.

tests/pass/tree_borrows/cell-inside-struct.rs
yoctocell: OK, then I can add the test to
`tests/pass/both_borrows/basic_aliasing_model.rs` since I don't think
some thing like this is tested (at least not directly).

src/borrow_tracker/tree_borrows/mod.rs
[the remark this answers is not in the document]
JoJoDeveloping: what should the test test?

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -133,27 +137,47 @@ impl<'tcx> NewPermission {
...
-            Mutability::Not if ty_is_freeze => (Permission::new_frozen(), true),
-            Mutability::Not if !ty_is_freeze => (Permission::new_cell(), false),
-            // Raw pointers never enter this function so they are not handled.
-            // However raw pointers are not the only pointers that take the parent
-            // tag, this also happens for `!Unpin` `&mut`s, which are excluded above.
+            // As demonstrated by `tests/fail/tree_borrows/reservedim_spurious_write.rs`,
+            // interior mutability and protectors interact poorly.
+            // To eliminate the case of Protected Reserved IM we override interior mutability
+            // in the case of a protected reference: protected references are always considered
+            // "freeze" in their reservation phase.
```
JoJoDeveloping, line 150: This comment should really be in `Permission::new_reserved` where this logic is implemented. Apart from this I think everything has been addressed.

src/borrow_tracker/tree_borrows/mod.rs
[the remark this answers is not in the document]
RalfJung: Well, jsut because there's a confusing comment in master doesn't mean it should stay confusing. ;)

src/borrow_tracker/tree_borrows/mod.rs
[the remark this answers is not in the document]
RalfJung: It should be a test that fails when we change the `true` to `false` here.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -285,43 +314,76 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let prot = new_perm.protector.is_some();
+
+        // Store initialized permissions and their corresponding range.
+        let mut perms_map: RangeMap<LocationState> = RangeMap::new(
+            base_offset + ptr_size,
+            LocationState::new_uninit(
+                Permission::new_disabled(),
+                foreign_access_skipping::IdempotentForeignAccess::None,
+            ),
+        );
```
RalfJung, line 327: I think this should be initialized with size `ptr_size`, and only cover the place we are retagging. That way we can avoid introducing some `disabled` states here that I don't think we actually want.

RalfJung (commented):
Thanks! This is a lot better.

I don't fully understand the logic around the new `perms_map` though. It seems like the map is created to cover all the bytes from the beginning of the allocation to the end of the retagged place, which is odd -- it should be either the entire allocation, or just the place. And given that we do *not* want to actually put anything into the `perms` map outside the place, I think it should be just the place.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -31,7 +33,7 @@ mod tests;
 
 /// Data for a single *location*.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-pub(super) struct LocationState {
+pub(crate) struct LocationState {
```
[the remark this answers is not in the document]
yoctocell: Because the compiler would complain about the visibility of
`LocationState` being private than `new_child`.  But seems like `super`
should be enough for `new_child`, so I will change both of the to `super`.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -613,18 +615,27 @@ impl Tree {
...
-        reborrow_range: AllocRange,
+        perm: NewPermission,
+        perms_map: RangeMap<LocationState>,
         span: Span,
-        prot: bool,
     ) -> InterpResult<'tcx> {
-        assert!(!self.tag_mapping.contains_key(&new_tag));
         let idx = self.tag_mapping.insert(new_tag);
         let parent_idx = self.tag_mapping.get(&parent_tag).unwrap();
+        let default_initial_perm = perm.freeze_perm;
```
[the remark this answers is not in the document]
RalfJung:
```rust
fn foo(x: &Cell<i32>) {
  let ptr = x as *const Cell<i32> as *mut _  as *mut i32;
  ptr.offset(1).write(0);
}

let arr = [Cell::new(1), Cell::new(1)];
foo(&arr[0]);
```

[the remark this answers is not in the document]
yoctocell: But something like this should still be UB, right?

```rust
fn foo(x: &Cell<i32>) {
  let ptr = x as *const Cell<i32> as *mut _  as *mut i32;
  ptr.offset(1).write(0);
}

let pair = (Cell::new(1), 1);
foo(&pair.0);
```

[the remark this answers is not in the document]
RalfJung: Hm... I expected that also to be allowed (for now), since it's basically the same as my example. We need some way to express "a shared reference with the permission to access interior mutable things outside the type it references", and your example is collateral damage that gets accepted as well then.

The key point achieved by your work is that this becomes UB:
```rust
fn foo(x: &(Cell<i32>, i32)) {
  let ptr = x as *const _ as *mut _  as *mut i32;
  ptr.offset(1).write(0);
}

let pair = (Cell::new(1), 1);
foo(&pair);
```

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -317,32 +317,33 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
         );
+        // Keep track of whether the node has any part that allows for interior mutability.
+        let mut is_frozen = true;
 
-        let NewPermission { freeze_perm, freeze_access, nonfreeze_perm, nonfreeze_access, .. } =
-            new_perm;
         this.visit_freeze_sensitive(place, ptr_size, |mut range, frozen| {
+            is_frozen = frozen;
             // Adjust range.
             range.start += base_offset;
```
RalfJung, line 334: This is no longer correct now that `perms_map` is relative to the place, I think?

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -317,32 +317,33 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                 foreign_access_skipping::IdempotentForeignAccess::None,
             ),
         );
+        // Keep track of whether the node has any part that allows for interior mutability.
+        let mut is_frozen = true;
 
-        let NewPermission { freeze_perm, freeze_access, nonfreeze_perm, nonfreeze_access, .. } =
-            new_perm;
         this.visit_freeze_sensitive(place, ptr_size, |mut range, frozen| {
+            is_frozen = frozen;
```
RalfJung, line 332:
```suggestion
            is_frozen = is_frozen || frozen;
```
Otherwise this does not seem to match the comment?

src/borrow_tracker/tree_borrows/mod.rs
RalfJung: It is quite strange that this does not cause any test failures. Can you investigate that?

src/borrow_tracker/tree_borrows/mod.rs
RalfJung: Ah right. We'd need a bunch of negations. I suggest you rename the variable to `has_unsafe_cell`, then the logic becomes a lot simpler -- `has_unsafe_cell = has_unsafe_cell || !frozen`.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -613,18 +615,30 @@ impl Tree {
...
-        assert!(!self.tag_mapping.contains_key(&new_tag));
         let idx = self.tag_mapping.insert(new_tag);
         let parent_idx = self.tag_mapping.get(&parent_tag).unwrap();
+        // Allow lazily writing to surrounding data if we have a reference to interior mutable data.
+        let default_initial_perm =
+            if has_unsafe_cell { perm.nonfreeze_perm } else { perm.freeze_perm };
+        let prot = perm.protector.is_some();
+
+        // SIFA of the frozen part must be weaker than SIFA of the non-frozen part, otherwise
+        // `self.update_last_accessed_after_retag` will break the SIFA invariant (see `foreign_access_skipping.rs`).
```
RalfJung, line 637: I don't understand why this assertion must hold. `foreign_access_skipping` doesn't explain it either, it does not mention frozen vs non-frozen parts at all.

RalfJung:
I have pushed a commit with various minor cleanups -- please take a look, I hope they make sense. :)

However, in doing so I stumbled upon a [question](https://github.com/rust-lang/miri/pull/4314#discussion_r2104168200) I could not answer.

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: We compute the `strongest_idempotent_foreign_access` separately for each byte in the `perms_map`. So I don't understand what you are saying.

> since the strongest_idempotent argument might come from the frozen permission

It can only come from the frozen permission for bytes that we freeze. Right?

yoctocell: Ah right, that's true.  Then the assert doesn't seem so necessary in this case.

JoJoDeveloping: This assertion is necessary, but the code wrong elsewhere.

It is correct for the case where there is no `UnsafeCell`: We use `freeze_perm` everywhere, and this has some SIFA, and after the retag we reset the parent's recorded accesses to be no stronger than that SIFA. (Example: `freeze_perm = Frozen`, this survives foreign reads but not writes so the SIFA is `Read`, and if a parent indicated it survived foreign `Write`s, we reset that to `Read`).

But if there is an `UnsafeCell`, things go wrong. The code uses the SIFA of the `default_permission`, which is `nonfreeze_perm` (think `Cell`, which survives all foreign accesses so the SIFA is `Write`). But not all permissions across the entire range survive that permissions's strongest initial foreign access, some are outside of the `UnsafeCell` and thus have `freeze_perm` (`Frozen`), which has a weaker SIFA as shown by the above assert. (here: `Read`).

But the correctness criterion is that we call `update_last_accessed_after_retag` with a foreign access that *all* permissions across the entire range of the newly inserted node are idempotent under. (We could always pass `None`, as mentioned by the documentation, but that would be inefficient). So we need to always pass in the `nonfreeze_perm`'s SIFA here.

I'm working on a counterexample showing that the current implementation is wrong, which I'll post soon.

RalfJung: > So we need to always pass in the nonfreeze_perm's SIFA here.

More generally, we need to pass in the minimum of the two SIFAs for those two modes, right? Then we don't need an assertion either.
(Or is it the maximum?)

JoJoDeveloping: This whole SIFA code is very confusing, but I think the code is correct currently, but only accidentally. The call to `update_last_accessed_after_retag` is only necessary for the permissions outside of the "active" area of the retag (where the implicit read happens), and there we use the SIFA of the permission that is uniformly set around the active area. So it's correct, and the assertion is not necessary.

But this is rather surprising, since this is not mentioned everywhere. I was originally confused by this, which is why I posted a wrong comment :see_no_evil: So either
* we remove the assertion, and add a large comment explaining this, or
* we keep the assertion, and always use `freeze_perm`'s SIFA there (the weaker one), which only relies on the functions' documented behavior but is maybe a bit less efficient.

yoctocell: Yeah, it would be the minimum

JoJoDeveloping: > we need to pass in the minimum of the two SIFAs

yes, that's option 2 but replacing the assertion by computing the minimum. But this should always be the one for the non-UnsafeCelly bits, since UnsafeCell allows more behavior (and thus survives more foreign accesses)

RalfJung: >  The call to update_last_accessed_after_retag is only necessary for the permissions outside of the "active" area of the retag (where the implicit read happens), and there we use the SIFA of the permission that is uniformly set around the active area. 

Ah, great. :)

I am all for updating the comments and removing the assertion, then.

RalfJung: But note that `access` can be `false`, i.e. we don't always do an implicit read...

RalfJung: Yeah but we should have logic that is either correct for everything we can put into `NewPermission`, or have assertions ruling out the cases where the logic would be incorrect.

JoJoDeveloping: > correct for everything we can put into `NewPermission`

Then I would say we add some more assertions than we have currently. Just to rule out nasty corner cases before someone thinks of them. At the very least, keep the current one.

PS: I am not yet 100% convinced the code is correct as-is, I need to ponder it some more.

JoJoDeveloping: We restructured it this morning, I pushed the results. This replaces the SIFA assertion by one ensuring that `nonfreeze_access` is always true. Presumably we want to remove that field then?

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -323,6 +323,20 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        // * If we do have an `UnsafeCell` (`has_unsafe_cell` becomes true), then
+        //   + some parts inside the initial range of the reference have `nonfreeze_perm`
+        //   + the "dynamic" area outside the initial range of the reference also have `nonfreeze_perm`
+        //   and thus it is sufficient to reset the strongest foreign access of all parents to the
+        //   `nonfreeze_perm` SIFA, because all the parts where we put `freeze_perm` already had a read
+        //   access happen, and thus the SFAs there are already reset. The assert below ensures that this
+        //   read access to the `freeze_perm` happens.
+        // * If there is no `UnsafeCell`, then `freeze_perm` is used everywhere (both inside and outside the initial range),
+        //   and we update everything to have the `freeze_perm`'s SIFA, so there are no issues. (And this assert below is not
+        //   actually needed in this case).
```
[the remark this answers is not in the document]
JoJoDeveloping: Yes that looks correct. I would pick one of `(accessed|initialized)` and use it consistently 🙃 but naming is hard

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -671,6 +671,7 @@ impl<'tcx> Tree {
         // We now weaken the recorded SIFA for our parents, until the invariant is restored.
         // We could weaken them all to `LocalAccess`, but it is more efficient to compute the SIFA
         // for the new permission statically, and use that.
+        // See the comment in `tb_reborrow` for why it is correct to use the SIFA of `default_uninit_perm`.
```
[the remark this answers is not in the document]
RalfJung: Yeah, something like that.

We could even have (debug) assertions in this function to check that, right?

src/borrow_tracker/tree_borrows/mod.rs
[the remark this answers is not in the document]
RalfJung: We don't mention this concept in the paper, do we?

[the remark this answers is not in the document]
RalfJung: (after brief discussion with Johannes) let's go for "accessed" everywhere. It should also be renamed in MiniRust then.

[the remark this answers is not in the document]
RalfJung: The "init" in `perm_init` is for "initial", as in "initial state", not for "initialized". So that one should not be renamed. Only the ones that refer to "initialized" should be renamed.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -612,46 +613,66 @@ impl Tree {
...
         {
-            perms.insert(idx, perm);
+            assert!(perm.is_initial());
+            for (_perms_range, perms) in self
+                .rperms
+                .iter_mut(Size::from_bytes(start) + base_offset, Size::from_bytes(end - start))
+            {
+                perms.insert(
+                    idx,
+                    LocationState::new_init(
```
JoJoDeveloping, line 662: After discussion with Ralf, we realized that this is not quite correct. Whether this is `init` or not depends on whether there was an initial read (i.e. the corresponding `access` bool was set in the `NewPermission` struct).

Presumably, the new `LocationState`s should constructed outside this method, and this method should instead be given a `RangeMap<LocationState>`.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -285,43 +311,90 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // We are only ever `Frozen` inside the frozen bits.
+            let (perm_init, access) = if frozen {
+                (new_perm.freeze_perm, new_perm.freeze_access)
+            } else {
+                (new_perm.nonfreeze_perm, new_perm.nonfreeze_access)
+            };
+
+            // Store initial permissions.
+            for (_perm_range, perm) in perms_map.iter_mut(range.start, range.size) {
+                *perm = perm_init;
```
JoJoDeveloping, line 352: here, it should create a new `LocationState` that is init if any only if `access` is set.

JoJoDeveloping (changes requested):
There's one more nit about the fact that we use `LocationState::new_init` even if no initial access happens.

RalfJung:
Thanks, I think the logic is all good now! Now it's just fine-tuning comments and names.

Also there is [one case](https://github.com/rust-lang/miri/pull/4314#discussion_r2111671711) where I wonder if we want a test to ensure that this is what we are doing -- if there even could be such a test.

@rustbot author

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -348,8 +352,13 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             // Store initial permissions.
-            for (_perm_range, perm) in perms_map.iter_mut(range.start, range.size) {
-                *perm = perm_init;
+            for (_loc_range, loc) in perms_map.iter_mut(range.start, range.size) {
+                let sifa = perm_init.strongest_idempotent_foreign_access(protected);
+                if access {
+                    *loc = LocationState::new_accessed(perm_init, sifa);
+                } else {
+                    *loc = LocationState::new_non_accessed(perm_init, sifa);
+                }
```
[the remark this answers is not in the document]
yoctocell: On first thought, I would say no, because the `access` is false if and
only if `perm` is Cell.  Since every kind of access is a noop for Cell,
I don't think it should have any impact, whether the location is
`accessed` or not.  I will see if I can find any counterexample.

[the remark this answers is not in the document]
JoJoDeveloping: Yes it's only for `Cell`, so it does not affect observable behavior. But it might in the future, so it's good for this to be tracked properly?

RalfJung:
All right, I think this is read. :)
Please squash the commit history so that each commit makes independent sense. If that's too much work you can also squash everything into a single commit. (`./miri squash` can help you do that.)

@rustbot author

RalfJung:
FWIW the comment was outdated, it should say `default_perm` now.

RalfJung:
Looks like a timeout, let's try this again...

PR #4322, opened 2025-05-16
Run some tests in ``disable-isolation``

tiif:
I was confused why tests that require ``disable-isolation`` are currently being run in isolation mode.

So I just took a stab and just move tests around.

### DIFF
--- a/tests/pass-dep/libc/libc-fs-with-isolation.rs
+++ b/tests/pass-dep/libc/libc-fs-with-isolation.rs
@@ -1,26 +1,8 @@
 //@ignore-target: windows # File handling is not implemented yet
-//@compile-flags: -Zmiri-isolation-error=warn-nobacktrace
-//@normalize-stderr-test: "(stat(x)?)" -> "$$STAT"
-
-use std::fs;
-use std::io::{Error, ErrorKind};
 
 fn main() {
     // test `fcntl(F_DUPFD): should work even with isolation.`
     unsafe {
         assert!(libc::fcntl(1, libc::F_DUPFD, 0) >= 0);
     }
-
-    // test `readlink`
-    let mut buf = vec![0; "foo_link.txt".len() + 1];
-    unsafe {
-        assert_eq!(libc::readlink(c"foo.txt".as_ptr(), buf.as_mut_ptr(), buf.len()), -1);
-        assert_eq!(Error::last_os_error().raw_os_error(), Some(libc::EACCES));
-    }
-
-    // test `stat`
-    let err = fs::metadata("foo.txt").unwrap_err();
-    assert_eq!(err.kind(), ErrorKind::PermissionDenied);
-    // check that it is the right kind of `PermissionDenied`
-    assert_eq!(err.raw_os_error(), Some(libc::EACCES));
 }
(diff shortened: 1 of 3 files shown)


tests/pass-dep/libc/libc-fs.rs
```diff
@@ -451,3 +454,20 @@ fn test_ioctl() {
...
+    }
+}
+
+/// Basic test for `stat`.
+fn test_stat() {
+    let err = fs::metadata("foo.txt").unwrap_err();
+    assert_eq!(err.kind(), ErrorKind::NotFound);
+    // check that it is the right kind of `PermissionDenied`
+    assert_eq!(err.raw_os_error(), Some(libc::ENOENT));
+}
```
tiif, line 473: There is a subtle difference from previous test. The error for the previous test is ``libc::EACCES``.

RalfJung: That's wrong then. The comment even says that we want `PermissionDenied`.

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -451,3 +454,20 @@ fn test_ioctl() {
...
 }
+
+/// Basic test for `readlink`.
+fn test_readlink() {
+    let mut buf = vec![0; "foo_link.txt".len() + 1];
+    unsafe {
+        assert_eq!(libc::readlink(c"foo.txt".as_ptr(), buf.as_mut_ptr(), buf.len()), -1);
+        assert_eq!(Error::last_os_error().raw_os_error(), Some(libc::ENOENT));
+    }
+}
```
RalfJung, line 465: Why does this not give a loud error with isolation? That seems like a pretty bad bug?

RalfJung:
Looking at this again when I am more awake ;)

These tests do not "require" `disable-isolation`. These tests test what happens when you have isolation disabled, and set `-Zmiri-isolation-error=warn-nobacktrace`. It's important to test that. So we definitely do not want to move these tests anywhere else -- we want to test the behavior of all our APIs both with and without isolation.

tiif:
> These tests do not "require disable-isolation". These tests test what happens when you have isolation enabled, and set -Zmiri-isolation-error=warn-nobacktrace. It's important to test that. So we definitely do not want to move these tests anywhere else -- we want to test the behavior of all our APIs both with and without isolation.

Ahh this makes sense, thanks for explaining! I will move the test back and add this to comment later.

tiif:
Alright, I just add more comment there

@rustbot ready

tests/pass-dep/libc/libc-fs-with-isolation.rs
```diff
@@ -11,6 +11,10 @@ fn main() {
         assert!(libc::fcntl(1, libc::F_DUPFD, 0) >= 0);
     }
 
+    // Although `readlink` and `stat` require disable-isolation mode
+    // to properly run, they are tested with isolation mode on to check the error emitted
+    // after the isolation error is thrown.
```
RalfJung, line 16:
```suggestion
    // to properly run, they are tested with isolation mode on to check the error emitted
    // with `-Zmiri-isolation-error=warn-nobacktrace`.
```

RalfJung:
This looks great, thanks! After resolving the last nit, please squash the commits using `./miri squash`. Then write `@rustbot ready` after you pushed the squashed PR.

@rustbot author

Issue #11, opened 2016-04-30
Consider using libffi to allow for external calls


drbo:
It would be nice to use [libffi][1] to allow for external calls.

[1]: https://sourceware.org/libffi/

eddyb:
Previously suggested on IRC based on @cmr's mentions in a different context.
The [closure example](http://www.atmark-techno.com/~yashi/libffi.html#index-closures-34) presents what I think is one of the most valuable features of this library: the ability to JIT a C function which hardcodes extra owned values, a whole closure masquerading in just a function pointer.

tsion:
@drbo We've definitely been considering that.

For a bit of context, there are two versions of Miri we'll have in the future. If the Rust teams decide to use a MIR-based constant evaluator, some version of Miri will get incorporated into rustc, and that version will probably never use libffi, because you probably want to ban arbitrary FFI calls in constant evaluation.

But it would be a lot of fun to try adding libffi to this external version of Miri. I would probably wait until after any Miri → rustc merging happens, though.

BatmanAoD:
Now that Miri is indeed in rustc (wooooo!), any further thoughts on this?

If I understand correctly, FFI would be necessary for us to use Miri as the basis for a Jupyter kernel.

eddyb:
@BatmanAoD I'd advise against using miri for such a purpose - have you looked at webassembly?

BatmanAoD:
@eddyb I don't actually know what the requirements are for a Jupyter kernel; I was basing my understanding on [an old discussion](https://github.com/murarth/rusti/issues/43) about using `rusti` instead of Miri for the basis of the kernel.

I can't actually find documentation on how the kernel API works, though (I probably just haven't spent enough time looking). It may be that communication between JupyterLab and the kernel is entirely over HTTP.

In any case: I'm not sure how WebAssembly would solve the problem. The Jupyter kernel is a REPL for a given language; the goal of a Rust kernel would be to permit people to create Jupyter notebooks that would use the Rust language instead of IPython (or any of the other languages that are [already supported via kernels](https://github.com/jupyter/jupyter/wiki/Jupyter-kernels)). So the need is for a Rust REPL that supports whatever interface JupyterLab uses.

BatmanAoD:
@eddyb Yeah, I think the terminology is a bit confusing!

oli-obk:
[...]

PR #4326, opened 2025-05-17
(more) precisely track memory accesses and allocations across FFI

nia-e:
This greatly expands the information we have access to when code is executed across the FFI bound, including tracking exact allocations/deallocations (where possible) and accesses to Miri memory exposed to foreign code.

The current architecture forks Miri into 2 processes at the point of the first FFI call, with the parent `ptrace`ing the child. Communication is done via IPC channels; the child process signals when to start or stop tracing (during FFI), and the parent returns information about memory accesses that occurred, if any.

Alternatives considered:
- FUSE: It would be possible to obtain most of the same information about accesses to *our* memory (but not allocations done on C's side) by registering a FUSE fs which lives entirely in memory and `mmap`ing files from there to act as the machine's memory. The main benefits of this approach would be that it would be more portable as we'd no longer need to disassemble the foreign code to examine it (though we will still need to read arbitrary registers on the fly) and that it would give us slightly more accurate information about the sizes and types of memory accesses where the current approach lightly conservatively overestimates. However, I decided not to do this because it requires FUSE utils to actually be installed on the host system and would add a lot more complexity to the logic
- Plain `sigaction`s: We can also get data by registering `sigactions` to trigger when foreign code accesses our memory, but even though this is much simpler it would only give us page-grained information about what was performed (where the current approach is more or less byte-grained)

As it stands, this PR is not ready for merging pending *at least* the following being resolved. Still, I am opening this because significant sections of the code are ready for (preliminary, this is still very WIP) review. In ~decreasing order of severity as I see it:
- [x] Actually incorporate the obtained information into Miri's memory tracking instead of just printing to stderr
- [x] General cleanup such as removing hardcoded unwraps, adding more documentation, etc
- [x] Dynamically fall back to the old behaviour without panicking if the parent fails to trace properly, e.g. if `ptrace` is disabled
- [x] Add more tests
- [x] Add logic for non-x86 architectures
- [ ] Handle the foreign code spawning multiple threads

Open questions / fixable current limitations:
- [x] Do we want to pull in an existing allocator instead of having our own for managing the machine memory? (No; we merged our own)
- [x] Can this be made to be combineable with a multi-seeded Miri run? (Yes!)
- [ ] What happens if the FFI code forks further (do we even allow it)?
- [x] Should we switch out `iced_x86` with something else (e.g. go ahead with the FUSE idea, or use Capstone/llvm-sys for the better compatibility and eat the binary size & compile time increase)? (Switched to Capstone)

Zulip thread: https://rust-lang.zulipchat.com/#narrow/channel/269128-miri/topic/Initializing.20Miri.20memory.20from.20C.20FFI/with/518542321

Works toward (I hesitate to say closes) rust-lang/miri#11 

edit: relevant PRs so far working towards this:
- rust-lang/miri#4401
- rust-lang/miri#4405
- rust-lang/miri#4435
- rust-lang/miri#4456
- rust-lang/miri#4622

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -122,6 +122,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "capstone"
+version = "0.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "015ef5d5ca1743e3f94af9509ba6bd2886523cfee46e48d15c2ef5216fd4ac9a"
+dependencies = [
+ "capstone-sys",
+ "libc",
+]
+
+[[package]]
+name = "capstone-sys"
+version = "0.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2267cb8d16a1e4197863ec4284ffd1aec26fe7e57c58af46b02590a0235809a0"
+dependencies = [
+ "cc",
+ "libc",
+]
+
+[[package]]
 name = "cargo-platform"
 version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -591,6 +611,7 @@ version = "0.1.0"
 dependencies = [
  "aes",
  "bitflags",
+ "capstone",
  "chrono",
  "chrono-tz",
  "colored",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ libloading = "0.8"
 nix = { version = "0.30.1", features = ["mman", "ptrace", "signal"] }
 ipc-channel = "0.19.0"
 serde = { version = "1.0.219", features = ["derive"] }
+capstone = "0.13"
 
 [dev-dependencies]
 ui_test = "0.29.1"
--- a/src/alloc/isolated_alloc.rs
+++ b/src/alloc/isolated_alloc.rs
@@ -1,5 +1,6 @@
 use std::alloc::{self, Layout};
 
+use nix::sys::mman;
 use rustc_index::bit_set::DenseBitSet;
 
 /// How many bytes of memory each bit in the bitset represents.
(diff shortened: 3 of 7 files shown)


RalfJung:
Write-up of design options and choices: https://hackmd.io/@scZ140SZTaOb1VASX_ju-Q/SJUzIeo-eg

nia-e:
This doesn't build or pass tests right now since it builds against my fork of rustc, but against it it passes all tests on x86 linux and (assuming I didn't mess them up) on other platforms as well!

RalfJung:
@rustbot author
based on discussion on Zulip re: splitting up the PR

nia-e:
Now that the isolated allocator bit is done (hooray!) I can get back to focusing on this. I assume it'll be a while until it gets merged in any capacity, but still :D There's still some work to be done, once it has been done I'll clean up the code and mark it ready for review.

I recall you asking that I split out the malloc/mmap-tracing bits for a separate later PR, I assume that still holds? The code size of each option is about the same (though perhaps the malloc-tracing will require fewer/no changes to rustc, I'll have to see).

Also, unfortunately macOS has a very janky & incomplete implementation of `ptrace`, so while not impossible (`lldb` et al. work on it) it would require a different approach at least partially. Judging by their docs the BSDs should be fine, but there's always the chance of subtly different behaviour so only linux is enabled for now.

RalfJung:
> I recall you asking that I split out the malloc/mmap-tracing bits for a separate later PR, I assume that still holds?

Yes, the first PR should focus on the read/write tracking.

> Also, unfortunately macOS has a very janky & incomplete implementation of ptrace, so while not impossible (lldb et al. work on it) it would require a different approach at least partially. Judging by their docs the BSDs should be fine, but there's always the chance of subtly different behaviour so only linux is enabled for now.

It's fine if the read/write tracking only works on Linux. Basic native-libs mode should keep working on macOS though.

nia-e:
I think this is ready for at least a preliminary review? The rustc side I'm pretty settled on, though, so I'll push and mark that as ready also. There's every chance I missed something really obvious and a second pair of eyes probably helps.

@rustbot ready

src/bin/miri.rs
```diff
@@ -192,6 +192,13 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
             todo!("GenMC mode not yet implemented");
         };
 
+        #[cfg(target_os = "linux")]
+        if !config.native_lib.is_empty() && !config.force_old_native_lib {
+            // FIXME: This should display a diagnostic / warning on error
+            // SAFETY: No other threads have spawned yet
```
RalfJung: This comment is almost certainly wrong. rustc can spawn an arbitrary amount of threads before calling us here.

The forking should happen way earlier, probably right after argument parsing but before we invoke rustc_driver.

no_alloc.json
RalfJung, line 1: This file probably should not be here. ;)

RalfJung (commented):
I started idly scrolling through this, but then realized this is way too much for me at the moment.^^ It is unlikely I'll be able to review a PR of this size and complexity before mid-July. I expect I have to book at least half a day for this, and I don't have that much time to spare earlier than that. If there's some way we could arrange real-time communication (e.g. via zoom) for a quicker feedback cycle, that could be useful.

Is there any way to split this up? Like
- first land some basic ptrace setup that does basically nothing
- then add catching of memory events in the server (but just logging the info to stderr or so without communicating them to Miri proper)
- then add a communication channel so that core Miri learns about these events and uses them for something useful

Also, the IPC protocol between server and client should be explicitly documented somewhere: what messages are there, what is their format, what do they mean, when can which message be sent by whom and what is the reply. Since it seems there are three channels, such docs are needed for each of them. Probably the file that defines the message types it the best place for that. Currently that's just the crate root of `trace` I think; this should be a proper module on its own.

nia-e:
I'd be more than happy to do that, yeah! I've also been quite busy the last couple of weeks and so haven't had much time to address reviews, but I'm okay with talking directly and splitting this up further. Let me know what and when works for you, I've had a bit of trouble with Zoom on my device in the past so I have a slight preference for Jitsi but it's not a huge deal. I'm likely to be free again and address review comments + be able to talk sometime early next week

no_alloc.json
nia-e: gulp

nia-e:
@rustbot author
based on discussion above re: splitting this up further

RalfJung:
Okay, sounds good! Let's see if it will be needed with the split-up approach and Oli helping with reviews.

RalfJung:
@nia-e I am going to close this PR since it's become inactive and work has moved to smaller follow-up PRs. Maybe it'd make sense to open an issue summarizing the plan and the current status?

PR #4329, opened 2025-05-18
enable clippy::as_conversions to fully rule out as-casts

RalfJung:


RalfJung:
@rust-lang/miri what do you think, should we do this?

src/shims/windows/fs.rs
```diff
@@ -401,7 +401,7 @@ fn extract_windows_epoch<'tcx>(
         Some(time) => {
             let duration = ecx.system_time_since_windows_epoch(&time)?;
             let duration_ticks = ecx.windows_ticks_for(duration)?;
-            #[allow(clippy::cast_possible_truncation)]
+            #[allow(clippy::as_conversions)]
```
saethlin, line 404: Should this be an expect?

saethlin:
Seems like forward progress to me :+1: 

Still not ideal in a few places as you pointed out in the diff, but better.

PR #4333, opened 2025-05-19
run tests on mips-unknown-linux-gnu

RalfJung:
That target has two unique properties we are not capturing yet:
- It is a 32bit big-endian target
- It does not have 64bit atomics (so std uses fallback paths for some logic)


ci/ci.sh
RalfJung, line 1: I also added MANY_SEEDS=16 so we have some coverage of those tests on our better-supported tier 2 targets. Let's see what that does to CI times... the macOS runner often is the first runner to finish, but sometimes it is the last to finish.

PR #4338, opened 2025-05-21
FileDescription: improve read/write docs

RalfJung:
@CraftSpider is this clear enough, or should we go into more detail?


CraftSpider (approved):
Looks much clearer to me, thanks

PR #4342, opened 2025-05-22
implement shim_sig macro

geetanshjuneja:
Implements `shim_sig` macro.
More details here - https://github.com/rust-lang/miri/issues/3842#issuecomment-2811896286


src/shims/unix/foreign_items.rs
```diff
@@ -112,26 +112,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                    abi,
-                    ExternAbi::C { unwind: false },
-                    [this.machine.layouts.const_raw_ptr.ty],
-                    this.machine.layouts.mut_raw_ptr.ty,
-                    args,
-                )?;
+                let (callee_abi, arg_types, ret_type) =
+                    shim_sig!(this, extern "C" fn((*const _)) -> (*const _));
+                let [name] =
+                    this.check_shim_abi(link_name, abi, callee_abi, arg_types, ret_type, args)?;
```
CraftSpider, line 118: Outsider question: How many places is `check_shim_abi` used so far? Could abi, args, and return get packaged into one argument so you could just use the macro directly in the parameters?

src/helpers.rs
```diff
@@ -1435,3 +1435,36 @@ impl ToU64 for usize {
...
+            std::str::FromStr::from_str($abi).expect("incorrect abi specified"),
+            [$(layout!($this, $arg)),*],
+            layout!($this, $ret)
+        )
+    };
+
+    // default Rust abi
+    ($this:ident, extern $abi:literal fn($($arg:tt),*) -> $ret:tt) => {
+        shim_sig!($this, extern "Rust" fn($($arg),*) -> $ret)
+    };
```
RalfJung, line 1452: Having to use `:tt` instead of `:ty` here is unfortunate.  I guess that's because `*const _` is not a valid type? Maybe we can use `*const T` instead?

RalfJung (commented):
@rustbot author

src/helpers.rs
geetanshjuneja:
```rust
shim_sig!(this, extern "C" fn() -> "pid_t")
````
"pid_t" wo

Comment (RalfJung) [reply src/helpers.rs:1452]: Why not? That's a valid type, syntactically.

Comment (geetanshjuneja) [reply src/helpers.rs:1452]: > Oh wait you put it in string quotes? Why that? It should be
> 
> ```rust
> shim_sig!(this, extern "C" fn() -> pid_t)
> ```
> 
> Or maybe we use
> 
> ```rust
> shim_sig!(this, extern "C" fn() -> libc::pid_t)
> ```
> 
> t make it more clear.

thx

Comment (geetanshjuneja) [reply src/helpers.rs:1452]:
```rust
macro_rules! layout_and_ty {
    ($this:ident, *const T) => {
        $this.machine.layouts.const_raw_ptr.ty
    };
    ($this:ident, *mut T) => {
        $this.machine.layouts.mut_raw_ptr.ty
    };
    ($this:ident, i32) => {
        $this.tcx.types.i32
    };
    ($this:ident, usize) => {
        $this.tcx.types.usize
    };
    ($this:ident, isize) => {
        $this.tcx.types.isize
    };
    ($this:ident, $x:ty) => {
        $this.libc_ty_layout(stringify!($x)).ty
    };
}
```

When I call this macro `layout_and_ty($this, i32)` in `shim_sig` macro, it matches with `($this:ident, $x:ty) ` not `($this:ident, i32) `

geetanshjuneja: In the macro arm of `layout_and_ty`, is it taking `i32` as a keyword not a rust type?

geetanshjuneja: > `i32` is not a keyword though...

I think the issue is that when I pass already matched `ty` token in `shim_sig` to `layout_and_ty`, then I will only be matching `$x:ty` arm. But if would have directly called `layout_and_ty` macro then I would have been able to match `i32` arm. See [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=c7ac20037edd62d8a7ee8fe695421907). 
Am I correct here?
So I think I have do `if-else` inside $x:ty$ arm and return the correct layout.

RalfJung: Hm, I don't know if there is a trick to "degenerate" a `ty` token back to a series of `tt`. Maybe ask on Zulip? I'm not a macro expert unfortunately.

geetanshjuneja:
@RalfJung I have created the topic [#general > Convert ty token to series of tt token](https://rust-lang.zulipchat.com/#narrow/channel/122651-general/topic/Convert.20ty.20token.20to.20series.20of.20tt.20token/with/521240018) on zulip. Can you tag someone who can help with this?

RalfJung:
If I knew who could help I'd have already done that.^^

RalfJung:
I looked into this a bit myself and I agree that string matching is the best we can do. So based on that I implemented a version of this macro with the API I was imagining: https://github.com/rust-lang/miri/pull/4497.

I'll close this PR as it is superseded by the #4497. Thanks for your experiments in this direction! If you want to make more PRs once #4497 lands to convert more code to the new check function, then your help will definitely be appreciated. :)

PR #4343, opened 2025-05-22
Isolate MiriMachine memory from Miri's

nia-e:
Based on discussion surrounding rust-lang/miri#4326, this merges in the (very simple) discrete allocator that the MiriMachine will have. See the design document linked there there for considerations, but in brief: we could pull in an off the shelf allocator for this, but performance isn't a massive worry and doing it this way might make it easier to enable support for doing multi-seeded runs in the future (without a lot more extraneous plumbing, at least)


Cargo.toml
```diff
@@ -39,6 +39,7 @@ features = ['unprefixed_malloc_on_supported_platforms']
 libc = "0.2"
 libffi = "4.0.0"
 libloading = "0.8"
+nix = { version = "0.30.1", features = ["mman", "ptrace", "signal"] }
```
RalfJung, line 42: We don't need all these features for this PR, do we?

src/discrete_alloc.rs
RalfJung, line 1: Can you move this and `alloc_bytes` into a new `alloc` folder so we avoid growing the number of files/dirs in the root? This is really an `alloc_bytes` implementation detail, right?

src/discrete_alloc.rs
```diff
@@ -0,0 +1,271 @@
...
+use std::sync;
+
+use nix::sys::mman::ProtFlags;
+
+use crate::helpers::ToU64;
+
+static ALLOCATOR: sync::Mutex<MachineAlloc> = sync::Mutex::new(MachineAlloc::empty());
+
+/// A distinct allocator for the `MiriMachine`, allowing us to manage its
+/// memory separately from that of Miri itself.
```
RalfJung, line 11:
```suggestion
/// A distinct allocator for interpreter memory contents, allowing us to manage its
/// memory separately from that of Miri itself. This is very useful for native-lib mode.
```

src/discrete_alloc.rs
```diff
@@ -0,0 +1,271 @@
...
+#[derive(Debug)]
+pub struct MachineAlloc {
+    pages: Vec<*mut u8>,
+    huge_allocs: Vec<(*mut u8, usize)>,
+    allocated: Vec<Box<[u8]>>,
+    page_size: usize,
+    enabled: bool,
```
RalfJung, line 18: Some of these fields look like they have interesting invariants attached to them. Surely this is not a `Vec` of arbitrary raw pointers. Please document all those invariants here in field-level doc comments.

src/discrete_alloc.rs
```diff
@@ -0,0 +1,271 @@
...
+}
+
+// SAFETY: We only point to heap-allocated data
+unsafe impl Send for MachineAlloc {}
+
+impl MachineAlloc {
+    // Allocation-related methods
+
+    /// Initializes the allocator with placeholder 4k pages.
```
RalfJung, line 27: I don't understand the comment. This doesn't seem to actually put any placeholder pages into the allocator?

src/discrete_alloc.rs
```diff
@@ -0,0 +1,271 @@
...
+    pub unsafe fn enable() {
+        let mut alloc = ALLOCATOR.lock().unwrap();
+        alloc.enabled = true;
+        // This needs to specifically be the system pagesize!
+        alloc.page_size = unsafe {
+            let ret = libc::sysconf(libc::_SC_PAGE_SIZE);
+            if ret > 0 {
+                ret.try_into().unwrap()
+            } else {
+                4096 // fallback
```
RalfJung, line 48: Given that this apparently must be the page size, how is this fallback justified?

src/discrete_alloc.rs
```diff
@@ -0,0 +1,271 @@
...
+            }
+        }
+    }
+
+    /// Returns a vector of page addresses managed by the allocator.
+    #[expect(dead_code)]
+    pub fn pages() -> Vec<u64> {
```
RalfJung, line 55: Why is this `u64`, not `usize`? These are host pages after all, right?

Probably better to remove the function entirely given that it is dead code, so I cannot judge how it will be used.

src/discrete_alloc.rs
```diff
@@ -0,0 +1,271 @@
...
+        let alloc = ALLOCATOR.lock().unwrap();
+        alloc.pages.clone().into_iter().map(|p| p.addr().to_u64()).collect()
+    }
+
+    fn add_page(&mut self) {
+        let page_layout =
+            unsafe { Layout::from_size_align_unchecked(self.page_size, self.page_size) };
+        let page_ptr = unsafe { alloc::alloc(page_layout) };
+        if page_ptr.is_null() {
+            panic!("aligned_alloc failed!!!")
```
RalfJung, line 65: There's no `aligned_alloc` here? Also no need for triple-`!` ;)

src/discrete_alloc.rs
```diff
@@ -0,0 +1,271 @@
...
+        }
+        self.allocated.push(vec![0u8; self.page_size / 8].into_boxed_slice());
```
RalfJung, line 67: Uh so we allocate a second allocation here that's 1/8th the size...? This is where the lack of comments on the type itself that would explain the invariant starts to really become a problem.^^

src/machine.rs
```diff
@@ -739,6 +739,8 @@ impl<'tcx> MiriMachine<'tcx> {
                 // undefined behaviour in Miri itself!
                 (
                     unsafe {
+                        #[cfg(all(unix, any(target_arch = "x86", target_arch = "x86_64")))]
+                        discrete_alloc::MachineAlloc::enable();
```
RalfJung, line 743: First of all, the safety requirement of calling this only once is not satisfied.

Secondly, `AllocBytes` seems to *always* use that allocator. What happens when it has not been "enabled"...?

RalfJung (commented):
Thanks for the PR! I left some first comments, but this is not a full review. I'd rather not reverse-engineer the invariants of `MachineAlloc` myself, so I'll wait for you to document them, which will make review a lot easier.

Furthermore, all `pub fn` in `discrete_alloc` should have proper doc comments, not just a safety comment. Please also add some basic unit tests -- we don't use them much in Miri, but this is one of the cases where they would make sense.

On Zulip you mentioned some benchmarks. Can you put benchmark results for the variant that you ended up going for here into the PR?

Cargo.toml
nia-e: Correct, oops. I'll change that

src/discrete_alloc.rs
nia-e: These all are used only for integrating with the ptrace stuff, but I can remove them for now, sure. The reason it was `u64`s is so that we don't need to iteratively call `to_u64()` in the ptrace code on what this returns, as everything there uses `u64`s internally to be the same as the rest of Miri

src/discrete_alloc.rs
RalfJung: The easiest thing would be to just panic instead. Or are there real situations where we cannot determine the page size?

src/discrete_alloc.rs
RalfJung: Generally in Miri we should use `u64` for target addresses and `usize` for host addresses.

But with native-libs mode, target==host, so... I guess that blurs the line.

Anyway let's discuss that in the ptrace PR; please remove these functions for now.

src/discrete_alloc.rs
RalfJung: Using `alloc::alloc` is the right thing to use. Just the panic message needs fixing :)

src/machine.rs
nia-e: Maybe I misphrased a comment someplace - it's fine for it to be called multiple times as long as it's called before any `MiriAllocBytes` have been instantiated. If it hasn't been enabled, `MachineAlloc::alloc()` and others just act as wrappers for `alloc::alloc()` and such

src/discrete_alloc.rs
nia-e: I guess I was just paranoid of extra unwraps, but you're probably right. I'll change this

src/machine.rs
RalfJung: > If it hasn't been enabled, MachineAlloc::alloc() and others just act as wrappers for alloc::alloc() and such

That definitely should be explained in the top-level `MachineAlloc` comment, and then also in the `enabled()` doc comment.

RalfJung: Also, the reasoning is still wrong: what if I create two miri instances with different settings in the same process, some with and some without native-libs mode? We don't currently do that, but the unsafe code you write has to be resilient for arbitrary uses of the safe public API of the Miri crate.

RalfJung: Let's take a step back and figure out what we are designing for. All this happens because `AllocBytes` can't access the machine. Maybe there's some way we can give it access to something though? That will require rustc changes, but I think it shouldn't be too hard:
- change `trait AllocBytes` to `trait AllocBytes<M>` and taking an `&M` parameter in the constructor
- the default impl becomes `impl<M> AllocBytes<M> for Box<[u8]>`, doesn't care about the machine
- the `Decodable` instance should not be needed for all `AllocBytes`, only for `Box<[u8]>`
- the big `Machine` trait requires `AllocBytes<&Self>`
- `MiriAllocBytes` implements `AllocBytes<&MiriMachine>` and can hence access its configuration
- the alloc functions in `memory.rs` plumb a machine reference through to the constructors

Then we don't need any global state at all, we can just keep the allocator state in the `MiriMachine`.

With this, we probably want the logic for deciding which allocator to use to be in `MiriAllocBytes`. The allocator could then be called `IsolatedPageAllocator` or so, reflecting its purpose and the fact that it is not always the Miri allocator.

src/discrete_alloc.rs
nia-e: Oh oops I realise what happened, this used to call `libc::aligned_alloc` and I never updated the comment. Oops!

src/machine.rs
nia-e: Oh, that's perfect then! I didn't think it worth it to plumb through another generic parameter on the rustc side but this actually solves a lot of problems :D I'll structure it that way and open a PR on rustc

RalfJung: Let's hope I was not too optimistic when I said this should be doable :)

nia-e: Well I have time to kill, I'll at least give it a shot. I'll push my changes as-is for now and then start working on that so you can see the option I found without touching rustc at all

RalfJung: Yeah that also works, though it will put allocations of both machines into the "discrete" pool in the scenario I described above.

nia-e:
I'll post benchmarks in a bit! I realised there might be some speed gains to be made with very simple changes, so I'll just experiment a little first. Thanks for the comments ^^

src/machine.rs
RalfJung: > he machine trait is pretty far from where AllocBytes is described and we'd have to specifically bound the generic impl as AllocBytes<M: Machine> if we wanted it to be usable here. 

No. My design specifically avoids that. Just make it an arbitrary `M`.

The impls either don't care (the `Box<[u8]>` one) so they can leave it generic, or they want to be an impl for a particular machine anyway and have that machine in scope (the Miri one).

nia-e:
Baseline is set to having the allocator fully disabled. It's only marginally slower in most cases, though it struggles with large allocations it seems. I wonder how much work it would be to improve that, but if we go down the "only machines using this will touch it" path I hope it's not too bad? I got a slight (~4%) improvement from calling `mmap` directly instead of calling `alloc::alloc()` but I doubt that's worth it. Is Miri built with optimisations on by default when invoking `./miri bench`?
```
Comparison with baseline (relative speed, lower is better for the new results):
  backtraces: 1.36 ± 0.08
  big-allocs: 32.71 ± 1.60
  mse: 1.05 ± 0.10
  range-iteration: 1.06 ± 0.03
  serde1: 1.00 ± 0.03
  serde2: 1.06 ± 0.03
  slice-chunked: 1.03 ± 0.04
  slice-get-unchecked: 0.93 ± 0.08
  string-replace: 1.06 ± 0.04
  unicode: 1.06 ± 0.05
  zip-equal: 1.08 ± 0.02
```

RalfJung:
> Is Miri built with optimisations on by default when invoking ./miri bench?

Yes.

A 32x slowdown with big allocations is hefty.^^ Shouldn't those just forward to `alloc::alloc` anyway as part of the huge allocs treatment? Without looking at the code, what I'd assume happens when the allocation request is close to the page size is: round up to multiple of page size, and then just allocate that and use it directly without any kind of tracking of "which parts of this page are used" or so. That should have basically no overhead.

nia-e:
I checked; seems like the `big_allocs` test specifically calls `alloc_zeroed` which was the reason for the slowdown (apparently `ptr.write_bytes()` is a *really* slow way to zero out bytes). I changed up the logic a bit to make the logic generic over calling `alloc::alloc()` vs `alloc::alloc_zeroed()` and these are the (much better!) results:
```
Comparison with baseline (relative speed, lower is better for the new results):
  backtraces: 1.23 ± 0.06
  big-allocs: 1.03 ± 0.05
  mse: 0.99 ± 0.09
  range-iteration: 1.01 ± 0.03
  serde1: 0.99 ± 0.02
  serde2: 1.03 ± 0.02
  slice-chunked: 0.97 ± 0.04
  slice-get-unchecked: 0.89 ± 0.08
  string-replace: 1.00 ± 0.03
  unicode: 1.01 ± 0.04
  zip-equal: 1.05 ± 0.04
```
I might be able to squeeze a bit more perf out by actually making the functions generic instead of just passing in a function pointer but shrug, unsure if it's necessary

RalfJung:
> What kind of unit tests do you think belong here? I assumed functionality is covered by the usual tests, but I'll happily add in some stuff if you think it's relevant

Similar to `range_map`: just add some functions with `#[test]` in the file with the allocator implementation to test some particular corner cases -- or at least a basic smoke test if there are no corner cases.

nia-e:
Openen [the PR on the main repo](https://github.com/rust-lang/miri/pull/4343), I'll get to adding tests if everything there is okay

nia-e:
Expecting the build to fail for now since it's adapted to the changes from the PR (but also Miri seems to be having trouble on the current upstream master commit, so I guess it's pending that being fixed too)

nia-e:
Tests added :D let me know if there's anything more to do

src/alloc/isolated_alloc.rs
```diff
@@ -0,0 +1,332 @@
+use std::alloc::{self, Layout};
+use std::sync;
+
+static ALLOCATOR: sync::Mutex<IsolatedAlloc> = sync::Mutex::new(IsolatedAlloc::empty());
```
RalfJung, line 4: I was hoping we wouldn't need this any more and could instead store this in the `Machine` and pass it to the allocation via the `params`... otherwise, the allocations of multiple native-libs machines could still get mixed here.

But that's harder now that parameters are their own associated types. It should still be possible though, AFAIK lifetime generic associated types are stable?

src/alloc/isolated_alloc.rs
```diff
@@ -0,0 +1,332 @@
...
+        Self { page_ptrs: Vec::new(), huge_ptrs: Vec::new(), page_infos: Vec::new(), page_size: 0 }
+    }
+
+    /// Expands the available memory pool by adding one page.
+    fn add_page(&mut self) -> (*mut u8, &mut Box<[u8]>) {
+        assert_ne!(self.page_size, 0);
+
+        let page_layout =
+            unsafe { Layout::from_size_align_unchecked(self.page_size, self.page_size) };
+        // We don't overwrite the bytes we hand out so make sure they're zeroed by default!
```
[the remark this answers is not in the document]
nia-e: Leftover from an earlier thing, oops!

src/alloc/isolated_alloc.rs
```diff
@@ -0,0 +1,332 @@
...
+    /// SAFETY: `page` must be a page-aligned pointer to an allocated page,
+    /// where the allocation is (at least) `page_size` bytes.
+    unsafe fn alloc_from_page(
+        page_size: usize,
+        layout: Layout,
+        page: *mut u8,
+        pinfo: &mut Box<[u8]>,
+        zeroed: bool,
+    ) -> Option<*mut u8> {
+        let (size, align) = IsolatedAlloc::normalized_layout(layout);
```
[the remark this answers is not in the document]
RalfJung: If that's the representation then it should probably be a `Vec<bool>`.

But, rustc also has a bunch of bitset types, one of which is surely useful here?
https://doc.rust-lang.org/nightly/nightly-rustc/rustc_index/bit_set/index.html
`DenseBitSet` sounds like pretty much what you need.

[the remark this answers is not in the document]
nia-e: Oh neat! Then I can use that :D ty

src/alloc/isolated_alloc.rs
nia-e: Ended up going with a hashmap, but I can change that if you think it's not the right call. I just couldn't see a non-intrusive way to do it outside of either (a) passing a raw pointer instead of a reference (yikes) or (b) using something like `Arc<RefCell<IsolatedAllocator>>` which I'm not sure is worth the overhead and runtime panic risk

nia-e:
Ah seems like it never got passed back to waiting on author so in case it didn't ping you, cc @RalfJung

src/alloc/isolated_alloc.rs
nia-e: I think I found a way to make the Rc-only thing work so I'll do that! + it might fix other stuff that's currently bothering me, lol

RalfJung: Ah I guess now I nerd-sniped you into making the many-seeds thing works for native-libs mode... sorry for that.^^

Just for the allocator, you don't need an ID -- you can entirely avoid even having a global variable here and just store another `Rc<...>` pointing to the allocator in the machine itself. That would only hold the allocations for that one machine so it can easily list "its" pages (and it also neatly involves having concurrent machines content on the same lock). But I don't know which further constraints are imposed by the ptrace thing.

RalfJung:
Based on https://github.com/rust-lang/miri/pull/4343#discussion_r2111274513 I assume this is @rustbot author again.

src/alloc/isolated_alloc.rs
nia-e: Sorry, I mean many-seeds can coexist but only one can be in "ffi mode" at a time, with the others blocking until that one finishes. Otherwise it becomes borderline impossible to trace properly, and perhaps actually impossible when it comes to e.g. intercepting syscalls

src/alloc/alloc_bytes.rs
```diff
@@ -80,22 +103,32 @@ impl MiriAllocBytes {
...
-    fn from_bytes<'a>(slice: impl Into<Cow<'a, [u8]>>, align: Align, _params: ()) -> Self {
+    fn from_bytes<'a>(
+        slice: impl Into<Cow<'a, [u8]>>,
+        align: Align,
+        params: MiriAllocParams,
+    ) -> Self {
         let slice = slice.into();
         let size = slice.len();
         let align = align.bytes();
+        let p_clone = params.clone();
```
RalfJung, line 122: Why is this getting cloned here? That should not be needed.

src/alloc/alloc_bytes.rs
```diff
@@ -105,12 +138,19 @@ impl AllocBytes for MiriAllocBytes {
         alloc_bytes
     }
 
-    fn zeroed(size: Size, align: Align, _params: ()) -> Option<Self> {
+    fn zeroed(size: Size, align: Align, params: MiriAllocParams) -> Option<Self> {
         let size = size.bytes();
         let align = align.bytes();
+        let p_clone = params.clone();
```
RalfJung, line 144: Same here regarding cloning.

src/concurrency/thread.rs
```diff
@@ -905,7 +906,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     interp_ok(MiriAllocBytes::from_bytes(
                         std::borrow::Cow::Borrowed(bytes),
                         align,
-                        (),
+                        params.clone(),
```
RalfJung, line 909: Ah damn, this closure should be a `FnOnce` and not `FnMut` and then we can avoid a clone here...
Could you make a rustc PR for that?

src/machine.rs
```diff
@@ -638,7 +642,8 @@ impl<'tcx> MiriMachine<'tcx> {
             let path = Path::new(out).join(filename);
             measureme::Profiler::new(path).expect("Couldn't create `measureme` profiler")
         });
-        let rng = StdRng::seed_from_u64(config.seed.unwrap_or(0));
+        let seed = config.seed.unwrap_or(0);
+        let rng = StdRng::seed_from_u64(seed);
```
RalfJung, line 646: Why did you change this?

src/machine.rs
```diff
@@ -532,6 +532,10 @@ pub struct MiriMachine<'tcx> {
     /// Needs to be queried by ptr_to_int, hence needs interior mutability.
     pub(crate) rng: RefCell<StdRng>,
 
+    /// The allocator used for the machine's `AllocBytes`.
```
RalfJung, line 535:
```suggestion
    /// The allocator used for the machine's `AllocBytes` in native-libs mode.
```

src/machine.rs
```diff
@@ -532,6 +532,10 @@ pub struct MiriMachine<'tcx> {
...
+    #[cfg(target_os = "linux")]
+    pub(crate) allocator: Rc<RefCell<crate::alloc::isolated_alloc::IsolatedAlloc>>,
```
RalfJung, line 537: This should be `Option<Rc<...>>` since it should only exist in native-libs mode.

RalfJung (commented):
I didnt look into the actual allocator for this, just the glue code around it.

src/alloc/isolated_alloc.rs
```diff
@@ -0,0 +1,306 @@
...
+            .position(|pg| ptr.addr() == pg.0.addr())
+            .expect("Freeing unallocated pages");
+        let ptr = self.huge_ptrs.remove(idx).0;
+        unsafe {
+            alloc::dealloc(ptr, layout);
+        }
+    }
+}
+/*
+#[cfg(test)]
```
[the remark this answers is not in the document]
nia-e: Had to comment them out when I refactored everything, I'll reenable them as I address the review!

src/alloc/isolated_alloc.rs
```diff
@@ -0,0 +1,306 @@
...
+        }
+
+        let mut free = vec![];
+        let page_layout =
+            unsafe { Layout::from_size_align_unchecked(self.page_size, self.page_size) };
+        for (idx, pinfo) in self.page_infos.iter().enumerate() {
+            if pinfo.is_empty() {
+                free.push(idx);
+            }
+        }
```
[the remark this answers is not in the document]
nia-e: Fair enough, I think at one point I considered allowing cross-page allocs and this is one of the "primordial" bits of code from the first draft. I'll change it up!

nia-e:
Hopefully that's better?

src/alloc/isolated_alloc.rs
```diff
@@ -39,38 +42,57 @@ impl IsolatedAlloc {
...
 
+    /// If the allocation is greater than a page, then round to the nearest page #.
+    /// Since we pass this into the global allocator, it's more useful to return
+    /// a `Layout` instead of a pair of usizes.
+    #[inline]
+    fn huge_normalized_layout(layout: Layout, page_size: usize) -> Layout {
+        // Allocate in page-sized chunks
+        let size = layout.size().next_multiple_of(page_size);
+        // Align probably *shouldn't* ever be greater than the page size,
+        // but just in case
```
RalfJung, line 72: It's completely legal to ask for a 16k align on a 4k page size system.

nia-e: Fair enough, lol

src/alloc/isolated_alloc.rs
```diff
@@ -0,0 +1,356 @@
...
+}
+
+impl IsolatedAlloc {
+    /// Creates an empty allocator.
+    pub fn new() -> Self {
+        Self {
+            page_ptrs: Vec::new(),
+            huge_ptrs: Vec::new(),
+            page_infos: Vec::new(),
+            page_size: unsafe { libc::sysconf(libc::_SC_PAGESIZE).try_into().unwrap() },
```
RalfJung, line 39: Please add a safety comment for all `unsafe` blocks

src/alloc/isolated_alloc.rs
```diff
@@ -0,0 +1,356 @@
...
+        }
+    }
+
+    /// Expands the available memory pool by adding one page.
+    fn add_page(&mut self, page_size: usize) -> (*mut u8, &mut DenseBitSet<usize>) {
+        let page_layout = unsafe { Layout::from_size_align_unchecked(page_size, page_size) };
+        let page_ptr = unsafe { alloc::alloc(page_layout) };
+        // `page_infos` has to have one-eighth as many bits as a page has bytes
+        // (or one-64th as many bytes)
```
RalfJung, line 48:
```suggestion
        // `page_infos` has to have one bit for each `COMPRESSION_FACTOR`-sized chunk of bytes in the page.
        assert!(page_size % COMPRESSION_FACTOR == 0);
```

src/alloc/isolated_alloc.rs
```diff
@@ -0,0 +1,356 @@
...
+        Self {
+            page_ptrs: Vec::new(),
+            huge_ptrs: Vec::new(),
+            page_infos: Vec::new(),
+            page_size: unsafe { libc::sysconf(libc::_SC_PAGESIZE).try_into().unwrap() },
+        }
+    }
+
+    /// Expands the available memory pool by adding one page.
+    fn add_page(&mut self, page_size: usize) -> (*mut u8, &mut DenseBitSet<usize>) {
```
RalfJung, line 44: Why does this take a page size argument when it could use `self.page_size`?

src/alloc/isolated_alloc.rs
```diff
@@ -0,0 +1,356 @@
...
+            let pinfo = std::iter::zip(&mut self.page_ptrs, &mut self.page_infos)
+                .find(|(page, _)| page.addr() == page_addr);
+            let Some((_, pinfo)) = pinfo else {
+                panic!(
+                    "Freeing in an unallocated page: {ptr:?}\nHolding pages {:?}",
+                    self.page_ptrs
+                )
+            };
+            // Since each bit of the bitset represents COMPRESSION_FACTOR bytes,
+            // we need to divide by that when getting the indices
```
RalfJung, line 206: The most important part is to explain what it is you are doing, not how you are doing it.
```suggestion
            // Mark this range as available in the page.
```

src/alloc/isolated_alloc.rs
```diff
@@ -0,0 +1,356 @@
...
+            for idx in ptr_idx_pinfo..ptr_idx_pinfo + size_pinfo {
+                pinfo.remove(idx);
+            }
+
+            // We allocated all the pages with this layout
+            let page_layout =
+                unsafe { Layout::from_size_align_unchecked(self.page_size, self.page_size) };
+            // Only 1 page might have been freed after a dealloc, so if it exists,
+            // find it and free it (and adjust the vectors)
+            if let Some(free_idx) = self.page_infos.iter().position(|pinfo| pinfo.is_empty()) {
```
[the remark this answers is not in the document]
nia-e: 🤦🏾

src/alloc/isolated_alloc.rs
```diff
@@ -183,6 +203,8 @@ impl IsolatedAlloc {
         let (size, align) = IsolatedAlloc::normalized_layout(layout);
 
         if IsolatedAlloc::is_huge_alloc(size, align, self.page_size) {
+            // SAFETY: Partly upheld by caller, and we checked that the size
+            // and align mean this must have been allocated via `alloc_huge`
```
RalfJung, line 207: "mean" -> "meaning"?

nia-e: It was intentional but I realise it sounds kinda weird, I'll edit it

src/concurrency/thread.rs
RalfJung: Once https://github.com/rust-lang/miri/pull/4355 lands you should be able to rebase & remove the `clone`.

nia-e:
Ok! So there's a small problem, namely that the jemalloc implementation rust uses by default is really slow with page-aligned allocations it seems; the reason perf was good before is that I forgot to actually enforce page-alignment in the `alloc_huge` calls. I checked; even passing in a constant `4096` as the alignment causes `big_allocs` to see a very significant slowdown depending on the run (seems quite highly variable - it was as low as 8.9x at one point, but I didn't save those results)

*However*, if we do the huge allocs by directly calling `mmap`, it drops back down to parity. Here's the old benchmark run:
```
Comparison with baseline (relative speed, lower is better for the new results):
  backtraces: 1.18 ± 0.05
  big-allocs: 22.30 ± 1.89
  mse: 1.01 ± 0.07
  range-iteration: 1.08 ± 0.02
  serde1: 1.06 ± 0.03
  serde2: 1.04 ± 0.04
  slice-chunked: 0.98 ± 0.04
  slice-get-unchecked: 1.03 ± 0.04
  string-replace: 1.00 ± 0.06
  unicode: 1.02 ± 0.04
  zip-equal: 1.01 ± 0.03
```
And this is what I got calling `mmap` directly (albeit this means the pointer will only ever be page aligned at most - unsure if relying on greater-than-page-alignment is even legal given that it's impossible to guarantee?)
```
Comparison with baseline (relative speed, lower is better for the new results):
  backtraces: 1.18 ± 0.05
  big-allocs: 1.07 ± 0.10
  mse: 1.00 ± 0.07
  range-iteration: 1.03 ± 0.02
  serde1: 1.02 ± 0.04
  serde2: 1.00 ± 0.04
  slice-chunked: 0.97 ± 0.05
  slice-get-unchecked: 0.97 ± 0.02
  string-replace: 0.96 ± 0.08
  unicode: 1.01 ± 0.04
  zip-equal: 1.00 ± 0.03
```

RalfJung:
> the jemalloc implementation rust uses by default is really slow with page-aligned allocations it seems; the reason perf was good before is that I forgot to actually enforce page-alignment in the alloc_huge calls.

What alignment did you set before?

> relying on greater-than-page-alignment is even legal given that it's impossible to guarantee

It's definitely legal, and it's also possible to guarantee. You "just" have to allocate more pages than needed and then round up the pointer you got to the needed alignment. jemalloc likely does that if you ask for a high alignment. It's non-trivial to implement this correctly though, in particular regarding deallocation, so I'd rather not have this in the codebase.

Given that the regression only affects native-libs mode, I am also fine with just taking the hit for now and having an issue to track the problem. We can then ping some people there that might know more about this; I am fairly clueless when it comes to allocators.

RalfJung:
Please squash, I'll take a look later :)

PR #4344, opened 2025-05-22
Implement file cloning on Windows

CraftSpider:
Just requires expanding what we can do with handles - get the current process pseudo-handle, and the Windows version of `dup`.

### DIFF
--- a/src/shims/windows/handle.rs
+++ b/src/shims/windows/handle.rs
@@ -9,6 +9,7 @@ use crate::*;
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub enum PseudoHandle {
     CurrentThread,
+    CurrentProcess,
 }
 
 /// Miri representation of a Windows `HANDLE`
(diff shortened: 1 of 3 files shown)


src/shims/windows/handle.rs
```diff
@@ -244,6 +248,76 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        if target_proc != Handle::Pseudo(PseudoHandle::CurrentProcess) {
+            throw_unsup_format!(
+                "`DuplicateHandle` `hSourceProcessHandle` parameter is not the current process, which is unsupported"
+            );
+        }
+
+        if this.ptr_is_null(target_handle_ptr)? {
+            throw_unsup_format!(
+                "`DuplicateHandle` `lpTargetHandle` parameter is null, which is unsupported"
```
oli-obk, line 289: The official docs say

> If lpTargetHandle is NULL, the function duplicates the handle, but does not return the duplicate handle value to the caller. This behavior exists only for backward compatibility with previous versions of this function. You should not use this feature, as you will lose system resources until the target process terminates.

So unsupported is correct, but that doesn't have to be in the error message. The message could just say that the pointer must not be null because the handle must be written somewhere or it would get leaked

src/shims/windows/handle.rs
```diff
@@ -244,6 +248,76 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    this.invalid_handle("DuplicateHandle")?
+                };
+                Handle::File(this.machine.fds.insert(fd))
+            }
+            Handle::Thread(_) => {
+                throw_unsup_format!(
+                    "`DuplicateHandle` called on a thread handle, which is unsupported"
+                );
+            }
+            Handle::Pseudo(pseudo) => Handle::Pseudo(pseudo),
```
oli-obk, line 311: Do we have tests for this code path?

CraftSpider: No, we don't currently directly test any of the handle shims outside of ones used directly for filesystem stuff. I can add such tests when I get a chance.

src/shims/windows/handle.rs
CraftSpider: I wasn't sure if we wanted to consider this case 'unsupported by Miri' or 'viewed as a bug by Miri'. I'm fine with considering it to be a program bug.

Issue #4340, opened 2025-05-21
`./miri squash` fails on Windows


RalfJung:
@CraftSpider reports that `./miri squash` fails on Windows.

@CraftSpider could you post the error message you are getting?

CraftSpider:
I'll do so when I have time - the problem may end up being not Windows-specific, I had to go to work basically immediately after seeing the error. It looked so mostly because it involved something with a mangled path - an error about no existing path for `D:somepathhere/foo`, but that's from memory.

ChrisDenton:
Is `./miri` using a mingw bash shell (or is git)? Mingw would mangle windows paths by default so they'd need to be converted to the cygwin path (e.g. by calling `cygpath 'D:\some\path\here'`).

RalfJung:
`./miri` invokes git directly, not via a shell... so sure where paths would be mangled.

ChrisDenton:
git itself might be using the bash shell. I'm pretty sure it still uses some internal scripts.

CraftSpider:
It's the path set in `GIT_SEQUENCE_EDITOR` - that path apparently is expected to point to a tool that git will execute in mingw, and thus should be a linux-like path, not a Windows-like one.

CraftSpider:
Error:
```
hint: Waiting for your editor to close the file... C:\Dev\Git\miri\miri-script\target\debug\miri-script.exe: line 1: C:DevGitmirimiri-scripttargetdebugmiri-script.exe: command not found
error: there was a problem with the editor 'C:\Dev\Git\miri\miri-script\target\debug\miri-script.exe'
Error: `git rebase` failed
```

RalfJung:
Uh, wtf? Why does it interpret escape sequences in an environment variable?!? Or how else are the backslashes disappearing?
It makes no sense to me that we'd have to specifically mangle this path, this seems like a MinGW / git-for-windows bug to me.

RalfJung:
https://github.com/rust-lang/miri/pull/4351 should fix this. But this is clearly a hack working around a bug somewhere else in the ecosystem, probably in MinGW. It'd be great if someone using these tools could report the bug, so that we don't need to have hundreds of work-arounds for this nonsense everywhere but can have it just fixed once and forall.

Meanwhile however, I think this issue can be closed.

RalfJung:
In a year or two we probably should remove that work-around. We'll never get reliable systems if we keep piling hacks upon hacks rather than fixing problems at their root.

PR #4351, opened 2025-05-27
attempt to fix squash on Windows

RalfJung:
Attempt to work around https://github.com/rust-lang/miri/issues/4340.
@CraftSpider could you check if this helps?


CraftSpider:
Yep, `./miri squash` works for me with this PR

RalfJung:
Thanks for testing!

I'd still appreciate if you could also file a bug with git-for-Windows -- clearly it should be possible to use blackslash paths with a windows application; the current behavior makes it unnecessarily hard to write portable code that interacts with `git`.

PR #4362, opened 2025-05-31
isolated_alloc: directly use mmap for allocations

nia-e:
In case there isn't a resolution to #4357, this will fix it at least. Benchmark with the allocator force-enabled vs baseline:
```
Comparison with baseline (relative speed, lower is better for the new results):
  backtraces: 1.10 ± 0.06
  big-allocs: 0.94 ± 0.07
  mse: 0.96 ± 0.03
  range-iteration: 0.99 ± 0.05
  serde1: 1.03 ± 0.05
  serde2: 1.02 ± 0.05
  slice-chunked: 0.99 ± 0.04
  slice-get-unchecked: 0.93 ± 0.04
  string-replace: 0.84 ± 0.04
  unicode: 1.03 ± 0.04
  zip-equal: 0.95 ± 0.07
```


src/alloc/isolated_alloc.rs
```diff
@@ -60,12 +60,13 @@ impl IsolatedAlloc {
 
     /// If the allocation is greater than a page, then round to the nearest page #.
     #[inline]
-    fn huge_normalized_layout(layout: Layout, page_size: usize) -> Layout {
+    fn huge_normalized_layout(&self, layout: Layout, page_size: usize) -> (Layout, usize) {
```
RalfJung: This is odd, since the resulting `Layout` is not what we are actually using -- we can't even tell `mmap` an alignment.

So really the "number of pages to allocate" is the only thing this should return I think.

nia-e: Made it return the size directly instead of page count since that's what we use, but good point. Also got rid of the page_layout function since we only care about the size when mmapping

src/alloc/isolated_alloc.rs
```diff
@@ -52,20 +52,18 @@ impl IsolatedAlloc {
         Layout::from_size_align(size, align).unwrap()
     }
 
-    /// Returns the layout used to allocate the pages that hold small allocations.
+    /// For greater-than-page-sized allocations, returns the minimum allocation
+    /// size to make sure the alignment requested can be given.
```
RalfJung:
```suggestion
    /// For greater-than-page-sized allocations, returns the allocation size we need to request
    /// including the slack we need to satisfy the alignment request.
```

src/alloc/isolated_alloc.rs
```diff
@@ -52,20 +52,18 @@ impl IsolatedAlloc {
...
     #[inline]
-    fn page_layout(&self) -> Layout {
-        Layout::from_size_align(self.page_size, self.page_size).unwrap()
-    }
-
-    /// If the allocation is greater than a page, then round to the nearest page #.
-    #[inline]
-    fn huge_normalized_layout(layout: Layout, page_size: usize) -> Layout {
+    fn huge_normalized_layout(&self, layout: Layout, page_size: usize) -> usize {
```
RalfJung: Since this takes `self`, why does it need `page_size` separately?

src/alloc/isolated_alloc.rs
```diff
@@ -52,20 +52,18 @@ impl IsolatedAlloc {
...
         // Allocate in page-sized chunks
         let size = layout.size().next_multiple_of(page_size);
+        let pg_count = size.div_ceil(page_size);
         // And make sure the align is at least one page
         let align = std::cmp::max(layout.align(), page_size);
-        Layout::from_size_align(size, align).unwrap()
+        let extra_pages = align.strict_div(page_size).saturating_sub(1);
```
RalfJung: Please add a comment explaining this logic.

src/alloc/isolated_alloc.rs
```diff
@@ -183,13 +192,28 @@ impl IsolatedAlloc {
     /// Allocates in multiples of one page on the host system.
```
RalfJung, line 199:
```suggestion
    /// Allocates in multiples of one page on the host system.
    /// Will always be zeroed.
```

src/alloc/isolated_alloc.rs
```diff
@@ -183,13 +192,28 @@ impl IsolatedAlloc {
...
+                libc::PROT_READ | libc::PROT_WRITE,
+                libc::MAP_PRIVATE | libc::MAP_ANONYMOUS,
+                -1,
+                0,
+            )
+            .cast::<u8>()
+        };
+        assert_ne!(ret.addr(), usize::MAX);
+        // Align the pointer as requested
+        let ofs = ret.addr().next_multiple_of(layout.align()).strict_sub(ret.addr());
```
RalfJung: Instead of this "sub" with a later "add", you could do something like `ret.map_addr(|a| a.next_multiple_of(...))`.

src/alloc/isolated_alloc.rs
```diff
@@ -218,15 +242,15 @@ impl IsolatedAlloc {
                 let page_ptr = self.page_ptrs.remove(idx);
                 // SAFETY: We checked that there are no outstanding allocations
                 // from us pointing to this page, and we know it was allocated
-                // with this layout
+                // with this layout whose size is the system page size
```
RalfJung:
```suggestion
                // with this layout whose size is the system page size.
```

src/alloc/isolated_alloc.rs
```diff
@@ -171,8 +168,20 @@ impl IsolatedAlloc {
 
     /// Expands the available memory pool by adding one page.
     fn add_page(&mut self) -> (*mut u8, &mut DenseBitSet<usize>) {
-        // SAFETY: The system page size, which is the layout size, cannot be 0
-        let page_ptr = unsafe { alloc::alloc(self.page_layout()) };
+        // SAFETY: mmap is safe to call as long as the length is a multiple of
+        // the system page size
```
RalfJung: Where do the docs say that the size needs to be a multiple of the page size?

It seems to me like the call is always valid. The system will just give us more memory than we asked for if we request half a page.

src/alloc/isolated_alloc.rs
```diff
@@ -255,18 +279,19 @@ impl IsolatedAlloc {
     /// SAFETY: Same as `dealloc()` with the added requirement that `layout`
     /// must ask for a size larger than the host pagesize.
     unsafe fn dealloc_huge(&mut self, ptr: *mut u8, layout: Layout) {
-        let layout = IsolatedAlloc::huge_normalized_layout(layout, self.page_size);
+        let size = self.huge_normalized_layout(layout, self.page_size);
         // Find the pointer matching in address with the one we got
         let idx = self
             .huge_ptrs
             .iter()
             .position(|pg| ptr.addr() == pg.0.addr())
```
RalfJung: This will not find anything if the pointer had to be offset to ensure the requested alignment.

Please add a native-lib test that requests 4k of memory with alignment 64k. That will then likely explode here.

src/alloc/isolated_alloc.rs
```diff
@@ -255,18 +279,19 @@ impl IsolatedAlloc {
...
             .expect("Freeing unallocated pages");
         // And kick it from the list
-        self.huge_ptrs.remove(idx);
-        // SAFETY: Caller ensures validity of the layout
+        let (un_offset_ptr, _) = self.huge_ptrs.remove(idx);
+        // SAFETY: huge_normalized_layout ensures the validity of the size, and
+        // we know the pointer was the original page-aligned one
         unsafe {
-            alloc::dealloc(ptr, layout);
+            assert_eq!(libc::munmap(un_offset_ptr.cast(), size), 0);
```
RalfJung:
```suggestion
            let ret = libc::munmap(un_offset_ptr.cast(), size);
            assert_eq!(ret, 0);
```
I prefer to not have side-effects inside `assert!`.

src/alloc/isolated_alloc.rs
```diff
@@ -255,18 +279,19 @@ impl IsolatedAlloc {
...
             .huge_ptrs
             .iter()
             .position(|pg| ptr.addr() == pg.0.addr())
             .expect("Freeing unallocated pages");
         // And kick it from the list
-        self.huge_ptrs.remove(idx);
-        // SAFETY: Caller ensures validity of the layout
+        let (un_offset_ptr, _) = self.huge_ptrs.remove(idx);
+        // SAFETY: huge_normalized_layout ensures the validity of the size, and
+        // we know the pointer was the original page-aligned one
```
RalfJung: I have no idea what this safety comment is supposed to mean.
```suggestion
        // SAFETY: huge_ptrs contains allocations that we own, and huge_normalized_layout
        // computed the size of the original allocation (if the user gave us the right `layout`,
        // which we may assume).
```

src/alloc/isolated_alloc.rs
nia-e: oops

src/alloc/isolated_alloc.rs
nia-e: I think I mixed up the length and offset fields, you're right. I'll edit the comment

RalfJung:
Hint: you can accept multiple suggestions into a single commit if you go to the "diff" view of the PR.

Also please avoid force-pushing when review is still in progress as that makes it a lot harder to see what changed since the last review.

nia-e:
Gah, sorry 😓 I can try to push back to the old commit hashes?

src/alloc/isolated_alloc.rs
```diff
@@ -181,15 +197,29 @@ impl IsolatedAlloc {
...
     /// SAFETY: Same as `alloc()`.
-    unsafe fn alloc_huge(&mut self, layout: Layout, zeroed: bool) -> *mut u8 {
-        let layout = IsolatedAlloc::huge_normalized_layout(layout, self.page_size);
-        // SAFETY: Upheld by caller
-        let ret =
-            unsafe { if zeroed { alloc::alloc_zeroed(layout) } else { alloc::alloc(layout) } };
+    unsafe fn alloc_huge(&mut self, layout: Layout) -> *mut u8 {
+        let size = self.huge_normalized_layout(layout);
+        // SAFETY: Layout size being a multiple of the system pagesize is enforced
+        // by huge_normalized_layout
```
RalfJung: Why do we care about layout size being a multiple of the system pagesize?

src/alloc/isolated_alloc.rs
```diff
@@ -181,15 +197,29 @@ impl IsolatedAlloc {
...
+                -1,
+                0,
+            )
+            .cast::<u8>()
+        };
+        assert_ne!(ret.addr(), usize::MAX);
         self.huge_ptrs.push((ret, layout.size()));
-        ret
+        // huge_normalized_layout ensures that we've overallocated enough space
+        // for this to be valid
```
RalfJung:
```suggestion
        // for this to be valid.
```

src/alloc/isolated_alloc.rs
```diff
@@ -171,8 +176,19 @@ impl IsolatedAlloc {
...
+                std::ptr::null_mut(),
+                self.page_size,
+                libc::PROT_READ | libc::PROT_WRITE,
+                libc::MAP_PRIVATE | libc::MAP_ANONYMOUS,
+                -1,
+                0,
+            )
+            .cast::<u8>()
+        };
+        assert_ne!(page_ptr.addr(), usize::MAX);
```
RalfJung:
```suggestion
        assert_ne!(page_ptr.addr(), usize::MAX, "mmap failed");
```

src/alloc/isolated_alloc.rs
```diff
@@ -181,15 +197,29 @@ impl IsolatedAlloc {
...
+                std::ptr::null_mut(),
+                size,
+                libc::PROT_READ | libc::PROT_WRITE,
+                libc::MAP_PRIVATE | libc::MAP_ANONYMOUS,
+                -1,
+                0,
+            )
+            .cast::<u8>()
+        };
+        assert_ne!(ret.addr(), usize::MAX);
```
RalfJung:
```suggestion
        assert_ne!(ret.addr(), usize::MAX, "mmap failed");
```

src/alloc/isolated_alloc.rs
```diff
@@ -218,15 +248,15 @@ impl IsolatedAlloc {
                 let page_ptr = self.page_ptrs.remove(idx);
                 // SAFETY: We checked that there are no outstanding allocations
                 // from us pointing to this page, and we know it was allocated
-                // with this layout
+                // with this layout whose size is the system page size.
```
RalfJung:
```suggestion
                // in add_page as exactly a single page.
```

src/alloc/isolated_alloc.rs
RalfJung, line 1: The comments in `struct IsolatedAlloc` also need updating, in particular the ones referring to the global allocator.

src/alloc/isolated_alloc.rs
```diff
@@ -255,18 +285,21 @@ impl IsolatedAlloc {
...
     /// must ask for a size larger than the host pagesize.
     unsafe fn dealloc_huge(&mut self, ptr: *mut u8, layout: Layout) {
-        let layout = IsolatedAlloc::huge_normalized_layout(layout, self.page_size);
+        let size = self.huge_normalized_layout(layout);
         // Find the pointer matching in address with the one we got
         let idx = self
             .huge_ptrs
             .iter()
-            .position(|pg| ptr.addr() == pg.0.addr())
+            .position(|pg| ptr.addr() == pg.0.addr().next_multiple_of(layout.align()))
```
RalfJung: This logic does not seem right to me... why should this be correct? I think this could return the wrong page! Consider that `pg` here may have been allocated with a smaller alignment, so `next_multiple_of` might return something way outside the range of `pg`.

Please think very slowly and carefully about this, and don't expect the reviewer to find flaws in your logic! The soundness of Miri depends on getting this right.

The right logic should be something like `pg.0.addr() <= ptr.addr() && ptr.addr() < pg.0.addr().strict_add(pg.1)`.

nia-e: ~~I might be misunderstanding but isn't this the exact logic we used when allocating and returning the pointer? Breaking it down case by case:~~
- ~~align ≤ page_size -> the pointer stored in the allocator is not offset, and `next_multiple_of` is a no-op since the pagesize is a multiple of the alignment (both are strictly powers of 2)~~
- ~~align > page_size -> the pointer we returned was `result_of_mmap.next_multiple_of(align)` and therefore this is the same thing being done and this will match~~

~~I had initially implemented basically the logic you requested but then concluded it's overcomplicated when this should work. Sorry if I'm making you do extra work, I thought I did think this through but I'll triple-check it just to be safe~~

Nevermind, I think I understand the problem. I did not realise that was a possibility, guh

src/alloc/isolated_alloc.rs
nia-e: Ah, I guess we don't anymore. I had initially misunderstood the docs for mmap/munmap as requiring page-aligned lengths and addresses, but that was wrong. I'll update this, sorry!

src/alloc/isolated_alloc.rs
nia-e: Should I put periods at the end of every comment? So far I have been using comment-terminal periods and proper capitalisation for doc comments and skipping those for normal comments

src/alloc/isolated_alloc.rs
RalfJung: > isn't this the exact logic we used when allocating and returning the pointer? 

Yes, so indeed the allocation we want to find will return `true` here. But you also must ensure that all other allocations return `false`. Those may have been allocated with a *different* layout and then your logic goes wrong.

RalfJung:
(FYI this is still marked as waiting-for-author.)

src/alloc/isolated_alloc.rs
RalfJung: Generally yes all sentences should use proper spelling, end in a period, and start with a capital letter. IOW, standard orthographic rules apply. :)

No need to change comments you're not otherwise touching in your PR, however.

src/alloc/isolated_alloc.rs
nia-e: Yeah, I realised that eventually x.x

nia-e:
I know, had been waiting for tests to finish to be safe

@rustbot ready

src/alloc/isolated_alloc.rs
nia-e: On the topic of linguistic bikeshedding, I've generally used british spellings for most things - should I switch to solely american english?

RalfJung: I use american spelling, but I won't bother trying to enforce that consistently as that would be hopeless.^^

src/alloc/isolated_alloc.rs
```diff
@@ -255,18 +284,23 @@ impl IsolatedAlloc {
     /// SAFETY: Same as `dealloc()` with the added requirement that `layout`
     /// must ask for a size larger than the host pagesize.
     unsafe fn dealloc_huge(&mut self, ptr: *mut u8, layout: Layout) {
-        let layout = IsolatedAlloc::huge_normalized_layout(layout, self.page_size);
+        let size = self.huge_normalized_layout(layout);
         // Find the pointer matching in address with the one we got
```
RalfJung:
```suggestion
        // Find the huge allocation containing `ptr`.
```

RalfJung:
This looks great, thanks! After resolving the last comment nit, please squash the commits using `./miri squash`. Then write `@rustbot ready` after you force-pushed the squashed PR.

@rustbot author

src/alloc/isolated_alloc.rs
```diff
@@ -255,18 +284,23 @@ impl IsolatedAlloc {
...
             .iter()
-            .position(|pg| ptr.addr() == pg.0.addr())
+            .position(|&(pg, size)| {
+                pg.addr() <= ptr.addr() && ptr.addr() < pg.addr().strict_add(size)
+            })
             .expect("Freeing unallocated pages");
         // And kick it from the list
-        self.huge_ptrs.remove(idx);
-        // SAFETY: Caller ensures validity of the layout
+        let (un_offset_ptr, _) = self.huge_ptrs.remove(idx);
```
RalfJung:
```suggestion
        let (un_offset_ptr, size2) = self.huge_ptrs.remove(idx);
        assert_eq!(size, size2, "got wrong layout in dealloc");
```

src/alloc/isolated_alloc.rs
```diff
@@ -255,18 +284,23 @@ impl IsolatedAlloc {
...
+        // SAFETY: huge_ptrs contains allocations that we own, and huge_normalized_layout
+        // computed the size of the original allocation (if the user gave us the right `layout`,
+        // which we may assume).
```
RalfJung:
```suggestion
        // SAFETY: huge_ptrs contains allocations made with mmap with the size recorded there.
```

PR #4363, opened 2025-05-31
bench: separate baseline results in gitignored dir

nia-e:
To hopefully avoid repeats of rust-lang/miri#4361.

Would there be any interest in extending the benchmark tool so that it could e.g. directly compare results from two git hashes? I found myself trying to mess with things for performance and having something like that might be helpful

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,5 +13,6 @@ flamegraph*.svg
 rustc-ice*.txt
 tests/native-lib/libtestlib.so
 .auto-*
+*.json
 
 /genmc/
--- a/miri-script/src/commands.rs
+++ b/miri-script/src/commands.rs
@@ -568,11 +568,17 @@ impl Command {
             Ok(results)
         };
 
-        if let Some(baseline_file) = save_baseline {
+        if let Some(mut baseline_file) = save_baseline {
+            if !baseline_file.ends_with(".json") {
+                baseline_file.push_str(".json");
+            }
             let results = gather_results()?;
             let baseline = File::create(baseline_file)?;
             serde_json::to_writer_pretty(BufWriter::new(baseline), &results)?;
-        } else if let Some(baseline_file) = load_baseline {
+        } else if let Some(mut baseline_file) = load_baseline {
+            if !baseline_file.ends_with(".json") {
+                baseline_file.push_str(".json");
+            }
             let new_results = gather_results()?;
             let baseline_results: BTreeMap<String, BenchResult> = {
                 let f = File::open(baseline_file)?;


RalfJung:
Thanks! However, TBH I'd rather not have a separate directory for this.

But I'd be fine with adding `*.json` to the gitignore. At least my baseline files usually have that file ending due to their format (we could nudge people in that direction by also using such a filename in the example, if we don't already do that).

nia-e:
Hmm, should I make it so it auto-appends .json if it's not specified? I didn't personally notice anything that would suggest that's the preferred naming scheme and it's probably more flexible if I can just pass `--load-baseline filename` and it grabs `filename.json`. I'll push that idea to this branch and wait for feedback

RalfJung:
I don't think we should be altering the filename the user chose.

nia-e:
I guess I'm just unsure where to put an example, I don't see a nonintrusive way to put it into the `--help` menu.  I'm also fine just carrying that original separate directory patch myself and hooking up a script to make sure I never push it by accident, since this wasn't a problem anyone else has had at this point

RalfJung:
There's docs in CONTRIBUTING.md but those already use the .json file extension. So yeah not sure there's anything to be done here.

One reason why I don't want to change this is that tab completion should work for `--load-baseline`, and it won't work if we append a suffix or prepend a directory name.

RalfJung:
> Would there be any interest in extending the benchmark tool so that it could e.g. directly compare results from two git hashes? I found myself trying to mess with things for performance and having something like that might be helpful

How do you imagine that to work? I don't want the tool to run any git commands (at least none that would mutate the repo or worktree), that seems like it could get too messy too quickly.

nia-e:
Similar tools I saw in the past would generally check that the current tree is clean and if so check out the first hash provided, benchmark it, then the 2nd, benchmark it, then check out the original head. I think it can be set up to handle user interrupts a la ctrl-c such that it stops the running benchmark, deletes temp files if any, and checks out the original head. But if that's too flimsy that's fair enough

nia-e:
> There's docs in CONTRIBUTING.md but those already use the .json file extension. So yeah not sure there's anything to be done here.
> 
> One reason why I don't want to change this is that tab completion should work for `--load-baseline`, and it won't work if we append a suffix or prepend a directory name.

Fair enough, though for what it's worth the auto-appending code I added wouldn't trigger if it saw the json suffix was already specified so tab completion would be unaffected

RalfJung:
Ah, I see. That feels a bit too magical to me...

nia-e:
That's fair. I can close the PR and just carry the patch myself, I find it useful

RalfJung:
Yeah maybe that's best for now. Thanks for looking into ways to improve this!

PR #4368, opened 2025-06-01
Fix tokio/file-io.rs test relying on `read`/`write` not being short

### LINKED ISSUE #4367: tokio/file-io test is non-deterministic
~~`./miri run --dep tests/pass-dep/tokio/file-io.rs -Zmiri-disable-isolation -Zmiri-many-seeds` fails on seed 40 for me. Overall, around 90% of seeds seem to fail. No idea why this did not explode earlier, maybe the distribution changed recently.~~

`tests/pass-dep/tokio/file-io.rs` has [failed on rustc CI](https://github.com/rust-lang/rust/pull/141834#issuecomment-2926854470); it seems to be non-deterministic. I haven't yet been able to reproduce this locally. (The reproducer I had before was completely bogus.)

@tiif can you look into this? This is causing issues for rustc CI so it is somewhat urgent.

Noratrieb:
The test did `write` and `read` and hoped that it would read/write everything, which doesn't always happen and caused CI failures. Switch to `write_all` and `read_to_end` to make it more reliable.

Probably fixes rust-lang/miri#4367.

### DIFF
--- a/tests/pass-dep/tokio/file-io.rs
+++ b/tests/pass-dep/tokio/file-io.rs
@@ -20,7 +20,7 @@ async fn test_create_and_write() -> io::Result<()> {
     let mut file = File::create(&path).await?;
 
     // Write 10 bytes to the file.
-    file.write(b"some bytes").await?;
+    file.write_all(b"some bytes").await?;
     assert_eq!(file.metadata().await.unwrap().len(), 10);
 
     remove_file(&path).unwrap();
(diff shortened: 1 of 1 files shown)


RalfJung:
This does not help, I'm still getting a 90% failure rate.

~~Miri doesn't do partial reads/writes so I would have been surprised if this made a difference.~~ It actually can do them for file accesses as there we just use the host methods.

That said, this still clearly makes the test more correct, so let's land this anyway.

RalfJung:
Actually my reproducer was wrong so this will probably fix the issue.

PR #4371, opened 2025-06-02
Add shim test for `DuplicateHandle`

CraftSpider:


RalfJung:
There are some CI failures, could you look into that?
@rustbot author

CraftSpider:
Well, I've just learned that attempting to read from a write-only file on Unix returns `EBADF`, which Rust maps to `Uncategorized`. Which surprises me for at least two reasons. But either way, I should be able to fix it by either:
- Mapping EBADF specifically on Linux
- Making the test allow this error.

The first though would likely be the first in a long line of this kind of change - several currently-disabled filesystem-related tests appear to rely on the specific OS error returned by functions, but this is a property of the _backend_, not the emulated target. If it's desired to always match the OS error kinds returned by the target, I'd want to brainstorm a more holistic way to handle the likely large amount of mapping that will require (and in some cases probably be impossible, since not all OSes will provide the same error granularity)

RalfJung:
EBADF is the error you get when the FD doesn't even exist, so it sounds to me like something else is going on.

But, anyway, why does this test have to involve "reading from a write-only file"? You can mirror the test at
https://github.com/rust-lang/miri/blob/57bac2e8bcf5d925126a42634bebf44437c4d5fb/tests/pass-dep/libc/libc-fs.rs#L82-L105
by having two FDs read from the same file and observing that reading one advances the position of the other. (The `dup` test doesn't even quite do that, it just checks that the position is copied over on `dup`; https://github.com/rust-lang/miri/pull/4382 fixes that.)

CraftSpider:
I was hoping to verify that permissions were copied correctly, was the idea. I can just test position matching. That does still leave the underlying issue of the error kind being based on the miri target, not the emulated one. But that's also pre-existing.

Also, EBADF being the error for no read permissions, outside of Miri and with no `dup` involved, can be seen here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=7574e0de8939c17772ca441a6c64e1d8. I checked before concluding that was just the expected error.

RalfJung:
> Also, EBADF being the error for no read permissions, outside of Miri and with no dup involved, can be seen here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=7574e0de8939c17772ca441a6c64e1d8. I checked before concluding that was just the expected error.

:+1: 

> That does still leave the underlying issue of the error kind being based on the miri target, not the emulated one. But that's also pre-existing.

Yeah, that's not surprising, though maybe we should have an issue for this. It'd take a huge amount of code to match up the error kinds, I don't think it's worth it.

CraftSpider:
Just a note: Didn't forget this, I just bought a house, I'll get to this when I have time again.

RalfJung:

@rustbot author

PR #4372, opened 2025-06-02
native-lib: allow multiple libraries and/or dirs

nia-e:
Since I'm expanding native-lib mode it will hopefully become more useful and we may want to pass in more complex libraries with dependencies. This makes that possible. Since defining the same symbol multiple times in the one native-lib was already UB, I didn't add a note warning against defining the same symbol multiple times across libs but maybe I should?

I'm also looking into how we could use the ptrace stuff to lift some of the other requirements, like how native-lib mode is not allowed to spawn threads that run concommittantly with the Rust code - since we're ptracing we might be able to e.g. park all C threads, let Miri do its thing for x amount of time, then re-enable FFI mode and let the foreign threads do their thing for x amount of time and track their accesses. This way some much more complex multithreaded programs might be runnable. But, for now that's in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -398,9 +398,11 @@ to Miri failing to detect cases of undefined behavior in a program.
   **unsound** since the fallback body might not be checking for all UB.
 * `-Zmiri-native-lib=<path to a shared object file>` is an experimental flag for providing support
   for calling native functions from inside the interpreter via FFI. The flag is supported only on
-  Unix systems. Functions not provided by that file are still executed via the usual Miri shims.
+  Unix systems. Functions not provided by that file are still executed via the usual Miri shims. If
+  a path to a directory is specified, all files in that directory are included nonrecursively. This
+  flag can be passed multiple times to specify multiple files and/or directories.
   **WARNING**: If an invalid/incorrect `.so` file is specified, this can cause Undefined Behavior in
-  Miri itself! And of course, Miri cannot do any checks on the actions taken by the native code.
+  Miri itself! And of course, Miri often cannot do any checks on the actions taken by the native code.
   Note that Miri has its own handling of file descriptors, so if you want to replace *some*
   functions working on file descriptors, you will have to replace *all* of them, or the two kinds of
   file descriptors will be mixed up. This is **work in progress**; currently, only integer and
--- a/src/shims/native_lib.rs
+++ b/src/shims/native_lib.rs
@@ -87,47 +87,52 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
     }
 
     /// Get the pointer to the function of the specified name in the shared object file,
-    /// if it exists. The function must be in the shared object file specified: we do *not*
-    /// return pointers to functions in dependencies of the library.  
+    /// if it exists. The function must be in one of the shared object files specified:
+    /// we do *not* return pointers to functions in dependencies of libraries.
     fn get_func_ptr_explicitly_from_lib(&mut self, link_name: Symbol) -> Option<CodePtr> {
         let this = self.eval_context_mut();
-        // Try getting the function from the shared library.
-        let (lib, lib_path) = this.machine.native_lib.as_ref().unwrap();
-        let func: libloading::Symbol<'_, unsafe extern "C" fn()> =
-            unsafe { lib.get(link_name.as_str().as_bytes()).ok()? };
-        #[expect(clippy::as_conversions)] // fn-ptr to raw-ptr cast needs `as`.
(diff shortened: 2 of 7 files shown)


src/shims/native_lib.rs
```diff
@@ -87,47 +87,52 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    let info = info.assume_init();
+                    #[cfg(target_os = "cygwin")]
+                    let fname_ptr = info.dli_fname.as_ptr();
+                    #[cfg(not(target_os = "cygwin"))]
+                    let fname_ptr = info.dli_fname;
+                    assert!(!fname_ptr.is_null());
+                    if std::ffi::CStr::from_ptr(fname_ptr).to_str().unwrap()
+                        != lib_path.to_str().unwrap()
+                    {
+                        continue;
```
oli-obk: if we've found the symbol, but the loading fails, should that really just continue to the next lib?

nia-e: Fair enough, I guess it's on the user to make sure not to pass us any broken/wonky shared objects. I'll change that

README.md
```diff
@@ -398,9 +398,11 @@ to Miri failing to detect cases of undefined behavior in a program.
   **unsound** since the fallback body might not be checking for all UB.
 * `-Zmiri-native-lib=<path to a shared object file>` is an experimental flag for providing support
   for calling native functions from inside the interpreter via FFI. The flag is supported only on
-  Unix systems. Functions not provided by that file are still executed via the usual Miri shims.
+  Unix systems. Functions not provided by that file are still executed via the usual Miri shims. If
+  a path to a directory is specified, all files in that directory are included nonrecursively. This
+  flag can be passed multiple times to specify multiple files and/or directories.
```
RalfJung, line 403: What does passing a directory here even do?

src/shims/native_lib.rs
RalfJung: I think `continue` actually makes sense here... this check failing means the symbol wasn't found in this library (only in one of its dependencies). The symbol might still be found in another library.

This is not about the loading of the library failing.

src/shims/native_lib.rs
```diff
@@ -87,47 +87,52 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // library if it can't find the symbol in the library itself.
+            // So, in order to check if the function was actually found in the specified
+            // `machine.external_so_lib` we need to check its `dli_fname` and compare it to
+            // the specified SO file path.
+            // This code is a reimplementation of the mechanism for getting `dli_fname` in `libloading`,
+            // from: https://docs.rs/libloading/0.7.3/src/libloading/os/unix/mod.rs.html#411
+            // using the `libc` crate where this interface is public.
+            let mut info = std::mem::MaybeUninit::<libc::Dl_info>::zeroed();
+            unsafe {
+                if libc::dladdr(fn_ptr, info.as_mut_ptr()) != 0 {
```
RalfJung, line 117: The more concerning point is that we just ignore `dladdr` failing...

README.md
nia-e: `-Zmiri-native-lib=/path/to/dir/` -> `-Zmiri-native-lib=/path/to/dir/file1 -Zmiri-native-lib=/path/to/dir/file2 -Zmiri-native-lib=/path/to/dir/file3 ...`

src/shims/native_lib.rs
nia-e: I can change that back then and put an `else` on the dladdr bit which returns none to address the other comment also?

README.md
RalfJung: Okay; the docs should say that.

src/shims/native_lib.rs
RalfJung: I think the else branch should panic, TBH.

README.md
RalfJung: Oh wait, they do, I just missed it. :facepalm:

src/shims/native_lib.rs
nia-e: Do we want to panic on fail? With the current dir implementation it would allow e.g. passing a directory which also contains things that aren't libraries and it will just ignore those. For instance it could be a directory of random build artifacts from some library, where only some are actually .so files, and this will transparently ignore all other files

RalfJung: I already implemented this in https://github.com/rust-lang/miri/pull/4389. At least for single-file arguments I think it should definitely be an error if that file cannot be loaded.

PR #4376, opened 2025-06-04
TB: Add flag to disable the more precise interior mutability tracking

yoctocell:
This adds a `-Zmiri-disable-tb-precise-interior-mut` flag that makes the tracking of `UnsafeCell` less precise (like how it was in rust-lang/miri#4273), i.e. Miri defaults to the new, more precise tracking.

cc @RalfJung @JoJoDeveloping

### DIFF
--- a/README.md
+++ b/README.md
@@ -458,6 +458,10 @@ to Miri failing to detect cases of undefined behavior in a program.
   This is much less likely with Stacked Borrows.
   Using Tree Borrows currently implies `-Zmiri-strict-provenance` because integer-to-pointer
   casts are not supported in this mode, but that may change in the future.
+* `-Zmiri-tree-borrows-no-precise-interior-mut` makes Tree Borrows
+  track interior mutable data on the level of references instead of on the
+  byte-level as is done by default.  Therefore, with this flag, Tree
+  Borrows will be more permissive.
 * `-Zmiri-force-page-size=<num>` overrides the default page size for an architecture, in multiples of 1k.
   `4` is default for most targets. This value should always be a power of 2 and nonzero.
 
--- a/src/bin/miri.rs
+++ b/src/bin/miri.rs
@@ -35,7 +35,7 @@ use std::sync::{Arc, Once};
 
 use miri::{
     BacktraceStyle, BorrowTrackerMethod, GenmcConfig, GenmcCtx, MiriConfig, MiriEntryFnType,
-    ProvenanceMode, RetagFields, ValidationMode,
+    ProvenanceMode, RetagFields, TreeBorrowsParams, ValidationMode,
 };
 use rustc_abi::ExternAbi;
 use rustc_data_structures::sync;
@@ -554,8 +554,21 @@ fn main() {
         } else if arg == "-Zmiri-disable-stacked-borrows" {
             miri_config.borrow_tracker = None;
         } else if arg == "-Zmiri-tree-borrows" {
-            miri_config.borrow_tracker = Some(BorrowTrackerMethod::TreeBorrows);
+            miri_config.borrow_tracker =
+                Some(BorrowTrackerMethod::TreeBorrows(TreeBorrowsParams {
+                    precise_interior_mut: true,
+                }));
             miri_config.provenance_mode = ProvenanceMode::Strict;
+        } else if arg == "-Zmiri-tree-borrows-no-precise-interior-mut" {
+            match &mut miri_config.borrow_tracker {
+                Some(BorrowTrackerMethod::TreeBorrows(params)) => {
+                    params.precise_interior_mut = false;
+                }
+                _ =>
+                    show_error!(
+                        "`-Zmiri-tree-borrows` is required before `-Zmiri-tree-borrows-no-precise-interior-mut`"
+                    ),
+            };
         } else if arg == "-Zmiri-disable-data-race-detector" {
             miri_config.data_race_detector = false;
             miri_config.weak_memory_emulation = false;
@@ -725,7 +738,7 @@ fn main() {
         }
     }
     // Tree Borrows implies strict provenance, and is not compatible with native calls.
-    if matches!(miri_config.borrow_tracker, Some(BorrowTrackerMethod::TreeBorrows)) {
+    if matches!(miri_config.borrow_tracker, Some(BorrowTrackerMethod::TreeBorrows { .. })) {
         if miri_config.provenance_mode != ProvenanceMode::Strict {
             show_error!(
                 "Tree Borrows does not support integer-to-pointer casts, and hence requires strict provenance"
@@ -735,6 +748,7 @@ fn main() {
             show_error!("Tree Borrows is not compatible with calling native functions");
         }
     }
+
     // Native calls and strict provenance are not compatible.
     if miri_config.native_lib.is_some() && miri_config.provenance_mode == ProvenanceMode::Strict {
         show_error!("strict provenance is not compatible with calling native functions");
--- a/src/borrow_tracker/mod.rs
+++ b/src/borrow_tracker/mod.rs
@@ -226,7 +226,13 @@ pub enum BorrowTrackerMethod {
     /// Stacked Borrows, as implemented in borrow_tracker/stacked_borrows
     StackedBorrows,
     /// Tree borrows, as implemented in borrow_tracker/tree_borrows
-    TreeBorrows,
+    TreeBorrows(TreeBorrowsParams),
+}
+
+/// Parameters that Tree Borrows can take.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct TreeBorrowsParams {
+    pub precise_interior_mut: bool,
 }
 
 impl BorrowTrackerMethod {
@@ -236,6 +242,13 @@ impl BorrowTrackerMethod {
             config.tracked_pointer_tags.clone(),
             config.retag_fields,
         ))
+    }
+
+    pub fn get_tree_borrows_params(self) -> TreeBorrowsParams {
+        match self {
+            BorrowTrackerMethod::TreeBorrows(params) => params,
+            _ => panic!("can only be called when `BorrowTrackerMethod` is `TreeBorrows`"),
+        }
     }
 }
 
@@ -252,7 +265,7 @@ impl GlobalStateInner {
                 AllocState::StackedBorrows(Box::new(RefCell::new(Stacks::new_allocation(
                     id, alloc_size, self, kind, machine,
                 )))),
-            BorrowTrackerMethod::TreeBorrows =>
+            BorrowTrackerMethod::TreeBorrows { .. } =>
                 AllocState::TreeBorrows(Box::new(RefCell::new(Tree::new_allocation(
                     id, alloc_size, self, kind, machine,
                 )))),
@@ -271,7 +284,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let method = this.machine.borrow_tracker.as_ref().unwrap().borrow().borrow_tracker_method;
         match method {
             BorrowTrackerMethod::StackedBorrows => this.sb_retag_ptr_value(kind, val),
-            BorrowTrackerMethod::TreeBorrows => this.tb_retag_ptr_value(kind, val),
+            BorrowTrackerMethod::TreeBorrows { .. } => this.tb_retag_ptr_value(kind, val),
         }
     }
 
@@ -284,7 +297,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let method = this.machine.borrow_tracker.as_ref().unwrap().borrow().borrow_tracker_method;
         match method {
             BorrowTrackerMethod::StackedBorrows => this.sb_retag_place_contents(kind, place),
-            BorrowTrackerMethod::TreeBorrows => this.tb_retag_place_contents(kind, place),
+            BorrowTrackerMethod::TreeBorrows { .. } => this.tb_retag_place_contents(kind, place),
         }
     }
 
@@ -293,7 +306,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let method = this.machine.borrow_tracker.as_ref().unwrap().borrow().borrow_tracker_method;
         match method {
             BorrowTrackerMethod::StackedBorrows => this.sb_protect_place(place),
-            BorrowTrackerMethod::TreeBorrows => this.tb_protect_place(place),
+            BorrowTrackerMethod::TreeBorrows { .. } => this.tb_protect_place(place),
         }
     }
 
@@ -302,7 +315,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let method = this.machine.borrow_tracker.as_ref().unwrap().borrow().borrow_tracker_method;
         match method {
             BorrowTrackerMethod::StackedBorrows => this.sb_expose_tag(alloc_id, tag),
-            BorrowTrackerMethod::TreeBorrows => this.tb_expose_tag(alloc_id, tag),
+            BorrowTrackerMethod::TreeBorrows { .. } => this.tb_expose_tag(alloc_id, tag),
         }
     }
 
@@ -319,7 +332,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.tcx.tcx.dcx().warn("Stacked Borrows does not support named pointers; `miri_pointer_name` is a no-op");
                 interp_ok(())
             }
-            BorrowTrackerMethod::TreeBorrows =>
+            BorrowTrackerMethod::TreeBorrows { .. } =>
                 this.tb_give_pointer_debug_name(ptr, nth_parent, name),
         }
     }
@@ -333,7 +346,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let method = borrow_tracker.borrow().borrow_tracker_method;
         match method {
             BorrowTrackerMethod::StackedBorrows => this.print_stacks(alloc_id),
-            BorrowTrackerMethod::TreeBorrows => this.print_tree(alloc_id, show_unnamed),
+            BorrowTrackerMethod::TreeBorrows { .. } => this.print_tree(alloc_id, show_unnamed),
         }
     }
 
--- a/src/borrow_tracker/tree_borrows/mod.rs
+++ b/src/borrow_tracker/tree_borrows/mod.rs
@@ -312,8 +312,6 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         }
 
         let span = this.machine.current_span();
-        let alloc_extra = this.get_alloc_extra(alloc_id)?;
-        let mut tree_borrows = alloc_extra.borrow_tracker_tb().borrow_mut();
 
         // Store initial permissions and their corresponding range.
         let mut perms_map: RangeMap<LocationState> = RangeMap::new(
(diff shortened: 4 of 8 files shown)


yoctocell:
Hmm, `tests/pass/tree_borrows/cell-inside-struct.rs` fails on CI but passes on my machine.  On my machine it prints the following tree

```
0..   4..   8
| Act | Act |    └─┬──<248=root of the allocation>
|?Cel |?Cel |      └────<249=a>
```

But on CI it prints

```
  0..   8
  | Act |    └─┬──<248=root of the allocation>
  |?Cel |      └────<249=a>
```

which should be the same thing? Maybe I should remove the printing of the tree from the test?

src/bin/miri.rs
```diff
@@ -554,8 +554,12 @@ fn main() {
         } else if arg == "-Zmiri-disable-stacked-borrows" {
             miri_config.borrow_tracker = None;
         } else if arg == "-Zmiri-tree-borrows" {
-            miri_config.borrow_tracker = Some(BorrowTrackerMethod::TreeBorrows);
+            miri_config.borrow_tracker =
+                Some(BorrowTrackerMethod::TreeBorrows { precise_interior_mut: true });
             miri_config.provenance_mode = ProvenanceMode::Strict;
+        } else if arg == "-Zmiri-disable-tb-precise-interior-mut" {
+            // Should this automatically imply "-Zmiri-tree-borrows"?
+            miri_config.tb_precise_interior_mut = false;
```
RalfJung: So there's two places in the config now that control whether TB uses precise-interior-mut? That doesn't make sense.

This should probably set the `precise_interior_mut` field inside `borrow_tracker`.

src/bin/miri.rs
```diff
@@ -554,8 +554,12 @@ fn main() {
         } else if arg == "-Zmiri-disable-stacked-borrows" {
             miri_config.borrow_tracker = None;
         } else if arg == "-Zmiri-tree-borrows" {
-            miri_config.borrow_tracker = Some(BorrowTrackerMethod::TreeBorrows);
+            miri_config.borrow_tracker =
+                Some(BorrowTrackerMethod::TreeBorrows { precise_interior_mut: true });
             miri_config.provenance_mode = ProvenanceMode::Strict;
+        } else if arg == "-Zmiri-disable-tb-precise-interior-mut" {
```
RalfJung: Let's have all TB flags start with a common prefix:
`-Zmiri-tree-borrows-no-precise-interior-mut`

Also, please add the flag to the README.

src/bin/miri.rs
JoJoDeveloping: Which would also answers the question of whether this implies TB ("yes it does")

JoJoDeveloping:
The Miri GC can "compact" RangeMaps if two adjacent ranges have the same content. So maybe add `-Zmiri-provenance-gc=0` to the flags for that test. I think there's a CI pass that runs with a more aggressive GC which probably triggers this.

src/borrow_tracker/mod.rs
```diff
@@ -237,6 +237,13 @@ impl BorrowTrackerMethod {
             config.retag_fields,
         ))
     }
+
+    pub fn has_precise_interior_mut(self) -> bool {
```
RalfJung: This function doesn't make much sense outside TB and doesn't return a sensible result there... I don't think we want to have such a function.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -367,10 +367,40 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // we treat the whole range as an `UnsafeCell`.
+            if has_unsafe_cell && !precise_interior_mut {
+                let nonfreeze_perm = new_perm.nonfreeze_perm;
+                let sifa = nonfreeze_perm.strongest_idempotent_foreign_access(protected);
+                if new_perm.nonfreeze_access {
+                    *perm = LocationState::new_accessed(nonfreeze_perm, sifa);
+                } else {
+                    *perm = LocationState::new_non_accessed(nonfreeze_perm, sifa);
+                }
+            }
```
RalfJung: This logic is unnecessarily convoluted. You're first computing the entire state with `UnsafeCell`, and then you *overwrite it*. Mutable state is an anti-pattern, so please avoid that. Instead, adjust the logic so that it constructs the right values from the start.

src/borrow_tracker/mod.rs
RalfJung: I'd propose adding a new type, `TreeBorrowsParams`, as the only field of the `TreeBorrows` variant. Then you can have a `get_tree_borrows_params` method here that panics when the current tracker is not TB. In fact you maybe want to add that further up to make it easier to call.

src/bin/miri.rs
RalfJung: I think the flag should error if TB hasn't been enabled by this point in the argument parsing.

src/borrow_tracker/tree_borrows/mod.rs
yoctocell: The problem with computing the correct permission from the start is that we don't know in advance whether there will be an `UnsafeCell` in the range, we have to go through the range to obtain this knowledge.  So when we go through the bytes in the range, and we see that `frozen` is true, we don't know whether the permission should be `freeze_perm` or `nonfreeze_perm` because later in the range, we might or might not have an `UnsafeCell`.

src/borrow_tracker/mod.rs
yoctocell: What should `TreeBorrowsParams` look like? Should it be a struct like this 

```rust
struct TreeBorrowsParams { precise_interior_mut: bool }
```

 Or what did you have in mind?

src/borrow_tracker/tree_borrows/mod.rs
JoJoDeveloping: So use `is_ty_freeze`? Which differs in the handling of `UnsafeCell<()>` struct fields however

RalfJung: When "precise interior mutability tracking" is disabled, `visit_freeze_sensitive` should never be called. That's how the code worked before your PR, after all.

src/borrow_tracker/mod.rs
RalfJung: Yes, that looks good.

src/bin/miri.rs
```diff
@@ -555,11 +555,22 @@ fn main() {
...
             // Should this automatically imply "-Zmiri-tree-borrows"?
-            miri_config.tb_precise_interior_mut = false;
+            if matches!(miri_config.borrow_tracker, Some(BorrowTrackerMethod::TreeBorrows { .. })) {
+                miri_config.borrow_tracker =
+                    Some(BorrowTrackerMethod::TreeBorrows(TreeBorrowsParams {
+                        precise_interior_mut: false,
+                    }));
+            } else {
+                show_error!(
+                    "`-Zmiri-tree-borrows` must also be set when setting `-Zmiri-tree-borrows-no-precise-interior-mut"
```
JoJoDeveloping: The error message should in particular say that it has to set "before" that flag.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -342,35 +342,74 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
         let protected = new_perm.protector.is_some();
-        this.visit_freeze_sensitive(place, ptr_size, |range, frozen| {
-            has_unsafe_cell = has_unsafe_cell || !frozen;
-
-            // We are only ever `Frozen` inside the frozen bits.
-            let (perm, access) = if frozen {
+        // NOTE: Using `ty_is_freeze` is not as precise as going through the range
+        // and computing `has_unsafe_cell`.  It can happen that `has_unsafe_cell`
+        // is false, but `!ty_is_freeze` is true.
+        let ty_is_freeze = place.layout.ty.is_freeze(*this.tcx, this.typing_env());
```
JoJoDeveloping: `ty_is_freeze` should only be called if `!precise_interior_mut`

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -404,7 +442,11 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             new_tag,
             perms_map,
             // Allow lazily writing to surrounding data if we found an `UnsafeCell`.
-            if has_unsafe_cell { new_perm.nonfreeze_perm } else { new_perm.freeze_perm },
+            if has_unsafe_cell || (!precise_interior_mut && !ty_is_freeze) {
```
JoJoDeveloping: in other words, compute this permission here inside the `if` above.

tests/pass/tree_borrows/cell-inside-struct.rs
```diff
@@ -0,0 +1,34 @@
+//! The same as `tests/fail/tree-borrows/cell-inside-struct` but with
+//! precise tracking of interior mutability disabled.
+//@compile-flags: -Zmiri-tree-borrows -Zmiri-tree-borrows-no-precise-interior-mut
```
JoJoDeveloping, line 3: I think this still needs to disable the GC?

src/borrow_tracker/tree_borrows/mod.rs
yoctocell: You mean like this?

```rust
let default_perm = if !precise_interior_mut {
    ...
    <permission>
} else {
    this.visit_freeze_sensitive(...);
    <permission>
}

File: tests/pass/tree_borrows/cell-inside-struct.rs (hunk 8 above)
Comment (yoctocell) [reply tests/pass/tree_borrows/cell-inside-struct.rs:3]: Now the tree printed is also [compacted](https://github.com/rust-lang/miri/pull/4376/files/1aeac15008767ba13a8ea157f1811a4318054c37..adf6838d1aa39407004229fb2c138ddf0ebbdabf#diff-da183195935df371d899392cf40694d044d25cd22986fa1bd4ae5bbf6ab9f028), so I don't think this is necessary?

File: src/borrow_tracker/tree_borrows/mod.rs (hunk 7 above)
Comment (JoJoDeveloping) [reply src/borrow_tracker/tree_borrows/mod.rs]: Yes that'd make sense

File: tests/pass/tree_borrows/cell-inside-struct.rs (hunk 8 above)
Comment (JoJoDeveloping) [reply tests/pass/tree_borrows/cell-inside-struct.rs:3]: Yes then it presumably works.

**Comment by RalfJung** [comment]
@rustbot author
(please ping me when this is ready for another review from me :)

File: src/borrow_tracker/tree_borrows/mod.rs [hunk 9]
```diff
@@ -342,35 +342,80 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                 (new_perm.nonfreeze_perm, new_perm.nonfreeze_access)
             };
+            let sifa = perm.strongest_idempotent_foreign_access(protected);
+            let new_loc = if access {
+                LocationState::new_accessed(perm, sifa)
+            } else {
+                LocationState::new_non_accessed(perm, sifa)
+            };
+
+            for (_loc_range, loc) in perms_map.iter_mut_all() {
```
JoJoDeveloping, line 370: Just to make sure, this will create a range map with only one range, i.e. it is maximally compacted, right?

src/bin/miri.rs
```diff
@@ -554,8 +554,23 @@ fn main() {
...
             miri_config.borrow_tracker = None;
         } else if arg == "-Zmiri-tree-borrows" {
-            miri_config.borrow_tracker = Some(BorrowTrackerMethod::TreeBorrows);
+            miri_config.borrow_tracker =
+                Some(BorrowTrackerMethod::TreeBorrows(TreeBorrowsParams {
+                    precise_interior_mut: true,
+                }));
             miri_config.provenance_mode = ProvenanceMode::Strict;
+        } else if arg == "-Zmiri-tree-borrows-no-precise-interior-mut" {
+            // Should this automatically imply "-Zmiri-tree-borrows"?
```
JoJoDeveloping: this comment is now outdated.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -342,35 +342,80 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            .borrow_tracker
+            .as_ref()
+            .unwrap()
+            .borrow()
+            .borrow_tracker_method
+            .get_tree_borrows_params()
+            .precise_interior_mut;
+
+        let default_perm = if !precise_interior_mut {
+            // NOTE: Using `ty_is_freeze` is not as precise as going through the range
```
JoJoDeveloping: "precise" is the wrong word here, isn't it? There is a difference in the handling of zero-sized UnsafeCells, but that's not really about precision, more about design decisions.

JoJoDeveloping (commented):
(whoops these comments were supposed to be part of the review above)

src/borrow_tracker/tree_borrows/mod.rs
yoctocell: I would think so, since we first create `perms_map` with `RangeMap::new`, which creates one range with a dummy `LocationState` , and here we just change it to the correct `LocationState`.

src/borrow_tracker/tree_borrows/mod.rs
yoctocell: Sometimes they also differ (`has_unsafe_cell` is false and `!ty_is_freeze` is true) when the size isn't 0.  For example

```
~/src/miri $ ./miri run tests/pass/tree_borrows/cell-alternate-writes.rs
...
 false, true, Size(8 bytes), MPlaceTy { mplace: MemPlace { ptr: 0x20168[alloc25]<22>, meta: Meta(0x20180[alloc26]<23>), misaligned: None }, ty: dyn std::ops::Fn() -> i32 + std::marker::Sync + std::panic::RefUnwindSafe } 
...
```

From what I have tested, every time this happens, the pointee type is a function.

README.md
```diff
@@ -458,6 +458,10 @@ to Miri failing to detect cases of undefined behavior in a program.
   This is much less likely with Stacked Borrows.
   Using Tree Borrows currently implies `-Zmiri-strict-provenance` because integer-to-pointer
   casts are not supported in this mode, but that may change in the future.
+* `-Zmiri-tree-borrows-no-precise-interior-mut` makes Tree Borrows
+  track interior mutable data on the node-level instead of on the
+  byte-level as is done by default.  Therefore, with this flag, Tree
+  Borrows will be more permissive.
```
RalfJung:
```suggestion
* `-Zmiri-tree-borrows-no-precise-interior-mut` makes Tree Borrows
  track interior mutable data on the level of references instead of on the
  byte-level as is done by default.  Therefore, with this flag, Tree
  Borrows will be more permissive.
```

src/bin/miri.rs
```diff
@@ -554,8 +554,22 @@ fn main() {
...
+                Some(BorrowTrackerMethod::TreeBorrows(TreeBorrowsParams {
+                    precise_interior_mut: true,
+                }));
             miri_config.provenance_mode = ProvenanceMode::Strict;
+        } else if arg == "-Zmiri-tree-borrows-no-precise-interior-mut" {
+            if matches!(miri_config.borrow_tracker, Some(BorrowTrackerMethod::TreeBorrows { .. })) {
+                miri_config.borrow_tracker =
+                    Some(BorrowTrackerMethod::TreeBorrows(TreeBorrowsParams {
+                        precise_interior_mut: false,
+                    }));
```
RalfJung: The more idiomatic way to do this is to capture a reference to the options inside `miri_config.borrow_tracker` and mutate them in-place. This also makes the logic more robust as it will then remain correct when more options are added in the future.

src/borrow_tracker/mod.rs
```diff
@@ -226,7 +226,12 @@ pub enum BorrowTrackerMethod {
     /// Stacked Borrows, as implemented in borrow_tracker/stacked_borrows
     StackedBorrows,
     /// Tree borrows, as implemented in borrow_tracker/tree_borrows
-    TreeBorrows,
+    TreeBorrows(TreeBorrowsParams),
+}
+/// Parameters that Tree Borrows can take.
```
RalfJung, line 232:
```suggestion

/// Parameters that Tree Borrows can take.
```

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -342,35 +342,80 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-            has_unsafe_cell = has_unsafe_cell || !frozen;
-
-            // We are only ever `Frozen` inside the frozen bits.
-            let (perm, access) = if frozen {
+        let precise_interior_mut = this
+            .machine
+            .borrow_tracker
+            .as_ref()
+            .unwrap()
+            .borrow()
```
RalfJung:
```suggestion
            .get_mut()
```
Let's avoid the run-time checks of `borrow()`, we don't need them here.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -342,35 +342,80 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        for (perm_range, perm) in perms_map.iter_mut_all() {
+            if perm.is_accessed() {
+                // Some reborrows incur a read access to the parent.
                 // Adjust range to be relative to allocation start (rather than to `place`).
-                let mut range_in_alloc = range;
+                let mut range_in_alloc = AllocRange {
+                    start: Size::from_bytes(perm_range.start),
+                    size: Size::from_bytes(perm_range.end - perm_range.start),
+                };
                 range_in_alloc.start += base_offset;
```
RalfJung: You can avoid a `let mut` here. There's no reason to first compute the wrong `start` here and then later mutate it.

RalfJung (commented):
Thanks! This is much better. I still have some comments, though.

@rustbot author

src/borrow_tracker/tree_borrows/mod.rs
yoctocell: For this to work, I had to move down

```rust
        let mut tree_borrows = alloc_extra.borrow_tracker_tb().borrow_mut();
```

from line 315.

src/bin/miri.rs
```diff
@@ -560,16 +560,17 @@ fn main() {
...
-                miri_config.borrow_tracker =
-                    Some(BorrowTrackerMethod::TreeBorrows(TreeBorrowsParams {
-                        precise_interior_mut: false,
-                    }));
-            } else {
-                show_error!(
-                    "`-Zmiri-tree-borrows` must also be set before the `-Zmiri-tree-borrows-no-precise-interior-mut` flag"
-                );
-            }
+            miri_config.borrow_tracker.as_mut().map(|method| {
```
RalfJung: This means there'll be no error if someone does `-Zmiri-disable-stacked-borrows -Zmiri-tree-borrows-no-precise-interior-mut`.

You'll have to do something like
```rust
match &mut miri_config.borrow_tracker {
  Some(BorrowTrackerMethod::TreeBorrows(ref mut params)) => ...
  _ => { /* show errror */ }
}
```

src/bin/miri.rs
```diff
@@ -560,17 +560,15 @@ fn main() {
...
-                        "`-Zmiri-tree-borrows` must also be set before the `-Zmiri-tree-borrows-no-precise-interior-mut` flag"
-                    )
+            match &mut miri_config.borrow_tracker {
+                Some(BorrowTrackerMethod::TreeBorrows(params)) => {
+                    params.precise_interior_mut = false;
                 }
-            });
+                _ =>
+                    show_error!(
+                        "`-Zmiri-tree-borrows` must also be set before the `-Zmiri-tree-borrows-no-precise-interior-mut` flag"
```
```suggestion
                        "`-Zmiri-tree-borrows` is required before `-Zmiri-tree-borrows-no-precise-interior-mut`"
```

RalfJung:

@rustbot author

PR #4383, opened 2025-06-07
use RwLockRef instead of RwLockId

LorrensP-2158466:
cc rust-lang/miri#3967 

This cleans up the RwLockId and uses a Ref instead.

### DIFF
--- a/src/concurrency/sync.rs
+++ b/src/concurrency/sync.rs
@@ -75,8 +75,6 @@ impl VisitProvenance for MutexRef {
     }
 }
 
-declare_id!(RwLockId);
-
 /// The read-write lock state.
 #[derive(Default, Debug)]
 struct RwLock {
@@ -109,6 +107,21 @@ struct RwLock {
     /// locks.
     /// This is only relevant when there is an active reader.
     clock_current_readers: VClock,
+}
+
+#[derive(Default, Clone, Debug)]
+pub struct RwLockRef(Rc<RefCell<RwLock>>);
+
+impl RwLockRef {
+    fn new() -> Self {
+        RwLockRef(Rc::new(RefCell::new(RwLock::default())))
+    }
+}
+
+impl VisitProvenance for RwLockRef {
+    fn visit_provenance(&self, _visit: &mut VisitWith<'_>) {
+        // RwLockRef contains no provenance.
+    }
 }
 
 declare_id!(CondvarId);
@@ -164,7 +177,6 @@ struct FutexWaiter {
 /// The state of all synchronization objects.
 #[derive(Default, Debug)]
 pub struct SynchronizationObjects {
-    rwlocks: IndexVec<RwLockId, RwLock>,
     condvars: IndexVec<CondvarId, Condvar>,
     pub(super) init_onces: IndexVec<InitOnceId, InitOnce>,
 }
@@ -196,8 +208,8 @@ impl SynchronizationObjects {
     pub fn mutex_create(&mut self) -> MutexRef {
         MutexRef::new()
     }
-    pub fn rwlock_create(&mut self) -> RwLockId {
-        self.rwlocks.push(Default::default())
+    pub fn rwlock_create(&mut self) -> RwLockRef {
+        RwLockRef::new()
     }
 
     pub fn condvar_create(&mut self) -> CondvarId {
@@ -447,12 +459,10 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
     #[inline]
     /// Check if locked.
-    fn rwlock_is_locked(&self, id: RwLockId) -> bool {
-        let this = self.eval_context_ref();
-        let rwlock = &this.machine.sync.rwlocks[id];
+    fn rwlock_is_locked(&self, rw_lock_ref: &RwLockRef) -> bool {
+        let rwlock = rw_lock_ref.0.borrow();
         trace!(
-            "rwlock_is_locked: {:?} writer is {:?} and there are {} reader threads (some of which could hold multiple read locks)",
-            id,
+            "rwlock_is_locked: writer is {:?} and there are {} reader threads (some of which could hold multiple read locks)",
             rwlock.writer,
             rwlock.readers.len(),
         );
@@ -461,21 +471,20 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
     /// Check if write locked.
     #[inline]
-    fn rwlock_is_write_locked(&self, id: RwLockId) -> bool {
-        let this = self.eval_context_ref();
-        let rwlock = &this.machine.sync.rwlocks[id];
-        trace!("rwlock_is_write_locked: {:?} writer is {:?}", id, rwlock.writer);
+    fn rwlock_is_write_locked(&self, rwlock_ref: &RwLockRef) -> bool {
+        let rwlock = rwlock_ref.0.borrow();
+        trace!("rwlock_is_write_locked: writer is {:?}", rwlock.writer);
         rwlock.writer.is_some()
     }
 
     /// Read-lock the lock by adding the `reader` the list of threads that own
     /// this lock.
-    fn rwlock_reader_lock(&mut self, id: RwLockId) {
+    fn rwlock_reader_lock(&mut self, rwlock_ref: &RwLockRef) {
         let this = self.eval_context_mut();
         let thread = this.active_thread();
-        assert!(!this.rwlock_is_write_locked(id), "the lock is write locked");
-        trace!("rwlock_reader_lock: {:?} now also held (one more time) by {:?}", id, thread);
-        let rwlock = &mut this.machine.sync.rwlocks[id];
+        assert!(!this.rwlock_is_write_locked(rwlock_ref), "the lock is write locked");
+        trace!("rwlock_reader_lock: now also held (one more time) by {:?}", thread);
+        let mut rwlock = rwlock_ref.0.borrow_mut();
         let count = rwlock.readers.entry(thread).or_insert(0);
         *count = count.strict_add(1);
         if let Some(data_race) = this.machine.data_race.as_vclocks_ref() {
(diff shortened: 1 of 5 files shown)


src/concurrency/sync.rs
```diff
@@ -574,68 +591,78 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    (None, std::mem::take(&mut rwlock.reader_queue))
+                }
+            } else {
+                return interp_ok(false);
             }
-            true
+        };
+
+        // We have no mutable borrow of `rwlock` anymore, so do the actual unblocking,
+        // which can trigger callbacks that use the `rwlock`.
```
LorrensP-2158466: This was the only way I didn't get any runtime borrow errors without using an explicit `drop(rwlock)`.

Also, is the comment necessary?

src/concurrency/sync.rs
```diff
@@ -485,41 +494,45 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    *count -= 1;
+                    if *count == 0 {
+                        trace!("rwlock_reader_unlock: no longer held by {:?}", thread);
+                        entry.remove();
+                    } else {
+                        trace!("rwlock_reader_unlock: held one less time by {:?}", thread);
+                    }
                 }
+                Entry::Vacant(_) => return interp_ok(false), // we did not even own this lock
+            }
```
LorrensP-2158466: Scoping instead of an explicit `drop(rwlock)`. I find it cleaner.

src/concurrency/sync.rs
```diff
@@ -111,6 +109,21 @@ struct RwLock {
     clock_current_readers: VClock,
 }
 
+#[derive(Default, Clone, Debug)]
+pub struct RwLockRef(Rc<RefCell<RwLock>>);
```
LorrensP-2158466, line 138: Would it be nicer to have a `Deref` on a Ref sync object? Because now we have to do this every time:
```rust
let rwlock = rwlock_ref.0.borrow();
```

RalfJung: `Deref` doesn't work with `RefCell`.

src/concurrency/sync.rs
```diff
@@ -485,41 +494,45 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        if this.rwlock_is_locked(rwlock_ref).not() {
             // All the readers are finished, so set the writer data-race handle to the value
             // of the union of all reader data race handles, since the set of readers
             // happen-before the writers
-            let rwlock = &mut this.machine.sync.rwlocks[id];
-            rwlock.clock_unlocked.clone_from(&rwlock.clock_current_readers);
+            let mut rwlock = rwlock_ref.0.borrow_mut();
+            let vclock = &rwlock.clock_current_readers.clone();
+            // TODO: clone needed here?
+            rwlock.clock_unlocked.clone_from(vclock);
```
RalfJung: This should be just
```suggestion
            rwlock.clock_unlocked.clone_from(&rwlock.clock_current_readers);
```
and then you need to `drop(rwlock)` since "unblock_thread" can run almost arbitrary code.

src/concurrency/sync.rs
RalfJung: I think an explicit `drop` here would be more readable than this. Please just use a `drop`, it's a fairly standard thing to do when using `RefCell`.

src/concurrency/sync.rs
```diff
@@ -461,21 +471,20 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    fn rwlock_is_write_locked(&self, rwlock_ref: &RwLockRef) -> bool {
+        let rwlock = rwlock_ref.0.borrow();
+        trace!("rwlock_is_write_locked: writer is {:?}", rwlock.writer);
         rwlock.writer.is_some()
     }
 
     /// Read-lock the lock by adding the `reader` the list of threads that own
     /// this lock.
-    fn rwlock_reader_lock(&mut self, id: RwLockId) {
+    fn rwlock_reader_lock(&mut self, rwlock_ref: RwLockRef) {
```
RalfJung: The `writer_lock` method takes `&RwLockRef`, so please do the same here for consistency.

src/shims/unix/sync.rs
```diff
@@ -702,16 +712,17 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        if this.rwlock_is_locked(&rwlock.rwlock_ref) {
             throw_ub_format!("destroyed a locked rwlock");
         }
 
         // This might lead to false positives, see comment in pthread_mutexattr_destroy
         this.write_uninit(
             &this.deref_pointer_as(rwlock_op, this.libc_ty_layout("pthread_rwlock_t"))?,
         )?;
+        // REVIEW: Is this the fixme related to the issue? I think so.
         // FIXME: delete interpreter state associated with this rwlock.
```
RalfJung: The FIXME should stay. To resolve it, we'd want to delete the "lazy_sync data" associated with this address.

src/concurrency/sync.rs
LorrensP-2158466: Aha, didn't know that, thanks 👍!

src/concurrency/sync.rs
LorrensP-2158466: I then get this error:
```bash
error[E0502]: cannot borrow `rwlock` as immutable because it is also borrowed as mutable
   --> src/concurrency/sync.rs:530:47
    |
530 |             rwlock.clock_unlocked.clone_from(&rwlock.clock_current_readers);
    |             ------                ----------  ^^^^^^ immutable borrow occurs here
    |             |                     |
    |             |                     mutable borrow later used by call
    |             mutable borrow occurs here
```

RalfJung: Yeah you need to store the reference.
```rust
let rwlock_ref = &mut *rwlock;
rwlock_ref.clock_unlocked.clone_from(&rwlock_ref.clock_current_readers);
```

LorrensP-2158466: wow what. Thanks, learned something new.

RalfJung: This is needed because `rwlock` here is a [`RefMut`](https://doc.rust-lang.org/nightly/std/cell/struct.RefMut.html). This `DerefMut`s to the type we care about, but when we deref it twice in the same statement it calls `deref_mut` twice, each time passing a full `&mut self` to `deref_mut`, and then the borrows overlap.

LorrensP-2158466:
That's weird, my lsp should run clippy because I'm using [rust_analyzer_helix.toml](https://github.com/rust-lang/miri/blob/master/etc/rust_analyzer_helix.toml). Gotta fix that and remember to still run it.

The other tests passed, so I think everything should be fine now.

@rustbot ready

src/concurrency/sync.rs
RalfJung: You forgot this part:

> and then you need to drop(rwlock) since "unblock_thread" can run almost arbitrary code.

src/concurrency/sync.rs
```diff
@@ -574,35 +588,37 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             //
             // We are prioritizing writers here against the readers. As a
             // result, not only readers can starve writers, but also writers can
             // starve readers.
             if let Some(writer) = rwlock.writer_queue.pop_front() {
-                this.unblock_thread(writer, BlockReason::RwLock(id))?;
+                drop(rwlock);
+                this.unblock_thread(writer, BlockReason::RwLock)?;
             } else {
-                // Take the entire read queue and wake them all up.
```
RalfJung, line 602: Please don't just remove comments when they are still valid.

LorrensP-2158466: Oh my, that should not happen, sorry.

src/concurrency/sync.rs
LorrensP-2158466: But `rwlock` is a `&mut RwLock`. If I do `drop(rwlock)` clippy warns me:
```rust
warning: calls to `std::mem::drop` with a reference instead of an owned value does nothing
   --> src/concurrency/sync.rs:532:17
    |
532 |                 drop(rwlock);
    |                 ^^^^^------^
    |                      |
    |                      argument has type `&mut concurrency::sync::RwLock`
    |
```

The code now looks like this:
```rust
let rwlock = &mut *rwlock_ref.0.borrow_mut();
rwlock.clock_unlocked.clone_from(&rwlock.clock_current_readers);
```

So I don't think dropping it really matters here.

RalfJung: Of course you have to drop the `RefMut`, not the mutable reference you got from it.

> But rwlock is a &mut RwLock

Not in the code as I suggested it above. There's a reason I suggested `let rwlock_ref = &mut *rwlock;`.

LorrensP-2158466: Okay, I get it now. Took some time...

src/concurrency/sync.rs
```diff
@@ -574,35 +590,38 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             // result, not only readers can starve writers, but also writers can
             // starve readers.
             if let Some(writer) = rwlock.writer_queue.pop_front() {
-                this.unblock_thread(writer, BlockReason::RwLock(id))?;
+                drop(rwlock);
+                this.unblock_thread(writer, BlockReason::RwLock)?;
             } else {
                 // Take the entire read queue and wake them all up.
                 let readers = std::mem::take(&mut rwlock.reader_queue);
+                drop(rwlock);
```
RalfJung:
```suggestion
                drop(rwlock); // make RefCell available for unblock callback
```

src/concurrency/sync.rs
```diff
@@ -574,35 +590,38 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             }
+
             // The thread was a writer.
             //
             // We are prioritizing writers here against the readers. As a
             // result, not only readers can starve writers, but also writers can
             // starve readers.
             if let Some(writer) = rwlock.writer_queue.pop_front() {
-                this.unblock_thread(writer, BlockReason::RwLock(id))?;
+                drop(rwlock);
```
RalfJung:
```suggestion
                drop(rwlock); // make RefCell available for unblock callback
```

src/concurrency/sync.rs
```diff
@@ -509,17 +518,20 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-            let rwlock = &mut this.machine.sync.rwlocks[id];
-            rwlock.clock_unlocked.clone_from(&rwlock.clock_current_readers);
+            let mut rwlock = rwlock_ref.0.borrow_mut();
+            let rwlock_ref = &mut *rwlock;
+            rwlock_ref.clock_unlocked.clone_from(&rwlock_ref.clock_current_readers);
             // See if there is a thread to unblock.
-            if let Some(writer) = rwlock.writer_queue.pop_front() {
-                this.unblock_thread(writer, BlockReason::RwLock(id))?;
+            if let Some(writer) = rwlock_ref.writer_queue.pop_front() {
+                drop(rwlock);
```
RalfJung:
```suggestion
                drop(rwlock); // make RefCell available for unblock callback
```

src/concurrency/sync.rs
```diff
@@ -509,17 +518,20 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 rwlock.clock_current_readers.join(clock)
             });
         }
+        drop(rwlock);
```
RalfJung:
```suggestion
        drop(rwlock); // FIXME can we avoid releasing and re-acquiring the RefCell?
```

RalfJung: I have some ideas for how this could be avoided but that does not have to happen in this PR.

LorrensP-2158466: Maybe in the next one, that one also closes the related issue.

RalfJung:

@rustbot author

src/concurrency/sync.rs
RalfJung: I've added a commit doing it for RwLock (and Mutex), please use that as template for condvar :)

PR #4384, opened 2025-06-07
diagnotics: do not repeat the entire message in the span label

RalfJung:
r? @oli-obk


RalfJung:
@oli-obk @saethlin what do you think about this change?

oli-obk:
yes! avoiding duplication *and* having more useful information

PR #4387, opened 2025-06-08
Include the name of the created tag in TB's Reborrow events

### LINKED ISSUE #4386: Log more info about reborrows which cause writes to become invalid under TB
There currently doesn't seem to be a good way to debug errors involving pointer invalidation due to reborrows under the tree borrows model, i.e. errors of the following form:
```
error: Undefined Behavior: write access through <31989684> at alloc18827939[0x10] is forbidden
...

     = help: the accessed tag <31989684> is a child of the conflicting tag <31895995>
     = help: the conflicting tag <31895995> has state Frozen which forbids this child write access
help: the accessed tag <31989684> was created here
...
help: the conflicting tag <31895995> later transitioned to Frozen due to a reborrow (acting as a foreign read access) at offsets [0x0..0x30]
     ...
     = help: this transition corresponds to a loss of write permissions
```

miri prints a backtrace of the problematic write, but only displays the line of code at which the the problematic reborrow happened. Notably, it *doesnt* print the tag created by the reborrow. `-Zmiri-track-pointer-tag` doesn't log anything for the reborrow event either.

I've hit a few instances of this error recently and had to resort to writing one-off repro tests and fussing around with breakpoints in my debugger to try to catch the reborrow event manually.

I'd be happy if the error logged the tag created by the reborrow, so that I could then get a backtrace with `-Zmiri-track-pointer-tag`. It might also make sense to just print the full backtrace of the reborrow in the error - I needed to know both the write backtrace and the reborrow backtrace in every case that I personally hit.

CC:@saethlin

saethlin:
This fixes https://github.com/rust-lang/miri/issues/4386 by recording and mentioning the created tag in a Reborrow event.

### DIFF
--- a/src/borrow_tracker/tree_borrows/diagnostics.rs
+++ b/src/borrow_tracker/tree_borrows/diagnostics.rs
@@ -15,7 +15,7 @@ use crate::*;
 #[derive(Clone, Copy, Debug)]
 pub enum AccessCause {
     Explicit(AccessKind),
-    Reborrow,
+    Reborrow(BorTag),
     Dealloc,
     FnExit(AccessKind),
 }
@@ -24,7 +24,7 @@ impl fmt::Display for AccessCause {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             Self::Explicit(kind) => write!(f, "{kind}"),
-            Self::Reborrow => write!(f, "reborrow"),
+            Self::Reborrow(tag) => write!(f, "reborrow to create {tag:?}"),
             Self::Dealloc => write!(f, "deallocation"),
             // This is dead code, since the protector release access itself can never
             // cause UB (while the protector is active, if some other access invalidates
@@ -40,7 +40,8 @@ impl AccessCause {
         let rel = if is_foreign { "foreign" } else { "child" };
         match self {
             Self::Explicit(kind) => format!("{rel} {kind}"),
-            Self::Reborrow => format!("reborrow (acting as a {rel} read access)"),
+            Self::Reborrow(tag) =>
+                format!("reborrow to create {tag:?} (acting as a {rel} read access)"),
             Self::Dealloc => format!("deallocation (acting as a {rel} write access)"),
             Self::FnExit(kind) => format!("protector release (acting as a {rel} {kind})"),
         }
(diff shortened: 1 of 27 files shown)


src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -15,7 +15,7 @@ use crate::*;
 #[derive(Clone, Copy, Debug)]
 pub enum AccessCause {
     Explicit(AccessKind),
-    Reborrow,
+    Reborrow(BorTag),
```
RalfJung, line 18:
```suggestion
    /// The tag here is the one *newly created* by the reborrow that caused the access.
    Reborrow(BorTag),
```

tests/fail/both_borrows/pass_invalid_shr_option.tree.stderr
```diff
@@ -1,11 +1,11 @@
-error: Undefined Behavior: reborrow through <TAG> at ALLOC[0x0] is forbidden
+error: Undefined Behavior: reborrow to create <TAG> through <TAG> at ALLOC[0x0] is forbidden
```
RalfJung, line 1: The new tag doesn't matter here, does it? Creating it failed, so it can't possibly occur anywhere else in the state already.

tests/fail/async-shared-mutable.tree.stderr
```diff
@@ -21,7 +21,7 @@ help: the accessed tag <TAG> later transitioned to Active due to a child write a
 LL |             *x = 1;
    |             ^^^^^^
    = help: this transition corresponds to the first write to a 2-phase borrowed mutable reference
-help: the accessed tag <TAG> later transitioned to Frozen due to a reborrow (acting as a foreign read access) at offsets [OFFSET]
+help: the accessed tag <TAG> later transitioned to Frozen due to a reborrow to create <TAG> (acting as a foreign read access) at offsets [OFFSET]
```
RalfJung, line 24: Is it clear that the newly created tag is always the more interesting one here, rather than the tag of the pointer that is the input to the reborrow?

JoJoDeveloping: Yes, because it can be traced with `-Zmiri-track-pointer-tag` (which tells you the parent tag), but you can't find the child tag from the parent.

saethlin: Yeah, that's why the alternative solution to the linked issue is to make tag tracking also log reborrows of the tracked tag. But I suspect that will be an unusable amount of information written to the terminal.

tests/fail/both_borrows/pass_invalid_shr_option.tree.stderr
RalfJung: If the above is correct then I don't think the new tag should be shown here.

RalfJung:
@saethlin do you plan to get back to this any time soon?
If not, let's close the PR to better reflect its status.

saethlin:
I thought I closed this a while ago :facepalm:

PR #4396, opened 2025-06-12
Allow building Miri with --features from miri-script

Stypox:
I added the `--features` flag to all `Command`s for which `cargo` can accept `--features`, i.e. install, build, check, test, run, doc. Those who can't accept `--features` are clippy, fmt, toolchain, rustc pull, rustc push, squash (the last 4 are not even cargo commands). Furthermore, I did not add it to the `Bench` command since that command doesn't take any "--flags" either. 

This PR is needed because there is no way to pass e.g. `--features tracing` just to the `cargo` commands issued on the root repository. Unfortunately `CARGO_EXTRA_FLAGS` applies the flags to the "cargo-miri" crate, too, which does not make sense for crate-specific features.

This PR also includes a small commit to make the interface of `install_to_sysroot` clearer.

Note: this PR is the counterpart to https://github.com/rust-lang/rust/pull/142379, which is for when Miri is built in the rustc repo.

### DIFF
--- a/miri-script/src/commands.rs
+++ b/miri-script/src/commands.rs
@@ -32,6 +32,7 @@ impl MiriEnv {
         &mut self,
         quiet: bool,
         target: Option<impl AsRef<OsStr>>,
+        features: &[String],
     ) -> Result<PathBuf> {
         if let Some(miri_sysroot) = self.sh.var_os("MIRI_SYSROOT") {
             // Sysroot already set, use that.
@@ -39,8 +40,8 @@ impl MiriEnv {
         }
 
         // Make sure everything is built. Also Miri itself.
-        self.build(".", &[], quiet)?;
-        self.build("cargo-miri", &[], quiet)?;
+        self.build(".", features, &[], quiet)?;
+        self.build("cargo-miri", &[], &[], quiet)?;
 
         let target_flag = if let Some(target) = &target {
             vec![OsStr::new("--target"), target.as_ref()]
@@ -58,7 +59,7 @@ impl MiriEnv {
         }
 
         let mut cmd = self
-            .cargo_cmd("cargo-miri", "run")
+            .cargo_cmd("cargo-miri", "run", &[])
             .arg("--quiet")
             .arg("--")
             .args(&["miri", "setup", "--print-sysroot"])
@@ -90,7 +91,9 @@ impl Command {
             Self::fmt(vec![])?;
         }
         if auto_clippy {
-            Self::clippy(vec![])?;
+            // no features for auto actions, see
+            // https://github.com/rust-lang/miri/pull/4396#discussion_r2149654845
+            Self::clippy(vec![], vec![])?;
         }
 
         Ok(())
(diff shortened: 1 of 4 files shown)


output.tracy
RalfJung, line 1: What is this file about?

miri-script/src/main.rs
```diff
@@ -144,10 +168,10 @@ impl Command {
...
-            | Self::Build { flags }
-            | Self::Check { flags }
-            | Self::Doc { flags }
+            Self::Install { flags, .. }
+            | Self::Build { flags, .. }
+            | Self::Check { flags, .. }
+            | Self::Doc { flags, .. }
             | Self::Fmt { flags }
             | Self::Toolchain { flags }
             | Self::Clippy { flags }
```
RalfJung: I think it'd make sense to also support this with clippy.

miri-script/src/commands.rs
```diff
@@ -175,14 +176,14 @@ impl Command {
...
-            Command::Test { bless, flags, target, coverage } =>
-                Self::test(bless, flags, target, coverage),
-            Command::Run { dep, verbose, target, edition, flags } =>
-                Self::run(dep, verbose, target, edition, flags),
-            Command::Doc { flags } => Self::doc(flags),
+            Command::Install { features, flags } => Self::install(features, flags),
+            Command::Build { features, flags } => Self::build(features, flags),
+            Command::Check { features, flags } => Self::check(features, flags),
+            Command::Test { bless, flags, target, coverage, features } =>
+                Self::test(bless, flags, target, coverage, features),
```
RalfJung: For consistency, please also stick to `features, flags` here. Might make sense to reorder the `test(...)` arguments to match the order in which the fields are declared in the enum.

miri-script/src/main.rs
Stypox: I used this to check which cargo commands can take --features but cargo clippy instead had a line pointing to `check`. Fixed now.
```sh
for i in install build check clippy test run doc fmt bench toolchain; do echo $i; cargo $i --help | grep features; done
 ```

File: miri-script/src/commands.rs [hunk 3]
```diff
@@ -159,14 +160,14 @@ impl Command {
...
-            | Command::Doc { .. }
-            | Command::Clippy { .. } => Self::auto_actions()?,
+            Command::Install { features, .. }
+            | Command::Build { features, .. }
+            | Command::Check { features, .. }
+            | Command::Test { features, .. }
+            | Command::Run { features, .. }
+            | Command::Doc { features, .. }
+            | Command::Clippy { features, .. } => Self::auto_actions(features.clone())?,
+            | Command::Fmt { .. } => Self::auto_actions(vec![])?,
```
Stypox: `Fmt` does not have features, however the features might be needed for some auto action. Should I add `features` to `Fmt` too, and expand its doc to explain how it's not actually used for `Fmt` but rather for auto actions?

Stypox (commented):
Thanks for the review! I force pushed to remove the unrelated file (I made the commit a few weeks ago and cherry picked it a few times, and that file was committed accidentally at some point, sorry for not noticing). I also pushed a commit to add --features to clippy and to auto_actions.

miri-script/src/commands.rs
RalfJung: Adding this to `fmt` just for auto-clippy doesn't make sense IMO. I don't think auto-clippy should enable any features... but I don't use auto-clippy anyway. @oli-obk what do you think?

oli-obk: auto-clippy should be indistinguishable from `./miri clippy`, so no features, yes.

Stypox: Ok, I removed `features` from auto-actions altogether, thanks for the input

RalfJung:
I assume this is
@rustbot ready
?

RalfJung:

@rustbot author

PR #4397, opened 2025-06-13
Add shims for `gettid`-esque functions

tgross35:
Various platforms provide a function to return the current OS thread ID, but they all use a slightly different name. Add shims for these functions for Apple, FreeBSD, and Windows, with tests to account for a few more platforms that are not yet supported by Miri.

These should be useful in general but will also help support printing the OS thread ID in panic messages [1].

[1]: https://github.com/rust-lang/rust/pull/115746

### DIFF
--- a/src/shims/env.rs
+++ b/src/shims/env.rs
@@ -110,8 +110,30 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         }
     }
 
+    /// Get the process identifier.
     fn get_pid(&self) -> u32 {
         let this = self.eval_context_ref();
         if this.machine.communicate() { std::process::id() } else { 1000 }
+    }
+
+    /// Get an "OS" thread ID for the current thread.
+    fn get_current_tid(&self) -> u32 {
+        let this = self.eval_context_ref();
+        self.get_tid(this.machine.threads.active_thread())
+    }
+
+    /// Get an "OS" thread ID for any thread.
+    fn get_tid(&self, thread: ThreadId) -> u32 {
+        let this = self.eval_context_ref();
+        let index = thread.to_u32();
+        let target_os = &this.tcx.sess.target.os;
+        if target_os == "linux" || target_os == "netbsd" {
+            // On Linux, the main thread has PID == TID so we uphold this. NetBSD also appears
+            // to exhibit the same behavior, though I can't find a citation.
+            this.get_pid().strict_add(index)
+        } else {
+            // Other platforms do not display any relationship between PID and TID.
+            index
+        }
     }
 }
--- a/src/shims/extern_static.rs
+++ b/src/shims/extern_static.rs
@@ -66,7 +66,7 @@ impl<'tcx> MiriMachine<'tcx> {
                     ecx,
                     &["__cxa_thread_atexit_impl", "__clock_gettime64"],
                 )?;
-                Self::weak_symbol_extern_statics(ecx, &["getrandom", "statx"])?;
+                Self::weak_symbol_extern_statics(ecx, &["getrandom", "gettid", "statx"])?;
             }
             "freebsd" => {
                 Self::null_ptr_extern_statics(ecx, &["__cxa_thread_atexit_impl"])?;
--- a/src/shims/unix/env.rs
+++ b/src/shims/unix/env.rs
@@ -274,15 +274,52 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         interp_ok(Scalar::from_u32(this.get_pid()))
     }
 
-    fn linux_gettid(&mut self) -> InterpResult<'tcx, Scalar> {
+    /// The `gettid`-like function for Unix platforms that take no parameters and return a 32-bit
+    /// integer. It is not always named "gettid".
+    fn unix_gettid(&mut self, link_name: &str) -> InterpResult<'tcx, Scalar> {
         let this = self.eval_context_ref();
-        this.assert_target_os("linux", "gettid");
+        this.assert_target_os_is_unix(link_name);
 
-        let index = this.machine.threads.active_thread().to_u32();
+        // For most platforms the return type is an `i32`, but some are unsigned. The TID
+        // will always be positive so we don't need to differentiate.
+        interp_ok(Scalar::from_u32(this.get_current_tid()))
+    }
 
-        // Compute a TID for this thread, ensuring that the main thread has PID == TID.
-        let tid = this.get_pid().strict_add(index);
+    /// The Apple-specific `int pthread_threadid_np(pthread_t thread, uint64_t *thread_id)`, which
+    /// allows querying the ID for arbitrary threads, identified by their pthread_t.
+    ///
+    /// API documentation: <https://www.manpagez.com/man/3/pthread_threadid_np/>.
+    fn apple_pthread_threadip_np(
+        &mut self,
+        thread_op: &OpTy<'tcx>,
+        tid_op: &OpTy<'tcx>,
+    ) -> InterpResult<'tcx, Scalar> {
+        let this = self.eval_context_mut();
+        this.assert_target_os("macos", "pthread_threadip_np");
 
-        interp_ok(Scalar::from_u32(tid))
+        let tid_dest = this.read_pointer(tid_op)?;
+        if this.ptr_is_null(tid_dest)? {
+            // If NULL is passed, an error is immediately returned
+            return interp_ok(this.eval_libc("EINVAL"));
+        }
+
+        let thread = this.read_scalar(thread_op)?.to_int(this.libc_ty_layout("pthread_t").size)?;
+        let thread = if thread == 0 {
+            // Null thread ID indicates that we are querying the active thread.
+            this.machine.threads.active_thread()
+        } else {
+            // Our pthread_t is just the raw ThreadId.
+            let Ok(thread) = this.thread_id_try_from(thread) else {
+                return interp_ok(this.eval_libc("ESRCH"));
+            };
+            thread
+        };
+
+        let tid = this.get_tid(thread);
+        let tid_dest = this.deref_pointer_as(tid_op, this.machine.layouts.u64)?;
+        this.write_int(tid, &tid_dest)?;
+
+        // Possible errors have been handled, return success.
+        interp_ok(Scalar::from_u32(0))
     }
 }
--- a/src/shims/unix/linux_like/syscall.rs
+++ b/src/shims/unix/linux_like/syscall.rs
@@ -4,6 +4,7 @@ use rustc_span::Symbol;
 use rustc_target::callconv::FnAbi;
 
 use crate::helpers::check_min_vararg_count;
+use crate::shims::unix::env::EvalContextExt;
 use crate::shims::unix::linux_like::eventfd::EvalContextExt as _;
 use crate::shims::unix::linux_like::sync::futex;
 use crate::*;
@@ -24,6 +25,7 @@ pub fn syscall<'tcx>(
     let sys_getrandom = ecx.eval_libc("SYS_getrandom").to_target_usize(ecx)?;
     let sys_futex = ecx.eval_libc("SYS_futex").to_target_usize(ecx)?;
     let sys_eventfd2 = ecx.eval_libc("SYS_eventfd2").to_target_usize(ecx)?;
+    let sys_gettid = ecx.eval_libc("SYS_gettid").to_target_usize(ecx)?;
 
     match ecx.read_target_usize(op)? {
         // `libc::syscall(NR_GETRANDOM, buf.as_mut_ptr(), buf.len(), GRND_NONBLOCK)`
@@ -52,6 +54,10 @@ pub fn syscall<'tcx>(
 
             let result = ecx.eventfd(initval, flags)?;
             ecx.write_int(result.to_i32()?, dest)?;
+        }
+        num if num == sys_gettid => {
+            let result = ecx.unix_gettid("SYS_gettid")?;
+            ecx.write_int(result.to_u32()?, dest)?;
         }
         num => {
             throw_unsup_format!("syscall: unsupported syscall number {num}");
--- a/src/shims/windows/foreign_items.rs
+++ b/src/shims/windows/foreign_items.rs
@@ -629,6 +629,23 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.write_scalar(name, &name_ptr)?;
                 this.write_scalar(res, dest)?;
             }
+            "GetThreadId" => {
+                let [handle] = this.check_shim(abi, sys_conv, link_name, args)?;
+                let handle = this.read_handle(handle, "GetThreadId")?;
+                let thread = match handle {
+                    Handle::Thread(thread) => thread,
+                    Handle::Pseudo(PseudoHandle::CurrentThread) => this.active_thread(),
+                    _ => this.invalid_handle("GetThreadDescription")?,
+                };
+                let tid = this.get_tid(thread);
+                this.write_scalar(Scalar::from_u32(tid), dest)?;
+            }
+            "GetCurrentThreadId" => {
+                let [] = this.check_shim(abi, sys_conv, link_name, args)?;
+                let thread = this.active_thread();
+                let tid = this.get_tid(thread);
+                this.write_scalar(Scalar::from_u32(tid), dest)?;
+            }
 
             // Miscellaneous
             "ExitProcess" => {
--- a/tests/pass-dep/shims/gettid.rs
+++ b/tests/pass-dep/shims/gettid.rs
@@ -1,0 +1,183 @@
+//! Test for `gettid` and similar functions for retrieving an OS thread ID.
+//@ revisions: with_isolation without_isolation
+//@ [without_isolation] compile-flags: -Zmiri-disable-isolation
+
+#![feature(linkage)]
+
+fn gettid() -> u64 {
+    cfg_if::cfg_if! {
+        if #[cfg(any(target_os = "android", target_os = "linux"))] {
+            gettid_linux_like()
+        } else if #[cfg(target_os = "nto")] {
+            unsafe { libc::gettid() as u64 }
+        } else if #[cfg(target_os = "openbsd")] {
+            unsafe { libc::getthrid() as u64 }
+        } else if #[cfg(target_os = "freebsd")] {
+            unsafe { libc::pthread_getthreadid_np() as u64 }
+        } else if #[cfg(target_os = "netbsd")] {
+            unsafe { libc::_lwp_self() as u64 }
+        } else if #[cfg(any(target_os = "solaris", target_os = "illumos"))] {
+            // On Solaris and Illumos, the `pthread_t` is the OS TID.
+            unsafe { libc::pthread_self() as u64 }
+        } else if #[cfg(target_vendor = "apple")] {
+            let mut id = 0u64;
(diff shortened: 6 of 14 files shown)


src/shims/env.rs
```diff
@@ -110,8 +110,24 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+    /// Get an "OS" thread ID for any thread.
+    fn get_tid(&self, thread: ThreadId) -> u32 {
+        let this = self.eval_context_ref();
+        let index = thread.to_u32();
+
+        // todo: is this also true on Windows?
+        // Compute a TID for this thread, ensuring that the main thread has PID == TID.
+        this.get_pid().strict_add(index)
+    }
```
tgross35: @ChrisDenton is there any relationship between the PID and TIDs on Windows?

This (preexisting) logic works for Unix because the first TID for a process is the same as the PID, but if that logic isn't true on Windows then I'll need to update this. If it is true on Windows, then I think the `CreateThread` shim needs to get updated to return the same TID logic in `lpThreadId` (hence the failing test).

ChrisDenton: On Windows thread IDs and process IDs are unrelated.

RalfJung: But it doesn't hurt to use the Unix scheme on Windows as well I guess?

tgross35: Great, that makes things easier. Thanks.

tgross35: Oops, I replied before seeing your message Ralf. Apparently it seems like a Linux thing more specifically than a Unix thing, I can't find documentation for other platforms about a relationship between PID and TID.

I just updated to only offset by PID on Linux, but do you prefer I put it back on all platforms and update the ID saved by `CreateThread` instead?

tgross35: Just checked a handful of different platforms on cfarm:

Linux

```text
main tid 2212321 0x21c1e1
main pid 2212321 0x21c1e1
```

FreeBSD: 

```text
main tid 0 0
main pid 23750 0x5cc6
```

OpenBSD:

```text
main tid 137368 0x21898
main pid 93069 0x16b8d
```

NetBSD:

```text
main tid 5057 0x13c1
main pid 5057 0x13c1
```

NetBSD checking a spawned thread

```text
in spawn: pid 1474 0x5c2 tid 13912 0x3658
in main:  pid 1474 0x5c2 tid 1474 0x5c2
```

MacOS:

```text
main tid 7763176 0x7674e8
main pid 51507 0xc933
```

Based on that nonscientific experiment I'll add NetBSD to the list of platforms that says tid is an offset from pid. For the others it's probably actually better that they don't line up, so users don't rely on that non-cross-platform behavior.

tgross35:
Updated to include the `SYS_gettid` syscall

src/shims/unix/linux_like/syscall.rs
```diff
@@ -53,6 +55,11 @@ pub fn syscall<'tcx>(
             let result = ecx.eventfd(initval, flags)?;
             ecx.write_int(result.to_i32()?, dest)?;
         }
+        num if num == sys_gettid => {
+            let [] = check_exact_vararg_count("syscall(SYS_gettid, ...)", varargs)?;
```
tgross35: The other syscalls here seem to allow more than the required number of arguments. Should this just do the same and disregard arguments?

On one hand, it can't be UB to call varidics with more arguments than they expect, unless the function you're calling is doing something sadistic with asm. On the other hand, if you are passing more arguments here, you're doing something outside of the documented API.

tgross35:
Rebased because of the test failures but that didn't help; I don't know why this seems to have an effect on `alloc-access-tracking.rs`.

src/shims/unix/linux_like/syscall.rs
RalfJung: Please remain consistent with the other variadic functions -- I think last time my conclusion was that we should never error when there are extra arguments, for anything variadic, so something like `check_exact_vararg_count` deliberately does not exist.

src/shims/extern_static.rs
```diff
@@ -66,7 +66,7 @@ impl<'tcx> MiriMachine<'tcx> {
                     ecx,
                     &["__cxa_thread_atexit_impl", "__clock_gettime64"],
                 )?;
-                Self::weak_symbol_extern_statics(ecx, &["getrandom", "statx"])?;
+                Self::weak_symbol_extern_statics(ecx, &["getrandom", "gettid", "statx"])?;
```
RalfJung, line 69: You're adding a new global variable here, that's why alloc-access-tracking changes. You'll have to increment the ID there by 1.

src/shims/unix/env.rs
```diff
@@ -275,15 +276,56 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        // Various platforms have a similar function with different names.
+        match (target_os.as_ref(), link_name.as_str()) {
+            ("linux" | "android", "gettid") => (),
+            ("openbsd", "getthrid") => (),
+            ("freebsd", "pthread_getthreadid_np") => (),
+            ("netbsd", "_lwp_self") => (),
+            (_, name) => panic!("`{name}` is not supported on {target_os}"),
+        };
+
```
RalfJung: I don't think it's worth having such a table, this should be handled by the dispatch logic before already. In here we should just assert that this is truly a unix platform.

src/shims/env.rs
```diff
@@ -110,8 +110,29 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
 
+    /// Get the process identifier.
     fn get_pid(&self) -> u32 {
         let this = self.eval_context_ref();
         if this.machine.communicate() { std::process::id() } else { 1000 }
     }
+
+    /// Get an "OS" thread ID for the current thread.
+    fn get_current_tid(&self) -> u32 {
+        self.get_tid(self.eval_context_ref().machine.threads.active_thread())
```
RalfJung: Please follow the usual pattern.
```suggestion
        let this = self.eval_context_ref();
        this.get_tid(this.machine.threads.active_thread())
```

src/shims/unix/env.rs
```diff
@@ -275,15 +276,56 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    fn apple_pthread_threadip_np(
+        &mut self,
+        thread_op: &OpTy<'tcx>,
+        tid_op: &OpTy<'tcx>,
+    ) -> InterpResult<'tcx, Scalar> {
+        let this = self.eval_context_mut();
+
+        let target_vendor = &this.tcx.sess.target.vendor;
+        assert_eq!(
+            target_vendor, "apple",
```
RalfJung: We usually query the OS, not the vendor... so please remain consistent and use `assert_target_os`, there's no point in having different logic for this in different places.

src/shims/unix/env.rs
```diff
@@ -275,15 +276,56 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        };
+
+        // For most platforms the return type is an `i32`, but some are unsigned. The TID
+        // will always be positive so we don't need to differentiate.
+        interp_ok(Scalar::from_u32(this.get_current_tid()))
+    }
 
-        let index = this.machine.threads.active_thread().to_u32();
+    /// The Apple-specific `int pthread_threadid_np(pthread_t thread, uint64_t *thread_id)`, which
+    /// allows querying the ID for arbitrary threads.
```
RalfJung:
```suggestion
    /// allows querying the ID for arbitrary threads identified by their pthread_t.
```

src/shims/unix/env.rs
```diff
@@ -275,15 +276,56 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            target_vendor, "apple",
+            "`pthread_threadid_np` is not supported on target vendor {target_vendor}",
+        );
+
+        let thread = this.read_scalar(thread_op)?.to_int(this.libc_ty_layout("pthread_t").size)?;
+        let thread = if thread == 0 {
+            // Null thread ID indicates that we are querying the active thread.
+            this.machine.threads.active_thread()
+        } else {
+            let Ok(thread) = this.thread_id_try_from(thread) else {
```
RalfJung, line 312:
```suggestion
            // Our pthread_t is just the raw ThreadId.
            let Ok(thread) = this.thread_id_try_from(thread) else {
```

src/shims/unix/env.rs
```diff
@@ -275,15 +276,56 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                return interp_ok(this.eval_libc("ESRCH"));
+            };
+            thread
+        };
 
-        // Compute a TID for this thread, ensuring that the main thread has PID == TID.
-        let tid = this.get_pid().strict_add(index);
+        let tid_dest = this.deref_pointer_as(tid_op, this.machine.layouts.u64)?;
+        this.write_int(this.get_tid(thread), &tid_dest)?;
```
RalfJung: Please let-bind the `this.get_tid(thread)`, so that the core logic is a bit easier to spot among the boilerplate.

src/shims/unix/linux_like/syscall.rs
```diff
@@ -53,6 +55,11 @@ pub fn syscall<'tcx>(
...
+            let result = ecx.unix_gettid(Symbol::intern("gettid"))?;
```
RalfJung:
```suggestion
            let result = ecx.unix_gettid(Symbol::intern("SYS_gettid"))?;
```

tests/pass-dep/shims/gettid.rs
```diff
@@ -0,0 +1,167 @@
...
+        0
+    }
+
+    fn thread_body() {
+        SPAWNED_TID.store(gettid(), Ordering::Relaxed);
+        let sleep_duration = time::Duration::from_millis(10);
+
+        // Spin until the main thread has a chance to read this thread's ID
+        while !CAN_JOIN.load(Ordering::Relaxed) {
+            thread::sleep(sleep_duration);
```
RalfJung: Just use `yield_now` here, so need to have arbitrary sleep durations.

tests/pass-dep/shims/gettid.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    assert_eq!(from_libc, from_sys);
+    assert_eq!(from_libc, from_static);
+
+    from_libc
+}
+
+/// Specific platforms can query the tid of arbitrary threads; test that here.
+#[cfg(any(target_vendor = "apple", windows))]
+mod queried {
```
RalfJung: Why does this use the low-level OS APIs for thread creation, instead of the Rust std APIs?

tests/pass-dep/shims/gettid.rs
```diff
@@ -0,0 +1,167 @@
+//@ ignore-target: illumos solaris wasm
```
RalfJung:
```suggestion
//@ ignore-target: illumos solaris
```
We don't run most tests on wasm anyway.

Why are illumos and solaris skipped? That should have a comment at least.
(These are not primary-support targets, so it's okay to leave them until later... except if the planned change would break all panics on those targets, then it'd be better to do something or else our CI will break. We have some people to ping who said they'd help with illumos/solaris issues, if you'd like.)

RalfJung (commented):
Lots of little comments :)
@rustbot author

src/shims/unix/env.rs
tgross35: I had this because it does seem to be supported on iOS and other Apple targets, but I suppose they probably aren't otherwise supported anyway :)

tests/pass-dep/shims/gettid.rs
tgross35: Windows and apple have a way to get the thread ID of threads other than the current one, which is being tested here. But to do this you need the `pthread_t` or the `HANDLE`, which as far as I know we can't get from `std`.

I'll comment this a bit better

src/shims/unix/env.rs
RalfJung: Yeah the only apple target we support atm is macOS.

tests/pass-dep/shims/gettid.rs
tgross35: I originally thought they didn't have a function equivalent to `gettid` (returning OS TID), but on the r-l/r thread the maintainers confirmed that the value returned by `pthread_self` is the OS TID on those platforms. So I'll just expand the tests.

src/shims/extern_static.rs
tgross35: It seems I needed to increment by 2, not sure if that is expected

tgross35:
I believe I addressed all of the concerns. I also updated the Windows and Mac implementations (and tests) since they return gracefully rather than crashing with null or (on Windows) invalid pointers.

@rustbot ready

tests/pass-dep/shims/gettid.rs
```diff
@@ -0,0 +1,187 @@
...
+        let spawned_tid_from_handle = spawn_update_join();
+        let spawned_tid_from_self = SPAWNED_TID.load(Ordering::Relaxed);
+        let current_tid = gettid();
+
+        // Ensure that we got a different thread ID.
+        assert_ne!(spawned_tid_from_handle, current_tid);
+
+        // Ensure that we get the same result from `gettid` and from querying a thread's handle
+        assert_ne!(spawned_tid_from_handle, 0);
+        assert_eq!(spawned_tid_from_handle, spawned_tid_from_self);
```
RalfJung: The comment refers to the 2nd check. What's the 0 check about?

src/shims/unix/env.rs
```diff
@@ -310,6 +310,12 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             "`pthread_threadid_np` is not supported on target vendor {target_vendor}",
         );
 
+        let tid_dest = this.read_pointer(tid_op)?;
+        if this.ptr_is_null(tid_dest)? {
+            // If NULL is passed, an error is immediately returned
```
RalfJung, line 302: Is this documented behavior or just observed?

src/shims/unix/env.rs
```diff
@@ -274,15 +274,50 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         interp_ok(Scalar::from_u32(this.get_pid()))
     }
 
-    fn linux_gettid(&mut self) -> InterpResult<'tcx, Scalar> {
+    /// The `gettid`-like function for Unix platforms that take no parameters and return a 32-bit
+    /// integer. It is not always named "gettid".
+    fn unix_gettid(&mut self, link_name: &str) -> InterpResult<'tcx, Scalar> {
         let this = self.eval_context_ref();
-        this.assert_target_os("linux", "gettid");
+        this.assert_target_os_is_unix(link_name.as_ref());
```
RalfJung:
```suggestion
        this.assert_target_os_is_unix(link_name);
```
This should also work, doesn't it?

src/shims/windows/foreign_items.rs
```diff
@@ -631,19 +632,26 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             "GetThreadId" => {
                 let [handle] = this.check_shim(abi, sys_conv, link_name, args)?;
-                let handle = this.read_handle(handle, "GetThreadId")?;
 
-                let thread = match handle {
-                    Handle::Thread(thread) => thread,
-                    Handle::Pseudo(PseudoHandle::CurrentThread) => this.active_thread(),
-                    _ => this.invalid_handle("GetThreadDescription")?,
+                let handle = this.read_scalar(handle)?;
+                let tid = match Handle::try_from_scalar(handle, this)? {
```
RalfJung: The expectation is that you'll usually call read_handle instead (so `try_from_scalar` can remain private). That's what you used to do and now you changed it... why? At the very least this needs a long comment explaining why the consistent behavior `read_handle` is intended to enforce everywhere is somehow not right here.

tests/pass-dep/shims/gettid.rs
```diff
@@ -133,7 +133,8 @@ mod queried {
             let res = WaitForSingleObject(handle, INFINITE);
             assert_ne!(res, WAIT_FAILED, "failed to join thread");
 
-            invalid_handle_tid = GetThreadId(usize::MAX as HANDLE);
+            // Ensure we return 0 for invalid threads rather than crashing
```
RalfJung: We have generally taken the stance that we'll detect invalid handles early, even if Windows doesn't do that. The consensus was that invalid handles are an error and we should report them.

src/shims/extern_static.rs
RalfJung: Ah right, there's two new allocations: one to hold the static `gettid`, and one that represent the function which that static points to.

src/shims/unix/env.rs
tgross35: The possible error codes are documented https://www.manpagez.com/man/3/pthread_threadid_np/, the priority (null check before valid thread check) is observed

RalfJung: That's apparently a BSD manpage, not an Apple one... but it's better than no reference. Would be good to have some sort of link somewhere in the function for what we used as reference.

tests/pass-dep/shims/gettid.rs
tgross35: In Windows it indicates the error condition, but it's already checked in the relevant function. It still shouldn't happen on Unix but the check probably isn't very useful there, so I've remove this.

src/shims/unix/env.rs
tgross35: Indeed, I don't know why I thought it didn't like that...

src/shims/unix/env.rs
tgross35: It is from MacOS (see the table at the very bottom). Source for reference https://github.com/apple-oss-distributions/libpthread/blob/1ebf56b3a702df53213c2996e5e128a535d2577e/src/pthread.c#L1118-L1146 (there is also a manpage in that repo)

src/shims/windows/foreign_items.rs
tgross35: This is to match the documented behavior of returning 0 on failure https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getthreadid. But with https://github.com/rust-lang/miri/pull/4397#discussion_r2173773555, this can be dropped.

tests/pass-dep/shims/gettid.rs
tgross35: Is this Windows-only or is something similar done for invalid `pthread_t`s on other platforms?

src/shims/unix/env.rs
RalfJung: > It is from MacOS (see the table at the very bottom). 

Ah, I missed that. Please link to that then :)

tests/pass-dep/shims/gettid.rs
RalfJung: This is specifically for windows handles.

tgross35:
Switched back to the old Windows behavior and the other small things, on to the next round

@rustbot review

RalfJung:
This looks great, thanks! I have pushed a commit with a small comment tweak. Please squash the commits using `./miri squash`. Then write `@rustbot ready` after you force-pushed the squashed PR.

@rustbot author

tgross35:
Thanks!

RalfJung:
We just did a rustc-push sync, so ordinarily it would take around a week until this one gets synced. Let me know if something is blocked on this PR then I can do an earlier sync.

tgross35:
https://github.com/rust-lang/rust/pull/115746 still has ~6 days left on fcp so there isn't really any rush here, but thank you for offering

PR #4401, opened 2025-06-15
trace: add barebones ptrace setup

nia-e:
Per the review in rust-lang/miri#4326, this implements the absolute bare minimum ptrace bits; no access is actually intercepted or logged, but when running in native-lib mode Miri will now fork itself and trace the child process unless this is disabled via CLI flag. Currently the docs are left the same as in the original PR i.e. they might reference things that aren't implemented yet, but I can change that.

### DIFF
--- a/src/alloc/isolated_alloc.rs
+++ b/src/alloc/isolated_alloc.rs
@@ -266,6 +266,18 @@ impl IsolatedAlloc {
             alloc::dealloc(ptr, layout);
         }
     }
+
+    /// Returns a vector of page addresses managed by the allocator.
+    pub fn pages(&self) -> Vec<usize> {
+        let mut pages: Vec<_> =
+            self.page_ptrs.clone().into_iter().map(|p| p.expose_provenance()).collect();
+        for (ptr, size) in &self.huge_ptrs {
+            for i in 0..size / self.page_size {
+                pages.push(ptr.expose_provenance().strict_add(i * self.page_size));
+            }
+        }
+        pages
+    }
 }
 
 #[cfg(test)]
--- a/src/bin/miri.rs
+++ b/src/bin/miri.rs
@@ -227,10 +227,11 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
         } else {
             let return_code = miri::eval_entry(tcx, entry_def_id, entry_type, &config, None)
                 .unwrap_or_else(|| {
+                    #[cfg(target_os = "linux")]
+                    miri::register_retcode_sv(rustc_driver::EXIT_FAILURE);
                     tcx.dcx().abort_if_errors();
                     rustc_driver::EXIT_FAILURE
                 });
-
             std::process::exit(return_code);
         }
 
@@ -722,6 +723,8 @@ fn main() {
             } else {
                 show_error!("-Zmiri-native-lib `{}` does not exist", filename);
             }
+        } else if arg == "-Zmiri-force-old-native-lib-mode" {
+            miri_config.force_old_native_lib = true;
         } else if let Some(param) = arg.strip_prefix("-Zmiri-num-cpus=") {
             let num_cpus = param
                 .parse::<u32>()
@@ -792,6 +795,16 @@ fn main() {
 
     debug!("rustc arguments: {:?}", rustc_args);
     debug!("crate arguments: {:?}", miri_config.args);
+    #[cfg(target_os = "linux")]
+    if !miri_config.native_lib.is_empty() && !miri_config.force_old_native_lib {
+        // FIXME: This should display a diagnostic / warning on error
+        // SAFETY: If any other threads exist at this point (namely for the ctrlc
+        // handler), they will not interact with anything on the main rustc/Miri
+        // thread in an async-signal-unsafe way such as by accessing shared
+        // semaphores, etc.; the handler only calls `sleep()` and `exit()`, which
+        // are async-signal-safe, as is accessing atomics
+        let _ = unsafe { miri::init_sv() };
+    }
     run_compiler_and_exit(
         &rustc_args,
         &mut MiriCompilerCalls::new(miri_config, many_seeds, genmc_config),
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -133,6 +133,7 @@ pub enum NonHaltingDiagnostic {
         details: bool,
     },
     NativeCallSharedMem,
+    NativeCallNoTrace,
     WeakMemoryOutdatedLoad {
         ptr: Pointer,
     },
@@ -629,6 +630,8 @@ impl<'tcx> MiriMachine<'tcx> {
             Int2Ptr { .. } => ("integer-to-pointer cast".to_string(), DiagLevel::Warning),
             NativeCallSharedMem =>
                 ("sharing memory with a native function".to_string(), DiagLevel::Warning),
+            NativeCallNoTrace =>
+                ("unable to trace native code memory accesses".to_string(), DiagLevel::Warning),
             ExternTypeReborrow =>
                 ("reborrow of reference to `extern type`".to_string(), DiagLevel::Warning),
             CreatedPointerTag(..)
@@ -664,6 +667,10 @@ impl<'tcx> MiriMachine<'tcx> {
                 format!("progress report: current operation being executed is here"),
             Int2Ptr { .. } => format!("integer-to-pointer cast"),
             NativeCallSharedMem => format!("sharing memory with a native function called via FFI"),
+            NativeCallNoTrace =>
+                format!(
+                    "sharing memory with a native function called via FFI, and unable to use ptrace"
+                ),
             WeakMemoryOutdatedLoad { ptr } =>
                 format!("weak memory emulation: outdated value returned from load at {ptr}"),
             ExternTypeReborrow =>
@@ -712,6 +719,22 @@ impl<'tcx> MiriMachine<'tcx> {
             NativeCallSharedMem => {
                 vec![
                     note!(
+                        "when memory is shared with a native function call, Miri can only track initialisation and provenance on a best-effort basis"
+                    ),
+                    note!(
+                        "in particular, Miri assumes that the native call initializes all memory it has written to"
+                    ),
+                    note!(
+                        "Miri also assumes that any part of this memory may be a pointer that is permitted to point to arbitrary exposed memory"
+                    ),
+                    note!(
+                        "what this means is that Miri will easily miss Undefined Behavior related to incorrect usage of this shared memory, so you should not take a clean Miri run as a signal that your FFI code is UB-free"
+                    ),
+                ]
+            }
+            NativeCallNoTrace => {
+                vec![
+                    note!(
                         "when memory is shared with a native function call, Miri stops tracking initialization and provenance for that memory"
                     ),
                     note!(
@@ -722,6 +745,10 @@ impl<'tcx> MiriMachine<'tcx> {
                     ),
                     note!(
                         "what this means is that Miri will easily miss Undefined Behavior related to incorrect usage of this shared memory, so you should not take a clean Miri run as a signal that your FFI code is UB-free"
+                    ),
+                    #[cfg(target_os = "linux")]
+                    note!(
+                        "this is normally partially mitigated, but either -Zmiri-force-old-native-lib-mode was passed or ptrace is disabled on your system"
                     ),
                 ]
             }
--- a/src/shims/native_lib.rs
+++ b/src/shims/native_lib.rs
@@ -1,5 +1,7 @@
 //! Implements calling functions from a native library.
 use std::ops::Deref;
+#[cfg(target_os = "linux")]
+use std::{cell::RefCell, rc::Rc};
 
 use libffi::high::call as ffi;
 use libffi::low::CodePtr;
@@ -8,7 +10,15 @@ use rustc_middle::mir::interpret::Pointer;
 use rustc_middle::ty::{self as ty, IntTy, UintTy};
 use rustc_span::Symbol;
 
+#[cfg(target_os = "linux")]
+use crate::alloc::isolated_alloc::IsolatedAlloc;
 use crate::*;
+
+#[cfg(target_os = "linux")]
+type CallResult<'tcx> =
+    InterpResult<'tcx, (ImmTy<'tcx>, Option<shims::trace::messages::MemEvents>)>;
+#[cfg(not(target_os = "linux"))]
+type CallResult<'tcx> = InterpResult<'tcx, (ImmTy<'tcx>, Option<!>)>;
 
 impl<'tcx> EvalContextExtPriv<'tcx> for crate::MiriInterpCx<'tcx> {}
 trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
@@ -19,8 +29,13 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
         dest: &MPlaceTy<'tcx>,
         ptr: CodePtr,
         libffi_args: Vec<libffi::high::Arg<'a>>,
-    ) -> InterpResult<'tcx, ImmTy<'tcx>> {
+    ) -> CallResult<'tcx> {
         let this = self.eval_context_mut();
+        #[cfg(target_os = "linux")]
+        let alloc = this.machine.allocator.clone();
+        #[cfg(not(target_os = "linux"))]
+        let alloc = ();
+        let maybe_memevents;
 
         // Call the function (`ptr`) with arguments `libffi_args`, and obtain the return value
         // as the specified primitive integer type
@@ -30,60 +45,71 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // Unsafe because of the call to native code.
                 // Because this is calling a C function it is not necessarily sound,
                 // but there is no way around this and we've checked as much as we can.
-                let x = unsafe { ffi::call::<i8>(ptr, libffi_args.as_slice()) };
-                Scalar::from_i8(x)
+                let x = unsafe { do_native_call::<i8>(ptr, libffi_args.as_slice(), alloc) };
+                maybe_memevents = x.1;
+                Scalar::from_i8(x.0)
             }
             ty::Int(IntTy::I16) => {
-                let x = unsafe { ffi::call::<i16>(ptr, libffi_args.as_slice()) };
-                Scalar::from_i16(x)
+                let x = unsafe { do_native_call::<i16>(ptr, libffi_args.as_slice(), alloc) };
+                maybe_memevents = x.1;
+                Scalar::from_i16(x.0)
             }
             ty::Int(IntTy::I32) => {
-                let x = unsafe { ffi::call::<i32>(ptr, libffi_args.as_slice()) };
-                Scalar::from_i32(x)
+                let x = unsafe { do_native_call::<i32>(ptr, libffi_args.as_slice(), alloc) };
+                maybe_memevents = x.1;
+                Scalar::from_i32(x.0)
             }
             ty::Int(IntTy::I64) => {
-                let x = unsafe { ffi::call::<i64>(ptr, libffi_args.as_slice()) };
-                Scalar::from_i64(x)
+                let x = unsafe { do_native_call::<i64>(ptr, libffi_args.as_slice(), alloc) };
+                maybe_memevents = x.1;
+                Scalar::from_i64(x.0)
             }
             ty::Int(IntTy::Isize) => {
-                let x = unsafe { ffi::call::<isize>(ptr, libffi_args.as_slice()) };
-                Scalar::from_target_isize(x.try_into().unwrap(), this)
+                let x = unsafe { do_native_call::<isize>(ptr, libffi_args.as_slice(), alloc) };
+                maybe_memevents = x.1;
+                Scalar::from_target_isize(x.0.try_into().unwrap(), this)
             }
             // uints
             ty::Uint(UintTy::U8) => {
-                let x = unsafe { ffi::call::<u8>(ptr, libffi_args.as_slice()) };
-                Scalar::from_u8(x)
+                let x = unsafe { do_native_call::<u8>(ptr, libffi_args.as_slice(), alloc) };
+                maybe_memevents = x.1;
+                Scalar::from_u8(x.0)
             }
             ty::Uint(UintTy::U16) => {
-                let x = unsafe { ffi::call::<u16>(ptr, libffi_args.as_slice()) };
-                Scalar::from_u16(x)
+                let x = unsafe { do_native_call::<u16>(ptr, libffi_args.as_slice(), alloc) };
+                maybe_memevents = x.1;
+                Scalar::from_u16(x.0)
             }
             ty::Uint(UintTy::U32) => {
-                let x = unsafe { ffi::call::<u32>(ptr, libffi_args.as_slice()) };
-                Scalar::from_u32(x)
+                let x = unsafe { do_native_call::<u32>(ptr, libffi_args.as_slice(), alloc) };
+                maybe_memevents = x.1;
+                Scalar::from_u32(x.0)
             }
             ty::Uint(UintTy::U64) => {
-                let x = unsafe { ffi::call::<u64>(ptr, libffi_args.as_slice()) };
-                Scalar::from_u64(x)
+                let x = unsafe { do_native_call::<u64>(ptr, libffi_args.as_slice(), alloc) };
+                maybe_memevents = x.1;
+                Scalar::from_u64(x.0)
             }
             ty::Uint(UintTy::Usize) => {
-                let x = unsafe { ffi::call::<usize>(ptr, libffi_args.as_slice()) };
-                Scalar::from_target_usize(x.try_into().unwrap(), this)
+                let x = unsafe { do_native_call::<usize>(ptr, libffi_args.as_slice(), alloc) };
+                maybe_memevents = x.1;
+                Scalar::from_target_usize(x.0.try_into().unwrap(), this)
             }
             // Functions with no declared return type (i.e., the default return)
             // have the output_type `Tuple([])`.
             ty::Tuple(t_list) if (*t_list).deref().is_empty() => {
-                unsafe { ffi::call::<()>(ptr, libffi_args.as_slice()) };
-                return interp_ok(ImmTy::uninit(dest.layout));
+                let (_, mm) = unsafe { do_native_call::<()>(ptr, libffi_args.as_slice(), alloc) };
+                return interp_ok((ImmTy::uninit(dest.layout), mm));
             }
             ty::RawPtr(..) => {
-                let x = unsafe { ffi::call::<*const ()>(ptr, libffi_args.as_slice()) };
-                let ptr = Pointer::new(Provenance::Wildcard, Size::from_bytes(x.addr()));
+                let x = unsafe { do_native_call::<*const ()>(ptr, libffi_args.as_slice(), alloc) };
+                maybe_memevents = x.1;
+                let ptr = Pointer::new(Provenance::Wildcard, Size::from_bytes(x.0.addr()));
                 Scalar::from_pointer(ptr, this)
             }
             _ => throw_unsup_format!("unsupported return type for native call: {:?}", link_name),
         };
-        interp_ok(ImmTy::from_scalar(scalar, dest.layout))
+        interp_ok((ImmTy::from_scalar(scalar, dest.layout), maybe_memevents))
     }
 
     /// Get the pointer to the function of the specified name in the shared object file,
@@ -179,6 +205,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // The first time this happens, print a warning.
                 if !this.machine.native_call_mem_warned.replace(true) {
                     // Newly set, so first time we get here.
+                    #[cfg(target_os = "linux")]
+                    if shims::trace::Supervisor::poll() {
+                        this.emit_diagnostic(NonHaltingDiagnostic::NativeCallSharedMem);
+                    } else {
+                        this.emit_diagnostic(NonHaltingDiagnostic::NativeCallNoTrace);
+                    }
+                    #[cfg(not(target_os = "linux"))]
                     this.emit_diagnostic(NonHaltingDiagnostic::NativeCallSharedMem);
                 }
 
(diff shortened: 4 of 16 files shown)


src/alloc/isolated_alloc.rs
```diff
@@ -266,6 +266,18 @@ impl IsolatedAlloc {
             alloc::dealloc(ptr, layout);
         }
     }
+
+    /// Returns a vector of page addresses managed by the allocator.
+    pub fn pages(&self) -> Vec<usize> {
+        let mut pages: Vec<_> =
+            self.page_ptrs.clone().into_iter().map(|p| p.expose_provenance()).collect();
+        self.huge_ptrs.iter().for_each(|(ptr, size)| {
```
oli-obk:
```suggestion
        for (ptr, size) in &self.huge_ptrs {
```

src/alloc/isolated_alloc.rs
```diff
@@ -266,6 +266,18 @@ impl IsolatedAlloc {
...
+            for i in 0..size / self.page_size {
+                pages.push(unsafe { ptr.add(i * self.page_size).expose_provenance() });
```
oli-obk: Since we're moving to usize anyway, can't you expose the `ptr` first and then just do integer ops on it, completely in safe code?

src/bin/miri.rs
```diff
@@ -792,6 +795,12 @@ fn main() {
 
     debug!("rustc arguments: {:?}", rustc_args);
     debug!("crate arguments: {:?}", miri_config.args);
+    #[cfg(target_os = "linux")]
+    if !miri_config.native_lib.is_empty() && !miri_config.force_old_native_lib {
+        // FIXME: This should display a diagnostic / warning on error
+        // SAFETY: No other threads have spawned yet
```
oli-obk: This seems like a footgun. If we register the ctlrc handler before this then suddenly it's UB? Not sure how to improve this, maybe there's a way to ask the process how many threads it has or sth?

nia-e: Hm, to be fair we can play a little fast and loose here. [The docs for `fork()`](https://www.man7.org/linux/man-pages/man2/fork.2.html) claim that after forking in a multithreaded context the child process must only do async-signal-safe things, but that's in a "it might be safe to do other things but it depends on what the other threads are doing" way, not in an "it's always UB to call async-signal-unsafe code" way. As long as we know what the ctrlc handler / other threads are doing, this is still sound to do in a multithreaded context. I'll check the code but given we're talking about an interrupt handler, I find it extremely unlikely that it's an issue

oli-obk: ok, leaving some more info on this safety comment about what you just told me works for me

src/shims/trace/child.rs
```diff
@@ -0,0 +1,231 @@
...
+                // system resources
+                let init = std::panic::catch_unwind(|| {
+                    let listener =
+                        ChildListener { message_rx, attached: false, override_retcode: None };
+                    // Trace as many things as possible, to be able to handle them as needed
+                    let options = ptrace::Options::PTRACE_O_TRACESYSGOOD
+                        | ptrace::Options::PTRACE_O_TRACECLONE
+                        | ptrace::Options::PTRACE_O_TRACEFORK;
+                    // Attach to the child process without stopping it
+                    match ptrace::seize(child, options) {
```
oli-obk, line 187: Except for this part, most of the logic here is duplicating some parts of `mitosis`. But I see no way to get the right handle out of the child in https://docs.rs/mitosis/0.1.1/mitosis/struct.JoinHandle.html

While I don't want to put the burden on you to upstream whatever is needed to support this use case, please leave a FIXME so we'll investigate handrolling less of the logic around forking.

src/shims/trace/mod.rs
oli-obk, line 1: @RalfJung should we move this module and IsolatedAlloc into a separate crate within this repo that we depend on via a path dependency? Should speed up edit cycles by a small amount and generally improve separation (and thus allow unit testing nicely and potentially moving out of tree at some point). It worked for ui_test 😆

src/shims/trace/parent.rs
```diff
@@ -0,0 +1,265 @@
+use ipc_channel::ipc;
+use nix::sys::{ptrace, signal, wait};
+use nix::unistd;
+
+use crate::shims::trace::StartFfiInfo;
+use crate::shims::trace::messages::{Confirmation, MemEvents, TraceRequest};
```
oli-obk:
```suggestion
use super::StartFfiInfo;
use super::messages::{Confirmation, MemEvents, TraceRequest};
```

src/shims/trace/child.rs
```diff
@@ -0,0 +1,231 @@
...
+use std::rc::Rc;
+
+use ipc_channel::ipc;
+use nix::sys::{ptrace, signal};
+use nix::unistd;
+
+use crate::alloc::isolated_alloc::IsolatedAlloc;
+use crate::shims::trace::messages::{Confirmation, MemEvents, TraceRequest};
+use crate::shims::trace::parent::{ChildListener, sv_loop};
+use crate::shims::trace::{FAKE_STACK_SIZE, StartFfiInfo};
```
oli-obk:
```suggestion
use super::messages::{Confirmation, MemEvents, TraceRequest};
use super::parent::{ChildListener, sv_loop};
use super::{FAKE_STACK_SIZE, StartFfiInfo};
```

oli-obk (commented):
yea this lgtm. Really cool that it's possible

src/shims/trace/child.rs
nia-e: Wasn't familiar with that crate, but I'll take a look. If I have time later to upstream some of this and then just use that, that could be nice

nia-e:
Think this all addresses the review, thanks!

src/shims/trace/mod.rs
oli-obk: ping @RalfJung

thoughts?

oli-obk (approved):
please rebase and squash

RalfJung:
I don't think it's worth a separate crate tbh. But I have also not looked into this PR at all yet.^^

nia-e:
In transit currently so I can't build Miri locally to test this but if CI passes this should be fine, ty!

Out of sheer curiosity regarding the code which caused the merge conflict, is there any semantic difference between `(*t_list).deref()` and just `(**t_list)`? Can't say I've seen the former pattern before

oli-obk:
Oh that was fallout from the metasized work. This may just be stuff from early drafts we forgot to fix later or it's something about inference. Unsure. But it should almost always be equivalent

src/alloc/isolated_alloc.rs
```diff
@@ -266,6 +266,18 @@ impl IsolatedAlloc {
             alloc::dealloc(ptr, layout);
         }
     }
+
+    /// Returns a vector of page addresses managed by the allocator.
+    pub fn pages(&self) -> Vec<usize> {
+        let mut pages: Vec<_> =
+            self.page_ptrs.clone().into_iter().map(|p| p.expose_provenance()).collect();
```
RalfJung, line 273: This allocates the buffer for these pages *twice*... that seems rather unnecessary.

src/bin/miri.rs
```diff
@@ -227,10 +227,11 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
         } else {
             let return_code = miri::eval_entry(tcx, entry_def_id, entry_type, &config, None)
                 .unwrap_or_else(|| {
+                    #[cfg(target_os = "linux")]
+                    miri::register_retcode_sv(rustc_driver::EXIT_FAILURE);
```
RalfJung, line 231: Why is this done here but not all the other places where we `exit`?

src/bin/miri.rs
```diff
@@ -792,6 +795,16 @@ fn main() {
...
     debug!("crate arguments: {:?}", miri_config.args);
+    #[cfg(target_os = "linux")]
+    if !miri_config.native_lib.is_empty() && !miri_config.force_old_native_lib {
+        // FIXME: This should display a diagnostic / warning on error
+        // SAFETY: If any other threads exist at this point (namely for the ctrlc
+        // handler), they will not interact with anything on the main rustc/Miri
+        // thread in an async-signal-unsafe way such as by accessing shared
+        // semaphores, etc.; the handler only calls `sleep()` and `exit()`, which
+        // are async-signal-safe, as is accessing atomics
+        let _ = unsafe { miri::init_sv() };
```
RalfJung, line 806: Why do we just *ignore* errors here...?

src/shims/trace/child.rs
```diff
@@ -0,0 +1,238 @@
...
+    }
+}
+
+/// Initialises the supervisor process. If this function errors, then the
+/// supervisor process could not be created successfully; else, the caller
+/// is now the child process and can communicate via `start_ffi`/`end_ffi`,
+/// receiving back events through `get_events`.
+///
+/// # Safety
+/// The invariants for `fork()` must be upheld by the caller.
```
RalfJung, line 138: I don't understand this safety comment. It should at least provide an easily clickable references to those `fork` invariants.

src/bin/miri.rs
```diff
@@ -792,6 +795,16 @@ fn main() {
...
     debug!("rustc arguments: {:?}", rustc_args);
     debug!("crate arguments: {:?}", miri_config.args);
+    #[cfg(target_os = "linux")]
+    if !miri_config.native_lib.is_empty() && !miri_config.force_old_native_lib {
+        // FIXME: This should display a diagnostic / warning on error
+        // SAFETY: If any other threads exist at this point (namely for the ctrlc
+        // handler), they will not interact with anything on the main rustc/Miri
+        // thread in an async-signal-unsafe way such as by accessing shared
+        // semaphores, etc.; the handler only calls `sleep()` and `exit()`, which
+        // are async-signal-safe, as is accessing atomics
```
RalfJung, line 805: Is sounds like this relying on the ctrlc internal implementation details that could change any time...?

I also have no clue why we are talking about threads here, but that may be because `init_sv` doesn't have a self-contained safety comment.

src/diagnostics.rs
```diff
@@ -664,6 +667,10 @@ impl<'tcx> MiriMachine<'tcx> {
                 format!("progress report: current operation being executed is here"),
             Int2Ptr { .. } => format!("integer-to-pointer cast"),
             NativeCallSharedMem => format!("sharing memory with a native function called via FFI"),
+            NativeCallNoTrace =>
+                format!(
+                    "sharing memory with a native function called via FFI, and unable to use ptrace"
```
RalfJung, line 672: So all macOS and Windows users will see this? I don't think that's what we want. Tracing mode is anyway not perfect so the warning should largely be the same either way.

src/shims/trace/child.rs
```diff
@@ -0,0 +1,238 @@
...
+    event_rx: ipc::IpcReceiver<MemEvents>,
+}
+
+/// Marker representing that an error occurred during creation of the supervisor.
+#[derive(Debug)]
+pub struct SvInitError;
+
+impl Supervisor {
+    /// Returns `true` if the supervisor process exists, and `false` otherwise.
+    pub fn poll() -> bool {
```
RalfJung, line 33: Why is this called `poll`? It doesn't poll anything, it just checks whether the supervisor is enabled...?

src/shims/trace/child.rs
```diff
@@ -0,0 +1,238 @@
...
+use ipc_channel::ipc;
+use nix::sys::{ptrace, signal};
+use nix::unistd;
+
+use super::messages::{Confirmation, MemEvents, TraceRequest};
+use super::parent::{ChildListener, sv_loop};
+use super::{FAKE_STACK_SIZE, StartFfiInfo};
+use crate::alloc::isolated_alloc::IsolatedAlloc;
+
+static SUPERVISOR: std::sync::Mutex<Option<Supervisor>> = std::sync::Mutex::new(None);
```
RalfJung, line 13: Uh... we generally don't use globals in Miri, everything should be in the machine. I think we need more design discussion here, and at the very least an extensive comment explaining why this is not in the machine.

src/diagnostics.rs
```diff
@@ -710,6 +717,22 @@ impl<'tcx> MiriMachine<'tcx> {
                 v
             }
             NativeCallSharedMem => {
+                vec![
+                    note!(
+                        "when memory is shared with a native function call, Miri can only track initialisation and provenance on a best-effort basis"
+                    ),
+                    note!(
+                        "in particular, Miri assumes that the native call initializes all memory it has written to"
```
RalfJung, line 725: This isn't even correct, is it? None of this extra tracing is actually wired up yet...

src/shims/native_lib.rs
```diff
@@ -196,12 +229,55 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    args: &[ffi::Arg<'_>],
+    alloc: Option<Rc<RefCell<IsolatedAlloc>>>,
+) -> (T, Option<shims::trace::messages::MemEvents>) {
+    use shims::trace::Supervisor;
+
+    unsafe {
+        if let Some(alloc) = alloc {
+            // SAFETY: We don't touch the machine memory past this point
+            let (guard, stack_ptr) = Supervisor::start_ffi(alloc.clone());
+            // SAFETY: Upheld by caller
```
RalfJung, line 255: Wouldn't it be easier to do this in `call_native_with_args` before and after the big match? That then also requires less passing-around of things...

src/shims/trace/child.rs
```diff
@@ -0,0 +1,238 @@
...
+    /// the last possible moment before entering said call. `alloc` points to
+    /// the allocator which handed out the memory used for this machine.
+    ///
+    /// As this locks the supervisor via a mutex, no other threads may enter FFI
+    /// until this one returns and its guard is dropped via `end_ffi`. The
+    /// pointer returned should be passed to `end_ffi` to avoid a memory leak.
+    ///
+    /// SAFETY: The resulting guard must be dropped *via `end_ffi`* immediately
+    /// after the desired call has concluded.
+    pub unsafe fn start_ffi(
```
RalfJung, line 47: Seems better to return a guard type with a destructor? That's the usual Rust way of having an API with some cleanup happening later.

src/shims/native_lib.rs
```diff
@@ -19,8 +29,13 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
         dest: &MPlaceTy<'tcx>,
         ptr: CodePtr,
         libffi_args: Vec<libffi::high::Arg<'a>>,
-    ) -> InterpResult<'tcx, ImmTy<'tcx>> {
+    ) -> CallResult<'tcx> {
```
RalfJung, line 32: I am very confused by this API. Why do we get events *at the end*? Don't we need to deal with events in real time, while the native code is running?

src/shims/native_lib.rs
```diff
@@ -196,12 +229,55 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    ptr: CodePtr,
+    args: &[ffi::Arg<'_>],
+    alloc: Option<Rc<RefCell<IsolatedAlloc>>>,
+) -> (T, Option<shims::trace::messages::MemEvents>) {
+    use shims::trace::Supervisor;
+
+    unsafe {
+        if let Some(alloc) = alloc {
+            // SAFETY: We don't touch the machine memory past this point
+            let (guard, stack_ptr) = Supervisor::start_ffi(alloc.clone());
```
RalfJung, line 254: Seems like `(guard, stack_ptr)` is not something the user should actually see in pieces, it's just a package that should be passed to `end_call`? Please make a proper abstraction here by making this a new type.

src/shims/trace/child.rs
```diff
@@ -0,0 +1,238 @@
...
+    ) -> (std::sync::MutexGuard<'static, Option<Supervisor>>, Option<*mut [u8; FAKE_STACK_SIZE]>)
+    {
+        let mut sv_guard = SUPERVISOR.lock().unwrap();
+        // If the supervisor is not initialised for whatever reason, fast-fail.
+        // This might be desired behaviour, as even on platforms where ptracing
+        // is not implemented it enables us to enforce that only one FFI call
+        // happens at a time
+        let Some(sv) = sv_guard.take() else {
+            return (sv_guard, None);
+        };
```
RalfJung, line 58: The comment says "fast-fail", but this doesn't seem to return a failure, the remaining logic keeps doing the native call anyway. So how does that make sense?

src/shims/trace/child.rs
```diff
@@ -0,0 +1,238 @@
...
+            .try_recv_timeout(std::time::Duration::from_secs(5))
+            .map_err(|e| {
+                match e {
+                    ipc::TryRecvError::IpcError(_) => (),
+                    ipc::TryRecvError::Empty =>
+                        eprintln!("Waiting for accesses from supervisor timed out!"),
+                }
+            })
+            .ok();
+        // Do *not* leave the supervisor empty, or else we might get another fork...
```
RalfJung, line 126: What does this mean?

src/shims/trace/child.rs
```diff
@@ -0,0 +1,238 @@
...
+        }
+        // On the off-chance something really weird happens, don't block forever
+        let ret = sv
+            .event_rx
+            .try_recv_timeout(std::time::Duration::from_secs(5))
+            .map_err(|e| {
+                match e {
+                    ipc::TryRecvError::IpcError(_) => (),
+                    ipc::TryRecvError::Empty =>
+                        eprintln!("Waiting for accesses from supervisor timed out!"),
```
RalfJung, line 122: Shouldn't we abort here? Continuing seems like it could lead to arbitrary garbage behavior?

src/shims/trace/child.rs
```diff
@@ -0,0 +1,238 @@
...
+/// Instruct the supervisor process to return a particular code. Useful if for
+/// whatever reason this code fails to be intercepted normally. In the case of
+/// `abort_if_errors()` used in `bin/miri.rs`, the return code is erroneously
+/// given as a 0 if this is not used.
+pub fn register_retcode_sv(code: i32) {
+    let mut sv_guard = SUPERVISOR.lock().unwrap();
+    if let Some(sv) = sv_guard.take() {
+        sv.message_tx.send(TraceRequest::OverrideRetcode(code)).unwrap();
+        *sv_guard = Some(sv);
+    }
```
RalfJung, line 237: Why does this take-and-put-back `sv`? It could just borrow it, no?

src/shims/trace/mod.rs
```diff
@@ -0,0 +1,31 @@
+mod child;
+pub mod messages;
+mod parent;
+
+use std::ops::Range;
+
+pub use self::child::{Supervisor, init_sv, register_retcode_sv};
+
+/// The size used for the array into which we can move the stack pointer.
+const FAKE_STACK_SIZE: usize = 1024;
```
RalfJung, line 10: So we have just a 1KB stack here? Is there some particular constraint on this value? What happens if the program needs more stack?

What even is the point of the "fake stack"?

src/shims/trace/parent.rs
```diff
@@ -0,0 +1,263 @@
+use ipc_channel::ipc;
+use nix::sys::{ptrace, signal, wait};
+use nix::unistd;
+
+use super::StartFfiInfo;
+use super::messages::{Confirmation, MemEvents, TraceRequest};
+
+/// The flags to use when calling `waitid()`.
+/// Since bitwise OR on the nix version of these flags is implemented as a trait,
+/// we can't use them directly so we do it this way
```
RalfJung, line 10: I don't understand this comment.

src/shims/trace/parent.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    // Memory allocated on the MiriMachine
+    let mut _ch_pages = Vec::new();
+    let mut _ch_stack = None;
+
+    // The pid of the last process we interacted with, used by default if we don't have a
+    // reason to use a different one
+    let mut curr_pid = init_pid;
+
+    // There's an initial sigstop we need to deal with
```
RalfJung, line 160: Where does that come from?

src/shims/trace/parent.rs
```diff
@@ -0,0 +1,263 @@
...
+                _ch_stack = Some(ch_info.stack_ptr);
+
+                // We received the signal and are no longer in the main listener loop,
+                // so we can let the child move on to the end of start_ffi where it will
+                // raise a SIGSTOP. We need it to be signal-stopped *and waited for* in
+                // order to do most ptrace operations!
+                confirm_tx.send(Confirmation).unwrap();
+                // We can't trust simply calling `Pid::this()` in the child process to give the right
+                // PID for us, so we get it this way
+                curr_pid = wait_for_signal(None, signal::SIGSTOP, false).unwrap();
```
RalfJung, line 184: Who's ending this SIGSTOP and where? Seems like there's a protocol here that I haven't seen documented so far.

src/shims/trace/parent.rs
```diff
@@ -0,0 +1,263 @@
...
+                eprintln!("Process unexpectedly got {signal}; continuing...");
+                // In case we're not tracing
+                if ptrace::syscall(pid, signal).is_err() {
+                    // If *this* fails too, something really weird happened
+                    // and it's probably best to just panic
+                    signal::kill(pid, signal::SIGCONT).unwrap();
+                }
+            }
+            // Child entered a syscall; we wait for exits inside of this, so it
+            // should never trigger on return from a syscall we care about
```
RalfJung, line 210: I don't understand this comments. Who's waiting for what where and why?

tests/native-lib/pass/ptr_read_access.stderr
RalfJung, line 1: I am very surprised that this passed CI since we should be running this on macOS as well. Any idea what is happening here?

src/shims/trace/child.rs
RalfJung, line 1: `shims/trace` is a bit ambiguous IMO, let's move this inside a `native_lib` module.

RalfJung (commented):
I still have a bunch of questions and comments. Some of them are resolved in https://github.com/rust-lang/miri/pull/4418, but a bunch are clarification questions, so I'd appreciate a follow-up PR extending the comments here and possibly doing some refactoring.

src/bin/miri.rs
RalfJung: AFAIK `fork` suspends/deletes all threads except for the one doing the `fork`. So why does it matter what those other threads do?

RalfJung: The problem with `fork` in a concurrent program is that those other threads may hold locks, which will now never be released. That's why async-signal-safety matters -- it's a lot like a signal handler, where the corresponding main thread may hold locks that will never get released while the signal handler runs.

src/bin/miri.rs
nia-e: It seems to have something to do with the way `abort_if_errors()` exits not properly getting us a return code. I'm unsure of what the root cause is - I spent a lot of time trying to debug that but it seems to be the only thing that causes this behaviour, so I special-cased it

src/bin/miri.rs
nia-e: If it errored and the sv process failed to init, then we catch that when calling `poll()` on it. I would like this to emit a diagnostic on error, though; it just didn't seem necessary to include that as part of this PR

src/shims/trace/child.rs
nia-e: I can add that, sure thing.

src/diagnostics.rs
nia-e: I'm fine changing that back, sure

src/shims/trace/child.rs
nia-e: Couldn't think of a better term for it I suppose. Maybe `check()`?

src/shims/trace/child.rs
nia-e: Doing this per-machine is very difficult without much deeper changes I believe; the `ipc-channel`s are MPSC so we can't have each machine store a copy of them. I don't think it's even realistically possible to have a single supervisor process effectively trace what's happening in multiple FFI calls in parallel, so we'd just be reimplementing a lock instead of doing it like this with a global mutex. The alternative is having 2 processes per `MiriMachine` which sounds even worse to me both complexity- and performance-wise.

src/diagnostics.rs
nia-e: I mentioned in the PR description that I left comments/docs "as if" everything was fully hooked up, but again I'm fine changing this

src/shims/native_lib.rs
nia-e: It would be, yes. I did it this way to keep everything as self-contained as possible for the sake of easing the maintenance burden, so that in the case of any future refactor in the native-libs code all that needs to be done is call `start_ffi()` and `end_ffi()` and not worry about anything else, but again I can change this

src/shims/trace/child.rs
nia-e: Fair enough, I'll do that

src/shims/native_lib.rs
nia-e: I couldn't really see an effective way to do that, since it might need more arch-specific code (e.g. setting up the stack in a sane way to allow us to longjmp into arbitrarily complex internal Miri logic). I don't see any major disadvantage of doing it this way, save for it not allowing us to do lazy initialisation of statics and such. I thought I explained this was the design I was going for back in my original design doc?

src/shims/trace/child.rs
nia-e: I should clarify the comment. It will do the native call, just fall back to the old logic instead, since `start_ffi()` will take this path iff `poll()` returned false

src/shims/trace/child.rs
nia-e: If the global is left as a `None` that's taken as a signal that the supervisor process has not been initialised. It would take a refactor of the current code to actually have this result in a second fork, but that is how `init_sv()` determines whether it does anything or is just a no-op, by checking if that mutex holds a `None`.

src/shims/trace/child.rs
nia-e: As I understand it, it would lead to Miri potentially thinking UB happened when it did not. If that's a panic-worthy situation, I'm fine changing the code to do so

src/shims/trace/child.rs
nia-e: That's what I initially wanted to do, but for whatever reason it complained when I tried that and the type was behind a `MutexGuard`. I can try again, maybe I missed something

src/shims/trace/mod.rs
nia-e: It just needs to be big enough to allow us to jump into `mempr_on` and `mempr_off`. If we don't do this, jumping into them might mess up the stack and cause problems once we jump back into the C code. The size was chosen semi arbitrarily, since a function that small will not take up more than a hundred or so bytes for its stackframe so I just made sure it had some padding on top of that to be safe. So, only those functions ever touch this fake stack. I considered having this be a global too but decided against it since I knew the general policy was to minimise the number of those.

src/shims/trace/parent.rs
nia-e: We can't have `const WAIT_FLAGS: wait::WaitPidFlag = wait::WaitPidFlag::WUNTRACED | wait::WaitPidFlag::WEXITED` because `nix` uses `bitflags` and this issue is blocking them being const: https://github.com/bitflags/bitflags/issues/180. I could do this with `union()` instead; this just seemed clearer I suppose

src/shims/trace/parent.rs
nia-e: `init_sv()` raises a sigstop on the child arm, to wait in place and make sure that the parent process can `ptrace` it properly. If it can't, then the child is killed instead of being resumed from sigstop

src/shims/trace/parent.rs
nia-e: This is the sigstop that the child raises in `start_ffi()`. Then calling `ptrace::syscall()` is what ends it (it's the same as `ptrace::cont()` but also pauses the child when it enters or exits a syscall; right now unused, but will be important for the malloc tracing later)

src/shims/trace/parent.rs
nia-e: Bah, this is wrong. I should have edited this when I gutted the malloc tracing bits, apologies

tests/native-lib/pass/ptr_read_access.stderr
nia-e: I `cfg(linux)`'ed all of the trace-related bits since macOS has a very barebones (bad) ptrace implementation, so when running on it it'll just use the old behaviour

src/bin/miri.rs
nia-e: I find it somewhat hard to imagine how the ctrlc handler could realistically be changed to act in ways that may break the safety invariants for signal safety, but for clarity I'll add a better safety comment for `init_sv`.

src/shims/trace/child.rs
nia-e: I have no idea how I forgot `as_mut()` exists for optionals... oops

src/bin/miri.rs
RalfJung: `abort_if_errors` doesn't exit, it unwinds. `catch_with_exit_code` then turns that into an `exit`.

I think you should be able to remove the random `register_retcode` here, and then once https://github.com/rust-lang/miri/pull/4406 lands add this inside the new `fn exit` there.

Or does ptrace halt execution on an unwind? It should just resume execution then to let the program run its normal course.

src/bin/miri.rs
RalfJung: Are you saying we dynamically fall back to calling native code without fine-grained tracing if the setup fails?

That makes sense, but having a `let _` then does not. Also, there should be a comment.

> emit a diagnostic on error, though; it just didn't seem necessary to include that as part of this PR

If you have confusing things like `let _` instead you need to explain those, or at least add `FIXME(ptrace): emit a warning here` or so.

EDIT: There actually is a FIXME, fair. I think I'd have rather seen this staged a bit differently but that's getting into very subjective territory. ;)

src/shims/trace/child.rs
RalfJung: What about the obvious choice of `is_enabled`...?

src/shims/trace/child.rs
RalfJung: There are three options:
- Ensure there's only one machine per process in ptrace mode
- Do ptrace per-machine
- Make a single shared ptrace supervisor work with mutliple machines

It seems you are going for the latter, which is fine, but I see 0 comments explaining this. You should explicitly say that this is what happens (since it is unusual), and add an explanation of why it works. The doc comment for the global variable would be a good place for that, I think.

src/diagnostics.rs
RalfJung: It's not okay to ship code to users that is enabled by default (if they use `-Zmiri-native-lib`) and then prints diagnostics that we know to be just wrong.

I have now made it so your code isn't enabled by default, mitigating this. But please keep this in mind in the future: either the code needs to be production-ready, or it needs to be off-by-default.

src/shims/native_lib.rs
RalfJung: I don't understand how this even answers my question. See what I did in my PR -- it's still start_ffi / end_ffi, but requiring much less awkward passing around of the remaining state.

src/shims/native_lib.rs
RalfJung: >  I thought I explained this was the design I was going for back in my original design doc?

If the design doc said that the events are only processed after the native code returns, then I entirely missed it. I was sure you'd be doing live processing.

RalfJung: Okay I now see it says

> This consists of making use of debug tooling to intercept each read and write happening once FFI has been initiated and then reporting these back **at the end**.

So I guess I speed-read that document too hard, sorry for that.

src/shims/trace/child.rs
RalfJung: Looks like this needs a big comment at that global to explain its invariant.

src/shims/trace/child.rs
RalfJung: I don't understand your reply. This isn't about UB or not, it's about an apparent communication breakdown between child and supervisor? That shouldn't say "there was UB" of course, it should say "there was a communication breakdown between child and supervisor".

src/shims/trace/mod.rs
RalfJung: Okay, makes sense -- but needs more comments explaining everything. :)

src/shims/trace/parent.rs
RalfJung: Oh, just say that then -- the trait can't be used in `const`. And link to that issue.

src/shims/trace/parent.rs
RalfJung: Seems like we need documentation somewhere for the overarching protocol of who raises which signal when and what happens then.

tests/native-lib/pass/ptr_read_access.stderr
RalfJung: Right, so macos doesn't get the fine-grained ptrace. But macos apparently still prints the error that should only be printed when using fine-grained ptrace. So there's clearly a bug here, and your reply doesn't explain it.

src/bin/miri.rs
RalfJung: The main point is that I don't think it even matters what the ctrlc handler does. I think you misunderstood the safety requirements of `fork`.

src/shims/trace/parent.rs
nia-e: I put that in `trace/messages.rs` but I can expand on it if it's insufficient as-is

src/shims/trace/child.rs
nia-e: Fair, I'll make it a proper error.

src/shims/trace/parent.rs
RalfJung: This code here runs before the main event loop, the comment in `messages.rs` explains the body of the loop IIUC.

nia-e: I'll put it in the docs for `init_sv` then since that's the most obvious place I think

RalfJung: It's part of the protocol, I think it makes sense to have the entire protocol in one place somewhere. So I'd add it to `messages.rs`. The text there could also make it more clear that those steps there are being repeated arbitrarily often, and which message types are used for which of these messages.

PR #4405, opened 2025-06-18
trace: implement supervisor components for tracing

nia-e:
Per #4326, this is the second of the main tracer bringup PRs. Let me know what needs fixing / changing / docs / justification / etc :D


nia-e:
Had a couple dumb bugs to iron out but I think this is good to go now.

@rustbot ready

src/shims/native_lib.rs
```diff
@@ -229,7 +229,15 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             .collect::<Vec<libffi::high::Arg<'_>>>();
 
         // Call the function and store output, depending on return type in the function signature.
-        let (ret, _) = this.call_native_with_args(link_name, dest, code_ptr, libffi_args)?;
+        let (ret, maybe_memevents) =
+            this.call_native_with_args(link_name, dest, code_ptr, libffi_args)?;
+
+        #[cfg(target_os = "linux")]
```
oli-obk: Could use `if cfg!(target_os = "linux")`, which will get optimized, but borrowck and lints don't see it as a compile-time condition, so you don't need to suppress in the else path

src/alloc/isolated_alloc.rs
```diff
@@ -271,13 +272,54 @@ impl IsolatedAlloc {
...
+    }
+
+    /// Applies `prot` to every page managed by the allocator.
+    ///
+    /// SAFETY: Accessing memory in violation of the protection flags will
+    /// trigger a segfault.
+    unsafe fn mprotect(&mut self, prot: mman::ProtFlags) -> Result<(), nix::errno::Errno> {
+        for &pg in &self.page_ptrs {
+            unsafe {
+                // We already know only non-null ptrs are pushed to self.pages
```
oli-obk: could make `page_ptrs` and `huge_ptrs` contain NonNull directly

src/shims/trace/parent.rs
```diff
@@ -257,7 +404,305 @@ fn wait_for_signal(
...
+                _ => unimplemented!(),
+            }
+        }
+
+        Ok(())
+    }
+
+    // Get information on what caused the segfault. This contains the address
+    // that triggered it.
+    let siginfo = ptrace::getsiginfo(pid).map_err(|_| ExecError::Shrug)?;
```
oli-obk: When would this fail?

src/shims/trace/parent.rs
```diff
@@ -257,7 +404,305 @@ fn wait_for_signal(
...
+        )
+        .unwrap();
+
+        // Check if we also own the next page, and if so unprotect it in case
+        // the access spans the page boundary.
+        if ch_pages.contains(&page_addr.strict_add(page_size)) {
+            ptrace::write(pid, (&raw const PAGE_COUNT).cast_mut().cast(), 2).unwrap();
+        } else {
+            ptrace::write(pid, (&raw const PAGE_COUNT).cast_mut().cast(), 1).unwrap();
+        }
```
oli-obk:
```suggestion
        let flag = if ch_pages.contains(&page_addr.strict_add(page_size)) {
            2
        } else {
            1
        };
        ptrace::write(pid, (&raw const PAGE_COUNT).cast_mut().cast(), flag).unwrap();
```

src/shims/trace/parent.rs
nia-e: It's possible that a segfault was delivered via `kill` instead of it being a real segfault, which is unusual but would result in this failing. Though it's very rare - I've only seen 1 bit of C in the wild use `SIGSEGV`s in this way in some ancient embedded code - so I'm fine removing it and just unwrapping

oli-obk: Yea let's go with unwrap until we have a real use case

src/alloc/isolated_alloc.rs
```diff
@@ -287,30 +289,67 @@ impl IsolatedAlloc {
...
+
+    /// Deprotects all owned memory by setting it to RW. Erroring here is very
+    /// likely unrecoverable, so it may panic if applying those permissions
+    /// fails.
+    pub fn unprep_ffi(&mut self) {
+        let prot = mman::ProtFlags::PROT_READ | mman::ProtFlags::PROT_WRITE;
+        unsafe {
+            self.mprotect(prot).unwrap();
+        }
+    }
```
RalfJung, line 334: Elsewhere this is called start_ffi and end_ffi, right?

Why do we need to give a list of pages to the trace mechanism if we then still do this work here in the allocator...?

src/shims/trace/parent.rs
```diff
@@ -1,16 +1,108 @@
...
+    fn ip(&self) -> usize;
+    /// Set the instruction pointer; remember to also set the stack pointer, or
+    /// else the stack might get messed up!
+    fn set_ip(&mut self, ip: usize);
+    /// Set the stack pointer, ideally to a zeroed-out area.
+    fn set_sp(&mut self, sp: usize);
+}
+
+// It's fine / desirable behaviour for values to wrap here, we care about just
+// preserving the bit pattern.
```
RalfJung, line 57: But the bit pattern may change if we cast to a smaller type...

src/shims/trace/parent.rs
```diff
@@ -257,7 +388,302 @@ fn wait_for_signal(
...
+                                arch::arm64::Arm64Vas::ARM64_VAS_16B
+                                | arch::arm64::Arm64Vas::ARM64_VAS_8H
+                                | arch::arm64::Arm64Vas::ARM64_VAS_4S
+                                | arch::arm64::Arm64Vas::ARM64_VAS_2D
+                                | arch::arm64::Arm64Vas::ARM64_VAS_1Q => 16,
+                            };
+                            let push = addr..addr.strict_add(size);
+                            // FIXME: This now has access type info in the latest
+                            // git version of capstone because this pissed me off
+                            // and I added it. Change this when it updates.
```
RalfJung, line 475: :-)

src/shims/trace/parent.rs
```diff
@@ -257,7 +388,302 @@ fn wait_for_signal(
...
+    let page_addr = addr.strict_sub(addr.strict_rem(page_size));
+
+    if ch_pages.iter().any(|pg| (*pg..pg.strict_add(page_size)).contains(&addr)) {
+        // Overall structure:
+        // - Get the address that caused the segfault
+        // - Unprotect the memory
+        // - Step 1 instruction
+        // - Parse executed code to estimate size & type of access
+        // - Reprotect the memory
+        // - Continue
```
RalfJung, line 539: And somewhere along the line we apparently use a fake stack for something?

src/shims/trace/parent.rs
```diff
@@ -257,7 +388,302 @@ fn wait_for_signal(
...
+
+        // Check if we also own the next page, and if so unprotect it in case
+        // the access spans the page boundary.
+        let flag = if ch_pages.contains(&page_addr.strict_add(page_size)) { 2 } else { 1 };
+        ptrace::write(pid, (&raw const PAGE_COUNT).cast_mut().cast(), flag).unwrap();
+
+        ptrace::setregs(pid, new_regs).unwrap();
+
+        // Our mempr_* functions end with a raise(SIGSTOP).
+        wait_for_signal(Some(pid), signal::SIGSTOP, true)?;
```
RalfJung, line 575: That `true` argument is not very self-explaining, please use a dedicated enum instead.

src/shims/trace/parent.rs
```diff
@@ -257,7 +388,302 @@ fn wait_for_signal(
...
+        for a in (ch_stack..ch_stack.strict_add(FAKE_STACK_SIZE)).step_by(ARCH_WORD_SIZE) {
+            ptrace::write(pid, std::ptr::with_exposed_provenance_mut(a), 0).unwrap();
         }
+
+        // Save registers and grab the bytes that were executed. This would
+        // be really nasty if it was a jump or similar but those thankfully
+        // won't do memory accesses and so can't trigger this!
+        let regs_bak = ptrace::getregs(pid).unwrap();
+        new_regs = regs_bak;
+        let ip_poststep = regs_bak.ip();
```
RalfJung, line 595: Should we assert that ip_poststep is (a) bigger than ip_prestep and (b) not too much bigger, just to be safe?

src/shims/trace/parent.rs
```diff
@@ -257,7 +388,302 @@ fn wait_for_signal(
...
+                    .unwrap()
+                    .to_ne_bytes()
+                    .to_vec(),
+            );
+            ret
+        });
+
+        // Now figure out the size + type of access and log it down
+        // This will mark down e.g. the same area being read multiple times,
+        // since it's more efficient to compress the accesses at the end.
```
RalfJung, line 611: Where are we "compressing" anything...?

src/shims/trace/parent.rs
```diff
@@ -257,7 +388,302 @@ fn wait_for_signal(
...
+        if capstone_disassemble(&instr, addr, cs, acc_events).is_err() {
+            // Read goes first because we need to be pessimistic.
```
RalfJung, line 613: Why is read-first "pessimistic"?

RalfJung (commented):
I still have a bunch of questions and comments. Some of them are resolved in https://github.com/rust-lang/miri/pull/4418, but a bunch are clarification questions, so I'd appreciate a follow-up PR extending the comments here and possibly doing some refactoring.

src/shims/trace/parent.rs
```diff
@@ -144,82 +234,123 @@ pub fn sv_loop(
...
         match evt {
-            // start_ffi was called by the child, so prep memory
+            // start_ffi was called by the child, so prep memory.
             ExecEvent::Start(ch_info) => {
-                // All the pages that the child process is "allowed to" access
-                _ch_pages = ch_info.page_ptrs;
-                // And the fake stack it allocated for us to use later
-                _ch_stack = Some(ch_info.stack_ptr);
+                // All the pages that the child process is "allowed to" access.
+                ch_pages = ch_info.page_ptrs;
```
RalfJung, line 265: Why does this even work? Wasn't this allocated in the child after the `fork`, so the allocation holding this `Vec`'s buffer only exists in the child now, not in the parent?

RalfJung: Oh I get it, this is properly (de)serialized over the IPC channel.

src/shims/trace/parent.rs
```diff
@@ -257,7 +388,302 @@ fn wait_for_signal(
...
+                            }
+                        }
+                        _ => (),
+                    },
+                #[cfg(any(target_arch = "riscv32", target_arch = "riscv64"))]
+                arch::ArchOperand::RiscVOperand(risc_voperand) => {
+                    match risc_voperand {
+                        arch::riscv::RiscVOperand::Mem(_) => {
+                            // We get basically no info here.
+                            let push = addr..addr.strict_add(size);
```
RalfJung, line 509: The variable `size` does not exist. Apparently this code did not even get build-tested. Please be more careful next time, this now causes a bunch of extra work to clean it all up.

RalfJung: So, in other words, before we can re-land any riscv-specific code, we need CI set up for it. I have no idea how to set up CI that uses riscv as host architecture, probably that requires some sort of emulation.

We actually have the same problem with the arm code: it is not covered by CI at all, since we don't have a linux arm runner.

RalfJung: Turns out the 32bit ARM code also does not build.

So yeah we definitely need a CI setup for every single target we have specific code for, everything else is not sustainable.

src/alloc/isolated_alloc.rs
nia-e: I can change the naming if that's preferrable, sure.

The point is that here in the allocator we're doing the initial and final bulk protection/deprotection to make it so accesses to the machine's memory result in a segfault; the list of pages passed to the supervisor is used to tell when a segfault is "real" (e.g. the C code wrote through a null ptr) or if not (that is, if its address matches any of these pages) we log it down instead. I guess we could delete this and make the supervisor process do the protection to the child process through those `mempr_on()`/`mempr_off()` functions, but that would be quite messy

src/shims/trace/parent.rs
nia-e: I should clarify. These fields on `libc::user_regs_struct` are always `usize`/`isize`-sized, i.e. on x64 the type of `self.rip` is `i64` while on x86 the equivalent `self.eip` is `i32` (since these are defined as wobbly C types). What I meant by wrapping is just casting from signed to unsigned; this should never cause the size to decrease

src/shims/trace/parent.rs
nia-e: I can document that here as well, sure

src/shims/trace/parent.rs
nia-e: Makes sense, I'll add that

src/shims/trace/parent.rs
nia-e: Ah, this is outdated, oops. I had a bold idea at one point of compressing down the list of which reads and writes happened in some fancy way until I realised it requires we always have perfect info on the size of accesses that never conservatively overestimates said size and I forgot to update this comment when I gutted that

src/shims/trace/parent.rs
nia-e: Also an artifact of the idea I mentioned above, I'll remove this comment. But in brief I wanted to be able to assert e.g. if a write happens to some area of memory whose bytes had a provenance associated with them, then we can delete said provenance and a later read to that area won't expose it, thus the read going first being "pessimistic". I'd like to keep the ordering though, just in case at some later point capstone improves enough / I find some way to implement that (maybe having some way to signal when we do know the exact size of accesses vs guessing?)

src/shims/trace/parent.rs
nia-e: Ough. I had tested this on 64-bit arm, sorry for the inconvenience!

src/alloc/isolated_alloc.rs
RalfJung: > I guess we could delete this and make the supervisor process do the protection to the child process through those mempr_on()/mempr_off() functions, but that would be quite messy

What I was thinking was more along the lines of having all `mprotect` calls reasonably close to each other -- right now, some are in native_lib and some in the allocator. That does *not* mean they have to be done by the supervisor. They can be done by the child in start_ffi / end_ffi, but with the logic in the tracing code, not here.

If you also change the pages getter to return an `impl Iterator` instead of a `Vec`, you can avoid some unnecessary allocation in end_ffi.

src/shims/trace/parent.rs
RalfJung: Please use `cast_signed`/`cast_unsinged` for casting between signed and unsigned, do not use `as`.

src/shims/trace/parent.rs
RalfJung: I don't see the point of ordering, it's just confusing if it's not actually needed and also it's unnecessary work.

src/shims/trace/parent.rs
RalfJung: > I'll look to see if I can do anything about setting up CI, not really my area of expertise but I'll give it a shot

I spent most of Sunday setting that up. :)

src/shims/trace/parent.rs
nia-e: Should I split it up and just have it return separate vectors for the reads and writes then?

RalfJung: I think ideally it's one vector that has all events in-order. That is the simplest thing for the post-processing logic, and it can even help avoid unnecessary pointer exposure if C overwrites memory before reading it (meaning the old provenance stored there is definitely lost).

nia-e: That was my idea too. The problem was that when we can't tell the size of the access (e.g. on aarch64) we default to the worst case size, so it might be possible that - say - a 4-byte write happens, we log it down as 8 bytes bc we don't know the real size, and in doing so accidentally delete those next 4 bytes' provenance info. I am considering adding an extra field to specify when we know for sure the exact size vs when we're conservatively overestimating, which would allow us to do this

RalfJung: Ah... good point.
The general interface here is to include both a lower bound and an upper bound for the size of the access. Then the post-processing code can do the best possible approximation.

PR #4406, opened 2025-06-18
Add tracing_chrome under "tracing" feature

Stypox:
- Added a "tracing" feature that enables Chrome traces, and sets `Machine::TRACING_ENABLED` to `true`
- I ended up adding the `tracing_chrome` crate by copy-pasting [this ~600 line file](https://github.com/thoren-d/tracing-chrome/blob/7e2625ab4aeeef2f0ef9bde9d6258dd181c04472/src/lib.rs).
	- As discussed previously, depending on the *tracing-chrome* crate from crates.io is unfortunately not possible since it depends on `tracing_core` which conflicts with *rustc_private*'s `tracing_core` (meaning it would not be possible to use the `ChromeLayer` in a context that expects a `Layer` from from *rustc_private*'s `tracing_core` version).
	- I tried to use cargo's `[patch]` and `[replace]` sections, but although they would work for normal libraries, they don't seem to behave as expected when the crate to replace comes from `rustc_private`, see [this Zulip comment](https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/Adding.20a.20dependency.20to.20Miri.20that.20depends.20on.20a.20rustc.20dep/near/517896708) for a list of experiments
	- Also see more dicussion in [this Zulip thread](https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/Adding.20a.20dependency.20to.20Miri.20that.20depends.20on.20a.20rustc.20dep)
	- I am open to trying other stuff to avoid copy-pasting foreign code into Miri, but I didn't want to waste more time on this right now, as it's blocking any work on tracing.
	- Do I need to mention the author and license at the top of the copied file, or is a link (like I've done now) enough for attribution? 
- I moved the logger/tracing setup functions out of `miri.rs` so the file is less cluttered.
	- Should I put those new files under `trace/` (like done now) or should I rather move them under `bin/`, since only `bin/miri.rs` uses those functions?
- The `ChromeLayer` internally starts a thread to write data to file, and thus relies on a guard to properly flush the file and terminate the thread when dropped. Since `std::process::exit()` was being called in a few places, I had to restructure the code a bit to avoid exiting directly (which wouldn't call `drop()` on the guard).
    - As far as I understand rustc does not call exit(), but rather raises an unwinding panic which is then caught by `rustc_driver::catch_fatal_errors()`. After modifying `run_compiler_then_exit` to not exit directly, I could confirm that the tracing file was being flushed correctly in case of a compiler error by running `RUSTC_LOG=1 ./miri run --features tracing FILE_WITH_SYNTAX_ERRORS`.
    - I moved the call to `init_early_loggers()` after argument parsing, is this ok? I did not see any log/trace call during argument parsing anyway. This avoids having to refactor `show_error!()` to not exit() directly

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -158,6 +158,15 @@ compiler that has `debug=true` set in `bootstrap.toml`.
 You can set `MIRI_BACKTRACE=1` to get a backtrace of where an
 evaluation error was originally raised.
 
+#### Tracing
+
+You can generate a Chrome trace file from a Miri execution by passing `--features=tracing` during the
+build and then setting `MIRI_TRACING=1` when running Miri. This will generate a `.json` file that
+you can visualize in [Perfetto](https://ui.perfetto.dev/). For example:
+
+```sh
+MIRI_TRACING=1 ./miri run --features=tracing tests/pass/hello.rs
+```
 
 ### UI testing
 
--- a/src/bin/log/setup.rs
+++ b/src/bin/log/setup.rs
@@ -1,0 +1,126 @@
+use std::env::{self, VarError};
+use std::str::FromStr;
+use std::sync::{Mutex, OnceLock};
+
+use rustc_middle::ty::TyCtxt;
+use rustc_session::{CtfeBacktrace, EarlyDiagCtxt};
+
+/// The tracing layer from `tracing-chrome` starts a thread in the background that saves data to
+/// file and closes the file when stopped. If the thread is not stopped properly, the file will be
+/// missing end terminators (`]` for JSON arrays) and other data may also not be flushed. Therefore
+/// we need to keep a guard that, when [Drop]ped, will send a signal to stop the thread. Make sure
+/// to manually drop this guard using [deinit_loggers], if you are exiting the program with
+/// [std::process::exit]!
+#[must_use]
+struct TracingGuard {
+    #[cfg(feature = "tracing")]
+    _chrome: super::tracing_chrome::FlushGuard,
+    _no_construct: (),
+}
+
+// This ensures TracingGuard is always a drop-type, even when the `_chrome` field is disabled.
+impl Drop for TracingGuard {
+    fn drop(&mut self) {}
+}
+
+fn rustc_logger_config() -> rustc_log::LoggerConfig {
+    // Start with the usual env vars.
+    let mut cfg = rustc_log::LoggerConfig::from_env("RUSTC_LOG");
+
+    // Overwrite if MIRI_LOG is set.
+    if let Ok(var) = env::var("MIRI_LOG") {
+        // MIRI_LOG serves as default for RUSTC_LOG, if that is not set.
+        if matches!(cfg.filter, Err(VarError::NotPresent)) {
+            // We try to be a bit clever here: if `MIRI_LOG` is just a single level
+            // used for everything, we only apply it to the parts of rustc that are
+            // CTFE-related. Otherwise, we use it verbatim for `RUSTC_LOG`.
+            // This way, if you set `MIRI_LOG=trace`, you get only the right parts of
+            // rustc traced, but you can also do `MIRI_LOG=miri=trace,rustc_const_eval::interpret=debug`.
+            if tracing::Level::from_str(&var).is_ok() {
+                cfg.filter = Ok(format!(
+                    "rustc_middle::mir::interpret={var},rustc_const_eval::interpret={var},miri={var}"
+                ));
+            } else {
+                cfg.filter = Ok(var);
+            }
+        }
+    }
+
+    cfg
+}
+
+/// The global logger can only be set once per process, so track whether that already happened and
+/// keep a [TracingGuard] so it can be [Drop]ped later using [deinit_loggers].
+static LOGGER_INITED: OnceLock<Mutex<Option<TracingGuard>>> = OnceLock::new();
+
+fn init_logger_once(early_dcx: &EarlyDiagCtxt) {
+    // If the logger is not yet initialized, initialize it.
+    LOGGER_INITED.get_or_init(|| {
+        let guard = if env::var_os("MIRI_TRACING").is_some() {
+            #[cfg(not(feature = "tracing"))]
+            {
+                crate::show_error!(
+                    "fatal error: cannot enable MIRI_TRACING since Miri was not built with the \"tracing\" feature"
+                );
+            }
+
+            #[cfg(feature = "tracing")]
+            {
+                let (chrome_layer, chrome_guard) =
+                    super::tracing_chrome::ChromeLayerBuilder::new().include_args(true).build();
+                rustc_driver::init_logger_with_additional_layer(
+                    early_dcx,
+                    rustc_logger_config(),
+                    || {
+                        tracing_subscriber::layer::SubscriberExt::with(
+                            tracing_subscriber::Registry::default(),
+                            chrome_layer,
+                        )
+                    },
+                );
+
+                Some(TracingGuard { _chrome: chrome_guard, _no_construct: () })
+            }
+        } else {
+            // initialize the logger without any tracing enabled
+            rustc_driver::init_logger(early_dcx, rustc_logger_config());
+            None
+        };
+        Mutex::new(guard)
+    });
+}
+
+pub fn init_early_loggers(early_dcx: &EarlyDiagCtxt) {
+    // We only initialize `rustc` if the env var is set (so the user asked for it).
+    // If it is not set, we avoid initializing now so that we can initialize later with our custom
+    // settings, and *not* log anything for what happens before `miri` starts interpreting.
+    if env::var_os("RUSTC_LOG").is_some() {
+        init_logger_once(early_dcx);
+    }
+}
+
+pub fn init_late_loggers(early_dcx: &EarlyDiagCtxt, tcx: TyCtxt<'_>) {
+    // If the logger is not yet initialized, initialize it.
+    init_logger_once(early_dcx);
+
+    // If `MIRI_BACKTRACE` is set and `RUSTC_CTFE_BACKTRACE` is not, set `RUSTC_CTFE_BACKTRACE`.
+    // Do this late, so we ideally only apply this to Miri's errors.
+    if let Some(val) = env::var_os("MIRI_BACKTRACE") {
+        let ctfe_backtrace = match &*val.to_string_lossy() {
+            "immediate" => CtfeBacktrace::Immediate,
+            "0" => CtfeBacktrace::Disabled,
+            _ => CtfeBacktrace::Capture,
+        };
+        *tcx.sess.ctfe_backtrace.borrow_mut() = ctfe_backtrace;
+    }
+}
+
+/// Must be called before the program terminates to ensure the trace file is closed correctly. Not
+/// doing so will result in invalid trace files. Also see [TracingGuard].
+pub fn deinit_loggers() {
+    if let Some(guard) = LOGGER_INITED.get()
+        && let Ok(mut guard) = guard.lock()
+    {
+        std::mem::drop(guard.take());
+    }
+}
--- a/src/bin/log/setup.rs
+++ b/src/bin/log/setup.rs
@@ -59,7 +59,7 @@ fn init_logger_once(early_dcx: &EarlyDiagCtxt) {
         let guard = if env::var_os("MIRI_TRACING").is_some() {
             #[cfg(not(feature = "tracing"))]
             {
-                crate::show_error!(
+                crate::fatal_error!(
                     "fatal error: cannot enable MIRI_TRACING since Miri was not built with the \"tracing\" feature"
                 );
             }
--- a/src/bin/log/tracing_chrome.rs
+++ b/src/bin/log/tracing_chrome.rs
@@ -1,0 +1,625 @@
+// SPDX-License-Identifier: MIT
+// SPDX-FileCopyrightText: Copyright (c) 2020 Thoren Paulson
+//! This file is taken unmodified from the following link, except for file attributes and
+//! `extern crate` at the top.
+//! https://github.com/thoren-d/tracing-chrome/blob/7e2625ab4aeeef2f0ef9bde9d6258dd181c04472/src/lib.rs
+//! Depending on the tracing-chrome crate from crates.io is unfortunately not possible, since it
+//! depends on `tracing_core` which conflicts with rustc_private's `tracing_core` (meaning it would
+//! not be possible to use the [ChromeLayer] in a context that expects a [Layer] from
+//! rustc_private's `tracing_core` version).
+#![allow(warnings)]
+#![cfg(feature = "tracing")]
+
+// This is here and not in src/lib.rs since it is a direct dependency of tracing_chrome.rs and
+// should not be included if the "tracing" feature is disabled.
+extern crate tracing_core;
+
+use tracing_core::{field::Field, span, Event, Subscriber};
+use tracing_subscriber::{
+    layer::Context,
+    registry::{LookupSpan, SpanRef},
+    Layer,
+};
+
+use serde_json::{json, Value as JsonValue};
+use std::{
+    marker::PhantomData,
+    path::Path,
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc, Mutex,
+    },
+};
+
+use std::io::{BufWriter, Write};
+use std::sync::mpsc;
+use std::sync::mpsc::Sender;
+use std::{
+    cell::{Cell, RefCell},
+    thread::JoinHandle,
+};
+
+thread_local! {
+    static OUT: RefCell<Option<Sender<Message>>> = const { RefCell::new(None) };
+    static TID: RefCell<Option<usize>> = const { RefCell::new(None) };
+}
+
+type NameFn<S> = Box<dyn Fn(&EventOrSpan<'_, '_, S>) -> String + Send + Sync>;
+type Object = serde_json::Map<String, JsonValue>;
+
+/// A [`Layer`](tracing_subscriber::Layer) that writes a Chrome trace file.
+pub struct ChromeLayer<S>
+where
+    S: Subscriber + for<'span> LookupSpan<'span> + Send + Sync,
+{
+    out: Arc<Mutex<Sender<Message>>>,
+    start: std::time::Instant,
+    max_tid: AtomicUsize,
+    include_args: bool,
+    include_locations: bool,
+    trace_style: TraceStyle,
+    name_fn: Option<NameFn<S>>,
+    cat_fn: Option<NameFn<S>>,
+    _inner: PhantomData<S>,
+}
+
+/// A builder for [`ChromeLayer`](crate::ChromeLayer).
+#[derive(Default)]
+pub struct ChromeLayerBuilder<S>
+where
+    S: Subscriber + for<'span> LookupSpan<'span> + Send + Sync,
+{
+    out_writer: Option<Box<dyn Write + Send>>,
+    name_fn: Option<NameFn<S>>,
+    cat_fn: Option<NameFn<S>>,
+    include_args: bool,
+    include_locations: bool,
+    trace_style: TraceStyle,
+    _inner: PhantomData<S>,
+}
+
+/// Decides how traces will be recorded.
+#[derive(Default)]
+pub enum TraceStyle {
+    /// Traces will be recorded as a group of threads.
+    /// In this style, spans should be entered and exited on the same thread.
+    #[default]
+    Threaded,
+
+    /// Traces will recorded as a group of asynchronous operations.
+    Async,
+}
+
+impl<S> ChromeLayerBuilder<S>
+where
+    S: Subscriber + for<'span> LookupSpan<'span> + Send + Sync,
+{
+    pub fn new() -> Self {
+        ChromeLayerBuilder {
+            out_writer: None,
+            name_fn: None,
+            cat_fn: None,
+            include_args: false,
+            include_locations: true,
+            trace_style: TraceStyle::Threaded,
+            _inner: PhantomData,
+        }
+    }
+
+    /// Set the file to which to output the trace.
+    ///
+    /// Defaults to `./trace-{unix epoch in micros}.json`.
+    ///
+    /// # Panics
+    ///
+    /// If `file` could not be opened/created. To handle errors,
+    /// open a file and pass it to [`writer`](crate::ChromeLayerBuilder::writer) instead.
+    pub fn file<P: AsRef<Path>>(self, file: P) -> Self {
+        self.writer(std::fs::File::create(file).expect("Failed to create trace file."))
+    }
+
+    /// Supply an arbitrary writer to which to write trace contents.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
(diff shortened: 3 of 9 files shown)


Stypox:
I am not sure why tests fail, when I tried locally they failed even on the latest commit on `master` ( 2f4f9ac15ae5f2c4e1c95b4dc38172e4be3ec4bb ). I did a `git bisect` and it told me that either ab135f0ea77aed3ed95c358a4a8969f471f56748 or 42f66f42c031936189921805e245b22e0f94109a are the first bad commit (the first one doesn't build, so I skipped it, while the second is the first where `./miri test` failed). However this does not explain why the CI on 2f4f9ac15ae5f2c4e1c95b4dc38172e4be3ec4bb passed.

tiif:
I am on ``x86_64-unknown-linux-gnu``, ``./miri test`` for current master branch (2f4f9ac15) passed locally for me. 

May I know what os are you currently using? And is the failure the same as the one in the CI?

Stypox:
@tiif On my PC the tests that fail run into errors like "error: unsupported operation: extern static `__rust_no_alloc_shim_is_unstable` is not supported by Miri". Even just running ` ./miri run tests/pass/hello.rs` triggerrs the same error. But the CI exits with code 143 without printing any error (I'm looking at ubuntu-latest):
```
2025-06-19T10:55:01.6057618Z ##[error]The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.
2025-06-19T10:55:01.6062974Z ##[error]Process completed with exit code 143.
2025-06-19T10:55:01.6983400Z Cleaning up orphan processes
```


I am also on `x86_64-unknown-linux-gnu`, my OS is Manjaro Linux with kernel 5.15.182-1-MANJARO, I have a AMD Ryzen 7 5800H cpu, let me know if you need anything else.

tiif:
I suspect the CI failure might be triggered by the change in this PR, as there were successful CI runs after this one in https://github.com/rust-lang/miri/actions. (I haven't read the code closely, so this is just a guess  :)

About the unsupported error, I think maybe the toolchain is outdated. You can try to follow the steps here https://github.com/rust-lang/miri/blob/master/CONTRIBUTING.md#preparing-the-build-environment to update it if you haven't done so recently. 

If the problem persists, feel free to ask for help [in zulip](https://rust-lang.zulipchat.com/#narrow/channel/269128-miri).

Stypox:
The tests pass locally. I rebased on master and force pushed to make the CI rerun, but it still fails.

Both now and earlier, I think the cause for the CI failure was "memory allocation of 408 bytes failed" on windows-latest (see [here](https://github.com/rust-lang/miri/actions/runs/15872693022/job/44752825107?pr=4406#step:5:761) and [here](https://github.com/rust-lang/miri/actions/runs/15755727571/job/44410624914?pr=4406#step:5:743)). mac also fails due to "No space left on device (os error 28)" (see [here](https://github.com/rust-lang/miri/actions/runs/15872693022/job/44752825087?pr=4406#step:5:810)). I don't understand why ubuntu fails with code 143 without printing any error.

I don't understand how changes in this PR can cause the tests to use more memory (assuming that's the culprit). The behavior of the code in this PR should be the same as before when no `--tracing` is passed.

Stypox:
After having everything already compiled in cache, `/usr/bin/time -v ./miri test` reports "Maximum resident set size (kbytes): 284044" on this PR and "Maximum resident set size (kbytes): 284608" on the master branch, so there does not seem to be any difference I can reproduce on my device.

ci/ci.sh
```diff
@@ -28,16 +28,16 @@ begingroup "Building Miri"
 export RUSTFLAGS="-D warnings"
 export CARGO_INCREMENTAL=0
 export CARGO_EXTRA_FLAGS="--locked"
+# We enable all features (except tracing) to make sure the Stacked Borrows consistency check runs.
+export FEATURES="--features=genmc,stack-cache,stack-cache-consistency-check"
 
 # Determine configuration for installed build (used by test-cargo-miri and `./miri bench`).
 echo "Installing release version of Miri"
-time ./miri install
+time ./miri install "$FEATURES"
```
Stypox: The change in this line changes the behavior of `test-cargo-miri` and `./miri bench` down the line by enabling the `genmc` and `stack-cache-consistency-check` features which would otherwise be disabled by default. Since I couldn't find a comment explaining why those features are not being enabled here too, I made this change, but it's not really necessary. Is this ok or should I revert this change?

RalfJung: `stack-cache-consistency-check` makes things quite slow so it was deliberately not enabled here.

src/trace/tracing_chrome.rs
```diff
@@ -0,0 +1,623 @@
+//! This file is taken unmodified from the following link, except for file attributes at the top:
+//! https://github.com/thoren-d/tracing-chrome/blob/7e2625ab4aeeef2f0ef9bde9d6258dd181c04472/src/lib.rs
+//! Depending on the tracing-chrome crate from crates.io is unfortunately not possible, since it
+//! depends on `tracing_core` which conflicts with rustc_private's `tracing_core` (meaning it would
+//! not be possible to use the [ChromeLayer] in a context that expects a [Layer] from from
+//! rustc_private's `tracing_core` version).
```
RalfJung: Please explain the licensing situation to ensure we got that covered, too.

src/bin/miri.rs
```diff
@@ -63,6 +64,7 @@ struct MiriCompilerCalls {
     many_seeds: Option<ManySeedsConfig>,
     /// Settings for using GenMC with Miri.
     genmc_config: Option<GenmcConfig>,
+    tracing_guard: Option<TracingGuard>,
```
RalfJung: Why do we need a new field here and what does this do?

src/bin/miri.rs
```diff
@@ -211,6 +219,10 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
                 if return_code != rustc_driver::EXIT_SUCCESS {
                     eprintln!("FAILING SEED: {seed}");
                     if !many_seeds.keep_going {
+                        // drop the tracing guard before exiting, so tracing calls are flushed correctly
+                        if let Ok(mut lock) = tracing_guard.try_lock() {
+                            let _guard_being_dropped = (*lock).take();
```
RalfJung: An explicit call to `drop` seems more clear here.

But it's not great tha we apparently have to remember to do this in multiple places...

src/bin/miri.rs
```diff
@@ -223,6 +235,9 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
             if num_failed > 0 {
                 eprintln!("{num_failed}/{total} SEEDS FAILED", total = many_seeds.seeds.count());
             }
+            // drop the tracing guard before exiting, so tracing calls are flushed correctly
+            #[allow(clippy::drop_non_drop)]
```
RalfJung: You can probably avoid the lint by adding `impl Drop for TracingGuard` to make sure it is always a drop-type.

src/bin/miri.rs
```diff
@@ -336,74 +355,19 @@ macro_rules! show_error {
...
-            "0" => CtfeBacktrace::Disabled,
-            _ => CtfeBacktrace::Capture,
-        };
-        *tcx.sess.ctfe_backtrace.borrow_mut() = ctfe_backtrace;
-    }
-}
-
 /// Execute a compiler with the given CLI arguments and callbacks.
+/// Does not call [std::process::exit] directly, but rather returns an [ExitCode],
+/// to allow the tracing guard to be dropped before exiting.
```
RalfJung: With those new semantics, the function name no longer makes sense.

src/trace/setup.rs
```diff
@@ -0,0 +1,94 @@
...
+    let mut guard = None;
+    LOGGER_INITED.call_once(|| {
+        #[cfg(feature = "tracing")]
+        let (chrome_layer, chrome_guard) =
+            super::tracing_chrome::ChromeLayerBuilder::new().include_args(true).build();
+        guard = Some(TracingGuard {
+            #[cfg(feature = "tracing")]
+            _chrome: chrome_guard,
+            _no_construct: (),
+        });
```
RalfJung: Why does this guard have toe be created here? This then causes all sorts of complexity since you can't know whether it gets created in early-init or late-init.

It seems better to just create the guard separately?

src/trace/setup.rs
RalfJung, line 1: I'm not convinced that this file should live here. That makes the responsibility sharing between the bin crate and the lib crate a lot more complicated.

Can this live as a separate file in the `bin` folder?

src/trace/setup.rs
```diff
@@ -0,0 +1,94 @@
+use std::env::{self, VarError};
+use std::str::FromStr;
+use std::sync::Once;
+
+use rustc_middle::ty::TyCtxt;
+use rustc_session::{CtfeBacktrace, EarlyDiagCtxt};
+use tracing_subscriber::Registry;
+
+#[must_use]
+pub struct TracingGuard {
```
RalfJung: Please add a doc comment to this type.

src/trace/tracing_chrome.rs
```diff
@@ -0,0 +1,623 @@
+//! This file is taken unmodified from the following link, except for file attributes at the top:
```
RalfJung: And except for the `extern crate` I guess?

src/trace/tracing_chrome.rs
```diff
@@ -0,0 +1,623 @@
...
+//! Depending on the tracing-chrome crate from crates.io is unfortunately not possible, since it
+//! depends on `tracing_core` which conflicts with rustc_private's `tracing_core` (meaning it would
+//! not be possible to use the [ChromeLayer] in a context that expects a [Layer] from from
+//! rustc_private's `tracing_core` version).
+#![allow(warnings)]
+#![cfg(feature = "tracing")]
+
+// this is here and not in src/lib.rs since it is a direct dependency of tracing_chrome.rs and
+// should not be included if the "tracing" feature is disabled
+extern crate tracing_core;
```
RalfJung:
```suggestion
// This is here and not in src/lib.rs since it is a direct dependency of tracing_chrome.rs and
// should not be included if the "tracing" feature is disabled.
extern crate tracing_core;
```

src/bin/miri.rs
Stypox: I added a doc comment

src/bin/miri.rs
Stypox: Yeah I also don't like that much, but it felt the safest option. An alternative would be to call `libc::atexit()` or to use a crate like https://crates.io/crates/ctor, though both options involve `unsafe`.

src/trace/setup.rs
Stypox: What do you mean by creating the guard separately? An alternative would be to pass a parameter around but then we'd have to do the `.is_none()` check anyway so I don't see better alternatives.

Stypox:
- Removed $FEATURES from `ci.sh`'s `./miri install`
- Added docs to TracingGuard, improved docs in tracing_chrome.rs and made other comments follow the "Abc def." style (instead of "abc def")
- `impl Drop for TracingGuard`
- Renamed `run_compiler_and_exit` to `run_compiler_return_exit_code`
- Moved tracing_chrome and logger setup into `bin/trace` folder. Putting them under `bin/` directly meant that each file would be interpreted as a binary as far as I understand, that's why I had to put them under an additional `trace/` subfolder.

src/bin/miri.rs
```diff
@@ -156,7 +159,11 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
...
 
         let early_dcx = EarlyDiagCtxt::new(tcx.sess.opts.error_format);
-        init_late_loggers(&early_dcx, tcx);
+        let tracing_guard = init_late_loggers(&early_dcx, tcx);
+        if self.tracing_guard.is_none() {
+            // either tracing_guard or self.tracing_guard are None, due to LOGGER_INITED.call_once()
+            self.tracing_guard = tracing_guard;
+        }
         if !tcx.crate_types().contains(&CrateType::Executable) {
             tcx.dcx().fatal("miri only makes sense on bin crates");
```
RalfJung, line 164: FWIW this also exits the process, so the `init_late_loggers` should probably be moved down.

src/bin/miri.rs
RalfJung: Here's an idea: move everything in `fn after_analysis` that is after `init_late_loggers` into a separate inherent method on `MiriCompilerCalls` that returns an `ExitCode`. Have the guard inside that separate method so it gets dropped when the method returns, and then there's just a single `exit` call in `after_analysis` afterwards.

Stypox: I thought of that but `exit()` is also used inside the `par_for_each_in` to exit immediately as soon as anything fails (when `many_seeds.keep_going` is false) and afaik there is no easy and safe way to stop executing threads to then drop the guard and return

src/bin/miri.rs
Stypox: Right, fixed

src/trace/setup.rs
RalfJung: I mean to have `init_*_logger` store the guard somewhere, e.g. inside `LOGGER_INITED`, and then have a separate function that fetches the guard from there (and panics if it has already been fetched / not yet been placed there).

src/bin/miri.rs
RalfJung: Hm... yeah true, `par_for_each_in` does not support early-return.

Well, then let's at least have a local `exit` closure inside this function to reduce the code duplication.

src/trace/setup.rs
RalfJung: `LOGGER_INITED` should probably become a [`OnceLock`](https://doc.rust-lang.org/nightly/std/sync/struct.OnceLock.html) for this.

Stypox: The guard needs to be fetched from `main()` and also from `after_analysis`. Having the guard in `main()` is needed to do the cleanup when the compiler exits. I pushed a commit that implements your suggestion and takes this observation into consideration. While the interface is nicer, it may lead to more confusion, what do you think?

RalfJung: Why do we even care about the contents of the trace file when there is an error?^^

But anyway, that last version works for me, thanks.

src/bin/miri.rs
RalfJung: > Well, then let's at least have a local exit closure inside this function to reduce the code duplication.

I don't think this has been done?
With the new API, what we probably want is a `fn exit` that calls `deinit_loggers`, and then use that instead of the one from `std::process`.

Stypox: Oh, I had extracted the `exit()` out of the ifs, but ok now I made an fn

src/bin/trace/tracing_chrome.rs
```diff
@@ -0,0 +1,624 @@
+//! This file is taken unmodified from the following link, except for file attributes and
+//! `extern crate` at the top. It is licensed under MIT.
```
RalfJung: The code is this repo is otherwise under MIT OR Apache2. I don't know if we have to do anything to license one file differently... I will have to [ask about this](https://rust-lang.zulipchat.com/#narrow/channel/392734-council/topic/Vendoring.20an.20MIT-licensed.20file/with/526194636).

src/bin/miri.rs
```diff
@@ -336,74 +350,19 @@ macro_rules! show_error {
...
-    // Invoke compiler, and handle return code.
-    let exit_code =
-        rustc_driver::catch_with_exit_code(move || rustc_driver::run_compiler(args, callbacks));
-    std::process::exit(exit_code)
+) -> ExitCode {
+    // Invoke compiler, catch any unwinding panics and handle return code.
+    match rustc_driver::catch_fatal_errors(move || rustc_driver::run_compiler(args, callbacks)) {
+        Ok(()) => ExitCode::SUCCESS,
+        _ => ExitCode::FAILURE,
+    }
```
RalfJung: This changes behavior -- we no longer use the exit code that `catch_with_exit_code` gives us. Please don't change behavior.

src/bin/miri.rs
```diff
@@ -336,74 +350,19 @@ macro_rules! show_error {
...
-        *tcx.sess.ctfe_backtrace.borrow_mut() = ctfe_backtrace;
-    }
-}
-
 /// Execute a compiler with the given CLI arguments and callbacks.
-fn run_compiler_and_exit(
+/// Does not call [std::process::exit] directly, but rather returns an [ExitCode],
+/// to allow the tracing guard to be dropped before exiting.
+#[must_use]
+fn run_compiler_return_exit_code(
```
RalfJung: Now that we have a new `exit` function, I think the easiest thing to do is to keep the old `run_compiler_and_exit` and just use the new `exit`. Then we don't need the `DeinitLoggersOnDrop` at all I think.

src/bin/miri.rs
```diff
@@ -793,6 +749,12 @@ fn main() {
         );
     };
 
+    // Init loggers the Miri way. Do so after arguments have been parsed, so no tracing file is
+    // created if argument parsing fails, and also so that we don't have to worry about dropping
+    // the guard before calling std::process::exit() in show_error!().
```
RalfJung: `show_error` could just use the new `exit`?

ci/ci.sh
```diff
@@ -28,16 +28,17 @@ begingroup "Building Miri"
 export RUSTFLAGS="-D warnings"
 export CARGO_INCREMENTAL=0
 export CARGO_EXTRA_FLAGS="--locked"
+# We enable all features (except tracing) to make sure the Stacked Borrows consistency check runs.
+export FEATURES="--features=genmc,stack-cache,stack-cache-consistency-check"
```
RalfJung: I am not happy about having to list the features here. I think my proposal would be that enabling the `tracing` feature just adds the corresponding code to the binary, but doesn't actually enable tracing. For that, we additionally have an env var, like `MIRI_TRACING` or so. Then one can also switch between traced and untraced builds more easily at runtime.

src/bin/trace/tracing_chrome.rs
RalfJung, line 2:
```suggestion
//! `extern crate` at the top. It is licensed under the MIT license,
//! Copyright (c) 2020 Thoren Paulson.
```

src/bin/miri.rs
```diff
@@ -232,8 +246,8 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
                     tcx.dcx().abort_if_errors();
                     rustc_driver::EXIT_FAILURE
                 });
-            std::process::exit(return_code);
-        }
+            exit(return_code);
+        };
```
```suggestion
        }
```

src/bin/trace/mod.rs
RalfJung, line 1: This doesn't just handle tracing, it handles all logging. So maybe `log` is a better name for the module? And then `logger_setup` can become just `setup`.

src/bin/trace/tracing_chrome.rs
RalfJung:
```suggestion
// SPDX-License-Identifier: MIT
// Copyright (c) 2020 Thoren Paulson
//! This file is taken unmodified from the following link, except for file attributes and
//! `extern crate` at the top:
```

src/bin/miri.rs
Stypox: Actually I had just inlined `catch_with_exit_code` to use `ExitCode::SUCCESS` and `ExitCode::FAILURE` instead of the constants `EXIT_SUCCESS` and `EXIT_FAILURE`, but their values are the same. Anyway, I reverted to the previous behavior, with in addition `deinit_loggers`.

RalfJung: Ah, I see. Still better if we can share that logic.

src/bin/trace/tracing_chrome.rs
Stypox: Thanks for asking there, I checked how it's done in other files and posted a comment to get confirmation that it's ok

Stypox:
Note that building with `--features=tracing` and then not passing `MIRI_TRACING` will still make the tracing calls into the tracing library (which will be printed to terminal if MIRI_LOG is set), it just won't save a trace file.

@rustbot ready

src/bin/log/tracing_chrome.rs
```diff
@@ -0,0 +1,625 @@
+// SPDX-License-Identifier: MIT
+// SPDX-FileCopyrightText: 2020 Thoren Paulson
```
RalfJung: Let's preserve the copyright note in full, matching the examples for this SPDX field
```suggestion
// SPDX-FileCopyrightText: Copyright (c) 2020 Thoren Paulson
```

src/bin/log/tracing_chrome.rs
```diff
@@ -0,0 +1,625 @@
...
+//! This file is taken unmodified from the following link, except for file attributes and
+//! `extern crate` at the top. It is licensed under MIT.
```
```suggestion
//! `extern crate` at the top.
```
This is now redundant.

CONTRIBUTING.md
```diff
@@ -158,6 +158,15 @@ compiler that has `debug=true` set in `bootstrap.toml`.
 You can set `MIRI_BACKTRACE=1` to get a backtrace of where an
 evaluation error was originally raised.
 
+#### Tracing
+
+You can generate a Chrome trace file from a Miri execution by passing `--features=tracing` during
```
RalfJung:
```suggestion
You can generate a Chrome trace file from a Miri execution by passing `--features=tracing` during the
```

src/bin/log/setup.rs
```diff
@@ -0,0 +1,127 @@
...
+fn init_logger_once(early_dcx: &EarlyDiagCtxt) {
+    // If the logger is not yet initialized, initialize it.
+    LOGGER_INITED.get_or_init(|| {
+        let guard = if env::var_os("MIRI_TRACING").is_some() {
+            #[cfg(not(feature = "tracing"))]
+            {
+                eprintln!(
+                    "fatal error: cannot enable MIRI_TRACING since Miri was not built with the \"tracing\" feature"
+                );
+                std::process::exit(1)
```
RalfJung: Can this use the `show_error!` macro that we use elsewhere in the miri binary setup?

RalfJung:
I have taken care of the last minor nits, let's get this finally landed. :)

CONTRIBUTING.md
```diff
@@ -158,6 +158,15 @@ compiler that has `debug=true` set in `bootstrap.toml`.
 You can set `MIRI_BACKTRACE=1` to get a backtrace of where an
 evaluation error was originally raised.
 
+#### Tracing
+
+You can generate a Chrome trace file from a Miri execution by passing `--features=tracing` during the
+build and then setting `MIRI_TRACING=1` when running Miri. This will generate a `.json` file that
+you can visualize in [Perfetto](https://ui.perfetto.dev/). For example:
```
RalfJung, line 165: I wonder, does https://profiler.firefox.com/ also work? Apparently it can do more than just Firefox traces.

Stypox: Yeah it works too, though it visualizes the trace without the tree structure. Should I add it there?

image

RalfJung: Is it still useful or does one need the tree to deal with the data?
I have no idea what it looks like in Perfetto.

Stypox: This is how it looks in Perfetto. I feel like the tree view is especially useful to see which spans are nested, which is hard to do in the firefox visualizer without zooming in enough. 

image

PR #4408, opened 2025-06-19
Only write join return value once join succeeds

Patrick-6:
This PR improves how thread joining is handled, by ensuring that the write of the return value only happens once the join succeeds. Before this write happened immediately, even if the thread is blocked by the join.

### DIFF
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -582,88 +582,6 @@ impl<'tcx> ThreadManager<'tcx> {
         interp_ok(())
     }
 
-    /// Mark that the active thread tries to join the thread with `joined_thread_id`.
-    fn join_thread(
-        &mut self,
-        joined_thread_id: ThreadId,
-        data_race_handler: &mut GlobalDataRaceHandler,
-    ) -> InterpResult<'tcx> {
-        if self.threads[joined_thread_id].join_status == ThreadJoinStatus::Detached {
-            // On Windows this corresponds to joining on a closed handle.
-            throw_ub_format!("trying to join a detached thread");
-        }
-
-        fn after_join<'tcx>(
-            threads: &mut ThreadManager<'_>,
-            joined_thread_id: ThreadId,
-            data_race_handler: &mut GlobalDataRaceHandler,
-        ) -> InterpResult<'tcx> {
-            match data_race_handler {
-                GlobalDataRaceHandler::None => {}
-                GlobalDataRaceHandler::Vclocks(data_race) =>
-                    data_race.thread_joined(threads, joined_thread_id),
-                GlobalDataRaceHandler::Genmc(genmc_ctx) =>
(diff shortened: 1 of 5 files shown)


src/concurrency/thread.rs
```diff
@@ -1114,20 +1032,103 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-    #[inline]
-    fn join_thread(&mut self, joined_thread_id: ThreadId) -> InterpResult<'tcx> {
+    /// Mark that the active thread tries to join the thread with `joined_thread_id`.
+    ///
+    /// When the join is successful (immediately, or as soon as the joined thread finishes), `success_retval` will be written to `return_dest`.
+    fn join_thread(
+        &mut self,
+        joined_thread_id: ThreadId,
+        success_retval: Scalar,
+        return_dest: &MPlaceTy<'tcx>,
```
Patrick-6, line 1042: Should these be an `Option`?

src/concurrency/thread.rs
```diff
@@ -1114,20 +1032,103 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        assert!(
+            this.machine
+                .threads
+                .threads
+                .iter()
+                .all(|thread| { !thread.state.is_blocked_on(BlockReason::Join(joined_thread_id)) }),
+            "this thread already has threads waiting for its termination"
+        );
+
+        self.join_thread(joined_thread_id, success_retval, return_dest)
```
```suggestion
        this.join_thread(joined_thread_id, success_retval, return_dest)
```

src/concurrency/thread.rs
```diff
@@ -1114,20 +1032,103 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        joined_thread_id: ThreadId,
+        success_retval: Scalar,
+        return_dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx> {
         let this = self.eval_context_mut();
-        this.machine
-            .threads
-            .join_thread_exclusive(joined_thread_id, &mut this.machine.data_race)?;
-        interp_ok(())
+        if this.machine.threads.threads[joined_thread_id].join_status == ThreadJoinStatus::Joined {
```
RalfJung: Seems like it might be worth to have a `let thread_mgt = &this.machine.threads;` alias?

(We should probably rename `machine.threads` to `machine.thread_mgr`... another time)

src/concurrency/thread.rs
RalfJung: What would be the point of that?

Patrick-6: In case that any future join implementation doesn't do a write, but I guess it can be changed if that ever happens.

RalfJung:

@rustbot author

PR #4411, opened 2025-06-20
Remove leaky synchronisation objects.

### LINKED ISSUE #3967: Get rid of IDs for synchronization primitives, make pthread shims not leaky

LorrensP-2158466:
fixes #3967 

The pr is structured in 4 commits, the first two replace the use of `Id` with `Ref` objects. The third commit refactors the `declare_id!` macro into a macro that creates these `Ref` objects for us. The last commit removes the usage of `SynchronisationObjects`.

### DIFF
--- a/src/concurrency/sync.rs
+++ b/src/concurrency/sync.rs
@@ -8,16 +8,16 @@ use std::time::Duration;
 
 use rustc_abi::Size;
 use rustc_data_structures::fx::FxHashMap;
-use rustc_index::IndexVec;
 
-use super::init_once::InitOnce;
 use super::vector_clock::VClock;
+use crate::concurrency::init_once::InitOnceRef;
 use crate::*;
 
 /// We cannot use the `newtype_index!` macro because we have to use 0 as a
 /// sentinel value meaning that the identifier is not assigned. This is because
 /// the pthreads static initializers initialize memory with zeros (see the
 /// `src/shims/sync.rs` file).
+#[allow(unused)]
 macro_rules! declare_id {
     ($name: ident) => {
         /// 0 is used to indicate that the id was not yet assigned and,
@@ -45,7 +45,6 @@ macro_rules! declare_id {
         }
     };
 }
-pub(super) use declare_id;
 
 /// The mutex state.
 #[derive(Default, Debug)]
@@ -226,9 +225,7 @@ struct FutexWaiter {
 
 /// The state of all synchronization objects.
 #[derive(Default, Debug)]
-pub struct SynchronizationObjects {
-    pub(super) init_onces: IndexVec<InitOnceId, InitOnce>,
-}
+pub struct SynchronizationObjects {}
 
 // Private extension trait for local helper methods
 impl<'tcx> EvalContextExtPriv<'tcx> for crate::MiriInterpCx<'tcx> {}
@@ -265,8 +262,8 @@ impl SynchronizationObjects {
         CondvarRef::new()
     }
 
-    pub fn init_once_create(&mut self) -> InitOnceId {
-        self.init_onces.push(Default::default())
+    pub fn init_once_create(&mut self) -> InitOnceRef {
+        InitOnceRef::new()
     }
 }
 
--- a/src/concurrency/sync.rs
+++ b/src/concurrency/sync.rs
@@ -1,9 +1,7 @@
-use std::cell::RefCell;
 use std::collections::VecDeque;
 use std::collections::hash_map::Entry;
 use std::default::Default;
 use std::ops::Not;
-use std::rc::Rc;
 use std::time::Duration;
 
 use rustc_abi::Size;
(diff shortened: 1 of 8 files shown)


src/concurrency/sync.rs
```diff
@@ -739,18 +748,19 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
 
         // Each condvar signal happens-before the end of the condvar wake
         if let Some(data_race) = this.machine.data_race.as_vclocks_ref() {
-            data_race.release_clock(&this.machine.threads, |clock| condvar.clock.clone_from(clock));
+            data_race.release_clock(&this.machine.threads, |clock| {
+                condvar_ref.0.borrow_mut().clock.clone_from(clock)
+            });
         }
-        let Some(waiter) = condvar.waiters.pop_front() else {
+        let Some(waiter) = condvar_ref.0.borrow_mut().waiters.pop_front() else {
```
RalfJung: Please don't call `borrow_mut` twice.

src/concurrency/sync.rs
```diff
@@ -13,39 +11,30 @@ use super::vector_clock::VClock;
...
-#[allow(unused)]
-macro_rules! declare_id {
-    ($name: ident) => {
-        /// 0 is used to indicate that the id was not yet assigned and,
-        /// therefore, is not a valid identifier.
-        #[derive(Clone, Copy, Debug, PartialOrd, Ord, PartialEq, Eq, Hash)]
-        pub struct $name(std::num::NonZero<u32>);
-
-        impl $crate::VisitProvenance for $name {
+macro_rules! sync_obj_ref {
```
RalfJung: I don't think this macro carries its weight -- and it makes some hidden assumptions, such as there not being any provenance in the `$sync_obj` type.

I'd prefer to just remove this commit.

RalfJung (commented):
Thanks a lot! Looks good overall, except I don't think this macro is worth it -- sorry.

src/concurrency/sync.rs
LorrensP-2158466: Yeah, understandable, I'll remove it!

RalfJung:
Ah, you force-pushed... please avoid that if at all possible (except when I ask you to do it), it makes it a lot harder to track what happened since the last review. (This is a pretty fatal flaw in Github, but unless they eventually get better at dealing with force-pushes we all have to work around it manually.)

RalfJung:
In this case you probably had to force-push to remove that macro commit, but you did not have to incorporate any of the recent changes in master. `git rebase --keep-base` helps in that case.

src/concurrency/sync.rs
```diff
@@ -59,24 +26,29 @@ struct Mutex {
     clock: VClock,
 }
 
+impl Mutex {
+    /// Get the id of the thread that currently owns this lock, or `None` if it is not locked.
+    pub fn owner(&self) -> Option<ThreadId> {
+        self.owner
+    }
```
RalfJung: I had deliberately avoided adding these methods when they are not needed, as it seemed like pointless boilerplate. What is the reason why you added them now?

src/concurrency/sync.rs
```diff
@@ -137,8 +109,8 @@ impl RwLock {
 pub struct RwLockRef(Rc<RefCell<RwLock>>);
 
 impl RwLockRef {
-    fn new() -> Self {
-        RwLockRef(Rc::new(RefCell::new(RwLock::default())))
+    pub fn new() -> Self {
+        Self(Rc::new(RefCell::new(Default::default())))
```
RalfJung: Why did you make this pub?

And why did you remove the explicit type? If you want to go for maximal implicitness, `Self(Default::default())` should work.

RalfJung:
The existing concerns are resolved, but you added a bunch of new changes I didn't ask for so there's new concerns.^^
@rustbot author

src/concurrency/sync.rs
LorrensP-2158466: > Why did you make this pub?
Shims from specific platforms use these `new()` functions to create their sync objects

Previously this was done with the `SynchronisationObjects` struct, but this was now redundant because we no longer use `ID`s.

> And why did you remove the explicit type? If you want to go for maximal implicitness, Self(Default::default()) should work.

I just copied and pasted the macro expansion results. Is there a specific way you'd prefer?

LorrensP-2158466:
> Ah, you force-pushed... please avoid that if at all possible (except when I ask you to do it), it makes it a lot harder to track what happened since the last review.

Ah shit, sorry. I'll remember that!

src/concurrency/sync.rs
RalfJung: Not sure what you mean? Functions like this already exist previously:

https://github.com/rust-lang/miri/blob/8d809daaffbf158db617c76a3831294405460cb9/src/concurrency/sync.rs#L67-L69

You just changed them a bit in style, not content, it seems. Or am I missing something?

src/concurrency/sync.rs
LorrensP-2158466: It seems cleaner to me to have this specific function on the object itself and not on the ref, but I'll move them!

src/concurrency/sync.rs
LorrensP-2158466: Yes, but the only place that it is called in is in the constructor functions of `SynchronisationObjects`:
https://github.com/rust-lang/miri/blob/efc63a41c61f2ffb3b81a8005709d0ccf6c9d36b/src/concurrency/sync.rs#L240-L246

So there wasn't any need for them to be public, now there is because `SynchronisationObjects` doesn't exist anymore because it's redundant.

RalfJung: Ah right, that explains the `pub`. :)
It doesn't explain the other refactoring, but :shrug:

LorrensP-2158466: > It doesn't explain the other refactoring, but 🤷

Yeah, I do not have a specific reason why I did it this way, just copied and pasted the macro expansion. 
> If you want to go for maximal implicitness, Self(Default::default())

Looks nice, so I'll be doing that. :)

LorrensP-2158466:
Tests should be fine, the latest changes were just some refactorings.

@rustbot ready

RalfJung:

@rustbot author

PR #4414, opened 2025-06-26
Support weak definitions

bjorn3:
When a symbol only has a weak definition, this definition will be picked. When a symbol has both a weak and a regular definition, the regular definition will be picked instead.

In the future the allocator shim may be replaced by weak definitions. This already adds support to simplify things in the future.

### DIFF
--- a/src/shims/foreign_items.rs
+++ b/src/shims/foreign_items.rs
@@ -5,6 +5,7 @@ use std::path::Path;
 use rustc_abi::{Align, AlignFromBytesError, CanonAbi, Size};
 use rustc_apfloat::Float;
 use rustc_ast::expand::allocator::alloc_error_handler_name;
+use rustc_hir::attrs::Linkage;
 use rustc_hir::def::DefKind;
 use rustc_hir::def_id::CrateNum;
 use rustc_middle::middle::codegen_fn_attrs::CodegenFnAttrFlags;
@@ -138,7 +139,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             Entry::Occupied(e) => e.into_mut(),
             Entry::Vacant(e) => {
                 // Find it if it was not cached.
-                let mut instance_and_crate: Option<(ty::Instance<'_>, CrateNum)> = None;
+
+                struct SymbolTarget<'tcx> {
+                    instance: ty::Instance<'tcx>,
+                    cnum: CrateNum,
+                    is_weak: bool,
+                }
+                let mut symbol_target: Option<SymbolTarget<'tcx>> = None;
                 helpers::iter_exported_symbols(tcx, |cnum, def_id| {
                     let attrs = tcx.codegen_fn_attrs(def_id);
                     // Skip over imports of items.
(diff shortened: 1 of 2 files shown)


bjorn3:
This has been extracted out of https://github.com/rust-lang/rust/pull/134522.

tests/ui.rs
```diff
@@ -198,6 +198,11 @@ fn run_tests(
     }
     config.program.args.push("-Zui-testing".into());
 
+    // Necessary for aux-build to be able to build libraries
+    config.custom_comments.insert("miri-be-rustc", |parser, _args, _span| {
+        parser.env_vars.push(("MIRI_BE_RUSTC".to_owned(), "target".to_owned()));
+    });
```
bjorn3: Seems like the test I added is the first time aux-build is used in the miri test suite.

RalfJung: Yeah, we don't use aux-builds, it was never even meant to work.^^

bjorn3: CI fails because MIRI_BE_RUSTC causes miri to error on `-Zmiri-*` flags, but CI passes one through `MIRIFLAGS`. ui_test doesn't seem to have a way to remove arguments. I guess I could use a single file and take advantage of the fact that miri (currently) skips the duplicate symbol check.

bjorn3: Merged it into a single file for now.

src/shims/foreign_items.rs
```diff
@@ -138,7 +139,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             Entry::Occupied(e) => e.into_mut(),
             Entry::Vacant(e) => {
                 // Find it if it was not cached.
-                let mut instance_and_crate: Option<(ty::Instance<'_>, CrateNum)> = None;
+                let mut instance_and_crate: Option<(ty::Instance<'_>, CrateNum, bool)> = None;
```
RalfJung: Please explain what the `bool` means, or use a self-explaining type. It might be worth using a struct instead of a tuple.

src/shims/foreign_items.rs
```diff
@@ -155,39 +156,70 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                                    link_name,
-                                    first: span,
-                                    first_crate: tcx.crate_name(cnum),
-                                    second: original_span,
-                                    second_crate: tcx.crate_name(original_cnum),
-                                });
+                        if let Some((original_instance, original_cnum, original_is_weak)) =
+                            instance_and_crate
+                        {
+                            match (is_weak, original_is_weak) {
```
RalfJung:
```suggestion
                            // There is more than one definition with this name. What we do now
                            // depends on whether one or both definitions are weak.
                            match (is_weak, original_is_weak) {
```

src/shims/foreign_items.rs
```diff
@@ -155,39 +156,70 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                                    instance_and_crate =
+                                        Some((ty::Instance::mono(tcx, def_id), cnum, is_weak));
+                                }
+                                (true, false) => {
+                                    // Current definition is a weak definition. Keep the original one.
+                                }
+                                (true, true) | (false, false) => {
+                                    // Either both definitions are non-weak or both are weak. In
+                                    // either case return an error. For weak definitions we error
+                                    // because it is undefined which definition would have been
```
RalfJung:
```suggestion
                                    // because it is unspecified which definition would have been
```

src/shims/foreign_items.rs
```diff
@@ -155,39 +156,70 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                 })?;
 
+                if let Some((instance, _, _)) = instance_and_crate {
+                    if !matches!(tcx.def_kind(instance.def_id()), DefKind::Fn | DefKind::AssocFn) {
+                        throw_ub_format!(
+                            "attempt to call an exported symbol that is not defined as a function"
+                        );
+                    }
+                }
+
```
RalfJung: I guess this logic can live here as well, but is there a specific reason you moved it?

tests/pass/function_calls/weak_definition.rs
RalfJung, line 1: Please rename this to `exported_symbol_weak` for better test grouping.

jdonszelmann:
what's the status of this one @bjorn3? (I expect just little time which is very fair <3)

bjorn3:
Somehow I didn't notice there were outstanding review comments.

src/shims/foreign_items.rs
bjorn3: Not that I recall.

src/shims/foreign_items.rs
bjorn3: Introduced a struct.

src/shims/foreign_items.rs
```diff
@@ -155,40 +162,82 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                                    first: span,
-                                    first_crate: tcx.crate_name(cnum),
-                                    second: original_span,
-                                    second_crate: tcx.crate_name(original_cnum),
-                                });
+                        if let Some(SymbolTarget {
+                            instance: original_instance,
+                            cnum: original_cnum,
+                            is_weak: original_is_weak,
+                        }) = symbol_target
```
RalfJung:
```suggestion
                        if let Some(original) = symbol_target
```
seems more concise and more clear

src/shims/foreign_items.rs
RalfJung: It will make a difference if there's a weak symbol that's not a function and a "strong" symbol that is. That's probably fine?

src/shims/foreign_items.rs
```diff
@@ -155,40 +162,82 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                                "attempt to call an exported symbol that is not defined as a function"
-                            );
-                        }
-                        instance_and_crate = Some((ty::Instance::mono(tcx, def_id), cnum));
                     }
                     interp_ok(())
                 })?;
 
-                e.insert(instance_and_crate.map(|ic| ic.0))
+                if let Some(SymbolTarget { instance, .. }) = symbol_target {
```
RalfJung, line 230:
```suggestion
                // Once we identified the instance corresponding to the symbol, ensure
                // it is a function. It is okay to encounter non-functions in the search above
                // as long as the final instance we arrive at is a function.
                if let Some(SymbolTarget { instance, .. }) = symbol_target {
```

RalfJung:
Thanks!
This is mostly ready, but CI isn't happy, and also please squash the commits.
@rustbot author

bjorn3:
Rebased and squashed.

@rustbot ready

PR #4418, opened 2025-06-28
various minor native-lib-tracing tweaks

RalfJung:
Follow-up to https://github.com/rust-lang/miri/pull/4401 and https://github.com/rust-lang/miri/pull/4405. In particular, disable native-lib tracing mode by default, since it is not yet fully implemented.

Cc @nia-e


nia-e:
I see most of the points from #4401 were addressed here, ty!

Issue #4419, opened 2025-06-28
Check that Miri builds on Linux ARM hosts


RalfJung:
Ever since https://github.com/rust-lang/miri/pull/4405, we have some `cfg` logic in Miri that specifically only kicks in on Linux ARM builds (with somewhat separate codepaths for 32bit and 64bit). However, we don't have any such build in CI. In fact, we don't even have an x86-32 Linux build in CI.

We should either find a way to at least do a `./miri check` for that host in CI, or remove the code. Entirely untested code is just a disaster waiting to happen. And sadly Rust isn't very good at cross-compiling so this will probably require somehow running on a host with the desired architecture.

@marcoieni rustc now seems to have aarch64 Linux runners for their CI, or some sort of emulation setup. Is that a setup we can easily copy for Miri?

RalfJung:
I tried to do this with cross-rs but sadly ran into problems: https://github.com/cross-rs/cross/issues/1694

RalfJung:
I just learned from @bjorn3 that cross-building just doesn't work for us due to the rustc_private dependency.

So, ugh, is there anything we can do here? The status quo is that we edit the code in this repo and then on the next sync to rustc we learn whether it actually builds on riscv or ARM-Linux, which is pretty bad.

RalfJung:
https://github.com/rust-lang/miri/pull/4424 added an aarch64-linux runner, which closes this issue.

https://github.com/rust-lang/miri/pull/4422 will hopefully add a bunch more architectures. :)

marcoieni:
> rustc now seems to have aarch64 Linux runners for their CI, or some sort of emulation setup. Is that a setup we can easily copy for Miri?

I'm answering even though you figured this out already. There are linux arm runners in github now, such as `ubuntu-24.04-arm` that you added in https://github.com/rust-lang/miri/pull/4424

So no emulation setup required 👍

RalfJung:
We also have riscv and s390x now via qemu-user and binfmt, which is a totally seamless emulation setup :)

PR #4421, opened 2025-06-29
CI: do a check build on more targets

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/4419


RalfJung:
Turns out this is hopeless, rustc_private crates cannot be cross-built. :cry:

PR #4422, opened 2025-06-29
run CI on a bunch more architectures

RalfJung:
We only have 64bit machines for those so we have to do some cross-compilation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,13 +75,13 @@ jobs:
 
       # The main test job! We don't run this in qemu as that is quite slow,
       # so those targets only get the clippy check below.
-      - name: Test Miri
+      - name: test Miri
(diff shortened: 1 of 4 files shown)


RalfJung:
Using the proper cross-compilation approach for both x86 and ARM (instead of multilib) actually means this already works, we don't even have to wait for https://github.com/libffi-rs/libffi-rs/pull/160. :)

However this does uncover that the build is broken on arm32 in the ptrace decompilation stuff, so we can only land this once https://github.com/rust-lang/rust/pull/143137 syncs back into this repo.

RalfJung:
Wow, with qemu + binfmt we can even just install and run the s390x version of rustc on an x86 host. :D

RalfJung:
Aaand even more native-lib tracing build failures, since the "fallback" path also doesn't build apparently.

PR #4434, opened 2025-07-01
disable type validity assertions for better error messages

RalfJung:
Miri has pretty nice errors for type validity violations, printing which field in the type the problem occurs at and so on.

However, we don't see these errors when using e.g. `mem::zeroed` as that uses `assert_zero_valid` to bail out before Miri can detect the UB.

Similar to what we did with @saethlin's UB checks, I think we should disable such language UB checks in Miri so that we can get better error messages. If we go for this we should probably say this in the intrinsic docs as well so that people don't think they can rely on these intrinsics catching anything.

### DIFF
--- a/src/intrinsics/mod.rs
+++ b/src/intrinsics/mod.rs
@@ -30,6 +30,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         unwind: mir::UnwindAction,
     ) -> InterpResult<'tcx, Option<ty::Instance<'tcx>>> {
         let this = self.eval_context_mut();
+        let intrinsic_name = this.tcx.item_name(instance.def_id());
 
         // Force use of fallback body, if available.
         if this.machine.force_intrinsic_fallback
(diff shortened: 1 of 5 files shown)


saethlin:
Did you have a specific reason for putting the logic here instead of using `cfg`s in the standard library code?

RalfJung:
No, both seemed about equally valid to me and this approach keeps the concern local to Miri. It also means const-eval code still gets the assertion  even when running in Miri.

But I can make this cfg(miri) in libs if you prefer *shrug*

src/intrinsics/mod.rs
```diff
@@ -41,11 +42,23 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             }));
         }
 
+        // Overwrite some core engine intrinsics that are used to catch UB -- we have better
```
oli-obk, line 45: Shouldn't we remove them from the core engine and add them to the ctfe intrinsic handler if miri doesn't want them?

RalfJung: That's also an option, yeah. Maybe that's more clear.

So what should it be:
- do this in the library?
- or do it by neutering the intrinsic in Miri?

oli-obk: I like this variant as it keeps the checks in miri const eval as you said

saethlin: Oh, because we don't do validation in normal const eval? So if we removed these checks for all interpreters, we'd miss even more UB in const eval?

RalfJung: > Oh, because we don't do validation in normal const eval? So if we removed these checks for all interpreters, we'd miss even more UB in const eval?

Exactly.

RalfJung:
Closing in favor of https://github.com/rust-lang/rust/pull/143327.

PR #4435, opened 2025-07-02
native_lib/trace: fix and reenable

nia-e:
Per discussion in various PRs (https://github.com/rust-lang/miri/pull/4401, https://github.com/rust-lang/miri/pull/4405), this addresses (some of) the points raised. Also using this to check if CI passes on all architectures.

### DIFF
--- a/src/alloc/isolated_alloc.rs
+++ b/src/alloc/isolated_alloc.rs
@@ -302,23 +302,20 @@ impl IsolatedAlloc {
         }
     }
 
-    /// Returns a vector of page addresses managed by the allocator.
-    pub fn pages(&self) -> Vec<usize> {
-        let mut pages: Vec<usize> =
-            self.page_ptrs.iter().map(|p| p.expose_provenance().get()).collect();
-        for (ptr, size) in self.huge_ptrs.iter() {
-            for i in 0..size / self.page_size {
-                pages.push(ptr.expose_provenance().get().strict_add(i * self.page_size));
-            }
-        }
-        pages
+    /// Returns a list of page addresses managed by the allocator.
+    pub fn pages(&self) -> impl Iterator<Item = usize> {
+        let pages = self.page_ptrs.iter().map(|p| p.expose_provenance().get());
+        pages.chain(self.huge_ptrs.iter().flat_map(|(ptr, size)| {
+            (0..size / self.page_size)
+                .map(|i| ptr.expose_provenance().get().strict_add(i * self.page_size))
+        }))
     }
 
     /// Protects all owned memory as `PROT_NONE`, preventing accesses.
     ///
     /// SAFETY: Accessing memory after this point will result in a segfault
     /// unless it is first unprotected.
-    pub unsafe fn prepare_ffi(&mut self) -> Result<(), nix::errno::Errno> {
+    pub unsafe fn start_ffi(&mut self) -> Result<(), nix::errno::Errno> {
         let prot = mman::ProtFlags::PROT_NONE;
         unsafe { self.mprotect(prot) }
     }
@@ -326,7 +323,7 @@ impl IsolatedAlloc {
     /// Deprotects all owned memory by setting it to RW. Erroring here is very
     /// likely unrecoverable, so it may panic if applying those permissions
     /// fails.
-    pub fn unprep_ffi(&mut self) {
+    pub fn end_ffi(&mut self) {
         let prot = mman::ProtFlags::PROT_READ | mman::ProtFlags::PROT_WRITE;
         unsafe {
             self.mprotect(prot).unwrap();
--- a/src/bin/miri.rs
+++ b/src/bin/miri.rs
@@ -233,8 +233,6 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
         } else {
             let return_code = miri::eval_entry(tcx, entry_def_id, entry_type, &config, None)
                 .unwrap_or_else(|| {
-                    //#[cfg(target_os = "linux")]
-                    //miri::native_lib::register_retcode_sv(rustc_driver::EXIT_FAILURE);
                     tcx.dcx().abort_if_errors();
                     rustc_driver::EXIT_FAILURE
                 });
@@ -337,6 +335,9 @@ impl rustc_driver::Callbacks for MiriBeRustCompilerCalls {
 fn exit(exit_code: i32) -> ! {
     // Drop the tracing guard before exiting, so tracing calls are flushed correctly.
     deinit_loggers();
+    // Make sure the supervisor knows about the code code.
+    #[cfg(target_os = "linux")]
+    miri::native_lib::register_retcode_sv(exit_code);
     std::process::exit(exit_code);
 }
 
@@ -355,6 +356,11 @@ fn run_compiler_and_exit(
     args: &[String],
     callbacks: &mut (dyn rustc_driver::Callbacks + Send),
 ) -> ! {
+    // Install the ctrlc handler that sets `rustc_const_eval::CTRL_C_RECEIVED`, even if
+    // MIRI_BE_RUSTC is set. We do this late so that when `native_lib::init_sv` is called,
+    // there are no other threads.
+    rustc_driver::install_ctrlc_handler();
+
     // Invoke compiler, catch any unwinding panics and handle return code.
     let exit_code =
         rustc_driver::catch_with_exit_code(move || rustc_driver::run_compiler(args, callbacks));
@@ -438,10 +444,6 @@ fn main() {
 
     let args = rustc_driver::catch_fatal_errors(|| rustc_driver::args::raw_args(&early_dcx))
         .unwrap_or_else(|_| std::process::exit(rustc_driver::EXIT_FAILURE));
-
-    // Install the ctrlc handler that sets `rustc_const_eval::CTRL_C_RECEIVED`, even if
-    // MIRI_BE_RUSTC is set.
-    rustc_driver::install_ctrlc_handler();
 
     // If the environment asks us to actually be rustc, then do that.
     if let Some(crate_kind) = env::var_os("MIRI_BE_RUSTC") {
@@ -750,15 +752,15 @@ fn main() {
 
     debug!("rustc arguments: {:?}", rustc_args);
     debug!("crate arguments: {:?}", miri_config.args);
-    #[cfg(target_os = "linux")]
     if !miri_config.native_lib.is_empty() && miri_config.native_lib_enable_tracing {
-        // FIXME: This should display a diagnostic / warning on error
-        // SAFETY: If any other threads exist at this point (namely for the ctrlc
-        // handler), they will not interact with anything on the main rustc/Miri
-        // thread in an async-signal-unsafe way such as by accessing shared
-        // semaphores, etc.; the handler only calls `sleep()` and `exit()`, which
-        // are async-signal-safe, as is accessing atomics
-        //let _ = unsafe { miri::native_lib::init_sv() };
+        // SAFETY: No other threads are running
+        #[cfg(target_os = "linux")]
+        if unsafe { miri::native_lib::init_sv() }.is_err() {
+            eprintln!(
+                "warning: The native-lib tracer could not be started. Is this an x86 Linux system, and does Miri have permissions to ptrace?\n\
+                Falling back to non-tracing native-lib mode."
+            );
+        }
     }
     run_compiler_and_exit(
         &rustc_args,
--- a/src/shims/native_lib/mod.rs
+++ b/src/shims/native_lib/mod.rs
@@ -1,9 +1,5 @@
 //! Implements calling functions from a native library.
 
-// FIXME: disabled since it fails to build on many targets.
-//#[cfg(target_os = "linux")]
-//pub mod trace;
-
 use std::ops::Deref;
 
 use libffi::high::call as ffi;
@@ -13,14 +9,55 @@ use rustc_middle::mir::interpret::Pointer;
 use rustc_middle::ty::{self as ty, IntTy, UintTy};
 use rustc_span::Symbol;
 
-//#[cfg(target_os = "linux")]
-//use self::trace::Supervisor;
+#[cfg_attr(
+    not(all(
+        target_os = "linux",
+        target_env = "gnu",
+        any(target_arch = "x86", target_arch = "x86_64")
+    )),
+    path = "trace/stub.rs"
+)]
+pub mod trace;
+
 use crate::*;
 
-//#[cfg(target_os = "linux")]
-//type CallResult<'tcx> = InterpResult<'tcx, (ImmTy<'tcx>, Option<self::trace::messages::MemEvents>)>;
-//#[cfg(not(target_os = "linux"))]
-type CallResult<'tcx> = InterpResult<'tcx, (ImmTy<'tcx>, Option<!>)>;
+/// The final results of an FFI trace, containing every relevant event detected
+/// by the tracer.
+#[allow(dead_code)]
+#[cfg_attr(target_os = "linux", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug)]
+pub struct MemEvents {
+    /// An list of memory accesses that occurred, in the order they occurred in.
+    pub acc_events: Vec<AccessEvent>,
+}
+
+/// A single memory access.
+#[allow(dead_code)]
+#[cfg_attr(target_os = "linux", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug)]
+pub enum AccessEvent {
+    /// A read may have occurred on this memory range.
+    /// Some instructions *may* read memory without *always* doing that,
+    /// so this can be an over-approximation.
+    /// The range info, however, is reliable if the access did happen.
+    Read(AccessRange),
+    /// A read may have occurred on this memory range.
+    /// Some instructions *may* write memory without *always* doing that,
+    /// so this can be an over-approximation.
+    /// The range info, however, is reliable if the access did happen.
+    Write(AccessRange),
+}
+
+/// The memory touched by a given access.
+#[allow(dead_code)]
+#[cfg_attr(target_os = "linux", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Clone, Debug)]
+pub struct AccessRange {
+    /// The base address in memory where an access occurred.
+    pub addr: usize,
+    /// The number of bytes affected from the base.
+    pub size: usize,
+}
 
 impl<'tcx> EvalContextExtPriv<'tcx> for crate::MiriInterpCx<'tcx> {}
 trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
@@ -31,18 +68,17 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
         dest: &MPlaceTy<'tcx>,
         ptr: CodePtr,
         libffi_args: Vec<libffi::high::Arg<'a>>,
-    ) -> CallResult<'tcx> {
+    ) -> InterpResult<'tcx, (crate::ImmTy<'tcx>, Option<MemEvents>)> {
         let this = self.eval_context_mut();
-        //#[cfg(target_os = "linux")]
-        //let alloc = this.machine.allocator.as_ref().unwrap();
+        #[cfg(target_os = "linux")]
+        let alloc = this.machine.allocator.as_ref().unwrap();
+        #[cfg(not(target_os = "linux"))]
+        // Placeholder value.
+        let alloc = ();
 
-        // SAFETY: We don't touch the machine memory past this point.
-        //#[cfg(target_os = "linux")]
-        //let (guard, stack_ptr) = unsafe { Supervisor::start_ffi(alloc) };
-
-        // Call the function (`ptr`) with arguments `libffi_args`, and obtain the return value
-        // as the specified primitive integer type
-        let res = 'res: {
+        trace::Supervisor::do_ffi(alloc, || {
+            // Call the function (`ptr`) with arguments `libffi_args`, and obtain the return value
+            // as the specified primitive integer type
             let scalar = match dest.layout.ty.kind() {
                 // ints
                 ty::Int(IntTy::I8) => {
@@ -93,7 +129,7 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // have the output_type `Tuple([])`.
                 ty::Tuple(t_list) if (*t_list).deref().is_empty() => {
                     unsafe { ffi::call::<()>(ptr, libffi_args.as_slice()) };
-                    break 'res interp_ok(ImmTy::uninit(dest.layout));
+                    return interp_ok(ImmTy::uninit(dest.layout));
                 }
                 ty::RawPtr(..) => {
                     let x = unsafe { ffi::call::<*const ()>(ptr, libffi_args.as_slice()) };
@@ -101,23 +137,14 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     Scalar::from_pointer(ptr, this)
                 }
                 _ =>
-                    break 'res Err(err_unsup_format!(
+                    return Err(err_unsup_format!(
                         "unsupported return type for native call: {:?}",
                         link_name
                     ))
                     .into(),
             };
             interp_ok(ImmTy::from_scalar(scalar, dest.layout))
-        };
-
-        // SAFETY: We got the guard and stack pointer from start_ffi, and
-        // the allocator is the same
-        //#[cfg(target_os = "linux")]
-        //let events = unsafe { Supervisor::end_ffi(alloc, guard, stack_ptr) };
-        //#[cfg(not(target_os = "linux"))]
-        let events = None;
-
-        interp_ok((res?, events))
+        })
     }
 
     /// Get the pointer to the function of the specified name in the shared object file,
@@ -214,10 +241,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 if !this.machine.native_call_mem_warned.replace(true) {
                     // Newly set, so first time we get here.
                     this.emit_diagnostic(NonHaltingDiagnostic::NativeCallSharedMem {
-                        //#[cfg(target_os = "linux")]
-                        //tracing: self::trace::Supervisor::is_enabled(),
-                        //#[cfg(not(target_os = "linux"))]
-                        tracing: false,
+                        tracing: self::trace::Supervisor::is_enabled(),
                     });
                 }
 
--- a/src/shims/native_lib/trace/child.rs
+++ b/src/shims/native_lib/trace/child.rs
@@ -4,12 +4,22 @@ use std::rc::Rc;
 use ipc_channel::ipc;
 use nix::sys::{ptrace, signal};
 use nix::unistd;
+use rustc_const_eval::interpret::InterpResult;
 
 use super::CALLBACK_STACK_SIZE;
-use super::messages::{Confirmation, MemEvents, StartFfiInfo, TraceRequest};
+use super::messages::{Confirmation, StartFfiInfo, TraceRequest};
 use super::parent::{ChildListener, sv_loop};
 use crate::alloc::isolated_alloc::IsolatedAlloc;
+use crate::shims::native_lib::MemEvents;
 
+/// A handle to the single, shared supervisor process across all `MiriMachine`s.
+/// Since it would be very difficult to trace multiple FFI calls in parallel, we
+/// need to ensure that either (a) only one `MiriMachine` is performing an FFI call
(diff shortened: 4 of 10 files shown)


nia-e:
The only options I see for making the builds work nicely regardless of arch are either (a) copy-pasting a massive `cfg` macro everywhere; (b) copy-pasting said massive macro only a couple times to declare some stub functions; or (c) give in and add a small build script. Given the size of Miri I assume the choice to not have said build script was very deliberate, so I can try to do one of the other options (or maybe I'm missing something!) but for now this hopefully will build. [This](https://github.com/rust-lang/cargo/issues/1197) would have solved everything trivially, but oh well

RalfJung:
It should be entirely possible to do this with a single `cfg` that decides whether we include the "real" implementation of `native_lib::tracing`, or a fallback.

Cargo.toml
```diff
@@ -67,6 +67,7 @@ default = ["stack-cache"]
 genmc = []
 stack-cache = []
 stack-cache-consistency-check = ["stack-cache"]
+trace = []
```
RalfJung: Even if we did use the build script approach (which I don't think we should), this makes no sense. Cargo features and cfg flags are entirely different things.

Also note that "trace" is ambiguous, we recently landed https://github.com/rust-lang/rust/pull/140969 which also involves the term "trace" but is something very different.

src/bin/miri.rs
```diff
@@ -796,15 +796,15 @@ fn main() {
 
     debug!("rustc arguments: {:?}", rustc_args);
     debug!("crate arguments: {:?}", miri_config.args);
-    #[cfg(target_os = "linux")]
+    #[cfg(trace)]
     if !miri_config.native_lib.is_empty() && miri_config.native_lib_enable_tracing {
         // FIXME: This should display a diagnostic / warning on error
```
```suggestion
        // FIXME(ptrace): This should display a diagnostic / warning on error
```

src/bin/miri.rs
```diff
@@ -796,15 +796,15 @@ fn main() {
...
         // SAFETY: If any other threads exist at this point (namely for the ctrlc
         // handler), they will not interact with anything on the main rustc/Miri
         // thread in an async-signal-unsafe way such as by accessing shared
         // semaphores, etc.; the handler only calls `sleep()` and `exit()`, which
         // are async-signal-safe, as is accessing atomics
```
RalfJung: As discussed elsewhere, I don't think this comment makes sense. You seem to think that when you `fork`, there's some requirement that spans all currently running threads. That's not the case; only the thread that calls `fork` is being carried over to the new process, the other ones just disappear, so what they would have done does not matter.

What matters is that the code in the main thread that runs after `fork` is async-signal-safe. And, uh... is it? There's a ton of code still running there, like all the IPC and tracing logic.^^

src/shims/native_lib/mod.rs
```diff
@@ -1,8 +1,7 @@
 //! Implements calling functions from a native library.
 
-// FIXME: disabled since it fails to build on many targets.
-//#[cfg(target_os = "linux")]
-//pub mod trace;
+#[cfg(trace)]
+pub mod trace;
```
RalfJung: To reduce the amount of `cfg` in this file, I would suggest something like this:
```rust
#[cfg_attr(not(all(target_os = "linux", any(target_arch = "x86_64", ...), path = "trace/stub.rs")]
mod trace;
```
and then the `stub.rs` file contains just a bunch of empty functions and dummy type aliases so that to the outside, it looks like all the same functions and types exist.

src/alloc/mod.rs
```diff
@@ -1,5 +1,5 @@
 mod alloc_bytes;
-#[cfg(target_os = "linux")]
```
RalfJung: Let's just keep using that allocator on all Linux targets, no harm in that.

src/bin/miri.rs
```diff
@@ -227,8 +227,8 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
         } else {
             let return_code = miri::eval_entry(tcx, entry_def_id, entry_type, &config, None)
                 .unwrap_or_else(|| {
-                    //#[cfg(target_os = "linux")]
-                    //miri::native_lib::register_retcode_sv(rustc_driver::EXIT_FAILURE);
+                    #[cfg(trace)]
+                    miri::native_lib::register_retcode_sv(rustc_driver::EXIT_FAILURE);
```
RalfJung: At the very least, add a comment explaining why you are doing this here and nowhere else.

Does putting this into `fn exit` (which will exist after a rebase) suffice?

src/alloc/isolated_alloc.rs
RalfJung, line 1: `fn pages` can also still avoid an unnecessary `clone()...collect()`.

Cargo.toml
nia-e: The feature was left in by accident, sorry! But I refactored per your suggestion anyways to just cfg_attr a path with a stub implementation

src/bin/miri.rs
nia-e: The other threads disappear in the child process, sure. However, calling `fork` in a single-threaded process does not come with any extra requirements that the child act in a signal-safe way, but calling it in a multithreaded one does. ("After a fork() in a multithreaded program, the child can safely call only async-signal-safe functions (see signal-safety(7)) until such time as it calls execve(2).") As I understand it this isn't some inherent safety invariant of `fork` itself; from the signal-safety docs,

"To avoid problems with unsafe functions, there are two possible choices:

(a) Ensure that (1) the signal handler [or child process after a fork] calls only async-signal-safe functions, and (2) the signal handler [or child] itself is reentrant with respect to global variables in the main program.
(b) Block signal delivery in the main program when calling functions that are unsafe or operating on global data that is also accessed by the signal handler [or child].

Generally, the second choice is difficult in programs of any complexity, so the first choice is taken."

In this case we can ensure the second requirement because we know that the one other thread (the ctrlc handler) does not operate on global data that might be ever accessed by the child process or vice versa. That's what that comment tries to explain.

src/bin/miri.rs
nia-e: It's probably wise to put it there to be safe, but it should also be here above `abort_if_errors()` since that may exit the program without calling this new exit function. I'll add a comment, though

src/bin/miri.rs
RalfJung: Okay, I see. I still disagree with the claim that we "know" anything about the ctrlc handler though, we are just making an educated guess.

Is there any reason we have to do the forking *after* installing the ctrlc handler?

src/bin/miri.rs
RalfJung: > it should also be here above abort_if_errors() since that may exit the program without calling this new exit function.

No, it cannot exit the program. It'll unwind, and we'll end up [here](https://github.com/rust-lang/miri/blob/7035160b93cd78abbfb5fd50ac840ed73dfe4ca4/src/bin/miri.rs#L360).

src/bin/miri.rs
nia-e: We can install it after if that's preferred. Maybe put it in `run_compiler_and_exit`?

src/bin/miri.rs
nia-e: Checked, seems to work if we put it there. Ty!

src/bin/miri.rs
RalfJung: Definitely preferred to making assumptions about 3rd-party crates. :)

src/alloc/isolated_alloc.rs
RalfJung: Oh, the clone is already gone apparently, sorry for that... but that's a nice use of combinators, too :)

src/shims/native_lib/mod.rs
```diff
@@ -12,15 +9,27 @@ use rustc_middle::mir::interpret::Pointer;
...
 
-#[cfg(trace)]
+#[cfg_attr(
+    all(
+        target_os = "linux",
+        target_env = "gnu",
+        any(target_arch = "x86", target_arch = "x86_64", target_arch = "aarch64")
+    ),
+    path = "trace/mod.rs"
+)]
```
RalfJung: This attribute can just be removed as that's the default path it will use.

src/shims/native_lib/mod.rs
```diff
@@ -111,10 +118,8 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
         // SAFETY: We got the guard and stack pointer from start_ffi, and
         // the allocator is the same
-        #[cfg(trace)]
         let events = unsafe { Supervisor::end_ffi(alloc, guard, stack_ptr) };
-        #[cfg(not(trace))]
-        let events = None;
+        //let events = None;
```
RalfJung: Please remove commented-out code

src/shims/native_lib/trace/mod.rs
```diff
@@ -6,3 +6,8 @@ pub use self::child::{Supervisor, init_sv, register_retcode_sv};
 
 /// The size of the temporary stack we use for callbacks that the server executes in the client.
 const CALLBACK_STACK_SIZE: usize = 1024;
+
+pub(super) type CallResult<'tcx> = rustc_const_eval::interpret::InterpResult<
```
RalfJung: Please add a doc comment

src/shims/native_lib/mod.rs
```diff
@@ -31,14 +39,12 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    ) -> trace::CallResult<'tcx> {
         let this = self.eval_context_mut();
-        //#[cfg(target_os = "linux")]
-        //let alloc = this.machine.allocator.as_ref().unwrap();
+        let alloc = this.machine.allocator.as_ref().unwrap();
 
         // SAFETY: We don't touch the machine memory past this point.
-        //#[cfg(target_os = "linux")]
-        //let (guard, stack_ptr) = unsafe { Supervisor::start_ffi(alloc) };
+        let (guard, stack_ptr) = unsafe { Supervisor::start_ffi(alloc) };
```
RalfJung: Just to make sure this doesn't get lost, this should become just `let guard = ` with a new `FfiCallGuard` type or so that stores the `stack_ptr` internally.

Actually, I think it would be even better to make this a callback-based API, and thus fully exclude any chance of the user forgetting to call `end_ffi`:
```rust
unsafe fn do_ffi(
  alloc: &Rc<RefCell<IsolatedAlloc>>,
  f: impl FnOnce(),
) -> Option<MemEvents>;
```
This function does the setup to prepare an FFI call, then falls `f` which should do the call, then does the cleanup, and finally returns what happened while `f` was running.

The only annoying thing is this puts the entire closure inside the unsafe block, but the closure is already half-unsafe so that's not such a big deal.

RalfJung: Ah, we also need the closure to return something:
```rust
unsafe fn do_ffi<R>(
  alloc: &Rc<RefCell<IsolatedAlloc>>,
  f: impl FnOnce() -> InterpResult<'tcx, R>,
) -> InterpResult<'tcx, (R, Option<MemEvents>)>
```
That then also avoids the slightly awkward `break 'ret` stuff we currently use, we can just `return` from the closure.

i like this, let's go for this. :)

nia-e: I had just written something which I pushed before noticing this, but I'll switch to this ^^ ty

src/shims/native_lib/trace/child.rs
```diff
@@ -10,6 +10,11 @@ use super::messages::{Confirmation, MemEvents, StartFfiInfo, TraceRequest};
 use super::parent::{ChildListener, sv_loop};
 use crate::alloc::isolated_alloc::IsolatedAlloc;
 
+/// A handle to the single, shared supervisor process across all `MiriMachine`s.
+/// Since it would be very difficult to trace multiple FFI calls in parallel, we
+/// need to ensure that either (a) only one `MiriMachine` is performing an FFI call
+/// at any given time, or (b) there are distinct supervisor and child processes for
+/// each machine. The former was chosen here.
```
RalfJung, line 19: Please also explain why there is an Option here. Is it just for initialization? Or does the supervisor sometimes get taken out?

nia-e: Right now I think it's just for initialising it, yes. I'll mention that

src/shims/native_lib/trace/mod.rs
```diff
@@ -2,14 +2,34 @@ mod child;
...
+    f: impl FnOnce() -> InterpResult<'tcx, crate::ImmTy<'tcx>>,
+) -> CallResult<'tcx> {
+    // SAFETY: We don't touch the machine memory past this point.
+    let guard = unsafe { Supervisor::start_ffi(alloc) };
+
+    f().map(|v| {
+        let memevents = unsafe { Supervisor::end_ffi(guard) };
+        (v, memevents)
+    })
+}
```
RalfJung: Instead of adding another layer, I think this code becomes a lot clearer if you *replace* the existing start_ffi and end_ffi with the new function. No need to stack abstractions on top of each other here, that just increases the mental burden. That also keeps the setup and teardown code nicely close to each other within a single function, and you don't have to spell out the type of all the stuff that needs to be remembered in between.

src/shims/native_lib/trace/mod.rs
```diff
@@ -2,14 +2,34 @@ mod child;
...
 const CALLBACK_STACK_SIZE: usize = 1024;
 
-pub(super) type CallResult<'tcx> = rustc_const_eval::interpret::InterpResult<
-    'tcx,
-    (crate::ImmTy<'tcx>, Option<messages::MemEvents>),
->;
+/// Wrapper type for what we get back from an FFI call; the former is its actual
+/// return value, and the latter is the list of memory accesses that occurred during
+/// this call.
+pub type CallResult<'tcx> = InterpResult<'tcx, (crate::ImmTy<'tcx>, Option<messages::MemEvents>)>;
```
RalfJung: I think you can avoid this type alias if you move the `MemEvents` type outside `trace`, into `native_lib`. That also means the logic that processes the even is not infected by `cfg`.

src/shims/native_lib/trace/mod.rs
nia-e: Alright, I'll do that

nia-e:
Seems like CI is fine.

@rustbot ready

src/shims/native_lib/trace/child.rs
```diff
@@ -34,32 +44,23 @@ impl Supervisor {
...
-    ) -> (std::sync::MutexGuard<'static, Option<Supervisor>>, Option<*mut [u8; CALLBACK_STACK_SIZE]>)
-    {
+        f: impl FnOnce() -> InterpResult<'tcx, crate::ImmTy<'tcx>>,
+    ) -> InterpResult<'tcx, (crate::ImmTy<'tcx>, Option<MemEvents>)> {
         let mut sv_guard = SUPERVISOR.lock().unwrap();
-        // If the supervisor is not initialised for whatever reason, fast-fail.
+        // If the supervisor is not initialised for whatever reason, fast-return.
         // This might be desired behaviour, as even on platforms where ptracing
         // is not implemented it enables us to enforce that only one FFI call
         // happens at a time.
```
RalfJung: It's not very reassuring if the code says "might be desired behavior".^^
```suggestion
        // As a side-effect, even on platforms where ptracing
        // is not implemented, we enforce that only one FFI call
        // happens at a time.
```
It's kind of odd though that we would do this for "Linux without ptrace", but *not* do it on macos...

src/shims/native_lib/trace/parent.rs
```diff
@@ -5,24 +5,23 @@ use nix::sys::{ptrace, signal, wait};
...
-/// this cannot be const directly so we do it this way.
 const WAIT_FLAGS: wait::WaitPidFlag =
-    wait::WaitPidFlag::from_bits_truncate(libc::WUNTRACED | libc::WEXITED);
+    wait::WaitPidFlag::WUNTRACED.union(wait::WaitPidFlag::WEXITED);
 
 /// Arch-specific maximum size a single access might perform. x86 value is set
 /// assuming nothing bigger than AVX-512 is available.
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 const ARCH_MAX_ACCESS_SIZE: usize = 64;
 /// The largest arm64 simd instruction operates on 16 bytes.
```
RalfJung: Is there any citation or so you can give for this?
Seems pretty important that we don't get this wrong.^^

src/shims/native_lib/trace/parent.rs
RalfJung, line 1: Can you explain a bit more what happens in `capstone_disassemble`? There's a *loop* there that I did not expect...
```rust
for op in arch_detail.operands()
```
The code might actually end up adding *multiple events* to the list? But they are all *for the same address*? How does that make sense?

src/shims/native_lib/trace/child.rs
nia-e: Outside of many-seeds performance-sensitive scenarios, I find it hard to imagine when we *wouldn't* be okay with this blocking. I do recall running into a concrete case where this was desired; the `ptr_write_access` test will write to statics, so performing it in in many-seeds mode without the tracing caused it to randomly segfault (but it was fine with, as it forced the threads to take turns)

RalfJung: Yeah, I'm fine with the blocking.

But then the `do_ffi` implementation in `trace/stub.rs` should also do it to ensure feature parity. :)

src/shims/native_lib/trace/parent.rs
nia-e: I can add a comment explaining the rationale, but there isn't really a canonical set of flags to use for this that I know of. `WEXITED` makes sure we actually kill the child process when it exits instead of leaving it as a zombie, and WUNTRACED is mostly helpful at the very start before we've ptraced

src/shims/native_lib/trace/parent.rs
nia-e: There might be multiple operands, but we only care about the memory one. Thankfully, enforcing that there's only one memory operand per instruction is a rule for x86 (can look up a source if needed) and while I didn't check for ARM I highly doubt it's not

src/shims/native_lib/trace/parent.rs
RalfJung: Uh, I think you misunderstood my comment.^^ It refers to this:
> The largest arm64 simd instruction operates on 16 bytes.

nia-e: oh! oops. Yes, I'll add a citation

src/shims/native_lib/trace/parent.rs
RalfJung: But your code doesn't enforce this in any way. The code should be structurally unable to add more than one event. In fact I'd say `capstone_disassemble` should just return an `AccessEvent` rather than getting access to the full event list.

And then the surrounding code should probably be something like
```rust
let mut mem_operands = arch_detail.operands().filter(...);
let mem_operand = mem_operands.next(). /* error handling somehow */;
if let Some(snd_operand) = mem_operands.next() { /* uh? */ }
```
Or, maybe keep the loop but have some assertions there about not finding more than one operand. The details don't matter, what matters is that we should triple-check that everything makes sense and bail out on the slightest sign of something being off. Silently misbehaving in a 1% situation is *not* acceptable in Miri. People use Miri exactly to find those 1% bugs.

RalfJung: > and while I didn't check for ARM I highly doubt it's not

Every case where you are not 100% confident should be covered by runtime assertions and discussed in comments. We don't play guessing games in Miri. So when making your patches ready for upstream, please shift your mindset to ensuring 100% reliability -- it's better to panic in some odd cases than to *ever* silently do the wrong thing.

nia-e: That's fair. Turns out I was wrong, this raised a bit of doubt in my mind so I checked :< but I can fix it relatively easily, at least. Sorry ab this x.x

RalfJung: Given that we have a disassembler, is it practical to just *not need* such a fallback assumption? Can we just halt execution when capstone fails, and tell people to try again without tracing?

Having always-reliable size information would also make the post-processing logic simpler. So if that's practical, it seems like a good starting point?

nia-e: I'm slightly worried about weird edge cases here - say, someone running with weird extensions that just haven't yet made their way into the version of capstone we use. I'll try running some tests on really big libraries that do heavy C FFI on my raspi and see if at least the common cases are always correct

RalfJung: I think it's okay if for some cases people still have to use the non-tracing mode. Still a lot better than everyone having to use the non-tracing mode. :)

nia-e: Honestly for ARM we might be better off doing a lot of this manually. There's very few non-vector instructions which actually access memory (see https://developer.arm.com/documentation/dui0231/b/arm-instruction-reference/arm-memory-access-instructions/) and they have decently complex behaviour; we can just use capstone for the vector ones, and implement some manual logic based off of the reference in all other cases

RalfJung: Uh... I'm not excited at the prospect of maintaining code that parses machine code.^^ What's wrong with just using capstone?

nia-e: It doesn't provide the exact access size for non-x86 instructions currently. We can get decent approximations in some cases - like for NEON vector instructions, where we can tell the size based on the vector arrangement specifier - but generic non-simd loads and stores don't have any associated info. Plus, there's the somewhat big point that I ran into just now that sometimes loads and stores will proceed going backwards instead of forwards from the starting address which is... fun.

Honestly, I'm very tempted to just write a small[citation needed] crate for this because the actual logic judging from the docs is not that complex and I feel some amount of indignation that a proper tool does not already exist for this. In the meantime we can either keep the current behaviour which has nonexact sizes, or just Be Sad and turn off arm64 as well

RalfJung: I'm confused, how can capstone not have the information about how large the access is? Like, maybe it doesn't have that as a convenient getter, but surely we can just... match on the decoded instruction, and then know how big that is? That still beats decoding them ourselves.^^ It's like the difference between dealing with a raw text file vs a parsed AST. At least, that's what I am naively imagining, knowing nothing about aarch64 assembly nor capstone. ;)

But if we do that I think that's worth a separate PR. The first PR should land some protoype that makes everything as easy as possible. It sounds like that excludes aarch64, so I'd say turn that off *for now* and put it on the todo list.

nia-e: To clarify - I'm not suggesting directly parsing machine code. Just using the info capstone gives us a bit more creatively (e.g. matching on the mnemonic shorthand for a given instruction, with our custom logic living there)

nia-e: Ah, okay it seems like we had the same idea. Yeah, capstone doesn't have a simple getter like on x86 but we can still very much take advantage of a lot of the info it gives us

RalfJung: > To clarify - I'm not suggesting directly parsing machine code. Just using the info capstone gives us a bit more creatively (e.g. matching on the mnemonic shorthand for a given instruction, with our custom logic living there)

Right, yes that sounds much more sensible than what I thought you were suggesting. :)

src/bin/miri.rs
```diff
@@ -750,15 +751,14 @@ fn main() {
...
-        // handler), they will not interact with anything on the main rustc/Miri
-        // thread in an async-signal-unsafe way such as by accessing shared
-        // semaphores, etc.; the handler only calls `sleep()` and `exit()`, which
-        // are async-signal-safe, as is accessing atomics
-        //let _ = unsafe { miri::native_lib::init_sv() };
+        // SAFETY: No other threads are running
+        #[cfg(target_os = "linux")]
+        if unsafe { miri::native_lib::init_sv() }.is_err() {
+            eprintln!(
+                "warning: The native-lib tracer could not be started. Is this a Linux system, and does Miri have permissions to ptrace?"
```
```suggestion
                "warning: The native-lib tracer could not be started. Is this a Linux system, and does Miri have permissions to ptrace?\n\
                Falling back to non-tracing native-lib mode."
```

src/shims/native_lib/trace/stub.rs
```diff
@@ -12,13 +14,17 @@ impl Supervisor {
...
 }
 
 #[allow(dead_code, clippy::missing_safety_doc)]
-#[inline(always)]
 pub unsafe fn init_sv() -> Result<(), !> {
+    let mut sv_guard = SUPERVISOR.lock().unwrap();
+    if sv_guard.is_none() {
+        *sv_guard = Some(Supervisor);
+    }
     Ok(())
```
RalfJung: Shouldn't this return `Err`, to trigger the error that we are using fallback mode?

src/shims/native_lib/trace/stub.rs
```diff
@@ -0,0 +1,33 @@
...
+    #[inline(always)]
+    pub fn is_enabled() -> bool {
+        false
+    }
+
+    pub fn do_ffi<'tcx, T>(
+        _: T,
+        f: impl FnOnce() -> InterpResult<'tcx, crate::ImmTy<'tcx>>,
+    ) -> InterpResult<'tcx, (crate::ImmTy<'tcx>, Option<super::MemEvents>)> {
+        let _g = SUPERVISOR.lock().unwrap();
```
RalfJung, line 21:
```suggestion
        // We acquire the lock to ensure that no two FFI calls run concurrently.
        let _g = SUPERVISOR.lock().unwrap();
```

src/shims/native_lib/trace/stub.rs
```diff
@@ -0,0 +1,33 @@
...
+        f().map(|v| (v, None))
+    }
+}
+
+#[allow(dead_code, clippy::missing_safety_doc)]
+pub unsafe fn init_sv() -> Result<(), !> {
+    let mut sv_guard = SUPERVISOR.lock().unwrap();
+    if sv_guard.is_none() {
+        *sv_guard = Some(Supervisor);
+    }
```
RalfJung: There's not really any reason to have anything in that lock, is there?

nia-e: I suppose not, just added it for parity with the other implementation. I'm fine changing it

src/shims/native_lib/trace/stub.rs
nia-e: Ah, fair point

RalfJung:
@rustbot ready
(I know this is redundant but I need something in my inbox to represent this ;)

src/shims/native_lib/trace/parent.rs
```diff
@@ -417,71 +412,13 @@ fn handle_segfault(
                             if acc_ty.is_writable() {
```
RalfJung: Can it actually happen that the same operation is both a read and a write? Is that used for instance for atomic fetch-and-add and operations like that?

RalfJung (commented):
I have rebased and pushed a commit with some cleanups, please check if those make sense.

nia-e:
They all make sense, yeah. Figured I could leave the correct bits of the arm64 code as long as they're disabled, but I'll just add them all back at once in a later PR ^^

src/shims/native_lib/trace/parent.rs
nia-e: I don't know which ops exactly (I know on arm64 there's an explicit swap register with memory op for example), but yes there are x86 ops that do both. The write isn't always done on all ops; per `iced` which I was using before and is explicitly x86-focused so I'll trust as a source, the possible options are Read, CondRead, Write, CondWrite, ReadWrite, and ReadCondWrite. Since this only triggers when an access occurs, we can treat CondRead and CondWrite the same as their normal versions bc we know the condition was met, and putting a write after the read on a ReadCondWrite if the condition wasn't met is still fine in the sense that we don't lose information, just might accidentally mark uninit stuff as init.

We could maybe get around this with another `mempr_*` function so that instead we first set perms to read instead of RW, and if it retriggers a segfault we know for sure the write happens, but that might add a good bit of code complexity. It's an idea I've had for a bit though and plan on implementing at some point.

RalfJung:
> They all make sense, yeah. Figured I could leave the correct bits of the arm64 code as long as they're disabled, but I'll just add them all back at once in a later PR ^^

I'd rather not have code in the repo that doesn't even get built.

src/shims/native_lib/trace/parent.rs
RalfJung: Oh so these events are still approximate, we don't know if both a read and a write actually happened? Good to know, we have to keep that in mind when interpreting the events. That reduces the benefit of having precise range information...

nia-e: Well, that's only on the RCW events. I can add a boolean `is_certain` field on the write and if it's false we double-check that the allocation it's writing to is mutable first, but that will be a very rare edge case

RalfJung: It is UB to run `compare_exchange` on an immutable allocation even if it ends up not writing anything. So that particular check should be fine, if it only affects such atomic operations.

nia-e: I'll pour over the iced source later then, just to see which instructions that flag is set for. Ty!

PR #4443, opened 2025-07-04
declare data race and weak memory support as non-experimental

RalfJung:
I think data race detection and weak memory emulation are pretty solid now, and in particular they are (to our knowledge) sound wrt C++20. So we don't need to label them as experimental any more.


saethlin:
Agreed.

Issue #4379, opened 2025-06-05
Miri does not support uid, gid, and several other metadata fields


DianaNites:
`rustc 1.89.0-nightly (4b27a04cc 2025-06-04)`
`miri 0.1.0 (4b27a04cc8 2025-06-04)`

This program returns different, and incorrect, results when run in miri with isolation disabled vs without miri

```rust
use std::os::unix::fs::MetadataExt as _;

fn main() {
    let fs = std::fs::metadata("/proc/self/").unwrap();
    let fr_uid = fs.uid();
    let fr_gid = fs.gid();
    std::dbg!(fr_uid);
    std::dbg!(fr_gid);
}
```

This reproduces on the [playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=ff6ec386a93f141693121c95834d4b45), as well as my local machine. This applies to the directory itself and files within it

The playground returns user and group IDs of 1001, and my machine returns 1000, when run normally, and under miri both return 0/root. This is incorrect and does not match the host, and this can be verified by adding following snippet

```rust
std::dbg!(std::fs::read_link("/proc/self"));
loop {}
```

and then manually inspecting the correct `/proc` directory for the running miri test.

When isolation is disabled, miri reports the `metadata` call as unsupported with `error: unsupported operation: `statx` not available when isolation is enabled`

RalfJung:
Yeah we do not support uid/gid metadata as it is not exposed by std to my knowledge. So we have to fill in some dummy data.

RalfJung:
Seems like these days, std actually does expose uid/gid, so we could make use of that -- but only when we are on a Unix host. We typically try to avoid such host-specific behavior, but it might be better to support this on some hosts than on none.

This would have to be added around here:

https://github.com/rust-lang/miri/blob/02917f35e7e303b718062d95b62619c49cfb37b0/src/shims/unix/fs.rs#L745-L746

and here:

https://github.com/rust-lang/miri/blob/02917f35e7e303b718062d95b62619c49cfb37b0/src/shims/unix/fs.rs#L139-L140

PR #4444, opened 2025-07-05
shims::fs adding more fields to FileMetadata

devnexen:
addressing, partially at least, FIXME comment and
targetting unixes, adding device, user and group ids.

Fixes https://github.com/rust-lang/miri/issues/4379

### DIFF
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -128,19 +128,16 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let (created_sec, created_nsec) = metadata.created.unwrap_or((0, 0));
         let (modified_sec, modified_nsec) = metadata.modified.unwrap_or((0, 0));
         let mode = metadata.mode.to_uint(this.libc_ty_layout("mode_t").size)?;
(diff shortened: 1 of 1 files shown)


src/shims/unix/fs.rs
```diff
@@ -1601,6 +1610,32 @@ impl FileMetadata {
         let modified = extract_sec_and_nsec(metadata.modified())?;
 
         // FIXME: Provide more fields using platform specific methods.
-        interp_ok(Ok(FileMetadata { mode, size, created, accessed, modified }))
+
+        #[cfg(unix)]
```
RalfJung: Please rearrange this so that not *all* the fields need to be repeated everywhere. You can use `cfg_select!` to insert let-bindings into the surrounding scope like [here](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=76c91051796053974cba43aa5ec27204).

RalfJung:
Thanks! I left a comment for reducing code duplication. :)

@rustbot author

RalfJung:
Thanks! I did some more minor cleanup, that's easier than trying to explain it via comments. :) Please take a look, could be useful feedback for future PRs.

PR #4453, opened 2025-07-07
Implement building GenMC C++ code

Patrick-6:
This PR is in preparation for adding GenMC support to Miri.
It adds the `genmc-sys` crate, which can automatically load GenMC from a git repo (temporarily pointing to a private repo, but will point to the [GenMC Github repo](https://github.com/MPI-SWS/genmc) after everything is implemented)

This PR doesn't add any functionality for GenMC mode, only a test for calling C++ functions from Rust code through the `cxx` crate.

### DIFF
--- a/doc/genmc.md
+++ b/doc/genmc.md
@@ -60,6 +60,7 @@ The build script in the `genmc-sys` crate handles locating, downloading, buildin
 
 To determine which GenMC repo path will be used, the following steps are taken:
 - If the env var `GENMC_SRC_PATH` is set, it's value is used as a path to a directory with a GenMC repo (e.g., `GENMC_SRC_PATH="path/to/miri/genmc-sys/genmc-sys-local"`).
+  - Note that this variable must be set wherever Miri is built, e.g., in the terminal, or in the Rust Analyzer settings.
 - If the path `genmc-sys/genmc-src/genmc` exists, try to set the GenMC repo there to the commit we need.
 - If the downloaded repo doesn't exist or is missing the commit, the build script will fetch the commit over the network.
   - Note that the build script will *not* access the network if any of the steps previous steps succeeds.


Cargo.toml
```diff
@@ -28,6 +28,7 @@ chrono-tz = "0.10"
 directories = "6"
 bitflags = "2.6"
 serde_json = { version = "1.0", optional = true }
+genmc-sys = { path = "./genmc-sys/", version = "0.11.0", optional = true }
```
RalfJung: Does GenMC even compile on all architectures we support in Miri? I thought there was something about it not supporting 32bit hosts? What about Windows and macOS?

genmc-sys/build.rs
```diff
@@ -0,0 +1,184 @@
...
+/// Name of the library of the GenMC model checker.
+const GENMC_MODEL_CHECKER: &str = "model_checker";
+
+const RUST_CXX_BRIDGE_FILE_PATH: &str = "src/lib.rs";
+
+// FIXME(GenMC, build): decide whether to keep debug enabled or not (without this: calling BUG() ==> UB)
+const ENABLE_GENMC_DEBUG: bool = true;
+
+#[cfg(feature = "vendor_genmc")]
+mod vendoring {
```
RalfJung: I think "vendoring" is the wrong term here, as that refers to the practice of wholly copying a dependency's source code into our source code. I'd call it something involving `download`.

Cargo.toml
Patrick-6: GenMC seems to support at least [Linux and Mac OS X](https://github.com/MPI-SWS/genmc?tab=readme-ov-file#gnulinux) (likely only 64 bit), all other configurations are untested and might not work.

Also, even if it compiles on a platform, there could be issues with differently sized integers/endianness changing behavior (especially `uintptr_t` on a 32-bit OS).

I'll keep all platforms active for now, that way I see what fails where, but we should probably limit the `genmc` feature to just supported platforms before this is merged.

RalfJung: We don't run any tests on big-endian hosts so if that is not supported upstream we should disable it, even if we don't see test failures.

Patrick-6: GenMC is also not tested on any big endian hosts, so disabling those is probably a good idea.

I don't think there should be any endianness issues as long as the host and the target have the same endianness (GenMC always uses the host's endianness, it cannot be switched currently).
We will have to enforce equal endianness when we start merging the actual Miri-GenMC functionality.

genmc-sys/build.rs
```diff
@@ -0,0 +1,183 @@
...
+    // sizes and field offsets in the cxx bridge library compared to the model_checker library.
+    // This will lead to data corruption in these fields, which can be hard to debug (fields change values randomly).
+    if ENABLE_GENMC_DEBUG {
+        bridge.define("ENABLE_GENMC_DEBUG", "1");
+    }
+
+    bridge.compile("genmc_interop");
+
+    // Link the Rust-C++ interface library generated by cxx_build:
+    println!("cargo::rustc-link-lib=static=genmc_interop");
```
Patrick-6: Does Miri need to be compatible with older cargo versions (before 1.77)? If yes, we should use the old syntax `cargo:` (single colon) instead of `cargo::` as mentioned [here](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script).

Cargo.toml
RalfJung: That, and equal pointer size as well IIUC.

genmc-sys/build.rs
RalfJung: No, Miri only needs to build with nightly toolchains.

But please be consistent in whether you use one or two colons. :)

genmc-sys/build.rs
```diff
@@ -0,0 +1,183 @@
...
+
+const GENMC_LOCAL_PATH_STR: &str = "./genmc/";
+
+/// Name of the library of the GenMC model checker.
+const GENMC_MODEL_CHECKER: &str = "model_checker";
+
+const RUST_CXX_BRIDGE_FILE_PATH: &str = "src/lib.rs";
+
+// FIXME(GenMC, build): decide whether to keep debug enabled or not (without this: calling BUG() ==> UB)
+const ENABLE_GENMC_DEBUG: bool = true;
```
Patrick-6: GenMC has a `BUG()` statement that prints an error when reached with this setting enabled, and acts like `unreachable_unchecked` otherwise. There's also a `BUG_ON(cond)` that only checks its condition with this setting enabled.

So far I was unable to measure any meaningful performance impact from this setting for Miri-GenMC, and for GenMC the difference seems to be <5%.

I think we can leave this enabled for now, since it might help catch some bugs.

Cargo.toml
Patrick-6: Right, pointers must be 64-bit currently, because that's the only size the current memory allocation scheme supports.

genmc-sys/build.rs
RalfJung: Fine for me for now; before shipping enabling this in the Miri we distribute, we should also consult Michalis.

Cargo.toml
RalfJung: Oh, even the *target* pointers must be 64bit? That is unfortunate, but we have to start somewhere. :)

Patrick-6: It's sadly not a simple fix, the allocation scheme will need to be reworked for 32-bit systems to be useful. It currently uses 4 reserved bits, then 28 bits to indicate the thread the memory was allocated on, then 32 bits for the address (4 GiB per thread).

If we used the same scheme on a 32-bit system this tradeoff would be a lot worse, e.g., 4 bits reserved, 6 bits for threads (64 threads max), leaving only 22 bits for the address (4 MiB per thread, possibly not enough to be very useful).

We should focus on 64-bit for now and leave 32-bit as future work. :)

RalfJung: Yeah agreed, we just need a good error message. :)

(FWIW, that sounds like 64bit targets on a 32bit host should still work, if GenMC used less host-dependent types. But I assume very few people run Miri on 32bit hosts so it's very low priority.)

Patrick-6: I've looked into how we might have the `genmc` feature only available on supported targets, but looking at [cargo #1197](https://github.com/rust-lang/cargo/issues/1197), I don't think this is currently possible.
We can have the `genmc-sys` crate only linked on supported platforms, but this doesn't work for features:
```toml
[target.'cfg(all(any(target_os = "linux", target_os = "macos"), target_pointer_width = "64"))'.dependencies]
genmc-sys = { path = "./genmc-sys/", version = "0.11.0", optional = true }
```

We could add extra `cfg`s to still use the dummy module even with the `genmc` feature enabled, if the platform is unsupported ([here](https://github.com/Patrick-6/miri/blob/04102b2875dca9a6b8432f6b6dbd2b2439abb1a1/src/concurrency/mod.rs#L12)). (also, disable the GenMC tests on those platforms ([here](https://github.com/Patrick-6/miri/blob/04102b2875dca9a6b8432f6b6dbd2b2439abb1a1/tests/ui.rs#L340-L343)))

RalfJung: Yeah that's by design, features are global.

So what you suggest for the dependency sounds right, and then the logic inside Miri should plug in `genmc/dummy.rs` for other targets even if the feature is enabled.

genmc-sys/build.rs
```diff
@@ -0,0 +1,183 @@
...
+    // FIXME(GenMC, build): can we use c++23? Does CXX support that? Does rustc CI support that?
+    bridge
+        .opt_level(2)
+        .debug(true) // Same settings that GenMC uses ("-O2 -g")
+        .warnings(false) // NOTE: enabling this produces a lot of warnings.
+        .std("c++20")
+        .include(genmc_common_include_path)
+        .include(model_checker_include_path)
+        .include("./src_cpp")
+        .define("PACKAGE_BUGREPORT", "\"FIXME(GenMC) determine what to do with this!!\"") // FIXME(GenMC): HACK to get stuff to compile (this is normally defined by cmake)
```
Patrick-6: We don't *need* this define if we have `GENMC_DEBUG` enabled permanently, because the only place that uses this define (printing a bug report message) is replaced with a `BUG()` statement in that case.

Still, we should find a solution for keeping this (and other) settings in sync between the cxx_bridge and the cmake compilation steps in the build script.

Some compiler settings (e.g., `GENMC_DEBUG`) can even change the sizes of structs used by both compilation steps, which can lead to very weird bugs.

I haven't found a good solution yet for preventing such issues, since any added/changed settings in the cmake build will not automatically be picked up for the cxx_bridge build (unless we can somehow read those out, but that sounds like a bad idea).

RalfJung: I don't understand the question... Does genmc have *mandatory* settings in its cmake configuration? Most programs can be built without setting anything and pick reasonable defaults, so there's no reason Miri needs to know a list of all the settings.

Obviously if cmake settings can alter the ABI then we need to set them the right way and ensure they are fixed that way.

Patrick-6: We have some global compile definitions for the cmake build [here](https://github.com/Patrick-6/genmc/blob/bd4496acaf0994b2515f58f75d21ad8dd15f6603/CMakeLists.txt#L10-L13), without which certain files won't compile at all. The build script uses cmake to build the GenMC `model_checker` library, everything works there.

The issue comes from the second build with `cxx_bridge`. This step uses the files in the [`genmc-sys/src_cpp` directory](https://github.com/Patrick-6/miri/tree/miri-genmc-build/genmc-sys/src_cpp) to build the Rust-C++ interop library. Those C++ files include some header files from GenMC, which may not compile without those defines. Since this step is completely independent of the cmake build, we have to somehow set these defines ourselves.

RalfJung: Why can't this reuse the cmake build? Is this the officially recommended way to use cxx?

Patrick-6: That should possible, but then we need to have a way to add an extra CMakeLists.txt file to the GenMC build (which should be doable with an extra cmake parameter), and we will need to have the [cxxbridge-cmd](https://github.com/dtolnay/cxx?tab=readme-ov-file#non-cargo-setup) executable installed (in CI and locally for development) to be able to build the interop code from cmake.
We could run `cargo install cxxbridge-cmd` in the build script, but I'm not sure if that is the proper way to do it.

RalfJung: Are you saying that cxxbridge-cmd has features that are not available via the Rust API?

cxxbridge-cmd is explicitly indicated as being for "non-Cargo builds". I asked about what the official recommendation for this situation is, and cxxbridge-cmd is definitely not it. :)

Patrick-6: Well, the [official solution for a using cxx with cmake](https://cxx.rs/build/cmake.html) is 

> There is not an officially endorsed CMake setup for CXX, but a few developers have shared one that they got working. You can try one of these as a starting point.

The first example uses a separate cmake build and a `cxx_build::bridge` in the [build.rs](https://github.com/XiangpengHao/cxx-cmake-example/blob/master/rust_part/build.rs), which is what we have now.

Patrick-6: [The second example uses `cxxbridge-cmd`](https://github.com/david-cattermole/cxx-demo-example/blob/main/src/CMakeLists.txt#L16)

genmc-sys/build.rs
```diff
@@ -0,0 +1,183 @@
...
+    use std::path::PathBuf;
+    use std::str::FromStr;
+
+    use git2::{Oid, Repository};
+
+    use super::GENMC_LOCAL_PATH_STR;
+
+    pub(crate) const GENMC_GITHUB_URL: &str = "https://github.com/Patrick-6/genmc.git";
+    pub(crate) const GENMC_COMMIT: &str = "bd4496acaf0994b2515f58f75d21ad8dd15f6603";
+    pub(crate) const GENMC_DOWNLOAD_PATH_STR: &str = "./downloaded/genmc/";
```
Patrick-6: My idea was that this downloaded version is for building Miri-GenMC automatically (shouldn't be modified manually), and for development you can clone the repo into the `./genmc` path.

Should this `downloaded` path be inside the `target` directory instead of here?

- If we place it here, it won't be automatically cleaned up by `cargo clean`, but it also shouldn't change much, so maybe we don't need to.
- In `target`, it will also be re-downloaded for every configuration (Debug, Release), which isn't necessary.
- If someone wants to look at the downloaded code (or set up their IDE to scan it), placing it here would make that easier, compared to some subdirectory in `target` with a randomized path name.

RalfJung: > Should this downloaded path be inside the target directory instead of here?

Where exactly is "here"?

Patrick-6: `miri/genmc-sys/downloaded/genmc`

RalfJung: :shrug: I don't have a strong opinion, but if you put it inside `genmc-sys` then it should be listed in `.gitignore`.

Patrick-6: I think I'll keep it in `genmc-sys`.

Should this directory be included in the `rerun-if-changed` for the build script?
It currently is, and it causes an unnecessary rebuild on the second `./miri build` after downloading the files (probably because the download is seens as a modification). This directory isn't meant for development anyway (i.e., shouldn't be modified manually), so checking for changes is probably not needed.

RalfJung: If the user brings their own genmc rather than having it auto-downloaded, changing that should cause rebuilds.

For the auto-downloaded one that does not seem necessary. Though it'd be unfortunate if the user got confused about that...
Have you tried putting it in `CARGO_TARGET_TMPDIR`? Or does that also change between debug and release builds?

genmc-sys/build.rs
Patrick-6: Which option should I go with?

I think in terms of maintainability, build times, etc., going with just 1 compilation that's hooking into the existing cmake setup is easier, but requiring an install of `cxxbridge` is annoying.

We _could_ add `cargo install cxxbridge-cmd` to `./miri toolchain`?

genmc-sys/src_cpp/MiriInterface.hpp
```diff
@@ -0,0 +1,42 @@
...
+		std::cerr << "C++: GenMC handle created!" << std::endl;
+	}
+
+	virtual ~MiriGenMCShim()
+	{
+		std::cerr << "C++: GenMC handle destroyed!" << std::endl;
+	}
+
+	static std::unique_ptr<MiriGenMCShim> createHandle(const GenmcParams &config);
+};
```
Patrick-6: Since clippy and rustfmt don't apply to the C++ files, how should they be handled?
- Naming style? Like in Rust? (currently, the naming mostly follows GenMC's naming conventions)
- Formatting? `clang-format`?
- Linting? `clang-tidy`?

Should these be hooked into `./miri fmt` and `./miri clippy`? Maybe just if the `clang-*` tools are installed?

genmc-sys/build.rs
Patrick-6: I'm not sure that is the correct path to use, [according to the docs](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates):

> CARGO_TARGET_TMPDIR — Only set when building [integration test](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#integration-tests) or benchmark code. This is a path to a directory inside the target directory where integration tests or benchmarks are free to put any data needed by the tests/benches.

We already need the GenMC files for `./miri build`.

Patrick-6: > For the auto-downloaded one that does not seem necessary. Though it'd be unfortunate if the user got confused about that...

I'll add a check in the downloading code to give an error if the downloaded repo is not in a clean state, and note that changes should only be made in `genmc-sys/genmc/`.

I could also create a  `downloaded/README.md` file with the same info.

genmc-sys/build.rs
RalfJung: Let's not use cxxbridge-cmd. Surely there's some way we can share the logic for which definitions to add?

I don't think I fully understand the problem; let's discuss this at the next meeting.

genmc-sys/src_cpp/MiriInterface.hpp
RalfJung: > Should these be hooked into ./miri fmt and ./miri clippy? 

Not for the MVP. Maybe in a later PR we can add this -- but only if the tools are installed, which we should not require.

Regarding the naming style, no preference from my side. Going with the GenMC style works for me.

genmc-sys/build.rs
RalfJung: Oh right CARGO_TARGET_TMPDIR is wrong.

The actual cmake build files are put somewhere in the target dir, I assume? This is just about the source code?

Patrick-6: Yes, this is just about the GenMC source code.

The cmake and cxx build files are in `target/debug/genmc-sys-[HASH]/out/`. This path is set by cargo as the environment variable [`OUT_DIR`](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates):

> OUT_DIR — If the package has a build script, this is set to the folder where the build script should place its output. See below for more information. (Only set during compilation.)

RalfJung:
@rustbot author
Please mark it as ready when CI is green.

genmc-sys/build.rs
```diff
@@ -0,0 +1,200 @@
...
+            match repo.find_remote("origin") {
+                Ok(mut remote) =>
+                    match remote.fetch(&[GENMC_COMMIT], None, None) {
+                        Ok(_) =>
+                            println!(
+                                "cargo::warning=Successfully fetched commit '{GENMC_COMMIT:?}'"
+                            ),
+                        Err(e) => panic!("Failed to fetch from remote: {e}"),
+                    },
+                Err(e) => println!("cargo::warning=Could not find remote 'origin': {e}"),
```
Patrick-6: Which (if any) messages should be printed from the build script?
It might be useful to know that the build script is using the local `./genmc` repo vs cloning from git.
Sadly there does not seem to be a way to emit info messages from a build script, only warnings.

RalfJung: If everything works, the build script should be silent. Cargo will hide the output so `println!` is fine -- it will be shown if the build script fails, which is useful for debugging.

Cargo.toml
```diff
@@ -47,6 +47,10 @@ ipc-channel = "0.19.0"
 serde = { version = "1.0.219", features = ["derive"] }
 capstone = "0.13"
 
+# FIXME(genmc,macos): Add `target_os = "macos"` once `https://github.com/dtolnay/cxx/issues/1535` is fixed.
```
RalfJung:
```suggestion
# FIXME(genmc,macos): Add `target_os = "macos"` once https://github.com/dtolnay/cxx/issues/1535 is fixed.
```

genmc-sys/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "genmc-sys"
+version = "0.11.0" # NOTE: should match the GenMC version
```
RalfJung: This seems fragile. Can we avoid it?

Is there a specific file in the genmc repo that contains the genmc version?

genmc-sys/build.rs
```diff
@@ -0,0 +1,200 @@
...
+            std::process::exit(1);
+        }
+
+        // Check if there are any updates:
+        let commit = if let Ok(oid) = Oid::from_str(GENMC_COMMIT)
+            && let Ok(commit) = repo.find_commit(oid)
+        {
+            commit
+        } else {
+            match repo.find_remote("origin") {
```
RalfJung:
```suggestion
            // The commit is not in the checkout. Try `git fetch` and hope that we find the commit then.
            match repo.find_remote("origin") {
```

genmc-sys/build.rs
RalfJung: So please remove all of the `cargo::warning=Successfully ...` -- success is not worth a warning.

genmc-sys/build.rs
```diff
@@ -0,0 +1,200 @@
...
+                        Err(e) => panic!("Failed to fetch from remote: {e}"),
+                    },
+                Err(e) => println!("cargo::warning=Could not find remote 'origin': {e}"),
+            }
+            let oid = Oid::from_str(GENMC_COMMIT).unwrap();
+            repo.find_commit(oid).unwrap()
+        };
+
+        // Set the repo to the correct branch:
+        checkout_commit(&repo, GENMC_COMMIT);
```
RalfJung: You have the `commit` here already, can't you save some work in `checkout_commit` by passing it there?

genmc-sys/build.rs
```diff
@@ -0,0 +1,200 @@
...
+    fn checkout_commit(repo: &Repository, refname: &str) {
+        let (object, reference) = repo.revparse_ext(refname).expect("Object not found");
+
+        repo.checkout_tree(&object, None).expect("Failed to checkout");
+
+        match reference {
+            // `gref` is an actual reference like branches or tags.
+            Some(gref) => repo.set_head(gref.name().unwrap()),
+            // This is a commit, not a reference.
+            None => repo.set_head_detached(object.id()),
```
RalfJung: On the terminal I could just do `git checkout $commit` or `git reset --hard $commit`. Why does this need a match? Is there no command that always works?

genmc-sys/build.rs
```diff
@@ -0,0 +1,200 @@
...
+        .file("./src_cpp/MiriInterface.hpp")
+        .file("./src_cpp/MiriInterface.cpp");
+
+    // NOTE: It is very important to ensure that this and similar flags are set/unset both here and
+    // for the cmake build below, otherwise, certain structs/classes can have different
+    // sizes and field offsets in the cxx bridge library compared to the model_checker library.
+    // This will lead to data corruption in these fields, which can be hard to debug (fields change values randomly).
+    if ENABLE_GENMC_DEBUG {
+        bridge.define("ENABLE_GENMC_DEBUG", "1");
+    }
```
RalfJung: And because it is so similar, they should have shared logic -- one central table of flags, applied in two places.

Why is the flag called ENABLE_GENMC_DEBUG here and GENMC_DEBUG below?

genmc-sys/build.rs
```diff
@@ -0,0 +1,200 @@
...
+    let dst = config.build();
+
+    println!("cargo::rustc-link-search=native={}/build/{GENMC_MODEL_CHECKER}", dst.display());
+    println!("cargo::rustc-link-lib=static={GENMC_MODEL_CHECKER}");
+}
+
+fn main() {
+    // Select between local GenMC repo, or downloading GenMC from a specific commit.
+    let Ok(genmc_local_path) = PathBuf::from_str(GENMC_LOCAL_PATH_STR);
+    let genmc_path = if genmc_local_path.exists() {
```
RalfJung: So the local path, if it exists always takes precedence?

This should be documented somewhere. I'd say it's finally time to create a `doc` folder, and then you can have a `genmc.md` file inside it to explain things like this.

genmc-sys/build.rs
```diff
@@ -0,0 +1,200 @@
+use std::path::{Path, PathBuf};
+use std::str::FromStr;
+
+const GENMC_LOCAL_PATH_STR: &str = "./genmc/";
```
RalfJung:
```suggestion
const GENMC_LOCAL_PATH: &str = "./genmc/";
```
Please also add a doc comment.

src/bin/miri.rs
```diff
@@ -598,7 +598,11 @@ fn main() {
         } else if arg == "-Zmiri-many-seeds-keep-going" {
             many_seeds_keep_going = true;
         } else if let Some(trimmed_arg) = arg.strip_prefix("-Zmiri-genmc") {
-            // FIXME(GenMC): Currently, GenMC mode is incompatible with aliasing model checking.
+            if !miri_config.genmc_mode {
+                miri_config.genmc_mode = true;
+                // FIXME(GenMC): Currently, GenMC mode is incompatible with aliasing model checking.
+                miri_config.borrow_tracker = None;
+            }
             miri_config.borrow_tracker = None;
```
RalfJung: What is going on here, why do you set borrow_tracker twice? And why is there an `if`?

src/bin/miri.rs
```diff
@@ -598,7 +598,11 @@ fn main() {
...
             GenmcConfig::parse_arg(&mut genmc_config, trimmed_arg);
```
RalfJung: So we have `genmc_mode` and `genmc_config`... isn't that redundant? `genmc_mode` is always the same as `genmc_config.is_some()`?

src/concurrency/genmc/dummy.rs
```diff
@@ -228,9 +228,19 @@ impl VisitProvenance for GenmcCtx {
...
-        unimplemented!(
-            "GenMC feature im Miri is disabled, cannot handle argument: \"-Zmiri-genmc{trimmed_arg}\""
-        );
+        if cfg!(all(
+            feature = "genmc",
+            any(target_os = "linux", target_os = "macos"),
+            target_pointer_width = "64"
+        )) {
+            unimplemented!(
+                "GenMC feature im Miri is disabled, cannot handle argument: \"-Zmiri-genmc{trimmed_arg}\""
```
RalfJung:
```suggestion
                "GenMC is disabled, cannot handle argument: \"-Zmiri-genmc{trimmed_arg}\""
```

src/concurrency/mod.rs
```diff
@@ -9,7 +9,12 @@ mod vector_clock;
 pub mod weak_memory;
 
 // Import either the real genmc adapter or a dummy module.
-#[cfg_attr(not(feature = "genmc"), path = "genmc/dummy.rs")]
+// On unsupported platforms, we still include the dummy module, even if the `genmc` feature is enabled.
```
RalfJung:
```suggestion
// On unsupported platforms, we always include the dummy module, even if the `genmc` feature is enabled.
```

Cargo.toml
```diff
@@ -47,6 +47,10 @@ ipc-channel = "0.19.0"
...
+[target.'cfg(all(target_os = "linux", target_pointer_width = "64"))'.dependencies]
```
RalfJung: You said genmc assume little-endian, right? Please also add that to the conditions everywhere, then.

tests/ui.rs
```diff
@@ -338,6 +338,13 @@ fn main() -> Result<()> {
         ui(Mode::Fail, "tests/native-lib/fail", &target, WithoutDependencies, tmpdir.path())?;
     }
 
+    // We only enable GenMC tests when the `genmc` feature is enabled, but also only on platforms we support:
+    // FIXME(genmc,macos): Add `target_os = "macos"` once `https://github.com/dtolnay/cxx/issues/1535` is fixed.
+    if cfg!(all(feature = "genmc", target_os = "linux", target_pointer_width = "64")) {
+        ui(Mode::Pass, "tests/genmc/pass", &target, WithDependencies, tmpdir.path())?;
+        ui(Mode::Fail, "tests/genmc/fail", &target, WithDependencies, tmpdir.path())?;
+    }
```
RalfJung: Remember that `cfg!` checks host conditions, even when it says `target_*`.
So this runs only on Linux hosts, but it runs on all targets, e.g. if I do `--target i686-unknown-linux-gnu`.

(a) you need logic somewhere that rejects unsupported targets when genmc is enabled -- must be 64bit little-endian
(b) here you need to avoid running tests on the wrong targets... the easiest thing is to check `host == target`. That's not great as then we don't do any cross-target checking, but that's fine for now.

RalfJung:
Here's my first round of comments. :)

@rustbot author

genmc-sys/Cargo.toml
Patrick-6: This doesn't have any effects, we can keep the version at 0.1.0 like Miri's.

The version is defined in the [main `CMakeLists.txt` file](https://github.com/Patrick-6/genmc/blob/bd4496acaf0994b2515f58f75d21ad8dd15f6603/CMakeLists.txt#L2).

genmc-sys/build.rs
Patrick-6: As discussed, I'll swap over to using a common `config.h` file for both compilation steps.

`GENMC_DEBUG` is the name used for cmake (i.e., `cmake -D GENMC_DEBUG=ON`), while `ENABLE_GENMC_DEBUG` is for the C++ code (i.e., `g++ -D ENABLE_GENMC_DEBUG`)

genmc-sys/build.rs
Patrick-6: `miri/doc` or `miri/genmc-sys/doc` ?

Cargo.toml
Patrick-6: GenMC main assumption is that host and target have the same endianness, which we will need to check for once we add the actual functionality.
I don't think GenMC specifically assumes little-endian anywhere, but it is not tested on any big-endian systems.

This will likely only get relevant once support for mixed-size accesses is added to GenMC.
Should I still disable big-endian systems for now?

genmc-sys/Cargo.toml
RalfJung: >  we can keep the version at 0.1.0 like Miri's.

Then let's do that.

genmc-sys/build.rs
RalfJung: `miri/doc`. You said we'd probably want a bunch of docs specifically for genmc mode anyway, right?

Cargo.toml
RalfJung: > GenMC main assumption is that host and target have the same endianness, which we will need to check for once we add the actual functionality.

Ah okay. Then please add a FIXME in `GenmcCtx::new` indicating we need to check for that.

Where does it even assume that? If it just treats program values as opaque blobs, it shouldn't care whether host and target have the same endianess.

src/bin/miri.rs
Patrick-6: Yeah, I'll remove the `.genmc_mode` and replace it with a `.genmc_config`.

src/bin/miri.rs
Patrick-6: The duplicate is a mistake. The if is there in case someone does `-Zmiri-genmc -Zmiri-tree-borrows -Zmiri-genmc-something` or similar. Without the if, this would silently disable the `-Zmiri-tree-borrows` setting. With the if, there will instead be an error in the check further down in this file. (Note that depending on the order of arguments this can still happen, e.g., `-Zmiri-tree-borrows -Zmiri-genmc`).

Since the goal is to add borrow checking support in the future anyway, should we just go without the if?

RalfJung: I think we should stop mutating `borrow_tracker` here -- for the reasons you state, that becomes messy.

Instead, after all arguments are parsed, we can have a check whether *both* genmc and borrow tracking are enabled. Then we emit a warning saying that genmc is incompatible with borrow tracking and therefore borrow tracking is being disabled.

genmc-sys/build.rs
Patrick-6: Yes, there will be quite a lot of things to describe, such as usage, limitations, tricks/workarounds, development.

Currently, `/doc` [is gitignored](https://github.com/rust-lang/miri/blob/861fb3b6f42f6ca1952c475c0a4ad4cf14f72612/.gitignore#L2), can I remove that?

Cargo.toml
Patrick-6: IIRC that's what Michalis mentioned at some point, but I'm not sure endianness even matters currently, since there is no mixed-size support yet.

Patrick-6:
Should the build script not download anything if cargo is invoked with [`--offline` or `--frozen`](https://doc.rust-lang.org/cargo/commands/cargo.html?highlight=frozen#manifest-options) (which causes cargo to not access the network)? I can't really find anything about how a build script should behave in such a case, or if it can even detect it. There is the [`CARGO_NET_OFFLINE` env. var](https://doc.rust-lang.org/cargo/reference/environment-variables.html#configuration-environment-variables), but that appears to be only for changing cargo's behavior, it's not one of the [env. vars set for build scripts.](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts).

genmc-sys/build.rs
```diff
@@ -0,0 +1,235 @@
...
+    config.define("BUILD_MODEL_CHECKER", "ON");
+
+    config.build_target(GENMC_MODEL_CHECKER);
+    let _dst = config.build();
+
+    let genmc_model_checker_build_directory = genmc_build_path.join(GENMC_MODEL_CHECKER);
+    println!("cargo::rustc-link-search=native={}", genmc_model_checker_build_directory.display());
+    println!("cargo::rustc-link-lib=static={GENMC_MODEL_CHECKER}");
+
+    genmc_build_path
```
Patrick-6: I'm not quite happy with this code yet:
- The Hack at the top seems to be required for building on older Ubuntu, which isn't relevant yet for Miri's CI, but might be if this should build in rustc's CI (or someone tries to compile Miri-GenMC on an older OS).
  - I suspect it's due to version differences with the older cmake that's installed there, possibly an edge case that's not handled by the cmake crate. We can either leave it as is or try to investigate further.
- The "build" path extension seems to be undocumented by the cmake crate (it is added [here](https://docs.rs/cmake/0.1.54/src/cmake/lib.rs.html#557)), so it might not be a good idea to depend on that.
  - This might be unlikely to change, especially since the [cmake crate](https://github.com/rust-lang/cmake-rs) lives in the `rust-lang` Github project too.
  - We may be able to use the return value (`_dst`) from `config.build()` instead (which should be the install directory), but I haven't yet figured out why the produced `model_checker` library is not in that directory. I'll have to investigate this further.

As discussed, I've switched to using a config.h file instead of compile definitions. This config file lives in the build directory (at `{genmc_build_path}/config.h`), which is then included in the `cxx_bridge` build step (That's why we return that path here).
With this config.h, no more manual settings syncing is required.

genmc-sys/build.rs
```diff
@@ -0,0 +1,235 @@
...
+    // Build all required components:
+    let genmc_build_path = build_genmc_model_checker(&genmc_path);
+    build_cxx_bridge(&genmc_path, &genmc_build_path);
+
+    // Only rebuild if anything changes:
+    // Note that we don't add the downloaded GenMC repo, since that should never be modified manually.
+    // Adding that path here would also trigger an unnecessary rebuild after the repo is cloned (since cargo detects that as a file modification).
+    println!("cargo::rerun-if-changed={RUST_CXX_BRIDGE_FILE_PATH}");
+    println!("cargo::rerun-if-changed=./src_cpp");
+    println!("cargo::rerun-if-changed={GENMC_LOCAL_PATH}");
```
Patrick-6: For some reason, adding this path to the `rerun-if-changed` if the path doesn't exist yet causes full rebuilds on every build. I think the non-existing directory is seen as a change by cargo.

Here is the output from running `CARGO_LOG=cargo::core::compiler::fingerprint=info ./miri build --features=genmc` after a complete build without any file modifications (as described [here](https://doc.rust-lang.org/cargo/faq.html#why-is-cargo-rebuilding-my-code)):
```
cargo::core::compiler::fingerprint: stale: missing "/path/to/miri/genmc-sys/./genmc/"
cargo::core::compiler::fingerprint: fingerprint dirty for miri v0.1.0 (/path/to/miri)
```

Is this intended behavior for cargo?

genmc-sys/build.rs
Patrick-6: The main difficulty is the lifetime in `Commit<'repo>` on `repo`, which made it hard to reuse the commit for the 2 match arms. I've reworked how this code works, it should be less duplicate work now.

genmc-sys/build.rs
Patrick-6: (I've moved over to using the `config.h` file now.)

src/bin/miri.rs
Patrick-6: So this warning would always show up unless the user manually passes arguments to disable borrow tracking (since it's on by default)?

genmc-sys/build.rs
```diff
@@ -0,0 +1,235 @@
...
+/// Path where the `cxx_bridge!` macro is used to define the Rust-C++ interface.
+const RUST_CXX_BRIDGE_FILE_PATH: &str = "src/lib.rs";
+
+// FIXME(GenMC, build): decide whether to keep debug enabled or not (without this: calling BUG() ==> UB).
+// FIXME(GenMC, build): decide on which profile to use.
+
+/// Enable/disable additional debug checks, prints and options for GenMC.
+const ENABLE_GENMC_DEBUG: bool = true;
+/// The profile with which to build GenMC.
+const GENMC_CMAKE_PROFILE: &str = "RelWithDebInfo";
```
Patrick-6: Reminder to discuss this with Michalis before merging since the original comment is now marked as outdated.

genmc-sys/build.rs
RalfJung: I have no idea what it is gitignored... yes please remove that.

Cargo.toml
RalfJung: Ah, but with mixed-size it might indeed matter, fair... let's just limit to little-endian for now if that's the only thing that is tested.

RalfJung:
> Should the build script not download anything if cargo is invoked with [`--offline` or `--frozen`](https://doc.rust-lang.org/cargo/commands/cargo.html?highlight=frozen#manifest-options) (which causes cargo to not access the network)? I can't really find anything about how a build script should behave in such a case, or if it can even detect it. There is the [`CARGO_NET_OFFLINE` env. var](https://doc.rust-lang.org/cargo/reference/environment-variables.html#configuration-environment-variables), but that appears to be only for changing cargo's behavior, it's not one of the [env. vars set for build scripts.](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts).

Yeah I don't think we have a way to check for that.

Please ask on some RUst forum (URLO on Zulip) what the typical approach here is, I don't know.

genmc-sys/build.rs
RalfJung: - What is that `-B` flag and why is setting it a hack?
- If the cmake crate currently provides no way of getting the build dir, then maybe open an issue to ask for such an API, and reference it here?

genmc-sys/build.rs
```diff
@@ -0,0 +1,235 @@
...
+                let commit = update_local_repo(&repo, commit_oid);
+                checkout_commit(&repo, &commit);
+            }
+            Err(_) => {
+                let repo = clone_remote_repo(&genmc_download_path);
+                let Ok(commit) = repo.find_commit(commit_oid) else {
+                    println!(
+                        "cargo::error=Cloned GenMC repository does not contain required commit '{GENMC_COMMIT}'"
+                    );
+                    fatal_error();
```
RalfJung: Can't this just be a `panic!`? That's the usual way build scripts report an error, I think.

genmc-sys/build.rs
```diff
@@ -0,0 +1,235 @@
...
+        if statuses.is_empty() {
+            return;
+        }
+
+        /// Printing too many files makes reading the error message difficult, so we limit the number.
+        const PRINT_LIMIT: usize = 8;
+
+        println!();
+        println!(
+            "cargo::error=Downloaded GenMC repository at path '{GENMC_DOWNLOAD_PATH}' has been modified:"
```
RalfJung: I think you can just make all of these regular `eprintln!` without the `cargo::error` stuff, and then `panic!` at the end, and all that output will be shown.

Also, it is enough to just say that the folder has been modified, no need to have all this logic that lists the differences.

genmc-sys/build.rs
```diff
@@ -0,0 +1,235 @@
...
+        .file("./src_cpp/MiriInterface.hpp")
+        .file("./src_cpp/MiriInterface.cpp")
+        .compile("genmc_interop");
+
+    // Link the Rust-C++ interface library generated by cxx_build:
+    println!("cargo::rustc-link-lib=static=genmc_interop");
+}
+
+/// Build the GenMC model checker library.
+/// Returns the path
```
RalfJung: Which path?

genmc-sys/build.rs
```diff
@@ -0,0 +1,235 @@
...
+    let Ok(genmc_local_path) = PathBuf::from_str(GENMC_LOCAL_PATH);
+    let genmc_path = if genmc_local_path.exists() {
+        genmc_local_path
+    } else if cfg!(feature = "download_genmc") {
+        downloading::download_genmc()
+    } else {
+        println!(
+            "cargo::error=GenMC not found in path '{GENMC_LOCAL_PATH}', and downloading GenMC is disabled."
+        );
+        fatal_error();
```
RalfJung: More cases that could be just a panic. Probably applies to all `fatal_error` users.

genmc-sys/build.rs
RalfJung: Can you just add `GENMC_LOCAL_PATH` only if you actually use the local path, not the download?

genmc-sys/build.rs
Patrick-6: The `-B` flags sets the build directory, which I would expect to be taken care of automatically by the `cmake` crate. When I tested it on Ubuntu 22.04, the first of the two `cmake` command invocation used the wrong directory without this (first invocation is a configure step, the second invocation is the actual build).
The difference between the OSes makes me think this is not intended behavior, maybe I should make a bug report for that too.

genmc-sys/build.rs
Patrick-6: Yes, but I think then we wouldn't trigger an automatic rebuild if a user adds the `./genmc` directory. That's probably still better than constantly rebuilding, since this case should be rare.

src/bin/miri.rs
RalfJung: Yes.

genmc-sys/build.rs
RalfJung: Oh, you are using the existence of the directory as the signal for using it... yeah that doesn't work with cargo's model, it seems. I'll ask the cargo devs about it, but for now, let's use an env var to signal explicitly that you want a local build instead of a download, and then add that env var to rerun-if-changed.

genmc-sys/build.rs
RalfJung: What is the default used by GenMC itself?

Patrick-6: Profile: `RelWithDebInfo` (with `-O2 -g`)
GENMC_DEBUG: `OFF`

genmc-sys/build.rs
RalfJung: Also see https://github.com/rust-lang/cargo/issues/6003

genmc-sys/build.rs
RalfJung: We should do the same then for release builds. For debug builds it'd make sense to set GENMC_DEBUG to On.

src/bin/miri.rs
Patrick-6: For the Miri-GenMC tests (not in this PR), should I disable borrow tracking for all of them or add the warning to the expected output?

Patrick-6: For the warning, should I use `tracing::warn!`? We don't have a `tcx` [where the check will be](https://github.com/Patrick-6/miri/blob/c9415f8148e3c394db2804fc396f1b2b9e67ae8c/src/bin/miri.rs#L767-L772) (for using `tcx.dcx().warn` like in `after_analysis`).
Or, should I delay the warning until we reach `after_analysis`?

RalfJung: > For the Miri-GenMC tests (not in this PR), should I disable borrow tracking for all of them or add the warning to the expected output?

Yes.


> For the warning, should I use `tracing::warn!`? We don't have a `tcx` [where the check will be](https://github.com/Patrick-6/miri/blob/c9415f8148e3c394db2804fc396f1b2b9e67ae8c/src/bin/miri.rs#L767-L772) (for using `tcx.dcx().warn` like in `after_analysis`). Or, should I delay the warning until we reach `after_analysis`?

You can just use `eprintln!("warning: ...")`.

Patrick-6: > > For the Miri-GenMC tests (not in this PR), should I disable borrow tracking for all of them or add the warning to the expected output?
> 
> Yes.

which one, adding the warning?

RalfJung:
> which one, adding the warning?

Ah, sorry, stopped reading too early...
Add the flag to disable SB to the tests.

genmc-sys/build.rs
Patrick-6: I can't find any way for the build script to even detect if `debug_assertions` is enabled.
In the [list of env vars set for a build script](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts), there is only:
- `DEBUG`, which I assume is whether or not debug info is included (assuming it matches [the "debug" in Cargo.toml](https://doc.rust-lang.org/cargo/reference/profiles.html#debug))
- `PROFILE`, which I don't think will help us much for Miri since IIRC `miri-script` never uses `--release` anyway (?)

I can't really imagine that we are the first who would like to know if debug assertions are enabled from a build script, so I might be missing something here.

RalfJung: > PROFILE, which I don't think will help us much for Miri since IIRC miri-script never uses --release anyway (?)

`./miri install` uses `--release`, as does the dist build.
So if debug_assertions cannot be checked (I didn't find a way either) I think PROFILE is exactly what you should check.

genmc-sys/build.rs
```diff
@@ -0,0 +1,201 @@
...
+    use std::path::{Path, PathBuf};
+    use std::str::FromStr;
+
+    use git2::{Commit, Oid, Repository, StatusOptions};
+
+    use super::GENMC_LOCAL_PATH;
+
+    pub(crate) const GENMC_GITHUB_URL: &str = "https://github.com/Patrick-6/genmc.git";
+    pub(crate) const GENMC_COMMIT: &str = "2f503036ae14dc91746bfc292d142f332f31727e";
+    pub(crate) const GENMC_DOWNLOAD_PATH: &str = "./downloaded/genmc/";
```
RalfJung: For all these paths it'd be good to know what they are relative to.^^

genmc-sys/build.rs
```diff
@@ -0,0 +1,201 @@
...
+        genmc_local_path
+    } else if cfg!(feature = "download_genmc") {
+        downloading::download_genmc()
+    } else {
+        panic!("GenMC not found in path '{GENMC_LOCAL_PATH}', and downloading GenMC is disabled.");
+    };
+
+    // Build all required components:
+    let genmc_install_dir = build_genmc_model_checker(&genmc_path);
+    build_cxx_bridge(&genmc_path, &genmc_install_dir);
```
RalfJung: This should only use `genmc_install_dir` since the headers should also be installed.

doc/GenMC.md
```diff
@@ -0,0 +1,42 @@
...
+## Development
+
+GenMC is written in C++, which complicates development a bit.
+For Rust-C++ interop, Miri uses [CXX.rs](https://cxx.rs/), and all handling of C++ code is contained in the `genmc-sys` crate (located in the Miri repository root directory: `miri/genmc-sys/`).
+
+The actual code for GenMC is not contained in the Miri repo itself, but in a [separate GenMC repo](https://github.com/MPI-SWS/genmc) (with different maintainers).
+Note that this repo is just a mirror repo.
+<!-- FIXME(genmc): define how submitting code to GenMC should be handled. -->
+
+<!-- FIXME(genmc): explain development. -->
```
Patrick-6: I've added some basic documentation, anything else the should already be added in this PR?

RalfJung: Please document how the genmc source code is fetched: there is some env var (?) that lets you decide whether it downloads genmc or uses a version stored in some folder.

doc/GenMC.md
RalfJung, line 1: Please rename the file to `genmc`; filenames default to lowercase.

doc/GenMC.md
Patrick-6: Currently there isn't such an env var, how should it be called? `GENMC_PATH`?

RalfJung: I'd suggest `GENMC_SRC_PATH`.

genmc-sys/build.rs
Patrick-6: [The docs](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts) advise against using this variable:

> PROFILE — release for release builds, debug for other builds. This is determined based on if the [profile](https://doc.rust-lang.org/cargo/reference/profiles.html) inherits from the [dev](https://doc.rust-lang.org/cargo/reference/profiles.html#dev) or [release](https://doc.rust-lang.org/cargo/reference/profiles.html#release) profile. Using this environment variable is not recommended. Using other environment variables like OPT_LEVEL provide a more correct view of the actual settings being used.

It's probably fine for our use case, but should we make a feature request from cargo to expose the `debug_assertion` setting?

RalfJung: Yeah I'm aware of the downsides but we can't query the debug assertions so :shrug: 


> It's probably fine for our use case, but should we make a feature request from cargo to expose the debug_assertion setting?

It's called `debug_assertions`, but yes that seems like a reasonable question.

Patrick-6:
How should I resolve the merge conflicts? Using the Github UI or by rebasing on the latest `master` branch?

RalfJung:
Please do a rebase. (I have no idea what GH's UI does, I never used it. If it can resolve conflicts in a way that the result is a rebased PR, feel free to use it. What is important is that the PR has no merge commits.)

genmc-sys/build.rs
```diff
@@ -0,0 +1,258 @@
...
+        panic!("llvm-config command failed: {}", String::from_utf8_lossy(&output.stderr));
+    }
+    let cpp_flags = String::from_utf8(output.stdout)
+        .expect("llvm-config output should be valid UTF-8")
+        .trim()
+        .to_string();
+
+    println!("cargo::warning=LLVM cpp_flags: '{cpp_flags}'");
+    cpp_flags
+}
```
Patrick-6: This is the best solution I've found so far.
For the second compilation step in [cmake build, cxx build], we need to add the LLVM compile definitions (which consists of the include path and some preprocessor definitions), otherwise it cannot find the required header files (and it might misbehave due to the missing defines).

This works as follows:
- Get the LLVM version from the `config.h` file generated by the cmake step. This way it doesn't matter which LLVM version that the GenMC build chooses. The syntax of that file is controlled by GenMC and unlikely to change.
- Link that LLVM version (dynamic linking). This is always `LLVM-{llvm_version}`, at least for existing LLVM versions.
- Call `llvm-config-{llvm_version} --cpp_flags`, which output all the compiler flags we need.
- Add these flags to the cxx build step.

I think this should be reasonably stable, and it is only temporary until the LLVM dependency is no longer required (famous last words ^^).

I've looked into dynamically linking GenMC so we don't have to link LLVM manually, but that has its own difficulties (e.g., where to install the genmc `.so` file), and then we still need the compile definitions for the cxx build, so almost no benefits.

RalfJung: Why does the cxx build need the LLVM info? Does LLVM leak into the *public interface* of GenMC somehow?

Patrick-6: We haven't yet clearly defined what the "public interface" for the GenMC library should contain, so currently this interface includes everything that is in any header file.

The most important header file we include is `GenMCDriver.hpp`, which transitively includes headers from all over GenMC (including the LLI interpreter code, which depends on LLVM).
This is one of the reasons I've been pushing to get rid of the LLVM dependency from the GenMC parts we need for Miri as fast as possible.

genmc-sys/build.rs
Patrick-6: There was [a discussion on Zulip](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/Allowing.20build.20scripts.20to.20determine.20debug_assertion.20status/with/319890956) about this in 2023, which looks like it didn't lead to any changes.

RalfJung: Ah, good find.

Seems useful to have an issue tracking this, so please file one.

Patrick-6: [Done, issue is created.](https://github.com/rust-lang/cargo/issues/15760)

RalfJung: Great, thanks.

For now, check `PROFILE` as the closest approximation.

genmc-sys/build.rs
```diff
@@ -0,0 +1,260 @@
...
+// Check out doc/GenMC.md for more info.
+
+/// Path used for development of Miri-GenMC.
+/// A GenMC repository in this directory (relative to the `genmc-sys` directory) will take precedence over the downloaded GenMC repository.
+/// If the `download` feature is disabled, this path must contain a GenMC repository.
+const GENMC_LOCAL_PATH: &str = "./genmc/";
+
+/// Path where the downloaded GenMC repository will be stored (relative to the `genmc-sys` directory).
+/// Note that this directory is *not* cleaned up automatically by `cargo clean`.
+#[cfg(feature = "download_genmc")]
```
RalfJung: Why is this a separate `feature`?

genmc-sys/build.rs
```diff
@@ -0,0 +1,260 @@
...
+            genmc_src_path.display()
+        );
+        genmc_src_path
+    } else if genmc_local_path.exists() {
+        // If the local repository exists, cargo should watch it for changes:
+        // FIXME(genmc,cargo): We could always watch this path even if it doesn't (yet) exist, depending on how `https://github.com/rust-lang/cargo/issues/6003` is resolved.
+        //                     Adding it here means we don't rebuild if a user creates `GENMC_LOCAL_PATH`, which isn't ideal.
+        //                     Cargo currently always rebuilds if a watched directory doesn't exist, so we can only add it if it exists.
+        println!("cargo::rerun-if-changed={GENMC_LOCAL_PATH}");
+        genmc_local_path
```
RalfJung:
```suggestion
```
The point of the env var was to control whether we download or not, wasn't it?

genmc-sys/build.rs
Patrick-6: No particular reason, I've just noticed that it can be a feature since it's cleanly separated from the rest.
Should I remove that (i.e., make it always enabled)?

genmc-sys/build.rs
Patrick-6: I thought we would keep this so you can use the normal `./miri` commands for development, without having to set the env var, by having your development copy of GenMC in `genmc-sys/genmc/`.

RalfJung: Normal `./miri` doesn't turn on the feature. `./miri build --feature genmc` should use the downloaded version -- that's the right default for most contributors.

genmc-sys/build.rs
RalfJung: Features have a huge cost in the sense that every feature doubles the size of the test matrix of possible build settings.

So, yeah, if there's no particular reason then please make it always-enabled.

genmc-sys/build.rs
Patrick-6: So for Miri-GenMC contributors the way to do it is `GENMC_SRC_PATH="/some/path/" ./miri build --features=genmc` ? (or `export GENMC_SRC_PATH=...`)

RalfJung: For the extremely rare contributor that needs a custom genmc, yeah.

It'd be nicer if you just had to create a folder, but that doesn't seem to be possible with cargo currently.

Patrick-6: But that is what we have if I don't remove that `if`

Patrick-6: The current implementation uses:
- path in `GENMC_SRC_PATH` if set
- `genmc-sys/genmc` if present
- `genmc-sys/downloaded/genmc` if present
- do the download into `downloaded/genmc`

RalfJung: > But that is what we have if I don't remove that `if`

But it doesn't work: if you first build without the folder existing, and then you create the folder, it won't rebuild and will keep using the downloaded genmc.

Patrick-6: Nothing a little `touch genmc-sys/build.rs` doesn't fix ;)

But ok, we can remove it.

RalfJung: > Nothing a little touch genmc-sys/build.rs doesn't fix ;)

If you prefer that, we can also do that, and carefully document it.

But we don't want that *and* the env var. Let's have one way and one way only for local builds.

genmc-sys/build.rs
```diff
@@ -0,0 +1,259 @@
...
+/// Info required to be passed to subsequent compilation steps.
+struct CompileInfo {
+    /// Directory where cmake installs the GenMC library into.
+    pub genmc_install_dir: PathBuf,
+    // FIXME(genmc,llvm): remove once LLVM dependency is removed.
+    pub llvm_definitions: String,
+    pub llvm_include_dirs: String,
+    // FIXME(genmc,cmake): Remove once the GenMC debug setting is available in the config.h file.
+    pub enable_genmc_debug: bool,
+}
```
Patrick-6: I've switched to a struct here since the amount of parameters passed between the two compilation steps got a bit too big. Maybe we should just combine both compilation steps into one function?

For the last parameter `enable_genmc_debug`: Due to time constraints this one will likely not be available the `config.h` file in the GenMC release, so I'm passing it to the second compilation step manually again for now.

RalfJung: Yeah, having them both in one function is fine. You can use comments to make the structure clear.

genmc-sys/build.rs
Patrick-6: I've switched to only the env var + download ways, it is still reasonably usable by doing `export GENMC_SRC_PATH=...` in the current shell, then the path doesn't need to be passed every time.

etc/rust_analyzer_helix.toml
```diff
@@ -5,6 +5,7 @@ source = "discover"
 linkedProjects = [
     "Cargo.toml",
     "cargo-miri/Cargo.toml",
+    "genmc-sys/Cargo.toml",
     "miri-script/Cargo.toml",
 ]
```
Patrick-6, line 10: Apparently using the `linkedProjects` setting means that rust-analyzer auto-discovery is turned off, so you get no RA help at all in directories no listed here. Adding the `genmc-sys` lines fixes it for VSCode, but I can't test it for helix (no reason to believe it wouldn't work though).

There still another issue (maybe not for this PR): I still don't get any RA help in any Rust files in the `miri/tests/` directory, but maybe that could be added too somehow? I'm unsure how this would work though, since the settings file here also overrides the check command to use `./miri clippy/check` instead of `cargo clippy/check`, which may not work with the tests directory.

Also, this setting expects paths to either `Cargo.toml`, `rust-project.json`, or loose `.rs` files, and we definitely don't want to add all tests here.

RalfJung: Yeah RA doesn't have support for ui test suites like we're using them, unfortunately.

miri-script/src/commands.rs
```diff
@@ -757,8 +757,8 @@ impl Command {
                 if ty.is_file() {
                     name.ends_with(".rs")
                 } else {
-                    // dir or symlink. skip `target` and `.git`.
-                    &name != "target" && &name != ".git"
+                    // dir or symlink. skip `target`, `.git` and `downloaded` (last is used in `genmc-sys` subcrate of Miri).
+                    &name != "target" && &name != ".git" && &name != "downloaded"
```
Patrick-6: This was an interesting issue, `./miri fmt` formatted the new Rust frontend tests in the downloaded GenMC repo, which the `genmc-sys` build script then detects as a disallowed modification. ^^

This fix is not perfect, there might be a name collision in the future (probably unlikely). More important, for development I usually placed the GenMC repo in the `genmc-sys/genmc` directory, and this will path will still be formatted like before.

We would need to match on something like `name.starts_with("genmc") && entry.path().parent().unwrap().name().unwrap() == "genmc-sys"` here too, but that seems quite complicated (we can't just match on `genmc`, since that would exclude `src/concurrency/genmc` too).

genmc-sys/build.rs
```diff
@@ -0,0 +1,261 @@
...
+    use std::str::FromStr;
+
+    use git2::{Commit, Oid, Remote, Repository, StatusOptions};
+
+    use super::GENMC_DOWNLOAD_PATH;
+
+    /// The GenMC repository the we get our commit from.
+    pub(crate) const GENMC_GITHUB_URL: &str = "https://github.com/MPI-SWS/genmc.git";
+    /// The GenMC commit we depend on. It must be available on the specified GenMC repository.
+    pub(crate) const GENMC_COMMIT: &str = "3438dd2c1202cd4a47ed7881d099abf23e4167ab";
```
Patrick-6, line 31: We are now using the official GenMC release, not my cloned repo

miri-script/src/commands.rs
RalfJung: Just give it a more unique name, like `genmc-src`.

Patrick-6:
As mentioned on Zulip, the license changes are still missing, so this probably shouldn't be merged yet, but otherwise:

- From a technical point of view, everything should be in order now.
- The documentation might need to be improved/cleaned up/expanded.
- We don't yet have any formatting & linting rules (and automatic checks) for the C++ code that lives in the Miri repo (but that's a low priority for now).

@rustbot ready

genmc-sys/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "genmc-sys"
+version = "0.1.0"
+edition = "2024"
```
RalfJung, line 8: This need a license declaration, and it should probably be the genmc GPL license to ensure this is reflected somewhere on the Rust side.

RalfJung:
```suggestion
edition = "2024"
license = "GPL-3.0-or-later"
```

genmc-sys/build.rs
```diff
@@ -0,0 +1,261 @@
...
+        .file("./src_cpp/MiriInterface.hpp")
+        .file("./src_cpp/MiriInterface.cpp")
+        .compile("genmc_interop");
+
+    // Link the Rust-C++ interface library generated by cxx_build:
+    println!("cargo::rustc-link-lib=static=genmc_interop");
+}
+
+fn main() {
+    // Select which path to use for the GenMC repo:
```
RalfJung, line 245:
```suggestion
    // Make sure we don't accidentally distribute a binary with GPL code.
    if option_env!("RUSTC_STAGE").is_some() {
        panic!("genmc should not be enabled in the rustc workspace since it includes a GPL dependency");
    }

    // Select which path to use for the GenMC repo:
```

Cargo.toml
```diff
@@ -66,7 +70,7 @@ harness = false
 
 [features]
 default = ["stack-cache", "native-lib"]
-genmc = []
+genmc = ["dep:genmc-sys"]
```
RalfJung:
```suggestion
genmc = ["dep:genmc-sys"] # this enables a GPL dependency!
```

genmc-sys/LICENSE
Patrick-6, line 1: FWIU, this file is required too for the GPL license

genmc-sys/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+license = "GPL-3.0-or-later"
+name = "genmc-sys"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+cxx = { version = "1.0.160", features = ["c++20"] }
+
+[build-dependencies]
+cc = "1.2.30"
```
Patrick-6, line 14: I think this line should fix the failing build in the rustc workspace

genmc-sys/Cargo.toml
Patrick-6: I've added the license already locally before I saw your comment

genmc-sys/build.rs
```diff
@@ -0,0 +1,268 @@
...
+            PathBuf::from_str(&genmc_src_path).expect("GENMC_SRC_PATH should contain a valid path");
+        assert!(
+            genmc_src_path.exists(),
+            "GENMC_SRC_PATH={} does not exist!",
+            genmc_src_path.display()
+        );
+        genmc_src_path
+    } else {
+        downloading::download_genmc()
+    };
```
Patrick-6, line 257: I've just noticed that when using rust-analyzer, this behavior means you need to set/unset the `GENMC_SRC_PATH` env var both for the terminal and in the RA settings, otherwise one will use the downloaded path instead.

It's a bit annoying (to figure out too), so we should add it to the docs.

RalfJung: I mean you also need to enable the feature explicitly in the RA settings so yeah that's expected.

RalfJung:
Thanks!

It's probably easiest for me to just do the remaining changes directly instead of you trying to read my mind. Usually I would just push that to your branch, but unfortunately you have disabled the setting that would let me do that. So I had to make a new PR instead: https://github.com/rust-lang/miri/pull/4498. Please go through the 2nd commit there to see what I changed.

doc/genmc.md
RalfJung, line 1: I have drastically shortened these docs since they should explain the interface o how to build and work with Miri-GenMC, not all the implementation details.

src/concurrency/thread.rs
```diff
@@ -677,8 +677,6 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
     fn run_on_stack_empty(&mut self) -> InterpResult<'tcx, Poll<()>> {
         let this = self.eval_context_mut();
         // Inform GenMC that a thread has finished all user code. GenMC needs to know this for scheduling.
-        // FIXME(GenMC): Thread-local destructors *are* user code, so this is odd. Also now that we
-        // support pre-main constructors, it can get called there as well.
```
RalfJung, line 681: Why did you remove this comment? It is still correct, AFAIK.

PR #4456, opened 2025-07-08
trace: incorporate events

nia-e:
If tracing is enabled, use the info from the trace to actually update Miri's state. Pending a small change to rustc to expose some necessary interfaces to handle writes, this should be good!


nia-e:
Ah, I forgot I rebased this on top of #4435. Just the last commit is relevant for now

src/shims/native_lib/mod.rs
```diff
@@ -169,6 +211,52 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            let (alloc_addr, alloc_len) = {
+                let alloc = this.get_alloc_raw(alloc_id)?;
+                (alloc.get_bytes_unchecked_raw().addr(), alloc.len())
+            };
+
+            let unshifted_overlap = std::cmp::max(evt_rg.addr, alloc_addr)
+                ..std::cmp::min(
+                    evt_rg.addr.strict_add(evt_rg.size),
+                    alloc_addr.strict_add(alloc_len),
+                );
```
nia-e, line 238: Given that we assert 1 access = 1 allocation, should we skip this entirely and just error if the access is out of bounds?

src/shims/native_lib/mod.rs
```diff
@@ -169,6 +211,52 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // Get the (size, len) pair for the current allocation.
+            let (alloc_addr, alloc_len) = {
+                let alloc = this.get_alloc_raw(alloc_id)?;
+                (alloc.get_bytes_unchecked_raw().addr(), alloc.len())
+            };
+
+            let unshifted_overlap = std::cmp::max(evt_rg.addr, alloc_addr)
+                ..std::cmp::min(
+                    evt_rg.addr.strict_add(evt_rg.size),
+                    alloc_addr.strict_add(alloc_len),
```
nia-e: Given that we assert 1 access = 1 allocation, should we skip this step and just error if it's out of bounds?

RalfJung:
What this PR definitely needs is some tests. :) Remember that tracing mode is off-by-default; none of the tests hit the tracing infra. So this needs some new native-mode tests that specifically check that we detect the ranges properly. For instance, sharing a 2-field uninit struct with C, having C initialize the first field, and then reading the 2nd field -- that should be UB.

nia-e:
I actually wrote some! Just didn't know if they belong here. I'll add them, though

src/shims/native_lib/mod.rs
```diff
@@ -169,6 +211,52 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            if evt.is_read() {
+                let alloc = this.get_alloc_raw(alloc_id)?;
+                let p_map = alloc.provenance();
+                for idx in overlap {
+                    // If a provenance was read by the foreign code, expose it.
+                    if let Some(prov) = p_map.get(Size::from_bytes(idx), this) {
+                        this.expose_provenance(prov)?;
+                    }
+                }
+            } else {
```
nia-e: `else if info.mutability.is_mut()` pending the discussion in the last PR

nia-e:
I did readd logic (not much code) for whether an access spands multiple allocations since from what I understand this is theoretically allowed by LLVM at least and I'm a little scared it could happen in the wild; but I can replace that with just an assertion that accesses span a single allocation and ICE/error otherwise.

@rustbot ready

src/shims/native_lib/mod.rs
```diff
@@ -196,6 +218,60 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
     }
+
+    /// Applies the `events` to Miri's internal state. The event vector must be
+    /// ordered sequentially by when the accesses happened, and the sizes are
+    /// assumed to be exact.
+    fn tracing_apply_accesses(&mut self, events: MemEvents) -> InterpResult<'tcx> {
+        let this = self.eval_context_mut();
+
+        for evt in events.acc_events {
+            //eprintln!("evt: {evt:?}");
```
nia-e: Oops, left in some debug bits. I'll delete

tests/native-lib/fail/trace_read.rs
```diff
@@ -0,0 +1,26 @@
+//@only-target: linux
+//@only-target: gnu
+//@only-target: x86
```
RalfJung: You can just do `only-target: x86_64-unknown-linux-gnu`.

tests/native-lib/fail/trace_read.rs
RalfJung, line 1: Please add a comment explaining what the test does. Please put the tests that need tracing into a `tracing` subfolder. And please give it a name that reflects its purpose, e.g. `tracing/unread_ptr_not_exposed.rs`.

tests/native-lib/fail/trace_read.rs
```diff
@@ -0,0 +1,26 @@
...
+    unexposed_reachable_alloc();
+}
+
+fn unexposed_reachable_alloc() {
+    let inner = 42;
+    let intermediate = &raw const inner;
+    let exposed = (&raw const intermediate).expose_provenance();
+    unsafe { do_nothing() };
+    let invalid: *const i32 = std::ptr::with_exposed_provenance(intermediate.addr());
+    let valid: *const *const i32 = std::ptr::with_exposed_provenance(exposed);
```
RalfJung: What's the point of `valid`?

tests/native-lib/fail/trace_read.stderr
```diff
@@ -0,0 +1,40 @@
...
+   = help: alternatively, `MIRIFLAGS=-Zmiri-permissive-provenance` disables this warning
+   = note: BACKTRACE:
+   = note: inside `unexposed_reachable_alloc` at tests/native-lib/fail/trace_read.rs:LL:CC
+note: inside `main`
+  --> tests/native-lib/fail/trace_read.rs:LL:CC
+   |
+LL |     unexposed_reachable_alloc();
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+error: Undefined Behavior: pointer not dereferenceable: pointer must be dereferenceable for 4 bytes, but got $HEX[noalloc] which is a dangling pointer (it has no provenance)
```
RalfJung: This error does not look like it has anything to do with tracing?

tests/native-lib/fail/trace_write.rs
RalfJung, line 1: Also needs a more clear filename, e.g. `tracing/uninit.rs`.

tests/native-lib/fail/trace_write.rs
```diff
@@ -0,0 +1,24 @@
...
+
+fn main() {
+    partial_init();
+}
+
+fn partial_init() {
+    let mut slice = std::mem::MaybeUninit::<[u8; 3]>::uninit();
+    let slice_ptr = slice.as_mut_ptr().cast::<u8>();
+    unsafe {
+        init_n(2, slice_ptr);
```
RalfJung:
```suggestion
        // Initialize the first two elements.
        init_n(2, slice_ptr);
```

tests/native-lib/fail/trace_write.rs
```diff
@@ -0,0 +1,24 @@
...
+        assert!(*slice_ptr == 0);
+        assert!(*slice_ptr.offset(1) == 0);
+        let _val = *slice_ptr.offset(2); //~ ERROR: Undefined Behavior: using uninitialized data
```
RalfJung:
```suggestion
        // Reading the third is UB!
        let _val = *slice_ptr.offset(2); //~ ERROR: Undefined Behavior: using uninitialized data
```

tests/native-lib/fail/trace_read.rs
```diff
@@ -0,0 +1,26 @@
+//@only-target: linux
+//@only-target: gnu
+//@only-target: x86
+//@only-on-host
```
RalfJung: You don't need this, the native tests anyway run only on the host.

src/shims/native_lib/mod.rs
```diff
@@ -269,15 +346,20 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             // be read by FFI. The `black_box` is defensive programming as LLVM likes
             // to (incorrectly) optimize away ptr2int casts whose result is unused.
             std::hint::black_box(alloc.get_bytes_unchecked_raw().expose_provenance());
-            // Expose all provenances in this allocation, since the native code can do $whatever.
-            for prov in alloc.provenance().provenances() {
-                this.expose_provenance(prov)?;
+
+            if !tracing {
+                // Expose all provenances in this allocation, since the native code can do $whatever.
```
RalfJung, line 355:
```suggestion
                // Expose all provenances in this allocation, since the native code can do $whatever.
                // Can be skipped when tracing; in that case we'll expose just the actually-read parts later.
```

src/shims/native_lib/mod.rs
```diff
@@ -269,15 +346,20 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                for prov in alloc.provenance().provenances() {
+                    this.expose_provenance(prov)?;
+                }
             }
 
             // Prepare for possible write from native code if mutable.
             if info.mutbl.is_mut() {
                 let (alloc, cx) = this.get_alloc_raw_mut(alloc_id)?;
-                alloc.process_native_write(&cx.tcx, None);
+                if !tracing {
```
RalfJung, line 367:
```suggestion
                // These writes could initialize everything and wreck havoc with the pointers.
                // We can skip that when tracing; in that case we'll later do that only for the memory that got actually written.
                if !tracing {
```

src/shims/native_lib/mod.rs
```diff
@@ -196,6 +218,60 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        for evt in events.acc_events {
+            //eprintln!("evt: {evt:?}");
+            let evt_rg = evt.get_range();
+            // LLVM at least permits vectorising accesses to adjacent allocations,
+            // so we cannot assume 1 access = 1 allocation. :(
+            let mut rg = evt_rg.addr..evt_rg.end();
+            while let Some(curr) = rg.next() {
+                let alloc_id = this
+                    .alloc_id_from_addr(curr.to_u64(), rg.len().try_into().unwrap(), true)
+                    .expect("Foreign code did an out-of-bounds access!");
```
RalfJung: That shouldn't ICE Miri, that should be a `throw_ub_format!`

src/shims/native_lib/mod.rs
```diff
@@ -196,6 +218,60 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let alloc = this.get_alloc_raw(alloc_id)?;
+                let alloc_addr = alloc.get_bytes_unchecked_raw().addr();
+
+                // Skip forward however many bytes of the access are contained in the current allocation.
```
RalfJung: Please explain the logic that happens here.

src/shims/native_lib/mod.rs
```diff
@@ -196,6 +218,60 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let overlap = std::cmp::max(alloc_addr, curr).strict_sub(alloc_addr)
+                    ..std::cmp::min(alloc_addr.strict_add(alloc.len()), rg.end)
+                        .strict_sub(alloc_addr);
+                let _ = rg.advance_by(overlap.len());
```
RalfJung: Add a comment explaining why we ignore this `Result`.

src/shims/native_lib/mod.rs
```diff
@@ -196,6 +218,60 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                //eprintln!("overlap: {overlap:?}");
+
+                // Get an overlap range as an offset from the allocation base addr.
+                //let overlap = unshifted_overlap.start..unshifted_overlap.start.strict_add(unshifted_overlap.len());
+
+                // Reads are infallible, writes might not be.
```
RalfJung: I don't understand this comment.

src/shims/native_lib/mod.rs
```diff
@@ -196,6 +218,60 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                if evt.is_read() {
```
RalfJung: I think an exhaustive match on `AccessType` would be more clear here than an if-else.

src/shims/native_lib/mod.rs
```diff
@@ -196,6 +218,60 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    let p_map = alloc.provenance();
+                    for idx in overlap {
+                        // If a provenance was read by the foreign code, expose it.
+                        if let Some(prov) = p_map.get(Size::from_bytes(idx), this) {
+                            this.expose_provenance(prov)?;
+                        }
+                    }
+                } else if evt.definitely_happened() || alloc.mutability.is_mut() {
```
RalfJung: So we skip this for maybe-writes on const allocs... basically assuming those didn't actually happen? That makes sense, but needs a comment.

src/shims/native_lib/trace/parent.rs
```diff
@@ -386,7 +386,11 @@ fn capstone_find_events(
                         acc_events.push(AccessEvent::Read(push.clone()));
                     }
                     if acc_ty.is_writable() {
-                        acc_events.push(AccessEvent::Write(push));
+                        // FIXME: This could be made certain; either determine all cases where
+                        // only reads happen, or have an intermediate mempr_* function to first
+                        // map the page(s) as readonly and check if a segfault occurred.
+                        // If this did a read as well, it's possible the write didn't happen.
+                        acc_events.push(AccessEvent::Write(push, !acc_ty.is_readable()));
```
RalfJung, line 399: Why does `!readable` imply "definitely written"?

RalfJung (commented):
Okay that's really cool that this works, this is pure magic.^^

I still have a bunch of comments, of course. ;)

tests/native-lib/fail/trace_read.rs
nia-e: Just a sanity check I guess; I can remove it

tests/native-lib/fail/trace_read.stderr
nia-e: It's because `let invalid: *const i32 = std::ptr::with_exposed_provenance(intermediate.addr())` creates an invalid pointer (as the alloc is not exposed). If I remove `-Zmiri-native-lib-enable-tracing`, the test xpasses with no errors

src/shims/native_lib/mod.rs
nia-e: The bit with the overlaps was meant to be deleted (oops). The bit about reads and writes is regarding the discussion with when we can know an access happened for sure; in all cases (on x86), if the tracer process detected a segfault at an instruction that might read, it did read. There exists a possibility on instructions that might write that they didn't write if they also did a read. So, reads are always certain, writes are sometimes certain and sometimes not

src/shims/native_lib/trace/parent.rs
nia-e: Per the options that `iced_x86` offered for [what kind of access an x86 op can do](https://docs.rs/iced-x86/latest/iced_x86/enum.OpAccess.html), the options are
```rust
enum OpAccess {
    Read,
    CondRead,
    Write,
    CondWrite,
    ReadWrite,
    ReadCondWrite,
    // None and NoMemAccess also exist but are irrelevant
}
```
If we got a segfault, then some kind of access did happen for sure, so `CondRead` and `CondWrite` can safely become `Read` and `Write` respectively. The only scenario we might be unsure of is with possibly-RW accesses since those might be `ReadCondWrite` where we only know the read happened for sure

tests/native-lib/fail/trace_read.stderr
RalfJung: > If I remove -Zmiri-native-lib-enable-tracing, the test xpasses with no errors

True, I misunderstood.
Please add more comments in this test, it is quite subtle. :)

tests/native-lib/fail/trace_read.rs
nia-e: Expanded this so that a native read actually happens and we properly check that it does expose whatever it reads but no more

nia-e:
This should address review, lmk if there's anything more!

nia-e:
nvm, seems like I messed something up, just a sec 😭

src/shims/native_lib/mod.rs
RalfJung: Okay so "infallible" is just the wrong word here? What you say has nothing to do with the read/write being fallible.

src/shims/native_lib/trace/parent.rs
RalfJung: Okay. Please include that reasoning as a comment in the code.

src/shims/native_lib/mod.rs
```diff
@@ -226,46 +217,56 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             let evt_rg = evt.get_range();
             // LLVM at least permits vectorising accesses to adjacent allocations,
             // so we cannot assume 1 access = 1 allocation. :(
             let mut rg = evt_rg.addr..evt_rg.end();
             while let Some(curr) = rg.next() {
-                let alloc_id = this
-                    .alloc_id_from_addr(curr.to_u64(), rg.len().try_into().unwrap(), true)
-                    .expect("Foreign code did an out-of-bounds access!");
+                let Some(alloc_id) =
+                    this.alloc_id_from_addr(curr.to_u64(), rg.len().try_into().unwrap(), true)
```
RalfJung: Please add a comment to the `true` argument -- unnamed `bool` need explanation.
```rust
..., /* only_exposed_allocations */ true)
```

src/shims/native_lib/mod.rs
nia-e: Yes, it was just wrong and I updated it in the last commit

src/shims/native_lib/mod.rs
```diff
@@ -226,46 +217,56 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                else {
+                    throw_ub_format!("Foreign code did an out-of-bounds access!")
+                };
                 let alloc = this.get_alloc_raw(alloc_id)?;
                 let alloc_addr = alloc.get_bytes_unchecked_raw().addr();
 
                 // Skip forward however many bytes of the access are contained in the current allocation.
+                // The start of the overlap range will be the greater of the alloc base address and the
+                // lowest remaining address in the access' range; the end will be the lesser of the end
```
RalfJung: Isn't the start of the overlap always `curr` since we go left-to-right?

src/shims/native_lib/mod.rs
```diff
@@ -226,46 +217,56 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                        // If a provenance was read by the foreign code, expose it.
-                        if let Some(prov) = p_map.get(Size::from_bytes(idx), this) {
-                            this.expose_provenance(prov)?;
+
+                match evt {
+                    AccessEvent::Read(_) => {
+                        let p_map = alloc.provenance();
+                        for idx in overlap {
+                            // If a provenance was read by the foreign code, expose it.
+                            if let Some(prov) = p_map.get(Size::from_bytes(idx), this) {
```
RalfJung, line 254: Please add a FIXME saying that we should really have a `get_range` on `ProvenanceMap`.

(Here's a next PR for you if you are interested. ;)

tests/native-lib/pass/ptr_read_access.rs
```diff
@@ -1,8 +1,5 @@
 //@revisions: trace notrace
-//@[trace] only-target: linux
-//@[trace] only-target: gnu
-//@[trace] only-target: x86
-//@[trace] only-on-host
+//@[trace] only-target: x86_64-unknown-linux-gnu x86-unknown-linux-gnu
```
RalfJung: What is `x86-unknown-linux-gnu`? There is no such target. Do you mean `i686-unknown-linux-gnu`?

tests/ui.rs
```diff
@@ -338,6 +343,20 @@ fn main() -> Result<()> {
     if cfg!(unix) && target == host {
         ui(Mode::Pass, "tests/native-lib/pass", &target, WithoutDependencies, tmpdir.path())?;
         ui(Mode::Fail, "tests/native-lib/fail", &target, WithoutDependencies, tmpdir.path())?;
+        if cfg!(all(
+            any(target_arch = "x86", target_arch = "x86_64"),
+            target_os = "linux",
+            target_env = "gnu"
+        )) {
+            // No pass tests here currently.
```
RalfJung: Uh, why did you do this?

If you thought I asked for this, then you misunderstood.
There should be a `tracing` folder inside the existing pass/fail folders.

tests/native-lib/pass/ptr_read_access.rs
nia-e: facepalm

tests/ui.rs
nia-e: oh! that makes more sense, okay

src/shims/native_lib/mod.rs
nia-e: Fair point!

src/shims/native_lib/mod.rs
```diff
@@ -222,21 +222,22 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                // lowest remaining address in the access' range; the end will be the lesser of the end
-                // of the allocation and the end of the access' range. Both are then shifted by `alloc_addr`
+                // The start of the overlap range will be `curr`; the end will be the lesser of the end of
+                // the allocation and the end of the access' range. Both are then shifted by `alloc_addr`
                 // so that the overlap range would have index 0 as the first byte of the allocation (which
                 // is how the methods on the allocation expect it to be).
-                let overlap = std::cmp::max(alloc_addr, curr).strict_sub(alloc_addr)
+                let overlap = curr.strict_sub(alloc_addr)
                     ..std::cmp::min(alloc_addr.strict_add(alloc.len()), rg.end)
                         .strict_sub(alloc_addr);
```
RalfJung: I think the end is easier expressed as `min(rg.end.strict_sub(alloc_addr), alloc.len())`.

src/shims/native_lib/mod.rs
```diff
@@ -222,21 +222,22 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                     throw_ub_format!("Foreign code did an out-of-bounds access!")
                 };
                 let alloc = this.get_alloc_raw(alloc_id)?;
                 let alloc_addr = alloc.get_bytes_unchecked_raw().addr();
 
                 // Skip forward however many bytes of the access are contained in the current allocation.
-                // The start of the overlap range will be the greater of the alloc base address and the
-                // lowest remaining address in the access' range; the end will be the lesser of the end
-                // of the allocation and the end of the access' range. Both are then shifted by `alloc_addr`
+                // The start of the overlap range will be `curr`; the end will be the lesser of the end of
```
RalfJung: Please make it clear that this range is in terms of offsets inside the allocation, i.e., relative to `alloc_addr`.

tests/ui.rs
```diff
@@ -206,6 +206,11 @@ fn run_tests(
         config.program.args.push(flag);
     }
 
+    // Pass on the flag for enabling tracing also when performing those tests.
+    if path.starts_with("tests/native-lib/tracing") {
+        config.program.args.push(std::ffi::OsString::from("-Zmiri-native-lib-enable-tracing"));
+    }
```
RalfJung: This looks like a leftover from the misunderstanding

RalfJung:
This looks great, thanks! After resolving the last nits, please squash the commits using `./miri squash`. Then write `@rustbot ready` after you force-pushed the squashed PR.

@rustbot author

RalfJung:
Lol that commit message is something.^^
image

When you squash and it asks you for the commit message, please take care to write a reasonable commit message. I've cleaned it up for you in this PR, and also did some minor tweaks.

src/shims/native_lib/mod.rs
RalfJung: Seems like the argument is "it's always `Ok`"? That is *not* a valid reason to ignore a result! If it's always `Ok`, then add an `unwrap` to indeed ensure that it is always `Ok`.

nia-e:
oops! normally I just mark everything as fixups and leave the top commit message, didn't realise `./miri squash` uses the actual squash function, sorry x.x

nia-e:
Good thing that unwrap was added, the logic was off by 1 >< this should fix it, I'll --fixup if it passes CI

RalfJung:
That's exactly why Result should not be ignored. ;)

Please update the comment as well.

nia-e:
should be good; ci seemed to pass, just squashed and updated the comment

Issue #450, opened 2018-09-05
Evaluate global constructors (life before main)


oli-obk:
@eddyb is lawful evil https://github.com/neon-bindings/neon/blame/master/src/lib.rs#L49

basically we should look in platform specific link sections for statics and run their value before anything else

```rust
#[cfg_attr(target_os = "linux", link_section = ".ctors")]
#[cfg_attr(target_os = "macos", link_section = "__DATA,__mod_init_func")]
#[cfg_attr(target_os = "windows", link_section = ".CRT$XCU")]
pub static __FOOMP: extern "C" fn() = {
    extern "C" fn __foomp() {
        // life before main goes here
    }
    __foomp
}
```

Ericson2314:
🤮 @eddyb!

RalfJung:
Are you sure about the "lawful" part? :P

More seriously though -- shouldn't we instead lobby for some way to express this in Rust proper? I already feel bad about that TLS emulation we are doing...

eddyb:
> shouldn't we instead lobby for some way to express this in Rust proper?

No, Rust as a language is uninterested in life-before-main. The only reason any of this works is because it's an OS feature we can't really stop you from opting into.

RalfJung:
See https://github.com/rust-lang/rust/issues/66862#issuecomment-559862805 for more information on these sections.

joshtriplett:
You might find the following C program useful for experimenting with init section order:

```c
#include <stdio.h>

typedef void (*FN)(void);

#define MKINIT(function, section) \
    void function(void) { puts(section); } \
    __attribute__((__section__(section))) FN function##_section[] = { function };

MKINIT(init_array, ".init_array")
MKINIT(init_array_00001, ".init_array.00001")
MKINIT(init_array_00002, ".init_array.00002")
MKINIT(init_array_65533, ".init_array.65533")
MKINIT(init_array_65534, ".init_array.65534")
MKINIT(fini_array, ".fini_array")
MKINIT(fini_array_00001, ".fini_array.00001")
MKINIT(fini_array_00002, ".fini_array.00002")
MKINIT(fini_array_65533, ".fini_array.65533")
MKINIT(fini_array_65534, ".fini_array.65534")
MKINIT(ctors, ".ctors")
MKINIT(ctors_00001, ".ctors.00001")
MKINIT(ctors_00002, ".ctors.00002")
MKINIT(ctors_65533, ".ctors.65533")
MKINIT(ctors_65534, ".ctors.65534")
[...]

PR #4459, opened 2025-07-10
Add support for global constructors (i.e. life before main)

ibraheemdev:
This should resolve https://github.com/rust-lang/miri/issues/450.

I verified that the `inventory` crate now works locally on my machine (Linux), but I'm not sure how this should be tested. Would adding `inventory` as a dependency for tests be acceptable, or should we setup the ctors manually?


RalfJung:
Thanks for the PR. :)

I would appreciate a test that does the setup itself, that's much easier for debugging future issues should they arise.

src/shims/ctor.rs
```diff
@@ -0,0 +1,87 @@
...
+        this: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, Poll<()>> {
+        use GlobalCtorStatePriv::*;
+        let new_state = 'new_state: {
+            match &mut self.0 {
+                Init => {
+                    let this = this.eval_context_mut();
+
+                    // Lookup constructors from the relevant magic link section.
+                    let link_section = match this.tcx.sess.target.os.as_ref() {
```
Noratrieb: there's a binary_format on target

tests/pass/shims/ctor.rs
```diff
@@ -0,0 +1,43 @@
...
+                target_os = "solaris",
+                target_os = "none",
+                target_family = "wasm",
+            )),
+            link_section = ".init_array"
+        )]
+        #[cfg_attr(windows, link_section = ".CRT$XCU")]
+        #[cfg_attr(
+            any(target_os = "macos", target_os = "ios"),
+            link_section = "__DATA,__mod_init_func"
```
bjorn3, line 33:
```suggestion
            link_section = "__DATA,__mod_init_func,mod_init_funcs"
```

Mach-O requires the `S_MOD_INIT_FUNC_POINTERS` section flag (`mod_init_funcs` in `link_section`). LLVM was accidentally adding this for `__DATA,__mod_init_func` making it work without explicitly adding `,mod_init_funcs`, but I don't know if it guarantees this. See also https://github.com/rust-lang/rustc_codegen_cranelift/issues/1588

src/shims/ctor.rs
```diff
@@ -0,0 +1,87 @@
...
+                Init => {
+                    let this = this.eval_context_mut();
+
+                    // Lookup constructors from the relevant magic link section.
+                    let link_section = match this.tcx.sess.target.binary_format {
+                        // Read the CRT library section on Windows.
+                        BinaryFormat::Coff => ".CRT$XCU",
+
+                        // Read `__mod_init_func` on macOS.
+                        BinaryFormat::MachO => "__DATA,__mod_init_func",
```
bjorn3: This also needs to be updated. Also strictly speaking it would be necessary to parse this into segment, section and section flags. For compat with rust-ctors and the like you may also need to keep accepting `__DATA,__mod_init_funcs` for now.

ibraheemdev: I guess this will break support for the ctor and inventory crates until they are updated?

ibraheemdev: Just saw the updated comment, I assume you mean keep accepting `__DATA,__mod_init_func` (without the `s`). Do we also need to add handling for `__DATA_CONST`?

tests/pass/shims/ctor.rs
ibraheemdev: From the [Mach-O reference](https://github.com/aidansteele/osx-abi-macho-file-format-reference), it looks like `mod_init_funcs` is not strictly required for the `S_MOD_INIT_FUNC_POINTERS` to be set. It's possible that this is outdated and retained for backwards compatibility.

> `S_MOD_INIT_FUNC_POINTERS`—This section contains pointers to module initialization functions.
The standard tools create `__DATA,__mod_init_func`sections of this type

src/shims/ctor.rs
bjorn3: > Just saw the updated comment, I assume you mean keep accepting __DATA,__mod_init_func (without the s).

Yes

> Do we also need to add handling for __DATA_CONST?

No clue.

src/eval.rs
```diff
@@ -215,27 +215,47 @@ impl Default for MiriConfig {
     }
 }
 
+struct MainThread<'tcx> {
```
RalfJung: The name is too generic, given that this is specifically the main-thread-stack-empty-handler.

Also, all this data is only needed in `MainThreadState::GlobalCtors` right? They become dead weight afterwards. Please structure the enum to make it clear which data is actually needed when.

src/eval.rs
```diff
@@ -289,6 +309,139 @@ impl<'tcx> MainThreadState<'tcx> {
         }
         interp_ok(Poll::Pending)
     }
+
+    fn eval_entry(&mut self, ecx: &mut MiriInterpCx<'tcx>) -> InterpResult<'tcx, ()> {
```
RalfJung: Please move all this code into a function on `MiriInterpCx`, right after `create_ecx`, so that the diff remains small.

Also, please call the function `call_main`. The function does not do the actual evaluation so `eval_entry` does not make a lot of sense as a name.

src/helpers.rs
```diff
@@ -1236,7 +1236,12 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     }
 
     /// Lookup an array of immediates stored as a linker section of name `name`.
```
RalfJung: The `name` argument does not exist any more so the comment needs updating. Also the comment still implies that this is for a *single* section but now it can actually be more than one.

src/helpers.rs
```diff
@@ -1236,7 +1236,12 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-    fn lookup_link_section(&mut self, name: &str) -> InterpResult<'tcx, Vec<ImmTy<'tcx>>> {
+    ///
+    /// Any sections to which the provided equality function returns `true` will be included.
```
RalfJung: That function will not be `eq` in many cases, so the name is confusing. Please call it `include_name`.

src/shims/ctor.rs
```diff
@@ -0,0 +1,100 @@
...
+                            this.lookup_link_section(|section| {
+                                let mut parts = section.splitn(3, ',');
+                                let (segment_name, section_name, section_type) =
+                                    (parts.next(), parts.next(), parts.next());
+
+                                segment_name == Some("__DATA")
+                                    && section_name == Some("__mod_init_func")
+                                    // The `mod_init_funcs` directive ensures that the `S_MOD_INIT_FUNC_POINTERS` flag
+                                    // is set on the section, but it is not strictly required.
+                                    && matches!(section_type, None | Some("mod_init_funcs"))
```
RalfJung: This can match two sections at the same time, right? And then we... run them in some arbitrary order?

src/shims/ctor.rs
```diff
@@ -0,0 +1,100 @@
...
+                            })?,
+
+                        // Read the standard `.init_array` section on platforms that use ELF, or WASM,
+                        // which supports the same linker directive.
+                        BinaryFormat::Elf | BinaryFormat::Wasm =>
+                            this.lookup_link_section(|section| section == ".init_array")?,
```
RalfJung: What about all the other section names mentioned in https://github.com/rust-lang/miri/issues/450#issuecomment-559868385 ? If we don't support them for now (which is fine for me), please add a FIXME.

src/shims/ctor.rs
```diff
@@ -0,0 +1,100 @@
...
+        };
+
+        self.0 = new_state;
+        interp_ok(Poll::Pending)
+    }
+}
+
+impl<'tcx> EvalContextPrivExt<'tcx> for crate::MiriInterpCx<'tcx> {}
+trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    fn schedule_ctor(&mut self, ctor: ImmTy<'tcx>) -> InterpResult<'tcx> {
```
RalfJung: This is not worth a function, please just inline it at its only call site.

src/eval.rs
```diff
@@ -215,27 +215,47 @@ impl Default for MiriConfig {
...
+                    }
+                }
+            }
             Running => {
-                *self = TlsDtors(Default::default());
+                // Inform GenMC that a thread has finished all user code. GenMC needs to know this for scheduling.
+                if let Some(genmc_ctx) = this.machine.data_race.as_genmc_ref() {
+                    let thread_id = this.active_thread();
+                    genmc_ctx.handle_thread_stack_empty(thread_id);
+                }
```
RalfJung: Previously the callback was called for *all* threads when their stack goes empty; now it only gets called on the main thread. That seems wrong.

Cc @Patrick-6 -- it's a bit unclear to me what the contract is here for when the callback needs calling. Already before this PR, a thread with destructors would call the callback more than once. What changes with this PR is that the main thread stack being empty does *not* imply that the main thread is done running "user code".

What does "user code" even mean here? TLS destructors are user code. So already before this PR there was user code running after this callback.

RalfJung (commented):
Thanks! I'm onboard with the overall architecture, but have a bunch of comments. :)

src/shims/ctor.rs
bjorn3: There is one section (`__DATA,__mod_init_func`) which may optionally have the `mod_init_funcs` flag set.

RalfJung: That's all part of the section name though, so in terms of what we see here there might be both `__DATA,__mod_init_func` and `__DATA,__mod_init_func,mod_init_funcs`.

bjorn3: Yes. They ordered the same as if you have two statics in the `__DATA,__mod_init_func` section. That is in some arbitrary order.

src/eval.rs
```diff
@@ -215,27 +215,47 @@ impl Default for MiriConfig {
     }
 }
 
+struct MainThread<'tcx> {
+    entry_id: DefId,
+    entry_type: MiriEntryFnType,
+    args: Vec<String>,
```
RalfJung: It should be enough to have just an `argc` and `argv`, both of type `ImmTy` here. You can do the setup of the arguments before calling the global ctors.

In fact you probably have to do that since the ctors may call shims that give access to the arguments and I assume that is supposed to work.

src/shims/ctor.rs
ibraheemdev: I believe `.ctors` and `.dtors` are legacy sections compared to `.init_array` and `.fini_array` (so `.fini_array` is for destructors, which is not relevant for this PR). I'm not sure about `.init_array.N` though, I added a comment for that.

src/eval.rs
ibraheemdev: I moved it back to its original location but yeah, I'm unsure if it's correct to call this after the constructors complete.

src/eval.rs
```diff
@@ -234,6 +240,15 @@ impl<'tcx> MainThreadState<'tcx> {
     ) -> InterpResult<'tcx, Poll<()>> {
         use MainThreadState::*;
         match self {
+            GlobalCtors { ctor_state, entry_id, entry_type, argc, argv } => {
+                match ctor_state.on_stack_empty(this)? {
+                    Poll::Pending => {} // just keep going
+                    Poll::Ready(()) => {
+                        call_main(this, *entry_id, *entry_type, argc.clone(), argv.clone())?;
```
ibraheemdev, line 249: Is cloning the `ImmTy`s fine here? I'm not sure if it's worth it to `Option` wrap them or do `replace_with` on the enum.

RalfJung:
Please avoid rebasing over new commits in master and do not force-push unless there are conflicts. Such rebases cause me a bunch of work since Github is pretty bad at dealing with force-pushes.

src/eval.rs
RalfJung: Yeah cloning them is fine, immediates are fairly cheap to clone.

src/eval.rs
```diff
@@ -395,11 +390,51 @@ pub fn create_ecx<'tcx>(
...
+
+    // Make sure we have MIR. We check MIR for some stable monomorphic function in libcore.
+    let sentinel =
+        helpers::try_resolve_path(tcx, &["core", "ascii", "escape_default"], Namespace::ValueNS);
+    if !matches!(sentinel, Some(s) if tcx.is_mir_available(s.def.def_id())) {
+        tcx.dcx().fatal(
+            "the current sysroot was built without `-Zalways-encode-mir`, or libcore seems missing. \
+            Use `cargo miri setup` to prepare a sysroot that is suitable for Miri."
+        );
+    }
```
RalfJung: Why did you move this code down?

src/shims/ctor.rs
```diff
@@ -0,0 +1,94 @@
...
+                                segment_name == Some("__DATA")
+                                    && section_name == Some("__mod_init_func")
+                                    // The `mod_init_funcs` directive ensures that the `S_MOD_INIT_FUNC_POINTERS` flag
+                                    // is set on the section, but it is not strictly required.
+                                    && matches!(section_type, None | Some("mod_init_funcs"))
+                            })?,
+
+                        // Read the standard `.init_array` section on platforms that use ELF, or WASM,
+                        // which supports the same linker directive.
+                        // FIXME: Add support for `.init_array.N`.
```
RalfJung:
```suggestion
                        // FIXME: Add support for `.init_array.N` and `.ctors`?
```
"Legacy" doesn't mean "nobody uses them so we can ignore them silently"...

tests/pass/shims/ctor.rs
RalfJung: The docs seem to say that S_MOD_INIT_FUNC_POINTERS indicates init functions, which sounds to me like the flag is required to designate a section of init functions?

src/eval.rs
ibraheemdev: I moved the `argc` and `argv` setup before it as you had suggested.

tests/pass/shims/ctor.rs
ibraheemdev: I believe the flag is implied on `__DATA,__mod_init_func` sections, but I'm not sure. LLVM seems to implement it that way and the ecosystem relies on it, so I think it makes sense for Miri to follow.

For this specific test I don't think it matters much whether we specify the flag given that we support it either way, but I can add it if you prefer.

src/eval.rs
ibraheemdev: Do you think this is a blocker from getting this merged, given that the current semantics are already unclear? Let me know if there's anything I can do to push this PR forward.

src/eval.rs
RalfJung: Ah, fair, I didn't realize this would require moving the `late_init`... I'll have to take a closer look at this then.

tests/pass/shims/ctor.rs
RalfJung: If it is questionable whether the version without the extra tag *should* work (as opposed to "happens to work due to undocumented LLVM accidents"), then at the very least we need a FIXME comment here clarifying this.

> the ecosystem relies on it

Do you have a link for that?

src/eval.rs
RalfJung: I'll discuss with Patrick.

tests/pass/shims/ctor.rs
ibraheemdev: The [ctor](https://docs.rs/ctor/latest/src/ctor/macros/mod.rs.html#533) and [inventory](https://docs.rs/inventory/latest/src/inventory/lib.rs.html#550) crates do not set it explicitly.

I can add a comment to the code. I'm not even sure where to find the official Mach-O reference, all I can find are outdated mirrors.

bjorn3: LLVM currently implies the flag for `__DATA,__mod_init_func` (https://github.com/llvm/llvm-project/blob/2e8e254d18f51b6ca898bf0b1e4d12109b5b16c7/llvm/lib/CodeGen/TargetLoweringObjectFileImpl.cpp#L1292-L1304), but I don't know if it actually guarantees this. The implication is because it creates a section with this name with the flag set somewhere and then merges the attributes of all definitions of the same section.

Edit: https://github.com/llvm/llvm-project/commit/cb307a27bf19ac76eeaf16ec6c8a4a698fad88d1 seems to have introduced this behavior as part of moving away from using the `.mod_init_func` assembly directive which implied this flag.

RalfJung: I have added a comment, that should suffice for now -- thanks for the links!

ibraheemdev:
Thanks for pushing this forward!

src/eval.rs
RalfJung: Given that genmc is not yet functional, we can postpone figuring out what the actual purpose of this callback is.

PR #4466, opened 2025-07-14
native-lib: allow passing structs as arguments

nia-e:
Blocked on #4456 currently to make rebasing easier when that lands. This allows passing arbitrary structs as arguments to native code; receiving a struct as a return value is not yet implemented. Also includes a couple of tests, though we may want to add more (e.g. ensuring it's properly detecting UB when a non-`#[repr(C)]` struct is passed).

I figured the review queue looked a bit too empty :D


tests/native-lib/fail/struct_not_extern_c.rs
```diff
@@ -0,0 +1,19 @@
...
+    pub other_value: i16,
+}
+
+extern "C" {
+    fn pass_struct(s: PassMe) -> i32;
+}
+
+fn main() {
+    let pass_me = PassMe { value: 42, other_value: 1337 };
+    unsafe { pass_struct(pass_me) }; //~ ERROR: Undefined Behavior: passing a non-#[repr(C)] struct over FFI
```
nia-e: Is this correct? I think it's UB, but maybe if the user intentionally sets `#[allow(improper_ctypes)]` we might just want to report this as unsupported instead of UB?

nia-e:
Think this is ready for review. I split off a bunch of the new code this introduces into a different file to not pollute `native_lib/mod.rs` too much more than it already is

@rustbot ready

src/shims/native_lib/mod.rs
```diff
@@ -276,6 +279,115 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                // This relies on the `expose_provenance` in the `visit_reachable_allocs` callback
+                // above.
+                CPrimitive::RawPtr(std::ptr::with_exposed_provenance_mut(s.bytes_usize())).into()
+            }
+            // For ADTs, create a libffi::middle::Type from their fields.
+            ty::Adt(adt_def, args) => {
+                let strukt = this.adt_to_carg(v.layout.ty, *adt_def, args)?;
+
+                // The raw bytes backing this arg.
+                let bytes = match v.as_mplace_or_imm() {
```
nia-e: Can a struct ever be an imm? From my testing it seems like it's always an mplace here but I'm not sure

nia-e:
Seems like there's some trouble here, nvm.

@rustbot author

src/shims/native_lib/mod.rs
RalfJung: Structs can be immediates, yes. `repr(C)` structs currently cannot, but that may change in the future.

nia-e:
Should be fixed. Also bumped the libffi version since apparently there was a [soundness issue](https://github.com/libffi-rs/libffi-rs/commit/23726dca1f6e9164894978bc71e7229eb3b43c2a) that got resolved since the previous version we were on; that wasn't the cause of my test failures, but it's probably smart to include it anyways. There's a couple `// TODO:`s, but those are either out of scope for this first PR (unions) or things I'm just uncertain of and need advice (the other 2).

@rustbot ready

src/shims/native_lib/ffi.rs
```diff
@@ -0,0 +1,113 @@
...
+    let mut arg_ptrs = vec![];
+    for arg in args {
+        arg_tys.push(arg.ty);
+        arg_ptrs.push(arg.ptr)
+    }
+    let cif = Cif::new(arg_tys, R::reify().into_middle());
+    unsafe { cif.call(fun, &arg_ptrs) }
+}
+
+/// A wrapper type for `libffi::middle::Type` which also holds a pointer to the data.
```
RalfJung: Can you explain what this means in a self-contained way, rather than referencing other concepts the reader is unlikely to know?

Also, it's odd to lead with this being a "type" when it is called "arg".

src/shims/native_lib/ffi.rs
```diff
@@ -0,0 +1,113 @@
...
+    _p: std::marker::PhantomData<&'a [u8]>,
+}
+
+impl<'a> FfiArg<'a> {
+    fn new(ty: FfiType, ptr: ArgPtr) -> Self {
+        Self { ty, ptr, _p: std::marker::PhantomData }
+    }
+}
+
+/// An owning form of `FfiArg`.
```
RalfJung: The prefixes "C" and "Ffi" don't do a great job reflecting "owned" and borrowed.

So what about `OwnedArg` and `BorrowedArg`? The "downcast" function (also an odd name, what does it downcast?) can then be called `borrow`.

`CPrimitive` could be called `ScalarArg` as that seems to match what we use it for.

Also, please generally define types bottom-up and with increasing complexity/subtlety, to simplify reading the file in order:
- ScalarArg
- OwnedArg
- BorrowedArg

src/shims/native_lib/ffi.rs
```diff
@@ -0,0 +1,113 @@
...
+/// We introduce this enum instead of just calling `Arg::new` and storing a list of
+/// `libffi::middle::Arg` directly, because the `libffi::middle::Arg` just wraps a reference to
+/// the value it represents and we need to store a copy of the value, and pass a reference to
+/// this copy to C instead.
+#[derive(Debug, Clone)]
+pub enum CArg {
+    /// Primitive type.
+    Primitive(CPrimitive),
+    /// Struct with its computed type layout and bytes.
+    Struct(FfiType, Box<[u8]>),
```
RalfJung: This doesn't have to be a struct, right? It can be any type that `FfiType` can represent, which includes basic integers.

tests/native-lib/fail/struct_not_extern_c.rs
RalfJung: Yeah, reporting this as "unsupported" is better.

tests/native-lib/pass/scalar_arguments.rs
```diff
@@ -39,5 +39,54 @@ fn main() {
...
         // test void function that prints from C
         printer();
+
+        test_pass_struct();
+        test_pass_struct_complex();
+    }
+}
+
+/// Test passing a basic struct as an argument.
+fn test_pass_struct() {
```
RalfJung: The file these tests are in is called "scalar_arguments". Do these arguments you are passing here ,ook like scalars to you? :)

Please make a new file for aggregate arguments.

tests/native-lib/scalar_arguments.c
```diff
@@ -30,6 +30,43 @@ EXPORT int64_t add_short_to_long(int16_t x, int64_t y) {
...
+
+typedef struct ComplexStruct {
+    Part1 part_1;
+    Part2 part_2;
+    uint32_t part_3;
+} ComplexStruct;
+
+EXPORT int32_t pass_struct_complex(const ComplexStruct complex) {
+  if ((((uint32_t)complex.part_1.high) << 16 | (uint32_t)complex.part_1.low) == complex.part_2.bits
+      && complex.part_2.bits == complex.part_3)
```
RalfJung: What is happening here...?

src/shims/native_lib/mod.rs
```diff
@@ -276,6 +278,147 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                        }
+                    }
+                    either::Either::Right(imm) => {
+                        // For immediates, we know the backing scalar is going to be 128 bits,
+                        // so we can just copy that.
+                        // TODO: is it possible for this to be a scalar pair?
+                        let scalar = imm.to_scalar();
+                        if scalar.size().bytes() > 0 {
+                            let bits = scalar.to_bits(scalar.size())?;
+                            bits.to_ne_bytes().to_vec().into_boxed_slice()
```
RalfJung: There's no way this is correct.^^ You're constructing a 16-byte slice here no matter the actual size of the type. And yes this can be a scalar pair. On a big-endian target, the data would even be at the wrong end of the buffer.

However, this should be dead code for `repr(C)` structs, so I'd say just emit "unsupported" here for now.

src/shims/native_lib/mod.rs
```diff
@@ -276,6 +278,147 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    either::Either::Left(mplace) => {
+                        // We do all of this to grab the bytes without actually
+                        // stripping provenance from them, since it'll later be
+                        // exposed recursively.
+                        let ptr = mplace.ptr();
+                        // Make sure the provenance of this allocation is exposed;
+                        // there must be one for this mplace to be valid at all.
+                        // The interpreter-level allocation itself is exposed in
+                        // visit_reachable_allocs.
+                        this.expose_provenance(ptr.provenance.unwrap())?;
```
RalfJung: This makes no sense to me, why would the provenance of the allocation holding the struct be exposed? That's just the memory we are copying from.

At the same time, this fails to properly expose provenance that a field might be pointing to. Please add a test involving a `repr(C)` struct that has a field of pointer type.

Getting the list of provenance you have to expose here is the least trivial remaining piece of work. You may even need to add a new API to rustc since I am not sure you can currently get that list from the allocation.

src/shims/native_lib/mod.rs
```diff
@@ -276,6 +278,147 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                            .alloc_id_from_addr(
+                                ptr.addr().bytes(),
+                                mplace.layout.size.bytes_usize().try_into().unwrap(),
+                                /* only_exposed_allocations */ true,
+                            )
+                            .unwrap();
+                        let ptr_raw = this.get_alloc_bytes_unchecked_raw(id)?;
+                        // SAFETY: We know for sure that at ptr_raw the next layout.size bytes
+                        // are part of this allocation and initialised. They might be marked as
+                        // uninit in Miri, but all bytes returned by the isolated allocator are
```
RalfJung: FWIW before https://github.com/rust-lang/miri/pull/4475 this argument was wrong on macOS where we didn't use the isolated allocator.

And in general there's no need to be so specific.
```suggestion
                        // uninit in Miri, but all bytes returned by `MiriAllocBytes` are
```

src/shims/native_lib/mod.rs
```diff
@@ -276,6 +278,147 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+    /// Parses an ADT to construct the matching libffi type.
+    fn adt_to_ffitype(
+        &self,
+        orig_ty: Ty<'_>,
+        adt_def: ty::AdtDef<'tcx>,
+        args: &'tcx ty::List<ty::GenericArg<'tcx>>,
+    ) -> InterpResult<'tcx, FfiType> {
+        // TODO: is this correct? Maybe `repr(transparent)` when the inner field
+        // is itself `repr(c)` is ok?
```
RalfJung: Yes, that would be okay. But let's leave it to a future PR, since a `repr(transparent)` wrapper around e.g. `i32` should also be supported and that'll be a bit more interesting -- it's going to be an `Immediate`.

src/shims/native_lib/mod.rs
```diff
@@ -276,6 +278,147 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        if !adt_def.repr().c() {
+            throw_ub_format!("passing a non-#[repr(C)] struct over FFI: {orig_ty}")
+        }
+        // TODO: unions, etc.
+        if !adt_def.is_struct() {
+            throw_unsup_format!("unsupported argument type for native call: {orig_ty}");
```
```suggestion
            throw_unsup_format!("unsupported argument type for native call: {orig_ty} is an enum or union");
```

src/shims/native_lib/mod.rs
```diff
@@ -276,6 +278,147 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
         interp_ok(())
     }
+
+    /// Extract the value from the result of reading an operand from the machine
+    /// and convert it to a `CArg`.
+    fn op_to_carg(&self, v: &OpTy<'tcx>, tracing: bool) -> InterpResult<'tcx, CArg> {
```
```suggestion
    fn op_to_ffi_arg(&self, v: &OpTy<'tcx>, tracing: bool) -> InterpResult<'tcx, CArg> {
```

src/shims/native_lib/mod.rs
```diff
@@ -276,6 +278,147 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                // Pointer without provenance may not access any memory anyway, skip.
+                if let Some(prov) = ptr.provenance {
+                    // The first time this happens, print a warning.
+                    if !this.machine.native_call_mem_warned.replace(true) {
+                        // Newly set, so first time we get here.
+                        this.emit_diagnostic(NonHaltingDiagnostic::NativeCallSharedMem { tracing });
+                    }
+
+                    this.expose_provenance(prov)?;
+                };
```
RalfJung: The same snippet will be needed in the struct case, so please make this a helper function.

RalfJung (commented):
Thanks! libffi is much cooler than I expected. :)

I have some ideas for how we could unify the scalar and struct handling... but I think that's better done in a follow-up PR, so I'll post some things on Zulip.

@rustbot author

tests/native-lib/pass/scalar_arguments.rs
```diff
@@ -39,5 +39,54 @@ fn main() {
...
+    extern "C" {
+        fn pass_struct_complex(s: ComplexStruct) -> i32;
+    }
+
+    let complex = ComplexStruct {
+        part_1: Part1 { high: 0xabcd, low: 0xef01 },
+        part_2: Part2 { bits: 0xabcdef01 },
+        part_3: 0xabcdef01,
+    };
+    assert_eq!(unsafe { pass_struct_complex(complex) }, 0);
```
RalfJung: Can you add a fail test that looks roughly like this, to ensure we do check initialization properly?
```rust
let arg = MaybeUninit::<ComplexStruct>::uninit();
pass_struct_complex(*arg.as_ptr());
```
This needs `derive(Copy, Clone)` on the struct. (Don't use `.read()`, that might defeat the test.)

tests/native-lib/scalar_arguments.c
nia-e: The test just ensures all 3 fields of the struct are passed correctly; in this case `part_1` is a 32-bit value split into 2 16-bit fields, and the other 2 are directly 32-bit values, and this pieces said value back together and ensures it was received correctly

src/shims/native_lib/mod.rs
nia-e: Hm, I figured doing this makes it so the `visit_reachable_allocs` call later exposes any provenances in the struct's fields no? Though I realise exposing the provenance for the struct itself is unnecessary

src/shims/native_lib/ffi.rs
nia-e: I'll change the name, presumably the next PR idea which you brought up on Zulip would make this enum redundant anyway

tests/native-lib/scalar_arguments.c
RalfJung: Instead of doing some hard-to-follow shifts and whatever here, just make the function take 3 more arguments (all scalars) that contain the expected value for the 3 fields.

src/shims/native_lib/mod.rs
RalfJung: Ah, I didn't understand the last sentence.

So yeah this will massively over-expose -- not just exposing the provenance of the struct itself is unnecessary, also consider that the struct might just span 10 bytes in a 1KB allocation and exposing all the other provenance in that allocation is equally unnecessary.

nia-e:
Apologies for the delay, IRL stuff came up, but this should be good now. I refactored the internals a little to do away with the need for a distinct borrowed/owned arg type, which should make the internal logic simpler, and added tests for a few more edge cases (notably the case of multiple structs in the same allocation; seems like I didn't need to change anything in rustc after all). Also rebased on main but all of the changes are in the last commit.

@rustbot ready

nia-e:
Hm, I rebased on master to make use of `get_range()` here as well and it seems like `repr(C)` structs *can* be immediates now (or at least, it causes tests to fail with that reason). Should I do the refactor you mentioned on Zulip in this branch? It's already mostly written locally

RalfJung:
> Hm, I rebased on master to make use of `get_range()` here as well and it seems like `repr(C)` structs _can_ be immediates now (or at least, it causes tests to fail with that reason). Should I do the refactor you mentioned on Zulip in this branch? It's already mostly written locally

Turns out `repr(C)` types cannot be `Scalar`, but they can be `ScalarPair`... so yeah we'll have to include that refactor in this PR I guess.

src/shims/native_lib/mod.rs
```diff
@@ -267,6 +269,157 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let bytes = match v.as_mplace_or_imm() {
+                    either::Either::Left(mplace) => {
+                        // Get the alloc id corresponding to this mplace, alongside
+                        // a pointer that's offset to point to this particular
+                        // mplace (not one at the base addr of the allocation).
+                        let mplace_ptr = mplace.ptr();
+                        let sz = mplace.layout.size.bytes_usize();
+                        let id = this
+                            .alloc_id_from_addr(
+                                mplace_ptr.addr().bytes(),
```
RalfJung: Here you are doing the equivalent of casting a pointer to an integer and back. Please don't. You can get the alloc ID from the pointer directly.

src/shims/native_lib/mod.rs
```diff
@@ -267,6 +269,157 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                        // SAFETY: We know for sure that at mplace_ptr.addr() the next layout.size
+                        // bytes are part of this allocation and initialised. They might be marked
+                        // as uninit in Miri, but all bytes returned by `MiriAllocBytes` are
+                        // initialised.
+                        unsafe {
+                            std::slice::from_raw_parts(
+                                alloc_ptr.with_addr(mplace_ptr.addr().bytes_usize()),
+                                mplace.layout.size.bytes_usize(),
+                            )
+                            .to_vec()
```
RalfJung: Round-tripping through a `Vec` seems odd?

src/shims/native_lib/mod.rs
RalfJung: Specifically, please use `ptr_get_alloc_id`. That also gives you the offset inside the allocation.

However, you'll need to first separately handle the zero-sized case. `mplace.ptr()` might not point to an `Allocation` if the type has size 0.

nia-e: Are ZSTs allowed over FFI? I was under the impression that there's no way to represent them in C so I figured just erroring here makes sense

src/shims/native_lib/mod.rs
nia-e: Fair, I guess it's pretty easily avoidable

nia-e:
Should be good. I couldn't use your suggestion of `write_target_uint` since that only takes a `u128` while a scalar pair could be up to 256 bits, so I had to adjust its logic to account for 2 scalars. Frankly that whole section could belong in a separate function to turn an immediate into bytes, but if one exists I couldn't find it.

I had also added some tests for that code (passing structs that were `a: i32, b: i16` and `a: i16, b: i32` so that there might be alignment padding between the fields) but removed them from here since `test_pass_struct` already handles the tricky case (second field with greater align than the first).

src/shims/native_lib/mod.rs
RalfJung: I guess erroring is fine but that still require a separate check to print a non-obscure error.

RalfJung:
> Should be good. I couldn't use your suggestion of write_target_uint since that only takes a u128 while a scalar pair could be up to 256 bits

You should be calling `write_target_uint` twice then, once for each scalar.

src/shims/native_lib/mod.rs
```diff
@@ -267,6 +270,189 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        // Now grab the bytes of the argument.
+        let bytes = match v.as_mplace_or_imm() {
+            either::Either::Left(mplace) => {
+                // Get the alloc id corresponding to this mplace, alongside
+                // a pointer that's offset to point to this particular
+                // mplace (not one at the base addr of the allocation).
+                let mplace_ptr = mplace.ptr();
+                let sz = mplace.layout.size.bytes_usize();
+                if sz == 0 {
+                    throw_unsup_format!("Attempting to pass a ZST over FFI");
```
```suggestion
                    throw_unsup_format!("attempting to pass a ZST over FFI");
```

src/shims/native_lib/mod.rs
```diff
@@ -267,6 +270,189 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                // initialised.
+                unsafe {
+                    Box::from(std::slice::from_raw_parts(
+                        alloc_ptr.add(ofs),
+                        mplace.layout.size.bytes_usize(),
+                    ))
+                }
+            }
+            either::Either::Right(imm) => {
+                let (first, maybe_second) = imm.to_scalar_and_meta();
```
RalfJung: `to_scalar_and_meta` is meant for wide pointers, you don't have a wide pointer here. You only need to handle `ScalarPair` since `Scalar` should be truly impossible. It seems worth a little helper that deals with one scalar so that you can call it twice.

src/shims/native_lib/mod.rs
```diff
@@ -267,6 +270,189 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                // TODO: Is this actually ok? Seems like the only way to figure
+                // out how the scalars are laid out relative to each other.
+                let align_second = match imm.layout.backend_repr {
+                    rustc_abi::BackendRepr::Scalar(_) => 1,
+                    rustc_abi::BackendRepr::ScalarPair(_, sc2) => sc2.align(this).bytes_usize(),
+                    _ => unreachable!(),
+                };
+                // How many bytes to skip between scalars if necessary for alignment.
+                let skip = sz_first.next_multiple_of(align_second).strict_sub(sz_first);
+
```
RalfJung: Have a look at https://github.com/rust-lang/rust/blob/ece1397e3ff40e7f8a411981844a8214659da970/compiler/rustc_const_eval/src/interpret/place.rs#L716-L732 for how to determine the offset of the 2nd field.

src/shims/native_lib/mod.rs
```diff
@@ -267,6 +270,189 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let mut bytes: Box<[u8]> = (0..sz).map(|_| 0u8).collect();
+
+                // Copy over the bytes in an endianness-agnostic way. Since each
+                // scalar may be up to 128 bits and write_target_uint doesn't
+                // give us an easy way to do multiple writes in a row, we
+                // adapt its logic for two consecutive writes.
+                let mut bytes_wr = bytes.as_mut();
+                match this.data_layout().endian {
+                    rustc_abi::Endian::Little => {
```
RalfJung: You should *not* be doing your own endianess handling. `write_target_uint` can do that for you.

src/shims/native_lib/mod.rs
nia-e: Sure, a `repr(C)` struct must be a scalar pair but this also needs to handle primitives, no? I can just match on the immediate directly instead of calling `to_scalar_and_meta()`, though

RalfJung: > but this also needs to handle primitives, no?

Ah, is this also the codepath for regular scalars now? :+1: to matching on the type of the immediate then.

src/shims/native_lib/mod.rs
```diff
@@ -267,6 +269,158 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    ))
+                }
+            }
+            either::Either::Right(imm) => {
+                // A little helper to write scalars to our byte array.
+                let write_scalar = |this: &MiriInterpCx<'tcx>, sc: Scalar, bytes: &mut [u8]| {
+                    // If a scalar is a pointer, then expose its provenance.
+                    if let interpret::Scalar::Ptr(p, _) = sc {
+                        // This relies on the `expose_provenance` in the `visit_reachable_allocs` callback
+                        // below to expose the actual interpreter-level allocation.
```
RalfJung: This comment is entirely in the wrong place now... previously it referred to a `std::ptr::with_exposed_provenance_mut`, now it stopped making any sense. Please ask before just taking random guesses as to where to put such comments.

nia-e: Is it wrong? If either scalar is a pointer, the code under the comment exposes provenance only in the AM; then later when `visit_reachable_allocs` is called it'll do `alloc.get_bytes_unchecked_raw().expose_provenance()` on the allocation said pointer points to. Unless I misunderstood the meaning of that comment, that's what it referred to in the first place

RalfJung:
I've done some refactoring and cleanup, found it easier to just edit the code than write it out.^^ Please take a look at the commit I added and use that to train your model for future Miri PRs. ;)

src/shims/native_lib/mod.rs
```diff
@@ -279,63 +279,67 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                 let alloc = this.get_alloc_raw(id)?;
-                let alloc_ptr = this.get_alloc_bytes_unchecked_raw(id)?;
-                for prov in alloc.provenance().get_range(this, (ofs..ofs.strict_add(sz)).into()) {
-                    this.expose_provenance(prov)?;
-                }
-                // SAFETY: We know for sure that at alloc_ptr + ofs the next layout.size
-                // bytes are part of this allocation and initialised. They might be marked
-                // as uninit in Miri, but all bytes returned by `MiriAllocBytes` are
-                // initialised.
-                unsafe {
```
RalfJung: In particular, I got rid of some unsafe code here.

src/shims/native_lib/mod.rs
RalfJung: It's correct that the code below only exposes the ptr in the AM. But the claim that this act somehow relies on the ptr also being exposed in the "host" makes no sense. That referred to the `with_exposed_provenance_mut`, which as per its documentation requires a prior call to `expose_provenance`, and the comment was explaining where that `expose_provenance` happened.

The new code side-steps the `with_exposed_provenance_mut`, making it less clear where to put the comment. But this here is definitely the wrong spot.

nia-e:
Ty! Glad this is good now ^^

PR #4469, opened 2025-07-15
Add std::hint::spin_loop() to tests

Patrick-6:
Some Miri tests have spin loops that weren't marked by `std::hint::spin_loop()`. Adding this hint makes these tests run slightly faster (on `x86_64`), likely because Miri was spinning in that loop until the thread got randomly preempted.

Some tests use `thread::yield` instead of the hint, these are unchanged, but some of them could likely be switched to the `spin_loop` hint too (e.g., [windows_join_multiple.rs](https://github.com/rust-lang/miri/blob/9b0962c547dc35820775f29c39c40378b89bbcae/tests/pass-dep/concurrency/windows_join_multiple.rs#L21)).

Tests that have a comment about not yielding are also unchanged (e.g., [concurrency/spin_loop.rs](https://github.com/rust-lang/miri/blob/9b0962c547dc35820775f29c39c40378b89bbcae/tests/pass/concurrency/spin_loop.rs#L8-L10)).

r? @RalfJung

### DIFF
--- a/tests/pass/0weak_memory_consistency.rs
+++ b/tests/pass/0weak_memory_consistency.rs
@@ -41,7 +41,7 @@ fn static_atomic_bool(val: bool) -> &'static AtomicBool {
 }
 
 /// Spins until it acquires a pre-determined value.
-fn loads_value(loc: &AtomicI32, ord: Ordering, val: i32) -> i32 {
+fn spin_until_i32(loc: &AtomicI32, ord: Ordering, val: i32) -> i32 {
     while loc.load(ord) != val {
         std::hint::spin_loop();
     }
@@ -49,7 +49,7 @@ fn loads_value(loc: &AtomicI32, ord: Ordering, val: i32) -> i32 {
 }
 
 /// Spins until it acquires a pre-determined boolean.
-fn loads_bool(loc: &AtomicBool, ord: Ordering, val: bool) -> bool {
+fn spin_until_bool(loc: &AtomicBool, ord: Ordering, val: bool) -> bool {
     while loc.load(ord) != val {
         std::hint::spin_loop();
     }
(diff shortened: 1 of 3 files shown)


tests/pass/0weak_memory_consistency.rs
```diff
@@ -48,6 +48,14 @@ fn loads_value(loc: &AtomicI32, ord: Ordering, val: i32) -> i32 {
     val
 }
 
+/// Spins until it acquires a pre-determined boolean.
+fn loads_bool(loc: &AtomicBool, ord: Ordering, val: bool) -> bool {
```
RalfJung: A name like `spin_until` seems more clear.

Patrick-6: - Should it still take the `val: bool` argument (always waiting until true would be enough for this test)?
- Renaming just for the boolean case, or function for waiting on an integer above?
  - The integer cases here uses `loads_value`, [this one here uses `reads_value`](https://github.com/Patrick-6/miri/blob/71c0b25cdb3f8a6dc0515e985f2fcc98ab0e02cb/tests/pass/weak_memory/weak.rs#L27), should I make them all the same?

RalfJung: Taking the `val` makes sense, yeah.

> The integer cases here uses loads_value, [this one here uses reads_value](https://github.com/Patrick-6/miri/blob/71c0b25cdb3f8a6dc0515e985f2fcc98ab0e02cb/tests/pass/weak_memory/weak.rs#L27), should I make them all the same?

I wasn't going to ask you to fix existing code, but sure, you can rename that as well.

Patrick-6: - `loads_value` and `spins_until`, or
- `spins_until_value` and `spins_until_bool`?

RalfJung: `spins_until` everywhere -- except if we need the bool and until version in the same file, then *both* should get the type name as a suffix.

PR #4470, opened 2025-07-16
allow using different cargo binary

nia-e:
Currently `./miri` and `miri-script` always assume that the correct cargo binary is just named `cargo`; this allows setting an env var to overwrite this. Useful for my setup, felt like a small enough change. The alternative was aliasing something to `cargo` instead but that can break things

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -348,6 +348,7 @@ https. Add the following to your `.gitconfig`:
 
 The following environment variables are relevant to `./miri`:
 
+* `CARGO` sets the binary used to execute Cargo; if none is specified, defaults to `cargo`.
 * `MIRI_AUTO_OPS` indicates whether the automatic execution of rustfmt, clippy and toolchain setup
   (as controlled by the `./auto-*` files) should be skipped. If it is set to `no`, they are skipped.
   This is used to allow automated IDE actions to avoid the auto ops.
--- a/miri
+++ b/miri
@@ -15,8 +15,9 @@ if [ -n "$MIRI_IN_RA" ]; then
   CARGO_FLAGS+=("--message-format=json" "-Zroot-dir=$ROOT_DIR")
   TARGET_DIR="$ROOT_DIR"/target
 fi
+
 # Run cargo.
-cargo $TOOLCHAIN build --manifest-path "$ROOT_DIR"/miri-script/Cargo.toml \
+${CARGO:-cargo} $TOOLCHAIN build --manifest-path "$ROOT_DIR"/miri-script/Cargo.toml \
   --target-dir "$TARGET_DIR" "${CARGO_FLAGS[@]}" || \
   ( echo "Failed to build miri-script. Is the 'stable' toolchain installed?"; exit 1 )
 # Instead of doing just `cargo run --manifest-path .. $@`, we invoke miri-script binary directly.
--- a/miri-script/src/util.rs
+++ b/miri-script/src/util.rs
@@ -38,6 +38,8 @@ pub struct MiriEnv {
     pub miri_dir: PathBuf,
     /// active_toolchain is passed as `+toolchain` argument to cargo/rustc invocations.
     toolchain: String,
+    /// The cargo binary to use.
+    cargo_bin: String,
     /// Extra flags to pass to cargo.
     cargo_extra_flags: Vec<String>,
     /// The rustc sysroot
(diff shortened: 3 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -83,6 +83,9 @@ will cause any `./miri` command to automatically call `./miri toolchain`, `clipp
 for you. If you don't want all of these to happen, you can add individual `.auto-toolchain`,
 `.auto-clippy` and `.auto-fmt` files respectively.
 
+The actual Cargo binary that `./miri` uses is by default `cargo`; set the environment variable
+`MIRI_CARGO_CMD` to override this.
```
RalfJung: Let's use the `CARGO` env var for this which is more standard. Also please move the docs down to "Further environment variables".

miri
```diff
@@ -15,8 +15,16 @@ if [ -n "$MIRI_IN_RA" ]; then
...
+# Set the binary used for cargo if the user specifies a different one.
+if [[ -z "${MIRI_CARGO_CMD}" ]]; then
+  CARGO_BIN="cargo"
+else
+  CARGO_BIN="${MIRI_CARGO_CMD}"
+fi
+
 # Run cargo.
-cargo $TOOLCHAIN build --manifest-path "$ROOT_DIR"/miri-script/Cargo.toml \
+$CARGO_BIN $TOOLCHAIN build --manifest-path "$ROOT_DIR"/miri-script/Cargo.toml \
```
RalfJung:
```suggestion
${CARGO:-cargo} $TOOLCHAIN build --manifest-path "$ROOT_DIR"/miri-script/Cargo.toml \
```
Then you can remove the if-then-else.

miri-script/src/util.rs
```diff
@@ -133,9 +133,10 @@ impl MiriEnv {
         let MiriEnv { toolchain, cargo_extra_flags, .. } = self;
         let manifest_path = path!(self.miri_dir / crate_dir.as_ref() / "Cargo.toml");
         let features = features_to_args(features);
+        let cargo_bin = std::env::var("MIRI_CARGO_CMD").unwrap_or("cargo".to_string());
         cmd!(
             self.sh,
-            "cargo +{toolchain} {cmd} {cargo_extra_flags...} --manifest-path {manifest_path} {features...}"
+            "{cargo_bin} +{toolchain} {cmd} {cargo_extra_flags...} --manifest-path {manifest_path} {features...}"
```
RalfJung, line 143: There's also `install_to_sysroot` which needs a similar patch.

miri-script/src/util.rs
```diff
@@ -133,9 +133,10 @@ impl MiriEnv {
         let MiriEnv { toolchain, cargo_extra_flags, .. } = self;
         let manifest_path = path!(self.miri_dir / crate_dir.as_ref() / "Cargo.toml");
         let features = features_to_args(features);
+        let cargo_bin = std::env::var("MIRI_CARGO_CMD").unwrap_or("cargo".to_string());
```
RalfJung: This seems worth storing inside `MiriEnv` rather than re-doing each time.

nia-e:
should be good now!

CONTRIBUTING.md
```diff
@@ -356,6 +356,7 @@ The following environment variables are relevant to `./miri`:
   tests. By default, the number of cores is used.
 * `MIRI_SKIP_UI_CHECKS` (recognized by `./miri test`) disables checking that the `stderr` or
   `stdout` files match the actual output.
+* `CARGO` sets the binary used to execute Cargo; if none is specified, defaults to `cargo`.
```
RalfJung: Please keep the list alphabetically sorted.

miri-script/src/util.rs
```diff
@@ -106,6 +108,9 @@ impl MiriEnv {
             sh.set_var("PATH", new_path);
         }
 
+        // Get the cargo binary to use, if one is set.
+        let cargo_bin = std::env::var("CARGO").unwrap_or("cargo".to_string());
```
RalfJung:
```suggestion
        let cargo_bin = std::env::var("CARGO").unwrap_or_else(|| "cargo".to_string());
```
Let's avoid allocating a string when we don't need to.

RalfJung:
(That squash also changed the base commit, making re-review harder. Please use `./miri squash`, or set `--keep-base` when squashing manually.)

PR #4471, opened 2025-07-16
triagebot welcome message for new users

RalfJung:
It seems we have to set the message twice, unfortunately.

Cc @Urgau

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@ for a list of Miri maintainers.
 
 [Rust Zulip]: https://rust-lang.zulipchat.com
 
-### Pull review process
+### PR review process
 
 When you get a review, please take care of the requested changes in new commits. Do not amend
 existing commits. Generally avoid force-pushing. The only time you should force push is when there
--- a/triagebot.toml
+++ b/triagebot.toml
@@ -17,6 +17,17 @@ allow-unauthenticated = [
 [assign]
 warn_non_default_branch = true
 contributing_url = "https://github.com/rust-lang/miri/blob/master/CONTRIBUTING.md"
+[assign.custom_welcome_messages]
+welcome-message = """
+Welcome, and thank you for contributing to Miri!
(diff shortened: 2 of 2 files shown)


oli-obk:
Do we even have the first message? We never auto assign reviewers

RalfJung:
I am hoping we can have a welcome message even if we don't auto-assign, but the docs don't make that fully clear so we may just have to try it.

Urgau:
> It seems we have to set the message twice, unfortunately.

Yeah, this is on purpose, we didn't wanted repos to have half custom/half default message.

> I am hoping we cam have a welcome message even if we don't auto-assign

Yes, contrary to the default auto-assign message, the custom ones (that you just set here) are pretty much always triggered (modulo if there is a manual assignment `r?` in the PR description).

> but the docs don't make that fully clear

Will make a note to improve them.

-----

Btw, I noticed that in addition to the custom messages, you also have the `contributing_url` option set, which wasn't used before since there wasn't any message but now that there is one it will be used (if you don't want it, just remove it from the config).

RalfJung:
> Yeah, this is on purpose, we didn't wanted repos to have half custom/half default message.

But one of them is dead weight for us because we are not using auto-assign, right?

> Btw, I noticed that in addition to the custom messages, you also have the contributing_url option set, which wasn't used before since there wasn't any message but now that there is one it will be used (if you don't want it, just remove it from the config).

Ah, thanks for the heads-up! I'll wait to see how it looks for real... but the message already links to the same document so yeah we should probably remove it.

RalfJung:
Btw, can that message use markdown?

Urgau:
> But one of them is dead weight for us because we are not using auto-assign, right?

Yeah, it is, I wonder if on triagebot side we should make the check a bit smarter by only requiring `welcome-message` (not the `-no-reviewer` variant) if auto-assign is enabled.

> Btw, can that message use markdown?

Yes, absolutely, triagebot forwards the content as-is to GitHub, so you use whatever GitHub supports.

PR #4472, opened 2025-07-16
triagebot: tweak welcome message

RalfJung:
based on https://github.com/rust-lang/miri/pull/4471#issuecomment-3079524562


RalfJung:
@Urgau is it expected that I get the message even though I do have prior contributions? :)

RalfJung:
Also, it doesn't seem to use the `contributing_url` after all?

Urgau:
> is it expected that I get the message even though I do have prior contributions? :)

Yes, despite being called "welcome" message it's also used to tell the assignment (none in you're case). That's why there is a `-no-reviewer` variant (for when we have no-one to assign).

> Also, it doesn't seem to use the `contributing_url` after all?

It's only shown for first timer contributor and you're not. (it's the first time we hit this case, we can change the rule if you want)

RalfJung:
Oh, so we can only customize the repeated message, not the first-time message? "welcome" is a confusing name then, yeah.^^


I'll remove the "welcome" part then, and make the wording more suited for being always repeated.

Urgau:
Yeah, it's not very well named, the feature grew out of the needs for the rustc-dev-guide (see https://github.com/rust-lang/rustc-dev-guide/pull/2502#issuecomment-3056202025) where they also don't use auto-assign. I will see about renaming it.

Would you prefer your message to only be shown for new comers? (aka be a true welcome message)

RalfJung:
I think reminding repeat contributors of this isn't bad, I adjusted the message accordingly in this PR.

Urgau:
I looked at what exactly the contributing url actually adds and it adds all these instructions. I don't know if they are all relevant for you. 

----------

Thank you for contributing to Miri!
Please remember to not force-push to the PR branch except when you need to rebase due to a conflict or when the reviewer asks you for it.

Please see [the contribution instructions](https://github.com/rust-lang/miri/blob/master/CONTRIBUTING.md) for more information. Namely, in order to ensure the minimum review times lag, PR authors and assigned reviewers should ensure that the review label (`S-waiting-on-review` and `S-waiting-on-author`) stays updated, invoking these commands when appropriate:

- `@rustbot author`: the review is finished, PR author should check the comments and take action accordingly
- `@rustbot review`: the author is ready for a review, this PR will be queued again in the reviewer's queue

RalfJung:
We use that process in Miri so yeah that seems fine.

Urgau:
Well `S-waiting-on-review` is *not* set automatically[^1], I could see that could be confusing since the message ask for it to be set accurately.

[^1]: if you would like it to be set automatically, add it to `[autolabel]` like [this](https://github.com/rust-lang/rust/blob/5795086bdfe7ed988aa53a110bd0692c33d8755b/triagebot.toml#L174-L175)

PR #4479, opened 2025-07-19
CI: add windows-arm runner

RalfJung:


RalfJung:
This doesn't work yet, rustup-toolchain-install-master currently cannot be built on windows-arm since it depends on winapi:
```
rustup-toolchain-install-master v1.8.1 (/home/r/src/rust/rustup-toolchain-install-master)
├── ansi_term v0.12.1
│   └── winapi v0.3.9
...
├── pbr v1.1.1
│   ├── crossbeam-channel v0.5.15
│   │   └── crossbeam-utils v0.8.21
│   ├── libc v0.2.174
│   └── winapi v0.3.9 (*)
```
Cc @kennytm 
Might be relevant for https://github.com/rust-lang/rfcs/pull/3817

RalfJung:
@kennytm judging from https://github.com/a8m/pb/issues/123 it seems like actually https://github.com/kennytm/rustup-toolchain-install-master/pull/61 is enough to fix this... any chance you could cut a release?

kennytm:
@RalfJung published 1.9.0.

RalfJung:
Okay, that seems to work. :)  But it is super slow.

cargo-miri/src/phases.rs
```diff
@@ -222,12 +222,12 @@ pub fn phase_cargo_miri(mut args: impl Iterator<Item = String>) {
...
     //
     // In `main`, we need the value of `RUSTC` to distinguish RUSTC_WRAPPER invocations from rustdoc
-    // or TARGET_RUNNER invocations, so we canonicalize it here to make it exceedingly unlikely that
+    // or TARGET_RUNNER invocations, so we make it absolute to make it exceedingly unlikely that
     // there would be a collision with other invocations of cargo-miri (as rustdoc or as runner). We
     // explicitly do this even if RUSTC_STAGE is set, since for these builds we do *not* want the
     // bootstrap `rustc` thing in our way! Instead, we have MIRI_HOST_SYSROOT to use for host
     // builds.
-    cmd.env("RUSTC", fs::canonicalize(find_miri()).unwrap());
+    cmd.env("RUSTC", path::absolute(find_miri()).unwrap());
```
RalfJung, line 230: There seems to be some behavior difference between x86-win32 and arm-win64 somewhere which lead to things breaking if we use `canonicalize` here. :shrug:

RalfJung:
After disabling GC_STRESS, the runtime is reasonable. So, the runner seems to be pretty slow, but that's a problem with all Windows runners.

Issue #4369, opened 2025-06-01
Miri should artifically shorten some reads/writes


RalfJung:
The issue in https://github.com/rust-lang/miri/issues/4367 would likely have been found a lot faster if Miri artificially shortened reads/writes sometimes. I'm thinking of a heuristic like, with 50% probability only try to read/write half as much as was requested. (But we probably need to still read/write at least 1 byte -- 0-byte reads have different semantics IIRC?)

Cc @tiif @Noratrieb

tiif:
If we have this feature, it'd be helpful for us to figure out which part of the test might fail spuriously. ~~But I haven't had any ideas on how to make the test better.~~ (I have an idea below)

This kind of failure also can also happen for [other writes in std](https://doc.rust-lang.org/std/io/trait.Write.html#tymethod.write). 

>If this method consumed n > 0 bytes of buf it must return Ok(n). If the return value is Ok(n) then n must satisfy n <= buf.len(). A return value of Ok(0) typically means that the underlying object is no longer able to accept bytes and will likely not be able to in the future as well, or that the buffer provided is empty.

EDIT: Maybe if we write n bytes, we should accept the value from 0 to n as the number of bytes written instead of only n?

RalfJung:
> This kind of failure also can also happen for [other writes in std](https://doc.rust-lang.org/std/io/trait.Write.html#tymethod.write).

What do you mean, "other writes"? That's the writes I am talking about.

In the end this is about the underlying libc functions `read`/`write` though, not about the public functions in std.

tiif:
I initially thought the failure can only happen in tokio's write.

But then I figured out the failure can also happen for other non-tokio write.

RalfJung:
@ChrisDenton @CraftSpider are short reads/writes also a thing on Windows, or do `NtReadFile` and `NtWriteFile` always return a "maximally" sized result?

ChrisDenton:
For normal filesystem files, they'll always return the maximum. For reading from devices such as pipes it can return shorter reads.

RalfJung:
Ah, fun. Well, we don't support pipes on Windows yet.

IIUC, Linux permits short reads even for regular files. That means when you run Miri with a Windows target on a Linux host, you can get incorrect behavior if the underlying system decides to cut some reads/writes short...

ChrisDenton:
[...]

PR #4481, opened 2025-07-19
Read write truncate

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/4369: with 50% change, we only read/write half the requested buffer. This quickly found some issues in our test suite.^^


src/shims/windows/fs.rs
```diff
@@ -462,6 +463,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         };
         let io_status_info = this.project_field_named(&io_status_block, "Information")?;
 
+        // It seems like short writes are not a thing on Windows, so we don't truncate `count` here.
+        // FIXME: if we are on a Unix host, short host writes host are visible to the program!
```
RalfJung: @CraftSpider is it correct that partial reads/writes are not a thing on Windows?
https://stackoverflow.com/questions/62438021/can-ntreadfile-produce-a-short-read-without-reaching-eof seems fairly clear here.

OTOH that also says that `io_status_info` would not be written on failure, whereas we always reset it to 0...

PR #4489, opened 2025-07-22
Miscellaneous fixes for tracing

Stypox:
This PR contains a few misc fixes for the tracing code that was recently added:
- The "id" saved to tracing chrome files is now cast from `u64` to `i64` so that it can always be loaded in Perfetto (otherwise it would sometimes be `>2^63-1` and Perfetto wouldn't be able to parse it properly, e.g. test these: [trace-ok.json](https://github.com/user-attachments/files/21371988/trace-ok.json), [trace-error.json](https://github.com/user-attachments/files/21371990/trace-error.json)).
- In https://github.com/rust-lang/miri/pull/4452 I forgot a `$` in the macro (sorry about this, I probably erased it by accident right before committing)
- I removed a duplicate "fatal_error: " being printed out when `MIRI_TRACING=1` is provided without `--features=tracing`, ref: https://github.com/rust-lang/miri/pull/4406#discussion_r2176537025

@rustbot ready

### DIFF
--- a/src/bin/log/setup.rs
+++ b/src/bin/log/setup.rs
@@ -60,7 +60,7 @@ fn init_logger_once(early_dcx: &EarlyDiagCtxt) {
             #[cfg(not(feature = "tracing"))]
             {
                 crate::fatal_error!(
-                    "fatal error: cannot enable MIRI_TRACING since Miri was not built with the \"tracing\" feature"
+                    "Cannot enable MIRI_TRACING since Miri was not built with the \"tracing\" feature"
                 );
             }
 


src/bin/log/tracing_chrome.rs
```diff
@@ -519,14 +519,16 @@ where
...
+    /// error if an id does not fit in an 64-bit signed integer in 2's complement.
+    fn get_root_id(&self, span: SpanRef<S>) -> Option<i64> {
         match self.trace_style {
             TraceStyle::Threaded => {
                 if span.fields().field("tracing_separate_thread").is_some() {
                     // assign an independent "id" to spans with argument "tracing_separate_thread",
                     // so they appear a separate trace line in trace visualization tools, see
                     // https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview#heading=h.jh64i9l3vwa1
-                    Some(span.id().into_u64())
+                    Some(span.id().into_u64() as i64)
```
RalfJung: This might wrap around -- isn't that a problem?

src/bin/log/tracing_chrome.rs
```diff
@@ -519,14 +519,16 @@ where
         }
     }
 
-    fn get_root_id(&self, span: SpanRef<S>) -> Option<u64> {
+    /// Returns [Option]`<i64>` instead of [Option]`<u64>` because apparently Perfetto gives an
```
RalfJung: What is up with the markdown here?
```suggestion
    /// Returns `Option<i64>` instead of `Option<u64>` because apparently Perfetto gives an
```

src/bin/log/tracing_chrome.rs
Stypox: No, that's by design, I tested and Perfetto accepts 64-bit 2s complement signed integers (i.e. from -2^63 to 2^63-1), so wrapping around is intended

Stypox:
I improved the markdown comment and also included info about why we cast `u64` to `i64` with wraparound

src/bin/log/tracing_chrome.rs
RalfJung: Then please call `cast_signed` instead, and ideally add a comment saying that perfetto is fine with negative numbers that for in an i64, but not fine with numbers above i64::MAX. Bare `as` casts are always sus.

src/bin/log/tracing_chrome.rs
```diff
@@ -519,14 +520,17 @@ where
         }
     }
 
-    fn get_root_id(&self, span: SpanRef<S>) -> Option<u64> {
+    /// Returns `Option<i64>` instead of `Option<u64>` because apparently Perfetto gives an
+    /// error if an id does not fit in a 64-bit signed integer in 2's complement. We cast the
+    /// span id from `u64` to `i64` with wraparound, since negative values are fine.
```
RalfJung: I don't think the reader has any reason to expect `u64` here... the comment is probably better placed near the cast.

Stypox: Done

src/bin/log/tracing_chrome.rs
RalfJung: > Then please call cast_signed instead

This is still unresolved

Stypox: Oh I didn't see the comment. Fixed now.

PR #4490, opened 2025-07-23
Use `josh-sync` instead of `miri-script` for Josh synchronization

Kobzol:
Discussed on Zulip. Removes local code, configures josh-sync, updates documentation and CI workflow.

Now that I think about it, I'm not sure if `--allow-noop` will be enough here. Because `rustc-josh-sync` will automatically rollup the commit and changes if it fails to pull anything, so checking for `HEAD^` probably won't do what we want. Maybe we can ignore the follow-up steps if `josh-sync` returns exit status 2? (in that case there's nothing to pull)


CONTRIBUTING.md
```diff
@@ -297,25 +297,27 @@ You can also directly run Miri on a Rust source file:
...
+We use the [`josh-sync`](https://github.com/rust-lang/josh-sync) tool to transmit changes between the
 rustc and Miri repositories. You can install it as follows:
 
 ```sh
-cargo +stable install josh-proxy --git https://github.com/josh-project/josh --tag r24.10.04
+cargo install --locked --git https://github.com/rust-lang/josh-sync
 ```
 
-Josh will automatically be started and stopped by `./miri`.
+[Josh](https://github.com/josh-project/josh) will automatically be started and stopped by the commands below.
```
RalfJung:
```suggestion
The commands below will automatically install and manage the [Josh](https://github.com/josh-project/josh) proxy that performs the actual work.
```

CONTRIBUTING.md
```diff
@@ -297,25 +297,27 @@ You can also directly run Miri on a Rust source file:
...
 
 ### Importing changes from the rustc repo
 
 *Note: this usually happens automatically, so these steps rarely have to be done by hand.*
 
 We assume we start on an up-to-date master branch in the Miri repo.
 
+1) First, create a branch for the pull, e.g. `git checkout -b rustc-pull`
```
RalfJung:
```suggestion
1) First, create a branch for the pull, e.g. `git checkout -b rustup`
```
Just because that's traditionally the branch name we use ;)

CONTRIBUTING.md
```diff
@@ -328,12 +330,12 @@ needed.
...
 ### Exporting changes to the rustc repo
 
-We will use the josh proxy to push to your fork of rustc. Run the following in the Miri repo,
+We will use the `josh-sync` tool to push to your fork of rustc. Run the following in the Miri repo,
 assuming we are on an up-to-date master branch:
 
 ```sh
 # Push the Miri changes to your rustc fork (substitute your github handle for YOUR_NAME).
-./miri rustc-push YOUR_NAME miri
+rustc-josh-sync push YOUR_NAME miri
```
RalfJung: The docs were outdated, the branch name is optional now.

It seems like josh-sync still requires it? I guess I can get used to that again.

RalfJung: Hm, this doesn't work:
```
$ RUSTC_GIT=../rustc rustc-josh-sync push RalfJung miri
Updating/installing josh-proxy binary...
+ cd "/home/r/src/rust/miri" && "cargo" "install" "--locked" "--git" "https://github.com/josh-project/josh" "--tag" "r24.10.04" "josh-proxy"
+ cd "/home/r/src/rust/miri" && "git" "status" "--untracked-files=no" "--porcelain"
josh up and running
Preparing https://github.com/miri/rust (base: 9748d87dc70a9a6725c5dbd76ce29d04752b4f90)...
+ cd "../rustc" && "git" "fetch" "https://github.com/miri/rust" "RalfJung"
Username for 'https://github.com': ^C
```
That command also makes no sense, `https://github.com/miri/rust` doesn't exist...

miri-script/src/commands.rs
```diff
@@ -99,66 +92,6 @@ impl Command {
         Ok(())
     }
 
-    fn start_josh() -> Result<impl Drop> {
-        // Determine cache directory.
-        let local_dir = {
-            let user_dirs =
-                directories::ProjectDirs::from("org", "rust-lang", "miri-josh").unwrap();
```
RalfJung, line 106: How does josh-sync manage the josh cache?

CONTRIBUTING.md
Kobzol: It does require it now, yeah. We could add a default branch name for pushes to `josh-sync.toml`, although I find it a bit weird, because the branch is pushed to a rustc fork of the person doing the push, so in a sense it seems like a "user-centric" property, rather than something that is shared across everyone doing miri synces (unless it's just a single person :laughing:).

miri-script/src/commands.rs
Kobzol: In exactly the [same way](https://github.com/rust-lang/josh-sync/blob/main/src/josh.rs#L26). The cache directory is called `rustc-josh`.

CONTRIBUTING.md
Kobzol: :facepalm: Sorry, it's `<branch-name> <gh-username>` for `rustc-josh-sync`. I kept the original order from `./miri push` and didn't realize it's reverted.

RalfJung: Ah, that's a confusing order.^^ Both the URL and the natural "coarse-grained to fine-grained" would put the username first...

miri-script/src/commands.rs
RalfJung: Okay so it's shared across all tools if I sync multiple different tools... that should be fine, they're also all using the same josh after all.

CONTRIBUTING.md
Kobzol: Yeah, fair enough. I could change it ofc, but that would require updating docs of a bunch of repos, so if you don't mind it too much, I'd just keep it as it is now.

RalfJung: I guess I'll be able to learn to live with it...

Having a default branch would also avoid that. ;)

RalfJung: It's true that that is more of a per-user thing, but  if the name is sufficiently clear, like `miri-sync`, it's unlikely to collide. I copied this idea of having a default from RA, if I remember correctly.

RalfJung:
I can't really test pulling since there's currently no changes, but pushing works fine and so surely pulling will also work.

Can you please clean up the git history? Then we can land this. :)  We can always adjust the docs later e.g. if default branches become a thing.

Kobzol:
> Now that I think about it, I'm not sure if --allow-noop will be enough here. Because rustc-josh-sync will automatically rollup the commit and changes if it fails to pull anything, so checking for HEAD^ probably won't do what we want. Maybe we can ignore the follow-up steps if josh-sync returns exit status 2? (in that case there's nothing to pull)

There is still this issue ^ I think that the current CI workflow won't work correctly, because when there's nothing to pull, rustc-josh-sync will rollback the created commits, so when the CI workflow compares diff against `HEAD^`, it will be looking at an unrelated previous commit.

I changed the workflow to compare with `$GITHUB_SHA` instead of with `HEAD^`. Let me know what you think.

.github/workflows/ci.yml
```diff
@@ -147,16 +147,16 @@ jobs:
...
+        run: cargo +stable install --locked --git https://github.com/rust-lang/josh-sync
       - name: setup bot git name and email
         run: |
           git config --global user.name 'The Miri Cronjob Bot'
           git config --global user.email 'miri@cron.bot'
       - name: Install nightly toolchain
         run: rustup toolchain install nightly --profile minimal
       - name: get changes from rustc
-        run: ./miri rustc-pull
+        run: rustc-josh-sync pull --allow-noop
```
RalfJung: Didn't you do some exit code things that we should check here to determine whether it's a NOP, and then just skip the rest of the steps?

Kobzol: Well, yes, by default the pull will exit with status code 2. But that also fails the whole job, which is not what we want. And if it exits with status code 0 (which is what `--allow-noop` does), the job will continue. There is AFAIK not an easy way to tell GH "stop this job, but also mark it as success". We would need to e.g. export some step output or an environment variable and then skip all the subsequent steps if it is set to true, which is what the default workflow does (https://github.com/rust-lang/josh-sync/blob/main/.github/workflows/rustc-pull.yml#L79). I didn't use the default workflow here though, because we also need to do the `miri fmt` postprocessing.

If you're fine with the approach of passing data between steps, I can replicate it here. But comparing with `GITHUB_SHA` seemed easier.

RalfJung: Those GITHUB_OUTPUT hacks don't look pleasant at all.

To me the cleanest solution seems to be to simply merge all these steps into one, so we can nicely express early-successful-exit.

Kobzol: I don't see it as a hack, it's a normal way to pass output between steps, even if a bit verbose. That being said, I was actually hoping that you would suggest combining everything into a single step :laughing: Done that in the latest commit: https://github.com/rust-lang/miri/pull/4490/commits/339dbf7effeeef6f4e8499eaef5fcf3eb64e2523

RalfJung: Right, I guess "excessively verbose" is more accurate than "hack". IMO Github didn't design this very well, but whatever.

.github/workflows/ci.yml
```diff
@@ -147,27 +147,40 @@ jobs:
...
       - name: Install nightly toolchain
         run: rustup toolchain install nightly --profile minimal
-      - name: get changes from rustc
-        run: ./miri rustc-pull
       - name: Install rustup-toolchain-install-master
         run: cargo install -f rustup-toolchain-install-master
-      - name: format changes (if any)
+      - name: Push changes to a branch and create PR
         run: |
+          # Temporarily disable early exit to examine the status code of rustc-josh-sync
```
RalfJung, line 164:
```suggestion
          # Make it easier to see what happens.
          set -x
          # Temporarily disable early exit to examine the status code of rustc-josh-sync.
```

.github/workflows/ci.yml
```diff
@@ -147,27 +147,40 @@ jobs:
...
+          set +e
+          rustc-josh-sync pull
+          exitcode=$?
+          set -e
+
+          # If there were no changes to pull, rustc-josh-sync returns status code 2
+          # In that case skip the rest of the job
```
RalfJung:
```suggestion
          # If there were no changes to pull, rustc-josh-sync returns status code 2.
          # In that case, skip the rest of the job.
```

.github/workflows/ci.yml
```diff
@@ -147,27 +147,40 @@ jobs:
...
+          if [ $exitcode -eq 2 ]; then
+            echo "Nothing changed in rustc, skipping PR"
+            exit 0
+          elif [ $exitcode -ne 0 ]; then
+            # If return code was not 0 or 2, rustc-josh-sync actually failed
+            echo "rustc-josh-sync failed"
```
RalfJung:
```suggestion
            echo "error: rustc-josh-sync failed"
```

.github/workflows/ci.yml
```diff
@@ -147,27 +147,40 @@ jobs:
...
-      - name: Push changes to a branch and create PR
-        run: |
-          # `git diff --exit-code` "succeeds" if the diff is empty.
-          if git diff --exit-code HEAD^; then echo "Nothing changed in rustc, skipping PR"; exit 0; fi
-          # The diff is non-empty, create a PR.
+
+          # Create a PR
           BRANCH="rustup-$(date -u +%Y-%m-%d)"
           git switch -c $BRANCH
           git push -u origin $BRANCH
```
RalfJung, line 188: While we are at it, can you remove the ZULIP env vars a few lines down from here? We don't need them any more since we no longer send a Zulip message in this job.

RalfJung (commented):
LGTM, just some comment nits :)

Kobzol:
Done.

RalfJung:
Awesome, thanks!
I guess we'll see soon whether the new cronjob actually works. ;)

PR #4498, opened 2025-07-27
Add support for building and linking against genmc

RalfJung:
This is https://github.com/rust-lang/miri/pull/4453 with some minor changes from my side.

### DIFF
--- a/src/concurrency/genmc/config.rs
+++ b/src/concurrency/genmc/config.rs
@@ -1,19 +1,41 @@
-use crate::MiriConfig;
+use super::GenmcParams;
 
+/// Configuration for GenMC mode.
+/// The `params` field is shared with the C++ side.
+/// The remaining options are kept on the Rust side.
 #[derive(Debug, Default, Clone)]
 pub struct GenmcConfig {
-    // TODO: add fields
+    pub(super) params: GenmcParams,
+    do_estimation: bool,
+    // FIXME(GenMC): add remaining options.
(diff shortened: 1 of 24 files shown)


Patrick-6:
LGTM, we just need to make sure to add the platform checks once the actual functionality is added.

RalfJung:
Which platform checks? `src/concurrency/genmc/config.rs` only gets built when genmc is supported so it makes no sense to check the platform inside of it.

src/concurrency/genmc/config.rs
```diff
@@ -17,23 +17,17 @@ impl GenmcConfig {
...
-        if !cfg!(all(
-            feature = "genmc",
-            any(target_os = "linux", target_os = "macos"),
-            target_pointer_width = "64",
-            target_endian = "little"
-        )) {
-            unimplemented!(
-                "GenMC mode is not supported on this platform, cannot handle argument: \"-Zmiri-genmc{trimmed_arg}\""
-            );
-        }
```
Patrick-6: These checks. Maybe we could put them [here in miri.rs](https://github.com/rust-lang/miri/blob/9fb3bbb646c5ce491f7b6dcc9744c276e686d806/src/bin/miri.rs#L737)?

Patrick-6: Ah, no, the dummy version will be used on unsupported platforms, it is ok then.

PR #4523, opened 2025-08-11
Add tracing to data race functions

Stypox:
This PR adds tracing calls to the `Vclocks` data race checks. 

`data_race` seem to amount to around 5% of the program time, although given the sheer amount of very short-lived tracing calls I think most of that 5% is spent within the tracing infrastructure. The trace file for analysis below was obtained by running the multithreaded program at the bottom with `n=100` and `thread_count=5`; let me know if there is a better multithreaded example I could try where data race checks are more prominent.

The following table was obtained with the SQL query below in ui.perfetto.dev, on the trace file mentioned above.

```sql
select "TOTAL PROGRAM DURATION" as name, count(*), max(ts + dur) as "sum(dur)", 100.0 as "%", null as "min(dur)", null as "max(dur)", null as "avg(dur)", null as "stddev(dur)" from slices union select "TOTAL OVER ALL SPANS (excluding events)" as name, count(*), sum(dur), cast(cast(sum(dur) as float) / (select max(ts + dur) from slices) * 1000 as int) / 10.0 as "%", min(dur), max(dur), cast(avg(dur) as int) as "avg(dur)", cast(sqrt(avg(dur*dur)-avg(dur)*avg(dur)) as int) as "stddev(dur)" from slices where parent_id is null and name != "frame" and name != "step" and dur > 0 union select name, count(*), sum(dur), cast(cast(sum(dur) as float) / (select max(ts + dur) from slices) * 1000 as int) / 10.0 as "%", min(dur), max(dur), cast(avg(dur) as int) as "avg(dur)", cast(sqrt(avg(dur*dur)-avg(dur)*avg(dur)) as int) as "stddev(dur)" from slices where parent_id is null and name != "frame" and name != "step" group by name order by sum(dur) desc, count(*) desc
```

image

This is the partitioning of calls among the various `data_race` spans being traced (even nested ones are included here, that's why the count(*) don't add up to 498874. It was obtained with the SQL query below on the same trace file as before.

```sql
select args.string_value as name, count(*), sum(dur), min(dur), max(dur), cast(avg(dur) as int) as "avg(dur)", cast(sqrt(avg(dur*dur)-avg(dur)*avg(dur)) as int) as "stddev(dur)" from slices inner join args USING (arg_set_id) where args.key = "args." || slices.name and name = "data_race" group by args.string_value order by count(*) desc
```

image

<details>
<summary>Rust code that was used as a benchmark to pass to Miri to collect traces</summary>

```rs
fn main() {
    let n: usize = std::env::args().nth(1).unwrap().parse().unwrap();
    let thread_count: usize = std::env::args().nth(2).unwrap().parse().unwrap();

    let threads = (0..thread_count).map(|_| std::thread::spawn(move || {
        let mut v = (0..n).into_iter().collect::<Vec<_>>();
        for i in &mut v {
            *i += 1;
        }
    })).collect::<Vec<_>>();

    for thread in threads {
        thread.join().unwrap();
    }
}
```

</details>

### DIFF
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -1394,6 +1394,7 @@ impl<'tcx> Machine<'tcx> for MiriMachine<'tcx> {
             GlobalDataRaceHandler::Genmc(genmc_ctx) =>
                 genmc_ctx.memory_load(machine, ptr.addr(), range.size)?,
             GlobalDataRaceHandler::Vclocks(_data_race) => {
+                let _trace = enter_trace_span!(data_race::before_memory_read);
                 let AllocDataRaceHandler::Vclocks(data_race, weak_memory) = &alloc_extra.data_race
                 else {
                     unreachable!();
@@ -1429,6 +1430,7 @@ impl<'tcx> Machine<'tcx> for MiriMachine<'tcx> {
                 genmc_ctx.memory_store(machine, ptr.addr(), range.size)?;
             }
             GlobalDataRaceHandler::Vclocks(_global_state) => {
+                let _trace = enter_trace_span!(data_race::before_memory_write);
                 let AllocDataRaceHandler::Vclocks(data_race, weak_memory) =
                     &mut alloc_extra.data_race
                 else {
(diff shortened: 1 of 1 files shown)


RalfJung:
LGTM, thanks!

PR #4524, opened 2025-08-12
Account for time spent tracing, use RDTSC for faster time

Stypox:
This PR does two things to drastically improve the tracing experience by removing some biases in timestamps caused by overheads in the tracing machinery:
1. It measures how much time is spent inside the `tracing_chrome` machinery and subtracts it from all subsequent timestamps saved to the trace file. This is not very elegant, but does the job quite well. The `tracing_chrome` machinery is inherently slow as it has to format traced arguments (e.g. call their Debug implementation), and this obviously can't be offloaded to a background thread, since it would involve passing references to data to another thread which would break memory safety.
2. It uses `rdtsc` to measure time on Linux x86/x86_64, which only takes a few nanoseconds and returns nanosecond-precise timestamps. `std::time::Instant` on the other hand takes `~1.3us` which completely biases the length of spans shorter than a few microseconds. See at the bottom for the options considered for time measurements.

The following table was obtained by interpreting the Rust script at the bottom in Miri. "Execution time (cpu)" is obtained with Linux's `time` (and is a sum of the time spent on the Miri thread and on the tracing thread), and "Max time in trace file" is obtained by looking at the `ts` value of the last entry of the produced trace file. The sensitivity of the values in the table above seems to be ~10%. We can see that 1. (which measures time twice instead of once per trace event) adds no overhead if 2. is also employed. Using both 1. and 2. reduces the real runtime from 1.7s to 1.4s and especially reduces the time observed in trace files from 1.7s to 0.3s (though the tracing overhead still accounts for $1-91ms/273ms \approx 67\\%$ of the time seen in the traces).   

<table>
<tr>
 <th>
 <th>Execution time (cpu)
 <th>Max time in trace file
<tr>
 <td>Tracing disabled
 <td>91ms
 <td>-
<tr>
 <td>Before this PR
 <td>1678ms
 <td>1650ms
<tr>
 <td>With just 1.
 <td>3599ms
 <td>1511ms
<tr>
 <td>With just 2.
 <td>1424ms
 <td>915ms
<tr>
 <td>With 1. and 2.
 <td>1383ms
 <td>273ms
</table>

<details>
<summary>Rust script used for benchmarking with `n=100`</summary>

```rs
fn main() {
    let n: usize = std::env::args().nth(1).unwrap().parse().unwrap();
    let mut v = (0..n).into_iter().collect::<Vec<_>>();
    for i in &mut v {
        *i += 1;
    }
}
```

</details>

<details>
<summary>Options considered</summary>

I was spending some time trying to figure out in which other places time is spent during the execution of Miri, and by using `perf` I noticed that a tangible amount of time is spent in code under the `concurrency/` subfolder. The functions there are called very frequently and are all quite fast (they take just a few hundreds nanoseconds to execute), and when I added tracing calls to them the execution time of Miri nearly doubled. It turns out that tracing calls **on Linux** take a few microseconds each, and they spend most of the time just for measuring time. Here are the options we have:
1. The current implementation uses `std::time::Instant`, which internally uses `clock_gettime(CLOCK_MONOTONIC)` at least on Linux (see [here](https://doc.rust-lang.org/src/std/sys/pal/unix/time.rs.html#280) and [here](https://doc.rust-lang.org/src/std/sys/pal/unix/time.rs.html#124)). This system call has a `~1.3µs` latency on my PC, and has nanosecond precision. This is not ideal for very short spans, but is ok-ish for spans of at least a few microseconds.
2. I considered `clock_gettime(CLOCK_MONOTONIC_COARSE)`. The system call takes as little as `~5ns` on my pc, but the clock's precision is a few milliseconds. Using this clock makes traces useless, because events happen much quicker than every few milliseconds, as shown [here](https://www.pingcap.com/blog/how-we-trace-a-kv-database-with-less-than-5-percent-performance-impact/) in section "Coarse time".
3. A third option exists, which is both fast and precise, but has portability and stability issues. x86's `rdtsc` instruction reads the TSC counter in the CPU, which was born to continuously count the clock cycles, and was later adapted to just count at a constant rate (due to CPU frequency scaling). This counter only exists on x86/x86_64 platforms, and suffers from stability issues:  every core has its own TSC and they may be out-of-sync, so time might appear to go backwards if the scheduler decides to move Miri's thread from one core to another. To avoid this, Miri's thread can be forced to execute on just one core with `sched_setaffinity`. The implementation gets slightly more complicated if Miri uses more than one thread (e.g. with `-Zmiri-many-seeds`) but is still doable (and also I guess it is not so useful to trace Miri with `-Zmiri-many-seeds` anyway).  Another issue is that some older CPUs have TSC counters that count with a varying frequency (due to frequency scaling) and are thus not usable to measure time, and this can be detected by reading a system file (read more [here](https://www.pingcap.com/blog/how-we-trace-a-kv-database-with-less-than-5-percent-performance-impact) under "TSC"). `rdtsc` has nanosecond latency and nanosecond precision which would be perfect for tracing.

Note that all Apple devices have [a clock that runs at 24MHz](https://stackoverflow.com/a/15125418) that can be read with `mach_absolute_time` (and `std::time::Instant` already uses it), so the latency there is 40ns. Linux on ARM also seems to have a pretty fast time implementation backing `std::time::Instant`, because at least on my Android phone `std::time::Instant` has a latency of `~250ns`. The Windows system call is `QueryPerformanceCounter`, which should internally use TSC and hence be fast (see [here](https://learn.microsoft.com/en-us/windows/win32/dxtecharts/game-timing-and-multicore-processors), recommendation 1.). Also see [this](https://stackoverflow.com/q/55583503) for the system calls used by `std::time::Instant` on the various platforms.  Therefore the above problem only applies to Linux on x86/x86_64.

What would you suggest doing here? I would continue using `std::time::Instant` everywhere except for x86/x86_64 Linux devices that support constant-rate TSC, and instead use TSC there despite all of its caveats. This change can be made inside `tracing_chrome.rs`'s `get_ts` function.

</details>


saethlin:
> The following table was obtained by interpreting the Rust script at the bottom in Miri.

FYI we have a benchmark suite that tries to tickle some pathological cases in the interpreter. What you've shared clearly demonstrates the value of this PR, but the benchmark suite programs may also be educational.

src/bin/log/tracing_chrome.rs
```diff
@@ -50,9 +53,16 @@ use std::{
     thread::JoinHandle,
 };
 
+use crate::log::tracing_chrome_instant::TracingChromeInstant;
+
+struct ThreadData {
+    tid: usize,
+    out: Sender<Message>,
+    start: TracingChromeInstant,
+}
```
RalfJung, line 68: Please add comments explaining what this state does.

src/bin/log/tracing_chrome.rs
```diff
@@ -548,31 +532,55 @@ where
...
+    /// inside the trace file (i.e. `ts`), even if such functions are slow (e.g. because they need
+    /// to format arguments on the same thread those arguments are collected on, otherwise memory
+    /// safety would be broken).
+    ///
+    /// `f` is called with the microseconds elapsed since the current thread was started (**not**
+    /// since the program start!), with the current thread ID (i.e. `tid`), and with a [Sender] that
+    /// can be used to send a [Message] to the thread that collects [Message]s and saves them to the
+    /// trace file.
+    #[inline(always)]
+    fn with_elapsed_micros_subtracting_tracing<T: Fn(f64, usize, &Sender<Message>)>(&self, f: T) {
```
RalfJung: Please add some comments inside this function, explaining what happens and pointing out anything subtle.

```suggestion
    fn with_elapsed_micros_subtracting_tracing(&self, f: impl Fn(f64, usize, &Sender<Message>)) {
```

src/bin/log/tracing_chrome.rs
```diff
@@ -548,31 +532,55 @@ where
         }
     }
 
-    fn enter_span(&self, span: SpanRef<S>, ts: f64) {
-        let callsite = self.get_callsite(EventOrSpan::Span(&span));
+    fn enter_span(&self, span: SpanRef<S>, ts: f64, tid: usize, out: &Sender<Message>) {
+        let callsite = self.get_callsite(EventOrSpan::Span(&span), tid);
         let root_id = self.get_root_id(span);
-        self.send_message(Message::Enter(ts, callsite, root_id));
+        let _ignored = out.send(Message::Enter(ts, callsite, root_id));
```
RalfJung, line 544: What are these things being ignored everywhere?

src/bin/log/tracing_chrome_instant.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+impl TracingChromeInstant {
+    /// Can be thought of as the same as [std::time::Instant::now()], but also does some setup to
+    /// make TSC stable in case TSC is available. This is supposed to be called (at most) once per
+    /// thread since the thread setup takes a few milliseconds.
+    ///
+    /// WARNING: If TSC is available, `incremental_thread_id` is used to pick to which CPU to pin
+    /// the current thread. It should be an incremental number that starts from 0. Be aware that
+    /// the current thread will be restricted to one CPU for the rest of the execution!
```
RalfJung: What is an "incremental" number?

src/bin/log/tracing_chrome_instant.rs
```diff
@@ -0,0 +1,170 @@
...
+        #[cfg(target_arch = "x86_64")]
+        use core::arch::x86_64::{_mm_lfence, _rdtsc};
+        use core::sync::atomic::{Ordering, compiler_fence};
+
+        unsafe {
+            _mm_lfence();
+            compiler_fence(Ordering::SeqCst);
+            let tsc = _rdtsc();
+            compiler_fence(Ordering::SeqCst);
+            _mm_lfence();
```
RalfJung: What's the point of these fences...?

src/bin/log/tracing_chrome_instant.rs
```diff
@@ -0,0 +1,170 @@
...
+/// This measures time using [std::time::Instant], except for x86/x86_64 Linux machines, where
+/// [std::time::Instant] is too slow (~1.5us) and thus `rdtsc` is used instead (~5ns).
+pub enum TracingChromeInstant {
+    WallTime {
+        start_instant: std::time::Instant,
+    },
+    #[cfg(all(target_os = "linux", any(target_arch = "x86", target_arch = "x86_64")))]
+    Tsc {
+        start_tsc: u64,
+        tsc_to_microseconds: f64,
```
RalfJung, line 37: Please add a comment explaining the meaning (and unit) of this field.

src/bin/log/tracing_chrome_instant.rs
```diff
@@ -0,0 +1,170 @@
...
+
+        let mut buf = String::with_capacity(1024);
+        while cpuinfo.read_line(&mut buf).ok()? > 0 {
+            if buf.starts_with("flags") {
+                return Some(buf.contains("constant_tsc"));
+            }
+            buf.clear();
+        }
+        None // EOF
+    }
```
RalfJung: Is that really the recommended way to deal with TSC? *Measure* the frequency?!? And parse some text file in `/proc` to check whether it is present...?

src/bin/log/tracing_chrome_instant.rs
```diff
@@ -0,0 +1,170 @@
...
+    pub(super) fn is_tsc_available() -> Option<bool> {
+        use std::io::{BufRead, BufReader};
+
+        let cpuinfo = std::fs::File::open("/proc/cpuinfo").ok()?;
+        let mut cpuinfo = BufReader::new(cpuinfo);
+
+        let mut buf = String::with_capacity(1024);
+        while cpuinfo.read_line(&mut buf).ok()? > 0 {
+            if buf.starts_with("flags") {
+                return Some(buf.contains("constant_tsc"));
```
RalfJung: This will also fire if there is some other flag that contains `constant_tsc` as a substring... doesn't seem ideal?

src/bin/log/tracing_chrome_instant.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    /// Forces the current thread to run on a single CPU, which ensures the TSC counter is monotonic
+    /// (since TSCs of different CPUs might be out-of-sync). `incremental_thread_id` is used to pick
+    /// to which CPU to pin the current thread, and should be an incremental number that starts from
+    /// 0.
+    pub(super) fn set_cpu_affinity(incremental_thread_id: usize) {
+        let cpu_id = match std::thread::available_parallelism() {
+            Ok(available_parallelism) => incremental_thread_id % available_parallelism,
+            _ => {
+                eprintln!("Could not determine CPU count to properly set CPU affinity");
```
RalfJung: Should this be a panic? As-is the error will look pretty random and it's not even clear that it comes from Miri (rather than the interpreted program).

src/bin/log/tracing_chrome_instant.rs
```diff
@@ -0,0 +1,170 @@
...
+    /// spent executing other `f`s passed to this function. This behavior allows subtracting time
+    /// spent in functions that log tracing data (which `f` is supposed to be) from the tracing time
+    /// measurements.
+    #[inline(always)]
+    pub fn with_elapsed_micros_subtracting_tracing<T: Fn(f64)>(&mut self, f: T) {
+        match self {
+            TracingChromeInstant::WallTime { start_instant } => {
+                let instant_before_f = std::time::Instant::now();
+                let ts = (instant_before_f - *start_instant).as_nanos() as f64 / 1000.0;
+                f(ts);
```
RalfJung: It seems like we usually use nanoseconds in an `f64` in these APIs. Why are you using microseconds?

src/bin/log/tracing_chrome_instant.rs
```diff
@@ -0,0 +1,170 @@
...
+                *start_instant += std::time::Instant::now() - instant_before_f;
```
RalfJung, line 96: Please explain this logic, it's not entirely obvious.

src/bin/log/tracing_chrome_instant.rs
Stypox: I couldn't find a "recommended" way to deal with TSC, as documentation is quite scarce, but in the various sources I used as a reference they obtained the frequency by measuring it:
- https://github.com/fast/fastant/blob/27c9ec5ec90b5b67113a748a4defee0d2519518c/src/tsc_now.rs#L234
- https://gist.github.com/pmttavara/6f06fc5c7679c07375483b06bb77430c
- https://github.com/sheroz/tick_counter/blob/main/src/lib.rs#L184

I also found [this article](https://blog.trailofbits.com/2019/10/03/tsc-frequency-for-all-better-profiling-and-benchmarking/) explaining how to install a kernel driver that publishes the TSC frequency from the kernel, so I guess there is no way to obtain such frequency without having access to the kernel.

I also don't like this approach at all but it should calculate the right frequency quite reliably.

Stypox: > And parse some text file in /proc to check whether it is present...?

That's also done [here](https://github.com/fast/fastant/blob/27c9ec5ec90b5b67113a748a4defee0d2519518c/src/tsc_now.rs#L194)

src/bin/log/tracing_chrome_instant.rs
Stypox: `tracing_chrome` was using microseconds before, too, since that's the unit saved in the .json trace file

src/bin/log/tracing_chrome_instant.rs
Stypox: They ensure that instructions are not reordered around `rdtsc()`  by the compiler or by the CPU, otherwise it might not measure the time at the expected point in the code

src/bin/log/tracing_chrome.rs
Stypox: They say whether `mpsc::Sender::send()` had success or not (it might not have success if the receiver was dropped). They were being ignored before, too, so I kept the same behavior. I guess an alternative would be to just call `.unwrap()`

src/bin/log/tracing_chrome_instant.rs
Stypox: Not being able to determine the number of CPUs can only cause issues when `-Zmany-seeds` is greater than the number of cores on the device, because in that case `incremental_thread_id` would at some point be greater than the number of cores, and setting the CPU affinity would not work. That's why there is a panic just below that handles the case where `libc::sched_setaffinity` does not succeed. So even when this message is shown it might be unlikely that there is an actual problem, but it's still worth printing it out so in case the `panic!()` below fires the user would have a clue why.

src/bin/log/tracing_chrome_instant.rs
Stypox: What I mean is that for every new thread that enters the tracing infrastructure, the counter should increase by 1, so if there are 5 threads then their ids are 0 1 2 3 4.

src/bin/log/tracing_chrome_instant.rs
RalfJung: Please read the compiler_fence docs. I don't think that fence makes any sense here.

What you probably want is an empty inline asm block. These fences have concurrency semantics which isn't strong enough here.

src/bin/log/tracing_chrome.rs
RalfJung: Ah okay so this is unrelated to the PR then? In that case it's fine, I was just confused.

src/bin/log/tracing_chrome_instant.rs
RalfJung: The way it is printed is quite confusing though.

For all similar situations I think Miri just unwraps, so IMO that also makes most sense here.

src/bin/log/tracing_chrome_instant.rs
RalfJung:
```suggestion
    /// WARNING: If TSC is available, `incremental_thread_id` is used to pick to which CPU to pin
    /// the current thread. Thread IDs should be assigned contiguously starting from 0. Be aware that
    /// the current thread will be restricted to one CPU for the rest of the execution!
```

src/bin/log/tracing_chrome_instant.rs
Stypox: Ok, I will use `asm!("", options(nostack, preserves_flags))` instead. Let me know if there are other options I should add to the `asm!` call. I found a suggestion to use  `asm!("" ::: "memory" : "volatile")` [here](https://rust-lang.github.io/rfcs/0888-compiler-fence-intrinsics.html) but that syntax for `asm!` is outdated and I am not sure if my version is equivalent.

src/bin/log/tracing_chrome_instant.rs
Stypox: Actually I found a better way: running the `__cpuid` instruction with the correct parameters and reading the correct field. See how it's done in the `raw_cpuid` crate ([here](https://docs.rs/raw-cpuid/latest/src/raw_cpuid/extended.rs.html#965-967)). I implemented that now and it works on my machine:

```rust
const LEAF: u32 = 0x80000007; // this is the leaf for "advanced power management info"
let cpuid = unsafe { __cpuid(LEAF) };
(cpuid.edx & (1 << 8)) != 0 // EDX bit 8 indicates invariant TSC
 ```

File: src/bin/log/tracing_chrome_instant.rs (hunk 8 above)
Comment (Stypox) [reply src/bin/log/tracing_chrome_instant.rs]: Fixed by getting the flag from `cpuid` instead

File: src/bin/log/tracing_chrome_instant.rs (hunk 10 above)
Comment (Stypox) [reply src/bin/log/tracing_chrome_instant.rs]: I added a comment to `with_elapsed_micros_subtracting_tracing`

File: src/bin/log/tracing_chrome_instant.rs (hunk 4 above)
Comment (Stypox) [reply src/bin/log/tracing_chrome_instant.rs]: Yeah this is a better phrasing, thank you

**Comment by Stypox** [comment]
Thank you for the review, I have addressed your comments.

Let me know if the fencing approach with `asm!` and the detection approach with `cpuid` look good to you.

Now in `with_elapsed_micros_subtracting_tracing` I added duplicate comments between the two `match` branches (because they do the same thing except with different types), also let me know if that's fine.

I also kept the `_ignored` as-is since that's how it also was before.

File: src/bin/log/tracing_chrome_instant.rs (hunk 9 above)
Comment (Stypox) [reply src/bin/log/tracing_chrome_instant.rs]: Turned it into a panic

File: src/bin/log/tracing_chrome_instant.rs (hunk 5 above)
Comment (RalfJung) [reply src/bin/log/tracing_chrome_instant.rs]: I would recommend just `asm!("")`, that is the most conservative.

File: src/bin/log/tracing_chrome_instant.rs [hunk 12]
```diff
@@ -61,20 +68,49 @@ impl TracingChromeInstant {
...
+                // instantiated, accounting for any time that was previously spent executing `f`.
+                // The "accounting" part is not computed in this line, but is rather done by
+                // shifting forward the `start_instant` down below.
                 let ts = (instant_before_f - *start_instant).as_nanos() as f64 / 1000.0;
+
+                // Run the function (supposedly a function internal to the tracing infrastructure).
                 f(ts);
+
+                // Measure how much time was spent executing `f` and shift `start_instant` forward
+                // by that amount.
```
RalfJung:
```suggestion
                // by that amount. This "removes" that time from the trace.
```

src/bin/log/tracing_chrome_instant.rs
```diff
@@ -61,20 +68,49 @@ impl TracingChromeInstant {
...
             #[cfg(all(target_os = "linux", any(target_arch = "x86", target_arch = "x86_64")))]
             TracingChromeInstant::Tsc { start_tsc, tsc_to_microseconds } => {
+                // Obtain the current time (before executing `f`).
                 let tsc_before_f = tsc::rdtsc();
+
+                // Using the current time (`instant_before_f`) and the `start_instant` stored in
+                // `self`, calculate the elapsed time (in microseconds) since this instant was
+                // instantiated, accounting for any time that was previously spent executing `f`.
+                // The "accounting" part is not computed in this line, but is rather done by
+                // shifting forward the `start_instant` down below.
```
RalfJung: You don't have to repeat the comments since it's the same logic as above.

src/bin/log/tracing_chrome_instant.rs
```diff
@@ -94,13 +130,15 @@ mod tsc {
         use core::arch::x86::{_mm_lfence, _rdtsc};
         #[cfg(target_arch = "x86_64")]
         use core::arch::x86_64::{_mm_lfence, _rdtsc};
-        use core::sync::atomic::{Ordering, compiler_fence};
+        use std::arch::asm;
 
         unsafe {
+            // Fence the execution with `_mm_lfence` and `asm!` to ensure that neither the compiler
+            // nor the CPU reorder instructions, as that would make _rdtsc() run at the wrong time.
```
RalfJung: Is the lfence really needed? And why is it the right CPU fence here?

Stypox: I did see `_mm_lfence()` used somewhere and took it from there, but I don't know why the load+store `mfence` was not being used instead. I tried looking up more information about this online and found many implementations that don't fence at all, [one](https://github.com/koeninger/tsc-trace/blob/main/src/lib.rs#L110) that also fences with `lfence` (but only optionally), [one](https://github.com/sheroz/tick_counter/blob/main/src/lib.rs#L169) that mixes `lfence` and `mfence` (but only when starting a timer and nevertheless it doesn't make much sense).

None of the implementations I found online fence against the compiler, so that's probably not needed after all. And given that we are not trying to measure the time taken for single/(very few) assembly instructions, it might not matter in any case whether we fence or not, since CPU reordering doesn't reorder instructions that are too far apart.

I tried removing the fencing completely and Miri+tracing got ~15% faster, which is something we might want to keep. The two following images show the usual table obtained with the query below on two trace files: the first one obtained with fences (lfence + `asm!("", options(nostack, preserves_flags))`) and the other with no fencing at all, and the percentages are very close (if not identical), indicating that removing fences does not change much what we actually measure.

image

image

```sql
select "TOTAL PROGRAM DURATION" as name, count(*), max(ts + dur) as "sum(dur)", 100.0 as "%", null as "min(dur)", null as "max(dur)", null as "avg(dur)", null as "stddev(dur)" from slices union select "TOTAL OVER ALL SPANS (excluding events)" as name, count(*), sum(dur), cast(cast(sum(dur) as float) / (select max(ts + dur) from slices) * 1000 as int) / 10.0 as "%", min(dur), max(dur), cast(avg(dur) as int) as "avg(dur)", cast(sqrt(avg(dur*dur)-avg(dur)*avg(dur)) as int) as "stddev(dur)" from slices where parent_id is null and name != "frame" and name != "step" and dur > 0 union select name, count(*), sum(dur), cast(cast(sum(dur) as float) / (select max(ts + dur) from slices) * 1000 as int) / 10.0 as "%", min(dur), max(dur), cast(avg(dur) as int) as "avg(dur)", cast(sqrt(avg(dur*dur)-avg(dur)*avg(dur)) as int) as "stddev(dur)" from slices where parent_id is null and name != "frame" and name != "step" group by name order by sum(dur) desc, count(*) desc
```

So given the above observations, I'd remove fencing completely. I pushed a commit to do so.

RalfJung:

@rustbot author

src/bin/log/tracing_chrome_instant.rs
matthieu-m: AFAIK the fences (or `rdtscp`) are typically used to ensure measurement accuracy by preventing the CPU from reordering the `rdtsc` instruction and thus "blur" the measurement. It's mostly useful when very accurate timing is necessary, and overkill otherwise.

matthieu-m:
A few years ago I went down the rabbit hole of constructing the lowest overhead clock measurement possible.

The heart of it was relatively simple:

 1. Use `rdtsc` to obtain the number of cycles.
 2. Use an affine formula to scale that to the number of nanoseconds since the start of the unix epoch.
 3. Regularly re-evaluate the result of the affine formula against the real clock, to catch any drift.

The latter is actually surprisingly hard to get right, and it took a couple of tries and measurements to figure out exactly how to adjust the affine formula to get a monotonic timestamp & avoid "swerves" on readjustments.

And the whole exercise proved near pointless, in practice:

 - This neat little clock took about 14ns to compute the real time.
 - A call to [`gettimeofday`](https://man7.org/linux/man-pages/man2/settimeofday.2.html), which is vDSO accelerated, takes 16ns.

It turns out that this whole thing is exactly how `gettimeofday` is essentially implemented under the hood, and while there's possibly a slight overhead (+2ns) in calling it, it saves a lot of hassle.

Is there a reason not to use `gettimeofday`, at least when available and vDSO-accelerated, instead of trying to implement a custom clock?

RalfJung:
`clock_gettime` is also on the vDSO, AFAIK. And that's what libstd uses. But somehow @Stypox got a lot of overhead with that, right?

Stypox:
Thanks for the insights! On my PC rdtsc is not considered usable as a source for time measurements by the kernel, which thus relies on slower sources. My CPU is quite recent (AMD Ryzen 7 5800H) so I am not sure why the kernel can't use rdtsc with it.

I guess this PR could be improved by checking which time source the kernel is using, and if it's already vDSO, we can use Instant normally.

jsimmons:
> Thanks for the insights! On my PC rdtsc is not considered usable as a source for time measurements by the kernel, which thus relies on slower sources. My CPU is quite recent (AMD Ryzen 7 5800H) so I am not sure why the kernel can't use rdtsc with it.
> 
> I guess this PR could be improved by checking which time source the kernel is using, and if it's already vDSO, we can use Instant normally.

The kernel disables tsc when it detects that it's not stable. It should say something in dmesg during boot about the time stamp source and measured frequency. Note that new hardware does not guarantee that your hardware tsc works, the clocks are notorious for being buggy on some platforms.

It's still possible to get an improvement from directly sampling with rdtsc, but if it's huge, it's a system configuration problem.

matthieu-m:
For the sake of completeness, I benchmarked today `std::time::Instant::now()` on 2 distinct x64 machines:

1. A "gaming" work desktop computer, running Ubuntu 18 via WSL2.
2. One of the largest AWS instances I have access to (96 cores), running Ubuntu 22.

I don't have the exact CPU versions off-hand -- I should have taken notes --though both support x86-64-v4 (ie, AVX512) and are thus relatively recent.

On my desktop computer, I got timings of about 12ns, while on the AWS server, I got timings of about 16ns-20ns. I expect difference is mostly due to clocking -- those giant AWS instances are at about 3GHz, my own work computer would be pushing 4GHz.

Needless to say, it's a far cry from 1.2us, and is more in line with roughly the expected performance of `rdtsc` (~6ns) followed by clock frequency adjustment (specifically, multiplying the number cycles by the inverse of the clock frequency).

RalfJung:
Okay now I also have to try this.^^ What's a good quick way to benchmark `Instant::now`?

RalfJung:
I ran the example I found [here](https://stackoverflow.com/questions/55583503/is-stdtimeduration-as-precise-as-timeprecise-time-ns-from-time-crate) and I'm seeing numbers around 16-22ns.

So... yeah that manual rdtsc doesn't usually seem necessary. Odd that `Instant::now` is so slow on your system, @Stypox.

jsimmons:
Sorry, I wasn't on my actual PC before. But to elaborate a bit, you're looking for these lines in your dmesg output. There's tsc-early too, which maybe says something interesting.

```
Aug 23 12:55:34 archlinux kernel: tsc: Fast TSC calibration using PIT
Aug 23 12:55:34 archlinux kernel: tsc: Detected 3393.477 MHz processor
Aug 23 12:55:34 archlinux kernel: tsc: Refined TSC clocksource calibration: 3393.623 MHz
Aug 23 12:55:34 archlinux kernel: clocksource: tsc: mask: 0xffffffffffffffff max_cycles: 0x30eac649395, max_idle_ns: 440795252668 ns
Aug 23 12:55:34 archlinux kernel: clocksource: Switched to clocksource tsc
```

On x86 you can also have other, dramatically slower, clock sources. Either virtualized timestamps in a vm, or things like hpet. Linux will try to default to tsc, but if it fails to calibrate because your time stamp counter isn't behaving, it will fall back to slower methods.

Generally speaking, if you aren't seeing tsc it's a big performance issue, and you might want to investigate more deeply, especially if you like to play video games on the machine. :) one thing to check is that you have the latest bios for your motherboard, since sometimes they fix things like this in patches.

Stypox:
It seems like there is [a BIOS bug](https://forums.lenovo.com/t5/Other-Linux-Discussions/TSC-clocksource-not-work/m-p/5126977) on my PC which has never been fixed, and TSC is therefore always unstable. I was able to force the kernel to use TSC anyway with `tsc=reliable clocksource=tsc hpet=disable`, which is however risky and problematic, but does make `Instant` much faster (~30ns), almost on par with reading raw TSC (~13ns). But anyway, it seems like it was just my PC's BIOS having problems with unstable TSC, not a widespread issue, so I guess parts of this PR could be reverted.

RalfJung:
Yeah, maybe we should remove the RDTSC code again? We can have some note in the docs for how people can tell whether they get the fast clock on Linux, and what they can do if they get a slow clock.

PR #4535, opened 2025-08-21
Rustup

RalfJung:
Bisecting for https://github.com/rust-lang/miri/pull/4534


RalfJung:
Same error, which makes sense if the cause of the problem is https://github.com/rust-lang/rust/pull/145622.

PR #4539, opened 2025-08-22
Add documentation for tracing

Stypox:
This PR adds documentation under the `doc/` folder, for the work that was done on the tracing infrastructure so far. I spellchecked the file but there might still be typos, let me know if you can find any.  Also let me know if there is anything I missed in the file.

[Rendered](https://github.com/Stypox/miri/blob/tracing-docs/doc/tracing.md)


doc/tracing.md
```diff
@@ -0,0 +1,334 @@
+# Documentation for the tracing infrastructure in Miri
+
+Miri can be traced to understand how much time is spent in its various components (e.g. borrow tracker, data race checker, etc.). When tracing is enabled, running Miri will create a `.json` trace file that can be opened and analyzed in [Perfetto](https://ui.perfetto.dev/). The events and spans generated throughout the codebase are collected by [the `tracing` crate](https://crates.io/crates/tracing), which then dispatches them to the code that writes to the trace file, but also to the logger if logging is enabled. For any questions regarding this documentation you may contact [Stypox](https://rust-lang.zulipchat.com/#narrow/dm/627563-Stypox) on Zulip.
```
RalfJung:
```suggestion
Miri can be traced to understand how much time is spent in its various components (e.g. borrow tracker, data race checker, etc.). When tracing is enabled, running Miri will create a `.json` trace file that can be opened and analyzed in [Perfetto](https://ui.perfetto.dev/). For any questions regarding this documentation you may contact [Stypox](https://rust-lang.zulipchat.com/#narrow/dm/627563-Stypox) on Zulip. 
```

doc/tracing.md
```diff
@@ -0,0 +1,334 @@
...
+### The timeline
+
+You will see the boxes "Global Legacy Events" and "Process 1" on the left of the workspace: after clicking on either of them their timeline will expand and you will be able to zoom in and look at individual spans (and events).
+
+- "Process 1" contains tracing spans for the various components of Miri, all in a single timeline line (e.g. borrow tracker, data race checker, etc.)
+- "Global Legacy Events" contains auxiliary spans on two separate lines that allow understanding what code is being executed at any point in time:
+    - "frame": what is the current stack frame in the interpreted program
+    - "step": what statement/terminator in the MIR of the interpreted program is being executed
+
+Spans are represented as colored boxes in the timeline, while instantaneous events are represented by tiny arrows. Events were there even before any tracing was added, because they are the same events that would get logged if logging was enabled.
```
RalfJung:
```suggestion
Spans are represented as colored boxes in the timeline, while instantaneous events are represented by tiny arrows. (Events exist because rustc and Miri also use the `tracing` crate for debug logging, and those logs turn into events in the trace.)
```

doc/tracing.md
```diff
@@ -0,0 +1,334 @@
...
+
+```rust
+let _trace = enter_trace_span!(M, layouting::fn_abi_of_instance, ?instance, ?extra_args);
+```
+
+![](./img/perfetto_span.png)
+
+### SQL tables
+
+Perfetto supports querying the span/event database using SQLite queries. Just type `:` in the search bar at the top to enter SQL mode, and then you will be able to enter SQL queries there. The relevant SQL tables are:
```
RalfJung:
```suggestion
Perfetto supports querying the span/event database using SQL queries. Just type `:` in the search bar at the top to enter SQL mode, and then you will be able to enter SQL queries there. The relevant SQL tables are:
```
I assume there's no SQLite involved here?

doc/tracing.md
```diff
@@ -0,0 +1,334 @@
...
+order by sum(dur) desc, count(*) desc
+```
+
+This is the kind of table you would get out:
+
+![](./img/perfetto_aggregate_statistics_sql.png)
+
+### Statistics about subnames of a span
+
+
```
RalfJung:
```suggestion

```

doc/tracing.md
```diff
@@ -0,0 +1,334 @@
...
+The solution to make it possible to check whether tracing is enabled at compile time even in `rustc_const_eval` was to add a function with this signature to the `Machine` trait:
+```rust
+fn enter_trace_span(span: impl FnOnce() -> tracing::Span) -> impl EnteredTraceSpan
+```
+
+where `EnteredTraceSpan` is just a marker trait implemented by `()` and `tracing::span::EnteredSpan`. This function returns `()` by default (without calling the `span` closure), except in `MiriMachine` where if tracing is enabled it will return `span().entered()`.
+
+The code in `rustc_const_eval` calls this function when it wants to do tracing, and the compiler will supposedly optimize out tracing calls when tracing is disabled.
```
RalfJung:
```suggestion
The code in `rustc_const_eval` calls this function when it wants to do tracing, and the compiler will (hopefully) optimize out tracing calls when tracing is disabled.
```

doc/tracing.md
```diff
@@ -0,0 +1,334 @@
...
+
+### Making a flamegraph
+
+After compiling Miri, you can run the following command to make a flamegraph using Linux' `perf`. It can be useful to spot functions that use up a significant amount of time but that are not yet covered by tracing calls.
+
+```sh
+perf record  --call-graph dwarf -F 999 ./miri/target/debug/miri --edition 2021 --sysroot ~/.cache/miri ./tests/pass/hashmap.rs && perf script | inferno-collapse-perf | inferno-flamegraph > flamegraph.svg
+```
+
+![](./img/flamegraph.svg)
```
RalfJung: This file does not seem to exist?

doc/tracing.md
```diff
@@ -0,0 +1,334 @@
...
+
+### Time measurements
+
+tracing-chrome originally used `std::time::Instant` to measure time, however on some x86/x86_64 Linux systems it might be unbearably slow since the underlying system call (`clock_gettime`) would take ≈1.3µs. Read more [here](https://btorpey.github.io/blog/2014/02/18/clock-sources-in-linux/) about how the Linux kernel chooses the clock source.
+
+Therefore, on x86/x86_64 Linux systems with a CPU that has an invariant TSC counter, we read from that instead to measure time, which takes only ≈13ns. There are unfortunately a lot of caveats to this approach though, as explained [in the code](https://github.com/rust-lang/miri/blob/master/src/bin/log/tracing_chrome_instant.rs) and [in the PR](https://github.com/rust-lang/miri/pull/4524). The most impactful one is that: every thread spawned in Miri that wants to trace something (which requires measuring time) needs to pin itself to a single CPU core (using `sched_setaffinity`).
+
+## Other useful stuff
+
+### Good code snippets to analyze traces of
```
RalfJung: Not sure if thee are really better than the benchmarks we already have?

doc/tracing.md
```diff
@@ -0,0 +1,334 @@
...
+select a.ts+a.dur as ts, b.ts-a.ts-a.dur as dur, a.id, a.track_id, a.category, a.depth, a.stack_id, a.parent_stack_id, a.parent_id, a.arg_set_id, a.thread_ts, a.thread_instruction_count, a.thread_instruction_delta, a.cat, a.slice_id, "empty" as name
+from ordered as a inner join ordered as b on a.rn=b.rn-1
+order by b.ts-a.ts-a.dur desc
+```
+
+### Saving Perfetto's state as a preset
+
+Unfortunately Perfetto does not seem to support saving the UI state as a preset that can be used to repeat the same analysis on multiple traces. You have to click through the various menus or run the various SQL queries every time to setup the UI as you want.
+
+## Adding new tracing calls to the code
```
RalfJung, line 183:
```suggestion
## Implementation details

Here we explain how tracing is implemented internally, and how to add new tracing calls to the code.

The events and spans generated throughout the codebase are collected by [the `tracing` crate](https://crates.io/crates/tracing), which then dispatches them to the code that writes to the trace file, but also to the logger if logging is enabled. 
```

doc/tracing.md
Stypox: The SQL data is stored in a SQLite database: https://perfetto.dev/docs/analysis/perfetto-sql-syntax

I guess I could explain it better and add that link

RalfJung: I think just saying "SQL queries" and linking to that page is fine.

doc/tracing.md
Stypox: Yes this was left here by accident, removed now

doc/tracing.md
Stypox: I would still keep two separate sections for implementation details and how to add tracing calls to the code: when just adding new tracing calls one needs to know about the latter but does not care about the former.

RalfJung: Ah right, I missed that there already is an "impl details" section. This paragraph should probably go there then.

doc/tracing.md
Stypox: Yeah I guess they are not real benchmarks, but their duration (and also the size of the trace file) is customizable by choosing a different `n` or `thread_count` which is useful. 

Should I remove this section altogether? Or replace it with something else?

Stypox:
Thanks for the review! I pushed a commit addressing all comments but the last two

doc/tracing.md
RalfJung: I don't think the examples are important enough to have them here. So probably best to just remove the section.

doc/tracing.md
Stypox: Yeah I already did so in [`993d964` (#4539)](https://github.com/rust-lang/miri/pull/4539/commits/993d96420a91b641302f043adb614c08e72d6066)

Stypox:
I removed the section with examples, all of the review comments should be addressed now

RalfJung:

@rustbot author

Issue #4483, opened 2025-07-19
Windows: File::create fails when file already exists (cannot call SetFileInformationByHandle)


RalfJung:
When calling `File::create` and the file already exists, we hit an error on Windows targets:
```
  error: unsupported operation: can't call foreign function `SetFileInformationByHandle` on OS `windows`
    --> C:\Users\runneradmin\.rustup\toolchains\miri\lib\rustlib\src\rust\library\std\src\sys\pal\windows\api.rs:231:26
     |
  LL |             let result = c::SetFileInformationByHandle(handle, class, info, size);
     |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ unsupported operation occurred here
     |
     = help: this means the program tried to do something Miri does not support; it does not indicate a bug in the program
     = note: BACKTRACE:
     = note: inside `std::sys::pal::windows::api::set_file_information_by_handle::set_info` at C:\Users\runneradmin\.rustup\toolchains\miri\lib\rustlib\src\rust\library\std\src\sys\pal\windows\api.rs:231:26: 231:82
     = note: inside `std::sys::pal::windows::api::set_file_information_by_handle::<std::sys::pal::windows::c::windows_sys::FILE_ALLOCATION_INFO>` at C:\Users\runneradmin\.rustup\toolchains\miri\lib\rustlib\src\rust\library\std\src\sys\pal\windows\api.rs:236:14: 236:68
     = note: inside `std::sys::fs::windows::File::open_native` at C:\Users\runneradmin\.rustup\toolchains\miri\lib\rustlib\src\rust\library\std\src\sys\fs\windows.rs:333:17: 333:79
     = note: inside `std::sys::fs::windows::File::open` at C:\Users\runneradmin\.rustup\toolchains\miri\lib\rustlib\src\rust\library\std\src\sys\fs\windows.rs:305:9: 305:39
     = note: inside `std::fs::OpenOptions::_open` at C:\Users\runneradmin\.rustup\toolchains\miri\lib\rustlib\src\rust\library\std\src\fs.rs:1711:9: 1711:42
     = note: inside `std::fs::OpenOptions::open::<&std::path::Path>` at C:\Users\runneradmin\.rustup\toolchains\miri\lib\rustlib\src\rust\library\std\src\fs.rs:1707:9: 1707:34
[...]

PR #4547, opened 2025-08-29
Implement SetFileInformationByHandle, as well as direct shim test for it

CraftSpider:
Fixes rust-lang/miri#4483

### DIFF
--- a/src/shims/windows/fs.rs
+++ b/src/shims/windows/fs.rs
@@ -5,6 +5,7 @@ use std::path::PathBuf;
 use std::time::SystemTime;
 
 use bitflags::bitflags;
+use rustc_abi::Size;
 use rustc_target::spec::Os;
 
 use crate::shims::files::{FdId, FileDescription, FileHandle};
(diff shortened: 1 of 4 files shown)


tests/pass/shims/fs.rs
RalfJung, line 1: Can you also add a `File::create` on a file that already exists somewhere suitable in this file so we ensure we cover that codepath?

tests/pass/shims/fs.rs
```diff
@@ -209,7 +208,10 @@ fn test_file_set_len() {
 
     // Can't use set_len on a file not opened for writing
     let file = OpenOptions::new().read(true).open(&path).unwrap();
-    assert_eq!(ErrorKind::InvalidInput, file.set_len(14).unwrap_err().kind());
+    assert_eq!(
+        if cfg!(windows) { ErrorKind::PermissionDenied } else { ErrorKind::InvalidInput },
```
RalfJung: Since we pass through the host OS error, I doubt this will work properly. Just check that it's one of the two (with a comment explaining why).

src/shims/windows/fs.rs
```diff
@@ -373,6 +374,64 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            err_unsup_format!(
+                "`SetFileInformationByHandle` is only supported on file-backed file descriptors"
+            )
+        })?;
+
+        if class == this.eval_windows_u32("c", "FileEndOfFileInfo") {
+            let ptr = this.deref_pointer_as(
+                file_information,
+                this.windows_ty_layout("FILE_END_OF_FILE_INFO"),
+            )?;
```
RalfJung: This re-reads the pointer. Since you already read the pointer, you can just call `ptr_to_mplace` directly.

src/shims/windows/fs.rs
```diff
@@ -373,6 +374,64 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        buffer_size: &OpTy<'tcx>,      // DWORD
+    ) -> InterpResult<'tcx, Scalar> {
+        // ^ Returns BOOL (i32 on Windows)
+        let this = self.eval_context_mut();
+        this.assert_target_os("windows", "SetFileInformationByHandle");
+        this.check_no_isolation("`SetFileInformationByHandle`")?;
+
+        let class = this.read_scalar(class)?.to_u32()?;
+        let buffer_size = this.read_scalar(buffer_size)?.to_u32()?;
+        let file_info = this.read_pointer(file_information)?;
```
RalfJung: Please pick a consistent name (`file_information` vs `file_info`).

src/shims/windows/fs.rs
```diff
@@ -373,6 +374,64 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        this.check_ptr_access(
+            file_info,
+            Size::from_bytes(buffer_size),
+            CheckInAllocMsg::MemoryAccess,
+        )?;
+
+        let file = this.read_handle(file, "GetFileInformationByHandle")?;
```
RalfJung:
```suggestion
        let file = this.read_handle(file, "SetFileInformationByHandle")?;
```
Happens a few more times below.

RalfJung (commented):
Thanks!

RalfJung:
@rustbot author

There's also a conflict that prevents CI from running, so please rebase to resolve that.

CraftSpider:
@rustbot ready

Still pending CI, but rebased and responded to review (finally)

src/shims/windows/fs.rs
```diff
@@ -419,6 +417,35 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     interp_ok(this.eval_windows("c", "FALSE"))
                 }
             }
+        } else if class == this.eval_windows_u32("c", "FileAllocationInfo") {
+            // File allocation size is independent of file EOF position on Windows, except that
+            // EOF must be <= allocation size. So we only do something if the file is bigger than
+            // the requested allocation size.
```
RalfJung: So in other words, we don't grow, we only shrink? I don't understand what that has to do with "EOF position" vs "allocation size". I also don't know what those terms even mean. ;)

src/shims/windows/fs.rs
```diff
@@ -406,10 +406,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         })?;
 
         if class == this.eval_windows_u32("c", "FileEndOfFileInfo") {
-            let ptr = this.deref_pointer_as(
-                file_information,
-                this.windows_ty_layout("FILE_END_OF_FILE_INFO"),
-            )?;
+            let ptr = this
```
RalfJung: It is somewhat odd to call this a `ptr` when it is *not* a `Pointer`, unlike `file_information` which is a `Pointer` but not called `ptr`. ;)

Name suggestions: `place`, `file_information`.

RalfJung (changes requested):
Thanks! Unfortunately there's some new code which causes confusion, probably due to my complete lack of understanding of Windows APIs. ;)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4547/dc1ede04df7ff834f2a025c59618dd7bcd2cc244..e9b084e3b514cdbaf8126ce397082905941a35eb)*

src/shims/windows/fs.rs
```diff
@@ -372,6 +373,87 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            let new_len =
+                this.read_scalar(&this.project_field_named(&ptr, "EndOfFile")?)?.to_i64()?;
+            match file.file.set_len(new_len.try_into().unwrap()) {
+                Ok(_) => interp_ok(this.eval_windows("c", "TRUE")),
+                Err(e) => {
+                    this.set_last_error(e)?;
+                    interp_ok(this.eval_windows("c", "FALSE"))
+                }
+            }
+        } else if class == this.eval_windows_u32("c", "FileAllocationInfo") {
```
RalfJung, line 420: This doesn't have a direct test.

src/shims/windows/fs.rs
CraftSpider: EOF position is the file's length. Allocation size is generally a multiple of the underlying storage's sector size, and is effectively the file's pre-allocated capacity before writing will have to map more physical storage. 

Rust `std` doesn't expose this cross-platform, and we don't support it in GetFileInformationByHandle, so the only _observable_ side effect of setting it is truncating length to be less than capacity.

src/shims/windows/fs.rs
```diff
@@ -372,6 +373,87 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // File allocation size is independent of file EOF position on Windows, except that
+            // EOF must be <= allocation size. So we only do something if the file is bigger than
+            // the requested allocation size.
```
RalfJung: So, is this correct?
```suggestion
            // On Windows, files are somewhat similar to a `Vec` in that they have a separate
            // "length" (called "EOF position") and "capacity" (called "allocation size").
            // Growing the allocation size is largely a performance hint which we can
            // ignore -- it can also be directly queried, but we currently do not support that.
            // So we only need to do something if this operation shrinks the allocation size
            // so far that it affects the EOF position.
```

src/shims/windows/fs.rs
RalfJung: Okay. I have made a suggestion for updating the comment.

However, I am a bit concerned about just silently doing nothing for the other case, where the allocation size is grown beyond the EOF position. Can't that break stuff? Or is it okay to consider this more of a hint, in some sense?

CraftSpider: I _think_ this is okay to consider as a hint.
- Extra allocation beyond file size is automatically freed when the last HANDLE to a file is closed
- Programs should be resilient to different cluster allocation sizes, since it's configurable.
- A quick look through the first couple pages of a GitHub search shows mostly uses to set a file size to 0, and a couple uses to pre-allocate space solely for write efficiency.
- I think we could emulate it with a side map tracking a HANDLE's allocation size, updated on writes beyond it, if we need to. This may be worth it if we want/need to support `GetFileInformationByHandle` for allocation size.

@ChrisDenton, I know you're familiar with Windows, so I'd appreciate your thoughts. Is this value guaranteed observable in other ways that I'm not finding it my search?

ChrisDenton: Yes, that sounds right. Its intended use is as a performance optimisation. It's similar to `Vec::with_capacity` in that you can reserve some space upfront then write into it later. It's not really observable outside of directly querying it. And of course write performance if your writes would have otherwise triggered the allocation to grow.

RalfJung: Makes sense, thanks! I have updated the suggested comment accordingly.

RalfJung:

@rustbot author

PR #4549, opened 2025-08-30
native-lib mode: avoid unsoundness due to mrpotect

RalfJung:
I think the way we used mprotect was unsound since LLVM could have moved accesses to the protected memory into the section where that memory is protected, causing a segfault.

To prevent that, we should *first* negotiate with the supervisor to set up the segfault handler, and *then* do the page protecting business. That ensures that there is no moment in time where accessing this memory aborts the program: a segfault does get triggered, but it is resolved transparently to the program, which is fine.

@nia-e is there any issue with reordering things like this?

### DIFF
--- a/src/shims/native_lib/trace/child.rs
+++ b/src/shims/native_lib/trace/child.rs
@@ -90,14 +90,6 @@ impl Supervisor {
         // Unwinding might be messed up due to partly protected memory, so let's abort if something
         // breaks inside here.
         let res = std::panic::abort_unwind(|| {
-            // SAFETY: We do not access machine memory past this point until the
-            // supervisor is ready to allow it.
-            // FIXME: this is sketchy, as technically the memory is still in the Rust Abstract Machine,
-            // and the compiler would be allowed to reorder accesses below this block...
-            unsafe {
-                Self::protect_pages(alloc.pages(), mman::ProtFlags::PROT_NONE).unwrap();
-            }
-
             // Send over the info.
             // NB: if we do not wait to receive a blank confirmation response, it is
             // possible that the supervisor is alerted of the SIGSTOP *before* it has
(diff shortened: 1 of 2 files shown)


src/shims/native_lib/trace/child.rs
```diff
@@ -130,6 +120,18 @@ impl Supervisor {
                 .unwrap();
             }
 
+            // We can't use IPC channels here to signal that FFI mode has ended,
+            // since they might allocate memory which could get us stuck in a SIGTRAP
+            // with no easy way out! While this could be worked around, it is much
```
RalfJung: I don't understand this comment. What is the problem with allocating memory / SIGTRAP?

nia-e: Right now there's none (I guess it's a leftover comment I didn't properly update when I cut out the malloc tracing), but I implemented it this way from the start because of the way I intended to implement said malloc tracing bits. Shouldn't be of concern to this PR, though

nia-e:
LGTM, this should be inconsequential unless I messed up something quite badly ^^

src/shims/native_lib/trace/parent.rs
```diff
@@ -289,8 +276,7 @@ pub fn sv_loop(
                         }
                     }
                 },
-            // Child entered a syscall; we wait for exits inside of this, so it
-            // should never trigger on return from a syscall we care about.
```
RalfJung, line 293: The part of the comment that said "we wait for exits inside of this" is also outdated, right? I don't see any logic waiting for anything to exit here.

nia-e: Yep, that was about trapping `mmap`

RalfJung:
Thanks!

src/shims/native_lib/trace/child.rs
RalfJung: FWIW, LLVM is in principle allowed to insert allocations anywhere or at least move them around... we may have to do some not-entirely-kosher reasoning for the syscall tracing then. ;)

Issue #4550, opened 2025-08-31
There are such things as jump instructions that access memory


comex:
From [native_lib/trace/parent.rs](https://github.com/rust-lang/miri/blob/faf02b64d65ea082743585572a8e7032688efc2a/src/shims/native_lib/trace/parent.rs#L515):
```
    // Save registers and grab the bytes that were executed. This would
    // be really nasty if it was a jump or similar but those thankfully
    // won't do memory accesses and so can't trigger this!
```

But x86_64 actually can perform a memory load and jump in a single instruction, such as `callq *(%rax)`.  (Not to be confused with `callq *%rax`, which doesn't load from memory, despite the `*`.  AT&T syntax is great.)

Disclaimer: I'm writing this off-the-cuff and haven't tried to actually reproduce this issue with Miri.

Other notes:

The `ret` instruction likewise performs a memory access and jumps, but the access is to the stack, so it would only become an issue if the native library placed its stack in the guarded memory region.  There are probably 

32-bit x86 has the same issue as x86-64.  As for other architectures (that aren't currently supported anyway), most of them can't jump and access memory in a single instruction, but 32-bit ARM can.

RalfJung:
Cc @nia-e

nia-e:
I knew about the `ret` case, but thankfully that should never be an issue in our usecase. I didn't know this was possible for `callq` and admittedly that does make this annoying - I'm not sure how machine code is generated for e.g. passing pointers to local closures but I imagine that might trigger it.

Either way, I can rework the logic to be resilient against this, it's just going to be more annoying. Thanks for the heads up!

PR #4551, opened 2025-08-31
native-lib: more resilient grabbing of instruction bytes

nia-e:
Fixes rust-lang/miri#4550.

### DIFF
--- a/src/shims/native_lib/trace/parent.rs
+++ b/src/shims/native_lib/trace/parent.rs
@@ -18,6 +18,11 @@ const ARCH_WORD_SIZE: usize = 4;
 #[cfg(target_arch = "x86_64")]
 const ARCH_WORD_SIZE: usize = 8;
 
+// x86 max instruction length is 15 bytes:
+// https://www.intel.com/content/www/us/en/developer/articles/technical/intel-sdm.html
+// See vol. 3B section 24.25.
+const ARCH_MAX_INSTR_SIZE: usize = 15;
+
 /// The address of the page set to be edited, initialised to a sentinel null
 /// pointer.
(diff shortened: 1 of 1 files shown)


src/shims/native_lib/trace/parent.rs
```diff
@@ -472,7 +477,26 @@ fn handle_segfault(
...
     let mut new_regs = regs_bak;
-    let ip_prestep = regs_bak.ip();
+
+    // Read at least one instruction from the ip. It's possible that the instruction
+    // that triggered the segfault was short and at the end of the mapped text area,
+    // so some of these reads may fail; in that case, just write empty bytes. If all
+    // reads failed, the disassembler will report an error.
+    let instr = (0..(ARCH_MAX_INSTR_SIZE.div_ceil(ARCH_WORD_SIZE)))
+        .flat_map(|ofs| {
+            ptrace::read(
```
RalfJung, line 488: This reads an entire word, not just one byte, right? I have a hard time following the logic since it seems to then... concatenate 15 words of memory, or so?

RalfJung: Ah, that's what the `div_ceil` is for, I see.

RalfJung, line 488:
```suggestion
            // This reads one word of memory; we divided by `ARCH_WORD_SIZE` above to compensate for that.
            ptrace::read(
```

RalfJung:
This looks great, thanks! Please squash the suggestion into the existing commits using `./miri squash`. Then write `@rustbot ready` after you force-pushed the squashed PR.

@rustbot author

Issue #4556, opened 2025-09-02
Applied floating-point error seems too large


RalfJung:
`apply_random_float_error` is supposed to apply a random error in the range of a few ULP. However, this does not seem to behave quite as expected...

I tired this code:
```rust
        let exact = f64::from_bits(4602308182625945072);
        for _ in 0..100 {
            let val = (0.5f64).sin();
            let diff_ulp = 'diff: {
                if val >= exact {
                    let mut cur = exact;
                    for i in 0i32..128 {
                        if cur == val {
                            break 'diff i;
                        }
                        cur = cur.next_up();
                    }
                    unreachable!()
                } else {
                    let mut cur = exact;
                    for i in 0i32..128 {
                        if cur == val {
                            break 'diff -i;
                        }
                        cur = cur.next_down();
                    }
                    unreachable!()
                }
            };
            println!("Got a difference of {diff_ulp} ULP");
        }
```
Here 4602308182625945072 is the bit representation of the value that `(0.5f64).sin()` returns on my system when no error is applied.

That produces the following results:
```
Got a difference of -5 ULP
Got a difference of 8 ULP
Got a difference of -4 ULP
Got a difference of -1 ULP
Got a difference of -4 ULP
Got a difference of 6 ULP
Got a difference of 0 ULP
Got a difference of -4 ULP
Got a difference of -4 ULP
Got a difference of 4 ULP
Got a difference of -4 ULP
Got a difference of -1 ULP
Got a difference of -2 ULP
Got a difference of -4 ULP
Got a difference of -1 ULP
Got a difference of -7 ULP
Got a difference of 6 ULP
Got a difference of 2 ULP
Got a difference of -3 ULP
Got a difference of 8 ULP
Got a difference of 6 ULP
Got a difference of 4 ULP
Got a difference of 6 ULP
Got a difference of 4 ULP
Got a difference of -5 ULP
Got a difference of 6 ULP
Got a difference of -6 ULP
Got a difference of 4 ULP
Got a difference of 6 ULP
Got a difference of 4 ULP
Got a difference of 0 ULP
Got a difference of 4 ULP
Got a difference of 4 ULP
Got a difference of 0 ULP
Got a difference of -4 ULP
Got a difference of -1 ULP
Got a difference of 0 ULP
Got a difference of 0 ULP
Got a difference of -1 ULP
Got a difference of -3 ULP
Got a difference of -7 ULP
Got a difference of 2 ULP
Got a difference of 6 ULP
[...]

PR #4558, opened 2025-09-02
fix mangitude of applied float error

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/4556.

I honestly don't understand where this off-by-1 comes from, though...

### DIFF
--- a/src/math.rs
+++ b/src/math.rs
@@ -6,10 +6,6 @@ use rustc_middle::ty::{self, FloatTy, ScalarInt};
 use crate::*;
 
 /// Disturbes a floating-point result by a relative error in the range (-2^scale, 2^scale).
-///
-/// For a 2^N ULP error, you can use an `err_scale` of `-(F::PRECISION - 1 - N)`.
-/// In other words, a 1 ULP (absolute) error is the same as a `2^-(F::PRECISION-1)` relative error.
-/// (Subtracting 1 compensates for the integer bit.)
 pub(crate) fn apply_random_float_error<F: rustc_apfloat::Float>(
     ecx: &mut crate::MiriInterpCx<'_>,
     val: F,
@@ -17,11 +13,13 @@ pub(crate) fn apply_random_float_error<F: rustc_apfloat::Float>(
 ) -> F {
     if !ecx.machine.float_nondet
         || matches!(ecx.machine.float_rounding_error, FloatRoundingErrorMode::None)
+        // relative errors don't do anything to zeros... avoid messing up the sign
+        || val.is_zero()
     {
         return val;
     }
-
     let rng = ecx.machine.rng.get_mut();
+
     // Generate a random integer in the range [0, 2^PREC).
     // (When read as binary, the position of the first `1` determines the exponent,
     // and the remaining bits fill the mantissa. `PREC` is one plus the size of the mantissa,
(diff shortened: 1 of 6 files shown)


RalfJung:
Even with this, the distribution is far from uniform, which doesn't make me happy as it means the most extreme error occurs so rarely it's hard to test for them:
```
num(-4) = 20
num(-3) = 97
num(-2) = 73
num(-1) = 74
num(0) = 74
num(1) = 69
num(2) = 75
num(3) = 72
num(4) = 22
```

LorrensP-2158466:
Yeah, that's a bit unfortunate. Maybe we could generate some value between `0..=ulp` and based on a random bool, do `next_up` or `next_down`?

src/math.rs
```diff
@@ -37,27 +36,31 @@ pub(crate) fn apply_random_float_error<F: rustc_apfloat::Float>(
...
-    let n = i32::try_from(ulp_exponent)
-        .expect("`err_scale_for_ulp`: exponent is too large to create an error scale");
-    // we know this fits
+    let n = i32::try_from(ulp_exponent).unwrap();
     let prec = i32::try_from(F::PRECISION).unwrap();
-    let err_scale = -(prec - n - 1);
+    // Math would tell us we should set `-(prec - 1 - n)`, since a 1 ULP (absolute) error is the
+    // same as a `2^-(F::PRECISION-1)` relative error. However,
+    // `tests/pass/float_extra_rounding_error.rs` confirms that `-(prec - n)` behaves as expected.
+    // FIXME: figure out why...
```
eduardosm: Lets consider an hypothetical floating point format with `F::PRECISION=4`.

With a value of x=1.000<sub>2</sub> and an error of 1 ULP, we would have a range of 0.111<sub>2</sub> ≤ x ≤ 1.001<sub>2</sub>.

If we apply `-(prec - 1 - n)` (n=0 for 1 ULP):

|e|=x·2^(4-1-0)=1.000<sub>2</sub>·2^(-3)=0.001<sub>2</sub>.
x+|e|=1.000<sub>2</sub>+0.001<sub>2</sub>=1.001<sub>2</sub>
x-|e|=1.000<sub>2</sub>-0.001<sub>2</sub>=0.111<sub>2</sub>

So far, so good.

Now lets try with x=1.110<sub>2</sub>. With 1 ULP error 1.101<sub>2</sub> ≤ x ≤ 1.111<sub>2</sub>

Applying `-(prec - 1 - n)`:
|e|=x·2^(4-1-0)=1.110<sub>2</sub>·2^(-3)=0.00111<sub>2</sub>.
x+|e|=1.110<sub>2</sub>+0.00111<sub>2</sub>=10.00<sub>2</sub>
x-|e|=1.110<sub>2</sub>-0.00111<sub>2</sub>=1.100<sub>2</sub>

For both values (1.000<sub>2</sub> and 1.110<sub>2</sub>), the magnitude of an error of 1 ULP is less or equal to 0.001. But trying to apply a relative error will result in larger absolute errors for the larger value. The error value should be generated without taking the mantissa into account, only the exponent.

src/math.rs
```diff
@@ -37,43 +35,61 @@ pub(crate) fn apply_random_float_error<F: rustc_apfloat::Float>(
...
+            },
+        FloatRoundingErrorMode::None => unreachable!(),
+    };
+    let ulp = (val.next_up().value - val).value;
+    // Shift the value by N times the ULP
+    (val + (ulp * F::from_i128(err_ulp.into()).value).value).value
 }
 
-/// Applies a random 16ULP floating point error to `val` and returns the new value.
+/// Applies a random 2^N ULP floating point error to `val` and returns the new value.
```
LorrensP-2158466: I think this needs to be consistent with the docs of `apply_random_float_error_ulp`.

LorrensP-2158466:
Nice, I like this! It should make it easier for testing.

src/math.rs
RalfJung: Thanks for the analysis!

> The error value should be generated without taking the mantissa into account, only the exponent.

I don't know how that could look like. Instead I implemented something that works directly with the ULP and its multiples, which seems to behave as expected.

oli-obk (approved):

PR #4559, opened 2025-09-02
Add Zed editor config

nia-e:
Nearly identical to the vscode one, just that Zed requires different formatting for json and has an extra `initialization_options` block.


RalfJung:
That is quite annoying that we need a dozen different config files for RA that will inevitably go out of sync. :/

nia-e:
There's basically just one basic schema for all of them, right? I feel like it should be possible to auto-generate them somehow

PR #4506, opened 2025-07-29
Add minimal functionality for using GenMC mode

Patrick-6:
This PR is the MVP for Miri-GenMC, which adds enough to be able to run some simple programs.

A lot of functionality is missing and will be added in a future PR (e.g., atomic read-modify-write/compare_exchange/fences, mutex support, support for mixed atomic and non-atomic memory accesses to the same location, etc.)
(The commit history shows some of the features that were removed for this PR.)

The current branch points to a fork of GenMC, since there are some changes there that must be upstreamed first. This PR should not be merged until the required changes are available in upstream GenMC.

PR #4566, opened 2025-09-04
Add more features for GenMC mode (RMW, fences, new printing options)

Patrick-6:
This PR is a followup to https://github.com/rust-lang/miri/pull/4506

New features for GenMC mode include:
- Support for read-modify-write operations (no compare-exchange yet).
- Support for fences.
- New option to print GenMC output message on errors.
- New option to print GenMC execution graphs after every execution.
- New option to set memory model to "Sequential Consistency" (SC).
- ~Throw an error when attempting to use a futex, since they are unsupported.~
- Throw an error on `acquire_clock` and `release_clock` in GenMC mode.

There is also some cleanup for changes from the last PR, including the switch to `std::process::abort` in tests.

### DIFF
--- a/src/concurrency/data_race.rs
+++ b/src/concurrency/data_race.rs
@@ -958,16 +958,20 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
     /// with this program point.
     ///
     /// The closure will only be invoked if data race handling is on.
-    fn release_clock<R>(&self, callback: impl FnOnce(&VClock) -> R) -> Option<R> {
+    fn release_clock<R>(
+        &self,
+        callback: impl FnOnce(&VClock) -> R,
+    ) -> InterpResult<'tcx, Option<R>> {
         let this = self.eval_context_ref();
-        // FIXME: make this a proper error instead of ICEing the interpreter.
-        assert!(
-            this.machine.data_race.as_genmc_ref().is_none(),
-            "this operation performs synchronization that is not supported in GenMC mode"
-        );
-        Some(
-            this.machine.data_race.as_vclocks_ref()?.release_clock(&this.machine.threads, callback),
-        )
+        interp_ok(match &this.machine.data_race {
+            GlobalDataRaceHandler::None => None,
+            GlobalDataRaceHandler::Genmc(_genmc_ctx) =>
+                throw_unsup_format!(
+                    "this operation performs synchronization that is not supported in GenMC mode"
+                ),
+            GlobalDataRaceHandler::Vclocks(data_race) =>
+                Some(data_race.release_clock(&this.machine.threads, callback)),
+        })
     }
 
     /// Acquire the given clock into the current thread, establishing synchronization with
--- a/src/concurrency/sync.rs
+++ b/src/concurrency/sync.rs
@@ -204,7 +204,7 @@ pub(super) trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
             this.mutex_enqueue_and_block(mutex_ref, Some((retval, dest)));
         } else {
             // We can have it right now!
-            this.mutex_lock(&mutex_ref);
+            this.mutex_lock(&mutex_ref)?;
             // Don't forget to write the return value.
             this.write_scalar(retval, &dest)?;
         }
@@ -338,7 +338,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     }
 
     /// Lock by setting the mutex owner and increasing the lock count.
-    fn mutex_lock(&mut self, mutex_ref: &MutexRef) {
+    fn mutex_lock(&mut self, mutex_ref: &MutexRef) -> InterpResult<'tcx> {
         let this = self.eval_context_mut();
         let thread = this.active_thread();
         let mut mutex = mutex_ref.0.borrow_mut();
@@ -352,9 +352,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             mutex.owner = Some(thread);
         }
         mutex.lock_count = mutex.lock_count.strict_add(1);
-        if let Some(data_race) = this.machine.data_race.as_vclocks_ref() {
-            data_race.acquire_clock(&mutex.clock, &this.machine.threads);
-        }
+        this.acquire_clock(&mutex.clock)?;
+        interp_ok(())
     }
 
     /// Try unlocking by decreasing the lock count and returning the old lock
@@ -377,11 +376,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // The mutex is completely unlocked. Try transferring ownership
                 // to another thread.
 
-                if let Some(data_race) = this.machine.data_race.as_vclocks_ref() {
-                    data_race.release_clock(&this.machine.threads, |clock| {
-                        mutex.clock.clone_from(clock)
-                    });
-                }
+                this.release_clock(|clock| mutex.clock.clone_from(clock))?;
                 let thread_id = mutex.queue.pop_front();
                 // We need to drop our mutex borrow before unblock_thread
                 // because it will be borrowed again in the unblock callback.
@@ -425,7 +420,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     assert_eq!(unblock, UnblockKind::Ready);
 
                     assert!(mutex_ref.owner().is_none());
-                    this.mutex_lock(&mutex_ref);
+                    this.mutex_lock(&mutex_ref)?;
 
                     if let Some((retval, dest)) = retval_dest {
                         this.write_scalar(retval, &dest)?;
@@ -439,7 +434,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
     /// Read-lock the lock by adding the `reader` the list of threads that own
     /// this lock.
-    fn rwlock_reader_lock(&mut self, rwlock_ref: &RwLockRef) {
+    fn rwlock_reader_lock(&mut self, rwlock_ref: &RwLockRef) -> InterpResult<'tcx> {
         let this = self.eval_context_mut();
         let thread = this.active_thread();
         trace!("rwlock_reader_lock: now also held (one more time) by {:?}", thread);
@@ -447,9 +442,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         assert!(!rwlock.is_write_locked(), "the lock is write locked");
         let count = rwlock.readers.entry(thread).or_insert(0);
         *count = count.strict_add(1);
-        if let Some(data_race) = this.machine.data_race.as_vclocks_ref() {
-            data_race.acquire_clock(&rwlock.clock_unlocked, &this.machine.threads);
-        }
+        this.acquire_clock(&rwlock.clock_unlocked)?;
+        interp_ok(())
     }
 
     /// Try read-unlock the lock for the current threads and potentially give the lock to a new owner.
@@ -472,12 +466,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             }
             Entry::Vacant(_) => return interp_ok(false), // we did not even own this lock
         }
-        if let Some(data_race) = this.machine.data_race.as_vclocks_ref() {
-            // Add this to the shared-release clock of all concurrent readers.
-            data_race.release_clock(&this.machine.threads, |clock| {
-                rwlock.clock_current_readers.join(clock)
-            });
-        }
+        // Add this to the shared-release clock of all concurrent readers.
+        this.release_clock(|clock| rwlock.clock_current_readers.join(clock))?;
 
         // The thread was a reader. If the lock is not held any more, give it to a writer.
         if rwlock.is_locked().not() {
@@ -521,7 +511,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 }
                 |this, unblock: UnblockKind| {
                     assert_eq!(unblock, UnblockKind::Ready);
-                    this.rwlock_reader_lock(&rwlock_ref);
+                    this.rwlock_reader_lock(&rwlock_ref)?;
                     this.write_scalar(retval, &dest)?;
                     interp_ok(())
                 }
@@ -531,7 +521,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
     /// Lock by setting the writer that owns the lock.
     #[inline]
-    fn rwlock_writer_lock(&mut self, rwlock_ref: &RwLockRef) {
+    fn rwlock_writer_lock(&mut self, rwlock_ref: &RwLockRef) -> InterpResult<'tcx> {
         let this = self.eval_context_mut();
         let thread = this.active_thread();
         trace!("rwlock_writer_lock: now held by {:?}", thread);
@@ -539,9 +529,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let mut rwlock = rwlock_ref.0.borrow_mut();
         assert!(!rwlock.is_locked(), "the rwlock is already locked");
         rwlock.writer = Some(thread);
-        if let Some(data_race) = this.machine.data_race.as_vclocks_ref() {
-            data_race.acquire_clock(&rwlock.clock_unlocked, &this.machine.threads);
-        }
+        this.acquire_clock(&rwlock.clock_unlocked)?;
+        interp_ok(())
     }
 
     /// Try to unlock an rwlock held by the current thread.
@@ -559,11 +548,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             rwlock.writer = None;
             trace!("rwlock_writer_unlock: unlocked by {:?}", thread);
             // Record release clock for next lock holder.
-            if let Some(data_race) = this.machine.data_race.as_vclocks_ref() {
-                data_race.release_clock(&this.machine.threads, |clock| {
-                    rwlock.clock_unlocked.clone_from(clock)
-                });
-            }
+            this.release_clock(|clock| rwlock.clock_unlocked.clone_from(clock))?;
 
             // The thread was a writer.
             //
@@ -613,7 +598,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 }
                 |this, unblock: UnblockKind| {
                     assert_eq!(unblock, UnblockKind::Ready);
-                    this.rwlock_writer_lock(&rwlock_ref);
+                    this.rwlock_writer_lock(&rwlock_ref)?;
                     this.write_scalar(retval, &dest)?;
                     interp_ok(())
                 }
@@ -663,12 +648,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     match unblock {
                         UnblockKind::Ready => {
                             // The condvar was signaled. Make sure we get the clock for that.
-                            if let Some(data_race) = this.machine.data_race.as_vclocks_ref() {
-                                data_race.acquire_clock(
+                            this.acquire_clock(
                                     &condvar_ref.0.borrow().clock,
-                                    &this.machine.threads,
-                                );
-                            }
+                                )?;
                             // Try to acquire the mutex.
                             // The timeout only applies to the first wait (until the signal), not for mutex acquisition.
                             this.condvar_reacquire_mutex(mutex_ref, retval_succ, dest)
@@ -695,9 +677,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let mut condvar = condvar_ref.0.borrow_mut();
 
         // Each condvar signal happens-before the end of the condvar wake
-        if let Some(data_race) = this.machine.data_race.as_vclocks_ref() {
-            data_race.release_clock(&this.machine.threads, |clock| condvar.clock.clone_from(clock));
-        }
+        this.release_clock(|clock| condvar.clock.clone_from(clock))?;
         let Some(waiter) = condvar.waiters.pop_front() else {
             return interp_ok(false);
         };
@@ -736,9 +716,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                         UnblockKind::Ready => {
                             let futex = futex_ref.0.borrow();
                             // Acquire the clock of the futex.
-                            if let Some(data_race) = this.machine.data_race.as_vclocks_ref() {
-                                data_race.acquire_clock(&futex.clock, &this.machine.threads);
-                            }
+                            this.acquire_clock(&futex.clock)?;
                         },
                         UnblockKind::TimedOut => {
                             // Remove the waiter from the futex.
@@ -766,9 +744,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let mut futex = futex_ref.0.borrow_mut();
 
         // Each futex-wake happens-before the end of the futex wait
-        if let Some(data_race) = this.machine.data_race.as_vclocks_ref() {
-            data_race.release_clock(&this.machine.threads, |clock| futex.clock.clone_from(clock));
-        }
+        this.release_clock(|clock| futex.clock.clone_from(clock))?;
 
         // Remove `count` of the threads in the queue that match any of the bits in the bitset.
         // We collect all of them before unblocking because the unblock callback may access the
--- a/src/alloc_addresses/mod.rs
+++ b/src/alloc_addresses/mod.rs
@@ -498,6 +498,8 @@ impl<'tcx> MiriMachine<'tcx> {
         // Also remember this address for future reuse.
         let thread = self.threads.active_thread();
         global_state.reuse.add_addr(rng, addr, size, align, kind, thread, || {
+            // We already excluded GenMC above. We cannot use `self.release_clock` as
+            // `self.alloc_addresses` is borrowed.
             if let Some(data_race) = self.data_race.as_vclocks_ref() {
                 data_race.release_clock(&self.threads, |clock| clock.clone())
             } else {
--- a/src/concurrency/init_once.rs
+++ b/src/concurrency/init_once.rs
@@ -96,10 +96,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         init_once.status = InitOnceStatus::Complete;
 
         // Each complete happens-before the end of the wait
-        if let Some(data_race) = this.machine.data_race.as_vclocks_ref() {
-            data_race
-                .release_clock(&this.machine.threads, |clock| init_once.clock.clone_from(clock));
-        }
+        this.release_clock(|clock| init_once.clock.clone_from(clock))?;
 
         // Wake up everyone.
         // need to take the queue to avoid having `this` be borrowed multiple times
@@ -125,10 +122,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         init_once.status = InitOnceStatus::Uninitialized;
 
         // Each complete happens-before the end of the wait
-        if let Some(data_race) = this.machine.data_race.as_vclocks_ref() {
-            data_race
-                .release_clock(&this.machine.threads, |clock| init_once.clock.clone_from(clock));
-        }
+        this.release_clock(|clock| init_once.clock.clone_from(clock))?;
 
         // Wake up one waiting thread, so they can go ahead and try to init this.
         if let Some(waiter) = init_once.waiters.pop_front() {
--- a/src/shims/unix/linux_like/epoll.rs
+++ b/src/shims/unix/linux_like/epoll.rs
@@ -608,7 +608,7 @@ fn check_and_update_one_event_interest<'tcx>(
         // If we are tracking data races, remember the current clock so we can sync with it later.
         ecx.release_clock(|clock| {
             event_instance.clock.clone_from(clock);
-        });
+        })?;
         // Triggers the notification by inserting it to the ready list.
         ready_list.insert(epoll_key, event_instance);
         interp_ok(true)
--- a/src/shims/unix/linux_like/eventfd.rs
+++ b/src/shims/unix/linux_like/eventfd.rs
@@ -199,7 +199,7 @@ fn eventfd_write<'tcx>(
             // Future `read` calls will synchronize with this write, so update the FD clock.
             ecx.release_clock(|clock| {
                 eventfd.clock.borrow_mut().join(clock);
-            });
+            })?;
 
             // Store new counter value.
             eventfd.counter.set(new_count);
--- a/src/shims/unix/macos/sync.rs
+++ b/src/shims/unix/macos/sync.rs
@@ -296,7 +296,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
             this.mutex_enqueue_and_block(mutex_ref, None);
         } else {
-            this.mutex_lock(&mutex_ref);
+            this.mutex_lock(&mutex_ref)?;
         }
 
         interp_ok(())
@@ -321,7 +321,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             // reentrancy.
             this.write_scalar(Scalar::from_bool(false), dest)?;
         } else {
-            this.mutex_lock(&mutex_ref);
+            this.mutex_lock(&mutex_ref)?;
             this.write_scalar(Scalar::from_bool(true), dest)?;
         }
 
--- a/src/shims/unix/sync.rs
+++ b/src/shims/unix/sync.rs
@@ -498,14 +498,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     MutexKind::Normal => throw_machine_stop!(TerminationInfo::Deadlock),
                     MutexKind::ErrorCheck => this.eval_libc_i32("EDEADLK"),
                     MutexKind::Recursive => {
-                        this.mutex_lock(&mutex.mutex_ref);
+                        this.mutex_lock(&mutex.mutex_ref)?;
                         0
                     }
                 }
             }
         } else {
             // The mutex is unlocked. Let's lock it.
-            this.mutex_lock(&mutex.mutex_ref);
+            this.mutex_lock(&mutex.mutex_ref)?;
             0
         };
         this.write_scalar(Scalar::from_i32(ret), dest)?;
@@ -525,14 +525,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     MutexKind::Default | MutexKind::Normal | MutexKind::ErrorCheck =>
                         this.eval_libc_i32("EBUSY"),
                     MutexKind::Recursive => {
-                        this.mutex_lock(&mutex.mutex_ref);
+                        this.mutex_lock(&mutex.mutex_ref)?;
                         0
                     }
                 }
             }
         } else {
             // The mutex is unlocked. Let's lock it.
-            this.mutex_lock(&mutex.mutex_ref);
+            this.mutex_lock(&mutex.mutex_ref)?;
             0
         }))
     }
@@ -600,7 +600,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 dest.clone(),
             );
         } else {
-            this.rwlock_reader_lock(&rwlock.rwlock_ref);
+            this.rwlock_reader_lock(&rwlock.rwlock_ref)?;
             this.write_null(dest)?;
         }
 
@@ -615,7 +615,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         if rwlock.rwlock_ref.is_write_locked() {
             interp_ok(Scalar::from_i32(this.eval_libc_i32("EBUSY")))
         } else {
-            this.rwlock_reader_lock(&rwlock.rwlock_ref);
+            this.rwlock_reader_lock(&rwlock.rwlock_ref)?;
             interp_ok(Scalar::from_i32(0))
         }
     }
@@ -648,7 +648,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 dest.clone(),
             );
         } else {
-            this.rwlock_writer_lock(&rwlock.rwlock_ref);
+            this.rwlock_writer_lock(&rwlock.rwlock_ref)?;
             this.write_null(dest)?;
         }
 
@@ -663,7 +663,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         if rwlock.rwlock_ref.is_locked() {
             interp_ok(Scalar::from_i32(this.eval_libc_i32("EBUSY")))
         } else {
-            this.rwlock_writer_lock(&rwlock.rwlock_ref);
+            this.rwlock_writer_lock(&rwlock.rwlock_ref)?;
             interp_ok(Scalar::from_i32(0))
         }
     }
--- a/src/shims/unix/unnamed_socket.rs
+++ b/src/shims/unix/unnamed_socket.rs
@@ -259,7 +259,7 @@ fn anonsocket_write<'tcx>(
         // Remember this clock so `read` can synchronize with us.
         ecx.release_clock(|clock| {
             writebuf.clock.join(clock);
-        });
+        })?;
         // Do full write / partial write based on the space available.
         let write_size = len.min(available_space);
         let actual_write_size = ecx.write_to_host(&mut writebuf.buf, write_size, ptr)?.unwrap();


genmc-sys/cpp/src/MiriInterface/EventHandling.cpp
```diff
@@ -89,6 +89,72 @@
...
+    MemOrdering store_ordering,
+    RMWBinOp rmw_op,
+    GenmcScalar rhs_value,
+    GenmcScalar old_val
+) -> ReadModifyWriteResult {
+    const auto addr = SAddr(address);
+    // `type` is only used for printing.
+    const auto type = AType::Unsigned;
+
+    const auto rhsVal = GenmcScalarExt::to_sval(rhs_value);
```
Patrick-6: These can be inlined.

src/concurrency/genmc/helper.rs
```diff
@@ -95,3 +98,55 @@ impl AtomicWriteOrd {
...
+/// Convert a possibly signed Miri min/max operation to its GenMC counterpart.
+pub(super) fn min_max_to_genmc_rmw_op(min: bool, is_signed: bool) -> RMWBinOp {
+    // FIXME(genmc): is there a use for FMin/FMax? GenMC has (Min, UMin, FMin)
+    match (min, is_signed) {
+        (true, true) => RMWBinOp::Min,
+        (false, true) => RMWBinOp::Max,
+        (true, false) => RMWBinOp::UMin,
+        (false, false) => RMWBinOp::UMax,
+    }
+}
```
Patrick-6: I just realized that GenMC no longer supports atomic FMin and FMax, but Rust doesn't seem to support atomic floats either, so my comment here can probably be removed.

src/concurrency/genmc/mod.rs
```diff
@@ -125,22 +130,19 @@ impl GenmcCtx {
 
     /// Get the number of blocked executions encountered by GenMC.
     pub fn get_blocked_execution_count(&self) -> u64 {
-        let mc = self.handle.borrow();
-        mc.as_ref().unwrap().get_blocked_execution_count()
+        self.handle.borrow().get_blocked_execution_count()
```
Patrick-6, line 133: This makes use of `UniquePtr`s [`impl Deref`](https://docs.rs/cxx/1.0.175/src/cxx/unique_ptr.rs.html#154-169), which can panic if the `UniquePtr` is `null`, but we never move it, so it should always be non-null.

Ideally, this would be some kind of `NonNullUniquePtr`, but CXX does not support anything like that currently.

The new code variant here at least prevents having to manually `unwrap` everywhere.

src/concurrency/genmc/mod.rs
```diff
@@ -317,39 +330,48 @@ impl GenmcCtx {
...
+        is_signed: bool,
+        rhs_scalar: Scalar,
+        old_value: Scalar,
     ) -> InterpResult<'tcx, (Scalar, Option<Scalar>)> {
-        assert!(
-            !self.get_alloc_data_races(),
-            "atomic min/max operation with data race checking disabled."
-        );
-        throw_unsup_format!("FIXME(genmc): Add support for atomic min/max.")
+        // FIXME(genmc): can `rhs_scalar` be a pointer? Should this be allowed?
```
Patrick-6: `AtomicPtr` doesn't have `fetch_{min,max}` methods, so should either arguments here be prohibited from being a pointer?

src/concurrency/genmc/mod.rs
```diff
@@ -575,9 +597,9 @@ impl GenmcCtx {
         let genmc_parent_tid = thread_infos.get_genmc_tid(curr_thread_id);
         let genmc_new_tid = thread_infos.add_thread(new_thread_id);
 
-        let mut mc = self.handle.borrow_mut();
-        mc.as_mut().unwrap().handle_thread_create(genmc_new_tid, genmc_parent_tid);
+        self.handle.borrow_mut().pin_mut().handle_thread_create(genmc_new_tid, genmc_parent_tid);
 
+        // TODO GENMC (ERROR HANDLING): can this ever fail?
         interp_ok(())
```
Patrick-6: (Leftover comment from a rebase) I don't think thread creation can ever fail? At least GenMC doesn't give an error in any case.

src/concurrency/genmc/mod.rs
```diff
@@ -762,6 +777,65 @@ impl GenmcCtx {
...
+            genmc_rhs_scalar,
+            genmc_old_value,
+        );
+
+        if let Some(error) = rmw_result.error.as_ref() {
+            // FIXME(genmc): error handling
+            throw_ub_format!("{}", error.to_string_lossy());
+        }
+
+        // FIXME(genmc): Check that both RMW arguments have sane provenance.
```
Patrick-6: We cannot check if the rhs is a pointer before calling into GenMC, since we only know which value will be read at this point.
(We can't yet send pointers to GenMC with this PR, so I've left out the code that will go here.)

tests/genmc/fail/loom/buggy_inc.rs
Patrick-6, line 1: [Loom is licensed under MIT](https://github.com/tokio-rs/loom/blob/dbf32b04bae821c64be44405a0bb72ca08741558/LICENSE), so we should be able to use its test.

tests/genmc/pass/atomics/rmw_edge_cases.rs
Patrick-6, line 1: Initially I had 10 separate tests here (using revisions), one per type, but I think this macro approach is cleaner.

tests/utils/genmc.rs
```diff
@@ -51,6 +51,14 @@ pub unsafe fn join_pthread(thread_id: pthread_t) {
...
 }
 
+/// Spawn `N` threads using `pthread_create` without any arguments, abort the process on any errors.
+pub unsafe fn create_pthreads_no_params<const N: usize>(
+    functions: [extern "C" fn(*mut c_void) -> *mut c_void; N],
+) -> [pthread_t; N] {
+    let value = std::ptr::null_mut();
+    functions.map(|func| spawn_pthread(func, value))
+}
+
```
Patrick-6: Not sure if `create_*` is a good name, maybe `spawn_*`?

Patrick-6:
r? @RalfJung 

Next part of merging GenMC mode, but a bit less code this time ^^

src/concurrency/genmc/mod.rs
```diff
@@ -317,39 +330,48 @@ impl GenmcCtx {
...
-        _min: bool,
-        _is_signed: bool,
-        _rhs_scalar: Scalar,
-        _old_value: Scalar,
+        ecx: &InterpCx<'tcx, MiriMachine<'tcx>>,
+        address: Size,
+        size: Size,
+        ordering: AtomicRwOrd,
+        min: bool,
+        is_signed: bool,
```
RalfJung: This screams for an enum representing the operation, rather than two `bool` flags.
```rust
enum MinMaxOp {
  Min { signed: bool },
  Max { signed: bool },
}
```

src/concurrency/genmc/mod.rs
RalfJung: Yeah there can't be any pointers here.

src/concurrency/genmc/mod.rs
```diff
@@ -295,19 +298,29 @@ impl GenmcCtx {
...
+        rhs_scalar: Scalar,
+        old_value: Scalar,
     ) -> InterpResult<'tcx, (Scalar, Option<Scalar>)> {
-        assert!(
-            !self.get_alloc_data_races(),
-            "atomic read-modify-write operation with data race checking disabled."
-        );
-        throw_unsup_format!("FIXME(genmc): Add support for atomic RMW.")
+        // Ensure the the right argument is never a pointer.
+        if matches!(rhs_scalar, Scalar::Ptr(..)) {
```
RalfJung: `atomic_rmw_op` in `atomic.rs` already checks these things, no reason to repeat such assertions here.

src/concurrency/genmc/mod.rs
RalfJung: In fact we should really refactor this a bit on the Miri side, splitting out min/max like this makes little sense. I'll prepare a PR for that.

src/concurrency/genmc/mod.rs
RalfJung: I don't understand what you even want to check here.

genmc-sys/cpp/src/MiriInterface/EventHandling.cpp
```diff
@@ -89,6 +89,72 @@
...
+
+    const auto* ret_val = std::get_if<SVal>(&load_ret);
+    if (nullptr == ret_val) {
+        ERROR("Unimplemented: read-modify-write returned unexpected result.");
+    }
+    const auto read_old_val = *ret_val;
+    const auto new_value = executeRMWBinOp(read_old_val, rhsVal, size, rmw_op);
+
+    const auto storePos = inc_pos(thread_id);
+    const auto store_ret = GenMCDriver::handleStore<EventLabel::EventLabelKind::FaiWrite>(
```
RalfJung, line 134: So an RMW is just a load followed by a store? Doesn't this need some kind of marker that ties them together because RMW behave different than a normal load-store sequence?

genmc-sys/src/lib.rs
```diff
@@ -120,6 +132,19 @@ mod ffi {
         Debug3ReadsFrom,
     }
 
+    #[derive(Debug)]
+    /// Setting to control which execution graphs GenMC prints after every execution.
+    enum ExecutiongraphPrinting {
+        /// Print no graphs.
+        None,
```
RalfJung, line 136: Can you use attributes here or does cxx not like then? If you can, you can avoid the manual `Default` impl by using https://doc.rust-lang.org/std/default/trait.Default.html#enums

genmc-sys/src/lib.rs
```diff
@@ -188,6 +226,21 @@ mod ffi {
         SequentiallyConsistent = 6,
     }
 
+    #[derive(Debug)]
+    enum RMWBinOp {
```
RalfJung, line 228: Does this correspond to anything in genmc or can we choose arbitrary values here?

src/concurrency/genmc/helper.rs
```diff
@@ -95,3 +98,55 @@ impl AtomicWriteOrd {
...
+            AtomicFenceOrd::AcqRel => MemOrdering::AcquireRelease,
+            AtomicFenceOrd::SeqCst => MemOrdering::SequentiallyConsistent,
+        }
+    }
+}
+
+impl AtomicRwOrd {
+    /// Split up the atomic success ordering of a read-modify-write operation into GenMC's representation.
+    /// Note that both returned orderings are currently identical, because this is what GenMC expects.
+    pub(super) fn to_genmc_memory_orderings(self) -> (MemOrdering, MemOrdering) {
```
RalfJung:
```suggestion
    pub(super) fn to_genmc(self) -> (MemOrdering, MemOrdering) {
```

src/concurrency/genmc/helper.rs
```diff
@@ -95,3 +98,55 @@ impl AtomicWriteOrd {
...
+            AtomicFenceOrd::Release => MemOrdering::Release,
+            AtomicFenceOrd::AcqRel => MemOrdering::AcquireRelease,
+            AtomicFenceOrd::SeqCst => MemOrdering::SequentiallyConsistent,
+        }
+    }
+}
+
+impl AtomicRwOrd {
+    /// Split up the atomic success ordering of a read-modify-write operation into GenMC's representation.
+    /// Note that both returned orderings are currently identical, because this is what GenMC expects.
```
RalfJung: If genmc expects a single ordering then it'd make more sense to just return a single ordering here.

src/concurrency/genmc/helper.rs
```diff
@@ -95,3 +98,55 @@ impl AtomicWriteOrd {
...
+        (true, false) => RMWBinOp::UMin,
+        (false, false) => RMWBinOp::UMax,
+    }
+}
+
+/// Convert a possibly negated binary operation to its GenMC counterpart.
+pub(super) fn to_genmc_rmw_op(bin_op: rustc_middle::mir::BinOp, negate: bool) -> RMWBinOp {
+    match bin_op {
+        rustc_middle::mir::BinOp::Add => RMWBinOp::Add,
+        rustc_middle::mir::BinOp::Sub => RMWBinOp::Sub,
```
RalfJung: Here you are silently ignoring `negate`... that should never be set but better safe than sorry.

src/concurrency/genmc/mod.rs
RalfJung: In reality it can fail e.g. when the kernel runs out of memory... but in Miri it can't fail.

genmc-sys/cpp/src/MiriInterface/EventHandling.cpp
```diff
@@ -89,6 +89,72 @@
...
+    GenmcScalar rhs_value,
+    GenmcScalar old_val
+) -> ReadModifyWriteResult {
+    const auto addr = SAddr(address);
+    // `type` is only used for printing.
+    const auto type = AType::Unsigned;
+
+    const auto rhsVal = GenmcScalarExt::to_sval(rhs_value);
+
+    const auto load_ret = handle_load_reset_if_none<EventLabel::EventLabelKind::FaiRead>(
```
RalfJung, line 112: Why `FaiRead`? That seems specific to fetch-and-increment?

src/concurrency/genmc/run.rs
```diff
@@ -32,8 +34,15 @@ pub fn run_genmc_mode<'tcx>(
...
 
         // Execute the program until completion to get the return value, or return if an error happens:
-        // FIXME(genmc): add an option to allow the user to see the GenMC output message when the verification is done.
-        let return_code = eval_entry(genmc_ctx.clone())?;
+        let Some(return_code) = eval_entry(genmc_ctx.clone()) else {
+            // If requested, print the output GenMC produced:
+            if genmc_config.print_genmc_output {
+                eprintln!("=============================================");
+                eprintln!("{}", genmc_ctx.get_result_message());
+                eprintln!("=============================================");
```
RalfJung:
```suggestion
                eprintln!("== raw GenMC output =========================");
                eprintln!("{}", genmc_ctx.get_result_message());
                eprintln!("== end of raw GenMC output ==================");
```

src/concurrency/sync.rs
```diff
@@ -716,6 +716,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         callback: DynUnblockCallback<'tcx>,
     ) {
         let this = self.eval_context_mut();
+        if this.machine.data_race.as_genmc_ref().is_some() {
+            unimplemented!("Futexes are currently not supported in GenMC mode");
+        }
```
RalfJung: I'd prefer if we didn't have to put this in so many different places.

It should be enough to put the error in [these two](https://github.com/rust-lang/miri/blob/68b74cbbc6fec64d2a529432d21f1e44029738d2/src/concurrency/data_race.rs#L962-L981) functions and then ensure that we always call those (i.e., reduce the use of `as_vclocks_ref` elsewhere in Miri).

Example diff:
```diff
-        if let Some(data_race) = this.machine.data_race.as_vclocks_ref() {
-            data_race.acquire_clock(&rwlock.clock_unlocked, &this.machine.threads);
-        }
+        this.acquire_clock(&rwlock.clock_unlocked)?;
```
```diff
-        if let Some(data_race) = this.machine.data_race.as_vclocks_ref() {
-            // Add this to the shared-release clock of all concurrent readers.
-            data_race.release_clock(&this.machine.threads, |clock| {
-                rwlock.clock_current_readers.join(clock)
-            });
-        }
+        // Add this to the shared-release clock of all concurrent readers.
+        this.release_clock(|clock| rwlock.clock_current_readers.join(clock));
```

I'd suggest doing this in a separate PR.

tests/genmc/fail/loom/buggy_inc.rs
```diff
@@ -0,0 +1,63 @@
+//@compile-flags: -Zmiri-genmc -Zmiri-disable-stacked-borrows
+
+// This tests is the test `checks_fail` from loom/test/smoke.rs adapted for Miri-GenMC.
```
RalfJung:
```suggestion
// This is the test `checks_fail` from loom/test/smoke.rs adapted for Miri-GenMC.
```

tests/genmc/fail/loom/buggy_inc.rs
RalfJung: MIT != MIT+Apache2.

Please add:
```
// SPDX-License-Identifier: MIT
// SPDX-FileCopyrightText: Copyright (c) 2019 Carl Lerche
```

RalfJung (commented):
First round of feedback

tests/utils/genmc.rs
RalfJung: Keep the name consistent with the other functions -- so `spawn` is better indeed.

tests/utils/genmc.rs
```diff
@@ -51,6 +51,14 @@ pub unsafe fn join_pthread(thread_id: pthread_t) {
     }
 }
 
+/// Spawn `N` threads using `pthread_create` without any arguments, abort the process on any errors.
+pub unsafe fn create_pthreads_no_params<const N: usize>(
+    functions: [extern "C" fn(*mut c_void) -> *mut c_void; N],
+) -> [pthread_t; N] {
+    let value = std::ptr::null_mut();
```
RalfJung: This can be inlined

tests/genmc/pass/atomics/rmw_simple.rs
```diff
@@ -0,0 +1,30 @@
...
+    if x != y {
+        std::process::abort();
+    }
+}
+
+#[unsafe(no_mangle)]
+fn miri_start(_argc: isize, _argv: *const *const u8) -> isize {
+    VALUE.store(1, ORD);
+
+    // Try all RMW operations:
```
RalfJung: These are not all RMW operations...
- ~~compare_exchange(_weak)~~
- or
- xor
- and
- nand
- signed min/max

tests/genmc/pass/atomics/rmw_simple.rs
RalfJung, line 1: Why is this test separate from `tests/genmc/pass/atomics/rmw_edge_cases.stderr`?

Seems best to have one test that covers all operations on all AtomicInt types.

tests/genmc/pass/litmus/IRIWish.rs
```diff
@@ -0,0 +1,40 @@
...
+            let _r1 = X.load(Relaxed);
+            std::sync::atomic::fence(AcqRel);
+            let _r2 = Y.load(Relaxed);
+        });
+        spawn_pthread_closure(|| {
+            let _r1 = Y.load(Relaxed);
+            std::sync::atomic::fence(AcqRel);
+            let _r2 = X.load(Relaxed);
+        });
+        0
```
RalfJung, line 64: This seems like a good test to check the values that are being loaded.

RalfJung (commented):
And now I also went over the tests

src/concurrency/genmc/mod.rs
Patrick-6: I think we discussed at one point that (except for atomic `swap` and `compare_exchange`), no RMW operation should have a RHS-argument that has provenance.

The implementation of provenance in my complete branch (on the GenMC side) always uses the provenance of the LHS argument, RHS provenance is lost.
[The check I have on my branch](https://github.com/Patrick-6/miri/blob/47cb38a8b472d5ed1f8870e70bc79cd989ec78b2/src/concurrency/genmc/mod.rs#L900-L920) prevents this from happening unnoticed.

src/concurrency/sync.rs
Patrick-6: Should I drop the commit adding the futex errors?

tests/genmc/fail/loom/buggy_inc.rs
Patrick-6: I modified more than half the lines in the test and added comments compared to [the original](https://github.com/tokio-rs/loom/blob/dbf32b04bae821c64be44405a0bb72ca08741558/tests/smoke.rs) (to make it work with Miri-GenMC). This should be fine, right?

src/concurrency/genmc/mod.rs
RalfJung: This is all already checked by earlier layers. It's not worth checking again here (and definitely not worth giving nice errors since they can't be reached).

Matching on `Scalar` is a highly unusual operation; this code is not special enough to justify going so far down in the Miri stack of abstractions.

In fact I realized the checks in the earlier layers can be tightened due to some recent refactoring; https://github.com/rust-lang/miri/pull/4570 does that.

src/concurrency/sync.rs
RalfJung: Yes, please do.

tests/genmc/fail/loom/buggy_inc.rs
RalfJung: There's no harm in just adding the comment.

tests/genmc/pass/atomics/rmw_simple.rs
Patrick-6: Yeah I can put all operations into one test.

Ideally we would also have this test, but this is currently bugged with GenMC (my patch has not yet been merged):
```rust
let x = AtomicU8::new(0);
x.store(150, SeqCst); // 150
x.fetch_add(150, SeqCst); // 300 % 256 == 44
x.fetch_max(200, SeqCst); // should be 200, but is currently 300 (atomic values are stored as 64 bit ints)

let a = x.load(SeqCst); // returns 44 instead of 200
if a != 200 {
    std::process::abort();
}
```
I don't think we can do anything about this bug on the Miri side sadly.

RalfJung: Interesting that your corner-case-tests still work despite this.

You can add this in commented-out form with a FIXME in that test. Is there a genmc issue tracker we can reference?

genmc-sys/cpp/src/MiriInterface/EventHandling.cpp
Patrick-6: The `EventLabel::EventLabelKind::FaiWrite` is the marker (FAI = Fetch-And-Increment, due to historical reasons). This parameter makes GenMC create special load/store labels (`FaiReadLabel` and `FaiWriteLabel` in GenMCs codebase).

genmc-sys/src/lib.rs
Patrick-6: Yes, [this is from GenMC](https://github.com/Patrick-6/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/src/Support/RMWOps.hpp#L25-L38).

This type still corresponds to [this comment](https://github.com/rust-lang/miri/blob/8c8a1d69e05e422bb11ad325ddc2dc994fe3dce3/genmc-sys/src/lib.rs#L204-L242), but maybe there should be more warning that these should absolutely not be changed unless GenMC changes them.

genmc-sys/cpp/src/MiriInterface/EventHandling.cpp
Patrick-6: Yes, apparently it is called that for historical reasons

tests/genmc/pass/atomics/rmw_simple.rs
Patrick-6: No, it's only a PR on a private repo

Patrick-6: > Interesting that your corner-case-tests still work despite this.

It's not actually surprising, I've chosen the values in that test such that they work for all bit widths from 8 to 64 (so I can use the same values for all types), which avoided this specific edge case.

genmc-sys/cpp/src/MiriInterface/EventHandling.cpp
RalfJung: Interesting...

The comment [here](https://github.com/MPI-SWS/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/src/ExecutionGraph/EventLabel.hpp#L773) however says that this is *not* used for compare-and-swap? That has a separate `CasReadLabel`, it seems. There's also `CasWriteLabel`.

genmc-sys/src/lib.rs
RalfJung: You can just add the usual `/// Corresponds to GenMC's type with the same name.`.

The `/// Should only be modified if changed by GenMC.` parts can probably be removed.

> `Types shared between Miri/Rust and GenMC/C++ through cxx_bridge:`

I don't quite get this comment since all types above are also shared between Rust and C++.

genmc-sys/src/lib.rs
Patrick-6: No, those are not supported currently.

RalfJung: Bummer.

genmc-sys/src/lib.rs
```diff
@@ -53,7 +53,11 @@ impl GenmcScalar {
 
 impl Default for GenmcParams {
     fn default() -> Self {
-        Self { print_random_schedule_seed: false, do_symmetry_reduction: false }
+        Self {
+            print_random_schedule_seed: false,
+            do_symmetry_reduction: false,
+            print_execution_graphs: ExecutiongraphPrinting::default(),
```
```suggestion
            // GenMC graphs can be quite large since Miri produces a lot of (non-atomic) events.
            // GenMC also does not print graphs by default.
            print_execution_graphs: ExecutiongraphPrinting::None,
```
I don't think it's worth having a default impl just for this. Seeing all the default params in one place is nice as well.

genmc-sys/src/lib.rs
Patrick-6: > > `Types shared between Miri/Rust and GenMC/C++ through cxx_bridge:`
> 
> I don't quite get this comment since all types above are also shared between Rust and C++.

The ones on top are all inside the Miri repo (just shared with the C++ side), we can modify them however we want.

The ones at the bottom are between shared with GenMC, these cannot be changed easily

genmc-sys/cpp/src/MiriInterface/EventHandling.cpp
Patrick-6: Yeah CAS uses different labels, support for that will be in one of the next PRs

RalfJung: CAS and compare-exchange are the same thing though, and you have a comment saying that this function can be used with compare-exchange? I am confused.

src/concurrency/genmc/mod.rs
```diff
@@ -762,6 +777,65 @@ impl GenmcCtx {
         interp_ok(store_result.is_coherence_order_maximal_write)
     }
 
+    /// Inform GenMC about an atomic read-modify-write operation.
+    /// For GenMC, compare-exchange and atomic-swap are also RMW (see `RMWBinOp` for full list of operations).
```
```suggestion
    /// This includes atomic swap (also often called "exchange"), but does *not*
    /// include compare-exchange (see `RMWBinOp` for full list of operations).
```

genmc-sys/src/lib.rs
```diff
@@ -163,6 +188,19 @@ mod ffi {
...
+        error: UniquePtr<CxxString>,
+        /// The value that was read by the RMW operation as the left operand.
+        old_value: GenmcScalar,
+        /// The value that was produced by the RMW operation.
+        new_value: GenmcScalar,
+        /// `true` if the write should also be reflected in Miri's memory representation.
+        is_coherence_order_maximal_write: bool,
+    }
+
     /**** Types shared between Miri/Rust and GenMC/C++ through cxx_bridge: ****/
```
```suggestion
    /**** These are GenMC types that we have to copy-paste here since cxx does not support
      "importing" externally defined C++ types. ****/
```

genmc-sys/src/lib.rs
RalfJung: For the code that links Rust and LLVM (inside rustc codegen), we avoid this problem by having our own version of these enums and then having C++ code that translates between the two. That may also be a good idea here? I am slightly nervous that future versions of GenMC could change some of these values and then things could go wrong in really weird ways.

genmc-sys/cpp/src/MiriInterface/EventHandling.cpp
RalfJung: I've found the confusing comment and proposed an update.

RalfJung, line 112:
```suggestion
    // Somewhat confusingly, the GenMC term for RMW read/write labels is
    // `FaiRead` and `FaiWrite`.
    const auto load_ret = handle_load_reset_if_none<EventLabel::EventLabelKind::FaiRead>(
```

src/concurrency/genmc/mod.rs
RalfJung: Please rebase over latest master to include my PR. Sorry for the merge conflicts, but overall it should make your code a bit simpler.

genmc-sys/src/lib.rs
RalfJung, line 228: FWIW under the Rust naming convention this type should be called `RmwBinOp`.

Patrick-6: > we avoid this problem by having our own version of these enums and then having C++ code that translates between the two
Doesn't that have the same problem, but between the C++ code and LLVM?

I don't think there should be weird issues from such changes, since cxx_bridge inserts many static assertions, e.g., for `ActionKind`, the generated code looks like this:
```C++
static_assert(::std::is_enum<ActionKind>::value, "expected enum");
static_assert(sizeof(ActionKind) == sizeof(::std::uint8_t), "incorrect size");
static_assert(static_cast<::std::uint8_t>(ActionKind::Load) == 0, "disagrees with the value in #[cxx::bridge]");
static_assert(static_cast<::std::uint8_t>(ActionKind::NonLoad) == 1, "disagrees with the value in #[cxx::bridge]");
```

One thing that might be a problem is if any public GenMC function ever swaps the order of some arguments with the same type (or types convertible into each other). I can't really see how something like that could be caught though.

Patrick-6: Yeah, that name comes from GenMC. We could typedef it to the proper spelling?

The issue is that the entire `ffi::*` is public (since there are so many functions and types in there). In that case both the renamed and original would be visible.

RalfJung: Ah, I missed that these are still "extern shared" types... apparently only extern shared structs got deprecated, not extern shared enums.

genmc-sys/src/lib.rs
```diff
@@ -205,6 +258,7 @@ mod ffi {
         /**** Types shared between Miri/Rust and GenMC/C++: ****/
```
```suggestion
        /**** Types shared between Miri/Rust and GenMC/C++:
            (This tells cxx that the enums defined above are already defined on the C++ side;
            it will emit assertions to ensure that the two definitions agree.) ****/
```

genmc-sys/src/lib.rs
RalfJung: > Yeah, that name comes from GenMC.

Right... yeah not worth renaming then.

genmc-sys/build.rs
```diff
@@ -184,13 +184,11 @@ fn compile_cpp_dependencies(genmc_path: &Path) {
...
-    let cmakelists_path = genmc_path.join("CMakeLists.txt");
-
     // FIXME(genmc,cargo): Switch to using `CARGO_CFG_DEBUG_ASSERTIONS` once https://github.com/rust-lang/cargo/issues/15760 is completed.
     // Enable/disable additional debug checks, prints and options for GenMC, based on the Rust profile (debug/release)
     let enable_genmc_debug = matches!(std::env::var("PROFILE").as_deref().unwrap(), "debug");
 
-    let mut config = cmake::Config::new(cmakelists_path);
+    let mut config = cmake::Config::new(genmc_path);
     config
         .always_configure(false) // We don't need to reconfigure on subsequent compilation runs.
```
RalfJung: Ah I forgot about this in the last PR... I think we should only set this to `false` when *not* using the downloaded genmc.

genmc-sys/src/lib.rs
```diff
@@ -97,6 +98,8 @@ mod ffi {
         pub print_random_schedule_seed: bool,
         pub do_symmetry_reduction: bool,
         pub print_execution_graphs: ExecutiongraphPrinting,
+        /// Enabling this will set the memory model used by GenMC to "Sequential Consistency" (SC).
+        pub disable_weak_memory_emulation: bool,
```
RalfJung, line 103: Instead of duplicating the field, can you pass in the main miri config object?

tests/genmc/pass/litmus/Z6_U.weak.stderr
RalfJung, line 1: It seems like every weak behavior can also occur with SC. Is that expected?

tests/genmc/pass/atomics/rmw_simple.rs
RalfJung: Why did you mark this as resolved?

I asked to add the test you mentioned above with a FIXME to ensure we have it recorded somewhere in this repo. I can't find that in the PR.

RalfJung: FWIW it should be entirely possible to hit that case with a test where values are the same for all bitwidths:
```rust
let x = <$atomic>::new(0);
x.store(10, SeqCst);
x.fetch_add(<$int>::MAX, SeqCst); // definitely overflows, so new value of x is smaller than 10
x.fetch_max(10, SeqCst); // new value of x should be 10
let a = x.load(SeqCst);
if a != 10 {
    std::process::abort();
}
```
The reason your test doesn't this this is that it doesn't do "overflowing add" followed by "max" (it uses the opposite order).

genmc-sys/src/lib.rs
Patrick-6: No, since this `struct GenmcParam` is shared with the C++ side, it cannot contain anything not supported by CXX, such as `Option` (which `MiriConfig` contains).

tests/genmc/pass/litmus/Z6_U.weak.stderr
Patrick-6: It is consistent with running the C version of the same test with GenMC and printing its output (with a different output order though).

In this case, disallowing weak behaviors removed 4 possible executions, but these had identical outputs as some non-weak executions. There are only 8 possible outputs for 22 and 18 total execs., so there must be duplicate outputs.

For example, reading from Y in t2, then writing to Y from t1 then from t3 compared to reversing the order of t3 and t1 are two different executions, but the same output.

If you want to see the exact executions that are disallowed, you can run with the flag `-Zmiri-genmc-print-exec-graphs`.

genmc-sys/build.rs
Patrick-6: FWIU this should be ok:
- First compilation: configure will be executed by cmake-rs
- Nth compilation: cmake runs configure itself if required (i.e., environment or `CMakeLists.txt` changed)

genmc-sys/src/lib.rs
RalfJung: Hm... but I think we can do better than have all this genmc logic in `miri.rs`. Maybe that block (inside the `if let Some(genmc_ctx) = &mut miri_config.genmc_config {`) should be moved to a function in `genmc/config.rs`?

tests/genmc/pass/litmus/Z6_U.weak.stderr
RalfJung: Is there some way to differentiate those in the output, e.g. by printing the final value for X and Y?

tests/genmc/pass/litmus/Z6_U.rs
```diff
@@ -1,35 +1,63 @@
...
+            }),
+            spawn_pthread_closure(|| {
+                Y.store(3, SeqCst);
+                b = X.load(SeqCst);
+            }),
+        ];
+        join_pthreads(ids);
+
+        // Print the values to check that we get all of them:
+        if writeln!(MiriStderr, "({a}, {b})").is_err() {
```
RalfJung:
```suggestion
        if writeln!(MiriStderr, "a={a}, b={b}").is_err() {
```

genmc-sys/build.rs
RalfJung: What is the downside to `always_configure(true)`? I suppose the cmake crate made that the default for a reason.

tests/genmc/pass/litmus/Z6_U.weak.stderr
Patrick-6: In general, no, for example, two threads running the following closure each would yield the exact same output, but their order gives two possible executions (t1 goes first, or t2 goes first):
```Rust
|| {
  x.fetch_add(10, SeqCst);
}
```
In the case of the Z6_U test, printing `(a,b,X,Y)` yields 15 vs 12 unique results (RC11 vs SC) compared to 8 for both from just printing `(a,b)`.

genmc-sys/build.rs
Patrick-6: Always configuring makes the [build slower](https://github.com/rust-lang/cmake-rs/blob/b59be3b5cce397de4018aacd51501f925ce22ecd/src/lib.rs#L396-L403). I don't know why `true` it is the default, but maybe there were some issues with older versions of cmake not handling this properly?
When I use cmake by hand, I never run the configure step manually, only the build step, unless I want to change the build settings or after having completely deleted the build directory.

tests/genmc/pass/litmus/Z6_U.weak.stderr
RalfJung: Right, I wasn't asking in general, I was asking about this particular test :)

genmc-sys/build.rs
RalfJung: Yeah but how much slower?

For all we know, the build settings changed since the last time cmake was run. So please make it reconfigure for the default cause of using the downloaded source so we minimize the risk of people getting weird issues due to outdated configs or whatnot.

Patrick-6: It adds ~3 seconds every time `./miri check`, `clippy`, `build`, `run` or `test` runs after a change to `genmc-sys/`.

tests/genmc/pass/litmus/Z6_U.rs
Patrick-6: I think this is easier to visually compare the lines compared to `(a={a}, b={b}, X=...)`:
```rust
let result = (a, b, X.load(Relaxed), Y.load(Relaxed));
if writeln!(MiriStderr, "(a,b,X,Y)={result:?}").is_err() {
```

genmc-sys/build.rs
RalfJung: 3s is low enough (that folder won't change *that* often), seems worth it to ensure an up-to-date config.

src/concurrency/genmc/config.rs
```diff
@@ -69,4 +70,39 @@ impl GenmcConfig {
     pub fn disable_weak_memory_emulation(&mut self) {
         self.params.disable_weak_memory_emulation = true;
     }
```
RalfJung: Now you can remove this method and just modify the field directly.

tests/genmc/pass/litmus/Z6_U.rs
RalfJung: Hm, don't think I agree. As the number of variables increases the tuple becomes harder and harder to read.

genmc-sys/build.rs
Patrick-6: I don't really agree that this extra time doesn't matter, it can make actively working on `genmc-sys` more annoying (note that this extra time also applies to changing any Rust files in that directory), especially trying to quickly iterate on changes (e.g., make change, run `./miri run`) or quickly debug print something: every tiny change costs 3 seconds extra. 

cmake already runs the configure step itself if _anything_ about the config changes, even just the timestamp on any `CMakeLists.txt` or any build parameter.
If the source directory changes, cmake-rs handles it [here](https://github.com/rust-lang/cmake-rs/blob/b59be3b5cce397de4018aacd51501f925ce22ecd/src/lib.rs#L991-L999) by deleting the build directory.

I've personally never had an issue with an inconsistent config before, even after switching branches or rebasing.

But, I can change it if you want, I just think it will only have drawbacks🤷‍♂️

tests/genmc/pass/litmus/Z6_U.rs
Patrick-6: I don't even know if we need to print the variable names in the output at all, since it is mainly meant for the UI test to ensure we get everything, not to be human readable. (Also, any printed char makes the test a bit slower ;) )

tests/genmc/pass/atomics/rmw_simple.rs
Patrick-6: FIXME is added now

tests/genmc/pass/atomics/rmw_simple.rs
Patrick-6: Extra tests are added now (`rmw_ops.rs`)

tests/genmc/pass/litmus/Z6_U.rs
RalfJung: I think it's quite useful to be able to look at the file and understand what it says about the possible behaviors of the program.

genmc-sys/build.rs
RalfJung: I assumed you'd use local sources instead of download mode anyway when actually working on this code so you wouldn't get any extra build time then?

Patrick-6: I can see someone work on the C++ or Rust code in `genmc-sys/`, but without modifying the GenMC code, that would work without using a local GenMC repo.
I guess we could just warn against working in `genmc-sys/` without local GenMC in the `doc/genmc.md`.

RalfJung: I don't think a 3s wait time is bad enough to justify a warning.

Can the download-genmc logic determine whether it *changed* anything, or whether the intended commit was already checked out before? Then we could do the reconfigure only when the commit changed.

src/concurrency/data_race.rs
```diff
@@ -961,6 +960,10 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
     /// The closure will only be invoked if data race handling is on.
     fn release_clock<R>(&self, callback: impl FnOnce(&VClock) -> R) -> Option<R> {
         let this = self.eval_context_ref();
+        assert!(
+            this.machine.data_race.as_genmc_ref().is_none(),
+            "release_clock is not supported in GenMC mode."
```
RalfJung:
```suggestion
            "this operation performs synchronization that is not supported in GenMC mode"
```

src/concurrency/data_race.rs
```diff
@@ -973,7 +976,7 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
         match &this.machine.data_race {
             GlobalDataRaceHandler::None => {}
             GlobalDataRaceHandler::Genmc(_genmc_ctx) =>
-                throw_unsup_format!("acquire_clock is not (yet) supported in GenMC mode."),
+                throw_unsup_format!("acquire_clock is not supported in GenMC mode."),
```
RalfJung:
```suggestion
                throw_unsup_format!("this operation performs synchronization that is not supported in GenMC mode"),
```

src/concurrency/data_race.rs
```diff
@@ -961,6 +960,10 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
     /// The closure will only be invoked if data race handling is on.
     fn release_clock<R>(&self, callback: impl FnOnce(&VClock) -> R) -> Option<R> {
         let this = self.eval_context_ref();
+        assert!(
```
RalfJung:
```suggestion
        // FIXME: make this a proper error instead of ICEing the interpreter.
        assert!(
```
(This is not a request to do this change in this PR, just to leave a note so it's clear in the code that this is not the final form of this function)

tests/genmc/pass/atomics/rmw_ops.rs
```diff
@@ -0,0 +1,91 @@
...
+        // SWAP
+        x.store(!0, ORD);
+        assert_eq(!0, x.swap(pattern, ORD));
+        assert_eq(pattern, x.load(ORD));
+
+        // Check correct behavior of atomic min/max combined with overflowing add/sub.
+        x.store(10, ORD);
+        assert_eq(10, x.fetch_add(<$int>::MAX, ORD)); // definitely overflows, so new value of x is smaller than 10
+        assert_eq(<$int>::MAX.wrapping_add(10), x.fetch_max(10, ORD)); // new value of x should be 10
+        // assert_eq(10, x.load(ORD)); // FIXME(genmc,BUG): enable this check once GenMC correctly handles min/max truncation.
```
RalfJung:
```suggestion
        // assert_eq(10, x.load(ORD)); // FIXME(genmc,#4572): enable this check once GenMC correctly handles min/max truncation.
```

RalfJung:
LGTM, please squash. :)
@rustbot author

RalfJung:
Thanks! I have added a patch for the release/acquire clock error messages.

PR #4571, opened 2025-09-05
Enable MacOS support for GenMC mode.

Patrick-6:
With https://github.com/dtolnay/cxx/issues/1535 fixed, we may be able to support GenMC mode on MacOS now!


RalfJung:
Looks like the macos runners don't have LLVM development headers preinstalled. I also have no idea how to install them there.^^

This is fairly low priority compared to the rest of your thesis. :)

RalfJung:
I'll close this and take care of it in a new PR, based on this patch. Thanks for the help!

PR #4575, opened 2025-09-10
weak memory tests: add more info on where they come from, and test list of behaviors more thoroughly

RalfJung:
@cbeuw does this sound right?


tests/pass/0weak_memory_consistency.rs
```diff
@@ -56,6 +49,7 @@ fn spin_until_bool(loc: &AtomicBool, ord: Ordering, val: bool) -> bool {
     val
 }
 
+/// Test matching https://www.doc.ic.ac.uk/~afd/homepages/papers/pdfs/2017/POPL.pdf, Figure 7
```
RalfJung: We also have a test for Figure 8, but it is a test of opposite polarity -- it ensure that we do see a particular weak behavior, but does not exclude that we don't see the "wrong" behavior demonstrated in that figure.

Is there some way we could test that?

tests/pass/0weak_memory/weak.rs
RalfJung, line 1: I've refactored these tests so we now check the list of all outcomes, instead of just checking whether a particular outcome occurs.

RalfJung, line 1: @cbeuw If I understood you correctly, our data race detection can exhibit behaviors that were not legal in C++11 and not exhibited by the POPL17 work, when it comes to release sequences. Do we have a test ensuring that we do indeed see such behaviors?

cbeuw: No we don't, the release sequence changes were made by @JCTyblaidd and predates my weak memory work. There is a test case we could easily implement though https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0982r0.html

RalfJung: Thanks for the pointer! I added that test, and it behaves as expected. :)

PR #4577, opened 2025-09-11
Fix release/scquire synchonization for loads from the store buffer

RalfJung:
We should probably have that. :)  I also clarified the comment on `sc_loaded`.

### DIFF
--- a/src/concurrency/data_race.rs
+++ b/src/concurrency/data_race.rs
@@ -183,6 +183,9 @@ struct AtomicMemoryCellClocks {
     /// contains the vector of timestamps that will
     /// happen-before a thread if an acquire-load is
     /// performed on the data.
+    ///
+    /// With weak memory emulation, this is the clock of the most recent write. It is then only used
+    /// for release sequences, to integrate the most recent clock into the next one for RMWs.
     sync_vector: VClock,
 
     /// The size of accesses to this atomic location.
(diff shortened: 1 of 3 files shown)


oli-obk:
Kind of makes sense that it works but without picking up atomics and locks again and reading it I don't understand it fully.

A test for the behaviour is good nontheless

RalfJung:
I finally found out where release sequences are implemented, I think. :) @cbeuw would be great if you could take a look.

However, now I have more questions... why is there only a single `sync_vector` per atomic? Shouldn't that be stored inside the store buffer? This way it seems like even if we end up reading from an old value, we'll still acquire the latest vector clock? How is that even correct? The vector clocks of consecutive store buffer elements do not have to be ordered...

RalfJung:
(There is another test I want to add.)

RalfJung:
Uh... this test fails:
```rust
/// Ensure that when we read from an outdated release store, we acquire its clock.
fn old_release_store() {
    check_all_outcomes([None, Some(1)], || {
        let x = static_atomic(0);
        let y = static_atomic(0);

        let t1 = spawn(move || {
            y.store(1, Relaxed);
            x.store(1, Release); // this is what we want to read from
            x.store(3, Relaxed);
        });
        let t2 = spawn(move || {
            if x.load(Acquire) == 1 {
                // We must have acquired the `y.store` so we cannot see the initial value any more.
                Some(y.load(Relaxed))
            } else {
                None
            }
        });

        t1.join().unwrap();
        t2.join().unwrap()
    });
}
```
We in fact do see `Some(0)` here.

I think our logic is just wrong? Or did I miss something?

RalfJung:
Turns out yes, the logic was indeed wrong. The mistake is already present in the paper, but not in the original tsan11 implementation.

PR #4578, opened 2025-09-11
Add compare_exchange support for GenMC mode

Patrick-6:
Followup to https://github.com/rust-lang/miri/pull/4566 and https://github.com/rust-lang/miri/pull/4506.

This PR adds support for using `compare_exchange{_weak}` for GenMC mode, and adds code to emit warnings in cases where bugs may be missed. This includes the missing support for spurious failures for the weak variant, and missing modelling for the failure memory ordering leading to certain executions that cannot be explored.

This PR also adds tests `compare_exchange`.


Patrick-6:
r? @RalfJung 

I have more tests that use `compare_exchange`, but they require supporting provenance, which needs a change on the GenMC side first.

src/concurrency/genmc/mod.rs
```diff
@@ -348,21 +357,82 @@ impl GenmcCtx {
...
+            maybe_upgrade_compare_exchange_success_orderings(success, fail);
+
+        if let Some(warnings_cache) = &self.global_state.warnings_cache {
+            // FIXME(genmc): remove once GenMC supports failure memory ordering in `compare_exchange`.
+            warnings_cache.warn_once_rmw_failure_ordering(
+                ecx,
+                success,
+                upgraded_success_ordering,
+                fail,
+            );
```
RalfJung: Please have the logic that decides whether we will warn here, instead of splitting it across files.

I don't think you need an entire file for this, all you need is something like
```rust
type WarningCache = RwLock<FxHashSet<Span>>;

fn emit_warning(cache: &WarningCache, diagnostic: impl FnOnce() -> NonHaltingDiagnostic) { ... }
```

src/diagnostics.rs
```diff
@@ -637,6 +644,8 @@ impl<'tcx> MiriMachine<'tcx> {
                 ("sharing memory with a native function".to_string(), DiagLevel::Warning),
             ExternTypeReborrow =>
                 ("reborrow of reference to `extern type`".to_string(), DiagLevel::Warning),
+            GenmcCompareExchangeWeak | GenmcCompareExchangeOrderingMismatch { .. } =>
+                ("warning was triggered here".to_string(), DiagLevel::Warning),
```
```suggestion
                ("GenMC might miss possible behaviors of this code".to_string(), DiagLevel::Warning),
```

tests/genmc/fail/data_race/mpu2_rels_rlx.rs
```diff
@@ -0,0 +1,47 @@
+//@compile-flags: -Zmiri-genmc -Zmiri-disable-stacked-borrows
+
+// Translated from GenMC's test `wrong/racy/MPU2+rels+rlx`.
+// Test if Miri with GenMC can detect the data race on `X`.
+// The data race only occurs if thread 1 finishes, then threads 3 and 4 run, then thread 1.
```
RalfJung: One of the `1` here should be a `2` I assume?

tests/genmc/fail/loom/store_buffering.rs
```diff
@@ -0,0 +1,59 @@
+//@ revisions: non_genmc genmc
+//@[non_genmc] compile-flags:
```
RalfJung:
```suggestion
```

tests/genmc/fail/loom/store_buffering.rs
```diff
@@ -0,0 +1,59 @@
...
+
+#[unsafe(no_mangle)]
+fn miri_start(_argc: isize, _argv: *const *const u8) -> isize {
+    // For normal Miri, we need multiple repetitions, but GenMC should find the bug with only 1.
+    const REPS: usize = if cfg!(any(non_genmc, non_genmc_std)) { 128 } else { 1 };
+
+    for _ in 0..REPS {
+        // Reset for each iteration.
+        X.store(0, SeqCst);
+        Y.store(0, SeqCst);
```
RalfJung: Usually we create new atomics for each iteration, to be entirely sure that they don't interfere with each other.

You can just make these local variables instead of `static`, can't you?

tests/genmc/pass/atomics/cas_failure_ord_racy_key_init.rs
```diff
@@ -0,0 +1,62 @@
+//@compile-flags: -Zmiri-genmc -Zmiri-disable-stacked-borrows -Zmiri-ignore-leaks
+// -Zmiri-genmc-print-exec-graphs=all
```
RalfJung: This line here does nothing, what is its point?

tests/genmc/pass/atomics/cas_failure_ord_racy_key_init.rs
```diff
@@ -0,0 +1,62 @@
...
+use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::Ordering::*;
+
+use crate::genmc::*;
+
+const KEY_SENTVAL: usize = usize::MAX;
+
+static KEY: AtomicUsize = AtomicUsize::new(KEY_SENTVAL);
+
+static mut VALUES: [usize; 2] = [0, 0];
```
RalfJung, line 25: What is the point of this array?

This seems to just check that `get_or_init` actually returns a unique value and properly synchronizes so one can then do non-atomic accesses?

tests/genmc/pass/atomics/cas_simple.rs
```diff
@@ -0,0 +1,47 @@
+//@compile-flags: -Zmiri-genmc -Zmiri-disable-stacked-borrows
+
+#![no_main]
+
+use std::sync::atomic::*;
+
+static VALUE: AtomicUsize = AtomicUsize::new(0);
+
+const SUCCESS_ORD: Ordering = Ordering::SeqCst;
+const FAILURE_ORD: Ordering = Ordering::SeqCst;
```
RalfJung: Please inline these, this is unnecessarily verbose.

tests/genmc/fail/loom/store_buffering.rs
```diff
@@ -0,0 +1,59 @@
...
+
+// This is the test `store_buffering` from `loom/test/litmus.rs`, adapted for Miri-GenMC.
+// https://github.com/tokio-rs/loom/blob/dbf32b04bae821c64be44405a0bb72ca08741558/tests/litmus.rs
+
+// This test shows the comparison between running Miri with or without GenMC.
+// Without GenMC, Miri requires multiple iterations of the loop to detect the error.
+
+#![no_main]
+
+#[cfg(not(any(non_genmc_std, genmc_std)))]
```
RalfJung:
```suggestion
```

tests/genmc/fail/loom/store_buffering.rs
```diff
@@ -0,0 +1,59 @@
...
+
+use crate::genmc::*;
+
+static X: AtomicUsize = AtomicUsize::new(0);
+static Y: AtomicUsize = AtomicUsize::new(0);
+
+#[unsafe(no_mangle)]
+fn miri_start(_argc: isize, _argv: *const *const u8) -> isize {
+    // For normal Miri, we need multiple repetitions, but GenMC should find the bug with only 1.
+    const REPS: usize = if cfg!(any(non_genmc, non_genmc_std)) { 128 } else { 1 };
```
RalfJung:
```suggestion
    const REPS: usize = if cfg!(non_genmc) { 128 } else { 1 };
```

tests/genmc/pass/litmus/casdep.rs
RalfJung, line 1: What is this test about? Please at least link to the corresponding GenMC test.

Same for ccr and cii.

Also, none of the actually concurrent tests are checking the values we are seeing, would one of them be suited for that?

tests/genmc/pass/litmus/MPU2_rels_acqf.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    unsafe {
+        spawn_pthread_closure(|| {
+            X.store(1, Relaxed);
+
+            Y.store(0, Release);
+            Y.store(1, Relaxed);
+        });
+        spawn_pthread_closure(|| {
+            let expected = 2;
```
RalfJung: Please inline these trivial let-bindings that are only used once (in all tests).

tests/genmc/pass/atomics/cas_failure_ord_racy_key_init.rs
Patrick-6: Leftovers from debugging ^^

tests/genmc/pass/atomics/cas_failure_ord_racy_key_init.rs
Patrick-6: This was the case that showed I had to upgrade the success ordering to match the failure ordering, otherwise this test would show a data race. I'll write a better comment.

tests/genmc/pass/litmus/casdep.rs
Patrick-6: I'll add result printing to the `MPU2_rels_acqf` test.

tests/genmc/pass/litmus/MPU2_rels_acqf.rs
Patrick-6: Ah these are some leftovers from C/C++, where the `compare_exchange` functions take a pointer/reference, so they can update the expected value ^^

tests/genmc/fail/data_race/mpu2_rels_rlx.rs
Patrick-6: Yes, one at the end.

src/concurrency/genmc/mod.rs
```diff
@@ -84,21 +85,42 @@ impl PerExecutionState {
     }
 }
 
+/// Type for storing spans for already emitted warnings.
+type WarningCache = RwLock<FxHashSet<Span>>;
+
+/// Emit a warning if it hasn't already been reported for current span.
+fn emit_warning<'tcx>(
```
RalfJung: This looks like something for helper.rs.

tests/genmc/pass/atomics/cas_failure_ord_racy_key_init.rs
```diff
@@ -1,5 +1,12 @@
 //@compile-flags: -Zmiri-genmc -Zmiri-disable-stacked-borrows -Zmiri-ignore-leaks
-// -Zmiri-genmc-print-exec-graphs=all
+
+// Adapted from: `impl LazyKey`, `fn lazy_init`: rust/library/std/src/sys/thread_local/key/racy.rs
+// Two threads race to initialize a key, which is just an index into an array in this test.
+// The (Acquire, Release) orderings on the compare_exchange prevent any data races for reading from `VALUES[key]`.
+
+// FIXME(genmc): GenMC does not model the failure ordering of compare_exchange currently.
+//               Miri thus upgrades the success ordering to prevent showing any false data races in cases like this one.
+//               Once GenMC supports the failure ordering, this test should work without the upgrading.
```
RalfJung:
```suggestion
// Miri thus upgrades the success ordering to prevent showing any false data races in cases like this one.
// Once GenMC supports the failure ordering, this test should work without the upgrading.
```

src/concurrency/genmc/mod.rs
```diff
@@ -81,16 +85,43 @@ impl PerExecutionState {
...
+        return;
+    }
+    // This span has not yet been reported, so we insert it into the cache and report it.
+    let mut cache = cache.write().unwrap();
+    if cache.insert(span) {
+        // Some other thread may have added this span while we didn't hold the lock, so we only emit it if the insertions succeeded.
+        ecx.emit_diagnostic(diagnostic());
+    }
+}
+
```
Patrick-6: This approach with just 1 map has a limitation: for `compare_exchange_weak` with a non-matching failure ordering, this will now only report the incorrect failure ordering warning, but not the warning for the missing spurious failures (since they have the same span), is this ok?

tests/genmc/pass/litmus/MPU2_rels_acqf.rs
```diff
@@ -1,46 +1,70 @@
...
+                    X.store(2, Relaxed);
+                }
+            }),
+        ];
+        join_pthreads(ids);
+
+        // Print the values to check that we get all of them:
+        if writeln!(
+            MiriStderr,
+            "X={}, Y={}, a={a:?}, b={b:?}, c:{c}",
```
RalfJung:
```suggestion
            "X={}, Y={}, a={a:?}, b={b:?}, c={c}",
```

tests/genmc/pass/litmus/MPU2_rels_acqf.rs
```diff
@@ -1,46 +1,70 @@
...
+            X.load(Relaxed),
+            Y.load(Relaxed)
+        )
+        .is_err()
```
RalfJung:
```suggestion
        .unwrap_or_else(|| abort())
```
Then you don't need the `if`.

src/concurrency/genmc/mod.rs
RalfJung: I didn't actually mean to go for a single map, you can still have one map for each warning kind.

Patrick-6:
[That CI failure](https://github.com/rust-lang/miri/actions/runs/17668837874/job/50215803835?pr=4578#step:5:87) doesn't look like it is caused by my change?

RalfJung:
Yeah, the Ubuntu repos are having some sort of trouble.

RalfJung:
Please rebase, that should fix the CI.

RalfJung:
Oh and you can also squash the commits, this is ready otherwise. :)

PR #4580, opened 2025-09-11
Fix #4579 by checking if the strong protector is actually "active".

### LINKED ISSUE #4579: [TB] Strong protectors prevent deallocation even when they are zero-sized
Consider the following program:

```rust
fn protectme(_x: &mut (), y: *mut u8) {
    let y = unsafe { Box::from_raw(y) };
    drop(y)
}

fn main() {
    let y = Box::new(42u8);
    let y = Box::leak(y);
    let x = unsafe { &mut *(y as *mut u8 as *mut ()) };
    protectme(x, y);
}
```

The deallocation of the `Box` in `protectme` is currently reported as UB by Tree Borrows (but not by SB) because of the protector on `_x`. However, this protector is zero-sized (no offset have yet been accessed) so it's not actually protecting anything. It should not prevent the deallocation.

JoJoDeveloping:
Where "active" means that the accessed bit is set.
    
Fixes rust-lang/miri#4579. This also reverts rust-lang/miri#3831, or rather turns this into a must-pass test.


tests/pass/both_borrows/zero-sized-protected.rs
RalfJung, line 1: This could become part of the basic_aliasing_model test, no?

JoJoDeveloping: like so?

RalfJung: Yes, thanks :)

PR #4582, opened 2025-09-12
make a basic hello world work on wasip2

RalfJung:
With https://github.com/rust-lang/rust/pull/146207 finally switching to a provenance-compatible API, this can now be made to work.

### DIFF
--- a/src/shims/wasi/foreign_items.rs
+++ b/src/shims/wasi/foreign_items.rs
@@ -35,6 +35,74 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.write_pointer(res, dest)?;
             }
 
+            // Standard input/output
+            // FIXME: These shims are hacks that just get basic stdout/stderr working. We can't
+            // constrain them to "std" since std itself uses the wasi crate for this.
+            "get-stdout" => {
+                let [] =
+                    this.check_shim_sig(shim_sig!(extern "C" fn() -> i32), link_name, abi, args)?;
+                this.write_scalar(Scalar::from_i32(1), dest)?; // POSIX FD number for stdout
+            }
+            "get-stderr" => {
+                let [] =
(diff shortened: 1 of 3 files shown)


src/shims/wasi/foreign_items.rs
```diff
@@ -35,6 +35,74 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let [handle] =
+                    this.check_shim_sig(shim_sig!(extern "C" fn(i32) -> ()), link_name, abi, args)?;
+                let handle = this.read_scalar(handle)?.to_i32()?;
+
+                if !(handle == 1 || handle == 2) {
+                    throw_unsup_format!("wasm output-stream: unsupported handle");
+                }
+                // We don't actually close these FDs, so this is a NOP.
+            }
+            "[method]output-stream.blocking-write-and-flush" => {
```
oli-obk, line 61: I found https://github.com/bytecodealliance/wasi-rs/blob/01af74f3c71dee1afebb75612031732ee3346e4f/crates/wasip2/src/imports.rs#L4516C21-L4517C84 but do you know where these things are actually documented?

RalfJung: I found https://wasmcloud.com/docs/reference/wasi/interfaces/  but that uses higher-level types, so I ended up reversing this based on the file you linked.

bjorn3: The high level interfaces are specified in https://github.com/WebAssembly/WASI/tree/main/wasip2 The exact symbol names are merely a contract between `wit-bindgen` and `wasm-tools component new`. They are not part of the ABI after linking. Based on https://github.com/bytecodealliance/wasm-tools/pull/1828 it seems that wit-bindgen is going to change the used names in the future once wasm-tools versions with support for the new naming scheme are old enough.

PR #4583, opened 2025-09-12
Add GenMC estimation mode.

Patrick-6:
Followup to https://github.com/rust-lang/miri/pull/4578.

This PR adds support for the GenMC estimation mode.
If enabled, before the actual verification, estimation mode which runs the program with random scheduling to estimate the size of the concurrent execution space and required verification time.

Also, add support for switching the scheduling policy used by GenMC.

### DIFF
--- a/src/concurrency/genmc/run.rs
+++ b/src/concurrency/genmc/run.rs
@@ -70,7 +70,7 @@ fn run_genmc_mode_impl<'tcx>(
 
         // Execute the program until completion to get the return value, or return if an error happens:
         let Some(return_code) = eval_entry(genmc_ctx.clone()) else {
-            genmc_ctx.print_genmc_output(genmc_config);
+            genmc_ctx.print_genmc_output(genmc_config, tcx);
             return None;
         };
 
@@ -97,7 +97,7 @@ fn run_genmc_mode_impl<'tcx>(
                 // Since we don't have any span information for the error at this point,
                 // we just print GenMC's error string, and the full GenMC output if requested.
                 eprintln!("(GenMC) Error detected: {error}");
-                genmc_ctx.print_genmc_output(genmc_config);
+                genmc_ctx.print_genmc_output(genmc_config, tcx);
                 return None;
             }
         }
@@ -110,13 +110,13 @@ impl GenmcCtx {
     ///
     /// This message can be very verbose and is likely not useful for the average user.
     /// This function should be called *after* Miri has printed all of its output.
-    fn print_genmc_output(&self, genmc_config: &GenmcConfig) {
+    fn print_genmc_output(&self, genmc_config: &GenmcConfig, tcx: TyCtxt<'_>) {
         if genmc_config.print_genmc_output {
             eprintln!("GenMC error report:");
             eprintln!("{}", self.get_result_message());
         } else {
-            eprintln!(
-                "(Add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report.)"
+            tcx.dcx().note(
+                "add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report"
             );
         }
     }
--- a/tests/genmc/fail/data_race/mpu2_rels_rlx.stderr
+++ b/tests/genmc/fail/data_race/mpu2_rels_rlx.stderr
@@ -16,6 +16,7 @@ note: inside `genmc::spawn_pthread_closure::thread_func::<{closure@tests/genmc/f
 LL |         f();
    |         ^^^
 
-(Add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report.)
+note: add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report
+
 error: aborting due to 1 previous error
 
--- a/tests/genmc/fail/data_race/weak_orderings.rel_rlx.stderr
+++ b/tests/genmc/fail/data_race/weak_orderings.rel_rlx.stderr
@@ -16,6 +16,7 @@ note: inside `genmc::spawn_pthread_closure::thread_func::<{closure@tests/genmc/f
 LL |         f();
    |         ^^^
 
-(Add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report.)
+note: add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report
+
 error: aborting due to 1 previous error
 
--- a/tests/genmc/fail/data_race/weak_orderings.rlx_acq.stderr
+++ b/tests/genmc/fail/data_race/weak_orderings.rlx_acq.stderr
@@ -16,6 +16,7 @@ note: inside `genmc::spawn_pthread_closure::thread_func::<{closure@tests/genmc/f
 LL |         f();
    |         ^^^
 
-(Add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report.)
+note: add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report
+
 error: aborting due to 1 previous error
 
--- a/tests/genmc/fail/data_race/weak_orderings.rlx_rlx.stderr
+++ b/tests/genmc/fail/data_race/weak_orderings.rlx_rlx.stderr
@@ -16,6 +16,7 @@ note: inside `genmc::spawn_pthread_closure::thread_func::<{closure@tests/genmc/f
 LL |         f();
    |         ^^^
 
-(Add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report.)
+note: add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report
+
 error: aborting due to 1 previous error
 
--- a/tests/genmc/fail/loom/buggy_inc.stderr
+++ b/tests/genmc/fail/loom/buggy_inc.stderr
@@ -10,6 +10,7 @@ LL |             std::process::abort();
 
 note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace
 
-(Add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report.)
+note: add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report
+
 error: aborting due to 1 previous error
 
--- a/tests/genmc/fail/loom/store_buffering.genmc.stderr
+++ b/tests/genmc/fail/loom/store_buffering.genmc.stderr
@@ -10,6 +10,7 @@ LL |             std::process::abort();
 
 note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace
 
-(Add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report.)
+note: add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report
+
 error: aborting due to 1 previous error
 
--- a/tests/genmc/fail/simple/2w2w_weak.relaxed4.stderr
+++ b/tests/genmc/fail/simple/2w2w_weak.relaxed4.stderr
@@ -10,6 +10,7 @@ LL |             std::process::abort();
 
 note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace
 
-(Add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report.)
+note: add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report
+
 error: aborting due to 1 previous error
 
--- a/tests/genmc/fail/simple/2w2w_weak.release4.stderr
+++ b/tests/genmc/fail/simple/2w2w_weak.release4.stderr
@@ -10,6 +10,7 @@ LL |             std::process::abort();
 
 note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace
 
-(Add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report.)
+note: add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report
+
 error: aborting due to 1 previous error
 
--- a/tests/genmc/fail/simple/2w2w_weak.sc3_rel1.stderr
+++ b/tests/genmc/fail/simple/2w2w_weak.sc3_rel1.stderr
@@ -10,6 +10,7 @@ LL |             std::process::abort();
 
 note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace
 
-(Add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report.)
+note: add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report
+
 error: aborting due to 1 previous error
 


genmc-sys/src/lib.rs
```diff
@@ -165,6 +189,22 @@ mod ffi {
...
+        /// Standard deviation of the total executions estimate.
+        sd: i64,
+        /// Number of explored executions during the estimation.
+        explored_execs: u64,
+        /// Number of encounteded blocked executions during the estimation.
+        blocked_execs: u64,
+        /// Number of encounteded moot executions during the estimation.
+        /// NOTE: moot executions are only counted in debug mode.
+        moot_execs: u64,
+    }
```
Patrick-6: We cannot share the `struct VerificationResult` where GenMC has the actual result, since that is very large (hard to maintain the CXX mapping), and contains types CXX doesn't support.
This struct bridges the gap: we populate this on the C++ side, then print the data on the Rust side.

`mean` and `sd` are `i64` not `u64` since there is no unsigned variant of C++'s `std::llround` function. This function is used to round an `long double`, which we can't share with the Rust side FWIU.

For the other fields, GenMC uses `unsigned`, so `u64` should be more than enough.

src/concurrency/genmc/config.rs
```diff
@@ -57,6 +62,21 @@ impl GenmcConfig {
                         "Invalid suffix to GenMC argument '-Zmiri-genmc-print-exec-graphs', expected '', '=none', '=explored', '=blocked' or '=all'"
                     )),
             }
+        } else if let Some(trimmed_arg) = trimmed_arg.strip_prefix("schedule-policy=") {
+            genmc_config.params.schedule_policy = trimmed_arg.parse().map_err(|_err| {
+                format!("-Zmiri-genmc-schedule-policy={{wf, wfr, ltr, arbitrary}}")
+            })?;
+        } else if trimmed_arg == "estimate" {
+            // FIXME(genmc): should this be on by default (like for GenMC)?
+            // Enable estimating the execution space and require time before running the actual verification.
```
Patrick-6, line 66: GenMC has estimation mode on by default, which would also be an option for Miri. This will make verification a lot slower though, since it requires at least 10 extra executions for an estimation.
This may be better to wait until Miri-GenMC is more optimized.

src/concurrency/genmc/mod.rs
```diff
@@ -128,13 +129,41 @@ pub struct GenmcCtx {
...
+    /// print an estimation for how many executions the entire verification will require and give a total time estimate.
+    pub fn print_estimation_result(&self, elapsed_time: Duration) {
+        let result = self.handle.borrow().get_estimation_results();
+        let elapsed_time_sec = elapsed_time.as_secs_f64();
+        let estimated_time_sec = elapsed_time_sec * result.mean as f64
+            / (u64::saturating_add(result.explored_execs, result.blocked_execs)) as f64;
+
+        if result.mean < 0 || result.sd < 0 {
+            eprintln!("WARNING: Estimation gave weird results, there may have been an overflow.");
+        }
```
Patrick-6: I'm not sure if this can actually happen, but I guess it doesn't hurt to print a warning if it does.

There is definitely another issue where the estimation mean can be randomly extremely high for some programs. I don't know why that happens, but the issue is on the GenMC side somewhere.
The tests we have currently merged don't seem to show this bug.
Maybe we should mention this in the doc somewhere?

src/concurrency/genmc/mod.rs
```diff
@@ -128,13 +129,41 @@ pub struct GenmcCtx {
...
+        if result.moot_execs != 0 {
+            // Moot executions are only counted by GenMC in debug mode.
+            eprintln!("  Moot     executions: {}", result.moot_execs);
+        }
+        eprintln!("Total executions estimate: {} (+- {})", result.mean, result.sd);
+        eprintln!("Time to completion estimate: {estimated_time_sec:.2}s");
+        eprintln!();
+    }
+
```
Patrick-6: Can we write any tests for using estimation mode?
The actual time and number of executions will change every run (due to a random seed used), so we can't hardcode the expected output. Maybe there is some way to replace the values in the output with placeholders?

src/concurrency/genmc/run.rs
```diff
@@ -1,32 +1,57 @@
...
+    Estimation,
+    Verification,
+}
+
+impl GenmcMode {
+    /// Return whether warnings on unsupported features should be printed in this mode.
+    fn print_unsupported_warnings(self) -> bool {
+        self == GenmcMode::Verification
+    }
+}
```
Patrick-6, line 24: As we discussed at some point, I've disabled warnings in estimation mode to avoid duplication.

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
```diff
@@ -40,3 +40,18 @@ auto MiriGenmcShim::handle_execution_end() -> std::unique_ptr<std::string> {
     GenMCDriver::handleExecutionEnd();
     return {};
 }
+
+/**** Estimation mode result ****/
+
+auto MiriGenmcShim::get_estimation_results() const -> EstimationResult {
+    const auto& res = getResult();
+    uint64_t moot = 0;
+    GENMC_DEBUG(moot = static_cast<uint64_t>(res.exploredMoot););
```
Patrick-6: Moot executions are only counted by GenMC if `ENABLE_GENMC_DEBUG` is enabled, so I've just set it to 0 if it isn't, and only print it if it's non-zero.

genmc-sys/cpp/src/MiriInterface/Setup.cpp
```diff
@@ -119,12 +120,15 @@ static auto to_genmc_verbosity_level(const LogLevel log_level) -> VerbosityLevel
...
     conf->symmetryReduction = params.do_symmetry_reduction;
 
-    // FIXME(genmc): expose this setting to Miri (useful for testing Miri-GenMC).
-    conf->schedulePolicy = SchedulePolicy::WF;
+    // Set the scheduling policy. GenMC uses `WFR` in estimation mode.
+    conf->schedulePolicy = estimation_mode ? SchedulePolicy::WFR : params.schedule_policy;
 
+    // Set the min and max number of executions tested in estimation mode.
+    conf->estimationMin = 10;
+    conf->estimationMax = params.estimation_max;
```
Patrick-6: This actually causes 10+1 executions at a minimum, but it's what GenMC uses too.

doc/genmc.md
```diff
@@ -24,6 +24,9 @@ Note that `cargo miri test` in GenMC mode is currently not supported.
 ### Supported Parameters
 
 - `-Zmiri-genmc`: Enable GenMC mode (not required if any other GenMC options are used).
+- `-Zmiri-genmc-estimate`: This enables estimation of the concurrent execution space and verification time, before running the full verification. This should help users detect when their program is to complex to fully verify in a reasonable time. This will explore enough executions to make a good estimation, but at least 10 and at most `estimation-max` executions.
```
RalfJung:
```suggestion
- `-Zmiri-genmc-estimate`: This enables estimation of the concurrent execution space and verification time, before running the full verification. This should help users detect when their program is too complex to fully verify in a reasonable time. This will explore enough executions to make a good estimation, but at least 10 and at most `estimation-max` executions.
```

doc/genmc.md
```diff
@@ -24,6 +24,9 @@ Note that `cargo miri test` in GenMC mode is currently not supported.
...
+- `-Zmiri-genmc-estimation-max={MAX_ITERATIONS}`: Set the maximum number of executions that will be explored during estimation (default: 1000).
+- `-Zmiri-genmc-schedule-policy={wf, wfr, ltr, arbitrary | random}`: Change the scheduling policy used by GenMC (writes-first, writes-first-randomize, left-to-right, arbitrary/random next thread). The scheduling policy should not affect correctness, but may make verification performance worse than using the default (WF == writes-first).
```
RalfJung: If the only thing it can do is make performance worse... why should one ever use this flag?

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
```diff
@@ -40,3 +40,18 @@ auto MiriGenmcShim::handle_execution_end() -> std::unique_ptr<std::string> {
...
+    return EstimationResult {
+        .mean = static_cast<int64_t>(std::llround(res.estimationMean)),
+        .sd = static_cast<int64_t>(std::llround(std::sqrt(res.estimationVariance))),
```
RalfJung: So what are the original types of these? You said `long double`? Wouldn't it then make more sense to cast them to `double` which corresponds to `f64` in Rust?

genmc-sys/cpp/src/MiriInterface/Setup.cpp
```diff
@@ -119,12 +120,17 @@ static auto to_genmc_verbosity_level(const LogLevel log_level) -> VerbosityLevel
...
+    // Deviation threshold % under which estimation is deemed good enough.
+    conf->sdThreshold = 10;
```
RalfJung: Where do the fix values here (estimationMin, sdThreshold) come from? Did you copy them from GenMC?

src/concurrency/genmc/dummy.rs
```diff
@@ -39,6 +46,10 @@ mod run {
 impl GenmcCtx {
     // We don't provide the `new` function in the dummy module.
 
+    pub fn print_estimation_result(&self) {
+        unreachable!()
+    }
```
RalfJung: Why is this here? The function is only called inside genmc as far as I can see, so it shouldn't have to exist in the dummy module?

src/bin/miri.rs
```diff
@@ -186,17 +187,26 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
...
             };
 
-            // FIXME(genmc): add estimation mode here.
-
-            let return_code = run_genmc_mode(&config, eval_entry_once, tcx).unwrap_or_else(|| {
+            // Estimate the execution space and runtime, if enabled.
+            if genmc_config.do_estimation()
+                && run_genmc_mode(&config, eval_entry_once, tcx, GenmcMode::Estimation).is_none()
+            {
+                // We might already find an error during estimation, then we should stop here.
```
RalfJung: Does this logic have to live here? Seems like it could also live inside the genmc module.

src/concurrency/genmc/mod.rs
```diff
@@ -128,13 +129,42 @@ pub struct GenmcCtx {
...
     }
 
+    /// Given the time taken for the estimation mode run,
+    /// print an estimation for how many executions the entire verification will require and give a total time estimate.
+    pub fn print_estimation_result(&self, elapsed_time: Duration) {
+        let result = self.handle.borrow().get_estimation_results();
+        let elapsed_time_sec = elapsed_time.as_secs_f64();
+        #[allow(clippy::as_conversions)]
+        let estimated_time_sec = elapsed_time_sec * result.mean as f64
+            / (u64::saturating_add(result.explored_execs, result.blocked_execs)) as f64;
```
RalfJung: Does this match what GenMC does?

If you cast the `u64` to `f64` first you can do the addition without worrying about saturation/overflow.

RalfJung (commented):
Thanks! I have to admit I am not entirely convinced by the concept of a separate "estimation mode", but if it's even on-by-default in GenMC then I guess it makes sense for us to have it set up, and hopefully turn it on when performance improves.

Please add a test that actually uses these new codepaths.

genmc-sys/cpp/src/MiriInterface/Setup.cpp
Patrick-6: When we create a new `Config`, we get a default initialized struct (C++ default for all fields, i.e., 0 for integers, denoted by the `{}` syntax [here](https://github.com/Patrick-6/genmc/blob/af9cc9ccd5d412b16defc35dbf36571c63a19c76/src/Verification/Config.hpp#L53-L55)).
The actual defaults for GenMC come from the argument parsing code [here](https://github.com/Patrick-6/genmc/blob/af9cc9ccd5d412b16defc35dbf36571c63a19c76/src/main.cpp#L203-L210) (based on the LLVM CLI arg parser), so we currently have to copy all those that we need.

doc/genmc.md
Patrick-6: About the usefulness of the estimation mode. It was introduced in [1]. The reasoning was quote:

> Time Unpredictability (§3): Non-expert users of GenMC were finding it
difficult to estimate how long verification will take, and whether it is worth
waiting for the verification result or give up. To address this problem, we
implement a procedure that produces a probabilistic estimate of the size of
the state space so that users can anticipate the total verification cost, and
perhaps revise their code as necessary.

I've noticed this problem with Loom, my program is checked quickly with some small number of threads, but then never seems to finish when I add even one more. Due to the super-exponential runtime, it can be very difficult to know if it will take 5 minutes or 5 days to finish.

Also, we made the estimation mode part of the core goals of my thesis, so I had to include it ^^

[1] Michalis Kokologiannakis, Rupak Majumdar, and Viktor Vafeiadis. Enhancing genmc’s usability and performance. (2024)

doc/genmc.md
Patrick-6: It is not very useful for a normal user, but GenMC also exposes it too (maybe we could make a "developer options" section in the doc?).
This option is very useful for testing, since there should be no differences in outcomes (except order), when verifying a correct program (for programs with multiple possible errors, the schedule affects which one is encountered first and thus shown).
Ideally, we would run all `genmc/pass/` tests once with the default schedule, and once with `arbitrary` (except for tests where we print the outputs).

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
Patrick-6: Yeah that's probably better. For very large results, it wouldn't fit in an `i64` or `f64`, so we have rounding/truncation either way, but the `f64` could at least be `infinite` to show that it happened.

src/concurrency/genmc/mod.rs
Patrick-6: GenMC uses `unsigned` (32 bit on most platforms) wrapping addition. I've used u64 so there's never an issue even if that is ever changed by GenMC, and anyway it should never be higher than `estimation_max + 1` (`1001` by default), so I guess the saturating add is overkill.

genmc-sys/cpp/src/MiriInterface/Setup.cpp
```diff
@@ -119,12 +120,17 @@ static auto to_genmc_verbosity_level(const LogLevel log_level) -> VerbosityLevel
     );
     conf->symmetryReduction = params.do_symmetry_reduction;
 
-    // FIXME(genmc): expose this setting to Miri (useful for testing Miri-GenMC).
-    conf->schedulePolicy = SchedulePolicy::WF;
+    // Set the scheduling policy. GenMC uses `WFR` in estimation mode.
+    conf->schedulePolicy = estimation_mode ? SchedulePolicy::WFR : params.schedule_policy;
 
+    // Set the min and max number of executions tested in estimation mode.
+    conf->estimationMin = 10;
```
RalfJung:
```suggestion
    conf->estimationMin = 10; // default taken from GenMC
```

genmc-sys/cpp/src/MiriInterface/Setup.cpp
RalfJung:
```suggestion
    conf->sdThreshold = 10; // default taken from GenMC
```

genmc-sys/src/lib.rs
```diff
@@ -54,10 +55,12 @@ impl GenmcScalar {
 impl Default for GenmcParams {
     fn default() -> Self {
         Self {
+            estimation_max: 1000,
```
```suggestion
            estimation_max: 1000, // default taken from GenMC
```

doc/genmc.md
RalfJung: GenMC exposes things that are useful to test GenMC; that doesn't mean they are useful to expose in Miri as well. We don't have to test the GenMC algorithm here, but just the glue code that wires it into Miri. Is the scheduler policy useful for that?

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
RalfJung: f64::MAX is around 10^300. Seems unlikely the estimate will be that big. ;)

src/concurrency/genmc/mod.rs
RalfJung: Then just use `strict_add`.

src/concurrency/genmc/dummy.rs
Patrick-6: That was from an earlier version where more GenMC code lived in `bin/miri.rs`, I'll remove this and the other remnants of that from `dummy.rs`

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
Patrick-6: It's possible, if you have 167 threads all `fetch_add` an atomic integer (without symmetry reduction), you get `167! ~= 1.5*10^300` execs. ^^
(GenMC already returns a negative estimate starting at 21 threads in this case due to the cast to `long long` truncating `21!`)

doc/genmc.md
Patrick-6: We had to rework the GenMC scheduler specifically for Miri to respect the po-rf-relation in the program, so testing this wouldn't hurt. Having such tests should also prevent future changes to the scheduling algorithm in GenMC from breaking Miri (GenMC does no tests to ensure that the po-rf-order is always upheld, so changes may accidentally break this promise)

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
RalfJung: Note that in C++, casting a float to an int outside the range of the int is UB. Truncation is not guaranteed. (`llround` is not UB in that case, just "implementation-defined".)

RalfJung: Anyway, I would suggest the `EstimationResult` type should use the same types as C++ and just forward the data, so that all the logic lives in Rust.

doc/genmc.md
RalfJung: That sounds like Miri should specifically test the po-rf property. But how does testing the random scheduler help?

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
```diff
@@ -43,13 +45,23 @@ auto MiriGenmcShim::handle_execution_end() -> std::unique_ptr<std::string> {
 
 /**** Estimation mode result ****/
 
+/// Cast a `long double` to a `double`, converting any values not fitting a `double` into
+/// `infinite`.
+static auto cast_f128_f64(long double value) -> double {
```
RalfJung: Why does this need a helper function...?

doc/genmc.md
Patrick-6: In some tests, deterministic scheduling without po-rf may coincidentally match the po-rf schedule, or it doesn't match, but didn't cause a difference visible to the program. Randomized scheduling can help detect such cases.

We could probably make a test that has multiple threads reading from each other non-atomically in a way that we can detect wrong values, and just have that run with randomized scheduling, e.g., [Treiber-stack-dynamic](https://github.com/Patrick-6/miri/blob/6a219792abeab1440340d4c08c671dd51b080f6d/tests/genmc/pass/data-structures/treiber_stack_dynamic.rs), with enough threads and additional value checks (that test doesn't work yet due to missing pointer provenance support).

My next PR has one test where I'm trying all schedules, since the comment on the GenMC version mentions that there could be problems if the scheduling was done wrong, for a different reason than po-rf (assume-ctrl: [GenMC version](https://github.com/Patrick-6/genmc/blob/af9cc9ccd5d412b16defc35dbf36571c63a19c76/tests/correct/litmus/assume-ctrl/assume-ctrl.c), [Miri version](https://github.com/Patrick-6/miri/blob/6a219792abeab1440340d4c08c671dd51b080f6d/tests/genmc/pass/litmus/assume-ctrl.rs)).

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
Patrick-6: `long double` to `double` casting is also "implementation-defined" from what I found, so I've written my own cast function that replaces all out-of-range values with `infinite`.

I don't think we can pass [an f128 value](https://doc.rust-lang.org/nightly/std/primitive.f128.html) between the languages, since `f128` isn't even stable in Rust, and CXX doesn't support it.

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
Patrick-6: FWIU, `long double` to `double` cast is "implementation-defined" if the value is outside the range of a `double`, so this prevents any weirdness.
It is a function since I wanted to add the comment of what it is doing, and it is used twice down below.

RalfJung: Where do you get that "implementation-defined" from?
If it was implementation-defined, GCC should document it at https://gcc.gnu.org/onlinedocs/gcc/C-Implementation.html.

RalfJung: Also, `long double` is not always f128. On x86-32 it's f80, a weird type specific to the x87 floating-point unit (that is, for some reason, 96 bits in size).

doc/genmc.md
RalfJung: I was more thinking, Miri could have debug assertions which ensure that the schedule respects po-rf. (Or GenMC could have such assertions internally if that's easier.)

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
Patrick-6: > Where do you get that "implementation-defined" from? If it was implementation-defined, GCC should document it at https://gcc.gnu.org/onlinedocs/gcc/C-Implementation.html.

There are various conflicting sources mentioning it being either "implementation-defined" or UB, but the most official looking source I could find is the [C++ reference on implicit casts](https://en.cppreference.com/w/cpp/language/implicit_cast.html), which says:

> If the source value can be represented exactly in the destination type, it does not change.
    If the source value is between two representable values of the destination type, the result is one of those two values (it is implementation-defined which one, although if IEEE arithmetic is supported, rounding defaults [to nearest](https://en.cppreference.com/w/cpp/numeric/fenv/FE_round.html)).
    Otherwise, the behavior is undefined.

This implies it is actually UB to cast a `long double` to a `double` if it is out of range (?)

doc/genmc.md
Patrick-6: I can't think of a way to even assert that; Miri just reads (non-atomically) from its memory, but doesn't know which event it is supposed to read from. If the replay doesn't respect `rf`, then this will read from the wrong store event, but Miri doesn't even know what the correct store should be.
There can be two cases if `rf` is not upheld: Miri reads from uninit memory, or from the wrong store, depending on whether there was another store before the read or not. The first case is easy to detect, the second requires assertions in the test.

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
RalfJung: The only actually relevant source is the C++ standard. :)

C and C++ leave the semantics of floats entirely open to the implementation. However, GCC and most compilers promise to implement IEEE 754 semantics, and I am pretty sure they define casts with reasonable behavior.

So, just use a cast here please.

RalfJung: At least C also seems to clearly define that this will round:

> A value of any real floating type can be implicitly converted to any other real floating type.

> - if the value can be represented, but cannot be represented exactly, the result is the nearest higher or the nearest lower value (in other words, rounding direction is implementation-defined), although if IEEE arithmetic is supported, rounding is to nearest.

tests/genmc/pass/litmus/viktor-relseq.stderr
```diff
@@ -1,2 +1,8 @@
+(GenMC) Estimation complete. No errors were detected.
```
RalfJung: Seems potentially confusing to say "No errors were detected" here, since there may well be errors reported later.

tests/genmc/pass/litmus/viktor-relseq.rs
```diff
@@ -1,6 +1,15 @@
-//@compile-flags: -Zmiri-genmc -Zmiri-disable-stacked-borrows
+//@ normalize-stderr-test: "Finished estimation in [0-9]+\.[0-9]+ seconds\." -> "Finished estimation in [TIME] seconds."
+//@ normalize-stderr-test: "Explored executions: [0-9]+" -> "Explored executions: [MEAN]"
```
RalfJung: This is not the mean, or is it?

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
RalfJung: What even is a moot execution?^^

Seems like the docs have some explaining to do.

tests/genmc/pass/litmus/viktor-relseq.rs
Patrick-6: Ah right, not the mean, but the number of explored execs

doc/genmc.md
RalfJung: I don't think it's worth having a configurable scheduler just for testing purposes. Randomized testing has its own issues anyway.

I'm entirely fine with adding test cases that (under the default scheduler) hit interesting corner cases regarding po-rf, if there is such a thing.

Patrick-6: > Randomized testing has its own issues anyway.

~~Which issues?~~ (Note that the randomized schedule is still exhaustive, it just does the exploration in a different order)

EDIT: Oh, you meant that the test is randomized. Yeah, GenMC has a test pass that runs all its test with random schedule 10 times, only running a randomized test once may have it pass by coincidence.

RalfJung: They tend to cause spurious failures that are hard to reproduce.

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
Patrick-6: I've removed all mentions of "moot executions", those should only be relevant for developers of GenMC, not for Miri-GenMC.

doc/genmc.md
```diff
@@ -35,15 +34,22 @@ Note that `cargo miri test` in GenMC mode is currently not supported.
...
 
 #### Regular Miri parameters useful for GenMC mode
 
 - `-Zmiri-disable-weak-memory-emulation`: Disable any weak memory effects (effectively upgrading all atomic orderings in the program to `SeqCst`). This option may reduce the number of explored program executions, but any bugs related to weak memory effects will be missed. This option can help determine if an error is caused by weak memory effects (i.e., if it disappears with this option enabled).
 
+#### Developer options
+
+These options are likely not useful for most users, but may be helpful for developers working on Miri-GenMC.
+
+- `-Zmiri-genmc-schedule-policy={wf, wfr, ltr, arbitrary | random}`: Change the scheduling policy used by GenMC (writes-first, writes-first-randomize, left-to-right, arbitrary/random next thread). The scheduling policy should not affect correctness, but may make verification performance worse than using the default (WF == writes-first).
```
RalfJung: As I said, please remove this entirely. I don't think it's worth it.

doc/genmc.md
```diff
@@ -35,15 +34,22 @@ Note that `cargo miri test` in GenMC mode is currently not supported.
   - `error`:    Print errors.
   - `warning`:  Print errors and warnings.
   - `tip`:      Print errors, warnings and tips.
-  - If Miri is built with debug assertions, there are additional log levels available (downgraded to `tip` without debug assertions):
-    - `debug1`:   Print revisits considered by GenMC.
-    - `debug2`:   Print the execution graph after every memory access.
-    - `debug3`:   Print reads-from values considered by GenMC.
+- `-Zmiri-genmc-verbose`: Show more information, such as estimated number of executions, and time taken for verification.
```
RalfJung, line 41: The default output also shows the estimated number of executions, doesn't it?

EDIT: Ah, you implemented something slightly different than the most recent proposal I saw before.

src/concurrency/genmc/mod.rs
```diff
@@ -156,18 +158,15 @@ impl GenmcCtx {
...
+            eprintln!("  Explored executions: {}", result.explored_execs);
+            if result.blocked_execs != 0 {
+                eprintln!("  Blocked  executions: {}", result.blocked_execs);
+            }
+            eprintln!("Expected number of executions: {:.0} ± {:.0}", result.mean, result.sd);
         }
-        eprintln!("Total executions estimate: {} (+- {})", result.mean, result.sd);
-        eprintln!("Time to completion estimate: {estimated_time_sec:.2}s");
-        eprintln!();
+        eprintln!("Expected verification time: {estimated_mean_sec:.2} ± {estimated_sd_sec:.2}s");
```
```suggestion
        eprintln!("Expected verification time: {estimated_mean_sec:.2}s ± {estimated_sd_sec:.2}s");
```

src/concurrency/genmc/run.rs
```diff
@@ -105,20 +97,29 @@ fn run_genmc_mode_impl<'tcx>(
...
+            "Verification complete with {} executions. No errors found.",
+            explored_execution_count + blocked_execution_count
+        );
+        if genmc_config.verbose_output {
+            let elapsed_time = Instant::now().duration_since(time_start);
+            eprintln!();
+            eprintln!("Verification took {elapsed_time:.2?}.");
+            eprintln!("Number of complete executions explored: {explored_execution_count}");
+            if blocked_execution_count > 0 {
+                eprintln!("Number of blocked executions seen: {blocked_execution_count}");
```
RalfJung:
```suggestion
            eprintln!("Verification took {elapsed_time:.2?}.");
            if blocked_execution_count > 0 {
                eprintln!("Number of complete executions explored: {explored_execution_count}");
                eprintln!("Number of blocked executions seen: {blocked_execution_count}");
```
Seems pointless to print explored_execution_count when it equals the total count.

tests/ui.rs
```diff
@@ -277,6 +277,8 @@ regexes! {
     r"\bsys/([a-z_]+)/[a-z]+\b"     => "sys/$1/PLATFORM",
     // erase paths into the crate registry
     r"[^ ]*/\.?cargo/registry/.*/(.*\.rs)"  => "CARGO_REGISTRY/.../$1",
+    // remove time print from GenMC estimation mode output.
+    "Expected verification time: .*±.*" => "Total executions estimate: [MEAN] ± [SD]s",
```
RalfJung:
```suggestion
    "^Expected verification time: .*±.*" => "Total executions estimate: [MEAN] ± [SD]s",
```

doc/genmc.md
Patrick-6: Only the documentation, or also the option parsing itself?

RalfJung: Please remove the entire option. That code doesn't carry its weight IMO.

tests/ui.rs
Patrick-6: The regex with the `^` doesn't match anymore, and none of the other regexes in this file use `^` either, is this really correct?

RalfJung: Hm, odd... I guess we don't do line-by-line matching. Maybe this?
```suggestion
    "\nExpected verification time: .*±.*" => "\nExpected verification time: [MEAN] ± [SD]s",
```

Patrick-6: No, that results in this:
```
Estimating GenMC verification time...Expected verification time: [MEAN] ± [SD]
```

Patrick-6: Yes, that works, but is it that important to match the newline?

RalfJung: Seems just safer to not match in the middle of a line.

doc/genmc.md
Patrick-6: > I'm entirely fine with adding test cases that (under the default scheduler) hit interesting corner cases regarding po-rf, if there is such a thing.

I think the treiber-stack-dynamic test I'll add once we have provenance support is a good fit, so I think we can leave this for a later PR.

src/concurrency/genmc/run.rs
RalfJung: I think ideally we'd preserve the warning cache from the estimation to the main execution... but we can leave a FIXME and file an issue for that if you prefer, it doesn't have to be in this PR.

src/concurrency/genmc/run.rs
RalfJung: You didn't fully resolve this, there's still that `eprintln!();` at the top. You really like your empty lines, don't you? ;)

EDIT: In fact it seems you didn't resolve this at all? No part of my proposal seems to be included in the PR?

Patrick-6: Yeah, I don't like the output to be too crammed ;)

For the change, I seem to have applied it to a different file in [this commit](https://github.com/rust-lang/miri/pull/4583/commits/f25303d96947e4294a3c475a824de60c6971a15f#diff-45c317e5fff8b77b3aec64f4f867c3ccf87df029744187810fcd2b87e1020ea9R164-R169), didn't realize this is for another file, I'll make the change here too.

RalfJung: Ah, the output logic is spread across two files... is there a good way to have it all in the same file?
Given that `mod.rs` is so big, maybe move this function to `run.rs`.

Patrick-6: Should the output be a separate functions or inline in `run_genmc_mode_impl` (in the main GenMC loop)? I think a separate function may be better for readability.

RalfJung: Yeah it can be a separate function. Maybe the output after the verification run should also be a separate function, then the two functions nicely mirror each other (and they should be next to each other ofc).

src/concurrency/genmc/mod.rs
```diff
@@ -169,7 +203,8 @@ impl GenmcCtx {
     /// This function determines if we should continue exploring executions or if we are done.
     ///
     /// In GenMC mode, the input program should be repeatedly executed until this function returns `true` or an error is found.
-    pub fn is_exploration_done(&self) -> bool {
+    fn is_exploration_done(&self) -> bool {
+        // FIXME(genmc): `is_exploration_done` can cause GenMC errors in some cases (see `try_get_error`)
         self.handle.borrow_mut().pin_mut().is_exploration_done()
     }
```
Patrick-6: About the possible error mentioned in the comment, this can happen in GenMC, if [`done`](https://github.com/Patrick-6/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/src/Verification/GenMCDriver.cpp#L383-L394), calling `checkForRaces` through [`isRevisitValid`](https://github.com/Patrick-6/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/src/Verification/GenMCDriver.cpp#L460-L477) finds a race. I'm still checking if this is something to report to the user or if it is a bug in GenMC.

Patrick-6: Simplest fix for now would be to add `self.try_get_error().unwrap();` (similar to `handle_execution_end`)

RalfJung: That panics on error instead of reporting them, though?

src/concurrency/genmc/run.rs
```diff
@@ -79,31 +80,33 @@ fn run_genmc_mode_impl<'tcx>(
...
+            ExecutionEndResult::Error(error) => {
+                // This can be reached for errors that affect the entire execution, not just a specific event.
+                // For instance, linearizability checking and liveness checking report their errors this way.
+                // Neither are supported by Miri-GenMC at the moment though. However, GenMC also
+                // treats races on deallocation as global errors, so this code path is still reachable.
+                // Since we don't have any span information for the error at this point,
+                // we just print GenMC's error message.
+                eprintln!("(GenMC) Error detected: {error}");
+                eprintln!();
+                eprintln!("{}", genmc_ctx.get_result_message());
```
RalfJung: Another empty line I don't agree with. ;)
This makes it look like the message is separate from there being an error.

What is an example of how this looks like?

Patrick-6: We could use the same format as a few lines up:
```rust
eprintln!("(GenMC) Error detected: {error}");
eprintln!("== raw GenMC output =========================");
eprintln!("{}", genmc_ctx.get_result_message());
eprintln!("== end of raw GenMC output ==================");
```
`error` here is [one of these strings](https://github.com/Patrick-6/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/src/Verification/VerificationError.cpp#L32-L83), e.g., "Non-atomic race".
The output message contains the full GenMC output, since we don't have any span information we can't really provide anything better.

src/concurrency/genmc/mod.rs
Patrick-6: I've checked with Michalis, these errors should be reported to the user. The latest commit added proper error handling for these.

src/concurrency/genmc/run.rs
RalfJung: I proposed that there because I thought this output would be quite verbose.
Why do we sometimes tie this printing of `get_result_message` to `genmc_config.print_genmc_output` and sometimes not?

Patrick-6: GenMC produces this result message always, but it is very verbose (contains the entire execution graph, so possibly thousands of lines of output) and contains some messages/warnings not (yet) relevant to Miri.

It can contain useful info, therefore the config option `print_genmc_output` to optionally print it exists.
In some cases (these "end of execution" errors), we have no span info, so the GenMC result message is the only way to get any feedback about where the error happened, that's why we print it there.

src/concurrency/genmc/run.rs
Patrick-6: Note that any warning we encounter during estimation will also be encountered during verification (although possibly in a different order in estimation mode), so we don't miss any warnings with this setup.

Should there be an option to turn off these warnings completely? There may be quite a lot of them for a large program.

Lastly, there is also the issue of the regular Miri warnings, which are currently reported once per execution, since they are tied to the `MiriMachine`. How should we deal with these? (e.g., `int2ptr_warned`)

src/concurrency/genmc/run.rs
RalfJung: Okay... sure, that works. I'd have to see actual output to really comment on this, but we don't have any test triggering this so it's tricky.

We should probably have a shared function for printing that big result message to make it consistent.

> It can contain useful info, therefore the config option print_genmc_output to optionally print it exists.

That makes me wonder why this is not tied to genmc_verbose...

src/concurrency/genmc/run.rs
RalfJung: > Note that any warning we encounter during estimation will also be encountered during verification (although possibly in a different order in estimation mode), so we don't miss any warnings with this setup.

Unless GenMC tells us that estimation already completed the search...

> Should there be an option to turn off these warnings completely? There may be quite a lot of them for a large program.

So far that wasn't necessary for the warnings of that sort that we have so I don't think it is needed for now.

> Lastly, there is also the issue of the regular Miri warnings, which are currently reported once per execution, since they are tied to the MiriMachine. How should we deal with these? (e.g., int2ptr_warned)

Open an issue.
Maybe we should just go back to using global `static` variables for this...

Patrick-6: > Unless GenMC tells us that estimation already completed the search...

That can't happen, estimation and verification are completely separate (we create a new `GenMCDriver` for each).
This means, even if the estimation explores all possible executions, the verification will explore them again (side note: this can be a large overhead for tiny programs (0 to 10 executions), since estimation always does a minimum of 10 executions)

src/concurrency/genmc/run.rs
Patrick-6: genmc_verbose causes < 10 extra lines, but printing the GenMC message prints possibly thousands of lines (one per graph event).
If you want, I can create a tiny test to show what the full output looks like, but that output may not be very stable across Miri/GenMC/rustc/std-lib updates (any change that causes different events to be created changes the output message).

src/concurrency/genmc/run.rs
RalfJung: I thought you added `ExecutionEndResult` to handle that but apparently I misunderstood.

src/concurrency/genmc/run.rs
RalfJung: > genmc_verbose causes < 10 extra lines, but printing the GenMC message prints possibly thousands of lines (one per graph event).

If you ask for verbose, you get verbose. ;)

> If you want, I can create a tiny test to show what the full output looks like, but that output may not be very stable across Miri/GenMC/rustc/std-lib updates (any change that causes different events to be created changes the output message).

What's a command I could run to see the output?

src/concurrency/genmc/run.rs
Patrick-6: No, that is only for within one verification run or one estimation run, it just packages the two previous calls to `handle_execution_end` and `is_exploration_done` together and hides the details by just returning the `ExecutionEndResult`.

src/concurrency/genmc/run.rs
Patrick-6:
```bash
./miri run --features=genmc --dep -Zmiri-genmc -Zmiri-disable-stacked-borrows -Zmiri-genmc-print-genmc-output tests/genmc/fail/simple/2w2w_weak.rs
```

src/concurrency/genmc/run.rs
RalfJung: Ah okay.

Maybe just file an issue regarding the warning situation.

src/concurrency/genmc/run.rs
Patrick-6: For this test case the output contains a warning for unordered writes, but no error, since the error was caught by Miri (`process::abort()`).
If it is an error that GenMC caught (e.g., data races), that error will be printed there too. The error output looks very similar (it also contains the execution graph like the warning does).

Patrick-6: This test shows such an error (place in top level Miri directory and run the same command)
```rust
#![no_main]

#[path = "tests/utils/genmc.rs"]
mod genmc;
use crate::genmc::*;

static mut X: u64 = 0;

#[unsafe(no_mangle)]
fn miri_start(_argc: isize, _argv: *const *const u8) -> isize {
    unsafe {
        spawn_pthread_closure(|| {
            X = 1;
        });
        spawn_pthread_closure(|| {
            X = 2;
        });

        0
    }
}
```

RalfJung: Okay that output is completely useless for the uninitiated, I don't see why we'd enable it by default...

Patrick-6: Would you prefer just a short string like "Non-atomic race" without a span in this case, and possibly a hint to use `Zmiri-genmc-print-genmc-output`? (For the case where we don't have a span, since the error was returned _after_ the execution is complete)

RalfJung: Yeah I think doing that consistently is better.
By default:
```
(Add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report.)
```
And then when that flag is set:
```
GenMC error report:
<error>
```
(and that should be the last thing we print so we don't need an "end" marker)

Patrick-6: Did you leave out the leading `-` in `-Zmiri-genmc-...` on purpose?

RalfJung: Oops.^^

src/concurrency/genmc/run.rs
```diff
@@ -113,6 +117,14 @@ fn run_genmc_mode_impl<'tcx>(
 }
 
 impl GenmcCtx {
+    /// Print the full output message produced by GenMC.
+    /// This message can be very verbose and is likely not useful for the average user.
+    /// This function should be called *after* Miri has printed all of its output.
+    fn print_genmc_output(&self) {
+        eprintln!("GenMC error report:");
+        eprintln!("{}", self.get_result_message());
+    }
```
RalfJung: What I meant was to make the body of this function something like
```rust
if enmc_config.print_genmc_output {
   // ...
} else {
  eprintln!(
            "(Add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report.)"
        );
}
```
We definitely don't want to print that hint *all the time*.

Patrick-6: Ah, wrong default ^^

src/concurrency/genmc/run.rs
```diff
@@ -108,6 +102,10 @@ fn run_genmc_mode_impl<'tcx>(
                 eprintln!("(GenMC) Error detected: {error}");
                 if genmc_config.print_genmc_output {
                     genmc_ctx.print_genmc_output();
+                } else {
+                    eprintln!(
+                        "(Add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report.)"
+                    );
```
RalfJung: I suggested to have the if-then-else in `fn print_genmc_output` so it is shared. Is there some reason not to do that?

Patrick-6: So you would also print the hint in the case further up in this file (for regular errors where we do have span info)?

RalfJung: Yes.

Patrick-6: Sure, we can do that

RalfJung:
LGTM, thanks! Please squash :)

Issue #4587, opened 2025-09-16
read_dir broken on FreeBSD


RalfJung:
Some recent rustc changes broke read_dir on FreeBSD:
```
error: unsupported operation: can't call foreign function `readdir` on OS `freebsd`
    --> /home/runner/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/sys/fs/unix.rs:733:50
     |
  LL |                 let entry_ptr: *const dirent64 = readdir64(self.inner.dirp.0);
     |                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ unsupported operation occurred here
     |
     = help: this means the program tried to do something Miri does not support; it does not indicate a bug in the program
     = note: BACKTRACE:
     = note: inside `<std::sys::fs::unix::ReadDir as std::iter::Iterator>::next` at /home/runner/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/sys/fs/unix.rs:733:50: 733:78
     = note: inside `<std::fs::ReadDir as std::iter::Iterator>::next` at /home/runner/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/fs.rs:2303:9: 2303:22
     = note: inside `<std::iter::Map<std::fs::ReadDir, {closure@tests/pass/shims/fs.rs:307:18: 307:21}> as std::iter::Iterator>::next` at /home/runner/.rustup/toolchains/miri/lib/rustlib/src/rust/library/core/src/iter/adapters/map.rs:107:9: 107:25
     = note: inside `<std::iter::Map<std::fs::ReadDir, {closure@tests/pass/shims/fs.rs:307:18: 307:21}> as std::iter::Iterator>::collect::<std::vec::Vec<(std::ffi::OsString, bool)>>` at /home/runner/.rustup/toolchains/miri/lib/rustlib/src/rust/library/core/src/iter/traits/iterator.rs:2027:9: 2027:38
     = note: inside `<std::collections::BTreeMap<std::ffi::OsString, bool> as std::iter::FromIterator<(std::ffi::OsString, bool)>>::from_iter::<std::iter::Map<std::fs::ReadDir, {closure@tests/pass/shims/fs.rs:307:18: 307:21}>>` at /home/runner/.rustup/toolchains/miri/lib/rustlib/src/rust/library/alloc/src/collections/btree/map.rs:2347:34: 2347:60
[...]

PR #4589, opened 2025-09-16
support `readdir` on FreeBSD

LorrensP-2158466:
Fixes rust-lang/miri#4587.

Modifies the previously named `linux_solarish_readdir64` (now `readdir64`) to also work on `freebsd`.

### DIFF
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -900,14 +900,10 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         }
     }
 
-    fn linux_solarish_readdir64(
-        &mut self,
-        dirent_type: &str,
-        dirp_op: &OpTy<'tcx>,
-    ) -> InterpResult<'tcx, Scalar> {
+    fn readdir64(&mut self, dirent_type: &str, dirp_op: &OpTy<'tcx>) -> InterpResult<'tcx, Scalar> {
         let this = self.eval_context_mut();
 
-        if !matches!(&*this.tcx.sess.target.os, "linux" | "solaris" | "illumos") {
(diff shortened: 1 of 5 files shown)


src/shims/unix/fs.rs
```diff
@@ -900,14 +900,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-        dirent_type: &str,
-        dirp_op: &OpTy<'tcx>,
-    ) -> InterpResult<'tcx, Scalar> {
+    fn readdir64(&mut self, dirent_type: &str, dirp_op: &OpTy<'tcx>) -> InterpResult<'tcx, Scalar> {
         let this = self.eval_context_mut();
 
-        if !matches!(&*this.tcx.sess.target.os, "linux" | "solaris" | "illumos") {
+        if !matches!(
+            &*this.tcx.sess.target.os,
+            "linux" | "solaris" | "illumos" | "macos" | "freebsd"
```
RalfJung: Why "macos"? I don't think std uses this on macos. That means we have no tests, and we shouldn't add this without tests.

RalfJung (commented):
Thanks!

Could you please also rebase and remove the freebsd exception in `tests/pass/shims/fs.rs` so that this gets tested again?

RalfJung:
I have also pushed a commit to further simplify that entry / entry_place business; please make sure to pull that.

src/shims/unix/fs.rs
LorrensP-2158466: The docs for macos and freebsd are the same for this operation, so I thought to also support it for macos. I'll remove it!

RalfJung:
Thanks :)

PR #4591, opened 2025-09-18
Improve GenMC mode scheduling and add assume statements

Patrick-6:
Followup to https://github.com/rust-lang/miri/pull/4583.

Adds improvements to scheduling by determining the next instruction type more accurately.
Also, adds support for assume statements, which allow the user to optimize the verification by skipping certain executions from being explored.


src/concurrency/thread.rs
```diff
@@ -705,12 +708,29 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
...
+                    reason: block_reason @ (BlockReason::Mutex | BlockReason::GenmcAssume),
+                    ..
+                } => {
+                    info!(
+                        "GenMC: schedule returned thread {next_thread_id:?}, which is blocked, so we unblock it now."
+                    );
+                    this.unblock_thread(next_thread_id, *block_reason)?;
+                }
+                _ => {}
+            }
```
Patrick-6, line 720: Should Miri block a thread when it is blocked by an "assume" statement (and later on, from an intercepted `Mutex::lock` call)?
We could just leave the blocking up to GenMC (and leave the threads unblocked in Miri), then we don't need to do the unblocking logic here.
I think blocking the threads in Miri too is better, since it keeps the two tool's view of the thread state in sync.

src/shims/foreign_items.rs
```diff
@@ -485,6 +486,19 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // GenMC mode: Assume statements block the current thread when their condition is false.
+            "miri_genmc_verifier_assume" => {
+                let [condition] =
+                    this.check_shim_sig_lenient(abi, CanonAbi::Rust, link_name, args)?;
+                if this.machine.data_race.as_genmc_ref().is_some() {
+                    this.handle_genmc_verifier_assume(condition)?;
+                } else {
+                    tracing::warn!(
+                        "GenMC: function `miri_genmc_verifier_assume` used, but GenMC mode is not active, skip ..."
+                    );
```
Patrick-6: Not sure how to handle this case, maybe an error (or one-time warning) would be better.

doc/genmc.md
```diff
@@ -50,6 +50,46 @@ Note that `cargo miri test` in GenMC mode is currently not supported.
...
+}
+
+fn spin_until_true(flag: &AtomicBool) {
+  #[cfg(not(feature = "genmc"))]
+  while (!flag.load(Relaxed)) {}
+
+  #[cfg(feature = "genmc")]
+  unsafe { miri_genmc_verifier_assume(flag.load(Relaxed)); }
+}
+```
```
Patrick-6, line 119: I was thinking about how such manual spinloop bounding could look in actual code, but I don't know if a crate feature is the best for this. Maybe `RUSTFLAGS="--cfg genmc"` (like Loom uses) would be better?

tests/utils/miri_extern.rs
```diff
@@ -147,4 +147,7 @@ extern "Rust" {
     /// "symbolic" alignment checks. Will fail if the pointer is not actually aligned or `align` is
     /// not a power of two. Has no effect when alignment checks are concrete (which is the default).
     pub fn miri_promise_symbolic_alignment(ptr: *const (), align: usize);
+
+    /// Blocks the current execution if the argument is false
+    pub fn miri_genmc_verifier_assume(condition: bool);
```
Patrick-6: - Should this function be `unsafe`?
  - Maybe it should, since misusing it can make GenMC miss possibly important executions.
- What should the name be?
  - GenMC uses `__VERIFIER_assume()`
- Should it return the boolean it takes (so it could be chained or used in a `if` or `while`)?

tests/genmc/pass/litmus/assume-ctrl.rs
```diff
@@ -0,0 +1,54 @@
+//@ revisions: order12 order21
+//@compile-flags: -Zmiri-genmc -Zmiri-disable-stacked-borrows -Zmiri-genmc-verbose
+//@normalize-stderr-test: "Verification took .*\n" -> "Verification took [TIME]\n"
```
Patrick-6: I'm using `verbose` here to ensure we actually get the expected number of blocked executions.
GenMC doesn't have this blocked execution since it provides the model checker with annotations for determining when the `assume` condition is true. If Miri ever does that too, this blocked execution should disappear.

src/concurrency/genmc/scheduling.rs
```diff
@@ -1,58 +1,120 @@
...
+        ));
+    };
+    let intrinsice_name = intrinsic_def.name.as_str();
+    let Some(suffix) = intrinsice_name.strip_prefix("atomic_") else {
+        return interp_ok(NonAtomic); // Non-atomic, so guaranteed not an atomic load
+    };
+    // `atomic_store` is guaranteed not a load.
+    // Any future `atomic_*` intrinsics may have load semantics, so we err on the side of caution and classify them as "maybe loads".
+    // FIXME(genmc): make this more precise. (`atomic_fence`? `atomic_singlethreadfence`?)
+    interp_ok(MaybeAtomic(if suffix == "store" { ActionKind::NonLoad } else { ActionKind::Load }))
```
Patrick-6: I've checked the list of atomic symbols [here](https://github.com/rust-lang/rust/blob/0c0c58b8e453f552ebd7f3a1545acdd109de028c/compiler/rustc_span/src/symbol.rs#L521-L540), and I think the only one we can safely say doesn't have any load semantics is `atomic_store`, and possible the two fence symbols.

src/concurrency/thread.rs
RalfJung: I see no good reason to block this in Miri if it requires extra work.

src/concurrency/genmc/scheduling.rs
RalfJung: Regarding the fences, that depends on what GenMC wants. Are fences considered loads or not?

src/concurrency/thread.rs
Patrick-6: I originally added the blocking since we need a way to detect when we have to ask GenMC to schedule again.
By default, something like an assume statement looks to Miri like we can continue on with the current thread afterwards.
The two options I see are: block the thread in that case, or have a boolean flag in the `GenmcCtx` indicating that we have to schedule again (we would set the flag instead of blocking the thread, then reset it on the next schedule call).

src/concurrency/genmc/scheduling.rs
Patrick-6: The GenMC logic for this is [here](https://github.com/MPI-SWS/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/src/Static/LLVMUtils.cpp#L180-L188), which calls [`hasGlobalLoadSemantics`](https://github.com/MPI-SWS/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/src/Runtime/InterpreterEnumAPI.hpp#L117-L125). From this I would say that fences are not considered to have load semantics.

(I think we should treat the intercepted `sync::Mutex::{lock,try_lock}` methods as loads, once we add the interception in a later PR, which should just require adding [these two symbols](https://github.com/rust-lang/rust/blob/0c0c58b8e453f552ebd7f3a1545acdd109de028c/compiler/rustc_span/src/symbol.rs#L2159-L2160) to the list).

src/concurrency/thread.rs
RalfJung: Don't we ask GenMC to schedule before all terminators anyway?

src/concurrency/genmc/scheduling.rs
RalfJung: That function is odd since it doesn't even seem to consider atomic loads.

Please clarify this with Michalis instead of guessing. :)

src/concurrency/thread.rs
Patrick-6: No, only before atomic terminators currently, but we could change that (it would be a bit slower, since we would have to call into GenMC more often).

RalfJung: Ah okay. I think we need a classification for terminators that returns one of 3 categories:
- not interesting for genmc
- interesting but definitely not an atomic load
- interesting and may be an atomic load

That can then be used both to decide whether to invoke the scheduler, and the "is it a load" logic.

RalfJung: Hm, I also don't quite understand how assume fits into this though, since that only blocks once it actually gets evaluated. IOW, there we want to reschedule *after* a terminator, not before, right?

Maybe we should make the genmc wrapper honor the yield_current_thread flag (i.e., it then actually asks genmc for what to schedule next), and genmc_verifier_assume can just yield if the condition does not hold?

Patrick-6: - What is the next instruction after we interpret a custom Miri function (e.g., `miri_genmc_verifier_assume`)? This is one point where scheduling is required (if the condition in the assume is false).
- How can we detect if a function call may have atomic properties, such as thread creation? Should we just assume any function call is possibly atomic, and possibly a load, unless we can determine it to be an `atomic_store`/`fence`?

Patrick-6: Yeah, it should be _after_ the function call, so yield may be the cleanest option

src/concurrency/genmc/scheduling.rs
Patrick-6: I've asked Michalis: fences and non-atomic loads don't count as `ActionKind::Load`.

src/concurrency/thread.rs
Patrick-6: I've looked into this again. We could use yielding for assume statements, but for `Mutex::lock` interception, we need to:
- Call `handle_mutex_lock`
- if the lock is not taken:
  - wait until GenMC schedules this thread again
  - call `handle_mutex_lock` again (This 2. call should always succeed in locking unless there's a bug IIUC)

The blocking callback allows us to do this 2. call, but the yielding makes that harder.

For `Mutex::lock`, could we instead reset the current MIR instruction in the current thread to right before the `lock` call? That way, when the thread is scheduled again, it will call the lock function again.

RalfJung: > * What is the next instruction after we interpret a custom Miri function (e.g., `miri_genmc_verifier_assume`)? This is one point where scheduling is required (if the condition in the assume is false).
> 
>     * How can we detect if a function call may have atomic properties, such as thread creation? Should we just assume any function call is possibly atomic, and possibly a load, unless we can determine it to be an `atomic_store`/`fence`?

I don't follow, those are things you must already handle, no?

> For Mutex::lock, could we instead reset the current MIR instruction in the current thread to right before the lock call? That way, when the thread is scheduled again, it will call the lock function again.

No, I don't like this kind of re-execution, we tried this in the past and it didn't work very well.

Looks like there are actually good reasons to block these threads then. :)

src/concurrency/thread.rs
```diff
@@ -705,12 +708,29 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
...
         if let Some(genmc_ctx) = this.machine.data_race.as_genmc_ref() {
-            let next_thread_id = genmc_ctx.schedule_thread(this)?;
+            let Some(next_thread_id) = genmc_ctx.schedule_thread(this)? else {
+                return interp_ok(SchedulingAction::ExecuteStep);
+            };
+            // GenMC determined that we can schedule this thread again, so we unblock it:
+            // FIXME(genmc): should we even block the threads in Miri?
+            match &this.machine.threads.threads[next_thread_id].state {
+                ThreadState::Blocked {
+                    reason: block_reason @ (BlockReason::Mutex | BlockReason::GenmcAssume),
```
RalfJung: I would say use a single block reason for "blocked on GenMC". Let's not mix this up with normal mutexes.

src/concurrency/thread.rs
```diff
@@ -705,12 +708,29 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
 
         // In GenMC mode, we let GenMC do the scheduling.
         if let Some(genmc_ctx) = this.machine.data_race.as_genmc_ref() {
-            let next_thread_id = genmc_ctx.schedule_thread(this)?;
+            let Some(next_thread_id) = genmc_ctx.schedule_thread(this)? else {
+                return interp_ok(SchedulingAction::ExecuteStep);
+            };
+            // GenMC determined that we can schedule this thread again, so we unblock it:
+            // FIXME(genmc): should we even block the threads in Miri?
```
```suggestion
            // If a thread is blocked on GenMC, we have to implicitly unblock it when it gets scheduled again.
```

src/concurrency/thread.rs
```diff
@@ -110,6 +110,8 @@ pub enum BlockReason {
     Eventfd,
     /// Blocked on unnamed_socket.
     UnnamedSocket,
+    /// Blocked on a GenMC `assume` statement (GenMC mode only).
```
```suggestion
    /// Blocked on a GenMC `assume` statement (GenMC mode only).
    /// Will be implicitly unblocked when GenMC schedules this thread again.
```

src/concurrency/thread.rs
```diff
@@ -572,6 +574,7 @@ impl<'tcx> ThreadManager<'tcx> {
     /// See <https://docs.microsoft.com/en-us/windows/win32/procthread/thread-handles-and-identifiers>:
     /// > The handle is valid until closed, even after the thread it represents has been terminated.
     fn detach_thread(&mut self, id: ThreadId, allow_terminated_joined: bool) -> InterpResult<'tcx> {
+        // FIXME(genmc): does GenMC need to know about detached threads?
```
RalfJung: Well, does it? Does LLI do anything for them? Did you ask Michalis?

src/concurrency/thread.rs
```diff
@@ -705,12 +708,29 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
...
+                    this.unblock_thread(next_thread_id, *block_reason)?;
+                }
+                _ => {}
+            }
 
             let thread_manager = &mut this.machine.threads;
             thread_manager.active_thread = next_thread_id;
 
             assert!(thread_manager.threads[thread_manager.active_thread].state.is_enabled());
+
```
RalfJung: Why the new empty line?

tests/genmc/pass/litmus/assume-ctrl.order12.stderr
```diff
@@ -0,0 +1,5 @@
+Running GenMC Verification...
+Verification complete with 5 executions. No errors found.
+Number of complete executions explored: 4
+Number of blocked executions seen: 1
+Verification took [TIME]\n
```
RalfJung: Something is wrong with your normalization patterns it seems

tests/genmc/pass/litmus/assume-ctrl.rs
```diff
@@ -0,0 +1,54 @@
...
+
+// Translated from GenMC's test "litmus/assume-ctrl".
+// Shortened quote:
+// "An example demonstrating why we should treat assume() statements like
+// if statements, in the sense that the former also create control dependencies."
+//
+// Without this treatment of `assume()`, using a different a different scheduling policies or thread orders may make GenMC miss certain executions.
+// We test this by switching the order in which we spawn the threads.
```
RalfJung: I don't understand this comment at all. There's no concept of a "control dependency" anywhere in Miri or the GenMC glue, as far as I have seen?

tests/utils/miri_extern.rs
RalfJung: - All extern functions are unsafe.
- `miri_genmc_assume` works IMO. `genmc_verifier` seems redundant.
- Does the GenMC version do that?

src/concurrency/genmc/intercept.rs
```diff
@@ -0,0 +1,45 @@
...
+            None,
+            callback!(
+                @capture<'tcx> {
+                    condition: OpTy<'tcx>,
+                }
+                |this, unblock: UnblockKind| {
+                    assert_eq!(unblock, UnblockKind::Ready);
+
+                    let condition = this.run_for_validation_ref(|this| this.read_scalar(&condition))?.to_bool()?;
+                    assert!(condition);
```
RalfJung: This is quite odd. GenMC has no way of knowing what re-doing this *non-atomic* read will do. In almost all cases this will read from a location that no other thread has access to.

Can this code even be reached? I thought an execution just gets blocked entirely when the condition is `false`?

tests/genmc/pass/litmus/assume-ctrl.rs
RalfJung: > I'm using verbose here to ensure we actually get the expected number of blocked executions.

Then please add a comment explaining that.

src/concurrency/thread.rs
Patrick-6: > > * What is the next instruction after we interpret a custom Miri function (e.g., `miri_genmc_verifier_assume`)? This is one point where scheduling is required (if the condition in the assume is false).
> >   
> >   * How can we detect if a function call may have atomic properties, such as thread creation? Should we just assume any function call is possibly atomic, and possibly a load, unless we can determine it to be an `atomic_store`/`fence`?
> 
> I don't follow, those are things you must already handle, no?

I'm speaking of functions like `pthread_create`, `miri_genmc_verifier_assume`, `std::sync::Mutex::lock`, etc. that are not intrinsics, but regular functions (possibly intercepted by Miri)

tests/utils/miri_extern.rs
Patrick-6: > * Does the GenMC version do that?

No. I thought there might be cases where it is easier to `cfg!()` include the assume in the code if it returns the condition, but it is probably not necessary.

src/concurrency/thread.rs
Patrick-6: Should it be one `BlockReason::Genmc` or more specific with `BlockReason::Genmc(GenmcBlockReason)`?

Currently it is only for assume, but I'll add mutexes later, and other features may want to block threads too.
It might be better to have separate ones, especially for debugging, to track which type a thread is blocked on.

tests/genmc/pass/litmus/assume-ctrl.rs
Patrick-6: Yeah this is GenMC internal. I've had multiple variants of this test using different schedules, since these would fail if GenMC didn't do this treatment of assume internally.
I think I can remove the comment, since I had to remove the scheduling policy variants after we've removed them in the previous PR. (There are now two variants with different thread orders to simulate the scheduling policy differences)

src/concurrency/genmc/intercept.rs
Patrick-6: Currently, it can't, once a thread is blocked by an assume, it will not be executed anymore in the current execution (thus causing a "blocked" execution).

If Miri were to supply the annotations for when to unblock this assume (optimization from "[Unblocking dynamic partial order reduction](https://link.springer.com/chapter/10.1007/978-3-031-37706-8_12)"), then this thread can be unblocked once GenMC determines that the condition now succeeds.
I don't know how that will work exactly, so we can probably replace this code with a comment about needing to be careful if that is ever implemented.

src/concurrency/thread.rs
RalfJung: Yes, what about them? Your code already handles them, right?

Anyway if we keep the blocking state then this discussion is resolved, right?

src/concurrency/thread.rs
RalfJung: Since GenMC doesn't tell you why a thread got unblocked, we can't use it for any actual sanity checks, so I'd rather not track a more specific reason. Fro debugging there's already trace logging saying when a thread gets blocked and why.

tests/genmc/pass/litmus/assume-ctrl.rs
```diff
@@ -0,0 +1,54 @@
...
+
+#[unsafe(no_mangle)]
+fn miri_start(_argc: isize, _argv: *const *const u8) -> isize {
+    // FIXME(genmc,HACK): remove these initializing writes once Miri-GenMC supports mixed atomic-non-atomic accesses.
+    X.store(0, Relaxed);
+    Y.store(0, Relaxed);
+
+    unsafe {
+        let t0 = || {
+            miri_genmc_verifier_assume(2 > Y.load(Relaxed) || Y.load(Relaxed) > 3);
```
RalfJung: Why load Y twice?

tests/genmc/pass/litmus/assume-ctrl.rs
RalfJung: So this was testing for a bug in your changes to GenMC? But not in the glue code that's in this repo?

Maybe just say that this hits a corner case in the logic for assume or so.

We should obviously have some test that actually checks that assume works, I don't understand why this is the best test for that. Shouldn't that be more like the example code you added in the docs?

doc/genmc.md
```diff
@@ -50,6 +50,46 @@ Note that `cargo miri test` in GenMC mode is currently not supported.
 
 <!-- FIXME(genmc): add tips for using Miri-GenMC more efficiently. -->
 
+### Optimizing verification times
+
+<!-- FIXME(genmc): explain how to use miri_start and pthreads to reduce workload -->
```
RalfJung: Given that Michalis is fairly close to merging changes that should drastically improve performance when there are many non-atomic accesses, let's wait to see if that is even still necessary.

doc/genmc.md
```diff
@@ -50,6 +50,46 @@ Note that `cargo miri test` in GenMC mode is currently not supported.
...
+#[cfg(feature = "genmc")]
+extern "Rust" {
+  // This is a special function that Miri provides.
+  // It blocks the thread calling this function if the condition is false.
+  pub fn miri_genmc_verifier_assume(condition: bool);
+}
+
+fn spin_until_true(flag: &AtomicBool) {
+  #[cfg(not(feature = "genmc"))]
+  while (!flag.load(Relaxed)) {}
```
RalfJung:
```suggestion
  while (!flag.load(Relaxed)) { std::hint::spin_loop(). }
```

doc/genmc.md
RalfJung: I'm not sure we should further pollute the cfg namespace. Also, changing RUSTFLAGS depending on Miri runtime flags breaks cargo's caching.

I think eventually we could have a Miri interface crate, and we can have a `miri_genmc_enabled` magic function to decide which one to use here. For now, I think it's too early to worry about this.

src/concurrency/genmc/intercept.rs
RalfJung: Yeah please just make this `unreachable!()`.

Currently the argument to miri_genmc_assume is a boolean *value*, not an expression. So it makes no conceptual sense to even ask about whether the condition will ever succeed in the future. That would need an entirely different API.

tests/genmc/pass/litmus/assume-ctrl.rs
Patrick-6: This is part of the [original test](https://github.com/MPI-SWS/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/tests/correct/litmus/assume-ctrl/assume-ctrl.c), it requires two loads for the assume to cause the issue described by the comment on the test.

RalfJung: I think this test only makes sense when the body of the `assume` is an expression, which it probably is in GenMC. But here it is a value. There isn't actually any control dependency. Your code is equivalent to
```rust
let b = 2 > Y.load(Relaxed) || Y.load(Relaxed) > 3;
miri_genmc_verifier_assume(b);
```

src/concurrency/thread.rs
Patrick-6: GenMC does not support detached threads (`pthread_detach` commented out [here](https://github.com/MPI-SWS/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/include/pthread.h#L180-L184), causes `unknown external function` error if added back and used).
I think we can just rely on Miri's checks for these (e.g., cannot join detached threads), but I'll check with Michalis.

src/concurrency/thread.rs
Patrick-6: This is just about marking them as `ActionKind::Load` (maybe loads) or `ActionKind::Store` (definitely not loads). It might be simpler to treat all function calls as "maybe loads", since keeping a complete and accurate list may be hard to maintain.

genmc-sys/cpp/src/MiriInterface/EventHandling.cpp
```diff
@@ -30,6 +30,12 @@
 #include <memory>
 #include <utility>
 
+/**** Blocking instructions ****/
+
+void MiriGenmcShim::handle_thread_block(ThreadId thread_id) {
+    GenMCDriver::handleAssume(inc_pos(thread_id), AssumeType::User);
```
RalfJung: There's an odd mismatch in the naming here?

src/concurrency/genmc/mod.rs
```diff
@@ -903,8 +894,15 @@ impl GenmcCtx {
...
     /// or added by some automated program transformation, e.g., for spinloops.
-    fn handle_user_block<'tcx>(&self, _machine: &MiriMachine<'tcx>) -> InterpResult<'tcx> {
-        todo!()
+    fn handle_user_block<'tcx>(&self, machine: &MiriMachine<'tcx>) -> InterpResult<'tcx> {
+        let thread_infos = self.exec_state.thread_id_manager.borrow();
+        let curr_thread = machine.threads.active_thread();
+        let genmc_curr_thread = thread_infos.get_genmc_tid(curr_thread);
+        debug!("GenMC: handle_user_block, blocking thread {curr_thread:?} ({genmc_curr_thread:?})");
+
+        self.handle.borrow_mut().pin_mut().handle_thread_block(genmc_curr_thread);
```
RalfJung: And then more odd naming mismatches here ("user block" vs "thread block")

src/concurrency/genmc/scheduling.rs
```diff
@@ -1,58 +1,120 @@
...
+}
+
+fn get_function_kind<'tcx>(
+    ecx: &InterpCx<'tcx, MiriMachine<'tcx>>,
+    func_ty: Ty<'tcx>,
+) -> InterpResult<'tcx, NextInstrKind> {
+    use NextInstrKind::*;
+    let callee_def_id = match func_ty.kind() {
+        ty::FnDef(def_id, _args) => def_id,
+        _ => return interp_ok(MaybeAtomic(ActionKind::Load)), // we don't know the callee, might be an intrinsic or pthread_join
```
RalfJung:
```suggestion
        _ => return interp_ok(MaybeAtomic(ActionKind::Load)), // we don't know the callee, might be pthread_join
```
Function pointers to intrinsics are not a thing.

src/concurrency/genmc/scheduling.rs
```diff
@@ -1,58 +1,120 @@
...
+    let Some(intrinsic_def) = ecx.tcx.intrinsic(callee_def_id) else {
+        // FIXME(genmc): Make this work for other platforms.
+        let item_name = ecx.tcx.item_name(*callee_def_id);
+        return interp_ok(MaybeAtomic(
+            if matches!(item_name.as_str(), "pthread_join" | "WaitForSingleObject") {
+                ActionKind::Load
+            } else {
+                ActionKind::NonLoad
+            },
+        ));
```
RalfJung: You already handled all foreign items above. What's the point of this here? This should just
```suggestion
        // The next step is a call to a regular Rust function.
        return interp_ok(NonAtomic);
```

tests/genmc/pass/litmus/assume-ctrl.rs
Patrick-6: I have a variant for `treiber-stack-dynamic` that uses assumes like in the docs, but we can't merge that yet (needs pointer provenance, and possible temporal atomic/NA mixing).
There is another GenMC's litmus test using assume that Michalis suggested at one point, I can translate that one too, and maybe I can come up with one myself too.

src/concurrency/genmc/scheduling.rs
```diff
@@ -1,58 +1,120 @@
...
+    use NextInstrKind::*;
+    let callee_def_id = match func_ty.kind() {
+        ty::FnDef(def_id, _args) => def_id,
+        _ => return interp_ok(MaybeAtomic(ActionKind::Load)), // we don't know the callee, might be an intrinsic or pthread_join
+    };
+    if ecx.tcx.is_foreign_item(*callee_def_id) {
+        // Some shims, like pthread_join, must be considered loads. So just consider them all loads,
+        // these calls are not *that* common.
+        return interp_ok(MaybeAtomic(ActionKind::Load));
+    }
```
RalfJung: Please move this inside the `else` below. Intrinsics are currently not considered foreign items I think, but they used to be considered foreign items and I'd rather not rely on this not changing again.

src/shims/foreign_items.rs
RalfJung: I think this should `throw_unsup_format!("miri_genmc_verifier_assume is only supported in GenMC mode")` or so.

tests/genmc/pass/litmus/assume-ctrl.rs
RalfJung:
```suggestion
//@normalize-stderr-test: "Verification took .*s" -> "Verification took [TIME]s"
```

src/concurrency/thread.rs
RalfJung: Your logic already treats all *foreign* function calls as "maybe load", which indeed seems like the right choice to me.

tests/genmc/pass/litmus/assume-ctrl.rs
RalfJung: A simple message-passing test with a spin loop for the receiver should be sufficient.

src/concurrency/genmc/scheduling.rs
RalfJung: `Load` should also be renamed to `MaybeLoad`.

tests/genmc/pass/litmus/assume-ctrl.rs
Patrick-6: FWIU GenMC takes a boolean, not an expression ([__VERIFIER_assume](https://github.com/MPI-SWS/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/include/genmc.h#L15), [__VERIFIER_assume_internal](https://github.com/MPI-SWS/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/include/genmc_internal.h#L136-L137)).

But yeah, this test may not make sense to keep for Miri, I'll add another one (maybe remove this one?).

doc/genmc.md
Patrick-6: From what I heard, the "event coalescing" that may drastically increase NA event handling is only in the conceptual stage currently, there is nothing yet implemented (unless you've heard something different in the meantime).

Even then, there are differences, at least between std threads and pthreads:
- Spawning and not joining N empty pthreads results in 1 execution
- Spawning and not joining N empty std::threads result in 6^(N-1) executions (joining them results in only 1). I haven't yet investigated why this happens, but I suspect it might be some kind of race on some cleanup code (?). Maybe we should document that too/instead?

Also, if someone tries to verify a program with many execs, not having to run the `fn main` code millions of times seems like it might be worth switching to `miri_start` for, even if it is made more efficient in the future.

tests/genmc/pass/litmus/assume-ctrl.rs
RalfJung: Hm, maybe the transformations treat the call special somehow? I don't see how this can work as a boolean...

Yeah, please remove this test and use one that more obviously is just about `assume`.

src/concurrency/genmc/scheduling.rs
```diff
@@ -1,58 +1,120 @@
...
     InterpCx, InterpResult, MiriMachine, TerminationInfo, ThreadId, interp_ok, throw_machine_stop,
 };
 
+enum NextInstrKind {
+    MaybeAtomic(ActionKind),
+    NonAtomic,
+}
+
+/// Check if a call or tail-call could have atomic load semantics.
+fn get_next_instruction_kind<'tcx>(
```
RalfJung, line 16: This improved logic doesn't seem to affect any tests. Shouldn't this reduce the number of executions because we now schedule better?

genmc-sys/cpp/src/MiriInterface/EventHandling.cpp
Patrick-6: Ah, that function was renamed in GenMC, but I didn't update the Miri side.

doc/genmc.md
RalfJung: I really don't like having to tell people to rewrite their code (then they might as well use loom). Let's see what the performance is like after the NA change (IIUC this work has been going on in parallel with yours).

> Spawning and not joining N empty std::threads result in 6^(N-1) executions

That's odd, and seems worth an issue.

src/concurrency/genmc/scheduling.rs
Patrick-6: That is part of GenMC, I was unable to have it changed so far. Should there be a `ActionKind` equivalent enum on the Miri side with proper naming? That may also be easier to see the definition without having to go to GenMC's code (i.e., keep the enums coming from GenMC inside the `genmc-sys` crate).

src/concurrency/genmc/scheduling.rs
Patrick-6: No, the number of explored executions is identical, this change just makes the handling of the executions possibly more efficient (spending less time in GenMC's code).

src/concurrency/genmc/scheduling.rs
RalfJung: Ah, right, I forgot... 
Let;s just adjust the docs then, in the enum declaration, to make it clear that `Load` means "maybe load" and `NonLoad` means "definitely not a load".

doc/genmc.md
Patrick-6: There are currently cases where the users *must* modify their code to be able to use GenMC mode, such as the [`spinloop_assume`](https://github.com/rust-lang/miri/blob/9b933aff7b09c0bd63cc103c96de9e74c6e2c302/tests/genmc/pass/intercept/spinloop_assume.rs) test I've just added. A spinloop without a bound will cause infinitely many possible executions, so GenMC never terminates.

The simplest option to allow such programs without user code modification would be to add something similar to GenMC's `--unroll=N` parameter, which limits _ALL_ loops to `N` iterations (at the LLVM IR level), thus making the program bounded.
For Miri-GenMC, I think it would be better to only bound loops that contain atomic operations, otherwise it will block programs with any loops, such as string equality with more than `N` characters (this could still include loops that should not be bounded, but would be better than bounding all loops).
This would also need to be a dynamic transformation, detecting loops and counting how many iterations, including handling of nested loops, so I don't think this will trivial to implement (GenMC does it statically on the LLVM IR).

src/concurrency/genmc/scheduling.rs
Patrick-6: I've updated the comment [here](https://github.com/Patrick-6/miri/blob/9b933aff7b09c0bd63cc103c96de9e74c6e2c302/genmc-sys/src/lib.rs#L261-L264)

doc/genmc.md
```diff
@@ -70,19 +70,27 @@ Since there are no side effects, replacing the loop with one iteration is safe.
 This can be done conditionally, for example by adding a feature to your `Cargo.toml` specifically for running with Miri-GenMC:
 
 ```rust
-#[cfg(feature = "genmc")]
+#[cfg(miri)]
 extern "Rust" {
```
RalfJung:
```suggestion
unsafe extern "Rust" {
```

doc/genmc.md
```diff
@@ -70,19 +70,27 @@ Since there are no side effects, replacing the loop with one iteration is safe.
...
   // This is a special function that Miri provides.
   // It blocks the thread calling this function if the condition is false.
-  pub fn miri_genmc_verifier_assume(condition: bool);
+  pub fn miri_genmc_assume(condition: bool);
```
RalfJung:
```suggestion
  pub unsafe fn miri_genmc_assume(condition: bool);
```
You said it is marked unsafe, so let's actually do that.

genmc-sys/cpp/src/MiriInterface/EventHandling.cpp
RalfJung: Why did you resolve this? The names still don't match.

tests/genmc/pass/intercept/spinloop_assume.rs
```diff
@@ -0,0 +1,98 @@
+//@ revisions: bounded123 bounded321 replaced123 replaced321
+//@compile-flags: -Zmiri-genmc -Zmiri-disable-stacked-borrows -Zmiri-genmc-verbose
+//@normalize-stderr-test: "Verification took .*s" -> "Verification took [TIME]s"
+
+// This test uses assume GenMC statements to bound or replace spinloops.
```
RalfJung:
```suggestion
// This test uses GenMC assume statements to bound or replace spinloops.
```

tests/genmc/pass/intercept/spinloop_assume.rs
```diff
@@ -0,0 +1,98 @@
...
+// Three threads pass a value to each other, spinning on an atomic FLAG to wait for the previous thread.
+//
+// There are two variants, one limits the spinloop to three iterations, and one that completely replaces the spin loop.
+// Without this loop bounding, this test *cannot* be verified, since GenMC will have to explore infinitely many executions (one per possible number of loop iterations).
+
+// FIXME(genmc): GenMC provides the `--unroll=N` option, which limits all loops to at most N iterations (at the LLVM IR level).
+// Such an option for Miri would allow a variant of this test without manual bounding, using this automatic loop bounding instead.
+
+// We use different thread orders to ensure if doesn't just pass by chance (each thread order should give the same result).
```
RalfJung:
```suggestion
// We use different thread orders to ensure it doesn't just pass by chance (each thread order should give the same result).
```

tests/genmc/pass/litmus/assume-ctrl.rs
RalfJung, line 1: Didn't we say we'd remove this test entirely?

genmc-sys/src/lib.rs
```diff
@@ -258,7 +258,9 @@ mod ffi {
     /// Corresponds to GenMC's type with the same name.
     /// Should only be modified if changed by GenMC.
     enum ActionKind {
-        /// Any Mir terminator that's atomic and has load semantics.
+        /// Any MIR terminator that's atomic and that may have load semantics.
+        /// This includes functions with atomic properties, such as `pthread_create`.
+        /// If the exact type of the terminator cannot be determined, load is a safe default `Load`.
         Load,
         /// Anything that's not a `Load`.
```
```suggestion
        /// Anything that's definitely not a `Load`.
```

doc/genmc.md
RalfJung: > There are currently cases where the users must modify their code to be able to use GenMC mode, such as the [spinloop_assume](https://github.com/rust-lang/miri/blob/9b933aff7b09c0bd63cc103c96de9e74c6e2c302/tests/genmc/pass/intercept/spinloop_assume.rs) test I've just added. A spinloop without a bound will cause infinitely many possible executions, so GenMC never terminates.

Fair, but that's separate from "write your Rust code so that it looks like C code".

tests/genmc/pass/intercept/spinloop_assume.rs
RalfJung, line 1: I like this test :)

genmc-sys/cpp/src/MiriInterface/EventHandling.cpp
Patrick-6: The naming is a bit weird on the GenMC side, this `handleAssume` is only called to block the thread, meaning only when the assume condition is false (the [doc comment](https://github.com/Patrick-6/genmc/blob/75218c458fefa2dafe96200e528dc28f7da00030/src/Verification/GenMCDriver.hpp#L160-L161) also mentions this, and [the implementation](https://github.com/Patrick-6/genmc/blob/75218c458fefa2dafe96200e528dc28f7da00030/src/Verification/GenMCDriver.cpp#L1638) calls `handleBlock` internally).

Maybe `handle_user_assume_block` or something is a better name for the Miri side?

tests/genmc/pass/litmus/assume-ctrl.rs
Patrick-6: Yes, it's removed now

genmc-sys/cpp/src/MiriInterface/EventHandling.cpp
RalfJung: Let's go for `handle_assume_block`. Not sure where you are getting the "user" part from?

RalfJung:
Still has a few unresolved discussions.

@rustbot author

EDIT: Ah, it's also still marked as a draft. :)

genmc-sys/cpp/src/MiriInterface/EventHandling.cpp
Patrick-6: `user` comes from the [`AssumeType::User`](https://github.com/Patrick-6/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/src/ExecutionGraph/LoadAnnotation.hpp#L24-L25), used [here](https://github.com/Patrick-6/miri/blob/5ccb51314a3df23a727a25b85430f54d25447512/genmc-sys/cpp/src/MiriInterface/EventHandling.cpp#L36).

Patrick-6: This will become more relevant once different types of "assumes" are added, then it could either be separate functions for each type, or to share the `AssumeType` with the Miri side.
What should we go with for now?

src/concurrency/genmc/scheduling.rs
Patrick-6: I think we can resolve this? I've added the two fence types to the check.

genmc-sys/cpp/src/MiriInterface/EventHandling.cpp
RalfJung: I'd say let's call it `handle_assume_block` and have a comment that for now `AssumeType::User` is hard-coded, and it can become a parameter in the future.

src/concurrency/genmc/intercept.rs
```diff
@@ -32,11 +32,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 }
```
RalfJung: Now you can remove the condition field above from the callback.

src/concurrency/thread.rs
RalfJung: This may have to do with the weirdness around `thread::spawn` that you mentioned elsewhere? When the join handle of a `thread::spawn` gets dropped, `pthread_detach` is called. Did the test where you tried "Spawning and not joining N empty pthreads" also call pthread_detach?

Patrick-6: Yeah, that seems to be the cause, `std::mem::forget`ting the `std::thread` handles leads to only 1 execution again.
The opposite is not true though, `pthread_detach` does _not_ cause these additional executions.

I think this is caused by GenMC exploring all possible ways to drop the [`Arc<Packet>`](https://github.com/rust-lang/rust/blob/dd7fda570040e8a736f7d8bc28ddd1b444aabc82/library/std/src/thread/mod.rs#L1798-L1801) in the `JoinHandle` ([here](https://github.com/rust-lang/rust/blob/dd7fda570040e8a736f7d8bc28ddd1b444aabc82/library/std/src/thread/mod.rs#L569) and [here](https://github.com/rust-lang/rust/blob/dd7fda570040e8a736f7d8bc28ddd1b444aabc82/library/std/src/thread/mod.rs#L512-L517), `my_packet` is dropped with the `JoinHandle`, `their_packet` when the spawned thread finishes).
There are a few RMW and fence operations involved in this, which likely come from [`std::sync::Arc::drop`](https://github.com/rust-lang/rust/blob/dd7fda570040e8a736f7d8bc28ddd1b444aabc82/library/alloc/src/sync.rs#L2637-L2687).

RalfJung: Ah right, there's a (benign) race here about who drops their `Packet` first.

doc/genmc.md
```diff
@@ -50,6 +50,52 @@ Note that `cargo miri test` in GenMC mode is currently not supported.
...
+  while (!flag.load(Relaxed)) {
+    std::hint::spin_loop();
+  }
+}
+
+/// Bounding this loop means the only execution that is explored where there is only 1 load.
+/// Incorrect use of such assume statements can lead GenMC to miss important executions, so it is marked `unsafe`.
+///
+/// Note that such bounding is similar to what can be achieved in Loom using `skip_branch` (`https://docs.rs/loom/latest/loom/fn.skip_branch.html`)
+fn spin_until_true_bounded(flag: &AtomicBool) {
```
RalfJung:
```suggestion
fn spin_until_true_genmc(flag: &AtomicBool) {
```

doc/genmc.md
```diff
@@ -83,18 +83,29 @@ fn spin_until_true(flag: &AtomicBool) {
...
 /// Incorrect use of such assume statements can lead GenMC to miss important executions, so it is marked `unsafe`.
 ///
 /// Note that such bounding is similar to what can be achieved in Loom using `skip_branch` (`https://docs.rs/loom/latest/loom/fn.skip_branch.html`)
-fn spin_until_true_bounded(flag: &AtomicBool) {
+fn spin_until_true_genmc(flag: &AtomicBool) {
   unsafe { miri_genmc_assume(flag.load(Relaxed)); }
 }
+
+/// This function shows how to limit the number of iterations.
+fn spin_until_true_genmc_(flag: &AtomicBool) {
```
RalfJung: Why the trailing `_`?

Patrick-6: To avoid using the same name as the function above.
Should I call them both the same? Or one `spin_until_true_genmc_replaced` and one `spin_until_true_genmc_bounded`?

RalfJung: Oh I didn't realize you had the `0..3` function here... why should that be here at all? This is never what one would want to use, is it?

doc/genmc.md
```diff
@@ -65,7 +65,7 @@ If the loop has side effects, bounding it means that certain executions are not
 
 The following code gives an example for how to replace a loop that waits for a boolean to be true.
 Since there are no side effects, replacing the loop with one iteration is safe.
-This can be done conditionally, for example by adding a feature to your `Cargo.toml` specifically for running with Miri-GenMC:
+The last function also shows how to limit the number of executions if there are side-effects that should be explored.
```
RalfJung: Oh, this is why.

If we explain how to manually do loop unrolling it should be separate explanations for "simple spinloop with no side-effect" and "loop with side-effect".

doc/genmc.md
```diff
@@ -1,15 +1,24 @@
...
-**NOTE: Currently, no actual GenMC functionality is part of Miri, this is still WIP.**
+Miri in GenMC mode takes a program as input like regular Miri, but instead of running it once, the program is executed repeatedly, until all possible executions allowed by the Rust memory model are explored.
+This includes all possible thread interleavings and all allowed return values for atomic operations, including cases that are very rare to encounter on actual hardware.
 
-<!-- FIXME(genmc): add explanation. -->
+GenMC requires the input program to be bounded, i.e., have finitely many possible executions, otherwise it will not terminate.
+Any loops that may run infinitely must be replaced or bounded.
+
+GenMC makes use of Dynamic Partial Order Reduction (DPOR) to reduce the number of executions that must be explored, but the runtime can still be super-exponential in the size of the input program (number of threads and amount of interaction between threads).
+Large programs may not be verifiable in a reasonable amount of time.
```
Patrick-6, line 18: I've added an actual explanation for GenMC mode and a link to the correctness issues, since there is some available functionality now.

doc/genmc.md
```diff
@@ -1,15 +1,24 @@
...
 
+**NOTE: GenMC mode is not yet fully implemented, and has [several correctness issues](https://github.com/rust-lang/miri/issues/4572). Using GenMC mode currently requires manually compiling Miri, see [Usage](#usage).**
+
+
 [GenMC](https://github.com/MPI-SWS/genmc) is a stateless model checker for exploring concurrent executions of a program.
 Miri-GenMC integrates that model checker into Miri.
 
-**NOTE: Currently, no actual GenMC functionality is part of Miri, this is still WIP.**
+Miri in GenMC mode takes a program as input like regular Miri, but instead of running it once, the program is executed repeatedly, until all possible executions allowed by the Rust memory model are explored.
+This includes all possible thread interleavings and all allowed return values for atomic operations, including cases that are very rare to encounter on actual hardware.
```
RalfJung, line 10:
```suggestion
This includes all possible thread interleavings and all allowed return values for atomic operations, including cases that are very rare to encounter on actual hardware.
(However, this does not include other sources of non-determinism, such as the absolute addresses of allocations.
It is hence still possible to have latent bugs in a test case even if they passed GenMC.)
```

doc/genmc.md
```diff
@@ -1,15 +1,24 @@
...
+
+GenMC requires the input program to be bounded, i.e., have finitely many possible executions, otherwise it will not terminate.
+Any loops that may run infinitely must be replaced or bounded.
```
RalfJung:
```suggestion
Any loops that may run infinitely must be replaced or bounded (see below).
```

doc/genmc.md
```diff
@@ -52,20 +61,18 @@ Note that `cargo miri test` in GenMC mode is currently not supported.
 
 ### Optimizing verification times
```
RalfJung: Based on the text above, this isn't just about "optimizing", right?
```suggestion
### Eliminating unbounded loops
```

doc/genmc.md
```diff
@@ -76,33 +83,43 @@ unsafe extern "Rust" {
 }
 
 /// This functions loads an atomic boolean in a loop until it is true.
-/// GenMC will explore all executions where this does 1, 2, 3, ... loads, which can drastically increase the number of explored executions.
+/// GenMC will explore all executions where this does 1, 2, ..., ∞ loads, which means the verification will never terminate.
+/// GenMC requires all input programs to be bounded (have finitely many possible executions), so we cannot verify a program containing this function.
```
RalfJung:
```suggestion
/// GenMC will explore all executions where this does 1, 2, ..., ∞ loads, which means the verification will never terminate.
```

doc/genmc.md
```diff
@@ -76,33 +83,43 @@ unsafe extern "Rust" {
...
 fn spin_until_true(flag: &AtomicBool) {
   while (!flag.load(Relaxed)) {
     std::hint::spin_loop();
   }
 }
 
-/// Fully replacing this loop means the only execution that is explored is where there is only 1 load.
-/// Incorrect use of such assume statements can lead GenMC to miss important executions, so it is marked `unsafe`.
+/// By this loop with an assume statement, the only executions that will be explored are those with exactly 1 load.
```
RalfJung:
```suggestion
/// By replacing this loop with an assume statement, the only executions that will be explored are those with exactly 1 load.
```

doc/genmc.md
```diff
@@ -76,33 +83,43 @@ unsafe extern "Rust" {
...
+/// Incorrect use of assume statements can lead GenMC to miss important executions, so it is marked `unsafe`.
 ///
 /// Note that such bounding is similar to what can be achieved in Loom using `skip_branch` (`https://docs.rs/loom/latest/loom/fn.skip_branch.html`)
 fn spin_until_true_genmc(flag: &AtomicBool) {
```
RalfJung:
```suggestion
fn spin_until_true_genmc(flag: &AtomicBool) {
```
Seems a bit arbitrary to just randomly mention Loom here.

doc/genmc.md
Patrick-6: Yeah, that's a better title for the section.

src/concurrency/genmc/scheduling.rs
```diff
@@ -1,58 +1,115 @@
...
+    let basic_block = &frame.body().basic_blocks[loc.block];
+    if let Some(_statement) = basic_block.statements.get(loc.statement_index) {
+        // Statements can't be atomic.
+        return interp_ok(NonAtomic);
+    }
+    match &basic_block.terminator().kind {
+        // All atomics are modeled as function calls to intrinsic functions.
+        // The one exception is thread joining, but those are also calls.
+        TerminatorKind::Call { func, .. } | TerminatorKind::TailCall { func, .. } => {
+            let frame = thread_manager.active_thread_stack().last().unwrap();
```
RalfJung: You already have a `frame` in scope, why do you get a 2nd one?

Patrick-6: I'll remove it.

RalfJung:
Looks good, please squash. :)

Patrick-6:
> Looks good, please squash. :)

Done :)

PR #4595, opened 2025-09-19
TB: update terminology to match paper & MiniRust

RalfJung:


src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -618,14 +618,14 @@ impl Tree {
...
-    /// map are relative to `base_offset`.
+    /// `initial_perms` defines the initial permissions for a block of memory starting at
+    /// `base_offset`. These may nor may not be already marked as "accessed".
     /// `default_perm` defines the initial permission for the rest of the allocation.
+    /// These are definitely not "accessed".
     ///
     /// For all non-accessed locations in the RangeMap (those that haven't had an
     /// implicit read), their SIFA must be weaker than or as weak as the SIFA of
-    /// `default_perm`.
+    /// `default_perm`; otherwise, this will panic.
```
RalfJung: @JoJoDeveloping  I am actually a bit confused by this. The assertion looks as follows:
```rust
                assert!(
                    default_strongest_idempotent
                        >= perm.permission.strongest_idempotent_foreign_access(protected)
                );
```
This seems to run for all locations in `initial_perms`, not just the non-accessed ones. Why that?

JoJoDeveloping: It could be changed to the following, I think:
```
                assert!(
                    perm.is_accessed()
                        || default_strongest_idempotent
                            >= perm.permission.strongest_idempotent_foreign_access(protected)
                );
```
But the way it is currently makes it easier to reason about because the SIFA correction can be proven correct without considering the initial read.

RalfJung: Then we should update the comments to match, though.

RalfJung: Instead of tracking this assertion all the way through, wouldn't it be easier to just compute the maximum SIFA and pass that to `update_last_accessed_after_retag`? I've done that now in the 2nd commit, could you take a look if that makes sense?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -618,67 +618,63 @@ impl Tree {
...
         // Create the node
         self.nodes.insert(
             idx,
             Node {
                 tag: new_tag,
                 parent: Some(parent_idx),
                 children: SmallVec::default(),
-                default_initial_perm: default_perm,
+                default_initial_perm: outside_perm,
                 default_initial_idempotent_foreign_access: default_strongest_idempotent,
```
RalfJung, line 636: What I am not entirely sure about is whether this is still correct, or whether this also needs to be set to the maximum.

RalfJung: Ah, this is only used for "outside" states so no max is needed.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -618,67 +618,63 @@ impl Tree {
...
 
+        // We need to know the biggest SIFA for `update_last_accessed_after_retag` below.
+        let mut max_sifa = default_strongest_idempotent;
         for (Range { start, end }, &perm) in
-            initial_perms.iter(Size::from_bytes(0), initial_perms.size())
+            inside_perms.iter(Size::from_bytes(0), inside_perms.size())
         {
             assert!(perm.is_initial());
+            max_sifa =
+                cmp::max(max_sifa, perm.permission.strongest_idempotent_foreign_access(protected));
```
RalfJung: `perm.permission.strongest_idempotent_foreign_access` is odd, isn't it? `perm` already contains a SIFA. We should just read that.

JoJoDeveloping:
Looks reasonable now.

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: I think we could. If that field is wrongly initialized it is a bug anyways.

RalfJung: Yeah my latest revision does that.

RalfJung:
Thanks for taking a look!

Issue #4593, opened 2025-09-19
Check bootstrap build in CI


RalfJung:
The rustc repo builds Miri with the beta bootstrap compiler and standard library. Multiple times now this has led to problems because that build configuration is not tested in the Miri repo. We should find some way to test this.

That's not entirely trivial though since we have to somehow figure out the correct bootstrap compiler to use.

RalfJung:
This is actually quite tricky. We need to build against the nightly private rustc crates but the beta standard library -- I don't think that artifact is distributed anywhere. So we'd probably have to actually build the Rust compiler ourselves somehow, within Miri CI?

PR #4596, opened 2025-09-19
CI: test bootstrap build

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/4593


RalfJung:
Oh, this doesn't work...

RalfJung:
I guess we can actually just do a full check-build of rustc + Miri in a few minutes. Nice.

The failure is expected since right now Miri indeed does not build against bootstrap std.

RalfJung:
I've integrated this into https://github.com/rust-lang/miri/pull/4598.

PR #4600, opened 2025-09-22
fix SIFA logic

RalfJung:
At least I think the old logic was wrong? Cc @JoJoDeveloping 

We should definitely have a test for this.


RalfJung:
I think this is the actual assertion we needed before I added the `max` earlier today:
```rust
// Accessed nodes have the SIFA fixed up as part of the access. For everything else,
// let's ensure default_strongest_idempotent has the minimum SIFA so the fixup
// below does the right thing.
if !perm.accessed {
    assert!(default_strongest_idempotent <= perm.idempotent_foreign_access);
}
```
This assertion also passes. The thing we checked instead (`default_strongest_idempotent >= perm.idempotent_foreign_access`) seems just entirely pointless, but it also passed since for non-accessed nodes we in fact always have `default_strongest_idempotent == perm.idempotent_foreign_access` (because if a non-accessed node exists, then `outside_perm` is the same as `nonfreeze_perm`).

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -640,13 +640,13 @@ impl<'tcx> Tree {
...
         self.nodes.get_mut(parent_idx).unwrap().children.push(idx);
 
-        // We need to know the biggest SIFA for `update_last_accessed_after_retag` below.
-        let mut max_sifa = default_strongest_idempotent;
+        // We need to know the weakest SIFA for `update_last_accessed_after_retag` below.
+        let mut min_sifa = default_strongest_idempotent;
         for (Range { start, end }, &perm) in
             inside_perms.iter(Size::from_bytes(0), inside_perms.size())
         {
             assert!(perm.permission.is_initial());
```
JoJoDeveloping, line 648: can we add an assert here that `perm.idempotent_foreign_access == perm.permission.strongest_idempotent_foreign_access(protected)`? I do not fully recall what sets `perm.idempotent_foreign_access` initially and since it's "far away" from this function it might be easy to miss that this is used here.

JoJoDeveloping:
Good catch!

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: The constructor sets it and it's a private field so nobody outside this files can mess with it, but sure.

PR #4603, opened 2025-09-24
Implement Pointer conversions to and from GenMC.

Patrick-6:
Add support for `AtomicPtr` in GenMC mode by adding provenance info to values sent to GenMC.
This requires a change to GenMC, so we now depend on a [new commit](https://gitlab.inf.ethz.ch/public-plf/genmc/-/tree/cd01c12032bdd71df742b41c7817f99acc72e7ab/) (still not upstream GenMC).


genmc-sys/build.rs
```diff
@@ -227,12 +227,14 @@ fn compile_cpp_dependencies(genmc_path: &Path, always_configure: bool) {
...
-        "./cpp/src/MiriInterface/Exploration.cpp",
-        "./cpp/src/MiriInterface/Setup.cpp",
-        "./cpp/src/MiriInterface/ThreadManagement.cpp",
-    ];
+        "MiriInterface/EventHandling.cpp",
+        "MiriInterface/Exploration.cpp",
+        "MiriInterface/Setup.cpp",
+        "MiriInterface/ThreadManagement.cpp",
+    ]
+    .map(|file| std::path::absolute(cpp_files_base_path.join(file)).unwrap());
```
Patrick-6, line 239: Having absolute paths here improves the developer experience, since you can now CTRL-click on paths in the error messages to open the corresponding file (e.g., in VSCode), which doesn't work well with `./` paths.

genmc-sys/src/lib.rs
```diff
@@ -45,10 +45,17 @@ pub fn create_genmc_driver_handle(
 }
 
 impl GenmcScalar {
-    pub const UNINIT: Self = Self { value: 0, is_init: false };
+    pub const UNINIT: Self = Self { value: 0, extra: 0, is_init: false };
+
+    pub const MUTEX_LOCKED_STATE: Self = Self::from_u64(1);
+    pub const MUTEX_UNLOCKED_STATE: Self = Self::from_u64(0);
```
Patrick-6: Rebasing error, I'll remove these in my next push.

src/concurrency/genmc/helper.rs
```diff
@@ -101,11 +109,39 @@ pub fn scalar_to_genmc_scalar<'tcx>(
...
+        // TODO GENMC: is using `size: Size` ok here? Can we ever have `size != sizeof pointer`?
+
+        // FIXME: Currently GenMC mode incompatible with aliasing model checking
+        let tag = BorTag::default();
+        let provenance = crate::machine::Provenance::Concrete { alloc_id, tag };
+        let offset = addr;
+        let ptr = rustc_middle::mir::interpret::Pointer::new(provenance, offset);
+
+        let size = size.bytes().try_into().unwrap();
+        return interp_ok(Scalar::Ptr(ptr, size));
```
Patrick-6: Should there be an assumption that the `size` we get from [here](https://github.com/Patrick-6/miri/blob/f3170664d91f3f5ceb4a611e567d9cf5623ec4ed/src/concurrency/data_race.rs#L737) is pointer sized? (or, currently: exactly 64 bits)

Patrick-6 (commented):
r? @RalfJung Just this and 2 more PRs remaining (temporal mixing and mutex interception) ;)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4603/dc87a815ccb3e4e0fe1e4742b1f9e04fc7c1e6df..f3170664d91f3f5ceb4a611e567d9cf5623ec4ed)*

src/provenance_gc.rs
```diff
@@ -220,6 +220,11 @@ impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
     fn run_provenance_gc(&mut self) {
         let this = self.eval_context_mut();
+        // FIXME(genmc): In GenMC mode, we skip running the garbage collector.
+        // Borrow tracking is not yet supported in GenMC mode, and allocations are never deleted.
+        if this.machine.data_race.as_genmc_ref().is_none() {
+            return;
+        }
```
Patrick-6, line 227: Instead of disabling garbage collection entirely, would it be possible to keep a list of all `AllocId`s that we've sent to GenMC, and retain those, instead of retaining all allocations?

src/provenance_gc.rs
```diff
@@ -220,6 +220,11 @@ impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
     fn run_provenance_gc(&mut self) {
         let this = self.eval_context_mut();
+        // FIXME(genmc): In GenMC mode, we skip running the garbage collector.
+        // Borrow tracking is not yet supported in GenMC mode, and allocations are never deleted.
+        if this.machine.data_race.as_genmc_ref().is_some() {
+            return;
+        }
```
Patrick-6: Instead of disabling garbage collection entirely, would it be possible to keep a list of all AllocIds that we've sent to GenMC, and retain those, instead of retaining all allocations?

RalfJung: Yeah that's definitely possible. A bit annoying that we'd have to store them a 2nd time though...

You can just have a hash set of AllocId in the genmc per-exec state and then visit that during GC collection.

genmc-sys/build.rs
RalfJung: Okay, please add a comment explaining that.

src/concurrency/genmc/helper.rs
RalfJung: Do *not* use the `Scalar::Ptr`/`Scalar::Int` constructors directly. They are semi-private. Only use the functions to construct these values.

Here, you want to construct a `Pointer` first, and then convert that with `Scalar::from_pointer`.

src/concurrency/genmc/helper.rs
```diff
@@ -101,11 +109,39 @@ pub fn scalar_to_genmc_scalar<'tcx>(
...
 /// For pointers, attempt to convert the stored base address of their allocation back into an `AllocId`.
 pub fn genmc_scalar_to_scalar<'tcx>(
-    _ecx: &MiriInterpCx<'tcx>,
+    ecx: &MiriInterpCx<'tcx>,
     scalar: GenmcScalar,
     size: Size,
 ) -> InterpResult<'tcx, Scalar> {
-    // FIXME(genmc): Add GenmcScalar to Miri Pointer conversion.
+    if scalar.extra != 0 {
+        // We have a pointer!
```
RalfJung: This code is unnecessarily bloated, it can be half as long if you don't let-bind every single thing.

src/concurrency/genmc/helper.rs
```diff
@@ -101,11 +109,39 @@ pub fn scalar_to_genmc_scalar<'tcx>(
...
+        let alloc_size = 0; // TODO GENMC: what is the correct size here? Is 0 ok?
+        let only_exposed_allocations = false;
+        let Some(alloc_id) =
+            ecx.alloc_id_from_addr(base_addr, alloc_size, only_exposed_allocations)
+        else {
+            // TODO GENMC: what is the correct error in this case? Maybe give the pointer wildcard provenance instead?
+            throw_unsup_format!(
+                "Cannot get allocation id of pointer received from GenMC (base address: 0x{base_addr:x}, pointer address: 0x{:x})",
+                addr.bytes()
+            );
```
RalfJung: Isn't this just `unreachable!()`?

genmc-sys/src/lib.rs
```diff
@@ -166,6 +173,7 @@ mod ffi {
     #[derive(Debug, Clone, Copy)]
     struct GenmcScalar {
         value: u64,
+        extra: u64,
```
RalfJung, line 176: Please add a doc comment explaining how we use this field.

genmc-sys/cpp/src/MiriInterface/Setup.cpp
RalfJung, line 1: These changes are just formatting changes right?

Patrick-6: Yeah, the `.clang-format` I've made has a column width limit, which applies to comments too. Sometimes, running formatting (I think mainly if there is a syntax error), it indents the comment more, and then wraps the text. Fixing the error and running formatting again doesn't fix the comment then.

I've never noticed this issue with rustfmt, I suspect since it doesn't wrap comments that are too long.
Should I disable the column limit for comments?

RalfJung: This file was already formatted before this PR though, wasn't it?

src/provenance_gc.rs
RalfJung: Taking a step back here, we're piling up a bunch of hacks that all have the same cause:
- not removing AllocIds from the int2ptr map
- not running the GC at all (making the above superfluous)
- even having `alloc_id_from_addr` work on non-exposed allocations

We can avoid all of that if we store the mapping between AllocId and addresses for genmc-relevant pointers inside GenmcCtx. Then the normal Miri logic doesn't have to worry about catering for all the ways in which genmc is different, and we even know which AllocId to preserve (since they will be in that map).

It's a bit of a shame to duplicate that mapping, but given how many special cases we are piling up, it feels worth it.

tests/genmc/pass/atomics/atomic_ptr_multi_load.rs
```diff
@@ -0,0 +1,71 @@
+//@ revisions: ptr08 ptr16 ptr32 ptr64
+//@compile-flags: -Zmiri-genmc -Zmiri-disable-stacked-borrows
+
+// Test that we can pass pointers to different allocation types to GenMC, including pointers not pointing to the start of the allocation.
+// There should be one explored execution where the second thread read null, and one execution per possible pointer to read.
```
RalfJung:
```suggestion
// There should be one explored execution where the second thread reads null, and one execution per possible pointer to read.
```

tests/genmc/pass/atomics/atomic_ptr_multi_load.rs
```diff
@@ -0,0 +1,71 @@
...
+use genmc::*;
+
+#[cfg(ptr08)]
+type Int = u8;
+#[cfg(ptr16)]
+type Int = u16;
+#[cfg(ptr32)]
+type Int = u32;
+#[cfg(not(any(ptr08, ptr16, ptr32)))]
+type Int = u64;
```
RalfJung: Why would the pointee type make any difference?

tests/genmc/pass/atomics/atomic_ptr_ops.rs
```diff
@@ -0,0 +1,87 @@
...
+        array[2] = 0xC;
+        if *ptr != 0xC {
+            std::process::abort();
+        }
+        // The other array elements should be unchanged.
+        for (i, x) in array.iter().enumerate() {
+            if i != 2 && *x != 0xAAAA {
+                std::process::abort();
+            }
+        }
```
RalfJung: This is testing basic pointer properties, not operations, right? Seems to duplicate the tests in `ptr_multi_load`.

OTOH there are a bunch of operations on the AtomicPtr type that are not tested here. Can you add them?

tests/genmc/pass/atomics/atomic_ptr_multi_load.rs
```diff
@@ -0,0 +1,71 @@
+//@ revisions: ptr08 ptr16 ptr32 ptr64
+//@compile-flags: -Zmiri-genmc -Zmiri-disable-stacked-borrows
+
+// Test that we can pass pointers to different allocation types to GenMC, including pointers not pointing to the start of the allocation.
```
RalfJung:
```suggestion
// Test that we can pass pointers to different allocation types to GenMC and back, including pointers not pointing to the start of the allocation.
```

tests/genmc/pass/atomics/atomic_ptr_multi_load.rs
RalfJung, line 1: Maybe call this `atomic_ptr_roundtrip` or so? It's testing that the pointer passes to genmc and back correctly, IIUC.

tests/genmc/pass/data-structures/ms_queue_dynamic.rs
```diff
@@ -0,0 +1,206 @@
+//@ revisions: default_RW2 default_R1W1 default_R1W2 spinloop_assume_RW2 spinloop_assume_R1W2
```
RalfJung: Please add a comment explaining the naming scheme here. Also, explain what is the point of trying various different numbers of readers and writers (as opposed to just a single fixed set of threads, like 2 of each kind or so).

tests/genmc/pass/atomics/atomic_ptr_ops.rs
Patrick-6: Do you want to test nightly functionality of `AtomicPtr`? I think I'm already using basically all (atomic) non-nightly functionality (except [`fetch_update`](https://doc.rust-lang.org/src/core/sync/atomic.rs.html#2019-2026) / `update`, which are just a `compare_exchange_weak` loop internally). `fetch_ptr_add`, [`fetch_and`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicPtr.html#method.fetch_and) etc. are all nightly.

tests/genmc/pass/atomics/atomic_ptr_multi_load.rs
Patrick-6: Different alignment constraints per type (to ensure it works with less than 8 byte alignment), but I guess just testing u8 should be enough then.

tests/genmc/pass/atomics/atomic_ptr_ops.rs
RalfJung: fetch_and etc have recently been stabilized and are riding the train towards a stable release, see https://doc.rust-lang.org/nightly/std/sync/atomic/struct.AtomicPtr.html#method.fetch_and.

tests/genmc/pass/atomics/atomic_ptr_multi_load.rs
RalfJung: Why is the alignment relevant?

genmc-sys/cpp/src/MiriInterface/Setup.cpp
Patrick-6: Yeah, but I've missed some cases where this happened back then (the "but GenMC" being on the new line before).

(btw, the `skipUninitLoadChecks` change at the bottom of the file is _not_ a formatting change, that is due to a change on the GenMC side, where the API changed slightly (different parameter))

genmc-sys/build.rs
Patrick-6: Should we detect all the C++ files in `genmc-sys/cpp` automatically here? That would avoid having to manually add new files to the list in the future.

src/concurrency/genmc/helper.rs
Patrick-6: I tried to find an example for when it might happen, but found a different bug instead. ^^

I now also think that this case is unreachable, since reading a pointer in a thread means there is a reads-from relation to the thread that made the allocation. The GenMC replay ensures that the write event is replayed before the load that reads-from it, and the allocation must happen before the write. Therefore, Miri always knows about the allocation at the point where GenMC returns a pointer to it, and this lookup here should never fail.

tests/genmc/pass/atomics/atomic_ptr_multi_load.rs
Patrick-6: I don't see a reason to think that there could be a case where pointers with arbitrary alignment are not supported (accidental truncation maybe?), but it may be good to test for that anyway. The other `AtomicPtr` tests are all just using u64, so such an issue would not be detected if we never use a type with lower alignment requirements.

genmc-sys/build.rs
RalfJung: No, I prefer being explicit about which files are included in the build.

tests/genmc/pass/atomics/atomic_ptr_multi_load.rs
RalfJung: Okay, that doesn't sound like we need 4 revisions of this test then.

tests/genmc/pass/data-structures/ms_queue_dynamic.rs
Patrick-6: I'm not sure how many tests we actually need here, but we can probably get rid of the reader-writer tests, and just use separate reader and writer threads, e.g., 1 reader and 1, 2, 3 writers, and the same with assume statements to compare.
That should show if scaling improves with updates to GenMC mode in the future.

tests/genmc/pass/atomics/atomic_ptr_ops.rs
Patrick-6: I've added these functions now too

genmc-sys/build.rs
Patrick-6: > Okay, please add a comment explaining that.

Done

genmc-sys/src/lib.rs
Patrick-6: Done

src/concurrency/genmc/scheduling.rs
```diff
@@ -107,6 +107,11 @@ impl GenmcCtx {
         // Depending on the exec_state, we either schedule the given thread, or we are finished with this execution.
         match result.exec_state {
             ExecutionState::Ok => interp_ok(Some(thread_infos.get_miri_tid(result.next_thread))),
+            ExecutionState::Error => {
+                // We have an error, but we don't know what it is.
+                // We stop the execution, since we ask GenMC for errors at the end, which will catch that error.
+                throw_machine_stop!(TerminationInfo::Exit { code: 0, leak_check: false });
+            }
```
Patrick-6: I've realized that there can be the case where an error occurs, but we never reached a program exit in the current execution, in which case the `expect` below would panic. This `Error` case prevents that problem.

tests/genmc/fail/data_race/atomic_ptr_alloc_write_race.rs
```diff
@@ -0,0 +1,42 @@
...
+    unsafe {
+        let ids = [
+            spawn_pthread_closure(|| {
+                let a: *mut u64 = alloc(Layout::new::<u64>()) as *mut u64;
+                X.store(a, Relaxed); // Relaxed ordering does not synchronize the alloc with the other thread.
+            }),
+            spawn_pthread_closure(|| {
+                let b = X.load(Relaxed);
+                if !b.is_null() {
+                    *b = 42; //~ ERROR: Undefined Behavior
```
Patrick-6: I would like to have a test case where this is replaced with a `dealloc` (to cause a malloc-free race), but that is one of the cases where we currently don't get a span (GenMC doesn't return the error immediately).
This is the output:
```
Running GenMC Verification...
(GenMC) Error detected: Attempt to access freed memory
note: add `-Zmiri-genmc-print-genmc-output` to MIRIFLAGS to see the detailed GenMC error report
```
How can I do this with the UI test? It expects me to give an `//~ ERROR:` annotation, but there isn't one I could add.

Otherwise: I could add my fix for this issue on the GenMC side, and hope the fix will also be included in the official GenMC release once that is completed (or deal with the problem of the missing span then).

src/concurrency/genmc/helper.rs
```diff
@@ -80,19 +82,27 @@ pub fn split_access(address: Size, size: Size) -> impl Iterator<Item = (u64, u64
...
     scalar: Scalar,
 ) -> InterpResult<'tcx, GenmcScalar> {
     interp_ok(match scalar {
         rustc_const_eval::interpret::Scalar::Int(scalar_int) => {
             // FIXME(genmc): Add u128 support once GenMC supports it.
             let value: u64 = scalar_int.to_uint(scalar_int.size()).try_into().unwrap();
-            GenmcScalar { value, is_init: true }
+            GenmcScalar { value, extra: 0, is_init: true }
+        }
+        rustc_const_eval::interpret::Scalar::Ptr(pointer, size) => {
```
Patrick-6, line 95: Should this use `scalar.to_int` and `scalar.to_pointer()`?
That would lose the compile time exhaustiveness check (if there are ever new enum variants added).

src/concurrency/genmc/helper.rs
```diff
@@ -101,16 +111,29 @@ pub fn scalar_to_genmc_scalar<'tcx>(
...
-    // NOTE: GenMC always returns 64 bit values, and the upper bits are not yet truncated.
-    // FIXME(genmc): GenMC should be doing the truncation, not Miri.
-    let (value_scalar_int, _got_truncated) = ScalarInt::truncate_from_uint(scalar.value, size);
-    interp_ok(Scalar::Int(value_scalar_int))
+    // If `extra` is zero, we have a regular integer.
+    if scalar.extra == 0 {
+        // NOTE: GenMC always returns 64 bit values, and the upper bits are not yet truncated.
+        // FIXME(genmc): GenMC should be doing the truncation, not Miri.
+        let (value_scalar_int, _got_truncated) = ScalarInt::truncate_from_uint(scalar.value, size);
+        return interp_ok(Scalar::Int(value_scalar_int));
```
Patrick-6: Should this be changed to not use `Scalar::Int`?
This way is convenient due to the `truncate_from_uint` function. `Scalar::from_int` panics if the value is too large, so that is not a good alternative.

src/provenance_gc.rs
Patrick-6: Yeah I agree, I think that is a much cleaner solution.

Patrick-6: I've implemented a map and removed the `alloc_id_from_addr` parameter for non-exposed allocs, but there might be more hacks we can remove.

tests/genmc/pass/data-structures/ms_queue_dynamic.rs
RalfJung: Whatever it is, it needs comments explaining it. It's far from clear that one is supposed to look at the execution numbers for these variants and compare them somehow.

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
```diff
@@ -26,6 +26,8 @@ auto MiriGenmcShim::schedule_next(
         return SchedulingResult { ExecutionState::Ok, static_cast<int32_t>(result.value()) };
     if (getExec().getGraph().isBlocked())
         return SchedulingResult { ExecutionState::Blocked, 0 };
+    if (getResult().status.has_value())
+        return SchedulingResult { ExecutionState::Error, 0 };
```
RalfJung, line 30: Why does `getResult().status.has_value()` indicate an error?

genmc-sys/src/lib.rs
RalfJung: You made it a comment on the type. Can't this be a comment on the field?

genmc-sys/src/lib.rs
```diff
@@ -167,6 +167,11 @@ mod ffi {
 
     /// This type corresponds to `Option<SVal>` (or `std::optional<SVal>`), where `SVal` is the type that GenMC uses for storing values.
     /// CXX doesn't support `std::optional` currently, so we need to use an extra `bool` to define whether this value is initialized or not.
+    ///
+    /// If `extra` is zero, `value` is a plain integer.
+    /// If `extra` is non-zero, it contains information about the provenance of the pointer stored in `value`.
+    /// `extra` contain the base address of the allocation that this pointer belongs to, since `AllocId` is not consistent across executions.
```
```suggestion
    /// `extra` contains the base address of the allocation that this pointer belongs to.
    /// This address takes the role of `AllocId` in Miri's pointers since `AllocId` is not consistent across executions.
```

src/concurrency/genmc/helper.rs
RalfJung: No need to, this is the kind of low-level operation that needs direct access to the scalar.

src/concurrency/genmc/helper.rs
RalfJung: `value_scalar_int.into()` should work.

src/concurrency/genmc/helper.rs
```diff
@@ -113,40 +115,25 @@ pub fn genmc_scalar_to_scalar<'tcx>(
...
+    let alloc_id = ecx
+        .alloc_id_from_addr(
+            /* base_addr */ scalar.extra,
+            /* alloc_size */ 0,
+            /* only_exposed_allocations */ false,
+        )
+        .unwrap();
+    // FIXME(genmc,borrow tracking): Borrow tracking not yet supported.
+    let provenance = machine::Provenance::Concrete { alloc_id, tag: BorTag::default() };
+    let ptr = interpret::Pointer::new(provenance, /* offset */ Size::from_bytes(scalar.value));
```
RalfJung:
```suggestion
    let ptr = interpret::Pointer::new(provenance, Size::from_bytes(scalar.value));
```

src/concurrency/genmc/scheduling.rs
RalfJung: That's abusing `TerminationInfo::Exit`...

Can't we get error info from GenMC and call `throw_ub_format` with that?

Or, if this can only occur due to some bug in GenMC, we can also just panic.

tests/genmc/fail/data_race/atomic_ptr_dealloc_write_race.rs
RalfJung, line 1: Is this different from a hypothetical unsync_free test (that matches unsync_malloc)?

tests/genmc/fail/data_race/atomic_ptr_alloc_write_race.rs
RalfJung: We have `//@error-in-other-file: ` but I don't know if it would work here. We'd probably have to change how that error is emitted.

tests/genmc/pass/atomics/atomic_ptr_ops.rs
```diff
@@ -4,84 +4,170 @@
...
-            Err(ptr) =>
-                if ptr != x_ptr || *ptr != *x_ptr {
-                    std::process::abort();
-                } else {
-                    *ptr = *ptr;
-                },
-        }
+fn assert_equals<T: Eq + Copy + Debug>(a: T, b: T) {
+    if a != b {
+        writeln!(MiriStderr, "{:?}, {:?}", a, b).unwrap_or_else(|_| std::process::abort());
```
RalfJung:
```suggestion
        writeln!(MiriStderr, "{:?}, {:?}", a, b).ok();
```
This can't actually fail, let's not clutter the code.

src/alloc_addresses/mod.rs
```diff
@@ -243,7 +237,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         }?;
 
         // We only use this provenance if it has been exposed, or if the caller requested also non-exposed allocations
```
RalfJung: This comment needs updating.

src/alloc_addresses/mod.rs
RalfJung, line 1: The genmc stuff in `free_alloc_id` can also be removed now I think?

src/concurrency/genmc/helper.rs
```diff
@@ -83,6 +84,7 @@ pub fn split_access(address: Size, size: Size) -> impl Iterator<Item = (u64, u64
 /// Pointers with `Wildcard` provenance are not supported.
 pub fn scalar_to_genmc_scalar<'tcx>(
     ecx: &MiriInterpCx<'tcx>,
+    exposed_allocs_map: &RefCell<ExposedAllocsMap>,
```
RalfJung: "exposed" has a very specific meaning in Miri, and that's not how you are using this term here.

Maybe `genmc_shared_allocs_map`?

src/concurrency/genmc/mod.rs
```diff
@@ -268,13 +275,13 @@ impl GenmcCtx {
     ) -> InterpResult<'tcx, Scalar> {
         assert!(!self.get_alloc_data_races(), "atomic load with data race checking disabled.");
         let genmc_old_value = if let Some(scalar) = old_val {
-            scalar_to_genmc_scalar(ecx, scalar)?
+            scalar_to_genmc_scalar(ecx, &self.exec_state.exposed_allocs_map, scalar)?
```
RalfJung: Can you just pass `self` instead?

src/concurrency/genmc/mod.rs
```diff
@@ -905,7 +927,10 @@ impl GenmcCtx {
 }
 
 impl VisitProvenance for GenmcCtx {
-    fn visit_provenance(&self, _visit: &mut VisitWith<'_>) {
-        // We don't have any tags.
+    fn visit_provenance(&self, visit: &mut VisitWith<'_>) {
+        let exposed_allocs_map = self.exec_state.exposed_allocs_map.borrow();
+        for alloc_id in exposed_allocs_map.values().cloned() {
```
```suggestion
        for alloc_id in exposed_allocs_map.values().copied() {
```

genmc-sys/src/lib.rs
Patrick-6: Only on `extra`, or should all fields have comments?

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
Patrick-6: `status` is an [`std::optional<VerificationError>`](https://gitlab.inf.ethz.ch/public-plf/genmc/-/blob/cd01c12032bdd71df742b41c7817f99acc72e7ab/src/Verification/VerificationResult.hpp#L14) which is set to the error if one happened (indicated by `has_value`, which is like Rust `Option::is_some`).

src/concurrency/genmc/scheduling.rs
RalfJung: I am trying to understand what is going on with these errors...

So, IIUC, genmc internally has some state recording information about the most recent error. Furthermore, the "handle" functions can also return error information. So there's two error channels, which is kind of odd. If a handler returns an error, and then we ask for the most recent error, will we get the same error again? If that's the case, couldn't e.g. a load just return a boolean saying whether there was an error, instead of the `UniquePtr<CxxString>` that we currently have?

And then apparently it can happen that a handler does not fail but there's still an error recorded internally inside GenMC. The new code you added in this PR detects this as part of scheduling. (Why scheduling?) Is that expected or a bug? Can we work around it by checking the internal error state after calling such a problematic handler, so that to the rest of Miri this still looks like a normal handler that raised an error?

tests/genmc/fail/data_race/atomic_ptr_dealloc_write_race.rs
Patrick-6: Yes: this test is malloc --> free --> use allocation in different thread (use-after-free across threads) (malloc and free are implicit here, since `z` is stack allocated).
The not yet added test is malloc --> free in other thread without synchronization (race between malloc and free).

genmc-sys/src/lib.rs
RalfJung: For this type it probably makes sense to document all fields. (Then you can remove the -- implicit -- reference to `is_init` from the type-level docs.)

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
RalfJung: But then we ignore the actual `VerificationError` here?

tests/genmc/fail/data_race/atomic_ptr_dealloc_write_race.rs
RalfJung: But here the error is reported by Miri, not by GenMC. Does GenMC not detect a race between the write in thread 2 and the dealloc in thread 1?

This is concerning since a difference in scheduling could make the write happen first, so Miri wouldn't see an issue, but there's still a race.

src/concurrency/genmc/scheduling.rs
Patrick-6: > Can't we get error info from GenMC and call `throw_ub_format` with that?

We could, but that would point to the wrong span (the span at the time of the scheduling call, not the span when the error happened).

> So, IIUC, genmc internally has some state recording information about the most recent error. Furthermore, the "handle" functions can also return error information. So there's two error channels, which is kind of odd. If a handler returns an error, and then we ask for the most recent error, will we get the same error again? If that's the case, couldn't e.g. a load just return a boolean saying whether there was an error, instead of the `UniquePtr<CxxString>` that we currently have?

Originally, GenMC only had the second error channel. With LLI, it didn't matter if the execution continued after an error, instead, the execution will be aborted at the next scheduling call, which checks if an error happened and returns no next thread to schedule in that case.

With Miri, we want to use Miri's span info, so we need to know immediately when an error happens. This can either happen by:
- polling GenMC for errors after every (fallible) operation (requires knowing which operations are fallible)
- returning a boolean from every fallible operation, then asking for the error afterwards
- returning an optional error after every fallible operation

We went with the last option, since returning an error better shows the meaning compared to a bool, and is not that much harder to implement.

There's 2 problems: Not all fallible operations currently return an error, and some errors do not have a corresponding span (they are about the entire execution).
This change to scheduling is to fix the first of these problems, but I think I found a better solution, by checking for the error in the cases where we currently don't get an error return (e.g., `handle_free` [on the C++ side](https://github.com/Patrick-6/miri/blob/096e1c61186c3af165e05715bd03f7b8d7a05c60/genmc-sys/cpp/src/MiriInterface/EventHandling.cpp#L256-L257) and [the Rust side](https://github.com/Patrick-6/miri/blob/096e1c61186c3af165e05715bd03f7b8d7a05c60/src/concurrency/genmc/mod.rs#L659-L662)).
We should probably still have this change to scheduling, but panic (since we should be catching all errors immediately).

tests/genmc/fail/data_race/atomic_ptr_alloc_write_race.rs
Patrick-6: I managed to fix it with the change I've mentioned in another comment about polling for an error in [`handle_free`](https://github.com/Patrick-6/miri/blob/096e1c61186c3af165e05715bd03f7b8d7a05c60/genmc-sys/cpp/src/MiriInterface/EventHandling.cpp#L256-L257), and [the malloc-free-race test](https://github.com/Patrick-6/miri/blob/096e1c61186c3af165e05715bd03f7b8d7a05c60/tests/genmc/fail/atomics/atomic_ptr_unsync_malloc_free.rs#L26-L37) now [gives a proper error span](https://github.com/Patrick-6/miri/blob/096e1c61186c3af165e05715bd03f7b8d7a05c60/tests/genmc/fail/atomics/atomic_ptr_unsync_malloc_free.stderr).

src/concurrency/genmc/scheduling.rs
RalfJung: > I think I found a better solution, by checking for the error in the cases where we currently don't get an error return (e.g., handle_free [on the C++ side](https://github.com/Patrick-6/miri/blob/096e1c61186c3af165e05715bd03f7b8d7a05c60/genmc-sys/cpp/src/MiriInterface/EventHandling.cpp#L256-L257) and [the Rust side](https://github.com/Patrick-6/miri/blob/096e1c61186c3af165e05715bd03f7b8d7a05c60/src/concurrency/genmc/mod.rs#L659-L662)).
We should probably still have this change to scheduling, but panic (since we should be catching all errors immediately).

I think that's exactly what I suggested above, so -- sounds good.

tests/genmc/fail/data_race/atomic_ptr_dealloc_write_race.rs
Patrick-6: Isn't that error detected by the rustc const eval machinery ([this error here](https://github.com/rust-lang/rust/blob/5b9007bfc358817cf066ee27c6b440440727d3a7/compiler/rustc_const_eval/messages.ftl#L329C1-L330))? This is [the same message shown in the stderr](https://github.com/rust-lang/miri/blob/3f83c5e88783062c343bc044019961c9850564d1/tests/genmc/fail/atomics/atomic_ptr_use_after_free.stderr#L2C1-L2C99).
I would assume that GenMC doesn't even know about this memory access since the error is detected before calling into GenMC.

Patrick-6: It is possible to modify the test so that the write is scheduled before the free, in which case the rustc const eval doesn't detect the error, but GenMC still detects the data race (as soon as the memory is deallocated):
```
error: Undefined Behavior: Attempt to access freed memory
  --> tests/genmc/fail/atomics/atomic_ptr_use_after_free_rev.rs:44:13
   |
44 |             }),
   |             ^ Undefined Behavior occurred here
```
The line it points to is where the scope of the stack allocation `z` ends.
This is not a very clear error since we currently can't show the counterpart (the write), but it doesn't miss the data race.

RalfJung: > It is possible to modify the test so that the write is scheduled before the free, in which case the rustc const eval doesn't detect the error, but GenMC still detects the data race (as soon as the memory is deallocated):

Okay, please add revisions to make the test cover both cases.

RalfJung: > I would assume that GenMC doesn't even know about this memory access since the error is detected before calling into GenMC.

Ah indeed, we call `before_memory_deallocation` *after* checking that the pointer points to an actual allocation. I guess that's consistent with the read/write hooks.

genmc-sys/src/lib.rs
```diff
@@ -166,16 +166,15 @@ mod ffi {
...
-    /// If `extra` is non-zero, it contains information about the provenance of the pointer stored in `value`.
-    /// `extra` contain the base address of the allocation that this pointer belongs to, since `AllocId` is not consistent across executions.
-    /// Operations on values in GenMC (e.g., `fetch_add`) preserve the `extra` of the left argument (`left.fetch_add(right, ...)`).
     #[derive(Debug, Clone, Copy)]
     struct GenmcScalar {
+        /// If the `extra` field is zero, `value` represents a plain integer.
+        /// If the `extra` field is non-zero, it contains information about the provenance of the pointer stored in `value`.
         value: u64,
+        /// This field is either zero, or the base address of the allocation that this pointer belongs to.
+        /// Operations on `SVal` in GenMC (e.g., `fetch_add`) preserve the `extra` of the left argument (`left.fetch_add(right, ...)`).
```
```suggestion
        /// The raw byte-level value (discarding provenance, if any) of this scalar.
       value: u64,
        /// This is zero for integer values. For pointers, this encodes the provenance by
        /// storing the base address of the allocation that this pointer belongs to.
        /// Operations on `SVal` in GenMC (e.g., `fetch_add`) preserve the `extra` of the left argument (`left.fetch_add(right, ...)`).
```

genmc-sys/src/lib.rs
```diff
@@ -166,16 +166,15 @@ mod ffi {
...
         extra: u64,
+        /// CXX doesn't support `std::optional` currently, so we need to use an extra `bool` to define whether this value is initialized or not.
```
```suggestion
        /// Indicates whether this value is initialized. If this is `false`, the other fields do not matter.
        /// (Ideally we'd use `std::optional` but CXX does not support that.)
```

tests/genmc/pass/atomics/atomic_ptr_roundtrip.rs
RalfJung, line 1: Can you add a test as follows (doesn't even need concurrency), to ensure that indeed provenance is properly preserved in genmc:
- Create two pointers `p` and `q` with the same address but different provenance. Let's say `p` is valid but `q` is dangling.
- Store `p` in an atomic, load it, deref the pointer -- this should be fine.
- Store `q` in an atomic, load it, deref the pointer -- this should be UB.

Patrick-6: I've added 2 tests for this, one where we create the pointer using non-atomic operations, and one where we use `fetch_byte_{add,sub}` to create the invalid pointer on the GenMC side. (Both work as expected)

src/concurrency/genmc/mod.rs
```diff
@@ -644,7 +649,10 @@ impl GenmcCtx {
         let curr_thread = machine.threads.active_thread();
         let genmc_tid = thread_infos.get_genmc_tid(curr_thread);
 
-        self.handle.borrow_mut().pin_mut().handle_free(genmc_tid, address.bytes());
+        if self.handle.borrow_mut().pin_mut().handle_free(genmc_tid, address.bytes()) {
+            // An error was detected, so we get the error string from GenMC.
+            throw_ub_format!("{}", self.try_get_error().unwrap());
+        }
```
Patrick-6, line 656: Which approach do you prefer?
- Return a string when an error happens (this would allow us to add more info to GenMC in the C++ function where the error happens) (this is used in all other `handle_*` functions currently)
- Return a boolean, then call `self.try_get_error()` instead. (Only used for `handle_free` currently).

src/concurrency/genmc/helper.rs
```diff
@@ -129,10 +128,11 @@ pub fn genmc_scalar_to_scalar<'tcx>(
     }
     // `extra` is non-zero, we have a pointer.
     // When we get a pointer from GenMC, then we must have sent it to GenMC before in the same execution (since the reads-from relation is always respected).
-    let alloc_id = *exposed_allocs_map.borrow().get(&scalar.extra).unwrap();
+    let alloc_id =
+        *genmc_ctx.exec_state.genmc_shared_allocs_map.borrow().get(&scalar.extra).unwrap();
```
RalfJung: Isn't this a HashMap?
```suggestion
        genmc_ctx.exec_state.genmc_shared_allocs_map.borrow()[&scalar.extra];
```

src/concurrency/genmc/mod.rs
RalfJung: This is temporary anyway, right? So I don't really care. Just put a FIXME somewhere to indicate that we'll want to do error handling like the other handlers do eventually.

tests/genmc/fail/atomics/atomic_ptr_invalid_provenance_make.rs
```diff
@@ -0,0 +1,48 @@
...
+use std::sync::atomic::AtomicPtr;
+use std::sync::atomic::Ordering::*;
+
+#[unsafe(no_mangle)]
+fn miri_start(_argc: isize, _argv: *const *const u8) -> isize {
+    unsafe {
+        let atomic_ptr = AtomicPtr::new(std::ptr::null_mut());
+        let mut a = Box::new(0u64);
+        let mut b = Box::new(0u64);
+        // FIXME: use `Box::as_mut_ptr` once stabilized (`https://github.com/rust-lang/rust/issues/129090`).
```
RalfJung: There's no reason for such a FIXME. It doesn't really matter when whether you use the method or not, and if you want to use it you can -- just enable the feature gate.

tests/genmc/fail/atomics/atomic_ptr_invalid_provenance.rs
```diff
@@ -0,0 +1,48 @@
+//@compile-flags: -Zmiri-genmc -Zmiri-disable-stacked-borrows
+
+// Test that we can distinguish two pointers with the same address, but different provenance, after they are sent to GenMC and back.
+// We have two variants, one where we send such a pointer to GenMC, and one where we make it on the GenMC side.
```
RalfJung, line 5: These two variants should be almost the same except for a few lines, so please use revisions to avoid the code duplication.

Also, please use the `with_addr` method on pointers instead of manually repeating it in your code.

tests/genmc/fail/atomics/atomic_ptr_invalid_provenance_make.rs
```diff
@@ -0,0 +1,48 @@
...
+
+        // Create a pointer with the provenance of `b_ptr`, but the address of `a_ptr` on the GenMC side.
+        atomic_ptr.store(b_ptr, Relaxed);
+        atomic_ptr.fetch_byte_add(a_ptr as usize, Relaxed);
+        atomic_ptr.fetch_byte_sub(b_ptr as usize, Relaxed);
+        let ptr = atomic_ptr.load(Relaxed);
+        if a_ptr as usize != ptr as usize {
+            std::process::abort();
+        }
+        *ptr = 44; //~ ERROR: Undefined Behavior
```
RalfJung:
```suggestion
        // This pointer has the same address as `a_ptr`, but not the same
        // provenance, so writing to it fails.
        *ptr = 44; //~ ERROR: points to before the beginning of the allocation
```

tests/genmc/fail/data_race/atomic_ptr_alloc_race.rs
RalfJung, line 1: The naming of your new "unsync"/data race tests is all over the place. I'd suggest to follow the existing naming scheme of the non-genmc tests:  `alloc_write_race`, `alloc_dealloc_race`, etc (where the first access listed is the one that goes "first" in Miri's execution order).

Also why are these not in the data_race folder?

tests/genmc/fail/data_race/atomic_ptr_write_dealloc_race.rs
```diff
@@ -0,0 +1,50 @@
+//@compile-flags: -Zmiri-genmc -Zmiri-disable-stacked-borrows
+
+// Test that use-after-free bugs involving atomic pointers are detected in GenMC mode.
+// This variant checks that the data race is still detected, even if the write happens before the free.
```
RalfJung: I think these will just be `dealloc_write_race` vs `write_dealloc_race` then.

tests/genmc/pass/data-structures/ms_queue_dynamic.rs
```diff
@@ -149,13 +153,13 @@ impl MyStack {
 #[unsafe(no_mangle)]
 fn miri_start(_argc: isize, _argv: *const *const u8) -> isize {
     // We try multiple different parameters for the number and types of threads:
-    let (readers, writers, rdwr) = if cfg!(any(default_RW2, spinloop_assume_RW2)) {
-        (0, 0, 2)
-    } else if cfg!(any(default_R1W1)) {
-        (1, 1, 0)
-    } else {
-        // default_R1W2, spinloop_assume_R1W2
+    let (readers, writers, rdwr) = if cfg!(any(default_R1W3, spinloop_assume_R1W3)) {
```
RalfJung: Looks like rdwr is always 0 so please remove it.

tests/genmc/pass/data-structures/treiber_stack_dynamic.rs
```diff
@@ -106,13 +108,13 @@ fn miri_start(_argc: isize, _argv: *const *const u8) -> isize {
     unsafe { STACK.top.store(std::ptr::null_mut(), Relaxed) };
 
     // We try multiple different parameters for the number and types of threads:
-    let (readers, writers, rdwr) = if cfg!(any(default_RW2, spinloop_assume_RW2)) {
-        (0, 0, 2)
-    } else if cfg!(any(default_R2W2, spinloop_assume_R2W2)) {
-        (2, 2, 0)
-    } else {
-        // default_R1W2, spinloop_assume_R1W2
+    let (readers, writers, rdwr) = if cfg!(any(default_R1W3, spinloop_assume_R1W3)) {
```
RalfJung: Same here, please remove rdwr.

src/concurrency/genmc/scheduling.rs
RalfJung: Didn't we agree that this should just `panic!` now?

tests/genmc/fail/atomics/atomic_ptr_invalid_provenance.rs
Patrick-6: This doesn't work for the `malloc_free` and `malloc_access` tests, right? Each variant there complains about an unmatched error diagnostic from the other revision, like this:
```rust
#[cfg(never_true)]
let x = 0; //~ ERROR: Undefined Behavior
           ^^^ (diagnostic not found)
```
(For the invalid provenance test I can just ensure both errors happen on the same line)

src/concurrency/genmc/mod.rs
Patrick-6: I'm not sure how the final error handling will look like, so I can't say which of the two approaches is better. I've added a generic FIXME about error handling for now, like for the other places related to that.

src/concurrency/genmc/scheduling.rs
Patrick-6: I've made it a panic now, but still have it print the detected error in the panic message, since this case is both a bug in GenMC (missing error return for a `handle*` function), _and_ a bug in the user program (the detected error is still real).

Patrick-6:
oops

r? @RalfJung

tests/genmc/fail/atomics/atomic_ptr_invalid_provenance.rs
RalfJung: You can make diagnostics conditional on the revision
```
//~[rev1] ERROR: blah
```

tests/genmc/fail/data_race/atomic_ptr_alloc_race.rs
Patrick-6: I've moved and renamed the tests.

tests/genmc/fail/data_race/atomic_ptr_alloc_free_race.rs
RalfJung, line 1: It's called `dealloc`, not `free`, for the existing tests and the other tests you are adding here.

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
```diff
@@ -24,8 +24,10 @@ auto MiriGenmcShim::schedule_next(
 
     if (const auto result = GenMCDriver::scheduleNext(threads_action_))
         return SchedulingResult { ExecutionState::Ok, static_cast<int32_t>(result.value()) };
-    if (GenMCDriver::isExecutionBlocked())
+    if (getExec().getGraph().isBlocked())
         return SchedulingResult { ExecutionState::Blocked, 0 };
+    if (getResult().status.has_value())
```
RalfJung:
```suggestion
    if (getResult().status.has_value()) // the "value" here is a `VerificationError`
```

tests/genmc/fail/data_race/atomic_ptr_write_dealloc_race.rs
RalfJung: It's not clear here what "this variant" is referring to -- what's the other variant? The comment should make that clear.

tests/genmc/fail/data_race/atomic_ptr_alloc_free_race.rs
Patrick-6: Should I rename `atomic_ptr_double_free.rs` too? I think `double-free` is more commonly used for that kind of bug than `double-dealloc` (at least in C/C++).

RalfJung: I'd frame it as a `dealloc_dealloc` race.

Patrick-6: I don't think the bug is that there is a race between two deallocations, more that freeing the same memory twice is never allowed (whether synchronized or not).

RalfJung: Why would we have a genmc test for that? It's an entirely standard Miri check.

Patrick-6: To ensure that such bugs are still detected in GenMC mode?

RalfJung: You are not suggesting that we duplicate the entire `tests/fail` suite for genmc mode, right?

What makes double-free special that we should duplicate it?

Patrick-6: I guess it isn't special, except that the bug should always be detected in GenMC mode, even without running the test with multiple seeds or running the buggy code in a loop.

You've mentioned that it may be good to run the entire Miri test suite in GenMC mode too, so double-free bugs will be tested once that is implemented.

Should I remove the test?

tests/genmc/fail/atomics/atomic_ptr_double_free.rs
```diff
@@ -0,0 +1,49 @@
...
+    unsafe {
+        let ids = [
+            spawn_pthread_closure(|| {
+                let a: *mut u64 = alloc(Layout::new::<u64>()) as *mut u64;
+                X.store(a, SeqCst);
+                // If the other thread runs here, there will be a double free.
+                let b = X.swap(std::ptr::null_mut(), SeqCst);
+                if b.is_null() {
+                    std::process::abort();
+                }
```
RalfJung:
```suggestion
```

tests/genmc/fail/data_race/atomic_ptr_alloc_free_race.rs
RalfJung: > I guess it isn't special, except that the bug should always be detected in GenMC mode, even without running the test with multiple seeds or running the buggy code in a loop.

I see. Yeah we can do that. Then the comment should be clarified to say that we specifically check that genmc can find the one interleaving where there's UB.

tests/genmc/fail/atomics/atomic_ptr_double_free.rs
```diff
@@ -0,0 +1,49 @@
...
+fn miri_start(_argc: isize, _argv: *const *const u8) -> isize {
+    // FIXME(genmc,HACK): remove this initializing write once Miri-GenMC supports mixed atomic-non-atomic accesses.
+    X.store(std::ptr::null_mut(), SeqCst);
+
+    unsafe {
+        let ids = [
+            spawn_pthread_closure(|| {
+                let a: *mut u64 = alloc(Layout::new::<u64>()) as *mut u64;
+                X.store(a, SeqCst);
+                // If the other thread runs here, there will be a double free.
```
RalfJung:
```suggestion
                // We have to yield to the other thread exactly here to reproduce the double-free.
```

RalfJung:
This looks great, thanks! Please [squash the commits](https://rustc-dev-guide.rust-lang.org/git.html#squash-your-commits), then we can land this. Please use the `--keep-base` flag when squashing so that the force-push diff is easier to review; do *not* rebase on the latest master branch. Then write `@rustbot ready` after you force-pushed the squashed PR.

@rustbot author

Patrick-6:
Done! @rustbot ready

PR #4614, opened 2025-10-02
Implement std::sync::Mutex interception for GenMC mode.

Patrick-6:
This PR adds support for Miri to intercept uses of `std::sync::Mutex` using the rustc diagnostic items added in https://github.com/rust-lang/rust/pull/141690.


genmc-sys/cpp/include/MiriInterface.hpp
```diff
@@ -234,6 +245,14 @@ struct MiriGenmcShim : private GenMCDriver {
...
      */
     std::vector<Action> threads_action_;
+
+    /**
+     * Map of already used annotation ids (e.g., for mutexes).
+     * FIXME(GenMC): Ensure that these are consistent with how GenMC expects them once Miri supports
+     * multithreading in GenMC mode.
+     */
+    std::unordered_map<uint64_t, ModuleID::ID> annotation_id {};
+    ModuleID::ID annotation_id_counter = 0;
```
Patrick-6: I'm not sure how this should look like for multithreading, but just making this map global should work (but might be slow).
GenMC uses the `llvm::instruction` of the mutex lock call from the interpreted program, and assigns each instruction an ID. Maybe Miri could also do something similar, e.g, using the span for identification.
The current solution should be fine as long as there is no multithreading support for GenMC mode.

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
```diff
@@ -0,0 +1,142 @@
...
+        EventDeps()
+    );
+    if (const auto* err = std::get_if<VerificationError>(&load_ret))
+        return MutexLockResultExt::from_error(format_error(*err));
+    // If we get a `Reset`, GenMC decided that this lock operation should not yet run, since it
+    // would not acquire the mutex. Like the handling of the case further down where we read a `1`
+    // ("Mutex already locked"), Miri should call the handle function again once the current thread
+    // is scheduled by GenMC the next time.
+    if (std::holds_alternative<Reset>(load_ret))
+        return MutexLockResultExt::ok(false);
```
Patrick-6: We don't trigger this branch with any of our current mutex tests, but this is the same behavior that GenMC has for mutex lock ([here, see the `!ret.has_value()`](https://github.com/MPI-SWS/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/src/Runtime/Execution.cpp#L2921-L2922)).

src/concurrency/genmc/intercept.rs
```diff
@@ -1,16 +1,137 @@
...
+            .as_genmc_ref()
+            .expect("This function should only be called in GenMC mode.");
+
+        let get_mutex_call_infos = || {
+            // assert!(!args.is_empty());
+            assert_eq!(args.len(), 1);
+            let arg = this.copy_fn_arg(&args[0]);
+            let addr = this.read_target_usize(&arg)?;
+            // FIXME(genmc): assert that we have at least 1 byte.
+            // FIXME(genmc): maybe use actual size of mutex here?.
```
Patrick-6: Ah, missed a comment at the top here.

For the bottom comments, how can I assert that the location pointed to by `arg` is at least 1 byte large?

src/concurrency/genmc/intercept.rs
```diff
@@ -1,16 +1,137 @@
...
+                            .handle_mutex_lock(genmc_curr_thread, addr, size);
+                        if let Some(error) = result.error.as_ref() {
+                            // FIXME(genmc): improve error handling.
+                            throw_ub_format!("{}", error.to_string_lossy());
+                        }
+                        // FIXME(genmc): The reported error message is bad, it does not point to the second lock call involved in the deadlock.
+                        // FIXME(genmc): there can be cases where not acquiring a mutex after the second attempt is *not* a deadlock. Reliably detecting deadlocks requires extra analysis (in GenMC).
+                        if !result.is_lock_acquired {
+                            throw_unsup_format!("Could not lock Mutex, which may indicate a deadlock. (GenMC mode does not fully support deadlock detection yet).")
+                        }
```
Patrick-6: I'm not sure how to frame this error (thus "unsupported" error for now). From my discussion with Michalis I think not acquiring the mutex on the second try indicates a deadlock in most cases, but not always. So far I wasn't able to create a test case without a deadlock that triggers this though.

src/concurrency/genmc/intercept.rs
```diff
@@ -1,16 +1,137 @@
...
+            // FIXME(genmc): maybe use actual size of mutex here?.
+
+            let thread_infos = genmc_ctx.exec_state.thread_id_manager.borrow();
+            let curr_thread = this.machine.threads.active_thread();
+            let genmc_curr_thread = thread_infos.get_genmc_tid(curr_thread);
+            interp_ok((genmc_curr_thread, addr, 1))
+        };
+
+        use rustc_span::sym;
+        interp_ok(if this.tcx.is_diagnostic_item(sym::sys_mutex_lock, instance.def_id()) {
```
Patrick-6: Is it possible to `match` on `def_id()`? So far I didn't find a way, since it must go through the `is_diagnostic_item` call.

src/concurrency/genmc/scheduling.rs
```diff
@@ -63,15 +63,24 @@ fn get_function_kind<'tcx>(
...
         }
+        // Some functions, like `sys::Mutex::lock` are intercepted by Miri, instead of running their implementation.
+        // These should be treated as atomics, and some of them have load semantics.
+        if ecx.tcx.is_diagnostic_item(rustc_span::sym::sys_mutex_lock, callee_def_id)
+            || ecx.tcx.is_diagnostic_item(rustc_span::sym::sys_mutex_try_lock, callee_def_id)
+        {
+            return interp_ok(MaybeAtomic(ActionKind::Load));
+        } else if ecx.tcx.is_diagnostic_item(rustc_span::sym::sys_mutex_unlock, callee_def_id) {
+            return interp_ok(MaybeAtomic(ActionKind::NonLoad));
+        }
```
Patrick-6: Future additions to the list of intercepted functions should be added here too, otherwise they will not be treated as atomics, messing up the scheduling. Should there be a comment about this somewhere, and if yes, where?

Patrick-6 (commented):
r? @RalfJung 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4614/017326671dac651ff38216a40b017ab77da0ced8..9a2ec6983a9c99b09aadc2b42dfaf821904c36c6)*

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
```diff
@@ -0,0 +1,142 @@
...
+        // of the store.
+        const bool* is_coherence_order_maximal_write = std::get_if<bool>(&store_ret);
+        ERROR_ON(
+            nullptr == is_coherence_order_maximal_write,
+            "Unimplemented: store part of mutex try_lock returned unexpected result."
+        );
+    } else {
+        // We did not acquire the mutex, so we tell GenMC to block the thread until we can acquire
+        // it.
+        GenMCDriver::handleAssume(inc_pos(thread_id), AssumeType::Spinloop);
```
RalfJung: Where does the "until we can acquire it" part happen? I see there is `Spinloop` instead of `User`, but it doesn't say which mutex we are waiting for.

genmc-sys/cpp/include/MiriInterface.hpp
RalfJung: That sounds like genmc is using IDs that are static and identify the *instruction*, whereas you are using the address which is dynamic and identifies the mutex *instance*.

What does genmc actually need the ID for an which properties does it need to have?

genmc-sys/cpp/include/MiriInterface.hpp
```diff
@@ -234,6 +245,14 @@ struct MiriGenmcShim : private GenMCDriver {
      * indices, since GenMC expects us to do that.
      */
     std::vector<Action> threads_action_;
+
+    /**
+     * Map of already used annotation ids (e.g., for mutexes).
+     * FIXME(GenMC): Ensure that these are consistent with how GenMC expects them once Miri supports
+     * multithreading in GenMC mode.
+     */
+    std::unordered_map<uint64_t, ModuleID::ID> annotation_id {};
```
RalfJung: Judging from the rest of the code here, this uint64_t is the address of the mutex, right? That must be explained in a comment here.

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
```diff
@@ -0,0 +1,142 @@
...
+    } else {
+        annot_id = it->second;
+    }
+    const auto annot = std::move(Annotation(
+        AssumeType::Spinloop,
+        Annotation::ExprVP(NeExpr<AnnotID>::create(
+                               RegisterExpr<AnnotID>::create(ASize(size).getBits(), annot_id),
+                               ConcreteExpr<AnnotID>::create(ASize(size).getBits(), SVal(1))
+        )
+                               .release())
```
RalfJung: What is all of this about? Why does `ASize(size).getBits()` show up twice and what does the `1` mean?

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
```diff
@@ -0,0 +1,142 @@
...
+    ));
+
+    // Mutex starts out unlocked, so we always say the previous value is "unlocked".
+    const auto old_val = SVal(0);
```
RalfJung: This comment makes no sense at all. This is the usual CAS from 0 to 1 that acquire a mutex, isn't it? It has nothing to do with whether a mutex starts out locked or unlocked upon initialization.

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
```diff
@@ -0,0 +1,142 @@
...
+        // it.
+        GenMCDriver::handleAssume(inc_pos(thread_id), AssumeType::Spinloop);
+    }
+    return MutexLockResultExt::ok(is_lock_acquired);
+}
+
+auto MiriGenmcShim::handle_mutex_try_lock(ThreadId thread_id, uint64_t address, uint64_t size)
+    -> MutexLockResult {
+    auto& currPos = threads_action_[thread_id].event;
+    // Mutex starts out unlocked, so we always say the previous value is "unlocked".
```
RalfJung: This comment also makes no sense to me.

src/concurrency/genmc/intercept.rs
```diff
@@ -1,16 +1,136 @@
...
 };
 
 // Handling of code intercepted by Miri in GenMC mode, such as assume statement or `std::sync::Mutex`.
 
-/// Other functionality not directly related to event handling
 impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    /// Given a `ty::Instance<'tcx>`, do any required special handling.
+    /// Returns true if this `instance` should be skipped (i.e., no MIR should be executed for it).
+    fn check_genmc_intercept_function(
```
RalfJung:
```suggestion
    fn genmc_intercept_function(
```
That should be clear enough.

src/concurrency/genmc/intercept.rs
RalfJung: What does LLI do?
This error isn't very helpful for users. "Maybe this is a deadlock and maybe it's... what else?"

What I'd expect here is that this just tries again in a loop.

src/concurrency/genmc/intercept.rs
RalfJung: Yeah if-then-else is what I'd expect here.

At least the more complicated ones like `lock` should be implemented as a separate method.

src/concurrency/genmc/intercept.rs
```diff
@@ -1,16 +1,136 @@
...
+        let get_mutex_call_infos = || {
+            assert_eq!(args.len(), 1);
+            let arg = this.copy_fn_arg(&args[0]);
+            let addr = this.read_target_usize(&arg)?;
+            // FIXME(genmc): assert that we have at least 1 byte.
+            // FIXME(genmc): maybe use actual size of mutex here?.
+
+            let thread_infos = genmc_ctx.exec_state.thread_id_manager.borrow();
+            let curr_thread = this.machine.threads.active_thread();
+            let genmc_curr_thread = thread_infos.get_genmc_tid(curr_thread);
```
RalfJung: This looks like you may want a little helper `active_thread_genmc_tid` or so.

And then the closure isn't worth its weight any more. It's anyway usually easier to follow when each shim starts by parsing its arguments, thus making it clear what the signature is. You probably want a little helper for that, something similar to `check_shim_sig_lenient`.

src/concurrency/genmc/scheduling.rs
RalfJung: A comment in `genmc_intercept_function` sounds good. A bit of refactoring of that function to separate the dispatch from the actual logic of the intercepted functions should help avoid drowning that comment in the rest of the code.

tests/genmc/pass/shims/mutex_simple.rs
```diff
@@ -0,0 +1,82 @@
...
+fn miri_start(_argc: isize, _argv: *const *const u8) -> isize {
+    std::panic::abort_unwind(main_);
+    0
+}
+
+fn main_() {
+    let mut guard = LOCK.lock().unwrap();
+    for &v in guard.iter() {
+        assert!(v == 1234); // Check that mutex values are initialized correctly
+    }
```
RalfJung: What is the point of this entirely random-looking check (and same below)?

tests/genmc/pass/intercept/mutex_simple.rs
```diff
@@ -0,0 +1,82 @@
...
+
+use crate::genmc::*;
+
+const REPS: u64 = if cfg!(reps3) {
+    3
+} else if cfg!(reps2) {
+    2
+} else {
+    1
+};
```
RalfJung: I don't see why it's worth trying this with different numbers of REPS.

src/concurrency/genmc/intercept.rs
Patrick-6: LLI just ignores this deadlock (for an equivalent C program that uses pthread mutexes), counting it as any other blocked execution in the final count.

RalfJung: So the thread is always blocked again if the lock can't be acquired? That sounds like the loop that I expected.

So what should happen here then is that after unblocking we simply recursively call the mutex lock interceptor again.

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
Patrick-6: An unlocked mutex is indicated by the value being `0`, but `old_val` is _not_ used as the "expected" value for a CAS however (this is just the load part of a CAS, no comparison happens during the load).
This value is used for the temporal mixing of atomic/non-atomic accesses from the previous PR. We inform GenMC, that _if_ there is no previous atomic store to this location, `old_val` is the value that should be used (which means this is the initial value we want to give any mutex).

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
Patrick-6: This is the same as GenMC does with LLI for mutex lock [here](https://github.com/MPI-SWS/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/src/Runtime/Execution.cpp#L2908-L2914).
This expression tells GenMC that the load is part of a spinloop that will stop looping if this expression holds (the value at that location being different from `SVal(1)`, where `1` indicates that the Mutex is locked).

`ASize(size).getBits()` shows up twice to avoid having too many local variable declarations (which you've told me repeatedly that I add too many of those ^^).
I can switch it to be this if you prefer:
EDIT: I think we shouldn't use `getBits`, since it uses `CHAR_BIT`, which can be more than 8 on some platforms. FWIU, Rust uses 8 bits per byte on all platforms, so I think it's better to use this:
```C++
const auto size_bits = size * 8;
// (or: inline size * 8)
```
I'll add a comments for the `SVal(1)` and meaning of the annotation.

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
Patrick-6: FWIU the spinloop annotation we add on the load at the start of this function is what GenMC uses to know when to schedule this thread again.

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
RalfJung: Oh, that's what you mean...

That value should be entirely irrelevant since we never do normal accesses to that location?

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
RalfJung: >  (which you've told me repeatedly that I add too many of those ^^).

I did, for cases where they were only used once. :)

CHAR_BIT is 8 on all targets we support so it doesn't matter much how you do the conversion.

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
RalfJung: So there's some sort of pattern matching here where the previous events are also taken into account?

That seems worth mentioning.

genmc-sys/cpp/include/MiriInterface.hpp
Patrick-6: The annotation ID is used to uniquely define a "register", and is passed as a parameter for the `RegisterExpr` class (see code and comment [here](https://github.com/MPI-SWS/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/src/Support/SExpr.hpp#L230-L242)). From the comment there I'm not sure what exact conditions must be met by this annotation ID, since it also mentions being a `void *` (but it is an `unsigned int` currently, see [ModuleID::ID](https://github.com/MPI-SWS/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/src/Static/ModuleID.hpp#L19-L24)). IIRC the current implementation for getting IDs for Miri is based on a discussion I had with Michalis.

src/concurrency/genmc/intercept.rs
Patrick-6: Yes, I had code like that at one point, but it required a loop [here](https://github.com/Patrick-6/miri/blob/0aefe3476ba97edeaa640b070a0d26e3d5b4a117/src/concurrency/thread.rs#L712-L720), since the unblocking may cause the thread to be blocked again, at which point we need to call into GenMC to schedule a different thread.
Also, we would miss deadlocks involving mutexes if we do that.

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
Patrick-6: It is only relevant for the first `lock` or `try_lock` call, since those would otherwise read uninit memory (they need to read `SVal(0)` so they can succeed).
For `unlock`, it shouldn't matter (since it should always happen _after_ a `lock` call), so we could also pass `std::nullopt` here.

tests/genmc/pass/shims/mutex_simple.rs
Patrick-6: I thought we should check that our interception and custom handling of the mutex functions doesn't cause any of these unrelated values to be modified, but it could be unnecessary.

RalfJung: Yeah this is unnecessary. It's also the wrong test for that; the concern here should be getting genmc's state confused so we'd want to do *atomic* reads. But without talking about the data layout of `Mutex` all that makes little sense.

genmc-sys/cpp/include/MiriInterface.hpp
RalfJung: "IIRC" does not exactly fill me with confidence. ;)

src/concurrency/genmc/intercept.rs
RalfJung: > Also, we would miss deadlocks involving mutexes if we do that.

An error saying "this may or may not be a deadlock, have fun figuring it out" isn't really better than no deadlock detection.

The question is what information genmc gives us. When and why does it unblock a thread waiting for a mutex? What does it mean if that thread then fails to acquire the mutex? These questions need to be answered by understanding genmc, not by poking at it with a stick.

RalfJung: If in doubt, we should do what the LLI frontend does. You said it "ignores" the deadlock; what does that mean? In the code, what does it do in that very place we are discussing here?

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
RalfJung: How do these locations usually get their values? Does genmc usually require a pthread_mutex_init call? Mutexes can have static initializers though and those don't have to be all-zero...

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
```diff
@@ -0,0 +1,142 @@
...
+    auto [it, inserted] = annotation_id.try_emplace(address, annotation_id_counter);
+    if (inserted) {
+        annot_id = annotation_id_counter++;
+    } else {
+        annot_id = it->second;
+    }
+    const auto annot = std::move(Annotation(
+        AssumeType::Spinloop,
+        Annotation::ExprVP(NeExpr<AnnotID>::create(
+                               RegisterExpr<AnnotID>::create(ASize(size).getBits(), annot_id),
```
RalfJung: After discussing with Michalis we concluded that the ID here should be just 0. LLI needs this to support annotations with multiple variables but inside GenMC, `RegisterExpr` marks the "value of the current expression" and the ID does not really matter.

src/concurrency/genmc/intercept.rs
RalfJung: After discussing with Michalis, we think the right thing to do here is to do the equivalent of `assume(false)`: mark the thread as blocked *again*, and it will now definitely not be woken up again.

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
RalfJung: There's a difference between "reset" and "didn't get the lock", as your comment also explains. So it seems best to actually expose this to Miri, where we can use it for sanity checks.

The protocol should be something like:
first time, we get the lock or reset. Second time, we either get or don't get the lock (in the latter case the thread is permanently blocked).

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
RalfJung: I think I understand now what happens; we need to just work on the comment a bit.
```suggestion
    // As usual, we need to tell GenMC which value was stored at this location before this atomic
    // access, if there previously was a non-atomic initializing access. We just always set this to 0,
    // indicating that the lock is not locked.
    const auto old_val = SVal(0);
```

src/concurrency/genmc/intercept.rs
```diff
@@ -1,16 +1,136 @@
...
+            .data_race
+            .as_genmc_ref()
+            .expect("This function should only be called in GenMC mode.");
+
+        let get_mutex_call_infos = || {
+            assert_eq!(args.len(), 1);
+            let arg = this.copy_fn_arg(&args[0]);
+            let addr = this.read_target_usize(&arg)?;
+            // FIXME(genmc): assert that we have at least 1 byte.
+            // FIXME(genmc): maybe use actual size of mutex here?.
```
RalfJung: Using the actual size of the pointee type of `arg` here would be good, yeah.

RalfJung (commented):
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4614/017326671dac651ff38216a40b017ab77da0ced8..cdcf6e09b2d176024a1ef30ccea31b5964a5b428)*

src/concurrency/genmc/intercept.rs
Patrick-6: How can I get the size of the pointee?
`arg` is an `OpTy`, which has a `layout` field, but I think this is this for the argument itself not the pointee. With `arg.layout.ty.pointee_metadata_ty_or_projection(this.tcx.tcx)` I can get the `Ty` of the pointee, but I couldn't figure out how to get the size of that type.

RalfJung: In this situation, the best way is probably to deref the pointer to get a place:
```rust
let place = this.deref_pointer(&arg)?;
let addr = place.ptr.addr();
let size = playe.layout.size;
```

src/concurrency/genmc/intercept.rs
```diff
@@ -9,64 +10,56 @@ use crate::{
...
+    fn parse_mutex_method_args(
+        &self,
+        args: &[rustc_const_eval::interpret::FnArg<'tcx, crate::Provenance>],
+    ) -> InterpResult<'tcx, MutexMethodArgs> {
+        assert_eq!(args.len(), 1, "Mutex lock/unlock/try_lock should take exactly 1 argument.");
+        let this = self.eval_context_ref();
+        let arg = this.copy_fn_arg(&args[0]);
+        // FIXME(genmc): use actual size of the pointee of `arg`.
+        let size = 1;
+        // GenMC does not support large accesses, we limit the size to the maximum access size.
```
RalfJung: Something is very strange if the mutex is too big for this.

Papering over that by clamping the size doesn't sound good. Let's just have an assertion in `handle_mutex_lock` (like we already have for some other handlers I think?)

src/concurrency/genmc/intercept.rs
```diff
@@ -9,64 +10,56 @@ use crate::{
...
-            // FIXME(genmc): assert that we have at least 1 byte.
-            // FIXME(genmc): maybe use actual size of mutex here?.
-
-            let thread_infos = genmc_ctx.exec_state.thread_id_manager.borrow();
-            let curr_thread = this.machine.threads.active_thread();
-            let genmc_curr_thread = thread_infos.get_genmc_tid(curr_thread);
-            interp_ok((genmc_curr_thread, addr, 1))
-        };
+#[derive(Clone, Copy)]
+struct MutexMethodArgs {
```
RalfJung: Why do we need a new *type* for this...??

Miri has hundreds of shim/intercept functions already. This should follow the same pattern as everything else, not make up something entirely new.

RalfJung: I have pushed a commit adjusting the shims here to better match the rest of Miri. That also gets rid of this type.

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
Patrick-6: By "permanently blocked", do you mean report that as a deadlock, or should I make it an `assume(false)`, like you've mentioned in another comment? (The current implementation reports it as a deadlock).

tests/genmc/fail/intercept/mutex_deadlock_std.stderr
Patrick-6, line 1: I've added this test temporarily to show the current error message we get for deadlocks in `sys::Mutex::lock`. Using either pthreads or `std::thread`, the deadlock error is shown 3 times, once on each of the two involved lock calls, and once where the join of the threads happens.

- Do you know why the error is shown on the join calls?
- For the lock calls, the error points towards the `sys::lock` function, not the `std::sync::Mutex::lock` function. Could this be changed by adding a `#[cfg_attr(miri, track_caller)]` to `sys::Mutex::lock` ([here](https://github.com/rust-lang/rust/blob/981353ca16018bb35e232684b5f91f0943bf45bf/library/std/src/sys/sync/mutex/futex.rs#L28-L35))?
- Which do you prefer, the pthread or the `std::thread` version of this test? The std thread version has more output due to the warnings about compare_exchange, and takes almost twice as long to finish (~1.1 sec vs ~2.1 sec on my machine), but is simpler in terms of code in the test file.

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
Patrick-6: There are also these two cases:
- getting the result "lock not acquired" AND it not being a reset on the first lock call (this happens in the `genmc/fail/mutex_deadlock.rs` test)
- getting "lock not acquired" on the second lock call (in the callback when the thread is unblocked).

Currently I report both cases as deadlocks, but I couldn't trigger with any test yet.

Patrick-6:
This failure is weird, the `mutex_poison.rs` test passes on my machine with the expected output, but not in CI?
Maybe something changed in the upstream implementation of one of the functions used in the test, so a rebase could potentially fix this?

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
RalfJung: > By "permanently blocked", do you mean report that as a deadlock, or should I make it an assume(false), like you've mentioned in another comment? 

I mean `assume(false)`.

> getting the result "lock not acquired" AND it not being a reset on the first lock call (this happens in the genmc/fail/mutex_deadlock.rs test)

Odd, I thought that was not possible. But regardless, any non-reset non-acquired should trigger `assume(false)`. We can probably avoid some code duplication then by simply having the unblock callback recursively call the lock handler function.

> getting "lock not acquired" on the second lock call (in the callback when the thread is unblocked).

That's also supposedly impossible.

tests/genmc/pass/intercept/mutex_poison.rs
RalfJung, line 1: Poisoning is entirely implemented inside the library on top of the basic lock we are intercepting. So how could this possibly *not* work if the lock itself works correctly...?

tests/genmc/fail/intercept/mutex_deadlock_std.stderr
RalfJung: > Do you know why the error is shown on the join calls?

When there is a deadlock, that means all threads are blocked, and Miri then shows where all threads are blocked. Miri can't know which threads "wait on" which other threads so they might all be part of the waiting loop.

> For the lock calls, the error points towards the sys::lock function, not the std::sync::Mutex::lock function. Could this be changed by adding a #[cfg_attr(miri, track_caller)] to sys::Mutex::lock ([here](https://github.com/rust-lang/rust/blob/981353ca16018bb35e232684b5f91f0943bf45bf/library/std/src/sys/sync/mutex/futex.rs#L28-L35))?

Yeah, it could. We haven't spend much effort on making deadlock errors nicer. One can see the lock call in the backtrace so all the necessary information is present at least.

> Which do you prefer, the pthread or the std::thread version of this test? The std thread version has more output due to the warnings about compare_exchange, and takes almost twice as long to finish (~1.1 sec vs ~2.1 sec on my machine), but is simpler in terms of code in the test file.

We can go for the no_std version then, with a FIXME saying to use std once that's not quite so slow any more.

RalfJung:
You probably need to rebase over https://github.com/rust-lang/miri/pull/4617 yeah.

src/concurrency/genmc/intercept.rs
```diff
@@ -27,22 +28,32 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let this = self.eval_context_mut();
         let genmc_ctx = this.machine.data_race.as_genmc_ref().unwrap();
 
+        let size = mutex.layout.size.bytes();
+        assert!(
+            size < MAX_ACCESS_SIZE,
```
RalfJung:
```suggestion
            size <= MAX_ACCESS_SIZE,
```
Equal to the max should be fine, right?

(Occurs again below)

Patrick-6:
The rebase fixed the previous CI failure, but there is [another one](https://github.com/rust-lang/miri/actions/runs/18273391512/job/52019964749?pr=4614#step:6:87) now, possibly spurious? Nothing in this PR should affect Windows, or Linux on Arm specifically.

RalfJung:
Yeah those are just network timeouts, you can ignore them.

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
```diff
@@ -0,0 +1,158 @@
...
+    -> MutexLockResult {
+    // This annotation informs GenMC about the condition required to make this lock call succeed.
+    // It stands for `value_read_by_load != MUTEX_LOCKED`.
+    const auto size_bits = size * 8;
+    const auto annot = std::move(Annotation(
+        AssumeType::Spinloop,
+        Annotation::ExprVP(
+            NeExpr<ModuleID::ID>::create(
+                // `RegisterExpr` marks the value of the current expression, i.e., the loaded value.
+                // The id is zero, since this annotation only uses one input variable.
```
RalfJung:
```suggestion
                // The `id` is ignored by GenMC; it is only used by the LLI frontend to substitute
                // other variables from previous expressions that may be used here.
```

src/concurrency/genmc/intercept.rs
Patrick-6: I don't think so, since a Mutex may be just an `AtomicU8`, `*U32` or even `*U128` depending on the implementation (possibly more if there are any extra fields in the Mutex), so the size might no exactly equal the maximum size (u128 would currently even cause a crash).

For example, on Linux, the size is only 4 bytes (see [here](https://github.com/rust-lang/rust/blob/1a3cdd34629306fa67624eaa60d73687e7fcf855/library/std/src/sys/sync/mutex/futex.rs#L4-L8) and [here](https://github.com/rust-lang/rust/blob/1a3cdd34629306fa67624eaa60d73687e7fcf855/library/std/src/sys/pal/hermit/futex.rs#L11-L14))
On itron, [the mutex](https://github.com/rust-lang/rust/blob/1a3cdd34629306fa67624eaa60d73687e7fcf855/library/std/src/sys/sync/mutex/itron.rs#L9-L12) is `sizeof(usize) + sizeof(bool)` large ([from the `id` and `spin` fields](https://github.com/rust-lang/rust/blob/1a3cdd34629306fa67624eaa60d73687e7fcf855/library/std/src/sys/pal/itron/spin.rs#L52-L56)), which currently crash Miri on a 64-bit system.
(Cases like this is also why I had a clamp on the size initially.)

RalfJung: What are you talking about...?

You are currently forbidding 8-byte mutexes. Why?

Patrick-6: oh, I misread, I thought you meant to change it to `size == MAX_ACCESS_SIZE`.

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
```diff
@@ -80,7 +80,7 @@ auto MiriGenmcShim::handle_mutex_lock(ThreadId thread_id, uint64_t address, uint
         // We did not acquire the mutex, so we tell GenMC to block the thread until we can acquire
         // it. GenMC determines this based on the annotation we pass with the load further up in
         // this function, namely when that load will read a value other than `MUTEX_LOCKED`.
-        GenMCDriver::handleAssume(inc_pos(thread_id), AssumeType::Spinloop);
+        // NOTE: The blocking happens on the Miri side.
```
RalfJung: I am confused, why did you move this around?

src/concurrency/genmc/shims.rs
```diff
@@ -47,13 +96,15 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
         if !result.is_reset {
-            throw_machine_stop!(TerminationInfo::Deadlock);
+            // GenMC expects us to block the thread if we did get a `reset`.
+            genmc_ctx.handle_assume_block(&this.machine, AssumeType::Spinloop)?;
         }
-        // NOTE: We don't write anything back to Miri's memory, the Mutex state is handled only by GenMC.
+        // Block the thread on the Miri side.
         this.block_thread(
                 crate::BlockReason::Genmc,
                 None,
```
RalfJung, line 100: This still duplicates the `handle_mutex_lock` call.

Based on my understanding of what happens in LLI, this should look roughly as follows:
```rust
fn intercept_mutex_lock(...) {
  let result = genmc_ctx.handle_mutex_lock();
  if result.reset {
    // GenMC will schedule us again later when the time is right for acquiring this lock.
    // Then we try again.
    this.block_thread(
      crate::BlockReason::Genmc, None,
      callback!( @capture<'tcx> { }, |...| { ecx.intercept_mutex_lock(...) } ),
    );
  } else if result.is_lock_acquired {
    // All good!
  } else {
    // We won't get this lock in this execution, just permanently block this thread.
    this.block_thread(
      crate::BlockReason::Genmc, None,
      callback!( @capture<'tcx> { }, |...| { unreachable!() } ),
    );
  }
}
```
I don't know when LLI emits which kind of `handleAssume` around locks; please figure that out.

You said earlier this needs a loop in the scheduler since it can now happen that the thread genmc scheduled immediately blocks again within its unblock handler -- that makes sense.

RalfJung: I tried finding the LLI code for that; I think it's [here](https://gitlab.inf.ethz.ch/public-plf/genmc/-/blob/e175ca8c0f740c1ed382d08ba44c0cdef5606d1d/src/Runtime/Execution.cpp#L2923)?

So it looks like on a reset, we... do nothing? Well we still have to block the Miri thread (so that we can run the lock logic again in the unblock handler) but we don't emit any GenMC events.

If we actually get a result and it's that we did not acquire the lock, we call handleAssume with Spinloop. Seems fine to do that already on the C++ side like you did before. On the Miri side we behave like `assume(false)`, i.e. we block and panic if we ever get unblocked.

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
Patrick-6: I think it is better to have all blocking to be initiated from the Rust side, not split across both Rust and C++ code. GenMC has no assertions about not blocking a thread twice (it just segfaults), so I've added an assertion for that [here](https://github.com/Patrick-6/miri/blob/093381e052d30248981487b1d278409257666702/genmc-sys/cpp/src/MiriInterface/EventHandling.cpp#L36) and moved all blocking calls to the Rust side through that function.

src/concurrency/genmc/shims.rs
Patrick-6: The `CALL_DRIVER_RESET_IF_FALSE` macro does something: it [sets back the current instruction in LLI by one](https://gitlab.inf.ethz.ch/public-plf/genmc/-/blob/ce775ccd7866db820fa12ffca66463087a11dd96/src/Runtime/Interpreter.h#L328) if the [`HandleResult<SVal>`](https://gitlab.inf.ethz.ch/public-plf/genmc/-/blob/ce775ccd7866db820fa12ffca66463087a11dd96/src/Verification/GenMCDriver.hpp#L105-109) returned by `handleLoad` does not contain a value.

Patrick-6: Side note: We currently never handle the `Invalid` case of `HandleResult` for any `handleLoad` or other `handle*` function. `Invalid` seems to be returned in "moot" executions only, which none of our current tests encounter. These will definitely show up once Miri supports bounded exploration with GenMC, but we can probably ignore them for now (Miri will crash with an error if a `HandleResult::Invalid` is ever returned).
EDIT: [This is one such place where it could error out on `Invalid`](https://github.com/Patrick-6/miri/blob/093381e052d30248981487b1d278409257666702/genmc-sys/cpp/src/MiriInterface/EventHandling.cpp#L63-L64).

RalfJung: > The `CALL_DRIVER_RESET_IF_FALSE` macro does something: it [sets back the current instruction in LLI by one](https://gitlab.inf.ethz.ch/public-plf/genmc/-/blob/ce775ccd7866db820fa12ffca66463087a11dd96/src/Runtime/Interpreter.h#L328) if the [`HandleResult<SVal>`](https://gitlab.inf.ethz.ch/public-plf/genmc/-/blob/ce775ccd7866db820fa12ffca66463087a11dd96/src/Verification/GenMCDriver.hpp#L105-109) returned by `handleLoad` does not contain a value.

We achieve the same by calling `intercept_mutex_lock` from the unblock callback.

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
RalfJung: It just seems a bit odd that a handle_mutex_lock would not actually handle that call entirely -- I don't think it is good to have that logic split across the Rust and C++ code like you have now.

genmc-sys/cpp/src/MiriInterface/Mutex.cpp
```diff
@@ -0,0 +1,159 @@
...
+        const bool* is_coherence_order_maximal_write = std::get_if<bool>(&store_ret);
+        ERROR_ON(
+            nullptr == is_coherence_order_maximal_write,
+            "Unimplemented: store part of mutex try_lock returned unexpected result."
+        );
+    } else {
+        // We did not acquire the mutex, so we tell GenMC to block the thread until we can acquire
+        // it. GenMC determines this based on the annotation we pass with the load further up in
+        // this function, namely when that load will read a value other than `MUTEX_LOCKED`.
+        GenMCDriver::handleAssume(inc_pos(thread_id), AssumeType::Spinloop);
```
RalfJung: You could call `handle_assume_block` here to still get the assertion, couldn't you?

src/concurrency/genmc/intercept.rs
```diff
@@ -35,4 +71,164 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        );
+        if let Some(error) = result.error.as_ref() {
+            // FIXME(genmc): improve error handling.
+            throw_ub_format!("{}", error.to_string_lossy());
+        }
+        debug!(
+            "GenMC: Mutex::try_lock(): is_reset: {}, is_lock_acquired: {}",
+            result.is_reset, result.is_lock_acquired
+        );
+        assert!(!result.is_reset, "GenMC returned 'reset' for a mutex lock multiple times.");
```
RalfJung:
```suggestion
        assert!(!result.is_reset, "GenMC returned 'reset' for a mutex try_lock.");
```

src/concurrency/genmc/mod.rs
```diff
@@ -200,6 +200,14 @@ impl GenmcCtx {
     fn get_alloc_data_races(&self) -> bool {
         self.exec_state.allow_data_races.get()
     }
+
+    /// Get the id of the currently active thread mapped to it's thread ID for GenMC.
```
```suggestion
    /// Get the GenMC id of the currently active thread.
```

tests/genmc/pass/intercept/mutex_poison.rs
RalfJung: Based on what did you resolve this discussion?

Patrick-6: I assumed this was about the `mutex_poison.rs` test failing in CI (fixed by the rebase), since that coincided with your comment.

> Poisoning is entirely implemented inside the library on top of the basic lock we are intercepting.

Is it correct to assume that this will stay true with all future changes to the standard library? [`poison::Flag`](https://github.com/rust-lang/rust/blob/d773bd07d63a74adcf25ea5f4aae986be94cac5e/library/std/src/sync/poison.rs#L91) _is_ currently target independent, but I did not find this documented as an invariant.

> So how could this possibly _not_ work if the lock itself works correctly...?

Are we already confident that the lock itself works correctly based on just the other mutex tests?

In this test, we check that the interception also works for unlocks caused by a `panic` (correctly poisoning the mutex), and that we can still lock the Mutex afterwards (after catching the panic and acquiring the mutex from the main thread).

RalfJung: > Is it correct to assume that this will stay true with all future changes to the standard library? [poison::Flag](https://github.com/rust-lang/rust/blob/d773bd07d63a74adcf25ea5f4aae986be94cac5e/library/std/src/sync/poison.rs#L91) is currently target independent, but I did not find this documented as an invariant.

I think that is a reasonable assumption, given that (a) I cannot imagine any way in which platform-specific poison handling would help, and (b) nopoison mutexes are being worked on and are surely meant to share most of their code with the poisoning mutexes.

> Are we already confident that the lock itself works correctly based on just the other mutex tests?

If not, we should extend those tests. Which case is not covered?

> In this test, we check that the interception also works for unlocks caused by a panic (correctly poisoning the mutex), and that we can still lock the Mutex afterwards (after catching the panic and acquiring the mutex from the main thread).

The API we are shiming doesn't even mention unwinding. It has 3 functions with a fairly clear API contract, and that is what we should focus on testing.

Panic unwinding is just a normal form of control flow. If we tested every shim both during "normal" execution and unwinding we'd double the size of our test suite for no reason at all. We're also not testing separately whether mutexes work both inside and outside of loops, for instances.

src/concurrency/thread.rs
```diff
@@ -708,21 +708,31 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
...
+                    info!(
+                        "GenMC: scheduling blocked thread {next_thread_id:?}, so we unblock it now."
+                    );
+                    this.unblock_thread(next_thread_id, BlockReason::Genmc)?;
+                }
+                // Set the new active thread.
+                let thread_manager = &mut this.machine.threads;
+                thread_manager.active_thread = next_thread_id;
+                // A thread blocked by GenMC may get blocked again during the unblocking callback.
+                // In that case, we need to ask for a different thread to run next.
```
```suggestion
                // The thread we just unblocked may have been blocked again during the unblocking callback.
                // In that case, we need to ask for a different thread to run next.
```

src/concurrency/thread.rs
```diff
@@ -708,21 +708,31 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
...
+                    .is_blocked_on(BlockReason::Genmc)
+                {
+                    info!(
+                        "GenMC: scheduling blocked thread {next_thread_id:?}, so we unblock it now."
+                    );
+                    this.unblock_thread(next_thread_id, BlockReason::Genmc)?;
+                }
+                // Set the new active thread.
+                let thread_manager = &mut this.machine.threads;
+                thread_manager.active_thread = next_thread_id;
```
RalfJung: Seems better to only do this assignment after we decided that we'll really run this thread (i.e. inside the `if ... is_enabled`).

tests/genmc/pass/intercept/mutex_poison.rs
Patrick-6: > > Are we already confident that the lock itself works correctly based on just the other mutex tests?
> 
> If not, we should extend those tests. Which case is not covered?

Without anything related to unwinding, I can only see some missing cases, but these are likely not possible to trigger with `std::sync::Mutex` (since we don't use `pthread_mutex_*` directly like GenMC does for C code):
- attempting to lock an uninitialized / destroyed mutex
  - ([GenMC test `unalloc-lock`](https://github.com/Patrick-6/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/tests/wrong/infr/unalloc-lock/variants/unalloc-lock0.c))
  - ([GenMC test `destroyed-mutex`](https://github.com/Patrick-6/genmc/blob/afabd75b92f9b58b83465f009ccf1ce8820571ef/tests/wrong/locking/destroyed-mutex/variants/destroyed-mutex0.c#L1))
- double unlocking a mutex (prevented by automatic unlocking in `MutexGuard::drop`)
- unlocking a mutex on a different thread (prevented by `MutexGuard` not being `sync`).

It is probably fine to just drop the `mutex_poison` test.
Should there be tests for any of the cases above?

RalfJung: You can test most of these things with sufficiently terrible unsafe code. ;)

- We're not intercepting init/destroy since those are not atomic operations so this one is moot.
- You can double unlock with something like
```rust
let mut guard = mutex.lock().unwrap();
ptr::drop_in_place(&raw mut guard);
drop(guard);
```
- You can unlock on a different thread using an unsound send wrapper (you said `Sync` but I don't think that is right)
```rust
struct MakeSend<T>(T):
unsafe impl<T> Send for MakeSend<T> {}
```

tests/genmc/pass/shims/mutex_simple.rs
RalfJung: (This is still open)

tests/genmc/pass/intercept/mutex_poison.rs
Patrick-6: Should there be tests for unlocking on a different thread?
[The docs for `MutexGuard`](https://doc.rust-lang.org/std/sync/struct.MutexGuard.html#impl-Send-for-MutexGuard%3C'_,+T%3E) seem to imply that there are (non-POSIX) platforms that do allow unlocking on a different thread, so testing for that on all platforms would be wrong.

---------------------------------------------------------------

For double unlocking, this program:
```rust
use std::ptr;
use std::sync::Mutex;

fn main() {
    unsafe {
        let mutex = Mutex::new(1234u64);
        let mut guard = mutex.lock().unwrap();
        ptr::drop_in_place(&raw mut guard);
        drop(guard);
    }
}
```
- For regular Miri, this gives _no errors_! Is this expected behavior? ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d921f1156089aa6de2dc9615e57ee457))
- With GenMC, it causes an ICE in [`schedule_thread`](https://github.com/Patrick-6/miri/blob/b38d2cd4275f1684b5b741ceb80a1aefcde48e9e/src/concurrency/genmc/scheduling.rs#L130-L137) due to an unhandled `Mixed-size accesses` error, I'll try to investigate that.

RalfJung: > Should there be tests for unlocking on a different thread?
[The docs for MutexGuard](https://doc.rust-lang.org/std/sync/struct.MutexGuard.html#impl-Send-for-MutexGuard%3C'_,+T%3E) seem to imply that there are (non-POSIX) platforms that do allow unlocking on a different thread, so testing for that on all platforms would be wrong.

Seems good to test that Miri+GenMC doesn't crash or so when that happens. But it's fine for that test to just pass.

> For regular Miri, this gives no errors! Is this expected behavior?

Yeah that's fine -- this is library UB but not language UB. The underlying futex implementation is entirely fine with this. (The second unlock might unlock the mutex held by someone else, which can then cause language UB that Miri would detect.)

Patrick-6: > * With GenMC, it causes an ICE in [`schedule_thread`](https://github.com/Patrick-6/miri/blob/b38d2cd4275f1684b5b741ceb80a1aefcde48e9e/src/concurrency/genmc/scheduling.rs#L130-L137) due to an unhandled `Mixed-size accesses` error, I'll try to investigate that.

Found the issue for the ICE: GenMC does not return an error from `handleLoad` for mixed-size accesses due to a missing change on the GenMC side ([private repo link](https://gitlab.mpi-sws.org/sav/genmc-tool/-/merge_requests/166/diffs?commit_id=46a4564d9797319305e900293c291a7d9fd1da4a)).

Looks like Mutexes on the stack or heap are currently not supported, since they cause mixed size accesses (initial write causes accesses with `sizeof<Mutex>`, but the locking load/stores are `sizeof<u32>` (on Linux)). Mutexes in static allocations don't have this initializing write (they use the temporal mixing API), so they work.

RalfJung: I am confused -- all our accesses go through the mutex shims and they all use the same size, don't they?

src/concurrency/genmc/shims.rs
RalfJung, line 1: We call what we do here "shims" in the rest of Miri, so it'd probably make sense to rename this file and the test folder to `shims` as well.

tests/genmc/pass/intercept/mutex_poison.rs
Patrick-6: Not all accesses: for a stack/heap allocated Mutex, we have a `copy_nonoverlapping` (I think for returning the Mutex from `Mutex::<T>::new()`), which is at least one load + store to the address of the Mutex with size `(4 + 1 + padding + sizeof(T)) bytes` (for the Futex, the poison flag, and the `UnsafeCell<T>`).
The lock call then accesses the first 4 bytes of the Mutex, which causes a mixed size access error (same address as copy, but different size).

For the static allocation, there is no `copy_nonoverlapping` (the initialization is handled somewhere in the consteval code in rustc IIUC). On the first (try_)lock call, we pass the "initial" value `MUTEX_UNLOCKED` to GenMC, which updates the value of the first 4 bytes of the Mutex without causing a mixed-size access error.

RalfJung: Ah so this is just the normal mixed-size issue then, paired with the fact that a mutex is 4 bytes and we only properly support ptr-sized atomics at the moment.

Maybe just add this to the pass test commented-out with a FIXME saying this should pass once mixed-size stuff works properly (it passes because we don't model POSIX mutexes here, but a more low-level mutex API)?

Patrick-6: I've added a comment to the `mutex_simple.rs` test about adding checks using stack/heap allocated mutexes.

> (it passes because we don't model POSIX mutexes here, but a more low-level mutex API)?

The test for unlocking on a different thread (using a static Mutex) will give an error on _all targets_, since GenMC always assumes `pthread_mutex`, so unlocking on a different thread is not allowed. Both this and double-unlock cause a GenMC [`VerificationError::VE_InvalidUnlock`](https://gitlab.inf.ethz.ch/public-plf/genmc/-/blob/e175ca8c0f740c1ed382d08ba44c0cdef5606d1d/src/Verification/VerificationError.cpp#L65), see the 2 new tests I've just pushed)

tests/genmc/fail/shims/mutex_double_unlock.rs
```diff
@@ -0,0 +1,20 @@
+//@compile-flags: -Zmiri-genmc -Zmiri-disable-stacked-borrows
+//@error-in-other-file: Undefined Behavior
+
+// Test that GenMC can detect a double unlock of a mutex.
```
RalfJung, line 4:
```suggestion
// Test that GenMC can detect a double unlock of a mutex.
// This test will cause an error even if the program actually would work entirely fine despite the double-unlock
// because GenMC always assumes a `pthread`-like API.
```

RalfJung:

@rustbot author

Issue #4503, opened 2025-07-28
Support libc memset


bjorn3:
This is very common in C code. Supporting it in miri would make it easier to detect UB in C code translated to rust using c2rust.

RalfJung:
I'm surprised we don't already have this.^^ I guess Rust code never uses it since we have `write_bytes`.

PR #4618, opened 2025-10-07
feat: add support for libc::memset

vishruth-thimmaiah:
closes: rust-lang/miri#4503

### DIFF
--- a/src/shims/foreign_items.rs
+++ b/src/shims/foreign_items.rs
@@ -837,6 +837,7 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 #[expect(clippy::as_conversions)]
                 let val = val as u8;
 
+                // C requires that this must always be a valid pointer, even if `n` is zero, so we better check that.
                 this.ptr_get_alloc_id(ptr_dest, 0)?;
 
                 let bytes = std::iter::repeat_n(val, n.try_into().unwrap());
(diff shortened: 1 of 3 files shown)


tests/pass-dep/libc/libc-mem.rs
RalfJung, line 1: We'll also need a test in `fail-dep/libc/memset_null` doing a zero-sized memset on a null pointer -- which likely is UB for the C function, even though we allow it for the intrinsic.

src/shims/foreign_items.rs
```diff
@@ -827,6 +827,22 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let [ptr_dest, val, n] =
+                    this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
+                let ptr_dest = this.read_pointer(ptr_dest)?;
+                let val = this.read_scalar(val)?.to_i32()?;
+                let n = this.read_target_usize(n)?;
+                // The docs say val is "interpreted as unsigned char".
+                #[expect(clippy::as_conversions)]
+                let val = val as u8;
+
+                this.ptr_get_alloc_id(ptr_dest, 0)?;
```
RalfJung, line 841:
```suggestion
                // C requires that this must always be a valid pointer, even if `n` is zero, so we better check that.
                this.ptr_get_alloc_id(ptr_dest, 0)?;
```

RalfJung (commented):
Thanks for the PR!

RalfJung:
Looks good, thanks :)

PR #4622, opened 2025-10-11
native-lib/trace: Intercept mmap and munmap

nia-e:
Adds support for tracing `mmap`s in native code and logging that. Notably does *not* let us call `libc::mmap(..)` on the Rust side and then `munmap` it on the native side / the other way; I have ideas for maybe doing that, but they'd be quite a bit more involved since it requires us doing some syscall shimming.

### DIFF
--- a/src/alloc/alloc_bytes.rs
+++ b/src/alloc/alloc_bytes.rs
@@ -12,8 +12,12 @@ use crate::helpers::ToU64 as _;
 
 #[derive(Clone, Debug)]
 pub enum MiriAllocParams {
+    /// Allocation was created by calling `alloc::alloc()`.
     Global,
+    /// Allocation came from the isolated allocator.
     Isolated(Rc<RefCell<IsolatedAlloc>>),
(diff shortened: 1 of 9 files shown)


RalfJung:
I saw the PR and it's in my queue, but it's a big one so it may take a bit until I have time to dig into this -- sorry.

nia-e:
Realised this actually has a couple issues that need to be addressed and I *could* do that in this implementation relatively simply, or rewrite it to be better in every way (so we can call `mmap` on one side of the ffi boundary and `munmap` on the other w/ the same pointer, and vice versa) with a tiny tiny bit of inline assembly. I think I'll do that and mark this as a draft

oli-obk:
Nia gave me a crash course in various of those ideas at eurorust. I'll do the first reviews and assign to you (Ralf) once I think it's ready

RalfJung:
@nia-e I am going to close this PR since it's become inactive and work has moved to smaller follow-up PRs. Maybe it'd make sense to open an issue summarizing the plan and the current status?

Issue #4625, opened 2025-10-13
native-lib mode does not support null function pointers passed as `Option<fn()>`


weiznich:
I believe that this is more a feature request than anything else. I've tried again running some of diesels tests via miri and with the `Zmiri-native-lib` support I think it should be finally nearly possible now. 

The remaining roadblock seems to be errors like the following:

```
error: unsupported operation: passing a non-#[repr(C)] struct over FFI: std::option::Option<unsafe extern "C" fn(*mut std::ffi::c_void, i32, *mut *mut i8, *mut *mut i8) -> i32>
   --> /home/weiznich/Documents/rust/diesel/diesel/src/sqlite/connection/raw.rs:75:13
    |
 75 | /             ffi::sqlite3_exec(
 76 | |                 self.internal_connection.as_ptr(),
 77 | |                 query.as_ptr(),
 78 | |                 callback_fn,
 79 | |                 callback_arg,
 80 | |                 ptr::null_mut(),
 81 | |             )
    | |_____________^ unsupported operation occurred here
```

Relevant code link: https://github.com/diesel-rs/diesel/blob/ed01866ec0bf63070d179da4ce3d37df582f3e79/diesel/src/sqlite/connection/raw.rs#L71-L82

The [`sqlite3_exec`](https://docs.rs/libsqlite3-sys/latest/libsqlite3_sys/fn.sqlite3_exec.html) function is provided by libsqlite3-sys as a generated binding for the underlying [`sqlite3_exec`](https://sqlite.org/c3ref/exec.html) function from sqlite itself. These bindings are generated by bindgen, so my expectation here would be that miri is able to deal with bindings generated by bindgen or that bindgen generates bindings that are accepted by miri. Right of now there seems to be some sort of disagreement of what's accepted and what not. 

Technically speaking I believe the code generated by bindgen is actually sound as the documentation of [`Option`](https://doc.rust-lang.org/std/option/index.html#representation) explicitly calls out that specifically `Option<unsafe extern "C" fn (…)>::None` is equivalent to a null function pointer.

So this is more or less the request to support "non `#[repr(C)]`" rust types with an explicitly specified representation in FFI code.

RalfJung:
Yeah, Miri's native-lib mode does not yet support all types that are actually ABI-safe.

RalfJung:
[...]

PR #4628, opened 2025-10-13
native-lib: support all types with Scalar layout

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/4625

However, this also means we'll silently just pass function pointers to C and things will explode in horrendous ways if C then ends up calling such a function pointer. So I don't think we can land this as-is. I am not sure what else to do... we could have a flag for people to opt-in to this and promise the function pointers will never be called?

### DIFF
--- a/src/shims/native_lib/trace/parent.rs
+++ b/src/shims/native_lib/trace/parent.rs
@@ -530,7 +530,18 @@ fn handle_segfault(
     // Don't use wait_for_signal here since 1 instruction doesn't give room
     // for any uncertainty + we don't want it `cont()`ing randomly by accident
     // Also, don't let it continue with unprotected memory if something errors!
-    let _ = wait::waitid(wait::Id::Pid(pid), WAIT_FLAGS).map_err(|_| ExecEnd(None))?;
+    let stat = wait::waitid(wait::Id::Pid(pid), WAIT_FLAGS).map_err(|_| ExecEnd(None))?;
+    match stat {
+        wait::WaitStatus::Signaled(_, s, _)
+        | wait::WaitStatus::Stopped(_, s)
+        | wait::WaitStatus::PtraceEvent(_, s, _) =>
+            assert!(
+                !matches!(s, signal::SIGSEGV),
+                "native code segfaulted when re-trying memory access\n\
+                is the native code trying to call a Rust function?"
+            ),
+        _ => (),
+    }
 
     // Zero out again to be safe
     for a in (ch_stack..ch_stack.strict_add(CALLBACK_STACK_SIZE)).step_by(ARCH_WORD_SIZE) {


nia-e:
Could we detect if a pointer is a function pointer? If so, we can change its address (if non-null) to point to some preallocated single page of memory mapped R-X which raises a signal that the supervisor can intercept & print an error

RalfJung:
We can't just change the value when it is being passed to C -- the program could realize this with some ptr2int casts. (Also, for fields in structs/unions behind ptr indirections, determining their type becomes more of a heuristic.)

However, we could adjust the logic that determines the addresses of function pointer allocations to begin with, to put them all onto some R-X page. Interesting idea.

src/shims/native_lib/mod.rs
```diff
@@ -386,34 +386,44 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        // stable, but that's fine -- we anyway can't check that the callee has the intended ABI.
+        if let BackendRepr::Scalar(s) = layout.backend_repr {
+            // Simple sanity-check: this cannot be `repr(C)`.
+            assert!(!layout.ty.ty_adt_def().is_some_and(|adt| adt.repr().c()));
+            return interp_ok(match s.primitive() {
+                Primitive::Int(Integer::I8, /* signed */ true) => FfiType::i8(),
+                Primitive::Int(Integer::I16, /* signed */ true) => FfiType::i16(),
+                Primitive::Int(Integer::I32, /* signed */ true) => FfiType::i32(),
+                Primitive::Int(Integer::I64, /* signed */ true) => FfiType::i64(),
+                Primitive::Int(Integer::I8, /* signed */ false) => FfiType::u8(),
```
nia-e, line 411: Should this (+ those below) be `Primitive::Int(Integer::U8)`? I'm not super sure about the backend repr stuff so maybe this is correct because of that `false` idk

RalfJung: There is no `Integer::U8`: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_abi/enum.Integer.html.

The `rustc_abi::Integer` type only encodes the size of an int, not its sign.

oli-obk:
Since we need to convert values into their FFI repr, could we reject all those with function provenance and error?

I guess a fully ptr2int converted function pointer would still get missed.

RalfJung:
Our conversion function isn't type-driven -- it just takes the raw bytes of the argument values and sends them over.

We could scan that value *and all memory reachable from it* for function pointers, but currently we don't do such a recursive traversal so... not sure how I feel about that. If we can pull of the R-X page, that feels more slick (and it should reliably segfault even if we don't have the supervisor that can give a nice error).

EDIT: Actually, we do have the `visit_reachable_allocs` scan. That would then error even for function pointers that aren't actually used. Hm...

nia-e:
Do we care about function pointers having different addresses? If not then we can do this even more simply, just have

```rust
pub extern "C" fn one_pointee_to_rule_them_all() {
    panic!("unsupported operation: native code tried to call a Rust function");
}
```

where all function pointers take its address.

RalfJung:
yeah every function pointer needs to have a unique address for Miri itself to be able to handle them properly.

nia-e:
Then we can have that function delcared somewhere, allocate as many R-X pages as needed, fill them with `jmp one_pointee_to_rule_them_all` calls, and get `page_size / sizeof(jmp)` distinct addresses per page :D

RalfJung:
That still needs platform-specific code to generate the assembly. So we still need fallback code that creates an R-- page where just jumping there segfaults (which I assume can be done with mmap).

---

Anyway turns out Oli's suggestion is trivial to implement since we already have the `visit_reachable_allocs` scan; I pushed that.  However, this does mean that if any function pointer gets cast to an integer ever, one can't make any more FFI calls... so maybe this shouldn't stop execution, and should just emit a warning saying "if you ever call that function, glhf"?

I assume that `malloc`d memory is non-executable everywhere these days, so calling such a function pointer *should* reliably crash and not execute garbage instructions.

nia-e:
Fair, Oli's is probably the reasonable approach. We could do the other you suggested; there are "weird" architectures where [`R` implies `X`](https://www.man7.org/linux/man-pages/man2/mmap.2.html#VERSIONS) (iirc m68k has this?) but if you're somehow running Miri on that it's likely fair to say that's a you problem

RalfJung:
Okay let's do that. :D

Interestingly, when I do actually call that function, the program just hangs. I think there's a segfault and then the supervisor is getting confused since the segfault is indeed in the address range it cares about...

nia-e:
I assume this is the issue, that we're not checking that the segfault is cleared?
```rust
// in handle_segfault
-    let _ = wait::waitid(wait::Id::Pid(pid), WAIT_FLAGS).map_err(|_| ExecEnd(None))?;
+    let stat = wait::waitid(wait::Id::Pid(pid), WAIT_FLAGS).map_err(|_| ExecEnd(None))?;
+    match stat {
+        wait::WaitStatus::Signaled(_, s, _) | wait::WaitStatus::Stopped(_, s) | wait::WaitStatus::PtraceEvent(_, s, _) => assert!(!matches!(s, signal::SIGSEGV), "Native code segfaulted in correctly allocated memory; this is probably a function call!"),
+        _ => (),
+    }
```

RalfJung:
That works well, thanks :)

Issue #4633, opened 2025-10-15
Bump libffi dependency


RalfJung:
There's a new major version of libffi. We should update so that we can benefit from https://github.com/libffi-rs/libffi-rs/issues/180. However apparently the APIs Miri uses changed so this will not be a trivial update.

@nia-e any chance you could take a look at this?

nia-e:
on it!

PR #4634, opened 2025-10-15
native-lib: bump libffi

nia-e:
Closes rust-lang/miri#4633 

The actual breaking change in libffi 5.0 seems to be the lifetime on the `libffi::middle::Arg` type, which we were already more-or-less respecting the semantics of. This lets us simplify the code that handles it so hooray!


RalfJung:
That was very quick, thanks :)

PR #4643, opened 2025-10-22
Avoid panicking when `Cargo.toml` is not present

FranciscoTGouveia:
Hello!
This is my first contribution here to Miri, and hopefully the first of many! :)

When running `cargo clippy` in a directory that does not contain a `Cargo.toml`, it displays an informative message such as:
`could not find Cargo.toml in [dir] or any parent directory`.

In contrast, running `cargo miri [test|run]` without a `Cargo.toml` currently causes a panic.
While it is expected that this command should fail, a more user-friendly behavior, similar to `clippy`, would be to print a descriptive error message instead of panicking.

I'd appreciate your feedback on the following:
I was considering simply calling `bail!` on the error, as done in `cargo` (see [here](https://github.com/rust-lang/cargo/blob/344c4567c634a25837e3c3476aac08af84cf9203/src/cargo/util/important_paths.rs#L21-L33)), but that would require adding `anyhow` as a dependency to `cargo-miri`. Since we already have a `show_error!` macro available, I believe it might be better to use that instead.

Thank you in advance for your time and input!

### DIFF
--- a/cargo-miri/src/util.rs
+++ b/cargo-miri/src/util.rs
@@ -213,7 +213,11 @@ fn cargo_extra_flags() -> Vec<String> {
 
 pub fn get_cargo_metadata() -> Metadata {
     // This will honor the `CARGO` env var the same way our `cargo()` does.
-    MetadataCommand::new().no_deps().other_options(cargo_extra_flags()).exec().unwrap()
+    MetadataCommand::new()
+        .no_deps()
+        .other_options(cargo_extra_flags())
+        .exec()
+        .unwrap_or_else(|err| show_error!("{}", err))
 }
 
 /// Pulls all the crates in this workspace from the cargo metadata.


cargo-miri/src/util.rs
```diff
@@ -213,7 +213,11 @@ fn cargo_extra_flags() -> Vec<String> {
 
 pub fn get_cargo_metadata() -> Metadata {
     // This will honor the `CARGO` env var the same way our `cargo()` does.
-    MetadataCommand::new().no_deps().other_options(cargo_extra_flags()).exec().unwrap()
+    MetadataCommand::new()
+        .no_deps()
+        .other_options(cargo_extra_flags())
+        .exec()
+        .unwrap_or_else(|err| show_error!("{}", err))
```
RalfJung, line 220: What does this error look like?

I am wondering if we should add something like "`cargo metadata` failed:\n{}" or so.

RalfJung (commented):
Thanks for the PR!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4643/ce6f40d9c4f529868435c7d410ca83fb244ecb39..60776a43b02ecd26638094d722310fa47dc8fc14)*

cargo-miri/src/util.rs
FranciscoTGouveia: The error message currently looks like this:
```
fatal error: `cargo metadata` exited with an error: error: could not find `Cargo.toml` in `/home/ftg` or any parent directory
```

For a visual comparison, an image of the previous and current behavior is available below:

image

Apologies for not including this information when I initially opened the PR.

RalfJung: Thanks! And no worries. :)

That does sound a bit redundant, but the message seems to be entirely controlled by cargo_metadata, and it's definitely a big step up from before, so -- LGTM.

Issue #1705, opened 2021-02-10
Can't find crate with `hyper` crate as a dependency


saruman9:
## MCVE

`Cargo.toml`:

```toml
[dependencies]
hyper = "0.14.4"
```

## Error

```
...
Compiling hyper v0.14.4
error[E0463]: can't find crate for `http`
  --> /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.4/src/lib.rs:61:9
   |
61 | pub use http;
   |         ^^^^ can't find crate

error: aborting due to previous error
```

RalfJung:
Thanks for the report! I can reproduce this locally.

This has something to do with the tricks we pull to do a check-only-build (even when cargo instructs us to do a full build). When I [disable those hacks](https://github.com/RalfJung/miri/tree/miri-full-build) and set `XARGO_CHECK=xargo`, the issue disappears.

hyd-dev:
[Copy and paste](https://github.com/rust-lang/miri/commit/12dcbaddcb21119d066a34d84f0a49a85432ddd5) the [`--extern` flag patching from `phase_cargo_runner`](https://github.com/rust-lang/miri/blob/e09dce0f43993faf03cc46ac365b6a46287f90e7/cargo-miri/bin.rs#L749-L750) fixes it.

Looks like `hyper` is [`crate-type = ["lib", "staticlib", "cdylib"]`](https://github.com/hyperium/hyper/blob/196d9bd69c0059dbff6771f6dce9188d4841c185/Cargo.toml#L23), so Cargo [decided to pass `--extern http=*.rlib` instead of `---extern http=*.rmeta`](https://github.com/rust-lang/cargo/blob/b842849732f89df8675eb2d933c384d6338e4466/src/cargo/core/compiler/mod.rs#L1091-L1105) to the compiler, but the `.rlib` file [is only a stub](https://github.com/rust-lang/miri/blob/e09dce0f43993faf03cc46ac365b6a46287f90e7/cargo-miri/bin.rs#L690-L697).

When compiling a normal library crate (`--extern http=/my/home/directory/test/target/x86_64-unknown-linux-gnu/debug/deps/libhttp-8f0bd6ee31f077df.rmeta`):
```
   Compiling http-body v0.4.0
[...]

PR #1710, opened 2021-02-14
 Patch `--extern` arguments in `phase_cargo_rustc` as well

hyd-dev:
Fixes #1705.

PR #4648, opened 2025-10-23
Use the dummy codegen backend

bjorn3:
This allows getting rid of all the argument mangling hacks as well as the exported_non_generic_symbols override hack by doing cargo build instead of cargo check. This is also a prerequisite for using native jit mode support in miri that I hope to add to rustc in the future to be used with cg_clif too.


src/bin/miri.rs
```diff
@@ -250,71 +246,6 @@ struct MiriBeRustCompilerCalls {
...
-    #[allow(rustc::potential_query_instability)] // rustc_codegen_ssa (where this code is copied from) also allows this lint
-    fn config(&mut self, config: &mut Config) {
-        if config.opts.prints.is_empty() && self.target_crate {
-            // Queries overridden here affect the data stored in `rmeta` files of dependencies,
-            // which will be used later in non-`MIRI_BE_RUSTC` mode.
-            config.override_queries = Some(|_, local_providers| {
-                // `exported_non_generic_symbols` and `reachable_non_generics` provided by rustc always returns
-                // an empty result if `tcx.sess.opts.output_types.should_codegen()` is false.
-                // In addition we need to add #[used] symbols to exported_symbols for `lookup_link_section`.
-                local_providers.exported_non_generic_symbols = |tcx, LocalCrate| {
```
bjorn3, line 262: On second thought we may still need this for `#[used]`.

cargo-miri/src/phases.rs
```diff
@@ -472,29 +471,6 @@ pub fn phase_rustc(mut args: impl Iterator<Item = String>, phase: RustcPhase) {
...
-                        val.push("metadata");
-                    }
-                }
-                cmd.arg(format!("{emit_flag}={}", val.join(",")));
-                continue;
-            }
-            // Patch `--extern` filenames, since Cargo sometimes passes stub `.rlib` files:
-            // https://github.com/rust-lang/miri/issues/1705
-            if arg == "--extern" {
-                forward_patched_extern_arg(&mut args, &mut cmd);
```
bjorn3, line 495: The `forward_patched_extern_arg` function can likely be removed as the fake lib file hack has been removed.

src/lib.rs
```diff
@@ -165,6 +165,7 @@ pub use crate::shims::unwind::{CatchUnwindData, EvalContextExt as _};
 /// Also disable the MIR pass that inserts an alignment check on every pointer dereference. Miri
 /// does that too, and with a better error message.
 pub const MIRI_DEFAULT_ARGS: &[&str] = &[
+    "-Zcodegen-backend=dummy",
```
bjorn3, line 168: The dummy codegen backend is a builtin codegen backend that is only capable of emitting rlibs containing the crate metadata. For every other crate type it emits an error after codegen. This is fine for miri as for `--crate-type bin`, miri exits before codegen.

cargo-miri/src/phases.rs
```diff
@@ -407,7 +395,7 @@ pub fn phase_rustc(mut args: impl Iterator<Item = String>, phase: RustcPhase) {
                 assert_eq!(val, "metadata");
             } else {
                 // For all other kinds of tests, we can just add our flag.
-                cmd.arg("--emit=metadata");
+                cmd.arg("--emit=metadata,link");
```
bjorn3: Is this still needed in the first place?

bjorn3:
Almost there. With https://github.com/rust-lang/rust/pull/148177 and https://github.com/rust-lang/rust/pull/148253 merged, the entire miri test suite passes after blessing a single change to a query cycle error. Doctests are broken however due to those requiring a check build for bin crates first, which the dummy codegen backend doesn't yet allow. Opened https://github.com/rust-lang/rust/pull/148299 for this.

cargo-miri/src/phases.rs
bjorn3: The `--emit metadata` is still necessary. And it shouldn't use `--emit link`.

bjorn3:
I hadn't noticed the cargo-miri-test test suite before. It currently has trouble with non-rlib/bin crate types. In particular cargo thinks it needs to build crates only available as cdylib (despite those never being usable) and the `dropping unsupported crate type` warnings have a non-deterministic order for some reason (edit: I think this is because of non-deterministic order in which rustc is executed. I assumed cargo would skip unsupported `--crate-type` args and report those warnings itself, but it seems like cargo just forwards them to rustc and expects rustc to produce the warnings.).

cargo-miri/src/setup.rs
```diff
@@ -160,7 +160,7 @@ pub fn setup(
 
     // Do the build.
     let status = SysrootBuilder::new(&sysroot_dir, target)
-        .build_mode(BuildMode::Check)
+        .build_mode(BuildMode::Build)
```
RalfJung:
```suggestion
        .build_mode(BuildMode::Build) // not a real build, since we use dummy codegen
```

test-cargo-miri/Cargo.toml
```diff
@@ -10,7 +10,7 @@ edition = "2024"
 
 [dependencies]
 byteorder = "1.0"
-cdylib = { path = "cdylib" }
+#cdylib = { path = "cdylib" }
```
RalfJung: What's up with this?
We added cdylib dependencies here because we got bug reports about Miri acting strange on them on real code, so we can't just stop supporting that.

RalfJung (commented):
If we can get rid of the hacks that patch `--emit` and `--extern`, that would definitely neat.

However I am concerned about those "unsupported crate type" warnings, that seems like something Miri users will then also see in the real world. We also need to ensure crate graphs that include cdylib keep working.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4648/a62e8f69d9244b784c663de3fa4489fe2ae6038f..c4b94af5cee4fdd33c20d1e352d29ccdfd46fae1)*

test-cargo-miri/Cargo.toml
bjorn3: Dependencies that are both an rlib and cdylib work just fine. However dependencies that are solely a cdylib won't work. But you wouldn't be able to use them anyway. Cargo for some reason insists on still building them, but you would never be able to use them in any other way than copying them from the target dir. It is not an artifact dependency that you can link against through a build script.

RalfJung: That looks deliberate: https://github.com/rust-lang/miri/commit/1923044705b2e9d54f7fa8f40e1c0bab5ea081ff explicitly talks about "cdylib-only". That commit was part of https://github.com/rust-lang/miri/pull/1710 which fixed https://github.com/rust-lang/miri/issues/1705.

RalfJung: https://github.com/rust-lang/miri/issues/1705 was about a crate that mixed cdylib and other crate types.

But still, a cdylib-only crate might be odd or useless, but it worked fine so far so I am concerned about breaking it as people might be relying on it.

test-cargo-miri/run.custom-target-dir.stderr.ref
```diff
@@ -1,2 +1,6 @@
+UNSUPPORTED CRATE TYPE
```
RalfJung: As I mentioned in the review summary, this isn't great. A lot of crate graphs contain cdylibs, so we'd be showing this warning to users a lot.

test-cargo-miri/Cargo.toml
bjorn3: hyper was an rlib+cdylib crate, not a cdylib-only crate. rlib+cdylib still works. Digging through the linked issues, all of them seem to be for rlib+cdylib crates and none for cdylib-only crates.

cargo-miri/src/phases.rs
RalfJung, line 1: We used to have code here that skips the build if the crate type is cdylib. That was obviously wrong, I just didn't realize at the time that there can be multiple crate types. But we could maybe instead filter the crate types and remove cdylib (and anything else that's unsupported), and then skip the build if there's no crate type left?

test-cargo-miri/Cargo.toml
RalfJung: Yeah that's what I just wrote in the [2nd comment](https://github.com/rust-lang/miri/pull/4648#discussion_r2484756232). Hard to say what people rely on though.

What's the error one gets when having a cdylib-only crate?

bjorn3: > What's the error one gets when having a cdylib-only crate?

```
error: cannot produce cdylib for `cdylib v0.1.0 (/home/gh-bjorn3/miri/test-cargo-miri/cdylib)` as the target `aarch64-unknown-linux-gnu` does not support these crate types
```

I've now moved the cdylib package from a dependency to just a workspace member. This way we still check that the mere presence of a cdylib-only package doesn't break miri, only when using it as dependency.

RalfJung: That error is kind of misleading since it's not about the target, it's about the codegen backend.

bjorn3: Yeah, this is unfortunately hard coded in cargo. I thought it was also hard coded in the code parsing the rustc warnings, but looks like that is not the case after all: https://github.com/rust-lang/cargo/blob/4406c1b96413e79ee770f81506a0697fb8f11404/src/cargo/core/compiler/build_context/target_info.rs#L669-L670 I will open a PR to change the rustc warning and see if I can open a PR to change the cargo error.

cargo-miri/src/phases.rs
bjorn3: I've now pushed a commit to do that.

RalfJung: That should then also fix the cdylib-only-dependency, shouldn't it?

bjorn3: It doesn't. The dummy codegen backend tells cargo that cdylib is not supported and then cargo reports this error before we can filter the `--crate-type cdylib`.

RalfJung: Ah, unfortunate.

cargo-miri/src/setup.rs
RalfJung: You didn't actually apply this suggestion?

src/bin/miri.rs
```diff
@@ -253,12 +233,15 @@ impl rustc_driver::Callbacks for MiriBeRustCompilerCalls {
     #[allow(rustc::potential_query_instability)] // rustc_codegen_ssa (where this code is copied from) also allows this lint
     fn config(&mut self, config: &mut Config) {
         if config.opts.prints.is_empty() && self.target_crate {
+            // Avoid warnings about unsupported crate types
+            #[allow(rustc::bad_opt_access)]
+            config.opts.crate_types.retain(|&c| c == CrateType::Executable || c == CrateType::Rlib);
```
RalfJung: We should probably only do this for target crates?

What happens if the list of crate types is empty after the filtering?

Also, why is it okay to allow `bad_opt_access` here?

src/bin/miri.rs
```diff
@@ -253,12 +233,15 @@ impl rustc_driver::Callbacks for MiriBeRustCompilerCalls {
...
+
             // Queries overridden here affect the data stored in `rmeta` files of dependencies,
             // which will be used later in non-`MIRI_BE_RUSTC` mode.
             config.override_queries = Some(|_, local_providers| {
-                // `exported_non_generic_symbols` and `reachable_non_generics` provided by rustc always returns
-                // an empty result if `tcx.sess.opts.output_types.should_codegen()` is false.
-                // In addition we need to add #[used] symbols to exported_symbols for `lookup_link_section`.
+                // We need to add #[used] symbols to exported_symbols for `lookup_link_section`.
+                // FIXME handle this somehow in rustc itself to avoid this hack
```
```suggestion
                // FIXME handle this somehow in rustc itself to avoid this hack.
```

test-cargo-miri/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = ["subcrate", "issue-1567", "exported-symbol-dep", "test-local-crate-detection"]
+members = ["cdylib", "subcrate", "issue-1567", "exported-symbol-dep", "test-local-crate-detection"]
```
RalfJung: At this point it is probably less confusing to just remove the `cdylib` crate entirely.

src/bin/miri.rs
bjorn3: > We should probably only do this for target crates?

This is gated behind `if self.target_crate`.

> What happens if the list of crate types is empty after the filtering?

Currently rustc treats that as implicit `--crate-type lib`, but in any case it shouldn't be possible to get an empty list here. Cargo will error if none of the crate types are bin or rlib as the dummy codegen backend only lists those crate types as supported.

> Also, why is it okay to allow bad_opt_access here?

This is long before the `TyCtxt` gets created that you are normally supposed to access `crate_types` through.

cargo-miri/src/setup.rs
bjorn3: I though I did. Should be fixed now.

test-cargo-miri/Cargo.toml
bjorn3: Removed

src/bin/miri.rs
RalfJung: > Currently rustc treats that as implicit --crate-type lib, but in any case it shouldn't be possible to get an empty list here. Cargo will error if none of the crate types are bin or rlib as the dummy codegen backend only lists those crate types as supported.

Please add an assertion then.

src/bin/miri.rs
```diff
@@ -253,12 +233,15 @@ impl rustc_driver::Callbacks for MiriBeRustCompilerCalls {
     #[allow(rustc::potential_query_instability)] // rustc_codegen_ssa (where this code is copied from) also allows this lint
     fn config(&mut self, config: &mut Config) {
         if config.opts.prints.is_empty() && self.target_crate {
+            // Avoid warnings about unsupported crate types
+            #[allow(rustc::bad_opt_access)]
```
```suggestion
            #[allow(rustc::bad_opt_access)] // tcx does not exist yet
```

src/bin/miri.rs
```diff
@@ -253,12 +233,15 @@ impl rustc_driver::Callbacks for MiriBeRustCompilerCalls {
     #[allow(rustc::potential_query_instability)] // rustc_codegen_ssa (where this code is copied from) also allows this lint
     fn config(&mut self, config: &mut Config) {
         if config.opts.prints.is_empty() && self.target_crate {
+            // Avoid warnings about unsupported crate types
```
```suggestion
            // Avoid warnings about unsupported crate types.
```

src/bin/miri.rs
bjorn3: > Currently rustc treats that as implicit --crate-type lib

Actually, it treats it as `--crate-type bin` just like if the user didn't pass `--crate-type` at all. I got it confused with the behavior when `--crate-type` is present on the cli, but all listed crate types are unsupported by the target or codegen backend.

In any case I added an assertion.

RalfJung:

@rustbot author

RalfJung:
All right, let's see how well this works :)

Issue #4642, opened 2025-10-22
Support `f32`/`f64` in native function calls.


weiznich:
Currently this results in an error like that one:

```
error: unsupported operation: unsupported return type for native call: "sqlite3_value_double"
    --> /home/weiznich/Documents/rust/diesel/diesel/src/sqlite/connection/sqlite_value.rs:186:18
     |
 186 |         unsafe { ffi::sqlite3_value_double(self.value.as_ptr()) }
     |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ unsupported operation occurred here
     |
     = help: this is likely not a bug in the program; it indicates that the program performed an operation that Miri does not support
     = note: BACKTRACE on thread `types::f32_from`:
     = note: inside `diesel::sqlite::SqliteValue::<'_, '_, '_>::read_double` at /home/weiznich/Documents/rust/diesel/diesel/src/sqlite/connection/sqlite_value.rs:186:18: 186:64
```

It would be nice to support floating point types there as well.

RalfJung:
Marking as good-first-issue since this should be fairly easy:

- add `f32` and `f64` in this match (for float arguments): https://github.com/rust-lang/miri/blob/ce6f40d9c4f529868435c7d410ca83fb244ecb39/src/shims/native_lib/mod.rs#L407
- and also add them in this match (for float return values): https://github.com/rust-lang/miri/blob/ce6f40d9c4f529868435c7d410ca83fb244ecb39/src/shims/native_lib/mod.rs#L95
- ensure we have tests for floats in both argument and return position

PR #4650, opened 2025-10-26
Support `f32`/`f64` in native function calls

FranciscoTGouveia:
Fixes rust-lang/miri#4642.

I would appreciate feedback on the following:
- Initially, I decided to add two test cases -- one for `f32` and another for `f64` -- but I later concluded that this might be redundant. If not, I am happy to reintroduce the second one;
- I have thoroughly looked into the documentation of `rustc_apfloat`, and the only suitable solution I have found was [`from_bits`](https://docs.rs/rustc_apfloat/0.2.3+llvm-462a31f5a5ab/rustc_apfloat/trait.Float.html#tymethod.from_bits). I may have overlooked something, but going through the codebase, I did not find any other way to do this;
- As all operations done (`to_bits()` -> cast -> `from_bits()`) are bit-preserving, I do not think we need to worry about signaling NaNs here. Am I correct thinking this should not be problematic?

Thank you in advance for your time and input!

### DIFF
--- a/src/shims/native_lib/mod.rs
+++ b/src/shims/native_lib/mod.rs
@@ -8,9 +8,11 @@ use libffi::middle::Type as FfiType;
 use rustc_abi::{HasDataLayout, Size};
 use rustc_data_structures::either;
 use rustc_middle::ty::layout::{HasTypingEnv, TyAndLayout};
-use rustc_middle::ty::{self, IntTy, Ty, UintTy};
+use rustc_middle::ty::{self, FloatTy, IntTy, Ty, UintTy};
 use rustc_span::Symbol;
 use serde::{Deserialize, Serialize};
+
+use self::helpers::ToSoft;
 
 mod ffi;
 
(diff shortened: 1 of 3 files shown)


src/shims/native_lib/mod.rs
```diff
@@ -138,6 +140,14 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     let x = unsafe { ffi::call::<usize>(fun, libffi_args) };
                     Scalar::from_target_usize(x.try_into().unwrap(), this)
                 }
+                ty::Float(FloatTy::F32) => {
+                    let x = unsafe { ffi::call::<f32>(fun, libffi_args) };
+                    Scalar::from_f32(Single::from_bits(u128::from(x.to_bits())))
```
```suggestion
                    Scalar::from_f32(x.to_soft())
```
This should work after importing the `ToSoft` trait.

FranciscoTGouveia: Oh, I overlooked that.
My apologies -- I have corrected it now.

RalfJung: No worries, this isn't easy to find. :)

RalfJung:

@rustbot author

FranciscoTGouveia:
I believe this was a spurious failure in the CI.

@rustbot ready

PR #4654, opened 2025-10-27
Refactor TreeVisitor and Access Relatedness.

royAmmerschuber:
Refactors `TreeVisitor` in preparation for wildcard support for tree borrows.

Also simplifies `AccessRelatedness` to only track if an access is foreign or local.

### DIFF
--- a/src/borrow_tracker/tree_borrows/perms.rs
+++ b/src/borrow_tracker/tree_borrows/perms.rs
@@ -640,7 +640,7 @@ mod propagation_optimization_checks {
     impl Exhaustive for AccessRelatedness {
         fn exhaustive() -> Box<dyn Iterator<Item = Self>> {
             use AccessRelatedness::*;
-            Box::new(vec![This, StrictChildAccess, AncestorAccess, CousinAccess].into_iter())
+            Box::new(vec![ForeignAccess, LocalAccess].into_iter())
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -716,13 +716,11 @@ mod propagation_optimization_checks {
                 // We now assert it is idempotent, and never causes UB.
                 // First, if the SIFA includes foreign reads, assert it is idempotent under foreign reads.
                 if access >= IdempotentForeignAccess::Read {
-                    // We use `CousinAccess` here. We could also use `AncestorAccess`, since `transition::perform_access` treats these the same.
-                    // The only place they are treated differently is in protector end accesses, but these are not handled here.
```
RalfJung, line 720: This says something about one place treating cousin and ancestor differently... so maybe we do have to keep that distinction? Or is the comment outdated?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -344,12 +346,11 @@ where
 {
     fn should_continue_at(
         &self,
-        this: &mut TreeVisitor<'_>,
+        this: &'_ mut TreeVisitor<'_>,
```
```suggestion
        this: &mut TreeVisitor<'_>,
```
Why this change?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -812,17 +813,21 @@ impl<'tcx> Tree {
         // `perms_range` is only for diagnostics (it is the range of
         // the `RangeMap` on which we are currently working).
         let node_skipper = |access_kind: AccessKind, args: &NodeAppArgs<'_>| -> ContinueTraversal {
-            let NodeAppArgs { node, perm, rel_pos } = args;
+            let node = args.nodes.get(args.idx).unwrap();
+            let perm = args.perms.get(args.idx);
+            let rel_pos = args.rel_pos;
```
RalfJung: Please just inline `rel_pos`, it's not worth a variable.

src/borrow_tracker/tree_borrows/perms.rs
royAmmerschuber: This seems outdated. The only difference the protector end access has to the normal one, is which locations it accesses.

RalfJung:

@rustbot author

Issue #4655, opened 2025-10-29
Incorrect data race UB reported when mixing atomic and non-atomic accesses


RalfJung:
Miri reports UB in this program (even with `-Zmiri-fixed-schedule`):
```rust
use std::sync::atomic::{AtomicUsize, Ordering};
use std::{ptr, thread};

const SIZE: usize = 256;

static mut ARRAY: [u8; SIZE] = [0; _];
// Everything strictly less than this has been initialized by the sender.
static POS: AtomicUsize = AtomicUsize::new(0);

fn main() {
    // Sender
    let th = std::thread::spawn(|| {
        for i in 0..SIZE {
            unsafe { ptr::write(&raw mut ARRAY[i], 1) };
            POS.store(i + 1, Ordering::Release);

            thread::yield_now();

            // We are the only writer, so we can do non-atomic reads as well.
            unsafe { (&raw const POS).cast::<usize>().read() };
        }
    });

    // Receiver
    loop {
        let i = POS.load(Ordering::Acquire);

        if i > 0 {
            unsafe { ptr::read(&raw const ARRAY[i - 1]) };
        }

        if i == SIZE {
            break;
        }

        thread::yield_now();
    }

    th.join().unwrap();
}
```
This is related to the non-atomic read; if we remove that, the program is fine.

Mixed atomic and non-atomic accesses are a Rust extension to the C++ memory model. Does our extension allow this code? I would think so; the non-atomic load is fully ordered wrt all writes to that location. Tokio certainly would like this to be allowed (https://github.com/tokio-rs/tokio/issues/7712). But apparently our extension to the data race checker disallows this...

Cc @cbeuw

RalfJung:
This is related to
```
    /// When a non-atomic access happens on a location that has been atomically accessed
    /// before without data race, we can determine that the non-atomic access fully happens
    /// after all the prior atomic writes so the location no longer needs to exhibit
    /// any weak memory behaviours until further atomic accesses.
    pub fn memory_accessed(&self, range: AllocRange, global: &DataRaceState) {
```
For non-atomic reads, that reasoning just isn't correct any more. (I think the comment was written before we allowed atomic non-atomic races.)

However, even so I would not expect removing the store buffers to ever introduce incorrect UB... somehow https://github.com/rust-lang/miri/pull/4577 made it so that that can happen now, apparently.

RalfJung:
The fix should be in the next nightly, 2025-11-03.

PR #4658, opened 2025-10-30
weak memory: fix non-atomic read clearing store buffer

RalfJung:
We used to clear the store buffer on every non-atomic read, which made sense back when non-atomic accesses were not allowed to race with atomic accesses -- but these days, this makes little sense, and it prevents legal behaviors from being explored (see the change in the `weak.rs` test). Furthermore, when we reconstruct the store buffer on the next atomic store, we assume that the "prior" state we are inserting corresponds to a non-atomic write, and thus we do not give it a sync clock -- but if we cleared the store buffer, that prior state might in fact be an atomic write, so the incorrect sync clock then leads to incorrect vector clocks which causes wrong UB and potentially also wrong behavior.

Fixes https://github.com/rust-lang/miri/issues/4655

### DIFF
--- a/src/concurrency/data_race.rs
+++ b/src/concurrency/data_race.rs
@@ -688,6 +688,8 @@ impl MemoryCellClocks {
         self.write = (index, thread_clocks.clock[index]);
         self.write_type = write_type;
         self.read.set_zero_vector();
+        // This is not an atomic location any more.
+        self.atomic_ops = None;
         Ok(())
     }
 }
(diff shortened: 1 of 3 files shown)


src/concurrency/data_race.rs
```diff
@@ -675,24 +674,23 @@ impl MemoryCellClocks {
...
+            atomic.write_vector <= thread_clocks.clock && atomic.read_vector <= thread_clocks.clock
+        }) {
+            return Err(DataRace);
+        }
+        // Record this access.
+        self.write = (index, thread_clocks.clock[index]);
+        self.write_type = write_type;
+        self.read.set_zero_vector();
+        // This is not an atomic location any more.
+        self.atomic_ops = None;
```
RalfJung, line 692: @cbeuw do you know if there is a particular reason why we kept the `atomic_ops` around even after a non-atomic write?

src/machine.rs
```diff
@@ -1533,14 +1533,11 @@ impl<'tcx> Machine<'tcx> for MiriMachine<'tcx> {
...
                 let _trace = enter_trace_span!(data_race::before_memory_read);
-                let AllocDataRaceHandler::Vclocks(data_race, weak_memory) = &alloc_extra.data_race
+                let AllocDataRaceHandler::Vclocks(data_race, _weak_memory) = &alloc_extra.data_race
                 else {
                     unreachable!();
                 };
-                data_race.read(alloc_id, range, NaReadType::Read, None, machine)?;
-                if let Some(weak_memory) = weak_memory {
-                    weak_memory.memory_accessed(range, machine.data_race.as_vclocks_ref().unwrap());
-                }
```
RalfJung, line 1546: @cbeuw is it correct to do absolutely nothing with the store buffers on a non-atomic read? I would think so, but better to double-check.^^

RalfJung:
I'm going to land this for now to fix the bug; we can always adjust more later.

Issue #4612, opened 2025-09-30
GenMC patches missing upstream


RalfJung:
@Patrick-6 can you make a list of GenMC changes that *need* to land upstream for us to be able to stop using our fork?
It's okay if those are links to the private repo, I should still be able to access them.

Patrick-6:
The changes are listed in this issue on the private repo: https://gitlab.mpi-sws.org/sav/genmc-tool/-/issues/34
Should I duplicate the list here?

RalfJung:
That's helpful, thanks.

PR #4661, opened 2025-11-03
Use Github URL to fetch GenMC

michaliskok:
Until now, GenMC was fetched from an "unofficial" repository. Now that all relevant changes for Miri have been merged upstream, GenMC can be fetched from its Github repository.

This PR changes the relevant URL, and suitably adapts MiriInterface.hpp.

r? @RalfJung 
Fixes https://github.com/rust-lang/miri/issues/4612


RalfJung:
Thanks! I have adjusted the `GenmcScalar` type to also use the new field name, so that they remain consistent.

PR #4663, opened 2025-11-04
Restore support for the wasm32-wasip2 target

bjorn3:
This adjusts find_mir_or_eval_fn to respect #[wasm_import_module] on wasm. To avoid having to thread through the import module everywhere, I opted to merge it with the item name with a $$$ as separator. This is unlikely to occur in a real import module name or item name.

This reverts commit c38128b0d43ae39f93bc8039b7950ab2000ec925 and makes a few minor adjustments.

### DIFF
--- a/src/shims/env.rs
+++ b/src/shims/env.rs
@@ -51,7 +51,7 @@ impl<'tcx> EnvVars<'tcx> {
         } else if ecx.tcx.sess.target.os == "windows" {
             EnvVars::Windows(WindowsEnvVars::new(ecx, env_vars)?)
         } else {
-            // Used e.g. for wasi
+            // For "none" targets (i.e., without an OS), and wasi.
             EnvVars::Uninit
         };
         ecx.machine.env_vars = env_vars;


README.md
```diff
@@ -220,6 +220,7 @@ degree documented below):
   - `solaris` / `illumos`: maintained by @devnexen. Supports the entire test suite.
   - `freebsd`: maintained by @YohDeadfall and @LorrensP-2158466. Supports the entire test suite.
   - `android`: **maintainer wanted**. Support very incomplete, but a basic "hello world" works.
+  - `wasi`: **maintainer wanted**. Support very incomplete, but a basic "hello world" works.
```
RalfJung, line 223: Are you willing to be listed as target maintainer for this?
I don't think I want to re-land this without a target maintainer.

src/machine.rs
```diff
@@ -1243,7 +1243,23 @@ impl<'tcx> Machine<'tcx> for MiriMachine<'tcx> {
             // foreign function
             // Any needed call to `goto_block` will be performed by `emulate_foreign_item`.
             let args = ecx.copy_fn_args(args); // FIXME: Should `InPlace` arguments be reset to uninit?
-            let link_name = Symbol::intern(ecx.tcx.symbol_name(instance).name);
+
+            let link_name = if ecx.tcx.sess.target.is_like_wasm
+                && let Some(module) =
+                    ecx.tcx.wasm_import_module_map(instance.def_id().krate).get(&instance.def_id())
```
RalfJung: I hope this code doesn't have to live in `machine.rs`.

If possible, please move it inside the `wasi` folder.
If that doesn't work, at least push it down to inside `emulate_foreign_item`.

src/shims/env.rs
```diff
@@ -51,7 +51,7 @@ impl<'tcx> EnvVars<'tcx> {
         } else if ecx.tcx.sess.target.os == "windows" {
             EnvVars::Windows(WindowsEnvVars::new(ecx, env_vars)?)
         } else {
-            // For "none" targets (i.e., without an OS).
+            // Used e.g. for wasi
```
RalfJung:
```suggestion
            // For "none" targets (i.e., without an OS), and wasi.
```

src/machine.rs
bjorn3: `emulate_foreign_item` is also called from `emulate_dyn_sym` which doesn't have access to the `Instance` as `dlsym` never gets passed one. I guess I could pass an `Option<Instance>` to `emulate_foreign_item` if you prefer that.

bjorn3: That wasn't as bad as I was afraid of.

RalfJung:
Marking as "blocked" since we should have a target maintainer.

src/shims/wasi/foreign_items.rs
```diff
@@ -0,0 +1,135 @@
...
+        link_name: Symbol,
+        abi: &FnAbi<'tcx, Ty<'tcx>>,
+        args: &[OpTy<'tcx>],
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx, EmulateItemResult> {
+        let this = self.eval_context_mut();
+
+        let (interface, name) = if let Some(instance) = instance
+            && let Some(module) =
+                this.tcx.wasm_import_module_map(instance.def_id().krate).get(&instance.def_id())
```
RalfJung, line 27: FWIW I think it is a bug in rustc that link_name returns the wrong name here... link_name should return whatever is necessary to find the symbol to link to. Apparently that's a tuple of two names for some targets -- fine. But instead we return a single name that's sometimes one component of that tuple, and sometimes completely irrelevant to linking, and then one needs to use out-of-band mechanisms to get the actual tuple -- that's not a good way to deal with this IMO.

bjorn3: With core wasm the import name is a tuple of two names. However for the [linking tool convention](https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md) the symbol name used for resolving symbols within the linker session is entirely separate from the name used for importing items from outside the final linked wasm module. The symbol name is ignored after linking, while the import name tuple is used at runtime. Rustc's link_name returns the symbol name, while this code computes the import name tuple.

RalfJung: > Rustc's link_name returns the symbol name, while this code computes the import name tuple.

That's at best very confusing. I would expect that query to return the information about what symbol this links to (what I would call the "link name", and what wasm apparently calls "import name"). Apparently we don't have any query that returns this information, instead every backend duplicates this logic?

As you said elsewhere, this isn't even wasm-specific. Windows also uses a tuple for names, IIUC.

bjorn3: For Windows we also use the link name and there is no option to use the import name. The import library is what defines the mapping from link name to dll name + import name. So the linker only sees the symbol name (as returned by `tcx.symbol_name()` in rustc) when resolving symbols. On wasm the symbol table directly contains the import name tuple rather than putting it in a separate file, but I believe the linker still only matches on the symbol name and handles the imports later.

(I don't think wasm formally uses symbol name/import name. I made this naming scheme up to help describe what I meant)

RalfJung:
I will close this for now since there's nothing actionable to review until we have someone who volunteers to be target maintainer.

Issue #4464, opened 2025-07-11
`posix_fallocate` not supported


nazar-pc:
```
error: unsupported operation: can't call foreign function `posix_fallocate` on OS `linux`
   --> /home/nazar-pc/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/fs2-0.4.3/src/unix.rs:103:24
    |
103 |     let ret = unsafe { libc::posix_fallocate(file.as_raw_fd(), 0, len as libc::off_t) };
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ unsupported operation occurred here
    |
    = help: this means the program tried to do something Miri does not support; it does not indicate a bug in the program
    = note: BACKTRACE on thread `tests::other_op`:
    = note: inside `fs2::unix::allocate` at /home/nazar-pc/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/fs2-0.4.3/src/unix.rs:103:24: 103:86
    = note: inside `<std::fs::File as fs2::FileExt>::allocate` at /home/nazar-pc/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/fs2-0.4.3/src/lib.rs:104:9: 104:33
```

It is meant to pre-allocate disk space to ensure writes do not fail within provided length with lack of disk space afterward. I'd say it is mostly the same as `File::set_len()` except it avoids sparse files on disk.

RalfJung:
I think the best we can do here is implement it like set_len. Would that be useful?

We're generally avoiding non-portable implementations unless they are needed for something super common, so the part about the disk space is not something we're likely going to get.

nazar-pc:
I think `set_len()`-like implementation is reasonable and helpful already

PR #4664, opened 2025-11-04
Implement `posix_fallocate` with `set_len()` functionality

LorrensP-2158466:
should close rust-lang/miri#4464.

I used [this man page](https://man7.org/linux/man-pages/man3/posix_fallocate.3p.html) for the implementation. 
Changes included in this pr:
- impl `posix_fallocate` and `posix_fallocate64` (same as truncate versions because of libc::off_t)
- use the impl in `unix/foreign_items`.
- tests in `pass-dep/libc/libc-fs.rs`.

### DIFF
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -487,6 +487,48 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // fadvise is only informational, we can ignore it.
                 this.write_null(dest)?;
             }
+
+            "posix_fallocate" => {
+                // posix_fallocate is not supported by macos.
+                this.check_target_os(
+                    &[Os::Linux, Os::FreeBsd, Os::Solaris, Os::Illumos, Os::Android],
+                    link_name,
+                )?;
+                let [fd, offset, len] = this.check_shim_sig(
+                    shim_sig!(extern "C" fn(i32, libc::off_t, libc::off_t) -> i32),
+                    link_name,
+                    abi,
+                    args,
+                )?;
+
+                let fd = this.read_scalar(fd)?.to_i32()?;
+                // We don't support platforms which have libc::off_t bigger than 64 bits.
+                let offset =
+                    i64::try_from(this.read_scalar(offset)?.to_int(offset.layout.size)?).unwrap();
+                let len = i64::try_from(this.read_scalar(len)?.to_int(len.layout.size)?).unwrap();
+
+                let result = this.posix_fallocate(fd, offset, len)?;
+                this.write_scalar(result, dest)?;
+            }
+
+            "posix_fallocate64" => {
+                // posix_fallocate64 is only supported on Linux and Android
+                this.check_target_os(&[Os::Linux, Os::Android], link_name)?;
+                let [fd, offset, len] = this.check_shim_sig(
+                    shim_sig!(extern "C" fn(i32, libc::off64_t, libc::off64_t) -> i32),
+                    link_name,
+                    abi,
+                    args,
+                )?;
+
+                let fd = this.read_scalar(fd)?.to_i32()?;
+                let offset = this.read_scalar(offset)?.to_i64()?;
+                let len = this.read_scalar(len)?.to_i64()?;
+
+                let result = this.posix_fallocate(fd, offset, len)?;
+                this.write_scalar(result, dest)?;
+            }
+
             "realpath" => {
                 let [path, resolved_path] = this.check_shim_sig(
                     shim_sig!(extern "C" fn(*const _, *mut _) -> *mut _),
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -1202,6 +1202,65 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         }
     }
 
+    /// NOTE: According to the man page of `possix_fallocate`, it returns the error code instead
+    /// of setting `errno`.
+    fn posix_fallocate(
+        &mut self,
+        fd_num: i32,
+        offset: i64,
+        len: i64,
+    ) -> InterpResult<'tcx, Scalar> {
+        let this = self.eval_context_mut();
+
+        // Reject if isolation is enabled.
+        if let IsolatedOp::Reject(reject_with) = this.machine.isolated_op {
+            this.reject_in_isolation("`posix_fallocate`", reject_with)?;
+            // Return error code "EBADF" (bad fd).
+            return interp_ok(this.eval_libc("EBADF"));
+        }
+
+        // EINVAL is returned when: "offset was less than 0, or len was less than or equal to 0".
+        if offset < 0 || len <= 0 {
+            return interp_ok(this.eval_libc("EINVAL"));
+        }
+
+        // Get the file handle.
+        let Some(fd) = this.machine.fds.get(fd_num) else {
+            return interp_ok(this.eval_libc("EBADF"));
+        };
+        let file = match fd.downcast::<FileHandle>() {
+            Some(file_handle) => file_handle,
+            // Man page specifies to return ENODEV if `fd` is not a regular file.
+            None => return interp_ok(this.eval_libc("ENODEV")),
+        };
+
+        if !file.writable {
+            // The file is not writable.
+            return interp_ok(this.eval_libc("EBADF"));
+        }
+
+        let current_size = match file.file.metadata() {
+            Ok(metadata) => metadata.len(),
+            Err(err) => return this.io_error_to_errnum(err),
+        };
+        // Checked i64 addition, to ensure the result does not exceed the max file size.
+        let new_size = match offset.checked_add(len) {
+            // `new_size` is definitely non-negative, so we can cast to `u64`.
+            Some(new_size) => u64::try_from(new_size).unwrap(),
+            None => return interp_ok(this.eval_libc("EFBIG")), // new size too big
+        };
+        // If the size of the file is less than offset+size, then the file is increased to this size;
+        // otherwise the file size is left unchanged.
+        if current_size < new_size {
+            interp_ok(match file.file.set_len(new_size) {
+                Ok(()) => Scalar::from_i32(0),
+                Err(e) => this.io_error_to_errnum(e)?,
+            })
+        } else {
+            interp_ok(Scalar::from_i32(0))
+        }
+    }
+
     fn fsync(&mut self, fd_op: &OpTy<'tcx>) -> InterpResult<'tcx, Scalar> {
         // On macOS, `fsync` (unlike `fcntl(F_FULLFSYNC)`) does not wait for the
         // underlying disk to finish writing. In the interest of host compatibility,
--- a/tests/pass-dep/libc/libc-fs.rs
+++ b/tests/pass-dep/libc/libc-fs.rs
@@ -36,6 +36,10 @@ fn main() {
     test_posix_realpath_errors();
     #[cfg(target_os = "linux")]
     test_posix_fadvise();
+    #[cfg(not(target_os = "macos"))]
+    test_posix_fallocate::<libc::off_t>(libc::posix_fallocate);
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    test_posix_fallocate::<libc::off64_t>(libc::posix_fallocate64);
     #[cfg(target_os = "linux")]
     test_sync_file_range();
     test_isatty();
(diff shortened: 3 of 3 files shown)


src/shims/unix/foreign_items.rs
```diff
@@ -486,6 +486,41 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // fadvise is only informational, we can ignore it.
                 this.write_null(dest)?;
             }
+
+            // only macos doesn't support `posix_fallocate`
+            "posix_fallocate" if &*this.tcx.sess.target.os != "macos" => {
```
LorrensP-2158466: Not sure if this is the correct way, from what I could find online, only macos doesn't implement `posix_fallocate`.

src/shims/unix/fs.rs
```diff
@@ -1197,6 +1198,65 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        // EINVAL is returned when: "offset was less than 0, or len was less than or equal to 0".
+        if offset < 0 || len <= 0 {
+            return interp_ok(einval);
+        }
+
+        if file.writable {
+            let current_size = match file.file.metadata() {
+                Ok(metadata) => metadata.size(),
+                Err(err) => return this.io_error_to_errnum(err),
+            };
```
LorrensP-2158466: I could not find another way to get the file size in Miri, is this correct?

src/shims/unix/foreign_items.rs
RalfJung: If that is the case then the test should also be for all OSes except macos.

RalfJung: Also, please check `pipe2` for how we make a shim available only on some OSes. Please do not use a match guard.

src/shims/unix/fs.rs
```diff
@@ -1197,6 +1197,65 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        offset: i128,
+        len: i128,
+    ) -> InterpResult<'tcx, Scalar> {
+        let this = self.eval_context_mut();
+
+        // According to the man page of `possix_fallocate`, it returns the error code instead
+        // of setting `errno`.
+        let ebadf = Scalar::from_i32(this.eval_libc_i32("EBADF"));
+        let einval = Scalar::from_i32(this.eval_libc_i32("EINVAL"));
+        let enodev = Scalar::from_i32(this.eval_libc_i32("ENODEV"));
```
RalfJung: I think it's better to just inline these where they are needed; the error path shouldn't take so much space on the happy path.

Also, you can use `eval_libc` instead of `Scalar::from_i32(eval_libc_i32)`.

src/shims/unix/fs.rs
```diff
@@ -1197,6 +1197,65 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // Man page specifies to return ENODEV if `fd` is not a regular file.
+            None => return interp_ok(enodev),
+        };
+
+        // EINVAL is returned when: "offset was less than 0, or len was less than or equal to 0".
+        if offset < 0 || len <= 0 {
+            return interp_ok(einval);
+        }
+
+        if file.writable {
```
RalfJung: Please make this more like
```rust
if !file.writable { return error; }
```
to avoid rightwards drift

src/shims/unix/fs.rs
```diff
@@ -1197,6 +1197,65 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            let current_size = match file.file.metadata() {
+                Ok(metadata) => metadata.len(),
+                Err(err) => return this.io_error_to_errnum(err),
+            };
+            let new_size = match offset.checked_add(len) {
+                Some(size) => size.try_into().unwrap(), // We just checked negative `offset` and `len`.
```
RalfJung: Please change this code so it is clear what type we are converting this into.

Also the comment does not explain why the size would fit into the target type.

src/shims/unix/fs.rs
```diff
@@ -1197,6 +1197,65 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                None => return interp_ok(einval),
+            };
+            // `posix_fallocate` only specifies increasing the file size.
+            if current_size < new_size {
+                let result = file.file.set_len(new_size);
+                let result = this.try_unwrap_io_result(result.map(|_| 0i32))?;
```
RalfJung: This can set the errno. Didn't you say the function does not do that?

src/shims/unix/fs.rs
```diff
@@ -1197,6 +1197,65 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        if offset < 0 || len <= 0 {
+            return interp_ok(einval);
+        }
+
+        if file.writable {
+            let current_size = match file.file.metadata() {
+                Ok(metadata) => metadata.len(),
+                Err(err) => return this.io_error_to_errnum(err),
+            };
+            let new_size = match offset.checked_add(len) {
```
RalfJung: IIUC, if this operation overflows we should return EFBIG.

RalfJung (changes requested):
Thanks for the PR!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4664/3ab6735110fb308177900c33ff865b5a767979a8..64be5bb1c9cc7df28913e5e2208d0e6172085d9a)*

src/shims/unix/fs.rs
LorrensP-2158466: Oops, used the wrong function, thanks!

src/shims/unix/fs.rs
```diff
@@ -1197,6 +1197,69 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            Ok(metadata) => metadata.len(),
+            Err(err) => return this.io_error_to_errnum(err),
+        };
+        let new_size = match offset.checked_add(len) {
+            // u64::from(i128) can fail if:
+            //  - the value is negative, but we checed this above with `offset < 0 || len <= 0`
+            //  - the value is too big/small to fit in u64, this should not happen since the callers
+            //    check if the value is a `i32` or `i64`.
+            // So this unwrap is safe to do.
+            Some(size) => u64::try_from(size).unwrap(),
```
LorrensP-2158466: I'm actually unsure if this is allowed in Miri, the callers to check if they fit inside of the `layout.size`, which can only be `libc::off_t`(i32) or `libc::off64_t`(i64).

LorrensP-2158466:
Had to rebase because of the new `Os` enum.

src/shims/unix/fs.rs
RalfJung: Oh so we are adding two `i64` here and hence the result fits in a `u64`?

Is that what the docs say for this function? I would kind of expect that there would be a "too big" error if the result exceeds `i64::MAX`.

LorrensP-2158466: I should have asked this sooner, excuse me. The docs do not say anything about integer overflows, but it does specify returning "EFBIG" when this size is larger than the maximum file size. I assumed we should thus return "EFBIG" when the addition overflows.

But I don't know how to get the maximum file size in Miri.

RalfJung: I think the maximum file size is i64::MAX.

LorrensP-2158466: then this should be correct right?

RalfJung: No, it seems wrong.

LorrensP-2158466: You're correct, we're doing the `checked_add` on a `i128`, so we won't report an error if we go above `i64::MAX`. I'll update it.

RalfJung: I think you are still not erroring if both `offset` and `len` are `i64::MAX` (to just give an example)?

RalfJung: Ah no, the `checked_add` is now on i64, I see.

src/shims/unix/fs.rs
```diff
@@ -1202,6 +1202,67 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        let current_size = match file.file.metadata() {
+            Ok(metadata) => metadata.len(),
+            Err(err) => return this.io_error_to_errnum(err),
+        };
+        let new_size = match offset.checked_add(len) {
+            // u64::from(i64) can fail if:
+            //  - the value is negative, but we checed this above with `offset < 0 || len <= 0`
+            // So this `unwrap` is safe to do so.
+            Some(new_size) => u64::try_from(new_size).unwrap(),
```
RalfJung:
```suggestion
            // `new_size` is definitely non-negative, so we can cast to `u64`.
            Some(new_size) => u64::try_from(new_size).unwrap(),
```

src/shims/unix/fs.rs
```diff
@@ -1202,6 +1202,67 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        if !file.writable {
+            // The file is not writable.
+            return interp_ok(this.eval_libc("EBADF"));
+        }
+
+        let current_size = match file.file.metadata() {
+            Ok(metadata) => metadata.len(),
+            Err(err) => return this.io_error_to_errnum(err),
+        };
+        let new_size = match offset.checked_add(len) {
```
RalfJung, line 1247:
```suggestion
        // Checked i64 addition, to ensure the result does not exceed the max file size.
        let new_size = match offset.checked_add(len) {
```

src/shims/unix/fs.rs
```diff
@@ -1202,6 +1202,67 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // u64::from(i64) can fail if:
+            //  - the value is negative, but we checed this above with `offset < 0 || len <= 0`
+            // So this `unwrap` is safe to do so.
+            Some(new_size) => u64::try_from(new_size).unwrap(),
+            None => return interp_ok(this.eval_libc("EFBIG")), // new size to big
+        };
+        // `posix_fallocate` only specifies increasing the file size.
```
RalfJung:
```suggestion
        // If the size of the file is less than offset+size, then the file is increased to this size;
        // otherwise the file size is left unchanged.
```

src/shims/unix/fs.rs
```diff
@@ -1202,6 +1202,67 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        if current_size < new_size {
+            let result = match file.file.set_len(new_size) {
+                Ok(()) => Scalar::from_i32(0),
+                Err(e) => this.io_error_to_errnum(e)?,
+            };
+
+            interp_ok(result)
```
RalfJung:
```suggestion
            interp_ok(match file.file.set_len(new_size) {
                Ok(()) => Scalar::from_i32(0),
                Err(e) => this.io_error_to_errnum(e)?,
            })
```

src/shims/unix/fs.rs
```diff
@@ -1202,6 +1202,67 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        // EINVAL is returned when: "offset was less than 0, or len was less than or equal to 0".
+        if offset < 0 || len <= 0 {
+            return interp_ok(this.eval_libc("EINVAL"));
+        }
+
+        if !file.writable {
+            // The file is not writable.
+            return interp_ok(this.eval_libc("EBADF"));
+        }
```
RalfJung, line 1240: The order in which you do these checks looks a bit random, let's clean that up
```suggestion
        // EINVAL is returned when: "offset was less than 0, or len was less than or equal to 0".
        if offset < 0 || len <= 0 {
            return interp_ok(this.eval_libc("EINVAL"));
        }

        // Get the file handle.
        let Some(fd) = this.machine.fds.get(fd_num) else {
            return interp_ok(this.eval_libc("EBADF"));
        };
        let file = match fd.downcast::<FileHandle>() {
            Some(file_handle) => file_handle,
            // Man page specifies to return ENODEV if `fd` is not a regular file.
            None => return interp_ok(this.eval_libc("ENODEV")),
        };
        if !file.writable {
            // The file is not writable.
            return interp_ok(this.eval_libc("EBADF"));
        }
```

RalfJung (dismissed):
Thanks, just some minor nits. :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4664/37c660f4e1e6b742f2d4c3f7edd43752c7f98581..c5d164a8d0291a8bf15b57d43b5cad6a3e98b381)*

RalfJung:

@rustbot author

PR #4666, opened 2025-11-04
x86/rounding-error test: add a bit of slack

RalfJung:
@eduardosm either something is wrong with the `rsqrt` implementation, or with the test... when run with many seeds, it fails saying
```
correct result: 0.5, got: 0.49987793
that's an error of -0.00024414063 = 2^-11.999999
```

### DIFF
--- a/tests/pass/shims/x86/rounding-error.rs
+++ b/tests/pass/shims/x86/rounding-error.rs
@@ -11,25 +11,27 @@ use std::arch::x86_64::*;
 use std::collections::HashSet;
 
 fn main() {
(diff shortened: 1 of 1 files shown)


eduardosm:
The relative error is exactly 2^-12, but calculating the `log2` introduces some error, I believe that due to Miri introducing non-determinism to math functions. I think we should compare the relative error against `1.0/4096.0` and only use the log value in the assert message.

RalfJung:
>  but calculating the log2 introduces some error, 

:facepalm: of course. Thanks!

Issue #4670, opened 2025-11-05
Immovable synchronization primitives do not detect swapping


RalfJung:
pthread synchronization primitives must not be moved. We currently do detect some cases of that, but if you just swap the contents of two synchronization primitives, that's a total NOP and the program will continue -- whereas a this is clearly UB in a real execution (and in some pthread implementations, e.g. on macOS). We should get better at detecting such things.

This also affects os_unfair_lock to some extent, which we currently implement as kind-of movable, though arguably we are being too permissive here.

We probably want to extend the before_memory_write hook to also do *something* with the synchronization primitive metadata at the affected bytes -- for futexes, that's completely fine, but all constructs that store data "in userspace" need to be marked as broken on such a write. Also, ifn we refactor this anyway it would be great to get rid of the LAZY_INIT_COOKIE scheme... that's such a hack.

PR #4672, opened 2025-11-07
Make pthread and other "native" synchronization objects more resilient

RalfJung:
Overwriting the object will destroy its magic metadata, so it needs to be re-initialized (which will fail since the bytes no longer match the static initializer). Furthermore, reading or writing the object *while threads are queued* is insta-UB.

Fixes https://github.com/rust-lang/miri/issues/4670

### DIFF
--- a/src/shims/unix/sync.rs
+++ b/src/shims/unix/sync.rs
@@ -1,6 +1,6 @@
 use rustc_abi::Size;
 
-use crate::concurrency::sync::SyncObj;
+use crate::concurrency::sync::{AccessKind, SyncObj};
 use crate::*;
 
 /// Do a bytewise comparison of the two places. This is used to check if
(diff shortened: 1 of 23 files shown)


src/shims/unix/sync.rs
```diff
@@ -1,39 +1,28 @@
...
-        }
-    }
+    let left_bytes = ecx.read_bytes_ptr_strip_provenance(left.ptr(), size)?;
+    let right_bytes = ecx.read_bytes_ptr_strip_provenance(right.ptr(), size)?;
 
-    interp_ok(true)
+    interp_ok(left_bytes == right_bytes)
 }
 
+// The in0memory marker values we use to indicate whether objects have been initialized.
```
oli-obk:
```suggestion
// The in memory marker values we use to indicate whether objects have been initialized.
```

oli-obk (approved):
Oh nice scheme. I like how our existing before_memory_read and extra alloc data system makes this all fairly straight forward to support

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4672/0e6246b229ce27b599ce787632b5a024df5626ce..4c37f175f868cdf29f940ac56e7647ec8ea9e4c0)*

Issue #4374, opened 2025-06-04
Miri spurious wake up in epoll?


tiif:
Miri unexpectedly hits ``_ = &mut write_fut => unreachable!()``. This is a case of Miri having different behavior from real system.  Also see the brief discussion [here](https://github.com/tokio-rs/tokio/pull/7382#discussion_r2124242611).

```rust
#[tokio::test]
async fn poll_fns() {
    let (a, b) = socketpair();
    let afd_a = Arc::new(AsyncFd::new(a).unwrap());
    let afd_b = Arc::new(AsyncFd::new(b).unwrap());

    // Fill up the write side of A
    let mut bytes = 0;
    while let Ok(amt) = afd_a.get_ref().write(&[0; 512]) {
        bytes += amt;
    }

    let waker = TestWaker::new();

    assert_pending!(afd_a.as_ref().poll_read_ready(&mut waker.context()));

    let afd_a_2 = afd_a.clone();
    let r_barrier = Arc::new(tokio::sync::Barrier::new(2));
    let barrier_clone = r_barrier.clone();

    let read_fut = tokio::spawn(async move {
        // Move waker onto this task first
        assert_pending!(poll!(std::future::poll_fn(|cx| afd_a_2
            .as_ref()
            .poll_read_ready(cx))));
        barrier_clone.wait().await;

        let _ = std::future::poll_fn(|cx| afd_a_2.as_ref().poll_read_ready(cx)).await;
    });

    let afd_a_2 = afd_a.clone();
    let w_barrier = Arc::new(tokio::sync::Barrier::new(2));
    let barrier_clone = w_barrier.clone();

    let mut write_fut = tokio::spawn(async move {
        // Move waker onto this task first
        assert_pending!(poll!(std::future::poll_fn(|cx| afd_a_2
            .as_ref()
            .poll_write_ready(cx))));
        barrier_clone.wait().await;

        let _ = std::future::poll_fn(|cx| afd_a_2.as_ref().poll_write_ready(cx)).await;
    });

    r_barrier.wait().await;
    w_barrier.wait().await;

    let readable = afd_a.readable();
    tokio::pin!(readable);

    tokio::select! {
        _ = &mut readable => unreachable!(),
        _ = tokio::task::yield_now() => {}
    }

    // Make A readable. We expect that 'readable' and 'read_fut' will both complete quickly
    afd_b.get_ref().write_all(b"0").unwrap();

    let _ = tokio::join!(readable, read_fut);

[...]

PR #4676, opened 2025-11-10
epoll: do proper edge detection inside the epoll system

RalfJung:
Previously, the individual FDs were expected to only call `epoll_send_fd_ready_events` when something changed. However, they couldn't actually know which changes are relevant to the epoll interests, so instead move the change detection into the epoll core. To ensure we behave consistently with Linux, some events need to force a notification even if there was no change.

Also, fix a bug where when an event got un-available, we'd still keep it in the ready list. (Without the above, that would cause tons of existing tests to fail.) And fix EPOLL_CTL_MOD to make it update the `data` of events already in the ready list.

Fixes https://github.com/rust-lang/miri/issues/4374.

### DIFF
--- a/tests/deps/Cargo.lock
+++ b/tests/deps/Cargo.lock
@@ -73,6 +73,79 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37909eebbb50d72f9059c3b6d82c0463f2ff062c9e95845c43a6c9c0355411be"
 
 [[package]]
+name = "futures"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "65bc07b1a8bc7c85c5f2e110c476c7389b4554ba72af57d8445ea63a576b0876"
+dependencies = [
+ "futures-channel",
+ "futures-core",
+ "futures-io",
+ "futures-sink",
+ "futures-task",
+ "futures-util",
+]
+
+[[package]]
+name = "futures-channel"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2dff15bf788c671c1934e366d07e30c1814a8ef514e1af724a602e8a2fbe1b10"
+dependencies = [
+ "futures-core",
+ "futures-sink",
+]
+
+[[package]]
+name = "futures-core"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "05f29059c0c2090612e8d742178b0580d2dc940c837851ad723096f87af6663e"
+
+[[package]]
+name = "futures-io"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9e5c1b78ca4aae1ac06c48a526a655760685149f0d465d21f37abfe57ce075c6"
+
+[[package]]
+name = "futures-macro"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "162ee34ebcb7c64a8abebc059ce0fee27c2262618d7b60ed8faf72fef13c3650"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
+name = "futures-sink"
+version = "0.3.31"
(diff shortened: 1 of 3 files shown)


src/shims/unix/linux_like/eventfd.rs
```diff
@@ -217,7 +217,9 @@ fn eventfd_write<'tcx>(
 
             // The state changed; we check and update the status of all supported event
             // types for current file description.
-            ecx.epoll_send_fd_ready_events(eventfd)?;
+            // Linux seems to cause spurious wakeups here, and Tokio seems to rely on that
+            // (see <https://github.com/rust-lang/miri/issues/4374#issuecomment-3508873493>).
+            ecx.epoll_send_fd_ready_events(eventfd, /* force_edge */ true)?;
```
RalfJung, line 223: @Darksonn this is the suspicious part... if I make this `false`, the [Tokio test](https://github.com/rust-lang/miri/issues/4374#issuecomment-3508873493) deadlocks....

tests/pass-dep/libc/libc-epoll-no-blocking.rs
```diff
@@ -254,6 +258,34 @@ fn test_epoll_eventfd() {
...
     let expected_value = u64::try_from(fd).unwrap();
     check_epoll_wait::<8>(epfd, &[(expected_event, expected_value)]);
+
+    // Write to the eventfd again.
+    let res = unsafe { libc_utils::write_all(fd, sized_8_data.as_ptr() as *const libc::c_void, 8) };
+    assert_eq!(res, 8);
+
+    // This does not change the status, so we should get no event.
+    // However, Linux performs a spurious wakeup.
+    check_epoll_wait::<8>(epfd, &[(expected_event, expected_value)]);
```
RalfJung, line 290: ... Setting `force_edge` to `false` makes it so that we get the empty event list here. Which is what I would expect since the readiness did not change. However, Linux does seem to always emit an epoll event on eventfd writes, even if readiness does not change -- and it looks like Tokio relies on that? At least that is the only explanation I have for the deadlocks I have been seeing.

src/shims/unix/linux_like/epoll.rs
```diff
@@ -345,31 +347,37 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                        e.insert(EpollEventInterest { events, data, prev_events: 0 }),
                 }
             } else {
                 // Modify the existing interest.
                 let Some(interest) = interest_list.get_mut(&epoll_key) else {
                     return this.set_last_error_and_return_i32(LibcError("ENOENT"));
                 };
-                *interest = new_interest;
+                interest.events = events;
+                // FIXME what about the data?
```
RalfJung: I need to add a test ensuring that when we change the data part, this actually takes effect. And I need to check whether it affects things that are already in the ready list.

tests/pass-dep/libc/libc-epoll-no-blocking.rs
Darksonn: Yes, it looks like Tokio does rely on that. Mio has no logic to consume the notification except when the 64-bit counter overflows. I guess this avoids a `read` syscall, so it makes some sort of sense. See [`eventfd.rs`][1] for details.

[1]: https://github.com/tokio-rs/mio/blob/master/src/sys/unix/waker/eventfd.rs

RalfJung: Okay, thanks for confirming!

RalfJung: Fun, Mio also references https://www.illumos.org/issues/16700 -- looks like Illumos was also requested to be bug-for-bug compatible with Linux here.

I wouldn't mind this so much if it was documented, but epoll overall has pretty bad documentation I have to say. :/

Darksonn: :D

RalfJung: @Darksonn FWIW it seems like Linux triggers a wakeup even if you write `[0u8; 8]` to the eventfd. That way you'd never even have to worry about overflow.

PR #4678, opened 2025-11-10
add shim for avx512 ternarylogic functions

folkertdev:
Adds the first dedicated avx512 shim, so that requires some setup.

The ternary logic functions are used in avx512 crc32 implementations, I'm specifically implementing one in zlib-rs right now.

### DIFF
--- a/src/shims/x86/avx512.rs
+++ b/src/shims/x86/avx512.rs
@@ -1,0 +1,85 @@
+use rustc_abi::CanonAbi;
+use rustc_middle::ty::Ty;
+use rustc_span::Symbol;
+use rustc_target::callconv::FnAbi;
+
+use crate::*;
+
+impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
+pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    fn emulate_x86_avx512_intrinsic(
+        &mut self,
+        link_name: Symbol,
+        abi: &FnAbi<'tcx, Ty<'tcx>>,
+        args: &[OpTy<'tcx>],
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx, EmulateItemResult> {
+        let this = self.eval_context_mut();
+        // Prefix should have already been checked.
+        let unprefixed_name = link_name.as_str().strip_prefix("llvm.x86.avx512.").unwrap();
+
+        match unprefixed_name {
+            // Used by the ternarylogic functions.
+            "pternlog.d.128" | "pternlog.d.256" | "pternlog.d.512" => {
+                this.expect_target_feature_for_intrinsic(link_name, "avx512f")?;
+                if matches!(unprefixed_name, "pternlog.d.128" | "pternlog.d.256") {
+                    this.expect_target_feature_for_intrinsic(link_name, "avx512vl")?;
+                }
+
+                let [a, b, c, imm8] =
+                    this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
+
+                assert_eq!(dest.layout, a.layout);
+                assert_eq!(dest.layout, b.layout);
+                assert_eq!(dest.layout, c.layout);
+
+                // The signatures of these operations are:
+                //
+                // ```
+                // fn vpternlogd(a: i32x16, b: i32x16, c: i32x16, imm8: i32) -> i32x16;
+                // fn vpternlogd256(a: i32x8, b: i32x8, c: i32x8, imm8: i32) -> i32x8;
+                // fn vpternlogd128(a: i32x4, b: i32x4, c: i32x4, imm8: i32) -> i32x4;
+                // ```
+                //
+                // The element type is always a 32-bit integer, the width varies.
+
+                let (a, _a_len) = this.project_to_simd(a)?;
+                let (b, _b_len) = this.project_to_simd(b)?;
+                let (c, _c_len) = this.project_to_simd(c)?;
+                let (dest, dest_len) = this.project_to_simd(dest)?;
+
+                // Compute one lane with ternary table.
+                let tern = |xa: u32, xb: u32, xc: u32, imm: u32| -> u32 {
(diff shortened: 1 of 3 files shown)


src/shims/x86/avx512.rs
```diff
@@ -0,0 +1,77 @@
...
+
+                let (a, _a_len) = this.project_to_simd(a)?;
+                let (b, _b_len) = this.project_to_simd(b)?;
+                let (c, _c_len) = this.project_to_simd(c)?;
+                let (dest, dest_len) = this.project_to_simd(dest)?;
+
+                let lane_size = dest.layout.field(this, 0).size;
+
+                // compute one lane with ternary table.
+                let tern = |xa: u32, xb: u32, xc: u32, imm: u32| -> u32 {
```
RalfJung, line 52: So the elements always fit in a u32?

Please add a comment explaining why this logic is correct even if the type is actually smaller.

folkertdev: Actually the elements are always 32-bit, so I've clarified the code a bit and added the signatures of the LLVM intrinsics.

tests/pass/shims/x86/intrinsics-x86-avx512.rs
RalfJung, line 1: In all of these tests, the result is 0, right? So the test isn't actually checking that we are doing any of this fancy bit shifting and so... could you add some tests where something actually happens? Preferably with different values on different lanes so we test multiple things at once (and we see that the lanes aren't all mixed up), and with some different values for the IMM8 as well. Since it's all the same implementation for us, just doing it for `_mm512_ternarylogic_epi32` is enough.

folkertdev: Guess I'll be updating some stdarch tests again, because that is where these were from.

I did keep the smaller vector sizes for now because they require an additional target feature, so testing for that seems useful.

RalfJung:
Thanks!

I meant to add more tests, not replace the existing ones. But this is fine too as long as all 3 are covered. The masking seems to be implemented in stdtach so we don't really need to test that here.

Did you confirm these on real HW?

tests/pass/shims/x86/intrinsics-x86-avx512.rs
```diff
@@ -191,6 +192,77 @@ unsafe fn test_avx512vpopcntdq() {
...
+        // Identity of A.
+        let r = _mm512_ternarylogic_epi32::<0b1010_1010>(a, b, c);
+        assert_eq_m512i(r, a);
+
+        // Bitwise or.
+        let r = _mm512_ternarylogic_epi32::<0b1111_1110>(a, b, c);
+        let e = _mm512_set4_epi32(0b111, 0b111, 0b001, 0b101);
+        assert_eq_m512i(r, e);
+
+        // Majority.
```
RalfJung: What is "majority"?

tests/pass/shims/x86/intrinsics-x86-avx512.rs
```diff
@@ -191,6 +192,77 @@ unsafe fn test_avx512vpopcntdq() {
...
+        let e = _mm256_set4_epi32(0b000, 0b010, 0b001, 0b101);
+        assert_eq_m256i(r, e);
+    }
+    test_mm256_ternarylogic_epi32();
+
+    #[target_feature(enable = "avx512f,avx512vl")]
+    unsafe fn test_mm_ternarylogic_epi32() {
+        let a = _mm_setr_epi32(0b100, 0b110, 0b001, 0b101);
+        let b = _mm_setr_epi32(0b010, 0b011, 0b001, 0b101);
+        let c = _mm_setr_epi32(0b001, 0b000, 0b001, 0b101);
```
RalfJung: Since this is a bitwise function, the `0b001` already covers the case of things being the same everywhere.
What's missing is 0-1-1 and 1-0-1, so let's do this (for all 3 tests):
```suggestion
        let a = _mm_setr_epi32(0b100, 0b110, 0b001, 0b101);
        let b = _mm_setr_epi32(0b010, 0b011, 0b001, 0b110);
        let c = _mm_setr_epi32(0b001, 0b000, 0b001, 0b111);
```

tests/pass/shims/x86/intrinsics-x86-avx512.rs
```diff
@@ -191,6 +192,77 @@ unsafe fn test_avx512vpopcntdq() {
...
+
+        // Bitwise or.
+        let r = _mm256_ternarylogic_epi32::<0b1111_1110>(a, b, c);
+        let e = _mm256_set4_epi32(0b111, 0b111, 0b001, 0b101);
+        assert_eq_m256i(r, e);
+
+        // Majority.
+        let r = _mm256_ternarylogic_epi32::<0b1110_1000>(a, b, c);
+        let e = _mm256_set4_epi32(0b000, 0b010, 0b001, 0b101);
+        assert_eq_m256i(r, e);
```
RalfJung, line 239: Can you add a ternary function that's not monotone, like a XOR b XOR c?

(This can replace the "bitwise or" or "majority" if you prefer.)

RalfJung:

@rustbot author

folkertdev:
Is CI flaky here?

```
E: Failed to fetch http://ports.ubuntu.com/ubuntu-ports/dists/noble-security/main/binary-armhf/Packages.xz  File has unexpected size (442940 != 443108). Mirror sync in progress? [IP: 91.189.91.102 80]
   Hashes of expected file:
    - Filesize:443108 [weak]
    - SHA256:94e16562de548af81cd7fe9f2c811beac145a8578215679cf9e859980d78c1db
    - SHA1:0754fc5a9770987305b40d119f88cfaf4530246b [weak]
    - MD5Sum:b0e683c5e432662b34d2c45edc9822ca [weak]
   Release file created at: Tue, 11 Nov 2025 15:57:34 +0000
E: Some index files failed to download. They have been ignored, or old ones used instead.
Error: Process completed with exit code 100.
```

RalfJung:
Yeah the Ubuntu ports mirrors are pretty bad, we already disabled most ports CI -- armv7 has been better so far but also sometimes fails. :/

PR #4679, opened 2025-11-11
Add libc direct tests for (get/set/unset)env functions

eduardorittner:
This PR adds tests for the `getenv`, `setenv` and `unsetenv` functions directly through the libc api. Related to rust-lang/miri#3179

### DIFF
--- a/tests/pass-dep/libc/libc-env.rs
+++ b/tests/pass-dep/libc/libc-env.rs
@@ -1,0 +1,96 @@
+//@ignore-target: windows # No libc
+//@compile-flags: -Zmiri-disable-isolation
+
+use std::ffi::CStr;
+
+fn test_getenv() {
+    let s = unsafe { libc::getenv(b"MIRI_ENV_VAR_TEST\0".as_ptr().cast()) };
+    assert!(!s.is_null());
+    let value = unsafe { CStr::from_ptr(s).to_str().unwrap() };
+    assert_eq!(value, "0");
+
+    // Get a non-existing environment variable
+    let s = unsafe { libc::getenv(b"MIRI_TEST_NONEXISTENT_VAR\0".as_ptr().cast()) };
+    assert!(s.is_null());
+
+    // Empty string should not crash
+    let s = unsafe { libc::getenv(b"\0".as_ptr().cast()) };
+    assert!(s.is_null());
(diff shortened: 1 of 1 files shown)


tests/pass-dep/libc/libc-env.rs
```diff
@@ -0,0 +1,98 @@
+//@ignore-target: windows
```
RalfJung:
```suggestion
//@ignore-target: windows # No libc
```

tests/pass-dep/libc/libc-env.rs
```diff
@@ -0,0 +1,98 @@
...
+
+use std::ffi::CStr;
+
+#[path = "../../utils/mod.rs"]
+mod utils;
+
+fn test_getenv() {
+    // PATH should exist and not be null
+    let s = unsafe { libc::getenv(b"PATH\0".as_ptr().cast()) };
+    assert!(!s.is_null());
```
RalfJung: Please check `MIRI_ENV_VAR_TEST` instead, which is set by the test harness so it should always have the value "0".

tests/pass-dep/libc/libc-env.rs
```diff
@@ -0,0 +1,98 @@
+//@ignore-target: windows
+//@compile-flags: -Zmiri-disable-isolation
+
+use std::ffi::CStr;
+
+#[path = "../../utils/mod.rs"]
+mod utils;
```
RalfJung: This does not seem to be used?

RalfJung (changes requested):
Thanks! I left some minor comments.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4679/751fca02946d501c3ffcdd8e63793546a8d85b31..a1adc38e885b62ecd100e54e94c8e11f2743e9d9)*

tests/pass-dep/libc/libc-env.rs
RalfJung: Please also check that the value is indeed "0".

eduardorittner: Sure! Sorry about that

RalfJung:

@rustbot author

PR #4697, opened 2025-11-16
genmc: Build without LLVM; adjust to API changes

michaliskok:
This PR bumps the GenMC version to 0.15.1:
  - LLVM is no longer required to build the GenMC library. This results to faster build times and likely unblocks rust-lang/miri#4571 
  - Invalid configs do not cause GenMC to panic now. Instead, GenMC returns a list of warnings/errors that Miri can use as it sees fit. (The code currently just aborts.)
  - Miri may now pass debugging information to GenMC so that better error reports are produced. (Currently unused.)
  - GenMC's scheduler returns the reason a given execution was dropped, so no heuristics have to be used in Miri.

r? @RalfJung
(Each commit can be reviewed separately.)

### DIFF
--- a/genmc-sys/build.rs
+++ b/genmc-sys/build.rs
@@ -28,7 +28,7 @@ mod downloading {
     /// The GenMC repository the we get our commit from.
     pub(crate) const GENMC_GITHUB_URL: &str = "https://github.com/MPI-SWS/genmc.git";
     /// The GenMC commit we depend on. It must be available on the specified GenMC repository.
-    pub(crate) const GENMC_COMMIT: &str = "d9527280bb99f1cef64326b1803ffd952e3880df";
+    pub(crate) const GENMC_COMMIT: &str = "aa10ed65117c3291524efc19253b5d443a4602ac";
 
     /// Ensure that a local GenMC repo is present and set to the correct commit.
     /// Return the path of the GenMC repo and whether the checked out commit was changed.


RalfJung:
Looks like CI has some trouble:
```

    /home/runner/work/miri/miri/genmc-sys/genmc-src/src/ExecutionGraph/ExecutionGraph.cpp: In member function 'void ExecutionGraph::validate()':
    /home/runner/work/miri/miri/genmc-sys/genmc-src/src/ExecutionGraph/ExecutionGraph.cpp:496:38: error: 'print' is not a member of 'std'; did you mean 'printf'?
      496 |                                 std::print(std::cerr, "Non-existent RF: {}\n{}\n", rLab->getPos(),
          |                                      ^~~~~
          |                                      printf
```
CI should be using g++ 13.2. Is that not new enough for std::print...?

RalfJung:
Yeah apparently that needs GCC 14.1. I don't know whether there's a way to get that on Github's CI. But even if there is, if not even Ubuntu 24 has a new enough GCC I'm quite concerned that this might just be too recent. New GCC versions tend to spread very slowly and we can't make it too hard for people to build Miri.

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
```diff
@@ -22,13 +22,24 @@ auto MiriGenmcShim::schedule_next(
...
+                return SchedulingResult { ExecutionState::Ok, static_cast<int32_t>(arg) };
+            else if constexpr (std::is_same_v<T, Blocked>)
+                return SchedulingResult { ExecutionState::Blocked, 0 };
+            else if constexpr (std::is_same_v<T, Error>)
+                return SchedulingResult { ExecutionState::Error, 0 };
+            else if constexpr (std::is_same_v<T, Finished>)
+                return SchedulingResult { ExecutionState::Finished, 0 };
+            else
+                static_assert(false, "non-exhaustive visitor!");
+            return SchedulingResult { ExecutionState::Finished, 0 };
```
RalfJung: This last line here shouldn't be reachable, right?

michaliskok:
> Yeah apparently that needs GCC 14.1. I don't know whether there's a way to get that on Github's CI. But even if there is, if not even Ubuntu 24 has a new enough GCC I'm quite concerned that this might just be too recent. New GCC versions tend to spread very slowly and we can't make it too hard for people to build Miri.

I'm a bit surprised that cmake configure didn't fail. But anyway I'll ensure it builds with 13.2.

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
michaliskok: Right (IIRC it was necessary for type checking). Will add a comment.

RalfJung:
```suggestion
            static_assert(false, "non-exhaustive visitor!");
```
Does this work for type-checking? EDIT: No it does not.

RalfJung:
```suggestion
            else
                static_assert(false, "non-exhaustive visitor!");
```
Or this builds for me locally.

genmc-sys/cpp/src/MiriInterface/EventHandling.cpp
```diff
@@ -255,7 +259,7 @@ auto MiriGenmcShim::handle_malloc(ThreadId thread_id, uint64_t size, uint64_t al
 
 auto MiriGenmcShim::handle_free(ThreadId thread_id, uint64_t address) -> bool {
     const auto pos = inc_pos(thread_id);
-    GenMCDriver::handleFree(pos, SAddr(address), EventDeps());
+    GenMCDriver::handleFree(nullptr, pos, SAddr(address), EventDeps());
```
RalfJung, line 262: `handleFree` seems to return a `std::optional<VerificationError>` now, does that mean we can resolve the FIXME below and remove this hack?

RalfJung: Ah never mind, that was already the return type before, it just maybe didn't always return an error when it should have or so?

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
michaliskok: I would swear I had tried that... But yeah seems to be working, thanks!

genmc-sys/cpp/src/MiriInterface/EventHandling.cpp
michaliskok: I guess an appropriate error type could have existed in cxx though (irrespective of what GenMC returns). How about I fix this in a separate PR, along with a couple of other open issues?

RalfJung: IIRC the problem was that `handleFree` didn't actually return an error, so we had to call `getResult` to check whether there was an error.

Yeah, let's move that to a future PR.

RalfJung:
Looks great, thanks!

PR #4700, opened 2025-11-17
./miri toolchain: also install rust-analyzer

RalfJung:
This is required for working editor integration if the editor extension does not ship an RA binary itself.

The component is around 8.9 MiB in size, adding around 2.5% to the overall download size. So the question is, do we want to add this for everyone even though most people won't need it, or do we instead provide some other way for people to add more components to the miri toolchain? @rust-lang/miri what do you think?

### DIFF
--- a/miri-script/src/commands.rs
+++ b/miri-script/src/commands.rs
@@ -153,7 +153,7 @@ impl Command {
         // Install and setup new toolchain.
         cmd!(sh, "rustup toolchain uninstall miri").run()?;
 
-        cmd!(sh, "rustup-toolchain-install-master -n miri -c cargo -c rust-src -c rustc-dev -c llvm-tools -c rustfmt -c clippy {flags...} -- {new_commit}")
+        cmd!(sh, "rustup-toolchain-install-master -n miri -c cargo -c rust-src -c rustc-dev -c llvm-tools -c rustfmt -c clippy -c rust-analyzer {flags...} -- {new_commit}")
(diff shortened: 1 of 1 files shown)


RalfJung:
@oli-obk @saethlin any preferences between these two options?
- always install RA component, increases download size by 2.5%
- add a mechanism for people to configure extra components that should be installed into the miri toolchain

oli-obk:
The latter is probably generally useful.

Installing r-a will cater to the majority of ppl tho afaict, as only jetbrains users won't use it

bjorn3:
Vscode and Zed uses wouldn't use the rustup component either most certainly. Both ship/download the rust-analyzer executable independently from rustup unless you explicitly reconfigure them to use a specific executable.

RalfJung:
> The latter is probably generally useful.

Yeah, but it means we need a config file.

OTOH maybe we want a config file anyway to configure things like features so I don't have to pass `--feature=genmc` all the time when working on genmc.^^

oli-obk:
Ok with that knowledge I think a miri.toml is probably best, yea

saethlin:
No objection, but as someone who doesn't even use r-a I don't really have anything to add. I've learned a few things from reading these comments ^^

RalfJung:
> Ok with that knowledge I think a miri.toml is probably best, yea

I finally got around to implementing that: https://github.com/rust-lang/miri/pull/5289

PR #4701, opened 2025-11-17
Enable MacOS support for GenMC mode

RalfJung:


RalfJung:
@michaliskok seems to work on macos now. :)

However we do get some warnings during the build:
```
  warning: genmc-sys@0.1.0: In file included from /Users/runner/work/miri/miri/target/debug/build/genmc-sys-901a62d9deac67c4/out/cxxbridge/sources/genmc-sys/src/lib.rs.cc:1:
  warning: genmc-sys@0.1.0: In file included from ./cpp/include/MiriInterface.hpp:14:
  warning: genmc-sys@0.1.0: /Users/runner/work/miri/miri/target/debug/build/genmc-sys-901a62d9deac67c4/out/genmc/include/genmc/ExecutionGraph/EventLabel.hpp:26:10: warning: non-portable path to file '"runtime/InterpreterEnumAPI.hpp"'; specified path differs in case from file name on disk [-Wnonportable-include-path]
  warning: genmc-sys@0.1.0:    26 | #include "Runtime/InterpreterEnumAPI.hpp"
  warning: genmc-sys@0.1.0:       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  warning: genmc-sys@0.1.0:       |          "runtime/InterpreterEnumAPI.hpp"
```

IIRC macos has a case-insensitive file system and somehow that is unhappy here?

RalfJung:
Maybe cmake is renaming things to lower-case when copying them to the build dir? I don't know why it makes a copy TBH but it seems header files get copied for some reason. I found this in the cmake docs

> On Windows and macOS, globbing is case-insensitive even if the underlying filesystem is case-sensitive (both filenames and globbing expressions are converted to lowercase before matching). On other platforms, globbing is case-sensitive.

That's about globbing but maybe it applies more globally.

I am inclined to just ignore this, the build works fine and I'm not going to try to debug this remotely on the GHA machines.^^

RalfJung:
Hm, we also seem stuck in some infinite loop in CI... I guess macos support has to wait until someone with a mac is willing to figure out what is going on.

RalfJung:
Fun, when I run this locally I get
```
error: Undefined Behavior: Attempt to read from uninitialized memory
  --> /home/r/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/sys/pal/unix/sync/mutex.rs:113:26
   |
LL |         let r = unsafe { libc::pthread_mutex_destroy(self.raw()) };
   |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
   |
```
This is still running Miri's versions of the pthread shims, of course... but I'm not sure why it causes that sort of error.

RalfJung:
For now I think let's just stick to Linux only, where we can actually test locally what happens. If someone with a Mac wants to help get genmc to work there, please let us know.

PR #4705, opened 2025-11-17
Add avx512 permute and pmaddbw

folkertdev:
Turns out we need a couple more, but the implementation can be shared with existing versions.

### DIFF
--- a/src/shims/x86/mod.rs
+++ b/src/shims/x86/mod.rs
@@ -1132,6 +1132,44 @@ fn pmaddbw<'tcx>(
     interp_ok(())
 }
 
+/// Shuffle 32-bit integers in `values` across lanes using the corresponding
+/// index in `indices`, and store the results in dst.
+///
+/// <https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_permutevar8x32_epi32>
+/// <https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_permutevar8x32_ps>
+/// <https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_permutexvar_epi32>
+fn permute<'tcx>(
+    ecx: &mut crate::MiriInterpCx<'tcx>,
+    values: &OpTy<'tcx>,
+    indices: &OpTy<'tcx>,
+    dest: &MPlaceTy<'tcx>,
+) -> InterpResult<'tcx, ()> {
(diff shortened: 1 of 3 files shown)


src/shims/x86/mod.rs
```diff
@@ -1086,6 +1086,89 @@ fn psadbw<'tcx>(
...
+
+    // fn permd(a: u32x8, b: u32x8) -> u32x8;
+    // fn permps(a: __m256, b: i32x8) -> __m256;
+    // fn vpermd(a: i32x16, idx: i32x16) -> i32x16;
+    assert_eq!(dest_len, values_len);
+    assert_eq!(dest_len, indices_len);
+
+    // Only use the lower 3 bits to index into a vector with 8 lanes,
+    // or the lower 4 bits when indexing into a 16-lane vector.
+    let mask = u32::try_from(dest_len).unwrap().strict_sub(1);
```
RalfJung, line 1160: Please add an assertion that dest_len is a power of 2.

tests/pass/shims/x86/intrinsics-x86-avx512.rs
```diff
@@ -55,6 +55,93 @@ unsafe fn test_avx512() {
         assert_eq_m512i(r, e);
     }
     test_mm512_sad_epu8();
+
+    #[target_feature(enable = "avx512bw")]
+    unsafe fn test_mm512_maddubs_epi16() {
+        #[rustfmt::skip]
```
RalfJung, line 62:
```suggestion
        // `a` is interpreted as `u8x16`, but `_mm512_set_epi8` expects `i8`, so we have to cast.
        #[rustfmt::skip]
```

folkertdev:
> I can't say I can make sense of this maddubs test (it looks like lots of random numbers), but the code is the same as before so 🤷

The numbers attempt to exercise some of the interesting behavior with the saturation. But yeah it's tough.

> I assume you ran the tests on real hardware?

Yes

@rustbot ready

folkertdev:
Oh btw has your opinion on force pushes changed now that we have the improved diffs? Or should I just squash all changes into one commit anyway?

RalfJung:
Yeah I guess we can be more relaxed about force-pushes now, though that does require some discipline to always submit a comment with the review so that we get the diff links.

This one is ready to have the fixup commit squashed in whatever way you think makes sense (the one-commit `./miri squash` automation is just there to make it easier for people). Thanks!

@rustbot author

folkertdev:
I absorbed the changes into the earlier commits

@rustbot ready

PR #4707, opened 2025-11-17
Support retagging of wildcard references in tree borrows

royAmmerschuber:
Adds proper suport for reborrowing wildcard pointers to tree borrows.


src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1169,95 +1161,88 @@ impl<'tcx> Tree {
...
+                        let node = args.nodes.get(args.idx).unwrap();
+                        let perm = args.loc.perms.get(args.idx);
+                        let wildcard_state =
+                            args.loc.wildcard_accesses.get(args.idx).cloned().unwrap_or_default();
+
+                        let old_state =
+                            perm.copied().unwrap_or_else(|| node.default_location_state());
+                        // If we know where, relative to this node, the wildcard access occurs,
+                        // then check if we can skip the entire subtree because the access might not
+                        // change any permissions here anyway.
```
RalfJung: I'm confused by the "might not" here. (Isn't that something you already had before and we removed it in the previous PR?)

royAmmerschuber: Yes, your right. This got lost in a rebase.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -847,73 +874,29 @@ impl<'tcx> Tree {
...
-                        access_kind: AccessKind,
-                        access_cause: diagnostics::AccessCause,
-                        args: NodeAppArgs<'_>|
-         -> Result<(), _> {
-            let node = args.nodes.get_mut(args.idx).unwrap();
-            let mut perm = args.loc.perms.entry(args.idx);
-
-            let state = perm.or_insert(node.default_location_state());
+        #[cfg(feature = "expensive-consistency-checks")]
+        self.verify_wildcard_consistency(global);
```
RalfJung: We should only need this if a wildcard is involved in the access, right? I.e. if the tag is wildcard, or if there are multiple roots.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -294,8 +295,9 @@ pub struct Tree {
     pub(super) nodes: UniValMap<Node>,
     /// Associates with each location its state and wildcard access tracking.
     pub(super) locations: DedupRangeMap<LocationTree>,
-    /// The index of the root node.
-    pub(super) root: UniIndex,
+    /// Contains both the root of the main tree as well as the roots of the wildcard subtrees.
+    /// Sorted according to `BorTag` from low to high. This also means the main root is `root[0]`.
+    pub(super) roots: SmallVec<[UniIndex; 2]>,
```
RalfJung, line 312: This seems like a good place to describe what these multiple trees even mean.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -657,17 +665,23 @@ impl<'tcx> Tree {
...
-        let parent_node = self.nodes.get_mut(parent_idx).unwrap();
-        // Register new_tag as a child of parent_tag
-        parent_node.children.push(idx);
+        if let Some(parent_idx) = parent_idx {
+            let parent_node = self.nodes.get_mut(parent_idx).unwrap();
+            // Register new_tag as a child of parent_tag
+            parent_node.children.push(idx);
+        } else {
+            // If the parent had wildcard provenance, then register the idx
+            // as a new wildcard root.
```
RalfJung, line 720: Please explain why this preserves sortedness.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -692,18 +706,24 @@ impl<'tcx> Tree {
         // We need to ensure the consistency of the wildcard access tracking data structure.
         // For this, we insert the correct entry for this tag based on its parent, if it exists.
```
RalfJung, line 747: The comment does not cover the else branch.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1095,117 +1093,331 @@ impl Tree {
...
+        // We use this to further narrow access_relatedness on the wildcard accesses treating
+        // all tags larger than this as only_foreign.
+        //
+        // If we accessed the main tree then its root will be the smallest tag in the entire
+        // allocation, meaning we perform only a foreign access on all other subtrees.
+        //
+        // Foreign accesses are always possible on wildcard subtrees, as they have
+        // `max_foreign_access==Write` on their root. However on the main tree this can fail
+        // resulting in an access error.
+        let accessed_root_tag = accessed_root.map(|idx| nodes.get(idx).unwrap().tag);
```
RalfJung, line 1187: This doesn't seem to directly relate to the comment. You can just move it up.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1095,117 +1093,331 @@ impl Tree {
...
+            }
+            // We don't perform a wildcard access on the tree we already performed a
+            // normal access on.
+            if Some(root) == accessed_root {
+                continue;
+            }
+            self.perform_wildcard_access(
+                root,
+                access_source,
+                /*max_local_tag*/ accessed_root_tag,
```
RalfJung, line 1210: This needs more explanation:

We can only be a local access for nodes that are a *parent* of us. We could pass `access_source` here, but using `accessed_root` is better since that will be smaller, and it is still a node that must be on all parent paths starting at the accessed node.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1095,117 +1093,331 @@ impl Tree {
...
+        loc_range: Range<u64>,
+        access_range: Option<AllocRange>,
+        access_kind: AccessKind,
+        access_cause: diagnostics::AccessCause,
+        global: &GlobalState,
+        alloc_id: AllocId, // diagnostics
+        span: Span,        // diagnostics
+        visit_children: ChildrenVisitMode,
+    ) -> InterpResult<'tcx, UniIndex> {
+        let root = OnceCell::new();
```
RalfJung: Please add a comment.

Also, do we really need interior mutability?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1095,117 +1093,331 @@ impl Tree {
...
+        // exposed children then some of the wildcard subtrees could also be children of
+        // the accessed node and would also need to be skipped. We can narrow down which
+        // child trees are children by comparing their root tag to the minimum exposed
+        // child of the accessed node. As the parent tag is always smaller than the child
+        // tag this means we only need to skip subtrees with a root tag larger than
+        // `min_exposed_child`.
+        let min_exposed_child = match visit_children {
+            ChildrenVisitMode::SkipChildrenOfAccessed =>
+                Self::get_min_exposed_child(access_source.unwrap(), nodes),
+            _ => None,
```
RalfJung: If this is just a clever optimization, let's do that in a later PR.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1095,117 +1093,331 @@ impl Tree {
...
+        interp_ok(root.into_inner().unwrap())
+    }
+
+    /// Performs a wildcard access on the tree with root `root`. Takes the `access_relatedness`
+    /// for each node from the `WildcardState` datastructure.
+    /// * `root`: Root of the tree being accessed.
+    /// * `access_source`: the index of the accessed tag, if any.
+    ///   This is only used for printing the correct tag on errors.
+    /// * `max_local_tag`: For any tag larger than this we do not perform local accesses. If no
+    ///   foreign access to a node with a larger tag is possible then we return an access error.
```
RalfJung: Is there a test that relies on this logic?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1095,117 +1093,331 @@ impl Tree {
...
+                root,
+                |args: &NodeAppArgs<'_>| -> ContinueTraversal {
+                    let node = args.nodes.get(args.idx).unwrap();
+                    let perm = args.loc.perms.get(args.idx);
+                    let wildcard_state =
+                        args.loc.wildcard_accesses.get(args.idx).cloned().unwrap_or_default();
+
+                    let only_foreign = max_local_tag
+                        .map(|max_local_tag| max_local_tag < node.tag)
+                        .unwrap_or(false);
```
```suggestion
                    let only_foreign = max_local_tag
                        .is_some_and(|max_local_tag| max_local_tag < node.tag);
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1095,117 +1093,331 @@ impl Tree {
...
+                    let old_state = perm.copied().unwrap_or_else(|| node.default_location_state());
+                    // If we know where, relative to this node, the wildcard access occurs,
+                    // then check if we can skip the entire subtree.
+                    if let Some(relatedness) =
+                        wildcard_state.access_relatedness(access_kind, only_foreign)
```
RalfJung: Can we avoid duplicating the access_relatedness code by moving it into a local closure?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1095,117 +1093,331 @@ impl Tree {
...
+                    // The `TreeVisitor` doesn't call `f_continue` on the `start_idx`
+                    // so we check here if we need to skip it.
+                    if args.idx == root
+                        && matches!(
+                            perm.skip_if_known_noop(access_kind, relatedness),
+                            ContinueTraversal::SkipSelfAndChildren
+                        )
+                    {
+                        return Ok(());
+                    }
```
RalfJung: Is it possible to just call f_continue on the root instead?

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -131,6 +147,12 @@ impl WildcardState {
             ..Default::default()
         }
     }
+    /// Crates the initial `WildcardState` for a wildcard root.
+    /// This has `max_foreign_access==Write` as any access that isn't from a child of
+    /// this tag is foreign.
```
RalfJung:
```suggestion
    /// This has `max_foreign_access==Write` as it actually is the child of *some* exposed node
    /// from which we can receive foreign accesses.
```

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -88,10 +88,26 @@ impl WildcardState {
...
-        match kind {
+    /// If `only_foreign` is true then we treat `LocalAccess` as impossible. This means we return
+    /// `None` if only a `LocalAccess` is possible, and we treat `EitherAccess` as a
+    /// `ForeignAccess`.
+    pub fn access_relatedness(
+        &self,
+        kind: AccessKind,
+        only_foreign: bool,
+    ) -> Option<WildcardAccessRelatedness> {
+        use WildcardAccessRelatedness as E;
```
RalfJung: Please move this inside the `if`.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1219,10 +1431,10 @@ impl Node {
 
 impl VisitProvenance for Tree {
     fn visit_provenance(&self, visit: &mut VisitWith<'_>) {
-        // To ensure that the root never gets removed, we visit it
-        // (the `root` node of `Tree` is not an `Option<_>`)
-        visit(None, Some(self.nodes.get(self.root).unwrap().tag));
-
+        // To ensure that the roots never get removed, we visit them.
```
RalfJung, line 1436:
```suggestion
        // To ensure that the roots never get removed, we visit them.
        // FIXME: it should be possible to GC wildcard tree roots.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -775,48 +795,55 @@ impl<'tcx> Tree {
...
+            // arbitrary one.
+            if let Some(start_idx) = start_idx {
+                TreeVisitor { nodes: &mut self.nodes, loc }.traverse_this_parents_children_other(
+                    start_idx,
+                    // Visit all children, skipping none.
+                    |_| ContinueTraversal::Recurse,
+                    |args| check_strong_protector(args.idx, args.nodes, &args.loc.perms),
+                )?;
+            }
+            // Afterward we check all tags in arbitrary order, so that we also catch
```
```suggestion
            // Afterwards we check all tags in arbitrary order, so that we also catch
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -775,48 +795,55 @@ impl<'tcx> Tree {
...
+            // protectors on different subtrees.
+            // (This unnecessarily checks the tags of `start_idx`s subtree again)
+            for (idx, _) in self.nodes.iter() {
+                check_strong_protector(idx, &self.nodes, &loc.perms)?;
+            }
```
RalfJung, line 897: We should definitely not to this when there's only one tree.

Maybe it's better to just traverse the other trees as well, so that we conceptually remain on the tree level.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -241,16 +241,12 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         };
 
         log_creation(this, Some((alloc_id, base_offset, parent_prov)))?;
-
-        let orig_tag = match parent_prov {
-            ProvenanceExtra::Wildcard => return interp_ok(place.ptr().provenance), // TODO: handle retagging wildcard pointers
-            ProvenanceExtra::Concrete(tag) => tag,
-        };
+        let new_prov = Provenance::Concrete { alloc_id, tag: new_tag };
```
RalfJung: Maybe move this to just after the previous `let` where we get the alloc_id? Grouping with `log` doesn't make that much sense.

src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -561,6 +563,11 @@ impl DisplayFmt {
             })
             .unwrap_or("")
     }
+
+    /// Print extra text if the tag is exposed.
+    fn print_exposed(&self, exposed: bool) -> S {
+        if exposed { " Exposed" } else { "" }
```
RalfJung:
```suggestion
        if exposed { " (exposed)" } else { "" }
```

tests/pass/tree_borrows/wildcard/reborrow.rs
```diff
@@ -0,0 +1,167 @@
...
+    //   │ ref1(Res)* │     │ ref2(Res)* │    │  reb(Res)  │
+    //   │            │     │            │    │            │
+    //   └────────────┘     └────────────┘    └────────────┘
+
+    // Could either have as a parent ref1 or ref2.
+    // So we can't disable either of them.
+    *reb = 13;
+
+    // We can still access both ref1, ref2.
+    assert_eq!(*ref2, 13);
```
RalfJung: Comment and code don't quite match... we can access both, after all.
You can just explain that this is technically UB we don't catch.

tests/pass/tree_borrows/wildcard/reborrow.rs
```diff
@@ -0,0 +1,167 @@
...
+    //                      │            │     │            │
+    //                      └────────────┘     └────────────┘
+
+    // Writes either through ref1, ref2 or ptr_base, which are all exposed.
+    // Since we don't know which we do not apply any transitions to any of
+    // the references.
+    unsafe { wild.write(13) };
+
+    // We should be able to access any of the references.
+    assert_eq!(*ref2, 13);
```
RalfJung: Again this doesn't quite test what the comment says.

tests/pass/tree_borrows/wildcard/reborrow.rs
```diff
@@ -0,0 +1,167 @@
...
+    //                             │
+    //                             │
+    //                             ▼
+    //                      ┌────────────┐
+    //                      │            │
+    //                      │ ref1(Res)* │
+    //                      │            │
+    //                      └────────────┘
+
+    // This is the only valid ordering these accesses can happen in.
```
RalfJung, line 144: This should refer back to the comment for the test: accessing reb2 doesn't invalidate reb1 because ref1 exists. (Also, please pick more distinguishable names.)

tests/fail/tree_borrows/wildcard/cross_tree_update_main_invalid_exposed.rs
RalfJung, line 1: Please explain in the comment that this tests the max_local_tag logic.

RalfJung (changes requested):
Overall approach looks good. Here's a bunch of comments. :)

Please also add the 3 SB tests in this area: `illegal_read/write_despite_exposed*`.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4707/5fbfc88282ba0db4a3f22bfdaa9c5556f4019eb9..52f2087f1722ba249ac92ccf0cdab0feb4b6da76)*

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber: reborrow5/8 rely on this

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -775,48 +811,72 @@ impl<'tcx> Tree {
...
+                            |_| ContinueTraversal::Recurse,
+                            check_strong_protector,
+                        )?,
+                )
+            } else {
+                None
+            };
+            // Afterwards we check all tags in arbitrary order, so that we also catch
+            // protectors on different subtrees.
+            // (This unnecessarily checks the tags of `start_idx`s subtree again)
```
RalfJung: This comment seems outdated now.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -775,48 +811,72 @@ impl<'tcx> Tree {
...
+            for root in self.roots.iter() {
+                if Some(*root) == accessed_root {
+                    continue;
+                }
+                TreeVisitor { nodes: &mut self.nodes, loc }.traverse_this_parents_children_other(
+                    *root,
+                    // Visit all children, skipping none.
+                    |_| ContinueTraversal::Recurse,
+                    check_strong_protector,
+                )?;
```
RalfJung: This part is still duplicated between the two branches... maybe it makes more sense to share the entire visitor invocation, not just the `check_strong_protector` closure?

src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -607,23 +614,21 @@ fn char_repeat(c: char, n: usize) -> String {
...
+        extraction_aux(tree, root, show_unnamed, &mut v);
         let Some(root) = v.pop() else {
             if show_unnamed {
                 unreachable!(
                     "This allocation contains no tags, not even a root. This should not happen."
                 );
             }
-            eprintln!(
-                "This allocation does not contain named tags. Use `miri_print_borrow_state(_, true)` to also print unnamed tags."
-            );
```
RalfJung, line 626: Why did this move around?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -294,8 +294,22 @@ pub struct Tree {
...
+    /// with this new reference as a root. We call this additional tree a wildcard subtree.
+    ///
+    /// Sorted according to `BorTag` from low to high. This also means the main root is `root[0]`.
+    ///
+    /// The actual structure should be a single tree but with wildcard provenance we approximate
+    /// this with this ordered set of trees. Each wildcard subtree could be the direct child of
+    /// any exposed tag (that is smaller than the root). This also means that it can only be the
+    /// child of a tree that comes before it in the vec ensuring we don't have any cycles in our
+    /// approximated tree.
+    ///
```
RalfJung, line 308:
```suggestion
    /// If we reborrow a reference which has wildcard provenance, then we do not know where in
    /// the tree to attach them. Instead we create a new additional tree for this allocation
    /// with this new reference as a root. We call this additional tree a wildcard subtree.
    ///
    /// The actual structure should be a single tree but with wildcard provenance we approximate
    /// this with this ordered set of trees. Each wildcard subtree is the direct child of *some* exposed
    /// tag (that is smaller than the root), but we do not know which. This also means that it can only be the
    /// child of a tree that comes before it in the vec ensuring we don't have any cycles in our
    /// approximated tree.
    ///
    /// Sorted according to `BorTag` from low to high. This also means the main root is `root[0]`.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -410,7 +425,7 @@ where
         this: &mut TreeVisitor<'_>,
         accessed_node: UniIndex,
         visit_children: ChildrenVisitMode,
-    ) -> Result<(), Err> {
+    ) -> Result<UniIndex, Err> {
```
RalfJung, line 429: Please explain the return value in a doc comment for this function.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -639,15 +680,20 @@ impl<'tcx> Tree {
...
+        parent_prov: ProvenanceExtra,
         new_tag: BorTag,
         inside_perms: DedupRangeMap<LocationState>,
         outside_perm: Permission,
         protected: bool,
         span: Span,
     ) -> InterpResult<'tcx> {
         let idx = self.tag_mapping.insert(new_tag);
-        let parent_idx = self.tag_mapping.get(&parent_tag).unwrap();
+
```
```suggestion
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -772,52 +834,67 @@ impl<'tcx> Tree {
...
+            // arbitrary one.
+            let accessed_root = start_idx.map(&mut check_tree).transpose()?;
+            // Afterwards we check all other trees.
+            // We iterate over the list in reverse order to ensure that we do not visit
+            // a parent before its child.
+            for root in self.roots.iter().rev() {
+                if Some(*root) == accessed_root {
+                    continue;
+                }
+                check_tree(*root)?;
```
```suggestion
            for &root in self.roots.iter().rev() {
                if Some(root) == accessed_root {
                    continue;
                }
                check_tree(root)?;
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -920,7 +997,9 @@ impl<'tcx> Tree {
 /// Integration with the BorTag garbage collector
 impl Tree {
     pub fn remove_unreachable_tags(&mut self, live_tags: &FxHashSet<BorTag>) {
-        self.remove_useless_children(self.root, live_tags);
+        for root in self.roots.clone().iter() {
```
RalfJung: Cloning the roots list isn't great -- do we have to?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1097,13 +1176,40 @@ impl<'tcx> LocationTree {
                 alloc_id,
                 span,
                 visit_children,
-            )
+            )?)
         } else {
             // `SkipChildrenOfAccessed` only gets set on protector release.
             // Since a wildcard reference are never protected this assert shouldn't fail.
```
```suggestion
            // `SkipChildrenOfAccessed` only gets set on protector release, which only
            // occurs on a known node.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1097,13 +1176,40 @@ impl<'tcx> LocationTree {
...
+            if Some(root) == accessed_root {
+                continue;
+            }
+            // This can only be a local access for nodes that are a parent of the accessed node.
+            // We could pass access_source as `max_local_tag`, but using accessed_root is better
+            // since that will be smaller, and it is still a node that must be
+            // on all parent paths starting at the accessed node.
+            //
+            // If we accessed the main tree then its root will be the smallest tag in the entire
+            // allocation, meaning we perform only a foreign access on all other subtrees.
```
```suggestion
            // The choice of `max_local_tag` requires some thought.
            // This can only be a local access for nodes that are a parent of the accessed node
            // and are therefore smaller, so the accessed node itself is a valid choice for `max_local_tag`.
            // However, using `accessed_root` is better since that will be smaller. It is still a valid choice
            // because for nodes *in other trees*, if they are a parent of the accessed node then they
            // are a parent of `accessed_root`.
            //
            // If we accessed the main tree then its root will be the smallest tag in the entire
            // allocation, meaning we perform only a foreign access on all other subtrees.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1097,13 +1176,40 @@ impl<'tcx> LocationTree {
...
+            //
+            // Foreign accesses are always possible on wildcard subtrees, as they have
+            // `max_foreign_access==Write` on their root. However on the main tree this can fail
+            // resulting in an access error.
```
RalfJung: I don't understand this last paragraph. What is it referring to?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1206,95 +1323,101 @@ impl<'tcx> LocationTree {
...
-            |idx: UniIndex, nodes: &UniValMap<Node>, loc: &LocationTree| -> ContinueTraversal {
-                let node = nodes.get(idx).unwrap();
-                let perm = loc.perms.get(idx);
-                let wildcard_state = loc.wildcard_accesses.get(idx).cloned().unwrap_or_default();
+        let get_relatedness = |idx: UniIndex, node: &Node, loc: &LocationTree| {
+            let wildcard_state = loc.wildcard_accesses.get(idx).cloned().unwrap_or_default();
+            // We can use the invariant that child tags are larger than parent tags
+            // to further refine the relatedness.
+            // If we know that the access comes through a certain tag, then we know
+            // for any larger tag that the access cannot be local.
```
RalfJung: This comment is a bit odd since it seems to refer to the reasoning inside `Tree::perform_access`. It should instead refer to what is written in the doc comment of the current function.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1206,95 +1323,101 @@ impl<'tcx> LocationTree {
...
+            let only_foreign = max_local_tag.is_some_and(|max_local_tag| max_local_tag < node.tag);
+            wildcard_state.access_relatedness(access_kind, only_foreign)
+        };
+
+        // This does a traversal across the tree updating children before their parents. The
+        // difference to `perform_normal_access` is that we take the access relatedness from
+        // the wildcard tracking state of the node instead of from the visitor itself.
+        //
+        // Different from a normal access the iteration order is important for correctness.
```
```suggestion
        // Unlike for a normal access, the iteration order is important for correctness.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1206,95 +1323,101 @@ impl<'tcx> LocationTree {
...
+        // Specifically if `max_local_tag` is Some. A local access cannot come from a child
+        // with a tag larger than `max_local_tag` however in general a parent cannot know if
+        // a local access is impossible because all its exposed children have a too large tag.
+        // However, in the case were our exposed children transition to a lower permission
+        // during this access, since we update the children before their parents this means
+        // that when we are updating the parent the child counts have already been updated to
+        // reflect the new possibly lower access level. This is still only an approximation of
+        // the correct access level.
```
RalfJung: I don't understand this comment. Why is it relevant that "a parent cannot know if a local access is impossible because all its exposed children have a too large tag"? That sounds like it would make skipping harder, sure, but why is it relevant for correctness?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1206,95 +1323,101 @@ impl<'tcx> LocationTree {
...
-                        .wildcard_accesses
-                        .get(args.idx)
-                        .and_then(|s| s.access_relatedness(access_kind))
-                    else {
-                        // There doesn't exist a valid exposed reference for this access to
-                        // happen through.
-                        // If this fails for one id, then it fails for all ids so this.
-                        // Since we always check the root first, this means it should always
-                        // fail on the root.
-                        assert_eq!(root, args.idx);
```
RalfJung, line 1254: Does this assertion not work any more?

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -435,6 +460,10 @@ impl Tree {
     /// Checks that the wildcard tracking data structure is internally consistent and
     /// has the correct `exposed_as` values.
     pub fn verify_wildcard_consistency(&self, global: &GlobalState) {
+        // We rely on the fact that wildcard_roots is ordered according to tag from low to high.
```
RalfJung:
```suggestion
        // We rely on the fact that `roots` is ordered according to tag from low to high.
```

tests/fail/both_borrows/illegal_read_despite_exposed2.rs
```diff
@@ -7,6 +9,9 @@ fn main() {
         let exposed_ptr = addr as *mut i32;
         // From the exposed ptr, we get a new unique ptr.
         let root2 = &mut *exposed_ptr;
+        // Tree borrows doesn't activate its mut references before the first write.
+        // So tree borrow would let this pass if we removed this line.
```
RalfJung:
```suggestion
        // Activate the reference (unnecessary on Stacked Borrows).
```

tests/fail/tree_borrows/wildcard/cross_tree_from_main.rs
```diff
@@ -0,0 +1,36 @@
+//@compile-flags: -Zmiri-tree-borrows -Zmiri-permissive-provenance
+
+/// Checks how accesses from one subtree effect other subtrees.
```
RalfJung:
```suggestion
/// Checks how accesses from one subtree affect other subtrees.
```
This typo occurs quite a lot, I won't annotate it everywhere. :)

tests/pass/tree_borrows/wildcard/reborrow.rs
```diff
@@ -0,0 +1,222 @@
...
+    //       ┌────────────┐
+    //       │            │
+    //       │ ref3(Res)* │
+    //       │            │
+    //       └────────────┘
+
+    // Performs a foreign read on ref3 and either read on reb1.
+    // This temporarily treats ref3 as only foreign.
+    let _x = *reb2;
+    // Performs an either write on ref3 and reb1.
```
RalfJung: What is an "either write"?

tests/pass/tree_borrows/wildcard/reborrow.rs
```diff
@@ -0,0 +1,222 @@
...
+    // This should stop treating ref3 as only foreign.
```
RalfJung: Why?

tests/pass/tree_borrows/wildcard/undetected_ub.rs
```diff
@@ -101,45 +101,57 @@ pub fn protected_exposed() {
...
-    // ref2 is disabled, so this read causes UB, but we currently don't protect this.
+    // ref2 should be disabled, so this read causes UB, but we currently don't detect this.
     let _fail = *ref2;
 }
 
-/// Currently, we do not assign protectors to wildcard references.
-/// This test has UB because it does a foreign write to a protected reference.
-/// However, that reference is a wildcard, so this doesn't get detected.
-#[allow(unused_variables)]
-pub fn protected_wildcard() {
```
RalfJung, line 112: The test you removed -- I presume we now detect the UB there? Is that UB covered by one of the tests you added?

tests/pass/tree_borrows/wildcard/undetected_ub.rs
```diff
@@ -101,45 +101,57 @@ pub fn protected_exposed() {
...
+    //                      ┌────────────┐
+    //                      │            │
+    //                      │ ref3(Res)* │
+    //                      │            │
+    //                      └────────────┘
+
+    // This access doesn't freeze reb1 even though no access could have come from its
+    // child ref3 (since ref3>reb2).
+    //
+    // ref3 doesn't get frozen, because It's still reserved.
```
RalfJung:
```suggestion
    // ref3 doesn't get frozen because it's still reserved.
```

tests/pass/tree_borrows/wildcard/undetected_ub.rs
```diff
@@ -101,45 +101,57 @@ pub fn protected_exposed() {
...
+    //                             │
+    //                             ▼
+    //                      ┌────────────┐
+    //                      │            │
+    //                      │ ref3(Res)* │
+    //                      │            │
+    //                      └────────────┘
+
+    // This access doesn't freeze reb1 even though no access could have come from its
+    // child ref3 (since ref3>reb2).
```
RalfJung: Why does the max_local_access logic not detect this?

RalfJung (changes requested):
Thanks! Looks good overall but I have plenty of minor comments.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4707/b39f1e797f0906bdc83a829f9ceedb8c213c192b..40ed9a1ac32f546a90a02c5f17ed900440105260)*

src/borrow_tracker/tree_borrows/diagnostics.rs
royAmmerschuber: Since we now have multiple trees, we only want to fail if no trees have named tags

tests/pass/tree_borrows/wildcard/undetected_ub.rs
royAmmerschuber: covered by the test `protected_wildcard.rs`

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber: Did you mean the comment at the beginning of `perform_normal_access`? I copied that one from `Tree::perform_access`. This comment looks fine to me.

RalfJung: It talks about children and parents and accesses "coming through" a certain tag. But all we know is we have max_local_tag and a promise that all larger tags are foreign. We don't know how max_local_tag is picked, that's not our business.

We could change the doc comment to say "max_local_tag is the tag of a parent of the accessed node such that all the nodes in between the two are definitely not in the current tree", or so. But that's not what we document so we shouldn't write a comment assuming anything about max_local_tag being in any relation to the accessed node.

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber: What I was trying to say, with correctness, is that the current implementation would be less correct if we changed the iteration order. Specifically, test `cross_tree_update_older_invalid_exposed2.rs` fails if we update parents before children.
This isn't as good as the *idealized* version of the algorithm, where we recalculate the entire wildcard datastructure before each access, but in the cases where the exposed tags happen to get disabled during this access it behaves the same as the idealized version.

RalfJung: I don't know what "less correct" means. :) Either it is correct or not.  The wildcard logic is *correct* if it never reports false UB. It seems you are talking about how much UB we can find, but when it comes to wildcards, that's not a correctness concern, that's best-effort anyway.

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber: Yes, `cross_tree_update_main_invalid_exposed.rs` fails the assert.
It will now either fail on the root or the first tag it encounters with `tag>max_local_tag`.

royAmmerschuber: I think it's not even necessarily the first. It's just the first one with an exposed child (through which an access could happen).

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -91,11 +91,11 @@ impl LocationState {
...
-        access_cause: AccessCause,
-        access_range: Option<AllocRange>,
+        access_cause: AccessCause,        //diagnostics
+        access_range: Option<AllocRange>, //diagnostics
         relatedness: AccessRelatedness,
-        span: Span,
-        location_range: Range<u64>,
+        span: Span,                 //diagnostics
+        location_range: Range<u64>, //diagnostics
         protected: bool,
```
royAmmerschuber, line 99: It might make sense to move all the diagnostics arguments into a single struct. Since these just get passed through unchanged all the way from `Tree::perform_access`.

RalfJung: That seems like a good idea for a future PR. :)

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -405,12 +420,13 @@ where
         (self.f_propagate)(NodeAppArgs { idx, rel_pos, nodes: this.nodes, loc: this.loc })
     }
 
+    /** Returns the root of this tree. */
```
```suggestion
    /// Returns the root of this tree.
```
Please stick to our usual comment style.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1097,13 +1176,42 @@ impl<'tcx> LocationTree {
...
+            // and are therefore smaller, so the accessed node itself is a valid choice for `max_local_tag`.
+            // However, using `accessed_root` is better since that will be smaller. It is still a valid choice
+            // because for nodes *in other trees*, if they are a parent of the accessed node then they
+            // are a parent of `accessed_root`.
+            //
+            // If we accessed the main tree then its root will be the smallest tag in the entire
+            // allocation, meaning we perform only a foreign access on all other subtrees.
+            //
+            // If `root` is the main root and it doesn't have any valid exposed children then this access
+            // will return an access error. This error doesn't get triggered on wildcard subtrees because we
```
RalfJung: Is there a test checking for this kind of error?

Not super sure why the comment bothers to spell this out -- this is just an instance of the usual general principles we implement, right?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1191,110 +1303,125 @@ impl<'tcx> LocationTree {
...
         .into()
     }
+
     /// Performs a wildcard access on the tree with root `root`. Takes the `access_relatedness`
     /// for each node from the `WildcardState` datastructure.
     /// * `root`: Root of the tree being accessed.
+    /// * `access_source`: the index of the accessed tag, if any.
+    ///   This is only used for printing the correct tag on errors.
+    /// * `max_local_tag`: The access cannot come through any node in this tree that has a larger
+    ///   tag than this tag.
```
RalfJung: Not sure why you changed this, but I find it more confusing than before now.
```suggestion
    /// * `max_local_tag`: The access can only be local for nodes whose tag is at most `max_local_tag`.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1191,110 +1303,125 @@ impl<'tcx> LocationTree {
...
         span: Span,        // diagnostics
     ) -> InterpResult<'tcx> {
-        let f_continue =
-            |idx: UniIndex, nodes: &UniValMap<Node>, loc: &LocationTree| -> ContinueTraversal {
-                let node = nodes.get(idx).unwrap();
-                let perm = loc.perms.get(idx);
-                let wildcard_state = loc.wildcard_accesses.get(idx).cloned().unwrap_or_default();
+        let get_relatedness = |idx: UniIndex, node: &Node, loc: &LocationTree| {
+            let wildcard_state = loc.wildcard_accesses.get(idx).cloned().unwrap_or_default();
+            // If the tag is larger than `max_local_tag` then we treat the access as only foreign.
```
```suggestion
            // If the tag is larger than `max_local_tag` then the access can only be foreign.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1191,110 +1303,125 @@ impl<'tcx> LocationTree {
...
+        };
+
+        // This does a traversal across the tree updating children before their parents. The
+        // difference to `perform_normal_access` is that we take the access relatedness from
+        // the wildcard tracking state of the node instead of from the visitor itself.
+        //
+        // Unlike for a normal access, the iteration order is important for improving the
+        // accuracy of wildcard accesses.
+        // Ideally we would recalculate the wildcard datastructure on every access while
+        // accounting for `max_local_tag` however to improve performance we only approximate
```
```suggestion
        // accounting for `max_local_tag`. However, to improve performance we only approximate
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1191,110 +1303,125 @@ impl<'tcx> LocationTree {
...
+        // this behavior.
+        // This is only less accurate if `max_local_tag` is Some. In this case accesses cannot
+        // happen from a tag that is larger than `max_local_tag` this is easy to check for nodes
```
```suggestion
        // happen from a tag that is larger than `max_local_tag`. This is easy to check for nodes
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1191,110 +1303,125 @@ impl<'tcx> LocationTree {
...
+        // whose tag is larger, but otherwise we do not check for each node if its children have
+        // too large tags leading to further inaccuracies.
+        // However, in the case where all the exposed child nodes also happen to lose their access
+        // permissions during this access, we can achieve the same precision as the ideal algorithm.
+        // Since we update the children before their parent this means the wildcard datastructure
+        // has already been updated to reflect the children's new access level. So when we update
+        // the parent of them we see that no local access is possible anymore.
```
RalfJung: This is still quite confusing. The update of the data structure does not take into account `max_local_tag`, right? That would be incorrect since the next access may have a different `max_local_tag`. So this seems to be mixing the discussion of two separate concerns?

Is there a test checking for this extra precision you mention?

RalfJung (changes requested):
Thanks! However, there are still some comments that need further clarification.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4707/b39f1e797f0906bdc83a829f9ceedb8c213c192b..5881323a91d7e03673cbab7e8534adb5c0e8f7ab)*

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber: `cross_tree_from_main.rs` tests that an access from the main tree is foreign to wildcard subtrees.
`cross_tree_update_main_invalid_exposed.rs` tests the case where main doesn't have any valid exposed references.

The comment was probably a leftover from when the main root was separate

RalfJung: Then it's probably better to remove this paragraph.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1097,13 +1176,42 @@ impl<'tcx> LocationTree {
...
+            // However, using `accessed_root` is better since that will be smaller. It is still a valid choice
+            // because for nodes *in other trees*, if they are a parent of the accessed node then they
+            // are a parent of `accessed_root`.
+            //
+            // If we accessed the main tree then its root will be the smallest tag in the entire
+            // allocation, meaning we perform only a foreign access on all other subtrees.
+            //
+            // If `root` is the main root and it doesn't have any valid exposed children then this access
+            // will return an access error. This error doesn't get triggered on wildcard subtrees because we
+            // set `max_foreign_access==Write` on their root.
```
```suggestion
            // As a consequence of this, since the root of the main tree is the smallest tag in the entire
            // allocation, if the access occurred in the main tree then other subtrees will only see foreign accesses.
```

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber: The datastructure doesn't get updated to reflect `max_local_tag` but it does get updated to reflect the new permission of the child node.
So if our exposed child has initially `Unique` permission, and we now perform a wildcard write with a `max_local_tag` smaller than the child, then the child gets `Disabled` and the datastructure gets updated to reflect this.
The parent gets updated after the child, so it sees the updated state.

`cross_tree_update_older_invalid_exposed2.rs` tests that the extra precision is there.
`only_foreign_is_temporary()` in `reborrow.rs` tests that `max_local_tag` doesn't itself effect the datastructure.

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber:
```rust
            // If `root` is the main root and it doesn't have any valid exposed children then this access
            // will return an access error. This error doesn't get triggered on wildcard subtrees because we
            // set `max_foreign_access==Write` on their root.
```
This is still relevant.
It's about what happens if an access happens on another subtree but no exposed children exist on the main tree.
`cross_tree_update_main_invalid_exposed.rs` is the relevant test case.

RalfJung: I am confused now, above you said

> The comment was probably a leftover from when the main root was separate

and now you said actually it is still relevant...?

royAmmerschuber: This "comment" is split into three paragraphs.
The second paragraph is what I meant with leftover.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1191,110 +1303,125 @@ impl<'tcx> LocationTree {
...
+        let get_relatedness = |idx: UniIndex, node: &Node, loc: &LocationTree| {
+            let wildcard_state = loc.wildcard_accesses.get(idx).cloned().unwrap_or_default();
+            // If the tag is larger than `max_local_tag` then we treat the access as only foreign.
+            let only_foreign = max_local_tag.is_some_and(|max_local_tag| max_local_tag < node.tag);
+            wildcard_state.access_relatedness(access_kind, only_foreign)
+        };
+
+        // This does a traversal across the tree updating children before their parents. The
+        // difference to `perform_normal_access` is that we take the access relatedness from
+        // the wildcard tracking state of the node instead of from the visitor itself.
```
RalfJung, line 1333:
```suggestion
        // the wildcard tracking state of the node instead of from the visitor itself.
        // Note, however, that this is an approximation: there can be situations where a node is
        // marked as having an exposed foreign node, but actually that foreign node cannot be
        // the source of the access due to `max_local_tag`. The wildcard tracking cannot know
        // about `max_local_tag` so we will incorrectly assume that this might be a foreign access.
```

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber: The old comment only described the reasoning we do on each node only based on the data of the node itself. However, in the current implementation nodes are also affected by changes in their children (at least in some cases), so nodes with a smaller tag could also be affected.
This is why I thought it made more sense to describe what the value represents and not how we use it in the function.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1191,110 +1303,125 @@ impl<'tcx> LocationTree {
...
+        //
+        // Unlike for a normal access, the iteration order is important for improving the
+        // accuracy of wildcard accesses.
```
```suggestion
        // Unlike for a normal access, the iteration order is important for improving the
        // accuracy of wildcard accesses: processing the effects of this access further down the tree can
        // cause exposed nodes to lose permissions, thus updating the wildcard data structure, which
        // will be taken into account when processing the parent nodes. Also see the test
        // `cross_tree_update_older_invalid_exposed2.rs`.
```
And then the rest if this text below can be removed as now everything has been said, right? Or is some important point missing?

What is still not clear to me is why doing the children first is guaranteed to be more precise. If we did parents first, that would also cause updates of the wildcard data structure that propagate throughout the tree; is that never useful?

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: Ah. The second paragraph makes perfect sense to me. It is the third paragraph that I do not understand. That seems like some random implementation detail of the function we call, why do we need to discuss it here?

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: > This is why I thought it made more sense to describe what the value represents and not how we use it in the function.

I don't follow. Both your old and your new comment and my comment are all different ways of saying what the value represents.

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber: For the parent before children update order to be better there would need to be cases where the `max_foreign_access` of a node changes during a wildcard access.

During a wildcard access without `max_local_tag` the access order isn't important.
This is because for an exposed node a wildcard access could always be local which never disables the node. So the wildcard datastructure doesn't get updated during the access.

The only nodes whose exposure can be updated are those with `tag > max_local_tag` as a local access isn't possible on them.
This also means that for any node with `tag < max_local_tag` `max_foreign_access` never changes.

If we are updating a wildcard subtree then the parents (or cousins) exposure doesn't affect its children. This is because a foreign access can always come through the root. So `max_foreign_access` always remains the same.

For an access on the main subtree to not be UB there must exist a node $A$ with `tag < max_local_tag`. This node is foreign to all nodes with `tag > max_local_tag`. This means the `max_foreign_access` doesn't change for nodes with `tag > max_local_tag` because an access through $A$ is possible.

If such a node doesn't exist it is UB, and it doesn't matter in which order we update the other nodes. However, I think currently we don't detect the UB when two exposed cousins both have `tag > max_local_tag`.

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber: You're right. I moved it to the location where the error gets thrown.

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: Okay, thanks. I'm fine with sticking to the child-parent order for now. Maybe add a comment like
```
(Doing accesses in the opposite order cannot help with precision but the reasons are complicated;
see <https://github.com/rust-lang/miri/pull/4707#discussion_r2581661123>.)
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1337,19 +1333,15 @@ impl<'tcx> LocationTree {
...
-        // This is only less accurate if `max_local_tag` is Some. In this case accesses cannot
-        // happen from a tag that is larger than `max_local_tag` this is easy to check for nodes
-        // whose tag is larger, but otherwise we do not check for each node if its children have
-        // too large tags leading to further inaccuracies.
-        // However, in the case where all the exposed child nodes also happen to lose their access
-        // permissions during this access, we can achieve the same precision as the ideal algorithm.
-        // Since we update the children before their parent this means the wildcard datastructure
-        // has already been updated to reflect the children's new access level. So when we update
-        // the parent of them we see that no local access is possible anymore.
+        // accuracy of wildcard accesses if `max_local_tag` is some: processing the effects of this
```
RalfJung: "some" -> "`Some`"

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1191,110 +1299,124 @@ impl<'tcx> LocationTree {
...
+                let node = args.nodes.get_mut(args.idx).unwrap();
+
+                let protected = global.borrow().protected_tags.contains_key(&node.tag);
+
+                let Some(wildcard_relatedness) = get_relatedness(args.idx, node, args.loc) else {
+                    // There doesn't exist a valid exposed reference for this access to
+                    // happen through.
+                    // This can only happen if `root` is the main root. Since we set
+                    // `max_foreign_access==Write` on all wildcard roots so at least a foreign access
+                    // is always possible on all nodes in a wildcard subtrees.
```
```suggestion
                    // This can only happen if `root` is the main root: We set
                    // `max_foreign_access==Write` on all wildcard roots, so at least a foreign access
                    // is always possible on all nodes in a wildcard subtree.
```

tests/pass/tree_borrows/wildcard/reborrow.rs
```diff
@@ -0,0 +1,224 @@
...
+    //       │            │
+    //       │ ref3(Res)* │
+    //       │            │
+    //       └────────────┘
+
+    // Performs a foreign read on ref3 and doesn't update reb1.
+    // This temporarily treats ref3 as if only foreign accesses are possible to
+    // it. This is because the accessed tag reb2 has a larger tag than ref3.
+    let _x = *reb2;
+    // Should not update ref3, reb1 as we dont know if the access is local or foreign.
```
RalfJung:
```suggestion
    // Should not update ref3, reb1 as we don't know if the access is local or foreign.
```

RalfJung:

@rustbot author

Issue #4684, opened 2025-11-13
Support "fstat" on Linux


RalfJung:
The standard library uses `statx` instead of `fstat` on Linux, but Rust programs may still call `fstat` directly, so it makes sense to also support `fstat`. In particular, this is needed for [Tokio](https://github.com/tokio-rs/tokio/blob/6b705b3053d2c777e05cb60c758202ff9d4b2e7d/tokio/tests/net_unix_pipe.rs#L228-L229).

We already support `fstat` on a bunch of Unixes so this should be fairly easy. However, some slight adjustments may be needed since the exact set of fields in the `stat` type tends to be platform-dependent.

hulxv:
Hey. I would like to work on this issue. Can you please assign me and give me more details about it?

RalfJung:
That's great! You can assign yourself with `@rustbot claim`.

For the implementation side, you can take a look at how the fstat shim is wired up on macOS and other targets. You'll want to move that to `src/shims/unix/foreign_items.rs` where we have the operations that exist on all Unixes. `macos_fbsd_solarish_fstat` should be renamed to `fstat`, and `macos_fbsd_solarish_write_stat_buf` renamed to just `write_stat_buf`. You'll have to check which `stat` fields exist on Linux and adjust `write_stat_buf` accordingly. Finally, add a new test function in `tests/pass-dep/libc/libc-fs.rs`.

PR #4714, opened 2025-11-21
feat: Support `fstat` in linux

hulxv:
Close rust-lang/miri#4684

### DIFF
--- a/src/shims/unix/freebsd/foreign_items.rs
+++ b/src/shims/unix/freebsd/foreign_items.rs
@@ -148,7 +148,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let result = this.macos_fbsd_solarish_lstat(path, buf)?;
                 this.write_scalar(result, dest)?;
             }
-            "fstat" | "fstat@FBSD_1.0" => {
+            "fstat@FBSD_1.0" => {
                 let [fd, buf] = this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
                 let result = this.fstat(fd, buf)?;
                 this.write_scalar(result, dest)?;
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -130,7 +130,11 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let (modified_sec, modified_nsec) = metadata.modified.unwrap_or((0, 0));
         let mode = metadata.mode.to_uint(this.libc_ty_layout("mode_t").size)?;
 
-        let buf = this.deref_pointer_as(buf_op, this.libc_ty_layout("stat"))?;
+        // We do *not* use `deref_pointer_as` here since determining the right pointee type
+        // is highly non-trivial: it depends on which exact alias of the function was invoked
+        // (e.g. `fstat` vs `fstat64`), and then on FreeBSD it also depends on the ABI level
+        // which can be different between the libc used by std and the libc used by everyone else.
+        let buf = this.deref_pointer(buf_op)?;
         this.write_int_fields_named(
             &[
                 ("st_dev", metadata.dev.into()),
@@ -156,9 +160,6 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
         if matches!(&this.tcx.sess.target.os, Os::MacOs | Os::FreeBsd) {
             this.write_int_fields_named(
                 &[
-                    ("st_atime_nsec", access_nsec.into()),
-                    ("st_mtime_nsec", modified_nsec.into()),
-                    ("st_ctime_nsec", 0),
                     ("st_birthtime", created_sec.into()),
                     ("st_birthtime_nsec", created_nsec.into()),
                     ("st_flags", 0),
--- a/src/shims/unix/linux/foreign_items.rs
+++ b/src/shims/unix/linux/foreign_items.rs
@@ -53,11 +53,6 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let result = this.linux_statx(dirfd, pathname, flags, mask, statxbuf)?;
                 this.write_scalar(result, dest)?;
             }
-            "fstat" => {
-                let [fd, buf] = this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
-                let result = this.fstat(fd, buf)?;
-                this.write_scalar(result, dest)?;
-            }
             // epoll, eventfd
             "epoll_create1" => {
                 let [flag] = this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
--- a/src/shims/unix/solarish/foreign_items.rs
+++ b/src/shims/unix/solarish/foreign_items.rs
@@ -100,11 +100,6 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let result = this.macos_fbsd_solarish_lstat(path, buf)?;
                 this.write_scalar(result, dest)?;
             }
-            "fstat" | "fstat64" => {
-                let [fd, buf] = this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
-                let result = this.fstat(fd, buf)?;
-                this.write_scalar(result, dest)?;
-            }
             "readdir" => {
                 let [dirp] = this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
                 let result = this.readdir64("dirent", dirp)?;
--- a/tests/pass-dep/libc/libc-fs.rs
+++ b/tests/pass-dep/libc/libc-fs.rs
@@ -461,13 +461,15 @@ fn test_fstat() {
     let file = File::open(&path).unwrap();
     let fd = file.as_raw_fd();
 
-    let mut stat: libc::stat = unsafe { MaybeUninit::zeroed().assume_init() };
-    let res = unsafe { libc::fstat(fd, &mut stat) };
+    let mut stat = MaybeUninit::<libc::stat>::uninit();
+    let res = unsafe { libc::fstat(fd, stat.as_mut_ptr()) };
     assert_eq!(res, 0);
+    let stat = unsafe { stat.assume_init_ref() };
 
     assert_eq!(stat.st_size, 5);
     assert_eq!(stat.st_mode & libc::S_IFMT, libc::S_IFREG);
 
+    // Check that all fields are initialized.
     let _st_nlink = stat.st_nlink;
     let _st_blksize = stat.st_blksize;
     let _st_blocks = stat.st_blocks;


src/shims/unix/fs.rs
```diff
@@ -118,7 +118,7 @@ impl UnixFileDescription for FileHandle {
 
 impl<'tcx> EvalContextExtPrivate<'tcx> for crate::MiriInterpCx<'tcx> {}
 trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
-    fn macos_fbsd_solarish_write_stat_buf(
```
RalfJung, line 121: I said in the issue that you have to check whether Linux has any fields in its `stat` that we need to add here, on top of the "basics". Did you do that? It seems like you did not, because there *are* extra fields: `st_mtime_nsec` and friends.

src/shims/unix/linux/foreign_items.rs
```diff
@@ -53,7 +53,11 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let result = this.linux_statx(dirfd, pathname, flags, mask, statxbuf)?;
                 this.write_scalar(result, dest)?;
             }
-
+            "fstat" => {
+                let [fd, buf] = this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
+                let result = this.fstat(fd, buf)?;
+                this.write_scalar(result, dest)?;
+            }
```
RalfJung: As explained in the issue, this should be moved into `src/shims/unix/foreign_items.rs`, and then the corresponding blocks in the other unixes can be removed. (Some of them also make the function available under other names; those cannot be removed.)

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -380,6 +382,25 @@ fn test_sync_file_range() {
...
+    let path = utils::prepare_with_content("miri_test_libc_fstat.txt", b"hello");
+    let file = File::open(&path).unwrap();
+    let fd = file.as_raw_fd();
+
+    let mut stat: libc::stat = unsafe { MaybeUninit::zeroed().assume_init() };
+    let res = unsafe { libc::fstat(fd, &mut stat) };
+    assert_eq!(res, 0);
+
+    assert_eq!(stat.st_size, 5);
+    assert_eq!(stat.st_mode & libc::S_IFMT, libc::S_IFREG);
```
RalfJung, line 470: The test should also read all of the other fields that POSIX mandates, to make sure we have initialized them: https://pubs.opengroup.org/onlinepubs/009695299/basedefs/sys/stat.h.html.

And on systems that have `st_mtime_nsec` and friends, those should be read as well. This seems to be Linux, FreeBSD, macOS, Solarish... so, everything we test.

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -38,6 +38,8 @@ fn main() {
     test_posix_fadvise();
     #[cfg(target_os = "linux")]
     test_sync_file_range();
+    #[cfg(target_os = "linux")]
+    test_fstat();
```
RalfJung, line 45: fstat exists on all Unixes we test. So this new test should run everywhere.

RalfJung (changes requested):
Thanks for the PR! However, there are some things missing.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4714/5fbfc88282ba0db4a3f22bfdaa9c5556f4019eb9..8731b63d85307fce3bc87e14add529aa2dcf5d40)*

src/shims/unix/linux/foreign_items.rs
hulxv: I tried to remove them from other unixes but it seems all of them are under different names

RalfJung:
Please use `@rustbot ready` when the PR is ready for review again.

However, it seems like there are some CI failures that need to be dealt with before this is ready?

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -380,6 +381,43 @@ fn test_sync_file_range() {
...
+    assert!(stat.st_dev > 0);
+    // we can't predict uid/gid, but they should be accessible.
+    let _ = stat.st_uid;
+    let _ = stat.st_gid;
+    // for a regular file, rdev is 0.
+    assert_eq!(stat.st_rdev, 0);
+    // times should be non-zero (epoch is in the past).
+    assert!(stat.st_atime >= 0);
+    assert!(stat.st_mtime >= 0);
+    assert!(stat.st_ctime >= 0);
```
RalfJung: For most of these, checking a particular value does not make sense. You can just read the field.

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -380,6 +381,43 @@ fn test_sync_file_range() {
...
+    assert_eq!(stat.st_size, 5);
+    assert_eq!(stat.st_mode & libc::S_IFMT, libc::S_IFREG);
+    assert_eq!(stat.st_nlink, 0);
+    assert_eq!(stat.st_blksize, 0);
+    assert_eq!(stat.st_blocks, 0);
+    assert_eq!(stat.st_ino, 0);
+    assert!(stat.st_dev > 0);
+    // we can't predict uid/gid, but they should be accessible.
+    let _ = stat.st_uid;
+    let _ = stat.st_gid;
```
RalfJung: This does not actually read the field! You have to store it in an actual variable.
```suggestion
    let _uid = stat.st_uid;
    let _gid = stat.st_gid;
```
The same applies below with the nsec fields.

src/shims/unix/linux/foreign_items.rs
RalfJung: You can add `fstat64` here as well.

src/shims/unix/macos/foreign_items.rs
```diff
@@ -58,7 +58,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             }
             "fstat" | "fstat64" | "fstat$INODE64" => {
```
RalfJung:
```suggestion
            "fstat$INODE64" => {
```

hulxv:
for some reason, it fails on `aarch64` only
```
  full stderr:
  hello dup fd
  error: Undefined Behavior: memory access failed: attempting to access 8 bytes, but got alloc23440+0x78 which is at or beyond the end of the allocation of size 120 bytes
    --> tests/pass-dep/libc/libc-fs.rs:465:24
     |
  LL |     let res = unsafe { libc::fstat(fd, &mut stat) };
     |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
     |
  help: alloc23440 was allocated here:
    --> tests/pass-dep/libc/libc-fs.rs:464:9
     |
  LL |     let mut stat: libc::stat = unsafe { MaybeUninit::zeroed().assume_init() };
     |         ^^^^^^^^
     = note: inside `test_fstat` at tests/pass-dep/libc/libc-fs.rs:465:24: 465:50
  note: inside `main`
    --> tests/pass-dep/libc/libc-fs.rs:45:5
     |
  LL |     test_fstat();
     |     ^^^^^^^^^^^^
  
  full stdout:
  hello dup fd
  ```
  
maybe the problem comes from `libc` itself in aarch64? It's 128 bits, but it allocates 120. I think it's related to the size of `libc::stat`. Maybe the problem is related to the padding size on aarch64. We can do something like that:
```rust
#[repr(C)]
struct SafeStat {
    stat: libc::stat,
    _pad: [u64; 16],
}
let mut safe_stat: SafeStat = unsafe { MaybeUninit::zeroed().assume_init() };
let stat_ptr = &mut safe_stat as *mut SafeStat as *mut libc::stat;

let res = unsafe { libc::fstat(fd, stat_ptr) };
```

what do you think?

RalfJung:
Note that the target it failed on is x86_64-unknown-freebsd. So you are looking at the wrong target.

We definitely should *not* do anything like that in the test. If a test fails, usually that means you have to fix the code, not the test.

src/shims/unix/solarish/foreign_items.rs
```diff
@@ -102,7 +102,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             }
             "fstat" | "fstat64" => {
```
RalfJung: As I have already said twice, this match arm should be removed now. Please make sure you understand my comments before marking the PR as ready, and ask if anything is unclear. Do not just guess and update the PR. Only mark it as ready once you are confident everything is ready. It's a lot of extra work for me if I have to keep re-checking and repeating everything I already said.

src/shims/unix/fs.rs
```diff
@@ -141,8 +141,11 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 ("st_gid", metadata.gid.into()),
                 ("st_rdev", 0),
                 ("st_atime", access_sec.into()),
+                ("st_atime_nsec", access_nsec.into()),
                 ("st_mtime", modified_sec.into()),
+                ("st_mtime_nsec", modified_nsec.into()),
                 ("st_ctime", 0),
+                ("st_ctime_nsec", 0),
```
RalfJung, line 152: You are now writing these fields twice on macos and freebsd.

src/shims/unix/linux/foreign_items.rs
RalfJung: You didn't yet add `fstat64` here. What was your plan?

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -150,7 +150,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             }
             "fstat" | "fstat@FBSD_1.0" => {
```
RalfJung:
```suggestion
            "fstat@FBSD_1.0" => {
```
As elsewhere, the "fstat" case should be removed here, since it is already handled in the "unix" handler now.

src/shims/unix/linux/foreign_items.rs
RalfJung: Hm, if we are fstat64 we should probably use the stat64 type, not the stat type... but somehow this worked before? Not entirely sure what is happening. This might be related to the freebsd test failure.

RalfJung:
I looked into the BSD issue. Apparently on BSD there is `libc::unix::bsd::freebsdlike::freebsd::freebsd12::stat` and `libc::stat` and those are not the same type. Great. Who on Earth screwed that up?!??

The fix for us is fairly simple though -- we just trust the type signature used by the program, since double-checking it is apparently impractically difficult:
```diff
diff --git a/src/shims/unix/fs.rs b/src/shims/unix/fs.rs
index 16214d7ef..c33100b24 100644
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -130,7 +130,11 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let (modified_sec, modified_nsec) = metadata.modified.unwrap_or((0, 0));
         let mode = metadata.mode.to_uint(this.libc_ty_layout("mode_t").size)?;
 
-        let buf = this.deref_pointer_as(buf_op, this.libc_ty_layout("stat"))?;
+        // We do *not* use `deref_pointer_as` here since determining the right pointee type
+        // is highly non-trivial: it depends on which exact alias of the function was invoked
+        // (e.g. `fstat` vs `fstat64`), and then on FreeBSD it also depends on the ABI level
+        // which can be different between the libc used by std and the libc used by everyone else.
+        let buf = this.deref_pointer(buf_op)?;
         this.write_int_fields_named(
             &[
                 ("st_dev", metadata.dev.into()),
```

RalfJung:
I have taken the liberty to fix the last few remaining issues. Thanks for giving this a shot!

hulxv:
> I have taken the liberty to fix the last few remaining issues. Thanks for giving this a shot!

Thanks for that and I am sorry, I was very busy past days. I will take a look on new issues

PR #4723, opened 2025-11-25
preliminary refactor for wildcard reborrows

royAmmerschuber:
This is part of rust-lang/miri#4707 and only contains the refactoring & moving around of code. Functionality should be exactly the same as before.

### DIFF
--- a/src/borrow_tracker/tree_borrows/tree.rs
+++ b/src/borrow_tracker/tree_borrows/tree.rs
@@ -772,15 +772,16 @@ impl<'tcx> Tree {
             span,
         )?;
 
+        // The order in which we check if any nodes are invalidated only
+        // matters to diagnostics, so we use the root as a default tag.
+        let start_idx = match prov {
+            ProvenanceExtra::Concrete(tag) => self.tag_mapping.get(&tag).unwrap(),
+            ProvenanceExtra::Wildcard => self.root,
+        };
+
         // Check if this breaks any strong protector.
         // (Weak protectors are already handled by `perform_access`.)
         for (loc_range, loc) in self.locations.iter_mut(access_range.start, access_range.size) {
-            // The order in which we check if any nodes are invalidated only
-            // matters to diagnostics, so we use the root as a default tag.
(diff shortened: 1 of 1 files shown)


src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -781,41 +781,43 @@ impl<'tcx> Tree {
...
                 ProvenanceExtra::Wildcard => self.root,
             };
-            TreeVisitor { nodes: &mut self.nodes, loc }.traverse_this_parents_children_other(
-                start_idx,
-                // Visit all children, skipping none.
-                |_| ContinueTraversal::Recurse,
-                |args: NodeAppArgs<'_>| {
-                    let node = args.nodes.get(args.idx).unwrap();
-                    let perm = args.loc.perms.entry(args.idx);
+            let check_strong_protector = |args: NodeAppArgs<'_>| {
```
RalfJung: Why are you let-binding this closure?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -927,16 +887,25 @@ impl<'tcx> Tree {
             // `tests/pass/tree_borrows/tree-borrows.rs` for an example of
             // why this is important.
             for (loc_range, loc) in self.locations.iter_mut_all() {
+                let source_idx = source_idx.unwrap();
```
RalfJung: This can be moved out of the loop.

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber: because #4707 needs to call it outside of the TreeVisitor.

RalfJung: See https://github.com/rust-lang/miri/pull/4707#discussion_r2559382962 -- I think there's a better way to share the code here. Probably best to leave that to the other PR.

RalfJung:

@rustbot author

PR #4727, opened 2025-11-27
CI: ensure we build on Ubuntu 22

RalfJung:


RalfJung:
As expected this currently fails; GenMC needs fixing to be buildable on Ubuntu 22.

RalfJung:
I'll close this and hope that we can instead update the runners used by rustc CI to Ubuntu 24. :)

Issue #4639, opened 2025-10-21
native-lib mode segfaults when calling Rust function pointer from C


weiznich:
With #4625 closed (thanks again for the fast turn around :heart:) I gave it another try to run some diesel tests using miri. This time I do not hit an unsupported operation anymore, but a segfault.

Steps to reproduce:

```
git clone https://github.com/diesel-rs/diesel
cd diesel
git checkout 4200a2c # should be main as of today
cd diesel_tests
# likely needs to adjust the libsqlite.so path to your system
MIRIFLAGS=-Zmiri-native-lib=/usr/lib64/libsqlite3.so.0.8.6 DATABASE_URL=":memory:" cargo +nightly miri test --features sqlite  types::i32_to_sql
```

This results in the following output in the end (after some warnings that this test calls into FFI code):

```
  process didn't exit successfully: `/home/weiznich/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/cargo-miri runner /home/weiznich/Documents/rust/diesel/target/miri/x86_64-unknown-linux-gnu/debug/deps/integration_tests-dc79c07a95be5a22 'types::i32_to_sql'` (signal: 11, SIGSEGV: invalid memory reference)

```

Running the code under gdb results in the following backtrace:

```
#0  0x00007fffe5817408 in ?? ()
#1  0x00007fffd6553be9 in sqlite3LeaveMutexAndCloseZombie.lto_priv.0 () from /usr/lib64/libsqlite3.so.0.8.6
#2  0x00007fffd655421c in sqlite3Close () from /usr/lib64/libsqlite3.so.0.8.6
#3  0x000055555603f052 in ffi_call_unix64 ()
#4  0x000055555603deff in ffi_call_int ()
#5  0x000055555603dbda in ffi_call ()
#6  0x0000555555d50192 in <rustc_const_eval::interpret::eval_context::InterpCx<miri::machine::MiriMachine> as miri::shims::native_lib::EvalContextExtPriv>::call_native_with_args::{closure#0} ()
#7  0x0000555555e7fae6 in <rustc_const_eval::interpret::eval_context::InterpCx<miri::machine::MiriMachine> as miri::shims::foreign_items::EvalContextExtPriv>::emulate_foreign_item_inner ()
[...]

PR #4728, opened 2025-11-27
Basic support for FFI callbacks (at least stop segfaulting and report a proper error)

weiznich:
This commit adds basic support for FFI callbacks by registering a shim function via libffi. This shim function currently only notes if it's actually called and registers an error for these cases. The main motivation for this is to prevent miri segfaulting as described in rust-lang/miri#4639. Obviously the code is likely highly unsafe, especially if something goes wrong, etc as it's going back and forth over an ffi boundary and just casts pointers as it likes.

In the future miri could try to continue execution in the registered callback, although as far as I understand Ralf that is no easy problem. There are already preparations for this, like actually receiving the arguments and setting up the structure to return something.

This produces the following error for diesel:

```
error: unsupported operation: Tried to call a function pointer via FFI boundary. That's not supported yet by miri
        This function pointer was registered by a call to `sqlite3_create_function_v2` using an argument of the type `std::option::Option<unsafe extern "C" fn(*mut std::ffi::c_void)>`
    --> /home/weiznich/Documents/rust/diesel/diesel/src/sqlite/connection/raw.rs:270:37
     |
 270 |         let close_result = unsafe { ffi::sqlite3_close(self.internal_connection.as_ptr()) };
     |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ unsupported operation occurred here
     |
     = note: BACKTRACE on thread `types::i32_to_s`:
     = note: inside `<diesel::sqlite::connection::raw::RawConnection as std::ops::Drop>::drop` at /home/weiznich/Documents/rust/diesel/diesel/src/sqlite/connection/raw.rs:270:37: 270:90
     = note: inside `std::ptr::drop_in_place::<diesel::sqlite::connection::raw::RawConnection> - shim(Some(diesel::sqlite::connection::raw::RawConnection))` at /home/weiznich/.rustup/toolchains/miri/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:805:1: 807:25
     = note: inside `std::ptr::drop_in_place::<diesel::SqliteConnection> - shim(Some(diesel::SqliteConnection))` at /home/weiznich/.rustup/toolchains/miri/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:805:1: 807:25
note: inside `types::query_to_sql_equality::<diesel::sql_types::Integer, i32>`
    --> diesel_tests/tests/types.rs:1428:1
     |
1428 | }
     | ^
note: inside `types::i32_to_sql_integer`
    --> diesel_tests/tests/types.rs:192:13
     |
 192 |     assert!(query_to_sql_equality::<Integer, i32>("0", 0));
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside closure
    --> diesel_tests/tests/types.rs:191:24
     |
 190 | #[diesel_test_helper::test]
     | --------------------------- in this attribute macro expansion
 191 | fn i32_to_sql_integer() {
     |                        ^
     = note: this error originates in the attribute macro `test` which comes from the expansion of the attribute macro `diesel_test_helper::test` (in Nightly builds, run with -Z macro-backtrace for more info)

```

I feel that's better than nothing, although a better error would include trace information about the registering side and the call side as well.

There are no tests for this yet, as I'm not familiar with miri's test setup, how to structure them and where to put them. Any pointers for this would be helpful.

### DIFF
--- a/src/alloc_addresses/mod.rs
+++ b/src/alloc_addresses/mod.rs
@@ -169,8 +169,15 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                         this.tcx.try_get_global_alloc(alloc_id)
                     {
                         let fn_sig = this.tcx.fn_sig(instance.def_id()).skip_binder().skip_binder();
-                        let ptr = crate::shims::native_lib::build_libffi_closure_ptr(this, fn_sig)?;
-                        ptr.cast()
+                        let fn_ptr = crate::shims::native_lib::build_libffi_closure(this, fn_sig)?;
+
+                        #[expect(
+                            clippy::as_conversions,
+                            reason = "No better way to cast a function ptr to a ptr"
+                        )]
+                        {
+                            fn_ptr as *const _
+                        }
                     } else {
                         dummy_alloc(params)
                     }
--- a/src/shims/native_lib/mod.rs
+++ b/src/shims/native_lib/mod.rs
@@ -75,9 +75,8 @@ impl AccessRange {
     }
 }
 
-/// The data passed to the closure shim function used
-/// to intercept function pointer calls from native
-/// code
+/// The data passed to the closure shim function used to intercept function pointer calls from
+/// native code.
 struct CallbackData<'tcx, 'this> {
     ecx: &'this InterpCx<'tcx, MiriMachine<'tcx>>,
 }
@@ -338,9 +337,7 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // Read the bytes that make up this argument. We cannot use the normal getter as
                 // those would fail if any part of the argument is uninitialized. Native code
                 // is kind of outside the interpreter, after all...
-                let ret: Box<[u8]> =
-                    Box::from(alloc.inspect_with_uninit_and_ptr_outside_interpreter(range));
-                ret
+                Box::from(alloc.inspect_with_uninit_and_ptr_outside_interpreter(range))
             }
             either::Either::Right(imm) => {
                 let mut bytes: Box<[u8]> = vec![0; imm.layout.size.bytes_usize()].into();
@@ -446,7 +443,7 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
             // Scalar types have already been handled above.
             ty::Adt(adt_def, args) => self.adt_to_ffitype(layout.ty, *adt_def, args)?,
             // Rust uses `()` as return type for `void` function, which becomes `Tuple([])`.
-            ty::Tuple(t_list) if (*t_list).deref().is_empty() => FfiType::void(),
+            ty::Tuple(t_list) if t_list.len() == 0 => FfiType::void(),
             _ => {
                 throw_unsup_format!("unsupported type for native call: {}", layout.ty)
             }
@@ -454,22 +451,17 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
     }
 }
 
-/// This function sets up a new libffi Closure to intercept
+/// This function sets up a new libffi closure to intercept
 /// calls to rust code via function pointers passed to native code.
-///
-/// It manages tho setup the underlying libffi datastructure by initializing the
-/// right number of arguments with the right types, as well a setting
-/// the right return type
 ///
 /// Calling this function leaks the data passed into the libffi closure as
 /// these need to be available until the execution terminates as the native
 /// code side could store a function pointer and only call it at a later point.
-/// This is for example done by SQLite, which uses function pointers as destructors
-/// which then only run at teardown time
-fn build_libffi_closure<'tcx, 'this>(
+pub fn build_libffi_closure<'tcx, 'this>(
     this: &'this MiriInterpCx<'tcx>,
     fn_sig: rustc_middle::ty::FnSig<'tcx>,
-) -> InterpResult<'tcx, libffi::middle::Closure<'this>> {
+) -> InterpResult<'tcx, unsafe extern "C" fn()> {
+    // Compute argument and return types in libffi representation.
     let mut args = Vec::new();
     for input in fn_sig.inputs().iter() {
         let layout = this.layout_of(*input)?;
@@ -481,38 +473,19 @@ fn build_libffi_closure<'tcx, 'this>(
         let layout = this.layout_of(res_type)?;
         this.ty_to_ffitype(layout)?
     };
+
+    // Build the actual closure.
     let closure_builder = libffi::middle::Builder::new().args(args).res(res_type);
     let data = CallbackData { ecx: this };
     let data = Box::leak(Box::new(data));
-    let closure = closure_builder.into_closure(callback_callback, data);
-    interp_ok(closure)
-}
-
-/// This function wraps build_libffi_closure
-/// to return the correct pointer that is expected to be
-/// passed into native code as function pointer.
-///
-/// This function returns a `*const c_void` ptr to erase the
-/// actual function type, that is only known at runtime.
-pub fn build_libffi_closure_ptr<'tcx, 'this>(
-    this: &'this MiriInterpCx<'tcx>,
-    fn_sig: rustc_middle::ty::FnSig<'tcx>,
-) -> InterpResult<'tcx, *const std::ffi::c_void> {
-    let closure = build_libffi_closure(this, fn_sig)?;
+    let closure = closure_builder.into_closure(libffi_closure_callback, data);
     let closure = Box::leak(Box::new(closure));
+
+    // The actual argument/return type doesn't matter.
+    let fn_ptr = unsafe { closure.instantiate_code_ptr::<unsafe extern "C" fn()>() };
     // Libffi returns a **reference** to a function ptr here.
-    // (The actual argument type doesn't matter)
-    let fn_ptr =
-        unsafe { closure.instantiate_code_ptr::<unsafe extern "C" fn(*const std::ffi::c_void)>() };
     // Therefore we need to dereference the reference to get the actual function pointer.
-    let fn_ptr = *fn_ptr;
-    #[expect(clippy::as_conversions, reason = "No better way to cast a function ptr to a ptr")]
-    {
-        // After that we need to cast the function pointer to the
-        // expected pointer type. At this point we don't actually care about the
-        // type of this pointer.
-        interp_ok(fn_ptr as *const std::ffi::c_void)
-    }
+    interp_ok(*fn_ptr)
 }
 
 /// A shim function to intercept calls back from native code into the interpreter
@@ -520,7 +493,7 @@ pub fn build_libffi_closure_ptr<'tcx, 'this>(
 ///
 /// For now this shim only reports that such constructs are not supported by miri.
 /// As future improvement we might continue execution in the interpreter here.
-unsafe extern "C" fn callback_callback<'tcx, 'this>(
+unsafe extern "C" fn libffi_closure_callback<'tcx, 'this>(
     _cif: &libffi::low::ffi_cif,
     _result: &mut c_void,
     _args: *const *const c_void,


src/shims/native_lib/mod.rs
```diff
@@ -310,6 +333,44 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // casting the integer in `byte` to a pointer and using that.
         let bytes = match v.as_mplace_or_imm() {
             either::Either::Left(mplace) => {
+                let ptr_overwrite = match v.layout.ty.kind() {
+                    ty::Adt(_adt_def, args) =>
+                        if let ty::FnPtr(fn_ptr, _header) = args.type_at(0).kind() {
```
oli-obk: What is this adt + fndef match detecting? I assumed you'd just want to detect fn ptrs being converted to FFI data and inject your callback there, but that doesn't seem to be what is happening.

oli-obk:
Please add a test for this, too.

src/shims/native_lib/mod.rs
weiznich: You are correct that this is likely wrong. The underlying problem there is that this does not only need to handle plain function pointers but also something like `Option<unsafe fn()>`. Now I do not know much about the involved compiler types, so that is my crude attempt to get the relevant information. If you know a better way I'm certainly open for suggestions. 

(Beside of that it seems like I did not actually handle the plain function pointer case, I should definitely add that)

weiznich:
> Please add a test for this, too.

I'm more than happy to do that, but as pointed out in the OP I don't know where and how. Is there any documentation for this or can you provide a pointer to the right direction?

src/shims/native_lib/mod.rs
RalfJung: The purpose of this function is to transmit raw bytes to the C side. So matching on the type at all here is almost certainly not the right call.

RalfJung: Ah I think I see what you are doing here... you are *changing* the bytes if they represent a function pointer. That doesn't work, we need the bytes to be exactly the same on both sides.  After all, both sides might be casting those pointers to integers and print them, and something would be seriously cursed if the same function pointer then printed as different addresses on the two sides of the FFI call. Also, if you store a function pointer in a `static` and then pass a *pointer* to that static to C, this function will never see the function pointer, it will only ever see the pointer to the function pointer.

The code that determines the addresses of allocations (which includes both data allocations for normal static/stack/heap memory, and function allocations for the "memory that a function pointer points to") is here:

https://github.com/rust-lang/miri/blob/5a47c2535b171ac9e576c32b9a2692112eae6940/src/alloc_addresses/mod.rs#L139-L178

The FFI closure allocation needs to happen in the `AllocKind::Function` case there, so that we can then make the "virtual" address (i.e., the address in Miri's purely logical interpreter memory) of this function pointer the same as the real address of the closure. This is a key invariant of Miri in native_lib mode: Miri's logical/virtual addresses are the same as the real underlying addresses, and therefore C code can follow pointers stored in Miri memory and everything works out.

src/shims/native_lib/mod.rs
```diff
@@ -439,7 +528,12 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
         interp_ok(match layout.ty.kind() {
             // Scalar types have already been handled above.
             ty::Adt(adt_def, args) => self.adt_to_ffitype(layout.ty, *adt_def, args)?,
-            _ => throw_unsup_format!("unsupported argument type for native call: {}", layout.ty),
+            // Functions with no declared return type (i.e., the default return)
+            // have the output_type `Tuple([])`.
+            ty::Tuple(t_list) if (*t_list).deref().is_empty() => FfiType::void(),
```
RalfJung: This should not be needed -- function pointers are scalar types.

RalfJung:
> There are no tests for this yet, as I'm not familiar with miri's test setup, how to structure them and where to put them. Any pointers for this would be helpful.

The native-lib tests are in `tests/native-lib`. You can just add new `.rs` files inside `pass` or `fail` there and they will be picked up by the test suite. Use `./miri test` to run the entire test suite; `./miri test native` runs only tests whose name contains "native" which gets you the relevant results for this PR more quickly. See https://github.com/rust-lang/miri/blob/master/CONTRIBUTING.md#building-and-testing-miri for more details.

src/shims/native_lib/mod.rs
```diff
@@ -536,3 +630,92 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+unsafe extern "C" fn callback_callback(
+    cif: &libffi::low::ffi_cif,
+    result: &mut c_void,
+    args: *const *const c_void,
+    infos: &CallbackData<'_, '_>,
+) {
+    debug_assert_eq!(cif.nargs as usize, infos.args.len());
+    let mut rust_args = Vec::with_capacity(infos.args.len());
+    // cast away the pointer to pointer
```
RalfJung: As a general style note, please use full sentences in comments, including an upper-case first word and a period at the end.

src/shims/native_lib/mod.rs
```diff
@@ -536,3 +630,92 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                      That's not supported yet by miri\n This function pointer was registered by a call to `{}` \
+                      using an argument of the type `{}`", infos.link_name, infos.ty)
+                .into(),
+        });
+    });
+
+    // write here the output
+    // For now we just try to write some dummy output
+    // by using some "reasonable" default values
+    // to prevent crashing
```
RalfJung: I don't think we should return to the FFI code here. That code clearly wanted this function pointer to do something, and arbitrary nonsense could happen if we just skip whatever that is and return. This could be worse than a segfault. We have to abort execution here.

src/shims/native_lib/mod.rs
weiznich: The function I happened to use for testing returned `void`. As this function is now also called to construct the return value types for the libffi closure type this was required to make my test case working. 

That written: You are correct that this is not needed for a minimal support of callback over fro, it just happens to help with my particular test case and was rather straightforward to add.

src/shims/native_lib/mod.rs
weiznich: Thanks for the pointer. This sounds reasonable, but I fail to see how I would access information about the callback/function/allocation type there. Without this information it's not possible to construct the corresponding libffi type.

src/shims/native_lib/mod.rs
RalfJung: > this function is now also called to construct the return value types for the libffi closure type 

Ah, that makes sense.

Given that I think we should never return from that closure (see my other comments), I think we should not need this either.

src/shims/native_lib/mod.rs
RalfJung: There is `get_fn_alloc` but it's private (in `compiler/rustc_const_eval/src/interpret/memory.rs` in the rustc tree). We should probably make it public, and then also rename it to `try_get_alloc_fn` for better consistency with other, similar methods.

weiznich: This should hopefully be addressed now with https://github.com/rust-lang/miri/pull/4728/commits/9ccb197c323baf3a88062cb6c7753382b2b72551. Thanks for pointing to the right locations, it looks much robuster now :tada:

src/shims/native_lib/mod.rs
weiznich: This is still required as we need the `FfiType` while constructing the libffi Closure type to get the correct ABI as far as I understand.

src/shims/native_lib/mod.rs
```diff
@@ -536,3 +595,45 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    }
+
+    // We abort the execution at this point as we cannot return the
+    // expected value here.
+    eprintln!(
+        "Tried to call a function pointer via FFI boundary. \
+         That's not supported yet by miri\nThis function pointer was registered by a call to `{}` \
+         using an argument of the type `{}`",
+        "todo: fill in name", "todo: fill in type"
+    );
```
weiznich: This is the place I'm currently rather unhappy with, but I don't know how to solve better.

I would like to emit a proper "error" here and not just print something. Is there a easy way to do this?

Also I would like to include the name of the function that registered the closure and the type of the argument, but I'm not sure how I get that information in `src/alloc_address/mod.rs`. 

Maybe you have some pointer for both problems as well?

Also I expect that you want to tweak this message to something better, so consider that a placeholder until someone suggest a better wording.

tests/native-lib/fail/call_function_ptr.rs
```diff
@@ -0,0 +1,21 @@
...
+fn pass_fn_ptr() {
+    extern "C" {
+        fn call_fn_ptr(s: Option<extern "C" fn()>);
+    }
+
+    extern "C" fn nop() {}
+
+    unsafe {
+        call_fn_ptr(None); // this one is fine
+        call_fn_ptr(Some(nop)); // this one is not
```
weiznich: As of now this test fails, even as it emits the "right" output as it doesn't emit an actual error, but just a `eprintln` + non-zero exist code.

src/shims/native_lib/mod.rs
```diff
@@ -536,3 +595,45 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let ptr = StrictPointer::new(Provenance::Wildcard, Size::from_bytes(args.addr()));
+                // This offset moves the pointer to the next argument
+                args = unsafe { args.offset(1) };
+                Scalar::from_pointer(ptr, infos.this)
+            }
+            // the other types
+            _ => todo!(),
+        };
+        rust_args.push(scalar);
+    }
```
weiznich: Not sure if we should remove this block as well or fill in more types instead? For now this is unused but as soon as this callback starts actually to continue the execution of the rust code here, this would be required as first step.

weiznich:
@rustbot ready

(Not really ready as in ready to merge, but more ready as in this is waiting on reviewer input for now)

src/shims/native_lib/mod.rs
oli-obk: Could handle this separately in another PR and make all tests here have an already implemented return type. But not too important

src/alloc_addresses/mod.rs
```diff
@@ -169,9 +169,39 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                         Align::from_bytes(1).unwrap(),
                         params,
                     );
-                    let ptr = alloc_bytes.as_ptr();
+                    let mut ptr = alloc_bytes.as_ptr();
                     // Leak the underlying memory to ensure it remains unique.
                     std::mem::forget(alloc_bytes);
+                    if let Some(GlobalAlloc::Function { instance, .. }) =
```
oli-obk, line 168: The above code is irrelevant for this code path. Split the function code path from the vtable code path and unwrap the function global alloc as all AllocKind::Function should be backed by a GlobalAlloc::Function I think

src/shims/native_lib/mod.rs
oli-obk: Via the InterpCx you should be able to report an error, but we def still need the exit as there is no way to leave this FFI code and go back to the original Rust/miri code

oli-obk (changes requested):

src/shims/native_lib/mod.rs
RalfJung: The problem will be passing in the InterpCx. We can *not* just use the `this` that was passed to `build_libffi_closure` -- that reference gets invalidated when `build_libffi_closure` returns.

We need some way to pass this value from `call_native_with_args` to the callback that gets invoked while this native code runs.

weiznich: Would it be possible to just put the context in a (thread local) global variable just before calling a native function and take it from there in the callback?

RalfJung: That seems like the easiest option. The closure may find itself called on a different thread, at which point we have to just hard abort without a nice error, but at least for the common case this should work (and anyway if the C code spawns threads that interact in any way with Rust code, this native code integration is pretty much completely broken).

src/shims/native_lib/mod.rs
weiznich: I can do this, but in that case I likely also need to change existing tests to make it work with the implemented feature as they also use void as return/argument type for the callback.

src/alloc_addresses/mod.rs
RalfJung: > all AllocKind::Function should be backed by a GlobalAlloc::Function I think

Not quite, there's also the machine-specific function pointer types, aka `dlsym` fn ptrs. So the `if let` is probably a good idea.

src/alloc_addresses/mod.rs
```diff
@@ -162,18 +163,41 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                    std::mem::forget(alloc_bytes);
-                    ptr
+                #[cfg(all(unix, feature = "native-lib"))]
+                AllocKind::Function => {
+                    if let Some(GlobalAlloc::Function { instance, .. }) =
+                        this.tcx.try_get_global_alloc(alloc_id)
+                        && let rustc_middle::ty::InstanceKind::Item(def_id) = instance.def
+                        && let Some(closure) = crate::shims::native_lib::build_libffi_closure(
+                            this,
+                            this.tcx.fn_sig(def_id).skip_binder().skip_binder(),
```
RalfJung: If all you need is a `def_id`, you can get it from `instance.def_id()`; no need to match on the instance.

src/alloc_addresses/mod.rs
```diff
@@ -162,18 +163,41 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                        )
+                    {
+                        let closure = Box::leak(Box::new(closure));
+                        // Libffi returns a **reference** to a function ptr here
+                        // (The actual argument type doesn't matter)
+                        let fn_ptr = unsafe {
+                            closure.instantiate_code_ptr::<unsafe extern "C" fn(*const std::ffi::c_void)>()
+                        };
+                        // Therefore we need to dereference the reference to get the actual function pointer
```
RalfJung: As a general note, when you write a sentence, please end it in a full stop.

src/alloc_addresses/mod.rs
```diff
@@ -162,18 +163,41 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                            this,
+                            this.tcx.fn_sig(def_id).skip_binder().skip_binder(),
+                        )
+                    {
+                        let closure = Box::leak(Box::new(closure));
+                        // Libffi returns a **reference** to a function ptr here
+                        // (The actual argument type doesn't matter)
+                        let fn_ptr = unsafe {
+                            closure.instantiate_code_ptr::<unsafe extern "C" fn(*const std::ffi::c_void)>()
+                        };
```
RalfJung: Why is this done here and not inside `build_libffi_closure`?

src/shims/native_lib/mod.rs
```diff
@@ -439,11 +450,62 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            _ => {
+                throw_unsup_format!("unsupported argument type for native call: {}", layout.ty)
+            }
         })
     }
 }
 
+pub fn build_libffi_closure<'tcx, 'this>(
+    this: &'this MiriInterpCx<'tcx>,
+    fn_ptr: rustc_middle::ty::FnSig<'tcx>,
```
```suggestion
    fn_sig: rustc_middle::ty::FnSig<'tcx>,
```
This isn't a pointer.

src/shims/native_lib/mod.rs
```diff
@@ -439,11 +450,62 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+) -> Option<libffi::middle::Closure<'this>> {
+    let mut args = Vec::new();
+    for input in fn_ptr.inputs().iter() {
+        let layout = match this.layout_of(*input) {
```
RalfJung: Just use `?`. Something is very very odd when layout fails.

src/shims/native_lib/mod.rs
```diff
@@ -439,11 +450,62 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            Ok(layout) => layout,
+            Err(e) => {
+                tracing::info!(?e, "Skip closure");
+                return None;
+            }
+        };
+        let ty = match this.ty_to_ffitype(layout).report_err() {
```
RalfJung: This violates the contract of `report_err`, which is that you must report an error.

Please just use `?`.

src/shims/native_lib/mod.rs
```diff
@@ -439,11 +450,62 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
         interp_ok(match layout.ty.kind() {
             // Scalar types have already been handled above.
             ty::Adt(adt_def, args) => self.adt_to_ffitype(layout.ty, *adt_def, args)?,
-            _ => throw_unsup_format!("unsupported argument type for native call: {}", layout.ty),
+            // Functions with no declared return type (i.e., the default return)
+            // have the output_type `Tuple([])`.
```
```suggestion
            // Rust uses `()` as return type for `void` function, which becomes `Tuple([])`.
```

src/shims/native_lib/mod.rs
```diff
@@ -439,11 +450,62 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            ty::Tuple(t_list) if (*t_list).deref().is_empty() => FfiType::void(),
+            _ => {
+                throw_unsup_format!("unsupported argument type for native call: {}", layout.ty)
```
```suggestion
                throw_unsup_format!("unsupported type for native call: {}", layout.ty)
```

src/shims/native_lib/mod.rs
```diff
@@ -439,11 +450,62 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        match this.ty_to_ffitype(layout).report_err() {
+            Ok(ty) => ty,
+            Err(e) => {
+                tracing::info!(?e, "Skip closure");
+                return None;
+            }
+        }
+    };
+    let closure_builder = libffi::middle::Builder::new().args(args).res(res_type);
+    let data = CallbackData {};
```
RalfJung: So we can pass data to the closure?

Could we use this to avoid using thread-locals, and instead pass in a pointer to somewhere in memory that the current `ecx` pointer will be stored in before invoking native code?

src/shims/native_lib/mod.rs
```diff
@@ -536,3 +598,38 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         interp_ok(true)
     }
 }
+
+struct CallbackData {}
```
RalfJung: Please generally declare types before they are used.
Also, please add a doc comment.

src/shims/native_lib/mod.rs
```diff
@@ -439,11 +450,62 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // have the output_type `Tuple([])`.
+            ty::Tuple(t_list) if (*t_list).deref().is_empty() => FfiType::void(),
+            _ => {
+                throw_unsup_format!("unsupported argument type for native call: {}", layout.ty)
+            }
         })
     }
 }
 
+pub fn build_libffi_closure<'tcx, 'this>(
```
RalfJung, line 460: Please add a doc comment.
Same for `callback_callback` below -- it's private, but this code is tricky enough to warrant good documentation.

src/shims/native_lib/mod.rs
```diff
@@ -536,3 +598,38 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+struct CallbackData {}
+
+unsafe extern "C" fn callback_callback(
+    _cif: &libffi::low::ffi_cif,
+    _result: &mut c_void,
+    _args: *const *const c_void,
+    _infos: &CallbackData,
+) {
+    if let Some(this) = INTERP_CTX.get() {
+        let ctx = unsafe { this.cast::<MiriInterpCx<'_>>().as_ref() };
```
```suggestion
        let ecx = unsafe { this.cast::<MiriInterpCx<'_>>().as_ref() };
```

src/shims/native_lib/mod.rs
```diff
@@ -536,3 +598,38 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let stacktrace = ctx.generate_stacktrace();
+        crate::diagnostics::report_msg(
+            DiagLevel::Error,
+            "tried to call a function pointer through the FFI boundary".into(),
```
```suggestion
            "calling a function pointer through the FFI boundary".into(),
```

src/shims/native_lib/mod.rs
```diff
@@ -536,3 +598,38 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+unsafe extern "C" fn callback_callback(
+    _cif: &libffi::low::ffi_cif,
+    _result: &mut c_void,
+    _args: *const *const c_void,
+    _infos: &CallbackData,
+) {
+    if let Some(this) = INTERP_CTX.get() {
+        let ctx = unsafe { this.cast::<MiriInterpCx<'_>>().as_ref() };
+        let stacktrace = ctx.generate_stacktrace();
+        crate::diagnostics::report_msg(
```
RalfJung: Nice idea to use the normal diagnostics machinery here!

Instead of directly invoking `report_msg` (that function really should be private), can you call `report_result`? That makes this then use the same codepaths as other "unsupported" messages.

`err_unsup_format!` should let you create a `InterpErrorKind`, on which you can call `into()` to construct an `InterpErrorInfo`.

tests/native-lib/fail/call_function_ptr.notrace.stderr
```diff
@@ -0,0 +1,61 @@
...
+   = help: what this means is that Miri will easily miss Undefined Behavior related to incorrect usage of this shared memory, so you should not take a clean Miri run as a signal that your FFI code is UB-free
+   = note: BACKTRACE:
+   = note: inside `pass_fn_ptr` at tests/native-lib/fail/call_function_ptr.rs:LL:CC
+note: inside `main`
+  --> tests/native-lib/fail/call_function_ptr.rs:LL:CC
+   |
+LL |     pass_fn_ptr()
+   |     ^^^^^^^^^^^^^
+
+warning: sharing a function pointer with a native function called via FFI
```
RalfJung: Do we even still need this warning if we have the more informative and more precise error later?

src/alloc_addresses/mod.rs
weiznich: Hopefully I managed to fix all new comment.

RalfJung:
Looks great, thanks. :) I pushed a commit with a few minor cleanups; let's land this.

RalfJung:
Thinking about this some more, I think storing the `InterpCx` in a closure isn't quite right... "technically" that is an alias.

We'll have to allocate memory somewhere that serves as interchange between the FFI code and the closure. I can do that in a follow-up PR; I have some other libffi changes planned anyway.

PR #4729, opened 2025-11-27
test that target features given in the target spec are enabled

RalfJung:
This currently fails... not sure if that is a regression, but it might be the same as https://github.com/rust-lang/rust/issues/149314.
@bjorn3 do you think your PR will fix this?

### DIFF
--- a/tests/pass/target-spec-implies-target-feature.rs
+++ b/tests/pass/target-spec-implies-target-feature.rs
@@ -1,0 +1,8 @@
+//! Ensure that the target features given in the target spec are actually enabled.
+//@only-target: armv7
+
+fn main() {
+    assert!(cfg!(target_feature = "v7"));
+    assert!(cfg!(target_feature = "vfp2"));
+    assert!(cfg!(target_feature = "thumb2"));
+}


RalfJung:
I could just test this, I still have your PR checked out after all... no, turns out it does not fix this.
```
$ ./x run miri --target armv7-unknown-linux-gnueabihf -- ../miri/tests/pass/target-spec-implies-target-feature.rs
thread 'main' (1000) panicked at ../miri/tests/pass/target-spec-implies-target-feature.rs:6:5:
assertion failed: cfg!(target_feature = "vfp3")
```

RalfJung:
However it also already fails before https://github.com/rust-lang/miri/pull/4648. So we probably need a different test.

tests/pass/target-spec-implies-target-feature.rs
```diff
@@ -0,0 +1,8 @@
+//! Ensure that the target features given in the target spec are actually enabled.
+//@only-target: armv7
+
+fn main() {
+    assert!(cfg!(target_feature = "v7"));
+    assert!(cfg!(target_feature = "vfp3"));
+    assert!(cfg!(target_feature = "thumb2"));
```
bjorn3, line 7: `rustc --target armv7-unknown-linux-gnueabihf --print cfg` shows v7, vfp2 and thumb2, but not vfp3.

```
[...]
target_feature="aclass"
target_feature="dsp"
target_feature="fpregs"
target_feature="thumb2"
target_feature="v5te"
target_feature="v6"
target_feature="v6k"
target_feature="v6t2"
target_feature="v7"
target_feature="vfp2"
[...]
```

bjorn3:
My PR should fix this for v7 and thumb2. vfp3 is not a target feature recognized by rustc.

```
warning: unstable feature specified for `-Ctarget-feature`: `vfp3`
  |
  = note: this feature is not stably supported; its behavior can change in the future
```

RalfJung:
vfp3 is missing due to https://github.com/rust-lang/rust/issues/149399.

RalfJung:
FWIW I also tried `std::arch::is_arm_feature_detected` and that still fails even when I set the LLVM backend. Not sure what is up with that.
```
error: "v7" feature cannot be detected at run-time
 --> ../miri/tests/pass/target-spec-implies-target-feature.rs:7:13
  |
7 |     assert!(std::arch::is_arm_feature_detected!("v7"));
  |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
```
The macro docs make it sound like if the feature is available at compile time, the macro will work. Maybe I am misunderstanding the comment.

EDIT: I just misunderstood the code, these never work.

RalfJung:
> My PR should fix this for v7 and thumb2

Indeed it does, I confirmed that locally. Great!

RalfJung:
I integrated this test into https://github.com/rust-lang/miri/pull/4730.

PR #4731, opened 2025-11-28
Call rustc_driver::main() for MIRI_BE_RUSTC=host

bjorn3:
This way we don't need a special case in `MiriBeRustCompilerCalls`.

### DIFF
--- a/src/bin/miri.rs
+++ b/src/bin/miri.rs
@@ -272,18 +272,13 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
     }
 }
 
-struct MiriBeRustCompilerCalls {
-    target_crate: bool,
-}
+/// This compiler produces rlibs that are meant for later consumption by Miri.
+struct MiriDepCompilerCalls;
 
-impl rustc_driver::Callbacks for MiriBeRustCompilerCalls {
+impl rustc_driver::Callbacks for MiriDepCompilerCalls {
     #[allow(rustc::potential_query_instability)] // rustc_codegen_ssa (where this code is copied from) also allows this lint
     fn config(&mut self, config: &mut Config) {
-        if !self.target_crate {
-            // For a host crate, we fully behave like rustc.
-            return;
-        }
-        // For a target crate, we emit an rlib that Miri can later consume.
+        // We don't need actual codegen, we just emit an rlib that Miri can later consume.
         config.make_codegen_backend = Some(Box::new(make_miri_codegen_backend));
 
         // Avoid warnings about unsupported crate types. However, only do that we we are *not* being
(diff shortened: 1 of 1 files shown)


src/bin/miri.rs
```diff
@@ -457,32 +450,24 @@ fn main() {
 
     // If the environment asks us to actually be rustc, then do that.
     if let Some(crate_kind) = env::var_os("MIRI_BE_RUSTC") {
+        if crate_kind == "host" {
+            rustc_driver::main();
```
```suggestion
            // For host crates like proc macros and build scripts, we are an entirely normal rustc.
            // These eventually produce actual binaries and never run in Miri.
            return rustc_driver::main();
```

src/bin/miri.rs
```diff
@@ -272,17 +272,11 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
     }
 }
 
-struct MiriBeRustCompilerCalls {
-    target_crate: bool,
-}
+struct MiriBeRustCompilerCalls;
```
RalfJung: I feel like we should rename this to indicate that it is only used for target crates. Maybe `MiriDepCompilerCalls`?
```suggestion
/// This compiler produces rlibs that are meant for later consumption by Miri.
struct MiriDepCompilerCalls;
```

src/bin/miri.rs
```diff
@@ -272,17 +272,11 @@ impl rustc_driver::Callbacks for MiriCompilerCalls {
...
 
 impl rustc_driver::Callbacks for MiriBeRustCompilerCalls {
     #[allow(rustc::potential_query_instability)] // rustc_codegen_ssa (where this code is copied from) also allows this lint
     fn config(&mut self, config: &mut Config) {
-        if !self.target_crate {
-            // For a host crate, we fully behave like rustc.
-            return;
-        }
         // For a target crate, we emit an rlib that Miri can later consume.
```
```suggestion
        // We don't need actual codegen, we just emit an rlib that Miri can later consume.
```

src/bin/miri.rs
```diff
@@ -367,16 +361,15 @@ impl rustc_driver::Callbacks for MiriBeRustCompilerCalls {
...
-            // cargo-miri has patched the compiler flags to make these into check-only builds,
-            // but we are still emulating regular rustc builds, which would perform post-mono
-            // const-eval during collection. So let's also do that here, even if we might be
-            // running with `--emit=metadata`. In particular this is needed to make
-            // `compile_fail` doc tests trigger post-mono errors.
-            // In general `collect_and_partition_mono_items` is not safe to call in check-only
-            // builds, but we are setting `-Zalways-encode-mir` which avoids those issues.
-            let _ = tcx.collect_and_partition_mono_items(());
-        }
+        // cargo-miri has patched the compiler flags to make these into check-only builds,
```
RalfJung: That's not even correct any more, is it? You removed that flag patching in https://github.com/rust-lang/miri/pull/4648.

bjorn3: The call is still necessary here as `DummyCodegenBackend` doesn't have it. I will update the comment though.

RalfJung:
That's annoying that we can't use `return` to make it clear that the call won't return... is there some other nice way to make that syntactically obvious?

bjorn3:
`match rustc_driver::main() {}` works.

src/bin/miri.rs
```diff
@@ -457,32 +448,26 @@ fn main() {
 
     // If the environment asks us to actually be rustc, then do that.
     if let Some(crate_kind) = env::var_os("MIRI_BE_RUSTC") {
+        if crate_kind == "host" {
+            // For host crates like proc macros and build scripts, we are an entirely normal rustc.
+            // These eventually produce actual binaries and never run in Miri.
+            match rustc_driver::main() {}
```
```suggestion
            match rustc_driver::main() {
                // Empty match proves this function will never return.
            }
```

RalfJung:
LGTM, thanks! r=me after adding the comment and squashing.

Issue #4572, opened 2025-09-07
GenMC correctness issues


RalfJung:
Let's keep a list of all the cases where GenMC mode is not able to execute correct Rust code properly
- [ ] Temporal mixing of atomic and non-atomic accesses: GenMC doesn't know about the non-atomic accesses that occurred, so an atomic load can return invalid results
- [ ] Mixed-size accesses: Rust allows racing mixed-size accesses as long as all accesses are reads; GenMC does not support this, which can lead to loads returning invalid results
- [ ] Optimized local variables being moved into memory: when a local variable is moved from its efficient representation as an `Immediate` to its in-memory representation, GenMC thinks the write happens then and there, whereas really the write happened earlier. This can leas to GenMC reporting a data race when there is none.
- [ ] Incorrect wrapping behavior of fetch_add/sub: GenMC internally always uses 64bit arithmetic, so its internal value can diverge from what he program sees. This is relevant when a fetch_add is followed by a fetch_max, as the latter can then pick the wrong result.

Cc @Patrick-6 -- please let me know if I missed anything

Patrick-6:
- The temporal mixing issue is also linked to reading the initial value of globals (which currently returns a dummy value).
(Miri tests with a comment like this are affected by this issue: `// FIXME(genmc,HACK)`)
- For mixed-size accesses: This limitation also affects non-racing mixed-size accesses (e.g., single threaded).
The basic assumption of GenMC currently is: once any memory location is accessed with a given address and size combination, any future accesses to the same location use the same address and size. (different size but same address gives an error, different address but overlapping access results in incorrect "uninitialized read" errors or returning garbage values).
Note that this should only affect _atomic_ loads (non-atomic loads are served directly from Miri's memory, which handles them correctly).
- Memory Allocator: Incorrect behavior if any thread requests more than 4 GiB of memory over the course of one program run (this includes losses from alignment constraints. Note that memory is never reused: Any allocations count towards the limit).
At that point, GenMC will (silently) return out addresses that were already handed out earlier (due to an integer truncation).
Requests with sufficiently large sizes or alignments can also cause integer overflows in the allocator.
(I've written a fix for these issues, but it has not yet been upstreamed in GenMC.)

Patrick-6:
One more:
- GenMC doesn't support Rust programs that use uninitialized memory in specific ways, such as uninitialized globals (assumed to always be initialized by GenMC) or to re-uninitialized memory, such as with `MaybeUninit::uninit()`.
[...]

PR #4733, opened 2025-11-28
genmc: Bug fixes

michaliskok:
This PR bumps the GenMC version to 0.16.1:
 - Fixes the incorrect treatment of `atomic_fetch_{umax,umin}` (rust-lang/miri#4572)
 - Fixes overflowing on large allocations (rust-lang/miri#4572)
 - Collects all wrappers around the GenMC API in a single file (`Exploration.cpp`)
 - Performs minor cleanups
 - Fixes a couple of... FIXMEs

r? @RalfJung
(Each commit includes a description and can be reviewed separately.)

### DIFF
--- a/genmc-sys/build.rs
+++ b/genmc-sys/build.rs
@@ -28,7 +28,7 @@ mod downloading {
     /// The GenMC repository the we get our commit from.
     pub(crate) const GENMC_GITHUB_URL: &str = "https://github.com/MPI-SWS/genmc.git";
     /// The GenMC commit we depend on. It must be available on the specified GenMC repository.
-    pub(crate) const GENMC_COMMIT: &str = "aa10ed65117c3291524efc19253b5d443a4602ac";
+    pub(crate) const GENMC_COMMIT: &str = "22d3d0b44dedb4e8e1aae3330e546465e4664529";
 
     /// Ensure that a local GenMC repo is present and set to the correct commit.
     /// Return the path of the GenMC repo and whether the checked out commit was changed.
--- a/tests/genmc/pass/atomics/rmw_ops.rs
+++ b/tests/genmc/pass/atomics/rmw_ops.rs
@@ -67,7 +67,7 @@ macro_rules! test_rmw_edge_cases {
         x.store(10, ORD);
         assert_eq(10, x.fetch_add(<$int>::MAX, ORD)); // definitely overflows, so new value of x is smaller than 10
         assert_eq(<$int>::MAX.wrapping_add(10), x.fetch_max(10, ORD)); // new value of x should be 10
-        // assert_eq(10, x.load(ORD)); // FIXME(genmc,#4572): enable this check once GenMC correctly handles min/max truncation.
+        assert_eq(10, x.load(ORD));
     }};
 }
 


genmc-sys/cpp/include/MiriInterface.hpp
```diff
@@ -126,7 +126,7 @@ struct MiriGenmcShim : private GenMCDriver {
 
     /**** Memory (de)allocation ****/
     auto handle_malloc(ThreadId thread_id, uint64_t size, uint64_t alignment) -> uint64_t;
-    auto handle_free(ThreadId thread_id, uint64_t address) -> bool;
+    auto handle_free(ThreadId thread_id, uint64_t address) -> std::unique_ptr<std::string>;
```
RalfJung, line 132: This is an odd return type, I don't think we use that anywhere else? Needs at least a comment explaining the meaning of it.

src/concurrency/genmc/mod.rs
```diff
@@ -622,15 +622,15 @@ impl GenmcCtx {
...
+        let free_result =  self
             .handle
             .borrow_mut()
             .pin_mut()
-            .handle_free(self.active_thread_genmc_tid(machine), address.bytes())
-        {
+            .handle_free(self.active_thread_genmc_tid(machine), address.bytes());
+        if let Some(error) = free_result.as_ref() {
             // FIXME(genmc): improve error handling.
             // An error was detected, so we get the error string from GenMC.
```
RalfJung: "so we get the error string from GenMC" isn't the case any more, so the last line can be removed.

(Github does not support code suggestions for commit-by-commit review. :(  )

RalfJung (commented):
Thanks!

The last commit mentions issues with memory allocation; what exactly are you referring to?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4733/ea0d7001155fe2b822fd03eb39b6fca94d4d648e..4cbc75e9fdb628a24793172dfc878d8799e26827)*

RalfJung:
Also, please run `./miri fmt` to apply our standard formatting to the Rust code.
@rustbot author

RalfJung:
> The last commit mentions issues with memory allocation; what exactly are you referring to?

Ah, the PR description mentions "Fixes overflowing on large allocations". Is this about a single large allocation, or many allocations accumulating to exhaust the 4GB per-thread limit? How large is "large"? Is it possible to test for that? What happens now when the per-thread 4GB memory region is exhausted?

genmc-sys/cpp/include/MiriInterface.hpp
michaliskok: It's used in [`handle_execution_end`](https://github.com/rust-lang/miri/blob/59009fd7f18228439fa5623936478873d1c3a1c0/genmc-sys/src/lib.rs#L383).

I can add a `FreeResult` struct with a single `error` field if you prefer, but it's a bit verbose. (Other result types have more than one field.)

RalfJung: > It's used in [handle_execution_end](https://github.com/rust-lang/miri/blob/59009fd7f18228439fa5623936478873d1c3a1c0/genmc-sys/src/lib.rs#L383).

Ah, fair.

No need to add a new type, but please add a short comment here saying that a non-null return value indicates an error.

michaliskok:
> > The last commit mentions issues with memory allocation; what exactly are you referring to?
> 
> Ah, the PR description mentions "Fixes overflowing on large allocations". Is this about a single large allocation, or many allocations accumulating to exhaust the 4GB per-thread limit? How large is "large"? Is it possible to test for that? What happens now when the per-thread 4GB memory region is exhausted?

Each thread can allocate up to 4GB. If the limit is exceeded, the allocator would silently wrap around 0 and handle duplicate addresses. The new commit makes GenMC crash if the limit is exceeded, and doesn't overflow if a size/alignment >2^32 is passed to `alloc_aligned`.

genmc-sys/cpp/include/MiriInterface.hpp
RalfJung: And also add such a comment to `handle_execution_end`, please.

RalfJung:
> The new commit makes GenMC crash if the limit is exceeded, 

That may make it tricky to add a test then, depending on the exit status this will trigger.

michaliskok:
> > The new commit makes GenMC crash if the limit is exceeded,
> 
> That may make it tricky to add a test then, depending on the exit status this will trigger.

The alternative would be to return an error code to Miri and let Miri crash.

RalfJung:
I tried what happened when one thread allocates more than 4GB memory in total (but not at the same time, i.e. we allocate chunks of 512MB and free them again):
```
./miri run --features=genmc bench-cargo-miri/big-allocs/src/main.rs -Zmiri-genmc
```
With this PR, that leads to a panic:
```
thread 'rustc' (123993) panicked at src/concurrency/genmc/mod.rs:597:9:
assertion `left != right` failed: GenMC malloc returned nullptr.
```

EDIT: I get the same with an 8GB allocation, using this test:
```rust
fn main() {
    let _v = Vec::<u8>::with_capacity(8 * 1024 * 1024 * 1024);
}
```
So, I can't reproduce the GenMC crash.

RalfJung:
> The alternative would be to return an error code to Miri and let Miri crash.

The alternative would be to make Miri emit a nice error saying we ran out of memory for this thread. :)

We don't have to resolve that in this PR, but ideally we'd have some tests before we mark the item in https://github.com/rust-lang/miri/issues/4572 as resolved.

tests/genmc/fail/simple/alloc_large.rs
```diff
@@ -0,0 +1,25 @@
+//@revisions: single multiple
+//@compile-flags: -Zmiri-genmc -Zmiri-disable-stacked-borrows -Zmiri-ignore-leaks
+//@error-in-other-file: resource exhaustion
+
+// Ensure that we panic if GenMC fails to fulfill an allocation.
```
RalfJung:
```suggestion
// Ensure that we emit a proper error if GenMC fails to fulfill an allocation.
```

tests/genmc/fail/simple/alloc_large.rs
```diff
@@ -0,0 +1,25 @@
...
+// Two variants: one for a single large allocation, one for multiple large ones.
+//
+// We never deallocate the memory, so leak-checks must be disabled.
```
RalfJung: We don't have to disable them as the program stops with an error anyway. Only programs that actually finish executing are leak-checked.

tests/genmc/fail/simple/alloc_large.rs
```diff
@@ -0,0 +1,25 @@
+//@revisions: single multiple
+//@compile-flags: -Zmiri-genmc -Zmiri-disable-stacked-borrows -Zmiri-ignore-leaks
+//@error-in-other-file: resource exhaustion
+
+// Ensure that we panic if GenMC fails to fulfill an allocation.
+// Two variants: one for a single large allocation, one for multiple large ones.
```
RalfJung:
```suggestion
// Two variants: one for a single large allocation, one for multiple ones that are
// all individual below the limit, but together they are too big.
```

tests/genmc/fail/simple/alloc_large.rs
RalfJung: Also, the program actually *does* deallocate the memory. :) `Vec`s clean up when they get dropped.

RalfJung:
Here's a little patch to give the user some more explanation for why they may run out of address space in genmc mode:
```diff
diff --git a/src/diagnostics.rs b/src/diagnostics.rs
index bb8ba1969..8e252d306 100644
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -362,6 +362,10 @@ pub fn report_result<'tcx>(
                 vec![
                     note!("this is likely not a bug in the program; it indicates that the program performed an operation that Miri does not support"),
                 ],
+            ResourceExhaustion(ResourceExhaustionInfo::AddressSpaceFull) if ecx.machine.data_race.as_genmc_ref().is_some() =>
+                vec![
+                    note!("in GenMC mode, the address space is limited to 4GB per thread, and addresses cannot be reused")
+                ],
             UndefinedBehavior(AlignmentCheckFailed { .. })
                 if ecx.machine.check_alignment == AlignmentCheck::Symbolic
             =>
```

genmc-sys/cpp/include/MiriInterface.hpp
```diff
@@ -125,8 +125,11 @@ struct MiriGenmcShim : private GenMCDriver {
     void handle_fence(ThreadId thread_id, MemOrdering ord);
 
     /**** Memory (de)allocation ****/
+
     auto handle_malloc(ThreadId thread_id, uint64_t size, uint64_t alignment) -> uint64_t;
-    auto handle_free(ThreadId thread_id, uint64_t address) -> bool;
+
+    /** Returns null or an error string if an error occurs. */
```
RalfJung:
```suggestion
    /** Returns null on success, or an error string if an error occurs. */
```
This might not be proper English, but otherwise one can easily read this as "(returns or or an error string) if an error occurs"...

michaliskok:
> I tried what happened when one thread allocates more than 4GB memory in total (but not at the same time, i.e. we allocate chunks of 512MB and free them again):
> 
> ```
> ./miri run --features=genmc bench-cargo-miri/big-allocs/src/main.rs -Zmiri-genmc
> ```
> 
> With this PR, that leads to a panic:
> 
> ```
> thread 'rustc' (123993) panicked at src/concurrency/genmc/mod.rs:597:9:
> assertion `left != right` failed: GenMC malloc returned nullptr.
> ```
> 
> EDIT: I get the same with an 8GB allocation, using this test:
> 
> ```rust
> fn main() {
>     let _v = Vec::<u8>::with_capacity(8 * 1024 * 1024 * 1024);
> }
> ```
> 
> So, I can't reproduce the GenMC crash.

Indeed, sorry about that.

I thought I made GenMC panic in all cases, but that's not the case. For allocations taking place via `handleMalloc`, GenMC returns nullptr (and potentially an error trace, but error reporting is not hooked up yet anyway). I added a test ensuring that Miri produces an error report if addresses get depleted.

I also incorporated your other comments.

genmc-sys/cpp/include/MiriInterface.hpp
RalfJung: I had edited this suggestion after initially posting it, looks like that edit got lost?

RalfJung:
Looks great, thanks :)

PR #4739, opened 2025-12-04
Move `TreeVisitor` into its own file & make it generic.

royAmmerschuber:
Moves the tree visitor logic into its own file.

Also makes `TreeVisitor` generic instead of always taking a `LocationTree`. This isn't really necessary, but makes it clear that `TreeVisitor` doesn't depend on `LocationTree`.


RalfJung:
This looks great, thanks!

PR #4740, opened 2025-12-04
TreeBorrows: Put accesses diagnostic parameters into a single struct

royAmmerschuber:
All the `perform_access` functions take 5 parameters that are purely used for diagnostic purposes. This puts them all into a single struct and passes them by reference.

### DIFF
--- a/src/borrow_tracker/tree_borrows/diagnostics.rs
+++ b/src/borrow_tracker/tree_borrows/diagnostics.rs
@@ -77,6 +77,32 @@ pub struct Event {
     pub span: Span,
 }
 
+/// Diagnostics data about the current access and the location we are accessing.
+/// Used to create history events and errors.
+#[derive(Clone, Debug)]
+pub struct DiagnosticInfo {
+    pub alloc_id: AllocId,
+    pub span: Span,
+    /// The range the diagnostic actually applies to.
+    /// This is always a subset of `access_range`.
+    pub transition_range: Range<u64>,
+    /// The range the access is happening to. Is `None` if this is the protector release access
+    pub access_range: Option<AllocRange>,
+    pub access_cause: AccessCause,
+}
+impl DiagnosticInfo {
+    /// Creates a history event.
+    pub fn create_event(&self, transition: PermTransition, is_foreign: bool) -> Event {
+        Event {
+            transition,
+            is_foreign,
+            access_cause: self.access_cause,
+            access_range: self.access_range,
+            transition_range: self.transition_range.clone(),
+            span: self.span,
+        }
+    }
+}
 /// List of all events that affected a tag.
 /// NOTE: not all of these events are relevant for a particular location,
 /// the events should be filtered before the generation of diagnostics.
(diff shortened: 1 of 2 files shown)


src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -77,6 +77,32 @@ pub struct Event {
     pub span: Span,
 }
 
+/// Diagnostics data about the current access and the location we are accessing.
+/// Used to create history events and errors.
+#[derive(Clone, Debug)]
+pub struct DiagnosticsExtra {
```
RalfJung: Why "extra"?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -91,12 +91,9 @@ impl LocationState {
         nodes: &mut UniValMap<Node>,
         wildcard_accesses: &mut UniValMap<WildcardState>,
         access_kind: AccessKind,
-        access_cause: AccessCause,        //diagnostics
-        access_range: Option<AllocRange>, //diagnostics
         relatedness: AccessRelatedness,
-        span: Span,                 //diagnostics
-        location_range: Range<u64>, //diagnostics
         protected: bool,
+        diagnostics: &DiagnosticsExtra,
```
RalfJung: FWIW in Stacked Borrows we have a lot of `dcx: &mut DiagnosticCx<'_, '_, 'tcx>`; would that also make sense here? (apparently mutability is currently not needed in TB)

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -942,14 +939,16 @@ impl<'tcx> Tree {
...
                     source_idx,
-                    loc_range,
-                    Some(access_range),
                     access_kind,
-                    access_cause,
                     global,
-                    alloc_id,
-                    span,
                     ChildrenVisitMode::VisitChildrenOfAccessed,
+                    &DiagnosticsExtra {
```
RalfJung: Please let-bind this before the call.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -978,14 +977,16 @@ impl<'tcx> Tree {
...
                         Some(source_idx),
-                        loc_range,
-                        None,
                         access_kind,
-                        access_cause,
                         global,
-                        alloc_id,
-                        span,
                         ChildrenVisitMode::SkipChildrenOfAccessed,
+                        &DiagnosticsExtra {
```
RalfJung: Same here, please let-bind.

src/borrow_tracker/tree_borrows/diagnostics.rs
royAmmerschuber: Maybe it should be more something like `AccessContext`? It's just a bunch of metadata about the current access.

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber: I don't think mutability is necessary. We could also pass it by value.

src/borrow_tracker/tree_borrows/diagnostics.rs
RalfJung: I like to have "Diag(nostics)" in the name as that makes it purpose more clear.

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: Passing it by-ref is good, both for performance and to document that this isn't changing.

src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -80,7 +80,7 @@ pub struct Event {
 /// Diagnostics data about the current access and the location we are accessing.
 /// Used to create history events and errors.
 #[derive(Clone, Debug)]
-pub struct DiagnosticsExtra {
+pub struct AccessDiagnostics {
```
RalfJung: I'm not a big fan of this name -- "diagnostic" is the term used in the Rust compiler for a fully computed warning/error output, and that's not what this is.

What about "DiagnosticInfo"?

src/borrow_tracker/tree_borrows/diagnostics.rs
```diff
@@ -280,28 +306,25 @@ impl History {
...
     /// tag that lacked those permissions.
     pub conflicting_info: &'node NodeDebugInfo,
-    // What kind of access caused this error (read, write, reborrow, deallocation)
-    pub access_cause: AccessCause,
     /// Which tag, if any, the access that caused this error was made through, i.e.
     /// which tag was used to read/write/deallocate.
     /// Not set on wildcard accesses.
     pub accessed_info: Option<&'node NodeDebugInfo>,
+    /// Diagnostic data about the current access.
+    pub access_diagnostics: &'node DiagnosticInfo,
```
RalfJung:
```suggestion
    pub access_info: &'node DiagnosticInfo,
```
Seems more consistent?

royAmmerschuber: `accessed_info` and `access_info` seems confusing

RalfJung: Maybe rename `accessed_info` to `accessed_node_info`?

PR #4741, opened 2025-12-04
TreeBorrows: split `Tree::perform_protector_release_access` from `Tree::perform_access`

royAmmerschuber:
The `Tree::perform_access` function currently is used for both normal accesses and protector release accesses.
However, both of these are handled by different branches and don't share any code, so this splits them into two separate functions.

### DIFF
--- a/src/borrow_tracker/tree_borrows/tree.rs
+++ b/src/borrow_tracker/tree_borrows/tree.rs
@@ -828,7 +828,9 @@ impl<'tcx> Tree {
     ) -> InterpResult<'tcx> {
         self.perform_access(
             prov,
-            Some((access_range, AccessKind::Write, diagnostics::AccessCause::Dealloc)),
+            access_range,
+            AccessKind::Write,
+            diagnostics::AccessCause::Dealloc,
             global,
             alloc_id,
             span,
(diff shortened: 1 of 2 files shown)


src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -920,7 +914,9 @@ impl<'tcx> Tree {
     pub fn perform_access(
         &mut self,
         prov: ProvenanceExtra,
-        access_range_and_kind: Option<(AllocRange, AccessKind, diagnostics::AccessCause)>,
+        access_range: AllocRange,
+        access_kind: AccessKind,
+        access_cause: AccessCause,
```
RalfJung: `access_cause` is also just used for diagnostics I assume?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -934,61 +930,75 @@ impl<'tcx> Tree {
...
+                access_cause,
+                global,
+                alloc_id,
+                span,
+                ChildrenVisitMode::VisitChildrenOfAccessed,
+            )?;
+        }
+        interp_ok(())
+    }
+    /// this is the special access that is applied on protector release:
```
```suggestion

    /// This is the special access that is applied on protector release:
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -934,61 +930,75 @@ impl<'tcx> Tree {
...
+    /// - the access will be applied only to accessed locations of the allocation,
+    /// - it will not be visible to children,
+    /// - it will be recorded as a `FnExit` diagnostic access
+    /// - and it will be a read except if the location is `Unique`, i.e. has been written to,
+    ///   in which case it will be a write.
+    /// - otherwise identical to `Tree::perform_access`
+    pub fn perform_protector_release_access(
```
RalfJung: I think we more typically call this "protector end access", so
```suggestion
    pub fn perform_protector_end_access(
```

RalfJung (commented):
Makes sense! Just some minor nits.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4741/4e1f862c12596dbc53fa1bfee36c5079d2941c8c..079c67ad7983f5f20b77a6d5cb73a7f5216e7ebf)*

RalfJung:

@rustbot author

PR #4749, opened 2025-12-07
improve backtrace rendering

RalfJung:
There are several things I don't like about our current backtrace rendering:
- the first span it mentions is always the same span that the errors occurs at
- we do not say in which thread the error happened if the stack has just a single frame
- `= note: BACKTRACE:` looks clunky

So here I am trying to improve the situation. For the common case, the rendering looks something like this:
```
error: Undefined Behavior: entering unreachable code
  --> tests/fail/never_transmute_void.rs:LL:CC
   |
LL |         match v.0 {}
   |               ^^^ Undefined Behavior occurred here
   |
   = note: stack backtrace:
           0: m::f
               at tests/fail/never_transmute_void.rs:LL:CC
           1: main
               at tests/fail/never_transmute_void.rs:LL:CC
```

We also show a new span at the bottom of the backtrace for all stacks except the one of `main`. That span indicates why this particular function got called:
- For non-main threads, it points at the user-visible span corresponding to thread creation. This helps with https://github.com/rust-lang/miri/issues/4066, though I am not sure if we should consider it to resolve that issue since we only show one span, not a full backtrace (but that may be a better default as full cross-thread backtraces will be huge). The example in the issue now looks as follows:
```
error: unsupported operation: can't call foreign function `unknown` on OS `linux`
 --> thread.rs:5:21
  |
5 | fn bad() { unsafe { unknown() } }
  |                     ^^^^^^^^^ unsupported operation occurred here
  |
  = help: this means the program tried to do something Miri does not support; it does not indicate a bug in the program
  = note: this is on thread `unnamed-1`
note: the current function got called indirectly due to this code
 --> thread.rs:8:5
  |
8 |     std::thread::spawn(bad);
  |     ^^^^^^^^^^^^^^^^^^^^^^^
```
- For global constructors and destructors, it points at the `static` item that carries the magic linker annotation that turns into a constructor/desstructor. See the new `ctor_ub` test for an example.
- For TLS destructors it points at the user-visible span corresponding to where the destructor was registered.

@rust-lang/miri please let me know what you think.

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -576,11 +576,9 @@ fn report_msg<'tcx>(
     }
 
     // Show note and help messages.
-    let mut extra_span = false;
     for (span_data, note) in notes {
         if let Some(span_data) = span_data {
             err.span_note(span_data.span(), note);
-            extra_span = true;
         } else {
             err.note(note);
         }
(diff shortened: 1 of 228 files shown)


saethlin:
> The most tricky part is what to do with errors that point are more than one span.

What I've always wanted here is to do the multi-span diagnostics like the borrow checker uses. If we can pull that off, it may make it clearer because the help spans wouldn't be inserted between the top two components of the backtrace.

saethlin:
> The second commit adds one more span to the bottom of the backtrace for all stacks except the one of `main`. That span indicates why this particular function got called:

I think this change is definitely good, the information is important and if people read every word of the output text it's clear, but we may discover that some additional clarity is needed. I wouldn't rush it though. I like this part of the PR as-is.

RalfJung:
> What I've always wanted here is to do the multi-span diagnostics like the borrow checker uses. If we can pull that off, it may make it clearer because the help spans wouldn't be inserted between the top two components of the backtrace.

We can fairly easily just add multiple labels to the main diagnostic, yeah. However, in my brief experiments, we then lose control over the order in which they are rendered -- they are always shown in source code order. That's not ideal for us, the aliasing model errors work better if we show them in execution order.

RalfJung:
So as a concrete example, if I just replace the `span_help` by `span_label`, here are a data race error and an aliasing model error:
```
error: Undefined Behavior: Data race detected between (1) non-atomic write on thread `unnamed-1` and (2) non-atomic write on thread `unnamed-2` at alloc101
  --> tests/fail/data_race/write_write_race.rs:24:13
   |
19 |             *c.0 = 32;
   |             --------- and (1) occurred earlier here
...
24 |             *c.0 = 64; //~ ERROR: Data race detected between (1) non-atomic write on thread `unnamed-1` and (2) non-atomic write on thread `unnamed-2`
   |             ^^^^^^^^^ (2) just happened here
   |
   = note: this is on thread `unnamed-2`, inside closure
note: which got called indirectly due to this code
  --> tests/fail/data_race/write_write_race.rs:22:18
   |
22 |           let j2 = spawn(move || {
   |  __________________^
23 | |             let c = c; // avoid field capturing
24 | |             *c.0 = 64; //~ ERROR: Data race detected between (1) non-atomic write on thread `unnamed-1` and (2) non-atomic write on thread `unnamed-2`
25 | |         });
   | |__________^
```
```
error: Undefined Behavior: trying to retag from <317> for SharedReadWrite permission at alloc197[0x0], but that tag does not exist in the borrow stack for this location
  --> tests/fail/box-cell-alias.rs:9:14
   |
 9 |     unsafe { (*ptr).set(20) }; //~ ERROR: does not exist in the borrow stack
   |              ^^^^^^ this error occurs as part of retag at alloc197[0x0..0x1]
...
15 |     let ptr: *const Cell<u8> = &*val;
   |                                ----- <317> was created by a SharedReadWrite retag at offsets [0x0..0x1]
16 |     let res = helper(val, ptr);
   |                      --- <317> was later invalidated at offsets [0x0..0x1] by a Unique retag
   |
   = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
   = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
   = note: this is inside `helper`
note: which got called inside `main`
  --> tests/fail/box-cell-alias.rs:16:15
   |
16 |     let res = helper(val, ptr);
   |               ^^^^^^^^^^^^^^^^
```
In this example, the "was later invalidated" works out, but this may be shown above the other two labels.
Furthermore, in general, these multiple spans might not be in the same file; I am not sure what that would look like when rendered.

RalfJung:
@saethlin given those limitations of the borrow-checker-style multi-span diagnostics, what are your thoughts on this PR? Is making things better, is it a wash, or does it make things worse?

@oli-obk do you have any thoughts on this?

saethlin:
🤷 I think this PR is marginally better, but really the way out of this is to change the diagnostic system so that we can make it do what we want.

RalfJung:
I'm not sure I even know what I'd want this to look like -- the diagnostic system is designed around the idea of "showing the relevant part of the source code file", and that starts to fall apart if you want to show things out-of-order. But maybe someone finds a neat way to do this eventually. :)

But in terms of what we do now, can I take this as approval to land the PR @saethlin ?
The one alternative I can think of is to make the entire backtrace more like a traditional backtrace, and entirely forego the diagnostic rendering of spans for that part. That would definitely be a lot more compact. I find it hard to judge how useful showing the sources in the backtrace is.

saethlin:
> But in terms of what we do now, can I take this as approval to land the PR @saethlin ?

Yes.

> The one alternative I can think of is to make the entire backtrace more like a traditional backtrace, and entirely forego the diagnostic rendering of spans for that part. That would definitely be a lot more compact. I find it hard to judge how useful showing the sources in the backtrace is.

I agree on all counts. Maybe one thing we could do is change it for a few weeks and see if anyone complains? :shrug: 
(or ask people to respond somewhere if they like or dislike it?)

RalfJung:
I have pushed a commit that renders the backtrace very similar to the one for panics. What do you think?

oli-obk:
The deep indentation is a bit ugly (especially since the long source paths are 3 levels indented before they start) and yes, sad that we can't nicely do source highlighting 

But it's an improvement in readability either way, so lgtm

RalfJung:
We could use `eprintln!` instead of the diagnostic system to emit the backtrace... though then I am not sure how to best integrate the "origin span" (where the thread got spawned or where the TLS dtor got registered).

oli-obk:
Maybe the diagnostic system should generally just know about backtraces so that json backends can encode them directly instead of just generically as diagnostics and human backends can pretty them

RalfJung:
I tried the eprintln route and I think it is worse:
```
error: Undefined Behavior: interpreting an invalid 32-bit value as a char: 0xffffffff
  --> tests/fail/validity/invalid_char_cast.rs:12:13
   |
12 |             RET = *ptr as u32; //~ERROR: interpreting an invalid 32-bit value as a char
   |             ^^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
   |

stack backtrace:
0: cast
          at tests/fail/validity/invalid_char_cast.rs:12:13: 12:30
1: main
          at tests/fail/validity/invalid_char_cast.rs:20:5: 20:42
```
I don't think we can suppress that empty line.

So for now I'd stick to making it a note. I can reduce the amount of indentation, though it'll remain 3-level.

PR #4757, opened 2025-12-10
Tree Borrows: multiple invalid exposed nodes on main subtree

royAmmerschuber:
If there is no exposed node on the main subtree that allows a child write, then a wildcard access to the tree is invalid.
This wasn't caught before if there were multiple exposed nodes in the main subtree whose tag was larger than the accessed tag.

### DIFF
--- a/src/borrow_tracker/tree_borrows/tree.rs
+++ b/src/borrow_tracker/tree_borrows/tree.rs
@@ -111,7 +111,7 @@ impl LocationState {
             // We need to update the wildcard state, if the permission
             // of an exposed pointer changes.
             if node.is_exposed {
-                let access_type = self.permission.strongest_allowed_child_access(protected);
+                let access_type = self.permission.strongest_allowed_local_access(protected);
                 WildcardState::update_exposure(idx, access_type, nodes, wildcard_accesses);
             }
         }
@@ -1034,6 +1034,9 @@ impl<'tcx> LocationTree {
             wildcard_state.access_relatedness(access_kind, only_foreign)
         };
 
+        // Whether there is an exposed node in this tree that allows this access.
+        let mut has_valid_exposed = false;
+
         // This does a traversal across the tree updating children before their parents. The
         // difference to `perform_normal_access` is that we take the access relatedness from
         // the wildcard tracking state of the node instead of from the visitor itself.
@@ -1082,6 +1085,17 @@ impl<'tcx> LocationTree {
                     return Err(no_valid_exposed_references_error(diagnostics));
                 };
 
+                let mut entry = args.data.perms.entry(args.idx);
+                let perm = entry.or_insert(node.default_location_state());
+
+                // We only count exposed nodes through which an access could happen.
+                if node.is_exposed
+                    && perm.permission.strongest_allowed_local_access(protected).allows(access_kind)
+                    && max_local_tag.is_none_or(|max_local_tag| max_local_tag >= node.tag)
+                {
+                    has_valid_exposed = true;
+                }
+
                 let Some(relatedness) = wildcard_relatedness.to_relatedness() else {
                     // If the access type is Either, then we do not apply any transition
                     // to this node, but we still update each of its children.
@@ -1090,8 +1104,6 @@ impl<'tcx> LocationTree {
                     return Ok(());
                 };
 
-                let mut entry = args.data.perms.entry(args.idx);
-                let perm = entry.or_insert(node.default_location_state());
                 // We know the exact relatedness, so we can actually do precise checks.
                 perm.perform_transition(
                     args.idx,
@@ -1115,6 +1127,21 @@ impl<'tcx> LocationTree {
                 })
             },
         )?;
+        // If there is no exposed node in this tree that allows this access, then the
+        // access *must* be foreign. So we check if the root of this tree would allow this
+        // as a foreign access, and if not, then we can error.
+        // In practice, all wildcard trees accept foreign accesses, but the main tree does
+        // not, so this catches UB when none of the nodes in the main tree allows this access.
+        if !has_valid_exposed
+            && self
+                .wildcard_accesses
+                .get(root)
+                .unwrap()
+                .access_relatedness(access_kind, /* only_foreign */ true)
+                .is_none()
+        {
+            return Err(no_valid_exposed_references_error(diagnostics)).into();
+        }
         interp_ok(())
     }
 }
--- a/src/borrow_tracker/tree_borrows/wildcard.rs
+++ b/src/borrow_tracker/tree_borrows/wildcard.rs
@@ -20,6 +20,16 @@ pub enum WildcardAccessLevel {
     Read,
     Write,
 }
+impl WildcardAccessLevel {
+    /// Weather this access kind is allowed at this level.
+    pub fn allows(self, kind: AccessKind) -> bool {
+        let required_level = match kind {
+            AccessKind::Read => Self::Read,
+            AccessKind::Write => Self::Write,
+        };
+        required_level <= self
+    }
+}
 
 /// Where the access happened relative to the current node.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
@@ -430,7 +440,7 @@ impl Tree {
                 .map(|p| p.permission())
                 .unwrap_or_else(|| node.default_location_state().permission());
 
-            let access_type = perm.strongest_allowed_child_access(protected);
+            let access_type = perm.strongest_allowed_local_access(protected);
             WildcardState::update_exposure(
                 id,
                 access_type,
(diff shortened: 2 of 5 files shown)


src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1082,6 +1084,18 @@ impl<'tcx> LocationTree {
...
                 };
 
+                let mut entry = args.data.perms.entry(args.idx);
+                let perm = entry.or_insert(node.default_location_state());
+
+                // We only count exposed nodes through which an access could happen.
+                if node.is_exposed
+                    && perm.permission.strongest_allowed_child_access(protected)
+                        >= access_kind.into()
+                    && max_local_tag.is_none_or(|max_local_tag| max_local_tag > node.tag)
```
```suggestion
                    && max_local_tag.is_none_or(|max_local_tag| max_local_tag >= node.tag)
```
The max-local-tag itself can be the souce of local accesses, right?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1115,6 +1127,19 @@ impl<'tcx> LocationTree {
...
+        // If this is the main subtree and it doesn't contain any exposed children through which the access could happen then the access
+        // is invalid.
+        if !has_exposed
+            // Check that this is the main subtree.
+            && self
+                .wildcard_accesses
+                .get(root)
+                .unwrap()
+                .access_relatedness(access_kind, true)
+                .is_none()
```
RalfJung: How does this check that it is the main tree...?

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -20,6 +20,14 @@ pub enum WildcardAccessLevel {
     Read,
     Write,
 }
+impl From<AccessKind> for WildcardAccessLevel {
```
RalfJung: If we have to start converting between them, there's not really any point any more in them being different types.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1082,6 +1084,18 @@ impl<'tcx> LocationTree {
                     return Err(no_valid_exposed_references_error(diagnostics));
                 };
 
+                let mut entry = args.data.perms.entry(args.idx);
+                let perm = entry.or_insert(node.default_location_state());
+
+                // We only count exposed nodes through which an access could happen.
+                if node.is_exposed
+                    && perm.permission.strongest_allowed_child_access(protected)
```
RalfJung: We should probably rename that method to include `wildcard`; it's currently confusing when one reads this code and tries to understand it without knowing exactly by heart what each function does.

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber: `max_local_tag` is actually never equal to `node.tag` as we only call `perform_wildcard_access` on the trees which don't contain `max_local_tag`. But `>=` is more fitting for the name of `max_local_tag` so ill change this.

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber: This checks if no foreign access is possible on the root of the subtree. Only the main subtree doesn't allow foreign accesses, so if `access_relatedness()` is None then root is the main root.
I should add a comment to the bool flag (`/* only_foreign */`). Or I could make `max_foreign_access` public and read that directly.  

This way it's implemented technically makes it generic to all subtrees, But the only one it actually effects is the main subtree.

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber: I forgot to rename this function to `strongest_allowed_local_access` which would be the correct term. I think the function name (at least with this change) already tells you exactly what the function does.

src/borrow_tracker/tree_borrows/wildcard.rs
royAmmerschuber: We could replace `WildcardAccessLevel` with `Option<AccessKind>`. We would have to implement `Ord` for `AccessKind` and rely on `Option`'s `Ord` implementation, but I think that would still be the correct logic.

I do think it makes sense to keep the types separate since they represent different things. Access kind is the type of a concrete access, while access level represents the highest allowed access that is possible.

I only use this implementation to check if an access kind is allowed under a specific access level. Maybe I should replace this with a method `access_level.allows(access_kind)`.

royAmmerschuber:
> However, is there any way to generalize this so it doesn't just work for the main tree? Currently it looks like a pretty ad-hoc special case, and there are probably many of those that we could add.

This PR is trying to better approximate the ideal version of the algorithm (where we recalculate the wildcard data structure before each access).
It should now return a `no_valid_exposed_references_error()` in the same cases as the ideal algorithm. It doesn't cover the cases where a `TransitionError` would get returned, but for these there is no easy way to check them, so we would probably have to implement the ideal algorithm directly.

As for this only being implemented for the main subtree. This error already never gets raised on the other subtrees. An access is UB if there exist no valid exposed nodes on the main subtree, even if there are valid exposed nodes on other subtrees. This is because those other subtrees would also have to be attached to one of the valid exposed nodes on the main subtree.


> I'm also confused by the PR title. This is about whether there's any eligible exposed node on the main tree, right? Why do you talk about "multiple"?

> Are you saying this was caught if there was exactly one exposed node in the main subtree whose tag was larger than the accessed tag? How?

Yes, it's already handled for the single invalid exposed case (see test case `cross_tree_update_main_invalid_exposed.rs`). In the single case when we try to calculate the relatedness for the invalid exposed node we know that neither a local (because it's invalid) nor a foreign (because it's the only exposed node on the main subtree) access could happen so its UB.
However, if this node has an invalid exposed cousin, then it thinks a foreign access could have happened from that cousin.

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: The main problem here is that comment ant code don't match. Given that code, the comment should explain why this is correct for *any* tree that does not allow foreign accesses at the root.

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: Hm, I see... but then it is somewhat confusing that it returns a `WildcardAccess`.

src/borrow_tracker/tree_borrows/wildcard.rs
RalfJung: > I only use this implementation to check if an access kind is allowed under a specific access level. Maybe I should replace this with a method access_level.allows(access_kind).

That would also be reasonable.

RalfJung:
> This PR is trying to better approximate the ideal version of the algorithm (where we recalculate the wildcard data structure before each access).
It should now return a no_valid_exposed_references_error() in the same cases as the ideal algorithm.

I like that way of framing it, thanks. However, I would say the ideal (or rather, idealized) algorithm would just not have the data structure at all and do nested tree traversals.

This should be reflected in the comments.

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber: We could also rename `WildcardAccessLevel` to just `AccessLevel` or `MaxAccessLevel`.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1115,6 +1126,23 @@ impl<'tcx> LocationTree {
...
             },
         )?;
+        // For a wildcard access to be valid each subtree needs to either contain an exposed tag
+        // through which the access could have happened or a foreign access to the subtree must
+        // be possible. If neither of these is the case than the access is UB.
+        // In reality this is only ever UB on the main subtree as all other trees always allow
+        // foreign accesses.
+        if !has_exposed
+            // Check that no access that is foreign to this subtree is possible.
+            // (Its only impossible for the main subtree).
```
```suggestion
        // If there is no exposed node in this tree that allows this access, then the
        // access *must* be foreign. So we check if the root of this tree would allow this
        // as a foreign access, and if not, then we can error.
        // In practice, all wildcard trees accept foreign accesses, but the main tree does
        // not, so this catches UB when none of the nodes in the main tree allows this access.
        if !has_exposed
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1034,6 +1034,8 @@ impl<'tcx> LocationTree {
             wildcard_state.access_relatedness(access_kind, only_foreign)
         };
 
+        let mut has_exposed = false;
```
RalfJung: This isn't just about whether there is an exposed node, it's about whether there's an exposed node *that allows this access*. The name should at least indicate that, and there should be a comment here explaining that.

RalfJung:

@rustbot author

PR #4759, opened 2025-12-11
gitconfig is not a toml file

RalfJung:
This led to some odd syntax highlighting.


bjorn3:
Maybe `ini` would be a better match?

RalfJung:
:shrug: syntax highlighting is anyway mostly superfluous on that 2-line snippet

PR #4761, opened 2025-12-12
Implement FlushFileBuffers

CraftSpider:
Windows shim needed to support sync_data/sync_all


src/shims/windows/fs.rs
```diff
@@ -457,6 +457,39 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            this.invalid_handle("FlushFileBuffers")?
+        };
+        let file = desc.downcast::<FileHandle>().ok_or_else(|| {
+            err_unsup_format!(
+                "`FlushFileBuffers` is only supported on file-backed file descriptors"
+            )
+        })?;
+
+        if !file.writable {
+            this.set_last_error(IoError::WindowsError("ERROR_ACCESS_DENIED"))?;
```
RalfJung, line 480: So this matches the error code that Windows sets in this situation?

CraftSpider: Yes, I based this on a quick local test of calling `sync_all` on a read-only file.

RalfJung:
Looks great, thanks!

PR #4766, opened 2025-12-16
Tree Borrows: improve protector end access child skipping

royAmmerschuber:
Currently, if we do a protector release access we assume that any other subtree (including the main subtree) could be a child of the accessed node, meaning we conservatively don't update them based on the access.

This PR determines as accurately as possible (under the current model), which other subtrees could be children of the accessed tag.
This means we now accurately reflect:
- the main subtree is never a child of the accessed tag
- if the accessed tag doesn't have exposed children, then no other subtree is a child of it
- a subtree can only be a child of the accessed tag, if there exists an exposed child of the accessed tag with a tag less than the root of the subtree

### DIFF
--- a/src/borrow_tracker/tree_borrows/tree.rs
+++ b/src/borrow_tracker/tree_borrows/tree.rs
@@ -661,6 +661,7 @@ impl<'tcx> Tree {
                 global,
                 ChildrenVisitMode::VisitChildrenOfAccessed,
                 &diagnostics,
+                /* min_exposed_child */ None, // only matters for protector end access,
             )?;
         }
         interp_ok(())
@@ -685,6 +686,13 @@ impl<'tcx> Tree {
         }
 
         let source_idx = self.tag_mapping.get(&tag).unwrap();
+
+        let min_exposed_child = if self.roots.len() > 1 {
+            LocationTree::get_min_exposed_child(source_idx, &self.nodes)
+        } else {
+            // There's no point in computing this when there is just one tree.
+            None
+        };
 
         // This is a special access through the entire allocation.
         // It actually only affects `accessed` locations, so we need
@@ -716,6 +724,7 @@ impl<'tcx> Tree {
                     global,
                     ChildrenVisitMode::SkipChildrenOfAccessed,
                     &diagnostics,
+                    min_exposed_child,
                 )?;
             }
         }
@@ -876,9 +885,36 @@ impl Tree {
 }
 
 impl<'tcx> LocationTree {
+    /// Returns the smallest exposed tag, if any, that is a transitive child of `root`.
+    fn get_min_exposed_child(root: UniIndex, nodes: &UniValMap<Node>) -> Option<BorTag> {
+        // We cannot use the wildcard datastructure to improve this lookup. This is because
+        // the datastructure only tracks enabled nodes and we need to also consider disabled ones.
+        let mut stack = vec![root];
+        let mut min_tag = None;
+        while let Some(idx) = stack.pop() {
+            let node = nodes.get(idx).unwrap();
+            if min_tag.is_some_and(|min| min < node.tag) {
+                // The minimum we found before is bigger than this tag, and therefore
+                // also bigger than all its children, so we can skip this subtree.
+                continue;
+            }
+            stack.extend_from_slice(node.children.as_slice());
+            if node.is_exposed {
+                min_tag = match min_tag {
+                    Some(prev) if prev < node.tag => Some(prev),
+                    _ => Some(node.tag),
+                };
+            }
+        }
+        min_tag
+    }
+
     /// Performs an access on this location.
     /// * `access_source`: The index, if any, where the access came from.
     /// * `visit_children`: Whether to skip updating the children of `access_source`.
+    /// * `min_exposed_child`: The tag of the smallest exposed (transitive) child of the accessed node.
+    ///   This is only used with `visit_children == SkipChildrenOfAccessed`, where we need to skip children
+    ///   of the accessed node.
     fn perform_access(
         &mut self,
         roots: impl Iterator<Item = UniIndex>,
@@ -888,6 +924,7 @@ impl<'tcx> LocationTree {
         global: &GlobalState,
         visit_children: ChildrenVisitMode,
         diagnostics: &DiagnosticInfo,
+        min_exposed_child: Option<BorTag>,
     ) -> InterpResult<'tcx> {
         let accessed_root = if let Some(idx) = access_source {
             Some(self.perform_normal_access(
@@ -906,11 +943,22 @@ impl<'tcx> LocationTree {
         };
 
         let accessed_root_tag = accessed_root.map(|idx| nodes.get(idx).unwrap().tag);
-        if matches!(visit_children, ChildrenVisitMode::SkipChildrenOfAccessed) {
-            // FIXME: approximate which roots could be children of the accessed node and only skip them instead of all other trees.
-            return interp_ok(());
-        }
         for root in roots {
+            let tag = nodes.get(root).unwrap().tag;
+            // On a protector release access we have to skip the children of the accessed tag.
+            // However, if the tag has exposed children then some of the wildcard subtrees could
+            // also be children of the accessed node and would also need to be skipped. We can
+            // narrow down which wildcard trees might be children by comparing their root tag to the
+            // minimum exposed child of the accessed node. As the parent tag is always smaller
+            // than the child tag this means we only need to skip subtrees with a root tag larger
+            // than `min_exposed_child`. Once we find such a root, we can leave the loop because roots
+            // are sorted by tag.
+            if matches!(visit_children, ChildrenVisitMode::SkipChildrenOfAccessed)
+                && let Some(min_exposed_child) = min_exposed_child
+                && tag > min_exposed_child
+            {
+                break;
+            }
             // We don't perform a wildcard access on the tree we already performed a
             // normal access on.
             if Some(root) == accessed_root {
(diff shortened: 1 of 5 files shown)


src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -661,6 +661,7 @@ impl<'tcx> Tree {
                 global,
                 ChildrenVisitMode::VisitChildrenOfAccessed,
                 &diagnostics,
+                None, // min_exposed_child only matters for protector end access,
```
```suggestion
                /* min_exposed_child */ None, // only matters for protector end access,
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -686,6 +687,12 @@ impl<'tcx> Tree {
 
         let source_idx = self.tag_mapping.get(&tag).unwrap();
 
+        let min_exposed_child = if self.roots.len() > 1 {
+            LocationTree::get_min_exposed_child(source_idx, &self.nodes)
+        } else {
+            None
```
RalfJung, line 694: Is this correct?
```suggestion
            // There's no point in computing this when there is just one tree.
            None
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -876,9 +884,35 @@ impl Tree {
...
+    /// Returns the smallest exposed tag, if any, that is a transitive child of `root`.
+    fn get_min_exposed_child(root: UniIndex, nodes: &UniValMap<Node>) -> Option<BorTag> {
+        // We cannot use the wildcard datastructure to improve this lookup. This is because
+        // the datastructure only tracks enabled nodes and we need to also consider disabled ones.
+        let mut stack = vec![root];
+        let mut min_tag = None;
+        while let Some(idx) = stack.pop() {
+            let node = nodes.get(idx).unwrap();
+            if min_tag.is_some_and(|min| min < node.tag) {
+                continue;
```
RalfJung, line 899:
```suggestion
                // The minimum we found before is bigger than this tag, and therefore
                // also bigger than all its children, so we can skip this subtree.
                continue;
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -906,11 +941,21 @@ impl<'tcx> LocationTree {
...
+            // On a protector release access we skip the children of the accessed tag so that
+            // we can correctly return references from functions. However, if the tag has
+            // exposed children then some of the wildcard subtrees could also be children of
+            // the accessed node and would also need to be skipped. We can narrow down which
+            // child trees are children by comparing their root tag to the minimum exposed
+            // child of the accessed node. As the parent tag is always smaller than the child
+            // tag this means we only need to skip subtrees with a root tag larger than
+            // `min_exposed_child`.
+            if let Some(min_exposed_child) = min_exposed_child
+                && tag > min_exposed_child
```
RalfJung: Shouldn't this condition check `visit_children`?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -906,11 +941,21 @@ impl<'tcx> LocationTree {
...
 
         let accessed_root_tag = accessed_root.map(|idx| nodes.get(idx).unwrap().tag);
-        if matches!(visit_children, ChildrenVisitMode::SkipChildrenOfAccessed) {
-            // FIXME: approximate which roots could be children of the accessed node and only skip them instead of all other trees.
-            return interp_ok(());
-        }
         for root in roots {
+            let tag = nodes.get(root).unwrap().tag;
+            // On a protector release access we skip the children of the accessed tag so that
+            // we can correctly return references from functions. However, if the tag has
```
```suggestion
            // On a protector release access we have to skip the children of the accessed tag.
            // However, if the tag has
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -906,11 +941,21 @@ impl<'tcx> LocationTree {
...
+            // exposed children then some of the wildcard subtrees could also be children of
+            // the accessed node and would also need to be skipped. We can narrow down which
+            // child trees are children by comparing their root tag to the minimum exposed
```
```suggestion
            // wildcard trees might be children by comparing their root tag to the minimum exposed
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -906,11 +941,21 @@ impl<'tcx> LocationTree {
...
+            // child of the accessed node. As the parent tag is always smaller than the child
+            // tag this means we only need to skip subtrees with a root tag larger than
+            // `min_exposed_child`.
```
```suggestion
            // `min_exposed_child`. Once we find such a root, we can leave the loop because roots
            // are sorted by tag.
```

tests/fail/tree_borrows/wildcard/protector_release.rs
```diff
@@ -0,0 +1,77 @@
+//@compile-flags: -Zmiri-tree-borrows -Zmiri-permissive-provenance
+use std::cell::UnsafeCell;
+
+/// This is a variant of the test in `../protector-write-lazy.rs`, but with
+/// wildcard references.
+/// Checks that a protector release transitions a foreign reference on a reference,
```
RalfJung: What does "transitions a foreign reference on a reference" mean?

tests/fail/tree_borrows/wildcard/protector_release.rs
```diff
@@ -0,0 +1,77 @@
...
+/// if we can determine that it's not a child of the released tag.
+/// For this version we know the tag is not a child, because its wildcard root has
```
RalfJung:
```suggestion
/// For this version we know the tag is not a child because its wildcard root has
```

tests/fail/tree_borrows/wildcard/protector_release.rs
```diff
@@ -0,0 +1,77 @@
...
+        //          ▼                 ▼
+        //    ┌───────────┐     ┌───────────┐
+        //    │ ref2      │     │    ref3   │
+        //    │ Act │ Res │     │ Cel │ Cel │
+        //    └─────┬─────┘     └─────┬─────┘
+        //          │                 │
+        //          │                 │
+        //          ▼                 ▼
+        //    ┌───────────┐     ┌───────────┐
+        //    │ arg4      │     │       ref6│
```
RalfJung: Does the horizontal alignment of `ref6` mean anything?

tests/fail/tree_borrows/wildcard/protector_release2.rs
RalfJung, line 1: This test seems very similar to the previous one, why did you add it as a separate test? Is there some plausible bug in the code that could make a difference between these tests?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -876,9 +884,35 @@ impl Tree {
...
+        min_tag
+    }
+
     /// Performs an access on this location.
     /// * `access_source`: The index, if any, where the access came from.
     /// * `visit_children`: Whether to skip updating the children of `access_source`.
+    /// * `min_exposed_child`: any subtree that could be a child of `access_source` must
+    ///   have a root with a larger tag than this. If None then no other subtree can be
+    ///   a child of `access_source`.
+    ///   This only gets read if `visit_children==SkipChildrenOfAccessed`.
```
RalfJung: The comment doesn't match the name. What about something like this?
```suggestion
    /// * `min_exposed_child`: The tag of the smallest exposed (transitive) child of the accessed node.
    ///   This is useful with `visit_children == SkipChildrenOfAccessed`, where we need to skip children
    ///   of the accessed.
```

RalfJung (changes requested):
Mostly comments about comments, as usual. :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4766/088c054a78bacc0490e2c4116f274621bfd51e1e..69336b4d03945fc9b698128aea7c2338f3cd7f20)*

tests/fail/tree_borrows/wildcard/protector_release.rs
royAmmerschuber: ref6 was created as a reference to the second element of the array. arg4 to the first element

tests/fail/tree_borrows/wildcard/protector_release2.rs
royAmmerschuber: test 1 would also be satisfied by the simpler version, where we don't calculate `min_exposed_tag` and instead just use the tag of the accessed node.

tests/fail/tree_borrows/wildcard/protector_release.rs
RalfJung: Ah :) Maybe extend the vertical line upwards to make that more clear?

tests/fail/tree_borrows/wildcard/protector_release2.rs
```diff
@@ -0,0 +1,77 @@
...
+        // This should cause a foreign write on all foreign nodes,
+        // unless they could be a child of arg3.
+        // arg3 has an exposed child, so some tags with a wildcard
+        // ancestor could be its children.
+        // However, the root of ref6 was created before the exposed
+        // child ref5, so it cannot be a child of it. So it should
+        // get disabled (at location [0]).
+    };
+
+    let ptr = protect(ref2);
```
RalfJung, line 78: `ptr` does not show up in the diagram... which pointer is that?

tests/fail/tree_borrows/wildcard/protector_release2.rs
RalfJung: Okay. Please add that to the comment for test 2.

tests/fail/tree_borrows/wildcard/protector_release2.rs
royAmmerschuber: I think you're reviewing old code. I added a comment explaining it has provenance of `ref6` in the last commit.

RalfJung: You are right. :)

RalfJung:

@rustbot author

Issue #4725, opened 2025-11-25
Simplify libc-based tests with shared helpers


RalfJung:
Many of our libc-based tests are enormously verbose: they directly interact with the libc APIs, which is cumbersome from Rust. (They also often let-bind way more than is necessary or useful.)

The goal of this issue is to make the tests more concise and therefore more readable and easier to maintain. https://github.com/rust-lang/miri/pull/4724 and https://github.com/rust-lang/miri/pull/4721 and good examples for the kinds of things we can do:
- use C string literals instead of manually mucking about with `CString` or manually null-terminating byte slice literals
- use `errno_result` or `errno_check` instead of manually checking the result of functions that report their error via errno
- use `read_all_into_array` instead of `read`/`read_all`, and `write_all_from_slice` instead of `write`/`write_all` (this won't always work!)
- use the epoll helpers in the epoll tests (I only ported a few of them)

hulxv:
I would like to work on this :)

RalfJung:
Okay, great! Please change only 1 file (or a few small files) per PR. Also, make sure to look up the documentation of the involved libc function to understand how they report errors.

RalfJung:
A lot of progress has been made here, but there are still some things left to be done:


- [ ] `tests/pass-dep/libc/libc-fs.rs` still contains a bunch of stuff that could be cleaned up:
  - many calls to `libc::open`, `libc::dup`, `libc::dup2` that should be wrapped in `errno_result`
  - more than a dozen manual comparisons of a return value with `-1` that should probably also all be replaced by `errno_result` or `errno_check`

- [x] `tests/pass-dep/libc/libc-eventfd.rs` contains a bunch of `libc::write`. it also contains its own odd `read_bytes`/`write_bytes` helpers. that should probably all be `read_exact`/`write_all`.

beyond that file there are others that also could make better use of our utils:
- [x] `tests/fail-dep/libc/eventfd_block_read_twice.rs`
- [x] `tests/fail-dep/libc/eventfd_block_write_twice.rs`
- [x] `tests/fail-dep/libc/socketpair_block_read_twice.rs`
- [x] `tests/fail-dep/libc/socketpair_block_write_twice.rs`
- [x] `tests/fail-dep/libc/socketpair-data-race.rs`

PR #4768, opened 2025-12-17
Refactor libc pipe tests to use utility functions for error handling and data operations

hulxv:

### DIFF
--- a/tests/pass-dep/libc/libc-pipe.rs
+++ b/tests/pass-dep/libc/libc-pipe.rs
@@ -38,13 +38,11 @@ fn test_pipe() {
     let data = b"123";
     write_all_from_slice(fds[1], data).unwrap();
     let mut buf4: [u8; 5] = [0; 5];
-    let res = read_into_slice(fds[0], &mut buf4).unwrap().0.len();
-    assert_eq!(buf4[..res], data[..res]);
-    if res < 3 {
-        // Drain the rest from the read end.
-        let res = read_into_slice(fds[0], &mut buf4[res..]).unwrap().0.len();
-        assert!(res > 0);
-    }
+    let (part1, rest) = read_into_slice(fds[0], &mut buf4).unwrap();
+    assert_eq!(part1[..], data[..part1.len()]);
+    // Write 2 more bytes so we can exactly fill the `rest`.
+    write_all_from_slice(fds[1], b"34").unwrap();
+    read_all_into_slice(fds[0], rest).unwrap();
 }
 
 fn test_pipe_threaded() {
--- a/tests/utils/libc.rs
+++ b/tests/utils/libc.rs
@@ -40,19 +40,8 @@ pub unsafe fn read_all(
     return read_so_far as libc::ssize_t;
 }
 
-/// Read exactly `N` bytes from `fd`. Error if that many bytes could not be read.
+/// Try to fill the given slice by reading from `fd`. Error if that many bytes could not be read.
 #[track_caller]
-pub fn read_all_into_array<const N: usize>(fd: libc::c_int) -> Result<[u8; N], libc::ssize_t> {
-    let mut buf = [0; N];
-    let res = unsafe { read_all(fd, buf.as_mut_ptr().cast(), buf.len()) };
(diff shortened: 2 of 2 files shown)


tests/pass-dep/libc/libc-pipe.rs
```diff
@@ -25,24 +26,17 @@ fn main() {
...
 
     // Read size > data available in buffer.
-    let data = "123".as_bytes();
-    let res = unsafe { libc_utils::write_all(fds[1], data.as_ptr() as *const libc::c_void, 3) };
-    assert_eq!(res, 3);
+    let data = b"123";
+    write_all_from_slice(fds[1], data).unwrap();
     let mut buf4: [u8; 5] = [0; 5];
     let res = unsafe { libc::read(fds[0], buf4.as_mut_ptr().cast(), buf4.len() as libc::size_t) };
     assert!(res > 0 && res <= 3);
```
RalfJung: We should think about a way to also cover this case without having to directly invoke `read` (or `read_all` below).

I am imagining functions like this:
```rust
/// Fill `dest` by repeatedly reading from `fd`. Error if not enough bytes could be read.
fn read_all_into_slice(fd: libc::c_int, dest: &mut [u8]) -> Result<(), libc::ssize_t>;

/// Read up to `dest.len()` many bytes from `fd`. Returns a pair of the part of `dest` that has been
/// read into, and the tail that was left untouched.
fn read_into_slice(fd: libc::c_int, dest: &mut [u8]) -> Result<(&mut [u8], &mut [u8]), libc::ssize_t>;
```
Can you add such functions to `tests/utils/libc.rs` and use them here?

(Please let me know if this is too much, then I can take over for this part.)

tests/pass-dep/libc/libc-pipe.rs
```diff
@@ -57,37 +51,25 @@ fn test_pipe() {
...
-        assert_eq!(buf, "abcde".as_bytes());
+        let buf = read_all_into_array::<5>(fds[0]).unwrap();
+        assert_eq!(&buf, b"abcde");
     });
     thread::yield_now();
-    let data = "abcde".as_bytes().as_ptr();
-    let res = unsafe { libc_utils::write_all(fds[1], data as *const libc::c_void, 5) };
-    assert_eq!(res, 5);
+    let data = b"abcde";
+    write_all_from_slice(fds[1], data).unwrap();
```
RalfJung:
```suggestion
    write_all_from_slice(fds[1], b"abcde").unwrap();
```
Generally, if `data` is only used exactly once, then just inline it. This occurs some more times below.

tests/pass-dep/libc/libc-pipe.rs
```diff
@@ -5,6 +5,7 @@ use std::thread;
 
 #[path = "../../utils/libc.rs"]
 mod libc_utils;
+use libc_utils::{errno_check, errno_result, read_all_into_array, write_all_from_slice};
```
RalfJung:
```suggestion
use libc_utils::*;
```

tests/pass-dep/libc/libc-pipe.rs
hulxv: It seems clear. I will do changes very soon because currently I am busy with my exams but don't worry, I will do it!

Do you want to use these functions instead of `read` and `read_all` in all libc tests? also maybe adding them to all PRs do some conflicts so is there a way to avoid this? maybe open a separate pr?

RalfJung: This PR should use functions like that for all `read` in the pipe tests. Once this lands, the function is then available for the other PRs, so that should avoid conflicts.

RalfJung:
I can't tell if you intend to make more changes here or not -- as [indicated above](https://github.com/rust-lang/miri/pull/4768#issuecomment-3683772814), I will wait until you mark this PR as ready before I do another review.

hulxv:
Sorry, It seems the test fails without noticing it. it's fixed now

RalfJung:
Thanks! I made some further small tweaks, this is good to land now. :)

PR #4769, opened 2025-12-17
Replace assertions with errno_check for libc::strerror_r calls

hulxv:

### DIFF
--- a/tests/pass-dep/libc/libc-strerror_r.rs
+++ b/tests/pass-dep/libc/libc-strerror_r.rs
@@ -1,11 +1,15 @@
 //@ignore-target: windows # Supported only on unixes
 
+#[path = "../../utils/libc.rs"]
+mod libc_utils;
+use libc_utils::errno_check;
+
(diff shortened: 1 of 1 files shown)


tests/pass-dep/libc/libc-strerror_r.rs
```diff
@@ -1,11 +1,15 @@
 //@ignore-target: windows # Supported only on unixes
 
+#[path = "../../utils/libc.rs"]
+mod libc_utils;
+use libc_utils::errno_check;
+
 fn main() {
     unsafe {
         let mut buf = vec![0u8; 32];
-        assert_eq!(libc::strerror_r(libc::EPERM, buf.as_mut_ptr().cast(), buf.len()), 0);
```
RalfJung, line 6: `strerror_r` is not a usual function with errno error handling so I don't think we should use `errno_check` here. I think this test is fine as-is, it doesn't need any helpers.

Thanks for helping us check for this!

PR #4771, opened 2025-12-17
Refactor CPU affinity tests to use errno_check for error handling

hulxv:


RalfJung:
As discussed in the other PRs, marking as "waiting for author" until the first PRs landed, to avoid repeating the same review comments. We will get them all landed if you stick around, we should just take things one after the other. :)
@rustbot author

RalfJung:
LGTM, thanks!

PR #4772, opened 2025-12-17
Refactor epoll tests to use errno_result and improve notification checks

hulxv:

### DIFF
--- a/tests/pass-dep/libc/libc-epoll-blocking.rs
+++ b/tests/pass-dep/libc/libc-epoll-blocking.rs
@@ -21,9 +21,6 @@ fn main() {
     multiple_events_wake_multiple_threads();
 }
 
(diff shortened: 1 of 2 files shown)


RalfJung:
As discussed in the other PRs, marking as "waiting for author" until the first PRs landed, to avoid repeating the same review comments. We will get them all landed if you stick around, we should just take things one after the other. :)
@rustbot author

RalfJung:
Thanks! This is a good first step for these tests. However, they are not done yet: the `check_epoll_wait` function in "no-blocking.rs" should be removed and replaced by the one from utils; there are many calls to `write_all` that should use one of the fancier helpers; there are a bunch more opportunities to use `epoll_ctl_add`. This can be done in a follow-up PR. :)

PR #4776, opened 2025-12-18
add `vpdpbusd` avx512 intrinsic

folkertdev:
This intrinsic is useful for the adler32 checksum algorithm. 

The test attempts to hit a bunch of overflow and truncation cases, and I've validated it on real hardware.

### DIFF
--- a/src/shims/x86/avx512.rs
+++ b/src/shims/x86/avx512.rs
@@ -109,8 +109,66 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
                 pshufb(this, left, right, dest)?;
             }
+
+            // Used to implement the _mm512_dpbusd_epi32 function.
+            "vpdpbusd.512" | "vpdpbusd.256" | "vpdpbusd.128" => {
+                this.expect_target_feature_for_intrinsic(link_name, "avx512vnni")?;
+                if matches!(unprefixed_name, "vpdpbusd.128" | "vpdpbusd.256") {
+                    this.expect_target_feature_for_intrinsic(link_name, "avx512vl")?;
+                }
+
+                let [src, a, b] = this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
+
+                vpdpbusd(this, src, a, b, dest)?;
+            }
             _ => return interp_ok(EmulateItemResult::NotSupported),
         }
         interp_ok(EmulateItemResult::NeedsReturn)
     }
+}
+
+/// Multiply groups of 4 adjacent pairs of unsigned 8-bit integers in `a` with corresponding signed
+/// 8-bit integers in `b`, producing 4 intermediate signed 16-bit results. Sum these 4 results with
+/// the corresponding 32-bit integer in `src` (using wrapping arighmetic), and store the packed
+/// 32-bit results in `dst`.
+///
+/// <https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_dpbusd_epi32>
+/// <https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_dpbusd_epi32>
+/// <https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_dpbusd_epi32>
+fn vpdpbusd<'tcx>(
+    ecx: &mut crate::MiriInterpCx<'tcx>,
+    src: &OpTy<'tcx>,
+    a: &OpTy<'tcx>,
+    b: &OpTy<'tcx>,
+    dest: &MPlaceTy<'tcx>,
+) -> InterpResult<'tcx, ()> {
+    let (src, src_len) = ecx.project_to_simd(src)?;
+    let (a, a_len) = ecx.project_to_simd(a)?;
+    let (b, b_len) = ecx.project_to_simd(b)?;
+    let (dest, dest_len) = ecx.project_to_simd(dest)?;
+
+    // fn vpdpbusd(src: i32x16, a: i32x16, b: i32x16) -> i32x16;
+    // fn vpdpbusd256(src: i32x8, a: i32x8, b: i32x8) -> i32x8;
+    // fn vpdpbusd128(src: i32x4, a: i32x4, b: i32x4) -> i32x4;
+    assert_eq!(dest_len, src_len);
+    assert_eq!(dest_len, a_len);
+    assert_eq!(dest_len, b_len);
+
+    for i in 0..dest_len {
+        let src = ecx.read_scalar(&ecx.project_index(&src, i)?)?.to_i32()?;
+        let a = ecx.read_scalar(&ecx.project_index(&a, i)?)?.to_u32()?;
+        let b = ecx.read_scalar(&ecx.project_index(&b, i)?)?.to_u32()?;
+        let dest = ecx.project_index(&dest, i)?;
+
+        let zipped = a.to_le_bytes().into_iter().zip(b.to_le_bytes());
+        let intermediate_sum: i32 = zipped
+            .map(|(a, b)| i32::from(a).strict_mul(i32::from(b.cast_signed())))
+            .fold(0, |x, y| x.strict_add(y));
+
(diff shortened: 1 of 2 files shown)


src/shims/x86/mod.rs
```diff
@@ -1200,6 +1200,53 @@ fn pshufb<'tcx>(
...
+        let b = ecx.read_scalar(&ecx.project_index(&b, i)?)?.to_u32()?;
+        let dest = ecx.project_index(&dest, i)?;
+
+        let [a1, a2, a3, a4] = a.to_le_bytes();
+        let [b1, b2, b3, b4] = b.to_le_bytes();
+
+        let intermediate = i32::from(i16::from(a1).wrapping_mul(i16::from(b1 as i8)))
+            .wrapping_add(i32::from(i16::from(a2).wrapping_mul(i16::from(b2 as i8))))
+            .wrapping_add(i32::from(i16::from(a3).wrapping_mul(i16::from(b3 as i8))))
+            .wrapping_add(i32::from(i16::from(a4).wrapping_mul(i16::from(b4 as i8))));
```
folkertdev: the `as i8` is intentional here so that sign extension is used. So `try_from` would not work here, how does miri generally handle this?

RalfJung: I don't know the type of everything involved here, but there is `cast_signed`/`cast_unsigned` -- does that suffice?

src/shims/x86/mod.rs
```diff
@@ -1200,6 +1200,53 @@ fn pshufb<'tcx>(
...
 }
 
+/// Multiply groups of 4 adjacent pairs of unsigned 8-bit integers in a with corresponding signed
+/// 8-bit integers in b, producing 4 intermediate signed 16-bit results. Sum these 4 results with
+/// the corresponding 32-bit integer in src, and store the packed 32-bit results in dst.
+///
+/// <https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_dpbusd_epi32>
+/// <https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_dpbusd_epi32>
+/// <https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_dpbusd_epi32>
+fn vpdpbusd<'tcx>(
```
RalfJung: If this is only used in `avx512`, please move the function to that file.

src/shims/x86/mod.rs
```diff
@@ -1200,6 +1200,53 @@ fn pshufb<'tcx>(
     interp_ok(())
 }
 
+/// Multiply groups of 4 adjacent pairs of unsigned 8-bit integers in a with corresponding signed
```
RalfJung:
```suggestion
/// Multiply groups of 4 adjacent pairs of unsigned 8-bit integers in `a` with corresponding signed
```
same for all the other references to variable names in the doc comment

src/shims/x86/mod.rs
```diff
@@ -1200,6 +1200,53 @@ fn pshufb<'tcx>(
...
+        let dest = ecx.project_index(&dest, i)?;
+
+        let [a1, a2, a3, a4] = a.to_le_bytes();
+        let [b1, b2, b3, b4] = b.to_le_bytes();
+
+        let intermediate = i32::from(i16::from(a1).wrapping_mul(i16::from(b1.cast_signed())))
+            .wrapping_add(i32::from(i16::from(a2).wrapping_mul(i16::from(b2.cast_signed()))))
+            .wrapping_add(i32::from(i16::from(a3).wrapping_mul(i16::from(b3.cast_signed()))))
+            .wrapping_add(i32::from(i16::from(a4).wrapping_mul(i16::from(b4.cast_signed()))));
+
```
RalfJung: We should find a way to make this more readable...
As a start, why are you mixing i16 and i32? And why `wrapping_mul`? multiplying two `i8` as an `i16` cannot overflow, right? Same for `add`. If things can never overflow, please use the `strict` operations.

Also, I think it would make sense to let-bind the 4 multiplications. Maybe that could even be written in a loop, e.g. via [`from_fn`](https://doc.rust-lang.org/nightly/std/array/fn.from_fn.html)?

tests/pass/shims/x86/intrinsics-x86-avx512.rs
```diff
@@ -411,6 +413,96 @@ unsafe fn test_avx512ternarylogic() {
     test_mm_ternarylogic_epi32();
 }
 
+#[target_feature(enable = "avx512vnni")]
+unsafe fn test_avx512vnni() {
```
RalfJung, line 417: You mentioned that this is aiming to hit a bunch of overflow and truncation cases. Please add comments pointing that out.

src/shims/x86/avx512.rs
```diff
@@ -109,8 +109,64 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let a = ecx.read_scalar(&ecx.project_index(&a, i)?)?.to_u32()?;
+        let b = ecx.read_scalar(&ecx.project_index(&b, i)?)?.to_u32()?;
+        let dest = ecx.project_index(&dest, i)?;
+
+        let a = a.to_le_bytes();
+        let b = b.to_le_bytes();
+
+        let intermediate: [i32; 4] =
+            std::array::from_fn(|i| i32::from(a[i]).strict_mul(i32::from(b[i].cast_signed())));
+        let res = Scalar::from_i32(intermediate.into_iter().fold(src, |x, y| x.wrapping_add(y)));
```
RalfJung: There's still a `wrapping_add` here that needs explanation. Can this actually overflow?

tests/pass/shims/x86/intrinsics-x86-avx512.rs
```diff
@@ -411,6 +413,98 @@ unsafe fn test_avx512ternarylogic() {
...
+    #[target_feature(enable = "avx512vnni")]
+    unsafe fn test_mm512_dpbusd_epi32() {
+        // These inputs aim to hit overflow cases and test that A is interpreted as unsigned and B
```
RalfJung: How do they do that? Which of the constants below conspire how to make overflows happen?

src/shims/x86/avx512.rs
folkertdev: The addition with `src` can overflow because it is just an arbitrary `i32`. The intermediate additions cannot overflow, I've made them strict now.

src/shims/x86/avx512.rs
```diff
@@ -109,8 +109,65 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            }
             _ => return interp_ok(EmulateItemResult::NotSupported),
         }
         interp_ok(EmulateItemResult::NeedsReturn)
     }
 }
+
+/// Multiply groups of 4 adjacent pairs of unsigned 8-bit integers in `a` with corresponding signed
+/// 8-bit integers in `b`, producing 4 intermediate signed 16-bit results. Sum these 4 results with
+/// the corresponding 32-bit integer in `src`, and store the packed 32-bit results in `dst`.
```
RalfJung:
```suggestion
/// the corresponding 32-bit integer in `src` (using wrapping arithmetic), and store the packed 32-bit results in `dst`.
```

tests/pass/shims/x86/intrinsics-x86-avx512.rs
```diff
@@ -417,13 +417,13 @@ unsafe fn test_avx512ternarylogic() {
...
     unsafe fn test_mm512_dpbusd_epi32() {
-        // These inputs aim to hit overflow cases and test that A is interpreted as unsigned and B
-        // as signed.
         const SRC: [i32; 16] = [
             1,
             0,
             0,
             7,
+            // Using values close to the minimum/maximum here makes it very likely that the final
+            // addition with the element from this array overflows. The addition should wrap.
```
RalfJung: What do you mean by "very likely"? You are fixing all the inputs here, there should be no probabilities.^^

This pairs up with elements in A and B, right? So you can have comments there indicating which ones this is paired up with.

RalfJung:

@rustbot author

PR #4779, opened 2025-12-19
Convert Windows to use check_shim_sig instead of check_shim_sig_lenient

CraftSpider:


src/shims/sig.rs
```diff
@@ -53,6 +53,8 @@ macro_rules! shim_sig_arg {
             "*const _" => $this.machine.layouts.const_raw_ptr.ty,
             "*mut _" => $this.machine.layouts.mut_raw_ptr.ty,
             ty if let Some(libc_ty) = ty.strip_prefix("libc::") => $this.libc_ty_layout(libc_ty).ty,
+            ty if let Some(win_ty) = ty.strip_prefix("winapi::") =>
```
CraftSpider, line 62: This is shorthand for a longer path used by the windows_ty_layout function, I thought this made usage a lot nicer and more readable.

tests/pass-dep/getrandom.rs
```diff
@@ -12,6 +12,8 @@ fn main() {
     #[cfg(not(target_os = "solaris"))]
     getrandom_01::getrandom(&mut data).unwrap();
 
+    // TODO: getrandom 0.2 uses the wrong return type for BCrypteGenRandom
+    #[cfg(not(target_os = "windows"))]
```
CraftSpider, line 16: 0.2 uses `u32` for the return type, not `i32`. We now check this and report UB.  Is this worth reporting upstream for a backport?

src/shims/windows/foreign_items.rs
```diff
@@ -141,61 +141,100 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // https://github.com/rust-lang/rust/blob/fb00adbdb69266f10df95a4527b767b0ad35ea48/compiler/rustc_target/src/spec/mod.rs#L2766-L2768,
         // x86-32 Windows uses a different calling convention than other Windows targets
         // for the "system" ABI.
-        let sys_conv = if this.tcx.sess.target.arch == Arch::X86 {
-            CanonAbi::X86(X86Call::Stdcall)
-        } else {
-            CanonAbi::C
-        };
+        // TODO: Use this in shim-sig instead of 'system'? Or is that good enough for ExternAbi?
```
CraftSpider: Drawing attention to this - no tests are failing, but I need to double check whether this works, or if I need to make the `shim_sig` macro support dynamic ABIs.

src/shims/windows/foreign_items.rs
```diff
@@ -141,61 +141,100 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    args,
+                )?;
                 let result = this.SetCurrentDirectoryW(path)?;
                 this.write_scalar(result, dest)?;
             }
             "GetUserProfileDirectoryW" => {
-                let [token, buf, size] =
-                    this.check_shim_sig_lenient(abi, sys_conv, link_name, args)?;
+                let [token, buf, size] = this.check_shim_sig(
+                    shim_sig!(extern "system" fn(winapi::HANDLE, *mut _, *mut _) -> winapi::BOOL),
```
CraftSpider, line 212: General note - I preferred using fundamental types in most places, but I left a couple fundamental windows API types as their `winapi::*` variants, for clarity or because they have somewhat odd Rust representations.
- `BOOL`: Actually an i32, which is unexpected to most not familiar with it.
- `HANDLE`: This has switched between `isize` and `*mut _` in the past, and many Windows shims actually treat it as its own special thing.
- `HLOCAL` and `HMODULE`: Aliases for `HANDLE`, but with the implication they represent only certain types of handle.
- `FARPROC`: Actually `Option<extern "C" fn()>`. Much easier to load the resolved layout of that, instead of trying to add `Option` and `fn` special-cased to `shim_sig_arg`.

I can swap some of these to their literal representations if preferred, it may be a bit more performant, but it also may be worth adding caching to the `*_ty_layout` functions.

src/shims/windows/foreign_items.rs
```diff
@@ -838,64 +1170,99 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                )?;
                 // Return unknown file type.
                 this.write_null(dest)?;
             }
             "AddVectoredExceptionHandler" if this.frame_in_std() => {
-                #[allow(non_snake_case)]
-                let [_First, _Handler] =
-                    this.check_shim_sig_lenient(abi, sys_conv, link_name, args)?;
+                let [first, handler] = this.check_shim_sig(
+                    shim_sig!(extern "system" fn(u32, *mut _) -> *mut _),
```
CraftSpider, line 1239: I standardized all the names. I can undo this if desired, it just seemed like a good drive-by since most stuff uses Rust-style names, not the Windows API names.

src/shims/windows/foreign_items.rs
ChrisDenton: I'd add that a `HLOCAL` is a pointer to memory allocated by `LocalAlloc` and `HMODULE` is a pointer to a module's base address. Neither of these are handles in the same sense as `HANDLE` is despite being defined as a type alias (which is why they can't be passed to normal handle functions).

src/shims/sig.rs
RalfJung: Sounds good! Please document this in the doc comment for `shim_sig!`.

tests/pass-dep/getrandom.rs
RalfJung: There was a getrandom 0.2 release this year... so yeah you might as well report it upstream, but no reason to block this PR on that.

tests/pass-dep/getrandom.rs
```diff
@@ -12,6 +12,8 @@ fn main() {
     #[cfg(not(target_os = "solaris"))]
     getrandom_01::getrandom(&mut data).unwrap();
 
+    // TODO: getrandom 0.2 uses the wrong return type for BCrypteGenRandom
```
RalfJung:
```suggestion
    // On Windows, getrandom 0.2 uses the wrong return type for BCrypteGenRandom.
```

src/shims/windows/foreign_items.rs
RalfJung: "system" was added exactly to support the Windows API, so `extern "system"` really ought to work. Please just remove this comment.

src/shims/windows/foreign_items.rs
RalfJung: Sounds good! I will note though that this cuts both ways -- if someone reads just the Miri code, it's quite confusing to see `winapi::BOOL` and then later `i32` for the same argument. Not sure if there's much we can do about that.

src/shims/windows/foreign_items.rs
```diff
@@ -141,61 +141,100 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        // let sys_conv = if this.tcx.sess.target.arch == Arch::X86 {
+        //     CanonAbi::X86(X86Call::Stdcall)
+        // } else {
+        //     CanonAbi::C
+        // };
 
         // See `fn emulate_foreign_item_inner` in `shims/foreign_items.rs` for the general pattern.
 
         // Windows API stubs.
         // HANDLE = isize
```
RalfJung:
```suggestion
        // HANDLE = *mut c_void (formerly: isize)
```

src/shims/windows/foreign_items.rs
```diff
@@ -303,38 +395,73 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    args,
+                )?;
                 let res = this.DeleteFileW(file_name)?;
                 this.write_scalar(res, dest)?;
             }
             "SetFilePointerEx" => {
-                let [file, distance_to_move, new_file_pointer, move_method] =
-                    this.check_shim_sig_lenient(abi, sys_conv, link_name, args)?;
+                let [file, distance_to_move, new_file_pointer, move_method] = this.check_shim_sig(
+                    // i64 is actually a LARGE_INTEGER union of {u32, i32} and {i64}
```
RalfJung, line 435: And that becomes just `i64` on the ABI level? Those types are not ABI-compatible under Rust rules... but I guess they are compatible on Windows targets so "it's fine".^^

src/shims/windows/foreign_items.rs
RalfJung: Fine for me; so far we just didn't have anyone care enough to try and standardize this code.

FWIW, we also have a mix of styles for the functions that implement these shims, if they are implemented in separate functions.

RalfJung (changes requested):
Thanks! Nice to see the macro work at scale. :)
I will just trust that those signatures are correct based on the test passing; I have no intuition for Windows APIs so I'd have to check every one of them manually...

The only actually requested change here is a comment nit.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4779/117d9133a41b6875eac5868023f7154e87e4358b..25047ae2c68f93f638902ff9cb9985f0f6070cfe)*

src/shims/windows/foreign_items.rs
CraftSpider: All Rust code just defines this as an `i64`, including the official windows-sys bindings. So I think Windows considers it 'the same' for the system ABI.

tests/pass-dep/getrandom.rs
CraftSpider: Opened https://github.com/rust-random/getrandom/issues/775

src/shims/windows/foreign_items.rs
CraftSpider: I personally have preferred the external function method,, and would be willing to make a PR to standardize on that as a follow-up at some point.

josephlr:
Coming here from https://github.com/rust-random/getrandom/issues/775 it suprises me that using a `u32` instead of `i32` when binding to a Windows API triggers this check in Miri (given that `i32` and `u32` have identical ABIs on Windows).

Should this check be relaxed to not trigger when an ABI-identical type is used?

src/shims/windows/foreign_items.rs
RalfJung: It would still be UB to do this if caller and callee are both defined in Rust, but for functions only invoked via FFI the rules are a bit less clear.

RalfJung:
I am not aware of any sense in which i32 and u32 are ABI-identical at the language level. On some targets under some configurations (in particular, without CFI) they may be passed identically on the assembly level, but to my knowledge neither LLVM nor C nor Rust allow the programmer to make use of that.

src/shims/sig.rs
```diff
@@ -53,7 +56,9 @@ macro_rules! shim_sig_arg {
             "*const _" => $this.machine.layouts.const_raw_ptr.ty,
             "*mut _" => $this.machine.layouts.mut_raw_ptr.ty,
             ty if let Some(libc_ty) = ty.strip_prefix("libc::") => $this.libc_ty_layout(libc_ty).ty,
-            ty => panic!("unsupported signature type {ty:?}"),
+            ty if let Some(win_ty) = ty.strip_prefix("winapi::") =>
+                $this.windows_ty_layout(win_ty).ty,
+            ty => helpers::path_ty_layout($this, &ty.split("::").collect::<Vec<_>>()).ty,
```
RalfJung: Oh I only just realized you added a general fallback... not sure if we really want that. Is it needed for the PR?

RalfJung: Ah, it's needed for _Unwind_Reason_Code... hm

Issue #4636, opened 2025-10-15
Support for `io_uring`


sugarraysam:
Hello, as titled I am running into unsupported syscall error in my project, where I do use `io_uring`.

```bash
// from kernel - 425 maps to `io_uring_setup`
uapi/asm-generic/unistd.h
759:#define __NR_io_uring_setup 425
760:__SYSCALL(__NR_io_uring_setup, sys_io_uring_setup)

// Error log from miri
$ cargo +nightly test miri` test runtime::cancel::tests::test_cancel_timeout_multishot ... error: unsupported operation: syscall: unsupported syscall number 425
   --> /home/doodly/opt/io-uring/src/sys/mod.rs:115:15
    |
115 |     to_result(syscall(SYSCALL_SETUP, entries as c_long, p as c_long) as _)
    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ unsupported operation occurred here
```

I tried digging into the codebase, and am finding the `syscall()` function in `src/shims/unix/linux_like/syscall.rs`, but could not decipher exactly the path ahead to adding support for this syscall.

Thank you for the help.

RalfJung:
Yeah, we have no support for io_uring at the moment... and I assume that's pretty far out. We currently don't even support regular sockets.^^

sugarraysam:
Is there another path forward where we first "ignore" io_uring related syscalls so the rest of miri can still work on the project?

I don't care for full `io_uring` support, I just wanted to use regular miri in the codebase.

bjorn3:
You can use `#[cfg(miri)]` to run different code in miri than on regular builds. Miri can't just ignore unknown syscalls however. Syscalls can make arbitrary modifications to the program state, ignoring which can easily cause the program to have UB. ANd besides miri wouldn't know what return value to return for the syscall.

PR #2494, opened 2022-08-18
Add `gnu_get_libc_[release|version]` shims.

LegNeato:

PR #4784, opened 2025-12-22
Add a shim for uname systemcall

Thomasdezeeuw:
This system call is used to get information about the OS that is run. A10 for example uses this to determine what Linux version is being run and uses it to skip tests that are not supported by older Linux versions, see https://github.com/Thomasdezeeuw/a10/blob/87d04713a3ff5f9d60e1473a72939a1e1bcdaf13/tests/util/mod.rs#L44-L62.

When Miri runs in isolation mode this returns some dummy data based on a recent Linux version. Since the data should only be informative I think it makes sense to return something and allow the program to continue rather than stopping it.

### DIFF
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -172,7 +172,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let result = this.getpid()?;
                 this.write_scalar(result, dest)?;
             }
-            "uname" => {
+            "uname" if !matches!(&this.tcx.sess.target.os, Os::FreeBsd) => {
                 let [uname] = this.check_shim_sig(
                     shim_sig!(extern "C" fn(*mut _) -> i32),
                     link_name,


RalfJung:
Thanks for the PR!

Please extend the PR description, explaining the motivation for why adding this syscall is helpful and why it makes sense to return some hard-coded "dummy" data.

Thomasdezeeuw:
This is my first contribution to Miri, so I probably did something not quite right. If so, please let me know.

I tried to follow the conventions I found in the code as the contribution guide didn't provide a lot of guidance around adding support for new system calls.

Thomasdezeeuw:
Updated the pr description.

I think I also solved most issues found by the CI, but the Windows one I'm not sure of. In `src/shims/unix/env.rs` I would expect the code to be unix only, but it seems to be included on Windows as well, which then complains about missing the `libc` crate. @RalfJung any recommendations on what the best approach would be?

saethlin:
In general, it is wrong to implement shims in Miri by just calling the shimmed function in the shim, which is what you are trying to do. Miri is a cross-interpreter, so it needs to be able to run Linux code on Windows. `src/shims/unix` is shims used when the `--target` is unix-like, not when the OS running the interpreter is. I'm using a lot of words because this took me time to get used to and what you tried to do here is a very common mistake when first implementing a shim.

Ideally, you implement a shim by calling a portable function in the standard library. If that is not available, you might be able to use `cfg` to pick a reasonable implementation on a non-unix system.

src/shims/unix/env.rs
```diff
@@ -272,6 +272,61 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    fn uname(&mut self, uname: &OpTy<'tcx>) -> InterpResult<'tcx, Scalar> {
+        let this = self.eval_context_mut();
+        this.assert_target_os_is_unix("uname");
+
+        let uname_ptr = this.read_pointer(uname)?;
+        if this.ptr_is_null(uname_ptr)? {
+            return this.set_last_error_and_return_i32(LibcError("EFAULT"));
+        }
+
+        let mut uname_buf = unsafe { std::mem::zeroed() };
```
saethlin: FYI, we try to avoid unsafe in the interpreter except where it is absolutely necessary.

RalfJung:
> If that is not available, you might be able to use cfg to pick a reasonable implementation on a non-unix system.

I think we should avoid this -- I don't want Miri to become a platform abstraction layer full of `cfg` for all sorts of host systems. We have enough on our hands dealing with all sorts of *target* systems.

For uname, if we do have a shim at all, it should probably just return something plausible that is hard-coded. But I can easily see that causing more trouble than it solves.

What kernel features does a10 enable based on the kernel version? Chances are, Miri doesn't support them anyway. Everything we support is quite ancient, AFAIK. So it's probably best to just hard-code some old kernel version under `cfg(miri)` in your crate. (AFAIK, kernel version checking is an anti-pattern anyway. Isn't the usual approach to issue a syscall and check the error code?)

RalfJung:
Looking at the code, this seems to be mostly about io_uring... and sure enough, a10 is an io_uring library. Miri is quite far from having any sort of io_uring support (see https://github.com/rust-lang/miri/issues/4636). So even if uname works, I'm afraid Miri is not going to get very far in your test suite.

I appreciate the attempt to help, but I'm afraid Miri is not very suited for such super low-level platform-specific code.

Thomasdezeeuw:
> What kernel features does a10 enable based on the kernel version? Chances are, Miri doesn't support them anyway.

It doesn't change the code based on the kernel version, it only changes the tests to skip them if a certain feature is not supported and the OS would return some kind of not supported error.

> Looking at the code, this seems to be mostly about io_uring... and sure enough, a10 is an io_uring library. Miri is quite far from having any sort of io_uring support (see #4636). So even if uname works, I'm afraid Miri is not going to get very far in your test suite.

To clear up any potential confusion, `uname(2)` is part of POSIX 2001 and is unrelated to io_uring. 

This pr is an experiment for myself to get more familiar with Miri's codebase, I thought I do so while getting slightly closer to getting A10's test suite to run in Miri. The end goal for me would be to contribute support for io_uring (but I'm well aware that's a lot more work than this pr :) ).

Re: making the system call vs. hardcoded values. Let me know what is the best route here, I don't know enough about Miri's internals & goals to argue one way or the other.

RalfJung:
Related prior work: https://github.com/rust-lang/miri/pull/2494. We ended up not implementing that since it wouldn't really have helped for any usecase.

RalfJung:
Thinking a bit more specifically about this case -- IMO it makes little sense to query the kernel version of the host and use it in the interpreted program. I can't imagine any way of using `uname` where that would give the desired behavior.


> To clear up any potential confusion, uname(2) is part of POSIX 2001 and is unrelated to io_uring.

I understand it has nothing to do with io_uring. :)  But the usecase you presented is "I have code that checks the kernel version to see which io_uring flags are supported" (if I understood it correctly). I am not entirely convinced that's a sufficient motivation, for the reasons stated above. However, "it is part of POSIX" is a motivation I could be convinced of.

However, if we give any kind of version number, we have to be a bit careful -- your current implementation uses something involving "Linux" and "6.18" on *all* Unix targets, which makes little sense for e.g. FreeBSD or Illumos. You're also setting `machine` to "x86_64" even if the current target is something else. So that definitely needs to be fixed.

Wouldn't it be better to not even try to make this look like a real Linux, and instead just say "this is Miri 0.1" (based on `CARGO_PKG_VERSION`)? Any code that parses version numbers has to already be know this is fragile and have appropriate fallbacks in place.

@rustbot author

src/shims/unix/env.rs
Thomasdezeeuw: Pushed a commit to initialise all fields to zero. Do note that will make porting to new OS, which might have a different set of fields in `utsname`, a bit more annoying.

Thomasdezeeuw:
I've implemented your suggestion @RalfJung and now it return values that are unlike Like (mostly `Miri` and Miri's version). I've kept the `version` field empty.

Does it still make sense to have the actual call when not isolated? If not I can remove that code as well.

Otherwise @rustbot ready

src/shims/unix/env.rs
RalfJung: You shouldn't use any `libc` types or functions here. Those correspond to the current host, which is entirely unrelated to the current target.

RalfJung:
> Does it still make sense to have the actual call when not isolated? If not I can remove that code as well.

No, as I said above I think that does not make sense. That should then also avoid the CI issues.

@rustbot author

Thomasdezeeuw:
This required a bit more work:
* the returned architecture (machine field) is done retrieved from the virtual/Miri machine instead of the compiled (binary) constant
* We ignore fields that are not present in the type, mainly `domainname` which is Linux/Android only.
* For FreeBSD I added `__xuname` which is called by `uname` to do the actual implementation.

@rustbot ready

src/shims/unix/foreign_items.rs
```diff
@@ -172,6 +172,16 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let result = this.getpid()?;
                 this.write_scalar(result, dest)?;
             }
+            "uname" => {
```
RalfJung: Apparently that's not universally the symbol name then, given what you had to do for freebsd? Please add a check similar to what we do for `flock` and other shims to ensure this arm is only taken on OSes where it makes sense.

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -164,6 +164,23 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let [size, uname] = this.check_shim_sig(
+                    shim_sig!(extern "C" fn(i32, *mut _) -> i32),
+                    link_name,
+                    abi,
+                    args,
+                )?;
+                // The size determines the length of each field in the utsname
+                // structure. We don't really care about this, all we do check
+                // that the argument is an i32.
+                let _ = this.read_scalar(size)?.to_i32()?;
```
RalfJung:
```suggestion
                let _size = this.read_scalar(size)?.to_i32()?;
```

tests/pass-dep/libc/libc-uname.rs
```diff
@@ -0,0 +1,39 @@
...
+
+fn test_ok() {
+    // SAFETY: all zeros for `utsname` is valid.
+    let mut uname: libc::utsname = unsafe { std::mem::zeroed() };
+    let result = unsafe { libc::uname(&mut uname) };
+    if result != 0 {
+        panic!("failed to call uname");
+    }
+
+    // These values are only correct when running isolated.
```
RalfJung: That comment is no longer accurate, right?

src/shims/unix/env.rs
```diff
@@ -272,6 +272,37 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            return this.set_last_error_and_return_i32(LibcError("EFAULT"));
+        }
+
+        let uname = this.deref_pointer_as(uname, this.libc_ty_layout("utsname"))?;
+        let arch = this.machine.tcx.sess.target.arch.desc_symbol();
+        let values = [
+            ("sysname", "Miri"),
+            ("nodename", "Miri"),
+            ("release", env!("CARGO_PKG_VERSION")),
+            ("version", ""),
```
RalfJung: What does "version" usually contain and why is it empty here?

src/shims/unix/env.rs
```diff
@@ -272,6 +272,37 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            ("machine", arch.as_str()),
+            ("domainname", "(none)"),
+        ];
+        for (name, value) in values {
+            // Since not all OS have all fields (e.g. domainname), we ignore
+            // fields that are not present.
+            if let Some(field) = this.try_project_field_named(&uname, name)? {
```
RalfJung: Is domainname the only exception here?

The usual approach we take is to assert that the fields we expect to exist do exist. See `fn write_stat_buf`. Seems best to remain consistent and follow the same approach here.

RalfJung (changes requested):
Thanks! We are entering the fine-tuning phase of this PR. :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4784/4a8db0f1f0eed8c576ad5d7c4b5ff537d14d1451..30260e7daac0bab728e1eb88c070b051953b5e3a)*

src/shims/unix/env.rs
Thomasdezeeuw: I'm not 100% sure, but I think it's very much OS dependant. Posix defines it as (https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/sys_utsname.h.html):

> Current version level of this release.

Some examples:
* Arch Linux: `#1 SMP PREEMPT_DYNAMIC Thu, 18 Dec 2025 18:00:18 +0000`
* macOS: `Darwin Kernel Version 24.6.0: Wed Oct 15 21:12:15 PDT 2025; root:xnu-11417.140.69.703.14~1/RELEASE_ARM64_T6041`

I'm open to suggestion on what to put here for Miri, but realistically I don't think this info is useful in general (outside from maybe logging it).

src/shims/unix/env.rs
Thomasdezeeuw: > Is domainname the only exception here?

Yes. It's not the Posix standard, it seems like a glibc/GNU addition.

> The usual approach we take is to assert that the fields we expect to exist do exist. See `fn write_stat_buf`. Seems best to remain consistent and follow the same approach here.

I'm not sure if the approach in `write_stat_buf` is a good idea to be honest. It would mean a higher cost when porting to now OS as you need to check if the OS has the field or not. Furthermore if we don't fill the field for an OS that does define it will cause UB if the user didn't initialise the buffer before the call.

I've split the values in required fields (using `project_field_named`) and optional fields (using the `try_` variant, still ignoring if it's missing). Let me know what you think.

src/shims/unix/foreign_items.rs
Thomasdezeeuw: I think is the correct name for the function, but the implementation for FreeBSD within libc then calls `__xuname`. I'm not exactly sure what would happen if we remove it from the list here, but I think it would cause errors to be honest.

Thomasdezeeuw: I tried running it with `uname` here and with `--target x86_64-unknown-freebsd` and it does work, then I'll have to figure what the correct names are

RalfJung: This match arm here is just dead code for freebsd target. So just add a check to only allow this arm on the other Unixes.

src/shims/unix/env.rs
RalfJung: I don't really understand the distinction between "release" and "version"... but it seems like typically, both contain a version number. So maybe "version" should be "Miri 0.1" or so?

src/shims/unix/env.rs
RalfJung: I guess it's fine... I'm not a huge fan of `try_project_field_named` even existing but OTOH we already gave it and use it elsewhere (though not very often). The better API here would somehow check that we did not miss a field but until we have that, we can go with the "try" variant.

RalfJung:
@Thomasdezeeuw what is the status of this? Currently it is marked as waiting-for-author, which means we are waiting for you to indicate that you took care of all the feedback we gave you. Please request a review or write `@rustbot ready` when this is ready for review again.

src/shims/unix/env.rs
Thomasdezeeuw: I've changed this to be "Miri 0.1"

src/shims/unix/env.rs
Thomasdezeeuw: I've remove the `unsafe` block and the usage of libc types, so resolving this.

src/shims/unix/foreign_items.rs
Thomasdezeeuw: Done.

Thomasdezeeuw (commented):
I forgot about this pr, sorry about that. I've addressed the last point of feedback.

@rustbot ready

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4784/4a8db0f1f0eed8c576ad5d7c4b5ff537d14d1451..a592da111074ada170ab7e15cd1bd8efbc85dd0f)*

src/shims/unix/freebsd/foreign_items.rs
```diff
@@ -164,6 +164,23 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let [size, uname] = this.check_shim_sig(
+                    shim_sig!(extern "C" fn(i32, *mut _) -> i32),
+                    link_name,
+                    abi,
+                    args,
+                )?;
+                // The size determines the length of each field in the utsname
+                // structure. We don't really care about this, all we do check
+                // that the argument is an i32.
+                let _size = this.read_scalar(size)?.to_i32()?;
```
RalfJung, line 180: How can we not care about it? Won't we then just return nonsense when the user uses a size that's different from what we expect (which is apparently 256)?

src/shims/unix/foreign_items.rs
```diff
@@ -172,6 +172,16 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let result = this.getpid()?;
                 this.write_scalar(result, dest)?;
             }
+            "uname" if !matches!(&this.tcx.sess.target.os, Os::FreeBsd) => {
+                let [uname] = this.check_shim_sig(
+                    shim_sig!(extern "C" fn(*mut _) -> i32),
```
RalfJung, line 177: I had suggested to follow the "flock" pattern. This is not what flock does.

This is what flock does:
```suggestion
            "uname" => {
                // Not all Unixes have the `uname` symbol, e.g. FreeBSD does not.
                this.check_target_os(&[Os::Linux, Os::Android, Os::MacOs, Os::Illumos], link_name)?;
                let [uname] = this.check_shim_sig(
```
You'll probably have to add more OSes.

tests/pass-dep/libc/libc-uname.rs
```diff
@@ -0,0 +1,38 @@
...
+        std::env::consts::ARCH
+    );
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    assert_eq!(unsafe { CStr::from_ptr(&uname.domainname as *const _) }, c"(none)");
+}
+
+fn test_null_ptr() {
+    let result = unsafe { libc::uname(ptr::null_mut()) };
+    assert_eq!(result, -1);
+    assert_eq!(io::Error::last_os_error().raw_os_error(), Some(libc::EFAULT));
```
RalfJung, line 37: Please use `errno_result`/`errno_check` from the libc utils. (See other tests for how to import and use those functions.) Use it for all calls to `uname`.

Thomasdezeeuw:
It seems this wasn't as closed to be merged as I hoped. I don't want to spend significantly more time on this, so I'm going to close this. Sorry to waste everybody's time.

RalfJung:
Sorry for that. I'll probably pick up your commits to finish this. Thanks for the work you put into this!

PR #4788, opened 2025-12-25
Add supprot for FLS functions on Windows

ohadravid:
This PR adds support for the `Fls{Alloc,GetValue,SetValue,Free}` functions that are available on Windows. They enable native thread-local storage **with** destructor callbacks ([MSDN](https://learn.microsoft.com/en-us/windows/win32/api/fibersapi/nf-fibersapi-flsalloc)).

I also have a [PR to std](https://github.com/rust-lang/rust/pull/148799) that switches the default impl to use these functions, but I think supporting them in Miri can be useful anyway 🪟


RalfJung:
Thanks for the PR!
I would prefer to do this only if actually needed by std. So let's see how that PR goes.

RalfJung:
Let's just do the review in https://github.com/rust-lang/rust/pull/148799, that should be simpler.

PR #4790, opened 2025-12-27
show a warning when combing native-lib mode and many-seeds

RalfJung:


CatsAreFluffy:
Typo: "intersting"

RalfJung:
Ah, fair... just in a comment though, so not really worth a separate PR.

PR #4792, opened 2025-12-27
native-lib/trace: Intercept allocating calls, but do nothing

nia-e:
This is part 1 of rust-lang/miri#4791; it adds the bits needed to intercept allocating/deallocating calls to libc, but silently ignores them for now. In a follow-up PR this will forward them to shims we control.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ serde = { version = "1.0.219", features = ["derive"], optional = true }
 nix = { version = "0.30.1", features = ["mman", "ptrace", "signal"], optional = true }
 ipc-channel = { version = "0.20.0", optional = true }
 capstone = { version = "0.13", optional = true }
+proc-maps = { version = "0.4.0", optional = true }
 
 [target.'cfg(all(target_os = "linux", target_pointer_width = "64", target_endian = "little"))'.dependencies]
 genmc-sys = { path = "./genmc-sys/", version = "0.1.0", optional = true }
@@ -66,8 +67,8 @@ genmc = ["dep:genmc-sys"]
 stack-cache = []
 expensive-consistency-checks = ["stack-cache"]
 tracing = ["serde_json"]
-native-lib = ["dep:libffi", "dep:libloading", "dep:capstone", "dep:ipc-channel", "dep:nix", "dep:serde"]
 jemalloc = []
+native-lib = ["dep:libffi", "dep:libloading", "dep:capstone", "dep:ipc-channel", "dep:nix", "dep:serde", "dep:proc-maps"]
 
 [lints.rust.unexpected_cfgs]
 level = "warn"
--- a/src/shims/native_lib/mod.rs
+++ b/src/shims/native_lib/mod.rs
@@ -27,6 +27,7 @@ mod ffi;
 pub mod trace;
 
 use self::ffi::OwnedArg;
+use self::trace::EvalContextExt as _;
 use crate::*;
 
 /// The final results of an FFI trace, containing every relevant event detected
@@ -85,13 +86,9 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
         libffi_args: &mut [OwnedArg],
     ) -> InterpResult<'tcx, (crate::ImmTy<'tcx>, Option<MemEvents>)> {
         let this = self.eval_context_mut();
-        #[cfg(target_os = "linux")]
-        let alloc = this.machine.allocator.as_ref().unwrap();
-        #[cfg(not(target_os = "linux"))]
-        // Placeholder value.
-        let alloc = ();
 
-        trace::Supervisor::do_ffi(alloc, || {
+        let ty_is_sized = dest.layout.ty.is_sized(*this.tcx, this.typing_env());
+        this.do_ffi(|| {
             // Call the function (`ptr`) with arguments `libffi_args`, and obtain the return value
             // as the specified primitive integer type
             let scalar = match dest.layout.ty.kind() {
@@ -117,7 +114,12 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 }
                 ty::Int(IntTy::Isize) => {
                     let x = unsafe { ffi::call::<isize>(fun, libffi_args) };
-                    Scalar::from_target_isize(x.try_into().unwrap(), this)
+                    // We already know native-lib mode means target == host, so
+                    // this is ok.
+                    Scalar::from_int(
+                        i128::try_from(x).unwrap(),
+                        Size::from_bytes(size_of::<isize>()),
+                    )
                 }
                 // uints
                 ty::Uint(UintTy::U8) => {
@@ -138,7 +140,10 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 }
                 ty::Uint(UintTy::Usize) => {
                     let x = unsafe { ffi::call::<usize>(fun, libffi_args) };
-                    Scalar::from_target_usize(x.try_into().unwrap(), this)
+                    Scalar::from_uint(
+                        u128::try_from(x).unwrap(),
+                        Size::from_bytes(size_of::<usize>()),
+                    )
                 }
                 ty::Float(FloatTy::F32) => {
                     let x = unsafe { ffi::call::<f32>(fun, libffi_args) };
@@ -154,10 +159,10 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     unsafe { ffi::call::<()>(fun, libffi_args) };
                     return interp_ok(ImmTy::uninit(dest.layout));
                 }
-                ty::RawPtr(ty, ..) if ty.is_sized(*this.tcx, this.typing_env()) => {
+                ty::RawPtr(ty, ..) if ty_is_sized => {
                     let x = unsafe { ffi::call::<*const ()>(fun, libffi_args) };
                     let ptr = StrictPointer::new(Provenance::Wildcard, Size::from_bytes(x.addr()));
-                    Scalar::from_pointer(ptr, this)
+                    Scalar::Ptr(ptr, u8::try_from(size_of::<*const ()>()).unwrap())
                 }
                 _ =>
                     return Err(err_unsup_format!(
@@ -225,7 +230,6 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
     /// assumed to be exact.
     fn tracing_apply_accesses(&mut self, events: MemEvents) -> InterpResult<'tcx> {
         let this = self.eval_context_mut();
-
         for evt in events.acc_events {
             let evt_rg = evt.get_range();
             // LLVM at least permits vectorising accesses to adjacent allocations,
@@ -235,7 +239,11 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let Some(alloc_id) =
                     this.alloc_id_from_addr(curr.to_u64(), rg.len().try_into().unwrap())
                 else {
-                    throw_ub_format!("Foreign code did an out-of-bounds access!")
+                    throw_ub_format!(
+                        "Foreign code did an out-of-bounds access at {:#0x} for {:#0x} bytes!",
+                        curr,
+                        rg.len(),
+                    );
                 };
                 let alloc = this.get_alloc_raw(alloc_id)?;
                 // The logical and physical address of the allocation coincide, so we can use
@@ -526,7 +534,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             this.call_native_with_args(link_name, dest, code_ptr, &mut libffi_args)?;
 
         if tracing {
-            this.tracing_apply_accesses(maybe_memevents.unwrap())?;
+            let mm = maybe_memevents.unwrap();
+            this.tracing_apply_accesses(mm)?;
         }
 
         this.write_immediate(*ret, dest)?;
--- a/src/shims/native_lib/trace/child.rs
+++ b/src/shims/native_lib/trace/child.rs
@@ -1,6 +1,5 @@
-use std::cell::RefCell;
 use std::ptr::NonNull;
-use std::rc::Rc;
+use std::sync::atomic::Ordering;
 
 use ipc_channel::ipc;
 use nix::sys::{mman, ptrace, signal};
@@ -10,8 +9,9 @@ use rustc_const_eval::interpret::InterpResult;
 use super::CALLBACK_STACK_SIZE;
 use super::messages::{Confirmation, StartFfiInfo, TraceRequest};
 use super::parent::{ChildListener, sv_loop};
-use crate::alloc::isolated_alloc::IsolatedAlloc;
 use crate::shims::native_lib::MemEvents;
+use crate::shims::native_lib::trace::parent::{PAGE_ADDR, PAGE_COUNT, PAGE_SIZE};
+use crate::*;
 
 /// A handle to the single, shared supervisor process across all `MiriMachine`s.
 /// Since it would be very difficult to trace multiple FFI calls in parallel, we
@@ -32,7 +32,7 @@ pub struct Supervisor {
     /// parent process has handled the request from `message_tx`.
     confirm_rx: ipc::IpcReceiver<Confirmation>,
     /// Receiver for memory acceses that ocurred during the FFI call.
-    event_rx: ipc::IpcReceiver<MemEvents>,
+    event_rx: Option<ipc::IpcReceiver<MemEvents>>,
 }
 
 /// Marker representing that an error occurred during creation of the supervisor.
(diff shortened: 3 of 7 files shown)


Cargo.toml
```diff
@@ -40,6 +40,7 @@ serde = { version = "1.0.219", features = ["derive"], optional = true }
 nix = { version = "0.30.1", features = ["mman", "ptrace", "signal"], optional = true }
 ipc-channel = { version = "0.20.0", optional = true }
 capstone = { version = "0.13", optional = true }
+proc-maps = { version = "0.4.0", optional = true }
```
RalfJung, line 43: That's a huge pile of new dependencies. :/ and some duplicated ones, like `libloading`.

How long does this build script take? We *just* managed to finally get rid of the capstone overhead for check builds, I don't want to add back any new slow build scripts.

src/shims/native_lib/mod.rs
```diff
@@ -85,13 +86,9 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
     ) -> InterpResult<'tcx, (crate::ImmTy<'tcx>, Option<MemEvents>)> {
         let this = self.eval_context_mut();
-        #[cfg(target_os = "linux")]
-        let alloc = this.machine.allocator.as_ref().unwrap();
-        #[cfg(not(target_os = "linux"))]
-        // Placeholder value.
-        let alloc = ();
 
-        trace::Supervisor::do_ffi(alloc, || {
+        let ty_is_sized = dest.layout.ty.is_sized(*this.tcx, this.typing_env());
```
RalfJung, line 90: We do not support unsized return types for FFI calls.^^ So what is this about?

EDIT: Ah, it is factoring out the RawPtr check below -- but this is incorrect, it checks sizedness of the wrong thing. Anyway this logic looks quite different on master so hopefully a rebase will make this all go away.

src/shims/native_lib/mod.rs
```diff
@@ -154,10 +159,10 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     unsafe { ffi::call::<()>(fun, libffi_args) };
                     return interp_ok(ImmTy::uninit(dest.layout));
                 }
-                ty::RawPtr(ty, ..) if ty.is_sized(*this.tcx, this.typing_env()) => {
+                ty::RawPtr(ty, ..) if ty_is_sized => {
```
RalfJung, line 162: This changes what is being checked. Previously it checked whether the pointee is sized, now you are checking of the pointer is sized (which obviously it always is).

src/shims/native_lib/mod.rs
```diff
@@ -526,7 +534,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             this.call_native_with_args(link_name, dest, code_ptr, &mut libffi_args)?;
 
         if tracing {
-            this.tracing_apply_accesses(maybe_memevents.unwrap())?;
+            let mm = maybe_memevents.unwrap();
```
RalfJung, line 537: What's `mm` for? After unwrapping they are not "maybe" any more.

src/shims/native_lib/trace/child.rs
```diff
@@ -54,24 +54,37 @@ impl Supervisor {
...
         let mut sv_guard = SUPERVISOR.lock().unwrap();
         // If the supervisor is not initialised for whatever reason, fast-return.
         // As a side-effect, even on platforms where ptracing
         // is not implemented, we enforce that only one FFI call
         // happens at a time.
         let Some(sv) = sv_guard.as_mut() else { return f().map(|v| (v, None)) };
 
+        // Save the machine pointer to a location where the libc interceptors can use it,
+        // since we can't pass in arguments.
+        super::parent::MACHINE_PTR.store(machine_ptr.cast(), std::sync::atomic::Ordering::Relaxed);
```
RalfJung, line 79: This seems very similar to `native_lib_ecx_interchange` which has been added by a concurrent PR -- we don't need both of those, do we?

src/shims/native_lib/trace/parent.rs
```diff
@@ -430,16 +532,19 @@ fn capstone_disassemble(
     Ok(())
 }
 
+// THIS NEEDS TO SOMEHOW CATCH SEGFAULTS INSIDE IT!!!! AND ALSO IN THE FULL ONE IT
+// NEEDS TO GET THEM TO LOG THOSE ACCESSES AAAAAAAAAAAAAAAAA
```
RalfJung, line 536: Eh...?^^ This doesn't quite look ready for review. ;)

src/shims/native_lib/trace/child.rs
```diff
@@ -261,3 +290,52 @@ pub fn register_retcode_sv(code: i32) {
...
+// by offseting the instruction pointer into them. However, they are here because
+// they execute in the child process.
+
+/// Disables protections on the page whose address is currently in `PAGE_ADDR`.
+///
+/// SAFETY: `PAGE_ADDR` should be set to a page-aligned pointer to an owned page,
+/// `PAGE_SIZE` should be the host pagesize, and the range from `PAGE_ADDR` to
+/// `PAGE_SIZE` * `PAGE_COUNT` must be owned and allocated memory. No other threads
+/// should be running.
+pub unsafe extern "C" fn mempr_off() {
```
RalfJung, line 304: This and the next one only need to be `pub(super)`, right?

src/shims/native_lib/trace/child.rs
```diff
@@ -45,7 +45,7 @@ impl Supervisor {
         SUPERVISOR.lock().unwrap().is_some()
     }
 
-    unsafe fn protect_pages(
+    pub unsafe fn protect_pages(
```
RalfJung, line 48: Why is this `pub` now?

src/shims/native_lib/trace/child.rs
```diff
@@ -261,3 +290,52 @@ pub fn register_retcode_sv(code: i32) {
...
+        // It's up to the caller to make sure this doesn't actually overflow, but
+        // we mustn't unwind from here, so...
+        if libc::mprotect(
+            PAGE_ADDR.load(Ordering::SeqCst).cast(),
+            len,
+            libc::PROT_READ | libc::PROT_WRITE,
+        ) != 0
+        {
+            // Can't return or unwind, but we can do this.
+            std::process::exit(-1);
```
RalfJung, line 318: Looking at this... `exit` calls atexit handlers, who knows what those do. It may be safer to call `abort`?

src/shims/native_lib/trace/parent.rs
```diff
@@ -23,14 +24,42 @@ const ARCH_WORD_SIZE: usize = 8;
...
+/// A pointer to the `MiriInterpCx` for use within the libc shims.
+pub(super) static MACHINE_PTR: AtomicPtr<()> = AtomicPtr::new(std::ptr::null_mut());
+/// Is the return address within the libc-mapped area(s)?
+pub(super) static RET_IS_LIBC: AtomicBool = AtomicBool::new(false);
+
+/// Information about which pages were allocated/deallocated after a single
+/// libc intercepted event. After use, these are reset to 0.
+///
+/// INVARIANT: A single libc event can only allocate/deallocate one contiguous
+/// block of pages (as would be the case in a large `realloc`).
```
RalfJung, line 52: Would it make sense to bundle these in a struct, so they are more clearly grouped?

src/shims/native_lib/trace/parent.rs
```diff
@@ -310,6 +399,13 @@ fn get_disasm() -> capstone::Capstone {
     .unwrap()
 }
 
+struct SegfaultCatchingStuff<'a, 'b, 'c> {
```
RalfJung, line 402: Please add a doc comment. Also, maybe call it `SegfaultCatchingParams` or so?

src/shims/native_lib/trace/parent.rs
```diff
@@ -522,7 +627,7 @@ fn handle_segfault(
     ptrace::setregs(pid, new_regs).unwrap();
 
     // Our mempr_* functions end with a raise(SIGSTOP).
-    wait_for_signal(Some(pid), signal::SIGSTOP, InitialCont::Yes)?;
+    wait_for_signal(Some(pid), signal::SIGSTOP, InitialCont::Yes, None)?;
```
RalfJung, line 630:
```suggestion
    wait_for_signal(Some(pid), signal::SIGSTOP, InitialCont::Yes, /* catch_segfaults */ None)?;
```
Applies to all callsites.

Alternatively, make this an enum where one variant has all the params instead of reusing `Option`.

src/shims/native_lib/trace/parent.rs
```diff
@@ -553,64 +658,163 @@ fn handle_segfault(
...
-        ) != 0
-        {
-            // Can't return or unwind, but we can do this.
-            std::process::exit(-1);
+    /// Gets the libc function that a given instruction pointer corresponds to.
+    fn get_libc_fn(addr: usize) -> Option<LibcFn> {
+        // We'll be one instruction past the start
+        #[expect(clippy::as_conversions)]
+        match addr.strict_sub(BREAKPT_INSTR_SIZE) {
+            a if a == (libc::malloc as *const () as usize) => Some(LibcFn::Malloc),
```
RalfJung, line 696: So... we just compare this with a function pointer...?!? And we do this in the parent, even though this is a child address? And I guess this works because we forked and so addresses will be randomized the same way and... 
:exploding_head:  :dizzy_face:  This code is every abstract machine's worst nightmare.^^

src/shims/native_lib/trace/parent.rs
```diff
@@ -553,64 +658,163 @@ fn handle_segfault(
...
-pub unsafe extern "C" fn mempr_on() {
-    use std::sync::atomic::Ordering;
+    let page_size = PAGE_SIZE.load(Ordering::Relaxed);
+    let regs = ptrace::getregs(pid).unwrap();
+    match get_libc_fn(regs.ip()) {
+        Some(_) => {
+            // We'll possibly want to call libc functions in the interceptor shims,
+            // so make sure they're working.
+            fixup_libc(pid);
+            // On x86, the return address will be the last item on the stack.
```
RalfJung, line 713: How do we know that we are on x86 here?

src/shims/native_lib/trace/parent.rs
```diff
@@ -553,64 +658,163 @@ fn handle_segfault(
...
+                }
+            });
+            ptrace::write(pid, RET_IS_LIBC.as_ptr().cast(), ret_is_libc.into()).unwrap();
+
+            // Override the return address to give us another sigtrap
+            // (but save the original bytes).
+            let ret_addr_bytes =
+                ptrace::read(pid, std::ptr::without_provenance_mut(ret_addr)).unwrap();
+            ptrace::write(pid, std::ptr::without_provenance_mut(ret_addr), BREAKPT_INSTR.into())
+                .unwrap();
```
RalfJung, line 754: This does not overwrite the return address, it overwrites the memory the return address points to -- right?

src/shims/native_lib/trace/parent.rs
```diff
@@ -553,64 +658,163 @@ fn handle_segfault(
...
+            let catch_segfaults =
+                SegfaultCatchingStuff { ch_pages: &*pages, ch_stack, cs, acc_events };
+            // TODO: This should probably only log writes & not reads, since reads
+            // in these functions will never expose provenance to the rest of the native
+            // code. However, these functions likely won't even do any reads, so...
+            wait_for_signal(Some(pid), signal::SIGTRAP, InitialCont::Yes, Some(catch_segfaults))
+                .unwrap();
```
RalfJung, line 761: What are we even doing here? Who's logging stuff? What does that have to do with exposing anything? I am confused.

The commit message says "Handle accesses during libc fn calls also". Is this something we have to do for some reason, or something you thought would be better to do but it's not mandatory?

src/shims/native_lib/trace/parent.rs
```diff
@@ -553,64 +658,163 @@ fn handle_segfault(
...
+
+            // Unset the breakpoint stuff and move the ip back an instruction to compensate.
```
RalfJung, line 763: As in, undo the ret_addr editing we did above? Right?

src/shims/native_lib/trace/parent.rs
```diff
@@ -285,15 +326,63 @@ pub fn sv_loop(
...
+/// Fix up the libc values.
+#[expect(clippy::as_conversions)]
+fn fixup_libc(pid: unistd::Pid) {
+    unsafe {
+        ptrace::write(
+            pid,
+            libc::malloc as *mut _,
+            (libc::malloc as *mut libc::c_long).read_volatile(),
+        )
+        .unwrap();
```
RalfJung, line 356: Lol you're just copying the bytes from the parent (where we didn't change them) back to the child? :sob: :joy:

src/shims/native_lib/trace/parent.rs
```diff
@@ -285,15 +326,63 @@ pub fn sv_loop(
...
+            ExecEvent::Died(code) => return Err(ExecEnd(code)),
         }
     }
 
     unreachable!()
 }
 
+/// Set up SIGTRAPs on the first few bytes of malloc/free/etc.
+#[expect(clippy::as_conversions)]
+fn trap_libc(pid: unistd::Pid) {
```
RalfJung, line 338: I think it would be good to consistently use debugger terminology here: we are setting and unsetting breakpoints.

src/shims/native_lib/trace/parent.rs
```diff
@@ -553,64 +658,163 @@ fn handle_segfault(
...
+                if mp.filename().iter().any(|&name| name == libc_name) {
+                    (mp.start()..mp.start().strict_add(mp.size())).contains(&ret_addr)
+                } else {
+                    false
+                }
+            });
+            ptrace::write(pid, RET_IS_LIBC.as_ptr().cast(), ret_is_libc.into()).unwrap();
+
+            // Override the return address to give us another sigtrap
+            // (but save the original bytes).
```
RalfJung, line 750: This is another case of setting (and later unsetting) a breakpoint, right?

RalfJung (commented):
Thanks for the PR! I can already see I have to brace myself for some more abstraction-breaking shenanigans.^^ :)

Here's a first round of comments. As you said there's also conflicts to be resolved and CI failures to be looked into.

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4792/5779ab98cccdacb405c2b8b1ab611b0d40aa471d..966775e15dc69370d1a461229ae92357efea9374)*

Cargo.toml
nia-e: Is it? It seems like the only dependency on linux is libc: https://github.com/rbspy/proc-maps/blob/master/Cargo.toml

RalfJung: The lockfile has a load of new stuff in it -- that has to come from somewhere?

javierhonduco: > The lockfile has a load of new stuff in it -- that has to come from somewhere?

I believe the additional lockfile stuff comes from [this windows dependency](https://github.com/rbspy/proc-maps/blob/4977c4a3952001263e68fd6f755d898b2a9f1323/Cargo.toml#L20). While it's not compiled or downloaded in non-windows hosts, the lockfile tracks all the different targets I believe.

Since this is a Linux-only feature if I understand correctly, the `procfs` crate without any default features would work too. It pulls some other dependencies and it's slightly larger but if in the future miri needs other linux only data it could be useful

```
(nix:nix-shell-env) $ cargo tree 
proc-maps v0.1.0 (/home/fhonduvillac/src/miri-experiments/proc-maps)
└── proc-maps v0.5.0
    └── libc v0.2.189
(nix:nix-shell-env) $ ls -lah target/release/proc-maps
-rwxrwxr-x 2 fhonduvillac fhonduvillac 470K Jul 28 09:42 target/release/proc-maps
```
```
(nix:nix-shell-env) $ cargo tree
procfs v0.1.0 (/home/fhonduvillac/src/miri-experiments/procfs)
└── procfs v0.18.0
    ├── bitflags v2.13.1
    ├── procfs-core v0.18.0
    │   ├── bitflags v2.13.1
    │   └── hex v0.4.3
    └── rustix v1.1.4
        ├── bitflags v2.13.1
        └── linux-raw-sys v0.12.1
(nix:nix-shell-env) $ ls -lah target/release/procfs
-rwxrwxr-x 2 fhonduvillac fhonduvillac 519K Jul 28 09:42 target/release/procfs
```

src/shims/native_lib/trace/child.rs
```diff
@@ -54,24 +54,37 @@ impl Supervisor {
...
         // As a side-effect, even on platforms where ptracing
         // is not implemented, we enforce that only one FFI call
         // happens at a time.
         let Some(sv) = sv_guard.as_mut() else { return f().map(|v| (v, None)) };
 
+        // Save the machine pointer to a location where the libc interceptors can use it,
+        // since we can't pass in arguments.
+        super::parent::MACHINE_PTR.store(machine_ptr.cast(), std::sync::atomic::Ordering::Relaxed);
+        // Give the libc interceptors the event channel.
+        let mut e_rx = super::parent::EVT_RX.lock().unwrap();
```
javierhonduco, line 81: nit: Is there a reason why this error is not being handled? Just curious!

javierhonduco (commented):
Thank you so much for working on this, I am very interested in this PR! Left some drive-by comments, hope that's okay but feel free to ignore them. 

I maintain a bunch of Rust libraries with a significant amount of ffi code, so very excited to have better support for native code in the future 🎉 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4792/5779ab98cccdacb405c2b8b1ab611b0d40aa471d..966775e15dc69370d1a461229ae92357efea9374)*

PR #4805, opened 2026-01-04
Add a validity testcase for uninhabited variants

meithecatte:
Found this testcase in a random unmerged branch of mine. Looks like I wrote it as part of my work on rust-lang/rust#138961, and then forgot to submit it when plans wrt the scope of that PR have changed.

I have double checked by grepping for the error message and it looks like we still don't have any similar tests in miri.

### DIFF
--- a/tests/fail/validity/uninhabited_variant.rs
+++ b/tests/fail/validity/uninhabited_variant.rs
@@ -1,0 +1,17 @@
+// NOTE: this is essentially a smoke-test, with more comprehensive tests living in the rustc
+// repository at tests/ui/consts/const-eval/ub-enum.rs
+#![feature(never_type)]
+
+#[repr(C)]
+#[allow(dead_code)]
+enum E {
+    V1,    // discriminant: 0
+    V2(!), // 1
+}
+
+fn main() {
+    unsafe {
+        std::mem::transmute::<u32, E>(1);
+        //~^ ERROR: encountered an uninhabited enum variant
+    }
+}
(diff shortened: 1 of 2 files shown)


tests/fail/validity/uninhabited_variant.rs
```diff
@@ -0,0 +1,16 @@
...
+    V1,    // discriminant: 0
+    V2(!), // 1
+}
+
+fn main() {
+    assert_eq!(std::mem::size_of::<E>(), 4);
+
+    let val = 1u32;
+    let ptr = (&raw const val).cast::<E>();
+    unsafe { ptr.read() }; //~ ERROR: encountered an uninhabited enum variant
```
RalfJung: A simple transmute from `1u32` should have the same result, right? That would even take care of the size check.

RalfJung:
> I have double checked by grepping for the error message and it looks like we still don't have any similar tests in miri.

Yeah, we probably mostly test this via const-eval tests... but can't hurt to also have a Miri test for this.

tests/fail/validity/uninhabited_variant.rs
meithecatte: I'm not sure, I feel like you'd have a better understanding of what would happen here. Should I change it over to a transmute?

RalfJung: Yeah, I think that would make this easier to understand.

meithecatte:
So, having looked into this more closely, there's already a quite comprehensive rustc test for this at `tests/ui/consts/const-eval/ub-enum.rs`, not sure if duplicating it makes sense...

RalfJung:
Having at least a sort of smoke test also in Miri does make sense I think.

meithecatte:
Okay, I've switched it over to a transmute and added a comment to note where the more comprehensive tests can be found.

@rustbot ready

RalfJung:
Thanks!

PR #4806, opened 2026-01-06
Compile-time variant of shim_sig_arg

CraftSpider:
Tried to keep things as readable as possible by splitting out the macro magic into its own helper.

### DIFF
--- a/src/shims/sig.rs
+++ b/src/shims/sig.rs
@@ -28,45 +28,118 @@ pub struct ShimSig<'tcx, const ARGS: usize> {
 /// - `winapi::$ty` for a type from `std::sys::pal::windows::c`
 #[macro_export]
 macro_rules! shim_sig {
-    (extern $abi:literal fn($($arg:ty),* $(,)?) -> $ret:ty) => {
+    (extern $abi:literal fn($($args:tt)*) -> $($ret:tt)*) => {
         |this| $crate::shims::sig::ShimSig {
             abi: std::str::FromStr::from_str($abi).expect("incorrect abi specified"),
-            args: [$(shim_sig_arg!(this, $arg)),*],
-            ret: shim_sig_arg!(this, $ret),
+            args: shim_sig_args_sep!(this, [$($args)*]),
+            ret: shim_sig_arg!(this, $($ret)*),
         }
     };
 }
 
 /// Helper for `shim_sig!`.
+///
+/// Groups tokens into comma-separated chunks and calls the provided macro on them.
+///
+/// # Examples
+///
+/// ```ignore
+/// shim_sig_args_sep!(this, [*const _, i32, libc::off64_t]);
+/// // expands to:
+/// [shim_sig_arg!(*const _), shim_sig_arg!(i32), shim_sig_arg!(libc::off64_t)];
+/// ```
+#[macro_export]
+macro_rules! shim_sig_args_sep {
+    ($this:ident, [$($tt:tt)*]) => {
+        shim_sig_args_sep!(@ $this [] [] $($tt)*)
+    };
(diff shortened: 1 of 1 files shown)


src/shims/sig.rs
```diff
@@ -28,44 +28,96 @@ pub struct ShimSig<'tcx, const ARGS: usize> {
...
+            ret: shim_sig_arg!(this, $($ret)*),
         }
     };
 }
 
 /// Helper for `shim_sig!`.
+///
+/// Groups tokens into comma-separated chunks and calls the provided macro on them.
+#[macro_export]
+macro_rules! _comma_sep {
```
CraftSpider: This thing is a classic tt-muncher, which you can read up on elsewhere. But for a quick explanation here: It basically consumes the first token of the input, appends some tokens to a 'collector' argument as relevant, then at the end emits the collected tokens as its output.
You need the collector since a macro can't expand to an invalid expression, so you can't have `[foo!() bar!()]` where foo expands to `foo,` and bar expands to `bar,`.

A tt-muncher can have not great performance implications, since it's deeply recursive, but that's not likely to be an issue here since we're doing a linear scan and operating on relatively small chunks of tokens, all told. On the other hand, they're capable of basically arbitrary parsing operations since they are both stateful and not subject to the follow-rule and grouping limitations of syntactical matchers. They're basically the step in between a macro_rules and a proc macro.

RalfJung: Could you extend the doc comment to explain the usage of the macro, and have an internal comment explaining the grammar of the intermediate states (the `@ $mac $this [...] [...] ...` part)?

RalfJung: Also, with its `$this` argument and the output and output being `[...]` lists, I think it's not worth using a very generic name... please just call it `shim_sig_args_sep` or so.

RalfJung:
Very neat, thanks :)

src/shims/sig.rs
```diff
@@ -28,44 +28,115 @@ pub struct ShimSig<'tcx, const ARGS: usize> {
...
 /// Helper for `shim_sig!`.
+///
+/// Groups tokens into comma-separated chunks and calls the provided macro on them.
+///
+/// # Examples
+///
+/// ```
+/// shim_sig_args_sep!(this, [*const _, i32, libc::off64_t], shim_sig_arg)
+/// // expands to:
+/// [this.machine.layouts.const_raw_ptr.ty, this.tcx.types.i32, helpers::path_ty_layout(this, &["libc", "off64_t").ty]
```
RalfJung:
```suggestion
/// [shim_sig_arg!(*const _), shim_sig_arg!(i32), shim_sig_arg!(libc::off64_t)]
```

src/shims/sig.rs
```diff
@@ -28,44 +28,115 @@ pub struct ShimSig<'tcx, const ARGS: usize> {
...
+/// ```
+#[macro_export]
+macro_rules! shim_sig_args_sep {
+    ($this:ident, [$($tt:tt)*], $mac:ident) => {
```
RalfJung: Since `mac` is always `shim_sig_arg`, is there even a reason to make it a parameter?

RalfJung:
CI is also red.
@rustbot author

RalfJung:
This looks great, thanks! Please rebase to resolve the conflicts and squash the commits. You can squash manually if there are multiple independent commits you want to preserve, or use `./miri squash` (make sure to pick a suitable commit message). Then write `@rustbot ready` after you force-pushed the squashed PR.

@rustbot author

CraftSpider:
@rustbot ready

Assuming CI passes

Issue #4804, opened 2026-01-04
Add miri specific extern function for spawning a new thread in no_std projects


loftyinclination:
I'm attempting to run Miri in a `no_std` context, (testing some bindings against an FFI [embedded `thumbv7em-none-eabihf` C project](https://github.com/flipperzero-rs/flipperzero-rs/issues/100), for which we're attempting to get around the limitation of Miri working with FFI bindings by reimplementing the `extern C` methods in Rust), and have recently hit a roadblock. While we were able to get around the lack of std library implementation of panic handlers by [following the example in some internal Miri tests](https://github.com/rust-lang/miri/blob/master/tests/fail/panic/no_std.rs#L16), and the lack of alloc support by [utilizing the native shims in `miri_alloc` and `miri_dealloc`](https://github.com/rust-lang/miri/blob/master/tests/utils/miri_extern.rs#L117), there's no equivalent example for `thread::spawn`. 

I believe that it should be possible to create another `extern` function in https://github.com/rust-lang/miri/blob/master/tests/utils/miri_extern.rs, which would call into https://github.com/rust-lang/miri/blob/master/src/concurrency/thread.rs#L534, similar to how the windows and unix implementations handle things. 

Would this be a sensible way to go? Would this patch be welcome?

RalfJung:
That would be possible, yes. But is it realistic? How do you usually spawn threads on your target?

bjorn3:
It looks like the Flipper Zero has its own RTOS thing called Furi. Calling `furi_thread_alloc_ex` seems to be one option to spawn a thread.

loftyinclination:
I've not been eager to go down the route of bringing in the Furi/FreeRTOS implementations, mainly bcs I wasn't sure if MIRI would be able to track the allocations, and also because it seems somewhat out of scope to bring in and set up the entire OS on a MIRI entry point -- [projects on the flipper](https://github.com/flipperzero-rs/flipperzero-rs) tend not to deal with their launching, or with how any of the other threads (GUI, Bluetooth, ect) are running.

Even if we did use FreeRTOS to spawn the threads, wouldn't we still need to patch MIRI to have shims for that OS? I wasn't sure if that was worth maintaining compared to the simpler single extern function.

RalfJung:
I see. Yeah we probably do not want to have shims for all RTOSes out there, so if adjusting the code on your end using  a miri-"native" thread spawn function works for you, then we can add that.

loftyinclination:
[...]

PR #4807, opened 2026-01-07
Add miri specific extern functions for managing threads

loftyinclination:
See https://github.com/rust-lang/miri/issues/4804

### DIFF
--- a/src/shims/foreign_items.rs
+++ b/src/shims/foreign_items.rs
@@ -2,7 +2,7 @@ use std::collections::hash_map::Entry;
 use std::io::Write;
 use std::path::Path;
 
-use rustc_abi::{Align, CanonAbi, Size};
+use rustc_abi::{Align, CanonAbi, ExternAbi, Size};
 use rustc_ast::expand::allocator::NO_ALLOC_SHIM_IS_UNSTABLE;
 use rustc_data_structures::either::Either;
 use rustc_hir::attrs::Linkage;
@@ -434,6 +434,40 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     this.write_path_to_c_str(Path::new(&path), out, out_size)?;
                 // Return value: 0 on success, otherwise the size it would have needed.
                 this.write_int(if success { 0 } else { needed_size }, dest)?;
+            }
+            "miri_thread_spawn" => {
+                // FIXME: `check_shim_sig` does not work with function pointers.
+                let [start_routine, func_arg] =
+                    this.check_shim_sig_lenient(abi, CanonAbi::Rust, link_name, args)?;
+                let start_routine = this.read_pointer(start_routine)?;
+                let func_arg = this.read_immediate(func_arg)?;
+
+                this.start_regular_thread(
+                    Some(dest.clone()),
+                    start_routine,
+                    ExternAbi::Rust,
+                    func_arg,
+                    this.machine.layouts.unit,
+                )?;
+            }
+            "miri_thread_join" => {
+                let [thread_id] = this.check_shim_sig(
+                    shim_sig!(extern "Rust" fn(usize) -> bool),
+                    link_name,
+                    abi,
+                    args,
+                )?;
+
+                let thread = this.read_target_usize(thread_id)?;
+                if let Ok(thread) = this.thread_id_try_from(thread) {
+                    this.join_thread_exclusive(
+                        thread,
+                        /* success_retval */ Scalar::from_bool(true),
+                        dest,
+                    )?;
+                } else {
+                    this.write_scalar(Scalar::from_bool(false), dest)?;
+                }
             }
             // Hint that a loop is spinning indefinitely.
             "miri_spin_loop" => {
--- a/tests/pass/concurrency/miri_thread_join_spawned.rs
+++ b/tests/pass/concurrency/miri_thread_join_spawned.rs
@@ -1,0 +1,28 @@
+#![no_std]
+#![no_main]
+//@compile-flags: -Cpanic=abort
+//@ignore-target: windows # no-std not supported on Windows
+
+#[path = "../../utils/mod.no_std.rs"]
+mod utils;
+
+extern "Rust" fn thread_start(_null: *mut ()) {
+    unsafe {
+        utils::miri_spin_loop();
+        utils::miri_spin_loop();
+    }
+}
+
+#[no_mangle]
+fn miri_start(_argc: isize, _argv: *const *const u8) -> isize {
+    unsafe {
+        let thread_id = utils::miri_thread_spawn(thread_start, core::ptr::null_mut());
+        assert_eq!(utils::miri_thread_join(thread_id), true);
+    }
+    0
+}
+
+#[panic_handler]
+fn panic_handler(_: &core::panic::PanicInfo) -> ! {
+    loop {}
+}
--- a/tests/utils/miri_extern.rs
+++ b/tests/utils/miri_extern.rs
@@ -156,6 +156,19 @@ extern "Rust" {
     /// Blocks the current execution if the argument is false
     pub fn miri_genmc_assume(condition: bool);
 
+    /// Miri-provided extern function to spawn a new thread in the interpreter.
+    ///
+    /// Returns the thread id.
+    ///
+    /// This is useful when no fundamental way of spawning threads is available, e.g. when using
+    /// `no_std`.
+    pub fn miri_thread_spawn(t: extern "Rust" fn(*mut ()), data: *mut ()) -> usize;
+
+    /// Miri-provided extern function to join a thread that was spawned by Miri.
+    pub fn miri_thread_join(thread_id: usize) -> bool;
+
     /// Indicate to Miri that this thread is busy-waiting in a spin loop.
+    ///
+    /// As far as Miri is concerned, this is equivalent to `yield_now`.
     pub fn miri_spin_loop();
 }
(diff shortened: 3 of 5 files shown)


oli-obk:
Please also add a no_std test using this new function directly

src/shims/foreign_items.rs
```diff
@@ -435,6 +435,40 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // Return value: 0 on success, otherwise the size it would have needed.
                 this.write_int(if success { 0 } else { needed_size }, dest)?;
             }
+            "miri_thread_spawn" => {
+                let [start_routine, func_arg] =
+                    this.check_shim_sig_lenient(abi, CanonAbi::Rust, link_name, args)?;
```
RalfJung, line 441: Can you use the non-lenient version for new shims?
There are examples [here](https://github.com/rust-lang/miri/blob/0bbabdc126455a85beea71a352ade605cab269e0/src/shims/unix/foreign_items.rs).

src/shims/foreign_items.rs
```diff
@@ -435,6 +435,40 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    func_arg,
+                    this.machine.layouts.unit,
+                )?;
+            }
+            "miri_thread_join" => {
+                let [thread_id] =
+                    this.check_shim_sig_lenient(abi, CanonAbi::Rust, link_name, args)?;
+
+                let thread =
+                    this.read_scalar(thread_id)?.to_int(this.machine.layouts.usize.size)?;
```
RalfJung:
```suggestion
                    this.read_target_usize(thread_id)?;
```

tests/utils/miri_extern.rs
```diff
@@ -156,6 +156,20 @@ extern "Rust" {
...
+    /// Returns the thread id.
+    ///
+    /// This is useful when no fundamental way of spawning threads is available, e.g. when using
+    /// `no_std`.
+    pub fn miri_thread_spawn(t: extern "Rust" fn(*mut ()), data: *mut ()) -> usize;
+
+    /// Miri-provided extern function to join a thread that was spawned by Miri.
+    pub fn miri_thread_join(thread_id: usize) -> bool;
+
+    /// Miri-provided extern function to yield a thread.
```
RalfJung:
```suggestion
    /// Miri-provided extern function to yield the current thread.
```

tests/fail/concurrency/miri_thread_join_main.rs
RalfJung, line 1: We already have these tests for the other spawn functions and it's the same implementation... so I don't think we need these "fail" tests (and tests do cost execution time and maintenance after all).

loftyinclination: Fair enough -- if you're satisfied with the pass tests (and don't think there need to be any others), I'm happy.

src/shims/foreign_items.rs
loftyinclination: I've run into an issue with using the non-lenient version for the function pointer -- `unsupported signature type "extern \"Rust\" fn(*mut ()) -> ()"`

(as an aside, i'm not able to use `./miri run` for no_std. `/miri test` runs fine, but `run` fails with
```
error: unwinding panics are not supported without std
  |
  = help: using nightly cargo, use -Zbuild-std with panic="abort" to avoid unwinding
  = note: since the core library is usually precompiled with panic="unwind", rebuilding your crate with panic="abort" may not be enough to fix the problem
```
which is also the case for already existing tests such as tests/fail/panic/no_std.rs)

RalfJung: > (as an aside, i'm not able to use ./miri run for no_std. /miri test runs fine, but run fails with

Yeah, you need to pass the flags manually then. `//@compile-flags` etc only affetcs `./miri test`, not `./miri run`.

RalfJung: > I've run into an issue with using the non-lenient version for the function pointer -- unsupported signature type "extern \"Rust\" fn(*mut ()) -> ()"

Ah... right, function pointers will be tricky. So for this one please add a FIXME, but use the non-lenient version for the other 2.

loftyinclination: Done. I had to add support for `bool` to `shim_sig!` for `fn miri_thread_join(usize) -> bool`, not sure if that is a sign that the signature should be changed, or if adding that support was fine?

RalfJung: Adding `bool` is fine :)

tests/utils/miri_extern.rs
```diff
@@ -156,6 +156,20 @@ extern "Rust" {
...
+    pub fn miri_thread_spawn(t: extern "Rust" fn(*mut ()), data: *mut ()) -> usize;
+
+    /// Miri-provided extern function to join a thread that was spawned by Miri.
+    pub fn miri_thread_join(thread_id: usize) -> bool;
+
+    /// Miri-provided extern function to yield the current thread.
+    pub fn miri_thread_yield();
+
     /// Indicate to Miri that this thread is busy-waiting in a spin loop.
     pub fn miri_spin_loop();
```
RalfJung: Does it really make sense to have both of these? They do the exact same thing at the moment...

loftyinclination: I think so -- they've got different meanings outside of miri, I would personally consider that `miri_spin_loop` does the same thing to be an implementation detail. This is also somewhat similar to what the core method does;

> This function is different from [thread::yield_now](https://doc.rust-lang.org/stable/std/thread/fn.yield_now.html) which directly yields to the system’s scheduler, whereas spin_loop does not interact with the operating system.

> Note: On platforms that do not support receiving spin-loop hints this function does not do anything at all.

RalfJung: I can't quite imagine a future for Miri where we treat these differently, so I am inclined to say they should be merged. I do not have a strong opinion on how exactly that merged operation should be called.

How do you plan to use "yield", and why would "spin_loop" not make sense there?

loftyinclination: If we're going to merge them, I think that keeping the existing name so that there's no breaking changes would make sense.

I think that you're right though, and that most places that this would be used, spin_loop would make sense.

RalfJung: Okay, let's just keep miri_spin_loop and remove the new one then.

loftyinclination:
Ok, sounds good. Dyou think a comment on spin_loop mentioning yield_now would be useful, is that just duplicating logic from elsewhere?


________________________________

tests/pass/concurrency/miri_thread_join_spawned.rs
```diff
@@ -0,0 +1,31 @@
+#![no_std]
+#![no_main]
+//@compile-flags: -Cpanic=abort
+//@ignore-target: windows # no-std not supported on Windows
+
+extern "Rust" {
+    fn miri_thread_spawn(t: extern "Rust" fn(*mut ()), data: *mut ()) -> usize;
+    fn miri_thread_join(thread_id: usize) -> bool;
+    fn miri_spin_loop();
+}
```
RalfJung: Please import the utils module instead of repeating the definitions (following what other tests do).

RalfJung:
> Ok, sounds good. Dyou think a comment on spin_loop mentioning yield_now would be useful, is that just duplicating logic from elsewhere?

A comment sounds good, yeah.

tests/pass/concurrency/miri_thread_join_spawned.rs
loftyinclination: Okay :) 

I'd missed this, since I was looking at `tests/pass/miri_alloc.rs`. is there a reason that test isn't using imports in the way you're suggesting? would it be helpful if I changed that too?

RalfJung: The test probably predates that file. Sure, feel free to adjust it as well :)

loftyinclination: All done

tests/utils/miri_extern.rs
```diff
@@ -156,6 +156,19 @@ extern "Rust" {
...
+    /// This is useful when no fundamental way of spawning threads is available, e.g. when using
+    /// `no_std`.
+    pub fn miri_thread_spawn(t: extern "Rust" fn(*mut ()), data: *mut ()) -> usize;
+
+    /// Miri-provided extern function to join a thread that was spawned by Miri.
+    pub fn miri_thread_join(thread_id: usize) -> bool;
+
     /// Indicate to Miri that this thread is busy-waiting in a spin loop.
+    ///
+    /// As far as Miri is concerned, this is alternative in function to yield_now.
```
RalfJung:
```suggestion
    /// As far as Miri is concerned, this is equivalent to `yield_now`.
```

src/shims/foreign_items.rs
```diff
@@ -435,6 +435,41 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // Return value: 0 on success, otherwise the size it would have needed.
                 this.write_int(if success { 0 } else { needed_size }, dest)?;
             }
+            "miri_thread_spawn" => {
+                // TODO: switch to non-lenient check when shim_sig! macro supports function
+                // pointers
```
RalfJung:
```suggestion
                // FIXME: `check_shim_sig` does not work with function pointers.
```

PR #4808, opened 2026-01-07
add `pmaddwd` shim

folkertdev:
Previously `stdarch` implemented this function in terms of primitives miri can already evaluate. To work around a performance regression `stdarch` now uses the llvm intrinsic again. In zlib-rs we rely on miri being able to evaluate this function.

I've already fixed the LLVM bug, so in LLVM 22 we should be able to remove the shim, but the test coverage is still useful to ensure there are no further regressions.

I've only hooked up the avx512 version. I could add the sse2/avx2 variants too, but if nobody has complained about them yet it's just adding code that we'll remove in ~2 months.

I validated the test on real hardware.

cc https://github.com/rust-lang/stdarch/pull/1985

### DIFF
--- a/src/shims/x86/avx2.rs
+++ b/src/shims/x86/avx2.rs
@@ -6,7 +6,7 @@ use rustc_target::callconv::FnAbi;
 
 use super::{
     ShiftOp, horizontal_bin_op, mpsadbw, packssdw, packsswb, packusdw, packuswb, permute, pmaddbw,
-    pmulhrsw, psadbw, pshufb, psign, shift_simd_by_scalar,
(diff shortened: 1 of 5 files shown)


usamoi:
In case you didn't notice, SSE2 and AVX2 variants of `pmaddwd` have been already reimplemented; it's only the AVX512 variant that's missing.

folkertdev:
Thanks, I had missed that completely. I've now unified the implementations.

RalfJung:
LGTM, thanks!

I assume you tested this on real hardware?

folkertdev:
I did, yes.

PR #4812, opened 2026-01-11
Support fstat on non-file-backed FDs

### LINKED ISSUE #4753: Support fstat on non-file-backed FDs
`fstat` is supposed to work on all FDs, but we currently only support it on file-backed FDs.

I am not sure what exactly it should do for other FDs; figuring that out would be the first step in working on this issue. So to work on this you should be willing to do some research and explain the results so that we can then together figure out how to best implement this in Miri. However, I don't expect this to be complicated from the Miri side, so prior experience with Miri is not required.

enthropy7:
Implement fstat support for sockets, pipes, eventfd, epoll, and stdin/stdout/stderr file descriptors. Fixes scalar size mismatch error by using to_uint with scalar's own size instead of to_u32().

Fixes rust-lang/miri#4753

### DIFF
--- a/src/shims/files.rs
+++ b/src/shims/files.rs
@@ -1,6 +1,6 @@
 use std::any::Any;
 use std::collections::BTreeMap;
-use std::fs::{File, Metadata};
+use std::fs::File;
 use std::io::{ErrorKind, IsTerminal, Read, Seek, SeekFrom, Write};
 use std::marker::CoercePointee;
 use std::ops::Deref;
@@ -9,7 +9,7 @@ use std::{fs, io};
 
 use rustc_abi::Size;
 
-use crate::shims::unix::UnixFileDescription;
+use crate::shims::unix::{FileMetadata, UnixFileDescription};
 use crate::*;
 
 /// A unique id for file descriptions. While we could use the address, considering that
@@ -209,8 +209,21 @@ pub trait FileDescription: std::fmt::Debug + FileDescriptionExt {
         throw_unsup_format!("cannot close {}", self.name());
     }
 
-    fn metadata<'tcx>(&self) -> InterpResult<'tcx, io::Result<fs::Metadata>> {
+    /// Returns the host `fs::Metadata` for this FD, if available.
+    /// Used by host-aware shims like Windows's `GetFileInformationByHandle`.
+    /// Unrelated to Unix `fstat`, which goes through `fstat()`.
+    fn host_metadata<'tcx>(&self) -> InterpResult<'tcx, io::Result<fs::Metadata>> {
         throw_unsup_format!("obtaining metadata is only supported on file-backed file descriptors");
+    }
+
+    /// Return the metadata describing this FD for the `fstat`/`statx` family of syscalls.
+    /// File-backed FDs should call `FileMetadata::from_meta` with their host metadata.
+    /// Non-file-backed FDs should call `FileMetadata::synthetic` with an appropriate mode.
+    fn fstat<'tcx>(
+        &self,
+        _ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, Result<FileMetadata, IoError>> {
+        throw_unsup_format!("fstat is not supported on {}", self.name());
     }
 
     fn is_tty(&self, _communicate_allowed: bool) -> bool {
@@ -432,8 +445,15 @@ impl FileDescription for FileHandle {
         }
     }
 
-    fn metadata<'tcx>(&self) -> InterpResult<'tcx, io::Result<Metadata>> {
+    fn host_metadata<'tcx>(&self) -> InterpResult<'tcx, io::Result<fs::Metadata>> {
         interp_ok(self.file.metadata())
+    }
+
+    fn fstat<'tcx>(
+        &self,
+        ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, Result<FileMetadata, IoError>> {
+        FileMetadata::from_meta(ecx, self.file.metadata())
     }
 
     fn is_tty(&self, communicate_allowed: bool) -> bool {
--- a/src/shims/files.rs
+++ b/src/shims/files.rs
@@ -9,7 +9,7 @@ use std::{fs, io};
 
 use rustc_abi::Size;
 
-use crate::shims::unix::{FileMetadata, UnixFileDescription};
+use crate::shims::unix::UnixFileDescription;
 use crate::*;
 
 /// A unique id for file descriptions. While we could use the address, considering that
@@ -209,21 +209,12 @@ pub trait FileDescription: std::fmt::Debug + FileDescriptionExt {
         throw_unsup_format!("cannot close {}", self.name());
     }
 
-    /// Returns the host `fs::Metadata` for this FD, if available.
-    /// Used by host-aware shims like Windows's `GetFileInformationByHandle`.
-    /// Unrelated to Unix `fstat`, which goes through `fstat()`.
-    fn host_metadata<'tcx>(&self) -> InterpResult<'tcx, io::Result<fs::Metadata>> {
+    /// Returns the metadata for this FD, if available.
+    /// This is either host metadata, or a non-file-backed-FD type.
+    /// The latter is for new represented as a string storing a `libc` name so we only
+    /// support that kind of metadata on Unix targets.
+    fn metadata<'tcx>(&self) -> InterpResult<'tcx, Either<io::Result<fs::Metadata>, &'static str>> {
         throw_unsup_format!("obtaining metadata is only supported on file-backed file descriptors");
-    }
-
-    /// Return the metadata describing this FD for the `fstat`/`statx` family of syscalls.
-    /// File-backed FDs should call `FileMetadata::from_meta` with their host metadata.
-    /// Non-file-backed FDs should call `FileMetadata::synthetic` with an appropriate mode.
-    fn fstat<'tcx>(
-        &self,
-        _ecx: &mut MiriInterpCx<'tcx>,
-    ) -> InterpResult<'tcx, Result<FileMetadata, IoError>> {
-        throw_unsup_format!("fstat is not supported on {}", self.name());
     }
 
     fn is_tty(&self, _communicate_allowed: bool) -> bool {
@@ -445,15 +436,8 @@ impl FileDescription for FileHandle {
         }
     }
 
-    fn host_metadata<'tcx>(&self) -> InterpResult<'tcx, io::Result<fs::Metadata>> {
-        interp_ok(self.file.metadata())
-    }
-
-    fn fstat<'tcx>(
-        &self,
-        ecx: &mut MiriInterpCx<'tcx>,
-    ) -> InterpResult<'tcx, Result<FileMetadata, IoError>> {
-        FileMetadata::from_meta(ecx, self.file.metadata())
+    fn metadata<'tcx>(&self) -> InterpResult<'tcx, Either<io::Result<fs::Metadata>, &'static str>> {
+        interp_ok(Either::Left(self.file.metadata()))
     }
 
     fn is_tty(&self, communicate_allowed: bool) -> bool {
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -348,34 +348,6 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
     }
 }
 
-fn file_type_to_mode_name(file_type: std::fs::FileType) -> &'static str {
-    #[cfg(unix)]
-    use std::os::unix::fs::FileTypeExt;
-
-    if file_type.is_file() {
-        "S_IFREG"
-    } else if file_type.is_dir() {
-        "S_IFDIR"
-    } else if file_type.is_symlink() {
-        "S_IFLNK"
-    } else {
-        // Certain file types are only available when the host is a Unix system.
-        #[cfg(unix)]
-        {
-            if file_type.is_socket() {
-                return "S_IFSOCK";
-            } else if file_type.is_fifo() {
-                return "S_IFIFO";
-            } else if file_type.is_char_device() {
-                return "S_IFCHR";
-            } else if file_type.is_block_device() {
-                return "S_IFBLK";
-            }
-        }
-        "S_IFREG"
-    }
-}
-
 impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     fn open(
@@ -1662,9 +1634,37 @@ fn extract_sec_and_nsec<'tcx>(
     }
 }
 
+fn file_type_to_mode_name(file_type: std::fs::FileType) -> &'static str {
+    #[cfg(unix)]
+    use std::os::unix::fs::FileTypeExt;
+
+    if file_type.is_file() {
+        "S_IFREG"
+    } else if file_type.is_dir() {
+        "S_IFDIR"
+    } else if file_type.is_symlink() {
+        "S_IFLNK"
+    } else {
+        // Certain file types are only available when the host is a Unix system.
+        #[cfg(unix)]
+        {
+            if file_type.is_socket() {
+                return "S_IFSOCK";
+            } else if file_type.is_fifo() {
+                return "S_IFIFO";
+            } else if file_type.is_char_device() {
+                return "S_IFCHR";
+            } else if file_type.is_block_device() {
+                return "S_IFBLK";
+            }
+        }
+        "S_IFREG"
+    }
+}
+
 /// Stores a file's metadata in order to avoid code duplication in the different metadata related
 /// shims.
-pub struct FileMetadata {
+struct FileMetadata {
     mode: Scalar,
     size: u64,
     created: Option<(u64, u32)>,
@@ -1694,18 +1694,20 @@ impl FileMetadata {
         let Some(fd) = ecx.machine.fds.get(fd_num) else {
             return interp_ok(Err(LibcError("EBADF")));
         };
-        fd.fstat(ecx)
+        match fd.metadata()? {
+            Either::Left(host) => Self::from_meta(ecx, host),
+            Either::Right(name) => Self::synthetic(ecx, name),
+        }
     }
 
-    pub(crate) fn synthetic<'tcx>(
+    fn synthetic<'tcx>(
         ecx: &mut MiriInterpCx<'tcx>,
         mode_name: &str,
-        size: u64,
     ) -> InterpResult<'tcx, Result<FileMetadata, IoError>> {
         let mode = ecx.eval_libc(mode_name);
         interp_ok(Ok(FileMetadata {
             mode,
-            size,
+            size: 0,
             created: None,
             accessed: None,
             modified: None,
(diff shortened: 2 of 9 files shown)


src/shims/unix/fs.rs
```diff
@@ -128,17 +128,20 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-        let mode = metadata.mode.to_uint(this.libc_ty_layout("mode_t").size)?;
 
         // We do *not* use `deref_pointer_as` here since determining the right pointee type
         // is highly non-trivial: it depends on which exact alias of the function was invoked
         // (e.g. `fstat` vs `fstat64`), and then on FreeBSD it also depends on the ABI level
         // which can be different between the libc used by std and the libc used by everyone else.
         let buf = this.deref_pointer(buf_op)?;
+
+        // Extract mode value. write_int will handle size conversion when writing to st_mode field.
+        let mode = metadata.mode.to_uint(metadata.mode.size())?.try_into().unwrap_or(0u32);
```
RalfJung: This basically bypasses the sanity check that `made` has the right type. Why is this needed?

src/shims/unix/fs.rs
```diff
@@ -1621,9 +1624,126 @@ impl FileMetadata {
...
+        }
+    }
+
+    /// Create synthetic metadata for non-file-backed file descriptors.
+    fn for_non_file_fd<'tcx>(
+        ecx: &mut MiriInterpCx<'tcx>,
+        fd_name: &str,
+    ) -> InterpResult<'tcx, Result<FileMetadata, IoError>> {
+        // Determine the file type and mode based on the FD name
+        let (mode_name, size) = match fd_name {
```
RalfJung: The name is entirely a diagnostic tool. It should never be matched on in the program itself -- this will silently break if we change the string or add a typo. It also will silently do the wrong thing when new FD types are added.

Instead, the `metadata` method on those FD types should be adjusted to return suitable metadata.

src/shims/unix/fs.rs
```diff
@@ -1621,9 +1624,126 @@ impl FileMetadata {
...
+
+        // Start with the base file type (S_IFSOCK, S_IFIFO, etc.)
+        // Use eval_libc to get a Scalar with the correct size, just like in from_meta
+        let mode_base = ecx.eval_libc(mode_name);
+
+        // Add default permissions (0666 for sockets/pipes, 0600 for others).
+        // Convert Scalars to u32 using to_uint with their own size, then combine with bitwise OR.
+        let mode_base_u32 = mode_base.to_uint(mode_base.size())?.try_into().unwrap_or(0u32);
+        let permissions_u32 = if mode_name == "S_IFSOCK" || mode_name == "S_IFIFO" {
+            ecx.eval_libc("S_IRUSR")
```
RalfJung: All these look like they could just use `eval_libc_u32`.

tests/pass-dep/libc/libc-fstat-non-file.rs
```diff
@@ -0,0 +1,278 @@
...
+    let _st_dev = stat.st_dev;
+    let _st_uid = stat.st_uid;
+    let _st_gid = stat.st_gid;
+    let _st_rdev = stat.st_rdev;
+    let _st_atime = stat.st_atime;
+    let _st_mtime = stat.st_mtime;
+    let _st_ctime = stat.st_ctime;
+    let _st_atime_nsec = stat.st_atime_nsec;
+    let _st_mtime_nsec = stat.st_mtime_nsec;
+    let _st_ctime_nsec = stat.st_ctime_nsec;
```
RalfJung: Please move this into a function instead of repeating the code 3 times.

tests/pass-dep/libc/libc-fstat-non-file.rs
```diff
@@ -0,0 +1,278 @@
...
+    let stat = unsafe { stat.assume_init_ref() };
+
+    // stdout is typically a character device (S_IFCHR) or a regular file
+    // We just check that it's not an error
+    let file_type = stat.st_mode & libc::S_IFMT;
+    assert!(
+        file_type == libc::S_IFCHR || file_type == libc::S_IFREG,
+        "stdout should be S_IFCHR or S_IFREG, got {:#o}",
+        file_type
+    );
```
RalfJung: In this test we know there has been no output redirection to a file. So it can never be a regular file here, can it?

src/shims/unix/fs.rs
```diff
@@ -1621,9 +1624,126 @@ impl FileMetadata {
...
+        cfg_select! {
+            unix => {
+                use std::os::unix::fs::MetadataExt;
+                // Get current user/group IDs from a temporary file or current directory
+                let (dev, uid, gid) = match std::fs::metadata("/tmp")
+                    .or_else(|_| std::fs::metadata("."))
+                {
+                    Ok(temp_meta) => {
+                        (temp_meta.dev(), temp_meta.uid(), temp_meta.gid())
+                    }
```
RalfJung: This seems like a bad idea -- `'/tmp` and `.` have nothing to do with this FD so the values can easily be nonsense, and the overall behavior can be very confusing.

src/shims/unix/fs.rs
```diff
@@ -1621,9 +1624,126 @@ impl FileMetadata {
...
+        };
+
+        // Start with the base file type (S_IFSOCK, S_IFIFO, etc.)
+        // Use eval_libc to get a Scalar with the correct size, just like in from_meta
+        let mode_base = ecx.eval_libc(mode_name);
+
+        // Add default permissions (0666 for sockets/pipes, 0600 for others).
+        // Convert Scalars to u32 using to_uint with their own size, then combine with bitwise OR.
+        let mode_base_u32 = mode_base.to_uint(mode_base.size())?.try_into().unwrap_or(0u32);
+        let permissions_u32 = if mode_name == "S_IFSOCK" || mode_name == "S_IFIFO" {
```
RalfJung: From what I can see we currently entirely ignore the permission part in `FileMetadata::from_meta`. So seems like we could also get away with ignoring it for non-file-FDs for now?

RalfJung (dismissed):
Sorry for the long wait! The test looks good overall, but I think we should use a different implementation strategy. See the comments for details.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4812/f589bbb5bb0bddb29edcdfb4cabcd16575e14b0c..b186092e69bb3009587a5acd2a3d55ef26e40cae)*

enthropy7:
Hi, thanks for your answer! i’m still here, just a little busy with my work. i will answer to your comments as soon as i can and optimise my PR better. I think it would took around 3 days for me to fix this

enthropy7:
I reworked `fstat` for non-file-backed FDs so it no longer depends on host metadata: for `socketpair`/`pipe`/`eventfd`/`epoll` we now return synthetic metadata based on the FD type (correct `S_IF*`, size `0`, and no misleading host `dev/uid/gid`).  
I also removed the Unix-only host-metadata path in `unnamed_socket::metadata` that was breaking Windows CI. removed `fd.name()` matching in `fstat` logic (so it no longer relies on diagnostic strings), applied the requested test cleanup (`libc-fstat-non-file`: extracted repeated checks into a helper and tightened `stdout/stderr` expectations).

upd: fixed a cross-host issue that showed up in CI: on Windows hosts running Linux targets, `fstat` for stdio (`stdin`/`stdout`/`stderr`) could fail because host metadata is unavailable there.  
To make this host-independent, I added a synthetic stdio fallback in the Unix `fstat` path (same approach as for other non-file-backed FDs)

RalfJung:
So I take it you mean this is
@rustbot ready
? Please help us keep track of the PR state by using that command when the PR is ready for review again. :)

src/shims/unix/linux_like/epoll.rs
```diff
@@ -119,6 +119,28 @@ impl FileDescription for Epoll {
         "epoll"
     }
 
+    fn metadata<'tcx>(&self) -> InterpResult<'tcx, io::Result<std::fs::Metadata>> {
+        #[cfg(any(target_os = "linux", target_os = "android"))]
+        {
+            use std::fs::File;
+            use std::os::fd::FromRawFd;
+
+            let fd = unsafe { libc::epoll_create1(libc::EPOLL_CLOEXEC) };
```
RalfJung: I'm not sure what the point of this is, but we shouldn't be calling host operations here.

I don't think there should be any `cfg` usage in this PR. We only use `cfg` in truly exceptional circumstances, and I don't think this is one of them.

RalfJung:
That said, with all the `cfg` usage this is definitely not the right approach. The way we implement these APIs is by reading the documentation (ideally the POSIX documentation, not the glibc/Linux one), and then manually implementing what the docs say. If the docs aren't precise enough, we do experiments to figure out what happens on real systems. Ideally we do these experiments on multiple platforms (in the issue description you said you already did this). Then again we implement manually what the experiments showed the behavior is. Either way, there should be comments in the code explaining how the behavior was determined to be what eventually got implemented.

@rustbot author

src/shims/unix/fs.rs
```diff
@@ -1665,11 +1672,54 @@ impl FileMetadata {
...
+
+        if let Some(anon_socket) = fd.clone().downcast::<AnonSocket>() {
+            return FileMetadata::synthetic(ecx, anon_socket.metadata_mode_name(), 0);
+        }
+        if fd.clone().downcast::<EventFd>().is_some() {
+            return FileMetadata::synthetic(ecx, "S_IFREG", 0);
+        }
+        if fd.clone().downcast::<Epoll>().is_some() {
+            return FileMetadata::synthetic(ecx, "S_IFREG", 0);
+        }
```
RalfJung: No, this is not how you do dynamic dispatch. What are you listing all FD types here? The `fd.metadata` call below invokes the metadata function of the current FD type, that's where all the logic should live.

Meggy43 (approved):

enthropy7:
Reworked approach based on review - replaced the downcast chain and `fd.name()` matching in `from_fd_num` with a s`ynthetic_stat_mode()` trait method, so each FD type provides its own mode via proper dynamic dispatch. Removed all `cfg` blocks, host operations, `/tmp` metadata fetching, and permission bit computation. synthetic metadata now just uses the mode constant and zeros. Re `eval_libc_u32 `for mode: `mode_t` is `u16` on macOS in the libc crate, so `to_u32()` panics. kept `eval_libc `with `Scalar`.

src/shims/files.rs
```diff
@@ -202,6 +202,13 @@ pub trait FileDescription: std::fmt::Debug + FileDescriptionExt {
         throw_unsup_format!("obtaining metadata is only supported on file-backed file descriptors");
     }
 
+    /// If this FD supports fstat but is not backed by a host file, return the
+    /// mode flag name (e.g. "S_IFCHR") for synthetic stat metadata.
+    /// FDs that return `None` (default) will use real host file metadata via `metadata()`.
+    fn synthetic_stat_mode(&self) -> Option<&'static str> {
```
RalfJung: This API makes me wonder, why are those separate functions? Why don't have have a single function that returns something like `Either<io::Result<fs::Metadata>, &'static str>`?

Or even better, make it `Result<FileMetadata, IoError>`. Then file-backed FDs will call `FileMetadata::from_meta` and others can call `FileMetadata::synthetic`.

src/shims/unix/fs.rs
```diff
@@ -751,9 +759,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // the owner, its group and other users. Given that we can only provide the file type
         // without using platform specific methods, we only set the bits corresponding to the file
         // type. This should be an `__u16` but `libc` provides its values as `u32`.
-        let mode: u16 = metadata
-            .mode
-            .to_u32()?
+        let mode: u16 = (metadata.mode.to_uint(metadata.mode.size())? & u128::from(u16::MAX))
```
RalfJung: What is the point of this change? (I have [asked](https://github.com/rust-lang/miri/pull/4812#discussion_r2807427433) about this before.)

src/shims/unix/fs.rs
```diff
@@ -1681,13 +1709,38 @@ impl FileMetadata {
...
+                    "S_IFSOCK"
+                } else if file_type.is_fifo() {
+                    "S_IFIFO"
+                } else if file_type.is_char_device() {
+                    "S_IFCHR"
+                } else if file_type.is_block_device() {
+                    "S_IFBLK"
+                } else {
+                    "S_IFREG"
+                };
```
RalfJung: Please have a look at [this](https://github.com/rust-lang/miri/blob/2396c974e47ffa8680ec0e65443afa47c9445973/src/shims/unix/fs.rs#L289-L309) for how to write this case distinction with less code duplication.

src/shims/unix/fs.rs
```diff
@@ -1702,9 +1755,13 @@ impl FileMetadata {
...
-                let dev = metadata.dev();
-                let uid = metadata.uid();
-                let gid = metadata.gid();
+                let is_regular_fs_entry =
+                    file_type.is_file() || file_type.is_dir() || file_type.is_symlink();
+                let (dev, uid, gid) = if is_regular_fs_entry {
+                    (metadata.dev(), metadata.uid(), metadata.gid())
+                } else {
+                    (0, 0, 0)
+                };
```
RalfJung: Why do we actively ignore the metadata info from the host here if this is not a regular file/dir/symlink?

src/shims/files.rs
```diff
@@ -269,6 +279,10 @@ impl FileDescription for io::Stdout {
         "stdout"
     }
 
+    fn synthetic_stat_mode(&self) -> Option<&'static str> {
+        Some("S_IFCHR")
```
RalfJung: We don't actually know that stdin/stdout/stderr are S_IFCHR, right? It could be a file if some redirected the output there.

I'd propose to skip stdin/stdout/stderr in this first PR. We have to either use non-portable APIs to give a correct answer or make a guess, neither of which is great, so until we have a clear usecase IMO it's better to just not support them.

RalfJung (commented):
Sorry for the slow reply. This looks much better, thanks a lot! I still have some suggestions, but we're getting there. :)

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4812/19585c05d5b32a85cc648749e3ac41d20cf00c65..752fafc31d3d3c47389c92c40a2b97b0985cef11)*

src/shims/unix/fs.rs
RalfJung: Oh you replied elsewhere. Please put the replies in the proper threads so that I have a chance to find them.

> Re eval_libc_u32 for mode: mode_t is u16 on macOS in the libc crate, so to_u32() panics. kept eval_libc with Scalar.

Why was this not a problem before? We're already running tests like this on macOS, aren't we?

RalfJung: But the `&`-masking still doesn't make sense to me. This should then be
```rust
metadata.mode.to_uint(this.libc_ty_layout("mode_t").size)?.try_into().unwrap_or_else(...)
```

src/shims/files.rs
enthropy7: Done, merged into a `single fstat()` trait method returning `Result<FileMetadata, IoError>`. `FileHandle` calls `FileMetadata::from_meta`, non-file FDs call `FileMetadata::synthetic`. `FileMetadata` is re-exported through `unix/mod.rs`.

src/shims/files.rs
enthropy7: agreed, dropped stdio support from this PR. Removed the `fstat` impls on `io::Stdin/Stdout/Stderr/NullOutput `and the corresponding tests.

src/shims/unix/fs.rs
enthropy7: replaced the & `u16::MAX` mask with `libc_ty_layout("mode_t").size` to read the scalar at `mode_t's` actual target width (`u16` on macOS, `u32` on Linux). same fix applied in the statx writer. sorry about the misplaced reply earlier.

src/shims/unix/fs.rs
enthropy7: my mistake, removed the `is_regular_fs_entry` zeroing. `from_meta` now always use `MetadataExt::dev/uid/gid` as we always called with real host metadata.

src/shims/unix/fs.rs
enthropy7: extracted a free function `file_type_to_mode_name(FileType) -> &'static str` using `#[cfg(unix)]` per arm, matching the shape of` file_type_to_d_type`. `the cfg_select! `block is gone.

src/shims/files.rs
RalfJung: Now we still have both `metadata` and `fstat`. The point was to merge them, to avoid duplication. Does that not work?

src/shims/unix/fs.rs
```diff
@@ -1698,20 +1736,13 @@ impl FileMetadata {
...
-                let dev = metadata.dev();
-                let uid = metadata.uid();
-                let gid = metadata.gid();
-            }
-            _ => {
-                let dev = 0;
-                let uid = 0;
-                let gid = 0;
-            }
-        }
```
RalfJung, line 1714: Please just preserve the original code here, no need to change it at all (from what I can see).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4812/34e3128ea79093e1b9d16a1a32d1f99fb2ebf78e..a7993d98cacf05a737b80354b8f6a3e8dc5ef919)*

src/shims/files.rs
enthropy7: fully merged now. refactored `FileMetadata` to be cross-platform: mode is stored as `&'static str `name (e.g. `"S_IFREG"`) and resolved to a target-specific `Scalar` only at write time; timestamps are stored as raw `Option<SystemTime>` and converted to Unix-epoch-secs or Windows-epoch-ticks at their respective writers. Single `fstat()` trait method on `FileDescription`, no more `host_metadata`. Windows `GetFileInformationByHandle` now goes through `fstat()` too.

enthropy7: and yes, `FileMetadata` wasn't cross-platform at all. `mode: Scalar` came from `eval_libc`, which panics on Windows. timestamps were kept as (`u64, u32`) in Unix epoch, Windows uses a different approach. fixed both, now `fstat()` is the only method and `GetFileInformationByHandle` goes through it too.

RalfJung: Ah it's because of Windows support -- makes sense.

Thanks for putting in all this work! However, it's generally better to discuss design decisions with the maintainers and get their feedback before running ahead and implement something suboptimal. `FileMetadata` is still not target-independent if it has a `mode_name` field that stores a `libc` name (which doesn't cover Windows). Having the Windows code match on libc names is an abstraction layering violation.

I'll take over from here to avoid further miscommunication.

RalfJung:
I took the version of this PR before your mos recent changes and did a slightly different refactor., to avoid funneling the Windows logic through something from the unix shims. Thanks a lot for all your help!

enthropy7:
> I took the version of this PR before your mos recent changes and did a slightly different refactor., to avoid funneling the Windows logic through something from the unix shims. Thanks a lot for all your help!

Thank you too for collaborating! It was a pleasure to work on Miri. Will try to find new issues when I have free time and work together again. And about FileMetadata - I agree with your approach, I just thought you knew it's different for Windows and Unix and unifying it was on purpose.

RalfJung:
That's fair, but I had indeed entirely forgotten this and was just making guesses in this PR. :)

PR #4813, opened 2026-01-11
add avx512 `pack*` family of instructions to shims

ChillFish8:
Adds the pack family of instructions for the avx512 instruction set aka:

- `_mm512_packs_epi16`
- `_mm512_packus_epi16`
- `_mm512_packs_epi32`
- `_mm512_packus_epi32`

### DIFF
--- a/tests/pass/shims/x86/intrinsics-x86-avx512.rs
+++ b/tests/pass/shims/x86/intrinsics-x86-avx512.rs
@@ -1,6 +1,6 @@
 // We're testing x86 target specific features
 //@only-target: x86_64 i686
-//@compile-flags: -C target-feature=+avx512f,+avx512vl,+avx512bitalg,+avx512vpopcntdq,+avx512vnni
+//@compile-flags: -C target-feature=+avx512f,+avx512vl,+avx512bw,+avx512bitalg,+avx512vpopcntdq,+avx512vnni
 
 #[cfg(target_arch = "x86")]
 use std::arch::x86::*;
@@ -11,12 +11,14 @@ use std::mem::transmute;
 fn main() {
     assert!(is_x86_feature_detected!("avx512f"));
     assert!(is_x86_feature_detected!("avx512vl"));
+    assert!(is_x86_feature_detected!("avx512bw"));
     assert!(is_x86_feature_detected!("avx512bitalg"));
     assert!(is_x86_feature_detected!("avx512vpopcntdq"));
     assert!(is_x86_feature_detected!("avx512vnni"));
 
     unsafe {
         test_avx512();
+        test_avx512bw();
         test_avx512bitalg();
         test_avx512vpopcntdq();
         test_avx512ternarylogic();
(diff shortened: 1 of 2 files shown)


tests/pass/shims/x86/intrinsics-x86-avx512.rs
```diff
@@ -579,6 +581,97 @@ unsafe fn test_avx512vnni() {
     test_mm512_dpbusd_epi32();
 }
 
+#[target_feature(enable = "avx512bw")]
+unsafe fn test_avx512bw() {
+    #[target_feature(enable = "avx512bw")]
+    unsafe fn test_mm512_packs_epi16() {
+        let a = _mm512_set1_epi16(120);
+        let b = _mm512_set1_epi16(200);
```
RalfJung: Please ensure the tests cover saturation in both directions, i.e. for this test it should be towards i8::MAX and i8::MIN.

RalfJung:
Thanks for the PR :)

@rustbot author

tests/pass/shims/x86/intrinsics-x86-avx512.rs
ChillFish8: Good catch, I had forgotten that the behaviour goes the other way.

RalfJung:
Instead of adding a 2nd call, it may be better to actually also check the vector nature of these and have non-homogeneous inputs to cover these extra cases?

Also, please check the tests on real hardware if you haven't already done so.

ChillFish8:
> Instead of adding a 2nd call, it may be better to actually also check the vector nature of these and have non-homogeneous inputs to cover these extra cases?

Maybe, but I think doing the simple broadcasts are simpler for people to read and understand that the instruction does blocks of 128 bits. 

> Also, please check the tests on real hardware if you haven't already done so.

Have done, should be good to go.

RalfJung:
>  but I think doing the simple broadcasts are simpler for people to read and understand that the instruction does blocks of 128 bits.

Without comments I don't think this is clear at all... it took me quite a while to reverse engineer what is even meant by this, and why the tests look the way they do.

ChillFish8:
> Without comments I don't think this is clear at all... it took me quite a while to reverse engineer what is even meant by this, and why the tests look the way they do.

:sweat_smile: Ok fair enough, maybe I have just been staring at these instructions for so long now that I just implicitly associate it with that. 

I think in that case then either solution probably needs a comment, so whatever you prefer if you want it done in these broadcast steps or just combined into one check.

RalfJung:
I think I'd prefer having non-homogeneous inputs, i.e., combining them into one check.

ChillFish8:
Sounds good, I'll update it.

ChillFish8:
@rustbot ready

Updated the tests so it's a single check, and also added some comments so it's more clear how it works and what we expect the behaviour to be.

tests/pass/shims/x86/intrinsics-x86-avx512.rs
```diff
@@ -579,9 +581,148 @@ unsafe fn test_avx512vnni() {
...
 
+#[target_feature(enable = "avx512bw")]
+unsafe fn test_avx512bw() {
+    #[target_feature(enable = "avx512bw")]
+    unsafe fn test_mm512_packs_epi16() {
+        let a = _mm512_set1_epi16(120);
+
+        // Because `packs` instructions do signed saturation, we expect
+        // that any value over `u8::MAX` will be saturated to `i8::MAX`, and any value
+        // less than `u8::MIN` will also be saturated to `i8::MIN`.
```
RalfJung:
```suggestion
        // that any value over `i8::MAX` will be saturated to `i8::MAX`, and any value
        // less than `i8::MIN` will also be saturated to `i8::MIN`.
```

tests/pass/shims/x86/intrinsics-x86-avx512.rs
```diff
@@ -579,9 +581,148 @@ unsafe fn test_avx512vnni() {
...
+            300, 300, 300, 300, 300, 300, 300, 300, -200, -200, -200, -200, -200, -200, -200, -200,
+            300, 300, 300, 300, 300, 300, 300, 300, -200, -200, -200, -200, -200, -200, -200, -200,
+        );
+
+        // The pack* family of instructions in x86 operate in blocks
+        // of 128-bit lanes, meaning the first 128-bit lane in `a` is converted and written
+        // then the first 128-bit lane of `b`, followed by the second 128-bit lane in `a`, etc...
+        // Because we are going from 16-bits to 8-bits our 128-bit block becomes 64-bits in
+        // the output register.
+        // This leaves us with 8x 8-bit values interleaved in the final register.
```
RalfJung: You don't need to repeat this for every function.
```suggestion
        // See `test_mm512_packs_epi16` for an explanation of the output structure.
```

RalfJung:
Thanks! I assume this still passes on real hardware.

It would be nice if you could send the same test updates to stdarch, we are loosely keeping the tests here in sync with the ones there.

ChillFish8:
Can you link where that is? Happy to add it. 

I have a few other PRs in the works atm to add the unpack, compress and expand instructions as well.

RalfJung:
Ah, I thought you took the tests from stdarch. If not, never mind.

But it'd be worth checking if they have more/better tests we may want to add. You should be able to find them by grepping for the intrinsic name in https://github.com/rust-lang/stdarch/ .

RalfJung:
> I have a few other PRs in the works atm to add the unpack, compress and expand instructions as well.

Okay. So far my policy for avx512 intrinsics was that we will accept intrinsics that
- either are just the avx512 version of something we already have, so we can share the logic
- or have a clear motivation in terms of being sufficiently widely used

I am afraid that doing all of avx512 will get out of hand, hence the caution.

ChillFish8:
> either are just the avx512 version of something we already have, so we can share the logic

This should be fine for the unpack variants since they exist for avx2 already.

> or have a clear motivation in terms of being sufficiently widely used

Understandable, there are lot of very niche functions, that being said I think it is work supporting compress/expand as their base forms are part of the base foundation set `avx512f` and probably one of the most used avx512-only instrinsics since you'd want to use it for any sort of filtering or merge two registers.

RalfJung:
stdarch seems to have 2500 intrinsics in the avx512f "base". That's still a bit much. ;)

Anyway this is off-topic here. Probably better to discuss this on [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/269128-miri).

ChillFish8:
Sounds good, thanks for merging!

PR #4818, opened 2026-01-18
windows shims: throw unsupported error for DuplicateHandle on pseudo handle

beepster4096:
`DuplicateHandle` should create a new handle when called on a pseudo handle, instead of just returning it unchanged. Since we don't support that yet, it should error instead.

This PR doesn't add that support like I mentioned in https://github.com/rust-lang/miri/pull/2231#discussion_r2702734879 because that needs a big refactor actually. We'd need a way to track thread handles separately from the threads they point to, as they are all closed separately.

I also changed the `INVALID_HANDLE_VALUE` consts to be calculated instead of hardcoded while I was in the file.

### DIFF
--- a/tests/fail-dep/win/duplicate-pseudo-handle.rs
+++ b/tests/fail-dep/win/duplicate-pseudo-handle.rs
@@ -1,0 +1,22 @@
+//@only-target: windows # Uses win32 api functions
+use windows_sys::Win32::Foundation::{CloseHandle, DUPLICATE_SAME_ACCESS, DuplicateHandle};
+use windows_sys::Win32::System::Threading::{GetCurrentProcess, GetCurrentThread};
+
+fn main() {
+    unsafe {
+        let cur_proc = GetCurrentProcess();
+
+        let pseudo = GetCurrentThread();
+        let mut out = std::mem::zeroed();
+        let res =
+            DuplicateHandle(cur_proc, pseudo, cur_proc, &mut out, 0, 0, DUPLICATE_SAME_ACCESS);
+        //~^ERROR: pseudo handle
+        assert!(res != 0);
+        assert!(out.addr() != 0);
+        // Since the original handle was a pseudo handle, we must return something different.
+        assert!(out != pseudo);
+        // And closing it should work (which it does not for a pseudo handle).
+        let res = CloseHandle(out);
+        assert!(res != 0);
+    }
+}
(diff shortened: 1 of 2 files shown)


RalfJung:
Cc @CraftSpider 

> DuplicateHandle should create a new handle when called on a pseudo handle, instead of just returning it unchanged.

Is there some way this can actually be observed, so that we can have a test? Something showing that the old behavior was wrong.

src/shims/windows/handle.rs
```diff
@@ -55,9 +55,8 @@ impl Handle {
     const PSEUDO_DISCRIMINANT: u32 = 1;
     const THREAD_DISCRIMINANT: u32 = 2;
     const FILE_DISCRIMINANT: u32 = 3;
-    // Chosen to ensure Handle::Invalid encodes to -1. Update this value if there are ever more than
-    // 8 discriminants.
-    const INVALID_DISCRIMINANT: u32 = 7;
+    // Calculated to ensure Handle::Invalid encodes to -1.
+    const INVALID_DISCRIMINANT: u32 = 2u32.pow(Self::packed_disc_size()) - 1;
```
RalfJung: The problem with this variant is that it now becomes harder to tell whether the discriminants are all distinct.

tests/fail-dep/concurrency/windows_join_main.rs
```diff
@@ -12,7 +12,7 @@ use windows_sys::Win32::System::Threading::{INFINITE, WaitForSingleObject};
 
 // XXX HACK: This is how miri represents the handle for thread 0.
 // This value can be "legitimately" obtained by using `GetCurrentThread` with `DuplicateHandle`
-// but miri does not implement `DuplicateHandle` yet. (FIXME: it does now.)
+// but miri does not implement `DuplicateHandle` for thread handles yet.
```
RalfJung:
```suggestion
// but miri does not implement `DuplicateHandle` for pseudo handles yet.
// We also can't move the current thread handle across thread boundaries since it is a
// pseudo handle that always refers to the "current" thread.
```

CraftSpider: That last bit is how this would be testable, if it was implemented - `DuplicateHandle` on a pseudo-handle returns a new 'real' handle. So if you duplicate `CurrentThread`, you instead get a new non-pseudo handle that you can pass across a thread boundary.

CraftSpider:
Thanks for catching this, I missed this bit when implementing the function

RalfJung:
So for this PR I'd say
- Remove the 2nd commit. We already have a test ensuring the encoding is correct, and we need a good way to ensure that the discriminants are distinct.
- Add a test that passes the result of `GetCurrentThread` to `DuplicateHandle`.

beepster4096:
just going to @rustbot author this till I get back to it

RalfJung:
Given that this fixes a case of Miri doing the wrong thing, I'll take the first commit and add the test to get this landed soon. I hope that's okay. :)

Issue #4746, opened 2025-12-06
Support "readdir" on Linux (and all Unixes)


RalfJung:
We currently support `readdir64` on Linux but `readdir` only on other Unixes where we need it for the standard library. We should support "readdir" everywhere ("readdir64" is Linux-specific, so that remains only there).

This is similar to https://github.com/rust-lang/miri/issues/4684 so the PR will likely look a lot like https://github.com/rust-lang/miri/pull/4714:
- add a new case in `unix/foreign_item.rs` and remove the now-redundant cases in the other files
- add a test directly invoking `readdir` to ensure it actually works

Islam-Imad:
I want to work on this Issue :)

enthropy7:
@Islam-Imad hi, still working on this? i could finish it if you want

Islam-Imad:
> [@Islam-Imad](https://github.com/Islam-Imad) hi, still working on this? i could finish it if you want

I was working on some issues in llvm project. it takes my time :) , i will dedicate some time to this issue. i refer back to you for any updates.
I know that i got assigned for long time. but i will try it :) (also have some running exam)

Islam-Imad:
Sorry For that :)
i will start working on it and refer back for any questions :)

Islam-Imad:
unsafe {
        let dirp = libc::opendir(c_path.as_ptr()); // DIR *opendir(const char *name);
        assert!(!dirp.is_null());
        let mut entries = Vec::new();
        loop {
            let entry_ptr = libc::readdir(dirp);
            if entry_ptr.is_null() {
                break;
            }
            let name_ptr = std::ptr::addr_of!((*entry_ptr).d_name) as *const libc::c_char;
            let name = CStr::from_ptr(name_ptr);
            let name_str = name.to_string_lossy();
            if name_str != "." && name_str != ".." {
                entries.push(name_str.into_owned());
            }
        }
        assert_eq!(libc::closedir(dirp), 0);
        entries.sort();
        assert_eq!(entries, vec!["file1.txt", "file2.txt"]);
    }
this code always always fails on **FreeBSD** os.

Islam-Imad:
My assumption is that libc::readdir is portable but the running of Mac os fails sense it does't find a matching function to call.
FreeBSD read files but with empty names and the remaining OSs are fine.
i think that i have to add Mac os in readdir64 in matching expression.
@RalfJung 
should i open a draft to see the changes ?

RalfJung:
[...]

PR #4821, opened 2026-01-20
Unify  readdir across Unix targets and add test

Islam-Imad:
Closes rust-lang/miri#4746

### DIFF
--- a/tests/pass-dep/libc/libc-fs.rs
+++ b/tests/pass-dep/libc/libc-fs.rs
@@ -614,7 +614,6 @@ fn test_opendir_closedir() {
 fn test_readdir() {
     use std::fs::{create_dir, remove_dir, write};
 
-    use libc::opendir;
     let dir_path = utils::prepare_dir("miri_test_libc_readdir");
     create_dir(&dir_path).ok();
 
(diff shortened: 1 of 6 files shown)


RalfJung:
The freebsd issue seems to be a problem similar to last time with `fstat`, where it ends up picking the wrong type from libc because there are multiple libc crates. So the fix is the same, I made it use the type from the signature -- that means we trust the type the user gave us, but well, that's better than broken code.

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -579,3 +580,43 @@ fn test_ioctl() {
...
+        let mut entries = Vec::new();
+        loop {
+            let entry_ptr = libc::readdir(dirp);
+            if entry_ptr.is_null() {
+                break;
+            }
+            let name_ptr = std::ptr::addr_of!((*entry_ptr).d_name) as *const libc::c_char;
+            let name = CStr::from_ptr(name_ptr);
+            let name_str = name.to_string_lossy();
+            if name_str != "." && name_str != ".." {
```
RalfJung: Why did you skip the "." and ".." files here?

Those files should actually show up! This is a bug in Miri, I will file an issue.

RalfJung:
Thanks a lot! I fixed the typing issues as those are quite tricky, but having a test is super helpful.

tests/pass-dep/libc/libc-fs.rs
Islam-Imad: i used to write this condition while coding in `C` :) , i will give this bug some time (not nowadays) and refer back in zulip for any updates.

RalfJung: Thanks for the offer! However, I'll probably try to fix this bug tomorrow -- a bug where Miri does the wrong is something I'd like to have fixed ASAP.

PR #4824, opened 2026-01-22
Add test for libc opendir and closedir

Islam-Imad:
Add test for libc opendir and closedir
- open existing dir
- open not existing dir
- open file as if it was dir


RalfJung:
LGTM, thanks :)

Issue #3934, opened 2024-10-02
Add support for Linux timerfd


FrankReh:
Does or will Miri have the notion of elapsed time, or simulated elapsed time? Is there already anything close to the clocking necessary to support `timerfd_create` and the associated `_settime` and `_gettime`?

https://man7.org/linux/man-pages/man2/timerfd_create.2.html

oli-obk:
Yes, we support simulated time in isolation mode and use the real system time outside of it. We probably can only support CLOCK_MONOTONIC unless CLOCK_REALTIME without the ability to set the time is ok. The other clockids seem odd enough to be rarely used, but i'll happily be convinced otherwise and then we can think about what it would take for them

FrankReh:
Thank you. I found some nice examples of features already using the `timespec`and the MONOTIC and REALTIME clocks.

Let me keep this open for a little while. I may have questions once I understand how the timeout for Futex works. There the thread blocked on a Futex can be awoken if the duration is reached (I'm assuming). Here, the timerfd would become ready to read, and could be associated with an epoll.

FrankReh:
When might level-triggered epoll be supported by Miri? Would it be easy enough for a new-comer to support with what's already in Miri or is it waiting for blocking support or something else that's not there yet? Asking for a friend ...

Actually, asking because it is how the async_io::Timer from the smol projects is built. That's how I got to look at timerfd_create and timerfd_settime (for a monotonic clock) in the first place.

tiif:
> When might level-triggered epoll be supported by Miri? 

If I were to do it, I am currently left with #3665. 

> Would it be easy enough for a new-comer to support with what's already in Miri or is it waiting for blocking support or something else that's not there yet?

It should be easier as we already have structure provided for edge-triggered epoll. But I imagine this will still be a big feature that involves quite a bit of change. There will also be a bit of git conflict involved as it is overlapping with #3665. I can probably help with review and answer some questions, but I am not sure how much time I will be able to commit. Any opinion from @oli-obk as the primary reviewer of epoll architecture? 

> Actually, asking because it is how the async_io::Timer from the smol projects is built. That's how I got to look at timerfd_create and timerfd_settime (for a monotonic clock) in the first place.

Do they hit unsupported_error for level-triggered epoll? I am interested in seeing how they use it so we can design the test similar to how they use it.

FrankReh:
[...]

PR #4828, opened 2026-01-23
call unblock callbacks only when the thread is scheduled again

joboet:
I'm trying to shim [waitable timer objects](https://learn.microsoft.com/en-us/windows/win32/sync/waitable-timer-objects) since I want to make use of them in `std` (see https://github.com/rust-lang/rust/pull/151553). My idea is to modify the central `schedule` function to not only look for timed out blocking operations but also go through all active timers and look for ones whose deadline has passed. This however necessitates waking up multiple threads from `schedule` – which currently would run their wakeup callbacks, violating the principle that `schedule` is not supposed to run things. I could try to work my way around this, but I think a more principled solution is the following:

This PR changes the scheduling of callback functions: Instead of being immediately invoked when an operation unblocks a thread, the thread is instead put into the newly created `ThreadState::Unblocked`. Threads with `ThreadState::Unblocked` are eligible for scheduling, but will, upon being selected, execute a callback instead of executing the next `step`.

This has the nice side-effect of exploring more possible executions of the macOS futex functions – these retrieve the current futex waiter count in their wakeup callback, so by making this callback depend on scheduling, there are more possible values that can be returned. If I read the XNU sources correctly, that also matches the system behaviour (the waiter count is [retrieved](https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/kern/sys_ulock.c#L818) in `ulock_wait_cleanup` which is called from the `ulock_wait_continue` continuation).

I have no clue about how this interacts with GenMC, that could potentially cause issues with the `BlockReason::GenMC` hack.

### DIFF
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -25,8 +25,8 @@ use crate::*;
 enum SchedulingAction {
     /// Execute step on the active thread.
     ExecuteStep,
-    /// Execute a timeout callback.
-    ExecuteTimeoutCallback,
+    /// Execute a callback on the active thread.
+    ExecuteCallback,
     /// Wait for a bit, until there is a timeout to be called.
     Sleep(Duration),
 }
@@ -120,6 +120,9 @@ pub enum BlockReason {
 enum ThreadState<'tcx> {
     /// The thread is enabled and can be executed.
     Enabled,
+    /// The thread is enabled, but there is an unblock callback that needs to
+    /// be executed first.
+    Unblocked { kind: UnblockKind, callback: DynUnblockCallback<'tcx> },
     /// The thread is blocked on something.
     Blocked { reason: BlockReason, timeout: Option<Timeout>, callback: DynUnblockCallback<'tcx> },
     /// The thread has terminated its execution. We do not delete terminated
@@ -131,6 +134,8 @@ impl<'tcx> std::fmt::Debug for ThreadState<'tcx> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             Self::Enabled => write!(f, "Enabled"),
+            Self::Unblocked { kind, .. } =>
+                f.debug_struct("Unblocked").field("kind", kind).finish(),
             Self::Blocked { reason, timeout, .. } =>
                 f.debug_struct("Blocked").field("reason", reason).field("timeout", timeout).finish(),
             Self::Terminated => write!(f, "Terminated"),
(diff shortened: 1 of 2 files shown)


src/concurrency/thread.rs
```diff
@@ -768,29 +766,31 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
...
-            && !thread_manager.yield_active_thread
-        {
-            // The currently active thread is still enabled, just continue with it.
-            return interp_ok(SchedulingAction::ExecuteStep);
-        }
-        // The active thread yielded or got terminated. Let's see if there are any timeouts to take
-        // care of. We do this *before* running any other thread, to ensure that timeouts "in the
-        // past" fire before any other thread can take an action. This ensures that for
-        // `pthread_cond_timedwait`, "an error is returned if [...] the absolute time specified by
-        // abstime has already been passed at the time of the call".
```
joboet, line 782: That has nothing to do with the timing of the callback – `pthread_cond_timedwait` is allowed to yield even if the time has passed already. The important thing is that the timeout checks happen before selecting the next thread, otherwise you could get a deadlock.

RalfJung:
The logic for concurrency primitives such as mutexes relies on having the callback invoked immediately (for instance, in [this assertion](https://github.com/rust-lang/miri/blob/7dd34cd5f7219fa0011ca0e923d1b194ef082745/src/concurrency/sync.rs#L493)). So delaying this until the next time the thread is scheduled is not an option I think. I very deliberately chose the current design when I last refactored this code.

RalfJung:
> I'm trying to shim [waitable timer objects](https://learn.microsoft.com/en-us/windows/win32/sync/waitable-timer-objects)

Those sound similar to Linux's timerfd? What I kind of expected would happen there is that when a thread blocks on the FD, we compute when it has to wake up again and block the thread accordingly. That would avoid having the scheduler  worry about any non-thread objects. Does that not work?

joboet:
> Those sound similar to Linux's timerfd? What I kind of expected would happen there is that when a thread blocks on the FD, we compute when it has to wake up again and block the thread accordingly. That would avoid having the scheduler worry about any non-thread objects. Does that not work?

They are quite similar. The problem is that another thread could change the timer time while threads are blocked, which would invalidate their timeout time – they'd need to recompute it somehow. Also, the current approach necessitates going over all threads twice before executing the timeout. For `n` threads blocked on a timer, that results in an $O(n^2)$ wakeup operation. I think I can make everything work for the simplest case and support `std`'s usecase thereby, but I wanted to try propose this more principled approach first.

RalfJung:
> They are quite similar. The problem is that another thread could change the timer time while threads are blocked, which would invalidate their timeout time

Ah, so the docs say that if the timer gets updated when threads are already sleeping, those threads might get woken up earlier or later or even immediately? I was trying to find that out but it wasn't clear to me. Is it the same for timerfd?

Yeah that is more challenging to implement. I'll have to think about this.

lukeb36eu (approved):

RalfJung:
How did you plan to deal with this timeout-changed-when-thread-is-already-blocked situation? I don't see how "delay callback until thread gets scheduled again" makes that part easier.

> violating the principle that schedule is not supposed to run things

It's not supposed to run any *code* (as in, MIR), but these unblock callbacks aren't code, they are just interpreter internals. The scheduler already runs unblock callbacks when timeouts occur and that's fine.

> the current approach necessitates going over all threads twice before executing the timeout. For n threads blocked on a timer, that results in an O ( n 2 ) wakeup operation

I don't know what you are referring to here.

RalfJung:
I'm going to close this for now, due to the design problems mentioned above. If you want to get back to this I would suggest discussing the problem you need to solve first, maybe there is a less invasive way to solve it.

It would also be easier for me if the first primitive of this sort that we implement was timerfd (https://github.com/rust-lang/miri/issues/3934) rather than something Windows-specific.

PR #4836, opened 2026-01-28
add a bug to the trophy shelf

joboet:
Does it count if the issue reporter didn't use miri, but miri helped to find the bug?

CC https://github.com/rust-lang/rust/issues/151758


RalfJung:
I'll take it :) Thanks!

Issue #4352, opened 2025-05-27
Ignore file creation flag in ``fcntl(F_SETFL)``


tiif:
For ``fcntl(F_SETFL)``, some file creation flags supposed to be ignored:

> 
>       F_SETFL (int)
>             Set the file status flags to the value specified by arg.
>              File access mode (O_RDONLY, O_WRONLY, O_RDWR) and file
>             creation flags (i.e., O_CREAT, O_EXCL, O_NOCTTY, O_TRUNC)
>              in arg are ignored.  On Linux, this operation can change
>             only the O_APPEND, O_ASYNC, O_DIRECT, O_NOATIME, and
>              O_NONBLOCK flags.  It is not possible to change the O_DSYNC
>              and O_SYNC flags; see BUGS, below.
> source: https://man7.org/linux/man-pages/man2/fcntl.2.html

There is a list of file creation flags in https://man7.org/linux/man-pages/man2/open.2.html. 

To resolve this issue, we just need to ignore every file creation flag, except ``O_APPEND``, ``O_ASYNC``, ``O_DIRECT``, ``O_NOATIME`` and ``O_NONBLOCK``.

hulxv:
Hey, I would like to work on this. Can you give me more details?

tiif:
Currently, if the flag we passed in contains flags that supposed to be ignored, for example ``O_CREAT``, it will hit this unsupported path https://github.com/rust-lang/miri/blob/b00ed877740876e8d2eeb4087cbd9df2449278ac/src/shims/unix/unnamed_socket.rs#L192-L196. (This is a good test btw, you can write some tests to make sure that it actually failed with unsupported before making any change)

I think the correct fix here is going through all the flags that supposed to be ignored (see the issue description above for more information), and ignore it before passing it to ``set_flags`` [here](https://github.com/rust-lang/miri/blob/b00ed877740876e8d2eeb4087cbd9df2449278ac/src/shims/unix/fd.rs#L190-L200). [Here is an example of how flags are ignored](https://github.com/rust-lang/miri/blob/b00ed877740876e8d2eeb4087cbd9df2449278ac/src/shims/unix/unnamed_socket.rs#L188-L189).

omgupta-iitk:
Hi @hulxv, I had previously begun looking into this issue and would like to continue working on it. Would you mind if I take this one? If you’ve already started, no worries, just let me know. Thanks!

hulxv:
[...]

PR #4840, opened 2026-02-01
FCNTL F_SETFL Ignore creation flags

MousseARaser06:
This pull request aims at fixing rust-lang/miri#4352, it proceeds by having a list of platform-allowed flags (bits) for that operation (F_SETFL), and performing a bitwise AND assignment to exclude all of the flags that aren't allowed.
Allowed flags change from platform to platform

### DIFF
--- a/src/shims/unix/unnamed_socket.rs
+++ b/src/shims/unix/unnamed_socket.rs
@@ -173,9 +173,6 @@ impl FileDescription for AnonSocket {
         // FIXME: File creation flags should be ignored.
 
         let o_nonblock = ecx.eval_libc_i32("O_NONBLOCK");
-        let o_rdonly = ecx.eval_libc_i32("O_RDONLY");
-        let o_wronly = ecx.eval_libc_i32("O_WRONLY");
-        let o_rdwr = ecx.eval_libc_i32("O_RDWR");
 
         // O_NONBLOCK flag can be set / unset by user.
         if flag & o_nonblock == o_nonblock {
@@ -184,9 +181,6 @@ impl FileDescription for AnonSocket {
         } else {
             self.is_nonblock.set(false);
         }
-
-        // Ignore all file access mode flags.
-        flag &= !(o_rdonly | o_wronly | o_rdwr);
 
         // Throw error if there is any unsupported flag.
         if flag != 0 {


RalfJung:
Thanks for the PR :)

We'll get to reviewing it eventually, but there are other PRs in the queue that will be reviewed first.

src/shims/unix/fd.rs
```diff
@@ -194,7 +194,32 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                            | this.eval_libc_i32("O_APPEND")
+                            | this.eval_libc_i32("O_DIRECT"),
+                    // Linux + Android match case
+                    _ =>
+                        this.eval_libc_i32("O_NONBLOCK")
+                            | this.eval_libc_i32("O_APPEND")
+                            | this.eval_libc_i32("O_DIRECT")
+                            | this.eval_libc_i32("O_NOATIME")
+                            | this.eval_libc_i32("O_ASYNC"),
+                };
```
RalfJung, line 220: For Linux, we have docs saying "this operation can change      only the O_APPEND, O_ASYNC, O_DIRECT, O_NOATIME, and      O_NONBLOCK flags". Do we have anything similar for other OSes? If not, it seems kind of risky to just ignore everything we don't know about. The safer approach is to only ignore what we know can definitely be ignored.

src/shims/unix/fd.rs
```diff
@@ -194,7 +194,32 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+                flag &= allowed_flags;
```
RalfJung, line 222: It seems like the ignoring logic in `AnonSocket::set_flags` can now be removed?

RalfJung (changes requested):
Thanks for the PR! I have a concern regarding the fundamental approach.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4840/7d5b27865a3855e23eddf913e28a9e986f19d98a..9291577efc272fbb6dc1c1ae28614c061349da17)*

src/shims/unix/fd.rs
MousseARaser06: I searched the other docs (the ones in that match statement)
and they all more or less have the same ignored flags in their man pages
the only OS I'm not quite sure is Solaris and Illumos, but then if I don't include them inside the match statement, the test fails
I will re-check everything just to make sure, but I'm 95% sure about this

RalfJung: Well, the test should match reality. :)

Overall I think the easier approach is to just only ignore flags we are sure are ignored everywhere.

src/shims/unix/fd.rs
the8472, line 222: This looks like it's only whitelisting flags to be set. But flags can be toggled, so you need to check the diff between the old flags and the new flags.
A common pattern is to load the old flags, toggle just one bit and write back the new value.

This makes it brittle under multithreading, but I guess that's not a concern for miri.

RalfJung: That's handled in the actual per-FD-type implementations, this shared code is just meant to filter irrelevant flags away so each FD type can properly error on unsupported flags.

src/shims/unix/fd.rs
MousseARaser06: MacOs's man pages states the following

The flags for the F_GETFL and F_SETFL commands are as follows:

           O_NONBLOCK   Non-blocking I/O; if no data is available to a read
                        call, or if a write operation would block, the read or
                        write call returns -1 with the error EAGAIN.

           O_APPEND     Force each write to append at the end of file; corre-sponds corresponds
                        sponds to the O_APPEND flag of [open(2)](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/open.2.html#//apple_ref/doc/man/2/open).

           O_ASYNC      Enable the SIGIO signal to be sent to the process
                        group when I/O is possible, e.g., upon availability of
                        data to be read.

FreeBSD's man pages states the following 

The flags for the F_GETFL and F_SETFL commands are as follows:

       O_NONBLOCK   Non-blocking I/O; if no data is  available	to  a  [read(2)](https://man.freebsd.org/cgi/man.cgi?query=read&sektion=2&apropos=0&manpath=FreeBSD+15.0-RELEASE+and+Ports)
		    system  call,  or if a [write(2)](https://man.freebsd.org/cgi/man.cgi?query=write&sektion=2&apropos=0&manpath=FreeBSD+15.0-RELEASE+and+Ports) operation would block, the
		    read or write call returns -1 with the error EAGAIN.

       O_APPEND	    Force each write to	append at the end of file; corresponds
		    to the O_APPEND flag of [open(2)](https://man.freebsd.org/cgi/man.cgi?query=open&sektion=2&apropos=0&manpath=FreeBSD+15.0-RELEASE+and+Ports).

       O_DIRECT	    Minimize or	eliminate the cache  effects  of  reading  and
		    writing.   The  system  will  attempt to avoid caching the
		    data you read or write.  If	it cannot  avoid  caching  the
		    data,  it  will  minimize  the  impact the data has	on the
		    cache.  Use	of this	flag can  drastically  reduce  perfor-
		    mance if not used with care.

       O_ASYNC	    Enable  the	 SIGIO	signal to be sent to the process group
		    when I/O is	possible, e.g.,	upon availability of  data  to
		    be read.

       O_SYNC	    Enable synchronous writes.	Corresponds to the O_SYNC flag
		    of [open(2)](https://man.freebsd.org/cgi/man.cgi?query=open&sektion=2&apropos=0&manpath=FreeBSD+15.0-RELEASE+and+Ports).	 O_FSYNC is an historical synonym for O_SYNC.

       O_DSYNC	    Enable   synchronous  data	writes.	  Corresponds  to  the
		    O_DSYNC flag of [open(2)](https://man.freebsd.org/cgi/man.cgi?query=open&sektion=2&apropos=0&manpath=FreeBSD+15.0-RELEASE+and+Ports).

Illumos/Solaris's docs are very unclear, therefore only the most "common" flags are allowed (O_NONBLOCK, O_APPEND, O_DIRECT)

And Linux's man pages states the following

F_SETFL
              Set the file status flags to the value specified by arg.
              File access mode (O_RDONLY, O_WRONLY, O_RDWR) and file
              creation flags (i.e., O_CREAT, O_EXCL, O_NOCTTY, O_TRUNC)
              in arg are ignored.  On Linux, this operation can change
              only the O_APPEND, O_ASYNC, O_DIRECT, O_NOATIME, and
              O_NONBLOCK flags.  It is not possible to change the O_DSYNC
              and O_SYNC flags; see BUGS, below.

RalfJung: Okay. So sounds like ignoring O_RDONLY, O_WRONLY, O_RDWR, O_CREAT, O_EXCL, O_NOCTTY, O_TRUNC will work everywhere while also avoiding the risk of ignoring the wrong thing on new platforms / when new flags are added.

MousseARaser06: That sounds about right

RalfJung: Okay, so please adjust the PR accordingly then.

@rustbot author

MousseARaser06: That's what's the code is doing already though ?

RalfJung: The code has an `allowed_flags`, what I described is an `ignored_flags`.

MousseARaser06: I'm not sure how to do that, and I don't want to change it, closing PR

RalfJung: What I meant is fairly simple; sorry if I was not clear enough. In pseudo-code:
```rust
let ignored_flags = O_RDONLY | O_WRONLY | O_RDWR | O_CREAT | O_EXCL | O_NOCTTY | O_TRUNC;

let filtered_flags = flags & ~ignored_flags;
```
But, it's okay, me or someone else can take over the PR to finish this. Thanks for giving it a go!

Issue #3976, opened 2024-10-17
Poor error in `./miri run --dep` when using a wrong `-Zmiri` flag


RalfJung:
```
$ ./miri run tests/fail-dep/libc/socketpair-data-race.rs  -Zmiri-foo --dep
The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value: Errored { command: "\"/home/r/src/rust/miri/target/debug/miri\" \"--edition\" \"2021\" \"--sysroot\" \"/home/r/.cache/miri\" \"tests/fail-dep/libc/socketpair-data-race.rs\" \"-Zmiri-foo\" \"--out-dir\" \"/home/r/src/rust/miri/target/miri_ui\" \"--print=cfg\" \"--target\" \"x86_64-unknown-linux-gnu\"", errors: [], stderr: [101, 114, 114, 111, 114, 58, 32, 117, 110, 107, 110, 111, 119, 110, 32, 117, 110, 115, 116, 97, 98, 108, 101, 32, 111, 112, 116, 105, 111, 110, 58, 32, 96, 109, 105, 114, 105, 45, 102, 111, 111, 96, 10, 10], stdout: [] }
Location: tests/ui.rs:340
```
Showing stderr as its raw ASCII encoding isn't exactly readable. ;)

Cc @oli-obk

RalfJung:
This is caused by that unwrap here:

https://github.com/rust-lang/miri/blob/66366810988184e00ef17ac1f65b0106f4f96a36/tests/ui.rs#L340

However, as far as I can see there is no nice way to report a `test_result::Errored` -- the fields are private and the `Debug` impl produces the ugly output above. @oli-obk how is this intended to work? I can't quite figure it out...

oli-obk:
I just need to fix the debug impl.

RalfJung:
Ideally it can use the already existing reporting logic, would be unfortunate to have the same error reported in two different ways.

RalfJung:
@oli-obk this is still a problem; seems like at this point all we need is a ui_test bump?

oli-obk:
Well we had https://github.com/rust-lang/miri/pull/3948 but it had review requests 😛 and then I forgot about it

RalfJung:
Would be better to split the bump from the new feature. :) Which I think you have done now, thanks!

RalfJung:
This hasn't really gotten much better with the bump:
```
The application panicked (crashed).
[...]

PR #4842, opened 2026-02-02
`./miri run --dep`: do not run Miri during the dependency build

RalfJung:
Fixes https://github.com/rust-lang/miri/issues/3976

Add a `cargo miri build` variant specifically for the dependency build so we avoid spinning up an interpreter. When the interpreter itself is being debugged, having this extra interpreter during the build can be super confusing.


RalfJung:
Looks like we do need those cfgs after all... unfortunate.

Issue #4507, opened 2025-07-29
Only build capstone for the architectures we need


RalfJung:
The git version of capstone supports feature flags to control which architectures are supported. We should use that to limit it to x86 (and maybe eventually aarch64), hopefully saving some build time by skipping all the other architectures.

PR #4843, opened 2026-02-02
chore: limit capstone to x86 and arm

hsqStephenZhang:
Fix https://github.com/rust-lang/miri/issues/4507

### DIFF
--- a/src/shims/native_lib/trace/parent.rs
+++ b/src/shims/native_lib/trace/parent.rs
@@ -395,8 +395,6 @@ fn capstone_find_events(
                 _ => (),
             }
         }
-        // FIXME: arm64
-        _ => unimplemented!(),
     }
 
     false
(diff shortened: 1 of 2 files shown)


hsqStephenZhang:
## before

image


## after
image

and `cargo tree -e features` confirmed the result.

RalfJung:
Thanks for the PR!
Currently we don't even support ARM here so we can limit this to just x86.

However, you'll have to enable the `full` feature of capstone so that we can figure out the accesses of an instruction.

hsqStephenZhang:
Done. The newly compile binary is still much smaller than the original one.

image

src/shims/native_lib/trace/parent.rs
```diff
@@ -396,6 +396,7 @@ fn capstone_find_events(
             }
         }
         // FIXME: arm64
+        #[allow(unused)]
```
RalfJung: Why is this needed now?

If the enum now has only that one variant, please just entirely remove this arm.

RalfJung:

@rustbot author

Issue #3172, opened 2023-11-17
Add support for "llvm.aarch64.crypto.aese"


alex:
Currently this produces:

```
   --> /Users/alex_gaynor/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/../../stdarch/crates/core_arch/src/arm_shared/crypto.rs:69:5
    |
69  |     vaeseq_u8_(data, key)
    |     ^^^^^^^^^^^^^^^^^^^^^ can't call foreign function `llvm.aarch64.crypto.aese` on OS `macos`
    |
```

saethlin:
I should probably do a version of https://github.com/rust-lang/miri/issues/2057 for aarch64, all my current surveying is done based on x86-64-v2.

alex:
If there's a straightforward script for it (and you promise it won't destroy my computer :D), I'm happy to do a run on my ARM64 laptop.

saethlin:
It involves running the tests for every published crate so I feel like you're not up for that :)

alex:
I was thinking maybe I'd just do the first 500 or 1k or something :-)

But if your setup already works for it, that sounds good!

saethlin:
I hacked up https://github.com/saethlin/crater-at-home a bit to set the target to aarch64-unknown-linux-gnu and here's a thousand crates (hosted for now in my dev bucket): https://miri-bot-dev.s3.amazonaws.com/aarch64-1000.tar.xz

Missing LLVM intrinsics look like:
```
716 counts
(  1)      178 (24.9%, 24.9%): llvm.aarch64.neon.uminv.i32.v4i32
(  2)       62 ( 8.7%, 33.5%): llvm.aarch64.neon.uminv.i8.v16i8
(  3)       60 ( 8.4%, 41.9%): llvm.aarch64.neon.uminv.i16.v8i16
(  4)       50 ( 7.0%, 48.9%): llvm.aarch64.neon.tbl1.v16i8
(  5)       40 ( 5.6%, 54.5%): llvm.aarch64.neon.umaxp.v16i8
(  6)       26 ( 3.6%, 58.1%): llvm.aarch64.neon.ushl.v2i64
(  7)       24 ( 3.4%, 61.5%): llvm.aarch64.neon.ushl.v4i32
(  8)       18 ( 2.5%, 64.0%): llvm.aarch64.neon.uaddv.i32.v4i32
[...]

PR #4847, opened 2026-02-04
Feat: aarch64 intrinsics to make MIRI work on chacha20

hsqStephenZhang:
## Implement more intrinsics on aarch64

this PR intends to make miri work on [chacha20](https://github.com/RustCrypto/stream-ciphers/tree/master/chacha20). Partially fix this [issue](https://github.com/rust-lang/miri/pull/3172)

Newly implemented intrinsics include:
1. neon.uminv.i8.v16i8
2. neon.uminv.i8.v1616
3. neon.uminv.i8.v16i32
4. neon.tbl1.v16i8

## Test

1. unittest
2. run `MIRI=/path/to/miri cargo miri test --tests` in https://github.com/RustCrypto/stream-ciphers/tree/master/chacha20 and all 4 tests passed. (It took about half an hour on my m2 MacOS)

## Discussions

Though this [issue](https://github.com/rust-lang/miri/pull/3172) mentioned that more work should be put on stdarch, I think it would not be bad to support some common intrinsics on neon to make the crypto libs work. Thoughts?

### DIFF
--- a/src/shims/aarch64.rs
+++ b/src/shims/aarch64.rs
@@ -1,4 +1,4 @@
-use rustc_abi::{CanonAbi, Size};
+use rustc_abi::CanonAbi;
 use rustc_middle::mir::BinOp;
 use rustc_middle::ty::Ty;
 use rustc_span::Symbol;
@@ -60,8 +60,8 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             }
             // Vector table lookup: each index selects a byte from the 16-byte table, out-of-range -> 0.
             // Used to implement vtbl1_u8 function.
-            // LLVM does not have a portable shuffle that takes non-const indices.
-            // So we need to implement this ourselves
+            // LLVM does not have a portable shuffle that takes non-const indices
+            // so we need to implement this ourselves.
             // https://developer.arm.com/architectures/instruction-sets/intrinsics/vtbl1_u8
             "neon.tbl1.v16i8" => {
                 let [table, indices] =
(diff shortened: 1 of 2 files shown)


src/shims/aarch64.rs
```diff
@@ -58,9 +58,76 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    ecx: &mut crate::MiriInterpCx<'tcx>,
+    abi: &FnAbi<'tcx, Ty<'tcx>>,
+    link_name: Symbol,
+    args: &[OpTy<'tcx>],
+    dest: &MPlaceTy<'tcx>,
+    size: u64,
+) -> InterpResult<'tcx, ()> {
+    let [op] = ecx.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
+    let (op, op_len) = ecx.project_to_simd(op)?;
+    assert_eq!(op_len, size);
```
RalfJung: There's no reason to even pass the size as argument if it's already given by `op_len`.

hsqStephenZhang: fixed

hsqStephenZhang:
If i understand it correct, stdarch already used [`simd_reduce_min`](https://doc.rust-lang.org/src/core/stdarch/crates/core_arch/src/aarch64/neon/generated.rs.html#13827-13829) and [`simd_shuffle`](https://doc.rust-lang.org/src/core/stdarch/crates/core_arch/src/arm_shared/neon/generated.rs.html#7679-7681) for the intrinsics implemented here. But they are still lowered to llvm intrinsics?

So to support them, we have to implement in MIRI.

RalfJung:
> If i understand it correct, stdarch already used [simd_reduce_min](https://doc.rust-lang.org/src/core/stdarch/crates/core_arch/src/aarch64/neon/generated.rs.html#13827-13829) and [simd_shuffle](https://doc.rust-lang.org/src/core/stdarch/crates/core_arch/src/arm_shared/neon/generated.rs.html#7679-7681) for the intrinsics implemented here. But they are still lowered to llvm intrinsics?

Both of these cannot be true at the same time. If it uses simd_*, then no Miri patch is needed.

RalfJung:
Indeed the `uminv_reductions` test already passes on Miri without your changes.

The table lookup intrinsics [do not use](https://github.com/rust-lang/rust/blob/9fddd28d38eda100b07b0b11fa578da64c09da6b/library/stdarch/crates/core_arch/src/aarch64/neon/generated.rs#L19618-L19626) simd_* operations. Any idea why?
Cc @folkertdev

folkertdev:
LLVM does not have a portable shuffle that takes non-const indices. For `simd_shuffle` the indices must be a `const` vector.

RalfJung:
Okay, I guess we'll need that one then. But please remove the `uminv` ones, @hsqStephenZhang.

hsqStephenZhang:
Thanks for the context.

yep, only vqtbl1q needs to be supported to make chacha20 work.

RalfJung:
Thanks! I pushed some minor tweaks and squashed the commits.

PR #4851, opened 2026-02-05
Simplify wildcard datastructure.

royAmmerschuber:
Makes it so that each node only needs to count all local exposed nodes instead of also keeping track of foreign nodes. This makes the update logic significantly simpler.

### DIFF
--- a/src/borrow_tracker/tree_borrows/tree.rs
+++ b/src/borrow_tracker/tree_borrows/tree.rs
@@ -26,7 +26,7 @@ use super::foreign_access_skipping::IdempotentForeignAccess;
 use super::perms::{PermTransition, Permission};
 use super::tree_visitor::{ChildrenVisitMode, ContinueTraversal, NodeAppArgs, TreeVisitor};
 use super::unimap::{UniIndex, UniKeyMap, UniValMap};
-use super::wildcard::WildcardState;
+use super::wildcard::ExposedCache;
 use crate::borrow_tracker::{AccessKind, GlobalState, ProtectorKind};
 use crate::*;
 
@@ -89,7 +89,7 @@ impl LocationState {
         &mut self,
         idx: UniIndex,
         nodes: &mut UniValMap<Node>,
-        wildcard_accesses: &mut UniValMap<WildcardState>,
+        exposed_cache: &mut ExposedCache,
         access_kind: AccessKind,
         relatedness: AccessRelatedness,
         protected: bool,
@@ -99,7 +99,7 @@ impl LocationState {
         // ensures it is only called when `skip_if_known_noop` returns
         // `Recurse`, due to the contract of `traverse_this_parents_children_other`.
         self.record_new_access(access_kind, relatedness);
-
+        let old_access_level = self.permission.strongest_allowed_local_access(protected);
         let transition = self.perform_access(access_kind, relatedness, protected)?;
         if !transition.is_noop() {
             let node = nodes.get_mut(idx).unwrap();
@@ -111,8 +111,8 @@ impl LocationState {
             // We need to update the wildcard state, if the permission
             // of an exposed pointer changes.
             if node.is_exposed {
-                let access_type = self.permission.strongest_allowed_local_access(protected);
-                WildcardState::update_exposure(idx, access_type, nodes, wildcard_accesses);
+                let access_level = self.permission.strongest_allowed_local_access(protected);
+                exposed_cache.update_exposure(nodes, idx, old_access_level, access_level);
             }
         }
         Ok(())
@@ -261,14 +261,8 @@ pub struct LocationTree {
     ///
     /// We do uphold the fact that `keys(perms)` is a subset of `keys(nodes)`
     pub perms: UniValMap<LocationState>,
-    /// Maps a tag and a location to its wildcard access tracking information,
-    /// with possible lazy initialization.
-    ///
-    /// If this allocation doesn't have any exposed nodes, then this map doesn't get
-    /// initialized. This way we only need to allocate the map if we need it.
-    ///
-    /// NOTE: same guarantees on entry initialization as for `perms`.
-    pub wildcard_accesses: UniValMap<WildcardState>,
+    /// Caches information about the relatedness of nodes for a wildcard access.
+    pub exposed_cache: ExposedCache,
 }
 /// Tree structure with both parents and children since we want to be
 /// able to traverse the tree efficiently in both directions.
@@ -276,7 +270,7 @@ pub struct LocationTree {
 pub struct Tree {
     /// Mapping from tags to keys. The key obtained can then be used in
     /// any of the `UniValMap` relative to this allocation, i.e.
-    /// `nodes`, `LocationTree::perms` and `LocationTree::wildcard_accesses`
+    /// `nodes`, `LocationTree::perms` and `LocationTree::exposed_cache`
     /// of the same `Tree`.
     /// The parent-child relationship in `Node` is encoded in terms of these same
     /// keys, so traversing the entire tree needs exactly one access to
@@ -372,8 +366,8 @@ impl Tree {
                     IdempotentForeignAccess::None,
                 ),
             );
-            let wildcard_accesses = UniValMap::default();
-            DedupRangeMap::new(size, LocationTree { perms, wildcard_accesses })
+            let exposed_cache = ExposedCache::default();
+            DedupRangeMap::new(size, LocationTree { perms, exposed_cache })
         };
         Self { roots: SmallVec::from_slice(&[root_idx]), nodes, locations, tag_mapping }
     }
@@ -451,19 +445,9 @@ impl<'tcx> Tree {
             }
         }
 
-        // We need to ensure the consistency of the wildcard access tracking data structure.
-        // For this, we insert the correct entry for this tag based on its parent, if it exists.
-        // If we are inserting a new wildcard root (with Wildcard as parent_prov) then we insert
-        // the special wildcard root initial state instead.
-        for (_range, loc) in self.locations.iter_mut_all() {
-            if let Some(parent_idx) = parent_idx {
-                if let Some(parent_access) = loc.wildcard_accesses.get(parent_idx) {
-                    loc.wildcard_accesses.insert(idx, parent_access.for_new_child());
-                }
-            } else {
-                loc.wildcard_accesses.insert(idx, WildcardState::for_wildcard_root());
-            }
-        }
+        // We don't have to update `exposed_cache` as the new node is not exposed and
+        // has no children so the default counts of 0 are correct.
+
         // If the parent is a wildcard pointer, then it doesn't track SIFA and doesn't need to be updated.
         if let Some(parent_idx) = parent_idx {
             // Inserting the new perms might have broken the SIFA invariant (see
@@ -807,7 +791,7 @@ impl Tree {
         let node = self.nodes.remove(this).unwrap();
         for (_range, loc) in self.locations.iter_mut_all() {
             loc.perms.remove(this);
-            loc.wildcard_accesses.remove(this);
+            loc.exposed_cache.remove(this);
         }
         self.tag_mapping.remove(&node.tag);
     }
@@ -943,7 +927,7 @@ impl<'tcx> LocationTree {
         };
 
         let accessed_root_tag = accessed_root.map(|idx| nodes.get(idx).unwrap().tag);
-        for root in roots {
+        for (i, root) in roots.enumerate() {
             let tag = nodes.get(root).unwrap().tag;
             // On a protector release access we have to skip the children of the accessed tag.
             // However, if the tag has exposed children then some of the wildcard subtrees could
@@ -981,6 +965,7 @@ impl<'tcx> LocationTree {
                 access_kind,
                 global,
                 diagnostics,
+                /*is_wildcard_tree*/ i != 0,
             )?;
         }
         interp_ok(())
@@ -1029,7 +1014,7 @@ impl<'tcx> LocationTree {
                 .perform_transition(
                     args.idx,
                     args.nodes,
-                    &mut args.data.wildcard_accesses,
+                    &mut args.data.exposed_cache,
                     access_kind,
                     args.rel_pos,
                     protected,
@@ -1074,12 +1059,18 @@ impl<'tcx> LocationTree {
         access_kind: AccessKind,
         global: &GlobalState,
         diagnostics: &DiagnosticInfo,
+        is_wildcard_tree: bool,
     ) -> InterpResult<'tcx> {
         let get_relatedness = |idx: UniIndex, node: &Node, loc: &LocationTree| {
-            let wildcard_state = loc.wildcard_accesses.get(idx).cloned().unwrap_or_default();
             // If the tag is larger than `max_local_tag` then the access can only be foreign.
             let only_foreign = max_local_tag.is_some_and(|max_local_tag| max_local_tag < node.tag);
-            wildcard_state.access_relatedness(access_kind, only_foreign)
+            loc.exposed_cache.access_relatedness(
+                root,
+                idx,
+                access_kind,
+                is_wildcard_tree,
+                only_foreign,
+            )
         };
 
         // Whether there is an exposed node in this tree that allows this access.
@@ -1156,7 +1147,7 @@ impl<'tcx> LocationTree {
                 perm.perform_transition(
                     args.idx,
                     args.nodes,
-                    &mut args.data.wildcard_accesses,
+                    &mut args.data.exposed_cache,
                     access_kind,
                     relatedness,
                     protected,
@@ -1175,19 +1166,11 @@ impl<'tcx> LocationTree {
                 })
             },
         )?;
-        // If there is no exposed node in this tree that allows this access, then the
-        // access *must* be foreign. So we check if the root of this tree would allow this
-        // as a foreign access, and if not, then we can error.
-        // In practice, all wildcard trees accept foreign accesses, but the main tree does
-        // not, so this catches UB when none of the nodes in the main tree allows this access.
-        if !has_valid_exposed
-            && self
-                .wildcard_accesses
-                .get(root)
-                .unwrap()
-                .access_relatedness(access_kind, /* only_foreign */ true)
-                .is_none()
-        {
+        // If there is no exposed node in this tree that allows this access, then the access *must*
+        // be foreign to the entire subtree. Foreign accesses are only possible on wildcard subtrees
+        // as there are no ancestors to the main root. So if we do not find a valid exposed node in
+        // the main tree then this access is UB.
+        if !has_valid_exposed && !is_wildcard_tree {
             return Err(no_valid_exposed_references_error(diagnostics)).into();
         }
         interp_ok(())
--- a/src/borrow_tracker/tree_borrows/wildcard.rs
+++ b/src/borrow_tracker/tree_borrows/wildcard.rs
@@ -1,4 +1,3 @@
-use std::cmp::max;
 use std::fmt::Debug;
 
 use super::Tree;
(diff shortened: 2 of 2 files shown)


src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -261,14 +261,8 @@ pub struct LocationTree {
...
     pub perms: UniValMap<LocationState>,
-    /// Maps a tag and a location to its wildcard access tracking information,
-    /// with possible lazy initialization.
-    ///
-    /// If this allocation doesn't have any exposed nodes, then this map doesn't get
-    /// initialized. This way we only need to allocate the map if we need it.
-    ///
-    /// NOTE: same guarantees on entry initialization as for `perms`.
-    pub wildcard_accesses: UniValMap<WildcardState>,
+    /// Stores how nodes are related to exposed nodes,
```
```suggestion
    /// Stores how nodes are related to exposed nodes.
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1180,14 +1186,7 @@ impl<'tcx> LocationTree {
         // as a foreign access, and if not, then we can error.
         // In practice, all wildcard trees accept foreign accesses, but the main tree does
         // not, so this catches UB when none of the nodes in the main tree allows this access.
```
RalfJung: This comment is outdated now. Why did you change this to instead pass around a boolean?

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -1074,12 +1056,36 @@ impl<'tcx> LocationTree {
...
+            if max_local_tag.is_some_and(|max_local_tag| max_local_tag < node.tag) {
+                match relatedness {
+                    Some(LocalAccess) => None,
+                    Some(ForeignAccess) => Some(ForeignAccess),
+                    Some(EitherAccess) => Some(ForeignAccess),
+                    None => None,
+                }
+            } else {
+                relatedness
+            }
```
RalfJung: This seems to move some of the logic of determining the relatedness here into `tree.rs`. Seems better to have that in a helper in `wildcard.rs` like before.

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -50,403 +49,139 @@ impl WildcardAccessRelatedness {
...
-pub struct WildcardState {
-    /// How many of this node's direct children have `max_local_access()==Write`.
-    child_writes: u16,
-    /// How many of this node's direct children have `max_local_access()>=Read`.
-    child_reads: u16,
-    /// The maximum access level that could happen from an exposed node
-    /// that is foreign to this node.
+/// Only gets allocated if we add an exposed node.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct WildcardData(UniValMap<WildcardState>);
```
RalfJung: Please generally define types before using them. `WildcardState` as at the end of the file so currently one has to jump around a lot to understand this code.

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -50,403 +49,139 @@ impl WildcardAccessRelatedness {
...
+/// State per location per node keeping track of where relative to this
+/// node exposed nodes are and what access permissions they have.
+#[derive(Clone, Default, Debug, PartialEq, Eq)]
+pub struct WildcardState {
+    /// How many local nodes are exposed with write permissions.
+    local_writes: u16,
+    /// How many local nodes are exposed with read permissions.
+    local_reads: u16,
+}
 impl Tree {
```
RalfJung, line 191:
```suggestion

impl Tree {
```

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -50,403 +49,139 @@ impl WildcardAccessRelatedness {
...
-/// direct children's max_local_access/max_foreign_access.
-#[derive(Clone, Default, PartialEq, Eq)]
-pub struct WildcardState {
-    /// How many of this node's direct children have `max_local_access()==Write`.
-    child_writes: u16,
-    /// How many of this node's direct children have `max_local_access()>=Read`.
-    child_reads: u16,
-    /// The maximum access level that could happen from an exposed node
-    /// that is foreign to this node.
+/// Only gets allocated if we add an exposed node.
```
RalfJung: What does it mean for this to be "not allocated"? There's no `Option` here so it's not clear how the "absent" state is represented.

I think what you actually mean is that if the state is absent, then it is the default state, i.e. all counts are 0?

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -50,403 +49,139 @@ impl WildcardAccessRelatedness {
...
 }
 
+/// State per location per node keeping track of where relative to this
+/// node exposed nodes are and what access permissions they have.
+#[derive(Clone, Default, Debug, PartialEq, Eq)]
+pub struct WildcardState {
+    /// How many local nodes are exposed with write permissions.
+    local_writes: u16,
+    /// How many local nodes are exposed with read permissions.
+    local_reads: u16,
```
RalfJung: These numbers do *not* consider nodes in other trees, right? Some of those might be local, so that's not clear here.

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -50,403 +49,139 @@ impl WildcardAccessRelatedness {
...
     ) -> Option<WildcardAccessRelatedness> {
-        let rel = match kind {
-            AccessKind::Read => self.read_access_relatedness(),
-            AccessKind::Write => self.write_access_relatedness(),
+        // All nodes in the tree are local to the root, so we can use the root to get the total
+        // number of valid exposed nodes in the tree.
+        let root = self.0.get(root).cloned().unwrap_or_default();
+        let node = self.0.get(id).cloned().unwrap_or_default();
+
+        let (root, node) = match kind {
```
RalfJung: `root` isn't a great name for this variable, maybe call it `total_num` (and `local_num`)?

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -50,403 +49,139 @@ impl WildcardAccessRelatedness {
...
-
-        // Stack of nodes for which the max_foreign_access field needs to be updated.
-        // Will be filled with the children of this node and its parents children before
-        // we begin downwards traversal.
-        let mut stack: Vec<(UniIndex, WildcardAccessLevel)> = Vec::new();
-
-        // Add the direct children of this node to the stack.
-        {
+        let mut next_id = Some(id);
+        while let Some(id) = next_id {
```
RalfJung, line 164: This seems to be some sort of tree traversal; please explain that in a comment.

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -50,403 +49,139 @@ impl WildcardAccessRelatedness {
...
-                }
-                if old_parent_local_access == new_parent_local_access {
-                    // We didn't change `max_local_access()` for parent, so we don't need to propagate further upwards.
-                    break;
-                }
-
-                old_child_access = old_parent_local_access;
-                new_child_access = new_parent_local_access;
-                child = parent_id;
+            let mut wc = self.0.entry(id);
```
RalfJung: What's `wc`?

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -50,403 +49,139 @@ impl WildcardAccessRelatedness {
...
-                access_type,
-                &self.nodes,
-                &mut loc.wildcard_accesses,
-            );
+        if !node.is_exposed {
+            node.is_exposed = true;
+            let node = self.nodes.get(id).unwrap();
+
+            // When the first tag gets exposed then we initialize the
+            // wildcard state for every node and location in the tree.
```
RalfJung: This doesn't seem to be correct? We're only initializing it from the exposed node upwards.

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -457,10 +192,24 @@ impl Tree {
...
+        if node.is_exposed {
+            for (_, loc) in self.locations.iter_mut_all() {
+                let perm = loc
+                    .perms
+                    .get(idx)
+                    .map(|p| p.permission())
+                    .unwrap_or_else(|| node.default_location_state().permission());
+
+                let old_access_type = perm.strongest_allowed_local_access(true);
+                let access_type = perm.strongest_allowed_local_access(false);
```
RalfJung: Please explain those boolean arguments.

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -495,72 +239,25 @@ impl Tree {
...
-                );
-                let child_reads: usize = state.child_reads.into();
+                let (child_reads, child_writes) = node
+                    .children
+                    .iter()
+                    .copied()
+                    .map(|id| wildcard_accesses.0.get(id).cloned().unwrap_or_default())
+                    .fold((0, 0), |acc, wc| (acc.0 + wc.local_reads, acc.1 + wc.local_writes));
+                let expected_reads =
+                    child_reads + Into::<u16>::into(exposed_as >= WildcardAccessLevel::Read);
```
RalfJung:
```suggestion
                    child_reads + u16::from(exposed_as >= WildcardAccessLevel::Read);
```
Same below

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -451,19 +445,6 @@ impl<'tcx> Tree {
             }
         }
```
RalfJung, line 447:
```suggestion

        // We don't have to update `wildcard_accesses` as the new node is not exposed and
        // has no children so the default counts of 0 are correct.

```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -261,14 +261,8 @@ pub struct LocationTree {
...
-    /// Maps a tag and a location to its wildcard access tracking information,
-    /// with possible lazy initialization.
-    ///
-    /// If this allocation doesn't have any exposed nodes, then this map doesn't get
-    /// initialized. This way we only need to allocate the map if we need it.
-    ///
-    /// NOTE: same guarantees on entry initialization as for `perms`.
-    pub wildcard_accesses: UniValMap<WildcardState>,
+    /// Stores how nodes are related to exposed nodes,
+    pub wildcard_accesses: WildcardData,
```
RalfJung: The name of the field and the type are inconsistent.

And generally this is more about exposed nodes than about wildcard pointers, so maybe the name should reflect that? Something like `exposed_cache: ExposedNodeCache` or so? It's a cache in the sense that it just caches information that could also be recomputed.

RalfJung (commented):
That's a very nice overall code reduction :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4851/39c535f88079a4170bf0abadae2b49a8a9c48634..2d0044f51bfd69db2021c6fb96ec5aa05bed8528)*

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber: Previously we checked the `max_foreign_access` field which would be set to `Write` on wildcard roots. This PR removes `max_foreign_access` from `WildcardState` so we can't use that anymore.

src/borrow_tracker/tree_borrows/wildcard.rs
royAmmerschuber: `UniValMap` only allocates its backing storage if you insert an element, and we only insert an element if a node gets exposed.

RalfJung: That's pulling out implementation details. And in fact you don't even fully allocate the map when a pointer gets exposed, right? So just say that there's an implied default and that should cover all the cases.

royAmmerschuber: The original reason why we put this wildcard state into a separate UniValMap instead of storing it as part of LocationState is because we did not want to increase the memory usage for Trees without exposed nodes.
This is why I documented the fact nothing gets allocated if we do not expose nodes.

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: Thanks, makes sense.  Please update the comment then.

src/borrow_tracker/tree_borrows/wildcard.rs
RalfJung: The current comment is confusing, and fails to explain the most important point which is the implicit default.

If you want to also explain the allocation point, sure, but you need to be a bit less terse about it then.

src/borrow_tracker/tree_borrows/tree.rs
royAmmerschuber: I think it makes sense to change it to `wildcard_cache` as a name. I feel like `exposed` is a too general term considering this specifically tracks the relatedness for a wildcard access. 

This also means the doc-comment is wrong as we only care about exposed nodes through which an access is possible.

RalfJung: This specifically tracks how many accessible exposed children every node has. So having "exposed" in the name seems quite apt to me?

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -51,373 +50,142 @@ impl WildcardAccessRelatedness {
     }
 }
 
+/// Caches information about the relatedness of nodes for a wildcard access.
+///
+/// Initially treats all nodes as unexposed. Needs to be updated when a new node gets exposed
+/// or an exposed nodes permission changes.
+///
+/// Only allocates memory if we add an exposed node.
```
RalfJung:
```suggestion
/// Nodes not in this map have a default WildcardState, i.e. they have no exposed children.
/// In particular, this map remains empty (and thus consumes no memory) until the first
/// node in the tree gets exposed.
```

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -51,373 +50,142 @@ impl WildcardAccessRelatedness {
...
-                Write => write_foreign_access,
-                Read => read_foreign_access,
-                None => none_foreign_access,
-            };
+                // If some but not all of the valid exposed nodes are local then we cannot determine the correct relatedness.
+                EitherAccess
+            })
+        };
+        // If this is a wildcard tree then an access can always be foreign as
+        // it could come from another tree.
```
RalfJung: Wouldn't it be easier to handle this by incrementing `total_num` by 1 if `is_wildcard_tree` is true, before the logic above?

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -51,373 +50,142 @@ impl WildcardAccessRelatedness {
...
+        } else {
+            relatedness
+        };
 
-            if new_foreign_access != state.max_foreign_access {
-                Some((child, new_foreign_access))
-            } else {
-                Option::None
+        if only_foreign {
+            match relatedness {
```
RalfJung, line 130:
```suggestion
            // The is definitely not a local access; clamp the result accordingly.
            match relatedness {
```

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -428,25 +196,27 @@ impl Tree {
...
+            for (_, loc) in self.locations.iter_mut_all() {
+                let perm = loc
+                    .perms
+                    .get(id)
+                    .map(|p| p.permission())
+                    .unwrap_or_else(|| node.default_location_state().permission());
+
+                let access_level = perm.strongest_allowed_local_access(protected);
+                // an unexposed node gets treated as access level `None`.
+                // So initial exposure transitions from `None` to the nodes actual `access_level`.
```
RalfJung:
```suggestion
                // An unexposed node gets treated as access level `None`. Therefore,
                // the initial exposure transitions from `None` to the node's actual `access_level`.
```

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: Specifically, the data structure speeds up the search of "figure out whether there is an exposed node (with particular permissions) in this part of the tree". So the more I think about it the stronger I think the key term here is "exposed", not "wildcard".

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -261,22 +261,16 @@ pub struct LocationTree {
...
     pub perms: UniValMap<LocationState>,
-    /// Maps a tag and a location to its wildcard access tracking information,
-    /// with possible lazy initialization.
-    ///
-    /// If this allocation doesn't have any exposed nodes, then this map doesn't get
-    /// initialized. This way we only need to allocate the map if we need it.
-    ///
-    /// NOTE: same guarantees on entry initialization as for `perms`.
-    pub wildcard_accesses: UniValMap<WildcardState>,
+    /// Caches information about the relatedness of nodes for a wildcard access.
```
RalfJung, line 264:
```suggestion
    /// Caches information about where in the tree active exposed nodes
    /// (i.e. exposed nodes that can perform read/write accesses) are located.
```

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -51,374 +50,138 @@ impl WildcardAccessRelatedness {
...
-impl Debug for WildcardState {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("WildcardState")
-            .field("child_r/w", &(self.child_reads, self.child_writes))
-            .field("foreign", &self.max_foreign_access)
-            .field("exposed_as", &self.exposed_as)
-            .finish()
-    }
+#[derive(Clone, Default, Debug, PartialEq, Eq)]
+struct ExposedState {
```
RalfJung: `ExposedState` is an odd name for this. What about `NodeExposedCache` or `ExposedCacheNode` or something like that?

src/borrow_tracker/tree_borrows/wildcard.rs
```diff
@@ -98,6 +97,13 @@ impl WildcardCache {
             AccessKind::Read => (root.local_reads, node.local_reads),
             AccessKind::Write => (root.local_writes, node.local_writes),
         };
+
+        // If this is a wildcard tree then an access can always be foreign as
+        // it could come from another tree.
+        // We can represent this by adding 1 to the total which means there
+        // always exists a foreign exposed node.
+        let total_num = total_num + u16::from(is_wildcard_tree);
```
RalfJung, line 109: This makes me wonder, should we maybe store that +1 already in the root of wildcard trees directly, rather than having to add it here?

src/borrow_tracker/tree_borrows/wildcard.rs
RalfJung: Oh wait, `ExposedCache` is already per-node, right?
So... `LocationExposedCache` or something like that?

royAmmerschuber: `ExposedCache` is per location and stores an `ExposedState` for each node.
I'd go with `ExposedCacheNode`.

RalfJung: Ah okay. :+1: 

> ExposedCache is per location

Please add that to the doc comment.

src/borrow_tracker/tree_borrows/wildcard.rs
royAmmerschuber: That would be incorrect for the root. The root would think this extra exposed node was a child of it. So the access would always be local to it instead of foreign/either.

RalfJung: Good point!

RalfJung:
Thanks a lot! I just did some comment tweaks and squashed the commits.

Issue #4096, opened 2024-12-16
Miri hangs when reading file at `/proc/self/fd/X`


stevenengler:
If your program attempts to open and read from a file in `/proc/self/fd/`, then `cargo miri run` (with isolation disabled) hangs.

```rust
use std::os::fd::AsRawFd;
use std::fs::File;
use std::io::{Read, Seek, Write};

fn main() {
    let mut file = tempfile::tempfile().unwrap();
    file.write_all(b"hello").unwrap();
    file.rewind().unwrap();

    // opening and reading from /proc cause miri to become stuck
    let mut new_file = File::open(format!("/proc/self/fd/{}", file.as_raw_fd())).unwrap();
    //let mut new_file = file;

    let mut buf = Vec::new();
    new_file.read_to_end(&mut buf).unwrap();
    println!("{buf:?}");
}
```

```text
$ cargo --version
cargo 1.85.0-nightly (769f622e1 2024-12-14)
$ MIRIFLAGS=-Zmiri-disable-isolation cargo miri run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.01s
     Running `/home/steve/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/cargo-miri runner target/miri/x86_64-unknown-linux-gnu/debug/miri-test`
```

This is also reproducible at https://play.rust-lang.org/?version=stable&mode=debug&edition=2021

I'm guessing that miri virtualizes filesystem operations in some way, and maybe files in `/proc` just aren't supported?

RalfJung:
Miri virtualizes file descriptors, so the FD number you see may have nothing to do with the FD numbers in `/proc`.

So I am inclined to close this as not-a-bug; we don't support bypassing Miri and directly accessing the underlying OS. You will run into similar issues trying to access `/proc/self/mem`. Maybe we should print a warning or even halt execution when we detect a path in `/proc`, but that would just be a best-effort check.

stevenengler:
Is there a general guideline about what Linux features Miri emulates (or wants to emulate) and which ones it doesn't? For example Miri is emulating Linux file descriptor handles (and I'm guessing some syscalls), but is not emulating `/proc`.

I think it sounds fine to not support `/proc`, but it would be nice if a warning was shown when possible. Running tests in miri can already take ~20 minutes for some projects, and having a correct test which can block indefinitely under miri can be a bit difficult to debug.

[...]

PR #4857, opened 2026-02-13
Print a warning when trying to open a file in `/proc`

y1lan:
This diagnostic has `DiagLevel::Warning` and is emitted when opening a file under `/proc` on Unix platforms as discussed in issue rust-lang/miri#4096

### DIFF
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -354,23 +354,15 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let this = self.eval_context_mut();
 
         let path_raw = this.read_pointer(path_raw)?;
-        let path = this.read_path_from_c_str(path_raw)?;
-
-        match path::absolute(&path) {
-            Ok(path) => {
-                // "/proc" does not work properly from inside Miri, warn about opening files there.
-                if matches!(
-                    this.tcx.sess.target.os,
-                    Os::Linux | Os::Android | Os::Illumos | Os::Solaris
-                ) && (path == Path::new("/proc") || path.starts_with("/proc/"))
-                {
-                    this.machine.emit_diagnostic(NonHaltingDiagnostic::FileInProcOpened);
-                }
-            }
-            Err(_) => { /* when the path is empty, it is also unrelative to "/proc" issue */ }
-        }
-
         let flag = this.read_scalar(flag)?.to_i32()?;
+
+        let path = this.read_path_from_c_str(path_raw)?;
+        // Files in `/proc` won't work properly.
+        if matches!(this.tcx.sess.target.os, Os::Linux | Os::Android | Os::Illumos | Os::Solaris)
+            && path::absolute(&path).is_ok_and(|path| path.starts_with("/proc"))
+        {
+            this.machine.emit_diagnostic(NonHaltingDiagnostic::FileInProcOpened);
+        }
 
         let mut options = OpenOptions::new();
 


y1lan:
previous CI failed due to network error so I force-pushed the branch to trigger CI aging

RalfJung:
Thanks for the PR!

Please update the PR title to be about the user-visible effect (warn on `/proc` accesses), not implementation details like which enum gets a new variant.

Also, please add a test.

@rustbot author

y1lan:
How can I disable the test being ran on host windows  for linux target ?

y1lan:
> How can I disable the test being ran on host windows for linux target ?

I have found the configuration entry `ignore-host`

tests/pass/open_a_file_in_proc.rs
```diff
@@ -0,0 +1,15 @@
+//@compile-flags: -Zmiri-disable-isolation
+//@only-target: linux android illumos
+//@ignore-host: windows
+use std::io::Read;
+
+fn main() {
+    let _ = match std::fs::File::open("/proc/self/cmdline") {
```
RalfJung: Maybe make the test open `/proc/doesnotexist` so that we *expect* the file to not exist? That will make the test work on e.g. macOS hosts as well.

src/shims/unix/fs.rs
```diff
@@ -355,6 +355,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
         let path_raw = this.read_pointer(path_raw)?;
         let path = this.read_path_from_c_str(path_raw)?;
+        if path.starts_with("/proc") {
```
RalfJung: Please check `tcx.sess.target` to restrict the warning to Linux (and Android) targets.

y1lan: what about illumos?

RalfJung: Do they have `/proc`? What about other Unixes?

y1lan: As I have searched, Illumos and Solaris do have `/proc` and `/proc/self`. In FreeBSD and NetBSD, procfs is optional.        In MacOs and Darwin, procfs does not exist. these left unixes are hard to check one by one.

src/diagnostics.rs
```diff
@@ -701,6 +703,7 @@ impl<'tcx> MiriMachine<'tcx> {
                 };
                 format!("GenMC currently does not model the failure ordering for `compare_exchange`. {was_upgraded_msg}. Miri with GenMC might miss bugs related to this memory access.")
             }
+            FileInProcOpened => format!("miri virtualizes file descriptors, so opening a file in `/proc` may not work as expected as it would on the host.")
```
```suggestion
            FileInProcOpened => format!("files in `/proc` can bypass the Abstract Machine and might not work properly in Miri")
```

src/shims/unix/fs.rs
```diff
@@ -355,6 +355,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
         let path_raw = this.read_pointer(path_raw)?;
         let path = this.read_path_from_c_str(path_raw)?;
+        if path.starts_with("/proc")
+            && matches!(
+                this.tcx.sess.target.os,
+                Os::Linux | Os::Android | Os::Illumos | Os::Solaris
+            )
+        {
+            this.machine.emit_diagnostic(NonHaltingDiagnostic::FileInProcOpened);
+        }
```
RalfJung:
```suggestion
        // "/proc" does not work properly from inside Miri, warn about opening files there.
        if matches!(
                this.tcx.sess.target.os,
                Os::Linux | Os::Android | Os::Illumos | Os::Solaris
            )
            && path.starts_with("/proc")
        {
            this.machine.emit_diagnostic(NonHaltingDiagnostic::FileInProcOpened);
        }

```

RalfJung:
Thanks! I have simplified the check a bit, please have a look at the new commit I added. I also moved the cheap check (target OS) before the expensive path check to avoid unnecessary work.

Issue #4563, opened 2025-09-03
Remove our RDTSC code?


RalfJung:
In https://github.com/rust-lang/miri/pull/4524 we gained some code to directly access the x86 RDTSC clock for profiling purposes. However, on almost all systems, `std::time::Instant` should already be using that clock. So maybe instead of carrying such low-level code we should document how to diagnose and potentially work-around the case where the kernel doesn't trust the RDTSC clock? If the clock is so unstable that the kernel doesn't use it, maybe it's also not a great fit for profiling...

See the discussion [here](https://github.com/rust-lang/miri/pull/4524#issuecomment-3234290380).
Cc @Stypox

bjorn3:
`std::time::Instant` goes through libc, which in turn likely goes through the VDSO which finally may use rdtsc. This is more overhead than directly using rdtsc, as libc and the VDSO can't be inlined.

Edit: Never mind, the comment above the linked one says that the overhead is pretty small.

Stypox:
I am ok with reverting the RDTSC part of #4524. Furthermore, it seems to cause another problem now: when a parameter passed to a tracing call uses a `Debug` implementation that, while producing the debug string, makes another (nested) tracing call, rustc will crash. The reason is that the thread-local data used for subtracting away the time spent within the tracing infrastructure is stored in a thread-local `RefCell`, see [here](https://github.com/rust-lang/miri/blob/master/src/bin/log/tracing_chrome.rs#L564), and a nested tracing call will cause this `RefCell` to be borrowed twice.

[This](https://github.com/rust-lang/rust/blame/master/compiler/rustc_span/src/lib.rs#L2302) is the tracing call happening while formatting the span parameter passed to [this other tracing call](https://github.com/rust-lang/rust/blob/a41214f9bd7ce869d21898f31107f92c90a9112f/compiler/rustc_const_eval/src/interpret/step.rs#L82-L89), causing:

```
thread 'rustc' panicked at src/tools/miri/src/bin/log/tracing_chrome.rs:568:41:
RefCell already borrowed
...
  16:     0x561ad80b2591 - {closure#0}<tracing_subscriber::registry::sharded::Registry, miri::log::tracing_chrome::{impl#4}::on_event::{closure_env#0}<tracing_subscriber::registry::sharded::Registry>>
                               at ./src/tools/miri/src/bin/log/tracing_chrome.rs:568:41
...
  24:     0x7f16af671ff7 - {closure#1}
[...]

Issue #4752, opened 2025-12-08
Fine-grained event tracing does not work any more


RalfJung:
https://github.com/rust-lang/rust/pull/147613 caused a regression in our [tracing system](https://github.com/rust-lang/miri/blob/master/doc/tracing.md): the usual log filters now also get applied to the tracing events, meaning they are all filtered out and never make it into the JSON trace. See [this](https://github.com/rust-lang/rust/pull/147613#issuecomment-3415780630) and the following discussion for more details.

PR #4860, opened 2026-02-16
Miscellaneous tracing fixes and improvements

Stypox:
Fixes `RefCell` being borrowed mutably twice, in case of nested tracing calls while formatting an argument of a tracing call, see https://github.com/rust-lang/miri/issues/4563#issuecomment-3415871750. There are some places in rustc where that happens, e.g. [here](https://github.com/rust-lang/rust/blob/f6092f224d2b1774b31033f12d0bee626943b02f/compiler/rustc_const_eval/src/interpret/step.rs#L82-L89)->[here](https://github.com/rust-lang/rust/blob/f6092f224d2b1774b31033f12d0bee626943b02f/compiler/rustc_public/src/ty.rs#L265)->[here](https://github.com/rust-lang/rust/blob/f6092f224d2b1774b31033f12d0bee626943b02f/compiler/rustc_span/src/lib.rs#L2302).

After applying the fix in 3367573b6565581b0d457b587a3888e43ffebfff (which avoids keeping the `RefCell` borrowed while formatting arguments of tracing calls), I was surprised that the code was still crashing on [the same tracing call](https://github.com/rust-lang/rust/blob/f6092f224d2b1774b31033f12d0bee626943b02f/compiler/rustc_const_eval/src/interpret/step.rs#L82-L89), this time with a different backtrace that didn't contain any mention of our `tracing-chrome` layer. So I investigated a bit and it turned out to be a bug in `tracing-tree`, for which I could create a minimum reproducible example that has nothing to do with `tracing-chrome`: https://github.com/davidbarsky/tracing-tree/issues/93.

This PR also closes rust-lang/miri#4563 by removing unnecessary RDTSC code. This could possibly go in a separate PR but it touches the same code as the fix above, so I preferred to just make it a separate commit.

This PR finally updates the documentation for tracing, to explain how to ensure the system uses `tsc` as a timer, and to point out the two blockers that prevent tracing from actually being used at the moment.

Note: this PR uses f6466ce65 as a base, instead of the latest commit on `master`, because that's the last Miri commit that is not affected by https://github.com/rust-lang/miri/issues/4752, and where `tracing` could still be tested. Before merging I will rebase onto `master`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -189,6 +189,8 @@ you can visualize in [Perfetto](https://ui.perfetto.dev/). For example:
 MIRI_TRACING=1 ./miri run --features=tracing tests/pass/hello.rs
 ```
 
+See [doc/tracing.md](./doc/tracing.md) for more information.
+
 ### UI testing
 
 We use ui-testing in Miri, meaning we generate `.stderr` and `.stdout` files for the output
--- a/doc/tracing.md
+++ b/doc/tracing.md
@@ -2,6 +2,9 @@
 
 Miri can be traced to understand how much time is spent in its various components (e.g. borrow tracker, data race checker, etc.). When tracing is enabled, running Miri will create a `.json` trace file that can be opened and analyzed in [Perfetto](https://ui.perfetto.dev/). For any questions regarding this documentation you may contact [Stypox](https://rust-lang.zulipchat.com/#narrow/dm/627563-Stypox) on Zulip.
 
+> [!WARNING]
+> Tracing in Miri at the moment is broken due to two bugs in tracing libraries: https://github.com/tokio-rs/tracing/pull/3392 and https://github.com/davidbarsky/tracing-tree/issues/93. Also see https://github.com/rust-lang/miri/issues/4752.
+
 ## Obtaining a trace file
 
 ### From the Miri codebase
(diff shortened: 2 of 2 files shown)


src/bin/log/tracing_chrome.rs
```diff
@@ -562,28 +564,46 @@ where
...
-                Some(thread_data) => (thread_data, false),
-                None => {
-                    let tid = self.max_tid.fetch_add(1, Ordering::SeqCst);
-                    let out = self.out.lock().unwrap().clone();
-                    let start = TracingChromeInstant::setup_for_thread_and_start(tid);
-                    *thread_data = Some(ThreadData { tid, out, start });
-                    (thread_data.as_mut().unwrap(), true)
+            // make sure not to keep `value` borrowed when calling `f` below, since the user tracing
+            // code that `f` might invoke (e.g. fmt::Debug argument formatting) may contain nested
+            // tracing calls that would cause `value` to be doubly-borrowed mutably
```
```suggestion
            // Make sure not to keep `value` borrowed when calling `f` below, since the user tracing
            // code that `f` might invoke (e.g. fmt::Debug argument formatting) may contain nested
            // tracing calls that would cause `value` to be doubly-borrowed mutably.
```

RalfJung (commented):
Looks great, thanks a lot! Please rebase.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4860/ef30e906aae9520104dd54b81f648f770149d6bb..c313aa3adaed985ee370f449c0ea61f93b0d87bd)*

PR #4862, opened 2026-02-17
use the unstable `carryless_mul` in the `pclmul` implementation

folkertdev:
Eventually we should be able to use the intrinsics in `stdarch` so that miri no longer needs a custom implementation. Unfortunately the required optimizations didn't make it into LLVM 22, so we'll have to wait ~6 months.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,6 +22,7 @@
 #![feature(derive_coerce_pointee)]
 #![feature(arbitrary_self_types)]
 #![feature(macro_metavar_expr)]
+#![feature(uint_carryless_mul)]
 // Configure clippy and other lints
 #![allow(
     clippy::collapsible_else_if,
--- a/src/shims/x86/mod.rs
+++ b/src/shims/x86/mod.rs
@@ -1178,21 +1178,7 @@ fn pclmulqdq<'tcx>(
         let index = if (imm8 & 0x10) == 0 { lo } else { hi };
         let right = ecx.read_scalar(&ecx.project_index(&right, index)?)?.to_u64()?;
 
-        // Perform carry-less multiplication.
-        //
-        // This operation is like long multiplication, but ignores all carries.
-        // That idea corresponds to the xor operator, which is used in the implementation.
-        //
-        // Wikipedia has an example https://en.wikipedia.org/wiki/Carry-less_product#Example
-        let mut result: u128 = 0;
-
-        for i in 0..64 {
-            // if the i-th bit in right is set
-            if (right & (1 << i)) != 0 {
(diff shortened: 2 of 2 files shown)


folkertdev:
CI uses `2026-01-21/rustc-nightly-x86_64-unknown-linux-gnu.tar.xz` for that bootstrap job, How regularly does that update?

RalfJung:
It always the current rust nightly bootstrap compiler (the file storing that is [here](https://github.com/rust-lang/rust/blob/main/src/stage0)). If the job here fails, that means merging this PR would cause the next sync into the Rust repo to fail.

RalfJung:
The rust bootstrap compiler always gets updated around a release, so the next update will be around Feb 27th.

It is a fundamental part of the rustc development process that new standard library features can only be used inside rustc/miri after the next bootstrap bump (when nightly becomes beta and then that beta becomes the bootstrap compiler).

src/shims/x86/mod.rs
```diff
@@ -1178,24 +1178,8 @@ fn pclmulqdq<'tcx>(
...
-            // if the i-th bit in right is set
-            if (right & (1 << i)) != 0 {
-                // xor result with `left` shifted to the left by i positions
-                result ^= u128::from(left) << i;
-            }
-        }
-
         let dest = ecx.project_index(&dest, i)?;
-        ecx.write_scalar(Scalar::from_u128(result), &dest)?;
+        ecx.write_scalar(Scalar::from_u128(left.widening_carryless_mul(right)), &dest)?;
```
RalfJung: I think the better approach here would be to share this with the interpreter-level implementation of the intrinsic that landed in https://github.com/rust-lang/rust/pull/152132.

folkertdev: We reverted to just using the naive implementation in that PR, so there is no specific support in the interpreter for this operation currently.

RalfJung: Please separately let-bind the result; currently it is easy to miss the main computation among all the other stuff.

RalfJung:
Once https://github.com/rust-lang/rust/pull/153462 lands, the next sync should make this PR work.

src/lib.rs
```diff
@@ -18,6 +18,7 @@
 #![feature(arbitrary_self_types)]
 #![feature(iter_advance_by)]
 #![feature(macro_metavar_expr)]
+#![feature(uint_carryless_mul)]
```
kellyerinmontgomery426-ui, line 25: Not the intended abrivaints or hyphens and words scrutiny

RalfJung:
Thanks!

PR #4868, opened 2026-02-22
Start network socket shim

WhySoBad:
Hi,

This pull request is the start for the network socket support. Currently, the `libc::socket` and `libc::close` function calls work.

I also left some comments about uncertainties with flags but I've seen that rust-lang/miri#4840 addresses those. So, I'd suggest keeping it with the comments for now and as soon as rust-lang/miri#4840 is merged I'll update it.


RalfJung:
r? me

RalfJung:
Please remove the last commit from this PR again. Reviews are much easier in smaller chunks, and when the PR doesn't keep changing. :)

src/shims/unix/socket.rs
```diff
@@ -0,0 +1,176 @@
...
+enum SocketFamily {
+    // IPv4 internet protocols
+    IPv4,
+    // IPv6 internet protocols
+    IPv6,
+}
+
+#[derive(Debug)]
+enum SocketType {
+    // Reliable full-duplex communication, based on connections
```
RalfJung: Please write full sentences including punctuation.
```suggestion
    /// Reliable full-duplex communication, based on connections.
```

src/shims/unix/socket.rs
```diff
@@ -0,0 +1,186 @@
...
+    /// Family of the socket, used to ensure socket only binds/connects to address of
+    /// same family.
+    family: SocketFamily,
+    /// Type of the socket, either datagram or stream.
+    /// Only stream is supported at the moment!
+    socket_type: SocketType,
+    /// Inner standard library socket used for shimming.
+    /// Depending on whether `bind` or `connect` is called, we use a listener or a stream.
+    /// This is `None` until either of those methods is called.
+    socket: RefCell<Option<SocketKind>>,
```
RalfJung: Since `bind` and `connect` are not in this PR, please just remove this field.

src/shims/unix/socket.rs
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        // Ignore all file access mode flags
+        flag &= !(o_rdonly | o_wronly | o_rdwr);
+
+        if flag != 0 {
+            throw_unsup_format!("fcntl: only O_NONBLOCK is supported for F_SETFL on sockets");
+        }
+
+        interp_ok(Scalar::from_i32(0))
```
RalfJung: Just make this entire thing `throw_unsup_format` for now. We said only `socket` and `close` should work :)

I expect this to change quite a bit when https://github.com/rust-lang/miri/pull/4840 lands.

src/shims/unix/socket.rs
```diff
@@ -0,0 +1,186 @@
...
+            this.set_last_error(LibcError("EACCES"))?;
+            return interp_ok(Scalar::from_i32(-1));
+        }
+
+        let mut is_sock_nonblock = false;
+
+        // Interpret the flag. Every flag we recognize is "subtracted" from `flags`, so
+        // if there is anything left at the end, that's an unsupported flag.
+        if matches!(this.tcx.sess.target.os, Os::Linux | Os::Android) {
+            // SOCK_NONBLOCK only exists on Linux.
```
RalfJung: What about SOCK_CLOEXEC?

src/shims/unix/socket.rs
```diff
@@ -0,0 +1,186 @@
...
+            let sock_nonblock = this.eval_libc_i32("SOCK_NONBLOCK");
+            let sock_cloexec = this.eval_libc_i32("SOCK_CLOEXEC");
+            if flags & sock_nonblock == sock_nonblock {
+                is_sock_nonblock = true;
+                flags &= !sock_nonblock;
+            }
+            if flags & sock_cloexec == sock_cloexec {
+                flags &= !sock_cloexec;
```
RalfJung, line 118:
```suggestion
                // We don't support `exec` so we can ignore this.
                flags &= !sock_cloexec;
```

src/shims/unix/socket.rs
```diff
@@ -0,0 +1,186 @@
...
+            );
+        };
+
+        if flags != this.eval_libc_i32("SOCK_STREAM") {
+            throw_unsup_format!(
+                "socket: type {:#x} is unsupported, only SOCK_STREAM, \
+                                SOCK_CLOEXEC and SOCK_NONBLOCK are allowed",
+                flags
+            );
+        } else if protocol != 0 {
```
```suggestion
        }
        if protocol != 0 {
```
These are logically unrelated, `else` is just confusing.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -0,0 +1,11 @@
+//@ignore-target: windows # No libc socket on Windows
+//@ignore-target: solaris # Does socket is a macro for __xnet7_socket which has no shim
+//@ignore-target: illumos # Does socket is a macro for __xnet7_socket which has no shim
+//@compile-flags: -Zmiri-disable-isolation
+
+fn main() {
```
RalfJung, line 10: In preparation for later adding more tests here: please put this test into a function like `test_socket_close` or so and then have `main` just call that function.

src/shims/unix/socket.rs
WhySoBad: Oops, I misread the "validate socket flags" as "set socket flags". Removed it.

src/shims/unix/socket.rs
WhySoBad: I've updated the comment and also found that both of these options also [exist on FreeBSD](https://man.freebsd.org/cgi/man.cgi?query=socket&apropos=0&sektion=0&manpath=FreeBSD+15.0-RELEASE+and+Ports&format=html).

tests/pass-dep/libc/libc-socket-with-isolation.rs
RalfJung, line 1: This test did not need the change to separate function, it won't get any more cases anyway since sockets don't work with isolation. So please change it back. I deliberately only mentioned this change for the other test. :)

In fact, shouldn't the test assert that `libc::socket` fails? `socket` has the usual "-1 is error" convention so all calls to it should use `errno_result`. Grep the test suite for examples; you'll have to add something like this to import the helper:
```
#[path = "../../utils/libc.rs"]
mod libc_utils;
```

Please use `errno_result` in both tests.

RalfJung:

@rustbot author

PR #4882, opened 2026-03-01
Add network socket `listen` and `bind` shims

WhySoBad:
Hi,

This pull request adds shims for the `bind` and `listen` syscalls. To support the implementation of `TcpListener::bind` from the standard library, a small set of options for the `setsockopt` syscall are added as well.

For the `bind` shim I'm a bit uncertain about the validation we should perform there. Since we effectively only execute this syscall when seeing a `listen` later on, the current implementation tries to validate the things it can easily do with the data available (e.g. whether socket and address family match or whether address length is valid). 
Things like `EADDRINUSE` for example wouldn't be caught with the current implementation as it would require looking up the address manually and I don't know whether that would be worth it. 

I can see that it might be confusing for users when we only throw a subset of errors after the `bind` syscall and also when we then afterwards throw errors for the `bind` syscall once they call `listen` but I think this problem is unavoidable.

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -150,6 +150,13 @@ pub enum NonHaltingDiagnostic {
         effective_failure_ordering: AtomicReadOrd,
     },
     FileInProcOpened,
+    SocketListenUnsupportedBacklog {
+        details: bool,
+        /// Unsupported backlog value provided the by the program.
+        provided: i32,
+        /// Supported backlog value by Miri.
+        supported: i32,
+    },
 }
 
 /// Level of Miri specific diagnostics
@@ -656,6 +663,8 @@ impl<'tcx> MiriMachine<'tcx> {
             | WeakMemoryOutdatedLoad { .. } =>
                 ("tracking was triggered here".to_string(), DiagLevel::Note),
             FileInProcOpened => ("open a file in `/proc`".to_string(), DiagLevel::Warning),
+            SocketListenUnsupportedBacklog { .. } =>
+                ("call to `listen` with unsupported backlog value".to_string(), DiagLevel::Warning),
         };
 
         let title = match &e {
@@ -703,13 +712,18 @@ impl<'tcx> MiriMachine<'tcx> {
                 };
                 format!("GenMC currently does not model the failure ordering for `compare_exchange`. {was_upgraded_msg}. Miri with GenMC might miss bugs related to this memory access.")
             }
-            FileInProcOpened => format!("files in `/proc` can bypass the Abstract Machine and might not work properly in Miri")
+            FileInProcOpened => format!("files in `/proc` can bypass the Abstract Machine and might not work properly in Miri"),
+            SocketListenUnsupportedBacklog { provided, supported, .. } => format!("called `listen` on socket with backlog value of {provided} but only {supported} is supported")
         };
 
         let notes = match &e {
             ProgressReport { block_count } => {
                 vec![note!("so far, {block_count} basic blocks have been executed")]
             }
+            SocketListenUnsupportedBacklog { details: true, supported, .. } =>
+                vec![note!(
+                    "the given value will be ignored and a backlog of {supported} will be used instead"
+                )],
             _ => vec![],
         };
 
--- a/src/shims/unix/socket.rs
+++ b/src/shims/unix/socket.rs
@@ -1,12 +1,17 @@
-use std::cell::Cell;
-use std::net::{TcpListener, TcpStream};
+use std::cell::{Cell, RefCell};
+use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6, TcpListener};
 
+use rustc_abi::Size;
 use rustc_const_eval::interpret::{InterpResult, interp_ok};
 use rustc_middle::throw_unsup_format;
 use rustc_target::spec::Os;
 
+use crate::diagnostics::SpanDedupDiagnostic;
 use crate::shims::files::{FdId, FileDescription};
 use crate::{OpTy, Scalar, *};
+
+/// Backlog value passed to the `listen` syscall by the standard library
+const SUPPORTED_LISTEN_BACKLOG: i32 = 128;
 
 #[derive(Debug, PartialEq)]
 enum SocketFamily {
@@ -17,27 +22,25 @@ enum SocketFamily {
 }
 
 #[derive(Debug)]
-enum SocketType {
-    /// Reliable full-duplex communication, based on connections.
-    Stream,
+enum SocketState {
+    /// No syscall after `socket` has been made.
+    Initial,
+    /// The `bind` syscall has been called on the socket.
+    /// This is only reachable from the [`SocketState::Initial`] state.
+    Bound(SocketAddr),
+    /// The `listen` syscall has been called on the socket.
+    /// This is only reachable from the [`SocketState::Bound`] state.
+    #[expect(unused)]
+    Listening(TcpListener),
 }
 
-#[allow(unused)]
-#[derive(Debug)]
-enum SocketKind {
-    TcpListener(TcpListener),
-    TcpStream(TcpStream),
-}
-
-#[allow(unused)]
 #[derive(Debug)]
 struct Socket {
     /// Family of the socket, used to ensure socket only binds/connects to address of
     /// same family.
     family: SocketFamily,
-    /// Type of the socket, either datagram or stream.
-    /// Only stream is supported at the moment!
-    socket_type: SocketType,
+    /// Current state of the inner socket.
+    state: RefCell<SocketState>,
     /// Whether this fd is non-blocking or not.
     is_non_block: Cell<bool>,
 }
@@ -97,16 +100,19 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // Reject if isolation is enabled
         if let IsolatedOp::Reject(reject_with) = this.machine.isolated_op {
             this.reject_in_isolation("`socket`", reject_with)?;
-            this.set_last_error(LibcError("EACCES"))?;
-            return interp_ok(Scalar::from_i32(-1));
+            return this.set_last_error_and_return_i32(LibcError("EACCES"));
         }
 
         let mut is_sock_nonblock = false;
 
         // Interpret the flag. Every flag we recognize is "subtracted" from `flags`, so
         // if there is anything left at the end, that's an unsupported flag.
-        if matches!(this.tcx.sess.target.os, Os::Linux | Os::Android | Os::FreeBsd) {
-            // SOCK_NONBLOCK and SOCK_CLOEXEC only exist on Linux, Android and FreeBSD.
+        if matches!(
+            this.tcx.sess.target.os,
+            Os::Linux | Os::Android | Os::FreeBsd | Os::Solaris | Os::Illumos
+        ) {
+            // SOCK_NONBLOCK and SOCK_CLOEXEC only exist on Linux, Android, FreeBSD,
+            // Solaris, and Illumos targets
             let sock_nonblock = this.eval_libc_i32("SOCK_NONBLOCK");
             let sock_cloexec = this.eval_libc_i32("SOCK_CLOEXEC");
             if flags & sock_nonblock == sock_nonblock {
@@ -148,10 +154,311 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let fds = &mut this.machine.fds;
         let fd = fds.new_ref(Socket {
             family,
+            state: RefCell::new(SocketState::Initial),
             is_non_block: Cell::new(is_sock_nonblock),
-            socket_type: SocketType::Stream,
         });
 
         interp_ok(Scalar::from_i32(fds.insert(fd)))
+    }
+
+    fn bind(
+        &mut self,
+        socket: &OpTy<'tcx>,
+        address: &OpTy<'tcx>,
+        address_len: &OpTy<'tcx>,
+    ) -> InterpResult<'tcx, Scalar> {
+        let this = self.eval_context_mut();
+
+        let socket = this.read_scalar(socket)?.to_i32()?;
+        let address = match this.socket_address(address, address_len, "bind")? {
+            Ok(addr) => addr,
+            Err(e) => return this.set_last_error_and_return_i32(e),
+        };
+
+        // Get the file handle
+        let Some(fd) = this.machine.fds.get(socket) else {
+            return interp_ok(this.eval_libc("EBADF"));
+        };
+
+        let Some(socket) = fd.downcast::<Socket>() else {
+            // Man page specifies to return ENOTSOCK if `fd` is not a socket.
+            return interp_ok(this.eval_libc("ENOTSOCK"));
+        };
+
+        assert!(this.machine.communicate(), "cannot have `Socket` with isolation enabled!");
+
+        let mut state = socket.state.borrow_mut();
+
+        match *state {
+            SocketState::Initial => {
+                let address_family = match &address {
+                    SocketAddr::V4(_) => SocketFamily::IPv4,
+                    SocketAddr::V6(_) => SocketFamily::IPv6,
+                };
+
+                if socket.family != address_family {
+                    // Attempted to bind an address from a family that doesn't match
+                    // the family of the socket.
+                    let err = if matches!(this.tcx.sess.target.os, Os::Linux | Os::Android) {
+                        // Linux man page states that `EINVAL` is used when there is an addres family mismatch.
+                        // See <https://man7.org/linux/man-pages/man2/bind.2.html>
+                        LibcError("EINVAL")
+                    } else {
+                        // POSIX man page states that `EAFNOSUPPORT` should be used when there is an address
+                        // family mismatch.
+                        // See <https://man7.org/linux/man-pages/man3/bind.3p.html>
+                        LibcError("EAFNOSUPPORT")
+                    };
+                    return this.set_last_error_and_return_i32(err);
+                }
+
+                *state = SocketState::Bound(address);
+            }
+            SocketState::Bound(_) | SocketState::Listening(_) =>
+                throw_unsup_format!(
+                    "bind: socket is already bound and binding a socket \
+                                    multiple times is unsupported"
+                ),
+        }
+
+        interp_ok(Scalar::from_i32(0))
+    }
+
+    fn listen(&mut self, socket: &OpTy<'tcx>, backlog: &OpTy<'tcx>) -> InterpResult<'tcx, Scalar> {
+        let this = self.eval_context_mut();
+
+        let socket = this.read_scalar(socket)?.to_i32()?;
+        let backlog = this.read_scalar(backlog)?.to_i32()?;
+
+        // Get the file handle
+        let Some(fd) = this.machine.fds.get(socket) else {
+            return interp_ok(this.eval_libc("EBADF"));
+        };
+
+        let Some(socket) = fd.downcast::<Socket>() else {
+            // Man page specifies to return ENOTSOCK if `fd` is not a socket.
+            return interp_ok(this.eval_libc("ENOTSOCK"));
+        };
+
+        assert!(this.machine.communicate(), "cannot have `Socket` with isolation enabled!");
+
+        // Only allow the same backlog value as the standard library uses since the standard library
+        // doesn't provide a way to set a custom value.
+        if backlog != SUPPORTED_LISTEN_BACKLOG {
+            // The first time this happens at a particular location, print a warning.
+            static DEDUP: SpanDedupDiagnostic = SpanDedupDiagnostic::new();
+            this.dedup_diagnostic(&DEDUP, |first| {
+                NonHaltingDiagnostic::SocketListenUnsupportedBacklog {
+                    details: first,
+                    provided: backlog,
+                    supported: SUPPORTED_LISTEN_BACKLOG,
+                }
+            });
+        }
+
+        let mut state = socket.state.borrow_mut();
+
+        match &*state {
+            SocketState::Bound(socket_addr) =>
+                match TcpListener::bind(socket_addr) {
+                    Ok(listener) => *state = SocketState::Listening(listener),
+                    Err(e) => return this.set_last_error_and_return_i32(e),
+                },
+            SocketState::Initial => {
+                throw_unsup_format!(
+                    "listen: listening on a socket which isn't bound is unsupported"
+                )
+            }
+            SocketState::Listening(_) => {
+                throw_unsup_format!("listen: listening on a socket multiple times is unsupported")
+            }
+        }
+
+        interp_ok(Scalar::from_i32(0))
+    }
+
+    fn setsockopt(
+        &mut self,
+        socket: &OpTy<'tcx>,
+        level: &OpTy<'tcx>,
+        option_name: &OpTy<'tcx>,
+        option_value: &OpTy<'tcx>,
+        option_len: &OpTy<'tcx>,
+    ) -> InterpResult<'tcx, Scalar> {
+        let this = self.eval_context_mut();
+
+        let socket = this.read_scalar(socket)?.to_i32()?;
+        let level = this.read_scalar(level)?.to_i32()?;
+        let option_name = this.read_scalar(option_name)?.to_i32()?;
+        let socklen_layout = this.libc_ty_layout("socklen_t");
+        let option_len = this.read_scalar(option_len)?.to_int(socklen_layout.size)?;
+
+        // Get the file handle
+        let Some(fd) = this.machine.fds.get(socket) else {
+            return interp_ok(this.eval_libc("EBADF"));
+        };
+
+        let Some(_socket) = fd.downcast::<Socket>() else {
+            // Man page specifies to return ENOTSOCK if `fd` is not a socket.
+            return interp_ok(this.eval_libc("ENOTSOCK"));
+        };
+
+        if level == this.eval_libc_i32("SOL_SOCKET") {
+            let opt_so_reuseaddr = this.eval_libc_i32("SO_REUSEADDR");
+
+            if matches!(this.tcx.sess.target.os, Os::MacOs | Os::FreeBsd | Os::NetBsd) {
+                // SO_NOSIGPIPE only exists on MacOS, FreeBSD, and NetBSD.
+                let opt_so_nosigpipe = this.eval_libc_i32("SO_NOSIGPIPE");
+
+                if option_name == opt_so_nosigpipe {
+                    if option_len != 4 {
+                        // Option value should be C-int which is usually 4 bytes.
+                        return this.set_last_error_and_return_i32(LibcError("EINVAL"));
+                    }
+                    let option_value =
+                        this.deref_pointer_as(option_value, this.machine.layouts.i32)?;
+                    let _val = this.read_scalar(&option_value)?.to_i32()?;
+                    // We entirely ignore this value since we do not support signals anyway.
+
+                    return interp_ok(Scalar::from_i32(0));
+                }
+            }
+
+            if option_name == opt_so_reuseaddr {
+                if option_len != 4 {
+                    // Option value should be C-int which is usually 4 bytes.
+                    return this.set_last_error_and_return_i32(LibcError("EINVAL"));
+                }
+                let option_value = this.deref_pointer_as(option_value, this.machine.layouts.i32)?;
+                let _val = this.read_scalar(&option_value)?.to_i32()?;
+                // We entirely ignore this: std always sets REUSEADDR for us, and in the end it's more of a
+                // hint to bypass some arbitrary timeout anyway.
+                return interp_ok(Scalar::from_i32(0));
+            } else {
+                throw_unsup_format!(
+                    "setsockopt: option {option_name:#x} is unsupported for level SOL_SOCKET",
+                );
+            }
+        }
+
+        throw_unsup_format!(
+            "setsockopt: level {level:#x} is unsupported, only SOL_SOCKET is allowed"
+        );
+    }
+}
+
+impl<'tcx> EvalContextPrivExt<'tcx> for crate::MiriInterpCx<'tcx> {}
+trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    /// Attempt to turn an address and length operand into a standard library socket address.
+    ///
+    /// Returns an IO error should the address length not match the address family length.
+    fn socket_address(
+        &self,
+        address: &OpTy<'tcx>,
+        address_len: &OpTy<'tcx>,
+        foreign_name: &'static str,
+    ) -> InterpResult<'tcx, Result<SocketAddr, IoError>> {
+        let this = self.eval_context_ref();
+
+        let socklen_layout = this.libc_ty_layout("socklen_t");
+        // We only support address lengths which can be stored in a u64 since the
+        // size of a layout is also stored in a u64.
+        let address_len: u64 =
+            this.read_scalar(address_len)?.to_int(socklen_layout.size)?.try_into().unwrap();
+
+        // Initially, treat address as generic sockaddr just to extract the family field.
+        let sockaddr_layout = this.libc_ty_layout("sockaddr");
+        if address_len < sockaddr_layout.size.bytes() {
+            // Address length should be at least as big as the generic sockaddr
+            return interp_ok(Err(LibcError("EINVAL")));
+        }
+        let address = this.deref_pointer_as(address, sockaddr_layout)?;
+
+        let family_field = this.project_field_named(&address, "sa_family")?;
+        let family_layout = this.libc_ty_layout("sa_family_t");
+        let family = this.read_scalar(&family_field)?.to_int(family_layout.size)?;
+
+        // Depending on the family, decide whether it's IPv4 or IPv6 and use specialized layout
+        // to extract address and port.
+        let socket_addr = if family == this.eval_libc_i32("AF_INET").into() {
+            let sockaddr_in_layout = this.libc_ty_layout("sockaddr_in");
+            if address_len != sockaddr_in_layout.size.bytes() {
+                // Address length should be exactly the length of an IPv4 address.
+                return interp_ok(Err(LibcError("EINVAL")));
+            }
+            let address = address.transmute(sockaddr_in_layout, this)?;
+
+            let port_field = this.project_field_named(&address, "sin_port")?;
+            // Read bytes and treat them as big endian since port is stored in network byte order.
+            let port_bytes: [u8; 2] = this
+                .read_bytes_ptr_strip_provenance(port_field.ptr(), Size::from_bytes(2))?
+                .try_into()
+                .unwrap();
+            let port = u16::from_be_bytes(port_bytes);
+
+            let addr_field = this.project_field_named(&address, "sin_addr")?;
+            let s_addr_field = this.project_field_named(&addr_field, "s_addr")?;
+            // Read bytes and treat them as big endian since address is stored in network byte order.
+            let addr_bytes: [u8; 4] = this
+                .read_bytes_ptr_strip_provenance(s_addr_field.ptr(), Size::from_bytes(4))?
+                .try_into()
+                .unwrap();
+            let addr_bits = u32::from_be_bytes(addr_bytes);
+
+            SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::from_bits(addr_bits), port))
+        } else if family == this.eval_libc_i32("AF_INET6").into() {
+            let sockaddr_in6_layout = this.libc_ty_layout("sockaddr_in6");
+            if address_len != sockaddr_in6_layout.size.bytes() {
+                // Address length should be exactly the length of an IPv6 address.
+                return interp_ok(Err(LibcError("EINVAL")));
+            }
+            // We cannot transmute since the `sockaddr_in6` layout is bigger than the `sockaddr` layout.
+            let address = address.offset(Size::ZERO, sockaddr_in6_layout, this)?;
+
+            let port_field = this.project_field_named(&address, "sin6_port")?;
+            // Read bytes and treat them as big endian since port is stored in network byte order.
+            let port_bytes: [u8; 2] = this
+                .read_bytes_ptr_strip_provenance(port_field.ptr(), Size::from_bytes(2))?
+                .try_into()
+                .unwrap();
+            let port = u16::from_be_bytes(port_bytes);
+
+            let addr_field = this.project_field_named(&address, "sin6_addr")?;
+            let s_addr_field = this
+                .project_field_named(&addr_field, "s6_addr")?
+                .transmute(this.machine.layouts.u128, this)?;
+            // Read bytes and treat them as big endian since address is stored in network byte order.
+            let addr_bytes: [u8; 16] = this
+                .read_bytes_ptr_strip_provenance(s_addr_field.ptr(), Size::from_bytes(16))?
+                .try_into()
+                .unwrap();
+            let addr_bits = u128::from_be_bytes(addr_bytes);
+
+            let flowinfo_field = this.project_field_named(&address, "sin6_flowinfo")?;
+            // flowinfo doesn't get the big endian treatment as this field is stored in native byte order
+            // and not in network byte order.
+            let flowinfo = this.read_scalar(&flowinfo_field)?.to_u32()?;
+
+            let scope_id_field = this.project_field_named(&address, "sin6_scope_id")?;
+            // scope_id doesn't get the big endian treatment as this field is stored in native byte order
+            // and not in network byte order.
+            let scope_id = this.read_scalar(&scope_id_field)?.to_u32()?;
+
+            SocketAddr::V6(SocketAddrV6::new(
+                Ipv6Addr::from_bits(addr_bits),
+                port,
+                flowinfo,
+                scope_id,
+            ))
+        } else {
+            // Socket of other types shouldn't be created in a first place and
+            // thus also no address family of another type should be supported.
+            throw_unsup_format!(
+                "{foreign_name}: address family {family:#x} is unsupported, \
+                                only AF_INET and AF_INET6 are allowed"
+            );
+        };
+
+        interp_ok(Ok(socket_addr))
     }
 }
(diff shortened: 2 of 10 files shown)


src/shims/unix/socket.rs
```diff
@@ -29,6 +32,19 @@ enum SocketKind {
     TcpStream(TcpStream),
 }
 
+#[derive(Debug)]
+enum SocketState {
+    /// No syscall after `socket` has been made.
+    Initial,
+    /// The `bind` syscall has been called on the socket.
+    /// This is only reachable from the [`SocketState::Initial`] state.
+    Bind(SocketAddr),
```
RalfJung: To make this more of a self-explaining state name, maybe `Bound`?

src/shims/unix/socket.rs
```diff
@@ -29,6 +32,19 @@ enum SocketKind {
...
+    /// The `listen` syscall has been called on the socket.
+    /// This is only reachable from the [`SocketState::Bind`] state.
+    #[allow(unused)]
+    TcpListener(TcpListener),
```
RalfJung: `Listening`?

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        address: &OpTy<'tcx>,
+        address_len: &OpTy<'tcx>,
+    ) -> InterpResult<'tcx, Scalar> {
+        let this = self.eval_context_mut();
+
+        let socket = this.read_scalar(socket)?.to_i32()?;
+        let address = match socket_address(address, address_len, "bind", this)? {
+            Ok(addr) => addr,
+            Err(e) => return this.set_last_error_and_return_i32(e),
+        };
```
RalfJung: This can use `try_unwrap_io_result`.

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+/// Attempt to turn an address and length operand into a standard library socket address.
+///
+/// Returns an IO error should the address length not match the address family length.
+fn socket_address<'tcx>(
+    address: &OpTy<'tcx>,
+    address_len: &OpTy<'tcx>,
+    foreign_name: &'static str,
+    this: &mut InterpCx<'tcx, MiriMachine<'tcx>>,
+) -> InterpResult<'tcx, Result<SocketAddr, IoError>> {
```
RalfJung: Instead of making this a free function with an `InterpCx` parameter, please add a private extension trait so it can be called like `this.foo`.

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let address = match socket_address(address, address_len, "bind", this)? {
+            Ok(addr) => addr,
+            Err(e) => return this.set_last_error_and_return_i32(e),
+        };
+
+        // Reject if isolation is enabled
+        if let IsolatedOp::Reject(reject_with) = this.machine.isolated_op {
+            this.reject_in_isolation("`bind`", reject_with)?;
+            return this.set_last_error_and_return_i32(LibcError("EACCES"));
+        }
```
RalfJung: This should not be needed; instead, after checking the socket type, *assert* that we are not in isolation.

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            SocketState::Initial => {
+                let address_family = match &address {
+                    SocketAddr::V4(_) => SocketFamily::IPv4,
+                    SocketAddr::V6(_) => SocketFamily::IPv6,
+                };
+
+                // The Dragonfly man pages don't mention an error for mismatched
+                // socket and address families.
+                if socket.family == address_family
+                    || matches!(this.tcx.sess.target.os, Os::Dragonfly)
```
RalfJung: Let's ignore unsupported OSes like Dragonfly

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                {
+                    *state = SocketState::Bind(address);
+                } else {
+                    // Attempted to bind an address from a family that doesn't match
+                    // the family of the socket.
+                    let err = if matches!(this.tcx.sess.target.os, Os::Linux | Os::Android) {
+                        LibcError("EINVAL")
+                    } else {
+                        LibcError("EAFNOSUPPORT")
+                    };
```
RalfJung: This would be easier to read as
```rust
if socket.family != address_family {
  ...
  return ...;
}
```

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    || matches!(this.tcx.sess.target.os, Os::Dragonfly)
+                {
+                    *state = SocketState::Bind(address);
+                } else {
+                    // Attempted to bind an address from a family that doesn't match
+                    // the family of the socket.
+                    let err = if matches!(this.tcx.sess.target.os, Os::Linux | Os::Android) {
+                        LibcError("EINVAL")
+                    } else {
+                        LibcError("EAFNOSUPPORT")
```
RalfJung, line 210: Please add comments explaining the error code choices (ideally with links)

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    };
+                    return this.set_last_error_and_return_i32(err);
+                }
+            }
+            SocketState::Bind(_) | SocketState::TcpListener(_) => {
+                // Linux and OpenBSD explicitly disallow binding an already bound socket
+                // to another address. For the other targets it's protocol dependent and
+                // thus we error.
+                if matches!(this.tcx.sess.target.os, Os::Linux | Os::Android | Os::OpenBsd) {
```
RalfJung: This doesn't seem worth an OS match; let's just always throw_unsup

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    foreign_name: &'static str,
+    this: &mut InterpCx<'tcx, MiriMachine<'tcx>>,
+) -> InterpResult<'tcx, Result<SocketAddr, IoError>> {
+    let socklen_layout = this.libc_ty_layout("socklen_t");
+    let address_len = this.read_scalar(address_len)?.to_int(socklen_layout.size)?;
+    // We only support socklen_t sizes which can be converted to u64 since the layout
+    // byte sizes are also at most an u64.
+    let Ok(address_len) = u64::try_from(address_len) else {
+        throw_unsup_format!("{foreign_name}: socket address length is not valid u64");
+    };
```
RalfJung: It's fine to just try_into().unwrap() this.

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    let sockaddr_layout = this.libc_ty_layout("sockaddr");
+    if address_len < sockaddr_layout.size.bytes() {
+        // Address length should be at least as big as the generic sockaddr
+        return interp_ok(Err(LibcError("EINVAL")));
+    }
+    let address = this.deref_pointer_as(address, sockaddr_layout)?;
+
+    let family_field = this.project_field_named(&address, "sa_family")?;
+    let family_layout = this.libc_ty_layout("sa_family_t");
+    let family = this.read_scalar(&family_field)?.to_int(family_layout.size)?;
```
RalfJung: Another option would be this -- slightly shorter; both are fine.
```suggestion
    let family = this.read_scalar(&family_field)?.to_int(family_field.layout.size)?;
```

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+    // Depending on the family, decide whether it's IPv4 or IPv6 and use specialized layout
+    // to extract address and port.
+    let socket_addr = if family == this.eval_libc_i32("AF_INET").into() {
+        let sockaddr_in_layout = this.libc_ty_layout("sockaddr_in");
+        if address_len != sockaddr_in_layout.size.bytes() {
+            // Address length should be exactly the length of an IPv4 address
+            return interp_ok(Err(LibcError("EINVAL")));
+        }
+        let address = address.offset(Size::ZERO, sockaddr_in_layout, this)?;
```
RalfJung: [`address.transmute(...)`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_const_eval/interpret/trait.Projectable.html#method.transmute) should work

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        let port_field = this.project_field_named(&address, "sin_port")?;
+        let port = this.read_scalar(&port_field)?.to_u16()?;
+
+        let addr_field = this.project_field_named(&address, "sin_addr")?;
+        let s_addr_field = this.project_field_named(&addr_field, "s_addr")?;
+        let addr_bits = this.read_scalar(&s_addr_field)?.to_u32()?;
+
+        SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::from_bits(addr_bits.to_be()), port.to_be()))
```
RalfJung: What's going on with `to_be`? These are native APIs, they all want native endianess.

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    let socket_addr = if family == this.eval_libc_i32("AF_INET").into() {
+        let sockaddr_in_layout = this.libc_ty_layout("sockaddr_in");
+        if address_len != sockaddr_in_layout.size.bytes() {
+            // Address length should be exactly the length of an IPv4 address
+            return interp_ok(Err(LibcError("EINVAL")));
+        }
+        let address = address.offset(Size::ZERO, sockaddr_in_layout, this)?;
+
+        let port_field = this.project_field_named(&address, "sin_port")?;
+        let port = this.read_scalar(&port_field)?.to_u16()?;
```
RalfJung: The endianess conversion should happen here since it is about getting the data out of the machine, not about using the `SocketAddrV4` API.

Probably the best way of doing this is to load this as an array of N bytes and then use `from_be` on the host.

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        let addr_field = this.project_field_named(&address, "sin6_addr")?;
+        let s_addr_field = this.project_field_named(&addr_field, "s6_addr")?.offset(
+            Size::ZERO,
+            this.machine.layouts.u128,
+            this,
+        )?;
+        let addr_bits = this.read_scalar(&s_addr_field)?.to_u128()?;
+
+        let flowinfo_field = this.project_field_named(&address, "sin6_flowinfo")?;
```
RalfJung: Needs a comment saying why flowinfo and scope_id don't get the endianess treatment

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            128
+        };
+
+        // Only allow same backlog values as the standard library uses since the standard library
+        // doesn't provide a way to use custom values.
+        if backlog != allowed_backlog {
+            throw_unsup_format!(
+                "listen: backlog {backlog} is unsupported, only {allowed_backlog} is allowed"
+            )
+        }
```
RalfJung: This does not seem worth a hard error, let's make this a warning. See here for an example warning:

https://github.com/rust-lang/miri/blob/2ea5cf4e8033e96debce3927ed045884a7933455/src/alloc_addresses/mod.rs#L368-L372

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        let mut state = socket.state.borrow_mut();
+
+        match &*state {
+            SocketState::Bind(socket_addr) =>
+                match TcpListener::bind(socket_addr) {
+                    Ok(listener) => {
+                        *state = SocketState::TcpListener(listener);
```
RalfJung: Another use for the try_unwrap_io_result helper?

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        let socket = this.read_scalar(socket)?.to_i32()?;
+        let level = this.read_scalar(level)?.to_i32()?;
+        let option_name = this.read_scalar(option_name)?.to_i32()?;
+
+        // Reject if isolation is enabled
+        if let IsolatedOp::Reject(reject_with) = this.machine.isolated_op {
+            this.reject_in_isolation("`setsockopt`", reject_with)?;
+            return this.set_last_error_and_return_i32(LibcError("EACCES"));
+        }
```
RalfJung: This function is entirely interpreter-local so we don't care about isolation here.

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                Os::MacOs | Os::FreeBsd | Os::NetBsd | Os::Dragonfly
+            ) {
+                available_options.push("SO_NOSIGPIPE");
+                // SO_NOSIGPIPE only exists on MacOS, FreeBSD, NetBSD and Dragonfly.
+                let opt_so_nosigpipe = this.eval_libc_i32("SO_NOSIGPIPE");
+
+                if option_name == opt_so_nosigpipe {
+                    let option_value =
+                        this.deref_pointer_as(option_value, this.machine.layouts.i32)?;
+                    let flag_set = this.read_scalar(&option_value)?.to_u32()? == 1;
```
RalfJung: So `2` is `false`? That seems odd.

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                if matches!(*state, SocketState::Initial) {
+                    // On non-windows targets this is set to 1 before calling `bind` in the standard library implementation.
+                    // Since the standard library doesn't provide a way of setting this flag, we just ignore it.
+                    if flag_set {
+                        socket.has_reuse_addr.set(true);
+                    } else {
+                        throw_unsup_format!(
+                            "setsockopt: option SO_REUSEADDR on level SOL_SOCKET can only be set to 1"
+                        )
+                    }
```
RalfJung: I don't think this makes sense; instead, bind or listen (please figure out which) should check that the value has been set.

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,355 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        let state = socket.state.borrow();
+
+        // Set options on socket level, needed for
+        // - [`TcpListener::bind`]: SO_REUSEADDR
+        // - [`TcpListener::bind`]: SO_NOSIGPIPE no MacOS, FreeBSD, NetBSD and Dragonfly
+        if level == this.eval_libc_i32("SOL_SOCKET") {
+            // A list of available options which can later be displayed in the error should
+            // `option_name` be an invalid socket option.
+            let mut available_options = vec!["SO_REUSEADDR"];
```
RalfJung: I don't think this is worth it.

src/shims/unix/socket.rs
WhySoBad: `transmute` requires that the current and the new layout have the same sizes. This only works in the IPv4 case but in the IPv6 case the `sockaddr_in6` struct is bigger than the `sockaddr` struct. 

I now use `transmute` for the IPv4 case and leave it as-is for the IPv6 case with a comment.

src/shims/unix/socket.rs
WhySoBad: `try_unwrap_io_result` cannot be used here since this is only applicable to `T: From<i32>`

src/shims/unix/socket.rs
WhySoBad: `try_unwrap_io` cannot be used here either as we also don't have `std::net::TcpListener: From<i32>`

bjorn3:
> For the bind shim I'm a bit uncertain about the validation we should perform there. Since we effectively only execute this syscall when seeing a listen later on

A bind can also precede a connect. All bind does is set the local ip address to use. For listen this would be the ip on which you listen, while for connect that would be the ip from which you send packets to the remote. Also with UDP I think you can freely change the bind later on (or omit it entirely if recvfrom/recvmsg/sendto/sendmsg is used), but you can definitively change the connect later on.

src/shims/unix/socket.rs
RalfJung: Ah, right... we could consider adding an internal function that returns `io::Result<i32>`, or adding a macro that basically acts like `?` but with `set_last_error_and_return_i32`... but let's just open-code this for now and see how often the pattern arises.

src/shims/unix/socket.rs
RalfJung: Sounds good.

src/shims/unix/socket.rs
WhySoBad: Why should this function be entirely interpreter-local? A program can always call `libc::setsockopt`. I think it should receive the same isolation assertion as `bind` and `listen`.

WhySoBad:
> A bind can also precede a connect.

Yes, but we can't support this since we can only represent a `bind` followed by a `listen` using the standard library `TcpListener` abstraction. The `TcpStream` abstraction has no way of calling `bind` and thus we need to reject all cases where a `bind` is followed by a `connect`.  

> Also with UDP I think you can freely change the bind later on

At the moment we don't support UDP at all -- I probably should've mentioned in the description that this PR is just about TCP sockets. But I also don't think we could represent multiple `bind`s with `UdpSocket` as it's constructed using a single `bind` [in the standard library](https://github.com/rust-lang/rust/blob/c043085801b7a884054add21a94882216df5971c/library/std/src/sys/net/connection/socket/mod.rs#L669). However, multiple `connect`s would be allowed for `UdpSocket`s.

src/shims/unix/socket.rs
WhySoBad: Yes, I think a macro would probably be a nice solution in the future. I suspect that this pattern will arise at least a few additional times.

src/shims/unix/socket.rs
WhySoBad: I'll leave it as-is but good to know that one could also use the operand's layout.

src/shims/unix/socket.rs
RalfJung: It is interpreter-local in the sense that is has no effect on and no dependency on the outside world. That is what isolation means: behavior is fully deterministic and side-effect-free.

WhySoBad: Ah okay, I see. Yes, at the moment it's interpreter local but in the future it will get side-effects as some methods on `TcpStream` and `TcpListener` are just wrappers for `setsockopt`. 
I removed the isolation assertion for now and will add it back in when those methods get implemented.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -17,3 +23,123 @@ fn test_socket_close() {
...
+        errno_check(libc::bind(
+            sockfd,
+            (&addr as *const libc::sockaddr_in).cast::<libc::sockaddr>(),
+            size_of::<libc::sockaddr_in>() as libc::socklen_t,
+        ));
+    }
+}
+
+#[cfg(any(target_os = "macos", target_os = "dragonfly"))]
+fn test_bind_ipv4_nosigpipe() {
```
RalfJung: This is... an exact copy of `test_bind_ipv4`, isn't it? Just with a slightly different route towards calling `nosigpipe`.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -17,3 +23,123 @@ fn test_socket_close() {
...
     }
 }
+
+fn test_bind_ipv4() {
+    let sockfd =
+        unsafe { errno_result(libc::socket(libc::AF_INET, libc::SOCK_STREAM, 0)).unwrap() };
+    let addr = net::ipv4_sock_addr("127.0.0.1:1234").unwrap();
+    #[cfg(any(target_os = "macos", target_os = "freebsd", target_os = "netbsd"))]
+    try_set_nosigpipe(sockfd);
+    set_reuseaddr(sockfd);
```
RalfJung: The basic test should not set these flags IMO. Having a separate test that sets them is a good idea, but needs a comment.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -17,3 +23,123 @@ fn test_socket_close() {
...
+    unsafe {
+        errno_check(libc::bind(
+            sockfd,
+            (&addr as *const libc::sockaddr_in).cast::<libc::sockaddr>(),
+            size_of::<libc::sockaddr_in>() as libc::socklen_t,
+        ));
+    }
+}
+
+fn test_bind_ipv4_invalid_addr_len() {
```
RalfJung, line 120: All non-obvious tests should have comments explaining what they are testing.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -17,3 +23,123 @@ fn test_socket_close() {
...
+    let sockfd =
+        unsafe { errno_result(libc::socket(libc::AF_INET, libc::SOCK_STREAM, 0)).unwrap() };
+    let addr = net::ipv4_sock_addr("127.0.0.1:1234").unwrap();
+    try_set_nosigpipe(sockfd);
+    set_reuseaddr(sockfd);
+    let err = unsafe {
+        errno_result(libc::bind(
+            sockfd,
+            (&addr as *const libc::sockaddr_in).cast::<libc::sockaddr>(),
+            (size_of::<libc::sockaddr_in>() + 1) as libc::socklen_t,
```
RalfJung, line 129: I guess the "+1" here is the key part? Please point that out.

tests/pass-dep/shims/socket.rs
RalfJung, line 1: This is not a libc test. It doesn't use libc. So please move it to `tests/pass/shims`.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -17,3 +23,123 @@ fn test_socket_close() {
...
+            (&addr as *const libc::sockaddr_in).cast::<libc::sockaddr>(),
+            size_of::<libc::sockaddr_in>() as libc::socklen_t,
+        ));
+    }
+}
+
+fn test_bind_ipv4_invalid_addr_len() {
+    let sockfd =
+        unsafe { errno_result(libc::socket(libc::AF_INET, libc::SOCK_STREAM, 0)).unwrap() };
+    let addr = net::ipv4_sock_addr("127.0.0.1:1234").unwrap();
```
RalfJung: I don't think we need a helper for these, do we? `[127, 0, 0, 1]` seems understandable on its own.

Helpers are a last resort for when directly using libc functions is extremely awkward. The point of these tests is to test the libc API surface so we should stay as close to that as possible.

tests/fail-dep/libc/socket-reuseaddr-after-bind.stderr
```diff
@@ -0,0 +1,19 @@
+error: unsupported operation: setsockopt: option SO_REUSEADDR on level SOL_SOCKET can only be set before calling `bind` on a socket
```
RalfJung: This limitation seems unnecessary, given that we don't actually do anything until `listen` anyway?

After `listen` is might be too late, but what's the issue with doing it between bind and listen?

src/shims/unix/socket.rs
```diff
@@ -148,10 +181,350 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        if matches!(this.tcx.sess.target.os, Os::MacOs | Os::FreeBsd | Os::NetBsd)
+            && !socket.has_no_sig_pipe.get()
+        {
+            // The standard library implementation sets this flag immediately after calling
+            // the `socket` syscall on MacOS, FreeBSD, and NetBSD targets
+            throw_unsup_format!(
+                "listen: listening without SO_NOSIGPIPE set is unsupported on MacOS, \
+                                FreeBSD, and NetBSD targets"
+            )
```
RalfJung: Seeing how awkward this gets in the tests, I don't think we should do this. Miri doesn't handle signals anyway...

src/shims/unix/socket.rs
```diff
@@ -148,10 +181,350 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            return interp_ok(this.eval_libc("ENOTSOCK"));
+        };
+
+        assert!(this.machine.communicate(), "cannot have `Socket` with isolation enabled!");
+
+        if !socket.has_reuse_addr.get() {
+            // The standard library implementation sets this flag before binding
+            // for all non-windows targets.
+            throw_unsup_format!("bind: binding without SO_REUSEADDR set is unsupported")
+        }
```
RalfJung: Given that this is mostly just a hint to prevent the OS from adding an arbitrary timeout, I don't think we need to force the user to set this flag.

src/shims/unix/socket.rs
```diff
@@ -38,11 +42,11 @@ enum SocketState {
     Initial,
     /// The `bind` syscall has been called on the socket.
     /// This is only reachable from the [`SocketState::Initial`] state.
-    Bind(SocketAddr),
+    Bound(SocketAddr),
     /// The `listen` syscall has been called on the socket.
-    /// This is only reachable from the [`SocketState::Bind`] state.
+    /// This is only reachable from the [`SocketState::Bound`] state.
     #[allow(unused)]
```
RalfJung: Is this still unused?

src/shims/unix/socket.rs
```diff
@@ -436,92 +413,118 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                .try_into()
+                .unwrap();
+            let port = u16::from_be_bytes(port_bytes);
+
+            let addr_field = this.project_field_named(&address, "sin_addr")?;
+            let s_addr_field = this.project_field_named(&addr_field, "s_addr")?;
+            // Read bytes and treat them as big endian since address is stored in network byte order.
+            let addr_bytes = this
+                .read_bytes_ptr_strip_provenance(s_addr_field.ptr(), Size::from_bytes(4))?
+                .try_into()
```
RalfJung: Please add a type annotation to `let addr_bytes` to that the `try_into` here isn't quite as non-local. (Same below)

src/diagnostics.rs
```diff
@@ -657,6 +661,8 @@ impl<'tcx> MiriMachine<'tcx> {
             | WeakMemoryOutdatedLoad { .. } =>
                 ("tracking was triggered here".to_string(), DiagLevel::Note),
             FileInProcOpened => ("open a file in `/proc`".to_string(), DiagLevel::Warning),
+            SocketListenInvalidBacklog { .. } =>
+                ("call to `listen` with invalid backlog value".to_string(), DiagLevel::Warning),
```
```suggestion
                ("call to `listen` with unsupported backlog value".to_string(), DiagLevel::Warning),
```

src/diagnostics.rs
```diff
@@ -704,13 +710,18 @@ impl<'tcx> MiriMachine<'tcx> {
                 };
                 format!("GenMC currently does not model the failure ordering for `compare_exchange`. {was_upgraded_msg}. Miri with GenMC might miss bugs related to this memory access.")
             }
-            FileInProcOpened => format!("files in `/proc` can bypass the Abstract Machine and might not work properly in Miri")
+            FileInProcOpened => format!("files in `/proc` can bypass the Abstract Machine and might not work properly in Miri"),
+            SocketListenInvalidBacklog { actual, allowed, .. } => format!("called `listen` on socket with backlog value of {actual} but only {allowed} is allowed")
```
```suggestion
            SocketListenInvalidBacklog { actual, allowed, .. } => format!("called `listen` on socket with backlog value of {actual} but only {allowed} is supported"),
```

src/diagnostics.rs
```diff
@@ -704,13 +710,18 @@ impl<'tcx> MiriMachine<'tcx> {
...
         };
 
         let notes = match &e {
             ProgressReport { block_count } => {
                 vec![note!("so far, {block_count} basic blocks have been executed")]
             }
+            SocketListenInvalidBacklog { details: true, allowed, .. } =>
+                vec![note!(
+                    "Miri only allows a backlog value of {allowed} since this value is used by the standard library on the current target"
```
```suggestion
                    "the given value will be ignored and a backlog of {actual} will be used instead"
```

src/diagnostics.rs
RalfJung: Please add a test hitting this warning.

src/shims/unix/socket.rs
WhySoBad: Yes, this sounds right. I was a bit confused about the option as I thought it would introduce cases where Miri errors but the standard library wouldn't. After thinking more about it, I realized that it's actually exactly the opposite where Miri won't cause an `EADDRNOTAVAIL` error in cases where the OS _could_. 

Do you think this should also get a warning or should we just silently accept the missing flag?

src/shims/unix/socket.rs
WhySoBad: Okay. I also thought about that Miri could emulate the signal. Isn't the [std::io::ErrorKind::ConnectionAborted](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.ConnectionAborted) exactly for this case? Once Miri supports signals, it could match for this error kind and turn this into a `SIGPIPE` signal if the flag shouldn't be set.

Also, the same question as below, should this be turned into a warning or should we just silently ignore it?

src/shims/unix/socket.rs
RalfJung: No need to emit a warning.

src/shims/unix/socket.rs
RalfJung: > Once Miri supports signals, it could match for this error kind and turn this into a SIGPIPE signal if the flag shouldn't be set.

I'd say we worry about that if we ever support signals.^^
No reason to emit a warning for now.

src/shims/unix/socket.rs
WhySoBad: Yes, since we never use the listener.

tests/fail-dep/libc/socket-reuseaddr-after-bind.stderr
WhySoBad: Yes, you're right. Given that we ignore the value of this option anyways, it doesn't matter whether the user sets it before or after a `bind`.
For the same reason, I've now removed the state-checking logic when attempting to set this socket option.

tests/pass-dep/libc/libc-socket.rs
WhySoBad: I'd argue that it's kind of awkward when we need to construct structs like the following without some sort of helper instead of just being able to write `[::1]:1234`.
```rust
pub struct sockaddr_in6 {
    pub sin6_family: libc::sa_family_t,
    pub sin6_port: libc::in_port_t,
    pub sin6_flowinfo: u32,
    pub sin6_addr: libc::in6_addr,
    pub sin6_scope_id: u32,
}
```

src/diagnostics.rs
WhySoBad: `{allowed}` shouldn't be replaced by `{actual}` here

RalfJung: Wait, "actual" isn't the actual value being used?

tests/pass-dep/libc/libc-socket.rs
RalfJung: What does the code for that look like?

I think going through the standard library functions adds too many layers here. The entire point of these tests is to check the APIs directly, not via the standard library.

src/diagnostics.rs
WhySoBad: No, `actual` is the unsupported value provided by the user whilst `allowed` (now called `supported`) is the value supported by Miri. Since the point of the warning is to tell the user that Miri won't use their provided value, we print the value of `allowed`/`supported` to inform them which value is used instead. 

I can see that `actual` is a badly chosen name. I'll rename it to `provided` to avoid future confusion.

tests/pass-dep/libc/libc-socket.rs
WhySoBad: That sounds fair. I've now simplified the helpers to directly create the `sockaddr_in6` and `sockaddr_in` respectively without relying on the standard library abstractions.

tests/utils/libc.rs
```diff
@@ -157,3 +157,49 @@ pub mod epoll {
...
+    /// IPv6 localhost address bytes
+    pub const IPV6_LOCALHOST: [u8; 16] = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1];
+
+    /// Create a libc representation of an IPv4 address given the address bytes and a port.
+    pub fn ipv4_sock_addr(addr_bytes: [u8; 4], port: u16) -> libc::sockaddr_in {
+        libc::sockaddr_in {
+            sin_family: libc::AF_INET as libc::sa_family_t,
+            sin_port: port.to_be(),
+            // The native endian conversion method is used so that endianness is not swapped
+            // since the `addr_bytes` already stored the address in big endian.
```
RalfJung:
```suggestion
            // We want to transmute `addr_bytes` to a `u32`: `addr_bytes` is already in big-endian
            // and that's the format `sin_addr` expects. `from_ne_bytes` is a safe function
            // doing exactly this transmute.
```
It took me a bit to understand the comment here, so let me try to reword. I honestly wonder if it'd not be less confusing to just use `transmute`.^^

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -17,3 +33,163 @@ fn test_socket_close() {
...
+    errno_result(unsafe {
+        libc::setsockopt(
+            sockfd,
+            level,
+            option_name,
+            (&raw const option_value) as *const _,
+            option_len,
+        )
+    })?;
+    Ok(())
```
RalfJung, line 193:
```suggestion
    })
```

RalfJung: Ah no that doesn't work, never mind

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -17,3 +33,163 @@ fn test_socket_close() {
...
+
+#[cfg(any(
+    target_os = "macos",
+    target_os = "freebsd",
+    target_os = "netbsd",
+    target_os = "dragonfly"
+))]
+/// Tests binding after the `SO_NOSIGPIPE` socket option has been set on the newly created socket.
+/// This mimics the standard library's [`TcpListener`] implementation on MacOS, FreeBSD,
+/// NetBSD, and Dragonfly targets.
```
```suggestion
/// Tests binding after the `SO_NOSIGPIPE` socket option has been set on the newly created socket.
/// That flag only exists on BSD-like OSes.
```

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -1,14 +1,30 @@
...
+        target_os = "netbsd",
+        target_os = "dragonfly"
+    ))]
+    {
+        test_bind_ipv4_nosigpipe();
+        test_set_nosigpipe_invalid_len();
+    }
+    test_bind_ipv4_invalid_addr_len();
+    test_bind_ipv6();
+    test_set_reuseaddr_invalid_len();
```
RalfJung: Please keep the order in `main` consistent with the order of definitions below. Currently it looks kind of random.

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,310 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let opt_so_nosigpipe = this.eval_libc_i32("SO_NOSIGPIPE");
+
+                if option_name == opt_so_nosigpipe {
+                    if option_len != 4 {
+                        // Option value should be C-int which is usually 4 bytes.
+                        return this.set_last_error_and_return_i32(LibcError("EINVAL"));
+                    }
+                    let option_value =
+                        this.deref_pointer_as(option_value, this.machine.layouts.i32)?;
+                    this.read_scalar(&option_value)?.to_i32()?;
```
```suggestion
                    let _val = this.read_scalar(&option_value)?.to_i32()?;
                    // We entirely ignore this value since we do not support signals anyway.
```

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,310 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                }
+            }
+
+            if option_name == opt_so_reuseaddr {
+                if option_len != 4 {
+                    // Option value should be C-int which is usually 4 bytes.
+                    return this.set_last_error_and_return_i32(LibcError("EINVAL"));
+                }
+                let option_value = this.deref_pointer_as(option_value, this.machine.layouts.i32)?;
+                this.read_scalar(&option_value)?.to_i32()?;
```
```suggestion
                let _val = this.read_scalar(&option_value)?.to_i32()?;
                // We entirely ignore this: std always sets REUSEADDR for us, and in the end it's more of a
                // hint to bypass some arbitrary timeout anyway.
```

src/shims/unix/socket.rs
```diff
@@ -148,10 +173,310 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            } else {
```
RalfJung, line 337: It's confusing to use `else` here given that not all options are part of the if-else chain. You are relying on the logic for every option to early-`return`. So please rely on it all the way and don't use `else`.
```suggestion
            }
            
```

tests/utils/libc.rs
WhySoBad: True, I think it's clearer when we manually do the transmute.

src/shims/unix/socket.rs
```diff
@@ -29,15 +36,30 @@ enum SocketKind {
...
+#[derive(Debug)]
+enum SocketState {
+    /// No syscall after `socket` has been made.
+    Initial,
+    /// The `bind` syscall has been called on the socket.
+    /// This is only reachable from the [`SocketState::Initial`] state.
+    Bound(SocketAddr),
+    /// The `listen` syscall has been called on the socket.
+    /// This is only reachable from the [`SocketState::Bound`] state.
+    #[allow(unused)]
```
```suggestion
    #[expect(unused)]
```

src/shims/unix/socket.rs
```diff
@@ -29,15 +36,30 @@ enum SocketKind {
...
+
 #[derive(Debug)]
 struct Socket {
     /// Family of the socket, used to ensure socket only binds/connects to address of
     /// same family.
     family: SocketFamily,
     /// Type of the socket, either datagram or stream.
     /// Only stream is supported at the moment!
+    #[allow(unused)]
     socket_type: SocketType,
```
RalfJung: If this enum has only one value and we ignore the field... then why do field and enum even exist?

tests/utils/libc.rs
RalfJung: Okay, let's do that. Still with a short comment though:
```
// `addr_bytes` is in big-endian and that's the format `sin_addr` expects.
```

WhySoBad:
> You didn't ask for a review again but here are some comments anyway.

I wanted to first wait for the CI to complete. But thanks for the review anyways :) 

@rustbot ready

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -98,24 +88,6 @@ fn test_bind_ipv4_nosigpipe() {
...
-fn test_set_nosigpipe_invalid_len() {
-    let sockfd =
-        unsafe { errno_result(libc::socket(libc::AF_INET, libc::SOCK_STREAM, 0)).unwrap() };
-    // Value should be of type `libc::c_int` which has size 4 bytes.
-    // By providing a u64 of size 8 bytes we trigger an invalid length error.
-    let err = setsockopt(sockfd, libc::SOL_SOCKET, libc::SO_NOSIGPIPE, 1u64).unwrap_err();
-    assert_eq!(err.kind(), ErrorKind::InvalidInput);
-    // check that it is the right kind of `InvalidInput`
-    assert_eq!(err.raw_os_error(), Some(libc::EINVAL));
-}
```
RalfJung: Why did you reorder the functions? They were nicely grouped. Just the order in `main` was strange.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -190,6 +192,6 @@ fn setsockopt<T>(
             (&raw const option_value) as *const _,
             option_len,
         )
-    })?;
-    Ok(())
+    })
+    .map(|_| ())
```
RalfJung: Please go back to `?` for this -- using a higher-order function here just obscures what happens IMO.

tests/utils/libc.rs
```diff
@@ -169,9 +169,11 @@ pub mod net {
         libc::sockaddr_in {
             sin_family: libc::AF_INET as libc::sa_family_t,
             sin_port: port.to_be(),
-            // The native endian conversion method is used so that endianness is not swapped
-            // since the `addr_bytes` already stored the address in big endian.
-            sin_addr: libc::in_addr { s_addr: u32::from_ne_bytes(addr_bytes) },
+            // `addr_bytes` is already in big-endian and that's the format `sin_addr` expects.
+            #[allow(unnecessary_transmutes)]
```
RalfJung: Also use `expect` here

tests/pass-dep/libc/libc-socket.rs
WhySoBad: I thought having all IPv4 bind tests, then all IPv6 bind tests and then all setsockopt tests would be a better order. 

Otherwise we have some setsockopt tests between bind tests which seemed a bit off.

We kind of still have setsockopt tests between the bind tests but in my opinion they belong to bind as we're testing binding with these options set. 
Given that we'll have more setsockopt tests for the other options which will be supported, I think the current order is better.

RalfJung: Whatever you do, please keep the socketopt tests together. Currently, they are not. test_bind_ipv4_nosigpipe and test_set_nosigpipe_invalid_len should remain next to each other. (No idea why one has ipv4 in the name and the other does not...)

RalfJung:

@rustbot author

Issue #3100, opened 2023-10-02
Add support for setting dynamically available target features


RalfJung:
Currently one can tell Miri that, e.g., AVX is available by setting `-C target-feature="avx"`. However, that corresponds to building a binary that statically requires AVX to be available. For real Rust, there's also the possibility of building code without AVX being globally available, but then at run-time detecting AVX to be available. It would be great if Miri could emulate that.

This will require at least:
- Some way to set which features are dynamically available. Probably something like `-Zmiri-target-feature`.
- Making sure that `is_target_feature_enabled!` respects this dynamic set.

The test added in https://github.com/rust-lang/miri/pull/3099 can then be made more subtle by enabling AVX dynamically, at which point the currently reported UB should disappear -- but we should still see UB due to the [ABI mismatch](https://github.com/rust-lang/miri/issues/3095)!

PR #4885, opened 2026-03-04
explain how to control target features with Miri

RalfJung:
Trigger by a [Zulip thread](https://rust-lang.zulipchat.com/#narrow/channel/269128-miri/topic/miri.20and.20CPU.20target.20features).

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,16 @@ platform. For example `cargo miri test --target s390x-unknown-linux-gnu`
 will run your test suite on a big-endian target, which is useful for testing
 endian-sensitive code.
 
+### Controlling target features
+
+Controlling target features works similar to regular rustc invocations:
+`RUSTFLAGS="-Ctarget-features=+avx512f" cargo miri test` runs the tests with AVX512 enabled. (Miri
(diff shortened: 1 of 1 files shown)


Shnatsel:
Does `-Ctarget-cpu=` also work?

Please also mention that I can set `MIRIFLAGS` to tell miri to emulate those features but not force the code to be compiled with them. This is very useful for testing code with runtime CPU feature detection, which all practical x86 SIMD code is forced to do.

RalfJung:
> Does -Ctarget-cpu= also work?

Yeah it should work just like in rustc, we don't even have any of our own code there. We just get the list of available target features from rustc once it is done computing that. I'll mention that in the docs.

> Please also mention that I can set MIRIFLAGS to tell miri to emulate those features but not force the code to be compiled with them

Turns out I spoke too quickly on Zulip, the situation is a bit tricky...  the binary crate actually still gets built with those features if you put them in MIRIFLAGS, just dependencies don't. Setting target features just at runtime is currently not implemented. That's tracked in https://github.com/rust-lang/miri/issues/3100.

Shnatsel (approved):

RalfJung:
Thanks for taking a look!

Issue #4879, opened 2026-02-28
Post-monomorphization error may be triggered by any thread local key usage when using `-Zmiri-native-lib` flag


wuwbobo2021:
This is originally discovered by @rib in <https://github.com/jni-rs/jni-rs/issues/771#issuecomment-3977500816>.

### Simple test case

```toml
[package]
name = "miri-test"
version = "0.1.0"
edition = "2024"
```

```rust
use std::cell::RefCell;

#[derive(Debug)]
struct SomeStruct {}

thread_local! {
    static KEY: RefCell<SomeStruct> =  SomeStruct {}.into()
}

fn main() {
    KEY.with(|_f| {
        println!("Hello, world!");
    });
}
```

Run:

```sh
MIRIFLAGS="-Zmiri-native-lib=/home/wuwbobo/tmp/miri-test/mylib/target/debug/libmylib.so -Zmiri-backtrace=full" cargo +nightly miri run
```

(In which `libmylib.so` is an empty dummy `cdylib`, this can be changed to any existing dynamic library)

Error output: 

```
error: post-monomorphization error: the type `Self` does not have a fixed layout
  --> src/main.rs:11:5
   |
11 |     KEY.with(|_f| {
   |     ^^^ post-monomorphization error occurred here
   |
   = note: stack backtrace:
           0: main
               at src/main.rs:11:5: 11:8
           1: <fn() as std::ops::FnOnce<()>>::call_once - shim(fn())
               at /home/wuwbobo/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5: 250:71
           2: std::sys::backtrace::__rust_begin_short_backtrace::<fn(), ()>
               at /home/wuwbobo/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/sys/backtrace.rs:166:18: 166:21
           3: std::rt::lang_start::<()>::{closure#0}
               at /home/wuwbobo/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/rt.rs:206:18: 206:75
           4: std::ops::function::impls::<impl std::ops::FnOnce<()> for &dyn std::ops::Fn() -> i32 + std::marker::Sync + std::panic::RefUnwindSafe>::call_once
[...]

PR #4891, opened 2026-03-08
native calls: more gracefully handle function pointers with unsupported types

RalfJung:
This is an attempt to fix https://github.com/rust-lang/miri/issues/4879. When we encounter a function pointer whose argument/return type we cannot turn into a libffi type, we now construct a libffi closure without out setting argument/return types (which I assume is equivalent to `fn() -> ()`. Since the closure doesn't actually access its arguments and never returns, I hope that's okay? I don't know what actually happens when libffi constructs such a closure; does it actually run any code that assumes the argument/return types are correct before invoking our own callback? If yes, then what I do here is probably UB so we'd have to find another approach...

Cc @nia-e @bjorn3


nia-e:
AFAIK, libffi only manipulates the arguments as opaque bytes and so this should be fine

PR #4892, opened 2026-03-09
Handle blocking I/O using Mio, add `accept` and `connect` network socket shims

WhySoBad:
Hi,

This pull request adds a new blocking I/O manager to the Miri machine. This manager can be used to register blocking I/O operations using `block_thread_for_io` and unblock them once an event for this I/O source happened. 

To test the new non-blocking I/O, this pull request also adds support for the `accept` and `connect` TCP socket operations. Miri now supports blocking and non-blocking accepts and connects.


Cargo.lock
```diff
@@ -431,7 +431,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
```
RalfJung: Why is windows-sys getting downgraded?

src/concurrency/blocking_io.rs
```diff
@@ -0,0 +1,209 @@
...
+///
+/// Since blocking host I/O is inherently non-deterministic, no method on this
+/// manager should be called when isolation is enabled. The only exception is
+/// the [`BlockingIoManager::new`] function to create the manager. Everywhere else,
+/// we assert that isolation is disabled!
+pub struct BlockingIoManager {
+    /// Poll instance to monitor I/O events from the OS.
+    /// This is only [`None`] when Miri is run with isolation enabled.
+    poll: Option<Poll>,
+    /// Buffer used to store the ready I/O events when calling [`Poll::poll`].
```
RalfJung, line 33: The comment should clarify that there isn't actually any state here, we just reuse the buffer.

src/concurrency/blocking_io.rs
```diff
@@ -0,0 +1,209 @@
...
+        kind: BlockingIoKind,
+        source: SourceRef,
+        thread: ThreadId,
+        interests: Interest,
+    ) {
+        let poll =
+            self.poll.as_ref().expect("Blocking I/O should not be called with isolation enabled");
+
+        #[allow(clippy::as_conversions)]
+        let token = Token(thread.to_u32() as usize);
```
RalfJung:
```suggestion
        let token = Token(thread.to_u32().to_usize());
```
and remove the `allow`

src/concurrency/blocking_io.rs
```diff
@@ -0,0 +1,209 @@
...
+
+/// Capacity of the event queue which can be polled at a time.
+/// Since we don't expect many simultaneous blocking I/O events
+/// this value can be set rather low.
+const IO_EVENT_CAPACITY: usize = 16;
+
+// Supertrait to enforce that all sources implement [`Source`] as
+// well as [`VisitProvenance`].
+pub trait SourceRefExt: Source + VisitProvenance {}
+impl<T> SourceRefExt for T where T: Source + VisitProvenance {}
```
RalfJung: We currently assume FDs cannot hold provenance. Maybe we should remove that assumption, but then we'll do that globally, not just for blocking... so please remove this here for now.

src/concurrency/blocking_io.rs
```diff
@@ -0,0 +1,209 @@
...
+    /// This is only [`None`] when Miri is run with isolation enabled.
+    poll: Option<Poll>,
+    /// Buffer used to store the ready I/O events when calling [`Poll::poll`].
+    events: Events,
+    /// Map between threads which are currently blocked, the kind of I/O
+    /// they are blocked on and the underlying I/O source.
+    sources: FxHashMap<ThreadId, (BlockingIoKind, SourceRef)>,
+    /// List of threads which are ready to be unblocked together with the I/O kind
+    /// they were blocked on.
+    ready: VecDeque<(ThreadId, BlockingIoKind)>,
```
RalfJung: Seems cleaner if polling just returns which threads are ready now, rather than storing that in global state.

That means we have to do the call-unblock-callbacks-loop after each `poll`; that's a new shared helper in the scheduler that's then called in two places.

src/concurrency/blocking_io.rs
```diff
@@ -0,0 +1,209 @@
...
+
+        // We need to clone the iterator here since it holds an immutable reference to `self.events`.
+        // This doesn't work out since we need a mutable self-reference inside the loop body.
+        let events = self.events.iter().cloned().collect::<Vec<_>>();
+        for event in events {
+            let token = event.token();
+            // It's safe to convert the token identifier back to an u32
+            // since we only create tokens from thread id's which are u32.
+            #[expect(clippy::as_conversions)]
+            let thread = ThreadId::new_unchecked(token.0 as u32);
```
RalfJung: No `as` casts please

src/concurrency/thread.rs
```diff
@@ -27,8 +27,11 @@ enum SchedulingAction {
     ExecuteStep,
     /// Execute a timeout callback.
     ExecuteTimeoutCallback,
-    /// Wait for a bit, until there is a timeout to be called.
-    Sleep(Duration),
+    /// Wait for a bit but at most as long as the duration specified.
+    /// We wake up early if an I/O event happened.
+    /// If the duration is [`None`], we sleep indefinitely. This is
+    /// only allowed when isolation is disabled!
```
```suggestion
    /// only allowed when isolation is disabled and there are threads waiting for I/O!
```

src/concurrency/thread.rs
```diff
@@ -27,8 +27,11 @@ enum SchedulingAction {
...
+    Sleep(Option<Duration>),
```
RalfJung: If this does more than sleeping we should rename. `SleepAndWaitForIo`?

src/concurrency/thread.rs
RalfJung: At least parts of this comment should be above the IO handling and also cover that.

src/concurrency/thread.rs
```diff
@@ -1304,7 +1345,20 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                 SchedulingAction::Sleep(duration) => {
-                    this.machine.monotonic_clock.sleep(duration);
+                    if this.machine.communicate() {
+                        // When we're running with isolation disabled, instead of
+                        // strictly sleeping the duration we allow waking up
+                        // early for I/O events from the OS.
+                        //
+                        // We ignore the result from the poll as it's just used as a sleep.
+                        // In the scheduler, before scheduling anything, we poll again anyways.
+                        this.machine.blocking_io.poll(duration).ok();
```
RalfJung: We should default to aborting on errors (for all `poll` calls)

src/concurrency/blocking_io.rs
```diff
@@ -0,0 +1,209 @@
...
+
+        assert!(
+            !self.sources.contains_key(&thread),
+            "A thread cannot be registered twice at the same time"
+        );
+
+        // Treat errors from registering as fatal. On UNIX hosts this can only
+        // fail due to system resource errors (e.g. ENOMEM or ENOSPC).
+        poll.registry().register(&mut *source.borrow_mut(), token, interests).unwrap();
+        self.sources.insert(thread, (kind, source));
```
RalfJung:
```suggestion
        self.sources.try_insert(thread, (kind, source)).unwrap();
```
Then you can remove the `assert!` above

src/shims/unix/socket.rs
```diff
@@ -32,6 +36,62 @@ enum SocketState {
     /// The `listen` syscall has been called on the socket.
     /// This is only reachable from the [`SocketState::Bound`] state.
     Listening(TcpListener),
+    /// The `connect` syscall has been called and we weren't yet able
+    /// to ensure the connection is established. This is only reachable
+    /// from the [`SocketState::Initial`] state.
+    Connecting(TcpStream),
```
RalfJung, line 53: If you add a `RefCell` here, can you avoid the double-`Rc`?

src/shims/unix/socket.rs
```diff
@@ -32,6 +36,62 @@ enum SocketState {
...
+    ///
+    /// If the connection is established or the socket is in any other state,
+    /// [`Ok`] is returned.
+    ///
+    /// **Important**: On Windows hosts this function cannot be used to ensure the socket is connected.
+    /// Windows treats sockets which are connecting as connected until either the connection timeout hits
+    /// or an error occurs. Thus, the [`TcpStream::peer_addr`] method returns [`Ok`] with the provided peer
+    /// address even when the connection might not yet be established.
+    /// Because of this, on Windows hosts, it's only allowed to call this function after receiving a [`Interest::WRITABLE`]
+    /// for the connecting socket!
```
RalfJung: This seems odd -- if it doesn't work on Windows we shouldn't use it.

But I think what you are saying is: only call this function after you got a writable event. Then it always works. Right? Just leave that in the doc comment, then maybe explain inside the function in a non-doc event why we need that assumption.

src/shims/unix/socket.rs
```diff
@@ -32,6 +36,62 @@ enum SocketState {
...
+    pub fn try_set_connected(&mut self) -> io::Result<()> {
+        let SocketState::Connecting(stream) = self else { return Ok(()) };
+
+        if let Ok(Some(e)) = stream.take_error() {
+            // There was an error whilst connecting.
+            // We won't get EINPROGRESS or ENOTCONNECTED here
```
RalfJung, line 84: If that's true we should `assert!` it.

src/shims/unix/socket.rs
```diff
@@ -32,6 +36,62 @@ enum SocketState {
...
+            // so we need to reset the state.
+            *self = SocketState::Initial;
```
RalfJung, line 89: Please explain that the only state that can transition to `Connecting` is `Initial`.

src/shims/unix/socket.rs
```diff
@@ -232,18 +297,21 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
         };
 
         assert!(this.machine.communicate(), "cannot have `Socket` with isolation enabled!");
 
         // Only allow the same backlog value as the standard library uses since the standard library
         // doesn't provide a way to set a custom value.
+        // FIXME: We probably want to remove this warning as Mio sockets use a backlog value of -1 on
+        //        most targets which defaults to the platform limit. This would mean that we get the
+        //        warning for every program using standard library sockets.
         if backlog != SUPPORTED_LISTEN_BACKLOG {
```
RalfJung: Yeah that makes sense to me. We can just have a comment here saying this is mostly a performance hint that we'll ignore.

src/shims/unix/socket.rs
```diff
@@ -272,11 +340,220 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        let Some(socket) = fd.downcast::<Socket>() else {
+            // Man page specifies to return ENOTSOCK if `fd` is not a socket
+            return this.set_last_error_and_return(LibcError("ENOTSOCK"), dest);
+        };
+
+        assert!(this.machine.communicate(), "cannot have `Socket` with isolation enabled!");
+
+        match &*socket.state.borrow() {
+            SocketState::Initial => {}
```
```suggestion
            SocketState::Initial => { /* fall-through to below */ }
```

src/shims/unix/socket.rs
```diff
@@ -272,11 +340,220 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        // [`TcpStream::take_err`] and [`TcpStream::peer_addr`] both
+        // don't return errors.
+        // For non-blocking sockets we need to check that for every
+        // [`Interest::WRITEABLE`] event on the stream.
+        match TcpStream::connect(address) {
+            Ok(stream) => *socket.state.borrow_mut() = SocketState::Connecting(stream),
+            Err(e) => return this.set_last_error_and_return(e, dest),
+        };
+
+        if cfg!(target_os = "windows") || !socket.is_non_block.get() {
```
RalfJung: For now let's just throw_unsupported for non-blocking `connect`/`accept`

src/shims/unix/socket.rs
```diff
@@ -672,4 +949,189 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let socket = socket_source.borrow();
+                let mut state = socket.state.borrow_mut();
+
+                if let UnblockKind::TimedOut = kind {
+                    // Connecting the socket failed so we need to reset the state.
+                    let old_state = std::mem::replace(&mut *state, SocketState::Initial);
+                    // Socket should have been in connecting state before.
+                    assert!(matches!(old_state, SocketState::Connecting(_)));
+                    return this.set_last_error_and_return(LibcError("ETIMEDOUT"), &dest);
+                };
```
RalfJung: This seems currently unreachable?

src/shims/unix/socket.rs
```diff
@@ -672,4 +949,189 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+                match state.try_set_connected() {
+                    Ok(_) => this.write_scalar(Scalar::from_i32(0), &dest),
+                     // We need to block the thread again as the connection is still not yet ready.
+                     Err(e) if e.kind() == io::ErrorKind::NotConnected || e.kind() == io::ErrorKind::InProgress => {
```
RalfJung: We're checking for those two variants quite a bit... seems like we want an enum representing the two cases?

src/shims/unix/linux_like/syscall.rs
```diff
@@ -59,6 +61,11 @@ pub fn syscall<'tcx>(
             let result = ecx.unix_gettid("SYS_gettid")?;
             ecx.write_int(result.to_u32()?, dest)?;
         }
+        num if num == sys_accept4 => {
+            let [socket, address, address_len, flags] =
+                check_min_vararg_count("syscall(SYS_accept4, ...)", varargs)?;
+            ecx.accept4(socket, address, address_len, Some(flags), dest)?;
```
RalfJung, line 68: Why do we also support the syscall version of this?

src/shims/unix/socket.rs
```diff
@@ -272,11 +340,220 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let fd = this.machine.fds.new_ref(Socket {
+            family,
+            state: RefCell::new(SocketState::Connected(stream)),
+            is_non_block: Cell::new(is_sock_nonblock),
+        });
+        let sockfd = this.machine.fds.insert(fd);
+        // We need to create the scalar using the destination size since
+        // `SYS_accept4` returns a c_long (i64) which doesn't match
+        // the c_int (i32) returned from the `accept`/`accept4` syscalls.
+        this.write_scalar(Scalar::from_int(sockfd, dest.layout.size), dest)
```
RalfJung: I assume tis also all disappears them if we only support blocking sockets for now.

src/shims/unix/socket.rs
```diff
@@ -272,11 +340,220 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        }
+
+        if flags != 0 {
+            throw_unsup_format!(
+                "accept4: flag {flags:#x} is unsupported, only SOCK_CLOEXEC \
+                                and SOCK_NONBLOCK are allowed",
+            );
+        }
+
+        if !socket.is_non_block.get() {
```
RalfJung: This is an odd double negation -- "if it is not non-blocking"

src/shims/unix/socket.rs
```diff
@@ -672,4 +949,189 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    // We checked that the socket is in listening state before blocking
+                    // and since there is no outgoing transition from that state this
+                    // should be unreachable.
+                    unreachable!()
+                };
+
+                if let UnblockKind::TimedOut = kind {
+                    // We pretend the syscall was interrupted by a signal as there usually
+                    // is no timeout on `accept` syscalls.
+                    return this.set_last_error_and_return(LibcError("EINTR"), &dest);
```
RalfJung: Is this even reachable? Sounds to me like we should just `assert!` that `kind` has the right value.

src/shims/unix/socket.rs
```diff
@@ -672,4 +949,189 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            None,
+            callback!(@capture<'tcx> {
+                address_ptr: Pointer,
+                address_len_ptr: Pointer,
+                is_sock_nonblock: bool,
+                socket_source: Rc<RefCell<FileDescriptionRef<Socket>>>,
+                dest: MPlaceTy<'tcx>,
+            } |this, kind: UnblockKind| {
+                let socket = socket_source.borrow();
+                let state = socket.state.borrow();
```
RalfJung, line 920: So there's a `RefCell` inside a `RefCell` here? That seems odd, why does that happen?

src/shims/unix/socket.rs
```diff
@@ -672,4 +949,189 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    // address and we don't need to write anything.
+                    if let Err(e) = this.write_socket_address(&addr, address_ptr, address_len_ptr, "accept4")? {
+                      return this.set_last_error_and_return(e, &dest);
+                    };
+                }
+
+                let fd = this.machine.fds.new_ref(Socket {
+                    family,
+                    state: RefCell::new(SocketState::Connected(stream)),
+                    is_non_block: Cell::new(is_sock_nonblock),
```
RalfJung: So, the accepted connection inherits this flag from the listening socket?

src/shims/unix/socket.rs
```diff
@@ -672,4 +949,189 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                });
+                let sockfd = this.machine.fds.insert(fd);
+                // We need to create the scalar using the destination size since
+                // `SYS_accept4` returns a c_long (i64) which doesn't match
```
RalfJung: > `SYS_accept4` returns a c_long (i64)

Do you have a link documenting this?

src/shims/unix/socket.rs
```diff
@@ -672,4 +949,189 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        match &mut *state {
+            SocketState::Listening(listener) => listener.register(registry, token, interests),
+            SocketState::Connecting(stream) | SocketState::Connected(stream) =>
+                stream.register(registry, token, interests),
+            SocketState::Initial | SocketState::Bound(_) =>
+                Err(io::Error::new(
+                    io::ErrorKind::InvalidInput,
+                    "Only sockets which are listening or connected can \
+                                            be registered for polling",
+                )),
```
RalfJung: Can't this just be `unreachable!`?

src/shims/io_error.rs
```diff
@@ -172,6 +173,7 @@ const WINDOWS_IO_ERROR_TABLE: &[(&str, std::io::ErrorKind)] = {
         ("ERROR_TOO_MANY_LINKS", TooManyLinks),
         ("ERROR_CALL_NOT_IMPLEMENTED", Unsupported),
         ("WSAEWOULDBLOCK", WouldBlock),
+        ("WSAEINPROGRESS", InProgress),
```
RalfJung: This mapping does not seem to be present in the Rust standard library, so I am not sure we should have it here either.

tests/pass-dep/libc/libc-socket.rs
RalfJung, line 1: What about the blocking sockets? Those don't seem to have direct libc tests at the moment?

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -338,27 +425,31 @@ enum LibcSocketAddr {
 /// Wraps a call to a platform function that returns a socket address.
 /// This is very much the same as the function with the same name in the
 /// standard library implementation.
-fn sockname<F>(f: F) -> io::Result<LibcSocketAddr>
+/// Returns a tuple containing the actual return address of the performed
```
```suggestion
/// Returns a tuple containing the actual return value of the performed
```

src/shims/unix/socket.rs
WhySoBad: It's just that the `syscall(...)` syscall always returns a long. I've now added a link to the manpages of syscall.

src/shims/unix/socket.rs
WhySoBad: Yes, it's written like this in the man pages:
>If flags is 0, then accept4() is the same as accept().  The following values can be bitwise ORed in flags to  obtain different behavior:
     SOCK_NONBLOCK   Set the O_NONBLOCK file status flag on the open file description (see open(2)) referred to by the new file descriptor.  Using this flag saves extra calls to fcntl(2) to achieve the same result.

src/shims/unix/linux_like/syscall.rs
WhySoBad: Because on android the `accept4` syscall is manually implemented as a raw syscall: https://github.com/rust-lang/libc/blob/25f7037a3577c0669fc4473be0ab21a28030f68f/src/unix/linux_like/android/b32/arm.rs#L480

src/concurrency/blocking_io.rs
WhySoBad: Yes, however, I cannot unwrap/expect because our value type doesn't implement `Debug`. I've now changed it to manually check whether there was an error and to panic if it was the case.

src/shims/io_error.rs
WhySoBad: I've now suggested adding this variant in the tracking issue of the standard library. If they're okay with that I'll open a pull request there and implement it here as well. For now, I'll remove it.

src/concurrency/thread.rs
```diff
@@ -765,26 +772,45 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
...
+
+        // We also check timeouts before running any other thread, to ensure that timeouts
+        // "in the past" fire before any other thread can take an action. This ensures that for
         // `pthread_cond_timedwait`, "an error is returned if [...] the absolute time specified by
         // abstime has already been passed at the time of the call".
         // <https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_cond_timedwait.html>
         let potential_sleep_time = thread_manager.next_callback_wait_time(clock);
         if potential_sleep_time == Some(Duration::ZERO) {
-            return interp_ok(SchedulingAction::ExecuteTimeoutCallback);
+            // The timeout exceeded for some thread so we execute the timeout callback.
```
WhySoBad: Since we unblock all ready threads it seems a bit odd that we only execute the timeout callback on a single thread even if multiple threads might have an exceeded callback. Should I change this to also execute the callback for all threads with expired timeouts?

tests/pass-dep/libc/libc-socket.rs
WhySoBad: Yeah, I forgot to write a test for that. But now there is one :)

src/shims/unix/socket.rs
RalfJung: Oh, I thought `is_sock_nonblock` is the same as `socket.is_non_block`. Maybe pick a less ambiguous variable name. :)

src/concurrency/thread.rs
RalfJung: Yeah that makes sense. Keep in mind that this also has to do the right thing when communication is off and timeouts are governed by the "fake" clock.

src/concurrency/blocking_io.rs
RalfJung: In that case you can use `unwrap_or_else(|| panic!(...))`. Or, just add a few `derive(Debug)` to make the `unwrap` work.

src/concurrency/thread.rs
WhySoBad: I just had a look at this and it is a bit bigger than I imagined it to be. I think it's better do that in another PR if that's okay.

RalfJung: Sure, as long as this PR doesn't leave things in an incorrect state where we fail to run some of the timeout callbacks.

src/shims/unix/linux_like/syscall.rs
```diff
@@ -59,6 +61,11 @@ pub fn syscall<'tcx>(
             let result = ecx.unix_gettid("SYS_gettid")?;
             ecx.write_int(result.to_u32()?, dest)?;
         }
+        num if num == sys_accept4 => {
+            let [socket, address, address_len, flags] =
```
RalfJung, line 66:
```suggestion
            // Used on Android.
            let [socket, address, address_len, flags] =
```

src/concurrency/blocking_io.rs
```diff
@@ -0,0 +1,193 @@
...
+        let ready = events
+            .into_iter()
+            .map(|event| {
+                let token = event.token();
+                let thread = ThreadId::new_unchecked(token.0.try_into().unwrap());
+                // Deregister this source as we only want to receive one event per thread.
+                let kind = self.deregister(thread);
+                (thread, kind)
+            })
+            .collect::<Vec<_>>();
```
RalfJung, line 77: This creates *two* new vectors. That should not be necessary.

src/concurrency/thread.rs
```suggestion
            // The timeout exceeded for some thread so we unblock one thread and execute its
            // timeout callback.
```

src/concurrency/thread.rs
```diff
@@ -1363,4 +1348,38 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    .filter(|(thread_id, _)| {
+                        let thread = &threads[*thread_id];
+                        // Filter out threads whose timeout already expired to prevent unblocking those because
+                        // of an I/O event.
+                        !matches!(
+                            &thread.state,
+                            ThreadState::Blocked { timeout: Some(timeout), .. } if timeout.get_wait_time(clock) == Duration::ZERO
+                        )
+                    })
+                    .collect::<Vec<_>>(),
```
RalfJung: And *another* `collect`? That makes *three* vector allocations per poll. Please bring that down to 1.

src/concurrency/thread.rs
```diff
@@ -1300,13 +1331,55 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // We can ignore errors originating from interrupts since subsequent calls
+            // to poll are not affected.
```
```suggestion
            // We can ignore errors originating from interrupts; that's just a spurious wakeup.
```

src/concurrency/thread.rs
```diff
@@ -1300,13 +1331,55 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let clock = &this.machine.monotonic_clock;
+        let threads = &this.machine.threads.threads;
+
+        let ready = match this.machine.blocking_io.poll(timeout) {
+            Ok(ready) =>
+                ready
+                    .into_iter()
+                    .filter(|(thread_id, _)| {
+                        let thread = &threads[*thread_id];
+                        // Filter out threads whose timeout already expired to prevent unblocking those because
```
RalfJung: Hm, that is an odd choice. If the I/O event completed just around the timeout, it seems better to treat it as slightly before the deadline rather than discard it as having barely missed the deadline?

src/shims/unix/socket.rs
```diff
@@ -672,4 +899,146 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            None,
+            callback!(@capture<'tcx> {
+                socket: FileDescriptionRef<Socket>,
+                dest: MPlaceTy<'tcx>,
+            } |this, kind: UnblockKind| {
+                assert_eq!(kind, UnblockKind::Ready);
+
+                let mut state = socket.state.borrow_mut();
+
+                match state.try_set_connected() {
```
RalfJung, line 985:
```suggestion
                // We received a "writable" notification so `try_set_connected` is safe to call.
                match state.try_set_connected() {
```
Is this correct?

src/shims/unix/socket.rs
```diff
@@ -57,30 +70,42 @@ impl SocketState {
...
+            drop(stream);
+            // Go back to initial state as the only way of getting into the
+            // `Connecting` state is from the `Initial` state.
             *self = SocketState::Initial;
             return Err(e);
         }
 
         if let Err(e) = stream.peer_addr() {
-            if e.kind() != io::ErrorKind::NotConnected {
+            let e = e.into();
```
RalfJung: This is very magical type inference, please make it more explicit, e.g. `let e = SocketIoError::from(e)`.

src/shims/unix/socket.rs
```diff
@@ -1018,55 +964,44 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                    let old_state = std::mem::replace(&mut *state, SocketState::Initial);
-                    // Socket should have been in connecting state before.
-                    assert!(matches!(old_state, SocketState::Connecting(_)));
-                    return this.set_last_error_and_return(LibcError("ETIMEDOUT"), &dest);
-                };
+                let mut state = socket.state.borrow_mut();
 
                 match state.try_set_connected() {
                     Ok(_) => this.write_scalar(Scalar::from_i32(0), &dest),
                      // We need to block the thread again as the connection is still not yet ready.
```
RalfJung: This comment should be inside the block that does that; currently it looks like it applies to *both* of the following match arms.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -234,22 +210,36 @@ fn test_accept_nonblock() {
...
         // Instantly yield to main thread to ensure that the `connect` syscall
         // was called before we call the `accept` on the server.
         thread::yield_now();
 
+        let (_peerfd, _peer_addr) =
+            sockname(|storage, len| unsafe { libc::accept(server_sockfd, storage, len) }).unwrap();
+
+        // Yield back to the client thread to test whether calling `connect` first also
+        // works.
+        thread::yield_now();
```
RalfJung: This test runs under normal concurrency so it may yield in many places, making this comment somewhat aspirational.

If you want to control the interleaving, either set the `-Zmiri-deterministic-concurrency` flag so that threads *only* get swapped when they are blocked on yield, or add a little `sleep` here to force Miri to run the other thread.

src/concurrency/blocking_io.rs
```diff
@@ -15,11 +14,32 @@ use crate::*;
...
+    fn register(&self, registry: &Registry, token: Token, interests: Interest) -> io::Result<()> {
+        registry.register(&mut *self.borrow_mut(), token, interests)
+    }
+
+    fn deregister(&self, registry: &Registry) -> io::Result<()> {
+        registry.deregister(&mut *self.borrow_mut())
+    }
+}
+
+type BlockingIoSourceRef = Rc<dyn BlockingIoSource>;
```
RalfJung: I don't think this type alias is worth it -- things are more clear when it just gets inlined everywhere IMO.

src/shims/files.rs
```diff
@@ -120,6 +120,36 @@ impl<T: FileDescription + 'static> FileDescriptionExt for T {
...
+/// I/O source to a reference to the blocking I/O source.
+pub trait FileDescriptionBlockingIoSourceExt {
+    fn into_source_rc(self: FileDescriptionRef<Self>) -> Rc<dyn BlockingIoSource>;
+}
+
+impl<T: FileDescription + BlockingIoSource + 'static> FileDescriptionBlockingIoSourceExt for T {
+    fn into_source_rc(self: FileDescriptionRef<Self>) -> Rc<dyn BlockingIoSource> {
+        self.0
+    }
+}
```
RalfJung: This is quite clever, but what it boils down to is "leaking" the internal `Rc` of a `FileDescriptionRef`. It's also maybe a bit too clever, took me a bit to disentangle the new traits and extensions here. ;)

I think we can do something simpler. I see two options:
- hard-code blocking_io to work with `FileDescriptionRef<dyn BlockingIoSource>` or something like that
- accept an extra `Box` allocation

src/shims/unix/socket.rs
```diff
@@ -35,17 +48,17 @@ enum SocketState {
     Bound(SocketAddr),
     /// The `listen` syscall has been called on the socket.
     /// This is only reachable from the [`SocketState::Bound`] state.
-    Listening(TcpListener),
+    Listening(RefCell<TcpListener>),
```
RalfJung: `SocketState` is already a RefCell, so these are nested RefCell. That shouldn't be necessary.

src/shims/unix/socket.rs
WhySoBad: yes

src/concurrency/thread.rs
WhySoBad: Yes I also thought of that case. However, on the other hand, when I specify a timeout, I'd assume that this timeout is strict. It would be kind of confusing if it was strict _except_ in some places where it's okay to exceed by a certain threshold. 

If you're okay with that, then I could change it to allow such behavior. What's the magnitude of threshold you're thinking of?

WhySoBad:
> This also demonstrates how we can avoid the nested `RefCell`

Oh yeah, this was a total oversight. We could've avoided the nested `RefCell` even with the previous implementation.

> I think we can do something simpler. I see two options:
>
>-    hard-code blocking_io to work with `FileDescriptionRef<dyn BlockingIoSource>` or something like that
>-   accept an extra `Box` allocation


I've now implemented the `Box` variant since I think having a more generic API is nicer than saving those, potentially few, heap allocations. 
With my previous implementation I really tried to avoid going onto the heap whilst also not restricting ourselves to only support `FileDescriptionRef`'s. However, I agree that it wasn't very nice to retrace what's going on with that implementation :)

@rustbot ready

src/concurrency/thread.rs
RalfJung: I don't think there should be any threshold. If we see that an event is available for a thread, we should just report that event to the thread.

The point of the timeout is to prevent the thread from stalling too long when nothing happens. But if we're here and the timeout passed, we already screwed that up -- we slept longer than intended instead of waking up the thread. That should be rare, but if it happens, there's no point in throwing away good data like the event we just got.

src/shims/unix/socket.rs
```diff
@@ -22,6 +21,23 @@ enum SocketFamily {
     IPv6,
 }
 
+enum SocketIoError {
+    /// The socket is not yet ready. Either EINPROGRESS or ENOTCONNECTED occurred.
+    #[expect(unused)]
+    NotReady(io::Error),
```
RalfJung: Why is this unused?

RalfJung: Oh, the error inside is unused. Please just remove that then.

src/concurrency/blocking_io.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// Poll instance to monitor I/O events from the OS.
+    /// This is only [`None`] when Miri is run with isolation enabled.
+    poll: Option<Poll>,
+    /// Buffer used to store the ready I/O events when calling [`Poll::poll`].
+    /// This is not part of the state and only stored to avoid allocating a
+    /// new buffer for every poll.
+    events: Events,
+    /// Map between threads which are currently blocked, the kind of I/O
+    /// they are blocked on and the underlying I/O source.
+    sources: FxHashMap<ThreadId, (BlockingIoKind, Box<dyn WithSource>)>,
```
RalfJung: Given that the `BlockingIoKind` tracking is done entirely inside this abstraction, I don't think it is worth its overhead. The existing block kind check exists because block and unblock can be fairly far apart from each other, and they can be anywhere in Miri's shims. We can't always trust all shims to be correct. But here, everything is tightly controlled inside `blocking_io`, there's no way some code inside `socket` could screw up the block kind. IMO that makes it not worth it, the boilerplate for tracking this outweighs the benefit in terms of catching bugs. A bunch of things will become simpler when you remove `BlockingIoKind`, in particular you don't need the `mem::replace` dance in `poll` any more.

src/shims/unix/socket.rs
```diff
@@ -272,11 +343,160 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // The socket is already in a connecting state.
+            SocketState::Connecting(_) =>
+                return this.set_last_error_and_return(LibcError("EALREADY"), dest),
+            // We don't return EISCONN for already connected sockets, for which we're
+            // sure that the connection is established, since TCP sockets are usually
+            // allowed to be connected multiple times.
+            _ =>
+                throw_unsup_format!(
+                    "connect: connecting is only supported for sockets which are neither bound, \
+                                    listening nor already connected"
```
RalfJung: For some reason the second line is much more indented here.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -166,12 +172,84 @@ fn test_listen() {
...
+
+        // Yield back to the client thread to test whether calling `connect` first also
+        // works.
+        thread::sleep(Duration::from_millis(10));
+
+        let (_peerfd, _peer_addr) =
+            sockname(|storage, len| unsafe { libc::accept(server_sockfd, storage, len) }).unwrap();
+    });
+
+    // Test connecting to an already accepting server.
```
RalfJung, line 231: I think you put the `sleep` in the wrong place? Currently, this here will block as `t1` is sleeping before its first `accept`.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -166,12 +172,84 @@ fn test_listen() {
...
+    unsafe {
+        errno_check(libc::connect(
+            client_sockfd,
+            (&addr as *const libc::sockaddr_in).cast::<libc::sockaddr>(),
+            size_of::<libc::sockaddr_in>() as libc::socklen_t,
+        ));
+    }
+
+    // Test connecting first without an accepting server.
```
```suggestion
    // Thread `t1` should now be in its second `sleep`.
    // Test connecting when there is no actively ongoing `accept`.
```

src/concurrency/thread.rs
```diff
@@ -1300,13 +1331,54 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                            let thread = &threads[*thread_id];
+                            // Filter out threads whose timeout already expired to prevent unblocking those because
+                            // of an I/O event.
+                            !matches!(
+                                &thread.state,
+                                ThreadState::Blocked { timeout: Some(timeout), .. } if timeout.get_wait_time(clock) == Duration::ZERO
+                            )
+                        });
+                ready
+            }
```
```suggestion
            Ok(mut ready) => ready,
```

src/shims/unix/socket.rs
WhySoBad: Could you elaborate what you mean? It's 4 tabs like everywhere else.

RalfJung: Just look at it, it's not aligned at all.

image


Also we don't use tabs, it's spaces everywhere.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -227,6 +223,10 @@ fn test_accept_connect() {
             sockname(|storage, len| unsafe { libc::accept(server_sockfd, storage, len) }).unwrap();
     });
 
+    // Yield to server thread to ensure `accept` is called before we try
```
RalfJung, line 227: It's not called the "server thread" anywhere else. I like that term but then you should make it more explicit, like use `let server = thread::spawn` and also have a comment there saying "spawn the server thread" or so.

src/shims/unix/socket.rs
WhySoBad: So it should be aligned to the `:`? Then I'll change it everywhere, because I've always used 4 tabs/16 spaces just like it was done in the socketpair shim .

RalfJung: All lines of the string should start in the same column, like for example [here](https://github.com/rust-lang/rust/blob/40857fcd00c7982cb7ca5f32674f42ce6e1e9688/src/tools/miri/src/concurrency/data_race.rs#L1438-L1457).

That code in the socketpair shim probably got moved around so it's less indented now, and rustfmt doesn't auto-fix indentation so it never got adjusted.

RalfJung:

@rustbot author

PR #4897, opened 2026-03-12
Make protector end accesses weak memory/data race aware

### LINKED ISSUE #3266: Detecting races when releasing Tree Borrows protectors?

JoJoDeveloping:
Protector end accesses are there to ensure that the optimizer can insert accesses throughout the entire lifespan of the protector. But in weak memory, these newly inserted accesses could cause data races. This means that the protector end access must also "count" for data races, to ensure that newly inserted accesses do not introduce new races as the race "already happened" due to the protector end access. But the protector end access is no ordinary access, since it only accesses "upwards" in the borrow tree. In particular children of the accessed node are not affected, and for these nodes it should _not_ count as a data race.

To turn this into a semantics, we declare that protector ends themselves (not the accesses) are weak memory events. This means that a protector might still be active for some other thread if the thread has not yet acquired the protector end (i.e. it the end did not happen-before that thread's clock), and in this case the protector still triggers UB even though it has already ended.

The tests show two examples where it is important that UB is triggered independently of the access under the protector actually happening. This is the case now but was not the case before.

The implementation turns out to be quite complicated. Currently it stores three vector clocks in every node, two of which are (only) needed for the foreign access skipping of rust-lang/miri#4006.

This is WIP. There need to be more tests and it is unknown if we actually want the semantics implemented by this PR. It also needs to be benchmarked; although I'm not sure where things could be sped up if this turns out to be slow.

fixes rust-lang/miri#3266 

see-also https://github.com/rust-lang/unsafe-code-guidelines/issues/585


JoJoDeveloping:
I added some comments and think this is ready for review. But it's still WIP since it's unsure the implemented semantics make sense.

I would recommend that one first reviews the first commit which does some general refactoring, and then the remaining commits.

How can I test the standard library with this changeset?

RalfJung:
> How can I test the standard library with this changeset?

- `./miri install`
- clone https://github.com/rust-lang/miri-test-libstd
- make the `miri` toolchain the default in that folder
- follow the instructions from that repo

PR #4929, opened 2026-03-29
fix ICE when no mangle generic

HerrCai0907:
fixed: https://github.com/rust-lang/rust/issues/154385

Extern symbols can be generic functions because it is a compilation warning instead of error.
But when instancing `Instance::mono`, it requires non generic functions.

e.g.
```rust
#[unsafe(no_mangle)]
fn foo<T: std::fmt::Debug>(value: T)
```

The fix is adding an additional check before create instance.


RalfJung:
Thanks for the PR!

However, before we do anything like this and risk papering over a deeper problem, we need to figure out why this code isn't rejected by rustc. Generic no_mangle items make no sense. Why is that not a hard error?

saethlin:
The lint dates back to https://github.com/rust-lang/rust/pull/30269 and I've been unable to find any conversation about the decision to make it a warning. IMO the most sensible thing here is a hard error, the compiler ignoring `#[no_mangle]` seems really bad.

oli-obk:
Yea, we should just crater turning the lint into an error

HerrCai0907:
for this kind of language change(warn -> error), should i create a rfc firstly? sorry i am a new comer of rust community.

oli-obk:
You can open a PR for rustc directly. You can assign it to me with `r? @oli-obk` in the main PR message or a comment.

I'll crater it (run it on all crates on crates.io), and then will send it to the lang team for FCPing it. This change doesn't need an RFC

RalfJung:
I'll mark this PR as waiting-on-author then while alternative avenues are being explored.

@rustbot author

RalfJung:
I just ran into the same ICE in https://github.com/rust-lang/miri/issues/5128, without realizing it, and wrote a fix in https://github.com/rust-lang/miri/pull/5129. My fix filters out generics earlier, inside `iter_exported_symbols`, which I think is preferable. I will hence close this PR, but I'll cherry-pick your test so that this is not just tested by test-cargo-miri.

PR #4943, opened 2026-04-06
Fix deallocation span for locals freed by `become` tail calls

### LINKED ISSUE #4863: The deallocation span of locals that are deallocated by a tail call `become` is nonsense
Test code:
```rust
#![feature(explicit_tail_calls)]
#![allow(incomplete_features)]

fn g(x: *const i32) {
    let _val = unsafe { *x }; //~ERROR: has been freed, so this pointer is dangling
}

fn f(_x: *const i32) {
    let local = 0;
    let ptr = &local as *const i32;
    become g(ptr)
}

fn main() {
    f(std::ptr::null());
}
```
The deallocation span used to be
```
help: ALLOC was deallocated here:
  --> tests/fail/tail_calls/dangling-local-var.rs:LL:CC
   |
LL |     f(std::ptr::null());
   |     ^^^^^^^^^^^^^^^^^^^
```
now it is
```
help: ALLOC was deallocated here:
  --> tests/fail/tail_calls/dangling-local-var.rs:LL:CC
   |
LL |     let _val = unsafe { *x };
   |         ^^^^
```
Both of these make no sense. I think the ideal span would be to point at the `become`.

LaneAsade:
This PR fixes rust-lang/miri#4863

The problem was due to the fact that `before_memory_deallocation` used `current_user_relevant_span` to record where an allocation was freed, which produced wrong spans for the locals deallocated by a `become` tail call. 

The solution was to introduce a new field called `tail_call_span` on `MiriMachine` which records `become` expression's span in `before_terminator` if a `TailCall` terminator is detected and then use that span in `before_memory_deallocation` instead of the current one.

### DIFF
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -662,7 +662,7 @@ pub struct MiriMachine<'tcx> {
     pub short_fd_operations: bool,
 
     // Stores the span of the `TailCall` (`become`) terminator
-    pub tail_call_span:Option<Span>,
+    pub tail_call_span: Option<Span>,
 }
 
 impl<'tcx> MiriMachine<'tcx> {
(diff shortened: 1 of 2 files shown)


RalfJung:
Thanks for looking into this, and sorry for the slow review!

However, I don't think this logic is correct. Once you set `tail_call_span`, it will never be unset, so it becomes the deallocation span of *all allocations henceforth*. That's clearly wrong. And this will be hard to fix since you really only want to affect the deallocations that happen *while the `TailCall` terminator is executed*, not anything that happens after.

I think this will require work in the core interpreter. Here on the Miri side, the best we can do are hard-to-get-right and, worse, hard-to-maintain hacks.

@rustbot author

RalfJung:
I'm going to close this due to inactivity. @LaneAsade if you want to get back to this, I'd suggest to start a discussion on Zulip where you sketch a design for how you'd like to fix this issue. Note that this is a very non-trivial issue so I would recommend to maybe start with something easier (like something marked "good first issue").

PR #4944, opened 2026-04-06
feat: tid support to sched_getaffinity/sched_setaffinity

### LINKED ISSUE #4062: Add TID support to sched_getaffinity/sched_setaffinity

hsqStephenZhang:
Fix https://github.com/rust-lang/miri/issues/4062

### DIFF
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -420,9 +420,14 @@ pub enum TimeoutAnchor {
     Absolute,
 }
 
-/// An error signaling that the requested thread doesn't exist.
+/// An error signaling that the requested thread doesn't exist or has terminated.
 #[derive(Debug, Copy, Clone)]
-pub struct ThreadNotFound;
+pub enum ThreadLookupError {
+    /// No thread with this ID exists.
+    InvalidId,
+    /// The thread exists but has already terminated.
+    Terminated(ThreadId),
+}
 
 /// A set of threads.
 #[derive(Debug)]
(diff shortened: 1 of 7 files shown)


src/concurrency/thread.rs
```diff
@@ -498,6 +498,12 @@ impl<'tcx> ThreadManager<'tcx> {
         }
     }
 
+    /// Find a thread by its internal index, if the thread is still alive (not terminated).
+    pub fn thread_id_from_index_if_alive(&self, index: u32) -> Option<ThreadId> {
```
RalfJung: Just above this we can find `thread_id_try_from`, and looking at how it is used it probably also should return `Err` if the thread is terminated.

So I'd suggest that instead of adding a new function, you change the behavior of the old function (and add a doc comment explaining its behavior).

RalfJung (commented):
Thanks a lot for the PR, and sorry for the slow review.

This looks great! I think we can avoid adding a new function here. Also the test will need minor adjustments once https://github.com/rust-lang/miri/pull/4994 lands to be sure that the new part is only executed on Linux/Android.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4944/2f07bd6affcd0247f96f71f4c9cc5306612b16d8..6750bb4df994b8bedcb72cd8d45f235c70268e48)*

RalfJung:
Ah, fair, there are some operations that should work on terminated threads... nice way of handling that. :)

@rustbot author

src/shims/windows/handle.rs
```diff
@@ -185,10 +184,13 @@ impl Handle {
 
         match Self::from_packed(handle) {
             Some(Self::Thread(thread)) => {
-                // validate the thread id
+                // validate the thread id; Windows handles remain valid even after thread termination
```
RalfJung:
```suggestion
                // Validate the thread id. Windows handles remain valid even after thread termination.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4944/2f07bd6affcd0247f96f71f4c9cc5306612b16d8..b0c9ddfa861f13b9172b1eb197ba3ef13c39f027)*

RalfJung: FWIW you forgot the `.` at the end. I recommend copy-pasting or clicking the button to apply the suggestion.

But we can just land it like this, it's fine.

hsqStephenZhang: ops, sorry about that. 

thanks.

PR #4947, opened 2026-04-08
tree borrows: implicit writes

quiode:
This PR tries to address https://github.com/rust-lang/unsafe-code-guidelines/issues/584#issue-3440604811. It is part of a bachelor thesis supervised by @JoJoDeveloping and @RalfJung, for more information, see: [Project_Description.pdf](https://github.com/user-attachments/files/26537277/Project_Description.pdf).
This implements the checking for implicit writes for Tree Borrows. It is disabled by default but can be enabled using the `-Zmiri-tree-borrows-implicit-writes` flag.
When it is enabled, Miri inserts a write for all mutable borrows on function entry. This enables the optimization implemented here: https://github.com/rust-lang/rust/pull/155207


src/bin/miri.rs
```diff
@@ -531,6 +532,16 @@ fn main() -> ExitCode {
                         "`-Zmiri-tree-borrows` is required before `-Zmiri-tree-borrows-no-precise-interior-mut`"
                     ),
             };
+        } else if arg == "-Zmiri-tree-borrows-no-writable" {
```
RalfJung: We want "no writable" (the status quo) to be the default. So please invert the meaning of the flag and call it something like `-Zmiri-tree-borrows-implicit-writes`.

src/borrow_tracker/mod.rs
```diff
@@ -224,6 +224,7 @@ pub enum BorrowTrackerMethod {
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub struct TreeBorrowsParams {
     pub precise_interior_mut: bool,
+    pub writable: bool,
```
RalfJung:
```suggestion
    /// Controls whether `&mut` function arguments are immediately activated with
    /// an implicit write.
    pub implicit_writes: bool,
```

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -138,6 +138,17 @@ impl<'tcx> NewPermission {
         let ty_is_freeze = pointee.is_freeze(*cx.tcx, cx.typing_env());
         let is_protected = retag_kind == RetagKind::FnEntry;
 
+        // Check if the strong mode / writable optimization has been disabled for this function using the `-Zmiri-tree-borrows-no-writable` flag
+        let no_writable = !cx
```
RalfJung: You negate this here and then negate `no_writable` again later. Please avoid this needless double negation.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -159,14 +170,32 @@ impl<'tcx> NewPermission {
             _ => Permission::new_reserved_im(),
         };
 
-        // Everything except for `Cell` gets an initial access.
-        let initial_access = |perm: &Permission| !perm.is_cell();
+        // [0] contains freeze information, [1] contains non freeze information
+        let mut access_perms = [(freeze_perm, None), (nonfreeze_perm, None)];
```
RalfJung: Is there a nicer, more declarative way to write this that doesn't rely on mutable state?

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -159,14 +170,32 @@ impl<'tcx> NewPermission {
...
+        for (permission, access) in &mut access_perms {
+            // Everything except for `Cell` gets an initial read.
+            if !permission.is_cell() {
+                // Every explicit mutable reference that gets an initial read also gets an initial write (except if it has been explicitly disabled using the `-Zmiri-tree-borrows-no-writable` flag).
```
RalfJung: This comment does not match the code -- the code also checks for `RetagKind::FnEntry`

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -159,14 +170,32 @@ impl<'tcx> NewPermission {
...
+                // Thus implicit mutable references (Two Phase borrowing) and Raw pointers are excluded.
+                if !no_writable
+                    && Some(Mutability::Mut) == ref_mutability
+                    && matches!(retag_kind, RetagKind::FnEntry)
+                {
+                    *access = Some(AccessKind::Write);
+
+                    // if cell gets a write, permission is gonna be set to Unique
```
RalfJung: Please use proper English grammar and punctuation.
```suggestion
                    // If cell gets a write, permission is going to be set to Unique.
```
But also... there's no "cell" here? We are in the `!is_cell` branch!

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -287,31 +316,36 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                 ptr_size,
-                LocationState::new_accessed(
-                    Permission::new_disabled(),
-                    IdempotentForeignAccess::None,
+                (
+                    LocationState::new_accessed(
+                        Permission::new_disabled(),
+                        IdempotentForeignAccess::None,
+                    ),
+                    None,
```
RalfJung: Please add a comment explaining the meaning of this `None`.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -324,42 +358,71 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        for (perm_range, (_perm, access)) in inside_perms.iter_all() {
+            if let Some(access) = access {
+                // Some reborrows incur a read/write access to the parent.
+                // As a write also implies a read, a single write is performed instead of a read and a write.
+
+                // writing to an immutable allocation (static variables) is UB, check this here
+                if *access == AccessKind::Write
+                    && this.get_alloc_mutability(alloc_id).unwrap().is_not()
+                {
+                    throw_ub!(WriteToReadOnly(alloc_id))
```
RalfJung, line 372: It is curious that Stacked Borrows does not need an explicit check like this. Any idea why?

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -109,13 +109,13 @@ impl<'tcx> Tree {
 pub struct NewPermission {
     /// Permission for the frozen part of the range.
     freeze_perm: Permission,
-    /// Whether a read access should be performed on the frozen part on a retag.
-    freeze_access: bool,
+    /// Whether and what kind of access should be performed on the frozen part on a retag.
+    freeze_access: Option<AccessKind>,
+    /// Whether a write access should be performed on the frozen part on a retag.
```
RalfJung: This comment here doesn't seem to make sense?  It is part of the doc comment for `nonfreeze_perm`...

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -265,14 +265,19 @@ impl Permission {
...
 
-    /// Default initial permission of the root of a new tree at inbounds positions.
-    /// Must *only* be used for the root, this is not in general an "initial" permission!
+    #[cfg(test)]
+    pub fn is_unique(&self) -> bool {
+        self.inner == Unique
+    }
+
+    /// Default initial permission for mutable references at the start of a function that is either
+    /// protected or not interior mutable (see *writable* and *strong mode*) and of the root of a new tree at inbounds positions.
```
RalfJung: Better have the comment just say what it does without referring to how it is used -- such comments should go into the NewPermission logic.

(I know you just updated the old comments, but now that the logic got more complicated it's not worth duplicating it here.)

Same below for `new_reserved_frz`.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -73,6 +73,11 @@ impl LocationState {
         Self { permission, accessed: true, idempotent_foreign_access: sifa }
     }
 
+    #[cfg(test)]
+    pub fn possible(&self) -> bool {
```
RalfJung, line 78: What is possible here? Please add a doc comment.

src/data_structures/dedup_range_map.rs
```diff
@@ -55,6 +55,22 @@ impl<T> DedupRangeMap<T> {
             .unwrap()
     }
 
+    /// Creates a new [DedupRangeMap] of type `N` by applying `f` to all it's elements.
+    ///
+    /// The range for each element remains unchanged, use [DedupRangeMap::iter_mut] to change it. For more fine-grained control, the range is provided to `f`.
+    pub fn transform<F, N>(self, mut f: F) -> DedupRangeMap<N>
```
RalfJung: `map` would be a more natural name I think.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -324,42 +358,71 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                        };
                     }
                 }
             }
         }
+
+        let mut tree_borrows = this.get_alloc_extra(alloc_id)?.borrow_tracker_tb().borrow_mut();
+
+        // remove access information
+        let inside_perms = inside_perms.transform(|_, (loc_state, _)| loc_state);
```
RalfJung: Creating an entire new map for this seems rather wasteful. Is that truly the best approach?

RalfJung (commented):
I didn't look at the tests yet, and have to dig more into the logic, but here are some first comments.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4947/667f5c0e34aca5f32e66379107ae40e23adfd09e..7f8b7052ab122bd1423de6583863db166a5e1392)*

src/borrow_tracker/tree_borrows/tree.rs
JoJoDeveloping: I wrote this function, the answer is whether this location state is "legal," where illegal is something like "Unique without the init flag" or other stuff removed as part of https://github.com/minirust/minirust/pull/276

src/borrow_tracker/tree_borrows/mod.rs
JoJoDeveloping: note that this takes ownership of the old map.

src/borrow_tracker/tree_borrows/mod.rs
RalfJung: Sketch for cleaned-up version:
```suggestion
        let unique_perm = if writable { new_unique } else { new_reserved_frz };
        let freeze_perm = match ref_mutability {
```
Then use this instead of new_reserved_frz but only if protected.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -159,14 +170,32 @@ impl<'tcx> NewPermission {
...
+            }
+        }
 
         Some(NewPermission {
-            freeze_perm,
-            freeze_access: initial_access(&freeze_perm),
-            nonfreeze_perm,
-            nonfreeze_access: initial_access(&nonfreeze_perm),
+            freeze_perm: access_perms[0].0,
+            freeze_access: access_perms[0].1,
```
RalfJung: access is entirely a function of perm... we probably want to encode it as such, i.e., remove the `access` fields from `NewPermission`. We can then have a method on `Permission` to compute the initial access... in fact that method already exits, it just needs to be renamed! It is currently called `protector_end_access`. New name suggestion: `associated_access`.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -287,31 +316,36 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             .precise_interior_mut;
 
         // Compute initial "inside" permissions.
-        let loc_state = |frozen: bool| -> LocationState {
+        let loc_state = |frozen: bool| -> (LocationState, Option<AccessKind>) {
```
RalfJung: Adding `Option<AccessKind>` here should not be necessary any more once the initial access is a method on `Permission`

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -324,42 +358,71 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-            if perm.accessed() {
-                // Some reborrows incur a read access to the parent.
                 // Adjust range to be relative to allocation start (rather than to `place`).
                 let range_in_alloc = AllocRange {
                     start: Size::from_bytes(perm_range.start) + base_offset,
                     size: Size::from_bytes(perm_range.end - perm_range.start),
                 };
 
+                let alloc_extra = this.get_alloc_extra(alloc_id)?;
+                let mut tree_borrows = alloc_extra.borrow_tracker_tb().borrow_mut();
```
RalfJung: Why does this happen each time through the loop? Would be better to only do it once upfront.

src/borrow_tracker/tree_borrows/mod.rs
RalfJung: This goes away when we do the other planned refactors.

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -265,14 +265,19 @@ impl Permission {
         self.inner == Cell
     }
 
-    /// Default initial permission of the root of a new tree at inbounds positions.
-    /// Must *only* be used for the root, this is not in general an "initial" permission!
+    #[cfg(test)]
+    pub fn is_unique(&self) -> bool {
```
RalfJung, line 269: Please add a brief doc comment.

src/borrow_tracker/tree_borrows/tree.rs
RalfJung: That sounds reasonable. Maybe call it `reachable`? And the comment could then exactly what you said. :)

src/data_structures/dedup_range_map.rs
RalfJung: The entire function can be removed again I think, after the aforementioned refactors.

tests/pass/tree_borrows/implicit_writes/flag.rs
RalfJung, line 1: The point of all these new `pass` tests is not entirely clear to me. I would suggest to go find all existing tests that have the following
```
//@revisions: stack tree
//@[tree]compile-flags: -Zmiri-tree-borrows
```
and then add a new revision, tree_implicit_writes or so, that also enable the new flag there.

We also have a bunch of TB-specific tests. Please discuss with Johannes which of those should get an "implicit writes" revision. Once we have all that, do we even need any more `pass` tests?

tests/fail/both_borrows/aliasing_mut1.rs
```diff
@@ -1,9 +1,11 @@
-//@revisions: stack tree
+//@revisions: stack tree iwrites
```
RalfJung: `tree_iwrites` maybe? The current name is not clear enough IMO.

tests/fail/function_calls/arg_inplace_locals_alias_ret.tree.stderr
```diff
@@ -7,7 +7,7 @@ LL | fn callee(x: S) -> S {
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Tree Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/tree-borrows.md for further information
    = help: the accessed tag <TAG> (root of the allocation) is foreign to the protected tag <TAG> (i.e., it is not a child)
-   = help: this reborrow (acting as a foreign read access) would cause the protected tag <TAG> (currently Unique) to become Disabled
+   = help: this reborrow (acting as a foreign write access) would cause the protected tag <TAG> (currently Unique) to become Disabled
```
RalfJung: Why does this output change when the flag is not set for the test?

tests/fail/tree_borrows/implicit_writes/flag.rs
RalfJung, line 1: This test looks like a duplicate of `as_mut_ptr`?

tests/fail/tree_borrows/implicit_writes/fnentry_invalidation.rs
```diff
@@ -0,0 +1,19 @@
+//@compile-flags: -Zmiri-tree-borrows -Zmiri-tree-borrows-implicit-writes
+// This shows that with checking for implicit writes, Tree Borrows behaves more like Stacked Borrows, and the additional write in `fail/tree_borrows/fnentry_invalidation.rs` is not needed detect / cause UB.
```
RalfJung: The end of this sentence is not grammatical so I am not quite sure what you mean. Is this exactly identical to the SB test `tests/fail/stacked_borrows/fnentry_invalidation.rs`? If so, please say so. If not -- why not?

tests/fail/tree_borrows/implicit_writes/mut_option.rs
RalfJung, line 1: How is this different from as_mut_ptr? It just hides the argument in an `Option` but that's not worth a separate test...

tests/fail/tree_borrows/implicit_writes/ptr_write2.rs
```diff
@@ -0,0 +1,16 @@
+// Tests that UB exists for an explicit write, even without checking for implicit writes.
+// Thus to allow the compiler to perform an optimization where he inserts a write at this position (when we don't have the explicit one), we have to check that this does not cause UB -> reason for `-Zmiri-tree-borrows-implicit-writes`.
+//@compile-flags: -Zmiri-tree-borrows
```
RalfJung: This isn't even an implicit-write test? We already have this case covered for regular TB with existing tests.

RalfJung (commented):
And here's the second round, now also with comments for the logic and the tests. :)  Please ask either in this PR, or Johannes on Zulip, if you have questions.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4947/667f5c0e34aca5f32e66379107ae40e23adfd09e..982b2e1a4ca02fb0fb49d39ed31c9e889561bf1e)*

tests/fail/tree_borrows/implicit_writes/ptr_write.rs
RalfJung, line 1: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4947/667f5c0e34aca5f32e66379107ae40e23adfd09e..982b2e1a4ca02fb0fb49d39ed31c9e889561bf1e)*

Please also add a version of this test where `x` has type `&mut UnsafeCell<u8>`.

src/borrow_tracker/mod.rs
```diff
@@ -224,6 +224,7 @@ pub enum BorrowTrackerMethod {
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub struct TreeBorrowsParams {
     pub precise_interior_mut: bool,
+    pub implicit_writes: bool,
```
RalfJung, line 228: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4947/667f5c0e34aca5f32e66379107ae40e23adfd09e..7826cfacf26562332e5f6904713fb202e4cd150d)*

This is the place where the comment you just added should go -- it goes with the struct definition, not with the use.

src/borrow_tracker/tree_borrows/mod.rs
quiode: it was needed in a previous iteration of the code but it seems like it's no longer necessary

RalfJung:
> Reminder, once the PR becomes ready for a review, use `@rustbot ready`.

@rustbot ready
I need this for my TODO tracking.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -138,35 +133,49 @@ impl<'tcx> NewPermission {
...
             _ => Permission::new_reserved_frz(),
         };
         let nonfreeze_perm = match ref_mutability {
             // Shared references are "transparent".
             Some(Mutability::Not) => Permission::new_cell(),
-            // *Protected* mutable references and boxes are reserved without regarding for interior mutability.
+            // *Protected* mutable references are unique or reserved.
+            Some(Mutability::Mut) if is_protected => unique_perm,
+            // *Protected* boxes are reserved without regarding for interior mutability.
             _ if is_protected => Permission::new_reserved_frz(),
```
RalfJung:
```suggestion
            None if is_protected => Permission::new_reserved_frz(),
```
To make it more clear that this is just for boxes.

src/borrow_tracker/tree_borrows/perms.rs
```diff
@@ -266,18 +266,17 @@ impl Permission {
     }
 
     #[cfg(test)]
```
RalfJung: Attributes generally go after the doc comment, not before.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -73,6 +73,12 @@ impl LocationState {
         Self { permission, accessed: true, idempotent_foreign_access: sifa }
     }
 
+    #[cfg(test)]
+    /// Tests that an initial location is only Unique if implicit writes are checked, thus if it is accessed on initial creation.
```
```suggestion
    /// Checks whether the current location state is ever reachable in a real execution.
    #[cfg(test)]
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -73,6 +73,12 @@ impl LocationState {
...
+    pub fn possible(&self) -> bool {
+        self.accessed || !self.permission.is_unique()
```
RalfJung, line 80:
```suggestion
        // `Unique` can only be reached on actually accessed locations.
        self.accessed || !self.permission.is_unique()
```

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -73,6 +73,12 @@ impl LocationState {
         Self { permission, accessed: true, idempotent_foreign_access: sifa }
     }
 
+    #[cfg(test)]
+    /// Tests that an initial location is only Unique if implicit writes are checked, thus if it is accessed on initial creation.
+    pub fn possible(&self) -> bool {
```
RalfJung, line 78: Please also add a debug assertion somewhere in the state transition logic which checks that indeed we never reach an "impossible" state. (This serves to check that this function is correct.)

src/borrow_tracker/mod.rs
```diff
@@ -224,6 +224,7 @@ pub enum BorrowTrackerMethod {
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub struct TreeBorrowsParams {
     pub precise_interior_mut: bool,
+    // Controls whether `&mut` function arguments are immediately activated with an implicit write.
```
RalfJung: Please make this a doc comment (`/// ...`).

tests/pass/tree_borrows/implicit_writes/flag.rs
RalfJung: This still hasn't been done: please make sure the existing both-borrows pass tests are also run with implicit writes.

tests/pass/tree_borrows/implicit_writes/ptr_write.rs
```diff
@@ -0,0 +1,36 @@
...
+
+fn normal() {
+    let mut x = 0u8;
+    let ptr = &raw mut x;
+    let res = foo(&mut x, ptr);
+    assert_eq!(res, 0);
+}
+
+fn foo(_x: &mut u8, y: *mut u8) -> u8 {
+    // implicit write inserted here for x
```
RalfJung: These comments are confusing since no implicit write is being inserted here in this test.

README.md
```diff
@@ -515,6 +515,7 @@ to Miri failing to detect cases of undefined behavior in a program.
   track interior mutable data on the level of references instead of on the
   byte-level as is done by default.  Therefore, with this flag, Tree
   Borrows will be more permissive.
+* `-Zmiri-tree-borrows-implicit-writes` enables the tracking of implicit / spurious writes. This only has an effect in Miri if [Tree Borrows] is enabled. This makes Tree Borrows less permissive.
```
RalfJung:
```suggestion
* `-Zmiri-tree-borrows-implicit-writes` enables implicit writes for all `&mut` function arguments.
  This makes Tree Borrows less permissive.
```
Please also keep the list alphabetic.

RalfJung (commented):
Next round of comments. There are not that many left. :)
(But you apparently missed one comment in my previous review.)

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4947/667f5c0e34aca5f32e66379107ae40e23adfd09e..581375d72694b955a467154938ab0d4b75c6d5a7)*

RalfJung:
Please also update the PR description; it no longer reflects the current state of this PR.

src/bin/miri.rs
```diff
@@ -531,6 +532,16 @@ fn main() -> ExitCode {
                         "`-Zmiri-tree-borrows` is required before `-Zmiri-tree-borrows-no-precise-interior-mut`"
                     ),
             };
+        } else if arg == "-Zmiri-tree-borrows-implicit-writes" {
```
JoJoDeveloping, line 535: I feel like we should at least warn the user if they have `-Z mutable-writable` in their flags, and is in TB mode without the correct Miri flag. From a usability perspective I would ideally like the Miri flag to there be enabled automatically.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4947/667f5c0e34aca5f32e66379107ae40e23adfd09e..581375d72694b955a467154938ab0d4b75c6d5a7)*

RalfJung: The current proposal calls the flag `-Zllvm-writable`. That seems fairly clear that that flag does not affect Miri.

tests/pass/tree_borrows/implicit_writes/flag.rs
quiode: I'm waiting for an answer from @JoJoDeveloping for the second part. I will implement it as soon as it is clear which tests should get a new revision.

RalfJung: In a first step, everything in `pass` that has `//@revisions: stack tree` should also get a `tree_implicit_writes` revision.

quiode: Some tests fail with implicit writes, for example they use `as_mut_ptr`. Should I put a copy of them in the fails directory, just add the failing output for the revision or remove the revision entirely?

RalfJung: Ah, I see. Don't add the revision for those then.  We can add it once your PR with the attribute landed.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -138,35 +133,49 @@ impl<'tcx> NewPermission {
...
+            None if is_protected => Permission::new_reserved_frz(),
             // Unprotected mutable references and boxes start in `ReservedIm`.
             _ => Permission::new_reserved_im(),
         };
 
-        // Everything except for `Cell` gets an initial access.
-        let initial_access = |perm: &Permission| !perm.is_cell();
-
         Some(NewPermission {
             freeze_perm,
```
JoJoDeveloping: this actually has a subtle bug, since this permission will be used for the "outside" permission. But this should always be `Reserved` (not `Reserved initialized` and not `Unique`), even if there are initial writes happening "inside."

To solve this properly, we can either add an "outside perm" field to `NewPermission`. Or we can do a larger refactor where first insert the node with the `freeze_perm` and then perform the initial access on the tree-with-the-new-node such that its permissions are affected accordingly. This means that we would have logic for selecting the initial permission (`Frozen`/`Reserved`/`Cell`) and the initial access (`None`/`Read`/`Write`) unlike the current thing which tries to do both together. It would also mean that we stop "overloading" the `initialized` field. I would find this approach much cleaner overall, since it is also compatible with https://github.com/minirust/minirust/pull/276 and mirrors the reborrow logic from [minirust](https://github.com/minirust/minirust/blob/master/spec/mem/tree_borrows/reborrow_settings.md?plain=1#L122); for this reason I prefer the latter.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4947/667f5c0e34aca5f32e66379107ae40e23adfd09e..46d78c2b213c75e962def59a182d57a0ab355672)*

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -138,35 +133,49 @@ impl<'tcx> NewPermission {
...
         };
         let nonfreeze_perm = match ref_mutability {
             // Shared references are "transparent".
             Some(Mutability::Not) => Permission::new_cell(),
-            // *Protected* mutable references and boxes are reserved without regarding for interior mutability.
-            _ if is_protected => Permission::new_reserved_frz(),
+            // *Protected* mutable references are unique or reserved.
+            Some(Mutability::Mut) if is_protected => unique_perm,
+            // *Protected* boxes are reserved without regarding for interior mutability.
+            None if is_protected => Permission::new_reserved_frz(),
```
JoJoDeveloping: we should double-check how https://github.com/rust-lang/rust/pull/155207 treats Boxes.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4947/667f5c0e34aca5f32e66379107ae40e23adfd09e..46d78c2b213c75e962def59a182d57a0ab355672)*

JoJoDeveloping (changes requested):

src/borrow_tracker/tree_borrows/mod.rs
quiode: Like this:
```rust
// CHECK: @box_moved(ptr noalias noundef nonnull align 4 %0)
#[no_mangle]
pub fn box_moved(_: Box<i32>) {}

// CHECK: @box_borrow(ptr noalias noundef readonly align 8 captures(address, read_provenance) dereferenceable(8) %_1)
#[no_mangle]
pub fn box_borrow(_: &Box<i32>) {}

// CHECK: @box_borrow_mut(ptr noalias noundef writable align 8 dereferenceable(8) %_1)
#[no_mangle]
pub fn box_borrow_mut(_: &mut Box<i32>) {}

// CHECK: @box_interior_mut(ptr noalias noundef nonnull align 8 %0)
#[no_mangle]
pub fn box_interior_mut(_: Box<&mut i32>) {}
```

quiode:
While looking over the code, an issue came forward which is mentioned in https://github.com/rust-lang/miri/pull/4947#discussion_r3095265028. 29d5336cb9d7a5c335606e80d7f2920aef1d94a7 tries to address this issue by "constraining" the outside permission to Reserved if it is Unique. A test that currently breaks (but should not), but with the changes not anymore, has been implemented in `array-pointer-access.rs`. Also added additional assertions when using `is_initial()` as this now allows Unqiue which some functions do not expect.

src/borrow_tracker/tree_borrows/tree.rs
```diff
@@ -121,6 +122,7 @@ impl LocationState {
                 exposed_cache.update_exposure(nodes, idx, old_access_level, access_level);
             }
         }
+        debug_assert!(self.possible());
```
RalfJung, line 125: The `self.perform_access` call in this function already checked this, didn't it? No need to check it again.

tests/fail/tree_borrows/implicit_writes/ptr_write.rs
```diff
@@ -2,7 +2,6 @@
 //@compile-flags: -Zmiri-tree-borrows -Zmiri-tree-borrows-implicit-writes
 
 fn foo(_x: &mut u8, y: *mut u8) -> u8 {
-    // implicit write inserted here for x
```
RalfJung: Why did you remove this comment?

tests/fail/tree_borrows/implicit_writes/ptr_write_unsafe_cell.rs
```diff
@@ -4,7 +4,6 @@
 use std::cell::UnsafeCell;
 
 fn foo(_x: &mut UnsafeCell<u8>, y: *mut u8) -> u8 {
-    // implicit write inserted here for x
```
RalfJung: Same here, this comment seems useful to point out what the read conflicts with. Maybe it could have been more explicit, but better some comment than no comment. ;)

tests/pass/tree_borrows/array-pointer-access.rs
RalfJung, line 1: You said this test is supposed to check the "outside accesses" problem that we discussed? I am very confused.
- The test name and description never even mention this.
- The test does not enable implicit writes so it can't possibly test anything about them.

RalfJung (commented):
Thanks for the revision! However I am confused by some of the changes.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4947/667f5c0e34aca5f32e66379107ae40e23adfd09e..29d5336cb9d7a5c335606e80d7f2920aef1d94a7)*

RalfJung:
I have pushed a refactor of the permission logic that is a bit longer than what you had but IMO also easier to follow. Please let me know what you think.

src/borrow_tracker/tree_borrows/tree.rs
quiode: indeed

src/borrow_tracker/tree_borrows/mod.rs
JoJoDeveloping: The first one is important here, and shows that `Box`es do not become `writable` in the LLVM back-end. I don't remember really discussing whether this is the desirable semantics. Under Stacked Borrows, adding this attribute would be sound, right? @RalfJung

RalfJung: `writable` is meaningless without `dereferenceable` and `Box` cannot be `dereferenceable`.

But thinking about it, that's an LLVM limitation, nothing we need to tie ourselves to. For TB it'd make sense that `Box` also starts out as `Unique`, IMO.

JoJoDeveloping: Yes, `Box` should have initial writes happen like for mutable references. @quiode can you make it so, and also add a test for that?

For further context, `Box` can not be `dereferenceable` in LLVM because that attribute declares dereferenceability throughout the entire function, but `Box` can be deallocated. If LLVM had a `dereferenceable_on_entry` we would use it for `Box`, and then we could also use `writable` (which already only has "on entry" semantics). But being dereferenceable just on entry is not very powerful in terms of allowed optimizations, so there are no real plans currently to add it to LLVM.

Either way, we should trigger enough UB here to later allow adding such attributes if it ever becomes relevant.

tests/fail/tree_borrows/implicit_writes/ptr_write.rs
quiode: i assumed that was meant by the comment here: https://github.com/rust-lang/miri/pull/4947#discussion_r3079841976

quiode: but in that case I'm gonna rephrase them

RalfJung: That link doesn't go anywhere for me unfortunately. Github is really bad at linking to comments in PRs sadly...

quiode: image
that was the comment I meant to reference

RalfJung: Ah. Yes what I meant was to clarify the comment, e.g.
```
// The TB retag performs an implicit write for x here.
```
I think when I first read the comment I interpreted "implicit write" as writes LLVM actually inserts during transforms, not as whatever happens during a TB retag, and that's what confused me.

src/borrow_tracker/tree_borrows/mod.rs
quiode: For clarification, that new implicit write should only happen for mutable boxes? And also the change is only in Miri, not adding the `writable` attribute in rustc too, even though it currently wouldn't do anything. @JoJoDeveloping

RalfJung: It should happen for all `Box`; there is no distinction of mutability for them. And yes, only in Miri.

tests/pass/tree_borrows/read_retag_no_race.rs
```diff
@@ -1,5 +1,6 @@
 //@revisions: tree tree_implicit_writes
 //@[tree_implicit_writes]compile-flags: -Zmiri-tree-borrows-implicit-writes
+//@[tree_implicit_writes]ignore-target: x86_64-apple-darwin x86_64-pc-solaris
```
JoJoDeveloping: Hold on, this test is supposed to fail with `tree_implicit_writes` (see the comments in this file). And indeed it does so... but only sometimes.

Can you turn this into a test that always fails with `-Zmiri-tree-borrows-implicit-writes`, for all architectures? Doing so will likely require some tricks with `-Zmiri-deterministic-concurrency`, disabling all printing, and using `thread::yield_now()`. We can discuss more over Zulip.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4947/855aa6e2b6fa7a1ef0f228a9c723bbbd60972407..211bc15a6ab269b35e7bc4305eec48c9f3504cb8)*

RalfJung: Also, please never just ignore some targets without explanation. That always needs a proper explanation.

JoJoDeveloping: PS: if we already have a test testing that we do the data race tracking correctly here, then this porting-over is not needed. But I think we currently don't do so, do we?

quiode: Sorry for the confusion, I just wanted a quick and dirty way to see which targets fail as I didn't know exactly which local test command to use to do the extensive CI tests.

RalfJung: Always a good idea to add a `// FIXME` with a brief comment when pushing such a quick-and-dirty change to a branch where other people will look at it. :)

quiode:
All requested changes should be implemented now.
@rustbot ready

tests/fail/tree_borrows/implicit_writes/read_retag.rs
```diff
@@ -0,0 +1,110 @@
...
+// with just barriers. We must remove preemption so that the execution and the
+// error messages are deterministic.
+//@compile-flags: -Zmiri-deterministic-concurrency
+use std::ptr::addr_of_mut;
+use std::sync::{Arc, Barrier};
+use std::thread;
+
+unsafe extern "Rust" {
+    safe fn miri_write_to_stdout(bytes: &[u8]);
+}
```
JoJoDeveloping: the printing is not an essential part of the test, please just remove it entirely.

tests/fail/tree_borrows/implicit_writes/read_retag.rs
```diff
@@ -0,0 +1,110 @@
...
+
+#[derive(Copy, Clone)]
+struct SendPtr(*mut u8);
+
+unsafe impl Send for SendPtr {}
+
+// This test features the problematic pattern
```
JoJoDeveloping: the introductory test here is not helpful since it yaps about two interleavings and why we're only testing one etc. This might have been useful for the original test, but you're already referencing that from this file so there's no need to repeat it. Just say plainly that this test tests that the implicit write also counts in the data race model or something like that 🙃

tests/fail/tree_borrows/implicit_writes/read_retag.rs
```diff
@@ -0,0 +1,110 @@
...
+        barrier.wait();
+    }};
+}
+
+fn thread_1(x: SendPtr, barrier: IdxBarrier) {
+    let x = unsafe { &mut *x.0 };
+    synchronized!(barrier, "spawn");
+    thread::yield_now();
+
+    synchronized!(barrier, "read x || retag y");
```
JoJoDeveloping: these two `synchronized!` calls can be combined, right? Nothing happens in-between them in both threads.

tests/fail/tree_borrows/implicit_writes/read_retag.rs
```diff
@@ -0,0 +1,110 @@
...
+    thread::yield_now();
```
JoJoDeveloping: is it truly necessary to yield in-between every barrier? I don't think so...

tests/fail/tree_borrows/implicit_writes/read_retag.rs
```diff
@@ -0,0 +1,110 @@
...
+    fn write(y: &mut u8, v: u8, barrier: &IdxBarrier) {
+        //~^ ERROR: Data race detected
+        synchronized!(barrier, "write y");
+        thread::yield_now();
+        *y = v;
+        thread::yield_now();
+    }
+
+    synchronized!(barrier, "read x || retag y");
+    thread::yield_now();
```
JoJoDeveloping: I am not 100% convinced the yielding here is robust. I would have expected something like:

Thread 1 (with the "actual" access) exists the final barrier, performs the read, and exits/blocks.

Thread 2 (with the retag) exists the final barrier, yields a bunch to ensure thread 1 can do the actual access first no matter who continues first after the barrier, and then performs the retag which triggers the data race.

This somewhat elaborate scheme should ensure a deterministic error location, which will be at the second access, i.e. the implicit (retag) write.

tests/fail/tree_borrows/implicit_writes/read_retag.rs
```diff
@@ -0,0 +1,110 @@
...
+    let y = unsafe { &mut *y.0 };
+    synchronized!(barrier, "spawn");
+    thread::yield_now();
+
+    fn write(y: &mut u8, v: u8, barrier: &IdxBarrier) {
+        //~^ ERROR: Data race detected
+        synchronized!(barrier, "write y");
+        thread::yield_now();
+        *y = v;
+        thread::yield_now();
```
JoJoDeveloping: there is no need to keep the code after the UB, right?

tests/fail/tree_borrows/implicit_writes/read_retag.stdout
JoJoDeveloping, line 1: if we delete the printing, we can delete this file entirely.

JoJoDeveloping (changes requested):
the test is good, but I have some comments mostly affecting the things beside the code that is run as part of this test. (I.e. it's mostly bikeshedding, sorry.)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4947/855aa6e2b6fa7a1ef0f228a9c723bbbd60972407..2a947c6b0a714cfc19c5f34128d4da6e879d66f0)*

JoJoDeveloping:
I will try to have another look over the entire code soon :tm: but it's a very busy week for me.

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -138,36 +133,78 @@ impl<'tcx> NewPermission {
...
+                None =>
+                    if is_protected && implicit_writes && !matches!(part, Outside) {
+                        // Boxes are treated the same as mutable references.
+                        Permission::new_unique()
+                    } else if is_protected || frozen {
+                        // We also use this for protected `Box<UnsafeCell>` as otherwise adding
+                        // `noalias` would not be sound.
+                        Permission::new_reserved_frz()
+                    } else {
+                        Permission::new_reserved_im()
```
JoJoDeveloping, line 199: @RalfJung I don't think it makes sense to mark Boxes as `ReservedIM`, does it? It's what the logic was doing before and we likely didn't notice... (Note that Minirust does the same).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4947/855aa6e2b6fa7a1ef0f228a9c723bbbd60972407..cce5ab71e3ee13709cb298c15f545086b3baffbf)*

tests/fail/both_borrows/aliasing_mut1.rs
JoJoDeveloping, line 1: @RalfJung We added the `tree_iwrites` revision to *some* of the fail tests. There are many more fail tests that we could add this to. Is there a reason why not to do so?

Specifically, this should probably be on all tests in `fail/both_borrows` and on selected tests in `fail/tree_borrows` even outside the `fail/tree_borrows/implicit_writes` folder...

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4947/855aa6e2b6fa7a1ef0f228a9c723bbbd60972407..cce5ab71e3ee13709cb298c15f545086b3baffbf)*

src/borrow_tracker/tree_borrows/mod.rs
RalfJung: Why does it not make sense? `Box<T>` as a function argument should be treated the same as `&mut T`, just without the protector.

JoJoDeveloping: this is about `Box` in non-function-argument position, i.e. when moved inside a method.

JoJoDeveloping:
> I will try to have another look over the entire code soon ™️ but it's a very busy week for me.

I had another look at it and am now happy with it :tada: (except for the things noted above)

src/borrow_tracker/tree_borrows/mod.rs
RalfJung: What is your proposal for what their initial state should be?

JoJoDeveloping: just Reserved. For mutable references we have ReservedIM because of the safe code in [this test](https://github.com/rust-lang/miri/blob/master/tests/pass/tree_borrows/reserved.rs#L72), but I don't think that is possible with boxes.

RalfJung: Interesting. We figured it'd be like `&mut` because "why not".

I'd say lets leave this unchanged for now, but please open an issue or PR (depending on what you have time for) specifically for this change, also including an example of code that you think should have UB but currently doesn't because of this.

tests/pass/tree_borrows/array-pointer-access.rs
```diff
@@ -0,0 +1,30 @@
+// Tests that accessing the same array using disjoint pointers does not cause UB under Tree Borrows.
+// The pointer will access the memory range using the outside permission, thus testing if it has been set correctly.
```
RalfJung:
```suggestion
// The pointer will access the memory range using the outside permission, thus testing if it has been set correctly.
// In particular, in implicit_writes mode we used to set the "outside" permissions to
// `Unique`, which caused an ICE.
// Put differently, this checks the absence of subobject provenance for arrays.
```

tests/fail/tree_borrows/implicit_writes/ptr_write_box.rs
```diff
@@ -0,0 +1,16 @@
+// same test as `ptr_write.rs` but with `Box`
+//@compile-flags: -Zmiri-tree-borrows -Zmiri-tree-borrows-implicit-writes
+
+fn main() {
+    let mut x: Box<u8> = Box::new(0u8);
+    let ptr = &mut *x as *mut u8;
```
RalfJung: There's still a mutable reference here. If that isn't relevant, maybe use `&raw mut *x` instead?

tests/fail/tree_borrows/implicit_writes/retag_is_race.rs
JoJoDeveloping, line 1: can you rename this test into `retag_is_data_race.rs` or something like that? The current name is in praticular wrong as it is not a "read" retag.

tests/fail/tree_borrows/implicit_writes/read_retag.rs
```diff
@@ -0,0 +1,47 @@
...
+    barrier.wait(); // write
+}
+
+fn thread_2(y: SendPtr, barrier: Arc<Barrier>) {
+    let y = unsafe { &mut *y.0 };
+    barrier.wait(); // init
+    thread::yield_now(); // force other thread to read first
+
+    fn write(y: &mut u8, v: u8, barrier: &Arc<Barrier>) {
+        //~^ ERROR: Data race detected
```
JoJoDeveloping: can you make the error message more specific, to match on the retag being called a *write*?

tests/pass/tree_borrows/tree-borrows.rs
```diff
@@ -1,4 +1,6 @@
-//@compile-flags: -Zmiri-tree-borrows
```
RalfJung, line 1: You can keep this line so that all revisions use TB.

tests/pass/tree_borrows/no_implicit_writes.rs
RalfJung, line 1: This is quite confusing to be in this folder as this test is about  *not* having implicit writes.

Maybe call it `tests/pass/tree_borrows/no_implicit_writes.rs`?

tests/pass/tree_borrows/implicit_writes/as_mut_ptr.rs
RalfJung, line 1: This can be merged with `tests/pass/tree_borrows/implicit_writes/ptr_write.rs`.

tests/pass/tree_borrows/array-pointer-access.rs
RalfJung, line 1: Maybe make this just a function in `tests/pass/tree_borrows/tree-borrows.rs`? That's where we have most of these kinds of tests.

RalfJung (commented):
Thanks! Just a few more comments for the tests, the rest looks good. :)

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4947/855aa6e2b6fa7a1ef0f228a9c723bbbd60972407..cce5ab71e3ee13709cb298c15f545086b3baffbf)*

tests/fail/both_borrows/aliasing_mut1.rs
RalfJung: You already added this is more than enough places I think. The pass tests are more critical since the flag makes TB less permissive.

JoJoDeveloping (approved):
some other cosmetic things in tests I noticed while reading over this with Ralf.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4947/855aa6e2b6fa7a1ef0f228a9c723bbbd60972407..cce5ab71e3ee13709cb298c15f545086b3baffbf)*

quiode:
all done

@rustbot ready

RalfJung:

@rustbot author

PR #4954, opened 2026-04-10
Per-byte handling for UnsafePinned

### LINKED ISSUE #4819: Make UnsafePinned handling properly per-byte
UnsafeCell gets treated in a per-byte way by SB and TB (at least in the default configuration). https://github.com/rust-lang/rust/pull/151365 adds support for UnsafePinned, but so far we just handle this in an all-or-nothing way. We should make it per-byte as well.

LaneAsade:
Fixes rust-lang/miri#4819

This PR ensures that `UnsafePinned` is handled in per-byte way just like `UnsafeCell` by adding a check for `UnsafePinned` alongside the existing `UnsafeCell` check in `visit_value`. This makes sure that only the bytes actually covered by the `UnsafePinned` opt out and not the entire allocation.

### DIFF
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -443,7 +443,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     }
 
     /// Visits the memory covered by `place`, sensitive to freezing: the 2nd parameter
-    /// of `action` will be true if this is frozen, false if this is in an `UnsafeCell`.
+    /// of `action` will be true if this is frozen, false if this is in an `UnsafeCell` or `UnsafePinned`.
     /// The range is relative to `place`.
     fn visit_freeze_sensitive(
         &self,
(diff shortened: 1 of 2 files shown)


RalfJung:
Thanks! However, I think it is too early for this. We currently still honor the `Unpin` trait, for which it is rather unclear what it even means to do things per-byte. We have to first land https://github.com/rust-lang/rust/pull/152946 before we can remove the `Unpin` hack from Miri, and then we can consider doing per-byte `UnsafePinned`.

Also, this PR treats UnsafeCell and UnsafePinned the same, which I don't think is right.
And finally, the test doesn't actually test anything. The same code already gets accepted by today's Miri.

LaneAsade:
No worries! I was actually confused by `UnsafeCell` and `UnsafePinned` because it seemed like there weren't any differences between them. I genuinely thought that was how it was. Maybe its cuz I couldn't find a lot of information regarding `UnsafePinned`. I would love to work on this once #152946 is done (Of course, only after I have a full understanding of miri internals). Also, sorry for the trouble you went through to review my PR.

RalfJung:
>  I was actually confused by UnsafeCell and UnsafePinned because it seemed like there weren't any differences between them. 

As a general rule, please always point out your confusions when submitting a PR. The people reviewing your PRs will thank you. :)

LaneAsade:
Yes, I will keep that in mind from now on. Thank you for the advice. Again, I apologize for any inconvenience caused.

RalfJung:
No worries. :)

PR #4955, opened 2026-04-12
Allow changing socket blocking state

WhySoBad:
Hi,

This pull request adds the required functionality to change the network socket blocking state through `ioctl` (used by the standard library) and `fcntl`.

Since the `ioctl` shim now allows changing the socket blocking state through the standard library, there are now also some tests for testing non-blocking sockets using the standard library. However, we cannot test non-blocking connects this way, as we would need mio for that.

Is it worth adding mio as a test dependency for this in a future PR? Because we already test non-blocking connects using libc.

Also, because setting the blocking state of a socket through `fcntl` is supported on all targets, we can now also run the libc non-blocking socket tests on MacOS targets.

### DIFF
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -62,6 +62,20 @@ pub trait UnixFileDescription: FileDescription {
         throw_unsup_format!("cannot flock {}", self.name());
     }
 
+    /// Modifies device parameters.
+    /// `op` is the device-dependent operation code. It's either a `c_long` or `c_int`, depending on
+    /// the target and whether it uses glibc or musl.
+    /// `arg` is the optional third argument which exists depending on the operation code. It's either
+    /// an integer or a pointer.
+    fn ioctl<'tcx>(
+        &self,
+        _op: Scalar,
+        _arg: Option<&OpTy<'tcx>>,
+        _ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, i32> {
+        throw_unsup_format!("cannot use ioctl on {}", self.name());
+    }
+
     /// Return which epoll events are currently active.
     fn epoll_active_events<'tcx>(&self) -> InterpResult<'tcx, EpollEvents> {
         throw_unsup_format!("{}: epoll does not support this file description", self.name());
@@ -127,6 +141,39 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // return `0` if flock is successful
         let result = result.map(|()| 0i32);
         interp_ok(Scalar::from_i32(this.try_unwrap_io_result(result)?))
+    }
+
+    fn ioctl(
+        &mut self,
+        fd: &OpTy<'tcx>,
+        op: &OpTy<'tcx>,
+        varargs: &[OpTy<'tcx>],
+    ) -> InterpResult<'tcx, Scalar> {
+        let this = self.eval_context_mut();
+
+        let fd = this.read_scalar(fd)?.to_i32()?;
+        let op = this.read_scalar(op)?;
+        // There is at most one relevant variadic argument.
+        // It exists depending on the device and the opcode and thus we can't
+        // use `check_min_vararg_count` here.
+        let arg = varargs.first();
+
+        let Some(fd) = this.machine.fds.get(fd) else {
+            return this.set_last_error_and_return_i32(LibcError("EBADF"));
+        };
+
+        // Handle common opcodes.
+        let fioclex = this.eval_libc("FIOCLEX");
+        let fionclex = this.eval_libc("FIONCLEX");
+        if op == fioclex || op == fionclex {
+            // Since we don't support `exec`, those are NOPs.
+            return interp_ok(Scalar::from_i32(0));
+        }
+
+        // Since some ioctl operations use the return value as an output parameter, we cannot strictly use the convention of
+        // zero indicating success and -1 indicating an error.
+        let return_value = fd.as_unix(this).ioctl(op, arg, this)?;
+        interp_ok(Scalar::from_i32(return_value))
     }
 
     fn fcntl(
--- a/src/shims/unix/socket.rs
+++ b/src/shims/unix/socket.rs
@@ -15,7 +15,8 @@ use rustc_target::spec::Os;
 
 use crate::concurrency::blocking_io::InterestReceiver;
 use crate::shims::files::{EvalContextExt as _, FdId, FileDescription, FileDescriptionRef};
-use crate::{OpTy, Scalar, *};
+use crate::shims::unix::UnixFileDescription;
+use crate::*;
 
 #[derive(Debug, PartialEq)]
 enum SocketFamily {
@@ -171,6 +172,10 @@ impl FileDescription for Socket {
         true
     }
 
+    fn as_unix<'tcx>(&self, _ecx: &MiriInterpCx<'tcx>) -> &dyn UnixFileDescription {
+        self
+    }
+
     fn get_flags<'tcx>(&self, ecx: &mut MiriInterpCx<'tcx>) -> InterpResult<'tcx, Scalar> {
         let mut flags = ecx.eval_libc_i32("O_RDWR");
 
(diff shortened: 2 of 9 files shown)


src/shims/unix/fd.rs
```diff
@@ -129,6 +143,44 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let op = this.read_scalar(op)?;
+        let arg = if varargs.len() <= 1 {
+            varargs.first()
+        } else {
+            throw_ub_format!(
+                "ioctl: got {} variadic parameters but only a single optional \
+                variadic parameter is allowed",
+                varargs.len()
+            )
+        };
```
RalfJung: This needs a comment explaining that there is at most one relevant argument here. Also, if there are more arguments that simply does not matter and can be ignored.

src/shims/unix/fd.rs
```diff
@@ -129,6 +143,44 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            return this.set_last_error_and_return_i32(LibcError("EBADF"));
+        };
+
+        let fioclex = this.eval_libc("FIOCLEX");
+        let fionclex = this.eval_libc("FIONCLEX");
+
+        if op == fioclex || op == fionclex {
+            // Since we don't support `exec`, those are NOPs.
+            return interp_ok(Scalar::from_i32(0));
+        }
```
RalfJung, line 171:
```suggestion
        // Handle common opcodes.
        let fioclex = this.eval_libc("FIOCLEX");
        let fionclex = this.eval_libc("FIONCLEX");
        if op == fioclex || op == fionclex {
            // Since we don't support `exec`, those are NOPs.
            return interp_ok(Scalar::from_i32(0));
        }
```

src/shims/unix/socket.rs
```diff
@@ -183,10 +188,59 @@ impl FileDescription for Socket {
...
+                // FIONBIO cannot be used to change the blocking mode of a socket on solarish targets:
+                // <https://github.com/rust-lang/rust/commit/dda5c97675b4f5b1f6fdab64606c8a1f21021b0a>
+                throw_unsup_format!(
+                    "ioctl: setting FIONBIO on sockets is unsupported on solarish targets"
+                );
+            }
+
+            let Some(value_ptr) = arg else {
+                // FIONBIO requires the third argument.
+                return ecx.set_last_error(LibcError("EINVAL")).map(|_| -1);
```
RalfJung: This should raise UB -- there's normally no way to check whether a vararg is present.

src/shims/unix/socket.rs
```diff
@@ -183,10 +188,59 @@ impl FileDescription for Socket {
...
+        op: Scalar,
+        arg: Option<&OpTy<'tcx>>,
+        ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, i32> {
+        assert!(ecx.machine.communicate(), "cannot have `Socket` with isolation enabled!");
+
+        let fionbio = ecx.eval_libc("FIONBIO");
+
+        if op == fionbio {
+            if matches!(ecx.tcx.sess.target.os, Os::Solaris | Os::Illumos) {
```
RalfJung: Let's make this an allowlist rather than a denylist. Who knows which other targets are weird about this...

tests/pass-dep/libc/libc-socket-no-blocking.rs
```diff
@@ -18,26 +18,128 @@ use libc_utils::*;
...
+}
+
+/// Test that nonblocking TCP server sockets return [`ErrorKind::WouldBlock`] when trying
 /// to accept when no incoming connection exists. This also tests that nonblocking server sockets
 /// are still able to accept incoming connections should they already exist before the `accept` or
 /// `accept4` syscall is called.
 fn test_accept_nonblock() {
-    // Create a new non-blocking server socket.
-    let (server_sockfd, addr) = net::make_listener_ipv4(libc::SOCK_NONBLOCK).unwrap();
+    let (server_sockfd, addr) = net::make_listener_ipv4(0).unwrap();
```
RalfJung: This looks like we can remove the flag argument now?

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -49,10 +49,15 @@ fn main() {
 }
 
 fn test_socket_close() {
```
RalfJung: Maybe pick a new name?

tests/pass-dep/libc/libc-socket-no-blocking.rs
```diff
@@ -18,26 +18,128 @@ use libc_utils::*;
...
     test_send_recv_dontwait();
     test_write_read_nonblock();
 
     test_getpeername_ipv4_nonblock();
     test_getpeername_ipv4_nonblock_no_peer();
 }
 
-/// Test that nonblocking TCP server sockets return [`io::ErrorKind::WouldBlock`] when trying
+/// Test that setting the O_NONBLOCK flag changes the blocking state of a socket.
+fn test_fcntl_nonblock_opt() {
```
RalfJung, line 59: Please validate at least these new flag checking tests (ideally the entire file) on a real Linux system.

tests/pass/shims/socket-no-blocking.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    // Start server thread.
+    let handle = thread::spawn(move || {
+        // Yield back to client thread to ensure that we have an incoming connection.
+        thread::sleep(Duration::from_millis(10));
+
+        // Accepting when there is an existing incoming connection should
+        // succeed without blocking.
+
+        let (_stream, _peer_addr) = listener.accept().unwrap();
```
RalfJung: Let's use `-Zmiri-fixed-schedule` and careful `yield` rather than relying on timing for this.

RalfJung (commented):
Thanks! :) Just a bunch of fairly minor comments.

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4955/42fef81c68bf2a50c1e4120aac7818d8c80cc689..2951dfbebef218022c59e8e699803d8e5db1af46)*

tests/pass-dep/libc/libc-socket-no-blocking.rs
WhySoBad: I've ran the tests locally without any errors/failed assertions.

RalfJung: Nice, thanks :)

src/shims/unix/socket.rs
```diff
@@ -222,17 +222,20 @@ impl UnixFileDescription for Socket {
         let fionbio = ecx.eval_libc("FIONBIO");
 
         if op == fionbio {
-            if matches!(ecx.tcx.sess.target.os, Os::Solaris | Os::Illumos) {
+            if !matches!(ecx.tcx.sess.target.os, Os::Linux | Os::Android | Os::MacOs | Os::FreeBsd)
+            {
                 // FIONBIO cannot be used to change the blocking mode of a socket on solarish targets:
                 // <https://github.com/rust-lang/rust/commit/dda5c97675b4f5b1f6fdab64606c8a1f21021b0a>
```
RalfJung, line 230: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4955/42fef81c68bf2a50c1e4120aac7818d8c80cc689..780ee1244a289ecfaa5809dff1378d61d97f3657)*

That's a curious comment in that commit

> The ioctl(FIONBIO) method of setting a file descriptor to be
non-blocking does not notify the underlying resource in the same way
that fcntl(F_SETFL, O_NONBLOCK) does on illumos and Solaris.

How does F_SETFL "notify" the underlying resource, what does that entail? We don't do any notifying here either...

tests/pass/shims/socket-no-blocking.rs
```diff
@@ -81,7 +79,8 @@ fn test_send_recv_nonblock() {
         match stream.read(&mut buffer[bytes_read..]) {
             Ok(read) => bytes_read += read,
             Err(err) if err.kind() == ErrorKind::WouldBlock => {
-                // Not all data is written into the stream, wait for a bit.
+                // Not all data is written into the stream, wait for a bit
+                // and yield to the server thread if it's not terminated.
                 thread::sleep(Duration::from_millis(10));
```
RalfJung: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4955/42fef81c68bf2a50c1e4120aac7818d8c80cc689..780ee1244a289ecfaa5809dff1378d61d97f3657)*

This can be just `yield_now` now since we know exactly what the scheduler will do.

src/shims/unix/socket.rs
WhySoBad: I think they differ in how the kernel stores the non-blocking flag. From the following reply it seems as if solarish systems have two different non-blocking flags and thus `ioctl` cannot be used to set the flag read by `fcntl`: https://community.hpe.com/t5/operating-system-hp-ux/ioctl-fionbio/m-p/2721984/highlight/true#M63364

I can't find much online about this so it could also be that there is another problem why you cannot use `ioctl` on soarlish.

RalfJung: Curious.

For the platforms which you have now allowed here, are there docs about how this behaves / how it relates to the POSIX variant?

WhySoBad: I couldn't find a man page documenting how FIONBIO is implemented on the linux, macos and freebsd man pages.

I've looked it up in the linux souce code, and there it's basically the same as `fcntl` (I'd assume for android it's the same as well). In the freebsd source it calls the device specific ioctl (which I didn't look up any further). And well, for macos I couldn't look up anything.

RalfJung: FWIW the macos kernel and even some of the userland are actually open source. See for instance [here](https://github.com/apple-oss-distributions/libplatform/blob/a00a4cc36da2110578bcf3b8eeeeb93dcc7f4e11/src/os/lock.c#L645).

But, given that Rust uses the ioctl on these targets, I guess it's fine...

src/shims/unix/socket.rs
```diff
@@ -183,10 +188,62 @@ impl FileDescription for Socket {
...
+        op: Scalar,
+        arg: Option<&OpTy<'tcx>>,
+        ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, i32> {
+        assert!(ecx.machine.communicate(), "cannot have `Socket` with isolation enabled!");
+
+        let fionbio = ecx.eval_libc("FIONBIO");
+
+        if op == fionbio {
+            if !matches!(ecx.tcx.sess.target.os, Os::Linux | Os::Android | Os::MacOs | Os::FreeBsd)
```
RalfJung, line 227: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4955/42fef81c68bf2a50c1e4120aac7818d8c80cc689..9218194591150fda3f9556e8d9ca8846f2bbbbe1)*

```suggestion
            // On these OSes, Rust uses the ioctl, so we trust that it is reasonable and controls
            // the same internal flag as fcntl.
            if !matches!(ecx.tcx.sess.target.os, Os::Linux | Os::Android | Os::MacOs | Os::FreeBsd)
```

RalfJung:
This looks great, thanks! After resolving the last comment nit, please squash the commits. You can squash manually if there are multiple independent commits you want to preserve, or use `./miri squash` (make sure to pick a suitable commit message). Then write `@rustbot ready` after you force-pushed the squashed PR.

@rustbot author

src/shims/unix/socket.rs
WhySoBad: Interesting, I didn't know that.

WhySoBad:
@RalfJung some test failed on the CI on the PR on my branch, can you still remove it from the merge queue?

WhySoBad:
I think this is a timing issue. Probably, the socket is still connecting when we're calling `recv`. If that's the case, we also need to loop there, but this time when ENOTCONN is returned:
https://github.com/WhySoBad/miri/actions/runs/24403117527/job/71278916255

WhySoBad:
> If that's the case, we also need to loop there, but this time when ENOTCONN is returned

I've now added a new test, which only tests non-blocking connect and awaits the connection using such a loop. Everywhere else I changed the tests to connect whilst the client socket is blocking and only afterwards set the socket to be non-blocking.

tests/pass-dep/libc/libc-socket-no-blocking.rs
```diff
@@ -467,12 +502,12 @@ fn test_write_read_nonblock() {
...
 
-    // Non-blocking connects always "fail" with EINPROGRESS.
-    let err = net::connect_ipv4(client_sockfd, addr).unwrap_err();
-    assert_eq!(err.kind(), ErrorKind::InProgress);
+    unsafe {
+        // Change client socket to be non-blocking.
+        errno_check(libc::fcntl(client_sockfd, libc::F_SETFL, libc::O_NONBLOCK));
+    }
 
     // We are connecting and the server socket is not writing.
```
RalfJung: Shouldn't this say "connected" now?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4955/5cc22e9a6d705a723bd9c803f50cfb423bda8946..08d9e0e5856c4c3b0f4bebe52b3207f786dc1597)*

RalfJung:
:+1: please re-squash. (You can do this preemptively for such tiny fixes if you want.)

PR #4956, opened 2026-04-12
GenMC: Don't split up non-atomic accesses to smaller chunks

michaliskok:
GenMC is now able to handle larger and mixed-size non-atomic accesses. (Follow up on rust-lang/miri#4949).
This deals with one of the checkmarks in https://github.com/rust-lang/miri/issues/4572.

r? @RalfJung


RalfJung:
Nice :)

tests/genmc/pass/atomics/u64_as_u32_array.rs
RalfJung, line 1: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4956/0722600f188a328d457eeb8608103bbbbd57b682..400f5c23157cfe696c918cb186f7a1ebf11f9c6f)*

Actually I just realized this test doesn't turn on GenMC mode. 😂

RalfJung: Maybe we should set `-Zmiri-genmc` automatically for tests in that folder, though this could be tricky to implement in our test runner.

michaliskok: Nah, it's my fault for copy-pasting test code left and right :-)

PR #4959, opened 2026-04-13
Miri-GenMC API refactor

michaliskok:
Another follow up on rust-lang/miri#4949 making the following changes:

- Replaces bool with per-operation status enums in CXX types
- Renames `*Ext` factory namespaces to `Make*` and drops `from_` prefix in CXX
- Uses `std::visit` with `overloaded{}` for `HandleResult` dispatch in `MiriGenmcShim`
- Adds `GenmcHandleResult<T>` to collapse 3-field status checks in Miri
- Flattens `genmc-sys/cpp/src/` folder
- Renames MiriGenmcShim to MiriGenMCInterface

These map to the respective commits.

r? @RalfJung

I'm not sure how this looks to you. I'm happy to split to smaller PRs or revise as you see fit.


RalfJung:
> Renames MiriGenmcShim to MiriGenMCInterface

The usual naming convention in Rust is CamelCase, including for acronyms, entirely disregarding the usual capitalization of the component words. That's why it's Arc, not ARC or ARc, and IoSliceMut, not IOSliceMut. So it should be MiriGenmcInterface.

genmc-sys/src/lib.rs
```diff
@@ -209,69 +209,111 @@ mod ffi {
         blocked_execs: u64,
     }
 
+    #[derive(Debug, Clone, Copy)]
```
RalfJung, line 212: The usual style is to put attributes after the doc comment, so please move them down.

This applies consistently everywhere; I won't add comments for every instance of this.

genmc-sys/src/lib.rs
```diff
@@ -209,69 +209,111 @@ mod ffi {
...
+    /// Status tag shared by operations with a single success mode
+    /// (atomic loads/stores, non-atomic loads/stores, and FAIs.
```
RalfJung, line 214: Missing closing parenthesis before the `)`.

(And GH doesn't allow suggestions when one does per-commit review...)

genmc-sys/src/lib.rs
```diff
@@ -209,69 +209,111 @@ mod ffi {
...
+    /// For operations with multiple success modes, see [`CasStatus`] and [`MutexLockStatus`].
```
RalfJung, line 215: Hm, is that really worth a separate enum? Wouldn't it be more uniform to always use this enum and treat the success mode as "payload"?

genmc-sys/src/lib.rs
```diff
@@ -209,69 +209,111 @@ mod ffi {
...
+        Reset,
+        /// The lock was successfully acquired by this thread.
+        Acquired,
+        /// The lock is already held; this thread has been blocked on the GenMC side.
+        NotAcquired,
+    }
+
     #[must_use]
     #[derive(Debug)]
+    /// Result from an atomic load. When `status == Ok`, `read_value` is valid.
```
RalfJung, line 256: The 2nd sentence is redundant with the per-field comments.

genmc-sys/src/lib.rs
```diff
@@ -209,69 +209,111 @@ mod ffi {
...
-        /// The value that was read. Should not be used if `has_value` is `false`.
+        /// Valid when `status == Ok`.
         read_value: GenmcScalar,
     }
 
+    #[must_use]
+    #[derive(Debug)]
+    /// Result from a non-atomic load or store. GenMC uses these only for data-race detection;
+    /// no value is communicated back to Miri. When `status == Ok` there is no payload.
+    struct NonAtomicResult {
```
RalfJung, line 269: Please don't put this *in between* LoadResult and StoreResult. Those two should be next to each other.

src/concurrency/genmc/mod.rs
```diff
@@ -451,24 +451,29 @@ impl GenmcCtx {
...
+                throw_ub_format!(
+                    "{}",
+                    cas_result.error.as_ref().unwrap().to_string_lossy()
+                ),
+            CasStatus::Failed | CasStatus::Succeeded => {}
+            status => unreachable!("unexpected CasStatus: {status:?}"),
         }
 
         let return_scalar = genmc_scalar_to_scalar(ecx, self, cas_result.old_value, size)?;
+        let is_success = matches!(cas_result.status, CasStatus::Succeeded);
```
RalfJung, line 467: You're matching on the enum a 2nd time here, that's an anti-pattern. You can instead extract this info already in the first match:
```rust
let is_success = match status {
  ...
  Succeeded => true,
  Faild => false,
};
```

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
RalfJung, line 501: Can load_ret also be handled with a visitor? I guess that's tricky to express because `std::visit` does not allow early-return.

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
```diff
@@ -106,7 +106,10 @@ void MiriGenmcShim::handle_assume_block(ThreadId thread_id, AssumeType assume_ty
             [](const VerificationError& err) { return MakeLoadResult::error(format_error(err)); },
             [](const Invalid&) { return MakeLoadResult::invalid(); },
             [](const SVal& v) { return MakeLoadResult::value(v); },
-            [](const auto&) { UNREACHABLE(); },
+            [](const auto&) {
+                UNREACHABLE();
+                return MakeLoadResult::invalid();
```
RalfJung, line 111: Why is this needed now? Or should this have been in the previous commit?

genmc-sys/src/lib.rs
```diff
@@ -5,6 +5,28 @@ pub use cxx::UniquePtr;
...
+/// - [`LoadResult`]:           `T = GenmcScalar`                      (the value read)
+/// - [`NonAtomicResult`]:      `T = ()`                               (no payload)
+/// - [`StoreResult`]:          `T = bool`                             (is the write co-max)
+/// - [`ReadModifyWriteResult`]:`T = (GenmcScalar, GenmcScalar, bool)` (old, new, is_co_max)
+///
+/// For operations with more than one success mode ([`CompareExchangeResult`],
+/// [`MutexLockResult`]) the status enums carry sufficient information and should be
+/// matched directly.
+#[must_use]
+pub enum GenmcHandleResult<T> {
```
RalfJung, line 21: "handle" sounds like a verb here. Maybe call this just `GenmcResult`?

genmc-sys/src/lib.rs
```diff
@@ -5,6 +5,28 @@ pub use cxx::UniquePtr;
...
 pub use self::ffi::*;
 
+/// The result of a GenMC handle operation, after converting the raw C++ result struct into an
+/// idiomatic Rust type.
+///
+/// The payload type `T` is operation-specific:
+/// - [`LoadResult`]:           `T = GenmcScalar`                      (the value read)
+/// - [`NonAtomicResult`]:      `T = ()`                               (no payload)
+/// - [`StoreResult`]:          `T = bool`                             (is the write co-max)
+/// - [`ReadModifyWriteResult`]:`T = (GenmcScalar, GenmcScalar, bool)` (old, new, is_co_max)
```
RalfJung, line 15: No need to give an exhaustive list.

genmc-sys/src/lib.rs
```diff
@@ -44,6 +66,81 @@ pub fn create_genmc_driver_handle(
...
+            OperationStatus::Ok => GenmcHandleResult::Ok(self.is_coherence_order_maximal_write),
+            OperationStatus::Error => GenmcHandleResult::Error(cxx_string_to_owned(&self.error)),
+            OperationStatus::Invalid => GenmcHandleResult::Invalid,
+            _ => unreachable!("unexpected OperationStatus"),
+        }
+    }
+}
+
+impl ReadModifyWriteResult {
+    pub fn into_genmc_result(self) -> GenmcHandleResult<(GenmcScalar, GenmcScalar, bool)> {
```
RalfJung, line 107: I'm not a fan of how this conversion into a tuple loses the field names that distinguish old and new value. So I don't think for RMWs this is a win in its current form.

src/concurrency/genmc/mod.rs
```diff
@@ -820,12 +837,16 @@ impl GenmcCtx {
...
+                // FIXME(genmc): error handling
+                throw_ub_format!(
+                    "{}",
+                    store_result.error.as_ref().unwrap().to_string_lossy()
+                ),
+            OperationStatus::Ok => {}
+            status => unreachable!("unexpected OperationStatus: {status:?}"),
         }
         // Miri will always write non-atomic stores to memory. Make sure GenMC agrees with that.
         assert!(store_result.is_coherence_order_maximal_write);
```
RalfJung, line 852: Doesn't this fail to compile now since that field does not exist on NonAtomicStore?

src/concurrency/genmc/mod.rs
```diff
@@ -116,7 +116,7 @@ impl GlobalState {
 /// Some state is reset between each execution in the same run.
 pub struct GenmcCtx {
     /// Handle to the GenMC model checker.
-    handle: RefCell<UniquePtr<MiriGenmcShim>>,
+    handle: RefCell<UniquePtr<MiriGenMCInterface>>,
```
RalfJung, line 119: Is `handle` truly the best name for this? We often use "handle" as a verb in this code, so also using it as a noun can become confusing.
Would it make sense to just call this `genmc`?

RalfJung (commented):
Thanks! Some nice refactors. :)
@rustbot author

Comments for the commit messages:

> This made valid field combinations implicit and documented only in
comments.

FWIW they are still only documented in comments, right?

> `Ext` suggests an "extension"

Yeah, that's exactly what it is -- it's kind of like an extension trait in Rust.
But the new names are also fine 🤷 .

> provides compile-time exhaustiveness checking

There seems to always be an `UNREACHABLE` fallback. So is this really compile-time exhaustiveness checking? Looks more like runtime exhaustiveness checking.

> genmc/build,cpp: Flatten genmc-sys/cpp/src/ folder

This commit also seems to do some of the renaming of MiriGenmcShim already...

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4959/b9968c94af889723c98255c7c357f064d3f414eb..5f76116e70bdfb0a3cfdc5b0691b6efa4a29c713)*

michaliskok:
> Thanks! Some nice refactors. :) @rustbot author
> 
> Comments for the commit messages:
> 
> > This made valid field combinations implicit and documented only in
> > comments.
> 
> FWIW they are still only documented in comments, right?

Yes; fixed the commit message.

> > `Ext` suggests an "extension"
> 
> Yeah, that's exactly what it is -- it's kind of like an extension trait in Rust. But the new names are also fine 🤷 .

I dropped this commit to reduce the diff then.

> > provides compile-time exhaustiveness checking
> 
> There seems to always be an `UNREACHABLE` fallback. So is this really compile-time exhaustiveness checking? Looks more like runtime exhaustiveness checking.

Good catch --- I removed the `const auto` fallback.

> > genmc/build,cpp: Flatten genmc-sys/cpp/src/ folder
> 
> This commit also seems to do some of the renaming of MiriGenmcShim already...

I split the changed into commits post-hoc; some hiccups are because of that. Should be cleaner now.

> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4959/b9968c94af889723c98255c7c357f064d3f414eb..5f76116e70bdfb0a3cfdc5b0691b6efa4a29c713)_

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
michaliskok: Indeed, having visit subsume control flow is tricky...

genmc-sys/src/lib.rs
michaliskok: I didn't want to have both `OperationStatus` and `bool` discriminators for CASes and mutexes --- the point of this commit was to get rid of the boolean discriminators.

In a subsequent commit, all operations (including CASes/mutexes) were made to return ` GenmcHandlerResult<T>`, which does use `OperationStatus` and then an operation-specific payload. What do you think about that?

genmc-sys/cpp/src/MiriInterface/Exploration.cpp
michaliskok: Should've been in the previous one. Fixed.

genmc-sys/src/lib.rs
michaliskok: Renamed to `GenmcHandlerResult` so that it keeps the link to GenMC's `HandleResult`. I'd prefer not to name this `GenmcResult` to avoid confusing it with GenMC's `VerificationResult`.

genmc-sys/src/lib.rs
michaliskok: I replaced the tuple with `RmwOutcome { old, new, is_co_max }`.

src/concurrency/genmc/mod.rs
michaliskok: The PR as a whole compiled, but indeed this commit was not properly split. Fixed.

src/concurrency/genmc/mod.rs
michaliskok: Done.

genmc-sys/src/lib.rs
```diff
@@ -5,6 +5,47 @@ pub use cxx::UniquePtr;
 
 pub use self::ffi::*;
 
+/// The result of a GenMC handle operation, after converting the raw C++ result struct into an
+/// idiomatic Rust type.
+#[must_use]
+pub enum GenmcHandlerResult<T> {
+    /// This execution should be dropped.
+    Invalid,
```
RalfJung, line 13: This corresponds to `GenmcSkip`, right? Feels like we should have more consistent naming. I guess you call this "invalid" everywhere inside GenMC so maybe let's rename `GenmcSkip` to `GenmcInvalid` (with an explanatory comment)?

genmc-sys/src/lib.rs
RalfJung: > In a subsequent commit, all operations (including CASes/mutexes) were made to return  GenmcHandlerResult<T>, which does use OperationStatus and then an operation-specific payload. What do you think about that?

That does make the success mode part of the payload so :+1:

genmc-sys/cpp/src/Exploration.cpp
RalfJung:
```suggestion
    ERROR_ON(!ret_val, "Unimplemented: mutex lock load returned unexpected result.");
```

genmc-sys/cpp/src/Exploration.cpp
```diff
@@ -100,19 +100,24 @@ void MiriGenmcShim::handle_assume_block(ThreadId thread_id, AssumeType assume_ty
...
-    return LoadResultExt::from_value(*ret_val);
+    return std::visit(
+        overloaded {
+            [](const VerificationError& err) { return LoadResultExt::from_error(format_error(err)); },
+            [](const Invalid&) { return LoadResultExt::from_invalid(); },
+            [](const SVal& v) { return LoadResultExt::from_value(v); },
+            [](const Reset&) {
+                UNREACHABLE();
+                return LoadResultExt::from_invalid();
+            },
```
RalfJung, line 114: I'm confused by these branches... if they are unreachable then why do we even have them?

genmc-sys/cpp/src/Exploration.cpp
```diff
@@ -457,25 +494,31 @@ auto MiriGenmcShim::handle_mutex_lock(ThreadId thread_id, uint64_t address, uint
...
-    ERROR_ON(!is_co_max, "Unimplemented: mutex_try_lock store returned unexpected result.");
-    return MutexLockResultExt::ok(true);
+    // was the co-maximal store.
+    return std::visit(
+        overloaded {
+            [](const VerificationError& err) {
+                return MutexLockResultExt::from_error(format_error(err));
+            },
+            [](const GenMCDriver::Invalid&) { return MutexLockResultExt::from_invalid(); },
+            [](bool) { return MutexLockResultExt::acquired(); },
```
RalfJung: So we just ignore the `bool` here?

src/concurrency/genmc/mod.rs
```diff
@@ -558,10 +556,12 @@ impl GenmcCtx {
             genmc_size,
             alignment.bytes(),
         );
-        if let Some(_error) = malloc_result.error.as_ref() {
-            throw_exhaust!(AddressSpaceFull);
-        }
-        let chosen_address = malloc_result.address;
+        // FIXME(genmc): error handling
```
RalfJung: We are handling the error here... what is there to FIXME?

genmc-sys/src/lib.rs
```diff
@@ -5,6 +5,47 @@ pub use cxx::UniquePtr;
 
 pub use self::ffi::*;
 
+/// The result of a GenMC handle operation, after converting the raw C++ result struct into an
+/// idiomatic Rust type.
+#[must_use]
+pub enum GenmcHandlerResult<T> {
+    /// This execution should be dropped.
```
```suggestion
    /// This execution should be dropped/skipped.
```

src/concurrency/genmc/mod.rs
RalfJung, line 1: It seems like the only thing we ever do with a `GenmcHandlerResult` is
```rust
        let outcome = match rmw_result.into_genmc_result() {
            GenmcHandlerResult::Invalid => throw_machine_stop!(TerminationInfo::GenmcSkip),
            GenmcHandlerResult::Error(e) => throw_ub_format!("{e}"),
            GenmcHandlerResult::Ok(outcome) => outcome,
        };
```
So... let's try to extract that pattern into a function?
```rust
fn get_outcome<T>(result: GenmcHandlerResult<T>) -> InterpResult<'_, T> {
  ...
}
```
and then you can use `get_outcome(rmw_result.into_genmc_result())?` etc in the actual handlers, reducing the duplication quite a bit.

RalfJung (commented):
I have a few more questions and comments :)
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4959/c3ec75a6fd02f90fc842e0b1fbe657cc1d9c5257..d866669d5c3cd2fc79113a053b0036c4e061bf42)*

genmc-sys/cpp/src/Exploration.cpp
michaliskok: All GenMC handlers return the same type (`T | VerificationError | Reset | Invalid`). Some of these alternatives are not possible for certain handlers. Also, alternatives that are possible in general are sometimes not possible in Miri.

So we do need to handle them, I just dropped the `const auto &` version to get compile-time exhaustiveness.

genmc-sys/cpp/src/Exploration.cpp
michaliskok: Added a check.

src/concurrency/genmc/mod.rs
michaliskok: Most GenMC result handling had such comments. I think it was `improve error handling` here (?), but indeed this is one of the parts we handle rather well.

Now that handling is centralized in the new `get_outcome` you suggested, I added a single FIXME there.

src/concurrency/genmc/mod.rs
michaliskok: Looks much better, thanks! 

Note that `handle_alloc` doesn't use this because it throws a different error. I assumed it's an overkill to parameterize the error arm for a single caller.

src/concurrency/genmc/shims.rs
```diff
@@ -167,16 +166,13 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             genmc_ctx.active_thread_genmc_tid(&this.machine),
             mutex.ptr().addr().bytes(),
             mutex.layout.size.bytes(),
         );
-        if let Some(error) = result.error.as_ref() {
-            // FIXME(genmc): improve error handling.
-            throw_ub_format!("{}", error.to_string_lossy());
-        }
-        // NOTE: We don't write anything back to Miri's memory where the Mutex is located, that state is handled only by GenMC.}
+        get_outcome(result.into_genmc_result())?;
```
RalfJung: This returns a `bool` that we ignore. What is that `bool`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4959/c7d7d96c3dcfcb2743693a17e2714180f558a9e8..e259b2ea06ca580c338c368959f3bfa0bc72655d)*

RalfJung: It's apparently a  `is_co_max`.

RalfJung:
LGTM, thanks a lot :)

PR #4961, opened 2026-04-14
genmc tests: add -Zmiri-genmc by default

RalfJung:


RalfJung:
@michaliskok does this seem reasonable to you?

michaliskok:
> @michaliskok does this seem reasonable to you?

Yes. The only issue I see is that someone reading the tests might not realize that all these imply `-Zmiri-genmc`, but I guess that's fine.

PR #4963, opened 2026-04-14
Implement `_mm512_permutex2var_epi64` shim

elichai:
This implements the `_mm512_permutex2var_epi64` intrinsic, which is `vpermi2var.q.512`, which permutes 64-bit elements from two input vectors according to an index vector, producing a destination vector.

### DIFF
--- a/src/shims/x86/avx512.rs
+++ b/src/shims/x86/avx512.rs
@@ -3,7 +3,9 @@ use rustc_middle::ty::Ty;
 use rustc_span::Symbol;
 use rustc_target::callconv::FnAbi;
 
-use super::{packssdw, packsswb, packusdw, packuswb, permute, pmaddbw, pmaddwd, psadbw, pshufb};
+use super::{
+    packssdw, packsswb, packusdw, packuswb, permute, permute2, pmaddbw, pmaddwd, psadbw, pshufb,
+};
 use crate::*;
 
 impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
@@ -110,6 +112,13 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
 
                 permute(this, left, right, dest)?;
+            }
+            // Used to implement the _mm512_permutex2var_epi64 intrinsic.
+            "vpermi2var.q.512" => {
+                let [left, indices, right] =
+                    this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
+
+                permute2(this, left, indices, right, dest)?;
             }
             // Used to implement the _mm512_shuffle_epi8 intrinsic.
             "pshuf.b.512" => {
--- a/src/shims/x86/mod.rs
+++ b/src/shims/x86/mod.rs
@@ -1107,6 +1107,54 @@ fn permute<'tcx>(
     interp_ok(())
 }
 
+/// Shuffle elements from *two* source registers (`left` and `right`) using
+/// the corresponding index in `indices`, and store the results in `dest`.
+///
+/// For a vector with `N` lanes, the low `log2(N)` bits of each index select a
+/// lane within a source vector. Bit `log2(N)` selects the source vector (`0` =>
+/// `left`, `1` => `right`), and all higher bits are ignored.
+/// Equivalently, lane `i` of the result is copied from
+/// `src[indices[i] & (N - 1)]` where
+/// `src = if indices[i] & N == 0 { left } else { right }`.
+///
+/// <https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_permutex2var_epi64>
+fn permute2<'tcx>(
+    ecx: &mut crate::MiriInterpCx<'tcx>,
+    left: &OpTy<'tcx>,
+    indices: &OpTy<'tcx>,
+    right: &OpTy<'tcx>,
+    dest: &MPlaceTy<'tcx>,
+) -> InterpResult<'tcx, ()> {
+    let (left, left_len) = ecx.project_to_simd(left)?;
+    let (indices, indices_len) = ecx.project_to_simd(indices)?;
+    let (right, right_len) = ecx.project_to_simd(right)?;
+    let (dest, dest_len) = ecx.project_to_simd(dest)?;
+
+    assert_eq!(dest_len, left_len);
+    assert_eq!(dest_len, indices_len);
+    assert_eq!(dest_len, right_len);
+
+    // Use the low bits to select a lane within either input vector, and the next bit to
+    // choose between the two vectors.
+    assert!(dest_len.is_power_of_two());
+    let lane_mask = u128::from(dest_len).strict_sub(1);
+    let vector_select_bit = u128::from(dest_len);
+
+    for i in 0..dest_len {
+        let dest = ecx.project_index(&dest, i)?;
+        let index_place = ecx.project_index(&indices, i)?;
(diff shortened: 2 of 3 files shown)


elichai:
Improved the test code, and I've ran it on a real x86-64 machine(AMD EPYC 9B45) with AVX512:
`rustc ./tests/pass/shims/x86/intrinsics-x86-avx512.rs -C target-feature=+avx512f,+avx512vl,+avx512bw,+avx512bitalg,+avx512vpopcntdq,+avx512vnni -o avx512-test && ./avx512-test`

elichai:
I also have an implementation of `_mm512_permutex2var_epi8` built on top of this `permute2` function, should I add it here in a separate commit or open a new PR after this merges?
Thanks!

RalfJung:
Please slow down a little, my review queue is already overloaded. :)

elichai:
> Please slow down a little, my review queue is already overloaded. :)

Hehe yeah I'm sorry, don't worry, I only needed these 3 intrinsics to run my code with miri :) 
It's very cool that miri supports so many shims <3

src/shims/x86/mod.rs
```diff
@@ -1107,6 +1107,50 @@ fn permute<'tcx>(
...
 }
 
+/// Shuffle elements from `left` and `right` using the corresponding index in `indices`,
+/// and store the results in `dest`.
+///
+/// For each lane, the lower `log2(N)` bits select a lane within the chosen input vector,
+/// where `N` is the vector length. The next bit selects between `left` and `right`.
+///
+/// <https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_permutex2var_epi64>
+fn permute2<'tcx>(
```
RalfJung, line 1121: That name is not so self-explaining. ;)

Why do we need two different `permute`  functions?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4963/0a272829b77a88fa85a5a3a8d9f0eeda793687c2..045adf10fef4a1b7c1ad50ac40eb3d16fc01b93a)*

src/shims/x86/mod.rs
```diff
@@ -1107,6 +1107,50 @@ fn permute<'tcx>(
     interp_ok(())
 }
 
+/// Shuffle elements from `left` and `right` using the corresponding index in `indices`,
+/// and store the results in `dest`.
+///
+/// For each lane, the lower `log2(N)` bits select a lane within the chosen input vector,
+/// where `N` is the vector length. The next bit selects between `left` and `right`.
```
RalfJung: And the remaining bits? Are they ignored?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4963/0a272829b77a88fa85a5a3a8d9f0eeda793687c2..045adf10fef4a1b7c1ad50ac40eb3d16fc01b93a)*

src/shims/x86/mod.rs
elichai: Sorry, yeah it's similar to the other `permute` function but accepts 2 source registers instead of 1 (allowing you to choose lanes from each of the 2 source registers into the destination)

We could modify the `fn permute` to accept `sources: &[&OpTy<'tcx>]` and generalize over multiple sources, but I don't think Intel has any shuffling instruction with more than 2 source registers

From Intel docs:
Regular permute is:
```
__m512i _mm512_permutexvar_epi64 (__m512i idx, __m512i a)
#include <immintrin.h>
Instruction: vpermq zmm, zmm, zmm
CPUID Flags: AVX512F
Description: Shuffle 64-bit integers in a across lanes using the corresponding index in idx, and store the results in dst.
Operation:
FOR j := 0 to 7
	i := j*64
	id := idx[i+2:i]*64
	dst[i+63:i] := a[id+63:id]
ENDFOR
dst[MAX:512] := 0
```

permute2:
```
__m512i _mm512_permutex2var_epi64 (__m512i a, __m512i idx, __m512i b)
#include <immintrin.h>
Instruction: vpermi2q zmm, zmm, zmm
             vpermt2q zmm, zmm, zmm
CPUID Flags: AVX512F
Description: Shuffle 64-bit integers in a and b across lanes using the corresponding selector and index in idx, and store the results in dst.
Operation
FOR j := 0 to 7
	i := j*64
	off := idx[i+2:i]*64
	dst[i+63:i] := idx[i+3] ? b[off+63:off] : a[off+63:off]
ENDFOR
dst[MAX:512] := 0
```

src/shims/x86/mod.rs
```diff
@@ -1107,6 +1107,50 @@ fn permute<'tcx>(
     interp_ok(())
 }
 
+/// Shuffle elements from `left` and `right` using the corresponding index in `indices`,
```
RalfJung:
```suggestion
/// Shuffle elements from *two* source registers (`left` and `right`) using the corresponding index in `indices`,
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4963/0a272829b77a88fa85a5a3a8d9f0eeda793687c2..045adf10fef4a1b7c1ad50ac40eb3d16fc01b93a)*

src/shims/x86/mod.rs
RalfJung: Ah I see. Overeager generalization would probably make the code quite hard to follow so let's not do that. But the doc comment if this function should make it clear why it has this name.

src/shims/x86/avx512.rs
```diff
@@ -111,6 +113,13 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
                 permute(this, left, right, dest)?;
             }
+            // Used to implement the _mm512_permutex2var_epi64 intrinsic.
+            "vpermi2var.q.512" => {
+                let [left, indices, right] =
+                    this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
+
+                permute2(this, left, indices, right, dest)?;
```
RalfJung, line 121: That argument order is certainly a Choice. Wow. But I guess it's least confusing to be consistent with the intrinsic here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4963/0a272829b77a88fa85a5a3a8d9f0eeda793687c2..045adf10fef4a1b7c1ad50ac40eb3d16fc01b93a)*

src/shims/x86/mod.rs
elichai: AFAIU from https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_permutex2var_epi64&expand=4250&ig_expand=5002 it seems like bits [63:4] are ignored

src/shims/x86/mod.rs
elichai: @RalfJung Should I rename to `permute_pair`? or `permute_two_sources`?
Or we can rename `permute` to `permvar` and `permute2` to `vpermi2`? not sure if it's better but it seems more consistent with existing things like `pshufb` and `vpdpbusd`

Also, `vpdpbusd` is in a different file (`src/shims/x86/avx512.rs`) I can move `permute2` there if it's preferred

RalfJung: `premute2` isn't bad, but the comment should make it clear why that's the name -- see the suggested edit to the doc comment.

> existing things like pshufb and vpdpbusd

Those are meaningless random sequences of letters as far as I am concerned so no please let's not use those. ;)

src/shims/x86/mod.rs
RalfJung: This comment is still unresolved. Please extend the doc comment to explicitly say that the remaining bits are ignored.

@rustbot author

RalfJung:
Thanks, LGTM :)

PR #4964, opened 2026-04-15
Support running no_core Linux programs

### LINKED ISSUE #4512: Miri should not assume paths exist by string in `core`
I'm working on a custom core called `corps` which breaks some compatibility with Rust. While I realize I'm way off in the weeds and don't expect much support from things, I ran into an issue with `miri` where it resolves `core::ffi::c_ulong` as well as `core::ascii::escape_default` unconditionally, which causes miri to not work in my custom core.

"Fixing" the issue (temporarily for me) was quite a simple fix, and if miri was much further from working in my environment I wouldn't really bother making an issue.

The issues are in `chunk_size` in `src/tools/miri/src/concurrency/cpu_affinity.rs` which uses `core::ffi::c_ulong`, and the "sentinel" value checked in `src/tools/miri/src/eval.rs` of `core::ascii::escape_default`.

This is the diff I used to make miri work in my environment. Obviously, this is a hack workaround where I just remove the check if libcore is present in eval, and hard-code a `c_ulong` size. But I imagine that there would be a way to do these things through lang items which are accessible regardless of the path names.

```diff
diff --git a/src/tools/miri/src/concurrency/cpu_affinity.rs b/src/tools/miri/src/concurrency/cpu_affinity.rs
index 9583de5a483..0b6cd95a10c 100644
--- a/src/tools/miri/src/concurrency/cpu_affinity.rs
+++ b/src/tools/miri/src/concurrency/cpu_affinity.rs
@@ -32,10 +32,11 @@ pub fn new<'tcx>(cx: &impl LayoutOf<'tcx>, cpu_count: u32) -> Self {
         this
     }
 
-    pub fn chunk_size<'tcx>(cx: &impl LayoutOf<'tcx>) -> u64 {
+    pub fn chunk_size<'tcx>(_cx: &impl LayoutOf<'tcx>) -> u64 {
         // The actual representation of the CpuAffinityMask is [c_ulong; _].
-        let ulong = helpers::path_ty_layout(cx, &["core", "ffi", "c_ulong"]);
-        ulong.size.bytes()
+        //let ulong = helpers::path_ty_layout(cx, &["core", "ffi", "c_ulong"]);
+        //ulong.size.bytes()
+        8
     }
 
     fn set<'tcx>(&mut self, cx: &impl LayoutOf<'tcx>, cpu: usize) {
diff --git a/src/tools/miri/src/eval.rs b/src/tools/miri/src/eval.rs
index 4c531a8d1f5..ca179a8a486 100644
--- a/src/tools/miri/src/eval.rs
+++ b/src/tools/miri/src/eval.rs
@@ -9,7 +9,7 @@
 
 use rustc_abi::ExternAbi;
 use rustc_data_structures::fx::{FxHashMap, FxHashSet};
-use rustc_hir::def::Namespace;
+//use rustc_hir::def::Namespace;
 use rustc_hir::def_id::DefId;
 use rustc_middle::ty::layout::{HasTyCtxt, HasTypingEnv, LayoutCx};
 use rustc_middle::ty::{self, Ty, TyCtxt};
@@ -329,6 +329,7 @@ pub fn create_ecx<'tcx>(
         MiriMachine::new(config, layout_cx, genmc_ctx),
     );
 
+    /*
     // Make sure we have MIR. We check MIR for some stable monomorphic function in libcore.
     let sentinel =
         helpers::try_resolve_path(tcx, &["core", "ascii", "escape_default"], Namespace::ValueNS);
@@ -337,7 +338,7 @@ pub fn create_ecx<'tcx>(
             "the current sysroot was built without `-Zalways-encode-mir`, or libcore seems missing.\n\
             Note that directly invoking the `miri` binary is not supported; please use `cargo miri` instead."
         );
-    }
+    }*/
```

saethlin:
Fixes https://github.com/rust-lang/miri/issues/4512

### DIFF
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -962,8 +962,11 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let old_thread_id = this.machine.threads.set_active_thread_id(new_thread_id);
 
         // The child inherits its parent's cpu affinity.
-        if let Some(cpuset) = this.machine.thread_cpu_affinity.get(&old_thread_id).cloned() {
-            this.machine.thread_cpu_affinity.insert(new_thread_id, cpuset);
+        // Skips this if `machine.thread_cpu_affinity` is not initialized.
+        if let Some(thread_cpu_affinity) = &mut this.machine.thread_cpu_affinity
+            && let Some(cpuset) = thread_cpu_affinity.get(&old_thread_id).cloned()
+        {
+            thread_cpu_affinity.insert(new_thread_id, cpuset);
         }
 
         // Perform the function pointer load in the new thread frame.
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -21,6 +21,7 @@ use rustc_target::spec::Os;
 use crate::concurrency::GenmcCtx;
 use crate::concurrency::thread::TlsAllocAction;
 use crate::diagnostics::report_leaks;
+use crate::helpers::is_no_core;
 use crate::shims::{global_ctor, tls};
 use crate::*;
 
@@ -289,14 +290,20 @@ pub fn create_ecx<'tcx>(
         MiriMachine::new(config, layout_cx, genmc_ctx),
     );
 
-    // Make sure we have MIR. We check MIR for some stable monomorphic function in libcore.
-    let sentinel =
-        helpers::try_resolve_path(tcx, &["core", "ascii", "escape_default"], Namespace::ValueNS);
-    if !matches!(sentinel, Some(s) if tcx.is_mir_available(s.def.def_id())) {
-        tcx.dcx().fatal(
-            "the current sysroot was built without `-Zalways-encode-mir`, or libcore seems missing.\n\
-            Note that directly invoking the `miri` binary is not supported; please use `cargo miri` instead."
+    // Make sure we have MIR. We check MIR for some stable monomorphic function in libcore, except
+    // if the current crate is #![no_core] it's fine to be missing the usual items from libcore.
+    if !is_no_core(tcx) {
+        let sentinel = helpers::try_resolve_path(
+            tcx,
+            &["core", "ascii", "escape_default"],
+            Namespace::ValueNS,
         );
+        if !matches!(sentinel, Some(s) if tcx.is_mir_available(s.def.def_id())) {
+            tcx.dcx().fatal(
+                "the current sysroot was built without `-Zalways-encode-mir`, or libcore seems missing.\n\
+                Note that directly invoking the `miri` binary is not supported; please use `cargo miri` instead."
+            );
+        }
     }
 
     // Compute argc and argv from `config.args`.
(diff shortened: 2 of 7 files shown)


src/machine.rs
```diff
@@ -731,7 +732,8 @@ impl<'tcx> MiriMachine<'tcx> {
         );
         let threads = ThreadManager::new(config);
         let mut thread_cpu_affinity = FxHashMap::default();
-        if matches!(&tcx.sess.target.os, Os::Linux | Os::FreeBsd | Os::Android) {
+        if matches!(&tcx.sess.target.os, Os::Linux | Os::FreeBsd | Os::Android) && !is_no_core(tcx)
+        {
             thread_cpu_affinity
                 .insert(threads.active_thread(), CpuAffinityMask::new(&layout_cx, config.num_cpus));
```
RalfJung, line 738: This means that if the program later calls `sched_getaffinity` it may get nonsensical errors about threads not existing.

Might be better to turn this into an `Option<FxHashMap>` so that we can leave it uninitialized and give proper errors in the shims that need it?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4964/82be54a310a37db9d54358260cddb43126162470..9837d3d6c9753e729560ca1d9fc6a3300b5b26d8)*

src/concurrency/thread.rs
```diff
@@ -962,8 +962,10 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let old_thread_id = this.machine.threads.set_active_thread_id(new_thread_id);
 
         // The child inherits its parent's cpu affinity.
```
RalfJung, line 964:
```suggestion
        // The child inherits its parent's cpu affinity.
        // Skips this if `machine.thread_cpu_affinity` is not initialized.
```

src/eval.rs
```diff
@@ -289,14 +290,20 @@ pub fn create_ecx<'tcx>(
...
 
-    // Make sure we have MIR. We check MIR for some stable monomorphic function in libcore.
-    let sentinel =
-        helpers::try_resolve_path(tcx, &["core", "ascii", "escape_default"], Namespace::ValueNS);
-    if !matches!(sentinel, Some(s) if tcx.is_mir_available(s.def.def_id())) {
-        tcx.dcx().fatal(
-            "the current sysroot was built without `-Zalways-encode-mir`, or libcore seems missing.\n\
-            Note that directly invoking the `miri` binary is not supported; please use `cargo miri` instead."
+    // Make sure we have MIR. We check MIR for some stable monomorphic function in libcore, except
+    // if the current crate is #![no_core] it's fine to be missing the usual items from libcore.
```
RalfJung:
```suggestion
    // Make sure we have MIR. We check MIR for some stable monomorphic function in libcore. However,
    // if the current crate is #![no_core] it's fine to be missing the usual items from libcore.
```

RalfJung (commented):
Just some comment nits. :)
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4964/55962eb26a4733e5db315da892df18c142d008cf..27f38058558c3ed859b38b94e62c824f97b608d1)*

RalfJung:
r=me after squashing. :)

gamozolabs:
Thanks so much! This is exciting. Might be useful to some of the small internal rustc tests/snippits that are no core as well!

PR #4967, opened 2026-04-17
remove vec_unique.default.stderr

quiode:
This very small PR removes the file `tests/pass/tree_borrows/vec_unique.default.stderr`. It has no matching test and seems to be forgotten when the test was removed / changed.


RalfJung:
Good catch, thanks!

PR #4973, opened 2026-04-20
Add epoll integration for network sockets

WhySoBad:
Hi,

This pull request adds integration for Miri's epoll shim to the TCP socket shim. 

Due to the platform specific differences of mio's poll, the blocking I/O manager had to be refactored to a high degree. Instead of having sources only registered in the poll when we're currently interested in an event, we now have the sources registered in the poll with all available interests for their entire lifespan. We now also store the current readiness for every source in the blocking I/O manager.
When for example a thread should be blocked until some I/O interest is fulfilled on a source, we just add the receiver to the blocking I/O manager without changing anything in the poll itself. In every call to `poll` we now return all interest receivers whose interests are currently fulfilled -- no longer only those which are newly fulfilled.   
Once we hit an `EWOULDBLOCK` in a source operation (e.g. read, write, accept, ...) we need to falsify the corresponding readiness in the blocking I/O manager to keep it up to date. 

With this refactor, the actual epoll implementation is pretty straightforward -- we just need to return the readiness stored in the blocking I/O manager from the `epoll_active_events` method of host-backed source file descriptions.

### DIFF
--- a/src/concurrency/blocking_io.rs
+++ b/src/concurrency/blocking_io.rs
@@ -1,12 +1,15 @@
+use std::cell::RefMut;
 use std::collections::BTreeMap;
 use std::io;
+use std::ops::BitOrAssign;
 use std::time::Duration;
 
 use mio::event::Source;
 use mio::{Events, Interest, Poll, Token};
-use rustc_data_structures::fx::FxHashMap;
 
-use crate::shims::{FdId, FileDescriptionRef};
+use crate::shims::{
+    EpollEvalContextExt, FdId, FileDescription, FileDescriptionRef, WeakFileDescriptionRef,
+};
 use crate::*;
 
 /// Capacity of the event queue which can be polled at a time.
@@ -14,23 +17,112 @@ use crate::*;
 /// this value can be set rather low.
 const IO_EVENT_CAPACITY: usize = 16;
 
-/// Trait for values that contain a mio [`Source`].
-pub trait WithSource {
+/// Trait for file descriptions that contain a mio [`Source`].
+pub trait SourceFileDescription: FileDescription {
     /// Invoke `f` on the source inside `self`.
     fn with_source(&self, f: &mut dyn FnMut(&mut dyn Source) -> io::Result<()>) -> io::Result<()>;
+
+    /// Get a mutable reference to the readiness of the source.
+    fn get_readiness_mut(&self) -> RefMut<'_, BlockingIoSourceReadiness>;
 }
 
-/// An interest receiver defines the action that should be taken when
-/// the associated [`Interest`] is fulfilled.
-#[derive(Debug, Hash, PartialEq, Clone, Copy, Eq, PartialOrd, Ord)]
-pub enum InterestReceiver {
-    /// The specified thread should be unblocked.
-    UnblockThread(ThreadId),
+/// An I/O interest for a blocked thread. Note that all threads are always considered
+/// to be interested in "error" events.
+#[derive(Debug, Clone, Copy)]
+pub enum BlockingIoInterest {
+    /// The blocked thread is interested in [`Interest::READABLE`].
+    Read,
+    /// The blocked thread is interested in [`Interest::WRITABLE`].
+    Write,
+    /// The blocked thread is interested in [`Interest::READABLE`] and
+    /// [`Interest::WRITABLE`].
+    ReadWrite,
+}
+
+/// Struct reflecting the readiness of a source file description.
+#[derive(Debug)]
+pub struct BlockingIoSourceReadiness {
+    /// Boolean whether the source is currently readable.
+    pub readable: bool,
+    /// Boolean whether the source is currently writable.
+    pub writable: bool,
+    /// Boolean whether the read end of the source has been
+    /// closed.
+    pub read_closed: bool,
+    /// Boolean whether the write end of the source has been
+    /// closed.
+    pub write_closed: bool,
+    /// Boolean whether the source currently has an error.
+    pub error: bool,
+}
+
+impl BlockingIoSourceReadiness {
+    pub fn empty() -> Self {
+        Self {
+            readable: false,
+            writable: false,
+            read_closed: false,
+            write_closed: false,
+            error: false,
+        }
+    }
+
+    /// Check whether the current readiness fulfills the blocking I/O interest of
+    /// `interest`.
+    /// This function also returns `true` if the error readiness is set
+    /// even when the requested interest might not be fulfilled.
+    fn fulfills_interest(&self, interest: &BlockingIoInterest) -> bool {
+        match interest {
+            BlockingIoInterest::Read => self.readable || self.error,
+            BlockingIoInterest::Write => self.writable || self.error,
+            BlockingIoInterest::ReadWrite => self.readable || self.writable || self.error,
+        }
+    }
+}
+
+impl BitOrAssign for BlockingIoSourceReadiness {
+    fn bitor_assign(&mut self, rhs: Self) {
+        self.readable |= rhs.readable;
+        self.writable |= rhs.writable;
+        self.read_closed |= rhs.read_closed;
+        self.write_closed |= rhs.write_closed;
+        self.error |= rhs.error;
+    }
+}
+
+impl From<&mio::event::Event> for BlockingIoSourceReadiness {
+    fn from(event: &mio::event::Event) -> Self {
+        Self {
+            readable: event.is_readable(),
+            writable: event.is_writable(),
+            read_closed: event.is_read_closed(),
+            write_closed: event.is_write_closed(),
+            error: event.is_error(),
+        }
+    }
+}
+
+struct BlockingIoSource {
+    /// The source file description which is registered into the poll.
+    /// We only store weak references such that source file descriptions
+    /// can be destroyed whilst they are registered. However, they are required
+    /// to deregister themselves when [`FileDescription::destroy`] is called.
+    fd: WeakFileDescriptionRef<dyn SourceFileDescription>,
+    /// The threads which are blocked on the I/O source, and the interest indicating
+    /// when they should be unblocked.
+    blocked_threads: BTreeMap<ThreadId, BlockingIoInterest>,
 }
 
 /// Manager for managing blocking host I/O in a non-blocking manner.
 /// We use [`Poll`] to poll for new I/O events from the OS for sources
 /// registered using this manager.
+///
+/// The semantics of this manager are that host I/O sources are registered
+/// to a [`Poll`] for their entire lifespan. Once host readiness events happen
+/// on a registered source, its internal epoll readiness gets updated -- even
+/// when the source isn't part of an active epoll instance. Also, for the entire
+/// lifespan of the source, threads can be added which should be unblocked
+/// once a certain [`BlockingIoSourceReadiness`] for an I/O source is satisfied.
 ///
 /// Since blocking host I/O is inherently non-deterministic, no method on this
 /// manager should be called when isolation is enabled. The only exception is
@@ -44,10 +136,9 @@ pub struct BlockingIoManager {
     /// This is not part of the state and only stored to avoid allocating a
     /// new buffer for every poll.
     events: Events,
-    /// Map from source ids to the actual sources and their registered receivers
-    /// together with their associated interests.
-    sources:
-        BTreeMap<FdId, (FileDescriptionRef<dyn WithSource>, FxHashMap<InterestReceiver, Interest>)>,
+    /// Map from source file description ids to the actual sources and their
+    /// blocked threads.
+    sources: BTreeMap<FdId, BlockingIoSource>,
 }
 
 impl BlockingIoManager {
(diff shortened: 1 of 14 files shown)


RalfJung:
Feedback was given IRL.
@rustbot author

WhySoBad:
I noticed that I wrote the review comments of today's meeting in the PR on my fork, but every of those comments should now be resolved anyways.

The `BlockingIoManager:poll` now updates the epoll readiness of all sources. I've still kept the `InterestReceiver`s even though we only need it for unblocking threads at the moment (maybe there will be something in the future which also needs to be added to the blocking I/O manager).
I've also renamed the `EpollEvents` struct to `EpollReadiness`. This is more in line with the new `BlockingIoSourceReadiness` struct of the blocking I/O manager. 

Also, as you requested, the epoll test cases should now fail/block indefinitely when we forget to register a socket or when we forget to remove the readiness after receiving EWOULDBLOCK.

@rustbot ready

RalfJung:
> The BlockingIoManager:poll now updates the epoll readiness of all sources. I've still kept the InterestReceivers even though we only need it for unblocking threads at the moment (maybe there will be something in the future which also needs to be added to the blocking I/O manager).

I prefer clean code over premature generalization. So unless we have at least a concrete idea for what other interest receivers there may be, I think it'd be less confusing to just hard-code for now that these are all blocked threads.

WhySoBad:
Okay, so I'll change it that we just register thread id's together with an interest. 
The unblocking can then also be done in the `poll`, what about the deregistering? Should we implicitly deregister the thread id's once we unblocked the thread in the blocking I/O manager (how it's on current `master`) or should it always happen explicitly in the unblock callback (how it's in this PR)?

RalfJung:
Both are fine for me.

WhySoBad:
> Both are fine for me.

I've now implemented it such that we always need to deregister the interests explicitly.

src/concurrency/blocking_io.rs
```diff
@@ -1,31 +1,109 @@
...
+            writable: false,
+            read_closed: false,
+            write_closed: false,
+            error: false,
+        }
+    }
+
+    pub fn fulfills_interest(&self, interest: &BlockingIoInterest) -> bool {
+        match interest {
+            BlockingIoInterest::Read => self.readable || self.error,
```
RalfJung, line 76: Is there some documentation somewhere saying that "error" sources should be "ready" for reads and writes?

src/concurrency/blocking_io.rs
```diff
@@ -1,31 +1,109 @@
...
+            write_closed: event.is_write_closed(),
+            error: event.is_error(),
+        }
+    }
+}
+
+struct BlockingIoSource {
+    /// The source file description which is registered into the poll.
+    fd: WeakFileDescriptionRef<dyn SourceFileDescription>,
+    /// The threads which are blocked on the I/O source.
```
RalfJung:
```suggestion
    /// The threads which are blocked on the I/O source, and the interest indicating
    /// when they should be unblocked.
```

src/concurrency/blocking_io.rs
```diff
@@ -1,31 +1,109 @@
...
+            read_closed: event.is_read_closed(),
+            write_closed: event.is_write_closed(),
+            error: event.is_error(),
+        }
+    }
+}
+
+struct BlockingIoSource {
+    /// The source file description which is registered into the poll.
+    fd: WeakFileDescriptionRef<dyn SourceFileDescription>,
```
RalfJung, line 110: Please add a comment explaining why this has to be a weak ref.

src/concurrency/blocking_io.rs
```diff
@@ -70,133 +147,170 @@ impl BlockingIoManager {
...
-                // Because we allow spurious wake-ups, we mark all interests as ready even
-                // though some may not have been fulfilled.
-                interests.keys().map(move |receiver| (*receiver, source.clone()))
+                let source = ecx
+                    .machine
+                    .blocking_io
+                    .sources
+                    .get(&fd_id)
+                    .expect("Source should be registered");
+                let fd = source.fd.clone().upgrade().expect(
```
RalfJung:
```suggestion
                let fd = source.fd.upgrade().expect(
```

src/concurrency/blocking_io.rs
```diff
@@ -70,133 +147,170 @@ impl BlockingIoManager {
...
+                    "Source file description shouldn't be destroyed whilst being registered",
+                );
+
+                assert_eq!(fd.id(), fd_id);
+                // Update the readiness of the source.
+                *fd.get_readiness_mut() |= BlockingIoSourceReadiness::from(event);
+                fd
```
RalfJung, line 204:
```suggestion
                // Put FD into `event_fds` list.
                fd
```

src/concurrency/blocking_io.rs
```diff
@@ -70,133 +147,170 @@ impl BlockingIoManager {
...
             })
             .collect::<Vec<_>>();
 
-        // Deregister all ready sources as we only want to receive one event per receiver.
-        ready.iter().for_each(|(receiver, source)| self.deregister(source.id(), *receiver));
+        for fd in event_fds.into_iter() {
```
RalfJung, line 210:
```suggestion
        // Process events: update readiness, unblock threads.
        for fd in event_fds.into_iter() {
```

src/concurrency/thread.rs
```diff
@@ -787,34 +788,49 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
...
+            // At least one thread doesn't have a timeout set, and is blocked on host I/O or is waiting on an
+            // epoll instance which contains a host source interest. Hence, we sleep indefinitely in the
+            // hope that eventually an I/O event happens.
             interp_ok(SchedulingAction::SleepAndWaitForIo(None))
         } else {
             throw_machine_stop!(TerminationInfo::GlobalDeadlock);
         }
     }
 
+    /// Get whether the provided thread is currently blocked on host I/O.
```
```suggestion
    /// Check whether the provided thread is currently blocked on host I/O.
```

src/concurrency/blocking_io.rs
```diff
@@ -70,133 +147,170 @@ impl BlockingIoManager {
     ///   specified duration.
     /// - If the timeout is [`None`] the poll blocks indefinitely until an event occurs.
     ///
-    /// Returns the interest receivers for all file descriptions which received an I/O event together
-    /// with the file description they were registered for.
-    pub fn poll(
-        &mut self,
+    /// Returns the list of [`ThreadId`]s which can be unblocked because their interests are currently fulfilled.
+    /// Note that the events are returned in a level-triggered way, which means that [`ThreadId`]s whose interests
+    /// were already fulfilled before the poll will be returned again.
```
RalfJung: Didn't we say this loop should simply also wake up the relevant threads rather than returning them? That's what I meant, anyway. :)

src/shims/unix/linux_like/epoll.rs
```diff
@@ -547,6 +562,20 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    fn has_epoll_host_interests(&self, epfd_id: &FdId) -> bool {
+        let this = self.eval_context_ref();
+        this.machine
+            .epoll_interests
+            .0
+            .iter()
+            .filter_map(|(fd_id, epolls)| {
+                epolls.binary_search_by_key(&epfd_id, |(id, _)| id).ok().map(|_| fd_id)
+            })
+            .any(|fd_id| this.machine.blocking_io.contains_source(fd_id))
```
RalfJung: This is a complete linear scan of all epoll interests. Ugh.

https://github.com/rust-lang/miri/pull/4991 removes `Copy` from `BlockReason` so you can more easily add a `FileDescriptorRef` instead of an `FdId`. (`FdId` should be used extremely rarely! It exists basically only to support epoll in identifying FDs.) You'll just have to add `PartialEq` and `Eq` for `FileDescriptorRef` (make it compare the IDs).

Then you can directly follow that ref and see if that epoll is interested in any host sources, and you don't have to scan *all epoll interests of the entire interpreter* all the time.

src/shims/unix/linux_like/epoll.rs
```diff
@@ -612,7 +641,7 @@ fn return_ready_list<'tcx>(
         for (key, interest) in interest_list.iter() {
             // Ensure this matches the latest readiness of this FD.
             // We have to do an FD lookup by ID for this. The FdNum might be already closed.
-            let fd = &ecx.machine.fds.fds.values().find(|fd| fd.id() == key.0).unwrap();
+            let fd = ecx.machine.fds.fds.values().find(|fd| fd.id() == key.0).unwrap().clone();
```
RalfJung: Why did you add a `clone` here?

src/shims/unix/socket.rs
```diff
@@ -1395,6 +1422,13 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
         let (stream, addr) = match listener.accept() {
             Ok(peer) => peer,
+            Err(e) if e.kind() == io::ErrorKind::WouldBlock => {
+                // We know that the source is not readable so we need to update it's readiness.
```
```suggestion
                // We know that the source is not readable so we need to update its readiness.
```

src/shims/unix/socket.rs
```diff
@@ -1395,6 +1422,13 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                socket.io_readiness.borrow_mut().readable = false;
+                this.update_epoll_active_events(socket.clone(), false)?;
```
```suggestion
                this.update_epoll_active_events(socket.clone(), /* force_edge */ false)?;
```
Same for all the other calls.

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -0,0 +1,299 @@
...
+        errno_check(libc::fcntl(client_sockfd, libc::F_SETFL, libc::O_NONBLOCK));
+    }
+
+    // Spawn the server thread.
+    let server_thread = thread::spawn(move || {
+        net::accept_ipv4(server_sockfd).unwrap();
+    });
+
+    // Yield to server thread to ensure that it's currently accepting.
+    thread::sleep(Duration::from_millis(10));
```
RalfJung: Wouldn't this test make more sense if the *client* goes first, i.e. the *server* has a `sleep`?

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -0,0 +1,299 @@
...
+
+    // Non-blocking connects always "fail" with EINPROGRESS.
+    let err = net::connect_ipv4(client_sockfd, addr).unwrap_err();
+    assert_eq!(err.kind(), ErrorKind::InProgress);
+
+    // Add client socket with WRITABLE interest to epoll.
+    epoll_ctl_add(epfd, client_sockfd, EPOLLOUT | EPOLLET | EPOLLERR).unwrap();
+
+    check_epoll_wait::<8>(epfd, &[Ev { events: EPOLLOUT, data: client_sockfd }], -1);
```
RalfJung, line 61:
```suggestion
    // Wait until we are done connecting.
    check_epoll_wait::<8>(epfd, &[Ev { events: EPOLLOUT, data: client_sockfd }], -1);
```

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -0,0 +1,299 @@
...
+        // Accepting should now trigger an EWOULDBLOCK.
+        let err = net::accept_ipv4(server_sockfd).unwrap_err();
+        assert_eq!(err.kind(), ErrorKind::WouldBlock);
+
+        // Ensure that the server is no longer readable because we just
+        // attempted to accept without an incoming connection.
+        assert_eq!(&current_epoll_readiness::<8>(server_sockfd, EPOLLIN | EPOLLET), &[]);
+    });
+
+    net::connect_ipv4(client_sockfd, addr).unwrap();
```
RalfJung, line 110: Would it make sense to sleep before doing this so the server is more likely to have to actually wait?

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -0,0 +1,299 @@
...
+    let server_thread = thread::spawn(move || {
+        let (peerfd, _) = net::accept_ipv4(server_sockfd).unwrap();
+        // `peerfd` is a blocking socket now. But that's okay, the client still does non-blocking
+        // reads/writes.
+
+        // Yield back to client so that it starts receiving before we start sending.
+        thread::sleep(Duration::from_millis(10));
+
+        unsafe {
+            errno_result(libc_utils::write_all_generic(
```
RalfJung: You could just use `write_all`, right? That'd be a bit simpler?

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -0,0 +1,299 @@
...
+
+    let mut bytes_received = 0;
+
+    while bytes_received != buffer.len() {
+        check_epoll_wait::<8>(epfd, &[Ev { events: EPOLLIN, data: client_sockfd }], -1);
+
+        // Receive until we get an EWOULDBLOCK or we read everything.
+        // We're only allowed to call `epoll_wait` again once we received
+        // an EWOULDBLOCK because otherwise we could deadlock.
+        while bytes_received != buffer.len() {
```
RalfJung: I think this doesn't have to be a nested loop. You can just put the `check_epoll_wait` into the `if err.kind() == ErrorKind::WouldBlock` arm, right?

tests/utils/libc.rs
```diff
@@ -204,6 +204,29 @@ pub mod epoll {
     pub fn check_epoll_wait_noblock<const N: usize>(epfd: i32, expected: &[Ev]) {
         check_epoll_wait::<N>(epfd, expected, 0);
     }
+
+    /// Query the current epoll readiness of a file descriptor.
+    /// This is done by creating a new epoll instance, adding the
+    /// fd to the epoll interests and then performing a zero-timeout
+    /// wait.
+    pub fn current_epoll_readiness<const N: usize>(fd: i32, interests: i32) -> Vec<Ev> {
```
RalfJung: I would design this API a bit differently: more like
```suggestion
    pub fn current_epoll_readiness<const N: usize>(fd: i32) -> c_int {
```
and it simply returns a bitmask indicating the current readiness.

RalfJung (commented):
I like it, very nice work. :)  But I do of course have some comments. ;)

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4973/98cb1d3d48fa7c8e97994050bfd69e8f3da2a1a7..ebd1e70fd918a47595119942173b028082f9ff44)*

src/concurrency/blocking_io.rs
WhySoBad: For mio we have this: https://docs.rs/mio/latest/mio/struct.Poll.html#readiness-operations
Should I link this here?

src/shims/unix/linux_like/epoll.rs
WhySoBad: I've also realized that the current implementation would break down for nested epoll instances. From what I can tell the current epoll shim supports nested epoll instances despite never checking for loops?

src/concurrency/blocking_io.rs
WhySoBad: It's now changed :)

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
WhySoBad: Yes, that makes more sense.

src/shims/unix/linux_like/epoll.rs
RalfJung: No we don't support nested epoll.

src/concurrency/blocking_io.rs
RalfJung: That does not say that read implies error. It says that read *may* imply error.

I guess this is actually mostly a decision for us -- should we wake up threads on an error signal? That probably makes sense, but then we should explain why.

src/concurrency/blocking_io.rs
```diff
@@ -1,31 +1,118 @@
...
+/// Trait for file descriptions that contain a mio [`Source`].
+pub trait SourceFileDescription: FileDescription {
     /// Invoke `f` on the source inside `self`.
     fn with_source(&self, f: &mut dyn FnMut(&mut dyn Source) -> io::Result<()>) -> io::Result<()>;
+
+    /// Get a mutable reference to the readiness of the source.
+    fn get_readiness_mut(&self) -> RefMut<'_, BlockingIoSourceReadiness>;
+}
+
+/// An I/O interest for a blocked thread.
```
RalfJung:
```suggestion
/// An I/O interest for a blocked thread. Note that all threads are always considered
/// to be interested in "error" events.
```

src/concurrency/blocking_io.rs
```diff
@@ -1,31 +1,118 @@
...
+        }
+    }
+
+    /// Check whether the current readiness fulfills the blocking I/O interest of
+    /// `interest`.
+    /// This function also returns `true` if the error readiness is set
+    /// even when the requested interest might not be fulfilled.
+    /// Usually, a source with the error readiness set won't become readable or writable
+    /// anymore. So, adding the error interest implicitly prevents that the thread is
+    /// blocked indefinitely in such cases.
```
RalfJung:
```suggestion
    /// anymore. By waking up the thread, it will retry its operation and then notice and
    /// process the error.
```

src/concurrency/blocking_io.rs
```diff
@@ -70,133 +156,195 @@ impl BlockingIoManager {
...
     /// - If the timeout is [`None`] the poll blocks indefinitely until an event occurs.
     ///
-    /// Returns the interest receivers for all file descriptions which received an I/O event together
-    /// with the file description they were registered for.
-    pub fn poll(
-        &mut self,
+    /// Unblocks all threads whose interests are currently fulfilled. Note that the events are handled
+    /// in a level-triggered way. This means that threads whose interests were already fulfilled
+    /// before the poll will be unblocked again. The unblock callback functions need to manually
+    /// remove the interest from the poll to prevent multiple unblocks.
```
RalfJung:
```suggestion
    /// The events also immediately get processed: threads get unblocked, and epoll readiness gets updated.
```
The rest of the comment (about de-registering) should be placed at `add_blocked_thread`; it is the callers of that method that need to be aware of this requirement.

src/concurrency/blocking_io.rs
```diff
@@ -70,133 +156,195 @@ impl BlockingIoManager {
...
+                            .expect(
+                                "Source file description shouldn't be destroyed whilst being registered",
+                            )
+                            .get_readiness_mut()
+                            .fulfills_interest(interest)
+                            .then_some(thread_id)
+                    })
+                    .copied()
             })
             .collect::<Vec<_>>();
```
RalfJung: Why do you re-iterate over everything here...?
I expected that you'd just iterate the events we just received and wake up all those threads.

src/concurrency/blocking_io.rs
```diff
@@ -70,133 +156,195 @@ impl BlockingIoManager {
...
-    /// verify the requested I/O interests are really ready and to register again if they're not.
-    ///
-    /// It's assumed that no interest is already registered for this source with the same reason!
-    pub fn register(
-        &mut self,
-        source_fd: FileDescriptionRef<dyn WithSource>,
-        receiver: InterestReceiver,
-        interest: Interest,
-    ) {
+    /// Get whether a source file description is currently registered in the
```
RalfJung:
```suggestion
    /// Return whether a source file description is currently registered in the
```

src/concurrency/blocking_io.rs
```diff
@@ -70,133 +156,195 @@ impl BlockingIoManager {
...
-            self.sources.get_mut(&source_id).expect("Source should be registered");
-
-        current_interests
-            .remove(&receiver)
-            .unwrap_or_else(|| panic!("Receiver should be registered for source"));
-
-        let Some(new_interest) = interest_union(current_interests) else {
-            // There are no longer any interests in this source.
-            // We can thus deregister the source from the poll.
+        // We cannot use the stored `fd` because it's only a `WeakFileDescriptionRef` and the
```
RalfJung: Use it for what? I am confused by this comment.

src/concurrency/blocking_io.rs
```diff
@@ -70,133 +156,195 @@ impl BlockingIoManager {
...
+    /// Add a new blocked thread to a registered source. The thread gets unblocked
+    /// once its [`BlockingIoInterest`] is fulfilled when calling
+    /// [`BlockingIoManager::poll_and_unblock`].
+    ///
+    /// Note that an error interest is implicitly added to `interest`.
+    /// This means that the thread will also be unblocked when the error
+    /// readiness gets set for the source even when the requested interest
+    /// might not be fulfilled. More about that can be read in the
+    /// [`BlockingIoSourceReadiness::fulfills_interest`] doc comment.
+    ///
```
RalfJung:
```suggestion
    /// might not be fulfilled.
```
That's a private function, it shouldn't show up in a public function's documentation.

src/concurrency/blocking_io.rs
```diff
@@ -70,133 +156,195 @@ impl BlockingIoManager {
...
+    /// Also, as the OS can always produce spurious wake-ups, it's the callers responsibility
+    /// to verify the requested I/O interests are really fulfilled when an event for this
+    /// thread is returned from [`BlockingIoManager::poll_and_unblock`].
+    ///
+    /// It's assumed that the thread of `thread_id` isn't already blocked on
+    /// the source with id `source_id` and that this source is currently
+    /// registered.
+    pub fn add_blocked_thread(
```
RalfJung: Why is this public? Isn't `block_thread_for_io` the entry point that the rest of Miri should use?

If you make this private then it also doesn't need to repeat the entire doc comment.

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -0,0 +1,299 @@
...
+    // call `epoll_wait` multiple times until we received everything.
+
+    // Add client socket with READABLE interest to epoll.
+    epoll_ctl_add(epfd, client_sockfd, EPOLLIN | EPOLLET | EPOLLERR).unwrap();
+
+    let mut bytes_received = 0;
+
+    // Receive until we get an EWOULDBLOCK or we read everything.
+    // We're only allowed to call `epoll_wait` again once we received
+    // an EWOULDBLOCK because otherwise we could deadlock.
```
```suggestion
```

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -0,0 +1,299 @@
...
+                buffer.as_mut_ptr().byte_add(bytes_received).cast(),
+                buffer.len() - bytes_received,
+                0,
+            ))
+        };
+
+        match read_result {
+            Ok(received) => bytes_received += received as usize,
+            Err(err) if err.kind() == ErrorKind::WouldBlock => {
+                check_epoll_wait::<8>(epfd, &[Ev { events: EPOLLIN, data: client_sockfd }], -1);
```
RalfJung, line 185:
```suggestion
                // Use epoll to block until there's data available again.
                check_epoll_wait::<8>(epfd, &[Ev { events: EPOLLIN, data: client_sockfd }], -1);
```

src/concurrency/blocking_io.rs
```diff
@@ -70,133 +156,195 @@ impl BlockingIoManager {
...
     ///
-    /// Returns the interest receivers for all file descriptions which received an I/O event together
-    /// with the file description they were registered for.
-    pub fn poll(
-        &mut self,
+    /// Unblocks all threads whose interests are currently fulfilled. Note that the events are handled
+    /// in a level-triggered way. This means that threads whose interests were already fulfilled
+    /// before the poll will be unblocked again. The unblock callback functions need to manually
+    /// remove the interest from the poll to prevent multiple unblocks.
+    pub fn poll_and_unblock<'tcx>(
```
RalfJung: This does "poll and unblock and update readiness"... I think just `poll` is fine as the name. Or maybe `poll_and_process_events` or so.

RalfJung (commented):
We're getting there. :)  I'm confused by some of the code in the main `poll` function though.
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4973/0853747bf78b8b6e314dcdf14ade9c83a6c54f3e..ee92ad6e3788f0b3d84a2a0d12455818b9557a1c)*

src/concurrency/blocking_io.rs
WhySoBad: This would break when we add a blocked thread for an already fulfilled interest. 
Either we iterate over everything at the end of the poll or we check whether the interest is already satisfied when adding the blocked thread. 
But you're right, it shouldn't be done there because `poll` gets called extensively and thus it's better when we can save this loop. I'll add the check in the `add_blocked_thread` method.

WhySoBad: Or maybe we should do the check in the `block_thread_for_id` method directly? There we already have access to an `ecx` and it's the only public way to add a thread anyways.

src/concurrency/blocking_io.rs
WhySoBad: We cannot use this FD for deregistering the source and instead need to use the source provided via the `source` parameter instead.

src/concurrency/blocking_io.rs
RalfJung: > This would break when we add a blocked thread for an already fulfilled interest.

Would we ever do that? We only block when we get an EAGAIN, don't we?
I noticed the comment that says that thread wakeup is level-triggered and not edge-triggered. But why is that what we want?

Unblocking the thread immediately should also work I think. That means that we won't even go through the scheduler at all when doing the 0-timeout trick if the underlying socket is already ready, but that should be fine?

src/concurrency/blocking_io.rs
RalfJung: I still don't know what else this function even could do. `sources` is keyed by `FdId` so of course you need that to do anything with the map...? You seem to be comparing with an alternative implementation that you have in your head but that the comment does not explain and I can't think of any possible alternative.

RalfJung: Oh wait is the comment just placed in the wrong spot? Is it about the `poll.registry().deregister` below rather than the `self.sources.remove` that it sits next to?

src/concurrency/blocking_io.rs
```diff
@@ -1,37 +1,132 @@
...
 
 /// Manager for managing blocking host I/O in a non-blocking manner.
 /// We use [`Poll`] to poll for new I/O events from the OS for sources
 /// registered using this manager.
 ///
+/// The semantics of this manager are that host I/O sources are registered
+/// to a [`Poll`] for their entire lifespan. Once host readiness events happen
+/// on a registered source, its internal epoll readiness gets updated -- even
+/// when the source isn't part of an active epoll instance. Also, for the entire
+/// lifespan of the source blocked, threads can be added which should be unblocked
```
RalfJung:
```suggestion
/// lifespan of the source, threads can be added which should be unblocked
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4973/0853747bf78b8b6e314dcdf14ade9c83a6c54f3e..b26bc701af7e93289160877418c8a00458bf62e9)*

src/concurrency/blocking_io.rs
WhySoBad: > We only block when we get an EAGAIN, don't we?

No, we don't. If the socket is blocking, we call the `block_for_*` method which blocks the thread and only executes the operation inside the callback function.  

> That means that we won't even go through the scheduler at all when doing the 0-timeout trick if the underlying socket is already ready, but that should be fine?

Yes that should be fine. We'll still go through the scheduler if the socket is not yet writable and then we'll get woken up because of an exceeded timeout.

src/concurrency/blocking_io.rs
WhySoBad: > Oh wait is the comment just placed in the wrong spot?

Yes, it belongs to the `poll.registry().deregister`.

src/concurrency/blocking_io.rs
RalfJung: > If the socket is blocking, we call the block_for_* method which blocks the thread and only executes the operation inside the callback function.

Hm, interesting, I wonder why we do that rather than "`read`, and on EAGAIN, block and try again (recursive call)". But that's not for this PR to change.

If calling the callback immediately works, that seems good for me.

src/concurrency/blocking_io.rs
```diff
@@ -205,23 +342,38 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
...
+    ) -> InterpResult<'tcx> {
         let this = self.eval_context_mut();
-        this.machine.blocking_io.register(
-            source_fd,
-            InterestReceiver::UnblockThread(this.active_thread()),
-            interests,
-        );
-        this.block_thread(BlockReason::IO, timeout, callback);
+
+        this.machine.blocking_io.add_blocked_thread(source_fd.id(), this.active_thread(), interest);
```
RalfJung, line 364:
```suggestion
        // We always have to do this since the thread will de-register itself.
        this.machine.blocking_io.add_blocked_thread(source_fd.id(), this.active_thread(), interest);
```

src/concurrency/blocking_io.rs
```diff
@@ -356,25 +345,35 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
...
     /// really ready and to block again if they're not.
     ///
     /// It's the callers responsibility to remove the [`BlockingIoInterest`]
-    /// from the blocking I/O manager in the callback function.
+    /// from the blocking I/O manager in the provided callback function.
     #[inline]
     fn block_thread_for_io(
         &mut self,
-        source_id: FdId,
+        source_fd: FileDescriptionRef<dyn SourceFileDescription>,
```
RalfJung, line 356: You could get away with an `&FileDescriptionRef` here and thus avoid the `clone`s on the caller side I think?

src/concurrency/blocking_io.rs
```diff
@@ -1,37 +1,132 @@
...
+    }
+
+    /// Check whether the current readiness fulfills the blocking I/O interest of
+    /// `interest`.
+    /// This function also returns `true` if the error readiness is set
+    /// even when the requested interest might not be fulfilled.
+    /// Usually, a source with the error readiness set won't become readable or writable
+    /// anymore. By waking up the thread, it will retry its operation and then notice and
+    /// process the error.
+    pub fn fulfills_interest(&self, interest: &BlockingIoInterest) -> bool {
```
RalfJung: Why is this function public? It's just called in this module, I think?

That also means it doesn't need such a detailed comment. The more relevant place for documentation is on public APIs, e.g. the `BlockingIoInterest` type itself.

RalfJung (commented):
Nice, that seems to work well. :)
Just fine-tuning some comments and method visibility.

You can also squash already if you want.

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4973/0853747bf78b8b6e314dcdf14ade9c83a6c54f3e..7f0161950feecdbae3f7a26ef21e71d31c1c895c)*

src/concurrency/blocking_io.rs
WhySoBad: We can't save the `clone` on any of the uses we currently have because we always need the FD inside the callback function. Should I still change it to take a reference as there might be uses in the future where the FD isn't needed in the callback function in the future?

RalfJung: Oh right the Rc is moved into the closure argument so it can't be borrowed by this argument... bummer.

Nah it's not worth it then.

RalfJung:
Amazing, let's finally land this then. :-)

PR #4977, opened 2026-04-23
Remove artificial short reads/writes from TCP sockets

WhySoBad:
For the same reason as in rust-lang/miri#4975, this pull request removes short reads/writes for the TCP socket shim.


RalfJung:
Thanks :)

PR #4978, opened 2026-04-24
bump openssl-sys to support OpenSSL 4.0.x

heitbaum:
The previously pinned version of openssl-sys is not compatible with OpenSSL 4.0.x.

- `openssl-sys`: 0.9.111 -> 0.9.114

Also see https://github.com/rust-lang/rust/issues/155397

r? @RalfJung


RalfJung:
> Closes https://github.com/rust-lang/rust/issues/155397

This PR definitely does not close that issue -- the lockfile is entirely ignored in the rustc repo.

But we can still bump the dependency. :shrug:

PR #4981, opened 2026-04-26
[ci] Update GitHub Actions to last major release

gruenich:
Fix deprecations warnings for Node20.js that will stop working in June.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
     env:
       HOST_TARGET: ${{ matrix.host_target }}
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - name: install multiarch
         if: ${{ matrix.multiarch != '' }}
         run: |
@@ -105,7 +105,7 @@ jobs:
     name: style checks
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - uses: ./.github/workflows/setup
 
       - name: rustfmt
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -205,7 +205,7 @@ jobs:
       - name: Install rustup-toolchain-install-master
         run: cargo install -f rustup-toolchain-install-master
       # Create a token for the next step so it can create a PR that actually runs CI.
-      - uses: actions/create-github-app-token@v2
+      - uses: actions/create-github-app-token@v3
```
RalfJung, line 208: I think this is the only one that's actually  necessary? checkout v5 doesn't seem to emit any deprecation warnings.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4981/0a272829b77a88fa85a5a3a8d9f0eeda793687c2..7e7d1f19a81a3895e9aebd01c7a216f52b34d87b)*

RalfJung:
Thanks for the PR!

This GHA version silliness is getting seriously tiring.

.github/workflows/ci.yml
gruenich: Sure. The other upadtes do not hurt, do they?

RalfJung: Uh so you just decided yourself that your answer is good enough and unilaterally marked the discussion as resolved?

gruenich: Apologies, that was too fast, indeed.

RalfJung:
I looked through all the changelogs and I think it's fine. If there are problems we'll just revert.

PR #4984, opened 2026-04-28
Add `getaddrinfo` and `freeaddrinfo` shims

WhySoBad:
Hi,

This pull request adds shims for the `getaddrinfo` and `freeaddrinfo` syscalls. They are used by the `ToSocketAddrs` trait in the standard library to resolve hostnames to socket addresses. 
Because the abstraction of the standard library is very tight, we're pretty much only able to shim the exact use case of the standard library. 

Also, we cannot shim `gai_strerror` because the standard library doesn't have `ErrorKind`s for the DNS errors. This is a bit unfortunate because Miri will throw an unsupported error when there is a resolution error (e.g. hostname doesn't exist).

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -155,6 +155,9 @@ pub enum NonHaltingDiagnostic {
     },
     FileInProcOpened,
     ConnectingSocketGetsockname,
+    SocketAddressResolution {
+        error: std::io::Error,
+    },
 }
 
 /// Level of Miri specific diagnostics
@@ -663,6 +666,8 @@ impl<'tcx> MiriMachine<'tcx> {
             FileInProcOpened => ("open a file in `/proc`".to_string(), DiagLevel::Warning),
             ConnectingSocketGetsockname =>
                 ("Called `getsockname` on connecting socket".to_string(), DiagLevel::Warning),
+            SocketAddressResolution { .. } =>
+                ("error during address resolution".to_string(), DiagLevel::Warning),
         };
 
         let title = match &e {
@@ -711,7 +716,8 @@ impl<'tcx> MiriMachine<'tcx> {
                 format!("GenMC currently does not model the failure ordering for `compare_exchange`. {was_upgraded_msg}. Miri with GenMC might miss bugs related to this memory access.")
             }
             FileInProcOpened => format!("files in `/proc` can bypass the Abstract Machine and might not work properly in Miri"),
-            ConnectingSocketGetsockname => format!("connecting sockets return unspecified socket addresses on Windows hosts")
+            ConnectingSocketGetsockname => format!("connecting sockets return unspecified socket addresses on Windows hosts"),
+            SocketAddressResolution { error } => format!("address resolution failed: {error}"),
         };
 
         let notes = match &e {
@@ -727,6 +733,10 @@ impl<'tcx> MiriMachine<'tcx> {
                         "an unspecified socket address (e.g. `0.0.0.0:0`) will be returned instead"
                     ),
                 ],
+            SocketAddressResolution { .. } =>
+                vec![note!(
+                    "Miri cannot return proper error information from this call; only a generic error code is being returned"
+                )],
             _ => vec![],
         };
 
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -202,8 +202,10 @@ pub enum MiriMemoryKind {
     /// Memory for thread-local statics.
     /// This memory may leak.
     Tls,
-    /// Memory mapped directly by the program
+    /// Memory mapped directly by the program.
     Mmap,
+    /// Memory allocated for `getaddrinfo` result.
+    SocketAddress,
 }
 
 impl From<MiriMemoryKind> for MemoryKind {
@@ -219,7 +221,7 @@ impl MayLeak for MiriMemoryKind {
         use self::MiriMemoryKind::*;
         match self {
             Rust | Miri | C | WinHeap | WinLocal | Runtime => false,
-            Machine | Global | ExternStatic | Tls | Mmap => true,
+            Machine | Global | ExternStatic | Tls | Mmap | SocketAddress => true,
         }
     }
 }
@@ -232,7 +234,7 @@ impl MiriMemoryKind {
             // Heap allocations are fine since the `Allocation` is created immediately.
             Rust | Miri | C | WinHeap | WinLocal | Mmap => true,
             // Everything else is unclear, let's not show potentially confusing spans.
-            Machine | Global | ExternStatic | Tls | Runtime => false,
+            Machine | Global | ExternStatic | Tls | Runtime | SocketAddress => false,
         }
     }
 }
@@ -252,6 +254,7 @@ impl fmt::Display for MiriMemoryKind {
             ExternStatic => write!(f, "extern static"),
             Tls => write!(f, "thread-local static"),
             Mmap => write!(f, "mmap"),
+            SocketAddress => write!(f, "socket address"),
         }
     }
 }
--- a/src/shims/unix/socket.rs
+++ b/src/shims/unix/socket.rs
@@ -1,13 +1,12 @@
 use std::cell::{Cell, RefCell, RefMut};
+use std::io;
 use std::io::Read;
-use std::net::{Ipv4Addr, Ipv6Addr, Shutdown, SocketAddr, SocketAddrV4, SocketAddrV6};
+use std::net::{Ipv4Addr, Shutdown, SocketAddr, SocketAddrV4};
 use std::sync::atomic::AtomicBool;
 use std::time::Duration;
-use std::{io, iter};
 
 use mio::event::Source;
 use mio::net::{TcpListener, TcpStream};
-use rustc_abi::Size;
 use rustc_const_eval::interpret::{InterpResult, interp_ok};
 use rustc_middle::throw_unsup_format;
 use rustc_target::spec::Os;
@@ -15,6 +14,7 @@ use rustc_target::spec::Os;
 use crate::shims::files::{EvalContextExt as _, FdId, FileDescription, FileDescriptionRef};
 use crate::shims::unix::UnixFileDescription;
 use crate::shims::unix::linux_like::epoll::{EpollReadiness, EvalContextExt as _};
+use crate::shims::unix::socket_address::EvalContextExt as _;
 use crate::*;
 
 #[derive(Debug, PartialEq)]
@@ -354,7 +354,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let this = self.eval_context_mut();
 
         let socket = this.read_scalar(socket)?.to_i32()?;
-        let address = match this.socket_address(address, address_len, "bind")? {
+        let address = match this.read_socket_address(address, address_len, "bind")? {
             Ok(addr) => addr,
             Err(e) => return this.set_last_error_and_return_i32(e),
         };
@@ -570,7 +570,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let this = self.eval_context_mut();
 
         let socket = this.read_scalar(socket)?.to_i32()?;
-        let address = match this.socket_address(address, address_len, "connect")? {
+        let address = match this.read_socket_address(address, address_len, "connect")? {
             Ok(address) => address,
             Err(e) => return this.set_last_error_and_return(e, dest),
         };
@@ -1018,10 +1018,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             SocketState::Initial => SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::UNSPECIFIED, 0)),
         };
 
-        match this.write_socket_address(&address, address_ptr, address_len_ptr, "getsockname")? {
-            Ok(_) => interp_ok(Scalar::from_i32(0)),
-            Err(e) => this.set_last_error_and_return_i32(e),
-        }
+        this.write_socket_address(&address, address_ptr, address_len_ptr, "getsockname")
+            .map(|_| Scalar::from_i32(0))
     }
 
     fn getpeername(
@@ -1078,15 +1076,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                         Err(e) => return this.set_last_error_and_return(e, &dest),
                     };
 
-                    match this.write_socket_address(
+                    this.write_socket_address(
                         &address,
                         address_ptr,
                         address_len_ptr,
                         "getpeername",
-                    )? {
-                        Ok(_) => this.write_scalar(Scalar::from_i32(0), &dest),
-                        Err(e) => this.set_last_error_and_return(e, &dest),
-                    }
+                    )?;
+                   this.write_scalar(Scalar::from_i32(0), &dest)
                 }
             ),
         )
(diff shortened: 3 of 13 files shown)


RalfJung:
> Also, we cannot shim gai_strerror because the standard library doesn't have ErrorKinds for the DNS errors. This is a bit unfortunate because Miri will throw an unsupported error when there is a resolution error (e.g. hostname doesn't exist).

Can't we at least return something generic, like what our `strerror_r` shim does when encountering an unknown error kind?

WhySoBad:
Yeah we can do that.

WhySoBad:
Unlike `strerror_r`, `gai_strerror` doesn't take a buffer argument so we need to do static allocations for the error messages. Since we have the `AddressInfoStore` anyways, I was thinking about adding the error allocations there. 

And we also have the problem that `io_error_to_errnum` throws an unsupported error here: https://github.com/rust-lang/miri/blob/3f8007e19b8cac3d08fb07579bde3217c58807e4/src/shims/io_error.rs#L249-L260

Are there any other shims which needed to circumvent this method?

RalfJung:
> And we also have the problem that io_error_to_errnum throws an unsupported error here: 

What are the errnums you are getting?

WhySoBad:
Well, we only get a an `Uncategorized` `io::ErrorKind` because the `io::Error` is constructed manually here: https://github.com/rust-lang/rust/blob/HEAD/library/std/src/sys/net/connection/socket/unix.rs#L53 

The problem is that the `UNIX_IO_ERROR_TABLE` doesn't contain an entry for `Uncategorized` (which it isn't exposed to because we shouldn't match against that).

RalfJung:
Ah so the actual problem is that
- getaddrinfo returns an error code (but not a standard errnum code, these are specific DNS error codes)
- std translates that into a string
- we now have to somehow get a code back out of that again

There's no standard Unix errnum involved here ever, std just shoves this into an `io::Error` because it has to.

src/shims/unix/socket.rs
```diff
@@ -1053,6 +1093,126 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let flags_field = this.project_field_named(&hints_mplace, "ai_flags")?;
+        let flags = this.read_scalar(&flags_field)?;
+        if flags != Scalar::from_i32(0) {
+            // We cannot provide any flag hints to the standard library implementation.
+            throw_unsup_format!("getaddrinfo: flag hints are not supported")
+        }
+
+        let socket_addrs = match (node_str, port).to_socket_addrs() {
+            Ok(addrs) => addrs,
+            Err(e) => return this.set_last_error_and_return_i32(e),
```
RalfJung: So, this `set_last_error_and_return_i32` call here doesn't really make sense. It means we return `-1` on error but that's not how `getaddrinfo` works.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4984/98cb1d3d48fa7c8e97994050bfd69e8f3da2a1a7..3f8007e19b8cac3d08fb07579bde3217c58807e4)*

WhySoBad:
No, we get standard UNIX error codes back -- they are just DNS specific:
```
RETURN VALUE
     getaddrinfo() returns 0 if it succeeds, or one of the following nonzero error codes:

     EAI_ADDRFAMILY
            The specified network host does not have any network addresses in the requested address family.

     EAI_AGAIN
            The name server returned a temporary failure indication.  Try again later.

     EAI_BADFLAGS
            hints.ai_flags contains invalid flags; or, hints.ai_flags included AI_CANONNAME and node was NULL.

     EAI_FAIL
            The name server returned a permanent failure indication.

     EAI_FAMILY
            The requested address family is not supported.

     EAI_MEMORY
            Out of memory.

     EAI_NODATA
            The specified network host exists, but does not have any network addresses defined.

     EAI_NONAME
            The node or service is not known; or both node and service are NULL; or AI_NUMERICSERV was specified in hints.ai_flags and
            service was not a numeric port-number string.

     EAI_SERVICE
            The requested service is not available for the requested socket type.  It may be available through  another  socket  type.
            For  example,  this  error  could  occur  if  service was "shell" (a service available only on stream sockets), and either
            hints.ai_protocol was IPPROTO_UDP, or hints.ai_socktype was SOCK_DGRAM; or the error could occur if service was not  NULL,
            and hints.ai_socktype was SOCK_RAW (a socket type that does not support the concept of services).

     EAI_SOCKTYPE
            The requested socket type is not supported.  This could occur, for example, if hints.ai_socktype and hints.ai_protocol are
            inconsistent (e.g., SOCK_DGRAM and IPPROTO_TCP, respectively).

     EAI_SYSTEM
            Other system error; errno is set to indicate the error.

     The gai_strerror() function translates these error codes to a human readable string, suitable for error reporting.
```
And the standard library has by far not all UNIX error codes mapped to `io::ErrorKind`s (there are some experimental features to add more variants, but even with those features the DNS errorcodes aren't covered).
For unmapped `io::ErrorKind`s there is just the `io::ErrorKind::Uncategorized` which shouldn't be matched against because error kinds which map to might receive their own `io::ErrorKind` variant in the future.

src/shims/unix/socket.rs
WhySoBad: Oh, I see, it shouldn't set `errno` but instead just return the negative integer instead. I've misread the man page. But the problem with unmapped error codes remains for #4987.

RalfJung:
> No, we get standard UNIX error codes back

These are not standard unix error codes. `EAI_*` are negative numbers directly returned from that function. Normal unix error codes are positive and they are stored in `errnum` and the function returns `-1`.

This is basically a separate error enum specifically for `getaddrinfo`.

RalfJung:
> For unmapped io::ErrorKinds there is just the io::ErrorKind::Uncategorized which shouldn't be matched against because error kinds which map to might receive their own io::ErrorKind variant in the future.

No, `io::Error` can represent all host errors, even "unmapped" ones. That's what `Error::from_raw_os_error` is for.

But that's only for "normal" host error codes, not for this other kind of error code that `getaddrinfo` returns.

src/shims/unix/socket.rs
RalfJung: Yeah, and also here we have the problem of even figuring out what to return. std hides the actual error code from us. We can just guess something but it's not great.

WhySoBad: Yes that's a problem. So should I just return something like `EAI_FAIL` because it's kind of a permanent failure?

RalfJung: I was thinking EAI_SYSTEM and then set errno to the closest approximation of "we have no idea". Maybe EPROTO?

WhySoBad: Sounds reasonable. Maybe we should also emit a warning every time we return a bogus error such that it's clear that this wasn't the real error?

RalfJung: Usually we only emit warnings the first time this happens, but that seems reasonable yeah.

src/shims/unix/socket.rs
```diff
@@ -1114,6 +1133,134 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    // numeric ports.
+                    throw_unsup_format!(
+                        "getaddrinfo: non-numeric `service` arguments aren't supported"
+                    )
+                }
+            }
+        }
+
+        let node_c_str = this.read_c_str(node_ptr)?;
+        let node_str = String::from_utf8_lossy(node_c_str).to_string();
```
RalfJung: I think I'd prefer to throw an "unsupported" error if the string is not valid unicode, rather than silently doing the wrong thing.

src/shims/unix/socket.rs
```diff
@@ -1114,6 +1133,134 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            );
+        }
+
+        let mut port = 0;
+        if service_ptr != Pointer::null() {
+            // The C-string at `service_ptr` is either a port number or a name of a
+            // well-known service.
+            let service_c_str = this.read_c_str(service_ptr)?;
+            let service_string = String::from_utf8_lossy(service_c_str);
+            match service_string.parse::<u16>() {
```
```suggestion
            // Try to parse `service_c_str` as a number -- the only case we support.
            let port = str::from_utf8(service_c_str).ok().and_then(|s| s.parse::<u16>().ok());
            match port {
```

src/shims/unix/socket.rs
```diff
@@ -1114,6 +1133,134 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let this = self.eval_context_mut();
+
+        let res_ptr = this.read_pointer(res)?;
+
+        if !this.machine.address_store.0.remove(&res_ptr) {
+            // The pointer was not stored in the set.
+            throw_ub_format!(
+                "freeaddrinfo: attempted to free linked list which wasn't allocated using getaddrinfo"
+            );
+        }
```
RalfJung: The usual way we'd ensure this is by adding a new variant to `MiriMemoryKind`. Does that not work here?

src/shims/unix/socket.rs
```diff
@@ -1114,6 +1133,134 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             Err(e) => this.set_last_error_and_return_i32(e),
         }
     }
+
+    fn getaddrinfo(
```
RalfJung: This is a lot of code that seems to have nothing to do with the socket FD type. Please move all the getaddrinfo-related shims to a new file.

src/shims/unix/socket.rs
```diff
@@ -1386,6 +1533,106 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        let Some(address) = addresses.next() else {
+            // Iterator is empty; we return a null pointer.
+            return interp_ok(Ok(Pointer::null()));
+        };
+
+        let addrinfo_layout = this.libc_ty_layout("addrinfo");
+
+        let addrinfo_mplace = this.allocate(addrinfo_layout, MiriMemoryKind::Machine.into())?;
+        // Zero the newly allocated address info struct.
```
RalfJung: Do we have to do this? We usually only do that if we can't fill / know all the fields or are too lazy to figure out all the fields.

src/shims/unix/socket.rs
```diff
@@ -1386,6 +1533,106 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            SocketAddr::V4(_) => this.eval_libc("AF_INET"),
+            SocketAddr::V6(_) => this.eval_libc("AF_INET6"),
+        };
+        this.write_scalar(family, &family_mplace)?;
+
+        let socktype_mplace = this.project_field_named(&addrinfo_mplace, "ai_socktype")?;
+        this.write_scalar(this.eval_libc("SOCK_STREAM"), &socktype_mplace)?;
+
+        let protocol_mplace = this.project_field_named(&addrinfo_mplace, "ai_protocol")?;
+        this.write_int(0, &protocol_mplace)?;
```
RalfJung: What does "0" mean here?

src/shims/unix/socket.rs
```diff
@@ -1386,6 +1533,106 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            "getaddrinfo",
+        )? {
+            return interp_ok(Err(err));
+        };
+        this.write_pointer(sockaddr_mplace.ptr(), &addr_mplace)?;
+
+        let canonname_mplace = this.project_field_named(&addrinfo_mplace, "ai_canonname")?;
+        this.write_pointer(Pointer::null(), &canonname_mplace)?;
+
+        let next_mplace = this.project_field_named(&addrinfo_mplace, "ai_next")?;
```
```suggestion
        // Allocate remaining list and store a pointer to it.
        let next_mplace = this.project_field_named(&addrinfo_mplace, "ai_next")?;
```

src/shims/unix/socket.rs
```diff
@@ -1386,6 +1533,106 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let next_ptr = match this.allocate_address_infos(addresses)? {
+            Ok(next_ptr) => next_ptr,
+            Err(e) => return interp_ok(Err(e)),
```
RalfJung: All these `return Err` will leak memory. Is that expected?

Can this even ever error? I found no error path in `write_socket_address`.

src/diagnostics.rs
```diff
@@ -711,7 +716,8 @@ impl<'tcx> MiriMachine<'tcx> {
                 format!("GenMC currently does not model the failure ordering for `compare_exchange`. {was_upgraded_msg}. Miri with GenMC might miss bugs related to this memory access.")
             }
             FileInProcOpened => format!("files in `/proc` can bypass the Abstract Machine and might not work properly in Miri"),
-            ConnectingSocketGetsockname => format!("connecting sockets return unspecified socket addresses on Windows hosts")
+            ConnectingSocketGetsockname => format!("connecting sockets return unspecified socket addresses on Windows hosts"),
+            SocketAddressResolution { .. } => format!("address resolution errors aren't handled properly in Miri")
```
```suggestion
            SocketAddressResolution { .. } => format!("address resolution errors aren't handled properly in Miri"),
```

src/diagnostics.rs
```diff
@@ -727,6 +733,13 @@ impl<'tcx> MiriMachine<'tcx> {
                         "an unspecified socket address (e.g. `0.0.0.0:0`) will be returned instead"
                     ),
                 ],
+            SocketAddressResolution { error } =>
+                vec![
+                    note!(
+                        "instead of the actual error, Miri just returns a generic protocol error"
+                    ),
+                    note!("the actual error is '{error}'"),
```
RalfJung: Seems odd to only put the actual error into the note. Why not put it in the main message?

tests/pass-dep/libc/libc-socket-invalid-addr.rs
```diff
@@ -0,0 +1,31 @@
+//@ignore-target: windows # No libc socket on Windows
+//@compile-flags: -Zmiri-disable-isolation
+//@normalize-stderr-test: "the actual error is '.*'" -> "the actual error is 'host I/O error'"
```
RalfJung:
```suggestion
//@normalize-stderr-test: "the actual error is '.*'" -> "the actual error is '$MSG'"
```

tests/utils/libc.rs
```diff
@@ -369,7 +371,7 @@ pub mod net {
     /// syscall and the written address of it.
     pub fn sockname_ipv4<F>(f: F) -> io::Result<(libc::c_int, libc::sockaddr_in)>
     where
-        F: FnOnce(*mut libc::sockaddr, *mut libc::socklen_t) -> libc::c_int,
+        F: FnOnce(*mut libc::sockaddr_storage, *mut libc::socklen_t) -> libc::c_int,
```
RalfJung: Why did you change this type? That seems to just lead to lots of annoying casts.

RalfJung (commented):
Here's the first proper round of comments. :)
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4984/e0b3afbdd5362aa348abc6661ed6417d1a21f58c..d811aaf958594156e2ce38365b835c361aa88de9)*

src/shims/unix/socket.rs
WhySoBad: No this wasn't expected.
And you're right, there doesn't seem to be an error path in `write_socket_address` anymore.

src/shims/unix/socket.rs
WhySoBad: I didn't use that because I didn't know about it. This works fine.

src/shims/unix/socket.rs
WhySoBad: I've also moved the `read_socket_address` and `write_socket_address` methods into a new `socket_address` shim.

tests/utils/libc.rs
WhySoBad: Because `sockaddr` doesn't actually fit all socket address types. Especially, IPv6 addresses don't fit into `sockaddr` structs.

src/shims/unix/socket.rs
WhySoBad: This just means we want to fall back to the default protocol for the provided socket type, I've now added a comment there.

tests/utils/libc.rs
RalfJung: Given that `getsockname` uses `sockaddr`, IMO it'd make most sense to also use it here. I can't see how forcing basically every caller to cast helps anyone.

tests/pass-dep/libc/libc-socket-invalid-addr.rs
```diff
@@ -1,6 +1,6 @@
 //@ignore-target: windows # No libc socket on Windows
 //@compile-flags: -Zmiri-disable-isolation
-//@normalize-stderr-test: "the actual error is '.*'" -> "the actual error is 'host I/O error'"
+//@normalize-stderr-test: "address resolution failed: .*" -> "address resolution failed: $MSG"
```
RalfJung: Ah I screwed up, this should be `$$MSG`

src/diagnostics.rs
```diff
@@ -727,6 +733,13 @@ impl<'tcx> MiriMachine<'tcx> {
                         "an unspecified socket address (e.g. `0.0.0.0:0`) will be returned instead"
                     ),
                 ],
+            SocketAddressResolution { .. } =>
+                vec![
+                    note!("address resolution errors aren't handled properly in Miri"),
+                    note!(
+                        "instead of the actual error, Miri just returns a generic protocol error"
+                    ),
```
```suggestion
                vec![
                    note!("Miri cannot return proper error information from this call; only a generic error code is being returned"),
```

src/diagnostics.rs
```diff
@@ -663,6 +666,8 @@ impl<'tcx> MiriMachine<'tcx> {
             FileInProcOpened => ("open a file in `/proc`".to_string(), DiagLevel::Warning),
             ConnectingSocketGetsockname =>
                 ("Called `getsockname` on connecting socket".to_string(), DiagLevel::Warning),
+            SocketAddressResolution { .. } =>
+                ("error during socket address resolution".to_string(), DiagLevel::Warning),
```
```suggestion
                ("error during address resolution".to_string(), DiagLevel::Warning),
```

RalfJung (commented):
Just a few minor things (and that discussion about the type for the `sockname*` helpers)

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4984/e0b3afbdd5362aa348abc6661ed6417d1a21f58c..26a6bc618dceffa546e2cced619dc8402fc16818)*

tests/pass-dep/libc/libc-socket-invalid-addr.rs
WhySoBad: I think it's: 
```
//@normalize-stderr-test: "address resolution failed: (<MSG>.*)" -> "address resolution failed: $$MSG"
```

But is this actually what we want? On Windows we get different error messages and with this we again have the message inside the .stderr file.

RalfJung: `$$MSG` should render as "$MSG" in the output, that was the goal.

Your original pattern confused me because it looked like "host I/O error" was the actual error message.

src/machine.rs
```diff
@@ -202,8 +202,10 @@ pub enum MiriMemoryKind {
     /// Memory for thread-local statics.
     /// This memory may leak.
     Tls,
-    /// Memory mapped directly by the program
+    /// Memory mapped directly by the program.
     Mmap,
+    /// Memory containing linked lists of address info structs.
```
RalfJung:
```suggestion
    /// Memory allocated for `getaddrinfo` result.
```

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -50,6 +51,8 @@ fn main() {
     test_shutdown();
     test_shutdown_readable_after_write_close();
     test_shutdown_writable_after_read_close();
+
+    test_getaddrinfo_freeaddrinfo();
```
RalfJung: Please move this to a separate file... it doesn't have that much to do with sockets themselves after all.

RalfJung (commented):
Thanks, looks great! Two last minor comments, and then please squash the commits.
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4984/e0b3afbdd5362aa348abc6661ed6417d1a21f58c..bb10a02a38d4803d83c58f55b0aa150375817608)*

tests/pass-dep/libc/libc-socket.rs
WhySoBad: I've also moved the address resolution test from `tests/shims/socket.rs` into a separate file now.

tests/pass/shims/socket-address.rs
```diff
@@ -0,0 +1,33 @@
+//@ignore-target: windows # No libc socket on Windows
```
RalfJung: This isn't a libc test so the comment doesn't make much sense.
```suggestion
//@ignore-target: windows # No socket address support on Windows
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4984/2c6e263a237469674d4b1d525435312b36a62e1f..a5e33c078dd1f16ada5e273a329a0000fff3b807)*

PR #4985, opened 2026-04-29
Add `getsockname` shim for connecting and connected sockets

WhySoBad:
Hi,

This pull request adds the `getsockname` shim for connecting and connected sockets. At the moment, we just return `0.0.0.0` in those cases but because connecting sockets are already locally bound we can just return the value from `TcpStream::local_addr`.

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -144,6 +144,7 @@ pub enum NonHaltingDiagnostic {
         effective_failure_ordering: AtomicReadOrd,
     },
     FileInProcOpened,
+    ConnectingSocketGetsockname,
 }
 
 /// Level of Miri specific diagnostics
@@ -650,6 +651,8 @@ impl<'tcx> MiriMachine<'tcx> {
             | WeakMemoryOutdatedLoad { .. } =>
                 ("tracking was triggered here".to_string(), DiagLevel::Note),
             FileInProcOpened => ("open a file in `/proc`".to_string(), DiagLevel::Warning),
+            ConnectingSocketGetsockname =>
+                ("Called `getsockname` on connecting socket".to_string(), DiagLevel::Warning),
         };
 
         let title = match &e {
@@ -698,12 +701,22 @@ impl<'tcx> MiriMachine<'tcx> {
                 format!("GenMC currently does not model the failure ordering for `compare_exchange`. {was_upgraded_msg}. Miri with GenMC might miss bugs related to this memory access.")
             }
             FileInProcOpened => format!("files in `/proc` can bypass the Abstract Machine and might not work properly in Miri"),
+            ConnectingSocketGetsockname => format!("connecting sockets return unspecified socket addresses on Windows hosts")
         };
 
         let notes = match &e {
             ProgressReport { block_count } => {
                 vec![note!("so far, {block_count} basic blocks have been executed")]
             }
+            ConnectingSocketGetsockname =>
+                vec![
+                    note!(
+                        "Windows hosts do not provide `local_addr` information while the socket is still connecting, which might break the assumptions of code compiled for Unix targets"
+                    ),
+                    note!(
+                        "an unspecified socket address (e.g. `0.0.0.0:0`) will be returned instead"
+                    ),
+                ],
             _ => vec![],
         };
 
--- a/src/shims/unix/socket.rs
+++ b/src/shims/unix/socket.rs
@@ -1,6 +1,7 @@
 use std::cell::{Cell, RefCell};
 use std::io::Read;
 use std::net::{Ipv4Addr, Ipv6Addr, Shutdown, SocketAddr, SocketAddrV4, SocketAddrV6};
+use std::sync::atomic::AtomicBool;
 use std::time::Duration;
 use std::{io, iter};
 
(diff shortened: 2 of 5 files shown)


src/shims/unix/socket.rs
```diff
@@ -975,6 +975,11 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                     Err(e) => return this.set_last_error_and_return_i32(e),
                 },
+            SocketState::Connecting(stream) | SocketState::Connected(stream) =>
+                match stream.local_addr() {
+                    Ok(address) => address,
+                    Err(e) => return this.set_last_error_and_return_i32(e),
+                },
             // For non-bound sockets the POSIX manual says the returned address is unspecified.
             // Often this is 0.0.0.0:0 and thus we set it to this value.
             _ => SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::UNSPECIFIED, 0)),
```
RalfJung: Please spell out which states are handled by this case (i.e., replace the `_` by a concrete pattern).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4985/98cb1d3d48fa7c8e97994050bfd69e8f3da2a1a7..7b74d77d7c74a8ba5903a2bb0fc8c9ead598c6ae)*

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -407,6 +408,34 @@ fn test_getsockname_ipv4_unbound() {
     assert_eq!(addr.sin_addr.s_addr, sock_addr.sin_addr.s_addr);
 }
 
+/// Test the `getsockname` syscall on a connect IPv4 socket.
```
```suggestion
/// Test the `getsockname` syscall on a connected IPv4 socket.
```
I assume?

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -407,6 +408,34 @@ fn test_getsockname_ipv4_unbound() {
...
+    net::connect_ipv4(client_sockfd, addr).unwrap();
+
+    let (_, sock_addr) = net::sockname_ipv4(|storage, len| unsafe {
+        libc::getsockname(client_sockfd, storage, len)
+    })
+    .unwrap();
+
+    // We want to ensure that the local address is not the unspecified address.
+    // Because the bound address could be of any local interface, we cannot
+    // test for localhost.
```
RalfJung, line 449: The socket API would let us `bind` before calling `connect` to set the local address, right?
But I guess that's not implemented yet / can't even be implemented with the mio API?

tests/pass-dep/libc/libc-socket-no-blocking.rs
```diff
@@ -574,6 +576,52 @@ fn test_write_read_nonblock() {
...
+        libc::getsockname(client_sockfd, storage, len)
+    })
+    .unwrap();
+
+    // The unspecified IPv4 address.
+    let addr = net::sock_addr_ipv4([0, 0, 0, 0], 0);
+
+    assert_eq!(addr.sin_family, sock_addr.sin_family);
+    if cfg!(windows_host) {
+        // On Windows hosts a connecting socket is bound to the unspecified address.
```
RalfJung, line 615: That seems potentially problematic, doesn't it? Unix code may rely on knowing what the bound address is immediately after calling `connect`?

Is this entirely determined by `local_addr` on the host or is there some of our own logic involved here?

RalfJung (commented):
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4985/98cb1d3d48fa7c8e97994050bfd69e8f3da2a1a7..7b74d77d7c74a8ba5903a2bb0fc8c9ead598c6ae)*

tests/pass-dep/libc/libc-socket.rs
WhySoBad: Yes, the socket API allows to do this but we cannot represent this with mio because we only have `TcpStream::connect` there.

tests/pass-dep/libc/libc-socket-no-blocking.rs
WhySoBad: Ouh, you're right. There is no of our logic involved here and it's documented behavior on Windows (https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-getsockname):

>A Windows Sockets application must not assume that the address will be specified unless the socket is connected.

I guess the only way to prevent this would be to ensure that the socket is actually connected (and block for it otherwise) when calling `local_addr` on a Windows host.

RalfJung: We can't really have that call block...

We should definitely add a FIXME in the code. And maybe emit a warning the first time this happens?

src/shims/unix/socket.rs
```diff
@@ -975,14 +976,30 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    && matches!(&*state, SocketState::Connecting(_))
+                    && this.tcx.sess.target.os != Os::Windows
+                {
+                    // FIXME: On Windows hosts `TcpStream::local_addr` returns `0.0.0.0:0` whilst
+                    // the socket is connecting:
+                    // <https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-getsockname#remarks>
+                    // This is problematic because UNIX targets could expect a real local address even
+                    // for a connecting non-blocking socket.
+
+                    static DEDUP: SpanDedupDiagnostic = SpanDedupDiagnostic::new();
```
RalfJung: I don't think we need per-span deduplication here. Just use a simple `AtomicBool` and show the warning only once globally.

src/shims/unix/socket.rs
```diff
@@ -975,14 +976,30 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     Ok(address) => address,
                     Err(e) => return this.set_last_error_and_return_i32(e),
                 },
-            SocketState::Connecting(stream) | SocketState::Connected(stream) =>
+            SocketState::Connecting(stream) | SocketState::Connected(stream) => {
+                if cfg!(windows)
+                    && matches!(&*state, SocketState::Connecting(_))
+                    && this.tcx.sess.target.os != Os::Windows
```
RalfJung: This code lives in the "unix" folder. I sure hope the target OS can never be Windows :)

src/diagnostics.rs
```diff
@@ -698,12 +701,22 @@ impl<'tcx> MiriMachine<'tcx> {
...
         };
 
         let notes = match &e {
             ProgressReport { block_count } => {
                 vec![note!("so far, {block_count} basic blocks have been executed")]
             }
+            ConnectingSocketGetsockname =>
+                vec![
+                    note!(
+                        "UNIX targets expect a local socket address from `getsockname` even for connecting sockets"
```
```suggestion
                        "Windows hosts do not provide `local_addr` information while the socket is still connecting, which might break the assumptions of code compiled for Unix targets"
```

src/diagnostics.rs
```diff
@@ -698,12 +701,22 @@ impl<'tcx> MiriMachine<'tcx> {
...
+                    ),
+                    note!(
+                        "an unspecified socket address (e.g. 0.0.0.0:0) will be returned instead"
```
```suggestion
                        "an unspecified socket address (e.g. `0.0.0.0:0`) will be returned instead"
```

RalfJung (commented):
Next round :)
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4985/98cb1d3d48fa7c8e97994050bfd69e8f3da2a1a7..7a301d743268f6b9489b072ec3c24d01fb44b27e)*

RalfJung:

@rustbot author

PR #4987, opened 2026-04-29
Add network socket `shutdown` shim

WhySoBad:
Hi,

This pull request adds the shim for the `shutdown` syscall for TCP sockets. 
I've ran the tokio TCP tests using Miri and noticed that many of them require `TcpStream::shutdown` so it would be nice to also have this shimmed.

### DIFF
--- a/src/shims/io_error.rs
+++ b/src/shims/io_error.rs
@@ -251,6 +251,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     fn io_error_to_errnum(&self, err: std::io::Error) -> InterpResult<'tcx, Scalar> {
         let this = self.eval_context_ref();
         let target = &this.tcx.sess.target;
+
         if target.families.iter().any(|f| f == "unix") {
             for &(name, kind) in UNIX_IO_ERROR_TABLE {
                 if err.kind() == kind {
--- a/tests/pass-dep/libc/libc-socket.rs
+++ b/tests/pass-dep/libc/libc-socket.rs
@@ -44,6 +44,10 @@ fn main() {
 
     test_getpeername_ipv4();
     test_getpeername_ipv6();
+
+    test_shutdown();
+    test_shutdown_readable_after_write_close();
+    test_shutdown_writable_after_read_close();
 }
 
 /// Test creating a socket and then closing it afterwards.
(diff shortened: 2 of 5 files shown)


WhySoBad:
The CI fails for the same reason mentioned here: https://github.com/rust-lang/miri/pull/4984#issuecomment-4338105485.
It's because `WSAESHUTDOWN` is not a known error kind in the standard library.

What would you suggest doing to handle such unknown I/O errors?

src/shims/io_error.rs
```diff
@@ -255,7 +255,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // FIXME: This is a temporary workaround for handling WSAESHUTDOWN errors
         // on Windows. A discussion on how those errors should be handled can be found here:
         // <https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/WSAESHUTDOWN.20error.20on.20Windows/near/591883531>
-        let err = if matches!(err.raw_os_error(), Some(/* libc::WSAESHUTDOWN */ 10058)) {
+        let err = if cfg!(windows)
+            && matches!(err.raw_os_error(), Some(/* libc::WSAESHUTDOWN */ 10058))
```
RalfJung: Again, please just add it to the table. These special hacks seem both incorrect and unnecessary.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4987/f1a393bd9726374fdb7b4880f8427a37d6b76997..7bd742fde6b79f5ab883e36e7dc7957896f3d92d)*

RalfJung: Ah no wait you're right, this is a host problem.

But then please move the hack into `socket.rs`. This is specifically about `write`/`send` returning WSAESHUTDOWN, I don't think we want this applied globally.

Also, `libc::WSAESHUTDOWN` is not a thing.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -504,3 +490,50 @@ fn test_getpeername_ipv6() {
...
+
+    let mut byte = [0u8];
+
+    net::connect_ipv4(client_sockfd, addr).unwrap();
+    let client_dup_sockfd = unsafe { libc::dup(client_sockfd) };
+
+    // Closing should prevent reads/writes.
+    unsafe {
+        libc::shutdown(client_sockfd, libc::SHUT_WR);
+        errno_result(libc::write(client_sockfd, [0].as_ptr().cast(), 1)).unwrap_err();
```
RalfJung: Please check that you get the right error code here.

Also... reads should still be possible here, right?

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -504,3 +490,50 @@ fn test_getpeername_ipv6() {
...
+        libc::shutdown(client_sockfd, libc::SHUT_RD);
+        let bytes_read =
+            errno_result(libc::read(client_sockfd, byte.as_mut_ptr().cast(), 1)).unwrap();
+        assert_eq!(bytes_read, 0);
+    }
+
+    // TODO: Once epoll is available for TCP sockets, ensure that the rdhup and hup readiness
+    // are set.
+
+    // Closing should affect previous handles.
```
```suggestion
    // Closing should affect previously duplicated handles.
```

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -504,3 +490,50 @@ fn test_getpeername_ipv6() {
...
+    unsafe {
+        errno_result(libc::write(client_dup_sockfd, [0].as_ptr().cast(), 1)).unwrap_err();
+        let bytes_read =
+            errno_result(libc::read(client_dup_sockfd, byte.as_mut_ptr().cast(), 1)).unwrap();
+        assert_eq!(bytes_read, 0);
+    }
+
+    // Closing should affect new handles.
```
```suggestion
    // Closing should affect newly duplicated handles.
```

tests/pass/shims/socket.rs
```diff
@@ -113,3 +115,41 @@ fn test_peer_addr() {
...
+    let mut stream = TcpStream::connect(address).unwrap();
+    let mut stream_clone = stream.try_clone().unwrap();
+
+    // Closing should prevent reads/writes.
+    stream.shutdown(Shutdown::Write).unwrap();
+    stream.write(&[0]).unwrap_err();
+    stream.shutdown(Shutdown::Read).unwrap();
+    assert_eq!(stream.read(&mut byte).unwrap(), 0);
+
+    // Closing should affect previous handles.
```
RalfJung:
```suggestion
    // Closing should affect previously cloned handles.
```

tests/pass/shims/socket.rs
```diff
@@ -113,3 +115,41 @@ fn test_peer_addr() {
...
+    stream_clone.write(&[0]).unwrap_err();
+    assert_eq!(stream_clone.read(&mut byte).unwrap(), 0);
+
+    // Closing should affect new handles.
```
RalfJung:
```suggestion
    // Closing should affect newly cloned handles.
```

tests/pass/shims/socket.rs
```diff
@@ -113,3 +115,41 @@ fn test_peer_addr() {
...
+    // we need to connect to.
+    let address = listener.local_addr().unwrap();
+
+    // Start server thread.
+    let handle = thread::spawn(move || {
+        let (_stream, _addr) = listener.accept().unwrap();
+
+        // Yield back to client thread to ensure that the peer
+        // is only dropped at the end of the test.
+        thread::sleep(Duration::from_millis(10));
```
RalfJung: What happens if we are too slow? Seems problematic if we then get spurious errors.

RalfJung (commented):
I didn't even know this operation exists.^^ I wonder what the point is, why wouldn't you just `close`?

Anyway, this mostly looks good, except for some test nits.
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4987/f1a393bd9726374fdb7b4880f8427a37d6b76997..df98139fc40404fce165e2d008db8767d5686db1)*

tests/pass-dep/libc/libc-socket.rs
WhySoBad: Yes, I've extended the test to assert that.

tests/pass/shims/socket.rs
WhySoBad: I've now changed it such that the thread returns the stream, this way it only gets dropped once we join the thread at the end of the test.

tests/pass-dep/libc/libc-socket.rs
RalfJung: Can you also add a test checking that if we only SHUT_RD, then writing still works?

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -508,7 +510,9 @@ fn test_shutdown() {
     // Closing should prevent reads/writes.
     unsafe {
         libc::shutdown(client_sockfd, libc::SHUT_WR);
```
RalfJung: Since you have separate tests for WR/RD closing, let's make this test close both at once.

RalfJung (commented):
Just one final nit. :) Please also squash the commits.

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4987/f1a393bd9726374fdb7b4880f8427a37d6b76997..89b417ba99d6e378dca2f75045c2ee7c0d81e2c7)*

PR #4988, opened 2026-04-29
tree borrows (implicit writes): add support for #[rustc_no_writable]

quiode:
This PR enables support for the `#[rustc_no_writable]` attribute. If implicit writes checking in tree borrows is enabled, this attribute can be used to disable the checking on a per function basis.
A test case has been provided to demonstrate the usage and tests that it works as intended.

### DIFF
--- a/src/borrow_tracker/tree_borrows/mod.rs
+++ b/src/borrow_tracker/tree_borrows/mod.rs
@@ -134,9 +134,11 @@ impl<'tcx> NewPermission {
         let ty_is_freeze = pointee.is_freeze(*cx.tcx, cx.typing_env());
         let is_protected = retag_kind == RetagKind::FnEntry;
 
-        // Check if the implicit writes feature is globally enabled, using the `-Zmiri-tree-borrows-implicit-writes` flag, and not locally disabled using the `#[rustc_no_writable]` attribute.
-        // For performance reasons, only performs the lookup if is_protected is true as implicit write checking only works for protected references.
-        let implicit_writes_enabled = if is_protected {
+        // Check if the implicit writes feature is globally enabled, using the
+        // `-Zmiri-tree-borrows-implicit-writes` flag, and not locally disabled using the
+        // `#[rustc_no_writable]` attribute. For performance reasons, only performs the lookup if
+        // is_protected is true as implicit writes are only performed for protected references.
+        let implicit_writes_enabled = is_protected && {
             let implicit_writes = cx
                 .machine
                 .borrow_tracker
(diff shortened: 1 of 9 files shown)


RalfJung:
Thanks!

Are there any tests where, in the previous PR, you didn't add a "Tree Borrows + writable" revision because doing so would error (and now it works thanks to the attributes on `slice::as_mut_ptr` and `str::as_mut_ptr`)?

quiode:
Yes. I added the `tree_implicit_writes` to all remaining revisions (in pass) that had one for trees but not one for implicit writes. It seems all work now.

tests/pass/vec.rs
RalfJung, line 1: Nice, this is the test where my previous experiment for implicit writes in Tree Borrows came to a halt. :)

src/borrow_tracker/tree_borrows/mod.rs
```diff
@@ -143,6 +144,9 @@ impl<'tcx> NewPermission {
             .borrow_tracker_method
             .get_tree_borrows_params()
             .implicit_writes;
+        let def_id = cx.frame().instance().def_id();
+        let no_writable = find_attr!(cx.tcx, def_id, RustcNoWritable);
+        let implicit_writes_enabled = implicit_writes && !no_writable;
```
RalfJung: Can you make it so that all this code only runs if `is_protected` is true?

tests/pass/tree_borrows/rustc_no_writable.rs
```diff
@@ -0,0 +1,19 @@
...
+fn main() {
+    let mut x = 0u8;
+    let ptr = &raw mut x;
+    let res = dereference(&mut x, ptr);
+    assert_eq!(*res, 0);
+}
+
+#[rustc_no_writable]
+fn dereference<T>(x: T, y: *mut u8) -> T {
+    // miri inserts an implicit write here
```
RalfJung:
```suggestion
    // miri inserts *no* implicit write here due to the attribute.
    // Therefore we end up only reading from `x` and `y` which is fine.
```

tests/pass/tree_borrows/rustc_no_writable.rs
```diff
@@ -0,0 +1,19 @@
...
+    let _ = unsafe { *y };
+    // x = 42; // we'd like to add this write, so there must already be UB without it because there sure is with it.
```
RalfJung: This comment makes no sense here I think?

RalfJung (commented):
Nice :)
Just a few minor comments.

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/4988/6961dabc90309b36a981e56e08522a38fc405e22..9329130427ed5ccd0abeb2ad191f835ac3d1ddbf)*

quiode:
done
@rustbot ready

PR #4996, opened 2026-05-02
freebsd: sched_get/setaffinity work on IDs returned by pthread_getthreadid_np

RalfJung:
Also see the discussion in https://github.com/rust-lang/libc/issues/5095.

### DIFF
--- a/src/concurrency/blocking_io.rs
+++ b/src/concurrency/blocking_io.rs
@@ -219,7 +219,7 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
         let this = self.eval_context_mut();
         this.machine.blocking_io.register(
             source_fd,
-            InterestReceiver::UnblockThread(this.machine.threads.active_thread()),
+            InterestReceiver::UnblockThread(this.active_thread()),
             interests,
         );
         this.block_thread(BlockReason::IO, timeout, callback);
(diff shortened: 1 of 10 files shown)


RalfJung:
@YohDeadfall  @LorrensP-2158466 any chance you could try the libc-affinity test from this PR on a real system to confirm that indeed `pthread_getthreadid_np` works with the affinity functions?

YohDeadfall:
I may have some time tomorrow, on Sunday, or during the next week, but no promises due to my workload. Though the next weekend seems absolute mine.

RalfJung:
Looking at https://man.freebsd.org/cgi/man.cgi?query=cpuset_getaffinity&sektion=2&manpath=FreeBSD+10.3-RELEASE and https://man.freebsd.org/cgi/man.cgi?query=cpuset&sektion=2&apropos=0&manpath=FreeBSD+10.3-RELEASE, I am less and less sure that this is actually correct. Sadly the docs are pretty unclear.

It's probably best to avoid adding assumptions like "threads IDs and process IDs are basically the same thing". That is true on Linux, but FreeBSD has separate CPU_WHICH_TID and CPU_WHICH_PID and it has separate types `lwpid_t` and `pid_t` so... let's wait until someone actually asks for anything like this; that person will then hopefully have the knowledge and motivation to figure out and explain the rules for these operations on FreeBSD. :)

PR #5004, opened 2026-05-04
Remove `WSAESHUTDOWN` workaround for network socket writes

WhySoBad:
With the merger of https://github.com/rust-lang/rust/pull/156063, we no longer need to have a workaround for mapping WSAESHUTDOWN error codes to `io::ErrorKind::BrokenPipe`. This is now part of the standard library.

The error kind mapping is not added to the `UNIX_IO_ERROR_TABLE` because `ERROR_NO_DATA` already maps to `io::ErrorKind::BrokenPipe`.


RalfJung:
> The error kind mapping is not added to the UNIX_IO_ERROR_TABLE because ERROR_NO_DATA already maps to io::ErrorKind::BrokenPipe.

As the comment explains, multi-mappings are possible and useful -- but only for Windows *targets* which isn't what you are working on so :shrug:

PR #5005, opened 2026-05-06
Add basic `getsockopt` shim for reading `SO_ERROR`

WhySoBad:
Hi,

This pull request adds the `getsockopt` shim. Currently, it only supports reading the `SO_ERROR` option. 

By this, Miri is now able to run programs which use tokio/mio sockets. Thus, I also added a simple tokio socket test which tests connecting/accepting and reading/writing. Let me know if I should test more things with the tokio sockets. 
I've also ran the tokio TCP test suite and we pass everything except:
- Some tests which are using syscalls which aren't shimmed (mainly socket options)
- One test failing due to a Miri panic when the write end of a connecting socket is closed before it received a writable event
- One test failing due to Miri not clearing the readiness after a short read/write on a linux target 

For the latter two points I'll soon open pull requests here.

### DIFF
--- a/src/shims/unix/socket.rs
+++ b/src/shims/unix/socket.rs
@@ -7,6 +7,7 @@ use std::time::Duration;
 
 use mio::event::Source;
 use mio::net::{TcpListener, TcpStream};
+use rustc_abi::Size;
 use rustc_const_eval::interpret::{InterpResult, interp_ok};
 use rustc_middle::throw_unsup_format;
 use rustc_target::spec::Os;
@@ -58,6 +59,8 @@ struct Socket {
     is_non_block: Cell<bool>,
     /// The current blocking I/O readiness of the file description.
     io_readiness: RefCell<BlockingIoSourceReadiness>,
+    /// [`Some`] when the socket had an async error which has not yet been fetched via `SO_ERROR`.
+    error: RefCell<Option<io::Error>>,
 }
 
 impl FileDescription for Socket {
@@ -340,6 +343,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             state: RefCell::new(SocketState::Initial),
             is_non_block: Cell::new(is_sock_nonblock),
             io_readiness: RefCell::new(BlockingIoSourceReadiness::empty()),
+            error: RefCell::new(None),
         });
 
         interp_ok(Scalar::from_i32(fds.insert(fd)))
@@ -950,6 +954,152 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         );
     }
 
+    fn getsockopt(
+        &mut self,
+        socket: &OpTy<'tcx>,
+        level: &OpTy<'tcx>,
+        option_name: &OpTy<'tcx>,
+        option_value: &OpTy<'tcx>,
+        option_len: &OpTy<'tcx>,
+    ) -> InterpResult<'tcx, Scalar> {
+        let this = self.eval_context_mut();
+
+        let socket = this.read_scalar(socket)?.to_i32()?;
+        let level = this.read_scalar(level)?.to_i32()?;
+        let option_name = this.read_scalar(option_name)?.to_i32()?;
+        // These two pointers are used to return the value: `len_ptr` initially stores how much space
+        // is available. If the actual value fits into that space, it is written to
+        // `value_ptr` and `len_ptr` is updated to represent how many bytes
+        // were actually written. If the value does not fit, it is silently truncated.
+        // Also see <https://pubs.opengroup.org/onlinepubs/9799919799/functions/getsockopt.html>.
+        let option_value_ptr = this.read_pointer(option_value)?;
+        let option_len_ptr = this.read_pointer(option_len)?;
+
+        // Get the file handle
+        let Some(fd) = this.machine.fds.get(socket) else {
+            return this.set_last_error_and_return_i32(LibcError("EBADF"));
+        };
+
+        let Some(socket) = fd.downcast::<Socket>() else {
+            // Man page specifies to return ENOTSOCK if `fd` is not a socket.
+            return this.set_last_error_and_return_i32(LibcError("ENOTSOCK"));
+        };
+
+        if option_value_ptr == Pointer::null() || option_len_ptr == Pointer::null() {
+            // This socket option returns a value and thus we need to return EFAULT
+            // when either the value or the length pointers are null pointers.
+            return this.set_last_error_and_return_i32(LibcError("EFAULT"));
+        }
+
+        let socklen_layout = this.libc_ty_layout("socklen_t");
+        let option_len_ptr_mplace = this.ptr_to_mplace(option_len_ptr, socklen_layout);
+        let option_len: usize = this
+            .read_scalar(&option_len_ptr_mplace)?
+            .to_int(socklen_layout.size)?
+            .try_into()
+            .unwrap();
+
+        // We need a temporary buffer as `option_value_ptr` might not point to a large enough
+        // buffer, in which case we have to truncate.
+        let value_buffer = if level == this.eval_libc_i32("SOL_SOCKET") {
+            let opt_so_error = this.eval_libc_i32("SO_ERROR");
+
+            if option_name == opt_so_error {
+                // Because `TcpStream::take_error()` and `TcpListener::take_error()` consume the latest async
+                // error, we know that our stored `socket.error` is outdated when `TcpStream::take_error()`/
+                // `TcpListener::take_error()` returns `Ok(Some(...))`.
+                // If they return `Ok(None)`, then we fall back to the stored `socket.error`.
+                let error = match &*socket.state.borrow() {
+                    SocketState::Initial | SocketState::Bound(_) => socket.error.take(),
+                    SocketState::Listening(listener) =>
+                        listener.take_error().unwrap_or(socket.error.take()),
+                    SocketState::Connecting(stream) | SocketState::Connected(stream) =>
+                        stream.take_error().unwrap_or(socket.error.take()),
+                };
+                // Clear our own stored error -- it was either `take`n above or it is outdated.
+                socket.error.replace(None);
+
+                // We know there is no longer an async error and thus we need to update the
+                // I/O and epoll readiness of the socket.
+                socket.io_readiness.borrow_mut().error = false;
+                this.update_epoll_active_events(socket, /* force_edge */ false)?;
+
+                let return_value = match error {
+                    Some(err) => this.io_error_to_errnum(err)?.to_i32()?,
+                    // If there is no error, we write 0 into the option value buffer.
+                    None => 0,
+                };
+
+                // Allocate new buffer on the stack with the `i32` layout.
+                let value_buffer = this.allocate(this.machine.layouts.i32, MemoryKind::Stack)?;
+                this.write_int(return_value, &value_buffer)?;
+                value_buffer
+            } else {
+                throw_unsup_format!(
+                    "getsockopt: option {option_name:#x} is unsupported for level SOL_SOCKET",
+                );
+            }
+        } else if level == this.eval_libc_i32("IPPROTO_IP") {
+            let opt_ip_ttl = this.eval_libc_i32("IP_TTL");
+
+            if option_name == opt_ip_ttl {
+                let ttl = match &*socket.state.borrow() {
+                    SocketState::Initial | SocketState::Bound(_) =>
+                        throw_unsup_format!(
+                            "getsockopt: reading option IP_TTL on level IPPROTO_IP is only supported \
+                            on connected and listening sockets"
+                        ),
+                    SocketState::Listening(listener) => listener.ttl(),
+                    SocketState::Connecting(stream) | SocketState::Connected(stream) =>
+                        stream.ttl(),
+                };
+
+                let ttl = match ttl {
+                    Ok(ttl) => ttl,
+                    Err(e) => return this.set_last_error_and_return_i32(e),
+                };
+
+                // Allocate new buffer on the stack with the `u32` layout.
+                let value_buffer = this.allocate(this.machine.layouts.u32, MemoryKind::Stack)?;
+                this.write_int(ttl, &value_buffer)?;
+                value_buffer
+            } else {
+                throw_unsup_format!(
+                    "getsockopt: option {option_name:#x} is unsupported for level IPPROTO_IP",
+                );
+            }
+        } else {
+            throw_unsup_format!(
+                "getsockopt: level {level:#x} is unsupported, only SOL_SOCKET is allowed"
+            )
+        };
+
+        // Truncated size of the output value.
+        let output_value_len = value_buffer.layout.size.min(Size::from_bytes(option_len));
+        // Copy the truncated value into the buffer pointed to by `option_value_ptr`.
+        this.mem_copy(
+            value_buffer.ptr(),
+            option_value_ptr,
+            // Truncate the value to fit the provided buffer.
+            output_value_len,
+            // The buffers are guaranteed to not overlap since the `value_buffer`
+            // was just newly allocated on the stack.
+            true,
+        )?;
+        // Deallocate the value buffer as it was only needed to store the value and
+        // copy it into the buffer pointed to by `option_value_ptr`.
+        this.deallocate_ptr(value_buffer.ptr(), None, MemoryKind::Stack)?;
+
+        // On output, the length pointer contains the amount of bytes written -- not the size
+        // of the value before truncation.
+        this.write_scalar(
+            Scalar::from_uint(output_value_len.bytes(), socklen_layout.size),
+            &option_len_ptr_mplace,
+        )?;
+
+        interp_ok(Scalar::from_i32(0))
+    }
+
     fn getsockname(
         &mut self,
         socket: &OpTy<'tcx>,
@@ -1232,6 +1382,7 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             state: RefCell::new(SocketState::Connected(stream)),
             is_non_block: Cell::new(is_client_sock_nonblock),
             io_readiness: RefCell::new(BlockingIoSourceReadiness::empty()),
+            error: RefCell::new(None),
         });
         // Register the socket to the blocking I/O manager because
         // there is an associated host socket.
@@ -1490,17 +1641,18 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     };
 
                     // Manually check whether there were any errors since calling `connect`.
-                    if let Ok(Some(_)) = stream.take_error() {
+                    if let Ok(Some(err)) = stream.take_error() {
                         // There was an error during connecting and thus we
                         // return ENOTCONN. It's the program's responsibility
                         // to read SO_ERROR itself.
-                        //
+
+                        // Store the error such that we can return it when
+                        // `getsockopt(SOL_SOCKET, SO_ERROR, ...)` is called on the socket.
+                        socket.error.replace(Some(err));
+
                         // Go back to initial state since the only way of getting into the
                         // `Connecting` state is from the `Initial` state and at this point
                         // we know that the connection won't be established anymore.
-                        //
-                        // FIXME: We're currently just dropping the error information. Eventually
-                        // we'll have to store it so that it can be recovered by the user.
                         *state = SocketState::Initial;
                         drop(state);
                         return action.call(this, Err(()))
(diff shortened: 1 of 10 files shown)


src/shims/unix/socket.rs
```diff
@@ -58,6 +58,9 @@ struct Socket {
     is_non_block: Cell<bool>,
     /// The current blocking I/O readiness of the file description.
     io_readiness: RefCell<BlockingIoSourceReadiness>,
+    /// [`Some`] when the socket had an async error which was consumed by
+    /// [`TcpStream::take_error`] but not yet handled.
```
RalfJung: I am confused by this comment.
```suggestion
    /// [`Some`] when the socket had an async error which has not yet been fetched via `SO_ERROR`.
```
Is that what you mean? Why is `take_error` being mentioned?

src/shims/unix/socket.rs
```diff
@@ -950,6 +954,100 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    // when either the value or the length pointers are null pointers.
+                    return this.set_last_error_and_return_i32(LibcError("EFAULT"));
+                }
+
+                let error = match &*socket.state.borrow() {
+                    SocketState::Initial | SocketState::Bound(_) => socket.error.take(),
+                    SocketState::Listening(listener) =>
+                        listener.take_error().unwrap_or(socket.error.take()),
+                    SocketState::Connecting(stream) | SocketState::Connected(stream) =>
+                        stream.take_error().unwrap_or(socket.error.take()),
```
RalfJung, line 1017: Naively I would have expected the order to be the other way around (first check `socket.error`, then fall back to the actual stream)... or can this not matter?

src/shims/unix/socket.rs
```diff
@@ -950,6 +954,100 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                };
+                // Clear the stored error in case there was a new error and
+                // we thus didn't consume the stored error.
+                socket.error.replace(None);
+
+                // We know there is no longer an async error and thus we need to update the
+                // readiness in the blocking I/O manager.
+                socket.io_readiness.borrow_mut().error = false;
```
RalfJung, line 1024: Remember to call the method that updates the epoll based on this!

src/shims/unix/socket.rs
```diff
@@ -950,6 +954,100 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    .unwrap();
+
+                let return_value = match error {
+                    Some(err) => this.io_error_to_errnum(err)?.to_i32()?,
+                    // If there is no error, we write 0 into the option value buffer.
+                    None => 0,
+                };
+
+                // Value is at most 4 bytes because it's an i32.
+                let value_len = option_len.min(4);
```
RalfJung: I can't find anything in the docs that says that we do partial writes if the length is too short? Shouldn't we return an error instead?

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -112,6 +116,50 @@ fn test_accept_nonblock() {
...
+        // Change client socket to be non-blocking.
+        errno_check(libc::fcntl(client_sockfd, libc::F_SETFL, libc::O_NONBLOCK));
+    }
+
+    // We cannot attempt to connect to a localhost address because
+    // it could be the case that a socket from another test is
+    // currently listening on `localhost:12321` because we bind to
+    // random ports everywhere. For `127.0.1.1` we know that it's a loopback
+    // address and thus exists but because it's not the standard loopback
+    // address we also that nothing is bound to it.
```
```suggestion
    // address we also assume that nothing is bound to it.
```

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -112,6 +116,50 @@ fn test_accept_nonblock() {
...
+        epfd,
+        &[Ev { events: libc::EPOLLERR | EPOLLOUT | EPOLLHUP, data: client_sockfd }],
+        -1,
+    );
+
+    let errno =
+        net::getsockopt::<libc::c_int>(client_sockfd, libc::SOL_SOCKET, libc::SO_ERROR).unwrap();
+    // Depending on the host we receive different error kinds. Thus, we only check
+    // that it's a nonzero error code.
+    assert!(errno != 0)
```
```suggestion
    assert!(errno != 0);
```

RalfJung (commented):
Amazing how everything is coming together and now we can actually run that Tokio smoke test. :)

I still have some comments though, of course. ;)
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5005/0f465c2ef352cba1299a8debc057c7f93cdeb3f6..8b09c4b4afb131afff857f5a37d3f63a19be825a)*

src/shims/unix/socket.rs
WhySoBad: The POSIX man page about `getsockopt` states:
"The `getsockopt()` function shall retrieve the value for the option specified by the `option_name` argument for the socket specified by the socket argument. If the size of the option value is greater than `option_len`, the value stored in the object pointed to by the `option_value` argument shall be silently truncated. Otherwise, the object pointed to by the `option_len` argument shall be modified to indicate the actual length of the value."

However, the Linux man page indeed doesn't contain this section.

src/shims/unix/socket.rs
WhySoBad: Yes, that's what I meant :)

src/shims/unix/socket.rs
WhySoBad: The POSIX man page states:
"The SO_ERROR option is used only on [getsockopt()](https://pubs.opengroup.org/onlinepubs/009695399/functions/getsockopt.html). When this option is specified, [getsockopt()](https://pubs.opengroup.org/onlinepubs/009695399/functions/getsockopt.html) shall return any pending error on the socket and clear the error status."

As far as I understood, SO_ERROR should always contain the latest async error. Thus, when `TcpStream::take_error` or `TcpListener::take_error` returns an `Ok(Some(...))` then this error should always be newer than the error we already stored. 
I've looked at how TCP errors are set [1] and consumed (via `getsockopt(..., SO_ERROR, ...)`) [2] in the Linux kernel, and as far as I can tell this just overwrites the previous error:
[1]: https://github.com/torvalds/linux/blob/fcee7d82f27d6a8b1ddc5bbefda59b4e441e9bc0/net/ipv4/tcp_ipv4.c#L644-L650
[2]: https://github.com/torvalds/linux/blob/fcee7d82f27d6a8b1ddc5bbefda59b4e441e9bc0/include/net/sock.h#L2531-L2543

src/shims/unix/socket.rs
RalfJung: Ah, that's why I didn't see it locally.

Can you confirm that Linux indeed silently truncates this? Also, we should have tests for the "`option_len` too big" and "`option_len` too small" cases.

src/shims/unix/socket.rs
RalfJung: >  SO_ERROR should always contain the latest async error

If that's the principle, then please explain it in a comment, and also explain why the stream's error is newer than what we stored in `socket.error`.

tests/utils/libc.rs
```diff
@@ -388,6 +388,32 @@ pub mod net {
...
+
+        errno_result(unsafe {
+            libc::getsockopt(
+                sockfd,
+                level,
+                option_name,
+                option_value.as_mut_ptr().cast(),
+                &mut option_len,
+            )
+        })?;
```
RalfJung, line 413: Please add an `assert!` ensuring that after the call, `option_len` is still the same size, to ensure the function actually returned as many bytes as we expected it to return.

src/shims/unix/socket.rs
```diff
@@ -950,6 +953,101 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+                // If the buffer is smaller than 4 bytes, silently truncate the bytes to fill the
+                // provided buffer.
+                this.write_bytes_ptr(
+                    option_value_ptr,
+                    return_value.to_ne_bytes().into_iter().take(value_len),
+                )?;
```
RalfJung: This looks wrong, the endianess likely has to depend on the target.

Also the entire logic will just be annoying to deal with and has to be repeated for each case. I'd recommend a different approach, where we allocate a buffer that has the right size, and then at the bottom of this function there's code that copies a part of the buffer to `option_value_ptr`. Then we only have to handle this annoying truncation stuff once.

Or maybe you can think of another nice way to do this -- but I'd really like to avoid having endianess logic here.

src/shims/unix/socket.rs
```diff
@@ -950,6 +953,101 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        option_name: &OpTy<'tcx>,
+        option_value: &OpTy<'tcx>,
+        option_len: &OpTy<'tcx>,
+    ) -> InterpResult<'tcx, Scalar> {
+        let this = self.eval_context_mut();
+
+        let socket = this.read_scalar(socket)?.to_i32()?;
+        let level = this.read_scalar(level)?.to_i32()?;
+        let option_name = this.read_scalar(option_name)?.to_i32()?;
+        let option_value_ptr = this.read_pointer(option_value)?;
```
RalfJung, line 975:
```suggestion
        // These two pointers are used to return the value: `len_ptr` initially stores how much space
        // is available. If the actual value fits into that space, it is written to
        // `value_ptr` and `len_ptr` is updated to represent how many bytes
        // were actually written. If the value does not fit, it is silently truncated.
        // Also see <https://pubs.opengroup.org/onlinepubs/9799919799/functions/getsockopt.html>.
        let option_value_ptr = this.read_pointer(option_value)?;
```

RalfJung (commented):
That truncation of the value is really annoying, we may have to iterate a bit until we find a good way to deal with that.
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5005/0f465c2ef352cba1299a8debc057c7f93cdeb3f6..601a3c62ad7c0b852558b0ad792c3c9456ce728f)*

src/shims/unix/socket.rs
WhySoBad: >Can you confirm that Linux indeed silently truncates this?

Yes, it does silently truncate it on my host.

>Also, we should have tests for the "option_len too big" and "option_len too small" cases.

It's a bit difficult to test that with only SO_ERROR because the error codes differ on the different hosts. 
Ideally, we would test this with a socket option where we can use `setsockopt` to set it to a specified value and then compare it with the result of `getsockopt` with a too big/small buffer.
I'd suggest adding such tests once we add more socket options (which will be pretty soon) or if you want I can also just add another socket option in this PR such that we can test it better.

RalfJung: > It's a bit difficult to test that with only SO_ERROR because the error codes differ on the different hosts.

Yeah the test would then just check that it's a prefix of the actual value used on the current host (which it can determine by running it once with the right size).

> or if you want I can also just add another socket option in this PR such that we can test it better.

What's the simplest socketoption you could add to be used for the test?

WhySoBad: > What's the simplest socketoption you could add to be used for the test?

`IP_TTL` should be pretty easy, it's just an `u32` and we can use the `TcpStream::set_ttl()`/`TcpStream::ttl()` methods for setting/getting it.

RalfJung: So that would be only the "get" part then for this PR? Okay, that works for me. Though couldn't that also differ between targets?

src/shims/unix/socket.rs
WhySoBad: I've now changed it to use a temporary buffer for storing the actual value. The value is then copied truncated into the actual buffer.
Is this what you were thinking of?

src/shims/unix/socket.rs
```diff
@@ -950,6 +954,123 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                // If they return `Ok(None)`, then we fall back to the stored `socket.error`.
+                let error = match &*socket.state.borrow() {
+                    SocketState::Initial | SocketState::Bound(_) => socket.error.take(),
+                    SocketState::Listening(listener) =>
+                        listener.take_error().unwrap_or(socket.error.take()),
+                    SocketState::Connecting(stream) | SocketState::Connected(stream) =>
+                        stream.take_error().unwrap_or(socket.error.take()),
+                };
+                // Clear the stored error in case there was a new error and
+                // we thus didn't consume the stored error.
```
```suggestion
                // Clear our own stored error -- it was either `take`n above or it is outdated.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5005/0f465c2ef352cba1299a8debc057c7f93cdeb3f6..947542f855b884671c4a565a5b9c777953aa4337)*

src/shims/unix/socket.rs
```diff
@@ -1005,47 +1023,52 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                // of the value before truncation.
-                this.write_scalar(
-                    Scalar::from_int(i128::try_from(value_len).unwrap(), socklen_layout.size),
-                    &option_len_ptr_mplace,
-                )?;
-
-                return interp_ok(Scalar::from_i32(0));
+                // Allocate new buffer on the stack with the `i32` layout.
+                // We need a temporary buffer as `option_value_ptr` might not point to a large enough
+                // buffer, in which case we have to truncate.
```
RalfJung: This comment should be at the top near the `let (value_buffer, value_layout)`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5005/0f465c2ef352cba1299a8debc057c7f93cdeb3f6..947542f855b884671c4a565a5b9c777953aa4337)*

src/shims/unix/socket.rs
```diff
@@ -1005,47 +1023,52 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
 
-        throw_unsup_format!(
-            "getsockopt: level {level:#x} is unsupported, only SOL_SOCKET is allowed"
-        );
+        // Copy the truncated value into the buffer pointed to by `option_value_ptr`.
+        this.mem_copy(
+            value_buffer.ptr(),
+            option_value_ptr,
+            // Truncate the value to fit the provided buffer.
+            value_layout.size.min(Size::from_bytes(option_len)),
```
RalfJung: Please let-bind this so you can reuse it below.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5005/0f465c2ef352cba1299a8debc057c7f93cdeb3f6..947542f855b884671c4a565a5b9c777953aa4337)*

src/shims/unix/socket.rs
RalfJung: Yes that looks great. :)

src/shims/unix/socket.rs
WhySoBad: Yeah, but we could also just get the value once with all 4 bytes and then check whether the first two bytes match when we only get it with a 2 byte buffer.

src/shims/unix/socket.rs
```diff
@@ -1002,50 +1019,56 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 socket.error.replace(None);
 
                 // We know there is no longer an async error and thus we need to update the
-                // readiness in the blocking I/O manager.
+                // I/O and epoll readiness of the socket.
```
RalfJung, line 1023: Is that really how this works? The error readiness gets cleared on getsockopt(SO_ERROR)?

Is that covered by a test? Please confirm that test on a native host.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5005/0f465c2ef352cba1299a8debc057c7f93cdeb3f6..947542f855b884671c4a565a5b9c777953aa4337)*

src/shims/unix/socket.rs
RalfJung: Sure, that works.

src/shims/unix/socket.rs
WhySoBad: > Please confirm that test on a native host.

Yes, it gets cleared on my host. I'll add a test for this.

src/shims/unix/socket.rs
```diff
@@ -950,6 +954,123 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let return_value = match error {
+                    Some(err) => this.io_error_to_errnum(err)?.to_i32()?,
+                    // If there is no error, we write 0 into the option value buffer.
+                    None => 0,
+                };
+
+                // Allocate new buffer on the stack with the `i32` layout.
+                let value_buffer = this.allocate(this.machine.layouts.i32, MemoryKind::Stack)?;
+                this.write_int(return_value, &value_buffer)?;
+                (value_buffer, this.machine.layouts.i32)
```
RalfJung: Can't we access the layout below via `value_buffer.layout`? It shouldn't be necessary to store it separately.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5005/0f465c2ef352cba1299a8debc057c7f93cdeb3f6..9e750568f45ccc17d260350d4fe14f7030bacfa1)*

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -112,6 +116,54 @@ fn test_accept_nonblock() {
...
+    );
+
+    let errno =
+        net::getsockopt::<libc::c_int>(client_sockfd, libc::SOL_SOCKET, libc::SO_ERROR).unwrap();
+    // Depending on the host we receive different error kinds. Thus, we only check
+    // that it's a nonzero error code.
+    assert!(errno != 0);
+
+    let readiness = current_epoll_readiness::<8>(client_sockfd, EPOLLET | EPOLLOUT | EPOLLERR);
+    // Ensure that error readiness is cleared after reading SO_ERROR.
```
```suggestion
    // Ensure that error readiness is cleared after reading SO_ERROR.
    let readiness = current_epoll_readiness::<8>(client_sockfd, EPOLLET | EPOLLOUT | EPOLLERR);
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5005/0f465c2ef352cba1299a8debc057c7f93cdeb3f6..9e750568f45ccc17d260350d4fe14f7030bacfa1)*

src/shims/unix/socket.rs
WhySoBad: I've now also changed this for the `write_socket_address` method.

WhySoBad:
I think if you restart the failed CI job then everything should be fine.

@rustbot ready

src/shims/unix/socket.rs
```diff
@@ -1030,28 +1031,63 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             }
+        } else if level == this.eval_libc_i32("IPPROTO_IP") {
+            let opt_ip_ttl = this.eval_libc_i32("IP_TTL");
+
+            if option_name == opt_ip_ttl {
+                if option_value_ptr == Pointer::null() || option_len_ptr == Pointer::null() {
+                    // This socket option returns a value and thus we need to return EFAULT
+                    // when either the value or the length pointers are null pointers.
+                    return this.set_last_error_and_return_i32(LibcError("EFAULT"));
+                }
```
RalfJung: Can't we have this check just once at the top for all options?

RalfJung (commented):
Very nice. :)

Just one last question. Please also squash the commits.
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5005/0f465c2ef352cba1299a8debc057c7f93cdeb3f6..97411706556fd290a405cf4ca114e27a950438b2)*

src/shims/unix/socket.rs
```diff
@@ -950,6 +954,158 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+                let ttl = match &*socket.state.borrow() {
+                    SocketState::Initial | SocketState::Bound(_) =>
+                        throw_unsup_format!(
+                            "getsockopt: reading option IP_TTL on level IPPROTO_IP is only supported \
+                            on connected and listening sockets"
+                        ),
+                    SocketState::Listening(listener) => listener.ttl(),
+                    SocketState::Connecting(stream) | SocketState::Connected(stream) =>
+                        stream.ttl(),
```
RalfJung, line 1054: Does this mean we also support calling `ttl()` via std? Please add a test for that.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5005/0f465c2ef352cba1299a8debc057c7f93cdeb3f6..97411706556fd290a405cf4ca114e27a950438b2)*

PR #5007, opened 2026-05-07
try to remove generic normalization from backtraces

RalfJung:
I forgot why we do this, but it leads to some very strange results like
```
           0: <std::sync::MutexGuard<'_, u64> as std::ops::Drop>::drop
               at RUSTLIB/std/src/sync/poison/mutex.rs:LL:CC
           1: std::ptr::drop_glue))
               at RUSTLIB/core/src/ptr/mod.rs:LL:CC
           2: std::ptr::drop_glue))
               at RUSTLIB/core/src/ptr/mod.rs:LL:CC
```


RalfJung:
@oli-obk I think you added that normalization filter? I can't quite remember. Assuming CI passes, do you have any objections to removing it again? With the way we now rendering regular errors, it also applies to them, and I think it'd be good to see what we actually show to users here -- and also the normalization introduces some quite confusing artifacts.

oli-obk:
I think i only added it because there were weird CI failures from it somewhere. Let's remove it and see if anything crops up

PR #5009, opened 2026-05-08
Remove `TcpStream::peer_addr` check for network sockets, test epoll readiness for `shutdown`

WhySoBad:
As mentioned in rust-lang/miri#5005, this pull request fixes a bug with `shutdown` and non-blocking sockets. Additionally, it also adds tests for `shutdown` in combination with `epoll`.

The bug occurs when we connect a non-blocking socket and shut down the write end of the connecting socket before the socket became connected. As a consequence, the `TcpStream::peer_addr` call in the `ensure_connected` method fails because after closing the write end, we can no longer get the peer. Our current logic treats this failure as an OS error and panics.
Because tokio doesn't implement this (see https://github.com/tokio-rs/mio/issues/1942#issuecomment-4162607761) suggested check with `TcpStream::peer_addr` and everything seems to work fine, we should also be fine with removing the check and by this resolving the panic.

### DIFF
--- a/src/shims/unix/socket.rs
+++ b/src/shims/unix/socket.rs
@@ -602,11 +602,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 ),
         }
 
-        // Mio returns a potentially unconnected stream.
-        // We can be ensured that the connection is established when
-        // [`TcpStream::take_err`] and [`TcpStream::peer_addr`] both
-        // don't return an error after receiving an [`Interest::WRITEABLE`]
-        // event on the stream.
+        // This begins establishing the connection, but does not block until the stream is fully connected.
+        // We deal with that below.
         match TcpStream::connect(address) {
             Ok(stream) => {
                 *socket.state.borrow_mut() = SocketState::Connecting(stream);
@@ -1112,22 +1109,45 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             return this.set_last_error_and_return_i32(LibcError("ENOTCONN"));
         };
 
-        let shut_rd = this.eval_libc_i32("SHUT_RD");
-        let shut_wr = this.eval_libc_i32("SHUT_WR");
-        let shut_rdwr = this.eval_libc_i32("SHUT_RDWR");
+        let is_read_shutdown = how == this.eval_libc_i32("SHUT_RD");
+        let is_write_shutdown = how == this.eval_libc_i32("SHUT_WR");
+        let is_read_write_shutdown = how == this.eval_libc_i32("SHUT_RDWR");
 
         let how = match () {
-            _ if how == shut_rd => Shutdown::Read,
-            _ if how == shut_wr => Shutdown::Write,
-            _ if how == shut_rdwr => Shutdown::Both,
+            _ if is_read_shutdown => Shutdown::Read,
+            _ if is_write_shutdown => Shutdown::Write,
+            _ if is_read_write_shutdown => Shutdown::Both,
             // An invalid value was passed to `how`.
             _ => return this.set_last_error_and_return_i32(LibcError("EINVAL")),
         };
 
-        match stream.shutdown(how) {
-            Ok(_) => interp_ok(Scalar::from_i32(0)),
-            Err(e) => this.set_last_error_and_return_i32(e),
-        }
+        if let Err(e) = stream.shutdown(how) {
+            return this.set_last_error_and_return_i32(e);
+        };
+
+        drop(state);
+
+        // Because we map cross platform mio readiness to epoll readiness and
+        // the different platforms don't treat `shutdown` the same way, we set
+        // the readiness after a `shutdown` manually to achieve more consistent
+        // epoll readiness. Otherwise we do not generate enough epoll events
+        // on partial shutdowns on Windows hosts.
+        let mut readiness = socket.io_readiness.borrow_mut();
+        // Closing the read end of a socket causes an EPOLLRDHUP event.
+        readiness.read_closed |= is_read_shutdown || is_read_write_shutdown;
+        // Only shutting down the write end doesn't cause an EPOLLHUP event
+        // and thus we won't set the `write_closed` readiness for it here.
+        readiness.write_closed |= is_read_write_shutdown;
+        // The Linux kernel also sets EPOLLIN when both ends of a socket are closed:
+        // <https://github.com/torvalds/linux/blob/HEAD/net/ipv4/tcp.c#L584-L588>
+        readiness.readable |= is_read_write_shutdown;
+
+        drop(readiness);
+
+        // Update the epoll readiness for the socket.
+        this.update_epoll_active_events(socket, /* force_edge */ false)?;
+
+        interp_ok(Scalar::from_i32(0))
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/shims/unix/socket.rs
```diff
@@ -1506,24 +1523,18 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                        Err(_) => {
-                            // For all other errors the socket is connected. Since we're not interested in the
-                            // peer address and only want to know whether the socket is connected, we can ignore
-                            // the error and continue.
-                        }
-                    }
+                    //
+                    // Attempting to read the peer address would introduce an edge-case where the
+                    // write end of the socket could already be closed before it received a
+                    // writable event. When we then call [`TcpStream::peer_addr`] we receive an
```
```suggestion
                    // write end of the socket could already be shutdown before it received a
                    // writable event. When we then call [`TcpStream::peer_addr`] we receive an
```

src/shims/unix/socket.rs
```diff
@@ -1506,24 +1523,18 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                    // error. This would need extra state for storing whether the write end was
+                    // manually closed using `shutdown`.
+                    // Also, tokio doesn't read the peer address and everything seems to be fine,
+                    // so we don't do that either:
+                    // <https://github.com/tokio-rs/mio/issues/1942#issuecomment-4162607761>
```
RalfJung, line 1540:
```suggestion
                    // <https://github.com/tokio-rs/mio/issues/1942#issuecomment-4162607761>
                    // In other words, we are assuming that there will be no spurious
                    // wakeups while establishing the connection.
```

src/shims/unix/socket.rs
```diff
@@ -1112,22 +1110,41 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        // Because we map cross platform mio readiness to epoll readiness and
+        // the different platforms don't treat `shutdown` the same way, we set
+        // the readiness after a `shutdown` manually to achieve more consistent
+        // epoll readiness.
+        let mut readiness = socket.io_readiness.borrow_mut();
+        // Closing the read end of a socket causes an EPOLLRDHUP event.
+        readiness.read_closed |= is_read_shutdown || is_read_write_shutdown;
+        // Only shutting down the write end doesn't cause an EPOLLHUP event
+        // and thus we won't set the `write_closed` readiness for it here.
+        readiness.write_closed |= is_read_write_shutdown;
```
RalfJung, line 1140: Have you confirmed that this matches what Linux does?

src/shims/unix/socket.rs
```diff
@@ -1112,22 +1110,41 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        if let Err(e) = stream.shutdown(how) {
+            return this.set_last_error_and_return_i32(e);
+        };
+
+        drop(state);
+
+        // Because we map cross platform mio readiness to epoll readiness and
+        // the different platforms don't treat `shutdown` the same way, we set
+        // the readiness after a `shutdown` manually to achieve more consistent
+        // epoll readiness.
```
RalfJung: What goes wrong (on which hosts) if we don't do this?

src/shims/unix/socket.rs
```diff
@@ -603,10 +603,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         }
 
         // Mio returns a potentially unconnected stream.
```
```suggestion
        // Mio does not wait until the stream is fully connected.
```

src/shims/unix/socket.rs
```diff
@@ -603,10 +603,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-        // We can be ensured that the connection is established when
-        // [`TcpStream::take_err`] and [`TcpStream::peer_addr`] both
-        // don't return an error after receiving an [`Interest::WRITEABLE`]
-        // event on the stream.
+        // To ensure that the stream is connected, the
+        // [`ensure_connected`] method should be used.
```
```suggestion
        // We deal with that below.
```

RalfJung (commented):
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5009/a3517b45f998164a4d96f71714b8ac0d45592cc8..2a96cd7c7cb7ca7c067f9a083fde23af1c8784f1)*

tests/pass-dep/libc/libc-socket.rs
RalfJung, line 1: > The bug occurs when we connect a non-blocking socket and shut down the write end of the connecting socket before the socket became connected. As a consequence, the TcpStream::peer_addr call in the ensure_connected method fails because after closing the write end, we can no longer get the peer. Our current logic treats this failure as an OS error and panics.

Is it possible to add / extend a test to cover that case?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5009/a3517b45f998164a4d96f71714b8ac0d45592cc8..2a96cd7c7cb7ca7c067f9a083fde23af1c8784f1)*

src/shims/unix/socket.rs
WhySoBad: Yes, it's pretty much the exact same as the kernel does:
https://github.com/torvalds/linux/blob/aa54b1d27fe0c2b78e664a34fd0fdf7cd1960d71/net/ipv4/tcp.c#L584-L588

src/shims/unix/socket.rs
WhySoBad: On Windows hosts the tests which only shut down the read/write end of the socket hang indefinitely otherwise.

tests/pass-dep/libc/libc-socket.rs
WhySoBad: We cannot really test this because we need a socket which is stuck at connecting to a address + port and to which address + port we later bind a listener. Because we need to use randomized ports to avoid "address is already used" errors, we cannot know such a port before binding (listening to get a `TcpListener`).

src/shims/unix/socket.rs
WhySoBad: (and the tests also pass on my Linux host)

RalfJung: The kernel seems to also set `EPOLLIN`?

src/shims/unix/socket.rs
RalfJung:
```suggestion
        // Because we map cross platform mio readiness to epoll readiness and
        // the different platforms don't treat `shutdown` the same way, we set
        // the readiness after a `shutdown` manually to achieve more consistent
        // epoll readiness. Otherwise we do not generate enough epoll events
        // on partial shutdowns on Windows hosts.
```

tests/pass-dep/libc/libc-socket.rs
RalfJung: We could pick something like `127.0.0.208:23754` that is unlikely to be used.

src/shims/unix/socket.rs
```diff
@@ -602,11 +602,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 ),
         }
 
-        // Mio returns a potentially unconnected stream.
-        // We can be ensured that the connection is established when
-        // [`TcpStream::take_err`] and [`TcpStream::peer_addr`] both
-        // don't return an error after receiving an [`Interest::WRITEABLE`]
-        // event on the stream.
+        // Mio does not wait until the stream is fully connected.
```
RalfJung: On second thought, here's a better wording:
```suggestion
        // This begins establishing the connection, but does not block until the stream is fully connected.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5009/a3517b45f998164a4d96f71714b8ac0d45592cc8..5b543297262eb7aebee5f83326fa50ee0147c4b4)*

tests/pass-dep/libc/libc-socket.rs
WhySoBad: No, because we run multiple targets on the same runner. This means that a socket from a previous target could still be bound to this address and thus the test would fail.

RalfJung: Jobs for different targets run one after the other.

WhySoBad: Yes, but I think the address is still bound for the `TIME_WAIT` duration to ensure the TCP close handshake is finished. 

Even if this weren't the case, we still have the problem that we need to attempt to connect to a local address because we need to listen on it later. But because at the time of the `connect` nothing is bound/listening on the local address, the kernel immediately "fails" the connection attempt and so we cannot start listening later on to achieve a successful connection. When we don't achieve a successful connection, then we never enter the branch which would've paniced.

RalfJung: > Even if this weren't the case, we still have the problem that we need to attempt to connect to a local address because we need to listen on it later. But because at the time of the connect nothing is bound/listening on the local address, the kernel immediately "fails" the connection attempt and so we cannot start listening later on to achieve a successful connection.

Hm... yeah I think I see.

src/shims/unix/socket.rs
```diff
@@ -1130,13 +1130,17 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        // on partial shutdowns on Windows hosts.
         let mut readiness = socket.io_readiness.borrow_mut();
         // Closing the read end of a socket causes an EPOLLRDHUP event.
         readiness.read_closed |= is_read_shutdown || is_read_write_shutdown;
         // Only shutting down the write end doesn't cause an EPOLLHUP event
         // and thus we won't set the `write_closed` readiness for it here.
         readiness.write_closed |= is_read_write_shutdown;
+        // The Linux kernel also sets EPOLLIN when both ends of a socket are closed:
+        // <https://github.com/torvalds/linux/blob/HEAD/net/ipv4/tcp.c#L584-L588>
+        readiness.readable |= is_read_write_shutdown;
```
RalfJung, line 1143: *This* however we can have a test for, right?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5009/a3517b45f998164a4d96f71714b8ac0d45592cc8..7546cae5b845d0ef21297eb4c036569fd84269a4)*

WhySoBad: Yes :)

RalfJung:

@rustbot author

PR #5015, opened 2026-05-09
default all unknown weak extern statics to null

RalfJung:
Unfortunately we can't easily create a zero-filled allocation on-demand when encountering a weak extern. So instead we pre-allocate a ptr-sized null-filled allocation, and then make all ptr-sized weak externs aliases of that allocation. (Non-ptr-sized weak externs are just not supported for now.)

@bjorn3 does that seem like a plausible implementation strategy or is it silly?
In particular I wonder if LLVM is allowed to assume that different weak extern statics resolve to different addresses.

### DIFF
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -17,7 +17,7 @@ use rustc_data_structures::either::Either;
 use rustc_data_structures::fx::{FxHashMap, FxHashSet};
 #[allow(unused)]
 use rustc_data_structures::static_assert_size;
-use rustc_hir::attrs::InlineAttr;
+use rustc_hir::attrs::{InlineAttr, Linkage};
 use rustc_log::tracing;
 use rustc_middle::middle::codegen_fn_attrs::TargetFeatureKind;
 use rustc_middle::mir;
@@ -573,6 +573,8 @@ pub struct MiriMachine<'tcx> {
 
     /// Mapping extern static names to their pointer.
     pub(crate) extern_statics: FxHashMap<Symbol, StrictPointer>,
+    /// A pointer to the allocation we provide for non-existent weak symbols.
+    pub(crate) missing_weak_symbol: Option<StrictPointer>,
 
     /// The random number generator used for resolving non-determinism.
     /// Needs to be queried by ptr_to_int, hence needs interior mutability.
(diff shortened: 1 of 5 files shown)


src/machine.rs
```diff
@@ -1402,8 +1409,22 @@ impl<'tcx> Machine<'tcx> for MiriMachine<'tcx> {
...
             }
             interp_ok(ptr)
+        } else if ecx.tcx.codegen_fn_attrs(def_id).import_linkage == Some(Linkage::ExternalWeak) {
+            // Symbols with weak linkage default to null if they are not defined. However we can't
+            // create new allocations here. So instead we require the static to be ptr-sized and
+            // return a pointer to a pre-allocated allocation that filled with 0s.
+            if extern_decl_layout.size != ecx.tcx.data_layout.pointer_size() {
+                throw_unsup_format!(
+                    "`{link_name}` is a non-pointer-sized weak symbol, which is not supported by Miri",
+                )
```
bjorn3: I this wouldn't even typecheck: https://doc.rust-lang.org/nightly/error_codes/E0791.html

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5015/dd8994ddd4f5dc8564353854f5ea8513aa81807f..0b5faab718187b615ba4f14f6d8477c221487277)*

RalfJung: Ah nice :) Then I can just make this panic.

RalfJung:
All right, based on [Zulip discussions](https://rust-lang.zulipchat.com/#narrow/channel/136281-t-opsem/topic/weak.20symbols/with/594377515) and https://github.com/rust-lang/rust/pull/156467, I'd say this is good to go. :)

src/shims/extern_static.rs
```diff
@@ -77,6 +57,12 @@ impl<'tcx> MiriMachine<'tcx> {
             }
             _ => {} // No "extern statics" supported on this target
         }
+
+        // Also initialize `missing_wea`
```
bjorn3: *missing_weak

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5015/18f9d37db2dbe82d01da75d3d636c4e7bedc8e59..4d8d0f7c511edfa75122cb8cb9bd7d2bb92bc2e8)*

PR #5019, opened 2026-05-09
Add vectored read and vectored write shims for Unix targets

### LINKED ISSUE #4048: unix: add support for libc::readv and libc::writev

WhySoBad:
Hi,

This pull request adds shims for the `readv`, `preadv`, `preadv2`, `writev`, `pwritev` and `pwritev2` functions on Unix targets.

It's implemented the same way as the standard library does it per default for any file description type:
https://github.com/rust-lang/rust/blob/HEAD/library/std/src/io/mod.rs#L538-L552
We search for the first non-zero buffer and read into or write from it.

closes rust-lang/miri#4048

### DIFF
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -5,10 +5,10 @@ use std::io;
 use std::io::ErrorKind;
 
 use rand::RngExt;
-use rustc_abi::Size;
+use rustc_abi::{Align, Size};
 use rustc_target::spec::Os;
 
-use crate::shims::files::FileDescription;
+use crate::shims::files::{DynFileDescriptionRef, FileDescription};
 use crate::shims::sig::check_min_vararg_count;
 use crate::shims::unix::linux_like::epoll::EpollReadiness;
 use crate::shims::unix::*;
@@ -316,7 +316,6 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             .min(u64::try_from(this.target_isize_max()).unwrap())
             .min(u64::try_from(isize::MAX).unwrap());
         let count = usize::try_from(count).unwrap(); // now it fits in a `usize`
-        let communicate = this.machine.communicate();
 
         // Get the FD.
         let Some(fd) = this.machine.fds.get(fd_num) else {
@@ -324,33 +323,17 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             return this.set_last_error_and_return(LibcError("EBADF"), dest);
         };
 
-        // Handle the zero-sized case. The man page says:
-        // > If count is zero, read() may detect the errors described below.  In the absence of any
-        // > errors, or if read() does not check for errors, a read() with a count of 0 returns zero
-        // > and has no other effects.
-        if count == 0 {
-            this.write_null(dest)?;
-            return interp_ok(());
-        }
-        // Non-deterministically decide to further reduce the count, simulating a partial read (but
-        // never to 0, that would indicate EOF).
-        let count = if this.machine.short_fd_operations
-            && fd.short_fd_operations()
-            && count >= 2
-            && this.machine.rng.get_mut().random()
-        {
-            count / 2 // since `count` is at least 2, the result is still at least 1
-        } else {
-            count
-        };
-
         trace!("read: FD mapped to {fd:?}");
         // We want to read at most `count` bytes. We are sure that `count` is not negative
         // because it was a target's `usize`. Also we are sure that it's smaller than
         // `usize::MAX` because it is bounded by the host's `isize`.
 
-        let finish = {
-            let dest = dest.clone();
+        let dest = dest.clone();
+        this.read_from_fd(
+            fd,
+            buf,
+            count,
+            offset,
             callback!(
                 @capture<'tcx> {
                     count: usize,
@@ -363,22 +346,10 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                             // This must fit since `count` fits.
                             this.write_int(u64::try_from(read_size).unwrap(), &dest)
                         }
-                        Err(e) => {
-                            this.set_last_error_and_return(e, &dest)
-                        }
+                        Err(e) => this.set_last_error_and_return(e, &dest)
                 }}
-            )
-        };
-        match offset {
-            None => fd.read(communicate, buf, count, this, finish)?,
-            Some(offset) => {
-                let Ok(offset) = u64::try_from(offset) else {
-                    return this.set_last_error_and_return(LibcError("EINVAL"), dest);
-                };
-                fd.as_unix(this).pread(communicate, offset, buf, count, this, finish)?
-            }
-        };
-        interp_ok(())
+            ),
+        )
     }
 
     fn write(
@@ -402,39 +373,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             .min(u64::try_from(this.target_isize_max()).unwrap())
             .min(u64::try_from(isize::MAX).unwrap());
         let count = usize::try_from(count).unwrap(); // now it fits in a `usize`
-        let communicate = this.machine.communicate();
 
         // We temporarily dup the FD to be able to retain mutable access to `this`.
         let Some(fd) = this.machine.fds.get(fd_num) else {
             return this.set_last_error_and_return(LibcError("EBADF"), dest);
         };
 
-        // Handle the zero-sized case. The man page says:
-        // > If count is zero and fd refers to a regular file, then write() may return a failure
-        // > status if one of the errors below is detected.  If no errors are detected, or error
-        // > detection is not performed, 0 is returned without causing any other effect.   If  count
-        // > is  zero  and  fd refers to a file other than a regular file, the results are not
-        // > specified.
-        if count == 0 {
-            // For now let's not open the can of worms of what exactly "not specified" could mean...
-            this.write_null(dest)?;
-            return interp_ok(());
-        }
-        // Non-deterministically decide to further reduce the count, simulating a partial write.
-        // We avoid reducing the write size to 0: the docs seem to be entirely fine with that,
-        // but the standard library is not (https://github.com/rust-lang/rust/issues/145959).
-        let count = if this.machine.short_fd_operations
-            && fd.short_fd_operations()
-            && count >= 2
-            && this.machine.rng.get_mut().random()
-        {
-            count / 2
-        } else {
-            count
-        };
-
-        let finish = {
-            let dest = dest.clone();
+        let dest = dest.clone();
+        this.write_to_fd(
+            fd,
+            buf,
+            count,
+            offset,
             callback!(
                 @capture<'tcx> {
                     count: usize,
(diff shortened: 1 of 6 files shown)


RalfJung:
> We search for the first non-zero buffer and read into or write from it.

Hm, interesting idea. However it seems a bit unfortunate to never fill more than one of the buffers that we got from "userspace".

The original idea was to
- compute how long all buffers are combined
- create a temporary allocation with that size
- call the normal `read` pointing it at that allocation
- in the `finish` callback, copy things from the temporary allocation to the user-provided buffers

That seems better because then we'll actually exhibit the parts of the program that deal with having output in more than one of these buffers.

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -846,3 +857,244 @@ pub fn check_stat_fields(stat: &libc::stat) {
     let _st_mtime_nsec = stat.st_mtime_nsec;
     let _st_ctime_nsec = stat.st_ctime_nsec;
 }
+
+/// Test that vectored reads perform a single read into the
+/// first non-empty buffer.
+fn test_readv() {
```
RalfJung, line 864: Generally our tests should also pass on real systems, so actively testing for non-standard behavior is... not great.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5019/c70eccef0af89c75e08b4bb9c157e6b0dc37c95d..014c01d40f5f9582c5f1db858f7a042b84b57e08)*

tests/pass-dep/libc/libc-fs.rs
RalfJung, line 1: Can you also add a test that uses the std APIs? (`tests/pass/shims/fs.rs` is probably a good place for that)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5019/c70eccef0af89c75e08b4bb9c157e6b0dc37c95d..014c01d40f5f9582c5f1db858f7a042b84b57e08)*

tests/pass-dep/libc/libc-fs.rs
WhySoBad: Makes sense. The new tests now also pass on my host machine.

WhySoBad:
I've now implemented the vectored reads/writes using an intermediate buffer whose size is equal to the combined size of all provided buffers. This way, we should get the same semantics as ordinary `writev`, `readv`, ... on real systems.

src/shims/unix/fd.rs
```diff
@@ -464,4 +464,210 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        // Allocate a buffer which has the combined size of all buffers provided in `iov`.
+        let read_buffer_ptr = this.allocate_ptr(
+            Size::from_bytes(total_bytes),
+            Align::ONE,
+            MemoryKind::Stack,
+            AllocInit::Uninit,
+        )?;
+
+        // Read bytes into the combined buffer. This ensures the read is atomic.
+        this.read(fd, read_buffer_ptr.into(), total_bytes, offset, dest)?;
```
RalfJung: This can return before the read is done (if the read blocks). So this needs to be refactored to use the `finish` callback.

src/shims/unix/fd.rs
```diff
@@ -464,4 +464,210 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            None
+        };
+        if let Some(flags) = flags {
+            let flags = this.read_scalar(flags)?.to_i32()?;
+            // `preadv2` with zero flags is the same as `preadv`.
+            if flags != 0 {
+                // We know that only `preadv2` provides the flags argument.
+                throw_unsup_format!("preadv2: non-zero flags are unsupported");
+            }
+        }
```
RalfJung: It doesn't seem worth supporting `preadv2` for now if std and tokio don't need it.

src/shims/unix/fd.rs
```diff
@@ -464,4 +464,210 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        // Check that the FD exists.
+        let Some(_fd) = this.machine.fds.get(fd) else {
+            return this.set_last_error_and_return(LibcError("EBADF"), dest);
+        };
+
+        let iovec_layout = this.libc_array_ty_layout("iovec", iovcnt);
+        let iov_ptr_mplace = this.ptr_to_mplace(iov_ptr, iovec_layout);
+
+        let mut buffers = Vec::new();
```
RalfJung, line 445:
```suggestion
        // Read list of buffers from `iov`.
        let mut buffers = Vec::new();
```

src/shims/unix/fd.rs
```diff
@@ -464,4 +464,210 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                .read_scalar(&iov_len_field)?
+                .to_int(iov_len_field.layout.size)?
+                .try_into()
+                .unwrap();
+
+            let iov_base_field = this.project_field_named(&iovec, "iov_base")?;
+            let iov_base_ptr = this.read_pointer(&iov_base_field)?;
+
+            buffers.push((iov_base_ptr, iov_len));
+            total_bytes = total_bytes.checked_add(iov_len).unwrap();
```
RalfJung: Looks like you can just compute `total_bytes` afterwards via `iter()...sum()`

src/shims/unix/fd.rs
```diff
@@ -464,4 +464,210 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        }
+
+        // Allocate a buffer which has the combined size of all buffers provided in `iov`.
+        let read_buffer_ptr = this.allocate_ptr(
```
RalfJung: I'd suggest calling this just `tmp_ptr` or so.

Or rename both of them, e.g. `user_buffer` and `tmp_buffer` or so.

src/shims/unix/fd.rs
```diff
@@ -464,4 +464,210 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // previous buffers.
+            let read_buffer_start_ptr = this.ptr_offset_inbounds(
+                read_buffer_ptr.into(),
+                i64::try_from(bytes_copied).unwrap(),
+            )?;
+
+            // Copy at most as many bytes as the buffer fits but without reading
+            // any uninitialized bytes from the combined buffer.
+            let copy_amount = buffer_len
+                .min(bytes_read.saturating_sub_signed(i64::try_from(bytes_copied).unwrap()));
```
RalfJung: Seems like this will become simpler if you count down `bytes_remaining` rather than counting up `bytes_read`

src/shims/unix/fd.rs
```diff
@@ -464,4 +464,210 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                true,
+            )?;
+
+            bytes_copied = bytes_copied.checked_add(buffer_len).unwrap();
+        }
+
+        // Write bytes from the combined buffer. This ensures the write is atomic.
+        this.write(fd, write_buffer_ptr.into(), total_bytes, offset, dest)?;
+
+        this.deallocate_ptr(write_buffer_ptr.into(), None, MemoryKind::Stack)
```
RalfJung: This also has to be moved into the continuation.

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -58,6 +59,16 @@ fn main() {
     test_statx_on_file_descriptor();
     #[cfg(target_os = "linux")]
     test_statx_empty_path_on_pipe();
+    test_readv();
+    test_readv_empty_bufs();
+    #[cfg(not(target_os = "solaris"))]
+    test_preadv();
```
RalfJung, line 67: Looks like we should have a check in the shim then to avoid running it on Solaris.

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -846,3 +857,254 @@ pub fn check_stat_fields(stat: &libc::stat) {
     let _st_mtime_nsec = stat.st_mtime_nsec;
     let _st_ctime_nsec = stat.st_ctime_nsec;
 }
+
+/// Test vectored reads with multiple buffers.
+fn test_readv() {
```
RalfJung, line 864: Would be good to also have a test with sockets that
- ensures we actually fill more than one buffer
- ensures things work correctly when the read/write blocks

tests/pass/shims/fs.rs
```diff
@@ -427,3 +433,77 @@ fn test_flock() {
...
+
+    // Reset file cursor to read the written bytes.
+    f.seek(SeekFrom::Start(bytes_read as u64)).unwrap();
+    let mut written_bytes = vec![0u8; bytes_written];
+    f.read_exact(&mut written_bytes).unwrap();
+    assert_eq!(written_bytes.as_slice(), &write_buffer[0..bytes_written]);
+}
+
+/// Test vectored reads and vectored writes with byte offsets.
+#[cfg(all(unix, not(any(target_os = "solaris", target_os = "android"))))]
```
RalfJung, line 474: Please explain in the comment why this is disabled on Solaris and Android.

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -846,3 +857,254 @@ pub fn check_stat_fields(stat: &libc::stat) {
...
+fn test_readv_empty_bufs() {
+    let path = utils::prepare_with_content("pass-libc-readv-empty-bufs.txt", &[1u8, 2, 3]);
+    let cpath = CString::new(path.into_os_string().into_encoded_bytes()).unwrap();
+    let fd = unsafe { libc::open(cpath.as_ptr(), libc::O_RDONLY) };
+    assert_ne!(fd, -1);
+    unsafe { assert_eq!(libc::readv(fd, ptr::null::<libc::iovec>(), 0), 0) };
+}
+
+/// Test vectored reads with multiple buffers and a byte offset.
+#[cfg(not(target_os = "solaris"))]
```
RalfJung, line 908: A comment should explain why this is disabled on Solaris.

RalfJung (commented):
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5019/48b0b923635281fad0c7597c7c704a7d7fe10cf2..25b82495e0d535f84b74baf2ec737881a86c28b2)*

src/shims/unix/fd.rs
RalfJung: Please also move the short read/write logic into the helpers, otherwise vectored reads are never short.

src/shims/unix/fd.rs
```diff
@@ -447,19 +441,305 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        )
+    }
+}
+
+impl<'tcx> EvalContextPrivExt<'tcx> for crate::MiriInterpCx<'tcx> {}
+trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    /// Read `len` bytes from the `fd` file description at `offset` into the buffer
+    /// pointed to by `ptr`.
+    /// If `offset` is [`None`] the [`FileDescription::read`] method is used, otherwise
+    /// the [`UnixFileDescription::pread`] method is used.
```
RalfJung:
```suggestion
    /// the [`UnixFileDescription::pread`] method is used.
    /// `finish` will be invoked when the read is done (which might be way after
    /// this function returns as the read may block).
```
and same for the write helper.

src/shims/unix/fd.rs
```diff
@@ -447,19 +441,305 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    fn read_at_offset(
```
RalfJung: The offset is optional so this name doesn't seem great.

Maybe `read_from_fd` or so?

src/shims/unix/fd.rs
RalfJung:
```suggestion
    /// If `offset` is [`Some`], the read occurs at the given absolute position rather than the current file position (`read_at` semantics rather than `read`).
```

RalfJung (commented):
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5019/48b0b923635281fad0c7597c7c704a7d7fe10cf2..07dcaefb91bf313ac5acfee23286764519846c69)*

src/shims/unix/fd.rs
```diff
@@ -331,25 +331,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             this.write_null(dest)?;
             return interp_ok(());
         }
```
RalfJung: What about this, shouldn't that also be in the shared helpers?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5019/48b0b923635281fad0c7597c7c704a7d7fe10cf2..e690c7e1b48593c11b907f176641cb7dbd749640)*

RalfJung:

@rustbot author

WhySoBad:
After this is merged, the `pread` and `pwrite` checkboxes can be checked in #3179.

PR #5023, opened 2026-05-10
Handle short reads and writes for TCP sockets

WhySoBad:
Hi,

As mentioned in rust-lang/miri#5005, this pull request adds handling for short reads and writes on TCP sockets. 
Because short reads/writes imply an empty/full buffer on Unix hosts, we need to remove the readiness in those cases. On Windows hosts, we need to forcefully emit a new readiness event to prevent applications relying on this behavior from being stuck forever.

The `tcp_echo` test would not finish with the current implementation, thus I adjusted this test case and added it here, to ensure that it finishes with the new handling.

### DIFF
--- a/src/shims/unix/socket.rs
+++ b/src/shims/unix/socket.rs
@@ -1470,8 +1470,6 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
         // This is a *non-blocking* write.
         let result = this.write_to_host(stream, length, buffer_ptr)?;
-        // FIXME: When the host does a short write, we should emit an epoll edge -- at least for targets for which tokio assumes no short writes:
-        // <https://github.com/tokio-rs/tokio/blob/6c03e03898d71eca976ee1ad8481cf112ae722ba/tokio/src/io/poll_evented.rs#L240>
         match result {
             Err(IoError::HostError(e))
                 if matches!(e.kind(), io::ErrorKind::NotConnected | io::ErrorKind::WouldBlock) =>
@@ -1483,6 +1481,45 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // On Windows hosts, `send` can return WSAENOTCONN where EAGAIN or EWOULDBLOCK
                 // would be returned on UNIX-like systems. We thus remap this error to an EWOULDBLOCK.
                 interp_ok(Err(IoError::HostError(io::ErrorKind::WouldBlock.into())))
+            }
+            Ok(bytes_written) if bytes_written < length => {
+                // We had a short write. On Unix hosts using the `epoll` and `kqueue` backends, a
+                // short write means that the write buffer is full. We update the readiness
+                // accordingly, which means that next time we see "writable" we will report an epoll
+                // edge. Some applications (e.g. tokio) rely on this behavior; see
+                // <https://github.com/tokio-rs/tokio/blob/HEAD/tokio/src/io/poll_evented.rs#L244-L264>.
+                if cfg!(any(
+                    // epoll
+                    target_os = "android",
+                    target_os = "illumos",
+                    target_os = "linux",
+                    target_os = "redox",
+                    // kqueue
+                    target_os = "dragonfly",
+                    target_os = "freebsd",
+                    target_os = "ios",
+                    target_os = "macos",
+                    target_os = "netbsd",
+                    target_os = "openbsd",
+                    target_os = "tvos",
+                    target_os = "visionos",
+                    target_os = "watchos",
+                )) {
+                    socket.io_readiness.borrow_mut().writable = false;
+                    this.update_epoll_active_events(socket.clone(), /* force_edge */ false)?;
+                } else {
+                    // On hosts which don't use the `epoll` or `kqueue` backends, a short write
+                    // doesn't imply a full write buffer. However, the target we are emulating might
+                    // guarantee this behavior. To prevent applications from being stuck on such
+                    // targets waiting on a new readiness event, we emit a new edge which still
+                    // contains a writable readiness. This should trick the applications into trying
+                    // another write which would then return EWOULDBLOCK should it really be full.
+                    // This results in an unrealistic execution but we don't have another way of
+                    // finding out whether the write buffer is full. The "default case" of linux
+                    // host and linux target isn't affected by this.
+                    this.update_epoll_active_events(socket.clone(), /* force_edge */ true)?;
+                }
+                interp_ok(result)
             }
             result => interp_ok(result),
         }
@@ -1556,8 +1593,6 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             length,
             buffer_ptr,
         )?;
-        // FIXME: When the host does a short read, we should emit an epoll edge -- at least for targets for which tokio assumes no short reads:
-        // <https://github.com/tokio-rs/tokio/blob/6c03e03898d71eca976ee1ad8481cf112ae722ba/tokio/src/io/poll_evented.rs#L182>
         match result {
             Err(IoError::HostError(e))
                 if matches!(e.kind(), io::ErrorKind::NotConnected | io::ErrorKind::WouldBlock) =>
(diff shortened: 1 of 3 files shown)


src/shims/unix/socket.rs
```diff
@@ -1464,6 +1473,27 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // would be returned on UNIX-like systems. We thus remap this error to an EWOULDBLOCK.
                 interp_ok(Err(IoError::HostError(io::ErrorKind::WouldBlock.into())))
             }
+            Ok(bytes_written) if bytes_written < length => {
+                // We had a short write.
+                if cfg!(not(windows)) {
+                    // On Unix hosts a short write means that the write buffer is full.
+                    // Because some applications (e.g. tokio) rely on this behavior, we thus
+                    // remove the writable readiness.
```
```suggestion
                    // remove the writable readiness. This means the next time
                    // we see "writable" we will report an epoll edge.
```

src/shims/unix/socket.rs
```diff
@@ -1464,6 +1473,27 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // would be returned on UNIX-like systems. We thus remap this error to an EWOULDBLOCK.
                 interp_ok(Err(IoError::HostError(io::ErrorKind::WouldBlock.into())))
             }
+            Ok(bytes_written) if bytes_written < length => {
+                // We had a short write.
+                if cfg!(not(windows)) {
```
RalfJung: Tokio doesn't do this for all non-Windows targets, right? Please use the same condition as tokio and add a link to it. We rely on tokio as having tested this guarantee for us.

src/shims/unix/socket.rs
```diff
@@ -1550,6 +1580,30 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // would be returned on UNIX-like systems. We thus remap this error to an EWOULDBLOCK.
                 interp_ok(Err(IoError::HostError(io::ErrorKind::WouldBlock.into())))
             }
+            Ok(bytes_read) if bytes_read < length && !socket.is_read_shut_down.get() => {
+                // We had a short read and the socket read end is not shut down.
+                // When the read end is shut down, we only get EOF (zero byte reads)
+                // and thus we don't want to remove the readable readiness because
+                // we would then block indefinitely on subsequent reads.
```
```suggestion
            Ok(bytes_read) if bytes_read < length && bytes_read > 0 => {
```
0 bytes read is not a short read, that's a guaranteed EOF. So I think we can just wholly exclude that case here.

Though, what does it mean for readiness when we encounter EOF?

tests/pass-dep/tokio/socket.rs
```diff
@@ -54,3 +55,42 @@ async fn test_read_write() {
 
     stream.write_all(TEST_BYTES).await.unwrap();
 }
+
+// Test that [`io::copy`] works for copying every socket input directly into it's
+// output. By this, we test whether Miri correctly removes read readiness when
+// the buffer is empty.
+// This test case is heavily inspired by the `tcp_echo` test case of the tokio
+// test suite.
+async fn test_echo_server() {
```
RalfJung, line 64: Is there a way to have a simpler test that directly uses libc APIs? That would be much easier to debug if this ever breaks in the future.

RalfJung (commented):
Here we go. :)

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5023/f3f1ed1b34320de40d6dfad8f937341be227876d..ff85bd94bc040af79ae5d6b6d0c721803f5c5332)*

src/shims/unix/socket.rs
WhySoBad: > Though, what does it mean for readiness when we encounter EOF?

I think this might be the reason why the kernel sets the readable readiness when shutting down both ends of a socket. However, with this logic it should also set the readable readiness after a shut down of the read end then (which it doesn't).
I've also found this list where someone experimentally collected which Unix sets which readiness when a fd is EOF: https://www.chiark.greenend.org.uk/greenend.org.uk/rjk/tech/poll.html
And by this, we should keep the readable readiness when a socket is EOF on Linux.

RalfJung: Yeah arguably the socket is readable, in the sense that a read will not block nor return EAGAIN.

tests/pass-dep/tokio/socket.rs
WhySoBad: Because we want to test whether Miri correctly removes the readiness (and by this emits a new edge on the next readiness event), using libc directly is very tedious. 
I've now simplified the tokio test such that it no longer requires `io::copy` and I've also added some more comments to better understand what's going on.

RalfJung: Hm, what makes it so tedious? Is it not just something like
- establish connection, setup epoll
- write 100 bytes
- read 200 bytes, get back a short read
- have the tokio logic: on linux and friends, do *not* read again, else read again to get the EAGAIN
- write 100 bytes
- check epoll: must be ready for reads now

tests/pass-dep/tokio/socket.rs
```diff
@@ -82,15 +82,29 @@ async fn test_echo_server() {
...
+    // Every read is thus a short read which removes the internal
+    // readable readiness.
+    let mut buffer = [0u8; 1024];
+    let mut bytes_copied = 0;
+    loop {
+        tokio::select! {
+            // The client finished writing into the input.
+            // We can thus no longer read any input and need to
+            // break.
+            Some(_) = rx.recv() => break,
```
RalfJung: This is confusing, can't it be the case that there are still bytes left to be read in the buffer?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5023/f3f1ed1b34320de40d6dfad8f937341be227876d..5c06c9585ab529f963393764a1a5d86f484aed81)*

WhySoBad: I've now switched to shutting down the write end of the peer once it's finished. This should cause EOF reads in the client and then we know we can exit the loop.

tests/pass-dep/tokio/socket.rs
WhySoBad: You're right, my approach would've been an overengineered thing to achieve the same. I've now added such a libc test as well.

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -421,3 +429,79 @@ fn test_shutdown_write() {
 
     server_thread.join().unwrap();
 }
+
+// Test that Miri correctly removes the read readiness or emits a new edge after a short read.
+fn test_readiness_after_short_read() {
```
RalfJung, line 426: Can you also add `test_readiness_after_short_write`?

(Note that I pushed to your branch and already squashed your commits before I realized that the write path is currently not tested.)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5023/2a297fe427d90eff0644c8e812db4102ddbd38fa..644a6b30ef8a1a4459f9f273deaa355265afcc4c)*

RalfJung, line 426: The test does not seem to work. I removed the short-read handling logic and the test still passes.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5023/2a297fe427d90eff0644c8e812db4102ddbd38fa..e6bd248e690f401496579aabf7c84b66735408e6)*

tests/pass-dep/tokio/socket.rs
RalfJung: Nice, thanks.

Feel free to revert the tokio test to its original form with the high-level APIs, it's nice to cover those as well.

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
WhySoBad: Yes, the problem is that we read until we get an EWOULDBLOCK and we already have the logic of removing the readiness when receiving an EWOULDBLOCK. I've now fixed it by already exiting the loop after the first short read.

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
RalfJung: This and the below check are wrong, it should be "windows_host", not "windows_hosts".

Seems like the test actually works fine on Windows...

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5023/2a297fe427d90eff0644c8e812db4102ddbd38fa..7492f695538031d1282027b1e17973c295f40fca)*

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -505,3 +514,77 @@ fn test_readiness_after_short_read() {
...
+    // Add client socket with writable interest to epoll.
+    epoll_ctl_add(epfd, client_sockfd, EPOLLET | EPOLLOUT).unwrap();
+
+    // Wait until the socket becomes writable.
+    check_epoll_wait::<8>(epfd, &[Ev { events: EPOLLOUT, data: client_sockfd }], -1);
+
+    // We now want to fill the write buffer of the socket by repeatedly writing
+    // `buffer` into it. The last write should then be a short write.
+    // We assume/hope that the write buffer length is not divisible by 11111.
+    let buffer = [123u8; 1111];
```
RalfJung: The comment says 11111 but the code says 1111?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5023/2a297fe427d90eff0644c8e812db4102ddbd38fa..7492f695538031d1282027b1e17973c295f40fca)*

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
WhySoBad: Yes, I also noticed that. I've now included the total bytes written in a panic to see how many bytes Windows allows us to write.

WhySoBad: Okay, it seems as if it's possible to fill the write buffer on Windows. I'll test this again for the other cases where we explicitly skipped Windows hosts and open another PR should it also no longer be a problem. 

However, I think we need to explicitly skip Windows and Apple hosts for this test. As it seems, Windows and Apple hosts don't end with a short write before filling up the buffer and instead return EWOULDBLOCK for which we already reset the writable readiness. It could also be that buffer size was divisible by 1111, but I think this is unlikely.

RalfJung: What happens if you do a single huge write to fill the buffer in one go? Do Windows and Apple really flip from "successfully wrote all bytes at once" to "EWOULDBLOCK, not writing anything" as buffer size increases?

WhySoBad: I've now tried repeatedly writing a 1 MB buffer. On MacOS it for some reason now succeeded. And on Windows
we achieved an ~600KB write buffer when writing chunks of 1111 bytes. Now with a write buffer of 1 MB we achieve a write buffer of 2 MB. However, in both cases it doesn't do a partial write.

RalfJung: Well that's kind of annoying.

I'd say then write the test in a way that it also works when it ends up getting an EAGAIN.
And make it write in chunks of some prime on the order of 1000, like 1039.

RalfJung:
Looks good, thanks. :) Please squash.
@rustbot author

PR #5030, opened 2026-05-12
Add _mm512_permutex2var_epi8 shim support

elichai:
This implements the _mm512_permutex2var_epi8 intrinsic, which is vpermi2var.qi.512, which permutes 8-bit elements from two input vectors according to an index vector, producing a destination vector.

This builds on the `permute2` function we've added to https://github.com/rust-lang/miri/pull/4963

I've tested it on a real machine with: `rustc ~/intrinsics-x86-avx512.rs -C target-feature=+avx512f,+avx512vl,+avx512bw,+avx512bitalg,+avx512vpopcntdq,+avx512vnni,+avx512vbmi -o ~/avx512-test && ~/avx512-test`


RalfJung:
Nice, thanks :)

Issue #4703, opened 2025-11-17
GenMC: support macOS


RalfJung:
GenMC should work fine on macOS, but there are multiple things that need to be done before we can enable it in Miri:
- (non-blocking) there's some [odd warning about filename cases](https://github.com/rust-lang/miri/pull/4701#issuecomment-3540470511) (also see [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/122651-general/topic/macOS.20case.20sensitity.20issues/with/556769933)).
- The test suite seems to get stuck on CI, I don't know which test it is stuck in (see [this CI log](https://github.com/rust-lang/miri/actions/runs/19423438947/job/55565370422?pr=4701)). This part will probably require actual macOS hardware to further analyze, which I do not have.
- When I run the test suite for an apple target locally, the `std` tests fail. This is somewhat expected because we didn't wire up the mutex shims on macOS, but what's surprising is that we get an "access to uninit memory" error. I think that's an instance of [this](https://github.com/rust-lang/miri/issues/4572#issuecomment-3541175255). We should probably redirect the pthread APIs to genmc when running in genmc mode.

y1lan:
In  MacOs, directory name is case-insensitive. Therefore, `runtime` and `Runtime` would be regarded as the same directory when `runtime` (lowercase one) has been created before.

y1lan:
I think differing `runtime` and `Runtime` would be a solution.

y1lan:
```
set(PKG_INCLUDE_DIR "${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/runtime")
```
https://github.com/MPI-SWS/genmc/blob/22d3d0b44dedb4e8e1aae3330e546465e4664529/CMakeLists.txt#L10

is the root cause.

y1lan:
it seems that ./miri script works no well with `--features=genmc`, is there any guidelines to run genmc tests?

RalfJung:
See https://github.com/rust-lang/miri/blob/master/doc/genmc.md. You'll have to be a bit more specific about what does not work if you want someone to help you.

RalfJung:
Current status: GenMC fails with assertion failures (https://github.com/rust-lang/miri/pull/5032).

PR #5032, opened 2026-05-12
Enable MacOS support for GenMC mode.

RalfJung:
Let's see what the current status of https://github.com/rust-lang/miri/issues/4703 is.


RalfJung:
Okay now that I fixed our CI we actually get assertion failures within GenMC on macOS. ;)
Cc @michaliskok 
```
  ********************************************************************************
  INTERNAL FAILURE: Assertion failed: nodePtr_
    at /Users/runner/work/miri/miri/genmc-sys/genmc-src/genmc/genmc/ADT/ilist.hpp:123:reference genmc::ilist_iterator_base<EventLabel, po_tag, true, true>::operator*() const [T = EventLabel, Tag = po_tag, IsConst = true, IsReverse = true]
  ********************************************************************************
```

PR #5036, opened 2026-05-13
Fix `connect` of blocking TCP sockets always returning `ENOTCONN` on error

WhySoBad:
This pull request fixes that failed blocking connects always return ENOTCONN instead of the actual error. 
The problem was that we also stored the result of `TcpStream::take_error` for blocking sockets instead of just returning this error.

### DIFF
--- a/src/shims/unix/socket.rs
+++ b/src/shims/unix/socket.rs
@@ -45,6 +45,13 @@ enum SocketState {
     /// For a socket created using the `connect` syscall, this is
     /// only reachable from the [`SocketState::Connecting`] state.
     Connected(TcpStream),
+    /// The SO_ERROR socket option has been set after calling
+    /// the `connect` syscall, indicating that the connection
+    /// attempt failed. By the POSIX specification, a socket is
+    /// is an unspecified state after a failed connection attempt
+    /// and thus nothing (except destroying the socket) should be
+    /// supported when a socket is in this state.
+    ConnectionFailed(TcpStream),
 }
 
 #[derive(Debug)]
@@ -77,7 +84,10 @@ impl FileDescription for Socket {
 
         if matches!(
             &*self.state.borrow(),
-            SocketState::Listening(_) | SocketState::Connecting(_) | SocketState::Connected(_)
+            SocketState::Listening(_)
+                | SocketState::Connecting(_)
+                | SocketState::Connected(_)
+                | SocketState::ConnectionFailed(_)
         ) {
             // There exists an associated host socket so we need to deregister it
             // from the blocking I/O manager.
@@ -373,6 +383,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         };
 
         assert!(this.machine.communicate(), "cannot have `Socket` with isolation enabled!");
+        this.ensure_not_failed(&socket, "bind")?;
 
         let mut state = socket.state.borrow_mut();
 
@@ -411,6 +422,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     "bind: socket is already bound and binding a socket \
                     multiple times is unsupported"
                 ),
+            SocketState::ConnectionFailed(_) => unreachable!(),
         }
 
         interp_ok(Scalar::from_i32(0))
@@ -434,6 +446,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         };
 
         assert!(this.machine.communicate(), "cannot have `Socket` with isolation enabled!");
+        this.ensure_not_failed(&socket, "listen")?;
 
         let mut state = socket.state.borrow_mut();
 
@@ -460,6 +473,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             SocketState::Connecting(_) | SocketState::Connected(_) => {
                 throw_unsup_format!("listen: listening on a connected socket is unsupported")
             }
+            SocketState::ConnectionFailed(_) => unreachable!(),
         }
 
         interp_ok(Scalar::from_i32(0))
@@ -495,6 +509,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         };
 
         assert!(this.machine.communicate(), "cannot have `Socket` with isolation enabled!");
+        this.ensure_not_failed(&socket, "accept4")?;
 
         if !matches!(*socket.state.borrow(), SocketState::Listening(_)) {
             throw_unsup_format!(
@@ -589,6 +604,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         };
 
         assert!(this.machine.communicate(), "cannot have `Socket` with isolation enabled!");
+        this.ensure_not_failed(&socket, "connect")?;
 
         match &*socket.state.borrow() {
             SocketState::Initial => { /* fall-through to below */ }
@@ -632,15 +648,20 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             let dest = dest.clone();
 
             this.ensure_connected(
-                socket,
+                socket.clone(),
                 /* should_wait */ true,
                 "connect",
                 callback!(
                     @capture<'tcx> {
+                        socket: FileDescriptionRef<Socket>,
                         dest: MPlaceTy<'tcx>
                     } |this, result: Result<(), ()>| {
                         if result.is_err() {
-                            this.set_last_error_and_return(LibcError("ENOTCONN"), &dest)
+                            // An error occurred whilst connecting. We know
+                            // that it has been consumed by `ensure_connected`
+                            // and is now stored in `socket.error`.
+                            let err = socket.error.take().unwrap();
+                            this.set_last_error_and_return(err, &dest)
                         } else {
                             this.write_scalar(Scalar::from_i32(0), &dest)
                         }
@@ -964,6 +985,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     SocketState::Listening(listener) => listener.set_ttl(ttl),
                     SocketState::Connecting(stream) | SocketState::Connected(stream) =>
                         stream.set_ttl(ttl),
+                    SocketState::ConnectionFailed(_) => unreachable!(),
                 };
 
                 return match result {
@@ -994,6 +1016,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                         ),
                     SocketState::Connecting(stream) | SocketState::Connected(stream) =>
                         stream.set_nodelay(nodelay),
+                    SocketState::ConnectionFailed(_) => unreachable!(),
                 };
 
                 return match result {
@@ -1064,17 +1087,16 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             let opt_so_error = this.eval_libc_i32("SO_ERROR");
 
             if option_name == opt_so_error {
-                // Because `TcpStream::take_error()` and `TcpListener::take_error()` consume the latest async
-                // error, we know that our stored `socket.error` is outdated when `TcpStream::take_error()`/
-                // `TcpListener::take_error()` returns `Ok(Some(...))`.
-                // If they return `Ok(None)`, then we fall back to the stored `socket.error`.
-                let error = match &*socket.state.borrow() {
-                    SocketState::Initial | SocketState::Bound(_) => socket.error.take(),
-                    SocketState::Listening(listener) =>
-                        listener.take_error().unwrap_or(socket.error.take()),
-                    SocketState::Connecting(stream) | SocketState::Connected(stream) =>
-                        stream.take_error().unwrap_or(socket.error.take()),
+                // Reading SO_ERROR should always return the latest async error. Because our stored
+                // `socket.error` could be outdated, we attempt to update it here.
+                this.update_last_error(&socket);
+
+                let return_value = match socket.error.take() {
+                    Some(err) => this.io_error_to_errnum(err)?.to_i32()?,
+                    // If there is no error, we return 0 as the option value.
+                    None => 0,
                 };
+
                 // Clear our own stored error -- it was either `take`n above or it is outdated.
                 socket.error.replace(None);
 
@@ -1082,12 +1104,6 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // I/O and epoll readiness of the socket.
                 socket.io_readiness.borrow_mut().error = false;
                 this.update_epoll_active_events(socket, /* force_edge */ false)?;
-
-                let return_value = match error {
-                    Some(err) => this.io_error_to_errnum(err)?.to_i32()?,
-                    // If there is no error, we write 0 into the option value buffer.
-                    None => 0,
-                };
 
                 // Allocate new buffer on the stack with the `i32` layout.
                 let value_buffer = this.allocate(this.machine.layouts.i32, MemoryKind::Stack)?;
@@ -1111,6 +1127,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     SocketState::Listening(listener) => listener.ttl(),
                     SocketState::Connecting(stream) | SocketState::Connected(stream) =>
                         stream.ttl(),
+                    SocketState::ConnectionFailed(_) => unreachable!(),
                 };
 
                 let ttl = match ttl {
@@ -1139,6 +1156,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                         ),
                     SocketState::Connecting(stream) | SocketState::Connected(stream) =>
                         stream.nodelay(),
+                    SocketState::ConnectionFailed(_) => unreachable!(),
                 };
 
                 let nodelay = match nodelay {
@@ -1211,6 +1229,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         };
 
         assert!(this.machine.communicate(), "cannot have `Socket` with isolation enabled!");
+        this.ensure_not_failed(&socket, "getsockname")?;
 
         let state = socket.state.borrow();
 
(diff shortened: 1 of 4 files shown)


src/shims/unix/socket.rs
```diff
@@ -1824,26 +1731,36 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+                        let io_error = if socket.is_non_block.get() {
+                            // For non-blocking sockets it's the program's responsibility to
+                            // read SO_ERROR itself.
+
+                            // Store the error such that we can return it when
+                            // `getsockopt(SOL_SOCKET, SO_ERROR, ...)` is called on the socket.
+                            socket.error.replace(Some(err));
+
+                            LibcError("ENOTCONN")
```
RalfJung: This looks strange -- we are *both* returning ENOTCONN *and* storing this for SO_ERROR? Isn't every error only returned once?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5036/c4a2b5e4ecf4cf825c5150b2284d5cb97768bfb4..0ead3622545f414669c80d2a7e8b42c7340ab1e2)*

WhySoBad: Yes, e.g. when we have a failed `connect` and we then want to `read` we would end up in this case. 
The `read` should return ENOTCONN as the `connect` failed and the reason for the failure is stored in the SO_ERROR socket option.

Also, I just noticed that our logic here is a bit flawed. There could be the case where we have a non-blocking connect which failed and we don't have an operation following it which calls `ensure_connected` (meaning no `read`/`write`) but instead we read SO_ERROR directly. When we then call `read` or `write` afterwards, the error check in `ensure_connected` succeeds and we would internally treat the socket as connected. 
I think we can solve this by storing an additional boolean tracking whether the current error has already been read but I'm not entirely sure whether this works out -- I'll have to think a bit more about this.

src/shims/unix/socket.rs
RalfJung: This comment seems outdated now, it's not always ENOTCONN.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5036/c4a2b5e4ecf4cf825c5150b2284d5cb97768bfb4..0ead3622545f414669c80d2a7e8b42c7340ab1e2)*

src/shims/unix/socket.rs
RalfJung: So the idea is that for a non-blocking socket, this is called before some *other* operation. That other operation should then return ENOTCONN.

But for blocking sockets, we call it during `connect` itself. So we want a different error treatment.

The real distinguishing factor hence should be "is this called from `connect` or some other operation", right? Maybe that should just be a flag. Remember that `is_non_block` can be changed any time by the program.

src/shims/unix/socket.rs
```diff
@@ -1301,9 +1208,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     address_ptr: Pointer,
                     address_len_ptr: Pointer,
                     dest: MPlaceTy<'tcx>,
-                } |this, result: Result<(), ()>| {
-                    if result.is_err() {
-                        return this.set_last_error_and_return(LibcError("ENOTCONN"), &dest)
+                } |this, result: Result<(), IoError>| {
+                    if let Err(e) = result {
+                        return this.set_last_error_and_return(e, &dest)
```
RalfJung: For instance, here we don't ever want to return anything other than ENOTCONN. Right?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5036/c4a2b5e4ecf4cf825c5150b2284d5cb97768bfb4..0ead3622545f414669c80d2a7e8b42c7340ab1e2)*

WhySoBad:
I've noticed that with the current implementation we cannot call `connect` again after a previous connection attempt failed. This is because we didn't deregister the source from the blocking I/O manager. 
I've now changed that (which also needed a slight change in the blocking I/O manager) and updated a test such that we test connecting multiple times. 

@rustbot ready

src/concurrency/blocking_io.rs
```diff
@@ -278,22 +278,17 @@ impl BlockingIoManager {
...
         let poll =
             self.poll.as_ref().expect("Blocking I/O should not be called with isolation enabled");
 
-        let stored_source = self.sources.remove(&source_id).expect("Source should be registered");
-        // Ensure that the source file description is already removed from the file
-        // description table.
-        assert!(
-            stored_source.fd.upgrade().is_none(),
-            "Sources must only be deregistered when they are destroyed"
-        );
```
RalfJung, line 293: Seems worth adding a comment explaining why the source might still exist.

src/shims/unix/socket.rs
```diff
@@ -1065,17 +1007,39 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             let opt_so_error = this.eval_libc_i32("SO_ERROR");
 
             if option_name == opt_so_error {
+                let state = socket.state.borrow();
+                let new_error = match &*state {
+                    SocketState::Initial | SocketState::Bound(_) => None,
+                    SocketState::Listening(listener) => listener.take_error().unwrap_or_default(),
+                    SocketState::Connecting(stream) | SocketState::Connected(stream) =>
+                        stream.take_error().unwrap_or_default(),
+                };
```
RalfJung: So if there's an error getting the error, we just ignore that?

src/shims/unix/socket.rs
```diff
@@ -1065,17 +1007,39 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                    SocketState::Listening(listener) =>
-                        listener.take_error().unwrap_or(socket.error.take()),
-                    SocketState::Connecting(stream) | SocketState::Connected(stream) =>
-                        stream.take_error().unwrap_or(socket.error.take()),
+                // `TcpListener::take_error()` returns `Ok(Some(...))`. When this is the case, `new_error` is
+                // `Some(...)`. Otherwise `new_error` is `None` and we fall back to the stored `socket.error`.
+                let return_value = match new_error.or(socket.error.take()) {
+                    Some(err) => this.io_error_to_errnum(err)?.to_i32()?,
+                    // If there is no error, we write 0 into the option value buffer.
+                    None => 0,
```
```suggestion
                let error = new_error.or(socket.error.take());
```
I think it's clearer to keep the error-to-retval logic separate the way it was before this PR.

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -154,6 +154,7 @@ fn test_connect_nonblock_err() {
         -1,
     );
 
+    // Because the socket is in `Connecting` state, this should reset it's state to `Initial`.
```
```suggestion
    // Because the socket is in `Connecting` state, this should reset its state to `Initial`.
```

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -163,6 +164,12 @@ fn test_connect_nonblock_err() {
     // Ensure that error readiness is cleared after reading SO_ERROR.
     let readiness = current_epoll_readiness::<8>(client_sockfd, EPOLLET | EPOLLOUT | EPOLLERR);
     assert!(readiness & EPOLLERR == 0);
+
+    // After a failed connection attempt we should be able to call `connect`
+    // again because the internal socket should be reset to `Initial` state.
+    let err = net::connect_ipv4(client_sockfd, addr).unwrap_err();
+    // Ensure we don't get EALREADY.
+    assert_eq!(err.kind(), ErrorKind::InProgress);
```
RalfJung: Did you confirm this on a real host?

src/shims/unix/socket.rs
```diff
@@ -1065,17 +1007,39 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             if option_name == opt_so_error {
+                let state = socket.state.borrow();
+                let new_error = match &*state {
+                    SocketState::Initial | SocketState::Bound(_) => None,
+                    SocketState::Listening(listener) => listener.take_error().unwrap_or_default(),
+                    SocketState::Connecting(stream) | SocketState::Connected(stream) =>
+                        stream.take_error().unwrap_or_default(),
+                };
+
+                if new_error.is_some() && matches!(&*state, SocketState::Connecting(_)) {
```
```suggestion
                // If we are still connecting, we may have to advance the state machine.
                // This is usually handled by `ensure_connected`, but here we
                // are clobbering the host socket's "last error" and we have to ensure this does not get lost.
                if new_error.is_some() && matches!(&*state, SocketState::Connecting(_)) {
```

src/shims/unix/socket.rs
```diff
@@ -1830,19 +1759,20 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                        // Go back to initial state since the only way of getting into the
-                        // `Connecting` state is from the `Initial` state and at this point
+                        // Deregister the socket from the blocking I/O manager and go back
+                        // to initial state since the only way of getting into the
+                        // `Connecting` state is from the `Initial` state, and at this point
                         // we know that the connection won't be established anymore.
-                        *state = SocketState::Initial;
-                        drop(state);
+                        this.machine.blocking_io.deregister(socket.id(), &*socket);
+                        socket.state.replace(SocketState::Initial);
```
RalfJung: This is basically a duplicate of what happens above in the SO_ERROR logic, isn't it?

Would it make sense to have a `update_last_error` method that
- calls `take_error`
- if there is an error there
  - overwrites `socket.error`
  - and advanced the state machine

That gives us a single place where we process incoming error information from the host socket. Then `ensure_connected` and SO_ERROR can both call this function.

RalfJung (commented):
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5036/c4a2b5e4ecf4cf825c5150b2284d5cb97768bfb4..22a05f3eddb44ac4584d5e1d5240310839687c9b)*

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
WhySoBad: No I did not. I've now tested it and after some research I've found that it's _sometimes_ possible to connect again to a socket whose connection failed. It depends on the kernel internal TCP state machine and the reason of the failure.
I'd therefore now 
- Revert the changes of the `blocking_io` manager
- Introduce a new `Failed(TcpStream)` state which we enter when we have an error whilst being in `Connecting(_)` state
- Return an unsupported error when attempting to `connect` to a socket in `Failed(_)` state

By this we also keep the semantics that "a single FileDescription is backed by at most one host file descriptor". 
What do you think about this?

RalfJung: >  I've found that it's sometimes possible to connect again to a socket whose connection failed. It depends on the kernel internal TCP state machine and the reason of the failure.

Have you figured out anything systematic /  any reliable rules? Which error codes are you seeing?

> Return an unsupported error when attempting to connect to a socket in Failed(_) state

Is this also needed for blocking sockets?

For non-blocking sockets there is the EALREADY error code, maybe we can slightly stretch its definition by saying that an attempt that fails is never considered "completed".

WhySoBad: Okay, I've now inspected the linux kernel's behavior a bit further: 
- After a non-blocking connection attempt failed SO_ERROR is set
- When calling `connect` again we end up here: https://github.com/torvalds/linux/blob/HEAD/net/ipv4/af_inet.c#L739-L745
where either ECONNABORTED or the SO_ERROR is returned, and the socket is reset to SS_UNCONNECTED.
- Calling `connect` a third time will then work again

I don't think we should implement that. I've also gave the `connect` POSIX man page a closer look and there I've found the following section:
```
APPLICATION USAGE
     If connect() fails, the state of the socket is unspecified. Conforming applications should close the file descriptor and create a new socket before attempting to reconnect.
```
and even in the Linux man page we have the following section:
```
NOTES
     If  connect()  fails,  consider  the  state of the socket as unspecified.  Portable applications should close the
     socket and create a new one for reconnecting.
```

So I'd now do the following:
- Revert the changes in the blocking I/O manager
- Introduce a new `Failed` state which we enter when we have an error whilst being in `Connecting(_)` state
- Return an UB error when any operation is done on a socket which is in the `Failed` state

RalfJung: > Return an UB error when any operation is done on a socket which is in the Failed state

UB seems too strong, let's make that "unsupported".

But other than that, sounds good. Thanks for checking!

src/shims/unix/socket.rs
```diff
@@ -1044,6 +1075,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             return this.set_last_error_and_return_i32(LibcError("ENOTSOCK"));
         };
 
+        this.ensure_not_failed(&socket, "getsockopt")?;
```
RalfJung: `getsockopt`/`setsockopt` should be fine on a failed socket I think? In particular SO_ERROR really should work?

src/shims/unix/socket.rs
```diff
@@ -1064,17 +1097,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                        stream.take_error().unwrap_or(socket.error.take()),
+                // Reading SO_ERROR should always return the latest async error. Because our stored
+                // `socket.error` could be outdated, we attempt to update it here.
+                if let Err(e) = this.update_last_error(&socket) {
+                    return this.set_last_error_and_return_i32(e);
                 };
+
+                let return_value = match socket.error.take() {
+                    Some(err) => this.io_error_to_errnum(err)?.to_i32()?,
+                    // If there is no error, we write 0 into the option value buffer.
```
```suggestion
                    // If there is no error, we return 0 as the the option value.
```

src/shims/unix/socket.rs
```diff
@@ -1871,6 +1903,66 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    /// Should `socket` be in the [`SocketState::Connecting`] state whilst there is
+    /// an error on the host socket, we transition into the [`SocketState::ConnectionFailed`]
+    /// state because we know that `socket` can no longer successfully establish a
+    /// connection.
+    fn update_last_error(&self, socket: &FileDescriptionRef<Socket>) -> Result<(), IoError> {
+        let mut state = socket.state.borrow_mut();
+
+        let new_error = match &*state {
+            SocketState::Listening(listener) =>
+                listener.take_error().map_err(IoError::HostError)?,
```
RalfJung: Can't you just `unwrap` here? Looking at the `getsockopt` docs, I don't see any reasonable way this can fail.

tests/fail-dep/libc/socket-connect-after-failed-connection.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    let errno =
+        net::getsockopt::<libc::c_int>(client_sockfd, libc::SOL_SOCKET, libc::SO_ERROR).unwrap();
+    // Depending on the host we receive different error kinds. Thus, we only check
+    // that it's a nonzero error code.
+    assert!(errno != 0);
+
+    // Ensure that error readiness is cleared after reading SO_ERROR.
+    let readiness = current_epoll_readiness::<8>(client_sockfd, EPOLLET | EPOLLOUT | EPOLLERR);
+    assert!(readiness & EPOLLERR == 0);
```
RalfJung, line 58: Why should it be cleared? Is that what the native host does?

RalfJung (commented):
Yes I think we are getting there. :)
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5036/5c23cbb102855122b70132051e8aece91799949c..31e980733c4fd7c78203c2ebd9d4fba7d9cb34cc)*

src/shims/unix/socket.rs
WhySoBad: Whilst not being part of the Linux man page, the POSIX man page specifies that `getsockopt` may fail if the socket has been shut down:
```
The getsockopt() function may fail if:

     EACCES
            The calling process does not have appropriate privileges.

     EINVAL
            The socket has been shut down.

     ENOBUFS
            Insufficient resources are available in the system to complete the function.
```
Since we don't check for that explicitly, I'd leave this here.

tests/fail-dep/libc/socket-connect-after-failed-connection.rs
WhySoBad: Yes, after reading SO_ERROR the error readiness is cleared. This also runs on my native host.

src/shims/unix/socket.rs
WhySoBad: > `getsockopt`/`setsockopt` should be fine on a failed socket I think?

If you want we can support that. However, the specification is very clear about the socket being in an unspecified state so I'd argue that even this should be unsupported.

>  In particular SO_ERROR really should work

Reading SO_ERROR works as we only enter the `ConnectionFailed` state _after_ reading and returning the error. Otherwise the `socket-connect-after-failed-connection.rs` test would already fail at the `getsockopt` instead of the `connect`.

src/shims/unix/socket.rs
RalfJung: But you are unwrapping this result in one of the call sites so that does not seem consistent.
Also, leaving the state machine in `Connecting` when SO_ERROR *fails* doesn't seem right.

src/shims/unix/socket.rs
```diff
@@ -1871,6 +1903,66 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            SocketState::Connecting(stream) | SocketState::Connected(stream) =>
+                stream.take_error().map_err(IoError::HostError)?,
+            _ => None,
```
RalfJung: Please spell out the states here so that we have to think about this case when adding more states.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5036/5c23cbb102855122b70132051e8aece91799949c..876982d61ea0930dd71594e50b3b96fd22db79ed)*

src/shims/unix/socket.rs
RalfJung: It doesn't really say what "unspecified state" means though. It also says "close and create new socket before attempting to reconnect", but doesn't talk about other operations like getsockopt/setsockopt.

It seems unlikely that basic option management would be affected by whatever the OS does on a failed connect.

src/shims/unix/socket.rs
```diff
@@ -787,6 +812,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             return this.set_last_error_and_return(LibcError("ENOTSOCK"), dest);
         };
 
+        this.ensure_not_failed(&socket, "recv")?;
```
RalfJung: Instead of having this check everywhere, we could just have it in `ensure_connected`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5036/5c23cbb102855122b70132051e8aece91799949c..876982d61ea0930dd71594e50b3b96fd22db79ed)*

src/shims/unix/socket.rs
WhySoBad: I've checked the Linux, NetBSD, FreeBSD and MacOS man pages and neither of them contain this error for `getsockopt`. Thus, I've now changed it to panic when there is an error reading SO_ERROR.

RalfJung:

@rustbot author

PR #5037, opened 2026-05-13
Fix last remnants of TB's Active -> Unique renaming

JoJoDeveloping:
Seems like we forgot this in rust-lang/miri#4595


RalfJung:
Good catch!

PR #5040, opened 2026-05-14
ci: declare contents:read on Tier 2 sysroots workflow

arpitjain099:
Pins `sysroots.yml` to `contents: read` at workflow scope. The `sysroots` job builds rust targets and uploads `failures.tar.gz` as a workflow artifact. The `sysroots-cron-fail-notify` job downloads that artifact and posts to Zulip via `ZULIP_API_TOKEN`. Neither writes to the repo or calls the GitHub API beyond `actions/upload-artifact` / `actions/download-artifact` (no explicit perms needed in v4+).

Defense-in-depth motivation is [CVE-2025-30066](https://github.com/advisories/GHSA-mrrh-fwg8-r2c3) on `tj-actions/changed-files`: a compromised third-party action runs inside the existing job context and exfiltrates the workflow `GITHUB_TOKEN` via build logs. The blast radius equals the token's issued scope.

Style matches the workflow-level `permissions:` block already in `ci.yml`. YAML validated locally with `yaml.safe_load`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,9 @@ defaults:
   run:
     shell: bash
 
+permissions:
+  contents: read
+
 jobs:
   test:
     name: test (${{ matrix.host_target }})
--- a/.github/workflows/sysroots.yml
+++ b/.github/workflows/sysroots.yml
@@ -8,6 +8,9 @@ defaults:
   run:
     shell: bash
 
+permissions:
+  contents: read
+
 jobs:
   sysroots:
     name: Build the sysroots


RalfJung:
I'm afraid I don't know GHA that well so I can't tell if this is a good idea or not.

@rust-lang/infra do you have a GHA expert who could help us here?

bjorn3:
Given that this is a public repo (and thus there is an implicit read permission for the repo contents on any github token), `permissions: {}` to not give any permissions at all should work too. By the way ci.yml doesn't contain a top-level `permissions` block, only an override for a single job. I don't know if that override is enough for GHA to disable all permissions for the rest of the workflow or not. Can't hurt to add it there too.

Kobzol:
I cannot manage to find the default GITHUB_TOKEN permissions in GitHub's docs, they are terrible in this.. But I think that `contents: read` is already the default?

bjorn3:
It depends on when the repo was created whether GHA by default only has read or also write permissions. You can change it in the repo settings.

ubiratansoares:
> I cannot manage to find the default GITHUB_TOKEN permissions in GitHub's docs, they are terrible in this.. But I think that contents: read is already the default?

Those are quite scattered, unfortunately.  `contents:read` are not necessarily the default, as @bjorn3 mentioned. I can recommend [zizmor docs](https://docs.zizmor.sh/audits/#excessive-permissions) on `excessive-permissions` as an alternative source on this topic. I think this PR put us in the right direction :)

Btw, running zizmor in this repos brings this warning (amongst other findings)

```
warning[excessive-permissions]: overly broad permissions
  --> .github/workflows/sysroots.yml:1:1
   |
 1 | / name: Tier 2 sysroots
 2 | |
 3 | | on:
 4 | |   schedule:
...  |
62 | |           Sincerely,
63 | |           The Miri Cronjobs Bot'
   | |_________________________________^ default permissions used due to no permissions: block
   |
   = note: audit confidence → Medium
```

RalfJung:
So the conclusion is, we should have `permissions: read` for both CI yml files?

ubiratansoares:
That's the recommendation at Workflow level, yes. Or, as @bjorn3 mentioned, setting `permissions: {}`. 

If for some reason some job requires more permissions, one can grant them at level Job with an additional `permissions` block

Edit : note that the block should be as defined in the PR, `permissions: read` probably would bring a Workflow parsing error

arpitjain099:
Thanks @ubiratansoares, @bjorn3, @Kobzol, @RalfJung for the GHA deep-dive. Sounds like the consensus matches the change in this PR: `permissions: contents: read` at workflow level on `sysroots.yml` (with `permissions: {}` as an even-tighter alternative since the workflow doesn't talk to the GitHub API at all).

Happy to wait on whatever direction you settle on. If you'd like me to extend the same to `ci.yml` (which @bjorn3 noted only has the per-job override), I can either roll it into this PR or open a follow-up, whichever fits your review preference. Just let me know.

RalfJung:
Yes please add a similar section to `ci.yml` as well.

arpitjain099:
> Yes please add a similar section to `ci.yml` as well.

Thanks @RalfJung will do.

arpitjain099:
@RalfJung added `permissions: contents: read` to `ci.yml` as well. Let me know if anything else.

RalfJung:
Thanks :)

Issue #4183, opened 2025-02-07
`std::fs::copy` is unsupported


anatawa12:
When we call `std::fs::copy` on macos, we encountered ``` unsupported operation: can't call foreign function `fclonefileat` on OS `macos` ``` error.

I think `std::fs::copy` is a basic API commonly used so I reported here as miri error message suggests.

I searched for `fclonefileat` but I could not found existing issue

RalfJung:
Thanks for the report!

Yes it would be reasonable to support this; no idea how tricky it would be though. I don't think we support `fs::copy` on any OS so far, so implementing it on Linux would likely be the first step.

RalfJung:
Looks like on Linux,` fs:;copy` falls back to `io::copy`, so this might already work.

RalfJung:
At least on Linux the main required shim for this seems to be "copy_file_range".  That would be the first step.

In a next step we also have to do something for "splice" and "sendfile64". However, we can hopefully tell std that those are not supported and therefore avoid having to implement them.

PR #5041, opened 2026-05-15
Add Linux copy_file_range shim

eval-exec:
Refs: https://github.com/rust-lang/miri/issues/4183#issuecomment-4413693178

> At least on Linux the main required shim for this seems to be "copy_file_range". That would be the first step.

Friendly invite @RalfJung to review.

https://codebrowser.dev/linux/linux/fs/read_write.c.html#1648

Some places I'm aware for `fn copy_file_range`'s limitation:

1. `fn copy_file_range` currently rejects same `file_in` and `file_out`, but linux does not reject “same file” by itself. It rejects “same file with overlapping source/destination ranges.” 

In https://man7.org/linux/man-pages/man2/copy_file_range.2.html:
>        fd_in and fd_out can refer to the same file.  If they refer to the
>       same file, then the source and target ranges are not allowed to
>       overlap.
>
>       The flags argument is provided to allow for future extensions and
>       currently must be set to 0.

But in this PR:

```rust
        if file_in == file_out {
            return this.set_last_error_and_return(LibcError("EINVAL"), dest);
        }
```

### DIFF
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -395,6 +395,185 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
 impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    fn copy_file_range(
+        &mut self,
+        fd_in_op: &OpTy<'tcx>,
+        off_in_op: &OpTy<'tcx>,
+        fd_out_op: &OpTy<'tcx>,
+        off_out_op: &OpTy<'tcx>,
+        len_op: &OpTy<'tcx>,
+        flags_op: &OpTy<'tcx>,
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx> {
+        let this = self.eval_context_mut();
+
+        let fd_in = this.read_scalar(fd_in_op)?.to_i32()?;
+        let fd_out = this.read_scalar(fd_out_op)?.to_i32()?;
+        let len = this.read_target_usize(len_op)?;
+        let flags = this.read_scalar(flags_op)?.to_u32()?;
+
+        if flags != 0 {
+            return this.set_last_error_and_return(LibcError("EINVAL"), dest);
+        }
+
+        let Some(fd_in) = this.machine.fds.get(fd_in) else {
+            return this.set_last_error_and_return(LibcError("EBADF"), dest);
+        };
+        let Some(fd_out) = this.machine.fds.get(fd_out) else {
+            return this.set_last_error_and_return(LibcError("EBADF"), dest);
+        };
+        let Some(file_in) = fd_in.downcast::<FileHandle>() else {
+            return this.set_last_error_and_return(LibcError("EINVAL"), dest);
+        };
+        let Some(file_out) = fd_out.downcast::<FileHandle>() else {
+            return this.set_last_error_and_return(LibcError("EINVAL"), dest);
+        };
+        if file_in == file_out {
+            return this.set_last_error_and_return(LibcError("EINVAL"), dest);
+        }
+        if !file_out.writable {
+            return this.set_last_error_and_return(LibcError("EBADF"), dest);
+        }
+
+        let loff_t = this.libc_ty_layout("loff_t");
+        let off_in_ptr = this.read_pointer(off_in_op)?;
+        let off_out_ptr = this.read_pointer(off_out_op)?;
+        let mut off_in = if this.ptr_is_null(off_in_ptr)? {
+            None
+        } else {
+            let place = this.deref_pointer_as(off_in_op, loff_t)?;
+            let offset = this.read_scalar(&place)?.to_i64()?;
+            if offset < 0 {
+                return this.set_last_error_and_return(LibcError("EINVAL"), dest);
+            }
+            Some((place, u64::try_from(offset).unwrap()))
+        };
+        let mut off_out = if this.ptr_is_null(off_out_ptr)? {
+            None
+        } else {
+            let place = this.deref_pointer_as(off_out_op, loff_t)?;
+            let offset = this.read_scalar(&place)?.to_i64()?;
+            if offset < 0 {
+                return this.set_last_error_and_return(LibcError("EINVAL"), dest);
+            }
+            Some((place, u64::try_from(offset).unwrap()))
+        };
+
+        // The kernel is permitted to perform a partial copy. Keep the host allocation bounded;
+        // std will keep calling this shim until EOF or an error is reached.
+        let count = len
+            .min(u64::try_from(this.target_isize_max()).unwrap())
+            .min(u64::try_from(isize::MAX).unwrap())
+            .min(1024 * 1024);
+        let count = usize::try_from(count).unwrap();
+        if count == 0 {
+            this.write_null(dest)?;
+            return interp_ok(());
+        }
+
+        let communicate = this.machine.communicate();
+        assert!(communicate, "isolation should have prevented even opening a file");
+
+        let mut src = &file_in.file;
+        let mut dst = &file_out.file;
+        let src_start = if off_in.is_none() {
+            match src.stream_position() {
+                Ok(pos) => Some(pos),
+                Err(err) => return this.set_last_error_and_return(err, dest),
+            }
+        } else {
+            None
+        };
+        let dst_start = if off_out.is_none() {
+            match dst.stream_position() {
+                Ok(pos) => Some(pos),
+                Err(err) => return this.set_last_error_and_return(err, dest),
+            }
+        } else {
+            None
+        };
+
+        let mut bytes = vec![0; count];
+        let read_result = if let Some((_, offset)) = off_in {
+            let cur = match src.stream_position() {
+                Ok(pos) => pos,
+                Err(err) => return this.set_last_error_and_return(err, dest),
+            };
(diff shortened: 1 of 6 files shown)


RalfJung:
Thanks for the PR!

Could you please confirm that you will be able to explain the choices that this code makes in your own words during the review? I expect that you are the author of the code in this PR, in the sense that you are willing to take responsibility for everything it does well and all its flaws.   It's okay to use tools for initial drafts, but only if you are willing to stake your name on what you ultimately end up submitting for review, and if you dig into the code and its design choices until you understand them yourself.  If you had to guess the answer to some questions in order to get the PR to work, that's fine; please put all those questions and your guesses into the PR description or add them as comments to the relevant part of the diff.

eval-exec:
> Could you please confirm that you will be able to explain the choices that this code makes in your own words during the review?

Yes, I confirm.

> please put all those questions and your guesses into the PR description or add them as comments to the relevant part of the diff.

Sure, I'll update, thank you. 

---

Updated, @RalfJung Hello, Does this PR match what you had in mind for the first Linux step?

RalfJung:
I have added this to my review queue and will take a look eventually. :) However, this may take a week or two as I also have other obligations (and next week all Rust project members will be busy meeting in Utrecht).

eval-exec:
> I have added this to my review queue and will take a look eventually. :) However, this may take a week or two as I also have other obligations (and next week all Rust project members will be busy meeting in Utrecht).

I really appreciate your work. Do you have any other issues you would recommend I work on?

RalfJung:
I appreciate your enthusiasm, but it is best to wait for the first round of feedback so you can apply what you learned to your second PR. :)

eval-exec:
> I appreciate your enthusiasm, but it is best to wait for the first round of feedback so you can apply what you learned to your second PR. :)

Thank you.

tests/pass/shims/fs-copy.rs
```diff
@@ -0,0 +1,27 @@
+//@compile-flags: -Zmiri-disable-isolation
+//@only-target: linux
```
RalfJung, line 2: Didn't you add the syscall support to make this also work on Android?

src/shims/unix/fs.rs
```diff
@@ -353,6 +353,204 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        len_op: &OpTy<'tcx>,
+        flags_op: &OpTy<'tcx>,
+        dest: &MPlaceTy<'tcx>,
+    ) -> InterpResult<'tcx> {
+        let this = self.eval_context_mut();
+
+        let fd_in = this.read_scalar(fd_in_op)?.to_i32()?;
+        let fd_out = this.read_scalar(fd_out_op)?.to_i32()?;
+        let len = this.read_target_usize(len_op)?;
+        let flags = this.read_scalar(flags_op)?.to_u32()?;
```
RalfJung, line 413: Please read *all* the arguments at the top of the function, following the pattern described [here](https://github.com/rust-lang/miri/blob/0f4e64d05388579bf15bbc5d0fbca44a49879a95/src/shims/foreign_items.rs#L279-L296).

src/shims/unix/fs.rs
```diff
@@ -353,6 +353,204 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            return this.set_last_error_and_return(LibcError("EINVAL"), dest);
+        };
+        let Some(file_out) = fd_out.downcast::<FileHandle>() else {
+            return this.set_last_error_and_return(LibcError("EINVAL"), dest);
+        };
+        // Linux permits same-file copies when the source and destination ranges do not overlap.
+        // We conservatively reject all same-file copies for now.
+        if file_in == file_out {
+            return this.set_last_error_and_return(LibcError("EINVAL"), dest);
+        }
```
RalfJung: This is not a waterproof test I think -- what if I just open the same file twice by calling `open` twice? That won't be caught here.

src/shims/unix/fs.rs
```diff
@@ -353,6 +353,204 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        let communicate = this.machine.communicate();
+        assert!(communicate, "isolation should have prevented even opening a file");
+
+        let mut src = &file_in.file;
+        let mut dst = &file_out.file;
+        let src_start = if off_in.is_none() {
+            match src.stream_position() {
+                Ok(pos) => Some(pos),
+                Err(err) => return this.set_last_error_and_return(err, dest),
```
RalfJung, line 482: Is it correct to forward errors from `stream_position` here? How could that even fail?

I think ideally you'd avoid any seeking if `off_in` is `None`.

src/shims/unix/fs.rs
```diff
@@ -353,6 +353,204 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            Err(err) => {
+                if let Some(src_start) = src_start {
+                    src.seek(SeekFrom::Start(src_start))
+                        .expect("failed to restore file position, this shouldn't be possible");
+                }
+                if let Some(dst_start) = dst_start {
+                    dst.seek(SeekFrom::Start(dst_start))
+                        .expect("failed to restore file position, this shouldn't be possible");
+                }
+                return this.set_last_error_and_return(err, dest);
```
RalfJung, line 565: Having to remember to restore the position on every early return seems very error-prone. Is there some way we can avoid that? For instance you could have a block and restore the desired positions afterwards:
```rust
let result = 'result: {
  ...
  if ... { break 'result this.set_last_error_and_return(err, dest); }
};
// ... fix file positions ...
result
```

src/shims/unix/fs.rs
```diff
@@ -353,6 +353,204 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            if write_size < read_size {
+                src.seek(SeekFrom::Start(src_start.strict_add(copied)))
+                    .expect("failed to restore file position, this shouldn't be possible");
+            }
+        }
+        if let Some((place, offset)) = off_in.as_mut() {
+            let Some(new_offset) = offset.checked_add(i64::try_from(write_size).unwrap()) else {
+                return this.set_last_error_and_return(LibcError("EOVERFLOW"), dest);
+            };
+            *offset = new_offset;
```
RalfJung, line 579: Seems a bit late to check this now? Shouldn't we check at the beginning that the requested offsets and size do not overflow?

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -60,6 +62,121 @@ fn main() {
...
+    use std::ptr;
+
+    let bytes = b"abcdef";
+    let src_path = utils::prepare_with_content("miri_test_libc_copy_file_range_source.txt", bytes);
+    let dst_path = utils::prepare("miri_test_libc_copy_file_range_destination.txt");
+    let src = File::open(&src_path).unwrap();
+    let dst = File::create(&dst_path).unwrap();
+
+    unsafe {
+        let ret = libc::copy_file_range(
```
RalfJung, line 90: Please use `errno_result` to invoke `copy_file_range` so you get a bit of automatic error handling.

tests/pass-dep/libc/libc-fs.rs
```diff
@@ -60,6 +62,121 @@ fn main() {
...
+            1,
+            0,
+        );
+        assert_eq!(ret, -1);
+        assert_eq!(libc::lseek(src.as_raw_fd(), 0, libc::SEEK_CUR), 0);
+    }
+    assert_eq!(Error::last_os_error().raw_os_error(), Some(libc::EOVERFLOW));
+    assert_eq!(overflow_off_out, i64::MAX as libc::loff_t);
+    assert_eq!(std::fs::read(&overflow_dst_path).unwrap(), b"");
+
```
RalfJung, line 167: Please add a few comments explaining what you are testing.

src/shims/unix/fs.rs
```diff
@@ -353,6 +353,204 @@ trait EvalContextExtPrivate<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        let mut off_in = if this.ptr_is_null(off_in_ptr)? {
+            None
+        } else {
+            let place = this.deref_pointer_as(off_in_op, loff_t)?;
+            let offset = this.read_scalar(&place)?.to_i64()?;
+            if offset < 0 {
+                return this.set_last_error_and_return(LibcError("EINVAL"), dest);
+            }
+            Some((place, offset))
+        };
```
RalfJung, line 450: Does the standard library use non-null values for the offset pointers?

If not, then maybe the MVP shouldn't have that either. It's easier to review smaller PRs so it'd be better to add this separately later.

RalfJung (commented):
Wow that function really does a lot.^^ Here's a first round of comments. :)

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5041/48b0b923635281fad0c7597c7c704a7d7fe10cf2..16872948272bef7b28461064f284fc489a21cb5f)*

RalfJung:
@eval-exec do you still plan to get back to this PR? It's fine if not, we should just know about it so someone else can then pick up the work. :)
I think it would help to start smaller, and implement just a particular special case of this function while throwing "unsupported" for everything else. That would also make it easier to review.

eval-exec:
@RalfJung  Yes, I still plan to get back to it. I've just been busy recently.
Thank you, I'll continue this PR tomorrow.

RalfJung:
No need to rush, I just wanted to prevent this from going stale. :)

PR #5042, opened 2026-05-15
Add network socket read and write timeout

WhySoBad:
Hi,

This pull request adds support for the `SO_SNDTIMEO` and `SO_RCVTIMEO` socket options. 
Because the mio sockets are non-blocking, we implement this ourselves on top of mio's sockets. 
By default both timeouts are `None`, which means that the socket will block indefinitely. When changed to `Some`, we provide a `timeout` argument to `this.block_thread_for_io` with a deadline which achieves the same as a read/write timeout on blocking sockets.
For non-blocking sockets those options are ignored.

### DIFF
--- a/src/clock.rs
+++ b/src/clock.rs
@@ -11,12 +11,12 @@ use crate::MiriMachine;
 const NANOSECONDS_PER_BASIC_BLOCK: u128 = 5000;
 
 /// An instant (a fixed moment in time) in Miri's monotone clock.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct Instant {
     kind: InstantKind,
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 enum InstantKind {
     Host(StdInstant),
     Virtual { nanoseconds: u128 },
@@ -134,7 +134,7 @@ impl MonotonicClock {
 }
 
 /// A deadline for some event to occur.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub enum Deadline {
     Monotonic(Instant),
     RealTime(SystemTime),
--- a/src/shims/unix/socket.rs
+++ b/src/shims/unix/socket.rs
@@ -3,6 +3,7 @@ use std::io;
 use std::io::Read;
 use std::net::{Ipv4Addr, Shutdown, SocketAddr, SocketAddrV4};
 use std::sync::atomic::AtomicBool;
+use std::time::Duration;
 
 use mio::event::Source;
 use mio::net::{TcpListener, TcpStream};
@@ -67,6 +68,18 @@ struct Socket {
     io_readiness: RefCell<BlockingIoSourceReadiness>,
     /// [`Some`] when the socket had an async error which has not yet been fetched via `SO_ERROR`.
     error: RefCell<Option<io::Error>>,
+    /// Read timeout of the socket. [`None`] means that reads can block indefinitely.
+    /// The timeout is applied to the monotonic clock (the Unix specification doesn't
+    /// specify which clock to use, but the monotonic clock is more common for
+    /// relative timeouts).
+    /// This is ignored when the socket is non-blocking.
+    read_timeout: Cell<Option<Duration>>,
+    /// Write timeout of the socket. [`None`] means that writes can block indefinitely.
+    /// The timeout is applied to the monotonic clock (the Unix specification doesn't
+    /// specify which clock to use, but the monotonic clock is more common
+    /// for relative timeouts).
+    /// This is ignored when the socket is non-blocking.
+    write_timeout: Cell<Option<Duration>>,
 }
 
 impl FileDescription for Socket {
@@ -108,14 +121,16 @@ impl FileDescription for Socket {
         assert!(communicate_allowed, "cannot have `Socket` with isolation enabled!");
 
         let socket = self;
+        let deadline = ecx.action_deadline(socket.is_non_block.get(), socket.read_timeout.get());
 
         ecx.ensure_connected(
             socket.clone(),
-            !socket.is_non_block.get(),
+            deadline.clone(),
             "read",
             callback!(
                 @capture<'tcx> {
                     socket: FileDescriptionRef<Socket>,
+                    deadline: Option<Deadline>,
                     ptr: Pointer,
                     len: usize,
                     finish: DynMachineCallback<'tcx, Result<usize, IoError>>,
@@ -134,8 +149,8 @@ impl FileDescription for Socket {
                         finish.call(this, result)
                     } else {
                         // The socket is in blocking mode and thus the read call should block
-                        // until we can read some bytes from the socket.
-                        this.block_for_recv(socket, ptr, len, /* should_peek */ false, finish)
+                        // until we can read some bytes from the socket or the timeout exceeded.
+                        this.block_for_recv(socket, deadline, ptr, len, /* should_peek */ false, finish)
                     }
                 }
             ),
@@ -153,14 +168,16 @@ impl FileDescription for Socket {
         assert!(communicate_allowed, "cannot have `Socket` with isolation enabled!");
 
         let socket = self;
+        let deadline = ecx.action_deadline(socket.is_non_block.get(), socket.write_timeout.get());
 
         ecx.ensure_connected(
             socket.clone(),
-            !socket.is_non_block.get(),
+            deadline.clone(),
             "write",
             callback!(
                 @capture<'tcx> {
                     socket: FileDescriptionRef<Socket>,
+                    deadline: Option<Deadline>,
                     ptr: Pointer,
                     len: usize,
                     finish: DynMachineCallback<'tcx, Result<usize, IoError>>
@@ -179,8 +196,8 @@ impl FileDescription for Socket {
                         return finish.call(this, result)
                     } else {
                         // The socket is in blocking mode and thus the write call should block
-                        // until we can write some bytes into the socket.
-                        this.block_for_send(socket, ptr, len, finish)
+                        // until we can write some bytes into the socket or the timeout exceeded.
+                        this.block_for_send(socket, deadline, ptr, len, finish)
                     }
                 }
             ),
@@ -353,6 +370,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             is_non_block: Cell::new(is_sock_nonblock),
             io_readiness: RefCell::new(BlockingIoSourceReadiness::empty()),
             error: RefCell::new(None),
+            read_timeout: Cell::new(None),
+            write_timeout: Cell::new(None),
         });
 
         interp_ok(Scalar::from_i32(fds.insert(fd)))
@@ -567,6 +586,17 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         } else {
             // The socket is in blocking mode and thus the accept call should block
             // until an incoming connection is ready.
+
+            if socket.read_timeout.get().is_some() {
+                // Some Unixes like Linux also apply the SO_RCVTIMEO socket option
+                // to `accept` calls:
+                // <https://github.com/torvalds/linux/blob/HEAD/net/ipv4/inet_connection_sock.c#L668-L675>
+                // This is currently not supported by Miri.
+                throw_unsup_format!(
+                    "accept4: blocking accept is not supported when SO_RCVTIMEO is non-zero"
+                )
+            }
+
             this.block_for_accept(
                 socket,
                 address_ptr,
@@ -645,11 +675,20 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             // The socket is in blocking mode and thus the connect call should block
             // until the connection with the server is established.
 
-            let dest = dest.clone();
+            if socket.write_timeout.get().is_some() {
+                // Some Unixes like Linux also apply the SO_SNDTIMEO socket option
+                // to `connect` calls:
+                // <https://github.com/torvalds/linux/blob/HEAD/net/ipv4/af_inet.c#L701-L710>
+                // This is currently not supported by Miri.
+                throw_unsup_format!(
+                    "connect: blocking connect is not supported when SO_SNDTIMEO is non-zero"
+                )
+            }
 
+            let dest = dest.clone();
             this.ensure_connected(
                 socket.clone(),
-                /* should_wait */ true,
+                /* deadline */ None,
                 "connect",
                 callback!(
                     @capture<'tcx> {
@@ -729,29 +768,29 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             );
         }
 
-        // If either the operation or the socket is non-blocking, we don't want
-        // to wait until the connection is established.
-        let should_wait = !is_op_non_block && !socket.is_non_block.get();
+        let is_non_block = is_op_non_block || socket.is_non_block.get();
+        let deadline = this.action_deadline(is_non_block, socket.write_timeout.get());
         let dest = dest.clone();
 
         this.ensure_connected(
             socket.clone(),
-            should_wait,
+            deadline.clone(),
             "send",
             callback!(
                 @capture<'tcx> {
                     socket: FileDescriptionRef<Socket>,
+                    deadline: Option<Deadline>,
                     flags: i32,
                     buffer_ptr: Pointer,
                     length: usize,
-                    is_op_non_block: bool,
+                    is_non_block: bool,
                     dest: MPlaceTy<'tcx>,
                 } |this, result: Result<(), ()>| {
                     if result.is_err() {
                         return this.set_errno_and_return_neg1(LibcError("ENOTCONN"), &dest)
                     }
 
-                    if is_op_non_block || socket.is_non_block.get() {
+                    if is_non_block {
                         // We have a non-blocking operation or a non-blocking socket and
                         // thus don't want to block until we can send.
                         match this.try_non_block_send(&socket, buffer_ptr, length)? {
@@ -760,9 +799,10 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                         }
                     } else {
                         // The socket is in blocking mode and thus the send call should block
-                        // until we can send some bytes into the socket.
+                        // until we can send some bytes into the socket or the timeout exceeded.
                         this.block_for_send(
                             socket,
+                            deadline,
                             buffer_ptr,
                             length,
                             callback!(@capture<'tcx> {
@@ -850,29 +890,29 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             );
         }
 
-        // If either the operation or the socket is non-blocking, we don't want
-        // to wait until the connection is established.
-        let should_wait = !is_op_non_block && !socket.is_non_block.get();
+        let is_non_block = is_op_non_block || socket.is_non_block.get();
+        let deadline = this.action_deadline(is_non_block, socket.read_timeout.get());
         let dest = dest.clone();
 
         this.ensure_connected(
             socket.clone(),
-            should_wait,
+            deadline.clone(),
             "recv",
             callback!(
                 @capture<'tcx> {
                     socket: FileDescriptionRef<Socket>,
+                    deadline: Option<Deadline>,
                     buffer_ptr: Pointer,
                     length: usize,
                     should_peek: bool,
-                    is_op_non_block: bool,
+                    is_non_block: bool,
                     dest: MPlaceTy<'tcx>,
                 } |this, result: Result<(), ()>| {
                     if result.is_err() {
                         return this.set_errno_and_return_neg1(LibcError("ENOTCONN"), &dest)
                     }
 
-                    if is_op_non_block || socket.is_non_block.get() {
+                    if is_non_block {
                         // We have a non-blocking operation or a non-blocking socket and
                         // thus don't want to block until we can receive.
                         match this.try_non_block_recv(&socket, buffer_ptr, length, should_peek)? {
@@ -881,9 +921,10 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                         }
                     } else {
                         // The socket is in blocking mode and thus the receive call should block
-                        // until we can receive some bytes from the socket.
+                        // until we can receive some bytes from the socket or the timeout exceeded.
                         this.block_for_recv(
                             socket,
+                            deadline,
                             buffer_ptr,
                             length,
                             should_peek,
@@ -930,6 +971,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         };
 
         if level == this.eval_libc_i32("SOL_SOCKET") {
+            let opt_so_rcvtimeo = this.eval_libc_i32("SO_RCVTIMEO");
+            let opt_so_sndtimeo = this.eval_libc_i32("SO_SNDTIMEO");
             let opt_so_reuseaddr = this.eval_libc_i32("SO_REUSEADDR");
 
             if matches!(this.tcx.sess.target.os, Os::MacOs | Os::FreeBsd | Os::NetBsd) {
@@ -948,6 +991,25 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
                     return interp_ok(Scalar::from_i32(0));
                 }
+            }
+
+            if option_name == opt_so_rcvtimeo || option_name == opt_so_sndtimeo {
+                let timeval_layout = this.libc_ty_layout("timeval");
+                let option_value = this.ptr_to_mplace(option_value_ptr, timeval_layout);
+
+                let timeout = match this.read_timeval(&option_value)? {
+                    None => return this.set_errno_and_return_neg1_i32(LibcError("EINVAL")),
+                    Some(Duration::ZERO) => None,
+                    Some(duration) => Some(duration),
+                };
+
+                if option_name == opt_so_rcvtimeo {
+                    socket.read_timeout.set(timeout);
+                } else {
+                    socket.write_timeout.set(timeout);
+                }
+
+                return interp_ok(Scalar::from_i32(0));
             }
 
             if option_name == opt_so_reuseaddr {
@@ -1085,6 +1147,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // buffer, in which case we have to truncate.
         let value_buffer = if level == this.eval_libc_i32("SOL_SOCKET") {
             let opt_so_error = this.eval_libc_i32("SO_ERROR");
+            let opt_so_rcvtimeo = this.eval_libc_i32("SO_RCVTIMEO");
+            let opt_so_sndtimeo = this.eval_libc_i32("SO_SNDTIMEO");
 
             if option_name == opt_so_error {
                 // Reading SO_ERROR should always return the latest async error. Because our stored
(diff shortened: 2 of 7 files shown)


src/shims/unix/socket.rs
```diff
@@ -125,8 +131,10 @@ impl FileDescription for Socket {
                         finish.call(this, result)
                     } else {
                         // The socket is in blocking mode and thus the read call should block
-                        // until we can read some bytes from the socket.
-                        this.block_for_recv(socket, ptr, len, /* should_peek */ false, finish)
+                        // until we can read some bytes from the socket or the timeout exceeded.
+                        let deadline = socket.read_timeout.get()
+                            .map(|duration| this.machine.monotonic_clock.now().add_lossy(duration));
```
RalfJung: So here you decided that the read timeout uses the "monotonic" clock rather than the "real time" clock. Is that backed by documentation? If so, please reference that.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5042/48b0b923635281fad0c7597c7c704a7d7fe10cf2..b4bbd327e1cc35bb1f951fbf9e05623894fbe50c)*

src/shims/unix/socket.rs
```diff
@@ -1513,30 +1586,45 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     fn block_for_send(
         &mut self,
         socket: FileDescriptionRef<Socket>,
+        deadline: Option<Instant>,
```
RalfJung: With https://github.com/rust-lang/miri/pull/5044, this should now be `Option<Deadline>`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5042/48b0b923635281fad0c7597c7c704a7d7fe10cf2..b4bbd327e1cc35bb1f951fbf9e05623894fbe50c)*

src/shims/unix/socket.rs
WhySoBad: I couldn't find documentation about whether it uses the monotonic clock or the real time clock on real Unix systems.

RalfJung: Hm, we should figure that out. I guess monotone makes sense since it is a relative timeout, but OTOH most old APIs use the real-time clock by default.

@the8472 do you happen to know which clock SO_RCVTIMEO/SO_SNDTIMEO use?

src/shims/unix/socket.rs
```diff
@@ -124,8 +131,10 @@ impl FileDescription for Socket {
                         finish.call(this, result)
                     } else {
                         // The socket is in blocking mode and thus the read call should block
-                        // until we can read some bytes from the socket.
-                        this.block_for_recv(socket, ptr, len, /* should_peek */ false, finish)
+                        // until we can read some bytes from the socket or the timeout exceeded.
```
RalfJung, line 152:
```suggestion
                        // until we can read some bytes from the socket or the timeout exceeded.
                        // FIXME: which clock should this deadline use?
```

src/shims/unix/socket.rs
```diff
@@ -124,8 +131,10 @@ impl FileDescription for Socket {
...
+                        let deadline = socket.read_timeout.get()
+                            .map(|duration| this.machine.monotonic_clock.now().add_lossy(duration).into());
```
RalfJung: Don't you have to call this *before* calling `ensure_connected` to avoid losing track of the time spent there?

src/shims/unix/socket.rs
```diff
@@ -1744,9 +1844,11 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     // Execute the provided callback function when the socket is either in
     // [`SocketState::Connected`] or an error occurred.
     /// If the socket is currently neither in the [`SocketState::Connecting`] nor
-    /// the [`SocketState::Connecting`] state, an ENOTCONN error is returned.
-    /// When the callback function is called with `Ok(_)`, then we're guaranteed
+    /// the [`SocketState::Connecting`] state, [`Err`] is returned.
+    /// When the callback function is called with [`Ok`], then we're guaranteed
     /// that the socket is in the [`SocketState::Connected`] state.
+    /// The boolean flag of the [`Err`] variant indicates whether the connection
+    /// failed because of an exceeded timeout.
```
```suggestion
    /// failed because of an exceeded timeout (`true`) or not (`false`).
```

src/shims/unix/socket.rs
```diff
@@ -638,11 +656,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                    } |this, result: Result<(), ()>| {
-                        if result.is_err() {
-                            this.set_last_error_and_return(LibcError("ENOTCONN"), &dest)
-                        } else {
-                            this.write_scalar(Scalar::from_i32(0), &dest)
+                    } |this, result: Result<(), bool>| {
+                        match result {
+                            Ok(_) => this.write_scalar(Scalar::from_i32(0), &dest),
+                            Err(/* timed out */ true) =>
+                                this.set_last_error_and_return(IoError::from(io::ErrorKind::InProgress), &dest),
```
RalfJung: Please use use `LibcError` here to remain consistent.

src/shims/unix/socket.rs
```diff
@@ -739,9 +759,12 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                         }
                     } else {
                         // The socket is in blocking mode and thus the send call should block
-                        // until we can send some bytes into the socket.
+                        // until we can send some bytes into the socket or the timeout exceeded.
+                        let deadline = socket.write_timeout.get()
+                            .map(|duration| this.machine.monotonic_clock.now().add_lossy(duration).into());
```
RalfJung: Looks like we have some code duplication between send and write (and recv and read). Would be good to clean that up (in a future PR).

src/shims/unix/socket.rs
```diff
@@ -929,6 +957,33 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let sec = this.read_scalar(&sec_field)?.to_int(sec_field.layout.size)?;
+
+                let usec_field = this.project_field_named(&option_value, "tv_usec")?;
+                let usec = this.read_scalar(&usec_field)?.to_int(usec_field.layout.size)?;
+
+                let timeout = if sec == 0 && usec == 0 {
+                    None
+                } else {
+                    let sec = u64::try_from(sec).unwrap();
+                    let nsec = u32::try_from(usec).unwrap().checked_mul(1000).unwrap();
```
```suggestion
                    let nsec = u32::try_from(usec).unwrap().strict_mul(1000);
```

src/shims/unix/socket.rs
```diff
@@ -1412,19 +1492,25 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-                assert_eq!(kind, UnblockKind::Ready);
-
                 // Remove the blocking I/O interest for unblocking this thread.
                 this.machine.blocking_io.remove_blocked_thread(socket.id(), this.machine.threads.active_thread());
 
+                match kind {
+                    UnblockKind::Ready => { /* fall-through to below */ },
+                    // When the read timeout is exceeded EAGAIN/EWOULDBLOCK is returned.
+                    UnblockKind::TimedOut =>
+                        return this.set_last_error_and_return(IoError::from(io::ErrorKind::WouldBlock), &dest)
```
RalfJung: Please use `LibcError`.

src/shims/unix/socket.rs
```diff
@@ -1520,22 +1609,27 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             } |this, kind: UnblockKind| {
-                assert_eq!(kind, UnblockKind::Ready);
-
                 // Remove the blocking I/O interest for unblocking this thread.
                 this.machine.blocking_io.remove_blocked_thread(socket.id(), this.machine.threads.active_thread());
 
+                match kind {
+                    UnblockKind::Ready => { /* fall-through to below */ },
+                    // When the write timeout is exceeded EAGAIN/EWOULDBLOCK is returned.
+                    UnblockKind::TimedOut => return finish.call(this, Err(io::ErrorKind::WouldBlock.into()))
```
RalfJung: More `LibcError`.

src/shims/unix/socket.rs
```diff
@@ -1635,23 +1730,28 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             } |this, kind: UnblockKind| {
-                assert_eq!(kind, UnblockKind::Ready);
-
                 // Remove the blocking I/O interest for unblocking this thread.
                 this.machine.blocking_io.remove_blocked_thread(socket.id(), this.machine.threads.active_thread());
 
+                match kind {
+                    UnblockKind::Ready => { /* fall-through to below */ },
+                    // When the read timeout is exceeded EAGAIN/EWOULDBLOCK is returned.
+                    UnblockKind::TimedOut => return finish.call(this, Err(io::ErrorKind::WouldBlock.into()))
```
RalfJung: More `LibcError`.

src/shims/unix/socket.rs
```diff
@@ -1768,19 +1870,25 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        // used as a deadline. Otherwise, we use a zero duration timeout, i.e. we return
+        // immediately (but we still go through the scheduler once -- which is fine).
+        let deadline = if should_wait {
+            socket
+                .write_timeout
+                .get()
+                .map(|duration| this.machine.monotonic_clock.now().add_lossy(duration).into())
+        } else {
+            Some(this.machine.monotonic_clock.now().into())
+        };
```
RalfJung: Why is it correct to always use the write timeout?

It seems to me we should just have the caller pass in an `Option<Deadline>`.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -664,3 +667,99 @@ fn test_getsockopt_truncate() {
...
+    assert_eq!(timeout.tv_usec, short_timeout.tv_usec);
+
+    // The read timeout also affects blocking `accept`s.
+    // Accepting without an incoming connection should not block
+    // indefinitely but at least 50 milliseconds.
+    let before = Instant::now();
+    let err = net::accept_ipv4(server_sockfd).unwrap_err();
+    assert_eq!(err.kind(), ErrorKind::WouldBlock);
+    // Ensure that we blocked for at least 50 milliseconds.
+    assert!(Instant::now().duration_since(before) > Duration::from_millis(50));
```
```suggestion
    assert!(Instant::now().duration_since(before) >= Duration::from_millis(50));
```

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -664,3 +667,99 @@ fn test_getsockopt_truncate() {
...
+    let mut buffer = [0u8; 16];
+    // The read should return an EAGAIN/EWOULDBLOCK after ~10ms instead of blocking indefinitely.
+    let before = Instant::now();
+    let err = unsafe {
+        errno_result(libc::read(client_sockfd, buffer.as_mut_ptr().cast(), buffer.len()))
+            .unwrap_err()
+    };
+    assert_eq!(err.kind(), ErrorKind::WouldBlock);
+    // Ensure that we blocked for at least 50 milliseconds.
+    assert!(Instant::now().duration_since(before) > Duration::from_millis(50))
```
```suggestion
    assert!(Instant::now().duration_since(before) >= Duration::from_millis(50))
```

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -664,3 +667,99 @@ fn test_getsockopt_truncate() {
...
+    loop {
+        let result = unsafe {
+            errno_result(libc::write(client_sockfd, buffer.as_ptr().cast(), buffer.len()))
+        };
+        match result {
+            Ok(_) => { /* continue to fill up buffer */ }
+            // When we get an EAGAIN/EWOULDBLOCK when writing into a blocking socket, we know
+            // it's because of the write timeout exceeding because the write buffer
+            // is full.
+            Err(err) if err.kind() == ErrorKind::WouldBlock => break,
```
RalfJung: Can you also measure the block time here and ensure it makes sense?

tests/pass/shims/socket.rs
```diff
@@ -167,3 +170,56 @@ fn test_sockopt_nodelay() {
...
+
+/// Test setting and reading the SNDTIMEO socket option.
+/// This also tests that a read won't block indefinitely
+/// when the read timeout is set to [`Some`] duration.
+fn test_sockopt_read_timeout() {
+    let listener = TcpListener::bind("127.0.0.1:0").unwrap();
+    let address = listener.local_addr().unwrap();
+
+    let mut stream = TcpStream::connect(address).unwrap();
+    let _other_end = listener.accept().unwrap();
```
RalfJung, line 182: Please also do the "timeout on listener affects `accept`" part here.

RalfJung (commented):
So this affects read, write, accept -- but not connect?

@rustbot  author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5042/1e193dc41c50121f36baa044053102d273d108da..f84946cea6a90b2ed5cd3312c91e4c06734d6885)*

src/shims/unix/socket.rs
the8472: I think it's unspecified in POSIX. I am not certain what linux uses but I assume it's CLOCK_MONOTONIC since both `nanosleep` and `epoll_wait` also use CLOCK_MONOTOIC for their timeouts.

the8472: There's this in clock_settime

> Setting the value of the CLOCK_REALTIME clock via clock_settime() shall have no effect on threads that are blocked waiting for a relative time service based upon this clock, including the [nanosleep()](https://pubs.opengroup.org/onlinepubs/009695399/functions/nanosleep.html) function; nor on the expiration of relative timers based upon this clock. Consequently, these time services shall expire when the requested relative interval elapses, independently of the new or old value of the clock.

If we consider timeouts relative timers then that makes them essentially monotonic, not realtime.

RalfJung: Thanks!

That's an odd thing to put in the docs, and makes me wonder why one can even select a clock e.g. for FUTEX_WAIT (i.e. does the FUTEX_CLOCK_REALTIME flag even make a difference then).

But let's assume that it's the monotone clock then, but put in a comment that that's just a guess.

the8472: CLOCK_REALTIME-without-settime-adjustments is not necessarily the same as CLOCK_MONOTONIC. It just makes it _a_ monotonic clock.
They're even allowed to come from different hardware clock sources that drift apart over time, though afaik linux does use the same underlying source to implement both where possible.

RalfJung: With absolute timeouts (deadlines) they are clearly different, but with relative timeouts they are "almost" the same assuming the drift is sufficiently small. (It also seems hard to imagine under which conditions they'd actually drift in practice.)

src/shims/unix/socket.rs
WhySoBad: I've now added the `action_deadline` helper because the `ensure_connected` now takes an `Option<Deadline>` and I didn't want to manually create the zero timeout at multiple places.

tests/pass/shims/socket.rs
WhySoBad: The standard library doesn't expose `set_read_timeout` for `TcpListener`s. We can only test it with libc directly.

RalfJung:
> So this affects read, write, accept -- but not connect?

I guess not, otherwise std wouldn't have to use `poll` for `connect_timeout`.
But still, an odd choice by POSIX I would say.

WhySoBad:
No, `connect` is also affected. My current implemented is just wrong. 

It's documented in `man 7 socket`:

> SO_RCVTIMEO
     SO_SNDTIMEO
            Specify the receiving or sending timeouts until reporting an error.  The argument is a struct timeval.  If
            an input or output function blocks for this period of time, and data has been sent or received, the return
            value  of  that  function  will be the amount of data transferred; if no data has been transferred and the
            timeout has been reached, then -1 is returned with errno set to EAGAIN or EWOULDBLOCK, or EINPROGRESS (for
            connect(2)) just as if the socket was specified to be nonblocking.  If the timeout is set to zero (the de‐
            fault), then the operation will never timeout.  Timeouts only have effect for system  calls  that  perform
            socket  I/O  (e.g., accept(2), connect(2), read(2), recvmsg(2), send(2), sendmsg(2)); timeouts have no ef‐
            fect for select(2), poll(2), epoll_wait(2), and so on.

tests/pass-dep/libc/libc-socket.rs
WhySoBad: It turns out the write timeout works a bit differently. When the buffer is already full, the kernel instantly returns `EWOULDBLOCK` (I've tested it using `--native`). We cannot really test that it's happens instantly but the `Instant::now().duration_since(before) >= Duration::from_millis(50)` assertion fails for both `--native` and Miri.

RalfJung: So to be sure I understand -- if the timeout is 0 (i.e., no timeout), then a write on a full buffer blocks, but if there is a timeout, then it returns EWOULDBLOCK instantaneously? That seems very odd. I guess the spec technically allows early wakeups. But why would Linux do that?

Could you check if it behaves the same for non-localhost sockets?

RalfJung:
Odd, that makes me wonder why std uses `poll`... but that's a separate question.

FWIW if you want we can make that PR about just read/write and leave the rest for a future PR, to avoid making this one too big.

tests/pass-dep/libc/libc-socket.rs
WhySoBad: > if the timeout is 0 (i.e., no timeout), then a write on a full buffer blocks, but if there is a timeout, then it returns EWOULDBLOCK instantaneously

Correct.

> Could you check if it behaves the same for non-localhost sockets?

I've now tested it with a remote socket and the behavior is the same. When attempting to write after the write buffer is full, EWOULDBLOCK is instantly returned without awaiting the write timeout:
```rust
// ... `test_sockopt_sndtimeo`

let buffer = [1u8; 1024];
    loop {
        let result = unsafe {
            errno_result(libc::write(client_sockfd, buffer.as_ptr().cast(), buffer.len()))
        };
        match result {
            Ok(_) => { /* continue to fill up buffer */ }
            // When we get an EAGAIN/EWOULDBLOCK when writing into a blocking socket, we know
            // it's because of the write timeout exceeding because the write buffer
            // is full.
            Err(err) if err.kind() == ErrorKind::WouldBlock => break,
            Err(err) => panic!("unexpected error whilst filling up buffer: {err}"),
        }
    }

    let before = Instant::now();
    let err = unsafe {
        errno_result(libc::write(client_sockfd, buffer.as_ptr().cast(), buffer.len())).unwrap_err()
    };
    assert_eq!(err.kind(), ErrorKind::WouldBlock);
    // This assertion fails.
    assert!(Instant::now().duration_since(before) >= Duration::from_millis(50));
```

WhySoBad:
> Odd, that makes me wonder why std uses `poll`... but that's a separate question.

True, I'll ask on the Zulip because just using the SO_SNDTIMEO seems like a much simpler solution than having to use `poll` there.

Edit: Oh I see you already opened an discussion there :)

WhySoBad:
The latest force-push was to make this branch compatible with the changes from https://github.com/rust-lang/miri/pull/5049.

src/shims/unix/socket.rs
```diff
@@ -67,6 +68,16 @@ struct Socket {
     io_readiness: RefCell<BlockingIoSourceReadiness>,
     /// [`Some`] when the socket had an async error which has not yet been fetched via `SO_ERROR`.
     error: RefCell<Option<io::Error>>,
+    /// Read timeout of the socket. [`None`] means that reads can block indefinitely.
+    /// The timeout is applied to the monotonic clock (the Unix specification doesn't
+    /// specify whether a monotonic or a real-time clock should be used).
```
```suggestion
    /// The timeout is applied to the monotonic clock (the Unix specification doesn't
    /// specify which clock to use, but the monotonic clock is more common
    /// for relative timeouts).
```
Same for writes.

src/shims/unix/socket.rs
```diff
@@ -1415,6 +1506,28 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
 impl<'tcx> EvalContextPrivExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    /// Get the deadline for an action (e.g. reading or writing).
+    /// When `is_non_block` is [`true`], the returned deadline is a zero timeout.
+    /// If `action_timeout` is [`Some`], the returned deadline is the current
+    /// monotonic clock timestamp offset by `action_timeout`. Otherwise, no
+    /// deadline ([`None`]) is returned, indicating that the action can block
+    /// indefinitely.
```
```suggestion
    /// When `is_non_block` is [`true`], the returned deadline is "now", i.e.,
    /// we wake up immediately if the action cannot be completed.
    /// If `action_timeout` is `Some(duration)`, the returned deadline is in the future
    /// be the specified `duration`. Otherwise, no
    /// deadline ([`None`]) is returned, indicating that the action can block
    /// indefinitely.
```

src/shims/unix/socket.rs
```diff
@@ -645,11 +671,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             // The socket is in blocking mode and thus the connect call should block
             // until the connection with the server is established.
 
-            let dest = dest.clone();
+            if socket.read_timeout.get().is_some() {
```
RalfJung: This checks read_timeout but the error talks about SO_SNDTIMEO.

Do the docs say anywhere which of the timeouts is used for `connect`?

src/shims/unix/socket.rs
```diff
@@ -1312,7 +1402,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // UNIX targets should return ENOTCONN when the connection is not yet established.
         this.ensure_connected(
             socket.clone(),
-            /* should_wait */ false,
+            // Check whether the socket is connected without blocking.
+            this.action_deadline(/* is_non_block */ true, /* action_timeout */ None),
```
RalfJung: I think it would be more clear to just call `now` here directly.

src/shims/unix/socket.rs
```diff
@@ -1656,23 +1782,28 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             } |this, kind: UnblockKind| {
-                assert_eq!(kind, UnblockKind::Ready);
-
                 // Remove the blocking I/O interest for unblocking this thread.
                 this.machine.blocking_io.remove_blocked_thread(socket.id(), this.machine.threads.active_thread());
 
+                match kind {
+                    UnblockKind::Ready => { /* fall-through to below */ },
+                    // When the read timeout is exceeded EAGAIN/EWOULDBLOCK is returned.
+                    UnblockKind::TimedOut => return finish.call(this, Err(LibcError("EWOULDBLOCK")))
```
RalfJung, line 1796: I would have expected EAGAIN.
Is there any Unix we support where this makes a difference?

tests/pass-dep/libc/libc-socket.rs
RalfJung: > We cannot really test that it's happens instantly but the Instant::now().duration_since(before) >= Duration::from_millis(50) assertion fails for both --native and Miri.

Why does it fail like that for both? I would have expected that Miri correctly applies its own timeout and doesn't wake up early.

src/shims/unix/socket.rs
```diff
@@ -729,29 +762,30 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             );
         }
 
-        // If either the operation or the socket is non-blocking, we don't want
-        // to wait until the connection is established.
-        let should_wait = !is_op_non_block && !socket.is_non_block.get();
+        // Either the operation or the socket is non-blocking.
```
```suggestion
```

src/shims/unix/socket.rs
```diff
@@ -850,29 +885,30 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
```
```suggestion
```

src/shims/unix/socket.rs
```diff
@@ -950,6 +989,33 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let sec = this.read_scalar(&sec_field)?.to_int(sec_field.layout.size)?;
+
+                let usec_field = this.project_field_named(&option_value, "tv_usec")?;
+                let usec = this.read_scalar(&usec_field)?.to_int(usec_field.layout.size)?;
+
+                let timeout = if sec == 0 && usec == 0 {
+                    None
+                } else {
+                    let sec = u64::try_from(sec).unwrap();
+                    let nsec = u32::try_from(usec).unwrap().strict_mul(1000);
```
RalfJung: This sounds like the user program could actually make it panic?

Probably best to have a `read_timeval` method similar to the existing `read_timespec` and have it indicate when the value is invalid (like usec >= 1000000). The existing helper is in helpers.rs but shims/time.rs would be a better location.

RalfJung (commented):
The deadline logic looks good now. :)
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5042/55962eb26a4733e5db315da892df18c142d008cf..91775221f1a94be2e1636b45ef72c7dd4e0c8d4c)*

src/shims/unix/socket.rs
WhySoBad: I've looked it up in the `libc` for Apple, FreeBSD, Linux, Android, Solarish, and NetBSD, and all of them have the same value for EAGAIN and EWOULDBLOCK. If you want I can change it to EAGAIN here.

tests/pass-dep/libc/libc-socket.rs
WhySoBad: Yes you're right. It would've been beneficial when I also changed the timeout to 50ms instead of 10ms...

https://github.com/rust-lang/miri/blob/91775221f1a94be2e1636b45ef72c7dd4e0c8d4c/tests/pass-dep/libc/libc-socket.rs#L784-L785

When changed to 50ms it also works on a native host.

src/shims/unix/socket.rs
WhySoBad: No, the docs don't specify this but it's used here in the kernel: https://github.com/torvalds/linux/blob/HEAD/net/ipv4/af_inet.c#L701-L710

src/shims/unix/socket.rs
WhySoBad: I've now moved both into `shims/time.rs`.

src/shims/unix/socket.rs
RalfJung: In that case the choice does not matter, we should just be consistent. It looks like in our backward mapping we prefer EWOULDBLOCK so let's also use it here then.

tests/pass-dep/libc/libc-socket.rs
RalfJung: > When changed to 50ms it also works on a native host.

Wait, I thought you said above the write timeout just gets ignored in native execution for full buffers...?

src/shims/unix/socket.rs
RalfJung: Okay, please add a link then.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -806,6 +806,14 @@ fn test_sockopt_sndtimeo() {
...
         }
     }
+
+    // The write should return an EAGAIN/EWOULDBLOCK after ~50ms instead of blocking indefinitely.
+    let before = Instant::now();
+    let err = unsafe {
+        errno_result(libc::write(client_sockfd, buffer.as_ptr().cast(), buffer.len())).unwrap_err()
+    };
+    assert_eq!(err.kind(), ErrorKind::WouldBlock);
+    assert!(Instant::now().duration_since(before) >= Duration::from_millis(50));
```
RalfJung: Instead of blocking a 2nd time, can you measure the time inside the loop?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5042/55962eb26a4733e5db315da892df18c142d008cf..63bfdebc12a73ca633f026a1d4daec1f35c6462c)*

tests/pass-dep/libc/libc-socket.rs
RalfJung: Oh I see, the previous test was wrong.

src/clock.rs
```diff
@@ -11,12 +11,12 @@ use crate::MiriMachine;
...
 
 /// An instant (a fixed moment in time) in Miri's monotone clock.
-#[derive(Debug)]
+#[derive(Copy, Clone, Debug)]
 pub struct Instant {
     kind: InstantKind,
 }
 
-#[derive(Debug)]
+#[derive(Copy, Clone, Debug)]
```
RalfJung: Let's make this only Clone, not Copy

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5042/55962eb26a4733e5db315da892df18c142d008cf..63bfdebc12a73ca633f026a1d4daec1f35c6462c)*

WhySoBad:
The PR should be ready but the CI gets some 403 errors.

@rustbot ready

RalfJung:

@rustbot author

PR #5043, opened 2026-05-15
Fix readable readiness removed after a short peek

WhySoBad:
Hi,

This pull request fixes a regression introduced by rust-lang/miri#5023.
We wrongfully remove the readable readiness after a short peek. This has the consequence that repeated short peeks (or a short peek followed by a read) will block indefinitely. 

I've now added a test case for this and disabled the "short read special treatment" for short peeks.

### DIFF
--- a/src/shims/unix/socket.rs
+++ b/src/shims/unix/socket.rs
@@ -1697,12 +1697,12 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // would be returned on UNIX-like systems. We thus remap this error to an EWOULDBLOCK.
                 interp_ok(Err(IoError::HostError(io::ErrorKind::WouldBlock.into())))
             }
-            Ok(bytes_read) if bytes_read < length && bytes_read > 0 => {
-                // We had a short read. (Note that reading 0 bytes is guaranteed to indicate EOF,
-                // and can never happen spuriously, so we have to exclude that case.) On Unix hosts
-                // using the `epoll` and `kqueue` backends, a short read means that the read buffer
-                // is empty. We update the readiness accordingly, which means that next time we see
-                // "readable" we will report an epoll edge. Some applications (e.g. tokio) rely on
+            Ok(bytes_read) if !should_peek && bytes_read < length && bytes_read > 0 => {
+                // We had a short read (and were not peeking). (Note that reading 0 bytes is guaranteed
+                // to indicate EOF, and can never happen spuriously, so we have to exclude that case.)
+                // On Unix hosts using the `epoll` and `kqueue` backends, a short read means that the
+                // read buffer is empty. We update the readiness accordingly, which means that next time
+                // we see "readable" we will report an epoll edge. Some applications (e.g. tokio) rely on
                 // this behavior; see
                 // <https://github.com/tokio-rs/tokio/blob/HEAD/tokio/src/io/poll_evented.rs#L190-L210>
                 if cfg!(any(
--- a/tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
+++ b/tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
@@ -25,6 +25,7 @@ fn main() {
     test_shutdown_read();
     test_shutdown_write();
     test_readiness_after_short_read();
+    test_readiness_after_short_peek();
     test_readiness_after_short_write();
 }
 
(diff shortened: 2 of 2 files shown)


tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -505,6 +506,54 @@ fn test_readiness_after_short_read() {
...
+    }
+
+    // Write some bytes into the peer socket.
+    libc_utils::write_all(peerfd, TEST_BYTES).unwrap();
+
+    // Add client socket with readable interest to epoll.
+    epoll_ctl_add(epfd, client_sockfd, EPOLLET | EPOLLIN).unwrap();
+
+    // Wait until the socket becomes readable.
+    check_epoll_wait::<8>(epfd, &[Ev { events: EPOLLIN, data: client_sockfd }], -1);
```
RalfJung: There's only one thread in this test so waiting seems rather pointless?

src/shims/unix/socket.rs
```diff
@@ -1697,12 +1697,12 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // would be returned on UNIX-like systems. We thus remap this error to an EWOULDBLOCK.
                 interp_ok(Err(IoError::HostError(io::ErrorKind::WouldBlock.into())))
             }
-            Ok(bytes_read) if bytes_read < length && bytes_read > 0 => {
-                // We had a short read. (Note that reading 0 bytes is guaranteed to indicate EOF,
-                // and can never happen spuriously, so we have to exclude that case.) On Unix hosts
-                // using the `epoll` and `kqueue` backends, a short read means that the read buffer
-                // is empty. We update the readiness accordingly, which means that next time we see
-                // "readable" we will report an epoll edge. Some applications (e.g. tokio) rely on
+            Ok(bytes_read) if bytes_read < length && bytes_read > 0 && !should_peek => {
```
```suggestion
            Ok(bytes_read) if !should_peek && bytes_read < length && bytes_read > 0 => {
```

RalfJung (commented):
Good catch!
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5043/48b0b923635281fad0c7597c7c704a7d7fe10cf2..6082e268627633bed74a796823b34c2ec5a8e929)*

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -505,6 +506,51 @@ fn test_readiness_after_short_read() {
...
+    unsafe {
+        // Change client socket to be non-blocking.
+        errno_check(libc::fcntl(client_sockfd, libc::F_SETFL, libc::O_NONBLOCK));
+    }
+
+    // Write some bytes into the peer socket.
+    libc_utils::write_all(peerfd, TEST_BYTES).unwrap();
+
+    // Add client socket with readable interest to epoll.
+    epoll_ctl_add(epfd, client_sockfd, EPOLLET | EPOLLIN).unwrap();
```
RalfJung: epfd is now never actually used, is it?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5043/48b0b923635281fad0c7597c7c704a7d7fe10cf2..b3e464d2467457f75435d3be88ed55f08a6ce2ff)*

WhySoBad: Yes, with the current implementation it would be useless. However, I've noticed that we need to have the `epoll_wait` because otherwise we won't enter the scheduler between the `write_all` and the `recv`. This has the consequence that the readable readiness is not set because the host I/O event was not processed (the reason why the CI currently fails). 

This does not reflect actual linux behavior. In my opinion we can only solve this by forcefully entering the scheduler and processing the pending host I/O events before any read/write by having a zero-timeout block.

RalfJung: It sounds more like we need to go into the scheduler briefly *after* every read/write as that may have made some other socket ready? OTOH for being able to reason about which thread runs when it'd be kind of annoying if every read/write is a "yield". So the other option is to regularly do `poll` in the scheduler even if the active thread did not yield.

For now, I'd say put a FIXME here and have the test call `thread::yield_now` instead of `epoll_wait` so that we go into the scheduler.

WhySoBad: > So the other option is to regularly do poll in the scheduler even if the active thread did not yield.

This sounds as if it might also not work. We can't just regularly `poll` because we need to be ensured that we `poll` directly after a read/write/etc. We could introduce a new boolean flag somewhere which indicates that the scheduler should do a `poll` without a yielding thread. This could then be set after a read/write/etc.

RalfJung: I don't quite follow, but let's move that discussion elsewhere. Can you open an issue?

WhySoBad: I've opened https://github.com/rust-lang/miri/issues/5047

RalfJung:
Thanks! I have added a little comment tweak, looks good otherwise. :)

PR #5044, opened 2026-05-15
refactor thread block timeout handling

RalfJung:


src/shims/unix/socket.rs
```diff
@@ -1780,16 +1779,13 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // If we should wait until the connection is established, the timeout is `None`.
         // Otherwise, we use a zero duration timeout, i.e. we return immediately
         // (but we still go through the scheduler once -- which is fine).
-        let timeout = if should_wait {
-            None
-        } else {
-            Some((TimeoutClock::Monotonic, TimeoutAnchor::Absolute, Duration::ZERO))
```
RalfJung, line 1786: @WhySoBad this PR has the refactor/docs I promised.

Turns out I missed an inconsequential mistake here when we landed this code -- this actually configured a negative-duration timeout, at "absolute" time 0 on the monotonic clock (rather than "relative" time 0).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5044/48b0b923635281fad0c7597c7c704a7d7fe10cf2..daf29140ddba293cc01d793a8ce8e153d5766143)*

PR #5046, opened 2026-05-16
libc shim fixes

RalfJung:
Fixes some problems I found while running these tests on the host: 
- eventfd writes with a larger-than-8-bytes buffer actually should fail with EINVAL, notwithstanding what the docs say
- an epoll-blocking test was blocking for `IN | OUT` but the socket is always `OUT`; we should block for just `IN`
- statx should always return the MODE
- mkstemp should not prefix the path with `/tmp`


eval-exec:
> Fixes some problems I found while running these tests on the host:

@RalfJung How did you execute test to find these?

RalfJung:
I used the new `./miri run --native`.

Note however that not all tests are expected to work that way, it takes a bit of looking at the failure to understand whether it is expected or a bug.

PR #5050, opened 2026-05-19
Support file locking via fcntl

### LINKED ISSUE #5038: File locking does not work on Solaris
We support `file.lock()`/`file.lock_shared()` on almost all targets, except Solaris. Would be nice to close this gap. :)

```
error: unsupported operation: fcntl: unsupported command 0x7
  --> /home/r/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/sys/fs/unix.rs:1509:22
   |
LL |         cvt(unsafe { libc::fcntl(self.as_raw_fd(), libc::F_SETLKW, &flock) })?;
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ unsupported operation occurred here
   |
   = note: stack backtrace:
           0: std::sys::fs::unix::File::lock_shared
               at /home/r/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/sys/fs/unix.rs:1509:22: 1509:75
           1: std::fs::File::lock_shared
               at /home/r/.rustup/toolchains/miri/lib/rustlib/src/rust/library/std/src/fs.rs:919:9: 919:33
           2: test_flock
               at tests/pass/shims/fs.rs:413:5: 413:24
           3: main
               at tests/pass/shims/fs.rs:41:9: 41:21
```

Cc @devnexen

Nandan-Tav27:
## Summary

On Solaris, `File::lock` / `lock_shared` / `try_lock` / `try_lock_shared` / `unlock` lower to `fcntl(F_SETLK)` or `fcntl(F_SETLKW)` rather than `flock(2)`. Miri's fcntl shim didn't handle these commands and any program calling the methods aborted with `unsupported command 0x6` or `0x7`.

This PR adds the two commands to the fcntl shim and reuses miri's existing flock backend internally. A check has been added to reject range locks as `flock` only supports whole-file locking and the std call from Solaris would request a whole-file lock.

## Tests

- `tests/pass/shims/fs.rs::test_flock` - This is the previously failing test case mentioned in the issue. The Solaris gate has been dropped and the test passes now.
- `tests/fail-dep/libc/fs/fcntl_setlk_range.rs` - A new fail test case to ensure range locks are rejected
- `tests/pass-dep/libc/fcntl_setlk.rs` - A direct fcntl locking test. Needed for Unix-based platforms other than Solaris, where lock calls in test_flock would lower to flock(2) and the new fcntl path would not be used.

Verified locally with:

- `./miri test --target x86_64-pc-solaris`
- `./miri test --target x86_64-unknown-linux-gnu`
- `./miri test --target aarch64-apple-darwin`
- `./miri test --target x86_64-unknown-illumos`

Closes rust-lang/miri#5038

cc: @devnexen

### DIFF
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -270,14 +270,14 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
                 this.ffullsync_fd(fd_num)
             }
-            cmd if this.tcx.sess.target.os == Os::Solaris
+            cmd if matches!(this.tcx.sess.target.os, Os::Solaris | Os::Illumos)
                 && (cmd == f_setlk || cmd == f_setlkw) =>
             {
                 let Some(fd) = this.machine.fds.get(fd_num) else {
-                    return this.set_last_error_and_return_i32(LibcError("EBADF"));
+                    return this.set_errno_and_return_neg1_i32(LibcError("EBADF"));
                 };
 
-                let [flock] = check_min_vararg_count("fcntl(fd, F_SETFLK*, ...)", varargs)?;
+                let [flock] = check_min_vararg_count("fcntl(fd, F_SETLK*, ...)", varargs)?;
                 let flock = this.deref_pointer_as(flock, this.libc_ty_layout("flock"))?;
                 let l_type =
                     this.read_scalar(&this.project_field_named(&flock, "l_type")?)?.to_i16()?;
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -274,10 +274,10 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     this.read_scalar(&this.project_field_named(&flock, "l_type")?)?.to_i16()?;
                 let l_whence =
                     this.read_scalar(&this.project_field_named(&flock, "l_whence")?)?.to_i16()?;
-                let l_start =
-                    this.read_scalar(&this.project_field_named(&flock, "l_start")?)?.to_i64()?;
-                let l_len =
-                    this.read_scalar(&this.project_field_named(&flock, "l_len")?)?.to_i64()?;
+                let l_start = this.project_field_named(&flock, "l_start")?;
+                let l_start = this.read_scalar(&l_start)?.to_int(l_start.layout.size)?;
+                let l_len = this.project_field_named(&flock, "l_len")?;
+                let l_len = this.read_scalar(&l_len)?.to_int(l_len.layout.size)?;
 
                 // flock only supports whole-file locking
                 let seek_set = this.eval_libc_i32("SEEK_SET");
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -288,7 +288,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let l_len =
                     this.read_scalar(&this.project_field_named(&flock, "l_len")?)?.to_i64()?;
 
-                // We call flock, which only supports full file locking unlike fcntl
+                // We call flock, which only supports whole-file locking unlike fcntl
                 let seek_set = this.eval_libc_i32("SEEK_SET");
                 if i32::from(l_whence) != seek_set || l_start != 0 || l_len != 0 {
                     throw_unsup_format!(
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -259,20 +259,11 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
                 fd.set_flags(flag & !ignored_flags, this)
             }
-            cmd if this.tcx.sess.target.os == Os::MacOs
-                && cmd == this.eval_libc_i32("F_FULLFSYNC") =>
-            {
-                // Reject if isolation is enabled.
-                if let IsolatedOp::Reject(reject_with) = this.machine.isolated_op {
-                    this.reject_in_isolation("`fcntl`", reject_with)?;
-                    return this.set_errno_and_return_neg1_i32(ErrorKind::PermissionDenied);
-                }
+            cmd if cmd == f_setlk || cmd == f_setlkw => {
+                // fcntl POSIX locks (per-process) and flock (per open file descriptor) are distinct
+                // here we reuse miri's flock implementation
 
-                this.ffullsync_fd(fd_num)
-            }
-            cmd if matches!(this.tcx.sess.target.os, Os::Solaris | Os::Illumos)
-                && (cmd == f_setlk || cmd == f_setlkw) =>
-            {
+                // Check if this is a valid open file descriptor.
                 let Some(fd) = this.machine.fds.get(fd_num) else {
                     return this.set_errno_and_return_neg1_i32(LibcError("EBADF"));
                 };
@@ -288,7 +279,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let l_len =
                     this.read_scalar(&this.project_field_named(&flock, "l_len")?)?.to_i64()?;
 
-                // We call flock, which only supports whole-file locking unlike fcntl
+                // flock only supports whole-file locking
                 let seek_set = this.eval_libc_i32("SEEK_SET");
                 if i32::from(l_whence) != seek_set || l_start != 0 || l_len != 0 {
                     throw_unsup_format!(
@@ -296,13 +287,16 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     );
                 }
 
-                let f_rdlck = this.eval_libc("F_RDLCK").to_i16()?;
-                let f_wrlck = this.eval_libc("F_WRLCK").to_i16()?;
-                let f_unlck = this.eval_libc("F_UNLCK").to_i16()?;
+                // These constants vary in size depending upon the platform
+                let const_ty_size = this.eval_libc("F_RDLCK").to_scalar_int()?.size();
+                let f_rdlck = this.eval_libc("F_RDLCK").to_int(const_ty_size)?;
+                let f_wrlck = this.eval_libc("F_WRLCK").to_int(const_ty_size)?;
+                let f_unlck = this.eval_libc("F_UNLCK").to_int(const_ty_size)?;
                 // F_SETLK = non-blocking; F_SETLKW = blocking
                 let nonblocking = cmd == f_setlk;
 
                 use FlockOp::*;
+                let l_type = i128::from(l_type);
                 let op = if l_type == f_rdlck {
                     SharedLock { nonblocking }
                 } else if l_type == f_wrlck {
@@ -317,6 +311,17 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // return `0` if flock is successful
                 let result = result.map(|()| 0i32);
                 interp_ok(Scalar::from_i32(this.try_unwrap_io_result(result)?))
+            }
+            cmd if this.tcx.sess.target.os == Os::MacOs
+                && cmd == this.eval_libc_i32("F_FULLFSYNC") =>
+            {
+                // Reject if isolation is enabled.
+                if let IsolatedOp::Reject(reject_with) = this.machine.isolated_op {
+                    this.reject_in_isolation("`fcntl`", reject_with)?;
+                    return this.set_errno_and_return_neg1_i32(ErrorKind::PermissionDenied);
+                }
+
+                this.ffullsync_fd(fd_num)
             }
             cmd => {
                 throw_unsup_format!("fcntl: unsupported command {cmd:#x}");
--- a/src/shims/unix/fd.rs
+++ b/src/shims/unix/fd.rs
@@ -192,6 +192,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         let f_dupfd_cloexec = this.eval_libc_i32("F_DUPFD_CLOEXEC");
         let f_getfl = this.eval_libc_i32("F_GETFL");
         let f_setfl = this.eval_libc_i32("F_SETFL");
+        let f_setlk = this.eval_libc_i32("F_SETLK");
+        let f_setlkw = this.eval_libc_i32("F_SETLKW");
 
         // We only support getting the flags for a descriptor.
         match cmd {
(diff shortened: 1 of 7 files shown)


src/shims/unix/fd.rs
```diff
@@ -268,6 +270,54 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
                 this.ffullsync_fd(fd_num)
             }
+            cmd if this.tcx.sess.target.os == Os::Solaris
```
devnexen: while at it, would it be possible to enable the code paths (and tests) for illumos too ?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5050/159054c60e08465357ab954745fc27318fad55bf..d8d12ef9eb475b362d40ef2e4eaae40c056d45f0)*

Nandan-Tav27: Have enabled the code path for illumos and added an explicit test since `tests/pass/shims/fs.rs::test_flock` would lower to flock when run with the target as illumos

Nandan-Tav27:
Rebased across master to pick up the rename `set_last_error_and_return_i32` → `set_errno_and_return_neg1_i32`

src/shims/unix/fd.rs
```diff
@@ -268,6 +270,54 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
                 this.ffullsync_fd(fd_num)
             }
+            cmd if matches!(this.tcx.sess.target.os, Os::Solaris | Os::Illumos)
+                && (cmd == f_setlk || cmd == f_setlkw) =>
```
RalfJung: Is there a specific reason to support these only on Solarish? Do they behave differently across Unixes?

tests/pass-dep/libc/fcntl_setlk.rs
```diff
@@ -0,0 +1,44 @@
...
+    let file1 = OpenOptions::new().read(true).write(true).open(&path).unwrap();
+    let file2 = OpenOptions::new().read(true).write(true).open(&path).unwrap();
+    let fd1 = file1.as_raw_fd();
+    let fd2 = file2.as_raw_fd();
+
+    let wrlck = make_flock(libc::F_WRLCK as libc::c_short);
+    let rdlck = make_flock(libc::F_RDLCK as libc::c_short);
+    let unlck = make_flock(libc::F_UNLCK as libc::c_short);
+
+    assert_eq!(unsafe { libc::fcntl(fd1, libc::F_SETLK, &wrlck) }, 0);
```
RalfJung: Please use `errno_check` to invoke `fcntl`, that will mostly do the error handling for you.

tests/pass-dep/libc/fcntl_setlk.rs
```diff
@@ -0,0 +1,44 @@
...
+        assert_eq!(Error::last_os_error().raw_os_error(), Some(libc::EAGAIN));
+    }
+
+    assert_eq!(unsafe { libc::fcntl(fd1, libc::F_SETLK, &unlck) }, 0);
+
+    assert_eq!(unsafe { libc::fcntl(fd1, libc::F_SETLKW, &rdlck) }, 0);
+    assert_eq!(unsafe { libc::fcntl(fd2, libc::F_SETLKW, &rdlck) }, 0);
+
+    assert_eq!(unsafe { libc::fcntl(fd1, libc::F_SETLK, &unlck) }, 0);
+    assert_eq!(unsafe { libc::fcntl(fd2, libc::F_SETLK, &unlck) }, 0);
```
RalfJung: I guess unlocking once more should fail? That seems worth testing.

RalfJung (commented):
This looks good, thanks. :)

However, I wonder -- is the fcntl API not portable across all Unixes? If it is then we might as well provide it everywhere, even if std does not use it.

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5050/45035df166f93b6b3576e597651efebcd9cfb9d0..f160093774608ba4a7b1b19f3fe464330e4e29a8)*

src/shims/unix/fd.rs
Nandan-Tav27: The behaviour will be the same, I have dropped the gate. My only concern was that someone could mistake the un-gated implementation for a fully functioning fcntl implementation and try using it with range-locks (we do throw an error in this case). If you think that could be a potential issue as well I can perhaps extend the implementation to support that in a future PR.

tests/pass-dep/libc/fcntl_setlk.rs
Nandan-Tav27: I checked the behaviour and unlocking once more should ideally pass, shall I add a test to ensure that?

src/shims/unix/fd.rs
RalfJung: As long as every unsupported case does `throw_unsup_format!`, we are good.

tests/pass-dep/libc/fcntl_setlk.rs
RalfJung: Yes please extend the test to cover unlocking more than once, and also to cover acquiring more than one lock on the same FD.

src/shims/unix/fd.rs
```diff
@@ -257,6 +259,59 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
 
                 fd.set_flags(flag & !ignored_flags, this)
             }
+            cmd if cmd == f_setlk || cmd == f_setlkw => {
+                // fcntl POSIX locks (per-process) and flock (per open file descriptor) are distinct
+                // here we reuse miri's flock implementation
```
RalfJung, line 264: Is this one sentence or two sentences? There's no indication of a sentence ending so it should be one but then I have trouble parsing it.

Also, doesn't the comment essentially say that the implementation is wrong because the locks should be distinct but we use the same lock for both?

src/shims/unix/fd.rs
```diff
@@ -257,6 +259,59 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let l_type =
+                    this.read_scalar(&this.project_field_named(&flock, "l_type")?)?.to_i16()?;
+                let l_whence =
+                    this.read_scalar(&this.project_field_named(&flock, "l_whence")?)?.to_i16()?;
+                let l_start = this.project_field_named(&flock, "l_start")?;
+                let l_start = this.read_scalar(&l_start)?.to_int(l_start.layout.size)?;
+                let l_len = this.project_field_named(&flock, "l_len")?;
+                let l_len = this.read_scalar(&l_len)?.to_int(l_len.layout.size)?;
+
+                // flock only supports whole-file locking
```
RalfJung, line 282:
```suggestion
                // flock only supports whole-file locking.
```
Comments generally should be proper sentences. That makes them easier to read.

src/shims/unix/fd.rs
```diff
@@ -257,6 +259,59 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let seek_set = this.eval_libc_i32("SEEK_SET");
+                if i32::from(l_whence) != seek_set || l_start != 0 || l_len != 0 {
+                    throw_unsup_format!(
+                        "fcntl: range locks are not supported (only whole-file: l_whence=SEEK_SET, l_start=0, l_len=0)"
+                    );
+                }
+
+                // These constants vary in size depending upon the platform
+                let const_ty_size = this.eval_libc("F_RDLCK").to_scalar_int()?.size();
```
RalfJung, line 291:
```suggestion
                // These constants vary in size depending upon the platform.
                // There also is no integer type alias so we have to get the size
                // from one of the constants.
                let fcntl_flag_size = this.eval_libc("F_RDLCK").to_scalar_int()?.size();
```

tests/pass-dep/libc/fcntl_setlk.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    // Test re-acquiring and converting the lock on the same FD
+    // This does not behave correctly on Windows hosts, see #miri/5074
+    if !cfg!(windows_host) {
+        // Re-acquiring the same lock on the same FD should succeed
+        errno_check(unsafe { libc::fcntl(fd1, libc::F_SETLK, &wrlck) });
+        // Downgrading to a read lock on the same FD should also succeed
+        errno_check(unsafe { libc::fcntl(fd1, libc::F_SETLK, &rdlck) });
+
+        errno_check(unsafe { libc::fcntl(fd1, libc::F_SETLK, &wrlck) });
```
RalfJung, line 46:
```suggestion
        // And upgrading it back to a write lock should also succeed.
        errno_check(unsafe { libc::fcntl(fd1, libc::F_SETLK, &wrlck) });
```

tests/pass-dep/libc/fcntl_setlk.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    // Attempting to take a second exclusive lock from a different fd should fail
+    unsafe {
+        let err = errno_result(libc::fcntl(fd2, libc::F_SETLK, &wrlck)).unwrap_err();
+        assert_eq!(err.raw_os_error(), Some(libc::EAGAIN));
+    }
+
+    errno_check(unsafe { libc::fcntl(fd1, libc::F_SETLK, &unlck) });
+
+    errno_check(unsafe { libc::fcntl(fd1, libc::F_SETLKW, &rdlck) });
```
RalfJung, line 57:
```suggestion
    // Getting two read locks should succeed.
    errno_check(unsafe { libc::fcntl(fd1, libc::F_SETLKW, &rdlck) });
```

tests/pass-dep/libc/fcntl_setlk.rs
```diff
@@ -0,0 +1,68 @@
...
+        errno_check(unsafe { libc::fcntl(fd1, libc::F_SETLK, &wrlck) });
+        // Downgrading to a read lock on the same FD should also succeed
+        errno_check(unsafe { libc::fcntl(fd1, libc::F_SETLK, &rdlck) });
+
+        errno_check(unsafe { libc::fcntl(fd1, libc::F_SETLK, &wrlck) });
+    }
+
+    // Attempting to take a second exclusive lock from a different fd should fail
+    unsafe {
+        let err = errno_result(libc::fcntl(fd2, libc::F_SETLK, &wrlck)).unwrap_err();
```
RalfJung, line 51: When I run this test on my Linux machine, this assertion fails.
```
panicked at tests/pass-dep/libc/fcntl_setlk.rs:51:73:
called `Result::unwrap_err()` on an `Ok` value: 0
```

RalfJung (commented):
Thanks!
That comment in the implementation however makes me wonder if this is even correct. We don't have a test for trying to acquire both an flock and a fnctl lock on the same file (via the same or different FDs)... and IIUC that would behave incorrectly?

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5050/b537f496edf90808f44b6d9e67e3238c6aa09f62..1cb1b09014f2e0b888c7343b78f1b1462ff3df8f)*

tests/pass-dep/libc/fcntl_setlk.rs
```diff
@@ -0,0 +1,68 @@
...
+    errno_check(unsafe { libc::fcntl(fd1, libc::F_SETLK, &unlck) });
+
+    errno_check(unsafe { libc::fcntl(fd1, libc::F_SETLKW, &rdlck) });
+    errno_check(unsafe { libc::fcntl(fd2, libc::F_SETLKW, &rdlck) });
+
+    errno_check(unsafe { libc::fcntl(fd1, libc::F_SETLK, &unlck) });
+    errno_check(unsafe { libc::fcntl(fd2, libc::F_SETLK, &unlck) });
+
+    // Redundant unlocks should succeed
+    // This does not behave correctly on Windows hosts, see #miri/5074
```
RalfJung, line 64:
```suggestion
    // Redundant unlocks should succeed.
    // FIXME(#miri/5074): This does not behave correctly on Windows hosts.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5050/b537f496edf90808f44b6d9e67e3238c6aa09f62..1cb1b09014f2e0b888c7343b78f1b1462ff3df8f)*

tests/pass-dep/libc/fcntl_setlk.rs
```diff
@@ -0,0 +1,68 @@
...
+    let fd2 = file2.as_raw_fd();
+
+    let wrlck = make_flock(libc::F_WRLCK as libc::c_short);
+    let rdlck = make_flock(libc::F_RDLCK as libc::c_short);
+    let unlck = make_flock(libc::F_UNLCK as libc::c_short);
+
+    errno_check(unsafe { libc::fcntl(fd1, libc::F_SETLK, &wrlck) });
+
+    // Test re-acquiring and converting the lock on the same FD
+    // This does not behave correctly on Windows hosts, see #miri/5074
```
RalfJung, line 39:
```suggestion
    // Test re-acquiring and converting the lock on the same FD.
    // FIXME(#miri/5074): This does not behave correctly on Windows hosts.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5050/b537f496edf90808f44b6d9e67e3238c6aa09f62..1cb1b09014f2e0b888c7343b78f1b1462ff3df8f)*

tests/pass-dep/libc/fcntl_setlk.rs
RalfJung: I dug a bit into this. This points to the heart of the problem I think: these locks are associated with the *process*, not the file handle. That's a completely ridiculous semantics and probably not possible for us to support.

How do these locks behave on Solaris? Are they also process-associated there? The [docs](https://docs.oracle.com/cd/E19109-01/tsolaris8/835-8003/6ruu1b0n7/index.html) seem to indicate that this is the case. If so then I don't understand why std uses this API, it seems to go entirely against the documented API contract of `File::lock`.
I filed an issue for that: https://github.com/rust-lang/rust/issues/157390

Nandan-Tav27: Yes, they are process-associated on Solaris as well. I don't have a Solaris machine however did check [a few Oracle blogs](https://blogs.oracle.com/solaris/python-exclusive-file-locking-on-solaris#:~:text=On%20Solaris%2C%20a%20file%20can%20be%20locked%20for%20exclusive%20access%20by%20any%20process%20with%20the%20help%20of%20fcntl()%20system%20call), in addtion to the `fcntl` man pages before starting with this to understand the behaviour correctly.

src/shims/unix/fd.rs
Nandan-Tav27: Sorry for the confusion, I will work on making my comments cleaner and more readable.

Yes, they should ideally be two distinct lock types. Since std lowers `File::lock` to `fcntl` (passing the parameters necessary for a whole-file lock), I thought of re-using the `flock` backend rather than fully implementing `fcntl`-like locking functionality. As you have mentioned in your other comment, the two differ substantially in their guarantees. 
Thus, if the `fcntl` shim is provided primarily for the case when `File::lock` lowers to it (like on Solaris) we could maybe have something like the above. However, if the behaviour should replicate `fcntl` functionality thoroughly then I could work on modifying the implementation accordingly. Do let me know how you think I should proceed!

RalfJung: I think for now we should put this on hold and see how https://github.com/rust-lang/rust/issues/157390 is resolved. Sorry for not realizing this earlier, I assumed `File::lock` was implemented correctly and consistently across platforms and didn't catch that `fcntl` has very strange and different semantics.

>  if the fcntl shim is provided primarily for the case when File::lock lowers to it

We can't know what people use the fcntl for once we implement it. And even for use with File::lock, it would be quite strange if it behaves different in Miri than on the real thing.

Nandan-Tav27: Sure, I should have added the comment highlighting the discrepancy in the two lock types earlier, so apologies for that. Will keep an eye on  https://github.com/rust-lang/rust/issues/157390.

RalfJung:
Closing as the Solaris implementation is being removed from std: https://github.com/rust-lang/rust/pull/157509.
Sorry for not realizing this issue earlier!

PR #5051, opened 2026-05-19
Optionally pass `--config` down to cargo when building the sysroot

erickt:
This allows us to do things like point cargo at a vendored directory to support offline builds of the sysroot.

Note: this is blocked on https://github.com/RalfJung/rustc-build-sysroot/pull/34 being landed and released in a build.

### DIFF
--- a/cargo-miri/src/setup.rs
+++ b/cargo-miri/src/setup.rs
@@ -161,14 +161,19 @@ pub fn setup(
     };
 
     // Do the build.
-    let status = SysrootBuilder::new(&sysroot_dir, target)
+    let mut builder = SysrootBuilder::new(&sysroot_dir, target)
         .build_mode(BuildMode::Build) // not a real build, since we use dummy codegen
         .rustc_version(rustc_version.clone())
         .sysroot_config(sysroot_config)
         .rustflags(rustflags)
         .cargo(cargo_cmd)
-        .when_build_required(notify)
-        .build_from_source(&rust_src);
+        .when_build_required(notify);
+
+    for cargo_config in get_arg_flag_values("--config") {
+        builder = builder.cargo_config(cargo_config);
+    }
+
+    let status = builder.build_from_source(&rust_src);
     match status {
         Ok(SysrootStatus::AlreadyCached) =>
             if !quiet && show_setup {


RalfJung:
It looks like rust-src since recently contains the vendored dependencies of std, so no optional config should be needed for this? See https://github.com/rust-lang/rust/pull/156196.

erickt:
Oh interesting! I’ll test it out with the latest rust-src and get back to you.

erickt:
Right now it looks like miri isn't able to find the vendored sources automatically if I block network access:

```
% unshare -r -n -- rustup run nightly cargo run -- miri setup --verbose
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/cargo-miri miri setup --verbose`
Preparing a sysroot for Miri (target: x86_64-unknown-linux-gnu) in .../.cache/miri...
warning: spurious network error (3 tries remaining): [6] Could not resolve hostname (Could not resolve host: static.crates.io)
...
```

That's also the case if I explicitly point it at the rust-src:

```
env MIRI_LIB_SRC=$HOME/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library unshare -r -n -- rustup run nightly cargo run -- miri setup --verbose --frozen
warning: spurious network error (3 tries remaining): [6] Could not resolve hostname (Could not resolve host: static.crates.io)
...
```

However if I use this patch and my patch to `rustc-build-sysroot` it works:

```
% unshare -r -n -- rustup run nightly cargo run -- miri setup --verbose --frozen --config ~/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/.cargo/config.toml
   Compiling compiler_builtins v0.1.160 (.../.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/compiler-builtins/compiler-builtins)
   Compiling core v0.0.0 (.../.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core)
...
```

Would you rather we hardcode `cargo-miri` or `rustc-build-sysroot` to always point the config at `$MIRI_LIB_SRC/.cargo/config.toml`? I'm not sure if that'll work with a normal rust checkout though, since it doesn't have a `.cargo/config.toml`.

RalfJung:
We could make it so that if the `.cargo/config.toml` exists, we use it -- would that suffice for your usecase? That would avoid having to add new flags.

erickt:
@RalfJung Yeah I think that would work. We may also be able to just set `$CARGO_HOME` to `$MIRI_LIB_SRC`? Otherwise we could explicitly pass down the config file with https://github.com/RalfJung/rustc-build-sysroot/pull/34.

RalfJung:
@rustbot author
based on the feedback above

PR #5054, opened 2026-05-20
Fix deallocating temporary buffer on failed `readv`

WhySoBad:
This pull request fixes a code path where the temporary read buffer for the `readv` implementation wouldn't be deallocated.
I've now also added tests for `readv`/`writev` returning an error which should ensure that this can't produce memory leaks in the future.


RalfJung:
Good catch!

Issue #4982, opened 2026-04-26
Use directory handles to implement openat and friends


RalfJung:
Miri currently does not support `openat`, which necessitates [using the fallback impl](https://github.com/rust-lang/rust/blob/7a47964d3f2826f5ea3070656b3b8a3a72b38e1a/library/std/src/sys/fs/unix.rs#L2512-L2522) for `remove_dir_all`. That's not great. Now that directory handles are [coming to std](https://github.com/rust-lang/rust/issues/120426), we hopefully can implement enough extra shims so that we can run the real `remove_dir_all` implementation at least on Unix targets.

PR #5057, opened 2026-05-21
feat: add required syscalls for remove_dir_all

phraakture:
## summary

implemented three syscalls- openat(), unlinkat() and fdopendir(), that the standard library's remove_dir_all uses. kept the fd type similar to windows's DirHandle.  The only difference is unix one additionally overrides path(), which the Windows doesn't need as they are unix specific.

related issue rust-lang/miri#4982

Note: I haven't included any test yet. Any guidance on how these syscalls should be tested, or references to similar existing tests, would be really appreciated.

### DIFF
--- a/src/shims/unix/fs.rs
+++ b/src/shims/unix/fs.rs
@@ -4,7 +4,7 @@ use std::borrow::Cow;
 use std::ffi::OsString;
 use std::fs::{self, DirBuilder, File, FileType, OpenOptions, TryLockError};
 use std::io::{self, ErrorKind, Read, Seek, SeekFrom, Write};
-use std::path::{self, Path};
+use std::path::{self, Path, PathBuf};
 use std::time::SystemTime;
 
 use rustc_abi::Size;
@@ -13,7 +13,7 @@ use rustc_data_structures::fx::FxHashMap;
 use rustc_target::spec::Os;
 
 use self::shims::time::system_time_to_duration;
-use crate::shims::files::FileHandle;
+use crate::shims::files::{DirHandle, FileHandle};
 use crate::shims::os_str::bytes_to_os_str;
 use crate::shims::sig::check_min_vararg_count;
 use crate::shims::unix::fd::{FlockOp, UnixFileDescription};
@@ -461,6 +461,12 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             }
         }
 
+        let o_directory = this.eval_libc_i32("O_DIRECTORY");
+        let is_directory = flag & o_directory == o_directory;
+        if is_directory {
+            flag &= !o_directory;
+        }
+
         let o_nofollow = this.eval_libc_i32("O_NOFOLLOW");
         if flag & o_nofollow == o_nofollow {
             flag &= !o_nofollow;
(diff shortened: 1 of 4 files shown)


RalfJung:
Thanks for giving this a try!

This week we Rust devs are all meeting at a conference so a backlog of PRs will pile up that'll take a bit to get through -- please be patient. However I can already tell you that the PR is missing tests. :) Specifically
- tests that invoke the std APIs
- and tests that directly invoke the libc functions

If you look around the test suite you can find examples for both, e.g. `tests/pass/shims/fs.rs` and `tests/pass-dep/libc/libc-fs.rs`. You can just add the new tests to those two files.

src/shims/unix/fs.rs
```diff
@@ -493,11 +499,234 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
         }
 
-        let fd = options
-            .open(path)
-            .map(|file| this.machine.fds.insert_new(FileHandle { file, writable }));
+        let fd = match options.open(&path) {
+            Ok(file) if is_directory =>
+                match file.metadata() {
+                    Ok(meta) if meta.is_dir() =>
+                        this.machine.fds.insert_new(DirHandle { path: path.to_path_buf() }),
```
RalfJung, line 506: Given that these APIs are meant specifically to avoid the risk of race conditions, I don't think we should implement them by just storing a path. Rather, a `DirHandle` should hold a [`fs::Dir`](https://doc.rust-lang.org/nightly/std/fs/struct.Dir.html) so that we can be actually sure that this refers to a stable directory.

RalfJung (commented):
Looking at the implementation, it seems you are trying to implement these APIs based on top of paths represented as strings. I don't think we should do that. Lucky enough the standard library is gaining support for directory handles so the hope is that we can use those. :)

It's bad enough that we have this hack on Windows, but you can't actually do much with a DirHandle there so it's not too big of a problem.

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5057/55962eb26a4733e5db315da892df18c142d008cf..f24df73295cffdaac6128f32e169080099571831)*

RalfJung:
So my suggestion would to instead add a new FD type that wraps `fs::Dir`, and restrict this PR to only those operations that can be implemented on top of `fs::Dir`. Making this PR smaller (fewer new shims at once) is anyway a good idea.

That means it's not yet enough to `remove_dir_all`. That's fine. If you want to make progress towards that, you can help by implementing more operations on `fs::Dir`. :)  (Note that there's work in progress for that, see https://github.com/rust-lang/rust/issues/120426.)

RalfJung:
Also, a note on terminology, since it also came up in another PR of course: the PR title talks about "sycalls" but that's not quite correct. What we are implementing here aren't syscalls, they are libc functions. These often but not always correspond to syscalls. If you say "syscall" in the context of Miri, then I will assume you are referring to supporting more cases in `libc::syscall`.

RalfJung:
@phraakture do you still plan to get back to this PR? It's fine if not, we should just know about it so someone else can then pick up the work. :)
As mentioned above, it would be better to start smaller, with just a single libc shim. Also the implementation of the shim in Miri should use std's directory handles.

phraakture:
> @phraakture do you still plan to get back to this PR? It's fine if not, we should just know about it so someone else can then pick up the work. :) As mentioned above, it would be better to start smaller, with just a single libc shim. Also the implementation of the shim in Miri should use std's directory handles.

Sorry for the delay. I do still plan to work on this PR.
If it turns out I can't continue for some reason, I'll let you know so someone else can pick it up.

RalfJung:
Okay, thanks for the update!

Note that `DirHandle` now exists in Miri. I would suggest the first PR to be very minimal and only add two things:
- "open" with `O_DIRECTORY` creates a `DirHandle`
- "openat" on a `DirHandle` can open a file

Ignore `remove_dir_all` for now and focus just on getting the basics of directory handles working. (Further directory handle functionality is anyway blocked on https://github.com/rust-lang/rust/pull/150679.)

PR #5059, opened 2026-05-23
Fix socket tests to run on native host

WhySoBad:
Hi,

Some socket tests weren't able to succeed on native Linux hosts. This pull request implements the needed changes to make them also work on a native host. 

It's mainly about some Linux-specific-ish behavior on how to treat the `option_len` parameter of `getsockopt` and `setsockopt` calls.

### DIFF
--- a/tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
+++ b/tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
@@ -294,10 +294,17 @@ fn test_send_nonblock() {
                 if written as usize == fill_buf.len() {
                     // When we didn't have a short write we should still be able to write more.
                     // Ensure the socket is still writable.
-                    assert_eq!(
-                        current_epoll_readiness::<8>(client_sockfd, EPOLLOUT | EPOLLET),
-                        EPOLLOUT
-                    );
+                    let readiness = current_epoll_readiness::<8>(client_sockfd, EPOLLOUT | EPOLLET);
+                    if cfg!(miri) {
+                        // With Miri we keep the writable readiness until EWOULDBLOCK is returned.
+                        assert_eq!(readiness, EPOLLOUT);
+                    } else {
+                        // On native Linux hosts, the writable readiness is removed when the buffer
+                        // is "almost" full. We can't emulate this with Miri.
+                        // The buffer must not be "almost" full at the first write.
+                        let is_not_first_write = total_written > fill_buf.len();
+                        assert!(readiness == EPOLLOUT || (is_not_first_write && readiness == 0));
+                    }
                 }
             }
             Err(err) if err.kind() == ErrorKind::WouldBlock => break,
--- a/tests/pass-dep/libc/libc-socket.rs
+++ b/tests/pass-dep/libc/libc-socket.rs
@@ -8,7 +8,7 @@ mod utils;
 
 use std::io::ErrorKind;
 use std::time::{Duration, Instant};
-use std::{ptr, thread};
+use std::{ptr, slice, thread};
 
 use libc_utils::*;
 
@@ -121,11 +121,26 @@ fn test_set_reuseaddr_invalid_len() {
     let sockfd =
         unsafe { errno_result(libc::socket(libc::AF_INET, libc::SOCK_STREAM, 0)).unwrap() };
     // Value should be of type `libc::c_int` which has size 4 bytes.
-    // By providing a u64 of size 8 bytes we trigger an invalid length error.
-    let err = net::setsockopt(sockfd, libc::SOL_SOCKET, libc::SO_REUSEADDR, 1u64).unwrap_err();
+    // By providing an u16 of size 2 bytes we trigger an invalid length error.
+    let err = net::setsockopt(sockfd, libc::SOL_SOCKET, libc::SO_REUSEADDR, 1u16).unwrap_err();
     assert_eq!(err.kind(), ErrorKind::InvalidInput);
     // Check that it is the right kind of `InvalidInput`.
     assert_eq!(err.raw_os_error(), Some(libc::EINVAL));
+
+    // By providing an u64 of size 8 bytes the behavior differs between native hosts and Miri.
+    let result = net::setsockopt(sockfd, libc::SOL_SOCKET, libc::SO_REUSEADDR, 1u64);
+    match result {
+        Err(err) => {
+            // Check that this is the right error.
+            assert_eq!(err.kind(), ErrorKind::InvalidInput);
+            assert_eq!(err.raw_os_error(), Some(libc::EINVAL));
+        }
+        Ok(_) => {
+            // Some native hosts just ignore too large inputs and only look at a prefix.
+            // On Miri we require an exact size.
+            assert!(!cfg!(miri));
+        }
+    }
 }
 
 #[cfg(any(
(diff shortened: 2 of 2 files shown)


src/shims/unix/socket_address.rs
```diff
@@ -156,8 +156,8 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // to extract address and port.
         let socket_addr = if family == this.eval_libc_i32("AF_INET").into() {
             let sockaddr_in_layout = this.libc_ty_layout("sockaddr_in");
-            if address_len != sockaddr_in_layout.size.bytes() {
-                // Address length should be exactly the length of an IPv4 address.
+            if address_len < sockaddr_in_layout.size.bytes() {
+                // Address length should be at least the length of an IPv4 address.
```
RalfJung: What do POSIX / non-Linux Unixes have to say about this?

src/shims/unix/socket.rs
```diff
@@ -1135,8 +1140,16 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                // Allocate new buffer on the stack. We use a `u8` layout when the value fits in
+                // one byte and `option_len` is nonzero and at most 4 bytes. Otherwise, a `u32`
+                // layout is used. This matches what Linux does:
+                // <https://github.com/torvalds/linux/blob/HEAD/net/ipv4/ip_sockglue.c#L1736-L1751>
+                let is_small_buffer = option_len > 0 && option_len < 4;
+                let value_buffer = if u8::try_from(ttl).is_ok() && is_small_buffer {
+                    this.allocate(this.machine.layouts.u8, MemoryKind::Stack)?
+                } else {
+                    this.allocate(this.machine.layouts.u32, MemoryKind::Stack)?
+                };
```
RalfJung: I don't think we want to copy whatever this Linux behavior is.^^

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -294,9 +294,13 @@ fn test_send_nonblock() {
...
-                        current_epoll_readiness::<8>(client_sockfd, EPOLLOUT | EPOLLET),
-                        EPOLLOUT
+                    let readiness = current_epoll_readiness::<8>(client_sockfd, EPOLLOUT | EPOLLET);
+                    // On native Linux hosts, the writable readiness can be removed whilst still some
+                    // writes succeed afterwards. We can't emulate this with Miri because we don't know
+                    // the buffer sizes and thus we just accept that the writable readiness might be
+                    // missing when we already wrote at least one `fill_buf`.
+                    assert!(
+                        readiness == EPOLLOUT || total_written > fill_buf.len() && readiness == 0
                     );
```
```suggestion
                    // On native Linux hosts, the writable readiness is removed when the buffer is "almost" full. We can't emulate this with Miri.
                    assert!(
                        readiness == EPOLLOUT || (!cfg!(miri) && total_written > fill_buf.len() && readiness == 0)
);
```

tests/utils/libc.rs
```diff
@@ -390,8 +390,9 @@ pub mod net {
                 &mut option_len,
             )
         })?;
-        // Ensure that there was no truncation.
-        assert!(option_len == provided_len);
+        // Ensure that there was no truncation because
+        // of the buffer being too small.
+        assert!(option_len <= provided_len);
```
RalfJung: This sounds wrong. If not all memory got initialized we should not continue.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -54,7 +54,7 @@ fn main() {
     test_shutdown_readable_after_write_close();
     test_shutdown_writable_after_read_close();
 
-    test_getsockopt_truncate();
+    // test_getsockopt_truncate();
```
RalfJung: Just skip this under `!cfg!(miri)` for now instead of disabling the test entirely.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -118,11 +118,14 @@ fn test_set_reuseaddr_invalid_len() {
...
-    // By providing a u64 of size 8 bytes we trigger an invalid length error.
-    let err = net::setsockopt(sockfd, libc::SOL_SOCKET, libc::SO_REUSEADDR, 1u64).unwrap_err();
+    // By providing an u16 of size 2 bytes we trigger an invalid length error.
+    let err = net::setsockopt(sockfd, libc::SOL_SOCKET, libc::SO_REUSEADDR, 1u16).unwrap_err();
     assert_eq!(err.kind(), ErrorKind::InvalidInput);
     // Check that it is the right kind of `InvalidInput`.
     assert_eq!(err.raw_os_error(), Some(libc::EINVAL));
+    // By providing an u64 of size 8 bytes no error should be returned because we
+    // only read the first four bytes.
+    net::setsockopt(sockfd, libc::SOL_SOCKET, libc::SO_REUSEADDR, 1u64).unwrap();
```
RalfJung: Miri should keep erroring here, at least for now.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -118,11 +118,14 @@ fn test_set_reuseaddr_invalid_len() {
     let sockfd =
         unsafe { errno_result(libc::socket(libc::AF_INET, libc::SOCK_STREAM, 0)).unwrap() };
     // Value should be of type `libc::c_int` which has size 4 bytes.
-    // By providing a u64 of size 8 bytes we trigger an invalid length error.
-    let err = net::setsockopt(sockfd, libc::SOL_SOCKET, libc::SO_REUSEADDR, 1u64).unwrap_err();
+    // By providing an u16 of size 2 bytes we trigger an invalid length error.
+    let err = net::setsockopt(sockfd, libc::SOL_SOCKET, libc::SO_REUSEADDR, 1u16).unwrap_err();
```
RalfJung, line 125: Let's test too-small in all cases and too-big under `cfg!(miri)`.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -175,14 +178,25 @@ fn test_bind_ipv4_invalid_addr_len() {
...
+            // Subtracting 1 from the address should make the size invalid.
+            (size_of::<libc::sockaddr_in>() - 1) as libc::socklen_t,
         ))
         .unwrap_err()
     };
     assert_eq!(err.kind(), ErrorKind::InvalidInput);
     // Check that it is the right kind of `InvalidInput`.
     assert_eq!(err.raw_os_error(), Some(libc::EINVAL));
+
+    // Adding to the address length shouldn't make the size invalid.
```
```suggestion
    // A too big size is not invalid.
```

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -175,14 +178,25 @@ fn test_bind_ipv4_invalid_addr_len() {
...
+    unsafe {
+        errno_result(libc::bind(
+            sockfd,
+            (&addr as *const libc::sockaddr_in).cast::<libc::sockaddr>(),
+            // Adding 1 to the address should not cause an error.
```
```suggestion
```

RalfJung (commented):
Linux truly has some "fun" behaviors 😂  Let's not copy all of them though...

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5059/55962eb26a4733e5db315da892df18c142d008cf..0a6523aa4f2bcddbafdd21d187ee20e1428ca398)*

src/shims/unix/socket_address.rs
WhySoBad: FreeBSD requires the address length to be exact e.g. in `udp_send` here: https://github.com/freebsd/freebsd-src/blob/HEAD/sys/netinet/udp_usrreq.c#L1139-L1140
And from what I can tell, MacOS also requires it to be equal: https://github.com/apple-oss-distributions/xnu/blob/HEAD/bsd/netinet/in_pcb.c#L867-L871

tests/pass-dep/libc/libc-socket.rs
WhySoBad: I've now changed this test to use the new `SO_RCVTIMEO`. This way it also works on a native Linux host (as long as `usec` is a multiple of `1000`).

src/shims/unix/socket_address.rs
RalfJung: Okay, then we should also require them to be equal.

tests/pass-dep/libc/libc-socket.rs
```diff
@@ -746,14 +746,20 @@ fn test_getsockopt_truncate() {
     let (sockfd, _) = net::make_listener_ipv4().unwrap();
 
     // Set the read timeout for the socket.
```
RalfJung, line 754:
```
// We use an integer number of milliseconds to ensure that the value
// round-trips properly on Linux (which apparently rounds to milliseconds).
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5059/cc0daa6770ee49b42a47f6883d6688407a5af9df..5aba29cc1d40e2fa18e32bc5efe2ff609a980651)*

RalfJung:
On my system Linux rounds the timeouts to a multiple of 4ms so I had to tweak the tests a bit more.

PR #5060, opened 2026-05-23
Perform host I/O poll before foreign item invocation

### LINKED ISSUE #5047: Socket readiness is only updated after a thread yielded
Currently, we only process pending host I/O events in the scheduler when a thread yielded:
https://github.com/rust-lang/miri/blob/1e193dc41c50121f36baa044053102d273d108da/src/concurrency/thread.rs#L689-L708

This leads to the problem that when reading/writing into a connected socket and it's peer from a single thread, the readable/writable readiness will not be updated correctly because the thread never yielded.

An example for this can be seen in the `test_readiness_after_short_peek` test in the `tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs` file.

As discussed in https://github.com/rust-lang/miri/pull/5043/changes/BASE..b3e464d2467457f75435d3be88ed55f08a6ce2ff#r3253001748, a solution for this would be to always have a zero-timeout block after each read/write to induce an artificial yield. However, this solution is suboptimal.
A better solution would be to have a boolean flag (e.g. `process_io_events`, probably on the blocking I/O manager) which can be read in the scheduler and when set to `true` we process host I/O events even when the active thread didn't yield.

WhySoBad:
Hi,

This pull request fixes that the host I/O readiness events are only processed when a thread yields. As discussed in the linked issue, we now also call `poll_and_unblock` before invoking any foreign item. By this, we ensure that the latest host I/O events are processed before any function is called which might rely on those updated events.

This also has the consequence that we no longer need to enter the scheduler for zero timeout blocks.

Closes rust-lang/miri#5047.

### DIFF
--- a/src/concurrency/blocking_io.rs
+++ b/src/concurrency/blocking_io.rs
@@ -154,15 +154,9 @@ impl BlockingIoManager {
     }
 
     /// Poll for new I/O events from the OS or wait until the timeout expired.
-    ///
-    /// - If the timeout is [`Some`] and contains [`Duration::ZERO`], the poll doesn't block and just
-    ///   reads all events since the last poll.
-    /// - If the timeout is [`Some`] and contains a non-zero duration, it blocks at most for the
-    ///   specified duration.
-    /// - If the timeout is [`None`] the poll blocks indefinitely until an event occurs.
-    ///
+    /// The timeout semantics are the same as described in [`Poll::poll`].
     /// The events also immediately get processed: threads get unblocked, and epoll readiness gets updated.
-    pub fn poll<'tcx>(
+    fn poll<'tcx>(
         ecx: &mut MiriInterpCx<'tcx>,
         timeout: Option<Duration>,
     ) -> InterpResult<'tcx, Result<(), io::Error>> {
@@ -344,6 +338,9 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
     /// readiness gets set for the source even when the requested interest
     /// might not be fulfilled.
     ///
+    /// The callback function will immediately be executed with [`UnblockKind::Ready`]
+    /// when `interest` is already fulfilled for `source_fd`.
+    ///
     /// There can also be spurious wake-ups by the OS and thus it's the callers
     /// responsibility to verify that the requested I/O interests are
     /// really ready and to block again if they're not.
@@ -372,6 +369,29 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
             // until the readiness is fulfilled and execute the callback then.
             this.block_thread(BlockReason::IO, deadline, callback);
             interp_ok(())
+        }
+    }
+
+    /// Poll for I/O events until either an I/O event happened or the timeout expired.
+    ///
+    /// - If the timeout is [`Some`] and contains [`Duration::ZERO`], the poll doesn't block and just
+    ///   reads all events since the last poll.
+    /// - If the timeout is [`Some`] and contains a non-zero duration, it blocks at most for the
+    ///   specified duration.
+    /// - If the timeout is [`None`] the poll blocks indefinitely until an event occurs.
+    ///
+    /// Unblocks all threads which are blocked on I/O and whose I/O interests
+    /// are currently fulfilled.
+    fn poll_and_unblock(&mut self, timeout: Option<Duration>) -> InterpResult<'tcx> {
+        let this = self.eval_context_mut();
+
+        match BlockingIoManager::poll(this, timeout)? {
+            Ok(_) => interp_ok(()),
+            // We can ignore errors originating from interrupts; that's just a spurious wakeup.
+            Err(e) if e.kind() == io::ErrorKind::Interrupted => interp_ok(()),
+            // For other errors we panic. On Linux and BSD hosts this should only be
+            // reachable when a system resource error (e.g. ENOMEM or ENOSPC) occurred.
+            Err(e) => panic!("unexpected error while polling: {e}"),
         }
     }
 }
--- a/src/concurrency/thread.rs
+++ b/src/concurrency/thread.rs
@@ -1,9 +1,9 @@
 //! Implements threads.
 
+use std::mem;
 use std::sync::atomic::Ordering::Relaxed;
 use std::task::Poll;
 use std::time::{Duration, SystemTime};
-use std::{io, mem};
 
 use rand::RngExt;
 use rand::seq::IteratorRandom;
@@ -698,8 +698,11 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
         // or timeouts to take care of.
 
         if this.machine.communicate() {
-            // When isolation is disabled we need to check for events for
-            // threads which are blocked on host I/O.
+            // When isolation is disabled we need to check for events for threads
+            // which are blocked on host I/O. Unlike the `poll_and_unblock` before
+            // any foreign item, the call here is needed to ensure that threads which
+            // are blocked on host I/O are woken up even if no shimmed functions are
+            // executed afterwards.
             // We do this before running any other threads such that the threads
             // which received events are available for scheduling afterwards.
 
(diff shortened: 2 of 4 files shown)


src/concurrency/blocking_io.rs
```diff
@@ -363,15 +361,51 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
         // We always have to do this since the thread will de-register itself.
         this.machine.blocking_io.add_blocked_thread(source_fd.id(), this.active_thread(), interest);
 
+        let is_deadline_exceeded = match deadline {
+            Some(Deadline::Monotonic(instant)) =>
+                instant.duration_since(this.machine.monotonic_clock.now()).is_zero(),
+            Some(Deadline::RealTime(time)) =>
+                time.duration_since(SystemTime::now()).unwrap_or(Duration::ZERO).is_zero(),
+            None => false,
+        };
```
RalfJung: I don't think we should do this. If we want to avoid blocking threads for some socket operations we should do that in the socket logic.

src/shims/unix/foreign_items.rs
```diff
@@ -108,6 +109,15 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     ) -> InterpResult<'tcx, EmulateItemResult> {
         let this = self.eval_context_mut();
 
+        if this.machine.communicate() {
+            // When isolation is disabled we need to check for new host I/O events before
+            // running any shimmed function. This is needed for keeping the internal I/O
+            // readiness of the FDs up-to-date.
```
RalfJung:
```suggestion
            // running any shimmed function. This is needed to ensure that the shim we
            // execute has up-to-date information about host readiness (as reflected e.g. by epoll) even if the current thread never yields.
```

src/clock.rs
```diff
@@ -11,12 +11,12 @@ use crate::MiriMachine;
...
 
 /// An instant (a fixed moment in time) in Miri's monotone clock.
-#[derive(Debug)]
+#[derive(Copy, Clone, Debug)]
 pub struct Instant {
     kind: InstantKind,
 }
 
-#[derive(Debug)]
+#[derive(Copy, Clone, Debug)]
```
RalfJung: Let's make this only `Clone`, not `Copy`

RalfJung (commented):
Just minor comments. :)

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5060/55962eb26a4733e5db315da892df18c142d008cf..69f0bbd67e864f4c4163e164728f6905a46a52c1)*

WhySoBad:
We said the `ensure_connected` call in `getpeername` could now be removed, but then we would still need a check like `matches!(&*socket.state, SocketState::Connected(_)) || matches!(&*socket.state, SocketState::Connecting(_) if socket.io_readiness.writable)`. This is basically a repetition of `ensure_connected` just without the `block_thread_for_io` part.

I'd thus now suggest that we store the callback of `ensure_connected` in a variable, and then should the socket be currently writable we directly invoke the callback and otherwise we block until we're writable. I think this way we have the lowest code duplication.

RalfJung:
Okay, let's do that for now and let's revisit this after the switch to socket2 in case that unlocks further simplicaition of the state machine handling.

src/concurrency/thread.rs
```diff
@@ -793,73 +796,49 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
...
-            // reachable when a system resource error (e.g. ENOMEM or ENOSPC) occurred.
-            Err(e) => panic!("unexpected error while polling: {e}"),
-        }
-    }
-
     /// Find all threads with expired timeouts, unblock them and execute their timeout callbacks.
     ///
     /// This method returns the minimum duration until the next thread deadline.
     /// If all ready threads have no deadline set, [`None`] is returned.
     fn unblock_expired_deadlines(&mut self) -> InterpResult<'tcx, Option<Duration>> {
```
RalfJung, line 803: Why are you changing this function?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5060/cc0daa6770ee49b42a47f6883d6688407a5af9df..b3cae0e0067f7e8b81e780d057e04fd859428081)*

src/shims/unix/socket.rs
```diff
@@ -1926,28 +1926,36 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-            BlockingIoInterest::Write,
-            deadline,
+        // We're currently connecting. The only way to determine whether we are done
+        // connecting is by checking that there were no errors during connecting after
+        // we received a writable readiness.
+
+        let is_already_writable = socket.io_readiness.borrow().writable;
+        let is_already_expired = deadline
+            .as_ref()
+            .is_some_and(|deadline| this.machine.duration_until_deadline(deadline).is_zero());
```
RalfJung: This is just re-computing information that the caller already has... can't we avoid that? I think we should just go through the normal `block_thread_for_io` codepath if `!is_already_writable`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5060/cc0daa6770ee49b42a47f6883d6688407a5af9df..b3cae0e0067f7e8b81e780d057e04fd859428081)*

src/shims/unix/socket.rs
```diff
@@ -1926,28 +1926,36 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                     foreign_name: &'static str,
+                    was_blocked: bool,
                     action: DynMachineCallback<'tcx, Result<(), ()>>,
                 } |this, kind: UnblockKind| {
-                    // Remove the blocking I/O interest for unblocking this thread.
-                    this.machine.blocking_io.remove_blocked_thread(socket.id(), this.machine.threads.active_thread());
+                    if was_blocked {
+                        // Remove the blocking I/O interest for unblocking this thread.
+                        this.machine.blocking_io.remove_blocked_thread(socket.id(), this.machine.threads.active_thread());
+                    }
```
RalfJung: This makes me reconsider the approach -- here you are basically mirroring the internal logic of `block_thread_for_io`. We should avoid that.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5060/cc0daa6770ee49b42a47f6883d6688407a5af9df..b3cae0e0067f7e8b81e780d057e04fd859428081)*

src/concurrency/thread.rs
WhySoBad: With the addition of the `duration_until_deadline` function, we no longer need to calculate the duration manually here. However, because `duration_until_deadline` has a reference onto `this.machine` and `this.machine.threads.threads.iter_enumerated_mut` had a mutable reference, this conflicted.
I've now just moved the stuff which needed mutable access to threads into the second loop such that we can use the new helper.

RalfJung: That sounds like the helper is doing more harm than good.

But I anyway don't think we should even need the helper, nothing outside this function should ever need `duration_until_deadline`.

WhySoBad:
I get the feeling that the special handling for `getpeername` might not be worth it as I don't really see how a nice implementation without duplicating some code of another function would look like.

RalfJung:
Given that we'll likely overhaul this when moving to `socket2`, let's do the minimal change here that keeps things working -- so let's just stick to the strategy where getpeername goes through the scheduler sometimes and get back to this later.

WhySoBad:
The latest force-push just undoes the commit which introduced the helper and added those special cases in `ensure_connected`.

@rustbot ready

src/concurrency/blocking_io.rs
```diff
@@ -344,6 +338,10 @@ pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
     /// readiness gets set for the source even when the requested interest
     /// might not be fulfilled.
     ///
+    /// The callback function will directly be executed when either `interest`
+    /// is already fulfilled for `source_fd` ([`UnblockKind::Ready`]) or when
+    /// `deadline` is [`Some`] and it isn't in the future ([`UnblockKind::TimedOut`]).
```
RalfJung: This does not seem accurate any more?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5060/cc0daa6770ee49b42a47f6883d6688407a5af9df..df15dc77d155ce2a939e7c234d527c255e4f5de8)*

RalfJung:

@rustbot author

Issue #3605, opened 2024-05-14
Support more ways of using mmap


RalfJung:
In a lengthy [Zulip discussion](https://rust-lang.zulipchat.com/#narrow/stream/136281-t-opsem/topic/Interaction.20of.20Strict.20Provenance.20and.20.60mmap.28.29.60), it was discovered that there are modes of use of mmap that are perfectly fine within the scope of Rust's memory model, but not supported by Miri's current implementation:
- do an initial big reservation with MAP_NORESERVE, letting the kernel pick a suitable memory range and reserve that address space. (Yes, MAP_NORESERVE still reserves the address space. Talk about confusing flag names...) This may overcommit if permitted by the kernel (that's what "noreserve" refers to), but the memory is now all read/write accessible.
- then later do smaller mmap in that range that actually "reserves" the memory (no more overcommit). This may set some flags to get huge tables if possible. It will (may?) also erase the previous contents of the re-mapped ranges, but doesn't change the range of memory that is read/write accessible, so it's fine with our current memory model.

See [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d2ea1888ffe60d05399f151af49bbf48) for some example code. Thanks to Nils for helping with the exploration here!

newpavlov:
I haven't read the discussion fully, but here unsupported cases which were encountered by me:
- Read-only mappings created using `PROT_READ` (i.e. without `PROT_WRITE`),
- Mappings populated with `MAP_POPULATE`,
- Reserving address space with `PROT_NONE`,
- Mappings with `MAP_FIXED` (to populate address range created with `PROT_NONE`).

saethlin:
I think at one point I started working on all of these then backed out. So here are some notes.

> * Mappings populated with `MAP_POPULATE`,

Is this useful without supporting file mappings? Adding file mappings would be a whole thing in itself; I don't think we can support `MAP_SHARED` so it would be an odd sort of thing to tell users about because most users in the ecosystem `MAP_SHARED` their files even if they only want to read from them.

> * Reserving address space with `PROT_NONE`,

I don't know what semantics people expect when a program tries to access `PROT_NONE` memory. Making the interpreter halt is probably the only viable option, because otherwise we'd have to... execute the segfault handler? I don't think it would be right to report UB here. Most likely `PROT_NONE` would have to be implemented in `before_memory_read`?

> * Mappings with `MAP_FIXED` (to populate address range created with `PROT_NONE`).

[...]

PR #5063, opened 2026-05-25
Minimal support for `mmap` protections, `mprotect`, `madvise`

foonathan:
This very partially address rust-lang/miri#3605.

A common patterns in memory allocators is to use `mmap` with `PROT_NONE` to reserve a big chunk of address space, and then selectively commit it by changing the protections with `mprotect` (and hinting how it will be used with `madvise`).

A minimal way to support is to ignore protections in `mmap`. This incorrectly allows programs that access memory in a way not consistent with the protections, but the kernel reliably rejects them anyway, so running it without miri will reliably discover this error. `mprotect` can then be trivially implemented as a no-op.

I also started a prototype implementation of properly modelling protections. The borrow trackers already have Permission tags which could be co-opted for that I think (I don't think the model cares whether a borrow is disabled due to Rust rules or whether the kernel disabled the memory). An `mmap` with protections other than read/write could then be implemented as a new allocation that sets the initial permission accordingly, and not blindly to Unique. `mprotect` is a bit more tricky due to the partial nature and because it essentially yanks out permissions from underneath previously valid references. My initial idea to model permission downgrades was to forcefully change everything in the stack/tree of items to the new permission, but it might break some assumptions the algorithm makes. So I've decided to just propose this minimum implementation and see whether there's interest in the first place :)

On the other hand, supporting `madvise` is trivial, as it is mostly advisory. There are flags that have observable effects on file mappings, but file mappings are not supported by miri to begin with. So the shim doesn't need to do anything - we could do more argument validation, but again, that's not really the error you need miri for.

### DIFF
--- a/src/shims/unix/foreign_items.rs
+++ b/src/shims/unix/foreign_items.rs
@@ -816,16 +816,44 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             }
 
             "mmap" => {
-                let [addr, length, prot, flags, fd, offset] =
-                    this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
+                let [addr, length, prot, flags, fd, offset] = this.check_shim_sig(
+                    shim_sig!(extern "C" fn(*mut _, usize, i32, i32, i32, libc::off_t) -> *mut _),
+                    link_name,
+                    abi,
+                    args,
+                )?;
                 let offset = this.read_scalar(offset)?.to_int(this.libc_ty_layout("off_t").size)?;
                 let ptr = this.mmap(addr, length, prot, flags, fd, offset)?;
                 this.write_scalar(ptr, dest)?;
             }
             "munmap" => {
-                let [addr, length] =
-                    this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
+                let [addr, length] = this.check_shim_sig(
+                    shim_sig!(extern "C" fn(*mut _, usize) -> i32),
+                    link_name,
+                    abi,
+                    args,
+                )?;
                 let result = this.munmap(addr, length)?;
+                this.write_scalar(result, dest)?;
+            }
+            "mprotect" => {
+                let [addr, length, prot] = this.check_shim_sig(
+                    shim_sig!(extern "C" fn(*mut _, usize, i32) -> i32),
+                    link_name,
+                    abi,
+                    args,
+                )?;
+                let result = this.mprotect(addr, length, prot)?;
+                this.write_scalar(result, dest)?;
+            }
+            "madvise" => {
+                let [addr, length, advice] = this.check_shim_sig(
+                    shim_sig!(extern "C" fn(*mut _, usize, i32) -> i32),
+                    link_name,
+                    abi,
+                    args,
+                )?;
+                let result = this.madvise(addr, length, advice)?;
                 this.write_scalar(result, dest)?;
             }
 
@@ -1386,7 +1414,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let [_, _] = this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
                 this.write_null(dest)?;
             }
-            "sigaction" | "mprotect" if this.frame_in_std() => {
+            "sigaction" if this.frame_in_std() => {
                 let [_, _, _] = this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
                 this.write_null(dest)?;
             }
--- a/src/shims/unix/mem.rs
+++ b/src/shims/unix/mem.rs
@@ -53,9 +53,6 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             return interp_ok(Scalar::from_maybe_pointer(Pointer::without_provenance(addr), this));
         }
 
-        let prot_read = this.eval_libc_i32("PROT_READ");
-        let prot_write = this.eval_libc_i32("PROT_WRITE");
-
         // First, we do some basic argument validation as required by mmap
         if (flags & (map_private | map_shared)).count_ones() != 1 {
             this.set_last_error(LibcError("EINVAL"))?;
@@ -80,13 +77,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             );
         }
 
-        // Miri doesn't support protections other than PROT_READ|PROT_WRITE.
-        if prot != prot_read | prot_write {
-            throw_unsup_format!(
-                "Miri does not support calls to mmap with protections other than \
-                 PROT_READ|PROT_WRITE",
-            );
-        }
+        verify_prot(this, prot)?;
 
         // Miri does not support shared mappings, or any of the other extensions that for example
         // Linux has added to the flags arguments.
@@ -103,14 +94,10 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         }
 
         let align = this.machine.page_align();
-        let Some(map_length) = length.checked_next_multiple_of(this.machine.page_size) else {
+        let Some(map_length) = round_up_to_page_size(this, length) else {
             this.set_last_error(LibcError("EINVAL"))?;
             return interp_ok(this.eval_libc("MAP_FAILED"));
         };
-        if map_length > this.target_usize_max() {
-            this.set_last_error(LibcError("EINVAL"))?;
-            return interp_ok(this.eval_libc("MAP_FAILED"));
-        }
 
         let ptr = this.allocate_ptr(
             Size::from_bytes(map_length),
@@ -135,13 +122,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             return this.set_errno_and_return_neg1_i32(LibcError("EINVAL"));
         }
 
-        let Some(length) = length.checked_next_multiple_of(this.machine.page_size) else {
+        let Some(length) = round_up_to_page_size(this, length) else {
             return this.set_errno_and_return_neg1_i32(LibcError("EINVAL"));
         };
-        if length > this.target_usize_max() {
-            this.set_last_error(LibcError("EINVAL"))?;
-            return interp_ok(this.eval_libc("MAP_FAILED"));
-        }
 
         let length = Size::from_bytes(length);
         this.deallocate_ptr(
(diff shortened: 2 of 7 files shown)


saethlin:
> A minimal way to support is to ignore protections in `mmap`. This incorrectly allows programs that access memory in a way not consistent with the protections, but the kernel reliably rejects them anyway, so running it without miri will reliably discover this error. `mprotect` can then be trivially implemented as a no-op.

I don't agree with this line of reasoning. Miri does a handful of things including having a weird scheduler and emulating weak memory effects that try to make the program encounter possible executions more easily than it would outside the interpreter. In addition, people use Miri's cross-interpreting support to test their code on other platforms, so it is definitely wrong to assume that such errors would be discovered by running without Miri.

> My initial idea to model permission downgrades was to forcefully change everything in the stack/tree of items to the new permission, but it might break some assumptions the algorithm makes

Just off the top of my head, how would you restore write permission to the correct tags if a page goes from `PROT_READ|PROT_WRITE` down to just `PROT_READ` then back up to `PROT_READ|PROT_WRITE`? I don't think you should be touching the aliasing model here _at all_, it would be better to store additional state about the permissions of mapped address ranges then check those in `before_memory_read`/`before_memory_write`.

> On the other hand, supporting `madvise` is trivial, as it is mostly advisory. There are flags that have observable effects on file mappings, but file mappings are not supported by miri to begin with. So the shim doesn't need to do anything - we could do more argument validation, but again, that's not really the error you need miri for.

That "mostly" is doing a lot of heavy lifting. `MADV_DONTNEED` on Linux changes the contents of memory so this implementation is wrong. Actually upon re-reading your comment in the code I think you have misunderstood the man page? `MADV_DONTNEED` documents that subsequent accesses to private anonymous mappings (the only kind we support) see zero-fill-on-demand pages.

RalfJung:
Thanks for the PR!

> A minimal way to support is to ignore protections in mmap. This incorrectly allows programs that access memory in a way not consistent with the protections, but the kernel reliably rejects them anyway, so running it without miri will reliably discover this error. mprotect can then be trivially implemented as a no-op.

As @saethlin already said, this is not an acceptable implementation for Miri. This implementation would miss UB which is the most critical kind of bug we can have in Miri.

I think currently there is no non-UB way to remove permissions from memory. LLVM simply does not allow this. So we should not accept it in Miri either -- someone first has to do the work of making LLVM compatible with such operations.

foonathan:
> I don't agree with this line of reasoning. Miri does a handful of things including having a weird scheduler and emulating weak memory effects that try to make the program encounter possible executions more easily than it would outside the interpreter. In addition, people use Miri's cross-interpreting support to test their code on other platforms, so it is definitely wrong to assume that such errors would be discovered by running without Miri.

> As @saethlin already said, this is not an acceptable implementation for Miri. This implementation would miss UB which is the most critical kind of bug we can have in Miri.

That is fair. I was planning on implementing it properly in a follow-up PR, and this just served as initial launchpad.

> Just off the top of my head, how would you restore write permission to the correct tags if a page goes from PROT_READ|PROT_WRITE down to just PROT_READ then back up to PROT_READ|PROT_WRITE? I don't think you should be touching the aliasing model here at all, it would be better to store additional state about the permissions of mapped address ranges then check those in before_memory_read/before_memory_write.

That does seem a lot easier than what I was trying to do, I'll do that then directly.

> I think currently there is no non-UB way to remove permissions from memory. LLVM simply does not allow this. So we should not accept it in Miri either -- someone first has to do the work of making LLVM compatible with such operations.

For my use case, I only care about adding permissions to memory (start with `PROT_NONE`, then add `PROT_READ | PROT_WRITE`). I can start by implementing that and defer removing permissions.

> That "mostly" is doing a lot of heavy lifting. MADV_DONTNEED on Linux changes the contents of memory so this implementation is wrong. Actually upon re-reading your comment in the code I think you have misunderstood the man page? MADV_DONTNEED documents that subsequent accesses to private anonymous mappings (the only kind we support) see zero-fill-on-demand pages.

True, but that's a Linux specific guarantee; POSIX does not require it. So I essentially implemented [`posix_madvise`](https://www.man7.org/linux/man-pages/man3/posix_madvise.3.html) not `madvise`.  But yes, if miri supports it, it should probably do the technically non-portable thing that people expect from Linux. Although at this point, you then do have to worry about aliasing, since its essentially doing a write.

RalfJung:
> For my use case, I only care about adding permissions to memory (start with PROT_NONE, then add PROT_READ | PROT_WRITE). I can start by implementing that and defer removing permissions.

Changing permissions at all is not something LLVM currently permits. So the best we can do is something like
- we consider the `mmap` with PROT_NONE to be a "NOP" on the Abtract Machine (except that the address range gets reserved)
- the allocation only gets created when the permissions get changed

Except that's not really correct either since then it's unclear how the pointer is supposed to get the right provenance. If the allocation only gets created by the permission change, the pointer must be created *after* the permission change. Otherwise the compiler could reorder a `ptr.write` up above the permission change and that would obviously be problematic. So you still need to pass the pointer through an inline asm block or so such that we can pretend that it gets the provenance that was generated when the allocation got created during the permission change.

Or alternatively we could adopt a model like https://github.com/rust-lang/unsafe-code-guidelines/issues/606, but that requires blessing from LLVM.

See [this blog post](https://www.ralfj.de/blog/2026/03/13/inline-asm.html) for the underlying reasoning principles I am using here.

> True, but that's a Linux specific guarantee; POSIX does not require it. 

This argument is the wrong way around. As long as POSIX *permits* implementations to forget memory, Miri should exhibit that behavior. Miri is supposed to be a UB checker which for POSIX APIs means it must be a maximally "adversarial" implementation.

RalfJung:
>  So the best we can do is something like

Nikita reminded me that there's something slightly better we can do, where `mmap` creates an allocation of size 0 and `mprotect` grows the existing allocation. That's allowed by the LLVM LangRef. But it requires the `mprotect` calls to be made strictly in increasing address order.

RalfJung:
Okay, to taking a step back:
- Misusing `mprotect` can easily cause UB, so we can't just make this a NOP in Miri.
- Same for `madvise(MADV_DONTNEED)`. We should have an allowlist of flags that we can ignore here, and throw an "unsupported" error for anything else. (We could also return `-1` and set errno, but usually we prefer to fail loudly as that's typically less confusing to debug.)
- We can support a specific version of the `mmap`+`mprotect` pattern, but implementing it will be a bit tricky.

I assume that the `mmap` call sets `addr` to 0 so that Miri can choose the address? If not, that'll be even more tricky. ;)

If this assumption is correct, we need to roughly do this:
- In an `mmap(PROT_NONE)` call of a given size, create a "growable" allocation of size 0 for which we reserve `length` amount of memory in the address space. Return a pointer to that allocation.
- In an `mprotect(PROT_READ|PROT_WRITE)` call, check if the pointer points to the end of such an allocation. If yes, grow it by the given size. Raise UB if that would make the allocation larger than the size originally determined by `mmap`.

This needs a fundamentally new concept in the interpreter: a growable allocation. That will need patches on the rustc side where the heart of the interpreter lives, we can't do this all in the Miri repo. I'm happy to provide further details if you want to look into this. :)

I would leave any other permission changes to future work if this is already useful -- they might need more care to ensure Rust/LLVM can even guarantee their correctness. I also agree with @saethlin that the aliasing model is the wrong tool here. `mprotect` can cause accesses to segfault which is very different from what the aliasing model is supposed to model.

@rustbot author

Fayti1703:
> > True, but that's a Linux specific guarantee; POSIX does not require it.
> 
> This argument is the wrong way around. As long as POSIX _permits_ implementations to forget memory, Miri should exhibit that behavior. Miri is supposed to be a UB checker which for POSIX APIs means it must be a maximally "adversarial" implementation.

`posix_madvise` (the variant specified in POSIX) _doesn't_ permit this:

> The _posix_madvise_() function shall have no effect on the semantics of access to memory in the specified range, although it may affect the performance of access.

(<https://pubs.opengroup.org/onlinepubs/9799919799/functions/posix_madvise.html>)

The unprefixed `madvise` function that, on Linux, changes the memory when called with `MADV_DONTNEED` does not exist in POSIX.

RalfJung:
I see, I got that mixed up then. But the PR implements "madvise" so the concerns do apply.

foonathan:
Okay, I've reverted my lenient protection support and implemented only the true no-ops (i.e. a redundant mprotect call with PROT_WRITE | PROT_READ, madvise other than `MADV_DONTNEED`). This actually cuts down the conditional `cfg(miri)` I need significantly - all I need to support my codebase is a conditional alias of `PROT_NONE ` to `PROT_READ | PROT_WRITE`, so this is a good start. I will explore the support for PROT_NONE and the incremental expansion with mprotect in a follow-up PR, thanks for the pointers!

About the adversial implementation of `MADV_DONTNEED`: While yes, POSIX does not require setting it to zero, but we should do it to catch bugs if someone relies on the values still being there after an `MADV_DONTNEED`, I can also imagine a scenario where setting it to zero can hide bugs where someone relied on `MADV_DONTNEED` clearing the memory before re-using it. I think a true adversial implementation should therefore set it to a non-zero value to catch both kinds of mistakes. I will look into that in a follow-up PR as well.

@rustbot ready

tests/fail-dep/libc/madvise_out_of_bounds.stderr
```diff
@@ -0,0 +1,25 @@
+error: Undefined Behavior: in-bounds pointer arithmetic failed: attempting to offset pointer by 8192 bytes, but got ALLOC which is only 4096 bytes from the end of the allocation
```
foonathan: How do I make the test check not dependent on the exact page size? MacOS fails because of their bigger pages.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5063/48777d7c2f7a273f976f6fa549b483f6c09f41e1..5bd288de1744637be3e4a64dccfe797ef91c812b)*

RalfJung: You can add a `//@normalize-stderr-test` (see other tests for examples) to remove the page size from the output and thus make it consistent across targets.

RalfJung:
> About the adversial implementation of MADV_DONTNEED: While yes, POSIX does not require setting it to zero, but we should do it to catch bugs if someone relies on the values still being there after an MADV_DONTNEED, I can also imagine a scenario where setting it to zero can hide bugs where someone relied on MADV_DONTNEED clearing the memory before re-using it. I think a true adversial implementation should therefore set it to a non-zero value to catch both kinds of mistakes. I will look into that in a follow-up PR as well.

I don't know how liberal the contract is. The most adversarial thing we could do in Miri is to reset the memory to uninitialized. I think that should safely model every possible real implementation, but it might be too adversarial for what programs want to do with it.

src/shims/unix/mem.rs
```diff
@@ -152,4 +135,106 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            && advise != madv_random
+            && advise != madv_sequential
+            && advise != madv_willneed
+        {
+            throw_unsup_format!(
+                "Miri does not support calls to madvise with advice other than MADV_NORMAL, MADV_RANDOM, MADV_SEQUENTIAL, or MADV_WILLNEED",
+            );
+        }
+
+        // the pages from `[addr, addr + length)` must be mapped, so length definitely must not overflow
```
RalfJung:
```suggestion
        // The pages from `[addr, addr + length)` must be mapped, so length definitely must not overflow.
```
Applies to all other new comments as well: please use full sentences with regular punctuation. Too often have I been confused by comments that just string together words without indicating where the sentences end (because new lines get added to such existing comments without adjusting the existing comments).

src/shims/unix/mem.rs
```diff
@@ -152,4 +135,106 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+        // All advises we support are no-ops.
+        interp_ok(Scalar::from_i32(0))
+    }
+}
+
+fn round_up_to_page_size(this: &MiriInterpCx<'_>, length: u64) -> Option<u64> {
+    length
+        .checked_next_multiple_of(this.machine.page_size)
+        .filter(|length| *length <= this.target_usize_max())
```
RalfJung: The maximum size of an allocation is `isize::MAX` so please check for that here.

(I realize this is pre-existing, but if we're reorganizing this logic we might as well fix it.)

src/shims/unix/mem.rs
```diff
@@ -152,4 +135,106 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            throw_unsup_format!(
+                "Miri does not support calls to madvise with advice other than MADV_NORMAL, MADV_RANDOM, MADV_SEQUENTIAL, or MADV_WILLNEED",
+            );
+        }
+
+        // the pages from `[addr, addr + length)` must be mapped, so length definitely must not overflow
+        let Some(length) = round_up_to_page_size(this, length) else {
+            return this.set_errno_and_return_neg1_i32(LibcError("ENOMEM"));
+        };
+        // we do pointer arithmetic, so it would be UB to go out of bounds
```
RalfJung: Not sure what you mean here, I can't see any pointer arithmetic?
The compiler gives no meaning to madvise so I think it'd be fine to just not care whether the memory really exists or not... but what would that do in a real execution?

src/shims/unix/mem.rs
```diff
@@ -152,4 +135,106 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        this.check_ptr_access(
+            addr,
+            Size::from_bytes(length),
+            CheckInAllocMsg::InboundsPointerArithmetic,
+        )?;
+
+        // We only support `PROT_READ|PROT_WRITE`, so there is nothing to do here.
+        interp_ok(Scalar::from_i32(0))
```
RalfJung: So here we assume that it is okay to call `mprotect(PROT_READ|PROT_WRITE)` on a read-only allocation, such as a non-mutable `static`. That seems fine to me, though note that it is still UB to write to that memory -- the compiler assumes it remains read-only. The compiler doesn't care about the page table entries though, it only cares that you don't mutate the memory.

So, this seems fine, but there should be a comment explaining that subtlety.

src/shims/unix/foreign_items.rs
```diff
@@ -828,6 +828,18 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
             }
+            "mprotect" => {
+                let [addr, length, prot] =
+                    this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
+                let result = this.mprotect(addr, length, prot)?;
+                this.write_scalar(result, dest)?;
+            }
+            "madvise" => {
+                let [addr, length, advice] =
+                    this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
```
RalfJung: Given that these are new shims, can you use `check_shim_sig` instead? You can grep for that to find a bunch of example usage. `check_shim_sig_lenient` is the deprecated alias because we haven't converted everything yet.

src/shims/unix/mem.rs
```diff
@@ -152,4 +135,106 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            return this.set_errno_and_return_neg1_i32(LibcError("EINVAL"));
+        }
+
+        verify_prot(this, prot)?;
+
+        // the pages from `[addr, addr + length)` must be mapped, so length definitely must not overflow
+        let Some(length) = round_up_to_page_size(this, length) else {
+            return this.set_errno_and_return_neg1_i32(LibcError("ENOMEM"));
+        };
+        // we do pointer arithmetic, so it would be UB to go out of bounds
```
RalfJung:
```suggestion
        // Ensure this is actually allocated memory.
```

src/shims/unix/mem.rs
```diff
@@ -152,4 +135,106 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+        this.check_ptr_access(
+            addr,
+            Size::from_bytes(length),
+            CheckInAllocMsg::InboundsPointerArithmetic,
+        )?;
```
RalfJung: The error talking about pointer arithmetic is a bit confusing. Also, is this truly UB? Maybe we should classify it as "unsupported".
```suggestion
        ).map_err_kind(|_| err_unsup_format!("`mprotect` called on non-dereferenceable memory"))?;
```

RalfJung (commented):
Thanks a lot! Overall this looks good, I have a bunch of minor feedback. :)
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5063/48777d7c2f7a273f976f6fa549b483f6c09f41e1..b9f6bd660bfe9b1f076d444ef1e3165afd952271)*

src/shims/unix/mem.rs
foonathan: The POSIX spec says you can call `mprotect`/`madvise` with any address and length and get `ENOMEM`  (Addresses in the specified range are not currently mapped, or are outside the address space of the process.).  However, I thought it might be more useful if miri fails loudly in those situations: Logically, when you're passing an `addr` and `length` pair to a C function you're defining a slice, after all, and so it "feels" like UB if this slice goes out of bounds.

But I'm also fine to implement the POSIX spec here (which might still involve calling check_allocation but then turning an error into returning `ENOMEM`?).

src/shims/unix/mem.rs
foonathan: see above

src/shims/unix/mem.rs
RalfJung: Whether or not memory is inbounds is a separate question from whether it is mapped... a bunch of out-of-bounds memory might still be mapped since the allocator decided it's better not (yet) to `munmap` those pages. Whether or not some memory is OOB is not really observable to the program (any way of "testing" it just triggers UB if you are wrong) 

So... okay, let's go with UB for now. That's consistent with what we do for `munmap` if you don't follow the discipline we expect.

src/shims/unix/mem.rs
RalfJung: Let's go with this then, for both cases
```suggestion
        ).map_err_kind(|_| err_ub_format!("`mprotect` called on out-of-bounds memory"))?;
```

RalfJung:
Since we're getting reading to land this, please also squash the commits. (`./miri squash` can help with that for the typical case of making it all a single commit.)

foonathan:
Sorry for the delay, was at a C++ committee meeting. Should all be done now.

RalfJung:
Thanks! Sorry for the slow response, now it was my turn to travel. ;)

PR #5064, opened 2026-05-26
aarch64: implement SHA256 intrinsic shims 

jrakibi:
This adds shims for the four AArch64 SHA256  intrinsics:
- `vsha256hq_u32`
- `vsha256h2q_u32`
- `vsha256su0q_u32`
- `vsha256su1q_u32`

The math helpers are adapted from RustCrypto's SHA-256 soft implementation:
https://github.com/RustCrypto/hashes/blob/3d2bc57/sha2/src/sha256/soft/compact.rs

The test in this PR follows the same structure as the existing x86 SHA intrinsic test in https://github.com/rust-lang/miri/pull/3752 , so hopefully it should make review easier.

I also tested this patch against the [rust-bitcoin](https://github.com/rust-bitcoin/rust-bitcoin/) `bitcoin_hashes` crate, which has the full [NIST SHAVS](https://csrc.nist.gov/projects/cryptographic-algorithm-validation-program/secure-hashing) SHA256 vectors as an extra local check.


jrakibi:
Force-pushed to fix the fmt CI failure.

Also, if it would make review easier, I can rewrite the SHA256 helpers to follow the [ARM pseudocode](https://developer.arm.com/architectures/instruction-sets/intrinsics/vsha256su0q_u32) directly.

RalfJung:
Thanks for the PR!

We already have some sha256 code in `src/shims/x86/sha.rs`. Can you just move that code to `shims/math.rs` and reuse it?
I guess there's no small crate we could import to avoid having to have our own implementation here?

@rustbot author

jrakibi:
> We already have some sha256 code in src/shims/x86/sha.rs. Can you just move that code to shims/math.rs and reuse it?

The x86 and ARM SHA256 instructions have different structures (different nmber of args, different rounds per call), so the functions themselves can't be directly reused between the two architectures. 

would you like me to move everything to `math.rs` ?

> I guess there's no small crate we could import to avoid having to have our own implementation here?

I didn't find any crate that handles emulation of SHA2 hardware intrinsics.

RalfJung:
> The x86 and ARM SHA256 instructions have different structures (different nmber of args, different rounds per call)

But surely the core logic for a single round is the same and can be shared?

jrakibi:
> can you just move that code to shims/math.rs and reuse it?

Done, moved the single round + sigma0/sigma1 into `math.rs`, both x86 and aarch64 call it now

@rustbot  ready

src/shims/math.rs
```diff
@@ -301,3 +301,38 @@ pub(crate) fn compute_crc32(crc: u32, data: u64, bit_size: u32, polynomial: u128
 
     u32::try_from(dividend).unwrap().reverse_bits()
 }
+
+// sha256 primitives shared by the x86 and aarch64 shims. Math helpers adapted from RustCrypto soft impl:
```
RalfJung: Please move these in a `mod sha256` block (in the same file) to group them more clearly.

src/shims/aarch64.rs
```diff
@@ -236,8 +236,159 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+    ecx: &mut MiriInterpCx<'c>,
+    dest: &MPlaceTy<'c>,
+    val: [u32; 4],
+) -> InterpResult<'c, ()> {
+    for (i, part) in val.into_iter().enumerate() {
+        let projected = &ecx.project_index(dest, i.to_u64())?;
+        ecx.write_scalar(Scalar::from_u32(part), projected)?;
+    }
+    interp_ok(())
+}
```
RalfJung: Please give these functions more specific names.

tests/pass/shims/aarch64/intrinsics-aarch64-sha.rs
```diff
@@ -0,0 +1,143 @@
+//@only-target: aarch64
+//@compile-flags: -C target-feature=+sha2
```
RalfJung, line 2:
```suggestion
//@compile-flags: -C target-feature=+sha2
//@run-native
```
Let's check that the tests are correct (assuming that github's aarch64 machines have this extension)

tests/pass/shims/aarch64/intrinsics-aarch64-sha.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    let mut hash = String::new();
+    for word in &state {
+        write!(hash, "{:08x}", word).expect("writing to String doesn't fail");
+    }
+    assert_eq!(hash, "1b2293d21b17a0cb0c18737307c37333dea775eded18cefed45e50389f9f8184");
+}
+
+// Adapted from RustCrypto's sha256 aarch64 implementation:
+// https://github.com/RustCrypto/hashes/blob/master/sha2/src/sha256/aarch64_sha2.rs
```
RalfJung: To simplify review, can you make this *exactly identical* to the RustCrypto code?

RalfJung (commented):
Thanks, and sorry for the slow reply!

Unfortunately, I can't review code like this. This isn't just re-arranging the existing x86 SHA256 code, it seems to be completely rewriting it.   I don't know how to accept such code, any one of these constants or operations could be subtly wrong and I'd never notice. As someone without any knowledge in cryptography, I cannot accept cryptography code from someone I don't know.


@newpavlov any chance you could have a look at the hashing parts of this?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5064/052dde604b83f5eb0312cbda178777cb48166747..eef40c63c3f9474f9cdbb739a0414df9301f9a97)*

RalfJung:
@rustbot author
for the minor changes mentioned above

newpavlov:
I haven't reviewed the implementation thoroughly, but it looks fine on the first glance. Although, I would generate random test vectors for each intrinsic instead of re-implementing the SHA-256 compress function and testing the intrinsics through it.

RalfJung:
I like having end-to-end tests, but I wouldn't say no to also testing some test vectors.

src/shims/aarch64.rs
jrakibi: changed the names to `read_u32x4`/`write_u32x4`

tests/pass/shims/aarch64/intrinsics-aarch64-sha.rs
jrakibi: Done. 
Only deviation: the original `vld1q_u32(&K32[i])` is UB (a &u32 read as 16 bytes) and Miri itself rejects it, so I changed those loads to use K32[i..i + 4].as_ptr() instead.

We had the same bug in rust-bitcoin too (rust-bitcoin/rust-bitcoin#6262) after running bitcoin_hashes under this branch, Miri detected the same UB.

src/shims/math.rs
jrakibi: done in 407afc5ba2dd36a702d77dc569f9b5cc2f95d59f

src/shims/aarch64.rs
```diff
@@ -273,8 +273,155 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+
+                write_u32x4(this, &dest, result)?;
+            }
             _ => return interp_ok(EmulateItemResult::NotSupported),
         }
         interp_ok(EmulateItemResult::NeedsReturn)
     }
 }
+
+/// Reads the four `u32` lanes of a `uint32x4_t` SIMD register.
```
RalfJung: There are no registers in Miri. ;)
```suggestion
/// Reads a `[u32; 4]` array.
```

src/shims/aarch64.rs
```diff
@@ -273,8 +273,155 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.write_scalar(Scalar::from_u128(result), &dest)?;
             }
 
+            "crypto.sha256h" => {
+                this.expect_target_feature_for_intrinsic(link_name, "sha2")?;
+
+                let [abcd, efgh, wk] = this.check_shim_sig_unadjusted(link_name, args)?;
+
+                let (abcd_reg, abcd_len) = this.project_to_simd(abcd)?;
```
RalfJung: Please don't use "reg(ister)" terminology, it doesn't really make sense here. These are all just values.
```suggestion
                let (abcd, abcd_len) = this.project_to_simd(abcd)?;
```

src/intrinsics/aarch64.rs
```diff
@@ -273,8 +273,155 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.write_scalar(Scalar::from_u128(result), &dest)?;
             }
 
+            "crypto.sha256h" => {
```
RalfJung, line 319: Please add comments explaining which user-visible stdarch intrinsics these names correspond to.

RalfJung (commented):
Thanks @newpavlov for taking a look and thanks @jrakibi for the revisions! I see we already do have a bunch of test vectors for the x86 ones as well so... I guess it's fine? I am relying on the fact that, as far as I know, small changes to the algorithm will likely completely perturb the output and hence the tests would catch that.

I just have some minor comments left. Please also squash the commits then as we are nearing the end of the review process.

@apoelstra @Kixunil if you happen to have any time to look at this, that would be great.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5064/db1e37db2fe9c926ddc3cea3bcf916fa2893c26f..86ddafa1b3b31a7c83536c98ddeac3432eb2773d)*

src/shims/aarch64.rs
jrakibi: Fixed, I took the naming from `x86/sha.rs` to keep the two implementations easy to compare. Happy to rename it there too, in this PR or a follow-up if needed.

RalfJung: Ah, that must have slipped through back then.

Yeah if you want to feel free to rename those as well, but keep that in a separate commit.

jrakibi: Done in f13b90bcf12e24e01e5420338f6c7a95b4f58053

RalfJung (commented):
I think we waited long enough, let's land this.
However, there are conflicts that need to be resolved (some of the files moved around -- sorry for that).
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5064/811aefc249581e05efd072f6e16d57f816e38eda..f13b90bcf12e24e01e5420338f6c7a95b4f58053)*

RalfJung:
Thanks :)

PR #5067, opened 2026-05-26
enforce noalias on custom allocator Box by default

RalfJung:
As requested by @nia-e, let's use the more conservative semantics for now. Also add a flag to disable noalias on custom allocator Boxes. That flag wins the price for the longest flag: `-Zmiri-tree-borrows-relax-custom-allocator-uniqueness`. I am open to suggestions for shorter names. :)

In Stacked Borrows we just always enforce full uniqueness. It's meant to be the strict model anyway.


nia-e:
libs-api convention dictates we call this `-Zbikeshed-aliasable-allocator` 😄

this will be very useful, ty!

PR #5068, opened 2026-05-27
[Priroda] initial standalone CLI crate

moabo3li:
This pull request introduces a new step-through debugger for Rust MIR called Priroda.

It adds a new `priroda` crate with a simple CLI for debugging Rust programs running under Miri, and makes several changes to the core Miri codebase to support this integration. 

The changes include exposing internal APIs


moabo3li:
@oli-obk

priroda/README.md
```diff
@@ -0,0 +1,38 @@
...
+./miri install
+```
+
+Then build the Miri sysroot and export it for Priroda:
+
+```sh
+cd priroda
+cargo +miri miri setup
+export MIRI_SYSROOT="$(cargo +miri miri setup --print-sysroot)"
+```
```
oli-obk, line 27: Maybe the next step should be to add `./miri priroda` and `cargo miri --priroda` as commands so it all integrates well with local experiments :D 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5068/b2b7b7a0e661c7c5c0283f50ad571c496121c6b9..00a703c17f8efcceec2eae13cd6608ff1dbb71d4)*

priroda/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+name = "priroda"
+repository = "https://github.com/rust-lang/miri"
+version = "0.1.0"
+edition = "2024"
+
+
+[[bin]]
+name = "priroda"
+path = "src/main.rs"
+test = true          # we have unit tests
```
oli-obk: Copy paste mistake, priroda doesn't have unit tests

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5068/b2b7b7a0e661c7c5c0283f50ad571c496121c6b9..00a703c17f8efcceec2eae13cd6608ff1dbb71d4)*

priroda/Cargo.lock
oli-obk, line 1: does it work to use a workspace so we can share the `Cargo.lock` with the rest of miri?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5068/b2b7b7a0e661c7c5c0283f50ad571c496121c6b9..00a703c17f8efcceec2eae13cd6608ff1dbb71d4)*

priroda/src/main.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+
+impl rustc_driver::Callbacks for PrirodaCompilerCalls {
+    fn after_analysis<'tcx>(&mut self, _: &interface::Compiler, tcx: TyCtxt<'tcx>) -> Compilation {
+        tcx.dcx().emit_stashed_diagnostics();
+        tcx.dcx().abort_if_errors();
+
+        if !tcx.crate_types().contains(&CrateType::Executable) {
+            tcx.dcx().fatal("priroda only makes sense on bin crates");
```
oli-obk, line 60: ooh, far future thing, but we could def allow priroda to list functions and just call them with manually entered args or sth.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5068/b2b7b7a0e661c7c5c0283f50ad571c496121c6b9..00a703c17f8efcceec2eae13cd6608ff1dbb71d4)*

priroda/src/main.rs
```diff
@@ -0,0 +1,160 @@
...
+        let result = run_cli_loop(&mut session);
+
+        match result.report_err() {
+            Ok(()) => {}
+            Err(err) =>
+                if let Some((return_code, _leak_check)) = report_result(&session.ecx, err) {
+                    if return_code != 0 {
+                        std::process::exit(return_code);
+                    }
+                },
```
oli-obk, line 76: This will need adjusting in the future. Unlike miri, prioda should not exit when the evaluated program exits. Instead it should do the gdb-thing and print the exit code and await further commands

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5068/b2b7b7a0e661c7c5c0283f50ad571c496121c6b9..00a703c17f8efcceec2eae13cd6608ff1dbb71d4)*

priroda/src/main.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    let mut args: Vec<String> = std::env::args().collect();
+
+    args.splice(1..1, miri::MIRI_DEFAULT_ARGS.iter().map(ToString::to_string));
+
+    let sysroot_flag = String::from("--sysroot");
+    if !args.contains(&sysroot_flag) {
+        args.push(sysroot_flag);
+        args.push(find_sysroot());
+    }
```
oli-obk, line 40: future thing: we need to handle all the `-Z` flags that miri accepts

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5068/b2b7b7a0e661c7c5c0283f50ad571c496121c6b9..00a703c17f8efcceec2eae13cd6608ff1dbb71d4)*

oli-obk (changes requested):

src/concurrency/thread.rs
```diff
@@ -856,6 +842,21 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
 // Public interface to thread management.
 impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    /// Public because this is used by Priroda.
+    #[inline]
+    fn run_on_stack_empty(&mut self) -> InterpResult<'tcx, Poll<()>> {
```
RalfJung: Are we sure this is the right thing to expose publicly? This function is very internal and unsafe (in the sense of interpreter invariants, not meta-level UB). It seems better to expose a "step" function, and have Miri actually use that for itself as well.

moabo3li: That makes sense.  for more context I asked about that here [#miri/debugger > Priroda step loop and Miri stack-empty handling @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/586516-miri.2Fdebugger/topic/Priroda.20step.20loop.20and.20Miri.20stack-empty.20handling/near/597546939)

priroda/Cargo.toml
moabo3li: Done

priroda/Cargo.lock
moabo3li: Using a workspace shares `Cargo.lock` correctly in standalone Miri, but I noticed that Miri is synced into rustc’s existing workspace. Keeping `[workspace]` in `src/tools/miri/Cargo.toml` would create a nested workspace root there. Since is `./miri priroda` / `cargo miri --priroda`, would it will be better to make it another `[[bin]]` target of the existing miri package? The alternative is keeping it as a separate crate and coordinating a rustc-side workspace member.

src/concurrency/thread.rs
```diff
@@ -856,6 +856,24 @@ trait EvalContextPrivExt<'tcx>: MiriInterpCxExt<'tcx> {
 // Public interface to thread management.
 impl<'tcx> EvalContextExt<'tcx> for crate::MiriInterpCx<'tcx> {}
 pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
+    /// Public because this is used by Priroda.
+    fn miri_step(&mut self) -> InterpResult<'tcx, bool> {
```
RalfJung: What does the return value mean? Please add that to the doc comment.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5068/c74d8ad72ae8d20b1a95512e728aa4afe38d4123..6617b3b85cbeb6ad3fd85098874a856bc581f552)*

moabo3li: Looking at this again, I don't think the boolean should be part of the extracted API. It only indicates that the active thread terminated, while Priroda was treating it as whole-program completion. Miri already reports program exit through TerminationInfo::Exit, so I will change miri_step to return InterpResult<()>.

RalfJung (commented):
Miri parts LGTM. :)

The name `miri_step` isn't great but I can't think of anything better right now either, so we can go with it for now.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5068/b537f496edf90808f44b6d9e67e3238c6aa09f62..1b1e9027468f9af6fb768246229eaaa1e3c31638)*

priroda/README.md
moabo3li: I'll include that at the next pr

priroda/Cargo.lock
oli-obk: yea I think it should be a bin target, but let's go with what you had before for now and we'll make it a bin target when adding CI

oli-obk (approved):

PR #5070, opened 2026-05-27
add shim for `neon.pmull64`

folkertdev:
the aarch64 variant of x86's `pclmulqdq`, using carryless_mul for polynomial multiplication in GF(2). This intrinsic is useful for the crc32 algorithm.

The test has been validated on real hardware.

### DIFF
--- a/src/shims/aarch64.rs
+++ b/src/shims/aarch64.rs
@@ -214,6 +214,30 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let result = compute_crc32(crc, data, bit_size, polynomial);
                 this.write_scalar(Scalar::from_u32(result), dest)?;
             }
+            // Polynomial multiply long (64-bit x 64-bit -> 128-bit).
+            //
+            // This is the same as "carryless" multiplication, see
+            // <https://en.wikipedia.org/wiki/Carry-less_product#Multiplication_of_polynomials>.
+            //
+            // Used to implement the vmull_p64 and vmull_high_p64 functions.
+            // https://developer.arm.com/architectures/instruction-sets/intrinsics/vmull_p64
+            "neon.pmull64" => {
+                // LLVM and GCC group pmull with the AES intrinsics.
+                // Also see <https://gcc.gnu.org/pipermail/gcc-patches/2023-February/612088.html>.
+                this.expect_target_feature_for_intrinsic(link_name, "aes")?;
+
+                let [left, right] =
+                    this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
+                let left = this.read_scalar(left)?.to_u64()?;
+                let right = this.read_scalar(right)?.to_u64()?;
+
+                let result = left.widening_carryless_mul(right);
+
+                // dest is int8x16_t, transmute to u128 for the write.
+                let dest = dest.transmute(this.machine.layouts.u128, this)?;
+                this.write_scalar(Scalar::from_u128(result), &dest)?;
+            }
+
             _ => return interp_ok(EmulateItemResult::NotSupported),
         }
         interp_ok(EmulateItemResult::NeedsReturn)
--- a/tests/pass/shims/aarch64/intrinsics-aarch64-aes.rs
+++ b/tests/pass/shims/aarch64/intrinsics-aarch64-aes.rs
@@ -1,0 +1,45 @@
+// We're testing aarch64 AES target specific features.
+//@only-target: aarch64
+//@compile-flags: -C target-feature=+neon,+aes
+
+use std::arch::aarch64::*;
+use std::arch::is_aarch64_feature_detected;
+
+fn main() {
+    assert!(is_aarch64_feature_detected!("neon"));
+    assert!(is_aarch64_feature_detected!("aes"));
+
+    unsafe {
+        test_vmull_p64();
+        test_vmull_high_p64();
+    }
+}
+
+#[target_feature(enable = "neon,aes")]
+unsafe fn test_vmull_p64() {
+    assert_eq!(vmull_p64(0, 0), 0);
+    assert_eq!(vmull_p64(0, 0xffffffffffffffff), 0);
+    assert_eq!(vmull_p64(1, 1), 1);
+    assert_eq!(vmull_p64(1, 0x8000000000000000), 0x8000000000000000);
+
+    assert_eq!(vmull_p64(0b11, 0b11), 0b101);
+
+    // Check with the same inputs that are used in the x86_64 pclmulqdq test.
+    assert_eq!(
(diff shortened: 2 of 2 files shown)


src/shims/aarch64.rs
```diff
@@ -214,6 +214,25 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 let result = compute_crc32(crc, data, bit_size, polynomial);
                 this.write_scalar(Scalar::from_u32(result), dest)?;
             }
+            // Polynomial multiply long (64-bit x 64-bit -> 128-bit).
```
RalfJung, line 217: What does "polynomial" mean here? The implementation invokes `widening_carryless_mul` so I guess "polynomial" means "carryless"?

src/shims/aarch64.rs
```diff
@@ -214,6 +214,25 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            // Used to implement the vmull_p64 and vmull_high_p64 functions.
+            // https://developer.arm.com/architectures/instruction-sets/intrinsics/vmull_p64
+            "neon.pmull64" => {
+                this.expect_target_feature_for_intrinsic(link_name, "aes")?;
```
RalfJung, line 227: This is in the "aes" target feature? Odd. https://developer.arm.com/architectures/instruction-sets/intrinsics/vmull_p64 doesn't mention the target feature. Where would one have to look to verify this?

src/shims/aarch64.rs
```diff
@@ -214,6 +214,25 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                let [left, right] =
+                    this.check_shim_sig_lenient(abi, CanonAbi::C, link_name, args)?;
+                let left = this.read_scalar(left)?.to_u64()?;
+                let right = this.read_scalar(right)?.to_u64()?;
+
+                let result = left.widening_carryless_mul(right);
+
+                // dest is int8x16_t, transmute to u128 for the write.
+                let dest_layout = this.layout_of(this.tcx.types.u128)?;
+                let dest = dest.transmute(dest_layout, this)?;
```
RalfJung:
```suggestion
                let dest = dest.transmute(this.machine.layouts.u128, this)?;
```

tests/pass/shims/aarch64/intrinsics-aarch64-aes.rs
```diff
@@ -0,0 +1,45 @@
...
+    // Check with the same inputs as pclmulqdq.
+    assert_eq!(
+        vmull_p64(0x7fffffffffffffff, 0xdd358416f52ecd34),
+        (2704901987789626761u128 << 64) | 13036940098130298092u128,
+    );
+}
+
+#[target_feature(enable = "neon,aes")]
+unsafe fn test_vmull_high_p64() {
+    // The lower element is ignored.
```
RalfJung:
```suggestion
    // The lower (first) element is ignored.
```

tests/pass/shims/aarch64/intrinsics-aarch64-aes.rs
```diff
@@ -0,0 +1,45 @@
...
+#[target_feature(enable = "neon,aes")]
+unsafe fn test_vmull_p64() {
+    assert_eq!(vmull_p64(0, 0), 0);
+    assert_eq!(vmull_p64(0, 0xffffffffffffffff), 0);
+    assert_eq!(vmull_p64(1, 1), 1);
+    assert_eq!(vmull_p64(1, 0x8000000000000000), 0x8000000000000000);
+
+    assert_eq!(vmull_p64(0b11, 0b11), 0b101);
+
+    // Check with the same inputs as pclmulqdq.
```
RalfJung: There's no pclmulqdq anywhere in this file... is this comment trying to say?

RalfJung (commented):
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5070/5cf9e754762fcfb69bbc0ec3256c65ec131270e2..21e6a6c669b5e1e05fcc1ebb78df3b2c471674dc)*

src/shims/aarch64.rs
folkertdev: I'm going off of the `stdarch` target features, which correspond to the LLVM ones. There is some other information scattered on the internet, e.g.

- https://gcc.gnu.org/pipermail/gcc-patches/2023-February/612088.html
- https://github.com/dotnet/runtime/issues/35143#issuecomment-619900227

That last one is by an arm engineer, stating

> It's been confirmed to me that indeed we expect people to implement AES and PMULL together. At some point this will be updated in the Arm ARM, but this also matches with llvm and gcc do.

So it's established practice, and seems to be intentional.

tests/pass/shims/aarch64/intrinsics-aarch64-aes.rs
folkertdev: The inputs are the same as the ones used in the test for pclmulqdq, I'll clarify

src/shims/aarch64.rs
folkertdev: Eh, I don't fully understand the math but 

https://en.wikipedia.org/wiki/Carry-less_product#Multiplication_of_polynomials

> The carry-less product can also be seen as a multiplication of polynomials over the field [GF(2)](https://en.wikipedia.org/wiki/GF(2)). This is because the exclusive or corresponds to the addition in this field.

RalfJung:
```suggestion
            // Polynomial multiply long (64-bit x 64-bit -> 128-bit).
            // (This is the same as "carryless" multiplication, see <https://en.wikipedia.org/wiki/Carry-less_product#Multiplication_of_polynomials>.)
```

src/shims/aarch64.rs
RalfJung:
```suggestion
                // LLVM groups pmull with the AES intrinsics. Also see <https://gcc.gnu.org/pipermail/gcc-patches/2023-February/612088.html>.
                this.expect_target_feature_for_intrinsic(link_name, "aes")?;
```
That last one has the concerning statement "*optional* part of AES"... what's the point of an extension flag if that doesn't actually tell me whether the function is present?!? Oh well, if LLVM can get away with it then I guess so can we.

tests/pass/shims/aarch64/intrinsics-aarch64-aes.rs
RalfJung: And what on Earth is pclmulqdq? Looks like the output of pwgen. ;)

EDIT: Oh, it's an x86 intrinsic. I should have guessed, those generally have terrible names...

PR #5071, opened 2026-05-27
Add support for level-triggered epoll

### LINKED ISSUE #3448: Implement basic epoll support
The epoll API mainly consists of epoll_create1, epoll_ctl, and epoll_wait. We have basic shims for all these functions but currently our epoll support is mostly a stub, since the actual polling part is still missing. (Also the behavior of `dup` on an epoll file descriptor is likely wrong.)

Implementing this is non-trivial, since it introduces a new kind of blocking to Miri: block until *any* of a set of FDs is read for reads/writes. We currently don't have any notion of file descriptors blocking at all, so this will need some careful design We probably need some notion of a per-file-descriptor wait lists of threads that are blocked on this FD? We should probably have [eventfd](https://github.com/rust-lang/miri/issues/3445) and [socketpair](https://github.com/rust-lang/miri/issues/3442) implemented first so that we have some examples of blocking FDs to consider.

Also note some interesting tidbits in [this SO question](https://stackoverflow.com/questions/37787619/duplicating-epoll-file-descriptor): we may have to explicitly implement the concept of a file *description* as that seems to be observable through the epoll APIs. On the plus side this means we can implement `dup` once and for all and we don't have to rely on each and every FD type doing it correctly.

WhySoBad:
Hi,

As a basis for implementing a `poll` shim, this pull request adds support for level-triggered epolls. Besides needing to switch from a ready set to a ready queue, the changes in the epoll shim are relatively small. 

To test the functionality of the level-triggered epoll I've added `level_triggered` and `edge_triggered` revisions to the epoll libc tests. At the moment I can't come up with any level-triggered edge cases which aren't covered by this testing method.  

As mentioned in rust-lang/miri#4725, the non-blocking epoll tests didn't use the shared `check_epoll_wait_noblock` helper but used a custom helper instead. I've changed that in the first commit (a tiny bit also slipped through into the second one). 

Closes https://github.com/rust-lang/miri/issues/3448

### DIFF
--- a/src/shims/unix/linux_like/epoll.rs
+++ b/src/shims/unix/linux_like/epoll.rs
@@ -1,7 +1,6 @@
 use std::cell::RefCell;
-use std::collections::{BTreeMap, BTreeSet, VecDeque};
+use std::collections::{BTreeMap, VecDeque};
 use std::io;
-use std::ops::Bound;
 use std::time::Duration;
 
 use rustc_abi::FieldIdx;
@@ -23,7 +22,9 @@ pub struct Epoll {
     interest_list: RefCell<BTreeMap<EpollEventKey, EpollEventInterest>>,
     /// The subset of interests that is currently considered "ready". Stored separately so we
     /// can access it more efficiently.
-    ready_set: RefCell<BTreeSet<EpollEventKey>>,
+    /// This is implemented as a queue so that for level-triggered epoll, all events eventually
+    /// get returned from `epoll_wait`. The queue does not contain any duplicates.
+    ready_events: RefCell<VecDeque<EpollEventKey>>,
     /// The queue of threads blocked on this epoll instance.
     queue: RefCell<VecDeque<ThreadId>>,
 }
@@ -46,6 +47,9 @@ pub struct EpollEventInterest {
     relevant_events: u32,
     /// The currently active events for this file descriptor.
     active_events: u32,
+    /// Boolean whether this is an edge-triggered interest.
+    /// When [`false`] it's a level-triggered interest instead.
+    is_edge_triggered: bool,
     /// The vector clock for wakeups.
     clock: VClock,
     /// User-defined data associated with this interest.
@@ -203,12 +207,8 @@ impl EpollInterestTable {
                     .extract_if(range_for_id(id), |_, _| true)
                     // Consume the iterator.
                     .for_each(drop);
-                epoll
-                    .ready_set
-                    .borrow_mut()
-                    .extract_if(range_for_id(id), |_| true)
-                    // Consume the iterator.
-                    .for_each(drop);
+                // Remove the ready events for this file description.
+                epoll.ready_events.borrow_mut().retain(|(fd_id, _)| fd_id != &id);
             }
         }
     }
@@ -303,6 +303,13 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 this.read_scalar(&this.project_field(&event, FieldIdx::ZERO)?)?.to_u32()?;
             let data = this.read_scalar(&this.project_field(&event, FieldIdx::ONE)?)?.to_u64()?;
 
+            let is_edge_triggered = if events & epollet == epollet {
+                events &= !epollet;
+                true
+            } else {
+                false
+            };
+
             // Unset the flag we support to discover if any unsupported flags are used.
             let mut flags = events;
             // epoll_wait(2) will always wait for epollhup and epollerr; it is not
@@ -311,12 +318,6 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             events |= epollhup;
             events |= epollerr;
 
-            if events & epollet != epollet {
-                // We only support edge-triggered notification for now.
-                throw_unsup_format!("epoll_ctl: epollet flag must be included.");
-            } else {
-                flags &= !epollet;
-            }
             if flags & epollin == epollin {
                 flags &= !epollin;
             }
@@ -350,6 +351,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 }
                 let new_interest = EpollEventInterest {
                     relevant_events: events,
+                    is_edge_triggered,
                     data,
                     active_events: 0,
                     clock: VClock::default(),
@@ -364,6 +366,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                     return this.set_errno_and_return_neg1_i32(LibcError("ENOENT"));
                 };
                 interest.relevant_events = events;
+                interest.is_edge_triggered = is_edge_triggered;
                 interest.data = data;
             }
 
@@ -391,7 +394,11 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // We did not have interest in this.
                 return this.set_errno_and_return_neg1_i32(LibcError("ENOENT"));
             };
-            epfd.ready_set.borrow_mut().remove(&epoll_key);
+            // Remove the ready event for this key, should one exist.
+            let mut ready_events = epfd.ready_events.borrow_mut();
+            if let Some(idx) = ready_events.iter().position(|k| k == &epoll_key) {
+                ready_events.remove(idx);
+            }
             // If this was the last interest in this FD, remove us from the global list
             // of who is interested in this FD.
             if interest_list.range(range_for_id(id)).next().is_none() {
@@ -469,7 +476,7 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             return this.set_errno_and_return_neg1(LibcError("EBADF"), dest);
         };
 
-        if timeout == 0 || !epfd.ready_set.borrow().is_empty() {
+        if timeout == 0 || !epfd.ready_events.borrow().is_empty() {
             // If the timeout is 0 or there is a ready event, we can return immediately.
             return_ready_list(&epfd, dest, &event, this)?;
         } else {
@@ -590,18 +597,29 @@ fn update_readiness<'tcx>(
         &mut dyn FnMut(EpollEventKey, &mut EpollEventInterest) -> InterpResult<'tcx>,
     ) -> InterpResult<'tcx>,
 ) -> InterpResult<'tcx> {
-    let mut ready_set = epoll.ready_set.borrow_mut();
+    let mut ready_events = epoll.ready_events.borrow_mut();
     for_each_interest(&mut |key, interest| {
         // Update the ready events tracked in this interest.
         let new_readiness = interest.relevant_events & active_events;
         let prev_readiness = std::mem::replace(&mut interest.active_events, new_readiness);
         if new_readiness == 0 {
             // Un-trigger this, there's nothing left to report here.
-            ready_set.remove(&key);
+            if let Some(idx) = ready_events.iter().position(|k| k == &key) {
+                ready_events.remove(idx);
+            }
         } else if force_edge || new_readiness != prev_readiness & new_readiness {
-            // Either we force an "edge" to be detected, or there's a bit set in `new`
-            // that was not set in `prev`. In both cases, this is ready now.
-            ready_set.insert(key);
+            // Either we force an "edge" to be detected or there's a bit set in `new_readiness`
+            // that was not set in `prev_readiness`. In both cases, this is ready now.
+
+            // We need to ensure that this event is not already part of the
+            // `ready_events` queue before enqueueing:
+            // <https://github.com/torvalds/linux/blob/HEAD/fs/eventpoll.c#L1292-L1296>
+            if !ready_events.contains(&key) {
+                ready_events.push_back(key);
+            }
+
+            // No matter whether this is newly ready or just re-triggered,
+            // the `epoll_wait` fetching this event should sync with the current thread.
             ecx.release_clock(|clock| {
                 interest.clock.join(clock);
             })?;
@@ -609,12 +627,12 @@ fn update_readiness<'tcx>(
         interp_ok(())
     })?;
     // While there are events ready to be delivered, wake up a thread to receive them.
-    while !ready_set.is_empty()
+    while !ready_events.is_empty()
         && let Some(thread_id) = epoll.queue.borrow_mut().pop_front()
     {
-        drop(ready_set); // release the "lock" so the unblocked thread can have it
+        drop(ready_events); // release the "lock" so the unblocked thread can have it
         ecx.unblock_thread(thread_id, BlockReason::Epoll { epfd: epoll.clone() })?;
-        ready_set = epoll.ready_set.borrow_mut();
+        ready_events = epoll.ready_events.borrow_mut();
     }
 
     interp_ok(())
@@ -629,7 +647,7 @@ fn return_ready_list<'tcx>(
     ecx: &mut MiriInterpCx<'tcx>,
 ) -> InterpResult<'tcx, i32> {
     let mut interest_list = epfd.interest_list.borrow_mut();
-    let mut ready_set = epfd.ready_set.borrow_mut();
+    let mut ready_events = epfd.ready_events.borrow_mut();
     let mut num_of_events: i32 = 0;
     let mut array_iter = ecx.project_array_fields(events)?;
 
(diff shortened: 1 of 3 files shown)


WhySoBad:
As far as I can tell, this should also close #3448 (and by extension #602?).

tests/pass-dep/libc/libc-epoll-no-blocking.rs
```diff
@@ -60,7 +42,7 @@ fn test_epoll_socketpair() {
     write_all(fds[0], b"abcde").unwrap();
 
     // Register fds[1] with EPOLLIN|EPOLLOUT|EPOLLET|EPOLLRDHUP
-    epoll_ctl_add(epfd, fds[1], EPOLLIN | EPOLLOUT | EPOLLET | EPOLLRDHUP).unwrap();
+    epoll_ctl_add(epfd, fds[1], EPOLLIN | EPOLLOUT | EPOLLRDHUP | EPOLLET).unwrap();
```
RalfJung: Why did you reorder these? Now it does not match the comment any more.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5071/bba9f7e43428a0046c1faec8e43ce92504f3401a..d7138681ddcbae245acf72a20dbfaef48eef17f9)*

tests/pass-dep/libc/libc-epoll-no-blocking.rs
```diff
@@ -194,24 +174,20 @@ fn test_two_same_fd_in_same_epoll_instance() {
...
 
     // Register both fd to the same epoll instance.
-    let mut ev = libc::epoll_event {
-        events: (libc::EPOLLIN | libc::EPOLLOUT | libc::EPOLLET).cast_unsigned(),
-        u64: 5u64,
-    };
-    let res = unsafe { libc::epoll_ctl(epfd, libc::EPOLL_CTL_ADD, fds[1], &mut ev) };
+    let mut ev =
+        libc::epoll_event { events: (EPOLLIN | EPOLLOUT | EPOLLET).cast_unsigned(), u64: 5u64 };
+    let res = unsafe { libc::epoll_ctl(epfd, EPOLL_CTL_ADD, fds[1], &mut ev) };
```
RalfJung: This should use `epoll_ctl_add`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5071/bba9f7e43428a0046c1faec8e43ce92504f3401a..d7138681ddcbae245acf72a20dbfaef48eef17f9)*

tests/pass-dep/libc/libc-epoll-no-blocking.rs
WhySoBad: Oops, this is a remain from splitting up my initial commit into two commits. The comment is updated in the second commit.

src/shims/unix/linux_like/epoll.rs
```diff
@@ -23,7 +22,10 @@ pub struct Epoll {
     interest_list: RefCell<BTreeMap<EpollEventKey, EpollEventInterest>>,
     /// The subset of interests that is currently considered "ready". Stored separately so we
     /// can access it more efficiently.
-    ready_set: RefCell<BTreeSet<EpollEventKey>>,
+    /// This is implemented as a queue because Linux just re-inserts ready events at the end of
+    /// the list for level-triggered epoll. We need to manually ensure that this queue doesn't
+    /// contain any duplicate events.
```
RalfJung:
```suggestion
    /// This is implemented as a queue so that for level-triggered epoll, all events eventually
    /// get returned from `epoll_wait`. The queue does not contain any duplicates.
```

src/shims/unix/linux_like/epoll.rs
```diff
@@ -590,31 +589,39 @@ fn update_readiness<'tcx>(
...
+            // that was not set in `prev_readiness`. In both cases, this is ready now.
+
+            // We need to ensure that this event is not already part of the
+            // `ready_events` queue before enqueueing:
+            // <https://github.com/torvalds/linux/blob/HEAD/fs/eventpoll.c#L1292-L1296>
+            if !ready_events.contains(&key) {
+                ready_events.push_back(key);
+                ecx.release_clock(|clock| {
+                    interest.clock.join(clock);
+                })?;
```
RalfJung: I think the clock update should always happen, even if the event is already in the queue. Logically, fetching this event from the queue should synchronize-with all the events that caused it to be ready, not just the first one.

src/shims/unix/linux_like/epoll.rs
```diff
@@ -661,11 +670,15 @@ fn return_ready_list<'tcx>(
         num_of_events = num_of_events.strict_add(1);
         // Synchronize receiving thread with the event of interest.
         ecx.acquire_clock(&interest.clock)?;
-        // This was an edge-triggered event, so remove it from the ready set.
-        ready_set.remove(&key);
-        // Go find the next event.
-        event_lower_bound = Bound::Excluded(key);
+        if interest.relevant_events & epollet == 0 {
```
RalfJung: `relevant_events` is meant for the events, I don't think we should also store the "ET" part in there. Please make this a separate `bool` field in `EpollEventInterest`.

src/shims/unix/linux_like/epoll.rs
```diff
@@ -645,12 +653,13 @@ fn return_ready_list<'tcx>(
...
     // While there is a slot to store another event, and an event to store, deliver that event.
-    // We can't use an iterator over `ready_set` as we want to remove elements as we go,
-    // so we track the most recently delivered event to find the next one. We track it as a lower
-    // bound that we can pass to `BTreeSet::range`.
-    let mut event_lower_bound = Bound::Unbounded;
+    // Because `ready_events` is a queue, we just pop on the front until no more ready events exist.
+    // We also pop level-triggered events to prevent delivering them multiple times for a single
+    // `epoll_wait` call. Those level-triggered events are then stored in the `delivered_events`
+    // list such that we can re-insert them back into the queue afterwards.
+    let mut delivered_events = Vec::new();
```
RalfJung: The name isn't correct as we only collect level-triggered events here.

What about the following:
in the loop we directly do `ready_events.push_back`,
and also instead of `array_iter` we iterate over something like
```rust
// We will fill at most the first `ready_events.len()` slots of the array.
// Bounding the iterator this way ensures that we can re-add events
// to the end of the queue during the loop without having them show up in the array.
let mut array_iter = array_iter.take(ready_events.len());
```

tests/utils/libc.rs
```diff
@@ -172,11 +172,23 @@ pub mod epoll {
...
+        // `got` and `expected` might not have the same order and thus the assertion
+        // could fail. To circumvent this, we iterate through all events in `expected`
+        // and add the events from `got` into a new `got_ordered` list in the same order
+        // as `expected`.
+        let mut got_ordered = Vec::new();
+        for ev in expected {
+            if let Some(idx) = got.iter().position(|e| e == ev) {
+                got_ordered.push(got.remove(idx));
+            };
+        }
```
RalfJung: I think this can be simplified. You can get rid of `got_ordered`, emit a panic if `position` returns `None`, and after the loop also panic if there's anything left in `got`.

tests/pass-dep/libc/libc-epoll-blocking.rs
```diff
@@ -50,21 +65,32 @@ fn test_epoll_block_then_unblock() {
...
-    epoll_ctl_add(epfd, fds[0], libc::EPOLLIN | libc::EPOLLOUT | libc::EPOLLET).unwrap();
+    epoll_ctl_add(epfd, fds[0], EPOLLIN | EPOLLOUT | EPOLLET_OR_ZERO).unwrap();
 
     // epoll_wait to clear notification.
-    check_epoll_wait::<1>(epfd, &[Ev { events: libc::EPOLLOUT, data: fds[0] }], 0);
+    check_epoll_wait_noblock::<2>(epfd, &[Ev { events: EPOLLOUT, data: fds[0] }]);
 
-    // epoll_wait before triggering notification so it will block then get unblocked before timeout.
     let thread1 = thread::spawn(move || {
         thread::yield_now();
```
RalfJung, line 74:
```suggestion
        thread::yield_now();
        // Due to deterministic concurrency, we'll only get here when the other thread blocks.
```

tests/pass-dep/libc/libc-epoll-blocking.rs
```diff
@@ -31,16 +38,24 @@ fn test_epoll_block_without_notification() {
...
-    check_epoll_wait::<1>(epfd, &[], 5);
+    check_epoll_wait_noblock::<8>(epfd, &[Ev { events: EPOLLOUT, data: fd }]);
+
+    if cfg!(edge_triggered) {
+        // This epoll wait blocks, and timeout without notification.
+        check_epoll_wait::<1>(epfd, &[], 5);
+    } else {
+        // In level-triggered mode we should receive the same events
+        // as before without timing out.
+        check_epoll_wait_noblock::<8>(epfd, &[Ev { events: EPOLLOUT, data: fd }]);
```
RalfJung: This file isn't really meant to ever call `check_epoll_wait_noblock`, it's the "blocking epoll" file after all.

tests/utils/libc.rs
RalfJung, line 168: As you said there's a footgun here where one makes this `N` not big enough to get all the events we are expecting... so maybe we should just remove this `N` and have the function allocate a `vec![..., expected.len()+1]` to ensure we always know there would have been no more events?

tests/pass-dep/libc/libc-epoll-no-blocking.rs
```diff
@@ -589,24 +577,87 @@ fn test_epoll_ctl_notification() {
...
+    // Write to `fds[1]`.
+    libc_utils::write_all(fds[1], b"abcde").unwrap();
+
+    // Two events should be received.
+    check_epoll_wait_noblock::<3>(
+        epfd,
+        &[Ev { events: EPOLLIN | EPOLLOUT, data: fds[0] }, Ev { events: EPOLLOUT, data: fds[1] }],
+    );
+
+    // Only one event should be received because `fds[0]` is added in edge-triggered mode.
```
RalfJung:
```suggestion
    // If we call epoll_wait again immediately, only the level-triggered interests should be received again.
```

RalfJung (commented):
Looks good overall. :-)
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5071/bba9f7e43428a0046c1faec8e43ce92504f3401a..756c80e188966f7024cab60c6e2089afb8e6839c)*

src/shims/unix/linux_like/epoll.rs
WhySoBad: `array_iter` isn't a real `Iterator`, so `take()` doesn't exist. However, `array_iter.next()` gives a tuple where the first element is the index, so I've now added a check to the `while let Some(...)` loop with `idx < ready_events.len()` which achieves the same.

tests/utils/libc.rs
WhySoBad: I initially implemented it this way, however, I think especially for epoll events it would be nice to see the actual output compared to the expected output.
With the suggested implementation, we would also only get a panic for the first missing event which can be annoying to debug when multiple events are missing.

tests/utils/libc.rs
WhySoBad: I thought about adding this, however, this test would break because we want to test what happens when we don't read all events: https://github.com/rust-lang/miri/blob/b4a15e0a2f5e286e61c3f107d1def8e75061e09a/tests/pass-dep/libc/libc-epoll-no-blocking.rs#L514-L515

Maybe we want to add and expose a `check_epoll_wait_unsafe(epfd, expected, max_events, timeout)` and `check_epoll_wait` and `check_epoll_wait_unblock` call this method with `max_events = expected.len() + 1`?

src/shims/unix/linux_like/epoll.rs
```diff
@@ -590,31 +597,40 @@ fn update_readiness<'tcx>(
...
+            // Either we force an "edge" to be detected or there's a bit set in `new_readiness`
+            // that was not set in `prev_readiness`. In both cases, this is ready now.
+
+            // We need to ensure that this event is not already part of the
+            // `ready_events` queue before enqueueing:
+            // <https://github.com/torvalds/linux/blob/HEAD/fs/eventpoll.c#L1292-L1296>
+            if !ready_events.contains(&key) {
+                ready_events.push_back(key);
+            }
+
```
RalfJung, line 620:
```suggestion
            // No matter whether this is newly ready or just re-triggered,
            // the epoll_wait fetching this event should sync with the current thread.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5071/bba9f7e43428a0046c1faec8e43ce92504f3401a..b4a15e0a2f5e286e61c3f107d1def8e75061e09a)*

tests/pass-dep/libc/libc-epoll-blocking.rs
```diff
@@ -41,15 +41,15 @@ fn test_epoll_block_without_notification() {
     epoll_ctl_add(epfd, fd, EPOLLIN | EPOLLOUT | EPOLLET_OR_ZERO).unwrap();
 
     // epoll_wait to clear notification.
-    check_epoll_wait_noblock::<8>(epfd, &[Ev { events: EPOLLOUT, data: fd }]);
+    check_epoll_wait::<8>(epfd, &[Ev { events: EPOLLOUT, data: fd }], 0);
```
RalfJung: Does the timeout have to be "0" here? Or can we make it -1 as it will any return immediately since there are events to be delivered?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5071/bba9f7e43428a0046c1faec8e43ce92504f3401a..b4a15e0a2f5e286e61c3f107d1def8e75061e09a)*

tests/utils/libc.rs
RalfJung: You can put whatever details into the panic message that you'd like to have there. But that's not an argument for making the logic more complicated than it has to be.

tests/utils/libc.rs
RalfJung: I'd call it `check_epoll_wait_partial` or so.

Maybe it can then also be used for `multiple_events_wake_multiple_threads` where for some reason we still call `epoll_wait` directly?

tests/utils/libc.rs
RalfJung: Another implementation strategy could be
- replace `&[Ev]` by `[Ev; N]`
- sort both `expected` and `got`, and then `assert_eq!` them

WhySoBad:
I'd propose the following solution to the `check_epoll_wait` problem:
```rust
/// Call `epoll_wait` on `epfd` with the provided `timeout`.
/// It fetches at most `N` events from `epfd` and ensures that
/// the returned events match the `expected` events.
#[track_caller]
pub fn check_epoll_wait_explicit<const N: usize, const M: usize>(
    epfd: i32,
    mut expected: [Ev; M],
    timeout: i32,
) {
    assert!(N >= M, "Max event count should be at least as big as expected event count");
    let mut array: [libc::epoll_event; N] = [libc::epoll_event { events: 0, u64: 0 }; N];
    let num = errno_result(unsafe {
        libc::epoll_wait(epfd, array.as_mut_ptr(), N.try_into().unwrap(), timeout)
    })
    .expect("epoll_wait returned an error");
    let got = &mut array[..num.try_into().unwrap()];
    let mut got = got
        .iter()
        .map(|e| Ev { events: e.events.cast_signed(), data: e.u64.try_into().unwrap() })
        .collect::<Vec<_>>();
    // We sort `expected` and `got` to prevent assertion
    // failures due to wrong ordering.
    expected.sort();
    got.sort();
    assert_eq!(got, expected, "got wrong epoll events")
}

/// Call `epoll_wait` on `epfd` with the provided `timeout` and ensure
/// that the returned events match the `expected` events.
/// This function checks whether `expected` is equal to **all** ready events
/// of `epfd`. If you only want to ensure that `expected` matches a subset
/// of the ready events [`check_epoll_wait_explicit`] should be used instead.
#[track_caller]
pub fn check_epoll_wait<const N: usize>(epfd: i32, expected: [Ev; N], timeout: i32) {
    check_epoll_wait_explicit::<{ N + 1 }, N>(epfd, expected, timeout);
}

/// This does the same as [`check_epoll_wait`] just without blocking (zero `timeout`).
#[track_caller]
pub fn check_epoll_wait_noblock<const N: usize>(epfd: i32, expected: [Ev; N]) {
    check_epoll_wait::<N>(epfd, expected, 0);
}
```
The problem is that this now changes the function signature; should I still add this change as part of this PR or would you like to have this in a separate PR?

RalfJung:
I don't think you can do arithmetic in const generics currently, so ` check_epoll_wait_explicit::<{ N + 1 }, N>(...)` doesn't work I think. I'd suggest making `N` a runtime parameter rather than a const generic.

We can delay the signature change to another PR, but the current `got_ordered` thing is too confusing to be landed IMO.

And also... given that we now have an explicit queue, I am getting second thoughts about ignoring the order. Apparently order matters, otherwise we wouldn't use a queue. Where do we currently run into issues if we don't ignore the order?

WhySoBad:
> I don't think you can do arithmetic in const generics currently, so `check_epoll_wait_explicit::<{ N + 1 }, N>(...)` doesn't work I think. I'd suggest making N a runtime parameter rather than a const generic.

Oh yeah, sadly it's a nightly feature. 

> And also... given that we now have an explicit queue, I am getting second thoughts about ignoring the order. Apparently order matters, otherwise we wouldn't use a queue. Where do we currently run into issues if we don't ignore the order?

The test which previously failed on native hosts now also fails on Miri due to the queue. I've now changed the event order in this test and it now works on Miri as well as native hosts. 

@rustbot ready

RalfJung:

@rustbot author

PR #5073, opened 2026-05-28
flock tests: cover the case of asking for both locks on the same file handle

RalfJung:


RalfJung:
Strange, I would have expected this to work on Windows
```rust
    // Test that we can apply many shared locks
    file1.lock_shared().unwrap();
    file2.lock_shared().unwrap();
    // Test that shared lock prevents exclusive lock
    assert!(matches!(file1.try_lock().unwrap_err(), fs::TryLockError::WouldBlock));
    // Unlock one shared lock.
    file2.unlock().unwrap();
    // Now just file1 has a shared lock -- and we can convert that to an exclusive lock.
    file1.try_lock().unwrap(); // <-- fails here
```
But it seems like somehow releasing the lock on file2 does not actually make the lock available for file1? @ChrisDenton any idea what might be happening?

ChrisDenton:
Oh right, sorry, it's just the exclusive lock then shared lock case that works. Doing the opposite doesn't.

> A shared lock can overlap an exclusive lock if both locks were created using the same file handle.

RalfJung:
Wow that's... creative. Thanks for clarifying.

RalfJung:
FWIW I'd argue both locks were "created using the same file handle" in that example... they are both created via file1.

Issue #5065, opened 2026-05-26
Test `aarch64_be`?


folkertdev:
In https://github.com/rust-lang/miri/pull/5061 it is made explicit that for some targets we don't test on big-endian. However, we could run the Miri test suite for `aarch64_be`, a tier 3 big-endian target. 

Despite the tier 3 status, over in `rust-lang/stdarch` we try to keep it working. The target itself works fine, it's really the SIMD operations where subtle endianness issues sometimes pop up.

I believe that Miri already uses s390x, a tier 2 target, to check for endianness bugs. Testing `aarch64_be` is useful primarily to check that the shims for aarch644 SIMD intrinsics are implemented correctly also on BE.

A downside is that 

> libcore might fail to build on that target as we don't test it in rustc CI and I don't want that to impact Miri's CI.

I don't think that's ever happened in my time maintaining `stdarch`, but it might. In practice the only changes that would break this target specifically are:

- an LLVM update
- a stdarch subtree sync

Both of these must pass `rust-lang/rust` CI, which runs the Miri tests, so I think that would guarantee libcore builds?

Even so, the benefits are kind of small, and perhaps it's fine to just manually run the tests from time to time.

RalfJung:
My personal opinion is that we should only test targets that are at least tier 2. We can make an exception for targets that are "on their way towards tier 2" and that already have some basic rustc CI coverage, but I don't want Miri's nightly CI to become a sentinel for entirely unrelated breakage.

RalfJung:
> which runs the Miri tests, so I think that would guarantee libcore builds?

It only runs the Miri tests for a few major targets, much fewer than what we test in this repo.

folkertdev:
What https://github.com/rust-lang/miri/pull/5061 does though is actually disable the shims on `aarch64_be`. Not testing them is one thing, but that change makes it impossible to even run them.

> It only runs the Miri tests for a few major targets, much fewer than what we test in this repo.

If this was the only reason for hesitation we could add `aarch64_be` to the list of targets that builds `libcore`. We really do rely on that working in `stdarch`, it just hasn't broken so far.

RalfJung:
> What https://github.com/rust-lang/miri/pull/5061 does though is actually disable the shims on aarch64_be. Not testing them is one thing, but that change makes it impossible to even run them.

Yeah I disabled them as I had zero confidence in them being correct. We didn't even think of that case when landing them.
I didn't realize anyone relies on them -- sorry for that.

[...]

PR #5075, opened 2026-05-28
allow using shims for `aarch64_be`

folkertdev:
Related to https://github.com/rust-lang/miri/issues/5065, it's fine to just allow shims on `aarch64_be`.

I've also removed some transmutes from the tests, they are very likely to cause endianness issues.

### DIFF
--- a/src/shims/foreign_items.rs
+++ b/src/shims/foreign_items.rs
@@ -850,9 +850,9 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 );
             }
             name if name.starts_with("llvm.aarch64.")
-                && this.tcx.sess.target.arch == Arch::AArch64
-                && this.tcx.sess.target.endian == Endian::Little =>
+                && this.tcx.sess.target.arch == Arch::AArch64 =>
             {
+                // We currently only test on little-endian, but big-endian aarch64 does exist.
                 return shims::aarch64::EvalContextExt::emulate_aarch64_intrinsic(
                     this, link_name, abi, args, dest,
                 );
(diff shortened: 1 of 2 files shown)


src/shims/foreign_items.rs
```diff
@@ -850,21 +850,13 @@ trait EvalContextExtPriv<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 );
             }
             name if name.starts_with("llvm.aarch64.")
-                && this.tcx.sess.target.arch == Arch::AArch64
-                && this.tcx.sess.target.endian == Endian::Little =>
+                && this.tcx.sess.target.arch == Arch::AArch64 =>
             {
+                // We currently only test on little-endian, but big-endian aarch64 does exist.
                 return shims::aarch64::EvalContextExt::emulate_aarch64_intrinsic(
                     this, link_name, abi, args, dest,
```
RalfJung, line 857: So... are the implementations correct on BE? Do we have any way to check that?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5075/1cf48e9be10b8c15ae781a3fe521cd63e1ac1626..7a6f18c33abbca10b76762e630c936237dd31d35)*

folkertdev: > So... are the implementations correct on BE?

Unclear. I believe `stdarch` is kinda broken actually. Which is weird because we run the whole test suite etc but I think we've messed up loads and stores.

When I run the neon shim test file with `qemu` it falls over on the first assert. The test code translated to C works just fine, so `stdarch` is doing something wrong.

RalfJung: In that case I'd prefer to wait until we have confidence in our implementation, before enabling the intrinsics again. Or is that somehow making it harder to fix stdarch?

RalfJung: @rustbot author
based on above comment

PR #5076, opened 2026-05-29
./miri test: run some tests natively

### LINKED ISSUE #3810: Run pass tests against real rustc as well

RalfJung:
Some aspects of this are a bit quirky, but IMO acceptable:
- We ignore all `compile-flags` rather than just the `-Zmiri` ones. Seems fine for now, we don't have many (any?) other flags anyway.
- We do not compare stdout/stderr with the reference files. Seems fine for now, the relevant tests usually use assertions anyway.

I made the fs, socket, and epoll tests run natively in CI.

r? @oli-obk 
Fixes https://github.com/rust-lang/miri/issues/3810

### DIFF
--- a/tests/ui.rs
+++ b/tests/ui.rs
@@ -1,13 +1,15 @@
-use std::env;
+#![allow(clippy::let_and_return)]
 use std::num::NonZero;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::sync::OnceLock;
+use std::{env, fmt};
 
 use colored::*;
 use regex::bytes::Regex;
 use ui_test::build_manager::BuildManager;
 use ui_test::color_eyre::eyre::{Context, Result};
+use ui_test::custom_flags::Flag;
 use ui_test::custom_flags::edition::Edition;
 use ui_test::dependencies::DependencyBuilder;
 use ui_test::per_test_config::TestConfig;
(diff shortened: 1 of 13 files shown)


RalfJung:
Ah, some of our test runners are missing compilers for the "host" (32bit on 64bit).

Also, mkstemp on macOS succeeds when it should return an error...?

tests/ui.rs
```diff
@@ -147,12 +196,77 @@ fn miri_config(
...
+                config: &TestConfig,
+                _output: &std::process::Output,
+                build_manager: &BuildManager,
+            ) -> Result<(), ui_test::Errored> {
+                let mut cmd = native_run_cmd(config.build_command(build_manager)?, &config.config);
+                let exit_status = cmd
+                    .spawn()
+                    .expect("failed to spawn native test")
+                    .wait()
+                    .expect("failed to wait for native test");
```
RalfJung: Unlike `ui_test::Run`, we just spawn the command directly here. Is that a problem?

This makes me wonder if we could just dispatch to `ui_test::Run`... then we also wouldn't need a new ui_test release.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5076/1cf48e9be10b8c15ae781a3fe521cd63e1ac1626..c00fe13a2f183f699df8de63dc8030ed5b121877)*

RalfJung: Yeah using `ui_test::Run` seems to work pretty well. :)

RalfJung:
I think the problem is that we don't honor `CARGO_TARGET_$TARGET_LINKER` when invoking rustc directly. How does one control the linker when invoking rustc directly...?

EDIT: Ah, `-Clinker`. Maybe that helps.

RalfJung:
That seems to work. :)
@rustbot ready

RalfJung:
@oli-obk any chance you could have a look at my ui_test crimes here? Or are you okay with just landing this for now and we can fix it later if needed?

oli-obk (approved):
Hmm yea. This is the best you can do with ui_test rn, I'll take it as an example for thinking about how to improve the API

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5076/c74d8ad72ae8d20b1a95512e728aa4afe38d4123..0980c22eb81faa41e788a68b359f633cf41c0a6f)*

RalfJung:
Honestly it actually worked much better than I expected, the API is quite powerful.

RalfJung:
Where we really hit API limitations is "run_dep" mode. That's a hell of a hack.^^

PR #5083, opened 2026-05-31
Ensure no additional epoll events when using `check_epoll_wait`

WhySoBad:
As previously discussed, with the current implementation of `check_epoll_wait` the `const N: usize` generic is mostly set to `expected.len()` which means we don't ensure that there are no additional ready events other what we specified in `expected`. 

This PR now adds a new `check_epoll_wait_explicit` method, which takes a `max_events` argument. This method can be used when we intentionally don't want to read all ready events. Otherwise, the `check_epoll_wait` (as well as the non-blocking variant) no longer accept the generic `N` and instead call `check_epoll_wait_explicit` with `max_events = expected.len() + 1`. 

The first two commits of this PR also remove the custom `check_epoll_wait` implementations from two other tests.

### DIFF
--- a/tests/fail-dep/libc/libc-epoll-data-race.rs
+++ b/tests/fail-dep/libc/libc-epoll-data-race.rs
@@ -48,7 +48,7 @@ fn main() {
     thread::yield_now();
 
     // With room for one event: check result from epoll_wait.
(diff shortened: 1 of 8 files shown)


RalfJung:
LGTM, thanks. :)
I ended up renaming `check_epoll_wait_explicit` to `check_epoll_wait_partial` after thinking about the name a bit more.

PR #5088, opened 2026-06-03
test that even ZST references prevent deallocation via child pointers

RalfJung:


tests/fail/both_borrows/deallocate_against_protector2.rs
RalfJung, line 1: @JoJoDeveloping as far as I can tell, this fails in TB because just before deallocation, we do a write access with the pointer, which marks the memory range as accessed for this pointer and its parents, which also activates the strong protector, which then leads to UB as we violate said protector. Does that sound right?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5088/e5bd4a843abdbad43fd4c3ff4396b654e7b0a95f..d58db5ea6eb7d295d1ec9285184a2aed4f121992)*

PR #5091, opened 2026-06-05
move cron job earlier

RalfJung:
There seems to be a around 1.5h delay until the cron job actually starts. So let's move it earlier in the hope that then it is done by 6am UTC.

Cc @rust-lang/infra in case you have thoughts / know more about this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
     branches:
       - 'master'
   schedule:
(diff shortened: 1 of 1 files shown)


Kobzol:
Looking at https://github.com/rust-lang/miri/actions?query=event%3Aschedule, it seems to be quite random when the job starts. Seems like it's a known issue: https://github.com/orgs/community/discussions/156282

RalfJung:
Oh yeah that graph is... concerning.
Well let's see if moving it by half an hour does anything.

PR #5097, opened 2026-06-07
triagebot: remove assign.custom_welcome_messages

RalfJung:
With the nice diff links that we have now, not force-pushing has become less critical. So we don't need this reminder in every PR any more I think.

rust-lang/rust also does not have this section in their triagebot config.


RalfJung:
(sorry, misclick)

PR #5098, opened 2026-06-07
Move epoll readiness logic into `ReadinessWatcher`s

WhySoBad:
Hi,

This pull request adds a new `ReadinessManager` which receives readiness updates from file descriptions or the blocking I/O manager and then notifies interested `ReadinessConsumers` about the changed readiness.

Previously, all this logic was directly integrated into the epoll shim. In order to enable implementations for `poll`, `select`, etc. shims the logic is now outsourced into the readiness manager and the epoll shim is just another `ReadinessConsumer`. 

We also no longer have different types of readiness (`EpollReadiness`, `BlockingIoReadiness`) and instead everything just uses the new `Readiness` struct from the readiness manager.

### DIFF
--- a/src/shims/readiness.rs
+++ b/src/shims/readiness.rs
@@ -1,0 +1,532 @@
+use std::cell::{Ref, RefCell};
+use std::collections::{BTreeMap, VecDeque};
+use std::rc::{Rc, Weak};
+
+use crate::concurrency::VClock;
+use crate::shims::files::{DynFileDescriptionRef, FdNum};
+use crate::shims::*;
+use crate::*;
+
+/// Struct reflecting the readiness of a file description.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Readiness {
+    /// Boolean whether the file description is readable.
+    pub readable: bool,
+    /// Boolean whether the file description is writable.
+    pub writable: bool,
+    /// Boolean whether the read end of the file description
+    /// is closed.
+    pub read_closed: bool,
+    /// Boolean whether the write end of the file description
+    /// is closed.
+    pub write_closed: bool,
+    /// Boolean whether the file description has an error.
+    pub error: bool,
+}
+
+impl std::ops::BitAnd for Readiness {
+    type Output = Readiness;
+
+    fn bitand(self, rhs: Readiness) -> Self::Output {
+        Readiness {
+            readable: self.readable && rhs.readable,
+            writable: self.writable && rhs.writable,
+            read_closed: self.read_closed && rhs.read_closed,
+            write_closed: self.write_closed && rhs.write_closed,
+            error: self.error && rhs.error,
+        }
+    }
+}
+
+impl std::ops::BitOrAssign for Readiness {
+    fn bitor_assign(&mut self, rhs: Self) {
+        self.readable |= rhs.readable;
+        self.writable |= rhs.writable;
+        self.read_closed |= rhs.read_closed;
+        self.write_closed |= rhs.write_closed;
+        self.error |= rhs.error;
+    }
+}
+
+impl Readiness {
+    pub const EMPTY: Readiness = Readiness {
+        readable: false,
+        writable: false,
+        read_closed: false,
+        write_closed: false,
+        error: false,
+    };
+}
+
+pub type ReadinessInterestKey = (FdId, FdNum);
+
+/// Returns the range of all [`ReadinessInterestKey`] for the given FD ID.
+fn range_for_id(id: FdId) -> std::ops::RangeInclusive<ReadinessInterestKey> {
+    (id, 0)..=(id, FdNum::MAX)
+}
+
+#[derive(Debug, Clone)]
+pub struct ReadinessInterest {
+    /// The mask of events the interest is interested in
+    /// for this file descriptor.
+    pub relevant: Readiness,
+    /// Boolean whether this is an edge-triggered interest.
+    /// When [`false`] it's a level-triggered interest instead.
+    pub is_edge_triggered: bool,
+    /// Data attached to the interest.
+    // FIXME: In the future we might want to support more data types,
+    // then this should no longer be a `u64` but a `dyn Any` instead.
+    pub data: u64,
+    /// The currently active readiness for this file descriptor.
+    active: Readiness,
+    /// The vector clock for wakeups.
+    clock: VClock,
+}
+
+impl ReadinessInterest {
+    pub fn active(&self) -> &Readiness {
+        &self.active
+    }
+
+    pub fn clock(&self) -> &VClock {
+        &self.clock
+    }
+}
+
+/// A struct which stores [`ReadinessInterest`]s for a set of file descriptions
+/// together with which interests are currently satisfied, and a list of
+/// threads which should be unblocked once a [`ReadinessInterest`] of the
+/// watcher is fulfilled.
+#[derive(Debug)]
+pub struct ReadinessWatcher {
+    /// Globally unique identifier of the watcher.
+    id: usize,
+    /// A map of [`ReadinessInterest`]s registered for this watcher. Each entry is
+    /// identified using a [`FdId`] [`FdNum`] tuple.
+    interests: RefCell<BTreeMap<ReadinessInterestKey, ReadinessInterest>>,
+    /// The subset of interests that is currently considered "ready". Stored separately so we
+    /// can access it more efficiently.
+    /// This is implemented as a queue so that with level-triggered interests, all events eventually
+    /// get returned from [`ReadinessWatcher::get_ready_interests`]. The queue does not contain any
+    /// duplicates.
+    ready: RefCell<VecDeque<ReadinessInterestKey>>,
+    /// The queue of threads blocked on this watcher.
+    queue: RefCell<VecDeque<ThreadId>>,
+}
+
+impl ReadinessWatcher {
+    /// Get a reference to the map of registered interests of the watcher
+    /// together with their keys.
+    pub fn interests(&self) -> Ref<'_, BTreeMap<ReadinessInterestKey, ReadinessInterest>> {
+        self.interests.borrow()
+    }
+
+    /// Add an interest for the file description to which the file descriptor
+    /// `fd_num` belongs.
+    /// `relevant` contains the readiness mask of relevant events.
+    /// `is_edge_triggered` specifies whether the interest is edge-triggered
+    /// ([`true`]) or level-triggered ([`false`]).
+    /// `data` is the user-data which is associated with the interest.
+    ///
+    /// The function returns `Ok(())` when the interest was successfully
+    /// added, and `Err(())` when an interest with this key was already registered.
+    pub fn add_interest<'tcx>(
+        self: &Rc<Self>,
+        fd_num: FdNum,
+        relevant: Readiness,
+        is_edge_triggered: bool,
+        data: u64,
+        ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, Result<(), ()>> {
+        let fd_ref = ecx.machine.fds.get(fd_num).expect("File description should exist");
+        let fd_id = fd_ref.id();
+        let key = (fd_id, fd_num);
+
+        let interest = ReadinessInterest {
+            active: Readiness::EMPTY,
+            clock: VClock::default(),
+            relevant,
+            is_edge_triggered,
+            data,
+        };
+        let mut interests = self.interests.borrow_mut();
+        if interests.range(range_for_id(fd_id)).next().is_none() {
+            // This is the first time this FD got added to the watcher.
+            // We need to remember that in the global list such that we
+            // get notified about FD events.
+            ecx.machine.readiness_interests.insert(fd_id, self);
+        }
+        if interests.try_insert(key, interest).is_err() {
+            return interp_ok(Err(()));
+        }
+
+        // After adding a new interest for a fd, we need to forcefully update
+        // the readiness of this fd.
+
+        ecx.update_readiness(
+            self,
+            fd_ref.readiness()?,
+            /* force_edge */ true,
+            move |callback| {
+                // Need to release the RefCell when this closure returns, so we have to move
+                // it into the closure, so we have to do a re-lookup here.
+                callback(key, interests.get_mut(&key).unwrap())
+            },
+        )?;
+
+        interp_ok(Ok(()))
+    }
+
+    /// Update the interest which is registered for `key`.
+    /// `cb` gets invoked with a mutable reference to the registered
+    /// [`ReadinessInterest`].
+    ///
+    /// This function returns [`None`] when no interest is registered
+    /// for the specified `key`.
+    pub fn update_interest<'tcx>(
+        self: &Rc<Self>,
+        key: ReadinessInterestKey,
+        ecx: &mut MiriInterpCx<'tcx>,
+        cb: impl FnOnce(&mut ReadinessInterest),
+    ) -> InterpResult<'tcx, Option<()>> {
+        let mut interests = self.interests.borrow_mut();
+        let Some(interest) = interests.get_mut(&key) else { return interp_ok(None) };
+        cb(interest);
+
+        // After updating an interest for a fd, we need to forcefully update
+        // the readiness of this fd.
+
+        let fd_ref = ecx.machine.fds.get(key.1).expect("File description should exist");
+        ecx.update_readiness(
+            self,
+            fd_ref.readiness()?,
+            /* force_edge */ true,
+            move |callback| {
+                // Need to release the RefCell when this closure returns, so we have to move
+                // it into the closure, so we have to do a re-lookup here.
+                callback(key, interests.get_mut(&key).unwrap())
+            },
+        )?;
+
+        interp_ok(Some(()))
+    }
+
+    /// Remove the interest registered for `key`.
+    ///
+    /// This function returns [`None`] when no interest is registered
+    /// for the specified `key`.
+    pub fn remove_interest<'tcx>(
+        self: &Rc<ReadinessWatcher>,
+        key: ReadinessInterestKey,
+        ecx: &mut MiriInterpCx<'tcx>,
+    ) -> Option<()> {
+        self.interests.borrow_mut().remove(&key)?;
+
+        // Remove the ready event for this key, should one exist.
+        let mut ready = self.ready.borrow_mut();
+        if let Some(idx) = ready.iter().position(|k| k == &key) {
+            ready.remove(idx);
+        }
+
+        let is_last = self.interests.borrow().range(range_for_id(key.0)).next().is_none();
+        if is_last {
+            // If this was the last interest in this FD, we remove the watcher from
+            // the global list of who is interested in this FD.
+            ecx.machine.readiness_interests.remove(key.0, self);
+        }
+
+        Some(())
+    }
+
+    /// Add the thread with id `thread_id` to the queue of
+    /// blocked threads which will be unblocked when the
+    /// watcher becomes ready.
+    pub fn add_blocked_thread(&self, thread_id: ThreadId) {
+        self.queue.borrow_mut().push_back(thread_id);
+    }
+
+    /// Remove all threads with id `thread_id` from the queue
+    /// of blocked threads which will be unblocked when the
+    /// watcher becomes ready.
+    pub fn remove_blocked_thread(&self, thread_id: ThreadId) {
+        self.queue.borrow_mut().retain(|id| id != &thread_id);
+    }
+
+    /// Get the amount of interests which are registered to this
+    /// watcher and which are currently ready.
+    pub fn ready_count(&self) -> usize {
+        self.ready.borrow().len()
+    }
+
+    /// Get at most the first `count` ready interests from the ready queue.
+    ///
+    /// If the interest is a level-triggered interest, it's automatically
+    /// added to the end of the queue again such that it will only be reported
+    /// after all other ready interest have been returned.
+    ///
+    /// This method returns at most every event from the ready queue once.
+    /// This ensures that every returned interest is unique, even when there
+    /// are level-triggered interests.
+    pub fn get_ready_interests<'tcx>(
+        &self,
+        count: usize,
+        ecx: &mut MiriInterpCx<'tcx>,
+    ) -> InterpResult<'tcx, Vec<ReadinessInterest>> {
+        let interests = self.interests.borrow();
+        let mut ready = self.ready.borrow_mut();
+        let count = count.min(ready.len());
+        let mut ready_interests = Vec::with_capacity(count);
+
+        // Sanity-check to ensure that all event info is up-to-date.
+        if cfg!(debug_assertions) {
+            for (key, interest) in interests.iter() {
+                // Ensure this matches the latest readiness of this FD.
+                // We have to do an FD lookup by ID for this. The FdNum might be already closed.
+                let fd = ecx.machine.fds.fds.values().find(|fd| fd.id() == key.0).unwrap();
+                let current_active = fd.readiness()?;
+                assert_eq!(interest.active(), &(current_active & interest.relevant.clone()));
+            }
+        }
+
+        // We iterate over the first `count` entries from the `ready` queue
+        // and push the associated interests into the `ready_interests` list.
+        // If an interest is level-triggered, it's also added to the back of
+        // the queue again.
+        while ready_interests.len() < count
+            && let Some(key) = ready.pop_front()
+        {
+            let interest = interests.get(&key).expect("non-existing interest in ready set");
+
+            if !interest.is_edge_triggered {
+                // This is a level-triggered interest, so we need to re-add the event
+                // at the end of the ready queue like Linux does with epoll:
+                // <https://github.com/torvalds/linux/blob/HEAD/fs/eventpoll.c#L1835-L1847>
+                ready.push_back(key);
+            }
+
+            ready_interests.push(interest.clone());
+        }
+
+        interp_ok(ready_interests)
+    }
+
+    /// Destroy the watcher instance.
+    ///
+    /// This also deregisters all interests of the watcher
+    /// from the global readiness interest table.
+    pub fn destroy<'tcx>(self, ecx: &mut MiriInterpCx<'tcx>) {
+        // If we were interested in some FDs, we can remove that now.
+        let mut ids = self.interests.borrow().keys().map(|(id, _num)| *id).collect::<Vec<_>>();
+        // Because the ids come out of the map sorted,
+        // deduping only keeps all unique entries.
+        ids.dedup();
+        for id in ids {
+            ecx.machine.readiness_interests.remove(id, &self);
+        }
+    }
+}
+
+impl PartialEq for ReadinessWatcher {
+    fn eq(&self, other: &Self) -> bool {
+        self.id == other.id
+    }
+}
+
+impl Eq for ReadinessWatcher {}
+
+impl VisitProvenance for ReadinessWatcher {
+    fn visit_provenance(&self, _visit: &mut VisitWith<'_>) {}
+}
+
+/// The table with all [`ReadinessWatcher`]s.
+/// This tracks, for each file description, which watchers have an interest in events
+/// for this file description.
+pub struct ReadinessInterestTable {
+    /// The id of the next [`ReadinessWatcher`] created through
+    /// [`ReadinessInterestTable::new_watcher`].
+    next_watcher_id: usize,
+    /// Maps each file description (identified by its [`FdId`]) to the list of watchers that are
+    /// interested in that FD. We also store the [`ReadinessWatcher`]s ID
+    /// separately so we can access it without calling `upgrade`. The list
+    /// is sorted by that id.
+    interests: BTreeMap<FdId, Vec<(usize, Weak<ReadinessWatcher>)>>,
+}
+
+impl ReadinessInterestTable {
+    pub(crate) fn new() -> Self {
+        ReadinessInterestTable { interests: BTreeMap::new(), next_watcher_id: 0 }
+    }
+
+    /// Create a new [`ReadinessWatcher`] with a globally unique id.
+    /// Every watcher gets a sequentially increasing id such that no two
+    /// watchers ever get the same id.
+    pub fn new_watcher(&mut self) -> ReadinessWatcher {
+        let id = self.next_watcher_id;
+        self.next_watcher_id = id.strict_add(1);
+        ReadinessWatcher {
+            id,
+            interests: RefCell::new(BTreeMap::new()),
+            ready: RefCell::new(VecDeque::new()),
+            queue: RefCell::new(VecDeque::new()),
+        }
+    }
+
+    /// Add an interest for `watcher` for the file description with id `fd_id`.
+    fn insert(&mut self, fd_id: FdId, watcher: &Rc<ReadinessWatcher>) {
+        let watchers = self.interests.entry(fd_id).or_default();
+        let idx = watchers
+            .binary_search_by_key(&watcher.id, |&(id, _)| id)
+            .expect_err("watcher already has a registered interest in the provided fd");
+        watchers.insert(idx, (watcher.id, Rc::downgrade(watcher)));
+    }
+
+    /// Remove the interest of `watcher` for the file description with id `fd_id`.
+    fn remove(&mut self, fd_id: FdId, watcher: &ReadinessWatcher) {
+        let watchers = self.interests.entry(fd_id).or_default();
+        let idx = watchers
+            .binary_search_by_key(&watcher.id, |&(id, _)| id)
+            .expect("watcher has no registered interest in the provided fd");
+        watchers.remove(idx);
+    }
+
+    /// Get all watchers which have a registered interest in the file description
+    /// with id `fd_id`.
+    fn get_watchers_for_fd(&mut self, fd_id: &FdId) -> Option<Vec<Rc<ReadinessWatcher>>> {
+        let watchers = self.interests.get_mut(fd_id)?;
+        Some(
+            watchers
+                .iter()
+                .map(|(_id, watcher)| {
+                    watcher.upgrade().expect(
+                        "someone forgot to remove the garbage from `machine.readiness_interests`",
+                    )
+                })
+                .collect(),
+        )
+    }
+
+    /// Remove all watchers for the file description with id `fd_id`.
+    pub fn remove_watchers_for_fd(&mut self, fd_id: FdId) {
+        let Some(watchers) = self.interests.remove(&fd_id) else {
+            return;
+        };
+
+        for watcher in watchers.iter().filter_map(|(_id, watcher)| Weak::upgrade(watcher)) {
+            // This is a still-live watcher with interest in this FD. Remove all
+            // relevant interests (including from the ready set).
+            watcher
+                .interests
+                .borrow_mut()
+                .extract_if(range_for_id(fd_id), |_, _| true)
+                // Consume the iterator.
+                .for_each(drop);
+            // Remove the ready interests for this file description.
+            watcher.ready.borrow_mut().retain(|(id, _)| id != &fd_id);
+        }
+    }
+}
+
+impl<'tcx> EvalContextExt<'tcx> for MiriInterpCx<'tcx> {}
+pub trait EvalContextExt<'tcx>: MiriInterpCxExt<'tcx> {
+    /// Recursively check whether the [`ReadinessWatcher`] contains
+    /// interests which are host I/O source file descriptions.
+    fn has_watcher_host_interests(&self, watcher: &ReadinessWatcher) -> bool {
+        let this = self.eval_context_ref();
+        watcher.interests().keys().any(|(fd_id, _fd_num)| {
+            // By looking up whether the file description is currently registered,
+            // we get whether it's a host I/O source file description.
+            this.machine.blocking_io.contains_source(fd_id)
+        })
+    }
+
+    /// For a specific file description, get its current readiness and send it to everyone who
+    /// registered interest in this FD. This function must be called whenever the result of
+    /// `FileDescription::readiness` might change.
+    ///
+    /// If `force_edge` is set, edge-triggered interests will be triggered even if the set of
+    /// ready events did not change. This can lead to spurious wakeups. Use with caution!
+    fn update_fd_readiness(
+        &mut self,
+        fd: DynFileDescriptionRef,
+        force_edge: bool,
+    ) -> InterpResult<'tcx> {
+        let this = self.eval_context_mut();
+        let fd_id = fd.id();
+
+        let Some(watchers) = this.machine.readiness_interests.get_watchers_for_fd(&fd_id) else {
+            return interp_ok(());
+        };
+        let active_readiness = fd.readiness()?;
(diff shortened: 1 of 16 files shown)


src/concurrency/readiness.rs
```diff
@@ -102,7 +103,7 @@ pub struct ReadinessManager {
     /// The id with which the next readiness consumer will be registered.
     next_consumer_id: usize,
     /// The set of registered readiness consumers indexed by their ids.
-    consumers: BTreeMap<ReadinessConsumerId, Box<dyn ReadinessConsumer>>,
+    consumers: BTreeMap<ReadinessConsumerId, Rc<Box<dyn ReadinessConsumer>>>,
```
RalfJung: `Rc<Box<T>>` makes basically never sense, it should just be `Rc<T>` then.

But why is this even needed? This is introduced in a "fix" commit but without any tests or comments explaining the problem it is fixing.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5098/2265f0f03bbd704930af7c488e1a611d64e69489..8c17e77d999147cd204a49abf68fbed407d9efa7)*

src/concurrency/readiness.rs
RalfJung, line 1: Please add a module-level comment explaining the high-level overview of what happens in this file.

src/concurrency/readiness.rs
```diff
@@ -0,0 +1,205 @@
...
+        writable: false,
+        read_closed: false,
+        write_closed: false,
+        error: false,
+    };
+}
+
+/// A trait which can be implemented for things which need to receive notifications when
+/// the readiness of file descriptions change.
+pub trait ReadinessConsumer {
```
RalfJung: Why do we need a new trait here?

I would expect the first PR to just move the logic and make it non-Linux-specific, but it should not contain any overeager generalization. Doing both the moving of logic and changing of logic in a single PR is going to make this extremely hard to review.

src/concurrency/readiness.rs
```diff
@@ -0,0 +1,205 @@
...
+    fn bitor_assign(&mut self, rhs: Self) {
+        self.readable |= rhs.readable;
+        self.writable |= rhs.writable;
+        self.read_closed |= rhs.read_closed;
+        self.write_closed |= rhs.write_closed;
+        self.error |= rhs.error;
+    }
+}
+
+impl AsRef<Self> for Readiness {
```
RalfJung: What is this for?

src/concurrency/readiness.rs
```diff
@@ -0,0 +1,205 @@
...
+
+/// Struct storing the interests in file description readiness events.
+pub struct ReadinessManager {
+    /// The id with which the next readiness consumer will be registered.
+    next_consumer_id: usize,
+    /// The set of registered readiness consumers indexed by their ids.
+    consumers: BTreeMap<ReadinessConsumerId, Box<dyn ReadinessConsumer>>,
+    /// The registered interests indexed by file description ids.
+    interests: BTreeMap<FdId, BTreeSet<ReadinessConsumerId>>,
+}
```
RalfJung: This looks completely different than the `Epoll` type did. Why? Please keep the logic the same, just move it.

src/shims/unix/linux_like/epoll.rs
RalfJung: I expected this to mostly disappear here since it is now managed by the general OS-independent logic.

src/shims/unix/linux_like/eventfd.rs
```diff
@@ -228,7 +220,7 @@ fn eventfd_write<'tcx>(
             // Linux seems to cause spurious wakeups here, and Tokio seems to rely on that
             // (see <https://github.com/rust-lang/miri/pull/4676#discussion_r2510528994>
             // and also <https://www.illumos.org/issues/16700>).
-            ecx.update_epoll_active_events(eventfd, /* force_edge */ true)?;
+            ecx.notify_fd_readiness(eventfd, /* force_edge */ true)?;
```
RalfJung: `update_fd_readiness` would be a better name IMO.

RalfJung (commented):
Thanks! However, this contains way bigger changes than I expected, without sufficient motivation or high-level explanation of the new design.

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5098/2265f0f03bbd704930af7c488e1a611d64e69489..8c17e77d999147cd204a49abf68fbed407d9efa7)*

src/concurrency/readiness.rs
RalfJung: Also I think this file should live in `shims`, not `concurrency`. It is still more about implementing shims than about core interpreter concurrency. I also expect it will be directly tied to FDs which generally should only be the case for things inside `shims`.

src/concurrency/readiness.rs
WhySoBad: The problem was that `notify_fd_readiness` previously removed the interests and the consumers before calling `consumer.ready_event` and only re-inserted them back afterwards.

This made it impossible to deregister a consumer from within a `ready_event` implementation because at the time of calling it, the consumer wasn't stored in the list anymore (and thus temporarily "deregistered").

src/concurrency/readiness.rs
WhySoBad: `BitAnd` is implemented for `&Readiness` and writing `(&readiness1) & (&readiness2)` is a bit awkward to write, so I chose to implement it there such that we can have `readiness1.as_ref() & readiness2.as_ref()` instead.

src/concurrency/readiness.rs
WhySoBad: I don't fully understand what you mean. Moving the logic means that we still need some way of notifying the `Epoll` instances about new readiness events, that's what this trait is for.

src/shims/unix/linux_like/epoll.rs
WhySoBad: Is your expectation that the `ReadinessManager` stores a `Readiness` for every active FD? With the current implementation, I've chose to make the `ReadinessManager` relatively stateless, meaning it only stores which consumers are interested in events for which FDs and as soon as the manager receives such an event it notifies the consumer.

src/concurrency/readiness.rs
RalfJung: > BitAnd is implemented for &Readiness

Sounds like that is the problem -- implement it for `Readiness` instead.

src/shims/unix/linux_like/epoll.rs
RalfJung: I expected something along the lines of
- Most of `Epoll` gets factored into a `ReadinessWatcher`, so that `Epoll` itself just contains a `Rc<ReadinessWatcher>`. (We need an `Rc` so that we can have weak refs to it. It's going to be the only strong ref.)
- `EpollInterestTable` is renamed to `ReadinessInterestTable` and stores `Weak<ReadinessWatcher>`.

My hope is that we don't need the `FdId` of the epoll instance itself -- we can lazily remove dead entries from the `ReadinessInterestTable` when their weak refs are found to be dead.

WhySoBad: > My hope is that we don't need the FdId of the epoll instance itself 

Yes, I want to avoid that as well. Especially when thinking about implementing `poll` which hasn't an associated `FdId`. 

So, basically, your idea is that a consumer (e.g. epoll or poll) has one `ReadinessWatcher` which contains a map of `FdId, Readiness` of interests for the consumer? The `ReadinessInterestTable` would then be a list of `Weak<ReadinessWatcher>` and as soon as a new event is received we iterate over this list and somehow notify the consumer about the new event.

Wouldn't we also need a new trait for this scenario? I'm a bit hung up on how we want to notify the consumer about new events.

WhySoBad:
As discussed, the force-pushed commits now add a new `ReadinessWatcher` struct which basically does the same the `Epoll` struct did -- but just more generalized such that also different shims (like `poll`) can reuse this logic. 

The active `ReadinessWatcher`s are stored in the global `ReadinessInterestTable` on the Miri machine. 

One thing which slightly differs from what we discussed is that the `ReadinessWatcher`s have an id. This is needed as `BlockReason` is required to implement `PartialEq` since otherwise we can't check whether a thread was blocked for the same reason with which we unblocked it. The id of those watchers is fully private and is just a strictly monotonically increasing number to ensure uniqueness.

src/shims/readiness.rs
```diff
@@ -0,0 +1,480 @@
...
+    pub fn active(&self) -> &Readiness {
+        &self.active
+    }
+
+    pub fn clock(&self) -> &VClock {
+        &self.clock
+    }
+}
+
+/// A struct which stores [`ReadinessInterest`]s for a file description
```
RalfJung:
```suggestion
/// A struct which stores [`ReadinessInterest`]s for a set of file descriptions
```

src/shims/readiness.rs
```diff
@@ -0,0 +1,480 @@
...
+}
+
+impl ReadinessWatcher {
+    /// Get a reference to the map of registered interests of the watcher
+    /// together with their keys.
+    pub fn interests(&self) -> Ref<'_, BTreeMap<ReadinessInterestKey, ReadinessInterest>> {
+        self.interests.borrow()
+    }
+
+    /// Add an interest for the with `key`.
```
RalfJung: A word seems to be missing here.

Also that's pretty low-level, the caller could screw up and pass an FdId + FdNum that don't actually match. You could probably pass in just the FdNum, and query the FdId from `ecx`?

src/shims/readiness.rs
```diff
@@ -0,0 +1,480 @@
...
+        let is_first = interests.range(range_for_id(key.0)).next().is_none();
+        if interests.try_insert(key, interest).is_err() {
+            return interp_ok(Err(()));
+        }
+        if is_first {
+            // This is the first time this FD got added to the watcher.
+            // We need to remember that in the global list such that we
+            // get notified about FD events.
+            ecx.machine.readiness_interests.insert(key.0, self);
+        }
```
RalfJung: The original code didn't need a separate `is_first` variable here. You reordered things a bit. Why?

src/shims/readiness.rs
```diff
@@ -0,0 +1,480 @@
...
+    /// `cb` gets invoked with a mutable reference to the registered
+    /// [`ReadinessInterest`].
+    ///
+    /// This function returns [`None`] when no interest is registered
+    /// for the specified `key`.
+    pub fn update_interest<'tcx>(
+        self: &Rc<Self>,
+        key: ReadinessInterestKey,
+        cb: impl FnOnce(&mut ReadinessInterest),
+        ecx: &mut MiriInterpCx<'tcx>,
```
RalfJung: Closures should typically be the last argument.

src/shims/readiness.rs
```diff
@@ -0,0 +1,480 @@
...
+
+/// The table with all [`ReadinessWatcher`]s.
+/// This tracks, for each file description, which watchers have an interest in events
+/// for this file description.
+pub struct ReadinessInterestTable {
+    /// The id of the next [`ReadinessWatcher`] created through
+    /// [`ReadinessInterestTable::new_watcher`].
+    next_watcher_id: usize,
+    /// Map with registered watchers indexed by the file description
+    /// they're interested in.
```
RalfJung:
```suggestion
    /// Maps each file description (identified by its ID) to the list of watchers that are
    /// interested in that FD.
```

src/shims/readiness.rs
```diff
@@ -0,0 +1,480 @@
...
+        }
+        watchers.push(Rc::downgrade(watcher));
+    }
+
+    /// Remove the interest of `watcher` for the file description with id `fd_id`.
+    pub fn remove(&mut self, fd_id: FdId, watcher: &Rc<ReadinessWatcher>) {
+        let watchers = self.interests.entry(fd_id).or_default();
+        let idx = watchers
+            .iter()
+            .position(|stored| stored.upgrade().is_some_and(|stored| &stored == watcher));
```
RalfJung: In the `EpollInterestTable`, the `Vec` was sorted to avoid a linear scan here. Seems worth keeping that.

src/shims/readiness.rs
```diff
@@ -0,0 +1,480 @@
...
+
+        Some((next, interest))
+    }
+}
+
+impl PartialEq for ReadinessWatcher {
+    fn eq(&self, other: &Self) -> bool {
+        self.id == other.id
+    }
+}
```
RalfJung, line 335: Do we need this often enough to make it worth it? In the code below you can just compare the ID fields.

src/shims/readiness.rs
```diff
@@ -0,0 +1,480 @@
...
+                this.release_clock(|clock| {
+                    interest.clock.join(clock);
+                })?;
+            }
+            interp_ok(())
+        })?;
+
+        // While there are events ready to be delivered, wake up a thread to receive them.
+        while !ready.is_empty()
+            && let Some(thread_id) = queue.pop_front()
```
RalfJung: If you follow the original code and just use ` watcher.queue.borrow_mut().pop_front()` here, everything will become a bit simpler.

Is there anything else where you changed the original epoll logic? Please explicitly indicate any such change -- the PR *should* be just a code move without any logic change.

src/shims/unix/linux_like/epoll.rs
```diff
@@ -1,32 +1,26 @@
...
-    /// The queue of threads blocked on this epoll instance.
-    queue: RefCell<VecDeque<ThreadId>>,
+    /// Watcher used for registering interests in the global readiness
+    /// interest table.
+    watcher: Rc<ReadinessWatcher>,
+    /// Map from interests to their user-defined data.
+    /// libc's `data` field in `epoll_event` can store integer or pointer,
+    /// but only [`u64`] is supported for now.
+    /// <https://man7.org/linux/man-pages/man3/epoll_event.3type.html>
+    interest_data: RefCell<BTreeMap<ReadinessInterestKey, u64>>,
```
RalfJung: I think it's reasonable to also move the `data` field into the general infrastructure. Attaching extra data to an interest makes sense beyond just epoll. Maybe we'll eventually want to make that a `dyn Any` or so but for now it might just as well be a `u64`.

src/shims/unix/linux_like/epoll.rs
```diff
@@ -144,16 +42,16 @@ impl FileDescription for Epoll {
...
         ecx: &mut MiriInterpCx<'tcx>,
     ) -> InterpResult<'tcx, io::Result<()>> {
         // If we were interested in some FDs, we can remove that now.
-        let mut ids = self.interest_list.get_mut().keys().map(|(id, _num)| *id).collect::<Vec<_>>();
+        let mut ids = self.watcher.interests().keys().map(|(id, _num)| *id).collect::<Vec<_>>();
         ids.dedup(); // they come out of the map sorted
         for id in ids {
-            ecx.machine.epoll_interests.remove(id, self_id);
+            ecx.machine.readiness_interests.remove(id, &self.watcher);
         }
```
RalfJung: This should also be a method on `watcher`, like `fn destroy` or so. You can maybe even make that take `self`, and use `Rc::into_inner`.

src/shims/unix/linux_like/epoll.rs
```diff
@@ -531,161 +360,113 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-            // at the end of the ready queue:
-            // <https://github.com/torvalds/linux/blob/HEAD/fs/eventpoll.c#L1835-L1847>
-            ready_events.push_back(key);
+        // We will fill at most the first `ready_events_len` slots of the array.
+        // Bounding the iterator this way ensures that we can re-add events
+        // to the end of the queue during the loop without having them show up in the array.
+        let ready_events_len = u64::try_from(epfd.watcher.ready_count()).unwrap();
+        while let Some((idx, slot)) = array_iter.next(this)?
+            && idx < ready_events_len
+            && let Some((key, interest)) = epfd.watcher.next_ready()
```
RalfJung: This API with ready_count and next_ready seems a bit awkward to use.

We could maybe just have a `get_ready` method on `watcher` that takes an `n: usize` and returns a `Vec` of ready interests?

RalfJung (commented):
Yes that is much closer to the structure I was imagining, thanks :)
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5098/97eec7cf0c7ba44683e718869038f53a4b0931fb..ac73f473e4c290a8f802bf1d03cbc98b227f8f97)*

src/shims/readiness.rs
WhySoBad: I had a logic error where the current implementation would "leak" the registered interest in the table should the interest already be registered. However, when the interest is already registered then the `.next()` is always `Some(..)` so we don't need to worry about this case.

src/shims/readiness.rs
WhySoBad: If it wasn't for the `BlockReason` enum, we could remove this implementation. Theoretically, we could also just implement `PartialEq` and `Eq` manually for `BlockReason` but I think it's nicer to implement it for `ReadinessWatcher` such that we can derive it on `BlockReason`.

src/shims/readiness.rs
WhySoBad: No, there should be no changes to the original epoll logic.

RalfJung: Well, there were at least two that I noticed in my review (this one here, and the one about `let is_first`). Not sure how that happened, you must have deliberately edited the code after copying it -- so please let me know about any other places that you made similar edits.

WhySoBad: Yes, some code is edited to make it more readable. Another case is the `is_last` which is now a separate variable; but the logic is still the same. I've also changed `get_watchers_for_fd` (the replacement for `get_epolls`) to not return `&Weak<...>` and instead directly perform the upgrade itself. 
Otherwise, I've again looked at the original epoll implementation and how it's now implemented, and logically everything's the same.

src/shims/unix/linux_like/epoll.rs
```diff
@@ -424,13 +398,19 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
 
-        let interests = epfd.watcher.interests();
         let mut num_of_events = 0i32;
         let mut array_iter = this.project_array_fields(events)?;
 
+        #[expect(rustc::usage_of_ty_tykind)]
+        let rustc_middle::ty::TyKind::Array(_, events_len) = events.layout.ty.kind() else {
+            unreachable!("`events` needs to be an array type")
+        };
+        let max_events_num: usize =
```
RalfJung: You can just do `events.len(this)` :)

src/shims/unix/linux_like/epoll.rs
```diff
@@ -424,13 +398,19 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+            events_len.try_to_target_usize(this.tcx.tcx).unwrap().try_into().unwrap();
+
         // Sanity-check to ensure that all event info is up-to-date.
         if cfg!(debug_assertions) {
-            for (key, interest) in interests.iter() {
+            for (key, interest) in epfd.watcher.interests().iter() {
                 // Ensure this matches the latest readiness of this FD.
```
RalfJung: The sanity check should be inside `get_ready_interests`

src/shims/unix/linux_like/epoll.rs
```diff
@@ -439,26 +419,17 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
-        // to the end of the queue during the loop without having them show up in the array.
-        let ready_events_len = u64::try_from(epfd.watcher.ready_count()).unwrap();
-        while let Some((idx, slot)) = array_iter.next(this)?
-            && idx < ready_events_len
-            && let Some((key, interest)) = epfd.watcher.next_ready()
+        // We get up to the first `max_events_num` ready events from the
+        // watcher and fill them into the slots of the array.
+        let mut ready_interests_iter = epfd.watcher.get_ready_interests(max_events_num).into_iter();
+        while let Some((_idx, slot)) = array_iter.next(this)?
+            && let Some(interest) = ready_interests_iter.next()
```
RalfJung: This should just iterate over `ready_interests_iter`, and *assert* that it can actually store all those elements in the array.

src/shims/readiness.rs
```diff
@@ -247,28 +258,52 @@ impl ReadinessWatcher {
...
+        interests
+    }
+
+    /// Destroy the watcher instance.
+    ///
+    /// This also deregisters all interests of the watcher
+    /// from the global readiness interest table.
+    pub fn destroy<'tcx>(self, ecx: &mut MiriInterpCx<'tcx>) {
+        let mut ids = self.interests.borrow().keys().map(|(id, _num)| *id).collect::<Vec<_>>();
+        ids.dedup();
```
RalfJung, line 324: You lost a comment here.

src/shims/readiness.rs
```diff
@@ -0,0 +1,509 @@
...
+
+/// The table with all [`ReadinessWatcher`]s.
+/// This tracks, for each file description, which watchers have an interest in events
+/// for this file description.
+pub struct ReadinessInterestTable {
+    /// The id of the next [`ReadinessWatcher`] created through
+    /// [`ReadinessInterestTable::new_watcher`].
+    next_watcher_id: usize,
+    /// Maps each file description (identified by its ID) to the list of watchers that are
+    /// interested in that FD.
```
RalfJung:
```suggestion
    /// interested in that FD. We also store the ReadinessWatcher's ID
    /// separately so we can access it without calling `upgrade`. The `Vec`
    /// is sorted by that ID.
```

src/shims/readiness.rs
```diff
@@ -0,0 +1,509 @@
...
+}
+
+/// The table with all [`ReadinessWatcher`]s.
+/// This tracks, for each file description, which watchers have an interest in events
+/// for this file description.
+pub struct ReadinessInterestTable {
+    /// The id of the next [`ReadinessWatcher`] created through
+    /// [`ReadinessInterestTable::new_watcher`].
+    next_watcher_id: usize,
+    /// Maps each file description (identified by its ID) to the list of watchers that are
```
RalfJung:
```suggestion
    /// Maps each file description (identified by its `FdId`) to the list of watchers that are
```

RalfJung (commented):
I still need to also compare the new logic with the old one which is hard to do on Github... but here are some things I noticed while reading the latest diff.

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5098/97eec7cf0c7ba44683e718869038f53a4b0931fb..92397f8a9befc212ee7d2205f150dcf1252f1ce2)*

src/shims/readiness.rs
```diff
@@ -0,0 +1,529 @@
...
+        )?;
+
+        interp_ok(Some(()))
+    }
+
+    /// Remove the interest registered for `key`.
+    ///
+    /// This function returns [`None`] when no interest is registered
+    /// for the specified `key`.
+    pub fn remove_interest<'tcx>(
```
RalfJung, line 218: This logic looks quite different from the old epoll_ctl_del. Why?
(Especially after I asked you to point out every place where you changed the logic.)

This PR would be a lot easier to review if you just copy-paste the existing logic without changing it. (You also reworded some of the comments, which isn't problematic but also means that diffing shows spurious changes -- and it makes me wonder how it even happened since clearly it's not just copy-paste somehow?)

src/shims/readiness.rs
```diff
@@ -0,0 +1,529 @@
...
+            ecx.machine.readiness_interests.remove(key.0, self);
+        }
+
+        Some(())
+    }
+
+    /// Add the thread with id `thread_id` to the queue of
+    /// blocked threads which will be unblocked when the
+    /// watcher becomes ready.
+    pub fn add_thread(&self, thread_id: ThreadId) {
```
RalfJung: This function name is very vague. Maybe call it `add_blocked_thread` or so?

src/shims/readiness.rs
```diff
@@ -0,0 +1,529 @@
...
+                // This is a level-triggered interest, so we need to re-add the event
+                // at the end of the ready queue like Linux does with epoll:
+                // <https://github.com/torvalds/linux/blob/HEAD/fs/eventpoll.c#L1835-L1847>
+                ready.push_back(next);
+            }
+
+            interests.push(interest);
+        }
+
+        // Sanity-check to ensure that all event info is up-to-date.
```
RalfJung, line 281: Again you reorganized the code. This used to be before the loop. Please move it back up, I don't want to have to think through whether there's any difference between doing the check before and after... actually there probably is, with your PR it fails to actually sanity-check the events that just got removed from the queue.

src/shims/readiness.rs
```diff
@@ -0,0 +1,529 @@
...
+        let mut ready = self.ready.borrow_mut();
+        let count = count.min(ready.len());
+        let mut interests = Vec::with_capacity(count);
+
+        let mut i = 0;
+        while i < count
+            && let Some(next) = ready.pop_front()
+        {
+            i = i.strict_add(1);
+            let interest = Ref::map(self.interests.borrow(), |interests| {
```
RalfJung: Why are you re-`borrow`ing in each loop iteration? That wasn't necessary in the old code.

This should be almost exactly like the old `return_ready_list`.

src/shims/unix/linux_like/epoll.rs
```diff
@@ -144,17 +35,14 @@ impl FileDescription for Epoll {
...
         ecx: &mut MiriInterpCx<'tcx>,
     ) -> InterpResult<'tcx, io::Result<()>> {
-        // If we were interested in some FDs, we can remove that now.
-        let mut ids = self.interest_list.get_mut().keys().map(|(id, _num)| *id).collect::<Vec<_>>();
-        ids.dedup(); // they come out of the map sorted
-        for id in ids {
-            ecx.machine.epoll_interests.remove(id, self_id);
-        }
+        let watcher = Rc::into_inner(self.watcher)
+            .expect("Epoll instance should contain the only strong reference to the watcher");
```
RalfJung, line 44: There's also a strong reference in the `BlockedReason`. So... does that mean we'll ICE if an epoll FD gets closed while another thread is waiting on it?

src/shims/readiness.rs
```diff
@@ -0,0 +1,529 @@
...
+
+    /// Get all watchers which have a registered interest in the file description
+    /// with id `fd_id`.
+    fn get_watchers_for_fd(&mut self, fd_id: &FdId) -> Option<Vec<Rc<ReadinessWatcher>>> {
+        let watchers = self.interests.get_mut(fd_id)?;
+        Some(
+            watchers
+                .iter()
+                .map(|(_id, watcher)| {
+                    watcher.upgrade().expect("watcher has not been removed correctly")
```
RalfJung:
```suggestion
                    watcher.upgrade().expect("someone forgot to remove the garbage from `machine.readiness_interests`")
```

src/shims/readiness.rs
```diff
@@ -0,0 +1,529 @@
...
+        &mut self,
+        fd: DynFileDescriptionRef,
+        force_edge: bool,
+    ) -> InterpResult<'tcx> {
+        let this = self.eval_context_mut();
+        let fd_id = fd.id();
+
+        let Some(watchers) = this.machine.readiness_interests.get_watchers_for_fd(&fd_id) else {
+            return interp_ok(());
+        };
```
RalfJung, line 460: There's more logic changes here compared to the old code... it's fine, I have checked it now, but I wish you would have saved me that work.

RalfJung (commented):
Unfortunately comparing this with the old logic was a lot more work than it should have been since you introduced a bunch of unnecessary changes. And at least one of them is problematic I think. I left more comments.
@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5098/97eec7cf0c7ba44683e718869038f53a4b0931fb..fcaf0fac66ccf305092abb798ee2cb906a3a484e)*

src/shims/readiness.rs
WhySoBad: My idea was to not clone every `ReadinessInterest` which is returned when we would be fine with just having references to the interests. So re-borrowing the interests list a few times such that the borrows can be mapped to references of the individual interests made sense in my opinion.
And yes, this wasn't necessary in the old code, but only because we didn't need to return the interests which we got from `interest_list.get_mut(&key)`.  

If you want, I can change the code to clone the interests such that we don't need to re-borrow the interests multiple times and such that we can return a `Vec<ReadinessInterest>` instead.

src/shims/unix/linux_like/epoll.rs
WhySoBad: That's what would happen at the moment.

I see the following two solutions for solving this: 
- We store a strong reference in `BlockReason` and accept that the `Rc::into_inner` could fail
- We store a `(usize, Weak<ReadinessWatcher>)` in the `BlockReason` where the `usize` is the id of the watcher and then manually implement `PartialEq, Eq` for `BlockReason`. 

Whilst the second option would allow us to have the only strong reference to the watcher in the `Epoll` instance, it would also mean that we need to expose the watcher's id outside the readiness module which I originally didn't want to do.
What do you prefer?

src/shims/readiness.rs
WhySoBad: In my opinion there is no logical difference between what the `remove_interest` function does and what happened in the old `epoll_ctl_del` branch. Except maybe that the `remove_interest` function returns an `Option<()>` instead of an `InterpResult<'tcx, Scalar>`? But I'd argue that this generic API should not set errno, and instead that should be done by the caller like how it's implemented right now. 

The other differences are syntactical, namely that the boolean condition is separated out into the `is_last` variable, which I did because I found it easier to understand what's going on, and that the comment for the same if-statement is moved into the if-block. I did that because usually we have the comment about what we do when the condition is true inside the if-block.

Let me know if I should revert any of those changes.

RalfJung: Yes it is all those syntactical differences I am referring to. I *could* spend a bunch of time to figure out why none of them actually change the behavior, but that makes reviewing this PR a lot harder than it has to be. It's a bad idea to combine code movement and such syntactic changes in a single PR.

src/shims/readiness.rs
RalfJung: Please change the code to be syntactically as close as possible to the old `return_ready_list`.

src/shims/unix/linux_like/epoll.rs
RalfJung: Since we just need `Eq` (not `Ord`), it should also be okay to have only `Weak<...>` and compare those.

But... what do we do if we try to `upgrade` that reference and that fails? On a real system, I assume that threads get woken up even if the epoll they are waiting on gets closed after they got blocked?

WhySoBad: > `Weak<...>` and compare those.

I mainly wanted the id in the tuple such that we can still compare watchers whose `upgrade` failed. Otherwise, should we just treat all non-upgradable watchers as being equal?

> On a real system, I assume that threads get woken up even if the epoll they are waiting on gets closed after they got blocked?

I just tested it on my host and there the thread which called `epoll_wait` remained blocked, even after closing the epfd.

RalfJung: >  I just tested it on my host and there the thread which called epoll_wait remained blocked, even after closing the epfd.

What I meant is, they get woken up when one of the sources that was registered for their epoll is ready.

But in Miri if we end up calling `is_thread_blocked_on_host` after the epoll got closed, we'd panic trying to upgrade that `Weak`. Ergo it has to be a full `Rc`. But that means if this is the last `Rc` we drop we have to do the destructor logic I think? I think current Miri has a bug here: if we close an epoll while a thread blocks, we will *not* execute `FileDescription::destroy` as it's not the last ref. Later the BlockReason gets destroyed and then we actually decrement the refcount to 0, but we don't call `Epoll::destroy` there either, so that cleanup logic never runs... which I think can lead to panics later during the execution.

WhySoBad: > What I meant is, they get woken up when one of the sources that was registered for their epoll is ready.

I don't fully understand what you mean here. If the watcher is destroyed (which it needs to be in order for the `upgrade` to fail) all it's interests are removed. This means the epoll will never become ready?

> But in Miri if we end up calling `is_thread_blocked_on_host` after the epoll got closed, we'd panic trying to upgrade that `Weak`. Ergo it has to be a full `Rc`. But that means if this is the last `Rc` we drop we have to do the destructor logic I think?

The whole reason we have `is_thread_blocked_on_host` is to check whether a thread might get unblocked due to a host readiness event. So assuming we store a `Weak` watcher in the `BlockReason` and it cannot be upgraded, we know that this thread won't be unblocked due to a readiness event because the watcher has already been destroyed, and thus we can just return `false` from `is_thread_blocked_on_host`, no?

> Otherwise, should we just treat all non-upgradable watchers as being equal?

Also about this, I think we should just always return `false` from the `cmp` implementation in that case. This would mean we wouldn't be able to unblock any thread whose watcher has already been destroyed with `UnblockKind::Ready` -- which is essentially what we expect to happen; threads which are blocked on destroyed watchers can at most be unblocked due to the deadline being exceeded.

******

 So my suggestion would now be:
- Store a `Weak<ReadinessWatcher>` in the `BlockReason`
- Manually implement `PartialEq` and `Eq` for `BlockReason`; if a watcher cannot be upgraded, it's incomparable and we always return `false`, otherwise we use the `PartialEq` implementation from the `ReadinessWatcher` (I'd like to keep that, such that we don't need to expose the watcher id outside the readiness module)
- `is_blocked_on_host` returns `false` when the `Weak<...>` cannot be upgraded

RalfJung: > I don't fully understand what you mean here. If the watcher is destroyed (which it needs to be in order for the upgrade to fail) all it's interests are removed. This means the epoll will never become ready?

That sentence was talking about native behavior. I expect the kernel just keeps a reference to the epoll internally while the thread is blocked so closing the epoll does not actually destroy it inside the kernel, and the thread gets woken up as usual.

All of your arguments seem to be based on the assumption that a thread blocked on an epoll that is  closed while the thread is blocked will *never* wake up.

Please write a test to check what actually happens.

RalfJung: All of this complexity around `BlockReason` arises solely because we want to figure out whether it's worth sleeping and waiting for I/O indefinitely or not. Getting that check precisely right is just not worth all this effort. I think we should change `is_thread_blocked_on_host` to `is_thread_maybe_blocked_on_host` and return true for all socket/epoll blocks. Then we can remove the `epfd` from the `BlockReason` and simplify a bunch of stuff.

If we want to be fancy, we could check whether *there exists* an epoll that is waiting for the host, and only then return `true` from `is_thread_maybe_blocked_on_host`. But this slight increase in our ability to detect deadlocks is not worth tracking the *exact* epoll.

WhySoBad: > Please write a test to check what actually happens.

I've now written this test:
```rust
fn waiting_threads_unblocked_after_epoll_close() {
    // Create an epoll instance.
    let epfd = errno_result(unsafe { libc::epoll_create1(0) }).unwrap();

    // Create a socketpair instance.
    let mut fds = [-1, -1];
    errno_check(unsafe { libc::socketpair(libc::AF_UNIX, libc::SOCK_STREAM, 0, fds.as_mut_ptr()) });

    // Add `fds[0]` to epoll with readable interest.
    epoll_ctl_add(epfd, fds[0], EPOLLIN | EPOLLET_OR_ZERO).unwrap();

    let t1 = thread::spawn(move || {
        // Close epoll file descriptor.
        unsafe {
            errno_check(libc::close(epfd));
        }

        // Write some data into `fds[1]` which should make `fds[0]` readable.
        write_all(fds[1], b"abcde").unwrap();
    });

    // Indefinitely block until `fds[0]` becomes readable.
    check_epoll_wait(epfd, &[Ev { events: EPOLLIN, data: fds[0] }], -1);

    t1.join().unwrap();
}
```
And it succeeds with the `--native` flag, so your assumption is correct. I've noticed that my test from earlier today had a tiny flaw (maybe I should've written in Rust instead of C x.x).

This means we'll indeed need to store a strong reference to the watcher in the `BlockReason`.
However, I've realized that this wasn't the source of our problem: As long as the callback of the blocked thread also stores a strong reference to the epfd the file description wouldn't be destroyed anyways.
Currently, my only idea on how we could resolve this problem, is to use `Rc::strong_count` after the `epoll_wait` thread is unblocked to check whether the `epfd` has exactly two strong counts. One should be from the `epfd` stored in the `FdTable` and one should be from the `epfd` instance which is stored in the callback. If that's the case, we could drop the callback's `epfd` instance, and manually invoke a `close` on the `FdTable` (there should be a new public method for that, such that we don't need to use `OpTy`s).  

**Edit**: I've only saw your previous message now. However, what I wrote is still mostly valid: The problem is not that we store an epfd in the `BlockReason` (the new implementation would anyways only store a watcher), the problem is that we store a strong reference in the callback (this could potentially also be problematic for other FDs).

RalfJung: Ah, good point about the closures. However, manual arithmetic on the strong count sounds way too fragile.

This is a pre-existing issue. So I'd say let's finish this PR without worrying about it.

src/shims/readiness.rs
WhySoBad: I've realized that what we did is basically a manual implementation of `VecDeque::drain`, so I decided to iterate over the drained iterator instead of using the `while let Some(...)` loop.
Otherwise, I've tried to make everything as close as possible to the old `return_ready_list` implementation.

RalfJung: No, please don't do any logic changes (as in, any changes to the code at all other than making it compile in the new file with the new names). You're just making this code harder to review. I already said this multiple times, I don't know how much more clear I can make myself.

src/shims/readiness.rs
```diff
@@ -287,23 +288,21 @@ impl ReadinessWatcher {
...
-        {
-            i = i.strict_add(1);
-            let interest = Ref::map(self.interests.borrow(), |interests| {
-                interests.get(&next).expect("non-existing interest in ready set")
-            });
+        // We iterate over the first `count` entries from the `ready` queue
+        // and push the associated interests into the `ready_interests` list.
+        // If an interest is level-triggered, it's also added to the back of
+        // the queue again.
+        for key in ready.drain(0..count).collect::<Vec<_>>() {
```
RalfJung: Was there a `drain` in the original code? No. So why does moving the code suddenly introduce a `drain`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5098/97eec7cf0c7ba44683e718869038f53a4b0931fb..ef031538b8ac9022ea25a17845f4638d9fdea6b7)*

WhySoBad: Sorry, I thought when this change is the entire content of a commit a logic change should be fine. I've now reverted it back to a `while` loop.

src/shims/readiness.rs
RalfJung: The above comment was meant to say that yes please make it closer to the original code. I thought that was quite clear. I am now doing this myself...

RalfJung: Your new code had some unnecessary borrow-unborrow-borrow-unborrow of a RefCell. That would have been quite obvious if you had split the code movement from all the other unrelated refactor.

PR #5102, opened 2026-06-10
Mask CRC intrinsic inputs to match LoongArch hardware semantics

heiher:
LoongArch CRC b/h/w instructions only consume the low 8/16/32 bits of their input operand, ignoring any higher bits. However, Miri forwarded the full value to compute_crc32, which expects all bits above the specified width to be zero and may panic otherwise.

Mask off the unused high bits before calling compute_crc32 so the shim matches hardware behavior and correctly handles sign-extended intrinsic arguments.

### DIFF
--- a/src/shims/loongarch.rs
+++ b/src/shims/loongarch.rs
@@ -1,4 +1,4 @@
-use rustc_abi::CanonAbi;
+use rustc_abi::{CanonAbi, Size};
 use rustc_middle::ty::Ty;
 use rustc_span::Symbol;
 use rustc_target::callconv::FnAbi;
@@ -55,10 +55,17 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // b/h/w variants and i64 for the d variant, per the LLVM intrinsic
                 // definitions.
                 // https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/IR/IntrinsicsLoongArch.td
-                // If the higher bits are non-zero, `compute_crc32` will panic. We should probably
-                // raise a proper error instead, but outside stdarch nobody can trigger this anyway.
+                // LoongArch CRC b/h/w instructions ignore any bits above `bit_size`.
+                // https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html#crc-check-instructions
+                // Miri's `compute_crc32` requires all higher bits to be zero and may
+                // panic otherwise, so we explicitly mask them off here to reproduce the
+                // hardware behavior.
                 let crc = crc.to_u32()?;
-                let data = if bit_size == 64 { data.to_u64()? } else { u64::from(data.to_u32()?) };
+                let data = if bit_size == 64 {
+                    data.to_u64()?
+                } else {
+                    Size::from_bits(bit_size).truncate(data.to_u32()?.into()).try_into().unwrap()
+                };
 
                 let result = compute_crc32(crc, data, bit_size, polynomial);
                 this.write_scalar(Scalar::from_u32(result), dest)?;
--- a/tests/pass/shims/loongarch/intrinsics-loongarch64-crc.rs
+++ b/tests/pass/shims/loongarch/intrinsics-loongarch64-crc.rs
@@ -1,8 +1,19 @@
 // We're testing loongarch64-specific intrinsics
 //@only-target: loongarch64
-#![feature(stdarch_loongarch)]
+#![feature(abi_unadjusted, link_llvm_intrinsics, stdarch_loongarch)]
 
 use std::arch::loongarch64::*;
+
+unsafe extern "unadjusted" {
+    #[link_name = "llvm.loongarch.crc.w.b.w"]
+    fn _crc_w_b_w(a: i32, b: i32) -> i32;
+    #[link_name = "llvm.loongarch.crc.w.h.w"]
+    fn _crc_w_h_w(a: i32, b: i32) -> i32;
+    #[link_name = "llvm.loongarch.crcc.w.b.w"]
+    fn _crcc_w_b_w(a: i32, b: i32) -> i32;
+    #[link_name = "llvm.loongarch.crcc.w.h.w"]
+    fn _crcc_w_h_w(a: i32, b: i32) -> i32;
+}
 
 fn main() {
     test_crc_ieee();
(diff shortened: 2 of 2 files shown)


tests/pass/shims/loongarch/intrinsics-loongarch64-crc.rs
```diff
@@ -14,16 +14,19 @@ fn test_crc_ieee() {
     assert_eq!(crc_w_b_w(0x01, 0x00000000), 0x77073096);
     assert_eq!(crc_w_b_w(0x61, 0xffffffff_u32 as i32), 0x174841bc);
     assert_eq!(crc_w_b_w(0x2a, 0x2aa1e72b), 0x772d9171);
+    assert_eq!(crc_w_b_w(0xff_u8 as i8, 0x00000000), 0x2d02ef8d);
```
RalfJung: Hoes does this test make any of the higher bits non-0? 0xff still fits in a single byte.

How does this even compile? Doesn't the intrinsic take `i32` arguments?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5102/885c5408d71ae442a5e8d3f5dbe357c374b37a76..71c7ff11228cd049c91457078c40ca383d40d0dc)*

src/shims/loongarch.rs
```diff
@@ -55,10 +55,16 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                // LoongArch CRC b/h/w instructions ignore any bits above `bit_size`.
+                // Miri's `compute_crc32` requires all higher bits to be zero and may
+                // panic otherwise, so we explicitly mask them off here to reproduce the
+                // hardware behavior.
                 let crc = crc.to_u32()?;
-                let data = if bit_size == 64 { data.to_u64()? } else { u64::from(data.to_u32()?) };
+                let data = if bit_size == 64 {
+                    data.to_u64()?
+                } else {
+                    u64::from(data.to_u32()?) & (1u64.strict_shl(bit_size).wrapping_sub(1))
```
RalfJung: Is there a way to write this logic that's more readable? This bit masking here is very hard to follow.
For instance, maybe something like `Size::from_bits(bit_size).truncate(data)`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5102/885c5408d71ae442a5e8d3f5dbe357c374b37a76..71c7ff11228cd049c91457078c40ca383d40d0dc)*

src/shims/loongarch.rs
```diff
@@ -55,10 +55,16 @@ pub(super) trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // b/h/w variants and i64 for the d variant, per the LLVM intrinsic
                 // definitions.
                 // https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/IR/IntrinsicsLoongArch.td
-                // If the higher bits are non-zero, `compute_crc32` will panic. We should probably
-                // raise a proper error instead, but outside stdarch nobody can trigger this anyway.
+                // LoongArch CRC b/h/w instructions ignore any bits above `bit_size`.
```
RalfJung, line 58: Is there some documentation we can reference which confirms that the high bits are ignored?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5102/885c5408d71ae442a5e8d3f5dbe357c374b37a76..71c7ff11228cd049c91457078c40ca383d40d0dc)*

tests/pass/shims/loongarch/intrinsics-loongarch64-crc.rs
heiher: `0xff_u8 as i8` is simply `-1` as a signed value (writing `-1` directly may be clearer). When passed to `crc_w_b_w`, it is converted from `i8` to `i32`, which sign-extends the value and fills all upper bits with ones.

src/shims/loongarch.rs
heiher: `truncate` is still a nightly experimental API. I added a comment here to make the masking logic more readable.

src/shims/loongarch.rs
heiher: Yes. I've added it to the comment. Thanks!

https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html#crc-check-instructions

```
CRC.W.B.W rd, rj, rk:
    chksum = CRC32(GR[rk][31:0], GR[rj][7:0], 8, 0xEDB88320)
    GR[rd] = SignExtend(chksum, GRLEN)
```

src/shims/loongarch.rs
RalfJung: > truncate is still a nightly experimental API

No? It is a method on the compiler's `Size` type.
https://doc.rust-lang.org/nightly/nightly-rustc/rustc_abi/struct.Size.html#method.truncate

Also using nightly functions in Miri is entirely fine.

tests/pass/shims/loongarch/intrinsics-loongarch64-crc.rs
RalfJung: Ah, so this happens internally inside stdarch. And stdarch is changing this with https://github.com/rust-lang/stdarch/pull/2148.

Can you adjust the test to directly invoke the LLVM intrinsic (for the new cases -- leave the existing ones unchanged), so that we avoid relying on what stdarch does internally for our test coverage?

tests/pass/shims/loongarch/intrinsics-loongarch64-crc.rs
```diff
@@ -14,16 +14,19 @@ fn test_crc_ieee() {
     assert_eq!(crc_w_b_w(0x01, 0x00000000), 0x77073096);
     assert_eq!(crc_w_b_w(0x61, 0xffffffff_u32 as i32), 0x174841bc);
     assert_eq!(crc_w_b_w(0x2a, 0x2aa1e72b), 0x772d9171);
+    assert_eq!(crc_w_b_w(-1, 0x00000000), 0x2d02ef8d);
```
RalfJung: The test is supposed to demonstrate "high bits are ignored", so please make it so that this uses the same input as one of the cases above, only with different high bits, so one can immediately see that the result is the same.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5102/73cba51222cc836823795db8cd6a9d46d969d494..a96d9bee101002d69361febbe301e577039c9b4b)*

RalfJung:
Thanks! LGTM. Would be nice if you could run this test on real hardware just to be sure.

heiher:
> Thanks! LGTM. Would be nice if you could run this test on real hardware just to be sure.

Thanks! Confirmed on real hardware, the test passes there as well.

PR #5103, opened 2026-06-11
add borrow model test for mixed-mutability static

RalfJung:
I actually don't understand why this is UB in Tree Borrows -- somehow "precise interior mutable" tracking seems to be kicking in?
This came up in https://github.com/rust-lang/rust/issues/156789

Cc @JoJoDeveloping


RalfJung:
Ah. The MIR has this:
```
_2 = const {alloc1: &(i32, Atomic<i32>)};
```
With our new "implicit retag during assignment", this shared reference gets reborrowed and (with precise UnsafeCell tracking) it becomes immutable.

I think this was not UB before the "implicit retag" PR.
Seems useful to have this test so we notice if this ever changes.

PR #5106, opened 2026-06-13
Don't clear TCP socket readable readiness when read end is already closed

WhySoBad:
This pull request fixes the problems experienced in https://github.com/tokio-rs/tokio/pull/8205.

In the previous implementation we always removed the readable readiness for sockets after a short read -- even when the read end has been closed. When we still have some data in the read buffer whilst the read end gets closed, this could lead to the situation where we have a short read and thus remove the readable readiness before reading zero (indicating EOF).

During debugging I also noticed that we only set the `readable` readiness on `shutdown(SHUT_RDWR)`, however, it should also be set on `shutdown(SHUT_RD)`.

### DIFF
--- a/src/shims/unix/socket.rs
+++ b/src/shims/unix/socket.rs
@@ -1487,9 +1487,9 @@ pub trait EvalContextExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
         // Only shutting down the write end doesn't cause an EPOLLHUP event
         // and thus we won't set the `write_closed` readiness for it here.
         readiness.write_closed |= is_read_write_shutdown;
-        // The Linux kernel also sets EPOLLIN when both ends of a socket are closed:
+        // The Linux kernel also sets EPOLLIN when the read end of a socket is closed:
         // <https://github.com/torvalds/linux/blob/HEAD/net/ipv4/tcp.c#L584-L588>
-        readiness.readable |= is_read_write_shutdown;
+        readiness.readable |= is_read_shutdown || is_read_write_shutdown;
 
         drop(readiness);
 
@@ -1697,12 +1697,16 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     ) -> InterpResult<'tcx, Result<usize, IoError>> {
         let this = self.eval_context_mut();
 
-        let SocketState::Connected(stream) = &mut *socket.state.borrow_mut() else {
+        let mut state = socket.state.borrow_mut();
+        let SocketState::Connected(stream) = &mut *state else {
             panic!("try_non_block_send must only be called when the socket is connected")
         };
 
         // This is a *non-blocking* write.
         let result = this.write_to_host(stream, length, buffer_ptr)?;
+
+        drop(state);
+
         match result {
             Err(IoError::HostError(e))
                 if matches!(e.kind(), io::ErrorKind::NotConnected | io::ErrorKind::WouldBlock) =>
@@ -1715,8 +1719,13 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // would be returned on UNIX-like systems. We thus remap this error to an EWOULDBLOCK.
                 interp_ok(Err(IoError::HostError(io::ErrorKind::WouldBlock.into())))
             }
-            Ok(bytes_written) if bytes_written < length => {
-                // We had a short write. On Unix hosts using the `epoll` and `kqueue` backends, a
+            Ok(bytes_written)
+                if bytes_written < length && !socket.io_readiness.borrow().write_closed =>
+            {
+                // We had a short write. (Note that we don't want to clear the writable readiness for
+                // sockets whose write end has already been closed as those never block a write, i.e.,
+                // they are always write-ready.)
+                // On Unix hosts using the `epoll` and `kqueue` backends, a
                 // short write means that the write buffer is full. We update the readiness
                 // accordingly, which means that next time we see "writable" we will report an epoll
                 // edge. Some applications (e.g. tokio) rely on this behavior; see
@@ -1820,7 +1829,8 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
     ) -> InterpResult<'tcx, Result<usize, IoError>> {
         let this = self.eval_context_mut();
 
-        let SocketState::Connected(stream) = &mut *socket.state.borrow_mut() else {
+        let mut state = socket.state.borrow_mut();
+        let SocketState::Connected(stream) = &mut *state else {
             panic!("try_non_block_recv must only be called when the socket is connected")
         };
 
@@ -1832,6 +1842,9 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
             length,
             buffer_ptr,
         )?;
+
+        drop(state);
+
         match result {
             Err(IoError::HostError(e))
                 if matches!(e.kind(), io::ErrorKind::NotConnected | io::ErrorKind::WouldBlock) =>
@@ -1844,9 +1857,16 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
                 // would be returned on UNIX-like systems. We thus remap this error to an EWOULDBLOCK.
                 interp_ok(Err(IoError::HostError(io::ErrorKind::WouldBlock.into())))
             }
-            Ok(bytes_read) if !should_peek && bytes_read < length && bytes_read > 0 => {
+            Ok(bytes_read)
+                if !should_peek
+                    && bytes_read < length
+                    && bytes_read > 0
+                    && !socket.io_readiness.borrow().read_closed =>
+            {
                 // We had a short read (and were not peeking). (Note that reading 0 bytes is guaranteed
-                // to indicate EOF, and can never happen spuriously, so we have to exclude that case.)
+                // to indicate EOF, and can never happen spuriously, so we have to exclude that case.
+                // We also don't want to clear the readable readiness for sockets whose read end has
+                // already been closed as those never block a read, i.e., they are always read-ready.)
                 // On Unix hosts using the `epoll` and `kqueue` backends, a short read means that the
                 // read buffer is empty. We update the readiness accordingly, which means that next time
                 // we see "readable" we will report an epoll edge. Some applications (e.g. tokio) rely on
--- a/tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
+++ b/tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
@@ -1,8 +1,5 @@
 //@only-target: linux android illumos
 //@compile-flags: -Zmiri-disable-isolation
-//@revisions: windows_host unix_host
-//@[unix_host] ignore-host: windows
-//@[windows_host] only-host: windows
 //@run-native
 
 #![feature(io_error_inprogress)]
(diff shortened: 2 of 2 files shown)


src/shims/unix/socket.rs
```diff
@@ -1844,9 +1852,16 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
+                if !should_peek
+                    && bytes_read < length
+                    && bytes_read > 0
+                    && !socket.io_readiness.borrow().read_closed =>
+            {
                 // We had a short read (and were not peeking). (Note that reading 0 bytes is guaranteed
-                // to indicate EOF, and can never happen spuriously, so we have to exclude that case.)
+                // to indicate EOF, and can never happen spuriously, so we have to exclude that case.
+                // We also don't want to clear the readable readiness for sockets whose read end has
+                // already been closed as this would prevent us from reading the EOF.)
```
```suggestion
                // already been closed as those never block on a read, i.e., they are always read-ready.)
```

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -514,6 +515,102 @@ fn test_readiness_after_short_read() {
...
+    // We had a short read because `buffer` is bigger than `TEST_BYTES`.
+    // Because the read end of the socket is closed, we should still be able to
+    // read to detect EOFs.
+
+    let mut buffer = [1u8; 16];
+    let bytes_read = unsafe {
+        errno_result(libc::read(client_sockfd, buffer.as_mut_ptr().cast(), buffer.len())).unwrap()
+    };
+    // The read should return 0, indicating EOF.
+    assert_eq!(bytes_read, 0);
```
RalfJung, line 711: Please move this *after* the `current_epoll_readiness` check.

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -514,6 +515,102 @@ fn test_readiness_after_short_read() {
...
+
+    // Ensure that the "readable" and "read closed" readiness flags
+    // are still set.
+    assert_eq!(
+        current_epoll_readiness::<8>(client_sockfd, EPOLLIN | EPOLLET | EPOLLRDHUP),
```
RalfJung, line 702: Note for a future PR: this should probably also not take a const generic argument any  more.

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -514,6 +515,102 @@ fn test_readiness_after_short_read() {
...
+    net::connect_ipv4(client_sockfd, addr).unwrap();
+
+    unsafe {
+        // Change client socket to be non-blocking.
+        errno_check(libc::fcntl(client_sockfd, libc::F_SETFL, libc::O_NONBLOCK));
+    }
+
+    server_thread.join().unwrap();
+
+    // Update client socket interest to also include readable interest.
```
RalfJung: Why are you first registering this with one interest and later updating the interest?

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -514,6 +515,102 @@ fn test_readiness_after_short_read() {
...
+        // Shutting down the write end of the peer socket should close
+        // the read end of the client socket.
+        unsafe {
+            errno_check(libc::shutdown(peerfd, libc::SHUT_WR));
+        }
+
+        // Add client socket with "read closed" interest to epoll.
+        epoll_ctl_add(epfd, client_sockfd, EPOLLET | EPOLLRDHUP).unwrap();
+        // Wait until the read end of the client socket is closed.
+        check_epoll_wait(epfd, &[Ev { events: EPOLLRDHUP, data: client_sockfd }], -1);
```
RalfJung: Where do we close the read end, to trigger this event...?

RalfJung (commented):
Nice catch, thanks!

Should something similar happen for writes? If the write end of a socket is closed, then presumably writes will never block, so the socket is always writable?

@rustbot author

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5106/28b3a0aad5cc0fc7d2fbaf480f50db823f7fe31d..abbd1796bca00f1a7a697f5b780d8d9f1c4be67b)*

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
WhySoBad: It was closed implicitly by closing the write end of the peer socket. I've now changed it to closing the read end of the client socket instead.

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
WhySoBad: Should we just use a big enough default value for `N` (e.g. 128) which could still be manually overwritten as a const generic argument?

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
RalfJung: Ah, good point, there are 2 ways to close either side of the socket... it may be worth testing both of them.

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
RalfJung: A hard-coded buffer size of 1 should always work, no? We are just adding a single FD to that epoll anyway.

src/shims/unix/socket.rs
```diff
@@ -1715,8 +1719,13 @@ trait EvalContextPrivExt<'tcx>: crate::MiriInterpCxExt<'tcx> {
...
                 interp_ok(Err(IoError::HostError(io::ErrorKind::WouldBlock.into())))
             }
-            Ok(bytes_written) if bytes_written < length => {
-                // We had a short write. On Unix hosts using the `epoll` and `kqueue` backends, a
+            Ok(bytes_written)
+                if bytes_written < length && !socket.io_readiness.borrow().write_closed =>
+            {
+                // We had a short write. (Note that we don't want to clear the writable readiness for
+                // sockets whose write end has already been closed as those never block a write, i.e.,
+                // they are always write-ready -- even though the writes fail with EPIPE.)
```
```suggestion
                // they are always write-ready.)
```

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -25,8 +28,14 @@ fn main() {
     test_shutdown_read_write();
     test_shutdown_read();
     test_shutdown_write();
+    test_shutdown_write_full();
     test_readiness_after_short_read();
-    test_readiness_after_short_read_after_shutdown();
+    if !cfg!(windows_host) {
+        // Shutting down the read end of a socket sets
+        // EPOLLHUP on Windows hosts which isn't set on
+        // Unix hosts.
```
RalfJung: What's happening on Windows? This seems concerning.

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -431,6 +441,73 @@ fn test_shutdown_write() {
...
+
+    net::connect_ipv4(client_sockfd, addr).unwrap();
+
+    unsafe {
+        // Change client socket to be non-blocking.
+        errno_check(libc::fcntl(client_sockfd, libc::F_SETFL, libc::O_NONBLOCK));
+    }
+
+    // The peer socket is a blocking socket.
+    server_thread.join().unwrap();
```
RalfJung: I don't understand what this comment has to do with the code.

Also, why is this a separate thread anyway? Seems like you could just do it all in the main thread.

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -431,6 +441,73 @@ fn test_shutdown_write() {
...
+                // Instead of doing a short write to completely fill the buffer, they can return an
+                // EWOULDBLOCK when the next write wouldn't fit into the buffer.
+                // When we get such an error, we also assume the write buffer is full.
+                break;
+            }
+            Err(err) => panic!("unexpected error whilst filling up buffer: {err}"),
+        }
+    }
+
+    // The write buffer is full; the socket should no longer be writable.
```
```suggestion
    // The write buffer is full; the socket would now block on writes, i.e. readiness is 0.
```

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -431,6 +441,73 @@ fn test_shutdown_write() {
...
+    assert_eq!(current_epoll_readiness::<8>(client_sockfd, EPOLLOUT), 0);
+
+    // Close the socket write end.
+    unsafe {
+        errno_check(libc::shutdown(client_sockfd, libc::SHUT_WR));
+    }
+
+    // The socket should be "writable" again after its write end is closed.
```
```suggestion
    // The socket no longer blocks on writes after its write end is closed.
```

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -431,6 +441,73 @@ fn test_shutdown_write() {
...
+    check_epoll_wait(epfd, &[Ev { events: EPOLLOUT, data: client_sockfd }], -1);
```
RalfJung: The code is quite confusing here... a bit up you used `current_epoll_readiness` but now you're back to using the old `epfd`?

Also why doesn't this signal "write closed"?

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -514,6 +591,90 @@ fn test_readiness_after_short_read() {
...
+    assert_eq!(
+        current_epoll_readiness::<8>(client_sockfd, EPOLLIN | EPOLLET | EPOLLRDHUP),
+        EPOLLIN | EPOLLRDHUP
+    );
+
+    let mut buffer = [1u8; 16];
+    let bytes_read = unsafe {
+        errno_result(libc::read(client_sockfd, buffer.as_mut_ptr().cast(), buffer.len())).unwrap()
+    };
+    // The read should return 0, indicating EOF.
```
```suggestion
    // The read should not block and return 0, indicating EOF.
```

RalfJung (commented):
I am afraid I am still very confused by the tests.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5106/28b3a0aad5cc0fc7d2fbaf480f50db823f7fe31d..33296add1318993667334540ab17c46522ca5609)*

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
WhySoBad: > a bit up you used current_epoll_readiness but now you're back to using the old epfd

I did that because `epfd` was edge-triggered. I've now changed it to be level-triggered and then we can remove the `current_epoll_readiness` invocation. 

> Also why doesn't this signal "write closed"?

"write closed" isn't a real epoll readiness. I sometimes call EPOLLHUP "write closed" (because mio maps it like this) but in reality EPOLLHUP only gets set when both ends of the full-duplex connection are closed.

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
WhySoBad: This is one of the cases where the cross platform mapping of mio is wrong. I've tested the following minimal program in my CI pipeline:
```rust
#![allow(unused, unused_assignments)]

use std::error::Error;
use std::io::Write;

use mio::net::{TcpListener, TcpStream};
use mio::{Events, Interest, Poll, Token};

const SERVER: Token = Token(0);
const CLIENT: Token = Token(1);

fn main() -> Result<(), Box<dyn Error>> {
    let mut poll = Poll::new()?;
    let mut events = Events::with_capacity(128);

    let addr = "127.0.0.1:13265".parse()?;
    let mut server = TcpListener::bind(addr)?;
    poll.registry()
        .register(&mut server, SERVER, Interest::READABLE)?;

    let mut client = TcpStream::connect(addr)?;
    poll.registry()
        .register(&mut client, CLIENT, Interest::READABLE | Interest::WRITABLE)?;

    let mut peer_opt = None;

    loop {
        poll.poll(&mut events, None)?;

        for event in events.iter() {
            match event.token() {
                SERVER => {
                    let (mut peer, _) = server.accept().unwrap();
                    println!("before write");
                    peer.write_all(b"asdfasdfasdfasdf").unwrap();
                    println!("after write");
                    peer_opt = Some(peer);

                    println!("before shutdown");
                    client.shutdown(std::net::Shutdown::Read).expect("should shut down read end");
                    println!("after shutdown");
                }
                CLIENT => {
                    println!(
                        "readable={} writable={} read_closed={} write_closed={} error={}",
                        event.is_readable(),
                        event.is_writable(),
                        event.is_read_closed(),
                        event.is_write_closed(),
                        event.is_error()
                    );
                }
                _ => unreachable!(),
            }
        }
    }
}
```
and on Ubuntu I get:
```bash
before write
after write
before shutdown
after shutdown
readable=false writable=true read_closed=false write_closed=false error=false
readable=true writable=true read_closed=true write_closed=false error=false
```
whilst on Windows I get:
```bash
before write
after write
before shutdown
after shutdown
readable=false writable=true read_closed=false write_closed=false error=false
readable=true writable=true read_closed=true write_closed=true error=false
```

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
RalfJung: Everything else in this file is edge-triggered so not sure it's better to randomly mix in something level-triggered... at the very least, make sure this is well-commented.

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
RalfJung: That's a lot of code but no explanation. Please don't make me reverse engineer the code -- just explain what happens.

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
```diff
@@ -25,7 +28,14 @@ fn main() {
     test_shutdown_read_write();
     test_shutdown_read();
     test_shutdown_write();
+    test_shutdown_write_full();
```
RalfJung: The name of the test should better reflect what the test actually does.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/miri/5106/28b3a0aad5cc0fc7d2fbaf480f50db823f7fe31d..0697c40e21ca13c634cf4a57e262f97175ec5af3)*

tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
RalfJung: Most of the test should still work on Windows though, so let's run it and just check the slightly different expectations, to clearly document what happens -- and with a FIXME link to a mio issue.

WhySoBad: Turns out this is actually documented Windows behavior when there is still some data in the socket at the time of closing the read end:
"For TCP sockets, if there is still data queued on the socket waiting to be received, or data arrives subsequently, the connection is reset, since the data cannot be delivered to the user." (https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-shutdown)

I've now adjusted the test to also expect `EPOLLHUP` on Windows hosts.

RalfJung: So... can we avoid this by ensuring the read buffer is empty, and only then closing the read end?

RalfJung: Oh wait we need data in that buffer so that we can still do reads later, to get a short read, to trigger the test.

And there's no other way to get to the same situation? Like, what if we close the write end from the server side? Something like this?
```rust
diff --git a/tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs b/tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
index fd657db58..6de67b7eb 100644
--- a/tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
+++ b/tests/pass-dep/libc/libc-socket-no-blocking-epoll.rs
@@ -646,6 +646,9 @@ fn test_readable_after_read_shutdown_and_short_read() {
 
         // Write `TEST_BYTES` into the stream.
         libc_utils::write_all(peerfd, TEST_BYTES).unwrap();
+
+        // Then promise that we will do no more writes.
+        unsafe { errno_check(libc::shutdown(peerfd, libc::SHUT_WR)) };
     });
 
     net::connect_ipv4(client_sockfd, addr).unwrap();
@@ -657,11 +660,6 @@ fn test_readable_after_read_shutdown_and_short_read() {
 
     server_thread.join().unwrap();
 
-    // Close the read end of the client socket.
-    unsafe {
-        errno_check(libc::shutdown(client_sockfd, libc::SHUT_RD));
-    }
-
     // Add client socket with "read closed" and "readable" interest to epoll.
     epoll_ctl_add(epfd, client_sockfd, EPOLLET | EPOLLIN | EPOLLRDHUP).unwrap();
 
```

RalfJung: Yeah that seems to work. :)

I'll finish up the PR with this.

